PR #6436, opened 2021-08-16
Add Splashscreen api endpoint

daullmer:
**Changes**
Adds a splashscreen api endpoint.
The generated splashscreen uses posters and thumbs (if available). In case there are no thumbs (no Fanart plugin, no images uploaded, not images with the media files), backdrops are used, since they are pretty much available for every movie at TMDb.

**Sample**:
splashscreen-True

~~The darkening is up for debate. I think most clients can add this filter themselves (eg. with a CSS filter in web), but maybe some can't, so it might be helpful to generate it on the server.~~

I'm also open to changing the output path. The data directory may not be the right folder to save the generated splashscreen.

In the future, we could make the used libraries and maximum parental rating configurable. For now I allowed all items with a max. rating of NC-17 (mainly to filter out nsfw libraries).

**Issues**


daullmer:
If someone has a better name for this than splashscreen, I'm open to changing the name. Maybe "library wall" or something like this could work as well.

Jellyfin.Api/Controllers/SplashscreenController.cs
```diff
@@ -0,0 +1,105 @@
...
+        /// <param name="darken">Darken the generated image.</param>
+        /// <param name="regenerate">Whether to regenerate the image, regardless if one already exists.</param>
+        /// <returns>The splashscreen.</returns>
+        [HttpGet]
+        [ProducesResponseType(StatusCodes.Status200OK)]
+        [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+        [ProducesImageFile]
+        public ActionResult GetSplashscreen(
+            [FromQuery] bool? darken = false,
+            [FromQuery] bool? regenerate = false)
```
nielsvanvelzen, line 63: This is a bad idea - someone can just spam api calls to regenerate the image. Creating a scheduled task might be better.

Jellyfin.Api/Controllers/SplashscreenController.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        /// <summary>
+        /// Generates or gets the splashscreen.
+        /// </summary>
+        /// <param name="darken">Darken the generated image.</param>
+        /// <param name="regenerate">Whether to regenerate the image, regardless if one already exists.</param>
+        /// <returns>The splashscreen.</returns>
+        [HttpGet]
+        [ProducesResponseType(StatusCodes.Status200OK)]
+        [ProducesResponseType(StatusCodes.Status500InternalServerError)]
```
nielsvanvelzen, line 59: I don't think we add the 500 errors normally as they should not happen. Not sure though.

Jellyfin.Api/Controllers/SplashscreenController.cs
```diff
@@ -0,0 +1,105 @@
...
+        [ProducesImageFile]
+        public ActionResult GetSplashscreen(
+            [FromQuery] bool? darken = false,
+            [FromQuery] bool? regenerate = false)
+        {
+            var outputPath = Path.Combine(_appPaths.DataPath, $"splashscreen-{darken}.jpg");
```
nielsvanvelzen, line 65: webp might work better and should work on all of our clients.

Jellyfin.Api/Controllers/SplashscreenController.cs
MrTimscampi: Perhaps link it to library scans or something, so that it gets updated from time to time.

MrTimscampi:
Instead of serving the physical file directly, would it be possible to hook this up to the already existing image serving mechanism? (Which supports sending WebP if the client supports it, allows us to resize the image, etc)

Having the image's URL be part of the branding endpoint would also be a nice touch, so clients don't need to be updated if we change the endpoint at any time (Unlikely to happen, but it's also annoying to have raw URLs for stuff in clients, when all the resource URLs we use should be sent from the server :yum: )

In order to make this customizable for users, would it be possible to have a way to override it? Then we could have something like this in the server settings:
image

It'd make it easier for users to override it (Especially since we'll likely have to set it with Javascript, on web at least) and bring some more customization to the stock Jellyfin experience (No CSS required)

daullmer:
Thanks for your feedback! I have made the following changes:

- the new API endpoint is `/Branding/Splashscreen`
- The endpoint is now part of the usual processing pipeline. This allows for the usual image query parameters, and always returns the correct file format, depending on the `Accepts` header
- remove the `darken` api parameter as it is essentially the same as the `foregroundLayer` parameter
- generate splashscreen during refresh library scheduled task
- add post endpoint to upload a custom splashscreen
- add a new interface `IImageGenerator` to generate the splashscreen which could be used in the future for plugins to generate a different type of splashscreen (other images like genres and library images could be added as well in the future)
- The branding endpoint returns a new property called `SplashscreenUrl` with the url to the splashscreen ( don't know if it should return `/Branding/Splashscreen` or `Branding/Splashscreen`)

MediaBrowser.Model/Branding/BrandingOptions.cs
```diff
@@ -15,5 +18,21 @@ public class BrandingOptions
...
+        /// Only used to save the custom uploaded user splashscreen in the configuration file.
+        /// </remarks>
+        [JsonIgnore]
+        public string? SplashscreenLocation { get; set; }
+
+        /// <summary>
+        /// Gets the splashscreen url.
+        /// </summary>
+        [XmlIgnore]
+        public string SplashscreenUrl => "/Branding/Splashscreen";
```
nielsvanvelzen, line 36: New API properties should be marked nullable to maintain backwards compatibility in the SDK's/apiclients

daullmer: This property has no setter. So it doesn't matter if the clients send data in this property or not. It will always have the value `/Branding/Splashscreen`. The serializer doesn't throw an error if you don't send data for this property. So not sending it at all isn't a problem.
For deserializing the json by the clients shouldn't be a problem either since most serializers just ignore json properties they don't know.
So I don't think backwards compatibility is a problem here.

Jellyfin.Api/Controllers/ImageController.cs
```diff
@@ -1692,6 +1702,129 @@ public async Task<ActionResult<IEnumerable<ImageInfo>>> GetItemImageInfos([FromR
...
+        [ProducesResponseType(StatusCodes.Status204NoContent)]
+        [ProducesResponseType(StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [AcceptsImageFile]
+        public async Task<ActionResult> UploadCustomSplashscreen()
+        {
+            await using var memoryStream = await GetMemoryStream(Request.Body).ConfigureAwait(false);
+
+            // Handle image/png; charset=utf-8
+            var mimeType = Request.ContentType.Split(';').FirstOrDefault();
```
crobibero, line 1807:
```suggestion
            var mimeType = MediaTypeHeaderValue.Parse(Request.ContentType).MediaType;
```

Jellyfin.Drawing.Skia/SplashscreenBuilder.cs
```diff
@@ -0,0 +1,152 @@
...
+        private const int FinalHeight = 1080;
+        // generated collage resolution should be higher than the final resolution
+        private const int WallWidth = FinalWidth * 3;
+        private const int WallHeight = FinalHeight * 2;
+        private const int Rows = 6;
+        private const int Spacing = 20;
+
+        private readonly SkiaEncoder _skiaEncoder;
+
+        private Random? _random;
```
crobibero, line 22: This can be created in the function that utilizes it

MediaBrowser.Model/Branding/BrandingOptions.cs
crobibero: I believe this property does need to be nullable as the Kotlin sdk doesn't like properties missing that are marked as non-null

Emby.Server.Implementations/ScheduledTasks/Tasks/RefreshMediaLibraryTask.cs
```diff
@@ -58,6 +71,8 @@ public Task Execute(CancellationToken cancellationToken, IProgress<double> progr
 
             progress.Report(0);
 
+            _imageGenerator.GenerateSplashscreen(Path.Combine(_applicationPaths.DataPath, "splashscreen.webp"));
```
crobibero, line 74: Will this override a manually-uploaded splashscreen?

daullmer: No, the manually uploaded splashcreen gets saved as `splashscreen-upload.extension`. This path gets saved in branding.xml. A GET request to `/Branding/Splashscreen` checks if a file exists at this location. If not, splashscreen.webp is used. Generating a new splashscreen during the scheduled task will not override the path in branding.xml or the file at `splashscreen-upload.extension`

MediaBrowser.Model/Branding/BrandingOptions.cs
daullmer: Then I'm sorry I misunderstood you @nielsvanvelzen. I thought you were talking about an issue validating the object server-side.

crobibero (approved):

nielsvanvelzen (approved):

Jellyfin.Api/Controllers/ImageController.cs
```diff
@@ -1692,6 +1702,128 @@ public async Task<ActionResult<IEnumerable<ImageInfo>>> GetItemImageInfos([FromR
...
+            {
+                return BadRequest("Error reading mimetype from uploaded image");
+            }
+
+            var filePath = Path.Combine(_appPaths.DataPath, "splashscreen-upload" + MimeTypes.ToExtension(mimeType.Value));
+            var brandingOptions = _serverConfigurationManager.GetConfiguration<BrandingOptions>("branding");
+            brandingOptions.SplashscreenLocation = filePath;
+            _serverConfigurationManager.SaveConfiguration("branding", brandingOptions);
+
+            // use FileShare.None as this bypasses dotnet bug dotnet/runtime#42790 .
```
Bond-009, line 1818: Not needed anymore

Emby.Server.Implementations/ScheduledTasks/Tasks/RefreshMediaLibraryTask.cs
Bond-009, line 74: Will this always serve webp regardless of browser support?

Jellyfin.Drawing.Skia/DefaultImageGenerator.cs
```diff
@@ -0,0 +1,83 @@
...
+                DtoOptions = new DtoOptions(false),
+                ImageTypes = new ImageType[] { imageType },
+                Limit = 30,
+                // todo max parental rating configurable
+                MaxParentalRating = 10,
+                OrderBy = new ValueTuple<string, SortOrder>[]
+                {
+                    new ValueTuple<string, SortOrder>(ItemSortBy.Random, SortOrder.Ascending)
+                },
+                IncludeItemTypes = new string[] { "Movie", "Series" }
```
Bond-009, line 79: Needs rebase

Jellyfin.Drawing.Skia/DefaultImageGenerator.cs
```diff
@@ -0,0 +1,83 @@
...
+using MediaBrowser.Model.Entities;
+using MediaBrowser.Model.Querying;
+using Microsoft.Extensions.Logging;
+
+namespace Jellyfin.Drawing.Skia
+{
+    /// <summary>
+    /// The default image generator.
+    /// </summary>
+    public class DefaultImageGenerator : IImageGenerator
```
Bond-009, line 18: Why not put it in `ImageEncoder` where `StripCollageBuilder` etc also are

Bond-009 (changes requested):

Emby.Server.Implementations/ScheduledTasks/Tasks/RefreshMediaLibraryTask.cs
crobibero: No, this goes through the normal image return process. I just tested it and got image/png

Jellyfin.Drawing.Skia/SplashscreenBuilder.cs
```diff
@@ -0,0 +1,150 @@
...
+
+        /// <summary>
+        /// Generates a collage of posters and landscape pictures.
+        /// </summary>
+        /// <param name="posters">The poster paths.</param>
+        /// <param name="backdrops">The landscape paths.</param>
+        /// <returns>The created collage as a bitmap.</returns>
+        private SKBitmap GenerateCollage(IReadOnlyList<string> posters, IReadOnlyList<string> backdrops)
+        {
+            var random = new Random();
```
Bond-009, line 55: Can use `Random.Shared` instead

MediaBrowser.Model/Branding/BrandingOptions.cs
```diff
@@ -1,19 +1,38 @@
...
-        /// Gets or sets the custom CSS.
-        /// </summary>
-        /// <value>The custom CSS.</value>
-        public string? CustomCss { get; set; }
-    }
+    /// <summary>
+    /// Gets the splashscreen url.
+    /// </summary>
+    [XmlIgnore]
+    public string SplashscreenUrl => "/Branding/Splashscreen";
```
Bond-009, line 37: Why?

Bond-009 (changes requested):
Still not sure about using webp to store it locally

MediaBrowser.Model/Branding/BrandingOptions.cs
crobibero: This was requested above, but I think we can just remove it

Jellyfin.Api/Controllers/ImageController.cs
```diff
@@ -1692,6 +1703,130 @@ public async Task<ActionResult<IEnumerable<ImageInfo>>> GetItemImageInfos([FromR
...
+        /// <exception cref="ArgumentException">Error reading the image format.</exception>
+        [HttpPost("Branding/Splashscreen")]
+        [Authorize(Policy = Policies.RequiresElevation)]
+        [ProducesResponseType(StatusCodes.Status204NoContent)]
+        [ProducesResponseType(StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [AcceptsImageFile]
+        public async Task<ActionResult> UploadCustomSplashscreen()
+        {
+            await using var memoryStream = await GetMemoryStream(Request.Body).ConfigureAwait(false);
```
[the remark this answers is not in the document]
crobibero: This is how all the other image upload endpoints work, I don't think it's implement on the client yet

Bond-009 (approved):

PR #6440, opened 2021-08-16
Remove support for IDs in paths

MrTimscampi:
**Changes**

This feature makes path resolving more complex than needed, and is a (worse) duplicate of the Kodi-compatible [Parsing NFOs](https://kodi.wiki/view/NFO_files/Parsing) (which we already support).

This removes the inferior and limited path parsing of IDs, in favor of keeping (and encouraging the use of) Parsing NFOs.

**Issues**


cvium:
It's a legitimate feature that I wouldn't have added, but I think it's a bad idea to remove it as nothing is gained.

Issue #6415, opened 2021-08-11
System.ObjectDisposedException: Cannot access a closed Stream.


jeremyjjbrown:
**Describe the bug**
In progress OTA Recording does not start on FF or Chrome WebClients.

**System (please complete the following information):**
 - OS: Arch Linux
 - Virtualization: Docker
 - Clients: Firefox, Chrome (on MacOS) 
 - Browser: firefox 90, chrome 92
 - Jellyfin Version: jellyfin/jellyfin:10.7.6
 - Playback: OTA TV recording playback
 - Hardware Acceleration: unknown
 - Installed Plugins: none
 - Reverse Proxy: none
 - Networking: host
 - Storage: disk

**To Reproduce**

1. schedule a dvr
2. play the in progress dvr
4. UI hangs, play never starts, errors in server log

**Expected behavior**
The video should play

**Server Logs**
[02:45:22] [ERR] [136] Jellyfin.Server.Middleware.ExceptionMiddleware: Error processing request. URL GET /LiveTv/LiveRecordings/2f01c22a8fdf4d4881546f6cbc4172d1/stream.
System.ObjectDisposedException: Cannot access a closed Stream.
   at System.IO.MemoryStream.Read(Byte[] buffer, Int32 offset, Int32 count)
   at System.IO.MemoryStream.ReadAsync(Byte[] buffer, Int32 offset, Int32 count, CancellationToken cancellationToken)
--- End of stack trace from previous location ---
   at Microsoft.AspNetCore.Http.StreamCopyOperationInternal.CopyToAsync(Stream source, Stream destination, Nullable`1 count, Int32 bufferSize, CancellationToken cancel)
   at Microsoft.AspNetCore.Mvc.Infrastructure.FileResultExecutorBase.WriteFileAsync(HttpContext context, Stream fileStream, RangeItemHeaderValue range, Int64 rangeLength)
   at Microsoft.AspNetCore.Mvc.Infrastructure.FileStreamResultExecutor.ExecuteAsync(ActionContext context, FileStreamResult result)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.<InvokeNextResultFilterAsync>g__Awaited|29_0[TFilter,TFilterAsync](ResourceInvoker invoker, Task lastTask, State next, Scope scope, Object state, Boolean isCompleted)
[...]

PR #6444, opened 2021-08-17
Use ProgressiveFileStream for LiveRecordings endpoint

cvium:
**Issues**
Fixes https://github.com/jellyfin/jellyfin/issues/6415

I have no means to test this, but I'm guessing the MemoryStream gets disposed after the controller method returns. `ProgressiveFileStream` is better suited for this kind of use anyway.

### DIFF
--- a/Jellyfin.Api/Controllers/LiveTvController.cs
+++ b/Jellyfin.Api/Controllers/LiveTvController.cs
@@ -1172,7 +1172,7 @@ namespace Jellyfin.Api.Controllers
         [ProducesResponseType(StatusCodes.Status200OK)]
         [ProducesResponseType(StatusCodes.Status404NotFound)]
         [ProducesVideoFile]
-        public async Task<ActionResult> GetLiveRecordingFile([FromRoute, Required] string recordingId)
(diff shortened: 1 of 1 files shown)


crobibero (approved):

crobibero:
- [x] On-going recording
- [x] Finished recording

Recorded content plays back for me

cvium:
@crobibero did you test that playback longer than 30s works?

crobibero:
I can play back the recording I made this morning without issue, currently 2+ minutes in. I should note that I'm only using this PR, I did not include the linked one

cvium:
Oh, have to use them both to test that part.

crobibero:
I added that commit, I can play LiveTV and recordings for more then 30s, but I never hit that line of code. To me it looks like the stopwatch case will only be triggered if we're at the end of the file and the transcoding job has finished? I don't know how I could trigger that for testing

cvium:
Yes, it's a somewhat rare case. As long as it doesn't break playback it's good enough for me

barronpm (approved):

Bond-009 (approved):

PR #6446, opened 2021-08-17
Fix embedded subtitles taking priority over external ones

### LINKED ISSUE #6407: Embedded subtitles take precedence over external subtitles
**Describe the bug**

When a video file has an embedded English subtitle, as well as an external English subtitle, the embedded subtitle will be automatically selected as default subtitle.

**System:**
 - OS: Windows
 - Virtualization: Docker
 - Clients: Jellyfin Media Player & Web
 - Browser: Edge 92
 - Jellyfin Version: 10.7.6
 - Playback: ?
 - Hardware Acceleration: None
 - Installed Plugins: None
 - Reverse Proxy: None
 - Base URL: None
 - Networking: Bridge
 - Storage: local

**To Reproduce**
1. Add a `.MKV` file with embedded English subtitle, for example, `Slimes 01.mkv`.
2. Add an `.ass` English subtitle file, for example, `Slimes 01.en.ass`.
3. Scan the file with *Jellyfin*.
4. Click `Play` on the episode. The embedded subtitles will be used.

**Expected behavior**

As an end-user, I expect that external subtitles take precedence over embedded subtitles. I took the time to add these external subtitles, so I would expect a player to respect that and use them by default. This is what happens in other software I use, such as *MPC*, *Plex Media Player* and *VLC*. Selecting the embedded subtitle also causes unnecessary `ffmpeg` subtitle extraction and delays.

**Screenshots**
image

Deathspike:
**Changes**
Fixes `MediaStreamSelector`.`GetDefaultSubtitleStreamIndex` prioritizing embedded subtitles over external subtitles. Each strategy (`Default`, `Smart`, `Always`, `OnlyForced`) will now prioritize external subtitles before embedded ones. If there are no external subtitles, behavior is unaltered. If there are external subtitles, then:

- `Default`: If there are one or more external subtitles, they're sorted according to the original behavior. Even if no external subtitle is `default` or `forced`, external subtitles always take priority over embedded subtitles.
- `Smart`: If there are one or more external subtitles, they're sorted according to the original behavior. If no external subtitle matches the preferred user language, embedded subtitles are used.
- `Always`: If there are one or more external subtitles, they're sorted according to the original behavior. If all external subtitles are `forced`, embedded subtitles are used.
- `OnlyForced`: If there are one or more external subtitles, they're sorted according to the original behavior. If all external subtitles are not `forced`, embedded subtitles are used.

**Issues**
Fixes #6407

### DIFF
--- a/Emby.Server.Implementations/Library/MediaStreamSelector.cs
+++ b/Emby.Server.Implementations/Library/MediaStreamSelector.cs
@@ -57,14 +57,18 @@ namespace Emby.Server.Implementations.Library
             {
                 // Prefer embedded metadata over smart logic
 
-                stream = streams.FirstOrDefault(s => s.IsForced && string.Equals(s.Language, audioTrackLanguage, StringComparison.OrdinalIgnoreCase)) ??
-                    streams.FirstOrDefault(s => s.IsForced) ??
-                    streams.FirstOrDefault(s => s.IsDefault);
+                stream = streams.FirstOrDefault(s => s.IsExternal && s.IsForced && string.Equals(s.Language, audioTrackLanguage, StringComparison.OrdinalIgnoreCase)) ??
+                         streams.FirstOrDefault(s => s.IsExternal && s.IsForced) ??
+                         streams.FirstOrDefault(s => s.IsExternal && s.IsDefault) ??
+                         streams.FirstOrDefault(s => s.IsExternal) ??
+                         streams.FirstOrDefault(s => s.IsForced && string.Equals(s.Language, audioTrackLanguage, StringComparison.OrdinalIgnoreCase)) ??
+                         streams.FirstOrDefault(s => s.IsForced) ??
+                         streams.FirstOrDefault(s => s.IsDefault);
 
                 // if the audio language is not understood by the user, load their preferred subs, if there are any
                 if (stream == null && !preferredLanguages.Contains(audioTrackLanguage, StringComparer.OrdinalIgnoreCase))
                 {
-                    stream = streams.Where(s => !s.IsForced).FirstOrDefault(s => preferredLanguages.Contains(s.Language, StringComparer.OrdinalIgnoreCase));
+                    stream = streams.FirstOrDefault(s => !s.IsForced && preferredLanguages.Contains(s.Language, StringComparer.OrdinalIgnoreCase));
                 }
             }
             else if (mode == SubtitlePlaybackMode.Smart)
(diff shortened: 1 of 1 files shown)


Emby.Server.Implementations/Library/MediaStreamSelector.cs
```diff
@@ -57,14 +57,18 @@ public static class MediaStreamSelector
...
-                stream = streams.FirstOrDefault(s => s.IsForced && string.Equals(s.Language, audioTrackLanguage, StringComparison.OrdinalIgnoreCase)) ??
-                    streams.FirstOrDefault(s => s.IsForced) ??
-                    streams.FirstOrDefault(s => s.IsDefault);
+                stream = streams.FirstOrDefault(s => s.IsExternal && s.IsForced && string.Equals(s.Language, audioTrackLanguage, StringComparison.OrdinalIgnoreCase)) ??
+                         streams.FirstOrDefault(s => s.IsExternal && s.IsForced) ??
+                         streams.FirstOrDefault(s => s.IsExternal && s.IsDefault) ??
+                         streams.FirstOrDefault(s => s.IsExternal) ??
+                         streams.FirstOrDefault(s => s.IsForced && string.Equals(s.Language, audioTrackLanguage, StringComparison.OrdinalIgnoreCase)) ??
+                         streams.FirstOrDefault(s => s.IsForced) ??
+                         streams.FirstOrDefault(s => s.IsDefault);
```
cvium, line 66: This is a lot of looping

cvium: Would it make sense to to an orderby instead?

Deathspike: Sure. The advantage of this approach is readability; it's perfectly clear what is prioritized with the code like this. Using a one-time pre-sorted collection will be faster (even though we're talking literal microseconds here), but will be far less readable. I'll have to rewrite most of the subtitle selection code. Is that preferable? If so, I'll get on that.

1337joe (approved):

Emby.Server.Implementations/Library/MediaStreamSelector.cs
1337joe: Just ordering by IsExternal and running the old logic could end in a case where non-external subtitles matching IsForced would come before an external subtitle.

You could filter streams into externalStreams and embeddedStreams and run the logic on them sequentially to avoid having so many lines of FirstOrDefault calls, but you don't save any loops and it adds complexity.

1337joe: Thinking about this more, I think it would be cleaner/more maintainable to first filter for external subs, then run the stream selection logic on just the external subs (if there are any) or the full list. That way there isn't duplicated logic, and it's clear that external subs are prioritized. Something like this (untested):
```c#
// only consider external streams if present
var filteredStreams = streams.Any(s => s.IsExternal) ? streams.Where(s => s.IsExternal) : streams;
// treat external streams as default
stream = filteredStreams.FirstOrDefault(s => s.IsForced && string.Equals(s.Language, audioTrackLanguage, StringComparison.OrdinalIgnoreCase)) ??
         filteredStreams.FirstOrDefault(s => s.IsForced) ??
         filteredStreams.FirstOrDefault(s => s.IsDefault || s.IsExternal);
```

The logic works here because an external sub will be chosen if there is one, so it would never fall through to the embedded subs.

PR #6453, opened 2021-08-19
Simplify logic in MediaSourceInfo

GodTamIt:
**Changes**
This change cleans up the logic inside `MediaSourceInfo.cs`. As part of the change, the file will no longer have to disable the nullable context.


MediaBrowser.Model/Dto/MediaSourceInfo.cs
```diff
@@ -116,139 +117,91 @@ public MediaSourceInfo()
...
-            {
-                if (!stream.IsExternal)
-                {
-                    bitrate += stream.BitRate ?? 0;
-                }
-            }
+            int bitrate = MediaStreams
+                .Where(stream => !stream.IsExternal)
+                .Select(stream => stream.BitRate ?? 0)
+                .Sum();
```
cvium, line 138: Performance is probably not a huge concern with this code, but we generally try to avoid using LINQ because it's allocating Enumerables, Method Groups etc. And our memory traffic is already quite painful.

MediaBrowser.Model/Dto/MediaSourceInfo.cs
```diff
@@ -116,139 +117,91 @@ public MediaSourceInfo()
...
-            if (numStreams == 0)
+        public int? GetStreamCount(MediaStreamType type)
+        {
+            if (MediaStreams.Count == 0)
             {
                 return null;
             }
 
-            return numMatches;
+            return MediaStreams.Where(i => i.Type == type).Count();
```
cvium, line 198:
```suggestion
            return MediaStreams.Count(i => i.Type == type);
```

MediaBrowser.Model/Dto/MediaSourceInfo.cs
GodTamIt: Ah, coming from a C++/Rust world, I had no idea about this!

Happy to leave this the way it was before if that's your preference?

MediaBrowser.Model/Dto/MediaSourceInfo.cs
```diff
@@ -22,33 +22,34 @@ public MediaSourceInfo()
             SupportsDirectStream = true;
             SupportsDirectPlay = true;
             SupportsProbing = true;
+            TranscodeReasons = Array.Empty<TranscodeReason>();
         }
 
         public MediaProtocol Protocol { get; set; }
 
-        public string Id { get; set; }
+        public string? Id { get; set; }
```
cvium, line 30: This is a big no-no combined with the use of the `!` operator. Either it can be null or it can't. Generally we shouldn't be using `!`.

GodTamIt:
Will abandon this change for now.

PR #6457, opened 2021-08-21
Add 'includeAdult' option for TheMovieDb

Izumiko:
**Changes**
Some TV shows and movies are labeled as 'adult' at TheMovieDb. The TMDbLib uses the 'includeAdult' parameter to show this content in search results. This PR adds a configuration option to support this feature.

### DIFF
--- a/MediaBrowser.Providers/Plugins/Tmdb/Configuration/PluginConfiguration.cs
+++ b/MediaBrowser.Providers/Plugins/Tmdb/Configuration/PluginConfiguration.cs
@@ -1,0 +1,15 @@
+using MediaBrowser.Model.Plugins;
+
+namespace MediaBrowser.Providers.Plugins.Tmdb
+{
+    /// <summary>
+    /// Plugin configuration class for TMDb library.
+    /// </summary>
+    public class PluginConfiguration : BasePluginConfiguration
+    {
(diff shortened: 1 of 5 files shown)


crobibero (approved):

MediaBrowser.Providers/Plugins/Tmdb/Configuration/PluginConfiguration.cs
```diff
@@ -0,0 +1,11 @@
+#pragma warning disable CS1591
```
Bond-009, line 1: Please add documentation to new files.

MediaBrowser.Providers/Plugins/Tmdb/Plugin.cs
```diff
@@ -0,0 +1,40 @@
+#pragma warning disable CS1591
```
Bond-009, line 1: Same

Bond-009 (changes requested):

Izumiko:
It is better to add this option to the library level, not to global. But I didn't find a proper way to achieve this due to my limited knowledge. So I added it to the plugin configuration page for now.

Bond-009 (approved):

PR #6466, opened 2021-08-25
Users/rmcn96/fix readonly image

rclabough:
**Changes**
To reproduce: Media is stored on a read-only medium, such as a readonly mount.  Have a piece of media that has a cover art automatically read in by the media scanner.  If you try to replace that cover art, the API would return a 500 error due to the underlying system being read only.

The issue is resolved by handling the read-only IO Exception in the same manner it handles files that are not found.  It now ignores the error (because it is not allowed to delete the file), and continues with the metadata changes.  In the internal DB it recognizes that the art has been changed, while the original file is untouched.

### DIFF
--- a/MediaBrowser.Providers/Manager/ImageSaver.cs
+++ b/MediaBrowser.Providers/Manager/ImageSaver.cs
@@ -189,6 +189,11 @@ namespace MediaBrowser.Providers.Manager
                 catch (FileNotFoundException)
                 {
                 }
+                catch (Exception ex)
+                {
+                    _logger.LogInformation("Exception caught: ", ex.Message);
+                    throw;
+                }
                 finally
                 {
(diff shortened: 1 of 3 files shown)


MediaBrowser.Providers/Manager/ImageSaver.cs
```diff
@@ -189,6 +189,10 @@ await using (source.ConfigureAwait(false))
                 catch (FileNotFoundException)
                 {
                 }
+                catch (IOException e) when (e.Message.Contains("Read-only file system", StringComparison.OrdinalIgnoreCase))
```
Bond-009, line 192: Exception messages get translated to the system lang, this won't work.

Bond-009 (changes requested):

Bond-009:
Closing as fundamentally broken atm

PR #6486, opened 2021-08-31
Support forwarded headers for api url

fredriklindberg:
Adds a configuration option (default false) that allows the LocalAddress field in the system info endpoint to be populated based on information received in the HTTP request. Takes into account data injected by reverse proxies, this allows the published server url to dynamically be populated with the correct hostname.

Handles the [Forwarded](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded) and X-Forwarded-{Host,Port,Proto} headers.

**Note** that this disables the `XForwardedProto` processing middleware as that dropped too much information making it impossible to determine if a header was received or not. The middleware does also not seem capable of handling `X-Forwarded-Port` nor the `Forwarded` header.

This makes it easier to run behind a reverse proxy with multiple hostnames (especially to support chromecast). For example the following nginx hack is no longer needed
```

location /System/Info/Public {
  add_header Content-Type "application/json; charset=utf8";
  return 200 '{"LocalAddress":"$host","ServerName":"some-server","Version":"10.8.0","ProductName":"Jellyfin Server","OperatingSystem":"Linux","Id":"abcdef1234567890abcdef1234567890"}';
}
```

Examples with the network option `EnablePublishedServerUriByRequest` enabled (default false)
```
curl -H "x-forwarded-host: example.com" -H "x-forwarded-proto: https" http://localhost:8096/System/Info/Public
{"LocalAddress":"https://example.com","ServerName":"mbp","Version":"10.8.0","ProductName":"Jellyfin Server","OperatingSystem":"Darwin","Id":"ebd3473116a948eb90754b012a8d0d15","StartupWizardCompleted":true}
```
```
curl http://localhost:8096/System/Info/Public
{"LocalAddress":"http://localhost:8096","ServerName":"mbp","Version":"10.8.0","ProductName":"Jellyfin Server","OperatingSystem":"Darwin","Id":"ebd3473116a948eb90754b012a8d0d15","StartupWizardCompleted":true}
```


fredriklindberg:
> What's the motivation behind this exactly? It looks somewhat out of scope for jellyfin.

It simplifies running jellyfin behind a reverse proxy. It's basically a dynamic version of the "Published Server URIs" feature that is already available. But instead of deriving the URI based on client IP or interface it's using industry standard HTTP headers to construct the URI based on what the reverse proxy injected.

It's the same as getting the real client ip from X-Forwarded-For, but [X-Forwarded-Host](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host) (and associated headers) is getting the original host requested by the client.

For example, it would simplify the setup of a TLS terminating reverse proxy.
```
+--------+                        +------+                          +----------+
| Client +----------------------->|Proxy +------------------------->| Jellyfin |
+--------+ https://jf.example.com +------+ http://jellyfin.internal +----------+
                                           host: jellyfin.internal
                                           x-forwarded-host: jf.example.com
                                           x-forwarded-proto: https
                                           x-forwarded-for: <client-ip>
```

It's only needed because Jellyfin is constructing its own base URIs (/System/Info/Public), which is not able to do properly in an environment where it's behind a proxy (since it's unaware of it).

cvium:
But isn't X forwarded host not available in HttpContext.Request.Host?

fredriklindberg:
If you set `ForwardedHeaders.XForwardedHost` in the middlewear it will populate `Host` with the value of the header. That's how I started out.

However the middlewear does not support `X-Forwarded-Port`. This together with the peculiar way the asp.net core middlewear is handling these headers (overwriting the original request data) makes it impossible to determine which port to use when re-constructing the URL.

It is supposed to populate `X-Original-X`, but that header seems to be populated regardless if a `X-Fowarded-X` is received or not (or I did something wrong while testing).

It would be possible to make it work if we'd require the `X-Forwarded-Host` header sent to jellyfin to contain a port number (`X-Forwarded-Host: jf.example.com:443`. It's not industry standard, but I guess it would be good enough if documented.

cvium:
It doesn't seem like there's much of a standard (no RFC at least) and I'd prefer not re-implementing handling of X-Forwarded-X. From some quick research it appears that the port being in `-Host` isn't entirely uncommon.

fredriklindberg:
The standard is the `Forwarded` header, defined in [RFC7239](https://www.rfc-editor.org/rfc/rfc7239). This one is not supported at all by the ASP.NET core middleware.

But if it's preferred to use the ASP.NET core middleware only, I'll update the pull request to do that.

fredriklindberg:
Updated the PR. It works if given a port in the host header, but breaks otherwise. It's good enough for my needs.

```
curl -H "x-forwarded-host: example.com:443" -H "x-forwarded-proto: https" http://localhost:8096/System/Info/Public
{"LocalAddress":"https://example.com","ServerName":"mbp","Version":"10.8.0","ProductName":"Jellyfin Server","OperatingSystem":"Darwin","Id":"ebd3473116a948eb90754b012a8d0d15","StartupWizardCompleted":true}
```

Without port
```
curl -H "x-forwarded-host: example.com" -H "x-forwarded-proto: https" http://localhost:8096/System/Info/Public
{"LocalAddress":"https://example.com:8096","ServerName":"mbp","Version":"10.8.0","ProductName":"Jellyfin Server","OperatingSystem":"Darwin","Id":"ebd3473116a948eb90754b012a8d0d15","StartupWizardCompleted":true}
```

No headers
```
curl http://localhost:8096/System/Info/Public
{"LocalAddress":"http://localhost:8096","ServerName":"mbp","Version":"10.8.0","ProductName":"Jellyfin Server","OperatingSystem":"Darwin","Id":"ebd3473116a948eb90754b012a8d0d15","StartupWizardCompleted":true}
```

Emby.Server.Implementations/ApplicationHost.cs
```diff
@@ -1150,6 +1150,18 @@ public string GetSmartApiUrl(IPAddress remoteAddr, int? port = null)
         /// <inheritdoc/>
         public string GetSmartApiUrl(HttpRequest request, int? port = null)
         {
+            // Return the host in the HTTP request as the API url
+            if (ConfigurationManager.GetNetworkConfiguration().EnablePublishedServerUriByRequest)
+            {
+                int? requestPort = request.Host.Port;
+                if ((requestPort == 80 && request.Scheme == "http") || (requestPort == 443 && request.Scheme == "https"))
```
crobibero, line 1157: If `Scheme` is always lowercase:
```suggestion
                if ((requestPort == 80 && string.Equals(request.Scheme, "http", StringComparison.Ordinal)) || (requestPort == 443 && string.Equals(request.Scheme, "https", StringComparison.Ordinal)))
```

Otherwise:
```suggestion
                if ((requestPort == 80 && string.Equals(request.Scheme, "http", StringComparison.OrdinalIgnoreCase)) || (requestPort == 443 && string.Equals(request.Scheme, "https", StringComparison.OrdinalIgnoreCase)))
```

fredriklindberg: Is string.Equals preferred over == ?

crobibero: Yes

fredriklindberg: Alright, fixed

crobibero (approved):

Emby.Server.Implementations/ApplicationHost.cs
```diff
@@ -1150,6 +1150,19 @@ public string GetSmartApiUrl(IPAddress remoteAddr, int? port = null)
         /// <inheritdoc/>
         public string GetSmartApiUrl(HttpRequest request, int? port = null)
         {
+            // Return the host in the HTTP request as the API url
+            if (ConfigurationManager.GetNetworkConfiguration().EnablePublishedServerUriByRequest)
+            {
+                int? requestPort = request.Host.Port;
+                string requestScheme = request.Scheme.ToLower();
```
Bond-009, line 1157: No need for this now

fredriklindberg: Missed that one, fixed.

PR #6494, opened 2021-09-01
Clean up VideoImageProvider

Bond-009:


barronpm (approved):

crobibero (approved):

MediaBrowser.Providers/MediaInfo/VideoImageProvider.cs
```diff
@@ -21,56 +21,55 @@ public class VideoImageProvider : IDynamicImageProvider, IHasOrder
...
 
+        /// <inheritdoc />
         // Make sure this comes after internet image providers
         public int Order => 100;
 
+        /// <inheritdoc />
         public IEnumerable<ImageType> GetSupportedImages(BaseItem item)
         {
-            return new List<ImageType> { ImageType.Primary };
+            return new[] { ImageType.Primary };
```
cvium, line 41: A micro-optimization for sure, but should we make a static "backing" field for this? Seems like a total waste to create new arrays all the time. And the caller of `GetSupportedImages` typically calls `.ToArray` on it as well :(

cvium (approved):

PR #6511, opened 2021-09-05
Refresh episodes before seasons

revam:
Refresh episodes before seasons to get the know which season to append the episode to. I don't know if this should be done on every refresh of the series item, or only on the initial refresh, but this PR adds it to to every refresh.

**Changes**

Refresh episodes before seasons

**Issues**

### DIFF
--- a/MediaBrowser.Providers/TV/SeriesMetadataService.cs
+++ b/MediaBrowser.Providers/TV/SeriesMetadataService.cs
@@ -137,6 +137,8 @@ namespace MediaBrowser.Providers.TV
             // Loop through the unique season numbers
             foreach (var episode in episodesInSeriesFolder)
             {
+                await episode.RefreshMetadata(new MetadataRefreshOptions(new DirectoryService(FileSystem)), cancellationToken).ConfigureAwait(false);
+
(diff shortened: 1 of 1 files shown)


revam:
Note: I may have introduced double-fetching of the episode metadata.

MediaBrowser.Providers/TV/SeriesMetadataService.cs
```diff
@@ -137,6 +137,8 @@ private async Task FillInMissingSeasonsAsync(Series series, CancellationToken ca
             // Loop through the unique season numbers
             foreach (var episode in episodesInSeriesFolder)
             {
+                await episode.RefreshMetadata(new MetadataRefreshOptions(new DirectoryService(FileSystem)), cancellationToken).ConfigureAwait(false);
```
cvium, line 140: This shouldn't be necessary. The missing seasons is supposed to execute after metadata has already been refreshed.

Even if this fixes it I'm not sure it's the correct solution.

revam: Then i will probe a little further to see if i can find out why the metadata is not already fetched.

PR #6049, opened 2021-05-15
EFCore Image model update and cleanup

ferferga:
Some of the entities defined in https://github.com/jellyfin/jellyfin/pull/2293 have changed, one of them being the images, which now include blurhashes.

**Changes**
- Removed DisplayPreferencesDto in MediaBrowserModel, in favor of the new one defined in Jellyfin.Api
- Renamed **Artwork** entity to **Image** (for clarity, as suggested by @dkanada https://github.com/jellyfin/jellyfin/pull/2293#issuecomment-619549080)
- Added different fields for tracking file creation/modification date and database creation/modification date.
- ImageType has been moved to ``Jellyfin.Data.Enums`` from ``MediaBrowser.Model.Entities``, to remove the duplication that previously existed with ``ArtworkKind``.
- Created a migration that updates the model as described below, being skipped if ``MigrateUserDb`` is not applied yet.
- Log a message when there are pending EFCore migrations, informing the user that they can take a while to complete and advising against shutting down the server while they're in progress.
- Merged **ImageInfos** with **Image**. Detail of the changes:

**New Image entity properties**

```
public int Id
public string Path
public ImageType Type
public string? Blurhash
public DateTime LastModifiedDate = DateTime.UtcNow;
public DateTime AddedDate
public DateTime? FileCreationDate
public DateTime? FileModificationDate
public uint RowVersion
```

*With ImageType being the current ImageType enum*

**Current Image entity properties**
```
public int Id
public string Path
public ArtKind Kind
public uint RowVersion
```

**Current ImageInfo properties**
```
public int Id
public Guid? UserId
public string Path
public DateTime LastModified
```

Every image we have in the client has (or can potentially have) all those fields in common. There's now a single 1:M or 1:1 relationship between any entity and ``Image``. Examples:

```
* Movie => 1 : M being M Image objects for the Primary, Background, Art, etc images
* Photo => 1 : 1 Photo entity storing shutter speed, GPS location, etc... And Image (type Primary, the default one) storing blurhash and the path on disk.
* User => 1 : 1 User storing all the user information and Image the profile photo.
```

In this PR, only used entities were changed to use the new ``Image`` entity, as the others haven't been migrated to EFCore yet and we have time to revisit the others as well.

**Issues**

[...]

PR #6522, opened 2021-09-06
EFCore improvements and cleanups

ferferga:
This PR cherrypicks some of the changes made in #6049 in order to ease the review:

* Remove all the unused usings in the solution (we might want to enforce not unused usings to avoid having this situation again)
* Logs EFcore migrations, so users are aware of them
* Uses appHost instead of creating a new context in Program.cs
* (Not in #6049) Decouples model configuration into its own separate classes, so they're easier to maintain and doesn't clog the ``JellyfinDb.cs`` file


Jellyfin.Server/CoreAppHost.cs
```diff
@@ -78,7 +78,8 @@ protected override void RegisterServices()
             }
 
             ServiceCollection.AddDbContextPool<JellyfinDb>(
-                 options => options.UseSqlite($"Filename={Path.Combine(ApplicationPaths.DataPath, "jellyfin.db")}"));
+                 options => options.UseLoggerFactory(LoggerFactory).
+                    UseSqlite($"Filename={Path.Combine(ApplicationPaths.DataPath, "jellyfin.db")}"));
```
crobibero, line 82:
```suggestion
                 options => options
                   .UseLoggerFactory(LoggerFactory)
                   .UseSqlite($"Filename={Path.Combine(ApplicationPaths.DataPath, "jellyfin.db")}"));
```

I'm sure I got indentation incorrect, but this is more readable

ferferga: Good catch!

crobibero (approved):

Jellyfin.Server.Implementations/JellyfinDb.cs
```diff
@@ -153,100 +153,10 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
-            modelBuilder.Entity<Permission>()
-                .HasIndex(p => new { p.UserId, p.Kind })
-                .HasFilter("[UserId] IS NOT NULL")
-                .IsUnique();
-
-            modelBuilder.Entity<Preference>()
-                .HasIndex(p => new { p.UserId, p.Kind })
-                .HasFilter("[UserId] IS NOT NULL")
-                .IsUnique();
+            // Configuration for each entity is in it's own class inside 'ModelConfiguratio'.
```
barronpm, line 158:
```suggestion
            // Configuration for each entity is in it's own class inside 'ModelConfiguration'.
```

barronpm (approved):

cvium (approved):

PR #6528, opened 2021-09-07
Add "Production" and "TV" episode orders to the metadata editor.

oledfish:
Currently, the metadata editor allows episodes from series to pull the TMDB information and be grouped by aired date, absolute order, or DVD order. [TMDbLib exposes several more episode ordering schemas](https://github.com/LordMike/TMDbLib/blob/ef5d852982422f95fb646034c56d9dc23ce40f27/TMDbLib/Objects/TvShows/TvGroupType.cs) yet not all are made available in Jellyfin. 

Since the library takes care of fetching the information, this change is as simple as adding the proper options in the server as well as the web client. Corresponding PR for jellyfin-web is [here](https://github.com/jellyfin/jellyfin-web/pull/2937).


crobibero (approved):

MediaBrowser.Providers/Plugins/Tmdb/TmdbClientManager.cs
```diff
@@ -148,8 +148,10 @@ public async Task<TvShow> GetSeriesAsync(int tmdbId, string language, string ima
         private async Task<TvGroupCollection> GetSeriesGroupAsync(int tvShowId, string displayOrder, string language, string imageLanguages, CancellationToken cancellationToken)
         {
             TvGroupType? groupType =
+                string.Equals(displayOrder, "production", StringComparison.Ordinal) ? TvGroupType.Production :
                 string.Equals(displayOrder, "absolute", StringComparison.Ordinal) ? TvGroupType.Absolute :
                 string.Equals(displayOrder, "dvd", StringComparison.Ordinal) ? TvGroupType.DVD :
+                string.Equals(displayOrder, "tv", StringComparison.Ordinal) ? TvGroupType.TV :
```
cvium, line 154: What is TV order?

oledfish: As far as I know it's supposed to be used for shows with episodes that were grouped differently in specific regions, yet unfortunately it's not a very consistent option:

- [Naruto](https://www.themoviedb.org/tv/46260) shows a five-season grouping as opposed to other options that show either a four-season grouping or a single season with all episodes lumped in.
- [One Piece](https://www.themoviedb.org/tv/37854) has all its episodes grouped into a single season in TV order, while other options are far better organized.

cvium: In that case I think "tv" might be a bad name. Not sure what it should be called though.

oledfish: Neither do I in that case, the names used by the website are as descriptive as they get.

cvium: Can we call it "regional"?

oledfish: I think it does, but it would be inconsistent and confusing compared to the TMDB name

WWWesten: Is not "TV" = "Syndication"? See https://en.wikipedia.org/wiki/Broadcast_syndication
Why not add other types? For example "La casa de papel" (Money Heist) had Season 1 Spain version (aired, 15 episodes) and Netflix version (digital, 22 episodes) with different durations and number.

oledfish: The other types I hadn't added in this PR have obvious names (OriginalAirDate, StoryArc, Digital) so maybe just calling TV "Syndication" would be enough to have all of them?

WWWesten: Why not list all types and in the same order [as here](https://github.com/LordMike/TMDbLib/blob/ef5d852982422f95fb646034c56d9dc23ce40f27/TMDbLib/Objects/TvShows/TvGroupType.cs)?

Kaduo:
When multiple episode orders of a same type are available on TMDB for a series (for instance, One Piece has [two different "absolute" orders](https://www.themoviedb.org/tv/37854/episode_groups)), is there a way to control (or predict for that matter) which episode order will be fetched by TMDBLib?

MediaBrowser.Providers/Plugins/Tmdb/TmdbClientManager.cs
oledfish: I have pushed a commit listing all the types found in the client API.

joshuaboniface: I think using the values provided by the API are sensible, if we rename them then it's on us to come up with better mappings.

joshuaboniface (approved):

joshuaboniface:
> When multiple episode orders of a same type are available on TMDB for a series (for instance, One Piece has [two different "absolute" orders](https://www.themoviedb.org/tv/37854/episode_groups)), is there a way to control (or predict for that matter) which episode order will be fetched by TMDBLib?

I think the question would be how the TMDBLib API returns them, i.e. what is matched to what for this particular edge case. Since this is an addition though, and the existing system is just as potentially ambiguous, I don't know how much we should dwell on it too much in this (simple) PR.

PR #6531, opened 2021-09-08
Fix images "not loading" after manual identification

### LINKED ISSUE #3676: Identify with replace image takes more than 1 minute
**Describe the bug**

When I want to identify movies with option "replace image" it takes too long
image

This is not performance error of my server (2 core  CPU with 16GB RAM). My internet connection is 30/2Mbps, so replacing some small images shouldn't take so long.

 - OS: Debian
 - Virtualization: Docker
 - Clients: does not matter
 - Browser: does not matter
 - Jellyfin Version: 10.6
 - Playback: -
 - Installed Plugins: -
 - Reverse Proxy: none
 - Base URL: -
 - Networking: host
 - Storage: local

**To Reproduce**

1. Go to movies and try to identify with option replace images

**Expected behavior**

**Logs**

### LINKED ISSUE #6467: [Video] Identify with replace existing images fails
https://www.youtube.com/watch?v=a4R1RmNFahY

GodTamIt:
**Changes**

This change fixes a bug where images (usually of type `Primary`) seem to "not load" after the user custom identifies and matches the metadata with a particular source. This is due to the fact that search results shown on the UI will show primary covers from another language if those are the only covers available (#6467 shows that it is probably common in anime). However, in the previous internal logic, the image filtering does not allow for images from all languages, leading to a confusing mismatch between UI/UX and internal metadata tagging logic.

The fix enables images from all languages to be included during internal image fetching but adds logic to order results by the preferred languages so that the preferred language settings are still honored. If there are images available in the preferred language, they will be ordered before other results.

**Issues**

Fixes #3676
Fixes #6467

### DIFF
--- a/MediaBrowser.Model/Extensions/EnumerableExtensions.cs
+++ b/MediaBrowser.Model/Extensions/EnumerableExtensions.cs
@@ -18,6 +18,12 @@ namespace MediaBrowser.Model.Extensions
         /// <returns>The ordered remote image infos.</returns>
         public static IEnumerable<RemoteImageInfo> OrderByLanguageDescending(this IEnumerable<RemoteImageInfo> remoteImageInfos, string requestedLanguage)
         {
+            if (string.IsNullOrWhiteSpace(requestedLanguage))
+            {
+                // Default to English if no requested language is specified.
+                requestedLanguage = "en";
+            }
+
             var isRequestedLanguageEn = string.Equals(requestedLanguage, "en", StringComparison.OrdinalIgnoreCase);
 
             return remoteImageInfos.OrderByDescending(i =>
@@ -27,14 +33,16 @@ namespace MediaBrowser.Model.Extensions
                         return 3;
                     }
 
-                    if (!isRequestedLanguageEn && string.Equals("en", i.Language, StringComparison.OrdinalIgnoreCase))
-                    {
-                        return 2;
-                    }
-
                     if (string.IsNullOrEmpty(i.Language))
                     {
+                        // Assume empty image language is likely to be English.
                         return isRequestedLanguageEn ? 3 : 2;
+                    }
+
+                    if (!isRequestedLanguageEn && string.Equals(i.Language, "en", StringComparison.OrdinalIgnoreCase))
+                    {
+                        // Prioritize English over non-requested languages.
+                        return 2;
                     }
 
                     return 0;
--- a/MediaBrowser.Providers/Manager/ItemImageProvider.cs
+++ b/MediaBrowser.Providers/Manager/ItemImageProvider.cs
@@ -275,7 +275,7 @@ namespace MediaBrowser.Providers.Manager
                     item,
                     new RemoteImageQuery(provider.Name)
                     {
-                        IncludeAllLanguages = false,
+                        IncludeAllLanguages = true,
                         IncludeDisabledProviders = false,
                     },
                     cancellationToken).ConfigureAwait(false);
@@ -469,7 +469,7 @@ namespace MediaBrowser.Providers.Manager
             CancellationToken cancellationToken)
         {
             var eligibleImages = images
-                .Where(i => i.Type == type && !(i.Width.HasValue && i.Width.Value < minWidth))
+                .Where(i => i.Type == type && i.Width >= minWidth)
                 .ToList();
 
             if (EnableImageStub(item) && eligibleImages.Count > 0)
--- a/MediaBrowser.Providers/Manager/ProviderManager.cs
+++ b/MediaBrowser.Providers/Manager/ProviderManager.cs
@@ -24,6 +24,7 @@ using MediaBrowser.Controller.Providers;
 using MediaBrowser.Controller.Subtitles;
 using MediaBrowser.Model.Configuration;
 using MediaBrowser.Model.Entities;
+using MediaBrowser.Model.Extensions;
 using MediaBrowser.Model.IO;
 using MediaBrowser.Model.Providers;
 using Microsoft.Extensions.Logging;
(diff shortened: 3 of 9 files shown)


MediaBrowser.Providers/Manager/ItemImageProvider.cs
```diff
@@ -470,7 +470,7 @@ private bool UpdateMultiImages(BaseItem item, IReadOnlyList<LocalImageInfo> imag
             CancellationToken cancellationToken)
         {
             var eligibleImages = images
-                .Where(i => i.Type == type && !(i.Width.HasValue && i.Width.Value < minWidth))
+                .Where(i => i.Type == type && (!i.Width.HasValue || i.Width.Value >= minWidth))
```
cvium, line 473: Comparisons with null return false.
```suggestion
                .Where(i => i.Type == type && i.Width.Value >= minWidth)
```

MediaBrowser.Providers/Manager/ProviderManager.cs
```diff
@@ -274,12 +276,20 @@ public async Task<IEnumerable<RemoteImageInfo>> GetAvailableRemoteImages(BaseIte
...
+                    //
+                    // TODO: should exception case of "en" (English) eventually be removed?
                     result = result.Where(i => string.IsNullOrEmpty(i.Language) ||
                                                preferredLanguages.Contains(i.Language, StringComparer.OrdinalIgnoreCase) ||
                                                string.Equals(i.Language, "en", StringComparison.OrdinalIgnoreCase));
                 }
+                else if (preferredLanguages.Count > 0)
+                {
+                    // Order the images by preferred languages first.
+                    result = result.OrderBy(i => !string.IsNullOrEmpty(i.Language) && preferredLanguages.Contains(i.Language, StringComparer.OrdinalIgnoreCase) ? 0 : 1);
```
cvium, line 291: I'm not sure. Most providers should already be using https://github.com/jellyfin/jellyfin/blob/e9524f89d63894ea9af62cd1b61ddd89cb8b9e82/MediaBrowser.Model/Extensions/EnumerableExtensions.cs#L19

GodTamIt: Ah yep, that seems to do what we want here anyways. Pretty easy change made.

MediaBrowser.Providers/Manager/ProviderManager.cs
```diff
@@ -276,19 +273,19 @@ public async Task<IEnumerable<RemoteImageInfo>> GetAvailableRemoteImages(BaseIte
...
-                                               preferredLanguages.Contains(i.Language, StringComparer.OrdinalIgnoreCase) ||
+                                               string.Equals(preferredLanguage, i.Language, StringComparison.OrdinalIgnoreCase) ||
                                                string.Equals(i.Language, "en", StringComparison.OrdinalIgnoreCase));
                 }
-                else if (preferredLanguages.Count > 0)
+                else if (hasPreferredLanguage)
                 {
                     // Order the images by preferred languages first.
-                    result = result.OrderBy(i => !string.IsNullOrEmpty(i.Language) && preferredLanguages.Contains(i.Language, StringComparer.OrdinalIgnoreCase) ? 0 : 1);
+                    result = result.OrderByLanguageDescending(preferredLanguage);
```
cvium, line 288: You should remove the call to OrderByLanguageDescending in the tmdb providers and remove the `else if (hasPreferredLanguage)`. The orderby should probably be executed no matter what.

MediaBrowser.Model/Extensions/EnumerableExtensions.cs
```diff
@@ -18,6 +18,13 @@ public static class EnumerableExtensions
         /// <returns>The ordered remote image infos.</returns>
         public static IEnumerable<RemoteImageInfo> OrderByLanguageDescending(this IEnumerable<RemoteImageInfo> remoteImageInfos, string requestedLanguage)
         {
+            if (string.IsNullOrWhiteSpace(requestedLanguage))
+            {
+                // No requested language, so just order by secondary criteria
```
cvium, line 23: Shouldn't we prioritize "en" in that case?

MediaBrowser.Providers/Plugins/Tmdb/BoxSets/TmdbBoxSetImageProvider.cs
```diff
@@ -101,7 +101,7 @@ public async Task<IEnumerable<RemoteImageInfo>> GetImages(BaseItem item, Cancell
                 });
             }
 
-            return remoteImages.OrderByLanguageDescending(language);
+            return remoteImages;
```
cvium, line 104: Forgot to remove `using MediaBrowser.Model.Extensions;` from the providers ;)

cvium (approved):

MediaBrowser.Providers/Manager/ProviderManager.cs
```diff
@@ -235,14 +236,8 @@ public async Task<IEnumerable<RemoteImageInfo>> GetAvailableRemoteImages(BaseIte
 
             var preferredLanguage = item.GetPreferredMetadataLanguage();
 
-            var languages = new List<string>();
-            if (!query.IncludeAllLanguages && !string.IsNullOrWhiteSpace(preferredLanguage))
-            {
-                languages.Add(preferredLanguage);
-            }
-
             // TODO include [query.IncludeAllLanguages] as an argument to the providers
```
crobibero, line 239:
```suggestion
```

I think this `TODO` has been implemented now

crobibero (approved):

MediaBrowser.Providers/Manager/ProviderManager.cs
cvium: No it hasn't

Bond-009 (approved):

Issue #5521, opened 2021-03-13
Jellyfin doesn't stream files if they are symbolic links


pixelomer:
**Bug description**
When a video file (or any other type of media) is placed directly into the media folder, it can be played just fine. However, if a symbolic link is placed instead of the file itself, the video is still recognized by Jellyfin but it becomes unplayable. This issue is not reproducible in Jellyfin 10.6.4 (stable).

**System:**
 - OS: Raspbian Lite
   - `uname -a`: `Linux 5.10.17-v7l+ #1403 SMP Mon Feb 22 11:33:35 GMT 2021 armv7l GNU/Linux`
 - Virtualization: Docker version 20.10.5, build 55c4c88
 - Clients: Infuse for iOS, jellyfin-mpv-shim
 - Browser: Firefox 86.0.1 on macOS
 - Jellyfin Version: 10.7.0 (stable)
   - Docker image: `jellyfin/jellyfin`
   - Tag: `10.7.0`
 - Playback: Direct Play
 - Installed Plugins: none
 - Reverse Proxy: none
 - Base URL: /
 - Networking: `--net=host`
 - Storage: Physically attached SSD mounted in the container using `-v`

**To Reproduce**

1. Put a video file in the media folder (eg. `/media/MyVideo.mkv`)
2. Create a symbolic link to this video file from somewhere else (eg. A TV show directory)
```
mkdir -p /media/shows/MyAmazingShow/Season\ 01
cd /media/shows/MyAmazingShow/Season\ 01
ln -s ../../../MyVideo.mkv S01E01.mkv
```
3. Try playing this video using a client
4. Get either no response or an error

**Expected behavior**
Jellyfin should be able to stream a media file regardless of whether the media folder contains the file itself or a symbolic link to it.

**Logs**
The server does not log anything about this issue, even with debug level logging. The logs below are from `jellyfin-mpv-desktop`.
```
2021-03-13 23:01:40,600 [   DEBUG] media: Using direct url.
2021-03-13 23:01:40,620 [   ERROR] mpv: mkv: Invalid EBML length at position 137
2021-03-13 23:01:40,620 [   ERROR] mpv: mkv: Invalid EBML length at position 139
2021-03-13 23:01:40,620 [ WARNING] mpv: mkv: Unexpected end of file (no clusters found)
2021-03-13 23:01:40,620 [ WARNING] mpv: mkv: mkv metadata beyond end of file - incomplete file?
[...]

PR #5824, opened 2021-04-15
Handle symbolic links as source files

ralf1070:
When replaying content via DLNA 'Microsoft.AspNetCore.Http.SendFileResponseExtensions.SendFileAsyncCore'
validates the length of the sent information. Unfortuanately this check fails for source files
which are symbolic links (aka ReparsePoint's). In this case the size of the filename of the referenced
file is reported, instead of the size of the content of this file.

  * To fix this in AspNetCore.Http would be best, but is out of scope.
  * Second best would be, to resolve the symbolic link target and return this as PhysicalFileResult. This
    would require to add Mono.Posix to this project, which would add another Assembly and which is - as far
    as I know - not portable to Windows and therefore would require conditional compilation.
  * The third way is to use FileStreamResult for files known to be symbolic links, as it correctly reports
    the size of the open file. For all other files PhysicalFileResult is kept, as other parts of this
    project may rely on using it's properties.

PR #5775, opened 2021-04-11
Work around symbolic link issue on Linux

Orochimarufan:
In 10.7.x the direct file streams were switched to PhysicalFileResult. This has introduced a regression because System.IO.FileInfo (and by extension PhysicalFileResultExecutor) don't handle symbolic links correctly.
More specifically, System.IO.FileInfo will report the symlink size instead of the target size. This means PhysicalFileResultExecutor will only ever try to send a few bytes of a symlinked file.
The fix for the upstream issue has been moved back to .NET 6 so it would be nice to work around the regression in the mean time.

This is a working replacement for PhysicalFileResultExecutor based on Mono.Posix which has proper handling of symbolic links. However, it does introduce an additional dependency. It has been tested on amd64 and aarch64 Linux. I'd expect it to work just as well on macOS, provided the platform check is extended, but I have no means of verifying this claim.

An analogous solution might also be viable on Windows, but I don't know enough about Win32 to take a shot at it.
Alternatively, it may be possible to achieve a similar effect using the open-seek-tell workaround.

Note that this is my first time working with ASP.NET, so I'd welcome suggestions.

**Changes**
Add custom ResultExecutor for PhysicalFileResult based on Mono.Posix

**Issues**
Fixes #5521 on Linux

PR #6541, opened 2021-09-11
Read file length for symlinks,

cvium:
**Changes**
Created a subclass of `PhysicalFileResultExecutor` and replaced it in the DI container. It unfortunately reads the file attributes 1-2 times more than its superclass, but `FileInfo` is called multiple times even outside `PhysicalFileResultExecutor`. Therefore I have copied parts of static extensions methods eg. https://github.com/dotnet/aspnetcore/blob/main/src/Http/Http/src/SendFileFallback.cs to get the desired result.

The new executor is only used for symlinks.

Supersedes #5775 and #5824. And a huge thanks for the research done by both PR authors.

**Issues**
Fixes https://github.com/jellyfin/jellyfin/issues/5521

### DIFF
--- a/Jellyfin.Server/Infrastructure/SymlinkFollowingPhysicalFileResultExecutor.cs
+++ b/Jellyfin.Server/Infrastructure/SymlinkFollowingPhysicalFileResultExecutor.cs
@@ -42,7 +42,7 @@ namespace Jellyfin.Server.Infrastructure
         /// <summary>
         /// Initializes a new instance of the <see cref="SymlinkFollowingPhysicalFileResultExecutor"/> class.
         /// </summary>
-        /// <param name="loggerFactory"></param>
+        /// <param name="loggerFactory">An instance of the <see cref="ILoggerFactory"/> interface.</param>
         public SymlinkFollowingPhysicalFileResultExecutor(ILoggerFactory loggerFactory) : base(loggerFactory)
         {
         }
@@ -95,13 +95,15 @@ namespace Jellyfin.Server.Infrastructure
 
             if (range != null)
             {
-                return SendFileAsync(result.FileName,
+                return SendFileAsync(
+                    result.FileName,
                     response,
                     offset: range.From ?? 0L,
                     count: rangeLength);
             }
 
-            return SendFileAsync(result.FileName,
+            return SendFileAsync(
+                result.FileName,
                 response,
                 offset: 0,
                 count: null);
--- a/Jellyfin.Server/Infrastructure/SymlinkFollowingPhysicalFileResultExecutor.cs
+++ b/Jellyfin.Server/Infrastructure/SymlinkFollowingPhysicalFileResultExecutor.cs
@@ -136,7 +136,7 @@ namespace Jellyfin.Server.Infrastructure
 
             fileStream.Seek(offset, SeekOrigin.Begin);
             await StreamCopyOperation
-                .CopyToAsync(fileStream, response.Body, count, bufferSize, CancellationToken.None)
+                .CopyToAsync(fileStream, response.Body, count, BufferSize, CancellationToken.None)
                 .ConfigureAwait(true);
         }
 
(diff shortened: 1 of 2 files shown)


Jellyfin.Server/Infrastructure/SymlinkFollowingPhysicalFileResultExecutor.cs
```diff
@@ -0,0 +1,145 @@
...
+            }
+
+            if (count.HasValue
+                && (count.Value < 0 || count.Value > fileInfo.Length - offset))
+            {
+                throw new ArgumentOutOfRangeException(nameof(count), count, string.Empty);
+            }
+
+            // Copied from SendFileFallback.SendFileAsync
+            const int bufferSize = 1024 * 16;
```
Bond-009, line 127:
```suggestion
            const int BufferSize = 1024 * 16;
```

Jellyfin.Server/Infrastructure/SymlinkFollowingPhysicalFileResultExecutor.cs
```diff
@@ -0,0 +1,145 @@
...
+            var fileInfo = GetFileInfo(filePath);
+            if (offset < 0 || offset > fileInfo.Length)
+            {
+                throw new ArgumentOutOfRangeException(nameof(offset), offset, string.Empty);
+            }
+
+            if (count.HasValue
+                && (count.Value < 0 || count.Value > fileInfo.Length - offset))
+            {
+                throw new ArgumentOutOfRangeException(nameof(count), count, string.Empty);
```
Bond-009, line 123: Pass a message?

Bond-009 (approved):

Jellyfin.Server/Infrastructure/SymlinkFollowingPhysicalFileResultExecutor.cs
cvium: I suppose. This is just copied from aspnetcore

Jellyfin.Server/Infrastructure/SymlinkFollowingPhysicalFileResultExecutor.cs
```diff
@@ -0,0 +1,145 @@
...
+
+            // Copied from SendFileFallback.SendFileAsync
+            const int BufferSize = 1024 * 16;
+
+            await using var fileStream = new FileStream(
+                filePath,
+                FileMode.Open,
+                FileAccess.Read,
+                FileShare.ReadWrite,
+                bufferSize: bufferSize,
```
cvium, line 134:
```suggestion
                bufferSize: BufferSize,
```

Jellyfin.Server/Infrastructure/SymlinkFollowingPhysicalFileResultExecutor.cs
```diff
@@ -0,0 +1,145 @@
...
+                options: (AsyncFile.UseAsyncIO ? FileOptions.Asynchronous : FileOptions.None) | FileOptions.SequentialScan);
+
+            fileStream.Seek(offset, SeekOrigin.Begin);
+            await StreamCopyOperation
+                .CopyToAsync(fileStream, response.Body, count, bufferSize, CancellationToken.None)
```
cvium, line 139:
```suggestion
                .CopyToAsync(fileStream, response.Body, count, BufferSize, CancellationToken.None)
```

crobibero (approved):
Confirmed the `Content-Length` header is the same for direct play (normal), and direct play (symlink)

dekadius:
I'm wondering how this will behave with dangling symlinks. 

The same problem with streaming from symlinks existed in a 10.6 version, and a fix was implemented in [ManagedFileSystem.GetFileSystemMetaData](https://github.com/jellyfin/jellyfin/blob/master/Emby.Server.Implementations/IO/ManagedFileSystem.cs#L244). I imagine that fix is still needed for getting the correct file size to store in the db, but I see that newer versions of that fix handles dangling symlinks. I haven't been able to this PR yet, so this might be a false alarm, but catching the same exception and setting `Exists` to `false` in that case seems logical to me. Also it might be worth noting the difference in how the symlink file stream is opened.

Sorry if this already has been tested / considered.

cvium:
I imagine it will fail, which is alright.

random-human-being:
Can we expect a release with this fix before 10.8? A lot of content on my server is effectively inaccessible with the latest release.

cvium:
Unfortunately not

Issue #4449, opened 2020-11-09
Jellyfin ignoring --ffmpeg CLI option


spacekookie:
**Describe the bug**

I'm running jellyfin on a NixOS system, meaning that my paths don't work the same way as on other Linux systems. For that reason, there is a wrapper script that makes sure to pass the correct `ffmpeg` path to jellyfin via the `--ffmpeg` option.

I have recently run into issues because this path is not persisted in the jellyfin configuration (`/var/lib/jellyfin/config/encoding.xml`), or even used.

It looks like the first time jellyfin started it persisted a path to this config, and has not updated it since. This is now causing issues while transcoding on my system.

**System (please complete the following information):**
 - OS: Linux (NixOS)
 - Virtualization: None
 - Clients: NA
 - Browser: NA
 - Jellyfin Version: current stable (10.6.4)
 - Playback: Transcode
 - Installed Plugins: NA
 - Reverse Proxy: NA
 - Base URL: None
 - Networking: Host
 - Storage: NA

**To Reproduce**

1. Start jellyfin with a path given via the `--ffmpeg` flag
2. Change ffmpeg version, prividing a different path
3. The config has not updated the path

**Expected behavior**

For one, I am surprised that the path provided via the `--ffmpeg` flag is persisted into the configuration at all. But even if it is, I feel like the `--ffmpeg` CLI flag should take precedence over the existing configuration.

**Logs**

Here are the logs from startup: https://paste.rs/V9Q

alarsyo:
This issue still affects me!

joshuaboniface:
Copying this in from the PR for posterity:

Based on some discussion and code analysis, we think we know what's happening here.

1. The `--ffmpeg` option is specified at startup which populates the "default".

2. When the "Transcoding" configuration is loaded in the WebUI, we populate the "default" value of the CLI flag into the UI.

3. If those settings are then saved, the configuration value populated by `--ffmpeg` is then saved into the configuration file (`system.xml`).

4. The `system.xml` config then overrides the `--ffmpeg` option on next startup (by design as mentioned), and thus future changes to `--ffmpeg` are not respected.

Thus ultimately, I think this is a bug in the handling of the configuration UI. If we load in the "default" CLI flag, it shouldn't be populated as if it was entered by the user. How to actually fix that I'm not certain.

[...]

PR #6544, opened 2021-09-11
Don't set ffmpeg path from null to its Display value + add 404

cvium:
**Changes**
If existing path is null it means that it hasn't been set by the user, so don't set it to its Display value.

**Issues**
Fixes https://github.com/jellyfin/jellyfin/issues/4449


crobibero (approved):

MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
```diff
@@ -175,9 +185,14 @@ public void UpdateEncoderPath(string path, string pathType)
                 newPath = path;
             }
 
+            // Don't save an invalid path
+            if (!File.Exists(path))
```
Bond-009, line 189: This doesn't work, right?
I can't enter "ffmpeg" in the web interface bc it will 404

cvium: Should you be able to?

Bond-009: IMO, yes

cvium:
@jellyfin-bot rebase

MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
cvium: Removed

PR #6553, opened 2021-09-12
Comparing PremiereDate when episode comparison would otherwise be equal.

### LINKED ISSUE #6269: AiredEpisodeOrderComparer does not respect dates.
**Describe the bug**
[AiredEpisodeOrderComparer.cs](https://github.com/jellyfin/jellyfin/blob/804614315819c8c17c576b2cdfb683177b59eb7a/Emby.Server.Implementations/Sorting/AiredEpisodeOrderComparer.cs#L155) looks to not compare the air ordering but rather the Parent Index (Season Number) modified by the Index Number (Episode number)

This leads to it being basically the same as absolute order

 - OS: Windows
 - Virtualization: None - Developing
 - Clients: Browser, Curl
 - Browser: Chrome 91.0.4472.77 (Official Build) (64-bit) (cohort: Stable)
 - Jellyfin Version: 1.7.5
 - Playback: n/a
 - Installed Plugins: n/a
 - Reverse Proxy: n/a
 - Base URL: web/index.html#!/details?id=c29715f7fb1f55c8ea99d6f668007266&serverId=5ff7b1beaec741feb30c711dbc0d6ad8 - Networking: Host
 - Storage: local

**To Reproduce**
1) Create a directory with a naming structure as follows
Shows
└── Series
├── YYYYMMDD-Title [ytid].mkv
2) Create a library pointing at this directory
3) Edit episodes to have episode number of 0

**Expected behavior**
It would be expected that since there is no Aired Date property that premier date would be respected. It however is not and you will notice that the ordering is not in the chronological order.

**Server Logs**
n/a

**FFmpeg Logs**
n/a

**Browser Console Logs**
n/a

**Screenshots**
n/a

**Additional context**
This was found during development of my Youtube plugin to add the following feature
https://github.com/ankenyr/jellyfin-youtube-metadata-plugin/issues/7

It seems that at one point PremiereDate was use but was [commented out](https://github.com/jellyfin/jellyfin/blame/726e116d5bb0038edfdaa7f1672fa22df499964b/Emby.Server.Implementations/Sorting/AiredEpisodeOrderComparer.cs#L37) and it is unclear why this is. I believe uncommenting that line would solve the issue but it would likely break other things. What seems like a better fix would be changing the [comparison on line 155](https://github.com/jellyfin/jellyfin/blame/726e116d5bb0038edfdaa7f1672fa22df499964b/Emby.Server.Implementations/Sorting/AiredEpisodeOrderComparer.cs#L155) to do the premiere date compare rather than what it is doing now.

Let me know if this seems sane and I can work on a PR with tests.

ankenyr:
In cases where two episodes have the same episode and season, comparing the PremiereDate to further sort them.

**Issues**
Fixes #6269

### DIFF
--- a/Emby.Server.Implementations/Sorting/AiredEpisodeOrderComparer.cs
+++ b/Emby.Server.Implementations/Sorting/AiredEpisodeOrderComparer.cs
@@ -151,7 +151,6 @@ namespace Emby.Server.Implementations.Sorting
             if (comparisonResult == 0 && x.PremiereDate.HasValue && y.PremiereDate.HasValue)
             {
                 comparisonResult = DateTime.Compare(x.PremiereDate.Value, y.PremiereDate.Value);
-                }
             }
 
             return comparisonResult;
--- a/Emby.Server.Implementations/Sorting/AiredEpisodeOrderComparer.cs
+++ b/Emby.Server.Implementations/Sorting/AiredEpisodeOrderComparer.cs
@@ -154,7 +154,7 @@ namespace Emby.Server.Implementations.Sorting
                 }
             }
 
-            return compare_val;
+            return comparisonResult;
         }
 
         /// <summary>
--- a/Emby.Server.Implementations/Sorting/AiredEpisodeOrderComparer.cs
+++ b/Emby.Server.Implementations/Sorting/AiredEpisodeOrderComparer.cs
@@ -146,12 +146,11 @@ namespace Emby.Server.Implementations.Sorting
         {
             var xValue = ((x.ParentIndexNumber ?? -1) * 1000) + (x.IndexNumber ?? -1);
             var yValue = ((y.ParentIndexNumber ?? -1) * 1000) + (y.IndexNumber ?? -1);
-            var compare_val = xValue.CompareTo(yValue);
-            if (compare_val == 0)
+            var comparisonResult = xValue.CompareTo(yValue);
+            // If equal, compare premiere dates
+            if (comparisonResult == 0 && x.PremiereDate.HasValue && y.PremiereDate.HasValue)
             {
-                if (x.PremiereDate.HasValue & y.PremiereDate.HasValue)
-                {
-                    compare_val = DateTime.Compare(x.PremiereDate.Value, y.PremiereDate.Value);
+                comparisonResult = DateTime.Compare(x.PremiereDate.Value, y.PremiereDate.Value);
                 }
             }
 
--- a/Emby.Server.Implementations/Sorting/AiredEpisodeOrderComparer.cs
+++ b/Emby.Server.Implementations/Sorting/AiredEpisodeOrderComparer.cs
@@ -28,16 +28,6 @@ namespace Emby.Server.Implementations.Sorting
                 throw new ArgumentNullException(nameof(y));
             }
 
-            if (x.PremiereDate.HasValue && y.PremiereDate.HasValue)
-            {
-                var val = DateTime.Compare(x.PremiereDate.Value, y.PremiereDate.Value);
-
-                if (val != 0)
-                {
-                    // return val;
-                }
-            }
-
             var episode1 = x as Episode;
             var episode2 = y as Episode;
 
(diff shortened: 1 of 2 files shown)


Emby.Server.Implementations/Sorting/AiredEpisodeOrderComparer.cs
```diff
@@ -156,8 +146,16 @@ private static int CompareEpisodes(Episode x, Episode y)
         {
             var xValue = ((x.ParentIndexNumber ?? -1) * 1000) + (x.IndexNumber ?? -1);
             var yValue = ((y.ParentIndexNumber ?? -1) * 1000) + (y.IndexNumber ?? -1);
+            var compare_val = xValue.CompareTo(yValue);
+            if (compare_val == 0)
+            {
+                if (x.PremiereDate.HasValue & y.PremiereDate.HasValue)
+                {
+                    compare_val = DateTime.Compare(x.PremiereDate.Value, y.PremiereDate.Value);
```
cvium, line 154:
```suggestion
            var comparisonResult = xValue.CompareTo(yValue);
            // If equal, compare premiere dates
            if (comparisonResult == 0)
            {
                if (x.PremiereDate.HasValue & y.PremiereDate.HasValue)
                {
                    comparisonResult = DateTime.Compare(x.PremiereDate.Value, y.PremiereDate.Value);
```

Emby.Server.Implementations/Sorting/AiredEpisodeOrderComparer.cs
```diff
@@ -156,8 +146,16 @@ private static int CompareEpisodes(Episode x, Episode y)
...
+                }
+            }
 
-            return xValue.CompareTo(yValue);
+            return compare_val;
```
cvium, line 158:
```suggestion
            return comparisonResult;
```

Emby.Server.Implementations/Sorting/AiredEpisodeOrderComparer.cs
```diff
@@ -156,8 +146,16 @@ private static int CompareEpisodes(Episode x, Episode y)
         {
             var xValue = ((x.ParentIndexNumber ?? -1) * 1000) + (x.IndexNumber ?? -1);
             var yValue = ((y.ParentIndexNumber ?? -1) * 1000) + (y.IndexNumber ?? -1);
+            var compare_val = xValue.CompareTo(yValue);
+            if (compare_val == 0)
+            {
+                if (x.PremiereDate.HasValue & y.PremiereDate.HasValue)
+                {
+                    compare_val = DateTime.Compare(x.PremiereDate.Value, y.PremiereDate.Value);
+                }
```
cvium, line 155:
```suggestion
```

ankenyr:
Thanks for the suggestions @cvium I think those make it better. I appreciate it.

Emby.Server.Implementations/Sorting/AiredEpisodeOrderComparer.cs
```diff
@@ -156,8 +146,14 @@ private static int CompareEpisodes(Episode x, Episode y)
         {
             var xValue = ((x.ParentIndexNumber ?? -1) * 1000) + (x.IndexNumber ?? -1);
             var yValue = ((y.ParentIndexNumber ?? -1) * 1000) + (y.IndexNumber ?? -1);
+            var comparisonResult = xValue.CompareTo(yValue);
+            // If equal, compare premiere dates
+            if (comparisonResult == 0 && x.PremiereDate.HasValue && y.PremiereDate.HasValue)
+            {
+                comparisonResult = DateTime.Compare(x.PremiereDate.Value, y.PremiereDate.Value);
```
cvium, line 153: What if there's slight timezone differences? Should the timestamp matter?

ankenyr: Are you asking about x.Timestamp or the PremiereDate as a timestamp since it is Datetime? If the former I am not sure how Timestamp is used but I don't think it has to do with the time the episode was premiered and is more used for encoding. If the later, I am unsure how the sorter would do anything more. The timezone would need to be supplied by the metadata server and be a part of PremiereDate when written to the database.

cvium: I'm asking about the case where the premiere date is coming from different metadata providers where the timestamp may differ.

What I'm asking is if the timestamp is even relevant?

crobibero (approved):

Emby.Server.Implementations/Sorting/AiredEpisodeOrderComparer.cs
ankenyr: I don't believe this would affect things. The timezone if one is available will be accounted for in the comparison. If the agent isn't recording the timezone information or the database provides incorrect data then the sorting function is not at fault here and cannot be made to fix that problem.

cvium: I guess that's fair.

cvium (approved):

PR #6556, opened 2021-09-15
Use standard Authorization header in OpenAPI specification

nielsvanvelzen:
**Changes**
- Changed the security scheme in the OpenAPI document to use `Authorization` instead of `X-Emby-Authorization` as the default authorization header.

Multiple clients have already moved to this new header and more are coming. Using HTTP-standards instead of magic is the recommended way of connecting to your Jellyfin server.

**Issues**


cvium (approved):
I wouldn't call custom headers magic

crobibero (approved):

PR #6558, opened 2021-09-16
Respond on request when ffmpeg fails

dmitrylyzo:
If the video file is totally bad, ffmpeg exits with a non-zero code, but the request remains unresponded.
HLS player spams the `levelLoadTimeOut` error.

**Changes**
Throw on FFmpeg non-zero exit code.

**Issues**
The request remains unresponded when ffmpeg fails.

**Sample**
This video is an HLS-fragment and must not be played.
https://user-images.githubusercontent.com/56478732/133689991-593b52d3-7c95-43fe-9c3d-7fdaa4ad2f9f.mp4

Ideally, we should probably add a client-side timeout as well.

### DIFF
--- a/Jellyfin.Api/Helpers/TranscodingJobHelper.cs
+++ b/Jellyfin.Api/Helpers/TranscodingJobHelper.cs
@@ -607,6 +607,10 @@ namespace Jellyfin.Api.Helpers
             {
                 StartThrottler(state, transcodingJob);
             }
+            else if (transcodingJob.ExitCode != 0)
+            {
+                throw new Exception(string.Format(CultureInfo.InvariantCulture, "FFmpeg exited with code {0}", transcodingJob.ExitCode));
+            }
 
             _logger.LogDebug("StartFfMpeg() finished successfully");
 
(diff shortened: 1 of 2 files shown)


Jellyfin.Api/Helpers/TranscodingJobHelper.cs
```diff
@@ -607,6 +607,10 @@ private void DeleteHlsPartialStreamFiles(string outputFilePath)
             {
                 StartThrottler(state, transcodingJob);
             }
+            else if (transcodingJob.ExitCode != 0)
+            {
+                throw new Exception(string.Format(CultureInfo.InvariantCulture, "FFmpeg exited with code {0}", transcodingJob.ExitCode));
```
crobibero, line 612: I know we shouldn’t throw the base Exception, but I don’t know what would be better here

crobibero (approved):

Jellyfin.Api/Helpers/TranscodingJobHelper.cs
dmitrylyzo: Yeah, there is a `FfmpegException`, but I cannot reach it.

Bond-009: Maybe we should move it to another project? (.Common for example)

PR #6567, opened 2021-09-19
Reduce allocations

Bond-009:


crobibero (approved):
Should add a test for when the needle is found in the haystack multiple times

barronpm (approved):

cvium (approved):

PR #6572, opened 2021-09-20
Fix unit test RecordingHelperTests.GetRecordingName_Success.

Dixin:
**Changes**
The string representation of the date should be local.

### DIFF
--- a/Emby.Server.Implementations/LiveTv/EmbyTV/RecordingHelper.cs
+++ b/Emby.Server.Implementations/LiveTv/EmbyTV/RecordingHelper.cs
@@ -34,16 +34,16 @@ namespace Emby.Server.Implementations.LiveTv.EmbyTV
                 {
                     if (info.OriginalAirDate.Value.Date.Equals(info.StartDate.Date))
                     {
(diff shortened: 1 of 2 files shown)


Bond-009 (changes requested):
IMO a better solution would be setting `DateTimeKind` to `Local` for `OriginalAirDate` in the test data

Dixin:
> IMO a better solution would be setting `DateTimeKind` to `Local` for `OriginalAirDate` in the test data

Thanks. IMO it is better to have non-local date in the test data, and verify the result and see if the test data has been successfully converted to local date.

Bond-009:
I would make that a separate test instead of cramming it in this one, the parsing, converting and formatting isn't really the cleanest solution

Dixin:
> I would make that a separate test instead of cramming it in this one, the parsing, converting and formatting isn't really the cleanest solution

I updated the code to have a DateTime instance and only calls ToLocalTime(), which is the same as RecordingHelper.GetRecordingName().

Bond-009:
#6900

Issue #6385, opened 2021-08-04
Plan to Upgrade to .NET 6


richlander:
If you read my [.NET blog posts](https://devblogs.microsoft.com/dotnet/announcing-net-6-preview-6/), you'll know that we will soon we releasing .NET 6 (November 2021) and .NET 5 will go EOL soonish after (May, 2022). It addition, .NET 6 RC1 will be a Go-Live release, meaning that RC1 (and after) is a high confidence release and supported in production.

I'm naturally biased. I'd suggest moving to .NET 6 sooner than later. Perhaps it makes sense to switch to .NET 6 in containers first since that's a more isolated scenario.

I don't know Jellyfin well enough to suggest which features make sense to adopt first, to take advantage of the new release. It might depend on the distribution type. For example, I'm going to suggest adopting ready-to-run compilation in (at least) the container images you ship. It appears that this feature improves startup time by ~1.5s from my testing.

Somewhat separately, .NET 6 Debian container images are based off of Bullseye instead of Buster (what you are using now). You could consider moving to Bullseye at the same time.

Anyone interested in targeting the RC1 release (when it comes)? Certainly, if Jellyfin adopted RC1 or RC2, I'd want to talk about the project in the .NET 6 final release blog post to draw attention to an early adopter.

What are your thoughts on moving to .NET 6?

dkanada:
If I recall correctly, we switched to .NET 5 extremely fast, so I don't think there will be any pushback from another quick update to the .NET 6 release candidates. Just for reference, what does .NET 6 support look like on Ubuntu?

crobibero:
We’re eager to move to .NET 6, we even have a branch where @Bond-009 was playing with the upgrade

https://github.com/jellyfin/jellyfin/tree/dotnet6

I believe we decided against ready to run as it increases the compiled size for a minimal startup increase- and since this is a server application we don’t expect the server to be restarted regularly. I could be remembering incorrectly, but that’s what I remember the response being last time I brought up ready to run.

It would be interesting to move to .NET 6 for release 10.8, but I’m not sure when our next release is scheduled for

richlander:
Cool. That sounds great.

I can get you the size increase. We could also push that out a little. We have a new [PGO](https://devblogs.microsoft.com/dotnet/conversation-about-pgo/) tech that can be used to significantly reduce the size of crossgen code.

[...]

PR #6590, opened 2021-09-21
Target net6.0

Bond-009:


MediaBrowser.Model/Net/MimeTypes.cs
```diff
@@ -164,15 +165,16 @@ public static class MimeTypes
...
+        public static string GetMimeType(string path) => GetMimeType(path, "application/octet-stream");
 
         /// <summary>
         /// Gets the type of the MIME.
         /// </summary>
         /// <param name="filename">The filename to find the MIME type of.</param>
-        /// <param name="enableStreamDefault">Whether of not to return a default value if no fitting MIME type is found.</param>
-        /// <returns>The worrect MIME type for the given filename, or `null` if it wasn't found and <paramref name="enableStreamDefault"/> is false.</returns>
-        public static string? GetMimeType(string filename, bool enableStreamDefault)
+        /// <param name="defaultValue">Theefault value to return if no fitting MIME type is found.</param>
```
barronpm, line 174:
```suggestion
        /// <param name="defaultValue">The default value to return if no fitting MIME type is found.</param>
```

Emby.Server.Implementations/Data/SqliteItemRepository.cs
```diff
@@ -1647,18 +1644,16 @@ IEnumerable<MetadataField> GetLockedFields(string s)
...
-                                if (Enum.TryParse(i, true, out TrailerType parsedValue))
-                                {
-                                    yield return parsedValue;
-                                }
+                                (types ??= new List<TrailerType>()).Add(parsedValue);
                             }
                         }
 
-                        trailer.TrailerTypes = GetTrailerTypes(trailerTypes).ToArray();
+                        trailer.TrailerTypes = types.ToArray() ?? Array.Empty<TrailerType>();
```
barronpm, line 1656: Couldn't `types` be null here?

barronpm:
Fixes #6385

crobibero (approved):

Jellyfin.Server/Middleware/UrlDecodeQueryFeature.cs
```diff
@@ -52,18 +52,8 @@ public IQueryCollection Query
...
-                if (string.Equals(unencodedKey, key, StringComparison.Ordinal))
-                {
-                    // Don't do anything if it's not encoded.
-                    _store = value;
-                    return;
-                }
-
                 var pairs = new Dictionary<string, StringValues>();
-                var queryString = unencodedKey.SpanSplit('&');
+                var queryString = HttpUtility.UrlDecode(key).SpanSplit('&');
```
cvium, line 56: You're wasting an allocation (dictionary) here if there's no ampersand in the query string

MediaBrowser.MediaEncoding/Subtitles/SubtitleEditParser.cs
```diff
@@ -38,7 +38,7 @@ public SubtitleTrackInfo Parse(Stream stream, CancellationToken cancellationToke
             subRip.LoadSubtitle(subtitle, lines, "untitled");
             if (subRip.ErrorCount > 0)
             {
-                _logger.LogError("{ErrorCount} errors encountered while parsing subtitle.");
+                _logger.LogError("{ErrorCount} errors encountered while parsing subtitle.", subRip.ErrorCount);
```
cvium, line 41: I believe logs should not end with full stops?

RSSDP/SsdpDevice.cs
```diff
@@ -134,11 +135,13 @@ public string FullDeviceType
         {
             get
             {
-                return String.Format("urn:{0}:{3}:{1}:{2}",
-                this.DeviceTypeNamespace ?? String.Empty,
-                this.DeviceType ?? String.Empty,
-                this.DeviceVersion,
-                this.DeviceClass ?? "device");
+                return String.Format(
```
cvium, line 138: Shouldn't this be `string`?

RSSDP/SsdpDevicePublisher.cs
```diff
@@ -542,15 +543,15 @@ private Task SendByeByeNotification(SsdpDevice device, string notificationType,
...
+            values["SERVER"] = string.Format(CultureInfo.InvariantCulture, "{0}/{1} UPnP/1.0 RSSDP/{2}", _OSName, _OSVersion, ServerVersion);
             values["NTS"] = "ssdp:byebye";
             values["NT"] = notificationType;
             values["USN"] = uniqueServiceName;
 
             var message = BuildMessage(header, values);
 
             var sendCount = IsDisposed ? 1 : 3;
-            WriteTrace(String.Format("Sent byebye notification"), device);
+            WriteTrace(String.Format(CultureInfo.InvariantCulture, "Sent byebye notification"), device);
```
cvium, line 554: `string`

RSSDP/SsdpDevice.cs
Bond-009: Old RSSDP code, out of scope for this PR

RSSDP/SsdpDevicePublisher.cs
Bond-009: Same

RSSDP/SsdpDevice.cs
cvium: Old or not, it's still a changed line. Why not just fix it?

Jellyfin.Server/Middleware/UrlDecodeQueryFeature.cs
Bond-009: Don't think this is true, if we only encoded a single value there wouldn't be an `&`We need to check for an `=`, if there isn't one we can return early

MediaBrowser.MediaEncoding/Subtitles/SubtitleEditParser.cs
Bond-009: Correct

RSSDP/SsdpDevice.cs
Bond-009: Because that codebase is full of this stuff, if you want you can change all of them at once.
I haven't because it's always been my plan to change back to upstream as soon as we can.
This change was only to make it compile on .net6

cvium (approved):

PR #6612, opened 2021-09-25
Remove sync FileStream hack

Bond-009:


crobibero (approved):

barronpm:
Should we also remove the hacks around dotnet/runtime#42790 now that it's fixed?

cvium (approved):

PR #4432, opened 2020-11-07
Add initial support for HEVC over FMP4-HLS

nyanmisaka:
**Web changes**
https://github.com/jellyfin/jellyfin-web/pull/2064

**Changes**
- Initial support for streaming FMP4-HLS.
- Direct streaming over FMP4 supports HEVC, H264, AAC, MP3, AC3, EAC3, FLAC, ALAC.
- HEVC Main 10 SDR direct streaming on supported devices.
- HEVC Main 10 HDR/HLG direct streaming on supported devices. (e.g. Apple A11 chip devices with iOS 14.1+ installed)
- Transcode and streaming in HEVC format. Compared with H264, the same quality can be achieved at a lower bandwidth.
- HWA HEVC encoding in NVENC, QSV, AMF and VAAPI.
- Enhance audio bitrate probing.
- Add an option to disable HEVC encoding.

AF65EEE7-6FCD-463F-8EDD-21D3ACF45234
9CE4C900-5B8F-498D-A76F-643B156FBA06
C37323D4-A838-4BA7-93F6-DF378547040B

PR #6617, opened 2021-09-26
FIX: cannot compare bitstream of different video codec

koying:
**Changes**

It doesn't really make sense to compare bitstream rate of 2 different codec.
Not downscaling because, e.g., a hevc 9Mb/s is lower than a h264 10Mb/s is comparing pears and apples.

Partial revert of https://github.com/jellyfin/jellyfin/pull/4432

**Issues**

### DIFF
--- a/Jellyfin.Api/Helpers/StreamingHelpers.cs
+++ b/Jellyfin.Api/Helpers/StreamingHelpers.cs
@@ -198,36 +198,14 @@ namespace Jellyfin.Api.Helpers
 
                 if (!EncodingHelper.IsCopyCodec(state.OutputVideoCodec) && state.OutputVideoBitrate.HasValue)
                 {
-                    var isVideoResolutionNotRequested = !state.VideoRequest.Width.HasValue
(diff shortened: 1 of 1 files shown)


Bond-009 (approved):
This makes sense, I think?

cvium (approved):

nyanmisaka (changes requested):
https://github.com/jellyfin/jellyfin/blob/d75c699ad4568fb2546588497c201fd6e4830477/MediaBrowser.Model/Dlna/ResolutionNormalizer.cs#L28-L39

Normalize() is used to downscale the resolution according to the hardcoded ResolutionConfigurations instead of checking the codecs, which is not reliable in some cases.

e.g. removing this will cause H.264 High10 1080p 4M video to be transcoded to H.264 720p 4M if the video resolution is not requested and the target bitrate is set to 4M. Downscaling is pointless in this case.

Bond-009:
Closing, see comment above

Issue #6560, opened 2021-09-17
Jellyfin fails to update plugins since upgrading to 10.8.0: `Directory does not exist to extract to`


hasezoey:
**Describe the bug**
Since upgrading to jellyfin 10.8.0 (docker `jellyfin/jellyfin:unstable`), plugin updates fail with `Directory does not exist to extract to`

**System (please complete the following information):**
 - OS:  Ubuntu Server 21.04
 - Virtualization: Docker
 - Clients: Browser
 - Browser: Firefox 92
 - Jellyfin Version: 10.8.0 (`jellyfin/jellyfin:unstable@sha256:b476a2e817200326c8dc10a3434e26a8c4d5193defe2f94b4abd8fde5fc2ef81`)
 - FFmpeg Version: [e.g. 4.3.2-Jellyfin]
 - Playback: [Direct Play, Remux, Direct Stream, Transcode] 
 - Hardware Acceleration: [e.g. none, VAAPI, NVENC, etc.]
 - Installed Plugins: `Anilist, AudioDB, Bookshelf, Cover Art Archive, MusicBrainz, OMDb, TMDb, TheTVDB`
 - Reverse Proxy: nginx
 - Base URL: none (`/` redirects to `/web`, setup like in the jellyfin-nginx example)
 - Networking: host
 - Storage: local

**To Reproduce**
1. create server with a lower jellyfin version than 10.8.0
2. install plugins (in my case `Anilist` (version 2.0.0)
3. stop server
4. upgrade to jellyfin to 10.8.0
5. observer logs

**Expected behavior**
Updates to not fail

**Server Logs**

<details>
<summary>The log part that is about the issue</summary>

```txt
[2021-09-17 14:33:05.904 +00:00] [INF] [1] Main: Startup complete 0:00:03.3366366
[2021-09-17 14:33:08.318 +00:00] [INF] [18] Emby.Server.Implementations.ScheduledTasks.TaskManager: "StartupTrigger" fired for task: "Update Plugins"
[2021-09-17 14:33:08.321 +00:00] [INF] [18] Emby.Server.Implementations.ScheduledTasks.TaskManager: Queuing task "PluginUpdateTask"
[2021-09-17 14:33:08.328 +00:00] [INF] [15] Emby.Server.Implementations.ScheduledTasks.TaskManager: Executing "Update Plugins"
[...]

PR #6618, opened 2021-09-26
Create output directory when extracting archive files

crobibero:
Fixes https://github.com/jellyfin/jellyfin/issues/6560
Exception was added in https://github.com/adamhathcock/sharpcompress/pull/614


barronpm (approved):

cvium (approved):
I disagree very much with the upstream change.

PR #6644, opened 2021-10-01
Enable nullable for ItemDataProvider

Bond-009:


Emby.Server.Implementations/LiveTv/EmbyTV/ItemDataProvider.cs
```diff
@@ -62,7 +68,7 @@ private void EnsureLoaded()
 
         private void SaveList()
         {
-            Directory.CreateDirectory(Path.GetDirectoryName(_dataPath));
+            Directory.CreateDirectory(Path.GetDirectoryName(_dataPath) ?? throw new ArgumentException("Path can't be a root directory.", nameof(_dataPath)));
```
cvium, line 71: Should we use ThrowIfNull instead?

cvium (approved):

crobibero (approved):

Emby.Server.Implementations/LiveTv/EmbyTV/ItemDataProvider.cs
Bond-009: No, `_dataPath` wasn't necessarily `null` so I don't think throwing a `ArgumentNullException` is the right thing to do.

PR #6663, opened 2021-10-04
Reference dotnet6-rc1 packages

crobibero:
Gotta get ahead of the game


Jellyfin.Server.Implementations/Users/UserManager.cs
```diff
@@ -701,6 +701,11 @@ public async Task UpdatePolicyAsync(Guid userId, UserPolicy policy)
         /// <inheritdoc/>
         public async Task ClearProfileImageAsync(User user)
         {
+            if (user.ProfileImage == null)
```
crobibero, line 704: `dbContext.Remove(TEntity entity)` now has nullability!

PR #6680, opened 2021-10-09
Add Docker Healthcheck

### LINKED ISSUE #5760: No CURL or WGET in official image to enable health-check
**Describe the bug**
There is a health-check api in jellyfin. However, it seems we can't check the status with official docker image.

 - Virtualization: Docker
 - Jellyfin Version: Docker latest image, 20210322

**Expected behavior**
Add curl or wget package to official docker image.

MarcelCoding:
**Changes**
Added Docker Healthcheck to ensure jellyfin is always available. (e.g. This enables reverse proxies (traefik) to remove jellyfin from the routes if it is overloaded.)

**Issues**
Fixed #5760 

PS: squash the pr, if the signature should be correct :) I did it in a codespace.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -90,4 +90,4 @@ ENTRYPOINT ["./jellyfin/jellyfin", \
     "--ffmpeg", "/usr/lib/jellyfin-ffmpeg/ffmpeg"]
 
 HEALTHCHECK --interval=30s --timeout=30s --start-period=10s --retries=3 \
-     CMD curl -L "${HEALTHCHECK_URL}" || exit 1
+     CMD curl -Lk "${HEALTHCHECK_URL}" || exit 1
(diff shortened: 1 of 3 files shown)


cvium:
> 
> 
> This also may not work if the server the server has `RequireHttps` enabled

@crobibero  Won't it also break if the server isn't listening on localhost?

crobibero:
> > This also may not work if the server the server has `RequireHttps` enabled
> 
> @crobibero Won't it also break if the server isn't listening on localhost?

Yeah that would also not work… maybe we should just have the health check url default to local host:8096 but allow it to be completely overridden using an ENV variable

MarcelCoding:
How about the `JELLYFIN_PublishedServerUrl` env var?

MarcelCoding:
I personally use a reverse proxy but without https between jellyfin and the proxy. But I could test it.

crobibero:
I believe that’s how most people have Jellyfin set up. Additionally, with 10.8 (unstable) there is an automatic redirect to remove the baseUrl when getting health status

MarcelCoding:
Als long a ceritificate w

MarcelCoding:
I also would like to add the option `-L` to curl that it follows redirects, just in case someone misses something in the configuration.

crobibero:
xxx/health -> xxx/baseurl/health
Implementation here: https://github.com/jellyfin/jellyfin/blob/master/Jellyfin.Server/Middleware/BaseUrlRedirectionMiddleware.cs

PR #6748, opened 2021-10-31
Add pt-pt as culture, fix localization typo

### LINKED ISSUE #6747: Pt-pt locale not selectable as Library Prefeered language / Metadata language
**Describe the bug**
As title describes, see image. Only pt and pt-br available, not pt-pt

 - OS: RBPI
 - Virtualization: Docker
 - Clients: Web, Android.
 - Browser: Firefox
 - Jellyfin Version: docker:latest
 - FFmpeg Version: [e.g. 4.3.2-Jellyfin]
 - Playback: [Direct Play, Remux, Direct Stream, Transcode] 
 - Hardware Acceleration: [e.g. none, VAAPI, NVENC, etc.]
 - Installed Plugins: [e.g. none, Fanart, Anime, etc.]
 - Reverse Proxy: [e.g. none, nginx, apache, etc.]
 - Base URL: [e.g. none, yes: /example]
 - Networking: [e.g. Host, Bridge/NAT]
 - Storage: [e.g. local, NFS, cloud]

**To Reproduce**

1. Go to library settings / metadata preference in dashboard -> libraries
2. Click on preferred language
3. Scroll down to P- letter

**Expected behavior**
See option to select Portuguese (portugal), present in /Emby.Server.Implementations/Localization/Core/pt-PT.json

**Screenshots**
imagem

ptalmeida:
**Changes**

* Add pt-pt as culture. This does not solve the problem with portuguese locales as described [here](https://github.com/jellyfin/jellyfin-web/pull/3114#issuecomment-954344003) in Jellyfin-web. Before this P
* Fix typo with Réunion

**Issues**

Possibly fix #6747, not tested.

### DIFF
--- a/Emby.Server.Implementations/Localization/iso6392.txt
+++ b/Emby.Server.Implementations/Localization/iso6392.txt
@@ -349,7 +349,8 @@ pli||pi|Pali|pali
 pol||pl|Polish|polonais
 pon|||Pohnpeian|pohnpei
 por||pt|Portuguese|portugais
-pob||pt-br|Portuguese (Brazil)|portugais
+pop||pt-pt|Portuguese (Portugal)|portugais (pt-pt)
+pob||pt-br|Portuguese (Brazil)|portugais (pt-br)
 pra|||Prakrit languages|prâkrit, langues
 pro|||Provençal, Old (to 1500)|provençal ancien (jusqu'à 1500)
 pus||ps|Pushto; Pashto|pachto
--- a/Emby.Server.Implementations/Localization/countries.json
+++ b/Emby.Server.Implementations/Localization/countries.json
@@ -630,7 +630,7 @@
         "TwoLetterISORegionName": "MD"
     },
     {
-        "DisplayName": "RÃ©union",
+        "DisplayName": "Réunion",
         "Name": "RE",
         "ThreeLetterISORegionName": "REU",
         "TwoLetterISORegionName": "RE"


crobibero (approved):

Emby.Server.Implementations/Localization/iso6392.txt
```diff
@@ -349,7 +349,8 @@ pli||pi|Pali|pali
 pol||pl|Polish|polonais
 pon|||Pohnpeian|pohnpei
 por||pt|Portuguese|portugais
-pob||pt-br|Portuguese (Brazil)|portugais
+pop||pt-pt|Portuguese (Portugal)|portugais (pt-pt)
```
cvium, line 352: Where did you get pop? When I look it up I get this https://iso639-3.sil.org/code/pop

ptalmeida: Like many others on that file, pob is a made up code (https://iso639-3.sil.org/code/pob).

 I just followed suit for a quick fix before #6750 can get done.

ptalmeida: ISO 639 really shouldn't be used for this, instead I believe we should be mirroring [Weblate country data](https://github.com/WeblateOrg/language-data), but that is a different issue.

cvium: I don't know. But you are editing a file called `iso6392.txt` (aka. ISO639-2), so I don't feel great about adding a bogus three letter code.

ptalmeida: Half of the file is bogus (take a look at French Canadian and Brazilian Portuguese per example). This standard isn't made for localisation and somewhere along the line instead of fixing the implementation it simply got these additions.


This is meant as a way to enable European Portuguese to be used, not a permanent fix as I've mentioned in the issue.

https://www.loc.gov/standards/iso639-2/php/code_list.php here is the unadulterated source.

cvium: > This standard isn't made for localisation and somewhere along the line instead of fixing the implementation it simply got these additions.

What do you propose?

cvium (approved):

Emby.Server.Implementations/Localization/iso6392.txt
ptalmeida: In the eventual PR addressing #6750, to replace this file entirely with this https://github.com/WeblateOrg/language-data/blob/main/languages.csv. This also contains the info needed for handling of plural forms in localization, which is an issue with Jellyfin as mentioned [here](https://github.com/jellyfin/jellyfin-vue/issues/1354).

As for this PR, it's just a quick fix so I can get my instance loading the correct translation.

PR #6752, opened 2021-11-01
Add Tests, Fix metadata refresh deletes backgrounds

### LINKED ISSUE #6310: Metadata refresh deletes existing background images when there are non to fetch
**Describe the bug**

Refreshing metadata of a TV Show season with "Replace existing images" enabled will delete pre existing background images when there are non to fetch.  

I am manually organizing UFC PPV in a tv show library with the metadata locked on the tv show so I can manually organize the show without Jellyfin making any changes to the metadata.  The season numbers and episodes numbers do not correspond with any metadata provider.  Because I lock the series from updating nfo metadata, no thumbnails are generated when I add content to the server (which is expected).  I do run the refresh metadata with replace existing images enabled so the TV Screen Grabber can grab a screenshot thumbnail for the episodes I add.  The main issue is that Jellyfin is deleting a background image i already have in place when i run the refresh metadata.  All other primary and thumbnail images are NOT deleted.  Only the background image is deleted on refresh.

 - OS: Debian Buster
 - Virtualization: Docker linuxserver.io
 - Jellyfin Version: 10.7.6
 - Storage: NFS

1337joe:
**Changes**
 - Add unit tests and documentation for `ItemImageProvider`
 - Fix failing tests:
   - Refreshing all images when the provider didn't have new images resulted in multi-image types (background, screenshots) being deleted while single-image types kept the old value.
   - Multi-image types could delete invalid paths without reporting changed, other image types had better tracking.
 - Refactor validation to use existing `BaseItem.ValidateImages` method up front instead of validating images piecemeal throughout the add/update handling.

**Issues**
Fixes #6310

### DIFF
--- a/MediaBrowser.Providers/Manager/ItemImageProvider.cs
+++ b/MediaBrowser.Providers/Manager/ItemImageProvider.cs
@@ -1,7 +1,5 @@
 #nullable disable
 
-#pragma warning disable CA1002, CS1591
-
 using System;
 using System.Collections.Generic;
 using System.IO;
@@ -25,6 +23,9 @@ using Microsoft.Extensions.Logging;
 
 namespace MediaBrowser.Providers.Manager
 {
+    /// <summary>
+    /// Utilities for managing images attached to items.
+    /// </summary>
     public class ItemImageProvider
     {
         private readonly ILogger _logger;
@@ -47,6 +48,12 @@ namespace MediaBrowser.Providers.Manager
             ImageType.Thumb
         };
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ItemImageProvider"/> class.
+        /// </summary>
+        /// <param name="logger">The logger.</param>
+        /// <param name="providerManager">The provider manager for interacting with provider image references.</param>
+        /// <param name="fileSystem">The filesystem.</param>
         public ItemImageProvider(ILogger logger, IProviderManager providerManager, IFileSystem fileSystem)
         {
             _logger = logger;
@@ -54,6 +61,13 @@ namespace MediaBrowser.Providers.Manager
             _fileSystem = fileSystem;
         }
 
+        /// <summary>
+        /// Verifies existing images have valid paths and adds any new local images provided.
+        /// </summary>
+        /// <param name="item">The <see cref="BaseItem"/> to validate images for.</param>
+        /// <param name="providers">The providers to use, must include <see cref="ILocalImageProvider"/>(s) for local scanning.</param>
+        /// <param name="directoryService">The directory service for <see cref="ILocalImageProvider"/>s to use.</param>
+        /// <returns><c>true</c> if changes were made to the item; otherwise <c>false</c>.</returns>
         public bool ValidateImages(BaseItem item, IEnumerable<IImageProvider> providers, IDirectoryService directoryService)
         {
             var hasChanges = false;
@@ -73,6 +87,15 @@ namespace MediaBrowser.Providers.Manager
             return hasChanges;
         }
 
+        /// <summary>
+        /// Refreshes from the providers according to the given options.
+        /// </summary>
+        /// <param name="item">The <see cref="BaseItem"/> to gather images for.</param>
+        /// <param name="libraryOptions">The library options.</param>
+        /// <param name="providers">The providers to query for images.</param>
+        /// <param name="refreshOptions">The refresh options.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <returns>The refresh result.</returns>
         public async Task<RefreshResult> RefreshImages(
             BaseItem item,
             LibraryOptions libraryOptions,
@@ -118,7 +141,7 @@ namespace MediaBrowser.Providers.Manager
         }
 
         /// <summary>
-        /// Refreshes from provider.
+        /// Refreshes from a dynamic provider.
         /// </summary>
         private async Task RefreshFromProvider(
             BaseItem item,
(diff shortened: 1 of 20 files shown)


tests/Jellyfin.Providers.Tests/Manager/ItemImageProviderTests.cs
```diff
@@ -0,0 +1,683 @@
...
+                    }
+                }
+            };
+        }
+
+        // Create a class that implements IHasScreenshots for testing since no BaseItem class is also IHasScreenshots
+        private class MovieWithScreenshots : Movie, IHasScreenshots
+        {
+            // No contents
+        }
```
1337joe, line 681: This is now the only class in my workspace that is a `BaseItem` that implements `IHasScreenshots`.

Assuming it's not used in some other plugin that I don't have checked out, should either the handling for `IHasScreenshots` be removed or the interface itself be removed as unused?

crobibero: I wouldn't remove the interface, we never know if there is a plugin that implements it.

MediaBrowser.Providers/Manager/ItemImageProvider.cs
```diff
@@ -73,31 +87,42 @@ public bool ValidateImages(BaseItem item, IEnumerable<IImageProvider> providers,
...
+        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <returns>The refresh result.</returns>
         public async Task<RefreshResult> RefreshImages(
             BaseItem item,
             LibraryOptions libraryOptions,
             List<IImageProvider> providers,
             ImageRefreshOptions refreshOptions,
             CancellationToken cancellationToken)
         {
+            List<ItemImageInfo> oldBackdropImages = new List<ItemImageInfo>();
```
cvium, line 106: You don't seem to add anything to the list, so just use an Array and save an allocation (hint: `Array.Empty<ItemImageInfo>() is free).

MediaBrowser.Providers/Manager/ItemImageProvider.cs
```diff
@@ -114,11 +139,22 @@ public bool ValidateImages(BaseItem item, IEnumerable<IImageProvider> providers,
                 }
             }
 
+            // only delete existing multi-images if new ones were added
+            if (oldBackdropImages.Count > 0 && oldBackdropImages.Count < item.GetImages(ImageType.Backdrop).Count())
+            {
+                PruneImages(item, oldBackdropImages);
```
cvium, line 145: What if some of the new images have overwritten the old ones? Can that happen?

1337joe: It cannot: the image saver always adds at a new index when `item.AllowsMultipleImages` and the checks to update in place only trigger when the same index is used.

While reviewing the code with your question in mind I did find a very specific edge case where backgrounds in both the item and the remote provider might not be re-downloaded and would then be pruned after updates completed, but I couldn't trigger it testing against live data. I pushed the fix for that and I'll add a test case tomorrow to exercise that path and prove it works as intended.

MediaBrowser.Providers/Manager/ItemImageProvider.cs
1337joe: It may be free to create an empty array, but BaseItem.RemoveImages (where this is used in PruneImages) requires a List and it seemed simpler and more readable to stick with a List type all the way through the handling of this.

Alternately, if you want me to change the `BaseItem.RemoveImages(List<ItemImageInfo> deletedImages)` call to `BaseItem.RemoveImages(IEnumerable<ItemImageInfo> deletedImages)` and use array throughout I can, I just was hesitant to mess with the interface there.

cvium: The only usage in the server for `BaseItem.RemoveImages` is in `ItemImageProvider`, so change away. It also shouldn't break any plugins if they were to use it as well as `IEnumerable` is implemented by IList

tests/Jellyfin.Providers.Tests/Manager/ItemImageProviderTests.cs
```diff
@@ -0,0 +1,692 @@
...
+            var changed = itemImageProvider.ValidateImages(new MovieWithScreenshots(), new List<ILocalImageProvider>(), null);
+
+            Assert.False(changed);
+        }
+
+        private static TheoryData<ImageType, int> GetImageTypesWithCount()
+        {
+            var theoryTypes = new TheoryData<ImageType, int>();
+
+            // shotgun approach; overkill for frequent runs
```
cvium, line 55: Shouldn't this just be removed?

cvium (approved):

tests/Jellyfin.Providers.Tests/Manager/ItemImageProviderTests.cs
1337joe: Good point. I started writing tests to understand everything ItemImageProvider was doing, so I saw value in hitting all the types, but with the class documented and a little improved I should clean up the tests.

Deleted the commented code and merged a couple tests using test-specific InlineData to reduce test duplication while maintaining coverage.

tests/Jellyfin.Providers.Tests/Manager/ItemImageProviderTests.cs
Bond-009: I'm fine with removing it, AFAIK it was only used by the `Game` item type https://github.com/MediaBrowser/Emby.Common/blob/6ef8705ed609666555894c698dfc140fd8314831/MediaBrowser.Controller/Entities/Game.cs#L12

crobibero (approved):

PR #6763, opened 2021-11-01
Clean up duplicate TmdbImageProvider code

1337joe:
**Changes**
- Extract copy-pasted code to convert `ImageData` to `RemoteImageInfo` and add test for new method
- Extract boilerplate GetTypeUrl code into a private helper method.

I was looking at the TMDb plugin to add logo support (next TMDbLib release should support logos) and I noticed there were some copy/paste errors (missing language where the API returns it, using `GetPosterUrl` for a backdrop, etc). Low impact, but seeing as the code showed proof that the duplication was hurting maintainability I figured it was worth cleaning up now.

### DIFF
--- a/MediaBrowser.Providers/Plugins/Tmdb/BoxSets/TmdbBoxSetImageProvider.cs
+++ b/MediaBrowser.Providers/Plugins/Tmdb/BoxSets/TmdbBoxSetImageProvider.cs
@@ -67,40 +67,12 @@ namespace MediaBrowser.Providers.Plugins.Tmdb.BoxSets
                 return Enumerable.Empty<RemoteImageInfo>();
             }
 
-            var remoteImages = new List<RemoteImageInfo>();
+            var posters = collection.Images.Posters;
+            var backdrops = collection.Images.Backdrops;
+            var remoteImages = new List<RemoteImageInfo>(posters.Count + backdrops.Count);
 
-            for (var i = 0; i < collection.Images.Posters.Count; i++)
-            {
-                var poster = collection.Images.Posters[i];
-                remoteImages.Add(new RemoteImageInfo
-                {
-                    Url = _tmdbClientManager.GetPosterUrl(poster.FilePath),
-                    CommunityRating = poster.VoteAverage,
-                    VoteCount = poster.VoteCount,
-                    Width = poster.Width,
-                    Height = poster.Height,
-                    Language = TmdbUtils.AdjustImageLanguage(poster.Iso_639_1, language),
-                    ProviderName = Name,
-                    Type = ImageType.Primary,
-                    RatingType = RatingType.Score
-                });
-            }
-
-            for (var i = 0; i < collection.Images.Backdrops.Count; i++)
-            {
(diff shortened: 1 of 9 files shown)


MediaBrowser.Providers/Plugins/Tmdb/BoxSets/TmdbBoxSetImageProvider.cs
```diff
@@ -67,40 +67,12 @@ public async Task<IEnumerable<RemoteImageInfo>> GetImages(BaseItem item, Cancell
...
-                    VoteCount = backdrop.VoteCount,
-                    Width = backdrop.Width,
-                    Height = backdrop.Height,
-                    ProviderName = Name,
-                    Type = ImageType.Backdrop,
-                    RatingType = RatingType.Score
-                });
-            }
+            TmdbUtils.ConvertToRemoteImageInfo(posters, _tmdbClientManager.GetPosterUrl, ImageType.Primary, language, remoteImages);
+            TmdbUtils.ConvertToRemoteImageInfo(backdrops, _tmdbClientManager.GetBackdropUrl, ImageType.Backdrop, language, remoteImages);
```
cvium, line 75: This seems equally error prone. I would perhaps create separate methods per image type.

1337joe: I found this easier to review than it used to be, but I can definitely see the argument for separate methods being even better.

Unfortunately, with `ConvertToRemoteImageInfo` private and only accessible through those new methods this falls into the bucket of things that aren't readily unit testable (without hitting the TMDb API during tests) due to the tight coupling to the TmdbClient. At least doing it this way drops the complexity compared to the original.

cvium: I have a couple of thoughts on that:

1. We use `internal` instead of `private` for the methods we wish to test. Some might consider it bad practice, but 🤷 
2. I don't like the tmdbutils class much. We could replace the methods `GetStillUrl` etc. with `ConvertToRemoteImageInfo`

crobibero (approved):

MediaBrowser.Providers/Plugins/Tmdb/BoxSets/TmdbBoxSetImageProvider.cs
1337joe: It's now more tightly coupled to the live tmdb client (and therefore not testable), but option 2 now seems like the obvious place that I should have put it in the first place. I couldn't delete all the `GetProfileUrl` methods because they're used in a couple places that aren't actually fetching the images, but I like this iteration.

cvium: It's generally tightly coupled by design. I was debating whether to use an adapter interface when I migrated it to TMDBLib but ultimately it felt like a waste of time.

It looks good now.

cvium (approved):

PR #6778, opened 2021-11-04
Add a bit of hardening to the systemd service

jvoisin:
**Changes**
Add a bit of hardening to the debian systemd unit.
Tested in an unprivileged lxc container, so it shouldn't™ break anything.


joshuaboniface (approved):
Seems fine on native Debian too.

crobibero (approved):

PR #6779, opened 2021-11-04
Generalize handling of multi-image types

1337joe:
I missed the window to comment on #6769 (didn't actually see it until I was half way through writing the code for this as my own approach to removing `IHasScreenshot` handling).

The way `IHasScreenshot` was removed from `ItemImageProvider` results in no way for a remote provider to return screenshots but left the code that determined how many should be downloaded and handled replacing them if any were found.

@Bond-009 If you would rather we not support screenshots at all I can revert this and instead strip out screenshot handling. Otherwise, I think this is a cleaner way to handle multi-images than writing special handling for each type.

**Changes**
- Replace specific handling for `Backdrop` and `Screenshot` with general handling for all image types that can supply multiple images.

### DIFF
--- a/MediaBrowser.Providers/Manager/ItemImageProvider.cs
+++ b/MediaBrowser.Providers/Manager/ItemImageProvider.cs
@@ -49,6 +49,15 @@ namespace MediaBrowser.Providers.Manager
         };
 
         /// <summary>
+        /// Image types that may contain more than one item.
+        /// </summary>
+        private static readonly ImageType[] _multiImages =
+        {
+            ImageType.Backdrop,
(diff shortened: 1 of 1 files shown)


cvium:
Screenshots seem to be unused, so feel free to remove. I think may have to keep the `ImageType` though as it might be considered a breaking API change to remove it.

nielsvanvelzen:
> It might be considered a breaking API change to remove it.

Yes, to avoid breaking the clients we should keep properties/enum members and annotate them with `[Deprecated]`. The values can be set to something compatible with the type (empty list / null / empty string etc.).

PR #6794, opened 2021-11-07
Remove some dead code

Bond-009:


crobibero (approved):

MediaBrowser.Controller/Providers/DynamicImageResponse.cs
```diff
@@ -20,25 +20,5 @@ public class DynamicImageResponse
         public ImageFormat Format { get; set; }
 
         public bool HasImage { get; set; }
-
-        public void SetFormatFromMimeType(string mimeType)
```
cvium, line 24: Unused? https://github.com/jellyfin/jellyfin-plugin-bookshelf/blob/b7d14d5e23520f95fa65aa0ea4b8289e9be3982c/Jellyfin.Plugin.Bookshelf/Providers/Epub/EpubMetadataImageProvider.cs#L161

cvium (approved):

PR #6818, opened 2021-11-10
Replace PBKDF2-SHA1 with PBKDF2-SHA512

Bond-009:
This also migrates already created passwords on login

Source for the number of iterations:
https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#pbkdf2


MediaBrowser.Model/Cryptography/Constants.cs
```diff
@@ -8,11 +8,16 @@ public static class Constants
         /// <summary>
         /// The default length for new salts.
         /// </summary>
-        public const int DefaultSaltLength = 64;
+        public const int DefaultSaltLength = 128 / 8;
+
+        /// <summary>
+        /// The default output length.
+        /// </summary>
+        public const int DefaultOutputLength = 512 / 8;
```
Bond-009, line 16: Not sure about this, maybe it should be `256 / 8`?

1337joe (approved):

crobibero (approved):

PR #6819, opened 2021-11-10
Try to use Width and Height from ImageInfo to determine aspect ratio

cvium:
**Changes**
Primary aspect ratio is computed in a prioritized order:
1. From ImageInfo
2. From ImageProcessor
3. BaseItem.DefaultPrimaryImageAspectRatio

**Issues**
Idk

### DIFF
--- a/Emby.Server.Implementations/Dto/DtoService.cs
+++ b/Emby.Server.Implementations/Dto/DtoService.cs
@@ -1398,41 +1398,33 @@ namespace Emby.Server.Implementations.Dto
                 return null;
             }
 
-            ImageDimensions size;
-
-            var defaultAspectRatio = item.GetDefaultPrimaryImageAspectRatio();
-
-            if (defaultAspectRatio > 0)
-            {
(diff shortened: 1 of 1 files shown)


Emby.Server.Implementations/Dto/DtoService.cs
```diff
@@ -1398,41 +1398,33 @@ public void AttachPrimaryImageAspectRatio(IItemDto dto, BaseItem item)
...
-
-            if (defaultAspectRatio > 0)
-            {
-                return defaultAspectRatio;
-            }
-
             if (!imageInfo.IsLocalFile)
             {
-                return null;
+                return item.GetDefaultPrimaryImageAspectRatio();
```
cvium, line 1403: Maybe this should default to null like it used to

crobibero (approved):
approved pending testing

1337joe (approved):

crobibero:
We're doing it live

Issue #2355, opened 2020-01-30
Screen Grabber\Image Extractor Not Always Working


Artiume:
**Describe the bug**

Found logs for FFmpeg Extractor trying to run. Everything is disabled so it shouldn't be running. 

The System Exception errors are also interesting, ``` MediaStream videoStream, Nullable`1 imageStreamIndex, MediaProtocol protocol, Boolean isAudio, Nullable`1 threedFormat, Nullable`1 offset```

**Expected behavior**

Extractor shouldn't be attempting to run when it's not called for. 
**Logs**

```
[2020-01-30 15:56:06.345 -05:00] [INF] [14] MediaBrowser.MediaEncoding.Encoder.MediaEncoder:Killing ffmpeg process
[2020-01-30 15:56:06.345 -05:00] [ERR] [14] MediaBrowser.MediaEncoding.Encoder.MediaEncoder:ffmpeg image extraction failed for file:"/media/movies/F/Freaks (1932)/Freaks (1932) WEBDL-1080p.mkv"
[2020-01-30 15:56:06.346 -05:00] [ERR] [14] MediaBrowser.MediaEncoding.Encoder.MediaEncoder:I-frame image extraction failed, will attempt standard way. Input: "file:\"/media/movies/F/Freaks (1932)/Freaks (1932) WEBDL-1080p.mkv\""
System.Exception: ffmpeg image extraction failed for file:"/media/movies/F/Freaks (1932)/Freaks (1932) WEBDL-1080p.mkv"
   at MediaBrowser.MediaEncoding.Encoder.MediaEncoder.ExtractImageInternal(String inputPath, String container, MediaStream videoStream, Nullable`1 imageStreamIndex, Nullable`1 threedFormat, Nullable`1 offset, Boolean useIFrame, CancellationToken cancellationToken)
   at MediaBrowser.MediaEncoding.Encoder.MediaEncoder.ExtractImage(String[] inputFiles, String container, MediaStream videoStream, Nullable`1 imageStreamIndex, MediaProtocol protocol, Boolean isAudio, Nullable`1 threedFormat, Nullable`1 offset, CancellationToken cancellationToken)
```

Artiume:
https://www.reddit.com/r/jellyfin/comments/ewrwxg/jellyfin_writes_to_disk_like_crazy_for_days_while/

Artiume:
every image extractor is turned off. found this for a music file. 
[...]

PR #6820, opened 2021-11-10
Add/expose image settings for video extras

1337joe:
Old state: If no library `TypeOptions` are found for a type a default option is used that enables all image providers. This allows the usage of a dummy object for populating the options panel, but results in image providers that the user disabled for `Movie` being run on any extras, trailers, or secondary video files also in the directory with the primary movie file.

**Changes**
- Fall back to `Owner` on missing `Parent` when searching up the tree for the library root to allow options to be found for owned items.
- Fall back to `Video` when type `Trailer` not found in library options.
- Add check to ensure owned videos know their owner on metadata refresh (videos created from the `Video.LocalAlternateVersions` list didn't know their owner).
- Add `Video` library options to allow the user to modify image collection settings for video extras.

"Extras" would be a better name for the library options, but options are looked up by type name, and in this case the problematic files were of type types `Trailer` or `Video`.

**Issues**
Possibly related to #2355

### DIFF
--- a/MediaBrowser.Model/Configuration/LibraryOptions.cs
+++ b/MediaBrowser.Model/Configuration/LibraryOptions.cs
@@ -93,12 +93,6 @@ namespace MediaBrowser.Model.Configuration
                 }
             }
 
-            if (type == "Trailer")
-            {
-                // fall back to generic extras options (if set)
-                return GetTypeOptions("Video");
-            }
-
             return null;
         }
     }
--- a/MediaBrowser.Model/Configuration/LibraryOptions.cs
+++ b/MediaBrowser.Model/Configuration/LibraryOptions.cs
@@ -93,6 +93,12 @@ namespace MediaBrowser.Model.Configuration
                 }
             }
 
+            if (type == "Trailer")
+            {
+                // fall back to generic extras options (if set)
+                return GetTypeOptions("Video");
+            }
+
             return null;
         }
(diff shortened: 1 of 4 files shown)


crobibero (approved):

MediaBrowser.Model/Configuration/LibraryOptions.cs
```diff
@@ -93,6 +93,12 @@ public LibraryOptions()
                 }
             }
 
+            if (type == "Trailer")
+            {
```
cvium, line 97: Not sure how I feel about this, but you should use nameof instead of hardcoding the type name

1337joe: I agree this is more hackish than the other workarounds I added. I was going to try adding a field to `TypeOptions` to add a secondary type so the `Video` option could also match `Trailer` but that looked like it was going to require a change to `LibraryTypeOptionsDto`, which seemed like larger scope than I wanted to hit with this.

If the other changes in this pull request look good I'd be happy to yank this part out and say so in a comment in the related issue.

1337joe:
Looking back at this it works for what it's intended to do, but feels like it makes the user experience worse for minimal gain (how many people will understand why there are image provider options for "Video"?). I'll try to revisit this later since it never got traction in this form.

Issue #6821, opened 2021-11-11
Jellyfin LDAP Authentication


NickSica:
### Please describe your bug

I was trying to connect Ombi to Jellyfin for user authentication and while the user would authenticate correctly on the backend, a null value error caused it to return a 401. I found the issue and my changes are in my [fork](https://github.com/NickSica/jellyfin/commit/4b2c40f71706d0b8e77925cb0103ae7b9fc65503). I am a bit wary about opening a pull request due to some behavior that was exhibited from taking the lines I ultimately changed out altogether. The behavior I noticed immediately was a broken the connection to jellyfin itself (on the server setup page in Ombi), but the authentication was working with existing Jellyfin users (The second log is from that). If all looks well, I can submit a pull request. 

Steps to Reproduce
1. Sign into Ombi
2. Connect to Jellyfin and Import Users
3. Try to login with an LDAP user

### Jellyfin Version

Other

### if other:

10.8.0

### Environment

```markdown
- OS: ArchLinux
- Virtualization: None
- Clients: Browser
- Browser: Chrome
- FFmpeg Version: 4.4-Jellyfin
- Playback Method: Direct Play
- Hardware Acceleration: none
- Plugins: LDAP
- Reverse Proxy: Nginx
- Base URL: none
- Networking: Host
- Storage: local
```

### Jellyfin logs

```shell
[2021-11-10 22:57:48.626 -05:00] [INF] [25] Emby.Server.Implementations.Session.SessionManager: Reissuing access token: "REDACTED"
[2021-11-10 22:57:48.630 -05:00] [ERR] [25] Jellyfin.Server.Middleware.ExceptionMiddleware: Error processing request. URL "POST" "/users/authenticatebyname".
System.ArgumentNullException: Value cannot be null. (Parameter 'deviceId')
   at Emby.Server.Implementations.Session.SessionManager.LogSessionActivity(String appName, String appVersion, String deviceId, String deviceName, String remoteEndPoint, User user) in /home/REDACTED/.cache/yay/jellyfin-git/src/jellyfin/Emby.Server.Implementations/Session/SessionManager.cs:line 280
   at Emby.Server.Implementations.Session.SessionManager.AuthenticateNewSessionInternal(AuthenticationRequest request, Boolean enforcePassword) in /home/REDACTED/.cache/yay/jellyfin-git/src/jellyfin/Emby.Server.Implementations/Session/SessionManager.cs:line 1501
[...]

PR #6822, opened 2021-11-11
Fix Ombi auth through Jellyfin

NickSica:
**Changes**
Adds a check for null or whitespace so the correct DeviceId, Version, and Device don't get overwritten

**Issues**
#6821

### DIFF
--- a/Jellyfin.Server.Implementations/Security/AuthorizationContext.cs
+++ b/Jellyfin.Server.Implementations/Security/AuthorizationContext.cs
@@ -199,6 +199,7 @@ namespace Jellyfin.Server.Implementations.Security
                     {
                         authInfo.Version = string.Empty;
                     }
+
                     authInfo.IsApiKey = true;
                 }
             }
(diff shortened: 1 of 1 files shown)


Jellyfin.Server.Implementations/Security/AuthorizationContext.cs
```diff
@@ -185,9 +185,20 @@ private async Task<AuthorizationInfo> GetAuthorization(HttpContext httpReq)
...
+
+                    if(string.IsNullOrWhiteSpace(authInfo.Device))
+                    {
+                        authInfo.Device = string.Empty;
+                    }
+
+                    if(string.IsNullOrWhiteSpace(authInfo.Version))
+                    {
+                        authInfo.Version = string.Empty;
+                    }
```
crobibero, line 201:
```suggestion
                    }

```

Jellyfin.Server.Implementations/Security/AuthorizationContext.cs
```diff
@@ -185,9 +185,21 @@ private async Task<AuthorizationInfo> GetAuthorization(HttpContext httpReq)
...
+                    {
+                        authInfo.DeviceId = string.Empty;
+                    }
+
+                    if(string.IsNullOrWhiteSpace(authInfo.Device))
+                    {
+                        authInfo.Device = string.Empty;
+                    }
+
+                    if(string.IsNullOrWhiteSpace(authInfo.Version))
```
1337joe, line 198: The suggestion tool isn't working for me, but there should be a single space between `if` and `(`, like so: `if (string.IsNull...`

1337joe (approved):

crobibero (approved):
While this won't fully solve the issue you ran into, it's still a good change

PR #6831, opened 2021-11-12
Image provider cleanup

1337joe:
**Changes**
- Clean up tests, merge similar tests using theories.
- Switch `IMediaEncoder` to take an enum type instead of string for image format to be less error-prone.
- Merge duplicated codec check in normalizer.
- Check codec for format of embedded image.

### DIFF
--- a/MediaBrowser.Providers/MediaInfo/EmbeddedImageProvider.cs
+++ b/MediaBrowser.Providers/MediaInfo/EmbeddedImageProvider.cs
@@ -156,13 +156,14 @@ namespace MediaBrowser.Providers.MediaInfo
                 }
             }
 
+            var format = ImageFormat.Jpg;
             string extractedImagePath =
-                await _mediaEncoder.ExtractVideoImage(item.Path, item.Container, mediaSource, imageStream, imageStream.Index, ".jpg", cancellationToken)
+                await _mediaEncoder.ExtractVideoImage(item.Path, item.Container, mediaSource, imageStream, imageStream.Index, format, cancellationToken)
                     .ConfigureAwait(false);
 
             return new DynamicImageResponse
             {
-                Format = ImageFormat.Jpg,
+                Format = format,
                 HasImage = true,
                 Path = extractedImagePath,
                 Protocol = MediaProtocol.File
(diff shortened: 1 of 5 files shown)


MediaBrowser.Providers/MediaInfo/EmbeddedImageProvider.cs
```diff
@@ -114,7 +114,7 @@ private async Task<DynamicImageResponse> GetEmbeddedImage(Video item, ImageType
                 ImageType.Primary => _primaryImageFileNames,
                 ImageType.Backdrop => _backdropImageFileNames,
                 ImageType.Logo => _logoImageFileNames,
-                _ => _primaryImageFileNames
+                _ => throw new ArgumentException("Unexpected image type: " + type)
```
cvium, line 117: Why this change?

1337joe: If the default case is hit then the caller is ignoring the response from `GetSupportedImages`. I noticed it was a line uncovered by the unit tests, and on considering how to hit it I realized that it shouldn't be a valid case.

Defaulting to primary would end up duplicating the primary image in whatever other image slot is requested. I can't see a case where that would be an intended behavior.

crobibero (approved):

MediaBrowser.Providers/MediaInfo/EmbeddedImageProvider.cs
cvium: But instead of crashing shouldn't it return `new DynamicImageResponse { HasImage = false }`?

1337joe: Well when you put it like that...

I was figuring the only person who would hit that case would be a developer in the process of adding new image types so failing fast and obviously was beneficial, but I see a very strong case for "don't cause errors unless actually necessary". Is it worth injecting a logger into the class for just this?

cvium: 1. You should probably add a "fast return" ie. `if (imageFileNames.Length == 0)`.
2. As long as we're not new'ing the class, inject the logger where you think it makes sense.

cvium (approved):

Issue #6832, opened 2021-11-13
Health endpoint returns authentication error, spams logs


joshuaboniface:
### Please describe your bug

Querying the API `/health` endpoint triggers an authentication error in the logs:

```
Nov 12 21:27:13 jf1.i.net jellyfin[2401927]: [21:27:13] [DBG] [23] Jellyfin.Api.Auth.CustomAuthenticationHandler: Error authenticating with CustomAuthenticationHandler
Nov 12 21:27:13 jf1.i.net jellyfin[2401927]: MediaBrowser.Controller.Authentication.AuthenticationException: Request does not contain a token.
Nov 12 21:27:13 jf1.i.net jellyfin[2401927]:    at Emby.Server.Implementations.HttpServer.Security.AuthService.Authenticate(HttpRequest request)
Nov 12 21:27:13 jf1.i.net jellyfin[2401927]:    at Jellyfin.Api.Auth.CustomAuthenticationHandler.HandleAuthenticateAsync()
Nov 12 21:27:13 jf1.i.net jellyfin[2401927]: [21:27:13] [DBG] [23] Jellyfin.Api.Auth.CustomAuthenticationHandler: AuthenticationScheme: CustomAuthentication was not authenticated.
```

However the actual request does return, e.g.

```
joshua@base git:fix-auth-health ~/Projects/Jellyfin/projects/server/jellyfin $ curl jf1.i.net:8096/health
Healthy
```

Couple points raised:

1. Why would this be an error instead of a `return false` situation at https://github.com/jellyfin/jellyfin/blob/fb5385f1df55c7a2ee35981ac138cee6c62c903d/Emby.Server.Implementations/HttpServer/Security/AuthService.cs#L25-L28

2. Should the `/health` endpoint even be protected? It looks autogenerated so I don't see a clear way to exclude it.

### Jellyfin Version

10.7.7, 10.8.0 alpha 1

### Environment

```markdown
- OS: Debian
- Virtualization: KVM (N/A)
- Clients: Curl
- Browser: N/A
- FFmpeg Version: N/A
- Playback Method: N/A
- Hardware Acceleration: N/A
- Plugins: N/A
- Reverse Proxy: HAProxy
- Base URL: `/`
- Networking: Standard
- Storage: Standard
```

### Jellyfin logs

```shell
Above.
```

### FFmpeg logs

```shell
N/A
```

### Please attach any browser or client logs here

N/A

### Please attach any screenshots here

N/A

### Code of Conduct

[...]

PR #6837, opened 2021-11-13
Don't throw exception on unauthenticated requests

crobibero:
Fixes https://github.com/jellyfin/jellyfin/issues/6832


cvium (approved):

joshuaboniface:
One Q: Will this still allow this endpoint to return if there's no auth, or will it start enforcing auth here?

crobibero:
> One Q: Will this still allow this endpoint to return if there's no auth, or will it start enforcing auth here?

Endpoints that don’t require auth work as intended with this change

joshuaboniface (approved):

PR #6848, opened 2021-11-15
Ensure the proper StartTimeTicks variable is forwarded to the AddDlnaHeaders function

marius-luca-87:
StreamingHelpers.GetStreamingState populates state.Request.StartTimeTicks based on the startTimeTicks query variable and DLNA headers.

Currently the query variable was used directly losing anything provided by "TimeSeekRange.dlna.org" .

### DIFF
--- a/Jellyfin.Api/Controllers/VideosController.cs
+++ b/Jellyfin.Api/Controllers/VideosController.cs
@@ -451,7 +451,7 @@ namespace Jellyfin.Api.Controllers
 
             if (@static.HasValue && @static.Value && state.DirectStreamProvider != null)
             {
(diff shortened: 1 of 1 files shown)


crobibero (changes requested):
These endpoints should prioritize the header if provided, but still fall back to using the query parameter.

marius-luca-87:
Query parameter is prioritized at:
https://github.com/jellyfin/jellyfin/blob/03435641c847acaf88c605ba55f5688d170a8251/Jellyfin.Api/Helpers/StreamingHelpers.cs#L67

This PR restores the original intention of the code that got lost in an old refactoring.

crobibero (approved):

cvium (approved):

PR #6850, opened 2021-11-15
Add Collection Validator, create collection based on nfo

PetrusZ:
Based on nfo's set element, automatically add movie to collection.

Signed-off-by: Petrus.Z <silencly07@gmail.com>

**Changes**
Implement the feature request posted in https://features.jellyfin.org/posts/73/create-collection-based-on-nfo.

I am new to C#, there may be some errors in the code. Thanks in advance for review.


Emby.Server.Implementations/Library/Validators/CollectionPostScanTask.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        /// <summary>
+        /// Runs the specified progress.
+        /// </summary>
+        /// <param name="progress">The progress.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <returns>Task.</returns>
+        public Task Run(IProgress<double> progress, CancellationToken cancellationToken)
+        {
+            return new CollectionValidator(_libraryManager, _collectionManager, _logger).Run(progress, cancellationToken);
```
cvium, line 47: I know you're probably just following the structure of the existing tasks, but it doesn't make much sense to put all the logic in a separate class.

Emby.Server.Implementations/Library/Validators/CollectionValidator.cs
```diff
@@ -0,0 +1,166 @@
...
+            progress.Report(100);
+        }
+
+        private async void AddMovieToCollection(Guid boxSetId, BoxSet boxSet, List<Movie> movieList)
+        {
+
+            var movieIds = new List<Guid>();
+            foreach (var movie in movieList)
+            {
+                if (!boxSet.ContainsLinkedChildByItemId(movie.Id))
```
cvium, line 156: This check is already performed by `AddToCollectionAsync`

Emby.Server.Implementations/Library/Validators/CollectionValidator.cs
```diff
@@ -0,0 +1,166 @@
...
+            var movies = _libraryManager.GetItemList(new InternalItemsQuery
+            {
+                IncludeItemTypes = new[] { nameof(Movie) },
+                IsVirtualItem = false,
+                OrderBy = new List<ValueTuple<string, SortOrder>>
+                {
+                new ValueTuple<string, SortOrder>(ItemSortBy.SortName, SortOrder.Ascending)
+                },
+                Recursive = true
+            }).Select(m => m as Movie).ToList();
```
cvium, line 66: Doing the cast and then calling ToList allocates a new list of the same objects. You already know they are `Movie` types, so just do the cast in the for-loop.

Emby.Server.Implementations/Library/Validators/CollectionValidator.cs
```diff
@@ -0,0 +1,166 @@
...
+
+            var boxSets = _libraryManager.GetItemList(new InternalItemsQuery
+            {
+                IncludeItemTypes = new[] { nameof(BoxSet) },
+                CollapseBoxSetItems = false,
+                Recursive = true
+            }).Select(b => b as BoxSet).ToList();
```
cvium, line 73: Same as above

Emby.Server.Implementations/Library/Validators/CollectionValidator.cs
```diff
@@ -0,0 +1,166 @@
...
+
+            var map = new Dictionary<string, List<Movie>>();
+            foreach (var movie in movies)
+            {
+                if (movie != null && movie.CollectionName != null)
+                {
+                    var movieList = new List<Movie>();
+                    if (map.TryGetValue(movie.CollectionName, out movieList))
+                    {
+                        if (!movieList.Where(m => m.Id == movie.Id).Any())
```
cvium, line 86:
```suggestion
                        if (!movieList.Any(m => m.Id == movie.Id))
```

Emby.Server.Implementations/Library/Validators/CollectionValidator.cs
```diff
@@ -0,0 +1,166 @@
...
+
+        /// <summary>
+        /// Runs the specified progress.
+        /// </summary>
+        /// <param name="progress">The progress.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <returns>Task.</returns>
+        public async Task Run(IProgress<double> progress, CancellationToken cancellationToken)
+        {
+            var movies = _libraryManager.GetItemList(new InternalItemsQuery
```
crobibero, line 57: I would probably paginate this query so it doesn’t pull all of the movies at once. It doesn’t need to be a small page query, 1,000 is probably good enough

PetrusZ:
Thanks @cvium , I fixed all issues your mentioned above.

Emby.Server.Implementations/Library/Validators/CollectionValidator.cs
PetrusZ: May I ask how to paginate this? I need to get all items which type is movie, the InternalItemsQuery don't have skip option. And if I paginate based itemIds, I have to get all movie's itemdId, this need to query all movies at once too. Thanks.

PetrusZ:
Oops, seems put all logic to PostScanTask somehow changed the OpenAPI specification?

cvium:
> Oops, seems put all logic to PostScanTask somehow changed the OpenAPI specification?

You can safely ignore it.

Emby.Server.Implementations/Library/Validators/CollectionPostScanTask.cs
```diff
@@ -0,0 +1,155 @@
...
+
+                numComplete++;
+                double percent = numComplete;
+                percent /= count * 2;
+                percent *= 100;
+
+                progress.Report(percent);
+            }
+
+            foreach (var pair in collectionNameMoviesMap)
```
cvium, line 112: Use deconstruction
```suggestion
            foreach (var (collectionName, movieList) in collectionNameMoviesMap)
```

Emby.Server.Implementations/Library/Validators/CollectionPostScanTask.cs
```diff
@@ -0,0 +1,155 @@
...
+            {
+                try
+                {
+                    var collectionName = pair.Key;
+                    var movieList = pair.Value;
+
+                    var boxSet = boxSets.FirstOrDefault(b => b != null ? b.Name == collectionName : false) as BoxSet;
```
cvium, line 119:
```suggestion
                    var boxSet = boxSets.FirstOrDefault(b => b?.Name == collectionName) as BoxSet;
```

Emby.Server.Implementations/Library/Validators/CollectionPostScanTask.cs
```diff
@@ -0,0 +1,155 @@
...
+                    if (boxSet == null)
+                    {
+                        // won't automatically create collection if only one movie in it
+                        if (movieList.Count >= 2)
+                        {
+                            boxSet = await _collectionManager.CreateCollectionAsync(new CollectionCreationOptions
+                            {
+                                Name = collectionName,
+                                IsLocked = true
```
cvium, line 128: Why locked?

Emby.Server.Implementations/Library/Validators/CollectionPostScanTask.cs
```diff
@@ -0,0 +1,155 @@
...
+            {
+                IncludeItemTypes = new[] { nameof(BoxSet) },
+                CollapseBoxSetItems = false,
+                Recursive = true
+            });
+
+            var numComplete = 0;
+            var count = movies.Count;
+
+            var collectionNameMoviesMap = new Dictionary<string, List<Movie>>();
```
cvium, line 82: You don't seem to be using the `Movie` class for much, so it should probably just be a `List<Guid>`. It will save you a `.Select` later.

Emby.Server.Implementations/Library/Validators/CollectionPostScanTask.cs
```diff
@@ -0,0 +1,155 @@
...
+                    numComplete++;
+                    double percent = numComplete;
+                    percent /= count * 2;
+                    percent *= 100;
+
+                    progress.Report(percent);
+                }
+                catch (Exception ex)
+                {
+                    _logger.LogError(ex, "Error refreshing {0}, {1}", pair.Key, pair.Value.ToString());
```
cvium, line 148:
```suggestion
                    _logger.LogError(ex, "Error refreshing {CollectionName} with {@MovieIds}", collectionName, movieList);
```

Emby.Server.Implementations/Library/Validators/CollectionPostScanTask.cs
```diff
@@ -0,0 +1,155 @@
...
+namespace Emby.Server.Implementations.Library.Validators
+{
+    /// <summary>
+    /// Class CollectionPostScanTask.
+    /// </summary>
+    public class CollectionPostScanTask : ILibraryPostScanTask
+    {
+        /// <summary>
+        /// The _library manager.
+        /// </summary>
```
cvium, line 23: Not necessary for private fields and properties.
```suggestion
```

Emby.Server.Implementations/Library/Validators/CollectionPostScanTask.cs
```diff
@@ -0,0 +1,155 @@
...
+        private readonly ILibraryManager _libraryManager;
+
+        /// <summary>
+        /// The collection manager.
+        /// </summary>
```
cvium, line 28:
```suggestion
```

Emby.Server.Implementations/Library/Validators/CollectionPostScanTask.cs
```diff
@@ -0,0 +1,155 @@
...
+        private readonly ICollectionManager _collectionManager;
+
+        /// <summary>
+        /// The logger.
+        /// </summary>
```
cvium, line 33:
```suggestion
```

Emby.Server.Implementations/Library/Validators/CollectionPostScanTask.cs
```diff
@@ -0,0 +1,155 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CollectionPostScanTask" /> class.
+        /// </summary>
+        /// <param name="libraryManager">The library manager.</param>
+        /// <param name="collectionManager">The collection manager.</param>
+        /// <param name="logger">The logger.</param>
+        public CollectionPostScanTask(
+            ILibraryManager libraryManager,
+            ILogger<CollectionPostScanTask> logger,
+            ICollectionManager collectionManager)
```
cvium, line 45: Minor thing
```suggestion
            ICollectionManager collectionManager,
            ILogger<CollectionPostScanTask> logger)
```

Emby.Server.Implementations/Library/Validators/CollectionPostScanTask.cs
```diff
@@ -0,0 +1,155 @@
...
+
+        /// <summary>
+        /// Runs the specified progress.
+        /// </summary>
+        /// <param name="progress">The progress.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <returns>Task.</returns>
+        public async Task Run(IProgress<double> progress, CancellationToken cancellationToken)
+        {
+
```
cvium, line 60:
```suggestion
```

cvium (changes requested):
It's generally fine. My comments are mostly style and performance related.

Another thing is that I think we should make this configurable at the Library level. Not everyone wants automatic collections.

Emby.Server.Implementations/Library/Validators/CollectionPostScanTask.cs
PetrusZ: `IsLocked = true` is also defualt when create collection manually. If you create a collection with `IsLocked = false` and the collection don't have correspond online data, then the collection will don't have cover/image. On the other hand, when set `IsLocked = true`, jellyfin will auto generate a cover/image based on local file/video. I think it's more friendly and beautiful.

cvium: Alright

Emby.Server.Implementations/Library/Validators/CollectionPostScanTask.cs
PetrusZ: To implement configureable automiclly create collection, I think `move.ContainingFolderPath` will be needed next. So I'll fix this later.

cvium: You should be skipping movies that are contained in a library with the option disabled, so `List<Guid>` still makes sense

PetrusZ:
Thank you @cvium , I fixed almost all issues you mentioned above. Next, I will look at how to implement configurable automatic creation of collections.

Emby.Server.Implementations/Library/Validators/CollectionPostScanTask.cs
PetrusZ: I'm thinking add a minimal variable to library options, when movie amount in same collection is less than this minimal variable, then it won't create or add it to collection.

So in the next foreach map loop, I will need `movie.ContainingFolderPath` to get library option and check whether should add it. It can't be done in first loop, beacause at that time, I don't know how many movies are in same collection.

cvium: You shouldn't be using `ContainingFolderPath`.

How will you handle two libraries with different collection requirements?

PetrusZ: Sorry, I'm not follow. What do you mean two libraries with different collection? I think there no other way to find out the movie belong to which library, except using `ContainingFolderPath`.

Emby.Server.Implementations/Library/Validators/CollectionValidator.cs
crobibero: There's a `StartIndex` parameter, which is used as an offset when querying.

Emby.Server.Implementations/Library/Validators/CollectionPostScanTask.cs
cvium: You can have as many Movie libraries as you want, but if the configuration is `MinimumNumberOfMoviesPerCollection`, how do you want to handle if library X is 2 and library Y is 5?

Maybe we'd save ourselves a headache by just forcing it to 2 and then make it a toggle instead (enabled/disabled).

>  I think there no other way to find out the movie belong to which library

We don't have to. LibraryManager.GetLibraryOptions is enough to get the setting.

PetrusZ: > You can have as many Movie libraries as you want, but if the configuration is MinimumNumberOfMoviesPerCollection, how do you want to handle if library X is 2 and library Y is 5?

Do you mean that one movie in two libraries, and library X is 2 and library Y is 5? That's a problem, but I think we can choose the smallest one. But you are right, the easiesy way is forcing it to 2. We can implement the simple one first.

> We don't have to. LibraryManager.GetLibraryOptions is enough to get the setting.

Thank you, I totally ignore `LibraryManager.GetLibraryOptions`, beacause I saw variable such as `CollectionFolders` in it function, I thought it was only can be used throught collection...

PetrusZ:
Hi, I add a option to library, so it can determine whether create or add movies to collection. I also submit a webui changed PR to jellyfin-web https://github.com/jellyfin/jellyfin-web/pull/3181 .

Emby.Server.Implementations/Library/Validators/CollectionValidator.cs
PetrusZ: Hi @crobibero , because I need to count how many movies in same collection, I have to store all movies to a map. It makes paginate movie query meanless. And I think the cost is more or less aceeptable, I have nearly 3000 movies, during first time validate , the memory rise about 100MB.

Emby.Server.Implementations/Library/Validators/CollectionPostScanTask.cs
```diff
@@ -0,0 +1,152 @@
...
+            }
+
+            progress.Report(100);
+        }
+
+        private List<Guid> FliterMoviesByOption(List<Movie> movieList) {
+            List<Guid> movieIds = new List<Guid>();
+            foreach (var movie in movieList)
+            {
+                if (_libraryManager.GetLibraryOptions(movie).AutoCollection)
```
cvium, line 144: I worry that this is inefficient. It is more likely that a user has a handful of movie libraries _at most_, but they could contain thousands of movies. That is a lot of unnecessary calls.

Maybe the whole thing should be flipped. If you look at this task https://github.com/jellyfin/jellyfin/blob/f5ca9cbc3bcffc13830669d5790525c6c11f9136/MediaBrowser.Providers/MediaInfo/SubtitleScheduledTask.cs#L73, you can see how it fetches items for specific libraries. We could have it do

1. Grab all libraries
2. Find libraries for which `AutoCollection` is true (note: I think the name should be changed to something like `AutomaticallyCreateCollections`)
3. Find all movies in the libraries (with AutoCollection=true) https://github.com/jellyfin/jellyfin/blob/f5ca9cbc3bcffc13830669d5790525c6c11f9136/MediaBrowser.Providers/MediaInfo/SubtitleScheduledTask.cs#L99-L108
4. Create the mapping like you do now
5. Create collections

PetrusZ: Great! I want to do this before, but I dont' know how to query movies in specfic library. Thank you.

As for library option, I choose `AutoCollection` beacause it not just create. If a collection already exits, it also will automatic add movie to collection.

Emby.Server.Implementations/Library/Validators/CollectionValidator.cs
PetrusZ: Improved a little, now query movies one time per library.

Emby.Server.Implementations/Library/Validators/CollectionPostScanTask.cs
crobibero: If this is complete, the `Movie` object isn't being used to build the collections, so you can just store the ID, as cvium suggested previously

PetrusZ: Sorry, forgot that. Fixed this in lastest commit.

Emby.Server.Implementations/Library/Validators/CollectionPostScanTask.cs
```diff
@@ -0,0 +1,137 @@
...
+                    Recursive = true
+                });
+
+                foreach (var m in movies)
+                {
+                    if (m is Movie movie && !string.IsNullOrEmpty(movie.CollectionName))
+                    {
+                        if (collectionNameMoviesMap.TryGetValue(movie.CollectionName, out var movieList))
+                        {
+                            if (!movieList.Any(m => m == movie.Id))
```
cvium, line 81: This should be equivalent I think
```suggestion
                            if (!movieList.Contains(movie.Id))
```

Emby.Server.Implementations/Library/Validators/CollectionPostScanTask.cs
```diff
@@ -0,0 +1,137 @@
...
+        public async Task Run(IProgress<double> progress, CancellationToken cancellationToken)
+        {
+            var boxSets = _libraryManager.GetItemList(new InternalItemsQuery
+            {
+                IncludeItemTypes = new[] { nameof(BoxSet) },
+                CollapseBoxSetItems = false,
+                Recursive = true
+            });
+
+            var collectionNameMoviesMap = new Dictionary<string, List<Guid>>();
```
cvium, line 57: I wonder if a HashSet is better here.
```suggestion
            var collectionNameMoviesMap = new Dictionary<string, HashSet<Guid>>();
```

MediaBrowser.Model/Configuration/LibraryOptions.cs
```diff
@@ -80,6 +81,7 @@ public LibraryOptions()
         public bool RequirePerfectSubtitleMatch { get; set; }
 
         public bool SaveSubtitlesWithMedia { get; set; }
+        public bool AutoCollection { get; set; }
```
cvium, line 84: I still believe a more descriptive name should be used. You have to keep in mind that this will be written to an XML file and it might not be clear what AutoCollection means.

PetrusZ: What do you think of `AutoAddToCollection`? Is it better?

Emby.Server.Implementations/Library/Validators/CollectionPostScanTask.cs
```diff
@@ -0,0 +1,137 @@
...
+                    Recursive = true
+                });
+
+                foreach (var m in movies)
+                {
+                    if (m is Movie movie && !string.IsNullOrEmpty(movie.CollectionName))
+                    {
+                        if (collectionNameMoviesMap.TryGetValue(movie.CollectionName, out var movieList))
+                        {
+                            if (!movieList.Contains(movie.Id))
```
crobibero, line 81: Don’t need to check if the element exists in a hash set before adding it, the add function already does the check

https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.hashset-1.add?view=net-6.0

crobibero (commented):
Now that we’re not keeping all of the movies in memory I think we should add pagination

Emby.Server.Implementations/Library/Validators/CollectionPostScanTask.cs
PetrusZ: Fixed. Thanks.

PetrusZ:
Hi @crobibero , I added pagination, don’t know if I did it correctly.

PetrusZ:
Sorry, made a silly mistake before. @crobibero Can you check if I have pagination correctly? Thank you.

Emby.Server.Implementations/Library/Validators/CollectionPostScanTask.cs
```diff
@@ -0,0 +1,150 @@
...
+                        break;
+                    }
+
+                    startIndex += pagesize;
+                }
+            }
+
+            var numComplete = 0;
+            var count = collectionNameMoviesMap.Count;
+
```
cvium, line 109:
```suggestion
            if (count == 0)
            {
                progress.Report(100);
                return;
            }

            var boxSets = _libraryManager.GetItemList(new InternalItemsQuery
            {
                IncludeItemTypes = new[] { nameof(BoxSet) },
                CollapseBoxSetItems = false,
                Recursive = true
            });
```

Emby.Server.Implementations/Library/Validators/CollectionPostScanTask.cs
```diff
@@ -0,0 +1,150 @@
...
+        public async Task Run(IProgress<double> progress, CancellationToken cancellationToken)
+        {
+            var boxSets = _libraryManager.GetItemList(new InternalItemsQuery
+            {
+                IncludeItemTypes = new[] { nameof(BoxSet) },
+                CollapseBoxSetItems = false,
+                Recursive = true
+            });
+
+            var collectionNameMoviesMap = new Dictionary<string, HashSet<Guid>>();
```
cvium, line 57: Move the query down below to save a query if there's no items
```suggestion
            var collectionNameMoviesMap = new Dictionary<string, HashSet<Guid>>();
```

MediaBrowser.Model/Configuration/LibraryOptions.cs
cvium: `AutomaticallyAddToCollection` seems fine to me

PetrusZ:
Replaced `AutoCollection` to `AutomaticallyAddToCollection`.

Emby.Server.Implementations/Library/Validators/CollectionPostScanTask.cs
PetrusZ: Fixed, thanks.

cvium (approved):

crobibero (approved):

cvium:
I wouldn't call them mistakes. We're just often very particular about the code we merge.

tamburra:
Out of curiosity, will this write 'set' info to nfo file when collection is assigned via the WebUI?
I have jellyfin configured to write nfo files along side my media.  It would make sense to have this write to the nfo file as well to complete the experience of using jellyfin to organize my medias nfo files.

PetrusZ:
> Out of curiosity, will this write 'set' info to nfo file when collection is assigned to media via the WebUI? I have jellyfin configured to write nfo files along side my media. It would make sense to have this write to the nfo file as well to complete the experience of using jellyfin to organize my media nfo files.

No, it won't. And this feature isn't really related to this PR. I think you can start with making a feature request. And I'll check out this feature when I have time. BTW, it not saying I'll implement it for certain.

tamburra:
Thank you for the reply.  I have added a feature request, https://features.jellyfin.org/posts/1406/write-set-collection-to-nfo-file-when-media-is-add-to-collection-for-webui

PR #6853, opened 2021-11-15
Add ~alpha3 to Release

brianjmurrell:
**Changes**
Add ~alpha2 to Version: to be more transparent about what this build really is.

Signed-off-by: Brian J. Murrell <brian@interlinx.bc.ca>

### DIFF
--- a/fedora/jellyfin.spec
+++ b/fedora/jellyfin.spec
@@ -7,7 +7,7 @@
 %endif
 
 Name:           jellyfin
-Version:        10.8.0
+Version:        10.8.0~alpha2
 Release:        1%{?dist}
 Summary:        The Free Software Media System
 License:        GPLv3
@@ -137,6 +137,8 @@ fi
 %systemd_postun_with_restart jellyfin.service
 
 %changelog
+* Mon Nov 15 2021 Brian J. Murrell <brian@interlinx.bc.ca>
+- Update to release to be more transparent about which 10.8.0 release this is
 * Fri Dec 04 2020 Jellyfin Packaging Team <packaging@jellyfin.org>
 - Forthcoming stable release
 * Mon Jul 27 2020 Jellyfin Packaging Team <packaging@jellyfin.org>
--- a/fedora/jellyfin.spec
+++ b/fedora/jellyfin.spec
@@ -7,7 +7,7 @@
 %endif
 
 Name:           jellyfin
-Version:        10.8.0~alpha2
+Version:        10.8.0~alpha3
 Release:        1%{?dist}
 Summary:        The Free Software Media System
 License:        GPLv3


joshuaboniface:
Setting this to alpha2 won't do much since that one is already "out", but still thinking of a sustainable way to handle these (and the Debian packages similarly) during the pre-release cycle. It will probably have to be automated based on the CI, unless there's a way to change the binary package afterwards somehow.

brianjmurrell:
> Setting this to alpha2 won't do much since that one is already "out",

Fair enough.  I will update.

> but still thinking of a sustainable way to handle these (and the Debian packages similarly) during the pre-release cycle.

I don't know what your "update the release info" process is so I cannot help much.  When doing this sort of thing from git, I generally take advantage of the "most recent" git tag.  Usually coupled with the "number of commits since tag" to indicate whether the release really is on the tag or is in between.

> It will probably have to be automated based on the CI, unless there's a way to change the binary package afterwards somehow.

No way to change binary packages.

joshuaboniface:
Let's avoid this at all, updating this for every alpha/beta/rc is going to be a pain in the ass and another thing to have to track.

The goal post-10.8.0 is to have CI deal with this automatically, but until then, just leaving it at 10.8.0 for these prereleases is preferred and is what we're doing with Debian packages too.

joshuaboniface:
I'm going to see if I can script this in the CI and autogen these entries automatically based on the tag after the release is tagged.

joshuaboniface:
Also just curious, should this be in `Release:` instead of `Version:`? I'm referencing this with it in `Version:` for my auto-bump.

brianjmurrell:
> Also just curious, should this be in `Release:` instead of `Version:`? I'm referencing this with it in `Version:` for my auto-bump.

It has to be in `Release:`.

I see this still is not happening in alpha3.  Unless these RPMs have proper `~` in their `Release:` they are not going to properly upgrade from each other.  In fact as it's being [not] done now, there will NEVER be an upgrade from the first alpha as **nothing** is changing from one release to the next.  There is nothing to tell the package manager that something is actually a newer release.

I can certainly empathize with your wanting to automate this, but seeing as automation is something that will take time to come to fruition, the manual (and yes, painful, I guess) process should be getting done in the meanwhile otherwise your users will not be getting new/updated packages and you will not be getting up-to-date testing and will be getting out-of-date bug reports, etc.

brianjmurrell:
> based on the tag after the release is tagged.

Maybe just poor choice of words but **after** is a bit alarming.  The change to the `jellyfin.spec` to add the `~alpha3` has to be done in the **exact** commit that the tag is pointing to.  It doesn't work to modify the `jellyfin.spec` in a commit after the tag as it's the tagged commit that gets built.

joshuaboniface:
No, that is indeed correct. The CI will change the version when it is building the packages, after the tag.

So if the tag has something like `Version: 10.8.0` in the `jellyfin.spec`, the CI would use `bump_version 10.8.0-alpha4` to alter that to `Version: 10.8.0~alpha4` during the build process. The binary and source packages would then have that (correct) version, without any code changes in the actual repository itself.

Note that this is only going to affect our CI-based builds; I'm not sure how this would work (if at all) with COPR.

joshuaboniface:
Also note that I'm doing it like this because committing intermediate versions like this will break unstable builds; `master` doesn't know or care about these pre-releases, because if it did suddenly the unstable builds would be `alpha3` or such.

brianjmurrell:
> No, that is indeed correct. The CI will change the version when it is building the packages, after the tag.

Altering the specfile only at build time is a broken process.  The specfile, in the SCM should reflect the actual proper version.  People who pull the source shouldn't have to modify the specfile to get an accurately versioned build.

> So if the tag has something like `Version: 10.8.0` in the `jellyfin.spec`, the CI would use `bump_version 10.8.0-alpha4` to alter that to `Version: 10.8.0~alpha4`

Yeah, again.  This belongs in the specfile at the source, not done at build time.

> Note that this is only going to affect our CI-based builds; I'm not sure how this would work (if at all) with COPR.

Right.  You are solving the problem for an isolated one-off build system and breaking every other standard (both RPM and debian) build system in the world with this type of solution.  You are asking standard build systems (i.e. build systems which find and build a `.spec` file, or build from `debian/`, unaltered) to all change the way they work to satisfy this one-off build system.

If you really want to have a fully automated updating of versions across the source, my suggestion would be that you automate that in some pre-commit fashion so that the files that need accurate versioning are committed to SCM with the accurate version.  Maybe a script that does much of what your `bump_version` script does but does it pre-commit not at build time.  You could even probably hook it into a git hook to run automatically when you commit a tag, for example.

Additionally, people looking at unpacked source (not everyone might be viewing the source in the context of an SCM such as git, with tags, etc. -- i.e. tarballs) should have a very clear indication of what the actual version is without having to decipher what a build-time script (i.e. `bump_version`) is doing to make the purported version more accurate.

PR #6862, opened 2021-11-17
Query media streams by type instead of filtering

1337joe:
**Changes**
- Query `MediaSourceManager` directly in image providers 
- Address ignored warnings/minor cleanup in `AudioImageProvider`

### DIFF
--- a/MediaBrowser.Controller/Entities/BaseItem.cs
+++ b/MediaBrowser.Controller/Entities/BaseItem.cs
@@ -1108,15 +1108,9 @@ namespace MediaBrowser.Controller.Entities
 
         public virtual List<MediaStream> GetMediaStreams()
         {
-            return GetMediaStreams(null);
-        }
-
-        public virtual List<MediaStream> GetMediaStreams(MediaStreamType? type)
-        {
             return MediaSourceManager.GetMediaStreams(new MediaStreamQuery
             {
-                ItemId = Id,
-                Type = type
+                ItemId = Id
             });
         }
 
--- a/MediaBrowser.Controller/Entities/BaseItem.cs
+++ b/MediaBrowser.Controller/Entities/BaseItem.cs
@@ -1108,9 +1108,15 @@ namespace MediaBrowser.Controller.Entities
 
         public virtual List<MediaStream> GetMediaStreams()
         {
+            return GetMediaStreams(null);
+        }
+
+        public virtual List<MediaStream> GetMediaStreams(MediaStreamType? type)
+        {
             return MediaSourceManager.GetMediaStreams(new MediaStreamQuery
             {
-                ItemId = Id
+                ItemId = Id,
+                Type = type
             });
         }
 
--- a/MediaBrowser.Providers/MediaInfo/VideoImageProvider.cs
+++ b/MediaBrowser.Providers/MediaInfo/VideoImageProvider.cs
@@ -4,7 +4,9 @@ using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using MediaBrowser.Controller.Entities;
+using MediaBrowser.Controller.Library;
 using MediaBrowser.Controller.MediaEncoding;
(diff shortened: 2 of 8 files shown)


MediaBrowser.Controller/Entities/BaseItem.cs
```diff
@@ -1107,10 +1107,16 @@ public PlayAccess GetPlayAccess(User user)
         }
 
         public virtual List<MediaStream> GetMediaStreams()
```
1337joe, line 1109: Instead of adding a new method I considered changing this to have an optional, null-default `MediaStreamType?` parameter. That requires changing the `IHasMediaSources` signature to include the optional parameter. Is that allowed/desired?

There are places where `GetMediaStreams` needs to return all streams but also several places that immediately filter the list of streams to a specific type so the method will be called with and without the type argument.

cvium: It would be better to simply use the MediaSourceManager directly instead of via BaseItem

1337joe: I was initially reluctant to do this when I remembered that `LiveTvChannel` overrides `GetMediaStreams` to return an empty list (and I had missed the corresponding override for my changes), but none of the places that I switched to the new call can be reached with that type anyway. Converting this to a draft until I finish switching over.

crobibero (approved):

MediaBrowser.Providers/MediaInfo/VideoImageProvider.cs
```diff
@@ -78,12 +83,15 @@ private async Task<DynamicImageResponse> GetVideoImage(Video item, CancellationT
 
             // If we know the duration, grab it from 10% into the video. Otherwise just 10 seconds in.
             // Always use 10 seconds for dvd because our duration could be out of whack
-            var imageOffset = item.VideoType != VideoType.Dvd && item.RunTimeTicks.HasValue &&
-                              item.RunTimeTicks.Value > 0
+            var imageOffset = item.VideoType != VideoType.Dvd && item.RunTimeTicks is > 0
```
cvium, line 86: Redundant
```suggestion
            var imageOffset = item.VideoType != VideoType.Dvd && item.RunTimeTicks > 0
```

MediaBrowser.Providers/MediaInfo/VideoImageProvider.cs
```diff
@@ -78,12 +83,15 @@ private async Task<DynamicImageResponse> GetVideoImage(Video item, CancellationT
...
                                   ? TimeSpan.FromTicks(item.RunTimeTicks.Value / 10)
                                   : TimeSpan.FromSeconds(10);
 
-            var videoStream = item.GetDefaultVideoStream() ?? item.GetMediaStreams().FirstOrDefault(i => i.Type == MediaStreamType.Video);
+            var defaultQuery = new MediaStreamQuery { ItemId = item.Id, Index = item.DefaultVideoStreamIndex };
+            var videoQuery = new MediaStreamQuery { ItemId = item.Id, Type = MediaStreamType.Video };
+
+            var videoStream = _mediaSourceManager.GetMediaStreams(defaultQuery).FirstOrDefault()
+                ?? _mediaSourceManager.GetMediaStreams(videoQuery).FirstOrDefault();
```
cvium, line 94: It's a wasted allocation if the first call doesn't return null
```suggestion
            var query = new MediaStreamQuery { ItemId = item.Id, Index = item.DefaultVideoStreamIndex };
            var videoStream = _mediaSourceManager.GetMediaStreams(defaultQuery).FirstOrDefault();
            if (videoStream == null)
            {
                videoQuery.Type = MediaStreamType.Video;
                videoQuery.Index = null;
                videoStream = _mediaSourceManager.GetMediaStreams(query).FirstOrDefault();
            }
```

cvium (approved):

Issue #3669, opened 2020-07-23
Jellyfin treats 1080 as the episode number if the title includes 1080p


Starmann:
**Describe the bug**

**System (please complete the following information):**
 - OS: Windows 10 64bit Education
 - Virtualization: None
 - Clients: Browser (though it doesn't matter)
 - Browser: Firefox 78
 - Jellyfin Version: 10.6.0
 - Playback: Direct Stream
 - Installed Plugins: Anime
 - Reverse Proxy: None
 - Base URL: /web/index.html#!/details
 - Networking: localhost
 - Storage: HDD

**To Reproduce**

1. Create a new video, or download one. Give it an episode number followed by 1080p. My title was "[ZetaRebel] Mobile Suit Gundam - 01 - Gundam Rising!! (BD x264 1080p 8-bit FLAC)"
2. Open Jellyfin
3. Look at the episode number
4. It will be 1080

**Expected behavior**
I expect the episode number to be 1.

**Logs**
No error message, just some incorrect parsing.

**Screenshots**

I think my explanation was clear enough but here's a screenshot just in case it'll help

![](https://cdn.discordapp.com/attachments/735650534207258715/735668485698617384/unknown.png)

**Additional context**

I feel like the code for parsing titles is the weakest part of Jellyfin. As a stopgap solution, I'd love a configuration option for defining a show's episode naming format. Adding a show is one of the first things people do, and thus should be a more polished experience. While parsing titles is a hard problem to solve, I feel like most title problems could be ignored by the default format discarding anything in brackets or parenthesis as those are often used for metadata.

Bond-009:
Can you give the full path of the file?

whalehub:
@Starmann All the major media server applications (Plex, Emby, Jellyfin) expect you to name your files properly. In this case, the file should be called:

`Mobile Suit Gundam (1979) - S01E01 - Gundam Rising!! (BD x264 1080p 8-bit FLAC)`

Starmann:
@Bond-009 `E:\Movies\[ZetaRebel] Mobile Suit Gundam (BD x264 1080p 8-bit FLAC)\[ZetaRebel] Mobile Suit Gundam - 01 - Gundam Rising!! (BD x264 1080p 8-bit FLAC).mkv`

Starmann:
@whalehub I wasn't aware of that. I feel like there should be a tool to help you rename them properly.

Bond-009:
[...]

PR #6864, opened 2021-11-18
Improve episode parser

fiochen:
**Changes**

1. Avoid parsing wrong number from media tag, like:
`[foo] seriesname 03 [HEVC-10bit].mkv` want "3", not "10"
`[foo] seriesname 03 [Ma10p-720p].mkv` want "3", not "720"

2. Improve parsing in brackets, like:
`[foo] seriesname Volume 2 [01].mkv`

**Issues**

#3669

### DIFF
--- a/MediaBrowser.Providers/Plugins/MusicBrainz/MusicBrainzAlbumProvider.cs
+++ b/MediaBrowser.Providers/Plugins/MusicBrainz/MusicBrainzAlbumProvider.cs
@@ -319,6 +319,12 @@ namespace MediaBrowser.Providers.Music
                     {
                         case "name-credit":
                         {
+                            if (reader.IsEmptyElement)
+                            {
+                                reader.Read();
+                                break;
+                            }
+
                             using var subReader = reader.ReadSubtree();
                             return ParseArtistNameCredit(subReader);
                         }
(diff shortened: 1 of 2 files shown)


cvium:
@fiochen Is this no longer needed?

xF4m3:
Okay So i don't want to open a new issue unless necessary, but i am coming from plex atm but I am also having issues with the Episode parser. 
As a heads up, I have this issue with lots of Series, not just with this particular one, and on plex this is not an issue at all.
My folder Structure looks like this: /The Daily Life of the Immortal King [1080p]/Season 01/
Seams to detect the Season just fine, but the Episode names:
[SubsPlease] The Daily Life of the Immortal King - 01 (1080p) [969786E2].mkv
[SubsPlease] The Daily Life of the Immortal King - 02 (1080p) [B1E3FFA7].mkv
[SubsPlease] The Daily Life of the Immortal King - 03 (1080p) [AA66696D].mkv

They all End up being called Episode one, so now in this Series I got Season 01 with 15x Episode 1.
Of couse I could rename all my current Series for a better detection, but the issue is that I have alot of Animes that get downloaded on a weekly basis, which are named similar to this and renaming them Each after download is very teadious and as Plex shows not needed. I think the Episode/Season Parser need to be improved.

In Plex it is possible to change the Episode/Series Scanner on a Library Basis, just like the Metadata can be in Jellyfin, wouldnt this be a good idea to do aswell? People could then build plugins/scanners based on their needs.

For example this scanner works nearly flawless  in all my Librarys, regardless of Anime/TV shows/Movies etc. Maybe there is something in here that can help improve the Season/Episode detection.
https://github.com/ZeroQI/Absolute-Series-Scanner

Should i open a new issue or can we just reopen this one? Since I think this is related.

PR #6868, opened 2021-11-19
Fix docker build error armhf

eralmansouri:
**Changes**

I added a repo that contains libraspberrypi0 and specified the image for Debian buster instead of bullseye to resolve the other issues. It's not an ideal fix but better that than it not building at all I think.

**Issues**

When building on a raspberry pi, I get errors related to libraspberrypi0 not being found, but also errors related to libx265.


ferferga:
This won't be needed when https://github.com/jellyfin/jellyfin-ffmpeg/pull/77 is in

Bond-009:
Outdated

PR #6871, opened 2021-11-19
Only force scheme and port when HTTPS is disallowed

cvium:
It shouldn't force a scheme and/or port when https is allowed. The call to `GetLocalApiUrl` will decide that.

### DIFF
--- a/Emby.Server.Implementations/ApplicationHost.cs
+++ b/Emby.Server.Implementations/ApplicationHost.cs
@@ -1128,12 +1128,12 @@ namespace Emby.Server.Implementations
         }
 
         /// <inheritdoc/>
-        public string GetApiUrlForLocalAccess(bool allowHttps)
+        public string GetApiUrlForLocalAccess(bool allowHttps = true)
         {
             // With an empty source, the port will be null
             string smart = NetManager.GetBindInterface(string.Empty, out _);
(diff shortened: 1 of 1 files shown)


1337joe (approved):

thornbill:
:+1: I have confirmed this fixes the issue I was having with streaming from a HDHR tuner.

crobibero (approved):

Emby.Server.Implementations/ApplicationHost.cs
```diff
@@ -1128,12 +1128,12 @@ public string GetSmartApiUrl(string hostname)
...
         /// <inheritdoc/>
-        public string GetApiUrlForLocalAccess(bool allowHttps)
+        public string GetApiUrlForLocalAccess(bool allowHttps = true)
         {
             // With an empty source, the port will be null
             string smart = NetManager.GetBindInterface(string.Empty, out _);
-            var scheme = allowHttps ? Uri.UriSchemeHttps : Uri.UriSchemeHttp;
-            var port = allowHttps ? HttpsPort : HttpPort;
+            var scheme = !allowHttps ? Uri.UriSchemeHttp : null;
+            int? port = !allowHttps ? HttpPort : null;
```
Bond-009, line 1136: IMO, the negation should be removed

PR #6874, opened 2021-11-19
Expose TMDb image scale options to user

1337joe:
Changes:
- Expose TMDb image scale options to user

image

### DIFF
--- a/MediaBrowser.Providers/Plugins/Tmdb/Api/TmdbController.cs
+++ b/MediaBrowser.Providers/Plugins/Tmdb/Api/TmdbController.cs
@@ -1,0 +1,41 @@
+using System.Net.Mime;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using TMDbLib.Objects.General;
+
+namespace MediaBrowser.Providers.Plugins.Tmdb.Api
+{
+    /// <summary>
+    /// The TMDb api controller.
+    /// </summary>
+    [ApiController]
+    [Authorize(Policy = "DefaultAuthorization")]
+    [Route("[controller]")]
+    [Produces(MediaTypeNames.Application.Json)]
+    public class TmdbController : ControllerBase
+    {
+        private readonly TmdbClientManager _tmdbClientManager;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TmdbController"/> class.
+        /// </summary>
+        /// <param name="tmdbClientManager">The TMDb client manager.</param>
+        public TmdbController(TmdbClientManager tmdbClientManager)
+        {
+            _tmdbClientManager = tmdbClientManager;
+        }
+
+        /// <summary>
+        /// Gets the TMDb image configuration options.
+        /// </summary>
+        /// <returns>The image portion of the TMDb client configuration.</returns>
+        [HttpGet("ClientConfiguration")]
+        [ProducesResponseType(StatusCodes.Status200OK)]
+        public async Task<ConfigImageTypes> TmdbClientConfiguration()
+        {
+            return (await _tmdbClientManager.GetClientConfiguration().ConfigureAwait(false)).Images;
+        }
+    }
+}
--- a/MediaBrowser.Providers/Plugins/Tmdb/Configuration/PluginConfiguration.cs
+++ b/MediaBrowser.Providers/Plugins/Tmdb/Configuration/PluginConfiguration.cs
@@ -1,3 +1,4 @@
+using System.Collections.Generic;
 using MediaBrowser.Model.Plugins;
 
 namespace MediaBrowser.Providers.Plugins.Tmdb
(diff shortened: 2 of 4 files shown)


cvium:
What's the motivation behind this?

1337joe:
Simple answer: Downloading scaled images results in consistently sized images for a given use (Primary/Backdrop/etc) and saves space over downloading original resolution for everything. For example, the original photo for a random actress in my library is 1270 x 1905 while the largest scaled size is 421 x 632, which is less than 1/5th the file size and is still plenty sharp enough to identify her in the thumbnail.

More complicated motivation:
I have a script that I've been using that pulls images from TMDb so that I can embed them directly in the mkv media files as I rip my movie collection. It pulls the largest scaled image instead of the original size, which gives me a consistent resolution for the images across my library. Configuring jellyfin to download the specified scaled images lets me just use the image browser to pick what I want, then jellyfin saves the metadata and image next to the file. Doesn't yet completely replace my other workflow, but is a step towards it.

cvium:
But Jellyfin will already scale the image if you ask it to. Granted it keeps the original, but still.

1337joe:
Correct, this is all about what's stored on disk and is more or less transparent from the UI (unless resolution is set low enough that the images aren't crisp at the size they're displayed).  Reducing the size of stored images reduces the size of the metadata directory, which for my docker install at least is the bulk of the size for the jellyfin config directory. Consistent resolution only matters to me because I'm taking the files jellyfin downloads and using them elsewhere.

There was a somewhat open-ended feature request to "Add some options for the TMDB metadata scraper." that got me looking at what settings would make sense to add. The cast member limit was called out in the code with a TODO, the image scaling options are a TMDb setting, and both were things I personally expected to use.

cvium:
Would it not make more sense to make the limit non-provider specific?

1337joe:
That would make sense, but it's also more involved to do it right:
- The image fetcher settings is the logical place (to me) to put this configuration, but:
  - Fetcher settings are by library and image type, Artists aren't tied to libraries, and Artists being the images I personally will scale down the most that makes this feel incomplete to me until that's resolved.
  - We don't have default/top-level fetcher settings. I think it would belong on the libraries > metadata settings tab like language, but I haven't looked into how that is inherited by libraries and I doubt there's something already in place to do that for the fetcher settings.
- I don't know the clients well enough to know what image sizes to offer as options for the various image types (besides the obvious >= 1280 width for backgrounds). If we're doing the scaling ourselves we should optimize sizes for how we use images.
- Unless we arbitrarily pick the TMDb image sizes as our scaling options we probably fall back to downloading the original size to scale ourselves. Not really a problem if we take that route, but I don't see provider-specific handling for it happening as you suggested.
- This adds more processing that will happen during the library scan/metadata fetch. I don't know how intensive a bunch of skia calls are, but limiting the size up front on the download from the image provider saves bandwidth instead.

If the condition for getting downloaded image scaling in at all is that it applies to everything this drops in priority on my list at least until we have a way to manage metadata providers for people.

The approach here is TMDb specific because TMDb actually has a list of image scaling sizes that they offer for download, so I just exposed that on the settings page instead of always downloading original resolution like we were doing before. I haven't explored other APIs as much, but I haven't noticed any others with scaling options.

cvium:
> Not really a problem if we take that route, but I don't see provider-specific handling for it happening as you suggested.

That's not a huge issue I think.

> This adds more processing that will happen during the library scan/metadata fetch. I don't know how intensive a bunch of skia calls are, but limiting the size up front on the download from the image provider saves bandwidth instead.

Correct, but it's more complex and every provider would be different. If the end goal is storage savings I don't think this is what we'd want.

> This adds more processing that will happen during the library scan/metadata fetch. I don't know how intensive a bunch of skia calls are, but limiting the size up front on the download from the image provider saves bandwidth instead.

That is true, but ffprobe is much more intensive. We could make an optional post-scan task that scales all images.

1337joe:
Adding an optional post-scan task is just another step of complexity for a feature that I haven't seen anyone besides me request and pushes this farther from where I'm comfortable trying to tackle it right now.

Getting back to the intent of this pull request:
1. The TMDb api has options for image size when downloading
2. The Jellyfin plugin selects a specific option
3. That option is a good default but not the one I would personally prefer

Is there something wrong with me adding to the TMDb plugin config so that the user can make the choice instead of being locked into the Jellyfin default selection?

MediaBrowser.Providers/Plugins/Tmdb/TmdbClientManager.cs
```diff
@@ -575,16 +576,51 @@ public void ConvertStillsToRemoteImageInfo(List<ImageData> images, string reques
...
+                }
+            }
+
             for (var i = 0; i < images.Count; i++)
             {
                 var image = images[i];
+
+                int width;
+                int height;
+                if (targetHeight.HasValue)
```
cvium, line 601: You don't have to go through all this trouble to set the dimensions. Most plugins don't even know the size in advance. You can set it to null and it will be read from the file.

MediaBrowser.Providers/Plugins/Tmdb/Configuration/PluginConfiguration.cs
```diff
@@ -26,5 +27,45 @@ public class PluginConfiguration : BasePluginConfiguration
...
+
+        /// <summary>
+        /// Gets or sets a value indicating the poster image size to fetch.
+        /// </summary>
+        public string? PosterSize { get; set; }
+
+        /// <summary>
+        /// Gets or sets the available options for poster size.
+        /// </summary>
+        public List<string> PosterSizeOptions { get; set; } = new List<string>();
```
cvium, line 39: This is not really part of the configuration. You're overwriting them every time you read the config from TMDb. A better approach is an XHR call to an endpoint that spits out the valid size options.

MediaBrowser.Providers/Plugins/Tmdb/Api/TmdbController.cs
```diff
@@ -0,0 +1,41 @@
...
+using TMDbLib.Objects.General;
+
+namespace MediaBrowser.Providers.Plugins.Tmdb.Api
+{
+    /// <summary>
+    /// The TMDb api controller.
+    /// </summary>
+    [ApiController]
+    [Authorize(Policy = "DefaultAuthorization")]
+    [Route("[controller]")]
```
cvium, line 15: Typically the route is `<PluginNamespace>`, but we haven't standardized it (we really should though).

1337joe: `[controller]` resolves to "tmdb", you're saying I should set it to `[Route("tmdb")]` directly?

Checking a few plugins that have controllers:
[Trakt](https://github.com/jellyfin/jellyfin-plugin-trakt/blob/a581d4dce1bbc3a3632841f2b6fb219f04ce4329/Trakt/Api/TraktController.cs#L23) and [TmdbBoxSet](https://github.com/jellyfin/jellyfin-plugin-tmdbboxsets/blob/9ede4a11f6a525c360bd21628a9dad689a9f086c/Jellyfin.Plugin.TMDbBoxSets/Api/TMDbBoxSetsController.cs#L17) set `Route("[controller]")` like I did here.

[OpenSubtitles](https://github.com/jellyfin/jellyfin-plugin-opensubtitles/blob/1e83683bc3d02a25422d9b9302165164583deac5/Jellyfin.Plugin.OpenSubtitles/API/OpenSubtitlesController.cs#L19) and [KodiSyncQueue](https://github.com/jellyfin/jellyfin-plugin-kodisyncqueue/blob/7af5fef9eaac55200e67786646a077350340a36e/Jellyfin.Plugin.KodiSyncQueue/API/KodiSyncQueueController.cs#L23) don't set a `Route` at all.

[AutoOrganize](https://github.com/jellyfin/jellyfin-plugin-autoorganize/blob/73cc10464fd514d73f0eae8f1b80c2f30dc16b35/Emby.AutoOrganize/Api/FileOrganizationController.cs#L19) sets it to `Library/FileOrganizations`

cvium (approved):

Issue #3016, opened 2020-04-30
Improve Prometheus Exporter with JF Metrics


sparky8251:
Some initial ideas for Jellyfin specific metrics that administrators could find useful (will try and document names and types here as part of prep for making them):

- [ ] number of active sessions (include data on device type/client)
- [ ] number of total sessions (inactive are included here so you can get a broader idea of the full user base you serve. likely require some scheduled task to cleanup very inactive devices over time, but thats neither here nor there right now)
- [ ] number of active streams of each type (direct play, direct stream, transcode)
- [ ] number of active streams + sw/hw transcoding + codec
- [ ] number of and reason why, each play method above was chosen
- [ ] number of and reason why sw/hw accel method was chosen
- [ ] number and type of each library
- [ ] number and type of media codecs in each library (video, audio, sub, book, container, etc)
- [ ] number of user accounts
- [ ] number of groups
- [ ] number of users given specific permissions
- [ ] number of groups given specific permissions
- [ ] number of users that authenticate using an external source (+ the external source type)
- [ ] number of groups that map to an external source (+ the external source type)

Artiume:
What about some backend metrics? How long did a scan take to perform. Disc Writing/Reading metrics. RAM/CPU usage.

sparky8251:
RAM and CPU usage is not easy to determine as JF calls a number of external programs to perform tasks. Disk IO is the same. There's no real way for JF to know what those spawned processes are consuming for resources as far as I'm aware.

Additionally, you can use node_exporter to get RAM/Disk stats for the whole machine and you can use that to at least see JF caused spikes on the machine with a little intuition.

Scan performance seems less useful to me? Scans are very uncommon events and I'm unsure how to expose the data in a way that's useful if you end up with 1 scan a day since we would likely want all historical data. The scan system is also a mess... Not sure it would be a good idea to try and plug in to it right now.

falstaff1288:
Maybe the IP of the clients?

frebib:
Metrics around periodic tasks including media import, transcode tasks, other scheduled tasks. Metadata updates. Failures and errors for anything/everything that could possibly fail in some way. Jellyfin version updates available

frenos:
Is there still any activity going on to get something like this working?

To make it even more complicated I would also love to have user specific stats in sub-directories. You could extend the `/metrics` endpoint with `/metrics/<user-id>` and let prometheus also scrape specific users you are interested in.
Here we could have more detailed metrics:
[...]

PR #2985, opened 2020-04-26
Add Prometheus exporters

sparky8251:
This PR adds some basic Prometheus specific functionality. No JF specific metrics as of yet, but hopefully sets the stage to let us make some in the future :)

**Changes**
Adds a `/metrics` endpoint with Prometheus stats about the dotnet runtime and aspnet http server (which will become more useful as more API endpoints migrate to aspnet)

ToDo:

- [x] Make config change and default Prometheus metrics to off since its served on the same port as everything else and can be considered a leak of sensitive information (securable with a reverse proxy on the relevant path)
- [x] Confirm respects baseurl and if not, decide if it should or not

PR #6882, opened 2021-11-21
Create prometheus metric for item counts

holahmeds:
**Changes**
Create prometheus metric for number of items in library, by item type:
* Movie
* Series
* Season
* Episode
* MusicArtist
* MusicAlbum
* MusicVideo
* Audio
* Book
* PhotoAlbum
* Photo

**Issues**

 #3016

### DIFF
--- a/Emby.Server.Implementations/Metrics/PrometheusMetricsCollector.cs
+++ b/Emby.Server.Implementations/Metrics/PrometheusMetricsCollector.cs
@@ -1,0 +1,75 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using MediaBrowser.Controller.Entities;
+using MediaBrowser.Controller.Entities.Audio;
+using MediaBrowser.Controller.Entities.Movies;
+using MediaBrowser.Controller.Entities.TV;
+using MediaBrowser.Controller.Library;
+using MediaBrowser.Controller.Metrics;
+using Prometheus;
+using Prometheus.DotNetRuntime;
+
+namespace Emby.Server.Implementations.Metrics
+{
+    /// <summary>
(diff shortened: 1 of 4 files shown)


holahmeds:
Hi all, first time contributor here, so if there is a better place to create this metric please let me know.

I plan to add more metrics (through subsequent PRs) as I get more familiar with the code.

cvium:
It feels like something that should be in a plugin. I definitely don't like having Prometheus directly in LibraryManager.

holahmeds:
The discussion in #2985 seems to indicate that all the planned metrics may not be possible with a plugin. It is more that a year old though, so if it is no longer true I can look into creating a plugin.

Emby.Server.Implementations/Metrics/PrometheusMetricsCollector.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="PrometheusMetricsCollector" /> class.
+        /// </summary>
+        /// <param name="libraryManager">The library manager</param>
+        public PrometheusMetricsCollector(ILibraryManager libraryManager)
+        {
+            _libraryManager = libraryManager;
+            _libraryManager.ItemAdded += IncrementItemCount;
+            _libraryManager.ItemRemoved += (s, e) => UpdateItemCount();
```
holahmeds, line 42: `ItemRemoved` only seems to provide the parent item for e.g. If a series is removed, it fires once for the series but not the episodes. Calling `GetRecursiveChildren()` on it does not give them either. Is there another way to get all individual items removed?

holahmeds:
I have moved the metric out of Library manager and into a new class. All prometheus metrics can be put here.

robertas90poe (approved):

Issue #143, opened 2018-12-14
Support external audio tracks


JustAMan:
This should allow easier support for selecting audio track without transcoding or remuxing.

PR #6898, opened 2021-11-22
Add support for external audio files

jonas-resch:
**Changes**
Adds support for external audio files. Works similar to external subtitles.
Once this feature is released the docs need to be updated ([https://jellyfin.org/docs/general/server/media/movies.html](https://jellyfin.org/docs/general/server/media/movies.html) or new subentry like Subtitles?)

**Issues**
[#143](https://github.com/jellyfin/jellyfin/issues/143) / [https://features.jellyfin.org/posts/177/support-external-audio-tracks](https://features.jellyfin.org/posts/177/support-external-audio-tracks)


MediaBrowser.Providers/MediaInfo/AudioResolver.cs
```diff
@@ -0,0 +1,275 @@
...
+
+        private readonly IMediaEncoder _mediaEncoder;
+
+        private readonly CancellationToken _cancellationToken;
+
+        public AudioResolver(ILocalizationManager localization, IMediaEncoder mediaEncoder, CancellationToken cancellationToken = default)
+        {
+            _localization = localization;
+            _mediaEncoder = mediaEncoder;
+            _cancellationToken = cancellationToken;
```
cvium, line 34: This shouldn't be a private field. It should be provided directly to the method that needs cancellation support.

MediaBrowser.Providers/MediaInfo/AudioResolver.cs
```diff
@@ -0,0 +1,275 @@
...
+
+                var fullName = files[i];
+                var extension = Path.GetExtension(fullName.AsSpan());
+                if (!IsAudioExtension(extension))
+                {
+                    continue;
+                }
+
+                Model.MediaInfo.MediaInfo mediaInfo = GetMediaInfo(fullName).Result;
+                MediaStream mediaStream = mediaInfo.MediaStreams.First();
```
cvium, line 109: `mka` is a matroska container and can contain multiple audio streams

MediaBrowser.Providers/MediaInfo/AudioResolver.cs
```diff
@@ -0,0 +1,275 @@
...
+                        languageSpan = languageSpan[(lastDot + 1)..];
+                        break;
+                    }
+
+                    // Try to translate to three character code
+                    // Be flexible and check against both the full and three character versions
+                    var language = languageSpan.ToString();
+                    var culture = _localization.FindLanguageInfo(language);
+
+                    language = culture == null ? language : culture.ThreeLetterISOLanguageName;
```
cvium, line 144: The language might be readily available in the mediainfo, so I would say that the filename should be a last resort

MediaBrowser.Providers/MediaInfo/AudioResolver.cs
```diff
@@ -0,0 +1,275 @@
+#nullable disable
```
cvium, line 1: New files should not have warnings or nullable disabled

cvium (changes requested):

MediaBrowser.Providers/MediaInfo/AudioResolver.cs
```diff
@@ -0,0 +1,275 @@
...
+                        Path = path,
+                        Protocol = MediaProtocol.File
+                    }
+                },
+                _cancellationToken);
+        }
+
+        private static ReadOnlySpan<char> NormalizeFilenameForAudioComparison(string filename)
+        {
+            // Try to account for sloppy file naming
```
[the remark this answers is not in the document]
jonas-resch: Where Should I put this method?

MediaBrowser.Providers/MediaInfo/AudioResolver.cs
```diff
@@ -0,0 +1,275 @@
...
+            {
+
+                var fullName = files[i];
+                var extension = Path.GetExtension(fullName.AsSpan());
+                if (!IsAudioExtension(extension))
+                {
+                    continue;
+                }
+
+                Model.MediaInfo.MediaInfo mediaInfo = GetMediaInfo(fullName).Result;
```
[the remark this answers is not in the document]
jonas-resch: I think I need some help here. I tried with async & await but I can't declare the method as async because some variables are ReadonlySpans which aren't allowed in async methods.

MediaBrowser.Providers/MediaInfo/AudioResolver.cs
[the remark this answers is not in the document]
cvium: I'm not sure. We can also just not have it. People should name their files better.

MediaBrowser.Providers/MediaInfo/AudioResolver.cs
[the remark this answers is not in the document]
cvium: So don't use Spans :) It's better to use strings and async than using spans and sync.

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -696,6 +696,12 @@ public string GetInputArgument(EncodingJobInfo state, EncodingOptions options)
                 arg.Append(" -i \"").Append(subtitlePath).Append('\"');
             }
 
+            if (state.AudioStream != null && state.AudioStream.IsExternal)
+            {
+                arg.Append(" -i ")
+                    .Append(string.Format(CultureInfo.InvariantCulture, "file:\"{0}\"", state.AudioStream.Path));
```
crobibero, line 702: Append multiple times instead of using string.format

MediaBrowser.Providers/MediaInfo/AudioResolver.cs
```diff
@@ -0,0 +1,164 @@
...
+            var streams = new List<MediaStream>();
+
+            if (!video.IsFileProtocol)
+            {
+                return streams;
+            }
+
+            List<string> paths = GetExternalAudioFiles(video, directoryService, namingOptions, clearCache);
+
+            await AddExternalAudioStreams(streams, paths, startIndex, localizationManager, mediaEncoder, cancellationToken);
```
crobibero, line 43:
```suggestion
            await AddExternalAudioStreams(streams, paths, startIndex, localizationManager, mediaEncoder, cancellationToken).ConfigureAwait(false);
```

MediaBrowser.Providers/MediaInfo/FFProbeVideoInfo.cs
```diff
@@ -214,6 +218,8 @@ public class FFProbeVideoInfo
 
             await AddExternalSubtitles(video, mediaStreams, options, cancellationToken).ConfigureAwait(false);
 
+            await AddExternalAudio(video, mediaStreams, options, cancellationToken);
```
crobibero, line 221:
```suggestion
            await AddExternalAudio(video, mediaStreams, options, cancellationToken).ConfigureAwait(false);
```

MediaBrowser.Providers/MediaInfo/AudioResolver.cs
```diff
@@ -0,0 +1,164 @@
...
+        {
+            List<string> paths = new List<string>();
+
+            if (!video.IsFileProtocol)
+            {
+                return paths;
+            }
+
+            paths.AddRange(GetAudioFilesFromFolder(video.ContainingFolderPath, video.Path, directoryService, namingOptions, clearCache));
+            paths.AddRange(GetAudioFilesFromFolder(video.GetInternalMetadataPath(), video.Path, directoryService, namingOptions, clearCache));
```
cvium, line 62: It's very unlikely that the audio files are in the internal metadata folder. Subtitles are a special case because the subtitle downloader plugins can opt to save them in there, but no such thing exists for audio files.
```suggestion
```

MediaBrowser.Providers/MediaInfo/AudioResolver.cs
```diff
@@ -0,0 +1,164 @@
...
+            bool clearCache)
+        {
+            List<string> paths = new List<string>();
+            string videoFileNameWithoutExtension = Path.GetFileNameWithoutExtension(videoFileName);
+
+            if (!Directory.Exists(folder))
+            {
+                return paths;
+            }
+
```
cvium, line 81: No need to allocate if we're gonna exit early
```suggestion
            if (!Directory.Exists(folder))
            {
                return paths;
            }

            List<string> paths = new List<string>();
            string videoFileNameWithoutExtension = Path.GetFileNameWithoutExtension(videoFileName);
```

MediaBrowser.Providers/MediaInfo/AudioResolver.cs
```diff
@@ -0,0 +1,164 @@
...
+            var files = directoryService.GetFilePaths(folder, clearCache, true);
+            for (int i = 0; i < files.Count; i++)
+            {
+                string fileNameWithoutExtension = Path.GetFileNameWithoutExtension(files[i]);
```
cvium, line 85: Indexing into an array is O(1), but you should still put it in a variable if you access it more than once

MediaBrowser.Providers/MediaInfo/AudioResolver.cs
```diff
@@ -0,0 +1,164 @@
...
+                if (!AudioFileParser.IsAudioFile(files[i], namingOptions))
+                {
+                    continue;
+                }
+
+                // The audio filename must either be equal to the video filename or start with the video filename followed by a dot
+                if (videoFileNameWithoutExtension.Equals(fileNameWithoutExtension, StringComparison.OrdinalIgnoreCase) ||
+                    (fileNameWithoutExtension.Length > videoFileNameWithoutExtension.Length
+                         && fileNameWithoutExtension[videoFileNameWithoutExtension.Length] == '.'
+                         && fileNameWithoutExtension.StartsWith(videoFileNameWithoutExtension, StringComparison.OrdinalIgnoreCase)))
```
cvium, line 96: Indentation is off
```suggestion
                if (videoFileNameWithoutExtension.Equals(fileNameWithoutExtension, StringComparison.OrdinalIgnoreCase)
                    || (fileNameWithoutExtension.Length > videoFileNameWithoutExtension.Length
                        && fileNameWithoutExtension[videoFileNameWithoutExtension.Length] == '.'
                        && fileNameWithoutExtension.StartsWith(videoFileNameWithoutExtension, StringComparison.OrdinalIgnoreCase)))
```

MediaBrowser.Providers/MediaInfo/AudioResolver.cs
```diff
@@ -0,0 +1,164 @@
...
+                foreach (MediaStream mediaStream in mediaInfo.MediaStreams)
+                {
+                    mediaStream.Index = startIndex++;
+                    mediaStream.Type = MediaStreamType.Audio;
+                    mediaStream.IsExternal = true;
+                    mediaStream.Path = path;
+                    mediaStream.IsDefault = false;
+                    mediaStream.Title = null;
+
+                    if (mediaStream.Language == null)
```
cvium, line 127:
```suggestion
                    if (string.IsNullOrEmpty(mediaStream.Language))
```

MediaBrowser.Providers/MediaInfo/AudioResolver.cs
```diff
@@ -0,0 +1,164 @@
...
+
+namespace MediaBrowser.Providers.MediaInfo
+{
+    public class AudioResolver
+    {
+        public async Task<List<MediaStream>> GetExternalAudioStreams(
+            Video video,
+            int startIndex,
+            IDirectoryService directoryService,
+            NamingOptions namingOptions,
```
cvium, line 28: Instead of having so many deps as arguments to the method, you should make a constructor. Just don't `new` the class whenever you use it :)

The deps in question are:
1. NamingOptions
2. ILocalizationManager
3. IMediaEncoder

MediaBrowser.Providers/MediaInfo/AudioResolver.cs
```diff
@@ -0,0 +1,164 @@
...
+using MediaBrowser.Model.Dto;
+using MediaBrowser.Model.Entities;
+using MediaBrowser.Model.Globalization;
+using MediaBrowser.Model.MediaInfo;
+
+namespace MediaBrowser.Providers.MediaInfo
+{
+    public class AudioResolver
+    {
+        public async Task<List<MediaStream>> GetExternalAudioStreams(
```
cvium, line 24: It's generally frowned upon to return a list
```suggestion
        public async Task<IAsyncEnumerable<MediaStream>> GetExternalAudioStreams(
```

MediaBrowser.Providers/MediaInfo/AudioResolver.cs
```diff
@@ -0,0 +1,164 @@
...
+                         && fileNameWithoutExtension.StartsWith(videoFileNameWithoutExtension, StringComparison.OrdinalIgnoreCase)))
+                {
+                    paths.Add(files[i]);
+                }
+            }
+
+            return paths;
+        }
+
+        public async Task AddExternalAudioStreams(
```
cvium, line 105: Afaict, this method is only used once in this class. You could just move the body of the method into the calling method `GetExternalAudioStreams`.

MediaBrowser.Providers/MediaInfo/AudioResolver.cs
```diff
@@ -0,0 +1,164 @@
...
+                return paths;
+            }
+
+            paths.AddRange(GetAudioFilesFromFolder(video.ContainingFolderPath, video.Path, directoryService, namingOptions, clearCache));
+            paths.AddRange(GetAudioFilesFromFolder(video.GetInternalMetadataPath(), video.Path, directoryService, namingOptions, clearCache));
+
+            return paths;
+        }
+
+        private List<string> GetAudioFilesFromFolder(
```
cvium, line 67: You can merge this method with `GetExternalAudioFiles` if you remove the internal metadata code path.

MediaBrowser.Providers/MediaInfo/AudioResolver.cs
```diff
@@ -0,0 +1,164 @@
...
+            }
+
+            List<string> paths = GetExternalAudioFiles(video, directoryService, namingOptions, clearCache);
+
+            await AddExternalAudioStreams(streams, paths, startIndex, localizationManager, mediaEncoder, cancellationToken);
+
+            return streams;
+        }
+
+        public List<string> GetExternalAudioFiles(
```
cvium, line 48:
```suggestion
        public IEnumerable<string> GetExternalAudioFiles(
```

MediaBrowser.Providers/MediaInfo/FFProbeProvider.cs
```diff
@@ -111,6 +116,15 @@ public bool HasChanged(BaseItem item, IDirectoryService directoryService)
                 return true;
             }
 
+            AudioResolver audioResolver = new AudioResolver();
```
cvium, line 119: This should be instantiated in the constructor of `FFProbeProvider`

MediaBrowser.Providers/MediaInfo/FFProbeProvider.cs
```diff
@@ -111,6 +116,15 @@ public bool HasChanged(BaseItem item, IDirectoryService directoryService)
...
+            if (item.SupportsLocalMetadata && video != null && !video.IsPlaceHolder
+                && !video.AudioFiles.SequenceEqual(
+                        audioResolver.GetExternalAudioFiles(video, directoryService, _namingOptions, false), StringComparer.Ordinal))
+            {
+                _logger.LogDebug("Refreshing {0} due to external audio change.", item.Path);
```
cvium, line 124: Always use meaningful names in the log format strings
```suggestion
                _logger.LogDebug("Refreshing {ItemPath} due to external audio change.", item.Path);
```

MediaBrowser.Providers/MediaInfo/AudioResolver.cs
jonas-resch: I'll implement this manually since the List<string> paths = ... must be done before the if(!Directory.Exists(folder)) :)

cvium: Right, sorry. With the return type changed to `IEnumerable` you can use `Enumerable.Empty<string>()`.

MediaBrowser.Providers/MediaInfo/FFProbeProvider.cs
jonas-resch: Sorry, but I literally copied the debug line from [a few lines above](https://github.com/jellyfin/jellyfin/blob/4ac2e24907e25b07979d76e571d73e102c1348e3/MediaBrowser.Providers/MediaInfo/FFProbeProvider.cs#L115). Should I change it nonetheless or change it for every debug line in this method? :)

MediaBrowser.Providers/MediaInfo/AudioResolver.cs
jonas-resch: With not newing the class you mean the three arguments mentioned, correct? Or do you mean the AudioResolver? :)

MediaBrowser.Providers/MediaInfo/FFProbeProvider.cs
cvium: It's only important that new code follows the standards. If there's some quick fixes in the surrounding code that's ok, but PRs can quickly balloon in size with unrelated changes.

MediaBrowser.Providers/MediaInfo/AudioResolver.cs
cvium: I mean `AudioResolver`. It should only be newed once if possible.

MediaBrowser.Providers/MediaInfo/AudioResolver.cs
```diff
@@ -0,0 +1,147 @@
...
+            }
+
+            // Check if video folder exists
+            string folder = video.ContainingFolderPath;
+            if (!Directory.Exists(folder))
+            {
+                yield break;
+            }
+
+            string videoFileNameWithoutExtension = Path.GetFileNameWithoutExtension(video.Path);
```
cvium, line 106: Since this method isn't async, could you not use spans here?

jonas-resch: Just this variable here? or also the others in this method?

cvium: I try to use Spans when possible. You don't need the extra string that is created by `Path.GetFileNameWithoutExtension`, you only use it for comparisons.

jonas-resch: Ahh, now I understand what you mean. Thought you meant the variable type span :) I'll change it this evening.

jonas-resch: I can't use spans here since the function returns an IEnumerable :)

cvium: It's only async methods that can't use spans directly.

MediaBrowser.Providers/MediaInfo/AudioResolver.cs
```diff
@@ -0,0 +1,146 @@
+#pragma warning disable CA1002, CS1591
+
+using System;
```
cvium, line 3: New files shouldn't need to disable warnings.

```suggestion
using System;
```

MediaBrowser.Providers/MediaInfo/AudioResolver.cs
```diff
@@ -0,0 +1,146 @@
...
+            for (int i = 0; i < files.Count; i++)
+            {
+                string file = files[i];
+                string fileNameWithoutExtension = Path.GetFileNameWithoutExtension(file);
+
+                if (!AudioFileParser.IsAudioFile(file, _namingOptions))
+                {
+                    continue;
+                }
+
```
cvium, line 117:
```suggestion
                if (!AudioFileParser.IsAudioFile(file, _namingOptions))
                {
                    continue;
                }

                string fileNameWithoutExtension = Path.GetFileNameWithoutExtension(file);
```

MediaBrowser.Providers/MediaInfo/FFProbeVideoInfo.cs
```diff
@@ -214,6 +217,8 @@ public class FFProbeVideoInfo
...
```
cvium, line 220:
```suggestion
            await AddExternalAudio(video, mediaStreams, options, cancellationToken).ConfigureAwait(false);
```

MediaBrowser.Providers/MediaInfo/AudioResolver.cs
```diff
@@ -0,0 +1,146 @@
...
+            if (!video.IsFileProtocol)
+            {
+                yield break;
+            }
+
+            IEnumerable<string> paths = GetExternalAudioFiles(video, directoryService, clearCache);
+            foreach (string path in paths)
+            {
+                string fileNameWithoutExtension = Path.GetFileNameWithoutExtension(path);
+                Model.MediaInfo.MediaInfo mediaInfo = await GetMediaInfo(path, cancellationToken);
```
cvium, line 57:
```suggestion
                Model.MediaInfo.MediaInfo mediaInfo = await GetMediaInfo(path, cancellationToken).ConfigureAwait(false);
```

cvium:
I did some testing and you need this patch to avoid some crashes. Another thing is that it doesn't properly handle shortcut file types like `.strm`.
```
Index: MediaBrowser.Controller/Entities/Video.cs
IDEA additional info:
<+>UTF-8
diff --git a/MediaBrowser.Controller/Entities/Video.cs b/MediaBrowser.Controller/Entities/Video.cs
--- a/MediaBrowser.Controller/Entities/Video.cs	(revision 6bbfcf1906b7a35da3db4954f9a3762bca9f3a93)
+++ b/MediaBrowser.Controller/Entities/Video.cs	(date 1638344442748)
@@ -33,6 +33,7 @@
             AdditionalParts = Array.Empty<string>();
             LocalAlternateVersions = Array.Empty<string>();
             SubtitleFiles = Array.Empty<string>();
+            AudioFiles = Array.Empty<string>();
             LinkedAlternateVersions = Array.Empty<LinkedChild>();
         }
 
Index: MediaBrowser.Providers/MediaInfo/AudioResolver.cs
IDEA additional info:
<+>UTF-8
diff --git a/MediaBrowser.Providers/MediaInfo/AudioResolver.cs b/MediaBrowser.Providers/MediaInfo/AudioResolver.cs
--- a/MediaBrowser.Providers/MediaInfo/AudioResolver.cs	(revision 6bbfcf1906b7a35da3db4954f9a3762bca9f3a93)
+++ b/MediaBrowser.Providers/MediaInfo/AudioResolver.cs	(date 1638345053869)
@@ -131,7 +131,7 @@
             for (int i = 0; i < files.Count; i++)
             {
                 string file = files[i];
-                if (!AudioFileParser.IsAudioFile(file, _namingOptions))
+                if (string.Equals(video.Path, file, StringComparison.OrdinalIgnoreCase) || !AudioFileParser.IsAudioFile(file, _namingOptions))
                 {
                     continue;
                 }

**Comment by jonas-resch** [comment]
> I did some testing and you need this patch to avoid some crashes. Another thing is that it doesn't properly handle shortcut file types like `.strm`.
> 
> ```
> Index: MediaBrowser.Controller/Entities/Video.cs
> IDEA additional info:
> Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
> <+>UTF-8
> ===================================================================
> diff --git a/MediaBrowser.Controller/Entities/Video.cs b/MediaBrowser.Controller/Entities/Video.cs
> --- a/MediaBrowser.Controller/Entities/Video.cs	(revision 6bbfcf1906b7a35da3db4954f9a3762bca9f3a93)
> +++ b/MediaBrowser.Controller/Entities/Video.cs	(date 1638344442748)
> @@ -33,6 +33,7 @@
>              AdditionalParts = Array.Empty<string>();
>              LocalAlternateVersions = Array.Empty<string>();
>              SubtitleFiles = Array.Empty<string>();
> +            AudioFiles = Array.Empty<string>();
>              LinkedAlternateVersions = Array.Empty<LinkedChild>();
>          }
>  
> Index: MediaBrowser.Providers/MediaInfo/AudioResolver.cs
> IDEA additional info:
> Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
> <+>UTF-8
> ===================================================================
> diff --git a/MediaBrowser.Providers/MediaInfo/AudioResolver.cs b/MediaBrowser.Providers/MediaInfo/AudioResolver.cs
> --- a/MediaBrowser.Providers/MediaInfo/AudioResolver.cs	(revision 6bbfcf1906b7a35da3db4954f9a3762bca9f3a93)
> +++ b/MediaBrowser.Providers/MediaInfo/AudioResolver.cs	(date 1638345053869)
> @@ -131,7 +131,7 @@
>              for (int i = 0; i < files.Count; i++)
>              {
>                  string file = files[i];
> -                if (!AudioFileParser.IsAudioFile(file, _namingOptions))
> +                if (string.Equals(video.Path, file, StringComparison.OrdinalIgnoreCase) || !AudioFileParser.IsAudioFile(file, _namingOptions))
>                  {
>                      continue;
>                  }
> ```

Thanks for your input :) I implemented the two changes (didn't push them yet).
Regarding the `.strm` files:

- What do I do if the `.strm` file or the stream url it contains is invalid?
- To parse the `.strm` file I am also using the `NormalizeStrmLine` method found in `FFProbeProvider`. For now I copied it into the `AudioResolver` class. Where should I put it so this code isn't duplicate?

File: MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs [hunk 27]
```diff
@@ -696,6 +696,11 @@ public string GetInputArgument(EncodingJobInfo state, EncodingOptions options)
                 arg.Append(" -i \"").Append(subtitlePath).Append('\"');
             }
 
+            if (state.AudioStream != null && state.AudioStream.IsExternal)
+            {
+                arg.Append(" -i \"").Append(state.AudioStream.Path).Append("\"");
```
crobibero, line 701:
```suggestion
                arg.Append(" -i \"").Append(state.AudioStream.Path).Append('"');
```

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -1999,10 +2004,24 @@ public string GetMapArgs(EncodingJobInfo state)
...
             if (state.AudioStream != null)
             {
-                args += string.Format(
-                    CultureInfo.InvariantCulture,
-                    " -map 0:{0}",
-                    state.AudioStream.Index);
+                if (state.AudioStream.IsExternal)
+                {
+                    int externalAudioMapIndex = state.SubtitleStream != null && state.SubtitleStream.IsExternal ? 2 : 1;
+                    int externalAudioStream = state.MediaSource.MediaStreams.Where(i => i.Path == state.AudioStream.Path).ToList().IndexOf(state.AudioStream);
```
crobibero, line 2010:
```suggestion
                    int externalAudioStream = state.MediaSource.MediaStreams.FindIndex(i => i.Path == state.AudioStream.Path);
```
Shouldn't there be only one audio stream that has the same path?

jonas-resch: No, `.mka` files can have multiple audio streams in one file (didn't know that either until @cvium told me) :)

jonas-resch:
> > What do I do if the .strm file or the stream url it contains is invalid?
> 
> Don't crash. Log an error and skip the file.
> 
> > To parse the .strm file I am also using the NormalizeStrmLine method found in FFProbeProvider. For now I copied it into the AudioResolver class. Where should I put it so this code isn't duplicate?
> 
> I'm not sure. Maybe a new class in the same folder called `ShortcutHelper`. You could also create an `IShortcutHandler` but I think that's out of scope.

Ok, thanks. I encountered one issue while implementing support for .strm files: the .strm file gets added as a secondary video version even if the stream it references only contains audio. Do you know where the scanning for additional video versions happens so I can fix this issue? I searched for ~1h but couldn't find it. Thanks :)

MediaBrowser.Providers/MediaInfo/FFProbeProvider.cs
```diff
@@ -39,10 +40,12 @@ public class FFProbeProvider : ICustomMetadataProvider<Episode>,
     {
         private readonly ILogger<FFProbeProvider> _logger;
         private readonly SubtitleResolver _subtitleResolver;
+        private readonly AudioResolver _audioResolver;
         private readonly FFProbeVideoInfo _videoProber;
         private readonly FFProbeAudioInfo _audioProber;
 
         private readonly Task<ItemUpdateType> _cachedTask = Task.FromResult(ItemUpdateType.None);
+        private readonly NamingOptions _namingOptions;
```
cvium, line 48:
```suggestion
```

MediaBrowser.Providers/MediaInfo/FFProbeProvider.cs
```diff
@@ -55,10 +57,11 @@ public class FFProbeProvider : ICustomMetadataProvider<Episode>,
...
             IChapterManager chapterManager,
-            ILibraryManager libraryManager)
+            ILibraryManager libraryManager,
+            NamingOptions namingOptions)
         {
             _logger = logger;
-
+            _audioResolver = new AudioResolver(localization, mediaEncoder, namingOptions);
             _subtitleResolver = new SubtitleResolver(BaseItem.LocalizationManager);
             _videoProber = new FFProbeVideoInfo(
```
cvium, line 66: I messed up the merge. Can you re-add `_audioResolver`?

cvium:
> > > What do I do if the .strm file or the stream url it contains is invalid?
> > 
> > 
> > Don't crash. Log an error and skip the file.
> > > To parse the .strm file I am also using the NormalizeStrmLine method found in FFProbeProvider. For now I copied it into the AudioResolver class. Where should I put it so this code isn't duplicate?
> > 
> > 
> > I'm not sure. Maybe a new class in the same folder called `ShortcutHelper`. You could also create an `IShortcutHandler` but I think that's out of scope.
> 
> Ok, thanks. I encountered one issue while implementing support for .strm files: the .strm file gets added as a secondary video version even if the stream it references only contains audio. Do you know where the scanning for additional video versions happens so I can fix this issue? I searched for ~1h but couldn't find it. Thanks :)

Apologies for the delay. I think we should just drop strm support for now. Change this line https://github.com/jellyfin/jellyfin/pull/6898/files#diff-6e6909cb7c6a12c02d72cb84949cd92585e22998a8b48e9bb716150e52f10c5fR134 to also filter out strm files.

jonas-resch:
> > > > What do I do if the .strm file or the stream url it contains is invalid?
> > > 
> > > 
> > > Don't crash. Log an error and skip the file.
> > > > To parse the .strm file I am also using the NormalizeStrmLine method found in FFProbeProvider. For now I copied it into the AudioResolver class. Where should I put it so this code isn't duplicate?
> > > 
> > > 
> > > I'm not sure. Maybe a new class in the same folder called `ShortcutHelper`. You could also create an `IShortcutHandler` but I think that's out of scope.
> > 
> > 
> > Ok, thanks. I encountered one issue while implementing support for .strm files: the .strm file gets added as a secondary video version even if the stream it references only contains audio. Do you know where the scanning for additional video versions happens so I can fix this issue? I searched for ~1h but couldn't find it. Thanks :)
> 
> Apologies for the delay. I think we should just drop strm support for now. Change this line https://github.com/jellyfin/jellyfin/pull/6898/files#diff-6e6909cb7c6a12c02d72cb84949cd92585e22998a8b48e9bb716150e52f10c5fR134 to also filter out strm files.

No problem :) I excluded .strm files. Maybe I'll implement it in the future.

MediaBrowser.Providers/MediaInfo/AudioResolver.cs
```diff
@@ -0,0 +1,174 @@
...
+                yield break;
+            }
+
+            string videoFileNameWithoutExtension = Path.GetFileNameWithoutExtension(video.Path);
+
+            var files = directoryService.GetFilePaths(folder, clearCache, true);
+            for (int i = 0; i < files.Count; i++)
+            {
+                string file = files[i];
+                if (string.Equals(video.Path, file, StringComparison.OrdinalIgnoreCase) || !AudioFileParser.IsAudioFile(file, _namingOptions) || Path.GetExtension(file).Equals(".strm", StringComparison.OrdinalIgnoreCase))
```
cvium, line 134:
```suggestion
                if (string.Equals(video.Path, file, StringComparison.OrdinalIgnoreCase)
                    || !AudioFileParser.IsAudioFile(file, _namingOptions)
                    || Path.GetExtension(file.AsSpan()).Equals(".strm", StringComparison.OrdinalIgnoreCase))
```

crobibero (approved):

PR #6920, opened 2021-11-28
Add an option for dropping specific subtitle formats using the DLNA SubtitleProfile

marius-luca-87:
Currently there is no option to stop DLNA from transcoding when encountering specific subtitle formats.
A new Drop option was added to the SubtitleProfile that should allow us to control when we want avoid transcoding.

For the end user it should be easier to ignore embedded PGSSUB subtitles from the DLNA profile and provide an external subtitle than to modify large video files.

### DIFF
--- a/Jellyfin.Api/Helpers/DynamicHlsHelper.cs
+++ b/Jellyfin.Api/Helpers/DynamicHlsHelper.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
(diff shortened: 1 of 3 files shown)


MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -1220,7 +1220,9 @@ private void LogConditionFailure(DeviceProfile profile, string type, ProfileCond
             {
                 var subtitleProfile = GetSubtitleProfile(item, subtitleStream, options.Profile.SubtitleProfiles, playMethod, _transcoderSupport, item.Container, null);
 
-                if (subtitleProfile.Method != SubtitleDeliveryMethod.External && subtitleProfile.Method != SubtitleDeliveryMethod.Embed)
+                if (subtitleProfile.Method != SubtitleDeliveryMethod.Drop &&
+                    subtitleProfile.Method != SubtitleDeliveryMethod.External &&
+                    subtitleProfile.Method != SubtitleDeliveryMethod.Embed)
```
crobibero, line 1225:
```suggestion
                if (subtitleProfile.Method != SubtitleDeliveryMethod.Drop
                    && subtitleProfile.Method != SubtitleDeliveryMethod.External
                    && subtitleProfile.Method != SubtitleDeliveryMethod.Embed)
```

Jellyfin.Api/Helpers/DynamicHlsHelper.cs
```diff
@@ -462,6 +462,10 @@ private bool EnableAdaptiveBitrateStreaming(StreamState state, bool isLiveStream
 
         private void AddSubtitles(StreamState state, IEnumerable<MediaStream> subtitles, StringBuilder builder, ClaimsPrincipal user)
         {
+            if (state.SubtitleDeliveryMethod == SubtitleDeliveryMethod.Drop)
+            {
+                return;
+            }
```
crobibero, line 468:
```suggestion
            }

```

IDisposable (approved):

crobibero (approved):

1337joe (approved):

Issue #494, opened 2019-01-07
Replace MediaBrowser.Model/Net/MimeTypes.cs to be based on the IANA mime-type XML.


EraYaN:
The current way is error prone (hence there are some mim-type errors)

So we should be using something like the IANA XML of the Apache mime.types file and a resource and just have this class parse it and classify things.

anthonylavado:
This still needs to be done, right?

EraYaN:
Yes.

cvium:
https://www.nuget.org/packages/MimeTypes/

forresthopkinsa:
I'll try to tackle this.

holahmeds:
Should this be closed now that #6941 is merged?

PR #6941, opened 2021-12-04
Use MimeTypes package to determine MIME type

holahmeds:
This simplifies the code since we don't have to keep large mappings of extensions and MIME types. The current method is also reportedly error prone (#494).

We still keep the ability to override the mappings for:
- filling in entries not present in the package, for e.g. ".azw3"
- picking preferred extensions, for e.g. MimeTypes provides ".conf" as a possible extionsion for "text/plain", and while that is correct, ".txt" would be preferrable
- compatibility reasons

**Issues**
#494

Library: https://github.com/khellang/MimeTypes


MediaBrowser.Model/Net/MimeTypes.cs
```diff
@@ -231,6 +176,12 @@ public static class MimeTypes
                 return result;
             }
 
+            var extensions = Model.MimeTypes.GetMimeTypeExtensions(mimeType);
+            if (extensions.Any())
+            {
+                return "." + extensions.First();
+            }
+
             return null;
```
crobibero, line 185:
```suggestion
            var extension = Model.MimeTypes.GetMimeTypeExtensions(mimeType).FirstOrDefault();
            return string.IsNullOrEmpty(extension) ? null : "." + extension;
```

MediaBrowser.Model/Net/MimeTypes.cs
```diff
@@ -12,6 +12,10 @@ namespace MediaBrowser.Model.Net
     /// <summary>
     /// Class MimeTypes.
     /// </summary>
+    ///
+    /// http://en.wikipedia.org/wiki/Internet_media_type
```
cvium, line 16: Maybe this should be put in a `<remarks>` tag?

tests/Jellyfin.Model.Tests/Net/MimeTypesTests.cs
```diff
@@ -0,0 +1,164 @@
...
+        [InlineData(".oga", "audio/ogg")]
+        [InlineData(".ogg", "audio/ogg")]
+        [InlineData(".opus", "audio/ogg")]
+        [InlineData(".vorbis", "audio/vorbis")]
+        [InlineData(".wav", "audio/wav")]
+        [InlineData(".webma", "audio/webm")]
+        [InlineData(".wma", "audio/x-ms-wma")]
+        [InlineData(".wv", "audio/x-wavpack")]
+        [InlineData(".xsp", "audio/xsp")]
+        public void GetMimeType(string input, string expectedResult)
```
cvium, line 90: We usually follow the `MethodName_State_Expectation` naming scheme.
```suggestion
        public void GetMimeType_Valid_ReturnsCorrectResult(string input, string expectedResult)
```

tests/Jellyfin.Model.Tests/Net/MimeTypesTests.cs
```diff
@@ -0,0 +1,164 @@
...
+        [InlineData("video/quicktime", ".mov")]
+        [InlineData("video/vnd.mpeg.dash.mpd", ".mpd")]
+        [InlineData("video/webm", ".webm")]
+        [InlineData("video/x-flv", ".flv")]
+        [InlineData("video/x-m4v", ".m4v")]
+        [InlineData("video/x-matroska", ".mkv")]
+        [InlineData("video/x-ms-asf", ".asf")]
+        [InlineData("video/x-ms-wmv", ".wmv")]
+        [InlineData("video/x-msvideo", ".avi")]
+        public void ToExtension(string input, string expectedResult)
```
cvium, line 159:
```suggestion
        public void ToExtension_Valid_ReturnsCorrectResult(string input, string expectedResult)
```

crobibero (approved):

MediaBrowser.Model/Net/MimeTypes.cs
holahmeds: Updated

cvium (approved):

1337joe (approved):

PR #6974, opened 2021-12-11
Automatically bump stable versions in build

joshuaboniface:
Should prevent strangeness with building these for pre-releases or actual release versions (if the script wasn't manually run before). Whatever the Git tag is, becomes the package version. Could be expanded to avoid needing to manually `bump_version` at all ever if we put in placeholders, though this might break other builds. Would need careful evaluation before expanding that.

With a properly bumped stable release this shouldn't change anything, but for pre-releases it will change the version automatically.

Avoids needing to do #6853 manually (for both Debian and Fedora/CentOS packages).

Basically, adds a run of `bump_version` to the build CI, taking the Git tag as the version to bump to.

### DIFF
--- a/.ci/azure-pipelines-package.yml
+++ b/.ci/azure-pipelines-package.yml
@@ -39,6 +39,14 @@ jobs:
     vmImage: 'ubuntu-latest'
 
   steps:
+  - script: echo "##vso[task.setvariable variable=JellyfinVersion]$( awk -F '/' '{ print $NF }' <<<'$(Build.SourceBranch)' | sed 's/^v//' )"
+    displayName: Set release version (stable)
+    condition: startsWith(variables['Build.SourceBranch'], 'refs/tags/v')
+
+  - script: './bump-version $(JellyfinVersion)'
+    displayName: Bump internal version (stable)
+    condition: startsWith(variables['Build.SourceBranch'], 'refs/tags/v')
+
   - script: 'docker build -f deployment/Dockerfile.$(BuildConfiguration) -t jellyfin-server-$(BuildConfiguration) deployment'
     displayName: 'Build Dockerfile'
 
@@ -80,6 +88,14 @@ jobs:
     vmImage: 'ubuntu-latest'
 
   steps:
+  - script: echo "##vso[task.setvariable variable=JellyfinVersion]$( awk -F '/' '{ print $NF }' <<<'$(Build.SourceBranch)' | sed 's/^v//' )"
+    displayName: Set release version (stable)
+    condition: startsWith(variables['Build.SourceBranch'], 'refs/tags/v')
+
+  - script: './bump-version $(JellyfinVersion)'
+    displayName: Bump internal version (stable)
+    condition: startsWith(variables['Build.SourceBranch'], 'refs/tags/v')
+
   - task: DownloadPipelineArtifact@2
     displayName: 'Download OpenAPI Spec'
     inputs:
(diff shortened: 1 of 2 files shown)


crobibero (approved):

brianjmurrell:
So how does this actually work?  Are the files (i.e. `jellyfin.spec`) actually modified in the commit that is pushed with a tag on it before it's actually committed to git?

Where is the (i.e.) "~alpha3" supposed to be coming from?  https://github.com/jellyfin/jellyfin/blob/master/SharedVersion.cs doesn't seem to have any `alpha*` included in it, which is where it seems that `bump_version` should be reading it from.

I just want to understand what the end result is with this PR so that I can be sure it's going to work with packaging as expected.

joshuaboniface:
In the Azure CI, the repo will be cloned, and then checked out to the tag. Then, bump_version will be run which will alter the `fedora.spec` (and `debian/changelog`) to the tag version. Then the builds happen. Once the build finishes, these changes will be lost and never committed.

brianjmurrell:
> In the Azure CI, the repo will be cloned, and then checked out to the tag. Then, bump_version will be run which will alter the `fedora.spec` (and `debian/changelog`) to the tag version. Then the builds happen. Once the build finishes, these changes will be lost and never committed.

Indeed.  So the RPM and `debian/` packaging specifications will never reflect the actual version of the software it came with.  This is a bad design.  These build systems are meant to carry accurate versioning.  They are not intended to have an accurate version shoved in at the last minute, out-of-band, right before build.

Both debian and RPM build systems are intended to work without SCMs that provide tagging context, etc.  They are supposed to be build-able from stand-alone release artifacts such as a tarball, and in the case of debian optionally, a `debian/` patch file.  None of these (standard!) build mechanisms have any out-of-band channel to convey that what is being built is not actually what the artifacts say is being built but rather is actually something else.

brianjmurrell:
The more I look at the number of files that are changed with `bump_version` the more I am convinced that in the name of transparency (at least!), this is something that should be done and committed, not just done right before build.

This could be automated with githooks.  Maybe `pre-commit`.

joshuaboniface:
I agree that this is not ideal, but I'm trying to work around the confines of our current Azure workflow while minimizing the problems that these pre-release builds might cause to `master`. Our post-10.8.0 CI revamp should, hopefully, fix a lot of this, because at that point pre-tag commits and such can be automated too, and the releases/tags generated automatically like plugins are now. But we haven't had a chance to fully implement that yet, and I'm doing this to fix the specific complaint that these alpha/beta/rc releases don't have a "real" version in their packages. It really has to be one or the other here.

joshuaboniface:
I agree that this is not ideal, but I'm trying to work around the confines of our current Azure workflow while minimizing the problems that these pre-release builds might cause to `master`. Our post-10.8.0 CI revamp should, hopefully, fix a lot of this, because at that point pre-tag commits and such can be automated too, and the releases/tags generated automatically like plugins are now. But we haven't had a chance to fully implement that yet, and I'm doing this to fix the specific complaint that these alpha/beta/rc releases don't have a "real" version in their packages. It really has to be one or the other here.

joshuaboniface:
(This was "merged" automatically by merging #6980, which is strange, but OK.)

brianjmurrell:
> (This was "merged" automatically by merging #6980, which is strange, but OK.)

Yeah.  I really had intended #6980 to be based on (and mergeable into) this PR, but per the second paragraph in #6980:

> This PR really should be based on https://github.com/joshuaboniface/jellyfin/tree/ci-autoversion-packages but for whatever reason, joshuaboniface is not showing up in the base repository selector. As such this PR includes all of the code from #6974. If somebody who can, wants to rebase this on #6974 and land it to there, that is fine. Or this can wait for #6974 to land and then it can be rebased on master and then landed to there.

I really wanted #6980 code to be an addition to this PR and only landed when this PR was approved and landed.  I'm still puzzled why I couldn't find your fork in the base repository selector.  AFAIU all forks are supposed to show up there to allow one to merge into another fork if one wants.

joshuaboniface:
Yea that's very strange; my fork is public.

brianjmurrell:
Indeed, it was very strange.  There were numerous forks available to be the base, but yours was not one of them.

PR #6978, opened 2021-12-12
Add tests for ProbeResultNormalizer.GetFrameRate

Bond-009:


MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
```diff
@@ -1027,27 +1027,32 @@ private bool IsClose(double d1, double d2, double variance = .005)
...
+
+            if (!float.TryParse(value[..index], NumberStyles.Integer, CultureInfo.InvariantCulture, out var dividend)
+                || !float.TryParse(value[(index + 1)..], NumberStyles.Integer, CultureInfo.InvariantCulture, out var divisor))
             {
-                result = float.Parse(parts[0], CultureInfo.InvariantCulture);
+                return null;
             }
 
-            return float.IsNaN(result) ? null : result;
+            return divisor == 0f ? 0f : dividend / divisor;
```
Bond-009, line 1055: Yes, it's fine to compare directly with 0 here,
it isn't the result of a calculation or something and I only care about exactly 0.

tests/Jellyfin.MediaEncoding.Tests/Probing/ProbeResultNormalizerTests.cs
```diff
@@ -18,6 +18,19 @@ public class ProbeResultNormalizerTests
         private readonly JsonSerializerOptions _jsonOptions = JsonDefaults.Options;
         private readonly ProbeResultNormalizer _probeResultNormalizer = new ProbeResultNormalizer(new NullLogger<EncoderValidatorTests>(), null);
 
+        [Theory]
+        [InlineData("2997/125", 23.976f)]
+        [InlineData("1/50", 0.02f)]
+        [InlineData("25/1", 25f)]
+        [InlineData("120/1", 120f)]
+        [InlineData("1704753000/71073479", 23.98578237601117f)]
+        [InlineData("0/0", 0f)]
```
Bond-009, line 27: I think 0 is the preferred output instead of `null`, but I'm not 100%

crobibero (approved):

1337joe (approved):

MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
```diff
@@ -1027,27 +1027,32 @@ private bool IsClose(double d1, double d2, double variance = .005)
...
-            var parts = value.Split('/');
-
-            float result;
-
-            if (parts.Length == 2)
+            int index = value.IndexOf('/');
+            if (index == -1)
             {
-                result = float.Parse(parts[0], CultureInfo.InvariantCulture) / float.Parse(parts[1], CultureInfo.InvariantCulture);
+                // REVIEW: is this brach actually required? (i.e. does ffprobe ever output something other then a fraction?)
```
cvium, line 1040: Does it matter? Better safe than sorry.
```suggestion
```

tests/Jellyfin.MediaEncoding.Tests/Probing/ProbeResultNormalizerTests.cs
cvium: I disagree. Division by 0 is undefined. I would expect it to return null.

MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
Bond-009: It's an extra branch to cover in our tests :p

cvium: Unless you can prove that it's unnecessary I say leave it in

Bond-009: That's what the comment is for...
The code stays until I'm sure it isn't needed

cvium: Then change the comment to reflect that ;) And fix the typos and spelling errors

cvium (approved):

1337joe (approved):

crobibero (approved):

PR #6979, opened 2021-12-12
Use BaseItemKind where possible

crobibero:
Got a little big...

### DIFF
--- a/Emby.Server.Implementations/Data/SqliteItemRepository.cs
+++ b/Emby.Server.Implementations/Data/SqliteItemRepository.cs
@@ -196,7 +196,7 @@ namespace Emby.Server.Implementations.Data
 
         private static readonly string _mediaAttachmentInsertPrefix;
 
-        private static readonly HashSet<BaseItemKind> _programTypes = new()
+        private static readonly BaseItemKind[] _programTypes = new[]
         {
             BaseItemKind.Program,
             BaseItemKind.TvChannel,
@@ -204,7 +204,7 @@ namespace Emby.Server.Implementations.Data
             BaseItemKind.LiveTvChannel
         };
 
-        private static readonly HashSet<BaseItemKind> _programExcludeParentTypes = new()
+        private static readonly BaseItemKind[] _programExcludeParentTypes = new[]
         {
             BaseItemKind.Series,
             BaseItemKind.Season,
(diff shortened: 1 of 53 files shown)


tests/Jellyfin.Api.Tests/Helpers/RequestHelpersTests.cs
```diff
@@ -59,31 +59,7 @@ public static void GetOrderBy_Success(IReadOnlyList<string> sortBy, IReadOnlyLis
...
-                "AggregateFolder",
-                "Audio",
-                "BasePluginFolder",
-                "CollectionFolder"
-            };
-
-            var res = RequestHelpers.GetItemTypeStrings(input);
-
-            Assert.Equal(expected, res);
+            Assert.Empty(Array.Empty<BaseItemKind>());
```
nielsvanvelzen, line 62: Isn't this test useless now?

nielsvanvelzen (approved):
API changes look fine!

Jellyfin.Api/Controllers/LibraryController.cs
```diff
@@ -715,26 +716,26 @@ public async Task<ActionResult> GetDownload([FromRoute, Required] Guid itemId)
...
-                includeItemTypes.Add(nameof(Series));
+                includeItemTypes.Add(BaseItemKind.Series);
             }
             else
             {
                 // For non series and movie types these columns are typically null
                 isSeries = null;
                 isMovie = null;
-                includeItemTypes.Add(item.GetType().Name);
+                includeItemTypes.Add(item.GetBaseItemKind());
```
cvium, line 738: Now that GetBaseItemKind is used so often, can we cache the value? Enum.Parse isn't cheap.

1337joe (approved):

Emby.Server.Implementations/Data/SqliteItemRepository.cs
```diff
@@ -196,57 +196,56 @@ public class SqliteItemRepository : BaseSqliteRepository, IItemRepository
...
         {
-            "Audio",
-            "MusicAlbum",
-            "MusicVideo",
-            "AudioBook",
-            "AudioPodcast"
+            BaseItemKind.Audio,
+            BaseItemKind.MusicAlbum,
+            BaseItemKind.MusicVideo,
+            BaseItemKind.AudioBook
```
Bond-009, line 248: I think `HashSet` is the best collection type for this now, plain arrays should be fine

Bond-009 (approved):

1337joe:
Doing this on insert:

https://github.com/jellyfin/jellyfin/blob/b5d4fdb56e59772b925d75cfe1c8ba691dbe1e77/Emby.Server.Implementations/Data/SqliteItemRepository.cs#L718

is incompatible with doing this on query (when `includeTypes` contains `BaseItemKind`):

https://github.com/jellyfin/jellyfin/blob/b5d4fdb56e59772b925d75cfe1c8ba691dbe1e77/Emby.Server.Implementations/Data/SqliteItemRepository.cs#L3584

Will also need some migration work to convert existing installs to use `BaseItemKind` as the database type.

PR #7006, opened 2021-12-16
Fix BaseItemKind Querying

crobibero:
Fixes regression from https://github.com/jellyfin/jellyfin/pull/6979

The following enum values are unmapped:
```
BaseItemKind.ManualPlaylistsFolder
BaseItemKind.ChannelFolderItem
BaseItemKind.Program
BaseItemKind.Recording
```

Also fixes the naming of `_retrieveItemColumns`

### DIFF
--- a/Emby.Server.Implementations/Data/SqliteItemRepository.cs
+++ b/Emby.Server.Implementations/Data/SqliteItemRepository.cs
@@ -3611,15 +3611,32 @@ namespace Emby.Server.Implementations.Data
                         whereClauses.Add("type<>@type");
                         statement?.TryBind("@type", excludeTypeName);
                     }
+                    else
+                    {
+                        Logger.LogWarning("Undefined BaseItemKind to Type mapping: {BaseItemKind}", excludeTypes[0]);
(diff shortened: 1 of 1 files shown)


Bond-009 (approved):
Should we log a warning if `_baseItemKindNames` doesn't contain the key?

Emby.Server.Implementations/Data/SqliteItemRepository.cs
```diff
@@ -3629,15 +3646,32 @@ private List<string> GetWhereClauses(InternalItemsQuery query, IStatement statem
...
+                    }
+                    else
+                    {
+                        Logger.LogWarning("Undefined BaseItemKind to Type mapping: {BaseItemKind}", includeType);
+                    }
+                }
+
+                // Remove trailing comma.
+                whereBuilder.Length--;
+                whereClauses.Add($"type in ({whereBuilder})");
```
Bond-009, line 3674: Use the stringbuilder instead of string interpolation

cvium (approved):

PR #5823, opened 2021-04-15
Change repeated library names generator

danieladov:
In the current implementation, if a library with a repeated name is created, a 1 will be add at the end of the string (movies,movies1) but if another one is created another 1 is added (movies, movies1, and movies11).

With my implementation, the last number is incremented each time a new library with a repeated name is created (movies, movies1, and movies2). The counter will work up to 19 which I think is good enough.

PR #7013, opened 2021-12-18
Increment library name number instead of appending

1337joe:
**Changes**
When a new library is created with a name that already exists:
- if the last character is not a number: a " 2" is appended.
- if the last character is a number: that number is incremented by 1.

**Issues**

Supersedes #5823

### DIFF
--- a/Emby.Server.Implementations/Library/LibraryManager.cs
+++ b/Emby.Server.Implementations/Library/LibraryManager.cs
@@ -2890,11 +2890,12 @@ namespace Emby.Server.Implementations.Library
 
             var rootFolderPath = _configurationManager.ApplicationPaths.DefaultUserViewsPath;
 
+            var existingNameCount = 1; // first numbered name will be 2
             var virtualFolderPath = Path.Combine(rootFolderPath, name);
             while (Directory.Exists(virtualFolderPath))
             {
-                name += "1";
-                virtualFolderPath = Path.Combine(rootFolderPath, name);
+                existingNameCount++;
+                virtualFolderPath = Path.Combine(rootFolderPath, name + " " + existingNameCount);
             }
 
             var mediaPathInfos = options.PathInfos;


Emby.Server.Implementations/Library/LibraryManager.cs
```diff
@@ -2893,7 +2894,16 @@ public async Task AddVirtualFolder(string name, CollectionTypeOptions? collectio
             var virtualFolderPath = Path.Combine(rootFolderPath, name);
             while (Directory.Exists(virtualFolderPath))
```
cvium, line 2895: Performance isn't exactly a concern here, so we could just use [`Directory.GetDirectories`](https://docs.microsoft.com/en-us/dotnet/api/system.io.directory.getdirectories?view=net-6.0#System_IO_Directory_GetDirectories_System_String_System_String_System_IO_SearchOption_), with a wildcard pattern (eg. `name + "*"`) and count them?

1337joe: Just counting the matches doesn't account for libraries having been deleted:

Say I create three libraries: "Movies", "Movies 2", and "Movies 3", then delete "Movies 2". If I try to create another library named "Movies" then by count of "Movies*" I would expect to be creating "Movies 3" (which conflicts with an existing library) but by iterating through the next sequential name I find that I can create "Movies 2" instead.

cvium: That is true. You could simplify it though. Just keep a separate counter and concat it with name until the loop condition is unsatisfied.

1337joe: Yeah, I definitely over thought that originally.

cvium (approved):

crobibero (approved):

PR #7037, opened 2021-12-22
Fix resolved type for Trailers

1337joe:
The refactored extras resolving produces trailers of type `Video` instead of their specific type `Trailer`, this corrects that.

From the front-end everything seems to work as-is, but there are image providers and a metadata provider that support `Trailer` (but not `Video`) that don't currently run.

### DIFF
--- a/Emby.Server.Implementations/Library/Resolvers/Movies/MovieResolver.cs
+++ b/Emby.Server.Implementations/Library/Resolvers/Movies/MovieResolver.cs
@@ -132,7 +132,15 @@ namespace Emby.Server.Implementations.Library.Resolvers.Movies
             // Handle owned items
             if (args.Parent == null)
             {
-                return base.Resolve(args);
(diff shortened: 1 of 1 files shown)


crobibero (approved):

cvium:
What about series trailers?

1337joe:
`SeriesResolver` doesn't handle owned items. The `MovieResolver` actually is the first in `LibraryManager.ResolveItem` to produce a result so this change fixes both.

Should I duplicate the owned item code from `MovieResolver` to `SeriesResolver`?

cvium:
Neither of them should be handling "owned" items tbh

1337joe:
That does sound cleaner. I'll try to work on that tomorrow, unless you'd like to do it since you've been working closely with all of this recently.

1337joe:
Replaced with #7044 to fix this more generally

PR #7041, opened 2021-12-24
Fix System.OverflowException with ffprobe

Maeur1:
**Changes**
I was getting errors with some media files where the `CodecTimeBase` was set to `0/1`. This caused the `ffprobe` math to generate a divide by zero bug which then flowed onto a full error with the following stack trace: 

```
System.OverflowException: Value was either too large or too small for an Int32.
   at System.Convert.ToInt32(Double value)
   at System.Convert.ToInt32(Single value)
   at MediaBrowser.MediaEncoding.Probing.ProbeResultNormalizer.GetMediaStream(Boolean isAudio, MediaStreamInfo streamInfo, MediaFormatInfo formatInfo)
   at MediaBrowser.MediaEncoding.Probing.ProbeResultNormalizer.<>c__DisplayClass10_0.<GetMediaInfo>b__0(MediaStreamInfo s)
   at System.Linq.Enumerable.SelectListIterator`2.MoveNext()
   at System.Linq.Enumerable.WhereEnumerableIterator`1.ToList()
   at MediaBrowser.MediaEncoding.Probing.ProbeResultNormalizer.GetMediaInfo(InternalMediaInfoResult data, Nullable`1 videoType, Boolean isAudio, String path, MediaProtocol protocol)
   at MediaBrowser.MediaEncoding.Encoder.MediaEncoder.GetMediaInfoInternal(String inputPath, String primaryPath, MediaProtocol protocol, Boolean extractChapters, String probeSizeArgument, Boolean isAudio, Nullable`1 videoType, Boolean forceEnableLogging, CancellationToken cancellationToken)
   at MediaBrowser.Providers.MediaInfo.FFProbeVideoInfo.ProbeVideo[T](T item, MetadataRefreshOptions options, CancellationToken cancellationToken)
   at MediaBrowser.Providers.Manager.MetadataService`2.RunCustomProvider(ICustomMetadataProvider`1 provider, TItemType item, String logName, MetadataRefreshOptions options, RefreshResult refreshResult, CancellationToken cancellationToken)
```

If the dividend is also 0, we should return `null` instead since a framerate of 0 seems kinda off? 

OR an alternative could be to add some logic before the `1 / GetFrameRate(....) ?? 0`. Not sure what would be best but this fixed my issues with my library.

**Issues**
Sorry didnt really make an issue for this, I just found this bug on my own server

### DIFF
--- a/MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
+++ b/MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
@@ -1057,7 +1057,7 @@ namespace MediaBrowser.MediaEncoding.Probing
                 return null;
             }
 
-            return divisor == 0f ? null : dividend / divisor;
+            return divisor == 0f || dividend == 0f ? null : dividend / divisor;
         }
 
         private void SetAudioRuntimeTicks(InternalMediaInfoResult result, MediaInfo data)
--- a/tests/Jellyfin.MediaEncoding.Tests/Probing/ProbeResultNormalizerTests.cs
+++ b/tests/Jellyfin.MediaEncoding.Tests/Probing/ProbeResultNormalizerTests.cs
@@ -25,6 +25,7 @@ namespace Jellyfin.MediaEncoding.Tests.Probing
         [InlineData("120/1", 120f)]
         [InlineData("1704753000/71073479", 23.98578237601117f)]
         [InlineData("0/0", null)]
+        [InlineData("0/1", null)]
         [InlineData("1/1000", 0.001f)]
         [InlineData("1/90000", 1.1111111E-05f)]
         [InlineData("1/48000", 2.0833333E-05f)]


crobibero (approved):

LinFor:
I have the same issue with ffprobe, this PR works fine.

MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
```diff
@@ -1052,7 +1052,7 @@ private bool IsClose(double d1, double d2, double variance = .005)
                 return null;
             }
 
-            return divisor == 0f ? null : dividend / divisor;
+            return divisor == 0f || dividend == 0f ? null : dividend / divisor;
```
Bond-009, line 1055: I think this one should return `0`, @cvium ?

Bond-009 (changes requested):

MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
cvium: 0/999999 is still 0, so that wouldn't change anything. I'm not sure if this is the place to fix it

Maeur1: This is meant to fix the issue where there is logic: `1 / GetFrameRate(....) ?? 0` which turns into Inifnity when you go `1/0`

Bond-009: You are right, the problem isn't here, I've made a PR that hopefully fixes the issue
https://github.com/jellyfin/jellyfin/pull/7120

tests/Jellyfin.MediaEncoding.Tests/Probing/ProbeResultNormalizerTests.cs
```diff
@@ -25,6 +25,7 @@ public class ProbeResultNormalizerTests
         [InlineData("120/1", 120f)]
         [InlineData("1704753000/71073479", 23.98578237601117f)]
         [InlineData("0/0", null)]
+        [InlineData("0/1", null)]
```
Maeur1, line 28: Not gonna lie, this C# testing is kinda alien magic to me coming from Java but I guess this should add the test for the circumstance I am looking for.

Read a bit here if anyone else is as confused as me: https://docs.microsoft.com/en-us/dotnet/core/testing/unit-testing-with-dotnet-test

Maeur1:
Added a test for the exact circumstance I am trying to avoid.

MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
Maeur1: LGTM thanks

hasezoey:
Did this PR get merged? it just lists it as "closed"

1337joe:
It was closed in favor of #7120, which is tagged as "needs-testing".

If you're seeing the problem this fixed and can test that pull request and comment on it that would be helpful.

hasezoey:
> If you're seeing the problem this fixed and can test that pull request and comment on it that would be helpful.

just wanted to know why this was closed, also i had the "same" error, but after updating jellyfin (unstable) to todays latest, the error was gone (so i wanted to know if this was a result of this being merged)

Maeur1:
> It was closed in favor of #7120, which is tagged as "needs-testing".
> 
> If you're seeing the problem this fixed and can test that pull request and comment on it that would be helpful.

Opened up https://github.com/jellyfin/jellyfin/pull/7167 with some tests if that helps! Happy to make changes based on any guidance as I am not 100% familiar with C# guidelines/testing.

PR #6069, opened 2021-05-17
Added ConfigurationChanging event to plugin class.

BaronGreenback:
Added ConfigurationChanging that gets called just prior to plugin configuration saving.

This enables the configuration to be validated prior to saving, stopping the need for it to be validated and saved a second time.

PR #7045, opened 2021-12-24
Add IHasConfigurationValidation

crobibero:
Superceeds https://github.com/jellyfin/jellyfin/pull/6069

### DIFF
--- a/MediaBrowser.Common/Plugins/BasePluginOfT.cs
+++ b/MediaBrowser.Common/Plugins/BasePluginOfT.cs
@@ -153,23 +153,16 @@ namespace MediaBrowser.Common.Plugins
             }
         }
 
-        /// <summary>
-        /// Saves the current configuration to the file system.
-        /// </summary>
-        public virtual void SaveConfiguration()
-        {
-            SaveConfiguration(Configuration);
-        }
-
         /// <inheritdoc />
         public virtual void UpdateConfiguration(BasePluginConfiguration configuration)
         {
-            if (configuration == null)
-            {
-                throw new ArgumentNullException(nameof(configuration));
-            }
-
+            ArgumentNullException.ThrowIfNull(configuration, nameof(configuration));
             Configuration = (TConfigurationType)configuration;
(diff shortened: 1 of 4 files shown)


MediaBrowser.Common/Plugins/BasePluginOfT.cs
```diff
@@ -153,24 +153,16 @@ public virtual void SaveConfiguration(TConfigurationType config)
...
-        public virtual void SaveConfiguration()
-        {
-            SaveConfiguration(Configuration);
-        }
-
         /// <inheritdoc />
         public virtual void UpdateConfiguration(BasePluginConfiguration configuration)
         {
-            if (configuration == null)
+            Configuration = (TConfigurationType)configuration ?? throw new ArgumentNullException(nameof(configuration));
```
cvium, line 159: Use ArgumentNullException.ThrowIfNull and the cast will not return null but throw an exception

crobibero:
@jellyfin-bot rebase

MediaBrowser.Model/Plugins/IHasConfigurationValidation.cs
```diff
@@ -0,0 +1,15 @@
...
+/// Interface to indicate that the plugin has configuration validation.
+/// </summary>
+/// <typeparam name="TPluginConfiguration">The type of plugin configuration.</typeparam>
+public interface IHasConfigurationValidation<in TPluginConfiguration>
+    where TPluginConfiguration : BasePluginConfiguration
+{
+    /// <summary>
+    /// Validate the plugin configuration.
+    /// </summary>
+    /// <param name="pluginConfiguration">The plugin configuration.</param>
```
1337joe, line 13: Should the documentation include reference to `ConfigurationValidationException`, since that's what should be thrown on failure?

1337joe (approved):

crobibero:
This isn't needed, the plugin can just throw an exception when overriding `SaveConfiguration`

Issue #7050, opened 2021-12-25
[Issue]: Unable to determine image file extension from mime type image/jpg


nyanmisaka:
### Please describe your bug

Images with `.jpg` extension uploading failed after https://github.com/jellyfin/jellyfin/commit/7ee96a59d3b69e15aec1df33ad46f47021f8a20b

https://github.com/jellyfin/jellyfin/blob/076a13abeb344f7a79d81cce030799c8e77dbfaf/MediaBrowser.Providers/Manager/ImageSaver.cs#L368-L373

### Jellyfin Version

Other

### if other:

master

### Environment

```markdown
- OS: Linux and Windows
- Virtualization: None
- Clients: Web
- Browser: Chrome
```

### Jellyfin logs

```shell
[2021-12-25 18:01:18.111 +08:00] [ERR] [126] Jellyfin.Server.Middleware.ExceptionMiddleware: Error processing request. URL "POST" "/Items/6f9b09b629663dcd11f5a9ed4213b95c/Images/Primary".
System.ArgumentException: Unable to determine image file extension from mime type image/jpg
   at MediaBrowser.Providers.Manager.ImageSaver.GetStandardSavePath(BaseItem item, ImageType type, Nullable`1 imageIndex, String mimeType, Boolean saveLocally)
   at MediaBrowser.Providers.Manager.ImageSaver.SaveImage(BaseItem item, Stream source, String mimeType, ImageType type, Nullable`1 imageIndex, Nullable`1 saveLocallyWithMedia, CancellationToken cancellationToken)
   at Jellyfin.Api.Controllers.ImageController.SetItemImage(Guid itemId, ImageType imageType)
   at Jellyfin.Api.Controllers.ImageController.SetItemImage(Guid itemId, ImageType imageType)
   at lambda_method1584(Closure , Object )
   at Microsoft.AspNetCore.Mvc.Infrastructure.ActionMethodExecutor.TaskOfActionResultExecutor.Execute(IActionResultTypeMapper mapper, ObjectMethodExecutor executor, Object controller, Object[] arguments)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.<InvokeActionMethodAsync>g__Awaited|12_0(ControllerActionInvoker invoker, ValueTask`1 actionResultValueTask)
[...]

PR #7052, opened 2021-12-25
Add image/jpg to extension lookup

crobibero:
Fixes https://github.com/jellyfin/jellyfin/issues/7050
Fixes https://github.com/jellyfin/jellyfin/issues/7097

`image/jpg` may not be a valid mime-type, but it's required for our workflow


cvium (approved):

nyanmisaka (approved):

Bond-009:
I think this is an issue with the JS api client?

crobibero:
This won't be required in next release of jellyfin/jellyfin-apiclient-javascript

PR #7054, opened 2021-12-26
Add DisableEmbeddedSubtitles setting

matthew-jones-uk:
Option to disable embedded subs being added to metadata on library scan. This helps with setups that handle embedded subs badly.

Implements https://features.jellyfin.org/posts/985/hide-embedded-subtitles

### DIFF
--- a/MediaBrowser.Model/Configuration/EmbeddedSubtitleOptions.cs
+++ b/MediaBrowser.Model/Configuration/EmbeddedSubtitleOptions.cs
@@ -1,0 +1,30 @@
+namespace MediaBrowser.Model.Configuration
+{
+    /// <summary>
+    /// An enum representing the options to disable embedded subs.
+    /// </summary>
+    public enum EmbeddedSubtitleOptions
+    {
+
+        /// <summary>
+        /// Allow all embedded subs.
+        /// </summary>
+        AllowAll,
+
+        /// <summary>
+        /// Allow only embedded subs that are text based.
+        /// </summary>
+        AllowText,
+
+        /// <summary>
+        /// Allow only embedded subs that are image based.
+        /// </summary>
+        AllowImage,
+
+        /// <summary>
+        /// Disable all embedded subs.
+        /// </summary>
+        AllowNone,
+    }
+
+}
--- a/MediaBrowser.Model/Configuration/EmbeddedSubtitleOptions.cs
+++ b/MediaBrowser.Model/Configuration/EmbeddedSubtitleOptions.cs
@@ -9,22 +9,22 @@ namespace MediaBrowser.Model.Configuration
         /// <summary>
         /// Allow all embedded subs.
         /// </summary>
-        AllowAll,
+        AllowAll = 0,
 
         /// <summary>
         /// Allow only embedded subs that are text based.
         /// </summary>
-        AllowText,
+        AllowText = 1,
 
         /// <summary>
         /// Allow only embedded subs that are image based.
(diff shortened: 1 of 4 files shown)


matthew-jones-uk:
Not sure if the logger should be changed to something like debug instead of info?

MediaBrowser.Providers/MediaInfo/FFProbeVideoInfo.cs
```diff
@@ -229,6 +229,12 @@ public class FFProbeVideoInfo
                 video.Video3DFormat ??= mediaInfo.Video3DFormat;
             }
 
+            if (libraryOptions.DisableEmbeddedSubtitles)
+            {
+                _logger.LogDebug("Disabling embedded subtitles due to DisableEmbeddedSubtitles setting");
```
crobibero, line 234:
```suggestion
                _logger.LogDebug("Disabling embedded subtitles for {Path} due to DisableEmbeddedSubtitles setting", video.Path);
```

So we know which file the embedded subtitles are disabled for

crobibero (approved):
Works great!

I was able to show/hide the embedded subtitles by running "Search for missing metadata" or "Replace all metadata" on the library or item

cvium:
Wouldn't it make more sense to make it similar to the subtitle burn-in option?

1.	Allow all
2.	Allow only text
3.	Allow only image
4. None

Another thing is that I'm not sure disabling it in the prober is the correct place. We could do the filtering when subtitle requests are made.

matthew-jones-uk:
> Wouldn't it make more sense to make it similar to the subtitle burn-in option?
> 
> 1. Allow all
> 2. Allow only text
> 3. Allow only image
> 4. None
>
> Another thing is that I'm not sure disabling it in the prober is the correct place. We could do the filtering when subtitle requests are made.

The former seems something that would be easier to do within the prober, no? This implementation was aimed to be quite simple (and quick) to do, but I can add this. If not in the prober then where is the sub request handling done?

Also, if memory serves, the sub burn-in options are client side I think. This feature in its current form was aimed to prevent problems relating to subtitles failing (or being extremely slow) to load on the server side, hence being in the library options. What are your thoughts on this?

cvium:
We can keep it in the prober for now but you'd have to make sure the stream indices are intact.

The settings would remain largely the same. I'm just proposing a more granular setting. I get the impression that it is mostly PGS subtitles that people want to get rid of.

matthew-jones-uk:
The stream indices are intact in its current form I believe. I'll add the extra options to configure it.

matthew-jones-uk:
Not sure if I should've force-pushed there, but @cvium that should integrate the options you requested 🙂

1337joe (approved):

MediaBrowser.Model/Configuration/EmbeddedSubtitleOptions.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// <summary>
+    /// An enum representing the options to disable embedded subs.
+    /// </summary>
+    public enum EmbeddedSubtitleOptions
+    {
+
+        /// <summary>
+        /// Allow all embedded subs.
+        /// </summary>
+        AllowAll,
```
crobibero, line 12: enums should have defined values

matthew-jones-uk: Done!

crobibero (approved):

PR #7077, opened 2021-12-31
Playlist Carousel as Home Screen Option

sdoan99:
**Changes**
Current Options for Home Screen customization is My Media, Continue Watching, Next Up, and Latest Media

Adding options for custom playlists or collections would add to the tailored solution that jellyfin is

**Issues**
this feature would make jellyfin media server one step closer to being the best


crobibero:
I think this was accidentally opened as a PR instead of an issue. Either way, feature requests belong at https://features.jellyfin.org

sdoan99:
left a feature request. thanks crobibero

PR #7078, opened 2021-12-31
Improve metadata merge for identify

### LINKED ISSUE #7040: [Issue]: Identify does not clear tagline or tags
### Please describe your bug

Had a video that was incorrectly labeled/identified as a move with a similar name.
I clicked on Identify and chose the correct movie, checking to replace all metadata and images.

The resulting entry showed the tagline and tags from the old movie.
They had to be manually cleared.

Upon re-identifying a movie, should the tagline and tags (really everything) not be cleared?  Hence I had tags and tagline that is totally inappropriate to the movie itself (which is in a different genre).

Expected behaviour is that when replace all is selected, it should replace entries that the new movie has, but also clear those which are not included in the newly identified movie.

### Jellyfin Version

10.7.7

### if other:

_No response_

### Environment

```markdown
- OS: Debian 11
- Virtualization: Docker (Offial)
- Clients: Browser
- Browser: Chrome
- FFmpeg Version: N/A
- Playback Method: N/A
- Hardware Acceleration: N/A
- Plugins: N/A
- Reverse Proxy: traefik 2.0
- Base URL: /
- Networking: Bridge
- Storage: local
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

1337joe:
**Changes**
- Pull default `MergeData` implementation to abstract class
- Merge `ProviderUtils` into `MetadataService` base class and add tests
- Add flag to indicate metadata refresh is called by identify, in which case existing metadata (and images, if replacing them) will be discarded even if there is no new value to replace with
- General cleanup of warnings and tests on affected classes

**Issues**
Fixes #7040


MediaBrowser.Providers/Manager/ProviderUtils.cs
```diff
@@ -13,8 +11,21 @@
 
 namespace MediaBrowser.Providers.Manager
 {
+    /// <summary>
+    /// Class ProviderUtils.
+    /// </summary>
     public static class ProviderUtils
```
cvium, line 17: Does it make sense to keep this class now?

1337joe:
Converting to draft pull request: existing images aren't removed on re-identify like metadata now is, going to do a bit of rework to make that happen.

Will also see about merging `ProviderUtils` into `MetadataService` and keeping it testable.

MediaBrowser.Providers/Manager/ItemImageProvider.cs
```diff
@@ -61,6 +61,30 @@ public ItemImageProvider(ILogger logger, IProviderManager providerManager, IFile
...
+                var currentImage = item.GetImageInfo(_singularImages[i], 0);
+                if (currentImage != null)
+                {
+                    singular.Add(currentImage);
+                }
+            }
+
+            var oldBackdropImages = item.GetImages(ImageType.Backdrop).ToArray();
+            var toRemove = singular.Concat(oldBackdropImages).ToArray();
+            PruneImages(item, toRemove);
```
crobibero, line 83:
```suggestion
            singular.AddRange(item.GetImages(ImageType.Backdrop);
            PruneImages(item, singular);
```

and change PruneImages to be
```patch
Index: MediaBrowser.Providers/Manager/ItemImageProvider.cs
IDEA additional info:
<+>UTF-8
diff --git a/MediaBrowser.Providers/Manager/ItemImageProvider.cs b/MediaBrowser.Providers/Manager/ItemImageProvider.cs
--- a/MediaBrowser.Providers/Manager/ItemImageProvider.cs	(revision 6d299eed495549fd92402571577d732c74a73b37)
+++ b/MediaBrowser.Providers/Manager/ItemImageProvider.cs	(date 1641255052822)
@@ -340,9 +340,9 @@
             return options.IsEnabled(type);
         }
 
-        private void PruneImages(BaseItem item, ItemImageInfo[] images)
+        private void PruneImages(BaseItem item, IReadOnlyList<ItemImageInfo> images)
         {
-            for (var i = 0; i < images.Length; i++)
+            for (var i = 0; i < images.Count; i++)
             {
                 var image = images[i];
```

crobibero (approved):

MediaBrowser.Controller/Providers/ImageRefreshOptions.cs
```diff
@@ -27,6 +27,11 @@ public ImageRefreshOptions(IDirectoryService directoryService)
 
         public bool IsAutomated { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether the action is an identify.
+        /// </summary>
+        public bool IsIdentify { get; set; }
```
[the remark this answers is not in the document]
1337joe: `IsAutomated` just indicates that the refresh was triggered by the user (I think, it doesn't appear to do much from a brief scan). A normal manual metadata refresh, even with replace all metadata set, should not clear existing fields for which metadata is not found.

`IsIdentify` specifies that metadata refresh is the result of re-identifying the item, which should trigger a full wipe of downloaded metadata and images (because they were downloaded for the wrong id).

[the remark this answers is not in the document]
cvium: IsIdentify is maybe a bit too specific though. `ClearOldMetadata` or whatever seems like a better fit to me

MediaBrowser.Controller/Providers/ImageRefreshOptions.cs
```diff
@@ -27,6 +27,11 @@ public ImageRefreshOptions(IDirectoryService directoryService)
 
         public bool IsAutomated { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether old metadata should be removed if it isn't be replaced.
```
cvium, line 31:
```suggestion
        /// Gets or sets a value indicating whether old metadata should be removed if it isn't replaced.
```

cvium (approved):

MediaBrowser.Providers/Manager/MetadataService.cs
```diff
@@ -882,16 +883,311 @@ private bool HasChanged(BaseItem item, IHasItemChangeMonitor changeMonitor, IDir
...
+                var personInSource = source.FirstOrDefault(i => string.Equals(i.Name.RemoveDiacritics(), normalizedName, StringComparison.OrdinalIgnoreCase));
+
+                if (personInSource != null)
+                {
+                    foreach (var providerId in personInSource.ProviderIds)
+                    {
+                        if (!person.ProviderIds.ContainsKey(providerId.Key))
+                        {
+                            person.ProviderIds[providerId.Key] = providerId.Value;
+                        }
```
Bond-009, line 1126: Use `Dictionary.TryAdd` instead
```suggestion
                        person.ProviderIds.TryAdd(providerId.Key, providerId.Value);
```

MediaBrowser.Providers/Manager/MetadataService.cs
```diff
@@ -882,16 +883,311 @@ private bool HasChanged(BaseItem item, IHasItemChangeMonitor changeMonitor, IDir
...
+
+            foreach (var id in source.ProviderIds)
+            {
+                var key = id.Key;
+
+                // Don't replace existing Id's.
+                if (replaceData || !target.ProviderIds.ContainsKey(key))
+                {
+                    target.ProviderIds[key] = id.Value;
+                }
```
Bond-009, line 1077: Maybe something like this is better here? @cvium opinion?
```suggestion
                if (replaceData)
                {
                    target.ProviderIds[key] = id.Value;
                }
                else
                {
                    target.ProviderIds.TryAdd(key, id.Value);
                }
```

tests/Jellyfin.Providers.Tests/Manager/ItemImageProviderTests.cs
```diff
@@ -163,45 +153,51 @@ public void MergeImages_PopulatedItemWithGoodPathsAndSameNewImages_NoChange(Imag
             var itemImageProvider = GetItemImageProvider(null, fileSystem);
             var changed = itemImageProvider.MergeImages(item, images);
 
-            Assert.False(changed);
+            if (updateTime)
+            {
+                Assert.True(changed);
+                // before and after paths are the same, verify updated by size reset to 0
+                Assert.Equal(imageCount, item.GetImages(imageType).Count());
+                foreach (var image in item.GetImages(imageType))
```
Bond-009, line 161: Even in tests I don't like calling `item.GetImages(imageType)` twice

Bond-009 (changes requested):

Bond-009 (approved):

PR #7080, opened 2022-01-01
Allow ApiKey query parameter for authentication

crobibero:
**Changes**
Adds the ability to use `ApiKey` instead of `api_key` to authenticate when connecting the websocket.


nielsvanvelzen:
Would it be possible to use the code inside AuthorizationContext instead? That way we could support authentication via headers too.

Bond-009 (approved):

nielsvanvelzen (approved):

PR #7107, opened 2022-01-04
Speed up Jellyfin.Naming.Tests

Bond-009:
In my limited testing this saves ~4 seconds, from 19 to 15 seconds (in
        Xunit itself)


tests/Jellyfin.Naming.Tests/TV/EpisodePathParserTest.cs
```diff
@@ -50,8 +51,7 @@ public void ParseEpisodesCorrectly(string path, bool isDirectory, string name, i
         [InlineData("/test/01-03.avi", true, true)]
         public void EpisodePathParserTest_DifferentExpressionsParameters(string path, bool? isNamed, bool? isOptimistic)
         {
-            NamingOptions o = new NamingOptions();
-            EpisodePathParser p = new EpisodePathParser(o);
+            EpisodePathParser p = new EpisodePathParser(_namingOptions);
```
ilaborde, line 54: if we add a constructor and dispose, we can initialize the EpisodePathParser in the constructor and remove all the repetitive lines.
private readonly EpisodePathParser _episodePathParser;
`
public EpisodePathParserTest() {
 _episodePathParser = new EpisodeResolver(_namingOptions);
}`

But I'm not sure if that will improve the speed

Bond-009: `NamingOptions` is very expensive to create, this won't make much difference

crobibero (approved):

ilaborde (approved):

1337joe (approved):

PR #7118, opened 2022-01-05
Add null check for audio stream

crobibero:
Sometimes the audio stream isn't known, or doesn't exist...


1337joe (approved):

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -1236,7 +1236,7 @@ private void LogConditionFailure(DeviceProfile profile, string type, ProfileCond
                 return (false, TranscodeReason.ContainerBitrateExceedsLimit);
             }
 
-            if (audioStream.IsExternal)
+            if (audioStream != null && audioStream.IsExternal)
```
cvium, line 1239: Bond and I are starting to prefer `something?.Prop == true`

1337joe (approved):

PR #7148, opened 2022-01-09
Fix #7147: Don't return subtitles in mismatched format

### LINKED ISSUE #7147: [Issue]: API returns subtitles when requesting invalid formats
### Please describe your bug

http://localhost:8096/Videos/{itemid}/{mediasourceid}/Subtitles/{subtitleindex}/Stream.{format}

If API request for subtitles is made with valid options, except for format, JF currently returns the requested subtitle stream as Stream.{format}. Invalid {format} would be a format that JF can't convert to, or one that the subtitle file is not already in.

Example:
external subtitle format on disk. `Movie.en.srt`
subtitle stream request `Stream.notaformat`
currently returns a file named `Stream.notaformat` with contents of `Movie.en.srt`

This could cause issues because one format could be requested, and the return file could be a different format.
Example:
external subtitle format on disk. `Movie.en.srt`
subtitle stream request `Stream.ssa` - There is currently not a SSA subtitle writer in JF
would return a file named `Stream.ssa` with contents in SRT format.

Proposed solution:
Don't return result if can't convert, or return matching existing file.

### Jellyfin Version

10.7.7

### if other:

and 10.8.0

### Environment

```markdown
- OS: Windows
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

SenorSmartyPants:


SenorSmartyPants:
Bad PR

Issue #6672, opened 2021-10-07
Picking integrated subtitles displays "Fetching additional data" and then does nothing


vgambier:
**Describe the bug**
I'm playing a mkv file with dozens of integrated subtitles. They all show up in the list of available subtitles but they do not do anything. If I switch to "English" for example, it displays "Fetching additional data" and then nothing happens. The subtitles are working fine in VLC.

A similar version of this bug occured in earlier episodes of the same show, where it would take a long time for subtitles to show up - however, like I said, in some files they just never show up, no matter how long I wait.

A web search returned people having similar bugs in older versions, but it appeared that the issue was thought to eventually be fixed.

**System (please complete the following information):**
 - OS: Windows 10
 - Virtualization: None
 - Clients: Browser
 - Browser: Firefox 92.0.1 (64 bits)
 - Jellyfin Version: 10.7.7
 - FFmpeg Version: 4.3.2-Jellyfin
 - Playback: Sorry, I am not sure. Default?
 - Hardware Acceleration: Sorry, I am not sure. Default?
 - Installed Plugins: AudioDB10.7.7.0, MusicBrainz10.7.7.0, OMDb10.7.7.0, Open Subtitles11.0.0.0, TheTVDB6.0.0.0, Trakt12.0.0.0
 - Reverse Proxy: nginx
 - Base URL: yes: /jellyfin
 - Networking: Host
 - Storage: local (external hard drive)

Note: I briefly tried streaming from my Android phone and I got the same results. I can provide more information if necessary.

**To Reproduce**

1. Try to play a file that is affected by this issue
2. Choose any subtitles from the list
3. Wait
4. Notice no subtitles appear.

**Expected behavior**
The subtitles should show up.

**Server Logs**

```
[22:26:33] [WRN] [86] Jellyfin.Server.Middleware.ResponseTimeMiddleware: Slow HTTP Response from https://localhost/jellyfin/Sessions/Playing/Progress to 127.0.0.1 in 0:00:00.7078701 with Status Code 204
[22:28:17] [INF] [72] Emby.Server.Implementations.HttpServer.WebSocketManager: WS 127.0.0.1 closed
[22:28:20] [INF] [61] Emby.Server.Implementations.HttpServer.WebSocketManager: WS 127.0.0.1 request
[...]

PR #7153, opened 2022-01-10
Increase timeout for subtitle extraction to 30min

vgambier:
**Changes**
Increase timeout value for subtitle extraction operation from 5 minutes to 30 minutes. In some cases, ffmpeg takes over 5 minutes to extract the subtitle even though nothing is wrong with the file and the extraction is possible.

Arguably the fact that some video files take this long might indicate that this extraction step should be (optionally?) done in advance, otherwise there will be 5 to 30 minutes of the file playing without subtitles.

**Issues**
Addresses #6672

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -151,6 +151,7 @@
  - [peterspenler](https://github.com/peterspenler)
  - [MBR-0001](https://github.com/MBR-0001)
  - [jonas-resch](https://github.com/jonas-resch)
+ - [vgambier](https://github.com/vgambier)
 
 # Emby Contributors
 
--- a/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
+++ b/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
@@ -440,7 +440,7 @@ namespace MediaBrowser.MediaEncoding.Subtitles
                     throw;
                 }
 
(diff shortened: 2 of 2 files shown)


CONTRIBUTORS.md
```diff
@@ -217,3 +217,4 @@
  - [olsh](https://github.com/olsh)
  - [lbenini](https://github.com/lbenini)
  - [gnuyent](https://github.com/gnuyent)
+ - [vgambier](https://github.com/vgambier)
```
1337joe, line 220: This section is contributors to the Emby code before Jellyfin was forked off, move this to above the "#Emby Contributors" line.

vgambier: Apologies. I force-pushed the change,

1337joe (approved):

crobibero (approved):

Bond-009 (approved):

Qiler:
This does not solve the issue it was attempting to solve because there is an additional 5min timeout [at this line](https://github.com/jellyfin/jellyfin/blob/3d38b92b6ecf645f36e192f354ad038f6891296a/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs#L588).
I'm still getting timeout after 5 minutes. @vgambier

RealGreenDragon:
Thanks @vgambier for this PR, and thanks @Qiler to good observation.
I created a new PR to complete the change.

RealGreenDragon:
I missed to back-port my PR also in 10.8.z branch, so I opened a new PR just now.

PR #7167, opened 2022-01-13
ProbeResultNormalizer: fix framerate compare + tests

Maeur1:
**Changes**
The same fixes from https://github.com/jellyfin/jellyfin/pull/7120, but instead this includes some tests to ensure that the logic is correct. Was unsure on how I could push some commits to the existing branch so I made a new PR completely.

I could see why we wanted some tests, as the math conversion did seem quite complex, but after some testing it looks to be functioning correctly.

I broke out the method to make it testable, as the main method this logic sat in would require quite a lot of mocking which I am not too comfortable doing just yet, this stuff is still kinda black magic to me as I haven't ever written anything in C#.

**Issues**
No issues here, just fixing some ffprobe errors I was having previously from https://github.com/jellyfin/jellyfin/pull/7041

### DIFF
--- a/MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
+++ b/MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
@@ -611,6 +611,17 @@ namespace MediaBrowser.MediaEncoding.Probing
         }
 
         /// <summary>
+        /// Determines whether a stream code time base is double the frame rate.
+        /// </summary>
+        /// <param name="averageFrameRate">average frame rate.</param>
+        /// <param name="codecTimeBase">codec time base string.</param>
+        /// <returns>true if the codec time base is double the frame rate.</returns>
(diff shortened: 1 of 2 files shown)


MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
```diff
@@ -720,17 +731,16 @@ private MediaStream GetMediaStream(bool isAudio, MediaStreamInfo streamInfo, Med
                 stream.AverageFrameRate = GetFrameRate(streamInfo.AverageFrameRate);
                 stream.RealFrameRate = GetFrameRate(streamInfo.RFrameRate);
 
+                bool videoInterlaced = !string.IsNullOrWhiteSpace(streamInfo.FieldOrder)
+                    && !string.Equals(streamInfo.FieldOrder, "progressive", StringComparison.OrdinalIgnoreCase);
+
```
Maeur1, line 736: Re-ordered this method so that the below comment makes sense as to how `h264MbaffCoded` is formulated

Bond-009 (approved):

PR #7170, opened 2022-01-13
Fix RPM build to not use dotnet 6.0

brianjmurrell:
**Changes**
Now that dotnet 6.0 is out, the RPM build needs to prevent dotnet-host-6.0 from being installed by the BR.

Knowing this can happen, this probably also needs to be forward ported to master to prevent this same kind of issue when dotnet 7.0 becomes available.

### DIFF
--- a/fedora/jellyfin.spec
+++ b/fedora/jellyfin.spec
@@ -24,10 +24,13 @@ Source16:       jellyfin-firewalld.xml
 %{?systemd_requires}
 BuildRequires:  systemd
 BuildRequires:  libcurl-devel, fontconfig-devel, freetype-devel, openssl-devel, glibc-devel, libicu-devel
-# Requirements not packaged in main repos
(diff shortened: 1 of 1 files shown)


crobibero:
Why is this needed? We are not releasing anymore 10.7 versions

brianjmurrell:
> Why is this needed?

Because with dotnet 6.0 in the MS repo now, this _latest stable release_ no longer builds RPMs successfully.  If somebody checks out this _stable branch_ and tries to build RPMs for EL7, it will now fail, due to the recent changes in the MS dotnet repo.

>We are not releasing anymore 10.7 versions

Fair enough.  As you wish.  This PR at least helps somebody if they do try to build a stable release RPM for EL7.

nielsvanvelzen:
Closing this PR as it's targeting an unsupported release branch.

PR #7176, opened 2022-01-14
Fix trailers duplication

dmitrylyzo:
**Changes**
- Fix duplication (same URL) of remote trailers
- Fix duplication of local trailers

**Issues**
- Wrong iterating over remote trailers causes duplication of remote trailers.
- `/LocalTrailers` endpoint returns doubled local trailers.
- `.LocalTrailerCount` is doubled as well.

### DIFF
--- a/Emby.Server.Implementations/Dto/DtoService.cs
+++ b/Emby.Server.Implementations/Dto/DtoService.cs
@@ -1102,12 +1102,14 @@ namespace Emby.Server.Implementations.Dto
 
             if (options.ContainsField(ItemFields.LocalTrailerCount))
             {
-                allExtras ??= item.GetExtras().ToArray();
-                dto.LocalTrailerCount = allExtras.Count(i => i.ExtraType == ExtraType.Trailer);
-
                 if (item is IHasTrailers hasTrailers)
                 {
-                    dto.LocalTrailerCount += hasTrailers.GetTrailerCount();
+                    dto.LocalTrailerCount = hasTrailers.GetTrailerCount();
+                }
+                else
+                {
+                    allExtras ??= item.GetExtras().ToArray();
+                    dto.LocalTrailerCount = allExtras.Count(i => i.ExtraType == ExtraType.Trailer);
                 }
             }
 
--- a/Emby.Server.Implementations/Dto/DtoService.cs
+++ b/Emby.Server.Implementations/Dto/DtoService.cs
@@ -1108,8 +1108,7 @@ namespace Emby.Server.Implementations.Dto
                 }
                 else
                 {
-                    allExtras ??= item.GetExtras().ToArray();
-                    dto.LocalTrailerCount = allExtras.Count(i => i.ExtraType == ExtraType.Trailer);
+                    dto.LocalTrailerCount = (allExtras ?? item.GetExtras()).Count(i => i.ExtraType == ExtraType.Trailer);
                 }
             }
 
--- a/Jellyfin.Api/Controllers/UserLibraryController.cs
+++ b/Jellyfin.Api/Controllers/UserLibraryController.cs
@@ -206,21 +206,19 @@ namespace Jellyfin.Api.Controllers
                 : _libraryManager.GetItemById(itemId);
 
             var dtoOptions = new DtoOptions().AddClientFields(Request);
-            var dtosExtras = item.GetExtras(new[] { ExtraType.Trailer })
-                .Select(i => _dtoService.GetBaseItemDto(i, dtoOptions, user, item))
-                .ToArray();
 
             if (item is IHasTrailers hasTrailers)
             {
                 var trailers = hasTrailers.LocalTrailers;
                 var dtosTrailers = _dtoService.GetBaseItemDtos(trailers, dtoOptions, user, item);
-                var allTrailers = new BaseItemDto[dtosExtras.Length + dtosTrailers.Count];
-                dtosExtras.CopyTo(allTrailers, 0);
-                dtosTrailers.CopyTo(allTrailers, dtosExtras.Length);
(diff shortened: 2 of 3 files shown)


Emby.Server.Implementations/Dto/DtoService.cs
```diff
@@ -1102,12 +1102,14 @@ private void AttachBasicFields(BaseItemDto dto, BaseItem item, BaseItem owner, D
...
-
                 if (item is IHasTrailers hasTrailers)
                 {
-                    dto.LocalTrailerCount += hasTrailers.GetTrailerCount();
+                    dto.LocalTrailerCount = hasTrailers.GetTrailerCount();
+                }
+                else
+                {
+                    allExtras ??= item.GetExtras().ToArray();
+                    dto.LocalTrailerCount = allExtras.Count(i => i.ExtraType == ExtraType.Trailer);
```
cvium, line 1112:
```suggestion
                    dto.LocalTrailerCount = (allExtras ?? item.GetExtras()).Count(i => i.ExtraType == ExtraType.Trailer);
```

dmitrylyzo: If we are sure we won't use `allExtras` (as a cache?) below in the future. Otherwise we have to touch this line again.

cvium: But this is now :)

dmitrylyzo: Isn't it like:
```cs
if (!allExtras) allExtras = item.GetExtras().ToArray();
```
We could remove `.ToArray()` though (from both lines).

cvium: ?? is the null coalescing operator. It will return the left hand side if it's not null, otherwise it returns the right hand side.

We can't remove the toarray since it will enumerate twice if both ifs are entered.

crobibero (approved):

Jellyfin.Api/Controllers/UserLibraryController.cs
```diff
@@ -206,21 +206,19 @@ public ActionResult<IEnumerable<BaseItemDto>> GetLocalTrailers([FromRoute, Requi
...
             if (item is IHasTrailers hasTrailers)
             {
                 var trailers = hasTrailers.LocalTrailers;
                 var dtosTrailers = _dtoService.GetBaseItemDtos(trailers, dtoOptions, user, item);
-                var allTrailers = new BaseItemDto[dtosExtras.Length + dtosTrailers.Count];
-                dtosExtras.CopyTo(allTrailers, 0);
-                dtosTrailers.CopyTo(allTrailers, dtosExtras.Length);
+                var allTrailers = new BaseItemDto[dtosTrailers.Count];
+                dtosTrailers.CopyTo(allTrailers, 0);
                 return allTrailers;
```
cvium, line 216:
```suggestion
```

Jellyfin.Api/Controllers/UserLibraryController.cs
```diff
@@ -206,21 +206,19 @@ public ActionResult<IEnumerable<BaseItemDto>> GetLocalTrailers([FromRoute, Requi
...
 
             var dtoOptions = new DtoOptions().AddClientFields(Request);
-            var dtosExtras = item.GetExtras(new[] { ExtraType.Trailer })
-                .Select(i => _dtoService.GetBaseItemDto(i, dtoOptions, user, item))
-                .ToArray();
 
             if (item is IHasTrailers hasTrailers)
             {
                 var trailers = hasTrailers.LocalTrailers;
                 var dtosTrailers = _dtoService.GetBaseItemDtos(trailers, dtoOptions, user, item);
```
cvium, line 213:
```suggestion
                return Ok(_dtoService.GetBaseItemDtos(trailers, dtoOptions, user, item));
```

Jellyfin.Api/Controllers/UserLibraryController.cs
```diff
@@ -206,21 +206,19 @@ public ActionResult<IEnumerable<BaseItemDto>> GetLocalTrailers([FromRoute, Requi
...
-                dtosTrailers.CopyTo(allTrailers, dtosExtras.Length);
+                var allTrailers = new BaseItemDto[dtosTrailers.Count];
+                dtosTrailers.CopyTo(allTrailers, 0);
                 return allTrailers;
             }
 
-            return dtosExtras;
+            return item.GetExtras(new[] { ExtraType.Trailer })
+                .Select(i => _dtoService.GetBaseItemDto(i, dtoOptions, user, item))
+                .ToArray();
```
cvium, line 221:
```suggestion
            return Ok(item.GetExtras()
                .Where(e => e.ExtraType == ExtraType.Trailer)
                .Select(i => _dtoService.GetBaseItemDto(i, dtoOptions, user, item)));
```

dmitrylyzo: Btw, the `GetExtras` overload used previously doesn't sort.
https://github.com/jellyfin/jellyfin/blob/7500c2b28b4603965461b7bce37413c0a53ff2d0/MediaBrowser.Controller/Entities/BaseItem.cs#L2634-L2640

cvium: Do you think this could be an issue?

dmitrylyzo: I don't think so. This is a question of consistency.
Moreover, perhaps we don't need to sort trailers at all, or even we need to randomize them.
Speaking of minimizing the number of operation, we should probably sort where we call `GetExtras` if necessary.

cvium (approved):

Bond-009 (approved):

PR #7184, opened 2022-01-15
Upgrade Docker image to use up-to-date Mesa

GaaH:
This PR upgrade the official Docker image to Debian bullseye in order to use an up-to-date version of Mesa.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,10 +7,11 @@ ARG DOTNET_VERSION=6.0
 FROM node:lts-alpine as web-builder
 ARG JELLYFIN_WEB_VERSION=master
 RUN apk add curl git zlib zlib-dev autoconf g++ make libpng-dev gifsicle alpine-sdk automake libtool make gcc musl-dev nasm python3 \
- && curl -L https://github.com/jellyfin/jellyfin-web/archive/${JELLYFIN_WEB_VERSION}.tar.gz | tar zxf - \
- && cd jellyfin-web-* \
- && npm ci --no-audit --unsafe-perm \
- && mv dist /dist
+  && curl -L https://github.com/jellyfin/jellyfin-web/archive/${JELLYFIN_WEB_VERSION}.tar.gz | tar zxf - \
+  && cd jellyfin-web-* \
+  && npx browserslist@latest --update-db \
+  && npm ci --no-audit --unsafe-perm \
+  && mv dist /dist
 
 FROM debian:stable-slim as app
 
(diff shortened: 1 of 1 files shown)


hogenf:
Are we not on bullseye with "FROM debian:stable-slim as app" all ready?
Why is "npx browserslist@latest --update-db" added?

This seams to be more about up-to-date version of Mesa than upgrade to bullseye.

GaaH:
> Are we not on bullseye with "FROM debian:stable-slim as app" all ready? 

Yes, you should be right, but the Docker image seems to be old and is still on buster. But re-building the image should be sufficient to upgrade to bullseye, I guess.

> Why is "npx browserslist@latest --update-db" added?

NPM was issuing a warning about the database not being up-to-date and suggests to do [this](https://github.com/browserslist/browserslist#browsers-data-updating).

> This seams to be more about up-to-date version of Mesa than upgrade to bullseye.

Indeed, sorry for the misleading title. I named the pull-request like this because of what is written in the [Hardware acceleration page from the documentation](https://jellyfin.org/docs/general/administration/hardware-acceleration.html#va-api): "As of 10.7.1, the Docker image uses Debian 10 and thus Mesa 18.3.6. To upgrade it, you'll need to make your own image based on a more recent distribution.".

crobibero:
If approved these changes.will need to be mirrored in https://github.com/jellyfin/jellyfin-metapackages

Shadowghost:
It would be great if this PR could wait for a result on https://github.com/jellyfin/jellyfin-build/issues/12 so we don't need to touch the Dockerfiles twice.

crobibero:
This PR is no longer blocked, please update this PR to remove merge conflicts.

GaaH:
I’m not sure this PR is required anymore (nor if it was required back then). As stated by @hogenf, the Docker image should be using an up-to-date mesa since it already relies on debian stable, aka bullseye. However, I do not understand why this is not reflected in the official Docker image. I guess that one should use the _unstable_ image to use the latest Dockerfile?

Furthermore, I do not understand the relationship between this repo and https://github.com/jellyfin/jellyfin-metapackages. Can  someone point me to some documentation?

Bond-009:
Closing as it's not needed

> I guess that one should use the unstable image to use the latest Dockerfile?

Correct
> Can someone point me to some documentation?

https://github.com/jellyfin/jellyfin-metapackages/blob/master/README.md
This is a simple set of Docker images that combine the [jellyfin-server](https://github.com/jellyfin/jellyfin) and [jellyfin-web](https://github.com/jellyfin/jellyfin-web) Docker images into one final jellyfin image for distribution.

PR #6454, opened 2021-08-20
Allow for partial transcoding to a DirectPlayProfile

GodTamIt:
**Changes**
This change allows for partial transcodes to be prioritized over full transcodes when a video stream can be decoded by the target client. In an instance where a video stream is fully compliant with the specifications of the client but transcoding is still required due to container being unsupported or other streams needing transcoding, this is where this change will prefer to target that direct profile instead of a full transcode to a `TranscodeProfile`.

**Note:** this change currently requires the resulting protocol to be HTTP progressive download. There is a `TODO` comment left there to detect whether the client may support something like HLS or DASH. I will say a codec-agnostic adaptive format like DASH seems kinda nice here.

Currently testing on Firefox, Chrome, Android TV, and Android.

**Issues**
Fixes #6149 
Fixes #6450
Fixes #7295

PR #7198, opened 2022-01-16
Add TranscodingProfile conditions

dmitrylyzo:
The idea is to add some control over the transcoding profile separately from the direct profiles.
For example, no resolution limit for DirectPlay, but no more screen size for transcoding.
Ideally, we need global conditions as well.

**Changes**
Add TranscodingProfile conditions

### DIFF
--- a/MediaBrowser.Model/Dlna/StreamBuilder.cs
+++ b/MediaBrowser.Model/Dlna/StreamBuilder.cs
@@ -699,6 +699,7 @@ namespace MediaBrowser.Model.Dlna
                     if (playlistItem.PlayMethod != PlayMethod.DirectPlay)
                     {
                         playlistItem.PlayMethod = PlayMethod.Transcode;
+                        ApplyTranscodingConditions(playlistItem, transcodingProfile.Conditions, null, true, true);
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


crobibero (approved):

Shadowghost:
Does this also work as intended with #7325 instead of #6454?

dmitrylyzo:
> Does this also work as intended with #7325 instead of #6454?

I haven't tested with them at all.
Now I tried with #7325 - it seems to work (but I need to move the call).

crobibero (approved):

dmitrylyzo:
Hmm, this breaks remuxing (it is also `Transcode`) :thinking:

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -710,6 +710,11 @@ private StreamInfo BuildVideoItem(MediaSourceInfo item, VideoOptions options)
                     if (playlistItem.PlayMethod != PlayMethod.DirectPlay)
                     {
                         playlistItem.PlayMethod = PlayMethod.Transcode;
+
+                        if ((playlistItem.TranscodeReasons & (VideoReasons | TranscodeReason.ContainerBitrateExceedsLimit)) != 0)
```
dmitrylyzo, line 714: Or maybe `VideoReasons' is enough (since that means using another container)?

joshuaboniface (approved):

PR #7199, opened 2022-01-16
Add 'IsHdr' codec condition

dmitrylyzo:
_This is my old PR. I haven't tested it since that day._

The idea is that the client can limit the color range - block HDR even if it is supported.

**Changes**
Add 'IsHdr' codec condition

_I think it would be better to replace `IsHdr` with a string of video ranges._


whiteowl3:
Is this still alive?  Has it been superceded?

dmitrylyzo:
Superseded by #7947

PR #7202, opened 2022-01-16
Add option to change unix socket permissions

### LINKED ISSUE #5311: Unix socket permissions
Listening on Unix socket is great feature, but without ability to specify permissions on that socket it is quite unusable for common setup.
I hope this qualifies as bug.

knackebrot:
Using P/Invoke might look scary, but since we're limited to unices and libc, this should work pretty reliably.

**Changes**

When new configuration parameter `kestrel:socketPermissions` is set (e.g. `0777`), try to change the unix socket's permissions to its value.

**Issues**

Fixes #5311

### DIFF
--- a/Jellyfin.Server/Program.cs
+++ b/Jellyfin.Server/Program.cs
@@ -5,6 +5,7 @@ using System.IO;
 using System.Linq;
 using System.Net;
 using System.Reflection;
+using System.Runtime.InteropServices;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
@@ -198,6 +199,13 @@ namespace Jellyfin.Server
                 try
                 {
                     await webHost.StartAsync(_tokenSource.Token).ConfigureAwait(false);
+
+                    if (startupConfig.UseUnixSocket() && Environment.OSVersion.Platform == PlatformID.Unix)
+                    {
+                        var socketPath = GetUnixSocketPath(startupConfig, appPaths);
+
+                        SetUnixSocketPermissions(startupConfig, socketPath);
+                    }
                 }
                 catch (Exception ex) when (ex is not TaskCanceledException)
                 {
@@ -327,20 +335,7 @@ namespace Jellyfin.Server
                     // Bind to unix socket (only on unix systems)
                     if (startupConfig.UseUnixSocket() && Environment.OSVersion.Platform == PlatformID.Unix)
                     {
-                        var socketPath = startupConfig.GetUnixSocketPath();
-                        if (string.IsNullOrEmpty(socketPath))
-                        {
-                            var xdgRuntimeDir = Environment.GetEnvironmentVariable("XDG_RUNTIME_DIR");
-                            if (xdgRuntimeDir == null)
-                            {
-                                // Fall back to config dir
-                                socketPath = Path.Join(appPaths.ConfigurationDirectoryPath, "socket.sock");
-                            }
-                            else
-                            {
-                                socketPath = Path.Join(xdgRuntimeDir, "jellyfin-socket");
-                            }
-                        }
+                        var socketPath = GetUnixSocketPath(startupConfig, appPaths);
 
                         // Workaround for https://github.com/aspnet/AspNetCore/issues/14134
                         if (File.Exists(socketPath))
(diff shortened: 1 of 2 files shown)


Jellyfin.Server/Program.cs
```diff
@@ -657,5 +674,8 @@ private static string NormalizeCommandLineArgument(string arg)
 
             return "\"" + arg + "\"";
         }
+
+        [DllImport("libc")]
+        private static extern int chmod(string pathname, int mode);
```
crobibero, line 679: This can be moved to L352 to bring it withing the `Unix` block

knackebrot: Moved to L355 since it's only being used there.

crobibero (approved):

Ta180m:
I'm getting the error
```
[INF] [1] Main: Kestrel listening on Any IP4 Address
[INF] [1] Main: Kestrel listening to unix socket /run/jellyfin/jellyfin.sock
[ERR] [1] Main: Failed to set unix socket permissions to 0770, exit code: -1
```
when running Jellyfin patched with this PR.

I've configured the Unix socket like this:
```
cat /etc/systemd/system/jellyfin.service.d/override.conf 
[Service]
RuntimeDirectory=jellyfin
Environment=JELLYFIN_kestrel__socket=true
Environment=JELLYFIN_kestrel__socketPath=/tmp/jellyfin.sock
Environment=JELLYFIN_kestrel__socketPermissions=0770
```

knackebrot:
> I'm getting the error
> 
> ```
> [INF] [1] Main: Kestrel listening on Any IP4 Address
> [INF] [1] Main: Kestrel listening to unix socket /run/jellyfin/jellyfin.sock
> [ERR] [1] Main: Failed to set unix socket permissions to 0770, exit code: -1
> ```
> 
> when running Jellyfin patched with this PR.
> 
> I've configured the Unix socket like this:
> 
> ```
> cat /etc/systemd/system/jellyfin.service.d/override.conf 
> [Service]
> RuntimeDirectory=jellyfin
> Environment=JELLYFIN_kestrel__socket=true
> Environment=JELLYFIN_kestrel__socketPath=/run/jellyfin/jellyfin.sock
> Environment=JELLYFIN_kestrel__socketPermissions=0770
> ```

Yeah, that was my fault, permissions were being set before the socket was created. Works fine for me after latest push:

`[20:56:35] [INF] Kestrel listening on Any IP4 Address
[20:56:35] [INF] Kestrel listening to unix socket /run/jellyfin/jellyfin.sock
[20:56:35] [INF] Unix socket permissions set to 0777
`

`ls -lah /run/jellyfin/
srwxrwxrwx  1 jellyfin jellyfin    0 jan 22 20:56 jellyfin.sock
`

Ta180m:
@knackebrot Thanks, the latest commit works for me as well!

Jellyfin.Server/Program.cs
```diff
@@ -657,5 +652,48 @@ private static string NormalizeCommandLineArgument(string arg)
...
+            {
+                var xdgRuntimeDir = Environment.GetEnvironmentVariable("XDG_RUNTIME_DIR");
+                if (xdgRuntimeDir == null)
+                {
+                    // Fall back to config dir
+                    socketPath = Path.Join(appPaths.ConfigurationDirectoryPath, "socket.sock");
+                }
+                else
+                {
+                    socketPath = Path.Join(xdgRuntimeDir, "jellyfin-socket");
```
crobibero, line 670: Can we have the socket name be `jellyfin.socket` in both cases?

knackebrot: if we're willing to break the previous behavior, I'm for changing it to `jellyfin.sock`. It's similar to `mysqld.sock`, `php7.4-fpm.sock` etc.

knackebrot:
I've changed the default unix socket name to `jellyfin.sock` as discussed

crobibero (approved):

Jellyfin.Server/Program.cs
```diff
@@ -664,5 +659,49 @@ private static string NormalizeCommandLineArgument(string arg)
...
+            return socketPath;
+        }
+
+        private static void SetUnixSocketPermissions(IConfiguration startupConfig, string socketPath)
+        {
+            var socketPerms = startupConfig.GetUnixSocketPermissions();
+
+            if (!string.IsNullOrEmpty(socketPerms))
+            {
+                [DllImport("libc")]
```
cvium, line 691: I think a dll import for a one-off during startup is sub-optimal. Can't we just call chmod directly?

joshuaboniface: Is the chmod cross-platform? I'm not sure if any of this is, but that could be why? @knackebrot

knackebrot: I thought P/Invoke would be more efficient than spawning a new process and a safer bet, since .net runtime is linked against libc. I don't know much about how P/Invoke works internally, can you please explain how is it sub-optimal?

joshuaboniface (approved):
With no additional comments, I think we should keep this as is and go with it. If it needs to change later so be it.

Jellyfin.Server/Program.cs
```diff
@@ -689,9 +689,9 @@ private static void SetUnixSocketPermissions(IConfiguration startupConfig, strin
             if (!string.IsNullOrEmpty(socketPerms))
             {
                 [DllImport("libc")]
-                static extern int chmod(string pathname, int mode);
+                static extern int Chmod(string pathname, int mode);
 
-                var exitCode = chmod(socketPath, Convert.ToInt32(socketPerms, 8));
+                var exitCode = Chmod(socketPath, Convert.ToInt32(socketPerms, 8));
```
knackebrot, line 695: is this going to work without the DllImport Entrypoint set to "chmod"?

crobibero: Good call, I've never used DllImport so it's very possible I broke it.
Here's a PR to fix the entrypoint: https://github.com/jellyfin/jellyfin/pull/7495

Jellyfin.Server/Program.cs
```diff
@@ -664,5 +659,49 @@ private static string NormalizeCommandLineArgument(string arg)
...
+            if (!string.IsNullOrEmpty(socketPerms))
+            {
+                [DllImport("libc")]
+                static extern int Chmod(string pathname, int mode);
+
+                var exitCode = Chmod(socketPath, Convert.ToInt32(socketPerms, 8));
+
+                if (exitCode < 0)
+                {
+                    _logger.LogError("Failed to set Kestrel unix socket permissions to {SocketPerms}, return code: {ExitCode}", socketPerms, exitCode);
```
kevincox, line 698: Shouldn't this raise an exception and cause the program yo fsil to startup?

Issue #6951, opened 2021-12-06
[Issue]: Grouped TV Show Library Folders thumb image never updates on home page


SenorSmartyPants:
### Please describe your bug

I have two tv show libraries, one for the DVR recorded shows, and one for other tv show videos. I set my home page to group these  library folders together so I see one "TV Shows" thumbnail image under "My Media" on the home page. Periodically JF will update the thumbnail images for Movies and Collections, but the combined tv show folder thumbnail never updates. If I ungroup these libraries I can see that the thumbnail for the individual tv show libraries have been updated. I can re-enable grouping and still see the same TV Shows thumbnail image. 

Combined library thumbnail images should be treated the same as individual library thumbnails. If individuals are updated, combined ones should be updated at the same time.

### Jellyfin Version

10.7.7

### Environment

```markdown
- OS: armbian/debian
- Virtualization: LSIO docker
- Clients: browser
- Browser: chrome 96
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

SenorSmartyPants:
This was successfully fixed by #7221 and deployed in JF 10.8.0-beta1

PR #7221, opened 2022-01-20
UserView Thumbnail fixes

SenorSmartyPants:
**Changes**

- UserView thumbnail (like when you group multiple libraries of the same type together on the homepage) was using poster images, not backdrop. 
- Image selected would never change, shuffle available items with images.


SenorSmartyPants:
Fixes #6951

crobibero (approved):

1337joe (approved):

PR #7233, opened 2022-01-22
Fix some warnings

Bond-009:


MediaBrowser.Model/Dto/DisplayPreferencesDto.cs
```diff
@@ -63,10 +63,10 @@ public DisplayPreferencesDto()
         public int PrimaryImageWidth { get; set; }
 
         /// <summary>
-        /// Gets or sets the custom prefs.
+        /// Gets the custom prefs.
         /// </summary>
         /// <value>The custom prefs.</value>
-        public Dictionary<string, string?> CustomPrefs { get; set; }
+        public Dictionary<string, string?> CustomPrefs { get; }
```
crobibero, line 69: I think this will break setting display preferences

Bond-009: You're right

crobibero (approved):

Emby.Server.Implementations/Library/MediaStreamSelector.cs
```diff
@@ -13,39 +11,24 @@ namespace Emby.Server.Implementations.Library
...
-                    return defaultStream.Index;
-                }
-            }
-
-            var stream = streams.FirstOrDefault();
-
-            if (stream != null)
-            {
-                return stream.Index;
+                return sortedStreams.FirstOrDefault(i => i.IsDefault)?.Index;
```
1337joe, line 20: Old behavior: if there is no default stream it falls through and returns the first stream
New behavior: if there is no default stream it returns null, even if there is a non-default stream

Was this intended?

Bond-009: Nice catch, fixed and added test for this case.

1337joe (approved):

Emby.Server.Implementations/Library/MediaStreamSelector.cs
```diff
@@ -95,26 +84,27 @@ public static class MediaStreamSelector
             return stream?.Index;
         }
 
-        private static IEnumerable<MediaStream> GetSortedStreams(IEnumerable<MediaStream> streams, MediaStreamType type, string[] languagePreferences)
+        private static IEnumerable<MediaStream> GetSortedStreams(IEnumerable<MediaStream> streams, MediaStreamType type, IReadOnlyList<string> languagePreferences)
         {
             // Give some preference to external text subs for better performance
-            return streams.Where(i => i.Type == type)
+            return streams
+            .Where(i => i.Type == type)
```
crobibero, line 91:
```suggestion
            return streams
            .Where(i => i.Type == type)
```

Indentation seems off?

crobibero (approved):

Issue #7237, opened 2022-01-22
[Issue]: Reduce the number of queries


jaysonsantos:
### Please describe your bug

Hi there, first thank you all for this amazing project, I really want to see Jellyfin as a 1st grade product :).
I'd like to open this issue in order to propose query reduction per routes, so this could make more effective usage of the disk when sqlite has to fetch data from it.
To try and measure the pain points of the app, I started a branch to instrumentate the code with opentelemetry (if it is something you would like to have upstream, let me know).
Here you can see an average of time per route, note that my local server uses HDD, so any random read would suffer.
![image](https://user-images.githubusercontent.com/51510/150653393-0f9d9909-96ef-451b-8568-912ab7c6b9db.png)
Then you can see on the details, that many queries will be issued (a bit hard to see here, but i can share the dashboard with whoever wants to see it).
![image](https://user-images.githubusercontent.com/51510/150653452-31dc5dae-8045-46b7-be9a-36b976fc0b42.png)

The effects of this is that you can also see that it seems to be locked in a transaction because other threads/tasks are using the db for a longer period.
![image](https://user-images.githubusercontent.com/51510/150653518-feb0da4b-cf83-47ec-a532-c358e54ea783.png)

What do you say? Do you see this as an improvement to be made? 
I see slow endpoints even when running from my dev machine with nvme, so I guess that the locks are the ones affecting its performance.

### Jellyfin Version

Other

### if other:

master branch

### Environment

```markdown
- OS: Ubuntu 20.04
- Virtualization: Bare metal
- Clients: Browser, Android, Fire TV Cube, Android TV, iOS
- Browser: Chrome
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- Plugins: trakt
- Reverse Proxy: direct connection
- Base URL: /
- Networking: host
- Storage: nfs
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cvium:
I would personally be interested in seeing how you did the instrumentation, but I'm not sure it's something we'd upstream.

jaysonsantos:
@cvium my current local patch you can find here https://gist.github.com/0ae616d0f3c73c577da8eff72707e1f6

jaysonsantos:
@cvium the cool thing is that it could be upstreamed without additional "weight" on standar installations because it only reports for subscribers of the ActivitySource's, and it could be only activated if provided a collector's url, i guess this would help observe patterns where the app is being slow

[...]

PR #7240, opened 2022-01-22
chore: Add a read only connection for routes like Shows/NextUp

jaysonsantos:
This is an atempt to reduce the number of locks reported on #7237 by having a connection that can be shared among the threads without a semaphore

### DIFF
--- a/Emby.Server.Implementations/Data/BaseSqliteRepository.cs
+++ b/Emby.Server.Implementations/Data/BaseSqliteRepository.cs
@@ -97,6 +97,7 @@ namespace Emby.Server.Implementations.Data
         /// </summary>
         /// <value>The write connection.</value>
         protected SQLiteDatabaseConnection WriteConnection { get; set; }
+
         protected SQLiteDatabaseConnection ReadConnection { get; set; }
 
         protected ManagedConnection GetConnection(bool readOnly = false)
@@ -104,7 +105,7 @@ namespace Emby.Server.Implementations.Data
             if (readOnly)
             {
                 ReadConnection ??= SQLite3.Open(DbFilePath, ConnectionFlags.ReadOnly, null);
-                return new ManagedConnection(ReadConnection, null!);
+                return new ManagedConnection(ReadConnection, null);
             }
 
             WriteLock.Wait();
--- a/Emby.Server.Implementations/Data/BaseSqliteRepository.cs
+++ b/Emby.Server.Implementations/Data/BaseSqliteRepository.cs
@@ -103,8 +103,8 @@ namespace Emby.Server.Implementations.Data
         {
             if (readOnly)
             {
-                ReadConnection ??= SQLite3.Open(DbFilePath, DefaultConnectionFlags | ConnectionFlags.ReadOnly, null);
-                return new ManagedConnection(ReadConnection, null);
+                ReadConnection ??= SQLite3.Open(DbFilePath, ConnectionFlags.ReadOnly, null);
+                return new ManagedConnection(ReadConnection, null!);
             }
 
             WriteLock.Wait();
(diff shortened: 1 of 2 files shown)


crobibero (approved):

jaysonsantos:
Hey there, please, put this on hold as I think that it seems that SQLitePCL.raw is not thrad safe, I am going to give a read on its source code to see what can be done there otherwise a `System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.` happens

nielsvanvelzen:
I marked the pull request as a draft, when it's ready again just press the "ready for review" button.

jaysonsantos:
Ok, this should be working now.
After reading on https://github.com/ericsink/SQLitePCL.raw/issues/430 I saw that the default `SQLitePCL.pretty.ConnectionFlags` comes with `NoMutex` enabled.
In another iteration I'd add a simple "pooler" to have N read only connections and randomly distribute them to `GetConnection` calls, this way we could mitigate slow queries blocking other read queries

jaysonsantos:
It is not perfect yet, but it seems to be a bit faster:
image
Navigation feels a bit faster as well.

1337joe (approved):

Emby.Server.Implementations/Data/BaseSqliteRepository.cs
```diff
@@ -97,9 +97,16 @@ protected BaseSqliteRepository(ILogger<BaseSqliteRepository> logger)
         /// </summary>
         /// <value>The write connection.</value>
         protected SQLiteDatabaseConnection WriteConnection { get; set; }
+        protected SQLiteDatabaseConnection ReadConnection { get; set; }
```
crobibero, line 100:
```suggestion

        protected SQLiteDatabaseConnection ReadConnection { get; set; }
```

Newline between properties

Emby.Server.Implementations/Data/ManagedConnection.cs
```diff
@@ -9,11 +9,12 @@ namespace Emby.Server.Implementations.Data
...
     public sealed class ManagedConnection : IDisposable
     {
-        private readonly SemaphoreSlim _writeLock;
+        private readonly SemaphoreSlim? _writeLock;
 
         private SQLiteDatabaseConnection? _db;
 
-        private bool _disposed = false;
+        private bool _disposed;
+
```
crobibero, line 17:
```suggestion
```

Emby.Server.Implementations/Data/ManagedConnection.cs
```diff
@@ -9,11 +9,12 @@ namespace Emby.Server.Implementations.Data
...
 
         public ManagedConnection(SQLiteDatabaseConnection db, SemaphoreSlim writeLock)
```
crobibero, line 19:
```suggestion
        public ManagedConnection(SQLiteDatabaseConnection db, SemaphoreSlim? writeLock)
```

Emby.Server.Implementations/Data/BaseSqliteRepository.cs
```diff
@@ -97,9 +97,16 @@ protected BaseSqliteRepository(ILogger<BaseSqliteRepository> logger)
...
 
         protected ManagedConnection GetConnection(bool readOnly = false)
         {
+            if (readOnly)
+            {
+                ReadConnection ??= SQLite3.Open(DbFilePath, ConnectionFlags.ReadOnly, null);
+                return new ManagedConnection(ReadConnection, null!);
```
crobibero, line 107:
```suggestion
                return new ManagedConnection(ReadConnection, null);
```

crobibero (approved):

jaysonsantos:
@crobibero how did you get to the point of that exception?
i've rebased my code with the current master and i cannot reproduce it

crobibero:
I'm fairly certain it was during scan.
I was also informed that we had attempted this previously and ran into other issues

jaysonsantos:
@crobibero i was taking a look to see if somehow there were more instances of the semaphore used there and when the problem happens, nothing is actually leaking, i am thinking that the problem could be with the sqlite library that is somehow not releasing the lock when disposed.
what do you say if i send a pr where I register the database connections as dependency injections and use microsoft.data.sqlite that is using newer versions of sqlitepcl.raw? the one jellyfin currentlyh uses is more than 6 years without a new release.

PR #7245, opened 2022-01-23
add stale issue check workflow

h1dden-da3m0n:
### Description

This PR aims to add a replacement for the discontinued Stale-Bot configuration in favour of a GitHub Action workflow.
It replicates the old config for Stale-Bot and uses the default GitHub Actions token to comment and close stale Issues.

> NOTE: It explicitly does not touch PRs

### Changes

* add a basic GitHub action workflow to monitor, label, and clos stale issues within this repository

### Issues

* successor to the discontinued Stale-Bot config


crobibero (approved):

.github/workflows/repo-stale.yaml
```diff
@@ -0,0 +1,31 @@
+name: Issue Stale Check
+
+on:
+  schedule:
+    - cron: '30 1 * * *'
+  workflow_dispatch:
+
+permissions:
+  issues: write
+  pull-requests: write
```
Bond-009, line 10: Is this needed even if it doesn't close PRs?

Bond-009 (approved):

.github/workflows/repo-stale.yaml
h1dden-da3m0n: tbh I should clean that up entirely. this would only be effectife for the `{{ github.token }}` that gets injected into the CI run. 
Usually for in repo triggered workflows the token has full write access on all levels of the repo, unless the 'restrictive' mode is used, then you would be required to use the permission block...
Regardless, you are indeed correct, this would not be needed in any case if the workflow wont ever touch PRs.

PR #7246, opened 2022-01-23
Remove some allocations

cvium:
**Changes**
Removed some cruft

### DIFF
--- a/MediaBrowser.Controller/Entities/BaseItem.cs
+++ b/MediaBrowser.Controller/Entities/BaseItem.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
+using System.IO;
 using System.Linq;
 using System.Text;
 using System.Text.Json.Serialization;
@@ -1286,7 +1287,7 @@ namespace MediaBrowser.Controller.Entities
                 {
                     if (IsFileProtocol)
                     {
-                        requiresSave = await RefreshedOwnedItems(options, GetFileSystemChildren(options.DirectoryService).ToList(), cancellationToken).ConfigureAwait(false);
+                        requiresSave = await RefreshedOwnedItems(options, GetFileSystemChildren(options.DirectoryService), cancellationToken).ConfigureAwait(false);
                     }
 
                     await LibraryManager.UpdateImagesAsync(this).ConfigureAwait(false); // ensure all image properties in DB are fresh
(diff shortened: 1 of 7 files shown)


MediaBrowser.Controller/Entities/BaseItem.cs
```diff
@@ -2041,27 +2042,32 @@ public virtual Task UpdateToRepositoryAsync(ItemUpdateType updateReason, Cancell
...
+            {
+                if (!imageInfo.IsLocalFile)
+                {
+                    continue;
+                }
 
-            var deletedImages = ImageInfos
-                .Where(image => image.IsLocalFile && !allFiles.Contains(image.Path, StringComparison.OrdinalIgnoreCase))
-                .ToList();
+                if (File.Exists(imageInfo.Path))
```
cvium, line 2056: This might be worse for network storage (as DirectoryService pulled all paths from the containing folder) but idk

crobibero (approved):

MediaBrowser.LocalMetadata/Images/LocalImageProvider.cs
```diff
@@ -449,32 +449,38 @@ private bool AddImage(List<FileSystemMetadata> files, List<LocalImageInfo> image
...
+        private static FileSystemMetadata? GetImage(IReadOnlyList<FileSystemMetadata> files, string name, string? prefix = null)
         {
+            var fileNameLength = name.Length + (prefix?.Length ?? 0);
             for (var i = 0; i < files.Count; i++)
             {
                 var file = files[i];
-                if (!file.IsDirectory
-                    && file.Length > 0
-                    && Path.GetFileNameWithoutExtension(file.FullName.AsSpan()).Equals(name, StringComparison.OrdinalIgnoreCase))
+                if (file.IsDirectory || file.Length <= 0 || file.FullName.Length != fileNameLength)
```
Bond-009, line 476: This won't work?
`file.FullName` return the whole path, not just the file name.

MediaBrowser.LocalMetadata/Images/LocalImageProvider.cs
```diff
@@ -449,32 +449,38 @@ private bool AddImage(List<FileSystemMetadata> files, List<LocalImageInfo> image
...
+                {
+                    continue;
+                }
+
+                var fileName = Path.GetFileNameWithoutExtension(file.FullName.AsSpan());
+                if (fileName.StartsWith(prefix, StringComparison.OrdinalIgnoreCase)
+                    && fileName.EndsWith(name, StringComparison.OrdinalIgnoreCase))
```
Bond-009, line 483: Is this really worth the effort?

Bond-009 (changes requested):

MediaBrowser.LocalMetadata/Images/LocalImageProvider.cs
cvium: Right, brainfart.

MediaBrowser.LocalMetadata/Images/LocalImageProvider.cs
Bond-009: :shrug: It's more brittle now (see example above :p )

cvium: That is true, but only because I was too lazy to add a test...

Bond-009: The other solution is to stack alloc a `Span<char>` and put it in there, which would make comparing easier, but with the added risk of having a string too large for the stack

Bond-009 (approved):

MediaBrowser.LocalMetadata/Images/LocalImageProvider.cs
cvium: Isn't it unlikely with file paths?

Bond-009: `unlikely` sure, but not impossible

PR #7256, opened 2022-01-27
Proxy remote image fetching through the server

lanslot6:
**Changes**
Proxy remote image fetching through the server:

- Add a `GET Items/{itemId}/RemoteImages/Fetch` endpoint which expects an `imageUrl` query parameter, and proxies the `GET {imageUrl}` request on behalf of the client.
- Modify the `GET Items/{itemId}/RemoteImages` endpoint to return remote image urls using the `Fetch` endpoint, i.e. `Items/{itemId}/RemoteImages/Fetch?imageUrl={rawImageUrl}` instead of plain `{rawImageUrl}`.
- Modify the `POST Items/{itemId}/RemoteImages/Download` endpoint to strip out the `Fetch` endpoint prefix, i.e. removing the `Items/{itemId}/RemoteImages/Fetch?imageUrl=` prefix to get the raw image url.

This is useful when:

- The user wants to make sure all scraping requests are initiated from the server side because, e.g., the server uses a VPN for scraping.
- The remote image url is not accessible from clients, e.g. when the remote image url points to an custom endpoint hosted next to the Jellyfin server.

Configuration option ideas not included in this PR:

1. Add a server configuration to globally enable or disable proxying for remote image requests.
2. Add a server configuration to selectively enable proxying for remote image requests if and only if the remote host or IP matches a configured list.

I think 1 is good but 2 might be an overkill. Alternatively I can also see a case for not providing an option at all.

**Issues**
n/a

### DIFF
--- a/Jellyfin.Api/Controllers/RemoteImageController.cs
+++ b/Jellyfin.Api/Controllers/RemoteImageController.cs
@@ -3,9 +3,15 @@ using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
 using System.IO;
 using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Net.Http.Headers;
 using System.Threading;
 using System.Threading.Tasks;
+using Jellyfin.Api.Attributes;
 using Jellyfin.Api.Constants;
+using MediaBrowser.Common.Extensions;
+using MediaBrowser.Common.Net;
 using MediaBrowser.Controller;
 using MediaBrowser.Controller.Library;
 using MediaBrowser.Controller.Providers;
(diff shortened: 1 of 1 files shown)


DMouse10462:
I'm curious how this relates to a similar security advisory in which the response was to *remove* this kind of functionality for non-Administrators.

https://github.com/jellyfin/jellyfin/security/advisories/GHSA-rgjw-4fwc-9v96

lanslot6:
> I'm curious how this relates to a similar security advisory in which the response was to _remove_ this kind of functionality for non-Administrators.
> 
> [GHSA-rgjw-4fwc-9v96](https://github.com/jellyfin/jellyfin/security/advisories/GHSA-rgjw-4fwc-9v96)

Thanks for the pointer. I am concerned about the security implications too. Given that said, I don't think this PR adds much additional risk on top of the existing issues.

I think the true fix should be modifying the web client to fetch and render images via XHR which can be authenticated. Native clients can already do this easily.

cvium:
It seems to me that this PR adds back the issue that was patched as well as a potential method for a DoS attack.

lanslot6:
I will look into the web client fix to use an authenticated endpoint to fetch images.

PR #7346, opened 2022-02-21
Optimize Guid comparisons

Bond-009:
* Use Guid.Equals(Guid) instead of the == override
* Ban the usage of Guid.Equals(Object) to prevent accidental boxing
* Compare to default(Guid) instead of Guid.Empty

### DIFF
--- a/BannedSymbols.txt
+++ b/BannedSymbols.txt
@@ -1,1 +1,4 @@
 P:System.Threading.Tasks.Task`1.Result
+M:System.Guid.op_Equality(System.Guid,System.Guid)
+M:System.Guid.op_Inequality(System.Guid,System.Guid)
+M:System.Guid.Equals(System.Object)
--- a/Emby.Dlna/Didl/DidlBuilder.cs
+++ b/Emby.Dlna/Didl/DidlBuilder.cs
@@ -160,7 +160,7 @@ namespace Emby.Dlna.Didl
             else
             {
                 var parent = item.DisplayParentId;
-                if (!parent.Equals(Guid.Empty))
+                if (!parent.Equals(default))
                 {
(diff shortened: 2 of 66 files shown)


crobibero (approved):

cvium (approved):

ravindUwU:
Randomly noticed [`Guid == Guid` listed as a banned symbol](https://github.com/jellyfin/jellyfin/blob/fbb0f1afbcd52a15d6e56742bba0f338a5ced88f/BannedSymbols.txt#L2-L3).

As far as I can see, `==` ([.NET](https://github.com/dotnet/runtime/blob/d2ea726dd3d8faad797755c9fbb1ff15226e25f0/src/libraries/System.Private.CoreLib/src/System/Guid.cs#L1182), [Reference Source](https://github.com/microsoft/referencesource/blob/ec9fa9ae770d522a5b5f0607898044b7478574a3/mscorlib/system/guid.cs#L1108-L1135)) and `.Equals` ([.NET](https://github.com/dotnet/runtime/blob/d2ea726dd3d8faad797755c9fbb1ff15226e25f0/src/libraries/System.Private.CoreLib/src/System/Guid.cs#L1086), [Reference Source](https://github.com/microsoft/referencesource/blob/ec9fa9ae770d522a5b5f0607898044b7478574a3/mscorlib/system/guid.cs#L965-L992)) do the same thing under the hood, I'm curious as to what prompted this change 🤔.

(I realise this was back in 2022, so it's entirely possible that the two weren't the same back then).

cvium:
It's in the description. To avoid boxing.

ravindUwU:
To avoid accidental `Guid == object` comparisons (because `Guid == Guid` doesn't box AFAIK?); I see. Thanks! 😄

PR #7347, opened 2022-02-21
Migrate from IWebHost to IHost

crobibero:
current error:
```
[08:17:28] [FTL] [1] Main: Error while starting server.
System.InvalidOperationException: Unable to resolve service for type 'MediaBrowser.Controller.Configuration.IServerConfigurationManager' while attempting to activate 'Jellyfin.Server.Startup'.
   at Microsoft.Extensions.DependencyInjection.ActivatorUtilities.ConstructorMatcher.CreateInstance(IServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ActivatorUtilities.CreateInstance(IServiceProvider provider, Type instanceType, Object[] parameters)
   at Microsoft.AspNetCore.Hosting.GenericWebHostBuilder.UseStartup(Type startupType, HostBuilderContext context, IServiceCollection services, Object instance)
   at Microsoft.AspNetCore.Hosting.GenericWebHostBuilder.<>c__DisplayClass13_0.<UseStartup>b__0(HostBuilderContext context, IServiceCollection services)
   at Microsoft.Extensions.Hosting.HostBuilder.CreateServiceProvider()
   at Microsoft.Extensions.Hosting.HostBuilder.Build()
   at Jellyfin.Server.Program.StartApp(StartupOptions options) in C:\Code\jellyfin\Jellyfin.Server\Program.cs:line 190
```

### DIFF
--- a/Jellyfin.Server/Jellyfin.Server.csproj
+++ b/Jellyfin.Server/Jellyfin.Server.csproj
@@ -41,7 +41,7 @@
     <PackageReference Include="Microsoft.Extensions.Diagnostics.HealthChecks.EntityFrameworkCore" Version="6.0.9" />
     <PackageReference Include="prometheus-net" Version="6.0.0" />
     <PackageReference Include="prometheus-net.AspNetCore" Version="6.0.0" />
-    <PackageReference Include="Serilog.AspNetCore" Version="4.1.0" />
+    <PackageReference Include="Serilog.AspNetCore" Version="6.0.1" />
     <PackageReference Include="Serilog.Enrichers.Thread" Version="3.1.0" />
     <PackageReference Include="Serilog.Settings.Configuration" Version="3.4.0" />
     <PackageReference Include="Serilog.Sinks.Async" Version="1.5.0" />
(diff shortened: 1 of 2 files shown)


cvium:
@crobibero Status?

crobibero:
@cvium Its been a while since I worked on this but I think this changes the pipeline order which caused DI errors. I think we should still migrate to the new host implementation but it mah require some real work

cvium:
Shouldn't we be using https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.webapplication.createbuilder?view=aspnetcore-6.0 anyway?

crobibero:
We should, but Serilog deprecated `UseSerilog()` on `IWebAppHostBuilder` in v5 so we need to use `IHostBuilder` if we want to continue use.

https://github.com/serilog/serilog-aspnetcore/releases/tag/v5.0.0

I spent some more time trying to get this working but I ran into the same service resolve issue

Bond-009:
Superseded by #9078

PR #7378, opened 2022-02-28
Add resume book as default home section

thornbill:
**Changes**
Updates the default home section to match web values

**Issues**
N/A

### DIFF
--- a/Jellyfin.Api/Controllers/DisplayPreferencesController.cs
+++ b/Jellyfin.Api/Controllers/DisplayPreferencesController.cs
@@ -126,9 +126,11 @@ namespace Jellyfin.Api.Controllers
                 HomeSectionType.SmallLibraryTiles,
                 HomeSectionType.Resume,
                 HomeSectionType.ResumeAudio,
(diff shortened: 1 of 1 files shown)


crobibero (approved):

1337joe (approved):

Jellyfin.Api/Controllers/DisplayPreferencesController.cs
```diff
@@ -182,7 +184,7 @@ public DisplayPreferencesController(IDisplayPreferencesManager displayPreference
                 var order = int.Parse(key.AsSpan().Slice("homesection".Length));
                 if (!Enum.TryParse<HomeSectionType>(displayPreferences.CustomPrefs[key], true, out var type))
                 {
-                    type = order < 7 ? defaults[order] : HomeSectionType.None;
+                    type = order < 8 ? defaults[order] : HomeSectionType.None;
```
nielsvanvelzen, line 187: I think this line doesn't make any sense anymore because there can only be 7 homesections in total. Same with the last item in the defaults list.

PR #7388, opened 2022-03-01
Hide ISO 639-2 special codes in display titles

GermanCoding:
This has bugged me for a while now.

Jellyfin shows the language for audio streams, if it is present in metadata. This is great and correct.

However, there are audio streams that make use of ISO 639-2 special codes. Those then look like this:

before

"und" is ISO-special for undetermined, which is confusing for users (especially as "und" is also a word in german). Generally I think some of these specials are rarely known, e.g who knows that "zxx" is "no linguistic content; not applicable".

Therefore I believe having these specials displayed prominently in display titles is wrong. I get that there is still value in displaying this information - it's technical data that may be of interest to more technical users. Hence I did **not** remove language codes displaying from metadata - just from the display title.

Not it looks like this, much cleaner IMHO:

after

Language codes are still shown in associated metadata:

metadata

There are also more language codes that Jellyfin currently can't expand (ISO 639-2 B codes for example), but here I think that there is still value in displaying them, hence they are not filtered. It's just the specials that cause more confusion than providing information to users.

**Changes**

Filters ISO639-2 special codes (except the reserved range - whoever uses that hopefully knows what they're doing) from display titles.

**Testing**

As shown in the screenshots above, manually verified that this works as intended.

### DIFF
--- a/MediaBrowser.Model/Entities/MediaStream.cs
+++ b/MediaBrowser.Model/Entities/MediaStream.cs
@@ -17,6 +17,18 @@ namespace MediaBrowser.Model.Entities
     /// </summary>
     public class MediaStream
     {
+        private static readonly string[] _specialCodes =
+        {
+            // Uncoded languages.
+            "mis",
+            // Multiple languages.
+            "mul",
+            // Undetermined.
+            "und",
+            // No linguistic content; not applicable.
+            "zxx"
+        };
+
         /// <summary>
         /// Gets or sets the codec.
         /// </summary>
@@ -137,7 +149,8 @@ namespace MediaBrowser.Model.Entities
                     {
                         var attributes = new List<string>();
 
-                        if (!string.IsNullOrEmpty(Language))
+                        // Do not display the language code in display titles if unset or set to a special code. Show it in all other cases (possibly expanded).
+                        if (!string.IsNullOrEmpty(Language) && !_specialCodes.Contains(Language, StringComparer.OrdinalIgnoreCase))
                         {
                             // Get full language string i.e. eng -> English. Will not work for some languages which use ISO 639-2/B instead of /T codes.
                             string fullLanguage = CultureInfo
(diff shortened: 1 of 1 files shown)


crobibero:
The list of codes should be initialized once, and you can probably use .Contains with string comparison instead of using Linq

MediaBrowser.Model/Entities/MediaStream.cs
```diff
@@ -17,6 +17,8 @@ namespace MediaBrowser.Model.Entities
     /// </summary>
     public class MediaStream
     {
+        private static readonly List<string> _specialCodes = new() { "mis" /*uncoded languages*/, "mul" /*multiple languages*/, "und" /*undetermined*/, "zxx" /*no linguistic content; not applicable*/ };
```
crobibero, line 20:
```suggestion
        private static readonly string[] _specialCodes =
        {
            // Uncoded languages.
            "mis",
            // Multiple languages.
            "mul",
            // Undetermined.
            "und",
            // No linguistic content; not applicable.
            "zxx"
        };
```

crobibero (approved):

Shadowghost:
I'd say squash it but we're not that concerned about it most of the time.

MediaBrowser.Model/Entities/MediaStream.cs
```diff
@@ -137,7 +149,8 @@ public string DisplayTitle
                     {
                         var attributes = new List<string>();
 
-                        if (!string.IsNullOrEmpty(Language))
+                        // Do not display the language code in display titles if unset or set to a special code. Show it in all other cases (possibly expanded).
+                        if (!string.IsNullOrEmpty(Language) && !_specialCodes.Contains(Language, StringComparer.OrdinalIgnoreCase))
```
cvium, line 153: We have our own extension method that is slightly better optimized and uses StringComparison (may require an additional `using`)
```suggestion
                        if (!string.IsNullOrEmpty(Language) && !_specialCodes.Contains(Language, StringComparison.OrdinalIgnoreCase))
```

1337joe (approved):

PR #7406, opened 2022-03-05
Fix build

crobibero:


Bond-009 (approved):

Jellyfin.Api/Controllers/ImageController.cs
```diff
@@ -1,4 +1,3 @@
-using System;
```
nielsvanvelzen, line 1: ^ I think this removal doesn't really fix the build

1337joe: Yeah, this is the cause of all these errors:
`The type or namespace name 'Guid' could not be found (are you missing a using directive or an assembly reference?)`

1337joe (approved):

PR #7432, opened 2022-03-08
SocketFactory: Remove redundant code

Bond-009:
ExclusiveAddressUse should be false by default afaik

### DIFF
--- a/Emby.Server.Implementations/Net/SocketFactory.cs
+++ b/Emby.Server.Implementations/Net/SocketFactory.cs
@@ -80,6 +80,8 @@ namespace Emby.Server.Implementations.Net
 
             var retVal = new Socket(AddressFamily.InterNetwork, SocketType.Dgram, ProtocolType.Udp);
 
+            retVal.ExclusiveAddressUse = false;
+
             try
             {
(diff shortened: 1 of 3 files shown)


crobibero (approved):

1337joe:
Examining it in the debugger `ExclusiveAddressUse` is not false by default (on my machine - arch linux running master).

It's not just a bool field, the Socket.cs code in question is:
```cs
    public bool ExclusiveAddressUse
    {
      get => (int) this.GetSocketOption(SocketOptionLevel.Socket, SocketOptionName.ExclusiveAddressUse) != 0;
      set
      {
        if (this.IsBound)
          throw new InvalidOperationException(SR.net_sockets_mustnotbebound);
        this.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.ExclusiveAddressUse, value ? 1 : 0);
      }
    }
```

Bond-009:
@1337joe In that case, the docs are lying
> The default is true for Windows Server 2003 and Windows XP Service Pack 2, and false for all other versions.

I'll fix it

1337joe (approved):

PR #7433, opened 2022-03-08
Rework SsdpHttpClient

Bond-009:
Log malformed responses instead of silently ignoring them
Use our own user agent


Emby.Dlna/PlayTo/DlnaHttpClient.cs
```diff
@@ -0,0 +1,105 @@
+#pragma warning disable CS1591
```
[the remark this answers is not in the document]
Bond-009: It's renamed ;)

Jellyfin.Server/Startup.cs
```diff
@@ -103,6 +104,22 @@ public void ConfigureServices(IServiceCollection services)
                 })
                 .ConfigurePrimaryHttpMessageHandler(defaultHttpClientHandlerDelegate);
 
+            services.AddHttpClient(NamedClient.Dlna, c =>
+                {
+                    c.DefaultRequestHeaders.UserAgent.ParseAdd(
+                        string.Format(
+                            CultureInfo.InvariantCulture,
+                            "{0}/{1} UPnP/1.1 {2}/{3}",
```
Bond-009, line 112: Should probably keep this as `UPnP/1.0`, right?

crobibero (approved):

Jellyfin.Server/Startup.cs
```diff
@@ -103,6 +104,22 @@ public void ConfigureServices(IServiceCollection services)
...
+                        string.Format(
+                            CultureInfo.InvariantCulture,
+                            "{0}/{1} UPnP/1.0 {2}/{3}",
+                            MediaBrowser.Common.System.OperatingSystem.Name,
+                            Environment.OSVersion,
+                            _serverApplicationHost.Name,
+                            _serverApplicationHost.ApplicationVersionString));
+
+                    c.DefaultRequestHeaders.Add("CPFN.UPNP.ORG", _serverApplicationHost.FriendlyName); // Required for UPnP DeviceArchitecture v2.0
+                    c.DefaultRequestHeaders.Add("FriendlyName.DLNA.ORG", _serverApplicationHost.FriendlyName); // REVIEW: where does this come from?
```
1337joe, line 119: Google hits on log entries with that header say jellyfin/emby are far from the only DLNA server to use it, and this is probably an old version but it defines it:

[DLNA Networked Device Interoperability Guidelines](https://docshare.tips/dlna-networked-device-interoperability-guidelines-architectures-and-protocols_58c06e18b6d87fac848b5393.html) Section 7.4.53:
> Requirement [7.4.53.1]: 
> HTTP Client Endpoints may use the friendlyName.dlna.org HTTP to communicate a user friendly name to the Content Source Endpoint in HTTP GET requests.

1337joe (approved):

Jellyfin.Server/Startup.cs
PrplHaz4: I believe it traces back to the Server Name field on Admin -> General

some practical examples located here: https://github.com/UniversalMediaServer/UniversalMediaServer/search?q=FriendlyName

Bond-009: @1337joe I think that's for clients, but I'm not 100% sure https://github.com/xindawndev/dlna-microstack/blob/master/include/dlna/DlnaHttp.h#L554

PR #7325, opened 2022-02-17
Rework StreamBuilder to avoid unnecessary transcoding

eyezak:
**Changes**
Adds preferred matching of video and audio streams to both DirectPlay and Transcode profiles. Multiple candidate profiles are considered after codec resolution, with preference to copy-codec. In the case of compatible video streams, remuxing is preferred to transcoding audio. In the cases of no compatible video streams,  multiple equivalent candidate profiles and !Allow*Copy - preference is given to the client-supplied profile order to support prefer fMP4/ts streams user-setting. In contrast to the approach of https://github.com/jellyfin/jellyfin/pull/6454, PlaybackInfo based apis now prefer transcoding to direct streaming to allow direct-stream equivalent transcoding to HLS (ts or fMP4).
MediaInfoHelper has been adapted to only call StreamBuilder once.

Converts TranscodeReason to Flags and serialize into strings in models for simplicity.
Adds a suite of tests for StreamBuilder based on real and imagined device profiles.

Currently testing on Firefox/Chrome/Safari with av1,divx,h264,hevc,msmp4,xvid x aac,ac3,eac3,mp3,flac,opus,vorbis x mp4,mkv,avi,webm encodes of the Sintel 2048 trailer. Verified HEVC10 and HEVC on safari with remux and transcode as direct-stream with web-player modifications (https://github.com/jellyfin/jellyfin-web/pull/3442) .

**Issues**
fixes #6450
fixes #6149
alternative approach to #6454 
Potentially others i have not validated yet

PR #7449, opened 2022-03-12
Revert using Math.Clamp

crobibero:
Fixes edgecase tests.

Reverts change suggested here: https://github.com/jellyfin/jellyfin/pull/7325#discussion_r820286509

### DIFF
--- a/MediaBrowser.Model/Dlna/StreamBuilder.cs
+++ b/MediaBrowser.Model/Dlna/StreamBuilder.cs
@@ -909,8 +909,9 @@ namespace MediaBrowser.Model.Dlna
                 }
 
                 // Make sure the video bitrate is lower than bitrate settings but at least 64k
+                // Don't use Math.Clamp as availableBitrateForVideo can be lower then 64k.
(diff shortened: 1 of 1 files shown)


1337joe (approved):

Bond-009 (approved):

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -910,7 +910,7 @@ private void BuildStreamVideoItem(StreamInfo playlistItem, VideoOptions options,
 
                 // Make sure the video bitrate is lower than bitrate settings but at least 64k
                 var currentValue = playlistItem.VideoBitrate ?? availableBitrateForVideo;
-                playlistItem.VideoBitrate = Math.Clamp(currentValue, 64_000, availableBitrateForVideo);
+                playlistItem.VideoBitrate = Math.Max(Math.Min(availableBitrateForVideo, currentValue), 64_000);
```
Bond-009, line 913: I'd add a comment that we can't use `Math.Clamp` here bc `availableBitrateForVideo` can be lower then `64_000`
Otherwise I'll run into this again in the future :p

PR #7466, opened 2022-03-17
Prefer Cuvid to Nvdec by default for better performance

nyanmisaka:
**Changes**
- Prefer Cuvid to Nvdec by default for better performance

### DIFF
--- a/MediaBrowser.Model/Configuration/EncodingOptions.cs
+++ b/MediaBrowser.Model/Configuration/EncodingOptions.cs
@@ -32,7 +32,7 @@ namespace MediaBrowser.Model.Configuration
             DeinterlaceMethod = "yadif";
             EnableDecodingColorDepth10Hevc = true;
             EnableDecodingColorDepth10Vp9 = true;
-            EnableEnhancedNvdecDecoder = true;
+            EnableEnhancedNvdecDecoder = false;
             PreferSystemNativeHwDecoder = true;
             EnableIntelLowPowerH264HwEncoder = false;
             EnableIntelLowPowerHevcHwEncoder = false;


crobibero (approved):

1337joe (approved):

Tontonjo:
I did some testing on this: 
The fact is that it comes at the cost of a much higher VRAM usage, wich could prevent, in the case  of some streams needs to be transcoded, to a fail because there's not enough VRAM left.

Setted OFF:
 0   N/A  N/A     25660      C   ...ib/jellyfin-ffmpeg/ffmpeg      861MiB
Setted ON:
0   N/A  N/A     26858      C   ...ib/jellyfin-ffmpeg/ffmpeg     1529MiB

So  in this case, if i have a Quadro P400 with 2GB of Vram, wich is very common, another transcode task would fail but if i left it OFF, i could have at least 2 streams running.

I'd suggest leaving this OFF as i thing it's better to have a maybe slower stream than no stream at all.

nyanmisaka:
> I did some testing on this: The fact is that it comes at the cost of a much higher VRAM usage, wich could prevent, in the case of some streams needs to be transcoded, to a fail because there's not enough VRAM left.
> 
> Setted OFF: 0 N/A N/A 25660 C ...ib/jellyfin-ffmpeg/ffmpeg 861MiB Setted ON: 0 N/A N/A 26858 C ...ib/jellyfin-ffmpeg/ffmpeg 1529MiB
> 
> So in this case, if i have a Quadro P400 with 2GB of Vram, wich is very common, another transcode task would fail but if i left it OFF, i could have at least 2 streams running.
> 
> I'd suggest leaving this OFF as i think it's better to have a maybe slower stream than no stream at all.
> 
> If you dont want to go that way, maybe a notice would be great under the option to notify that this cause a high VRAM usage.

https://github.com/jellyfin/jellyfin/pull/7387
https://github.com/jellyfin/jellyfin-web/pull/3523

Tontonjo:
> > I did some testing on this: The fact is that it comes at the cost of a much higher VRAM usage, wich could prevent, in the case of some streams needs to be transcoded, to a fail because there's not enough VRAM left.
> > Setted OFF: 0 N/A N/A 25660 C ...ib/jellyfin-ffmpeg/ffmpeg 861MiB Setted ON: 0 N/A N/A 26858 C ...ib/jellyfin-ffmpeg/ffmpeg 1529MiB
> > So in this case, if i have a Quadro P400 with 2GB of Vram, wich is very common, another transcode task would fail but if i left it OFF, i could have at least 2 streams running.
> > I'd suggest leaving this OFF as i think it's better to have a maybe slower stream than no stream at all.
> > If you dont want to go that way, maybe a notice would be great under the option to notify that this cause a high VRAM usage.
> 
> #7387 [jellyfin/jellyfin-web#3523](https://github.com/jellyfin/jellyfin-web/pull/3523)

Ho Thanks for clarifying! Very good  job.

PR #7470, opened 2022-03-17
Remove check for non-existent "emby rt" client in UserViewsController

nielsvanvelzen:
**Changes**
Removed a check in the GetUserViews operation that modifies the query based on a client name that we don't use anywhere[^1]

[^1]: https://cs.github.com/?scopeName=All+repos&scope=&q=org%3Ajellyfin+%22emby+rt%22

**Issues**


Bond-009 (approved):

Bond-009:
This doesn't have to be a Task anymore, no?

1337joe (approved):

PR #7477, opened 2022-03-19
Use requested bitrate for direct play calculations

crobibero:
Potentially fixes regression in https://github.com/jellyfin/jellyfin/pull/7325

### DIFF
--- a/MediaBrowser.Model/Dlna/StreamBuilder.cs
+++ b/MediaBrowser.Model/Dlna/StreamBuilder.cs
@@ -362,16 +362,6 @@ namespace MediaBrowser.Model.Dlna
             return playlistItem;
         }
 
-        private static long? GetBitrateForDirectPlayCheck(MediaSourceInfo item, AudioOptions options, bool isAudio)
-        {
-            if (item.Protocol == MediaProtocol.File)
-            {
-                return options.Profile.MaxStaticBitrate;
-            }
-
-            return options.GetMaxBitrate(isAudio);
-        }
-
         private (DirectPlayProfile Profile, PlayMethod? PlayMethod, TranscodeReason TranscodeReasons) GetAudioDirectPlayProfile(MediaSourceInfo item, MediaStream audioStream, AudioOptions options)
         {
             var directPlayProfile = options.Profile.DirectPlayProfiles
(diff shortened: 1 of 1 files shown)


nyanmisaka (approved):
It works for me.

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -395,7 +385,7 @@ private StreamInfo BuildAudioItem(MediaSourceInfo item, AudioOptions options)
             // If device requirements are satisfied then allow both direct stream and direct play
             if (item.SupportsDirectPlay)
             {
-                if (IsItemBitrateEligibleForDirectPlay(item, GetBitrateForDirectPlayCheck(item, options, true) ?? 0, PlayMethod.DirectPlay))
+                if (IsItemBitrateEligibleForDirectPlay(item, options.GetMaxBitrate(false) ?? 0, PlayMethod.DirectPlay))
```
dmitrylyzo, line 388:
```suggestion
                if (IsItemBitrateEligibleForDirectPlay(item, options.GetMaxBitrate(true) ?? 0, PlayMethod.DirectPlay))
```
because this is audio (also `true` as the last argument).
But I never got any other value than `MaxBitrate`, i.e. `MaxStaticBitrate` or `MaxStaticMusicBitrate` :man_shrugging: 

Fun fact, the other `GetBitrateForDirectPlayCheck` call also has `true` as the last argument.

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -615,7 +605,7 @@ private StreamInfo BuildVideoItem(MediaSourceInfo item, VideoOptions options)
             var videoStream = item.VideoStream;
 
             // TODO: This doesn't account for situations where the device is able to handle the media's bitrate, but the connection isn't fast enough
```
dmitrylyzo, line 607: Isn't this `TODO` solved by this change?

MediaBrowser.Model/Dlna/StreamBuilder.cs
crobibero: Yep that was my copy&paste error, good catch

nyanmisaka (approved):

1337joe (approved):

PR #7486, opened 2022-03-22
Move NormalizeDlnaMediaUrl logic inside the StreamInfo::ToUrl function.

marius-luca-87:
Moved the NormalizeDlnaMediaUrl logic inside StreamInfo::ToUrl function as the query parameters doesn't need to be applied when using PlayMethod.DirectPlay since the url is an actual file path.

Fix ArgumentOutOfRangeException present in
`var filename = url.Substring(0, url.IndexOf('?', StringComparison.Ordinal));`
when using PlayMethod.DirectPlay (url is a file path and has no '?').

### DIFF
--- a/Emby.Dlna/Didl/DidlBuilder.cs
+++ b/Emby.Dlna/Didl/DidlBuilder.cs
@@ -79,11 +79,6 @@ namespace Emby.Dlna.Didl
             _libraryManager = libraryManager;
         }
 
-        public static string NormalizeDlnaMediaUrl(string url)
-        {
-            return url + "&dlnaheaders=true";
-        }
-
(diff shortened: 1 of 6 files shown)


cvium:
Why is the url a file path?

marius-luca-87:
ToUrl returns [MediaSource.Path](https://github.com/jellyfin/jellyfin/blob/4fce423b47cb55e509c6320c0dd8a074a5174b50/MediaBrowser.Model/Dlna/StreamInfo.cs#L603) when PlayMethod == PlayMethod.DirectPlay .

Also note that since [#7325](https://github.com/jellyfin/jellyfin/pull/7325) when trying to browse DLNA with VLC PlayMethod  is now PlayMethod.DirectPlay instead of PlayMethod.DirectStream.

cvium:
DirectPlay should not return the file path. That is not _our_ intention at least. I don't think this PR is the correct fix.

PR #7497, opened 2022-03-27
Allow ExternalId to be used on all BaseItems

revam:
Allow the `ExternalId` field to be set for all `BaseItem`s. It is currently only used by the Live TV featues. My use case would be to use the field on `Series` entries as it would be a great escape hatch for plugins to forcefully merge entries without (invalidly) using the ImDB id — which I'm currently doing — or the TvDB id, which are the only other two ids besides the `ExternalId` allowed for merging series together. This behavior is hard coded in [the Series entry](https://github.com/jellyfin/jellyfin/blob/a26509a98a07889c9c947e8b3992ab00d6249960/MediaBrowser.Controller/Entities/TV/Series.cs#L177-L185) which results in me resulting to a hacky spaghetti :spaghetti: workaround like [this](https://github.com/ShokoAnime/Shokofin/blob/267e41249864cbdec8c178cce334fc7eea7b1344/Shokofin/Providers/SeriesProvider.cs#L172-L173) in order to merge the series before this change.

**Changes**
Change the `GetUserDataKeys` method for the `BaseItem` to allow the `ExternalId` to be used as an escape hatch for plugins using an "_unsupported_" id to merge series. This change makes the code more error prone if used incorrectly but also gives plugins (and the core) more control over the merging process if used correctly — as there is no collision detection if multiple plugins (and/or other providers) use the `ExternalId` so it would be up to each plugin to make sure it plays nice with the other plugins (and/or providers).

**Issues**

### DIFF
--- a/MediaBrowser.Controller/Entities/BaseItem.cs
+++ b/MediaBrowser.Controller/Entities/BaseItem.cs
@@ -1425,12 +1425,9 @@ namespace MediaBrowser.Controller.Entities
         {
             var list = new List<string>();
 
-            if (SourceType == SourceType.Channel)
+            if (!string.IsNullOrEmpty(ExternalId))
             {
-                if (!string.IsNullOrEmpty(ExternalId))
-                {
-                    list.Add(ExternalId);
-                }
+                list.Add(ExternalId);
             }
 
             list.Add(Id.ToString());


revam:
Just for the record, but if anyone have a better solution (that's not hard coded) to be able to merge series (and other entries for that matter) without relying on the TvDB, ImDB, and/or TMDB ids then I will gladly close this PR in favor of the other solution.

revam:
Anyone?

crobibero:
Right now we are focusing on bugfixes for the 10.8 release so other PRs have been lower priority

revam:
Stupid idea. Didn't work as intended. Will work a new PR with a better solution that's more universal and less hacky.

PR #7499, opened 2022-03-27
Fix QSV hardware encoding / decoding in official Dockerfile

ocervell:
fixes https://github.com/jellyfin/jellyfin/issues/5993

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,7 +31,7 @@ ARG LEVEL_ZERO_VERSION=1.3.22549
 # mesa-va-drivers: needed for AMD VAAPI. Mesa >= 20.1 is required for HEVC transcoding.
 # curl: healthcheck
 RUN apt-get update \
- && apt-get install --no-install-recommends --no-install-suggests -y ca-certificates gnupg wget apt-transport-https curl \
+ && apt-get install --no-install-recommends --no-install-suggests -y ca-certificates gnupg gpg-agent vainfo wget apt-transport-https curl \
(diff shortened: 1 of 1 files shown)


crobibero:
I don't personally have hardware that can use QSV, but as far as I know Jellyfin's ffmpeg build already provides the drivers required for iHD.

@Shadowghost 
@nyanmisaka

martadinata666:
As the user use QSV, i can it work well `i5-8400T`, cant really tell on 10.7, as i build 10.8 from git master. Maybe some package missing on 10.7

dlicois:
see #5993 current stable docker image doesn't work because of wrong jellyfin-ffmpeg.

Latest master docker might have fixed it already, but there is no stable image with updated jellyfin-ffmpeg.

nyanmisaka (changes requested):
The iHD driver and vainfo are already a part of jellyfin-ffmpeg since [4.4.1-2](https://github.com/jellyfin/jellyfin-ffmpeg/releases/tag/v4.4.1-2).
https://github.com/jellyfin/jellyfin-ffmpeg/pull/89
https://github.com/jellyfin/jellyfin-ffmpeg/pull/93

martadinata666:
Then should be 10.7.7 -> 10.7.8 with upgraded jellyfin? Dunno which one better, either upgrade jellyfin or add non-free driver

crobibero:
There will be no 10.7.8. The next release is 10.8.

Issue #7505, opened 2022-03-28
[Issue]: Unable to load user permission


crobibero:
### Please describe your bug

https://www.reddit.com/r/jellyfin/comments/tpqrfm/the_first_beta_for_1080_is_up_were_getting_there/i2durtt/

### Jellyfin Version

Other

### if other:

10.8.0-beta.1

### Environment

```markdown
- OS:
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

```shell
today at 6:24:52 PM[18:24:52] [ERR] [26] Jellyfin.Server.Middleware.ExceptionMiddleware: Error processing request. URL GET /users/public.
today at 6:24:52 PMSystem.InvalidOperationException: Sequence contains no matching element
today at 6:24:52 PM   at System.Linq.ThrowHelper.ThrowNoMatchException()
today at 6:24:52 PM   at System.Linq.Enumerable.First[TSource](IEnumerable`1 source, Func`2 predicate)
today at 6:24:52 PM   at Jellyfin.Data.Entities.User.HasPermission(PermissionKind kind)
today at 6:24:52 PM   at Jellyfin.Api.Controllers.UserController.<>c__DisplayClass25_0.<Get>b__0(User i)
today at 6:24:52 PM   at System.Linq.Utilities.<>c__DisplayClass1_0`1.<CombinePredicates>b__0(TSource x)
today at 6:24:52 PM   at System.Linq.Enumerable.WhereEnumerableIterator`1.ToArray()
today at 6:24:52 PM   at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
today at 6:24:52 PM   at System.Linq.OrderedEnumerable`1.GetEnumerator()+MoveNext()
today at 6:24:52 PM   at System.Linq.Enumerable.SelectIPartitionIterator`2.MoveNext()
today at 6:24:52 PM   at System.Text.Json.Serialization.Converters.IEnumerableDefaultConverter`2.OnWriteResume(Utf8JsonWriter writer, TCollection value, JsonSerializerOptions options, WriteStack& state)
today at 6:24:52 PM   at System.Text.Json.Serialization.JsonCollectionConverter`2.OnTryWrite(Utf8JsonWriter writer, TCollection value, JsonSerializerOptions options, WriteStack& state)
[...]

Issue #6109, opened 2021-05-26
GET /users/public -> 500, User.HasPermissions "Sequence contains no matching element" crash


chadmiller:
Starting a stopped Jellyfin instance in an up-to-date docker image from official "jellyfin/jellyfin" docker repo on Ubuntu 20.04-ish.

Initial log-in-screen asks to select server. (That's unlike first start!) Upon clicking on the only choice, and redirected to `/web/index.html`, the we page doesn't render anything. Opening browser tools shows network listing. The only error item is HTTP 500 from `/users/public`.

No unusual request headers are sent. Response is empty document after HTTP 500.

Log says `Jellyfin version: 10.7.5` ... 

```
System.InvalidOperationException: Sequence contains no matching element
   at System.Linq.ThrowHelper.ThrowNoMatchException()
   at System.Linq.Enumerable.First[TSource](IEnumerable`1 source, Func`2 predicate)
   at Jellyfin.Data.Entities.User.HasPermission(PermissionKind kind)
   at Jellyfin.Api.Controllers.UserController.<>c__DisplayClass23_0.<Get>b__0(User i)
   at System.Linq.Utilities.<>c__DisplayClass1_0`1.<CombinePredicates>b__0(TSource x)
   at System.Linq.Enumerable.WhereEnumerableIterator`1.ToArray()
   ...
```

chadmiller:
Don't be a punk, 'bot.

cvium:
Did you put the /data folder in a volume or bind mount?

chadmiller:
My docker-compose, using image "jellyfin/jellyfin" has volumes for `/media` (ro) , `/config` , and `/cache` .

Beneath the volume for `/config/` , there is a `data/` directory. data itself isn't unusual or named anywhere.

chadmiller:
Note, this doesn't happen presently. I suspect it has something to do with name resolution (browser to mdns-advertized name?) or a failure cascading off of that.

I am reporting this because there's probably a test higher that could avoid the crashing code path, like an emptiness test before first() .

covertneko:
Just adding that this happens to me almost every time I update my jellyfin docker container. No idea what's going on, but clearing my cache, trying different browsers, different machines, and connecting via IP all yield the same result until I roll back.

Salamandar:
Trying to use 10.8.0-alpha5 triggered this issue. I could not solve this.

marine1988:
i have same problem now i cant get back to stable version :(

[...]

PR #7506, opened 2022-03-28
Safely get/set User permission/preference

crobibero:
Fixes https://github.com/jellyfin/jellyfin/issues/7505
Fixes https://github.com/jellyfin/jellyfin/issues/6109

Note: I'm unable to reproduce the error but this should avoid the issue entirely.

### DIFF
--- a/Jellyfin.Data/Entities/User.cs
+++ b/Jellyfin.Data/Entities/User.cs
@@ -362,7 +362,7 @@ namespace Jellyfin.Data.Entities
         /// <returns><c>True</c> if the user has the specified permission.</returns>
         public bool HasPermission(PermissionKind kind)
         {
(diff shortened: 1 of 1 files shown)


Jellyfin.Data/Entities/User.cs
```diff
@@ -372,7 +372,21 @@ public bool HasPermission(PermissionKind kind)
         /// <param name="value">The value to set.</param>
         public void SetPermission(PermissionKind kind, bool value)
         {
-            Permissions.First(p => p.Kind == kind).Value = value;
+            var permissionUpdated = false;
+            foreach (var currentPermissions in Permissions)
```
cvium, line 376: Why not just FirstOrDefault?

cvium (approved):

PR #7507, opened 2022-03-28
Fix StudioImageProvider

crobibero:


thornbill (approved):
:+1: I have confirmed that this fixes setting a url in the studio image plugin settings.

PR #7521, opened 2022-03-29
Add fallback for image downloads with bad reported MediaType

1337joe:
`ImageSaver` uses `contentType` to determine the file extension to use when saving the image.

Some IPTV/EPG providers serve images that report as types that don't map to extensions (such as `binary/octet-stream`), resulting in no channel icon and logged exceptions every time one is downloaded and can't be saved.

This may be a more general fix for the tvheadend workaround on line 177, but I have been unable to reproduce that case on my dev machine to verify.

**Changes**
- Add fallback for image downloads with bad MediaType


crobibero (approved):

thornbill (approved):
I have tested this PR and confirm it fixes the issue for me. :tada:

PR #7527, opened 2022-03-30
Remove video file from file list before processing external files

Shadowghost:


1337joe:
This is just so the `files.Any()` call won't find any files if only the video file is in the directory?

Shadowghost:
We could leave it like it is and let the movie be dropped within the ExternalPathParser. Not sure what is cleaner tbh.

1337joe:
Like it is the `files.Any` check is useless because it will always find the movie file. I'm good with changing it, just wanted to make sure I understood the reasoning.

1337joe (approved):

crobibero (approved):

PR #7529, opened 2022-03-31
Properly handle stream addition and removal for strm use cases

Shadowghost:
Strm files are special:
* preserve non-external streams between scans
* add external streams before adding internal streams
* always recalculate the stream index of the selected stream within its source file 
* fix transcoding if external subtitle or audio isn't a single stream but a a stream in a container

We need to add the external streams before adding the internals because otherwise the index submitted by e.g. stream selection on jf-web would be wrong if the initial scan of the file did not already populate the internal streams or if the file changed. For strm files that's always the case if the file was never played before since probe on scan is skipped for linked remote files.

### DIFF
--- a/Emby.Server.Implementations/Library/MediaSourceManager.cs
+++ b/Emby.Server.Implementations/Library/MediaSourceManager.cs
@@ -151,7 +151,9 @@ namespace Emby.Server.Implementations.Library
         {
             var mediaSources = GetStaticMediaSources(item, enablePathSubstitution, user);
 
-            if (allowMediaProbe && mediaSources[0].Type != MediaSourceType.Placeholder && !mediaSources[0].MediaStreams.Any(i => i.Type == MediaStreamType.Audio || i.Type == MediaStreamType.Video))
+            if (allowMediaProbe && mediaSources[0].Type != MediaSourceType.Placeholder
+                && (item.MediaType == MediaType.Video && !mediaSources[0].MediaStreams.Any(i => i.Type == MediaStreamType.Video)
+                || item.MediaType == MediaType.Audio && !mediaSources[0].MediaStreams.Any(i => i.Type == MediaStreamType.Audio)))
             {
                 await item.RefreshMetadata(
                     new MetadataRefreshOptions(_directoryService)
--- a/Emby.Server.Implementations/Library/MediaSourceManager.cs
+++ b/Emby.Server.Implementations/Library/MediaSourceManager.cs
@@ -153,9 +153,9 @@ namespace Emby.Server.Implementations.Library
 
             // If file is strm or main media stream is missing, force a metadata refresh with remote probing
             if (allowMediaProbe && mediaSources[0].Type != MediaSourceType.Placeholder
-                && (item.Path.EndsWith(".strm")
-                || (item.MediaType == MediaType.Video && !mediaSources[0].MediaStreams.Any(i => i.Type == MediaStreamType.Video))
-                || (item.MediaType == MediaType.Audio && !mediaSources[0].MediaStreams.Any(i => i.Type == MediaStreamType.Audio))))
+                && (item.Path.EndsWith(".strm", StringComparison.OrdinalIgnoreCase)
+                    || (item.MediaType == MediaType.Video && !mediaSources[0].MediaStreams.Any(i => i.Type == MediaStreamType.Video))
+                    || (item.MediaType == MediaType.Audio && !mediaSources[0].MediaStreams.Any(i => i.Type == MediaStreamType.Audio))))
             {
                 await item.RefreshMetadata(
                     new MetadataRefreshOptions(_directoryService)
--- a/Emby.Server.Implementations/Library/MediaSourceManager.cs
+++ b/Emby.Server.Implementations/Library/MediaSourceManager.cs
@@ -151,9 +151,11 @@ namespace Emby.Server.Implementations.Library
         {
             var mediaSources = GetStaticMediaSources(item, enablePathSubstitution, user);
 
+            // If file is strm or main media stream is missing, force a metadata refresh with remote probing
             if (allowMediaProbe && mediaSources[0].Type != MediaSourceType.Placeholder
-                && (item.MediaType == MediaType.Video && !mediaSources[0].MediaStreams.Any(i => i.Type == MediaStreamType.Video)
-                || item.MediaType == MediaType.Audio && !mediaSources[0].MediaStreams.Any(i => i.Type == MediaStreamType.Audio)))
+                && (item.Path.EndsWith(".strm")
+                || (item.MediaType == MediaType.Video && !mediaSources[0].MediaStreams.Any(i => i.Type == MediaStreamType.Video))
+                || (item.MediaType == MediaType.Audio && !mediaSources[0].MediaStreams.Any(i => i.Type == MediaStreamType.Audio))))
             {
                 await item.RefreshMetadata(
                     new MetadataRefreshOptions(_directoryService)
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -2228,14 +2228,14 @@ namespace MediaBrowser.Controller.MediaEncoding
             }
 
             var args = string.Empty;
-            var numberOfExternalStreams = state.MediaSource.MediaStreams.Where(stream => stream.IsExternal == true).Count();
+            int videoStreamIndex = state.MediaSource.MediaStreams.Where(i => i.Path == state.VideoStream.Path).ToList().IndexOf(state.VideoStream);
 
             if (state.VideoStream != null)
             {
                 args += string.Format(
                     CultureInfo.InvariantCulture,
                     "-map 0:{0}",
-                    state.VideoStream.Index - numberOfExternalStreams);
+                    videoStreamIndex);
             }
             else
             {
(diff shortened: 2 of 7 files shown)


crobibero (approved):

MediaBrowser.Providers/MediaInfo/FFProbeVideoInfo.cs
```diff
@@ -213,14 +215,30 @@ public class FFProbeVideoInfo
...
+                    }
+
+                }
                 mediaAttachments = Array.Empty<MediaAttachment>();
                 chapters = Array.Empty<ChapterInfo>();
             }
 
-            await AddExternalSubtitlesAsync(video, mediaStreams, options, cancellationToken).ConfigureAwait(false);
+            // Only fetch and add external media if we have a video stream
+            if (currentMediaStreams != null && currentMediaStreams.Find(i => i.Type == MediaStreamType.Video) != null)
```
whiteowl3, line 236: I don't understand the purpose of this check.  It creates an undesirable situation where external subtitles are not revealed on the first playback attempt.  Reverting this change works better in my tests.

MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
```diff
@@ -195,7 +195,7 @@ private async Task<Stream> GetSubtitleStream(SubtitleInfo fileInfo, Cancellation
             MediaStream subtitleStream,
             CancellationToken cancellationToken)
         {
-            if (!subtitleStream.IsExternal)
+            if (!subtitleStream.IsExternal || subtitleStream.Path.EndsWith(".mks"))
```
crobibero, line 198:
```suggestion
            if (!subtitleStream.IsExternal || subtitleStream.Path.EndsWith(".mks", StringComparison.OrdinalIgnoreCase))
```

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -2226,13 +2226,14 @@ public string GetMapArgs(EncodingJobInfo state)
             }
 
             var args = string.Empty;
+            int videoStreamIndex = state.MediaSource.MediaStreams.Where(i => i.Path == state.VideoStream.Path).ToList().IndexOf(state.VideoStream);
```
cvium, line 2229: Given that you are now using this so much throughout the class, I think it should be a static method eg:

```C#
        private static int FindIndex(IReadOnlyList<MediaStream> mediaStreams, MediaStream streamToFind)
        {
            var index = 0;
            var length = mediaStreams.Count;
            for (var i = 0; i < length; i++)
            {
                var currentMediaStream = mediaStreams[i];
                if (currentMediaStream.Path == streamToFind.Path)
                {
                    index++;
                }

                if (currentMediaStream == streamToFind)
                {
                    return index;
                }
            }

            return -1;
        }
```

MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
```diff
@@ -411,6 +411,19 @@ public string GetInputArgument(string inputFile, MediaSourceInfo mediaSource)
...
 
+        /// <summary>
+        /// Gets the input argument for an external subtitle file.
+        /// </summary>
+        /// <param name="inputFile">The input file.</param>
+        /// <returns>System.String.</returns>
+        /// <exception cref="ArgumentException">Unrecognized InputType.</exception>
+        public string GetExternalSubtitleInputArgument(string inputFile)
+        {
+            var prefix = "file";
```
cvium, line 422:
```suggestion
            const string Prefix = "file";
```

MediaBrowser.Providers/MediaInfo/FFProbeVideoInfo.cs
```diff
@@ -173,9 +173,23 @@ public class FFProbeVideoInfo
...
             ChapterInfo[] chapters;
 
+            mediaStreams = new List<MediaStream>();
+
+            // Add external streams before adding the streams from the file to preserve stream IDs on remote videos
+            await AddExternalSubtitlesAsync(video, mediaStreams, options, cancellationToken).ConfigureAwait(false);
+
+            await AddExternalAudioAsync(video, mediaStreams, options, cancellationToken).ConfigureAwait(false);
+
+            var startIndex = mediaStreams.Count == 0 ? 0 : (mediaStreams.Select(i => i.Index).Max() + 1);
```
cvium, line 183: Don't need the Select
```suggestion
            var startIndex = mediaStreams.Count == 0 ? 0 : (mediaStreams.Max(i => i.Index) + 1);
```

MediaBrowser.Providers/MediaInfo/FFProbeVideoInfo.cs
```diff
@@ -213,15 +227,16 @@ public class FFProbeVideoInfo
             }
             else
             {
-                mediaStreams = new List<MediaStream>();
+                var nonExternalMediaStreams = video.GetMediaStreams().Where(i => !i.IsExternal);
```
cvium, line 230: The `!i.IsExternal` check should just be in the foreach imo

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -2226,13 +2226,14 @@ public string GetMapArgs(EncodingJobInfo state)
             }
 
             var args = string.Empty;
+            int videoStreamIndex = FindIndex(state.MediaSource.MediaStreams, state.VideoStream);
```
cvium, line 2229: Shouldn't this be inside the if? You're not using it anywhere else afaict

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -2242,24 +2243,25 @@ public string GetMapArgs(EncodingJobInfo state)
 
             if (state.AudioStream != null)
             {
+                int audioStreamIndex = FindIndex(state.MediaSource.MediaStreams, state.AudioStream);
                 if (state.AudioStream.IsExternal)
                 {
                     bool hasExternalGraphicsSubs = state.SubtitleStream != null && state.SubtitleStream.IsExternal && !state.SubtitleStream.IsTextSubtitleStream;
                     int externalAudioMapIndex = hasExternalGraphicsSubs ? 2 : 1;
-                    int externalAudioStream = state.MediaSource.MediaStreams.Where(i => i.Path == state.AudioStream.Path).ToList().IndexOf(state.AudioStream);
+                    int externalAudioStreamIndex = state.MediaSource.MediaStreams.Where(i => i.Path == state.AudioStream.Path).ToList().IndexOf(state.AudioStream);
```
cvium, line 2251: Is this used?

MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
```diff
@@ -511,12 +511,21 @@ private async Task ConvertTextSubtitleToSrtInternal(string inputPath, string lan
 
             await semaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
 
+            var subtitleStreamIndex = mediaSource.MediaStreams.Where(i => i.Path == subtitleStream.Path).ToList().IndexOf(subtitleStream);
```
cvium, line 514: Can't we place `FindIndex` somewhere that is visible for both projects?

Shadowghost: Somewhere in `Jellyfin.Extensions` maybe or do we have a internal untility namespace?

cvium: It's very specific to MediaStreams and Jellyfin.Extensions is for more generic stuff. Just make it public in EncodingHelper and use it.

Emby.Server.Implementations/Library/MediaSourceManager.cs
```diff
@@ -151,7 +151,11 @@ public async Task<List<MediaSourceInfo>> GetPlaybackMediaSources(BaseItem item,
         {
             var mediaSources = GetStaticMediaSources(item, enablePathSubstitution, user);
 
-            if (allowMediaProbe && mediaSources[0].Type != MediaSourceType.Placeholder && !mediaSources[0].MediaStreams.Any(i => i.Type == MediaStreamType.Audio || i.Type == MediaStreamType.Video))
+            // If file is strm or main media stream is missing, force a metadata refresh with remote probing
+            if (allowMediaProbe && mediaSources[0].Type != MediaSourceType.Placeholder
+                && (item.Path.EndsWith(".strm", StringComparison.OrdinalIgnoreCase)
+                || (item.MediaType == MediaType.Video && !mediaSources[0].MediaStreams.Any(i => i.Type == MediaStreamType.Video))
+                || (item.MediaType == MediaType.Audio && !mediaSources[0].MediaStreams.Any(i => i.Type == MediaStreamType.Audio))))
```
cvium, line 158:
```suggestion
                && (item.Path.EndsWith(".strm", StringComparison.OrdinalIgnoreCase)
                    || (item.MediaType == MediaType.Video && !mediaSources[0].MediaStreams.Any(i => i.Type == MediaStreamType.Video))
                    || (item.MediaType == MediaType.Audio && !mediaSources[0].MediaStreams.Any(i => i.Type == MediaStreamType.Audio))))
```

cvium (approved):

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -5389,6 +5397,29 @@ public string GetProgressiveAudioFullCommandLine(EncodingJobInfo state, Encoding
...
+                if (currentMediaStream == streamToFind)
+                {
+                    return index;
+                }
+
+                if (currentMediaStream.Path == streamToFind.Path)
+                {
+                    index++;
+                }
+
```
crobibero, line 5417:
```suggestion
```

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -5389,6 +5397,29 @@ public string GetProgressiveAudioFullCommandLine(EncodingJobInfo state, Encoding
...
+            for (var i = 0; i < length; i++)
+            {
+                var currentMediaStream = mediaStreams[i];
+                if (currentMediaStream == streamToFind)
+                {
+                    return index;
+                }
+
+                if (currentMediaStream.Path == streamToFind.Path)
```
crobibero, line 5413:
```suggestion
                if (string.Equals(currentMediaStream.Path, streamToFind.Path, StringComparison.Ordinal))
```

crobibero (approved):

PR #7557, opened 2022-04-04
Fix PGS burn-in on certain iGPU such as Iris Plus 655

nyanmisaka:
**Changes**
- Use a larger frame pool as per https://trac.ffmpeg.org/wiki/Hardware/QuickSync

**Issues**
Fixes https://github.com/jellyfin/jellyfin-ffmpeg/issues/95

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -3381,7 +3381,7 @@ namespace MediaBrowser.Controller.MediaEncoding
                     }
 
                     // qsv requires a fixed pool size.
-                    // default to 64 otherwise if will fails on certain iGPU.
+                    // default to 64 otherwise it will fail on certain iGPU.
                     subFilters.Add("hwupload=extra_hw_frames=64");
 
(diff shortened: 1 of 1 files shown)


MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -3381,7 +3381,8 @@ public static string GetHwTonemapFilter(EncodingOptions options, string hwTonema
                     }
 
                     // qsv requires a fixed pool size.
-                    subFilters.Add("hwupload=extra_hw_frames=32");
+                    // default to 64 otherwise if will fails on certain iGPU.
```
Shadowghost, line 3384:
```suggestion
                    // default to 64 otherwise it will fail on certain iGPU.
```

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -3589,7 +3590,8 @@ public static string GetHwTonemapFilter(EncodingOptions options, string hwTonema
...
```
Shadowghost, line 3593:
```suggestion
                    // default to 64 otherwise it will fail on certain iGPU.
```

Shadowghost (approved):

Bond-009 (approved):

PR #7561, opened 2022-04-05
Fix NamedConfiguration API Generation

### LINKED ISSUE #7536: [Issue]: OpenApi Definition for UpdateNamedConfiguration does not identify a Request Body
### Please describe your bug

A cursory reading of the Swashbuckle docs and the UpdateNamedConfiguration method itself in ConfigurationController.cs leads me to believe that this is caused by the manual processing of the request body in the method rather than defining it as a parameter with the [FromBody] annotation.

The method appears to be dynamically determining the type of the body based on the named configuration - for example, the "dlna" configuration casts the body to `DlnaOptions`. I'm not certain how this could be remedied in a way that's friendly to both OpenAPI and the dynamic behavior, but I have a couple ideas (possibly naive ones since Dotnet/C# isn't my specialty):

1. The half-baked method: parameterize it as some generic JSON-compatible type (string/base JsonObject-ish class) that OpenAPI can accept well enough to generate code that allows for a body to be passed in. This solves the problem but is terrible in the way of type safety. Not much worse than it is right now though.

2. The inheritance route: create something like `INamedConfiguration` that `DlnaOptions` and the other named configuration POCOs (is that a term like POJO is in Java?) can inherit from to use as the type for the paremeter; I'm not sure how happy this would make Swashbuckle. In the best case, it would create type definitions for all of the subclasses and we'd be set to use them as needed in clients. At its worst, it would create some garbage/blank definition for the spec and we'd have to either find a way to get Swashbuckle to attach type definitions for the subclasses or define them manually in the client implementations (not ideal).

It's also possible that both of my suggestions aren't the best route and there's an even better way to handle this. Suggestions are welcome - I'd like to try tackling this, but I'm a Groovy/Grails/Micronaut developer by day so it'll take me some time to get acquainted with the dev environment to try my approaches to see how they fit.

This issue affects all clients utilizing the OpenAPI output such as in-org projects like jellyfin-client-axios and consumers of such API clients like jellyfin-vue. I came across this issue while trying to implement DLNA settings in jellyfin-vue as a learning experience.

### Jellyfin Version

10.8.0

### Environment

```markdown
- OS: Any/All
- Virtualization: N/A
- Clients: OpenAPI Consumers
- Browser: N/A
- FFmpeg Version: N/A
- Playback Method: N/A
- Hardware Acceleration: N/A
- Plugins: None, but this may raise questions about how plugins define their types.
- Reverse Proxy: N/A
- Base URL: N/A
- Networking: N/A
- Storage: N/A
```

### Code of Conduct

DMouse10462:
**Changes**
- Parameterize request body (so it is visible to Swashbuckle)
- Manually add types accepted by configurations to OpenAPI documentation.

Areas that could use some feedback:
- The controller method is no longer async - is this a problem? Is there a better way to accomplish this? I'm a Java EE developer by day, so the C#/dotnet style of handling this is a bit foreign to me. I'm happy to implement suggestions.
- The manual addition of the types to OpenAPI seems a bit shady, but even with @crobibero's suggestion of using Swashbuckle's polymorphism/inheritance features I can't seem to find a solution that allows me to specify an interface as the parameter type for the request body. This step seems to be a prerequisite to actually utilizing the polymorphism/inheritance features mentioned.

**Issues**
Fixes #7536

### DIFF
--- a/Emby.Server.Implementations/AppBase/BaseConfigurationManager.cs
+++ b/Emby.Server.Implementations/AppBase/BaseConfigurationManager.cs
@@ -399,6 +399,12 @@ namespace Emby.Server.Implementations.AppBase
         }
 
         /// <inheritdoc />
+        public ConfigurationStore[] GetConfigurationStores()
+        {
+            return _configurationStores;
+        }
+
+        /// <inheritdoc />
         public Type GetConfigurationType(string key)
         {
             return GetConfigurationStore(key)
--- a/Jellyfin.Api/Controllers/ConfigurationController.cs
+++ b/Jellyfin.Api/Controllers/ConfigurationController.cs
@@ -86,21 +86,23 @@ namespace Jellyfin.Api.Controllers
         /// Updates named configuration.
         /// </summary>
         /// <param name="key">Configuration key.</param>
+        /// <param name="configuration">Configuration.</param>
         /// <response code="204">Named configuration updated.</response>
         /// <returns>Update status.</returns>
         [HttpPost("Configuration/{key}")]
         [Authorize(Policy = Policies.RequiresElevation)]
         [ProducesResponseType(StatusCodes.Status204NoContent)]
-        public async Task<ActionResult> UpdateNamedConfiguration([FromRoute, Required] string key)
+        public ActionResult UpdateNamedConfiguration([FromRoute, Required] string key, [FromBody, Required] JsonDocument configuration)
         {
             var configurationType = _configurationManager.GetConfigurationType(key);
-            var configuration = await JsonSerializer.DeserializeAsync(Request.Body, configurationType, _serializerOptions).ConfigureAwait(false);
-            if (configuration == null)
+            var deserializedConfiguration = configuration.Deserialize(configurationType, _serializerOptions);
+
+            if (deserializedConfiguration == null)
             {
                 throw new ArgumentException("Body doesn't contain a valid configuration");
             }
 
-            _configurationManager.SaveConfiguration(key, configuration);
+            _configurationManager.SaveConfiguration(key, deserializedConfiguration);
             return NoContent();
         }
 
--- a/Jellyfin.Server/Filters/AdditionalModelFilter.cs
+++ b/Jellyfin.Server/Filters/AdditionalModelFilter.cs
@@ -1,4 +1,5 @@
-﻿using MediaBrowser.Common.Plugins;
+using MediaBrowser.Common.Plugins;
+using MediaBrowser.Controller.Configuration;
 using MediaBrowser.Controller.LiveTv;
 using MediaBrowser.Model.ApiClient;
 using MediaBrowser.Model.Entities;
@@ -14,6 +15,17 @@ namespace Jellyfin.Server.Filters
     /// </summary>
     public class AdditionalModelFilter : IDocumentFilter
     {
+        private readonly IServerConfigurationManager _serverConfigurationManager;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AdditionalModelFilter"/> class.
+        /// </summary>
+        /// <param name="serverConfigurationManager">Instance of the <see cref="IServerConfigurationManager"/> interface.</param>
+        public AdditionalModelFilter(IServerConfigurationManager serverConfigurationManager)
+        {
+            _serverConfigurationManager = serverConfigurationManager;
+        }
+
         /// <inheritdoc />
         public void Apply(OpenApiDocument swaggerDoc, DocumentFilterContext context)
         {
@@ -29,6 +41,11 @@ namespace Jellyfin.Server.Filters
 
             context.SchemaGenerator.GenerateSchema(typeof(SessionMessageType), context.SchemaRepository);
             context.SchemaGenerator.GenerateSchema(typeof(ServerDiscoveryInfo), context.SchemaRepository);
+
+            foreach (var configuration in _serverConfigurationManager.GetConfigurationStores())
+            {
+                context.SchemaGenerator.GenerateSchema(configuration.ConfigurationType, context.SchemaRepository);
+            }
         }
     }
 }
--- a/MediaBrowser.Common/Configuration/IConfigurationManager.cs
+++ b/MediaBrowser.Common/Configuration/IConfigurationManager.cs
@@ -61,6 +61,12 @@ namespace MediaBrowser.Common.Configuration
         object GetConfiguration(string key);
 
         /// <summary>
+        /// Gets the array of coniguration stores.
+        /// </summary>
+        /// <returns>Array of ConfigurationStore.</returns>
+        ConfigurationStore[] GetConfigurationStores();
+
+        /// <summary>
         /// Gets the type of the configuration.
         /// </summary>
         /// <param name="key">The key.</param>


DMouse10462:
Marking as draft again; I'm going to look at other solutions. If the other solutions start breaking things, I'd like to suggest the current solution as a 10.8.x fix to get the APIs working for consumers ASAP while working on a "proper" solution in master with proper typing, inheritance, etc.

DMouse10462:
Re-opening for review against release-10.8.z per [below Matrix message](https://matrix.to/#/!YOoxJKhsHoXZiIHyBG:matrix.org/$ZdtzBkVjlgtYyWlSaj5MfpxBwQCSRfRF7M0b-xJLxqM?via=bonifacelabs.ca&via=t2bot.io&via=matrix.org):

> I want to look for other solutions, but it seems as though anything we do on this front will be quite large (making existing ConfigurationType classes inherit from a base class, reworking controllers, etc.). If the goal is to have less breakage in 10.8.x, I'd suggest making the changes I presented today, backporting to 10.8.x, then working on a more "permanent" solution in master.
>
> The only technical debt (if you want to call it that) with my current fix is the extra injection and GenerateSchema loop in AdditionalModelFilter. This will likely be replaced, not removed, with some inheritance configuration in the future, so this would likely be encountered during a re-work of this exact solution.
>
> The parameterization of the body, even with a generic JsonDocument type, is a net positive for the readability of the code (the signature now represents reality - we're accepting a body here - where before there was hidden functionality in the form of side-effects). That can be further expanded on in the future when we have a better solution, but for now I believe this will provide immediate benefits to consumers of the API, especially those using generated clients like jellyfin-client-axios (jellyfin-vue) which would go from being completely unable to send a body to a state that can be described as "somewhat type unsafe client-side, but functional".
>
> I'll rebase this against 10.8.x and target that branch in the PR before sending it out for review again.

nielsvanvelzen:
The request body now uses an empty schema (instead of none) so not much changes on that part.

Expand for openapi diff:
<details>
```diff
diff --git a/openapi.json b/openapihead.json
index 794c87b..1efd534 100644
--- a/openapi.json
+++ b/openapihead.json
@@ -3804,6 +3804,21 @@
             }
           }
         ],
+        "requestBody": {
+          "description": "Configuration.",
+          "content": {
+            "application/json": {
+              "schema": { }
+            },
+            "text/json": {
+              "schema": { }
+            },
+            "application/*+json": {
+              "schema": { }
+            }
+          },
+          "required": true
+        },
         "responses": {
           "204": {
             "description": "Named configuration updated."
@@ -47112,6 +47127,61 @@
         "additionalProperties": false,
         "description": "Defines the display preferences for any item that supports them (usually Folders)."
       },
+      "DlnaOptions": {
+        "type": "object",
+        "properties": {
+          "EnablePlayTo": {
+            "type": "boolean",
+            "description": "Gets or sets a value indicating whether gets or sets a value to indicate the status of the dlna playTo subsystem."
+          },
+          "EnableServer": {
+            "type": "boolean",
+            "description": "Gets or sets a value indicating whether gets or sets a value to indicate the status of the dlna server subsystem."
+          },
+          "EnableDebugLog": {
+            "type": "boolean",
+            "description": "Gets or sets a value indicating whether detailed dlna server logs are sent to the console/log.\r\nIf the setting \"Emby.Dlna\": \"Debug\" msut be set in logging.default.json for this property to work."
+          },
+          "EnablePlayToTracing": {
+            "type": "boolean",
+            "description": "Gets or sets a value indicating whether whether detailed playTo debug logs are sent to the console/log.\r\nIf the setting \"Emby.Dlna.PlayTo\": \"Debug\" msut be set in logging.default.json for this property to work."
+          },
+          "ClientDiscoveryIntervalSeconds": {
+            "type": "integer",
+            "description": "Gets or sets the ssdp client discovery interval time (in seconds).\r\nThis is the time after which the server will send a ssdp search request.",
+            "format": "int32"
+          },
+          "AliveMessageIntervalSeconds": {
+            "type": "integer",
+            "description": "Gets or sets the frequency at which ssdp alive notifications are transmitted.",
+            "format": "int32"
+          },
+          "BlastAliveMessageIntervalSeconds": {
+            "type": "integer",
+            "description": "Gets or sets the frequency at which ssdp alive notifications are transmitted. MIGRATING - TO BE REMOVED ONCE WEB HAS BEEN ALTERED.",
+            "format": "int32"
+          },
+          "DefaultUserId": {
+            "type": "string",
+            "description": "Gets or sets the default user account that the dlna server uses.",
+            "nullable": true
+          },
+          "AutoCreatePlayToProfiles": {
+            "type": "boolean",
+            "description": "Gets or sets a value indicating whether playTo device profiles should be created."
+          },
+          "BlastAliveMessages": {
+            "type": "boolean",
+            "description": "Gets or sets a value indicating whether to blast alive messages."
+          },
+          "SendOnlyMatchedHost": {
+            "type": "boolean",
+            "description": "gets or sets a value indicating whether to send only matched host."
+          }
+        },
+        "additionalProperties": false,
+        "description": "The DlnaOptions class contains the user definable parameters for the dlna subsystems."
+      },
       "DlnaProfileType": {
         "enum": [
           "Audio",
@@ -47154,6 +47224,150 @@
         ],
         "type": "string"
       },
+      "EncodingOptions": {
+        "type": "object",
+        "properties": {
+          "EncodingThreadCount": {
+            "type": "integer",
+            "format": "int32"
+          },
+          "TranscodingTempPath": {
+            "type": "string",
+            "nullable": true
+          },
+          "FallbackFontPath": {
+            "type": "string",
+            "nullable": true
+          },
+          "EnableFallbackFont": {
+            "type": "boolean"
+          },
+          "DownMixAudioBoost": {
+            "type": "number",
+            "format": "double"
+          },
+          "MaxMuxingQueueSize": {
+            "type": "integer",
+            "format": "int32"
+          },
+          "EnableThrottling": {
+            "type": "boolean"
+          },
+          "ThrottleDelaySeconds": {
+            "type": "integer",
+            "format": "int32"
+          },
+          "HardwareAccelerationType": {
+            "type": "string",
+            "nullable": true
+          },
+          "EncoderAppPath": {
+            "type": "string",
+            "description": "Gets or sets the FFmpeg path as set by the user via the UI.",
+            "nullable": true
+          },
+          "EncoderAppPathDisplay": {
+            "type": "string",
+            "description": "Gets or sets the current FFmpeg path being used by the system and displayed on the transcode page.",
+            "nullable": true
+          },
+          "VaapiDevice": {
+            "type": "string",
+            "nullable": true
+          },
+          "EnableTonemapping": {
+            "type": "boolean"
+          },
+          "EnableVppTonemapping": {
+            "type": "boolean"
+          },
+          "TonemappingAlgorithm": {
+            "type": "string",
+            "nullable": true
+          },
+          "TonemappingRange": {
+            "type": "string",
+            "nullable": true
+          },
+          "TonemappingDesat": {
+            "type": "number",
+            "format": "double"
+          },
+          "TonemappingThreshold": {
+            "type": "number",
+            "format": "double"
+          },
+          "TonemappingPeak": {
+            "type": "number",
+            "format": "double"
+          },
+          "TonemappingParam": {
+            "type": "number",
+            "format": "double"
+          },
+          "H264Crf": {
+            "type": "integer",
+            "format": "int32"
+          },
+          "H265Crf": {
+            "type": "integer",
+            "format": "int32"
+          },
+          "EncoderPreset": {
+            "type": "string",
+            "nullable": true
+          },
+          "DeinterlaceDoubleRate": {
+            "type": "boolean"
+          },
+          "DeinterlaceMethod": {
+            "type": "string",
+            "nullable": true
+          },
+          "EnableDecodingColorDepth10Hevc": {
+            "type": "boolean"
+          },
+          "EnableDecodingColorDepth10Vp9": {
+            "type": "boolean"
+          },
+          "EnableEnhancedNvdecDecoder": {
+            "type": "boolean"
+          },
+          "PreferSystemNativeHwDecoder": {
+            "type": "boolean"
+          },
+          "EnableIntelLowPowerH264HwEncoder": {
+            "type": "boolean"
+          },
+          "EnableIntelLowPowerHevcHwEncoder": {
+            "type": "boolean"
+          },
+          "EnableHardwareEncoding": {
+            "type": "boolean"
+          },
+          "AllowHevcEncoding": {
+            "type": "boolean"
+          },
+          "EnableSubtitleExtraction": {
+            "type": "boolean"
+          },
+          "HardwareDecodingCodecs": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "nullable": true
+          },
+          "AllowOnDemandMetadataBasedKeyframeExtractionForExtensions": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "nullable": true
+          }
+        },
+        "additionalProperties": false
+      },
       "EndPointInfo": {
         "type": "object",
         "properties": {
@@ -47701,6 +47915,10 @@
         },
         "additionalProperties": false
       },
+      "GuidStringValueTuple": {
+        "type": "object",
+        "additionalProperties": false
+      },
       "HardwareEncodingType": {
         "enum": [
           "AMF",
@@ -48609,6 +48827,72 @@
         },
         "additionalProperties": false
       },
+      "LiveTvOptions": {
+        "type": "object",
+        "properties": {
+          "GuideDays": {
+            "type": "integer",
+            "format": "int32",
+            "nullable": true
+          },
+          "RecordingPath": {
+            "type": "string",
+            "nullable": true
+          },
+          "MovieRecordingPath": {
+            "type": "string",
+            "nullable": true
+          },
+          "SeriesRecordingPath": {
+            "type": "string",
+            "nullable": true
+          },
+          "EnableRecordingSubfolders": {
+            "type": "boolean"
+          },
+          "EnableOriginalAudioWithEncodedRecordings": {
+            "type": "boolean"
+          },
+          "TunerHosts": {
+            "type": "array",
+            "items": {
+              "$ref": "#/components/schemas/TunerHostInfo"
+            },
+            "nullable": true
+          },
+          "ListingProviders": {
+            "type": "array",
+            "items": {
+              "$ref": "#/components/schemas/ListingsProviderInfo"
+            },
+            "nullable": true
+          },
+          "PrePaddingSeconds": {
+            "type": "integer",
+            "format": "int32"
+          },
+          "PostPaddingSeconds": {
+            "type": "integer",
+            "format": "int32"
+          },
+          "MediaLocationsCreated": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "nullable": true
+          },
+          "RecordingPostProcessor": {
+            "type": "string",
+            "nullable": true
+          },
+          "RecordingPostProcessorArguments": {
+            "type": "string",
+            "nullable": true
+          }
+        },
+        "additionalProperties": false
+      },
       "LiveTvServiceInfo": {
         "type": "object",
         "properties": {
@@ -49385,6 +49669,15 @@
         },
         "additionalProperties": false
       },
+      "MetadataConfiguration": {
+        "type": "object",
+        "properties": {
+          "UseFileCreationTimeForDateAdded": {
+            "type": "boolean"
+          }
+        },
+        "additionalProperties": false
+      },
       "MetadataEditorInfo": {
         "type": "object",
         "properties": {
@@ -49502,6 +49795,21 @@
         ],
         "type": "string"
       },
+      "MigrationOptions": {
+        "type": "object",
+        "properties": {
+          "Applied": {
+            "type": "array",
+            "items": {
+              "$ref": "#/components/schemas/GuidStringValueTuple"
+            },
+            "description": "Gets the list of applied migration routine names.",
+            "readOnly": true
+          }
+        },
+        "additionalProperties": false,
+        "description": "Configuration part that holds all migrations that were applied."
+      },
       "MovePlaylistItemRequestDto": {
         "type": "object",
         "properties": {
@@ -49754,6 +50062,172 @@
         },
         "additionalProperties": false
       },
+      "NetworkConfiguration": {
+        "type": "object",
+        "properties": {
+          "RequireHttps": {
+            "type": "boolean",
+            "description": "Gets or sets a value indicating whether the server should force connections over HTTPS."
+          },
+          "CertificatePath": {
+            "type": "string",
+            "description": "Gets or sets the filesystem path of an X.509 certificate to use for SSL."
+          },
+          "CertificatePassword": {
+            "type": "string",
+            "description": "Gets or sets the password required to access the X.509 certificate data in the file specified by Jellyfin.Networking.Configuration.NetworkConfiguration.CertificatePath."
+          },
+          "BaseUrl": {
+            "type": "string",
+            "description": "Gets or sets a value used to specify the URL prefix that your Jellyfin instance can be accessed at."
+          },
+          "PublicHttpsPort": {
+            "type": "integer",
+            "description": "Gets or sets the public HTTPS port.",
+            "format": "int32"
+          },
+          "HttpServerPortNumber": {
+            "type": "integer",
+            "description": "Gets or sets the HTTP server port number.",
+            "format": "int32"
+          },
+          "HttpsPortNumber": {
+            "type": "integer",
+            "description": "Gets or sets the HTTPS server port number.",
+            "format": "int32"
+          },
+          "EnableHttps": {
+            "type": "boolean",
+            "description": "Gets or sets a value indicating whether to use HTTPS."
+          },
+          "PublicPort": {
+            "type": "integer",
+            "description": "Gets or sets the public mapped port.",
+            "format": "int32"
+          },
+          "UPnPCreateHttpPortMap": {
+            "type": "boolean",
+            "description": "Gets or sets a value indicating whether the http port should be mapped as part of UPnP automatic port forwarding."
+          },
+          "UDPPortRange": {
+            "type": "string",
+            "description": "Gets or sets the UDPPortRange."
+          },
+          "EnableIPV6": {
+            "type": "boolean",
+            "description": "Gets or sets a value indicating whether gets or sets IPV6 capability."
+          },
+          "EnableIPV4": {
+            "type": "boolean",
+            "description": "Gets or sets a value indicating whether gets or sets IPV4 capability."
+          },
+          "EnableSSDPTracing": {
+            "type": "boolean",
+            "description": "Gets or sets a value indicating whether detailed SSDP logs are sent to the console/log.\r\n\"Emby.Dlna\": \"Debug\" must be set in logging.default.json for this property to have any effect."
+          },
+          "SSDPTracingFilter": {
+            "type": "string",
+            "description": "Gets or sets the SSDPTracingFilter\r\nGets or sets a value indicating whether an IP address is to be used to filter the detailed ssdp logs that are being sent to the console/log.\r\nIf the setting \"Emby.Dlna\": \"Debug\" msut be set in logging.default.json for this property to work."
+          },
+          "UDPSendCount": {
+            "type": "integer",
+            "description": "Gets or sets the number of times SSDP UDP messages are sent.",
+            "format": "int32"
+          },
+          "UDPSendDelay": {
+            "type": "integer",
+            "description": "Gets or sets the delay between each groups of SSDP messages (in ms).",
+            "format": "int32"
+          },
+          "IgnoreVirtualInterfaces": {
+            "type": "boolean",
+            "description": "Gets or sets a value indicating whether address names that match Jellyfin.Networking.Configuration.NetworkConfiguration.VirtualInterfaceNames should be Ignore for the purposes of binding."
+          },
+          "VirtualInterfaceNames": {
+            "type": "string",
+            "description": "Gets or sets a value indicating the interfaces that should be ignored. The list can be comma separated. <seealso cref=\"P:Jellyfin.Networking.Configuration.NetworkConfiguration.IgnoreVirtualInterfaces\" />."
+          },
+          "GatewayMonitorPeriod": {
+            "type": "integer",
+            "description": "Gets or sets the time (in seconds) between the pings of SSDP gateway monitor.",
+            "format": "int32"
+          },
+          "EnableMultiSocketBinding": {
+            "type": "boolean",
+            "description": "Gets a value indicating whether multi-socket binding is available.",
+            "readOnly": true
+          },
+          "TrustAllIP6Interfaces": {
+            "type": "boolean",
+            "description": "Gets or sets a value indicating whether all IPv6 interfaces should be treated as on the internal network.\r\nDepending on the address range implemented ULA ranges might not be used."
+          },
+          "HDHomerunPortRange": {
+            "type": "string",
+            "description": "Gets or sets the ports that HDHomerun uses."
+          },
+          "PublishedServerUriBySubnet": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "description": "Gets or sets the PublishedServerUriBySubnet\r\nGets or sets PublishedServerUri to advertise for specific subnets."
+          },
+          "AutoDiscoveryTracing": {
+            "type": "boolean",
+            "description": "Gets or sets a value indicating whether Autodiscovery tracing is enabled."
+          },
+          "AutoDiscovery": {
+            "type": "boolean",
+            "description": "Gets or sets a value indicating whether Autodiscovery is enabled."
+          },
+          "RemoteIPFilter": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "description": "Gets or sets the filter for remote IP connectivity. Used in conjuntion with <seealso cref=\"P:Jellyfin.Networking.Configuration.NetworkConfiguration.IsRemoteIPFilterBlacklist\" />."
+          },
+          "IsRemoteIPFilterBlacklist": {
+            "type": "boolean",
+            "description": "Gets or sets a value indicating whether <seealso cref=\"P:Jellyfin.Networking.Configuration.NetworkConfiguration.RemoteIPFilter\" /> contains a blacklist or a whitelist. Default is a whitelist."
+          },
+          "EnableUPnP": {
+            "type": "boolean",
+            "description": "Gets or sets a value indicating whether to enable automatic port forwarding."
+          },
+          "EnableRemoteAccess": {
+            "type": "boolean",
+            "description": "Gets or sets a value indicating whether access outside of the LAN is permitted."
+          },
+          "LocalNetworkSubnets": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "description": "Gets or sets the subnets that are deemed to make up the LAN."
+          },
+          "LocalNetworkAddresses": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "description": "Gets or sets the interface addresses which Jellyfin will bind to. If empty, all interfaces will be used."
+          },
+          "KnownProxies": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "description": "Gets or sets the known proxies. If the proxy is a network, it's added to the KnownNetworks."
+          },
+          "EnablePublishedServerUriByRequest": {
+            "type": "boolean",
+            "description": "Gets or sets a value indicating whether the published server uri is based on information in HTTP requests."
+          }
+        },
+        "additionalProperties": false,
+        "description": "Defines the Jellyfin.Networking.Configuration.NetworkConfiguration."
+      },
       "NewGroupRequestDto": {
         "type": "object",
         "properties": {
@@ -49829,6 +50303,62 @@
         ],
         "type": "string"
       },
+      "NotificationOption": {
+        "type": "object",
+        "properties": {
+          "Type": {
+            "type": "string",
+            "nullable": true
+          },
+          "DisabledMonitorUsers": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "description": "Gets or sets user Ids to not monitor (it's opt out)."
+          },
+          "SendToUsers": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "description": "Gets or sets user Ids to send to (if SendToUserMode == Custom)."
+          },
+          "Enabled": {
+            "type": "boolean",
+            "description": "Gets or sets a value indicating whether this MediaBrowser.Model.Notifications.NotificationOption is enabled."
+          },
+          "DisabledServices": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "description": "Gets or sets the disabled services."
+          },
+          "SendToUserMode": {
+            "allOf": [
+              {
+                "$ref": "#/components/schemas/SendToUserType"
+              }
+            ],
+            "description": "Gets or sets the send to user mode."
+          }
+        },
+        "additionalProperties": false
+      },
+      "NotificationOptions": {
+        "type": "object",
+        "properties": {
+          "Options": {
+            "type": "array",
+            "items": {
+              "$ref": "#/components/schemas/NotificationOption"
+            },
+            "nullable": true
+          }
+        },
+        "additionalProperties": false
+      },
       "NotificationResultDto": {
         "type": "object",
         "properties": {
@@ -51853,6 +52383,14 @@
         "type": "string",
         "description": "Enum SendCommandType."
       },
+      "SendToUserType": {
+        "enum": [
+          "All",
+          "Admins",
+          "Custom"
+        ],
+        "type": "string"
+      },
       "SeriesInfo": {
         "type": "object",
         "properties": {
@@ -52899,6 +53437,45 @@
         "type": "string",
         "description": "Delivery method to use during playback of a specific subtitle format."
       },
+      "SubtitleOptions": {
+        "type": "object",
+        "properties": {
+          "SkipIfEmbeddedSubtitlesPresent": {
+            "type": "boolean"
+          },
+          "SkipIfAudioTrackMatches": {
+            "type": "boolean"
+          },
+          "DownloadLanguages": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "nullable": true
+          },
+          "DownloadMovieSubtitles": {
+            "type": "boolean"
+          },
+          "DownloadEpisodeSubtitles": {
+            "type": "boolean"
+          },
+          "OpenSubtitlesUsername": {
+            "type": "string",
+            "nullable": true
+          },
+          "OpenSubtitlesPasswordHash": {
+            "type": "string",
+            "nullable": true
+          },
+          "IsOpenSubtitleVipAccount": {
+            "type": "boolean"
+          },
+          "RequirePerfectMatch": {
+            "type": "boolean"
+          }
+        },
+        "additionalProperties": false
+      },
       "SubtitlePlaybackMode": {
         "enum": [
           "Default",
@@ -54633,6 +55210,28 @@
         "additionalProperties": false,
         "description": "Provides the MAC address and port for wake-on-LAN functionality."
       },
+      "XbmcMetadataOptions": {
+        "type": "object",
+        "properties": {
+          "UserId": {
+            "type": "string",
+            "nullable": true
+          },
+          "ReleaseDateFormat": {
+            "type": "string"
+          },
+          "SaveImagePathsInNfo": {
+            "type": "boolean"
+          },
+          "EnablePathSubstitution": {
+            "type": "boolean"
+          },
+          "EnableExtraThumbsDuplication": {
+            "type": "boolean"
+          }
+        },
+        "additionalProperties": false
+      },
       "XmlAttribute": {
         "type": "object",
         "properties": {
```
</details>

DMouse10462:
@nielsvanvelzen: That additional "requestBody" object with the empty schema definitions allows for OpenAPI generator to create an API client that is aware of the need to send a body, like so:

<details>
<summary>Before (openapi-base)</summary>

configuration-api.d.ts
```typescript
/**
 * Request parameters for updateNamedConfiguration operation in ConfigurationApi.
 * @export
 * @interface ConfigurationApiUpdateNamedConfigurationRequest
 */
export interface ConfigurationApiUpdateNamedConfigurationRequest {
    /**
     * Configuration key.
     * @type {string}
     * @memberof ConfigurationApiUpdateNamedConfiguration
     */
    readonly key: string;
}
```
</details>

<details>
<summary>After (openapi-head)</summary>

configuration-api.d.ts
```typescript
/**
 * Request parameters for updateNamedConfiguration operation in ConfigurationApi.
 * @export
 * @interface ConfigurationApiUpdateNamedConfigurationRequest
 */
export interface ConfigurationApiUpdateNamedConfigurationRequest {
    /**
     * Configuration key.
     * @type {string}
     * @memberof ConfigurationApiUpdateNamedConfiguration
     */
    readonly key: string;
    /**
     * Configuration.
     * @type {any}
     * @memberof ConfigurationApiUpdateNamedConfiguration
     */
    readonly body: any;
}
```
</details>

Note also the differences in the resultant JS, such as the inclusion of the body parameter in the signature of ConfigurationApiAxiosParamCreator:

Before:
```typescript
updateNamedConfiguration: (key, options = {}) => __awaiter(this, void 0, void 0, function* () {
```

After:
```typescript
updateNamedConfiguration: (key, body, options = {}) => __awaiter(this, void 0, void 0, function* () {
```

Without these changes (or another set of changes that would accomplish adding a "requestBody" object to the endpoint's definition), we can't actually submit a configuration via the generated client. This got in the way as I was playing with implementing the DLNA configuration in jellyfin-vue.

crobibero (approved):
I'll say this is fine as a stop-gap, especially since it doesn't change anything except for the api spec

nielsvanvelzen (approved):
I'm not the biggest fan but it will do for now. Added support to the Kotlin SDK in jellyfin/jellyfin-sdk-kotlin#367.

PR #7567, opened 2022-04-07
Fix xmltv caching

cvium:
**Changes**
Compute the MD5 checksum of the gz files and use the hash as the temp folder name to avoid cluttering the temp folder with duplicate files.

**Issues**
Fixes #7565

Issue #7565, opened 2022-04-07
[Issue]: Live TV Guide data refresh creating junk file to fill up drive


ashipaek0:
### Please describe your bug

1. Add multiple EPG Guide xml to Live TV EPG section, Refresh Guide data
2. Go to /var/cache/jellyfin/temp/ on server
3. large random 'data.xml' fills up drive

### Jellyfin Version

10.8.0

### Environment

```markdown
- OS: Ubuntu Server
- Virtualization: KVM
- Clients: ANY
- Browser: ANY
- FFmpeg Version: Jellyfin-ffmpeg 4.4
- Playback Method:
- Hardware Acceleration:
- Plugins:
- Reverse Proxy:Cloudflare
- Base URL: None
- Networking: Host
- Storage:Local, Cloud
```

### Jellyfin logs

```shell
https://gist.github.com/ashipaek0/4ba860b4d5a36c859f52d65b3280790f
```

### Please attach any screenshots here

![Screenshot from 2022-04-07 12-03-14](https://user-images.githubusercontent.com/25457085/162185049-d7d17ee2-c525-4cc9-b933-b78b21c1147d.png)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ashipaek0:
Further Info: 
Apparently this happens when you add a xxxx.xml.gz link as EPG source

PR #7592, opened 2022-04-13
Fix xmltv caching part 2

1337joe:
Live TV Guide data in gz format is downloaded to `cache/xmltv` then uncompressed to `cache/temp`.

Issue #7565/PR #7567 fixed `cache/temp` filling up with old files, this PR addresses `cache/xmltv` having similar problems.

The old cache scheme for files in `cache/xmltv` is to create the name `[day of year]-[hour of day]-[id].xml` and see if that file exists before downloading a new one. There is nothing in place for deleting old data. In addition to wasting space storing old files, after a year it's possible for the naming system to repeat and read the cached files from the previous year.

**Changes:**
- Check cached file age directly and replace if too old.

**Issues**

### DIFF
--- a/Emby.Server.Implementations/LiveTv/Listings/XmlTvListingsProvider.cs
+++ b/Emby.Server.Implementations/LiveTv/Listings/XmlTvListingsProvider.cs
@@ -27,6 +27,8 @@ namespace Emby.Server.Implementations.LiveTv.Listings
 {
     public class XmlTvListingsProvider : IListingsProvider
     {
+        private static readonly TimeSpan _maxCacheAge = TimeSpan.FromHours(1);
+
         private readonly IServerConfigurationManager _config;
         private readonly IHttpClientFactory _httpClientFactory;
         private readonly ILogger<XmlTvListingsProvider> _logger;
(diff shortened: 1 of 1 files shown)


Emby.Server.Implementations/LiveTv/Listings/XmlTvListingsProvider.cs
```diff
@@ -70,13 +72,14 @@ private async Task<string> GetXml(ListingsProviderInfo info, CancellationToken c
                 return UnzipIfNeeded(info.Path, info.Path);
             }
 
-            string cacheFilename = DateTime.UtcNow.DayOfYear.ToString(CultureInfo.InvariantCulture) + "-" + DateTime.UtcNow.Hour.ToString(CultureInfo.InvariantCulture) + "-" + info.Id + ".xml";
+            string cacheFilename = info.Id + ".xml";
             string cacheFile = Path.Combine(_config.ApplicationPaths.CachePath, "xmltv", cacheFilename);
-            if (File.Exists(cacheFile))
+            if (File.Exists(cacheFile) && File.GetLastWriteTime(cacheFile) >= DateTime.Now.Subtract(_maxCacheAge))
```
crobibero, line 77: This should use UTC (both `GetLastWriteTime` and `DateTime.Now`

crobibero (approved):

cvium (approved):

Issue #7560, opened 2022-04-05
[10.8.0][Issue]: Very relevant bug in the search system appeared in version 10.8.0


BrunoPaiva513:
### Please describe your bug

Hi!

The update to the Beta 8.0 version of Jellyfin generated a major bug in the search engine.

**Accented words (very common in other languages) are no longer detected in the search. For example:**
When searching in Portuguese for Cidadão, the movie Cidadão Kane (the famous Citizen Kane in the original title) is not found due to the existence of “ã” in the word.

Previous versions of Jellyfin normally found titles like this. But I think it's also worth mentioning that my only annoyance with this wonderful platform was the fact that searches differentiate accented letters. In languages where the use of accents is common, search engines always understand searches like “cidadÃo” and “cidAo” in the same way. I believe it would be relevant to also solve this old problem, which is related.

Thank you very much
Bruno

### Jellyfin Version

10.8.0

### Environment

```markdown
- OS:
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

1337joe:
Other accent-related recent tickets:
#7450
#7484

Issue #7484, opened 2022-03-21
[Issue]: Search does not work properly for åäö


LinAGKar:
### Please describe your bug

Searching for a word containing å, ä or ö does not work properly, as Jellyfin replaces it with the similar-looking ASCII character. So for example, if you type "Jön", "Indiana Jones" will be included in the results, but not "Jönssonligan".

### Jellyfin Version

Other

### if other:

10.8-alpha5

### Environment

```markdown
- OS: OpenSUSE Leap 15.3
- Virtualization: Docker jellyfin/jellyfin
- Clients: Browser
- Browser: Firefox 98
- FFmpeg Version: 4.4.1-Jellyfin
- Hardware Acceleration: none
- Plugins: AudioDB 10.8.0.0, MusicBrainz 10.8.0.0, OMDb 10.8.0.0, Studio Images 10.8.0.0, TMDb 10.8.0.0
- Reverse Proxy: nginx
- Base URL: none
- Networking: Host
- Storage: local
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cbenco:
Seems like the Diacritics package doesn't strip this diaeresis correctly out of the film name retrieved from TMDB

![image](https://user-images.githubusercontent.com/13434960/160278849-ae6d134e-aa26-4543-8875-7d81c8d9d35c.png)

Jellifi007:
This bug is related to regression introduced by Bond-009@981cf4cfa0d45a3654578a1989738fc82469c889

Reverting to our own RemoveDiacritics extension should do the work (not sure why this has been changed).

I think we should put our own "reverted RemoveDiacritics extension" in a separated project ("Jellyfin.Diacritics" ?) to allow for less coupling with existing projects, rather than re-adding the extension into "MediaBrowser.Controller.Extensions;" like before.

Proposed target class optimized with shared thread safe regex object :
![Diacritics_corrected](https://user-images.githubusercontent.com/93096110/163390514-7b8d1a0f-f8c5-4bcb-9bd0-c84c8bf5dc54.png)

Proposed source :
[Extensions.cs.zip](https://github.com/jellyfin/jellyfin/files/8489164/Extensions.cs.zip)

Diacritics package has to be removed and import references corrected in following projects in order to restore correct Jellyfin behavior:
![Diacritics_projetcs](https://user-images.githubusercontent.com/93096110/163391403-98f548ea-bac1-4d92-a405-1446555a595a.png)

[...]

Issue #6393, opened 2021-08-04
NormalizationForm


sushilicious:
**Describe the bug**
`NormalizationForm.FormKD` is messing up korean titles with utf8.

**System (please complete the following information):**
 - OS: Archlinux and also on Debian
 - Virtualization: Docker and bare-metal
 - Clients: Browser
 - Browser: Firefox 91.0b
 - Jellyfin Version: master (commit 2521a13d7d8df3bfa5291386ba4389469f02ade8)
 - Playback: N/A
 - Hardware Acceleration: none
 - Installed Plugins: Anime, TVDB
 - Reverse Proxy: none
 - Base URL: none
 - Networking: Host
 - Storage: Local mounted and docker-mounted

**To Reproduce**
1. Add Korean files encoded in utf8.
2. Scan using one of the plugins using `FormKD`, (e.g., AniDB, TVDB)
3. Watch as logs send a very strangely encoded query.

**Expected behavior**
I expect to see a query encoded in utf8.

**Server Logs**
[2021-08-03 13:43:49.311 -07:00] [ERR] [29] Jellyfin.Plugin.Tvdb.Providers.TvdbSeriesProvider: No series results found for "운명처럼 널 사랑해"
(Supposed to be "운명처럼 널 사랑해")

[2021-08-03 13:44:18.087 -07:00] [ERR] [42] Jellyfin.Plugin.Tvdb.Providers.TvdbSeriesProvider: No series results found for "애타는 로맨스"
(Supposed to be "애타는 로맨스")

**FFmpeg Logs**
n/a

**Browser Console Logs**
n/a

**Screenshots**
n/a

**Additional context**
I've already added this as an issue to the tvdb plugin (jellyfin/jellyfin-plugin-tvdb#44), but this seems to many other plugins, so I'm creating another issue here.

Is there a reason why `NormalizationForm.FormKD` is used as opposed to something like `NormalizationForm.FormC`?

I'm asking because said normalization form is screwing up korean video titles encoded in utf8 (i.e., it works if I change it `FormC`). Are there any languages that depend on it? If so, I'll add a special case for korean, if not I'll add a pull request just changing it to `FormC`.

Perhaps jellyfin should provide a normalization API (e.g., handle encoding, remove "the"), so that issues like this can be changed managed in one location?

crobibero:
[...]

PR #7604, opened 2022-04-15
Fixes diacritics regression #7484 #7450 #7560

### LINKED ISSUE #7450: Accented characters in movies and people's name should be treated as non-accented
Take a Director, some metadata provider name him `Kieślowski` some others will name him `Kieslowski`

As of now (I'm using 10.8.0.-a5), these are two different directors.
He will appear as two different person, and related movies will be "director unrelated".

I guess that should be a very generic behaviours for all string metadata.

Moreover, I realize "original name" metadata for people (like a person's name in its own language script) are unsearchable.

Jellifi007:
**Changes**

Fixes diacritics regression introduced by Bond-009@981cf4cfa0d45a3654578a1989738fc82469c889

Example: searching "jon" will now correctly return movies containing "Jönsson":
image

**Issues**

Fixes #7450 #7484 #7560


Jellifi007:
Also note that:

- .NET 6.0 seems to replace automatically each invalid UTF-16 code point (ill-formed surrogates for example) by code point \uFFFD (replacement char).
- Xunit.InlineDataAttribute seems to replace each invalid UTF-16 code point by two \uFFFD chars (not sure why)

So the "_nonConformingUnicode" regex has to be slightly modified to take this in account and strip those unecessary \uFFFD chars.

crobibero:
I don't know enough about unicode, but https://github.com/jellyfin/jellyfin/issues/6393 requests `NormalizationForm.FormC`

Jellifi007:
FormC algorithm is already applied to the string stripped from diacritics before returning it.

So in the end, new tests with Korean string from https://github.com/jellyfin/jellyfin/issues/6393 does not show any trace of aforementionned issue :
image

if that's ok, I will add those new tests with Korean strings

Jellifi007:
About current AzurePipeline KO test status on all platforms : 
image
image

It looks like that #7624 merge broke some tests unrelated to this PR (see https://github.com/jellyfin/jellyfin/pull/7624/checks) :
image


(not sure if this is the right place to signal it)

src/Jellyfin.Extensions/StringExtensions.cs
```diff
@@ -7,6 +11,45 @@ namespace Jellyfin.Extensions
...
+            {
+                UnicodeCategory uc = CharUnicodeInfo.GetUnicodeCategory(c);
+                if (uc != UnicodeCategory.NonSpacingMark)
+                {
+                    withoutDiactritics.Append(c);
+                }
+            }
+
+            string normalizedText = withoutDiactritics.ToString().Normalize(NormalizationForm.FormC);
+            return normalizedText;
```
crobibero, line 40:
```suggestion
            return withoutDiactritics.ToString().Normalize(NormalizationForm.FormC);
```

crobibero (approved):

1337joe (approved):

Bond-009 (approved):

PR #7614, opened 2022-04-18
Rework Fedora build

### LINKED ISSUE #7471: [Issue]: No longer installs on Fedora 36, bump version?
### Please describe your bug

Installing the latest unstable build on Fedora 36 Prerelease, it returns:

```Error: 
 Problem: conflicting requests
  - nothing provides liblttng-ust.so.0()(64bit) needed by jellyfin-server-20220317.2-1.fc33.x86_64
```
As the Fedora release targets 33, I imagine targeting 35 or 36 is in order.

### Jellyfin Version

Other

### if other:

_No response_

### Environment

_No response_

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

### LINKED ISSUE #7504: [Issue]: Fedora 33 is EOL lets create packages for Fedora 34 and 35
### Please describe your bug

https://docs.fedoraproject.org/en-US/releases/eol/

Fedora 33 is EOL so lets create package to target Fedora 35

We should do like this : 

https://github.com/jellyfin/jellyfin-web/pull/2524/files

### Jellyfin Version

Other

### if other:

10.8

### Environment

```markdown
- OS: Fedora 
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

mihawk90:
# Changes
* Rewrite Fedora `Makefile` so we don't need to constantly update with every new Fedora release. This is especially useful when Fedora and Jellyfin release cycles don't line up. Version selection is as follows:
  1. TARGET environment variable, which is currently used already
  2. Currently running Fedora version
  3. Hardcoded Fallback version that can be updated occasionally
* Use ~~`latest`~~ `36` Fedora image in Fedora Docker builds
  * ~~Same reasoning as the Makefile, we don't need to constantly update ourselves when a base image tag can do that for us. The Fedora base image is maintained by the Fedora project (and from my understanding mostly automated), so we don't need to worry about this not being updated anytime soon.~~
  * ~~`latest` always points to the current *released* Fedora version. So while F36 is currently in beta, `latest` still points to F35.~~
  * changed to `36` as requested in review
* Remove additional dotnet-preview repo from Makefile
  * Repo doesn't seem maintained, and maintainers actively discourage the usage in production in its description
  * Considering this, we should build with stable .NET releases, which Fedora and RHEL 8+ repos already provide
* Use Fedora-version-specific runtime-identifier for `dotnet publish`
  * This has no actual effect right now judging by the [RID
        catalog](https://github.com/dotnet/docs/blob/3efd59151a7421172658a6fbcf88a0bd5fa7a92d/docs/core/rid-catalog.md), so this is just future proofing.
* Remove AutoReqProv
  * There's rarely a reason to use this feature, this is not one of them
  * In the [proposal of this feature](https://fedoraproject.org/wiki/AutoReqProv_(draft)#Usage) it is stated that this is not to be used on packages with binaries in /usr/bin and others, which is the case in this package.
  * also didn't seem to work since we were still having dependency issues with implicit dependencies anyway
* Removed `DOTNET_SKIP_FIRST_TIME_EXPERIENCE` as it is unused in .NET SDK
  * see dotnet/sdk#9945
  * it's already merged for removal in future versions
* Filter dependency to fix F36 build
  * Filter for F36+ since we don't know when (if ever) this issue will befixed (it's up to .NET SDK). If it's ever fixed we can adjust or remove this condition.
  * built on F34, F35, F36, and EPEL7 targets to confirm it only affects the F36 RPM. Can't test EPEL 8 (see below).
* general housecleaning in the spec file
  * Replaced shell script used to start JF with a symlink, since it didn't do anything else.
  * Sorted installs by category/path to make it easier to maintain the spec.
  * Defined `%defattr` in file section so we don't need to set it on a by-file level. This also helps with the messed up default file permissions `dotnet publish` produces.
  * Removed some duplicate attribute definitions in `%install` and `%files` sections. They are now all in the `%files` section because of `%defattr` being specified and overriding them otherwise.

## Minor adjustments:
* Move building process `dotnet publish` to `%build` section
  * Also removed `--output` from this due to an outstanding bug on SDK's side. This also separates building and installing as intended
* define LICENSE as `%license`, which automatically puts it in a standardised directory

## Tests done
:heavy_check_mark: `make rpms`
:heavy_check_mark: `TARGET=fedora-123 make rpms`
:heavy_check_mark: `TARGET=gibberish make rpms`
:heavy_check_mark: `TARGET=epel-7-x86_64 make rpms`
:x: `TARGET=epel-8-x86_64 make rpms` - `mock-core-configs`  doesn't come with a `.cfg` for this, but this is an issue with the current version already. Not sure what the correct replacement is, `rhel+epel-8`?

# Related Issues
* fixes #7471
* fixes #7504 
* fixes jellyfin/jellyfin-web#2059
* companion PR to jellyfin/jellyfin-web#3571

---

# Just for reference, these are pretty much solved:

**Open questions/potential changes**
@brianjmurrell sorry to ping you like this but it seems you did a lot of work on the Fedora builds already, so possibly you're best addressed with these
1. Should we possibly use rawhide as a fallback instead? Otherwise we will still have to update the fallback periodically. Rawhide might be the better option here.
2. Do we really need a fallback at all? The Azure pipelines are passing the environment variable from what I can tell, if building locally it will likely be on Fedora anyway. And if it's not Fedora it's probably going to be RHEL (or compatibles) where any of the released Fedoras or Rawhide are not going to be useful either.
3. Currently we're not dealing with RHEL-clones (Alma, CentOS, Rocky) at all, should we?
5. When passing unhandled targets (Centos, Alma, Rocky, SuSE(?)), there's no repo passed to `--addrepo` at all. Should we just error out and cancel the build?

Only tangentially related, but are the added repos even needed anymore? At least on F36 the required Dotnet packages are already in the Fedora repos:
```
% rpm -E %fedora
36

% dnf search dotnet-sdk              
Last metadata expiration check: 14:04:24 ago on Sat 16 Apr 2022 23:14:07 CEST.
===================== Name Matched: dotnet-sdk ======================
dotnet-sdk-3.1.x86_64 : .NET Core 3.1 Software Development Kit
dotnet-sdk-3.1-source-built-artifacts.x86_64 : Internal package for
     ...: building .NET Core 3.1 Software Development Kit
dotnet-sdk-6.0.x86_64 : .NET 6.0 Software Development Kit
dotnet-sdk-6.0-source-built-artifacts.x86_64 : Internal package for
     ...: building .NET 6.0 Software Development Kit

% dnf search dotnet-runtime
Last metadata expiration check: 14:04:47 ago on Sat 16 Apr 2022 23:14:07 CEST.
=================== Name Matched: dotnet-runtime ====================
dotnet-runtime-3.1.x86_64 : NET Core 3.1 runtime
dotnet-runtime-6.0.x86_64 : NET 6.0 runtime
```

Also one important note and also a reason for this being a draft: this does *not* solve the issue described in #7471.
While the RPM builds just fine, it cannot be installed on F36 due to said missing dependency:
```
% sudo dnf install ./jellyfin-10.8.0\~beta2-1.fc36.x86_64.rpm ./jellyfin-server-10.8.0\~beta2-1.fc36.x86_64.rpm ./jellyfin-web-10.8.0\~beta2-1.fc36.noarch.rpm
Last metadata expiration check: 2:04:57 ago on Mon 18 Apr 2022 02:28:42 CEST.
Error: 
 Problem 1: conflicting requests
  - nothing provides liblttng-ust.so.0()(64bit) needed by jellyfin-server-10.8.0~beta2-1.fc36.x86_64
 Problem 2: package jellyfin-10.8.0~beta2-1.fc36.x86_64 requires jellyfin-server = 10.8.0~beta2-1.fc36, but none of the providers can be installed
  - conflicting requests
  - nothing provides liblttng-ust.so.0()(64bit) needed by jellyfin-server-10.8.0~beta2-1.fc36.x86_64
(try to add '--skip-broken' to skip uninstallable packages)
```

Still need to find a solution to that.

### DIFF
--- a/deployment/Dockerfile.fedora.amd64
+++ b/deployment/Dockerfile.fedora.amd64
@@ -1,4 +1,4 @@
-FROM fedora:33
+FROM fedora:36
 # Docker build arguments
 ARG SOURCE_DIR=/jellyfin
 ARG ARTIFACT_DIR=/dist
--- a/deployment/Dockerfile.fedora.amd64
+++ b/deployment/Dockerfile.fedora.amd64
@@ -9,7 +9,7 @@ ENV IS_DOCKER=YES
 
 # Prepare Fedora environment
 RUN dnf update -yq \
-  && dnf install -yq @buildsys-build rpmdevtools git dnf-plugins-core libcurl-devel fontconfig-devel freetype-devel openssl-devel glibc-devel libicu-devel systemd wget
+  && dnf install -yq @buildsys-build rpmdevtools git dnf-plugins-core libcurl-devel fontconfig-devel freetype-devel openssl-devel glibc-devel libicu-devel systemd wget make
 
 # Install DotNET SDK
 RUN wget -q https://download.visualstudio.microsoft.com/download/pr/dc930bff-ef3d-4f6f-8799-6eb60390f5b4/1efee2a8ea0180c94aff8f15eb3af981/dotnet-sdk-6.0.300-linux-x64.tar.gz -O dotnet-sdk.tar.gz \
--- a/fedora/jellyfin.spec
+++ b/fedora/jellyfin.spec
@@ -10,7 +10,7 @@ Name:           jellyfin
 Version:        10.8.0
 Release:        1%{?dist}
 Summary:        The Free Software Media System
-License:        GPLv3
+License:        GPLv2
 URL:            https://jellyfin.org
 # Jellyfin Server tarball created by `make -f .copr/Makefile srpm`, real URL ends with `v%%{version}.tar.gz`
 Source0:        jellyfin-server-%{version}.tar.gz
--- a/fedora/jellyfin.spec
+++ b/fedora/jellyfin.spec
@@ -62,6 +62,7 @@ the Jellyfin server to bind to ports 80 and/or 443 for example.
 %prep
 %autosetup -n jellyfin-server-%{version} -b 0
 
+
 %build
 export DOTNET_CLI_TELEMETRY_OPTOUT=1
 export PATH=$PATH:/usr/local/bin
@@ -71,47 +72,62 @@ dotnet publish --configuration Release --self-contained --runtime %{dotnet_runti
 
 
 %install
-%{__mkdir} -p %{buildroot}%{_libdir}/%{name} %{buildroot}%{_bindir}
-%{__cp} -r Jellyfin.Server/bin/Release/net6.0/%{dotnet_runtime}/publish/* %{buildroot}%{_libdir}/%{name}
+# Jellyfin files
+%{__mkdir} -p %{buildroot}%{_libdir}/jellyfin %{buildroot}%{_bindir}
+%{__cp} -r Jellyfin.Server/bin/Release/net6.0/%{dotnet_runtime}/publish/* %{buildroot}%{_libdir}/jellyfin
+ln -srf %{_libdir}/jellyfin/jellyfin %{buildroot}%{_bindir}/jellyfin
+%{__install} -D %{SOURCE14} %{buildroot}%{_libexecdir}/jellyfin/restart.sh
 
-%{__install} -D -m 0644 %{SOURCE15} %{buildroot}%{_sysconfdir}/systemd/system/jellyfin.service.d/override.conf
-%{__install} -D -m 0644 %{SOURCE17} %{buildroot}%{_unitdir}/jellyfin.service.d/jellyfin-server-lowports.conf
-%{__install} -D -m 0644 Jellyfin.Server/Resources/Configuration/logging.json %{buildroot}%{_sysconfdir}/jellyfin/logging.json
-%{__mkdir} -p %{buildroot}%{_bindir}
-tee %{buildroot}%{_bindir}/jellyfin << EOF
-#!/bin/sh
-exec %{_libdir}/jellyfin/jellyfin \${@}
-EOF
+# Jellyfin config
+%{__install} -D Jellyfin.Server/Resources/Configuration/logging.json %{buildroot}%{_sysconfdir}/jellyfin/logging.json
+%{__install} -D %{SOURCE12} %{buildroot}%{_sysconfdir}/sysconfig/jellyfin
+
+# system config
+%{__install} -D %{SOURCE16} %{buildroot}%{_prefix}/lib/firewalld/services/jellyfin.xml
+%{__install} -D %{SOURCE13} %{buildroot}%{_sysconfdir}/sudoers.d/jellyfin-sudoers
+%{__install} -D %{SOURCE15} %{buildroot}%{_sysconfdir}/systemd/system/jellyfin.service.d/override.conf
+%{__install} -D %{SOURCE11} %{buildroot}%{_unitdir}/jellyfin.service
+
+# empty directories
 %{__mkdir} -p %{buildroot}%{_sharedstatedir}/jellyfin
 %{__mkdir} -p %{buildroot}%{_sysconfdir}/jellyfin
-%{__mkdir} -p %{buildroot}%{_var}/log/jellyfin
 %{__mkdir} -p %{buildroot}%{_var}/cache/jellyfin
+%{__mkdir} -p %{buildroot}%{_var}/log/jellyfin
 
-%{__install} -D -m 0644 %{SOURCE11} %{buildroot}%{_unitdir}/jellyfin.service
-%{__install} -D -m 0644 %{SOURCE12} %{buildroot}%{_sysconfdir}/sysconfig/jellyfin
-%{__install} -D -m 0600 %{SOURCE13} %{buildroot}%{_sysconfdir}/sudoers.d/jellyfin-sudoers
-%{__install} -D -m 0755 %{SOURCE14} %{buildroot}%{_libexecdir}/jellyfin/restart.sh
-%{__install} -D -m 0644 %{SOURCE16} %{buildroot}%{_prefix}/lib/firewalld/services/jellyfin.xml
+# jellyfin-server-lowports subpackage
+%{__install} -D -m 0644 %{SOURCE17} %{buildroot}%{_unitdir}/jellyfin.service.d/jellyfin-server-lowports.conf
+
 
 %files
 # empty as this is just a meta-package
 
 %files server
-%attr(755,root,root) %{_bindir}/jellyfin
-%{_libdir}/jellyfin/*
+%defattr(644,root,root,755)
+
+# Jellyfin files
+%{_bindir}/jellyfin
 # Needs 755 else only root can run it since binary build by dotnet is 722
+%attr(755,root,root) %{_libdir}/jellyfin/createdump
 %attr(755,root,root) %{_libdir}/jellyfin/jellyfin
-%{_unitdir}/jellyfin.service
-%{_libexecdir}/jellyfin/restart.sh
-%{_prefix}/lib/firewalld/services/jellyfin.xml
-%attr(755,jellyfin,jellyfin) %dir %{_sysconfdir}/jellyfin
+%{_libdir}/jellyfin/*
+%attr(755,root,root) %{_libexecdir}/jellyfin/restart.sh
+
+# Jellyfin config
+%config(noreplace) %attr(644,jellyfin,jellyfin) %{_sysconfdir}/jellyfin/logging.json
 %config %{_sysconfdir}/sysconfig/jellyfin
+
+# system config
+%{_prefix}/lib/firewalld/services/jellyfin.xml
+%{_unitdir}/jellyfin.service
 %config(noreplace) %attr(600,root,root) %{_sysconfdir}/sudoers.d/jellyfin-sudoers
 %config(noreplace) %{_sysconfdir}/systemd/system/jellyfin.service.d/override.conf
-%config(noreplace) %attr(644,jellyfin,jellyfin) %{_sysconfdir}/jellyfin/logging.json
+
+# empty directories
 %attr(750,jellyfin,jellyfin) %dir %{_sharedstatedir}/jellyfin
-%attr(-,jellyfin,jellyfin) %dir %{_var}/log/jellyfin
+%attr(755,jellyfin,jellyfin) %dir %{_sysconfdir}/jellyfin
 %attr(750,jellyfin,jellyfin) %dir %{_var}/cache/jellyfin
+%attr(-,  jellyfin,jellyfin) %dir %{_var}/log/jellyfin
+
 %license LICENSE
 
 
--- a/fedora/jellyfin.spec
+++ b/fedora/jellyfin.spec
@@ -30,6 +30,12 @@ BuildRequires:  libcurl-devel, fontconfig-devel, freetype-devel, openssl-devel, 
 BuildRequires:  dotnet-runtime-6.0, dotnet-sdk-6.0
 Requires: %{name}-server = %{version}-%{release}, %{name}-web = %{version}-%{release}
 
+# Temporary (hopefully?) fix for https://github.com/jellyfin/jellyfin/issues/7471
+%if 0%{?fedora} >= 36
+%global __requires_exclude ^liblttng-ust\\.so\\.0.*$
+%endif
+
+
 %description
 Jellyfin is a free software media system that puts you in control of managing and streaming your media.
 
@@ -107,6 +113,7 @@ EOF
 %attr(-,jellyfin,jellyfin) %dir %{_var}/log/jellyfin
 %attr(750,jellyfin,jellyfin) %dir %{_var}/cache/jellyfin
 %license LICENSE
+
 
 %files server-lowports
 %{_unitdir}/jellyfin.service.d/jellyfin-server-lowports.conf
--- a/fedora/jellyfin.spec
+++ b/fedora/jellyfin.spec
@@ -1,7 +1,7 @@
 %global         debug_package %{nil}
 # Set the dotnet runtime
 %if 0%{?fedora}
-%global         dotnet_runtime  fedora-x64
+%global         dotnet_runtime  fedora.%{fedora}-x64
 %else
 %global         dotnet_runtime  centos-x64
 %endif
@@ -25,13 +25,10 @@ Source17:       jellyfin-server-lowports.conf
 %{?systemd_requires}
 BuildRequires:  systemd
 BuildRequires:  libcurl-devel, fontconfig-devel, freetype-devel, openssl-devel, glibc-devel, libicu-devel
-# Requirements not packaged in main repos
-# COPR @dotnet-sig/dotnet or
+# Requirements not packaged in RHEL 7 main repos, added via Makefile
 # https://packages.microsoft.com/rhel/7/prod/
 BuildRequires:  dotnet-runtime-6.0, dotnet-sdk-6.0
 Requires: %{name}-server = %{version}-%{release}, %{name}-web = %{version}-%{release}
-# Disable Automatic Dependency Processing
-AutoReqProv:    no
 
 %description
 Jellyfin is a free software media system that puts you in control of managing and streaming your media.
@@ -60,14 +57,17 @@ the Jellyfin server to bind to ports 80 and/or 443 for example.
 %autosetup -n jellyfin-server-%{version} -b 0
 
 %build
+export DOTNET_CLI_TELEMETRY_OPTOUT=1
+export PATH=$PATH:/usr/local/bin
+# cannot use --output due to https://github.com/dotnet/sdk/issues/22220
+dotnet publish --configuration Release --self-contained --runtime %{dotnet_runtime} \
+    "-p:DebugSymbols=false;DebugType=none" Jellyfin.Server
+
 
 %install
-export DOTNET_CLI_TELEMETRY_OPTOUT=1
-export DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1
-export PATH=$PATH:/usr/local/bin
-dotnet publish --configuration Release --output='%{buildroot}%{_libdir}/jellyfin' --self-contained --runtime %{dotnet_runtime} \
-    "-p:DebugSymbols=false;DebugType=none" Jellyfin.Server
-%{__install} -D -m 0644 LICENSE %{buildroot}%{_datadir}/licenses/jellyfin/LICENSE
+%{__mkdir} -p %{buildroot}%{_libdir}/%{name} %{buildroot}%{_bindir}
+%{__cp} -r Jellyfin.Server/bin/Release/net6.0/%{dotnet_runtime}/publish/* %{buildroot}%{_libdir}/%{name}
+
 %{__install} -D -m 0644 %{SOURCE15} %{buildroot}%{_sysconfdir}/systemd/system/jellyfin.service.d/override.conf
 %{__install} -D -m 0644 %{SOURCE17} %{buildroot}%{_unitdir}/jellyfin.service.d/jellyfin-server-lowports.conf
 %{__install} -D -m 0644 Jellyfin.Server/Resources/Configuration/logging.json %{buildroot}%{_sysconfdir}/jellyfin/logging.json
@@ -106,7 +106,7 @@ EOF
 %attr(750,jellyfin,jellyfin) %dir %{_sharedstatedir}/jellyfin
 %attr(-,jellyfin,jellyfin) %dir %{_var}/log/jellyfin
 %attr(750,jellyfin,jellyfin) %dir %{_var}/cache/jellyfin
-%{_datadir}/licenses/jellyfin/LICENSE
+%license LICENSE
 
 %files server-lowports
 %{_unitdir}/jellyfin.service.d/jellyfin-server-lowports.conf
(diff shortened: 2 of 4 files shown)


crobibero:
This PR should target `release-10.8.z` so it gets into the 10.8 release, and will be merged into `master` after

mihawk90:
Thank you, getting on that.
Not very experienced with this so figuring it out :sweat_smile:

mihawk90:
I have no idea if I did this right, but my git graph looked fine and the parent commit also seems to be the tip of the release branch so... I hope so?

mihawk90:
OK... so, I might open up the scope of this PR a little bit if that is OK.

Particularly, I would suggest 2 further changes to the Fedora build process:

1. Remove `AutoReqProv:    no` from at least the server-spec, if not the web-spec as well.
    - I don't see any possible benefit of *not* having RPM figure out the dependencies on its own
    - in the [proposal of this feature](https://fedoraproject.org/wiki/AutoReqProv_(draft)#Usage), it was stated that this feature is not to be used for packages with executables:
    > These filtering macros MUST only be used with packages which meet one of the following criteria:
    > 
    > Noarch packages
    > Architecture specific packages with no binaries in $PATH (e.g. /bin, /usr/bin, /sbin, /sbin) or libexecdir and no system libs in libdir. This includes all of the subpackages generated from the spec file.
They are not permitted in any other cases, because the macros interfere with the "coloring" of elf32/64 executables done internally by RPM to support multilib installs.
2. Remove the additional Dotnet repos for builds on Fedora and RHEL/EPEL
    - The dotnet-sig repo doesn't seem to be maintained anymore. The last version is 6.0.0 for the runtime and 6.0.100 for the SDK on the COPR, while the official Fedora repo is shipping 6.0.3 and 6.0.103 respectively already, with further updates on the way (as of this writing).
    - Even if it were maintained, this repo is shipping preview versions, and even the [repo description](https://copr.fedorainfracloud.org/coprs/g/dotnet-sig/dotnet-preview/) itself states that these are probably unfit:
    > You probably want to the .NET Core packages in the main Fedora repositories for Fedora 32 and later. Otherwise, use https://copr.fedorainfracloud.org/coprs/g/dotnet-sig/dotnet/ instead.
    > 
    > The packages in this repo may have bugs and break between upgrades and do other bad things.
    - Considering this, we should be building stable releases with stable .NET builds too. There is no reason for running a stable release on a possibly-wonky .NET base.
    - Whether we could use these (if they are ever to be maintained again) in nightly builds or the CI would be a separate discussion to have.

Any objections to these? Otherwise I would push them on the same branch and change the PR title accordingly.

crobibero:
As long as it's changing fedora (or rpm) specific files I see no issue with having it in one PR

brianjmurrell:
If you are going to make these kinds of changes, you need to make sure you test on all of the supported platforms.  You could do that locally with `mock` easily enough, but you might want to consider setting up a copr project that will test all of your changes on all platforms.

BEWARE: the CI here is not good enough to test these changes.  It builds packages in non-standard ways (i.e. runs rpmbuild as root) and does not build for all supported platforms on PRs (last time I checked) so you could make changes that break on untested platforms.

mihawk90:
> You could do that locally with `mock` easily enough

Already doing that don't worry, see the "Tests done" section ;)

On that note, .NET takes a literal eternity to "Determine projects to restore". Is there a way to disable that? Considering `mock` cleans the build root on every compile there is never going to be any "project to restore".
Not really worried about that though, that can be done down the road.

Do you have any recommendations regarding the 4 posted questions in the original post?

I guess 3 and 4 sort of take care of themselves if we're going to remove the additional repos anyway (except for RHEL/EPEL 7), but 1 and 2 regarding rawhide as fallback are still open.

brianjmurrell:
> Already doing that don't worry, see the "Tests done" section ;)

Ideally you could modify this project's CI to actually do the full matrix of RPM builds on PRs so that PRs don't cause unknown regressions in RPM building.  I had already pushed a PR to add building an RPM on at least one platform on PRs.  I don't recall if it ever got merged though.

brianjmurrell:
> I was thinking about that, and not just for CI but for the releases too.

Yes, of course.  If it's not done for releases, there'd be no RPMs.  Unless somebody is doing that out-of-band.

> But I didn't want to cram it into this PR because it might be a bit too much.

I don't see adding more complete testing, particularly of changes being made in a PR as adding too much.  Rather I would say just the opposite.  It's adding too little to leave it out.

> And I also still need to look into how everything actually works. Ideally I'd like to build using mock in docker,

+1.  `mock` should be the minimum requirement as it finds all kinds of assumptions that other environments (even just a plain Docker image) build in that might not be universally true.

> but mock documentation mentions that it needs a specific Docker flag set for that to work

You do need CAP_SYSADM or privileged, yes.  But that should not pose any problems.

> so I'm not sure if we can even do that.

Why not?  Any basic Docker CI system should not prohibit either of the above.

> And I currently don't even have Docker set up to test this sooo yeah.

It's not difficult.
 
> edit: it seems there's also a transition from Azure to GitHub Actions on the Horizon (see the [Projects page](https://github.com/orgs/jellyfin/projects/31), so we need to coordinate with that too.

Perhaps.  But if you drop into Azure before the above is done, it should just all transition with everything else.

mihawk90:
> I don't see adding more complete testing, particularly of changes being made in a PR as adding too much. Rather I would say just the opposite. It's adding too little to leave it out.

Sure, I'm not saying we shouldn't do it, I would just probably do it as a separate PR since I don't even know yet what exactly needs to be changed and where.

> You do need CAP_SYSADM or privileged, yes. But that should not pose any problems.
> 
> > so I'm not sure if we can even do that.
> 
> Why not? Any basic Docker CI system should not prohibit either of the above.

Yeah I read what it needs, I just don't even know what they are or how they're set 😅

I'm just saying I have absolutely 0 experience with this so it'll take some time to look into it.

mihawk90:
Regarding that last commit:

It's a fairly big chunk of change and I triple- and quadruple checked that everything is fine, but if someone could take another pair of eyes and checked if I missed any permissions, that would be nice.

I built a package for both the `v10.8.0-beta2` tag and my branch, then `rpm -qp --dump`ed both of them and ran a diff on the outputs. Everything seemed fine, the only things that changed were the timestamps (obviously) and the permissions on the `.dll` and `.so` files, which was intended since they don't need the executable bit.

fedora/jellyfin.spec
```diff
@@ -71,47 +72,62 @@ dotnet publish --configuration Release --self-contained --runtime %{dotnet_runti
...
-%{__install} -D -m 0644 %{SOURCE15} %{buildroot}%{_sysconfdir}/systemd/system/jellyfin.service.d/override.conf
-%{__install} -D -m 0644 %{SOURCE17} %{buildroot}%{_unitdir}/jellyfin.service.d/jellyfin-server-lowports.conf
-%{__install} -D -m 0644 Jellyfin.Server/Resources/Configuration/logging.json %{buildroot}%{_sysconfdir}/jellyfin/logging.json
-%{__mkdir} -p %{buildroot}%{_bindir}
-tee %{buildroot}%{_bindir}/jellyfin << EOF
-#!/bin/sh
-exec %{_libdir}/jellyfin/jellyfin \${@}
-EOF
+# Jellyfin files
+%{__mkdir} -p %{buildroot}%{_libdir}/jellyfin %{buildroot}%{_bindir}
```
brianjmurrell, line 76: Why are you reverting `%{name}` to `jellyfin` here (and elsewhere)?

mihawk90: I had used `%{name}` only for those two lines and it was inconsistent with the rest of the file.

I was thinking about replacing `jellyfin` everywhere but I decided against it since I didn't know which of them were hardcoded in the .NET code and which were variable.

Essentially this just brings it in line with the rest of the file. The [current version](https://github.com/jellyfin/jellyfin/blob/a2abae3014578e8d9c7f1b54140fe3dcf1503ad2/fedora/jellyfin.spec) uses `%{name}` only for the relation to other pacakges (server and server-lowports).

joshuaboniface:
So I don't have any strong opinions at all on any of this, Fedora specs have never been my strong suit and as mentioned, Brian has done most of the work. I'll defer entirely to you guys.

CI is static for now until post-10.8.0 when we change the entire build system over to GitHub Actions. At that point, we can customize CI however is needed but for today/10.8.0, it must stay the same for consistency.

mihawk90:
Now that 10.8 is released, should I rebase this on [v10.8.0](https://github.com/jellyfin/jellyfin/tree/v10.8.0) ?

deployment/Dockerfile.fedora.amd64
```diff
@@ -1,4 +1,4 @@
-FROM fedora:33
+FROM fedora:latest
```
h1dden-da3m0n, line 1: I may not be able to comment on the rest of the changes, as distro packaging is not my strong suite nor focus area, however PLEASE NEVER USE `latest` IN CI!

```suggestion
FROM fedora:36
```

h1dden-da3m0n:
> **Note**: I am in favour of getting this in to 10.8.x as a lot of research on Fedora packaging seems to have gone into this and F36 being a target platform (even if its community maintained, as we have no Fedora specialist on the team).
> On another note though, distro packaging will be spun off into separate repositories once we reworked the CI (current plan for 10.9.x), so don't be surprised if your changes will then get migrated over to such a RPM packaging repository once that's ready.

Thank you for the work you put into this mihawk and hope we can get your changes in for 10.8.x

deployment/Dockerfile.fedora.amd64
oddstr13: I second this, chasing down bugs cause the build scripts silently changed a version without anyone knowing won't be fun, it's better to have an explicit pr for the versions, so that someone could have a chance of knowing what's up.

mihawk90: Alright fair enough, I can edit that commit during the rebase, not a big deal.
Ideally the packages would be built in `mock` inside the docker container (as discussed above) and then the base image doesn't matter all that much anyway.

h1dden-da3m0n: Understood, yet still, having versions of dependencies change unexpectedly in any CI environment is just asking for trouble, even if the RPM build itself may be encapsulated in `mock`.

Regardless, could you please rebase the PR, so that is ready for a final review?
We should get it into the next 10.8.x then :smiley:

brianjmurrell:
> Something that has been ignored so far and that should be taken note of for the RPM builds (it was brought up due to future CI discussions):
> 
> Our FFMPEG does currently not build an RPM, and the regular OS FFMPEG may cause unexpected behaviour with a 10.8.x server (if you intend to use HWA). As mentioned before, we don't have any RPM gurus on the team (we are all in Debian or Arch land) so we were unable to find a good packaging solution for RPM of our FFMPEG.

It would probably just be better for everyone if whatever you have in your FFMPEG fork is contributed back upstream, yes?

h1dden-da3m0n:
@brianjmurrell obviously, and we are trying to do that where we can, but if you check the FFMPEG project you will see that not only is an elaborated long process to get your changes in, but some areas wont get accepted due to lack of active maintainer of certain areas of FFMPEG.

Same goes for JSO, yet that is on another level of a nightmare to upstream changes to, yet we try to upstream the changes we do where we can.

mihawk90:
> we were unable to find a good packaging solution for RPM of our FFMPEG

In theory we could adapt [RPM Fusion's ffmpeg spec file](https://pkgs.rpmfusion.org/cgit/free/ffmpeg.git/tree/ffmpeg.spec) and change source and target path. Just not sure if installing multiple instances of ffmpeg is a good idea :thinking: 
Something to think about later I guess.

brianjmurrell:
> In theory we could adapt [RPM Fusion's ffmpeg spec file](https://pkgs.rpmfusion.org/cgit/free/ffmpeg.git/tree/ffmpeg.spec) and change source and target path.

Why change the source and target path?

> Just not sure if installing multiple instances of ffmpeg is a good idea thinking Something to think about later I guess.

Don't install multiple instances.  If the Jellyfin ffmpeg is truly a super-set of RPM Fusion's, just have the Jellyfin one supersede the RPM Fusion one.  To ensure it doesn't get upgraded over, add a new `Provides: jellyfin-ffmpeg` to it and then have `jellyfin-server` `Requires: jellyfin-ffmpeg`.  The only downside here is that new `ffmpeg` packages from Jellyfin will need to be rolled when RPM Fusion updates theirs **AND** something has a `Requires:` for the new version as a minimum.  That should be pretty rare though.

Rebuilding an updated Jellyfin `ffmpeg` when RPM Fusion updates theirs should be fully automatable in GitHub Actions so that it happens pretty quickly after RPM Fusion does update.

mihawk90:
> Why change the source and target path?

Well the Source obviously because we'd need the source of JFs ffmpeg ;)
And Target because I was thinking that JF needed it in a specific location and not systemwide.
But maybe not then.

---

I'm just wondering are there any plans to host an actual repository for the RPMs? Having the users require to download and install 3 (with ffmpeg 4) packages on their own on a regular basis seems a bit of a pain.
There used to be a COPR but that stopped building and besides technically it couldn't be hosted on COPR anyway because of the ffmpeg dependency (I ran into the same issue with OBS studio). SUSE's Open Build Service wasn't working for me either because I couldn't find the dependencies on there and they don't allow to add external repos at all.

So the only option I found for me was Packagecloud, but the free version could hit the limit fairly quickly with the install base of JF and the Pro plan is a bit pricey.
Maybe selfhosting is an option but I have no experience with that whatsoever.

h1dden-da3m0n:
Well in regards of system wide or non system wide FFMPEG: there is, to my knowledge, a EnvVar that Jellyfin server respects and uses `JELLYFIN_FFMPEG` (we use it in the Container Images)

In regards to RPM repo: this is a tat complicated, yes we could host it, but for now RPM was a community maintained effort, with brian (correct me if I am wrong) maintaining parts of that.
To my knowledge, we currently have no plan to host our own RPM repo in the foreseeable future.

brianjmurrell:
> Well the Source obviously because we'd need the source of JFs ffmpeg ;)

Ah, yes of course.  I was reading source as something different.

> I'm just wondering are there any plans to host an actual repository for the RPMs?

Indeed!  There is [# dnf copr enable brianjmurrell/jellyfin](https://copr.fedorainfracloud.org/coprs/brianjmurrell/jellyfin/) at the moment, but something more officially linked to the JF project would be good indeed.

> Having the users require to download and install 3 (with ffmpeg 4) packages on their own on a regular basis seems a bit of a pain.

Completely agreed which is why I maintain the above COPR.

>There used to be a COPR but that stopped building

Which COPR?  Mine is still building.  It's building 10.8.0 right now.

FWIW, I use RPM Fusion `ffmpeg` with my Fedora and EL8 JF machines.  I don't do any HWA transcoding though.

mihawk90:
> Which COPR? Mine is still building. It's building 10.8.0 right now.

Hm interesting, I haven't checked it in a while tbh, should have looked at it before posting :D I just remember there being an issue about it on here that got closed at some point (I think?).

And yeah I'm using 10.8.0 on Fedora 36 with no issues with the default ffmpeg, but not using HWA either so I couldn't say anything about it.

Shadowghost:
To bring some more info onto the ffmpeg topic:
It's not only about just building ffmpeg, that would only require a small adoption of the rpmfusion rpm stuff.
Depending on the used architecture we're building several libraries, tools and in case of Intel also their driver from source (either directly from master or the latest available versions) and ship those together with our ffmpeg which is built against those libraries.
The ffmpeg binaries themself are installed to a custom location and not added to any global paths, so if a user really wants to use it, manual linking or a direct call is required. This was done to not break any existing systems.

I took a look at how you'd do this properly with the RPM buildsystem some months ago and haven't seen any concrete solution to this kind of usecase (but as already said, we're all new to this on the team).

deployment/Dockerfile.fedora.amd64
mihawk90: Rebased and pushed, waiting for the CI to complain :)

brianjmurrell:
> Hm interesting, I haven't checked it in a while tbh, should have looked at it before posting :D I just remember there being an issue about it on here that got closed at some point (I think?).

Note that there are two different projects in my COPR:  [jellyfin](https://copr.fedorainfracloud.org/coprs/brianjmurrell/jellyfin/) which only builds stable GA releases and [jellyfin-prerelease](https://copr.fedorainfracloud.org/coprs/brianjmurrell/jellyfin-prerelease/) which builds _alpha_, _beta_, and _rc_ releases.  This is done so that people can have a stable channel and not get updated to prereleases if they don't want those.  You would have had to have the _jellyfin-prerelease_ project enabled to have gotten the 10.8.0 alpha. beta and release-candidates.

h1dden-da3m0n (approved):
Thank you for all the discussion and everyone chiming in!

> **Note**: we have to assume this works as expected and was tested by you and ideally also @brianjmurrell as we would like to minimise any possible chance to CI breaking before we finally split out packaging from the server repo

mihawk90:
Sorry, the push might have been a bit premature (although I don't see any reason why it wouldn't work anymore).

I'm still building and still need to rebase the companion-PR for jf-web.
Ran out of time right now but I'll get at it ASAP and comment back here after build and test.

h1dden-da3m0n:
Okay, no worries, thank you for the input brian.
As soon as the new CI takes form it will likely rely on the current way RPMs are built.
However, since the packaging will be decoupled per platform and if one fails not fail the entire publish, we should be able to try things out without influencing any other platform packaging. (i.e. if the RPM build breaks then it would only need a fix in the RPM git repo)

mihawk90:
OK so:

## Tests done
:heavy_check_mark: `make rpms`
:heavy_check_mark: `TARGET=fedora-34-x86_64 make rpms` (although EOL anyway)
:heavy_check_mark: `TARGET=fedora-35-x86_64 make rpms`
:heavy_check_mark: `TARGET=fedora-123 make rpms` (fails like it is supposed to)
:heavy_check_mark: `TARGET=gibberish make rpms` (fails like it is supposed to)
:heavy_check_mark: `TARGET=epel-7-x86_64 make rpms`
:heavy_check_mark: `TARGET=epel-8-x86_64 make rpms`
:heavy_check_mark: `TARGET=rocky+epel-8-x86_64 make rpms`

jf-web is rebased and pushed, local tests building right now and I'll comment over there separately.

deployment/Dockerfile.fedora.amd64
```diff
@@ -1,4 +1,4 @@
-FROM fedora:33
+FROM fedora:36
```
h1dden-da3m0n, line 1: Okay, good thing I tried playing through the workflow of our current CI.
It seems that `fedora:36` does not bundle `make`, nor will it get installed alongside the packages that we install below (odd that automake does not bundle it).
This obviously leads to a build error once we attempt to run the Makefile to build the fedora RPM. 
Regardless, if you add `make` to the `dnf install` call below it builds as you would expect.

Could you please add `make` to the package list below, so that it gets installed and won't error the CI, after that I consider this PR done from my point of view.

h1dden-da3m0n (changes requested):

deployment/Dockerfile.fedora.amd64
mihawk90: Woops, missed this review earlier.

Added and pushed

h1dden-da3m0n (approved):

mihawk90:
I'm not entirely sure what that failing OpenAPI CI is supposed to tell me... anything I need to do here?

h1dden-da3m0n:
> I'm not entirely sure what that failing OpenAPI CI is supposed to tell me... anything I need to do here?

It seems to be behaving a bit odd on this PR, as it really is meant to identify if server changes have any effect on the API spec.
Anyway, this came up while going once more over the PR, but did you change this on purpose and did you ensure that the server still manages to serve web properly without this?

https://github.com/mihawk90/jellyfin/blob/6d5c697183df6bed4d2617c56de4350d00750e4d/fedora/jellyfin.env#L24

mihawk90:
> Anyway, this came up while going once more over the PR, but did you change this on purpose and did you ensure that the server still manages to serve web properly without this?
> 
> https://github.com/mihawk90/jellyfin/blob/6d5c697183df6bed4d2617c56de4350d00750e4d/fedora/jellyfin.env#L24

Yes this was on purpose, as lined out in the companion PR for JF-web:
> * Move web-files to the default location
>   * JF didn't find the web-files when it was run as a user from a terminal. It would check the default location, but no find the files. So we move the files in the default location.
>   * Both the Docker build and Debian package use a sub-directory of jellyfin already, although Debian uses /web and points the env-var there.

I decided to leave the line in there as a comment for reference if anyone *does* want to change it.

h1dden-da3m0n:
Thank you for the clarification, I knew I read the reasoning, but wasn't sure anymore when prompted in the dev chat :sweat_smile:

joshuaboniface (approved):
This has my approval as long as it works :)

crobibero (approved):

PR #7654, opened 2022-04-24
Prefer MetadataProvider enum as provider id key over arbitrary strings

Shadowghost:
The key of provider IDs can be set to arbitrary strings from external sources like NFOs.
To properly handle cases where these strings resemble internally available a value in the `MetadataProvider` enum we should use those values and not the string for best compatibility.

### DIFF
--- a/MediaBrowser.Model/Entities/ProviderIdsExtensions.cs
+++ b/MediaBrowser.Model/Entities/ProviderIdsExtensions.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
+using System.Linq;
 
 namespace MediaBrowser.Model.Entities
 {
@@ -9,6 +10,16 @@ namespace MediaBrowser.Model.Entities
     /// </summary>
     public static class ProviderIdsExtensions
     {
+        /// <summary>
+        /// Case insensitive dictionary of <see cref="MetadataProvider"/> string representation.
+        /// </summary>
+        private static readonly Dictionary<string, string> _metadataProviderEnumDictionary =
+            Enum.GetValues<MetadataProvider>()
+                .ToDictionary(
+                    enumValue => enumValue.ToString(),
+                    enumValue => enumValue.ToString(),
+                    StringComparer.OrdinalIgnoreCase);
+
         /// <summary>
         /// Checks if this instance has an id for the given provider.
         /// </summary>
@@ -108,7 +119,7 @@ namespace MediaBrowser.Model.Entities
         /// <param name="instance">The instance.</param>
         /// <param name="name">The name.</param>
         /// <param name="value">The value.</param>
-        public static void SetProviderId(this IHasProviderIds instance, string name, string value)
+        public static void SetProviderId(this IHasProviderIds instance, string name, string? value)
         {
             if (instance == null)
             {
@@ -125,7 +136,15 @@ namespace MediaBrowser.Model.Entities
                 // Ensure it exists
                 instance.ProviderIds ??= new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
 
-                instance.ProviderIds[name] = value;
+                // Match on internal MetadataProvider enum string values before adding arbitrary providers
+                if (_metadataProviderEnumDictionary.TryGetValue(name, out var enumValue))
+                {
+                    instance.ProviderIds[enumValue] = value;
+                }
+                else
+                {
+                    instance.ProviderIds[name] = value;
+                }
             }
         }
 


MediaBrowser.Model/Entities/ProviderIdsExtensions.cs
```diff
@@ -137,7 +155,25 @@ public static void SetProviderId(this IHasProviderIds instance, string name, str
...
+            {
+                instance.ProviderIds?.Remove(providerString);
+            }
+            else
+            {
+                // Ensure it exists
+                instance.ProviderIds ??= new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+
+                instance.ProviderIds[providerString] = value;
+            }
```
crobibero, line 176: Why duplicate the logic?
`instance.SetProviderId(str, value)` already does this

MediaBrowser.XbmcMetadata/Parsers/BaseNfoParser.cs
```diff
@@ -768,7 +768,8 @@ protected virtual void FetchDataFromXmlNode(XmlReader reader, MetadataResult<T>
 
                         var provider = reader.GetAttribute("type");
                         var id = reader.ReadElementContentAsString();
-                        if (!string.IsNullOrWhiteSpace(provider) && !string.IsNullOrWhiteSpace(id))
+
+                        if (!string.IsNullOrEmpty(provider))
```
crobibero, line 772: Why remove the `id` check? This would allow nfo to unset a provider id

MediaBrowser.XbmcMetadata/Parsers/BaseNfoParser.cs
```diff
@@ -1136,20 +1134,20 @@ private void FetchFromRatingsNode(XmlReader reader, T item)
...
 
-                            using var subtree = reader.ReadSubtree();
-                            FetchFromRatingNode(subtree, item, ratingName);
+                                using var subtree = reader.ReadSubtree();
+                                FetchFromRatingNode(subtree, item, ratingName);
 
-                            break;
-                        }
+                                break;
+                            }
```
crobibero, line 1150: This just seems to change indentation?

Shadowghost: Correct but seems like now it's as it should be? At least all other switch statements are formatted like this too.

MediaBrowser.Model/Entities/ProviderIdsExtensions.cs
Shadowghost: Because that function is used a lot more often all over the code than the one with the string argument.
It skips the Enum matching.

cvium: Why should it skip?

MediaBrowser.XbmcMetadata/Parsers/BaseNfoParser.cs
cvium: The other statements are incorrectly formatted

MediaBrowser.Model/Entities/ProviderIdsExtensions.cs
Shadowghost: Because the input is already a value of that enum. Runnign toString on it and passing it to the function with the string arguent would just match that string against the wole enum again which is unnecessary.

cvium: It's a dictionary lookup with an `O(1)` time complexity. The code duplication is much worse.

MediaBrowser.Model/Entities/ProviderIdsExtensions.cs
```diff
@@ -125,7 +135,15 @@ public static void SetProviderId(this IHasProviderIds instance, string name, str
                 // Ensure it exists
                 instance.ProviderIds ??= new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
 
-                instance.ProviderIds[name] = value;
+                // Match on internal MetadataProvider enum string values before adding arbitrary providers
+                if (metadataProviderEnumDictionary.TryGetValue(name, out string? enumValue))
```
crobibero, line 139:
```suggestion
                if (metadataProviderEnumDictionary.TryGetValue(name, out var enumValue))
```

MediaBrowser.Model/Entities/ProviderIdsExtensions.cs
```diff
@@ -9,6 +10,15 @@ namespace MediaBrowser.Model.Entities
...
     public static class ProviderIdsExtensions
     {
+        /// <summary>
+        /// Case insensitive dictionary of <see cref="MetadataProvider"/> string representation.
+        /// </summary>
+        private static Dictionary<string, string> metadataProviderEnumDictionary = new Dictionary<string, string>(
+            Enum.GetValues(typeof(MetadataProvider))
+                .Cast<MetadataProvider>()
+                .ToDictionary(enumValue => enumValue.ToString(), enumValue => enumValue.ToString()),
+            StringComparer.OrdinalIgnoreCase);
```
crobibero, line 20:
```suggestion
        private static readonly Dictionary<string, string> _metadataProviderEnumDictionary =
            Enum.GetValues<MetadataProvider>()
                .ToDictionary(
                    enumValue => enumValue.ToString(),
                    enumValue => enumValue.ToString(),
                    StringComparer.OrdinalIgnoreCase);
```


also note the change from `metadataProviderEnumDictionary` -> `_metadataProviderEnumDictionary`

crobibero (approved):

1337joe (approved):

almightiest:
This doesn't work with Kodi NFO exports and the newest [TheMovieDb python scraper](https://kodi.wiki/view/Add-on:The_Movie_Database_Python) they recommend to use.  Long story short, they set the type as "themoviedb" instead of "tmdb" so the enum doesn't match.  May need to have some kind of dictionary lookup to map from themoviedb to the tmdb enum?

Shadowghost:
I can only see the long name being used by the rating and that one we handle (not properly but we handle it and use the rating)

almightiest:
Their new scraper sets both the rating name and uniqueid type to "themoviedb" from what I'm seeing. When I clear out the provider id in my Jellyfin metadata, update the NFO to use the new format, and initiate a Jellyfin metadata replace for the movie, it doesn't import the external id.  Once I change the uniqueid type back to "tmdb" (overriding what Kodi exports) it imports again in Jellyfin 10.8.0b3.

Shadowghost:
I tried all available metadata scrapers in the Kodi 20 alpha and the NFO was never written with `themoviedb` as a `uniqueid`.

almightiest:
Weird, maybe something else caused it to get changed.  I did look at the Kodi video119 database (from Kodi v19) and the uniqueid type exists as `themoviedb` after scanning in a movie with their TMDb python scraper.  Maybe the `uniqueid`.`type` is getting translated to `tmdb` on export in v20?

almightiest:
I re-ran some tests and isolated it to just Kodi and their newest python scraper, and it did as you said, `uniqueid`=`tmdb` and `rating`=`themoviedb`.  Sorry for the trouble!

PR #7663, opened 2022-04-26
Search items: Support sql-like for each word of searchTerm

### LINKED ISSUE #1674: Search for series expects exact punctuation between words in title

with_punctuation

no_punctuation

doc-code-hub:
## Introduction

Can't find `Mr. Robot` by searching for `mr robot`?
You are not the only one!

In fact, looking at [this issue](https://github.com/jellyfin/jellyfin/issues/1674) ( #1674 ), I also began to notice that if the titles contain punctuation or particular characters, these must be specified in the search for it to be successful. **_Clearly not good!_**
_When searching for titles that include characters like `.` `,` `"` In the middle (omitting them from searchTerm), the query fails._

To give some examples:
- Searching for `Mr Robot` does not find `Mr. Robot` "
- Searching for `Flash point` does not find `"Flash", point ...`
_and so on ( some more examples in the issue #1674 )..._ :) 

## Changes

To overcome the problem, I introduced a third and final LIKE on the items query.
_This acts, like the previous one, on the CleanName and OriginalTitle columns._

When my query kicks in?
- If the searchTerms are longer than 1
- and if the searchTerms contain 2 or more separate words

My idea is to search with the "**_contains_**" criterion but **_applied to each single word_**.

### What about now?

**Example**
Eg. the searchTerm `Mr Robot` becomes `%Mr%Robot%`, thus correctly extracts:
- `Mr. Robot`
- `Mr Robot`
- `Mr - Robot`
- `Mr, Robot`
- `Mr "Robot"`
_etc_

**Cool side effect**
Furthermore, in this way it becomes possible to extract also by acronym ( not bad imo :) )
Eg. `b b t` extracts `The Big Bang Theory`.

## Issues

fix #1674 

---
Just my second PR on this repo 😄

I do not think that what I have done hides dark sides () and I am convinced that the search experience is now simpler, intuitive and functional!

If you think it may hide any weaknesses, I would like to discuss it :) 
I'm avaliable for anything!


bradbeattie:
I wonder if instead of breaking the queries down by delimiters, if using https://github.com/schiffma/distlib to get Levenshtein or Jaro-Winkler would be preferable. Doing so would better handle typos and misspellings. Though I'm not clear on how to get distlib into Jellyfin...

doc-code-hub:
@bradbeattie 
It would be interesting! I hadn't thought of introducing more advanced search algorithms in fact.

I'm not familiar with the implementation of these search algorithms either sadly.
As soon as I can get back on the project I try to see if I find out anything about it (I have some recent conflicts to resolve anyway).

c4mz:
FYI, I'm not sure if my example is align with your PR but, if I search for the TV series "Dark", it does not show up in the results. I can't get it to show up at all.

sevenrats:
@bradbeattie it's more complex than simple metrics because we also track what attribute matched on the search. So, you would need multiple metric trees, one for every attribute you were tracking, and you would need to query all those metric trees each time you wanted to perform a search.  Not performant.

bradbeattie:
@sevenrats What I'd like to see is search being exposed to plugin capabilities. If motivated plugin devs can replace the standard search with something backed by ElasticSearch/Meilisearch/whatever, we may we'll get a much faster iteration loop _and_ get stemming/fuzzymatching/etc at the same time.

jumoog:
I have the same problem. Another possible solution is to save the CleanName as `mr robot`.

JPVenson:
I really like the idea and the function but it seems a bit problematic to further invest into the SqLite specific code while we are transitioning towards EFCore.

Bond-009:
Closing as we've migrated the DB to EFCore and this PR would have to be rewritten from scratch.

PR #7699, opened 2022-05-03
Fix streambuilder reasons for direct playback checks

Shadowghost:
The checks currently always returned the reason of the first failed DirectPlayProfile but that can be irritating because there may be other profiles which had reasons less resource intense to fix.

This PR fixes that behaviour by actually using the calculated `Rank` property.

I also renamed some variables and methods to properly reflect what they do.


crobibero:
ci is fail

tests/Jellyfin.Model.Tests/Dlna/StreamBuilderTests.cs
```diff
@@ -403,32 +403,39 @@ public async Task BuildVideoItemWithDirectPlayExplicitStreams(string deviceName,
...
                     Assert.Single(val.TargetVideoCodec);
 
                     if (transcodeMode == "DirectStream")
                     {
+                        // Check expected audio codecs (1)
                         if (!targetAudioStream.IsExternal)
                         {
-                            // check expected audio codecs (1)
-                            Assert.DoesNotContain(targetAudioStream.Codec, val.AudioCodecs);
+                            if (val.TranscodeReasons == TranscodeReason.ContainerNotSupported)
```
dmitrylyzo, line 421: Shouldn't it be like this?
```suggestion
                            if ((val.TranscodeReasons & TranscodeReason.ContainerNotSupported) != 0)
```

crobibero (approved):

cvium (approved):

PR #7705, opened 2022-05-04
Add scheduled task to extract subtitles

shemanaev:
"Fetching additional data" is really annoying to see when you're ready for movie with popcorn :)

There is already workaround from @iwalton3 [jellyfin_extract_sub.py](https://gist.github.com/iwalton3/f60f4741f561a742e6f8689a621c9824), but it requires python, cron and all that hassle.

So this is basically the same idea: extract all embedded subtitles in background beforehand.
There is no default trigger since task can be resource consuming and probably not for everyone.

**Changes**

* Add scheduled task to extract subtitles

**Issues**

Addresses #6672

And feature requests:
* [Automatic extraction of subtitles](https://features.jellyfin.org/posts/288/automatic-extraction-of-subtitles)
* [Subtitle improvements](https://features.jellyfin.org/posts/722/subtitle-improvements)

### DIFF
--- a/Emby.Server.Implementations/ScheduledTasks/Tasks/ExtractSubtitlesTask.cs
+++ b/Emby.Server.Implementations/ScheduledTasks/Tasks/ExtractSubtitlesTask.cs
@@ -101,7 +101,7 @@ public class ExtractSubtitlesTask : IScheduledTask
                     var format = stream.Codec;
 
                     // SubtitleEncoder has readers only for these formats, everything else converted to SRT.
-                    if (!_supportedFormats.Contains(format, StringComparer.OrdinalIgnoreCase))
+                    if (!_supportedFormats.Contains(format, StringComparison.OrdinalIgnoreCase))
                     {
                         format = SubtitleFormat.SRT;
                     }
--- a/Emby.Server.Implementations/ScheduledTasks/Tasks/ExtractSubtitlesTask.cs
+++ b/Emby.Server.Implementations/ScheduledTasks/Tasks/ExtractSubtitlesTask.cs
@@ -21,6 +21,8 @@ namespace Emby.Server.Implementations.ScheduledTasks.Tasks;
 /// </summary>
 public class ExtractSubtitlesTask : IScheduledTask
 {
+    private const int QueryPageLimit = 100;
+
     private readonly ILibraryManager _libraryManager;
     private readonly ISubtitleEncoder _subtitleEncoder;
     private readonly ILocalizationManager _localization;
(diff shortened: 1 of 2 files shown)


LogicalPhallacy (approved):
Looks good to me. There's potentially an argument for not hardcoding the subtitle formats like that (or putting it into a more configurable location) but that's a minor nitpick

Shadowghost:
We could also just dump all subs and attachments into an mks container.

crobibero:
I would prefer to have this paginated as it will pull all movies and episodes that have subtitles every run

shemanaev:
> hardcoding the subtitle formats like that

Made a little bit better.

> We could also just dump all subs and attachments into an mks container.

Does it help in this situation? AFAIK subs still must be extracted.

> I would prefer to have this paginated as it will pull all movies and episodes that have subtitles every run

Done.

Emby.Server.Implementations/ScheduledTasks/Tasks/ExtractSubtitlesTask.cs
```diff
@@ -0,0 +1,121 @@
...
+/// </summary>
+public class ExtractSubtitlesTask : IScheduledTask
+{
+    private const int QueryPageLimit = 100;
+
+    private readonly ILibraryManager _libraryManager;
+    private readonly ISubtitleEncoder _subtitleEncoder;
+    private readonly ILocalizationManager _localization;
+    private static readonly BaseItemKind[] _itemTypes = { BaseItemKind.Episode, BaseItemKind.Movie };
+    private static readonly List<string> _supportedFormats = new() { SubtitleFormat.SRT, SubtitleFormat.ASS, SubtitleFormat.SSA };
```
crobibero, line 30:
```suggestion
    private static readonly string[] _supportedFormats = { SubtitleFormat.SRT, SubtitleFormat.ASS, SubtitleFormat.SSA };
```

Emby.Server.Implementations/ScheduledTasks/Tasks/ExtractSubtitlesTask.cs
```diff
@@ -0,0 +1,121 @@
...
+                                     && stream.SupportsExternalStream
+                                     && !stream.IsExternal);
+                foreach (var stream in streams)
+                {
+                    var index = stream.Index;
+                    var format = stream.Codec;
+                    var mediaSourceId = video.Id.ToString("N", CultureInfo.InvariantCulture);
+
+                    // SubtitleEncoder has readers only for these formats, everything else converted to SRT.
+                    if (!_supportedFormats.Contains(format, StringComparer.OrdinalIgnoreCase))
```
crobibero, line 103: Make sure to use `Contains` from `Jellyfin.Extensions`

Emby.Server.Implementations/ScheduledTasks/Tasks/ExtractSubtitlesTask.cs
```diff
@@ -0,0 +1,121 @@
...
+                    {
+                        format = "srt";
+                    }
+
+                    await _subtitleEncoder.GetSubtitles(video, mediaSourceId, index, format, 0, 0, false, cancellationToken).ConfigureAwait(false);
+                }
+
+                completedVideos++;
+                double percent = (double)completedVideos / numberOfVideos;
+                progress.Report(100 * percent);
```
crobibero, line 113:
```suggestion
                progress.Report(100d * completedVideos / numberOfVideos);
```

Emby.Server.Implementations/ScheduledTasks/Tasks/ExtractSubtitlesTask.cs
```diff
@@ -0,0 +1,121 @@
...
+
+                var streams = video.GetMediaStreams()
+                    .Where(stream => stream.IsTextSubtitleStream
+                                     && stream.SupportsExternalStream
+                                     && !stream.IsExternal);
+                foreach (var stream in streams)
+                {
+                    var index = stream.Index;
+                    var format = stream.Codec;
+                    var mediaSourceId = video.Id.ToString("N", CultureInfo.InvariantCulture);
```
crobibero, line 100: Pull this line out of the inner loop

Emby.Server.Implementations/ScheduledTasks/Tasks/ExtractSubtitlesTask.cs
```diff
@@ -0,0 +1,121 @@
...
+
+                    // SubtitleEncoder has readers only for these formats, everything else converted to SRT.
+                    if (!_supportedFormats.Contains(format, StringComparer.OrdinalIgnoreCase))
+                    {
+                        format = "srt";
```
crobibero, line 105:
```suggestion
                        format = SubtitleFormat.SRT;
```

Emby.Server.Implementations/ScheduledTasks/Tasks/ExtractSubtitlesTask.cs
shemanaev: Is it enough to add `using`?

Shadowghost:
> > We could also just dump all subs and attachments into an mks container.
> 
> Does it help in this situation? AFAIK subs still must be extracted.

Extartcion from an mks is faster but ultimately it would only result in having another intermediate format. How you're doing it now is the correct way to do it.

Emby.Server.Implementations/ScheduledTasks/Tasks/ExtractSubtitlesTask.cs
crobibero: Yes, it should be enough to just add the using statement

crobibero (approved):

Emby.Server.Implementations/ScheduledTasks/Tasks/ExtractSubtitlesTask.cs
```diff
@@ -0,0 +1,121 @@
...
+                    .Where(stream => stream.IsTextSubtitleStream
+                                     && stream.SupportsExternalStream
+                                     && !stream.IsExternal);
+                foreach (var stream in streams)
+                {
+                    var index = stream.Index;
+                    var format = stream.Codec;
+
+                    // SubtitleEncoder has readers only for these formats, everything else converted to SRT.
+                    if (!_supportedFormats.Contains(format, StringComparer.OrdinalIgnoreCase))
```
crobibero, line 104:
```suggestion
                    if (!_supportedFormats.Contains(format, StringComparison.OrdinalIgnoreCase))
```

I was wrong, also need to change the second parameter

crobibero:
Upon internal discussion we decided that we would rather have this as a plugin as it relies on a side-effect. 
Plugin is available here: https://github.com/jellyfin/jellyfin-plugin-subtitleextract and in the official Jellyfin repo. 
It supports Jellyfin servers 10.8.1+

I've credited you in the original commit

PR #7708, opened 2022-05-04
Add Vulkan filtering support for AMD VAAPI (Vega/gfx9+)

nyanmisaka:
**Changes**
- Add Vulkan filtering support for AMD VAAPI (**Vega and newer/gfx9+, kernel>= 5.11**),
 including HWA subtitles burn-in and HDR tonemapping without using AMD's buggy OpenCL stack.

- Dolby Vision to SDR tonemapping is partially supported on AMD VAAPI.

**Note**
- This PR requires `jellyfin-ffmpeg5` 5.0.1-2 and newer.

- Linux kernel version should be equal or newer than 5.11 since some DRM changes are newly added. (`uname -r`)

- For Polaris and older/gfx8- cards, the essential vulkan extension `VK_EXT_image_drm_format_modifier` [hasn't been implemented](https://gitlab.freedesktop.org/mesa/mesa/-/issues/5882), which requires many changes in Mesa and linux kernel.


MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
```diff
@@ -279,6 +304,37 @@ private string GetEncoderPathFromDirectory(string path, string filename, bool re
             }
         }
 
+        private Version GetLinuxKernelVersion()
```
Bond-009, line 307: You should be able to use `Environment.OSVersion.Version`

nyanmisaka: Oh, thanks. I actually missed such an easy method. 🤣

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -3669,14 +3725,25 @@ public static string GetHwTonemapFilter(EncodingOptions options, string hwTonema
...
                 // Intel iHD path, with extra vpp tonemap and overlay support.
-                return GetVaapiFullVidFiltersPrefered(state, options, vidDecoder, vidEncoder);
+                return GetIntelVaapiFullVidFiltersPrefered(state, options, vidDecoder, vidEncoder);
+            }
+
+            // prefered vaapi + vulkan filters pipeline
+            if (_mediaEncoder.IsVaapiDeviceAmd
+                && isVaapiVkSupported
+                && _mediaEncoder.IsVaapiDeviceSupportVulkanFmtModifier
+                && Environment.OSVersion.Version != null
```
Bond-009, line 3735: I don't this this can be null?
https://source.dot.net/#System.Private.CoreLib/OperatingSystem.cs,74

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -2760,22 +2789,41 @@ public static string GetHwTonemapFilter(EncodingOptions options, string hwTonema
...
+                {
+                    args += ":range={6}";
+                }
+
+                if (string.Equals(options.TonemappingAlgorithm, "bt2390", StringComparison.OrdinalIgnoreCase))
+                {
+                    algorithm = "bt.2390";
+                }
+
+                if (string.Equals(options.TonemappingAlgorithm, "none", StringComparison.OrdinalIgnoreCase))
```
crobibero, line 2819:
```suggestion
                else if (string.Equals(options.TonemappingAlgorithm, "none", StringComparison.OrdinalIgnoreCase))
```

MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
```diff
@@ -70,6 +70,16 @@ public class MediaEncoder : IMediaEncoder, IDisposable
         private bool _isVaapiDeviceAmd = false;
         private bool _isVaapiDeviceInteliHD = false;
         private bool _isVaapiDeviceInteli965 = false;
+        private bool _isVaapiDeviceSupportVulkanFmtModifier = false;
+
+        private string[] _vulkanFmtModifierExts = {
```
crobibero, line 75:
```suggestion
        private static string[] _vulkanFmtModifierExts = {
```

crobibero (approved):

nyanmisaka:
Any updates on this? We need to merge this into master asap and get more feedback.

ri0thamus:
Just to add my two cents to this: I have tried out @nyanmisaka's Docker image for this and I was able to play HEVC 4k files whereas I was not able to do so by just attempting this with OpenCL (the open-source version). I still had trouble with AV1, but that is an issue with my hardware and not the patch.

TianZerL:
Also confirmed that the Vulkan tonemapping works well in AMD Vega3, which is almost impossible to install AMD's official OpenCL runtime in Docker.
And maybe we can add support for Mesa OpenCL (Clover) for AMD GPUs too, which is open source and much more convenient for lots of AMD GPUs than using the official closed source one, especially in Docker. This may allow more AMD GPUs and Linux distribution to enable tonemapping.
The current OpenCL implementation of jellyfin will detect whether the vendor is `Advanced Micro Devices Inc`, thus refusing to use Mesa's OpenCL whose vendor is `Mesa`.

nyanmisaka:
> Also confirmed that the Vulkan tonemapping works well in AMD Vega3, which is almost impossible to install AMD's official OpenCL runtime in Docker. And maybe we can add support for Mesa OpenCL (Clover) for AMD GPUs too, which is open source and much more convenient for lots of AMD GPUs than using the official closed source one, especially in Docker. This may allow more AMD GPUs and Linux distribution to enable tonemapping. The current OpenCL implementation of jellyfin will detect whether the vendor is `Advanced Micro Devices Inc`, thus refusing to use Mesa's OpenCL whose vendor is `Mesa`.

There's no image support in Clover and it doesn't support OpenCL 1.2, which is the minimal requirement in FFmpeg.
Also it doesn't support interop with VAAPI but Vulkan RADV supports this well.

TianZerL:
> There's no image support in Clover and it doesn't support OpenCL 1.2, which is the minimal requirement in FFmpeg.

Thank you for your reply. I tried to install mesa-opencl-icd in the container and i suppose it works, `clinfo` can identify the OpenCL device. And Clover actually has supported OpenCL 1.2 since Mesa 20.3, see [here](https://mesamatrix.net/#OpenCL1.2). (clinfo reports OpenCL1.1, because Clover does not support `printf()`)

> Also it doesn't support interop with VAAPI but Vulkan RADV supports this well.

This is very troublesome. So Clover cannot be used for tonemapping now. 
Anyway, thank you for your excellent work for AMD GPUs!

nyanmisaka:
> Thank you for your reply. I tried to install `mesa-opencl-icd` in the container and i suppose it works, `clinfo` can identify the OpenCL device. And Clover actually has supported OpenCL 1.2 since Mesa 20.3, see [here](https://mesamatrix.net/#OpenCL1.2). (clinfo reports OpenCL1.1, because Clover does not support `printf()`)

Lacking of image support in Clover is the elephant in the room.

> This is very troublesome. So Clover cannot be used for tonemapping now. Anyway, thank you for your excellent work for AMD GPUs!

That's right. Either using AMD ROCr/legacy OpenCL or using Vulkan compute via RADV open source driver.

ri0thamus:
> > Lacking of image support in Clover is the elephant in the room.
> 
> Just for verification, the [matrix](https://mesamatrix.net/#OpenCL1.2_Extension_clCreateImage) also reports that `clCreateImage` is supported, so I guess `image` feature is at least partially supported since Mesa 20.3?

Unfortunately, no. This is from clinfo on my system using Mesa 22.3-devel:
Image support                                   No

nyanmisaka:
> > Lacking of image support in Clover is the elephant in the room.
> 
> Just for verification, the [matrix](https://mesamatrix.net/#OpenCL1.2_Extension_clCreateImage) also reports that `clCreateImage` is supported, so I guess `image` feature is at least partially supported since Mesa 20.3?

You're likely to get "No image support in OpenCL driver (zero supported image formats)" from [FFmpeg](https://github.com/FFmpeg/FFmpeg/blob/dd2ea014ef273157fe9a0e928e77841fbbee9b2f/libavutil/hwcontext_opencl.c#L1525) when you invoke the buffer allocator in `hwcontext_opencl` since image formats in radeonsi backend is not implemented. Also you can check the `clinfo` to see if the driver have full image support.

RedHat devs [tried to add support for this](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/7241) last year but he start a new OpenCL project Rusticl soon, which is aimed to supersede Clover in the future.

ri0thamus:
Is the only thing that this PR is waiting on is for secondary approval? With Ubuntu 22.10 coming out this month and its use of the 5.19 kernel this would be a big win for AMD users.

sand-bit:
I have been using @nyanmisaka's jellyfin with this patch included on a Mesa + AMD setup, and the included tonemapping has been helpful. It would be appreciated if @Bond-009 could also review this patch!

nyanmisaka:
> Is the only thing that this PR is waiting on is for secondary approval? With Ubuntu 22.10 coming out this month and its use of the 5.19 kernel this would be a big improvement for AMD users.

Almost right but we still need this [jellyfin-ffmpeg PR](https://github.com/jellyfin/jellyfin-ffmpeg/pull/189) to fix a modifier issue in subtitle burn-in on AMD APU. Also the minimum required kernel version has been changed to 5.15 LTS.

Bond-009 (approved):

sand-bit:
It would be nice to get this into the 10.8.z branch or the next 10.8.6 release.

nyanmisaka:
> It would be nice to get this into the 10.8.z branch or the next 10.8.6 release.

You can already try this PR in the unstable branch.
It is unlikely that this PR will be backported to 10.8.z due to massive changes.

nyanmisaka:
> This PR actually does more than just adding support for AMD GPUs, it adds support Vulkan filtering to _all_ Vulkan capable GPUs supporting the required extensions. This should be renamed to
> 
> > Add Vulkan filtering support for VAAPI devices supporting `VK_EXT_image_drm_format_modifier`

Intel and Nvidia GPU don't require Vulkan to do zero-copy HDR transcoding, they have an excellent and readily available OpenCL/CUDA runtime on Linux. So this PR is only open for AMD.

aviallon:
@nyanmisaka you forgot about ARM devices.
And perhaps I'll want to use this Vulkan fltering for everything later (including Intel and NVIDIA GPUs)

nyanmisaka:
We need a cross platform solution for Intel, AMD and NVIDIA.
But the current Vulkan interop support in ffmpeg has some limitations.

|| Intel QSV<->VK | NVIDIA CUDA<->VK |
|--|--|--|
|Linux   | ✔️ | ❌|
|Win32 | ❌|✔️ |

However our custom OpenCL and CUDA filters covered these use cases perfectly.

Also the v3dv gpu in RPi 4 is too weak for transcoding with libplacebo on the fly.

Issue #7130, opened 2022-01-07
[Issue]: Directory watch disabled if subdirectory (e.g. lost+found) is unreadable


niclan:
### Please describe your bug

This morning I've reviewed my jellyfin logs for the first time.  I've been able to correct some permission issues and get them quite quiet. But this remains:

```
jan. 07 09:40:15 gont jellyfin[3320044]: System.UnauthorizedAccessException: Access to the path '/space/dvds/lost+found' is denied.
jan. 07 09:40:15 gont jellyfin[3320044]:  ---> System.IO.IOException: Permission denied
jan. 07 09:40:15 gont jellyfin[3320044]:    --- End of inner exception stack trace ---
jan. 07 09:40:15 gont jellyfin[3320044]: [09:40:15] [INF] Stopping directory watching for path /space/dvds
```

The lost+found directory should not be readable to non-admins and should in any case not be catalogued. It seems unnecessary to let a permission denied exception disable watching a directory hierarchy.

Regards,
  Nicolai

### Jellyfin Version

10.7.7

### Environment

```markdown
- OS: Linux, Ubuntu 20.04.3 LTS
- Virtualization: Running directly on host, not in VM or container
- Clients: This is a storage backend issue
- Browser: This is a storage backend issue
- FFmpeg Version: This is a storage backend issue
- Playback Method: This is a storage backend issue
- Hardware Acceleration: This is a storage backend issue
- Plugins: This is a storage backend issue
- Reverse Proxy: This is a storage backend issue
- Base URL: This is a storage backend issue
- Networking: This is a storage backend issue
- Storage: Local disk, ext4 filesystem
```

### Jellyfin logs

```shell
jan. 07 09:40:15 gont jellyfin[3320044]: [09:40:15] [ERR] Error in Directory watcher for: /space/dvds
jan. 07 09:40:15 gont jellyfin[3320044]: System.UnauthorizedAccessException: Access to the path '/space/dvds/lost+found' is denied.
jan. 07 09:40:15 gont jellyfin[3320044]:  ---> System.IO.IOException: Permission denied
jan. 07 09:40:15 gont jellyfin[3320044]:    --- End of inner exception stack trace ---
jan. 07 09:40:15 gont jellyfin[3320044]: [09:40:15] [INF] Stopping directory watching for path /space/dvds
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

niclan:
[...]

PR #7724, opened 2022-05-08
Don't let permission denied kill library scans

jtcasper:
**Changes**
I have a single `./Movies` directory that contains symlinks to many
other directories, some of which are in shared paths. Other daemons
sometimes overwrite permissions on these paths and prevent Jellyfin from
being able to scan. Because this exception is unhandled within a LINQ
statement it non-deterministically kills metadata refresh for my entire Movies
library.

Running with this patch has allowed me to at least add + scan for
movies in directories that Jellyfin can currently access and gives me
some notification of the symlinks failing due to permissions errors.

**Issues**
May be related to #7130

### DIFF
--- a/Emby.Server.Implementations/IO/ManagedFileSystem.cs
+++ b/Emby.Server.Implementations/IO/ManagedFileSystem.cs
@@ -262,6 +262,10 @@ namespace Emby.Server.Implementations.IO
                             _logger.LogError(ex, "Reading the file size of the symlink at {Path} failed. Marking the file as not existing.", fileInfo.FullName);
                             result.Exists = false;
                         }
+                        catch (UnauthorizedAccessException ex)
+                        {
+                            _logger.LogError(ex, "Reading the file at {Path} failed due to a permissions exception.", fileInfo.FullName);
+                        }
                     }
                 }
 


jtcasper:
I'd _like_ to add unit tests for this but I don't write C# in any capacity. I was just going to set a mock thrown in `jellyfin/tests/Jellyfin.Server.Implementations.Tests/IO/ManagedFileSystemsTests.cs` but I just don't have enough familiarity with the frameworks to do that.

Running the actual code on my box and verifying it works is not a test, but it's what I did.

Shadowghost:
Since this is a problem in 10.8 you should consider targeting the release-10.8.z branch to get this bug fix (at least I consider this to be a bug) into the upcoming release.

jtcasper:
@Shadowghost o7

jtcasper:
If you (a maintainer?) or other maintainers think I should publish this for review, let me know. I don't mind changing any of the verbiage for the log message.

crobibero (approved):

jtcasper:
I can't tell what happened with the `Check Backport` check, it seems to be attempting to merge a branch from inside `jellyfin/jellyfin` but using the branch name `origin/perms` which should be in `jtcasper/jellyfin`

crobibero:
Don't worry about that check, it doesn't currently work

cvium (approved):

Issue #7656, opened 2022-04-24
[Issue]: Jellyfin does not accurately label episodes numbered over 999


Sarcasthik:
### Please describe your bug

Hello there.

Here's the issue : I have an anime that has now over 1000+ episodes but Jellyfin does not seem to count above 999. 
If the episodes are labeled "My Super Anime 1000" and then "My Super Anime 1001" both of them are labeled as episode 100.

Now this is not a big issue so far but whenever I add a new episode it does not show up in the "Recently added" because Jellyfin thinks I have already watched it since it is also labeled episode 100 and not 100X.
Maybe it has already been solved in the 10.8 Beta but better safe than sorry.

Thank you !

### Jellyfin Version

Other

### if other:

10.7.5

### Environment

```markdown
- OS: Windows Server 2019
- Virtualization: no
- Clients: Chrome/Firefox
- Browser: Chrome/Firefox
- FFmpeg Version:
- Playback Method: Direct
- Hardware Acceleration:
- Plugins:
- Reverse Proxy: none
- Base URL:
- Networking:
- Storage: local
```

### Please attach any screenshots here

Here it's ok until episode 999 :
![image](https://user-images.githubusercontent.com/43416863/164997638-41c18032-233f-4cd5-8f3f-137bfbe096f9.png)

After episode 1000 it begins to bug :
![image](https://user-images.githubusercontent.com/43416863/164997679-baff40ef-379c-459f-8000-5770c646bccf.png)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Sarcasthik:
It goes like this "D:\Jellyfin\Séries Anime\One Piece\Saison 21\One Piece 100X VOSTFR"
I made my own seasons per arcs and it went well until recently. Could that be the issue ?

LewkyB:
@Sarcasthik Thanks for giving me that path. Looks like it was just set to check 3 digits instead of 4. I made a small change and it seems to be working great.

Sarcasthik:
Nice, thank you ! I know it's a small thing but you guys rock !

neilsb:
Fixed by #7741

PR #7741, opened 2022-05-12
Fix to allow for episode numbering over 999 in certain scenarios

LewkyB:
**Changes**
The capture group would only take up to 3 digits. When the episode count would get to 1000, it would only take out 100. I've increased the capture group to 4 digits and verified that it fixed the issue by adding some inline data to `GetEpisodeNumberFromFileTest` and also verifying that it worked on my local.

**Issues**
#7656


crobibero (approved):

Bond-009 (approved):
This may match years now, but since all tests succeed :man_shrugging:

gtrcoi:
is this not implemented into 10.8.11 yet? I'm still having the same issue

`E:\Videos\TV Shows\One Piece\One Piece - 1007 - Zoro`s Pursuit! Ice Oni Tag! [6B8A79B5].mkv` pulls episode 100 metadata for example

PR #7749, opened 2022-05-15
fix: disable "Automatically add to collection" by default

### LINKED ISSUE #7747: [Issue]: A library named "Collections" keeps reappearing even after manual deletion
### Please describe your bug

A library named "Collections" was automatically created by Jellyfin:
image

I removed it from settings. But after a while Jellyfin recreated it.

The path the library points to is `/var/lib/jellyfin/root/default/Collections`.

The content of it is the "The Lord of the Rings Collection", a colleciton of 3 movies. The actually file of each episode is stored as an individual movie in another library. I think this is some auto-merging feature that went buggy.

I don't know if the bug appeared after one update. But I'm pretty sure it wasn't there until recent weeks.

Plugins enabled:
image

### Jellyfin Version

10.8.0

### if other:

_No response_

### Environment

```markdown
- OS: Debian testing (bookworm)
- Virtualization: None
- FFmpeg Version: jellyfin-ffmpeg5 5.0.1-3-bullseye
- Plugins: See the screenshot above
- Base URL: /
- Storage: local
```

### Jellyfin logs

I don't see anything useful:
```shell
/var/log/jellyfin # rg 'Collections'
jellyfin20220511.log
2249:[2022-05-11 20:15:02.001 -07:00] [INF] Removing item, Type: "CollectionFolder", Name: "Collections", Path: "/var/lib/jellyfin/root/default/Collections", Id: [REDACTED-UUID]
2250:[2022-05-11 20:15:02.171 -07:00] [WRN] Slow HTTP Response from "https://[REDACTED-DOMAIN]/Library/VirtualFolders?refreshLibrary=true&name=Collections" to "[REDACTED-IP]" in 0:00:00.9389611 with Status Code 204
```

### Code of Conduct

cvium:
**Changes**
Disabled the feature by default. It's not something most people would want enabled.

**Issues**
Fixes #7747

### DIFF
--- a/MediaBrowser.Model/Configuration/LibraryOptions.cs
+++ b/MediaBrowser.Model/Configuration/LibraryOptions.cs
@@ -17,7 +17,7 @@ namespace MediaBrowser.Model.Configuration
             RequirePerfectSubtitleMatch = true;
             AllowEmbeddedSubtitles = EmbeddedSubtitleOptions.AllowAll;
 
-            AutomaticallyAddToCollection = true;
+            AutomaticallyAddToCollection = false;
             EnablePhotos = true;
             SaveSubtitlesWithMedia = true;
             EnableRealtimeMonitor = true;


nyanmisaka (approved):

crobibero (approved):

1337joe:
Just tested this - changing this value doesn't actually default the checkbox to false in library options when creating a new library.

Looks like it's enabled by default [in web](https://github.com/jellyfin/jellyfin-web/blob/dba10f923550ec39aab1437ec39ea9db854f15ac/src/components/libraryoptionseditor/libraryoptionseditor.template.html#L53), which should pick up what the server sends as the library options, but a new library only gets a `LibraryOptionsResultDto`, not a full `VirtualFolderInfo` with all the library option values.

Changing it to actually query for a default `LibraryOptions` to initialize new libraries from is beyond me at the moment, do we want to just change that line in web to match the default on the server side?

cvium:
> Changing it to actually query for a default `LibraryOptions` to initialize new libraries from is beyond me at the moment, do we want to just change that line in web to match the default on the server side?

Sure

Issue #7267, opened 2022-02-02
[Issue]: Audio missing in LiveTV with m3u tuner


hjschmidt99:
### Please describe your bug

Some german broadcasters (e.g. ZDF, ZDFneo, ZDFinfo, 3sat, Phoenix)) provide m3u8 files with seperate video and audio streams. 
For these programs Jellyfin LiveTV is missing audio on play or record.
Direct recording of the m3u8 URL using ffmpeg to a file works nicely with audio.

Attached zip file:
- Live streams.m3u8: My collection of URLs for the m3u tuner
- record.cmd: downlaod ZDF m3u8 file and record (copy) the stream with ffmpeg
- zdf.m3u8: the downloaded m3u8 file for ZDF showing the diffenert stream URLs for video and audio.
[test.zip](https://github.com/jellyfin/jellyfin/files/7989776/test.zip)

### Jellyfin Version

10.7.7

### Environment

```markdown
- OS: Win10 Pro
- Virtualization: ---
- Clients: Browser
- Browser: Firefox 96.0.3
- FFmpeg Version:  n4.3.2-Jellyfin
- Playback Method: Direct Play
- Hardware Acceleration: ---
- Plugins: ---
- Reverse Proxy: ---
- Base URL: ---
- Networking: Internet 250Mb/s
- Storage: local
```

### Jellyfin logs

```shell
[21:23:43] [INF] [56] Jellyfin.Api.Helpers.MediaInfoHelper: User policy for hjs1. EnablePlaybackRemuxing: True EnableVideoPlaybackTranscoding: True EnableAudioPlaybackTranscoding: True
[21:23:43] [INF] [56] Jellyfin.Api.Helpers.MediaInfoHelper: Profile: Unknown Profile, Path: http://zdf-hls-15.akamaized.net/hls/live/2016498/de/veryhigh/master.m3u8, isEligibleForDirectPlay: True, isEligibleForDirectStream: True
[21:23:43] [INF] [56] Jellyfin.Api.Helpers.MediaInfoHelper: Profile: Unknown Profile, No video direct play profiles found for http://zdf-hls-15.akamaized.net/hls/live/2016498/de/veryhigh/master.m3u8 with codec Unknown codec
[...]

PR #7776, opened 2022-05-21
Don't ignore DTS in M3UTuner to avoid missing audio

hjschmidt99:
Tested locally on win10 using the m3u8 file attached to the bug.
The offending channels now have audio in live TV and recordings.

**Changes**
Set IgnoreDTS to false

**Issues**
Fixes https://github.com/jellyfin/jellyfin/issues/7267


Shadowghost:
Since this is a bugfix you may want to target the 10.8.x release branch.

PR #7781, opened 2022-05-21
Fix closing livetv stream on session end

crobibero:
Fixes https://github.com/jellyfin/jellyfin/issues/7664

### DIFF
--- a/Emby.Server.Implementations/Session/SessionManager.cs
+++ b/Emby.Server.Implementations/Session/SessionManager.cs
@@ -413,6 +413,7 @@ namespace Emby.Server.Implementations.Session
             session.PlayState.IsPaused = info.IsPaused;
             session.PlayState.PositionTicks = info.PositionTicks;
             session.PlayState.MediaSourceId = info.MediaSourceId;
+            session.PlayState.LiveStreamId = info.LiveStreamId;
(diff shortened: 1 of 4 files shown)


Emby.Server.Implementations/Session/SessionManager.cs
```diff
@@ -329,13 +329,17 @@ public void OnSessionControllerConnected(SessionInfo session)
...
+        public async Task CloseIfNeededAsync(SessionInfo session)
         {
             if (!session.SessionControllers.Any(i => i.IsSessionActive))
             {
                 var key = GetSessionKey(session.Client, session.DeviceId);
 
                 _activeConnections.TryRemove(key, out _);
+                if (!string.IsNullOrEmpty(session.PlayState?.LiveStreamId))
+                {
+                    await _mediaSourceManager.CloseLiveStream(session.PlayState?.LiveStreamId).ConfigureAwait(false);
```
cvium, line 341: I don't remember if `.Value` is required, but the `?` is unnecessary
```suggestion
                    await _mediaSourceManager.CloseLiveStream(session.PlayState.LiveStreamId).ConfigureAwait(false);
```

cvium (approved):

Bond-009 (approved):

PR #7828, opened 2022-05-31
Fix Dolby Vision profile 5 and 8 to SDR HW tone-mapping

nyanmisaka:
**Changes**
- Fix Dolby Vision profile 5 and 8 to SDR HW tone-mapping

**Note**
- `Enable enhanced NVDEC decoder` should be enabled on NV.
- `Prefer OS native DXVA or VA-API hardware decoders` should be enabled on Intel.

- CUDA/OpenCL runtime is required as always.
- Intel VPP tonemap doesn't support DoVi.
- The latest `jellyfin-ffmpeg5` is required.
https://github.com/jellyfin/jellyfin-ffmpeg/pull/151


MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -144,15 +144,27 @@ private bool IsCudaFullSupported()
...
+                var vidDecoder = GetHardwareVideoDecoder(state, options) ?? string.Empty;
+                var isSwDecoder = string.IsNullOrEmpty(vidDecoder);
+                var isNvdecDecoder = vidDecoder.Contains("cuda", StringComparison.OrdinalIgnoreCase);
+                var isVaapiDecoder = vidDecoder.Contains("vaapi", StringComparison.OrdinalIgnoreCase);
+                var isD3d11vaDecoder = vidDecoder.Contains("d3d11va", StringComparison.OrdinalIgnoreCase);
+                return isSwDecoder || isNvdecDecoder || isVaapiDecoder || isD3d11vaDecoder;
+            }
+
+            return string.Equals(state.VideoStream.ColorTransfer, "smpte2084", StringComparison.OrdinalIgnoreCase)
+                    || string.Equals(state.VideoStream.ColorTransfer, "arib-std-b67", StringComparison.OrdinalIgnoreCase);
```
cvium, line 167:
```suggestion
            return string.Equals(state.VideoStream.ColorTransfer, "smpte2084", StringComparison.OrdinalIgnoreCase)
                   || string.Equals(state.VideoStream.ColorTransfer, "arib-std-b67", StringComparison.OrdinalIgnoreCase);
```

cvium (approved):

MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
```diff
@@ -372,7 +374,11 @@ public Task<MediaInfo> GetMediaInfo(MediaInfoRequest request, CancellationToken
 
             string analyzeDuration = string.Empty;
 
-            if (request.MediaSource.AnalyzeDurationMs > 0)
+            if (!string.IsNullOrEmpty(_ffmpegAnalyzeDuration))
+            {
+                analyzeDuration = "-analyzeduration " + _ffmpegAnalyzeDuration;
+            }
+	    else if (request.MediaSource.AnalyzeDurationMs > 0)
```
crobibero, line 381:
```suggestion
            else if (request.MediaSource.AnalyzeDurationMs > 0)
```

(I think I got the spacing correct)

crobibero (approved):

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -54,11 +56,13 @@ public class EncodingHelper
...
             IApplicationPaths appPaths,
             IMediaEncoder mediaEncoder,
-            ISubtitleEncoder subtitleEncoder)
+            ISubtitleEncoder subtitleEncoder,
+            IConfiguration config)
         {
             _appPaths = appPaths;
             _mediaEncoder = mediaEncoder;
             _subtitleEncoder = subtitleEncoder;
+            _ffmpegAnalyzeDuration = config.GetValue<string>(Controller.Extensions.ConfigurationExtensions.FfmpegAnalyzeDurationKey) ?? string.Empty;
```
cvium, line 65: Won't this require restarting Jellyfin every time you change the value?

nyanmisaka: Yes, change env requires a restart to take effect.
But our default value 200M is more than enough.

https://github.com/jellyfin/jellyfin/blob/990e4e8e58e2d15f9bbd010e06915a40c9180235/Emby.Server.Implementations/ConfigurationOptions.cs#L19

cvium: But by using IConfiguration to retrieve the value every time we could in theory avoid it

nyanmisaka: I have changed it to get from GetFFmpegAnalyzeDuration(). https://github.com/jellyfin/jellyfin/blob/4fa7f49ffc6cec2a5be3cf9e5ff0fb727815e6b1/MediaBrowser.Controller/Extensions/ConfigurationExtensions.cs#L79-L85

Bond-009 (approved):
Resolve @cviums comment and this should be good to go

cvium (approved):

PR #7842, opened 2022-06-04
Backport all dependency updates

crobibero:
Ensures all dependencies are in line with what is currently in master, besides OptimizedPriorityQueue.

That PR was closed last year without comment, so unsure what the reason was (https://github.com/jellyfin/jellyfin/pull/5329)

Note that the following were not applied as they have breaking changes:
https://github.com/jellyfin/jellyfin/pull/7658
https://github.com/jellyfin/jellyfin/pull/7345


joshuaboniface (approved):
YOLO

Bond-009 (approved):

PR #7843, opened 2022-06-04
Refresh live TV channel image when remapped (issue #7834)

### LINKED ISSUE #7834: [Issue]: In Live TV, refreshing guide data doesn't update channel images after remapping them manually
### Please describe your bug

I set up HD Homerun and Schedulesdirect, let the guide refresh and found a lot of channels incorrectly mapped. When I manually remapped them I tried refreshing it a few times, although the channel content was now correct, the image didn't get corrected.

I stepped through `LiveTvManager.GetChannelAsync` and found where it says:

```cs
if (!item.HasImage(ImageType.Primary))
```

So the channel already has a primary image (the incorrect one from the auto-mapped channel) and this seems to stop it from re-fetching. Should there be some kind of comparison between the old/new URL strings so it can detect when the cached image is wrong? I am happy to look into this further if this seems like the right direction.

### Jellyfin Version

Other

### if other:

master branch

### Environment

```markdown
- OS: MacOS
- Virtualization:
- Clients: Browser
- Browser: Safari 15.5
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

danielearwicker:
**Changes**

In `GetChannelAsync` the `if (!item.HasImage(ImageType.Primary))` check means once a channel has an image, it is never updated. So it continues to have the wrong image even after it has been manually remapped to the correct channel by the user.

To know when the user has done this it seems like we need to store the URL that was fetched to save the local cached image for the channel. If the URL from the EPG is different from this, we should clear the primary image so it is refetched.

That's almost the whole change, except that probably to avoid doing a lot of unnecessary work the existing code in `UpdateImagesAsync` never updates images for any live TV shows or channels. I've left that as it is except now it only applies to shows (of which there are huge numbers) but not channels (of which there are relatively few).

**Issues**

Fixes #7834


jurgenhaas:
Is there a way I could patch my running Jellyfin installation to give this a try?

Shadowghost:
@danielearwicker it would be appreciated if you can resolve the conflicts.

Shadowghost (approved):

1337joe (approved):

Emby.Server.Implementations/Library/LibraryManager.cs
```diff
@@ -1839,8 +1839,14 @@ public async Task UpdateImagesAsync(BaseItem item, bool forceUpdate = false)
             var outdated = forceUpdate
                 ? item.ImageInfos.Where(i => i.Path is not null).ToArray()
                 : item.ImageInfos.Where(ImageNeedsRefresh).ToArray();
-            // Skip image processing if current or live tv source
-            if (outdated.Length == 0 || item.SourceType != SourceType.Library)
+
+            var parentItem = item.GetParent();
+            var isLiveTvShow = item.SourceType != SourceType.Library &&
+                               parentItem != null &&
```
Bond-009, line 1845:
```suggestion
                               parentItem is not null &&
```

Bond-009 (changes requested):

JPVenson (approved):

Bond-009:
Superseded by #10841

Issue #6164, opened 2021-06-08
Custom Genres for TV/Film not working properly


moosepolice:
**Describe the bug**
Adding a custom genre will show it on the TV/Movie description page but are not being displayed in the Genre section. Clicking on the custom genre will take you to a new page with an endless loading icon. Custom genres are displayed in the Library filter and show content if the filter is selected. Affects both TV and Movies.

**System (please complete the following information):**
 - OS: Ubuntu
 - Virtualization: Linux (no docker)
 - Clients: Browser, Android
 - Browser: Firefox 89
 - Jellyfin Version: 10.7.5
 - Installed Plugins: AniDB, AudioDB, Fanart, OMDb, Open Subtitles, TheTVDB
 - Reverse Proxy: none
 - Base URL: none
 - Storage: local

**To Reproduce**
1. Go to a movie or a tv show and click edit Metadata
2. Add a new custom genre like Superhero or Cyberpunk, then save
3. Click on the new genre in the show or movie description page, this will trigger endless loading gif
4. Go to TV show or movie main library, select Genre in the top menu. You will not find the new custom genre
5. Select the library filter, you will be able to see the new custom genre and clicking on it will show items with said genre
6. The issue is apparent with both web client as well as the Android app.
7. For some reason, adding "Anime" as a new custom genre work, but nothing else I have tried.

**Expected behavior**
A new custom genre should act as the default genres fetched from metadata sources. It should appear in the Genre page and clicking on it on an item's description page should take you to the Genre page.

**Server Logs**
This appears in the logs when you click on the custom genre in a tv show or movie description page:

```
[2021-06-08 12:23:30.618 -04:00] [ERR] Error processing request. URL "GET" "/Users/36d6516a40f84dbbb84b1c6819e80270/Items/3419b6d0a34b7dd604f2e44ad086c9a9".
System.NullReferenceException: Object reference not set to an instance of an object.
   at Emby.Server.Implementations.Dto.DtoService.GetBaseItemDtoInternal(BaseItem item, DtoOptions options, User user, BaseItem owner)
   at Emby.Server.Implementations.Dto.DtoService.GetBaseItemDto(BaseItem item, DtoOptions options, User user, BaseItem owner)
   at Jellyfin.Api.Controllers.UserLibraryController.GetItem(Guid userId, Guid itemId)
   at lambda_method937(Closure , Object )
[...]

PR #7849, opened 2022-06-05
Fix null reference error when item is null

adrianocavalari:
**Changes**
Add null validator for items.
Add UserLibraryControllerTests
Replicate: Create a library(Keep empty), add a new genre, and click on the new genre

**Issues**
#6164


Emby.Server.Implementations/Dto/DtoService.cs
```diff
@@ -170,6 +170,11 @@ private static IList<BaseItem> GetTaggedItems(IItemByName byName, User user, Dto
 
         private BaseItemDto GetBaseItemDtoInternal(BaseItem item, DtoOptions options, User user = null, BaseItem owner = null)
         {
+            if (item == null)
```
Bond-009, line 173: Ideally, null shouldn't ever be passed to this function, but be caught beforehand

Bond-009 (changes requested):

Emby.Server.Implementations/Dto/DtoService.cs
adrianocavalari: Hi @Bond-009, that's true, I'll take a look at why this is being sent as null, but it would be good to have this validation to avoid the system crashing in any future changes that can make this to be null again(After fixed).

Bond-009: The tests should catch those regressions

Bond-009:
Superseded by #9253

PR #7868, opened 2022-06-10
chore: disable DLNA by default

cvium:
**Changes**
Disable DLNA by default. It does not support authentication, but it is LAN only, so the security implications aren't as big as UPnP, but it still seems unnecessary to enable by default.

Also added an ActionFilter to handle the "toggling" of the DLNA controller instead of the previous `if` mess.

### DIFF
--- a/Emby.Dlna/Configuration/DlnaOptions.cs
+++ b/Emby.Dlna/Configuration/DlnaOptions.cs
@@ -13,7 +13,7 @@ namespace Emby.Dlna.Configuration
         public DlnaOptions()
         {
             EnablePlayTo = true;
-            EnableServer = true;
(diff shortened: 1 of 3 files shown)


PrplHaz4:
Is it worth adding a note to the DLNA enable screen (or something) with this?  Also I don't think it's LAN-only, I believe the DLNA endpoints that expose the media can still be accessed outside of your LAN even though the media IDs are reasonably obfuscated.

> Enabling DLNA exposes certain media via an always-on unauthenticated endpoint (http://exampledlnaurl).  When enabled, imiting access to the exposed media can only be done by assigning a DLNA user with specific media permissions.

joshuaboniface (approved):

PR #7885, opened 2022-06-12
Prevent 400 error when using navigation buttons.

iwalton3:
**Changes**
 - Fix navigation buttons in web app. (API call in `jellyfin-web` currently returns a 400 error.)

This fix can be back-ported to the 10.8.x stable version and I have verified it works as such.

### DIFF
--- a/MediaBrowser.Model/Session/GeneralCommand.cs
+++ b/MediaBrowser.Model/Session/GeneralCommand.cs
@@ -14,9 +14,9 @@ public class GeneralCommand
     }
 
     [JsonConstructor]
-    public GeneralCommand(Dictionary<string, string> arguments)
+    public GeneralCommand(Dictionary<string, string>? arguments)
     {
-        Arguments = arguments;
+        Arguments = arguments ?? new Dictionary<string, string>();
     }
 
     public GeneralCommandType Name { get; set; }


nielsvanvelzen (approved):

MediaBrowser.Model/Session/GeneralCommand.cs
```diff
@@ -14,14 +14,16 @@ public GeneralCommand()
...
+            Arguments = arguments;
+        }
     }
 
     public GeneralCommandType Name { get; set; }
 
     public Guid ControllingUserId { get; set; }
 
-    public Dictionary<string, string> Arguments { get; }
+    public Dictionary<string, string> Arguments { get; } = new();
```
cvium, line 28: Better to do it in the constructor.

```suggestion
    public Dictionary<string, string> Arguments { get; }
```

MediaBrowser.Model/Session/GeneralCommand.cs
```diff
@@ -14,14 +14,16 @@ public GeneralCommand()
     }
 
     [JsonConstructor]
-    public GeneralCommand(Dictionary<string, string> arguments)
+    public GeneralCommand(Dictionary<string, string>? arguments = null)
     {
-        Arguments = arguments;
+        if (arguments != null) {
+            Arguments = arguments;
+        }
```
cvium, line 21:
```suggestion
        Arguments = arguments ?? new Dictionary<string, string>();
```

cvium (changes requested):

MediaBrowser.Model/Session/GeneralCommand.cs
```diff
@@ -14,14 +14,16 @@ public GeneralCommand()
     }
 
     [JsonConstructor]
-    public GeneralCommand(Dictionary<string, string> arguments)
+    public GeneralCommand(Dictionary<string, string>? arguments = null)
```
cvium, line 17: Is the default value necessary?

```suggestion
    public GeneralCommand(Dictionary<string, string>? arguments)
```

iwalton3: That’s a good question I’ll rebuild it and see

iwalton3: Yes this worked

iwalton3:
Looks like backport still failed with a rebase

crobibero (approved):

crobibero:
The backport command always fails, it's a relic of the way we used to do things

cvium (approved):

Issue #7883, opened 2022-06-11
[Issue]: searchHints API Endpoint sometimes return bad request when some characters is in the URL


tobycm:
### Please describe your bug

I am developing with Jellyfin API
Let's look at this endpoint: [/search/hints](http://swagger.emby.media/?staticview=true#/SearchService/getSearchHints)

I can authenticate with Jellyfin just fine, try searching some media and success doing so.

Demo:

### Jellyfin Version

10.7.7

### Environment

```markdown
- OS: Server: ARM Raspbian Linux, Client: Windows
- Virtualization: None
- Clients: Jellyfin Windows
- Browser: None
- FFmpeg Version: i don't know how to find this
- Playback Method: None
- Hardware Acceleration: None
- Plugins: AudioDB, AutoOrganize, MusicBrainz, OMDb, Playback Reporting, Reports, Webhook
- Reverse Proxy: nginx + Cloudflare, same thing happen on local too
- Base URL: /jellyfin
- Networking: Host
- Storage: local
```

### Jellyfin logs

```shell
[2022-06-11 15:21:50.518 -07:00] [ERR] Error processing request. URL "GET" "/Search/Hints".
System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
   at System.SZArrayHelper.get_Item[T](Int32 index)
   at Jellyfin.Api.Controllers.SearchController.GetSearchHintResult(SearchHintInfo hintInfo)
   at System.Linq.Enumerable.SelectListIterator`2.ToArray()
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at Jellyfin.Api.Controllers.SearchController.Get(Nullable`1 startIndex, Nullable`1 limit, Nullable`1 userId, String searchTerm, String[] includeItemTypes, String[] excludeItemTypes, String[] mediaTypes, Nullable`1 parentId, Nullable`1 isMovie, Nullable`1 isSeries, Nullable`1 isNews, Nullable`1 isKids, Nullable`1 isSports, Boolean includePeople, Boolean includeMedia, Boolean includeGenres, Boolean includeStudios, Boolean includeArtists)
   at lambda_method1203(Closure , Object , Object[] )
[...]

PR #7894, opened 2022-06-12
Switch to FirstOrDefault extension

crobibero:
Fixes https://github.com/jellyfin/jellyfin/issues/7883

Extension method was added because there is an analyzer to not use the Linq method

### DIFF
--- a/src/Jellyfin.Extensions/ReadOnlyListExtension.cs
+++ b/src/Jellyfin.Extensions/ReadOnlyListExtension.cs
@@ -57,5 +57,21 @@ namespace Jellyfin.Extensions
 
             return -1;
         }
+
+        /// <summary>
+        /// Get the first or default item from a list.
+        /// </summary>
+        /// <param name="source">The source list.</param>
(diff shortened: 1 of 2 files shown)


src/Jellyfin.Extensions/ReadOnlyListExtension.cs
```diff
@@ -57,5 +57,16 @@ public static int FindIndex<T>(this IReadOnlyList<T> source, Predicate<T> match)
...
+
+        /// <summary>
+        /// Get the first or default item from a list.
+        /// </summary>
+        /// <param name="source">The source list.</param>
+        /// <typeparam name="T">The type of item.</typeparam>
+        /// <returns>The first item or default if list is empty.</returns>
+        public static T? FirstOrDefault<T>(this IReadOnlyList<T> source)
+        {
+            return source.Count == 0 ? default : source[0];
```
cvium, line 69: But source can be null

cvium (approved):

Bond-009 (approved):

PR #7903, opened 2022-06-12
Make IgnoreDts configurable for M3U tuner

knackebrot:
**Changes**
- Added IgnoreDts to TunerHostInfo
- Added IgnoreDts checkbox to M3U tuner configuration page: https://github.com/jellyfin/jellyfin-web/pull/3696

**Issues**

Fixes https://github.com/jellyfin/jellyfin/issues/7267
Closes https://github.com/jellyfin/jellyfin/pull/7779 and https://github.com/jellyfin/jellyfin/pull/7776

### DIFF
--- a/MediaBrowser.Model/LiveTv/TunerHostInfo.cs
+++ b/MediaBrowser.Model/LiveTv/TunerHostInfo.cs
@@ -8,6 +8,7 @@ namespace MediaBrowser.Model.LiveTv
         public TunerHostInfo()
         {
             AllowHWTranscoding = true;
+            IgnoreDts = true;
         }
 
         public string Id { get; set; }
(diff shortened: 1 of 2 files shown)


MediaBrowser.Model/LiveTv/TunerHostInfo.cs
```diff
@@ -31,5 +31,7 @@ public TunerHostInfo()
         public int TunerCount { get; set; }
 
         public string UserAgent { get; set; }
+
+        public bool IgnoreDts { get; set; } = true;
```
crobibero, line 35: `AllowHWTranscoding` is set in the constructor, so might as well keep it consistent

knackebrot: Fixed

thornbill:
Given the nature of this change is to add a new feature (including what appears to be an API change) this might be better suited for merging to master?

crobibero (approved):

Bond-009 (approved):

rebel24:
I've updated to the latest build and don't see this option ? Is this in the latest unstable build, if not, can I set this parameter manually ?

PR #7940, opened 2022-06-15
Fix recommendations

Shadowghost:
Didn't find any issue describing this but there were reports on Matrix.

### DIFF
--- a/Emby.Server.Implementations/Data/SqliteItemRepository.cs
+++ b/Emby.Server.Implementations/Data/SqliteItemRepository.cs
@@ -2403,7 +2403,7 @@ namespace Emby.Server.Implementations.Data
                 }
 
                 // genres, tags, studios, person, year?
-                builder.Append("+ (Select count(1) * 10 from ItemValues where ItemId=Guid and CleanValue in (select CleanValue from itemvalues where ItemId=@SimilarItemId))");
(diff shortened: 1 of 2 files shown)


Bond-009 (approved):

1337joe (approved):

Emby.Server.Implementations/Data/SqliteItemRepository.cs
```diff
@@ -2403,7 +2403,7 @@ private void SetFinalColumnsToSelect(InternalItemsQuery query, List<string> colu
                 }
 
                 // genres, tags, studios, person, year?
-                builder.Append("+ (Select count(1) * 10 from ItemValues where ItemId=Guid and CleanValue in (select CleanValue from itemvalues where ItemId=@SimilarItemId))");
+                builder.Append("+ (Select count(1) * 10 from ItemValues where ItemId=Guid and CleanValue in (select CleanValue from ItemValues where ItemId=@SimilarItemId))");
```
cvium, line 2406: Unnecessary change, no? I'm pretty sure it's case insensitive

Shadowghost: Me too, but we did it case sensitive for all other tables except this one, so for consitencie's sake I changed it. I cna revert that if required.

PR #7941, opened 2022-06-15
Fix the int overflow issue and the wrong map index of external audio

nyanmisaka:
**Changes**
- Fix the int overflow issue in encoder bufsize,
cap the max target bitrate to int.MaxValue/2 = 1073741823, which is more than enough for any common videos.
- Fix the map index of external audio

**Issues**
Fixes https://github.com/jellyfin/jellyfin-plugin-tvheadend/issues/59
FIxes https://github.com/jellyfin/jellyfin/issues/7942

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -1893,7 +1893,7 @@ namespace MediaBrowser.Controller.MediaEncoding
             return request.EnableAutoStreamCopy;
         }
 
-        public int? GetVideoBitrateParamValue(BaseEncodingJobOptions request, MediaStream videoStream, string outputVideoCodec)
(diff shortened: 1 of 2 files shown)


MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -1925,7 +1925,7 @@ public bool CanStreamCopyAudio(EncodingJobInfo state, MediaStream audioStream, I
                 }
             }
 
-            return bitrate;
+            return Math.Min(bitrate ?? 0, int.MaxValue / 2);
```
Bond-009, line 1928: Shouldn't you still return null, or change the function signature?

nyanmisaka: Done.

cvium (approved):
Wouldn't hurt to add a code comment regarding Max / 2

articuno1au:
Thanks for the quick turn around on this one :clap:

Issue #7534, opened 2022-03-31
[Issue]: Next Up Rewatching - Last watched is a special, rewatching displays wrong episode.


SenorSmartyPants:
### Please describe your bug

I have several series where the last episode I watched was an end of season/series special. However, rewatching is displaying either that same last watched special as the next up, or displaying another special episode from after the same season.

I know that specials are processed a bit differently in the next up code. I must not have tested specials at all when I added rewatching.

### Jellyfin Version

10.8.0

### Environment

```markdown
- OS:
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Please attach any screenshots here

![image](https://user-images.githubusercontent.com/991618/161118358-e431a890-846b-4a91-bbb8-8fc97c9ab9a0.png)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

SenorSmartyPants:
Still an issue

SenorSmartyPants:
Still an issue, and there is a PR to address it.

SenorSmartyPants:
Don't close.

fuzxi:
Still an issue. (I hate stale bots)

SenorSmartyPants:
Still an issue

fuzxi:
Still fresh

felix920506:
I am not seeing this behavior in 10.9.9, instead the season I am rewatching simply doesn't show up in next up unless I manually unwatch the whole season

SenorSmartyPants:
Still a thing

SenorSmartyPants:
Still a thing

SenorSmartyPants:
Still a thing

SenorSmartyPants:
still a thing

stephenshutters:
I’m having this issue even with non-special episodes. A previously watched episode stays in Next Up marked as having been watched, and the actual next episode is listed beside it. And no amount of marking the episode as unwatched and then watched again actually removes it from Next Up.

![Example](https://github.com/user-attachments/assets/9f1229e4-7e61-4949-a16f-7c586f9ebe9f)

PR #7966, opened 2022-06-18
Check for last watched specials and next episode based on them

SenorSmartyPants:
**Changes**

Improved logic when last watched episode was a special when rewatching TV.

**Issues**
No issues entered but here are the problems solved.
- When last episode watched was a special, and no further episodes in series, special is always displayed in next up.
- Wrong next episode was displayed when last episode watched was a special


cvium:
Four extra queries per series is a little much

SenorSmartyPants:
> (Potentially) Four extra DB queries per series is a little much

Is there a way to return userdata with the watched episodes?

cvium:
Not right now. So I'm starting to think that maybe "rewatching" isn't ready for prime time and that we should revert it

SenorSmartyPants:
DTOOptions.EnableUserData is set to true on on the input parameter. But I don't see UserData returned. Why is that option there if it is not honored?

SenorSmartyPants:
It also appears that _userDataManager.GetUserData is caching those lookups. I'm not sure when that is initially looked up or how long it is cached.

SenorSmartyPants:
Addresses #7534

SenorSmartyPants:
Messed up PR with a rebase. Closing. New PR created.

PR #8001, opened 2022-06-24
Fix yuvj420p pixel format hardware decoding

### LINKED ISSUE #7823: [Synology DS220+] NAS stop responding during high bitrate 4k transcoding
Hi,

**Configuration:** Synology DS220+ (2GB RAM) last updates for DSM and packages, jellyfin 10.7.7 installed in Docker.

I recently purchased a GoPro Hero 10 Black and my settings for videos are : 4k HEVC, 50 fps, bitrate 100-120 Mbps (max settings).
These new videos are included in my jellyfin library.

On Chrome, in the jellyfin web client, i want to play these files (no bitrate limit in my profile, and playing from local network).
I can see in the jellyfin server logs that the video is transcoded from HEVC to h264, but some seconds after playing the video, the DS220+ becomes completly unresponding: no DSM interface available, no network ping possible so no SSH possible. Even, if i push the power button of the NAS to shut it down, it does not. I have to unplug it.
This is the same behavior for each GoPro 4k files like this.

The fact is that if i put a bitrate limit for streaming in my profile, for example 1080p 20 mbps, no problem during transcode.
If i change to 1080p 40 mbps, freeze.
I cannot log the memory of the NAS, i can see that it is growing but after it freezes so i cannot log it. And I don't see any logs in the NAS for that should contains the pattern "out of memory". So not sure about memory.
The configuration for transcoding should be correct because there is no problem for other files with lower bitrates, and during success transcoding, NAS CPU is not very high and memory does not explode as well.

When i play the same video from my android phone with the jellyfin app, transcode is made as well but seems locked to 8mbps (i don't know why because i disabled any external network limit), but it works at this bitrate.

My old action cam was 1080p 50fps with bitrate at 30mps and there was no problem.
I think of 4k/HEVC problem exclusively.

Any ideas or people experiencing transcode issues for 4k files ?
Attached, a screenshot of the media info of the file and the beginning of the log during transcoding.
image
[Log.txt](https://github.com/jellyfin/jellyfin/files/8799126/Log.txt)

Thanks.

nyanmisaka:
**Changes**
- Fix yuvj420p pixel format hardware decoding (yuvj420p is used on GoPro)

**Issues**
Fixes #7823

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -4288,6 +4288,7 @@ namespace MediaBrowser.Controller.MediaEncoding
                     return videoStream.BitDepth.Value;
                 }
                 else if (string.Equals(videoStream.PixelFormat, "yuv420p", StringComparison.OrdinalIgnoreCase)
+                         || string.Equals(videoStream.PixelFormat, "yuvj420p", StringComparison.OrdinalIgnoreCase)
                          || string.Equals(videoStream.PixelFormat, "yuv444p", StringComparison.OrdinalIgnoreCase))
                 {
                     return 8;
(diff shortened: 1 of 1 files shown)


crobibero (approved):

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -4578,7 +4579,8 @@ public string GetQsvHwVidDecoder(EncodingJobInfo state, EncodingOptions options,
             var hwSurface = (isIntelDx11OclSupported || isIntelVaapiOclSupported)
                 && _mediaEncoder.SupportsFilter("alphasrc");
 
-            var is8bitSwFormatsQsv = string.Equals("yuv420p", videoStream.PixelFormat, StringComparison.OrdinalIgnoreCase);
+            var is8bitSwFormatsQsv = string.Equals("yuv420p", videoStream.PixelFormat, StringComparison.OrdinalIgnoreCase)
+                || string.Equals("yuvj420p", videoStream.PixelFormat, StringComparison.OrdinalIgnoreCase);
```
cvium, line 4583: Please fix indentation for the next 4 additions too
```suggestion
                                     || string.Equals("yuvj420p", videoStream.PixelFormat, StringComparison.OrdinalIgnoreCase);
```

nyanmisaka: Done.

cvium (approved):

PR #8008, opened 2022-06-25
Add escape hatch for Series merging

revam:
**Changes**

This is an universal solution for plugins to override how series are merged.
The reason to override is so we can set the same provider id on multiple items without merging them, while using another id for merging them. Having an (optional) provider id not tied to any online database allows plugins to use their own rules for merging series.

**Issues**

Follow up PR for #7497

### DIFF
--- a/MediaBrowser.Model/Entities/MetadataProvider.cs
+++ b/MediaBrowser.Model/Entities/MetadataProvider.cs
@@ -8,6 +8,12 @@ namespace MediaBrowser.Model.Entities
     public enum MetadataProvider
     {
         /// <summary>
+        /// This metadata provider is for users and/or plugins to override the
+        /// default merging behaviour.
+        /// </summary>
+        Custom = 0,
+
+        /// <summary>
         /// The imdb.
         /// </summary>
         Imdb = 2,
--- a/MediaBrowser.Controller/Entities/TV/Series.cs
+++ b/MediaBrowser.Controller/Entities/TV/Series.cs
@@ -184,6 +184,11 @@ namespace MediaBrowser.Controller.Entities.TV
                 list.Insert(0, key);
             }
 
+            if (this.TryGetProviderId(MetadataProvider.Custom, out key))
+            {
+                list.Insert(0, key);
+            }
+
             return list;
         }
 


revam:
I don't know if it should show up in the UI or not, but I can add another commit to add the IExternalId needed if so desired.

revam:
Also, this doesn't change — or try to reuse — any existing logic, as opposed to what my previous attempt at solving this problem did (see the linked PR for details).

MediaBrowser.Model/Entities/MetadataProvider.cs
```diff
@@ -7,6 +7,12 @@ namespace MediaBrowser.Model.Entities
     /// </summary>
     public enum MetadataProvider
     {
+        /// <summary>
+        /// This provider is an escape hatch for the user to override the series
+        /// merging.
+        /// </summary>
+        User = 0,
```
cvium, line 14: Wouldn't it be better to call it `Custom`?

revam: That wpuød probably be better. Will redo it later today

cvium: And the xmldoc is also a bit too Series specific. It can technically be used by all entities.

revam: Updated the enum name and added a more generic description.

revam:
Is there anything more that should be done, or is there anything else holding back this PR from being merged? @cvium

revam:
Still waiting for a reply...

cvium:
We are not including further features in 10.8. You should rebase on master.

MediaBrowser.Model/Entities/MetadataProvider.cs
```diff
@@ -7,6 +7,12 @@ namespace MediaBrowser.Model.Entities
     /// </summary>
     public enum MetadataProvider
     {
+        /// <summary>
+        /// This metadata provider is for users and/or plugins to override the
+        /// default merging behaviour.
+        /// </summary>
+        Custom = 0,
```
cvium, line 14: Are we certain that 0 is unused? Could be 999 or something instead to avoid collisions

revam: I looked at the hit blame history, and only '1' was previously used.

revam: *git

revam: I can change it to either '20' (the next unused number) or '999' (your suggestion) if you don't want to use '0'

cvium: It's fine

cvium (approved):

Bond-009 (approved):

revam:
Rebased on latest master since the jobs failed (according to my mail notification)

Issue #8015, opened 2022-06-26
[Issue]: Wrong stream mapping in ffmpeg if video file contains program data (e.g. tv recordings) or contains index gaps


neseih:
### Please describe your bug

Hi,

I got the following streams in a video file (.ts file from tv recording), no stream #0:0, sometimes program data:
```
  Stream #0:1[0x13ed]: Video: h264 (High) ([27][0][0][0] / 0x001B), yuv420p(tv, bt709, progressive), 1280x720 [SAR 1:1 DAR 16:9], 50 fps, 50 tbr, 90k tbn
  Stream #0:2[0x13ee](deu): Audio: mp2 ([3][0][0][0] / 0x0003), 48000 Hz, stereo, fltp, 256 kb/s
  Stream #0:3[0x13ef](deu): Audio: mp2 ([3][0][0][0] / 0x0003), 48000 Hz, stereo, fltp, 256 kb/s (visual impaired) (descriptions)
  Stream #0:4[0x13f3](deu): Audio: mp2 ([3][0][0][0] / 0x0003), 48000 Hz, stereo, fltp, 192 kb/s (hearing impaired)
  Stream #0:5[0x13f0](deu): Subtitle: dvb_teletext ([6][0][0][0] / 0x0006), 492x250
  Stream #0:6[0x13f2](deu): Audio: ac3 ([6][0][0][0] / 0x0006), 48000 Hz, 5.1(side), fltp, 448 kb/s
  Stream #0:7[0x492]: Unknown: none ([5][0][0][0] / 0x0005)
  Stream #0:8[0x498]: Unknown: none ([12][0][0][0] / 0x000C)
...
```

When selecting an audio track from the video file, playback fails as the wrong mapping is created:
```
[...]

Issue #8016, opened 2022-06-26
No audio track encoded/streamed  by playing of recorded TVH streams (SD, HD) 


olili:
### Please describe your bug

all TVH records (profile passthrough, ie. all streams) are not playable - video is okay, but there is no audio stream included. 
Live TV is working okay. 
With V.10.7.7 I had no problem. 

Transcode log attached.

Any hints?

### Jellyfin Version

10.8.0

### Environment

```markdown
- OS: deb 11
- Virtualization: no, bare metal
- Clients: chrome
- Browser:Chrome
- FFmpeg Version: 4.4.1-Jellyfin
- Playback Method: all bandwiths 
- Hardware Acceleration: yes, vaapi
- Plugins: TVH
- Reverse Proxy: 
- Base URL:
- Networking:
- Storage:
```

### FFmpeg logs

```shell
/usr/lib/jellyfin-ffmpeg/ffmpeg -analyzeduration 200M -init_hw_device vaapi=va:,driver=iHD -filter_hw_device va -hwaccel vaapi -hwaccel_output_format vaapi -autorotate 0 -i "http://jellyfin:jellyfin@127.0.0.1:9981/dvrfile/1554984998" -map_metadata -1 -map_chapters -1 -threads 4 -map 0:1 -map 0:2 -map -0:s -codec:v:0 h264_vaapi -rc_mode VBR -b:v 292000 -maxrate 292000 -bufsize 584000 -profile:v:0 high -force_key_frames:0 "expr:gte(t,0+n_forced*3)" -vf "setparams=color_primaries=bt709:color_trc=bt709:colorspace=bt709,deinterlace_vaapi=rate=frame,scale_vaapi=w=426:h=340:format=nv12" -codec:a:0 aac -ac 2 -ab 128000 -copyts -avoid_negative_ts disabled -max_muxing_queue_size 2048 -f hls -max_delay 5000000 -hls_time 3 -hls_segment_type mpegts -start_number 0 -hls_segment_filename "/var/lib/jellyfin/transcodes/2fbeeb0c92548b35780b27a8f1eef5ad%d.ts" -hls_playlist_type vod -hls_list_size 0 -y "/var/lib/jellyfin/transcodes/2fbeeb0c92548b35780b27a8f1eef5ad.m3u8"


ffmpeg version 4.4.1-Jellyfin Copyright (c) 2000-2021 the FFmpeg developers
  built with gcc 8 (Debian 8.3.0-6)
[...]

PR #8017, opened 2022-06-26
Fix stream mapping if no external stream exists

nyanmisaka:
**Changes**
- Fix stream mapping if no external stream exists

**Issues**
Should fixes: #8015 #8016

> // Add external streams before adding the streams from the file to preserve stream IDs on remote videos

Remote video stream do not always start at 0:0, data stream start at 0:0 is possible.


nyanmisaka:
https://github.com/Shadowghost/jellyfin/commit/e8fbaabd7958e247b3805a755f69d61a513e8ee5 is better solution, close this for now.

PR #8019, opened 2022-06-26
Fix stream mapping in base video files

### LINKED ISSUE #8015: [Issue]: Wrong stream mapping in ffmpeg if video file contains program data (e.g. tv recordings) or contains index gaps

Shadowghost:
Assumptions:
* external files only have audio or sub streams, no weird other data streams or whatever
* the base file can have other data streams except audio and video 
----
Fixes #8015 #8016
Supersedes #8017


MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -5483,20 +5483,33 @@ public string GetProgressiveAudioFullCommandLine(EncodingJobInfo state, Encoding
...
             {
-                var currentMediaStream = mediaStreams[i];
-                if (currentMediaStream == streamToFind)
+                for (var i = externalStreams; i < streams; i++)
                 {
-                    return index;
+                    var currentMediaStream = mediaStreams[i];
+                    if (currentMediaStream == streamToFind)
+                    {
+                        return currentMediaStream.Index - externalStreams;
```
nyanmisaka, line 5496: If the user upgrades from 10.7.x to 10.8.x with this PR, and does not force a metadata refresh, the indices of external subtitle streams will follow the indices of main media streams.

`-map 0:-1 -map 0:0 -map -0:s`
In this case it will break the stream mapping.
video -> not found -> return -1;
audio -> found at 1, minus 1 -> return 0;

```
[matroska,webm @ 0x55ad455b4200] Invalid stream specifier: -1.
    Last message repeated 19 times
Stream map '0:-1' matches no streams.
To ignore this, add a trailing '?' to the map.
```

Shadowghost: You are correct but I don't know how we could recognize and handle this.

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -5483,20 +5483,33 @@ public string GetProgressiveAudioFullCommandLine(EncodingJobInfo state, Encoding
 
         public static int FindIndex(IReadOnlyList<MediaStream> mediaStreams, MediaStream streamToFind)
         {
-            var index = 0;
-            var length = mediaStreams.Count;
+            var streams = mediaStreams.Count;
+            var externalStreams = mediaStreams.Count(s => s.IsExternal);
```
nyanmisaka, line 5487:
```suggestion
            var externalStreams = mediaStreams.Count(s => s.IsExternal);

            // Fallback to the old behavior of external subtitle stream indices in 10.7 and older.
            if (externalStreams > 0 && !mediaStreams[0].IsExternal)
            {
                return streamToFind.Index;
            }            
```

This should help.

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
cvium: Database migration

Shadowghost: I'M not confident enough to do that tbh. Especially because this change depends not only on what the current index is

cvium: I don't really like handling backwards compatibility with code because it's something that will never be removed again :/

Shadowghost: We could mention the need of a metadata refresh in the release notes and either drop the workaround right now or keep it in place until 10.9 and remove it there.
I know all of this is suboptimal. :/
How would one approach an automated db migration in such a case?

neseih: > I don't really like handling backwards compatibility with code because it's something that will never be removed again :/

I think the same. The code is hard to understand (for me) and doesn‘t work in all cases. I would propose to just save the stream id from ffprobe as a clean solution. That would fix the bug for all later library reads.
Maybe it is possible to remove the stream mapping for all ts files and tv library and reparse these files.
Anyway, it would only improve the situation as these files are currently not working (even with this fix).
Nevertheless, thx @Shadowghost for trying hard to solve this!

Shadowghost: The reason why it's not like that already is the following:
When using files on a remote FS (like strm via Gdrive) they are not probed on scan because that takes really long. Instead they are probed when playback starts. If you have external audio/subs for that file and select them before starting playback they would be added after the probed streams from the file, geting new Indices and therefore breaking the selection you did when initiating the playback. That's why I changed this to first add the external streams and after that add the internal streams.
Ideally we'd have an independant ID in the database wich has nohing to do with the actual stream ID within the file.

neseih: Would it be possible to have external streams to have negative indexes? So you could know if its internal or external and if the internal streams of the file has already been probed. But that‘s a breaking change as well 😞

Shadowghost:
Not required anymore.

PR #8027, opened 2022-06-28
Remove r3 from clean strings

### LINKED ISSUE #6255: Titles from filenames are truncated upon any CleanStrings regex match
Filename `abc R2 def.mkv`  is given the title `abc R2 def`.

Filename `abc R3 def.mkv`  is given the title `abc`.

Why does the string "R3" cause the title to be truncated while the string "R2" does not?  

The library is of type "Movies" but contains personal media.  "Prefer embedded titles over filenames" is not checked.  All metadata sources are disabled.  

Update: 
`abc a3 def.mkv`  > `abc a3 def` 
`abc r3 def.mkv`  > `abc`

`r3` gets truncated but `a3` and `r2` do not.

Update 2:
Any string listed in the CleanStrings regex is removed and everything that follows the string is truncated from the title.  

https://github.com/jellyfin/jellyfin/blob/91d6ffd731e163bf281348872c2421598fa4edb2/Emby.Naming/Common/NamingOptions.cs#L140

Update 3:
The issue also affects "Other" media type libraries.  

Title string cleaning arguably applies to "Movies" except it would be nice to have an option for generating movie titles from filenames without regex cleaning or other interpretation.  

The CleanStrings regex is wholly specific to Movies.   Applying Cleanstrings to "Other" implies that "Other" does not include music, books, photos, etc.

Bond-009:
**Issues**
Fixes #6255

### DIFF
--- a/Emby.Naming/Common/NamingOptions.cs
+++ b/Emby.Naming/Common/NamingOptions.cs
@@ -153,7 +153,7 @@ namespace Emby.Naming.Common
 
             CleanStrings = new[]
             {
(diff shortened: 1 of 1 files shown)


Shadowghost:
Shouldn't this go against the 10.8 release branch since it's a bugfix?

cvium:
Why are we removing this? R5 is pretty old and unused nowadays, but it was a thing for a long while

Bond-009:
@cvium are you talking solely about r5 or also about r3?
I can change it to only remove r3 if that is what you'd prefer

cvium:
> @cvium are you talking solely about r5 or also about r3? I can change it to only remove r3 if that is what you'd prefer

I'm not sure if R3 was a thing, but R5 was. I don't know if it's worth keeping though.

cvium (approved):

crobibero (approved):

PR #8038, opened 2022-06-30
Remove mount and unmount permissions for jellyfin group from sudoers

### LINKED ISSUE #8037: [Issue]: Debian package allows privilege escalation from jellyfin user to root
### Please describe your bug

Jellyfin debian package includes [`/etc/sudoers.d/jellyfin-sudoers`](https://github.com/jellyfin/jellyfin/blob/2c9b24a1e76449697abdaae6c4ccdb515d2d1337/debian/conf/jellyfin-sudoers#L34) file which allows jellyfin user to execute `/bin/mount` without a password. This allows jellyfin user to perform privilege escalation to root.

Exploitation is rather trivial:
```
jellyfin@box:/tmp$ sudo mount -o bind /bin/bash /bin/mount
jellyfin@box:/tmp$ sudo mount
mount: cannot set terminal process group (1939091): Inappropriate ioctl for device
mount: no job control in this shell
box:/tmp# id
uid=0(root) gid=0(root) groups=0(root)
box:/tmp#
```
For this issue to be actual security issue, the attacker would first need to gain ability to execute commands (or code) as `jellyfin` user. This could also be a problem for services that provide jellyfin as a service that low privilege users are allowed to access.

Assuming mounting ISO images is a necessary feature  there are number of ways to fix this, one of which would be to employ FUSE and fuseiso9660. Another one would be to add a wrapper application for mounting that would restrict the parameters that can be given to mount, and then give this command sudo rights. Note that the wrapper should enforce 'nosuid', 'nodev' and 'noexec' mount options, and restrict where the mountpoints are located. Also, it would be prudent to restrict umount to only allow unmounting these specific mountpoints ncreated by the mount wrapper.

Meanwhile this issue can be mitigated manually by commenting out the following lines in the `/etc/sudoers.d/jellyfin-sudoers` file:
```
jellyfin ALL=(ALL) NOPASSWD: /bin/mount
jellyfin ALL=(ALL) NOPASSWD: /bin/umount
```

This issue was reported to `security@jellyfin.org` in December 2021 but it appears this resulted in no actions being taken.

### Jellyfin Version

Other

### if other:

at least 10.7.7 and git HEAD

### Environment

```markdown
- OS: Debian
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

Shadowghost:
Fixes #8037

We no longer mount ISOs nor do we support plugins mounting them. All playback and probing related handling is done by ffmpeg or other tools directly accessing the ISO without the OS being involved.

### DIFF
--- a/debian/conf/jellyfin-sudoers
+++ b/debian/conf/jellyfin-sudoers
@@ -30,8 +30,4 @@ Defaults!RESTARTSERVER_INITD !requiretty
 Defaults!STARTSERVER_INITD !requiretty
 Defaults!STOPSERVER_INITD !requiretty
 
-#Allow the server to mount iso images
-jellyfin ALL=(ALL) NOPASSWD: /bin/mount
-jellyfin ALL=(ALL) NOPASSWD: /bin/umount
-
 Defaults:jellyfin !requiretty
--- a/fedora/README.md
+++ b/fedora/README.md
@@ -18,14 +18,6 @@ $ sudo dnf install https://download1.rpmfusion.org/free/fedora/rpmfusion-free-re
 $ sudo yum localinstall --nogpgcheck https://download1.rpmfusion.org/free/el/rpmfusion-free-release-7.noarch.rpm
 ```
 
-## ISO mounting
-
-To allow Jellyfin to mount/umount ISO files uncomment these two lines in `/etc/sudoers.d/jellyfin-sudoers`
-```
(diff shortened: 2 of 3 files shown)


joshuaboniface (approved):
I thought for sure we removed this ages ago...

cvium (approved):

PR #8039, opened 2022-06-30
Implement a Modular Home screen and User facing plugin pages

IAmParadox27:
**Changes**

- Add support to have a modular (extensible) home screen that can be modified by the user and added to by plugins.
- Add support for plugins to serve pages that any user (not just admins) can see.
- The support is hidden behind a user preference and the old home screen layout is left untouched.

**Known Issues**
- Currently only 5 of the home screen sections have been ported into the new system: My Media, Continue Watching, Next Up, Latest Movies and Latest Shows. The other types will need to be added in due course.
- Settings are limited and currently use a json file for storing the settings rather than the database.

**Linked PRs**
https://github.com/jellyfin/jellyfin-web/pull/3740
https://github.com/jellyfin/jellyfin-apiclient-javascript/pull/422

### DIFF
--- a/Emby.Server.Implementations/ApplicationHost.cs
+++ b/Emby.Server.Implementations/ApplicationHost.cs
@@ -30,6 +30,7 @@ using Emby.Server.Implementations.Cryptography;
 using Emby.Server.Implementations.Data;
 using Emby.Server.Implementations.Devices;
 using Emby.Server.Implementations.Dto;
+using Emby.Server.Implementations.HomeScreen;
 using Emby.Server.Implementations.HttpServer.Security;
 using Emby.Server.Implementations.IO;
 using Emby.Server.Implementations.Library;
@@ -599,6 +600,10 @@ namespace Emby.Server.Implementations
             serviceCollection.AddSingleton<ISubtitleManager, SubtitleManager>();
 
             serviceCollection.AddSingleton<IProviderManager, ProviderManager>();
+
+            serviceCollection.AddSingleton<IHomeScreenManager, HomeScreenManager>();
+
+            serviceCollection.AddSingleton<IPluginPagesManager, PluginPagesManager>();
 
             // TODO: Refactor to eliminate the circular dependency here so that Lazy<T> isn't required
             serviceCollection.AddTransient(provider => new Lazy<ILiveTvManager>(provider.GetRequiredService<ILiveTvManager>));
--- a/Emby.Server.Implementations/HomeScreen/HomeScreenManager.cs
+++ b/Emby.Server.Implementations/HomeScreen/HomeScreenManager.cs
@@ -23,6 +23,8 @@ namespace Emby.Server.Implementations.HomeScreen
         private readonly IServiceProvider _serviceProvider;
         private readonly IApplicationPaths _applicationPaths;
 
+        private const string SettingsFile = "ModularHomeSettings.json";
+
         public HomeScreenManager(IServiceProvider serviceProvider, IApplicationPaths applicationPaths)
         {
             _serviceProvider = serviceProvider;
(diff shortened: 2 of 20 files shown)


thornbill:
These two features should definitely be separate pull requests:

- Add support to have a modular (extensible) home screen that can be modified by the user and added to by plugins.
- Add support for plugins to serve pages that any user (not just admins) can see.

It just makes everything easier if PRs are limited to one per feature. :slightly_smiling_face: 

I haven't reviewed how this all works yet, but I'm not sure what is really gained by moving control of the home screen to the server. Users are already able to manage what sections appear on their home screen and the order of the sections. It seems a lot less flexible having that aspect of the interface controlled by the server instead of letting the clients manage that.

On the web side of things we recently [merged support for react-router](https://github.com/jellyfin/jellyfin-web/pull/3706), so I doubt we would want to add new routes using the old method unless it was absolutely necessary for some reason.

IAmParadox27:
> I haven't reviewed how this all works yet, but I'm not sure what is really gained by moving control of the home screen to the server. Users are already able to manage what sections appear on their home screen and the order of the sections. It seems a lot less flexible having that aspect of the interface controlled by the server instead of letting the clients manage that.

The control is that custom sections can be added by plugins. My plan is to add several sections that the users can turn on if they want "Watch it again", "Because you watched...", "Next in Collection".

Clients still manage what sections they want to see, but the server controls which sections are available. Its plenty more extensible doing it this way.

> These two features should definitely be separate pull requests:

While I agree, the Modular Home feature requires the plugin pages and that feature was developed because of the modular home need, I just wrote it in a way that other plugins could use it if they wanted.

thornbill:
Sorry for the delay in getting a response here, these are kind of a substantial changes that will really impact future development... and things tend to go a bit slower in the summer anyway. :wink: 

I don't think there is really any objection to the general ideas you have worked on here as far as supporting additional options for home screens and non-admin pages for plugins. I think ideally though it would have been best to have started with a proposal or discussion about the implementation... We definitely don't want anyone to feel as though there time has been wasted if the implementation does not line up with what the team wants to support long term.

Personally I'm still not convinced that moving control of the home page to the server is a good thing. I think we should be more flexible and supportive of clients that want to try different things (see Swiftfin or jellyfin-vue) instead of locking them to a specification set in the server. Similarly I don't think I've seen an example of something that could be added to the home screen via a plugin that wouldn't be better to just have support for directly in the core offering. :man_shrugging: imho the main blocker there is that it is currently fairly difficult to add new sections in web not really anything to do with the server or how the functionality exists today. I have wanted to add support for "new releases" since we forked from emby, but just haven't bothered to do so because of that reason...

I do definitely see a need for non-admin pages from plugins though. Although it should definitely be a separate PR (or set of PRs). I understand that the home screen changes were dependent on it, but it's still much easier to handle single PRs that add a single feature. There has been some talk about having plugin configuration not be tightly coupled to jellyfin-web so that other clients could also be able to manage and configure plugins, so that maybe another point to consider in how we want plugins to work in the future. I think maybe that could be a later discussion though.

IAmParadox27:
Guess Jellyfin isn't truly open source then.

cvium:
> It's the servers job to provide the client with an easy route to display not for it to really do any functional that isn't essential for it to do. If something can be centralised (server) it should.

I agree... somewhat. I don't really understand why the server should be responsible for home screen management though. I'm all for offering up some API endpoints that provide the necessary data for clients to build the screens, but it should be mostly agnostic.

I would also be fine with generic user settings being stored in the server eg. sort order, home screen layout etc.

IAmParadox27:
That is effectively what this change is. The user still says what sections they're interested in, the client requests which sections the user has selected (much like it asks for the section order) and then builds the layout and requests the data for each one. That's literally what this change does.

IAmParadox27:
> I don't really understand why the server should be responsible for home screen management though.

The additional support this change allows is for another developer to come along and go "hey I want to add a 'because you watched' section to the Home Screen" (like Netflix). Currently they'd have to implement the endpoint to provide the data, add the functionality to the client to support the section. Whereas what this change allows is for the dev to make a plugin, add the section type into the HomeScreenManager and implement a function to return the data for what movies/tv shows they're recommending, and that data is in the same format as all the other data receivers that currently exist for the current sections

thornbill:
> I would also be fine with generic user settings being stored in the server eg. sort order, home screen layout etc.

afaik this describes exactly how home screens work today. 

> The additional support this change allows is for another developer to come along and go "hey I want to add a 'because you watched' section to the Home Screen" (like Netflix).

I still don't see why we wouldn't just want this as a core feature... and similar data is available via the recommendations endpoints already.

PR #8050, opened 2022-07-03
Fix rendering library thumbnail if it has an Arabic name

hadicharara:
Library thumbnail rendering were broken with Arabic names, due to a bug in the DrawText function of the canvas.
Example for the word أفلام:
لقطة شاشة 2022-07-02 في 20 55 50

**Changes**
I used CanvasExtensions.DrawShapedText to properly render Arabic fonts. However, this function does not consider TextAlign (and calculating text width for Arabic fonts is complicated). As such, I am only using it for Arabic library names, and have set a fixed position from the left for the library name.

لقطة شاشة 2022-07-02 في 20 57 53

Although not ideal, it is better than before. I am open to any feedback!


Bond-009:
> due to a bug in the DrawText function of the canvas.

Is this a Skia issue?

hadicharara:
> > due to a bug in the DrawText function of the canvas.
> 
> Is this a Skia issue?

Yes exactly! I had added a library that fixed this problem. Not sure if I should rebase though because of the lack of interest in this MR.

crobibero:
Please rebase (and update added deps)
Unsure if something like this will work, but seems better to me then using Regex.

https://stackoverflow.com/a/4331154

hadicharara:
> Unsure if something like this will work, but seems better to me then using Regex.

Considering this won't run often, I'm not sure if it's worth the performance improvement. However, if you insist, I'll look into it. Also, I just rebased and updated the dependencies.

crobibero:
> > Unsure if something like this will work, but seems better to me then using Regex.
> 
> Considering this won't run often, I'm not sure if it's worth the performance improvement. However, if you insist, I'll look into it. Also, I just rebased and updated the dependencies.

If my link works it should fix the image for all RTL languages, not just Arabic; so I think it should at least be tested

hadicharara:
Alright I'll test it and report back.

hadicharara:
I did it. It should work with all RTL languages. Result is still sub-par but better than before.

cvium:
I don't think this PR is entirely correct. Some of the values are outside char range. You need to convert them to utf32 like the SO post says

hadicharara:
Converted characters to UTF-32 now.

Khaled0Ebrahim:
I geuss it will work
just needs to be centered vertically and horizontally

Khaled0Ebrahim:
https://github.com/mono/SkiaSharp/pull/1910#issue-1090076882

Khaled0Ebrahim:
I don't know much about c# but if this worked it is much more cleaner

```
using System.Globalization;

namespace Jellyfin.Drawing.Skia
{
    /// <summary>
    /// Used to provide utility functions to the project.
    /// </summary>
    public static class Utils
    {
        /// <summary>
        /// Checks if the string contains a right-to-left character.
        /// </summary>
        /// <param name="text">The string to check for right-to-left characters.</param>
        public static bool HasRtlCharacters(string? text)
        {
            if (text == null)
            {
                return false;
            }

            foreach (char c in text)
            {
                if (CharUnicodeInfo.GetUnicodeCategory(c) == UnicodeCategory.RightToLeft ||
                    CharUnicodeInfo.GetUnicodeCategory(c) == UnicodeCategory.RightToLeftEmbedding)
                {
                    return true;
                }
            }

            return false;
        }
    }
}

```

Khaled0Ebrahim:
I am not familiar with C# but I assume this is a cleaner version of your check
```
namespace Jellyfin.Drawing.Skia
{
    using System.Globalization;

    /// <summary>
    /// Provides utility functions for the project.
    /// </summary>
    public static class Utils
    {
        /// <summary>
        /// Checks if the string contains any right-to-left characters.
        /// </summary>
        /// <param name="text">The string to check.</param>
        /// <returns>True if the string contains any right-to-left characters, false otherwise.</returns>
        public static bool HasRTLCharacters(string? text)
        {
            if (string.IsNullOrEmpty(text))
            {
                return false;
            }

            TextInfo textInfo = CultureInfo.CurrentCulture.TextInfo;
            return textInfo.IsRightToLeft(text);
        }
    }
}

```

PR #8080, opened 2022-07-08
set collection ProductionYear metadata to newest Movie instead of oldest

### LINKED ISSUE #8079: [Issue]: Collections ProductionYear based on oldest movie in collection instead of newest
### Please describe your bug

The metadata for ProductionYear of Collections is set based on the oldest movie in the collection instead of the newest.
This causes issues such as sorting Movies based on date with collections enabled.  The collection ends up being buried at an older place in the order rather than the newer location. 

### Jellyfin Version

10.8.0

### if other:

_No response_

### Environment

```markdown
- OS:
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

sbranden:
Set collection ProductionYear to newest Movie instead of oldest so
that collections show up in newer location during sorting.

Fixes #8079

Signed-off-by: Scott Branden <sbranden@gmail.com>

**Changes**

**Issues**

### DIFF
--- a/MediaBrowser.Providers/Manager/MetadataService.cs
+++ b/MediaBrowser.Providers/Manager/MetadataService.cs
@@ -424,7 +424,7 @@ namespace MediaBrowser.Providers.Manager
                 return updateType;
             }
 
-            var date = children.Select(i => i.PremiereDate ?? DateTime.MaxValue).Min();
+            var date = children.Select(i => i.PremiereDate ?? DateTime.MinValue).Max();
 
             var originalPremiereDate = item.PremiereDate;
             var originalProductionYear = item.ProductionYear;
@@ -436,7 +436,7 @@ namespace MediaBrowser.Providers.Manager
             }
             else
             {
-                var year = children.Select(i => i.ProductionYear ?? 0).Min();
+                var year = children.Select(i => i.ProductionYear ?? 0).Max();
 
                 if (year > 0)
                 {


cvium:
I don't disagree with the sorting, but changing the premiere date seems wrong to me.

sbranden:
> 

@cvium 
Changing the premiere date makes sense for the collection as the collection is effectively a boxset.  The latest boxset for a collection does not exist until the latest movie in the collection premieres.

sbranden:
@cvium, please see my response.  I think this is the correct fix.

sbranden (commented):
csharp analyze failure unrelated to this commit

cvium:
If anything, it should be a configurable option. I haven't seen anyone else request this behaviour.

sbranden:
> If anything, it should be a configurable option. I haven't seen anyone else request this behaviour.

@cvium, it's a bug.  I guess nobody else has found the pattern as to what is wrong.  But, if you need a configuration option for this could you please let me know how/where to add it?  I am not familiar with this code to easily figure such out.

cvium:
I apologize for the delay. No one in the team seems entirely on board with this change.

sbranden:
Change makes sense to me.  A boxset release date can't exist until newest movie exists.

JPVenson:
Closed as stale

sbranden:
> Closed as stale

Issue still exists and fix still valid.

nielsvanvelzen:
In BaseItemDto logic we should use:
- PremiereDate for the lowest date (e.g. birthdate for a person, premiere for a movie, airing start for a series etc.)
- EndDate for the highest date (e.g. date of death for a person, premiere for a movie, airing end for a series etc.)

So for a collection that would mean the item in the collection with the lowest premiereDate will be the premiereDate of the collection.

sbranden:
> In BaseItemDto logic we should use:
> 
> * PremiereDate for the lowest date (e.g. birthdate for a person, premiere for a movie, airing start for a series etc.)
> * EndDate for the highest date (e.g. date of death for a person, premiere for a movie, airing end for a series etc.)
> 
> So for a collection that would mean the item in the collection with the lowest premiereDate will be the premiereDate of the collection.

A premiereDate of a collection is like a boxset.  The boxset date is the date of the last move, not the first.  Having it work as you propose makes sorting collections rather useless.  There is no indication when new movies are added to the collections.  That is the problem.

nielsvanvelzen:
In Jellyfin, a collection is not a box set but a box set is a collection. If you want to sort a collection based on the premiere date of the newest item inside of it we need I believe there are other ways to achieve that.

Bond-009:
See comment from @nielsvanvelzen above

sbranden:
With such a definition for Collection the date field is very useless.  You have no way of finding the collections with the newest items in them (as a collection "date" is based on the oldest).

PR #8086, opened 2022-07-08
Add SAMI subtitle parser

Shadowghost:
**Changes**
* Adds a parser for SAMI subtitles so they can be converted

**Currently untested since I don't have any SAMI subtitles.**


Bond-009:
Replaced by #8087

PR #8087, opened 2022-07-08
feat: make subtitleeditparser generic

cvium:
**Changes**
Instead of hard coding and wrapping every `SubtitleFormat` sub class, this introduces a generic parser that tries all available `SubtitleFormat` based on extension.

**Issues**
Supercedes #8086


crobibero (approved):

MediaBrowser.MediaEncoding/Subtitles/SubtitleEditParser.cs
```diff
@@ -57,5 +85,94 @@ public SubtitleTrackInfo Parse(Stream stream, CancellationToken cancellationToke
...
+            var assembly = Assembly.GetAssembly(typeof(SubtitleFormat));
+            if (assembly == null)
+            {
+                _logger.LogError("Missing assembly containing {SubtitleFormatName}", nameof(SubtitleFormat));
+                return GetFallbackSubtitleFormats();
+            }
+
+            foreach (var type in assembly.GetTypes())
+            {
+                if (!type.IsSubclassOf(typeof(SubtitleFormat)))
```
Bond-009, line 105: Should we check to make sure it isn't abstract?
```suggestion
                if (!type.IsSubclassOf(typeof(SubtitleFormat)) || type.IsAbstract)
```

MediaBrowser.MediaEncoding/Subtitles/SubtitleEditParser.cs
```diff
@@ -57,5 +85,94 @@ public SubtitleTrackInfo Parse(Stream stream, CancellationToken cancellationToke
...
         }
+
+        /// <inheritdoc />
+        public bool SupportsFileExtension(string fileExtension)
+            => _subtitleFormats.ContainsKey(fileExtension);
+
+        private IEnumerable<SubtitleFormat> GetSubtitleFormats()
+        {
+            var subtitleFormats = new List<SubtitleFormat>();
+            var assembly = Assembly.GetAssembly(typeof(SubtitleFormat));
```
Bond-009, line 96:
```suggestion
            var assembly = typeof(SubtitleFormat).Assembly;
```
https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.getassembly?view=net-7.0#remarks

MediaBrowser.MediaEncoding/Subtitles/SubtitleEditParser.cs
```diff
@@ -57,5 +85,94 @@ public SubtitleTrackInfo Parse(Stream stream, CancellationToken cancellationToke
...
+            if (assembly == null)
```
Bond-009, line 97: Is this case ever hit?

Bond-009 (changes requested):

MediaBrowser.MediaEncoding/Subtitles/SubtitleEditParser.cs
cvium: I doubt it, but it's nullable 🤷🏼

Bond-009: Not anymore ;)

cvium: You can remove it then

MediaBrowser.MediaEncoding/Subtitles/SubtitleEditParser.cs
```diff
@@ -57,5 +85,89 @@ public SubtitleTrackInfo Parse(Stream stream, CancellationToken cancellationToke
...
+                new Csv3(),
+                new Csv4(),
+                new Csv5(),
+                new Ebu(),
+                new NetflixImsc11Japanese(),
+                new NetflixTimedText(),
+                new QuickTimeText(),
+                new RealTime(),
+                new SmpteTt2052()
+            };
```
Bond-009, line 171:
```suggestion
```
Isn't this dead code now?

Bond-009 (changes requested):

Bond-009 (approved):

PR #8104, opened 2022-07-11
Add resolution text output for more resolutions

### LINKED ISSUE #8099: [Issue]: 848x480 16:9 video reported as 540p
### Please describe your bug

848x480 16:9 video reported as 540p.

This should be reported as 480p.

### Jellyfin Version

10.8.0

### if other:

_No response_

### Environment

```markdown
- OS:
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

image

### Code of Conduct

Shadowghost:
The text response was somehow lacking, so I added some more distinct values and changed some. Also added tests.

**Changes**
* Add text output for resolutions:
  * 144p
  * 240p
  * 360p
* Add 16:9 resolution handling to 480p and 576p
* properly handle ultra wide resolutions for 1080p

**Issues**
Fixes #8099


SenorSmartyPants:
Thanks for addressing this so quickly.

crobibero (approved):

SenorSmartyPants:
I noticed 512x384 4:3 is reported as 480. It's closer to 360, so maybe it could be rounded down instead of up?

Shadowghost:
IMHO we should not try to do this but just stick to the maximum official widths and heights.

SenorSmartyPants:
I can understand the desire to stick with official numbers. But with some older videos and less common resolutions it gives an inaccurate description of the resolution. Most of my SD and low def. examples here have been from Classic Doctor Who episodes. I found another weird one. 514x362, reported as 480p.

image

Maybe the upper bound could be a little fuzzy? **25% of the difference to the next level over** would still be considered the lower resolution instead of the upper? I know this could get into a morass of where exactly to put the boundary for a resolution, I'm just hoping to get a more accurate representation of some of these less common in-between resolutions.

```c#
                // 256x144 (16:9 square pixel format)
                <= 256 when Height <= 168 => IsInterlaced ? "144i" : "144p",
                // 426x240 (16:9 square pixel format)
                <= 426 when Height <= 270 => IsInterlaced ? "240i" : "240p",
                // 640x360 (16:9 square pixel format)
                <= 640 when Height <= 390 => IsInterlaced ? "360i" : "360p",
                // 854x480 (16:9 square pixel format)
                <= 854 when Height <= 495 => IsInterlaced ? "480i" : "480p",
                // 960x544 (16:9 square pixel format)
                <= 960 when Height <= 549 => IsInterlaced ? "540i" : "540p",
```

Bond-009 (approved):

PR #8112, opened 2022-07-12
Add musl-linux-arm64 support

JinYi-Tsinghua:
This passed a build test. It will add support for musl-linux-arm64. The files are appropriately named and the contents are consistent with the existing ARM64 and musl build files.

This means support for postmarketOS, and aarch64/ARM64 versions of Void Linux and Alpine Linux.


joshuaboniface (approved):
Seems  sensible to me. It would have to be added in a couple other places too, e.g. into `build.yaml` and the pipelines Matrix, then I'd have to tweak the build scripts. But I'm good with that.

The idea is that this complements the existing arm64 Docker build right?

joshuaboniface:
Here's where it would go: https://github.com/jellyfin/jellyfin/blob/master/.ci/azure-pipelines-package.yml#L117 into that matrix as `arm64-musl`

joshuaboniface:
Or, did I complete misunderstand and this is for a generic `tar.gz` package?

JinYi-Tsinghua:
> Or, did I complete misunderstand and this is for a generic `tar.gz` package? If so just follow the naming convention of the `amd64-musl` one so it's consistent, and then update the matrix at the _top_ of that `azure-pipelines-package.yml` config.

musl-linux-arm64 is the correct name. If you modify the naming convention from what I have used, then the build.sh will fail because it depends on the output of 'dpkg --print-architecture'. I confirmed that the output of 'dpkg --print-architecture' is musl-linux-arm64 on both postmarketOS and Void Linux (the two I tested).

As for `azure-pipelines-package.yml`, yes, the top needs to be updated to include:
`      Linux.musl-linux-arm64:
        BuildConfiguration: linux.musl-linux-arm64`

joshuaboniface (approved):

crobibero (approved):

nielsvanvelzen:
I think this PR is causing master builds to fail

Issue #8141, opened 2022-07-19
[Issue]: SqliteItemRepository throws exception when trying to retrieve People information


joseph39:
### Please describe your bug

Due to how SqliteItemRepository attempts to [bind](https://github.com/jellyfin/jellyfin/blob/master/Emby.Server.Implementations/Data/SqliteItemRepository.cs#L4991-L4994) @UserId even when there may be no such parameter in the query, it can lead to an exception being thrown.

### Jellyfin Version

10.8.0

### Environment

```markdown
- OS:
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #8142, opened 2022-07-19
Bind @UserId only when it appears in the statement

joseph39:
Bind @UserId only when it appears in the statement

**Changes**
This commit fixes a case where SqliteItemRepository attempts to bind @UserId even when such parameter may not appear in the query.

**Issues**
https://github.com/jellyfin/jellyfin/issues/8141


cvium:
Can you re-target the `release-10.8.z` branch?

joseph39:
https://github.com/jellyfin/jellyfin/pull/8166

Issue #8145, opened 2022-07-19
Add partial content support to Static Remote Stream


joseph39:
Add partial content support to Static Remote Stream by allowing Range-related HTTP headers to be passed through.

The goal of this change ultimately is to re-enable DirectStream for HTTP-backed media items where clients can buffer the data as little as necessary.

joseph39:
The goal of this change ultimately is to re-enable DirectStream for HTTP-backed media items where clients can buffer the data as little as necessary.

PR #8146, opened 2022-07-19
Re-enable Remote Direct Stream

joseph39:
Add partial content support to Static Remote Stream

**Changes**
This change adds partial content support to Static Remote Stream by allowing Range-related HTTP headers to be passed through.

**Issues**
https://github.com/jellyfin/jellyfin/issues/8145

### DIFF
--- a/Jellyfin.Api/Helpers/FileStreamResponseHelpers.cs
+++ b/Jellyfin.Api/Helpers/FileStreamResponseHelpers.cs
@@ -84,16 +84,16 @@ namespace Jellyfin.Api.Helpers
 
             foreach (var headerToCopy in HeadersToCopy)
             {
-                if (response.Content.Headers.Contains(headerToCopy))
+                if (response.Content.Headers.TryGetValues(headerToCopy, out var headerValue))
                 {
-                    httpContext.Response.Headers.Add(headerToCopy, new StringValues(response.Content.Headers.GetValues(headerToCopy).ToArray()));
+                    httpContext.Response.Headers.Add(headerToCopy, headerValue.ToArray());
                 }
             }
 
             if (".m3u8".Equals(new FileInfo(mediaPath.AbsolutePath).Extension, StringComparison.OrdinalIgnoreCase))
             {
                 var playlistContent = await response.Content.ReadAsStringAsync(cancellationToken).ConfigureAwait(false);
-                return await ProcessPlaylist(state.Request.Id.ToString(), state.MediaSource.Id, accessToken, playlistContent);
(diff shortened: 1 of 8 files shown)


joseph39:
The goal ultimately is to re-enable DirectStream for HTTP-backed media items where clients can buffer the data as little as necessary.

Jellyfin.Api/Helpers/FileStreamResponseHelpers.cs
```diff
@@ -32,17 +35,31 @@ public static class FileStreamResponseHelpers
...
+                request.Headers.Add(HeaderNames.Range, values.ToArray());
             }
 
             // Can't dispose the response as it's required up the call chain.
-            var response = await httpClient.GetAsync(new Uri(state.MediaPath), cancellationToken).ConfigureAwait(false);
+            var response = await httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken).ConfigureAwait(false);
             var contentType = response.Content.Headers.ContentType?.ToString() ?? MediaTypeNames.Text.Plain;
 
-            httpContext.Response.Headers[HeaderNames.AcceptRanges] = "none";
+            var headersToCopy = new List<string>() { HeaderNames.ContentLength, HeaderNames.ContentRange, HeaderNames.ContentDisposition, HeaderNames.ContentEncoding };
```
cvium, line 53: This should either be a private static field or written out explicitly like below. You're unnecessarily creating a new list every time this method is called.

```
            var responseHeaders = response.Content.Headers;
            var headers = httpContext.Response.Headers;
            if (responseHeaders.ContentLength != null)
            {
                headers.ContentLength = responseHeaders.ContentLength;
            }

            if (responseHeaders.ContentEncoding.Count > 0)
            {
                headers.ContentEncoding = responseHeaders.ContentEncoding.ToArray();
            }

            if (responseHeaders.TryGetValues(HeaderNames.ContentRange, out var contentRange))
            {
                headers.ContentRange = contentRange.ToArray();
            }

            if (responseHeaders.TryGetValues(HeaderNames.ContentDisposition, out var contentDisposition))
            {
                headers.ContentDisposition = contentDisposition.ToArray();
            }

joseph39: Thanks for taking a look at the PR -- I've addressed your feedback.

I've also since made further progress and added support to re-write segment URIs within M3U playlist files so that clients can stream remote content directly (without needing to transcode). I think this should be sufficient to re-enable remote direct streams, and with local testing, it has shown to have made a day-and-night difference w.r.t. its performance & responsiveness.

Issue #8164, opened 2022-07-23
[Issue]: Transcoding is forced on video files with no default audio stream set. Reason - SecondaryAudioNotSupported 


gaudat:
### Steps to reproduce:

* Create an AVI file with the following FFMPEG command.
```sh
ffmpeg -re -f lavfi -i "smptehdbars=rate=30:size=1920x1080" -f lavfi -i "sine=frequency=1000:sample_rate=48000" -to 10 -c:v h264 -profile:v baseline -pix_fmt yuv420p -preset ultrafast -tune zerolatency -crf 28 -c:a mp3 testvid.avi
```
* Add the created "testvid.avi" to a Jellyfin library.
* Try to play the file on web UI.
* Observe transcoding is always on and there is no direct playback option.
* Transcode reason is listed as SecondaryAudioNotSupported even there is only one audio stream in the file.

### Expected behavior:
File is played directly with no transcoding as H264 and MP3 are supported by the browser.

### Root cause:
JSON output of ffprobe has `streams[*].disposition.default` == 0 for all streams.

### Proposed solution:
Add check and modification of parse result in `ProbeResultNormalizer`.

### Jellyfin Version

Other

### if other:

10.8.1

### Environment

```markdown
- OS: Raspbian 2022-01-28 on Raspberry 3B+
- Virtualization: None
- Clients: Stock web client
- Browser: Latest Edge/FF/Chrome desktop/mobile
- FFmpeg Version: 5.0.1-Jellyfin (/usr/lib/jellyfin-ffmpeg/ffmpeg)
- Playback Method: Video file on local drive
- Hardware Acceleration: None
- Plugins: None
- Reverse Proxy: None
- Base URL: http://<IPv4 address>:8096
- Networking: LAN
- Storage: External hard drive
```

### Jellyfin logs

On playback:
```
[...]

PR #8165, opened 2022-07-23
Make first stream default if none is set

gaudat:
Refer to #8164.


gaudat:
Logic seems not working on my side. Converting to draft while I fix it.

gaudat:
8957d5c passes unit tests with the "default" field in test data set to 0. Let me know if you want to add an extra test case for this situation. As I do not know how frequent this happens.

gaudat:
Tested working: deb arm64 on RPi 3B+. Reqiures a full library rescan to work as ffprobe is only run on scan. 

Alternatively we can put the logic in `StreamBuilder.cs`.

cvium:
> Alternatively we can put the logic in StreamBuilder.BuildAudioItem or StreamBuilder.BuildVideoItem.

If it's not too complex I think it might be preferable

gaudat:
With current master I cannot reproduce the reason SecondaryAudioNotSupported . Maybe this has been fixed between the old version I had and the latest version.

Closing this now.

PR #8174, opened 2022-07-26
Disable auto inserted SW scaler for HW decoders

nyanmisaka:
**Changes**
- Disable auto inserted SW scaler for HW decoders in case of changed resolution
- Fix DV P5 MKV remuxing

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -931,6 +931,13 @@ namespace MediaBrowser.Controller.MediaEncoding
                 arg.Append(" -i \"").Append(state.AudioStream.Path).Append('"');
             }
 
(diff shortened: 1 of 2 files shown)


MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -931,6 +931,13 @@ public string GetInputArgument(EncodingJobInfo state, EncodingOptions options, s
                 arg.Append(" -i \"").Append(state.AudioStream.Path).Append('"');
             }
 
+            // Disable auto inserted SW scaler for HW decoders in case of changed resolution.
+            var isSwDecoder = string.IsNullOrEmpty(GetHardwareVideoDecoder(state, options) ?? string.Empty);
```
cvium, line 935: The null coalescing is not necessary

cvium (approved):

crobibero (approved):

PR #8203, opened 2022-08-02
Implement NFO named season parsing

### LINKED ISSUE #8489: [Issue]: tvshow.nfo `namedseason` not working
### Please describe your bug

```
<?xml version='1.0' encoding='utf-8'?>
<tvshow>
  <genre>Music</genre>
  <namedseason number="1">Videos</namedseason>
  <namedseason number="2">Music Videos</namedseason>
  <title>Beyond the Guitar</title>
</tvshow>
```
The `tvshow.nfo` above does not recognize named seasons when importing custom tv shows. Names them as Season 1, Season 2, etc

See namedseason kodi usage here: https://kodi.wiki/view/NFO_files/TV_shows

### Jellyfin Version

10.8.0

### if other:

_No response_

### Environment

```markdown
- OS:
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

Shadowghost:
I'm not sure if I've done this correctly. The main problem is that season names are within the series NFO, so we need to process this data before the seasons themself are processed.

Fixes #8489

### DIFF
--- a/MediaBrowser.Controller/Entities/TV/Series.cs
+++ b/MediaBrowser.Controller/Entities/TV/Series.cs
@@ -28,11 +28,15 @@ namespace MediaBrowser.Controller.Entities.TV
         public Series()
         {
             AirDays = Array.Empty<DayOfWeek>();
+            SeasonNames = new Dictionary<int, string>();
         }
 
         public DayOfWeek[] AirDays { get; set; }
 
         public string AirTime { get; set; }
+
+        [JsonIgnore]
+        public Dictionary<int, string> SeasonNames { get; set; }
 
         [JsonIgnore]
         public override bool SupportsAddingToPlaylist => true;
--- a/MediaBrowser.Providers/Manager/MetadataService.cs
+++ b/MediaBrowser.Providers/Manager/MetadataService.cs
@@ -12,6 +12,7 @@ using Jellyfin.Extensions;
 using MediaBrowser.Controller.Configuration;
 using MediaBrowser.Controller.Entities;
 using MediaBrowser.Controller.Entities.Audio;
+using MediaBrowser.Controller.Entities.TV;
 using MediaBrowser.Controller.Library;
 using MediaBrowser.Controller.Providers;
 using MediaBrowser.Model.Configuration;
--- a/MediaBrowser.Providers/TV/SeriesMetadataService.cs
+++ b/MediaBrowser.Providers/TV/SeriesMetadataService.cs
@@ -79,10 +79,7 @@ namespace MediaBrowser.Providers.TV
                 foreach (var season in sourceSeasonNames)
                 {
                     var seasonNumber = season.Key;
-                    if (!targetSeasonNames.ContainsKey(seasonNumber))
-                    {
-                        targetItem.SeasonNames[seasonNumber] = season.Value;
-                    }
+                    targetSeasonNames.TryAdd(seasonNumber, season.Value);
                 }
             }
 
@@ -219,9 +216,9 @@ namespace MediaBrowser.Providers.TV
                 var existingSeason = seasons.FirstOrDefault(i => i.IndexNumber == seasonNumber);
                 string? seasonName = null;
 
-                if (seasonNumber.HasValue && seasonNames.ContainsKey(seasonNumber.Value))
+                if (seasonNumber.HasValue && seasonNames.TryGetValue(seasonNumber.Value, out var tmp))
                 {
-                    seasonName = seasonNames[seasonNumber.Value];
+                    seasonName = tmp;
                 }
 
                 if (existingSeason is null)
--- a/MediaBrowser.Providers/TV/SeriesMetadataService.cs
+++ b/MediaBrowser.Providers/TV/SeriesMetadataService.cs
@@ -41,7 +41,7 @@ namespace MediaBrowser.Providers.TV
 
             RemoveObsoleteEpisodes(item);
             RemoveObsoleteSeasons(item);
-            await FillInMissingSeasonsAsync(item, cancellationToken).ConfigureAwait(false);
+            await UpdateAndCreateSeasonsAsync(item, cancellationToken).ConfigureAwait(false);
         }
 
         /// <inheritdoc />
@@ -67,6 +67,24 @@ namespace MediaBrowser.Providers.TV
 
             var sourceItem = source.Item;
             var targetItem = target.Item;
+            var sourceSeasonNames = sourceItem.SeasonNames;
+            var targetSeasonNames = targetItem.SeasonNames;
+
+            if (replaceData || targetSeasonNames.Count == 0)
+            {
+                targetItem.SeasonNames = sourceSeasonNames;
+            }
+            else if (targetSeasonNames.Count != sourceSeasonNames.Count || !sourceSeasonNames.Keys.All(targetSeasonNames.ContainsKey))
+            {
+                foreach (var season in sourceSeasonNames)
+                {
+                    var seasonNumber = season.Key;
+                    if (!targetSeasonNames.ContainsKey(seasonNumber))
+                    {
+                        targetItem.SeasonNames[seasonNumber] = season.Value;
+                    }
+                }
+            }
 
             if (replaceData || string.IsNullOrEmpty(targetItem.AirTime))
             {
@@ -86,7 +104,7 @@ namespace MediaBrowser.Providers.TV
 
         private void RemoveObsoleteSeasons(Series series)
         {
-            // TODO Legacy. It's not really "physical" seasons as any virtual seasons are always converted to non-virtual in FillInMissingSeasonsAsync.
+            // TODO Legacy. It's not really "physical" seasons as any virtual seasons are always converted to non-virtual in UpdateAndCreateSeasonsAsync.
             var physicalSeasonNumbers = new HashSet<int>();
             var virtualSeasons = new List<Season>();
             foreach (var existingSeason in series.Children.OfType<Season>())
@@ -177,36 +195,43 @@ namespace MediaBrowser.Providers.TV
         }
 
         /// <summary>
-        /// Creates seasons for all episodes that aren't in a season folder.
+        /// Creates seasons for all episodes if they don't exist.
         /// If no season number can be determined, a dummy season will be created.
+        /// Updates seasons names.
         /// </summary>
         /// <param name="series">The series.</param>
         /// <param name="cancellationToken">The cancellation token.</param>
         /// <returns>The async task.</returns>
-        private async Task FillInMissingSeasonsAsync(Series series, CancellationToken cancellationToken)
+        private async Task UpdateAndCreateSeasonsAsync(Series series, CancellationToken cancellationToken)
         {
+            var seasonNames = series.SeasonNames;
             var seriesChildren = series.GetRecursiveChildren(i => i is Episode || i is Season);
-            var episodesInSeriesFolder = seriesChildren
+            var seasons = seriesChildren.OfType<Season>().ToList();
+            var uniqueSeasonNumbers = seriesChildren
                 .OfType<Episode>()
-                .Where(i => !i.IsInSeasonFolder);
-
-            List<Season> seasons = seriesChildren.OfType<Season>().ToList();
+                .Select(e => e.ParentIndexNumber >= 0 ? e.ParentIndexNumber : null)
+                .Distinct();
 
             // Loop through the unique season numbers
-            foreach (var episode in episodesInSeriesFolder)
+            foreach (var seasonNumber in uniqueSeasonNumbers)
             {
                 // Null season numbers will have a 'dummy' season created because seasons are always required.
-                var seasonNumber = episode.ParentIndexNumber >= 0 ? episode.ParentIndexNumber : null;
                 var existingSeason = seasons.FirstOrDefault(i => i.IndexNumber == seasonNumber);
+                string? seasonName = null;
+
+                if (seasonNumber.HasValue && seasonNames.ContainsKey(seasonNumber.Value))
+                {
+                    seasonName = seasonNames[seasonNumber.Value];
+                }
 
                 if (existingSeason is null)
                 {
-                    var season = await CreateSeasonAsync(series, seasonNumber, cancellationToken).ConfigureAwait(false);
-                    seasons.Add(season);
+                    var season = await CreateSeasonAsync(series, seasonName, seasonNumber, cancellationToken).ConfigureAwait(false);
+                    series.AddChild(season);
                 }
-                else if (existingSeason.IsVirtualItem)
+                else
                 {
-                    existingSeason.IsVirtualItem = false;
+                    existingSeason.Name = GetValidSeasonNameForSeries(series, seasonName, seasonNumber);
                     await existingSeason.UpdateToRepositoryAsync(ItemUpdateType.MetadataEdit, cancellationToken).ConfigureAwait(false);
                 }
             }
@@ -216,21 +241,17 @@ namespace MediaBrowser.Providers.TV
         /// Creates a new season, adds it to the database by linking it to the [series] and refreshes the metadata.
         /// </summary>
         /// <param name="series">The series.</param>
+        /// <param name="seasonName">The season name.</param>
         /// <param name="seasonNumber">The season number.</param>
         /// <param name="cancellationToken">The cancellation token.</param>
         /// <returns>The newly created season.</returns>
         private async Task<Season> CreateSeasonAsync(
             Series series,
+            string? seasonName,
             int? seasonNumber,
             CancellationToken cancellationToken)
         {
-            string seasonName = seasonNumber switch
-            {
-                null => _localizationManager.GetLocalizedString("NameSeasonUnknown"),
-                0 => LibraryManager.GetLibraryOptions(series).SeasonZeroDisplayName,
-                _ => string.Format(CultureInfo.InvariantCulture, _localizationManager.GetLocalizedString("NameSeasonNumber"), seasonNumber.Value)
-            };
-
+            seasonName = GetValidSeasonNameForSeries(series, seasonName, seasonNumber);
             Logger.LogInformation("Creating Season {SeasonName} entry for {SeriesName}", seasonName, series.Name);
 
             var season = new Season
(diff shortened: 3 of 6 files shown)


jmbannon:
This would be a great addition for Jellyfin, bumping to show support

cvium:
> This would be a great addition for Jellyfin, bumping to show support

There are Reactions for this sort of thing.

MediaBrowser.Controller/Entities/TV/Series.cs
```diff
@@ -34,6 +34,8 @@ public Series()
 
         public string AirTime { get; set; }
 
+        public Dictionary<int, string> SeasonNames { get; set; }
```
cvium, line 37: I'm still not sold on this

Shadowghost: I don't like it either but haven't found a better solution for it. Additional input is always welcome 😀

Shadowghost: Would this need to be saved to the database too?

jmbannon: Friendly bump, what can we do to make this work @joshuaboniface @cvium ?

joshuaboniface (approved):

MediaBrowser.Providers/Manager/MetadataService.cs
```diff
@@ -1090,6 +1125,18 @@ private bool HasChanged(BaseItem item, IHasItemChangeMonitor changeMonitor, IDir
                 }
             }
 
+            if (source is Series sourceSeries && target is Series targetSeries)
+            {
+                if (replaceData || string.IsNullOrEmpty(target.ForcedSortName))
```
1337joe, line 1130:
```suggestion
                if (replaceData || target.SeasonNames == null)
```

1337joe (changes requested):

1337joe (approved):

MediaBrowser.Providers/Manager/MetadataService.cs
1337joe: My mistake, this broke it: should have been `targetSeries.SeasonNames`

daullmer:
Maybe it would be nice to have infos about this new feature [here](https://jellyfin.org/docs/general/server/metadata/nfo) when this gets merged?

MediaBrowser.Providers/TV/SeriesMetadataService.cs
```diff
@@ -67,6 +67,24 @@ protected override void MergeData(MetadataResult<Series> source, MetadataResult<
...
+            if (replaceData || targetSeasonNames is null)
+            {
+                targetItem.SeasonNames = sourceSeasonNames;
+            }
+            else if (!targetSeasonNames.Equals(sourceSeasonNames))
+            {
+                foreach (var season in sourceSeasonNames)
+                {
+                    var seasonNumber = season.Key;
+                    if (string.IsNullOrEmpty(targetSeasonNames[seasonNumber]))
```
cvium, line 82: Won't this crash sometimes?

MediaBrowser.Providers/TV/SeriesMetadataService.cs
```diff
@@ -67,6 +67,24 @@ protected override void MergeData(MetadataResult<Series> source, MetadataResult<
...
             var sourceItem = source.Item;
             var targetItem = target.Item;
+            var sourceSeasonNames = sourceItem.SeasonNames;
+            var targetSeasonNames = targetItem.SeasonNames;
+
+            if (replaceData || targetSeasonNames is null)
+            {
+                targetItem.SeasonNames = sourceSeasonNames;
+            }
+            else if (!targetSeasonNames.Equals(sourceSeasonNames))
```
cvium, line 77: Is it supposed to be reference equals?

MediaBrowser.Providers/TV/SeriesMetadataService.cs
```diff
@@ -177,36 +195,42 @@ private void DeleteEpisode(Episode episode)
...
-                    seasons.Add(season);
+                    var seasonName = seasonNames?.Where(s => s.Key == sNumber).FirstOrDefault();
+                    var season = await CreateSeasonAsync(series, seasonName?.Value, sNumber, cancellationToken).ConfigureAwait(false);
+                    series.AddChild(season);
                 }
-                else if (existingSeason.IsVirtualItem)
+                else
                 {
                     existingSeason.IsVirtualItem = false;
+                    var seasonName = seasonNames?.Where(s => s.Key == existingSeason.IndexNumber).FirstOrDefault().Value;
```
cvium, line 232: Crash if Default

MediaBrowser.XbmcMetadata/Parsers/SeriesNfoParser.cs
```diff
@@ -110,6 +112,29 @@ protected override void FetchDataFromXmlNode(XmlReader reader, MetadataResult<Se
...
+                case "namedseason":
+                    {
+                        var parsed = int.TryParse(reader.GetAttribute("number"), NumberStyles.Integer, CultureInfo.InvariantCulture, out var seasonNumber);
+                        var name = reader.ReadElementContentAsString();
+
+                        if (!string.IsNullOrWhiteSpace(name) && parsed)
+                        {
+                            if (item.SeasonNames == null)
+                            {
+                                item.SeasonNames = new Dictionary<int, string>()
```
cvium, line 124: Aren't the parentheses redundant?

MediaBrowser.Providers/TV/SeriesMetadataService.cs
```diff
@@ -177,36 +195,42 @@ private void DeleteEpisode(Episode episode)
...
-                var seasonNumber = episode.ParentIndexNumber >= 0 ? episode.ParentIndexNumber : null;
-                var existingSeason = seasons.FirstOrDefault(i => i.IndexNumber == seasonNumber);
+                var sNumber = seasonNumber >= 0 ? seasonNumber : null;
+                var existingSeason = seasons.FirstOrDefault(i => i.IndexNumber == sNumber);
 
                 if (existingSeason is null)
                 {
-                    var season = await CreateSeasonAsync(series, seasonNumber, cancellationToken).ConfigureAwait(false);
-                    seasons.Add(season);
+                    var seasonName = seasonNames?.Where(s => s.Key == sNumber).FirstOrDefault();
```
cvium, line 225: Why a Where? Isn't it a dictionary?

MediaBrowser.Providers/TV/SeriesMetadataService.cs
cvium: .Equals for Collections are somewhat undefined, so it almost always falls back to ReferenceEquals

MediaBrowser.XbmcMetadata/Parsers/SeriesNfoParser.cs
Shadowghost: The outer if clause checks if the name is set and parsing the series number to an int was successfull, the inner if checks if the item already has a dictionary assigned, if it hasn't, one is created with the new K/V pair, if it has the K/V pair is added to the existing dict.

Or does the item automatically have a dictionary assigned because the property is not set as nullable in the class - that would render the check useless?

cvium: When you use an initializer, the parentheses are redundant e.g.

`var myList = new List<int> { 1, 2, 3, 4 };`

Bond-009: TBF, I still type them

MediaBrowser.Providers/TV/SeriesMetadataService.cs
```diff
@@ -67,6 +67,25 @@ protected override void MergeData(MetadataResult<Series> source, MetadataResult<
...
             var targetItem = target.Item;
+            var sourceSeasonNames = sourceItem.SeasonNames;
+            var targetSeasonNames = targetItem.SeasonNames;
+
+            if (replaceData || targetSeasonNames is null)
+            {
+                targetItem.SeasonNames = sourceSeasonNames;
+            }
+            else if (!(targetSeasonNames.Count == sourceSeasonNames.Count
+                && targetSeasonNames.Keys.All(sourceSeasonNames.ContainsKey)))
```
cvium, line 78: I don't think much is gained by having this check (line 77 is fine). `Keys` creates a `KeyCollection` (a class), which you then enumerate with LINQ, so nothing is really saved. I'm also not sure the check is correct. You only want to set _new_ key-value pairs in targetItem, but you check whether all targetItem keys are in sourceItem. Shouldn't it be the other way around?

cvium (changes requested):
See previous comment

MediaBrowser.Providers/TV/SeriesMetadataService.cs
Shadowghost: The condition in itself is inverted, so I think the logic is correct.
What I wanted to achieve is this: If lengths are not the same and the series does not have a name for a season the new season names are added.

But as you pointed out, the `SeasonNames` is not persisted within the database, so it is lost on restart, I guess we can circumvent it by manually defining a getter, building the dictionary based on the season metadata.

cvium: It works because if the count is the same, then they are either identical or not, but it's not "correct". It should've been `targetSeasonNames.Count != sourceSeasonNames.Count
			  || !sourceSeasonNames.Keys.All(targetSeasonNames.ContainsKey)`

MediaBrowser.Providers/TV/SeriesMetadataService.cs
```diff
@@ -177,36 +196,48 @@ private void DeleteEpisode(Episode episode)
...
             List<Season> seasons = seriesChildren.OfType<Season>().ToList();
 
             // Loop through the unique season numbers
-            foreach (var episode in episodesInSeriesFolder)
+            foreach (var seasonNumber in uniqueSeasonNumbers)
             {
                 // Null season numbers will have a 'dummy' season created because seasons are always required.
-                var seasonNumber = episode.ParentIndexNumber >= 0 ? episode.ParentIndexNumber : null;
-                var existingSeason = seasons.FirstOrDefault(i => i.IndexNumber == seasonNumber);
+                var sNumber = seasonNumber >= 0 ? seasonNumber : null;
```
cvium, line 221: Wouldn't it be better to put this in the Select on line 212?

MediaBrowser.Providers/TV/SeriesMetadataService.cs
Shadowghost: Right, the keys could be different if the length is the same.

cvium: It's the same result in the end, but the intention somewhat differs.

MediaBrowser.Providers/TV/SeriesMetadataService.cs
```diff
@@ -177,37 +195,51 @@ private void DeleteEpisode(Episode episode)
...
             var seriesChildren = series.GetRecursiveChildren(i => i is Episode || i is Season);
-            var episodesInSeriesFolder = seriesChildren
+            var uniqueSeasonNumbers = seriesChildren
                 .OfType<Episode>()
-                .Where(i => !i.IsInSeasonFolder);
-
-            List<Season> seasons = seriesChildren.OfType<Season>().ToList();
+                .Select(e => e.ParentIndexNumber)
+                .Distinct()
+                .Select(s => s >= 0 ? s : null);
```
cvium, line 212:
```suggestion
                .Select(e => e.ParentIndexNumber >= 0 ? e.ParentIndexNumber : null)
                .Distinct();
```

MediaBrowser.Providers/TV/SeriesMetadataService.cs
```diff
@@ -177,37 +195,51 @@ private void DeleteEpisode(Episode episode)
...
 
-            // Loop through the unique season numbers
-            foreach (var episode in episodesInSeriesFolder)
+            if (uniqueSeasonNumbers is not null)
```
cvium, line 214: It can never be null?

Shadowghost: it can be null if the LINQ statement returns nothing. It does not throw anywhere within the function though.

cvium: But uniqueSeasonNumbers is a LINQ statement. How can it be null?

Shadowghost: According to VSCode `uniqueSeasonNumbers` is of type `IEnumerable<int?>?`. Maybe it is possible if there is no child with `OfType<Episode>()`? Or should the latter result in an empty `IEnumerable`?

cvium: Whenever you use `var`, the type is nullable.

cvium: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/declarations#implicitly-typed-local-variables

MediaBrowser.Providers/TV/SeriesMetadataService.cs
```diff
@@ -177,36 +195,45 @@ private void DeleteEpisode(Episode episode)
...
+            foreach (var seasonNumber in uniqueSeasonNumbers)
             {
                 // Null season numbers will have a 'dummy' season created because seasons are always required.
-                var seasonNumber = episode.ParentIndexNumber >= 0 ? episode.ParentIndexNumber : null;
                 var existingSeason = seasons.FirstOrDefault(i => i.IndexNumber == seasonNumber);
+                string? seasonName = null;
+
+                if (seasonNumber.HasValue
+                    && seasonNames.ContainsKey(seasonNumber.Value)
+                )
```
cvium, line 224: I don't think we usually put it on a separate line
```suggestion
                    && seasonNames.ContainsKey(seasonNumber.Value))
```

cvium (approved):

MediaBrowser.Providers/TV/SeriesMetadataService.cs
```diff
@@ -67,6 +67,24 @@ protected override void MergeData(MetadataResult<Series> source, MetadataResult<
...
+            }
+            else if (targetSeasonNames.Count != sourceSeasonNames.Count || !sourceSeasonNames.Keys.All(targetSeasonNames.ContainsKey))
+            {
+                foreach (var season in sourceSeasonNames)
+                {
+                    var seasonNumber = season.Key;
+                    if (!targetSeasonNames.ContainsKey(seasonNumber))
+                    {
+                        targetItem.SeasonNames[seasonNumber] = season.Value;
+                    }
```
Bond-009, line 85:
```suggestion
                    targetSeasonNames.TryAdd(seasonNumber, season.Value);
```

MediaBrowser.Providers/TV/SeriesMetadataService.cs
```diff
@@ -177,36 +195,43 @@ private void DeleteEpisode(Episode episode)
...
+            foreach (var seasonNumber in uniqueSeasonNumbers)
             {
                 // Null season numbers will have a 'dummy' season created because seasons are always required.
-                var seasonNumber = episode.ParentIndexNumber >= 0 ? episode.ParentIndexNumber : null;
                 var existingSeason = seasons.FirstOrDefault(i => i.IndexNumber == seasonNumber);
+                string? seasonName = null;
+
+                if (seasonNumber.HasValue && seasonNames.ContainsKey(seasonNumber.Value))
+                {
+                    seasonName = seasonNames[seasonNumber.Value];
```
Bond-009, line 224:
```suggestion
                if (seasonNumber.HasValue && seasonNames.TryGetValue(seasonNumber.Value, out var tmp))
                {
                    seasonName = tmp;
```

Bond-009 (changes requested):

MediaBrowser.Providers/TV/SeriesMetadataService.cs
```diff
@@ -67,6 +67,21 @@ protected override void MergeData(MetadataResult<Series> source, MetadataResult<
...
             var sourceItem = source.Item;
             var targetItem = target.Item;
+            var sourceSeasonNames = sourceItem.SeasonNames;
+            var targetSeasonNames = targetItem.SeasonNames;
+
+            if (replaceData || targetSeasonNames.Count == 0)
+            {
+                targetItem.SeasonNames = sourceSeasonNames;
+            }
+            else if (targetSeasonNames.Count != sourceSeasonNames.Count || !sourceSeasonNames.Keys.All(targetSeasonNames.ContainsKey))
```
Bond-009, line 77: Is the is second part really necessary?

MediaBrowser.Providers/TV/SeriesMetadataService.cs
```diff
@@ -67,6 +67,21 @@ protected override void MergeData(MetadataResult<Series> source, MetadataResult<
...
+            {
+                foreach (var season in sourceSeasonNames)
+                {
+                    var seasonNumber = season.Key;
+                    targetSeasonNames.TryAdd(seasonNumber, season.Value);
```
Bond-009, line 82:
```suggestion
                foreach (var (number, name) in sourceSeasonNames)
                {
                    targetSeasonNames.TryAdd(number, name);
```

Bond-009 (approved):

MediaBrowser.Providers/TV/SeriesMetadataService.cs
Shadowghost: Yes in case the season names got changed. the main "problem" is that the season names are not persisted within the `Series` object.
I tried modifying the getter to either use the local prop if it has any length and if it has none get the season names from the database but it did never work as I wanted.

Might be worth taking another look though because it is a cleaner solution.

vegedb:
Since this is fixed, 
May we know what's the naming convention in nfo?
This was the previous,
` <namedseason number="1">Videos</namedseason>`
` <namedseason number="2">Music Videos</namedseason>`

@jmbannon @Bond-009  @Shadowghost

Shadowghost:
We didn't change the convention but are respecting it now.
This fix is not in the current stable release though.

vegedb:
> We didn't change the convention but are respecting it now. This fix is not in the current stable release though.

Does the recent update [10.8.12](https://github.com/jellyfin/jellyfin/releases/tag/v10.8.12) implement this?

crobibero:
> > We didn't change the convention but are respecting it now. This fix is not in the current stable release though.
> 
> Does the recent update [10.8.12](https://github.com/jellyfin/jellyfin/releases/tag/v10.8.12) implement this?

No, version 10.9 will be the first version with this change.

jmbannon:
Playing around with this and noticed named seasons require it to be in `season.nfo`, is this intended? Kodi defines `namedseason` in the `tvshow.nfo` (https://kodi.wiki/view/NFO_files/TV_shows). I'm not sure where the notion of `season.nfo` came from in the first place, but it's not part of the Kodi standard. IMO we should correct this so JF has 100% compatibility with Kodi

jmbannon:
Actually, disregard -- I think it was a user-error. Have it working with the following info:

versions: `jellyfin-ffmpeg5_5.1.3-7-jammy_amd64.deb jellyfin-server_20231115.15-unstable_amd64.deb`
tvshow.nfo:
```
<?xml version='1.0' encoding='utf-8'?>
<tvshow>
  <genre>ytdl-sub</genre>
  <mpaa>TV-14</mpaa>
  <namedseason number="1">Named Season 1</namedseason>
  <title>Best Prebuilt TV Show Collection</title>
</tvshow>
```
screenshot:
image

Thanks again for this feature and looking very forward to 10.9 becoming stable :slightly_smiling_face:

PR #8213, opened 2022-08-06
Fix high single thread usage in throttler

nyanmisaka:
**Changes**
- Fix high single thread usage in throttler (this requires [jellyfin-ffmpeg >= 5.0.1-8](https://github.com/jellyfin/jellyfin-ffmpeg/pull/165))

> free single core usage from [c]. And it has a faster resume speed than other methods. vf_realtime filter has also been considered but it doesn't support remux. [p] to pause, [u] or any other key to resume.

### DIFF
--- a/Jellyfin.Api/Models/PlaybackDtos/TranscodingThrottler.cs
+++ b/Jellyfin.Api/Models/PlaybackDtos/TranscodingThrottler.cs
@@ -2,6 +2,7 @@
 using System.Threading;
 using System.Threading.Tasks;
 using MediaBrowser.Common.Configuration;
+using MediaBrowser.Controller.MediaEncoding;
 using MediaBrowser.Model.Configuration;
 using MediaBrowser.Model.IO;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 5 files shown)


Shadowghost (approved):

crobibero (approved):

Jellyfin.Api/Models/PlaybackDtos/TranscodingThrottler.cs
```diff
@@ -55,7 +59,8 @@ public async Task UnpauseTranscoding()
 
                 try
                 {
-                    await _job.Process!.StandardInput.WriteLineAsync().ConfigureAwait(false);
+                    var resumeKey = _mediaEncoder.IsPkeyPauseSupported ? "u" : "\n";
```
cvium, line 62: Is `\n` sufficient on all platforms or should it be `Environment.NewLine`?

cvium (approved):

n-peugnet:
Thank you very much for this and https://github.com/jellyfin/jellyfin-ffmpeg/pull/165 @nyanmisaka !

PR #8246, opened 2022-08-11
Respect visibility for people items

LogicalPhallacy:
**Changes**
This PR updates the library manager and DtoService to respect the IsVisible check on People items, to address what I am calling the "Game of Thrones crossover" problem [reported on reddit](https://www.reddit.com/r/jellyfin/comments/wl0kpm/search_spans_all_people_in_all_libraries_ignores/). In short an actor might have a headshot photo that is not appropriate for some users, and currently all people show up in search regardless of content restrictions. The changes to the library address the first problem, but as elaborated on in the thread, shows like Game of Thrones pulled in performers from adult entertainment, and they could bring problematic headshot photos with them as well. This enables the ability for someone trying to restrict this to tag those people with a tag that is then applied as a Tag based parental control and have that tag prevent them from showing up for a restricted user, either in search or on a media item's people list.


wkearney99:
I am the user that raised this question on reddit.  I simply stumbled across the 'leak' of nsfw actor headshots showing up while logged in as a user that did not have access to any of the content in which that person was listed.  The source movie was in a library not accessible to the user.  But a search that happened to use part of the person name brought up People as part of the results, and there was a nsfw headshot picture for a matching person.

One fix would, of course, be to manually change the headshot pictures to something not nsfw.  But then there's the question of person names themselves being problematic showing up in results.  

I naively assumed that if a user didn't have access to content in a library then nothing (including the people) would be accessible in any form.  I didn't know that search was 'global' in nature and that people headshots would be included from inaccessible content.  

I get that it adds a degree of computational effort to do additional filtering.  I can't speak to that as I really don't know much at all about the internals JF uses to manage all this.

joshuaboniface:
This could probably target the 10.8.z release branch if desired, it's a pretty good bugfix.

LogicalPhallacy:
> This could probably target the 10.8.z release branch if desired, it's a pretty good bugfix.

If you think that's best I am fine with retargeting this PR against that branch. Given that this "bug" is less a bug and more an unintended omission, and that we've only had one report of it, I wouldn't think it merits a bugfix release on its own, but if there are a few others coming to 10.8.x I think it'd be a good fit.

I haven't profiled it yet, but I would imagine this will cause a slight perf hit on people searches due to the addition IsVisible computations, but they're infrequent enough from what I could see running this down that I don't think it will add too much strain overall.

Shadowghost:
We have other bugfixes scheduled for 10.8.4 so it should be fine to go in there too. Because imo it's still a bug worth fixing before the nex major version.

LogicalPhallacy:
Aw nuts, I forgot my fork was off master,. Let me rebase this against 10.8.z before any additional actions happen.

cvium:
> Aw nuts, I forgot my fork was off master,. Let me rebase this against 10.8.z before any additional actions happen.

@LogicalPhallacy What's the status on this?

Bond-009:
Superseded by #8399

PR #8247, opened 2022-08-11
Don't mark anything as unwatched when you start playing it

### LINKED ISSUE #7535: [Issue]: Playing a watched episode and then stopping within the first few minutes marks episode as unwatched.
### Please describe your bug

I noticed this in the web player and in a 3rd party client as well

1. Play a watched episode
2. let episode play for 10 seconds
3. stop playback
4. episode is marked as unwatched

Episode should not be marked as unwatched. It is still watched, just wasn't watched again.

Here is UserData for the episode after it was marked unwatched.

```
UserData: {
PlaybackPositionTicks: 0,
PlayCount: 2,
IsFavorite: false,
LastPlayedDate: "2022-03-31T18:03:30.8895109Z",
Played: false,
Key: "121361008006"
}
```

And then after I manually marked it as watched to fix the problem
```
UserData: {
PlaybackPositionTicks: 0,
PlayCount: 3,
IsFavorite: false,
LastPlayedDate: "2022-03-31T18:12:16.576Z",
Played: true,
Key: "121361008006"
},
```

### Jellyfin Version

10.8.0

### if other:

_No response_

### Environment

```markdown
- OS:
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

SenorSmartyPants:
**Changes**

Starting to play any item that is marked as played and supports resuming (a movie for example), immediately marks this item as unwatched. Starting to play an item should never mark it as unwatched.

Stable Backport

**Issues**

Fixes #7535

### DIFF
--- a/Emby.Server.Implementations/Session/SessionManager.cs
+++ b/Emby.Server.Implementations/Session/SessionManager.cs
@@ -738,10 +738,6 @@ namespace Emby.Server.Implementations.Session
             {
                 data.Played = true;
             }
-            else
-            {
-                data.Played = false;
-            }
 
             _userDataManager.SaveUserData(user, item, data, UserDataSaveReason.PlaybackStart, CancellationToken.None);
         }


crobibero:
- This changes existing (intended) behavior, so it will need to be user configurable.
- This isn't bugfix, and as such will not be ported into 10.8.

SenorSmartyPants:
Where can I find a description of the intended behavior? It currently doesn't make much sense to me to mark something as unwatched when you watch it again.

SenorSmartyPants:
Is there a specific settings area you think this should go in? I'd rather it be a server wide setting. Well I don't think it should be a setting at all but whatever. Options are better than no options.

cvium:
The play count still increases, so I don't see this as much of an issue. I believe Plex behaves in the same way (sort of).

zehnerGIT:
just my 2 cents:
I think it's good behaviour to mark an episode as unwatched if you rewatch it and stop the playback after a few minutes, but perhaps this should respect the settings under Server -> Playback -> Resume (Minimum resume percentage and/or duration). If the resume status doesn't switch, then the watched status should not switch, too.
From time to time I click the wrong button/area or start the wrong playback and immediately stop it. Then the file is marked as unwatched and I have to mark it as watched again, so the playcount is increased by 1 for a wrong playback attempt. If the marking would respect these settings, the scenario in the issue (10sec playback) and my scenario (1sec) are no longer a problem, at least with the default values (5% and 300sec).

SenorSmartyPants:
My interpretation based on property names of what these properties _**should**_ represent:
- `Played`: Has this item _**ever**_ been played to completion
- `PlayCount`: How many times has item been played to completion

After reviewing more code, what it seems like these properties represent:
- `Played`: Has this item been played to completion the last time it was played.
- `PlayCount`: How many times has item been played, even for a second. In other words how many times has the play button been pressed for this item.

> The play count still increases, so I don't see this as much of an issue. I believe Plex behaves in the same way (sort of). What is the motivation for this change exactly? What does it achieve?

**Played**
I can understand the current behavior for items that don't `SupportsPositionTicksResume`. Since there's no way to mark progess in those items, marking them as unplayed is the only way to show that info. I don't want to change that. 

Otherwise it should be left solely up to `_userDataManager.UpdatePlayState`, Played should not be set to false for items that `SupportsPositionTicksResume` during `OnPlaybackStart` as is the current behavior. Watching something again doesn't mean I never watched it.

**What does it achieve?**
If I accidentally start watching something and stop, it is now marked as unplayed. If I manually mark this as played to correct it, trakt plugin will sync this as another viewing. JF marking things as unplayed could remove play history during a trakt sync.

If I want to filter my library to unplayed items, I don't want to see anything I've seen before but might be marked as unplayed because of accidental clicks, or something I'm in progress rewatching.

**PlayCount**
The only use I can find for `PlayCount` is in sorting, it's not displayed anywhere in the UI. But how relevant is `PlayCount` when it can't tell the difference between me taking 3 sessions to finish a 3 hour movie (that's a `PlayCount` of 3) vs watching a movie to completion 3 times. That kind of information is useless to me.

SenorSmartyPants:
TL;DR 

The core of my argument is this, current behavior is:

1. Play item
2. Item marked as unplayed
This is not a good design choice. Prove be wrong.

PR #8259, opened 2022-08-14
Increase (really) timeout for subtitle extraction to 30min

RealGreenDragon:
**Changes**
The PR #7153 increases subtitle extraction operation timeout from 5 minutes to 30 minutes (to address long-running extractions due for example to a slow storage).
But such PR changes the timeout only in method that manage subtitle conversion and not in the method that manage subtitle extraction.
This PR completes the change.

**Issues**
Addresses #6672
Complete PR #7153


daullmer (approved):

crobibero (approved):

RealGreenDragon:
I missed to back-port this PR also in 10.8.z branch, so I opened a new PR just now.

PR #8261, opened 2022-08-14
Use guid for adjacentTo API parameter

nielsvanvelzen:
The adjacentTo parameter required an item id or null. Item ids are always guid.

**Changes**
- Use Guid for adjacentTo API parameter

**Issues**


Jellyfin.Api/Controllers/TvShowsController.cs
```diff
@@ -278,9 +278,9 @@ public class TvShowsController : BaseJellyfinApiController
             }
 
             // This must be the last filter
-            if (!string.IsNullOrEmpty(adjacentTo))
+            if (adjacentTo.HasValue)
```
Bond-009, line 281: Should we check for empty guids?

Bond-009 (approved):

crobibero (approved):

PR #8266, opened 2022-08-15
Only use single decimal digit for user rating from TMDb

### LINKED ISSUE #8238: [Issue]: Fetched community rating not accepted when editing metadata
### Please describe your bug

Apparently the metadata fetcher can get the community rating in higher precision (8.266 instead of the usual two-digit 8.3). However, when trying to edit and save metadata, these additional digits seem to cause trouble; jellyfin wants me to reduce it to the usual two digits.

Steps to reproduce:
* Add "Zack Snyder's Justice League" (IMDB ID tt12361974, TMDB ID 791373). Other movies may work as well but most of them only have one or two digit resolution for the community rating.
* Open the dialog to edit its metadata. You should see a community rating of 8.266
* Click save (no need to edit anything).
* A message appears that only 8.2 or 8.3 are valid community ratings. 

Desired behavior: 
* Either the metadata fetcher does something wrong and those additional digits are garbage (a.k.a. they shouldn't have been fetched at all).
* Or, if they're actually accurate, the dialog should accept them. The additional digits don't seem to cause any issues. The movie page displays properly 8.3 as rating. 

Kind regards,
Max

### Jellyfin Version

Other

### if other:

10.8.1

### Environment

```markdown
- OS: Windows 10 LTSC 2019
- Virtualization:No
- Clients: Browser
- Browser: Firefox Desktop 103.0.1 x64
- FFmpeg Version: doesn't apply
- Playback Method: doesn't apply
- Hardware Acceleration: doesn't apply
- Plugins: doesn't apply 
- Reverse Proxy: nginx
- Base URL: none
- Networking: host
- Storage: local
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

image

### Code of Conduct

daullmer:
**Changes**
Not sure if there's a difference between rounding the float or the double

**Issues**
Fixes #8238

### DIFF
--- a/MediaBrowser.Providers/Plugins/Tmdb/Movies/TmdbMovieProvider.cs
+++ b/MediaBrowser.Providers/Plugins/Tmdb/Movies/TmdbMovieProvider.cs
@@ -198,7 +198,7 @@ namespace MediaBrowser.Providers.Plugins.Tmdb.Movies
                 movie.CollectionName = movieResult.BelongsToCollection.Name;
             }
 
-            movie.CommunityRating = Convert.ToSingle(movieResult.VoteAverage);
+            movie.CommunityRating = Convert.ToSingle(Math.Round(movieResult.VoteAverage, 1));
 
             if (movieResult.Releases?.Countries != null)
             {


crobibero (approved):

cvium:
This seems like the wrong fix. The backend should never be the limiting factor for precision.

cvium:
We should these two lines to `step="any"` https://github.com/jellyfin/jellyfin-web/blob/5d8b527b92847c4c8be82940e52e65a951beaffc/src/components/metadataEditor/metadataEditor.template.html#L72-L75

daullmer:
Closing in favour of a fix in jellyfin/jellyfin-web

PR #8279, opened 2022-08-17
Fix operation name in SearchController and correct nullability for SearchHint

nielsvanvelzen:
_Untested, I don't have a proper test environment_

**Changes**
- Fix operation name in SearchController
  Operation names should be unique across the application. The search thingy used `get` as name which is not that unique of a name
- Correct nullability for SearchHint
  - Basically looked at the nullability of BaseItemDto (which all info is copied from) and what's conditionally set in the controller to determine nullability
- Use BaseItemKind enum in SearchHint
  - GetClientTypeName returns basically the same but as a string

**Issues**

### DIFF
--- a/MediaBrowser.Model/Search/SearchHint.cs
+++ b/MediaBrowser.Model/Search/SearchHint.cs
@@ -1,8 +1,6 @@
-#nullable disable
-#pragma warning disable CS1591
-
 using System;
 using System.Collections.Generic;
+using Jellyfin.Data.Enums;
 
 namespace MediaBrowser.Model.Search
 {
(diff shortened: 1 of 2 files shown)


MediaBrowser.Model/Search/SearchHint.cs
```diff
@@ -1,8 +1,8 @@
-#nullable disable
-#pragma warning disable CS1591
+#pragma warning disable CS0618
```
Bond-009, line 1: Why is this needed? `ItemId` is referenced here, is it?

Bond-009 (approved):

MediaBrowser.Model/Search/SearchHint.cs
nielsvanvelzen: Id and ItemId are the same, from SearchController:

```
            // legacy
            result.ItemId = result.Id;
```

So I want to mark ItemId as obsolete so it generates the `deprecated: true` field in the OpenAPI specification. Can't figure out how to disable the error during build though..

crobibero (approved):

Bond-009:
Build seems to fail?

MediaBrowser.Model/Search/SearchHint.cs
Bond-009: I would expect the warning to be inside the `SearchController` file :thinking:

nielsvanvelzen:
Yeah build was failing because of the [Obsolete] thingy, fixed now.

MediaBrowser.Model/Search/SearchHint.cs
nielsvanvelzen: Yeah that's why it didn't work

Bond-009 (approved):

Issue #3930, opened 2020-08-18
How to config "FFmpeg:probesize" and “FFmpeg:analyzeduration”


ssyangchn:
**Describe the bug**

1. Im trying config the "FFmpeg:probesize" and “FFmpeg:analyzeduration” but errors in the log
```
.\jellyfin.exe --datadir "d:\ProgramData\Jellyfin\Server" --FFmpeg:probesize "50M" --FFmpeg:analyzeduration "10M"
Jellyfin.Server 10.6.3.0
Copyright ?  2019 Jellyfin Contributors. Code released under the GNU General Public License

ERROR(S):
  Option 'FFmpeg:probesize' is unknown.
  Option 'FFmpeg:analyzeduration' is unknown.
```
2. How to config jellyfin use a proxy ? not a reverse proxy

The [jellyfin doc: ](https://jellyfin.org/docs/general/administration/configuration.html#main-configuration-options)

```
Main Configuration
The main server configuration is built upon the ASP .NET configuration framework, which provides a tiered approach to loading configuration. The base directory to locate the configuration files is set using the configuration directory setting. The configuration sources are as follows, with later sources having higher priority and overwriting the values in earlier sources.

Hard-coded default values: These defaults are specified specified in the Jellyfin source code and cannot be changed.
Default logging configuration file (logging.default.json): This file should not be modified manually by users. It is reserved by the server to be overwritten with new settings on each new release.
System-specific logging configuration file (logging.json): This is the file you should change if you want to have a custom logging setup. Jellyfin uses the Serilog logging framework, and you can read about the configuration options available in their documentation.

NOTE
This file can be changed at runtime, which will automatically reload the configuration and apply the changes immediately.

Environment variables: The documentation provided by Microsoft explains how to set these configuration options via environment variables. Jellyfin uses its own custom JELLYFIN_ prefix for these variables. For example, to set a value for the HttpListenerHost:DefaultRedirectPath setting, you would set a value for the JELLYFIN_HttpListenerHost__DefaultRedirectPath environment variable.

Command line options: Certain command line options are loaded into the configuration system and have the highest priority. The following command line options are mapped to associated configuration options.

--nowebcontent sets the hostwebclient configuration setting to false
--plugin-manifest-url sets a value for the InstallationManager:PluginManifestUrl configuration setting
Main Configuration Options
This section lists all the configuration options available and explains their function.
```
Key | Default Value | Description
-- | -- | --
[...]

PR #8280, opened 2022-08-17
Fix ffmpeg analyze duration env var taking priority over media source

thornbill:
**Changes**
This changes the order that analyze duration values are applied so media source values take precedence over the ffmpeg environment variable/setting.

**Issues**
Fixes the long delay when starting a live tv stream by honoring the analyze duration set at the media source level that has been ignored since 10.8.1.

Related discussion: https://github.com/jellyfin/jellyfin/issues/3930#issuecomment-1166172969

Related PR: https://github.com/jellyfin/jellyfin/pull/7828

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -4976,13 +4976,13 @@ namespace MediaBrowser.Controller.MediaEncoding
             // The default value of -probesize is more than enough, so leave it as is.
             var ffmpegAnalyzeDuration = _config.GetFFmpegAnalyzeDuration() ?? string.Empty;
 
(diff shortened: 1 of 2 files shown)


Bond-009 (approved):

crobibero (approved):

cvium (approved):

PrplHaz4:
doesn't this effectively make the environment variable useless?  it seems like there should be env variables per mediasource type to allow users to set an override explicitly where needed

thornbill:
I believe the only media source that sets the analyze duration currently is Live TV. The previous change would always use the ffmpeg setting since it had a default value which lead to a pretty poor experience for Live TV. Maybe we should have a setting per media source type, or just a way to override the default for Live TV, but I honestly don't know if there is a need for that. 🤷‍♂️

PR #8378, opened 2022-09-09
Don't allow throwing System.Exception

Bond-009:


Emby.Server.Implementations/LiveTv/Listings/SchedulesDirect.cs
```diff
@@ -662,7 +660,7 @@ private async Task<string> GetToken(ListingsProviderInfo info, CancellationToken
                 return root.Token;
             }
 
-            throw new Exception("Could not authenticate with Schedules Direct Error: " + root.Message);
+            throw new AuthenticationException("Could not authenticate with Schedules Direct Error: " + root.Message);
```
Bond-009, line 663: Not 100% sure this is the right exception to throw here

crobibero (approved):

1337joe (approved):

PR #8381, opened 2022-09-10
Add Lyrics API Endpoint

1hitsong:
**Changes**
Adds an API endpoint to get lyrics for a requested item. The lyrics are sourced from local .LRC or .TXT files and returned as JSON for client consumption.

Lyric files must match the filenames for their corresponding item.

For example: The lyric file for `01 Death Eternal.mp3` must be `01 Death Eternal.lrc` or `01 Death Eternal.txt`

**Sample Response**
```{
    "Results": [
        {
            "Start": 670200000,
            "Text": "'Cause it's a bittersweet symphony that's life."
        },
        {
            "Start": 780399999,
            "Text": "Trying to make ends meet,"
        },
        {
            "Start": 806700000,
            "Text": "you're a slave to money"
        }
    ]
}
```


Jellyfin.Api/Controllers/UserLibraryController.cs
```diff
@@ -381,5 +385,50 @@ private UserItemDataDto UpdateUserItemRatingInternal(Guid userId, Guid itemId, b
...
+        {
+            var user = _userManager.GetUserById(userId);
+
+            if (user == null)
+            {
+                List<Lyrics> lyricsList = new List<Lyrics>
+                {
+                    new Lyrics { Error = "User Not Found" }
+                };
+                return NotFound(new { Results = lyricsList.ToArray() });
```
crobibero, line 409:
```suggestion
                return NotFound();
```

Jellyfin.Api/Controllers/UserLibraryController.cs
```diff
@@ -381,5 +385,50 @@ private UserItemDataDto UpdateUserItemRatingInternal(Guid userId, Guid itemId, b
...
+                ? _libraryManager.GetUserRootFolder()
+                : _libraryManager.GetItemById(itemId);
+
+            if (item == null)
+            {
+                List<Lyrics> lyricsList = new List<Lyrics>
+                {
+                    new Lyrics { Error = "Requested Item Not Found" }
+                };
+                return NotFound(new { Results = lyricsList.ToArray() });
```
crobibero, line 422:
```suggestion
                return NotFound(new { Results = lyricsList.ToArray() });
```

Jellyfin.Api/Controllers/UserLibraryController.cs
```diff
@@ -381,5 +385,50 @@ private UserItemDataDto UpdateUserItemRatingInternal(Guid userId, Guid itemId, b
...
+            }
+
+            List<Lyrics> result = ItemHelper.GetLyricData(item);
+            if (string.IsNullOrEmpty(result.ElementAt(0).Error))
+            {
+                return Ok(new { Results = result });
+            }
+
+            return NotFound(new { Results = result.ToArray() });
```
crobibero, line 431:
```suggestion
            return NotFound();
```

crobibero (changes requested):

Jellyfin.Api/Controllers/UserLibraryController.cs
1hitsong: Did you mean for this one to be ```return NotFound();``` ?

crobibero: I did, sorry!

Jellyfin.Api/Controllers/UserLibraryController.cs
```diff
@@ -1,14 +1,18 @@
...
 using System.Threading;
 using System.Threading.Tasks;
 using Jellyfin.Api.Constants;
 using Jellyfin.Api.Extensions;
+using Jellyfin.Api.Helpers;
 using Jellyfin.Api.ModelBinders;
+using Jellyfin.Api.Models.UserDtos;
 using Jellyfin.Data.Enums;
 using Jellyfin.Extensions;
+using Kfstorm.LrcParser;
```
dkanada, line 15: Is there a specific reason this is precompiled?

Jellyfin.Api/Controllers/UserLibraryController.cs
```diff
@@ -381,5 +385,42 @@ private UserItemDataDto UpdateUserItemRatingInternal(Guid userId, Guid itemId, b
...
+            if (item == null)
+            {
+                return NotFound();
+            }
+
+            List<Lyrics> result = ItemHelper.GetLyricData(item);
+            if (string.IsNullOrEmpty(result.ElementAt(0).Error))
+            {
+                return Ok(new { Results = result });
+            }
```
dkanada, line 421: At minimum I would suggest an ILyricProvider and a lambda here that searches through existing implementations to find the first result. Bonus points if there's a setting somewhere to select a particular lyric source, but as long as the provider exists that can always be added later.

Jellyfin.Api/Controllers/UserLibraryController.cs
1hitsong: The one straight from NuGet targets the wrong TargetFramework. I recompiled it with a modified TargetFramework value of `net6.0`

crobibero: I thought I left a comment- we straight up cannot use this library as it is unlicensed. We also dont directly include dependencies.

Jellyfin.Api/Models/UserDtos/Lyrics.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// </summary>
+    public class Lyrics
+    {
+        /// <summary>
+        /// Gets or sets the start.
+        /// </summary>
+        public double? Start { get; set; }
+
+        /// <summary>
+        /// Gets or sets the test.
```
neilsb, line 14:
```suggestion
        /// Gets or sets the text.
```

nielsvanvelzen:
I've looked at the API part only thinking how I would implement a lyrics feature in our Android apps. These are my findings:

- Let the client know if there are any lyrics without calling the lyrics endpoint
 Just like with videos we know which video/audio/sub tracks are available before we start playback

- Use proper nullability in API
  I'd imagine the start & text properties should never be null

- Error for each line?
  If parsing the lyrics fails we should not respond instead of sending errors for each separate line

- Document properties
  Are the lines always ordered by timestamp?
  Is the "start" in ticks/milliseconds/seconds?
  Can text be empty? Can it contains line breaks? HTML? Markdown?

- ID tags
  According to Wikipedia lyric files may contain some metadata which we should also add to the response. I know a lot of my foobar_openlyrics files have this metadata.
  https://en.wikipedia.org/wiki/LRC_(file_format)

- What happens to my lyric files without timestamps?
  Is everything a single "line" with start set to 0? Are the separate lines all with timestamp 0?

- Enhanced format
  Some lrc files have "word time tags", how is that handled right now? (see Wikipedia page for example)

I think an API response might end up looking somewhat like this:

```json5
{
  "metadata": {
    "//": "Either a Dictionary<string,string> or a proper object with \"known\" id3 tags like Artist/Album etc",
    "artist": "Artist name",
    "title": "Track name",
  },
  "lines": [
    { "start": 0, "text": "Line 1" },
    { "start": 100, "text": "Line 2" },
    { "start": 300, "text": "Line 2" },
  ]
}
```

Some of the above mentioned stuff might be out of scope for this PR but these are my thoughts for now.

1hitsong:
Thanks for all the helpful comments y'all! I'll start working through these.

Jellyfin.Api/Controllers/UserLibraryController.cs
1hitsong: I've found a different NuGet package to use for the LRC parser that has an MIT license.

Jellyfin.Api/Controllers/UserLibraryController.cs
1hitsong: You'll have to forgive me, C# isn't my primary language. I think I'm getting what you're saying here, but if I have any followup questions, do you mind if I reach out for clarification?

1hitsong:
> Let the client know if there are any lyrics without calling the lyrics endpoint

Would this be satisfied by adding a HEAD endpoint that indicates whether or not lyrics are found?

> Error for each line?

It'll be all or nothing. Either you get something back, or you get a 404 response.

> Are the lines always ordered by timestamp?

Yes, they will be ordered in ascending order.

>Is the "start" in ticks/milliseconds/seconds?

They will be ticks

> Can text be empty? Can it contains line breaks? HTML? Markdown?

Yes across the board, but it will be encoded. Perhaps we could have an optional strip parameter to remove unwanted values (eg, strip HTML, string line breaks, etc.)

Example:
```
{
    "Results": [
        {
            "Start": 0,
            "Text": " <00:05.02>Some <00:07.23>lyrics"
        },
        {
            "Start": 75900000,
            "Text": ""
        },
        {
            "Start": 79400000,
            "Text": " Sing this song ##YELL!##"
        },
        {
            "Start": 150100000,
            "Text": " <strong>Yeah\\n\\r!</strong>"
        }
    ]
}
```

> ID tags

I'll look into what it'll take to add ID data to the response.

> What happens to my lyric files without timestamps?

What file format are they? Are the TXT or are they still LRC files?

> Enhanced format

The line level timestamp will be used as the Start value in the JSON, and the word times will be passed as part of the Text value. This way clients are free to do with them as they please (use them, ignore them, etc.).

Example: 
```
{
    "Start": 0,
    "Text": " <00:05.02>Some <00:07.23>lyrics"
},
```

nielsvanvelzen:
> Would this be satisfied by adding a HEAD endpoint that indicates whether or not lyrics are found?

I was thinking adding it to the BaseItemDto in some way, there is a mediasources property where each source has mediastreams. lyrics could maybe be one of them, idk.

> It'll be all or nothing. Either you get something back, or you get a 404 response.

So why is there an error property in the responses then?

> What file format are they? Are the TXT or are they still LRC files?

lrc files but without timestamps at the start of each line. The Foobar2000 plugin UI slowly scrolls those lyrics.

1hitsong:
> I was thinking adding it to the BaseItemDto in some way, there is a mediasources property where each source has mediastreams. lyrics could maybe be one of them, idk.

I'll take a look and see what I can come up with.

> So why is there an error property in the responses then?

Originally I was returning an error description along with the NotFound(), but after code suggestions from Cody, it's no longer doing that. I'm updating the responses now and the error property will be removed.

> lrc files but without timestamps at the start of each line. The Foobar2000 plugin UI slowly scrolls those lyrics.

Interesting. So essentially a TXT file in an LRC container. I'll take a look at it.

ferferga:
Thank you very much for this PR, been wanting to implement proper lyric support in Vue!

Two points though:

* Why not also add compatibility for lyrics included in IDv3/FLAC (etc) tags? They are fetched by ffprobe during scan (and, in fact, they appear in JF logs), but they're not saved anywhere. Those are usually not synced, but at least we have something

* I very much prefer @nielsvanvelzen proposed schema, as adding HTML tags to a response that will be served to multiple clients (not all of them web-based...) seems counterintuitive. I'd say just timestamps and strings should be present, all the other formatting should be handled by the client (which in the end is the one that knows how to show the info to the user), not server.

1hitsong:
> Why not also add compatibility for lyrics included in IDv3/FLAC (etc) tags? They are fetched by ffprobe during scan (and, in fact, they appear in JF logs), but they're not saved anywhere. Those are usually not synced, but at least we have something

I'd like to keep the scope of this PR focused on local file support, but that sounds like a wonderful followup PR.

> I very much prefer @nielsvanvelzen proposed schema...

I agree, so that's what I'm basing it on in the updated code. The server makes no assumptions about the data other than parsing meta and timestamps if present. Anything else in the text will get returned and clients can do what they want with it.

Jellyfin.Api/Controllers/UserLibraryController.cs
dkanada: No problem, you can find me on Matrix. For the record, the IDv3 lyric support mentioned by @ferferga is a great example of code that would go in a separate LyricProvider down the road. People could even implement new providers through plugins if they want support for a paid service or something.

1hitsong:
I think I covered everything mentioned above.

Below is some sample output.

**LRC File with MetaData**
```
{
    "MetaData": {
        "ar": " The Verve",
        "ti": " Bitter Sweet Symphony",
        "al": " This Is Music",
        "by": " lcf1002",
        "id": " da_qblwhlsa"
    },
    "lyrics": [
        {
            "Start": 7800000,
            "Text": ""
        },
        {
            "Start": 670200000,
            "Text": "'Cause it's a bittersweet symphony that's life."
        },
        {
            "Start": 753700000,
            "Text": ""
        },
        {
            "Start": 780400000,
            "Text": "Trying to make ends meet,"
        },
        {
            "Start": 806700000,
            "Text": "you're a slave to money"
        },
```
**LRC File without Metadata or TimeStamps**
```
{
    "lyrics": [
        {
            "Text": "this"
        },
        {
            "Text": "is"
        },
        {
            "Text": "some"
        },
        {
            "Text": "lyrics"
        }
    ]
}
```

**TXT File**
```
{
    "lyrics": [
        {
            "Text": "This is a lyric"
        },
        {
            "Text": "so is this"
        }
    ]
}
```

1hitsong:
Additionally, I added the field HasLocalLyricsFile the get item API call. If a local lyric file is found and has a corresponding provider, this value will be true.
```
{
    "Name": "02 Voices of the Dead",
    "ServerId": "466d44350b294c0c888a93d77c3d5684",
    "Id": "f4c0097608894ff3bebaf802e596e03f",
    "Etag": "56beecb1e99461374e9d88a843c6a05f",
    "DateCreated": "2022-09-08T23:17:45.4225411Z",
    "CanDelete": true,
    "CanDownload": true,
    "HasLocalLyricsFile": true,
    ...
```

MediaBrowser.Model/Dto/BaseItemDto.cs
```diff
@@ -76,6 +76,8 @@ public class BaseItemDto : IHasProviderIds, IItemDto, IHasServerId
 
         public bool? CanDownload { get; set; }
 
+        public bool? HasLocalLyricsFile { get; set; }
```
crobibero, line 79: I would just use `HasLyrics`

Jellyfin.Api/Helpers/ItemHelper.cs
```diff
@@ -0,0 +1,106 @@
...
+                .Where(type => typeof(ILyricsProvider).IsAssignableFrom(type) && !type.IsInterface);
+
+            if (!foundLyricProviders.Any())
+            {
+                return null;
+            }
+
+            foreach (var provider in foundLyricProviders)
+            {
+                ILyricsProvider? newProvider = Activator.CreateInstance(provider) as ILyricsProvider;
```
crobibero, line 38: `ILyricsProvider` should be added to the DI pipeline.

Jellyfin.Api/Models/UserDtos/ILyricsProvider.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// Gets a value indicating whether Process() generated data.
+        /// </summary>
+        /// <returns><c>true</c> if data generated; otherwise, <c>false</c>.</returns>
+        bool HasData { get; }
+
+        /// <summary>
+        /// Gets Data object generated by Process() method.
+        /// </summary>
+        /// <returns><c>Object</c> with data if no error occured; otherwise, <c>null</c>.</returns>
+        object? Data { get; }
```
crobibero, line 26: This needs to be an actual typed object.

Jellyfin.Api/Models/UserDtos/ILyricsProvider.cs
```diff
@@ -0,0 +1,34 @@
+using System.Collections.ObjectModel;
+using MediaBrowser.Controller.Entities;
+
+namespace Jellyfin.Api.Models.UserDtos
+{
+    /// <summary>
+    /// Interface ILyricsProvider.
+    /// </summary>
+    public interface ILyricsProvider
```
crobibero, line 9: I suggest this interface follows the existing `ISubtitleProvider` (Doesn't need `Search`)

put it in `MediaBrowser.Controller.Lyrics`

Jellyfin.Api/Models/UserDtos/LrcLyricsProvider.cs
```diff
@@ -0,0 +1,117 @@
...
+using LrcParser.Model;
+using LrcParser.Parser;
+using MediaBrowser.Controller.Entities;
+
+namespace Jellyfin.Api.Models.UserDtos
+{
+    /// <summary>
+    /// LRC File Lyric Provider.
+    /// </summary>
+    public class LrcLyricsProvider : ILyricsProvider
```
crobibero, line 16: `LrcLyricsProvider` and `TxtLyricsProvider` should be under `MediaBrowser.Providers.Lyric`

Jellyfin.Api/Models/UserDtos/TxtLyricsProvider.cs
```diff
@@ -0,0 +1,81 @@
...
+            string? lyricFilePath = Helpers.ItemHelper.GetLyricFilePath(item.Path);
+
+            if (string.IsNullOrEmpty(lyricFilePath))
+            {
+                return;
+            }
+
+            List<Lyric> lyricsList = new List<Lyric>();
+
+            string lyricData = System.IO.File.ReadAllText(lyricFilePath);
```
dkanada, line 61: Once you hook it up to the DI pipeline plugins will be possible so this looks great to me. My last comment is that there seems to be a mix of `lyric` and `lyrics` for variables and other references - we should pick one now and stick to it.

dkanada: Since `ISubtitleProvider` is already singular we should probably do that here, but as long as it's consistent everywhere I could go either way.

Jellyfin.Api/Helpers/ItemHelper.cs
```diff
@@ -0,0 +1,106 @@
...
+                if (newProvider is not null)
+                {
+                    newProvider.Process(item);
+                    if (newProvider.HasData)
```
cvium, line 42: Seems pointless to keep state in the provider. Why not just return the result (or null) directly?

Jellyfin.Api/Helpers/ItemHelper.cs
```diff
@@ -0,0 +1,106 @@
...
+    {
+        /// <summary>
+        /// Opens lyrics file, converts to a List of Lyrics, and returns it.
+        /// </summary>
+        /// <param name="item">Requested Item.</param>
+        /// <returns>Collection of Lyrics.</returns>
+        internal static object? GetLyricData(BaseItem item)
+        {
+            // Find all classes that implement ILyricsProvider Interface
+            var foundLyricProviders = System.Reflection.Assembly.GetExecutingAssembly()
```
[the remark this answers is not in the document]
1hitsong: Yeah. Based on Cody's comment above, I'm working to add this to the DI pipeline.

Jellyfin.Api/Helpers/ItemHelper.cs
1hitsong: I've been poking around the DI pipeline, but have yet to really grasp the flow. Do we have any documentation, or an example PR, that I can look at and study?

Honestly, right now I don't even know where to begin.

crobibero: I would add `ILyricManager` and follow the `ISubtitleManager` flow as subtitles can be very similar to lyrics. I know this is getting much larger then the initial implementation but we would like to make the provider easily extendible by plugins

sample: Emby.Server.Implementations/ApplicationHost.cs L787

1hitsong: Thanks man. I called in a friend who is a full time C# programmer to help me show me how to introduce ILyricsProvider into the DI pipeline. I have it working again, but we're adding the 2 providers individually to serviceCollection.

I suspect using an ILyricManager will allow only a single entry to serviceCollection to be used.

MediaBrowser.Providers/Lyric/LyricManager.cs
```diff
@@ -0,0 +1,97 @@
...
+        {
+            _logger = logger;
+            _fileSystem = fileSystem;
+            _monitor = monitor;
+            _mediaSourceManager = mediaSourceManager;
+            _localization = localizationManager;
+        }
+
+        /// <inheritdoc />
+        public void AddParts(IEnumerable<ILyricProvider> lyricProviders)
```
cvium, line 56: I would honestly prefer not having AddParts

1hitsong: What's the alternative? Can you point me to an example that doesn't use AddParts?

cvium: We don't have any examples I think. But here's one:

https://github.com/jellyfin/jellyfin/blob/104e36f2f9c6440a7547a4c76d80a69d5af84eea/Jellyfin.Server/CoreAppHost.cs#L89-L92 could be rewritten to 

```
            foreach (var type in GetExportTypes<IWebSocketListener>())
            {
                serviceCollection.AddSingleton(typeof(IWebSocketListener), type);
            }

Comment (1hitsong) [reply MediaBrowser.Providers/Lyric/LyricManager.cs:56]: 😆  Oh, so like the comment above that section says 

>"// TODO search the assemblies instead of adding them manually?"

I'd prefer to not change that logic as post of this PR since it's not directly part of adding Lyric functionality. It sounds like a good followup for a cleanup PR though.

Comment (cvium) [reply MediaBrowser.Providers/Lyric/LyricManager.cs:56]: Is it not directly part of it? Instead of using an `AddParts` method, you can scan the assemblies for `ILyricProvider` and add them to the servicecollection. Then put `IEnumerable<ILyricProvider>` in the LyricManager's constructor.

Comment (1hitsong) [reply MediaBrowser.Providers/Lyric/LyricManager.cs:56]: Oh, I misunderstood. I thought you were asking me to rework all the serviceCollection.AddSingleton calls as you showed. Which would be touching code unrelated to Lyrics.

After my morning coffee, I now understand you were creating an example for me. 🤦

File: Jellyfin.Api/Models/UserDtos/TxtLyricsProvider.cs (hunk 12 above)
Comment (1hitsong) [reply Jellyfin.Api/Models/UserDtos/TxtLyricsProvider.cs]: I've gone through all the references and updated filename, variables, references, documentation, etc.

You have lyric providers. Lyric files. A single line from a song is a Lyric.

Songs have Lyrics. You get Lyrics for a song.

**Comment by 1hitsong** [comment]
I believe I've covered everything asked up to this point.

File: MediaBrowser.Providers/Lyric/LyricManager.cs [hunk 16]
```diff
@@ -0,0 +1,91 @@
...
+{
+    public class LyricManager : ILyricManager
+    {
+        private readonly ILogger<LyricManager> _logger;
+        private readonly IFileSystem _fileSystem;
+        private readonly ILibraryMonitor _monitor;
+        private readonly IMediaSourceManager _mediaSourceManager;
+        private readonly ILocalizationManager _localization;
+
+        private IEnumerable<ILyricProvider> _lyricProviders;
```
cvium, line 39: You should convert it to an array

MediaBrowser.Controller/Lyrics/LyricResponse.cs
```diff
@@ -0,0 +1,24 @@
+#nullable disable
+
+#pragma warning disable CS1591
```
dkanada, line 3: seems like this line is no longer required

MediaBrowser.Controller/Lyrics/Lyric.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    /// <summary>
+    /// Lyric model.
+    /// </summary>
+    public class Lyric
+    {
+        /// <summary>
+        /// Gets or sets the start time in ticks.
+        /// </summary>
+        public double? Start { get; set; }
```
nielsvanvelzen, line 11: If it's in ticks I don't think we need a double, a long might make more sense.

MediaBrowser.Controller/Lyrics/LyricResponse.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    /// <summary>
+    /// LyricResponse model.
+    /// </summary>
+    public class LyricResponse
+    {
+        /// <summary>
+        /// Gets or sets MetaData.
+        /// </summary>
+        public IDictionary<string, object> MetaData { get; set; }
```
nielsvanvelzen, line 15:
```suggestion
        public IDictionary<string, object> Metadata { get; set; }
```
For consistency with other models

nielsvanvelzen, line 15: I'm not sure about using an object here. Probably better to always use strings.

MediaBrowser.Providers/Lyric/TxtLyricProvider.cs
```diff
@@ -0,0 +1,73 @@
...
+            string? lyricFilePath = LyricInfo.GetLyricFilePath(this, item.Path);
+
+            if (string.IsNullOrEmpty(lyricFilePath))
+            {
+                return null;
+            }
+
+            List<Controller.Lyrics.Lyric> lyricList = new List<Controller.Lyrics.Lyric>();
+
+            string lyricData = System.IO.File.ReadAllText(lyricFilePath);
```
crobibero, line 54: Does `File.ReadAllLines()` work in place of manual line reading?

crobibero (commented):
Note that this will add a decent amount of IO per api call, so follow up pr should add this to the scanner + database

MediaBrowser.Controller/Lyrics/LyricResponse.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    /// <summary>
+    /// LyricResponse model.
+    /// </summary>
+    public class LyricResponse
+    {
+        /// <summary>
+        /// Gets or sets Metadata.
+        /// </summary>
+        public IDictionary<string, string> Metadata { get; set; }
```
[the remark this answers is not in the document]
1hitsong: Looking at wikipedia and several lyric sites, there appear to be 9 base metadata fields that people use. Some do use other fields, but they appear to be the exception rather than the rule.

MediaBrowser.Controller/MediaBrowser.Controller.csproj
```diff
@@ -18,6 +18,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="AutoMapper" Version="11.0.1" />
```
cvium, line 21: We don't use automapper anywhere else. You should map it manually.

MediaBrowser.Controller/Lyrics/Metadata.cs
```diff
@@ -0,0 +1,54 @@
...
+
+/// <summary>
+/// Metadata model.
+/// </summary>
+public class Metadata
+{
+    /// <summary>
+    /// Gets or sets Artist - [ar:The song artist].
+    /// </summary>
+    public string? Ar { get; set; }
```
cvium, line 13: Can these not have more descriptive names?

dkanada: Possibly not unless the mapper can use different variable names similar to XML or JSON parsers.

cvium: We shouldn't use a mapper and yes it can

MediaBrowser.Controller/Lyrics/Metadata.cs
```diff
@@ -0,0 +1,54 @@
+using System.Collections.Generic;
+
+namespace MediaBrowser.Controller.Lyrics;
+
+/// <summary>
+/// Metadata model.
+/// </summary>
+public class Metadata
```
nielsvanvelzen, line 8:
```suggestion
public class LyricMetadata
```

Use unique and specific names for API models.

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
```diff
@@ -0,0 +1,168 @@
...
+    public string Name { get; } = "LrcLyricProvider";
+
+    /// <inheritdoc />
+    public IEnumerable<string> SupportedMediaTypes
+    {
+        get => new Collection<string>
+            {
+                "lrc"
+            };
+    }
```
crobibero, line 28:
```suggestion
    public IEnumerable<string> SupportedMediaTypes { get; } = new[] { "lrc" };
```

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
```diff
@@ -0,0 +1,168 @@
...
+
+            foreach (string metaDataRow in metaDataRows)
+            {
+                var metaDataField = metaDataRow.Split(':');
+                if (metaDataField.Length != 2)
+                {
+                    continue;
+                }
+
+                string metaDataFieldName = metaDataField[0][1..].Trim().ToLowerInvariant();
```
crobibero, line 72:
```suggestion
                string metaDataFieldName = metaDataField[0][1..].Trim();
```

Just add `StringComparer.OrdinalIgnoreCase` to the dictionary constructor

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
```diff
@@ -0,0 +1,168 @@
...
+                string metaDataFieldValue = metaDataField[1][..^1].Trim();
+
+                fileMetaData.Add(metaDataFieldName, metaDataFieldValue);
+            }
+        }
+        catch
+        {
+            return null;
```
crobibero, line 80: Would be nice to log so we know why the exception was thrown instead of just silently returning

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
```diff
@@ -0,0 +1,168 @@
...
+    /// <summary>
+    /// Converts metadata from an LRC file to LyricMetadata properties.
+    /// </summary>
+    /// <param name="metaData">The metadata from the LRC file.</param>
+    /// <returns>A lyricMetadata object with mapped property data.</returns>
+    private LyricMetadata MapMetadataValues(IDictionary<string, string> metaData)
+    {
+        LyricMetadata lyricMetadata = new LyricMetadata();
+
+        if (metaData.TryGetValue("ar", out var artist) && artist is not null)
```
crobibero, line 120:
```suggestion
        if (metaData.TryGetValue("ar", out var artist) && !string.IsNullOrEmpty(artist))
```

apply to all

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
```diff
@@ -0,0 +1,168 @@
...
+            lyricMetadata.Creator = creator;
+        }
+
+        if (metaData.TryGetValue("ve", out var version) && version is not null)
+        {
+            lyricMetadata.Version = version;
+        }
+
+        return lyricMetadata;
+
```
crobibero, line 166:
```suggestion
```

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
```diff
@@ -0,0 +1,168 @@
...
+        for (int i = 0; i < sortedLyricData.Count; i++)
+        {
+            var timeData = sortedLyricData[i].TimeTags.First().Value;
+            if (timeData is null)
+            {
+                continue;
+            }
+
+            long ticks = TimeSpan.FromMilliseconds((double)timeData).Ticks;
```
crobibero, line 96:
```suggestion
            long ticks = TimeSpan.FromMilliseconds(timeData.Value).Ticks;
```

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
```diff
@@ -0,0 +1,168 @@
...
+            lyricList.Add(new Controller.Lyrics.Lyric(sortedLyricData[i].Text, ticks));
+        }
+
+        if (fileMetaData.Any())
```
crobibero, line 100:
```suggestion
        if (fileMetaData.Count != 0)
```

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
```diff
@@ -0,0 +1,168 @@
...
+
+namespace MediaBrowser.Providers.Lyric;
+
+/// <summary>
+/// LRC Lyric Provider.
+/// </summary>
+public class LrcLyricProvider : ILyricProvider
+{
+    /// <inheritdoc />
+    public string Name { get; } = "LrcLyricProvider";
```
crobibero, line 19:
```suggestion
    public string Name => "LrcLyricProvider";
```

MediaBrowser.Providers/Lyric/TxtLyricProvider.cs
```diff
@@ -0,0 +1,57 @@
...
+
+namespace MediaBrowser.Providers.Lyric;
+
+/// <summary>
+/// TXT Lyric Provider.
+/// </summary>
+public class TxtLyricProvider : ILyricProvider
+{
+    /// <inheritdoc />
+    public string Name { get; } = "TxtLyricProvider";
```
crobibero, line 16:
```suggestion
    public string Name => "TxtLyricProvider";
```

MediaBrowser.Providers/Lyric/TxtLyricProvider.cs
```diff
@@ -0,0 +1,57 @@
...
+
+    /// <inheritdoc />
+    public IEnumerable<string> SupportedMediaTypes
+    {
+        get => new Collection<string>
+            {
+                "lrc", "txt"
+            };
+    }
```
crobibero, line 25:
```suggestion
    public IEnumerable<string> SupportedMediaTypes { get; } = new[] { "lrc", "txt" };
```

crobibero (commented):
We might also want to add a `Priority` to the `ILyricProvider` so we can ensure that LrcLyricProvider is used before TxtLyricProvider, right now I think it's sorted alphabetically

MediaBrowser.Controller/Lyrics/LyricResponse.cs
```diff
@@ -0,0 +1,19 @@
...
+
+/// <summary>
+/// LyricResponse model.
+/// </summary>
+public class LyricResponse
+{
+    /// <summary>
+    /// Gets or sets Metadata.
+    /// </summary>
+    public LyricMetadata? Metadata { get; set; }
```
nielsvanvelzen, line 13:
```suggestion
    public LyricMetadata Metadata { get; set; }
```

All properties of LyricMetadata are nullable so this one doesn't need to be nullable.

MediaBrowser.Controller/Lyrics/LyricResponse.cs
```diff
@@ -0,0 +1,19 @@
...
+
+    /// <summary>
+    /// Gets or sets Lyrics.
+    /// </summary>
+    public IEnumerable<Lyric>? Lyrics { get; set; }
```
nielsvanvelzen, line 18:
```suggestion
    public IEnumerable<Lyric> Lyrics { get; set; }
```
Should never be null because that would mean no lyrics

MediaBrowser.Controller/Lyrics/Lyric.cs
```diff
@@ -0,0 +1,28 @@
+namespace MediaBrowser.Controller.Lyrics;
+
+/// <summary>
+/// Lyric model.
+/// </summary>
+public class Lyric
```
nielsvanvelzen, line 6: I think calling this "LyricLine" might make more sense because it's the information for a single line in a lyrics file.

MediaBrowser.Controller/Lyrics/LyricMetadata.cs
```diff
@@ -0,0 +1,52 @@
...
+
+    /// <summary>
+    /// Gets or sets By - Creator of the LRC file.
+    /// </summary>
+    public string? By { get; set; }
+
+    /// <summary>
+    /// Gets or sets Offset - Offset:+/- Timestamp adjustment in milliseconds.
+    /// </summary>
+    public string? Offset { get; set; }
```
nielsvanvelzen, line 41: Since we're adding the properties explicitly now instead of the Dictionary<string,string> from before we can set the types for each property. Offset and Length should be long/int/numerical.

nielsvanvelzen (changes requested):

MediaBrowser.Controller/Lyrics/LyricResponse.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    /// <summary>
+    /// Gets or sets Metadata.
+    /// </summary>
+    public LyricMetadata Metadata { get; set; } = new LyricMetadata();
+
+    /// <summary>
+    /// Gets or sets Lyrics.
+    /// </summary>
+    public IReadOnlyCollection<LyricLine> Lyrics { get; set; } = new List<LyricLine>();
```
cvium, line 18: You're better off with `IReadOnlyList` as it lets you use `Array.Empty<LyricLine>()`

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
```diff
@@ -0,0 +1,213 @@
...
+    public LyricResponse? GetLyrics(BaseItem item)
+    {
+        string? lyricFilePath = LyricInfo.GetLyricFilePath(this, item.Path);
+
+        if (string.IsNullOrEmpty(lyricFilePath))
+        {
+            return null;
+        }
+
+        List<LyricLine> lyricList = new List<LyricLine>();
```
cvium, line 57: It's pretty redundant having the type on both sides of `=`. Either use `var` or `new()`

1.         var lyricList = new List<LyricLine>();
2.         List<LyricLine> lyricList = new();

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
```diff
@@ -0,0 +1,213 @@
...
+                    metaDataField = metaDataRow.Split(':');
+                    metaDataFieldName = metaDataField[0][1..].Trim();
+                    metaDataFieldValue = metaDataField[1][..^1].Trim();
+                }
+                else
+                {
+                    int colonIndex = metaDataRow.IndexOf(':', StringComparison.OrdinalIgnoreCase);
+                    metaDataFieldName = metaDataRow[..colonIndex][1..].Trim();
+                    metaDataFieldValue = metaDataRow[(colonIndex + 1)..][..^1].Trim();
+                }
```
cvium, line 100: I don't really understand this bit. Is it equivalent to `metaDataRow.Split(':', RemoveEmptyEntries | TrimEntries)`?

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
```diff
@@ -0,0 +1,213 @@
...
+        {
+            // Ensure minutes include leading zero
+            var lengthData = length.Split(':');
+            if (lengthData[0].Length == 1)
+            {
+                length = "0" + length;
+            }
+
+            // If only Minutes and Seconds were provided, prepend zeros for hours
+            if (lengthData.Length == 2)
```
cvium, line 177: Is all this necessary? There's an overload for `DateTime.TryParseExact` that takes in a list of possible formats.

MediaBrowser.Providers/Lyric/TxtLyricProvider.cs
```diff
@@ -0,0 +1,55 @@
...
+        List<LyricLine> lyricList = new List<LyricLine>();
+
+        if (lyricTextLines.Length == 0)
+        {
+            return null;
+        }
+
+        foreach (string lyricTextLine in lyricTextLines)
+        {
+            lyricList.Add(new LyricLine(lyricTextLine));
```
cvium, line 50: This is a great example of `IReadOnlyList` being the superior choice. This won't be necessary as you can just assign it to the `Lyrics` property directly. (But pretending that it had to be a list, a simple `.ToList` would be better than adding to a new list).

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
```diff
@@ -0,0 +1,213 @@
...
+    {
+        string? lyricFilePath = LyricInfo.GetLyricFilePath(this, item.Path);
+
+        if (string.IsNullOrEmpty(lyricFilePath))
+        {
+            return null;
+        }
+
+        List<LyricLine> lyricList = new List<LyricLine>();
+        List<LrcParser.Model.Lyric> sortedLyricData = new List<LrcParser.Model.Lyric>();
```
cvium, line 58: You're not using the list you're creating here.

MediaBrowser.Controller/Lyrics/LyricLine.cs
```diff
@@ -0,0 +1,28 @@
...
+/// <summary>
+/// Lyric model.
+/// </summary>
+public class LyricLine
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="LyricLine"/> class.
+    /// </summary>
+    /// <param name="start">The lyric start time in ticks.</param>
+    /// <param name="text">The lyric text.</param>
```
cvium, line 12: The order should be reversed

MediaBrowser.Providers/Lyric/TxtLyricProvider.cs
1hitsong: "This won't be necessary as you can just assign it to the Lyrics property directly."

This is my C# ignorance talking, but how can you directly assign it to `Lyrics` property when `lyricTextLines` is `string[]` but the `Lyrics` property  is wanting `IReadOnlyList<LyricLine>`?

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
1hitsong: I'm confused on this one. 

I use lyricList on 124, 132, 135.

I use sortedLyricData on 68, 110, 115, 124.

MediaBrowser.Providers/Lyric/TxtLyricProvider.cs
cvium: An array implements `IReadOnlyList`

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
cvium: You immediately assign a new value to `sortedLyricData` on line 68. There's no need to create a list prior to this point

1hitsong: If I don't create the list outside the Try/Catch, how can I use in on 97, 102, 104, etc? VS blows up with `Does not exist in this context` if I declare sortedLyricData and assign the value together on 68.

MediaBrowser.Providers/Lyric/TxtLyricProvider.cs
1hitsong: I mean the data itself. The loop is to convert from string to `Lyricline`. How can that be avoided?

cvium: Right, I was sleepy. It can't be avoided, but you can avoid the use of .Add creating a new array under the hood by either specifying the List size or by creating an array of size `lyricTextLines,Length`

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
cvium: You can restructure the code in a way that moves the instantiation of dictionaries etc. closer to their uses. And it shouldn't be necessary with such a large try-catch. `Decode` is probably the only call that should throw.

MediaBrowser.Controller/Lyrics/LyricResponse.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    /// <summary>
+    /// Gets or sets Metadata.
+    /// </summary>
+    public LyricMetadata Metadata { get; set; } = new();
+
+    /// <summary>
+    /// Gets or sets Lyrics.
+    /// </summary>
+    public IReadOnlyList<LyricLine> Lyrics { get; set; } = new List<LyricLine>();
```
crobibero, line 18:
```suggestion
    public IReadOnlyList<LyricLine> Lyrics { get; set; } = Array.Empty<LyricLine>();
```

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
```diff
@@ -0,0 +1,189 @@
...
+        {
+            int colonCount = metaDataRow.Count(f => (f == ':'));
+            if (colonCount == 0)
+            {
+                continue;
+            }
+
+            string[] metaDataField = metaDataRow.Split(':', 2, StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries);
+            string metaDataFieldName = metaDataField[0][1..].Trim();
+            string metaDataFieldValue = metaDataField[1][..^1].Trim();
```
crobibero, line 92:
```suggestion
            string metaDataFieldName = metaDataField[0][1..];
            string metaDataFieldValue = metaDataField[1][..^1];
```

Now that trim is handled by the split

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        if (metaData.TryGetValue("au", out var author) && !string.IsNullOrEmpty(author))
+        {
+            lyricMetadata.Author = author;
+        }
+
+        if (metaData.TryGetValue("length", out var length) && !string.IsNullOrEmpty(length))
+        {
+            if (DateTime.TryParseExact(length, new string[] { "HH:mm:ss", "H:mm:ss", "mm:ss", "m:ss" }, null, DateTimeStyles.None, out var value))
```
crobibero, line 158: The formats array should be `private static readonly` (at the class level)

1hitsong: `private static readonly string[] AcceptedTimeFormats => new[] { "HH:mm:ss", "H:mm:ss", "mm:ss", "m:ss" };`

I'm getting an error 'The modifier 'readonly' is not valid for this item.

1hitsong: The code I just pushed used `private static` let me know if I need to adjust to something else of if I can resolve this.

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
```diff
@@ -0,0 +1,195 @@
...
+    public ResolverPriority Priority => ResolverPriority.First;
+
+    /// <inheritdoc />
+    public IReadOnlyCollection<string> SupportedMediaTypes { get; } = new[] { "lrc", "elrc" };
+
+    /// <summary>
+    /// Gets the Accepted Time Formats for the metadata numeric values.
+    /// </summary>
+    /// <value>The AcceptedTimeFormats.</value>
+    private static string[] AcceptedTimeFormats => new[] { "HH:mm:ss", "H:mm:ss", "mm:ss", "m:ss" };
```
cvium, line 46:
```suggestion
    private static readonly IReadOnlyList<string> _acceptedTimeFormats = new string[] { "HH:mm:ss", "H:mm:ss", "mm:ss", "m:ss" };
```

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
```diff
@@ -0,0 +1,195 @@
...
+        // Parse metadata rows
+        var metaDataRows = lyricData.Lyrics
+            .Where(x => x.TimeTags.Count == 0)
+            .Where(x => x.Text.StartsWith('[') && x.Text.EndsWith(']'))
+            .Select(x => x.Text)
+            .ToList();
+
+        foreach (string metaDataRow in metaDataRows)
+        {
+            int colonCount = metaDataRow.Count(f => (f == ':'));
```
cvium, line 90: You're not using the count, so it'd be better to do `metaDataRow.Contains(':')`

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
```diff
@@ -0,0 +1,190 @@
...
+
+        IDictionary<string, string> fileMetaData = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+        string lrcFileContent = System.IO.File.ReadAllText(lyricFilePath);
+
+        Song lyricData;
+
+        try
+        {
+            // Parse and sort lyric rows
+            LyricParser lrcLyricParser = new LrcParser.Parser.Lrc.LrcParser();
```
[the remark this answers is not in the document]
1hitsong: Would it work if I made it a field?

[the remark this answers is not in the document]
crobibero: yes, make it just like `ILogger`, but create it in the constructor instead of passing it into the constructor

crobibero (approved):

Emby.Server.Implementations/ApplicationHost.cs
```diff
@@ -46,6 +46,7 @@
 using Emby.Server.Implementations.TV;
 using Emby.Server.Implementations.Updates;
 using Jellyfin.Api.Helpers;
+using Jellyfin.Api.Models.UserDtos;
```
cvium, line 49: Is this used?

MediaBrowser.Controller/Lyrics/LyricInfo.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// </summary>
+    /// <param name="lyricProvider">The lyricProvider interface to use.</param>
+    /// <param name="itemPath">Path of requested item.</param>
+    /// <returns>Lyric file path if passed lyric provider's supported media type is found; otherwise, null.</returns>
+    public static string? GetLyricFilePath(ILyricProvider lyricProvider, string itemPath)
+    {
+        foreach (string lyricFileExtension in lyricProvider.SupportedMediaTypes)
+        {
+            var lyricFilePath = Path.ChangeExtension(itemPath, lyricFileExtension);
+            if (File.Exists(lyricFilePath))
```
cvium, line 21: That's potentially 5 IO calls for every music file. I wonder if it would be better to use [`Directory.GetFiles`](https://learn.microsoft.com/en-us/dotnet/api/system.io.directory.getfiles?view=net-6.0#system-io-directory-getfiles(system-string-system-string)) eg.

```C#
            foreach (var filename in Directory.GetFiles(Path.GetDirectoryName(lyricFilePath), $"{Path.GetFilenameWithoutExtension(lyricFilePath)}.*")
```

MediaBrowser.Controller/Lyrics/LyricInfo.cs
```diff
@@ -0,0 +1,29 @@
...
+/// </summary>
+public static class LyricInfo
+{
+    /// <summary>
+    /// Gets matching lyric file for a requested item.
+    /// </summary>
+    /// <param name="lyricProvider">The lyricProvider interface to use.</param>
+    /// <param name="itemPath">Path of requested item.</param>
+    /// <returns>Lyric file path if passed lyric provider's supported media type is found; otherwise, null.</returns>
+    public static string? GetLyricFilePath(ILyricProvider lyricProvider, string itemPath)
```
cvium, line 16: It might look nicer if it's an extension method
```suggestion
    public static string? GetLyricFilePath(this ILyricProvider lyricProvider, string itemPath)
```

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
```diff
@@ -0,0 +1,196 @@
...
+    public IReadOnlyCollection<string> SupportedMediaTypes { get; } = new[] { "lrc", "elrc" };
+
+    /// <summary>
+    /// Opens lyric file for the requested item, and processes it for API return.
+    /// </summary>
+    /// <param name="item">The item to to process.</param>
+    /// <returns>If provider can determine lyrics, returns a <see cref="LyricResponse"/> with or without metadata; otherwise, null.</returns>
+    public LyricResponse? GetLyrics(BaseItem item)
+    {
+        string? lyricFilePath = LyricInfo.GetLyricFilePath(this, item.Path);
```
cvium, line 54: If you make it an extension method, this call could be changed to this
```suggestion
        string? lyricFilePath = this.GetLyricFilePath(item.Path);
```

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
```diff
@@ -0,0 +1,196 @@
...
+
+        if (string.IsNullOrEmpty(lyricFilePath))
+        {
+            return null;
+        }
+
+        IDictionary<string, string> fileMetaData = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
```
cvium, line 61:
```suggestion
        var fileMetaData = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
```

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
```diff
@@ -0,0 +1,196 @@
...
+
+        foreach (string metaDataRow in metaDataRows)
+        {
+            if (!metaDataRow.Contains(':', StringComparison.OrdinalIgnoreCase))
+            {
+                continue;
+            }
+
+            string[] metaDataField = metaDataRow.Split(':', 2, StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries);
+            string metaDataFieldName = metaDataField[0][1..];
```
cvium, line 93: Why `[1..]`? Do you have some sample data?

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
```diff
@@ -0,0 +1,196 @@
...
+        }
+
+        if (metaData.TryGetValue("au", out var author) && !string.IsNullOrEmpty(author))
+        {
+            lyricMetadata.Author = author;
+        }
+
+        if (metaData.TryGetValue("length", out var length) && !string.IsNullOrEmpty(length))
+        {
+            if (DateTime.TryParseExact(length, _acceptedTimeFormats.ToArray(), null, DateTimeStyles.None, out var value))
```
cvium, line 165: Change the field type to `string[]` instead
```suggestion
            if (DateTime.TryParseExact(length, _acceptedTimeFormats, null, DateTimeStyles.None, out var value))
```

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
1hitsong:
```
[ti:Death Eternal]
[ar: Calabrese]
[AU: 1hitsong]
[length: 1:02:23]
[offset: 300]
[00:00.78]
[01:07.02]'Cause it's a bittersweet symphony that's life.
[01:15.37]
[01:18.04]Trying to make ends meet,
[01:20.67]you're a slave to money
[01:23.66]then you die......
```

Emby.Server.Implementations/ApplicationHost.cs
1hitsong: Nope. Removed.

MediaBrowser.Controller/Lyrics/LyricInfo.cs
1hitsong: I agree. Done.

MediaBrowser.Controller/Lyrics/LyricInfo.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        // Ensure the directory path exists
+        if (!Directory.Exists(itemDirectoryPath))
+        {
+            return null;
+        }
+
+        foreach (var lyricFilePath in Directory.GetFiles(itemDirectoryPath, $"{Path.GetFileNameWithoutExtension(itemPath)}.*"))
+        {
+            if (lyricFilePath is null) { continue; }
```
cvium, line 42: How can it be null?

MediaBrowser.Controller/Lyrics/LyricInfo.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            if (lyricProvider.SupportedMediaTypes.Contains(Path.GetExtension(lyricFilePath)[1..]))
```
cvium, line 44: You're allocating 2 strings for every possible file path. You might be able to use a span instead and avoid the allocations. There should be a Contains extension in the Jellyfin.Extensions project.
```suggestion
            if (lyricProvider.SupportedMediaTypes.Contains(Path.GetExtension(lyricFilePath.AsSpan())[1..]))
```

MediaBrowser.Controller/Lyrics/LyricInfo.cs
1hitsong: That was part of my attempt to resolve the Azure build failure due to warnings. It looks like the real fix was moving the Path.GetDirectoryName outside the foreach and testing if it's null and if the Directory Exists before the loop.

So I can remove it.

cvium (approved):

MediaBrowser.Controller/Lyrics/LyricMetadata.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// </summary>
+    public string? Title { get; set; }
+
+    /// <summary>
+    /// Gets or sets Author - Creator of the lyric data.
+    /// </summary>
+    public string? Author { get; set; }
+
+    /// <summary>
+    /// Gets or sets Length - How long the song is.
```
nielsvanvelzen, line 31: In what unit? Ticks? Milliseconds? (I suspect milliseconds looking at the offset property)

MediaBrowser.Controller/Lyrics/LyricMetadata.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// </summary>
+    public string? Album { get; set; }
+
+    /// <summary>
+    /// Gets or sets Title - The title of the song.
+    /// </summary>
+    public string? Title { get; set; }
+
+    /// <summary>
+    /// Gets or sets Author - Creator of the lyric data.
```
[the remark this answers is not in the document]
nielsvanvelzen: Is author the author of the lyrics or the author of the song?

MediaBrowser.Controller/Lyrics/LyricMetadata.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// </summary>
+    public long? Offset { get; set; }
+
+    /// <summary>
+    /// Gets or sets Creator - The Software used to create the LRC file.
+    /// </summary>
+    public string? Creator { get; set; }
+
+    /// <summary>
+    /// Gets or sets Version - The version of the Creator used.
```
nielsvanvelzen, line 51:
```suggestion
    /// Gets or sets Version - The version of the creator used.
```

MediaBrowser.Controller/Lyrics/LyricMetadata.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// </summary>
+    public string? By { get; set; }
+
+    /// <summary>
+    /// Gets or sets Offset - Offset:+/- Timestamp adjustment in milliseconds.
+    /// </summary>
+    public long? Offset { get; set; }
+
+    /// <summary>
+    /// Gets or sets Creator - The Software used to create the LRC file.
```
nielsvanvelzen, line 46:
```suggestion
    /// Gets or sets Creator - The software used to create the LRC file.
```

MediaBrowser.Controller/Lyrics/LyricMetadata.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// </summary>
+    public long? Length { get; set; }
+
+    /// <summary>
+    /// Gets or sets By - Creator of the LRC file.
+    /// </summary>
+    public string? By { get; set; }
+
+    /// <summary>
+    /// Gets or sets Offset - Offset:+/- Timestamp adjustment in milliseconds.
```
nielsvanvelzen, line 41:
```suggestion
    /// Gets or sets Offset - Lyric offset compared to audio in milliseconds.
```

MediaBrowser.Controller/Lyrics/LyricLine.cs
```diff
@@ -0,0 +1,28 @@
...
+/// Lyric model.
+/// </summary>
+public class LyricLine
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="LyricLine"/> class.
+    /// </summary>
+    /// <param name="text">The lyric text.</param>
+    /// <param name="start">The lyric start time in ticks.</param>
+    public LyricLine(string text, long? start = null)
```
nielsvanvelzen, line 13: Parameter order is not the same as the property definition order. Triggers my OCD :p

MediaBrowser.Controller/Lyrics/LyricResponse.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace MediaBrowser.Controller.Lyrics;
+
+/// <summary>
+/// LyricResponse model.
+/// </summary>
+public class LyricResponse
+{
+    /// <summary>
+    /// Gets or sets Metadata.
```
nielsvanvelzen, line 12:
```suggestion
    /// Metadata for the lyrics
```

MediaBrowser.Controller/Lyrics/LyricResponse.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// </summary>
+    public LyricMetadata Metadata { get; set; } = new();
+
+    /// <summary>
+    /// Gets or sets Lyrics.
```
nielsvanvelzen, line 17:
```suggestion
    /// Collection of individual lyric lines
```

nielsvanvelzen (changes requested):
API looks fine, got a few comments on the property descriptions. The summaries for API models end up in the OpenAPI schema and eventually in the generated API code. So I'd like to have them somewhat descriptive for the client developers.

MediaBrowser.Controller/Lyrics/LyricMetadata.cs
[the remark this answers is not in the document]
1hitsong: According to the LRC documentation I've found.

`Artist` is the performer of the song.

`Author` is the person who typed up the lyrics. Meaning they created the lyric data itself.

`By` is the person who took the lyric data and created the LRC file using it.

`Creator` is the software used to create the LRC file.

So, `By` put the lyrics that were created by `Author` into LRC format using `Creator` for the song by `Artist`.

MediaBrowser.Controller/Lyrics/LyricMetadata.cs
```diff
@@ -0,0 +1,54 @@
...
+
+namespace MediaBrowser.Controller.Lyrics;
+
+/// <summary>
+/// LyricMetadata model.
+/// </summary>
+public class LyricMetadata
+{
+    /// <summary>
+    /// Gets or sets Artist - The song artist.
```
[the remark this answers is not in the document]
1hitsong: I can't make this change without the analyzer complaining.

MediaBrowser.Controller/Lyrics/LyricMetadata.cs
1hitsong: Updated the summary.

MediaBrowser.Controller/Lyrics/LyricResponse.cs
1hitsong: I can't make this change without the analyzer complaining.

However `Gets or sets Metadata for the lyrics` would work.

MediaBrowser.Controller/Lyrics/LyricResponse.cs
1hitsong: I can't make this change without the analyzer complaining.

However `Gets or sets a collection of individual lyric lines.` would work.

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
```diff
@@ -0,0 +1,200 @@
...
+            .ToList();
+
+        foreach (string metaDataRow in metaDataRows)
+        {
+            if (!metaDataRow.Contains(':', StringComparison.OrdinalIgnoreCase))
+            {
+                continue;
+            }
+
+            string[] metaDataField = metaDataRow.Split(':', 2, StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries);
```
Bond-009, line 92: Using `IndexOf(':')` and `AsSpan()` with ranges and `Trim()` would be way more efficient
You would only have 2 string allocations (after the span has been checked if it's empty) and only 1 strchr

MediaBrowser.Controller/Lyrics/LyricMetadata.cs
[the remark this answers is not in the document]
nielsvanvelzen: Yeah let's keep the "gets or sets" part for now but make it sound fluent like "Gets or sets the song artist".

MediaBrowser.Controller/Lyrics/LyricMetadata.cs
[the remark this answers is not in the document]
nielsvanvelzen: Right, that makes sense

MediaBrowser.Providers/Lyric/TxtLyricProvider.cs
```diff
@@ -0,0 +1,55 @@
...
+    public LyricResponse? GetLyrics(BaseItem item)
+    {
+        string? lyricFilePath = this.GetLyricFilePath(item.Path);
+
+        if (string.IsNullOrEmpty(lyricFilePath))
+        {
+            return null;
+        }
+
+        string[] lyricTextLines = System.IO.File.ReadAllLines(lyricFilePath);
```
Bond-009, line 39: Why can't you add `using System.IO;` to the file?
Also, would it make sense to make this method async?

MediaBrowser.Providers/Lyric/TxtLyricProvider.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        if (lyricTextLines.Length == 0)
+        {
+            return null;
+        }
+
+        List<LyricLine> lyricList = new(lyricTextLines.Length);
```
Bond-009, line 46: This can be an array, no?

Bond-009 (changes requested):

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
cvium: I don't know about way more efficient. It saves some array allocations, but internally Split already uses Spans

Bond-009: We already do a `Contains` for a fast path, might as well do everything so we don't have that array alloc

MediaBrowser.Providers/Lyric/TxtLyricProvider.cs
1hitsong: > Why can't you add using System.IO; to the file?

Done

> Also, would it make sense to make this method async?

I don't know. Is there a good way to determine if it would be helpful?

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
1hitsong: I'm not 100% sure how to apply what @Bond-009 is suggesting. Can someone post a suggested edit for this one?

cvium: The math is probably off, but this is one way
```C#
            var index = metaDataRow.IndexOf(':');
            if (index == -1)
            {
                continue;
            }

            var metaDataFieldName = metaDataRow.AsSpan(1, index - 1).Trim().ToString();
            var metaDataFieldValue = metaDataRow.AsSpan(index + 1, metaDataRow.Length - index).Trim().ToString();
```

1hitsong: The analyzer really wants me to use Contains() instead of IndexOf().

image

1hitsong: Ah, you're exactly right. I was working top down and stopped once the warning showed.

1hitsong: FYI, the final "math" that worked was
```
    var metaDataFieldNameSpan = metaDataRow.AsSpan(1, index - 1).Trim();
    var metaDataFieldValueSpan = metaDataRow.AsSpan(index + 1, metaDataRow.Length - index - 2).Trim();
```

MediaBrowser.Providers/Lyric/TxtLyricProvider.cs
1hitsong: I took a look at making GetLyrics() async. By doing this I'll need to update ILyricProvider, which leads to updating LRCLyricProvider.

I got the following error in LRCLyricProvider.
```
Parameters or locals of type 'ReadOnlySpan<char>' cannot be declared in async methods or async lambda expressions.
```

Do I need to create a new interface for GetLyrics() to allow it to be async and overload ILyricProvider? Unsure it it's worth all the effort.

cvium: You can get around the limitations by putting the span stuff in a local function

nielsvanvelzen (approved):
API part looks good to me :+1:

MediaBrowser.Providers/Lyric/TxtLyricProvider.cs
```diff
@@ -0,0 +1,58 @@
...
+        string[] lyricTextLines = await File.ReadAllLinesAsync(lyricFilePath).ConfigureAwait(false);
+
+        if (lyricTextLines.Length == 0)
+        {
+            return null;
+        }
+
+        LyricLine[] lyricList = new LyricLine[lyricTextLines.Length];
+
+        for (int lyricLine = 0; lyricLine < lyricTextLines.Length; lyricLine++)
```
cvium, line 51: `lyricLine` is a misleading variable name since it's actually an index

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
```diff
@@ -0,0 +1,211 @@
...
+        {
+            lyricMetadata.Version = version;
+        }
+
+        return lyricMetadata;
+    }
+
+    private static string GetMetadataFieldName(string metaDataRow, int index)
+    {
+        return metaDataRow.AsSpan(1, index - 1).Trim().ToString();
```
cvium, line 204: I'm not sure but I think it might be better to check `IsEmpty` and then return `string.Empty` instead of calling `ToString` if the span is empty

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
```diff
@@ -0,0 +1,213 @@
...
+            // Example 2: [ar: Calabrese]
+            var metaDataFieldName = GetMetadataFieldName(metaDataRow, index);
+            var metaDataFieldValue = GetMetadataValue(metaDataRow, index);
+
+            if (string.IsNullOrEmpty(metaDataFieldName) || string.IsNullOrEmpty(metaDataFieldValue))
+            {
+                continue;
+            }
+
+            fileMetaData[metaDataFieldName.ToString()] = metaDataFieldValue.ToString();
```
cvium, line 106: Aren't they already strings?
```suggestion
            fileMetaData[metaDataFieldName] = metaDataFieldValue;
```

MediaBrowser.Controller/Lyrics/LyricInfo.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        // Ensure the directory path exists
+        if (!Directory.Exists(itemDirectoryPath))
+        {
+            return null;
+        }
+
+        foreach (var lyricFilePath in Directory.GetFiles(itemDirectoryPath, $"{Path.GetFileNameWithoutExtension(itemPath)}.*"))
+        {
+            if (EnumerableExtensions.Contains(lyricProvider.SupportedMediaTypes, Path.GetExtension(lyricFilePath.AsSpan())[1..], StringComparison.OrdinalIgnoreCase))
```
Bond-009, line 41: It's an extension method, should just be able to do this:
```suggestion
            if (lyricProvider.SupportedMediaTypes.Contains(Path.GetExtension(lyricFilePath.AsSpan())[1..], StringComparison.OrdinalIgnoreCase))
```

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
```diff
@@ -0,0 +1,213 @@
...
+            long ticks = TimeSpan.FromMilliseconds(timeData.Value).Ticks;
+            lyricList.Add(new LyricLine(sortedLyricData[i].Text, ticks));
+        }
+
+        if (fileMetaData.Count != 0)
+        {
+            // Map metaData values from LRC file to LyricMetadata properties
+            LyricMetadata lyricMetadata = MapMetadataValues(fileMetaData);
+
+            return new LyricResponse { Metadata = lyricMetadata, Lyrics = lyricList };
```
Bond-009, line 133: Format this over multiple lines or use/create a constructor

MediaBrowser.Providers/Lyric/LrcLyricProvider.cs
```diff
@@ -0,0 +1,213 @@
...
+        }
+
+        return new LyricResponse { Lyrics = lyricList };
```
Bond-009, line 136: Same

MediaBrowser.Providers/Lyric/TxtLyricProvider.cs
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        LyricLine[] lyricList = new LyricLine[lyricTextLines.Length];
+
+        for (int lyricLineIndex = 0; lyricLineIndex < lyricTextLines.Length; lyricLineIndex++)
+        {
+            lyricList[lyricLineIndex] = new LyricLine(lyricTextLines[lyricLineIndex]);
+        }
+
+        return new LyricResponse { Lyrics = lyricList };
```
Bond-009, line 56: Same

Bond-009 (approved):
After these cosmetic changes it should be fine

Chaphasilor:
Hi, I'm trying to start implementing client support for the new lyrics endpoints, but the endpoints don't seems to be in the `unstable` build of the server. There's also no mention of these endpoints in the [OpenAPI doc for the unstable channel](https://api.jellyfin.org/openapi/jellyfin-openapi-unstable.json).  
Am I looking in the wrong place? I assumed these changes were merged into `master`, and the `unstable` channel was built from `master`?

nielsvanvelzen:
The lyric stuff is definitely there

image

Chaphasilor:
I found the problem. `HasLyrics` is not included when fetching multiple items (i.e. not specifying the `itemId` but only some filters like `ParentId`), even when including `Fields: HasLyrics` in the query parameters. I thought I'd tested that, but it seems like I got unlucky and used a song that didn't have lyrics (or at least the server hadn't picked them up yet).  
Thanks for your help!

Issue #8387, opened 2022-09-11
[Issue]: Special Features not sorting correctly (alphanumerically)


somewherebutnothere:
### Please describe your bug

Special Features are not sorting correctly, as in alphanumerically.

### Jellyfin Version

10.8.0

### Environment

```markdown
- OS: Ubuntu Server 22.04
- Virtualization: Docker
- Clients: Web
- Browser: Chrome
```

### Please attach any screenshots here

![image](https://user-images.githubusercontent.com/6793358/189545149-51fcad83-0f49-4455-9e78-e34fbd4e99a6.png)

![image](https://user-images.githubusercontent.com/6793358/189545685-5a14a546-5b3b-481d-ba62-96da29ba4027.png)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

RatVader:
I've experienced this same issue with unexpected sorting of a TV Show's "extras" directory in Ubuntu 18.04.6 LTS (GNU/Linux 4.15.0-204-generic x86_64) running Jellyfin.Server 10.8.9.0 on March 26, 2023.  I had files sorted like so, with Jellyfin seemingly ignoring the numbers at the beginning:
- 02. Ending 1.mkv
- 04. Ending 2.mkv
- 01. Opening 1.mkv
- 03. Opening 2.mkv

I was able to sort them by preceding the number with an 'E' but the same method failed when preceded by an 'O', 'S', or 'S00E':
- E01 Opening 1.mkv
- E02 Ending 1.mkv
- E03 Opening 2.mkv
- E04 Ending 2.mkv

PR #8401, opened 2022-09-13
Fix special features not being sorted

ZachPhelan:
**Changes**
Make the `GetSpecialFeatures` function consistent with the rest of the file by using `GetExtras()` and filtering down to what is needed. The overloaded `GetExtras(BaseItem.DisplayExtraTypes)` was only used for `GetSpecialFeature`, so I opted to delete it as it is no longer needed. 

**Issues**
[8387](https://github.com/jellyfin/jellyfin/issues/8387)

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -148,6 +148,7 @@
  - [xosdy](https://github.com/xosdy)
  - [XVicarious](https://github.com/XVicarious)
  - [YouKnowBlom](https://github.com/YouKnowBlom)
+ - [ZachPhelan](https://github.com/ZachPhelan)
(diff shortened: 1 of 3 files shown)


ZachPhelan:
> Breaks at least the reports plugin https://github.com/jellyfin/jellyfin-plugin-reports/blob/0c0a27abdb33b9b86d14806e4c53a61a5a21e07a/Jellyfin.Plugin.Reports/Api/Data/ReportBuilder.cs#L580

Yeah you're right. I did not think about plugins. Thanks for the feedback. I still think it's better form to use the base function without the parameter, but can keep the other for compatibility.

cvium (approved):

Bond-009 (approved):

PR #8407, opened 2022-09-14
Fix HTML for badges in README

hannesbraun:
**Changes**

The HTML for the badges in the README did not look valid to me. For example, my JetBrains IDE wasn't able to render the README correctly. This should fix the invalid HTML.


Bond-009 (approved):

crobibero (approved):
We should also remove the lgtm badge as it's shutting down at the end of this year...

PR #4546, opened 2020-11-22
If baseurl is not set, redirect relatively on the initial web/index.html redirect

oddstr13:
This fixes the redirect failure when nginx is doing baseurl magic while Jellyfin is unaware.
(regression from then baseurl was first introduced in server)

PR #8408, opened 2022-09-14
Redirect natively if no baseurl is set

Shadowghost:
Based on the code from #4546 but uses `System.Uri` to generate the proper redirect path.

**Changes**
Fix redirect failures by redirecting back to the index page.

----

Supersedes #4546

### DIFF
--- a/Jellyfin.Server/Middleware/BaseUrlRedirectionMiddleware.cs
+++ b/Jellyfin.Server/Middleware/BaseUrlRedirectionMiddleware.cs
@@ -68,13 +68,8 @@ namespace Jellyfin.Server.Middleware
                 // Always redirect back to the default path if the base prefix is invalid or missing
                 _logger.LogDebug("Normalizing an URL at {LocalPath}", localPath);
 
-                if (string.IsNullOrWhiteSpace(baseUrlPrefix) || baseUrlPrefix[^1] != '/')
-                {
-                    baseUrlPrefix += "/";
-                }
-
                 var uri = new Uri(localPath);
-                var redirectUri = new Uri(baseUrlPrefix + _configuration[DefaultRedirectKey]);
+                var redirectUri = new Uri(baseUrlPrefix + "/" + _configuration[DefaultRedirectKey]);
                 var target = uri.MakeRelativeUri(redirectUri).ToString();
                 _logger.LogDebug("Redirecting to {Target}", target);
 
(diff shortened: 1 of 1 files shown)


crobibero (commented):
This removes the health check redirect

Shadowghost:
I guess we need to ecplicitly exclude it. What about the metrics endpoint?

crobibero:
The `/health` redirect is there for mostly automatic docker health checks, so that redirect should still exist. The metrics endpoint shouldn't automatically redirect, the user should just configure their custom tools correctly

crobibero (approved):

Jellyfin.Server/Middleware/BaseUrlRedirectionMiddleware.cs
```diff
@@ -45,36 +45,40 @@ public async Task Invoke(HttpContext httpContext, IServerConfigurationManager se
...
                 }
 
-                if (!startsWithBaseUrl
-                    || localPath.Length == baseUrlPrefix.Length
-                    // Local path is /baseUrl/
-                    || (localPath.Length == baseUrlPrefix.Length + 1 && localPath[^1] == '/'))
+                // Always redirect back to the default path if the base prefix is invalid or missing
+                _logger.LogDebug("Normalizing an URL at {LocalPath}", localPath);
+
+                if (string.IsNullOrWhiteSpace(baseUrlPrefix) || baseUrlPrefix[^1] != '/')
```
cvium, line 71: This is always true https://github.com/jellyfin/jellyfin/blob/90174f68e2b477e613d756d78cbfd89c2a33f30a/Jellyfin.Networking/Configuration/NetworkConfiguration.cs#L64

Jellyfin.Server/Middleware/BaseUrlRedirectionMiddleware.cs
```diff
@@ -45,36 +45,35 @@ public async Task Invoke(HttpContext httpContext, IServerConfigurationManager se
             var localPath = httpContext.Request.Path.ToString();
             var baseUrlPrefix = serverConfigurationManager.GetNetworkConfiguration().BaseUrl;
 
-            if (!string.IsNullOrEmpty(baseUrlPrefix))
+            if (string.IsNullOrEmpty(localPath)
+                || string.Equals(localPath, "/", StringComparison.OrdinalIgnoreCase)
+                || string.Equals(localPath, "/web", StringComparison.OrdinalIgnoreCase)
+                || string.Equals(localPath, "/web/", StringComparison.OrdinalIgnoreCase)
```
cvium, line 51: These seem very redundant given the `!localPath.StartsWith(baseUrlPrefix, StringComparison.OrdinalIgnoreCase)` check

cvium (approved):

PR #8411, opened 2022-09-15
Allow direct play even if no audio stream is available

Maxr1998:
If a video doesn't have an audio stream, it cannot direct play right now, since no supported audio stream can is available (`FirstOrDefault` always returns `null` on empty collections).
This commit ensures that any audio stream is available before matching the device profiles.

**Changes**
* Wraps the audio stream check in a `.Any()` if clause to ensure at least one audio stream is available.
* The unused variable `list` is removed.
* `videoSupported` and `audioSupported` checks optimized to return faster – `null` streams are always “supported”.

**Issues**
* [jellyfin/jellyfin-android#28](https://github.com/jellyfin/jellyfin-android/issues/28#issuecomment-1221328279)

### DIFF
--- a/MediaBrowser.Model/Dlna/StreamBuilder.cs
+++ b/MediaBrowser.Model/Dlna/StreamBuilder.cs
@@ -436,9 +436,9 @@ namespace MediaBrowser.Model.Dlna
                 {
                     containerSupported = true;
 
-                    videoSupported = videoStream != null && profile.SupportsVideoCodec(videoStream.Codec);
+                    videoSupported = videoStream == null || profile.SupportsVideoCodec(videoStream.Codec);
 
-                    audioSupported = audioStream != null && profile.SupportsAudioCodec(audioStream.Codec);
+                    audioSupported = audioStream == null || profile.SupportsAudioCodec(audioStream.Codec);
 
                     if (videoSupported && audioSupported)
                     {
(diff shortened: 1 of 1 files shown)


nielsvanvelzen (approved):

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -1182,14 +1181,18 @@ private static int GetMaxAudioBitrateForTotalBitrate(long totalBitrate)
                     }
 
                     // Check audio codec
-                    var selectedAudioStream = candidateAudioStreams.FirstOrDefault(audioStream => directPlayProfile.SupportsAudioCodec(audioStream.Codec));
-                    if (selectedAudioStream == null)
+                    MediaStream selectedAudioStream = null;
+                    if (candidateAudioStreams.Any())
```
Bond-009, line 1185: Looks like `candidateAudioStreams` is iterated over multiple times, maybe it schould be a collection instead of an `IEnumerable?`

Bond-009 (changes requested):

MediaBrowser.Model/Dlna/StreamBuilder.cs
Maxr1998: Seems like a good idea, I'll look into it tomorrow.

Maxr1998: Done. Let me know if I missed anything.

cvium (approved):

PR #8472, opened 2022-09-28
Fix subtitle selection behaviour

### LINKED ISSUE #8251: [Issue]: Always defaults to English subtitles
### Please describe your bug

Even if you have set another language as default in the subtitle settings, Jellyfin always default to English subtitles.

### Jellyfin Version

10.8.4

### if other:

_No response_

### Environment

```markdown
- OS: OpenSUSE
- Virtualization: Docker
- Clients: Web, Android, Chromecast
- Browser: Firefox 103
- FFmpeg Version: 5.0.1-Jellyfin
- Playback Method: DirectPlay
- Hardware Acceleration: none
- Plugins: none
- Reverse Proxy: nginx
- Base URL: none
- Networking: host
- Storage: local
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

jgriff6:
**Changes**

Aligned the behaviour of the subtitle selection logic with the descriptions that are given on the front end. Currently the behaviour is not as the user would expect.

- "Default" should account for language preference when there are multiple options of equal priority, and should play subtitle tracks according to external, forced or default flag priority.
- "Smart" should respect forced tracks, then only attempt to play other subtitles if the user's subtitle language preference does not match the audio track language. If subtitles in the preferred language are not present then fall back on default behaviour.
- "Always Play" should always attempt to play subtitles matching the user's language preference, and fall back on default behaviour otherwise.

**Issues**

Fixes #8251

### DIFF
--- a/Emby.Server.Implementations/Library/MediaStreamSelector.cs
+++ b/Emby.Server.Implementations/Library/MediaStreamSelector.cs
@@ -45,42 +45,40 @@ namespace Emby.Server.Implementations.Library
                 .ThenByDescending(x => x.IsForced && string.Equals(x.Language, audioTrackLanguage, StringComparison.OrdinalIgnoreCase))
                 .ThenByDescending(x => x.IsForced)
                 .ThenByDescending(x => x.IsDefault)
+                .ThenByDescending(x => preferredLanguages.Contains(x.Language, StringComparison.OrdinalIgnoreCase))
                 .ToList();
 
             MediaStream? stream = null;
             if (mode == SubtitlePlaybackMode.Default)
             {
-                // Prefer embedded metadata over smart logic
-                stream = sortedStreams.FirstOrDefault(s => s.IsExternal || s.IsForced || s.IsDefault);
-
-                // if the audio language is not understood by the user, load their preferred subs, if there are any
-                if (stream == null && !preferredLanguages.Contains(audioTrackLanguage, StringComparison.OrdinalIgnoreCase))
-                {
-                    stream = sortedStreams.FirstOrDefault(s => !s.IsForced && preferredLanguages.Contains(s.Language, StringComparison.OrdinalIgnoreCase));
-                }
+                // Load subtitles according to external, forced and default flags.
+                stream = sortedStreams.FirstOrDefault(x => x.IsExternal || x.IsForced || x.IsDefault);
             }
             else if (mode == SubtitlePlaybackMode.Smart)
             {
-                // if the audio language is not understood by the user, load their preferred subs, if there are any
+                // Respect forced flag.
+                stream = sortedStreams.FirstOrDefault(x => x.IsForced);
+
+                // Only attempt to load subtitles if the audio language is not one of the user's preferred subtitle languages.
+                // If no subtitles of preferred language available, use default behaviour.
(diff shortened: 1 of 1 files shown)


Emby.Server.Implementations/Library/MediaStreamSelector.cs
```diff
@@ -45,42 +45,40 @@ public static class MediaStreamSelector
...
-                    stream = sortedStreams.FirstOrDefault(s => !s.IsForced && preferredLanguages.Contains(s.Language, StringComparison.OrdinalIgnoreCase));
-                }
+                // Load subtitles according to external, forced and default flags.
+                stream = sortedStreams.FirstOrDefault(x => x.IsExternal || x.IsForced || x.IsDefault);
             }
             else if (mode == SubtitlePlaybackMode.Smart)
             {
-                // if the audio language is not understood by the user, load their preferred subs, if there are any
+                // Respect forced flag.
+                stream = sortedStreams.FirstOrDefault(x => x.IsForced);
```
cvium, line 60: This should be in an else branch

jgriff6: For sure, I've gone ahead and changed that.

cvium (approved):
I haven't tested it but it makes sense to me

Emby.Server.Implementations/Library/MediaStreamSelector.cs
```diff
@@ -45,42 +45,42 @@ public static class MediaStreamSelector
...
+                    stream = sortedStreams.FirstOrDefault(x => x.IsForced);
                 }
             }
             else if (mode == SubtitlePlaybackMode.Always)
             {
-                // always load the most suitable full subtitles
-                stream = sortedStreams.FirstOrDefault(s => !s.IsForced);
+                // Always load subtitles of the user's preferred subtitle language if possible, otherwise default behaviour.
+                stream = sortedStreams.FirstOrDefault(x => preferredLanguages.Contains(x.Language, StringComparison.OrdinalIgnoreCase)) ??
+                    sortedStreams.FirstOrDefault(x => x.IsExternal || x.IsForced || x.IsDefault);
```
1337joe, line 76: Forced tracks sort first and are limited to what needs to be translated to the language of the sub track, where mode `Always` means that the user wants subtitles for everything. I think this would benefit from the original `!s.IsForced` check, at least when comparing to `preferredLanguages`.

jgriff6: It's a good point that the user would prefer full subtitles so I've re-added the check.

Do you think it's worth considering the case where an external track in an alternate language is chosen over a forced track in the desired language? I'm thinking that would generally be desired because downloading external tracks is nearly always done purposefully?

1337joe (approved):

Emby.Server.Implementations/Library/MediaStreamSelector.cs
1337joe: Yeah, I think anything not forced and in `preferredLanguages` is a valid choice, the rest of the sorting and selecting looks reasonable to me.

PR #8511, opened 2022-10-07
Use is and is not when comparing with null

Bond-009:


crobibero (approved):

Shadowghost (approved):

Shadowghost:
I'm not sure if changing so many files for nor real gain is a good idea though.

PR #8515, opened 2022-10-08
feat: macOS arm64 build support

gnattu:
Signed-off-by: gnattu <gnattuoc@me.com>

**Changes**
This add building support for macOS arm64 target which is pretty stable now on dotnet 6. One thing to note is that the old build script is renamed from `build.macos` to `build.macos.amd64`, I did not find direct use of these scripts other than the build script itself, if it breaks anything please let me know.

Already tested with a ubuntu 22.04 amd64 buiding host using docker crossbuild. 

**Issues**

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -13,5 +13,6 @@ packages:
   - centos.amd64
   - linux.amd64
   - windows.amd64
-  - macos
+  - macos.amd64
+  - macos.arm64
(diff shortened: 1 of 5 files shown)


joshuaboniface (approved):
Looks good.

Bond-009 (approved):

Issue #3636, opened 2020-07-20
10.6 Unable to View Playlists Unless User Has Created One


omniusaspirer:
**Describe the bug**
Upon updating to 10.6 I noticed that I was unable to view any of my previous music playlists until creating a new playlist after the update, which then allowed me to see all my preexisting playlists. This is a regression from 10.5.5 on which a user was able to see all server playlists regardless of whether they had personally created one. 

**System (please complete the following information):**
 - OS: Ubuntu 18.04
 - Virtualization: n/a
 - Clients: All
 - Browser: All
 - Jellyfin Version: 10.6
 - Playback: n/a
 - Installed Plugins: none
 - Reverse Proxy:none
 - Base URL: none
 - Storage: local

**To Reproduce**
Create a playlist on one user, then note that playlist is not visible on another user until 2nd user creates a playlist.

**Expected behavior**
All server playlists should be visible to all users regardless of whether they have personally created a playlist (ideally playlists could optionally be flagged as public/shared but that would be a feature request).

**Logs**
n/a

**Screenshots**
n/a

**Additional context**
n/a

crobibero:
I just tested 10.5, it looks like users without created playlists were not able to see other users' playlists.
Digging through the current code I see where playlists are being dropped, but this looks to require a much larger fix.

omniusaspirer:
There was absolutely a change somewhere in the 10.6 update which removed the ability for users to see even their own playlists unless they create a new playlist post 10.6. Worse still, if that user deletes the new 10.6 playlist they once again lose the ability to see their own pre-10.6 playlists, thus forcing the creation of permanent dummy playlists by each user with a previous playlist or the recreation of all old playlists. 

Just one of my personal playlists contained over 1200 items, so while I could manually go through them all and force my users to do the same if a lesser quick fix to seeing a users own playlists could be implemented it would be appreciated.

omniusaspirer:
I just recalled that the reason users were able to see playlists without creating them was because I'd put in m3u playlists which *were* visible to all users regardless of whether they'd created their own. This then allowed them to see other users playlists as well. Unfortunately the 10.6 update simply deleted any non XML playlist from the playlists directory so that workaround feels less viable after being once-bitten.

FazeThree71:
Regarding the m3u playlists tidbit - I've always removed these from my folders so that Jellyfin would handle the playlists internally. So for me, the m3u file doesn't apply but I have the same issue.

[...]

PR #8526, opened 2022-10-09
Overhaul content rating handling

### LINKED ISSUE #4363: Parental rating not being inherited
**Describe the bug**
If I change the rating of a certain anime show (saint seiya) TV-14 to FR-10 the changes are not transfered to episode, so the anime shows in the kid account, but empty.

 - OS: archlinux
 - Virtualization: docker
 - Jellyfin Version: 10.6.4

**Expected behavior**
Episode appears on the kid account

### LINKED ISSUE #5939: Movie list does not load when logged in as user with a maximum allowed parental rating and viewing a combined library folder
**Describe the bug**

I've been as accurate as I could be in the title of this issue. When the following conditions are met, movies do not populate the list in which they are supposed to show up:

- The current user is not an admin.
- The current user has a "maximum allowed parental rating" set by an admin
- The current user has two or more libraries combined for viewing
- The combined library is attempted to be viewed.

From the logs, this appears to be a permissions issue, perhaps due to the "maximum allowed parental rating". However, if the libraries are not combined, the same user can view all of the individual libraries and just see the movies they are allowed to see.

 - OS: Ubuntu 20.04
 - Virtualization: Docker
 - Clients: Windows
 - Browser: Chrome
 - Jellyfin Version: 10.7.2
 - Playback: n/a
 - Installed Plugins: Trakt, Playback Reporting, but I don't believe these apply here.
 - Reverse Proxy: Caddy, but really this doesn't apply either.
 - Base URL: none
 - Networking: Host
 - Storage: Local

**To Reproduce**

1. Have multiple libraries containing movies.
2. Create a new user (non-admin).
3. In the admin dashboard for that user, on the Parental Control tab, set a maximum-allowed parental rating, eg. CA-PG.
4. Log in as the new user.
5. In the user's settings, under "Home" in the "Library Folders" section, check at least two libraries containing movies.
6. Go back to the home screen for the user, and attempt to navigate into the combined "Movies" library.
7. The spinner gif spins forever, and no movies ever show.

**Expected behavior**

The list of movies in the combined library should show, filtered by the user's maximum-allowed parental rating.

**Logs**

```
Jellyfin.Api.Controllers.ItemsController: "MyKidsName" is not permitted to access Library "Movies"
```
In this log entry, "Movies" is the name of the combined library folder.

**Screenshots**

image
image
image

**Additional context**

### LINKED ISSUE #6046: Block items with tags doesn't hide from Latest
**Describe the bug**
Duplicate of [#2079](https://github.com/jellyfin/jellyfin/issues/2079), when you block tags, the users impacted can still see the tagged material in the 'Latest' views.

 - OS: Debian
 - Virtualization: N/A
 - Clients: All
 - Browser:
 - Jellyfin Version: 10.7.5
 - Playback: N/A
 - Installed Plugins: none
 - Reverse Proxy: none
 - Base URL: none
 - Networking: Host
 - Storage: local

**To Reproduce**
1. Go to Edit Item Metadata for a newly added item
2. Add tag 'x' and save
3. Create user, go to Parental Control, under Block Item with Tags add 'x'
4. Go to home page, see item still in 'Latest' despite not appearing when browsed to.

**Expected behavior**
Item shouldn't be visible anywhere to this user

### LINKED ISSUE #6514: Playlists 'infinite loading' if the user not have all access libraries
The Playlists 'infinite loading' if the user not have all access libraries.

**Describe the bug**
The playlists don't launch if I not have all access libraries.
Whith and whithout server plugins, the bug remains.
If I check the box 'Enable access to all libraries' for the user, it's ok.
If I check the box 'Allow this user to manage the server' for the user, it's Not ok.

 - OS: Debian 10
 - Virtualization: VM guest (KVM Proxmox)
 - Clients: Browser, Android (and app Gelli on Android)
 - Browser: Firefox 91.0.2 (64 bits), Chromium Version 93.0.4577.63 (Build officiel) snap (64 bits)
 - Jellyfin Version: 10.7.6
 - FFmpeg Version: don't know
 - Playback: don't know 
 - Hardware Acceleration: don't know
 - Installed Plugins: AudioDB10.7.6.0 , Auto Organize9.0.0.0 , Kodi Sync Queue6.0.0.0 , MusicBrainz10.7.6.0 , OMDb10.7.6.0, Playback Reporting10.0.0.0 , Reports12.0.0.0
 - Reverse Proxy: none
 - Base URL: none
 - Networking: NAT
 - Storage: local mount of SMB share on NAS
 
 **To Reproduce**
1. Go to 'Media'
2. Click on 'Playlists'
3. See infinite loading
 Or
1. Go to 'Media / Music'
2. Click on 'header section : Playlists'
3. Choose to open 'one of your playlists'
4. Click on 'header play button'
3. See infinite loading

**Screenshots**
1_JellyFin_info_srv

2_JellyFin_Access_Libraries

3_JellyFin_pb_playlists

4_JellyFin_pb_playlists

**Additional context**
The bug is still on WebClient (any browser, any PC, Linux and Windows), on Android App "Jellyfin (Mobile)" and "JellyFin (AndroidTV)" and "Gelli (Android FDroid).

When you check the box "Enable access to all libraries" for the user, it's ok.
All users are concerned except the admin.

### LINKED ISSUE #7637: [Issue]: Parental control doesn't seem to work
### Please describe your bug

I decided to create my eldest his own profile, so he could have access to my library, but with an age restriction of GB-15. He has access to all the GB-18 movies too though. I shutdown the server, fired it back up and logged in again to double check, but there are plenty of GB-18 movies visible. It seems that Parental Controls are not working.

### Jellyfin Version

10.7.7

### if other:

_No response_

### Environment

```markdown
- OS:Debian
- Virtualization: Docker
- Clients: Browser
- Browser: Firefox
- FFmpeg Version: 4.3.2-Jellyfin
- Playback Method: N/A
- Hardware Acceleration: N/A
- Plugins: Default
- Reverse Proxy: nginx
- Base URL: /jellyfin
- Networking: Ethernet/WiFi
- Storage: 1TB SSD
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

### LINKED ISSUE #7733: Parental Controls not respected when searching
### Please describe your bug

I am running Linux with Jellyfin 10.8
The search function is not excluding parental control tags or ratings and showing content that should be excluded. 

I have Parental Content setup for users with excluded tags

User Parental Content exclude tags, , See image
image

Live TV Meta Data (Channel is set to have tag, ADULT-XXX),, See image
image

This is being filtered out of the live tv correctly however when you search the channels that should be excluded show up as all kinds of media types, See image
image

### Jellyfin Version

10.8.0

### if other:

_No response_

### Environment

```markdown
- OS: Debian
- Virtualization:
- Clients: All
- Browser: Chrome
- FFmpeg Version: 4.32
- Playback Method:
- Hardware Acceleration:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

### LINKED ISSUE #8263: [Issue]: Merging multiple movie libraries into movies category requires full library access
### Please describe your bug

I have three different 'Movie' libraries and have set-up to collate all of these into the single Movies category.
If I don't enable a user with 'full access' but still give them access to all of the libraries individually, they don't see any movies in the grouped category at all, it's just a blank grid.
I haven't tried to split out one or more of the libraries from their access to test but I suspect if it happens when they have all access individually (but not access to all libraries) breaks things, I suspect the same would create the issue

### Jellyfin Version

Other

### if other:

10.8.3

### Environment

```markdown
- OS: Ubuntu 22.10
- Virtualization: N/A
- Clients: Android TV (NVIDIA Shield)
- Browser: N/A
- FFmpeg Version: Packaged
- Playback Method: N/A
- Hardware Acceleration: N/A
- Plugins: N/A
- Reverse Proxy: N/A
- Base URL: N/A
- Networking: N/A
- Storage: NFS
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

### LINKED ISSUE #8950: [Issue]: Private playlists are not properly hidden if users create playlists
### Please describe your bug

Reproduction steps:

1. Create an admin user
2. Create a playlist as the admin user for a private library
3. Create a user without admin permissions
4. Switch to the newly-created user account
5. Create a playlist on the new user account
6. Create another playlist and you will now be able to see the playlist for the private library 

### Jellyfin Version

10.7.7

### if other:

_No response_

### Environment

```markdown
- OS:
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

Shadowghost:
How content ratings are currently handled is a bit ominous and apparently incorrect (see #7637)

**Changes**
Align the country rating systems with:
* [IMDB Certification](https://help.imdb.com/article/contribution/titles/certificates/GU757M8ZJ9ZPXB39)
* [Wikipedia Motion Picture Content Rating Systems](https://en.wikipedia.org/wiki/Motion_picture_content_rating_system)
* [Wikipedia TV Parental Guidelines](https://en.wikipedia.org/wiki/TV_Parental_Guidelines)

Apply the following schema to all country rating systems:
```
null - Unrated Content (need to be explicitly excluded via user settings)
0 - 0+, Informational and Educational Content
1 - 1+
2 - 2+
3 - 3+
4 - 4+
5 - 5+
6 - 6+
7 - 7+
8 - 8+
9 - 9+
10 - 10+
11 - 11+
12 - 12+ 
13 - 13+
14 - 14+
15 - 15+
16 - 16+
17 - 17+
18 - 18+
1000 - Explicit Content
1001 - Prohibited/Not Approved/Banned Content
```

Furthermore another dummy content rating system was added with fallback values and arbitrary strings which take precedence over all others when matching against all available rating systems (this allows forcing specific behaviour if different systems use the same tag for different ratings)

Also keep in mind that the user specific setting in jellyfin-web sets the **maximum** allowed rating.
Since this setting is language specific some default values are added to ensure you're able to select common (US-based) ratings,

Additionally this PR solves some access control issues when blocking unrated content for specific item types and especially playlists.
The Playlist and UserLibrary endpoints now respect parental control when returning the items.
This prevents users from viewing items or playlists they have direct links or item ids for, they can only see and access their own playlists and only items which they have access to.

To make all this work a database migration is also included but due to inconsistencies in the current system it's impossible to properly map everything, so a full rescan of your media is required to make sure the values are set as intended.

**Issues**
Fixes #4363 
Fixes #5939 
Fixes #6046 
Fixes #6514 
Fixes #7637 
Fixes #7733
Fixes #8263
Fixes #8950


Emby.Server.Implementations/Localization/LocalizationManager.cs
```diff
@@ -260,6 +264,18 @@ public IEnumerable<ParentalRating> GetParentalRatings()
...
+            {
+                var trimmedRating = rating.AsSpan(index).TrimStart('-').Trim();
+
+                if (!trimmedRating.IsEmpty)
+                {
+                    return GetRatingLevel(trimmedRating.ToString());
+                }
+            }
+
             // TODO: Further improve by normalizing out all spaces and dashes
```
1337joe, line 279: Post-overhaul is this TODO still relevant?

nielsvanvelzen:
The table at https://en.wikipedia.org/wiki/Television_content_rating_system indicates some content rating systems include ages up to 21. Would it make sense to mirror that in this change? Might even be better to start counting down from int.maxvalue for the "special" values to be prepared for systems that include even more ages.

Emby.Server.Implementations/Localization/LocalizationManager.cs
Shadowghost: You never know what weird stuff people put into their NFOs...

Shadowghost:
I'm open to do that but we could also just use values of 30 and above imo.

cvium:
If we are doing special values, we should pick a number that isn't an age. Like 1000+.

Jellyfin.Api/Controllers/ItemUpdateController.cs
```diff
@@ -224,7 +224,7 @@ public ActionResult UpdateItemContentType([FromRoute, Required] Guid itemId, [Fr
             return NoContent();
         }
 
-        private void UpdateItem(BaseItemDto request, BaseItem item)
+        private async void UpdateItem(BaseItemDto request, BaseItem item)
```
crobibero, line 227: Don't use `async void`

Shadowghost: What would I use instead? Don't I need to await the saving on the item repo?

crobibero: Yes, but this function needs to be `async Task`

Emby.Server.Implementations/Localization/LocalizationManager.cs
```diff
@@ -184,7 +183,56 @@ public IEnumerable<CountryInfo> GetCountries()
...
+            }
+
+            // Catchall if max rating of country is less than 21
+            // Using 21 instead of 18 to be sure to allow access to all rated content except adult and banned
+            if (!ratings.Any(x => x.Value >= 21))
+            {
+                ratings.Add(new ParentalRating("21", 21));
+            }
+
+            // A lot of countries don't excplicitly have a seperate rating for adult content
```
cvium, line 222:
```suggestion
            // A lot of countries don't explicitly have a separate rating for adult content
```

Emby.Server.Implementations/Localization/LocalizationManager.cs
```diff
@@ -184,7 +183,56 @@ public IEnumerable<CountryInfo> GetCountries()
...
+            if (!ratings.Any(x => x.Value == 1000))
+            {
+                ratings.Add(new ParentalRating("XXX", 1000));
+            }
+
+            // A lot of countries don't excplicitly have a seperate rating for banned content
```
cvium, line 228:
```suggestion
            // A lot of countries don't explicitly have a separate rating for banned content
```

Emby.Server.Implementations/Localization/LocalizationManager.cs
```diff
@@ -257,7 +308,19 @@ public IEnumerable<ParentalRating> GetParentalRatings()
...
+            index = rating.IndexOf('-', StringComparison.Ordinal);
+            if (index != -1)
+            {
+                var trimmedRating = rating.AsSpan(index).TrimStart('-').Trim();
+
+                if (!trimmedRating.IsEmpty)
+                {
+                    return GetRatingLevel(trimmedRating.ToString());
+                }
+            }
```
Bond-009, line 321: https://github.com/jellyfin/jellyfin/blob/master/src/Jellyfin.Extensions/StringExtensions.cs#L74

candry7731:
Movies that have "unrated" rating is not filtered if user has a lower parental rating filter. 

IE. if I select Maximum allowed parental rating: TV-14
The user will still see unrated movies which in most cases is worse than TV-14

Shadowghost:
Unrated content does men exactly that: It has no rating.
To explicitly exclude such content for a user you need to enable this explicitly by setting the option `Block items with no or unrecognized rating information` for the user on the library.

Emby.Server.Implementations/Data/SqliteItemRepository.cs
```diff
@@ -4020,34 +4020,116 @@ private List<string> GetWhereClauses(InternalItemsQuery query, IStatement statem
...
-                if (statement is not null)
+                ratingClause += "InheritedParentalRatingValue not null";
+                if (query.MinParentalRating.HasValue)
                 {
-                    statement.TryBind("@MinParentalRating", query.MinParentalRating.Value);
+                    ratingClause += " AND InheritedParentalRatingValue >= @MinParentalRating";
+                    if (statement is not null)
+                    {
+                        statement.TryBind("@MinParentalRating", query.MinParentalRating.Value);
+                    }
```
Bond-009, line 4033: Where possible
```suggestion
                    statement?.TryBind("@MinParentalRating", query.MinParentalRating.Value);
```

Emby.Server.Implementations/Localization/Ratings/be.csv
```diff
@@ -1,6 +1,11 @@
...
-BE-16,8
+AL,0
+KT,0
+TOUS,0
+MG6,6
+6,6
+9,9
+KNT,12
+12,12
+BE_14,14
```
sjorge, line 9: Missed country prefix when they were removed?

Shadowghost: If I remember correctly `KNT` is not a prefix but a rating and `BE_14` was kept because the country prefix code only used `-` as a delimiter.

sjorge: `KNT` is indeed not a prefix, it's `Kinderen Niet Toegelaten`, Bit odd there is no french equivalent like `AL` (Dutch) has `TOUS` (french), then again Belgium is just a bit weird.

It's only the `BE_14` that seems aweird to me, I could also find no mention of it. There are however age categories 6,9,12,14 and 16. So my guess was BE_ was just a typo and should have been BE- and this also stripped.

But I had a very very hard time finding any form of official standard, closes I found was that the old one was replaced by the dutch one on January 2020 (https://www.brusselstimes.com/belgium/82830/films-broadcast-in-belgium-will-have-a-new-classification-system-from-2020.html) but could find no official government publication.

Emby.Server.Implementations/Data/SqliteItemRepository.cs
```diff
@@ -4020,34 +4020,116 @@ private List<string> GetWhereClauses(InternalItemsQuery query, IStatement statem
                 whereClauses.Add(clause);
             }
 
-            if (query.MinParentalRating.HasValue)
+            var ratingClause = "(";
+            if (query.HasParentalRating.HasValue && query.HasParentalRating.Value)
```
daullmer, line 4024:
```suggestion
            if (query.HasParentalRating ?? false)
```

Emby.Server.Implementations/Data/SqliteItemRepository.cs
```diff
@@ -4020,34 +4020,116 @@ private List<string> GetWhereClauses(InternalItemsQuery query, IStatement statem
...
+                }
 
-            if (query.HasParentalRating.HasValue)
+                if (query.MinParentalRating.HasValue)
+                {
+                    ratingClause += "InheritedParentalRatingValue >= @MinParentalRating";
+                    if (statement != null)
+                    {
+                        statement.TryBind("@MinParentalRating", query.MinParentalRating.Value);
+                    }
```
daullmer, line 4071:
```suggestion
                    statement?.TryBind("@MinParentalRating", query.MinParentalRating.Value);
```

Emby.Server.Implementations/Data/SqliteItemRepository.cs
```diff
@@ -4020,34 +4020,116 @@ private List<string> GetWhereClauses(InternalItemsQuery query, IStatement statem
...
+                    if (query.MinParentalRating.HasValue)
+                    {
+                        ratingClause += " AND ";
+                    }
+
+                    ratingClause += "InheritedParentalRatingValue <= @MaxParentalRating";
+                    if (statement != null)
+                    {
+                        statement.TryBind("@MaxParentalRating", query.MaxParentalRating.Value);
+                    }
```
daullmer, line 4085:
```suggestion
                    statement?.TryBind("@MaxParentalRating", query.MaxParentalRating.Value);
```

Emby.Server.Implementations/Data/SqliteItemRepository.cs
```diff
@@ -4020,34 +4020,116 @@ private List<string> GetWhereClauses(InternalItemsQuery query, IStatement statem
...
+                }
+            }
+            else if (query.MinParentalRating.HasValue)
             {
-                if (query.HasParentalRating.Value)
+                ratingClause += "InheritedParentalRatingValue is null OR (InheritedParentalRatingValue >= @MinParentalRating";
+                if (statement != null)
                 {
-                    whereClauses.Add("InheritedParentalRatingValue > 0");
+                    statement.TryBind("@MinParentalRating", query.MinParentalRating.Value);
```
daullmer, line 4103: `statement?.`...

Emby.Server.Implementations/Data/SqliteItemRepository.cs
```diff
@@ -4020,34 +4020,116 @@ private List<string> GetWhereClauses(InternalItemsQuery query, IStatement statem
...
+            }
+            else if (query.MaxParentalRating.HasValue)
+            {
+                ratingClause += "InheritedParentalRatingValue is null OR InheritedParentalRatingValue <= @MaxParentalRating";
+                if (statement != null)
+                {
+                    statement.TryBind("@MaxParentalRating", query.MaxParentalRating.Value);
+                }
+            }
+            else if (query.HasParentalRating.HasValue && !query.HasParentalRating.Value)
```
daullmer, line 4125:
```suggestion
            else if (!query.HasParentalRating ?? false)
```

Jellyfin.Api/Controllers/UserLibraryController.cs
```diff
@@ -162,7 +179,29 @@ public async Task<ActionResult<QueryResult<BaseItemDto>>> GetIntros([FromRoute,
...
+        {
+            return NotFound();
+        }
+
+        if (item is not UserRootFolder
+            // Check the item is visible for the user
+            && !item.IsVisible(user))
+        {
+            return Unauthorized($"{user.Username} is not permitted to access item {item.Name}.");
+        }
```
1337joe, line 202: This logic is shared between 4 methods, is there a reasonable way to extract it to a helper method to share code? Perhaps using an out parameter to pass back the return result if access is denied?

Jellyfin.Server/Migrations/Routines/MigrateRatingLevels.cs
```diff
@@ -0,0 +1,85 @@
...
+        {
+            _logger = logger;
+            _paths = paths;
+        }
+
+        /// <inheritdoc/>
+        public Guid Id => Guid.Parse("{67445D54-B895-4B24-9F4C-35CE0690EA07}");
+
+        /// <inheritdoc/>
+        public string Name => "RemoveDuplicateExtras";
```
1337joe, line 30:
```suggestion
        public string Name => "MigrateRatingLevels";
```

Jellyfin.Api/Controllers/UserLibraryController.cs
Shadowghost: I played around with it and it doesn't feel like a proper solution because the out parameters (you'd need one for the item and one for the user) can be null. You'd either need additional null checks or explicitly tell the compiler that the variable is not null there which feels kinda dirty.

```c#
    private ActionResult? CheckAccess (Guid userId, Guid itemId, out BaseItem? item, out User? user)
    {
        item = null;
        user = _userManager.GetUserById(userId);
        if (user is null)
        {
            return NotFound();
        }

        item = itemId.Equals(default)
            ? _libraryManager.GetUserRootFolder()
            : _libraryManager.GetItemById(itemId);

        if (item is null)
        {
            return NotFound();
        }

        if (item is not UserRootFolder
            // Check the item is visible for the user
            && !item.IsVisible(user))
        {
            return Unauthorized($"{user.Username} is not permitted to access item {item.Name}.");
        }

        return null;
    }
```

1337joe (approved):

Jellyfin.Api/Controllers/UserLibraryController.cs
1337joe: Fair enough, I suspected you hadn't done it because it was too unwieldy but I didn't have a dev environment up to play with it and couldn't immediately see why it wouldn't work out.

Emby.Server.Implementations/Data/SqliteItemRepository.cs
```diff
@@ -3674,12 +3671,8 @@ private List<string> GetWhereClauses(InternalItemsQuery query, IStatement statem
...
             {
                 whereClauses.Add("(CleanName like @NameContains or OriginalTitle like @NameContains)");
-                if (statement is not null)
-                {
-                    nameContains = FixUnicodeChars(nameContains);
-
-                    statement.TryBind("@NameContains", "%" + GetCleanValue(nameContains) + "%");
-                }
+                nameContains = FixUnicodeChars(nameContains);
+                statement?.TryBind("@NameContains", "%" + GetCleanValue(nameContains) + "%");
```
Bond-009, line 3675: Wrap this in a null check again, no need to call `FixUnicodeChars` if we won't use the result

Bond-009 (approved):

Emby.Server.Implementations/Data/SqliteItemRepository.cs
```diff
@@ -3622,12 +3623,8 @@ private List<string> GetWhereClauses(InternalItemsQuery query, IStatement statem
...
                         .Append(paramName)
                         .Append("))) OR ");
-
-                    if (statement is not null)
-                    {
-                        query.PersonIds[i].TryWriteBytes(idBytes);
-                        statement.TryBind(paramName, idBytes);
-                    }
+                    query.PersonIds[i].TryWriteBytes(idBytes);
+                    statement?.TryBind(paramName, idBytes);
```
Bond-009, line 3627: Wrap in if again

Bond-009 (changes requested):

Bond-009 (approved):

gitdeath:
This appears to fix my problem with series that have specific episodes rated lower than the series as a whole. Example: a TV-MA series with one episode showing as PG and that episode being available (to my mind improperly) to people with PG as max. 

Am I correct that this resolves? If not, is there something I can provide to get this included?

Also, any idea when this will go live because it is a huge QoL improvement as I'm manually catching these now.

Lastly thank you for the great work done on this overhaul already. Amazing.

SteveDinn:
This PR was merged back in February. It's been "live" for quite a while now. Just upgrade to the latest version and you should have it.

nielsvanvelzen:
The changes in this PR will be available in 10.9 which is not released yet. They are not included in the current version (10.8.10).

Shadowghost:
That is incorrect. It's been merged to master and therefore available in the unstable builds. No new major version was released since then, only patch releases for 10.8.
This PR will be part of 10.9

SteveDinn:
Whoops. I thought I had it! I stand corrected.

jussme:
I know this is not the official image but i tried the nightly tag [linuxserver/jellyfin:nightly-20231115.15-unstable-ls48](https://hub.docker.com/layers/linuxserver/jellyfin/nightly-20231115.15-unstable-ls48/images/sha256-77e05575b05a9f3417d95bb5d85729862a2a4611d9315f36bdfa0aea0d89f627?context=explore) (10.9.0), and it seems that this PR does not fix https://github.com/jellyfin/jellyfin/issues/6046

Shadowghost:
Please try with a current master build and rescan once. If it is still broken, we can reopen the issue.

jussme:
Stayed on the above mentioned tag. It seems the rescan was indeed needed. Tag exclusion works great, thanks for your work.

Wr0ngName:
Hi,

Running version 10.8.13 here (in docker). The issue #3636 is still a current one.

crobibero:
> Hi,
> 
> Running version 10.8.13 here (in docker). The issue #3636 is still a current one.

This PR is not in 10.8.13.

Wr0ngName:
> > Hi,
> > Running version 10.8.13 here (in docker). The issue #3636 is still a current one.
> 
> This PR is not in 10.8.13.

You're right (facepalm). Thanks for the quick reply. I will wait for 10.9 !

PR #5735, opened 2021-04-09
Happy Eyeballs (IP6 fallback to IP4) implementation of http client.

BaronGreenback:
Implementation of the default http client implementing happy eyeballs. https://github.com/jellyfin/jellyfin/issues/1002

Default implementation now uses happy eyeballs. (IP6 with IP4 fallback).

1) This implementation will cease using IPv6 in case of exception. 
If this exception is due to physical network changes NetworkManager will reset the IP6 capability, which will re-enable the functionality.

2) IP6 functionality is controlled via the network.xml. If, for some reason, NetworkManager hasn't been instantiated, IPv4 will be used. (thinking of testing)

3) This implementation is ignored by the DLNA sub-system, which continues to use single attempt due to it accessing IP addresses.

Issue #1002, opened 2019-02-25
Jellyfin does not retry IPv4 on IPv6 failure


brianjmurrell:
Getting this trying to view the Plugin Catalog on 10.2.1:
```
[2019-02-25 05:25:48.807 -05:00] [INF] "HTTP GET" "https://videoserver:8920/emby/Packages?TargetSystems=Server&IsAdult=false&IsAppStoreSafe=true&PackageType=UserInstalled". UserAgent: "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/72.0.3626.119 Safari/537.36"
[2019-02-25 05:25:48.825 -05:00] [INF] HttpClientManager "GET": "https://repo.jellyfin.org/releases/plugin/manifest.json"
[2019-02-25 05:25:48.827 -05:00] [INF] "HTTP GET" "https://videoserver:8920/emby/Plugins". UserAgent: "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/72.0.3626.119 Safari/537.36"
[2019-02-25 05:25:48.828 -05:00] [INF] HTTP Response 200 to "fd31:aeb1:48df:0:3b14:e643:83d8:7017". Time: "1""ms". "https://videoserver:8920/emby/Plugins" ""
[2019-02-25 05:26:08.828 -05:00] [ERR] Error Timeout getting response from "https://repo.jellyfin.org/releases/plugin/manifest.json"
System.Net.WebException: The operation has timed out.
   at System.Net.HttpWebRequest.EndGetResponse(IAsyncResult asyncResult)
   at System.Threading.Tasks.TaskFactory`1.FromAsyncCoreLogic(IAsyncResult iar, Func`2 endFunction, Action`1 endAction, Task`1 promise, Boolean requiresSynchronization)
[2019-02-25 05:26:08.853 -05:00] [ERR] Error processing request
MediaBrowser.Model.Net.HttpException: The operation has timed out. ---> System.Net.WebException: The operation has timed out.
   at System.Net.HttpWebRequest.EndGetResponse(IAsyncResult asyncResult)
   at System.Threading.Tasks.TaskFactory`1.FromAsyncCoreLogic(IAsyncResult iar, Func`2 endFunction, Action`1 endAction, Task`1 promise, Boolean requiresSynchronization)
   --- End of inner exception stack trace ---
[...]

PR #8598, opened 2022-10-20
Happy Eyeballs (IP6 fallback to IP4) implementation of http client.

Neuheit:
**Changes**
Continuation of #5735 
Made the IPv6 property threadsafe. 

**Issues**
#5735 
[#1002](https://github.com/jellyfin/jellyfin/issues/1002)

### DIFF
--- a/Jellyfin.Networking/HappyEyeballs/HttpClientExtension.cs
+++ b/Jellyfin.Networking/HappyEyeballs/HttpClientExtension.cs
@@ -36,6 +36,11 @@ namespace Jellyfin.Networking.HappyEyeballs
         private const int ConnectionEstablishTimeout = 2000;
 
         /// <summary>
+        /// Interlocked doesn't support bool types.
+        /// </summary>
+        private static int _useIPv6 = 0;
+
+        /// <summary>
         /// Gets a value indicating whether the initial IPv6 check has been performed (to determine whether v6 is available or not).
         /// </summary>
         private static bool _hasResolvedIPv6Availability;
@@ -43,7 +48,21 @@ namespace Jellyfin.Networking.HappyEyeballs
         /// <summary>
         /// Gets or sets a value indicating whether IPv6 should be preferred. Value may change based on runtime failures.
         /// </summary>
-        public static bool? UseIPv6 { get; set; } = null;
+        public static bool UseIPv6
+        {
+            get => Interlocked.CompareExchange(ref _useIPv6, 1, 1) == 1;
+            set
+            {
+                if (value)
+                {
+                    Interlocked.CompareExchange(ref _useIPv6, 1, 0);
+                }
+                else
+                {
+                    Interlocked.CompareExchange(ref _useIPv6, 0, 1);
+                }
+            }
+        }
 
         /// <summary>
         /// Implements the httpclient callback method.
@@ -56,7 +75,7 @@ namespace Jellyfin.Networking.HappyEyeballs
             // Until .NET supports an implementation of Happy Eyeballs (https://tools.ietf.org/html/rfc8305#section-2),
             // let's make IPv4 fallback work in a simple way. This issue is being tracked at https://github.com/dotnet/runtime/issues/26177
             // and expected to be fixed in .NET 6.
-            if (UseIPv6 == true)
+            if (UseIPv6)
             {
                 try
                 {
--- a/Jellyfin.Networking/HappyEyeballs/HttpClientExtension.cs
+++ b/Jellyfin.Networking/HappyEyeballs/HttpClientExtension.cs
@@ -1,3 +1,29 @@
+/*
+The MIT License (MIT)
+
+Copyright (c) .NET Foundation and Contributors
+
+All rights reserved.
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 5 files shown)


Jellyfin.Networking/HappyEyeballs/HttpClientExtension.cs
```diff
@@ -0,0 +1,136 @@
...
+
+        /// <summary>
+        /// Implements the httpclient callback method.
+        /// </summary>
+        /// <param name="context">The <see cref="SocketsHttpConnectionContext"/> instance.</param>
+        /// <param name="cancellationToken">The <see cref="CancellationToken"/> instance.</param>
+        /// <returns>The http steam.</returns>
+        public static async ValueTask<Stream> OnConnect(SocketsHttpConnectionContext context, CancellationToken cancellationToken)
+        {
+            // Until .NET supports an implementation of Happy Eyeballs (https://tools.ietf.org/html/rfc8305#section-2),
```
cvium, line 75: I have been thinking about this implementation today and read the old dotnet issue. This implementation doesn't really follow the RFC spec as the RFC states that both IPv6 and IPv4 should be called in parallel and whichever finishes first wins. This tries IPv6 first and then IPv4. And should IPv6 be disabled by one failure or is it more nuanced?

Neuheit: That is a good point I will get that changed.

Jellyfin.Networking/HappyEyeballs/HttpClientExtension.cs
```diff
@@ -72,45 +28,46 @@ public static bool UseIPv6
...
-                        using var linkedTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, quickFailCts.Token);
+            using var cancelIPv6 = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+            var tryConnectAsyncIPv6 = AttemptConnection(AddressFamily.InterNetworkV6, context, cancelIPv6.Token);
 
-                        localToken = linkedTokenSource.Token;
-                    }
+            if (await Task.WhenAny(tryConnectAsyncIPv6, Task.Delay(200, cancelIPv6.Token)).ConfigureAwait(false) == tryConnectAsyncIPv6 && tryConnectAsyncIPv6.IsCompletedSuccessfully)
+            {
+                cancelIPv6.Cancel();
+                return tryConnectAsyncIPv6.GetAwaiter().GetResult();
```
cvium, line 42: Why GetAwaiter.GetResult instead of await?

Jellyfin.Networking/HappyEyeballs/HttpClientExtension.cs
```diff
@@ -33,36 +13,12 @@ namespace Jellyfin.Networking.HappyEyeballs
...
-                {
-                    Interlocked.CompareExchange(ref _useIPv6, 1, 0);
-                }
-                else
-                {
-                    Interlocked.CompareExchange(ref _useIPv6, 0, 1);
-                }
-            }
-        }
+        // Implementation taken from https://github.com/rmkerr/corefx/blob/SocketsHttpHandler_Connect_HappyEyeballs/src/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/ConnectHelper.cs
```
cvium, line 21: You'll need to include the MIT license at the top of the file

cvium (changes requested):

Jellyfin.Networking/HappyEyeballs/HttpClientExtension.cs
Neuheit: The task has already completed so it won't have any performance effect. It also apparently is better at handling exceptions according to [this comment](https://github.com/dotnet/corefx/pull/29792/files#r189415885).

cvium: Sorry for letting it sit for so long. Can you add a code comment explaining this?

Jellyfin.Networking/HappyEyeballs/HttpClientExtension.cs
```diff
@@ -0,0 +1,122 @@
...
+            {
+                return await AttemptConnection(AddressFamily.InterNetwork, context, cancellationToken).ConfigureAwait(false);
+            }
+
+            using var cancelIPv6 = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+            var tryConnectAsyncIPv6 = AttemptConnection(AddressFamily.InterNetworkV6, context, cancelIPv6.Token);
+
+            //This connect task uses GetAwaiter().GetResult() as the appropriate task has already been completed.
+            //This results in improved exception handling.
+            //See https://github.com/dotnet/corefx/pull/29792/files#r189415885 for more details.
```
cvium, line 65:
```suggestion
            // GetAwaiter().GetResult() is used instead of .Result as this results in improved exception handling.
            // The tasks have already been completed.
            // See https://github.com/dotnet/corefx/pull/29792/files#r189415885 for more details.
```

Jellyfin.Networking/HappyEyeballs/HttpClientExtension.cs
```diff
@@ -0,0 +1,122 @@
...
+                cancelIPv6.Cancel();
+                return tryConnectAsyncIPv6.GetAwaiter().GetResult();
+            }
+
+            using var cancelIPv4 = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+            var tryConnectAsyncIPv4 = AttemptConnection(AddressFamily.InterNetwork, context, cancelIPv4.Token);
+
+            //Both connect tasks use GetAwaiter().GetResult() as the appropriate task has already been completed.
+            //This results in improved exception handling.
+            if (await Task.WhenAny(tryConnectAsyncIPv6, tryConnectAsyncIPv4).ConfigureAwait(false) == tryConnectAsyncIPv6)
```
cvium, line 77: One comment should suffice
```suggestion
            if (await Task.WhenAny(tryConnectAsyncIPv6, tryConnectAsyncIPv4).ConfigureAwait(false) == tryConnectAsyncIPv6)
```

Jellyfin.Server/Startup.cs
```diff
@@ -79,6 +81,13 @@ public void ConfigureServices(IServiceCollection services)
             var acceptJsonHeader = new MediaTypeWithQualityHeaderValue(MediaTypeNames.Application.Json, 1.0);
             var acceptXmlHeader = new MediaTypeWithQualityHeaderValue(MediaTypeNames.Application.Xml, 0.9);
             var acceptAnyHeader = new MediaTypeWithQualityHeaderValue("*/*", 0.8);
+            Func<IServiceProvider, HttpMessageHandler> eyeballsHttpClientHandlerDelegate = (_) => new SocketsHttpHandler()
+            {
+                AutomaticDecompression = DecompressionMethods.All,
+                RequestHeaderEncodingSelector = (_, _) => Encoding.UTF8,
+                ConnectCallback = HttpClientExtension.OnConnect
```
cvium, line 88: Would it make sense to make it configurable?

Neuheit: I didn't think so since the default client is not configurable and the callback needs to be set the HttpClientExtension, since it contains the happy eyeballs protocol.

cvium (approved):

Shadowghost (approved):

crobibero (approved):

Issue #7783, opened 2022-05-22
[Issue][10.8.X]: Dashboard causes high CPU usage and thread multiplication


yshsy:
### Please describe your bug

At seemingly random occurences, one of the Jellyfin threads will climb up to very high CPU usage (usually between 200 and 500% usage) while many others start to appear (usually 20-35 more of them). The issue is clearly visible on the screenshot attached to this issue.

This causes Jellyfin to become slow to the point of being completely unusable, freezing playback and the web interface.

The Jellyfin log doesn't seem to contain any relevant informations or warnings / errors. An anonymised snippet of this log right before the freezing can also be found below.

Restarting the Jellyfin service temporarly solves the issue, but it usually comes back relatively soon.

### Jellyfin Version

Other

### if other:

10.8.0 Beta 3

### Environment

```markdown
- OS:Debian 11
- Virtualization: No
- Clients:Any
- Browser:Any
- FFmpeg Version: 4.4.1
- Playback Method: Remux
- Hardware Acceleration: NVENC
- Plugins: 
- Reverse Proxy: apache
- Base URL: none
- Networking: Host
- Storage: local (ZFS)
```

### Jellyfin logs

```shell
# Sensitive data was redacted using {}

[2022-05-22 02:50:13.342 +02:00] [WRN] Slow HTTP Response from "http://{SERVER_URL}/Sessions/Playing/Progress" to "{IP}" in 0:00:00.6531287 with Status Code 204
[2022-05-22 02:50:16.826 +02:00] [INF] Processing playback tracker : "TW96aWxsYS81LjAgKFgxMTsgTGludXggeDg2XzY0KSBBcHBsZVdlYktpdC81MzcuMzYgKEtIVE1MLCBsaWtlIEdlY2tvKSBDaHJvbWUvMTAxLjAuNDk1MS42NCBTYWZhcmkvNTM3LjM2
IE9QUi84Ny4wLjQzOTAuMjV8MTY1MzE3Njc5MzU3Nw11-ad6f2a8fb5db4dac940d973094f18d00-a3eee41e85d2bb179da1938f5521d84d"
[2022-05-22 02:50:17.368 +02:00] [WRN] Slow HTTP Response from "http://{SERVER_URL}/Sessions/Playing/Progress" to "{IP}" in 0:00:00.5738518 with Status Code 204
[...]

PR #8601, opened 2022-10-21
remove JellyfinDbProvider and add second level caching

cvium:
**Changes**
Removed the JellyfinDbProvider and switched it to a pooled context factory. I don't think JellyfinDbProvider used the context pool that was configured previously. But the provider is also completely unnecessary now that DbContextFactory is a thing.
Added second level caching as api key authenticated requests would pull from the Device table every time.

**Issues**
Might be related #7783


crobibero (approved):

dot-i:
Just wanted to leave some feedback that this apparently doesn't solve the slow HTTP response during scans. I have a test-VM on which I can easily reproduce the slow UI problem by starting a scan of my full Shows library.

I see no difference when this patch is applied 😞 

That being said: this still looks like a good improvement regardless, it just doesn't seem like it will fix the slow or hanging UI some of us are seeing during library scans...

cvium:
It's not meant to fix that. It's an entirely different database.

Bond-009 (approved):

PR #4114, opened 2020-09-10
Add new files to rpm build

crobibero:
No idea if this will work, just based it on the other lines in the `%files server` block

PR #8638, opened 2022-10-28
Fix extra type differentiation

photonconvergence:
Change extra parsing rules for Featurettes and Shorts so they don't both get classed as ExtraType.Clip.

Fix test broken by these changes.

**Changes**
Change parsing rules for extras labeled "short" and "featurette." Both were previously classed as type "Clip" 
Add two new Extra types (short, featurette)
Add parsing rule to allow "clip" extras in folder form. Now accommodates all supported extra types in either suffix or folder form.

**Issues**
Fixes [#4114](https://github.com/jellyfin/jellyfin-web/issues/4114)

**Other**
PR for user facing portion: [4115](https://github.com/jellyfin/jellyfin-web/pull/4115)

### DIFF
--- a/Emby.Naming/Common/NamingOptions.cs
+++ b/Emby.Naming/Common/NamingOptions.cs
@@ -512,13 +512,13 @@ namespace Emby.Naming.Common
                     MediaType.Video),
 
                 new ExtraRule(
-                    ExtraType.Clip,
+                    ExtraType.Short,
                     ExtraRuleType.DirectoryName,
                     "shorts",
                     MediaType.Video),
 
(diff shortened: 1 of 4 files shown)


crobibero (approved):

cvium (approved):

1337joe:
Probably worth updating the docs to include "clips" in the list of folder types as well:
https://github.com/jellyfin/jellyfin.org/blob/master/docs/general/server/media/movies.md

1337joe (approved):

Bond-009 (approved):

Issue #8587, opened 2022-10-19
[Issue]: Incompatible with firewalld version 1.1.0


lopopylopy:
### Please describe your bug

In Fedora Linux, when trying to use the command `sudo dnf system-upgrade download --releasever=37` to upgrade to Fedora 37, a transaction error occurs.

> Error: Transaction test error: file /usr/lib/firewalld/services/jellyfin.xml from install of firewalld-1.2.1-1.fc37.n

The new version of firewalld in Fedora 37 "now ships a service definition file ... which is in conflict with the one shipped by Jellyfin" ([chrisawi](https://www.reddit.com/r/Fedora/comments/y89ttc/comment/isznvo8/?utm_source=share&utm_medium=web2x&context=3))

This prevents `system-upgrade` to actually upgrade the system unless the user is to permanently delete the Jellyfin package. Fedora 37 is still in beta, but this is potentially critical when it releases on October 25.

### Jellyfin Version

Other

### if other:

10.8.5

### Environment

```markdown
- OS: Fedora
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

yodatak:
Same for me

yodatak:
We should also bump all the reference of fedora 36 to 37 in [deployment/Dockerfile.fedora.amd64](https://github.com/jellyfin/jellyfin/blob/b1e0b4ee8940a7a38c766dff7a762d6e29b31412/deployment/Dockerfile.fedora.amd64)

and https://github.com/jellyfin/jellyfin/blob/c7edf8b565e7d52400cf65e747a8e800ecd3a036/fedora/jellyfin.spec

yodatak:
I cannot install after the upgrade 
Erreur : Erreur de la transaction de test :
  le fichier /usr/lib/firewalld/services/jellyfin.xml de l'installation de jellyfin-server-10.8.5-1.fc36.x86_64 entre en conflit avec le fichier du paquet firewalld-1.2.1-1.fc37.noarch

yodatak:
I try to make a fix here but as not tested yet https://github.com/jellyfin/jellyfin/pull/8644

yodatak:
[...]

PR #8644, opened 2022-10-28
Fix depencies issues for fedora 37

yodatak:
A try to fix issues #8587

### DIFF
--- a/deployment/Dockerfile.fedora.amd64
+++ b/deployment/Dockerfile.fedora.amd64
@@ -1,4 +1,4 @@
-FROM fedora:36
+FROM fedora:37
 # Docker build arguments
 ARG SOURCE_DIR=/jellyfin
 ARG ARTIFACT_DIR=/dist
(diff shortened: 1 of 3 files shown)


joshuaboniface (approved):
Looks good - tried to target the release branch but that didn't work - we'll just have to leave this until 10.9.0.

jlost:
Should I take this PR to mean Jellyfin doesn't work on Fedora 37 yet?

mihawk90:
> Should I take this PR to mean Jellyfin doesn't work on Fedora 37 yet?

Jellyfin itself should™ "just work"™, the problem here is the package. Theoretically you can just extract the contents of the package and run it that way.
Note: I don't advise this as it can cause conflicts when the package eventually becomes available, but it is an option for experienced users.

Shadowghost:
The issue this was meant to fix was fixed in upstream firewalld.

PR #8646, opened 2022-10-28
Remove firewalld spec and target Fedora 37

joshuaboniface:
Fedora 37 ships with this firewalld configuration by default, and thus our package conflicts with it.

For older releases, it should be added manually (and I don't believe those upgrading would have it removed) but I expect given Fedora's release schedule that most everyone will upgrade soon.

At the same time, upgrade the build Dockerfile to target 37 to ensure consistency.

### DIFF
--- a/deployment/Dockerfile.fedora.amd64
+++ b/deployment/Dockerfile.fedora.amd64
@@ -1,4 +1,4 @@
-FROM fedora:36
+FROM fedora:37
 # Docker build arguments
 ARG SOURCE_DIR=/jellyfin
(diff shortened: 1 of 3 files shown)


thornbill (approved):

joshuaboniface:
This breaks for some reason, closing for now. Will fix for 10.9.0

Oreloth:
Could you apply same fix for AlmaLinux/Rocky 9? They both have firewalld conflict, and some of my services needs firewalld to run properly.

PR #8703, opened 2022-11-06
Calculate output audio bitrate from output channel count

knackebrot:
384kbps is unnecessarily high for stereo output with our encoders and sending data over network is expensive. This commit introduces logic for calculating output bitrate based on output channel count, like this:
```csharp
(inputChannels, outputChannels) switch
{
    ( >= 6, >= 6 or 0) => Math.Min(640000, audioBitRate.Value),
    ( > 0, > 0) => Math.Min(outputChannels * 128000, audioBitRate.Value),
    ( > 0, _) => Math.Min(inputChannels * 128000, audioBitRate.Value),
    (_, _) => Math.Min(384000, audioBitRate.Value)
}
```

That means transcoded stereo output in hls is going to be **at most** 256kbps, mono 128kbps, surround 640kbps, etc.

The second commit adds VBR mode calculation from the output bitrate based on expected upper-bound average bitrate for high-quality audio. This is user-configurable and disabled by default.

Continued from https://github.com/jellyfin/jellyfin/pull/8113

### DIFF
--- a/Jellyfin.Api/Controllers/DynamicHlsController.cs
+++ b/Jellyfin.Api/Controllers/DynamicHlsController.cs
@@ -1685,9 +1685,20 @@ public class DynamicHlsController : BaseJellyfinApiController
 
             audioTranscodeParams += "-acodec " + audioCodec;
 
-            if (state.OutputAudioBitrate.HasValue)
+            var audioBitrate = state.OutputAudioBitrate;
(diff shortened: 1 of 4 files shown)


Shadowghost:
Would be nice if you could rebase on master.

knackebrot:
@Shadowghost Done.

Bond-009:
Included in #9388

Issue #8718, opened 2022-11-09
Library access controls do not prevent API listing of all server content


thornbill:
### Please describe your bug

Limiting a user's access to libraries does not prevent them from discovering and listing the contents of all libraries on a server via the API. Similarly disabling live TV access does not prevent a user from listing all TV channels via the API.

1. Disable a user's access to all libraries and live TV in the admin dashboard.
2. Verify that the `Policy` returned by `/Users/{USER_ID}` includes:
   * `EnableAllFolders: false`
   * `EnabledFolders: []`
   * `EnableLiveTvAccess: false`
3. Make a request to `/Library/MediaFolders` and observe that a list of ALL libraries are returned.
4. Using a library ID returned in the previous step, make a request to `/Users/{USER_ID}/Items?ParentId={LIBRARY_ID}` and observe that a list of ALL library contents are returned.
5. Make a request to `/LiveTv/Channels` and observe that a list of ALL channels are returned.

### Jellyfin Version

likely all, tested on 10.8.1, 10.8.7, and unstable

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

VideoFX:
Oh I got this too, i posted this before i noticed your post: https://github.com/jellyfin/jellyfin/issues/8730

PR #8732, opened 2022-11-12
Fix library access

thornbill:
**Changes**
Fixes the access control issues listed in #8718 in the `/Library/MediaFolders` and `/Users/{USER_ID}/Items` endpoints.

The Live TV access issue seems a bit more complex...

**Issues**
Partial fix for #8718

### DIFF
--- a/Jellyfin.Api/Controllers/ItemsController.cs
+++ b/Jellyfin.Api/Controllers/ItemsController.cs
@@ -282,39 +282,13 @@ namespace Jellyfin.Api.Controllers
                 includeItemTypes = new[] { BaseItemKind.Playlist };
             }
 
-            var enabledChannels = isApiKey
-                ? Array.Empty<Guid>()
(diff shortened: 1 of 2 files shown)


VideoFX:
Thank you @thornbill, and excellent work.

cvium (approved):
The Channel's IsVisible implementation: https://github.com/jellyfin/jellyfin/blob/7fdc0e3c3d65d2d408ce036063183b85d4bdb2e3/MediaBrowser.Controller/Channels/Channel.cs#L26

thornbill:
Yeah I saw the IsVisible implementation but the way the query is built I didn't see a good way to filter the results and still return valid counts. There is also seemingly identical functionality in the ChannelsController that does appear to check for visibility.

crobibero (approved):

Bond-009 (approved):

PR #8734, opened 2022-11-12
Add admin login to QuickConnect API

nielsvanvelzen:
**Changes**
- Change QC initiate to use HTTP POST, keep the old GET method for compat
- ~~Add QC revoke endpoint, requested by @thornbill~~
- Add userId parameter to QC authorize for jellyfin/jellyfin-web#3918

API changes are compatible with existing client implementations

**Issues**

API changes for https://github.com/jellyfin/jellyfin-web/issues/3918

### DIFF
--- a/Jellyfin.Api/Controllers/QuickConnectController.cs
+++ b/Jellyfin.Api/Controllers/QuickConnectController.cs
@@ -52,7 +52,7 @@ namespace Jellyfin.Api.Controllers
         /// <response code="200">Quick connect request successfully created.</response>
         /// <response code="401">Quick connect is not active on this server.</response>
         /// <returns>A <see cref="QuickConnectResult"/> with a secret and code for future use or an error message.</returns>
-        [HttpGet("Initiate")]
+        [HttpPost("Initiate")]
         [ProducesResponseType(StatusCodes.Status200OK)]
         public async Task<ActionResult<QuickConnectResult>> InitiateQuickConnect()
         {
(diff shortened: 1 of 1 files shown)


cvium:
I think the revoke endpoint is unnecessary. What is it supposed to solve, @thornbill?

thornbill:
I agree... after considering this further I don't think the revoke endpoint is necessary. The issue is that web allows you to generate multiple QC tokens which is not a great experience. I think we can just handle that in web though since the tokens expire after a short time.

nielsvanvelzen:
Updated the PR:
- Remove revoke endpoint
- Rebase on top of master
- Update the GET->POST change so the old endpoint is not added to the openapi doc

Changes might not create a merge conflict with #8812 but depending on which is merged first, the other needs updating.

Jellyfin.Api/Controllers/QuickConnectController.cs
```diff
@@ -66,6 +67,16 @@ public async Task<ActionResult<QuickConnectResult>> InitiateQuickConnect()
...
         }
 
+        /// <summary>
+        /// Old version of <see cref="InitiateQuickConnect" /> using a GET method.
+        /// Still available to avoid breaking compatibility.
+        /// </summary>
+        /// <returns>The result of <see cref="InitiateQuickConnect" />.</returns>
+        [Obsolete("Use POST request instead")]
+        [HttpGet("Initiate")]
+        [ApiExplorerSettings(IgnoreApi = true)]
```
crobibero, line 77: Won't removing this from the api spec be a breaking change for the generated sdks?

nielsvanvelzen: No because it uses the same name for the name method, so when using an old SDK version it will use the GET request and with newer SDK the POST request.

crobibero (approved):

Shadowghost (approved):

Bond-009 (approved):

PR #8547, opened 2022-10-13
Target .NET 7

Bond-009:

PR #8745, opened 2022-11-14
Add missing format providers (fix CA1305 errors)

Terrance:
This resolves the following compilation errors (originally encountered whilst trying to build for Arch Linux using an [AUR package](https://aur.archlinux.org/packages/jellyfin), but can be reproduced using the repo's own build script with the dpkg check commented out):

> Jellyfin.Api/Controllers/DisplayPreferencesController.cs(181,29): error CA1305: The behavior of '{0}' could vary based on the current user's locale settings. Provide a value for the 'IFormatProvider' argument. [Jellyfin.Api/Jellyfin.Api.csproj]
> Jellyfin.Api/Helpers/StreamingHelpers.cs(354,24): error CA1305: The behavior of '{0}' could vary based on the current user's locale settings. Provide a value for the 'IFormatProvider' argument. [Jellyfin.Api/Jellyfin.Api.csproj]
> Jellyfin.Server/Program.cs(614,30): error CA1305: The behavior of '{0}' could vary based on the current user's locale settings. Provide a value for the 'IFormatProvider' argument. [Jellyfin.Server/Jellyfin.Server.csproj]
> Jellyfin.Server/Program.cs(616,41): error CA1305: The behavior of '{0}' could vary based on the current user's locale settings. Provide a value for the 'IFormatProvider' argument. [Jellyfin.Server/Jellyfin.Server.csproj]

I've used `CultureInfo.InvariantCulture` as that seems to be used everywhere else, though I'm not familiar enough with the codebase to know if a better formatter would be more appropriate in these cases.

### DIFF
--- a/Jellyfin.Api/Controllers/DisplayPreferencesController.cs
+++ b/Jellyfin.Api/Controllers/DisplayPreferencesController.cs
@@ -178,7 +178,7 @@ namespace Jellyfin.Api.Controllers
 
             foreach (var key in displayPreferences.CustomPrefs.Keys.Where(key => key.StartsWith("homesection", StringComparison.OrdinalIgnoreCase)))
             {
-                var order = int.Parse(key.AsSpan().Slice("homesection".Length));
+                var order = int.Parse(key.AsSpan().Slice("homesection".Length), NumberStyles.Any, CultureInfo.InvariantCulture);
                 if (!Enum.TryParse<HomeSectionType>(displayPreferences.CustomPrefs[key], true, out var type))
                 {
                     type = order < 8 ? defaults[order] : HomeSectionType.None;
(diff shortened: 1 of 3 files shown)


crobibero:
Looks like you're building the server with net7 installed.
All of these changes are already included by https://github.com/jellyfin/jellyfin/pull/8547

Terrance:
Arch is indeed on .NET 7.0 now, though I believe I have the needed 6.0 runtime installed and as far as I can tell it's correctly targeting 6.0 (following `net6.0` in the project files, though forcing it in the `dotnet build` command doesn't help either).

Should I expect it to build correctly without these changes whilst still targeting 6.0?

Bond-009 (approved):

crobibero (approved):

PR #8752, opened 2022-11-15
Don't cancel DVR recordings when adjusting settings

### LINKED ISSUE #3523: LiveTV Recording Series - Adjusting Settings Cancels Future Recordings
**Describe the bug**
When you setup a series to be recorded, it shows that all subsequent episodes will be recorded.  However, if you adjust the series settings (ie: record on time, record 1 minute afterwards, record only new episodes, etc), the subsequent series are no longer showing a red dot and if you find them in the guide it shows they will not be recorded.  

Any adjustments to the settings seems to screw up the future recordings unless you manually set them to record again, which defeats the purpose of telling it to record the series.

 - OS: Windows
 - Virtualization: N/A
 - Clients: AndroidTV
 - Browser: Brave
 - Jellyfin Version: 10.5.5
 - Playback: Direct Play
 - Installed Plugins: Kodi Sync Queue, Playback Reporting, Telegram Notification
 - Reverse Proxy: caddy2
 - Base URL: none
 - Networking: N/A
 - Storage: local

**To Reproduce**

1. Select a TV series to record
2. Go into series settings
3. Adjust any setting (record only new episodes or adjust timing of when recording starts)
4. Go back to viewing anything else in Jellyfin
5. Go back into LIVE TV and view Series and see future recordings are no longer going to record.

**Expected behavior**
When adjusting series settings, future recordings remain selected to record.

**Logs**

**Screenshots**

**Additional context**

SenorSmartyPants:
Timers are getting added twice to the enabled list. This results in the second copy of the timer being detected as a duplicate of the showID and cancelled. Since both timers in the list are the same object, the original record is cancelled as a duplicate. 

I'm not sure exactly why it gets added twice in the first place, so I thought the easiest thing to do was to remove the duplicate timers (not duplicate shows)

Fixes #3523

### DIFF
--- a/Emby.Server.Implementations/LiveTv/EmbyTV/EmbyTV.cs
+++ b/Emby.Server.Implementations/LiveTv/EmbyTV/EmbyTV.cs
@@ -2201,7 +2201,7 @@ namespace Emby.Server.Implementations.LiveTv.EmbyTV
             }
         }
 
-        private void SearchForDuplicateShowIds(List<TimerInfo> timers)
+        private void SearchForDuplicateShowIds(IEnumerable<TimerInfo> timers)
         {
             var groups = timers.ToLookup(i => i.ShowId ?? string.Empty).ToList();
 
@@ -2322,7 +2322,7 @@ namespace Emby.Server.Implementations.LiveTv.EmbyTV
                 }
             }
 
-            SearchForDuplicateShowIds(enabledTimersForSeries.Distinct().ToList());
+            SearchForDuplicateShowIds(enabledTimersForSeries.Distinct());
 
             if (deleteInvalidTimers)
             {
--- a/Emby.Server.Implementations/LiveTv/EmbyTV/EmbyTV.cs
+++ b/Emby.Server.Implementations/LiveTv/EmbyTV/EmbyTV.cs
@@ -2282,39 +2282,13 @@ namespace Emby.Server.Implementations.LiveTv.EmbyTV
 
                     if (updateTimerSettings)
                     {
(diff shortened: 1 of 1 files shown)


SenorSmartyPants:
Suggestion for a JF-Web improvement: If editing series recording settings from the schedule page, reload page to display updated recordings.

crobibero (commented):
I would remove the `ToList()` and change the function signature to 
```c#
private void SearchForDuplicateShowIds(IEnumerable<TimerInfo> timers)
```

SenorSmartyPants:
> I would remove the `ToList()` and change the function signature to
> 
> ```cs
> private void SearchForDuplicateShowIds(IEnumerable<TimerInfo> timers)
> ```

Implemented.

crobibero (approved):

cvium:
> Timers are getting added twice to the enabled list. This results in the second copy of the timer being detected as a duplicate of the showID and cancelled. Since both timers in the list are the same object, the original record is cancelled as a duplicate.
> 
> I'm not sure exactly why it gets added twice in the first place, so I thought the easiest thing to do was to remove the duplicate timers (not duplicate shows)

I believe it is caused by a bad merge https://github.com/jellyfin/jellyfin/commit/49b61f238e634e8b2ed4af8a3e0036080cd023a6#diff-45966b4ae922d490eb075c7dd658808a1d38e31ad7ea58299b77c084d6d825d0R2413

cvium (approved):
I assume you have re-tested it

Bond-009 (approved):

PR #8768, opened 2022-11-17
DVR: Prefer HD channels then earliest showing when handling duplicate showings.

SenorSmartyPants:
Current behavior of the scheduler is undefined when handling duplicate showings. Order of timers is unknown. This PR sorts duplicate showings first by HD channels, then by earliest showing. Under the current behavior, my DVR could sometimes wait over a week to record a later episode for no discernable reason. 

Logs more details when scheduling DVR timers.

PR #8777, opened 2022-11-20
Save TVChannel Height if set

SenorSmartyPants:
**Changes**

Saves a height value if set for `LiveTVChannel` items. This allows an admin to specify the video height(ex. 480, 720, 1080) for channels when it is not automatically provided. This will then determine if the channel `isHD` or not.

**Issues**

Useful in conjunction with #8768

### DIFF
--- a/Jellyfin.Api/Controllers/ItemUpdateController.cs
+++ b/Jellyfin.Api/Controllers/ItemUpdateController.cs
@@ -246,6 +246,11 @@ public class ItemUpdateController : BaseJellyfinApiController
             episode.AirsBeforeSeasonNumber = request.AirsBeforeSeasonNumber;
         }
 
+        if (request.Height is not null && item is LiveTvChannel channel)
+        {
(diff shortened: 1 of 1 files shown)


Jellyfin.Api/Controllers/ItemUpdateController.cs
```diff
@@ -246,6 +246,11 @@ private void UpdateItem(BaseItemDto request, BaseItem item)
                 episode.AirsBeforeSeasonNumber = request.AirsBeforeSeasonNumber;
             }
 
+            if (item is LiveTvChannel channel)
+            {
+                channel.Height = request.Height ?? 0;
+            }
```
crobibero, line 252:
```suggestion
            if (request.Height is not null && item is LiveTvChannel channel)
            {
                channel.Height = request.Height.Value;
            }
```

crobibero (approved):

Shadowghost (approved):

Issue #6877, opened 2021-11-20
[Issue]: generating movie library thumbnails uses an crazy amount of memory


Thijzer:
### Please describe your bug

I noticed that when a movie library has to load in it's thumbnails by URL
<DOMAIN>/Items/<ID>/Images/Primary?fillHeight=342&fillWidth=228&quality=96&tag=0619c8b9bbbf8e088fb2f04be75864f0

- it needs to generate them on every new device, no cache pool?
- the generating of the thumbs reserves a lot of memory that doesn't get GB collected quickly

I'm willing to take a look into this, I just don't know why it has been setup this way, if this behavior is expected?

Can someone please give me some hints.

Thanks

### Jellyfin Version

10.7.7

### Environment

```markdown
- OS: 
- Virtualization: docker
- Clients: Brave Browser
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage: via docker volumes native ext4
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cvium:
Define a lot

hannesbraun:
I've had a similar experience using Jellyfin 10.8.4 inside a Docker container running on my Raspberry Pi 2. When the thumbnails are requested and aren't cached yet, I get a huge spike in memory usage that usually ends in Jellyfin crashing because the system's out of memory. To put it in numbers, my memory usage is about 340 MB when Jellyfin is running idle. During those requests, I can see values of up to 800 MB. (I'd guess it's even higher, but the refresh rate of my resource monitor is probably too slow for that.) The thumbnail page contained about 60 images.

> it needs to generate them on every new device, no cache pool?

As far as I was able to understand the current implementation, there is a cache pool. But different devices may require different resolutions. Thus, the images need to be resized.

> the generating of the thumbs reserves a lot of memory that doesn't get GB collected quickly

In the rare cases where Jellyfin didn't crash, the memory got garbage collected rather quickly, however. I didn't notice any significant change in "memory idle usage" after about 2 seconds or something like that.

From what I tried, limiting the number of parallel image encoding tasks should help to avoid those "big" memory spikes.
[...]

PR #8783, opened 2022-11-21
Allow limiting parallel image encodings to reduce memory usage

hannesbraun:
**Changes**
This change allows specifying the environment variable `JELLYFIN_PARALLEL_IMAGE_ENCODING_LIMIT` to limit the number of parallel image encodings as those tend to consume a lot of memory (potentially crashing the process as the system runs out of memory). If set, the number of parallel encodings will be limited to that number. Otherwise, the behavior should be unaffected.

The usage of an environment variable made sense here to me, as this setting conceptually didn't really seem to fit into the regular user settings.

**Issues**
This solves the issue I described in this comment: https://github.com/jellyfin/jellyfin/issues/6877#issuecomment-1257031371


Bond-009:
Use https://github.com/jellyfin/jellyfin/blob/master/Emby.Server.Implementations/ConfigurationOptions.cs instead of getting the env variable directly

hannesbraun:
I'm not quite sure how to use `ConfigurationOptions` directly... However, I found out that I can use an `IConfiguration` to retrieve the environment variable instead. Is this what you meant? I'm sorry for the inconvenience.

Bond-009:
Yes, that's correct. You would use `ConfigurationOptions` to set a default value

hannesbraun:
I see. Thank you. I think a default value shouldn't be required here as the implicit default value `null` is already sufficient and marks the absence of the setting. But if you'd like to have it written explicitly in the `ConfigurationOptions`, I don't have a problem with that either.

crobibero:
I think it is fair to limit the concurrent encodings by default. What value did you find worked best for you?

hannesbraun:
For me, 4 parallel encodings are about the maximum my Raspberry Pi 2 could reliably handle. The default value should probably be a bit higher as this will surely slow down loading a page with a lot of images when your system has more memory available. Maybe something like `2 * Environment.ProcessorCount` makes a good default limit. (Looks like I'm not able to retrieve the amount of physical memory installed.)

crobibero:
`2 * Environment.ProcessorCount` sounds like a good default to me

hannesbraun:
Thanks for the feedback. The default limit should now be implemented with `2 * Environment.ProcessorCount`.

Emby.Drawing/ImageProcessor.cs
```diff
@@ -48,18 +54,23 @@ public sealed class ImageProcessor : IImageProcessor, IDisposable
...
+            IConfiguration config)
         {
             _logger = logger;
             _fileSystem = fileSystem;
             _imageEncoder = imageEncoder;
             _mediaEncoder = mediaEncoder;
             _appPaths = appPaths;
+            _config = config;
+
+            _parallelEncodingLimit = new(_config.GetParallelImageEncodingLimit());
```
cvium, line 73: Isn't this the wrong constructor? Should we not set initial _and_ max to be safe?

cvium:
> The usage of an environment variable made sense here to me, as this setting conceptually didn't really seem to fit into the regular application settings.

Why does it not fit in? Personally, I don't like having configurations split between different providers and the .Net configuration framework is extensible enough that you can provide _any_ configuration key via environment variables.

Emby.Drawing/ImageProcessor.cs
hannesbraun: Yes, that would be better. I updated the PR accordingly.

hannesbraun:
> Why does it not fit in? Personally, I don't like having configurations split between different providers and the .Net configuration framework allows you to provide _any_ configuration key via environment variables.

As far as I know, the regular application settings don't contain any performance tuning parameters. (But I might as well have missed something there.)

Another reason for making it an environment variable was that you could set it without needing to run the Jellyfin server. In some very memory-constrained use cases, a server may already crash depending on the number of images that must be encoded when only accessing the startup screen. That would make it rather hard to navigate to the settings.

But in the end, I think it's also fine to move this setting into the usual application settings. Just let me know what you'd prefer.

hannesbraun:
The setting is now contained in the regular application settings.

Emby.Drawing/ImageProcessor.cs
```diff
@@ -199,8 +208,13 @@ public async Task<(string Path, string? MimeType, DateTime DateModified)> Proces
             {
                 if (!File.Exists(cacheFilePath))
                 {
+                    // Limit number of parallel (more precisely: concurrent) image encodings to prevent a high memory usage
+                    await _parallelEncodingLimit.WaitAsync().ConfigureAwait(false);
+
                     string resultPath = _imageEncoder.EncodeImage(originalImagePath, dateModified, cacheFilePath, autoOrient, orientation, quality, options, outputFormat);
 
+                    _parallelEncodingLimit.Release();
```
cvium, line 216: Shouldn't this be in a try-finally block?

hannesbraun: Yes, it should be in a try-finally block. Fixed.

crobibero (approved):

barronpm (approved):

hannesbraun:
The merge conflict should be resolved now.

crobibero (changes requested):
Upon internal discussion we think it may be better to default the `ParallelImageEncodingLimit` to 0 and if when creating the semaphore assume <1 to be `2 * Environment.ProcessorCount`

hannesbraun:
> Upon internal discussion we think it may be better to default the `ParallelImageEncodingLimit` to 0 and if when creating the semaphore assume <1 to be `2 * Environment.ProcessorCount`

Fixed

Bond-009 (approved):

crobibero (approved):

momala454:
how do I set ParallelImageEncodingLimit ?

hannesbraun:
There's a pending pull request for Jellyfin Web: https://github.com/jellyfin/jellyfin-web/pull/4262

momala454:
i saw it but in the mean time in which file do I set it ?
i tried 
<ParallelImageEncodingLimit>2</ParallelImageEncodingLimit>
inside system.xml to try to reduce memory usage, but it didn't reduced memory usage

momala454:
10.8.9

nielsvanvelzen:
The changes of this pull request will be available in 10.9.

momala454:
hello @nielsvanvelzen do you have an ETA for a possibility to include this into 10.8.* or first alphas for 10.9 ?
As this issue is preventing me to use jellyfin

nielsvanvelzen:
Features are not backported to 10.8 so it will be available with 10.9. There is no ETA for 10.9.

https://jellyfin.org/docs/general/faq#when-will-release-x-happen-when-will-feature-x-be-available

csaavedra:
Do I understand correctly that the default value will be 0, thus no limit?

nielsvanvelzen:
The default value is 0 which means `CPU core count * 2`.

momala454:
Take note that even with value 1, it still use too much ram on my 2gb raspberry so this feature alone is not enough

PR #8809, opened 2022-11-27
Disable loading MPlayer2

crobibero:
Related to https://github.com/jellyfin/jellyfin/issues/8198

This isn't a proper fix as we will likely get access denied further down the chain.. but this will allow Jellyfin to start.


cvium (changes requested):
We should be explicit about which ones we enable otherwise it's an endless game of whac-a-mole

PR #8812, opened 2022-11-27
Validate requested user id

crobibero:
Validates that the requested userid is accessible by the requesting user.


Jellyfin.Api/Helpers/RequestHelpers.cs
```diff
@@ -55,6 +55,38 @@ public static (string, SortOrder)[] GetOrderBy(IReadOnlyList<string> sortBy, IRe
...
+            var authenticatedUserId = claimsPrincipal.GetUserId();
+
+            // UserId not provided, fall back to authenticated user id.
+            if (userId is null || userId.Value.Equals(default))
+            {
+                return authenticatedUserId;
+            }
+
+            // Api key can always access requested user id.
+            if (claimsPrincipal.GetIsApiKey())
```
cvium, line 75: Api key already has admin role, so this check isn't necessary

Jellyfin.Api/Helpers/RequestHelpers.cs
```diff
@@ -55,6 +55,38 @@ public static (string, SortOrder)[] GetOrderBy(IReadOnlyList<string> sortBy, IRe
             return result;
         }
 
+        /// <summary>
+        /// Checks if the user can access a user.
+        /// </summary>
+        /// <param name="claimsPrincipal">The <see cref="ClaimsPrincipal"/> for the current request.</param>
+        /// <param name="userId">The user id.</param>
+        /// <returns>A <see cref="bool"/> whether the user can update the entry.</returns>
+        internal static Guid GetUserId(ClaimsPrincipal claimsPrincipal, Guid? userId)
```
cvium, line 64: I don't really understand this method tbh. If a user is trying to access another user's items, change their info or anything else, it should throw a Forbidden, not default to authenticated user.

crobibero: I went back and forth on whether it should return forbidden, will update in a few days

nielsvanvelzen:
FYI this PR needs to be updated now that #8734 is merged

Jellyfin.Api/Helpers/RequestHelpers.cs
```diff
@@ -55,6 +56,32 @@ public static (string, SortOrder)[] GetOrderBy(IReadOnlyList<string> sortBy, IRe
...
+            }
+
+            // User must be administrator to access another user.
+            var isAdministrator = claimsPrincipal.IsInRole(UserRoles.Administrator);
+            if (!userId.Value.Equals(authenticatedUserId) && !isAdministrator)
+            {
+                throw new SecurityException("Forbidden");
+            }
+
+            return userId.Value;
```
nielsvanvelzen, line 82: Should we check if the provided user id is valid before returning it here? I believe right now the API throws a 500 with an incorrect user id.

nielsvanvelzen: Actually at that point we might even change it to `GetUser()` and return the User(Dto?) object.

crobibero: I think that change is out of scope for this PR as it looks like there are a good amount of endpoints that still function if the user doesn't exist

Bond-009 (approved):

Shadowghost (approved):

Issue #1714, opened 2019-09-04
Lose entire library after drive disconnection


cactushydrocodone:
Hi,

Im currently not using jellyfin because of the android apps, but thats another point.
My question is, is this age old bug fixxed:
- Have medialibrary on a FUSE mountpoint
- Sometime you loose the mountpoint (Connection issue, Process reaper, unexpected circumstances, etc)
- Emby (bzw. jellyfin) deletes the entire library because it cant find the files anymore
- After a remount and restart of emby (bzw. jellyfin), the entire library is gone
- To restore, you need to do a full rescan, which can take days (My library is 60TB).

Workaround i had for this the problem:
- Keep a daily backup of the emby (bzw. jellyfin) data folder and restore it on mountpoint loss.

The original emby devs basically said: "ye we fix that, sometime."
That was 3 years ago.

Maybe i have luck with you guys.

cvium:
It's not fixed yet, but we want to.

EraYaN:
Kodi keeps the metadata in it's database for a while, and so re-adding the file at the same path will repopulate all data and watched status etc.

joshuaboniface:
I've never actually experienced this myself, either on Emby or Jellyfin, despite my CephFS volume going away all the time. Is there some particular action that triggers this? Even if the files aren't there, Jellyfin shouldn't be randomly deleting the entries in the DB, unless I'm missing something?

EraYaN:
I believe a scan can clean up “removed” files.

RaymondSchnyder:
> 
> 
> I've never actually experienced this myself, either on Emby or Jellyfin, despite my CephFS volume going away all the time. Is there some particular action that triggers this? Even if the files aren't there, Jellyfin shouldn't be randomly deleting the entries in the DB, unless I'm missing something?

Its as easy as this:

1. Emby / Jellyfun runs in a container, the FUSE mount is added with a docker mount
2. When the FUSE Mount on the host "goes away", it also goes away in the container
3. Emby / Jellyfin cant find its files anymore, at the next rescan, it begins to delete everything it doesnt have anymore (I do use a regular scan every 4 hours)
4. Even if you dont have a regular scan activated, at the next restart of the container (emby/jellyfin), it deletes ALL METADATA, EVERYTHING. All knowledge it ever had about those files.

RaymondSchnyder:
Another day on which ive lost a library.
The full rescan will take days, weeks.

Why did the emby devs did this like this?
[...]

PR #8814, opened 2022-11-28
Add opt-in folder stability feature for library scanning.

MattMckenzy:
Adding a ".jellycheck" file to a folder will register that folder for stability checks. Any future scan where this file is absent will have the scan skip the folder and not process it further.

Addresses #1714


MattMckenzy:
I just wanted to push this change out there in case there would be some constructive criticism on my changes. I still want to build a container from it so I can run it and test it on my home setup. Once I'm fully convinced it's working as it should, I'll flag it ready for review.

crobibero:
I would prefer the file be called ".jellyfin"
I havent gone through the code at all

cvium:
I don't think we can accept this solution. Many people mount their shares as read-only.

MattMckenzy:
The code is also read-only! People would signal to Jellyfin to check for the file in future scans by creating it themselves. This way it's completely transparent for anyone where this isn't an issue.

I do agree it's a bit hacky, yet this is the easiest and cleanest solution I could think of. 

No worries if you guys don't want to implement this, but I'll make use of it in my own setup to save me hours of effort every month.

MattMckenzy:
Apologies, just realized I closed this by accident while half-asleep haha. Feel free to close it again if you think there needs to be a better solution to this fix. 

I'd love to hear it if that's the case! Maybe I can spend sometime on making something a bit less hacky.

Shadowghost:
IMHO this kind of problem is not Jellyfin's responsibility, that's what your OS is for. Applications should not be responsible for handling filesystem or OS problems, that should be taken care of by the OS or service runner.

MattMckenzy:
> IMHO this kind of problem is not Jellyfin's responsibility, that's what your OS is for. Applications should not be responsible for handling filesystem or OS problems, that should be taken care of by the OS or service runner.

In a perfect world maybe, but good services should be built stable, anticipating issues with systems. Completely and irreversibly destroying a Library when a mount point is temporarily unavailable is not a reasonable response from Jellyfin.

da-anda:
instead of using a file in filesystem, wouldn't it be better if you could simply flag a library path as `removeable` in the library settings? This flag could then be used for other things as well, like forcing Jellyfin to cache the thumbnails in this path locally instead of relying on their availability while browsing the library etc.

armerpunkt:
> IMHO this kind of problem is not Jellyfin's responsibility, that's what your OS is for. Applications should not be responsible for handling filesystem or OS problems, that should be taken care of by the OS or service runner.

I recently had the SAS card die on me.  The OS knew just fine. There were messages in my syslog and smartmontools immediately sent me a notice they were gone. However, Jellyfin decided to delete my entire library before I could shut the system down for maintenance.  How's that the OS' fault? What exactly is the OS supposed to do? Should it shut the system down automatically taking all my other self-hosted services offline just to stop Jellyfin from deleting my library? That just seems a bit overkill when everything else continued running fine.

cvium:
I think we should emulate Plex

> On some operating systems—typically Windows, Linux, or some NAS devices—a content directory that is completely empty will appear to the filesystem the same as a directory that is inaccessible. If a content location is inaccessible, Plex Media Server will simply ignore it during a scan to help protect against incorrect removal of library content in situations such as an external hard drive being unplugged.

> In such a situation, you can either remove the source directory altogether or else place content in the directory again so that it can be scanned.

https://support.plex.tv/articles/201447966-why-does-content-i-deleted-on-my-drive-still-show-up/

sjorge:
If you are using systemd, you can also have systemd take down a service if a mount disappears, I have a drop-in for jellyfin.service that does this. It's not idea because if one of the mounts is gone it takes it down, but it's better than having to rescan the entire mount later.

Example drop-in:
```ini
[Unit]
BindsTo=mm-anime.mount
After=mm-anime.mount
BindsTo=mm-movies.mount
After=mm-movies.mount
BindsTo=mm-music.mount
After=mm-music.mount
BindsTo=mm-series-archive.mount
After=mm-series-archive.mount
BindsTo=mm-series-live.mount
After=mm-series-live.mount
```

It's the `BindsTo` specifically that will take the service down.

Having jellyfin somehow deal with it somehow, would be more ideal IMHO. I do like the way describe above by cvium.

krutburken:
My current workaround for this is to remove all the scheduled library scans in Jellyfin itself and replace them with a script in cron that first checks for all mounts before triggering a scan via api instead. The issue does need a real fix though.

MattMckenzy:
> I think we should emulate Plex
> 
> > On some operating systems—typically Windows, Linux, or some NAS devices—a content directory that is completely empty will appear to the filesystem the same as a directory that is inaccessible. If a content location is inaccessible, Plex Media Server will simply ignore it during a scan to help protect against incorrect removal of library content in situations such as an external hard drive being unplugged.
> 
> > In such a situation, you can either remove the source directory altogether or else place content in the directory again so that it can be scanned.
> 
> [support.plex.tv/articles/201447966-why-does-content-i-deleted-on-my-drive-still-show-up](https://support.plex.tv/articles/201447966-why-does-content-i-deleted-on-my-drive-still-show-up/)

This is a great simple solution that works perfectly for my use cases. I've made the changes to my fork [MattMckenzy/jellyfin](https://github.com/MattMckenzy/jellyfin) and have been using them on my home instance with no issues!

However, I think this change might cause confusion unless properly documented, so I decided to close this merge request for now.

If anyone from Jellyfin would like a new merge request with these changes, please let me know, I'd be glad to help!

sulisu:
I am using two external usb 5-disks enclosure to store Jellyfin Library. This is a must have feature for me.

Now I can get the jellyfin server up and run and scan newly added contents, without worrying about all libraries being deleted now and then.

> > I think we should emulate Plex
> > > On some operating systems—typically Windows, Linux, or some NAS devices—a content directory that is completely empty will appear to the filesystem the same as a directory that is inaccessible. If a content location is inaccessible, Plex Media Server will simply ignore it during a scan to help protect against incorrect removal of library content in situations such as an external hard drive being unplugged.
> > 
> > 
> > > In such a situation, you can either remove the source directory altogether or else place content in the directory again so that it can be scanned.
> > 
> > 
> > [support.plex.tv/articles/201447966-why-does-content-i-deleted-on-my-drive-still-show-up](https://support.plex.tv/articles/201447966-why-does-content-i-deleted-on-my-drive-still-show-up/)
> 
> This is a great simple solution that works perfectly for my use cases. I've made the changes to my fork [MattMckenzy/jellyfin](https://github.com/MattMckenzy/jellyfin) and have been using them on my home instance with no issues!
> 
> However, I think this change might cause confusion unless properly documented, so I decided to close this merge request for now.
> 
> If anyone from Jellyfin would like a new merge request with these changes, please let me know, I'd be glad to help!

cvium:
https://github.com/jellyfin/jellyfin/pull/9291

sulisu:
I port #9291 back to 10.8.9, it works well. Windows 10, library on two external usb disk enclosures.

ShrirajHegde:
> IMHO this kind of problem is not Jellyfin's responsibility, that's what your OS is for. Applications should not be responsible for handling filesystem or OS problems, that should be taken care of by the OS or service runner.

That is not an 'OS problem'. We can reasonably expect non-local drives to have this problem. Mounting a network drive is a pretty standard use case for Jellyfin or any similar media system. 

Please decide on something, even if it isn't "aesthetic" as many users are facing this problem from a LONG time.

One criteria I can think of is to set a timeout for missing files like N hours, days or weeks before purging from the library. This way, stable libraries can be set with a small timeout resulting in a dynamic library, while unstable mounts can avoid rescans with larger timeout. This way, no manual intervention is needed like Plex's behavior. The library eventually becomes consistent. This will also add the ability to customize the deletion behavior between a local and network library.

MattMckenzy:
Just to give some extra information on my case. Every time my libraries would get purged, it was when my container would get restarted. Maybe because I was using portainer and defining the libraries as volume mounts there would be some weird delay or something.

Regardless, I gave up on containerizing it and installed Jellyfin with a debian package instead. Installed this way, I've never had a mysteriously purged library again!

Just thought I'd mention it in case this helps someone.

ShrirajHegde:
I think this is a great solution as of now.
Using systemd service to mount and waiting for the mount before starting docker service (see step 11)

https://gist.github.com/Gyarbij/4dc1fe668b6e7d804b490bebddd3ac80

Issue #1888, opened 2019-10-12
Checkmark is off center when using AddPlayedIndicator=true in the API


cewert:
**Describe the bug**
The image returned when using AddPlayedIndicator=true has the circle with checkmark but the checkmark is off center

**To Reproduce**
Use the roku client

Or use a browser to access the API:
1. Use the web client to view a movie or tv show detail page
2. Copy the id from the url (id=)
3. Open jellyfin web client in a new tab
4. Replace "/web/index.html#!/home.html" in the URL with "/Items/{ITEMID}/Images/Primary?AddPlayedIndicator=true" replacing {ITEMID} with the id copied from step 2
5. Observe check mark in top right corner of image

**Expected behavior**
Check mark should be centered

**Screenshots**
![Screenshot from 2019-10-11 21-55-19](https://user-images.githubusercontent.com/16514652/66692983-dfb2b000-ec71-11e9-8208-a8a821436bbc.png)

**System (please complete the following information):**
 - OS: Docker
 - Browser: Firefox
 - Jellyfin Version: 10.4.0
 - Reverse proxy: none

onny:
Shouldn't we move this bug report to the Jellyfin-web repository?

anthonylavado:
> Shouldn't we move this bug report to the Jellyfin-web repository?
@onny The checkmark is rendered by the server, and returned in the API. This is the right repo, really. This is going to be somewhere in the Skia commands

dkanada:
@cewert is this still broken?

cewert:
I just tested and yes this is still broken for me. I accessed the API and observed the behavior using the latest version of firefox as well as the roku client.

OS: Docker
Jellyfin: 10.4.3

cewert:
I should have mentioned in the OP - the reason I created this issue is because the roku client uses this API param. @bisby also confirmed it was not working on his roku

anthonylavado:
I forget if Android TV uses this or not. Will check.

anthonylavado:
An update on what is now a very old thread. Android TV hasn't used this in a long time. The only known clients left are Roku and https://github.com/SenorSmartyPants/jellyfin-nmt by @SenorSmartyPants.

With Roku under increased development lately, we might be able to tackle generating client side watched indicators.

With that, though it may be some time away, we are considering deprecating this feature for 10.9.

Issue #312, opened 2018-12-29
Method for uploading media to the server


Strit:
One of the annoying things about running a media server, has always been to get the media onto the server itself.

I would like a way to upload media to the media folders on the server itself. Kind of like a file manager, but from the webUI.

LogicalPhallacy:
This is a good idea, but it might require n examination of user permission levels/the creation of a permissions system deeper than the admin/notadmin split that's there right now.

Strit:
Well, make it a setting in the user profile, just like the "delete media" one is.
But yes, probably require better system permissions to write to the filesystem.

joshuaboniface:
I'm kinda iffy on this, as I don't like what projects like Stream try to do regarding this, but if it can be implemented cleanly with a default-off flag I'd support it.

Strit:
> can be implemented cleanly with a default-off flag I'd support it.

That's what I was thinking.
Make it an Admin only thing, that is not on by default.

Thanks for considering it!

jaredallard:
You can always run a server on the host that allows you to upload media, i.e https://github.com/tritonmedia/twilight. Of course that's limited to in-network uploads. You could also use the entire platform (https://github.com/tritonmedia) to provide a pipeline to upload / download media.

Disclaimer: I'm the author of this, and it's not particularly ready for anyone else to deploy, but it might be useful per component.

SenorSmartyPants:
It seems like this is a feature that would be better left to something like a WebDAV server or other software built specifically to handle uploads. Let Jellyfin just handle the media that's there.

gsnerf:
I'm not sure it is a good idea for jellyfin to try to do to much stuff on it's own. Especially since this is crossplatform, it would probably need to support special cases for different filesystems. Also many users have different likes for stuff like folder structures naming etc. Trying to create a manager that can handle all that would essentially mean rebuilding already existing stuff like owncloud/nextcloud or similar stuff (why not use that instead?).

If this truly is something that should become a part of jellyfin I think its best seperated from the serving component into a separate file management component. Automatically adding the new files to the libaries should already be covered by the regular media scanner which monitors the library folders. Allowing this on a per user basis would definately need upload and filespace quotas, which might be overkill to start with. Better put it in the admin section and only allow admins to do this.

@joshuaboniface this request also heavily touches frontend stuff, is there a tag for that too?

SenorSmartyPants:
> > Let Jellyfin just handle the media that's there.
> 
[...]

PR #8819, opened 2022-11-29
Offset API Played Indicator to correct position

photonconvergence:
check

Fixes [#1888](https://github.com/jellyfin/jellyfin/issues/1888)
Fixes  [#312](https://github.com/jellyfin/jellyfin-roku/issues/312)


nielsvanvelzen (approved):
Change looks fine but I believe we're still planning to remove this option from the server.

photonconvergence:
Figured I might as well fix it anyway, just in case it survives for another release or two.

PR #8827, opened 2022-11-30
Improve DB perf for everyone not using debug logging

Bond-009:
* don't calculate elapsed time if we aren't going to use it
* increased accuracy of the elapsed time

### DIFF
--- a/Emby.Server.Implementations/Data/SqliteItemRepository.cs
+++ b/Emby.Server.Implementations/Data/SqliteItemRepository.cs
@@ -5,9 +5,11 @@
 using System;
 using System.Buffers.Text;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Runtime.CompilerServices;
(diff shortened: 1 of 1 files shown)


Emby.Server.Implementations/Data/SqliteItemRepository.cs
```diff
@@ -6246,5 +6202,49 @@ private MediaAttachment GetMediaAttachment(IReadOnlyList<ResultSetValue> reader)
...
+            {
+                _logger = logger;
+                _methodName = methodName;
+                _commandText = commandText;
+                _stopwatch = logger.IsEnabled(LogLevel.Debug) ? Stopwatch.StartNew() : null;
+            }
+
+            public void Dispose()
+            {
+                if (_stopwatch == null)
```
crobibero, line 6225:
```suggestion
                if (_stopwatch is null)
```

You changed all `== null` to `is null` in another pr :)

crobibero (approved):

Shadowghost (approved):

PR #8833, opened 2022-12-01
Add MusicBrainz server validation and fallback

Shadowghost:
Fixes one of the outstanding kinks of the new MusicBrainz plugin implementation. Makes this compatible with the config value used in 10.8.

**Changes**
* Validate server URI
* Properly set query options
* Fallback to official server if no valid URI is specified

### DIFF
--- a/MediaBrowser.Providers/Plugins/MusicBrainz/MusicBrainzAlbumProvider.cs
+++ b/MediaBrowser.Providers/Plugins/MusicBrainz/MusicBrainzAlbumProvider.cs
@@ -13,6 +13,7 @@ using MediaBrowser.Providers.Music;
 using MetaBrainz.MusicBrainz;
 using MetaBrainz.MusicBrainz.Interfaces.Entities;
 using MetaBrainz.MusicBrainz.Interfaces.Searches;
+using Microsoft.Extensions.Logging;
 
 namespace MediaBrowser.Providers.Plugins.MusicBrainz;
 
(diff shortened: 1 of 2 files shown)


Bond-009 (approved):

MediaBrowser.Providers/Plugins/MusicBrainz/MusicBrainzAlbumProvider.cs
```diff
@@ -21,16 +22,36 @@ namespace MediaBrowser.Providers.Plugins.MusicBrainz;
 /// </summary>
 public class MusicBrainzAlbumProvider : IRemoteMetadataProvider<MusicAlbum, AlbumInfo>, IHasOrder, IDisposable
 {
+    private readonly ILogger _logger;
```
crobibero, line 25:
```suggestion
    private readonly ILogger<MusicBrainzAlbumProvider> _logger;
```

MediaBrowser.Providers/Plugins/MusicBrainz/MusicBrainzArtistProvider.cs
```diff
@@ -22,16 +22,36 @@ namespace MediaBrowser.Providers.Plugins.MusicBrainz;
 /// </summary>
 public class MusicBrainzArtistProvider : IRemoteMetadataProvider<MusicArtist, ArtistInfo>, IDisposable
 {
+    private readonly ILogger _logger;
```
crobibero, line 25:
```suggestion
    private readonly ILogger<MusicBrainzArtistProvider> _logger;
```

crobibero (approved):

PR #8838, opened 2022-12-01
Add log point explaining why stream cant be shared

ajb142:
This change addresses https://github.com/jellyfin/jellyfin/issues/8756.

I believe this extra log point will help others set up and fault find issues with live tv and recording.

It was not clear to me or the op of the issue that we where getting certain response codes. So this extra log point would have saved a lot of debugging.


ajb142:
This change addresses #8756.

I believe this extra log point will help others set up and fault find issues with live tv and recording.

It was not clear to me or the op of the issue that we where getting certain response codes. So this extra log point would have saved a lot of debugging.

PR #8853, opened 2022-12-04
Drop dropped subtitles

sevenrats:
Remove subtitles with drop method from playbackinfo.

### DIFF
--- a/Jellyfin.Api/Controllers/MediaInfoController.cs
+++ b/Jellyfin.Api/Controllers/MediaInfoController.cs
@@ -228,31 +228,19 @@ namespace Jellyfin.Api.Controllers
                 }
             }
 
-            var dropList = new List<String>();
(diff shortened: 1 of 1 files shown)


nielsvanvelzen:
Is there an issue related to this change? If not, can you explain why this change is desired? This change might break certain clients that depend on all subtitle streams to exist.

sevenrats:
> Is there an issue related to this change? If not, can you explain why this change is desired? This change might break certain clients that depend on all subtitle streams to exist.

The benefit is for low-power servers where transcoding is undesirable or impossible.  For example, if a user has transcoding disabled, and clicks on a PGS sub, it fails to play.  On clients which do not advertise the subtitle codec this failure is confusing and frustrating and makes it seem like the system is broken.

@nielsvanvelzen would you prefer a new SubtitleDeliveryMethod, such as "Hide"

sevenrats:
This was to fix an edge case on roku that it doesn't actually fix.

PR #8887, opened 2022-12-10
Add permissions for collection management

### LINKED ISSUE #7179: [Issue]: all users can add to collections
### Please describe your bug

A regular user with no modify/edit permission is able to add movies to a new collection and add to an existing collection.

users without edit permission shouldn’t be able to do that as collections should be curated by the system 

### Jellyfin Version

10.7.7

### if other:

_No response_

### Environment

```markdown
- OS: Debian 11
- Virtualization: Docker 
- Clients: iOS App
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

Shadowghost:
**Changes**
* Adds a permission to manage collections (add, remove and delete)
* Enforce permission on the collection endpoints

Will require changes to jellyfin-web (https://github.com/jellyfin/jellyfin-web/pull/4199) to take the new permission into account when displaying the elements in the item dropdown and collection screens. Additionally requires a setting in jellyfin-web to be available to set it.

**Issues**
Fixes #7179

### DIFF
--- a/Jellyfin.Api/Controllers/CollectionController.cs
+++ b/Jellyfin.Api/Controllers/CollectionController.cs
@@ -1,6 +1,7 @@
 using System;
 using System.ComponentModel.DataAnnotations;
 using System.Threading.Tasks;
+using Jellyfin.Api.Constants;
 using Jellyfin.Api.Extensions;
 using Jellyfin.Api.ModelBinders;
 using MediaBrowser.Controller.Collections;
@@ -16,7 +17,7 @@ namespace Jellyfin.Api.Controllers;
 /// The collection controller.
 /// </summary>
 [Route("Collections")]
-[Authorize]
+[Authorize(Policy = Policies.CollectionManagement)]
 public class CollectionController : BaseJellyfinApiController
 {
     private readonly ICollectionManager _collectionManager;
--- a/Jellyfin.Server/Extensions/ApiServiceCollectionExtensions.cs
+++ b/Jellyfin.Server/Extensions/ApiServiceCollectionExtensions.cs
@@ -10,7 +10,6 @@ using Emby.Server.Implementations;
 using Jellyfin.Api.Auth;
 using Jellyfin.Api.Auth.AnonymousLanAccessPolicy;
 using Jellyfin.Api.Auth.DefaultAuthorizationPolicy;
-using Jellyfin.Api.Auth.DownloadPolicy;
 using Jellyfin.Api.Auth.FirstTimeSetupPolicy;
 using Jellyfin.Api.Auth.SyncPlayAccessPolicy;
 using Jellyfin.Api.Auth.UserPermissionPolicy;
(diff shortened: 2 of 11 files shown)


Jellyfin.Api/Controllers/CollectionController.cs
```diff
@@ -54,6 +60,12 @@ public class CollectionController : BaseJellyfinApiController
             [FromQuery] bool isLocked = false)
         {
             var userId = User.GetUserId();
+            var user = _userManager.GetUserById(userId);
+
+            if (!user.HasPermission(PermissionKind.IsAdministrator) || !user.HasPermission(PermissionKind.EnableCollectionManagement))
```
cvium, line 65: Shouldn't this be a policy?

Shadowghost: I guess that makes more sense here, will change and force push once done.

Shadowghost:
Changed to use a new auth permission.

Don't mind the other changed files, I just switched the files to file-based namespaces.

crobibero (approved):

Snipy2k4:
I don't understand how to fix this? Will this come in an update or do I have to download something here?!

crobibero:
This will come in an update

Jellyfin.Api/Auth/BaseAuthorizationHandler.cs
```diff
@@ -38,35 +39,35 @@ public abstract class BaseAuthorizationHandler<T> : AuthorizationHandler<T>
...
             if (isApiKey)
             {
                 return true;
             }
 
-            // Ensure claim has userId.
+            // Get user id from claim.
             var userId = claimsPrincipal.GetUserId();
-            if (userId.Equals(default))
+            if (userId.Equals(Guid.Empty))
```
crobibero, line 64: I'm pretty sure @Bond-009 converted to `.Equals(default)` because it has better performance

Jellyfin.Api/Auth/DefaultAuthorizationPolicy/DefaultAuthorizationHandler.cs
```diff
@@ -4,41 +4,39 @@
...
-        /// </summary>
-        /// <param name="userManager">Instance of the <see cref="IUserManager"/> interface.</param>
-        /// <param name="networkManager">Instance of the <see cref="INetworkManager"/> interface.</param>
-        /// <param name="httpContextAccessor">Instance of the <see cref="IHttpContextAccessor"/> interface.</param>
-        public DefaultAuthorizationHandler(
-            IUserManager userManager,
-            INetworkManager networkManager,
-            IHttpContextAccessor httpContextAccessor)
-            : base(userManager, networkManager, httpContextAccessor)
+        if (ValidateClaims(context.User, out var user))
```
crobibero, line 31:
```suggestion
        if (ValidateClaims(context.User, out _))
```

Should discard the parameter if it won't be used

Jellyfin.Api/Auth/BaseAuthorizationHandler.cs
Bond-009: Correct

Jellyfin.Server/Extensions/ApiServiceCollectionExtensions.cs
```diff
@@ -54,8 +51,6 @@ public static IServiceCollection AddJellyfinApiAuthorization(this IServiceCollec
         {
             // The default handler must be first so that it is evaluated first
             serviceCollection.AddSingleton<IAuthorizationHandler, DefaultAuthorizationHandler>();
-            serviceCollection.AddSingleton<IAuthorizationHandler, UserPermissionHandler>();
-            serviceCollection.AddSingleton<IAuthorizationHandler, FirstTimeSetupHandler>();
```
crobibero, line 58: Accidental removal?

Shadowghost: Fixed it.

crobibero (approved):

Bond-009 (approved):

PR #8842, opened 2022-12-03
Augment search to consider tags

bradbeattie:
**Changes**
This minor change permits users to search for items via tags.

**Issues**

* https://github.com/jellyfin/jellyfin/issues/7691
* https://features.jellyfin.org/posts/276/search-by-tag-genre
* https://features.jellyfin.org/posts/635/search-by-tag
* https://www.reddit.com/r/jellyfin/comments/paqi3t/is_there_a_way_of_searching_by_tag/
* https://www.reddit.com/r/jellyfin/comments/xp94po/am_i_doing_something_wrong_i_cannot_search_with/

**Commentary**

Seems to be a frequently requested feature. I considered searching with a join on the `ItemValues` table, but figured this direct approach would likely be sufficient.

PR #8889, opened 2022-12-11
Augment TMDB tags with original language

bradbeattie:
**Changes**
When generating tag metadata from TMDB, include the original language as a tag (e.g. `lang-en` or `lang-fr`).

The intent here is to piggyback on [SqliteItemRepository's use of tags to determine item similarity](https://github.com/jellyfin/jellyfin/blob/master/Emby.Server.Implementations/Data/SqliteItemRepository.cs#L2418) to also consider a movie's original (primary) language. The result being that Jellyfin's "More Like This" would give _some_ relevance to a movie's language.

Furthermore, in concert with https://github.com/jellyfin/jellyfin/pull/8842 this change would allow searching for content by language.

**Issues**

Partially addresses https://features.jellyfin.org/posts/39/filter-videos-by-audio-stream-language

### DIFF
--- a/MediaBrowser.Providers/Plugins/Tmdb/Movies/TmdbMovieProvider.cs
+++ b/MediaBrowser.Providers/Plugins/Tmdb/Movies/TmdbMovieProvider.cs
@@ -241,6 +241,7 @@ namespace MediaBrowser.Providers.Plugins.Tmdb.Movies
 
             if (movieResult.Keywords?.Keywords != null)
             {
+                movie.AddTag("lang-" + movieResult.OriginalLanguage);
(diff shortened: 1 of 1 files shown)


bradbeattie:
GitHub actions are showing a failure, though I suspect it isn't my code change but instead related to the dotnet 6->7 changes commited within the last 72 hours. I'll revisit this PR when that's resolved.

bradbeattie:
Closed and reopened as https://github.com/jellyfin/jellyfin/pull/8916. I suspect the dotnet changes have made their way in, but I couldn't seem to find a way to kick off the Github actions again without reopening the PR.

PR #8914, opened 2022-12-15
Augment tag searching to consider individual ItemValues instead of the delimited tags field

bradbeattie:
**Changes**
This change improves tag searching. Previously, `(Tags not null and Tags like @SearchTermContains)` searched against the composed `tag1|tag2|tag3` value. That approach had no good way of differentiating between `str1` and `prefix...str1...suffix` which might have entirely different meaning (e.g. searching `cia` would turn up content from `artifiCIAl intelligence`).

This PR provides bias towards startswith and equals, additionally biasing content with multiple matching tags.

### DIFF
--- a/Emby.Server.Implementations/Data/SqliteItemRepository.cs
+++ b/Emby.Server.Implementations/Data/SqliteItemRepository.cs
@@ -2461,7 +2461,9 @@ namespace Emby.Server.Implementations.Data
                 if (query.SearchTerm.Length > 1)
                 {
                     builder.Append("+ ((CleanName like @SearchTermContains or (OriginalTitle not null and OriginalTitle like @SearchTermContains)) * 10)");
-                    builder.Append("+ ((Tags not null and Tags like @SearchTermContains) * 5)");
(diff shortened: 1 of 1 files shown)


bradbeattie:
Still relevant, I'd argue.

bradbeattie:
Thanks. Still need one more approval!

cvium (approved):

richlawson:
Sorry to raise a question in a merged PR, but this augmentation should also include genres, correct? Would this have needed to be restricted to where `Type=4` to only include tags? Genres look like they're `Type=2`.

I'm looking at lines [3952](https://github.com/jellyfin/jellyfin/blob/master/Emby.Server.Implementations/Data/SqliteItemRepository.cs#L3952) and [3937](https://github.com/jellyfin/jellyfin/blob/master/Emby.Server.Implementations/Data/SqliteItemRepository.cs#L3937), noticing that the `ItemValues` specific to Tags and Genres have a where clause that includes that type. 

I was looking into how to add genres to the search after reviewing https://github.com/jellyfin/jellyfin/pull/8842, and it looks like that work is already completed here with this further enhancement.

PR #8916, opened 2022-12-16
Augment Tmdb tags to include original language

bradbeattie:
**Changes**
When generating tag metadata from TMDB, include the original language as a tag (e.g. `lang-en` or `lang-fr`).

The intent here is to piggyback on [SqliteItemRepository's use of tags to determine item similarity](https://github.com/jellyfin/jellyfin/blob/master/Emby.Server.Implementations/Data/SqliteItemRepository.cs#L2418) to also consider a movie's original (primary) language. The result being that Jellyfin's "More Like This" would give _some_ relevance to a movie's language.

Furthermore, in concert with https://github.com/jellyfin/jellyfin/pull/8842 this change would allow searching for content by language.

**Issues**

Partially addresses https://features.jellyfin.org/posts/39/filter-videos-by-audio-stream-language

### DIFF
--- a/MediaBrowser.Providers/Plugins/Tmdb/Movies/TmdbMovieProvider.cs
+++ b/MediaBrowser.Providers/Plugins/Tmdb/Movies/TmdbMovieProvider.cs
@@ -241,6 +241,7 @@ namespace MediaBrowser.Providers.Plugins.Tmdb.Movies
 
             if (movieResult.Keywords?.Keywords is not null)
             {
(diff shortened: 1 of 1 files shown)


Shadowghost:
IMHO this is a workaround. Ideally current and original content language would be saved explicitly in the database.

bradbeattie:
Yep. Happy to close if less patchy way of doing this is in the works.

PR #8928, opened 2022-12-19
Remove unused ImageByName API

nielsvanvelzen:
**Changes**
- Remove unused ImageByName API

The API contains 3 endpoints to list 3 metadata directories: general, ratings and mediainfo. The server never puts files in those places, they are non-existent by default. So the responses are all `[]`. Then there are endpoints to serve files from those directories, which are basically useless because there are no files.

I've searches with GH code search in our organization for usages of these endpoints and couldn't find any. My proposal is to remove the controller for now and if any issues arise during 10.9 testing (beta) we can revert this PR.

**Issues**


cvium (approved):
Seems safe to remove. There's more info here https://emby.media/community/index.php?/topic/695-imagesbyname-folder/ and if you search their forums, it seems like it was taken out of commission sometime during 2015.

crobibero (approved):

Bond-009 (approved):
Lets purge :fire:

PR #8933, opened 2022-12-20
Adds User Ratings

TheItalianGame:
**Changes**
Adds user rating functionality to the server portion of jellyfin
Specifically you can add a user rating sort by a user rating and filter by a user rating.

Currently working on it from the client side.

### DIFF
--- a/Emby.Server.Implementations/Data/SqliteItemRepository.cs
+++ b/Emby.Server.Implementations/Data/SqliteItemRepository.cs
@@ -2164,6 +2164,7 @@ namespace Emby.Server.Implementations.Data
                     || sortingFields.Contains(ItemSortBy.IsPlayed)
                     || sortingFields.Contains(ItemSortBy.IsUnplayed)
                     || sortingFields.Contains(ItemSortBy.PlayCount)
+                    || sortingFields.Contains(ItemSortBy.UserRating)
                     || sortingFields.Contains(ItemSortBy.DatePlayed)
                     || sortingFields.Contains(ItemSortBy.SeriesDatePlayed)
                     || query.IsFavoriteOrLiked.HasValue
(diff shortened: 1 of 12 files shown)


TheItalianGame:
I had to force push because the author was missing.

TheItalianGame:
API Changes.

Created two new routes:

`POST` `/Users/{userId}/UserRating/{itemId}` to update the rating and
`DELETE` `/Users/{userId}/UserRating/{itemId}` to remove the rating

Added query parameter to  `/Users/{userId}/Items` and `Users/{userId}/Items` called minUserRating to filter by user ratings.

In addition the `sortBy` parameter of the query routes now accepts the value `UserRating`

cmsimike:
*note* I am not associated with this project, just a user.
Thank you for this! I use UserRating very heavily in my previous Kodi library so this would be a great addition.

I recently migrated my metadata from Kodi to Jellyfin (through a custom end point meant to insert the data directly)

I dug through the API and found the `UserData` object on `Item`s which already contains a `Rating` property
image


I had assumed this was a user rating field already, so I populated my rating data in here. What's the difference between `Rating` and your new `UserRating`?

TheItalianGame:
I wasn't sure what that field was and neither was anyone else. It seems like many years ago there was plan to implement user ratings but it never went anywhere.


If the pr is approved I will open another that removes that column.

nielsvanvelzen:
I think it would be better to re-use the existing rating column/api member. From what I can see it's unused in the server, except for the likes boolean (which stores a 10 in rating for like and a 1 for dislike). We've deprecated the likes/dislikes (no client uses it) so it should be safe to use the rating column.

TheItalianGame:
Alright I'll do that.

I'd prefer if you delete this pr and I'll open another one when I change the column.

Bond-009:
Closing as requested

PR #8951, opened 2022-12-23
Explicitly setting DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=0 in Dockerfiles

bradbeattie:
# Context

https://github.com/jellyfin/jellyfin/pull/6629 implicitly disabled DOTNET_SYSTEM_GLOBALIZATION_INVARIANT in docker images. However, by only commenting out `=1` and not explicitly setting `=0`, there exists at least one side effect of this.

## Detected side effect

1. Using the [Jellyfin Docker image](https://hub.docker.com/layers/jellyfin/jellyfin/20221223.2-unstable/images/sha256-bd807f9aa0a0251bd537f30ae5973a9a7d80d1284529e62d59826098bc46f217?context=explore)
2. Install `dotnet-script` and generate a script with the following code from [StringExtensions.cs](https://github.com/jellyfin/jellyfin/blob/master/src/Jellyfin.Extensions/StringExtensions.cs#L22):
```C#
using System;
using System.Globalization;
using System.Text;
using System.Text.RegularExpressions;

static readonly Regex _nonConformingUnicode = new Regex("([\ud800-\udbff](?![\udc00-\udfff]))|((?<![\ud800-\udbff])[\udc00-\udfff])|(\ufffd)");

public static string RemoveDiacritics(this string text)
{
    string withDiactritics = _nonConformingUnicode
	.Replace(text, string.Empty)
	.Normalize(NormalizationForm.FormD);

    var withoutDiactritics = new StringBuilder();
    foreach (char c in withDiactritics)
    {
	UnicodeCategory uc = CharUnicodeInfo.GetUnicodeCategory(c);
	if (uc != UnicodeCategory.NonSpacingMark)
	{
	    withoutDiactritics.Append(c);
	}
    }

    return withoutDiactritics.ToString().Normalize(NormalizationForm.FormC);
}

Console.WriteLine(RemoveDiacritics("é"));
```
3. Run this script both with and without `DOTNET_SYSTEM_GLOBALIZATION_INVARIANT`:
```
# DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=0 /root/.dotnet/tools/dotnet-script helloworld.csx 
e
# DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=1 /root/.dotnet/tools/dotnet-script helloworld.csx 
é
```

# Changes

Explicitly setting `DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=0`

# Fixes

This change ensures that the unit tests we see in https://github.com/jellyfin/jellyfin/blob/master/tests/Jellyfin.Extensions.Tests/StringExtensionsTests.cs#L12 actually behave the same way in deployed Docker images. Otherwise the `CleanName` field in library.db contains accents, resulting in false negative searches for titles that actually contain diacritics.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -61,7 +61,7 @@ RUN apt-get update \
  && chmod 777 /cache /config /media \
  && sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && locale-gen
 
-# ENV DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=1
+ENV DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=0
 ENV LC_ALL en_US.UTF-8
 ENV LANG en_US.UTF-8
 ENV LANGUAGE en_US:en
--- a/Dockerfile.arm
+++ b/Dockerfile.arm
@@ -49,7 +49,7 @@ RUN apt-get update \
  && chmod 777 /cache /config /media \
  && sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && locale-gen
 
-# ENV DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=1
+ENV DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=0
 ENV LC_ALL en_US.UTF-8
 ENV LANG en_US.UTF-8
 ENV LANGUAGE en_US:en
--- a/Dockerfile.arm64
+++ b/Dockerfile.arm64
@@ -43,7 +43,7 @@ RUN apt-get update && apt-get install --no-install-recommends --no-install-sugge
  && chmod 777 /cache /config /media \
  && sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && locale-gen
 
-# ENV DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=1
+ENV DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=0
 ENV LC_ALL en_US.UTF-8
 ENV LANG en_US.UTF-8
 ENV LANGUAGE en_US:en


bradbeattie:
Short-term workaround for anyone else running into this: Just build your own docker image and explicitly override envvar like so.

```
FROM jellyfin/jellyfin:<whatever>
ENV DOTNET_SYSTEM_GLOBALIZATION_INVARIANT="0"
```

bradbeattie:
Possibly unnecessary given https://github.com/jellyfin/jellyfin/pull/8964? Though until that's in place, `ENV DOTNET_SYSTEM_GLOBALIZATION_INVARIANT="0"` is a reasonable workaround.

Bond-009:
Closing as this should be fixed now

TheGreatDoc:
Hi there!

I'm trying this solution and is not working as it should (or I'm missing something).

> ```
> FROM jellyfin/jellyfin:<whatever>
> ENV DOTNET_SYSTEM_GLOBALIZATION_INVARIANT="0"
> ```

Actually (without the "patch") when I search for `mana` or `sober` artists `maná` and `sôber` doesnt shows in search results, but they do if I explicitly search for `maná` or `sôber`.

With the builded image, using `ENV DOTNET_SYSTEM_GLOBALIZATION_INVARIANT="0"`, they never show in search result, even if you search explicitly using the diacritic.

Maybe I need to do anything else? some kind of database rebuild? I've tried redownloading metadata for all files, but didnt fix it.

Thanks in advance 

PS: Sorry for asking here for help, but didnt had luck in matrix

bradbeattie:
I suspect your `TypedBaseItems.CleanName` column is still populated with diacritics, but your search is being normalized. Might you try 

1. Running `sqlite3 <whatever path>/library.db "SELECT Name, CleanName FROM TypedBaseItems WHERE Name LIKE '%maná%'"` to confirm the above hypothesis
2. Refreshing the metadata on the affected content
3. Then running the above query again to confirm that the diacritics are maintained in `Name` but stripped in `CleanName`.

TheGreatDoc:
Worked like a charm. 

I had to disable vgmdb plugin as was throwing errors and I think that was the issue in my last attempt that didnt work.

Thank your very much sir.

PR #8952, opened 2022-12-24
Remove unused notification endpoints

nielsvanvelzen:
**Changes**

Remove some unused notification endpoints. Most of them were no-op since forever (at least since the API migration). The only endpoint that actually did something was `CreateAdminNotification` but I did not see any usages in any of our own clients.

This PR focuses on the API, the notification manager could be cleaned a bit too because it does contain code to have user-targeted notifications which is not used.

**Issues**


crobibero (approved):

Shadowghost (approved):
We should make sure to clean up the manager though.

PR #8964, opened 2022-12-27
Use Diacritics.NET

### LINKED ISSUE #8893: The search function cannot find results for the "œ" character when typing "oe"
**Describe The Bug**
When searching for a movie or show that has the "œ" character in the title, typing "oe" does not return any results but they should be considered the same.

**Steps To Reproduce**

1. Have a show or movie with a title like : Le cœur a ses raisons
2. Use the search bar and search for : Le coeur a ses raisons
3. The show or movie will not show up

**Expected Behavior**
Typing "oe" in the search bar should return the same results as "œ".

**Logs**

**Screenshots**
image
image

 - Platform: Ubuntu server (using docker)
 - Browser: Firefox (probably all others too)
 - Jellyfin Version: 10.8.7

**Additional Context**
N/A

Bond-009:
Last time we had to revert this due to regressions, now those regression tests seem to succeed with a newer version of Diacritics.NET

Proper fix for the issue mentioned in #8951

Fixes #8893

### DIFF
--- a/MediaBrowser.Providers/Manager/MetadataService.cs
+++ b/MediaBrowser.Providers/Manager/MetadataService.cs
@@ -444,8 +444,8 @@ namespace MediaBrowser.Providers.Manager
                 }
             }
 
-            if ((originalPremiereDate ?? DateTime.MinValue) != (item.PremiereDate ?? DateTime.MinValue) ||
-                (originalProductionYear ?? -1) != (item.ProductionYear ?? -1))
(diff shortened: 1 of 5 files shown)


crobibero (approved):

Shadowghost (approved):
I trust you tested search etc. Apart from that the code changes look fine :)

cvium (approved):

PR #8978, opened 2022-12-29
Fix `multi` cleaning

Xavier-Do:
Right now, a movie Name `Iron Man Multi 1080p.mkv` will be searched as `Iron Man Multi` leading to no result.

The cleaning regex was containing multi but it looks like a typo joined `multi` and `subs` in the same term.

### DIFF
--- a/Emby.Naming/Common/NamingOptions.cs
+++ b/Emby.Naming/Common/NamingOptions.cs
@@ -153,7 +153,7 @@ namespace Emby.Naming.Common
 
             CleanStrings = new[]
             {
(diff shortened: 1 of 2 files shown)


Bond-009 (approved):
Can you add a regression test for this?

Xavier-Do:
> Can you add a regression test for this?

I can try (Just starting to code in JellyFin :) )

Xavier-Do:
@Bond-009 Done, if it is ok for you :)

crobibero (approved):

PR #8946, opened 2022-12-23
Fix image searching to prioritize explicit language match if searching in English

bradbeattie:
**Context**
In the current master branch of Jellyfin, image ordering with a non-English requested language orders like so:

1. Images that match the requested language
2. English images _or_ images with no language
3. All other images

Whereas image ordering with English as the requested language orders like so:

1. Images matching your requested language _or_ images with no language
2. All other images

**Changes**

This change proposes a consistent ordering of images irrespective of the specific requested language:

1. Images that match the requested language
2. Images with no language
3. English images
4. All other images

PR #8982, opened 2022-12-30
Fix backdrop image fetching to prioritize non-language images

bradbeattie:
# Context
#8946 accidentally introduced a side effect of all backdrop images preferring the explicit language of the library. The purpose of backdrop images should actually deprioritize language.

We could additionally modify `OrderByLanguageDescending`, but how this gets used with both manual image searches and during metadata refresh makes it a little slippery to refactor properly. The proposed solution here is fairly simple and gets us the behaviour we're looking for.

# Changes
Image sorting prioritizes non-language images for backdrop metadata refreshes.

### DIFF
--- a/MediaBrowser.Providers/Manager/ItemImageProvider.cs
+++ b/MediaBrowser.Providers/Manager/ItemImageProvider.cs
@@ -313,7 +313,8 @@ namespace MediaBrowser.Providers.Manager
                 }
 
                 minWidth = savedOptions.GetMinWidth(ImageType.Backdrop);
-                await DownloadMultiImages(item, ImageType.Backdrop, refreshOptions, backdropLimit, provider, result, list, minWidth, cancellationToken).ConfigureAwait(false);
+                var listWithNoLangFirst = list.Where(i => string.IsNullOrEmpty(i.Language)).Concat(list.Where(i => !string.IsNullOrEmpty(i.Language)));
+                await DownloadMultiImages(item, ImageType.Backdrop, refreshOptions, backdropLimit, provider, result, listWithNoLangFirst, minWidth, cancellationToken).ConfigureAwait(false);
             }
             catch (OperationCanceledException)
             {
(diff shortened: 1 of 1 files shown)


bradbeattie:
Hmmm. This PR works for explicitly searching for an image, but not for replacing metadata + images...

bradbeattie:
Mixed feelings about the implementation approach, but happy with the resultant behaviour.

Shadowghost (approved):

crobibero:
I think a better change for this is to update `DownloadMultiImages`:

```diff
- foreach (var image in images.Where(i => i.Type == imageType))
+ foreach (var image in images.OrderBy(i => string.IsNullOrWhiteSpace(i.Language)).Where(i => i.Type == imageType))
```

bradbeattie:
> I think a better change for this is to update `DownloadMultiImages`:
> 
> ```diff
> - foreach (var image in images.Where(i => i.Type == imageType))
> + foreach (var image in images.OrderBy(i => string.IsNullOrWhiteSpace(i.Language)).Where(i => i.Type == imageType))
> ```

My concern with that change is that it binds `DownloadMultiImages` to specifically favour non-language images. True, it's only used for that purpose at the moment, but if `imageType` is an option, best to leave the sorting to calling function.

crobibero:
Ok, I would still change the sort method to what I suggested

bradbeattie:
Ah! So the intent behind `listWithNoLangFirst` is to say "I'll observe whatever ordering is defined in `OrderByLanguageDescending`, but otherwise prefer non-language images first". Because https://github.com/jellyfin/jellyfin/blob/master/MediaBrowser.Model/Extensions/EnumerableExtensions.cs#L19 might have complexity unknown to `listWithNoLangFirst`.

I'm unclear if C# ordering with `.OrderBy(i => string.IsNullOrWhiteSpace(i.Language))` will preserve the ordering of elements in its list that are equal in the lambda evaluation, but otherwise unique.

Bond-009 (approved):

PR #9001, opened 2023-01-02
Add stereo downmix algorithm selection.

### LINKED ISSUE #1986: Properly Downmix 5.1 Audio
**Disclaimer**: could not find the ffmpeg downmix options in the source so I do not know the exact state of this, but I've seen the my logs and jellyfin mostly uses the -ac 2 flag.

**Describe the bug**
Default downmix option '-ac 2' in ffmpeg produces audio what is not true to the original 5.1 audio.

**Logs**
An example from my log:
```
 /usr/bin/ffmpeg -ss 00:15:42.000 -fflags +genpts -noaccurate_seek -i file:"/HDD/Sorozatok/Avatar - The Legend of Korra/Season 2/The.Legend.of.Korra.S02E06.The.Sting.720p.HDTV.x264.mkv" -map_metadata -1 -map_chapters -1 -threads 0 -map 0:0 -map 0:1 -map -0:s -codec:v:0 copy -bsf:v h264_mp4toannexb -copyts -vsync -1 -codec:a:0 aac -strict experimental -ac 2 -ab 384000 -af "volume=2" -f hls -max_delay 5000000 -avoid_negative_ts disabled -start_at_zero -hls_time 6 -individual_header_trailer 0 -hls_segment_type mpegts -start_number 157 -hls_segment_filename "/var/lib/jellyfin/transcoding-temp/8e1aac6772385f39f58dd79afd65f4a5%d.ts" -hls_playlist_type vod -hls_list_size 0 -y "/var/lib/jellyfin/transcoding-temp/8e1aac6772385f39f58dd79afd65f4a5.m3u8"
```

**To Reproduce**
Watch a movie with 5.1 what is downmixed to stereo

**Additional context**
There are a couple of alternative formulas tested for 5.1 downmix options [here](https://superuser.com/a/1410620), - while I understand that this could be seen as personal preference - this could be explored further especially when others did the hard work already _(and the default '-ac 2' option discards the subwoofer channel entirely)_.

Shadowghost:
Implements different downmixing algorithms from https://superuser.com/questions/852400/properly-downmix-5-1-to-stereo-using-ffmpeg/1410620#1410620 which were discussed in https://github.com/jellyfin/jellyfin/issues/1986

**Changes**
* Add stereo downmixing algorithms

**Issues**
Fixes #1986

### DIFF
--- a/Jellyfin.Api/Controllers/DynamicHlsController.cs
+++ b/Jellyfin.Api/Controllers/DynamicHlsController.cs
@@ -22,6 +22,7 @@ using MediaBrowser.Controller.Library;
 using MediaBrowser.Controller.MediaEncoding;
 using MediaBrowser.Model.Configuration;
 using MediaBrowser.Model.Dlna;
+using MediaBrowser.Model.Entities;
 using MediaBrowser.Model.IO;
 using MediaBrowser.Model.Net;
 using Microsoft.AspNetCore.Authorization;
@@ -1731,7 +1732,12 @@ namespace Jellyfin.Api.Controllers
 
             var channels = state.OutputAudioChannels;
 
-            if (channels.HasValue)
+            if (channels.HasValue
+                && (channels.Value != 2
+                    || (state.AudioStream is not null
+                        && state.AudioStream.Channels.HasValue
+                        && state.AudioStream.Channels.Value > 5
+                        && _encodingOptions.DownMixStereoAlgorithm == DownMixStereoAlgorithms.None)))
             {
                 args += " -ac " + channels.Value;
             }
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -2129,15 +2129,30 @@ namespace MediaBrowser.Controller.MediaEncoding
 
             var filters = new List<string>();
 
-            // Boost volume to 200% when downsampling from 6ch to 2ch
             if (channels.HasValue
-                && channels.Value <= 2
+                && channels.Value == 2
                 && state.AudioStream is not null
                 && state.AudioStream.Channels.HasValue
-                && state.AudioStream.Channels.Value > 5
-                && !encodingOptions.DownMixAudioBoost.Equals(1))
+                && state.AudioStream.Channels.Value > 5)
             {
-                filters.Add("volume=" + encodingOptions.DownMixAudioBoost.ToString(CultureInfo.InvariantCulture));
+                switch (encodingOptions.DownMixStereoAlgorithm)
+                {
+                    case DownMixStereoAlgorithms.Dave750:
+                        filters.Add("volume=4.25");
+                        filters.Add("pan=stereo|c0=0.5*c2+0.707*c0+0.707*c4+0.5*c3|c1=0.5*c2+0.707*c1+0.707*c5+0.5*c3");
+                        break;
+                    case DownMixStereoAlgorithms.NightmodeDialogue:
+                        filters.Add("pan=stereo|c0=c2+0.30*c0+0.30*c4|c1=c2+0.30*c1+0.30*c5");
+                        break;
+                    case DownMixStereoAlgorithms.None:
+                    default:
+                        if (!encodingOptions.DownMixAudioBoost.Equals(1))
+                        {
+                            filters.Add("volume=" + encodingOptions.DownMixAudioBoost.ToString(CultureInfo.InvariantCulture));
+                        }
+
+                        break;
+                }
             }
 
             var isCopyingTimestamps = state.CopyTimestamps || state.TranscodingType != TranscodingJobType.Progressive;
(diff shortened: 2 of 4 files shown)


nielsvanvelzen (commented):
Does this need to be configured manually in the `encoding.xml` config file or do you plan to expose it in the API so it can be changed in the playback section in the dashboard?

Shadowghost:
If we have a final list of algorithms I plan on implementing a settings drop-down in web for it.

MediaBrowser.Model/Entities/DownMixStereoAlgorithms.cs
```diff
@@ -0,0 +1,24 @@
+namespace MediaBrowser.Model.Entities
```
crobibero, line 1: New files should use file-scoped namespace

MediaBrowser.Model/Configuration/EncodingOptions.cs
```diff
@@ -43,22 +50,54 @@ public EncodingOptions()
...
+        /// </summary>
         public double DownMixAudioBoost { get; set; }
 
+        /// <summary>
+        /// Gets or sets the algorithm used for downmixing audio to stereo.
+        /// </summary>
+        public DownMixStereoAlgorithms DownMixStereoAlgorithm { get; set; }
+
+        /// <summary>
+        /// Gets or sets the maximumu size of the muxing queue.
```
crobibero, line 84:
```suggestion
        /// Gets or sets the maximum size of the muxing queue.
```

MediaBrowser.Model/Configuration/EncodingOptions.cs
```diff
@@ -43,22 +50,54 @@ public EncodingOptions()
...
+        /// </summary>
         public bool EnableThrottling { get; set; }
 
+        /// <summary>
+        /// Gets or sets the delay after which throttling happens.
+        /// </summary>
         public int ThrottleDelaySeconds { get; set; }
 
+        /// <summary>
+        /// Gets or sets the hardware accelleration type.
```
crobibero, line 99:
```suggestion
        /// Gets or sets the hardware acceleration type.
```

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -2129,15 +2129,29 @@ public string GetAudioFilterParam(EncodingJobInfo state, EncodingOptions encodin
...
+                switch (encodingOptions.DownMixStereoAlgorithm)
+                {
+                    case DownMixStereoAlgorithms.Dave750:
+                        filters.Add("volume=4.25");
+                        filters.Add("pan=stereo|c0=0.5*c2+0.707*c0+0.707*c4+0.5*c3|c1=0.5*c2+0.707*c1+0.707*c5+0.5*c3");
+                        break;
+                    case DownMixStereoAlgorithms.NightmodeDialogue:
+                        filters.Add("pan=stereo|c0=c2+0.30*c0+0.30*c4|c1=c2+0.30*c1+0.30*c5");
+                        break;
+                    case DownMixStereoAlgorithms.None:
```
crobibero, line 2147:
```suggestion
                    case DownMixStereoAlgorithms.None:
                    default:
```

crobibero (approved):

Jellyfin.Api/Controllers/DynamicHlsController.cs
```diff
@@ -1731,7 +1732,13 @@ private string GetAudioArguments(StreamState state)
 
             var channels = state.OutputAudioChannels;
 
-            if (channels.HasValue)
+            if (channels.HasValue
+                && (channels.Value != 2
+                    || (channels.Value == 2
```
barronpm, line 1737: No need for this second check on `channels.value`

barronpm (changes requested):

Jellyfin.Api/Controllers/DynamicHlsController.cs
Shadowghost: Removed it.

barronpm (approved):

Bond-009 (approved):

false:
How do you currently use this feature ? It seems it is not visible in the playback menu, and editing the encoding.xml by adding <DownmixStereoAlgorithms>1</DownmixStereoAlgorithms> for exemple, just gets reseted and disappears from the config file.

Shadowghost:
It is only available in unstable and it is a global server setting.

Martan404:
Do you have any idea when we can expect to see it released to stable?

Gauvino:
29 of April 10.9 will be released with this changed in it

Gauvino:
> Do you have any idea when we can expect to see it released to stable?

29 of April 10.9 will be released with this changed in it

RhapsodySL:
> > Do you have any idea when we can expect to see it released to stable?
> 
> 29 of April 10.9 will be released with this changed in it

According to the changelog this is not included as of 10.9 despite being merged to master. If it is present, where is the option specifically? I'm unable to find it.

datsoy:
Is it possible to apply this globally? for all users

gnattu:
> Is it possible to apply this globally? for all users

It is a global option.

PR #9023, opened 2023-01-06
Zip downloads if item has multiple streams

GermanCoding:
**Changes**

Allows to conveniently download additional items (external subtitles, external audio etc) on items. These should now all be downloaded in a single zip.

The ZIP file is generated on the fly (streaming copy). No compression is used for performance, since media files are usually incompressible anyway this should not increase filesizes noticeably.

**Issues**
N/A

### DIFF
--- a/Jellyfin.Api/Controllers/LibraryController.cs
+++ b/Jellyfin.Api/Controllers/LibraryController.cs
@@ -3,8 +3,11 @@ using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
 using System.Globalization;
 using System.IO;
+using System.IO.Compression;
 using System.Linq;
 using System.Net;
+using System.Net.Http.Headers;
+using System.Net.Mime;
 using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
@@ -13,6 +16,7 @@ using Jellyfin.Api.Constants;
 using Jellyfin.Api.Extensions;
 using Jellyfin.Api.ModelBinders;
 using Jellyfin.Api.Models.LibraryDtos;
+using Jellyfin.Api.Results;
 using Jellyfin.Data.Entities;
 using Jellyfin.Data.Enums;
 using Jellyfin.Extensions;
@@ -30,6 +34,7 @@ using MediaBrowser.Model.Configuration;
 using MediaBrowser.Model.Dto;
 using MediaBrowser.Model.Entities;
 using MediaBrowser.Model.Globalization;
+using MediaBrowser.Model.IO;
 using MediaBrowser.Model.Net;
 using MediaBrowser.Model.Querying;
 using Microsoft.AspNetCore.Authorization;
@@ -309,7 +314,7 @@ namespace Jellyfin.Api.Controllers
             }
             catch (Exception ex)
             {
-                 _logger.LogError(ex, "Error refreshing library");
+                _logger.LogError(ex, "Error refreshing library");
             }
 
             return NoContent();
@@ -598,6 +603,7 @@ namespace Jellyfin.Api.Controllers
         /// Downloads item media.
         /// </summary>
         /// <param name="itemId">The item id.</param>
+        /// <param name="includeExternalStreams">Optional. Whether to include external streams (such as subtitles) in the download.</param>
         /// <response code="200">Media downloaded.</response>
         /// <response code="404">Item not found.</response>
         /// <returns>A <see cref="FileResult"/> containing the media stream.</returns>
@@ -606,8 +612,8 @@ namespace Jellyfin.Api.Controllers
         [Authorize(Policy = Policies.Download)]
         [ProducesResponseType(StatusCodes.Status200OK)]
         [ProducesResponseType(StatusCodes.Status404NotFound)]
-        [ProducesFile("video/*", "audio/*")]
-        public async Task<ActionResult> GetDownload([FromRoute, Required] Guid itemId)
+        [ProducesFile("video/*", "audio/*", "application/*")]
+        public async Task<ActionResult> GetDownload([FromRoute, Required] Guid itemId, [FromQuery] bool? includeExternalStreams)
         {
             var item = _libraryManager.GetItemById(itemId);
             if (item is null)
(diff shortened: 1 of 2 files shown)


Uruknara:
I forgot to ask for this but thank for making it its very cool

Jellyfin.Api/Controllers/LibraryController.cs
```diff
@@ -640,7 +643,47 @@ public async Task<ActionResult> GetDownload([FromRoute, Required] Guid itemId)
...
+            {
+                if (!(string.IsNullOrEmpty(stream.Path) || uniqueFiles.Contains(stream.Path)))
+                {
+                    uniqueFiles.Add(stream.Path);
+                }
+            }
+
+            uniqueFiles.Add(item.Path);
+
+            if (uniqueFiles.Count > 1)
```
cvium, line 658: This is a breaking change. It should at least be optional.

Jellyfin.Api/Controllers/LibraryController.cs
```diff
@@ -640,7 +643,47 @@ public async Task<ActionResult> GetDownload([FromRoute, Required] Guid itemId)
             // Quotes are valid in linux. They'll possibly cause issues here.
             var filename = Path.GetFileName(item.Path)?.Replace("\"", string.Empty, StringComparison.Ordinal);
 
-            return PhysicalFile(item.Path, MimeTypes.GetMimeType(item.Path), filename, true);
+            var streams = item.GetMediaStreams();
+            var uniqueFiles = new List<string>();
+            foreach (var stream in streams)
+            {
+                if (!(string.IsNullOrEmpty(stream.Path) || uniqueFiles.Contains(stream.Path)))
```
cvium, line 650: Why not a HashSet?

Jellyfin.Api/Controllers/LibraryController.cs
GermanCoding: Well, currently there is only a single download action. Making this optional is more difficult, as we would need to make it configureable somehow, i.e. by adding new parameters to the download endpoint. This would then require changes to clients to actually send these parameters which also needs new UI probably?

I found the current download option to be not really useful, since the current UI makes downloads of subtitles almost impossible (you have to talk to a SubtitleController endpoint directly, there's no UI). This change at least makes it behave it in a way I would expect from a download point of view. The current download logic already returned arbitrary filetypes (depending on what was stored on disk), so it's only semi-breaking IMHO.

If you have a suggestion on how to make this configurable, preferably without new UI, I'm happy to add switches.

crobibero (changes requested):
This is a breaking change. Either make it configurable or create a new endpoint that always returns a zip.

Jellyfin.Api/Controllers/LibraryController.cs
cvium: Why is a UI change such a bad thing? A simple query parameter like `?includeExternalStreams=true` would make the intention clear and without breaking anything

GermanCoding:
I have added an optional get parameter to toggle this behaviour.

Jellyfin.Api/Controllers/LibraryController.cs
```diff
@@ -640,7 +644,54 @@ public async Task<ActionResult> GetDownload([FromRoute, Required] Guid itemId)
...
+                    uniqueFiles.Add(stream.Path);
+                }
+            }
+
+            uniqueFiles.Add(item.Path);
+
+            if (uniqueFiles.Count > 1)
+            {
+                var filename_zip = (item.FileNameWithoutExtension ?? "download").Replace("\"", string.Empty, StringComparison.Ordinal) + ".zip";
+                return new FileCallbackResult(new MediaTypeHeaderValue("application/octet-stream"), async (outputStream, _) =>
```
crobibero, line 669:
```suggestion
                return new FileCallbackResult(new MediaTypeHeaderValue(MediaTypeNames.Application.Octet), async (outputStream, _) =>
```

Jellyfin.Api/Controllers/LibraryController.cs
```diff
@@ -640,7 +644,54 @@ public async Task<ActionResult> GetDownload([FromRoute, Required] Guid itemId)
...
+                {
+                    uniqueFiles.Add(stream.Path);
+                }
+            }
+
+            uniqueFiles.Add(item.Path);
+
+            if (uniqueFiles.Count > 1)
+            {
+                var filename_zip = (item.FileNameWithoutExtension ?? "download").Replace("\"", string.Empty, StringComparison.Ordinal) + ".zip";
```
crobibero, line 668:
```suggestion
                var filenameZip = (item.FileNameWithoutExtension ?? "download").Replace("\"", string.Empty, StringComparison.Ordinal) + ".zip";
```

Jellyfin.Api/Controllers/LibraryController.cs
```diff
@@ -640,7 +644,54 @@ public async Task<ActionResult> GetDownload([FromRoute, Required] Guid itemId)
...
+                {
+                    using var zipArchive = new ZipArchive(outputStream, ZipArchiveMode.Create);
+                    foreach (var file in uniqueFiles)
+                    {
+                        var zipEntry = zipArchive.CreateEntry(Path.GetFileName(file), CompressionLevel.NoCompression);
+
+                        var zipStream = zipEntry.Open();
+                        await using (zipStream.ConfigureAwait(false))
+                        {
+                            var data = System.IO.File.OpenRead(file);
```
crobibero, line 679:
```suggestion
                            var data = AsyncFile.OpenRead(file);
```

Jellyfin.Api/Results/FileCallbackResult.cs
```diff
@@ -0,0 +1,61 @@
...
+        /// <param name="contentType">The content type.</param>
+        /// <param name="callback">The callback.</param>
+        public FileCallbackResult(MediaTypeHeaderValue? contentType, Func<Stream, ActionContext, Task>? callback)
+            : base(contentType == null ? string.Empty : contentType.ToString())
+        {
+            if (callback == null)
+            {
+                throw new ArgumentNullException(nameof(callback));
+            }
+
```
crobibero, line 31:
```suggestion
            ArgumentNullException.ThrowIfNull(callback);
```

Jellyfin.Api/Results/FileCallbackResult.cs
```diff
@@ -0,0 +1,61 @@
...
+            _callback = callback;
+        }
+
+        /// <inheritdoc/>
+        public override Task ExecuteResultAsync(ActionContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
```
crobibero, line 41:
```suggestion
            ArgumentNullException.ThrowIfNull(context);
```

Jellyfin.Api/Results/FileCallbackResult.cs
```diff
@@ -0,0 +1,61 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Jellyfin.Api.Results
+{
+    /// <summary>
+    /// FileCallbackResult to allow for streaming files with custom actions (i.e. zip compression).
+    /// </summary>
+    public class FileCallbackResult : FileResult
+    {
+        private Func<Stream, ActionContext, Task> _callback;
```
crobibero, line 17:
```suggestion
        private readonly Func<Stream, ActionContext, Task> _callback;
```

Jellyfin.Api/Results/FileCallbackResult.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="FileCallbackResult"/> class.
+        /// </summary>
+        /// <param name="contentType">The content type.</param>
+        /// <param name="callback">The callback.</param>
+        public FileCallbackResult(MediaTypeHeaderValue? contentType, Func<Stream, ActionContext, Task>? callback)
+            : base(contentType == null ? string.Empty : contentType.ToString())
```
crobibero, line 25:
```suggestion
            : base(contentType?.ToString() ?? string.Empty)
```

GermanCoding:
Failing integration test is unrelated to this PR

nielsvanvelzen (changes requested):
I'm not a huge fan of this change because it adds 3 different behaviors to a "simple" download endpoint. I think it makes more sense to add a new endpoint (/Download/Bundle or something like that) that always returns an archive with the files in it configurable with query parameters.

That way we have an endpoint to download "raw" files and one to download files with related assets.

GermanCoding:
That sounds like an interesting idea, but sadly this isn't what I was looking for. If someone would like to implement this, I'm happy to help but in this case I don't think I will pursue this PR any further.

Issue #8780, opened 2022-11-20
[Issue]: Doesn't prevent Windows from sleeping with active streams


lsrom:
### Please describe your bug

Server is running on Windows desktop which is set to sleep after 20 minuts of idle. Jellyfin server should prevent Windows from sleeping when there are active streams.

### Jellyfin Version

Other

### if other:

10.8.5

### Environment

```markdown
- OS: Windows 10
- Virtualization: no
- Clients: Any client
```

### Jellyfin logs

```shell
No logs, this is not a crash just missing functionality.
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

nielsvanvelzen:
This is not a bug but a [feature request](https://features.jellyfin.org/posts/581/prevent-os-from-sleeping-hibernating-while-streaming)

PR #365, opened 2019-01-02
Remove unused PowerManagement

Bond-009:
It isn't up to the application to prevent the system from going to sleep

PR #9043, opened 2023-01-08
Prevent sleep / hibernate while streams are running (Draft)

jonschz:
**Changes**
This is a proof of concept to prevent the server from entering sleep modes while clients are connected and streaming. This missing feature has been demanded multiple times and has also been a problem for me personally.

In case there is interest in merging this feature, I will put more work into cleaning up the code; right now, the code quality is in a draft state. The following needs to be done at least:
- Make the feature an optional setting
- Make the code compile under Linux
- general cleanup

Future features could include
- Proper linux support
- Prevent sleep on encoding / recording

Let me know what you think. I am also aware of power management code in the past, removed here: https://github.com/jellyfin/jellyfin/pull/365

**Issues**

https://github.com/jellyfin/jellyfin/issues/8780
https://features.jellyfin.org/posts/581/prevent-os-from-sleeping-hibernating-while-streaming

**How to test**
- Set your Windows to go to sleep after 1 minute idle, and test if it actually does so.
- Start the modified Jellyfin server and start a stream on a different device.
- Let the stream run for 10 minutes.

With the modified code, the PC only enters sleep mode one minute after I close the stream on the other device. Without the change, the machine goes to sleep while the stream runs.

### DIFF
--- a/Emby.Server.Implementations/Session/SessionManager.cs
+++ b/Emby.Server.Implementations/Session/SessionManager.cs
@@ -7,6 +7,7 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
+using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
 using Jellyfin.Data.Entities;
@@ -40,6 +41,17 @@ using Episode = MediaBrowser.Controller.Entities.TV.Episode;
 
 namespace Emby.Server.Implementations.Session
 {
+    // based on https://stackoverflow.com/a/60512156
+    [Flags]
+    public enum EXECUTION_STATE : uint
+    {
+        ES_AWAYMODE_REQUIRED = 0x00000040,
+        ES_CONTINUOUS = 0x80000000,
+        ES_DISPLAY_REQUIRED = 0x00000002,
+        ES_SYSTEM_REQUIRED = 0x00000001
+        // ES_USER_PRESENT = 0x00000004
+    }
+
     /// <summary>
     /// Class SessionManager.
     /// </summary>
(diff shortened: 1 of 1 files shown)


nielsvanvelzen:
I don't think this feature should be in core. It could work as a plugin though.

kraM1tt:
I'd just like to say, I'm dying for this feature after moving over from Plex.
Energy prices are so high at the moment, I use Wake on LAN to wake my PC when I want to watch something on a TV or phone, then it's set to go to sleep after an hour, but with Plex it would activate powercfg AwayMode to keep it awake until user had logged off, it worked perfectly for me.

jonschz:
In response to your decision I have rewritten my code into a plugin: https://github.com/jonschz/jellyfin-plugin-preventsleep

From my personal user experience I still want to mention that if this feature had existed somewhere in the Jellyfin ecosystem, not only would I have expected it to be part of the core, but also to be enabled by default. It is quite surprising when you stream a full movie for the first time and your host computer suddenly goes into sleep mode, interrupting the stream. Some users may use Google to find that now this can be fixed by a plugin, but other users might be annoyed and look for (or go back to) alternatives. The number of reports and upvotes suggests that I am not the only one surprised by this default behaviour. My temporary solution was to change my power settings every time I start and stop using Jellyfin.

In case your decision remains, I will close this pull request and focus on the plugin for the time being.

nielsvanvelzen:
Jellyfin is maintained by a lot of people and I'm just one of them. My previous comment was my own opinion and others might look at it differently.

We discussed this shortly and one idea we had is to create a plugin specifically for the Windows tray app that contains this functionality. So that would mean that Jellyfin installs on Windows will get the feature without the need to install the plugin manually. A similar thing could be done for MacOS.

kraM1tt:
@nielsvanvelzen Sounds like a great plan to me, I appreciate your efforts, I'd help if I knew how

Bond-009:
Closing this PR here, any possible future work should directed at the windows tray app

PR #8127, opened 2022-07-16
Workaround for linux 5.18+ i915 hang at cost of performance

nyanmisaka:
**Changes**
- Workaround for linux 5.18+ i915 hang at cost of performance

PR #9050, opened 2023-01-09
Update workaround for i915 hang in linux 5.18 to 6.1.3

nyanmisaka:
The issue has been fixed in linux 6.2 recently: https://github.com/torvalds/linux/commit/3f882f2

And the fix was cherry picked into 6.0.18 and 6.1.4 (may be used by debian [bookworm](https://lists.debian.org/debian-kernel/2023/01/msg00066.html)).

**Changes**
- Update workaround for i915 hang in linux 5.18 to 6.1

**Issues**
- Fix the lower FPS on Intel QSV/VAAPI made by #8127

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -35,7 +35,11 @@ namespace MediaBrowser.Controller.MediaEncoding
         private readonly IMediaEncoder _mediaEncoder;
         private readonly ISubtitleEncoder _subtitleEncoder;
         private readonly IConfiguration _config;
-        private readonly Version _minKernelVersioni915Hang = new Version(5, 18);
+
+        // i915 hang was fixed by linux 6.2 (3f882f2)
+        private readonly Version _minKerneli915Hang = new Version(5, 18);
+        private readonly Version _maxKerneli915Hang = new Version(6, 1, 3);
+        private readonly Version _minFixedKernel60i915Hang = new Version(6, 0, 18);
 
         private static readonly string[] _videoProfilesH264 = new[]
         {
@@ -1309,7 +1313,7 @@ namespace MediaBrowser.Controller.MediaEncoding
             // which will reduce overhead in performance intensive tasks such as 4k transcoding and tonemapping.
             var intelLowPowerHwEncoding = false;
 
-            // Workaround for linux 5.18+ i915 hang at cost of performance.
+            // Workaround for linux 5.18 to 6.1.3 i915 hang at cost of performance.
             // https://github.com/intel/media-driver/issues/1456
             var enableWaFori915Hang = false;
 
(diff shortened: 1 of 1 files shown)


1337joe (approved):

Shadowghost (approved):

Bond-009 (approved):

snakeeater4526:
i swear the bug was fixed one week ago, and just today i used it and gpu hangs are back like before

NicholasFlamy:
I installed Jellyfin on TrueNAS very recently and Quicksync hardware acceleration worked for about 2 days before it stopped working and every time I tried to stream transcoded the client would say the media is not compatible. Have to use software transcoding for now (ugh, 90% CPU usage for 4K Blu-ray transcode to 1080p lower bitrate.)
I have an i7 8700K in my Media Server.

nyanmisaka:
> i swear the bug was fixed one week ago, and just today i used it and gpu hangs are back like before @nyanmisaka sorry to bother but can someone see again about this bug?

Please give more contexts about your kernel and distro. I believe the original issue I submitted here is already fixed in Linux 6.2 and some backports version.

https://github.com/intel/media-driver/issues/1456

snakeeater4526:
I actually run my jellyfin on a proxmox VM running ubuntu 22.04.1 with linux 5.15-60 ( was 57 before if i remember ).

I'm 100% sure the bug was fixed 2 weeks ago when the update appeared, and this week ( maybe with the install of the -60 kernel ) it's broken

snakeeater4526:
i'm actually gonna try to force update the kernel to a newer version... but it's clearly a problem for people like me that use ubuntu os as a server

snakeeater4526:
Ok, so the kernel update seems to have fixed the problem . But now the performance is tanked quite a lot and HDR to SDR mapping doesn't seem to work at all... but at least now it work and somewhat usable .

So for people that have the same problem, you have to update you kernel to a version higher than the original, rn i use the latest kernel available on ubuntu ( 6.1.11 )

( also there still this bug that make you unable to force the transcoding if you're stock file have lower bitrate, for exemple a well encoded 4k movie can be 10mbps in AV1, and it's impossible to transcode in h264 4k for device that don't support it .

but that's another unrelated issue ^^ )

snakeeater4526:
IT IS the version i use ^^

nyanmisaka:
If the tonemap speed is slow, you can choose the tonemap algorithm "mobius" to improve the speed.

Also, you can add `options i915 enable_guc=2` in `/etc/modprobe.d/i915.conf` and update initramfs. Then enable the low-power encoder in Jellyfin dashboard to improve performance.

https://jellyfin.org/docs/general/administration/hardware-acceleration/#intel-gen9-and-gen11-igpus

snakeeater4526:
it's not that the tonemap is slow, it's the whole transcode that is slow ( only a 45fps at 1080 10mps) 
and the tonemap is completely broken , even activated in the settings it's doesn't take.

Gonna see for the low power thingy but i use intel core 9th gen encoder ( coffee lake if i remember )

nyanmisaka:
Please share the ffmpeg transcode log.

snakeeater4526:
[transcode log.txt](https://github.com/jellyfin/jellyfin/files/10698960/transcode.log.txt)

nyanmisaka:
You were transcoding an AV1 4k HDR to H264 4k SDR.
So it make sense due to 9th Gen Intel doesn't support AV1 hardware decoding.

snakeeater4526:
decoding yes, but it shouldn't be a problem for hdr encoding anyway and anyway all of that was working before so idk... intel drivers seems a nightmare xD

snakeeater4526:
downgraded to 6.0.19 because there was a bug with 6.1.x was causing crash. but now somehow the hang is back... someday intel drivers will work...

snakeeater4526:
Strangely enough, this problem seems fixed when updated to ubuntu 22.10 ( was 22.04 before with kernel 6.0 ). today i'm still kernel 6.0 but with ubuntu 22.10, and the problem seems gone... strange

PR #8147, opened 2022-07-19
Rewrite NetworkManager and incorporate various fixes from open PRs

Shadowghost:
I tried debugging and fixing a problem but didn't understand all the custom code, so I reimplemented the NetworkManager functionality with native .NET `IPAddress` and `IPNetwork`  and a way more simpler object to hold both (e.g. for interfaces or in case we allow IPs and Subnets in a config)

**Changes**
* Migrate to native .NET IP objects and functions
* Incorporate changes from #5595
* Fix `LocalAddress` resolution in `System/Info/Public`
* Remove hack for .NET issues resolved in .NET 6
* Remove invalid hack for X-Forwarded-For headers when IPv6 is enabled
* Remove unused configuration values from network settings
* Group network settings by functionality (this only affects how they appear in `network.xml`)
* Handle subnets in `KnownProxies`
* Removed `UdpSocket` and replaced it with native .NET sockets
* DLNA/SSDP and AutoDiscovery now only bind to valid bind addresses (WIP)

**Notes**
* DLNA does not require IPs in the `Location` prop but a lot of clients do not work with domains, therefore we force IP+port
* DLNA and SSDP still only support IPv4

**Issues**
* Additional testing is welcome, especially DLNA/SSDP and AutoDiscovery

**Related**
Fixes #6272 
Fixes #9298
Fixes #9300
Partially fixes #8692
Might fix #4442
Might fix #7791
Might fix #9298
Might fix #9445
Might fix #9540
Might fix #9586

PR #9066, opened 2023-01-11
Change log level for bind address usage to debug

### LINKED ISSUE #9024: [Issue]: Jellyfin is spamming stdout with Using BindAddress during a library scan
### Please describe your bug

I have jellyfin running in a docker container. Today I did a full library scan and told it to rebuild everything. `docker logs` is filled with this:

```
[17:59:26] [INF] [104] Jellyfin.Networking.Manager.NetworkManager: 192.168.32.28/32: Using BindAddress https://jellyfin.redacted.net:null
[17:59:26] [INF] [104] Jellyfin.Networking.Manager.NetworkManager: 192.168.32.28/32: Using BindAddress https://jellyfin.redacted.net:null
[17:59:26] [INF] [55] Jellyfin.Networking.Manager.NetworkManager: 192.168.32.28/32: Using BindAddress https://jellyfin.redacted.net:null
[17:59:26] [INF] [55] Jellyfin.Networking.Manager.NetworkManager: 192.168.32.28/32: Using BindAddress https://jellyfin.redacted.net:null
[17:59:27] [INF] [55] Jellyfin.Networking.Manager.NetworkManager: 192.168.32.28/32: Using BindAddress https://jellyfin.redacted.net:null
[17:59:27] [INF] [55] Jellyfin.Networking.Manager.NetworkManager: 192.168.32.28/32: Using BindAddress https://jellyfin.redacted.net:null
[17:59:27] [INF] [55] Jellyfin.Networking.Manager.NetworkManager: 192.168.32.28/32: Using BindAddress https://jellyfin.redacted.net:null
```

### Jellyfin Version

10.8.0

### if other:

_No response_

### Environment

```markdown
- OS:
- Virtualization: Docker
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

barronpm:
Fixes #9024

### DIFF
--- a/Jellyfin.Networking/Manager/NetworkManager.cs
+++ b/Jellyfin.Networking/Manager/NetworkManager.cs
@@ -375,7 +375,7 @@ namespace Jellyfin.Networking.Manager
 
                 if (MatchesPublishedServerUrl(source, isExternal, out string res, out port))
                 {
-                    _logger.LogInformation("{Source}: Using BindAddress {Address}:{Port}", source, res, port);
+                    _logger.LogDebug("{Source}: Using BindAddress {Address}:{Port}", source, res, port);
                     return res;
                 }
             }


crobibero:
I think this only fixes the symptom, not the actual issue.. iirc the issue is that the network list that Jellyfin can bind to keeps changing

barronpm:
Based on the logs, and a quick look through the NetworkManager class, it seems like the problem is that something keeps calling this method, so even for the same source IP, this method gets called repeatedly. Admittedly, it would probably be better to find out where this is happening and look into caching the result, but for now I'm trying to minimize conflicts with #8147. Is there a GitHub issue showing that particular bug?

Shadowghost:
From experience the reason this gets spammed is a failing and restarting docker container.

barronpm:
Unless the issue author cut out other logs, that shouldn't be the case since there is no exception/startup logs

fhriley:
I can provide some color now that I understand where the log message is coming from. I have a scraper that scrapes from Jellyfin. It determines if it needs to log in (since login is expensive) by hitting the system info api endpoint. If it gets a 401, it needs to log in. If not, it can use its existing token. Thus, the spamming was because my scraper was calling that endpoint a lot.

crobibero (approved):

Bond-009 (approved):

PR #5373, opened 2021-03-04
Remove DvdLib

ferferga:
Always query ffprobe for DVD and bluray info

Close #4901

Issue #7786, opened 2022-05-23
[BUG] 10.8.0.0 beta3 DVD playback issue on Rocky Linux (CentOS builds)


jrvmedic:
1. Virgin installation of Rocky Linux 8.6
2. Virgin installation of  ffmpeg v5.0.1 static from ffmpeg.org website. Verfied it can transcode file VOB -> MP4
3.  Virgin installation of JellyFin v10.8.0.0 beta3 for Centos according to instructions on website.
4. Via browser to localhost:8096, added folder with two DVD images to create library, and scanned it.
5. From web browser attempted to play either of the DVDs yields "Playback Error" (see image).
6. Checking the .log shows a transcode problem when manually running the generated command line for the ffmpeg utitlity. (see below)
7. ffmpeg appears to complain that the input paramter (-i file:"/media/dvds/_YOUNG_SHELDON/YOUNG_SHELDON_S01_D2_2017" ) is a directory, and is not a file - Which is entirely true;
8. however, inside the folder (on disc) is the folder "VIDEO_TS", and therein are a dozen files (names VIDEO_TS.BUF/.IFO, etc and many .VOB files. This "dvd image" plays fine when directed to "play disc to VLC"
9. I would have expected JellyFin Server software to have generated an ffmpeg command line with one of the VOB files, or more probably, one of the .IFO files. Perhaps this is considered to be a "bug" in the ffmpeg software, and not a bug in the JellyFin software. But in any event, JellyFin server does not appear to play any DVD images that are put into its libraries.

May-23: I have confirmed that if the ffmpeg command -i parameter is changed from ]]]file:"/media/dvds/_YOUNG_SHELDON/YOUNG_SHELDON_S01_D2_2017"[[[ to ]]]file:"/media/dvds/_YOUNG_SHELDON/YOUNG_SHELDON_S01_D2_2017/VIDEO_TS/VIDEO_TS.VOB"[[[ then the .m3u8 file gets generated, and the .ts's files for it are created. The question therefore is: "What is the correct behavior for Jellyfin Server? Should it leave decoding of the sub-directories to ffmpeg? Or is its behavior correct and the containing folder above it is correct on the command line?"

I am unable to find documentation and/or examples of ffmpeg referencing a "dvd image" aka "a directory image of VOB,IFO, and BUF files", so my guess is that Jellyfin Server is creating the wrong -i parameter for ffmepg - but I cannot find any documentation as to what the new DVD behavior of Jellyfin should be.

[...]

PR #9068, opened 2023-01-11
Remove DvdLib

barronpm:
Based on the investigation done in #5373, the current code for DvdLib is unused, and it doesn't work with Windows paths.

This PR removes the unused/broken code. 

@ferferga, the original PR also removed BDInfo, was that unused as well?

Replaces #5373


Shadowghost (approved):

Shadowghost:
Afaik we're using ffmpeg for everything now.

Roardom:
> Afaik we're using ffmpeg for everything now.

Is this why playback of DVDs has been broken since 10.7.0? (https://github.com/jellyfin/jellyfin/issues/7786)

Bond-009 (approved):

MediaBrowser.Providers/MediaInfo/FFProbeVideoInfo.cs
```diff
@@ -85,40 +84,22 @@ public class FFProbeVideoInfo
...
                     var inputPath = item.Path;
-
                     blurayDiscInfo = GetBDInfo(inputPath);
 
-                    streamFileNames = blurayDiscInfo.Files;
-
-                    if (streamFileNames.Length == 0)
+                    if (blurayDiscInfo.Files.Length == 0)
                     {
                         _logger.LogError("No playable vobs found in bluray structure, skipping ffprobe.");
```
Shadowghost, line 98:
```suggestion
                        _logger.LogError("No playable filesfound in bluray structure, skipping ffprobe.");
```

Shadowghost (changes requested):
After looking at the playback code and fixing metadata retrival for DVD folders locally I came to the conclusion that playing folders outside of an ISO was never supported for DVDs.

It works for BDs because FFmpeg provides a special protocol handler (`bluray://`) for those folders - which is not the case for DVDs. We'd need to handle VOB selection and concatenation ourself to make playing DVDs from folders work (which we currently do not at all).

While doing so I found another function you can remove (also needs to be removed from the interface):
`MediaBrowser.Encoding.Encoder.MediaEncoder.GetPrimaryPlaylistVobFiles`

crobibero (approved):

PR #9071, opened 2023-01-11
Fix client supported image formats

teobaranga:
I recently noticed on the Android TV app that logos for TV shows with transparent backgrounds now had black backgrounds instead. I traced this to `GetClientSupportedFormats` short circuiting if the Accept header was missing.

These changes remove the short-circuit so that at least the jpg and png formats get added by default. Some code is adjusted so that it handles a missing Accept header.


crobibero (approved):

nielsvanvelzen (approved):
I'm working on a client-side fix for jellyfin-androidtv that will set the Accept header to `application/json, application/octet-stream;q=0.9, */*;q=0.8` (https://github.com/jellyfin/jellyfin-sdk-kotlin/pull/638).

Still good to fix it serverside aswell.

Bond-009 (approved):

teobaranga:
Nice, cheers for approving 👍

PR #9076, opened 2023-01-12
Remove unnecesary socket workaround

barronpm:
**Changes**
Removes code working around [ASP.NET Core #14134](https://github.com/aspnet/AspNetCore/issues/14134) not cleaning up unix socket files, as this has been fixed as of .NET 6

### DIFF
--- a/Jellyfin.Server/Program.cs
+++ b/Jellyfin.Server/Program.cs
@@ -339,12 +339,6 @@ namespace Jellyfin.Server
                     {
                         var socketPath = GetUnixSocketPath(startupConfig, appPaths);
 
-                        // Workaround for https://github.com/aspnet/AspNetCore/issues/14134
-                        if (File.Exists(socketPath))
-                        {
-                            File.Delete(socketPath);
-                        }
-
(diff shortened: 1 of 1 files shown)


crobibero:
What happens if Jellyfin crashes? Is the socket file left around? What happens if Jellyfin starts and the socket file exists?

Shadowghost (approved):

barronpm:
> What happens if Jellyfin crashes? Is the socket file left around? What happens if Jellyfin starts and the socket file exists?

If Jellyfin crashes, everything should still be disposed. If this doesn't happen, I would consider it a separate bug.

The behavior changes here occur in two circumstances:
1. Jellyfin is killed unexpectedly (kill -9, power loss, etc.)
2. Another application is configured to use the same socket that Jellyfin is using

In the first case, Jellyfin would fail to start without deleting the file. In the second, currently Jellyfin could potentially delete the socket being used by the other application and create its own socket.

IMO, either behavior is acceptable, but we should remove the comment if we decide to stay with the current behavior.

PR #9077, opened 2023-01-12
Adding a MusicBrainz provider for album images

GaaH:
**Changes**

Album Image provider from MusicBrainz / CoverArtArchive.

**Issues**


crobibero (changes requested):
New files shouldn't have warnings disabled.

GaaH:
I’ve done some changes to stop disabling warnings. Hope it’s better now.

GaaH:
I did not checked the available plugin before submitting this pull request, it seems that [this plugin](https://github.com/jellyfin/jellyfin-plugin-coverartarchive) should do the same thing (well, it does not work on my installation, but I suppose that changes should be done in this repo, not the jellyfin server one)

PR #9082, opened 2023-01-12
Fix unexpected remuxing of direct playable media

dmitrylyzo:
_Alternative to https://github.com/jellyfin/jellyfin-web/pull/3670_

**Changes**
Reset the maximum bitrate if the corresponding transcoding type is blocked for the user.

**Issues**
If the user has set a maximum bitrate and transcoding is blocked after that, direct playable video is never (if its bitrate exeeds the specified one) played direct because of this limit - it is remuxed.

**Steps To Reproduce**
1. With allowed transcoding, set the maximum bitrate in `Settings/Playback` to have the direct playable video transcoded.
2. Block transcoding.
3. Refresh page / relogin.
4. Start video.
5. Remuxing is used instead of DirectPlay. (_No DirectPlay because bitrate is exceeded_)
And there is no way to change this.

### DIFF
--- a/Jellyfin.Api/Helpers/MediaInfoHelper.cs
+++ b/Jellyfin.Api/Helpers/MediaInfoHelper.cs
@@ -239,6 +239,13 @@ namespace Jellyfin.Api.Helpers
 
             options.MaxBitrate = GetMaxBitrate(maxBitrate, user, ipAddress);
 
+            // If the user doesn't have access to transcoding, then ignore the bitrate limit
+            if ((item is Video && !user.HasPermission(PermissionKind.EnableVideoPlaybackTranscoding))
+                || (item is Audio && !user.HasPermission(PermissionKind.EnableAudioPlaybackTranscoding)))
+            {
+                options.MaxBitrate = null;
+            }
+
             if (!options.ForceDirectStream)
             {
                 // direct-stream http streaming is currently broken
--- a/Jellyfin.Api/Helpers/MediaInfoHelper.cs
+++ b/Jellyfin.Api/Helpers/MediaInfoHelper.cs
@@ -237,13 +237,13 @@ namespace Jellyfin.Api.Helpers
                     user.HasPermission(PermissionKind.EnableAudioPlaybackTranscoding));
             }
 
-            options.MaxBitrate = GetMaxBitrate(maxBitrate, user, ipAddress);
-
-            // If the user doesn't have access to transcoding, then ignore the bitrate limit
+            // If the user doesn't have access to transcoding or has disabled it, then ignore the bitrate limit of the client
             if (!mediaSource.SupportsTranscoding)
             {
-                options.MaxBitrate = null;
+                maxBitrate = null;
             }
+
+            options.MaxBitrate = GetMaxBitrate(maxBitrate, user, ipAddress);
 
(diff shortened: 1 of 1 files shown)


Shadowghost (approved):

silentTeee:
Nice fix! I'd like to weigh in as a user and say that this makes perfect sense: if a user can't transcode media, then they can't control the quality, period. Furthermore, having this done server-side seems like the better design choice because otherwise each client would need to address this edge case explicitly

_That said..._ I think that this should probably be paired with client UI changes anyway. My reasoning is that if the user can't transcode, then the clients shouldn't offer an option to select the quality at all (thinking about the new drop-downs that just arrived in the Android mobile client). But that's more of a client implementation detail, and allowing the server to have the final say to defend against a client misconfiguration/design flaw is probably for the best.

dmitrylyzo:
Some notes about the current implementation.

It ignores the global limit for the remote clients.
https://github.com/jellyfin/jellyfin/blob/9a9bf9aa1fe3acbcd8ab39318993fe4c9ea58ac7/Jellyfin.Api/Helpers/MediaInfoHelper.cs#L494
I was thinking to only null the `maxBitrate` (_move the block up and change to the `maxBitrate`_), allowing the server to limit the bandwidth. But we need to decide what to do if the user is not allowed to transcode, but the media exceeds the server limit. Just `media is not supported`?

There is more:
https://github.com/jellyfin/jellyfin/blob/464136cfc9113d33ed4698904a31249f1ed893bf/MediaBrowser.Model/Dlna/AudioOptions.cs#L85-L95
called from:
https://github.com/jellyfin/jellyfin/blob/464136cfc9113d33ed4698904a31249f1ed893bf/MediaBrowser.Model/Dlna/StreamBuilder.cs#L391
https://github.com/jellyfin/jellyfin/blob/464136cfc9113d33ed4698904a31249f1ed893bf/MediaBrowser.Model/Dlna/StreamBuilder.cs#L407
https://github.com/jellyfin/jellyfin/blob/464136cfc9113d33ed4698904a31249f1ed893bf/MediaBrowser.Model/Dlna/StreamBuilder.cs#L101
https://github.com/jellyfin/jellyfin/blob/464136cfc9113d33ed4698904a31249f1ed893bf/MediaBrowser.Model/Dlna/StreamBuilder.cs#L607-L608
and so on.

Shadowghost:
Ideally we need a specific error for this. Otherwise `Media not supported` seems okay to me but it's still a bit hacky.

Jellyfin.Api/Helpers/MediaInfoHelper.cs
```diff
@@ -237,6 +237,12 @@ public class MediaInfoHelper
                     user.HasPermission(PermissionKind.EnableAudioPlaybackTranscoding));
             }
 
+            // If the user doesn't have access to transcoding or has disabled it, then ignore the bitrate limit of the client
+            if (!mediaSource.SupportsTranscoding)
```
dmitrylyzo, line 241: `mediaSource.SupportsTranscoding` is `true` for Live TV regardless of the transcoding permission.

The previous condition also had no effect on Live TV.

Bond-009:
@dmitrylyzo what's the state of this?

Bond-009:
Closing as stale

PR #9093, opened 2023-01-14
Add option to specify user id on user creation

GermanCoding:
The issue being solved here is that custom user management plugins (such as [this ](https://github.com/GermanCoding/jellyfin-server-sync) one) cannot set a Guid when adding a new user - it is always choosen at random. Normally this is good behaviour, but for plugins trying to synchronize data this is problematic.

I did not find a good way to change the user id (shortly) after creation, so I propose to make it configurable (for plugins) instead. Implementations not requiring a specific ID can still operate as-is, API compatibility is maintained.

### DIFF
--- a/Jellyfin.Server.Implementations/Users/UserManager.cs
+++ b/Jellyfin.Server.Implementations/Users/UserManager.cs
@@ -1,4 +1,4 @@
-﻿#pragma warning disable CA1307
+#pragma warning disable CA1307
 
 using System;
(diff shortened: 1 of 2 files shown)


nielsvanvelzen (commented):
User identifiers are supposed to be unique and should not be provided by third party tooling. You can get the generated id when a user is created and store it in your tool to map to a Jellyfin user.

PR #9095, opened 2023-01-15
Feature: Provide service for File Transformation

JPVenson:
**Changes**
Added the `IWebFileTransformationWriteService` for plugins to add in-memory transformations to files provided under the /web prefix. Also added the `IWebFileTransformationReadService` for the newly created `WebFileTransformationService` to consume those changes.

This will allow plugin authors to no longer rely on modifying those files on disk or for the user to do that in the jellyfins web folder. Instead Plugin authors can just inject needed changes in memory in a structured way.

### DIFF
--- a/Jellyfin.Server/Extensions/ApiServiceCollectionExtensions.cs
+++ b/Jellyfin.Server/Extensions/ApiServiceCollectionExtensions.cs
@@ -27,7 +27,9 @@ using Jellyfin.Networking.Configuration;
 using Jellyfin.Server.Configuration;
 using Jellyfin.Server.Filters;
 using Jellyfin.Server.Formatters;
+using Jellyfin.Server.Infrastructure.FileTransformation;
 using MediaBrowser.Common.Net;
+using MediaBrowser.Controller.FileTransformation;
 using MediaBrowser.Model.Entities;
 using MediaBrowser.Model.Session;
 using Microsoft.AspNetCore.Authentication;
@@ -260,6 +262,18 @@ namespace Jellyfin.Server.Extensions
             }
 
             return mvcBuilder.AddControllersAsServices();
+        }
+
+        /// <summary>
+        /// Adds the File Transformation services to the service collection.
+        /// </summary>
+        /// <param name="serviceCollection">The service collection.</param>
+        /// <returns>The updated service collection.</returns>
+        public static IServiceCollection AddJellyfinFileTransformation(this IServiceCollection serviceCollection)
+        {
+            return serviceCollection.AddSingleton<WebFileTransformationService>()
+                .AddSingleton<IWebFileTransformationReadService>(s => s.GetRequiredService<WebFileTransformationService>())
+                .AddSingleton<IWebFileTransformationWriteService>(s => s.GetRequiredService<WebFileTransformationService>());
         }
 
         /// <summary>
--- a/Jellyfin.Server/Infrastructure/FileTransformation/PhysicalTransformedFileProvider.cs
+++ b/Jellyfin.Server/Infrastructure/FileTransformation/PhysicalTransformedFileProvider.cs
@@ -1,0 +1,64 @@
+using System.Buffers;
+using System.IO;
+using MediaBrowser.Controller.FileTransformation;
+using Microsoft.Extensions.FileProviders;
+using Microsoft.Extensions.FileProviders.Embedded;
+using Microsoft.Extensions.FileProviders.Physical;
+using Microsoft.Extensions.Primitives;
+
+namespace Jellyfin.Server.Infrastructure.FileTransformation;
+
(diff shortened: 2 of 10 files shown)


cvium:
I don't really understand the case here. It's for modifying the web client files from plugins? But the web client is packed into bundles with webpack

nielsvanvelzen:
We're already getting enough people reporting "bugs" when in reality they modified jellyfin-web causing those problems. We don't support such modifications and we should not make it easier to do it in an "official" way.
If you want to modify jellyfin-web, consider making an actual pull request to get the feature officially into Jellyfin or maintain your own fork.

JPVenson:
> I don't really understand the case here. It's for modifying the web client files from plugins? But the web client is packed into minified bundles with webpack

The case is for example the Plugin Jellyscrub or IntroSkipper. All want to add additional js files to be loaded. Jellyscrub does this by actually modifing the index.html to add a script Tag and introSkipper requires you to download there version of the web project. Another project has small adaptations like adding avatars for User profiles with buttons and everything. This all does require modification.

JPVenson:
> We're already getting enough people reporting "bugs" when in reality they modified jellyfin-web causing those problems. We don't support such modifications and we should not make it easier to do it in an "official" way. If you want to modify jellyfin-web, consider making an actual pull request to get the feature officially into Jellyfin or maintain your own fork.

I know but i would say as very popular plugins just require this and will always require this and there is an actual usecase for it (adding functionally), i hoped to make, what most do anyway a bit easier. 

I think this might help with that problem even. When the Plugin developer defines the method of modification and not the enduser, the chance is high that it is done in _one_ way and not a thousends like if any user of a plugin tires to modify the index.html. So it would reduce the chance of problems.

nielsvanvelzen:
> I know but i would say as very popular plugins just require this and will always require this and there is an actual usecase for it (adding functionally), i hoped to make, what most do anyway a bit easier.

Plugin developers could also help adding features to the server to avoid this need to modify the web client (which, again, we do not support). Both plugins you mentioned add functionality we want in the server. See https://github.com/jellyfin/jellyfin-meta/discussions/30. We see it as "server provides the HTTP API to get info and plugin API to provide it". We also reached out to the developer of jellyfin-intro-skipper (https://github.com/ConfusedPolarBear/intro-skipper/issues/53).

JPVenson:
> Plugin developers could also help adding features to the server
But dont you degrade plugins to 2nd class citizens with that line of thought? There will always be a plugin that wants to change something that needs code change. 
> avoid this need to modify the web client (which, again, we do not support)
But it is necessary for this type of plugins and there will always be plugins that may not be desired by you to be added to the main repro. They will require users to modify stuff on their own, causing more harm then good.

JPVenson:
Also just looking through the discussion, several people have mentioned that any way of modifying the WebUI is desired and it is a common theme among plugins or snippets for small modifications. This aims to standardize it and make it easier.

nielsvanvelzen:
> But dont you degrade plugins to 2nd class citizens with that line of thought? There will always be a plugin that wants to change something that needs code change.

We only have server-plugins and they are not second class citizens, they are just limited to what they can and cannot do. Modifying the frontend is **not** something they can or should do.

I'm primarily maintaining the Android TV client and regularly users complain that their customizations to jellyfin-web don't affect the Android TV client and it's a bother to constantly explain why that is. Jellyfin is, unlike a lot of other applications, software that needs clients for a lot of different platforms and not all of those use our web interface as a base (Android TV, Roku, Kodi, iOS etc.).

> But it is necessary for this type of plugins and there will always be plugins that may not be desired by you to be added to the main repro. They will require users to modify stuff on their own, causing more harm then good.

Those types of plugins are just not possible, that's just how it is. Users should not modify the jellyfin-web builds. Once again, plugin authors can work with us to get their ideas into the main repositories.

JPVenson:
> Those types of plugins are just not possible, that's just how it is. 

But it is possible and it is done in multiple ways that just causes at lot of problems and is cumbersome. 

> Users should not modify the jellyfin-web builds. Once again, plugin authors can work with us to get their ideas into the main repositories.

I agree that it is not the best way, but the barrier of entry into the jellyfin repro is higher then you might see it. There are some plugins that just want to exist as plugins and do not want to deal with lengthy review processes and slow PRs. So a plugin is the best choice for them. 
As long as jellyfin server bundles with the WebUI, direct modifications will always be a possibility and will be done. I propose to then at least do it "right" to minimize the impact.

cvium:
It is pretty easy to replace the web client. People are not forced to use the bundled web client.

JPVenson:
OK i proposed a more general way of doing client modifications https://github.com/jellyfin/jellyfin-meta/discussions/34 @cvium @nielsvanvelzen would you take a look if that is more what you would support?

PR #9103, opened 2023-01-15
Fix pre-existing chapter images not being deleted

### LINKED ISSUE #9102: Existing Chapters not deleted when setting "Enable chapter image extraction" back to false
### Please describe your bug
Existing video chapters are not removed when toggling the "Enable chapter image extraction" off - even when the "Extract Chapter Images" scheduled task runs. 

Looking through the codebase, it appears there is support for deleting existing chapters when this task runs, as can be seen here: https://github.com/jellyfin/jellyfin/blob/f3c57e6a0ae015dc51cf548a0380d1bed33959c2/Emby.Server.Implementations/MediaEncoder/EncodingManager.cs#L184-L189

However, in the [entrypoint for the scheduled task](https://github.com/jellyfin/jellyfin/blob/f3c57e6a0ae015dc51cf548a0380d1bed33959c2/Emby.Server.Implementations/ScheduledTasks/Tasks/ChapterImagesTask.cs#L91), a query is first made to get existing videos, filtering by those which _do not_ already have chapters: https://github.com/jellyfin/jellyfin/blob/f3c57e6a0ae015dc51cf548a0380d1bed33959c2/Emby.Server.Implementations/ScheduledTasks/Tasks/ChapterImagesTask.cs#L93-L105

I believe this may be the cause of the issue; the videos with pre-existing chapters are never sent downstream to the point where they could then be deleted.

### Steps to reproduce
1. Dashboard -> Libraries -> Select a movie/tv library -> Toggle "Enable chapter image extraction" option _on_
2. Dashboard -> Scheduled Tasks -> Run the "Extract Chapter Images" task
3. Verify chapter images have been generated for any media in the specified library
4. Rrepeat step 1. but toggle the option _off_ this time
5. Repeat step 2.
Repeat step 3. - notice the chapter images have not been deleted

### Environment
Synology DS920+ running Jellyfin in a Docker container

### Jellyfin Version
10.8.8

DarrenRuane:
**Changes**
Change `ChapterImagesTask` to query for videos with _and_ without pre-existing chapter images, rather than just those _without_ them. This makes sure that videos with pre-existing chapter images are sent downstream where they will be deleted if the client has disabled the "Enable chapter image extraction" setting in the dashboard.

**Issues**
Fixes #9102

### DIFF
--- a/Emby.Server.Implementations/MediaEncoder/EncodingManager.cs
+++ b/Emby.Server.Implementations/MediaEncoder/EncodingManager.cs
@@ -63,7 +63,7 @@ namespace Emby.Server.Implementations.MediaEncoder
         /// Determines whether [is eligible for chapter image extraction] [the specified video].
         /// </summary>
         /// <param name="video">The video.</param>
-        /// <param name="libraryOptions"></param>
+        /// <param name="libraryOptions">The library options for the video.</param>
         /// <returns><c>true</c> if [is eligible for chapter image extraction] [the specified video]; otherwise, <c>false</c>.</returns>
         private bool IsEligibleForChapterImageExtraction(Video video, LibraryOptions libraryOptions)
         {
--- a/Emby.Server.Implementations/MediaEncoder/EncodingManager.cs
+++ b/Emby.Server.Implementations/MediaEncoder/EncodingManager.cs
@@ -15,6 +15,7 @@ using MediaBrowser.Controller.Entities;
 using MediaBrowser.Controller.Library;
 using MediaBrowser.Controller.MediaEncoding;
 using MediaBrowser.Controller.Providers;
+using MediaBrowser.Model.Configuration;
 using MediaBrowser.Model.Dto;
 using MediaBrowser.Model.Entities;
 using MediaBrowser.Model.IO;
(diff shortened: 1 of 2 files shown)


crobibero:
I think this also has a side-effect of force the regeneration of chapter images but I haven't had a chance to test yet

DarrenRuane:
> I think this also has a side-effect of force the regeneration of chapter images but I haven't had a chance to test yet

Hey, thanks for the response.
Admittedly I'm not very familiar with the code-base but I believe that this change will not cause existing images to be regenerated as it will only extract (read: generate) images when they don't already exist, thanks to this line: https://github.com/jellyfin/jellyfin/blob/f3c57e6a0ae015dc51cf548a0380d1bed33959c2/Emby.Server.Implementations/MediaEncoder/EncodingManager.cs#L124

I did however notice a bug in my code so I've pushed a change up to fix that.
If you could get a chance to test this and verify that would be greatly apprciated.

1337joe (approved):
I tested this - works as expected, no sign of chapter images that already exist being re-extracted.

Shadowghost (approved):

Bond-009 (approved):

Issue #5576, opened 2021-03-20
Support for multiple video codecs in transcoding profile


dmitrylyzo:
I am trying to fix https://github.com/jellyfin/jellyfin-tizen/issues/69 with https://github.com/jellyfin/jellyfin-web/pull/2519.

This video is transcoded because audio isn't supported - that's right.
```
Video #1
ID                                       : 4113 (0x1011)
Menu ID                                  : 1 (0x1)
Format                                   : HEVC
Format/Info                              : High Efficiency Video Coding
Commercial name                          : HDR10
Format profile                           : Main 10@L5.1@High
Codec ID                                 : 36
Duration                                 : 1 min 2 s
Width                                    : 3 840 pixels
Height                                   : 2 160 pixels
Display aspect ratio                     : 16:9
Frame rate                               : 23.976 (24000/1001) FPS
Color space                              : YUV
Chroma subsampling                       : 4:2:0 (Type 2)
Bit depth                                : 10 bits
Color range                              : Limited
Color primaries                          : BT.2020
Transfer characteristics                 : PQ
Matrix coefficients                      : BT.2020 non-constant
Mastering display color primaries        : Display P3
Mastering display luminance              : min: 0.0050 cd/m2, max: 4000 cd/m2
Maximum Content Light Level              : 1000 cd/m2
Maximum Frame-Average Light Level        : 60 cd/m2

Audio #1
ID                                       : 4352 (0x1100)
Menu ID                                  : 1 (0x1)
Format                                   : AC-3 MLP FBA 16-ch
Format/Info                              : Audio Coding 3 + Meridian Lossless Packing FBA with 16-channel presentation
Commercial name                          : Dolby TrueHD with Dolby Atmos
Muxing mode                              : Stream extension
Codec ID                                 : 131
Duration                                 : 1 min 2 s
Bit rate mode                            : Variable
Bit rate                                 : 640 kb/s
Maximum bit rate                         : 6 183 kb/s
Channel(s)                               : 8 channels
Channel layout                           : L R C LFE Ls Rs Lb Rb
Sampling rate                            : 48.0 kHz
Frame rate                               : 31.250 FPS (1536 SPF)
Bit depth                                : 16 bits
Compression mode                         : Lossy
Stream size                              : 4.79 MiB (1%)
Service kind                             : Complete Main
Number of dynamic objects                : 11
Bed channel count                        : 1 channel
Bed channel configuration                : LFE
```

[...]

PR #9016, opened 2023-01-05
Add control over multiple codec streaming to the device profile

dmitrylyzo:
Using this flag we can disable unnecessary (for some clients) [backward compatibility streams](https://github.com/jellyfin/jellyfin/blob/c795f17fa6ae9e4f68f23cb569a1129de8fd7635/Jellyfin.Api/Helpers/DynamicHlsHelper.cs#L198-L237).

In fact, [this block](https://github.com/jellyfin/jellyfin/blob/c795f17fa6ae9e4f68f23cb569a1129de8fd7635/Jellyfin.Api/Helpers/DynamicHlsHelper.cs#L198-L237) doesn't work.
All streams in the master playlist are bound to the same `PlaySessionId`.
The server starts the transcoding process after receiving a request (_middle bitrate or lower_) from the device.
After sufficient buffering, the device requests a higher bitrate.
Since the `PlaySessionId` is the same, it seems that the transcoding job doesn't restart with the new parameters.

_Adaptive bitrate streaming probably doesn't work either._

**Changes**
Add the `EnableMultipleCodecStreaming` flag to the device profile to control multiple codec streaming

**Issues**
https://github.com/jellyfin/jellyfin-tizen/issues/163
https://github.com/jellyfin/jellyfin/issues/5576
https://github.com/jellyfin/jellyfin-webos/issues/105
https://github.com/jellyfin/jellyfin-web/issues/1190
https://github.com/jellyfin/jellyfin-webos/issues/59

Web part: https://github.com/jellyfin/jellyfin-web/pull/4260

PR #9121, opened 2023-01-18
Make the bitrate of the fallback stream the same as the original

### LINKED ISSUE #8043: [Issue]: HDR lost when audio track needs transcoding
### Please describe your bug

My media files with HVEC 4K HDR video tracks and DTS audio tracks aren't switching my device into HDR mode any longer. All of those files need to go through audio track transcoding, as my device isn't capable of playing DTS tracks.

This used to work thanks to https://github.com/jellyfin/jellyfin/pull/7325 until 10.8.0-beta2. It stopped working since 10.8.0-beta3 and still doesn't work with 10.8.0 and 10.8.1.

It still works if the media file is fully capable of direct play. So, using the same media files with HVEC 4K HDR video tracks but switching to an AC3 Dolby Digital audio track will switch the device into HDR mode as expected.

Please let me know if you need further information.

Related issues: https://github.com/jellyfin/jellyfin/issues/6450, https://github.com/jellyfin/jellyfin/issues/7295

Maybe related issues: https://github.com/jellyfin/jellyfin/issues/8040, https://github.com/jellyfin/jellyfin/issues/7835

### Environment

```markdown
- OS: Ubuntu 
- Virtualization: Docker
- Clients: LG 49UN71006LB. Tested with Jellyfin WebOS App and Stock Browser
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin Version

10.8.1

### Jellyfin logs

[jellyfin.log](https://github.com/jellyfin/jellyfin/files/9026712/jellyfin.log)

### Code of Conduct

dmitrylyzo:
_Alternative to #10751_

In addition to #9092
If HEVC encoding is enabled, we may still encounter unnecessary transcoding issue.

:warning: This PR is a hack/trick. It works on my Tizen 4 TV and Tizen 5 emulator (with `AllowHevcEncoding` condition removed).
I am for removing the problematic block, since we have `VideoRange` and probably no longer need a fallback stream. But this requires testing the Apple devices that require this fallback stream.

**Changes**
Make the bitrate of the fallback stream the same as the original, so that the client can choose by other attributes, such as color range.

**Issues**
Possible unnecessary transcoding (+tone mapping) in the same way as described [here](https://github.com/jellyfin/jellyfin/pull/9016#issue-1521365292).
> All streams in the master playlist are bound to the same `PlaySessionId`.
The server starts the transcoding process after receiving a request (_middle bitrate or lower_) from the device.
After sufficient buffering, the device requests a higher bitrate.
Since the `PlaySessionId` is the same, it seems that the transcoding job doesn't restart with the new parameters.

Fixes #8043
Fixes https://github.com/jellyfin/jellyfin-webos/issues/105
Fixes https://github.com/jellyfin/jellyfin-tizen/issues/163
Fixes https://github.com/jellyfin/jellyfin-webos/issues/59
Fixes https://github.com/jellyfin/jellyfin/issues/5576 (https://github.com/jellyfin/jellyfin/issues/5576#issuecomment-1373901085)
Fixes https://github.com/jellyfin/jellyfin-web/issues/1190

### DIFF
--- a/Jellyfin.Api/Helpers/DynamicHlsHelper.cs
+++ b/Jellyfin.Api/Helpers/DynamicHlsHelper.cs
@@ -211,19 +211,8 @@ public class DynamicHlsHelper
                     var sdrVideoUrl = ReplaceProfile(playlistUrl, "hevc", string.Join(',', requestedVideoProfiles), "main");
                     sdrVideoUrl += "&AllowVideoStreamCopy=false";
 
-                    var sdrOutputVideoBitrate = _encodingHelper.GetVideoBitrateParamValue(state.VideoRequest, state.VideoStream, state.OutputVideoCodec);
-                    var sdrOutputAudioBitrate = 0;
-                    if (EncodingHelper.LosslessAudioCodecs.Contains(state.VideoRequest.AudioCodec, StringComparison.OrdinalIgnoreCase))
-                    {
-                        sdrOutputAudioBitrate = state.AudioStream.BitRate ?? 0;
-                    }
-                    else
-                    {
-                        sdrOutputAudioBitrate = _encodingHelper.GetAudioBitrateParam(state.VideoRequest, state.AudioStream, state.OutputAudioChannels) ?? 0;
-                    }
-
-                    var sdrTotalBitrate = sdrOutputAudioBitrate + sdrOutputVideoBitrate;
-                    AppendPlaylist(builder, state, sdrVideoUrl, sdrTotalBitrate, subtitleGroup);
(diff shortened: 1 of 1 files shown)


nyanmisaka (approved):
Only very few videos require this fallback stream, so bitrate accuracy is not a big concern.

Bond-009 (approved):

crobibero (approved):

PR #9129, opened 2023-01-19
Added replaygain support for transcoded audio

TelepathicWalrus:
Changed the audio transcode settings so that replaygain metatags are utilized when ffmpeg is transcoding.

**Changes**
Add Replaygain support for transcoded audio.

**Issues**
None


TelepathicWalrus:
I would also like to add a toggle for this feature in the settings but to be honest I dont know which files i would need to edit for that. If someone could point me in the right direction then i would be more than happy to give it a go.

PR #9132, opened 2023-01-20
Switched to AsyncKeyedLock

MarkCiliaVincenti:
Switching the existing locking mechanism to use the AsyncKeyedLock library which has been heavily tested and benchmarked, offering superior performance and lower memory allocations (mainly due to the object pooling), while extra care ensured against race conditions.

**Changes**
The Jellyfin.Api now uses a more streamlined approach to locking concurrent tasks by key.


crobibero:
Please add a description to the PR woth reasoning behind this change.

MarkCiliaVincenti:
> Please add a description to the PR woth reasoning behind this change.

done.

MarkCiliaVincenti:
@crobibero ?

MarkCiliaVincenti:
@crobibero Please let me know your intentions so I see whether or not to resolve the conflicts.

cvium:
Personally, I'm not sure I see the benefit.

> offering superior performance and lower memory allocations 

The code paths that are now using your asynckeyedlock aren't exactly hot paths.

MarkCiliaVincenti:
> Personally, I'm not sure I see the benefit.
> 
> > offering superior performance and lower memory allocations
> 
> The code paths that are now using your asynckeyedlock aren't exactly hot paths.


Up to you. With this change you'd still have cleaner code and less stuff to maintain.

MarkCiliaVincenti:
@crobibero @cvium can we have some closure on this please?

MarkCiliaVincenti:
> @crobibero @cvium can we have some closure on this please? What are your intentions? If you want it, I will merge conflicts. If not, we can close the PR.

@crobibero @cvium

EraYaN:
AsyncKeyedLock seems like a fine library but in this case, locking with a string key process wide, wouldn't a named Mutex do essentially the same thing. Especially if the performance is not really all that interesting of a problem here. (Starting a transcode is infinitely more demanding that getting a named Mutex lock or like right now getting a SemaphoreSlim from a dictionary)

MarkCiliaVincenti:
> AsyncKeyedLock seems like a fine library but in this case, locking with a string key process wide, wouldn't a named Mutex do essentially the same thing. Especially if the performance is not really all that interesting of a problem here. (Starting a transcode is infinitely more demanding that getting a named Mutex lock or like right now getting a SemaphoreSlim from a dictionary)

It's not just for transcoding though but also for subtitle encoding and attachment extraction. In fact, the problem with the current implementation is that it does not clean up the dictionary after use and keeps it in memory. Given that this software is sometimes left running continuously, you can see this as a small memory leak. This PR solves this issue.

MarkCiliaVincenti:
@crobibero @cvium @EraYaN @nielsvanvelzen are you interested or shall I close this PR?

MarkCiliaVincenti:
@crobibero @cvium @EraYaN @nielsvanvelzen kindly revert.

JPVenson:
what shall be reverted? @MarkCiliaVincenti

MarkCiliaVincenti:
> what shall be reverted? @MarkCiliaVincenti

I meant revert back to me not reverting of code changes. If you're still interested I will fix the conflicts and update to the latest version of the library.

JPVenson:
i know the lib and i like the usage of the dispose pattern instead of the Barrier one so if you want to keep updating this please.

MarkCiliaVincenti:
Cool, will do so tomorrow then.

MarkCiliaVincenti:
Had to create a new branch and PR for this.

PR #9147, opened 2023-01-21
Remove Emby.Notifications

barronpm:
**Changes**
Removes Emby.Notifications and makes the API a no-op. the only non-archived plugin that uses this is the Webhook plugin, which can be updated to use existing events.


Bond-009 (approved):

nielsvanvelzen:
There is only a single usage in jellyfin-web (https://github.com/jellyfin/jellyfin-web/blob/856e0938cc3b0123fd50cc0638ad78501cbb0762/src/controllers/dashboard/notifications/notification/index.js#L22-L23). The controller can be completely removed instead of deprecated when that usage is removed.

cvium:
Is now the time to say that I think we should use something like Mediator for events? Even the new setup is fairly limiting.

Shadowghost (approved):

crobibero (approved):

PR #9148, opened 2023-01-21
Added replaygain support for transcoded audio

TelepathicWalrus:
Added replaygain support for transcoded audio. Because the audio has to be transcoded to enable replaygain, I have also added another option to force audio transcoding. 

Frontend settings here: https://github.com/jellyfin/jellyfin-web/pull/4294

**Changes**
Added replaygain support for transcoded audio
Added always transcode audio function

**Issues**


TelepathicWalrus:
There seems to be an issue with the current master branch that if you select transcoding in device (ie in finamp settings) the transcoding fails. Using the EnableAlwaysAudioTranscode setting bypasses this and transcoding and replaygain will work.

TelepathicWalrus:
It probably could be done without transcoding but that's a bit beyond my ability. I am like you and wanted replaygain support, or any audio normalisation. I have just implemented it this way as it was easiest for me to do. If anyone has any better suggestions on how to get it working without transcoding please let me know 😊

TelepathicWalrus:
I've given this a second attempt and seem to have it working without transcoding if you are interested. See https://github.com/jellyfin/jellyfin/pull/9222

PR #9175, opened 2023-01-25
Remove OS information from System Info

barronpm:
**Changes**
Removes OS information (Name, processor architecture) from system info endpoints.


Emby.Dlna/Main/DlnaEntryPoint.cs
```diff
@@ -262,7 +263,7 @@ public void StartDevicePublisher(Configuration.DlnaOptions options)
             {
                 _publisher = new SsdpDevicePublisher(
                     _communicationsServer,
-                    MediaBrowser.Common.System.OperatingSystem.Name,
+                    RuntimeInformation.OSDescription,
```
Bond-009, line 266:
```suggestion
                    Environment.OSVersion.Platform.ToString(),
```

Jellyfin.Server/Startup.cs
```diff
@@ -108,7 +109,7 @@ public void ConfigureServices(IServiceCollection services)
                         string.Format(
                             CultureInfo.InvariantCulture,
                             "{0}/{1} UPnP/1.0 {2}/{3}",
-                            MediaBrowser.Common.System.OperatingSystem.Name,
+                            RuntimeInformation.OSDescription,
```
Bond-009, line 112:
```suggestion
                            Environment.OSVersion.Platform,
```

Bond-009 (changes requested):
`RuntimeInformation.OSDescription` gives a bit too much info to put in a user agent

Bond-009 (approved):

Shadowghost (approved):

MediaBrowser.Model/System/SystemInfo.cs
```diff
@@ -137,6 +138,7 @@ public SystemInfo()
         [Obsolete("This isn't set correctly anymore")]
         public FFmpegLocation EncoderLocation { get; set; }
 
+        [Obsolete("This is no longer set")]
         public Architecture SystemArchitecture { get; set; }
```
nielsvanvelzen, line 142: Set a default value / make nullable so a generated SDK doesn't fail when a later release removes the property.

barronpm: Since this field is an enum, it has a default value (32bit for this specific one iirc). How does the default value affect a generated sdk when a property is removed?

nielsvanvelzen: When there is a default value it adds that value to the OpenAPI spec. An SDK will add that default value to it's source. When the server stops sending the key the SDK can fallback to that default value (or null if nullable).

nielsvanvelzen: Right now (10.8.9 OpenAPI spec) there is no default value or nullability.

barronpm: Fixed

nielsvanvelzen (approved):

crobibero (approved):

PR #9176, opened 2023-01-25
Save modified userData after parsing an nfo

### LINKED ISSUE #8415: [Issue]: NFO watched status not imported for movies or episodes
### Please describe your bug

Same issue as the previously resolved #1640 

I am working on migrating my server from an Ubuntu install to a Truenas Scale (using Truecharts) installation and since I already have all metadata and watched status saved in NFO files I tested the import process and it seems like it isn’t reading the watched tags in the NFO file. I disabled all other metadata providers and had disabled saving NFO before importing as well as setting the user for watched status. I import and the majority of data from the NFO is imported correctly but watched is always shown as false on the webui. Curious if there’s anything specific I need to set for it to work or if it might be a bug introduced since the previous issue was fixed. Thanks!

### Jellyfin Version

10.8.0

### if other:

_No response_

### Environment

```markdown
- OS: Truenas Scale
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

ProfessionalismIsMyMiddleName:
**Changes**
The user data object wasn't saved when modified after parsing an nfo, which made the watched, playcount and lastplayed tags functionally useless.  
I had to duplicate a few lines and move them in the switch in order to avoid trying to save new userData for every tag, but that shouldn't be a problem.  
Since it seems odd that this issue wasn't noticed for more than one year and the relevant tests were passing, this warrants a look by someone more experienced with the codebase, but that fixes #8415 for me.
I'm also not sure about the proper use of the "none" cancellation token sent during the data save, but I don't think there's an issue with not being able to cancel the save...

**Issues**
Fixes #8415

### DIFF
--- a/MediaBrowser.XbmcMetadata/Parsers/BaseNfoParser.cs
+++ b/MediaBrowser.XbmcMetadata/Parsers/BaseNfoParser.cs
@@ -266,11 +266,6 @@ namespace MediaBrowser.XbmcMetadata.Parsers
 
             var nfoConfiguration = _config.GetNfoConfiguration();
             UserItemData? userData = null;
-            if (!string.IsNullOrWhiteSpace(nfoConfiguration.UserId))
-            {
-                var user = _userManager.GetUserById(Guid.Parse(nfoConfiguration.UserId));
-                userData = _userDataManager.GetUserData(user, item);
-            }
 
             switch (reader.Name)
             {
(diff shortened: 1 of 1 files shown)


daullmer (approved):

Bond-009 (changes requested):
Can't you just add `_userDataManager.SaveUserData(user, item, userData, UserDataSaveReason.Import, CancellationToken.None);` under the switch statement instead of possibly getting and saving the user data 3 times?

daullmer:
> Can't you just add `_userDataManager.SaveUserData(user, item, userData, UserDataSaveReason.Import, CancellationToken.None);` under the switch statement instead of possibly getting and saving the user data 3 times?

Is this a good idea? It would save the user data for every tag, including title, release year, ... Would this be a huge overhead?

Bond-009:
@daullmer You're right, I missed the fact that this whole function is for 1 node.

Bond-009 (approved):

Shadowghost (approved):

PR #9178, opened 2023-01-25
Escape the path to pass as a command line argument

dmitrylyzo:
**Changes**
Escape the path to pass as a command line argument.

**Issues**
1. Transcoding doesn't work when `datadir` contains quotes in the path.
2. Attachment extraction doesn't work when `datadir` contains spaces in the path.

~_It would be better to extract these `Replace`s into a function. Where?_~
Use `MediaBrowser.MediaEncoding.NormalizePath` (https://github.com/jellyfin/jellyfin/pull/9178#issuecomment-1404675829)

Tested with datadir `jellyfin "test` on Ubuntu 20.04.

### DIFF
--- a/Jellyfin.Api/Controllers/DynamicHlsController.cs
+++ b/Jellyfin.Api/Controllers/DynamicHlsController.cs
@@ -1662,8 +1662,8 @@ namespace Jellyfin.Api.Controllers
                 startNumber.ToString(CultureInfo.InvariantCulture),
                 baseUrlParam,
                 isEventPlaylist ? "event" : "vod",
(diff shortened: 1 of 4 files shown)


Shadowghost (approved):

Bond-009:
We have this function in `MediaBrowser.MediaEncoding`
https://github.com/jellyfin/jellyfin/blob/master/MediaBrowser.MediaEncoding/Encoder/EncodingUtils.cs#L59

crobibero (approved):

Bond-009 (approved):

Issue #9120, opened 2023-01-18
[Issue]: API returns 500 rather than 404 if updating the played status of a non-existent item


fhriley:
### Please describe your bug

By chance I made a call to update the played status of an item that I had already deleted. The server returned `500 internal server error` rather than the `404 not found` I would expect.

```
DEBUG|urllib3.connectionpool|connectionpool.py:456|t:MainThread|https://redacted.com:443 DELETE /Users/d4d75bfbe12f420f9ad25dc7cd1de12a/PlayedItems/49d0dfe77579b7b1a4be184964a456b2 HTTP/1.1" 500 None
```
```
raise HTTPError(http_error_msg, response=self)
    requests.exceptions.HTTPError: 500 Server Error: Internal Server Error for url: https://redacted.com/Users/d4d75bfbe12f420f9ad25dc7cd1de12a/PlayedItems/49d0dfe77579b7b1a4be184964a456b2
```

### Jellyfin Version

Other

### if other:

10.8.8

### Environment

```markdown
- OS:
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

```
[07:12:54] [ERR] [152] Jellyfin.Server.Middleware.ExceptionMiddleware: Error processing request. URL DELETE /Users/d4d75bfbe12f420f9ad25dc7cd1de12a/PlayedItems/49d0dfe77579b7b1a4be184964a456b2.
System.NullReferenceException: Object reference not set to an instance of an object.
   at Jellyfin.Api.Controllers.PlaystateController.MarkUnplayedItem(Guid userId, Guid itemId)
   at lambda_method3845(Closure , Object )
   at Microsoft.AspNetCore.Mvc.Infrastructure.ActionMethodExecutor.AwaitableObjectResultExecutor.Execute(IActionResultTypeMapper mapper, ObjectMethodExecutor executor, Object controller, Object[] arguments)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.<InvokeActionMethodAsync>g__Awaited|12_0(ControllerActionInvoker invoker, ValueTask`1 actionResultValueTask)
[...]

PR #9211, opened 2023-01-30
Add 404 response to MarkPlayedItem/MarkUnplayedItem

chrisb92:
**Changes**
When calling the MarkPlayedItem and MarkUnplayedItem endpoints with an itemId which does not exist, the response should be 404 rather than the current behaviour of throwing a 500 error.

**Issues**
Fixes: #9120


crobibero (approved):

Jellyfin.Api/Controllers/PlaystateController.cs
```diff
@@ -65,21 +66,30 @@ public class PlaystateController : BaseJellyfinApiController
         /// <param name="itemId">Item id.</param>
         /// <param name="datePlayed">Optional. The date the item was played.</param>
         /// <response code="200">Item marked as played.</response>
-        /// <returns>An <see cref="OkResult"/> containing the <see cref="UserItemDataDto"/>.</returns>
+        /// <response code="400">Item not found.</response>
```
crobibero, line 69:
```suggestion
        /// <response code="404">Item not found.</response>
```

Jellyfin.Api/Controllers/PlaystateController.cs
```diff
@@ -91,18 +101,27 @@ public class PlaystateController : BaseJellyfinApiController
         /// <param name="userId">User id.</param>
         /// <param name="itemId">Item id.</param>
         /// <response code="200">Item marked as unplayed.</response>
-        /// <returns>A <see cref="OkResult"/> containing the <see cref="UserItemDataDto"/>.</returns>
+        /// <response code="400">Item not found.</response>
```
crobibero, line 104:
```suggestion
        /// <response code="404">Item not found.</response>
```

chrisb92: Doh! Silly me, that's what I get for copying the line above 😅 Fixed!

Shadowghost (approved):

Bond-009 (approved):

nielsvanvelzen (approved):

PR #9214, opened 2023-01-31
Secure API

Shadowghost:
The goal of this pull request is to secure the API by requiring authentication header or API key request parameter where applicable.
Additionally, proper response codes are added to endpoints if either auth failed or access is forbidden.

Endpoints which should be accessible without auth:
* Global system information
* Public user list
* Image endpoints (otherwise embedding is not possible)
* DLNA endpoints (only from local network) excluding playback endpoints (DLNA controller should return URIs with API key request parameter included)

All other endpoints should require auth by either header or request parameter while deprecating request parameter where possible.

**(Planned) Changes**
- [ ] Migrate API to file scoped namespaces
- [ ] Enforce auth on applicable endpoints
- [x] Add proper responses to endpoints
- [ ] Mark request parameter auth as deprecated where applicable

**Issues**
*To be filled when finished*

### DIFF
--- a/Jellyfin.Api/Controllers/ClientLogController.cs
+++ b/Jellyfin.Api/Controllers/ClientLogController.cs
@@ -38,12 +38,10 @@ public class ClientLogController : BaseJellyfinApiController
     /// Upload a document.
     /// </summary>
     /// <response code="200">Document saved.</response>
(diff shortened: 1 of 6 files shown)


Bond-009:
@Shadowghost What's the status of this?

PR #9219, opened 2023-01-31
Add a way to handle more invalid characters when sanitizing a filename

### LINKED ISSUE #2370: Recorded TV shows may have cryptic folder/file names rather than title of show
I am using JellyFin DVR to record TV shows OTA in the US using Schedules Direct data.

I frequently see irregularly named shows that are a sequence of letters/numbers/symbols. For example, see: 

irregular filenames

Within tvshow.nfo, the correct show name is written (and therefore is known for purposes of writing the show title in the folder/files). Further, the episode jpg shows the correct poster art for the show even though it is misnamed.

```
<?xml version="1.0" encoding="utf-8" standalone="yes"?>
<tvshow>
  <zap2itid>EP00373256</zap2itid>
  <title>Mexico: One Plate at a Time With Rick Bayless</title>
  <mpaa>TV-G</mpaa>
</tvshow>
```

ProfessionalismIsMyMiddleName:
**Changes**
GetValidFilename() Does not properly work for some filesystems.

I noticed the issue because the tmdb boxsets plugin failed to create collections when there was a colon in the collection name ([example collection](https://www.themoviedb.org/collection/661026)).

I traced the issue to Path.GetInvalidFileNameChars() not including all possible problematic characters on all filesystems, as mentioned in [the remarks of the doc](https://learn.microsoft.com/en-us/dotnet/api/system.io.path.getinvalidfilenamechars?view=net-7.0#remarks).
In my case (config on an external ntfs drive on ubuntu 22.04 through docker), the colon prevented the collection directory creation, but wasn't returned by the function.

As it would be impractical to rewrite a better equivalent of the function, I chose to simply add a manually defined array containing the problematic char, this way more can be easily added if more edge cases are discovered.

Also, the alternative option Path.GetInvalidPathChars() has the same problem, so it wouldn't help, but there's no other sane alternative from what I've seen...

**Issues**
Didn't end up creating an issue since I found the cause quickly enough, but I can if needed.
EDIT: 
Fixes #2370


daullmer:
This would fix #2370

1337joe (approved):

cvium (changes requested):
Path.GetInvalidFileNameChars relies on the OS it is running in and not the OS that is "hosting" the file system. I think it would be better to simply copy this list https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/IO/Path.Windows.cs#L15. It hasn't changed in 6 years and we can always update it.

Shadowghost:
@ProfessionalismIsMyMiddleName would be nice if you can address the comment.

ProfessionalismIsMyMiddleName:
I wholeheartedly agree with @cvium's comment, I will update the PR with his proposed character list, it does seem quite exhaustive.

Emby.Server.Implementations/IO/ManagedFileSystem.cs
```diff
@@ -275,7 +275,14 @@ private FileSystemMetadata GetFileSystemMetadata(FileSystemInfo info)
         /// <exception cref="ArgumentNullException">The filename is null.</exception>
         public string GetValidFilename(string filename)
         {
-            var invalid = Path.GetInvalidFileNameChars();
+            // using a character list instead of GetInvalidFileNameChars, as it is not exhaustive and may not return all invalid chars
+            char[] invalid = {
```
crobibero, line 279: Move this to the class level and make it `private static readonly`

ProfessionalismIsMyMiddleName:
@cvium Your requested change has been submitted.

Emby.Server.Implementations/IO/ManagedFileSystem.cs
ProfessionalismIsMyMiddleName: > Move this to the class level and make it `private static readonly`

Done. Sorry about that.

crobibero (approved):

cvium (approved):

Bond-009 (approved):

Emby.Server.Implementations/IO/ManagedFileSystem.cs
```diff
@@ -20,7 +20,13 @@ public class ManagedFileSystem : IFileSystem
...
         private static readonly bool _isEnvironmentCaseInsensitive = OperatingSystem.IsWindows();
-
+        private static readonly char[] _invalidPathCharacters = {
+                '\"', '<', '>', '|', '\0',
+                (char)1, (char)2, (char)3, (char)4, (char)5, (char)6, (char)7, (char)8, (char)9, (char)10,
+                (char)11, (char)12, (char)13, (char)14, (char)15, (char)16, (char)17, (char)18, (char)19, (char)20,
+                (char)21, (char)22, (char)23, (char)24, (char)25, (char)26, (char)27, (char)28, (char)29, (char)30,
+                (char)31, ':', '*', '?', '\\', '/'
+            };
         /// <summary>
```
Bond-009, line 30:
```suggestion
        private static readonly char[] _invalidPathCharacters =
        {
            '\"', '<', '>', '|', '\0',
            (char)1, (char)2, (char)3, (char)4, (char)5, (char)6, (char)7, (char)8, (char)9, (char)10,
            (char)11, (char)12, (char)13, (char)14, (char)15, (char)16, (char)17, (char)18, (char)19, (char)20,
            (char)21, (char)22, (char)23, (char)24, (char)25, (char)26, (char)27, (char)28, (char)29, (char)30,
            (char)31, ':', '*', '?', '\\', '/'
         };

        /// <summary>
```

Bond-009 (commented):
Debug build is failing

PR #9222, opened 2023-01-31
Audio normalization

TelepathicWalrus:
This is my second attempt at audio normalization. This uses FFmpeg to get the LUFS using the ebur128 filter. This is then is stored in the library database for later use by the htmlaudioplayer. 

Please let me know of any suggestions or oversights. 

Thanks!

Frontend here: https://github.com/jellyfin/jellyfin-web/pull/4318

**Changes**
Added LUFS data for audio files
Added LUFS to dto
Added LUFS grabber

**Issues**


Sjoerdfc:
It looks like this is working for me. I compiled & installed deb's on Ubuntu 20.04, and I have the checkbox in my user settings enabled. I'm playing through the web interface.

I did have to completely remove my library I think. After just a refresh, and selecting replace all metadata, I still could not hear a difference. But after removing it, and adding it again, I did. It sounds the same when using a separate player using replaygain now. I also see the LUFS key in the TypedBaseItems table in my database with all the values. Maybe I will check this again if works after just a refresh on a fresh install.

I normally use Gelli on Android to play music through Jellyfin, and am I correct in assuming this needs changes in each player to work? After installing this version Gelli doesn't work anymore, I just get a message "Couldn't play this song". But I think this is a problem in Gelli, because another android player, Finamp, does work. I just don't hear the normalization.

TelepathicWalrus:
Thats great to hear! Yeah so it will send the LUFS in the streaming dto so it shouldn't stop other players working, they will just receive the LUFS value for the track but this doesn't mean they have to use it. I'm not familiar with gelli but i would guess that with it being a native player extra support will need to be added. This will be the same for Finamp, it will need to be updated so that it uses this data.

Sjoerdfc:
And if it's implemented in a player it will work with and without transcoding?

TelepathicWalrus:
Yeah that's the plan. All the hard work has been done calculating the LUFS before hand, so once this data has been received by the player, the gain can be adjusted locally :)

Sjoerdfc:
And just curious, why do you scrape the LUFS data? Wasn't it easier just to read the replaygain data if already present, store that in the database, and send to the players?

TelepathicWalrus:
This is what i tried to do initially but i figured that not all music in a library will have the data so it would be just as easy to calculated it on the first probe. It may take longer, but it is only once and it will get the data for any audio file.

Sjoerdfc:
Yesterday when I said this was working for me, I deleted my entire library, and added only about 20 songs. This worked fine.

Today I started over with my original Jellyfin installation, and installed this version over it. I refreshed my entire library of ~10.000 songs, and selected scan for missing metadata. I checked the database, and saw the LUFS column was added, and after a while I saw a few values starting to appear. But now, after over 2 hours, it only scanned about 500 songs. At this rate, it will take two days to scan my entire library. So I'm not sure if scanning LUFS data is usable for big libraries of music. 

In comparison, scanning the exact same library (but over the network, so it should be even slower) with r128gain for replaygain data takes only 45 minutes or something.

TelepathicWalrus:
Interesting, my library is a lot smaller, only about 700 songs, but on my server that takes about 2-3 minutes. If you htop on your server can you see all the instances of ffmpeg running?

Sjoerdfc:
I stopped the scan, added more RAM and cores to my VM (its at 2.5GB ram, 6 cores now), and started the scan again.

When I start scanning I do see a few ffmpeg processes with 'ebur128' somewhere, so that should be it. But that stops after a few minutes of scanning, and then nothing. When I copy the library.db, open it, and sort by LUFS value to check the progress, nothing changed. It was at 781 when I stopped the scan earlier, and it still is now. In the webinterface the scan is stuck at 52%. So it seems like it completely stopped working now.

I don't know if this is a problem on my side or something else. Maybe wait for other people to test this.

TelepathicWalrus:
Okay yeah im running it in docker on my machine, so it has full access to all ram and 16 cores. I know it is reasonably intensive when it gets the data but it shouldn't grind to a halt. It seems by looking at my htop that ram shouldn't be the issue but threads are a key part.

TelepathicWalrus:
Just out of interest did you delete the library.db and retry or was it from your last installation? Just wondering if there was some sort of sql problem.

Sjoerdfc:
This was my original, old,  Jellyfin install, with my exisiting 10.000 song library, which I updated to your version, and refreshed the library.

TelepathicWalrus:
Ok so I've done some testing with a vm and building/using a docker image in the vm. I set the vm to have 1 thread and 2048MB of ram and it took around 10-15 minutes to search my library of around 700 songs. The only think I noticed when I was doing this was that some of the replaygain tags were not matching the tags of the database because of the ffmpeg arguments. I have updated these arguments so maybe this will help with your situation but I'm not sure. If you have a lot of high bitrate audio files then this should help if not then maybe not so much. The only other thing I can think of is to backup your library.db and then let jellyfin remake a whole new one and see how that goes.

Collcroc123:
Hi I'm a complete newbie to this, how would I get this set up on my unRAID server? I already have the official Jellyfin docker installed and set up. I assume I have to install this as a new docker, and if so, how?

TelepathicWalrus:
Hi, 
[This](https://jellyfin.org/docs/general/installation/source) has good information with how to build jellyfin from source, you just need to do an extra step. I am unfamiliar with unraid and how to specifically build it on that system but i assume it would be very similar to what I've outlined below.

1. Download repository with:
```
git clone -b audio-normalization https://github.com/TelepathicWalrus/jellyfin.git
cd jellyfin
```
2. Initialize the submodules with:
```
git submodule update --init
```
3. Edit Dockerfile with your editor of choice and change line 8 so that it reads:
```
ARG JELLYFIN_WEB_VERSION=audio-normalization
```
and change line 10 so that it reads:
```
 && curl -L https://github.com/TelepathicWalrus/jellyfin-web/archive/${JELLYFIN_WEB_VERSION}.tar.gz | tar zxf - \
```
This will change the front end to the modified version which allows you to enable/disable normalization.

4. Run `docker build -t $USERNAME/jellyfin .` to build the docker container. This will take some time

6. The docker container will now be ready to use with name "your username"/jellyfin (if your current username is 'jeff' the image would be jeff/jellyfin). You can edit your docker-compose file to run with this image or you can run it with `docker run -d -p 8096:8096 $USERNAME/jellyfin`

Hopefully this helps

Thanks

Bond-009:
Debug build fails with `MediaBrowser.Providers/MediaInfo/AudioFileProber.cs(9,1): error SA1210: Using directives should be ordered alphabetically by the namespaces`

TelepathicWalrus:
> Debug build fails with `MediaBrowser.Providers/MediaInfo/AudioFileProber.cs(9,1): error SA1210: Using directives should be ordered alphabetically by the namespaces`

Changed now

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
```diff
@@ -89,6 +98,34 @@ public class AudioFileProber
...
             }
 
+            string ffmpeg_path = _serverConfigurationManager.GetEncodingOptions().EncoderAppPathDisplay;
+            string args = "-hide_banner -i \"";
+            args += path;
+            args += "\" -af ebur128=framelog=verbose -f null -";
+
+            ProcessStartInfo ffmpeg = new ProcessStartInfo(ffmpeg_path);
+
+            System.Diagnostics.Process proc = new System.Diagnostics.Process();
```
Bond-009, line 108: use a `using` statement so this gets properly disposed

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
```diff
@@ -89,6 +98,34 @@ public class AudioFileProber
                 Fetch(item, result, cancellationToken);
             }
 
+            string ffmpeg_path = _serverConfigurationManager.GetEncodingOptions().EncoderAppPathDisplay;
```
Bond-009, line 101:
```suggestion
            string ffmpeg_path = _mediaEncoder.EncoderPath;
```

Bond-009 (changes requested):

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
TelepathicWalrus: i have added this with latest commit, please let me know if ok

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
TelepathicWalrus: i have added this with latest commit, please let me know if ok

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
```diff
@@ -89,6 +98,34 @@ public class AudioFileProber
...
+            string args = "-hide_banner -i \"";
+            args += path;
+            args += "\" -af ebur128=framelog=verbose -f null -";
```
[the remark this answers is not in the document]
TelepathicWalrus: i have added this with latest commit, please let me know if ok

Alistair1231:
Any updates on this?

MediaBrowser.Model/Dto/BaseItemDto.cs
```diff
@@ -779,6 +779,12 @@ public class BaseItemDto : IHasProviderIds, IItemDto, IHasServerId
         /// <value>The timer identifier.</value>
         public string TimerId { get; set; }
 
+        /// <summary>
+        /// Gets or sets the LUFS value.
+        /// </summary>
+        /// <value>The LUFS Value.</value>
+        public string LUFS { get; set; }
```
nielsvanvelzen, line 786: Shouldn't this be an integer/double/float? What kind of values can this be?

TelepathicWalrus:
Honestly don't know, I have requested a review from @Bond-009 but haven't heard anything.

MediaBrowser.Model/Dto/BaseItemDto.cs
TelepathicWalrus: Yeah i guess it should, tbh I left it as string as the output from ffmpeg is string.

TelepathicWalrus: I cannot update it ATM as I'm away from my pc but I probably can by Wednesday? If the reviewers can update it that would be great but if not then I'll test/update when I'm back

TelepathicWalrus: Okay so i managed to get hold of someones laptop and have changed the LUFS value to float from string. I have also added the option to disable the LUFS scan in the library options if people dont want this feature

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
```diff
@@ -89,6 +91,38 @@ public class AudioFileProber
                 Fetch(item, result, cancellationToken);
             }
 
+            var libraryOptions = _libraryManager.GetLibraryOptions(item);
+
+            if (libraryOptions.EnableLUFSScan)
+            {
+                string output;
+                using (System.Diagnostics.Process proc = new System.Diagnostics.Process())
```
cvium, line 99: Very minor thing, but you should follow the "coding style standards" for how we use Process. You can search around to find it.

TelepathicWalrus: Do you mean adding "using System.Diagnostics" to the top of the file and then changing line 99 so it reads "using (Process proc = new Process())"? if so i can implement that now.

cvium: No, I mean this https://github.com/jellyfin/jellyfin/blob/910617bbc3b960ff6a1e1c18c27a0b42a4be9ee0/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs#L398-L406. How we initialize it and the extra options.

TelepathicWalrus: Updated now please advise if it is not to code style standards. Thanks! :)

CONTRIBUTORS.md
```diff
@@ -224,6 +225,7 @@
  - [Detector1](https://github.com/Detector1)
  - [BlackIce013](https://github.com/blackice013)
  - [mporcas](https://github.com/mporcas)
+ - [TelepathicWalrus](https://github.com/TelepathicWalrus)
```
1337joe, line 228: This is the section for contributors to the Emby source code from before it was forked by Jellyfin. You should only be adding your name to the Jellyfin section.
```suggestion
```

cvium (approved):

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
```diff
@@ -89,6 +97,54 @@ public class AudioFileProber
...
+                    cancellationToken.ThrowIfCancellationRequested();
+                    MatchCollection split = Regex.Matches(output, @"I:\s+(.*?)\s+LUFS");
+
+                    if (split.Count != 0)
+                    {
+                        item.LUFS = float.Parse(split[0].Groups[1].ToString(), CultureInfo.InvariantCulture.NumberFormat);
+                    }
+                    else
+                    {
+                        item.LUFS = -18;
```
crobibero, line 137: I would move this `-18` to a const somewhere, especially since `-18` doesn't mean anything to me

TelepathicWalrus: I have added a const float for defualt(no db gain) LUFS value

crobibero (approved):

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
```diff
@@ -89,6 +100,54 @@ public class AudioFileProber
...
+                        throw;
+                    }
+
+                    output = await process.StandardError.ReadToEndAsync(cancellationToken).ConfigureAwait(false);
+                    cancellationToken.ThrowIfCancellationRequested();
+                    MatchCollection split = Regex.Matches(output, @"I:\s+(.*?)\s+LUFS");
+
+                    if (split.Count != 0)
+                    {
+                        item.LUFS = float.Parse(split[0].Groups[1].ToString(), CultureInfo.InvariantCulture.NumberFormat);
```
Bond-009, line 136:
```suggestion
                        item.LUFS = float.Parse(split[0].Groups[1].ValueSpan, CultureInfo.InvariantCulture.NumberFormat);
```
This should work and prevent a string allocation, no?

Bond-009 (changes requested):

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
TelepathicWalrus: Sure, quickly tried that and works fine if you think that'll be better :)

TelepathicWalrus:
Sorry @cvium, I accidentally double clicked for a review from bond :P

Bond-009 (approved):

lonebyte:
Thank you so much for this!

I have one follow-up question though. On albums with differing LUFS values (and especially without gaps between the tracks) this leads to jumps in the volume. Are you planning to implement a setting for using same LUFS for all tracks of an album (also called album gain)? This can be archived by simply using the lowest value of one particular album for all it's tracks.

If you don't plan to, I might be able to implement it, just let me know.

TelepathicWalrus:
Hey @lonebyte I've put in a pull request for this feature, if you wanted to download the frontend and backend and try it out :+1: 
https://github.com/jellyfin/jellyfin/pull/10085

Chaphasilor:
@Sjoerdfc just wanted to let you know that I'm in the process of implementing audio normalization for Finamp and there will be a beta with this feature soon. If you want me to, I could let you know when this is ready so you can try it out? :)

Sjoerdfc:
Hi @Chaphasilor, thats great. So you're using the LUFS values from this PR? Yeah, I'd be willing to test it if you want.

Chaphasilor:
Yes, exactly. Works with transcoding as well. I haven't implemented the Album Gain yet, and so far it only works on Android. I'll let you now when there's something to test! 😁

kdltr:
Is there, or would it be possible to get the option to get the loudness data from the ReplayGain tags? I’d rather not go through rescaning my 35k+ songs library ^^'

TelepathicWalrus:
> Is there, or would it be possible to get the option to get the loudness data from the ReplayGain tags? I’d rather not go through rescaning my 35k+ songs library ^^'

Okay so I've created a pull request which adds this. The only reason why I was hesitant to add it was for consistency sake. If the replaygain software didn't use the same method (although they all should) then some tracks could be very different volumes. Anyways, like i say pull request is up, please test and see what you think :+1:

https://github.com/jellyfin/jellyfin/pull/10566

Please note you will require new front end too which is linked in pull request.

Thanks

andzejsp:
can we have this on video media too?

KomGiann:
Please do something about movies and series too , right now I'm constantly adjusting volume to match each scene

NicholasFlamy:
Pardon me for the confusion, but does the current state of this feature and Jellyfin generate the data needed for audio normalization? Or do I need to use a separate tool to generate that data so that audio normalization works in Jellyfin?

Issue #1907, opened 2019-10-17
Unable to find linked item at path


ferferga:
**Describe the bug**
Changing the filenames or path of some items added in a collection results in "orphaned" items in the collection. This error is represented in the logs at Warning level like this:
``[WRN] Unable to find linked item at path path/to/the/original/file``

**To Reproduce**
1. Add items to a collection
2. Change the path or name of those items in your filesystem
3. Do an scan. Items will be automatically identified if you didn't change the name schema in a way that isn't recognizable by Jellyfin. If that's the case, Jellyfin will add the items to the original collection again.
4. Reboot the server and scan again (optional, but to make sure that Jellyfin is fully "aware" of the change)
5. See the logs after that last scan and grep with the filename of the old file: You will see that, although it handled the filename change gracefully in the first scan, it's still expecting the old file to be there, as that error just keeps spawning.

No matter how many library scans you do afterwards, those files will be stuck forever and Jellyfin will always be expecting them, although the user interface in all apps show the collection pointing to the new path correctly.

**Expected behavior**
Jellyfin forgets everything about all the old files in collections

**System (please complete the following information):**
 - OS: Debian Buster in Raspberry Pi
 - Jellyfin Version: [e.g. 10.4.0]
 - Reverse proxy: nginx

**Additional context**

I haven't tested this, but I think that probably this is happening if you just remove an item from the filesystem, as Jellyfin, when a file is altered/moved from the filesystem, removes it and adds it again.

anthonylavado:
In the case of music, this error can come up due to https://github.com/jellyfin/jellyfin/issues/1874

ferferga:
@anthonylavado In my case, it happened only with movies

snis:
Exakt same issue with removed media 
`[WRN] Unable to find linked item at path ...`

- OS: Arch linux x64
- Jellyfin v 10.4.3

Deepcuts:
Same problem here on 10.5.2

manzajaro:
> Same problem here on 10.5.2

me too,

OS: Linux  Arm64
Jellyfin v 10.5.2

dkanada:
I have looked into the scanning before and we have helper classes that should clean missing files and such from the database. I was looking at the album artist one but there might be one for other objects as well.

cognicom:
[...]

PR #9231, opened 2023-02-01
Added CleanupCollection task

JPVenson:
**Changes**
Added Task to run at startup to cleanup collection items that are not longer physical present.
Updated interface of ICollectionManager to get reference to the Boxsets location to prevent duplicate magic strings.

**Issues**
#1907

### DIFF
--- a/Emby.Server.Implementations/ScheduledTasks/Tasks/CleanupCollectionPathsTask.cs
+++ b/Emby.Server.Implementations/ScheduledTasks/Tasks/CleanupCollectionPathsTask.cs
@@ -1,0 +1,118 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Emby.Server.Implementations.Collections;
+using MediaBrowser.Controller.Collections;
+using MediaBrowser.Controller.Entities;
+using MediaBrowser.Controller.Entities.Movies;
+using MediaBrowser.Controller.Library;
+using MediaBrowser.Controller.Providers;
+using MediaBrowser.Model.Globalization;
+using MediaBrowser.Model.IO;
+using MediaBrowser.Model.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Emby.Server.Implementations.ScheduledTasks.Tasks;
+
+/// <summary>
+/// Deletes Path references from collections that no longer exists.
+/// </summary>
+public class CleanupCollectionPathsTask : IScheduledTask
+{
+    private readonly ILocalizationManager _localization;
+    private readonly ICollectionManager _collectionManager;
+    private readonly ILogger<CleanupCollectionPathsTask> _logger;
+    private readonly IProviderManager _providerManager;
+    private readonly IFileSystem _fileSystem;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="CleanupCollectionPathsTask"/> class.
+    /// </summary>
+    /// <param name="localization">Instance of the <see cref="ILocalizationManager"/> interface.</param>
+    /// <param name="collectionManager">Instance of the <see cref="ICollectionManager"/> interface.</param>
+    /// <param name="logger">The logger.</param>
+    /// <param name="providerManager">The provider manager.</param>
+    /// <param name="fileSystem">The filesystem.</param>
+    public CleanupCollectionPathsTask(
+        ILocalizationManager localization,
+        ICollectionManager collectionManager,
+        ILogger<CleanupCollectionPathsTask> logger,
+        IProviderManager providerManager,
+        IFileSystem fileSystem)
+    {
+        _localization = localization;
+        _collectionManager = collectionManager;
+        _logger = logger;
+        _providerManager = providerManager;
+        _fileSystem = fileSystem;
+    }
+
+    /// <inheritdoc />
+    public string Name => _localization.GetLocalizedString("TaskCleanCollections");
+
+    /// <inheritdoc />
+    public string Key => "CleanCollections";
+
+    /// <inheritdoc />
+    public string Description => _localization.GetLocalizedString("TaskCleanCollectionsDescription");
+
+    /// <inheritdoc />
+    public string Category => _localization.GetLocalizedString("TasksMaintenanceCategory");
+
+    /// <inheritdoc />
+    public async Task ExecuteAsync(IProgress<double> progress, CancellationToken cancellationToken)
+    {
+        var collectionsFolder = await _collectionManager.GetCollectionsFolder(true).ConfigureAwait(false);
+        if (collectionsFolder is null)
+        {
+            _logger.LogInformation("There is no collection folder to be found.");
+            return;
+        }
+
+        var collections = collectionsFolder.Children.OfType<BoxSet>()
+            .ToArray();
+        _logger.LogTrace("Found {CollectionLength} Boxsets.", collections.Length);
+        for (var index = 0; index < collections.Length; index++)
+        {
+            var collection = collections[index];
+            _logger.LogTrace("Check Boxset {CollectionName}.", collection.Name);
+            var itemsToRemove = new List<LinkedChild>();
+            foreach (var collectionLinkedChild in collection.LinkedChildren.ToArray())
+            {
+                if (!File.Exists(collectionLinkedChild.Path))
+                {
+                    _logger.LogInformation("Item in boxset {0} cannot be found at {1}.", collection.Name, collectionLinkedChild.Path);
+                    itemsToRemove.Add(collectionLinkedChild);
+                }
+            }
+
+            if (itemsToRemove.Any())
+            {
+                _logger.LogTrace("Update Boxset {CollectionName}.", collection.Name);
+                collection.LinkedChildren = collection.LinkedChildren.Except(itemsToRemove).ToArray();
+                await collection.UpdateToRepositoryAsync(ItemUpdateType.MetadataEdit, cancellationToken)
+                    .ConfigureAwait(false);
+
+                _providerManager.QueueRefresh(
+                    collection.Id,
+                    new MetadataRefreshOptions(new DirectoryService(_fileSystem))
+                    {
+                        ForceSave = true
+                    },
+                    RefreshPriority.High);
+            }
+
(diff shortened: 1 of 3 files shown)


Emby.Server.Implementations/ScheduledTasks/Tasks/CleanupCollectionPathsTask.cs
```diff
@@ -0,0 +1,117 @@
...
+    /// <inheritdoc />
+    public string Category => _localization.GetLocalizedString("TasksMaintenanceCategory");
+
+    /// <inheritdoc />
+    public async Task ExecuteAsync(IProgress<double> progress, CancellationToken cancellationToken)
+    {
+        var collectionsFolder = await _collectionManager.GetCollectionsFolder(false).ConfigureAwait(false);
+        if (collectionsFolder is null)
+        {
+            _logger.LogInformation("There is no collection folder to be found.");
```
crobibero, line 71:
```suggestion
            _logger.LogDebug("There is no collection folder to be found");
```

Emby.Server.Implementations/ScheduledTasks/Tasks/CleanupCollectionPathsTask.cs
```diff
@@ -0,0 +1,117 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Emby.Server.Implementations.Collections;
```
crobibero, line 7:
```suggestion
```

unused

Emby.Server.Implementations/ScheduledTasks/Tasks/CleanupCollectionPathsTask.cs
```diff
@@ -0,0 +1,117 @@
...
+    {
+        var collectionsFolder = await _collectionManager.GetCollectionsFolder(false).ConfigureAwait(false);
+        if (collectionsFolder is null)
+        {
+            _logger.LogInformation("There is no collection folder to be found.");
+            return;
+        }
+
+        var collections = collectionsFolder.Children.OfType<BoxSet>().ToArray();
+        _logger.LogTrace("Found {CollectionLength} Boxsets.", collections.Length);
```
crobibero, line 76:
```suggestion
        _logger.LogTrace("Found {CollectionLength} Boxsets", collections.Length);
```

`Trace` includes very verbose logging, I would stick to `Debug`

Emby.Server.Implementations/ScheduledTasks/Tasks/CleanupCollectionPathsTask.cs
```diff
@@ -0,0 +1,117 @@
...
+
+        for (var index = 0; index < collections.Length; index++)
+        {
+            var collection = collections[index];
+            _logger.LogTrace("Check Boxset {CollectionName}.", collection.Name);
```
crobibero, line 81:
```suggestion
            _logger.LogTrace("Check Boxset {CollectionName}", collection.Name);
```

Emby.Server.Implementations/ScheduledTasks/Tasks/CleanupCollectionPathsTask.cs
```diff
@@ -0,0 +1,117 @@
...
+            var itemsToRemove = new List<LinkedChild>();
+            foreach (var collectionLinkedChild in collection.LinkedChildren.ToArray())
+            {
+                if (!File.Exists(collectionLinkedChild.Path))
+                {
+                    _logger.LogInformation("Item in boxset {CollectionName} cannot be found at {ItemPath}.", collection.Name, collectionLinkedChild.Path);
```
crobibero, line 87:
```suggestion
                    _logger.LogInformation("Item in boxset {CollectionName} cannot be found at {ItemPath}", collection.Name, collectionLinkedChild.Path);
```

Emby.Server.Implementations/ScheduledTasks/Tasks/CleanupCollectionPathsTask.cs
```diff
@@ -0,0 +1,117 @@
...
+                    itemsToRemove.Add(collectionLinkedChild);
+                }
+            }
+
+            if (itemsToRemove.Any())
+            {
+                _logger.LogTrace("Update Boxset {CollectionName}.", collection.Name);
```
crobibero, line 94:
```suggestion
                _logger.LogTrace("Update Boxset {CollectionName}", collection.Name);
```

Emby.Server.Implementations/ScheduledTasks/Tasks/CleanupCollectionPathsTask.cs
```diff
@@ -0,0 +1,117 @@
...
+
+        var collections = collectionsFolder.Children.OfType<BoxSet>().ToArray();
+        _logger.LogTrace("Found {CollectionLength} Boxsets.", collections.Length);
+
+        for (var index = 0; index < collections.Length; index++)
+        {
+            var collection = collections[index];
+            _logger.LogTrace("Check Boxset {CollectionName}.", collection.Name);
+            var itemsToRemove = new List<LinkedChild>();
+            foreach (var collectionLinkedChild in collection.LinkedChildren.ToArray())
```
crobibero, line 83:
```suggestion
            foreach (var collectionLinkedChild in collection.LinkedChildren)
```

this is already an array

JPVenson: As this seems to be a shared object, i want to limit the possibility on a concurrent access to that collection. If you think this is unnecessary i will remove the enumeration.

BrianCArnold: Yes, we should remove calling .ToArray() as it won't do anything, except that it will copy the array items to a new array. `LinkedChildren` is reassigned to different arrays in other parts of the code, however which items are in the instance are never altered at any point. (i.e. we only ever do `LinkedChildren = ...`, not `LinkedChildren[3] = ...`).

Because an invocation of `foreach` like this:

```
foreach (var collectionLinkedChild in collection.LinkedChildren)
{
    // ...
}
```

Is, for all intents and purposes, just syntactic sugar for this:

```
var enumerator = collection.LinkedChildren.GetEnumerator();
while (enumerator.MoveNext())
{
    var collectionLinkedChild = enumerator.Current;   
    // ...
}
```

That means that even if the original array reference is assigned to a different array during execution, once the foreach loop starts executing, it will finish executing on the same array it started on, rather than switching to the new array partway through. 

Note: this is not true for `for`: 

```
for (int i = 0; i < collection.LinkedChildren; i++) 
{
    var collectionLinkedChild = collection.LinkedChildren[i];
    // ...
}
```

In this case, if the array was reassigned to a different array, it would actually switch to the new array reference, rather than complete execution on the original array, due to the fact that the array property is actually referenced on each iteration of the array. 

P.S. The compiler runs on eldritch horror and programmer tears, so it might decide to compile things differently due to deep magic if it's more efficient, but this is define behavior, so you can rely on it acting the same, even if the underlying execution is different.

BrianCArnold:
This issue doesn't just affect collections, is my understanding, but it also affects things like Playlists, as well as many other ways that ghost BaseItems show up in other parts of the system. Would it be possible to generalize this approach to clear out bad references in Playlists, etc?

Emby.Server.Implementations/ScheduledTasks/Tasks/CleanupCollectionPathsTask.cs
JPVenson: That is more to prevent modifications on that enumerator to cause this to throw
https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/List.cs#L1181

But i digress.

BrianCArnold:
If it's not an issue that's really affecting anyone, I can take a look at fixing it later. The only reason I want it is because having the Playlist etc. ghost items show up in the logs sets off my OCD, and that's not enough of a reason to ask someone else to make a PR.

Also, I know there's still some ghost items in my logs that aren't from a playlist, so I should probably dive in and find out where they're actually coming from before bothering to fix _just_ the playlist issue.

BrianCArnold:
But seriously, thanks for fixing this, it was causing a lot of people issues. 

Since the thing I'm complaining about doesn't affect users, I'll see if I can implement the fix as a plugin so I can stay on the mainline.

Shadowghost (approved):

1337joe (approved):

Emby.Server.Implementations/ScheduledTasks/Tasks/CleanupCollectionPathsTask.cs
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        var collections = collectionsFolder.Children.OfType<BoxSet>().ToArray();
+        _logger.LogDebug("Found {CollectionLength} Boxsets", collections.Length);
+
+        for (var index = 0; index < collections.Length; index++)
+        {
+            var collection = collections[index];
+            _logger.LogDebug("Check Boxset {CollectionName}", collection.Name);
+            var itemsToRemove = new List<LinkedChild>();
```
Bond-009, line 81: I'd rather reuse this list than allocating a new one for each collection, just clear it at the end of the for loop so it can be reused

Emby.Server.Implementations/ScheduledTasks/Tasks/CleanupCollectionPathsTask.cs
```diff
@@ -0,0 +1,116 @@
...
+            foreach (var collectionLinkedChild in collection.LinkedChildren)
+            {
+                if (!File.Exists(collectionLinkedChild.Path))
+                {
+                    _logger.LogInformation("Item in boxset {CollectionName} cannot be found at {ItemPath}", collection.Name, collectionLinkedChild.Path);
+                    itemsToRemove.Add(collectionLinkedChild);
+                }
+            }
+
+            if (itemsToRemove.Any())
```
Bond-009, line 91:
```suggestion
            if (itemsToRemove.Count != 0)
```

Bond-009 (approved):

PR #9242, opened 2023-02-03
Add uniqueid tag to nfo

fhriley:
This adds `<uniqueid>` to nfo files, which is required per Kodi's wiki, reference: [tv shows](https://kodi.wiki/view/NFO_files/TV_shows), [episodes](https://kodi.wiki/view/NFO_files/Episodes), [movies](https://kodi.wiki/view/NFO_files/Movies).

**Changes**

1) Adds `<uniqueid>` to nfo.
2) Adds jellyfin as a provider (and thus, a `uniqueid`) since it is a metadata provider. This change allows kodi plugins, sources, etc. to map kodi DB ids to jellyfin ids without having to do error prone title lookups.
3) Use DRY when adding provider ids.

### DIFF
--- a/MediaBrowser.Model/Entities/MetadataProvider.cs
+++ b/MediaBrowser.Model/Entities/MetadataProvider.cs
@@ -84,6 +84,11 @@ namespace MediaBrowser.Model.Entities
         /// <summary>
         /// The TvMaze provider.
         /// </summary>
-        TvMaze = 19
+        TvMaze = 19,
+
+        /// <summary>
+        /// The Jellyfin provider.
+        /// </summary>
+        Jellyfin = 20
     }
 }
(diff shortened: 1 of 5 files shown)


Shadowghost (dismissed):

MediaBrowser.Model/Entities/MetadataProvider.cs
```diff
@@ -84,6 +84,11 @@ public enum MetadataProvider
         /// <summary>
         /// The TvMaze provider.
         /// </summary>
-        TvMaze = 19
+        TvMaze = 19,
+
+        /// <summary>
+        /// The Jellyfin provider.
+        /// </summary>
+        Jellyfin = 20
```
cvium, line 92: Jellyfin is not a metadata provider

JPVenson:
closed due to inactivity

PR #9253, opened 2023-02-04
Return NotFound when itemId isn't found

Bond-009:


crobibero (approved):

Shadowghost (approved):

Jellyfin.Api/Models/UserDtos/ForgotPasswordDto.cs
```diff
@@ -11,5 +11,5 @@ public class ForgotPasswordDto
     /// Gets or sets the entered username to have its password reset.
     /// </summary>
     [Required]
-    public string? EnteredUsername { get; set; }
+    required public string EnteredUsername { get; set; }
```
crobibero, line 14: Pretty sure the order is supposed to be `public required string` but analyzer has a bug
https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3527

Jellyfin.Api/Auth/SyncPlayAccessPolicy/SyncPlayAccessHandler.cs
```diff
@@ -47,6 +48,10 @@ protected override Task HandleRequirementAsync(AuthorizationHandlerContext conte
 
             var userId = context.User.GetUserId();
             var user = _userManager.GetUserById(userId);
+            if (user is null)
```
cvium, line 51: If this happens, it's either an api key or the user was deleted between the authentication and authorization steps

Jellyfin.Api/Models/UserDtos/ForgotPasswordDto.cs
Bond-009: Thx I didn't know this, for now I'd just merge it as is (we already have code with this ordering) and fix it when the analyzer gets fixed.

Jellyfin.Api/Auth/SyncPlayAccessPolicy/SyncPlayAccessHandler.cs
Bond-009: What would you do here to prevent a nullref exception a bit lower?

Jellyfin.Api/Models/UserDtos/ForgotPasswordDto.cs
crobibero: I agree, just wanted to raise awareness

PR #9256, opened 2023-02-04
fix: swap to `jellyfin-ffmpeg5`

brettpetch:
**Changes**
- Swaps to building the Dockerfile with `jellyfin-ffmpeg5`

**Issues**

**Landmines**
Not sure about the state of arm support of JF-FFMPEG but might be worth a quick look if we update this on the amd64 build :)


Shadowghost (approved):

nyanmisaka (approved):
FYI the Dockerfile used in our releases is https://github.com/jellyfin/jellyfin-metapackages/blob/master/Dockerfile.amd64

PR #9264, opened 2023-02-06
Fix filter inaccurate TMDB search results

alamarre:
**Changes**
The TMDB API can yield undesirable results including the year filter being ignored, (e.g. Hercules (2014) yielding the 1997 result first) and partial matches coming before exact matches. (e.g. Cinderella (2008) yielding Cinderella Man (2005) as the first result)


Roardom:
Most people name their files after the year that imdb provides. If a film is released in a film festival the year prior to its theatrical release, imdb will use the year of the film festival, while tmdb will use the theatrical release year. Perhaps finding the result with the closest year (if it's within e.g. 3 years) instead of the exact year would work better for these cases.

MediaBrowser.Providers/Plugins/Tmdb/TmdbUtils.cs
```diff
@@ -188,5 +192,66 @@ public static string BuildParentalRating(string countryCode, string ratingValue)
...
+            var filtered = searchResults;
+
+            // partial matches can be returned before exact matches. Look for exact matches after removing invalid characters
+            var nearExactMatchFiltered = filtered.Where(s =>
+            {
+                var resultName = (s as SearchMovie)?.Title ?? (s as SearchTv)?.Name;
+                var resultOriginalName = (s as SearchMovie)?.OriginalTitle ?? (s as SearchTv)?.OriginalName;
+
+                return RemoveInvalidFileCharacters(resultName) == name
+                    || RemoveInvalidFileCharacters(resultOriginalName) == name;
```
crobibero, line 230:
```suggestion
                string? resultName;
                string? resultOriginalName;
                switch (s)
                {
                    case SearchMovie searchMovie:
                        resultName = searchMovie.Title;
                        resultOriginalName = searchMovie.OriginalTitle;
                        break;
                    case SearchTv searchTv:
                        resultName = searchTv.Name;
                        resultOriginalName = searchTv.OriginalName;
                        break;
                    default:
                        return false;
                }

                return string.Equals(RemoveInvalidFileCharacters(resultName), name, StringComparison.OrdinalIgnoreCase)
                       || string.Equals(RemoveInvalidFileCharacters(resultOriginalName), name, StringComparison.OrdinalIgnoreCase);
```

Longer code but more maintainable

MediaBrowser.Providers/Plugins/Tmdb/TmdbUtils.cs
```diff
@@ -188,5 +192,66 @@ public static string BuildParentalRating(string countryCode, string ratingValue)
...
+        }
+
+        /// <summary>
+        /// Identifies the nearest match to the search criteria. Addresses issues with search results from TMDB.
+        /// </summary>
+        /// <param name="name">The cleaned up name searched for.</param>
+        /// <param name="year">The year searched for.</param>
+        /// <param name="searchResults">The search results from a Movie or Tv search.</param>
+        /// <returns>The nearest match to the search criteria.</returns>
+        public static SearchMovieTvBase? GetBestMatch(string name, int year, IEnumerable<SearchMovieTvBase> searchResults)
```
crobibero, line 219:
```suggestion
        public static SearchMovieTvBase? GetBestMatch(string name, int year, IReadOnlyList<SearchMovieTvBase> searchResults)
```

using `IEnumerable` here raises warnings

Shadowghost (approved):

MediaBrowser.Providers/Plugins/Tmdb/TmdbUtils.cs
```diff
@@ -201,5 +205,79 @@ public static string BuildParentalRating(string countryCode, string ratingValue)
...
+                {
+                    DateTime? searchResultYear = (s as SearchMovie)?.ReleaseDate
+                        ?? (s as SearchTv)?.FirstAirDate;
+                    return searchResultYear.HasValue
+                        && Math.Abs(searchResultYear.Value.Year - year) < MaxYearDifference;
+                }).ToList();
+                if (yearFiltered.Any())
+                {
+                    filtered = yearFiltered.ToList();
+                }
```
crobibero, line 277: Can just return the first matching value instead of putting everything into a list

```suggestion
                var yearFiltered = searchResults
                    .Where(s =>
                    {
                        DateTime? searchResultYear = (s as SearchMovie)?.ReleaseDate
                                                     ?? (s as SearchTv)?.FirstAirDate;
                        return searchResultYear.HasValue
                               && Math.Abs(searchResultYear.Value.Year - year) < MaxYearDifference;
                    })
                    .FirstOrDefault();

                if (yearFiltered is not null)
                {
                    return yearFiltered;
                }
```

Shadowghost:
@alamarre would be nice if you could rebase and address the comments so we can get on with this

Roardom:
A new example that recently came up is Hamilton, originally widely released in 2020, but only just had a theatrical (re-)release this year in 2025. So, IMDB uses the year 2020 (https://www.imdb.com/title/tt8503618), and TMDB uses the year 2025 (https://www.themoviedb.org/movie/556574)

Bond-009:
Closing as stale, no response in 2+ years.

Issue #9230, opened 2023-02-01
[Issue]: Transcoding audio channels max should be 8, not 6


holow29:
### Please describe your bug

No matter what the transcoding profile on a device says for transcodingMaxAudioChannels or maxAudioChannels, the server will not allow more than 6. -ac 6 is always passed to ffmpeg even if higher channels are requested by client. Given the HLS spec and the fact that some devices support up to 8, this number should be bumped up or this restriction refactored.

As it is now, in [MediaBrowser.Controller](https://github.com/jellyfin/jellyfin/tree/master/MediaBrowser.Controller)/[MediaEncoding](https://github.com/jellyfin/jellyfin/tree/master/MediaBrowser.Controller/MediaEncoding)/[EncodingHelper.cs](https://github.com/jellyfin/jellyfin/blob/master/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs), GetNumAudioChannelsParam will not return higher than 6 because of transcoderChannelLimit which is set to 6 by default unless the codec is wma, mp3, or aac.

I think the server should only be restricting what is not possible for the server to do: in this case, I believe the limit would be based on HLS and ffmpeg. HLS allows for at least up to 8 channel audio. The max audio channels are already handled in multiple places client-side: in the transcoding profile, codec profile conditions, and potentially even HLS stream builder, though that might be the same as the transcoding profile.

We should increase the default transcoderChannelLimit to 8 or remove a default - the code already acts like there might not be a value for it even though it will always have a value of 6. We can add more specific codec cases where we know that ffmpeg will fail, as I think this feature was originally intentioned.

As an example: if a client is trying to get the server to transcode 8 channel audio to FLAC, which ffmpeg can do, the server will always pass -ac 6, limiting the channels to 6.
This is important for clients like those on Apple TV, which does not support passthrough audio but can decode up to 7.1 audio and ouput that as LPCM.

### Jellyfin Version

10.8.9

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #9266, opened 2023-02-06
Change transcoderChannelLimit default to 8

holow29:
**Changes**
Updates global transcoding max audio channels from 6 to 8.
Attempts to simplify some of the checks for max audio channels based on TranscodingMaxAudioChannels,resultChannels,transcoderChannelLimit.

If there are codecs that should be added that can't go up to 8 (from 6), we should probably add them individually. Alternatively, I suppose we could just add ones that go up to 8 individually, but that doesn't seem like the proper way to do this unless we got rid of the default altogether.

More discussion in the linked issue.

**Issues**
Fixes https://github.com/jellyfin/jellyfin/issues/9230


MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -2257,11 +2253,16 @@ public string GetAudioFilterParam(EncodingJobInfo state, EncodingOptions encodin
...
             {
                 resultChannels = resultChannels.HasValue
-                    ? Math.Min(resultChannels.Value, transcoderChannelLimit.Value)
-                    : transcoderChannelLimit.Value;
+                    ? Math.Min(resultChannels.Value, request.TranscodingMaxAudioChannels.HasValue
+                               ? Math.Min(request.TranscodingMaxAudioChannels.Value, transcoderChannelLimit.Value)
+                               : transcoderChannelLimit.Value)
+                    : request.TranscodingMaxAudioChannels.HasValue
+                        ? Math.Min(request.TranscodingMaxAudioChannels.Value, transcoderChannelLimit.Value)
+                        : transcoderChannelLimit.Value;
```
Bond-009, line 2265:
```cs
resultChannels = Math.Min(resultChannels.Value, transcoderChannelLimit.Value);
if (request.TranscodingMaxAudioChannels.HasValue && request.TranscodingMaxAudioChannels.Value < resultChannels)
{
    resultChannels = request.TranscodingMaxAudioChannels.Value;
}
```

Bond-009 (changes requested):

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
holow29: This does not take into account if resultChannels does not have a value, which the previous code did. Is this still something to worry about?

holow29: I've refactored this. Please take another look when you are able; if you'd prefer I use if statements instead of the ternary ones, I can switch, but I thought this made it a bit more readable because of the nullable types.

Bond-009 (approved):

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -2211,85 +2211,81 @@ public string GetAudioFilterParam(EncodingJobInfo state, EncodingOptions encodin
...
-
-            if (inputChannels.HasValue)
-            {
-                resultChannels = resultChannels.HasValue
-                    ? Math.Min(resultChannels.Value, inputChannels.Value)
-                    : inputChannels.Value;
-            }
+                // Set max transcoding channels to 8 to prevent encoding errors due to asking for too many channels
+                // AAC, FLAC, ALAC, libopus, libvorbis encoders all support at least 8 channels
+                int transcoderChannelLimit = 8;
```
nyanmisaka, line 2231: Please cap the max channel to 6 on `libfdk_aac`. Or it produces broken audio in HLS streaming.

nyanmisaka (changes requested):

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -2226,45 +2226,22 @@ public string GetAudioFilterParam(EncodingJobInfo state, EncodingOptions encodin
...
-                    // dca encoder currently only supports six channel output
-                    transcoderChannelLimit = 6;
-                }
-                else if (codec.IndexOf("mlp", StringComparison.OrdinalIgnoreCase) != -1)
-                {
-                    // mlp encoder currently only supports six channel output
-                    transcoderChannelLimit = 6;
-                }
-                else if (codec.IndexOf("truehd", StringComparison.OrdinalIgnoreCase) != -1)
+                switch (GetAudioEncoder(state))
```
Bond-009, line 2233: Don't use switch statement to compare strings, it doesn't behave the same as before

holow29: Is this still the case even though output of GetAudioEncoder() has ToLowerInvariant called if not otherwise defined?

Bond-009: I would probably work, but just to be consistent with the rest of the codebase I'd use the string comparison functions

holow29: I've switched to using string comparison. Let me test to make sure this is working. Is the switch expression I have now okay or would it be better to make into if statements instead?

holow29: I've changed to string comparison functions in a switch expression. If it would be better to do if statements instead, let me know.

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
holow29: I've switched to querying for the encoder vs. the codec so we can be more precise. Added libfdk_aac and a few others.

nyanmisaka (approved):

Shadowghost (approved):

PR #9278, opened 2023-02-08
Move Jellyfin.Api into src dir

Bond-009:
Be consistent with path separator in csproj files


crobibero (approved):

Shadowghost (approved):
But it does break all other API PRs...

Bond-009:
Closing as stale

PR #9282, opened 2023-02-08
refactor: simplify authz

cvium:
**Changes**
- Make DefaultAuthorizationRequirement a superclass to simplify calls in inherited requirements
- Simplify some handlers and requirements

**Issues**


Jellyfin.Api/Auth/DefaultAuthorizationPolicy/DefaultAuthorizationRequirement.cs
```diff
@@ -7,5 +7,18 @@ namespace Jellyfin.Api.Auth.DefaultAuthorizationPolicy
...
+        /// <param name="validateParentalSchedule">A value indicating whether to validate parental schedule.</param>
+        public DefaultAuthorizationRequirement(bool validateParentalSchedule = true)
+        {
+            ValidateParentalSchedule = validateParentalSchedule;
+        }
+
+        /// <summary>
+        /// Gets a value indicating whether to ignore parental schedule.
+        /// </summary>
+        public bool ValidateParentalSchedule { get; init; }
```
barronpm, line 22: If you're setting this in the constructor, is there any reason to also include the `init`?

cvium: No, it was from a previous iteration

Jellyfin.Data/DayOfWeekHelper.cs
```diff
@@ -17,5 +17,16 @@ public static DayOfWeek[] GetDaysOfWeek(DynamicDayOfWeek day)
                 _ => new[] { (DayOfWeek)day }
             };
         }
+
+        public static bool Contains(this DynamicDayOfWeek dynamicDayOfWeek, DayOfWeek dayOfWeek)
+        {
+            return dynamicDayOfWeek switch
+            {
+                DynamicDayOfWeek.Everyday => true,
+                DynamicDayOfWeek.Weekday => dayOfWeek is > DayOfWeek.Sunday and <= DayOfWeek.Friday,
```
cvium, line 26:
```suggestion
                DynamicDayOfWeek.Weekday => dayOfWeek is >= DayOfWeek.Monday and <= DayOfWeek.Friday,
```

Shadowghost (commented):
Functionality wise I verified this use cases:
* Parental control (working)
* Downloading (working)
* Elevation (working)

One thing I noticed though: When reloading the page I always need to log in again.

nielsvanvelzen (dismissed):
The OpenAPI specification is missing a lot (but not all) 401/403 responses with this PR. They need to be added back.

crobibero (approved):

Shadowghost (changes requested):
When reloading the page as a non-admin user I always need to log in again.

Shadowghost (approved):

crobibero (approved):

PR #9291, opened 2023-02-10
fix: skip library folders that are inaccessible or empty

cvium:
**Changes**
- Skip `Folder` where `IsTopParent == true` if the filesystem path is inaccessible or empty
- Don't remove anything when `IsRoot` is true (AggregateFolder and UserRootFolder will have this condition and they directly point to library folders)

I have only tested it on Windows with SMB mount and an empty folder. It needs to be tested more, especially Linux network shares, symlinks etc.

**Issues**

### DIFF
--- a/MediaBrowser.Controller/Entities/Folder.cs
+++ b/MediaBrowser.Controller/Entities/Folder.cs
@@ -440,7 +440,6 @@ namespace MediaBrowser.Controller.Entities
                     }
                 }
 
-                // If any items were added or removed....
                 if (newItems.Count > 0)
                 {
                     LibraryManager.CreateItems(newItems, this, cancellationToken);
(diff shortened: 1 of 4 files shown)


MediaBrowser.Controller/Entities/Folder.cs
```diff
@@ -416,7 +438,11 @@ private async Task ValidateChildrenInternal2(IProgress<double> progress, bool re
                             LibraryManager.DeleteItem(item, new DeleteOptions { DeleteFileLocation = false }, this, false);
                         }
                     }
+                }
 
+                // If any items were added or removed....
```
cvium, line 443:
```suggestion
```

JPVenson (approved):

ShrirajHegde:
Any update on this? This is still a very bad problem.

Especially since rescan takes like days for a library of few TBs.

crobibero (approved):

PR #9304, opened 2023-02-12
Always run code analyzers for tests projects

Bond-009:


crobibero (commented):
Can this just be moved to Dorectory.Build.props in the tests folder?

Shadowghost (approved):

crobibero (approved):

PR #9305, opened 2023-02-12
Use mediator instead of EventManager

barronpm:
**Changes**
Uses a custom mediator implementation instead of our `EventManager`.


Bond-009:
@barronpm What's the status of this?

barronpm:
Only one event has been moved over to the new system, but it's ready for a review on the concept. If we like the way this is set up, I can finish updating the events. Pinging @cvium since I was working with him on the initial draft.

Jellyfin.Server.Implementations/Mediator/JellyfinMediator.cs
```diff
@@ -0,0 +1,66 @@
...
+    {
+        _serviceProvider = serviceProvider;
+        _logger = logger;
+    }
+
+    /// <inheritdoc />
+    public async ValueTask Publish<TNotification>(TNotification notification, CancellationToken cancellationToken = default)
+        where TNotification : INotification
+    {
+        var handlers = _serviceProvider.GetServices<INotificationHandler<TNotification>>();
```
JPVenson, line 31: Generally i think it might be a good idea to allow Transient notification handlers as well to keep code separated. this would require the Mediator to create a scope.

cvium: By Transient I assume you mean handlers with Transient Lifetime? I don't see how that would aid in keeping code separate.

JPVenson: > By Transient I assume you mean handlers with Transient Lifetime?

Yes i thought of just services to handle stuff. Simply to follow the SOC principle.

For example i think of a service that handles just the event and then calls plugin specific code. For plugins that will handle multiple events, this would help to avoid registering 20ish interfaces on a single class.

cvium: Transient just means a new instance is created when the class/interface is pulled from the DI container. You can still fail to have proper separation of concerns. The general rule of thumb for the mediator handlers is that they should be stateless and one of the reasons for singleton handlers is to avoid unnecessary object creation.

JPVenson: Transient because they should not serve any other purpose. And that would play nicely with the stateless approach, because you know they do get recreated each time.

cvium: On a conceptual level, sure, but it doesn't stop someone from trying to shove in state into a transient service (not to mention, any of its deps can have state). You can also create one giant handler that subscribes to everything (bad).

In practice, we have a _lot_ of _frequent_ events, which makes transient a poor choice for older/smaller devices. Sometimes, compromises must be made.

JPVenson: > but it doesn't stop someone from trying to shove in state into a transient service

Well yes but i expect devs doing that having a whole other set of issues. :-D

> You can also create one giant handler that subscribes to everything (bad)

This is exactly the pattern i want to avoid.

> In practice, we have a lot of frequent events, which makes transient a poor choice for older/smaller devices. Sometimes, compromises must be made.

Valid point. I more thought of the "best case" scenario :-)

PR #9309, opened 2023-02-13
Add dependabot.yml

zcsizmadia:
**Changes**
Add dependabot.yml to let github to check and create PRs weekly if any nuget packages or github actions got updates.
See https://docs.github.com/en/code-security/dependabot


crobibero:
We use Renovate for dependency management (https://github.com/jellyfin/jellyfin/blob/master/.github/renovate.json)

Issue #8991, opened 2022-12-31
[Issue]: Unhandled WebSocket System.ObjectDisposedException


ipitio:
### Please describe your bug

After updating to v10.8.8, Jellyfin has been crashing once or twice a day due to an uncaught Exception. The logs indicate that the fatal error is thrown when a client improperly closes a connection. The connection is then closed on the server after it times out; yet, the server still tries to write to the stream.

### Jellyfin Version

Other

### if other:

10.8.8

### Environment

```markdown
- OS: Windows 10
- Virtualization: N/A
- Clients: Web (browser, mobile), Android TV, Roku, webOS
- Browser: Firefox latest
- FFmpeg Version: the one that was shipped with 10.8.8
- Playback Method: All
- Hardware Acceleration: NVENC
- Plugins: AniDB7.0.0.0,AniList5.0.0.0,AniSearch4.0.0.0,AudioDB10.8.8.0,Bookshelf9.0.0.0,Cover Art Archive7.0.0.0,Email Notifications9.0.0.0,Fanart10.0.0.0,IMVDb3.0.0.0,InfuseSync1.4.2.0,Intro Skipper0.1.7.0,Jellyscrub1.0.0.7,Kitsu4.0.0.0,Kodi Sync Queue9.0.0.0,MusicBrainz10.8.8.0,OMDb10.8.8.0,OPDS Feed4.0.0.0,Open Subtitles18.0.0.0,Playback Reporting13.0.0.0,Reports15.0.0.0,Studio Images10.8.8.0,TMDb10.8.8.0,TMDb Box Sets10.0.0.0,TMDb Trailers10.0.0.0,TVmaze10.0.0.0,TheTVDB8.0.0.0,Trakt22.0.0.0,VGMdb2.0.0.0,КиноПоиск10.7.5.4
- Reverse Proxy: Apache for internet clients, none for local/vpn
- Base URL: none
- Networking: Apache runs in WSL (Jellyfin doesn't), VPN is on the router (split tunnel)
- Storage: local
```

### Jellyfin logs

```shell
[WRN] [72] Emby.Server.Implementations.HttpServer.WebSocketConnection: WS "{CLIENT_IP_ADDR}" error receiving data: "The remote party closed the WebSocket connection without completing the close handshake."
```
```
[INF] [72] Emby.Server.Implementations.HttpServer.WebSocketManager: WS "{CLIENT_IP_ADDR}" closed
```
```
[...]

PR #9326, opened 2023-02-16
Catch Exception when disposing connection

ipitio:
**Changes**
Catch Exception when disposing connection.

**Issues**
jellyfin/jellyfin#8991

### DIFF
--- a/MediaBrowser.Controller/Net/BasePeriodicWebSocketListener.cs
+++ b/MediaBrowser.Controller/Net/BasePeriodicWebSocketListener.cs
@@ -232,6 +232,11 @@ namespace MediaBrowser.Controller.Net
                 // TODO Investigate and properly fix.
                 Logger.LogError(ex, "Object Disposed");
             }
+            catch (Exception ex)
+            {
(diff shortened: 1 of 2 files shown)


MediaBrowser.Controller/Net/BasePeriodicWebSocketListener.cs
```diff
@@ -232,6 +232,11 @@ private void DisposeConnection(Tuple<IWebSocketConnection, CancellationTokenSour
                 // TODO Investigate and properly fix.
                 Logger.LogError(ex, "Object Disposed");
             }
+            catch (Exception ex)
+            {
+                // TODO Investigate and properly fix.
+                Logger.LogError(ex, "Object Disposed Exception");
```
crobibero, line 238:
```suggestion
                Logger.LogError(ex, "Error disposing websocket");
```

We don't know if the exception is because the object was disposed or some other reason

crobibero (approved):

1337joe (approved):

Bond-009 (approved):

Issue #8971, opened 2022-12-28
Add "speed" property to playback state API's


nielsvanvelzen:
Currently, the remote control API does not support a "speed" property. We should add it to the API so we can:

1) interpolate the current progress without sending a playback update 10 times per second
2) view the playback speed in dashboard/remote control
3) allow for changing the speed using remote control

The property should be a float/double with a default value of 1.0. The server should clamp the value to avoid bad input (something like 0.1-10.0?).

From my understanding the new property should be added to (at least) PlaybackProgressInfo, PlaybackStartInfo and PlayerStateInfo. And likely more.

crobibero:
What happens in the kotlin sdk if we add a new, non-nullable propety to an object? I thought doing so broke existing versions

rzk3:
Hi! I'd like to take on this issue!

Bogdanf21:
Hi, is anyone currently working on this?

pmnlla:
Waiting to merge #9329 iirc

singlatushar07:
Hi, is this still being worked on? If not, I would like to work on this.

LezardSC:
Hi, I feel like no one is working on it anymore. If no one answered that they're working on it at December 1st I'll start working on it.


EDIT: I noticed that PR https://github.com/jellyfin/jellyfin/pull/9329 implements most of this but isn't merged.
I don't know why but I will not work on it since the work is already done.

Eyosias6981:
Hi maintainers! I'm a backend software engineer working on .NET systems and I would love to tackle this issue. I've tracked down the target model locations for PlaybackProgressInfo, PlaybackStartInfo, and PlayerStateInfo and am ready to implement the speed properties. Could you please assign this task to me?

PR #9329, opened 2023-02-16
Add PlaybackSpeed properties to state APIS

rzk3:
**Changes**
- Added Playback speed property to PlaybackProgressInfo and PlayerStateInfo
- Added checks to make sure PlaybackSpeed stays within 0.1 and 10.0

**Issues**
Adds requested properties from issue #8971


crobibero:
`Math.Clamp` looks like it would be useful for this :)

I think there are additional classes where this property should be added, such as in SyncPlay

rzk3:
Thanks for the advice on `Math.Clamp`, it's much more elegant than what I had before!

I added `PlaybackSpeed` to a few more classes relating to SyncPlay. I'm honestly not too sure where else it would be needed, I thought about adding it to the encoding related classes, but I'm not sure if that is better controlled by the `FrameRate` property they already have.
I would appreciate it if you could point me towards some more places where it would be useful!

Emby.Server.Implementations/SyncPlay/Group.cs
```diff
@@ -137,6 +142,30 @@ public class Group : IGroupStateContext
...
+                if (value is null)
+                {
+                    _playbackSpeed = null;
+                }
+                else
+                {
+                    _playbackSpeed = Math.Clamp((double)value, 0.1, 10.0);
+                }
+            }
+        }
```
crobibero, line 167:
```suggestion
        public double? PlaybackSpeed
        {
            get => _playbackSpeed;
            set => _playbackSpeed = value is null ? null : Math.Clamp(value.Value, 0.1, 10.0);
        }
```

daullmer (approved):

crobibero (approved):

Bond-009:
Maybe a nitpick, but I don't think the precision of a double is needed, a float should be fine, no?

rzk3:
I had chosen to go with doubles because I was concerned about the occasional float inaccuracy causing some desync during SyncPlay. Assuming that the range goes 0.1x to 10x I don't think it's too far-fetched to assume that it could result in a few second desync (I'm doing this math in my head so take it with a grain of salt). But if you are rounding or going with set values (e.x. 1x, 1.5x...) then going with floats would most likely be a better choice.

If you'd like me to change it please tell me and I'll do it as soon as I have time!

crobibero:
I think in addition to changing to use a float we should also round to the nearest tenths 

eg [...], 0.9, 1.0, 1.1, 1.2, 1.3, [...]

rzk3:
I made the changes to `float?` and to round the value to the nearest tenth. The code feels a little bit messy `set => _playbackSpeed = value is null ? null : (float?)Math.Round(Math.Clamp(value.Value, 0.1f, 10.0f), 1);` Because C#'s built-in rounding appears to only have support for double values. If you know of a better way to do this without needing to convert from float -> double -> float please tell me!

jbatt33:
What does it mean for the playback value to be null?

jbatt33:
> What does it mean for the playbackspeed to be null?

I see the commit now, 

[Made PlaybackSpeed nullable for kotlin SDK](https://github.com/jellyfin/jellyfin/pull/9329/commits/bb840a2e70262108bf7d2954cdb679eab394c63a)

@[rzk3](https://github.com/rzk3/jellyfin/commits?author=rzk3)
rzk3 committed on Feb 16

Thanks.

jbatt33:
> I made the changes to `float?` and to round the value to the nearest tenth. The code feels a little bit messy `set => _playbackSpeed = value is null ? null : (float?)Math.Round(Math.Clamp(value.Value, 0.1f, 10.0f), 1);` Because C#'s built-in rounding appears to only have support for double values. If you know of a better way to do this without needing to convert from float -> double -> float please tell me!

Did you consider using the decimal type?

Shadowghost:
Can you please resolve the conflicts, so this can be looked at further?

MediaBrowser.Model/SyncPlay/PlayQueueUpdateReason.cs
```diff
@@ -53,6 +53,11 @@ public enum PlayQueueUpdateReason
         /// <summary>
         /// A user is changing shuffle mode.
         /// </summary>
-        ShuffleMode = 9
+        ShuffleMode = 9,
+
+        /// <summary>
+        /// A user has changed the playback speed.
+        /// </summary>
+        PlaybackSpeedChange = 10
```
nielsvanvelzen, line 61: Changing the playback speed shouldn't update the queue so this feels unnecessary.

cvium: It affects Syncplay though

nielsvanvelzen: From my understanding this enum is only for when the actual queue changes (item added/removed/moved etc.) not the speed of a single item. That's just the state like play/pause/volume/track etc.

JPVenson (approved):

LezardSC:
Why isn't this PR merged?

Gauvino:
Does this pr need more review or anything else to be merged ?

jbatt33:
> Does this pr need more review or anything else to be merged ?

I think it just needs to fresh review.

nielsvanvelzen (changes requested):

Issue #4991, opened 2021-01-09
Jellyfin doesn't read track numbers from Vinyl records correctly e.g. A1, A2 & B1 etc


CluelessTechnologist:
**Description of issue**

So the usual way of numbering tracks on digital music is obviously:

> 01 - Song 1, 02 - Song 2, 03 - Song 3  

or
> 1 - Song 1, 2 - Song 2, 3 - Song 3  

But I sometimes listen to Vinyl records that have been digitized and they sometimes use track number like these:  

> A1 - Song 1, A2 - Song 2, B1 - Song 3

Now this is no problem in most media players, for example Plex has the tracks in the correct order. But in Jellyfin the order is by song name instead. My guess is that jellyfin don't recongize the track numbers correctly and therefore falls back to sorting by song name instead. 

**Media Info of the file**
```
General
Complete name                            : /The Vels/01. Danger Zone.flac
Format                                   : FLAC
Format/Info                              : Free Lossless Audio Codec
File size                                : 68.4 MiB
Duration                                 : 3 min 16 s
Overall bit rate mode                    : Variable
Overall bit rate                         : 2 923 kb/s
Album                                    : House Of Miracles
Track name                               : Danger Zone
Track name/Position                      : A1
Performer                                : The Vels
Producer                                 : Mercury, Mercury
Genre                                    : Rock, Pop
Recorded date                            : 1986
Cover                                    : Yes
Cover type                               : Cover (front)
Cover MIME                               : image/jpeg
CATALOGID                                : 826 804-1 M-1, 422-826 804-1M-1
COUNTRY                                  : US
DISCOGS_RELEASE_ID                       : 2190207
MEDIATYPE                                : Vinyl (LP, Album)
STYLE                                    : New Wave
WWW                                      : https://www.discogs.com/The-Vels-House-Of-Miracles/release/2190207

Audio
Format                                   : FLAC
Format/Info                              : Free Lossless Audio Codec
Duration                                 : 3 min 16 s
Bit rate mode                            : Variable
Bit rate                                 : 2 891 kb/s
Channel(s)                               : 2 channels
Channel layout                           : L R
Sampling rate                            : 96.0 kHz
Bit depth                                : 24 bits
Compression mode                         : Lossless
Stream size                              : 67.6 MiB (99%)
Writing library                          : libFLAC 1.3.0 (UTC 2013-05-26)

```

**Logs**
N/A

**FFmpeg Logs**
N/A

[...]

PR #7672, opened 2022-04-27
Allow pulling missing Track Number or other data from MusicBrainz

scottpk:
Add option to pull in missing track info from MusicBrainz

**Changes**
Add code that pulls info from MusicBrainz when missing
Add a configuration option to enable/disable this behavior

**Issues**
Relates to #4991 Jellyfin doesn't read track numbers from Vinyl records correctly e.g. A1, A2 & B1 etc (user's vinyl software did not fill in the track number)

PR #7603, opened 2022-04-15
Migrate MusicBrainz plugin to MetaBrainz.MusicBrainz

Shadowghost:
Rebased the original commits onto master, updated the library and adapted the code where necessary.

Supersedes #6369

PR #9331, opened 2023-02-16
Get track number/length from MusicBrainz where null

scottpk:
Add ability to get track number and length from MusicBrainz for tracks where this is null

**Changes**

Adds the ability to fill track number and length from MusicBrainz for tracks which have this data as null. This looks for null, so that existing metadata will not be overwritten.

image

Note that since this is a numeric field, it will be the numeric ordering (as in, for records B-side tracks will be after A-side tracks).

**Issues**
Relates to #4991 Jellyfin doesn't read track numbers from Vinyl records correctly e.g. A1, A2 & B1 etc (user's vinyl software did not fill in the track number)


CluelessTechnologist:
Please see discussion in previous PR: https://github.com/jellyfin/jellyfin/pull/7672

So since https://github.com/jellyfin/jellyfin/pull/7603 has been merged I see no blockers to have someone look over this PR? Or are there more rewrites planned to MusicBrainz plugin? @Shadowghost

Shadowghost:
Not from my side. This should be good to go if it works as intended

scottpk:
Apologies, I am less familiar with git still. How do I request a review?

CluelessTechnologist:
Any update on the review? @dmitrylyzo @thornbill 

I understand there might be other more pressing PRs to review. But maybe it's possible to take a quick look and let us know if there's any major issues we need to address? Thanks!

CluelessTechnologist:
I see. We need a backend maintainer to review this. Maybe @cvium or @crobibero since they reviewed https://github.com/jellyfin/jellyfin/pull/7603?

Shadowghost (approved):

MediaBrowser.Providers/Plugins/MusicBrainz/MusicBrainzAlbumProvider.cs
```diff
@@ -256,6 +256,71 @@ public async Task<MetadataResult<MusicAlbum>> GetMetadata(AlbumInfo info, Cancel
             }
         }
 
+        if (Plugin.Instance != null && Plugin.Instance.Configuration.GetMissingTrackInfo)
+        {
+            if (result.Item.Children.Any(s => s is Audio && s.IndexNumber == null))
+            {
+                var rspobj = _musicBrainzQuery.LookupRelease(Guid.Parse(releaseId ?? string.Empty), Include.Recordings);
```
cvium, line 263: String.empty doesn't make much sense here.

scottpk: `String.Empty` is there because the project's StyleCop settings are causing this:

image

That's the only reason I've done this, it prevents the build otherwise

cvium: It doesn't make sense to look up an empty id. Check that it isn't null before you try to parse it

MediaBrowser.Providers/Plugins/MusicBrainz/MusicBrainzAlbumProvider.cs
```diff
@@ -256,6 +256,71 @@ public async Task<MetadataResult<MusicAlbum>> GetMetadata(AlbumInfo info, Cancel
             }
         }
 
+        if (Plugin.Instance != null && Plugin.Instance.Configuration.GetMissingTrackInfo)
```
cvium, line 259: This entire section needs some refactoring. The nested indentation levels are too deep and it performs N + N + 3 database lookups. `result.Item.Children` always hits the db.

cvium, line 259: This entire section needs some refactoring. The nested indentation levels are too deep and it performs N + N + 3 database lookups. `result.Item.Children` always hits the db. The last two foreach could also be combined into one.

cvium (changes requested):

MediaBrowser.Providers/Plugins/MusicBrainz/MusicBrainzAlbumProvider.cs
```diff
@@ -256,6 +256,54 @@ public async Task<MetadataResult<MusicAlbum>> GetMetadata(AlbumInfo info, Cancel
...
+                    {
+                        c.IndexNumber = trk?.Position;
+                    }
+
+                    if ((ticks != c.RunTimeTicks) || (index != c.IndexNumber))
+                    {
+                        if (children.Where(p => Guid.Equals(p.Id, c.ParentId)).FirstOrDefault() is MusicAlbum parent)
+                        {
+                            parent.AddChild(c);
+                            result.Item.AddChild(parent);
```
cvium, line 300: Why is a musicalbum being set as the parent of another musicalbum?

MediaBrowser.Providers/Plugins/MusicBrainz/MusicBrainzAlbumProvider.cs
```diff
@@ -256,6 +256,54 @@ public async Task<MetadataResult<MusicAlbum>> GetMetadata(AlbumInfo info, Cancel
...
+                    }
+
+                    if (c.IndexNumber == null)
+                    {
+                        c.IndexNumber = trk?.Position;
+                    }
+
+                    if ((ticks != c.RunTimeTicks) || (index != c.IndexNumber))
+                    {
+                        if (children.Where(p => Guid.Equals(p.Id, c.ParentId)).FirstOrDefault() is MusicAlbum parent)
```
cvium, line 297: Can this ever happen. A MusicAlbum is not Audio, so there should be no musicalbums in children

MediaBrowser.Providers/Plugins/MusicBrainz/MusicBrainzAlbumProvider.cs
```diff
@@ -256,6 +256,54 @@ public async Task<MetadataResult<MusicAlbum>> GetMetadata(AlbumInfo info, Cancel
...
+        if (Plugin.Instance != null && Plugin.Instance.Configuration.GetMissingTrackInfo)
+        {
+            var children = result.Item.Children.ToArray();
+            if (children.Any(s => s is Audio && s.IndexNumber == null) && releaseId != null)
+            {
+                var rspobj = _musicBrainzQuery.LookupRelease(Guid.Parse(releaseId), Include.Recordings);
+
+                IList<ITrack> tracks = new List<ITrack>();
+
+                if (rspobj.Media != null)
```
cvium, line 268: If it's null, is there any reason to continue?

MediaBrowser.Providers/Plugins/MusicBrainz/MusicBrainzAlbumProvider.cs
scottpk: Yes, because this loop is filtered by instances where `RunTimeTicks` or `IndexNumber` is `null` meaning that this metadata is missing.

JPVenson (approved):

Bond-009:
Closing as stale, review unanswered for 2+ years

PR #9351, opened 2023-02-18
Fix EqualsAny condition check for int and double

### LINKED ISSUE #9350: [Issue]: Unable to use "EqualsAny" condition for property that has numeric values - when submitting to /Sessions/Capabilities
### Please describe your bug

When submitting device capabilities, I need to pass 2 values for AudioSampleRate. I expected to use EqualsAny, but it fails.

This works:
`{
                "Type": "Audio",
                "Codec": "flac",
                "Conditions": [
                    {
                        "Condition": "EqualsAny",
                        "Property": "AudioSampleRate",
                        "Value": "44100",
                        "IsRequired": true
                    }
                ]
            }`

This fails
`{
                "Type": "Audio",
                "Codec": "flac",
                "Conditions": [
                    {
                        "Condition": "EqualsAny",
                        "Property": "AudioSampleRate",
                        "Value": "44100|48000",
                        "IsRequired": true
                    }
                ]
            }`

Looking at the condition code here: https://github.com/jellyfin/jellyfin/blob/3e2fbb2685ab41827e6a34dd85162b2fe9bd5b67/MediaBrowser.Model/Dlna/ConditionProcessor.cs it looks like if the item value is numeric, EqualsAny doesn't work.

Look at line 143, if the item value is an int, equals and equalsany does the same thing. I expected it to split the expected value at the pipe delimiter as seen on line 173.

### Jellyfin Version

10.8.0

### if other:

10.8.9

### Environment

```markdown
- OS: Windows
- Virtualization: Direct Install
- Clients: Roku & Web
- Browser: Waterfox
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

image

### Code of Conduct

Shadowghost:
Currently `EqualsAny` is not implemented for `double` and the implementation for `int` does not work as it should (see #9350)

**Changes**
Split the conditional string value, parse the substrings into a int/double and check if one of them is equal to the current value.

**Issues**
Fixes #9350

### DIFF
--- a/MediaBrowser.Model/Dlna/ConditionProcessor.cs
+++ b/MediaBrowser.Model/Dlna/ConditionProcessor.cs
@@ -136,12 +136,26 @@ namespace MediaBrowser.Model.Dlna
                 return !condition.IsRequired;
             }
 
+            var conditionType = condition.Condition;
+            if (condition.Condition == ProfileConditionType.EqualsAny)
+            {
+                foreach (var singleConditionString in condition.Value.AsSpan().Split('|'))
+                {
+                    if (int.TryParse(singleConditionString, NumberStyles.Any, CultureInfo.InvariantCulture, out int conditionValue)
+                        && conditionValue.Equals(currentValue))
+                    {
+                        return true;
+                    }
+                }
+
+                return false;
+            }
+
             if (int.TryParse(condition.Value, NumberStyles.Any, CultureInfo.InvariantCulture, out var expected))
             {
-                switch (condition.Condition)
+                switch (conditionType)
                 {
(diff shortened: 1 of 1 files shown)


Bond-009 (changes requested):
Use a foreach and our own Split that works on ReadonlySpan<char>

Bond-009 (changes requested):
Maybe change the other ones in this file too, the default `NumberStyles` should be fine, no need to support currency symbols or hex etc

Bond-009 (approved):

crobibero (approved):

PR #6168, opened 2021-06-09
Reduce allocations, simplifed code, faster implementation, included tests - StreamInfo.ToUrl

BaronGreenback:
Method is used extensively in the device profile sub-system.

**Original code (ToUrl / BuildParam methods) :-**

Added values to a name/pair list, creating a new object for each entry. (checking and storing string.Empty for null)
parsed this list and ignored any string.Empty or default values and built a list<string> instance of individual name=value
joined the list<string> together with &
appending it at the end of the url
returned the url.

**Replacement code**
creates a string builder 2000 bytes long.
adds the url to string builder
adds the non null/default values to the string builder
returns the url

As it's a basic function re-write, included a testing unit to ensure the output remains the same.

PR #9369, opened 2023-02-21
Reduce allocations, simplifed code, faster implementation, included tests - StreamInfo.ToUrl

Shadowghost:
Rebases the code on current master and adopts it to changes happening since the original PR.

Supersedes #6168

### DIFF
--- a/MediaBrowser.Model/Dlna/StreamInfo.cs
+++ b/MediaBrowser.Model/Dlna/StreamInfo.cs
@@ -1,7 +1,12 @@
+#pragma warning disable CA1819 // Properties should not return arrays
+
 using System;
 using System.Collections.Generic;
 using System.Globalization;
+using System.Linq;
+using System.Text;
 using Jellyfin.Data.Enums;
+using Jellyfin.Extensions;
 using MediaBrowser.Model.Drawing;
 using MediaBrowser.Model.Dto;
 using MediaBrowser.Model.Entities;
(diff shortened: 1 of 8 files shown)


Bond-009 (changes requested):
The property based testing should be using FsCheck

Shadowghost:
Do you have an example? I never used FsCheck.

MediaBrowser.Model/Dlna/StreamInfo.cs
```diff
@@ -618,201 +620,296 @@ public string GetOption(string name)
...
 
-            if (!item.IsDirectStream)
+            if (!IsDirectStream)
             {
-                if (item.RequireNonAnamorphic)
+                if (RequireNonAnamorphic)
                 {
-                    list.Add(new NameValuePair("RequireNonAnamorphic", item.RequireNonAnamorphic.ToString(CultureInfo.InvariantCulture).ToLowerInvariant()));
+                    sb.Append("&RequireNonAnamorphic=");
+                    sb.Append(RequireNonAnamorphic.ToString(CultureInfo.InvariantCulture).ToLowerInvariant());
```
crobibero, line 809:
```suggestion
                    sb.Append(RequireNonAnamorphic.ToString(CultureInfo.InvariantCulture));
```

The `ToLowerInvariant` isn't required, the api is case-insensitive

MediaBrowser.Model/Dlna/StreamInfo.cs
```diff
@@ -618,201 +620,296 @@ public string GetOption(string name)
...
+                sb.Append('=');
+                sb.Append(pair.Value.Replace(" ", string.Empty, StringComparison.Ordinal));
             }
 
-            if (!item.IsDirectStream)
+            if (!IsDirectStream)
             {
-                list.Add(new NameValuePair("TranscodeReasons", item.TranscodeReasons.ToString()));
+                sb.Append("&TranscodeReasons=");
+                sb.AppendJoin(',', TranscodeReasons.ToString());
```
crobibero, line 882: `.ToString()` on the flags enum returns something like `"AudioIsExternal, DirectPlayError"` so this doesn't work as intended

MediaBrowser.Model/Dlna/StreamInfo.cs
```diff
@@ -618,201 +620,296 @@ public string GetOption(string name)
...
+
+            return sb.ToString();
+        }
+
+        public List<SubtitleStreamInfo> GetExternalSubtitles(ITranscoderSupport transcoderSupport, bool includeSelectedTrackOnly, string baseUrl, string accessToken)
+        {
+            return GetExternalSubtitles(transcoderSupport, includeSelectedTrackOnly, false, baseUrl, accessToken);
+        }
+
+        public List<SubtitleStreamInfo> GetExternalSubtitles(ITranscoderSupport transcoderSupport, bool includeSelectedTrackOnly, bool enableAllProfiles, string baseUrl, string accessToken)
```
crobibero, line 898:
```suggestion
        private List<SubtitleStreamInfo> GetExternalSubtitles(ITranscoderSupport transcoderSupport, bool includeSelectedTrackOnly, bool enableAllProfiles, string baseUrl, string accessToken)
```

tests/Jellyfin.Model.Tests/Dlna/StreamInfoTests.cs
```diff
@@ -0,0 +1,236 @@
...
+        return arr.ToArray();
+    }
+
+    /// <summary>
+    /// Fills most properties with random data.
+    /// </summary>
+    /// <param name="destination">The instance to fill with data.</param>
+    private static void FillAllProperties<T>(T destination)
+    {
+        var random = new Random();
```
crobibero, line 139: Randomness in a test isn't great.. should at least specify a seed so failures can be reproduced.

Shadowghost: I just copied them from the original PR. Still need to understand how they work...

src/Jellyfin.Extensions/EnumerableExtensions.cs
```diff
@@ -55,4 +57,84 @@ public static IEnumerable<T> SingleItemAsEnumerable<T>(this T item)
...
+    /// Gets all individual flags from an enum.
+    /// </summary>
+    /// <param name="value">The enum.</param>
+    /// <returns>The IEnumerable{Enum} containeing all individual flags.</returns>
+    public static IEnumerable<Enum> GetIndividualFlags(this Enum value)
+    {
+        return GetFlags(value, GetFlagValues(value.GetType()).ToArray());
+    }
+
+    private static IEnumerable<Enum> GetFlags(Enum value, Enum[] values)
```
crobibero, line 81: Just use the `.HasFlag()` function. Additionally, the result from `Enum.GetValues` should be cached.

Sample: https://github.com/jellyfin/jellyfin/blob/master/src/Jellyfin.Extensions/Json/Converters/JsonFlagEnumConverter.cs

MediaBrowser.Model/Dlna/StreamInfo.cs
crobibero: @Shadowghost I should have specified- the `ToLowerInvariant()` is used in multiple places in the builder- none of which are required

Shadowghost: Should be fixed now. Tests required some fixing too.

MediaBrowser.Model/Dlna/StreamInfo.cs
Shadowghost: Should work now as intended

JPVenson (approved):

crobibero (approved):

Bond-009 (approved):
We can always convert the tests to use FsCheck later

PR #6203, opened 2021-06-19
Optimizations & Simplifications - Removed multiple from String.Split to Span

BaronGreenback:
Replaced all string.Split with Spans.
Simplified container contains code
Commented.
Nullable'd.
Added more tests.

Changed Profile.Id to Guid - in prep for DLNA plugin.

Before anyone points it out - i've had to duplicate the SplitStringExtension.cs as it cannot be accessed from model.
Looking to move this code out post DLNA plugins, so this is really only a temp step.

PR #9374, opened 2023-02-22
Optimizations & Simplifications - Removed multiple from String.Split to Span

Shadowghost:
Rebases the code on current master and adopts it to changes happening since the original PR.

Supersedes #6203


Emby.Dlna/Server/DescriptionXmlBuilder.cs
```diff
@@ -38,24 +38,14 @@ public string GetXml()
...
-            attributes.Insert(0, new XmlAttribute
-            {
-                Name = "xmlns",
-                Value = "urn:schemas-upnp-org:device-1-0"
-            });
-
-            foreach (var att in attributes)
-            {
-                builder.AppendFormat(CultureInfo.InvariantCulture, " {0}=\"{1}\"", att.Name, att.Value);
+                var attributes = _profile.XmlRootAttributes.ToList();
```
cvium, line 44: Why ToList?

MediaBrowser.Model/Dlna/DeviceProfile.cs
```diff
@@ -1,506 +1,520 @@
...
+    {
+        DlnaProfileType.Audio,
+        DlnaProfileType.Video,
+        DlnaProfileType.Photo
+    };
+
+    /// <summary>
+    /// Contains the string representation of <see cref="_supportedMediaTypes"/>.
+    /// </summary>
+    private string _supportedMediaTypeString = AllMedia;
```
cvium, line 40: private fields/props shouldn't have xmldoc

MediaBrowser.Model/Dlna/DirectPlayProfile.cs
```diff
@@ -1,36 +1,64 @@
...
-            return (Type == DlnaProfileType.Audio || Type == DlnaProfileType.Video) && ContainerProfile.ContainsContainer(AudioCodec, codec);
-        }
+    /// <summary>
+    /// Returns whether the <see cref="AudioCodec"/> supports the <paramref name="codec"/>.
+    /// </summary>
+    /// <param name="codec">The codec to match against.</param>
+    /// <returns>True if supported.</returns>
+    public bool SupportsAudioCodec(string? codec)
+    {
+        return Type != DlnaProfileType.Photo && AudioCodec.ContainsContainer(codec);
```
cvium, line 62: Subtitle is also a valid dlnaprofiletype

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -60,11 +62,17 @@ public StreamBuilder(ITranscoderSupport transcoderSupport, ILogger logger)
                 if (streamInfo is not null)
                 {
                     streamInfo.DeviceId = options.DeviceId;
-                    streamInfo.DeviceProfileId = options.Profile.Id;
+                    streamInfo.DeviceProfileId = options.Profile.Id.ToString("N", CultureInfo.InvariantCulture);
                     streams.Add(streamInfo);
                 }
             }
 
+            foreach (var stream in streams)
```
cvium, line 70: This seems a little pointless? streams is initially an empty list and is populated with StreamInfo instances that already have deviceid and profile id set.

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -352,30 +360,28 @@ private static TranscodeReason GetTranscodeReasonForFailedCondition(ProfileCondi
...
-            if (profile is not null)
+            var formats = inputContainer.SpanSplit(',');
+            foreach (var format in formats)
             {
-                var playProfiles = playProfile is null ? profile.DirectPlayProfiles : new[] { playProfile };
-                foreach (var format in formats)
+                foreach (var directPlayProfile in profile.DirectPlayProfiles)
                 {
-                    foreach (var directPlayProfile in playProfiles)
+                    if (directPlayProfile.Type == type)
```
cvium, line 373: Please reduce the "nesting" (i.e. invert the ifs)

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -803,23 +807,28 @@ private StreamInfo BuildVideoItem(MediaSourceInfo item, MediaOptions options)
...
 
-            // Enforce HLS video codec restrictions
-            if (string.Equals(playlistItem.SubProtocol, "hls", StringComparison.OrdinalIgnoreCase))
+            if (videoCodec is not null)
             {
-                videoCodecs = videoCodecs.Where(codec => _supportedHlsVideoCodecs.Contains(codec)).ToArray();
+                var splitString = videoCodec.SpanSplit(',');
+                while (splitString.MoveNext())
+                {
+                    videoCodecs.Add(splitString.Current.ToString());
```
cvium, line 821: Just use .Split. You would have to benchmark whether enumerating is actually more time and memory saving than simply allocating an array and adding it to the list. I am guessing that it isn't worth the added complexity.

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -845,23 +854,36 @@ private StreamInfo BuildVideoItem(MediaSourceInfo item, MediaOptions options)
...
+            {
+                audioCodecs.Add(audioStream.Codec);
+            }
+
+            if (audioCodec is not null)
+            {
+                var splitString = audioCodec.SpanSplit(',');
+                while (splitString.MoveNext())
+                {
+                    audioCodecs.Add(splitString.Current.ToString());
```
cvium, line 868: Same as above

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -908,16 +930,20 @@ private StreamInfo BuildVideoItem(MediaSourceInfo item, MediaOptions options)
...
-            foreach (var i in appliedVideoConditions)
+            if (videoCodec is not null)
             {
-                var transcodingVideoCodecs = ContainerProfile.SplitValue(videoCodec);
-                foreach (var transcodingVideoCodec in transcodingVideoCodecs)
+                foreach (var condition in appliedVideoConditions)
                 {
-                    if (i.ContainsAnyCodec(transcodingVideoCodec, container))
+                    var transcodingVideoCodecs = videoCodec.SpanSplit(',');
+                    while (transcodingVideoCodecs.MoveNext())
```
cvium, line 938: You can just use a regular foreach loop

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -940,16 +966,20 @@ private StreamInfo BuildVideoItem(MediaSourceInfo item, MediaOptions options)
...
-            foreach (var codecProfile in appliedAudioConditions)
+            if (audioCodec is not null)
             {
-                var transcodingAudioCodecs = ContainerProfile.SplitValue(audioCodec);
-                foreach (var transcodingAudioCodec in transcodingAudioCodecs)
+                foreach (var codecProfile in appliedAudioConditions)
                 {
-                    if (codecProfile.ContainsAnyCodec(transcodingAudioCodec, container))
+                    var transcodingAudioCodecs = audioCodec.SpanSplit(',');
+                    while (transcodingAudioCodecs.MoveNext())
```
cvium, line 974: foreach

MediaBrowser.Model/Extensions/ContainerExtensions.cs
```diff
@@ -0,0 +1,124 @@
...
+
+    /// <summary>
+    /// Compares two containers, returning <paramref name="isNegativeList"/> if an item in <paramref name="inputContainer"/>
+    /// does not exist in <paramref name="profileContainers"/>.
+    /// </summary>
+    /// <param name="profileContainers">The comma-delimited string being searched.</param>
+    /// <param name="isNegativeList">The boolean result to return if a match is not found.</param>
+    /// <param name="inputContainer">The comma-delimited string being matched.</param>
+    /// <returns>The result of the operation.</returns>
+    public static bool ContainsContainer(this string? profileContainers, bool isNegativeList, string? inputContainer)
```
cvium, line 59: Why do you have this overload? A string is automatically cast to a ReadOnlySpan

tests/Jellyfin.Model.Tests/Dlna/ContainerProfileTests.cs
```diff
@@ -1,19 +1,53 @@
...
-        {
-            Assert.True(_emptyContainerProfile.ContainsContainer(containers));
-        }
+    [InlineData("mp3,mpeg", "avi")]
+    [InlineData("mp3,mpeg", null)]
+    [InlineData("mp3,mpeg,avi", "mp4,jpg")]
+    [InlineData("-mp3,mpeg", "mp3")]
+    [InlineData("-mp3,mpeg,avi", "mpeg,avi")]
+    [Theory]
+    public void ContainsContainer_Contains_False(string container, string extension)
```
cvium, line 38: Naming could be improved e.g. `ContainsContainer_NotInList_ReturnsFalse`

MediaBrowser.Model/Dlna/DeviceProfile.cs
```diff
@@ -1,506 +1,517 @@
...
-        /// <returns>The .</returns>
-        public string[] GetSupportedMediaTypes()
+        get => _supportedMediaTypeString;
+        set
         {
-            return ContainerProfile.SplitValue(SupportedMediaTypes);
+            _supportedMediaTypeString = value ?? AllMedia;
+            var test = _supportedMediaTypeString
+                .Split(',')
+                .SelectMany(i => Enum.TryParse<DlnaProfileType>(i, true, out var result) && Enum.IsDefined(typeof(DlnaProfileType), result)
```
cvium, line 93: Can IsDefined return false when TryParse returns true?

Shadowghost: I don't think so. Should I remove it?

MediaBrowser.Model/Extensions/ContainerExtensions.cs
Shadowghost: After looking further into it: This overload is required because the input of the Span overload is not nullable. And apparently we have different behaviour for `null` and empty string: empty string allows all containers, null none.

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -846,23 +851,36 @@ private StreamInfo BuildVideoItem(MediaSourceInfo item, MediaOptions options)
...
+            var audioCodecs = new List<string>();
+            if (audioStream is not null && audioCodec.ContainsContainer(audioStream.Codec))
+            {
+                audioCodecs.Add(audioStream.Codec);
+            }
+
+            if (audioCodec is not null)
+            {
+                var splitStrings = audioCodec.Split(',');
+                foreach (var splitString in splitStrings)
```
cvium, line 863: You can just use AddRange

MediaBrowser.Model/Dlna/DeviceProfile.cs
```diff
@@ -1,507 +1,518 @@
...
-        /// The GetSupportedMediaTypes.
-        /// </summary>
-        /// <returns>The .</returns>
-        public string[] GetSupportedMediaTypes()
+        get => _supportedMediaTypeString;
+        set
         {
-            return ContainerProfile.SplitValue(SupportedMediaTypes);
+            _supportedMediaTypeString = value ?? AllMedia;
+            var test = _supportedMediaTypeString
```
cvium, line 92: Why is this called test? :P

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -799,23 +799,28 @@ private StreamInfo BuildVideoItem(MediaSourceInfo item, MediaOptions options)
...
 
-            // Enforce HLS video codec restrictions
-            if (string.Equals(playlistItem.SubProtocol, "hls", StringComparison.OrdinalIgnoreCase))
+            if (videoCodec is not null)
             {
-                videoCodecs = videoCodecs.Where(codec => _supportedHlsVideoCodecs.Contains(codec)).ToArray();
+                var splitStrings = videoCodec.Split(',');
+                foreach (var splitString in splitStrings)
+                {
+                    videoCodecs.Add(splitString);
```
cvium, line 813: Why not AddRange?

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -799,23 +799,25 @@ private StreamInfo BuildVideoItem(MediaSourceInfo item, MediaOptions options)
...
-            var videoCodecs = ContainerProfile.SplitValue(videoCodec);
+            List<string> videoCodecs = new();
+            if (videoStream is not null && videoCodec.ContainsContainer(videoStream.Codec))
+            {
+                videoCodecs.Add(videoStream.Codec!);
+            }
 
-            // Enforce HLS video codec restrictions
-            if (string.Equals(playlistItem.SubProtocol, "hls", StringComparison.OrdinalIgnoreCase))
+            if (videoCodec is not null)
```
cvium, line 808: If it's null, wouldn't it crash on line 803?

Shadowghost: We are checking for non-null there too.

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -969,15 +979,18 @@ private void BuildStreamVideoItem(
                 // Reverse codec profiles for backward compatibility - first codec profile has higher priority
                 .Reverse();
 
-            foreach (var i in appliedVideoConditions)
+            if (videoCodec is not null)
```
dmitrylyzo, line 982: I think this check will become unnecessary when https://github.com/jellyfin/jellyfin/pull/12521 is backported.

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -1002,15 +1015,18 @@ private void BuildStreamVideoItem(
                 // Reverse codec profiles for backward compatibility - first codec profile has higher priority
                 .Reverse();
 
-            foreach (var codecProfile in appliedAudioConditions)
+            if (audioCodec is not null)
```
dmitrylyzo, line 1018: I think this check will become unnecessary when https://github.com/jellyfin/jellyfin/pull/12521 is backported.

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -388,30 +389,30 @@ private static TranscodeReason GetTranscodeReasonForFailedCondition(ProfileCondi
         /// <returns>The normalized input container.</returns>
         public static string? NormalizeMediaSourceFormatIntoSingleContainer(string inputContainer, DeviceProfile? profile, DlnaProfileType type, DirectPlayProfile? playProfile = null)
         {
-            if (string.IsNullOrEmpty(inputContainer))
+            if (!inputContainer.Contains(',', StringComparison.OrdinalIgnoreCase) || profile == null)
             {
-                return null;
+                return inputContainer;
```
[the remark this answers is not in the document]
Shadowghost: Yes but I'd argue previous behaviour was not correct since something like `mkv` would return `null` which for sure can't be the intended return value for à normalization method.

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -856,23 +855,25 @@ private void BuildStreamVideoItem(
             string? audioCodec)
         {
             // Prefer matching video codecs
-            var videoCodecs = ContainerProfile.SplitValue(videoCodec);
+            List<string> videoCodecs = [];
+            if (videoStream is not null && videoCodec.ContainsContainer(videoStream.Codec))
+            {
+                videoCodecs.Add(videoStream.Codec!);
```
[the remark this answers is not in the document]
Shadowghost: I'd argue that if the video codec is compatible, we should never transcode as long as it is in the compatibility list (which is what this condition checks).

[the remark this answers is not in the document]
dmitrylyzo: The problem is that you can check if the video is copyable (the codec exists in the list and matches the given attributes), but when it is not, you will select an unwanted codec.

gnattu (approved):
Does not break #12420

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -386,32 +387,33 @@ private static TranscodeReason GetTranscodeReasonForFailedCondition(ProfileCondi
         /// <param name="type">The <see cref="DlnaProfileType"/>.</param>
         /// <param name="playProfile">The <see cref="DirectPlayProfile"/> object to get the video stream from.</param>
         /// <returns>The normalized input container.</returns>
-        public static string? NormalizeMediaSourceFormatIntoSingleContainer(string inputContainer, DeviceProfile? profile, DlnaProfileType type, DirectPlayProfile? playProfile = null)
+        public static string NormalizeMediaSourceFormatIntoSingleContainer(string inputContainer, DeviceProfile? profile, DlnaProfileType type, DirectPlayProfile? playProfile = null)
         {
-            if (string.IsNullOrEmpty(inputContainer))
+            if (!inputContainer.Contains(',', StringComparison.OrdinalIgnoreCase) || profile == null)
```
Bond-009, line 392:
```suggestion
            if (!inputContainer.Contains(',', StringComparison.OrdinalIgnoreCase) || profile is null)
```

MediaBrowser.Model/Extensions/ContainerExtensions.cs
```diff
@@ -0,0 +1,100 @@
...
+{
+    /// <summary>
+    /// Compares two containers, returning true if an item in <paramref name="inputContainer"/> exists
+    /// in <paramref name="profileContainers"/>.
+    /// </summary>
+    /// <param name="profileContainers">The comma-delimited string being searched.
+    /// If the parameter begins with the <c>-</c> character, the operation is reversed.</param>
+    /// <param name="inputContainer">The comma-delimited string being matched.</param>
+    /// <returns>The result of the operation.</returns>
+    public static bool ContainsContainer(this string? profileContainers, string? inputContainer)
```
Bond-009, line 19: I'd prefer a static helper class instead of extensions since this only makes sense for 1 specific use-case and not strings in general.

tests/Jellyfin.Model.Tests/Dlna/StreamBuilderTests.cs
```diff
@@ -362,9 +362,9 @@ public async Task BuildVideoItemWithDirectPlayExplicitStreams(string deviceName,
             if (playMethod == PlayMethod.DirectPlay)
             {
                 // Check expected container
-                var containers = ContainerProfile.SplitValue(mediaSource.Container);
+                var containers = mediaSource.Container.Split(",");
```
Bond-009, line 365:
```suggestion
                var containers = mediaSource.Container.Split(',');
```

Bond-009 (changes requested):

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -386,32 +387,33 @@ private static TranscodeReason GetTranscodeReasonForFailedCondition(ProfileCondi
...
             {
-                return null;
+                return inputContainer;
             }
 
-            var formats = ContainerProfile.SplitValue(inputContainer);
-
-            if (profile is not null)
+            var formats = inputContainer.SpanSplit(',');
+            var playProfiles = playProfile is null ? profile.DirectPlayProfiles : new[] { playProfile };
```
dmitrylyzo, line 398: Since you want to get rid of `new[] { ... }`.
```suggestion
            var playProfiles = playProfile is null ? profile.DirectPlayProfiles : [playProfile];
```

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -825,6 +827,7 @@ private StreamInfo BuildVideoItem(MediaSourceInfo item, MediaOptions options)
             var audioBitrate = audioStream?.BitRate;
             var audioSampleRate = audioStream?.SampleRate;
             var audioBitDepth = audioStream?.BitDepth;
+            var useSubContainer = playlistItem.SubProtocol == MediaStreamProtocol.hls;
```
dmitrylyzo, line 830: I doubt that `playlistItem` is populated.
We only decide how to play.
We can probably always use a subcontainer if one exists.

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -856,11 +859,11 @@ private StreamInfo BuildVideoItem(MediaSourceInfo item, MediaOptions options)
                     {
                         var audioCodecs = ContainerProfile.SplitValue(transcodingProfile.AudioCodec);
```
dmitrylyzo: remove

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -621,8 +623,8 @@ private static void SetStreamInfoOptionsFromDirectPlayProfile(MediaOptions optio
             playlistItem.Container = container;
             playlistItem.SubProtocol = protocol;
 
-            playlistItem.VideoCodecs = new[] { item.VideoStream.Codec };
-            playlistItem.AudioCodecs = ContainerProfile.SplitValue(directPlayProfile?.AudioCodec);
+            playlistItem.VideoCodecs = [item.VideoStream.Codec];
+            playlistItem.AudioCodecs = directPlayProfile?.AudioCodec?.Split(',', StringSplitOptions.RemoveEmptyEntries) ?? Array.Empty<string>();
```
dmitrylyzo, line 627: Maybe also do it in `ContainerHelper.Split(...)`?

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -893,22 +894,31 @@ private void BuildStreamVideoItem(
             }
 
             // Prefer matching audio codecs, could do better here
-            var audioCodecs = ContainerProfile.SplitValue(audioCodec);
+            var audioCodecs = new List<string>();
+            if (audioStream is not null && audioCodec.ContainsContainer(audioStream.Codec))
+            {
+                audioCodecs.Add(audioStream.Codec);
```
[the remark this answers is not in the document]
dmitrylyzo: Reminder: should do the same as for the video codec.

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -899,23 +898,24 @@ private void BuildStreamVideoItem(
             string? audioCodec)
         {
             // Prefer matching video codecs
-            var videoCodecs = ContainerProfile.SplitValue(videoCodec);
-
-            // Enforce HLS video codec restrictions
-            if (playlistItem.SubProtocol == MediaStreamProtocol.hls)
+            List<string> videoCodecs = [];
+            if (videoCodec is not null)
```
dmitrylyzo, line 902: Hmm, what if `videoCodec` is not null, but gives an empty array?

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -1009,18 +1019,17 @@ private void BuildStreamVideoItem(
 
             var appliedVideoConditions = options.Profile.CodecProfiles
                 .Where(i => i.Type == CodecType.Video &&
-                    i.ContainsAnyCodec(videoCodecs, container, useSubContainer) &&
+                    i.ContainsAnyCodec([.. videoCodecs], container, useSubContainer) &&
```
dmitrylyzo, line 1022:
```suggestion
                    i.ContainsAnyCodec(playlistItem.VideoCodecs, container, useSubContainer) &&
```

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -1041,19 +1050,23 @@ private void BuildStreamVideoItem(
 
             var appliedAudioConditions = options.Profile.CodecProfiles
                 .Where(i => i.Type == CodecType.VideoAudio &&
-                    i.ContainsAnyCodec(audioCodecs, container) &&
+                    i.ContainsAnyCodec([.. audioCodecs], container) &&
```
dmitrylyzo, line 1053:
```suggestion
                    i.ContainsAnyCodec(playlistItem.AudioCodecs, container) &&
```

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -1041,19 +1050,23 @@ private void BuildStreamVideoItem(
...
                     i.ApplyConditions.All(applyCondition => ConditionProcessor.IsVideoAudioConditionSatisfied(applyCondition, audioChannels, inputAudioBitrate, inputAudioSampleRate, inputAudioBitDepth, audioProfile, isSecondaryAudio)))
                 // Reverse codec profiles for backward compatibility - first codec profile has higher priority
                 .Reverse();
 
-            foreach (var codecProfile in appliedAudioConditions)
+            var playlistAudioCodecs = playlistItem.AudioCodecs;
+            if (playlistAudioCodecs is not null)
             {
```
dmitrylyzo, line 1060: Remove this check - `appliedAudioConditions` will be empty if there's no audio codec.

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -1009,18 +1019,17 @@ private void BuildStreamVideoItem(
...
-                    i.ContainsAnyCodec(videoCodecs, container, useSubContainer) &&
+                    i.ContainsAnyCodec([.. videoCodecs], container, useSubContainer) &&
                     i.ApplyConditions.All(applyCondition => ConditionProcessor.IsVideoConditionSatisfied(applyCondition, width, height, bitDepth, videoBitrate, videoProfile, videoRangeType, videoLevel, videoFramerate, packetLength, timestamp, isAnamorphic, isInterlaced, refFrames, numVideoStreams, numAudioStreams, videoCodecTag, isAvc)))
                 // Reverse codec profiles for backward compatibility - first codec profile has higher priority
                 .Reverse();
-
-            foreach (var i in appliedVideoConditions)
+            foreach (var condition in appliedVideoConditions)
             {
                 foreach (var transcodingVideoCodec in videoCodecs)
```
dmitrylyzo, line 1028: Ideally, use the same array. I probably should have suggested it to Nyan right away.
```suggestion
                foreach (var transcodingVideoCodec in playlistItem.VideoCodecs)
```

MediaBrowser.Model/Dlna/StreamBuilder.cs
Shadowghost: It is explicitly marked as non-nullable...

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -936,22 +936,31 @@ private void BuildStreamVideoItem(
...
+            var audioCodecs = new List<string>();
+            if (audioStream is not null && ContainerHelper.ContainsContainer(audioCodec, audioStream.Codec))
+            {
+                audioCodecs.Add(audioStream.Codec);
+            }
+
+            if (audioCodec is not null)
+            {
+                audioCodecs.AddRange(audioCodec.Split(','));
+            }
```
dmitrylyzo, line 948:
```suggestion
            if (audioCodec is not null)
            {
                audioCodecs.AddRange(audioCodec.Split(',', StringSplitOptions.RemoveEmptyEntries));
            }

            if (audioCodecs.Length == 0 && audioStream is not null)
            {
                // Add the original codec if no codec is specified
                audioCodecs.Add(audioStream.Codec);
            }

```

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -899,23 +898,24 @@ private void BuildStreamVideoItem(
...
             // Prefer matching video codecs
-            var videoCodecs = ContainerProfile.SplitValue(videoCodec);
-
-            // Enforce HLS video codec restrictions
-            if (playlistItem.SubProtocol == MediaStreamProtocol.hls)
+            List<string> videoCodecs = [];
+            if (videoCodec is not null)
+            {
+                var splitStrings = videoCodec.Split(',');
+                videoCodecs.AddRange(splitStrings);
```
dmitrylyzo, line 905:
```suggestion
                videoCodecs.AddRange(videoCodec.Split(',', StringSplitOptions.RemoveEmptyEntries));
```

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -899,23 +898,24 @@ private void BuildStreamVideoItem(
...
+            }
+            else if (videoStream is not null)
```
dmitrylyzo, line 907: Hmm, what if `videoCodec` is not null, but gives an empty array?
```suggestion
            }

            if (videoCodecs.Length == 0 && videoStream is not null)
```

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -899,23 +898,24 @@ private void BuildStreamVideoItem(
...
             {
-                videoCodecs = videoCodecs.Where(codec => _supportedHlsVideoCodecs.Contains(codec)).ToArray();
+                videoCodecs.Add(videoStream.Codec!);
```
dmitrylyzo, line 909:
```suggestion
                // Add the original codec if no codec is specified
                videoCodecs.Add(videoStream.Codec!);
```

MediaBrowser.Model/Dlna/StreamBuilder.cs
dmitrylyzo: It will be set after we decide - that's the problem.

Shadowghost: So ignore it here? and reert to what was there before while not considering the subprotocol?

dmitrylyzo: I would say it needs to be reverted, for now.
But it might be a good idea to just change `ContainsAnyCodec` so that it always prefers `SubContainer` if set.
We can do it in another PR.

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -899,23 +898,25 @@ private void BuildStreamVideoItem(
             string? audioCodec)
         {
             // Prefer matching video codecs
-            var videoCodecs = ContainerProfile.SplitValue(videoCodec);
+            List<string> videoCodecs = [];
+            if (audioCodec is not null)
```
dmitrylyzo, line 902:
```suggestion
            if (videoCodec is not null)
```

MediaBrowser.Model/Dlna/StreamBuilder.cs
dmitrylyzo: Hmm, we already have `playlistItem.SubProtocol == MediaStreamProtocol.hls`.

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -711,15 +713,15 @@ private StreamInfo BuildVideoItem(MediaSourceInfo item, MediaOptions options)
...
+                            playlistItem.AudioCodecs = audioCodec is null ? Array.Empty<string>() : [audioCodec];
                         }
                     }
                     else if (directPlay == PlayMethod.DirectStream)
                     {
                         playlistItem.AudioStreamIndex = audioStream?.Index;
                         if (audioStream is not null)
                         {
-                            playlistItem.AudioCodecs = ContainerProfile.SplitValue(directPlayProfile?.AudioCodec);
+                            playlistItem.AudioCodecs = directPlayProfile?.AudioCodec?.Split(',', StringSplitOptions.RemoveEmptyEntries) ?? Array.Empty<string>();
```
dmitrylyzo, line 724:
```suggestion
                            playlistItem.AudioCodecs = ContainerHelper.Split(directPlayProfile?.AudioCodec);
```

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -899,23 +898,25 @@ private void BuildStreamVideoItem(
             string? audioCodec)
         {
             // Prefer matching video codecs
-            var videoCodecs = ContainerProfile.SplitValue(videoCodec);
+            List<string> videoCodecs = [];
+            if (audioCodec is not null)
+            {
+                videoCodecs.AddRange(ContainerHelper.Split(videoCodec));
+            }
```
dmitrylyzo, line 905: Btw, is this work? But we probably lose the benefit of the `List'.
```suggestion
            List<string> videoCodecs = ContainerHelper.Split(videoCodec);
```

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -936,22 +937,32 @@ private void BuildStreamVideoItem(
             }
 
             // Prefer matching audio codecs, could do better here
-            var audioCodecs = ContainerProfile.SplitValue(audioCodec);
+            List<string> audioCodecs = [];
+            if (audioCodec is not null)
+            {
+                audioCodecs.AddRange(ContainerHelper.Split(audioCodec));
+            }
```
dmitrylyzo, line 944: Btw, is this work? But we probably lose the benefit of the `List'.
```suggestion
            List<string> audioCodecs = ContainerHelper.Split(audioCodec);
```

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -936,22 +937,32 @@ private void BuildStreamVideoItem(
...
                 }
                 else
                 {
-                    audioCodecs = audioCodecs.Where(codec => _supportedHlsAudioCodecsTs.Contains(codec)).ToArray();
+                    audioCodecs = audioCodecs.Where(codec => _supportedHlsAudioCodecsTs.Contains(codec)).ToList();
                 }
             }
 
-            var audioStreamWithSupportedCodec = candidateAudioStreams.Where(stream => ContainerProfile.ContainsContainer(audioCodecs, stream.Codec)).FirstOrDefault();
+            var audioStreamWithSupportedCodec = candidateAudioStreams.Where(stream => playlistItem.AudioCodecs.Contains(stream.Codec, StringComparison.OrdinalIgnoreCase)).FirstOrDefault();
```
dmitrylyzo, line 965: To reuse the codec comparison logic. `playlistItem.AudioCodecs` is also not yet populated.
```suggestion
            var audioStreamWithSupportedCodec = candidateAudioStreams.Where(stream => ContainerHelper.ContainsContainer(audioCodecs, stream.Codec)).FirstOrDefault();
```

MediaBrowser.Model/Dlna/CodecProfile.cs
```diff
@@ -1,74 +1,94 @@
...
+    /// Checks to see whether the codecs and containers contain the given parameters.
+    /// </summary>
+    /// <param name="codec">The codec to match.</param>
+    /// <param name="container">The container to match.</param>
+    /// <param name="useSubContainer">Consider subcontainers.</param>
+    /// <returns>True if both conditions are met.</returns>
+    public bool ContainsAnyCodec(ReadOnlySpan<char> codec, string? container, bool useSubContainer = false)
+    {
+        var containerToCheck = useSubContainer && string.Equals(Container, "hls", StringComparison.OrdinalIgnoreCase) ? SubContainer : Container;
+        return ContainerHelper.ContainsContainer(containerToCheck, false, container) && ContainerHelper.ContainsContainer(containerToCheck, false, codec);
```
gnattu, line 92:
```suggestion
        return ContainerHelper.ContainsContainer(containerToCheck, false, container) && ContainerHelper.ContainsContainer(Codec, false, codec);
```

MediaBrowser.Model/Dlna/CodecProfile.cs
```diff
@@ -1,74 +1,94 @@
...
+    /// Checks to see whether the codecs and containers contain the given parameters.
+    /// </summary>
+    /// <param name="codec">The codec to match.</param>
+    /// <param name="container">The container to match.</param>
+    /// <param name="useSubContainer">Consider subcontainers.</param>
+    /// <returns>True if both conditions are met.</returns>
+    public bool ContainsAnyCodec(string? codec, string? container, bool useSubContainer = false)
+    {
+        var containerToCheck = useSubContainer && string.Equals(Container, "hls", StringComparison.OrdinalIgnoreCase) ? SubContainer : Container;
+        return ContainerHelper.ContainsContainer(containerToCheck, container) && ContainerHelper.ContainsContainer(Container, false, codec);
```
gnattu, line 79:
```suggestion
        return ContainerHelper.ContainsContainer(containerToCheck, container) && ContainerHelper.ContainsContainer(Codec, false, codec);
```

gnattu (changes requested):
We are comparing `Codec` with `Container` and breaks all condition checks

tests/Jellyfin.Model.Tests/Dlna/StreamBuilderTests.cs
```diff
@@ -297,7 +297,7 @@ public async Task BuildVideoItemWithFirstExplicitStream(string deviceName, strin
         [InlineData("RokuSSPlus", "mp4-h264-ac3-aac-srt-2600k", PlayMethod.DirectPlay, (TranscodeReason)0, "Remux")] // #6450
         [InlineData("RokuSSPlus", "mp4-hevc-ac3-aac-srt-15200k", PlayMethod.DirectPlay, (TranscodeReason)0, "Remux")] // #6450
         // no streams
-        [InlineData("Chrome", "no-streams", PlayMethod.Transcode, TranscodeReason.VideoCodecNotSupported, "Transcode")] // #6450
+        [InlineData("Chrome", "no-streams", PlayMethod.DirectPlay, (TranscodeReason)0, "Remux")]
```
dmitrylyzo, line 300: Just a note: this has changed because the `null` input codec is now accepted by any codec list. Previously it was only accepted by a negative codec list.

Before:
https://github.com/jellyfin/jellyfin/blob/c67b78bc68317c66f3324230b2fc36918b2b9d70/MediaBrowser.Model/Dlna/ContainerProfile.cs#L38
After:
https://github.com/jellyfin/jellyfin/blob/d9c05036b8e96e928d6943644d2002573b2a466c/MediaBrowser.Model/Extensions/ContainerHelper.cs#L22

Shadowghost: Thanks for the fine. I think how it is right now is more consistent.

dmitrylyzo: Well, it depends on how to interpret the `null` codec. Previously it probably meant "unknown", which makes sense.

It might be a good idea to treat `null` as "not supported" by both, positive and negative, codec lists. But this also changes the previous logic.

I would not change any logic in "optimization" PR (i.e. refactoring).

Shadowghost: So revert back to old behavior and ignore it only on negative list?

dmitrylyzo: Yes, for now.

MediaBrowser.Model/Dlna/CodecProfile.cs
```diff
@@ -1,74 +1,94 @@
...
+    /// Checks to see whether the codecs and containers contain the given parameters.
+    /// </summary>
+    /// <param name="codec">The codec to match.</param>
+    /// <param name="container">The container to match.</param>
+    /// <param name="useSubContainer">Consider subcontainers.</param>
+    /// <returns>True if both conditions are met.</returns>
+    public bool ContainsAnyCodec(ReadOnlySpan<char> codec, string? container, bool useSubContainer = false)
+    {
+        var containerToCheck = useSubContainer && string.Equals(Container, "hls", StringComparison.OrdinalIgnoreCase) ? SubContainer : Container;
+        return ContainerHelper.ContainsContainer(containerToCheck, false, container) && ContainerHelper.ContainsContainer(Codec, false, codec);
```
dmitrylyzo, line 92: It looks like `Container` supported inverse matching.
```suggestion
        return ContainerHelper.ContainsContainer(containerToCheck, container) && ContainerHelper.ContainsContainer(Codec, false, codec);
```

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -386,32 +387,33 @@ private static TranscodeReason GetTranscodeReasonForFailedCondition(ProfileCondi
...
+            if (!inputContainer.Contains(',', StringComparison.OrdinalIgnoreCase) || profile is null)
             {
-                return null;
+                return inputContainer;
             }
 
-            var formats = ContainerProfile.SplitValue(inputContainer);
-
-            if (profile is not null)
+            var formats = inputContainer.SpanSplit(',');
```
dmitrylyzo, line 397: Maybe add `ContainerHelper.SpanSplit` for consistency?

MediaBrowser.Model/Dlna/StreamBuilder.cs
dmitrylyzo: Remove `useSubContainer` from here, for now. It is always `false`.
As gnattu said, we probably need to rethink the use of `SubContainer`.

gnattu:
> Remove useSubContainer from here, for now. It is always false.

It *is* true when the client request it

MediaBrowser.Model/Dlna/StreamBuilder.cs
Shadowghost: This is a generic method from our string extension.

dmitrylyzo:
> It _is_ true when the client request it

This is a different function. You are talking about https://github.com/jellyfin/jellyfin/pull/9374#discussion_r1751009750

gnattu:
nvm, yes, in that line the protocol is not populated yet.

MediaBrowser.Model/Dlna/StreamBuilder.cs
dmitrylyzo: I was thinking of passing `RemoveEmptyEntries`.

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -1041,18 +1042,18 @@ private void BuildStreamVideoItem(
...
 
             foreach (var codecProfile in appliedAudioConditions)
             {
-                foreach (var transcodingAudioCodec in audioCodecs)
+                foreach (var transcodingAudioCodec in playlistItem.AudioCodecs)
                 {
                     if (codecProfile.ContainsAnyCodec(transcodingAudioCodec, container))
                     {
-                        ApplyTranscodingConditions(playlistItem, codecProfile.Conditions, transcodingAudioCodec, true, true);
+                        ApplyTranscodingConditions(playlistItem, codecProfile.Conditions, transcodingAudioCodec.ToString(), true, true);
```
dmitrylyzo, line 1056: It is already a string.
```suggestion
                        ApplyTranscodingConditions(playlistItem, codecProfile.Conditions, transcodingAudioCodec, true, true);
```

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -386,32 +387,33 @@ private static TranscodeReason GetTranscodeReasonForFailedCondition(ProfileCondi
...
+            if (!inputContainer.Contains(',', StringComparison.OrdinalIgnoreCase) || profile is null)
             {
-                return null;
+                return inputContainer;
             }
 
-            var formats = ContainerProfile.SplitValue(inputContainer);
-
-            if (profile is not null)
+            var formats = ContainerHelper.Split(inputContainer);
```
dmitrylyzo, line 397: I thought you wanted to minimize the allocation by using Span.
If you are going to leave it as is, remove `ToString` below.

dmitrylyzo (commented):
Another round. 😅

What about List vs Array?
In the worst case:
- VideoCodecs: 2 conversions (array -> list, list -> array), 1 addition and 1 filter.
- AudioCodecs: 3 conversions, 1 addition and 1 filter.

We can remove the conversions by reverting to array. Addition -> new array.
Is it worth it?

MediaBrowser.Model/Dlna/StreamBuilder.cs
Shadowghost: If I understood it correctly, we can't use Span if we want to remove empty. The current SplitSpan implementation just converts the string to a Span Enumerator.
This PR was initially only a forward-port of an already existing one 😅

dmitrylyzo: Oh, I thought that was a built-in feature. nvm then.

For further optimization (separate PR), we need to improve this by adding split flags:
https://github.com/jellyfin/jellyfin/blob/751e12e5b5c59f9df4494251166f47c8cf9ebfe1/src/Jellyfin.Extensions/SplitStringExtensions.cs#L93

gnattu (approved):
LGTM from the StreamBuilder and playback side. Things don't break even with the reworked unit tests.

MediaBrowser.Model/Dlna/CodecProfile.cs
```diff
@@ -1,74 +1,108 @@
...
-                return false;
-            }
+    /// <summary>
+    /// Checks to see whether the codecs and containers contain the given parameters.
+    /// </summary>
+    /// <param name="codecs">The codecs to match.</param>
+    /// <param name="container">The container to match.</param>
+    /// <param name="useSubContainer">Consider sub-containers.</param>
+    /// <returns>True if both conditions are met.</returns>
+    public bool ContainsAnyCodec(string[] codecs, string? container, bool useSubContainer = false)
```
Bond-009, line 64: Why does this overload exist? And if it's for API compat, why doesn't it just call the ReadOnlyList overload?

MediaBrowser.Model/Dlna/CodecProfile.cs
```diff
@@ -1,74 +1,108 @@
...
-                }
-            }
+    /// <summary>
+    /// Checks to see whether the codecs and containers contain the given parameters.
+    /// </summary>
+    /// <param name="codec">The codec to match.</param>
+    /// <param name="container">The container to match.</param>
+    /// <param name="useSubContainer">Consider sub-containers.</param>
+    /// <returns>True if both conditions are met.</returns>
+    public bool ContainsAnyCodec(string? codec, string? container, bool useSubContainer = false)
```
Bond-009, line 90: Same as above, but with the `ReadOnlySpan` overload

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -386,32 +387,33 @@ private static TranscodeReason GetTranscodeReasonForFailedCondition(ProfileCondi
         /// <param name="type">The <see cref="DlnaProfileType"/>.</param>
         /// <param name="playProfile">The <see cref="DirectPlayProfile"/> object to get the video stream from.</param>
         /// <returns>The normalized input container.</returns>
-        public static string? NormalizeMediaSourceFormatIntoSingleContainer(string inputContainer, DeviceProfile? profile, DlnaProfileType type, DirectPlayProfile? playProfile = null)
+        public static string NormalizeMediaSourceFormatIntoSingleContainer(string inputContainer, DeviceProfile? profile, DlnaProfileType type, DirectPlayProfile? playProfile = null)
         {
-            if (string.IsNullOrEmpty(inputContainer))
+            if (!inputContainer.Contains(',', StringComparison.OrdinalIgnoreCase) || profile is null)
```
Bond-009, line 392: I prefer putting the cheapest operation first (in this case we don't have to de the relatively expensive contains operation if profile is null)
```suggestion
            if (profile is null || !inputContainer.Contains(',', StringComparison.OrdinalIgnoreCase))
```

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -936,22 +930,28 @@ private void BuildStreamVideoItem(
             }
 
             // Prefer matching audio codecs, could do better here
-            var audioCodecs = ContainerProfile.SplitValue(audioCodec);
+            List<string> audioCodecs = [.. ContainerHelper.Split(audioCodec)];
```
Bond-009, line 933: I still prefer ToArray and ToList in these cases, way clearer IMHO
(not just here, but all over the PR BTW)

MediaBrowser.Model/Extensions/ContainerHelper.cs
```diff
@@ -0,0 +1,142 @@
...
+            return true;
+        }
+
+        var allInputContainers = inputContainer.Split(',');
+        var allProfileContainers = profileContainers.SpanSplit(',');
+        foreach (var container in allInputContainers)
+        {
+            foreach (var profile in allProfileContainers)
+            {
+                if (MemoryExtensions.Equals(profile, container, StringComparison.OrdinalIgnoreCase))
```
Bond-009, line 92: Can't you just do the following?
```suggestion
                if (container.Equals(profile, StringComparison.OrdinalIgnoreCase))
```

MediaBrowser.Model/Extensions/ContainerHelper.cs
```diff
@@ -0,0 +1,142 @@
...
+            // Empty profiles always support all containers/codecs.
+            return true;
+        }
+
+        var allInputContainers = inputContainer.Split(',');
+        foreach (var container in allInputContainers)
+        {
+            foreach (var profile in profileContainers)
+            {
+                if (MemoryExtensions.Equals(profile, container, StringComparison.OrdinalIgnoreCase))
```
Bond-009, line 123: Both are strings, no?
```suggestion
                if (string.Equals(profile, container, StringComparison.OrdinalIgnoreCase))
```

tests/Jellyfin.Model.Tests/Dlna/ContainerHelperTests.cs
```diff
@@ -0,0 +1,52 @@
...
+    public void ContainsContainer_NotInList_ReturnsFalse(string container, string? extension)
+    {
+        Assert.False(ContainerHelper.ContainsContainer(container, extension));
+    }
+
+    [InlineData("mp3,mpeg", "mp3")]
+    [InlineData("mp3,mpeg,avi", "mp3,avi")]
+    [InlineData("-mp3,mpeg", "avi")]
+    [InlineData("-mp3,mpeg,avi", "mp4,jpg")]
+    [Theory]
```
Bond-009, line 47:
```suggestion
    [Theory]
    [InlineData("mp3,mpeg", "mp3")]
    [InlineData("mp3,mpeg,avi", "mp3,avi")]
    [InlineData("-mp3,mpeg", "avi")]
    [InlineData("-mp3,mpeg,avi", "mp4,jpg")]
    public void ContainsContainer_InList_ReturnsTrue(string container, string? extension)
    {
        Assert.True(ContainerHelper.ContainsContainer(container, extension));
    }

    [Theory]
    [InlineData("mp3,mpeg", "avi")]
    [InlineData("mp3,mpeg,avi", "mp4,jpg")]
    [InlineData("mp3,mpeg", null)]
    [InlineData("mp3,mpeg", "")]
    [InlineData("-mp3,mpeg", "mp3")]
    [InlineData("-mp3,mpeg,avi", "mpeg,avi")]
    public void ContainsContainer_NotInList_ReturnsFalse(string container, string? extension)
    {
        Assert.False(ContainerHelper.ContainsContainer(container, extension));
    }

    [Theory]
    [InlineData("mp3,mpeg", "mp3")]
    [InlineData("mp3,mpeg,avi", "mp3,avi")]
    [InlineData("-mp3,mpeg", "avi")]
    [InlineData("-mp3,mpeg,avi", "mp4,jpg")]
```

Bond-009 (approved):
Mostly style stuff, some questions about duplicated functions.
I'd like to see the comments addressed but this can be merged without another look from me. Unless I need to clarify some things ofc

MediaBrowser.Model/Dlna/StreamBuilder.cs
Shadowghost: Agreed, I reverted it.

MediaBrowser.Model/Dlna/CodecProfile.cs
Shadowghost: I removed it since it is now unused anyways.

tests/Jellyfin.Model.Tests/Dlna/ContainerHelperTests.cs
```diff
@@ -0,0 +1,52 @@
...
+        Assert.True(ContainerHelper.ContainsContainer(container, extension));
+    }
+
+    [Theory]
+    [InlineData("mp3,mpeg", "avi")]
+    [InlineData("mp3,mpeg,avi", "mp4,jpg")]
+    [InlineData("mp3,mpeg", null)]
+    [InlineData("mp3,mpeg", "")]
+    [InlineData("-mp3,mpeg", "mp3")]
+    [InlineData("-mp3,mpeg,avi", "mpeg,avi")]
```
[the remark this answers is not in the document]
dmitrylyzo: > The enumerator extension returns if there are no characters left, so an actual test would be something like
> 
> ```cs
> [InlineData(",mp3", ",avi")] // empty values should be discarded
> ```

:+1: 
Added on both sides just to be sure.

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -1400,7 +1382,9 @@ private static int GetMaxAudioBitrateForTotalBitrate(long totalBitrate)
                         playMethod = PlayMethod.DirectStream;
                     }
 
-                    var ranked = rank(ref failureReasons);
+                    TranscodeReason[] rankings = [TranscodeReason.VideoCodecNotSupported, VideoCodecReasons, TranscodeReason.AudioCodecNotSupported, AudioCodecReasons, ContainerReasons];
```
dmitrylyzo, line 1385: Will this array be re-allocated each iteration? Or will it be considered const?
I would move it out of `Select`.

Generally speaking, I would leave the ranking untouched (except for `new[]`) and refactor it in the next PR. Current PR keeps growing.

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -388,32 +389,33 @@ private static TranscodeReason GetTranscodeReasonForFailedCondition(ProfileCondi
...
-                        {
-                            return format;
-                        }
+                        continue;
+                    }
+
+                    var formatStr = format.ToString();
+                    if (directPlayProfile.SupportsContainer(formatStr))
+                    {
+                        return formatStr;
```
dmitrylyzo, line 413: As I said before (https://github.com/jellyfin/jellyfin/pull/9374#discussion_r1756546326), unless we are using `SpanSplit`, `ToString` is redundant here.
```suggestion
                    if (directPlayProfile.SupportsContainer(format))
                    {
                        return format;
```

MediaBrowser.Model/Extensions/ContainerHelper.cs
```diff
@@ -0,0 +1,142 @@
...
+    /// <returns>The result of the operation.</returns>
+    public static bool ContainsContainer(IReadOnlyList<string>? profileContainers, bool isNegativeList, string inputContainer)
+    {
+        if (profileContainers is null)
+        {
+            // Empty profiles always support all containers/codecs.
+            return true;
+        }
+
+        var allInputContainers = inputContainer.Split(',');
```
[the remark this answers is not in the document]
dmitrylyzo: What about this? ☝️

tests/Jellyfin.Model.Tests/Dlna/ContainerHelperTests.cs
```diff
@@ -0,0 +1,54 @@
...
+    [InlineData(",mp3,", ",avi,")] // Empty values should be discarded
+    [InlineData("-,mp3,", ",mp3,")] // Empty values should be discarded
+    public void ContainsContainer_NotInList_ReturnsFalse(string container, string? extension)
+    {
+        Assert.False(ContainerHelper.ContainsContainer(container, extension));
```
dmitrylyzo, line 42: To test falsy span version.
```suggestion
        Assert.False(ContainerHelper.ContainsContainer(container, extension));
        
        if (extension is not null)
        {
            Assert.False(ContainerHelper.ContainsContainer(container, extension.AsSpan()));
        }
```
Tbf, I would implement all tests this way - the same test data.

MediaBrowser.Model/Extensions/ContainerHelper.cs
```diff
@@ -0,0 +1,145 @@
...
+
+    /// <summary>
+    /// Compares two containers, returning <paramref name="isNegativeList"/> if an item in <paramref name="inputContainer"/>
+    /// does not exist in <paramref name="profileContainers"/>.
+    /// </summary>
+    /// <param name="profileContainers">The profile containers being matched searched.</param>
+    /// <param name="isNegativeList">The boolean result to return if a match is not found.</param>
+    /// <param name="inputContainer">The comma-delimited string being matched.</param>
+    /// <returns>The result of the operation.</returns>
+    public static bool ContainsContainer(IReadOnlyList<string>? profileContainers, bool isNegativeList, string inputContainer)
```
dmitrylyzo, line 113: There's no test for this

MediaBrowser.Model/Extensions/ContainerHelper.cs
```diff
@@ -0,0 +1,142 @@
...
+    public static bool ContainsContainer(string? profileContainers, bool isNegativeList, ReadOnlySpan<char> inputContainer)
+    {
+        if (string.IsNullOrEmpty(profileContainers))
+        {
+            // Empty profiles always support all containers/codecs.
+            return true;
+        }
+
+        var allInputContainers = inputContainer.Split(',');
+        var allProfileContainers = profileContainers.SpanSplit(',');
```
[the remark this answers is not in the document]
Shadowghost: We're now checking `if (!container.IsEmpty)` within the loop which filters out empty values.

MediaBrowser.Model/Extensions/ContainerHelper.cs
```diff
@@ -0,0 +1,142 @@
...
+    /// <returns>The result of the operation.</returns>
+    public static bool ContainsContainer(string? profileContainers, bool isNegativeList, ReadOnlySpan<char> inputContainer)
+    {
+        if (string.IsNullOrEmpty(profileContainers))
+        {
+            // Empty profiles always support all containers/codecs.
+            return true;
+        }
+
+        var allInputContainers = inputContainer.Split(',');
```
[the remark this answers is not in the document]
Shadowghost: Added `!profile.IsEmpty` to the `if`

dmitrylyzo (approved):

crobibero (approved):

MediaBrowser.Model/Dlna/StreamBuilder.cs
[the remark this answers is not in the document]
gnattu: Late to the party but the old behavior is necessary because for LiveTV channel the inputContainer is indeed null upon first access because the source was never probed before.

PR #9381, opened 2023-02-23
Enable nullable for more files

Bond-009:


Shadowghost (approved):

1337joe (approved):

MediaBrowser.Providers/Plugins/Tmdb/TV/TmdbEpisodeImageProvider.cs
```diff
@@ -63,7 +61,7 @@ public async Task<IEnumerable<RemoteImageInfo>> GetImages(BaseItem item, Cancell
 
             var seriesTmdbId = Convert.ToInt32(series?.GetProviderId(MetadataProvider.Tmdb), CultureInfo.InvariantCulture);
 
-            if (seriesTmdbId <= 0)
+            if (series == null || seriesTmdbId <= 0)
```
crobibero, line 64:
```suggestion
            if (series is null || seriesTmdbId <= 0)
```

Bond-009: Muscle memory ;)

crobibero (approved):

PR #9403, opened 2023-02-26
Extras: Clean filename and process NFOs

SenorSmartyPants:
**Changes**

- Filenames are cleaned using CleanStrings regex
- new regex to remove extra suffixes
- NFOs for extras are now read

A file named `Interview with director-interview.mkv` will now be named `Interview with director` in JF.

NFO files named the same as extra file will be read for metadata.

### DIFF
--- a/Emby.Naming/Common/NamingOptions.cs
+++ b/Emby.Naming/Common/NamingOptions.cs
@@ -157,7 +157,8 @@ namespace Emby.Naming.Common
                 @"^(?<cleaned>.+?)(\[.*\])",
                 @"^\s*(?<cleaned>.+?)\WE[0-9]+(-|~)E?[0-9]+(\W|$)",
                 @"^\s*\[[^\]]+\](?!\.\w+$)\s*(?<cleaned>.+)",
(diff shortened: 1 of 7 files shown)


Shadowghost (approved):

SenorSmartyPants:
Any feedback on the unit test I changed? I'm not aware of which issues that might be related to, or if my change will affect them.

daullmer (approved):

1337joe:
> Any feedback on the unit test I changed? I'm not aware of which issues that might be related to, or if my change will affect them.

I added that test when I did some work on `ProviderManager` to ensure that behavior didn't change during my refactor. The check itself is from Emby, I just made the comment more relevant.

I've never used NFOs so I can't attest to the necessity of that check. I expect that if you're changing the behavior you've done some testing with various NFO and naming arrangements.

crobibero (commented):
I think some tests for the new naming would be helpful, without doing much checking I think you can plug directly into `FindExtrasTests`

SenorSmartyPants:
@crobibero Added one test of cleaning 3 extra names.

crobibero (approved):

PR #9388, opened 2023-02-24
Multiple HLS codec and bitrate fixes

Shadowghost:
**Changes**
* Rebase #8703 on master
* Add automatic DTS bitrate selection based on output channel count
* Calculate fallback bitrate based on output channel count
* Fix encoder check for DTS
* Add encoder check for TrueHD
* Add decoder check for DTS and TrueHD
* Fix condition check in `CanStreamCopyAudio`
* Enforce HLS audio and video codec restrictions on transcoding profiles

**Issues**
Fixes #9270
Fixes #9509
Supersedes #8703

PR #8546, opened 2022-10-13
Add support for OPUS and fixes for FLAC case issue in HLS

nyanmisaka:
`flac` is supported in Shaka but Safari native HLS only supports `fLaC`.

**Changes**
- Add support for OPUS and fixes for FLAC case issue in HLS

**Web changes**
https://github.com/jellyfin/jellyfin-web/pull/4041

PR #9409, opened 2023-02-27
Multiple HLS codec and bitrate fixes (10.8.z)

Shadowghost:
Backports #9388 and #8546

**Changes**
* Enforce HLS audio and video codec restrictions on transcoding profiles
* Properly calculate bitrates for transcoded codecs
* Fix encoder check for DTS
* Add encoder check for TrueHD
* Add decoder check for DTS and TrueHD
* Fix condition check in CanStreamCopyAudio
* Add FLAC/Opus workaround for HLS


MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -805,6 +813,23 @@ private void BuildStreamVideoItem(StreamInfo playlistItem, VideoOptions options,
...
+            if (string.Equals(playlistItem.SubProtocol, "hls", StringComparison.OrdinalIgnoreCase))
+            {
+                var supportedHlsAudioCodecs = new List<string> { "aac", "ac3", "eac3", "mp3" };
+                if (string.Equals(playlistItem.Container, "mp4", StringComparison.OrdinalIgnoreCase))
+                {
+                    // fMP4 supports more codecs than TS
+                    supportedHlsAudioCodecs.Add("alac");
+                    supportedHlsAudioCodecs.Add("flac");
+                    supportedHlsAudioCodecs.Add("opus");
+                    supportedHlsAudioCodecs.Add("vorbis");
```
nyanmisaka, line 827: We currently do not support vorbis in HLS.

https://github.com/jellyfin/jellyfin/blob/edc627fd5b5a5bd19c843dd9e2970b1ebce3fbfd/Jellyfin.Api/Helpers/DynamicHlsHelper.cs#L598-L637

nyanmisaka (approved):
Tested in both TS and fMP4.

crobibero (approved):

joshuaboniface:
So curiously, I've had a user test this and it seems like this alone isn't enough to fix the issue they were seeing. I'm planning to do some more extensive testing over the coming days to see how much I can replicate and confirm. Hold off on merge for a bit longer.

holow29:
Where are you getting supported audio/video codecs? Looking at RFC 8216, I see some information, but can't find reference to FLAC, for example, even though I know it works in HLS in fmp4. I'm mostly wondering because I'm not sure I've come across many codecs that will work in mp4 but won't in HLS fmp4, though more testing is probably merited.

For example, while devising profile for Swiftfin (https://github.com/jellyfin/Swiftfin/pull/519), I included a lot of codecs for HLS fmp4 that VLCKit supports to avoid transcoding for cases like TrueHD audio. With these server-side HLS restrictions, there will be cases like old Blurays with VC-1 or mpeg2video and TrueHD audio that will force a transcode instead of DirectStream. Obviously, I should do more testing to make sure these codecs will actually play in HLS fmp4, and there is an argument to be made that Jellyfin should be adhering to the specs instead of what is possible or non-standard.

nyanmisaka:
> 2.25. The container format for xHE-AAC, Apple Lossless, and FLAC audio MUST be fMP4.

Since HLS is an Apple standard, you should always refer to [their docs](https://developer.apple.com/documentation/http_live_streaming/http_live_streaming_hls_authoring_specification_for_apple_devices).

holow29:
Okay, I don't see opus mentioned in their docs, but it is in this PR. Is your position that Jellyfin server should restrict HLS based on Apple's authoring specs (vs. the standard defined in RFC 8216 or what individual clients can handle/request in their transcoding profile) ?

nyanmisaka:
Shaka-player. https://github.com/jellyfin/jellyfin-web/pull/4041

It can handle almost all the video and audio codecs supported by your browser. OPUS is no exception too.

nyanmisaka:
Use Apple specs as baseline and extend more codecs by the clients we used.
We may use AV1 in HLS in the future, which is certainly not supported by Apple for the time being.

holow29:
It really seems to me like relying on client device profile is a better solution; maybe there is a reason that won't work here that I'm not thinking of. However, I think the server should only be limiting what it can and can't handle given the restriction can actually be placed client-side. The server (and ffmpeg) can handle passing many other formats in HLS and fmp4. Shaka player, VLCkit-based players, etc. can handle things like opus and others that are not defined in Apple's HLS documentation/appendices. Is the idea to only support codecs that all clients can use?

In my example above for Swiftfin, if this PR merges, there will be no way to play old Blurays with VC-1 video or mpeg2video and TrueHD audio without transcoding the video - even though that isn't a limitation of the player itself. (Again, I need to test again because I don't recall, but I am pretty sure I was able to pass VC-1 and mpeg2video in a fmp4 HLS container). I also need to test DTS in fmp4 - which I think I've gotten to work before - because if this merges and Shadowghost's BD/DVD fix merges then Blurays will need to be streamed, and using HLS, any Bluray with DTS audio will automatically be forced to transcode it. Basically, I think this largely limits what will be able to remux over HLS, forcing either DirectStream or transcoding, when it is not a client limitation.

nyanmisaka:
@Shadowghost Better to add a `IsCopyCodec()` check before applying these restrictions.

We only need to ensure that the server can re-encode these codecs. Let the client handle the remuxing cases.

Shadowghost:
@holow29 the reason we are introducing this is that we had an issue where a client accidentially sent a codec profile not compliant with the HLS spec (DTS via HLS-TS) leading to no audio. This is reproducible if you try to send DTS via HLS (VLC on windows can't play those files, MPV can).
Most usecases should be fixed by addind a copycodec flag though since the main problem is if the sorce (audio or video) gets transcoded.

joshuaboniface:
OK definitely seems like we need the whole branch, I was able to replicate the issue still with just this PR. So far I haven't seen it introduce any new issues but you never know. But I'm willing to take the risk since this is a pretty major breakage.

Shadowghost:
I'll rebase the whole master PR (minus the VBR stuff) once we sorted out the remaining things.

Bond-009 (approved):

Jellyfin.Api/Helpers/DynamicHlsHelper.cs
```diff
@@ -710,7 +745,19 @@ private string ReplacePlaylistCodecsField(StringBuilder playlist, StringBuilder
...
+        }
+
+        private string ApplyFlacCaseWorkaround(StreamState state, string srcPlaylist)
+        {
+            if (!string.Equals(state.ActualOutputAudioCodec, "flac", StringComparison.OrdinalIgnoreCase))
+            {
+                return string.Empty;
+            }
+
+            var newPlaylist = srcPlaylist.Replace(",flac\"", ",fLaC\"", StringComparison.Ordinal);
```
crobibero, line 758: This seems to expect that `flac` is always the last audio codec appended
I'm no expert on the playlist code but that seems like a weird limitation to have?

It's also very possible that I am totally not understanding the contents of `srcPlaylist`

Shadowghost: We're directly modifying the HLS playlist which only includes one audio codec, so this is works as intended.

crobibero: For future readers this is what is being modified

```
#EXT-X-STREAM-INF:BANDWIDTH=8000000,VIDEO-RANGE=SDR,CODECS="avc1.640033,flac",RESOLUTION=1920x1080,FRAME-RATE=59.94
```

crobibero (approved):

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -23,6 +23,9 @@ public class StreamBuilder
 
         private readonly ILogger _logger;
         private readonly ITranscoderSupport _transcoderSupport;
+        private static readonly string[] _supportedHlsVideoCodecs = new string[] { "h264", "hevc" };
+        private static readonly string[] _supportedHlsAudioCodecsTs = new string[] { "aac", "ac3", "eac3", "mp3" };
+        private static readonly string[] _supportedHlsAudioCodecsMp4 = new string[] { "aac", "ac3", "eac3", "mp3", "alac", "flac", "opus" };
```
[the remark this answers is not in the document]
Shadowghost: If they work on fmp4 we can do that?

nyanmisaka (approved):

holow29:
Was the decision made to not use `IsCopyCodec()` check then? I think this will affect allowing codecs like vc1 and mpeg2video being passed untouched in HLS.

Shadowghost:
`IsCopyCodec()` can't be used in the places where the check would be needed, this was the best solution I can find. Officially HLS only supports h264 and h265 as video codec, if we need to add more to the list, we can do it in another PR.
Your thoughts on this @nyanmisaka?

nyanmisaka:
@holow29 Feel free to open a PR to add them in these if you support them on the client side.
- `EncodingHelper.cs/shiftVideoCodecs`
- `StreamBuilder.cs/_supportedHlsVideoCodecs`

PR #9410, opened 2023-02-27
Add config option to disable dummy chapter generation

Shadowghost:
Some people don't want dummy chapters.
This makes dummy chapter generation configurable and disables it by default.
It additionally remove the dummy chapter count setting. Either you want dummy chapters or you don't.

### DIFF
--- a/MediaBrowser.Model/Configuration/ServerConfiguration.cs
+++ b/MediaBrowser.Model/Configuration/ServerConfiguration.cs
@@ -243,16 +243,10 @@ namespace MediaBrowser.Model.Configuration
         public bool AllowClientLogUpload { get; set; } = true;
 
         /// <summary>
-        /// Gets or sets the dummy chapters duration in seconds.
+        /// Gets or sets the dummy chapter duration in seconds, use 0 (zero) or less to disable generation alltogether.
         /// </summary>
         /// <value>The dummy chapters duration.</value>
-        public int DummyChapterDuration { get; set; } = 300;
-
-        /// <summary>
-        /// Gets or sets the dummy chapter count.
-        /// </summary>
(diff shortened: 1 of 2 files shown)


Bond-009 (approved):

MediaBrowser.Providers/MediaInfo/FFProbeVideoInfo.cs
```diff
@@ -536,27 +536,25 @@ private ChapterInfo[] CreateDummyChapters(Video video)
...
-                    StartPositionTicks = currentChapterTicks
-                };
+                    chapters[i] = new ChapterInfo
+                    {
+                        StartPositionTicks = currentChapterTicks
+                    };
 
-                currentChapterTicks += dummyChapterDuration;
+                    currentChapterTicks += dummyChapterDuration;
+                }
```
1337joe, line 554:
```suggestion
                }

                return chapters;
```

MediaBrowser.Model/Configuration/ServerConfiguration.cs
```diff
@@ -243,16 +243,15 @@ public ServerConfiguration()
         public bool AllowClientLogUpload { get; set; } = true;
 
         /// <summary>
-        /// Gets or sets the dummy chapters duration in seconds.
+        /// Gets or sets a value indicating whether dummy chapter generation is enabled.
         /// </summary>
-        /// <value>The dummy chapters duration.</value>
-        public int DummyChapterDuration { get; set; } = 300;
+        public bool EnableDummyChapterGeneration { get; set; } = false;
```
cvium, line 248: I still think it's better to use 0 or -1 to disable it. We need fewer options, not more

crobibero (approved):

PR #9461, opened 2023-03-09
Fix MigrateRatingLevels

Shadowghost:
**Changes**
* Rewrite the MigrateRatingLevels migration to calculate the rating level based on the rating string by running it through the LocalizationProvider.


1337joe (approved):

MediaBrowser.Controller/Persistence/IItemRepository.cs
```diff
@@ -17,6 +17,13 @@ namespace MediaBrowser.Controller.Persistence
     /// </summary>
     public interface IItemRepository : IDisposable
     {
+        /// <summary>
+        /// Gets a database connection.
+        /// </summary>
+        /// <param name="readOnly">A value indicating whether the connection is readonly.</param>
+        /// <returns>The connection.</returns>
+        ManagedConnection GetConnection(bool readOnly = false);
```
crobibero, line 25: Note that this would give plugins direct access to the database

Shadowghost: How could we remedy this?

Jellyfin.Server/Migrations/Routines/MigrateRatingLevels.cs
```diff
@@ -0,0 +1,94 @@
...
+
+                // Migrate parental rating strings to new levels
+                _logger.LogInformation("Recalculating parental rating levels based on rating string.");
+                var queryResult = connection.Query("SELECT DISTINCT OfficialRating FROM TypedBaseItems");
+                foreach (var entry in queryResult)
+                {
+                    var ratingString = entry[0].ToString();
+                    if (string.IsNullOrEmpty(ratingString))
+                    {
+                        connection.Execute("UPDATE TypedBaseItems SET InheritedParentalRatingValue = NULL WHERE OfficialRating IS NULL OR OfficialRating=''");
```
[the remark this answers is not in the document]
Shadowghost: It's run a maxmimum of two times (since we select distinct): If we have a null and an empty string entry.

MediaBrowser.Controller/Persistence/IItemRepository.cs
crobibero: I think this extraction has the same effect
At the moment I'm not sure how to fix this but I don't believe direct db access is something we want plugins to have..

Shadowghost: Yeah I reverted it since it breaks too much anyway and does not provide soemthing.

But we still need a way to access the DB within the migrations. The only difference between now and before is that we provide a direct interface for accessing the DB. It's not like plugins were unable to do it before, just required more manual work.

crobibero (approved):

Bond-009:
With #9643 merged the DB isn't in exclusive mode anymore

crobibero (approved):

Bond-009 (approved):

PR #9491, opened 2023-03-15
Add auto-generated asyncapi docs

crobibero:
**Changes**
Adds generation of Websocket documentation.
Viewable at `/api-docs/asyncapi`

**Issues**
None, but requested by client devs

I'm not 100% happy with the generated docs but that may just be a limitation of how our websocket is implemented.

Current spec: [asyncapi.json.txt](https://github.com/jellyfin/jellyfin/files/10974824/asyncapi.json.txt)

### DIFF
--- a/Jellyfin.Server/Extensions/ApiServiceCollectionExtensions.cs
+++ b/Jellyfin.Server/Extensions/ApiServiceCollectionExtensions.cs
@@ -7,6 +7,7 @@ using System.Net.Sockets;
 using System.Reflection;
 using System.Security.Claims;
 using Emby.Server.Implementations;
+using Emby.Server.Implementations.HttpServer;
 using Jellyfin.Api.Auth;
 using Jellyfin.Api.Auth.AnonymousLanAccessPolicy;
 using Jellyfin.Api.Auth.DefaultAuthorizationPolicy;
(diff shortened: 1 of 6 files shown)


Shadowghost (approved):

Jellyfin.Server/Extensions/ApiServiceCollectionExtensions.cs
```diff
@@ -251,6 +257,36 @@ public static IServiceCollection AddJellyfinApiSwagger(this IServiceCollection s
...
         }
 
+        /// <summary>
+        /// Adds AsyncApi to the service collection.
+        /// </summary>
+        /// <param name="serviceCollection">The service collection.</param>
+        /// <returns>The updated service collection.</returns>
+        public static IServiceCollection AddAsyncApi(this IServiceCollection serviceCollection)
+        {
+            const string AsyncApiTitle = "Jellyfin Websocket Documentation";
```
nielsvanvelzen, line 267:
```suggestion
            const string AsyncApiTitle = "Jellyfin WebSocket Documentation";
```

nielsvanvelzen (changes requested):

crobibero:
- Missing `x-jellyfin-version` (openapi spec has it in case we decide to use separate jellyfin and api versions in the future)

I don't see a way to add this for the asyncapi docs...

- The server protocol is set to "wss" while using our HTTP port, should probably be "ws" or https port

fixed

- The generated server info is called "WebSocketConnection", should probably be something like "default" to avoid confusion with the channel name

fixed

- The channel name is called "WebSocketConnection", this must be "/socket" because it indicated the URL.

fixed

- No authentication specified on server (although the OpenAPI one isn't valid either)

I tried to fix it... but the library doesn't follow the spec

- Message schemas are a composite of webSocketMessage & an object containing their custom properties, which is fine except that both contain the same definition for `messageType`. I'd expect the specific message to use a const value. (so `activityLogEntryMessage` would force the value to `ActivityLogEntry`).

I originally tried to do this but couldn't figure out a way to do it.



---

Might be better to just properly shove these models into the openapi spec?

nielsvanvelzen:
> Might be better to just properly shove these models into the openapi spec?

I think this is the easiest solution for the Kotlin SDK but I'm not sure if this is the preferred way for the other clients/SDK's.

crobibero:
Closing in favor of https://github.com/jellyfin/jellyfin/pull/9682

PR #8004, opened 2022-06-24
Episode multiple versions

SenorSmartyPants:
**Changes**
Enable multiple versions for episodes. Episode suffixes work or episode may be placed in it's own directory. Will not collapse episodes in series directory. Unit tests included.

```
/TV/Dexter/Dexter - S01E01/Dexter - S01E01 - One.mkv
/TV/Dexter/Dexter - S01E01/Dexter - S01E01 - Two.mkv
/TV/Dexter/Dexter - S01E02 - Another.mkv
/TV/Dexter/Dexter - S01E03 - Again.mkv
/TV/Dexter/Season 2/Dexter - S02E01/Dexter - S02E01 - First.mkv
/TV/Dexter/Season 2/Dexter - S02E01/Dexter - S02E01 - Second.mkv
/TV/Dexter/Season 2/Dexter - S02E02 - More.mkv
```

JF will show 3 episodes in season 1. 2 episodes in season 2.

PR #9496, opened 2023-03-15
Fix file stacking

mammo0:
As discussed with @Shadowghost in #8004, there is a stacking rule (`VideoFileStackingRules`) in the file `Emby.Naming/Common/NamingOptions.cs` that makes no sense. It leads to a behavior that is not documented.

For example the files

- `300a.mkv`,
- `300b.mkv` and
- `300c.mkv`

will be treated as parts of the same movie. But according to the Shows documentation (https://jellyfin.org/docs/general/server/media/shows/#episodes-split-across-multiple-parts) _- don't know why this isn't also documented in the movie section -_ a `<parttype>` is required.
So normally the files should to be named like this:

- `300 parta.mkv`,
- `300 partb.mkv` and
- `300 partc.mkv`

I removed the corresponding rule and adjusted the affected test cases. Now the above naming (`300a`, ...) does no longer work. Instead the files have to be named like `300 parta`, etc. as it is documented.

### DIFF
--- a/tests/Jellyfin.Naming.Tests/Video/VideoListResolverTests.cs
+++ b/tests/Jellyfin.Naming.Tests/Video/VideoListResolverTests.cs
@@ -332,7 +332,7 @@ namespace Jellyfin.Naming.Tests.Video
                 files.Select(i => VideoResolver.Resolve(i, false, _namingOptions)).OfType<VideoFileInfo>().ToList(),
                 _namingOptions).ToList();
 
-            // The result should contain to individual movies
+            // The result should contain two individual movies
             // Version grouping should not work here, because the files are not in a directory with the name 'Four Sisters and a Wedding'
             Assert.Equal(2, result.Count);
         }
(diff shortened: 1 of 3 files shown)


tests/Jellyfin.Naming.Tests/Video/VideoListResolverTests.cs
```diff
@@ -332,7 +332,9 @@ public void TestFourSisters()
                 files.Select(i => VideoResolver.Resolve(i, false, _namingOptions)).OfType<VideoFileInfo>().ToList(),
                 _namingOptions).ToList();
 
-            Assert.Single(result);
+            // The result should contain to individual movies
```
Shadowghost, line 335:
```suggestion
            // The result should contain two individual movies
```

mammo0:
> I added the stacking documentation to shows [...]

You mean, you've now added it to the movies? Because the Shows documentation page had already that part.

Shadowghost:
He was the person who originally added the documentation.

We might want to move the stacking an versioning dosumentation to its own page though once the episdoe versioning PR is in.

Bond-009 (approved):

Issue #1946, opened 2019-10-27
[Privacy issue] User permissions don't work - Any user can play any media


omyno:
**Describe the bug**
A user without access to libraries can still access media in all libraries via search.

**To Reproduce**

1. Create user, don't grant permission to any libraries. 
2. Use search function to find media from other libraries
3. Find and play media that the user is not supposed to find.

**System (please complete the following information):**
 - OS: Docker
 - Browser: any
 - Jellyfin Version: https://hub.docker.com/layers/jellyfin/jellyfin/latest/images/sha256-af09b0edf3f5d0f6aecbe7745ff58097bb86b2c925f8c15488c3b47e6f9a5480
 - Reverse proxy: nginx

KeepBotting:
Confirmed.

OS: Docker on Void Linux
Browser: any
Jellyfin version: 10.4.0
Reverse proxy: yes, nginx

Z3ROGRAV1TYx:
Not having this exact issue personally, but would like to reference #2207 because I believe it's related.
OS: Windows 10
Jellyfin: 10.4.3
Reverse Proxy: Caddy

gnattu:
When I am testing this, I found that if we give the user access to at least one library with contents, they he will no longer able to search for media that he does not have access to, nor able to see the suggestions from other libraries. Could you confirm this?

sum01:
I just tested this. I have a user with access to all but 1 library. Tried to search for content in that disabled library and it doesn't show up. Tested with the web interface on server v10.6.4.

PR #9536, opened 2023-03-25
Implement check to hide all libraries when user has no access

elmuffo:
**Changes**
Implement the same check used at other `GetTopParentIdsForQuery` call sites so a user without any library access cannot bypass it by searching for media

**Issues**
Fixes [#1946](https://github.com/jellyfin/jellyfin/issues/1946)


Shadowghost (approved):

Bond-009 (approved):

crobibero (commented):
Would it be better to always append `Guid.Empty`?

```c#
query.TopParentIds = userViews.SelectMany(i => GetTopParentIdsForQuery(i, user)).Append(Guid.Empty).ToArray();
```

crobibero (approved):

PR #9551, opened 2023-03-28
Fix multiple NFO backdrops

### LINKED ISSUE #9456: [Issue]: NFO defined backdrops result in addition duplicates after each refresh all images
### Please describe your bug

I have an NFO for a TV season. I'm setting a backdrop using the NFO, remote image providers have no backdrops for this season. Every time I I run refresh metadata (replace all images) I get another backdrop of the same image from the NFO. I have no local images in the season directory, and JF does not create any there. Sounds like #7836 but that is reportedly fixed. I just tested against release-10.8.z branch

I think the root of the problem is here
https://github.com/jellyfin/jellyfin/blob/6821a2ab358761282a0030c42c837b39bad089e1/MediaBrowser.Providers/Manager/MetadataService.cs#L702
localItem.Images never has any images. Images are being added directly to item a few lines above. But I can't figure out what a fix should be

```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<season>
	<thumb aspect="poster">
		https://thetvdb.com/banners/seasons/72306-2-5.jpg
	</thumb>
	<fanart>
		<thumb>
			https://artworks.thetvdb.com/banners/fanart/original/72306-8.jpg
		</thumb>
	</fanart> 
</season>
```

### Jellyfin Version

10.8.0

### if other:

10.8.z release branch compiled locally for debugging

### Environment

```markdown
- OS:
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

image

### Code of Conduct

SenorSmartyPants:
Fixes #9456 

Force push to remove fix for separate issue.

### DIFF
--- a/MediaBrowser.Providers/Manager/MetadataService.cs
+++ b/MediaBrowser.Providers/Manager/MetadataService.cs
@@ -688,6 +688,12 @@ namespace MediaBrowser.Providers.Manager
 
                     if (localItem.HasMetadata)
                     {
+                        var oldBackdropImages = Array.Empty<ItemImageInfo>();
+                        if (options.IsReplacingImage(ImageType.Backdrop))
+                        {
+                            oldBackdropImages = item.GetImages(ImageType.Backdrop).ToArray();
+                        }
+
                         foreach (var remoteImage in localItem.RemoteImages)
                         {
                             try
@@ -708,6 +714,12 @@ namespace MediaBrowser.Providers.Manager
                             {
                                 Logger.LogError(ex, "Could not save {ImageType} image: {Url}", Enum.GetName(remoteImage.Type), remoteImage.Url);
                             }
+                        }
+
+                        // only delete existing multi-images if new ones were added
+                        if (oldBackdropImages.Length > 0 && oldBackdropImages.Length < item.GetImages(ImageType.Backdrop).Count())
+                        {
+                            imageService.PruneImages(item, oldBackdropImages);
                         }
 
                         if (foundImageTypes.Count > 0)
--- a/MediaBrowser.Providers/Manager/ItemImageProvider.cs
+++ b/MediaBrowser.Providers/Manager/ItemImageProvider.cs
@@ -357,7 +357,12 @@ namespace MediaBrowser.Providers.Manager
             return true;
         }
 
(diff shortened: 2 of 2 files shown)


Shadowghost:
Isn't this a general logic error? If you run `replace images` it should replace all images regardless if the resulting amount of images is 0. None of the existing images should be retained imo.

Shadowghost:
Okay, that's a valid concern.

Do the changes in this PR get rid of the duplicates if you run an update with replace all images?

SenorSmartyPants:
I believe it does, just been a while since I tested it.

Shadowghost:
Mind rebasing on master so I can test?

SenorSmartyPants:
I'll do that soon (<24 hours), but can't right now.

MediaBrowser.Providers/Manager/MetadataService.cs
```diff
@@ -713,6 +719,12 @@ public bool CanRefreshPrimary(Type type)
                             }
                         }
 
+                        // only delete existing multi-images if new ones were added
+                        if (oldBackdropImages.Length > 0 && oldBackdropImages.Length < item.GetImages(ImageType.Backdrop).Count())
```
Shadowghost, line 723: Wouldn't this be a problem if you adobt your referenced backdrops to an amount lower than the current one? In that case the changed backdrops would be dropped instead of applied.
Ideally we'd have a flag on images for deciding if they originate from a provider or were added by a user, so we can remove all from the providers but keep the manually added user images.

Shadowghost:
Did you fin the time to look into it again?

SenorSmartyPants:
No, been dealing with a lot IRL. But I will get back to this. Are there any 10.9 related deadlines I should be aware of?

Bond-009:
@SenorSmartyPants Not at the moment, still some stuff that has to be sorted out first.

SenorSmartyPants:
Rebase, fix merge conflict and force push.

JPVenson (approved):
LGTM

theguymadmax:
The linked issue can no longer be reproduced, closing as stale.

PR #9554, opened 2023-03-28
Trickplay preview images

nicknsy:
This adds most functionality from https://github.com/nicknsy/jellyscrub / https://github.com/iwalton3/jellyscrub to allow for a consistent way to generate preview images for media server-side and then access them in client-side apps through an authenticated API endpoint.

Since this change is on the larger side and modifies the HLS endpoint, media encoder, etc., this pull request is mainly for code review and other feedback on implementation that may need to be changed.

Currently, the design is as follows:
1. The TrickplayManager manages the generation and retrieval of trickplay images.
    - Trickplay manifest data is stored in the sqlite database like chapter images. Width+Id is primary key.
    - Trickplay images are generated by the MediaEncoder and stored in the metadata folder of the item.
 2. Generation is triggered from either the scheduled task, adding to library, or manually refreshing item metadata with "replace images" checked. Like chapters, must be enabled under library options to be generated.
 3. Trickplay manifest data (width, height, tile count, interval) is included when the web client requests item info and is opportunistically used to request trickplay images, taking precedence over chapter images.
 4. If an HLS stream, a trickplay playlist is generated and included based on manifest data.
 5. Images are retrieved using the ```Videos/{itemId}/Trickplay/{width}/{index}.jpg``` endpoint (modeled after /Videos/../Subtitles) as the mouse is scrubbed over the timeline. I chose not to use the /Images/ endpoint as it only passes image type and index which would make handling different trickplay resolutions more difficult or would require modifying that endpoint. 

If anyone is interested in helping test or leaving any feedback on the code/design please use the jellyfin-web fork [here](https://github.com/nicknsy/jellyfin-web/tree/trickplay). Configuration is done under Settings / Playback / Trickplay and must be enabled per library. Also, make sure to set the ffmpeg path to jellyfin-ffmpeg if using hw acceleration. Thanks. 

**Currently unimplemented**
- Saving trickplay to local media folders -- wanted to match chapter image functionality.
- Support for different trickplay resolutions are not implemented in the jellyfin-web fork I made for testing.
- No way to delete trickplay images, only regenerate.

**Currently untested**
- Whether or not the HLS playlist modifications will actually produce working previews on a compatible video player.

**Other discussions**
https://github.com/nicknsy/jellyscrub/issues/83
https://github.com/jellyfin/jellyfin-meta/discussions/33

**Example trickplay manifest**
```
"Trickplay": {
  "f37b3cd33269dff034ee51d4cda2c0e4": {
    "320": {
      "Width": 320,
      "Height": 180,
      "TileWidth": 10,
      "TileHeight": 10,
      "ThumbnailCount": 15,
      "Interval": 10000,
      "Bandwidth": 1161
    }
  },
  "244e7ede3885a7249de6308fd4d67816": {
    "320": {
      "Width": 320,
      "Height": 180,
      "TileWidth": 10,
      "TileHeight": 10,
      "ThumbnailCount": 15,
      "Interval": 10000,
      "Bandwidth": 1161
    }
  }
}
```


MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
```diff
@@ -772,6 +775,195 @@ private async Task<string> ExtractImageInternal(string inputPath, string contain
...
+            if (string.IsNullOrWhiteSpace(inputArg))
+            {
+                throw new InvalidOperationException("Empty or invalid input argument.");
+            }
+
+            // Output arguments
+            string fps = "fps=1/" + interval.TotalSeconds.ToString(CultureInfo.InvariantCulture);
+            if (string.IsNullOrWhiteSpace(filterParam))
+            {
+                filterParam = "-vf \"" + fps + "\"";
```
nyanmisaka, line 856: Better to move the `fps` filter to `GetVideoProcessingFilterParam()` and set the fps value in the request.
We may need reuse this filter to to limit the video frame rate.

MediaBrowser.Model/Configuration/EncodingOptions.cs
```diff
@@ -237,6 +238,11 @@ public EncodingOptions()
     /// </summary>
     public bool AllowHevcEncoding { get; set; }
 
+    /// <summary>
+    /// Gets or sets a value indicating whether MJPEG encoding is enabled.
+    /// </summary>
+    public bool AllowMjpegEncoding { get; set; }
```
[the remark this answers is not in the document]
nicknsy: Also, I still think this option could be used beyond trickplay--such as using hw encoding or hw tonemapping with extracted images like chapters or the preview image. In this case maybe it would be better to keep this is EncodingOptions but make the variable name clearer?

MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
nicknsy: I've made the change, but realized there is a possibility for both the ```-r``` option and ```fps``` option to be set when ```GetVideoQualityParam``` is used in conjunction with ```GetVideoProcessingFilterParam```. I'm not entirely sure if the simultaneous use of these options would cause ```-r``` to further lower the rate set by the fps filter, or if one would take precedence over the other. 

As far as code consistency and ffmpeg functionality, what are your thoughts? Would removing the  ```-r``` option from ```GetVideoQualityParam``` and leaving it solely up to ```GetVideoProcessingFilterParam``` lead to any issues? It seems like ```GetVideoQualityParam``` is always used after the input argument so I imagine fps should work to lower the output video fps as well.

Jellyfin.Api/Controllers/TrickplayController.cs
```diff
@@ -0,0 +1,178 @@
...
+        var path = _trickplayManager.GetTrickplayTilePath(item, width, index);
+        if (System.IO.File.Exists(path))
+        {
+            return PhysicalFile(path, MediaTypeNames.Image.Jpeg);
+        }
+
+        return NotFound();
+    }
+
+    private ActionResult GetTrickplayPlaylistInternal(int width, string mediaSourceId)
```
JPVenson, line 109: Maybe it would be a good idea to put that into its own service?

nicknsy: I wrote the playlist logic as part of the controller to mimic the current way external playlists seem to be structured, like the SubtitleController:
https://github.com/jellyfin/jellyfin/blob/bb5bf0277ae33fd4668e00bd2f1a85e89aef466a/Jellyfin.Api/Controllers/SubtitleController.cs#L324-L329

I'm not sure if the code for generating playlists does or doesn't have reusability outside of HLS, but if we did want it to be generated in a publicly accessible way the [DynamicHlsPlaylistGenerator](https://github.com/jellyfin/jellyfin/blob/master/src/Jellyfin.MediaEncoding.Hls/Playlist/DynamicHlsPlaylistGenerator.cs) is the only place I can think to put it.

JPVenson: I am not so much interested in reusability but more in another extension point for a plugin the adapt or rewrite the playlist generation. That would make a good point to join another playlist provider into.

JPVenson: Extending `DynamicHlsPlaylistGenerator` might be the right place as an alternative to its own service though.

nicknsy: It looks like trickplay might be outside the scope of ```DynamicHlsPlaylistGenerator``` given it's description: "Generator for dynamic HLS playlists where the segment lengths aren't known in advance." Given the relative scatter of Hls/Streaming code (e.g. DynamicHlsPlaylistGenerator, DynamicHlsHelper, StreamingHelpers, HlsHelpers, the different playlist controllers) and having just started working with the codebase, I'm going to stick with the ```SubtitleController```-type logic for now as it seems to be the closest analog.

If/when there is a general consensus on where specifically the playlist code should be moved or if it should be further abstracted altogether, it shouldn't be difficult to accommodate that.

JPVenson: Non-Action methods in controllers are possible but also not the cleanest thing (separation of concerns). I would still say an service for that kind of logic would be best as such business logic is (mostly) located inside services. I see not harm in creating a new interface and implementation alongside the `ITrickplayManager` like a `ITrickplayPlaylistGenerator`

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -152,6 +152,36 @@ private string GetH264OrH265Encoder(string defaultEncoder, string hwEncoder, Enc
...
+            var defaultEncoder = "mjpeg";
+
+            if (state.VideoType == VideoType.VideoFile)
+            {
+                var hwType = encodingOptions.HardwareAccelerationType;
+
+                var codecMap = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
+                {
+                    { "vaapi",                  defaultEncoder + "_vaapi" },
+                    { "qsv",                defaultEncoder + "_qsv" }
```
crobibero, line 166:
```suggestion
                    { "vaapi", defaultEncoder + "_vaapi" },
                    { "qsv", defaultEncoder + "_qsv" }
```

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -152,6 +152,36 @@ private string GetH264OrH265Encoder(string defaultEncoder, string hwEncoder, Enc
...
+                    && encodingOptions.EnableHardwareEncoding
+                    && codecMap.ContainsKey(hwType))
+                {
+                    var preferredEncoder = codecMap[hwType];
+
+                    if (_mediaEncoder.SupportsEncoder(preferredEncoder))
+                    {
+                        return preferredEncoder;
+                    }
+                }
```
crobibero, line 179:
```suggestion
                if (!string.IsNullOrEmpty(hwType)
                    && encodingOptions.EnableHardwareEncoding
                    && codecMap.TryGetValue(hwType, out var preferredEncoder)
                    && _mediaEncoder.SupportsEncoder(preferredEncoder))
                {
                    return preferredEncoder;
                }
```

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -152,6 +152,36 @@ private string GetH264OrH265Encoder(string defaultEncoder, string hwEncoder, Enc
...
 
+        private string GetMjpegEncoder(EncodingJobInfo state, EncodingOptions encodingOptions)
+        {
+            var defaultEncoder = "mjpeg";
+
+            if (state.VideoType == VideoType.VideoFile)
+            {
+                var hwType = encodingOptions.HardwareAccelerationType;
+
+                var codecMap = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
```
crobibero, line 163: this dictionary should be moved to `static readonly` at the class level

MediaBrowser.Controller/Trickplay/ITrickplayManager.cs
```diff
@@ -0,0 +1,54 @@
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using MediaBrowser.Controller.Entities;
+using MediaBrowser.Model.Entities;
+
+namespace MediaBrowser.Controller.Trickplay
```
crobibero, line 8: Use file-scoped namespace

MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
```diff
@@ -772,6 +775,184 @@ private async Task<string> ExtractImageInternal(string inputPath, string contain
...
+            TimeSpan interval,
+            bool allowHwAccel,
+            int? threads,
+            int? qualityScale,
+            ProcessPriorityClass? priority,
+            EncodingHelper encodingHelper,
+            CancellationToken cancellationToken)
+        {
+            var options = allowHwAccel ? _configurationManager.GetEncodingOptions() : new EncodingOptions();
+            threads = threads ?? _threads;
```
crobibero, line 793:
```suggestion
            threads ??= _threads;
```

MediaBrowser.Model/Configuration/TrickplayOptions.cs
```diff
@@ -0,0 +1,61 @@
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace MediaBrowser.Model.Configuration
```
crobibero, line 4: Use file-scoped namespace

MediaBrowser.Model/Entities/TrickplayTilesInfo.cs
```diff
@@ -0,0 +1,50 @@
+namespace MediaBrowser.Model.Entities
```
crobibero, line 1: use file-scoped namespace

MediaBrowser.Providers/Trickplay/TrickplayImagesTask.cs
```diff
@@ -0,0 +1,109 @@
...
+        public async Task ExecuteAsync(IProgress<double> progress, CancellationToken cancellationToken)
+        {
+            var items = _libraryManager.GetItemList(new InternalItemsQuery
+            {
+                MediaTypes = new[] { MediaType.Video },
+                IsVirtualItem = false,
+                IsFolder = false,
+                Recursive = true
+            }).OfType<Video>().ToList();
+
```
crobibero, line 81: This should be paginated. No reason to pull all videos into memory at once.

MediaBrowser.Providers/Trickplay/TrickplayManager.cs
```diff
@@ -0,0 +1,386 @@
...
+using MediaBrowser.Controller.MediaEncoding;
+using MediaBrowser.Controller.Persistence;
+using MediaBrowser.Controller.Trickplay;
+using MediaBrowser.Model.Configuration;
+using MediaBrowser.Model.Entities;
+using MediaBrowser.Model.IO;
+using Microsoft.Extensions.Logging;
+using SkiaSharp;
+
+namespace MediaBrowser.Providers.Trickplay
```
crobibero, line 20: file-scoped namespace

MediaBrowser.Providers/Trickplay/TrickplayProvider.cs
```diff
@@ -0,0 +1,127 @@
...
+using MediaBrowser.Controller.Entities;
+using MediaBrowser.Controller.Entities.Movies;
+using MediaBrowser.Controller.Entities.TV;
+using MediaBrowser.Controller.Library;
+using MediaBrowser.Controller.Providers;
+using MediaBrowser.Controller.Trickplay;
+using MediaBrowser.Model.Configuration;
+using Microsoft.Extensions.Logging;
+
+namespace MediaBrowser.Providers.Trickplay
```
crobibero, line 13: file-scoped namespace

MediaBrowser.Controller/Trickplay/ITrickplayManager.cs
```diff
@@ -0,0 +1,54 @@
...
+    public interface ITrickplayManager
+    {
+        /// <summary>
+        /// Generate or replace trickplay data.
+        /// </summary>
+        /// <param name="video">The video.</param>
+        /// <param name="replace">Whether or not existing data should be replaced.</param>
+        /// <param name="cancellationToken">CancellationToken to use for operation.</param>
+        /// <returns>Task.</returns>
+        Task RefreshTrickplayData(Video video, bool replace, CancellationToken cancellationToken);
```
crobibero, line 22:
```suggestion
        Task RefreshTrickplayDataAsync(Video video, bool replace, CancellationToken cancellationToken);
```

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -4717,6 +4752,15 @@ public static string GetHwTonemapFilter(EncodingOptions options, string hwTonema
...
             overlayFilters?.RemoveAll(filter => string.IsNullOrEmpty(filter));
 
+            var framerate = GetFramerateParam(state);
+            if (framerate.HasValue)
+            {
+                mainFilters.Insert(0, string.Format(
+                    CultureInfo.InvariantCulture,
+                    "fps={0}",
+                    framerate.Value));
+            }
```
nyanmisaka, line 4762: The actual fps can affect the key frame arguments. We may need some changes in `GetHlsVideoKeyFrameArguments()`.

https://github.com/jellyfin/jellyfin/blob/af5184f84170bde0d66d89be2b953f950c7dfc86/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs#L1401-L1413

MediaBrowser.Providers/Trickplay/TrickplayImagesTask.cs
nicknsy: I don't entirely understand how pagination would work with this as I had assumed all videos/library data would be loaded and stay in memory server-side when the server starts, or at least as clients and other tasks start to request them.

For reference, this is the code from the chapter images task (https://github.com/jellyfin/jellyfin/blob/08aac575290d0b94c0c4b7e6af6167867dc93efa/Emby.Server.Implementations/ScheduledTasks/Tasks/ChapterImagesTask.cs#L90-L106) that I based the trickplay task off of. Do you have any examples of pagination for item queries as it relates to server-sided tasks that I could look at as a starting point?

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
nicknsy: As mentioned in the original review, this could also potentially affect ```GetVideoQualityParam``` which uses the request Framerate/MaxFramerate to set the -r ffmpeg argument.

MediaBrowser.Model/Configuration/TrickplayOptions.cs
```diff
@@ -0,0 +1,60 @@
...
+
+    /// <summary>
+    /// Gets or sets the jpeg quality to use for image tiles.
+    /// </summary>
+    public int JpegQuality { get; set; } = 90;
+
+    /// <summary>
+    /// Gets or sets the number of threads to be used by ffmpeg.
+    /// </summary>
+    public int ProcessThreads { get; set; } = 0;
```
nicknsy, line 59: What are your guys' thoughts on changing the default thread count to 1 for trickplay generation? I think it would be better to err on the side of caution and prevent people from turning this on and then not knowing why their cpu is at high usage while trickplay images are generating, which can potentially take days or weeks depending on library size and processing power.

Jellyfin.Api/Controllers/TrickplayController.cs
```diff
@@ -0,0 +1,178 @@
...
+    {
+        if (_httpContextAccessor.HttpContext is null)
+        {
+            throw new ResourceNotFoundException(nameof(_httpContextAccessor.HttpContext));
+        }
+
+        var tilesResolutions = _trickplayManager.GetTilesResolutions(Guid.Parse(mediaSourceId));
+        if (tilesResolutions is not null && tilesResolutions.ContainsKey(width))
```
crobibero, line 117:
```suggestion
        if (tilesResolutions is not null && tilesResolutions.TryGetValue(width, out var tilesInfo))
```

And remove L120

Jellyfin.Api/Controllers/TrickplayController.cs
```diff
@@ -0,0 +1,178 @@
...
+    /// <returns>A <see cref="FileResult"/> containing the trickplay tiles image.</returns>
+    [HttpGet("Videos/{itemId}/Trickplay/{width}/{index}.jpg")]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    [ProducesImageFile]
+    public ActionResult GetTrickplayHlsPlaylist(
+        [FromRoute, Required] Guid itemId,
+        [FromRoute, Required] int width,
+        [FromRoute, Required] int index,
+        [FromQuery] string? mediaSourceId)
```
crobibero, line 92:
```suggestion
        [FromQuery] Guid? mediaSourceId)
```

And trickle down changes.

Jellyfin.Api/Controllers/TrickplayController.cs
```diff
@@ -0,0 +1,178 @@
...
+                        infDuration = tileDuration * tilesPerGrid;
+                    }
+
+                    var url = string.Format(
+                        CultureInfo.InvariantCulture,
+                        urlFormat,
+                        width.ToString(CultureInfo.InvariantCulture),
+                        i.ToString(CultureInfo.InvariantCulture),
+                        mediaSourceId,
+                        _httpContextAccessor.HttpContext.User.GetToken());
```
crobibero, line 157:
```suggestion
                        User.GetToken());
```

Jellyfin.Api/Controllers/TrickplayController.cs
```diff
@@ -0,0 +1,178 @@
...
+
+                    // EXTINF
+                    builder.Append("#EXTINF:").Append(string.Format(CultureInfo.InvariantCulture, decimalFormat, infDuration))
+                        .AppendLine(",");
```
crobibero, line 161:
```suggestion
                    builder
                        .Append("#EXTINF:")
                        .Append(string.Format(CultureInfo.InvariantCulture, decimalFormat, infDuration))
                        .AppendLine(",");
```

just changing formatting

Jellyfin.Api/Controllers/TrickplayController.cs
```diff
@@ -0,0 +1,178 @@
...
+
+                    // EXT-X-TILES
+                    builder.Append("#EXT-X-TILES:RESOLUTION=").Append(resolution).Append(",LAYOUT=").Append(layout).Append(",DURATION=")
+                        .AppendLine(string.Format(CultureInfo.InvariantCulture, decimalFormat, tileDuration));
```
crobibero, line 165:
```suggestion
                    builder
                        .Append("#EXT-X-TILES:RESOLUTION=")
                        .Append(resolution)
                        .Append(",LAYOUT=")
                        .Append(layout)
                        .Append(",DURATION=")
                        .AppendLine(string.Format(CultureInfo.InvariantCulture, decimalFormat, tileDuration));
```

Jellyfin.Api/Controllers/TrickplayController.cs
```diff
@@ -0,0 +1,178 @@
...
+
+            if (tilesInfo.TileCount > 0)
+            {
+                const string urlFormat = "Trickplay/{0}/{1}.jpg?MediaSourceId={2}&api_key={3}";
+                const string decimalFormat = "{0:0.###}";
+
+                var resolution = tilesInfo.Width.ToString(CultureInfo.InvariantCulture) + "x" + tilesInfo.Height.ToString(CultureInfo.InvariantCulture);
+                var layout = tilesInfo.TileWidth.ToString(CultureInfo.InvariantCulture) + "x" + tilesInfo.TileHeight.ToString(CultureInfo.InvariantCulture);
+                var tilesPerGrid = tilesInfo.TileWidth * tilesInfo.TileHeight;
+                var tileDuration = (decimal)tilesInfo.Interval / 1000;
```
crobibero, line 130:
```suggestion
                var tileDuration = tilesInfo.Interval / 1000m;
```

Jellyfin.Api/Controllers/TrickplayController.cs
```diff
@@ -0,0 +1,178 @@
...
+            var builder = new StringBuilder(128);
+            var tilesInfo = tilesResolutions[width];
+
+            if (tilesInfo.TileCount > 0)
+            {
+                const string urlFormat = "Trickplay/{0}/{1}.jpg?MediaSourceId={2}&api_key={3}";
+                const string decimalFormat = "{0:0.###}";
+
+                var resolution = tilesInfo.Width.ToString(CultureInfo.InvariantCulture) + "x" + tilesInfo.Height.ToString(CultureInfo.InvariantCulture);
+                var layout = tilesInfo.TileWidth.ToString(CultureInfo.InvariantCulture) + "x" + tilesInfo.TileHeight.ToString(CultureInfo.InvariantCulture);
```
crobibero, line 128:
```suggestion
                var resolution = $"{tilesInfo.Width}x{tilesInfo.Height}";
                var layout = $"{tilesInfo.TileWidth}x{tilesInfo.TileHeight}";
```

Jellyfin.Api/Controllers/TrickplayController.cs
crobibero: I think this function can live inside of TrickPlayManager? I agree that it doesn't belong in the controller

crobibero (commented):
Just a starter review for now

MediaBrowser.Providers/Trickplay/TrickplayImagesTask.cs
```diff
@@ -0,0 +1,108 @@
...
+    /// <inheritdoc />
+    public async Task ExecuteAsync(IProgress<double> progress, CancellationToken cancellationToken)
+    {
+        var items = _libraryManager.GetItemList(new InternalItemsQuery
+        {
+            MediaTypes = new[] { MediaType.Video },
+            IsVirtualItem = false,
+            IsFolder = false,
+            Recursive = true
+        }).OfType<Video>().ToList();
```
crobibero, line 80: This should be paginated, no reason to pull everything out of the database at once

MediaBrowser.Providers/Trickplay/TrickplayImagesTask.cs
```diff
@@ -0,0 +1,108 @@
...
+                cancellationToken.ThrowIfCancellationRequested();
+                await _trickplayManager.RefreshTrickplayDataAsync(item, false, cancellationToken).ConfigureAwait(false);
+            }
+            catch (OperationCanceledException)
+            {
+                break;
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError("Error creating trickplay files for {ItemName}: {Msg}", item.Name, ex);
```
crobibero, line 97:
```suggestion
                _logger.LogError(ex, "Error creating trickplay files for {ItemName}", item.Name);
```

MediaBrowser.Providers/Trickplay/TrickplayManager.cs
```diff
@@ -0,0 +1,385 @@
...
+        {
+            return;
+        }
+
+        var imgTempDir = string.Empty;
+        var outputDir = GetTrickplayDirectory(video, width);
+
+        try
+        {
+            await _resourcePool.WaitAsync(cancellationToken).ConfigureAwait(false);
```
crobibero, line 100: This should be outside of the `try`

MediaBrowser.Providers/Trickplay/TrickplayManager.cs
```diff
@@ -0,0 +1,385 @@
...
+                options.ProcessPriority,
+                _encodingHelper,
+                cancellationToken).ConfigureAwait(false);
+
+            if (string.IsNullOrEmpty(imgTempDir) || !Directory.Exists(imgTempDir))
+            {
+                throw new InvalidOperationException("Null or invalid directory from media encoder.");
+            }
+
+            var images = _fileSystem.GetFiles(imgTempDir, new string[] { ".jpg" }, false, false)
```
crobibero, line 142: the possible extensions should be pulled into a `static readonly`

MediaBrowser.Providers/Trickplay/TrickplayManager.cs
```diff
@@ -0,0 +1,385 @@
...
+                .Where(img => string.Equals(img.Extension, ".jpg", StringComparison.Ordinal))
```
crobibero, line 143: The `GetFiles` call already does this filter

MediaBrowser.Providers/Trickplay/TrickplayManager.cs
```diff
@@ -0,0 +1,385 @@
...
+        {
+            Width = width,
+            Interval = options.Interval,
+            TileWidth = options.TileWidth,
+            TileHeight = options.TileHeight,
+            TileCount = 0,
+            Bandwidth = 0
+        };
+
+        var firstImg = SKBitmap.Decode(images[0].FullName);
```
crobibero, line 207: Should use `IImageEncoder` instead of referencing Skia directly

MediaBrowser.Providers/Trickplay/TrickplayProvider.cs
```diff
@@ -0,0 +1,126 @@
...
+public class TrickplayProvider : ICustomMetadataProvider<Episode>,
+    ICustomMetadataProvider<MusicVideo>,
+    ICustomMetadataProvider<Movie>,
+    ICustomMetadataProvider<Trailer>,
+    ICustomMetadataProvider<Video>,
+    IHasItemChangeMonitor,
+    IHasOrder,
+    IForcedProvider
+{
+    private readonly ILogger<TrickplayProvider> _logger;
```
crobibero, line 28: Logger is unused

MediaBrowser.Providers/Trickplay/TrickplayProvider.cs
```diff
@@ -0,0 +1,126 @@
...
+            return ItemUpdateType.None;
+        }
+
+        if (_config.Configuration.TrickplayOptions.ScanBehavior == TrickplayScanBehavior.Blocking)
+        {
+            await _trickplayManager.RefreshTrickplayDataAsync(video, replace, cancellationToken).ConfigureAwait(false);
+        }
+        else
+        {
+            _ = _trickplayManager.RefreshTrickplayDataAsync(video, replace, cancellationToken).ConfigureAwait(false);
```
crobibero, line 120: I'm not a fan of having a nonblocking call for this

MediaBrowser.Model/Configuration/TrickplayOptions.cs
JPVenson: That is true, but on the other hand just 3 days ago i discovered that i can enable hardware acceleration for the jellyscrub task so now i enjoy some silence in my room again so that might be an better default value.

I think -1 (or use what is configured in transcoding) would be better. It is better to explain that the whole task is an resource intensive one then to explain why the new module is so much slower then the plugin one.

MediaBrowser.Providers/Trickplay/TrickplayImagesTask.cs
crobibero: The subtitle extract plugin has pagination
https://github.com/jellyfin/jellyfin-plugin-subtitleextract/blob/master/Jellyfin.Plugin.SubtitleExtract/Tasks/ExtractSubtitlesTask.cs

MediaBrowser.Providers/Trickplay/TrickplayProvider.cs
nicknsy: Blocking the library scan was a very unpopular choice that caused a ton of confusion which is why a non-blocking option was added and made the default. Plus, trickplay data is non-critical and not really necessary information when it comes to adding something to the library, so blocking the scan for upwards of 10 minutes per video depending on hardware would mean almost nobody would use this feature.

MediaBrowser.Model/Configuration/TrickplayOptions.cs
nicknsy: The default previously was to used the same values as the rest of jellyfin, but the problem is that trickplay generation often runs continuously on larger libraries or slower hardware, leading to confusion when people see high resource usage. Even today a came across such a thread: [reddit.com/r/jellyfin/comments/13jlop9/ffmpeg_is_using_over_100_cpu/](https://www.reddit.com/r/jellyfin/comments/13jlop9/ffmpeg_is_using_over_100_cpu/).

MediaBrowser.Providers/Trickplay/TrickplayManager.cs
nicknsy: Wasn't able to figure out a way to do this cleanly. IImageEncoder does not contain a Decode method--only SkiaEncoder does. With the system of ,NET DI I don't believe I can control which implementation of IImageEncoder the class gets. If I do type-checking that sort of breaks the point of the encoder abstraction, and I can't easily import Jellyfin.Drawing.Skia anyways since that seems to be built separately, so importing would break the whole MediaBrowser interface -> Jellyfin impl design.

Really it looks like there's only a SkiaEncoder and a NullImpl when no encoder is available. I could just add Decode to IImageEncoder.

Jellyfin.Api/Controllers/TrickplayController.cs
nicknsy: > I think this function can live inside of TrickPlayManager? I agree that it doesn't belong in the controller

It's difficult to implement in the manager in a way that makes sense. The trickplay playlist is intrinsically linked to the trickplay controller URL that serves the images, as well as the current http context to include the api key in the playlist.

The only way to make a playlist then would be to pass in the api key, which does make it acessible to other api controllers, but doesn't seem like a realistic use case. 

I could remove the authentication entirely, or implement it as stated above, but given that the current implementation is consistent with other HLS endpoints like the SubtitleController, I don't see a very compelling reason to do so. https://github.com/jellyfin/jellyfin/blob/bb5bf0277ae33fd4668e00bd2f1a85e89aef466a/Jellyfin.Api/Controllers/SubtitleController.cs#L324-L329

MediaBrowser.Providers/Trickplay/TrickplayManager.cs
crobibero: > so I could just add Decode to IImageEncoder. What are your thoughts?

Yeah that's what I was thinking. Skia isn't always available on the platform

MediaBrowser.Providers/Trickplay/TrickplayProvider.cs
nicknsy: I believe this is how the code currently functions. Users can selectively enable trickplay extraction and extraction on scan for each individual library. Only libraries with extraction enabled will ever have trickplay images generated, but libraries with trickplay enabled and scan generation disabled will have images generated through the scheduled task.

The enabled library check, along with other checks like media stream count, is done in the TrickplayManager to avoid code duplication across the task and metadata provider. The TrickplayProvider itself checks for enable on scan since that's the only code run on library scan. 

https://github.com/jellyfin/jellyfin/blob/f1874e2a489e0ed042021cefcf7b3d008ea3bf33/MediaBrowser.Providers/Trickplay/TrickplayManager.cs#L319-L323 

https://github.com/jellyfin/jellyfin/blob/f1874e2a489e0ed042021cefcf7b3d008ea3bf33/MediaBrowser.Providers/Trickplay/TrickplayProvider.cs#L101-L102

crobibero: Ah perfect, I missed that call in `CanGenerateTrickplay`

I'm still not a fan of the nonblocking call, but I guess it can stay for now

Jellyfin.Api/Controllers/TrickplayController.cs
```diff
@@ -0,0 +1,169 @@
...
+                        .AppendLine(",");
+
+                    // EXT-X-TILES
+                    builder
+                        .Append("#EXT-X-TILES:RESOLUTION=")
+                        .Append(resolution)
+                        .Append(",LAYOUT=")
+                        .Append(layout)
+                        .Append(",DURATION=")
+                        .AppendLine(string.Format(CultureInfo.InvariantCulture, decimalFormat, tileDuration));
```
crobibero, line 156:
```suggestion
                        .AppendFormat(CultureInfo.InvariantCulture, decimalFormat, tileDuration)
                        .AppendLine();
```

Jellyfin.Api/Controllers/TrickplayController.cs
```diff
@@ -0,0 +1,169 @@
...
+                        urlFormat,
+                        width.ToString(CultureInfo.InvariantCulture),
+                        i.ToString(CultureInfo.InvariantCulture),
+                        mediaSourceId.ToString("N"),
+                        User.GetToken());
+
+                    // EXTINF
+                    builder
+                        .Append("#EXTINF:")
+                        .Append(string.Format(CultureInfo.InvariantCulture, decimalFormat, infDuration))
```
crobibero, line 146:
```suggestion
                        .AppendFormat(CultureInfo.InvariantCulture, decimalFormat, infDuration)
```

Jellyfin.Api/Controllers/TrickplayController.cs
```diff
@@ -0,0 +1,169 @@
...
+                {
+                    // All tile grids before the last one must contain full amount of tiles.
+                    // The final grid will be 0 < count <= maxTiles
+                    if (i == tileGridCount - 1)
+                    {
+                        tilesPerGrid = tilesInfo.TileCount - (i * tilesPerGrid);
+                        infDuration = tileDuration * tilesPerGrid;
+                    }
+
+                    var url = string.Format(
```
crobibero, line 135: This can also be `.AppendFormat` as it is only used once below

Jellyfin.Api/Controllers/TrickplayController.cs
```diff
@@ -0,0 +1,169 @@
...
+                    // URL
+                    builder.AppendLine(url);
+                }
+
+                builder.AppendLine("#EXT-X-ENDLIST");
+                return new FileContentResult(Encoding.UTF8.GetBytes(builder.ToString()), MimeTypes.GetMimeType("playlist.m3u8"));
+            }
+        }
+
+        return new FileContentResult(Array.Empty<byte>(), MimeTypes.GetMimeType("playlist.m3u8"));
```
crobibero, line 167: I think it would be better to return `NotFound` here

Jellyfin.Api/Controllers/TrickplayController.cs
```diff
@@ -0,0 +1,169 @@
...
+
+            if (tilesInfo.TileCount > 0)
+            {
+                const string urlFormat = "Trickplay/{0}/{1}.jpg?MediaSourceId={2}&api_key={3}";
+                const string decimalFormat = "{0:0.###}";
+
+                var resolution = $"{tilesInfo.Width}x{tilesInfo.Height}";
+                var layout = $"{tilesInfo.TileWidth}x{tilesInfo.TileHeight}";
+                var tilesPerGrid = tilesInfo.TileWidth * tilesInfo.TileHeight;
+                var tileDuration = tilesInfo.Interval / 1000m;
```
crobibero, line 112:
```suggestion
                var tileDuration = tilesInfo.Interval / 1000d;
```

I think decimal is overkill here

Jellyfin.Api/Controllers/TrickplayController.cs
```diff
@@ -0,0 +1,169 @@
+using System;
+using System.Collections.Generic;
```
crobibero, line 2:
```suggestion
```

Jellyfin.Api/Controllers/TrickplayController.cs
```diff
@@ -0,0 +1,169 @@
...
+using System.ComponentModel.DataAnnotations;
+using System.Globalization;
+using System.IO;
+using System.Linq;
```
crobibero, line 6:
```suggestion
```

MediaBrowser.Providers/Trickplay/TrickplayManager.cs
nicknsy: Looked at it again and realized just adding a Decode to IImageEncoder doesn't solve the issue of abstraction since it would have to return a SkiaSharp bitmap, and TrickplayManager still uses things like SkCanvas. Would it be better to move the trickplay tile image creation to IImageEncoder as ```CreateTrickplayTile``` and implement it in SkiaEncoder similar to CreateSplashscreen and CreateImageCollage?

crobibero: Your suggestion of `CreateTrickplayTile` makes sense to me 👍

crobibero: @nicknsy Just an fyi- the only thing blocking approval from me is this thread

MediaBrowser.Model/Configuration/TrickplayOptions.cs
JPVenson: I see, however when i first run jellyscrub it took me well over a week with constant fan noise anyway. That came from the idea that the default settings where the most optimal so i assumed the whole process was just _slow_. I would assume that the default options are the fastest available so if that is no longer the case, the whole proccess would take months instead? that cannot be the solution, then a lot of people will complain that the scrub feature is not working because its so slow and i find that harder to explain to the avg user than a few days of constant usage.

But i see the dilemma here.

I would say, per default HW accelerated generation with max performance settings should be enabled.

crobibero (approved):

Shadowghost (approved):

Emby.Server.Implementations/Data/SqliteItemRepository.cs
```diff
@@ -383,6 +384,8 @@ public override void Initialize()
 
                 "create table if not exists " + ChaptersTableName + " (ItemId GUID, ChapterIndex INT NOT NULL, StartPositionTicks BIGINT NOT NULL, Name TEXT, ImagePath TEXT, PRIMARY KEY (ItemId, ChapterIndex))",
 
+                "create table if not exists " + TrickplayTableName + " (ItemId GUID, Width INT NOT NULL, Height INT NOT NULL, TileWidth INT NOT NULL, TileHeight INT NOT NULL, TileCount INT NOT NULL, Interval INT NOT NULL, Bandwidth INT NOT NULL, PRIMARY KEY (ItemId, Width))",
```
cvium, line 387: I feel we should be using EF for this especially if there's no foreign key.

Jellyfin.Api/Controllers/TrickplayController.cs
```diff
@@ -0,0 +1,168 @@
...
+        var path = _trickplayManager.GetTrickplayTilePath(item, width, index);
+        if (System.IO.File.Exists(path))
+        {
+            return PhysicalFile(path, MediaTypeNames.Image.Jpeg);
+        }
+
+        return NotFound();
+    }
+
+    private ActionResult GetTrickplayPlaylistInternal(int width, Guid mediaSourceId)
```
cvium, line 95: I prefer having thin controllers. This kind of logic should be in another class like the TrickplayManager

MediaBrowser.Model/Configuration/EncodingOptions.cs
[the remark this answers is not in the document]
cvium: @nyanmisaka thoughts?

MediaBrowser.Providers/Trickplay/TrickplayImagesTask.cs
```diff
@@ -0,0 +1,118 @@
...
+
+        var numberOfVideos = _libraryManager.GetCount(query);
+
+        var startIndex = 0;
+        var numComplete = 0;
+
+        while (startIndex < numberOfVideos)
+        {
+            query.StartIndex = startIndex;
+            var videos = _libraryManager.GetItemList(query).OfType<Video>().ToList();
```
cvium, line 94: Why ToList?

MediaBrowser.Providers/Trickplay/TrickplayManager.cs
```diff
@@ -0,0 +1,349 @@
...
+            tilesInfo.Bandwidth = Math.Max(tilesInfo.Bandwidth, bitrate);
+        }
+
+        /*
+         * Move trickplay tiles to output directory
+         */
+        Directory.CreateDirectory(outputDir);
+
+        // Replace existing tile grids if they already exist
+        if (Directory.Exists(outputDir))
```
cvium, line 252: This is always true since you're creating it on line 249?

Emby.Server.Implementations/Data/SqliteItemRepository.cs
Shadowghost: Can we just use EF on a non-EF database?

cvium: In theory yes, but it should be placed in the jellyfin.db

MediaBrowser.Model/Configuration/ServerConfiguration.cs
```diff
@@ -264,5 +264,7 @@ public ServerConfiguration()
         /// </summary>
         /// <value>The limit for parallel image encoding.</value>
         public int ParallelImageEncodingLimit { get; set; }
+
+        public TrickplayOptions TrickplayOptions { get; set; } = new TrickplayOptions();
```
Shadowghost, line 268: Documentation on this would be nice to have (and be consistent with the rest of the file)

Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
```diff
@@ -0,0 +1,468 @@
...
+    private readonly ILogger<TrickplayManager> _logger;
+    private readonly IMediaEncoder _mediaEncoder;
+    private readonly IFileSystem _fileSystem;
+    private readonly EncodingHelper _encodingHelper;
+    private readonly ILibraryManager _libraryManager;
+    private readonly IServerConfigurationManager _config;
+    private readonly IImageEncoder _imageEncoder;
+    private readonly IDbContextFactory<JellyfinDbContext> _dbProvider;
+
+    private static readonly SemaphoreSlim _resourcePool = new(1, 1);
```
crobibero, line 38: I think we can allow some concurrency here.. or is this a very intensive task?

crobibero (approved):
Please don't force push, it makes us do a full re-review :)

Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
nicknsy: Yeah this is a very intensive task and the intended route for improving speed is just upping the thread count and allowing ffmpeg to use multiple cores as needed which it seems to be very good at. Once a second ffmpeg process is added it just halves the speed and so on. 

We discussed the possibility of concurrency to make better use of the GPU in https://github.com/nicknsy/jellyscrub/issues/77, but even on high-end server GPUs we didn't see any improvement doing multiple ffmpeg extractions at once, though there could have been flaws in the testing methodology there.

Jellyfin.Api/Helpers/DynamicHlsHelper.cs
```diff
@@ -301,6 +307,13 @@ public class DynamicHlsHelper
             AppendPlaylist(builder, state, variantUrl, newBitrate, subtitleGroup);
         }
 
+        if (!isLiveStream && (state.VideoRequest?.EnableTrickplay).GetValueOrDefault(false))
```
cvium, line 310: For consistency with the code base
```suggestion
        if (!isLiveStream && (state.VideoRequest?.EnableTrickplay ?? false))
```

Jellyfin.Api/Helpers/DynamicHlsHelper.cs
```diff
@@ -529,6 +542,41 @@ private void AddSubtitles(StreamState state, IEnumerable<MediaStream> subtitles,
...
+
+            var line = string.Format(
+                CultureInfo.InvariantCulture,
+                playlistFormat,
+                trickplayInfo.Bandwidth.ToString(CultureInfo.InvariantCulture),
+                trickplayInfo.Width.ToString(CultureInfo.InvariantCulture),
+                trickplayInfo.Height.ToString(CultureInfo.InvariantCulture),
+                url);
+
+            builder.AppendLine(line);
```
cvium, line 576: Not that I think it matters too much in the grand scheme, but StringBuilder.AppendFormat is more memory efficient than string.Format and AppendLine

Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
```diff
@@ -0,0 +1,474 @@
...
+        var thumbnailsPerTile = trickplayInfo.TileWidth * trickplayInfo.TileHeight;
+        var requiredTiles = (int)Math.Ceiling((double)images.Count / thumbnailsPerTile);
+
+        for (int i = 0; i < requiredTiles; i++)
+        {
+            // Set output/input paths
+            var tilePath = Path.Combine(workDir, $"{i}.jpg");
+
+            imageOptions.OutputPath = tilePath;
+            imageOptions.InputPaths = images.Skip(i * thumbnailsPerTile).Take(thumbnailsPerTile).ToList();
```
cvium, line 242: Would be better to use List.GetRange I think? https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.getrange?view=net-7.0

MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
```diff
@@ -775,6 +778,184 @@ private async Task<string> ExtractImageInternal(string inputPath, string contain
...
+                    {
+                        if (await process.WaitForExitAsync(TimeSpan.FromSeconds(30)).ConfigureAwait(false))
+                        {
+                            ranToCompletion = true;
+                            break;
+                        }
+
+                        cancellationToken.ThrowIfCancellationRequested();
+
+                        var jpegCount = _fileSystem.GetFilePaths(targetDirectory)
```
cvium, line 928: Is it necessary to filter for jpgs? Isn't that the only files that will show up in that dir?

If it is necessary to filter, use the overload `GetFilePaths(string path, string[]? extensions, bool enableCaseSensitiveExtensions, bool recursive = false)` and use a static array for the `extensions`.

Emby.Server.Implementations/Dto/DtoService.cs
```diff
@@ -1058,6 +1062,15 @@ private void AttachBasicFields(BaseItemDto dto, BaseItem item, BaseItem owner, D
                     dto.Chapters = _itemRepo.GetChapters(item);
                 }
 
+                if (options.ContainsField(ItemFields.Trickplay))
+                {
+                    var manifest = _trickplayManager.GetTrickplayManifest(item).GetAwaiter().GetResult();
+
+                    // To stay consistent with other fields, this must go from a Guid to a non-dashed string.
+                    // This does not seem to occur automatically to dictionaries like it does with other Guid fields.
+                    dto.Trickplay = manifest.ToDictionary(x => x.Key.ToString("N", CultureInfo.InvariantCulture), y => y.Value);
```
cvium, line 1071: Isn't this the only place where `GetTrickplayManifest` is called? Shouldn't it just return this dict?

cvium (changes requested):

cvium (approved):

Shadowghost (approved):

Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
```diff
@@ -0,0 +1,474 @@
...
+    }
+
+    /// <inheritdoc />
+    public string GetTrickplayTilePath(BaseItem item, int width, int index)
+    {
+        return Path.Combine(GetTrickplayDirectory(item, width), index + ".jpg");
+    }
+
+    /// <inheritdoc />
+    public async Task<string?> GetHlsPlaylist(Guid itemId, int width, string? apiKey)
```
Bond-009, line 376: Would it make more sense to write to a stream instead of creating a stringbuilder to create a string only to convert it to a byte array?

nicknsy: The StringBuilder -> byte array pattern is just what was used in the SubtitleController that I modeled this after, but I am open to changing that.

Looks like I could use a MemoryStream + StreamWriter to avoid converting the playlist to a byte array in the end, though it won't be as elegant as using a StringBuilder since it requires lots of ```await stream.WriteLineAsync(string.Format(...))``` rather than being able to chain Appends,

What are your thoughts on just changing the return line to ```return Content(playlist, MimeTypes.GetMimeType("playlist.m3u8"), Encoding.UTF8);```. Both a MemoryStream or string would still need to be copied to the HTTP output stream so this should have a similar memory footprint as it avoids copying the string into an additional byte array before being written to the HTTP output (I assume).

Bond-009: Sounds good 👍

ebkalderon:
Just a curious passerby, but is trickplay a feature slated for the 10.9.0 release? Or will this be coming in a later release? I don't see a GitHub milestone assigned, so I was just wondering. Thanks! Feel free to mark this comment off-topic after replying, if you wish.

nielsvanvelzen:
Everything merged will be in 10.9

memehammad:
Any chance the option to save trickplay files next to the media could be added to this so we dont have to recreate the bif files when resetting our config or something?

nicknsy:
> Any chance the option to save trickplay files next to the media could be added to this so we dont have to recreate the bif files when resetting our config or something?

The official trickplay implementation doesn't work with BIF files regardless. Jellyscrub files will need to be converted to native files using the 10.9.0 plugin which I will release soon.

n00mkrad:
> Any chance the option to save trickplay files next to the media could be added to this so we dont have to recreate the bif files when resetting our config or something?

Bump, is this possible?
Seems like JF only stores it in the metadata folder right now.

crobibero:
Saving trickplay files next to media is a feature request. Please do not comment on old PRs requesting new features.

n00mkrad:
> Saving trickplay files next to media is a feature request. Please do not comment on old PRs requesting new features.

I assumed it was a setting somewhere since the plugin that got adopted (JellyScrub) had such a feature.

PR #9560, opened 2023-03-29
Sort embedded collections in Nfo files

IDisposable:
**Changes**
Because the Nfo files emit the collections as they are in-memory, the files are not stable in format, genres, tags, albums, people, etc. are emitted in random orders. Add ordering of the collections when emitting the Nfo files so the file remains stable (unchanged) when underlying media information doesn't change.

Specific ordering: (alphabetical/numeric ascending after trimming blanks and defaulting to zero for missing numbers)

This ALSO cleans up all the names/titles/artists/albums/tags/genres/etc. to ensure they are `Trim()` of leading and trailing blanks (because they often have blanks), and does this before the Distinct to ensure we don't end up with `'foo', 'foo ',` in the outputs.

BaseItem: Directors, Writers, Trailers (by Url), Production Locations, Genres, Studios, Tags, Custom Provider Data (by key), Linked Children  (by Path>LibraryItemId), Backdrop Images (by path), Actors (by SortOrder>Name)

AlbumNfo: Artists, Album Artists, Tracks (by ParentIndexNumber>IndexNumber>Name)

ArtistNfo: Albums (by Production Year>SortName>Name)

MovieNfo: Artists

**Issues**
N/A


IDisposable:
Fixing commit issues.

MediaBrowser.Controller/Entities/BaseItem.cs
```diff
@@ -904,7 +905,7 @@ protected virtual string CreateSortName()
                 // Remove from middle if surrounded by spaces
                 sortable = sortable.Replace(" " + search + " ", " ", StringComparison.Ordinal);
 
-                // Remove from end if followed by a space
+                // Remove from end if preceeded by a space
```
IDisposable, line 908: Just a typo I noticed.

MediaBrowser.Controller/Entities/BaseItem.cs
```diff
@@ -1780,7 +1781,7 @@ public void AddStudio(string name)
 
         public void SetStudios(IEnumerable<string> names)
         {
-            Studios = names.Distinct().ToArray();
+            Studios = names.Trimmed().Distinct().ToArray();
```
IDisposable, line 1784: `Trimmed()` against an `IEnumerable<string>` returns an enumeration of the values where each is `Trim()`ed.

MediaBrowser.Controller/Sorting/SortExtensions.cs
```diff
@@ -30,5 +30,15 @@ public static IOrderedEnumerable<T> ThenByStringDescending<T>(this IOrderedEnume
         {
             return list.ThenByDescending(getName, _comparer);
         }
+
+        public static string Trimmed(this string? value)
```
IDisposable, line 34: This one gets used when we have a possibly-null `string`

MediaBrowser.LocalMetadata/Parsers/BaseItemXmlParser.cs
```diff
@@ -449,7 +449,7 @@ protected virtual void FetchDataFromXmlNode(XmlReader reader, MetadataResult<T>
 
                     if (!string.IsNullOrWhiteSpace(val))
```
IDisposable, line 450: While it might seem this handles blanks, it doesn't handle _trailing_ or _leading_ blanks, just _only_ blanks.

MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
```diff
@@ -494,42 +495,44 @@ private List<NameValuePair> ReadValueArray(XmlReader reader)
         private void ProcessPairs(string key, List<NameValuePair> pairs, MediaInfo info)
         {
             IList<BaseItemPerson> peoples = new List<BaseItemPerson>();
+            var distinctPairs = pairs.Select(p => p.Value)
```
IDisposable, line 498: Do the trim and distincting (in that order!) for all of these people lists, not just studios

MediaBrowser.Providers/Manager/ProviderManager.cs
```diff
@@ -913,7 +913,8 @@ public void OnRefreshStart(BaseItem item)
             catch (Exception ex)
             {
                 // EventHandlers should never propagate exceptions, but we have little control over plugins...
-                _logger.LogError(ex, "Invoking {RefreshEvent} event handlers failed", nameof(RefreshStarted));
+                if (ex is not MissingMethodException)
```
IDisposable, line 916: Ooops, this probably shouldn't have been included with this patch... it's there because the providers are throwing errors due to API changes... ignore this.

MediaBrowser.XbmcMetadata/Savers/BaseNfoSaver.cs
```diff
@@ -763,7 +767,7 @@ public static void AddMediaInfo<T>(T item, XmlWriter writer)
                             XmlConvert.VerifyName(tagName);
                             Logger.LogDebug("Saving custom provider tagname {0}", tagName);
 
-                            writer.WriteElementString(GetTagForProviderKey(providerKey), providerId);
+                            writer.WriteElementString(tagName, providerId);
```
IDisposable, line 770: we just call that same method 4 lines up!

IDisposable:
I noticed the build with these changes is producing a difference in album.nfo and artist.nfo files compared to the 10.8.9 version I have actually running on my server. The new version of an example NFO shows that the Artist and AlbumArtist tags are **also** now being emitted as `<actor>` nodes. I don't see how my changes could have caused this, and I doubt it's intentional... should I fix this, or was the actor change intentional @crobibero or @Shadowghost?

```xml
<?xml version="1.0" encoding="utf-8" standalone="yes"?>
<album>
  <review />
  <outline />
  <lockdata>false</lockdata>
  <dateadded>2011-08-01 20:01:41</dateadded>
  <title>In Scarlet Storm</title>
  <year>1990</year>
  <premiered>1990-01-01</premiered>
  <releasedate>1990-01-01</releasedate>
  <runtime>37</runtime>
  <genre>Hard Rock</genre>
  <musicbrainzalbumid>fd0dd665-8f2e-421d-af80-6b08b120a13c</musicbrainzalbumid>
  <musicbrainzreleasegroupid>c522d109-09d8-3fdb-a9ba-627c24c6bb6a</musicbrainzreleasegroupid>
  <art>
    <poster>\\ZULU\Music\Z\David Zaffiro\In Scarlet Storm\Folder.jpg</poster>
  </art>
  <actor>
    <name>David Zaffiro</name>
    <type>AlbumArtist</type>
  </actor>
  <actor>
    <name>David Zaffiro</name>
    <type>Artist</type>
  </actor>
  <artist>David Zaffiro</artist>
  <albumartist>David Zaffiro</albumartist>
  <track>
    <position>1</position>
    <title>Through His Eyes</title>
    <duration>04:24</duration>
  </track>
   ...
</album>
```

Shadowghost:
That should not have happened. If changes by me caused it it was unintentional. If you know the reason, feel free to fix it.

IDisposable:
Now will not emit actors for MusicAlbum nor MusicArtist

IDisposable:
Finally clean build

MediaBrowser.Controller/Sorting/SortExtensions.cs
crobibero: I think I would prefer to just do `value?.Trim()` instead of having a function that turns null into an empty string.

IDisposable: That's fine. Excepting there's a lot of places where this is used in an OrderBy or ThenBy and was coalescing to a `string.Empty` or `""`. You make the call for what you want, I can rip through to swap it.

IDisposable: I've swapped over to `?.Trim()`

Emby.Server.Implementations/LiveTv/Listings/SchedulesDirect.cs
```diff
@@ -463,10 +463,10 @@ private static double GetAspectRatio(ImageDataDto i)
             }
 
             StringBuilder str = new StringBuilder("[", 1 + (programIds.Count * 13));
-            foreach (ReadOnlySpan<char> i in programIds)
+            foreach (var i in programIds)
```
IDisposable, line 466: This now generated a compilation error because you can't uses stack-allocated `ReadOnlySpan<char>` in an `async` method

IDisposable:
Rebased to current master

Shadowghost (approved):

IDisposable:
Resolved conflict :shipit:

MediaBrowser.XbmcMetadata/Savers/BaseNfoSaver.cs
```diff
@@ -511,7 +515,7 @@ public static void AddMediaInfo<T>(T item, XmlWriter writer)
                 writer.WriteElementString("credits", person);
             }
 
-            foreach (var trailer in item.RemoteTrailers)
+            foreach (var trailer in item.RemoteTrailers.OrderBy(t => t.Url?.Trim()))
```
cvium, line 518: This feels kinda hacky. Why do we need to trim strings coming from our DB?

IDisposable: We could remove this here and do a upgrade/conversion pass... but a _lot_ of fields we have in the database have trailing blanks because the source INFO provider had blanks and we didn't trim when copying the data over. As for this _specific_ field, probably could remove because the `Uri`s are unlikely to be bad, I was just following a consistent pattern (mini cargo cult for this PR).

IDisposable:
Rebased to current master to resolve conflicts

fogolin:
Hi there, can we get the second approval on this from and admin?

IDisposable:
If this is going to be merged eventually, I'll fix the conflicts. Just let me know

IDisposable:
Merge conflicts and CodeQL issues resolved. Builds cleanly.

JPVenson (approved):

MarkCiliaVincenti (approved):

IDisposable:
Fixed merge conflicts (again). This is GTG.

IDisposable:
Resolved conflict (again) :shipit:

Shadowghost:
@IDisposable could you please rebase again (sorry about that)

IDisposable:
Sure! I will probably have to do a little less now that thing have progressed. Will make it minimal

MediaBrowser.LocalMetadata/Parsers/BaseItemXmlParser.cs
```diff
@@ -234,8 +234,8 @@ protected virtual void FetchDataFromXmlNode(XmlReader reader, MetadataResult<T>
                     item.CustomRating = reader.ReadNormalizedString();
                     break;
                 case "RunningTime":
-                    var runtimeText = reader.ReadElementContentAsString();
-                    if (!string.IsNullOrWhiteSpace(runtimeText))
+                    var runtimeText = reader.ReadNormalizedString();
```
IDisposable, line 237: Moved to `ReadNormalizedString()` because it internally `.Trim()`s which means we can switch to `IsNullOrEmpty` now

MediaBrowser.LocalMetadata/Parsers/BaseItemXmlParser.cs
```diff
@@ -580,7 +580,12 @@ private void FetchFromTaglinesNode(XmlReader reader, T item)
                     switch (reader.Name)
                     {
                         case "Tagline":
-                            item.Tagline = reader.ReadNormalizedString();
+                            var val = reader.ReadNormalizedString();
+                            if (!string.IsNullOrEmpty(val))
```
IDisposable, line 584: This might be functionally different in that we won't set from a completely blank tagline, but that's now consistent with all the other properties.

MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
```diff
@@ -531,42 +533,44 @@ private List<NameValuePair> ReadValueArray(XmlReader reader)
         private void ProcessPairs(string key, List<NameValuePair> pairs, MediaInfo info)
         {
             List<BaseItemPerson> peoples = new List<BaseItemPerson>();
+            var distinctPairs = pairs.Select(p => p.Value)
+                    .Where(i => !string.IsNullOrWhiteSpace(i))
+                    .Trimmed()
+                    .Distinct(StringComparer.OrdinalIgnoreCase);
```
IDisposable, line 539: The purpose of this is to remove duplicates in an input list as we sometimes get the same person twice.

MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
```diff
@@ -607,8 +611,8 @@ private NameValuePair GetNameValuePair(XmlReader reader)
                 }
             }
 
-            if (string.IsNullOrWhiteSpace(name)
-                || string.IsNullOrWhiteSpace(value))
+            if (string.IsNullOrEmpty(name)
```
IDisposable, line 614: They're pretrimmed now.

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
```diff
@@ -309,13 +310,15 @@ private async Task FetchDataFromTags(Audio audio, Model.MediaInfo.MediaInfo medi
...
             {
-                var genres = string.IsNullOrEmpty(track.Genre) ? mediaInfo.Genres : track.Genre.Split(InternalValueSeparator).Distinct(StringComparer.OrdinalIgnoreCase).ToArray();
+                var genres = string.IsNullOrEmpty(track.Genre) ? mediaInfo.Genres : track.Genre.Split(InternalValueSeparator);
 
                 if (libraryOptions.UseCustomTagDelimiters)
                 {
                     genres = genres.SelectMany(g => SplitWithCustomDelimiter(g, libraryOptions.CustomTagDelimiters, libraryOptions.DelimiterWhitelist)).ToArray();
                 }
 
+                genres = genres.Trimmed().Distinct(StringComparer.OrdinalIgnoreCase).ToArray();
```
IDisposable, line 320: Have to do the `Distinct` after the sub-splitting because we could get duplicates now.

IDisposable:
All ready @Shadowghost

MediaBrowser.Controller/Sorting/SortExtensions.cs
```diff
@@ -30,5 +30,10 @@ public static IOrderedEnumerable<T> ThenByStringDescending<T>(this IOrderedEnume
         {
             return list.ThenByDescending(getName, _comparer);
         }
+
+        public static IEnumerable<string> Trimmed(this IEnumerable<string> values)
```
Bond-009, line 34: This isn't the right place for this, maybe somewhere in Jellyfin.Extensions?

IDisposable: Can do!

IDisposable:
Moved Trimmed to Jellyfin.Extensions.StringExtensions;

MediaBrowser.Controller/Sorting/SortExtensions.cs
IDisposable: Done

Bond-009 (approved):

IDisposable:
I cleaned up the excess `using`s and now honors the `.SortName` when specified.

IDisposable:
Fixed conflicts _again_.

IDisposable:
The _run-tests_ error is because the `IsDefault` isn't being set for the test of a single-stream file which has an subtitle stream and **does not** seem to be related to these changes at all.

I think this happens because that ignores in the stream's `stream.IsDefault` flag and only cares about `path.IsDefault` unlike what is done for `IsForced` and `IsHearingImpaired`. Not sure when this became a broken test.

See [MediaInfoResolver.cs  lines 116-127](https://github.com/jellyfin/jellyfin/blob/master/MediaBrowser.Providers/MediaInfo/MediaInfoResolver.cs#L116-L127)

Perhaps that test would be corrected by making line 122 this:
```c#
mediaStream.IsDefault = pathInfo.IsDefault || mediaStream.IsDefault;
```

PR #7039, opened 2021-12-23
Providermanager cleanup

1337joe:
A bunch of incremental improvements to ProviderManager.cs and related files, primarily focusing on `RefreshSingleItem`, `GetImageProviders`, and `GetMetadataProviders`.

**Changes**
- Pulled a couple public methods up to the interface, resolved `#nullable disable`, general minor cleanup
- Stored library type options and fetcher order logic up front instead of rechecking multiple times for each provider
- Merged and simplified similar/duplicated code for image and metadata sorting
- Added unit tests

PR #9566, opened 2023-03-31
Restore old interface for IBaseItemManager interface

IDisposable:
The changes made in PR #7039 break Metadata plug-ins that may not have yet been recompiled. Added the original methods back and made the base implementation mirror what the old code was doing to prevent the old plug-ins from throwing `MissingMethodExceptions`


crobibero:
This was an intentional change. Plug-ins will need to be updated when Jellyfin 10.9 is released. Plug-ins are not supported in unstable Jellyfin versions.

IDisposable:
Understood, I'll just keep this hack around until released

PR #9570, opened 2023-03-31
Modified await statement to remove compiler warning while awaiting on the file stream to be created.

jbatt33:
Working on compiler warning for this line of code. 

**Changes**
The changes to how the file stream is declared and added some brackets for the using.

The using will still be in the main context, while the flush will happen outside.

### DIFF
--- a/MediaBrowser.Controller/ClientEvent/ClientEventLogger.cs
+++ b/MediaBrowser.Controller/ClientEvent/ClientEventLogger.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.IO;
 using System.Threading.Tasks;
 
@@ -23,8 +23,12 @@ namespace MediaBrowser.Controller.ClientEvent
         {
             var fileName = $"upload_{clientName}_{clientVersion}_{DateTime.UtcNow:yyyyMMddHHmmss}_{Guid.NewGuid():N}.log";
             var logFilePath = Path.Combine(_applicationPaths.LogDirectoryPath, fileName);
-            await using var fileStream = new FileStream(logFilePath, FileMode.CreateNew, FileAccess.Write, FileShare.None);
-            await fileContents.CopyToAsync(fileStream).ConfigureAwait(false);
+            FileStream fileStream;
+            await using (fileStream = new FileStream(logFilePath, FileMode.CreateNew, FileAccess.Write, FileShare.None))
+            {
+                await fileContents.CopyToAsync(fileStream).ConfigureAwait(false);
+            }
+
             return fileName;
         }
     }


MediaBrowser.Controller/ClientEvent/ClientEventLogger.cs
```diff
@@ -23,8 +23,12 @@ public async Task<string> WriteDocumentAsync(string clientName, string clientVer
         {
             var fileName = $"upload_{clientName}_{clientVersion}_{DateTime.UtcNow:yyyyMMddHHmmss}_{Guid.NewGuid():N}.log";
             var logFilePath = Path.Combine(_applicationPaths.LogDirectoryPath, fileName);
-            await using var fileStream = new FileStream(logFilePath, FileMode.CreateNew, FileAccess.Write, FileShare.None);
-            await fileContents.CopyToAsync(fileStream).ConfigureAwait(false);
+            FileStream fileStream;
+            await using (fileStream = new FileStream(logFilePath, FileMode.CreateNew, FileAccess.Write, FileShare.None))
```
IDisposable, line 27: `await using (var fileStream = new FileStream(logFilePath, FileMode.CreateNew, FileAccess.Write, FileShare.None))` would be better... no reason for distinct declaration.

MediaBrowser.Controller/ClientEvent/ClientEventLogger.cs
```diff
@@ -23,8 +23,12 @@ public async Task<string> WriteDocumentAsync(string clientName, string clientVer
         {
             var fileName = $"upload_{clientName}_{clientVersion}_{DateTime.UtcNow:yyyyMMddHHmmss}_{Guid.NewGuid():N}.log";
             var logFilePath = Path.Combine(_applicationPaths.LogDirectoryPath, fileName);
-            await using var fileStream = new FileStream(logFilePath, FileMode.CreateNew, FileAccess.Write, FileShare.None);
-            await fileContents.CopyToAsync(fileStream).ConfigureAwait(false);
```
IDisposable, line 27: This seems to be a common anti-pattern... the `using` line should always have braces enclosing the body using the `IDisposable` object, that's the entire purpose OF the using statement. Seems that when used here, we're seeing a false understanding that the next line is the scope of the using, rather it is "anytime you want, but before the next-outer scope close" is the real spec.

I think we need to make a sweep looking for this (I just fixed an issue caused by this error in PR #9559), and I might be the right person to do it :)

crobibero: This isn't an anti-pattern but was a feature added in C# 8.
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/using

IDisposable: It does cause the class of bugs I fixed in PR #9559.

MediaBrowser.Controller/ClientEvent/ClientEventLogger.cs
jbatt33: I was not able to resolve the warning without writing it this way.

MediaBrowser.Controller/ClientEvent/ClientEventLogger.cs
jbatt33: Looks like some good reading here for me.

Bond-009:
I think this isn't needed anymore due to #10390

PR #9572, opened 2023-03-31
Resolving Warning related to ProviderIds

jbatt33:
**Changes**
Warning related to ProviderIds being initialized in the constructor and having its collection reference changed externally.
Changed the ProviderIds property while maintaining the interface's set implementation.

**Issues**
Compiler warning.

### DIFF
--- a/MediaBrowser.Controller/Collections/CollectionCreationOptions.cs
+++ b/MediaBrowser.Controller/Collections/CollectionCreationOptions.cs
@@ -25,7 +25,7 @@ namespace MediaBrowser.Controller.Collections
 
         public bool IsLocked { get; set; }
 
-        // private implementation of ProviderIds property from interface to allow setter access
+        // Private implementation of ProviderIds property from interface to allow setter access
         Dictionary<string, string> IHasProviderIds.ProviderIds
         {
             get => _providerIds;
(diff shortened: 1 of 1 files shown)


MediaBrowser.Controller/Collections/CollectionCreationOptions.cs
```diff
@@ -23,7 +25,14 @@ public CollectionCreationOptions()
 
         public bool IsLocked { get; set; }
 
-        public Dictionary<string, string> ProviderIds { get; set; }
+        // private implementation of ProviderIds property from interface to allow setter access
```
Shadowghost, line 28:
```suggestion
        // Private implementation of ProviderIds property from interface to allow setter access
```

Shadowghost (approved):

JPVenson (approved):

barronpm:
What warning does this fix?

jbatt33:
> What warning does this fix?

I need to check again, this is from a while back.

jbatt33:
> What warning does this fix?

Weird, looking at this code change I see what I was doing, but I do not seem to be able to generate this warning from the compiler any more. Also, checking the complete list of warnings for C#, I do not see one related to what this commit attempted to resolve. 

I'm currently not getting any warnings from my jellyfin master compile, so that's not good or is it...

jbatt33:
I'm going to close this since I can not reproduce the original warning I was attempting to resolve. Sorry about that.

Issue #7626, opened 2022-04-20
[10.8.0 beta 2][Issue]: Error: Cannot get image dimensions on all .svg logos


srcrist:
### Please describe your bug

During scans, Jellyfin is sending numerous errors to the logs about not being able to determine the image dimensions for .svg logo files in the metadata. Manual investigation has shown that the files are not corrupt, and otherwise load correctly in a viewer. 

### Jellyfin Version

10.8.0

### Environment

```markdown
- OS: Linux [Ubuntu 21.10]
- Virtualization: Bare Metal
- Clients: N/A
- Browser: N/A
- FFmpeg Version: Jellyfin-FFMPEG 4.4.1-4
- Playback Method: N/A
- Hardware Acceleration: N/A
- Plugins: AniDB, AudioDB, InfuseSync, Kodi Sync Queue, MusicBrainz, OMDb, Studio Images, TMDb, TVmaze, TheTVDB
- Reverse Proxy: nginx
- Base URL: none
- Networking: Direct 1gbps, Data Center
- Storage: local with cloud media
```

### Jellyfin logs

```shell
[2022-04-19 21:58:08.375 -04:00] [ERR] Cannot get image dimensions for "/var/lib/jellyfin/metadata/library/88/88bf6d9eb317fed754379a5d9a1dd814/logo.svg"
Non-success codec result: Unimplemented
Jellyfin.Drawing.Skia.SkiaCodecException: Exception of type 'Jellyfin.Drawing.Skia.SkiaCodecException' was thrown.
   at Jellyfin.Drawing.Skia.SkiaEncoder.GetImageSize(String path)
   at Emby.Drawing.ImageProcessor.GetImageDimensions(BaseItem item, ItemImageInfo info)
   at Emby.Server.Implementations.Library.LibraryManager.UpdateImagesAsync(BaseItem item, Boolean forceUpdate)
[2022-04-19 21:58:30.253 -04:00] [ERR] Cannot get image dimensions for "/var/lib/jellyfin/metadata/library/b4/b4ad202466414ae545d97fedf72a898b/logo.svg"
Non-success codec result: Unimplemented
Jellyfin.Drawing.Skia.SkiaCodecException: Exception of type 'Jellyfin.Drawing.Skia.SkiaCodecException' was thrown.
   at Jellyfin.Drawing.Skia.SkiaEncoder.GetImageSize(String path)
[...]

PR #9581, opened 2023-04-01
#7626 Added handling for common FormatExceptions with Skia loading sv…

JPVenson:
Handles the skia svg errors that will throw when skia loads an svg file that defines an element with style like this:

```
<g id="surface1">
<path style=" stroke:none;fill-rule:nonzero;fill:rgb(100%,100%,100%);fill-opacity:1;" ...
```

gracefully as this is unlikely to be fixed by the upstream repro.
This should reduce the log bloat for users with effected files.

**Changes**
Added try-catch and debug logging for effected code.

**Issues**
Mitigates changes done from PR #7946 for Bug #7626

### DIFF
--- a/src/Jellyfin.Drawing.Skia/SkiaEncoder.cs
+++ b/src/Jellyfin.Drawing.Skia/SkiaEncoder.cs
@@ -130,7 +130,7 @@ public class SkiaEncoder : IImageEncoder
                 // this exception is known to be thrown on vector images that define custom styles
                 // skia svg is not able to parse that and as the repository is quite stale and has not received updates we just catch them
                 _logger.LogDebug(skiaColorException, "There was a issue loading the requested svg file");
-                return new ImageDimensions(0, 0);
+                return default;
             }
         }
 
@@ -142,10 +142,10 @@ public class SkiaEncoder : IImageEncoder
                 return new ImageDimensions(info.Width, info.Height);
             case SKCodecResult.Unimplemented:
                 _logger.LogDebug("Image format not supported: {FilePath}", path);
-                return new ImageDimensions(0, 0);
+                return default;
             default:
                 _logger.LogError("Unable to determine image dimensions for {FilePath}: {SkCodecResult}", path, result);
-                return new ImageDimensions(0, 0);
+                return default;
         }
     }
 
--- a/src/Jellyfin.Drawing.Skia/SkiaEncoder.cs
+++ b/src/Jellyfin.Drawing.Skia/SkiaEncoder.cs
@@ -120,8 +120,18 @@ public class SkiaEncoder : IImageEncoder
         if (extension.Equals(".svg", StringComparison.OrdinalIgnoreCase))
         {
             var svg = new SKSvg();
-            svg.Load(path);
-            return new ImageDimensions(Convert.ToInt32(svg.Picture.CullRect.Width), Convert.ToInt32(svg.Picture.CullRect.Height));
+            try
+            {
+                svg.Load(path);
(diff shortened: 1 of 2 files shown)


src/Jellyfin.Drawing.Skia/SkiaEncoder.cs
```diff
@@ -120,8 +120,18 @@ public ImageDimensions GetImageSize(string path)
...
+            {
+                svg.Load(path);
+                return new ImageDimensions(Convert.ToInt32(svg.Picture.CullRect.Width), Convert.ToInt32(svg.Picture.CullRect.Height));
+            }
+            catch (FormatException skiaColorException)
+            {
+                // this exception is known to be thrown on vector images that define custom styles
+                // skia svg is not able to parse that and as the repository is quite stale and has not received updates we just catch them
+                _logger.LogDebug(skiaColorException, "There was a issue loading the requested svg file");
+                return new ImageDimensions(0, 0);
```
crobibero, line 133: What side-effect happens with a 0,0 image dimensions?

JPVenson: I am not 100% sure but it is the default used below for non-successfull return codes:
https://github.com/jellyfin/jellyfin/blob/c2684e537c5e52b2b4164565bee107d5aaeb8380/src/Jellyfin.Drawing.Skia/SkiaEncoder.cs#L145

crobibero (approved):

src/Jellyfin.Drawing.Skia/SkiaEncoder.cs
JPVenson: Edit:
Seems like 0,0 is handled as a special case in 5/6 callers as an default value specifically.

JPVenson: The 6th case also defaults to 0,0 on exceptions so i assume its fine

src/Jellyfin.Drawing.Skia/SkiaEncoder.cs
```diff
@@ -120,8 +120,18 @@ public ImageDimensions GetImageSize(string path)
...
-            return new ImageDimensions(Convert.ToInt32(svg.Picture.CullRect.Width), Convert.ToInt32(svg.Picture.CullRect.Height));
+            try
+            {
+                svg.Load(path);
+                return new ImageDimensions(Convert.ToInt32(svg.Picture.CullRect.Width), Convert.ToInt32(svg.Picture.CullRect.Height));
+            }
+            catch (FormatException skiaColorException)
+            {
+                // this exception is known to be thrown on vector images that define custom styles
+                // skia svg is not able to parse that and as the repository is quite stale and has not received updates we just catch them
```
Shadowghost, line 131:
```suggestion
                // This exception is known to be thrown on vector images that define custom styles
                // Skia SVG is not able to handle that and as the repository is quite stale and has not received updates we just catch them
```

src/Jellyfin.Drawing.Skia/SkiaEncoder.cs
Bond-009:
```suggestion
                return default;
```

This should work too and is maybe a bit more clear

JPVenson: @Bond-009 should i adjust related instances aswell? I mean if that style then everywhere.

Bond-009: That would be nice

Bond-009 (approved):

PR #9595, opened 2023-04-04
Add advanced options for Intel QSV encoders

knackebrot:
In my testing, enabling LookAhead and ExtBRC increases VMAF score by ~3 in low-bitrate scenarios (89 vs 92 for 1080p at 3Mbit) and Trellis by further ~0.2. The improvement largely depends on the content and I didn't find any sample that would suffer. 

Please note the actual parameters QSV applies internally for an encoding session depend on various factors, such as GPU generation, driver version and other supplied parameters, such as `preset` and `low_power`. For instance, LookAhead didn't seem to have an effect for me on a Gen9 GPU when combined with `low_power`.

More info on that here: https://github.com/intel/media-delivery/blob/master/doc/quality.rst

**Changes**
- Add options for LookAhead, ExtBrc, B-Pyramid, Adaptive I/B frames and Trellis
- Add extra_hw_frames to input if LookAhead is enabled
- Add calculation of the number of extra_hw_frames in filters (increased by LookAheadDepth) - still needs testing which ones actually need to be increased

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -896,6 +896,16 @@ namespace MediaBrowser.Controller.MediaEncoding
                 }
 
                 args.Append(GetQsvDeviceArgs(QsvAlias));
+
+                if (string.Equals(vidEncoder, "h264_qsv", StringComparison.OrdinalIgnoreCase) && options.EnableIntelLookAheadH264)
+                {
+                    args.Append(" -extra_hw_frames=" + options.IntelLookAheadDepthH264);
+                }
+                else if (string.Equals(vidEncoder, "hevc_qsv", StringComparison.OrdinalIgnoreCase) && options.EnableIntelLookAheadHevc)
+                {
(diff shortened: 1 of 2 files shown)


nyanmisaka:
The degree of support for these advanced features varies by Gen X, driver version, and system environment, so they cannot be directly exposed to users or user will mess up them easily. We must complete ffhwinfo and the correct ffmpeg fallback to continue enabling these advanced features. By then the transcoder needs a major overhaul.

nyanmisaka:
The low-power check box is already very problematic, at least it should not be placed with the basic options. It was introduced to solve the [encoding issue](https://gitlab.freedesktop.org/drm/intel/-/issues/8080) on Jasper Lake processors.

By looking at the media-driver/onevpl/msdk source code different combinations of these parameters under different bit rate controls can easily cause problems, which is why I didn't enable them. Not to mention that there are many outdated driver versions on Windows that may have missing functionality.

No driver capability checks and software fallbacks is a big problem I'm working on. After that I'd be happy to continue optimizing the video quality without compromising stability.

MarkCiliaVincenti (commented):
With so many changes being done to the string 'param' I wonder if it makes sense to switch to using StringBuilder here.

Bond-009:
Closing as stale, no response in 2+ years

PR #9607, opened 2023-04-06
Fix analyzers warnings and suggestions

goremykin:
**Changes**
This PR is mostly about cleanup and fixing analyzers warnings.
Some changes may seem controversial or a matter of taste, so all changes are grouped by type of warning/suggestion (see commit names) to make it easy to rollback any you don't like.

### DIFF
--- a/tests/Jellyfin.Model.Tests/Dlna/StreamBuilderTests.cs
+++ b/tests/Jellyfin.Model.Tests/Dlna/StreamBuilderTests.cs
@@ -162,7 +162,7 @@ namespace Jellyfin.Model.Tests
         [InlineData("Tizen4-4K-5.1", "mkv-vp9-aac-srt-2600k", PlayMethod.DirectPlay)]
         [InlineData("Tizen4-4K-5.1", "mkv-vp9-ac3-srt-2600k", PlayMethod.DirectPlay)]
         [InlineData("Tizen4-4K-5.1", "mkv-vp9-vorbis-vtt-2600k", PlayMethod.DirectPlay)]
-        public async Task BuildVideoItemSimple(string deviceName, string mediaSource, PlayMethod? playMethod, TranscodeReason why = 0, string transcodeMode = "DirectStream", string transcodeProtocol = "")
+        public async Task BuildVideoItemSimple(string deviceName, string mediaSource, PlayMethod? playMethod, TranscodeReason why = default, string transcodeMode = "DirectStream", string transcodeProtocol = "")
         {
             var options = await GetMediaOptions(deviceName, mediaSource);
             BuildVideoItemSimpleTest(options, playMethod, why, transcodeMode, transcodeProtocol);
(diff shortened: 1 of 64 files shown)


tests/Jellyfin.Model.Tests/Dlna/StreamBuilderTests.cs
```diff
@@ -162,7 +162,7 @@ public class StreamBuilderTests
         [InlineData("Tizen4-4K-5.1", "mkv-vp9-aac-srt-2600k", PlayMethod.DirectPlay)]
         [InlineData("Tizen4-4K-5.1", "mkv-vp9-ac3-srt-2600k", PlayMethod.DirectPlay)]
         [InlineData("Tizen4-4K-5.1", "mkv-vp9-vorbis-vtt-2600k", PlayMethod.DirectPlay)]
-        public async Task BuildVideoItemSimple(string deviceName, string mediaSource, PlayMethod? playMethod, TranscodeReason why = (TranscodeReason)0, string transcodeMode = "DirectStream", string transcodeProtocol = "")
+        public async Task BuildVideoItemSimple(string deviceName, string mediaSource, PlayMethod? playMethod, TranscodeReason why = 0, string transcodeMode = "DirectStream", string transcodeProtocol = "")
```
Bond-009, line 165: Maybe `default` would be better than `0`?

Bond-009 (approved):

tests/Jellyfin.Model.Tests/Dlna/StreamBuilderTests.cs
goremykin: @Bond-009, There is a chance to change the behavior of the code in this case as enum starts from 1

Bond-009: `default` for enums should always be `0` AFAIK?

goremykin: Except when explicitly specified :)
You can check the source code of the enum and make sure

Bond-009: Please show me how/where, I can only find this:
> The default value of an enumeration type E is the value produced by expression (E)0, even if zero doesn't have the corresponding enum member.

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/enum

goremykin: You are absolutely right. I thought it sets the first enum value (1), but it sets 0 even if there is such a value.

goremykin: So let's use default in this case

goremykin: @Bond-009, done

1337joe (approved):

PR #9625, opened 2023-04-10
chore: ffmpeg version bump & swap to env for file locations

brettpetch:
**Changes**

Moves to `jellyfin-ffmpeg6` builds and uses `ENV` for setting `JELLYFIN_CACHE_DIR` and `JELLYFIN_CONFIG_DIR` for easier changes by users via environment vars. (not everyone wants to mount `/config` and `/cache`)


crobibero:
As far as I know these dockerfiles aren't actually published anywhere, so changes will also need to be make in the metapackages repo https://github.com/jellyfin/jellyfin-metapackages

brettpetch:
> As far as I know these dockerfiles aren't actually published anywhere, so changes will also need to be make in the metapackages repo https://github.com/jellyfin/jellyfin-metapackages

Doing this mostly for myself as I build from here to run bleeding edge 😅

Shadowghost (approved):
I don't mind these changes but it would be nice if we can bump the Intel stuff to the most recent versions before merging.

Dockerfile
```diff
@@ -78,15 +78,15 @@ RUN dotnet publish Jellyfin.Server --disable-parallel --configuration Release --
 FROM app
 
 ENV HEALTHCHECK_URL=http://localhost:8096/health
+ENV JELLYFIN_CONFIG_DIR=/config
```
lonebyte, line 81: JELLYFIN_DATA_DIR needs to be set instead, as you want to replace `--datadir`. JELLYFIN_CONFIG_DIR would replace the option `--configdir`.

lonebyte (changes requested):

joshuaboniface (approved):
I see no issue here though Ghost's feedback should be addressed.

Gauvino:
> > I see no issue here though Ghost's feedback should be addressed.
> 
> 
> 
> Latest versions should be
> 
> 
> 
> ```dockerfile
> 
> ARG GMMLIB_VERSION=22.3.11
> 
> ARG IGC_VERSION=1.0.15136.4
> 
> ARG NEO_VERSION=23.35.27191.9
> 
> ARG LEVEL_ZERO_VERSION=1.3.27191.9
> 
> ```

Currently reworking this PR with latest version and a little bit more feature, refactoring the build process

brettpetch:
I think that it might be best to have it always pull the most recent. I don't mind writing that instead of doing args.

Gauvino:
> I think that it might be best to have it always pull the most recent. I don't mind writing that instead of doing args.

It would be a good idea I think

Shadowghost:
> I think that it might be best to have it always pull the most recent. I don't mind writing that instead of doing args.

Do not trust the version numbers written in the release description. They are wrong way too often, use the artifacts attached to the compute runtime release if possible.

crobibero:
closing in favor of https://github.com/jellyfin/jellyfin/pull/10603

Issue #2149, opened 2019-12-13
Eliminate All Compiler Warnings In Solution


mark-monteiro:
This is a general ticket to track progress on eliminating all warnings in the solution source code, including enabling missing analyzers and converting warnings to errors in release builds. @Bond-009, as I understand, you have already been working on this for a while. I thought I would start helping out with this task and we (and anyone else) can coordinate progress here.

The table below summarizes the progress on this task at the current HEAD commit on master:

|Project|Warnings|FxCop Analyzer|StyleCop Analyzer|Serilog Analyzer|Multithreading Analyzer|Warnings As Errors|
|-------|--------|--------------|-----------------|----------------|-----------------------|------------------|
|DvdLib|                               <ul><li>[x] </li></ul>|               <ul><li>[ ] </li></ul>|<ul><li>[ ] </li></ul>|<ul><li>[ ] </li></ul>|<ul><li>[ ] </li></ul>|<ul><li>[x] </li></ul>
|Emby.Dlna|                            <ul><li>[x] </li></ul>| <ul><li>[x] </li></ul>|<ul><li>[x] </li></ul>|<ul><li>[x] </li></ul>|<ul><li>[x] </li></ul>|<ul><li>[x] </li></ul>
|Emby.Drawing|                         <ul><li>[x] </li></ul>|               <ul><li>[x] </li></ul>|<ul><li>[x] </li></ul>|<ul><li>[x] </li></ul>|<ul><li>[x] </li></ul>|<ul><li>[x] </li></ul>
|Emby.Naming|                          <ul><li>[x] </li></ul>|               <ul><li>[ ] </li></ul>|<ul><li>[x] </li></ul>|<ul><li>[x] </li></ul>|<ul><li>[x] </li></ul>|<ul><li>[x] </li></ul>
|Emby.Notifications|                   <ul><li>[x] </li></ul>|               <ul><li>[x] </li></ul>|<ul><li>[x] </li></ul>|<ul><li>[x] </li></ul>|<ul><li>[x] </li></ul>|<ul><li>[x] </li></ul>
|Emby.Photos|                          <ul><li>[x] </li></ul>|               <ul><li>[x] </li></ul>|<ul><li>[x] </li></ul>|<ul><li>[x] </li></ul>|<ul><li>[x] </li></ul>|<ul><li>[x] </li></ul>
[...]

PR #9641, opened 2023-04-13
Fix compiler warnings for Emby.Server.Implementations

Daaiid:
Makes changes according to the warnings given on the project. See the separate commit messages for more details. I'd be glad to answer any questions and adjust my changes.

The change I would like to highlight as maybe misguided is the "ConfigureAwait(false)" suppression. I know that async/await is complicated and although I'm not completely unfamiliar with it this change is the one I'm the least confident in.

Partially addresses #2149

### DIFF
--- a/Emby.Server.Implementations/Data/SqliteItemRepository.cs
+++ b/Emby.Server.Implementations/Data/SqliteItemRepository.cs
@@ -205,7 +205,7 @@ namespace Emby.Server.Implementations.Data
         private static readonly string _mediaAttachmentSaveColumnsSelectQuery =
             $"select {string.Join(',', _mediaAttachmentSaveColumns)} from mediaattachments where ItemId=@ItemId";
 
-        private static readonly string _mediaAttachmentInsertPrefix;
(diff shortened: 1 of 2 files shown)


Shadowghost (approved):

Daaiid:
Fix the outstanding merge conflicts. Should be good to go now 👍

barronpm:
Looks like some of the .NET 8 changes wrt. analyzers means that this PR in it's original form doesn't quite work, but some of the changes here are definitely worth including. If you could just include the changes made in [313dac0](https://github.com/jellyfin/jellyfin/pull/9641/commits/313dac0fc330379fdaea3b4553ac5eac580de8b3) and [4232698](https://github.com/jellyfin/jellyfin/pull/9641/commits/423269824cb3d632b547b8e92148094e528c828b) that would be much simpler to review and merge.

Daaiid:
I've reworked this pull request to only include the changes as per @barronpm 's recommendation.
This way it is much easier to review and merge.

Bond-009 (approved):

barronpm (approved):

Issue #9634, opened 2023-04-12
[Issue]: Write Lock Contention Causing UI Freezing


Silvenga:
### Please describe your bug

I'm using a custom [image (silvenga-docker/jellyfin-debug)](https://github.com/silvenga-docker/jellyfin-debug) to get process dumps, but these issues are reproducible on the official image.

During library scans, there appears to be a large amount of lock contention around `GetConnection`. This causes many operations to completely freeze on the UI e.g. loading any media.

![image](https://user-images.githubusercontent.com/2753330/231603310-4f2d94ab-9032-4be7-ada1-1c7acc73b0a0.png)

As shown in the above screenshot, over 55 threads are waiting on a SQLite connection, to either read or write data.

I wonder if it would be possible to:

1. Allow reads to be multi-threaded (it looks like some work was started here?).
2. Allow some kind of interweaving - so instead of a single thread making many operations, limit the amount of work a single transaction can do, to reduce the contention (increase fairness).

Ultimately, my goal here is to improve UI responsiveness during load - for UX sakes.

### Jellyfin Version

Other

### if other:

10.9.0 (version on Master)

### Environment

```markdown
- OS: Ubuntu is the host, but the container is the Official Microsoft image for ASP.NET Core.
- Virtualization: Docker on top of Hyper-V
- Clients:
- Browser:
- FFmpeg Version: Version provided by the official image.
- Playback Method:
- Hardware Acceleration: No
- Plugins:
- Reverse Proxy: nginx
- Base URL: none
- Networking: Bridge 
- Storage: ext4 on top of RBD (Ceph, think block device on the network)
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

srcrist:
Issue should be closed by https://github.com/jellyfin/jellyfin/pull/9643.

PR #9643, opened 2023-04-14
sql connection pool

Bond-009:
Should improve #9634

### DIFF
--- a/Emby.Server.Implementations/Data/BaseSqliteRepository.cs
+++ b/Emby.Server.Implementations/Data/BaseSqliteRepository.cs
@@ -124,7 +124,7 @@ namespace Emby.Server.Implementations.Data
             ReadConnections = new ConnectionPool(ReadConnectionsCount, CreateReadConnection);
 
             // Configuration and pragmas can affect VACUUM so it needs to be last.
-            using (var connection = GetConnection(true))
+            using (var connection = GetConnection())
             {
                 connection.Execute("VACUUM");
             }
--- a/Emby.Server.Implementations/Data/BaseSqliteRepository.cs
+++ b/Emby.Server.Implementations/Data/BaseSqliteRepository.cs
@@ -122,17 +122,16 @@ namespace Emby.Server.Implementations.Data
         {
             WriteConnections = new ConnectionPool(WriteConnectionsCount, CreateWriteConnection);
             ReadConnections = new ConnectionPool(ReadConnectionsCount, CreateReadConnection);
+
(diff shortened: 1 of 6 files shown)


Emby.Server.Implementations/Data/BaseSqliteRepository.cs
```diff
@@ -101,63 +110,115 @@ protected BaseSqliteRepository(ILogger<BaseSqliteRepository> logger)
...
-            WriteConnection.Execute("VACUUM");
+            writeConnection.Execute("VACUUM");
 
-            return new ManagedConnection(WriteConnection, WriteLock);
+            return writeConnection;
+        }
+
+        protected SQLiteDatabaseConnection CreateReadConnection()
+        {
+            var writeConnection = SQLite3.Open(
```
AmbulantRex, line 184:
```suggestion
            var readConnection = SQLite3.Open(
```

Just a minor inconsistency with the naming.

Emby.Server.Implementations/Data/ConnectionPool.cs
```diff
@@ -0,0 +1,77 @@
...
+using SQLitePCL.pretty;
+
+namespace Emby.Server.Implementations.Data;
+
+/// <summary>
+/// A pool of SQLite Database connections.
+/// </summary>
+public sealed class ConnectionPool : IDisposable
+{
+    private readonly BlockingCollection<SQLiteDatabaseConnection> _connections = new();
```
crobibero, line 12: I don't think it _really_ matters but `BlockingCollection` is disposable

crobibero (approved):

Emby.Server.Implementations/Data/ConnectionPool.cs
```diff
@@ -0,0 +1,79 @@
...
+        if (_disposed)
+        {
+            ThrowObjectDisposedException();
+        }
+
+        return new ManagedConnection(_connections.Take(), this);
+
+        void ThrowObjectDisposedException()
+        {
+            throw new ObjectDisposedException(GetType().Name);
```
cvium, line 43: Wouldn't nameof be better? Compile-time vs runtime?

cvium (approved):

Emby.Server.Implementations/Data/ConnectionPool.cs
Bond-009: Since this class is sealed, yes that would be better

PR #9666, opened 2023-04-22
Source SQLite cache_size from an Environment Variable

### LINKED ISSUE #9646: [RFC] Allow cache_size to be set by Operator
Recently, Plex exposed the `cache_size` SQLite option to users, and reportedly this has a significant increase in performance in some deployments.

Right now `Emby.Server.Implementations.Data.SqliteItemRepository.CacheSize` sets the cache size for the SQLite connection. This is currently hard coded to `20_000` pages (about 82MB, assuming a page is 4KiB).

It would be ideal if Jellyfin could also expose this option so that operators could play around with it and see if that helps performance. I would propose, rather than exposing it in the UI (at least for now), allow operators to configure this via an environment variable (e.g. `JELLYFIN_SQLITE_CACHE_SIZE_MB`). Over time, the Jellyfin can see if increasing the default would be useful (or even exposing it in the UI). Other future niceties could include dynamically calculating this value based on restrictions such as Docker memory limits (as already done by the runtime).

I would love to make a PR if this change is acceptable.

Silvenga:
**Changes**

This pr adds support for an operator to modify the [`cache_size` PRAGMA option](https://www.sqlite.org/pragma.html#pragma_cache_size) dynamically. This allows an operator to increase the `cache_size` (currently hardcoded to about 82MiB) to troubleshoot performance or to decrease this if they are running in a memory constrained environment (the SQLite default value is about 2MiB).

I would propose this option as officially unsupported by Jellyfin project for now - e.g. "Here be dragons".

**Issues**

Resolves #9646.

### DIFF
--- a/Emby.Server.Implementations/Data/SqliteItemRepository.cs
+++ b/Emby.Server.Implementations/Data/SqliteItemRepository.cs
@@ -25,6 +25,7 @@ using MediaBrowser.Controller.Entities;
 using MediaBrowser.Controller.Entities.Audio;
 using MediaBrowser.Controller.Entities.Movies;
 using MediaBrowser.Controller.Entities.TV;
+using MediaBrowser.Controller.Extensions;
 using MediaBrowser.Controller.Library;
 using MediaBrowser.Controller.LiveTv;
 using MediaBrowser.Controller.Persistence;
@@ -34,6 +35,7 @@ using MediaBrowser.Model.Entities;
 using MediaBrowser.Model.Globalization;
 using MediaBrowser.Model.LiveTv;
 using MediaBrowser.Model.Querying;
+using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
 using SQLitePCL.pretty;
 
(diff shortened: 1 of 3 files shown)


Emby.Server.Implementations/Data/SqliteItemRepository.cs
```diff
@@ -6141,6 +6141,21 @@ private MediaAttachment GetMediaAttachment(IReadOnlyList<ResultSetValue> reader)
 
 #nullable enable
 
+        private static int GetCacheSizeOrDefault()
+        {
+            // https://www.sqlite.org/pragma.html#pragma_cache_size
+            // Positive values == pages to cache (by default, each pasge is 4KiB).
+            // Negative values == calculate number of pages to cache under this number (in KiB).
+
+            if (Environment.GetEnvironmentVariable("JELLYFIN_SQLITE_CACHE_SIZE") is { } environmentVariable
```
Bond-009, line 6150: We use `using Microsoft.Extensions.Configuration;` instead if getting the env var directly.
You can set the default value in `Emby.Server.Implementations//ConfigurationOptions.cs`

Bond-009 (changes requested):

Emby.Server.Implementations/Data/SqliteItemRepository.cs
Silvenga: So set this in the `ctor` instead of within this static context?

Silvenga:
> Parameters or locals of type 'ReadOnlySpan<char>' cannot be declared in async methods or async lambda expressions.

I hope that was just a flaky run...

crobibero (approved):

crobibero:
> Any suggestions on how to correct the bad tests? - Looks like the faker is faking bad data.

Just ignore them for now

Bond-009 (approved):

pranksta02:
Hi, is the environment variable to set for this SqliteCacheSizeKey ?

Silvenga:
@pranksta02 

> Hi, is the environment variable to set for this SqliteCacheSizeKey ?

It should be `JELLYFIN_SqliteCacheSizeKey` now.

Bond-009:
I think it is `JELLYFIN_sqlite__cacheSize`

Silvenga:
> I think it is `JELLYFIN_sqlite__cacheSize`

Oops. Thanks for the correction.

PR #9044, opened 2023-01-08
Fine tune DB settings

Bond-009:
I will be running this on my own server for a while to make sure it's fine

PR #9667, opened 2023-04-22
Increase journal_size_limit Defaults

Silvenga:
**Changes**

Per the [documentation](https://www.sqlite.org/pragma.html#pragma_journal_size_limit) for `journal_size_limit`, when set to `0`, SQLite will "always truncate rollback journals and WAL files to their minimum size".

This pr changes Jelly's defaults from `0` to `134_217_728` (aka, 128MiB).

While I have no performance tests, a value of `0` seems problematic given we are currently using the `journal_mode = WAL` option (since one of the purposes of using WAL is to prevent needing to remove/truncate the journal after each operation).

**Issues**

Related: #9044

### DIFF
--- a/Emby.Server.Implementations/Data/BaseSqliteRepository.cs
+++ b/Emby.Server.Implementations/Data/BaseSqliteRepository.cs
@@ -73,9 +73,10 @@ namespace Emby.Server.Implementations.Data
 
         /// <summary>
         /// Gets the journal size limit. <see href="https://www.sqlite.org/pragma.html#pragma_journal_size_limit" />.
+        /// The default (-1) is overriden to prevent unconstrained WAL size, as reported by users.
         /// </summary>
         /// <value>The journal size limit.</value>
-        protected virtual int? JournalSizeLimit => 0;
+        protected virtual int? JournalSizeLimit => 134_217_728; // 128MiB
 
         /// <summary>
         /// Gets the page size.


Silvenga:
@Bond-009, I'm not expert in SQLite to be sure. In your PR #9044 I wonder if `0` was a possible typo, or if you had another concern behind setting this explicitly?

Bond-009:
This was to prevent the infinitely increasing WAL issue that users experienced the first time we tried enabling WAL mode.

Silvenga:
I wonder if we could set this, to say, 100mb or so. That would maintain performance for most operations while preventing unconstrained WAL size.

Silvenga:
@Bond-009 I updated this PR from `-1` -> `134_217_728` (128MiB). The idea is that we want to perform append operations as much as possible since truncation is somewhat costly on newer filesystems that use COW, copy-on-write, semantics. Filesystems such as Window's new file system ReFS, BTRFS (likely the future replacement of ext4), my storage subsystem Ceph, etc.

Do you think that would work for the unconstrained WAL concern you mentioned?

Silvenga:
@Bond-009

Bond-009 (approved):
Size didn't balloon in testing :+1:

Silvenga:
Thanks @Bond-009 for looking into that!

Shadowghost (approved):

PR #9669, opened 2023-04-23
Fix subtitle encoder if webvtt is requested

sleepycatcoding:
**Changes**
Properly convert subtitles to webvtt if requested.

**Issues**
Fixes jellyfin/jellyfin-android#1028


nielsvanvelzen:
The app is requesting the wrong URL. The extension for WebVTT is `.vtt` and the app is requesting `.webvtt`. So IMO this is not a bug in the server.

sleepycatcoding:
Isn't then requesting `.subrip` also a bug in the app?

Shadowghost (approved):

sleepycatcoding:
> The app is requesting the wrong URL. The extension for WebVTT is `.vtt` and the app is requesting `.webvtt`. So IMO this is not a bug in the server.

This is a bug in the server, because the app uses the deliveryUrl from the server that is already wrong.

The URL is generated here:
https://github.com/jellyfin/jellyfin/blob/6e14f26ec85b8fad3c9412d395f0856a27421852/MediaBrowser.Model/Dlna/StreamInfo.cs#L894-L902

The subtitleProfile.Format is `webvtt`. A better fix probably would be to use file extensions instead of format in the URL.

cvium (approved):

aseering:
I'm running into this as well.  Is the patch in need of any tweaks or further changes?  Happy to help.

Issue #6052, opened 2021-05-15
Syncplay GroupUpdate class not properly added in OpenAPI spec


nielsvanvelzen:
**Describe the bug**

The GroupUpdate class uses a generic for the `Data` property. This might be cool to use inside the server code but it doesn't work for an API model. Swashbuckle is unable to properly create a type for it and ends up just not giving it a type at all... This causes the SDK's/apiclients to generate a model with an `any` type or something similar. While this may work in loosely typed languages like JavaScript it won't work in Kotlin/Java and therefor we're unable to implement SyncPlay in both the Android TV and Android mobile apps.

One potential fix I can think of is to remove the GroupUpdate class and replace it with individual classes for each type. This would be much easier to parse in the SDK. Unfortunately this would change the schema slightly and additional changes in existing Syncplay clients need to be made.

nielsvanvelzen:
Should be fixed for 10.9

PR #9682, opened 2023-04-24
Add all websocket messages to generated openapi spec

crobibero:
Alternative to #9491

This doesn't add any useful information to redoc / swagger but it will allow generated clients to have the models defined.

[openapi.10.9.json.txt](https://github.com/jellyfin/jellyfin/files/11311535/openapi.json.txt)
[openapi.10.8.json.txt](https://github.com/jellyfin/jellyfin/files/11344163/openapi.10.8.json.txt)

This also adds a default value for the generated MessageType

eg:
```json
{
	"MessageType": {
		"allOf": [{
			"$ref": "#/components/schemas/SessionMessageType"
		}],
		"description": "The different kinds of messages that are used in the WebSocket api.",
		"default": "ActivityLogEntryStop",
		"readOnly": true
	}
}
```


nielsvanvelzen:
I'll try to look at these changes later this week when I have some more time but I believe this fixes #6052 which would make it possible to add SyncPlay to our Kotlin SDK based clients.

nielsvanvelzen:
Based on the current 10.8 spec+PR with compile issues:

- Some .NET types have leaked into the spec (Module, Assembly, ConstructorInfo, CustomAttributeData etc.).
- There is an additional route "/Test" which uses an operation id that is already in use:
  ```

    "/Test": {
      "get": {
        "tags": [
          "Test"
        ],
        "operationId": "Get",
        "responses": {
          "200": {
            "description": "Success",
            "content": {
              "application/json": {
                "schema": {
                  "type": "string"
                }
              },
              "application/json; profile=\"CamelCase\"": {
                "schema": {
                  "type": "string"
                }
              },
              "application/json; profile=\"PascalCase\"": {
                "schema": {
                  "type": "string"
                }
              }
            }
          }
        }
      }
    },
  ```
  The operation id is "Get" which (in 10.8, I fixed it for 10.9) is used by the search hint endpoint. Duplicate operation ids cause the SDK compilation to fail.
  I'm expecting this endpoint is from your local branch or something and not related to this PR.

- The SDK currently uses manually written classes for the WebSocket messages and one thing I did was adding an interface (either OutgoingSocketMessage, IncomingSocketMessage or both). With the information in the spec right now I can't do that. I can think of two ways to fix that:
  1. Add a new property `x-jellyfin-type` with values `ws-in`, `ws-out` and `ws-shared`.
  2. Add a new schema `IncomingSocketMessage` with a value of `allOf: []` referencing the actual objects

crobibero:
@nielsvanvelzen This spec should fix all of your above points

  [openapi.10.8.json.txt](https://github.com/jellyfin/jellyfin/files/11348410/openapi.10.8.json.txt)

nielsvanvelzen:
This looks promising! I've found a bug in the Kotlin SDK generator that needs to be fixed and polymorphic (de)serialization should be implemented but this spec looks like it could fulfill that role and replace my manually written WebSocket classes.

For `InboundWebsocketMessage` and `OutboundWebsocketMessage` I believe it should use `oneOf` when reading the JSON schema docs: https://json-schema.org/understanding-json-schema/reference/combining.html?highlight=anyof#oneof

I'm also thinking we may want to remove the `ServerId` property from the spec so we can eventually drop it entirely. It doesn't make sense because the client should already know the id of the server it literally just connected to.

crobibero:
Suggestions applied: [openapi.10.8.json.txt](https://github.com/jellyfin/jellyfin/files/11357522/openapi.10.8.json.txt)

Jellyfin.Server/Filters/AdditionalModelFilter.cs
```diff
@@ -36,17 +35,38 @@ public AdditionalModelFilter(IServerConfigurationManager serverConfigurationMana
...
+                            && t != typeof(WebSocketMessageInfo))
+                .ToList();
+
+            context.SchemaRepository.AddDefinition(
+                nameof(InboundWebSocketMessage),
+                new OpenApiSchema
+                {
+                    Type = "object",
+                    Description = "Represents the list of possible inbound websocket types",
+                    AllOf = websocketTypes
```
nielsvanvelzen, line 52: Should also be oneOf

Jellyfin.Server/Filters/AdditionalModelFilter.cs
```diff
@@ -36,17 +35,38 @@ public AdditionalModelFilter(IServerConfigurationManager serverConfigurationMana
...
+            context.SchemaRepository.AddDefinition(
+                nameof(InboundWebSocketMessage),
+                new OpenApiSchema
+                {
+                    Type = "object",
+                    Description = "Represents the list of possible inbound websocket types",
+                    OneOf = websocketTypes
+                        .Where(t => typeof(IInboundWebSocketMessage).IsAssignableFrom(t))
+                        .Select(t => context.SchemaGenerator.GenerateSchema(t, context.SchemaRepository))
+                        .ToList()
```
nielsvanvelzen, line 55: Should also add this to the JSON output (for both inbound and outbound):

```
        "discriminator": {
          "propertyName": "MessageType"
        },
```

so it would look like this:
```
"InboundWebSocketMessage": {
        "type": "object",
        "oneOf": [
          {
            "$ref": "#/components/schemas/KeepAliveMessage"
          },
          {
            "$ref": "#/components/schemas/ActivityLogEntryStartMessage"
          },
          {
            "$ref": "#/components/schemas/ActivityLogEntryStopMessage"
          },
          {
            "$ref": "#/components/schemas/ScheduledTasksInfoStartMessage"
          },
          {
            "$ref": "#/components/schemas/ScheduledTasksInfoStopMessage"
          },
          {
            "$ref": "#/components/schemas/SessionsStartMessage"
          },
          {
            "$ref": "#/components/schemas/SessionsStopMessage"
          }
        ],
        "discriminator": {
          "propertyName": "MessageType"
        },
        "description": "Represents the list of possible inbound websocket types"
      },
```

as per my previous comment

crobibero:
New 10.8 spec with discriminators: [openapi.10.8.json.txt](https://github.com/jellyfin/jellyfin/files/11472189/openapi.10.8.json.txt)

nielsvanvelzen:
Ok last request and then I'm satisfied. Can we add another schema for WebSocketMessage that looks somewhat like this:

```

      "WebSocketMessage": {
        "type": "object",
        "oneOf": [
          {
            "$ref": "#/components/schemas/InboundWebSocketMessage"
          },
          {
            "$ref": "#/components/schemas/OutboundWebSocketMessage"
          }
        ]
      }
```

That way the SDK will generate an interface used by the inbound & outbound interfaces for the `messageId` property.

crobibero:
As you wish [openapi.10.8.json.txt](https://github.com/jellyfin/jellyfin/files/11472530/openapi.10.8.json.txt)

nielsvanvelzen (approved):

Shadowghost (approved):

PR #9714, opened 2023-05-01
Add auto close of an inactive session after X minutes

herby2212:
**Changes**
Added a internal timer in the `SessionManager` to check at `1` minute intervals for sessions that are paused longer then `10` minutes. For this a new property got added to the `SessionInfo` called `LastPausedDate` which contains the timestamp the session was last paused.
When a session passes this check it gets closed automatically without any user intervention needed.

The threshold can be configured via the new `InactiveSessionThreshold` property as part of the `ServerConfiguration`. Setting this property to `0` will disable the check and the auto close function.

**Features**
Implements feature [301](https://features.jellyfin.org/posts/301/option-to-kill-a-stream-from-a-user).

### DIFF
--- a/Jellyfin.Api/Helpers/StreamingHelpers.cs
+++ b/Jellyfin.Api/Helpers/StreamingHelpers.cs
@@ -191,6 +191,11 @@ public static class StreamingHelpers
             state.OutputAudioBitrate = encodingHelper.GetAudioBitrateParam(streamingRequest.AudioBitRate, streamingRequest.AudioCodec, state.AudioStream, state.OutputAudioChannels) ?? 0;
         }
 
+        if (outputAudioCodec.StartsWith("pcm_", StringComparison.Ordinal))
+        {
+            containerInternal = ".pcm";
+        }
+
         state.OutputAudioCodec = outputAudioCodec;
         state.OutputContainer = (containerInternal ?? string.Empty).TrimStart('.');
         state.OutputAudioChannels = encodingHelper.GetNumAudioChannelsParam(state, state.AudioStream, state.OutputAudioCodec);
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -6249,6 +6249,12 @@ namespace MediaBrowser.Controller.MediaEncoding
                 audioTranscodeParams.Add("-acodec " + GetAudioEncoder(state));
             }
 
+            if (GetAudioEncoder(state).StartsWith("pcm_", StringComparison.Ordinal))
+            {
+                audioTranscodeParams.Add(string.Concat("-f ", GetAudioEncoder(state).AsSpan(4)));
+                audioTranscodeParams.Add("-ar " + state.BaseRequest.AudioBitRate);
+            }
+
             if (!string.Equals(outputCodec, "opus", StringComparison.OrdinalIgnoreCase))
             {
                 // opus only supports specific sampling rates


crobibero (changes requested):
This needs to be configurable (both pause time and to disable this feature)

herby2212:
> This needs to be configurable (both pause time and to disable this feature)

Agreed. Is there documentation or a example of used classes regarding endpoints for user configuration?

herby2212:
> This needs to be configurable (both pause time and to disable this feature)

Option to disable the function and configure the threshold got added. Please see the updated PR description for more details.

Emby.Server.Implementations/Session/SessionManager.cs
```diff
@@ -564,9 +580,18 @@ private List<User> GetUsers(SessionInfo session)
             return users;
         }
 
-        private void StartIdleCheckTimer()
+        private void StartCheckTimers()
```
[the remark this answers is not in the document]
herby2212: I did the splitup on the `Stop` method as they are handled differently and are stopped on different occasions.
However as the callback functions are looking on the same `sessions` and have the same start requirements I kept them together.

Emby.Server.Implementations/Session/SessionManager.cs
```diff
@@ -613,10 +647,52 @@ private async void CheckForIdlePlayback(object state)
...
+                StopIdleCheckTimer();
+            }
+        }
+
+        private async void CheckForInactiveSteams(object state)
+        {
+            var pausedSessions = Sessions.Where(i =>
+                (i.NowPlayingItem is not null) &&
+                (i.PlayState.IsPaused == true) &&
+                (i.LastPausedDate is not null)).ToList();
```
crobibero, line 661:
```suggestion
            var pausedSessions = Sessions
                .Where(i =>
                    i.NowPlayingItem is not null
                    && i.PlayState.IsPaused
                    && i.LastPausedDate is not null)
                .ToList();
```

Emby.Server.Implementations/Session/SessionManager.cs
```diff
@@ -397,6 +404,15 @@ private async Task UpdateNowPlayingItem(SessionInfo session, PlaybackProgressInf
                 session.LastPlaybackCheckIn = DateTime.UtcNow;
             }
 
+            if (info.IsPaused && session.LastPausedDate.HasValue == false)
```
crobibero, line 407:
```suggestion
            if (info.IsPaused && session.LastPausedDate is null)
```

MediaBrowser.Model/Configuration/ServerConfiguration.cs
```diff
@@ -157,6 +157,12 @@ public ServerConfiguration()
         /// <value>The remaining time in minutes.</value>
         public int MaxAudiobookResume { get; set; } = 5;
 
+        /// <summary>
+        /// Gets or sets the threshold in minutes after a inactive session gets closed automatically.
+        /// </summary>
+        /// <value>The close inactive session threshold in minutes.</value>
+        public int InactiveSessionThreshold { get; set; } = 10;
```
crobibero, line 164: I think this should default to 0 and be opt-in.

herby2212: Agreed.

herby2212: I'm not sure if it should be disabled by default. As having it enabled by default has no negative effect. It saves server resources and if a session gets killed for a user he can just continue at the point he paused, as sessions only get killed if they are on pause, without any informations lost.

Emby.Server.Implementations/Session/SessionManager.cs
herby2212: It this personal preference or referenced by a code style guide regarding the indentation? Because while looking through the code (both Emby and Jellyfin packages) there is no clear style for a case like this.

I agree with your points, but would prefer to keep `.Where(i =>` in line 658 together with `Sessions`.

Shadowghost: The proposed style changes is what we try to use in new code. It's not documented anywere though, just preference.

crobibero: Correct- it's just preference. Moving `.Where` to L658 is fine by me

MediaBrowser.Model/Configuration/ServerConfiguration.cs
herby2212: I will keep it enabled by default as there have been no negative points mentioned for having it enabled by default.
If there are other points please un-resolve the conversation.

Emby.Server.Implementations/Session/SessionManager.cs
```diff
@@ -613,10 +647,53 @@ private async void CheckForIdlePlayback(object state)
...
+                    && i.LastPausedDate is not null)
+                .ToList();
+
+            if (pausedSessions.Count > 0)
+            {
+                var inactiveSessions = pausedSessions.Where(i => (DateTime.UtcNow - i.LastPausedDate).Value.TotalMinutes > _config.Configuration.InactiveSessionThreshold).ToList();
+
+                foreach (var session in inactiveSessions)
+                {
+                    _logger.LogDebug("Session {0} has been inactive for {1} minutes. Stopping it.", session.Id, _config.Configuration.InactiveSessionThreshold);
```
1337joe, line 670: Named placeholders are preferred, at least for new code.

```suggestion
                    _logger.LogDebug("Session {Session} has been inactive for {InactiveTime} minutes. Stopping it.", session.Id, _config.Configuration.InactiveSessionThreshold);
```

Emby.Server.Implementations/Session/SessionManager.cs
```diff
@@ -613,10 +647,53 @@ private async void CheckForIdlePlayback(object state)
...
+                            {
+                                Command = PlaystateCommand.Stop,
+                                ControllingUserId = session.UserId.ToString(),
+                                SeekPositionTicks = session.PlayState?.PositionTicks
+                            },
+                            CancellationToken.None).ConfigureAwait(true);
+                    }
+                    catch (Exception ex)
+                    {
+                        _logger.LogDebug(ex, "Error calling SendPlaystateCommand for stopping inactive session {0}.", session.Id);
```
1337joe, line 687:
```suggestion
                        _logger.LogDebug(ex, "Error calling SendPlaystateCommand for stopping inactive session {Session}.", session.Id);
```

1337joe (changes requested):

Emby.Server.Implementations/Session/SessionManager.cs
herby2212: Implemented.

herby2212:
Continued in PR #10357 to resolve merge conflicts and update to current master version.

Issue #2424, opened 2020-02-18
Support multiple codecs in Transcoding Profiles


PrplHaz4:
**Describe the bug**
Currently if you provide a list of codecs in the Transcoding Profile of a specific device, only the first one will be used for every transcode.  This reduces the likelihood of media being remuxed when possible and will force an actual transcode when it shouldn't be necessary.

**To Reproduce**

1. Play media with AAC audio on Samsung TV through DLNA
2. Audio is transcoded to AC3 when played (as that is what is in the Samsung TV profile)
3. Add AAC to Transcoding Profile for Samsung TV (`audio: ac3,aac`)
4. Audio is transcoded to AC3 instead of being `copy`d to AAC by ffmpeg

**Expected behavior**

Source codec is compared to a list of supported transcode codecs and if a stream copy is possible, that codec is used instead.

**Logs**

**Screenshots**

**System (please complete the following information):**
 - OS: Docker
 - Browser: Firefox
 - Jellyfin Version: 10.5-dev
 - Installed Plugins: none
 - Reverse proxy: none

**Additional context**
- Observed this behavior when investigating #2334 and #2409
- Not sure if this behavior is the same for device profiles included in clients (ie, web, android tv...etc) or just applies to DLNA clients.

Artiume:
So I'm having a similar issue on iPads. It's transcoding livetv to ac3 and isn't playing audio

JustAMan:
I'm pretty sure it's not a bug but an enhancement, so relabelling.

Artiume:
Could we switch the priority from ac3 to aac? I think either way priority is listed, devices will be impacted

PrplHaz4:
> 
> 
> Could we switch the priority from ac3 to aac? I think either way priority is listed, devices will be impacted

Changing priority for DLNA devices definitely impact things because for transcoding profiles it seems that priority isn't really a thing - the first one is the only one ever chosen (that's what this issue is).  For iPad/Safari it's probably in the `browserdeviceprofile.js` somewhere - which MAY behave differently from the DLNA profiles (regarding priority/multiple codecs) ...

sevenrats:
this stale issue references an obsolete unstable branch.

srcrist:
Issue closed by https://github.com/jellyfin/jellyfin/pull/7325

PR #9723, opened 2023-05-02
Fix multiple codec checking in CodecProfiles conditions

dmitrylyzo:
`ContainsAnyCodec` checks if its codec list contains the specified one.
#9411 limited this check to one codec per codec profile.
_However, I haven't seen us use more than one._

**Changes**
Allow (bring back) a list of codecs in `CodecProfiles`.

**Issues**
N/A

### DIFF
--- a/MediaBrowser.Model/Dlna/StreamBuilder.cs
+++ b/MediaBrowser.Model/Dlna/StreamBuilder.cs
@@ -746,11 +746,10 @@ namespace MediaBrowser.Model.Dlna
 
                     if (ContainerProfile.ContainsContainer(videoCodecs, item.VideoStream?.Codec))
                     {
-                        var videoCodec = transcodingProfile.VideoCodec;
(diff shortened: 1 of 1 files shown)


nyanmisaka (approved):
It fixes the original issue and has no side effects. Thanks!

Bond-009 (approved):

PrplHaz4:
I think this will also fix https://github.com/jellyfin/jellyfin/issues/2424 but I'm not sure I have the ability to test anymore.

dmitrylyzo:
> Tests are failing though.

I don't remember if they worked on the release branch before. 😅 
Could it be because 10.8 uses .NET 6 instead of 7?

> I think this will also fix #2424 but I'm not sure I have the ability to test anymore.

It was probably already fixed by #7325 and some subsequent PRs.
This one returns the ability to apply the same condition to multiple codecs which we don't use yet.

Issue #9665, opened 2023-04-22
[Question] Now Playing API endpoint


SamJongenelen:
Hey guys, I've been using Jellyfin with a lot of pleasure. I've connected my HomeAssistant to the REST API of JF, and was surprised to find only the `sessions` endpoint exists to gather info on who is now playing.  Of course, I was able to filter the Session response, but still I found it hard to combine different states (e.g. nothing playing but active session, playback started') and also, (the response is enormous :) ) into a nice sensor in HomeAssistant.

As I am a dotnet developer myself, I was wondering if a endpoint like `sessions/nowplaying`, that would return a simple object with something like Username, NowPlaying item name, would be accepted the code base. There might be a reason it isn't already. If it is oké - I can probably make it.

Basically the implementation could filter and project output of the `SessionManager`

crobibero:
Personally I don’t see the use in creating a simpler endpoint if all of the information is already returned by an existing (but complex) endpoint.

nielsvanvelzen:
Adding a new parameter to the existing endpoint to filter only sessions currently playing media might be a nice addition.

SamJongenelen:
Yeah i agree it maybe shouldn't be added, hence i ask here. 

Regarding the extra param: my 'problem' lies with the complex return type. It's hard to create a single HomeAssistant rest sensor that answers my question 'who is playing, what are they playing, and what is the bandwith consumption (based on content info)'. This is because the return object is complex and pretty big (it is correctly filled of course)

nielsvanvelzen:
Doesn't the Jellyfin integration in Home Assistant provide this information already?
https://www.home-assistant.io/integrations/jellyfin/

SamJongenelen:
Unfortunately not, it provides a media_player for all sessions as additional functionality indeed. I think i could use this, but I dont want the primary function of the integration - to control media - in HomeAssistant, I use it for network

nielsvanvelzen:
Our API is already quite bloated. Personally I think we shouldn't add another endpoint if an existing one already provides the necessary information (although with some extra bytes).

Extending the endpoint to allow additional filtering should be fine, but changing the response to return less data purely because that makes it easier to work with in Home Assistant feels like the wrong approach to me.

SamJongenelen:
Thanks for the feedback :) I personally think the SessionController isn't that big - but yes it is another addition.

Regarding the ' for home assistant '  - in Jellyfin Admin page i also wouldn't mind seeing only 'now playing' sessions.

PR #9726, opened 2023-05-03
API Sessions endpoint, added `nowPlaying` querystring param

SamJongenelen:
Only return sessions that have an item currently playing, when using optional parameter 'nowPlaying' on the `Sessions` endpoint.

This was a suggested workaround for my issue #9665.

**Changes**
Instead of adding extra endpoint,
querystringparam `nowPlaying` is added to `Sessions` endpoint,
to filter 'NowPlaying' sessions only.

**Issues**
https://github.com/jellyfin/jellyfin/issues/9665

### DIFF
--- a/Jellyfin.Api/Controllers/SessionController.cs
+++ b/Jellyfin.Api/Controllers/SessionController.cs
@@ -53,6 +53,7 @@ public class SessionController : BaseJellyfinApiController
     /// <param name="controllableByUserId">Filter by sessions that a given user is allowed to remote control.</param>
     /// <param name="deviceId">Filter by device Id.</param>
     /// <param name="activeWithinSeconds">Optional. Filter by sessions that were active in the last n seconds.</param>
+    /// <param name="nowPlaying">Optional. Filter by sessions that are currently playing content.</param>
     /// <response code="200">List of sessions returned.</response>
     /// <returns>An <see cref="IEnumerable{SessionInfo}"/> with the available sessions.</returns>
     [HttpGet("Sessions")]
@@ -61,7 +62,8 @@ public class SessionController : BaseJellyfinApiController
     public ActionResult<IEnumerable<SessionInfo>> GetSessions(
         [FromQuery] Guid? controllableByUserId,
         [FromQuery] string? deviceId,
-        [FromQuery] int? activeWithinSeconds)
+        [FromQuery] int? activeWithinSeconds,
+        [FromQuery] bool? nowPlaying)
     {
         var result = _sessionManager.Sessions;
 
@@ -108,6 +110,11 @@ public class SessionController : BaseJellyfinApiController
 
                 return true;
             });
+        }
+
+        if (nowPlaying.GetValueOrDefault())
+        {
+            result = result.Where(i => i.NowPlayingItem != null);
         }
 
         return Ok(result);
--- a/Jellyfin.Api/Controllers/SessionController.cs
+++ b/Jellyfin.Api/Controllers/SessionController.cs
@@ -112,9 +112,11 @@ public class SessionController : BaseJellyfinApiController
             });
         }
 
-        if (nowPlaying.GetValueOrDefault())
+        if (nowPlaying.HasValue)
         {
-            result = result.Where(i => i.NowPlayingItem != null);
+            result = nowPlaying.Value
+                ? result.Where(i => i.NowPlayingItem != null)
+                : result.Where(i => i.NowPlayingItem == null);
         }
 
         return Ok(result);
--- a/tests/Jellyfin.Server.Integration.Tests/Controllers/SessionControllerTests.cs
+++ b/tests/Jellyfin.Server.Integration.Tests/Controllers/SessionControllerTests.cs
@@ -25,15 +25,17 @@ public class SessionControllerTests : IClassFixture<JellyfinApplicationFactory>
         Assert.Equal(HttpStatusCode.NotFound, response.StatusCode);
     }
 
-    [Fact]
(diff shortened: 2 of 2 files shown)


Jellyfin.Api/Controllers/SessionController.cs
```diff
@@ -61,7 +62,8 @@ public class SessionController : BaseJellyfinApiController
     public ActionResult<IEnumerable<SessionInfo>> GetSessions(
         [FromQuery] Guid? controllableByUserId,
         [FromQuery] string? deviceId,
-        [FromQuery] int? activeWithinSeconds)
+        [FromQuery] int? activeWithinSeconds,
+        [FromQuery] bool? nowPlaying)
```
nielsvanvelzen, line 66: I would expect `nowPlaying=false` to only return sessions that are not playing anything but the current implementation doesn't differ between omitting the parameter and setting it to false.

tests/Jellyfin.Server.Integration.Tests/Controllers/SessionControllerTests.cs
```diff
@@ -24,4 +24,16 @@ public async Task GetSessions_NonExistentUserId_NotFound()
...
+    [Fact]
+    public async Task GetSessions_NowPlaying_Ok()
+    {
+        var client = _factory.CreateClient();
+        client.DefaultRequestHeaders.AddAuthHeader(_accessToken ??= await AuthHelper.CompleteStartupAsync(client).ConfigureAwait(false));
+
+        using var response = await client.GetAsync("Sessions?nowPlaying=true").ConfigureAwait(false);
+
+        Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+        Assert.Equal("[]", await response.Content.ReadAsStringAsync());
```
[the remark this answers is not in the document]
SamJongenelen: Yes, very true. I doubted to specifically test this. I chose only to validate that it does not break (even if nothing is playing) in the integration test project, and there seem to be no unittests (at least not in solution).

Suggestions?

Jellyfin.Api/Controllers/SessionController.cs
SamJongenelen: I had not thought of that angle, will implement

tests/Jellyfin.Server.Integration.Tests/Controllers/SessionControllerTests.cs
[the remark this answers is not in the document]
SamJongenelen: I've added all 3 combinations and simplified assert to only assert `Ok`

Jellyfin.Api/Controllers/SessionController.cs
SamJongenelen: updated the code

tests/Jellyfin.Server.Integration.Tests/Controllers/SessionControllerTests.cs
```diff
@@ -24,4 +24,18 @@ public async Task GetSessions_NonExistentUserId_NotFound()
         using var response = await client.GetAsync($"Session/Sessions?userId={Guid.NewGuid()}").ConfigureAwait(false);
         Assert.Equal(HttpStatusCode.NotFound, response.StatusCode);
     }
+
+    [Theory]
+    [InlineData("?nowPlaying=false")]
+    [InlineData("?nowPlaying=true")]
+    [InlineData("")]
+    public async Task GetSessions_NowPlaying_Ok(string querystring)
```
crobibero, line 32: This test is the same as testing for an invalid query parameter since the framework doesn't error on unknown query parameters. To actually test this you would need to set up the list of sessions to include playing & non playing ones.

SamJongenelen: Yes. I wouldnt mind to, but i am unfamiliar with the codebase still, and dont know how to do that.

Shall i remove the testcase?

JPVenson: The testcase itself is behaviour by the framework.

A test that you really get only active sessions when `nowPlaying=true` is in order

JPVenson (changes requested):

Jellyfin.Api/Controllers/SessionController.cs
```diff
@@ -106,16 +111,12 @@ public ActionResult<IEnumerable<SessionInfo>> GetSessions(
...
-            result = result.Where(i => i.UserId.IsEmpty() || i.ContainsUser(User.GetUserId()));
-        }
 
-        if (activeWithinSeconds.HasValue && activeWithinSeconds.Value > 0)
+        if (nowPlaying.HasValue)
         {
-            var minActiveDate = DateTime.UtcNow.AddSeconds(0 - activeWithinSeconds.Value);
-            result = result.Where(i => i.LastActivityDate >= minActiveDate);
+            result = nowPlaying.Value
+                ? result.Where(i => i.NowPlayingItem != null)
```
Bond-009, line 118:
```suggestion
                ? result.Where(i => i.NowPlayingItem is not null)
```

Jellyfin.Api/Controllers/SessionController.cs
```diff
@@ -106,16 +111,12 @@ public ActionResult<IEnumerable<SessionInfo>> GetSessions(
...
+                : result.Where(i => i.NowPlayingItem == null);
```
Bond-009, line 119:
```suggestion
                : result.Where(i => i.NowPlayingItem is null);
```

Bond-009 (approved):

PR #9746, opened 2023-05-09
Fix access to playlists not created by a user

Shadowghost:
**Changes**
* Introduce a new property `OpenAccess` to allow read access to the playlist for all users
* Automatically set `OpenAccess` to `true` if the playlist originates from the file system and was not created by any user
* Adopt playlist migration to handle those cases
* Add `403` response code documentation to the `PlaylistController` endpoints

### DIFF
--- a/Jellyfin.Api/Controllers/PlaylistsController.cs
+++ b/Jellyfin.Api/Controllers/PlaylistsController.cs
@@ -64,12 +64,15 @@ public class PlaylistsController : BaseJellyfinApiController
     /// <param name="userId">The user id.</param>
     /// <param name="mediaType">The media type.</param>
     /// <param name="createPlaylistRequest">The create playlist payload.</param>
+    /// <response code="200">Playlist created.</response>
+    /// <response code="403">User does not have permission to create playlists.</response>
     /// <returns>
     /// A <see cref="Task" /> that represents the asynchronous operation to create a playlist.
     /// The task result contains an <see cref="OkResult"/> indicating success.
     /// </returns>
     [HttpPost]
     [ProducesResponseType(StatusCodes.Status200OK)]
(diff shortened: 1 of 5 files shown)


nielsvanvelzen (approved):

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -102,9 +102,11 @@ public class PlaylistsController : BaseJellyfinApiController
     /// <param name="ids">Item id, comma delimited.</param>
     /// <param name="userId">The userId.</param>
     /// <response code="204">Items added to playlist.</response>
+    /// <response code="401">Unauthorized to add item to playlist.</response>
     /// <returns>An <see cref="NoContentResult"/> on success.</returns>
     [HttpPost("{playlistId}/Items")]
     [ProducesResponseType(StatusCodes.Status204NoContent)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
```
crobibero, line 109: I think this should be `403` instead of `401`

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -144,7 +147,7 @@ public class PlaylistsController : BaseJellyfinApiController
     /// <param name="playlistId">The playlist id.</param>
     /// <param name="entryIds">The item ids, comma delimited.</param>
     /// <response code="204">Items removed.</response>
-    /// <response code="401">Unauthorized to get playlist.</response>
+    /// <response code="403">User does not have permission to get playlist.</response>
```
crobibero, line 150: L154 should also be 403

crobibero (approved):

PR #9747, opened 2023-05-09
Fix FirstTimeSetupHandler not failing on invalid user if not in setup mode

Shadowghost:
`FirstTimeSetupHandler` did not prevent unauthed access after checking if the server is in setup mode.

This PR fixes that and adds 401 response annotations to applicable `SystemController` endpoints.

### DIFF
--- a/Jellyfin.Api/Auth/FirstTimeSetupPolicy/FirstTimeSetupHandler.cs
+++ b/Jellyfin.Api/Auth/FirstTimeSetupPolicy/FirstTimeSetupHandler.cs
@@ -38,7 +38,15 @@ namespace Jellyfin.Api.Auth.FirstTimeSetupPolicy
                 return Task.CompletedTask;
             }
 
-            if (requirement.RequireAdmin && !context.User.IsInRole(UserRoles.Administrator))
+            var contextUser = context.User;
+            if (requirement.RequireAdmin && !contextUser.IsInRole(UserRoles.Administrator))
+            {
+                context.Fail();
+                return Task.CompletedTask;
+            }
+
+            var userId = contextUser.GetUserId();
+            if (userId.Equals(default))
             {
                 context.Fail();
                 return Task.CompletedTask;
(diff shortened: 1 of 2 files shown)


nielsvanvelzen (approved):

Jellyfin.Api/Auth/FirstTimeSetupPolicy/FirstTimeSetupHandler.cs
```diff
@@ -38,7 +38,15 @@ protected override Task HandleRequirementAsync(AuthorizationHandlerContext conte
...
-            if (requirement.RequireAdmin && !context.User.IsInRole(UserRoles.Administrator))
+            var contextUser = context.User;
+            if (requirement.RequireAdmin && !contextUser.IsInRole(UserRoles.Administrator))
+            {
+                context.Fail();
+                return Task.CompletedTask;
+            }
+
+            var userId = contextUser.GetUserId();
+            if (userId.Equals(default))
```
crobibero, line 49: This makes it so api keys can't use the endpoint- is that what is desired?

Jellyfin.Api/Controllers/SystemController.cs
```diff
@@ -59,10 +59,12 @@ public class SystemController : BaseJellyfinApiController
     /// Gets information about the server.
     /// </summary>
     /// <response code="200">Information retrieved.</response>
+    /// <response code="403">User does not have permission to retrieve information.</response>
     /// <returns>A <see cref="SystemInfo"/> with info about the system.</returns>
     [HttpGet("Info")]
     [Authorize(Policy = Policies.FirstTimeSetupOrIgnoreParentalControl)]
     [ProducesResponseType(StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
```
crobibero, line 67: xmldoc says 403, response code says 401

Should be 403 everywhere

Jellyfin.Api/Auth/FirstTimeSetupPolicy/FirstTimeSetupHandler.cs
Shadowghost: API keys are handled in the `DefaultAuthorizationHandler`. The check over there only gets skipped if the request does not incude an API key and the default userID is used.

crobibero (approved):

Issue #9521, opened 2023-03-21
DLNA Play On remote is broken for Bose Soundtouch 309


espector:
I'm able to play to my Bose Soundtouch 300 sound bar. However, the remote screen is blank and doesn't show the play status. There's no way to control once the music stats casting. I can't pause, go to the next track, etc. 

I can control the volume on Soundtouch though. 

Setup: linuxserver jellyfin docker image, running on Synology. 

Screenshots below. 
![7939729A-577E-49AA-B028-34FEFC70439C](https://user-images.githubusercontent.com/3356907/226623820-0a8f2104-45a0-4d07-9c6c-95d7c012549c.jpeg)
![9830D10A-C002-4169-B7C7-5D930B584F15](https://user-images.githubusercontent.com/3356907/226623823-f45795f7-b3de-43e8-8d6b-475094389d26.png)

seaeagle1:
I'm having a similar issue with my Yamaha amplifier: PlayTo, song starts fine, but can't access the remote control function. Jellyfin gives the following error in the log:
```
[2023-05-12 12:19:16.075 +00:00] [ERR] [51] Emby.Dlna.Main.DlnaEntryPoint: Error updating device info for "R-N303D 42CCC0"
System.Xml.XmlException: '=' is an unexpected token. The expected token is ';'. Line 1, position 2605.
   at System.Xml.XmlTextReaderImpl.Throw(Exception e)
   at System.Xml.XmlTextReaderImpl.Throw(String res, String[] args)
   at System.Xml.XmlTextReaderImpl.ThrowUnexpectedToken(String expectedToken1, String expectedToken2)
   at System.Xml.XmlTextReaderImpl.HandleEntityReferenceAsync(Boolean isInAttributeValue, EntityExpandType expandType)
   at System.Xml.XmlTextReaderImpl.ParseTextAsync_ParseEntity(Int32 outOrChars, Char[] chars, Int32 pos, Int32 rcount, Int32 rpos, Int32 orChars, Char c)
   at System.Xml.XmlTextReaderImpl.ParseTextAsync_AsyncFunc(Task'1 task)
   at System.Xml.XmlTextReaderImpl.FinishPartialValueAsync()
   at System.Xml.XmlTextReaderImpl._GetValueAsync()
   at System.Xml.Linq.XContainer.ContentReader.ReadContentFromAsync(XContainer rootContainer, XmlReader r)
[...]

PR #9759, opened 2023-05-12
Rescue malformed Xml response in DLNA PlayTo

seaeagle1:
Not all devices (TVs/Receivers/Tuners) appear to provide well-formed Xml responses when using the 'remote control' / 'PlayTo' functionality. Currently, in such cases Jellyfin becomes unable to provide remote control or song progress, behaving like nothing happened (while the remote device starts playing), and throws an exception in the log. 
In this PR, when the XmlException is caught, it tries to escape unescaped & characters and then retries to read the Xml. If the Xml is now correct, Jellyfin will operate normally - if it is still not correct it will log the XmlException after all.

**Changes**
Add regex expression to the exception handler in DlnaHttpClient's SendRequestAsync function. Adding it there avoids interrupting the normal flow, but provides a rescue scenario for communicating with ill-behaving equipment.
Regex currently fixes unescaped ampersands, other corrections could easily be added.

**Issues**
Described in #9521. I'm not sure it fixes the original Bose issue, but does correct the behaviour with my amplifier.

### DIFF
--- a/Emby.Dlna/PlayTo/DlnaHttpClient.cs
+++ b/Emby.Dlna/PlayTo/DlnaHttpClient.cs
@@ -17,7 +17,10 @@ using Microsoft.Extensions.Logging;
 
 namespace Emby.Dlna.PlayTo
 {
-    public class DlnaHttpClient
+    /// <summary>
+    /// Http client for Dlna PlayTo function.
+    /// </summary>
+    public partial class DlnaHttpClient
     {
         private readonly ILogger _logger;
         private readonly IHttpClientFactory _httpClientFactory;
@@ -62,8 +65,7 @@ namespace Emby.Dlna.PlayTo
                 var xmlString = await response.Content.ReadAsStringAsync(cancellationToken).ConfigureAwait(false);
 
                 // find and replace unescaped ampersands (&)
-                Regex regex = new Regex(@"(&(?![a-z]*;))");
-                xmlString = regex.Replace(xmlString, @"&amp;");
+                xmlString = EscapeAmpersandRegex().Replace(xmlString, "&amp;");
 
                 try
                 {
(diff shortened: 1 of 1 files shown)


Shadowghost (approved):

Emby.Dlna/PlayTo/DlnaHttpClient.cs
```diff
@@ -54,15 +56,34 @@ private static string NormalizeServiceUrl(string baseUrl, string serviceUrl)
...
+                    var xmlReader = new StringReader(xmlString);
+                    return await XDocument.LoadAsync(
+                        xmlReader,
+                        LoadOptions.None,
+                        cancellationToken).ConfigureAwait(false);
                 }
+                catch (XmlException ex)
+                {
+                    _logger.LogError(ex, "Failed to parse response");
+                    if (_logger.IsEnabled(LogLevel.Debug))
```
cvium, line 80: Serilog already does this

Emby.Dlna/PlayTo/DlnaHttpClient.cs
```diff
@@ -54,15 +56,34 @@ private static string NormalizeServiceUrl(string baseUrl, string serviceUrl)
...
-            catch (XmlException ex)
+            catch (XmlException)
             {
-                _logger.LogError(ex, "Failed to parse response");
-                if (_logger.IsEnabled(LogLevel.Debug))
+                // try correcting the Xml response with common errors
+                var xmlString = await response.Content.ReadAsStringAsync(cancellationToken).ConfigureAwait(false);
+
+                // find and replace unescaped ampersands (&)
+                Regex regex = new Regex(@"(&(?![a-z]*;))");
```
cvium, line 65: Very inefficient. Either put it in a static variable or use the static methods i.e. `Regex.Replace(pattern, string)`

Emby.Dlna/PlayTo/DlnaHttpClient.cs
seaeagle1: Just a copy of what was there already, but I'm all for readability!

Emby.Dlna/PlayTo/DlnaHttpClient.cs
seaeagle1: Agreed. Replaced with the new GeneratedRegex.

Emby.Dlna/PlayTo/DlnaHttpClient.cs
Bond-009: The original intention of this was to only execute the expensive ReadAsString method when it would actually be logged

cvium (approved):

PR #9762, opened 2023-05-13
Convert string MediaType to enum MediaType

crobibero:
Requested by client devs

Works as normal from my testing


Jellyfin.Data/Enums/MediaType.cs
```diff
@@ -0,0 +1,32 @@
...
+
+/// <summary>
+/// Media types.
+/// </summary>
+public enum MediaType
+{
+    /// <summary>
+    /// Unknown media type.
+    /// </summary>
+    Unknown = 0,
```
nielsvanvelzen, line 11: Is it possible to get unknown media types via the API or do unknown media items only exist internally?

crobibero: I could probably go through and remove Unknown and have the default be null instead

EraYaN: It's probably sane to have a good default value where everyone knows to just error out because something is wrong, right?

Emby.Server.Implementations/Playlists/PlaylistManager.cs
```diff
@@ -84,7 +85,7 @@ public async Task<PlaylistCreationResult> CreatePlaylist(PlaylistCreationRequest
                         throw new ArgumentException("No item exists with the supplied Id");
                     }
 
-                    if (!string.IsNullOrEmpty(item.MediaType))
+                    if (item.MediaType == MediaType.Unknown)
```
1337joe, line 88: Missed the not
```suggestion
                    if (item.MediaType != MediaType.Unknown)
```

Emby.Server.Implementations/Playlists/PlaylistManager.cs
```diff
@@ -102,20 +103,20 @@ public async Task<PlaylistCreationResult> CreatePlaylist(PlaylistCreationRequest
                         {
                             options.MediaType = folder.GetRecursiveChildren(i => !i.IsFolder && i.SupportsAddingToPlaylist)
                                 .Select(i => i.MediaType)
-                                .FirstOrDefault(i => !string.IsNullOrEmpty(i));
+                                .FirstOrDefault(i => i != MediaType.Unknown);
                         }
                     }
 
-                    if (!string.IsNullOrEmpty(options.MediaType))
+                    if (options.MediaType is null || options.MediaType == MediaType.Unknown)
```
1337joe, line 110: Missed the not
```suggestion
                    if (options.MediaType is not null && options.MediaType != MediaType.Unknown)
```

1337joe (changes requested):

1337joe (approved):

Shadowghost (approved):

Bond-009 (approved):

PR #9764, opened 2023-05-13
Convert CollectionType, SpecialFolderType to enum

crobibero:
Requested by client devs

Draft pending testing


MediaBrowser.Controller/Entities/UserViewBuilder.cs
```diff
@@ -952,7 +952,7 @@ private BaseItem[] GetMediaFolders(User user, IEnumerable<string> viewTypes)
                     {
                         var folder = i as ICollectionFolder;
 
-                        return folder is not null && viewTypes.Contains(folder.CollectionType ?? string.Empty, StringComparison.OrdinalIgnoreCase);
+                        return folder?.CollectionType != null && viewTypes.Contains(folder.CollectionType.Value);
```
Shadowghost, line 955:
```suggestion
                        return folder?.CollectionType is not null && viewTypes.Contains(folder.CollectionType.Value);
```

Emby.Server.Implementations/Library/Resolvers/SpecialFolderResolver.cs
```diff
@@ -78,7 +79,8 @@ private string GetCollectionType(ItemResolveArgs args)
                     }
                 })
                 .Select(i => _fileSystem.GetFileNameWithoutExtension(i))
-                .FirstOrDefault();
+                .Select(i => Enum.TryParse<CollectionType>(i, out var collectionType) ? collectionType : (CollectionType?)null)
+                .FirstOrDefault(i => i != null);
```
Shadowghost, line 83:
```suggestion
                .FirstOrDefault(i => i is not null);
```

MediaBrowser.Controller/Library/ItemResolveArgs.cs
```diff
@@ -220,7 +221,7 @@ public bool ContainsFileSystemEntryByName(string name)
             return GetFileSystemEntryByName(name) != null;
```
Shadowghost, line 221:
```suggestion
            return GetFileSystemEntryByName(name) is not null;
```

Jellyfin.Data/Enums/CollectionType.cs
```diff
@@ -0,0 +1,147 @@
...
+
+    /// <summary>
+    /// Tv genres collection.
+    /// </summary>
+    TvGenres = 102,
+
+    /// <summary>
+    /// Tv genre collection.
+    /// </summary>
+    TvGenre = 103,
```
nielsvanvelzen, line 86: What's the difference between these two?

Jellyfin.Data/Enums/CollectionType.cs
```diff
@@ -0,0 +1,147 @@
...
+
+    /// <summary>
+    /// Movie genres collection.
+    /// </summary>
+    MovieGenres = 114,
+
+    /// <summary>
+    /// Movie genre collection.
+    /// </summary>
+    MovieGenre = 115
```
nielsvanvelzen, line 146: What's the difference between these two?

Jellyfin.Data/Enums/CollectionType.cs
crobibero: Yay legacy code where they added support for both…

Shadowghost: Can't we write a migration and get rid of them?

EraYaN: We could mark one as `[Obsolete]` or something right?

crobibero: The difference is actually as follows so it needs to stay... I've opted to just hide them from the api spec

https://github.com/jellyfin/jellyfin/blob/master/MediaBrowser.Controller/Entities/UserViewBuilder.cs#L73:L77

EraYaN: Aah so one is really a TvGenreList (list of TVGenres) and the other one a list of items in a specific genre.

crobibero:
@jellyfin-bot rebase

Bond-009 (approved):

Shadowghost (approved):

Jellyfin.Data/Attributes/OpenApiIgnoreEnumAttribute.cs
```diff
@@ -0,0 +1,11 @@
+using System;
+
+namespace Jellyfin.Data.Attributes;
+
+/// <summary>
+/// Attribute to specify that the enum value is to be ignored when generating the openapi spec..
```
barronpm, line 6:
```suggestion
/// Attribute to specify that the enum value is to be ignored when generating the openapi spec.
```

barronpm (approved):

PR #9776, opened 2023-05-15
Use default files to remove index.html from url

thornbill:
**Changes**
Use default files to remove "index.html" from the default web url

**Issues**
N/A

### DIFF
--- a/Jellyfin.Api/Middleware/BaseUrlRedirectionMiddleware.cs
+++ b/Jellyfin.Api/Middleware/BaseUrlRedirectionMiddleware.cs
@@ -48,8 +48,6 @@ public class BaseUrlRedirectionMiddleware
         if (string.IsNullOrEmpty(localPath)
             || string.Equals(localPath, baseUrlPrefix, StringComparison.OrdinalIgnoreCase)
             || string.Equals(localPath, baseUrlPrefix + "/", StringComparison.OrdinalIgnoreCase)
-            || string.Equals(localPath, baseUrlPrefix + "/web", StringComparison.OrdinalIgnoreCase)
-            || string.Equals(localPath, baseUrlPrefix + "/web/", StringComparison.OrdinalIgnoreCase)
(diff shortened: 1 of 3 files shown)


Jellyfin.Api/Middleware/BaseUrlRedirectionMiddleware.cs
```diff
@@ -48,8 +48,6 @@ public async Task Invoke(HttpContext httpContext, IServerConfigurationManager se
         if (string.IsNullOrEmpty(localPath)
             || string.Equals(localPath, baseUrlPrefix, StringComparison.OrdinalIgnoreCase)
             || string.Equals(localPath, baseUrlPrefix + "/", StringComparison.OrdinalIgnoreCase)
-            || string.Equals(localPath, baseUrlPrefix + "/web", StringComparison.OrdinalIgnoreCase)
-            || string.Equals(localPath, baseUrlPrefix + "/web/", StringComparison.OrdinalIgnoreCase)
```
thornbill, line 52: These would cause a redirect loop since the default url is now just `/web/`

Jellyfin.Server/Startup.cs
```diff
@@ -200,6 +205,7 @@ public void ConfigureServices(IServiceCollection services)
                     mainApp.UseRobotsRedirection();
                 }
 
+                mainApp.UseStaticFiles();
```
thornbill, line 208: I moved this because according to [the docs](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/static-files?view=aspnetcore-7.0#serve-default-documents): 

> `UseDefaultFiles` must be called before `UseStaticFiles` to serve the default file. `UseDefaultFiles` is a URL rewriter that doesn't serve the file.

Shadowghost (approved):

crobibero (approved):

Bond-009 (approved):

PR #9799, opened 2023-05-22
Use RegexGenerator where possible

Bond-009:
As far as I know this shouldn't have any  downsides?

https://learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-source-generators

### DIFF
--- a/Emby.Naming/Audio/AlbumParser.cs
+++ b/Emby.Naming/Audio/AlbumParser.cs
@@ -10,7 +10,7 @@ namespace Emby.Naming.Audio
     /// <summary>
     /// Helper class to determine if Album is multipart.
     /// </summary>
-    public class AlbumParser
+    public partial class AlbumParser
     {
         private readonly NamingOptions _options;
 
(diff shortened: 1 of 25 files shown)


Emby.Naming/Audio/AlbumParser.cs
```diff
@@ -23,6 +23,9 @@ public AlbumParser(NamingOptions options)
             _options = options;
         }
 
+        [GeneratedRegex(@"([-\.\(\)]|\s+)")]
+        private static partial Regex CleanRegex();
```
crobibero, line 27: This new regex doesn't have the same output

Old: 
`disc- 1` -> `disc 1`
New (note double spaces): 
`disc- 1` -> `disc  1`

Bond-009: Good catch

crobibero (approved):

Shadowghost (approved):

PR #9802, opened 2023-05-23
Fix: Always consider HLS supported for direct play

justjanne:
**Changes**
Always consider HLS supported, regardless of what the client actually reports. We're using HLS for transcoding anyway, so clients have to support it no matter what.

**Issues**
Without these changes, adding HLS streams referenced via .strm files will always fall back to remuxing.


Shadowghost (dismissed):

justjanne:
@nyanmisaka We're already checking the codecs of the included audio and video streams. This would just circumvent the checks for the container itself.

If you look a few lines further down, the

```cs
if (!directPlayProfile.SupportsVideoCodec(videoCodec))
```

checks aren't affected.

cvium:
> We're using HLS for transcoding anyway, so clients have to support it no matter what.

We also have progressive streams though. I don't think this is the right solution. It feels more like a hack than anything.

justjanne:
@cvium It's definitely an ugly hack. Ideally we'd modify the clients to correctly advertise HLS support (and make sure they also support switching audio tracks and video quality levels in a real HLS stream)

PR #9821, opened 2023-05-26
refactor: admin users must have a non-empty password

cvium:
**Changes**
It's annoying if your setup is strictly on LAN, but for the greater good, admin users must have a non-empty password


theGUI001 (approved):
The code is correct and the function follows security best practices, so the code is approved

Shadowghost (approved):

Bond-009 (approved):
I'd prefer passwords being mandatory for all users, but this is an improvement nonetheless

crobibero (approved):

PR #9863, opened 2023-06-06
Reduce bottlenecks scan code

Bond-009:
* Removes MetadataThrottler, we already limit the amount of directories to start scanning (this PR multiplies this limit by 2) in and the number of ffprobe instances, this extra throttling doesn't seem necessary.
* Increases the amount of thumbnail extraction processes from 2 to 2 * the number of threads

### DIFF
--- a/MediaBrowser.Providers/Manager/ProviderManager.cs
+++ b/MediaBrowser.Providers/Manager/ProviderManager.cs
@@ -131,12 +131,12 @@ namespace MediaBrowser.Providers.Manager
         {
             var type = item.GetType();
 
-            var service = _metadataServices.FirstOrDefault(current => current.CanRefreshPrimary(type));
-            service ??= _metadataServices.FirstOrDefault(current => current.CanRefresh(item));
+            var service = _metadataServices.FirstOrDefault(current => current.CanRefreshPrimary(type))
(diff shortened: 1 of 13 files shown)


crobibero (approved):

Shadowghost (approved):

MediaBrowser.Providers/Manager/ProviderManager.cs
```diff
@@ -809,27 +808,12 @@ private bool IsSaverEnabledForItem(IMetadataSaver saver, BaseItem item, LibraryO
         {
             var results = await provider.GetSearchResults(searchInfo, cancellationToken).ConfigureAwait(false);
 
-            var list = results.ToList();
```
goremykin, line 812: Multiple enumeration was introduced here and as a result potential bugs.
As I see, most of `GetSearchResults` implementations return arrays and there is no problem in this case.
But at least `MusicBrainzArtistProvider` returns a lazy enumeration using `yield` statement. And here we have a bug:
`MusicBrainzArtistProvider` lazily creates `RemoteSearchResult` while we iterate here to set `item.SearchProviderName` and then it creates new ones when we iterate this enumerable again outside. So we actually lose `SearchProviderName`

But I haven't tested yet

PR #9874, opened 2023-06-09
DLNA: Profiles: PS3: don't use AAC for transcoding

bibarub:
PS3 doesn't seem to like it, there's no sound

**Changes**
Removed the AAC codec for transcoding in the PS3 DLNA profile since it doesn't work

**Issues**
No Issues were opened, but it fixes sound on the PS3


Shadowghost (approved):

cvium:
Are you sure it's compatibility with the AAC codec and not something else? Plex and Sony both mention AAC as a supported format

https://support.plex.tv/articles/204377253-what-media-formats-are-supported/
https://manuals.playstation.net/document/en/ps3/current/music/filetypes.html

Shadowghost:
Might also be related to specific bitrates or channel amounts. I can't test it since I don't have a PS3...
@bibarub are you sure this is happening for AAC audio?

bibarub:
> Might also be related to specific bitrates or channel amounts. I can't test it since I don't have a PS3... @bibarub are you sure this is happening for AAC audio?

quite sure. one thing i'm unsure about is whether the ps3 doesn't support aac at all or it's just jellyfin somehow producing incompatible aac streams

barronpm:
DLNA functionality has been moved to a [plugin](https://github.com/jellyfin/jellyfin-plugin-dlna). If you're still interested in this change, please re-open this pull request in the plugin repository.

PR #9875, opened 2023-06-09
Various improvements

### LINKED ISSUE #9975: [Issue]: missing interpolation in the `MaskToCidr` method
### Please describe your bug

I've checked project sources with the PVS-Studio analyzer and found missing interpolation character:
```
....
if (!mask.TryWriteBytes(bytes, out var bytesWritten))
{
  Console.WriteLine("Unable to write address bytes, only {bytesWritten} bytes written.");
}
....
```
[Link to the sources](https://github.com/jellyfin/jellyfin/blob/5b711f38d143fc3eca231c745a52fedbe7a10092/Jellyfin.Networking/Extensions/NetworkExtensions.cs#L107).

The `bytesWritten` variable's value won't be written to the string as the `$` character is missed.

### Jellyfin Version

Other

### if other:

the `master` branch

### Environment

_No response_

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

Shadowghost:
**Changes**
* Add wrapper object for authentication events, removing stuff like API tokens from event messages (prevents leaking them to e.g. plugins)
* Add corresponding item id tp start/stop playback events
* Enforce access permissions on WebSocketListeners

Note: The now secured listeners aren't working in any meaningful way. The WebSockets handling really requrires an overhaul....

Fixes #9975

### DIFF
--- a/Emby.Server.Implementations/Session/SessionManager.cs
+++ b/Emby.Server.Implementations/Session/SessionManager.cs
@@ -24,6 +24,7 @@ using MediaBrowser.Controller.Drawing;
 using MediaBrowser.Controller.Dto;
 using MediaBrowser.Controller.Entities;
 using MediaBrowser.Controller.Events;
+using MediaBrowser.Controller.Events.Authentication;
 using MediaBrowser.Controller.Events.Session;
 using MediaBrowser.Controller.Library;
 using MediaBrowser.Controller.Net;
@@ -1462,7 +1463,7 @@ namespace Emby.Server.Implementations.Session
 
             if (user is null)
             {
-                await _eventManager.PublishAsync(new GenericEventArgs<AuthenticationRequest>(request)).ConfigureAwait(false);
+                await _eventManager.PublishAsync(new GenericEventArgs<AuthenticationRequestEventArgs>(new AuthenticationRequestEventArgs(request))).ConfigureAwait(false);
                 throw new AuthenticationException("Invalid username or password entered.");
             }
 
@@ -1498,7 +1499,7 @@ namespace Emby.Server.Implementations.Session
                 ServerId = _appHost.SystemId
             };
 
-            await _eventManager.PublishAsync(new GenericEventArgs<AuthenticationResult>(returnResult)).ConfigureAwait(false);
+            await _eventManager.PublishAsync(new GenericEventArgs<AuthenticationResultEventArgs>(new AuthenticationResultEventArgs(returnResult))).ConfigureAwait(false);
             return returnResult;
         }
 
--- a/Emby.Server.Implementations/Session/SessionManager.cs
+++ b/Emby.Server.Implementations/Session/SessionManager.cs
@@ -1463,7 +1463,7 @@ namespace Emby.Server.Implementations.Session
 
             if (user is null)
             {
-                await _eventManager.PublishAsync(new GenericEventArgs<AuthenticationRequestEventArgs>(new AuthenticationRequestEventArgs(request))).ConfigureAwait(false);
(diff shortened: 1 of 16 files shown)


Jellyfin.Server.Implementations/Events/Consumers/Session/PlaybackStopLogger.cs
```diff
@@ -73,7 +73,10 @@ public async Task OnEvent(PlaybackStopEventArgs eventArgs)
                         GetItemName(item),
                         eventArgs.DeviceName),
                     notificationType,
-                    user.Id))
+                    user.Id)
+                {
+                    ItemId = eventArgs.Item?.Id.ToString()
```
crobibero, line 78:
```suggestion
                    ItemId = eventArgs.Item?.Id.ToString("N", CultureInfo.InvariantCulture),
```

Jellyfin.Server.Implementations/Events/Consumers/Session/PlaybackStartLogger.cs
```diff
@@ -58,15 +58,18 @@ public async Task OnEvent(PlaybackStartEventArgs eventArgs)
...
+                string.Format(
+                    CultureInfo.InvariantCulture,
+                    _localizationManager.GetLocalizedString("UserStartedPlayingItemWithValues"),
+                    user.Username,
+                    GetItemName(eventArgs.MediaInfo),
+                    eventArgs.DeviceName),
+                GetPlaybackNotificationType(eventArgs.MediaInfo.MediaType),
+                user.Id)
+            {
+                ItemId = eventArgs.Item?.Id.ToString()
```
crobibero, line 70:
```suggestion
                ItemId = eventArgs.Item?.Id.ToString("N", CultureInfo.InvariantCulture),
```

Jellyfin.Api/WebSocketListeners/ActivityLogWebSocketListener.cs
```diff
@@ -56,6 +58,16 @@ protected override void Dispose(bool dispose)
         base.Dispose(dispose);
     }
 
+    private new void Start(WebSocketMessageInfo message)
```
cvium, line 61: Why new? Shouldn't it just be protected virtual?

cvium (approved):

Jellyfin.Api/WebSocketListeners/ActivityLogWebSocketListener.cs
Shadowghost: `protected virtual` throws `CS0108`, only way it did not throw this error was using `new`

cvium: Parent uses protected virtual. Child classes use protected override.

Emby.Server.Implementations/Session/SessionManager.cs
```diff
@@ -1462,7 +1463,7 @@ private async Task<AuthenticationResult> AuthenticateNewSessionInternal(Authenti
 
             if (user is null)
             {
-                await _eventManager.PublishAsync(new GenericEventArgs<AuthenticationRequest>(request)).ConfigureAwait(false);
+                await _eventManager.PublishAsync(new GenericEventArgs<AuthenticationRequestEventArgs>(new AuthenticationRequestEventArgs(request))).ConfigureAwait(false);
```
Bond-009, line 1466: Wouldn't it be better to make `AuthenticationRequestEventArgs` extend `EventArgs` instead of wrapping it in `GenericEventArgs`?

Bond-009 (approved):

PR #9882, opened 2023-06-12
Rearrage the Amd vaapi-vulkan pipeline for synchronization

nyanmisaka:
**Changes**
- Rearrage the Amd vaapi-vulkan pipeline for synchronization

**Issues**

The [vaSyncSurface()](https://gitlab.freedesktop.org/mesa/mesa/-/blob/23.1/src/gallium/frontends/va/surface.c?ref_type=heads#L123-L127) in radeonsi_drv_video only works for decoder or encoder entries, which means vpp frames never get synchronized in [vaapi->drm/vulkan "reverse" mapping](https://github.com/FFmpeg/FFmpeg/blob/ea3d24bbe3c58b171e55fe2151fc7ffaca3ab3d2/libavutil/hwcontext_vulkan.c#L2963-L2964). Use [vulkan_map_to_drm](https://github.com/FFmpeg/FFmpeg/blob/ea3d24bbe3c58b171e55fe2151fc7ffaca3ab3d2/libavutil/hwcontext_vulkan.c#L3306) instead.

- Fix tearing and artifacts when using vulkan subtitle burn-in

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -163,7 +163,8 @@ namespace MediaBrowser.Controller.MediaEncoding
 
         private bool IsVaapiFullSupported()
         {
-            return _mediaEncoder.SupportsHwaccel("vaapi")
+            return _mediaEncoder.SupportsHwaccel("drm")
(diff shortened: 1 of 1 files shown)


Bond-009 (approved):
Fixes the issue for me

PR #9894, opened 2023-06-16
Move hardcoded LibraryUpdateDuration to ServerConfiguration.

### LINKED ISSUE #9893: [Issue]: LibraryUpdateDuration is hardcoded
### Please describe your bug

`LibraryUpdateDuration` is [hardcoded to 30 seconds](https://github.com/jellyfin/jellyfin/blob/6771b5cabe96b4b3cbd1cd0c998d564f3dd17ed4/Emby.Server.Implementations/EntryPoints/LibraryChangedNotifier.cs#L32). It should go in ServerConfiguration so it can be changed (similar to [LibraryMonitorDelay](https://github.com/jellyfin/jellyfin/blob/6771b5cabe96b4b3cbd1cd0c998d564f3dd17ed4/MediaBrowser.Model/Configuration/ServerConfiguration.cs#L166)).

### Jellyfin Version

10.8.z

### if other:

_No response_

### Environment

_No response_

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

fhriley:
**Changes**

Move hardcoded LibraryUpdateDuration to ServerConfiguration.

**Issues**

Fixes #9893

### DIFF
--- a/MediaBrowser.Model/Configuration/ServerConfiguration.cs
+++ b/MediaBrowser.Model/Configuration/ServerConfiguration.cs
@@ -166,6 +166,12 @@ namespace MediaBrowser.Model.Configuration
         public int LibraryMonitorDelay { get; set; } = 60;
 
         /// <summary>
+        /// Gets or sets the duration in seconds that we will wait after a library updated event before executing the library changed notification.
+        /// </summary>
+        /// <value>The library update duration.</value>
+        public int LibraryUpdateDuration { get; set; } = 30;
+
+        /// <summary>
         /// Gets or sets the image saving convention.
         /// </summary>
         /// <value>The image saving convention.</value>
--- a/Emby.Server.Implementations/EntryPoints/LibraryChangedNotifier.cs
+++ b/Emby.Server.Implementations/EntryPoints/LibraryChangedNotifier.cs
@@ -12,6 +12,7 @@ using System.Threading.Tasks;
 using Jellyfin.Data.Entities;
 using Jellyfin.Data.Events;
 using MediaBrowser.Controller.Channels;
+using MediaBrowser.Controller.Configuration;
 using MediaBrowser.Controller.Entities;
 using MediaBrowser.Controller.Entities.Audio;
 using MediaBrowser.Controller.Library;
@@ -26,12 +27,8 @@ namespace Emby.Server.Implementations.EntryPoints
 {
     public class LibraryChangedNotifier : IServerEntryPoint
     {
-        /// <summary>
-        /// The library update duration.
-        /// </summary>
-        private const int LibraryUpdateDuration = 30000;
-
         private readonly ILibraryManager _libraryManager;
+        private readonly IServerConfigurationManager _configurationManager;
         private readonly IProviderManager _providerManager;
         private readonly ISessionManager _sessionManager;
         private readonly IUserManager _userManager;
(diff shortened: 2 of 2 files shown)


Shadowghost (approved):

MediaBrowser.Model/Configuration/ServerConfiguration.cs
```diff
@@ -165,6 +165,12 @@ public ServerConfiguration()
         /// <value>The file watcher delay.</value>
         public int LibraryMonitorDelay { get; set; } = 60;
 
+        /// <summary>
+        /// Gets or sets the duration in seconds that we will wait after a library updated event before executing the library changed notification.
+        /// </summary>
+        /// <value>The library update duration.</value>
+        public int LibraryUpdateDuration { get; set; } = 30;
```
crobibero, line 172: This is in seconds but the `dueTime` parameter expects milliseconds, making the timer now fire every 30ms.

crobibero (changes requested):

crobibero (commented):
I would change `Timeout.InfiniteTimeSpan` back to `Timeout.Infinite` as internally the constructor converts to a `uint`

fhriley:
> I would change `Timeout.InfiniteTimeSpan` back to `Timeout.Infinite` as internally the constructor converts to a `uint`

I can change this, but are you sure? The constructor takes a `TimeSpan` so to use `Timeout.Infinite` we'd have to pass something like `TimeSpan.FromMilliseconds(Timeout.Infinite)`. This codes an assumption that the infinite timespan is in milliseconds. Why not use the constant designed for this and not make the assumption? Or are you saying use the constructor that uses `int`s rather than the `TimeSpan` version?

crobibero:
>  Or are you saying use the constructor that uses ints rather than the TimeSpan version?

This, what the code was originally doing.

The docs show both work:
> `period`  [TimeSpan](https://learn.microsoft.com/en-us/dotnet/api/system.timespan?view=net-7.0) 
> The time interval between invocations of callback. Specify [InfiniteTimeSpan](https://learn.microsoft.com/en-us/dotnet/api/system.threading.timeout.infinitetimespan?view=net-7.0#system-threading-timeout-infinitetimespan) to disable periodic signaling.

> `period`  [Int32](https://learn.microsoft.com/en-us/dotnet/api/system.int32?view=net-7.0) 
> The time interval between invocations of callback, in milliseconds. Specify [Infinite](https://learn.microsoft.com/en-us/dotnet/api/system.threading.timeout.infinite?view=net-7.0#system-threading-timeout-infinite) to disable periodic signaling.

This is the constructor you moved to, which just takes the `TotalMilliseconds`, so there is no benefit besides requiring extra property access.
https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Threading/Timer.cs#L848-L862

fhriley:
I [matched the implementation](https://github.com/jellyfin/jellyfin/blob/c078391aece1b341343d743d7ecfb002ed177f6c/Emby.Server.Implementations/IO/FileRefresher.cs#L79) of `LibraryMonitorDelay`, but fixed the assumption.

fhriley:
I'll make one more pitch, and then if you still want it changed, I'll change it. The difference is this:

`_configurationManager.Configuration.LibraryUpdateDuration * 1000`

vs

`TimeSpan.FromSeconds(_configurationManager.Configuration.LibraryUpdateDuration)`

The former is not clear what time unit `LibraryUpdateDuration` is in. You only know it's in seconds if you look up the `Timer` constructor and see that it takes milliseconds. OTOH, the latter is very clear just by looking at the code. `LibraryUpdateDuration` is in seconds. Do you still want me to implement the int version?

crobibero:
🤦 sorry I totally missed that there wasn't a combined constructor. What you have is completely fine and I was just confused at the additional change

crobibero (approved):

Bond-009 (approved):

PR #9220, opened 2023-01-31
Add support for embedded lyrics to lyrics API

DomiStyle:
**Changes**
This PR adds support for loading embedded lyrics from music files in LRC and TXT format. Previously these were only loaded from external files.

In order to store the lyrics I enabled the JSON serialization for items of type Audio. If there's a problem with this it would also be possible to add a database field for it or retrieve them when requested.

PR #9920, opened 2023-06-23
Add support for lyric provider plugins

nielsvanvelzen:
**Changes**
This change splits the current LyricProvider into two interfaces; the provider and the parser. This allows for plugins to provide lyrics from different places (like online sources, although since nothing is cached in the DB this would be extremely inefficient at the moment).

An example for a plugin that provides lyrics in [this gist](https://gist.github.com/nielsvanvelzen/3c6fa26eab4f231096b0426a27f34f76).

Some eventual changes that could be made (but I don't have the time for):
- Cache lyrics in db
- Add options to library to enable/disable parsers/providers

**Issues**


MediaBrowser.Providers/Lyric/TxtLyricParser.cs
```diff
@@ -0,0 +1,49 @@
...
+
+    /// <inheritdoc />
+    public LyricResponse? ParseLyrics(LyricFile lyrics)
+    {
+        if (!_supportedMediaTypes.Contains(Path.GetExtension(lyrics.Name.AsSpan())[1..], StringComparison.OrdinalIgnoreCase))
+        {
+            return null;
+        }
+
+        string[] lyricTextLines = lyrics.Content.Split(new[] { "\r\n", "\r", "\n" }, StringSplitOptions.None);
```
crobibero, line 33: should move the split characters to `private static readonly`

crobibero (approved):

DomiStyle:
I like this, should also be useful for the embedded metadata provider PR I have here: https://github.com/jellyfin/jellyfin/pull/9220

Bond-009 (approved):

MediaBrowser.Providers/Lyric/LrcLyricParser.cs
```diff
@@ -3,34 +3,29 @@
...
+/// LRC Lyric Parser.
 /// </summary>
-public class LrcLyricProvider : ILyricProvider
+public class LrcLyricParser : ILyricParser
 {
-    private readonly ILogger<LrcLyricProvider> _logger;
-
     private readonly LyricParser _lrcLyricParser;
 
+    private static readonly string[] _supportedMediaTypes = { "lrc", "elrc" };
```
cvium, line 21: Wouldn't it be simpler if you included a dot here? That way you won't have to remove it from the GetExtension calls

cvium (changes requested):
Just the one comment (which applies to all occurrences)

cvium (approved):

MediaBrowser.Providers/Lyric/TxtLyricParser.cs
```diff
@@ -0,0 +1,50 @@
...
+    public LyricResponse? ParseLyrics(LyricFile lyrics)
+    {
+        if (!_supportedMediaTypes.Contains(Path.GetExtension(lyrics.Name.AsSpan()), StringComparison.OrdinalIgnoreCase))
+        {
+            return null;
+        }
+
+        string[] lyricTextLines = lyrics.Content.Split(_lineBreakCharacters, StringSplitOptions.None);
+
+        if (lyricTextLines.Length == 0)
```
cvium, line 36: This can't happen. It will always contain at least one element (unless the string is null, which results in NullReferenceException). You should check whether the string is null or empty prior to the split.

cvium (changes requested):
Sorry, I noticed something

MediaBrowser.Providers/Lyric/DefaultLyricProvider.cs
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    /// <inheritdoc />
+    public async Task<LyricFile?> GetLyrics(BaseItem item)
+    {
+        var path = GetLyricsPath(item);
+        if (path is not null)
+        {
+            var content = await File.ReadAllTextAsync(path).ConfigureAwait(false);
+            if (content.Length != 0)
```
cvium, line 35:
```suggestion
            if (!string.IsNullOrEmpty(content))
```

cvium (approved):

6xingyv:
Will it support karaoke lyrics?

NicholasFlamy:
Not sure if this is related, but the Image Extractor creates a .lrc file using the embedded lyrics of a file.
Screenshot 2024-05-04 150138

NicholasFlamy:
> Not sure if this is related, but the Image Extractor creates a .lrc file using the embedded lyrics of a file. Screenshot 2024-05-04 150138

Related to https://github.com/jellyfin/jellyfin/pull/11480

Issue #109, opened 2018-12-12
Support for edl/comskip files


kmarq:
Add support for edl/comskip files. Would be great to be able to provide a list of cut points and have playback follow this. 

See Kodi (https://kodi.wiki/view/Edit_decision_list) for an example.

PR #9924, opened 2023-06-24
Add muting/unmuting to plugin API

jacob-willden:
**Changes**

Adds the ability to programmatically mute, un-mute, or toggle muting via Jellyfin's plugin API (like with playing/pausing or seeking). Local tests via `dotnet test` all passed.

**Issues**

While not fully addressing #109, it does provide the opportunity for Edit Decision List (EDL) support through plugins, since EDLs are meant to allow both muting and skipping in video files.


MediaBrowser.Model/Session/PlaystateCommand.cs
```diff
@@ -48,6 +48,21 @@ public enum PlaystateCommand
...
+
+        /// <summary>
+        /// The unmute.
+        /// </summary>
+        Unmute,
+
+        /// <summary>
+        /// The toggle mute.
+        /// </summary>
+        ToggleMute
```
nielsvanvelzen, line 66: Changing this enum affects one of the API endpoints which I think is not intended. To mute/unmute/togglemute with the API you can already use the `/Sessions/[sessionid]/Command` endpoint.

jacob-willden: Oh interesting. So unlike with fast forwarding, rewinding, playing, pausing, and seeking, muting/un-muting or toggling muting does not need to be included in the `PlaystateCommand` enum. I can definitely revert that change.

jacob-willden: Hmm, removing those values from the enum made a test fail that didn't fail with a previous commit. Are we sure that we don't need them?

jacob-willden:
Once I added back the above values in PlaystateCommand.cs, the tests passed again, so I'm confident that the code in the other file depends on it.

jacob-willden:
Interestingly enough, even though all the tests passed, there are [a couple of errors in the CodeQL analysis that just ran today](https://github.com/jacob-willden/jellyfin/actions/workflows/codeql-analysis.yml). Would those be problematic?

Shadowghost:
CodeQL can be ignored most of the time, so it should be fine as is.

JPVenson:
@DerrikMilligan there are open merge issues and one open comment that needs to be resolved befor this can be merged

jacob-willden:
`Emby.Dlna/PlayTo/PlayToController.cs` has been deleted, without any apparent replacement. How are other `PlaystateCommand`s (like `Stop`, `Pause`, `Unpause`, or `NextTrack`) implemented now? Based on my search for `PlaystateCommand`, it appears that they aren't implemented anywhere.

Bond-009:
Closing as stale

PR #9938, opened 2023-06-28
Added Throttled Download as option

JPVenson:
- Added custom physical file provider that can throttle download speeds based on a new User setting.
- Allows in-progress downloads to change download speed
- Refactored UserManager to invoke UserChange event always when user has changed


Jellyfin.Server.Implementations/Users/UserManager.cs
```diff
@@ -608,9 +605,7 @@ await using (dbContext.ConfigureAwait(false))
                 user.SetPreference(PreferenceKind.MyMediaExcludes, config.MyMediaExcludes);
                 user.SetPreference(PreferenceKind.LatestItemExcludes, config.LatestItemsExcludes);
 
-                dbContext.Update(user);
-                _users[user.Id] = user;
-                await dbContext.SaveChangesAsync().ConfigureAwait(false);
+                await UpdateUserInternalAsync(dbContext, user).ConfigureAwait(false);
```
JPVenson, line 608: Everything done here is just to fix the inconsistent change event behavior

Jellyfin.Api/Helpers/ThrottledPhysicalFileResultExecutor.cs
```diff
@@ -0,0 +1,208 @@
...
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Microsoft.Net.Http.Headers;
+
+namespace Jellyfin.Api.Helpers;
+
+/// <summary>
+/// Provides methods to throttle the download of a physical file.
+/// </summary>
+public class ThrottledPhysicalFileResultExecutor : PhysicalFileResultExecutor, IActionResultExecutor<ThrottledPhysicalFileActionResult>
```
JPVenson, line 20: this whole class is in essence a copy-paste implementation of `PhysicalFileResultExecutor` plus the throttle behavior

Jellyfin.Api/Helpers/BandwidthLimitOption.cs
```diff
@@ -0,0 +1,19 @@
...
+/// </summary>
+public class BandwidthLimitOption
+{
+    /// <summary>
+    /// Gets or sets the Guid of the users id.
+    /// </summary>
+    public Guid User { get; set; }
+
+    /// <summary>
+    /// Gets or sets the bandwidth per second for a download.
```
crobibero, line 16: In what unit is this bandwith expected?

JPVenson: bytes. Same behavior as the Transcoding limit

JPVenson: Updated code comments to clarify unit

cvium:
Why is this a user setting? Why does a user want to throttle the download? I can see why a server admin would want to throttle.

JPVenson:
> Why is this a user setting? Why does a user want to throttle the download? I can see why a server admin would want to throttle.

Its a setting _to_ a user much like the throttle transcode/playback setting and should be handled more or less the same. Maybe i overlooked something to make it only settable for an admin but its build for an admin to set the same way as the Bandwidth setting.

cvium:
> > Why is this a user setting? Why does a user want to throttle the download? I can see why a server admin would want to throttle.
> 
> Its a setting _to_ a user much like the throttle transcode/playback setting and should be handled more or less the same. Maybe i overlooked something to make it only settable for an admin but its build for an admin to set the same way as the Bandwidth setting.

Why have a per-user limit? Should it not just be a global setting?

JPVenson:
> > > Why is this a user setting? Why does a user want to throttle the download? I can see why a server admin would want to throttle.
> > 
> > 
> > Its a setting _to_ a user much like the throttle transcode/playback setting and should be handled more or less the same. Maybe i overlooked something to make it only settable for an admin but its build for an admin to set the same way as the Bandwidth setting.
> 
> Why have a per-user limit? Should it not just be a global setting?

Ultimately, yes as it then would be the same as with the other bandwidth setting, however i did not had enough time to implement that aswell yet.

cvium:
> > > > Why is this a user setting? Why does a user want to throttle the download? I can see why a server admin would want to throttle.
> > > 
> > > 
> > > Its a setting _to_ a user much like the throttle transcode/playback setting and should be handled more or less the same. Maybe i overlooked something to make it only settable for an admin but its build for an admin to set the same way as the Bandwidth setting.
> > 
> > 
> > Why have a per-user limit? Should it not just be a global setting?
> 
> Ultimately, yes as it then would be the same as with the other bandwidth setting, however i did not had enough time to implement that aswell yet.

I don't think it makes sense to have this kind of granularity. I would approve a global setting since I can see it helping admins with limited upload speed. But having it per-user seems cumbersome and I don't see the need.

JPVenson:
Then why do we already have per-user bandwidth settings in addition to a global one?

cvium:
> Then why do we already have per-user bandwidth settings in addition to a global one?

Legacy? I don't agree with that one either, but keep in mind I'm just one person.

JPVenson:
No worries. I can just think of a good reason to keep both because for me the "localnetwork" detection does not work as i am a. behind a reverse proxy aswell as b. run a complex docker setup, where this is a great workaround to throttle external users only

cvium:
Shouldn't we fix the local detection instead then?

JPVenson:
Other topic, but that is not really fixable from anyone ^^. Its an issue with docker that runs inside WSL2 where WSL2 does proxy _everything_ and therefor the IP changes completely. Nothing JF can do. Its a total PITA but nothing we can do.
Just saying that there _are_ usecases for per-user bandwidth settings

EraYaN:
Well that just means people need to setup up their networking correctly, we can absolutely support forwarded for headers and the like to work behind a reverse proxy.

JPVenson:
> Well that just means people need to setup up their networking correctly, we can absolutely support forwarded for headers and the like to work behind a reverse proxy.

I dont really understand here. Reverse proxies behind bridged WSL2 networks are just not possible. Its a known limitation of the platform. Also not all RP's support bandwidth limiting.

EraYaN:
My point being that bridged WSL is not a correct network setup to host any services. Nothing to do with bandwidth limiting. If you want to run on a windows box run the Windows version. And if you want to run on Linux, run an actual linux box with proper networking.

JPVenson:
> My point being that bridged WSL is not a sane network setup to host any services. Nothing to do with bandwidth limiting. If you want to run on a windows box run the Windows version. And if you want to run on Linux, run an actual linux box with proper networking.

I dont think barring users from running on this or that is not a great idea. And i have to agree, we are getting lost in a side discussion here. Finally i want to say i have good reasons to do that, ping me on matrix if you want to hear them. 

If i get your current opinion on this feature correct, you are saying there is no need for it?

EraYaN:
I feel like this should be a global limit yes. IF people want local vs remote detection to work they will have to content with running sane server configurations.

JPVenson:
> I feel like this should be a global limit yes. IF people want local vs remote detection to work they will have to content with running sane server configurations.

Ok but to be clear, i dont implemented local vs remote limits, this PR adds user specific download restrictions. This PR does not differentiate between local network and external.

If there would be a global limit in addition to local ones, would that be fine with you?

EraYaN:
I still feel like its a bit of a work around and it opens up a ton of debugging problems if people forget that they have stuff configured on a per user basis. Especially since with don't really support groups for that. And our UI is just not very good at doing discovery of those things. I feel like it'd just be extra complexity that is not required.

And as for the global limit it can just work the same as the stream limit, which I believe is per stream (and only for external users). I called it "global" for it's configuration location not really for it's shared global resource nature.

cvium:
As I said before, I am totally on board with a throttle per stream that is set globally. I don't understand the use case for limiting specific users though.

JPVenson:
> I still feel like its a bit of a work around and it opens up a ton of debugging problems if people forget that they have stuff configured on a per user basis. 

I really dont see that issue, as this explanation is way to general in this case. You could apply it to nearly all features present in JF. A Per-User download limit is very self explanatory and easily debuggable with logs and just a single question: "Do all users have problems with downloads or just that one? If so check setting so-and-so" 

> As I said before, I am totally on board with a throttle per stream that is set globally

The current implementation is not that good for low-bandwidth scenarios like mine. I have 30mbit upload speed, and no matter what transcode limit i want to set, if its to low all users will have very bad streams even if there would be sufficient bandwidth and if i set it to high just a very small number of concurrent transcodes will saturate the available bandwidth. 
As i said, the only reasonable global limit would be a server-wide one, where a set amount of bandwidth can be split between users. Which is totally doable with downloads, just a lot of work.

If both of you still dont see the benefit, please say so.

EraYaN:
Doing full bandwidth server-wide management is way outside the scope of this PR, I'd be all for it, but that needs a ton of work and also updating transcoding bandwidth settings or adaptive streaming etc.  That is also something that would warrant some discussion in https://github.com/jellyfin/jellyfin-meta I feel and needs a proper design and quite some iteration. The current global settings are a balance between average concurrent users and total bandwidth available. Which I feel like is acceptable given how few people actually run very large installs with at the same time large changes in active users.

I still feel like it shouldn't be required at all to configure this stuff per user, it's just unnecessary complexity. Arguably the whole download thing is kind of a weird use-case anyway within Jellyfin but alas.

PR #9944, opened 2023-07-01
Close streams with no sessions on playback and session events

### LINKED ISSUE #9690: [Issue]: Jellyfin does not unsubscribe from Live TV streams
### Please describe your bug

I am trying to connect Jellyfin to my M3U/XML TV server. It mostly works but Jellyfin does not seem to unsubscribe from the HTTP stream when the content is closed. The TV server then thinks someone is still watching and it's impossible to switch to any other channel as it is not possible for the tuner to connect to multiple channels at the same time.

### Jellyfin Version

10.8.z

### if other:

_No response_

### Environment

```markdown
- OS: Windows 10
- Linux Kernel: none
- Virtualization: none
- Clients: Browser, Android
- Browser: Firefox
- FFmpeg Version: Unsure
- Playback Method: Transcode
- GPU Model: GTX1050Ti
- Plugins: none
- Reverse Proxy: no
- Base URL: no
- Networking: host
- Storage: n/a
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

DillonVarone:
Some of the Jellyfin clients (such as Android Mobile) do not seem to unsubscribe from Live Streams after they are closed. This can result in streams remaining open on the server indefinitely, which causes issues with M3UTuners that have a fixed number of active streams allowed. The server should be responsible for cleaning up after any irresponsible clients, or connection failures in this case.

**Changes**
Adds a new function to SessionManager to close any streams that are not currently "claimed" by a session. Adds behavior to trigger this cleanup on events that would require it such as playback stopping or sessions ending. Also allows the cleanup to be toggled for the case of Live TV Recording, which does not need to be cleaned up.

**Issues**
Fixes #9690

### DIFF
--- a/Emby.Server.Implementations/Session/SessionManager.cs
+++ b/Emby.Server.Implementations/Session/SessionManager.cs
@@ -323,6 +323,10 @@ namespace Emby.Server.Implementations.Session
                 {
                     await _mediaSourceManager.CloseLiveStream(session.PlayState.LiveStreamId).ConfigureAwait(false);
                 }
+                else
+                {
+                    await CleanupDanglingStreams();
+                }
 
                 OnSessionEnded(session);
             }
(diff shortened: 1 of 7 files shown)


JPVenson:
I see no way that `allowCleanup` to be ever set to true in the code

DillonVarone:
> I see no way that `allowCleanup` to be ever set to true in the code

It gets set to true or false directly in the call to OpenLiveStreamInternal.  In the call from MediaSourceManager/OpenLiveStream it is hardcoded true as that codepath is only hit when a client requests the stream, and its hardcoded to false from the call in EmbyTV/RecordStream as it shouldn't be needed or work as intended for streams opened by the server.

Emby.Server.Implementations/Session/SessionManager.cs
```diff
@@ -578,6 +582,66 @@ private void StopIdleCheckTimer()
...
+                            isStreamActive[streamIdx] = true;
+                            break;
+                        }
+                    }
+
+                    if (!isStreamActive[streamIdx])
+                    {
+                        try
+                        {
+                            await _mediaSourceManager.CloseLiveStream(liveStream.Key).ConfigureAwait(false);
```
IDisposable, line 632: Would be better to queue up all the `CloseLiveStream` returned Task and then Await them all after the loop, or even not await them at all since this is really a fire-and-forget operation.

Emby.Server.Implementations/Session/SessionManager.cs
```diff
@@ -945,18 +1025,6 @@ public async Task OnPlaybackStopped(PlaybackStopInfo info)
                 }
             }
 
-            if (!string.IsNullOrEmpty(info.LiveStreamId))
```
IDisposable, line 948: Seems like maybe this should be extracted to a helper method, not copy-pasted in a couple places.

Bond-009:
Closing as stale

PR #9951, opened 2023-07-02
Store lyrics in the database as media streams

crobibero:
**Changes**

Store the external lyric files as media streams like subtitles.

**Issues**

Working plugin: https://github.com/crobibero/jellyfin-plugin-lrclib


MediaBrowser.Providers/Lyric/ILyricProvider.cs
```diff
@@ -20,17 +21,10 @@ public interface ILyricProvider
...
-    bool HasLyrics(BaseItem item);
-
     /// <summary>
     /// Gets the lyrics.
     /// </summary>
-    /// <param name="item">The media item.</param>
+    /// <param name="path">The lyric path.</param>
     /// <returns>A task representing found lyrics.</returns>
-    Task<LyricFile?> GetLyrics(BaseItem item);
+    Task<LyricFile?> GetLyrics(string path);
```
nielsvanvelzen, line 29: This will break the plugin support I added in #9920. With BaseItem a plugin can find lyrics based on metadata (artist/track name) from various sources, including the file system but also online services etc.

crobibero: Good call, I've reverted that change and now plugins can specify whether the item has lyrics and can also provide them

nielsvanvelzen:
I think it would make more sense to have the LyricManager store the results in the database and not make the providers responsible for that. With the current approach a cloud-provider would still hold up item responses.

DomiStyle:
What's the preferred way to deal with embedded lyrics? They are currently not stored in database either.

They are provided by TagLib (`tags.Lyrics`) though, should they be added as media streams as well if available?

JPVenson (approved):

Gauvino (approved):

NicholasFlamy:
Thanks for your work. Can't wait! 👏👏👏

Shadowghost (approved):

Jellyfin.Api/Models/LyricDtos/UploadLyricDto.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// Gets or sets the subtitle format.
+    /// </summary>
+    [Required]
+    public string Format { get; set; } = string.Empty;
+
+    /// <summary>
+    /// Gets or sets a value indicating whether the lyric is synced.
+    /// </summary>
+    [Required]
+    public bool IsSynced { get; set; }
```
nielsvanvelzen, line 20: I think this isn't used

Jellyfin.Api/Controllers/LyricController.cs
```diff
@@ -0,0 +1,248 @@
...
+    /// Gets the remote lyrics.
+    /// </summary>
+    /// <param name="id">The remote provider item id.</param>
+    /// <response code="200">File returned.</response>
+    /// <returns>A <see cref="FileStreamResult"/> with the lyric file.</returns>
+    [HttpGet("Providers/Lyrics/{id}")]
+    [Authorize(Policy = Policies.LyricManagement)]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    [ProducesFile(MediaTypeNames.Text.Plain)]
+    public async Task<ActionResult> GetRemoteLyrics([FromRoute, Required] string id)
```
nielsvanvelzen, line 243:
```suggestion
    public async Task<ActionResult> GetRemoteLyrics([FromRoute, Required] string lyricId)
```

More consistent with downloadRemoteLyrics and indicative of what id we want

Jellyfin.Api/Controllers/LyricController.cs
```diff
@@ -0,0 +1,248 @@
...
+    {
+        var result = await _lyricManager.GetRemoteLyricsAsync(id, CancellationToken.None).ConfigureAwait(false);
```
nielsvanvelzen, line 245: What happens if the remote lyrics are not found or the resolver throws?

Jellyfin.Api/Controllers/LyricController.cs
```diff
@@ -0,0 +1,248 @@
...
+    /// Downloads a remote lyric.
+    /// </summary>
+    /// <param name="itemId">The item id.</param>
+    /// <param name="lyricId">The lyric id.</param>
+    /// <response code="204">Lyric downloaded.</response>
+    /// <returns>A <see cref="NoContentResult"/>.</returns>
+    [HttpPost("Items/{itemId}/RemoteSearch/Lyrics/{lyricId}")]
+    [Authorize(Policy = Policies.LyricManagement)]
+    [ProducesResponseType(StatusCodes.Status204NoContent)]
+    public async Task<ActionResult> DownloadRemoteLyrics(
```
nielsvanvelzen, line 209: Missing 404 response type

Jellyfin.Api/Controllers/LyricController.cs
```diff
@@ -0,0 +1,248 @@
...
+    /// <summary>
+    /// Search remote lyrics.
+    /// </summary>
+    /// <param name="itemId">The item id.</param>
+    /// <response code="200">lyrics retrieved.</response>
+    /// <returns>An array of <see cref="RemoteLyricInfo"/>.</returns>
+    [HttpGet("Items/{itemId}/RemoteSearch/Lyrics")]
+    [Authorize(Policy = Policies.LyricManagement)]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    public async Task<ActionResult<IEnumerable<RemoteLyricInfo>>> SearchRemotelyrics([FromRoute, Required] Guid itemId)
```
nielsvanvelzen, line 188: Missing 404 response type

Jellyfin.Api/Controllers/LyricController.cs
```diff
@@ -0,0 +1,248 @@
...
+    /// Gets an item's lyrics.
+    /// </summary>
+    /// <param name="itemId">Item id.</param>
+    /// <response code="200">Lyrics returned.</response>
+    /// <response code="404">Something went wrong. No Lyrics will be returned.</response>
+    /// <returns>An <see cref="OkResult"/> containing the item's lyrics.</returns>
+    [HttpGet("Audio/{itemId}/Lyrics")]
+    [Authorize]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    public async Task<ActionResult<LyricModel>> GetLyrics([FromRoute, Required] Guid itemId)
```
nielsvanvelzen, line 77: Missing 404 response type

MediaBrowser.Model/Providers/RemoteLyricInfo.cs
```diff
@@ -0,0 +1,24 @@
+using MediaBrowser.Model.Lyrics;
+
+namespace MediaBrowser.Model.Providers;
+
+/// <summary>
+/// The remote lyric info.
+/// </summary>
+public class RemoteLyricInfo
```
nielsvanvelzen, line 8: I think it would be worth adding the actual lyrics as well. This can be used to "preview" the lyrics when searching lyrics manually during playback.

nielsvanvelzen (changes requested):

Jellyfin.Api/Controllers/LyricController.cs
crobibero: If a provider errors it won't be added to the list
I'm not sure what you mean by resolver here?

nielsvanvelzen: Sorry I meant provider. When I call this endpoint with some random id that doesn't exist, what happens? I didn't see any checking here to respond with a 404.

Jellyfin.Api/Controllers/LyricController.cs
```diff
@@ -0,0 +1,252 @@
...
+        [FromRoute, Required] Guid itemId,
+        [FromBody, Required] UploadLyricDto body)
+    {
+        var audio = _libraryManager.GetItemById<Audio>(itemId);
+        if (audio is null)
+        {
+            return NotFound();
+        }
+
+        var stream = new MemoryStream(Encoding.UTF8.GetBytes(body.Data));
```
Bond-009, line 131:
```suggestion
        var bytes = Encoding.UTF8.GetBytes(body.Data);
        var stream = new MemoryStream(bytes, 0, bytes.Length, false, true);
```
We can expose the internal buffer for this MemoryStream, may speed something up down the line

Jellyfin.Api/Controllers/LyricController.cs
crobibero: Good call- this endpoint will now respond with a 404 if the provider isn't found

Jellyfin.Api/Controllers/SubtitleController.cs
```diff
@@ -407,7 +406,11 @@ await using (stream.ConfigureAwait(false))
         [FromBody, Required] UploadSubtitleDto body)
     {
         var video = (Video)_libraryManager.GetItemById(itemId);
-        var stream = new CryptoStream(Request.Body, new FromBase64Transform(), CryptoStreamMode.Read);
+
+        var memoryStream = new MemoryStream(Encoding.UTF8.GetBytes(body.Data));
+        using var transform = new FromBase64Transform();
+        var stream = new CryptoStream(memoryStream, transform, CryptoStreamMode.Read);
+        await using (memoryStream.ConfigureAwait(false))
```
Bond-009, line 413: https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync#unacceptable-pattern

Bond-009 (changes requested):

Jellyfin.Api/Controllers/LyricController.cs
```diff
@@ -0,0 +1,252 @@
...
+    /// </summary>
+    /// <param name="itemId">The item the lyric belongs to.</param>
+    /// <param name="body">The request body.</param>
+    /// <response code="204">Lyrics uploaded.</response>
+    /// <response code="400">Error processing upload.</response>
+    /// <response code="404">Item not found.</response>
+    /// <returns>The uploaded lyric.</returns>
+    [HttpPost("Audio/{itemId}/Lyrics")]
+    [Authorize(Policy = Policies.LyricManagement)]
+    [ProducesResponseType(StatusCodes.Status204NoContent)]
```
nielsvanvelzen, line 118: Should be 200 now

MediaBrowser.Model/Lyrics/UploadLyricDto.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// Gets or sets the lyric format, typically the file extension.
+    /// </summary>
+    [Required]
+    public string Format { get; set; } = string.Empty;
+
+    /// <summary>
+    /// Gets or sets the lyric data, plaintext.
+    /// </summary>
+    [Required]
+    public string Data { get; set; } = string.Empty;
```
nielsvanvelzen, line 21: I think "content" might be more descriptive.

MediaBrowser.Model/Lyrics/LyricModel.cs
```diff
@@ -1,12 +1,11 @@
...
 using System.Collections.Generic;
 
-namespace MediaBrowser.Controller.Lyrics;
+namespace MediaBrowser.Model.Lyrics;
 
 /// <summary>
 /// LyricResponse model.
 /// </summary>
-public class LyricResponse
+public class LyricModel
```
nielsvanvelzen, line 8:
```suggestion
public class LyricDto
```

This would make sense for the API part but might not make much sense for the other usages

Jellyfin.Api/Controllers/LyricController.cs
```diff
@@ -0,0 +1,253 @@
...
+    [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    public async Task<ActionResult<LyricModel>> UploadLyric(
```
nielsvanvelzen, line 121:
```suggestion
    public async Task<ActionResult<LyricModel>> UploadLyrics(
```

Consistency with other endpoints

MediaBrowser.Model/Lyrics/RemoteLyricInfoDto.cs
```diff
@@ -0,0 +1,27 @@
...
+
+    /// <summary>
+    /// Gets the provider name.
+    /// </summary>
+    public required string ProviderName { get; init; }
+
+    /// <summary>
+    /// Gets the lyric metadata.
+    /// </summary>
+    public required LyricMetadata Metadata { get; init; }
```
nielsvanvelzen, line 21: Metadata is also included in LyricModel so I guess we can drop it here

nielsvanvelzen (changes requested):

MediaBrowser.Model/Configuration/LibraryOptions.cs
```diff
@@ -92,6 +93,8 @@ public LibraryOptions()
 
         public bool SaveSubtitlesWithMedia { get; set; }
 
+        public bool SaveLyricsWithMedia { get; set; }
```
nielsvanvelzen, line 96: Add default value to OpenAPI spec

nielsvanvelzen (approved):
✔️ API changes

[Kotlin SDK diff](https://github.com/jellyfin/jellyfin-sdk-kotlin/compare/08c55eb27d45e195a4a229a6abb012628014bf1a...nielsvanvelzen:jellyfin-sdk-kotlin:tst-lyrics)

MediaBrowser.Providers/Lyric/LyricManager.cs
```diff
@@ -11,60 +27,428 @@ namespace MediaBrowser.Providers.Lyric;
...
+
+    /// <inheritdoc />
+    public async Task<LyricDto?> DownloadLyricsAsync(Audio audio, LibraryOptions libraryOptions, string lyricId, CancellationToken cancellationToken)
+    {
+        ArgumentNullException.ThrowIfNull(audio);
+        ArgumentNullException.ThrowIfNull(libraryOptions);
+        ArgumentException.ThrowIfNullOrWhiteSpace(lyricId);
+
+        var parts = lyricId.Split('_', 2);
+        var provider = GetProvider(parts[0]);
```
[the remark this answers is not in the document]
crobibero: Can't use `ReadOnlySpan` in async methods

I can create some string versions of `LeftPart` and `RightPart` if desired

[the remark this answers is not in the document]
Bond-009: You can use it, you just can't have a ref struct variable in an async method
```suggestion
        var provider = GetProvider(lyricId.AsSpan().LeftPart('_').ToString());
```
This should work (if you add `using Jellyfin.Extensions;`)

string versions of `LeftPart` and `RightPart` would make this a bit cleaner, but I'm not sure if that's needed

MediaBrowser.Providers/MediaInfo/MediaInfoResolver.cs
```diff
@@ -151,6 +152,48 @@ public abstract class MediaInfoResolver
...
+            }
+
+            var pathInfos = GetExternalFiles(audio, directoryService, clearCache);
+
+            if (pathInfos.Count == 0)
+            {
+                return Array.Empty<MediaStream>();
+            }
+
+            var mediaStreams = new List<MediaStream>(pathInfos.Count);
```
Bond-009, line 181: Can't this be an array?

MediaBrowser.Providers/MediaInfo/MediaInfoResolver.cs
```diff
@@ -151,6 +152,48 @@ public abstract class MediaInfoResolver
...
+                mediaStreams.Add(new MediaStream
+                {
+                    Type = MediaStreamType.Lyric,
+                    Path = pathInfo.Path,
+                    Language = pathInfo.Language,
+                    Index = startIndex++
+                });
+            }
+
+            return mediaStreams.AsReadOnly();
```
Bond-009, line 194: Normally we don't use `AsReadOnly()` bc every time you use it afterwards everything gets proxied trough `ReadOnlyCollection` which slows everything down

MediaBrowser.Providers/Lyric/LyricManager.cs
```diff
@@ -11,60 +28,427 @@ namespace MediaBrowser.Providers.Lyric;
...
+    /// <inheritdoc />
+    public async Task<IReadOnlyList<RemoteLyricInfoDto>> SearchLyricsAsync(LyricSearchRequest request, CancellationToken cancellationToken)
+    {
+        ArgumentNullException.ThrowIfNull(request);
+
+        var providers = _lyricProviders
+            .Where(i => !request.DisabledLyricFetchers.Contains(i.Name, StringComparer.OrdinalIgnoreCase))
+            .OrderBy(i =>
+            {
+                var index = request.LyricFetcherOrder.ToList().IndexOf(i.Name);
```
Bond-009, line 98: Maybe this `ToList` should happen outside of the `OrderBy`?

Bond-009 (changes requested):

MediaBrowser.Providers/Lyric/LyricManager.cs
crobibero: Turns out I had already added a `.IndexOf` for `IReadOnlyList`

That's what I get from just pulling code from the SubtitleManager

Bond-009 (approved):

nielsvanvelzen:
I will need some extra time to review with the latest changes because the Kotlin SDK does not currently support multipart/form-data so I need to implement that first.

nielsvanvelzen (changes requested):
The schema for the requestBody of the upload endpoint is now generated inline which is impossible to work with in the Kotlin SDK.

nielsvanvelzen (approved):
✔️ API changes

[Kotlin SDK diff](https://github.com/jellyfin/jellyfin-sdk-kotlin/compare/08c55eb27d45e195a4a229a6abb012628014bf1a...nielsvanvelzen:jellyfin-sdk-kotlin:tst-lyrics)

DomiStyle:
I would still like to add support for embedded lyrics but I'm unsure what the preferred way to achieve this with the new structure is.

Some things I thought of:

* Add a new `MediaStreamType`: `EmbeddedLyrics`
* Add a `Lyrics` or `EmbeddedLyrics` attribute to the `Audio` class - probably problematic since all the lyrics are assumed to be a `MediaStream` now
* Add the embedded lyrics as text directly as `MediaStreamType.Lyric` - would need special handling to separate lyrics that are paths vs lyrics that are text
* Add the path of the audio file as `MediaStreamType.Lyric` directly instead of to the `.lrc` file and get the embedded lyrics when lyrics are requested via the API - probably not ideal since the tags have to be read every time the lyrics are requested

Any preferred way to do this before I start work on a PR?

NicholasFlamy:
> I would still like to add support for embedded lyrics but I'm unsure what the preferred way to achieve this with the new structure is.
> 
> Some things I thought of:
> 
> * Add a new `MediaStreamType`: `EmbeddedLyrics`
> * Add a `Lyrics` or `EmbeddedLyrics` attribute to the `Audio` class - probably problematic since all the lyrics are assumed to be a `MediaStream` now
> * Add the embedded lyrics as text directly as `MediaStreamType.Lyric` - would need special handling to separate lyrics that are paths vs lyrics that are text
> * Add the path of the audio file as `MediaStreamType.Lyric` directly instead of to the `.lrc` file and get the embedded lyrics when lyrics are requested via the API - probably not ideal since the tags have to be read every time the lyrics are requested
> 
> Any preferred way to do this before I start work on a PR?

I seem to remember reading that the embedded metadata handling was going to be refactored or something.

crobibero:
@DomiStyle @NicholasFlamy please test out #11081

I only have audio files that I manually embedded lyrics into .mp3 and .flac using MediaMonkey5

NicholasFlamy:
> please test out #11081

I am not familiar with how to download the build artifacts from a Jellyfin PR.

NicholasFlamy:
Either this or #11081 might be causing Media Info to be blank.
Screenshot 2024-03-08 113309
It only happens to songs with lyrics, embedded or LRC file.

I just tested this. Some songs with embedded lyrics won't show Media Info, they look like that screenshot above. Then I tested a song that was encoded with the exact same encoder and settings and the same resolution (1200x1200) mozjpeg-encoded embedded cover art but no embedded lyrics or LRC file. Then I copied a random LRC file and moved it to the directory with this song with no lyrics and renamed the LRC file to match the song (but with the .lrc extension rather than .m4a of course) and rescanned and refreshed Jellyfin and the song had the same problem as the one with embedded lyrics.

My encodes are using these FFmpeg args `-c:a libfdk_aac -vbr 4 -cutoff 20000 -ar 44100 -movflags +faststart` in case it's specific to this.

I'm running the unstable docker image in TrueNAS Scale. I did not test this before these two issues were merged so I cannot confirm that these caused it.

Screenshot of song without lyrics (no embedded lyrics and no .lrc file with the same name as the file)
image

crobibero:
@NicholasFlamy please create an issue will full steps so it can be tracked appropriately

NicholasFlamy:
> @NicholasFlamy please create an issue will full steps so it can be tracked appropriately

#11130 Gotcha

PR #9958, opened 2023-07-03
Extract networking constants

Shadowghost:
Addresses changes proposed in #8147 after it got merged

**Changes**
* Extract networking constants
* Move networking extension to correct place


cvium (approved):

Jellyfin.Networking/Extensions/NetworkExtensions.cs
```diff
@@ -0,0 +1,351 @@
...
+
+/// <summary>
+/// Defines the <see cref="NetworkExtensions" />.
+/// </summary>
+public static partial class NetworkExtensions
+{
+    // Use regular expression as CheckHostName isn't RFC5892 compliant.
+    // Modified from gSkinner's expression at https://stackoverflow.com/questions/11809631/fully-qualified-domain-name-validation
+    [GeneratedRegex(@"(?im)^(?!:\/\/)(?=.{1,255}$)((.{1,63}\.){0,127}(?![0-9]*$)[a-z0-9-]+\.?)(:(\d){1,5}){0,1}$", RegexOptions.IgnoreCase, "en-US")]
+    private static partial Regex fqdnGeneratedRegex();
```
Bond-009, line 22: Pascal casing for functions

Jellyfin.Networking/Extensions/NetworkExtensions.cs
```diff
@@ -0,0 +1,351 @@
...
+        if (host.IsEmpty)
+        {
+            addresses = null;
+            return false;
+        }
+
+        host = host.Trim();
+
+        // See if it's an IPv6 with port address e.g. [::1] or [::1]:120.
+        if (host[0] == '[')
```
Bond-009, line 268: Trim before checking if empty, now it will throw when given only whitespace

Jellyfin.Networking/Extensions/NetworkExtensions.cs
```diff
@@ -0,0 +1,351 @@
...
+            {
+                try
+                {
+                    addresses = Dns.GetHostAddresses(hosts[0]);
+                    return true;
+                }
+                catch (SocketException)
+                {
+                    // Log and then ignore socket errors, as the result value will just be an empty array.
+                    Console.WriteLine("GetHostAddresses failed.");
```
Bond-009, line 299: Just remove this, it won't end up in our logs and doesn't tell us much

Bond-009 (changes requested):

Jellyfin.Networking/Extensions/NetworkExtensions.cs
```diff
@@ -337,14 +335,9 @@ public static bool TryParseHost(ReadOnlySpan<char> host, [NotNullWhen(true)] out
     public static IPAddress GetBroadcastAddress(IPNetwork network)
     {
         var addressBytes = network.Prefix.GetAddressBytes();
-        if (BitConverter.IsLittleEndian)
-        {
-            addressBytes = addressBytes.Reverse().ToArray();
-        }
```
Shadowghost: This was removed because the byte order returned by `GetAddressBytes()` is always big endian:
https://stackoverflow.com/questions/1407938/ipaddress-getaddressbytes-method-what-byte-order

Jellyfin.Networking/Constants/Network.cs
```diff
@@ -0,0 +1,75 @@
...
+
+    /// <summary>
+    /// IPv4 Loopback as defined in RFC 5735.
+    /// </summary>
+    public static readonly IPNetwork IPv4RFC5735Loopback = new IPNetwork(IPAddress.Loopback, 8);
+
+    /// <summary>
+    /// IPv4 private class A as defined in RFC 1918.
+    /// </summary>
+    public static readonly IPNetwork IPv4RFC1918PrivateClassA = new IPNetwork(IPAddress.Parse("10.0.0.0"), 8);
```
Bond-009, line 49: Using the ctor that takes a byte array is prob faster than parsing a string

Bond-009 (approved):

PR #9992, opened 2023-07-12
Change: Add TinyMediaManager compatibility for german parental rating

Blackskyliner:
**Changes**
This PR adds the format which gets written by
 the tool TinyMediaManager which can be used to
 manage large media databases comfortably.

TMM writes wither multiple Codes and/or space
 divided german FSK codes, but can't be configured
 to only use the number or the slash delimited variant.

After this change the parental control for Libraries
 managed with TMM and presented/loaded into Jellyfin
 should work again.

### DIFF
--- a/Emby.Server.Implementations/Localization/Ratings/de.csv
+++ b/Emby.Server.Implementations/Localization/Ratings/de.csv
@@ -1,12 +1,17 @@
 Educational,0
 Infoprogramm,0
 FSK-0,0
+FSK 0,0
 0,0
 FSK-6,6
+FSK 6,6
 6,6
 FSK-12,12
+FSK 12,12
 12,12
 FSK-16,16
+FSK 16,16
 16,16
 FSK-18,18
+FSK 18,18
 18,18
--- a/tests/Jellyfin.Server.Implementations.Tests/Localization/LocalizationManagerTests.cs
+++ b/tests/Jellyfin.Server.Implementations.Tests/Localization/LocalizationManagerTests.cs
@@ -100,7 +100,7 @@ namespace Jellyfin.Server.Implementations.Tests.Localization
             await localizationManager.LoadAll();
             var ratings = localizationManager.GetParentalRatings().ToList();
 
-            Assert.Equal(19, ratings.Count);
+            Assert.Equal(24, ratings.Count);
 
             var fsk = ratings.FirstOrDefault(x => x.Name.Equals("FSK-12", StringComparison.Ordinal));
             Assert.NotNull(fsk);


Shadowghost (approved):

Bond-009 (approved):

Hallo951:
Can you please check if this change (commit db347eb) is implemented in the current jellyfin version (10.8.13)? Because it does not work!!!

cvium:
It isn't in 10.8

ramsnerm:
When can it be expected that the German ratings are available in Jellyfin? THX

Blackskyliner:
It should be in the current 10.9.x release. I seem to have no problems in my library anymore after upgrading and not maintaining/applying my private custom patchset for 10.8.x anymore.

Blackskyliner:
The following TMM Setting must be used though.

image

ramsnerm:
So understanding you right: This change takes only effect in TMM and will not change the DropDown in Jellyfin itself (although I have set German as language)?

Blackskyliner:
Correct, the change only adds support for the metadata format which gets written by TMM into the NFO file which is then later read by jellyfin. In my instance it currently looks like this (which also seems kinda messy though I guess it reads all already used formats within the database and makes them available for selection):

image

I also think you need to correctly set the metadata settings to germany if not done already:

image

ramsnerm:
Thanks for the explanation. I checked my settings. I had German, Austria. When I change to German,Germany I get the German possibilities. I made then a test and set for example Switzerland as Country and it has the same Issue than Austria - only English age ratings are available.

So I think there is still a minor mistakes in the program, not considering that "FSK" is also applicable for Austria, Switzerland and some other countries. I think this modification is not complicate and should be considered in any of the upcoming updates.

In the meanwhile I take the workaround and take as Region "Germany".

@Blackskyliner Thanks for your support.

Blackskyliner:
Each country has their own official ratings list as .csv files (like the one I changes + the test so it can get merged).
I don't know if it makes much sense to now bloat up the ratings .csv files for all languages which support the FSK on that basis.

If I would have been a bit more knowledgable I would have patched the csv parsing to handle "FSK-12" and "FSK 12" the same but adding the csv values was faster and added less potential friction for breakages.

But if FSK is a thing all those both countries I don't see why whe can't/shouldn't be able to add them just as I did to de.csv as valid use case -- Just don't know if this is something the one how refactored the age restriction had in mind here. So yea, an issue seems to be the right intention to go through all possibilities.

The "simple" version I did for de.csv would be easy integratable in not much time at all.

ramsnerm:
Thanks for your perspective, I will setup a new issue.

PR #9996, opened 2023-07-13
Detect audio spatial format

mohd-akram:
Support was [added](https://github.com/FFmpeg/FFmpeg/compare/814178f...98a4699) to FFmpeg.

**Changes**
Add `AudioSpatialFormat` property to `MediaStream`.

### DIFF
--- a/MediaBrowser.Model/Entities/MediaStream.cs
+++ b/MediaBrowser.Model/Entities/MediaStream.cs
@@ -214,6 +214,26 @@ namespace MediaBrowser.Model.Entities
             }
         }
 
+        /// <summary>
+        /// Gets the audio spatial format.
+        /// </summary>
+        /// <value>The audio spatial format.</value>
+        public AudioSpatialFormat AudioSpatialFormat
+        {
+            get
+            {
+                if (Type != MediaStreamType.Audio || string.IsNullOrEmpty(Profile))
+                {
+                    return AudioSpatialFormat.None;
+                }
+
+                return
+                    Profile.Contains("Dolby Atmos", StringComparison.OrdinalIgnoreCase) ? AudioSpatialFormat.DolbyAtmos :
+                    Profile.Contains("DTS:X", StringComparison.OrdinalIgnoreCase) ? AudioSpatialFormat.DTSX :
+                    AudioSpatialFormat.None;
+            }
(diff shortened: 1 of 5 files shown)


mohd-akram:
Thanks, rebased.

Bond-009 (approved):

MediaBrowser.Model/Entities/MediaStream.cs
```diff
@@ -214,6 +214,26 @@ public string VideoDoViTitle
             }
         }
 
+        /// <summary>
+        /// Gets the audio spatial format.
+        /// </summary>
+        /// <value>The audio spatial format.</value>
+        public AudioSpatialFormat AudioSpatialFormat
```
crobibero, line 221: This would need to be nullable for backwards compatibility?

@nielsvanvelzen

nielsvanvelzen: Should be nullable or have a default value

Shadowghost:
@mohd-akram can you please address the proposed changes so we can get this merged? Thank you!

MediaBrowser.Model/Entities/MediaStream.cs
mohd-akram: I had tried to make it nullable before but it was problematic. What do you mean by a default value? I made it a getter like `VideoRange` and `VideoRangeType`, both of which are also enums with a similar zero value like `None`.

crobibero: Adding this attribute to this property changes the api spec
`[DefaultValue(AudioSpatialFormat.None)]`

```diff
"AudioSpatialFormat": {
	"enum": [
	  "None",
	  "DolbyAtmos",
	  "DTSX"
	],
	"allOf": [
	  {
		"$ref": "#/components/schemas/AudioSpatialFormat"
	  }
	],
	"description": "Gets the audio spatial format.",
+	"default": "None",
	"readOnly": true
},
```

crobibero, line 221:
```suggestion
        [DefaultValue(AudioSpatialFormat.None)]
        public AudioSpatialFormat AudioSpatialFormat
```

PR #10011, opened 2023-07-17
Prefer unplayed items when sorting randomly

DavidMikeSimon:
**Changes**
This PR adjusts the "Random" sorting strategy to prefer unplayed items over played items, as a simple way of addressing [feature request 717](https://features.jellyfin.org/posts/717/shuffle-weight-recently-played-media-should-be-less-likely-to-be-next).

The idea here is that when a user hits the "Shuffle" button in any given context (e.g. a show, a library, a collection), they are probably not interested in seeing items that they have recently already played via shuffle. Tying the priority to the "Played" flag gives the user some control over this behavior, e.g. if they want to "reset" their shuffle exclusion, they can mark a folder as unplayed.

**Looking for feedback and suggestions**
I'm definitely open to suggestions about this approach and implementation! In particular, I notice that besides the `ORDER BY RANDOM()` strategy implemented in `SqliteItemRepository`, there are two other possible ways for items to be shuffled: `RandomComparer` and `ShuffleExtensions`. I'm unsure about whether these need to be adjusted as well, and I was unable to figure how they could get invoked from front-end actions.

**Issues**
https://features.jellyfin.org/posts/717/shuffle-weight-recently-played-media-should-be-less-likely-to-be-next

### DIFF
--- a/Emby.Server.Implementations/Data/SqliteItemRepository.cs
+++ b/Emby.Server.Implementations/Data/SqliteItemRepository.cs
@@ -2150,6 +2150,7 @@ namespace Emby.Server.Implementations.Data
                     || sortingFields.Contains(ItemSortBy.PlayCount)
                     || sortingFields.Contains(ItemSortBy.DatePlayed)
                     || sortingFields.Contains(ItemSortBy.SeriesDatePlayed)
+                    || sortingFields.Contains(ItemSortBy.Random) // Needed because random sorting uses IsPlayed
                     || query.IsFavoriteOrLiked.HasValue
                     || query.IsFavorite.HasValue
                     || query.IsResumable.HasValue
@@ -2946,7 +2947,7 @@ namespace Emby.Server.Implementations.Data
                 if (hasSimilar)
                 {
                     prepend.Add(("SimilarityScore", SortOrder.Descending));
-                    prepend.Add((ItemSortBy.Random, SortOrder.Ascending));
+                    prepend.Add((ItemSortBy.SortName, SortOrder.Ascending));
                 }
 
                 var arr = new (string, SortOrder)[prepend.Count + orderBy.Count];
@@ -2982,7 +2983,8 @@ namespace Emby.Server.Implementations.Data
 
             if (string.Equals(name, ItemSortBy.Random, StringComparison.OrdinalIgnoreCase))
             {
-                return "RANDOM()";
+                // Prefer unplayed items
+                return "((ABS(RANDOM()) % 65536) + (Select Case When played Then 65536 Else 0 End))";
             }
 
             if (string.Equals(name, ItemSortBy.DatePlayed, StringComparison.OrdinalIgnoreCase))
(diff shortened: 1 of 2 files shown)


Emby.Server.Implementations/Data/SqliteItemRepository.cs
```diff
@@ -2946,7 +2947,7 @@ private string GetOrderByText(InternalItemsQuery query)
                 if (hasSimilar)
                 {
                     prepend.Add(("SimilarityScore", SortOrder.Descending));
-                    prepend.Add((ItemSortBy.Random, SortOrder.Ascending));
+                    prepend.Add((ItemSortBy.SortName, SortOrder.Ascending));
```
DavidMikeSimon, line 2950: Prevent `SimilarityScore` from needing to join with user data. Alternatively, I could add one more check to `EnableJoinUserData`.

DavidMikeSimon:
@Bond-009 Apologies for the bump, just wanted to make sure I didn't fall through the cracks. Anything I need to do to bring this PR up to snuff? I'm adding this feature because it's helpful for my own family's viewing habits, and I don't want to be stuck running a fork forever. :-)

cvium:
I believe this is a decent change in behaviour for video, but what about music?

DavidMikeSimon:
@cvium Music is automatically marked as `played` just like video, right? If so, once the shuffle gets around to playing every unplayed song, everything would be back to equal footing.

One possibly surprising behavior though is that any new items added afterward would always be the next to appear in shuffle. This might be helpful or annoying depending on the user's expectations for shuffling.

cvium:
I can only speak for myself, but I would hate for unplayed songs to be favored over played

DavidMikeSimon:
@cvium Hm, would the best solution be to make this behavior conditionally only occur on non-music items?

DavidMikeSimon:
On further thought: Only applying the penalty to non-music (e.g. anything with a type other than `Audio`, `MusicVideo`, `MusicAlbum`, `MusicArtist`, `MusicGenre`) would mean that a mixed-type shuffle would eventually _always_ sort all the music items first. I'm not sure how often mixed-type shuffles would happen in practice, but if they did, that might be weird behavior.

Maybe the best way would be to add a new `ItemSortBy` value (e.g. `"RandomPreferUnplayed"`), and then it would be the client's responsibility to determine which kind of randomness it wants depending on context.

cvium:
> Maybe the best way would be to add a new `ItemSortBy` value (e.g. `"RandomPreferUnplayed"`), and then it would be the client's responsibility to determine which kind of randomness it wants depending on context.

Sounds better.

> Also it just now occurred to me that the client can already do that with a multi-column sort, yeah? 🤦

Could probably do it by sorting unplayed and then by random, but someone will have to test it

DavidMikeSimon:
Closing in favor of https://github.com/jellyfin/jellyfin-web/pull/4769

PR #10049, opened 2023-07-28
Fix format normalizer for multiple input formats

### LINKED ISSUE #10046: [Issue]: Matroska Direct Play is broken
### Please describe your bug

Matroska Direct Play is broken with following commit:  [47aa07c3424ce0041e0a79eea1ab7f6621485b94](https://github.com/jellyfin/jellyfin/commit/47aa07c3424ce0041e0a79eea1ab7f6621485b94)

Matroska container name normalization was changed from:

```
format = format.Replace("matroska", "mkv", StringComparison.OrdinalIgnoreCase);
```

to:

```
if (string.Equals(format, "matroska", StringComparison.OrdinalIgnoreCase))
{
    return "mkv";
}
```

ffmpeg reports `matroska,webm` for all mkv files. With old code it will be transformed to `mkv,webm` and it's good and with new code it will remain as `matroska,webm`.

### Jellyfin Version

10.8.z

### if other:

_No response_

### Environment

_No response_

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

Shadowghost:
Seems like FFprobe can return a list of formats.

**Changes**
* Split format input before running normalization and run normalization on all input formats

**Issues**
Fixes #10046

### DIFF
--- a/MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
+++ b/MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
@@ -1,5 +1,4 @@
 #nullable disable
-#pragma warning disable CS1591
 
 using System;
 using System.Collections.Generic;
@@ -20,6 +19,9 @@ using Microsoft.Extensions.Logging;
 
 namespace MediaBrowser.MediaEncoding.Probing
 {
+    /// <summary>
+    /// Class responsible for normalizing FFprobe output.
+    /// </summary>
     public class ProbeResultNormalizer
     {
         // When extracting subtitles, the maximum length to consider (to avoid invalid filenames)
@@ -36,6 +38,11 @@ namespace MediaBrowser.MediaEncoding.Probing
 
         private string[] _splitWhiteList;
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ProbeResultNormalizer"/> class.
+        /// </summary>
+        /// <param name="logger">The <see cref="ILogger{ProbeResultNormalizer}"/> for use with the <see cref="ProbeResultNormalizer"/> instance.</param>
+        /// <param name="localization">The <see cref="ILocalizationManager"/> for use with the <see cref="ProbeResultNormalizer"/> instance.</param>
         public ProbeResultNormalizer(ILogger logger, ILocalizationManager localization)
         {
             _logger = logger;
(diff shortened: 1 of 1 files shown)


MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
```diff
@@ -20,6 +19,9 @@
 
 namespace MediaBrowser.MediaEncoding.Probing
 {
+    /// <summary>
+    /// Class ProbeResultNormalizer.
```
cvium, line 23: Kind of pointless, no?

MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
```diff
@@ -73,6 +80,15 @@ public ProbeResultNormalizer(ILogger logger, ILocalizationManager localization)
             "Smith/Kotzen",
         };
 
+        /// <summary>
+        /// Function to transform a FFprobe response into its <see cref="MediaInfo"/> equivalent.
```
cvium, line 84:
```suggestion
        /// Transforms a FFprobe response into its <see cref="MediaInfo"/> equivalent.
```

MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
```diff
@@ -252,25 +268,37 @@ private string NormalizeFormat(string format)
                 return null;
             }
 
-            // Handle MPEG-1 container
-            if (string.Equals(format, "mpegvideo", StringComparison.OrdinalIgnoreCase))
+            var formatString = string.Empty;
+            var splitFormat = format.Split(',');
+            for (var i = 0; i < splitFormat.Length; i++)
```
cvium, line 273: This feels overly complicated. Can't we just use string.Replace?

Shadowghost: I think we can, not sure why I didn't use it initially...

Shadowghost: As mentioned by @dmitrylyzo in chat there can be collisions if we use replace, so I opted to not change the approach.

crobibero (commented):
This is a bit cleaner:
```c#
// Input can be a list of multiple, comma-delimited formats - each of them needs to be checked
var splitFormat = format.Split(',');
for (var i = 0; i < splitFormat.Length; i++)
{
    // Handle MPEG-1 container
    if (string.Equals(splitFormat[i], "mpegvideo", StringComparison.OrdinalIgnoreCase))
    {
        splitFormat[i] = "mpeg";
    }

    // Handle MPEG-2 container
    else if (string.Equals(splitFormat[i], "mpeg", StringComparison.OrdinalIgnoreCase))
    {
        splitFormat[i] = "ts";
    }

    // Handle matroska container
    else if (string.Equals(splitFormat[i], "matroska", StringComparison.OrdinalIgnoreCase))
    {
        splitFormat[i] = "mkv";
    }
}

return string.Join(',', splitFormat);
```

crobibero (approved):

MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
```diff
@@ -73,6 +80,15 @@ public ProbeResultNormalizer(ILogger logger, ILocalizationManager localization)
             "Smith/Kotzen",
         };
 
+        /// <summary>
+        /// Transforms a FFprobe response into its <see cref="MediaInfo"/> equivalent.
+        /// </summary>
+        /// <param name="data">The <see cref="InternalMediaInfoResult"/>.</param>
+        /// <param name="videoType">The <see cref="VideoType"/>.</param>
+        /// <param name="isAudio">A boolean indicatin wether the media is audio.</param>
```
cvium, line 88:
```suggestion
        /// <param name="isAudio">A boolean indicating whether the media is audio.</param>
```

cvium (approved):

Bond-009 (approved):

PR #10053, opened 2023-07-28
Add parsing for dates with spaces

christianlegge:
**Changes**
I've encountered titles with the date formatted using spaces between the numbers recently ("2023 07 28"). This adds that as a date format.

**Issues**
None


crobibero (approved):

Shadowghost (approved):

cvium:
> I also noticed a slight bug in the date parser whereby any character could be used to separate the numbers in a dd-mm-yyyy format (including numbers, which would cause any 10-digit number to be interpreted as a date). I've fixed that as well.

Are you sure? I don't see it

christianlegge:
@cvium My mistake - I didn't realize `.`s were literal inside character classes.

Shadowghost (approved):

cvium (approved):

PR #10101, opened 2023-08-08
fix-cast-with-localhost-server

### LINKED ISSUE #10098: [Issue]: Function required to allow webclient to play on chromecast using localhost server url
### Please describe your bug

The Jellyfin-web client needs the real lan address of the server to play on chromecast using localhost server url. See jellyfin-web issue https://github.com/jellyfin/jellyfin-web/issues/4745

**Solution**
The 'real' lan address is the IPv4 private network address, or the IPv6 ULA (unique local address).
See https://en.wikipedia.org/wiki/Private_network

Obtain this address while initializing the network and pass it as LocalAddress in the GetPublicSystemInfo server API call.

Note: this modification only applies if the server API is called from a 'localhost' url, as for other cases the url is assumed to be sufficiently specific to serve as unique address for the chromecast device.

### Jellyfin Version

10.8.z

### if other:

current github version 10.9.0

### Environment

```markdown
- OS: Windows 10
- Clients: Windows 10
- Browser: Chrome, Edge
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

Sky-High:
Adds a function to supply the webclient with a unique server LAN address after a localhost server API call. This enables the webclient to play to a chromecast receiver in such a case.

**Changes**
- Determine unique local server address in Jellyfin.Networking/Manager/NetworkManager.cs
- Pass this to the webclient with the GetPublicSystemInfo API call in Emby.Server.Implementations/ApplicationHost.cs

**Issues**
Fixes #10098

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -27,13 +27,13 @@
     <PackageVersion Include="MetaBrainz.MusicBrainz" Version="5.0.0" />
     <PackageVersion Include="Microsoft.AspNetCore.Authorization" Version="7.0.12" />
     <PackageVersion Include="Microsoft.AspNetCore.HttpOverrides" Version="2.2.0" />
-    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Testing" Version="7.0.11" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Testing" Version="7.0.12" />
     <PackageVersion Include="Microsoft.CodeAnalysis.BannedApiAnalyzers" Version="3.3.4" />
-    <PackageVersion Include="Microsoft.Data.Sqlite" Version="7.0.11" />
-    <PackageVersion Include="Microsoft.EntityFrameworkCore.Design" Version="7.0.11" />
-    <PackageVersion Include="Microsoft.EntityFrameworkCore.Relational" Version="7.0.11" />
-    <PackageVersion Include="Microsoft.EntityFrameworkCore.Sqlite" Version="7.0.11" />
-    <PackageVersion Include="Microsoft.EntityFrameworkCore.Tools" Version="7.0.11" />
+    <PackageVersion Include="Microsoft.Data.Sqlite" Version="7.0.12" />
+    <PackageVersion Include="Microsoft.EntityFrameworkCore.Design" Version="7.0.12" />
+    <PackageVersion Include="Microsoft.EntityFrameworkCore.Relational" Version="7.0.12" />
+    <PackageVersion Include="Microsoft.EntityFrameworkCore.Sqlite" Version="7.0.12" />
+    <PackageVersion Include="Microsoft.EntityFrameworkCore.Tools" Version="7.0.12" />
     <PackageVersion Include="Microsoft.Extensions.Caching.Abstractions" Version="7.0.0" />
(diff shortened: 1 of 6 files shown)


cvium:
I am not sure why this is needed. We are already trying to find the bind address when starting the application.

PrplHaz4:
Shouldn't PublishedServerUrl be in the mix here too?  In a docker container without host networking or only accessible via reverse proxy, the LAN IP will not likely be reachable by any (DLNA or Chromecast) clients.

Sky-High:
@Shadowghost: The new constants and extensions look very nice, in particular when working with IPNetwork addresses and bindings. I would happily use more of that functionality but I can not really see how to make the InitialiseServerLanAddress code in NetworkManager.cs more efficient. For instance, GetBindAddress seems quite useful in retrieving a valid non-loopback bind address, but not all valid bind addresses are valid server address for the chromecast receiver.

The used approach already builds upon available bind addresses in _interfaces as that is a good start. These are filtered and ranked in preference in the following order:

1. All IPv6 global addresses are valid and unique everywhere, so do satisfy the server address requirement for chromecast receiver. And indeed, I tested that these addresses do work. Global addresses however do not seem to be the best choice for a local network, so in preference they rank lowest.

2. Preferable are local addresses. For IPv4 these are the common four digit private network addresses, and for IPv6 the Unique Local Addresses (ULA). Often more than one address is available, but not all are usable for the cast receiver.

- the local addresses must be advertised on the local network. In case of multiple real and virtual adapters, in the IPv4 case the most favorable is an adapter that also serves as gateway as there is a good chance that this adapter is also known to a router and thus is advertised, so the cast receiver will always be able to find the jellyfin server via this adapter address.

- Jellyfin server may have valid reasons to bind to Virtual Networks as applications in a virtual machine might want to communicate with Jellyfin. But a chromecast receiver is a 'real' physical device so will not be on a virtual network so virtual adapter addresses can not be used for this.

- In the IPv6 case local addresses can be ULA's or linklocal. Linklocal addresses do not seem to work with chromecast receivers. I don't know why, maybe because these addresses are very specific for the concerning adapters, on both the server and cast receiver end. The IPv6 UniqueLocal and SiteLocal addresses are valid however.

3. In case both IPv4 and IPv6 addresses are available, IPv6 is ranked higher in preference as it is the most state-of-the-art.

I am not a network engineer, so may be this reasoning has its flaws, but it seems to work.

Sky-High:
@PrplHaz4: The issue addressed in this PR originates from a single system running both Jellyfin-server and Jellyfin-web, and connecting via localhost. This seems to differ from your configuration, although of course comparable issues may play.

Sky-High:
@Shadowghost : I would be happy to use more of the recent constants for loopback addresses and the Network extensions, but as far as I can see that does not really seems to apply for this PR. Maybe you could give me some clues?
It would be nice to move forward with this PR.

Shadowghost:
I'll look at it over the weekend and write up where it might be applicable. It was just an idea to use them :)

Sky-High:
@Shadowghost I appreciate your great involvement, so if there is anything I can do to help ...

Shadowghost:
Sorry, started working full-time and free time is of an issue right now. But I will come back to this when I have time in the next two weeks (need to fix some other stuff in NetworkManager too anyway).

PR #10114, opened 2023-08-15
Reduce the amount of allocations in GetWhereClauses

Bond-009:
Reuse 1 StringBuilder to generate multiple clauses instead of creating Lists of strings to join

### DIFF
--- a/Emby.Server.Implementations/Data/SqliteItemRepository.cs
+++ b/Emby.Server.Implementations/Data/SqliteItemRepository.cs
@@ -26,7 +26,6 @@ using MediaBrowser.Controller.Entities.Audio;
 using MediaBrowser.Controller.Entities.Movies;
 using MediaBrowser.Controller.Entities.TV;
 using MediaBrowser.Controller.Extensions;
-using MediaBrowser.Controller.Library;
 using MediaBrowser.Controller.LiveTv;
 using MediaBrowser.Controller.Persistence;
 using MediaBrowser.Controller.Playlists;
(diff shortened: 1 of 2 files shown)


Emby.Server.Implementations/Data/SqliteItemRepository.cs
```diff
@@ -3819,266 +3757,270 @@ private List<string> GetWhereClauses(InternalItemsQuery query, IStatement? state
...
                 {
-                    for (var ind = 0; ind < query.BlockUnratedItems.Length; ind++)
-                    {
-                        statement.TryBind(paramName + ind, query.BlockUnratedItems[ind].ToString());
-                    }
+                    clauseBuilder.Append(ParamName).Append(i).Append(',');
+                    statement?.TryBind(ParamName + i, query.BlockUnratedItems[i].ToString());
                 }
 
+                // Remove last,
```
crobibero, line 3961:
```suggestion
                // Remove last ","
```

I initially thought this was an unfinished comment

Emby.Server.Implementations/Data/SqliteItemRepository.cs
```diff
@@ -3819,266 +3757,270 @@ private List<string> GetWhereClauses(InternalItemsQuery query, IStatement? state
...
+                clauseBuilder.Append("InheritedParentalRatingValue is null");
             }
 
-            var ratingClauseString = ratingClauseBuilder.ToString();
-            if (!string.Equals(ratingClauseString, "(", StringComparison.OrdinalIgnoreCase))
+            if (clauseBuilder.Length > 1)
             {
-                whereClauses.Add(ratingClauseString + ")");
+                whereClauses.Add(clauseBuilder.Append(')').ToString());
+                clauseBuilder.Length = 0;
```
crobibero, line 4023: Shouldn't we clear the clause builder even if length is 0?

cvium: Shouldn't be necessary afaik

Bond-009: All StringBuilder.Clear() does is setting length to 0

https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Text/StringBuilder.cs,12bafac7fd9481fd

cvium: I would argue the intent with  Clear is... Clearer

crobibero: I meant to say if the length is 1.. my mistake!

Bond-009: @cvium I used .length = 0 instead of .clear bc it's done this way in the existing code, but I could change all occurrences if that's preferred.

Emby.Server.Implementations/Data/SqliteItemRepository.cs
```diff
@@ -1305,88 +1304,27 @@ private bool TypeRequiresDeserialization(Type type)
...
-                return false;
-            }
-
-            if (type == typeof(MusicAlbum))
-            {
-                return false;
-            }
-
-            return true;
+            return !(type == typeof(Season)
```
cvium, line 1314: Why the negation and not just =! &&?

Emby.Server.Implementations/Data/SqliteItemRepository.cs
cvium: @Bond-009 Doesn't really matter to me

cvium (approved):

crobibero (approved):

PR #10125, opened 2023-08-18
fix: accessing Standard* of a Process requires manually disposing them afterwards

cvium:
**Changes**
Seems like we must dispose the streamreaders manually to avoid (temporarily) leaking SafeFileHandles: https://source.dot.net/#System.Diagnostics.Process/System/Diagnostics/Process.cs,894

### DIFF
--- a/MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
+++ b/MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
@@ -1033,6 +1033,14 @@ namespace MediaBrowser.MediaEncoding.Encoder
 
                 try
                 {
+                    process.StandardOutput.Dispose();
+                }
+                catch
+                {
+                }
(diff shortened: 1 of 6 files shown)


Bond-009 (approved):

MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
```diff
@@ -1031,6 +1031,14 @@ private void DisposeProcess(Process process)
                     _mediaEncoder._runningProcesses.Remove(this);
                 }
 
+                try
+                {
+                    process.StandardOutput.Dispose();
```
cvium, line 1036: There's some timing issues. This should be put where it's being read

Shadowghost (approved):

Bond-009 (approved):

PR #10138, opened 2023-08-21
refactor: use Microsoft.Data.Sqlite instead of SQLitePCL.pretty

cvium:
**Changes**
Essentially did a lift and shift, so the extension methods aren't the greatest. Microsoft.Data.Sqlite still uses SQLitePCL.Raw under the hood, but it's well-maintained since it's used by EFCore.


Emby.Server.Implementations/Data/SqliteExtensions.cs
```diff
@@ -1,11 +1,12 @@
-#nullable disable
+#nullable enable
```
Bond-009, line 1:
```suggestio
```

Emby.Server.Implementations/Data/SqliteExtensions.cs
```diff
@@ -115,335 +114,144 @@ public static bool TryReadDateTime(this IReadOnlyList<ResultSetValue> reader, in
...
-            if (value.HasValue)
-            {
-                TryBind(statement, name, value.Value);
-            }
-            else
-            {
-                TryBindNull(statement, name);
-            }
+            var result = command.ExecuteScalar();
+            return Convert.ToInt32(result!, CultureInfo.InvariantCulture);
```
Bond-009, line 247: Add a comment why this can't be null

Jellyfin.Server/Migrations/Routines/RemoveDuplicateExtras.cs
```diff
@@ -37,14 +38,12 @@ public void Perform()
...
-                dbPath,
-                ConnectionFlags.ReadWrite,
-                null))
+            using (var connection = new SqliteConnection($"Filename={dbPath}"))
             {
                 // Query the database for the ids of duplicate extras
                 var queryResult = connection.Query("SELECT t1.Path FROM TypedBaseItems AS t1, TypedBaseItems AS t2 WHERE t1.Path=t2.Path AND t1.Type!=t2.Type AND t1.Type='MediaBrowser.Controller.Entities.Video'");
-                var bads = string.Join(", ", queryResult.SelectScalarString());
+                // TODO does this LINQ execute before the reader is disposed?
+                var bads = string.Join(", ", queryResult.Select(x => x.GetString(0)).ToList());
```
Bond-009, line 46: Shouldn't even need the tolist here, no?

Bond-009 (changes requested):

Emby.Server.Implementations/Data/SqliteExtensions.cs
cvium: ?

Emby.Server.Implementations/Data/SqliteExtensions.cs
cvium: done

Jellyfin.Server/Migrations/Routines/RemoveDuplicateExtras.cs
cvium: No

Emby.Server.Implementations/Data/SqliteExtensions.cs
Bond-009: Nullable is enabled by default, you can remove the first line

MediaBrowser.Providers/TV/SeriesMetadataService.cs
```diff
@@ -228,13 +227,12 @@ private async Task UpdateAndCreateSeasonsAsync(Series series, CancellationToken
                 }
                 else
                 {
-                    var name = GetValidSeasonNameForSeries(series, seasonName, seasonNumber);
-                    if (string.Equals(existingSeason.Name, name, StringComparison.Ordinal))
+                    if (string.Equals(existingSeason.Name, seasonName, StringComparison.Ordinal))
```
Bond-009, line 230: Now you can just make this an else if

cvium: done

Emby.Server.Implementations/Data/SqliteItemRepository.cs
```diff
@@ -638,7 +624,7 @@ private void SaveItemsInTransaction(IDatabaseConnection db, IEnumerable<(BaseIte
                 {
                     if (requiresReset)
                     {
-                        saveItemStatement.Reset();
+                        // TODO saveItemStatement.Parameters.Clear();
```
crobibero, line 627: Does this TODO have a known fix?

crobibero (commented):
I may have missed it- but we don't seem to be using connection pooling? Is that something we want to enable?

Jellyfin.Server/Jellyfin.Server.csproj
```diff
@@ -47,8 +47,6 @@
     <PackageReference Include="Serilog.Sinks.Async" />
     <PackageReference Include="Serilog.Sinks.Console" />
     <PackageReference Include="Serilog.Sinks.File" />
-    <PackageReference Include="Serilog.Sinks.Graylog" />
```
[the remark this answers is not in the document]
cvium: Wouldn't surprise me if they could

Emby.Server.Implementations/Data/SqliteItemRepository.cs
cvium: It's not really necessary anymore, but I've added a clear as a safeguard

cvium:
> I may have missed it- but we don't seem to be using connection pooling? Is that something we want to enable?

It's enabled by default afaik?

crobibero:
> > I may have missed it- but we don't seem to be using connection pooling? Is that something we want to enable?
> 
> 
> 
> It's enabled by default afaik?

That's what it looked like, I think I just don't understand the library :)

crobibero (approved):

Bond-009 (approved):

Shadowghost (approved):

PR #10153, opened 2023-08-26
Add MediaStreamProtocol enum

nielsvanvelzen:
Was intended for the TranscodingProfile but it's used in some other API parts as well. We only support HLS (and HTTP is the "fallback"/default).

Ideally we'd use the term "progressive" instead of "http" but I don't want to break API compatibility, which from my understanding does not happen right now as the supported names are the same (http/hls).

**Changes**
- Add MediaStreamProtocol enum

**Issues**


nielsvanvelzen:
The tests are failing because the test data is using empty strings which are not a part of the enum. Empty strings should use the default (MediaStreamProtocol.Http) but I'm not sure how to accomplish that.

JPVenson (approved):

Shadowghost (approved):

nielsvanvelzen:
Can't approve because I made the initial changes but ✅ API changes. [Kotlin SDK diff](https://github.com/nielsvanvelzen/jellyfin-sdk-kotlin/commit/dcf57e991100df706af8bc2f5f7333ef404ceffd)

Bond-009 (approved):

PR #10154, opened 2023-08-26
Fix a few multiple enumerations

goremykin:
This PR suppresses one multiple enumeration warning and fixes one real multiple enumeration

**Changes**
See comments below

### DIFF
--- a/Emby.Server.Implementations/Library/LibraryManager.cs
+++ b/Emby.Server.Implementations/Library/LibraryManager.cs
@@ -358,7 +358,7 @@ namespace Emby.Server.Implementations.Library
 
             var children = item.IsFolder
                 ? ((Folder)item).GetRecursiveChildren(false)
-                : Enumerable.Empty<BaseItem>();
+                : Array.Empty<BaseItem>();
 
             foreach (var metadataPath in GetMetadataPaths(item, children))
             {
--- a/MediaBrowser.Providers/Manager/ProviderManager.cs
+++ b/MediaBrowser.Providers/Manager/ProviderManager.cs
@@ -807,13 +807,11 @@ namespace MediaBrowser.Providers.Manager
             where TLookupType : ItemLookupInfo
         {
             var results = await provider.GetSearchResults(searchInfo, cancellationToken).ConfigureAwait(false);
-
-            foreach (var item in results)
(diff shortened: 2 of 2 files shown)


Emby.Server.Implementations/Library/LibraryManager.cs
```diff
@@ -358,7 +358,7 @@ public void DeleteItem(BaseItem item, DeleteOptions options, BaseItem parent, bo
 
             var children = item.IsFolder
                 ? ((Folder)item).GetRecursiveChildren(false)
-                : Enumerable.Empty<BaseItem>();
+                : Array.Empty<BaseItem>();
```
goremykin, line 361: Here I just suppress false multiple enumeration warning. So now `children` is an `IList`

MediaBrowser.Providers/Manager/ProviderManager.cs
```diff
@@ -807,13 +807,11 @@ private bool IsSaverEnabledForItem(IMetadataSaver saver, BaseItem item, LibraryO
             where TLookupType : ItemLookupInfo
         {
             var results = await provider.GetSearchResults(searchInfo, cancellationToken).ConfigureAwait(false);
-
-            foreach (var item in results)
```
goremykin, line 811: Here we have a bug. If `provider.GetSearchResults` implementation returns lazy enumeration (like `MusicBrainzArtistProvider` for example), we lose `SearchProviderName`

cvium: Why not remove the method? It's a private method used in only one place.

cvium (approved):
LGTM

Bond-009 (approved):
Weird, multiple enumerations should cause errors when building debug mode :thinking:
What warnings did you get CA1851 or something else?

goremykin:
@Bond-009, I received these warnings from the jetbrains rider's analyzer.
There is even one more [here](https://github.com/jellyfin/jellyfin/blob/1f083c0de2a1be669c2af7e77cd4e7313ee8f449/RSSDP/IEnumerableExtensions.cs#L21). But I haven't come up with a good solution yet.

crobibero (approved):

Issue #8114, opened 2022-07-12
[Issue]: Jellyfin 10.8.1 not working with SELinux on Rocky Linux 8.6


bddali:
### Please describe your bug

Upgraded from Jellyfin 10.7.7-1 to 10.8.1-1.

After upgrade, the service didn't start. Errors related to SELinux were logged:

Jul 11 11:38:12 servername jellyfin[595608]: Failed to create CoreCLR, HRESULT: 0x80004005
Jul 11 11:38:12 servername systemd[1]: jellyfin.service: Main process exited, code=exited, status=137/n/a
Jul 11 11:38:12 servername systemd[1]: jellyfin.service: Failed with result 'exit-code'.
[...]
Jul 11 11:38:14 servername setroubleshoot: SELinux is preventing /usr/lib64/jellyfin/jellyfin from using the getsession access on a process.
Jul 11 11:38:14 servername setroubleshoot: SELinux is preventing /usr/lib64/jellyfin/jellyfin from using the execmem access on a process.

Downgraded to 10.8.0-1, SELinux issues went away and the service started successfully.

### Jellyfin Version

Other

### if other:

10.8.1-1

### Environment

```markdown
- OS: Rocky Linux 8.6
- Virtualization: No
- Clients: Browser
- Browser: Chrome, Firefox
- FFmpeg Version: ffmpeg-4.2.7-1
- Playback Method: Transcode
- Hardware Acceleration: None
- Plugins: None
- Reverse Proxy: None
- Base URL: None
- Networking: Host
- Storage: local
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PrivacyOsint2:
Samething: https://github.com/jellyfin/jellyfin/issues/8101

bddali:
@PrivacyOsint2 I don't see any similarity between the issues.

oneironaut2209:
I had same issue on Fedora 36. After setting SELinux to permissive the service starts normally. Also, running jellyfin directly  (/usr/bin/jellyfin) also works.

akolomitz:
> I had same issue on Fedora 36. After setting SELinux to permissive the service starts normally. Also, running jellyfin directly (/usr/bin/jellyfin) also works.

[...]

PR #10169, opened 2023-08-29
Add a small Bash script to launch Jellyfin, instead of a symlink.

gandihar:
Fix the Jellyfin launcher on SELinux-enabled systems. Replace /usr/bin/jellyfin symlink with a 2-line shell script, which itself calls the target of the symlink.
  - The symlink causes a problem with SELinux because it understands symlinks.
  - This shell script automatically inherits the correct SELinux context.

**Changes**
- Add 2-line shell script that calls the Jellyfin executable
- Update Fedora spec file to use new launcher script.

**Issues**
Fixes [Issue 8114](https://github.com/jellyfin/jellyfin/issues/8114)

### DIFF
--- a/fedora/jellyfin-selinux-launcher.sh
+++ b/fedora/jellyfin-selinux-launcher.sh
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+exec /usr/lib64/jellyfin/jellyfin ${@}
--- a/fedora/jellyfin.spec
+++ b/fedora/jellyfin.spec
@@ -14,6 +14,7 @@ License:        GPLv2
 URL:            https://jellyfin.org
 # Jellyfin Server tarball created by `make -f .copr/Makefile srpm`, real URL ends with `v%%{version}.tar.gz`
 Source0:        jellyfin-server-%{version}.tar.gz
+Source10:       jellyfin-selinux-launcher.sh
 Source11:       jellyfin.service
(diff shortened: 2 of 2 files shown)


cvium (approved):
I know nothing about Fedora and I can't test it, but LGTM

Shadowghost (approved):

lxwulf:
> I know nothing about Fedora and I can't test it, but LGTM

I do know Fedora or better say I use it since several years, maybe I can help to test it.

Bond-009 (approved):

fedora/jellyfin-selinux-launcher.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+exec /usr/lib64/jellyfin/jellyfin ${@}
```
brianjmurrell, line 3: You should [quote the `$@` to prevent expansion problems](https://www.shellcheck.net/wiki/SC2068) with arguments that might have spaces in them.

```suggestion
exec /usr/lib64/jellyfin/jellyfin "${@}"
```

brianjmurrell, line 3: This relies on the assumption that `%{_libdir}` is always going to be `/usr/lib64`.  A more portable and future-proof method would be to substitute `%{_libdir}` into this script before installing it.

PR #10186, opened 2023-09-02
feat: Return early if client requests segments just for buffering

ringus1:
As described in mentioned feature request, it would be great if Jellyfin would make use of multiple connections for streaming. Server basically supports this out-of-the-box and it's mostly up for the clients to make use of it. Yet, for transcoding scenario, client doesn't know how far it can do the actual buffering, without worrying about order in which those segments are pulled (like with 20 connections for example), before server tries to restart the transcoding from new timestamp.

Idea here is to make use of some header, which I propose here to be **X-Buffer-Only** to indicate that client requests given segment just for the sake of buffering and it's not intended as **seek** operation by user for that matter. So, if given segment is just not yet ready, it will return **412** status code, as precondition is not yet met. That will indicate to simply try again in a while.

Everything works as before for clients that will not use this flag.

**Changes**
Added logic to make use of optional **X-Buffer-Only** requested by the client, to not treat requested segments ahead of play-time as **seek** operation by the client. Instead it will return **412** status code if segment is not ready yet. Transcoding throttle will be unpaused if client is eager for more buffer.

**Issues**
Feature [2171](https://features.jellyfin.org/posts/2171/better-internet-streaming-using-multiple-parallel-connections-or-threads)

### DIFF
--- a/Jellyfin.Api/Controllers/DynamicHlsController.cs
+++ b/Jellyfin.Api/Controllers/DynamicHlsController.cs
@@ -1458,6 +1458,10 @@ public class DynamicHlsController : BaseJellyfinApiController
             _logger.LogDebug("returning {0} [it exists, try 1]", segmentPath);
             return await GetSegmentResult(state, playlistPath, segmentPath, segmentExtension, segmentId, job, cancellationToken).ConfigureAwait(false);
         }
+        else if (HttpContext.Request.Headers.ContainsKey("X-Buffer-Only"))
+        {
+            return new StatusCodeResult(412);
+        }
 
         var transcodingLock = _transcodingJobHelper.GetTranscodingLock(playlistPath);
         await transcodingLock.WaitAsync(cancellationToken).ConfigureAwait(false);
--- a/Jellyfin.Api/Controllers/DynamicHlsController.cs
+++ b/Jellyfin.Api/Controllers/DynamicHlsController.cs
@@ -1458,15 +1458,12 @@ public class DynamicHlsController : BaseJellyfinApiController
             _logger.LogDebug("returning {0} [it exists, try 1]", segmentPath);
             return await GetSegmentResult(state, playlistPath, segmentPath, segmentExtension, segmentId, job, cancellationToken).ConfigureAwait(false);
         }
-        else if (HttpContext.Request.Headers.ContainsKey("X-Buffer-Only"))
-        {
-            return new StatusCodeResult(412);
-        }
 
         var transcodingLock = _transcodingJobHelper.GetTranscodingLock(playlistPath);
         await transcodingLock.WaitAsync(cancellationToken).ConfigureAwait(false);
         var released = false;
         var startTranscoding = false;
+        bool bufferOnly = HttpContext.Request.Headers.ContainsKey("X-Buffer-Only");
 
         try
         {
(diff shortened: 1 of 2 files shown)


Shadowghost:
Just an FYI: using the X prefix for your own headers is discouraged by the IETF: https://stackoverflow.com/questions/3561381/custom-http-headers-naming-conventions

I suppose you tested that this actually works?

ringus1:
Interesting, I stumbled upon some materials that it is recommended. Perhaps **was** indeed. Though used widely by APIs like Amazon or Akamai IIRC.

Anyway, yes I've tested it. Not sure though what are standards for HLS clients. Perhaps `404` is more valid status code for segments that are yet to be populated?

JPVenson:
While i agree with you ghost, its still practically industry standard. Even JF uses it RN:
```
X-Content-Type-Options: nosniff
X-Response-Time-Ms: 0
X-Xss-Protection: 1;mode=block
```
And i think its fine to annotate custom headers.

JPVenson:
To further improve responsiveness, would it be better to block the request until the requested chunk is available?

ringus1:
I don't believe blocking would be better. It's _easier_ for naive clients in some aspects, true. But since it would need to pass this header explicitly to avoid **seeking** anyway, it needs to be conscious implementation.
Several issues that would come with that approach:
- It may be quite a while before given segment is ready, if client is too greedy (timeouts - client would need to interpret them somehow, and could be hard to tell if some are intentional, or not really)
- Async implementation or not, it blocks server resources for no good reason. Client can just execute backoff and try in few moments on expected status codes.
- Gets tricky also on client side, if user decides to actually do the **seek** and discard current buffering - connections would have to be aborted etc etc

nielsvanvelzen:
Can you link to the specification that defines the `X-Buffer-Only` header? I couldn't find anything about it.

JPVenson:
> Can you link to the specification that defines the `X-Buffer-Only` header? I couldn't find anything about it.

Its a custom header he created

nielsvanvelzen:
> > Can you link to the specification that defines the `X-Buffer-Only` header? I couldn't find anything about it.
> 
> Its a custom header he created

Didn't realize this 😅. In that case why is this not a query parameter or separate endpoint? Why use a header??

ringus1:
Yes, it's my invention. I don't think you will find anything like this for **closed** playlists. They simply indicate that content is ready, just fragmented. That's the issue with transcoded streams. _Readiness_ is basically faked.

ringus1:
> Didn't realize this 😅. In that case why is this not a query parameter or separate endpoint? Why use a header?

Could be. Figured header is minimum hassle to indicate intent. I'm mostly interested to hear, whether you guys think it's worth adding this way.

ringus1:
Especially that some major library like this: https://github.com/video-dev/hls.js/issues/5284 doesn't even seem to be considering to support something similar. :/

JPVenson:
> > Didn't realize this 😅. In that case why is this not a query parameter or separate endpoint? Why use a header?
> 
> Could be. Figured header is minimum hassle to indicate intent. I'm mostly interested to hear, whether you guys think it's worth adding this way at all.

If backwards compatibility is given, i am fine with the header if its the easy route for you.

JPVenson:
I want to test this, later this week. Otherwise OK for me

JPVenson:
i did not forget this. i just blown my dev env. need some time.

Bond-009:
Closing as not planned

PR #10196, opened 2023-09-05
Allow specifying an offset for a subtitle file

theshoeshiner:
**Changes**

This is a change to the subtitle file upload / retrieval logic to look for a token in the subtitle file name that tells the server to offset subtitles from that file a specified number of milliseconds. This is helpful when you have a subtitle file that has accurate text but is slightly out of sync. Because shifting subtitle files manually is difficult, and the use of online tools is still time consuming, it would be more desirable to have the server handle the slight discrepancy. When the offset token is found, the offset will be applied when the subtitles are sent to the client, thus no changes are necessary on the various clients. This largely removes the need for client-specific subtitle sync features, like that present in the web client.

I modelled this enhancement after the "Hearing Impaired" flag logic (and other similar features), which uses a token in the subtitle file name. For this logic the subtitle file needs to contain a token like "offset[#]" where [#] is an integer (negative values are valid) representing the number of milliseconds to offset the subtitles. e.g.

```
mymovie.eng.offset3200.srt
mymovie.spa.offset-2400.srt
```

In addition to applying the offset, the display name of the subtitle contains an attribute e.g.

```
English - Offset 3.2s - SUBRIP - External
Spanish - Offset -2.4s - SUBRIP - External
```

This change requires an additional nullable int database column to store the offset value. Ideally (in a separate PR) we would be able to change this offset value via the web UI rather than manually changing file names.

Some other thoughts/notes

- Currently when subtitle streams are retrieved, if the output format is the same as the source format then the server bypasses the `ConvertSubtitles` method where the shift takes place. With this new logic, if an offset is present, then the code must _always_ run the "convert" logic, even if the output format is the same as the source. An alternative to this would be to rewrite the source file, but this would pose other problems.
- I'm calling this property "Offset", however "Shift" also makes sense, but "Offset" was already used in the web client for this type of logic so I kept with that precedent.

**Issues**


EraYaN:
I feel like this might the wrong direction to take this feature. Most clients let you set an offset already, I'd be more in favor of supporting a way of saving that offset through the API, instead of this.

theshoeshiner:
I definitely agree. This PR is mostly to create the proper api to store the offset. The parsing of the offset from the file name is an initial / backup solution for users that need to edit things en masse and/or if the proposed changes to the admin app and web client cant be agreeed upon.

My ideal solution would be to update the server code to handle and persist the offset (what this PR does) and update the web client to 1) allow the user to edit the offset from the Edit Subtitles dialog and 2) have the clients integrate with the new API such that the user isn't potentially adjusting it in two places. Client adjustments would then ideally just be UI sugar since the real work is handled by the server.

theshoeshiner:
One thing to discuss - Right now as far as I can tell, subtitles are saved as media streams, and all their properties are collected at the time of upload/discovery and saved into that table. That table seems to mostly be intended as a place to store immutable properties of files i.e. things inherent to the data file that should not change. If that is a hard restriction, then allowing users to update the data that gets stored in that table may not be a good option.

If we do want to update that table directly then I wonder if we should also rename the file? As everything in that table seems to be information that can be derived from the file name and/or contents (e.g. language, format, etc). Setting the language of subtitles causes the file name to change, so it makes sense that setting the offset might do that as well?

EraYaN:
I feel like the file name is one of the worst places to have it, many people mount their media read-only. It's should be in the database somewhere for sure. Also a file name change means all backups tools and such will pick it up again.

theshoeshiner:
`I feel like the file name is one of the worst places to have it`

I agree in general, and would have gone that direction if this were brand new software, but so much of the existing code seems to be structured around using the file system / file names to hold metadata that I felt proposing such a significant change of direction wouldn't be appropriate for a first time contributor.

`It's should be in the database somewhere for sure`

It's in the mediastreams table, its just a question of how to keep it from getting overwritten.

EraYaN:
Really the only thing it should grab from the file name is potentially the language for subtitles and sidecar audio and for main media files stuff to use for looking up the metadata. I don't think we parse much from the file name for client settings, or at least we very much shouldn't.

I'd be open to having a new table for media streams that stores this, although putting in an offset is fine as well if the scanner gets updated to use upserts or merges correctly and leave existing data alone.

theshoeshiner:
As of right now there appear to be ~6 things being parsed from the file name:

1-2. The media file name and subtitle format (the obvious two)
3. The subtitle language
4. The subtitle forced flag
5. The default flag
6. The subtitle "hearing impaired" flag

Although I agree none of that is ideal for storing there, I figured consistently bad is better than inconsistently good.

And I don't think of this data as a client setting. It's not specific to any client. The offset data is specifically indicating to the server that "this subtitle file doesn't match unless shown with this offset", and _the server_ is responsible for applying the offset before sending anything to the client. The client doesn't even need to know about this setting.

The only reason a client might interact with this setting is to allow the user to adjust it on the fly, but that would no longer be a necessary feature to make the offset work, just an added bonus that it can be tweaked/adjusted from the player, as opposed to backing out, making the adjustment, and restarting.

JPVenson:
I like the idea to allow setting offset on the file, as i already had the issue that my external subtitle was out-of-sync for a series and i had to script a delay myself because the andriod tv client acted up and i had to set it every time when i watched that series. Storing the offset in the DB would be the right way though, but then you get the issue that your DB value can be out-of-sync with what the file provides. So the file would need to take prevalence over what is in the DB which makes the whole DB value worthless until we can edit it in the Web-UI.

cvium:
The parser is already pretty bad and working overtime. Let's not add to it.

theshoeshiner:
@JPVenson 

I've actually already updated the webapp with the ability to set the offset from the edit screen. Just haven't created the PR yet but I have it built and running on my server. I applied these changes to 10.8.10 so that I could use them immediately without worrying about 10.9 changes.

Obviously the webapp change will be useless unless the server changes as well. And it seems like there's still debate on how to handle that. I don't really have any good ideas about how to handle the offset outside the file name, because there is no DB table to store attributes like that. In my current implementation the file value takes precedence because the mediastreams data gets overwritten based on the file 

I haven't started making changes to the web player offset slider yet. Ideally that would sync with this property.

gaurovgiri (approved):
Seems good to me!

JPVenson (commented):
i actually did this review months ago but forgot to submit it sry :-D

warren-woolsey:
Has there been any action on getting the merge conflicts resolved, or has an alternative solution provided?

theshoeshiner:
I took a look the other day and I was torn between continuing with this and developing a more robust integration with one of the various external subtitle sync programs, which is what I'm now in need of. But I think this feature is still valuable for other users, given that most of them probably use formal releases of media (whereas I have a ton of DVR stuff) so if there is still a desire to merge it then I'll take a look at the conflicts this week.

zotabee:
I don't see the logic here. This isn't the role of Jellyfin Server to handle such thing in my opinion. This feels like implementing a fix in Jellyfin Server to fix something that shouldn't be broken in the first place. Fix your media files (your subtitles) not Jellyfin.

And I speak as someone who synced couple thousands of subtitles.

Either you don't want to do the syncing job beforehand or don't know how much to delay and don't care to have it saved it properly then, use a good video player / Jellyfin client to apply a manual offset (all good players offer buttons or shortcuts to apply a subtitles offset / delay).

Either if you want to do a proper job beforehand, just sync your subtitles. You should use Subtitle Edit : https://github.com/SubtitleEdit/subtitleedit

If you already know the offset, you can apply an offset / delay to the whole subs in two clicks, : Open your subtitles then go under "Synchronization" > "Adjust all times (show earlier/later)...", enter your offset, choose "Show earlier" or "Show later", save your subs, done. It takes less than 10 sec, faster than renaming your filenames...

If you don't know the offset, you can do a "Synchronization" > "Point Sync via another subtitle..." if you have other subs in sync found on internet. <1-2min.

Or do a visual sync with the audio waveform in extreme cases where you can't find corresponding subs.

And Subtitle Edit even has a "Batch convert" feature in "Tools" if you needs to do it repetitively (like for a whole season or a whole TV series, directly, if it's a constant/fixed offset/delay).

It even has a CLI version, that can probably be used with some scripting if you already have the offset stored in text somewhere.

theshoeshiner:
Most of what jellyfin does internally can be accomplished manually with other tools, including encoding, so the argument that it shouldn't do it simply because something else exists doesn't really hold water to me.

If the jellyfin client can do it, then we've already crossed the bridge of acknowledging that it's a useful feature and is perfectly in line with jellyfins "job". This is nothing more than an enhancement of that feature that prevents users from having to do that manual synchronization *in every single client*.

zotabee:
Possibly but using filenames to add an offset is like a custom personal use, at best. This feels like a PR for personal use.

And once you sync your subs, this will work with every players and every media servers, not only in Jellyfin. So there are only upsides in doing it properly before and not in Jellyfin.

theshoeshiner:
See the prior convo. The filename is already used to indicate 5+ pieces of metadata. That bridge has been crossed as well. This simply follows an already established JF pattern.

If your argument is that the proper way (sync program) is the only way, then why bring up the client feature? Its obvious that people want/use a feature that syncs subtitles  without having to run an external process.

This is not a complete replacement for proper syncing of an entire library. It is a convenience feature for people that dont have a large enough problem to be worthy of downloading and learning an independent program.

theshoeshiner:
FWIW my stance has always been that these criticisms are valid, but they are criticisms of *jellyfin*, not of this PR. This PR is in line with principles and goals already established in core.

If there is an existing effort to change or redesign the related features in the core and client, then this PR can and should be evaluated against those. But barring that, I dont feel this PR breaks any established patterns or goals, nor does it introduce any new ones.

zotabee:
Regarding your second-to-last post. It's not because Jellyfin is using filenames for other information that your idea is good.

And again, this doesn't make sense at all. If you already know the offset/delay, then sync your subs outside of Jellyfin instead of doing some custom filename renaming (that probably nobody is doing except you) and patching Jellyfin.

See my previous post, I even showed you how to do it with Subtitle Edit, which is the gold standard, with a modern and very intuitive GUI, and FOSS on GitHub. You won't even have to Google it to learn how to do it. It is as fast as renaming your filenames. Your initial assumption that syncing subs is hard and long is plain wrong. Again, if you already know the offset/delay, it literally takes under 10 secs max and couple clicks and it's as super easy. No excuse to not doing it, really. And you can even do it in batch.

You are trying to create a new naming convention, out of nowhere, that you are probably the only one using because you are just reticent and too lazy to learn how to sync your subs which should have been how you spent your time initially. This would have fixed your problem at the source directly instead of establishing a custom hack like this.

And you are confusing different scenarios, client-side subs delay/offset is usually used when you don't know the offset/delay and want to adjust on the fly (or to correct delay for subs getting slowly out-of-sync during playback because of speed/length not matching). So it's a must have feature for every good players.

But here you are describing a scenario where you already know the delay precisely, in advance and with the correct speed. Which is really odd and specific.

The problem is your media library and your out-of-sync subs, not Jellyfin lacking a "feature". So fix your subs and not Jellyfin.

theshoeshiner:
You're still missing some of the prior context.

It's not about forcing users to rename files. There would be follow on work, previously mentioned, that allowed users to quickly update the offset via the UI, exactly the same way they configure other properties *which also get persisted into the file name*. If you understood the code then you'd understand why the file name logic is necessary. Im not inventing a naming convention, because Jellyfin *has clearly already done that*. **Read**. **The**. **Code**. If you're not capable of doing that then you should move on.

There's no need for insults. You come here having obviously read none of the history, understanding none of the current implementation, and thus contributing _absolutely nothing_ to the conversation other than your own floating goalposts, and you have the nerve to get rude when your misunderstandings are called out. That doesn't have a place here. At this point, I could not care less about your thoughts on this matter. Feel free to continue to argue into the void, though.

Bond-009:
Closing as stale, review from 2023 hasn't been addressed.

PR #10197, opened 2023-09-05
Remove one session per device id limitation

thornbill:
**Changes**
Removes the limitation of only allowing one active auth token per device id. This removes the need for clients to implement hacks to enable users to switch logins without entering a password every time.

**Issues**
It's been mentioned in chat many times, but I don't see an issue or discussion for this :eyes:

### DIFF
--- a/Emby.Server.Implementations/Session/SessionManager.cs
+++ b/Emby.Server.Implementations/Session/SessionManager.cs
@@ -1519,24 +1519,9 @@ namespace Emby.Server.Implementations.Session
                     DeviceId = deviceId
                 }).ConfigureAwait(false)).Items;
 
-            foreach (var auth in allExistingForDevice)
-            {
-                if (existing is null || !string.Equals(auth.AccessToken, existing.AccessToken, StringComparison.Ordinal))
-                {
-                    try
-                    {
-                        await Logout(auth).ConfigureAwait(false);
-                    }
-                    catch (Exception ex)
-                    {
-                        _logger.LogError(ex, "Error while logging out.");
-                    }
-                }
-            }
-
             if (existing is not null)
             {
-                _logger.LogInformation("Reissuing access token: {Token}", existing.AccessToken);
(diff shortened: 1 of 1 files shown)


crobibero (commented):
Doesn't look like `allExistingForDevice` is used anymore so we can also remove that db call

Shadowghost (approved):

crobibero (approved):

cvium:
I am fairly certain this won't work. DeviceId is the primary key of the Device table. We currently don't support having multiple sessions for a device. There can only be one.

Emby.Server.Implementations/Session/SessionManager.cs
```diff
@@ -1513,30 +1513,9 @@ private async Task<string> GetAuthorizationToken(User user, string deviceId, str
                     Limit = 1
```
cvium, line 1513: What if the user somehow has multiple sessions on the same device? Shouldn't we remove the duplicates?

thornbill: How likely is that given the check that returns the same session when the same user authenticated on the same device?

cvium: It's probably very unlikely.

Another thing: If this flow is only triggered when the user attempts to log in, shouldn't we reset the token?

cvium: I think we should remove _all_ of the users sessions on that device and create a new one. I don't really understand why the token is just reused.

Shadowghost: Agreed

thornbill: Update to remove sessions for the same user and device. I also tested that I could use access tokens for different users without needing to re-auth (although web doesn't currently support this so I had to change local storage values directly).

PR #4628, opened 2020-12-01
Prevent GetUpNext from returning episodes in progress

MrTimscampi:
**Changes**

GetUpNext was returning episodes that are in progress, leading to this:

Screenshot_2020-12-01 Jellyfin

Its purpose should be to only return future episodes of a show that is already started, but not in progress episodes.

To achieve this, this gets the user data of the episode to be added to the list, checks if it's in progress and just returns null if it is.

**Issues**

PR #10200, opened 2023-09-06
Add option to include resumable items in next up requests

thornbill:
**Changes**
Adds an option to enable resumable items to be included in next up requests. This defaults to true to enable the original behavior of finding the next episode of a series that should be played.

Clients will probably want to add the new request parameter to prevent the duplication on the home screen between next up and continue watching. ~~Once this is merged I plan to add a new home screen section type that uses this to present a single row for next up + continue watching (as has been requested [here](https://features.jellyfin.org/posts/1055/give-the-option-to-combine-next-up-and-continue-watching)).~~ Actually this still won't fully meet the needs for a combined up next + continue watching home screen section since it does not include movies. :thinking: 

PR that introduced the initial breakage: https://github.com/jellyfin/jellyfin/pull/4628

**Issues**
N/A

### DIFF
--- a/Jellyfin.Api/Controllers/TvShowsController.cs
+++ b/Jellyfin.Api/Controllers/TvShowsController.cs
@@ -68,7 +68,8 @@ public class TvShowsController : BaseJellyfinApiController
     /// <param name="nextUpDateCutoff">Optional. Starting date of shows to show in Next Up section.</param>
     /// <param name="enableTotalRecordCount">Whether to enable the total records count. Defaults to true.</param>
     /// <param name="disableFirstEpisode">Whether to disable sending the first episode in a series as next up.</param>
-    /// <param name="enableRewatching">Whether to include watched episode in next up results.</param>
+    /// <param name="enableResumable">Whether to include resumable episodes in next up results.</param>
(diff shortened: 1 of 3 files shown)


Jellyfin.Api/Controllers/TvShowsController.cs
```diff
@@ -86,6 +87,7 @@ public class TvShowsController : BaseJellyfinApiController
         [FromQuery] DateTime? nextUpDateCutoff,
         [FromQuery] bool enableTotalRecordCount = true,
         [FromQuery] bool disableFirstEpisode = false,
+        [FromQuery] bool enableResumable = true,
```
crobibero, line 90: Should probably default this to false?

thornbill: I debated that but went with true since it effectively reverts the change that breaks resuming a show when an episode is in progress.

nielsvanvelzen: I believe this shouldn't break much. We should just keep our eyes out for any bug reports during the 10.9 beta cycle and if this causes issue change the default to `false`.

cvium (approved):

crobibero (approved):

PR #10218, opened 2023-09-13
Minor LibraryMonitor improvements

Bond-009:
* Enable nullable
* Add a fast return to ReportFileSystemChanged when path should be ignored
* Use Span overloads of Path.* functions where possible
* IFileSystem: remove NormalizePath as Path.TrimEndingDirectorySeparator already checks if it's a root path

### DIFF
--- a/Emby.Server.Implementations/IO/LibraryMonitor.cs
+++ b/Emby.Server.Implementations/IO/LibraryMonitor.cs
@@ -1,5 +1,3 @@
-#nullable disable
-
 #pragma warning disable CS1591
 
 using System;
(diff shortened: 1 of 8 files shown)


crobibero (approved):

Shadowghost (approved):

Emby.Server.Implementations/IO/LibraryMonitor.cs
```diff
@@ -417,7 +414,7 @@ private void CreateRefresher(string path)
                     }
 
                     // They are siblings. Rebase the refresher to the parent folder.
-                    if (string.Equals(parentPath, Path.GetDirectoryName(refresher.Path), StringComparison.Ordinal))
+                    if (parentPath is not null && string.Equals(parentPath, Path.GetDirectoryName(refresher.Path), StringComparison.Ordinal))
```
cvium, line 417: Shouldn't this use span as well?

cvium (approved):

Issue #10215, opened 2023-09-12
[Issue]: media directories served by ZFS have a ".zfs" directory, and with snapshots, duplicate the library contents, unable to exclude!


fermulator:
### Please describe your bug

Consider first, https://jellyfin.org/docs/general/server/media/excluding-directory/ ; OK we are able to exclude a directory by adding a `.ignore` file into it. I suppose that is "neat" (though slightly annoying to pollute the filesystem solely for a "hint" to Jellyfin).

Now consider this scenario;
* Admin has a ZFS storage filesystem on Linux zfsonlinux.org
* Admin uses ZFS snapshots https://docs.oracle.com/cd/E19253-01/819-5461/gbcya/index.html
* Admin remote mounts (or has local datasets)
* Admin is running jellyfin in docker (volume binding)
* Jellyfin imports media to these storage locations

The PROBLEM, is that the zfs dataset will contain a directory called `.zfs` (for zfs specific content) - and specifically in there, `.zfs/snapshot/*` contains DUPLICATIONS of any and all data within the dataset :( -- this means that jellyfin is trolling and finding duplicate media content for as many snapshots that may exist.

Cannot even workaround this problem by adding `.ignore` into there because the `.zfs` directory is READ-ONLY (by user or root).

Jellyfin should ideally provide a means to target `glob*` or `regex` exclusions - instead of relying on a `.ignore` file in the directory subsystem?

Open to alternative ways to fix/address.

---
ABOUT REFS

`[23:10:39] [INF] [1] Main: Jellyfin version: 10.8.10`

docker volume in docker-compose file:
```
      - /mnt/myserver/tv:/tv:ro
      - /mnt/myserver/movies:/movies:ro 
```

### Jellyfin Version

10.8.z

### Environment

```markdown
- OS: (N/A docker,, `jellyfin/jellyfin`)
- Linux Kernel:
- Virtualization:
- Clients: N/A
- Browser: N/A
- FFmpeg Version: `5.1.3`
- Playback Method:
- Hardware Acceleration: N/A
- GPU Model: N/A
- Plugins: N/A
- Reverse Proxy: N/A
- Base URL: N/A
- Networking: N/A
- Storage: `/movies` and `/tv`, via docker volume bind, to host `NFS` mount, backed by `zfs`
```

### Jellyfin logs

```shell
Interestingly, also jellyfin is throwing exceptions in logs :) 

[...]

PR #10219, opened 2023-09-13
Ignore .zfs folder

Bond-009:
Maybe helps with #10215 ?

### DIFF
--- a/Emby.Server.Implementations/Library/IgnorePatterns.cs
+++ b/Emby.Server.Implementations/Library/IgnorePatterns.cs
@@ -89,6 +89,10 @@ namespace Emby.Server.Implementations.Library
             // bts sync files
             "**/*.bts",
             "**/*.sync",
+
+            // zfs
+            "**/.zfs/**",
+            "**/.zfs"
         };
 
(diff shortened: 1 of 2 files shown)


crobibero (approved):

Shadowghost (approved):
I've been using ZFS for ages and didn't even know this exists...

systemcrash:
I think there's a tunable. 

`zfs set snapdir={hidden|visible} /mnt/zfspool`

cvium:
Should we not just ignore all hidden folders (parent folders do not count)?

Shadowghost:
Valid point. IMHO it would make sense.

Bond-009:
I'm sure some people keep their media in hidden folders (maybe the nsfw kind)
We should probably not show hidden dirs on the add media library page if this is really desired.

barronpm (approved):

JPVenson:
> I'm sure some people keep their media in hidden folders (maybe the nsfw kind) We should probably not show hidden dirs on the add media library page if this is really desired.

i think this would be a good compromise. Not showing them but if the user pastes the path directly still allow them.

PR #10222, opened 2023-09-13
PhotoResolver: change how generated images are detected

### LINKED ISSUE #7830: [Issue]: Jellyfin won't show me photos of my new Land Rover Discovery!
### Please describe your bug

(OK. The truth is I don't actually have a new Land Rover Discovery, but if I did and I included "discovery" in the filename, then Jellyfin would ignore the photo.)

[PhotoResolver.cs](https://github.com/jellyfin/jellyfin/blob/master/Emby.Server.Implementations/Library/Resolvers/PhotoResolver.cs) has a list of filenames which are ignored when scanning for photos including folder, thumb, landscape, fanart, backdrop, poster, cover, logo and default.

But not only does this exclude photos which **exactly** match those names, but also excludes any photos where any of this words are contained **anywhere** in the name.

Hence it ignores anything with discovery in the name. 

Also one of my photos named 42-factors-poster-design-01.png gets ignored.

For exact matches, I presume this was to exclude tiny thumbnails copies of the photos, but what was the reasoning behind excluding files which happen to contain any of these keywords?

Would be easy to remove the check, but Chesterton's Fence probably applies.

### Jellyfin Version

10.8.0

### if other:

_No response_

### Environment

_No response_

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

Bond-009:
Backdrops/fanart are generated as (backdrop)|(fanart)[0-9]*.extension

Fixes #7830

### DIFF
--- a/Emby.Server.Implementations/Library/Resolvers/PhotoResolver.cs
+++ b/Emby.Server.Implementations/Library/Resolvers/PhotoResolver.cs
@@ -1,5 +1,3 @@
-#nullable disable
-
 using System;
 using System.Collections.Generic;
 using System.IO;
@@ -25,7 +23,7 @@ namespace Emby.Server.Implementations.Library.Resolvers
         private readonly NamingOptions _namingOptions;
         private readonly IDirectoryService _directoryService;
 
-        private static readonly HashSet<string> _ignoreFiles = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
+        private static readonly string[] _ignoreFiles = new[]
         {
             "folder",
             "thumb",
@@ -56,7 +54,7 @@ namespace Emby.Server.Implementations.Library.Resolvers
         /// </summary>
         /// <param name="args">The args.</param>
         /// <returns>Trailer.</returns>
-        protected override Photo Resolve(ItemResolveArgs args)
+        protected override Photo? Resolve(ItemResolveArgs args)
         {
             if (!args.IsDirectory)
             {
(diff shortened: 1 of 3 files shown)


Emby.Server.Implementations/Library/Resolvers/PhotoResolver.cs
```diff
@@ -68,10 +66,10 @@ protected override Photo Resolve(ItemResolveArgs args)
                 {
                     if (IsImageFile(args.Path, _imageProcessor))
                     {
-                        var filename = Path.GetFileNameWithoutExtension(args.Path);
+                        var filename = Path.GetFileNameWithoutExtension(args.Path.AsSpan());
 
                         // Make sure the image doesn't belong to a video file
-                        var files = _directoryService.GetFiles(Path.GetDirectoryName(args.Path));
+                        var files = _directoryService.GetFiles(Path.GetDirectoryName(args.Path) ?? throw new InvalidOperationException("Path can't be a root directory."));
```
barronpm, line 72: IMO `x ?? throw` should be on it's own line, but maybe this check is unnecessary anyways, since we check `args.IsDirectory` earlier

JPVenson: If the directory name is null, there must be something very wrong with the path as GetDirectoryName gets you everything until the first `/`  or `\`. In that case we should return null and log the issue as debug.

JPVenson (changes requested):

Emby.Server.Implementations/Library/Resolvers/PhotoResolver.cs
```diff
@@ -25,7 +23,7 @@ public class PhotoResolver : ItemResolver<Photo>
         private readonly NamingOptions _namingOptions;
         private readonly IDirectoryService _directoryService;
 
-        private static readonly HashSet<string> _ignoreFiles = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
+        private static readonly string[] _ignoreFiles = new[]
```
cvium, line 26: Why change the type? Is Array.Contains faster than HashSet?

Bond-009: For very small arrays (< 5 elements) it can be but I chnged it here bc we don't even call `.Contains` anymore

barronpm (approved):

cvium (approved):

PR #10228, opened 2023-09-15
Replaced some magic values used in property attributes `[MaxLength]` and `[StringLength]` in data entities

Lehonti:
**Changes**
The length of strings in many data entities is constrained by `[MaxLength]` and `[StringLength]` attributes, which are passed a numeric value. However, these constraints come without an explanation. In this pull request, those values that could be neatly represented by the `MaxValue` constants of numeric value types were changed accordingly. Presumably, these values of these constraints mean _something_, and this is a first step towards making the code more self-documenting. I envision that, in the future, it will all be named constants.


cvium:
These are data annotations aka. database constraints. They are somewhat arbitrary and that is fine. Not all magic values need self-documenting constants.

PR #10239, opened 2023-09-17
Change default value of `UseFileCreationTimeForDateAdded` to `false`

### LINKED ISSUE #10001: [Issue]: Order By last added not working properly
### Please describe your bug

When I move a new Folder to the Video Library Folder the added Movie doesnt appear first when I order by last added. This happens obviously because the modification date of the movie isnt touch when moved under linux. If I copy the folder inside a windows environment modification date is updated and it shows as last added movie. 

It would be better when the 'order by' depends on when jellyfin added the movie to its database instead of relying on thos modification dates.

### Jellyfin Version

10.8.z

### if other:

_No response_

### Environment

```markdown
- OS: Linux/Synology DSM
- Clients:  jellyfin-androidtv, jellyfin-web
- Browser: Chromium
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

mkanilsson:
Use current timestamp instead of files `last modified` date by default. This change make jellyfin better match users expectations relating to ordering by "last added" and "Latest \<library name\>".

**Changes**
Change default value of `UseFileCreationTimeForDateAdded` to `false`

**Issues**
Fixes #10001

### DIFF
--- a/MediaBrowser.Model/Configuration/MetadataConfiguration.cs
+++ b/MediaBrowser.Model/Configuration/MetadataConfiguration.cs
@@ -6,7 +6,7 @@ namespace MediaBrowser.Model.Configuration
     {
         public MetadataConfiguration()
         {
-            UseFileCreationTimeForDateAdded = true;
+            UseFileCreationTimeForDateAdded = false;
         }
 
         public bool UseFileCreationTimeForDateAdded { get; set; }


nielsvanvelzen (commented):
I'm against this change. A lot of our users already have a media library when they start with Jellyfin. With this change all of the items in those existing libraries will have the same creation date (the timestamp the library is added to Jellyfin). This will likely cause more issue reports.

Once the server setup is done you can always change this option manually, but the default should be on in my opinion.

mkanilsson:
> A lot of our users already have a media library when they start with Jellyfin.

I did not think of this

> You can always change the option manually

I didn't find the setting yesterday, but I have found it now. I'll post a comment on the issue about how to change it, which is a better solution than this PR. 

Closing the PR.

PR #10242, opened 2023-09-18
Feature/pem ssl support

JPVenson:
Added pem file support


JPVenson:
Only for reference. After discussing the issue with the team, we decided to go ahead and remove the build-in SSL feature in favor of kestrel configuration and native SSL support via a configuration file.

PR #10244, opened 2023-09-19
Allow setting range of ports for dynamic assignment

### LINKED ISSUE #790: Randomized UDP may need to be open for DLNA renderer detection to work
My Jellyfin setup was unable to detect my Sony Blu-Ray player as a DLNA renderer until I opened the randomized UDP port that was previously thought to be only used for live TV in my local firewall.

I discovered this by running a JF server on my laptop and capturing all traffic to/from the Blu-Ray player. Presumably after some kind of unicast discovery, which I filtered out, the player starts sending what seems to be HTTP data to the randomized UDP port on the server.

This obviously makes setting up firewall rules pretty hard, is there any way we can instead hardcode the port without breaking anything?

 - OS: Docker on Debian 9 (with host networking)
 - Jellyfin Version: 10.1.0

tnyeanderson:
When running jellyfin in docker, it has to run in host mode since it needs to listen on random UDP ports. This PR gives the option for a user to set a range of ports that jellyfin will select from when it needs to bind to a random port. Now, the user may avoid host mode by setting aside a range of ports for jellyfin and forwarding all of them to the container.

**Changes**
Adds logic to select a random port between the configured range and try to bind to it. If it fails it will retry a configured number of times with different ports. If the max and min port values are not set (i.e. `0`) then no new behavior is introduced.

**Issues**
Fixes #790

## WIP

This PR is a proof of concept at the moment. This is my first time (since college) writing C# so open to any advice. Still need to wire up configuration of the parameters and add the retry functionality to the other `SocketFactory` methods, but wanted to see if there were any other ideas for implementing this by someone who is more familiar with the available libraries, etc.

I know this implementation isn't ideal, but the bind syscall doesn't allow setting this range. It is a kernel option at `/proc/sys/net/ipv4/ip_local_port_range`, but that can't be set solely for jellyfin (and in a container, is shared by the host and can't/shouldn't be edited). And others have noted that any solution which first checks if a port is available, then binds to it, could be undermined by some other process binding to the port between our check and bind calls anyways. This implementation does not have this problem, at the expense of blindly trying options a certain amount of times until it works (or gives up).

With the default value of 5 tries, I have no idea what the performance impact might be. It could be unacceptable. However, if the user provided a range including 100 available ports and 50 were currently in use, there is a 96.875% chance that the bind will be successful.

I really hope there is a better way to do this. Just wanted to float an idea and get feedback.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -168,6 +168,7 @@
  - [TheTyrius](https://github.com/TheTyrius)
  - [tallbl0nde](https://github.com/tallbl0nde)
  - [sleepycatcoding](https://github.com/sleepycatcoding)
+ - [tnyeanderson](https://github.com/tnyeanderson)
 
 # Emby Contributors
 
--- a/Emby.Server.Implementations/Net/SocketFactory.cs
+++ b/Emby.Server.Implementations/Net/SocketFactory.cs
@@ -20,22 +20,48 @@ namespace Emby.Server.Implementations.Net
                 throw new ArgumentException("localPort cannot be less than zero.", nameof(localPort));
             }
 
+            int min = 33000;
+            int max = 34000;
+            int remainingAttempts = 1;
+            Random rnd = new Random();
+
+            if (localPort == 0 && min != 0 && max != 0)
+            {
+                // If a min and max is set, try to use a port within that range
+                remainingAttempts = 5;
+                localPort = rnd.Next(min, max);
+            }
+
             var socket = new Socket(AddressFamily.InterNetwork, SocketType.Dgram, ProtocolType.Udp);
-            try
-            {
-                socket.EnableBroadcast = true;
-                socket.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.ReuseAddress, true);
-                socket.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.Broadcast, 1);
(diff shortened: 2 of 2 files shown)


JPVenson:
Maybe a bit of clarification. 
Host mode is only required for Chromecast. In all other cases, only ports for http and https(optionally) are needed.

PrplHaz4:
> Maybe a bit of clarification. Host mode is only required for Chromecast. In all other cases, only ports for http and https(optionally) are needed. And i dont think we can force chromecast onto using specific ports.

I believe Host mode also required for DLNA/UPnP.

JPVenson:
Oh yea sory litte mix up :-) yea only for DLNA. 
DLNA requires a multicast on the local network, that is why host mode is required formost as written in our docs. Even if you could make some dynamic ports open, we could still not just send the multicast package as all other kinds of networks for docker (except MacVlan which is also considered obsolete) would not be able to transfer them.

Shadowghost:
Since I changes code touched by this PR, please rebase on current master. Generally speaking I don't have anything against this as long as we keep the current behaviour as default.

gaurovgiri (approved):
LGTM, thanks for the contribution

tnyeanderson:
FYI this is far from complete, the implementation not only needs to be tested, but the configuration options and documentation need to be added. The current state of this PR is merely a proposal/POC

Bond-009:
Closing as stale

PR #10261, opened 2023-09-23
Use IHostLifetime for Shutdown/Restart

barronpm:
**Changes**
Uses the built in `IHostLifetime` infrastructure to handle shutdown/restart

### DIFF
--- a/Emby.Server.Implementations/ApplicationHost.cs
+++ b/Emby.Server.Implementations/ApplicationHost.cs
@@ -101,6 +101,7 @@ using Microsoft.AspNetCore.Mvc;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
 using Prometheus.DotNetRuntime;
 using static MediaBrowser.Controller.Extensions.ConfigurationExtensions;
@@ -848,6 +849,24 @@ namespace Emby.Server.Implementations
             {
                 EventHelper.QueueEventIfNotNull(HasPendingRestartChanged, this, EventArgs.Empty, Logger);
             }
+        }
+
+        /// <inheritdoc />
+        public void Restart()
+        {
+            ShouldRestart = true;
+            Shutdown();
+        }
+
+        /// <inheritdoc />
+        public void Shutdown()
+        {
+            Task.Run(async () =>
+            {
+                await Task.Delay(100).ConfigureAwait(false);
+                IsShuttingDown = true;
+                Resolve<IHostApplicationLifetime>().StopApplication();
+            });
         }
 
         /// <summary>
--- a/Emby.Server.Implementations/ApplicationHost.cs
+++ b/Emby.Server.Implementations/ApplicationHost.cs
@@ -12,7 +12,6 @@ using System.Linq;
 using System.Net;
 using System.Reflection;
 using System.Security.Cryptography.X509Certificates;
-using System.Threading;
 using System.Threading.Tasks;
 using Emby.Dlna;
 using Emby.Dlna.Main;
@@ -112,7 +111,7 @@ namespace Emby.Server.Implementations
     /// <summary>
     /// Class CompositionRoot.
     /// </summary>
-    public abstract class ApplicationHost : IServerApplicationHost, IAsyncDisposable, IDisposable
+    public abstract class ApplicationHost : IServerApplicationHost, IDisposable
     {
         /// <summary>
         /// The disposable parts.
@@ -127,7 +126,6 @@ namespace Emby.Server.Implementations
         private readonly IPluginManager _pluginManager;
 
         private List<Type> _creatingInstances;
-        private ISessionManager _sessionManager;
 
         /// <summary>
         /// Gets or sets all concrete types.
@@ -172,6 +170,8 @@ namespace Emby.Server.Implementations
                 ConfigurationManager.Configuration,
                 ApplicationPaths.PluginsPath,
                 ApplicationVersion);
+
+            _disposableParts.TryAdd((PluginManager)_pluginManager, byte.MinValue);
         }
 
         /// <summary>
@@ -202,7 +202,10 @@ namespace Emby.Server.Implementations
         public bool HasPendingRestart { get; private set; }
 
         /// <inheritdoc />
-        public bool IsShuttingDown { get; private set; }
+        public bool IsShuttingDown { get; set; }
+
+        /// <inheritdoc />
+        public bool ShouldRestart { get; set; }
 
         /// <summary>
         /// Gets the logger.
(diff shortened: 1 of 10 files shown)


Shadowghost (approved):

crobibero:
I get this exception when hitting restart

```
[09:44:10] [FTL] [1] Main: Unhandled Exception
System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'IServiceProvider'.
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.ThrowHelper.ThrowObjectDisposedException()
   at Microsoft.Extensions.DependencyInjection.ServiceProvider.GetService(Type serviceType, ServiceProviderEngineScope serviceProviderEngineScope)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService(IServiceProvider provider, Type serviceType)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService[T](IServiceProvider provider)
   at Jellyfin.Server.Program.StartServer(IServerApplicationPaths appPaths, StartupOptions options, IConfiguration startupConfig) in C:\Users\Cody\Code\jellyfin\Jellyfin.Server\Program.cs:line 187
   at Jellyfin.Server.Program.StartApp(StartupOptions options) in C:\Users\Cody\Code\jellyfin\Jellyfin.Server\Program.cs:line 119
   at Jellyfin.Server.Program.<Main>(String[] args)
Unhandled exception. System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'IServiceProvider'.
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.ThrowHelper.ThrowObjectDisposedException()
   at Microsoft.Extensions.DependencyInjection.ServiceProvider.GetService(Type serviceType, ServiceProviderEngineScope serviceProviderEngineScope)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService(IServiceProvider provider, Type serviceType)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService[T](IServiceProvider provider)
   at Jellyfin.Server.Program.StartServer(IServerApplicationPaths appPaths, StartupOptions options, IConfiguration startupConfig) in C:\Users\Cody\Code\jellyfin\Jellyfin.Server\Program.cs:line 187
   at Jellyfin.Server.Program.StartApp(StartupOptions options) in C:\Users\Cody\Code\jellyfin\Jellyfin.Server\Program.cs:line 119
   at Jellyfin.Server.Program.<Main>(String[] args)
```

crobibero (approved):

Emby.Server.Implementations/ApplicationHost.cs
```diff
@@ -1152,52 +1091,5 @@ protected virtual void Dispose(bool dispose)
...
-            }
-
-            if (_sessionManager is not null)
-            {
-                // used for closing websockets
-                foreach (var session in _sessionManager.Sessions)
-                {
-                    await session.DisposeAsync().ConfigureAwait(false);
-                }
-            }
```
Bond-009, line 1200: This should be added to the non async Dispose method

Emby.Server.Implementations/Session/SessionManager.cs
```diff
@@ -44,7 +45,7 @@ namespace Emby.Server.Implementations.Session
     /// <summary>
     /// Class SessionManager.
     /// </summary>
-    public class SessionManager : ISessionManager, IDisposable
+    public sealed class SessionManager : ISessionManager, IAsyncDisposable
```
Bond-009, line 48: You should implement IDisposable when implementing IAsyncDisposable

Jellyfin.Server/Program.cs
```diff
@@ -65,36 +63,9 @@ static Task ErrorParsingArguments(IEnumerable<Error> errors)
...
-
-            Shutdown();
-        }
-
         private static async Task StartApp(StartupOptions options)
         {
             _startTimestamp = Stopwatch.GetTimestamp();
-
-            // Log all uncaught exceptions to std error
-            static void UnhandledExceptionToConsole(object sender, UnhandledExceptionEventArgs e) =>
```
Bond-009, line 94: Is this not needed anymore?

Bond-009 (changes requested):

Jellyfin.Server/Program.cs
barronpm: The default behavior is already to log to console, so this was logging everything to the console twice.

Emby.Server.Implementations/ApplicationHost.cs
barronpm: This code has been moved to the SessionManager `OnApplicationStopping` method, as well as it's dispose method just in case that never gets properly called.

Emby.Server.Implementations/Session/SessionManager.cs
barronpm: Since the lifetime of this object is managed by the service collection, this is guaranteed to be correctly disposed by the container. 

See https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync

Emby.Server.Implementations/Session/SessionManager.cs
```diff
@@ -1833,5 +1778,51 @@ public Task SendMessageToUserDeviceSessions<T>(string deviceId, SessionMessageTy
...
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, "Error sending server shutdown notifications");
+            }
+
+            // Close open websockets to allow Kestrel to shut down cleanly
+            foreach (var session in _activeConnections.Values)
+            {
+                await session.DisposeAsync().ConfigureAwait(false);
```
cvium, line 1824: Aren't these disposed twice? Once here and then again in DisposeAsync?

Jellyfin.Server/Program.cs
```diff
@@ -210,7 +152,7 @@ private static async Task StartServer(IServerApplicationPaths appPaths, StartupO
 
                 try
                 {
-                    await host.StartAsync(_tokenSource.Token).ConfigureAwait(false);
+                    await host.StartAsync().ConfigureAwait(false);
```
cvium, line 155: Why remove the token?

barronpm: The token was only used in our listeners for SIGTERM/Ctrl+C, which is handled by the ConsoleLifetime

Emby.Server.Implementations/Session/SessionManager.cs
barronpm: Yes, kestrel waits for the web sockets to close before it is considered to have shut down, so it would hang if this wasn't here. This doesn't cause any problems, since disposing something a second time has no effect (or shouldn't, anyway), but ideally we would do something like passing the ApplicationStopping token to the web sockets so that they stop without needing this, but idk if that would introduce a race with sending out the shutdown messages, since they also happen on that event.

Bond-009 (approved):

cvium (approved):

Emby.Server.Implementations/ApplicationHost.cs
```diff
@@ -202,7 +203,10 @@ public abstract class ApplicationHost : IServerApplicationHost, IAsyncDisposable
         public bool HasPendingRestart { get; private set; }
 
         /// <inheritdoc />
-        public bool IsShuttingDown { get; private set; }
+        public bool IsShuttingDown { get; set; }
```
cvium, line 206:
```suggestion
        public bool IsShuttingDown { get; private set; }
```

Emby.Server.Implementations/ApplicationHost.cs
```diff
@@ -202,7 +203,10 @@ public abstract class ApplicationHost : IServerApplicationHost, IAsyncDisposable
...
+
+        /// <inheritdoc />
+        public bool ShouldRestart { get; set; }
```
cvium, line 209:
```suggestion
        public bool ShouldRestart { get; private set; }
```

PR #10324, opened 2023-10-02
Feature/enable gh codespaces

JPVenson:
Made nesseary updates to run the Jellyfin Server on github Codespaces.

**Changes**
Added `BindAllNetworkManager` for use with environment variable. This is necessary as ghcs will not detect jellyfins ports otherwise.
Updates readme.md to include information's about running in ghcs


Emby.Server.Implementations/ApplicationHost.cs
```diff
@@ -448,7 +448,16 @@ public void Init(IServiceCollection serviceCollection)
 
             ConfigurationManager.AddParts(GetExports<IConfigurationFactory>());
 
-            NetManager = new NetworkManager(ConfigurationManager, _startupConfig, LoggerFactory.CreateLogger<NetworkManager>());
+            // this is needed as Github Codespaces will not be able to find the process if we bind on interfaces directly.
+
+            if (Environment.GetEnvironmentVariable("JELLYFIN_NETWORKMANAGER_BINDALL") == "true")
```
crobibero, line 453:
```suggestion
            if (bool.TryParse(Environment.GetEnvironmentVariable("JELLYFIN_NETWORKMANAGER_BINDALL"), out var bindAll) && bindAll)
```

Doesn't really matter but might as well attempt to parse the bool

JPVenson: image

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,31 @@
+{
+    "name": "Plain JF Dotnet 8",
+    "image":"mcr.microsoft.com/devcontainers/dotnet:8.0-jammy",
+        //required to get skia to run
+	"postCreateCommand": "sudo apt-get update && sudo apt-get install -y libfontconfig1",
```
ferferga, line 5: Use features instead so all apt data is cleared afterwards. Like this:

"ghcr.io/devcontainers-contrib/features/apt-packages:1": {
      "preserve_apt_list": false,
      "packages": ["libfontconfig1"]
    }

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,31 @@
...
+    "customizations": {
+        "vscode": {
+            "extensions": [
+                "GitHub.vscode-github-actions",
+                "ms-dotnettools.vscode-dotnet-runtime",
+                "ms-dotnettools.csdevkit",
+                "ms-dotnetools.csharp"
+            ]
+        }
+    },
```
ferferga, line 26: Better if you do this so no need to keep 2 lists updated: https://github.com/jellyfin/jellyfin-vue/blob/master/.devcontainer/devcontainer.json#L28

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,31 @@
+{
+    "name": "Plain JF Dotnet 8",
+    "image":"mcr.microsoft.com/devcontainers/dotnet:8.0-jammy",
+        //required to get skia to run
+	"postCreateCommand": "sudo apt-get update && sudo apt-get install -y libfontconfig1",
+    "postStartCommand": "dotnet restore; dotnet workload update; dotnet dev-certs https --trust",
+    "containerEnv": {
+        "JELLYFIN_NETWORKMANAGER_BINDALL": "true"
+    },
+    "features": {
```
ferferga, line 10: I'd add the same features as in [Vue](https://github.com/jellyfin/jellyfin-vue/blob/master/.devcontainer/devcontainer.json#L11C60-L11C60)

- Having a setup ready for Docker is always a good idea
- gh cli is useful for checking out external's contributor's PRs, firing actions, etc etc

Emby.Server.Implementations/ApplicationHost.cs
```diff
@@ -448,7 +448,16 @@ public void Init(IServiceCollection serviceCollection)
...
+            // this is needed as Github Codespaces will not be able to find the process if we bind on interfaces directly.
+
+            if (bool.TryParse(Environment.GetEnvironmentVariable("JELLYFIN_NETWORKMANAGER_BINDALL"), out var bindAll) && bindAll)
+            {
+                NetManager = new BindAllNetworkManager(ConfigurationManager, _startupConfig, LoggerFactory.CreateLogger<NetworkManager>());
+            }
+            else
+            {
+                NetManager = new NetworkManager(ConfigurationManager, _startupConfig, LoggerFactory.CreateLogger<NetworkManager>());
+            }
```
ferferga, line 460: I don't understand why this is necessary. Why not [add the explicit port definition](https://github.com/jellyfin/jellyfin-vue/blob/master/.devcontainer/devcontainer.json#L14)?

README.md
```diff
@@ -142,6 +142,21 @@ cd Jellyfin.Server/bin/Debug/net8.0 # Change into the build output directory
 
 2. Execute the build output. On Linux, Mac, etc. use `./jellyfin` and on Windows use `jellyfin.exe`.
 
+### Running from GH-Codespaces
+
+Building and running the server from Github Codespaces require you to install the recommended VS-Code extensions as listed in the `.vscode/extensions.json` file. VsCode should ask you to install them each time you open the Server repro on GH-Codespaces.
+
```
ferferga, line 148: This is unnecessary if you follow my feedback

README.md
```diff
@@ -142,6 +142,21 @@ cd Jellyfin.Server/bin/Debug/net8.0 # Change into the build output directory
...
+As Jellyfin will run on a container on a github hosted server, JF needs to handle 3 things differently.
+1. Address Binding.
+Github will put all networked services behind a reverse proxy and will only autodetect it correctly if it will bind to `Ipv4.Any`. Because of this, there is an adapted `BindAllNetworkManager` that will overwrite all set bindings and bind to any IPV4 address.
+**NOTE:** If you want to access the JF instance from outside, like with a WebClient on another PC, remember to set the "ports" in the lower VsCode window to public.
+
+2. FFmpeg installation.
+Because sometimes you need FFMPEG to test certain cases, follow the instructions from the wiki on the dev enviorment:
+https://jellyfin.org/docs/general/installation/linux/#ffmpeg-installation
```
ferferga, line 156: Can't we do this as part of the container creation process?

ferferga (changes requested):

Emby.Server.Implementations/ApplicationHost.cs
JPVenson: i tried that but as i described in the readme docs, GH codespaces will not bind the proxy correctly to the ports. I think at the time JF starts the virtual interface might not be available yet and it binds its later and then JF does not react to it.

README.md
JPVenson: Yes absolutly but this is just the basic setup. I plan to create different modules for development and for testing for example. But this is the KISS approach (as it again already takes 2 months to merge this little)

.devcontainer/devcontainer.json
JPVenson: i can add docker-in-docker as its a great i dea but i dont think gh-cli is strictly nessary as i am currently writing this from vscode without it.

Emby.Server.Implementations/ApplicationHost.cs
JPVenson: It seems like the issue was already addressed by chance 2 months ago. i could remove the bind-all flag for what its worth but i sometimes see it pop up as an feature request so i would like to keep it anyway. Thoughts?

README.md
```diff
@@ -142,6 +142,19 @@ cd Jellyfin.Server/bin/Debug/net8.0 # Change into the build output directory
 
 2. Execute the build output. On Linux, Mac, etc. use `./jellyfin` and on Windows use `jellyfin.exe`.
 
+### Running from GH-Codespaces
+
+As Jellyfin will run on a container on a github hosted server, JF needs to handle 3 things differently.
+1. Address Binding.
+Github will put all networked services behind a reverse proxy and will only autodetect it correctly if it will bind to `Ipv4.Any`. Because of this, there is an adapted `BindAllNetworkManager` that will overwrite all set bindings and bind to any IPV4 address.
+**NOTE:** If you want to access the JF instance from outside, like with a WebClient on another PC, remember to set the "ports" in the lower VsCode window to public.
```
barronpm, line 150: This section is mostly outdated now

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,28 @@
+{
+    "name": "Plain JF Dotnet 8",
```
barronpm, line 2: Why this name, instead of just "Jellyfin" or similar? We always develop against the latest .NET version, so is there any reason to delineate?

JPVenson: Well i plan to make multiple versions for different scenarios for example:
- Development Jellyfin Server 
- Development Jellyfin Server - JFFmpeg
- Development Jellyfin Server - Standard FFmpeg
- Development Jellyfin Server - JFFmpeg with Media files
- Test Jellyfin Server
- Test Jellyfin Server - JFFmpeg
- Test Jellyfin Server - Standard FFmpeg
- Test Jellyfin Server - JFFmpeg with Media files
- Test Jellyfin Server - Multi User
- Test Jellyfin Server - DLNA

And so on. The dotnet8 moniker was because when i started server still targeted and older net version and you cannot use it currently to run the 10.8.13 release for example.

barronpm: So this one should be `Development Jellyfin Server` right? We shouldn't need to worry about ever supporting multiple versions, since once 10.9 is released, even when we upgrade master to .NET 9, the version of this file on the 10.9.x branch would still target .NET 8, no?

JPVenson: > So this one should be Development Jellyfin Server right? 

Yes
> even when we upgrade master to .NET 9, the version of this file on the 10.9.x branch would still target .NET 8, no?

That is actually a good point.

barronpm (approved):

JPVenson:
to whoever merges:

Please squash when merging

crobibero (approved):

PR #10328, opened 2023-10-03
Add functionality to remove empty media folders on item delete

### LINKED ISSUE #832: DVR program deletion leaves artifact files/directories
When deleting the last episode of a recorded program, JF should delete the entire program's directory tree.  This is not happening.  Artifacts are being left behind:

user@host:/data-drive/jellyfin_dvr_tv$ find Schooled/                                                                                                                   
Schooled/                                                                                                                                                                
Schooled/Season 1                                                                                                                                                        
Schooled/tvshow.nfo 

These files/directories should be cleaned up and removed to prevent a bunch of empty directories and orphaned files.

Dangles91:
**Changes**
- Make LibraryManager.DeleteItem async
- Remove LibraryManager.DeleteItem notifyParentItem param (unused)
- Delete all metadata and images on media item delete (when deleting media item from filesystem)
- Remove empty folders from filesystem on media item delete

Eempty media folders will be deleted when the last item is deleted. Eg. deleting the last episode from a season in a series would delete the season in the library and directory on disk. If that season was the last season for that series, the series would also be deleted from the library and disk.

Tested on Windows only, though I expect it should work for other operating systems. 

Detection for existing files uses the same ignore list as media import to ensure standard hidden files are not mistaked for user files. If there are any remaning files that are not safe to ignore and do not belong to Jellyfin, the delete action for the folder will be skipped.

**Issues**
ex. Fixes #832

### DIFF
--- a/Emby.Server.Implementations/Channels/ChannelManager.cs
+++ b/Emby.Server.Implementations/Channels/ChannelManager.cs
@@ -761,15 +761,14 @@ namespace Emby.Server.Implementations.Channels
                     var deadItem = _libraryManager.GetItemById(deadId);
                     if (deadItem is not null)
                     {
-                        _libraryManager.DeleteItem(
+                        await _libraryManager.DeleteItemAsync(
                             deadItem,
+                            parentItem,
                             new DeleteOptions
                             {
                                 DeleteFileLocation = false,
                                 DeleteFromExternalProvider = false
-                            },
-                            parentItem,
-                            false);
+                            });
                     }
                 }
             }
--- a/Emby.Server.Implementations/Channels/ChannelPostScanTask.cs
+++ b/Emby.Server.Implementations/Channels/ChannelPostScanTask.cs
@@ -38,15 +38,14 @@ namespace Emby.Server.Implementations.Channels
         /// <param name="progress">The progress.</param>
         /// <param name="cancellationToken">The cancellation token.</param>
         /// <returns>The completed task.</returns>
-        public Task Run(IProgress<double> progress, CancellationToken cancellationToken)
+        public async Task Run(IProgress<double> progress, CancellationToken cancellationToken)
         {
-            CleanDatabase(cancellationToken);
+            await CleanDatabaseAsync(cancellationToken);
 
             progress.Report(100);
-            return Task.CompletedTask;
         }
 
-        private void CleanDatabase(CancellationToken cancellationToken)
+        private async Task CleanDatabaseAsync(CancellationToken cancellationToken)
         {
             var installedChannelIds = ((ChannelManager)_channelManager).GetInstalledChannelIds();
 
@@ -60,11 +59,11 @@ namespace Emby.Server.Implementations.Channels
             {
                 cancellationToken.ThrowIfCancellationRequested();
 
-                CleanChannel((Channel)channel, cancellationToken);
+                await CleanChannelAsync((Channel)channel, cancellationToken);
             }
         }
 
-        private void CleanChannel(Channel channel, CancellationToken cancellationToken)
+        private async Task CleanChannelAsync(Channel channel, CancellationToken cancellationToken)
         {
             _logger.LogInformation("Cleaning channel {0} from database", channel.Id);
 
@@ -78,23 +77,21 @@ namespace Emby.Server.Implementations.Channels
             {
                 cancellationToken.ThrowIfCancellationRequested();
 
-                _libraryManager.DeleteItem(
+                await _libraryManager.DeleteItemAsync(
                     item,
                     new DeleteOptions
                     {
                         DeleteFileLocation = false
-                    },
-                    false);
+                    });
             }
 
             // Finally, delete the channel itself
-            _libraryManager.DeleteItem(
+            await _libraryManager.DeleteItemAsync(
                 channel,
                 new DeleteOptions
                 {
                     DeleteFileLocation = false
-                },
-                false);
+                });
         }
     }
 }
--- a/Emby.Server.Implementations/Data/CleanDatabaseScheduledTask.cs
+++ b/Emby.Server.Implementations/Data/CleanDatabaseScheduledTask.cs
@@ -22,11 +22,10 @@ namespace Emby.Server.Implementations.Data
 
         public Task Run(IProgress<double> progress, CancellationToken cancellationToken)
         {
-            CleanDeadItems(cancellationToken, progress);
-            return Task.CompletedTask;
+            return CleanDeadItemsAsync(cancellationToken, progress);
         }
 
-        private void CleanDeadItems(CancellationToken cancellationToken, IProgress<double> progress)
+        private async Task CleanDeadItemsAsync(CancellationToken cancellationToken, IProgress<double> progress)
         {
             var itemIds = _libraryManager.GetItemIds(new InternalItemsQuery
             {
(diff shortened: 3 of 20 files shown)


Dangles91:
Example screenshot of how this could look in jellyfin-web (working implementation)
image

Shadowghost:
I'm not sure if an option for this makes sense. If we properly check that the directory is empty (e.g. no hidden files) we should just remove the folder.

Dangles91:
If we can proceed without an option and have it as a default behaviour that would simplify things.

The current check is based on known items in the library db. We could extend that check and only delete the folder if there are no jellyfin managed files, Ie. no metadata, images, or media, and no other user or OS files?

splendid25:
I'm not sure I would want this enabled by default on my libraries.
I could empty a series folder because I've watched everything, and re-add videos later. The series custom metadata would be lost in that case (at least the "date added").

And what if the user running Jellyfin doesn't have write access to the folder?

JPVenson:
I dont think this is necessary good for jellyfin. There are already users complaining that Jellyfin deletes items, they dont expect (which is mostly a misconception on what JF does). The Delete should really only delete the one item its targeting not more.

Dangles91:
> I'm not sure I would want this enabled by default on my libraries.
> I could empty a series folder because I've watched everything, and re-add videos later. The series custom metadata would be lost in that case (at least the "date added").
> 
> And what if the user running Jellyfin doesn't have write access to the folder?

I'm not sure I understand this particular scenario. In the current implementation the episode metadata is already deleted, the same goes for movies.

When deleting all episodes on a series, jellyfin will keep the series and seasons and display empty media items. This would frustrate me as a user - if a show is not available to watch it should not appear as available.

If the user running jellyfin does not have write access then the delete action would fail?

However, having an option available for administrators to configure is never a bad thing. We just need a good place to put it. With the limitations of library management in general, I think this would need to be a global configuration. Any thoughts?

Dangles91:
> I dont think this is necessary good for jellyfin. There are already users complaining that Jellyfin deletes items, they dont expect (which is mostly a misconception on what JF does). The Delete should really only delete the one item its targeting not more.  

Based on a number issues entered around delete there also appears to be many complaining that it does not delete enough and does not clean up after itself well.

Perhaps having an option for administrators to configure would be a better approach?

Do you also have any examples of JF deleting things it should not? I've seen one misconception around collections, however this is just a communication issue with the message displayed in JF Web

JPVenson:
> > I dont think this is necessary good for jellyfin. There are already users complaining that Jellyfin deletes items, they dont expect (which is mostly a misconception on what JF does). The Delete should really only delete the one item its targeting not more.
> 
> Based on a number issues entered around delete there also appears to be many complaining that it does not delete enough and does not clean up after itself well.
> 
> Perhaps having an option for administrators to configure would be a better approach?
> 
> Do you have any examples of JF deleting things it should not? I've seen one misconception around collections, however this is just a communication issue with the message displayed in JF Web. But as I'm new to JF development I'm not aware of the history

Sory if i badly said that, its more like the other way around. People complaining in chat/gh that stuff was deleted that JF would never be able to delete. Which is just sometimes hard to explain to them that JF only deletes stuff in this very only function (apart from very badly configuring the system). 
As i see many of those complains in the chat, i am careful with expanding that method to delete even more stuff which can lead to users not "understanding" what they are about to do. 
Only really way i would say this would work if the webUI would __clearly__ list all items they are about to delete. Otherwise it will be near impossible even with logfiles to trace what JF does at this function at any given time.

Dangles91:
> > > I dont think this is necessary good for jellyfin. There are already users complaining that Jellyfin deletes items, they dont expect (which is mostly a misconception on what JF does). The Delete should really only delete the one item its targeting not more.
> > 
> > Based on a number issues entered around delete there also appears to be many complaining that it does not delete enough and does not clean up after itself well.
> > 
> > Perhaps having an option for administrators to configure would be a better approach?
> > 
> > Do you have any examples of JF deleting things it should not? I've seen one misconception around collections, however this is just a communication issue with the message displayed in JF Web. But as I'm new to JF development I'm not aware of the history
> 
> Sory if i badly said that, its more like the other way around. People complaining in chat/gh that stuff was deleted that JF would never be able to delete. Which is just sometimes hard to explain to them that JF only deletes stuff in this very only function (apart from very badly configuring the system). 
> As i see many of those complains in the chat, i am careful with expanding that method to delete even more stuff which can lead to users not "understanding" what they are about to do. 
> Only really way i would say this would work if the webUI would __clearly__ list all items they are about to delete. Otherwise it will be near impossible even with logfiles to trace what JF does at this function at any given time.

I think I understand what you mean now. This implementation does use a check to see if the folder is "empty" before deleting it, but it will first delete the metadata before performing that check.

And I say "empty" because it does ignore the same set of files which are ignored when running a library scan - system files, file system files, etc.

So in practice the folder cannot be deleted unless it is empty. We also ignore top level folders so entire libraries are also unaffected.

Adding a UI message with a list of all files would require the same effort as logging all file names before deletion, so having sufficient logs wouldn't be a problem (and maybe should be done anyway).

 A series of prompts would slow things down - as a power use I prefer to skip prompts where I can - so if prompts are required we should also provide an option to disable them.

Maybe a better prompt would be a message letting the user know there are no longer any media files associated with the parent and asking them if they would like to delete it?

JPVenson:
>Adding a UI message with a list of all files would require the same effort as logging all file names before deletion, so having sufficient logs wouldn't be a problem (and maybe should be done anyway).
Agreed. This will be a must have in any case.


> Maybe a better prompt would be a message letting the user know there are no longer any media files associated with the parent and asking them if they would like to delete it?

I think that would be a good compromise here.

Dangles91:
> >Adding a UI message with a list of all files would require the same effort as logging all file names before deletion, so having sufficient logs wouldn't be a problem (and maybe should be done anyway).
> 
> Agreed. This will be a must have in any case.
> 
> 
> > Maybe a better prompt would be a message letting the user know there are no longer any media files associated with the parent and asking them if they would like to delete it?
> 
> I think that would be a good compromise here.

I could also see a scenario where someone might want to retain their file based metadata but remove the parent from their media library.

Maybe two settings could be implemented. One for deleting empty directories and another to skip empty folders during a library scan?

splendid25:
> I'm not sure I understand this particular scenario. In the current implementation the episode metadata is already deleted, the same goes for movies.

Yeah I would definitely enable this setting for a movie library.
I would see this as a setting per library maybe ?

> If a show is not available to watch it should not appear as available.

I agree, but that's another issue concerning the WebUI that doesn't filter anything by default.
I mostly use Kodi, and it hides watched or empty series when you set the filters.

> Skip empty folders during a library scan

Isn't it the current implementation already? If not, yes it should be, I don't see the point of importing an empty folder.
But then you'd have some old empty folders never imported, so never deleted. And some new folders deleted when their content is gone...
It might not be easy to understand for users.

It's hard to find an absolute rule that suits every use case here, without a prompt at some point.

Dangles91:
What are people's thoughts on making this a server wide option as an incremental change? 

Library options aren't currently applied at a granular level when one item is a child of multiple libraries. Adding a library option would be misleading without significant change.

The prompt that is displayed when deleting media could be changed when this option is enabled informing users that the top level folder will be deleted if there are no other items associated with it.

splendid25:
I think you can skip File.Exists before File.Delete :

https://learn.microsoft.com/en-US/dotnet/api/system.io.file.delete?view=net-7.0#remarks
> If the file to be deleted does not exist, no exception is thrown.

Issue #10333, opened 2023-10-03
[Issue]: GitHub stale issue action doesn't seem to be closing old issues


jacobslusser:
### Please describe your bug

As of this writing, there are 703 open issues in the Jellyfin project. If you sort by least recently updated (and ignore those with the following tags: regression, security, roadmap, future, feature, enhancement, confirmed) there are a number of open issues that should have been closed by the GitHub stale action, [.github/workflows/repo-stale.yaml](https://github.com/jellyfin/jellyfin/blob/467aa40d215131e15f1ded162e8768a8df90e7d0/.github/workflows/repo-stale.yaml#L12C1-L34C1)

The action logs suggest a fix might be to [increase the number of operations](https://github.com/jellyfin/jellyfin/actions/runs/6387647162/job/17336290340#step:2:8461), however, that would need to be confirmed.

### Jellyfin Version

Other

### if other:

N/A

### Environment

```markdown
N/A
```

### Jellyfin logs

```shell
N/A
```

### FFmpeg logs

```shell
N/A
```

### Please attach any browser or client logs here

N/A

### Please attach any screenshots here

![image](https://github.com/jellyfin/jellyfin/assets/345183/842648e5-566b-4cea-8281-d0d2a3be2d07)

![image](https://github.com/jellyfin/jellyfin/assets/345183/854265d1-c44e-4279-af99-c3423dbf947c)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

jacobslusser:
I can submit a PR to fix this issue if the maintainers agree with the issue and probable cause.

sevenrats:
we just added a handful of members to the triage team, and we are in the process of cleaning this up now.  how do you want this issue to proceed? i suspect all these old issues are going to get handled by human intelligence is the next few days, but if you think your PR is  still a good idea it would be appreciated.

jacobslusser:
I have submitted the PR. It has been approved but not merged yet. Happy to help.

jacobslusser:
I don't believe the PR that was intended to fix this has worked. I still see the "stale issue" action starting from the most recent issue on every run, instead of from where it finished in the last run....

srcrist:
Jacob, did you have any thoughts on how to proceed here?

jacobslusser:
I can only conclude that the stale action doesn't work as advertised and is not persisting state between executions, despite their documentation and my every attempt to get it to do so.

Here is what I think I will do to work around the issue:

[...]

PR #10334, opened 2023-10-04
#10333 Updates issue stale action to fix issues with not running

jacobslusser:
GitHub actions are difficult to test locally, so I'm making some educated guesses about why the stale action was not running correctly. It looks like the primary culprit is that the action is not caching state and is subsequently starting with the most recent issue/PR and never getting past the first 75 as described in [this issue](https://github.com/actions/stale/issues/1090#issuecomment-1710940734). A view of the action logs corroberates this.

**Changes**
- Adds the `actions: write` permission to enable writing to the cache to maintain state
- Updates the `cron` to run twice daily instead of once
- Updates the verbiage on the `stale-issue-message` to be more concise
- Updates the link for getting help
- Adds a `close-issue-message`

**Issues**
#10333

### DIFF
--- a/.github/workflows/repo-stale.yaml
+++ b/.github/workflows/repo-stale.yaml
@@ -2,20 +2,21 @@ name: Stale Check
 
 on:
   schedule:
-    - cron: '30 1 * * *'
+    - cron: '30 */12 * * *'
   workflow_dispatch:
 
 permissions:
   issues: write
   pull-requests: write
+  actions: write
 
 jobs:
   issues:
-    name: Check issues
+    name: Check for stale issues
     runs-on: ubuntu-latest
     if: ${{ contains(github.repository, 'jellyfin/') }}
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/repo-stale.yaml
```diff
@@ -2,20 +2,21 @@ name: Stale Check
...
 
 jobs:
   issues:
-    name: Check issues
+    name: Check for stale issues
     runs-on: ubuntu-latest
     if: ${{ contains(github.repository, 'jellyfin/') }}
     steps:
-      - uses: actions/stale@1160a2240286f5da8ec72b1c0816ce2481aabf84 # v8.0.0
+      - uses: actions/stale@v8
```
Shadowghost, line 19: This will likely be reverted to the hash by renovate

Shadowghost (approved):
Sounds reasonable to me.

Bond-009 (approved):

crobibero (approved):

PR #10337, opened 2023-10-04
Add SystemManager service

barronpm:
**Changes**
Moves system info and shutdown/restart into a `SystemManager` service

### DIFF
--- a/Emby.Server.Implementations/ApplicationHost.cs
+++ b/Emby.Server.Implementations/ApplicationHost.cs
@@ -101,7 +101,6 @@ using Microsoft.AspNetCore.Mvc;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
 using Prometheus.DotNetRuntime;
 using static MediaBrowser.Controller.Extensions.ConfigurationExtensions;
@@ -133,7 +132,7 @@ namespace Emby.Server.Implementations
         /// <value>All concrete types.</value>
         private Type[] _allConcreteTypes;
 
-        private bool _disposed = false;
+        private bool _disposed;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="ApplicationHost"/> class.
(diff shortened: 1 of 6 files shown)


Emby.Server.Implementations/SystemManager.cs
```diff
@@ -0,0 +1,108 @@
...
+    public void Restart() => ShutdownInternal(true);
+
+    /// <inheritdoc />
+    public void Shutdown() => ShutdownInternal(false);
+
+    private void ShutdownInternal(bool restart)
+    {
+        Task.Run(async () =>
+        {
+            await Task.Delay(100).ConfigureAwait(false);
```
cvium, line 103: Isn't the Task.Run redundant since you're immediately calling Task.Delay? As I understand it, Task.Run will schedule the execution on a threadpool thread, `await Task.Delay` will then release the thread and continue on (potentially) another thread after 100ms

barronpm: Presumably it's to avoid blocking so the API has time to respond before the server starts to shut down

cvium: Yes, but in this case there's no difference between Task.Run and simply not awaiting it. With Task.Run you're using an extra thread. Doesn't matter much since it's going to restart the webserver though

Emby.Server.Implementations/ApplicationHost.cs
```diff
@@ -507,6 +496,8 @@ protected virtual void RegisterServices(IServiceCollection serviceCollection)
             serviceCollection.AddSingleton<IFileSystem, ManagedFileSystem>();
             serviceCollection.AddSingleton<IShortcutHandler, MbLinkShortcutHandler>();
 
+            serviceCollection.AddScoped<ISystemManager, SystemManager>();
```
cvium, line 499: I've started moving these into Startup btw

cvium (approved):

Emby.Server.Implementations/SystemManager.cs
barronpm: If there was no `Task.Run`, and we just didn't await the `Task.Delay`, wouldn't that mean that the `StopApplication()` call would happen immediately instead of in 100ms?

Bond-009 (approved):

PR #10339, opened 2023-10-05
Fix using movie.nfo first when <filename>.nfo also exists

### LINKED ISSUE #1558: Prefer movie.nfo when other nfo files may exist in the same directory within a Movie folder
**Is your feature request related to a problem? Please describe.**
If other nfo files exist within a Movie directory alongside movie.nfo, the movie fails to parse properly.

**Describe the solution you'd like**
Either prefer movie.nfo over other nfo files (could be hacky) OR if an nfo file exists that is not parsed as a true movie xml, catch the error properly and skip to the next nfo file in the Movie directory.

**Describe alternatives you've considered**
Manual move/rename of a large amount of nfo files that would be tough.

**Additional context**
Kodi (native) and Plex (Kodi import) both support this

leovan:
`<filename>.nfo` is used for store custom information of movie uploader on some private tracker bittorrent website, which is not a valid movie metadata nfo file. Overwrite `<filename>.nfo` will lead errors with file sharing. Therefor using `movie.nfo` first may be better when `<filename>.nfo` also exists.

**Changes**
Use `movie.nfo` first when `<filename>.nfo` also exists.

**Issues**
Fixes #1558

### DIFF
--- a/MediaBrowser.XbmcMetadata/Savers/MovieNfoSaver.cs
+++ b/MediaBrowser.XbmcMetadata/Savers/MovieNfoSaver.cs
@@ -60,13 +60,13 @@ namespace MediaBrowser.XbmcMetadata.Savers
             }
             else
             {
-                yield return Path.ChangeExtension(item.Path, ".nfo");
-
                 // only allow movie object to read movie.nfo, not owned videos (which will be itemtype video, not movie)
                 if (!item.IsInMixedFolder && item.ItemType == typeof(Movie))
                 {
                     yield return Path.Combine(item.ContainingFolderPath, "movie.nfo");
                 }
+
+                yield return Path.ChangeExtension(item.Path, ".nfo");
             }
         }
 


Shadowghost (approved):
I don't see an issue with this.

barronpm:
If there are multiple movies in a folder, would this force the same nfo onto all of them even if they have their own?

Shadowghost:
That is correct and a problem. Likely the reason why this check was placed after the direct match.

Not sure how we could remedy this though. I'd rather drop this PR if we don't find a way to handle this.

leovan:
> If there are multiple movies in a folder, would this force the same nfo onto all of them even if they have their own?

In this code, I see only when `item.IsInMixedFolder` is `false`, `movie.nfo` will be the candidate metadata file. And in `MovieResolver.cs` when `resolverResult.Count > 1`, which I think it means there are multiple video files in `parent` folder, then `isInMixedFolder` will be `true`, then `movie.nfo` won't be the candidate metadata file.

https://github.com/jellyfin/jellyfin/blob/40f7eb4e8cd9e250fb3870a4799a5a8d949e2068/Emby.Server.Implementations/Library/Resolvers/Movies/MovieResolver.cs#L285

Sorry I didn't fully read the codes. So please check whether the condition `!item.IsInMixedFolder && item.ItemType == typeof(Movie)` can avoid the problem you mentioned or not.

leovan:
> That is correct and a problem. Likely the reason why this check was placed after the direct match.
> 
> Not sure how we could remedy this though. I'd rather drop this PR if we don't find a way to handle this.

If the condition `!item.IsInMixedFolder && item.ItemType == typeof(Movie)` cannot avoid the problem. A better solution is to check whether `<filename>.nfo` is valid metadata file. If not then use `movie.nfo`. Is there a helper function to check whether a nfo file is a valid metadata file (at least is a valid xml file)? In the situation I mentioned before, the `<filename.nfo>` created by movie uploader is just a plain text file not a valid xml file.

leovan:
I made a minimal test on my machine and it seems working fine. I think `item.IsInMixedFolder` can tell whether a movie fold has multiple video files or not.

test-01

test-02

It uses their own nfo file to show the title.

Shadowghost:
What you tested is only different movies in a single folder if I understand the pictures correctly?
What we're concerned about is multiple versions in the same folder:
```
Matrix (1999) - 1080p.mkv
Matrix (1999) - 1080p.nfo
Matrix (1999) - 2160p.mkv
Matrix (1999) - 2160p.nfo
movie.nfo
```

leovan:
I made the test of multiple versions in the same folder, it works too.

test-03

test-04

Shadowghost:
Your structure is not correct though, you'd need to test it with a folder structure like this:

```
<library root>
| - Matrix (1999)
     | - Matrix (1999) - 1080p.mkv
     | - Matrix (1999) - 1080p.nfo
     | - Matrix (1999) - 2160p.mkv
     | - Matrix (1999) - 2160p.nfo
     | - movie.nfo
```

This is because the multi-version matching depends on the folder name (see our [naming guidelines](https://jellyfin.org/docs/general/server/media/movies))

leovan:
I can see the problem now after changing the parent folder name to `Matrix (1999)`, it use `Matrix movie.nfo` as title for the collection name which I think is reasonable.

屏幕截图 2023-10-06 163902

Problem is the detail of each version still uses `movie.nfo` as its metadata file. (Titles for 1080p and 2160p are same, `Matrix movie.nfo`)

屏幕截图 2023-10-06 164026
屏幕截图 2023-10-06 164034

But after removing `movie.nfo`, it use `Matrix 1080p` as title for the collection name which I think it use the first nfo file ordered by nfo filename.

屏幕截图 2023-10-06 164351

But in this case, it still has the problem. (Titles for 1080p and 2160p are same, `Matrix 1080p`)

屏幕截图 2023-10-06 164431
屏幕截图 2023-10-06 164441

I think this problem is caused by something beyond this PR to fix.

So, a better solution is to check whether `<filename>.nfo` is a valid xml file before adding it to the candidate nfo files. I will try to fix it in this way later.

leovan:
Now I can see the problem after changing the parent folder name to `Matrix (1999)`. It use `movie.nfo` for the collection title which I think is reasonable.

屏幕截图 2023-10-06 163902

Problem is that it also uses `movie.nfo` for details of different version. Same title for 1080p and 2160p, `Matrix movie.nfo`.

屏幕截图 2023-10-06 164026
屏幕截图 2023-10-06 164034

After removing `movie.nfo`, it use `Matrix (1999) - 1080p.nfo` for the collection title which I think it choose the first nfo file ordered by filename.

屏幕截图 2023-10-06 164351

But it still has the problem, it also has the same title for 1080p and 2160p, `Matrix 1080p`.

屏幕截图 2023-10-06 164431
屏幕截图 2023-10-06 164441

This problem may be beyond this PR to fix.

So a better way to fix the issue is to check whether `<filename>.nfo` is a valid xml file before add it to the candidate nfo file list. I will try to fix this issue in this way later.

Shadowghost:
This problem is know, so it's working as expected even if we have multiple versions. I guess we're fine now. Thanks for testing!

leovan:
> This problem is know, so it's working as expected even if we have multiple versions. I guess we're fine now. Thanks for testing!

You mean current simple solution of this PR for https://github.com/jellyfin/jellyfin/issues/1558 is OK?

Bond-009 (approved):

mikewesten:
@leovan
This caused several issues with multiple versioned movies.
It's now very broken compared to how it worked before.
-The initial scan after the version update created new "Date Added" fields" instead of copying from the pre-existing .nfo file that was filenamed after the first movie versions filename.
-The first movie version in the dropdown menu now has "Inherit" content type for metadata edit.
-Movie art cannot be searched and download in the desktop GUI anymore, unless you use the dropdown menu to select 2nd movie version before doing so.
-Metadata edits cannot be done in the desktop GUI unless we select the 2nd movie version in the dropdown menu.
-Android TV app is now confused and won't list movie versions in the correct order or playback the correct version anymore, in many cases.

This code introduced lots of problems for me.

Issue #5493, opened 2021-03-11
Manually entered Actor: Spinning Circle when clicking an actor's name


mr-simba:
**Describe the bug**
Jellyfin 10.7.0 in a docker file on a qnap nas.

**On a movie detail page:**
1) If the actor information comes from a metadata-service like TheMovieDB or The Open Movie Database:
All works as expected: When I click on an actor name, a page loads with the actor information and all his films in my database.
2) If I enter the actor manually to a movie:
Bug: When I click on a manually entered actor name, all I see is a black page with a spinning circle with different colors. 

**System (please complete the following information):**
 - OS: Linux 
 - Virtualization: Docker
 - Clients: Browser, Android
 - Jellyfin Version: 10.7.0

mod242:
I have the exact same issue. When adding new Movies with new Actors via NFO or adding Actors manually, they somehow aren't really added to Jellyfin. When clicking on the actor the spinning circle appears and the following is logged:

`Jellyfin.Server.Middleware.ExceptionMiddleware: Error processing request. URL "GET" "/Users/6ad9a332618f4ebe99cd77c3dfcd1534/Items/7cadcb1455ddb35f120872f5acab977d".
System.NullReferenceException: Object reference not set to an instance of an object.
   at Emby.Server.Implementations.Dto.DtoService.GetBaseItemDtoInternal(BaseItem item, DtoOptions options, User user, BaseItem owner)
   at Emby.Server.Implementations.Dto.DtoService.GetBaseItemDto(BaseItem item, DtoOptions options, User user, BaseItem owner)
   at Jellyfin.Api.Controllers.UserLibraryController.GetItem(Guid userId, Guid itemId)
   at lambda_method874(Closure , Object )
   at Microsoft.AspNetCore.Mvc.Infrastructure.ActionMethodExecutor.AwaitableObjectResultExecutor.Execute(IActionResultTypeMapper mapper, ObjectMethodExecutor executor, Object controller, Object[] arguments)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.<InvokeActionMethodAsync>g__Awaited|12_0(ControllerActionInvoker invoker, ValueTask`1 actionResultValueTask)
[...]

PR #10345, opened 2023-10-05
Don't create non existent persons in LibraryManager.GetPerson

### LINKED ISSUE #3901: LibraryManager seems to create new object with Id and other features when requesting non-existant object
When requesting GetPerson from the library manager providing it a string of a non existing persons name it will return an object a Person object.There is a Guid set in the Id property which is either deterministically created or a real object is created in the database upon first query. This is because a subsequent query for the same non existing user will result in the same Id

 - OS: Open Media Vault
 - Virtualization: docker
 - Clients: N/A
 - Browser: n/a
 - Jellyfin Version: 10.6.0
 - Playback: n/a
 - Installed Plugins: n/a
 - Reverse Proxy: n/a
 - Base URL: n/a
 - Networking: n/a
 - Storage: local

**To Reproduce**
```
var user_two = libraryManager.GetPerson("CGP asdasd");
//user_two.Id = 6b9beb03-c00e-5a25-495f-4923e211cfb2
var user_three = libraryManager.GetPerson("CGP asdasd");
//user_three.Id = 6b9beb03-c00e-5a25-495f-4923e211cfb2
var foo = libraryManager.GetItemById(user_three.Id);
// Same object as above
```

**Expected behavior**
It would be expected that either the item would be empty of an Id or there would be some indication it doesn't exist like an error. Certain fields could be used to determine if it doesn't exist such as DateLastRefreshed and DateLastSaved being from year epoch, and PresentationUniqueKey and PrimaryImagePath being null.

Bond-009:
return null instead.

GetStudio, GetGenre, GetMusicGenre, GetYear, GetArtist still create a new one when the requested one doesn't exist

Fixes #3901

### DIFF
--- a/Emby.Server.Implementations/Library/LibraryManager.cs
+++ b/Emby.Server.Implementations/Library/LibraryManager.cs
@@ -46,7 +46,6 @@ using MediaBrowser.Model.IO;
 using MediaBrowser.Model.Library;
 using MediaBrowser.Model.Querying;
 using MediaBrowser.Model.Tasks;
-using Microsoft.Extensions.Caching.Memory;
 using Microsoft.Extensions.Logging;
 using Episode = MediaBrowser.Controller.Entities.TV.Episode;
 using EpisodeInfo = Emby.Naming.TV.EpisodeInfo;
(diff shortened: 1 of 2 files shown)


cvium:
Did you test the 2nd scenario here https://github.com/jellyfin/jellyfin/issues/5493 ?

Bond-009:
Not yet

Bond-009:
Works fine

cvium (approved):

Shadowghost (approved):

crobibero (approved):

PR #5992, opened 2021-05-06
Make switch in nfo parser smaller

daullmer:
**Changes**

- Use custom methods to read strings, ints, floats, bools and dates from nfo
- Move subtree reader methods to separate file

The best way to review to PR would probably be to go through each commit which may make it easier to understand.

**Issues**
Fix multiple language tags resulting in only the last country to be added to the database

PR #10352, opened 2023-10-06
XML Parsing Cleanup

barronpm:
**Changes**
- De-duplicates some code
- Adds extension methods for common operations
- Cleans up some formatting

Supercedes #5992


MediaBrowser.Controller/Extensions/XmlReaderExtensions.cs
```diff
@@ -0,0 +1,217 @@
...
+        if (DateTime.TryParse(
+                text,
+                CultureInfo.InvariantCulture,
+                DateTimeStyles.AssumeUniversal | DateTimeStyles.AdjustToUniversal,
+                out value))
+        {
+            return true;
+        }
+
+        logger.LogWarning("Invalid date: {Date}", text);
```
cvium, line 64: Feels a bit weird to put a logger in the args

barronpm: I agree, I just added it to keep existing behavior. If you're fine with not logging on failure anymore I'll remove it.

MediaBrowser.Controller/Extensions/XmlReaderExtensions.cs
```diff
@@ -0,0 +1,217 @@
...
+
+                    break;
+                case "type":
+                case "Type":
+                    Enum.TryParse(subtree.ReadElementContentAsString(), true, out type);
+                    break;
+                case "order":
+                case "sortorder":
+                case "SortOrder":
+                    if (int.TryParse(
```
cvium, line 143: shouldn't it be using the TryReadInt here?

MediaBrowser.Controller/Extensions/XmlReaderExtensions.cs
```diff
@@ -0,0 +1,217 @@
...
+    /// <param name="reader">The <see cref="XmlReader"/>.</param>
+    /// <returns>IEnumerable{System.String}.</returns>
+    public static IEnumerable<string> GetStringArray(this XmlReader reader)
+    {
+        ArgumentNullException.ThrowIfNull(reader);
+        var value = reader.ReadElementContentAsString();
+
+        // Only split by comma if there is no pipe in the string
+        // We have to be careful to not split names like Matthew, Jr.
+        var separator = !value.Contains('|', StringComparison.Ordinal)
```
cvium, line 180: The rest of this method seems a bit inefficient

MediaBrowser.Controller/Extensions/XmlReaderExtensions.cs
barronpm: Yes, I've updated this method to use the other extension methods

MediaBrowser.Controller/Extensions/XmlReaderExtensions.cs
cvium: I don't think we've gotten that many reports about this, so it seems a little useless

MediaBrowser.Controller/Extensions/XmlReaderExtensions.cs
```diff
@@ -0,0 +1,203 @@
...
+        }
+
+        if (string.IsNullOrWhiteSpace(name))
+        {
+            return null;
+        }
+
+        return new PersonInfo
+        {
+            Name = name.Trim(),
```
cvium, line 160: It's already trimmed

MediaBrowser.Controller/Extensions/XmlReaderExtensions.cs
barronpm: Yeah, this is an ideal case for spans, but we don't have a `Split` method for multiple needles.
One solution is to do a pure fallback of `|` -> `;` -> `,` for separators, which is technically a breaking change, but I doubt there's a ton of files out there using both.

Alternatively, we would have to make another extension method, but imo that's a big enough change to warrant a separate PR.

cvium (approved):

Shadowghost (approved):

Bond-009 (approved):

PR #10354, opened 2023-10-07
Adding endpoints to support excluding items from "Continue Watching"

westonhowe98:
**Changes**
Resolves [this](https://features.jellyfin.org/posts/517/add-an-option-to-remove-an-item-from-continue-watching) feature request. (The backend piece at least)
Created from [this](https://github.com/jellyfin/jellyfin-meta/discussions/50#discussioncomment-7204164) discussion.

Adds 2 new end points for adding and removing a flag from an item. If the flag is present on the item then the item will not be considered for "Continue Watching"

The flag is removed anytime the playback progress of the item is updated so the item can start showing in "Continue Watching" again.

Handing for adding the flag column to the UserDatas table is added to the application startup to ensure the update process will include it.


westonhowe98:
Closing pull request. will reopen with PR on branch.

PR #10356, opened 2023-10-07
Change initialization of root folders to app startup

Dangles91:
**Changes**
- Remove lazy creation of library root folders
- Remove double check locking on library roots
- Move library root references to static properties on LibraryRoot class
- Reduce dependency on LibraryManager


Dangles91:
Having looked at some plugins, this would impact plugin developers.

Which leads into a larger discussion about separating the internal application structure from the plugin API.

Shadowghost:
I can't comment on how this should be handled since my understanding is limited. Maybe @cvium or @Bond-009 can comment on if this is doable or should be done at all.
Bigger architectural changes impacting clients and the majority of plugins should be proposed as discussion in [jellyfin-meta](https://github.com/jellyfin/jellyfin-meta)

gaurovgiri (approved):
LGTM

JPVenson:
Hey @Dangles91 
Please open a discussion on our meta repro. I will keep this PR open for the time being, i would however recommend you converting it into a draft until your discussion topic is resolved.

PR #10358, opened 2023-10-07
Exclude from continue watching

westonhowe98:
Changes
Resolves [this](https://features.jellyfin.org/posts/517/add-an-option-to-remove-an-item-from-continue-watching) feature request. (The backend piece at least)
Created from [this](https://github.com/jellyfin/jellyfin-meta/discussions/50#discussioncomment-7204164) discussion.

Adds 2 new end points for adding and removing a flag from an item. If the flag is present on the item then the item will not be considered for "Continue Watching"

The flag is removed anytime the playback progress of the item is updated so the item can start showing in "Continue Watching" again.

Handing for adding the flag column to the UserDatas table is added to the application startup to ensure the update process will include it.

Added new flag to DTOs and included in UserData reads.


thornbill:
This is blocked until the discussion for the _planned_ implementation in the meta repo is resolved. Making code changes before that happens is not the correct process.

westonhowe98:
Discussion topic closed

thornbill:
Sorry if the process hasn't been clear. The Jellyfin team needs to be involved in the discussion to make sure the implementation aligns with best practices, etc. That is the main reason a discussion is needed first for changes of a larger scope or fundamental nature.

gaurovgiri (approved):
Looks good to me

nathangur:
> Sorry if the process hasn't been clear. The Jellyfin team needs to be involved in the discussion to make sure the implementation aligns with best practices, etc. That is the main reason a discussion is needed first for changes of a larger scope or fundamental nature.


Could we begin the discussion on the implementation of both continue watching and the next up button implementations? I don't want these PR's to get buried.

JPVenson:
As already said in the discussion i will close this PR as currently no desired.

Wortex17:
This has 888 upvotes and is considered "not desired"? I absolutely get that you'd prefer a general refactoring. At the same time this topic is open for 4 yrs and at some point you'll have to take some cuts instead of waiting for big reactors. 

Kinda bumming me out that it feels like so much me and some pals would like to contribute is basically blocked by refactoring plans and I wonder how much potential contributors in generals are being discouraged by these stale PRs/Tickets

JPVenson:
@Wortex17 just for clarification: This kind of implementation, at this current state is not desired.
You have to keep in mind that the core team needs to be able to fix bugs, adapt the code and even want to do it. And there is currently no commitment to do that. I did not say that the feature itself was not desired, just that we currently have no capacity to support it proper as we are still missing lots of other stuff to do.

As any big OpenSource project we cannot give ETA's for a plethora of reasons.

Wortex17:
I see that and I understand the maintenance concerns. I did not want to dunk on the fact you decline this specific PR. 
The issue is that since the discussion with the server team seems to be something that happened internally, outsiders have no clue as to what would be required to chip in, instead it basically boils down to "thanks, nice try, we'll take it from there" but at the same you understandably don't have time for it.

As without any guideline, scope description or mission statement, it seems expected that a contributor throws a refactoring PR for the whole thing at and "hopes it fits/meets" the mistery criteria which can be pretty daunting.

(Should we move these to the discussion thread instead?)
(Thx for the quick response btw)

JPVenson:
> As without any guideline, scope description or mission statement, it seems expected that a contributor throws a refactoring PR for the whole thing at and "hopes it fits/meets" the mistery criteria which can be pretty daunting.

Been there not long ago trust me.
But i take that feedback on me and put a new ToDo on my list to expand the contribution guidelines:
https://jellyfin.org/docs/general/contributing/#developing-code

In general, we try to communicate that everyone can just hop onto matrix/discord and just ask if a certain feature needs a discussion in our GH Meta repro. This is the safeguard that should help anyone not doing some stuff that gets rejected because of architecture/undesirable. 

The overall issue here is that the Discussion was created one day before the PR, which is just not enough time for core members to deliberate and give feedback and the whys and wheres can be read up on my discussion thread
https://github.com/jellyfin/jellyfin-meta/discussions/50#discussioncomment-7658509
If i take a topic into internal discussions, i will always give a short summery why the result is as it is.

pto199:
This feature has been requested for nearly half a decade.  Somebody implements it and its blocked because there wasn't a discussion with the core team.  6 more months pass and its still nowhere and all the core team says is its not a proper fix.  So I guess we wait 3 more years for that? 

Just implement this band-aid solution now and do the proper fix later.  I have to delete my users profiles as of now and re-create them just to remove shows from next up for them.  That's ridiculous.

joshuaboniface:
>Just implement this band-aid solution now and do the proper fix later.

This is how you get spaghetti code littered with bugs that are impossible to fix, and features that are half-baked and are not extensible. Once something like this is implemented and live to thousands of instances, it is not trivial to just change it later.

I have to delete my users profiles as of now and re-create them just to remove shows from next up for them. That's ridiculous.

So, the issue is basically that how Next Up works is not intuitive and not easy to fix. This is one way to do it, but not one we would prefer as it just adds more spaghetti to the codebase. It's being, from my understanding, automatically determined from the fact that a show has any number of watched episodes and any number of episodes due to be watched. Currently the only way to cleanly clear that is to mark the entire series as unwatched, which is obviously not ideal, but you can do that instead of deleting users.

This extra flag is indeed one possible solution, but we don't think the design is good. As mentioned, it's just adding another DB field that needs to be tracked and managed for every series, and plus I don't think this would work intuitively the way one would think (for instance, how would one clear this field?).

A few team members have a few ideas to improve this, at a more fundamental level, by cleaning up how the various list types work. That's obviously not a quick and simple solution, but our goal with Jellyfin is not to just take the quick and simple solution every time - that's why the codebase we got was an absolute disaster. We want to ensure features are well designed and well implemented so that the software can be maintained into the future. That takes time and effort.

I'll ask @jellyfin/server here if they think a quick solution is ideal, but I suspect the answer remains "no" as undoing the quick and dirty solution later will be a lot harder than implementing it in the first place.

darkbepo:
Just an outsider non-coder here reading this and thinking out loudly - could this be made into a plugin instead so people can decide for themselves whether they want it or not?

PR #10363, opened 2023-10-07
Build Windows installer as part of packaging CI job

DavidFair:
This is more a proposal / proof of concept. I'm happy to take any suggestions and ideas, hence it's marked as a draft:

**Changes**
Using all the existing bits we can easily build the Windows installer for existing versions.

In this ~commit~ PR we:
- Grab the latest version of JF Web (based on the latest unstable)
- Grab the latest version of JF_UX
- Install nsis
- Build an installer as a dedicated artifact

Why; because being able to bundle JF installer as part of the daily builds increases the chances of finding bugs / getting fixes out. It also means that people don't have to figure out how to install ffmpeg...etc. if they want to test an unstable fix.

Limitations
- The installer tags itself as the real version, this needs fixing in the upstream installer

- We don't upload the installer as part of the CI to repos.jellyfin, this is partially because we don't want users easily finding "a released version", which is actually unstable but the installer is incorrect.

- Some assumptions have been made about how JF web will name its files. Ideally we want a symlink which is repointed on repos.jellyfin which points to the latest version. For the moment (ab)use the date naming.

**Testing**
Run the docker build as follows from the source dir:
```
cd deployment && docker build .  -f Dockerfile.windows.amd64 -t win_build
cd .. && docker run -v $(pwd)/deployment/dist:/dist -v $(pwd):/jellyfin -e IS_UNSTABLE="yes" -e BUILD_ID="local" win_build
```

There will be an installer in `deployment/dist/jellyfin-local-windows-x64.exe`
Test on a Windows machine, I've got it working locally with the limitations described above:
image

I've also tested it with a forked pipeline to verify it plays nice with the CI (albeit with most of the packaging / testing steps deleted for my sanity) [here](https://dev.azure.com/DavidFair/Jellyfin%20Fork/_build/results?buildId=11&view=logs&j=b239e49c-c2c4-5033-bc88-01d0d0cdac65)

### DIFF
--- a/deployment/Dockerfile.windows.amd64
+++ b/deployment/Dockerfile.windows.amd64
@@ -10,8 +10,8 @@ ENV IS_DOCKER=YES
 
 # Prepare Debian build environment
 RUN apt-get update -yqq \
-  && apt-get install -yqq --no-install-recommends \
-    debhelper gnupg devscripts unzip \
+    && apt-get install -yqq --no-install-recommends \
+    debhelper gnupg devscripts unzip nsis \
     mmv libcurl4-openssl-dev libfontconfig1-dev \
     libfreetype6-dev libssl-dev libssl1.1 liblttng-ust0 zip
 
--- a/deployment/build.windows.amd64
+++ b/deployment/build.windows.amd64
@@ -6,12 +6,30 @@ set -o errexit
 set -o xtrace
 
 # Version variables
+
+# Figure out the latest version by sorting based on the date then version number
+# ideally we should just have a symlink called latest that points to the latest version...
+JELLYFIN_WEB_LATEST_VERSION=$(curl -q https://repo.jellyfin.org/releases/server/portable/versions/unstable/web/ | grep -oP 'href="\K[0-9]+\.[0-9]+' | sort -t. -rn -k1,1 -k2,2 | head -1)
+
+JELLYFIN_WEB_FILENAME="jellyfin-web_${JELLYFIN_WEB_LATEST_VERSION}_portable.tar.gz"
+JELLYFIN_WEB_URL="https://repo.jellyfin.org/releases/server/portable/versions/unstable/web/${JELLYFIN_WEB_LATEST_VERSION}/${JELLYFIN_WEB_FILENAME}"
+
+JELLYFIN_INSTALLER_VERSION=master
+JELLYFIN_INSTALLER_URL="https://github.com/jellyfin/jellyfin-server-windows"
+
+JELLYFIN_UX_VERSION=master
+JELLYFIN_UX_URL="https://github.com/jellyfin/jellyfin-ux"
+
 NSSM_VERSION="nssm-2.24-101-g897c7ad"
 NSSM_URL="http://files.evilt.win/nssm/${NSSM_VERSION}.zip"
+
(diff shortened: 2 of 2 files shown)


crobibero:
We already have a workflow that has the ability to publish an unstable version of the windows Jellyfin installer, and we are [moving away from using Azure Pipelines](https://github.com/jellyfin-sandbox).


Additionally I'm of the opinion that users that need the installer shouldn't be using an unstable version of Jellyfin.

DavidFair:
> We already have a workflow that has the ability to publish an unstable version of the windows Jellyfin installer, and we are [moving away from using Azure Pipelines](https://github.com/jellyfin-sandbox).

Looking at the repository it's been mothballed for 1+ year. I'm a strong believer of [perfect is the enemy of good](https://en.wikipedia.org/wiki/Perfect_is_the_enemy_of_good):

I'd be happy to start porting things from pipelines in my fork, would you accept a PR onto this org which switches say the unit testing to GH workflows based on the existing work?
I'm not keen to point it to staging as I know it will likely sit unused for a long period of time, since the available dev effort (which is greatly appreciated, I'm not diminishing the great job you all do) is limited, so it will probably rot over there.

> Additionally I'm of the opinion that users that need the installer shouldn't be using an unstable version of Jellyfin.

I respectfully disagree; I don't think we should make it easy to find for example to prevent people accidentally installing it. But if we make it completely clear that it's unsupported, I would argue that [Linus's Law](https://en.wikipedia.org/wiki/Linus%27s_law) would find more bugs than it would generate support tickets.
Currently the only way to remember to update FFMPEG / Web / Server together is three manual downloads and extractions. This is even more painful when DLL names change, so you have to tidy your portable instances.

However, again this is my opinion hence I opened it as a draft PR. Arguably, this can still land regardless of the pipelines aspect, since I can't see anything for Windows packaging in the staging repo. If there's no interest though, I'm happy to close it since it's ultimately a proof-of-concept.

DavidFair:
> We would love help with the migration to github actions- I was surprised to see it had almost been a year since the changes were last touched. I'm not completely sure of the state of the workflows.

I've opened #10392 to get that ball rolling at least

> Your notes on creating the installer but not really "publishing" it make sense to me. One thing to note is that the ffmpeg link points to the latest 5.x version and not the 6.x version that Jellyfin 10.9 will be expecting.

I've bumped the version to 6.x, do you want to hold off on this PR until workflows fully switch to GitHub or something else? Or is it worth un-drafting this. I'm happy to tweak and change this as needed in the interim.

crobibero:
I find Azure devops to be a pain to use so I personally would prefer new ci use Github actions, but since this isn't creating a new workflow file it's a good start

gaurovgiri (approved):
LGTM

joshuaboniface:
@DavidFair Just seeing this now. I see you've done most of the work to get this automated and ported, so the new packaging is all at https://github.com/jellyfin/jellyfin-packaging. Happy to help integrate it there!

joshuaboniface:
cc @anthonylavado too not sure if you saw this, or how much integration with the tray app might take.

Bond-009:
Closing this as packaging code has its own repository now: https://github.com/jellyfin/jellyfin-packaging

PR #10366, opened 2023-10-07
Fix a few resharper warnings and appy a few its suggestions

goremykin:
**Changes**
This PR removes some minor redundancy and applies some resharper's suggestions (mostly modern sugar).
Commits are quite descriptive and split by suggestion / warning type, so I recommend review it commit by commit.


Emby.Server.Implementations/Channels/ChannelManager.cs
```diff
@@ -1156,7 +1156,7 @@ private async Task<BaseItem> GetChannelItemEntityAsync(ChannelItemInfo info, ICh
 
                 if (info.People is not null && info.People.Count > 0)
                 {
-                    _libraryManager.UpdatePeople(item, info.People);
+                    await _libraryManager.UpdatePeopleAsync(item, info.People, cancellationToken).ConfigureAwait(false);
```
goremykin, line 1159: Btw, do we have a synchronization context somewhere?
I mean, do we really need `ConfigureAwait` everywhere?

Emby.Server.Implementations/HttpServer/WebSocketConnection.cs
```diff
@@ -12,7 +12,6 @@
 using MediaBrowser.Controller.Net.WebSocketMessages;
 using MediaBrowser.Controller.Net.WebSocketMessages.Outbound;
 using MediaBrowser.Model.Session;
-using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Logging;
 
 namespace Emby.Server.Implementations.HttpServer
```
goremykin, line 17: Are we planning switching to the file-scoped namespace everywhere?

Emby.Server.Implementations/ScheduledTasks/Tasks/ChapterImagesTask.cs
```diff
@@ -115,7 +115,8 @@ public async Task ExecuteAsync(IProgress<double> progress, CancellationToken can
             {
                 try
                 {
-                    previouslyFailedImages = File.ReadAllText(failHistoryPath)
+                    var failHistoryText = await File.ReadAllTextAsync(failHistoryPath, cancellationToken).ConfigureAwait(false);
+                    previouslyFailedImages = failHistoryText
```
Bond-009, line 119: Why in 2 parts?

Bond-009 (approved):
For the future I'd prefer it if you wouldn't make cosmetic only changes to the RSSDP project.
Fewer changes compared to upstream will make it easier to diff and hopefully eventually replace.

Emby.Server.Implementations/ScheduledTasks/Tasks/ChapterImagesTask.cs
goremykin: @Bond-009, it looks a bit cumbersome. Let's fix it if you don't think so
```
previouslyFailedImages = (await File.ReadAllTextAsync(failHistoryPath, cancellationToken).ConfigureAwait(false))
    .Split('|', StringSplitOptions.RemoveEmptyEntries)
    .ToList();
```

goremykin: @Bond-009, and could you check my questions above?

Emby.Server.Implementations/Channels/ChannelManager.cs
Bond-009: No, but we should keep using them for the stuff that gets published to nuget at least

Emby.Server.Implementations/HttpServer/WebSocketConnection.cs
Bond-009: Maybe, but I wouldn't go and change them as that would mess with git blame

Emby.Server.Implementations/ScheduledTasks/Tasks/ChapterImagesTask.cs
Bond-009: I didn't notice the split below, I'd move that to the same line

Emby.Server.Implementations/HttpServer/WebSocketConnection.cs
goremykin: @Bond-009, At least in jetbrains rider it doesn't affect git blame at all. But I am not sure about VS and git cli.

Shadowghost (approved):
Looks fine but keep in mind that merging this will likely break a lot of existing PRs

PR #10367, opened 2023-10-08
Add end points for excluding an episode from next up.

westonhowe98:
**Changes**
Added flag to to exclude an episode from Next Up section.
Modified next up query to use flag.
Added flag to UserDataDTO.
Added 2 new end points to set and remove the flag accordingly.

Came about from [this](https://github.com/jellyfin/jellyfin-meta/discussions/50) discussion.


thornbill:
This is blocked until the discussion for the _planned_ implementation in the meta repo is resolved. Making code changes before that happens is not the correct process.

westonhowe98:
discussion topic closed

gaurovgiri (approved):
Looks good to me

JPVenson:
Hey @westonhowe98 

I have discussed this PR and the related meta discussion with the team and while we like the idea, we see it as adding more code to an outdated system. While we dont want to discard the idea, the general consensus is to postpone the change after a complete refactor of the "next-up" system.

PR #10379, opened 2023-10-10
Empty Guids shouldn't make it into the refresh queue

Bond-009:
```
System.ArgumentException: Guid can't be empty (Parameter 'id')
   at Emby.Server.Implementations.Library.LibraryManager.GetItemById(Guid id) in /home/loma/dev/jellyfin/Emby.Server.Implementations/Library/LibraryManager.cs:line 1224
   at MediaBrowser.Providers.Manager.ProviderManager.StartProcessingRefreshQueue() in /home/loma/dev/jellyfin/MediaBrowser.Providers/Manager/ProviderManager.cs:line 983
```
This doesn't fix the cause, but at least they won't make it into the queue


Shadowghost (approved):
Maybe add a comment that this should usually not happen.

crobibero (approved):

Issue #10359, opened 2023-10-07
[Issue]: Weird reason for transcoding, then proceeds to use copy for codec


felix920506:
### Please describe your bug

I am playing back some CD Rips in FLAC on my JF instance using both Web via Chrome and JMP.

I noticed this very weird transcode reason "Video bitrate exceeds the limit" (As in screenshots below)

It then proceeds to transcode the source file to another FLAC file with the flag `-acodec copy` (As shown in the ffmpeg log)

Steps to reproduce:
1. Playback FLAC files that exceed the bandwidth limits on the client side (I think)
2. See error

Sometimes it transcodes to AAC so this probably won't appear 100% of the time, however I've seen this behavior in 10+ files so I decided to open a bug report.

### Jellyfin Version

10.8.z

### if other:

10.8.11

### Environment

```markdown
- OS: Ubuntu 23.04
- Linux Kernel: 6.2
- Virtualization: None
- Clients: JMP
- Browser: N/A
- FFmpeg Version: 6.0
- Playback Method: "Transcoding"
- Hardware Acceleration: QSV
- GPU Model: Intel ARC A750
- Plugins: AniDB
- Reverse Proxy: Caddy
- Base URL: /jellyfin
- Networking: Reverse proxy
- Storage: Mounted SMB
```

### FFmpeg logs

```shell
Attached below (256kFLAC.txt)
```

### Please attach any browser or client logs here

ffmpeg log:
[256kFLAC.txt](https://github.com/jellyfin/jellyfin/files/12838839/256kFLAC.txt)

### Please attach any screenshots here

Media Info:
<img width="882" alt="截圖 2023-10-07 下午4 12 25" src="https://github.com/jellyfin/jellyfin/assets/25688628/75ee7e61-e2ae-4860-a5f7-ad1b514cdf7a">

Plaback status:
<img width="301" alt="截圖 2023-10-07 下午1 53 04" src="https://github.com/jellyfin/jellyfin/assets/25688628/1cf96586-565d-4fe5-be89-f161fa86ebd7">

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

felix920506:
I tried to reproduce the issue again today. Setting the bitrate limit to 1mbps immediately reproduced the issue for me.

Bond-009:
@felix920506 10.8.x or master, bc I tested on master

felix920506:
I've only tried Release 10.8.11 to this point. I will test master if needed.

[...]

PR #10387, opened 2023-10-11
StreamingHelpers.GetOutputFileExtension should never return null

Bond-009:
Fixes regression from #10348


barronpm (approved):

Shadowghost (approved):

felix920506:
I have just tested this PR for #10359 per @Bond-009 's request.
It is exhibiting the exact same behavior as 10.8.11
ffmpeg log:
[masterbranch20231011.txt](https://github.com/jellyfin/jellyfin/files/12872860/masterbranch20231011.txt)

PR #10392, opened 2023-10-12
Add GH Workflow for CI Tests

DavidFair:
**Changes**
Ports some of the workflow from the .ci/azure-pipelines-test.yml file:

- Establish a common naming pattern for the various workflows, e.g. `ci-*` or `issues-*` based on what it interacts with
- Split up existing workflows to match said pattern, without changing any details
- Port the test pipeline whilst leaving the original in place (until the project is happy to swap) with some caveats:

I've dropped SonarCloud, it seems to be a left-over which hasn't been enabled in a while.
There's a TODO on where we want to upload code coverage, since there's nothing natively built into GitHub (AFAIK), so we generate coverage but don't go any further.


.github/workflows/ci-tests.yml
```diff
@@ -0,0 +1,60 @@
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-dotnet@v3
+        with:
+          dotnet-version: ${{ env.SDK_VERSION }}
+
+      - name: Run DotNet CLI Tests
+        run: >
+          find tests -type f -iname "*.csproj" -print0 | xargs -0 -I{} -t
```
cvium, line 36: Our Azure pipeline uses the `"tests/**/*Tests.csproj"` pattern.

.github/workflows/ci-tests.yml
```diff
@@ -0,0 +1,60 @@
...
+          dotnet test '{}' ${{ env.TEST_ARGS }}
+        if: matrix.os != 'windows-latest'
+
+      - name: Run DotNet CLI Tests
+        run: >
+          Get-ChildItem -Recurse -Filter *.csproj -Path tests |
```
cvium, line 42: With the pattern above we won't need separate tasks

DavidFair: I couldn't get this to pick up all the tests when I tested it locally, and there doesn't seem to be anything on the [docs about recursive](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-test#arguments)

Attempting to run it gives:
```
  dotnet test 'tests/**/*.csproj' --configuration Release --collect:"XPlat Code Coverage" --settings tests/coverletArgs.runsettings --verbosity minimal
  shell: /usr/bin/bash -e {0}
  env:
    SDK_VERSION: 7.0.x
    TEST_ARGS: --configuration Release --collect:"XPlat Code Coverage" --settings tests/coverletArgs.runsettings --verbosity minimal
  
    DOTNET_ROOT: /usr/share/dotnet
  
MSBUILD : error MSB1009: Project file does not exist.
Switch: tests/**/*.csproj
```

And it looks like it's not planned: https://github.com/dotnet/sdk/issues/30655

I suspect the Azure equivalent is effectively a wrapper round a glob and something similar to this. I'm happy to take suggestions what we could do instead as something cross-platform would be a big plus in my books too

.github/workflows/ci-tests.yml
DavidFair: This find looks only in `tests` (recursively for `*.csproj`), it's case insensitive (to unify Windows and POSIX), so the equivalent shell glob would be: `tests/**/*.csproj` 

but I'm happy to tweak it back to the original equivalent still using find?

`find tests -type f -name "*Tests.csproj" -print0`

.github/workflows/ci-tests.yml
cvium: Unfortunate. We only have test projects in the tests folder. We could just use `dotnet test Jellyfin.sln`

DavidFair: :+1: I'd be happy with that
Let me rejig this PR and I can add an extra check too that all of the `*Tests.csproj` are named in the `.sln` too, so we don't have any slip through the cracks...

.github/workflows/ci-tests.yml
```diff
@@ -0,0 +1,56 @@
...
+
+    runs-on: "${{ matrix.os }}"
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-dotnet@v3
+        with:
+          dotnet-version: ${{ env.SDK_VERSION }}
+
+      - name: Ensure all tests are included in .sln
```
cvium, line 28: I don't think this step is necessary. If they aren't in the sln, then a reviewer has failed

DavidFair: Dropped :+1:

cvium (approved):

Bond-009 (approved):

gaurovgiri (approved):
Seems good to me!

DavidFair:
Thanks for resolving those merge conflicts for me :+1:

PR #5951, opened 2021-05-01
DLNA fix : Make media visible and playable

BaronGreenback:
Some devices / software do not show media if they cannot play use container, stopping the possibility of it be transcoded.
VLC and Hisense TV's both do this - i'm sure there are others.

This change returns all media via DLNA with the extension .ts. The original file extension is passed as a parameter.

Once the request to stream has been received, this correct file extension is used for transcoding selection / playback.

Have tried without using a file extension - but the same issue occurs.

Suspect this is the cause of https://github.com/jellyfin/jellyfin/issues/2371.

PR #10404, opened 2023-10-14
DLNA fix : Make media visible and playable

Shadowghost:
Some devices / software do not show media if they cannot play use container, stopping the possibility of it being played by transcoding.
 
**Changes**
* This change returns all media via DLNA with the extension .ts. The original file extension is passed as a parameter.
* Once the request to stream has been received, this correct file extension is used for transcoding selection / playback.

**Issues**
* Possible fix for  jellyfin/jellyfin-plugin-dlna#8
* Subersedes jellyfin/jellyfin#5951

### DIFF
--- a/Emby.Server.Implementations/LiveTv/EmbyTV/EmbyTV.cs
+++ b/Emby.Server.Implementations/LiveTv/EmbyTV/EmbyTV.cs
@@ -1616,7 +1616,7 @@ namespace Emby.Server.Implementations.LiveTv.EmbyTV
 
         private IRecorder GetRecorder(MediaSourceInfo mediaSource)
         {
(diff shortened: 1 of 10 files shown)


barronpm:
I'm not a big fan of editing our endpoints specifically for DLNA. IMO, we should extract common logic out of the controllers/helpers, and have DLNA create its own routes with the DLNA-specific functionality. This reduces coupling between DLNA and the rest of the server and makes it easier to extract DLNA into a plugin later.

Shadowghost:
I'm totally fine with that. I only resurrected the PR since the original PR already had approvals.
Feel free to close if unwanted.

gaurovgiri (approved):
Looks good to me

servimo:
So we have to wait till DNA plugin will be released?

PR #10406, opened 2023-10-15
Fix playlist listings

### LINKED ISSUE #10272: [Issue]: [Regression] User-created playlists not accessible via CollectionView in v10.8.11
### Please describe your bug

Any playlists manually created by a user (as opposed to automatically discovered from e.g. a m3u file) are not returned by the ID if the `parentId` is set to the ID of the `CollectionView`. Afaik, only CollectionViews of type `Music` make use of this feature in the web client.  

Before Jellyfin 10.8.11, the `parentId` would be ignored when fetching `Playlist` items from the API, resulting in the request `<baseUrl>/Users/<userId>/Items?IncludeItemTypes=Playlist&Recursive=true&StartIndex=0&ParentId=<musicTypeCollectionViewId>` to return *any* created playlist (scanned, manually created within the `Music` CollectionView, as well as created in any other CollectionView like `Shows`).  
This of course wasn't ideal either (I don't want to see a playlist of show episodes in my music client), and therefore this was "fixed" in https://github.com/jellyfin/jellyfin/pull/9909 and backported to 10.8.z in https://github.com/jellyfin/jellyfin/pull/9916.

Now only auto-discovered playlists are returned when performing the request above; the user-created playlists can only be found in the `UserView` of `collectionType: playlist`. If this behavior was by design, it would be extremely annoying when building music clients for Jellyfin, and would prevent separating music playlists from video playlists.

**To reproduce:**
1. Set up Jellyfin 10.8.11 (via docker)
2. Import at least a music library that contains at least one m3u playlist file
3. Jellyfin should auto-detect the playlist, and it should show up in the web client under Home > Music > Playlists
4. Create a new playlist and add at least one song to it
5. The newly created playlist is not shown under Home > Music > Playlists, but does show up under Home > Playlists

### Jellyfin Version

Other

### if other:

10.8.11

### Environment

```markdown
- OS: TrueNAS-SCALE-22.12.3.3
- Linux Kernel: ?
- Virtualization: Docker (k3s)
- Clients: Finamp 0.6.17, jellyfin-web 10.8.11
- Browser: Microsoft Edge Dev 119.0.2109.1
- FFmpeg Version: n/a
- Playback Method: n/a
- Hardware Acceleration: n/a
- GPU Model: n/a
- Plugins: none
- Reverse Proxy: none
- Base URL: none
- Networking: n/a
- Storage: host path / ZFS dataset
```

### Jellyfin logs

```shell
issue doesn't show up in any logs
```

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

thornbill:
This reverts commit 98c6c34fbbcdb556873bdcdb577fbc55a296b650.

**Changes**
It doesn't seem like playlists currently have parent ids, so including parent ids in requests cause no playlists to be displayed. I think reverting this is the best course of action for 10.8.

**Issues**
Fixes #10272

### DIFF
--- a/Jellyfin.Api/Controllers/ItemsController.cs
+++ b/Jellyfin.Api/Controllers/ItemsController.cs
@@ -244,7 +244,8 @@ namespace Jellyfin.Api.Controllers
                 .AddAdditionalDtoOptions(enableImages, enableUserData, imageTypeLimit, enableImageTypes);
 
             if (includeItemTypes.Length == 1
-                && includeItemTypes[0] == BaseItemKind.BoxSet)
+                && (includeItemTypes[0] == BaseItemKind.Playlist
+                    || includeItemTypes[0] == BaseItemKind.BoxSet))
             {
                 parentId = null;
             }


Shadowghost (approved):

cvium (approved):

gaurovgiri (approved):
LGTM

PR #10410, opened 2023-10-15
add EnableSubtitleManagement permission

Ch1nkara:
Add new feature: ability for a user to "Edit subtitles" without administrator rights

**Changes**
Create a new Jellyfin.Data/Enums/PermissionKind: EnableSubtitleManagement
Add that permission to the UserPolicy

Related PR
 - Front-end: [jellyfin-web](https://github.com/jellyfin/jellyfin-web/pull/4883)

TODO
 - Fix [permissions issue](https://github.com/jellyfin/jellyfin-plugin-opensubtitles/issues/131#issuecomment-1677299539) in opensubtitle plugin

**Issues**
https://features.jellyfin.org/posts/185/non-admin-users-download-subtitles

### DIFF
--- a/Jellyfin.Api/Constants/Policies.cs
+++ b/Jellyfin.Api/Constants/Policies.cs
@@ -84,4 +84,9 @@ public static class Policies
     /// Policy name for managing LiveTV.
     /// </summary>
     public const string LiveTvManagement = "LiveTvManagement";
+
+    /// <summary>
+    /// Policy name for accessing subtitles management.
+    /// </summary>
+    public const string SubtitleManagement = "SubtitleManagement";
 }
--- a/Jellyfin.Api/Controllers/SubtitleController.cs
+++ b/Jellyfin.Api/Controllers/SubtitleController.cs
@@ -115,7 +115,7 @@ public class SubtitleController : BaseJellyfinApiController
     /// <response code="200">Subtitles retrieved.</response>
     /// <returns>An array of <see cref="RemoteSubtitleInfo"/>.</returns>
     [HttpGet("Items/{itemId}/RemoteSearch/Subtitles/{language}")]
-    [Authorize]
+    [Authorize(Policy = Policies.SubtitleManagement)]
     [ProducesResponseType(StatusCodes.Status200OK)]
     public async Task<ActionResult<IEnumerable<RemoteSubtitleInfo>>> SearchRemoteSubtitles(
         [FromRoute, Required] Guid itemId,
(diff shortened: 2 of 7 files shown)


cvium:
The endpoint used to download subtitles does not require admin: https://github.com/jellyfin/jellyfin/blob/master/Jellyfin.Api/Controllers/SubtitleController.cs#L138

I think it makes sense to restrict it. Maybe. But right now your PR is incomplete. The user policy should be set as a req on that endpoint.

Ch1nkara:
I think you either misunderstood my PR or I misunderstand your message. 

The fact that the subtitle controller does not check for admin rights is fine in this case. 
Currently, a user with no admin rights can't see the "edit subtitle" menu entry (in a movie view). The goal of that PR is to remedy to that. 
An admin can choose to let a user add new subtitles to a movie (by using the opensubtile plugin for example).

Ideally, the subtitle controller should also restrict the rights to download, but the goal of that PR is to give more rights to non-admin users, not restrict them

cvium:
You've misunderstood. The server is the authority on permissions. It doesn't make sense to make arbitrary restrictions in web if the server doesn't have the same restrictions. Anyone can edit the source in browser dev tools.

The subtitlecontroller must enforce the new user policy.

Ch1nkara:
Understood, thanks, I'll see to it and update the PR

Ch1nkara:
I think now the PR is complete.
Just to make sure my intent is clear: currently only an administrator had the option to "edit subtitles"
admin view
admin view

user view
user view

I want to give normal user (without admin rights) the ability to "edit subtitles". I now see two different ways to do that:
1) This PR way, by adding a permission EnableSubtitleManagement (thus letting the admin choose who can "edit subtitles").
2) By only contributing to jellyfin-web and let every user have the "edit subtitle" menu. Since there is currently no restriction server-side, there is no need to make changes to jellyfin in order to let normal users "edit subtitles".

I honestly don't really mind which way you guys prefer, as long as there is a way for normal users to download subtitles.

I'm open to discussion if you see other ways to achieve this requested feature.

nielsvanvelzen:
I believe the server not checking for admin permissions right now is a bug. To fix it in a way that allows users to upload subtitles means adding the new permission for both the backend and frontend.

gaurovgiri (approved):
Seems good to me!

cvium (approved):
Seems good now

Bond-009 (approved):

PR #10415, opened 2023-10-16
Add function to fix cast with localhost server

### LINKED ISSUE #10098: [Issue]: Function required to allow webclient to play on chromecast using localhost server url

Sky-High:
Adds a function to supply the webclient with a unique server LAN address after a localhost server API call. This enables the webclient to play to a chromecast receiver in such a case.

**Changes**
- Determine unique local server address in Jellyfin.Networking/Manager/NetworkManager.cs
- Pass this to the webclient with the GetPublicSystemInfo API call in Emby.Server.Implementations/ApplicationHost.cs

**Explanation**
This PR originates from the jellyfin-web issue with chromecast: https://github.com/jellyfin/jellyfin-web/issues/4745

While starting jellyfin-server it finds the bind addresses to listen to. And when starting jellyfin-web, it is able to find jellyfin-server alright. But when both jellyfin-server and jellyfin-web run on the same machine, they by default connect on url http://localhost:8096.

This works fine for the communication between jellyfin-web and jellyfin-server, but as a result this localhost url is also submitted to the cast receiver as url to connect to jellyfin-server. And for the cast receiver, localhost is not a valid address for jellyfin-server (in fact, with this address the cast receiver would try to find jellyfin-server internally ;-) )

So in this specific case of jellyfin-server and jellyfin-web communicating via localhost, the url to be passed to the cast receiver needs to be the 'real' LAN address of jellyfin-server.

As jellyfin-web is not able to establish this LAN address, jellyfin-server should deliver that via the API to jellyfin-client. So this issue requires changes in both jellyfin-server and jellyfin-web.

**Issues**
Fixes #10098

**Remark**
This is a resubmit of PR #10101 which was closed while the concerning src files were heavily changed and improved recently.

### DIFF
--- a/Jellyfin.Networking/Manager/NetworkManager.cs
+++ b/Jellyfin.Networking/Manager/NetworkManager.cs
@@ -131,6 +131,11 @@ namespace Jellyfin.Networking.Manager
         public bool TrustAllIPv6Interfaces { get; private set; }
 
         /// <summary>
+        /// Gets the IP LAN address of Jellyfin server usable for a Google cast receiver and such.
+        /// </summary>
+        public static string ServerLanAddress { get; private set; } = string.Empty;
+
+        /// <summary>
         /// Gets the Published server override list.
         /// </summary>
         public IReadOnlyList<PublishedServerUriOverride> PublishedServerUrls => _publishedServerUrls;
@@ -631,6 +636,8 @@ namespace Jellyfin.Networking.Manager
             InitializeOverrides(config);
 
             PrintNetworkInformation(config, false);
+
+            InitializeServerLanAddress(config);
         }
 
         /// <summary>
@@ -1119,6 +1126,82 @@ namespace Jellyfin.Networking.Manager
                 _logger.Log(logLevel, "Remote IP filter is {0}", config.IsRemoteIPFilterBlacklist ? "Blocklist" : "Allowlist");
                 _logger.Log(logLevel, "Filter list: {0}", _remoteAddressFilter.Select(s => s.Prefix + "/" + s.PrefixLength));
             }
+        }
+
+        /// <summary>
+        /// Initializes the LAN IP address that the server advertises for cast receivers.
+        /// A LAN device like a Chromecast receiver needs an unique LAN address of the Jellyfin server.
+        /// For IPv4, a Private network address (like 192.168.0.0/16 )serves this purpose, and for IPv6
+        /// the Unique Local Address (ULA, fd00::/8) does. An IPv6 global address (2000::/3) also works
+        /// as unique server address but is ranked as less desirable than a local address.
+        /// See e.g. https://en.wikipedia.org/wiki/Private_network
+        ///
+        /// If multiple addresses are available:
+        /// - prefer Unique Local Addresses (ULA) or Site Local Addresses over other (global) addresses.
+        /// - all things equal, prefer IPv4 over IPv6 as the former in practice appears to be more stable.
+        /// </summary>
+        private void InitializeServerLanAddress(NetworkConfiguration config)
+        {
+            lock (_initLock)
+            {
+                int rank = 0;   // ranks most preferable ip address
+                IPAddress ipSrv = IPAddress.None;
+                NetworkInterface[] adapters = NetworkInterface.GetAllNetworkInterfaces();
+
+                foreach (var adapter in adapters)
+                {
+                    if (adapter.NetworkInterfaceType == NetworkInterfaceType.Loopback)
+                    {
+                        continue;
+                    }
+
+                    // check if the network interface serves as IPv4 gateway
+                    bool haveIP4GateWay = false;
+                    foreach (var address in adapter.GetIPProperties().GatewayAddresses)
+                    {
+                        haveIP4GateWay = haveIP4GateWay || address.Address.AddressFamily == AddressFamily.InterNetwork;
+                    }
+
+                    // if the network interface is used as gateway, and has an acceptable IP address for this server,
+                    // then this address is assumed to be accessible by other devices on the LAN.
+                    foreach (var inf in _interfaces)
+                    {
+                        IPAddress ip = inf.Address;
+
+                        if (!inf.Name.Equals(adapter.Name, StringComparison.OrdinalIgnoreCase) ||
+                            ip.Equals(IPAddress.Loopback) || ip.Equals(IPAddress.IPv6Loopback))
+                        {
+                            continue;
+                        }
+
+                        if (ip.AddressFamily == AddressFamily.InterNetworkV6)
+                        {
+                            if (IsIPv6Enabled)
+                            {
+                                if (rank < 2 && (ip.IsIPv6SiteLocal || ip.IsIPv6UniqueLocal))
+                                {
+                                    rank = 2;
+                                    ipSrv = ip;
+                                }
+                                else if (rank < 1)
+                                {
+                                    rank = 1;
+                                    ipSrv = ip;
+                                }
+                            }
+                        }
+                        else if (haveIP4GateWay && IsIPv4Enabled && rank < 3)
+                        {
+                            rank = 3;
+                            ipSrv = ip;
+                        }
+                    }
+
+                    ServerLanAddress = ipSrv.ToString();
+                }
+            }
+
+            _logger.LogInformation("ServerLanAddress: {0} ", ServerLanAddress);
         }
     }
 }
--- a/Emby.Server.Implementations/ApplicationHost.cs
+++ b/Emby.Server.Implementations/ApplicationHost.cs
@@ -918,7 +918,14 @@ namespace Emby.Server.Implementations
                 return GetLocalApiUrl(request.Host.Host, request.Scheme, requestPort);
             }
 
-            return GetSmartApiUrl(request.HttpContext.Connection.RemoteIpAddress ?? IPAddress.Loopback);
+            // Gets the server URL that can also be used by applications not on the server, for instance
+            // a cast receiver device. In case the request has a localhost url, LocalAddress will be set to a
+            // server Private network address (IPv4) or Unique Local Address (ULA, IPv6) which is valid for all
+            // devices on the LAN.
+            IPAddress localAddress = request.HttpContext.Connection.RemoteIpAddress ?? IPAddress.Loopback;
+            return NetworkConstants.IPv4RFC5735Loopback.Contains(localAddress) || NetworkConstants.IPv6RFC4291Loopback.Contains(localAddress)
+                ? GetSmartApiUrl(NetworkManager.ServerLanAddress)
+                : GetSmartApiUrl(localAddress);
         }
 
         /// <inheritdoc/>


Emby.Server.Implementations/SystemManager.cs
```diff
@@ -72,16 +76,32 @@ public SystemInfo GetSystemInfo(HttpRequest request)
...
+    /// devices on the LAN.
+    /// </summary>
+    /// <param name="request">Where this request originated.</param>
+    /// <returns>PublicSystemInfo.</returns>
     public PublicSystemInfo GetPublicSystemInfo(HttpRequest request)
     {
+        string host = request.Host.ToString();
+        string localAddress = _applicationHost.GetSmartApiUrl(request);
+
+        if (host.StartsWith("localhost:", StringComparison.OrdinalIgnoreCase) || host.StartsWith("[::1]", StringComparison.OrdinalIgnoreCase) || host.StartsWith("127.0", StringComparison.OrdinalIgnoreCase))
```
Shadowghost, line 93: We should have Subnet specifications for those networks inside the network extensions which you can check against, this should be more reliable.

Sky-High: Does this look better?

gaurovgiri (approved):
Seems good to me!

Emby.Server.Implementations/SystemManager.cs
```diff
@@ -72,16 +74,34 @@ public SystemInfo GetSystemInfo(HttpRequest request)
...
+    /// server Private network address (IPv4) or Unique Local Address (ULA, IPv6) which is valid for all
+    /// devices on the LAN.
+    /// </summary>
+    /// <param name="request">Where this request originated.</param>
+    /// <returns>PublicSystemInfo.</returns>
     public PublicSystemInfo GetPublicSystemInfo(HttpRequest request)
     {
+        string ip4LoopBk = Network.IPv4RFC5735Loopback.Prefix.ToString();
+        string ip6LoopBk = "[" + Network.IPv6RFC4291Loopback.Prefix.ToString() + "]";
+        string host = request.Host.Host;
```
cvium, line 90: Can't really trust this as it's a header. If possible, it's better to use `request.HttpContext.Connection.RemoteIpAddress`

Emby.Server.Implementations/SystemManager.cs
```diff
@@ -72,16 +74,34 @@ public SystemInfo GetSystemInfo(HttpRequest request)
...
+        string localAddress = _applicationHost.GetSmartApiUrl(request);
+
+        if (host.Equals("localhost", StringComparison.OrdinalIgnoreCase) || host.StartsWith(ip4LoopBk, StringComparison.OrdinalIgnoreCase) || host.StartsWith(ip6LoopBk, StringComparison.OrdinalIgnoreCase))
```
cvium, line 93: Is our extension method not sufficient? https://github.com/jellyfin/jellyfin/blob/20fd05b05081ad387e94128b4f26d907808c8f0c/MediaBrowser.Common/Extensions/HttpContextExtensions.cs#L16

Emby.Server.Implementations/SystemManager.cs
```diff
@@ -72,16 +74,34 @@ public SystemInfo GetSystemInfo(HttpRequest request)
...
+    /// devices on the LAN.
+    /// </summary>
+    /// <param name="request">Where this request originated.</param>
+    /// <returns>PublicSystemInfo.</returns>
     public PublicSystemInfo GetPublicSystemInfo(HttpRequest request)
     {
+        string ip4LoopBk = Network.IPv4RFC5735Loopback.Prefix.ToString();
+        string ip6LoopBk = "[" + Network.IPv6RFC4291Loopback.Prefix.ToString() + "]";
+        string host = request.Host.Host;
+        string localAddress = _applicationHost.GetSmartApiUrl(request);
```
cvium, line 91: Should be if-else. It's a waste to compute it and then discard it

Jellyfin.Networking/Manager/NetworkManager.cs
```diff
@@ -1123,5 +1130,79 @@ private void PrintNetworkInformation(NetworkConfiguration config, bool debug = t
...
+        /// Initializes the LAN IP address that the server advertises for cast receivers.
+        /// A LAN device like a Chromecast receiver needs an unique LAN address of the Jellyfin server.
+        /// For IPv4, a Private network address (like 192.168.0.0/16 )serves this purpose, and for IPv6
+        /// the Unique Local Address (ULA, fd00::/8) does. An IPv6 global address (2000::/3) also works
+        /// as unique server address but is ranked as less desirable than a local address.
+        /// See e.g. https://en.wikipedia.org/wiki/Private_network
+        ///
+        /// If multiple addresses are available:
+        /// - prefer Unique Local Addresses (ULA) or Site Local Addresses over other (global) addresses.
+        /// - all things equal, prefer IPv6 over IPv4 as the former is more state-of-the-art.
```
cvium, line 1144: IPv6 doesn't always work and I have seen reports that CC doesn't play well with it, so we should honestly prefer IPv4 for this

Jellyfin.Networking/Manager/NetworkManager.cs
```diff
@@ -1123,5 +1130,79 @@ private void PrintNetworkInformation(NetworkConfiguration config, bool debug = t
...
+                            }
+                            else if (haveIP4GateWay && IsIPv4Enabled && rank < 2)
+                            {
+                                rank = 2;
+                                ipSrv = ip;
+                            }
+                        }
+                    }
+
+                    ServerLanAddress = ipSrv.ToString();
```
cvium, line 1201: The server can listen on multiple NICs. Assigning a single IP as _the_ LAN address will work for most setups but might not work correctly when there's multiple LAN subnets involved. Shouldn't we take the caller IP into account?

cvium (changes requested):

Jellyfin.Networking/Manager/NetworkManager.cs
Sky-High: Ok, this also worked fine on my system but I don't have much experience with others. I changed the preferred ranking.

Jellyfin.Networking/Manager/NetworkManager.cs
Sky-High: Yes that is a tough one.

My main goal was to add functionality so that Chromecast works out of the box for simple systems. With this PR one can for instance just download https://repo.jellyfin.org/releases/server/windows/stable/installer/jellyfin_xxx_windows-x64.exe and have Jellyfin server and client running with Chromecast on one single PC without any additional configuration (apart from the media folders of course.) From my own experience this helps a lot in getting up to speed with Jellyfin for first time/non-expert users.

For more extended systems like multiple NIC's it might be possible in some cases to autoconnect chromecast receivers without additional configuration, but the solution probably will be more complicated: I guess that after the Jellyfin webclient connects (using the Cast button) to a cast device, it should use the friendlyName returned by the onSessionConnected event to identify the device. The IP address (subnet) of the cast device has to be found by mDNS discovery, which a javascript in a browser is not able to execute. Jellyfin server can do so, however, being a standard application, so Jellyfin-web should probably submit the friendlyName to the server in an API call with a request to resolve its IP address (or actually: the serveraddres on its subnet).

Apart from the coding complexity I am not sure if this approach works in all cases without configuration, and more advanced systems might require explicit configuration anyhow, including assigning of chromecast devices.

So, if there is a good use-case for autoconnect to chromecast I could investigate this in a future PR.

Note however that this PR does add autoconnect for "simple" system, and does *not* make the situation worse for more advanced systems: in the latter case the autoconnect might not work *with* the PR, but neither did if *without* the PR ;-)

I would be quite interested in your ideas and suggestions on this!

Jellyfin.Networking/Manager/NetworkManager.cs
```diff
@@ -1123,5 +1130,79 @@ private void PrintNetworkInformation(NetworkConfiguration config, bool debug = t
...
+                        haveIP4GateWay |= address.Address.AddressFamily == AddressFamily.InterNetwork;
+                    }
+
+                    // if the network interface is used as gateway, and has an acceptable IP address for this server,
+                    // then this address is assumed to be accessible by other devices on the LAN.
+                    foreach (var inf in _interfaces)
+                    {
+                        IPAddress ip = inf.Address;
+
+                        if (inf.Name.Equals(adapter.Name, StringComparison.OrdinalIgnoreCase) &&
```
cvium, line 1174: I would invert this `if`. The indentation gets pretty deep.

Jellyfin.Networking/Manager/NetworkManager.cs
```diff
@@ -1123,5 +1130,79 @@ private void PrintNetworkInformation(NetworkConfiguration config, bool debug = t
...
+                    if (adapter.NetworkInterfaceType == NetworkInterfaceType.Loopback)
+                    {
+                        continue;
+                    }
+
+                    // check if the network interface serves as IPv4 gateway
+                    bool haveIP4GateWay = false;
+                    foreach (var address in adapter.GetIPProperties().GatewayAddresses)
+                    {
+                        haveIP4GateWay |= address.Address.AddressFamily == AddressFamily.InterNetwork;
```
cvium, line 1165: I know that this technically works, but `|=` is a bit-wise operation. It's always seemed a bit hacky to me.
```suggestion
                        haveIP4GateWay = haveIP4GateWay || address.Address.AddressFamily == AddressFamily.InterNetwork;
```

Emby.Server.Implementations/SystemManager.cs
```diff
@@ -72,16 +74,30 @@ public SystemInfo GetSystemInfo(HttpRequest request)
...
+    /// </summary>
+    /// <param name="request">Where this request originated.</param>
+    /// <returns>PublicSystemInfo.</returns>
     public PublicSystemInfo GetPublicSystemInfo(HttpRequest request)
     {
+        IPAddress host = request.HttpContext.Connection.RemoteIpAddress ?? IPAddress.None;
+        string localAddress =
+            Network.IPv4RFC5735Loopback.Contains(host) || Network.IPv6RFC4291Loopback.Contains(host)
+            ? _applicationHost.GetSmartApiUrl(NetworkManager.ServerLanAddress)
+            : _applicationHost.GetSmartApiUrl(host);
```
cvium, line 92: This actually breaks `EnablePublishedServerUriByRequest` and then there's also a discrepancy between GetSystemInfo and GetPublicSystemInfo.

The check should probably be put into `public string GetSmartApiUrl(HttpRequest request)`

Jellyfin.Networking/Manager/NetworkManager.cs
Sky-High: Sure!

Emby.Server.Implementations/SystemManager.cs
Sky-High: Done!

Jellyfin.Networking/Manager/NetworkManager.cs
Sky-High: Done!

Sky-High:
Merge confilcts are solved.

Is there anything else to do in order to merge this PR?

Gauvino:
When this could be merge please ?

Sky-High:
@cvium could you have a look at this again?

Jellyfin.Networking/Manager/NetworkManager.cs
```diff
@@ -1122,5 +1129,81 @@ private void PrintNetworkInformation(NetworkConfiguration config, bool debug = t
...
+                            rank = 3;
+                            ipSrv = ip;
+                        }
+                    }
+
+                    ServerLanAddress = ipSrv.ToString();
+                }
+            }
+
+            _logger.LogInformation("ServerLanAddress: {0} ", ServerLanAddress);
```
cvium, line 1206:
```suggestion
            _logger.LogInformation("ServerLanAddress: {ServerLanAddress} ", ServerLanAddress);
```

Jellyfin.Networking/Manager/NetworkManager.cs
```diff
@@ -1122,5 +1129,81 @@ private void PrintNetworkInformation(NetworkConfiguration config, bool debug = t
...
+        /// - prefer Unique Local Addresses (ULA) or Site Local Addresses over other (global) addresses.
+        /// - all things equal, prefer IPv4 over IPv6 as the former in practice appears to be more stable.
+        /// </summary>
+        private void InitializeServerLanAddress(NetworkConfiguration config)
+        {
+            lock (_initLock)
+            {
+                int rank = 0;   // ranks most preferable ip address
+                IPAddress ipSrv = IPAddress.None;
+                NetworkInterface[] adapters = NetworkInterface.GetAllNetworkInterfaces();
```
cvium, line 1151: We should only consider the ones we're actually binding to

Sky-High: Yes I do that by filtering with

https://github.com/Sky-High/jellyfin/blob/653c718c4439f622d1a3dea3e11997f2123ea809/Jellyfin.Networking/Manager/NetworkManager.cs#L1173

This outer/inner loop combination is kind of inefficient but the outerloop is needed to find the adapter matching the interface that are actually bound to, which are in the inner loop.

Sky-High:
This PR is labeled with `awaiting-feedback` and `needs testing`, but I am a somewhat lost here. I already provided some feedback, but that was insufficient? And re testing: I obviously tested only on my own system, so that indeed is a restriction. Is there anything else I can do to move this PR forward?

JPVenson:
hey @Sky-High 
`awaiting-feedback` means feedback from the core devs. Sory i am not the fastest with keeping the tags up-to-date. For the `needs-testing` is something i add when i see external facing changes that are best if somebody (that did not made the change) tests this. This is no flat requirement just an information for other core devs to look into.

In this specific case its cvium you requested an rereview.

JPVenson:
Oh a related note, i dont fully get the contents here but from what i understand is i have the question if we dont already have this behavior with the `Published Server URIs`?

barronpm:
> Oh a related note, i dont fully get the contents here but from what i understand is i have the question if we dont already have this behavior with the `Published Server URIs`?
From my understanding, published Server URIs are publicly accessible, whereas this PR is adding functionality to get a LAN-accessible IP for use by a chromecast

Sky-High:
Maybe some more explanation of the background for this PR is helpful. Its intention is primarily to enable 'simple' jellyfin systems to use chromecast devices (almost) without configuration. In particular, installing jellyfin on a Windows system with the default installer from https://repo.jellyfin.org/releases/server/windows/stable/ yields a jellyfin server and client side by side on a single machine that supports chromecast devices just out of the box.

Chromecast support requires some dedicated functionality in view of specific behavior of the Google Cast SDK. We specifically have to deal with the Cast SDK requirement that the webclient calling the Cast SDK has to be loaded and run from a secure host. As the jellyfin webclient typically is hosted by the jellyfin server built-in webservice, jellyfin server needs to run as secure webserver, with the corresponding requirement for a suitable certificate.

There is one exception to the latter requirement for a certificate: a webclient run from a webhost by means of a localhost url is considered by the Cast SDK as a secure connection even without a certificate. This enables us to run a simple standalone jellyfin media system without the hassle of setting up certificates by running jellyfin server and client on the same machine.

To elaborate: the Cast SDK requirement for a secure connection is just for the webservice hosting the jellyfin webclient. Jellyfin server features two webservices: one is said hosting of the webclient, and the other one does the 'real Jellyfin work' by serving the api for jellyfin media functionality. The first service needs to be secure, but this is not required for the api connection to jellyfin server. Also, the communication between cast device and the jellyfin media service does not need to be secure; neither does the communication between cast device and webclient.

So, running jellyfin client from jellyfin server on the same system allows for a convenient simple installation. There is one caveat however: when jellyfin client is started with the localhost url, this url is also set as url for the jellyfin media api service. That works well for api calls by the webclient to the jellyfin media service, but localhost as url can not be used  by the cast device to communicate with jellyfin server: a 'real' lan address is needed for that.

This PR implements the functionality to determine that 'real' lan address of Jellyfin server, to be used by the cast device. Usually this is straightforward, but in case of multiple subnets not trivial. This implementation takes the approach that the serveraddress on the subnet that provides a gateway is a smart choice.

This undoubtedly does not work for all networks, and more complicated subnet configurations still do require manual setup. But this approach doesn't make things worse either, as without this implementation Chromecast wouldn't work at all, also not for simple configurations. As an additional safeguard against unexpected behavior, the implementation only acts in the intended specific case where jellyfin server and webclient connect as localhost, so run side by side on a single machine.

So although this PR has a limitation, it is still valuable as it will provide a simple 'out of the box' installation for Jellyfin with chromecast.

PR #10443, opened 2023-10-21
Added LatestRelease home section type for "New Releases" home section

herby2212:
**Changes**
This PR adds a new home section type `LatestRelease` as preperation for the `New Releases` home section in the `jellyfin-web` 
 project which shows newly added items ordered by their official release date and not when they got added to the library.

**Issues**
https://github.com/orgs/jellyfin/projects/43/views/1?pane=issue&itemId=30103153

### DIFF
--- a/Jellyfin.Api/Controllers/DisplayPreferencesController.cs
+++ b/Jellyfin.Api/Controllers/DisplayPreferencesController.cs
@@ -126,6 +126,7 @@ public class DisplayPreferencesController : BaseJellyfinApiController
             HomeSectionType.LiveTv,
             HomeSectionType.NextUp,
             HomeSectionType.LatestMedia,
+            HomeSectionType.LatestRelease,
             HomeSectionType.None,
         };
 
--- a/Jellyfin.Api/Controllers/DisplayPreferencesController.cs
+++ b/Jellyfin.Api/Controllers/DisplayPreferencesController.cs
@@ -126,7 +126,6 @@ public class DisplayPreferencesController : BaseJellyfinApiController
             HomeSectionType.LiveTv,
             HomeSectionType.NextUp,
             HomeSectionType.LatestMedia,
-            HomeSectionType.LatestRelease,
             HomeSectionType.None,
         };
 
--- a/Jellyfin.Server/Migrations/Routines/MigrateDisplayPreferencesDb.cs
+++ b/Jellyfin.Server/Migrations/Routines/MigrateDisplayPreferencesDb.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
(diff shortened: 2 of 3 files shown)


Jellyfin.Server/Migrations/Routines/MigrateDisplayPreferencesDb.cs
```diff
@@ -68,9 +68,11 @@ public void Perform()
                 HomeSectionType.SmallLibraryTiles,
                 HomeSectionType.Resume,
                 HomeSectionType.ResumeAudio,
+                HomeSectionType.ResumeBook,
                 HomeSectionType.LiveTv,
                 HomeSectionType.NextUp,
                 HomeSectionType.LatestMedia,
+                HomeSectionType.LatestRelease,
                 HomeSectionType.None,
             };
```
herby2212, line 77: The order here got matched with the one in the `DisplayPreferencesController`.

Jellyfin.Api/Controllers/DisplayPreferencesController.cs
```diff
@@ -126,6 +126,7 @@ public DisplayPreferencesController(IDisplayPreferencesManager displayPreference
             HomeSectionType.LiveTv,
             HomeSectionType.NextUp,
             HomeSectionType.LatestMedia,
+            HomeSectionType.LatestRelease,
             HomeSectionType.None,
```
nielsvanvelzen, line 130: There is a limit of 6 home sections so adding this one doesn't make any sense, in fact the latest ".None" value doesn't make sense either.

Jellyfin.Server/Migrations/Routines/MigrateDisplayPreferencesDb.cs
nielsvanvelzen: You should not edit old migrations

herby2212: > You should not edit old migrations

That was the question here if there are cases the migration is still used. If so it should be talked if a update here makes sense or doesn't.

Jellyfin.Api/Controllers/DisplayPreferencesController.cs
herby2212: It seems that the limit is not 6, but 7 in the current web version.
image

I agree that setting 7 or 8 different ones by default doesn't need to be. Instead the UI should offer a option to add a new section on demand.

thegauravgiri (approved):
Seems good to me!

Jellyfin.Server/Migrations/Routines/MigrateDisplayPreferencesDb.cs
herby2212: I removed the changes here to keep the PR small.

Jellyfin.Api/Controllers/DisplayPreferencesController.cs
herby2212: I removed the addition of the new home section type as it doesn't make a lot of sense if the UI needs to be changed everytime as previously stated.

Bond-009 (approved):

Shadowghost (approved):

nielsvanvelzen:
Can you link the jellyfin-web pull request this is needed for?

Bond-009:
Closing as stale, no response in 2+ years

PR #10463, opened 2023-10-23
Update to .NET 8

Bond-009:


gaurovgiri (approved):
Looks good to me

Shadowghost (approved):

Jellyfin.Api/Controllers/EnvironmentController.cs
```diff
@@ -169,7 +169,7 @@ public IEnumerable<FileSystemEntryInfo> GetDrives()
             // Check if unc share
             var index = path.LastIndexOf(UncSeparator);
 
-            if (index != -1 && path.IndexOf(UncSeparator, StringComparison.OrdinalIgnoreCase) == 0)
+            if (index != -1 && path[0] == UncSeparator)
```
cvium, line 172: This will crash if path is ""

Bond-009: How? We already check if it contains `UncSeparator` once, so it must have at least length 1, right?

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -866,16 +861,16 @@ private StreamInfo BuildVideoItem(MediaSourceInfo item, MediaOptions options)
...
-                playlistItem.SetOption(qualifier, "audiochannels", audioStream.Channels.ToString() ?? string.Empty);
+                playlistItem.SetOption(qualifier, "audiochannels", audioStream.Channels?.ToString(CultureInfo.InvariantCulture) ?? string.Empty);
 
                 if (!string.IsNullOrEmpty(audioStream.Profile))
                 {
                     playlistItem.SetOption(audioStream.Codec, "profile", audioStream.Profile.ToLowerInvariant());
                 }
 
-                if (audioStream.Level != 0)
+                if (audioStream.Level.HasValue)
```
cvium, line 871: This changes the behaviour?

Jellyfin.Networking/HappyEyeballs/HttpClientExtension.cs
```diff
@@ -65,7 +65,7 @@ public static async ValueTask<Stream> OnConnect(SocketsHttpConnectionContext con
             // See https://github.com/dotnet/corefx/pull/29792/files#r189415885 for more details.
             if (await Task.WhenAny(tryConnectAsyncIPv6, Task.Delay(200, cancelIPv6.Token)).ConfigureAwait(false) == tryConnectAsyncIPv6 && tryConnectAsyncIPv6.IsCompletedSuccessfully)
             {
-                cancelIPv6.Cancel();
+                await cancelIPv6.CancelAsync().ConfigureAwait(false);
```
barronpm, line 68: Why is this awaited but the next line calling `GetAwaiter().GetResult()`?

Bond-009: `tryConnectAsyncIPv6` Is already completed, so `GetAwaiter().GetResult()` just gets the result

Jellyfin.Api/Controllers/EnvironmentController.cs
cvium: True

crobibero (approved):

Gauvino:
Why not changing the net version to 8 in dockerfile ?

cvium (approved):

barronpm (approved):

Issue #10468, opened 2023-10-24
Support Dolby Vision (Profile 8) media having HDR10/HLG fallback (on webOS)


viktory36:
Currently, in [MediaBrowser.Model/Entities/MediaStream.cs](https://github.com/jellyfin/jellyfin/blob/master/MediaBrowser.Model/Entities/MediaStream.cs#L688C1-L699C1), the GetVideoColorRange function exits as soon as it finds the media file to have HDR10 or HLG data.
This return happens too soon, not accounting for Dolby Vision Profile 8 media that have [both DOVI and HDR10/HLG](https://en.wikipedia.org/wiki/Dolby_Vision#Profiles) data - leading to only HDR10 level playback for DOVI media on my TV with DOVI support.

My approach included adding two new possible enumerations to `VideoRangeType` - `DOVI_with_HDR10_Fallback` and `DOVI_with_HLG_Fallback`, and adding these in wherever i found `VideoRangeType.HDR10 / HLG` or `VideoRangeType.DOVI` in the code. 

--
It seems redundant to have the client send in that they support both `HDR10` and `DOVI_with_HDR10_Fallback`, so I'm open to suggestions on how to better implement this - please do point me towards the files that would need changing as well!
I think the best way would be to add support for multiple VideoRangeTypes, but I'm not sure how/how much would have to change for that.

Another open question is how should the IsHwTonemapAvailable functions in [MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs](https://github.com/jellyfin/jellyfin/blob/208b585a3f7d9764c023024b2a14e619464bdc04/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs#L253C1-L256C76) handle these new video range types? It makes sense to me that if these dovi+hdr files were being processed as just hdr until now and worked fine, we should be able to add them along with the hdr10/hlg support assertion no problem.

I look forward to knowing your thoughts on this. I'll open a PR with the mentioned approach now.

viktory36:
Actually, the current behavior should be like "report HDR10/HLG, but play content as DoVi on supported devices anyway", right?
If that's the case, this change isn't really required and I'll just have something to change on jellyfin-web to get my webOS setup working.

nyanmisaka:
In fact, reporting this DOVI flag here means that the client MUST support ALL common DOVI Profiles, including but not limited to Profile 5 as well as 7 and 8. An example is mainstream Apple devices, which are Dolby Vision certified.

[...]

PR #10469, opened 2023-10-24
Add new VideoRangeTypes to fully support DoVi on webOS

viktory36:
This commit adds two new values to the VideoRangeType enum to talk about media that have both DOVI and HDR10 or HLG data (DoVi Profile 8) to enable Dolby Vision playback on those browsers.

**Changes**
Introduced `DOVI_with_HDR10_Fallback` and `DOVI_with_HLG_Fallback` VideoRangeType enums, and added references to these new types wherever I found `VideoRangeType.HDR10 / HLG`.
MediaStream.cs will now test both of it's HDR10/HLG and DOVI conditions before returning the video range type.

**Issues**
Addresses #10468

### DIFF
--- a/Jellyfin.Api/Controllers/DynamicHlsController.cs
+++ b/Jellyfin.Api/Controllers/DynamicHlsController.cs
@@ -1804,8 +1804,8 @@ public class DynamicHlsController : BaseJellyfinApiController
         {
             if (EncodingHelper.IsCopyCodec(codec)
                 && (state.VideoStream.VideoRangeType == VideoRangeType.DOVI
-                    || state.VideoStream.VideoRangeType == VideoRangeType.DolbyVisionWithHDR10Fallback
-                    || state.VideoStream.VideoRangeType == VideoRangeType.DolbyVisionWithHLGFallback
+                    || state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithHDR10
+                    || state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithHLG
                     || string.Equals(state.VideoStream.CodecTag, "dovi", StringComparison.OrdinalIgnoreCase)
                     || string.Equals(state.VideoStream.CodecTag, "dvh1", StringComparison.OrdinalIgnoreCase)
                     || string.Equals(state.VideoStream.CodecTag, "dvhe", StringComparison.OrdinalIgnoreCase)))
--- a/Jellyfin.Api/Controllers/DynamicHlsController.cs
+++ b/Jellyfin.Api/Controllers/DynamicHlsController.cs
@@ -1797,8 +1797,6 @@ public class DynamicHlsController : BaseJellyfinApiController
 
         var args = "-codec:v:0 " + codec;
 
-        var requestedRange = state.GetRequestedRangeTypes(codec);
-
         if (string.Equals(state.ActualOutputVideoCodec, "h265", StringComparison.OrdinalIgnoreCase)
             || string.Equals(state.ActualOutputVideoCodec, "hevc", StringComparison.OrdinalIgnoreCase)
             || string.Equals(codec, "h265", StringComparison.OrdinalIgnoreCase)
@@ -1806,8 +1804,8 @@ public class DynamicHlsController : BaseJellyfinApiController
         {
             if (EncodingHelper.IsCopyCodec(codec)
                 && (state.VideoStream.VideoRangeType == VideoRangeType.DOVI
-                    || (state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithHDR10 && requestedRange.Contains(VideoRangeType.DOVIWithHDR10.ToString(), StringComparison.OrdinalIgnoreCase))
-                    || (state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithHLG && requestedRange.Contains(VideoRangeType.DOVIWithHLG.ToString(), StringComparison.OrdinalIgnoreCase))))
+                    || state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithHDR10
+                    || state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithHLG))
             {
                 // Prefer dvh1 to dvhe
                 args += " -tag:v:0 dvh1 -strict -2";
--- a/Jellyfin.Api/Controllers/DynamicHlsController.cs
+++ b/Jellyfin.Api/Controllers/DynamicHlsController.cs
@@ -1797,18 +1797,22 @@ public class DynamicHlsController : BaseJellyfinApiController
 
         var args = "-codec:v:0 " + codec;
 
-        var requestedRange = state.GetRequestedRangeTypes(codec);
-
         if (string.Equals(state.ActualOutputVideoCodec, "h265", StringComparison.OrdinalIgnoreCase)
             || string.Equals(state.ActualOutputVideoCodec, "hevc", StringComparison.OrdinalIgnoreCase)
             || string.Equals(codec, "h265", StringComparison.OrdinalIgnoreCase)
             || string.Equals(codec, "hevc", StringComparison.OrdinalIgnoreCase))
         {
+            var requestedRange = state.GetRequestedRangeTypes(codec);
+            var requestHasDOVI = requestedRange.Contains(VideoRangeType.DOVI.ToString(), StringComparison.OrdinalIgnoreCase);
+            var requestHasDOVIWithHDR10 = requestedRange.Contains(VideoRangeType.DOVIWithHDR10.ToString(), StringComparison.OrdinalIgnoreCase);
+            var requestHasDOVIWithHLG = requestedRange.Contains(VideoRangeType.DOVIWithHLG.ToString(), StringComparison.OrdinalIgnoreCase);
+            var requestHasDOVIWithSDR = requestedRange.Contains(VideoRangeType.DOVIWithSDR.ToString(), StringComparison.OrdinalIgnoreCase);
+
             if (EncodingHelper.IsCopyCodec(codec)
-                && ((state.VideoStream.VideoRangeType == VideoRangeType.DOVI && requestedRange.Contains(VideoRangeType.DOVI.ToString(), StringComparison.OrdinalIgnoreCase))
-                    || (state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithHDR10 && requestedRange.Contains(VideoRangeType.DOVIWithHDR10.ToString(), StringComparison.OrdinalIgnoreCase))
-                    || (state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithHLG && requestedRange.Contains(VideoRangeType.DOVIWithHLG.ToString(), StringComparison.OrdinalIgnoreCase))
-                    || (state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithSDR && requestedRange.Contains(VideoRangeType.DOVIWithSDR.ToString(), StringComparison.OrdinalIgnoreCase))))
+                && ((state.VideoStream.VideoRangeType == VideoRangeType.DOVI && requestHasDOVI)
+                    || (state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithHDR10 && requestHasDOVIWithHDR10)
+                    || (state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithHLG && requestHasDOVIWithHLG)
+                    || (state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithSDR && requestHasDOVIWithSDR)))
             {
                 // Prefer dvh1 to dvhe
                 args += " -tag:v:0 dvh1 -strict -2";
--- a/Jellyfin.Api/Controllers/DynamicHlsController.cs
+++ b/Jellyfin.Api/Controllers/DynamicHlsController.cs
@@ -1807,7 +1807,8 @@ public class DynamicHlsController : BaseJellyfinApiController
             if (EncodingHelper.IsCopyCodec(codec)
                 && ((state.VideoStream.VideoRangeType == VideoRangeType.DOVI && requestedRange.Contains(VideoRangeType.DOVI.ToString(), StringComparison.OrdinalIgnoreCase))
                     || (state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithHDR10 && requestedRange.Contains(VideoRangeType.DOVIWithHDR10.ToString(), StringComparison.OrdinalIgnoreCase))
-                    || (state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithHLG && requestedRange.Contains(VideoRangeType.DOVIWithHLG.ToString(), StringComparison.OrdinalIgnoreCase))))
+                    || (state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithHLG && requestedRange.Contains(VideoRangeType.DOVIWithHLG.ToString(), StringComparison.OrdinalIgnoreCase))
+                    || (state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithSDR && requestedRange.Contains(VideoRangeType.DOVIWithSDR.ToString(), StringComparison.OrdinalIgnoreCase))))
             {
                 // Prefer dvh1 to dvhe
                 args += " -tag:v:0 dvh1 -strict -2";
--- a/Jellyfin.Api/Controllers/DynamicHlsController.cs
+++ b/Jellyfin.Api/Controllers/DynamicHlsController.cs
@@ -1804,8 +1804,8 @@ public class DynamicHlsController : BaseJellyfinApiController
         {
             if (EncodingHelper.IsCopyCodec(codec)
                 && (state.VideoStream.VideoRangeType == VideoRangeType.DOVI
-                    || state.VideoStream.VideoRangeType == VideoRangeType.DOVI_with_HDR10_Fallback
-                    || state.VideoStream.VideoRangeType == VideoRangeType.DOVI_with_HLG_Fallback
+                    || state.VideoStream.VideoRangeType == VideoRangeType.DolbyVisionWithHDR10Fallback
+                    || state.VideoStream.VideoRangeType == VideoRangeType.DolbyVisionWithHLGFallback
                     || string.Equals(state.VideoStream.CodecTag, "dovi", StringComparison.OrdinalIgnoreCase)
                     || string.Equals(state.VideoStream.CodecTag, "dvh1", StringComparison.OrdinalIgnoreCase)
                     || string.Equals(state.VideoStream.CodecTag, "dvhe", StringComparison.OrdinalIgnoreCase)))
--- a/Jellyfin.Api/Controllers/DynamicHlsController.cs
+++ b/Jellyfin.Api/Controllers/DynamicHlsController.cs
@@ -1797,15 +1797,17 @@ public class DynamicHlsController : BaseJellyfinApiController
 
         var args = "-codec:v:0 " + codec;
 
+        var requestedRange = state.GetRequestedRangeTypes(codec);
+
         if (string.Equals(state.ActualOutputVideoCodec, "h265", StringComparison.OrdinalIgnoreCase)
             || string.Equals(state.ActualOutputVideoCodec, "hevc", StringComparison.OrdinalIgnoreCase)
             || string.Equals(codec, "h265", StringComparison.OrdinalIgnoreCase)
             || string.Equals(codec, "hevc", StringComparison.OrdinalIgnoreCase))
         {
             if (EncodingHelper.IsCopyCodec(codec)
-                && (state.VideoStream.VideoRangeType == VideoRangeType.DOVI
-                    || state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithHDR10
-                    || state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithHLG))
+                && ((state.VideoStream.VideoRangeType == VideoRangeType.DOVI && requestedRange.Contains(VideoRangeType.DOVI.ToString(), StringComparison.OrdinalIgnoreCase))
+                    || (state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithHDR10 && requestedRange.Contains(VideoRangeType.DOVIWithHDR10.ToString(), StringComparison.OrdinalIgnoreCase))
+                    || (state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithHLG && requestedRange.Contains(VideoRangeType.DOVIWithHLG.ToString(), StringComparison.OrdinalIgnoreCase))))
             {
                 // Prefer dvh1 to dvhe
                 args += " -tag:v:0 dvh1 -strict -2";
--- a/Jellyfin.Api/Controllers/DynamicHlsController.cs
+++ b/Jellyfin.Api/Controllers/DynamicHlsController.cs
@@ -1807,10 +1807,7 @@ public class DynamicHlsController : BaseJellyfinApiController
             if (EncodingHelper.IsCopyCodec(codec)
                 && (state.VideoStream.VideoRangeType == VideoRangeType.DOVI
                     || (state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithHDR10 && requestedRange.Contains(VideoRangeType.DOVIWithHDR10.ToString(), StringComparison.OrdinalIgnoreCase))
-                    || (state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithHLG && requestedRange.Contains(VideoRangeType.DOVIWithHLG.ToString(), StringComparison.OrdinalIgnoreCase))
-                    || string.Equals(state.VideoStream.CodecTag, "dovi", StringComparison.OrdinalIgnoreCase)
-                    || string.Equals(state.VideoStream.CodecTag, "dvh1", StringComparison.OrdinalIgnoreCase)
-                    || string.Equals(state.VideoStream.CodecTag, "dvhe", StringComparison.OrdinalIgnoreCase)))
+                    || (state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithHLG && requestedRange.Contains(VideoRangeType.DOVIWithHLG.ToString(), StringComparison.OrdinalIgnoreCase))))
             {
                 // Prefer dvh1 to dvhe
                 args += " -tag:v:0 dvh1 -strict -2";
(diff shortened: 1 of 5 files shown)


viktory36:
The build pipeline does not like the name `DOVI_with_HDR10_Fallback` as they have underscores. Do you have other suggestions for the enum name?

`
/home/vsts/work/1/s/Jellyfin.Data/Enums/VideoRangeType.cs(36,5): error CA1707: Remove the underscores from member name Jellyfin.Data.Enums.VideoRangeType.DOVI_with_HDR10_Fallback (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1707) [/home/vsts/work/1/s/Jellyfin.Data/Jellyfin.Data.csproj]
/home/vsts/work/1/s/Jellyfin.Data/Enums/VideoRangeType.cs(41,5): error CA1707: Remove the underscores from member name Jellyfin.Data.Enums.VideoRangeType.DOVI_with_HLG_Fallback (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1707) [/home/vsts/work/1/s/Jellyfin.Data/Jellyfin.Data.csproj]
`

crobibero:
> The build pipeline does not like the name DOVI_with_HDR10_Fallback as they have underscores. Do you have other suggestions for the enum name?


Maybe just DOVIHDR10 and DOVIHLG?

it's more difficult since both are acronyms..

viktory36:
> > The build pipeline does not like the name DOVI_with_HDR10_Fallback as they have underscores. Do you have other suggestions for the enum name?
> 
> Maybe just `DOVIHDR10` and `DOVIHLG`?
> 
> it's more difficult since both are acronyms..

Yup. I was thinking more of what the end user might see - if they navigate over to the playback info screen and find "DOVIHDR10" or "DOVIProfile8" I think that would be more confusing than the current.

DOVIwithHDR10
DOVIwithHLG

DoViHDR10
DoViHLG

I think I like this last one DoViHDR10 better, but not very much - don't want to mislead any users into thinking they're seeing DOVI when it's HDR.

Would you be opposed to disabling this warning CA1707? Would throwing that into the jellyfin.ruleset work for the build pipeline to pass?

viktory36:
Changed it to `DolbyVisionWithHDR10Fallback`. The only failing check now also failed for the initial PR introducing VideoRangeType enums. Not sure what needs to change to correct that.

I also reverted the "[Dont assume HLG/HDR10 tonemapper can support DOVI+HLG/HDR10](https://github.com/jellyfin/jellyfin/pull/10469/commits/d330436719ba2b567bcb948b86de1f329351c650)" commit - it makes sense that if the DOVI+HDR10 media were previously reported as HDR10 and had no issues with any HwTonemap, the new range types representing this media should fall in the same bucket.

viktory36:
Dear @nyanmisaka , please let me know your thoughts on this?

Jellyfin.Data/Enums/VideoRangeType.cs
```diff
@@ -30,6 +30,16 @@ public enum VideoRangeType
...
 
+    /// <summary>
+    /// Dolby Vision with HDR10 video range fallback (12bit or 10bit).
+    /// </summary>
+    DolbyVisionWithHDR10Fallback,
+
+    /// <summary>
+    /// Dolby Vision with HLG video range fallback (12bit or 10bit).
+    /// </summary>
+    DolbyVisionWithHLGFallback,
```
nyanmisaka, line 41: We can shorten its name a little bit.

```cs
    /// <summary>
    /// Dolby Vision video range type without any fallback (12bit).
    /// </summary>
    DOVI,

    /// <summary>
    /// Dolby Vision with HDR10 video range type fallback (12bit or 10bit).
    /// </summary>
    DOVIWithHDR10,

    /// <summary>
    /// Dolby Vision with HLG video range type fallback (12bit or 10bit).
    /// </summary>
    DOVIWithHLG,
```

MediaBrowser.Model/Entities/MediaStream.cs
```diff
@@ -712,9 +704,45 @@ public bool SupportsSubtitleConversionTo(string toCodec)
...
+            if (isDoVi)
+            {
+                if (isHDR10)
+                {
+                    return (VideoRange.HDR, VideoRangeType.DolbyVisionWithHDR10Fallback);
+                }
+                else if (isHLG)
+                {
+                    return (VideoRange.HDR, VideoRangeType.DolbyVisionWithHLGFallback);
+                }
```
nyanmisaka, line 731: IMO this is not that reliable. We have `dvBlCompatId` that can be used to identify the DOVI fallback types.

https://github.com/jellyfin/jellyfin/blob/8f2829e1aa696d37d338fceed2ea1d9edf1ca824/MediaBrowser.Model/Entities/MediaStream.cs#L206-L208

nyanmisaka:
> @nyanmisaka I have some changes locally that implement the suggestions above, but I can't push to the pull request branch. Should I open a new pull request, as the author seems inactive? @viktory36 I think there is an option to permit that, if you would be so kind as to enable it.

If the OP still doesn't respond, you can make your branch based on his commits and open a new PR.

viktory36:
Hey @GeorgeH005, I've enabled "allow edits by maintainers" to this PR, and added you as a collaborator to my fork if direct commits to the pr doesn't work.
If that's still not enough to be able to add a commit to this PR, please feel free to raise another.

Thanks and excited to have this on main!

Jellyfin.Api/Controllers/DynamicHlsController.cs
```diff
@@ -1804,6 +1804,8 @@ private string GetVideoArguments(StreamState state, int startNumber, bool isEven
         {
             if (EncodingHelper.IsCopyCodec(codec)
                 && (state.VideoStream.VideoRangeType == VideoRangeType.DOVI
+                    || state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithHDR10
+                    || state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithHLG
```
nyanmisaka, line 1808: Some clients do not support `dvh1` FourCC but they support HDR10 & HLG in `hvc1`, such as the HLS.js used in non-Safari browsers. Please refer to the [code here](https://github.com/jellyfin/jellyfin/blob/fab3151679999bddf767f8c1e5e0826f58f63e4f/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs#L2200-L2212) to set the appropriate video tag according to the VideoRangeType reported by the client.

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -6118,7 +6121,9 @@ public string GetVideotoolboxVidDecoder(EncodingJobInfo state, EncodingOptions o
                                     && state.VideoStream.VideoRange == VideoRange.HDR
                                     && (state.VideoStream.VideoRangeType == VideoRangeType.HDR10
                                         || state.VideoStream.VideoRangeType == VideoRangeType.HLG
-                                        || (state.VideoStream.VideoRangeType == VideoRangeType.DOVI
+                                        || ((state.VideoStream.VideoRangeType == VideoRangeType.DOVI
+                                            || state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithHDR10
+                                            || state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithHLG)
```
nyanmisaka, line 6126:
```suggestion
                                        || (state.VideoStream.VideoRangeType == VideoRangeType.DOVI
```

In VideoToolbox the `useOclToneMapping` is only enabled for DOVI Profile 5. No need to change here.

nyanmisaka (changes requested):
When everything is resolved, this should be merged after the relevant changes in jellyfin-web are merged. Merging this one alone will cause regression.

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
gnattu: I *think* it could be needed, as the user might have explicitly disabled VideoToolbox's native tone mapping. If `VideoRangeType.DOVI` does not imply `VideoRangeType.DOVIWithHDR10` or `VideoRangeType.DOVIWithHLG`, then users with such configurations might not have hwsurface initialized, which could cause hwmap to fail.

gnattu: > If so there is no reason to exclude VideoRangeType.{HDR10,HLG}.
They are already included, the code segment shown here might be incomplete, but all kind of HDR is already included
https://github.com/jellyfin/jellyfin/blob/6b4951fcf5c0fe2a473b2c13367fdf367a4a4c78/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs#L6121-L6128

nyanmisaka: Then this is necessary. I read the wrong line.

viktory36:
@GeorgeH005 , there's a PR open for web here with an open review comment: https://github.com/jellyfin/jellyfin-web/pull/4916

Let me take care of that one

Jellyfin.Api/Controllers/DynamicHlsController.cs
```diff
@@ -1797,13 +1797,17 @@ private string GetVideoArguments(StreamState state, int startNumber, bool isEven
...
             || string.Equals(codec, "h265", StringComparison.OrdinalIgnoreCase)
             || string.Equals(codec, "hevc", StringComparison.OrdinalIgnoreCase))
         {
             if (EncodingHelper.IsCopyCodec(codec)
                 && (state.VideoStream.VideoRangeType == VideoRangeType.DOVI
+                    || (state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithHDR10 && requestedRange.Contains(VideoRangeType.DOVIWithHDR10.ToString(), StringComparison.OrdinalIgnoreCase))
+                    || (state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithHLG && requestedRange.Contains(VideoRangeType.DOVIWithHLG.ToString(), StringComparison.OrdinalIgnoreCase))
                     || string.Equals(state.VideoStream.CodecTag, "dovi", StringComparison.OrdinalIgnoreCase)
                     || string.Equals(state.VideoStream.CodecTag, "dvh1", StringComparison.OrdinalIgnoreCase)
                     || string.Equals(state.VideoStream.CodecTag, "dvhe", StringComparison.OrdinalIgnoreCase)))
```
nyanmisaka, line 1813: We can drop these now.
```suggestion
```

nyanmisaka: ffmpeg currently does not support those profiles (handling metdata)... Profile 10 for AV1 and Profile 10 for 3D.

nyanmisaka: Currently, the support for DV in FFmpeg is limited to P5/7/8 implemented through reverse engineering. Additional metadata will be ignored, so setting dvh1 here won't help.

nyanmisaka (approved):

nyanmisaka:
> @nyanmisaka According to our conversation over in [jellyfin/jellyfin-web#5282](https://github.com/jellyfin/jellyfin-web/pull/5282), should I adjust the code to only tag profiles 5 and 8 as dolby vision, and the rest as regular hdr/hlg or sdr?

DV P5/8 => (HDR, DOVI*)
DV P7 => (HDR, HDR10) // no HLG in P7, i think?
Other DV => (SDR, SDR)

nyanmisaka:
> Should I also check if the client has requested DOVI as a VideoRangeType before tagging the stream as dvh1? Profile 5 will look wrong in both cases if dvh1 isn't supported ( I think ), but Profile 8.4 if tagged as hvc1 should fallback to SDR, which seems the better behaviour to me. These are the only two cases that I think this applies to

Just tested and found we should move that `requestedRange.Contains` to `Encodinghelper->CanStreamCopyVideo()`
https://github.com/jellyfin/jellyfin/blob/b6d130ae2d337294904f6d46d943e9f82c162978/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs#L2200-L2212

And enable `dvh1` for all DOVI*. Otherwise this will prevent remuxing and trigger transcoding.

nyanmisaka:
> @nyanmisaka That makes sense, I assume that if CanStreamCopyVideo is false then OutputVideoCodec is false, so EncodingHelper.IsCopyCodec is also false, so those checks in GetVideoArguments are redundant anyways, correct?

Yes. When the program runs to GetVideoArguments ->(IsCopyCodec && is dvh1), this means that the client and the server negotiate and prepare for remuxing instead of transcoding.

Just allow the server to enable stream copy of DV P8 for clients (Chrome/hls.js) that only support HDR10/HLG in `CanStreamCopyVideo()`.

nyanmisaka:
> Just allow the server to enable stream copy of DV P8 for clients (Chrome/hls.js) that only support HDR10/HLG in `CanStreamCopyVideo()`.

No. ^^^

nyanmisaka:
> I thought previously that tagging Profile 8.4 (DOVI + SDR) as DOVI would be correct, but since the stream isn't going to be copied in case DOVI is not supported, there is going to be unnecessary transcoding/tonemapping. Should I introduce a new type for 8.4? Handling it as SDR in the first place doesn't seem like a good idea.

```cs
                return dvProfile switch
                {
                    5 => (VideoRange.HDR, VideoRangeType.DOVI),
                    8 => dvBlCompatId switch
                    {
                        1 => (VideoRange.HDR, VideoRangeType.DOVIWithHDR10),
                        4 => (VideoRange.HDR, VideoRangeType.DOVIWithHLG), //< you mean this ? profile 8.4 HLG
                        _ => (VideoRange.HDR, VideoRangeType.DOVI)
                    },
                    7 => (VideoRange.HDR, VideoRangeType.HDR10),
                    _ => (VideoRange.SDR, VideoRangeType.SDR)
                };
```

Which profile is SDR?

nyanmisaka:
> > And enable `dvh1` for all DOVI*. Otherwise this will prevent remuxing and trigger transcoding.
> > If Chrome does not report DOVIWithHDR10 support and the video file is DV P8, then CanStreamCopyVideo() will return false.
> 
> @nyanmisaka But then the stream is going to be tagged as dvh1 even with Chrome as a client, and based on previous messages my understanding is that for the fallback to work, the stream needs to be hvc1.
> 
> > Just tested and found we should move that requestedRange.Contains to Encodinghelper->CanStreamCopyVideo()
> 
> So I think that check needs to exist in both places, not be moved

Yeah. You are right. I take it too easy.

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -2205,7 +2208,10 @@ public bool CanStreamCopyVideo(EncodingJobInfo state, MediaStream videoStream)
                     return false;
                 }
 
-                if (!requestedRangeTypes.Contains(videoStream.VideoRangeType.ToString(), StringComparison.OrdinalIgnoreCase))
+                // DOVIWithHDR10 should be compatible with HDR10 supporting players. Same goes with HLG. So allow copy of those formats
+                if (!requestedRangeTypes.Contains(videoStream.VideoRangeType.ToString(), StringComparison.OrdinalIgnoreCase)
+                     && !((requestedRangeTypes.Contains("HDR10", StringComparison.OrdinalIgnoreCase) && string.Equals(videoStream.VideoRangeType.ToString(), "DOVIWithHDR10", StringComparison.OrdinalIgnoreCase))
+                            || (requestedRangeTypes.Contains("HLG", StringComparison.OrdinalIgnoreCase) && string.Equals(videoStream.VideoRangeType.ToString(), "DOVIWithHLG", StringComparison.OrdinalIgnoreCase))))
```
nyanmisaka, line 2214:
```suggestion
                     && !((requestedRangeTypes.Contains("HDR10", StringComparison.OrdinalIgnoreCase) && videoStream.VideoRangeType == VideoRangeType.DOVIWithHDR10))
                            || (requestedRangeTypes.Contains("HLG", StringComparison.OrdinalIgnoreCase) && videoStream.VideoRangeType == VideoRangeType.DOVIWithHLG)))
```

nyanmisaka:
> @nyanmisaka My mistake, Profile 8.2. Sorry about that

This should be okay? We only care about HDR here. So just filter out the `dvBlCompatId` that are known to contain HDR compatible layers.
```
                    8 => dvBlCompatId switch
                    {
                        1 => (VideoRange.HDR, VideoRangeType.DOVIWithHDR10),
                        4 => (VideoRange.HDR, VideoRangeType.DOVIWithHLG), //< you mean this ? profile 8.4 HLG
                        _ => (VideoRange.SDR, VideoRangeType.SDR)
                    },
```

nyanmisaka:
Then I think you can add a `DOVIWithSDR`. Even though I don't have such a clip I know it exists.

nyanmisaka:
So it's an enhanced SDR video, so it *should* be (SDR, DOVIWithSDR).
It only makes sense during stream-copy. In other cases, there is no difference from traditional SDR.

nyanmisaka:
> @nyanmisaka Did some more research and found out that Apple's HLS spec supports profiles 5 and 8, but mpeg-ts should support dual layer 7 and every other available profile, at least according to Dolby. Not sure if we want to change our implementation now, at least when it comes to Profile 7, as I cannot speak at all for client support. I know we already had a conversation about Profile 7, but this is news to me. Given the option "Prefer fmp4-hls container..." in the playback settings, I assumed that there were mpeg-ts streaming capabilities. If I am wrong, of course disregard the above comment.

Apple just rejects the HEVC codec in MPEG-TS HLS, so this won't happen.

Jellyfin.Data/Enums/VideoRangeType.cs
```diff
@@ -26,10 +26,25 @@ public enum VideoRangeType
     HLG,
 
     /// <summary>
-    /// Dolby Vision video range type (12bit).
+    /// Dolby Vision video range type (10bit).
```
nyanmisaka, line 29: Profile 5 uses ICtCp instead of YCbCr, which is more efficient. So it's 10-Bit encoded/12-bit remapped.

MediaBrowser.Model/Entities/MediaStream.cs
```diff
@@ -707,34 +707,46 @@ public bool SupportsSubtitleConversionTo(string toCodec)
...
-
             var codecTag = CodecTag;
             var dvProfile = DvProfile;
             var rpuPresentFlag = RpuPresentFlag == 1;
             var blPresentFlag = BlPresentFlag == 1;
             var dvBlCompatId = DvBlSignalCompatibilityId;
 
             var isDoViHDRProfile = dvProfile == 5 || dvProfile == 7 || dvProfile == 8;
-            var isDoViHDRFlag = rpuPresentFlag && blPresentFlag && (dvBlCompatId == 0 || dvBlCompatId == 1 || dvBlCompatId == 4);
+            var isDoViHDRFlag = rpuPresentFlag && blPresentFlag && (dvBlCompatId == 0 || dvBlCompatId == 1 || dvBlCompatId == 4 || dvBlCompatId == 2 || dvBlCompatId == 6);
```
nyanmisaka, line 717: > dvBlCompatId == 2 || dvBlCompatId == 6

It's actually `isDoViSDRFlag`.

So just rename:
`isDoViHDRProfile` to `isDoViProfile`
`isDoViHDRFlag` to `isDoViFlag`

Jellyfin.Api/Controllers/DynamicHlsController.cs
```diff
@@ -1797,16 +1797,18 @@ private string GetVideoArguments(StreamState state, int startNumber, bool isEven
 
         var args = "-codec:v:0 " + codec;
 
+        var requestedRange = state.GetRequestedRangeTypes(codec);
+
```
nyanmisaka, line 1801:
```suggestion
```

Jellyfin.Api/Controllers/DynamicHlsController.cs
```diff
@@ -1797,16 +1797,18 @@ private string GetVideoArguments(StreamState state, int startNumber, bool isEven
...
         if (string.Equals(state.ActualOutputVideoCodec, "h265", StringComparison.OrdinalIgnoreCase)
             || string.Equals(state.ActualOutputVideoCodec, "hevc", StringComparison.OrdinalIgnoreCase)
             || string.Equals(codec, "h265", StringComparison.OrdinalIgnoreCase)
             || string.Equals(codec, "hevc", StringComparison.OrdinalIgnoreCase))
         {
```
nyanmisaka, line 1806:
```suggestion
        {
            var types = state.GetRequestedRangeTypes(codec);
            var haveDOVI = types.Contains(VideoRangeType.DOVI.ToString(), StringComparison.OrdinalIgnoreCase);
            var haveDOVIWithHDR10 = types.Contains(VideoRangeType.DOVIWithHDR10.ToString(), StringComparison.OrdinalIgnoreCase);
            var haveDOVIWithHLG = types.Contains(VideoRangeType.DOVIWithHLG.ToString(), StringComparison.OrdinalIgnoreCase);
            var haveDOVIWithSDR = types.Contains(VideoRangeType.DOVIWithSDR.ToString(), StringComparison.OrdinalIgnoreCase);

```

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -2205,7 +2208,11 @@ public bool CanStreamCopyVideo(EncodingJobInfo state, MediaStream videoStream)
                     return false;
                 }
 
-                if (!requestedRangeTypes.Contains(videoStream.VideoRangeType.ToString(), StringComparison.OrdinalIgnoreCase))
+                // DOVIWithHDR10 should be compatible with HDR10 supporting players. Same goes with HLG and of course SDR. So allow copy of those formats
+                if (!requestedRangeTypes.Contains(videoStream.VideoRangeType.ToString(), StringComparison.OrdinalIgnoreCase)
+                     && !((requestedRangeTypes.Contains("HDR10", StringComparison.OrdinalIgnoreCase) && videoStream.VideoRangeType == VideoRangeType.DOVIWithHDR10)
+                            || (requestedRangeTypes.Contains("HLG", StringComparison.OrdinalIgnoreCase) && videoStream.VideoRangeType == VideoRangeType.DOVIWithHLG)
+                            || (requestedRangeTypes.Contains("SDR", StringComparison.OrdinalIgnoreCase) && videoStream.VideoRangeType == VideoRangeType.DOVIWithSDR)))
```
nyanmisaka, line 2215:
```suggestion
                var haveDOVIWithHDR10 = requestedRangeTypes.Contains(VideoRangeType.DOVIWithHDR10.ToString(), StringComparison.OrdinalIgnoreCase);
                var haveDOVIWithHLG = requestedRangeTypes.Contains(VideoRangeType.DOVIWithHLG.ToString(), StringComparison.OrdinalIgnoreCase);
                var haveDOVIWithSDR = requestedRangeTypes.Contains(VideoRangeType.DOVIWithSDR.ToString(), StringComparison.OrdinalIgnoreCase);

                // DOVIWithHDR10 should be compatible with HDR10 supporting players. Same goes with HLG and of course SDR. So allow copy of those formats
                if (!requestedRangeTypes.Contains(videoStream.VideoRangeType.ToString(), StringComparison.OrdinalIgnoreCase)
                    && !((videoStream.VideoRangeType == VideoRangeType.DOVIWithHDR10 && haveDOVIWithHDR10)
                          || (videoStream.VideoRangeType == VideoRangeType.DOVIWithHLG && haveDOVIWithHLG)
                          || (videoStream.VideoRangeType == VideoRangeType.DOVIWithSDR && haveDOVIWithSDR)))
```

Each line is already too long, Extract the contains* to vars to increase readability.

Jellyfin.Api/Controllers/DynamicHlsController.cs
```diff
@@ -1797,16 +1797,18 @@ private string GetVideoArguments(StreamState state, int startNumber, bool isEven
...
-                && (state.VideoStream.VideoRangeType == VideoRangeType.DOVI
-                    || string.Equals(state.VideoStream.CodecTag, "dovi", StringComparison.OrdinalIgnoreCase)
-                    || string.Equals(state.VideoStream.CodecTag, "dvh1", StringComparison.OrdinalIgnoreCase)
-                    || string.Equals(state.VideoStream.CodecTag, "dvhe", StringComparison.OrdinalIgnoreCase)))
+                && ((state.VideoStream.VideoRangeType == VideoRangeType.DOVI && requestedRange.Contains(VideoRangeType.DOVI.ToString(), StringComparison.OrdinalIgnoreCase))
+                    || (state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithHDR10 && requestedRange.Contains(VideoRangeType.DOVIWithHDR10.ToString(), StringComparison.OrdinalIgnoreCase))
+                    || (state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithHLG && requestedRange.Contains(VideoRangeType.DOVIWithHLG.ToString(), StringComparison.OrdinalIgnoreCase))
+                    || (state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithSDR && requestedRange.Contains(VideoRangeType.DOVIWithSDR.ToString(), StringComparison.OrdinalIgnoreCase))))
```
nyanmisaka, line 1811:
```suggestion
                && ((state.VideoStream.VideoRangeType == VideoRangeType.DOVI && haveDOVI)
                     || (state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithHDR10 && haveDOVIWithHDR10)
                     || (state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithHLG && haveDOVIWithHLG)
                     || (state.VideoStream.VideoRangeType == VideoRangeType.DOVIWithSDR && haveDOVIWithSDR)))
```

Each line is already too long, Extract the contains* to vars to increase readability.

nyanmisaka:
Just want to double check, have you also tested these PRs on real webOS hardware and Chrome with fMP4-HLS enabled?

nyanmisaka:
> The hev1 mp4 might be something to look into. When converting that I did not specify a "-tag:v:0 hev1" argument I just did a "-i 12345.mkv -c:v copy -c:a copy -sn 12345.mp4".

Are the source files mentioned here DV P8? What results did you expect?

viktory36:
@GeorgeH005 nitpicked on the DOVIWithSDR enum description a little bit =] I reasoned SDR can commonly appear in 8bit or 10bit color depth, and if DV will be at 10bit depth, then (8bit / 10bit) is a fair statement to make -- though I haven't found a file that has DV with 8bit SDR fallback happening on the same media.

nyanmisaka:
> The source file was an mkv with Dolby Vision Profile 8 but I did not specify a tag when converting the file, so it might have been my fault

ffmpeg only passes DV metadata from demuxer to muxer when `-strict -2` is used.

nyanmisaka (approved):
Then I think this is good to go. The web side fixes should be merged first.

Considering this actually fixes Dolby Vision playback on webOS, IMO this isn't a new feature.

crobibero (approved):

nyanmisaka:
> @nyanmisaka The jellyfin-web team believes this should be merged before theirs.

~~1. https://github.com/jellyfin/jellyfin-web/pull/4916~~
2. https://github.com/jellyfin/jellyfin/pull/10469
3. https://github.com/jellyfin/jellyfin-web/pull/5282

All right. Let's merge this first. cc @crobibero

Issue #6247, opened 2021-06-30
Possible to set EnableMetrics via environment variable?


se-jaeger:
I tried to enable prometheus monitoring at startup but could not find the possibility to see `EnableMetrics` to true.

Does somebody know if and how this is possible?

I checked the code but I'm not familiar with C#.

BaronGreenback:
The setting is in the settings.xml file found in the data folder.

se-jaeger:
Thanks, I actually found it and it works as expected. My question was rather if it is possible to enable monitoring before the first startup. I use automated deployment on kubernetes and it would be great to set this with an env variable for example.

BaronGreenback:
Not at the moment. Might be able to have two config folders and do it via a different environmental var.

srcrist:
Closing this issue as it does not reference a Jellyfin bug. I took the liberty of migrating this feature request to https://features.jellyfin.org/posts/2397/enable-metrics-with-env-variable for you.

donaldgifford:
Add this `JELLYFIN_EnableMetrics=true` for it to use the env var to set it.

PR #10503, opened 2023-10-31
feat: add EnableMetrics override with environment variable

yohanb:
**Changes**
Adds the ability to override the `EnableMetrics` configuration value from the `system.xml` files.  The configuration value is added to the ASP.NET config framework so it can also be set through the other config sources (`appsettings.json`, etc.),

**Issues**
[Fixes issue 6247](https://github.com/jellyfin/jellyfin/issues/6247) / [Feature request](https://features.jellyfin.org/posts/2397/enable-metrics-with-env-variable)

### DIFF
--- a/Emby.Server.Implementations/ApplicationHost.cs
+++ b/Emby.Server.Implementations/ApplicationHost.cs
@@ -263,6 +263,8 @@ namespace Emby.Server.Implementations
         /// <inheritdoc/>
         public bool ListenWithHttps => Certificate is not null && ConfigurationManager.GetNetworkConfiguration().EnableHttps;
 
+        public bool EnableMetrics => ConfigurationManager.Configuration.EnableMetrics || this._startupConfig.GetValue<bool>(EnableMetricsKey);
+
         public string FriendlyName =>
             string.IsNullOrEmpty(ConfigurationManager.Configuration.ServerName)
                 ? Environment.MachineName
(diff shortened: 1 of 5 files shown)


yohanb:
anyone from the Jellyfin maintainers gonna chime in?

Emby.Server.Implementations/ApplicationHost.cs
```diff
@@ -263,6 +263,8 @@ public abstract class ApplicationHost : IServerApplicationHost, IDisposable
         /// <inheritdoc/>
         public bool ListenWithHttps => Certificate is not null && ConfigurationManager.GetNetworkConfiguration().EnableHttps;
 
+        public bool EnableMetrics => ConfigurationManager.Configuration.EnableMetrics || this._startupConfig.GetValue<bool>(EnableMetricsKey);
```
cvium, line 266: Two things:

1. We don't use `this`
2. One should take priority over the other (probably the env var). With this expression, both must either not be set or be false to disable it.

JPVenson, line 266: Also i dont like the location of this property. FeatureFlags should not be aggregated in the Host class. Why not keep it in the ConfigManager and extend the check there?

JPVenson (changes requested):

JPVenson:
closed due to inactivity

PR #10508, opened 2023-11-01
Added ActivityLogDto

### LINKED ISSUE #3122: Use DTO For ActivityLog
As discussed [here](https://github.com/jellyfin/jellyfin/pull/2970#discussion_r424909191), the database entity should not be returned directly from the API, but instead first be converted to DTOs.

Paulr756756:
Added Dto for ActivityLog Entity, updated ActivityManager.CreateAsync  to use a 
dto instead of directly accessing the entity. PR resolves #3122 

**Changes**

- New ActivityLogDto class created
- Updated ActivityManager.CreateAsync to use DTO
- New ActivityManager.GetActivityLogEntity method for conversion
- Refactored CreateAsync references 
- Updated CONTRIBUTORS.md

**Issues**

#3122


nielsvanvelzen (changes requested):
The OpenAPI spec still exposes ActivityLogEntry and doesn't even include the new ActivityLogDto class so this PR is incomplete.

Paulr756756:
> The OpenAPI spec still exposes ActivityLogEntry and doesn't even include the new ActivityLogDto class so this PR is incomplete.

Sorry I'm new to this so i couldn't understand. Could you please clarify.

nielsvanvelzen:
In short; you did not change the API at all, the new DTO class is not used by it.

Paulr756756:
**Further Changes**-

- Added Implicit and explicit operators in ActivityLogDto for better conversion performance.
- Removed direct exposure of ActivityLogEntry to the ActivityLogController by using dtos.
- More Refactoring.

Paulr756756:
@nielsvanvelzen I've made some changes can you review them when you get a chance?
Also, please let me know if i can further improve my code.

nielsvanvelzen:
Still used for WebSockets

Paulr756756:
@nielsvanvelzen Sorry for the inconvenience. I've modified ActivityLogWebSocketListener and ActivityLogEntryMessage classes.

barronpm:
It seems to me like the linked issue is somewhat outdated. Since the Activity log EF core migration `ActivityLogEntry` has been a DTO in all but name, so this PR could be accomplished by just renaming that class.

Paulr756756:
@barronpm I might be wrong but from what I've seen, `ActivityLogEntry` was being directly accessed by the DBContext and the Api Controllers as well. Shouldn't the main purpose of a DTO be to reduce that exposure?

barronpm:
We use DTOs to avoid exposing our db models directly through the API. `ActivityLogEntry` is our API model, while `ActivityLog` acts as the DB model. Before the activity database was migrated to EF Core, `ActivityLogEntry` was used both by the DB and API, thus the reasoning for the original issue.

Issue #8628, opened 2022-10-26
[Issue]: Jellyfin fails to recognize .sup (PGS) subtitle files, but plays them fine as .sub files


ArchemedIan:
### Please describe your bug

PGS Subtitle files with a .sup extension, which I believe is their default extension, are not detected as subtitle files by jellyfin.

(both the same file, with extension renamed) 

.sup file:
![image](https://user-images.githubusercontent.com/105951210/198051390-b1ab29fc-43c6-458b-801e-cc4c5a7bcb43.png)

.sub file:
![image](https://user-images.githubusercontent.com/105951210/198051634-92211d34-fa07-43c4-960c-753318755249.png)

plugins:

![image](https://user-images.githubusercontent.com/105951210/198052780-29678eff-9ac4-44bb-979b-42c18d4e7059.png)

### Jellyfin Version

Other

### if other:

10.8.5

### Environment

```markdown
- OS: debian (proxmox)
- Virtualization: none
- Clients: any
- Browser: any
- FFmpeg Version: jellyfin-ffmpeg5/unknown,now 5.1.2-2-bullseye amd64
- Playback Method: N/A
- Hardware Acceleration: none
- Plugins: https://user-images.githubusercontent.com/105951210/198052780-29678eff-9ac4-44bb-979b-42c18d4e7059.png
- Reverse Proxy: yes, nginx
- Base URL: none
- Networking: host
- Storage: btrfs raid0 array for media, btrfs single root
```

### Jellyfin logs

```shell
no log entry regarding the .sup files found
```

### FFmpeg logs

```shell
n/a
```

### Please attach any browser or client logs here

n/a

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

ArchemedIan:
https://github.com/jellyfin/jellyfin/blob/fca87bcd0c34fe738d7183269cf00c94da3d5dfa/MediaBrowser.Controller/Entities/BaseItem.cs#L58

ArchemedIan:
uh, bump?

crobibero:
Please try https://github.com/jellyfin/jellyfin/pull/8808

ArchemedIan:
> Please try #8808

seems to be working thanks guys

ivensfor:
Still unable to scan external subtitles in sup format

![2023-05-09_11-08-42](https://user-images.githubusercontent.com/8141652/236984242-1a037427-00cb-4fc3-9d49-a554f698ad8f.png)

[...]

PR #10537, opened 2023-11-06
Add support for .sup subtitle (#8808)

suyongxin4:
Fixes https://github.com/jellyfin/jellyfin/issues/8628

(cherry picked from commit 9a740344be757c2ab9a3f434295622573c39b181)

**Changes**
Cherry-pick the sup fix to 10.8.z so it gets delivered earlier

**Issues**


nielsvanvelzen:
This change is not suitable for backporting. It will be included in 10.9.

Issue #10417, opened 2023-10-16
[Issue]: Update user policy API endpoint does not validate required field 


mmedic:
### Please describe your bug

When calling POST `/Users/XXXXXXXXXXXXXXX/Policy` and `PasswordResetProviderId` is omitted, Exception is thrown and 500 is returned. This should be validated in the endpoint and 400 be returned.

[2023-10-16 19:42:31.263 +00:00] [ERR] [279] Jellyfin.Server.Middleware.ExceptionMiddleware: Error processing request. URL "POST" "/Users/XXXXXXXXXXXXXXX/Policy".
Microsoft.EntityFrameworkCore.DbUpdateException: An error occurred while saving the entity changes. See the inner exception for details.
 ---> Microsoft.Data.Sqlite.SqliteException (0x80004005): SQLite Error 19: 'NOT NULL constraint failed: Users.PasswordResetProviderId'.
   at Microsoft.Data.Sqlite.SqliteException.ThrowExceptionForRC(Int32 rc, sqlite3 db)
   at Microsoft.Data.Sqlite.SqliteDataReader.NextResult()
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteReader(CommandBehavior behavior)
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteReaderAsync(CommandBehavior behavior, CancellationToken cancellationToken)
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteDbDataReaderAsync(CommandBehavior behavior, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Storage.RelationalCommand.ExecuteReaderAsync(RelationalCommandParameterObject parameterObject, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Storage.RelationalCommand.ExecuteReaderAsync(RelationalCommandParameterObject parameterObject, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Update.ReaderModificationCommandBatch.ExecuteAsync(IRelationalConnection connection, CancellationToken cancellationToken)
   --- End of inner exception stack trace ---
   at Microsoft.EntityFrameworkCore.Update.ReaderModificationCommandBatch.ExecuteAsync(IRelationalConnection connection, CancellationToken cancellationToken)
[...]

PR #10553, opened 2023-11-08
Validate AuthenticationProviderId and PasswordResetProviderId

Chris-Codes-It:
**Changes**
POST /Users/{userId}/Policy returns BadRequest when PasswordResetProviderId is null or empty string

**Issues**
Issue [10417](https://github.com/jellyfin/jellyfin/issues/10417)


1337joe (approved):

crobibero:
Would adding `[Required(AllowEmptyStrings = false)]` to both `AuthenticationProviderId` and  `PasswordResetProviderId` in `UserPolicy` provide the same experience?

Chris-Codes-It:
> Would adding `[Required(AllowEmptyStrings = false)]` to both `AuthenticationProviderId` and `PasswordResetProviderId` in `UserPolicy` provide the same experience?

That's a fair comment; I did consider it however I didn't see any reference to `ModelState.IsValid` so I assumed no DataAnnotation binding had been wired-in. My bad. Have changed for both fields and updated the tests.

crobibero (approved):
https://learn.microsoft.com/en-us/aspnet/core/web-api/?view=aspnetcore-7.0#automatic-http-400-responses

The controller automatically validates models so the manual check isn't needed

1337joe (approved):

Bond-009 (approved):

PR #10566, opened 2023-11-11
Add option to use replaygain tags for audio

TelepathicWalrus:
Adds option to enable/disable scanning for replaygain tags. This is means that existing replaygain tags can be used instead of jellyfin scanning each track again. This may cause problems with consistency. If replaygain tags are enabled then they will override scanning for LUFS reducing computation, if they are not found and LUFS scan is still enabled then it will scan them tracks.

Frontend: https://github.com/jellyfin/jellyfin-web/pull/4984

### DIFF
--- a/MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
+++ b/MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
@@ -61,6 +61,9 @@ namespace MediaBrowser.Providers.MediaInfo
         [GeneratedRegex(@"I:\s+(.*?)\s+LUFS")]
         private static partial Regex LUFSRegex();
 
+        [GeneratedRegex(@"REPLAYGAIN_TRACK_GAIN:\s+(.*?)\s+dB")]
+        private static partial Regex ReplayGainTagRegex();
+
         /// <summary>
         /// Probes the specified item for metadata.
         /// </summary>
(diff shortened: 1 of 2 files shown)


MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
```diff
@@ -104,8 +107,50 @@ public partial class AudioFileProber
             }
 
             var libraryOptions = _libraryManager.GetLibraryOptions(item);
+            bool foundLUFSValue = false;
 
-            if (libraryOptions.EnableLUFSScan)
+            if (libraryOptions.UseReplayGainTags)
+            {
+                    using (var process = new Process()
+                {
```
crobibero, line 115:
```suggestion
                using (var process = new Process()
                {
```

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
```diff
@@ -61,6 +61,9 @@ public partial class AudioFileProber
         [GeneratedRegex(@"I:\s+(.*?)\s+LUFS")]
         private static partial Regex LUFSRegex();
 
+        [GeneratedRegex(@"REPLAYGAIN_TRACK_GAIN:\s+(.*?)\s+dB")]
```
crobibero, line 64:
```suggestion
        [GeneratedRegex(@"REPLAYGAIN_TRACK_GAIN:\s+([0-9.]+)\s+dB")]
```

Would this regex have the same desired group? I prefer having the group be specific in what it's trying to capture

TelepathicWalrus: Added "REPLAYGAIN_TRACK_GAIN:\s+-?([0-9.]+)\s+dB" to latest push. This allows negative gains too.

crobibero (approved):

1337joe (approved):

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
```diff
@@ -104,8 +107,50 @@ public partial class AudioFileProber
...
+                    {
+                        _logger.LogError(ex, "Error starting ffmpeg");
+
+                        throw;
+                    }
+
+                    using var reader = process.StandardError;
+                    var output = await reader.ReadToEndAsync(cancellationToken).ConfigureAwait(false);
+                    cancellationToken.ThrowIfCancellationRequested();
+                    MatchCollection split = ReplayGainTagRegex().Matches(output);
```
cvium, line 139: Why use Matches if you only expect one? The code would be a little clearer with just `.Match(output)`

Issue #10567, opened 2023-11-11
[Issue]: Calling DELETE /Users/{user_id}/PlayingItems/{item_id}/?positionTicks=8470000000 does nothing.


arabcoders:
### Please describe your bug

Using https://api.jellyfin.org/#tag/Playstate/operation/OnPlaybackStopped as reference, the API supposed to record user item play progress if `positionTicks` was supplied, However in reality it doesn't do anything and only issue the event. 

For example, choose any item in your database and try calling

```bash
$ curl -v -X DELETE 'http://localhost:8096/Users/{user_id}/PlayingItems/{item_id}?positionTicks=8470000000&ApiKey={api_id}'
```
you will get something like 

```bash
< HTTP/2 204
< alt-svc: h3=":443"; ma=2592000
< date: Sat, 11 Nov 2023 13:48:30 GMT
< server: Kestrel
< x-response-time-ms: 1
```

Which indicates that the API understood and accepted the request. However, nothing is actually done, and no progress has changed. As you can see from the attached photo. 

What i expect to happen is that jellyfin's API change the user positionTicks to that reported one at least according to the API docs. I work on WatchState tool which manage user watch state and progress and i managed to make it work for both emby and plex. Sadly, jellyfin proven to be a bit difficult to update user play progress, And i tried all the endpoints which deals with positionTick and non of them works as expected. 

Anyway i traced the codepath to this method 

https://github.com/jellyfin/jellyfin/blob/88873b6e9eedd1b600bd17c56ee66109d2ef566d/Emby.Server.Implementations/Session/SessionManager.cs#L911-L1019

I am not C# experienced dev it seems there is the code issue an event which leads to this code

https://github.com/jellyfin/jellyfin/blob/88873b6e9eedd1b600bd17c56ee66109d2ef566d/Jellyfin.Server.Implementations/Events/Consumers/Session/PlaybackStopLogger.cs#L39-L81

Reading this code, there is no mention of using the positionTicks, and it seems to only update the ActivityLog.

### Jellyfin Version

10.8.z

### Environment

```markdown
- Virtualization: linuxserver/jellyfin docker container.
```

### Jellyfin logs

```shell
[16:50:15] [INF] [42] Emby.Server.Implementations.Session.SessionManager: Playback stopped reported by app Syncer 10.8.12 playing Night Family. Stopped at 847000 ms
```

### Please attach any screenshots here

[...]

Issue #9614, opened 2023-04-07
[Issue]: Unable to set positionTicks for an item via API_KEY


luigi311:
### Please describe your bug

I am unable to set the positionTicks for an item for a user via the api key from the server. Using a delete request as following /Users/{userId}/PlayingItems/{itemId}?positionTicks={positionTicks} via a user token works but using that same request with a api key from the server does not. It also doesnt seem like theres any way to generate a user token via the api key so it doesnt look like theres a way to handle it that way either.

The logs report that the playback was stopped for the correct item at the correct time but since its with a api key it doesnt seem to handle it correctly even though the request has the user id in it so it should know who this belongs to.

### Jellyfin Version

10.8.0

### Environment

```markdown
- OS: Unraid
- Virtualization: Docker
- Clients: Python Script
- Browser: N/A
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- Plugins:
- Reverse Proxy: traefik
- Base URL:
- Networking: bridge
- Storage: local
```

### Jellyfin logs

```shell
[INF] [85] Emby.Server.Implementations.Session.SessionManager: Playback stopped reported by app "JellyPlex" "5.0.0" playing "Missing". Stopped at "1800000" ms
[INF] [137] Emby.Server.Implementations.Session.SessionManager: Playback stopped reported by app "JellyPlex" "5.0.0" playing "Just a Regular Irregular". Stopped at "900000" ms
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

luigi311:
This is still an issuee

luigi311:
this is still an issue

chuckgmthompson:
+1

luigi311:
This was resolved in 10.9

PR #10573, opened 2023-11-12
Add new API endpoint to view/update Item UserData

arabcoders:
**Changes**
This changes allow API users to update user data related to an item without the need to generate user specific token, right now, if you want to update for example, the played, favorites, likes, rating playback position ticks or any of the other related user data fields you have to send multiples requests. and some of them requires user access token `PlaybackPositionTicks` which is at least from my understanding is not possible to get unless you ask the user for username & password. So, this end point can serve to be entry point for all Item user data related requests and deprecate the other endpoints at some point.

This PR adds this new API endpoint `GET|POST /Users/{userId}/Items/{itemId}/UserData` 

**Issues**
This PR solves this #10567 and #9614


crobibero (changes requested):
Both of these endpoints allow any user to update any other user.
Example of how we validate access can be seen in `UserController::UpdateUser`

Additionally I would prefer to have the logic inside of the `UserDataManager` instead of in the controller.

ArabCoders:
> Both of these endpoints allow any user to update any other user. Example of how we validate access can be seen in `UserController::UpdateUser`
> 
> Additionally I would prefer to have the logic inside of the `UserDataManager` instead of in the controller.

Thanks for your comments, I'll Update the PR with the requested changes.

ArabCoders (commented):
I've updated the required changes, @crobibero Could you please review them and see if it's alright? Thank you.

MediaBrowser.Model/Dto/UserDataDto.cs
```diff
@@ -0,0 +1,43 @@
+#nullable disable
```
crobibero, line 1: Every property is nullable, so please remove this line

ArabCoders:
@crobibero Thanks again for pointing out the reflection usage, I thought it might be good idea to use reflection here to support future additions without having to change the code. Anyway the code is trivial to write, should the need arises in the future we can just refactor it again.

Thank you.

crobibero (approved):

ArabCoders:
I see CodeQL failing is it something i need to fix myself or is it from the project it self? i see references to `Mimetypes.cs` which i didn't interact with

MediaBrowser.Model/Dto/UserDataDto.cs
```diff
@@ -0,0 +1,40 @@
+namespace MediaBrowser.Model.Dto
+{
+    /// <summary>
+    /// Class UserDataDto extends UserItemDataDto to allow nullable members.
+    /// This change allow us to implement the new /Users/{UserId}/Items/{ItemId}/UserData endpoint.
+    /// This object allows the requestor to update all or specific user data fields without altering the non-nullable members state.
+    /// </summary>
+    public class UserDataDto : UserItemDataDto
```
nielsvanvelzen, line 8: This name is not descriptive enough. I'd suggest renaming it to "UpdateUserItemDataDto" which is similar to other models we already have like UpdateLibraryOptionsDto.

MediaBrowser.Model/Dto/UserDataDto.cs
```diff
@@ -0,0 +1,40 @@
+namespace MediaBrowser.Model.Dto
+{
+    /// <summary>
+    /// Class UserDataDto extends UserItemDataDto to allow nullable members.
+    /// This change allow us to implement the new /Users/{UserId}/Items/{ItemId}/UserData endpoint.
+    /// This object allows the requestor to update all or specific user data fields without altering the non-nullable members state.
```
nielsvanvelzen, line 6: This summary becomes the description for the API specification, we don't need developer notes for that. What benefit does this description have for client developers that use a generated SDK/apiclient?

nielsvanvelzen (changes requested):

nielsvanvelzen:
Should we also deprecate the old endpoints (markFavoriteItem etc.) in favor of this new endpoint?

MediaBrowser.Model/Dto/UserDataDto.cs
ArabCoders: Alright i have renamed them.

MediaBrowser.Model/Dto/UserDataDto.cs
ArabCoders: Updated to reflect the requested changes.

ArabCoders:
> Should we also deprecate the old endpoints (markFavoriteItem etc.) in favor of this new endpoint?

Unless we are sure the  Clients will be updated to reflect the change, i wouldn't deprecate them hastily.

nielsvanvelzen:
Deprecating means we annotate them with `[Obsolete]` but the endpoints still work. We do not currently have plans to remove deprecated parts of the API (perhaps in a major update like Jellyfin 11 which we also have no plans for yet).

nielsvanvelzen (approved):
✔️ API changes

Deprecating old API's could be done in a separate pull request

MediaBrowser.Model/Dto/UpdateUserItemDataDto.cs
```diff
@@ -0,0 +1,38 @@
+namespace MediaBrowser.Model.Dto
+{
+    /// <summary>
+    /// This is used by the api to get information about a item user data.
+    /// </summary>
+    public class UpdateUserItemDataDto : UserItemDataDto
```
Bond-009, line 6: Why does this inherit `UserItemDataDto`?

Bond-009 (changes requested):

MediaBrowser.Model/Dto/UpdateUserItemDataDto.cs
ArabCoders: I would have used `UserItemDataDto` directly if it was possible however due to the nature of possible fields not being sent it will alter user data. So, this class extends it and makes all fields nullable to not have unintended side effects. And i think the reviewers agree so far.

ArabCoders: If you want i can decouple the classes by duplicating the fields i feel it's really not good code case, however if that you want it's not a big deal

Bond-009: I think that would be best IMO

ArabCoders: Thats fine with me. Do you guys agree with this requested changes? @Shadowghost @nielsvanvelzen @crobibero seeing as you approved the previous code.

nielsvanvelzen: I've only reviewed the API specification which doesn't use inheritance anyway. There wouldn't be any difference on that part if this was changed, so no objection from me.

ArabCoders: @Bond-009 done.

MediaBrowser.Model/Dto/UpdateUserItemDataDto.cs
```diff
@@ -0,0 +1,77 @@
+#nullable disable
```
Bond-009, line 1: Remove this line and make the strings nullable

Bond-009 (changes requested):

Bond-009 (approved):

nielsvanvelzen:
Please rebase or merge master to fix failing CI

ArabCoders:
> Please rebase or merge master to fix failing CI

done

JPVenson (approved):

PR #10576, opened 2023-11-13
Serve all static filetypes from webroot

Jburso:
Allows all files, including files without extensions, to be served to clients.

This enables external acme tools to use the `wwwroot` for http-01 challenges, which don't contain file extensions.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -58,6 +58,7 @@
  - [HelloWorld017](https://github.com/HelloWorld017)
  - [ikomhoog](https://github.com/ikomhoog)
  - [iwalton3](https://github.com/iwalton3)
+ - [Jack Urso](https://github.com/Jburso)
  - [jftuga](https://github.com/jftuga)
  - [jmshrv](https://github.com/jmshrv)
  - [joern-h](https://github.com/joern-h)
--- a/Jellyfin.Server/Startup.cs
+++ b/Jellyfin.Server/Startup.cs
@@ -189,7 +189,11 @@ namespace Jellyfin.Server
                     mainApp.UseRobotsRedirection();
                 }
 
-                mainApp.UseStaticFiles();
+                mainApp.UseStaticFiles(new StaticFileOptions()
+                {
+                    ServeUnknownFileTypes = true
+                });
+
                 mainApp.UseAuthentication();
                 mainApp.UseJellyfinApiSwagger(_serverConfigurationManager);
                 mainApp.UseQueryStringDecoding();


Bond-009 (approved):

barronpm (approved):

cvium (changes requested):
I'm not a fan of this and neither is MS

https://learn.microsoft.com/en-us/aspnet/core/fundamentals/static-files?view=aspnetcore-8.0#non-standard-content-types

Jburso:
I've been thinking about ways to mitigate the security risk that Microsoft is referring to, but unfortunately they don't go into details of what the risk actually is.

As for the motivation for this PR, considering that Jellyfin doesn't currently have native ACME client support, I think it would be very beneficial to at least allow external clients to use its webroot for http-01 challenges. Yes, you can use a reverse-proxy, but you're still encrypting only half of the traffic. Behind the proxy is still plain HTTP. There is no reasonable way to have a full, end-to-end encrypted TLS tunnel from the Jellyfin server to a client -- a part of it will always be unencrypted. If we're talking about risk, I think that can be considered one as well. This can be mitigated by allowing auto-renewable certificates to exist directly on the server. 

In terms of fixes, I have two potential ideas.
1. Restrict `ServeUnknownFileTypes` to only files in `/.well-known` (or `/.well-known/acme-challenges`).
2. Specify a `FileExtensionContentTypeProvider()` with mappings for a null extension to `text/plain` (I don't know if this is allowed by the API).

Again, MS isn't clear with what the risk actually is, so it's not clear how to mitigate it.

cvium:
I think this issue is a combination of things. With Path Traversal, you could download any file that the webserver user has access to.

btw, we have never wanted to support something like LetsEncrypt.

Jburso:
I would hope that ASP.NET would not allow path traversal outside of the webroot by default regardless. And if it did, it would still allow you to access any files with a known extension... which is only of limited help.

I understand that Jellyfin doesn't want to support LetsEncrypt natively, but at the very least we can allow external tools to use our webroot to do it themselves. I think it's a relatively simple change with way more benefits (TLS everywhere, that's awesome!) than downsides.

crobibero:
The proper way to do this is to build an acme plugin instead of enabling non-recommended functionality to everyone.

JPVenson:
Hey @Jburso 
Thank you for the good idea but i discussed this with the team and we came to the conclusion that serving all files is just not in the scope of jellyfin. I would recommend discussion such changes first in our meta repro over at https://github.com/jellyfin/jellyfin-meta

PR #10600, opened 2023-11-18
Fix action building for .NET 8

Gauvino:
Switch docker building from CentOS 7 to 8 because of depracted package:
chrome_cb8dmNGMFw

Switch from Ubuntu 18 to 22 (LTS) and changed the package in consequence.


Gauvino:
The failed check is because I added backport with # in comment but I think the action don't want it

PR #10603, opened 2023-11-18
Refactor Dockerfile and build

Gauvino:
Rework of Dockerfile and build script

- Re-organized dockerfile for better syntax
- Changed Debian from 10 (Buster) to 12 (Bookworm)
- Changed Ubuntu from 18.04 (Bionic) to 22.04 (Jammy)
- Changed CentOS from 7 to CentOS Stream 9
- Changed Jammy Docker image to Built-in Jammy Microsoft .NET SDK image
- Upgrade version of Intel drivers ([23.39.27427.23](https://github.com/intel/compute-runtime/releases/tag/23.39.27427.23))
- Changed from debian to linux in debian/rules
- Changed jellyfin-ffmpeg5 to 6
- Changed GCC package version from 9 to 12
- Changed liblttng-ust from 0 to 1
- Changed libssl from 1.1 to 3
- Changed from using "Yum" to "Dnf" for CentOS and Fedora
- Arranged package installation
- Added ENV for ```${JELLYFIN_DATA_DIR} ${JELLYFIN_CACHE_DIR}```
- Added ```ENV NVIDIA_VISIBLE_DEVICES="all"``` 
- Added  ```ARG DOTNET_VERSION=8.0```
- Added "dnf clean all" and "rm -rf /var/cache/dnf" to the end of CentOS and Fedora Dockerfiles
- Added "apt-get clean", "apt-get autoremove", "rm -rf /var/lib/apt/lists/*" to the end of the Debian/Ubuntu Dockerfiles
- Removed --disable-parallel (previous problem for this has been fixed in .NET 8, now the build is way faster)
- Remove wget (only using curl now)
- Removed libcloog-isl-dev from ubuntu.armhf and ubuntu.arm64 (not available anymore)

Other PR :

 - jellyfin-web https://github.com/jellyfin/jellyfin-web/pull/5109
 - jellyfin-ffmpeg https://github.com/jellyfin/jellyfin-ffmpeg/pull/329
 - jellyfin-media-player https://github.com/jellyfin/jellyfin-media-player/pull/546
 - metapackages https://github.com/jellyfin/jellyfin-metapackages/pull/39
 - JavascriptSubtitlesOctopus https://github.com/jellyfin/JavascriptSubtitlesOctopus/pull/51

Supercedes https://github.com/jellyfin/jellyfin/pull/9625


crobibero (approved):

Gauvino:
@barronpm @Bond-009 @cvium @Shadowghost

deployment/Dockerfile.debian.amd64
```diff
@@ -14,7 +14,7 @@ RUN apt-get update -yqq \
   && apt-get install -yqq --no-install-recommends \
     debhelper gnupg devscripts build-essential mmv \
     libcurl4-openssl-dev libfontconfig1-dev libfreetype6-dev libssl-dev \
-    libssl1.1 liblttng-ust0
+    libssl3 liblttng-ust1
```
joshuaboniface, line 17: Please don't change anything about Debian packaging. This has a lot more implications than just changing these values imply.

joshuaboniface (changes requested):
I'm not in favour of this as-is.

Changing CentOS to 8 seems OK, though perhaps we should look at moving the base there to one of the derivatives.

The Debian/Ubuntu package changes I am not in favour of at all. There are too many implications to this to be done in a single PR right now, especially for core libraries like OpenSSL (changing that will break all older systems that link against openssl1.1 for instance).

Please submit the CentOS changes as one PR, and leave the Debian changes for me.

joshuaboniface:
@Gauvino All of our releases build from these configs. So, one set of configurations must support everything from Debian 10 and Ubuntu 20.04 through to present. Changing things like the SSL dependencies will break this. That's what I mean by it needs more work than a simply replace: all of these builds need to be thoroughly tested. As-is, I expect this will break nightly builds, hence my hesitation.

Gauvino:
> @Gauvino All of our releases build from these configs. So, one set of configurations must support everything from Debian 10 and Ubuntu 20.04 through to present. Changing things like the SSL dependencies will break this. That's what I mean by it needs more work than a simply replace: all of these builds need to be thoroughly tested. As-is, I expect this will break nightly builds, hence my hesitation.

@joshuaboniface As Debian 12 would become LTS in June and Ubuntu 23.10 LTS in July (6 months waiting), I'm thinking that it would not be a really bad idea at all cause Jellyfin 10.9 would not be released before that I think, a ton of work is pending in pull request and 10.9 would not release before merging all of this I think.

Else, it makes more sense now to move on to newer versions of the package and version cause Jellyfin can't be blocked with failed build and I don't think new package would break anything. .NET 8 is meant for new version, Debian 10 and Ubuntu 20 or better for the 10.8 build but not for the 10.9, I think it would be a good idea to go with the newer version for security and stability. Right now the action building is broken because of this only because the project is on .NET 8, I think you have multiple choice now: 
First one keeping build broken and wait for Debian and Ubuntu new LTS, second one is changing right now to new one and test everything out so the building action and Jellyfin Nightly could work great or third one reverts .NET 8 and change in .NET 7 waiting for the new LTS release of the OS.

Even so, 10.9 would be a really big release and a lot of changes planned, does changing to Debian 12 and Ubuntu 22 or 23 wouldn't be a problem at all now?

And does I can do something to test the build like building all of them by myself and test it out?

jumoog:
The next LTS Version is 24.04 not 23.10

Gauvino:
> The next LTS Version is 24.04 not 23.10

image

jumoog:
> > The next LTS Version is 24.04 not 23.10
> 
> image

The image only shows the release roadmap. The LTS release cycle is a new LTS version every 2 years in April.

14.04 / 16.04 / 18.04 / 20.04 / 22.04 so the next LTS release is 24.04 (April 2024). All releases in between are just STS releases.

Gauvino:
> > > The next LTS Version is 24.04 not 23.10
> > 
> > 
> > image
> 
> The image only shows the release roadmap. The LTS release cycle is a new LTS version every 2 years in April.
> 
> 14.04 / 16.04 / 18.04 / 20.04 / 22.04 so the next LTS release is 24.04 (April 2024). All releases in between are just STS releases.

Thanks you for the explication didn't knew that !

Gauvino:
image
Ok after so many change and optimization of the all the docker file I finally sucessfully build every image except 3 of them
image

joshuaboniface (approved):
Just going to YOLO and see how it goes. The changes seem OK from a functionality perspective, I'm just not 100% sure how everything will go when trying to run a full set of builds.

Gauvino:
> Just going to YOLO and see how it goes. The changes seem OK from a functionality perspective, I'm just not 100% sure how everything will go when trying to run a full set of builds.

If I ever made a mistake on building I would fix it immediately, no problem and from all the testing I made everything seems to build fine without an error.

Dockerfile
```diff
@@ -8,72 +8,68 @@ FROM node:20-alpine as web-builder
...
+ && curl -LO https://github.com/intel/intel-graphics-compiler/releases/download/igc-${IGC_VERSION}/intel-igc-core_${IGC_VERSION}_amd64.deb \
+         -LO https://github.com/intel/intel-graphics-compiler/releases/download/igc-${IGC_VERSION}/intel-igc-opencl_${IGC_VERSION}_amd64.deb \
+         -LO https://github.com/intel/compute-runtime/releases/download/${NEO_VERSION}/intel-level-zero-gpu_${LEVEL_ZERO_VERSION}_amd64.deb \
+         -LO https://github.com/intel/compute-runtime/releases/download/${NEO_VERSION}/intel-opencl-icd_${NEO_VERSION}_amd64.deb \
+         -LO https://github.com/intel/compute-runtime/releases/download/${NEO_VERSION}/libigdgmm12_${GMMLIB_VERSION}_amd64.deb \
  && dpkg -i *.deb \
  && cd .. \
  && rm -rf intel-compute-runtime \
- && apt-get remove gnupg wget -y \
+ && apt-get remove gnupg apt-transport-https -y \
```
Shadowghost, line 55: That package should not be needed anymore in bookworm

Dockerfile
```diff
@@ -8,72 +8,68 @@ FROM node:20-alpine as web-builder
...
+
 # https://github.com/intel/compute-runtime/releases
-ARG GMMLIB_VERSION=22.0.2
-ARG IGC_VERSION=1.0.10395
-ARG NEO_VERSION=22.08.22549
-ARG LEVEL_ZERO_VERSION=1.3.22549
+ARG GMMLIB_VERSION=22.3.11.ci17747749
+ARG IGC_VERSION=1.0.15136.4
+ARG NEO_VERSION=23.35.27191.9
+ARG LEVEL_ZERO_VERSION=1.3.27191.9
```
Shadowghost, line 34: Might need another update, there was a new release some days ago.

Dockerfile
```diff
@@ -83,11 +79,9 @@ COPY --from=builder /jellyfin /jellyfin
 COPY --from=web-builder /dist /jellyfin/jellyfin-web
 
 EXPOSE 8096
-VOLUME /cache /config
+VOLUME ${JELLYFIN_DATA_DIR} ${JELLYFIN_CACHE_DIR}
```
Shadowghost, line 82: Does this work? I always thought only ARG works outside of RUN statements

Dockerfile.arm
```diff
@@ -15,53 +14,46 @@ RUN apk add curl git zlib zlib-dev autoconf g++ make libpng-dev gifsicle alpine-
...
+ && apt-get update \
+ && apt-get install --no-install-recommends --no-install-suggests -y jellyfin-ffmpeg6 libssl-dev libfontconfig1 libfreetype6 vainfo libva2 locales \
+ && apt-get remove gnupg apt-transport-https -y \
  && apt-get clean autoclean -y \
  && apt-get autoremove -y \
  && rm -rf /var/lib/apt/lists/* \
- && mkdir -p /cache /config /media \
- && chmod 777 /cache /config /media \
+ && mkdir -p ${JELLYFIN_DATA_DIR} ${JELLYFIN_CACHE_DIR} \
+ && chmod 777 ${JELLYFIN_DATA_DIR} ${JELLYFIN_CACHE_DIR} \
```
Shadowghost, line 45: Same as above

Dockerfile.arm
```diff
@@ -72,11 +64,9 @@ COPY --from=builder /jellyfin /jellyfin
 COPY --from=web-builder /dist /jellyfin/jellyfin-web
 
 EXPOSE 8096
-VOLUME /cache /config
-ENTRYPOINT ["./jellyfin/jellyfin", \
-    "--datadir", "/config", \
-    "--cachedir", "/cache", \
-    "--ffmpeg", "/usr/lib/jellyfin-ffmpeg/ffmpeg"]
+VOLUME ${JELLYFIN_DATA_DIR} ${JELLYFIN_CACHE_DIR}
```
Shadowghost, line 67: Same as above

Dockerfile.arm64
```diff
@@ -66,11 +64,9 @@ COPY --from=builder /jellyfin /jellyfin
 COPY --from=web-builder /dist /jellyfin/jellyfin-web
 
 EXPOSE 8096
-VOLUME /cache /config
-ENTRYPOINT ["./jellyfin/jellyfin", \
-    "--datadir", "/config", \
-    "--cachedir", "/cache", \
-    "--ffmpeg", "/usr/bin/ffmpeg"]
+VOLUME ${JELLYFIN_DATA_DIR} ${JELLYFIN_CACHE_DIR}
```
Shadowghost, line 67: Same as above

Dockerfile
Gauvino: Actually it's the latest update available
image

Dockerfile
Gauvino: image
Does this respond to your question?

Dockerfile.arm
```diff
@@ -15,53 +14,46 @@ RUN apk add curl git zlib zlib-dev autoconf g++ make libpng-dev gifsicle alpine-
...
- libfreetype6 \
- vainfo \
- libva2 \
- locales \
- && apt-get remove gnupg -y \
+ && apt-get install --no-install-recommends --no-install-suggests -y ca-certificates gnupg curl apt-transport-https \
+ && curl -fsSL https://repo.jellyfin.org/jellyfin_team.gpg.key | gpg --dearmor -o /etc/apt/trusted.gpg.d/debian-jellyfin.gpg \
+ && curl -fsSL https://keyserver.ubuntu.com/pks/lookup?op=get\&search=0x6587ffd6536b8826e88a62547876ae518cbcf2f2 | gpg --dearmor -o /etc/apt/trusted.gpg.d/ubuntu-jellyfin.gpg \
+ && echo "deb [arch=$( dpkg --print-architecture )] https://repo.jellyfin.org/$( awk -F'=' '/^ID=/{ print $NF }' /etc/os-release ) $( awk -F'=' '/^VERSION_CODENAME=/{ print $NF }' /etc/os-release ) main" | tee /etc/apt/sources.list.d/jellyfin.list \
+ && echo "deb http://ppa.launchpad.net/ubuntu-raspi2/ppa/ubuntu bionic main">> /etc/apt/sources.list.d/raspbins.list \
```
Gauvino, line 37: @joshuaboniface @Shadowghost Does this repo is still needed ?

Dockerfile
Gauvino: Just so you can see when building image, in exemple it's the arm one 
image

Gauvino:
I checked if everything was fine after removing and build successfuly all images 
image

Dockerfile
Shadowghost: My bad I thought of the wrong repo (https://github.com/intel/media-driver)

Dockerfile.arm
Shadowghost: I think this can be removed since rpi HWA was removed.

Gauvino: Removed

Gauvino:
@Shadowghost @crobibero Could you look at this when you have time please. Thanks

Gauvino:
Made not to many changed just kinda optimise the dockerfile another time for more lightweight image with this I could gained about 200-300MB on some and for the majority 100MB so it's cool and I checked if every image could be build 
image
image

WizzardHub (approved):
lgtm ! <3

Gauvino:
Should I remake the PR with only 1 commit ?

crobibero (approved):

Gauvino:
I see they're still some error due to package missing I will fix this this weekend cause I don't have any pc right now ! @crobibero @joshuaboniface

PR #10613, opened 2023-11-21
Fixed disposable not being called

JPVenson:
Fixes memory leak that will be created because IDisposable will not be called on Closing a websocket.

**Changes**

**Issues**


crobibero (approved):

MediaBrowser.Controller/Session/SessionInfo.cs
```diff
@@ -391,6 +391,11 @@ public void Dispose()
                     _logger.LogDebug("Disposing session controller synchronously {TypeName}", disposable.GetType().Name);
                     disposable.Dispose();
                 }
+                else if (controller is IAsyncDisposable disposableAsync)
+                {
+                    _logger.LogDebug("Disposing session controller asynchronously {TypeName}", disposableAsync.GetType().Name);
+                    disposableAsync.DisposeAsync().GetAwaiter().GetResult();
+                }
```
Bond-009, line 398: This isn't needed, they should all implement `IDisposable`

JPVenson: Yes however making assumptions that one interface _should_ implement another while not declared was the underlying issue. This was added to ensure the same behaviour as `IAsyncDisposable` so future types will not have the same issue that they only implement either and not get called.

JPVenson: I have an alternative-

barronpm: What do you mean by "pulling the `IAsyncDispoable` usage up"?

If I'm reading this correctly, the changes you made to `DisposeAsync()` were what fixed the problem, and the [docs](https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync) clearly warn against implementing `IAsyncDisposable` without `IDisposable`, but the other way is fine. Put simply, not checking for `IDisposable` in `DisposeAsync` is a bug in this class, not implementing `IDisposable` in a session controller that needs it is a bug there.

JPVenson: > What do you mean by "pulling the `IAsyncDispoable` usage up"?
> 
> If I'm reading this correctly, the changes you made to `DisposeAsync()` were what fixed the problem, and the [docs](https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync) clearly warn against implementing `IAsyncDisposable` without `IDisposable`, but the other way is fine. Put simply, not checking for `IDisposable` in `DisposeAsync` is a bug in this class, not implementing `IDisposable` in a session controller that needs it is a bug there.

I am reading the warning but it does not sense to me to be honest:

> If you implement the [IAsyncDisposable](https://learn.microsoft.com/en-us/dotnet/api/system.iasyncdisposable) interface but not the [IDisposable](https://learn.microsoft.com/en-us/dotnet/api/system.idisposable) interface, your app can potentially leak resources. If a class implements [IAsyncDisposable](https://learn.microsoft.com/en-us/dotnet/api/system.iasyncdisposable), but not [IDisposable](https://learn.microsoft.com/en-us/dotnet/api/system.idisposable), and a consumer only calls Dispose, your implementation would never call DisposeAsync. This would result in a resource leak.

If the class only implements `IAsyncDisposable`, it implements the `DisposeAsync` method, but not the Dispose method so how would ` and a consumer only calls Dispose` on that class? How would that leak resources? 

If your Disposing method requires await, then you are more or less forced to implement `IAsyncDisposable` or shim them with `task.GetAwaiter().GetResult()` which is an anitpattern as well. We Require async disposal here so we should only need to implement that method. This is the first time i heard that only implementing IAsyncDisposable should be avoid, i only ever saw it for legacy reasons.

JPVenson: And yes by pulling up i mean, getting rid of the "IDisposable.Dispose" implementation and refactoring all callers to use "IAsyncDisposable.DisposeAsync" instead, which i already did on my PC and it works out fine

barronpm: The warning tells you to implement `IDisposable` as well as `IAsyncDisposable` in case the caller only checks for `IDisposable`, in which case `DisposeAsync` is never called. It's ok not to implement both if you know that `DisposeAsync` will be called, ie a DI service. SessionManager is in charge of disposal, so I'm fine with only implementing `IAsyncDisposable` here, but it still needs to check for `IDisposable` since session controllers could implement either.

JPVenson: Ah ok, i mean that statement is just misleading as it would always apply for sync/async interfaces, IEnumerable/IAsyncEnumerable etc etc... 

Yes, the benefit would be that we can await tasks from IAsyncDisposable and just run IDisposable in the Sessions IAsyncDisposable call, instead of wrapping the async call in the Sessions.IDisposable .

Will update the PR soon to use that.

Shadowghost (approved):

MediaBrowser.Controller/Session/ISessionManager.cs
```diff
@@ -111,7 +111,8 @@ public interface ISessionManager
         /// Reports the session ended.
         /// </summary>
         /// <param name="sessionId">The session identifier.</param>
-        void ReportSessionEnded(string sessionId);
+        /// <returns>Task.</returns>
+        ValueTask ReportSessionEnded(string sessionId);
```
cvium, line 115: Why ValueTask?

JPVenson: Because the AsyncDisposable may or may not return a task, for this usecase ValueTask was made.

JPVenson:
This will 100% break the recently merged DLNA plugin

cvium:
> This will 100% break the recently merged DLNA plugin @Bond-009

Why will it break?

JPVenson:
> > This will 100% break the recently merged DLNA plugin @Bond-009
> 
> Why will it break?

Because i originally changed the code myself to make it work again but that code was moved to the plugin, so per my latest update i removed that DLNA code as well. If (want i assume) he took that code and put it into the plugin, it needs changing.

servimo:
If I downgrade jellyfin, DLNA will appears?

servimo:
Sorry, wrong post.

JPVenson:
Apart from the issues with the DLNA caller, the change itself and the immediate (left) code is rather small. If you still consider the DLNA stuff, i would agree that its a bigger change. Looking at the PR now, its a better change i would say.

barronpm:
The changes you made to the method signatures affect the plugin API, which is why I said that. Based on a quick search though, it doesn't look like any of our plugins are affected (other than dlna) so I guess it's fine.

JPVenson:
Any change in Any public interface changes the plugin api if you want to see it that way xD.

MarkCiliaVincenti (approved):

Bond-009 (approved):

crobibero (approved):

PR #10623, opened 2023-11-25
[WIP] Move UserItemData to EF Core 

martimarkov:
**Changes**
Migrate the UserItemData entity to EF Core. 
Introduced a new LegacyEmbyDbContext to contain all Entities in the same DB. After we move all of them to EF Core it can be removed and merged to JellyfinDbContext. 

**Issues**

**TODO:**
Remove .Result from sync functions.


cvium:
.Result is banned. You must use GetAwaiter().GetResult() instead

Jellyfin.Server.Implementations/UserData/UserDataManager.cs
```diff
@@ -104,9 +104,9 @@ private async Task<UserItemData> GetUserDataAsync(User? user, BaseItem item)
     {
         ArgumentNullException.ThrowIfNull(user);
         ArgumentNullException.ThrowIfNull(item);
-        var dbContext = await _provider.CreateDbContextAsync().ConfigureAwait(false);
+        var dbContext = await _provider.CreateDbContextAsync().ConfigureAwait(true);
```
cvium, line 107: No, you almost never want to set it to true. And it's a no-op anyway. We are not hosting Jellyfin anywhere that has a synchronization context

Jellyfin.Server.Implementations/UserData/UserDataManager.cs
```diff
@@ -0,0 +1,273 @@
...
+
+    public void SaveUserData(Guid userId, BaseItem item, Data.Entities.UserItemData userData, UserDataSaveReason reason, CancellationToken cancellationToken)
+    {
+        var user = _userManager.GetUserById(userId);
+        SaveUserData(user, item, userData, reason, cancellationToken);
+    }
+
+    public void SaveUserData(User? user, BaseItem item, Data.Entities.UserItemData userData, UserDataSaveReason reason, CancellationToken cancellationToken)
+    {
+        SaveUserDataAsync(user, item, userData, reason, cancellationToken).ConfigureAwait(true);
```
JPVenson, line 100: same here

Jellyfin.Server.Implementations/UserData/UserDataManager.cs
```diff
@@ -0,0 +1,273 @@
...
+
+    private async Task<UserItemData> GetUserDataAsync(User? user, BaseItem item)
+    {
+        ArgumentNullException.ThrowIfNull(user);
+        ArgumentNullException.ThrowIfNull(item);
+        var dbContext = await _provider.CreateDbContextAsync().ConfigureAwait(true);
+        UserItemData? userItemData;
+        await using (dbContext.ConfigureAwait(true))
+        {
+            userItemData = await dbContext.UserDatas.FindAsync(new object[] { user.Id, item.Id.ToString() }).ConfigureAwait(true) ?? await CreateUserDataInternalAsync(user, item, dbContext).ConfigureAwait(true);
```
JPVenson, line 111: Creating a user on request does not sound like a good idea to me

Jellyfin.Server.Implementations/UserData/UserDataManager.cs
```diff
@@ -0,0 +1,273 @@
...
+    public UserItemData GetUserData(Guid userId, BaseItem item)
+    {
+        var user = _userManager.GetUserById(userId);
+
+        return GetUserData(user, item);
+    }
+
+    public UserItemData GetUserData(User? user, BaseItem item)
+    {
+        return GetUserDataAsync(user, item).GetAwaiter().GetResult();
```
JPVenson, line 140: I would either deprecate the sync function or make a parallel sync `GetUserDataAsync` one

Jellyfin.Server.Implementations/UserData/UserDataManager.cs
martimarkov: Fixed

Emby.Server.Implementations/ApplicationHost.cs
```diff
@@ -609,8 +606,18 @@ await using (jellyfinDb.ConfigureAwait(false))
                 }
             }
 
+            var legacyEmbyDb = await Resolve<IDbContextFactory<LibraryDbContext>>().CreateDbContextAsync().ConfigureAwait(false);
```
crobibero, line 609:
```suggestion
            var legacyLibraryDb = await Resolve<IDbContextFactory<LibraryDbContext>>().CreateDbContextAsync().ConfigureAwait(false);
```

Jellyfin.Data/Jellyfin.Data.csproj
```diff
@@ -39,11 +39,16 @@
...
   </ItemGroup>
 
   <ItemGroup>
     <Compile Include="..\SharedVersion.cs" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageVersion Update="Microsoft.EntityFrameworkCore.Abstractions" Version="8.0.0" />
+  </ItemGroup>
+
```
crobibero, line 53: This should be removed, veresions are managed in `Directory.Package.props`

Jellyfin.Server.Implementations/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -37,4 +37,30 @@ public static IServiceCollection AddJellyfinDbContext(this IServiceCollection se
 
         return serviceCollection;
     }
+
+    /// <summary>
+    /// Adds the <see cref="IDbContextFactory{TContext}"/> interface to the service collection with second level caching enabled.
+    /// </summary>
+    /// <param name="serviceCollection">An instance of the <see cref="IServiceCollection"/> interface.</param>
+    /// <returns>The updated service collection.</returns>
+    public static IServiceCollection AddLegacyEmbyDbContext(this IServiceCollection serviceCollection)
```
crobibero, line 46:
```suggestion
    public static IServiceCollection AddLegacyLibraryDbContext(this IServiceCollection serviceCollection)
```

Jellyfin.Server.Implementations/Migrations/DesignTimeLegacyEmbyDbFactory.cs
```diff
@@ -0,0 +1,20 @@
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Design;
+
+namespace Jellyfin.Server.Implementations.Migrations
+{
+    /// <summary>
+    /// The design time factory for <see cref="LibraryDbContext"/>.
+    /// This is only used for the creation of migrations and not during runtime.
+    /// </summary>
+    internal class DesignTimeLegacyEmbyDbFactory : IDesignTimeDbContextFactory<LibraryDbContext>
```
crobibero, line 10:
```suggestion
    internal class DesignTimeLegacyLibraryDbFactory : IDesignTimeDbContextFactory<LibraryDbContext>
```

Jellyfin.Server.Implementations/UserData/UserDataManager.cs
```diff
@@ -0,0 +1,273 @@
...
+
+        var keys = item.GetUserDataKeys();
+
+        var dbContext = await _provider.CreateDbContextAsync(cancellationToken).ConfigureAwait(false);
+        await using (dbContext.ConfigureAwait(false))
+        {
+            foreach (var key in keys)
+            {
+                dbContext.UserDatas.Update(userData);
+            }
```
crobibero, line 77: The looks like it's saving the same userData over and over?

MediaBrowser.Providers/Chapters/ChapterManager.cs
```diff
@@ -2,25 +2,52 @@
...
-        private readonly IItemRepository _itemRepo;
+        private readonly IDbContextFactory<LibraryDbContext> _provider;
+        private readonly ILogger<ChapterManager> _logger;
 
-        public ChapterManager(IItemRepository itemRepo)
+        public ChapterManager(IDbContextFactory<LibraryDbContext> provider, ILogger<ChapterManager> logger)
         {
-            _itemRepo = itemRepo;
+            _provider = provider;
+            this._logger = logger;
```
crobibero, line 23:
```suggestion
            _logger = logger;
```

We don't use `this`

MediaBrowser.Providers/Chapters/ChapterManager.cs
```diff
@@ -2,25 +2,52 @@
...
         }
 
         /// <inheritdoc />
-        public void SaveChapters(Guid itemId, IReadOnlyList<ChapterInfo> chapters)
+        public async void SaveChapters(Guid itemId, IReadOnlyList<ChapterInfo> chapters)
```
crobibero, line 27: Avoid `async void`

martimarkov: Can I ask why?

JPVenson: It swallows exceptions and is considered a dirty hack thats only existance is allowed on eventhandlers where there is no other option.

martimarkov: The exception part makes sense. I’ll fix it in the evening. 🙂

martimarkov: @JPVenson Would consider `async Task` better?

JPVenson: Yes that also means you have to await the task at the caller

Jellyfin.Server.Implementations/Extensions/ServiceCollectionExtensions.cs
martimarkov: Renamed

Jellyfin.Data/Jellyfin.Data.csproj
martimarkov: Removed

Emby.Server.Implementations/ApplicationHost.cs
```diff
@@ -608,8 +605,18 @@ await using (jellyfinDb.ConfigureAwait(false))
                 }
             }
 
+            var libraryDb = await Resolve<IDbContextFactory<LibraryDbContext>>().CreateDbContextAsync().ConfigureAwait(false);
+            await using (libraryDb.ConfigureAwait(false))
+            {
+                if ((await libraryDb.Database.GetPendingMigrationsAsync().ConfigureAwait(false)).Any())
+                {
+                    Logger.LogInformation("There are pending EFCore migrations in the database. Applying... (This may take a while, do not stop Jellyfin)");
```
crobibero, line 613: Probably want to specify what database the migrations are for since the same line is logged right above

Jellyfin.Server.Implementations/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -37,4 +37,30 @@ public static IServiceCollection AddJellyfinDbContext(this IServiceCollection se
...
+    public static IServiceCollection AddLibraryDbContext(this IServiceCollection serviceCollection)
+    {
+        serviceCollection.AddEFSecondLevelCache(options =>
+            options.UseMemoryCacheProvider()
+                .CacheAllQueries(CacheExpirationMode.Sliding, TimeSpan.FromMinutes(10))
+                .DisableLogging(true)
+                .UseCacheKeyPrefix("EF_")
+                // Don't cache null values. Remove this optional setting if it's not necessary.
+                .SkipCachingResults(result =>
+                    result.Value is null || (result.Value is EFTableRows rows && rows.RowsCount == 0)));
```
crobibero, line 55: We don't need to add the cache services twice

I would either stick the `LibraryDbContext` part into `AddJellyfinDbContext`, or pull `AddEFSecondLevelCache` to a separate function

Emby.Server.Implementations/Library/LibraryManager.cs
```diff
@@ -1483,6 +1495,18 @@ private void SetTopParentIdsOrAncestors(InternalItemsQuery query, IReadOnlyColle
             query.Parent = null;
         }
 
+        [Obsolete("LibraryManager.GetChapters is deprecated, please use ChapterManager.GetChapters instead.")]
```
crobibero, line 1498: The `[Obsolete]` attributes should be added at the interface level, nothing directly calls theses functions

Jellyfin.Server.Implementations/UserData/UserDataManager.cs
```diff
@@ -0,0 +1,262 @@
...
+
+    private async Task<UserItemData> GetUserDataAsync(User? user, BaseItem item)
+    {
+        ArgumentNullException.ThrowIfNull(user);
+        ArgumentNullException.ThrowIfNull(item);
+        var dbContext = await _provider.CreateDbContextAsync().ConfigureAwait(false);
+        UserItemData? userItemData;
+        await using (dbContext.ConfigureAwait(false))
+        {
+            userItemData = await dbContext.UserDatas.FindAsync(new object[] { user.Id, item.Id.ToString() }).ConfigureAwait(true);
```
crobibero, line 112:
```suggestion
            userItemData = await dbContext.UserDatas.Where(u => u.UserId.Equals(user.Id) && u.Key == item.Id.ToString()).FirstOrDefaultAsync().ConfigureAwait(false);
```

Emby.Server.Implementations/ScheduledTasks/Tasks/ChapterImagesTask.cs
```diff
@@ -142,7 +138,7 @@ public async Task ExecuteAsync(IProgress<double> progress, CancellationToken can
 
                 try
                 {
-                    var chapters = _itemRepo.GetChapters(video);
+                    var chapters = _libraryManager.GetChapters(video, CancellationToken.None).GetAwaiter().GetResult();
```
JPVenson, line 141: Please await this task instead

Jellyfin.Server.Implementations/UserData/UserDataManager.cs
```diff
@@ -0,0 +1,262 @@
...
+
+        var keys = item.GetUserDataKeys();
+
+        var dbContext = await _provider.CreateDbContextAsync(cancellationToken).ConfigureAwait(false);
+        await using (dbContext.ConfigureAwait(false))
+        {
+            foreach (var key in keys)
+            {
+                userData.Key = key;
+                dbContext.UserDatas.Update(userData);
```
JPVenson, line 77: I am not sure this does the trick. This should only mark the object as to be updated, not store the state. So the last value might win.

Jellyfin.Server.Implementations/UserData/UserDataManager.cs
JPVenson: Sory, that was maybe ambiguous.
What i meant, remove this method and update all callers to use the async instead, or make a totally Sync function.

Jellyfin.Server.Implementations/UserData/UserDataManager.cs
```diff
@@ -0,0 +1,262 @@
...
+
+    [Obsolete("Use GetUserDataAsync instead.")]
+    public UserItemData GetUserData(User? user, BaseItem item)
+    {
+        return GetUserDataAsync(user, item).GetAwaiter().GetResult();
+    }
+
+    public UserItemDataDto GetUserDataDto(BaseItem item, User user)
+    {
+             var userData = GetUserDataAsync(user, item).GetAwaiter().GetResult();
```
JPVenson, line 134: Same here

Jellyfin.Server.Implementations/UserData/UserDataManager.cs
```diff
@@ -0,0 +1,262 @@
...
+             var dto = GetUserItemDataDto(userData);
+
+             item.FillUserDataDtoValues(dto, userData, null, user, new DtoOptions());
+             return dto;
+    }
+
+    public UserItemDataDto GetUserDataDto(BaseItem item, BaseItemDto itemDto, User user, DtoOptions options)
+    {
+             var userData = GetUserDataAsync(user, item).GetAwaiter().GetResult();
```
JPVenson, line 143: Same here

Jellyfin.Server.Implementations/UserData/UserDataManager.cs
```diff
@@ -0,0 +1,262 @@
...
+             LastPlayedDate = data.LastPlayedDate,
+             Key = data.Key
+         };
+     }
+
+    public List<UserItemData> GetAllUserData(Guid userId)
+    {
+        var user = _userManager.GetUserById(userId);
+
+        return GetAllUserDataAsync(user).GetAwaiter().GetResult();
```
JPVenson, line 171: Same here

Jellyfin.Server.Implementations/UserData/UserDataManager.cs
JPVenson: > There are 20+ usages of GetUserDataDto and they are all in non-async code.

Then make then async

JPVenson: My issue with that is we get this PR in and the follow up one never gets though and we have another mess in the code where we supposed to use cleaned it up.

Then just make the whole stack sync and reintroduce the async version later.

Shadowghost: I'd say EF migration is the main point, everything else can be tackled later.

Currently we are bound to SQLite anyways which is mostly sychronous, so switching everything to async would not provide much right now. In the long run it is still something we should adopt to, but I don't think this is the scope of this PR.

tests/Jellyfin.XbmcMetadata.Tests/Parsers/MovieNfoParserTests.cs
```diff
@@ -52,7 +52,7 @@ public MovieNfoParserTests()
                 .Returns(_testUser);
 
             var userData = new Mock<IUserDataManager>();
-            userData.Setup(x => x.GetUserData(_testUser, It.IsAny<BaseItem>()))
+            userData.Setup(x => x.GetUserDataAsync(_testUser, It.IsAny<BaseItem>()).GetAwaiter().GetResult())
                 .Returns(new UserItemData());
```
JPVenson, line 56: Use `ReturnsAsync`

Jellyfin.Server.Implementations/UserData/UserDataManager.cs
```diff
@@ -0,0 +1,272 @@
...
+
+        var keys = item.GetUserDataKeys();
+
+        var dbContext = await _provider.CreateDbContextAsync(cancellationToken).ConfigureAwait(false);
+        await using (dbContext.ConfigureAwait(false))
+        {
+            foreach (var key in keys)
+            {
+                userData.Key = key;
+                var existingUserDataItem = await dbContext.UserDatas.FindAsync(new object[] { user.Id, userData.Key }, cancellationToken).ConfigureAwait(false);
```
crobibero, line 75: I prefer to use `Where/FirstOrDefault` over `Find` since it is more readable

Jellyfin.Server.Implementations/UserData/UserDataManager.cs
```diff
@@ -0,0 +1,272 @@
...
+        await using (dbContext.ConfigureAwait(false))
+        {
+            userItemDatas = dbContext.UserDatas.Where(entity => entity.UserId.Equals(user.Id));
+        }
+
+        return userItemDatas.ToList();
+    }
+
+    /// <inheritdoc />
+    public bool UpdatePlayState(BaseItem item, UserItemData data, long? reportedPositionTicks)
```
crobibero, line 197: Indentation is off in this function

kiawizard:
Guys, please resume this very important work if possible, this is our only chance to solve terrible performance of large libraries by migrating from sqlite to mysql

JPVenson:
Is this still `[WIP]`? We usually only want to review done PRs.

JPVenson:
The question before putting a PR out of draft is
- Are __you__ happy with the code
- Is the code feature complete
If both are yes, its not WIP.

Emby.Server.Implementations/Library/LibraryManager.cs
```diff
@@ -1859,7 +1871,7 @@ public async Task UpdateImagesAsync(BaseItem item, bool forceUpdate = false)
                 {
                     try
                     {
-                        var index = item.GetImageIndex(img);
+                        var index = item.GetImageIndexAsync(img).ConfigureAwait(false);
```
cvium, line 1874: Isn't this missing an await?

oddstr13:
Re-targeted to new feature branch.
Removed WIP tag to indicate readiness for reviews.

There is one unaddressed review comment.
There are new merge conflicts again :see_no_evil: 

I've requested more reviews, both from @barronpm which has done work with efcore migration in the past, and the server team as a whole.

@martimarkov does this PR stand on its own? As in, is Jellyfin in a sane state with just this PR merged?

Jellyfin.Server.Implementations/UserData/UserDataManager.cs
oddstr13: @JPVenson brings up good points, but this PR is already massive, and is only going to become larger before getting merged into master.

With touching almost 100 files, and with about 3500 lines to review, it is already a lot of work keeping on top of merge conflicts.

It would be best to keep the files touched minimal, even at some detriment of overall code quality and performance, in order to make it easier to deal with this massive task, that we've been postponing for many years already due to the amount of work that «needs to happen at once».

The requested change sound like would be relatively easy to do standalone, in one or more PRs down the line.
As such, I'm marking this conversation as resolved.

Consider making a cleanup issue with details on the cleanup task, pointing back to this conversation if you think that is desirable. Could possibly be a «good first issue»?

Jellyfin.Server.Implementations/LibraryDbContext.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        // Configuration for each entity is in it's own class inside 'ModelConfiguration'.
+        // modelBuilder.ApplyConfigurationsForEntitiesInContext();
+        modelBuilder.ApplyConfigurationsFromAssembly(
+            Assembly.GetExecutingAssembly(),
+            t => t.GetInterfaces().Any(i =>
+                i.IsGenericType &&
+                i.GetGenericTypeDefinition() == typeof(IEntityTypeConfiguration<>) &&
+                typeof(ILibraryModel).IsAssignableFrom(i.GenericTypeArguments[0])));
+        // modelBuilder.ApplyConfigurationsFromAssembly(typeof(LibraryDbContext).Assembly);
```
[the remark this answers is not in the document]
oddstr13: @crobibero is this conversation resolved?

EraYaN:
I'd be a huge fan if we could use one database instead of two. Especially if we ever want to support say Postgres, having two is more of a faff than just having one (which is what would be expected).

What was the rationale to add a second database?

JPVenson:
#12798 Moves all data from library.db into jellyfin.db with EFCore

0xBA5E64:
With #12798 merged will this now follow, or is this now redundant?

JPVenson:
> @JPVenson did AN AMAZING JOB!!!

when we re-enable unstable builds, we will see about that :-)

Issue #2538, opened 2020-03-08
Improve migrations and add support for plugins


mark-monteiro:
- [ ] Load migrations via reflection instead of hard-coding them into the migration runner. A timestamp property will need to be added to each migration and the migration runner updated to sort the migrations by timestamp in order to define the order that the migrations are run.
- [ ] Prefix each migration file name with the same timestamp as used in the property. This will keep the migrations sorted in the correct order when displayed in the filesystem, IDEs etc.
- [ ] Update migration runner to also log the migration namespace along side the name when executing migrations. This will be helpful when executing migrations from multiple plugins.
- [ ] Migration running logic should be in `Emby.Server.Implementations` instead of `Jellyfin.Server`
- [x] Instantiate each migration using the dependency injection framework instead of manually. This way each routine can specify its own dependencies independently and it will also not be necessary to pass `ApplicationHost` or `ILogger` instances to the `Run()` method (#2900)

sevenrats:
@mark-monteiro what is the status of this issue?

mark-monteiro:
Have not worked on any of the remaining tasks here and do not have time any more to work on this.

I created this issue when the migration framework was first implemented and had some ideas of how it could be improved. I'm not sure how much demand there actually is for migration support in plugins, I don't think it's something that has really come up.  I would be okay if this was closed, I'll leave it up to the triage team.

Bond-009:
This looks easy enough

srcrist:
@Bond-009 Were you interested in working on this? Should it be moved to "todo"?

LezardSC:
Hi, I'm searching for a not too hard issue to start working on the project. Is this issue still valid? If yes, I would love to give it a try.

mark-monteiro:
@LezardSC I haven't been involved in the project in a while so I'm not 100% sure, but as far as I know this still hasn't been implemented. It might be a good idea to also check in with the [development chat room](https://jellyfin.org/contact) before starting, especially since it's your first issue in the project.

AdrianEdelen:
I think with the recent changes with EF Core and migration behavior and overall database overhaul this maybe isn't needed anymore? One of the features being actively developed is custom database access for plugins, which should allow plugins to have their own database. In my opinion this effectively covers this issues goal.

mark-monteiro:
[...]

PR #10624, opened 2023-11-25
Improve migrations and add support for plugins

j-griffiths:
**Changes**

- Add interfaces for pre and post startup migration routines to differentiate between them when using reflection.
- Load migrations via reflection.
- Use timestamp property to define order of migrations.
- Rename files with timestamp prefixes to retain order in filesystem.
- Update migration runner to also log the migration namespace along side the name when executing migrations.
- Move migration logic to Emby.Server.Implementations.Migrations.

**Issues**

Implements suggestions from #2538


Emby.Server.Implementations/Migrations/MigrationRunner.cs
```diff
@@ -1,67 +1,51 @@
...
-            typeof(Routines.ReaddDefaultPluginRepository),
-            typeof(Routines.MigrateDisplayPreferencesDb),
-            typeof(Routines.RemoveDownloadImagesInAdvance),
-            typeof(Routines.MigrateAuthenticationDb),
-            typeof(Routines.FixPlaylistOwner),
-            typeof(Routines.MigrateRatingLevels),
-            typeof(Routines.AddDefaultCastReceivers)
-        };
+            return AppDomain.CurrentDomain.GetAssemblies()
+                .SelectMany(s => s.GetTypes())
```
JPVenson, line 28: Please use `GetExportedTypes` instead

Emby.Server.Implementations/Migrations/MigrationRunner.cs
```diff
@@ -1,67 +1,51 @@
...
+                .Where(p => p.IsClass && !p.IsAbstract && p.GetInterfaces().Contains(preStartupRoutines ? typeof(IPreStartupMigrationRoutine) : typeof(IPostStartupMigrationRoutine)));
```
JPVenson, line 29: Please use `p.IsAssignableTo(interfaceType)` instead

Emby.Server.Implementations/Migrations/IMigrationRoutine.cs
```diff
@@ -1,6 +1,6 @@
 using System;
 
-namespace Jellyfin.Server.Migrations
+namespace Emby.Server.Implementations.Migrations
```
crobibero, line 3: These should all be in the `Jellyfin.Server.Implementations` namespace if being moved. We don't add to `Emby.Server.Implementations` unless we have a good reason to

Emby.Server.Implementations/Migrations/IMigrationRoutine.cs
```diff
@@ -17,6 +17,11 @@ internal interface IMigrationRoutine
         /// </summary>
         public string Name { get; }
 
+        /// <summary>
+        /// Gets the timestamp from when this migration was written.
+        /// </summary>
+        public string Timestamp { get; }
```
crobibero, line 23: might as well use a `long` for the timestamp?

crobibero (commented):
Please don't force-push, it basically resets the review process

JPVenson:
I am not fully convinced that it is a good idea to allow plugins to modify the database. I have seen some really stupid stuff that plugins did to the database and i dont think we should facilitate that with allowing to execute plugin migrations that could potentially ruin everything and corrupt the DB even.

Emby.Server.Implementations/Migrations/IMigrationRoutine.cs
j-griffiths: There are database migrations already in Jellyfin.Server.Implementations.Migrations, do you want to be able to differentiate between them and the ones being moved or just put it all in the same directory? Alternatively I could just not move them, I only did it as it was one of the suggestions.

Bond-009:
Closing as stale

PR #10630, opened 2023-11-27
Added a check for the ReplaceAllImages value before updating an images

axl1232:
This check should prevent images from being updated unless the Replace existing images checkbox is checked in the refresh metadata dialog

**Changes**

**Issues**

### DIFF
--- a/Emby.Server.Implementations/Images/BaseDynamicImageProvider.cs
+++ b/Emby.Server.Implementations/Images/BaseDynamicImageProvider.cs
@@ -60,6 +60,11 @@ namespace Emby.Server.Implementations.Images
                 return ItemUpdateType.None;
             }
 
+            if (!options.ReplaceAllImages)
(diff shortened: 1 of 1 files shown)


Emby.Server.Implementations/Images/BaseDynamicImageProvider.cs
```diff
@@ -60,6 +60,11 @@ public async Task<ItemUpdateType> FetchAsync(T item, MetadataRefreshOptions opti
                 return ItemUpdateType.None;
             }
 
+            if (!options.ReplaceAllImages)
+            {
+                return ItemUpdateType.None;
+            }
```
1337joe, line 66: This looks like it will only download images when replace images is checked. What about the case where there are no images already downloaded? I would expect missing images to be downloaded without needing to check the "Replace All Images" box.

PR #10639, opened 2023-11-28
Fix Folder Child Count

LePips:
Why Luke?! (see blame)

### Changes

Fixes child count for folders/libraries. We already have the immediate children (we don't want recursive) so just ... count them instead of some random number.

My dev library is very small so I don't know impact on large libraries, or I may also be missing something that should be addressed.

### Example Usage

In clients we can use this number to show an amount of placeholder items in a list/grid while items are being fetched lazily in the background.

### DIFF
--- a/Emby.Server.Implementations/Dto/DtoService.cs
+++ b/Emby.Server.Implementations/Dto/DtoService.cs
@@ -431,14 +431,7 @@ namespace Emby.Server.Implementations.Dto
 
         private static int GetChildCount(Folder folder, User user)
         {
-            // Right now this is too slow to calculate for top level folders on a per-user basis
-            // Just return something so that apps that are expecting a value won't think the folders are empty
-            if (folder is ICollectionFolder || folder is UserView)
-            {
-                return Random.Shared.Next(1, 10);
-            }
-
-            return folder.GetChildCount(user);
+            return folder.Children.Count();
         }
 
         private static void SetBookProperties(BaseItemDto dto, Book item)


Emby.Server.Implementations/Dto/DtoService.cs
```diff
@@ -431,14 +431,7 @@ private void AttachUserSpecificInfo(BaseItemDto dto, BaseItem item, User user, D
...
         private static int GetChildCount(Folder folder, User user)
         {
-            // Right now this is too slow to calculate for top level folders on a per-user basis
-            // Just return something so that apps that are expecting a value won't think the folders are empty
-            if (folder is ICollectionFolder || folder is UserView)
-            {
-                return Random.Shared.Next(1, 10);
-            }
-
-            return folder.GetChildCount(user);
```
JPVenson, line 441: It seems like this was the only usage of `Folder.GetChildCount(User)`. Should we remove the unused method?

thornbill:
Does this account for user access controls (i.e. some content not being visible to the current user)?

LePips:
No this doesn't, so silly me on a caffeine high. There are a couple ways to do this:

- remove the type check and continue onto the item's `GetChildCount`, which performs a query based on user access
- just do a `return folder.Children.Where(i => i.IsVisible(user)).Count();` since we already have the children

However, my dev library is really small so I would want to check against a sample library with thousands of items on my machine and my RPI4. Frankly, I'm ignorant on this stuff and only thought of this as an easy win - so obviously performance is desirable above a UI/UX nice-to-have and this can be closed.

PR #10660, opened 2023-11-30
Move Jellyfin.Networking

barronpm:
**Changes**
- Moves `Jellyfin.Networking` to src
- Moves networking code from `Emby.Server.Implementations` to `Jellyfin.Networking`
- Fixes some warnings in the aforementioned networking code
- Uses file-scoped namespaces for moved code

Changes are probably best reviewed commit-by-commit


Shadowghost (approved):

src/Jellyfin.Networking/HappyEyeballs/HttpClientExtension.cs
```diff
@@ -0,0 +1,119 @@
...
+SOFTWARE.
+*/
+
+using System.IO;
+using System.Net.Http;
+using System.Net.Sockets;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Jellyfin.Networking.HappyEyeballs;
```
JPVenson, line 33: Wut?

crobibero: https://en.wikipedia.org/wiki/Happy_Eyeballs

src/Jellyfin.Networking/HappyEyeballs/HttpClientExtension.cs
```diff
@@ -0,0 +1,119 @@
...
+        using var cancelIPv6 = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+        var tryConnectAsyncIPv6 = AttemptConnection(AddressFamily.InterNetworkV6, context, cancelIPv6.Token);
+
+        // GetAwaiter().GetResult() is used instead of .Result as this results in improved exception handling.
+        // The tasks have already been completed.
+        // See https://github.com/dotnet/corefx/pull/29792/files#r189415885 for more details.
+        if (await Task.WhenAny(tryConnectAsyncIPv6, Task.Delay(200, cancelIPv6.Token)).ConfigureAwait(false) == tryConnectAsyncIPv6 && tryConnectAsyncIPv6.IsCompletedSuccessfully)
+        {
+            await cancelIPv6.CancelAsync().ConfigureAwait(false);
+            return tryConnectAsyncIPv6.GetAwaiter().GetResult();
```
JPVenson, line 69: As its a `Task` not a `ValueTask` its fine to `await` them anyway even if already completed. So either make `AttemptConnection` return a `ValueTask` (which would be the correct usage for it) and/or await it here anyway.

src/Jellyfin.Networking/Udp/UdpServer.cs
```diff
@@ -0,0 +1,136 @@
...
+
+    /// <summary>
+    /// Starts the specified port.
+    /// </summary>
+    /// <param name="cancellationToken">The cancellation token to cancel operation.</param>
+    public void Start(CancellationToken cancellationToken)
+    {
+        _udpSocket.Bind(_endpoint);
+
+        _ = Task.Run(async () => await BeginReceiveAsync(cancellationToken).ConfigureAwait(false), cancellationToken).ConfigureAwait(false);
```
JPVenson, line 97: I would prefer to use `Task.Factory.StartNew(..., TaskCreationOptions.LongRunning)` to not clog up the scheduler

src/Jellyfin.Networking/Udp/UdpServer.cs
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    private async Task BeginReceiveAsync(CancellationToken cancellationToken)
+    {
+        while (!cancellationToken.IsCancellationRequested)
+        {
+            try
+            {
+                var endpoint = (EndPoint)new IPEndPoint(IPAddress.Any, 0);
+                var result = await _udpSocket.ReceiveFromAsync(_receiveBuffer, endpoint, cancellationToken).ConfigureAwait(false);
```
JPVenson, line 107: Ideally _receiveBuffer should not live on the Heap as it has no use there. What is the idea behind putting it there?

src/Jellyfin.Networking/Udp/UdpServer.cs
```diff
@@ -0,0 +1,136 @@
...
+                var text = Encoding.UTF8.GetString(_receiveBuffer, 0, result.ReceivedBytes);
```
JPVenson, line 108: Best would be to get an encoder once and keep it.

src/Jellyfin.Networking/HappyEyeballs/HttpClientExtension.cs
barronpm: All of this code is being moved w/ as few changes as possible to make it compile, so any further changes aren't relevant for this PR.

src/Jellyfin.Networking/Udp/UdpServer.cs
cvium: It's a standard array. It always lives on the heap.

src/Jellyfin.Networking/Udp/UdpServer.cs
JPVenson: That is what `TaskCreationOptions.LongRunning` does. It does not schedule the Task on the local Thread pool thread.

src/Jellyfin.Networking/HappyEyeballs/HttpClientExtension.cs
JPVenson: await should always be considered for 3 reasons:
- Change of implementation. When the underlying implementation changes this could be a real task
- Again, that why a ValueTask should be used here. ValueTask represent exactly this behavior.
- Syntactic correctness. If we are in an async method, we should always await tasks.

src/Jellyfin.Networking/Udp/UdpServer.cs
cvium: Only the first part of the code will be on that longrunning thread. Thankfully, it's very easy to verify :)

```
            var task =  Task.Factory.StartNew(async () =>
            {
                while (true)
                {
                    Console.WriteLine($"Before await. Is thread pool thread? {Thread.CurrentThread.IsThreadPoolThread}");
                    await Task.Delay(100);
                    Console.WriteLine($"After await. Is thread pool thread? {Thread.CurrentThread.IsThreadPoolThread}");
                }
            }, TaskCreationOptions.LongRunning);
            Console.ReadKey();
```

Output:

```
Before await. Is thread pool thread? False
After await. Is thread pool thread? True
Before await. Is thread pool thread? True
After await. Is thread pool thread? True
Before await. Is thread pool thread? True
After await. Is thread pool thread? True
```

src/Jellyfin.Networking/HappyEyeballs/HttpClientExtension.cs
cvium: I prefer to do what Stephen Toub says: https://github.com/dotnet/corefx/pull/29792/files#r189415885. He seems like a smart enough guy :)

JPVenson: Understandable, i read the whole discussion before starting this :-D

But i would argue that even readability in our, way less critical scenario is more important then 5 bytes and 1ms additional time. 
Tbo i dont know why he did not even considered using ValueTask

src/Jellyfin.Networking/Udp/UdpServer.cs
cvium: What do you mean by this? Encoding.UTF8 is static.

src/Jellyfin.Networking/HappyEyeballs/HttpClientExtension.cs
cvium: Can't use a ValueTask in Task.WhenAny though, so how would you wait for a result but only for 200ms?

JPVenson: As its just one task you want to wait for 200ms, i would just use `await tryConnectAsyncIPv6.WaitAsync(200).ConfigureAwait`

src/Jellyfin.Networking/Udp/UdpServer.cs
cvium: That is correct, but when you use async-await it's not long running, and as you can see by the sample code I provided, the continuation is now bound to the Thread pool.

src/Jellyfin.Networking/HappyEyeballs/HttpClientExtension.cs
cvium: There is no WaitAsync for ValueTask :)

JPVenson: Oh really? Then i have to apologise :-)

jumoog (approved):

Bond-009 (approved):

PR #10672, opened 2023-12-03
Quote bash paths to prevent word splitting

Erotemic:
This PR makes a modification to the bash build scripts to quote variables that could contain directories with spaces. This prevents bash from word splitting the arguments, and will allow those unholy souls who use spaces in their path names to run these scripts.

I restricted this to `SOURCE_DIR` and  `ARTIFACT_DIR`, as it looked like the other paths were guaranteed to not have spaces in them.

Note, did automate this change using regex search / replace, and while I did check the diff to make sure I didn't introduce anything silly, reviewers should double check that I didn't make a mistake. 

**Changes**
Add quotes to bash names to allow directories with spaces for native builds.

### DIFF
--- a/deployment/build.centos.amd64
+++ b/deployment/build.centos.amd64
@@ -6,7 +6,7 @@ set -o errexit
 set -o xtrace
 
 # Move to source directory
-pushd ${SOURCE_DIR}
+pushd "${SOURCE_DIR}"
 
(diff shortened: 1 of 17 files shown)


joshuaboniface (approved):
Looks correct.

Shadowghost (approved):

Bond-009 (approved):

PR #10673, opened 2023-12-03
Add --output-dir argument to `build.sh`

Erotemic:
Currently the build script running in native mode will write the output to "../bin" with respect to the CWD, which is where jellyfin is cloned to. I keep all of my code repos in the same folder, so I don't want jellyfin to write to the folder above it. At the same time I don't want to break existing default behavior. Thus, I've added the option `--output-dir`, which lets the user specify the prefix where "bin/<platform>/..." outputs are written to.

I've added the appropriate help docs. Other than that, the change is straightforward. 

**Changes**
Add `-o / --output-dir` argument to `build.sh`

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -9,12 +9,13 @@ set -o pipefail
 usage() {
     echo -e "build.sh - Build Jellyfin binary packages"
     echo -e "Usage:"
-    echo -e "  $0 -t/--type <BUILD_TYPE> -p/--platform <PLATFORM> [-k/--keep-artifacts] [-l/--list-platforms]"
+    echo -e "  $0 -t/--type <BUILD_TYPE> -p/--platform <PLATFORM> [-o/--output-dir <OUTPUT_DIR>] [-k/--keep-artifacts] [-l/--list-platforms]"
     echo -e "Notes:"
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -101,7 +109,11 @@ if [[ -z ${BUILD_TYPE} || -z ${PLATFORM} ]]; then
 fi
 
 export SOURCE_DIR="$( pwd )"
-export ARTIFACT_DIR="${SOURCE_DIR}/../bin/${PLATFORM}"
+DEFAULT_OUTPUT_DIR="${SOURCE_DIR}/../"
+if [[ "$OUTPUT_DIR" == "" ]]; then
```
joshuaboniface, line 113: Use `[[ -z ${OUTPUT_DIR} ]]; then` here instead.

joshuaboniface (changes requested):

build.sh
Erotemic, line 113:
```suggestion
if [[ -z "${OUTPUT_DIR}" ]]; then
```

Erotemic: Done

Gauvino:
Can be closed due to https://github.com/jellyfin/jellyfin/pull/11162

PR #10702, opened 2023-12-08
Convert CollectionType to use lowercase enum names

crobibero:
I tried multiple methods of using a lowercasing string enum converter but wasn't able to find a method that worked fully.

I would much rather keep the enum to be pascalcase, so alternative suggestions are welcome

### DIFF
--- a/Emby.Server.Implementations/Images/CollectionFolderImageProvider.cs
+++ b/Emby.Server.Implementations/Images/CollectionFolderImageProvider.cs
@@ -32,26 +32,26 @@ namespace Emby.Server.Implementations.Images
 
             switch (viewType)
             {
-                case CollectionType.Movies:
+                case CollectionType.movies:
                     includeItemTypes = new[] { BaseItemKind.Movie };
                     break;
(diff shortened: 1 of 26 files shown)


1337joe (approved):

Shadowghost (approved):

Bond-009:
https://github.com/dotnet/runtime/issues/74385

Maybe we could use this solution? https://github.com/dotnet/runtime/issues/31081#issuecomment-848697673

crobibero:
> https://github.com/dotnet/runtime/issues/74385
> 
> 
> 
> Maybe we could use this solution? https://github.com/dotnet/runtime/issues/31081#issuecomment-848697673
> 
> 

This doesn't update the api spec, only the actual output. I also tried adding the custom json serializers to the openapi generator and that caused issues when trying to use swagger directly

jnorris441:
Wait does this mean the API spec still has the pascal case values but the server will be sending lowercase? 

Was just bit by this using the Swift SDK which is generated from the stable openapi json.

PR #10710, opened 2023-12-12
Use `eof_action=pass` when overlaying subtitles

### LINKED ISSUE #10698: [Issue]: Use `eof_action=pass` when overlaying subtitles
### Please describe your bug

When overlaying subtitles on video, jellyfin currently uses the following options with the `overlay` filter (there are multiple occurances, I only linked one):
https://github.com/jellyfin/jellyfin/blob/0d5bbfda321f64ad78b5b74e7bdfbbbb8b64c7af/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs#L3978

Aside the redundancy (`repeatlast` overrides `eof_action` and `shortest` overrides `repeatlast`), this causes the output to be terminated as the end of one of the input streams is reached. In some cases, the subtitles don't have the same duration as the main video and may end before the video, which causes the resulting video stream to be terminated prematurely.

Instead, I propose using `eof_action=pass:repeatlast=0`, which continues transcoding the video if the subtitle stream ends earlier.
If the subtitle stream has a higher duration than the video, overlaying the subtitles is stopped once the end of the video stream is reached even without `shortest=1`, since that's the default behavior. The `overlay` filter will never create an output stream with a higher duration than the main input.

See also: https://ffmpeg.org/ffmpeg-filters.html#Options-for-filters-with-several-inputs-_0028framesync_0029

### Jellyfin Version

10.8.z

### if other:

_No response_

### Environment

_No response_

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

jkhsjdhjs:
The previous behavior using `eof_action=endall` and `shortest=1` would end the video stream if one of the input stream (video, subtitle) ends. In some cases the duration of the overlayed subtitles is shorter than the video stream, causing the output to end when the subtitles end and dropping the remaining video stream.

This commit changes this behavior so `eof_action=pass` is used instead, which continues passing the video stream through even if the subtitles end earlier [1]. `shortest=1` is also removed, as this option implies `eof_action=endall`.

If the subtitle stream has a higher duration than the video stream, the output will also end with the video stream without `shortest=1`, as the video stream is the primary input to the `overlay` filter.

Fix #10698

[1] https://ffmpeg.org/ffmpeg-filters.html#Options-for-filters-with-several-inputs-_0028framesync_0029

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -3343,7 +3343,7 @@ namespace MediaBrowser.Controller.MediaEncoding
                 // [0:s]scale=s=1280x720
                 var subSwScaleFilter = GetCustomSwScaleFilter(inW, inH, reqW, reqH, reqMaxW, reqMaxH);
                 subFilters.Add(subSwScaleFilter);
-                overlayFilters.Add("overlay=eof_action=endall:shortest=1:repeatlast=0");
+                overlayFilters.Add("overlay=eof_action=pass:repeatlast=0");
             }
 
             return (mainFilters, subFilters, overlayFilters);
@@ -3520,7 +3520,7 @@ namespace MediaBrowser.Controller.MediaEncoding
                     }
 
                     subFilters.Add("hwupload=derive_device=cuda");
-                    overlayFilters.Add("overlay_cuda=eof_action=endall:shortest=1:repeatlast=0");
+                    overlayFilters.Add("overlay_cuda=eof_action=pass:repeatlast=0");
                 }
             }
             else
@@ -3529,7 +3529,7 @@ namespace MediaBrowser.Controller.MediaEncoding
                 {
                     var subSwScaleFilter = GetCustomSwScaleFilter(inW, inH, reqW, reqH, reqMaxW, reqMaxH);
                     subFilters.Add(subSwScaleFilter);
-                    overlayFilters.Add("overlay=eof_action=endall:shortest=1:repeatlast=0");
+                    overlayFilters.Add("overlay=eof_action=pass:repeatlast=0");
                 }
             }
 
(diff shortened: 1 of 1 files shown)


Bond-009 (approved):

nyanmisaka (changes requested):
`overlay_vulkan` should also apply the same change.

https://github.com/jellyfin/jellyfin/blob/82527314c1e92cce1bcf38140bbab3f99beea560/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs#L4619

jkhsjdhjs:
> `overlay_vulkan` should also apply the same change.
> 
> https://github.com/jellyfin/jellyfin/blob/82527314c1e92cce1bcf38140bbab3f99beea560/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs#L4619

Good catch! I initially did the change on the `release-10.8.z` branch, which doesn't have the `overlay_vulkan` line. I just fixed it.

nyanmisaka:
> Good catch! I initially did the change on the `release-10.8.z` branch, which doesn't have the `overlay_vulkan` line. I just fixed it.

Found another one. Please remove the `:shortest=1` in this line.
https://github.com/jellyfin/jellyfin/blob/bcc376a3520dadd43b67f869cac9ede368368b1f/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs#L4459

jkhsjdhjs:
> Found another one. Please remove the `:shortest=1` in this line.
> 
> https://github.com/jellyfin/jellyfin/blob/bcc376a3520dadd43b67f869cac9ede368368b1f/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs#L4459

Done. I found two other occurances of the same line and removed the `shortest=1` there as well.

nyanmisaka (approved):

cvium (approved):

Issue #10730, opened 2023-12-15
[Issue]: Multiple Artists Split Delimeter missing comma, or make it configurable


Shabinder:
### Please describe your bug

In Music, Artists tag when contains multiple artists, most common practice is to add all by splitting with `, ` , which for some reason jellyfin doesnt split itself.

This has critical in having a usable artists tab in media-player clients, 
Pls add comma as a delimeter, or make delimeters configurable.

Ex: 
https://musicbrainz.org/doc/Style/Classical/Release_Artist

### Jellyfin Version

10.8.z

### Environment

```markdown
- OS: Ubuntu 20.04 LTS
- Linux Kernel: GNU/Linux 5.15.0-1049
- Virtualization: Docker
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Shabinder:
A feature to add custom delimeters in settings would be the best ideal way to create and manage user libraries.

Shabinder:
Not Stale

PR #10735, opened 2023-12-16
Delimeters : "," added to split multiple value fields like artists

Shabinder:
**Delimeters** : `,` and `&` added to split multiple value fields like artists

Fixing: https://github.com/jellyfin/jellyfin/issues/10730

Allowing Edits by maintainer to add any nitpicks if necessary.

### DIFF
--- a/MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
+++ b/MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
@@ -29,7 +29,7 @@ namespace MediaBrowser.MediaEncoding.Probing
 
         private const string ArtistReplaceValue = " | ";
 
-        private readonly char[] _nameDelimiters = { '/', '|', ';', '\\', '&', ',' };
+        private readonly char[] _nameDelimiters = { '/', '|', ';', '\\', ',' };
 
         private readonly ILogger _logger;
         private readonly ILocalizationManager _localization;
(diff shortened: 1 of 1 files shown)


cvium:
Seems risky to me. & is also a no-go unless you believe Simon & Garfunkel are two separate artists.

Shadowghost:
I still think that us having this list of "split characters" is a bad thing. APE2 and ID3v2.4 both support multiple artist fields in tags.

Add ing `&` as a general delimiter is a bad idea, there are enough artists and groups out there which use it legitimately.

Shabinder:
How about having a **comma** only then ?
It is legitimately used by many taggers...
https://musicbrainz.org/doc/Style/Classical/Release_Artist

Also maybe we can provide custom delimeters using environment Variables or make configurable in Settings ?

Shabinder:
> Seems risky to me. & is also a no-go unless you believe Simon & Garfunkel are two separate artists.

Agree & is a risky, hence removed that, and only added a `,` which we can decide since its vastly used by taggers, lmk if any issue with that ?

TheMelmacian:
There are also artists with a `,` in the name.  On a quick search I found [Everything, Now!](https://rateyourmusic.com/artist/everything-now), [Go Robot, Go!](https://rateyourmusic.com/artist/go_robot__go_) and [Wow, Owls!](https://rateyourmusic.com/artist/wow-owls).

As Shadowghost mentioned, all modern tagging standards (APE2, ID3v2.4, Vorbis-Tags etc.) support multiple artist and genre fields. It's already a pain in the ass to maintain the exceptions for the existing delimiters, so I vote against it.

Shabinder:
understandable, lets make this user configurable as most other player do.
That will be the safest choice.

gnattu:
Replaced by #12385

Issue #10529, opened 2023-11-04
[Issue]: Collections missing artwork


mario-g98:
### Please describe your bug

I started uploading my media in the previous days so I don't know if this is a recent problem, but a bug number of automatically created collections by TMDb Box Sets Plugin don't get any artwork downloaded. Even if I click on Edit Images and then Search, no singe image is feched. 

I created a [post](https://forum.jellyfin.org/t-collections-don-t-have-thumbnails) in the forum, so apparently I'm not the only one with this problem.

### Jellyfin Version

10.8.z

### if other:

10.8.11

### Environment

```markdown
- OS: Unraid
```

### Jellyfin logs

```shell
This is the log when clicking on Refresh metadata:

[2023-11-04 21:10:26.110 +01:00] [INF] [76] Jellyfin.Plugin.KodiSyncQueue.EntryPoints.LibrarySyncNotification: Started library sync
[2023-11-04 21:10:26.168 +01:00] [INF] [76] Jellyfin.Plugin.KodiSyncQueue.EntryPoints.LibrarySyncNotification: Library Sync: Updated 1 items
[2023-11-04 21:10:26.168 +01:00] [INF] [76] Jellyfin.Plugin.KodiSyncQueue.EntryPoints.LibrarySyncNotification: Finished library sync, taking "00:00:00.0581887"
```

### Please attach any screenshots here

![Screenshot from 2023-11-04 21-12-29](https://github.com/jellyfin/jellyfin/assets/49582442/dffd24ad-bea0-4fdd-9829-eda972f8cf32)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

fogolin:
I'm also facing the same issue here. Even checked the TMDB to see if they were missing there as well, but they seem fine. One example is the John Wick collection, https://www.themoviedb.org/collection/404609-john-wick-collection, which is missing the artwork on my Jellyfin server.

unoriginaltom:
Seeing the same issue on my side with those created by the TMDB Box Sets plugin. Weirdly, though, it seemed to work for a single collection, but none of the others.

VigibotDev:
Same problem here.
Note few picture are downloaded of few collection, but vast majority won't.
What is surprising is that the problem is repeatable, collection of movie "xXx" for example always recovers the image correctly, but vast majority won't

mikewesten:
[...]

PR #10737, opened 2023-12-17
Make TMDb api key configurable, fix missing/wrong image urls

1337joe:
The API key is added to the config file but not the user-facing plugin page. Most users should not mess with the key, but as discussed in #10529 some users like tracking their own API usage or dislike that using the Jellyfin key includes their statistics where someone else could possibly see.

To change the key open the TMDb plugin settings and hit save, then open `Jellyfin.Plugin.Tmdb.xml` (at `~/.local/share/jellyfin/plugins/configurations/Jellyfin.Plugin.Tmdb.xml` in my dev environment), set the `TmdbApiKey` value, and restart Jellyfin to pick up changes.

An empty or missing `TmdbApiKey` value will cause Jellyfin to use the Jellyfin key, which will not be saved to the xml file.

**Changes**
- Add API key to configuration file, as well as error handling for the plugin config page in case the key is invalid
- Fix missing posters in collection identify page
- Fix a case where poster settings were used for person profile images

**Issues**

### DIFF
--- a/MediaBrowser.Providers/Plugins/Tmdb/BoxSets/TmdbBoxSetProvider.cs
+++ b/MediaBrowser.Providers/Plugins/Tmdb/BoxSets/TmdbBoxSetProvider.cs
@@ -75,12 +75,14 @@ namespace MediaBrowser.Providers.Plugins.Tmdb.BoxSets
             var collections = new RemoteSearchResult[collectionSearchResults.Count];
             for (var i = 0; i < collectionSearchResults.Count; i++)
             {
+                var result = collectionSearchResults[i];
                 var collection = new RemoteSearchResult
                 {
-                    Name = collectionSearchResults[i].Name,
-                    SearchProviderName = Name
+                    Name = result.Name,
+                    SearchProviderName = Name,
+                    ImageUrl = _tmdbClientManager.GetPosterUrl(result.PosterPath)
                 };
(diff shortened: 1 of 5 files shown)


Shadowghost:
I am not sure if we should allow users to set their own API keys. Looking back at when our Trakt plugin had issues, the fix for trakt was quite easy: revoke our API key. If users set their own this is not possible.

VigibotDev:
The problem retrieving collection posters is not related to the TMDb API key used. I still change the API key used by Jellyfin because there is no reason why my statistics should not arrive in my TMDb account. I use MITM-proxy to control outgoing data and/or have my own cache on the LAN so that Jellyfin does not go out on the internet in the event of a simple file movement.

nielsvanvelzen:
> I am not sure if we should allow users to set their own API keys.

Yes, always. We should always respect user privacy and allow them to self-host/change API keys/etc. just like how we added support for self hosted chromecast receivers in 10.9 and allow changing API keys for virtually any plugin. I'm actually surprised this is not the case for the TMDb plugin yet.

1337joe:
> I'd at least argue we should not provide a UI for it.

That's the approach I took. Originally the plugin page broke if the key was invalid so I added enough error handling to report the error to the user, but to actually view or change it you have to modify the plugin settings xml.

JPVenson (approved):

Shadowghost (approved):

crobibero (approved):

Bond-009 (approved):

PR #10751, opened 2023-12-20
Remove HEVC SDR fallback stream

### LINKED ISSUE #8043: [Issue]: HDR lost when audio track needs transcoding

dmitrylyzo:
_Alternative to #9121_

The client starts playback from a stream with an average bitrate. In some cases this is HEVC SDR. After a while, the client switches to the HDR stream, but the server doesn't restart ffmpeg and the client continues playing the old stream.

Since we have a `VideoRange` condition, the HEVC SDR fallback stream is no longer needed, IMO.

Alternatively, we can use bitrate copying - #9121

**Changes**
Remove HEVC SDR fallback stream.

**Issues**
Fixes #8043
Fixes https://github.com/jellyfin/jellyfin-webos/issues/105
Fixes https://github.com/jellyfin/jellyfin-tizen/issues/163
Fixes https://github.com/jellyfin/jellyfin-webos/issues/59
Fixes https://github.com/jellyfin/jellyfin/issues/5576 (https://github.com/jellyfin/jellyfin/issues/5576#issuecomment-1373901085)
Fixes https://github.com/jellyfin/jellyfin-web/issues/1190

### DIFF
--- a/Jellyfin.Api/Helpers/DynamicHlsHelper.cs
+++ b/Jellyfin.Api/Helpers/DynamicHlsHelper.cs
@@ -196,40 +196,6 @@ public class DynamicHlsHelper
         {
             var encodingOptions = _serverConfigurationManager.GetEncodingOptions();
 
-            // Provide SDR HEVC entrance for backward compatibility.
-            if (encodingOptions.AllowHevcEncoding
-                && !encodingOptions.AllowAv1Encoding
-                && EncodingHelper.IsCopyCodec(state.OutputVideoCodec)
-                && state.VideoStream.VideoRange == VideoRange.HDR
-                && string.Equals(state.ActualOutputVideoCodec, "hevc", StringComparison.OrdinalIgnoreCase))
-            {
-                var requestedVideoProfiles = state.GetRequestedProfiles("hevc");
-                if (requestedVideoProfiles is not null && requestedVideoProfiles.Length > 0)
-                {
-                    // Force HEVC Main Profile and disable video stream copy.
-                    state.OutputVideoCodec = "hevc";
-                    var sdrVideoUrl = ReplaceProfile(playlistUrl, "hevc", string.Join(',', requestedVideoProfiles), "main");
-                    sdrVideoUrl += "&AllowVideoStreamCopy=false";
-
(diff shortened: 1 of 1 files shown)


nyanmisaka:
Is there a way to check the client name and bypass this?
AFAIK Apple is still very picky about some stream-copied HDR clips even if HDR is definitely supported by the device.

41BA6084-434C-4C05-A6D2-50BA93B09945

dmitrylyzo:
> Is there a way to check the client name and bypass this?

I don't think we should use the client name to workaround this issue. Instead, we need to add some condition to the device profile.

> AFAIK Apple is still very picky about some stream-copied HDR clips even if HDR is definitely supported by the device.

There isn't much information in the playlist. 
Does it just try each stream until success?

#9121 is a softer solution.

PR #10756, opened 2023-12-21
Add stream compatibility flag and errors

dmitrylyzo:
_Server-side for https://github.com/jellyfin/jellyfin-web/pull/5252_

Every player has to recheck stream compatibility to determine which streams can be played directly.
[This example](https://github.com/jellyfin/jellyfin-web/blob/3d6b27e6a3ca8fa78f0d9050eb5dbfaf01cf82c8/src/plugins/htmlVideoPlayer/plugin.js#L728-L747) has only a generic test (container+codec), but no codec conditions.

On the other hand, the server performs a full validation for the provided device profile (only requested streams, but still).
This PR adds full validation and access to such information.

**Changes**
- Add stream compatibility flag (`SupportsDirectPlay`) and errors (`DirectPlayErrors`).
- Re-use stream compatibility errors.

**Issues**
N/A

**More Info**
It looks like we have a bug with detecting secondary audio: browsers skip unsupported streams, and the second audio stream (supported) becomes the first.
Moreover, Chrome 115+ detects only default audio tracks (bug?).
This can be fixed after complete stream analysis (this PR).

_And I need this for AVPlay player._ :relaxed:

### DIFF
--- a/MediaBrowser.Model/Entities/MediaStream.cs
+++ b/MediaBrowser.Model/Entities/MediaStream.cs
@@ -13,6 +13,7 @@ using Jellyfin.Extensions;
 using MediaBrowser.Model.Dlna;
 using MediaBrowser.Model.Extensions;
 using MediaBrowser.Model.MediaInfo;
+using MediaBrowser.Model.Session;
 
 namespace MediaBrowser.Model.Entities
 {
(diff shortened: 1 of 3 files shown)


MediaBrowser.Model/Entities/MediaStream.cs
```diff
@@ -615,6 +616,24 @@ public bool IsTextSubtitleStream
...
+        /// Gets or sets compatibility errors.
+        /// </summary>
+        /// <value>Compatibility errors.</value>
+        public TranscodeReason? CompatibilityErrors { get; set; }
+
+        /// <summary>
+        /// Gets whether this instance is compatible with device.
+        /// </summary>
+        /// <value><c>true</c> if this instance is compatible with device; otherwise, <c>false</c>.</value>
+        public bool? IsCompatible
```
dmitrylyzo, line 629: Or maybe name them `DirectPlayErrors` and `SupportsDirectPlay` so that they correspond to `MediaSourceInfo.SupportsDirectPlay`? :thinking:

MediaBrowser.Model/Dlna/SingleAudioMode.cs
```diff
@@ -0,0 +1,28 @@
+namespace MediaBrowser.Model.Dlna
+{
+    /// <summary>
+    /// Track selection mode when track switching is not supported.
+    /// </summary>
+    public enum SingleAudioMode
```
dmitrylyzo, line 6: Or `SingleAudioPolicy`? 🤔
Any suggestions?

Shadowghost: I think `SingleAudioPolicy` describes it better

JPVenson:
closed due to inactivity

dmitrylyzo:
Is it possible to reopen the PR? The branch has been updated.

JPVenson:
i only see merge conflicts since September 2024

nielsvanvelzen:
Change to branches don't show up after a PR is closed. If the branch is force pushed after closure it cannot be re-opened. So you need to create a new PR.

PR #10772, opened 2023-12-26
Discard WebM if there is an unsupported codec

dmitrylyzo:
**Changes**
Discard WebM if there is an unsupported codec.

**Issues**
MKV container is considered supported because it is also WebM.


Shadowghost (approved):

MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
```diff
@@ -30,6 +30,8 @@ public partial class ProbeResultNormalizer
         private const string ArtistReplaceValue = " | ";
 
         private readonly char[] _nameDelimiters = { '/', '|', ';', '\\' };
+        private readonly string[] _webmVideoCodecs = { "vp8", "vp9" };
```
dmitrylyzo, line 33: WebM unofficially supports AV1.
```suggestion
        private readonly string[] _webmVideoCodecs = { "av1", "vp8", "vp9" };
```

JPVenson (approved):

crobibero (approved):

Bond-009 (approved):

PR #10781, opened 2023-12-29
Remove UserManager Cache

barronpm:
**Changes**
- Removes the UserManager cache, as queries are already cached by the EFSecondLevelCache. 
This saved roughly ~1s on startup in my testing
- Uses DI for `IAuthenticationProvider` and `IPasswordResetProvider`

### DIFF
--- a/Jellyfin.Server/CoreAppHost.cs
+++ b/Jellyfin.Server/CoreAppHost.cs
@@ -14,6 +14,7 @@ using Jellyfin.Server.Implementations.Security;
 using Jellyfin.Server.Implementations.Trickplay;
 using Jellyfin.Server.Implementations.Users;
 using MediaBrowser.Controller;
+using MediaBrowser.Controller.Authentication;
 using MediaBrowser.Controller.BaseItemManager;
 using MediaBrowser.Controller.Devices;
 using MediaBrowser.Controller.Drawing;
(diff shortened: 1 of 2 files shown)


Shadowghost (approved):

Bond-009 (approved):

JPVenson (approved):

Jellyfin.Server/CoreAppHost.cs
```diff
@@ -78,6 +79,9 @@ protected override void RegisterServices(IServiceCollection serviceCollection)
 
             serviceCollection.AddSingleton<IActivityManager, ActivityManager>();
             serviceCollection.AddSingleton<IUserManager, UserManager>();
+            serviceCollection.AddSingleton<IAuthenticationProvider, DefaultAuthenticationProvider>();
+            serviceCollection.AddSingleton<IAuthenticationProvider, InvalidAuthProvider>();
+            serviceCollection.AddSingleton<IPasswordResetProvider, DefaultPasswordResetProvider>();
```
crobibero, line 84: Since this is moving to DI how are we going to handle external auth providers?
I would rather not move to a pattern where all plugins must register themselves

JPVenson: can't plugins just overwrite the registration by using TryAddSingelton?

barronpm: After discussion, we've decided to proceed with manual service registration

crobibero (approved):

Issue #10780, opened 2023-12-29
Path too long for metadata saver


backcountrymountains:
### Please describe your bug

Jellyfin went out and found some metadata for one of my music files/albums and it's a compilation with a bajillion artists/performers. The metadata saver attempts to create a file with the bajillion contributors, but fails because the path is too long. Then it tries to undo everything by ```Deleting dead``` but that seems to also fail.

It doesn't crash or anything but it's always showing up in my logs, mocking me.

Is there a way to handle the ```System.IO.PathTooLongException``` by truncating the path or something?

### Jellyfin Version

10.8.z

### if other:

10.8.11

### Environment

```markdown
Docker on Debian on WSL on Windows 11
The library volumes are Windows NTFS mounted in WSL Debian and mounted as volumes in Docker

WSL /etc/fstab is:
D:                  /mnt/d drvfs       defaults,metadata,uid=1000,gid=1000   0       0

Distributor ID: Debian
Description:    Debian GNU/Linux 11 (bullseye)
Release:        11
Codename:       bullseye

WSL version: 1.2.5.0
Kernel version: 5.15.90.1
WSLg version: 1.0.51
MSRDC version: 1.2.3770
Direct3D version: 1.608.2-61064218
DXCore version: 10.0.25131.1002-220531-1700.rs-onecore-base2-hyp
Windows version: 10.0.22621.2715

  jellyfin:
    image: jellyfin/jellyfin:latest
    hostname: jellyfin
    container_name: jellyfin
    user: 1000:1000
    depends_on:
      - nginx
    networks:
      - net
    ports:
      - 8096:8096
      - 1900:1900
      - 7359:7359
    environment:
      - JELLYFIN_DATA_DIR=/config
      - TZ=America/Denver
    volumes:
      - /mnt/d/jellyfin/config:/var/lib/jellyfin
      - /mnt/d/jellyfin/config:/config
      - /mnt/d/jellyfin/cache:/cache
      - /mnt/d/jellyfin/:/mnt/d/jellyfin/
    restart: 'unless-stopped'   
    deploy:
      resources:
        reservations:
          devices:
            - capabilities: [gpu]
```

### Jellyfin logs

```shell
[2023-12-29 08:52:10.859 -07:00] [ERR] [158] MediaBrowser.Providers.Manager.ProviderManager: Error in metadata saver
[...]

PR #10784, opened 2023-12-30
Add limit on metadata pathlength

devstra:
**Changes**
Modify BaseXmlSaver and BaseNfoSaver to prevent metadata directory of excessive length from being created (260 as per Windows PATH_MAX, also prevents common Unix NAME_MAX of 255 from being exceeded).

**Issues**
https://github.com/jellyfin/jellyfin/issues/10780

### DIFF
--- a/MediaBrowser.LocalMetadata/Savers/BaseXmlSaver.cs
+++ b/MediaBrowser.LocalMetadata/Savers/BaseXmlSaver.cs
@@ -94,7 +94,7 @@ namespace MediaBrowser.LocalMetadata.Savers
         {
             var path = GetSavePath(item);
             var directory = Path.GetDirectoryName(path) ?? throw new InvalidDataException($"Provided path ({path}) is not valid.");
-            Directory.CreateDirectory(directory);
+            _ = directory.Length > 260 ? Directory.CreateDirectory(directory[..260]) : Directory.CreateDirectory(directory);
 
             // On Windows, savint the file will fail if the file is hidden or readonly
             FileSystem.SetAttributes(path, false, false);
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -174,6 +174,7 @@
  - [Chris-Codes-It] (https://github.com/Chris-Codes-It)
  - [Pithaya](https://github.com/Pithaya)
  - [Çağrı Sakaoğlu](https://github.com/ilovepilav)
+ - [devstra](https://github.com/devstra)
 
 # Emby Contributors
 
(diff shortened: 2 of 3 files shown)


Shadowghost (approved):

MediaBrowser.LocalMetadata/Savers/BaseXmlSaver.cs
```diff
@@ -94,7 +94,7 @@ public async Task SaveAsync(BaseItem item, CancellationToken cancellationToken)
         {
             var path = GetSavePath(item);
             var directory = Path.GetDirectoryName(path) ?? throw new InvalidDataException($"Provided path ({path}) is not valid.");
-            Directory.CreateDirectory(directory);
+            _ = directory.Length > 260 ? Directory.CreateDirectory(directory[..260]) : Directory.CreateDirectory(directory);
```
JPVenson, line 97: just truncating the path is a poor way to fix this.
For windows you could got with the extended path sytax instead:
https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry

Or truncate it in the usual windows way by using the tilde character.

also for linux the max _folder_ and _filename_ length is 255 which will potentially break a lot of systems. The Path length is usually 4000.

JPVenson (changes requested):

MediaBrowser.LocalMetadata/Savers/BaseXmlSaver.cs
devstra: I see your point, it would make more sense to either use Windows extended path syntax or just cut the folder name short depending on OS. I'll close this PR and open a new one once I implement that.

JPVenson: hmm here it gets even worse, usually directory length restrictions are not OS limitations but filesystem ones. Each filesystem has other length limitations.

devstra: That does complicate things. Would it make sense to query the filesystem using [this property](https://learn.microsoft.com/en-us/dotnet/api/system.io.driveinfo.driveformat?view=net-8.0&redirectedfrom=MSDN#System_IO_DriveInfo_DriveFormat) and then map each filesystem to a NAME_MAX? Or would it be best to just use 255 as a catch-all for Unix systems? Very few (common) filesystems seem to deviate from 255.

JPVenson: Well for folder length anyway not path length. I think there is somewhere a variable you can get to get the max folder length yes but not the max path length ... or vice versa.

JPVenson: Tbo you could just catch the error and log it as is to not break the flow for other actions in the calling code, but i dont see how that can be fixed without opening a huge hole in the code (as this is literally everywhere in code "an issue")

devstra: If metadata wasn't automatically pulled then this wouldn't be an issue (or at least much less of one). I don't think just logging it would be an effective solution because of this: automatically pulling metadata is expected to work without user interaction. The majority of Unix-like filesystems don't have a PATH_MAX less than 4096, so I think it might be fair to leave it that way as a catch-all, as with with 255 for NAME_MAX.

PR #10787, opened 2023-12-31
Add limit on metadata pathlength (revised)

devstra:
**Changes**
Modify BaseXmlSaver and BaseNfoSaver to prevent metadata directory of excessive length from being created. On Windows, extended-path notation ("\\\\?\\") is used, and on Unix-like systems, the created directory is kept shorter than 255 characters (in-line with the most common Unix-like filesystem NAME_MAX value).

**Notes**
There are a couple of things to note about this changes:
1. This is a revision of #10784, which I decided to close after realizing that I had misunderstood the role of savers and created an ineffective solution.
2. This does not tailor a solution to individual filesystems, each of which usually have a different NAME_MAX. This is intentional, as every commonly-used filesystem has a NAME_MAX greater than or equal to 255, so keeping it at this value seems reasonable. 
3. I did not implement a solution to capping the path at 4096 characters (again, the most common Unix-like filesystem PATH_MAX by far) since this seemed redundant when the issue at hand is that automatically-retrieved metadata can result in folders of length greater than NAME_MAX being created. It's unlikely that any metadata item would come anywhere near this value, and I think it makes more sense to leave this as an error to be handled due to the difficulty in working around it.

**Issues**
#10780


MediaBrowser.LocalMetadata/Savers/BaseXmlSaver.cs
```diff
@@ -94,6 +94,14 @@ public async Task SaveAsync(BaseItem item, CancellationToken cancellationToken)
         {
             var path = GetSavePath(item);
             var directory = Path.GetDirectoryName(path) ?? throw new InvalidDataException($"Provided path ({path}) is not valid.");
+
+            if (directory.Length > 260 || item.Name.Length > 255)
+            {
+                _ = OperatingSystem.IsWindows() ? _ = directory.Insert(0, @"\\?\") : directory = Path.Combine(
```
JPVenson, line 100: I dont know if its handled differently on MacOS did you check that too?

JPVenson: also please expand that code, its very hard to read

devstra: Updated just now, see comments in most recent commit. Any metadata which doesn't inherit from Folder gets its save file name from an existing file, preventing the possibility of it exceeding the maximum path length.

MediaBrowser.LocalMetadata/Savers/BaseXmlSaver.cs
stesee, line 1: I am a bit late in the game. 

- There are edge cases where strings with a length of 255 char will not fit into the max length of 255byte filename on fs like ext or zfs.
- There are edge cases (e.g. emoticons, an I think non latin text too ) where truncating using .net substring will result in unexpected trailing chars.
Have a look into https://github.com/Codeuctivity/SanitizeFilename/blob/main/SanitizeFilenameTests/SanitizeTests.cs#L119 for such example values.

stesee: Trailing '.' and ' ' seems also be possible that is not recommended by https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions .
 I would be happy if you think about using https://www.nuget.org/packages/Codeuctivity.SanitizeFilename/ where these edge cases got solved.

JPVenson:
closed due to inactivity

PR #10793, opened 2024-01-01
Auto Discovery Cleanup

barronpm:
**Changes**
Rewrites auto-discovery as a `BackgroundService` using `UdpClient`. Tested on Windows+Linux using broadcast and unicast.


Shadowghost (approved):

Shadowghost:
This does not respect interface or bind address restrictions though, right? IMHO we should try to keep this functionality.

barronpm:
> This does not respect interface or bind address restrictions though, right? IMHO we should try to keep this functionality.

Good point, I've updated the PR.

src/Jellyfin.Networking/AutoDiscoveryHost.cs
```diff
@@ -0,0 +1,127 @@
...
+        foreach (var intf in _networkManager.GetInternalBindAddresses())
+        {
+            var address = OperatingSystem.IsLinux()
+                ? NetworkUtils.GetBroadcastAddress(intf.Subnet)
+                : intf.Address;
+
+            _ = Task.Run(() => ListenForAutoDiscoveryMessage(address, stoppingToken), stoppingToken);
+        }
+
+        return Task.CompletedTask;
```
JPVenson, line 74: You should return all created tasks so that "StopAsync" will await them when stop is requested to ensure they are properly stopped

https://learn.microsoft.com/en-us/aspnet/core/fundamentals/host/hosted-services?view=aspnetcore-8.0&tabs=visual-studio#backgroundservice-base-class

crobibero (approved):

JPVenson (approved):

Bond-009 (approved):

PR #10812, opened 2024-01-04
Remove some unused client capabilities and sync code

nielsvanvelzen:
Remove some legacy stuff that clutters the API.

**Changes**
- Remove unused stuff from client capabilities
- Remove left over sync code from various places

**Issues**

### DIFF
--- a/MediaBrowser.Model/Dto/BaseItemDto.cs
+++ b/MediaBrowser.Model/Dto/BaseItemDto.cs
@@ -85,11 +85,6 @@ namespace MediaBrowser.Model.Dto
 
         public string PreferredMetadataCountryCode { get; set; }
 
-        /// <summary>
-        /// Gets or sets a value indicating whether [supports synchronize].
-        /// </summary>
-        public bool? SupportsSync { get; set; }
-
         public string Container { get; set; }
 
         /// <summary>
--- a/Emby.Server.Implementations/Dto/DtoService.cs
+++ b/Emby.Server.Implementations/Dto/DtoService.cs
@@ -418,15 +418,6 @@ namespace Emby.Server.Implementations.Dto
             {
                 dto.PlayAccess = item.GetPlayAccess(user);
             }
-
-            if (options.ContainsField(ItemFields.BasicSyncInfo))
-            {
-                var userCanSync = user is not null && user.HasPermission(PermissionKind.EnableContentDownloading);
-                if (userCanSync && item.SupportsExternalTransfer)
-                {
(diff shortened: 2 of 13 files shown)


nielsvanvelzen:
(OpenAPI diff fails to generate because I touched BaseItemDto)

MediaBrowser.Model/Dto/BaseItemDto.cs
```diff
@@ -85,11 +85,6 @@ public class BaseItemDto : IHasProviderIds, IItemDto, IHasServerId
 
         public string PreferredMetadataCountryCode { get; set; }
 
-        /// <summary>
-        /// Gets or sets a value indicating whether [supports synchronize].
-        /// </summary>
-        public bool? SupportsSync { get; set; }
```
nielsvanvelzen, line 91: @thornbill the web client is the only repository that actually uses this property but I believe the code using it is legacy:

https://github.com/jellyfin/jellyfin-web/blob/e7c149dd93f6195f83fc7c596f98672bfbfda6d3/src/components/itemHelper.js#L157C17-L167

https://github.com/jellyfin/jellyfin-web/blob/master/src/components/itemContextMenu.js#L293-L299

crobibero (approved):

Bond-009 (approved):

MediaBrowser.Model/Dto/BaseItemDto.cs
thornbill: Yeah looks like it 👍

Shadowghost (approved):

MediaBrowser.Model/Dto/BaseItemDto.cs
nielsvanvelzen: https://github.com/jellyfin/jellyfin-web/pull/5082

janvs0:
Hi @nielsvanvelzen this breaks logging in using jellyfin-androidtv 0.16.4 for me (but still works fine on the non-tv android client). Looking at the logcat excerpt below I think the unused capabilities also need removing from jellyfin-sdk-kotlin.

```
kotlinx.serialization.MissingFieldException: Fields [SupportsContentUploading, SupportsSync] are required for type with serial name 'org.jellyfin.sdk.model.api.ClientCapabilities', but they were missing at path: $.SessionInfo.Capabilities 
```

nielsvanvelzen:
I expected this change to break the Android TV app yes. We will start on support for 10.9 once the first beta for it is out (although removing those fields from the request could be done now).

Issue #10800, opened 2024-01-03
[Issue]: /System/Info/Public returns 200 and empty response when remote access is disabled instead of an error


lf-:
### Please describe your bug

I forgot to enable remote access, and my jellyfin had absolutely incomprehensible symptoms of that (real) configuration issue.

When I have LAN networks set to the default empty value and access the server via a non-local subnet, I cannot access the web interface because it shows "undefined" on the server selection page, and there are strange JS errors. I traced it back to the fact that the server is responding with 200 and an empty payload for `/System/Info/Public`, which seems like a bug.

For your viewing pleasure I have reduced this bug to a relatively trivial `curl` command.

In the default state, if I execute this command (or, open the web ui while behind a reverse proxy):

```
jade@cube:~/ > curl  -vv -H 'Host: meowmeow.jade.fyi' -H 'X-forwarded-for: 100.67.51.25' http://localhost:8096/System/Info/Public
*   Trying [::1]:8096...
* connect to ::1 port 8096 failed: Connection refused
*   Trying 127.0.0.1:8096...
* Connected to localhost (127.0.0.1) port 8096
> GET /System/Info/Public HTTP/1.1
> Host: meowmeow.jade.fyi
> User-Agent: curl/8.4.0
> Accept: */*
> X-forwarded-for: 100.67.51.25
>
< HTTP/1.1 200 OK
< Content-Length: 0
< Date: Wed, 03 Jan 2024 11:48:20 GMT
< Server: Kestrel
< X-Response-Time-ms: 0
<
* Connection #0 to host localhost left intact
```

"Nothing" is not the expected response. This is the expected response, which is observed when the LAN addresses in the settings (`127.0.0.0/8,192.168.0.0/16,100.64.0.0/10`) match the X-forwarded-for address, or remote access is enabled:

```
jade@cube:~/ > curl -vv -H 'Host: meowmeow.jade.fyi' -H 'X-forwarded-for: 100.67.51.25' http://localhost:8096/System/Info/Public
*   Trying [::1]:8096...
* connect to ::1 port 8096 failed: Connection refused
*   Trying 127.0.0.1:8096...
* Connected to localhost (127.0.0.1) port 8096
> GET /System/Info/Public HTTP/1.1
> Host: meowmeow.jade.fyi
> User-Agent: curl/8.4.0
> Accept: */*
> X-forwarded-for: 100.67.51.25
>
< HTTP/1.1 200 OK
[...]

PR #10822, opened 2024-01-06
Respond with Service Unavailable if not in correct network

crobibero:
Fixes https://github.com/jellyfin/jellyfin/issues/10800

### DIFF
--- a/Jellyfin.Api/Middleware/IpBasedAccessValidationMiddleware.cs
+++ b/Jellyfin.Api/Middleware/IpBasedAccessValidationMiddleware.cs
@@ -42,7 +42,7 @@ public class IPBasedAccessValidationMiddleware
         if (!networkManager.HasRemoteAccess(remoteIP))
         {
             // No access from network, respond with 503 instead of 200.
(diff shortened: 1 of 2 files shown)


nielsvanvelzen (approved):

Jellyfin.Api/Middleware/IpBasedAccessValidationMiddleware.cs
```diff
@@ -41,6 +41,8 @@ public async Task Invoke(HttpContext httpContext, INetworkManager networkManager
 
         if (!networkManager.HasRemoteAccess(remoteIP))
         {
+            // No access from network, respond with 503 instead of 200.
+            httpContext.Response.StatusCode = (int)HttpStatusCode.ServiceUnavailable;
```
nielsvanvelzen, line 45: Would it make sense to use `StatusCodes.Status503ServiceUnavailable` instead?

barronpm (approved):

Shadowghost (approved):

PR #10825, opened 2024-01-06
Use helper function to compare guid

crobibero:
I found a few places where we were comparing a nullable guid to `default`, which isn't what was actually desired.

Double check to make sure I didn't replace anything in an EF call.


barronpm:
If this is Guid specific, does it make sense to name them `IsEmpty` and `IsNullOrEmpty`? IMO that's a bit clearer in what the intent is behind the comparisons.

MediaBrowser.Controller/Entities/Video.cs
```diff
@@ -456,7 +456,7 @@ private void RefreshLinkedAlternateVersions()
             foreach (var child in LinkedAlternateVersions)
             {
                 // Reset the cached value
-                if (child.ItemId.HasValue && child.ItemId.Value.Equals(default))
+                if (child.ItemId.IsNullOrEmpty())
```
barronpm, line 459: This changes behavior, the existing code checks that it is empty but _not_ `null`

crobibero: I don't think the difference is meaningful enough to keep the `.HasValue` check, now a Guid? that is already null can be set to null again

barronpm: Didn't see the body of the if, makes sense

barronpm (approved):

PR #10826, opened 2024-01-07
Use frozen collections in MimeTypes.cs

goremykin:
This PR just uses frozen collections in MimeTypes.cs, fixes a few minor warnings and gets rid of a few memory allocations.


JPVenson:
i think that in theory its a good idea to do that where applicable but i dont see that big of an impact in the grand scheme of things. If such methods need 10 or 60 ms does not really change anything. I would recommend that you open the general idea on our meta repro where you are discussing architectural things.

JPVenson (approved):

crobibero (approved):

Shadowghost (approved):

Bond-009 (approved):

Issue #7228, opened 2022-01-21
Music Continue Listening


fidoriel:
What ist the Issue?
Jellyfin is great but it lacks proper audiobook support(https://features.jellyfin.org/posts/243/audiobook-support). This feature request turns this year three years, so I decided to tackle one important feature for me. In a nutshell: I do not plan to make huge changes, just port the continue listening feature to the music library. 

How do I plan to solve it?
Continue listening is only supported for Audiobookfiles. Which can be really difficult to navigate because they can get really long(40h) and Jellyfin does not support m4b chapters as I know. because this is not only an issue in Jellyfin, I keep my Audiobooks in Multi files (chapters). (This allows a loses compression from audio dramas).
Nearly all of my Audiobooks are multi file, and ID3 tagged. If I add them to a Book lib in Jellyfin, the handling is horrible. It is just a Subfolder mess. To work around that I created a second Music library called Audiobooks and add them there. The yield I get from the ID3 tags and metadata provider MusicBrainz are really good. The only thing this is lacking is the Continue listening from .m4b audio files.

I can imagine that lots of people are doing it like me, so they already have created a Music library for Audiobooks in Jellyfin. My idea is to add an option to the library settings called "Enable Continue Listening", disabled by default. If ticked this should remember Album track and minute and add an entry to continue listening.
![Bildschirmfoto 2022-01-21 um 10 58 14](https://user-images.githubusercontent.com/49869342/150507090-03663a36-1df7-4c0a-b2f4-60269794c5da.png)

Maybe some Music enthusiast will find it great too. e.g. for long concerts?

I know it is not the best Audiobook support possible neither a great solution, but it would solve the issues I have in a convenient, time efficient and practical way.
I do not know who I have to ask for permission to tackle this issue to make sure it will find its way into the repo.

liponex:
Still affects me

foss-:
Is this still valid for Jellyfin 10.8.0?
Was struggling a lot with Jellyfin Book categorie and moved all Audiobooks to a second music library as per your suggestion and that seems to do the trick. Starting a track from an audiobook does remember the playback position by default. I did not alter the default library preferences.
The track also shows in the `Continue Listening` section.
There are a few remaining problems, like for me album section for Audiobooks not showing any album at all. But that would be a different problem than what this issue here is about.

foxhoundv:
[...]

Issue #2084, opened 2019-11-30
Resuming Audio


ovnx:
Resuming audio should be available, particularly for things like audio books. Menu in admin interface only shows "Resume Video" settings.

JustAMan:
This looks like a part of https://features.jellyfin.org/posts/243/audiobook-support

dkanada:
This has been implemented for AudioBook items, which are available when using the book library.

LeifAndersen:
Given that @Tolriq is unwilling to add syncing with book libraries in the Yatse app[1], would it be possible to enable this on music libraries as well as book ones?

[1]: https://community.yatse.tv/t/jellyfin-audiobooks-not-recognized-as-music/3180/5

PR #10834, opened 2024-01-09
Support resume on all audio

blakesmith:
I use Jellyfin to listen to multi-hour audio tracks. Not having Jellyfin remember my play position for audio files has been rough. I ended up patching my Jellyfin server build to allow all audio files to `SupportsPositionTicksResume`. This causes all playback positions for audio files to be saved.

I'm not sure if this is undesirable default behavior for others, but I'd love to have this change incorporated into upstream somehow, even if it needs to be behind a settings toggle (happy to help add this if that's the direction this needs to go).

Screenshot from 2024-01-08 21-29-44

**Changes**

All `Audio` entity types now default to saving their playback position via `SupportsPositionTicksResume`.

**Issues**

See: https://github.com/jellyfin/jellyfin/issues/7228 and https://github.com/jellyfin/jellyfin/issues/2084

### DIFF
--- a/MediaBrowser.Controller/Entities/Audio/Audio.cs
+++ b/MediaBrowser.Controller/Entities/Audio/Audio.cs
@@ -38,6 +38,9 @@ namespace MediaBrowser.Controller.Entities.Audio
         public IReadOnlyList<string> AlbumArtists { get; set; }
 
         [JsonIgnore]
+        public override bool SupportsPositionTicksResume => true;
+
+        [JsonIgnore]
         public override bool SupportsPlayedStatus => true;
 
         [JsonIgnore]


foss-:
Some thoughts: First, thank you for stepping up and creating this PR. As an avid audio book listener, it is frustrating, that I can't use Jellyfin to effectively listing to them (does not remember play position).

I don't think this is a good default for all audio files though. It means over time, you'll have a library where many files will start in the middle of the song if you frequently don't finish listening to a music file.

This request was initially brought up specifically for audio book content in https://github.com/jellyfin/jellyfin/issues/7228, where it is highly desirable to remember play position.

PR author added the use case of hour long music DJ sets where this also would be useful.

For audio books those can always be separated into their own section in Jellyfin and if this was a setting, that could be enabled for that section and not for the music section generally. Should default to off, due to argument above.

For DJ mixes that then would either mean to also move them into their own section or to have the setting applied to entire music folder with the mentioned drawback.

Overall I like this change and think it would improve Jellyfin UX a lot and allow more flexibility to cover more diverse use cases when it comes to playing audio files.

blakesmith:
Thanks for your thoughts @foss-! I might spend some time exploring your suggestions, especially if a library level setting would make this UX change more likely to be upstreamed (so I don't have to maintain an out-of-tree patch in perpetuity).

Chaphasilor:
Usually such an option is implemented in a way where playback position is only saved for tracks longer than x minutes. Alternatively, this could be a client setting where the clients are free to decide if they want to progress to be saved (e.g. by periodically submitting the current position).

8bitbuddhist:
For what it's worth, I [enabled this patch in Jellyfin 10.9.7 running on NixOS](https://github.com/8bitbuddhist/nix-configuration/blob/0b4422b1f5a4bac6051ac7495f2fed87ac5b08f7/modules/services/jellyfin.nix#L10) and it worked right away, so thank you for providing this 👍🏾

blakesmith:
@8bitbuddhist Nice! I ended up switching to [Navidrome](https://www.navidrome.org/) for self-hosted home audio, where playback saving can be more up to the client. I don't have the time or interest to move this PR forward anymore, if anyone else wants to pick up the torch.

JPVenson:
Closed as orphanaged

Issue #8502, opened 2022-10-06
[Issue]: Unable to get SchedulesDirect Data


ryantiger658:
### Please describe your bug

When I run TV Guide update my system throws a much of HTTP exceptions and no data is collected from SchedulesDirect.

I am able to see the Lineup options, but am unable to get the Schedule data.

I have updated and restarted the server, as well as deleted and re-created all of the Live TV settings

### Jellyfin Version

10.8.0

### Environment

```markdown
- OS: Docker
- Virtualization: Docker
- Clients: All
- Browser: Chrome
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking: Host
- Storage: Local
```

### Jellyfin logs

```shell
15:46:53] [INF] [26] Emby.Server.Implementations.LiveTv.Listings.SchedulesDirect: Authenticated with Schedules Direct token: REDACTED
[15:46:53] [ERR] [26] Emby.Server.Implementations.LiveTv.LiveTvManager: Error getting programs for channel MYST
System.InvalidOperationException: The request message was already sent. Cannot send the same request message multiple times.
   at System.Net.Http.HttpClient.CheckRequestMessage(HttpRequestMessage request)
   at System.Net.Http.HttpClient.CheckRequestBeforeSend(HttpRequestMessage request)
   at Emby.Server.Implementations.LiveTv.Listings.SchedulesDirect.Send(HttpRequestMessage options, Boolean enableRetry, ListingsProviderInfo providerInfo, CancellationToken cancellationToken, HttpCompletionOption completionOption)
   at Emby.Server.Implementations.LiveTv.Listings.SchedulesDirect.Send(HttpRequestMessage options, Boolean enableRetry, ListingsProviderInfo providerInfo, CancellationToken cancellationToken, HttpCompletionOption completionOption)
   at Emby.Server.Implementations.LiveTv.Listings.SchedulesDirect.GetChannels(ListingsProviderInfo info, CancellationToken cancellationToken)
   at Emby.Server.Implementations.LiveTv.EmbyTV.EmbyTV.GetEpgChannels(IListingsProvider provider, ListingsProviderInfo info, Boolean enableCache, CancellationToken cancellationToken)
[...]

PR #10838, opened 2024-01-09
Move Live TV code to Jellyfin.LiveTv

barronpm:
Moves Live TV code to separate project


nielsvanvelzen:
Would it make sense to rename `EmbyTV` -> `JellyfinTV` (both the namespace & class)?

barronpm:
> Would it make sense to rename `EmbyTV` -> `JellyfinTV` (both the namespace & class)?

EmbyTV will need to be refactored at some point, but in the interest of minimizing conflicts I've left that alone for now

Shadowghost (approved):

src/Jellyfin.LiveTv/Listings/SchedulesDirect.cs
```diff
@@ -613,6 +613,7 @@ private async Task<string> GetToken(ListingsProviderInfo info, CancellationToken
 
             // Response is automatically disposed in the calling function,
             // so dispose manually if not returning.
+#pragma warning disable IDISP016, IDISP017
```
Bond-009, line 616: This function is broken BTW
https://github.com/jellyfin/jellyfin/issues/8502#issuecomment-1341123200

Bond-009 (approved):

crobibero (approved):

PR #10841, opened 2024-01-09
Refresh live TV channel image when remapped (alt #7843) (fixes #7834)

### LINKED ISSUE #7834: [Issue]: In Live TV, refreshing guide data doesn't update channel images after remapping them manually

Bond-009:
Alternative solution that doesn't add another prop to `LiveTvChannel`

### DIFF
--- a/Emby.Server.Implementations/Library/LibraryManager.cs
+++ b/Emby.Server.Implementations/Library/LibraryManager.cs
@@ -1869,8 +1869,14 @@ namespace Emby.Server.Implementations.Library
             var outdated = forceUpdate
                 ? item.ImageInfos.Where(i => i.Path is not null).ToArray()
                 : item.ImageInfos.Where(ImageNeedsRefresh).ToArray();
-            // Skip image processing if current or live tv source
-            if (outdated.Length == 0 || item.SourceType != SourceType.Library)
+
+            var parentItem = item.GetParent();
+            var isLiveTvShow = item.SourceType != SourceType.Library &&
+                               parentItem is not null &&
+                               parentItem.SourceType != SourceType.Library; // not a channel
+
+            // Skip image processing if current or live tv show
+            if (outdated.Length == 0 || isLiveTvShow)
             {
                 RegisterItem(item);
                 return;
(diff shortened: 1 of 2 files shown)


Emby.Server.Implementations/LiveTv/LiveTvManager.cs
```diff
@@ -494,14 +494,17 @@ private async Task<LiveTvChannel> GetChannelAsync(ChannelInfo channelInfo, strin
 
             item.Name = channelInfo.Name;
 
-            if (!item.HasImage(ImageType.Primary))
+            var currentPrimary = item.GetImageInfo(ImageType.Primary, 0);
+            var imageUrlIsNull = string.IsNullOrWhiteSpace(channelInfo.ImageUrl);
+            if (currentPrimary is null
+                || (!imageUrlIsNull && currentPrimary.Path != channelInfo.ImageUrl))
```
barronpm, line 500: This could probably do with a comment explaining the rationale

Shadowghost (approved):

Shadowghost (approved):

theguymadmax (approved):
Tested: Channel images  were updated when channels were remapped. Prior to this PR this did not work.

PR #10850, opened 2024-01-11
Remove DLNA-specific DeviceProfile code

barronpm:
These will be moved to the plugin


MediaBrowser.Model/Dlna/DeviceProfile.cs
```diff
@@ -31,104 +26,6 @@ public class DeviceProfile
         [XmlIgnore]
```
nielsvanvelzen, line 26: Are those XmlIgnore attributes still needed? AFAIK only DLNA profiles are stored in XML.

barronpm: My plan was to move this code to a class called `DlnaDeviceProfile` that extends this class, so some of them are still needed unfortunately AFAIK. If there's an easy way to apply the attributes from the dlna plugin I'd prefer that, but idk.

nielsvanvelzen: Might be possible with an override.

nielsvanvelzen (approved):

Shadowghost (approved):

MediaBrowser.Model/Dlna/DeviceProfile.cs
barronpm: Looks like it's possible, but there are some caveats associated, so I'll leave this as-is for now and play around with it later

Bond-009 (approved):

crobibero (approved):

PR #10872, opened 2024-01-15
Update issue report.yml

felix920506:
**Changes**
Update the issue template to make initial reports clearer and easier to work with (hopefully)

**Issues**

None

### DIFF
--- a/.github/ISSUE_TEMPLATE/issue report.yml
+++ b/.github/ISSUE_TEMPLATE/issue report.yml
@@ -7,6 +7,7 @@ body:
     attributes:
       value: |
         Thanks for taking the time to fill out this bug report! Please provide as much detail as necessary, most questions may not be applicable to you. If you need real-time help, join us on [Matrix](https://matrix.to/#/#jellyfin-troubleshooting:matrix.org) or [Discord](https://discord.gg/zHBxVSXdBV).
+        Do keep in mind that Jellyfin is an English only project. Please submit bug reports in English if you can. If you don't feel comfortable using English, then submit in your native language, and wait for someone to translate it. Do not resort to tools like Google Translate.
   - type: textarea
     id: what-happened
     attributes:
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/issue report.yml
```diff
@@ -24,24 +26,14 @@ body:
           4. See error...
     validations:
       required: true
-  - type: dropdown
+  - type: input
     id: version
     attributes:
       label: Jellyfin Version
       description: What version of Jellyfin are you running?
-      options:
```
JPVenson, line 32: i would rather see this to be expanded to include recent minor versions instead of an input

felix920506: I don't think including recent minor versions would be sustainable.

Reasons as follows:
1. With how long it generally takes for things to get merged it may be months between a new release and the option being available in the issue report template.
2. Including a whole bunch of minor versions will make the dropdown very long and inconvenient 
3. Without an automated mechanism to update it every release, it would have to be done manually

JPVenson: i see but the effort then lies on our side, which we can include in the process, push with enough annoying the maintainers with relative ease. If we make it an free-text input the burden of putting something useful there lies on the user-side which will result in more issues/unclarities.

felix920506: If updating the list gets included into the formal release process then I'm ok with having all the subversions in the dropdown. How would the process of adding such thing into the release process work?

JPVenson: Annoying @joshuaboniface for as long as it needs so he remembers it.

felix920506: If this can be made a process I'll write up another PR for a more ideal (I think) template for it.

Issue #10873, opened 2024-01-15
[Issue]: "Group movies into collections" has no effect


mcarlton00:
### Please describe your bug

In current unstable, the "group movies into collections" tick box has no effect.  It shows all movies individually instead of grouping them into their respective collections.

Possibly related, but probably not, but on current stable (10.8.13) this setting drastically slows down the api response.  Without the option enabled, requesting all movies returns in .5 seconds.  With the option enabled, it takes 23 seconds.

### Jellyfin Version

Other

### if other:

Docker unstable: https://hub.docker.com/layers/jellyfin/jellyfin/unstable/images/sha256-d23ea94feca3263665cc5e905133c5d71b461fdeb9c105406ef7688b4a9dca35?context=explore

### Environment

```markdown
- OS: Debian 12
- Linux Kernel: 6.2.16
- Virtualization: Docker
- Clients: Web
- Browser: Firefox
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #10876, opened 2024-01-16
Fix incorrect path check in CleanupCollectionAndPlaylistPathsTask

Bond-009:


crobibero (approved):

barronpm (approved):

mcarlton00:
ref https://github.com/jellyfin/jellyfin/issues/10873.  setting has no effect because collections got emptied

PR #10878, opened 2024-01-17
Add IGuideManager service

barronpm:
Moves guide-related code to new `IGuideManager` service


src/Jellyfin.LiveTv/Guide/GuideManager.cs
```diff
@@ -0,0 +1,713 @@
...
+        if (coreService is not null)
+        {
+            await coreService.RefreshSeriesTimers(cancellationToken).ConfigureAwait(false);
+            await coreService.RefreshTimers(cancellationToken).ConfigureAwait(false);
+        }
+
+        // Load these now which will prefetch metadata
+        var dtoOptions = new DtoOptions();
+        var fields = dtoOptions.Fields.ToList();
+        dtoOptions.Fields = fields.ToArray();
```
Bond-009, line 151: What's the point of this `ToList().ToArray()`?

src/Jellyfin.LiveTv/Guide/GuideManager.cs
```diff
@@ -0,0 +1,713 @@
...
+                var isNews = false;
+                var isKids = false;
+                var isSeries = false;
+
+                var channelPrograms = (await service.GetProgramsAsync(currentChannel.ExternalId, start, end, cancellationToken).ConfigureAwait(false)).ToList();
+
+                var existingPrograms = _libraryManager.GetItemList(new InternalItemsQuery
+                {
+                    IncludeItemTypes = [BaseItemKind.LiveTvProgram],
+                    ChannelIds = new[] { currentChannel.Id },
```
Bond-009, line 235: Can't you use the new syntax here too?

Bond-009 (changes requested):

src/Jellyfin.LiveTv/Guide/GuideManager.cs
barronpm: Right, I meat to point this out. As far as I can tell, this whole block does nothing, I just wanted to double-check before getting rid of it.

src/Jellyfin.LiveTv/Guide/RefreshGuideScheduledTask.cs
```diff
@@ -0,0 +1,71 @@
...
+    /// <inheritdoc />
+    public Task ExecuteAsync(IProgress<double> progress, CancellationToken cancellationToken)
+        => _guideManager.RefreshGuide(progress, cancellationToken);
+
+    /// <inheritdoc />
+    public IEnumerable<TaskTriggerInfo> GetDefaultTriggers()
+    {
+        return new[]
+        {
+            // Every so often
```
crobibero, line 67: I love the "every so often" poetry we inherited. I don't see how this comment adds any actual info

src/Jellyfin.LiveTv/Guide/GuideManager.cs
```diff
@@ -0,0 +1,713 @@
...
+
+        progress.Report(100);
+    }
+
+    private double GetGuideDays()
+    {
+        var config = _config.GetLiveTvConfiguration();
+
+        return config.GuideDays.HasValue
+            ? Math.Max(1, Math.Min(config.GuideDays.Value, MaxGuideDays))
```
crobibero, line 161: Can change to `Math.Clamp`?

src/Jellyfin.LiveTv/Guide/GuideManager.cs
barronpm: Removed in latest commit

src/Jellyfin.LiveTv/Guide/GuideManager.cs
```diff
@@ -0,0 +1,709 @@
...
+    private double GetGuideDays()
+    {
+        var config = _config.GetLiveTvConfiguration();
+
+        return config.GuideDays.HasValue
+            ? Math.Clamp(config.GuideDays.Value, 1, MaxGuideDays)
+            : 7;
+    }
+
+    private async Task<Tuple<List<Guid>, List<Guid>>> RefreshChannelsInternal(ILiveTvService service, ActionableProgress<double> progress, CancellationToken cancellationToken)
```
Bond-009, line 161: Maybe we should change this to return a `ValueTuple` with named components

src/Jellyfin.LiveTv/Guide/GuideManager.cs
```diff
@@ -0,0 +1,709 @@
...
+    {
+        progress.Report(10);
+
+        var allChannelsList = (await service.GetChannelsAsync(cancellationToken).ConfigureAwait(false))
+            .Select(i => new Tuple<string, ChannelInfo>(service.Name, i))
```
Bond-009, line 166: This seems wasteful, no need to create a Tuple here, right? We can just use `service.Name` directly in the following foreach loop

Bond-009 (approved):
Small notes on existing code, not really important for this PR specifically

src/Jellyfin.LiveTv/Guide/GuideManager.cs
barronpm: Right, I was considering passing in the outer lists and using those directly instead of creating new lists each time.

crobibero (approved):

PR #10934, opened 2024-01-30
Add PlaybackOrder to Session state

dkacperski97:
Changes
~~Add ShuffleMode to Session state.~~
Add OrderMode to Session state.

SetShuffleQueue command is already listed as one of the supported commands by Session, but without the value of ShuffleMode it is not possible to send a valid request from jellyfin-web client.

Issues
Partially fixes https://github.com/jellyfin/jellyfin-web/issues/5012 (small changes in jellyfin-web client are required to fully fix this issue)

### DIFF
--- a/MediaBrowser.Model/Session/GeneralCommandType.cs
+++ b/MediaBrowser.Model/Session/GeneralCommandType.cs
@@ -48,6 +48,7 @@ namespace MediaBrowser.Model.Session
         PlayNext = 38,
         ToggleOsdMenu = 39,
         Play = 40,
-        SetMaxStreamingBitrate = 41
+        SetMaxStreamingBitrate = 41,
+        SetPlaybackOrder = 42
     }
 }
--- a/Emby.Server.Implementations/Session/SessionManager.cs
+++ b/Emby.Server.Implementations/Session/SessionManager.cs
@@ -394,6 +394,7 @@ namespace Emby.Server.Implementations.Session
             session.PlayState.SubtitleStreamIndex = info.SubtitleStreamIndex;
             session.PlayState.PlayMethod = info.PlayMethod;
             session.PlayState.RepeatMode = info.RepeatMode;
+            session.PlayState.PlaybackOrder = info.PlaybackOrder;
             session.PlaylistItemId = info.PlaylistItemId;
 
             var nowPlayingQueue = info.NowPlayingQueue;
--- a/MediaBrowser.Model/Session/PlaybackOrder.cs
+++ b/MediaBrowser.Model/Session/PlaybackOrder.cs
@@ -1,0 +1,18 @@
+namespace MediaBrowser.Model.Session
+{
+    /// <summary>
+    /// Enum PlaybackOrder.
+    /// </summary>
+    public enum PlaybackOrder
+    {
+        /// <summary>
+        /// Sorted playlist.
+        /// </summary>
+        Default = 0,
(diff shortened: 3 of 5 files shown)


nielsvanvelzen (commented):
I would personally prefer "OrderMode" as name instead of ShuffleMode but [I might be biased](https://github.com/jellyfin/jellyfin-androidtv/blob/master/playback/core/src/main/kotlin/model/PlaybackOrder.kt).

dkacperski97:
Yeah, "ShuffleMode" as a name is a little confusing, but that name is already used in SyncPlay, so it would probably need to be changed there too for consistency. It would also be nice if the `SetShuffleQueue` command was renamed. That would affect the clients. Maybe we could mark the old command as deprecated?

I could do it if I get the green light, but maybe as a separate PR? This PR is intended to fix functionally, that should already be there.

Or maybe you only meant the name of enum?

nielsvanvelzen:
If it is only SyncPlay that uses "ShuffleMode" right now we can just keep it like that for SyncPlay and use "OrderMode" for everything else.

Renaming is something we want to avoid because that would be a breaking change in the API.

dkacperski97:
I renamed `ShuffleMode` to `OrderMode` and I added `SetOrderMode` command.

crobibero (approved):

MediaBrowser.Model/Session/GeneralCommandType.cs
```diff
@@ -48,6 +48,7 @@ public enum GeneralCommandType
         PlayNext = 38,
         ToggleOsdMenu = 39,
         Play = 40,
-        SetMaxStreamingBitrate = 41
+        SetMaxStreamingBitrate = 41,
+        SetOrderMode = 42
```
nielsvanvelzen, line 52: I think we can re-use `SetShuffleQueue` for now.

MediaBrowser.Model/Session/OrderMode.cs
```diff
@@ -0,0 +1,18 @@
+namespace MediaBrowser.Model.Session
+{
+    /// <summary>
+    /// Enum OrderMode.
+    /// </summary>
+    public enum OrderMode
```
nielsvanvelzen, line 6: The enum would be called "PlaybackOrder" if we want to stay consistent with my example. With two values: Default and Shuffle

MediaBrowser.Model/Session/GeneralCommandType.cs
dkacperski97: I think the client code would get really confusing without the new command if we only add PlaybackOrder as a field: 
image
It would look like the setter is changing one field, but getter is returning another.

By adding the new command, we could start gradually replacing `SetShuffleQueue` in client code. When would be better time to change it than now, when we will be changing getters for shuffleMode in clients?

nielsvanvelzen: Fair enough, let's add it!

dkacperski97:
I renamed `OrderMode` to `PlaybackOrder` and I renamed added `SetOrderMode` command to `SetPlaybackOrder.`

nielsvanvelzen (approved):

Bond-009 (approved):

rasmuslos:
Just to be sure, this does not affect the `SetShuffleQueue` session command, right?

PR #10981, opened 2024-02-08
Add IListingsManager Service

barronpm:
**Changes**
Adds `IListingsManager` service to manage listings providers and map their channels to tuner channels


src/Jellyfin.LiveTv/Listings/ListingsManager.cs
```diff
@@ -0,0 +1,470 @@
...
+    }
+
+    /// <inheritdoc />
+    public async Task<ListingsProviderInfo> SaveListingProvider(ListingsProviderInfo info, bool validateLogin, bool validateListings)
+    {
+        ArgumentNullException.ThrowIfNull(info, nameof(info));
+
+        // Hack to make the object a pure ListingsProviderInfo instead of an AddListingProvider
+        // ServerConfiguration.SaveConfiguration crashes during xml serialization for AddListingProvider
+        info = JsonSerializer.Deserialize<ListingsProviderInfo>(JsonSerializer.SerializeToUtf8Bytes(info))!;
```
barronpm, line 62: Can this be removed? It looks like this is only called by the API endpoint, which takes a `ListingsProviderInfo` already.

src/Jellyfin.LiveTv/Listings/ListingsManager.cs
```diff
@@ -0,0 +1,470 @@
...
+    }
+
+    private static bool IsListingProviderEnabledForTuner(ListingsProviderInfo info, string tunerHostId)
+    {
+        if (info.EnableAllTuners)
+        {
+            return true;
+        }
+
+        ArgumentException.ThrowIfNullOrWhiteSpace(tunerHostId, nameof(tunerHostId));
```
crobibero, line 313:
```suggestion
        ArgumentException.ThrowIfNullOrWhiteSpace(tunerHostId);
```

Shadowghost (approved):

crobibero (approved):

MediaBrowser.Controller/LiveTv/IListingsManager.cs
```diff
@@ -0,0 +1,79 @@
...
+public interface IListingsManager
+{
+    /// <summary>
+    /// Saves the listing provider.
+    /// </summary>
+    /// <param name="info">The listing provider information.</param>
+    /// <param name="validateLogin">A value indicating whether to validate login.</param>
+    /// <param name="validateListings">A value indicating whether to validate listings..</param>
+    /// <returns>Task.</returns>
+    Task<ListingsProviderInfo> SaveListingProvider(ListingsProviderInfo info, bool validateLogin, bool validateListings);
```
Bond-009, line 23: Should we append Async to these function names while we're moving them?

barronpm: Generally I don't add the Async suffix unless there's a sync overload, but if we want to establish that as a convention we can do that.

Bond-009: IMO we should, makes it easier to see when a Task isn't properly awaited during code review

barronpm: Do analyzers/warnings not cover that already? I would expect build to fail if someone didn't await a method like that

Bond-009: They should, but in the past things have slipped through

barronpm: Alright, in that case I have one more queued PR for this sort of refactoring, then I can make a PR that applies that suffix everywhere in LiveTV

PR #10988, opened 2024-02-10
Fix Taiwan,China's display name

nyanmisaka:
Some users have pointed out to me these inaccurate names left over from Emby.
Make it correct now to avoid unnecessary controversy.

**Changes**
- Fix Taiwan,China's display name (As per https://en.wikipedia.org/wiki/Taiwan,_China)

**Issues**
emby
jf

### DIFF
--- a/Emby.Server.Implementations/Localization/countries.json
+++ b/Emby.Server.Implementations/Localization/countries.json
@@ -738,7 +738,7 @@
         "TwoLetterISORegionName": "SY"
     },
     {
-        "DisplayName": "Taiwan",
+        "DisplayName": "Taiwan, China",
         "Name": "TW",
         "ThreeLetterISORegionName": "TWN",
         "TwoLetterISORegionName": "TW"


oddstr13:
This change, in and of itself is a controversial one, in my humble opinion.

nyanmisaka:
> This change, in and of itself is a controversial one, in my humble opinion.

Nope. I don't think so. Ambiguous attitudes like yours are the root of the problem.

oddstr13:
The root of the problem here is an international territorial dispute.

Quoting the two first paragraphs of the Wikipedia article you linked, [Taiwan, China](https://en.wikipedia.org/wiki/Taiwan,_China);

> "Taiwan, China", "Taiwan, Province of China", and "Taipei, China" are controversial political terms that claim [Taiwan and its associated territories](https://en.wikipedia.org/wiki/Free_area_of_the_Republic_of_China) as a province or territory of the [People's Republic of China](https://en.wikipedia.org/wiki/China).
>
> The term "Taiwan, China" ([Chinese](https://en.wikipedia.org/wiki/Simplified_Chinese_characters): 中国台湾) is used by [Chinese state media](https://en.wikipedia.org/wiki/Chinese_state_media) even though the People's Republic of China – which is widely [recognized](https://en.wikipedia.org/wiki/United_Nations_General_Assembly_Resolution_2758) by the international community as the legitimate representative of "China" – does not exercise jurisdiction over Taiwan and other islands controlled by the [Republic of China](https://en.wikipedia.org/wiki/Taiwan) (ROC).


See also [Taiwan - Political status](https://en.wikipedia.org/wiki/Taiwan#Political_and_legal_status) and related articles.

nyanmisaka:
You're trying to introduce irrelevant politics into an open source project.

Gylesie:
I must agree with [oddstr13](https://github.com/oddstr13). Touching this subject is inherently political, and other subjects that provide service globally have been handling this subject carefully as well. Personally, I very much welcome this change. However, would there be any implications for Chinese users with this change?

nyanmisaka:
> I must agree with [oddstr13](https://github.com/oddstr13). Touching this subject is inherently political, and other subjects that provide service globally have been handling this subject carefully as well. Personally, I very much welcome this change. However, would there be any implications for Chinese users with this change?

Thanks for your input. I was born and live in mainland China. TBH I'm not interested in these political topics. I believe that I try to respect and treat everyone equally. I enjoy my hard-earned peace, so I firmly oppose actions that undermine national unity, such as placing the names of administrative provinces and countries side by side. In addition, from a racial perspective, we are all Chinese.

So avoiding talking about this will only exacerbate the gap between us, which should not be what open source developers and users want to see.

Shadowghost:
Regardless of the politics, we could circumvent this (and others like this) issue by change the naming from `Country` to `Region` or `Culture` and keep it as is - geographically `Taiwan` is correct term (it is the island of Taiwan).

nyanmisaka:
> Regardless of the politics, we could circumvent this issue (and others like this) by change the naming from `Country` to `Region` `Location` or `Culture` and keep it as is - geographically `Taiwan` is correct term (it is the island of Taiwan).

I personally accept this compromise solution.

Using `Country/Region` in the Web UI instead of only `Country` can avoid such controversy once and for all.

nyanmisaka:
Hopefully we can solve this in https://github.com/jellyfin/jellyfin-web/pull/5194
Thanks again for your discussions and suggestions.

sky-bro:
Why there's no China under Country/Region drop down menu??

nyanmisaka:
> Why there's no China under Country/Region drop down menu??

PRC.

PR #11010, opened 2024-02-14
Add missing MIME types for comicbook formats

### LINKED ISSUE #11009: [Issue]: Wrong MIME type returned for different comic book formats
### Please describe your bug

As described in https://github.com/jellyfin/jellyfin-plugin-opds/issues/38 , I noticed that the current MIME type implementation does not account for different comic book formats. It will always return `application/x-cbr`.

Sadly `cbr` (a.k.a. Rar) is a proprietary compression algorithm, so most FOSS eReaders won't be able to download or display such files.

I couldn't pinpoint the exact place where this goes wrong, but I noticed that this case is not covered by the tests in `MimeTypeTests.cs`. However, I also couldn't find a place where `x-cbr` is explicitly set, so I'm still not sure where this bug is coming from.

### Reproduction Steps

Is reproducible by adding `cbr` and `cbz` to the `MimeTypesTests` class like this:

```csharp
        [InlineData(".cbr", "application/x-cbr")]
        [InlineData(".cbz", "application/x-cbz")]
```

```text
[xUnit.net 00:00:00.73]     Jellyfin.Model.Tests.Net.MimeTypesTests.GetMimeType_Valid_ReturnsCorrectResult(input: ".cbz", expectedResult: "application/x-cbz") [FAIL]
  Fehler Jellyfin.Model.Tests.Net.MimeTypesTests.GetMimeType_Valid_ReturnsCorrectResult(input: ".cbz", expectedResult: "application/x-cbz") [1 ms]
  Fehlermeldung:
   Assert.Equal() Failure: Strings differ
                           ↓ (pos 16)
Expected: "application/x-cbz"
Actual:   "application/x-cbr"
                           ↑ (pos 16)
```

### Jellyfin Version

10.8.13

### if other:

_No response_

### Environment

```markdown
- OS:
- Linux Kernel:
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

```shell
-
```

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

rluetzner:
Until now all comicbook formats would identify themselves as `application/x-cbr`, which

1. is a deprecated MIME type and
2. indicates a comicbook compressed with the proprietary rar algorithm.

The latter point especially causes trouble with FOSS eReader software, which will refuse to open files that are seemingly rar compressed.

The two MIME types that are introduced here,

- `application/vnd.comicbook-rar` for cbr files and
- `application/vnd.comicbook+zip` for cbz files,

are officially listed by the Internet Assigned Numbers Authority (IANA) and will ensure that cbz files are properly announced as such.

Additionally, this adds a few missing deprecated MIME types to the backwards conversion from MIME type to file extension. This allows us to map MIME types to cb7, cba and cbt files. It also fixes an error with `application/x-cbr`, which would wrongly be identified as matching a file ending of cb7.

All of this works around bugs in the MimeTypes package Jellyfin uses. This should be resolved by [this pull request](https://github.com/khellang/MimeTypes/pull/24), but it's unclear how long this might take.

This fixes #11009 .


MediaBrowser.Model/Net/MimeTypes.cs
```diff
@@ -98,6 +100,12 @@ public static class MimeTypes
         private static readonly Dictionary<string, string> _extensionLookup = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
         {
             // Type application
+            { "application/vnd.comicbook-rar", ".cbr" },
+            { "application/vnd.comicbook+zip", ".cbz" },
+            { "application/x-cb7", ".cb7" },
+            { "application/x-cba", ".cba" },
+            { "application/x-cbr", ".cbr" },
+            { "application/x-cbt", ".cbt" },
```
crobibero, line 108: Why not add all of these to `_mimeTypeLookup`?

rluetzner: Good catch. I only did this, once I noticed the mismatch between `application/x-cbr` and the returned filtype `cbz`. I'll add the missing ones. :slightly_smiling_face:

rluetzner: Done.

crobibero (approved):

nielsvanvelzen (approved):

Shadowghost (approved):

Bond-009 (approved):

PR #11014, opened 2024-02-15
feat: Fully support hardware filters on Apple VideoToolbox

gnattu:
**Changes**

The upstream ffmpeg 6.1 released with a new `scale_vt` filter which provides support for scaling and tone mapping using VideoToolBox's native method.  This PR:
- Enables hardware scale filter if possible
- Enables hardware tone mapping is possible

One thing to note is that the VideoToolBox's tone mapping method has to use hard-coded Apple's parameters and does not have any tunable like other filters, but Apple's settings looks good enough to me.

Currently the WebUI does not have the options for enabling Tone Mapping for VideoToolBox, which need to be implemented separately as it does not have the tunable options available to other implementations.

**Issues**

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -6094,7 +6094,9 @@ namespace MediaBrowser.Controller.MediaEncoding
             // Hardware surface only make sense when interop with OpenCL
             // VideoToolbox's Hardware surface in ffmpeg is not only slower than hwupload, but also breaks HDR in many cases.
             // For example: https://trac.ffmpeg.org/ticket/10884
-            var useOclToneMapping = !IsVideoToolboxTonemapAvailable(state, options) && IsHwTonemapAvailable(state, options);
+            var useOclToneMapping = !IsVideoToolboxTonemapAvailable(state, options) &&
+                                    options.EnableTonemapping &&
+                                    state.VideoStream.VideoRangeType == VideoRangeType.DOVI;
             var useHwSurface = useOclToneMapping && IsVideoToolBoxFullSupported() && _mediaEncoder.SupportsFilter("alphasrc");
 
             if (is8bitSwFormatsVt)
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -2970,14 +2970,6 @@ namespace MediaBrowser.Controller.MediaEncoding
                 arg2 = (isSizeFixed ? ':' : '=') + arg2;
             }
 
-            if (string.Equals(hwScaleSuffix, "vt", StringComparison.OrdinalIgnoreCase))
-            {
-                // VideoToolBox scaling filter requires different syntax
-                arg1 = isSizeFixed ? ("=" + outWidth.Value + ":" + outHeight.Value) : string.Empty;
-                // VideoToolBox does format conversion automatically
-                arg2 = string.Empty;
-            }
-
             if (!string.IsNullOrEmpty(hwScaleSuffix) && (isSizeFixed || isFormatFixed))
             {
                 return string.Format(
@@ -5026,11 +5018,13 @@ namespace MediaBrowser.Controller.MediaEncoding
             newfilters.Add("hwupload");
             if (supportsHwScale)
             {
-                var hwScaleFilter = GetHwScaleFilter("vt", string.Empty, inW, inH, reqW, reqH, reqMaxW, reqMaxH);
-                if (true)
+                var hwScaleFilter = GetHwScaleFilter("vt", null, inW, inH, reqW, reqH, reqMaxW, reqMaxH);
+                if (useHwToneMapping)
                 {
-                    hwScaleFilter = string.IsNullOrEmpty(hwScaleFilter) ? "scale_vt=0:0:bt709:bt709:bt709"
-                        : string.Format(CultureInfo.InvariantCulture, hwScaleFilter, ":bt709:bt709:bt709");
+                    var tonemapArgs = "color_matrix=bt709:color_primaries=bt709:color_transfer=bt709";
+                    hwScaleFilter = string.IsNullOrEmpty(hwScaleFilter)
+                        ? "scale_vt=" + tonemapArgs
+                        : hwScaleFilter + ":" + tonemapArgs;
                 }
 
                 newfilters.Add(hwScaleFilter);
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -272,7 +272,8 @@ namespace MediaBrowser.Controller.MediaEncoding
                 var isNvdecDecoder = vidDecoder.Contains("cuda", StringComparison.OrdinalIgnoreCase);
                 var isVaapiDecoder = vidDecoder.Contains("vaapi", StringComparison.OrdinalIgnoreCase);
                 var isD3d11vaDecoder = vidDecoder.Contains("d3d11va", StringComparison.OrdinalIgnoreCase);
-                return isSwDecoder || isNvdecDecoder || isVaapiDecoder || isD3d11vaDecoder;
+                var isVideoToolBoxDecoder = vidDecoder.Contains("videotoolbox", StringComparison.OrdinalIgnoreCase);
+                return isSwDecoder || isNvdecDecoder || isVaapiDecoder || isD3d11vaDecoder || isVideoToolBoxDecoder;
             }
 
             return state.VideoStream.VideoRange == VideoRange.HDR
@@ -4988,6 +4989,8 @@ namespace MediaBrowser.Controller.MediaEncoding
             var noOverlay = swFilterChain.OverlayFilters.Count == 0;
             var supportsHwDeint = _mediaEncoder.SupportsFilter("yadif_videotoolbox");
             var supportsHwScale = _mediaEncoder.SupportsFilter("scale_vt");
+            // VideoToolbox is special. It does not use a separate tone mapping filter like others. Instead, it performs both tone mapping and scaling in a single filter.
+            var useHwToneMapping = IsHwTonemapAvailable(state, options) && supportsHwScale;
             // fallback to software filters if we are using filters not supported by hardware yet.
             var useHardwareFilters = noOverlay && (!doDeintH2645 || supportsHwDeint);
 
@@ -5008,6 +5011,11 @@ namespace MediaBrowser.Controller.MediaEncoding
             if (supportsHwScale)
             {
                 var hwScaleFilter = GetHwScaleFilter("vt", "", inW, inH, reqW, reqH, reqMaxW, reqMaxH);
+                if (useHwToneMapping)
+                {
+                    hwScaleFilter = string.IsNullOrEmpty(hwScaleFilter) ? "scale_vt=0:0:bt709:bt709:bt709"
+                        : string.Format(CultureInfo.InvariantCulture, hwScaleFilter, ":bt709:bt709:bt709");
+                }
                 newfilters.Add(hwScaleFilter);
             }
 
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -309,7 +309,7 @@ namespace MediaBrowser.Controller.MediaEncoding
                    && state.VideoStream.VideoRangeType == VideoRangeType.HDR10;
         }
 
-        private bool IsVideoToolboxVppTonemapAvailable(EncodingJobInfo state, EncodingOptions options)
+        private bool IsVideoToolboxTonemapAvailable(EncodingJobInfo state, EncodingOptions options)
         {
             if (state.VideoStream is null
                 || !options.EnableVideoToolboxTonemapping
@@ -5003,7 +5003,7 @@ namespace MediaBrowser.Controller.MediaEncoding
             var supportsHwDeint = _mediaEncoder.SupportsFilter("yadif_videotoolbox");
             var supportsHwScale = _mediaEncoder.SupportsFilter("scale_vt");
             // VideoToolbox is special. It does not use a separate tone mapping filter like others. Instead, it performs both tone mapping and scaling in a single filter.
-            var useHwToneMapping = IsVideoToolboxVppTonemapAvailable(state, options) && supportsHwScale;
+            var useHwToneMapping = IsVideoToolboxTonemapAvailable(state, options) && supportsHwScale;
             // fallback to software filters if we are using filters not supported by hardware yet.
             var useHardwareFilters = noOverlay && (!doDeintH2645 || supportsHwDeint);
 
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -5019,7 +5019,7 @@ namespace MediaBrowser.Controller.MediaEncoding
             var doOclTonemap = !doVtTonemap && IsHwTonemapAvailable(state, options);
 
             var scaleFormat = string.Empty;
-            if (GetVideoColorBitDepth(state) == 10)
+            if (!string.Equals(state.VideoStream.PixelFormat, "yuv420p", StringComparison.OrdinalIgnoreCase))
             {
                 // Use P010 for OpenCL tone mapping, otherwise force an 8bit output.
                 scaleFormat = doOclTonemap ? "p010le" : "nv12";
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -2954,6 +2954,13 @@ namespace MediaBrowser.Controller.MediaEncoding
             {
                 arg2 = (isSizeFixed ? ':' : '=') + arg2;
             }
+            if (string.Equals(hwScaleSuffix, "vt", StringComparison.OrdinalIgnoreCase))
+            {
+                // VideoToolBox scaling filter requires different syntax
+                arg1 = isSizeFixed ? ("=" + outWidth.Value + ":" + outHeight.Value) : string.Empty;
+                // VideoToolBox does format conversion automatically
+                arg2 = string.Empty;
+            }
 
             if (!string.IsNullOrEmpty(hwScaleSuffix) && (isSizeFixed || isFormatFixed))
             {
@@ -4980,6 +4987,7 @@ namespace MediaBrowser.Controller.MediaEncoding
             var newfilters = new List<string>();
             var noOverlay = swFilterChain.OverlayFilters.Count == 0;
             var supportsHwDeint = _mediaEncoder.SupportsFilter("yadif_videotoolbox");
+            var supportsHwScale = _mediaEncoder.SupportsFilter("scale_vt");
             // fallback to software filters if we are using filters not supported by hardware yet.
             var useHardwareFilters = noOverlay && (!doDeintH2645 || supportsHwDeint);
 
@@ -4988,14 +4996,20 @@ namespace MediaBrowser.Controller.MediaEncoding
                 return swFilterChain;
             }
 
-            // ffmpeg cannot use videotoolbox to scale
-            var swScaleFilter = GetSwScaleFilter(state, options, vidEncoder, inW, inH, threeDFormat, reqW, reqH, reqMaxW, reqMaxH);
-            newfilters.Add(swScaleFilter);
-
+            if (!supportsHwScale)
+            {
+                var swScaleFilter = GetSwScaleFilter(state, options, vidEncoder, inW, inH, threeDFormat, reqW, reqH, reqMaxW, reqMaxH);
+                newfilters.Add(swScaleFilter);
+            }
             // hwupload on videotoolbox encoders can automatically convert AVFrame into its CVPixelBuffer equivalent
             // videotoolbox will automatically convert the CVPixelBuffer to a pixel format the encoder supports, so we don't have to set a pixel format explicitly here
             // This will reduce CPU usage significantly on UHD videos with 10 bit colors because we bypassed the ffmpeg pixel format conversion
             newfilters.Add("hwupload");
+            if (supportsHwScale)
+            {
+                var hwScaleFilter = GetHwScaleFilter("vt", "", inW, inH, reqW, reqH, reqMaxW, reqMaxH);
+                newfilters.Add(hwScaleFilter);
+            }
 
             if (doDeintH2645)
             {
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -253,6 +253,14 @@ namespace MediaBrowser.Controller.MediaEncoding
                    && _mediaEncoder.SupportsFilterWithOption(FilterOptionType.OverlayVulkanFrameSync);
         }
 
+        private bool IsVideoToolBoxFullSupported()
+        {
+            return _mediaEncoder.SupportsHwaccel("videotoolbox")
+                && _mediaEncoder.SupportsFilter("yadif_videotoolbox")
+                && _mediaEncoder.SupportsFilter("overlay_videotoolbox")
+                && _mediaEncoder.SupportsFilter("scale_vt");
+        }
+
         private bool IsHwTonemapAvailable(EncodingJobInfo state, EncodingOptions options)
         {
             if (state.VideoStream is null
@@ -272,7 +280,8 @@ namespace MediaBrowser.Controller.MediaEncoding
                 var isNvdecDecoder = vidDecoder.Contains("cuda", StringComparison.OrdinalIgnoreCase);
                 var isVaapiDecoder = vidDecoder.Contains("vaapi", StringComparison.OrdinalIgnoreCase);
                 var isD3d11vaDecoder = vidDecoder.Contains("d3d11va", StringComparison.OrdinalIgnoreCase);
-                return isSwDecoder || isNvdecDecoder || isVaapiDecoder || isD3d11vaDecoder;
+                var isVideoToolBoxDecoder = vidDecoder.Contains("videotoolbox", StringComparison.OrdinalIgnoreCase);
+                return isSwDecoder || isNvdecDecoder || isVaapiDecoder || isD3d11vaDecoder || isVideoToolBoxDecoder;
             }
 
             return state.VideoStream.VideoRange == VideoRange.HDR
@@ -317,7 +326,7 @@ namespace MediaBrowser.Controller.MediaEncoding
                 return false;
             }
 
-            // Certain DV profile 5 video works in Safari with direct playing, but the VideoToolBox does not produce correct mapping results with trascoding.
+            // Certain DV profile 5 video works in Safari with direct playing, but the VideoToolBox does not produce correct mapping results with transcoding.
             // All other HDR formats working.
             return state.VideoStream.VideoRange == VideoRange.HDR
                    && state.VideoStream.VideoRangeType is VideoRangeType.HDR10 or VideoRangeType.HLG or VideoRangeType.HDR10Plus;
@@ -4976,12 +4985,6 @@ namespace MediaBrowser.Controller.MediaEncoding
                 return swFilterChain;
             }
 
-            if (_mediaEncoder.EncoderVersion.CompareTo(new Version("5.0.0")) < 0)
-            {
-                // All features used here requires ffmpeg 5.0 or later, fallback to software filters if using an old ffmpeg
-                return swFilterChain;
-            }
-
             var doDeintH264 = state.DeInterlace("h264", true) || state.DeInterlace("avc", true);
             var doDeintHevc = state.DeInterlace("h265", true) || state.DeInterlace("hevc", true);
             var doDeintH2645 = doDeintH264 || doDeintHevc;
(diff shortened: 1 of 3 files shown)


MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -2955,6 +2968,14 @@ public MediaStream GetMediaStream(IEnumerable<MediaStream> allStream, int? desir
...
             }
 
+            if (string.Equals(hwScaleSuffix, "vt", StringComparison.OrdinalIgnoreCase))
+            {
+                // VideoToolBox scaling filter requires different syntax
+                arg1 = isSizeFixed ? ("=" + outWidth.Value + ":" + outHeight.Value) : string.Empty;
+                // VideoToolBox does format conversion automatically
+                arg2 = string.Empty;
+            }
+
```
nyanmisaka, line 2978: These can actually be removed, see my other comments below.
```suggestion
```

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -4988,14 +5012,27 @@ public string GetHwTonemapFilter(EncodingOptions options, string hwTonemapSuffix
...
             // This will reduce CPU usage significantly on UHD videos with 10 bit colors because we bypassed the ffmpeg pixel format conversion
             newfilters.Add("hwupload");
+            if (supportsHwScale)
+            {
+                var hwScaleFilter = GetHwScaleFilter("vt", string.Empty, inW, inH, reqW, reqH, reqMaxW, reqMaxH);
+                if (useHwToneMapping)
+                {
+                    hwScaleFilter = string.IsNullOrEmpty(hwScaleFilter) ? "scale_vt=0:0:bt709:bt709:bt709"
+                        : string.Format(CultureInfo.InvariantCulture, hwScaleFilter, ":bt709:bt709:bt709");
+                }
```
nyanmisaka, line 5032: For hardware filters, it is best to use the option name explicitly.
Because these will go wrong once upstream changes the order of options.

```suggestion
                var hwScaleFilter = GetHwScaleFilter("vt", null, inW, inH, reqW, reqH, reqMaxW, reqMaxH);
                if (useHwToneMapping)
                {
                    var tonemapArgs = "color_matrix=bt709:color_primaries=bt709:color_transfer=bt709";
                    hwScaleFilter = string.IsNullOrEmpty(hwScaleFilter)
                        ? "scale_vt=" + tonemapArgs
                        : hwScaleFilter + ":" + tonemapArgs;
                }
```

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -272,7 +272,8 @@ private bool IsHwTonemapAvailable(EncodingJobInfo state, EncodingOptions options
                 var isNvdecDecoder = vidDecoder.Contains("cuda", StringComparison.OrdinalIgnoreCase);
                 var isVaapiDecoder = vidDecoder.Contains("vaapi", StringComparison.OrdinalIgnoreCase);
                 var isD3d11vaDecoder = vidDecoder.Contains("d3d11va", StringComparison.OrdinalIgnoreCase);
-                return isSwDecoder || isNvdecDecoder || isVaapiDecoder || isD3d11vaDecoder;
+                var isVideoToolBoxDecoder = vidDecoder.Contains("videotoolbox", StringComparison.OrdinalIgnoreCase);
+                return isSwDecoder || isNvdecDecoder || isVaapiDecoder || isD3d11vaDecoder || isVideoToolBoxDecoder;
             }
```
nyanmisaka, line 277:
```
            if (string.Equals(state.VideoStream.Codec, "hevc", StringComparison.OrdinalIgnoreCase)
                && state.VideoStream.VideoRange == VideoRange.HDR
                && state.VideoStream.VideoRangeType == VideoRangeType.DOVI)
...
```

Does it really handle the DoVi Profile 5 dynamic metadata?

nyanmisaka (changes requested):

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
gnattu: I'm still modifying the implementation and verifying each format. Unfortunately the DoVi Profile 5 only works for direct playing and the transcoding will produce incorrect color. HDR10, HDR10+ and HLG are verified to work.

These are leftovers to be removed as the tone mapping option is separated to make the front-end clients' config easier.

nyanmisaka: We already support remuxing DoVi P5. While tonemapping and transcoding it is only available in libplacebo and our custom CUDA and OpenCL filters. I don't think Apple will enable DoVi support in videotoolbox filters because for commercial products it usually requires licensing from Dolby. Also, HDR10 and HLG to SDR are easy because they use static metadata.

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -4980,22 +4995,40 @@ public string GetHwTonemapFilter(EncodingOptions options, string hwTonemapSuffix
             var newfilters = new List<string>();
             var noOverlay = swFilterChain.OverlayFilters.Count == 0;
             var supportsHwDeint = _mediaEncoder.SupportsFilter("yadif_videotoolbox");
+            var supportsHwScale = _mediaEncoder.SupportsFilter("scale_vt");
+            // VideoToolbox is special. It does not use a separate tone mapping filter like others. Instead, it performs both tone mapping and scaling in a single filter.
+            var useHwToneMapping = IsVideoToolboxTonemapAvailable(state, options) && supportsHwScale;
             // fallback to software filters if we are using filters not supported by hardware yet.
-            var useHardwareFilters = noOverlay && (!doDeintH2645 || supportsHwDeint);
+            var useHardwareFilters = noOverlay && (!doDeintH2645 || supportsHwDeint || supportsHwScale);
```
nyanmisaka, line 5002: So you don’t want users to enable tone-mapping and subtitle burn-in at the same time? It does result in a more complex filter pipeline, but I know many users want it.

gnattu: This decision was made a long time ago when the only hardware filter available was yadif.

It serves as a fallback to the safest software chain, as the hardware filter currently does not handle overlay.

However, our software filter chain also cannot perform tone mapping. So, to enable tone mapping functionality with subtitles burn-in, I propose do an `hwdownload` to yuv420 and attaching the software subtitle filters to the chain. Would this approach be acceptable? Or do I have to do something more.

nyanmisaka: We are not in a hurry, I would rather have a complete hardware pipeline. I'm actually more interested in whether Apple's fixed-function hardware supports overlay/image blending, or we can achieve through a Metal kernel.

If you are interested in it, we can try writing an `overlay_vt/videotoolbox` filter for ffmpeg (refer to [`overlay_cuda.cu`](https://github.com/FFmpeg/FFmpeg/blob/master/libavfilter/vf_overlay_cuda.cu) and port it to `overlay_videotoolbox.metal`, we already have a [`yadif_videotoolbox.metal`](https://github.com/FFmpeg/FFmpeg/blob/master/libavfilter/metal/vf_yadif_videotoolbox.metal) that can be used as a template). I don't have Apple hardware but can provide you with technical support in writing code for FFmpeg. Once completed, submit it to upstream FFmpeg and everything will be perfect.

gnattu: I'm currently working on a Metal implementation of the overlay filter. Apple's VideoToolbox does not expose any image blending function, so I'm required to use a custom Metal kernel. I already have a working kernel that performs basic alpha compositing and positioning, and I'm in the process of porting it to ffmpeg.

In examining the current usage pattern in Jellyfin, I've observed that most hardware overlay filters are used with an additional custom `alphaSrc` filter. I'm curious about the purpose of this filter. Perhaps I can adjust my implementation to eliminate the need for it and instead let the overlay filter to handle the issue more effectively.

nyanmisaka: You can't just consider PGS graphical subtitles. `alphasrc` is used in conjunction with the subtitles filter to convert SSA/ASS/SRT text subtitles to images. This is unavoidable.

gnattu: I have made some progress and the filter produces outputs(although not correct) in some cases. A big problem I found is that we cannot `hwupload` arbitrary image because the `HWContext` of `VideoToolBox` does not support common colorspaces seen in a lot of images like `rgba` used by pngs. Is this normal for HW overlay filters and could I do anything to convert that to something `VideoToolBox` could recognize(like yuv)?

nyanmisaka: The formats that can be used in HW filters are usually restricted in the [hwcontext_videotoolbox](https://github.com/FFmpeg/FFmpeg/blob/0895ef0d6d6406ee6cd158fc4d47d80f201b8e9c/libavutil/hwcontext_videotoolbox.c#L48).

For the subtitle burn-in use case, we need `main (nv12/yuv420p) + overlay (bgra/rgba/yuva420p)`, which means that the `overlay` input must contain at least an alpha channel.

Also, since the `scale_vt` filter (fixed-function HW) is not capable of pixel format conversion, you may need to add a [RGB2YUV/YUV2RGB](https://github.com/FFmpeg/FFmpeg/blob/0895ef0d6d6406ee6cd158fc4d47d80f201b8e9c/libavfilter/opencl/colorspace_common.cl#L137) function in the Metal kernel of `overlay_vt` to handle this.

nyanmisaka: It seems Apple [supports](https://developer.apple.com/documentation/corevideo/kcvpixelformattype_32rgba) `rgba` so you can add it. The difference between it and `bgra` is the channel order. Note that PGS subtitles natively use `bgra`.

If I understand correctly, [VTPixelTransferSessionTransferImage](https://developer.apple.com/documentation/videotoolbox/1503548-vtpixeltransfersessiontransferim) allows you to `convert` pixel formats. Assuming the vt encoders also support `bgra`. Then `overlay_vt` can be simplified to (bgra+bgra).

```
// pseudo filter pipeline
[0:s]hwupload[sub] // bgra, from pgssub or alphasrc
[0:v]scale_vt=format=bgra[main] // add convert support for `scale_vt`
[main][sub]overlay_vt[out] //bgra to encoder
```

gnattu: I *think* I got it. That method is calling `CVMetalTextureCacheCreateTextureFromImage` with width and height calculated from `CVPixelBufferGetWidthOfPlane`, but RGBA buffers are not planer, so that method would return 0, hence the `kCVReturnInvalidArgument`. I think I need to fix that first.

nyanmisaka: Only the video decoder generates yuv interlaced videos, so the author of `yadif_vt` probably only tested planar formats.

gnattu:
It now works with incorrect colorspace (as the main is in YUV and the overlay is in BGRA):

image

To handle the colorspace, we have multiple approaches here, and this could be very specific to `videotoolbox`, as the other approaches like the CUDA implementation assume the two layers have matching pixel formats, so they do not perform any conversion.

The most native way would be to use [CoreImage](https://developer.apple.com/documentation/coreimage?language=objc), generate an intermediate image with [`-[CIImage initWithCVPixelBuffer:]`](https://developer.apple.com/documentation/coreimage/ciimage/1438072-initwithcvpixelbuffer), then set [`kCIContextWorkingColorSpace`](https://developer.apple.com/documentation/coreimage/kcicontextworkingcolorspace) to the `CIContext`, and finally render it into another `CVPixelBuffer` with [`-[CIContext render:toCVPixelBuffer:bounds:colorSpace:]`](https://developer.apple.com/documentation/coreimage/cicontext/1437835-render). This approach could give us the most correct result, as it properly handles gamma correction, but having an intermediate image for every frame could be slower. (Note: CoreImage does have a Metal backend, but ffmpeg's own CoreImage filter only uses the old OpenGL backend. If I'm going with this approach, I will create my own `CIContext` with Metal.)

Another approach we could try is to bypass the `HWContenxt` of VideoToolBox and use a custom matrix to generate a `texture2d` with YCbCr, with an alpha channel. The problem here is that the colors could be wrong if we just use a basic matrix for such conversion. The matrix I plan to use is the following:

```
const float4x4 rgbaToYcbcrTransform = float4x4(
   float4(+0.2990, -0.1687, +0.5000, +0.0000),
   float4(+0.5870, -0.3313, -0.4187, +0.0000),
   float4(+0.1140, +0.5000, -0.0813, +0.0000),
   float4(+0.0000, +0.5000, +0.5000, +1.0000)
);
```

I'm going to try the CoreImage first to see if the performance is acceptable.

gnattu:
The CoreImage approach had initial success with color:
image

A big problem is that the frames appears not to be "in order", and it the video seems to be "vibrating" back and forth. What could be the cause of this? My assumption is that the frame_sync has something happened, but I'm not quite sure on how to fix that.

gnattu:
Attaching the example file for reference.
[test.mp4.zip](https://github.com/jellyfin/jellyfin/files/14322231/test.mp4.zip)

gnattu:
By adding `usleep(1600);` into the blending function body made the frame in-sync, so it seems like the function is called before the previous frame is ready? Why would this happen?

nyanmisaka:
ffmpeg's `frame_sync` utils only guarantees that frames are synchronized in timestamps, but does not guarantee that the underlying video data is synchronized.

nyanmisaka:
Perhaps you need some calls from CoreImage to ensure the video data has been synchronized correctly, similar to glflush()/glfinish() in OpenGL.

gnattu:
I got it. CoreImage defers rendering until users request access to it, that's why the frame is out of sync. I have to use `CVPixelBufferLockBaseAddress` to lock the rendered pixelbuffer first, and now the video is synced properly.

[test_fixed.mp4.zip](https://github.com/jellyfin/jellyfin/files/14323649/test_fixed.mp4.zip)

nyanmisaka:
Good catch. But i can still find some vibrating frames in this clip. Especially some frames at the beginning and end.

gnattu:
At this point it seems like it is unlikely a frame syncing issue in our filter. If I switch to `h264_videotoolbox` then there will be only vibrating frames at the beginning even I removed all  `CVPixelBufferLockBaseAddress`. It seems like the frame we output overloads the downstream pipeline which caused out of sync frames to be generated. 

If I forced it to run at a very slow speed (like 30fps), then even `hevc_videotoolbox` will not have any out of sync frames. 

Probably we should not output BGRA frames directly, but convert them back to planer YUV to make the hardware happier?

Or we need to find some way to throttle the frame output so that we never overloads the downstream.

[test.mp4_with_throttle.zip](https://github.com/jellyfin/jellyfin/files/14323882/test.mp4_with_throttle.zip)

nyanmisaka:
There is no overload in libavfilter. HW encoders can be overloaded but this is usually mitigated by dropping frames, some encoders allow this.

Such uncertainty means that there is still an out-of-sync somewhere in the overlay filter or the hwcontext of videotoolbox. It's best to start by reviewing your code.

gnattu:
By converting the output frame back to yuv fixed this, so this is indeed the behavior of `videotoolbox` when you feed it with bgra frames too quickly.  
[test.mp4_output_yuv.zip](https://github.com/jellyfin/jellyfin/files/14324778/test.mp4_output_yuv.zip)

nyanmisaka:
So the input and output of your filter was (yuv+bgra=>bgra)?
And you changed it to (yuv+bgra=>yuv).

gnattu:
Yes. That is what I did.

nyanmisaka:
Then it’s indeed an issue of the videotoolbox encoder when it comes to handling non-yuv input.

A common practice for an overlay filter is to keep the format of the output and main input the same, so the latter one is better.

gnattu:
The initial working tree is pushed here:
https://github.com/gnattu/jellyfin-ffmpeg/commit/587a97ae02c8052c65d5c266d6048539b9c0eebd

Before submitting this upstream, I'd like to receive some suggestions first.

nyanmisaka:
> The initial working tree is pushed here: [gnattu/jellyfin-ffmpeg@587a97a](https://github.com/gnattu/jellyfin-ffmpeg/commit/587a97ae02c8052c65d5c266d6048539b9c0eebd)
> 
> Before submitting this upstream, I'd like to receive some suggestions first.

Looks good at first glance. I left some suggestions.

gnattu:
Pushed v2 patch:
https://github.com/gnattu/jellyfin-ffmpeg/commit/d2276b009bc71cffc59477fe30bb3b0e11fe22d5

Now I'm thinking about using `VTPixelTransferSessionTransferImage` for every conversion by default. My initial testing shows that it is much slower than `CoreImage`, but after further testing I got no measurable performance difference, and `VTPixelTransferSessionTransferImage` handles more formats than `CoreImage`. The CoreImage can be used as a compatibility fallback, because `VTPixelTransferSessionTransferImage` is not available until iOS 16.

gnattu:
It seems like we have to manually set the timebase in `configure_output` like below:

```
link->time_base = avctx->inputs[0]->time_base;
ctx->fs.time_base = link->time_base;
```

The problem seems to gone with this.

gnattu:
This does not fix all combination. I'm confused.

gnattu:
Latest changes pushed to branch:
https://github.com/gnattu/jellyfin-ffmpeg/tree/vf_overlay_videotoolbox

nyanmisaka:
> This does not fix all combination. I'm confused.
> 
> Well at least the output is not affected. We still have `dup=91` in the log, but the output file seems ok.

`dup` is usually a timestamp or time_base issue. I added more suggestions [here](https://github.com/gnattu/jellyfin-ffmpeg/pull/1).

gnattu:
To support Dolby Vision Profile 5, my current plan is to introduce a simple filter `hwupload_videotoolbox` to inter-op with our current OpenCL kernel. The pipeline would be `hwupload->tonemap_cl->hwdownload->hwupload_videotoolbox->other_vt_filters`. I already have a working version base on `hwupload_cuda` and verified that it would work with our `overlay_videotoolbox` filter. Should I upstream this filter or keep it into our own fork?

nyanmisaka:
> To support Dolby Vision Profile 5, my current plan is to introduce a simple filter `hwupload_videotoolbox` to inter-op with our current OpenCL kernel. The pipeline would be `hwupload->tonemap_cl->hwdownload->hwupload_videotoolbox->other_vt_filters`. This will give us a support situation closer to what we are doing for Intel now: a fallback OpenCL tone mapping path and a more performant but less compatible VT tone mapping path.
> 
> I already have a working version base on `hwupload_cuda` and verified that it would work with our `overlay_videotoolbox` filter. Should I upstream this filter or keep it into our own fork?

I'm pretty sure upstream no longer accepts filters like `hwupload_cuda`. It's a legacy thing.
Is there any chance to add this to hwcontext->map_to/map_from?

gnattu:
> Is there any chance to add this to hwcontext->map_to/map_from?

It could be easier than this. The `derive_device` function is not implemented for `videotoolbox`, and actually, VideoToolbox does not need any kind of "context initialization" like other hardware. Therefore, I can simply implement a dummy no-op method that blindly returns 0. Then, the `hwdownload,format=FORMAT,hwupload=derive_device=videotoolbox` chain would work without the need for a separate `hwupload_videotoolbox` filter.

The direct frame conversion between OpenCL and VideoToolbox seems to be non-trivial and requires the use of very dated APIs. One approach I know of is to put the OpenCL image into an `IOSurface`, then generate a CoreImage `CIImage` from that. Next, render it into a `CVPixelBuffer` so that our other VideoToolbox filters can use it. The IOSurface does not reside in GPU memory and has to be copied to CPU memory, similar to `hwdownload`, so it is probably not faster than just using `hwdownload`.

nyanmisaka:
> > Is there any chance to add this to hwcontext->map_to/map_from?
> 
> It could be easier than this. The `derive_device` function is not implemented for `videotoolbox`, and actually, VideoToolbox does not need any kind of "context initialization" like other hardware. Therefore, I can simply implement a dummy no-op method that blindly returns 0. Then, the `hwdownload,format=FORMAT,hwupload=derive_device=videotoolbox` chain would work without the need for a separate `hwupload_videotoolbox` filter.

If you are okay with this then do it. Or eventually if you want, make our `tonemap_videotoolbox` filters for DoVi P5, like I ported `tonemap_opencl` to `tonemap_cuda`.

> The direct frame conversion between OpenCL and VideoToolbox seems to be non-trivial and requires the use of very dated APIs. One approach I know of is to put the OpenCL image into an `IOSurface`, then generate a CoreImage `CIImage` from that. Next, render it into a `CVPixelBuffer` so that our other VideoToolbox filters can use it. The IOSurface does not reside in GPU memory and has to be copied to CPU memory, similar to `hwdownload`, so it is probably not faster than just using `hwdownload`.

That makes sense. But it's hard to imagine that Apple created OpenCL but didn't provide any interoperability functions for VideoToolBox and it. It was supposed to be seamless thanks to UMA graphics.

nyanmisaka:
I have some new ideas. Vulkan is the future, you may be able to use [this extension](https://github.com/KhronosGroup/MoltenVK/issues/830) to implement map_to/from of vulkan<->videotoolbox, allowing us to use DoVi P5 tone-mapping in [`vf_libplacebo`](https://github.com/haasn/libplacebo).

gnattu:
>  It's hard to imagine that Apple created OpenCL but didn't provide any interoperability functions for VideoToolBox and it. It was supposed to be seamless thanks to UMA graphics.

By the time they created OpenCL their computers are not using UMA graphics except for Intel iGPUs, and now Metal Compute is their preferred way to do things, and it is indeed seamless. As you can see from the overlay filter implementation.

> you may be able to use https://github.com/KhronosGroup/MoltenVK/issues/830 to implement map_to/from of vulkan<->videotoolbox, allowing us to use DoVi P5 tone-mapping in [vf_libplacebo](https://github.com/haasn/libplacebo)

If libplacebo can provide the same level of functionality as our own kernel then I'm fine with that, but I'm going to using the hwdownload with OpenCL for now because it requires little to no work and is verified to work, so even if the libplacebo cannot work as expected, we still have a fallback plan. Having a `derive_device` function for VideoToolbox does not hurt anyway. The vulkan path would take longer as I'm not familiar with it, but probably not too long as the extension looks straightforward to me.

gnattu:
Oh, one more question. In case any objc api is needed, can I just rename `hwcontext_videotoolbox.c` to `hwcontext_videotoolbox.m`?

nyanmisaka:
> > you may be able to use [KhronosGroup/MoltenVK#830](https://github.com/KhronosGroup/MoltenVK/issues/830) to implement map_to/from of vulkan<->videotoolbox, allowing us to use DoVi P5 tone-mapping in [vf_libplacebo](https://github.com/haasn/libplacebo)
> 
> If libplacebo can provide the same level of functionality as our own kernel then I'm fine with that, but I'm going to using the hwdownload with OpenCL for now because it requires little to no work and is verified to work, so even if the libplacebo cannot work as expected, we still have a fallback plan. Having a `derive_device` function for VideoToolbox does not hurt anyway. The vulkan path would take longer as I'm not familiar with it, but probably not too long as the extension looks straightforward to me.

AFAIK we'll be entering feature freeze for the 10.9 release soon, maybe the end of the month. If you want this to come in, you'll also need to add *jellyfin-ffmpeg 6.1 for MacOS* and include the required patches for this feature. I'm not sure it's too late.

The CL kernel of the `tonemap_opencl` filter in ffmpeg [was originally based on](https://github.com/FFmpeg/FFmpeg/blob/4ea2b271eb58c05709e5fe53bf5642b33b7be6f4/libavfilter/opencl/tonemap.cl#L80) the shader of the MPV player. The author of libplacebo is also one of the main contributors to the MPV player, and the `gpu-next` renderer of the MPV player is libplacebo. It uses GLSL and therefore has the flexibility to dynamically generate shaders, which is hard to achieve in FFmpeg CUDA/CL filters.

So we are using a stripped-down version of MPV/libplacebo shader to provide usability for a wider range of GPUs, since Vulkan is only fully supported on devices of recent years.

nyanmisaka:
> Oh, one more question. In case any objc api is needed, can I just rename `hwcontext_videotoolbox.c` to `hwcontext_videotoolbox.m`?

I don't see a blocker for this. Only the compiler really cares about it. Just add something to the commit msg that supports your argument.

gnattu:
> AFAIK we'll be entering feature freeze for the 10.9 release soon, maybe the end of the month. If you want this to come in, you'll also need to add jellyfin-ffmpeg 6.1 for MacOS and include the required patches for this feature. I'm not sure it's too late.

Then the vulkan backend will definitely miss this. The ffmpeg problem is that it seems like we are on 6.0 so I'm not sure if the patches will apply. If the ffmpeg can be built then we can have this in with OpenCL + VT tone mapping.

gnattu:
Alternatively we can just port `scale_vt` to 6.0 as a workaround if it would take too long to make ffmpeg patches set.

nyanmisaka:
There are massive changes to `hwcontext_vulkan` of FF 6.1 that break our existing AMD VAAPI->Vulkan pipeline and it will take me some time to fix and test it. In addition, upstream reports more breakage related to Vulkan in FF 6.1.

I think we shouldn't do this rashly before the release of a new Jellyfin version. Similarly, FF 7.0 may arrive someday in the next few months, and it will introduce multi-threading globally and will require more testing. Therefore it is best to backport `scale_vt` to 6.0.

nyanmisaka:
Also I'm not sure if we need some documentation for our VideoToolBox HWA, if you have some ideas about it you can refer to this Intel documentation to add one.

https://jellyfin.org/docs/general/administration/hardware-acceleration/intel/
https://github.com/jellyfin/jellyfin.org/blob/master/docs/general/administration/hardware-acceleration/intel.md

gnattu:
After I raised the interop problem to upstream, my proposal of implementing derive_device as a no-op was rejected. Instead, they came up with a real implementation of hwmap between OpenCL and VideoToolbox instead of hacking around with hwupload=derive_device:

https://ffmpeg.org//pipermail/ffmpeg-devel/2024-February/321979.html

And that implementation is almost doing what I said: rendering it into an `IOSurface` and then deriving a `CVPixelBuffer` from it. However, it does not need an intermediate `CIImage`, as the `io_surface_ref` is created directly from the `CVPixelBuffer`, allowing anything rendered to it to be accessed directly in that buffer.

nyanmisaka:
> After I raised the interop problem to upstream, my proposal of implementing derive_device as a no-op was rejected. Instead, they came up with a real implementation of hwmap between OpenCL and VideoToolbox instead of hacking around with hwupload=derive_device:
> 
> https://ffmpeg.org//pipermail/ffmpeg-devel/2024-February/321979.html

Not surprising that upstream would reject it. This is also one of the reasons why we need to have our own fork, to achieve our goals as quickly as possible with the minimum cost, rather than rewriting `fftools/ffmpeg*`.

If you want to implement it more elegantly and have this feature benefit more than just videotoolbox, you might want to add a syntax like
```c
-init_hw_device videotoolbox=vt -init_hw_device opencl=ocl ... hwupload=derive_device=videotoolbox@vt ...
```
, which also requires adding additional APIs to lavf+lavu in order to be used in `fftools/ffmpeg_filter`.

> And that implementation is almost doing what I said: rendering it into an `IOSurface` and then deriving a `CVPixelBuffer` from it. However, it does not need an intermediate `CIImage`, as the `io_surface_ref` is created directly from the `CVPixelBuffer`, allowing anything rendered to it to be accessed directly in that buffer.

`clCreateImageFromIOSurfaceWithPropertiesAPPLE` that's what I was looking for. An OpenCL func/extension from Apple. Let's assume Apple did something internally to avoid extra copies. Can you try this patch and see how it performs?

nyanmisaka:
Btw, what is Apple's current attitude towards OpenCL? I remember they deprecated CL a long time ago and switched to Metal. Have they said in what version of MacOS they will remove OpenCL completely?

gnattu:
They actually ALREADY removed native OpenCL/OpenGL in their OS a long time ago. All the OpenCL and OpenGL code now is running on a compatibility layer on top of Metal and barely receives any updates. So their attitude is that you can keep using old code as long as it does not break.

nyanmisaka:
> They actually ALREADY removed native OpenCL/OpenGL in their OS a long time ago. All the OpenCL and OpenGL code now is running on a compatibility layer on top of Metal and barely receives any updates. So their attitude is that you can keep using old code as long as it does not break.

Then this should be safe enough. They took the same path as NVIDIA and used OpenCL as the wrapper for CUDA/Metal.

gnattu:
> Can you try this patch and see how it performs?

Just tried with that patch, it works and `reverse=1` has to be used as it in a one-way mapping if we want to chain filters.

The performance improvement is not significant though, only about 2% faster than `hwdownload, hwupload` (3 times average, 182 fps vs 178 fps) for a simple filter chain: `transpose_opencl->scale_vt` , probably due to the architecture of M1 Max makes such copy very cheap.

nyanmisaka:
> > Can you try this patch and see how it performs?
> 
> Just tried with that patch, it works and `reverse=1` has to be used as it in a one-way mapping if we want to chain filters.
> 
> The performance improvement is not significant though, only about 2% faster than `hwdownload, hwupload` (3 times average, 182 fps vs 178 fps) for a simple filter chain: `transpose_opencl->scale_vt` , probably due to the architecture of M1 Max makes such copy very cheap.

I'm assuming this is 4k resolution (3840x2160ish).

It should help with some Radeon GPUs whose host<->device transfer speed is limited by PCIe bandwidth. Intel HD/UHD should also benefit since it does not have a high on-die interconnect bandwidth like on M1 Max.

nyanmisaka:
I think with this interop patch, upstream should also accept your changes in derive_device no-op.

gnattu:
> I'm assuming this is 4k resolution (3840x2160ish).

It is not, but it is a 1080p video with relatively high bitrate(30M).

After a few more rounds I fount that it is the hardware hevc hardware encoder can only do this fast so the bottleneck is not in the filter, guess I have to use some low-res video to test.

nyanmisaka:
> > I'm assuming this is 4k resolution (3840x2160ish).
> 
> It is not, but it is a 1080p video with relatively high bitrate(30M).
> 
> After a few more rounds I fount that it is the hardware hevc hardware encoder can only do this fast so the bottleneck is not in the filter, guess I have to use some low-res video to test.

Can you also share the maximum throughput of VT decoders, encoders and filters on M1 Max? Apple never discloses this in their HW specs. Are there preset options included in the encoder to allow users to trade image quality for speed?

nyanmisaka:
On other platforms, I haven't seen a visible performance loss caused by bitrates at 1080p resolution. On the contrary, 4k 50M+ may run into this problem.

If we put the encoder aside. How many FPS can you get considering just the decoder and filters?

Decoder
```shell
ffmpeg -hwaccel videotoolbox -hwaccel_output_format videotoolbox_vld -i INPUT -an -sn -f null -
```

Decoder + filter
```shell
ffmpeg -hwaccel videotoolbox -hwaccel_output_format videotoolbox_vld -i INPUT -vf \
hwmap=derive_device=opencl:mode=read,transpose_opencl,hwmap=derive_device=videotoolbox:mode=write,scale_vt \
-an -sn -f null -
```

gnattu:
> If we put the encoder aside. How many FPS can you get considering just the decoder and filters?

This actually makes it run slower than putting an encoder after it, by a large percentage (~30%) slower than having a `hevc_videotoolbox` sink.

Looks like there are many magics inside the VideoToolbox that cannot be evaluated normally, or it could be some magic inside ffmpeg.

gnattu:
Updated to support all hardware filters of VideoToolbox. Now all main features: de-interlace,  scale, HDR->SDR tone mapping(including DOVI Profile 5), HDR pass-though and subtitles burn in should be supported by hardware.

nyanmisaka:
https://github.com/nyanmisaka/jellyfin/commit/d031f7daacfbc9499a7641efe5fbad5e8e6135e1

I simplified some redundant parts. And the order of the deint filter has been adjusted, because it must be chained after the decoder, not after the scale filter, otherwise the image quality will be degraded. Finally, I disabled hwSurface only for cases using VideoToolBox tone-mapping. Because for common transcoding use cases, it is not cost-effective to exchange higher system power consumption (memory copy from decoder buffer) and more heat for a small amount of extra FPS, especially for battery-powered notebooks.

gnattu:
> [nyanmisaka@d031f7d](https://github.com/nyanmisaka/jellyfin/commit/d031f7daacfbc9499a7641efe5fbad5e8e6135e1)
> 
> I simplified some redundant parts. And the order of the deint filter has been adjusted, because it must be chained after the decoder, not after the scale filter, otherwise the image quality will be degraded. Finally, I disabled hwSurface only for cases using VideoToolBox tone-mapping. Because for common transcoding use cases, it is not cost-effective to exchange higher system power consumption (2x memory copy: decoder buffer->copy to sw buffer->copy to hw buffer) and more heat for a small amount of extra FPS, especially for battery-powered notebooks.

No, you should not use hw surface, as it not only breaks tone mapping, but it also breaks HDR pass-through. The wrong colors will appear everywhere. It also cannot guarantee zero copy on VideoToolbox, as the hardware decoder, GPU, and hardware encoder could be located at different locations off the bus (not always in the same chip like other platforms). For Intel Macs, the hardware encoders/decoders could be the Intel-built-in, or they could be the T2 chip's built-in on some models. However, the filter context would be created on the amdgpu if available. In this case, hw surface not only makes it slower, but it also does not eliminate the copy either. If we have to use it, we should only use it for 8bit videos and take closer look at the performance and power consumption to see if it produces any measurable benefits.

nyanmisaka:
> but it also breaks HDR pass-through.

Actually we haven't implemented HDR passthrough yet because it requires HEVC Main10 instead of Main. Unless you want 8-bit HDR.

> It also cannot guarantee zero copy on VideoToolbox, as the hardware decoder, GPU, and hardware encoder could be located at different locations off the bus (not always in the same chip like other platforms). For Intel Macs, the hardware encoders/decoders could be the Intel-built-in, or they could be the T2 chip's built-in on some models. However, the filter context would be created on the amdgpu if available. In this case, hw surface not only makes it slower, but it also does not eliminate the copy either. If we have to use it, we should only use it for 8bit videos and take closer look at the performance and power consumption to see if it produces any measurable benefits.

Well, I'm convinced. hwSurface is now only enabled for OpenCL tone-mapping. https://github.com/nyanmisaka/jellyfin/commit/5d89a86d76a2499f08d23446fe2001ce9fb14b2f

gnattu:
Tested a few files and it appears to be OK. I merged most of your suggested code into my branch, only made some modifications:

- Manually check OpenCL Tone Mapping to avoid stack overflow
- Only do color override for OpenCL Tone Mapping because it is not needed when hw surface is not in use
- Return empty filter chain when nothing has to be done

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -4954,56 +4978,146 @@ public string GetHwTonemapFilter(EncodingOptions options, string hwTonemapSuffix
...
-            // ffmpeg cannot use videotoolbox to scale
-            var swScaleFilter = GetSwScaleFilter(state, options, vidEncoder, inW, inH, threeDFormat, reqW, reqH, reqMaxW, reqMaxH);
-            newfilters.Add(swScaleFilter);
+            /* Make main filters for video stream */
+            var mainFilters = new List<string>();
+
+            if (!(doTonemap || hasSubs || doDeintH2645))
+            {
+                // Dummy action to return empty filters when nothing to do.
+                return (mainFilters, mainFilters, mainFilters);
```
nyanmisaka, line 5043: You cannot return here or you will miss video scaling.

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -4954,56 +4978,147 @@ public string GetHwTonemapFilter(EncodingOptions options, string hwTonemapSuffix
...
+                return (null, null, null);
             }
 
-            // ffmpeg cannot use videotoolbox to scale
-            var swScaleFilter = GetSwScaleFilter(state, options, vidEncoder, inW, inH, threeDFormat, reqW, reqH, reqMaxW, reqMaxH);
-            newfilters.Add(swScaleFilter);
+            /* Make main filters for video stream */
+            var mainFilters = new List<string>();
+
+            if (!(doTonemap || doScale || hasSubs || doDeintH2645))
```
nyanmisaka, line 5042: I don't get it, why would you check these prematurely? Subsequent code has checked them and will eliminate empty filters.

When we enable more formats for the VT decoder in the future, or when upstream adds new ones to the format list of the VT encoder, the sw formats in this code path will not be subject to these known-working formats and may throw runtime error.

I think what we want is some concrete format rather than letting ffmpeg negotiate it.

```cs
            mainFilters.Add("format=nv12|p010le|videotoolbox_vld");
            mainFilters.Add("hwupload=derive_device=videotoolbox");
```

gnattu: This is because I've seen empty hwupload causing out-of-order frames during my early testing, but I'm not seeing them again since we stopped using hw surface. So, yes, perhaps we don't need to check this prematurely for now.

gnattu:
What we need to do is modify the upstream `scale_vt` filter to make it output NV12 frames, since we are not doing P010 encoding in any case for now. Currently, the `scale_vt` filter does not modify the pixel format, and all 10-bit inputs will have a 10-bit output, even after tone-mapping. This causes our encoder to complain about it (10-bit input using main profile), leading to frame errors for some videos. It will look like the same problem when we are feeding BGRA frames directly into it.

gnattu:
A more severe problem: it seems like that we cannot even use hw surface for OpenCL tone mapping. The `videotoolbox_vld` is hardcoded to be an 8bit input which means we cannot chain it after the `scale_vt` filter and OpenCL filter has to be the very first filter to get a 10bit input, which forces us to do tone mapping before scaling, and that would be very slow for 4K videos.

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -4954,56 +4978,145 @@ public string GetHwTonemapFilter(EncodingOptions options, string hwTonemapSuffix
...
+            var isVtEncoder = vidEncoder.Contains("videotoolbox", StringComparison.OrdinalIgnoreCase);
+
+            var doDeintH264 = state.DeInterlace("h264", true) || state.DeInterlace("avc", true);
+            var doDeintHevc = state.DeInterlace("h265", true) || state.DeInterlace("hevc", true);
+            var doDeintH2645 = doDeintH264 || doDeintHevc;
+            var doVtTonemap = IsVideoToolboxTonemapAvailable(state, options);
+            var doOclTonemap = !doVtTonemap && IsHwTonemapAvailable(state, options);
+
+            var scaleFormat = string.Empty;
+            if (GetVideoColorBitDepth(state) == 10)
```
nyanmisaka, line 5022: Does VtEncoder support the sub-sampling of input format and encode profile do not match?

Such as `nv16` Rext(4:2:2) -> `nv12` Main(4:2:0)

gnattu: To an extent. It does not straight error out, but the output frame may contain some artifacts.

gnattu: Do we have any convenient way to detect the input pixel format in the server for now? I don't want to force a pixel conversion for every inputs just for those very rare 422 videos.

nyanmisaka (approved):

Shadowghost (approved):

crobibero (approved):

PR #11037, opened 2024-02-20
Add ability to enable/disable libraries

btopherjohnson:
This change addresses users desires to have the option to disable particular libraries for all users at certain times. This is handy for seasonal libraries (like christmas movies/music/tv shows).

I added a "Disabled" option to LibraryOptions.cs and a check to see if a library is disabled in CollectionFolder.cs (it overrides Folder's IsVisible function).

I chose "disabled" over "enabled" because the default value for booleans loaded into the LibraryOptions class by the JsonDeserializer is false, so by default, existing libraries will remain enabled.

I also made a corresponding checkbox in the jellyfin-web repository.

This commit addresses the issue described here:
https://features.jellyfin.org/posts/2530/hide-a-library-from-all-users-disable-library

MediaBrowser.Controller/Entities/CollectionFolder.cs - Added override to IsVisible to check if the library is disabled
MediaBrowser.Model/Configuration/LibraryOptions.cs - Added "Disabled" as an option for libraries.

### DIFF
--- a/MediaBrowser.Model/Configuration/LibraryOptions.cs
+++ b/MediaBrowser.Model/Configuration/LibraryOptions.cs
@@ -25,6 +25,8 @@ namespace MediaBrowser.Model.Configuration
             SeasonZeroDisplayName = "Specials";
         }
 
+        public bool Disabled { get; set; }
+
         public bool EnablePhotos { get; set; }
 
         public bool EnableRealtimeMonitor { get; set; }
--- a/MediaBrowser.Model/Configuration/LibraryOptions.cs
+++ b/MediaBrowser.Model/Configuration/LibraryOptions.cs
@@ -25,7 +25,7 @@ namespace MediaBrowser.Model.Configuration
             SeasonZeroDisplayName = "Specials";
         }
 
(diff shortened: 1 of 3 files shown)


btopherjohnson:
I opened a corresponding PR in jellyfin-web:
[PR 5215](https://github.com/jellyfin/jellyfin-web/pull/5215)

JPVenson (approved):

MediaBrowser.Model/Configuration/LibraryOptions.cs
```diff
@@ -25,6 +25,8 @@ public LibraryOptions()
             SeasonZeroDisplayName = "Specials";
         }
 
+        public bool Disabled { get; set; }
```
cvium, line 28: You can just assign a default value.

```suggestion
        public bool Enabled { get; set; } = true;
```

Shadowghost:
Did you check if this properly enforced ACLs on the items below the library?

cvium (approved):

Bond-009 (approved):

PR #11067, opened 2024-02-26
Add helpers to register and get a plugin HttpClient

crobibero:
Usage:

Register:
```c#
public void RegisterServices(IServiceCollection serviceCollection, IServerApplicationHost applicationHost)
{
    applicationHost.AddPluginHttpClient<PluginType>(serviceCollection);
}
```

Get:
```c#
var httpClient = httpClientFactory.GetPluginHttpClient<PluginType>();
```

where `PluginType : BasePlugin`

### DIFF
--- a/Emby.Server.Implementations/ApplicationHost.cs
+++ b/Emby.Server.Implementations/ApplicationHost.cs
@@ -290,7 +290,7 @@ namespace Emby.Server.Implementations
         public void AddPluginHttpClient<T>(
             IServiceCollection serviceCollection,
             string basedOnClientName,
-            (string Name, string[] Values)[] customHeaders = null)
+            (string Name, string[] Values)[] customHeaders)
             where T : BasePlugin
         {
             ArgumentNullException.ThrowIfNull(serviceCollection);
--- a/Emby.Server.Implementations/ApplicationHost.cs
+++ b/Emby.Server.Implementations/ApplicationHost.cs
@@ -10,8 +10,12 @@ using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Net;
+using System.Net.Http;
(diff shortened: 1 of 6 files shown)


Shadowghost (approved):

barronpm (commented):
I'm not a big fan of adding more to ApplicationHost, IMO these are a good candidate for extension methods for `IServiceCollection`

cvium:
What is this trying to solve?


fre. 1. mar. 2024 kl. 15.55 skrev Patrick Barron ***@***.***>:

crobibero:
> What is this trying to solve?

I wanted to standardize how plugins configure HttpClients so they get the correct request headers without needing to implement it in every plugin

Evey plugin could just use `NamedClient.Default` but I would like to also add the plugin name added to the UserAgent

JPVenson:
Well now with Dotnet8 why not go the extra route and actually use named DI http clients as well as the new IHttpClientFactory
https://learn.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests

crobibero:
> Well now with Dotnet8 why not go the extra Kilometer and actually use named DI http clients as well as the new IHttpClientFactory [learn.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests](https://learn.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests)

I'm not sure what you're referring to- but we already do use IHttpClientFactory and named clients

JPVenson:
closed do to inactivity

PR #11071, opened 2024-02-27
Move userId in API from route to optional query parameter

nielsvanvelzen:
**This is a proof of concept, looking for feedback to see if this is the right way to fix this problem or if there are better methods.**

Right now it changes 2 operations: GetUserViews and GetGroupingOptions, both part of UserViewsController. There are 37 more operations that will need a similar change. I've tested this with the unstable version of jellyfin-web and the old endpoints still work so this is **not** a breaking API change.

**Changes**

Move userId in API from route to optional query parameter by adding new endpoints (with new routes!!) that move the userId to a query parameter. The old endpoints are kept but made obsolete and hidden from the OpenAPI specification.

With this change it is no longer required to provide a user id in the API, together with #11024 and #11028. This makes is a lot easier to use our API/SDK's and develop clients, as you no longer need to keep track of the current user id to use the API.

**Issues**

### DIFF
--- a/Jellyfin.Api/Controllers/UserViewsController.cs
+++ b/Jellyfin.Api/Controllers/UserViewsController.cs
@@ -4,6 +4,7 @@ using System.ComponentModel.DataAnnotations;
 using System.Globalization;
 using System.Linq;
 using Jellyfin.Api.Extensions;
+using Jellyfin.Api.Helpers;
 using Jellyfin.Api.ModelBinders;
 using Jellyfin.Api.Models.UserViewDtos;
 using Jellyfin.Data.Enums;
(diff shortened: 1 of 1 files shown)


Jellyfin.Api/Controllers/UserViewsController.cs
```diff
@@ -92,14 +91,33 @@ public class UserViewsController : BaseJellyfinApiController
...
+    /// <returns>An <see cref="OkResult"/> containing the user views.</returns>
+    [HttpGet("Users/{userId}/Views")]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    [Obsolete("Kept for backwards compatibility")]
+    [ApiExplorerSettings(IgnoreApi = true)]
+    public QueryResult<BaseItemDto> GetUserViews2(
+        [FromRoute, Required] Guid userId,
+        [FromQuery] bool? includeExternalContent,
+        [FromQuery, ModelBinder(typeof(CommaDelimitedArrayModelBinder))] CollectionType?[] presetViews,
+        [FromQuery] bool includeHidden = false) => GetUserViews(userId, includeExternalContent, presetViews, includeHidden);
```
crobibero, line 119:
```suggestion
        [FromQuery] bool includeHidden = false)
        => GetUserViews(userId, includeExternalContent, presetViews, includeHidden);
```

The only suggestion I have is to place the expression body on the next line- when I first read through this I thought it was magic

nielsvanvelzen: The server is full of crazy linters that complain about all my formatting, so I auto-formatted using Rider.

nielsvanvelzen: Like really, I don't want to start another tabs vs spaces like war but this is the only sane way to format it and somehow we decided to not do this.

91700002d0

crobibero: don't bring your kotlin into this :1st_place_medal:

PR #11072, opened 2024-02-27
Add SubtitleCodecPreference and improve smart subtitles

JeWe37:
This PR offers some simple improvements for the automatic subtitle pre-selection in Jellyfin.

Please note: I am not familiar with the migration system, and mostly simply copied it from the previous migration of the Users table, however it does seem to work.

jellyfin-web PR: jellyfin/jellyfin-web#5235

**Changes**
- Codecs cannot be ranked uniquely by quality, as compatibility and styling quality are both factors, therefore the preferred codec is made configurable
- The smart subtitle mode previously did not handle the forced flag very "smartly", this has been improved

**Issues**


Emby.Server.Implementations/Library/MediaStreamSelector.cs
```diff
@@ -60,13 +62,19 @@ public static class MediaStreamSelector
...
+                        .ThenByDescending(x => !x.IsForced)
+                        .ThenByDescending(x => preferredCodecs.Contains(x.Codec, StringComparison.OrdinalIgnoreCase))
+                        .FirstOrDefault();
                 }
                 else
                 {
-                    // Respect forced flag.
-                    stream = sortedStreams.FirstOrDefault(x => x.IsForced);
+                    // Respect forced flag, but make sure to match the language
+                    stream = sortedStreams.FirstOrDefault(x => x.IsForced && string.Equals(x.Language, audioTrackLanguage, StringComparison.OrdinalIgnoreCase));
```
TimGels, line 77: Just to make sure, this means Forced will only be considered when the audio track language is the same, correct?

JeWe37: Yes, I would say generally you would only want forced subtitles in the same language as the audio, if that language is your preferred subtitle language.

nielsvanvelzen:
What is the benefit of this change? Why should a user want to prefer a certain subtitle codec?

JeWe37:
The two most common subtitles are SRT and ASS. The latter has far better styling support, but compatibility with many players is rather poor, so some may choose to prefer SRT(particularly since even when ASS works, it may not be reliable or performant), depending on their clients. 

The alternative would be to impose some fixed codec ranking. One could conceive of having an option for preferring the highest compatibility or highest quality subtitle codecs. If that is preferred I'd be happy to implement that instead.

nielsvanvelzen:
Feels like a hack to me, in most cases having different formats of subtitles for a file does not mean they are the same subtitles. If a user chooses to prefer SRT they might get German subtitles when their preferred language is Dutch but those subtitles used the SSA format.

JeWe37:
That's accounted for by the codec having a low priority in the sorting. The subtitles must first be the same in terms of language preference, disposition etc. before the codec will be considered. Do you have any suggestions for how else to go about this?

nielsvanvelzen:
Personally I think it's a niche use-case, I can't imagine a lot of people having media with multiple subtitles for the same language. That's just my opinion though, others might think different about it.

EraYaN:
> Personally I think it's a niche use-case, I can't imagine a lot of people having media with multiple subtitles for the same language. That's just my opinion though, others might think different about it.

This actually happens a ton when ripping blurays, all the forced, SDH and random extra subtitle tracks are a menace. Not sure if this fixes it but if it sorts the Forced one underneath the normal one it would help a ton. And then you add some external sub and it gets even wilder.

TimGels:
Quite common for Anime to have both PGS and SSA/ ASS tracks with the same language (full/ dialogue). I do like the Force subtitle change, as it aligns with the Matroska spec to my knowledge.

Emby.Server.Implementations/Library/MediaSourceManager.cs
```diff
@@ -388,6 +388,18 @@ private void SetDefaultSubtitleStreamIndex(MediaSourceInfo source, UserItemData
             }
 
             var preferredSubs = NormalizeLanguage(user.SubtitleLanguagePreference);
+            var preferredCodec = user.SubtitleCodecPreference;
+            var preferredCodecs = preferredCodec switch
+            {
```
TimGels, line 393: Maybe add an option for PGS (HDMV PGS) as well? Quite a common format.

Emby.Server.Implementations/Library/MediaStreamSelector.cs
TimGels: I agree. To my knowledge that is also in line with the Matroska spec.

Emby.Server.Implementations/Library/MediaSourceManager.cs
JeWe37: Good point. None of the image based formats are handled. Jellyfin only had constants defined for text based formats. I'll look over the codebase and check if there's any other places image based subs are mentioned.

JeWe37:
Alright, to move this discussion forward let's summarize the options here:

Firstly, the  change to the forced subtitle ordering for smart subtitles seems pretty uncontroversial it seems. It also is rather non-invasive so I think that doesn't really present an issue.

As for the matter of the subtitle codec preference, there are broadly two options here that I think would work to achieve what I want:
- Keep the setting as is, meaning having one specific subtitle codec that if it exists will be preferred to all others.
  - Pros:
    - Very flexible, people can choose what they prefer themselves
  - Cons:
    - Rather technical, perhaps not terribly user friendly if you don't know about subtitle codecs
    - Does not allow ranking the subtitle codecs. So say you prefer ASS, but you have a file with SRT and PGS, you couldn't now choose to prefer PGS
- Change to choosing a preference for either highest quality or best supported subtitles
  - Pros:
    - More user friendly
    - Can rank all subtitles
  - Cons
    - Less flexible, hypothetically there could exist clients with which this wouldn't work well(in practice knowing most of the Jellyfin clients this isn't a real issue)
I do believe at least providing this option is valuable, see the examples given in the above comments. It is perhaps somewhat niche, but I think that is fine. After all this is an option that nobody has to use if they don't need it.

JeWe37:
With 10.9 now having been out for a while I thought I'd give this a rebase and see if this can't be completed. I'd love to get some input on my last comment @nielsvanvelzen so that I can incorporate any desired changes.

JPVenson:
closed due to inactivity

PR #11074, opened 2024-02-27
Move userId in API from route to optional query parameter

crobibero:
Superceeds https://github.com/jellyfin/jellyfin/pull/11071

I left the route parameter in `UserController` and `SessionController` since they are only really used for manipulating other users

### DIFF
--- a/Jellyfin.Api/Controllers/ImageController.cs
+++ b/Jellyfin.Api/Controllers/ImageController.cs
@@ -90,7 +90,7 @@ public class ImageController : BaseJellyfinApiController
     /// <response code="204">Image updated.</response>
     /// <response code="403">User does not have permission to delete the image.</response>
     /// <returns>A <see cref="NoContentResult"/>.</returns>
-    [HttpPost("Users/Images")]
+    [HttpPost("UserImage")]
     [Authorize]
     [AcceptsImageFile]
     [ProducesResponseType(StatusCodes.Status204NoContent)]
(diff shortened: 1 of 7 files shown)


nielsvanvelzen:
Overall a great change. I think it would make more sense to remove the `/Users/` prefix from the paths, like I did with the UserViews (e.g. /Users/{id}/Views -> /UserViews). So a few examples would be:
- /Users/{id}/Views -> /UserViews
- /Users/{id}/Suggestions -> /Suggestions (Or perhaps /Items/Suggestions)
- /Users/{userId}/PlayedItems/{itemId} -> /PlayedItems/{itemId}
etc.

You've also missed the UserController. Although most of these are for admins and are likely to be used to manipulate other users. Some could definitely be changed, like "UpdateUserConfiguration" and "UpdateUserPassword".

[Kotlin SDK diff](https://github.com/nielsvanvelzen/jellyfin-sdk-kotlin/commit/410db5e21e417646622fc9eb082af586d4b53ab6)

nielsvanvelzen (approved):
✅ API changes

[Kotlin SDK diff](https://github.com/jellyfin/jellyfin-sdk-kotlin/compare/a7fc77d141119479591950bd2bb666fd6d783075...nielsvanvelzen:jellyfin-sdk-kotlin:tst-no-userid)

Shadowghost (approved):

Bond-009 (approved):

PR #11075, opened 2024-02-28
Fix Grouped Alternate Version Behavior

### LINKED ISSUE #9281: [Issue]: Grouped version count is not changed when a version is deleted from the file system
### Please describe your bug

After grouping one or more versions, the group count is displayed on the poster that is displayed for that group. As well, Media Info shows all the versions in that group, and it's possible to select which version to play.

Subsequently deleting one or more versions from the file system and rescanning the library, the deleted versions are correctly removed from Media Info and version selection from Play. However, the version count that is displayed on the poster is not decremented to account for the deleted version(s).

Example/steps to reproduce:

[**_NOTE_**: this applies to multiple versions stored in their own individual folders, rather than multiple versions stored within the same folder]

1. Group two items/versions of the same movie, verifying that the poster count = 2
2. In the file system, delete one of the items that had been grouped in Jellyfin
3. Rescan the library containing the original group for changed/updated media
4. The poster count still = 2, even though Media Info now shows only one item
5. The Split function is no longer available, so there is no way to change the count shown on the poster

### Jellyfin Version

10.8.0

### if other:

10.8.9

### Environment

```markdown
- OS: Windows 11
- Virtualization: None
- Clients: Browser
- Browser: MS Edge
- FFmpeg Version: n/a
- Playback Method: n/a
- Hardware Acceleration: n/a
- Plugins: None
- Reverse Proxy: n/a
- Base URL: n/a
- Networking: n/a
- Storage: n/a
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

one-new-message:
**Changes**
Implements a basic check in `LibraryManager` to remove an item from containing collections before deletion.

**Issues**
Intended to fix #9281 and its related unexpected behavior.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -252,3 +252,4 @@
  - [JPUC1143](https://github.com/Jpuc1143/)
  - [0x25CBFC4F](https://github.com/0x25CBFC4F)
  - [Robert Lützner](https://github.com/rluetzner)
+ - [melanie gemini emory](https://github.com/one-new-message)
--- a/Emby.Server.Implementations/Library/LibraryManager.cs
+++ b/Emby.Server.Implementations/Library/LibraryManager.cs
@@ -374,6 +374,17 @@ namespace Emby.Server.Implementations.Library
 
                 try
                 {
+                    if (item.HasLocalAlternateVersions)
+                    {
+                        if (parent.GetBaseItemKind() is BaseItemKind.BoxSet)
+                        {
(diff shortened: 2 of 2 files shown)


one-new-message:
finally managed to get a working test environment, not noticing any odd behavior with duplicate/multiple versions and collections. will check if this behavior is present in 10.8.13 to compare.

one-new-message:
this PR does not fix this issue. closing and will take another look.

Issue #8835, opened 2022-12-01
Schedules direct artwork not refreshing automatically.  Refreshing guide manually works.


junobomb:
@rudism Although fixed in 10.8.6, now it requires refreshing Guide Data for images to appear.  After a while images disappear and do not refresh automatically.  Refreshing guide manually works.

_Originally posted by @junobomb in https://github.com/jellyfin/jellyfin/issues/7528#issuecomment-1333859271_
    [Issue]: Schedules direct artwork not loading #7528

aniel300:
is there an eta for when this will be fixed?

frabe8378910:
I still still do not get any images in Jellyfin with guide data from Schedules Direct, no matter if I run a refresh of the guide data manually or not. Is there any other way to force an image reload? 
Running on Debian Unstable and Jellyfin 20230209.8.

In the log there tons of this messages:

[2023-02-09 13:19:31.211 +01:00] [ERR] Error processing request. URL "GET" "/Items/3abe69fac82c545eddb28990d17bbcbc/Images/Primary".
System.InvalidOperationException: Operation is not valid due to the current state of the object.
   at Emby.Server.Implementations.Library.LibraryManager.ConvertImageToLocal(BaseItem item, ItemImageInfo image, Int32 imageIndex)
   at Jellyfin.Api.Controllers.ImageController.GetImageInternal(Guid itemId, ImageType imageType, Nullable`1 imageIndex, String tag, Nullable`1 format, Nullable`1 maxWidth, Nullable`1 maxHeight, Nullable`1 percentPlayed, Nullable`1 unplayedCount, Nullable`1 width, Nullable`1 height, Nullable`1 quality, Nullable`1 fillWidth, Nullable`1 fillHeight, Nullable`1 blur, String backgroundColor, String foregroundLayer, BaseItem item, ItemImageInfo imageInfo)
   at Jellyfin.Api.Controllers.ImageController.GetItemImage(Guid itemId, ImageType imageType, Nullable`1 maxWidth, Nullable`1 maxHeight, Nullable`1 width, Nullable`1 height, Nullable`1 quality, Nullable`1 fillWidth, Nullable`1 fillHeight, String tag, Nullable`1 cropWhitespace, Nullable`1 format, Nullable`1 percentPlayed, Nullable`1 unplayedCount, Nullable`1 blur, String backgroundColor, String foregroundLayer, Nullable`1 imageIndex)
   at lambda_method781(Closure, Object)
[...]

PR #11083, opened 2024-03-01
Precache livetv program images

crobibero:
Should fix https://github.com/jellyfin/jellyfin/issues/8835

### DIFF
--- a/MediaBrowser.Providers/Manager/ProviderManager.cs
+++ b/MediaBrowser.Providers/Manager/ProviderManager.cs
@@ -9,6 +9,7 @@ using System.Net.Http;
 using System.Net.Mime;
 using System.Threading;
 using System.Threading.Tasks;
+using AsyncKeyedLock;
 using Jellyfin.Data.Enums;
 using Jellyfin.Data.Events;
 using Jellyfin.Extensions;
@@ -30,6 +31,7 @@ using MediaBrowser.Model.Extensions;
 using MediaBrowser.Model.IO;
 using MediaBrowser.Model.Net;
 using MediaBrowser.Model.Providers;
+using Microsoft.Extensions.Caching.Memory;
 using Microsoft.Extensions.Logging;
 using Book = MediaBrowser.Controller.Entities.Book;
 using Episode = MediaBrowser.Controller.Entities.TV.Episode;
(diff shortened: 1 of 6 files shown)


Shadowghost (approved):

MediaBrowser.Providers/Manager/ProviderManager.cs
```diff
@@ -150,52 +162,86 @@ public Task<ItemUpdateType> RefreshSingleItem(BaseItem item, MetadataRefreshOpti
...
+                    {
+                        await SaveImage(
+                            item,
+                            cacheStream,
+                            cachedValue.Value.ContentType,
+                            type,
+                            imageIndex,
+                            cancellationToken).ConfigureAwait(false);
+                    }
+                }
```
Bond-009, line 182: Shouldn't we do an early return here if we find it in the cache?

crobibero: I swear I had the return there..

MediaBrowser.Providers/Manager/ProviderManager.cs
```diff
@@ -150,52 +162,87 @@ public Task<ItemUpdateType> RefreshSingleItem(BaseItem item, MetadataRefreshOpti
...
-            var contentType = response.Content.Headers.ContentType?.MediaType;
+                var httpClient = _httpClientFactory.CreateClient(NamedClient.Default);
+                using var response = await httpClient.GetAsync(url, cancellationToken).ConfigureAwait(false);
 
-            // Workaround for tvheadend channel icons
-            // TODO: Isolate this hack into the tvh plugin
-            if (string.IsNullOrEmpty(contentType))
-            {
-                if (url.Contains("/imagecache/", StringComparison.OrdinalIgnoreCase))
+                if (response.StatusCode != HttpStatusCode.OK)
```
JPVenson, line 188:
```suggestion
                response.EnsureSuccessStatusCode()
```

MediaBrowser.Providers/Manager/ProviderManager.cs
```diff
@@ -150,52 +162,84 @@ public Task<ItemUpdateType> RefreshSingleItem(BaseItem item, MetadataRefreshOpti
...
-                    cancellationToken).ConfigureAwait(false);
+                var stream = await response.Content.ReadAsStreamAsync(cancellationToken).ConfigureAwait(false);
+                await using (stream.ConfigureAwait(false))
+                {
+                    var memoryStream = new MemoryStream();
+                    await using (memoryStream.ConfigureAwait(false))
+                    {
+                        await stream.CopyToAsync(memoryStream, cancellationToken).ConfigureAwait(false);
+                        memoryStream.Seek(0, SeekOrigin.Begin);
+                        _memoryCache.Set(url, (contentType, memoryStream.GetBuffer()), TimeSpan.FromSeconds(10));
```
Bond-009, line 232: `MemoryStream.GetBuffer()` returns the underlying buffer, it's length may not match the actual content
I'd use `HttpContent.ReadAsByteArrayAsync` and construct a `MemoryStream` from that for use by `SaveImage`

Bond-009 (changes requested):

Bond-009 (approved):

JPVenson (approved):

Issue #5501, opened 2021-03-12
Incorrect GIF image display


kwaiksa:
**Describe the bug**
I use GIF image as thumb in some video.
After upgraded 10.7.0, GIF display incorrect.
And Image format showed in browser developer tools is image/png not image/gif.

**System (please complete the following information):**
 - OS: Windows 10
 - Virtualization: No
 - Clients: Browser
 - Browser: Chrome 89
 - Jellyfin Version: 10.7.0 release

**Screenshots**
![Honeycam 2021-03-12 14-00-25](https://user-images.githubusercontent.com/9345919/110909905-9aede580-834b-11eb-829b-9745cec42d39.gif)
10.7.0

![Honeycam 2021-03-12 14-12-14](https://user-images.githubusercontent.com/9345919/110909967-afca7900-834b-11eb-90a7-b17f8b486ada.gif)
10.6.4

boyxiaopang:
and so, it is a bug?

kwaiksa:
10.7.6 release still exist

kwaiksa:
Still in v10.6.4 due to this issue.
But i need new feature in v10.7.x, thought i should give up my GIF thumbnail now... :(

crobibero:
I just tested 10.8-alpha1 and don't have an issue with a gif as the primary image or as the thumb

agarnele:
i use the skin aeon mod silvo in kodi.... she will play gif aniations (extat in the standard skin too)

VinceGusmini:
[Report on Jellyfin (Windows) Version 10.7.7]
Client : Chrome Version 96.0.4664.110

GIF don't work anywhere for me but I tried to swap my GIF to APNG and it's only working in the Admin Dashboard of libraries.
Record :
![chrome](https://user-images.githubusercontent.com/29730840/147968063-8755c990-0f4e-4637-9e03-654ce1ea9676.gif)

Shirkit:
@crobibero any chance it gets extended to support animated WebP images asl well? They are smaller in size, it's the preferred method for storing lots of animated images.

felix920506:
@Shirkit Hi, please open a new issue for it as WebP is a different format than GIF

PR #11109, opened 2024-03-05
Don't decode animated images

crobibero:
Apparently animated WebP images are a thing
https://github.com/jellyfin/jellyfin/issues/5501#issuecomment-1977428971

This change makes it so an an image has any frames (is animated) we won't attempt to convert it from the original format


IDisposable (approved):

Shadowghost (approved):

cvium (approved):

Bond-009 (approved):

thornbill:
Could you add some context in the PR description for reference? 💜

crobibero:
> Could you add some context in the PR description for reference? 💜

Added a little bit of flavor the PR body as I should have way back when

PR #11120, opened 2024-03-08
IIS Hosting

kimboslice99:
As mentioned in #11116 this will allow for hosting InProcess on IIS

I would be happy to write some documentation for it as well but im unsure where such information should even go, maybe Installation->IIS?


JPVenson:
Hey @kimboslice99 
I am not sure if we _want_ IIS to be supported. This change would be better discussed beforehand in our meta repro.

kimboslice99:
> not sure if we _want_ IIS to be supported

Why _not_? 

.Net apps that run on kestrel can be run on IIS, this doesnt require much support. It just works.

Been running this way for several days now. Works well. I did open a discussion where I was then told to make a PR. I am not sure what you mean by "meta repro"?

Care to actually mention your objections with this? What possible downside is there to this small change?

JPVenson:
Well from one point, IIS is a pain to deal with compared to the kestrel webserver. I dont see any advantages allowing that apart from adding another feature, noone of the team _wants_ to deal with nor will use it which will lead to issues with it that no one from the team wants to work on and we get in the same situation we currently have with SyncPlay, The SSL feature and a few other parts. Speaking of which, we do support a few features like the explicit adapter binding and the SSL feature i am not certain will work under the IIS.

kimboslice99:
>  explicit adapter binding and the SSL feature i am not certain will work under the IIS.

Adapter binding and ssl is handled by IIS, of course. I dont believe those settings have any effect when behind IIS.

> noone of the team wants to deal with nor will use it which will lead to issues with it

There isnt anything specific about running on IIS that is really different from kestrel, to the application.

> IIS is a pain to deal with compared to the kestrel webserver.

Thats more a personally preference I'd say. IIS has a fantastic interface and allows for so much more to be done in the pipeline, and has _many_ security features


The entire point of the ANCM is that a .net app can run on IIS with a simple configuration option

> The ASP.NET Core Module (ANCM) is a native IIS module that plugs into the IIS pipeline, allowing ASP.NET Core applications to work with IIS

JPVenson:
Yes so that means, we would create support for a feature (as its in the code) that implicate other features, nobody from the team runs and nobody even if its a personal preference, wants to work with from the team. If this code gets ins, it means we have to support it, which there is no desire from the team members i talked about the issue.

This is by no means final. I just spoke with a small number from the team and others might chime in and give thier thoughts on the topic but from my perspective, its a feature that we dont want to support, that breaks existing other features and has a very limited audience for the price of impossible support.

kimboslice99:
I see your points and appreciate the input. I just dont understand the "support" issue, there is nothing to support, simply note in docs that ssl and port binding is handled out of iis when configured this way. To the rest of the application, it cant tell the difference, so I dont see how any issues could arise. the ancm handles this all very well.

Otherwise, I will maintain this fork 🤷 I just thought others could benefit

JPVenson:
For context, i am a member of the Triage team for jellyfin and a moderator in the chats. Together with others i try to resolve issues that users might have with jellyfin. 

If we were to put that code into jellyfin, we would tell anyone that we officially support running jellyfin within an IIS server. And there is the issue, were would be in a situation where the code would support that feature, but we as a team have no desire to work with IIS or debug its quirks when they come up. The existing infastructure, docker or barebones works well enough and has enough issues we need to be aware of so adding that would make anything just way more difficult. Also it just does not stop at those two features, someone needs to figure out what exactly the implications if running JF in a IIS would be and that everything still works as expected. Formost permission issues i expect to be a nightmare to debug. 

I understand that this works great for you, but we have to ensure that it works great for everyone else and this implicates more then just 3 lines of code to be checked.

kimboslice99:
Permissions are no different than running JF in any other way, just that now the user account (if using app pool identity) would be IIS AppPool\sitename. 

Debugging issues on IIS is very easy, I find (FREB, http.sys err log, etc). 

I would be happy to provide any information as I'm pretty well versed in the configuration and setup of IIS. 

ffmpeg hardware transcoding works, for one, so this is ideal compared to a service install on windows. 

I can understand your point, that perhaps nobody on the Triage team is experienced with IIS, however I am willing to provide support on this (I have quite a bit of free time to devote).

Xhanti:
The point is you are familiar with IIS and its quirks, they are not and don't care to become familiar with it. That's okay, if you want to fork and maintain it that's fine. If they have no interest in maintaining it, there's no need to force the issue. And IIS support is not free, teaching people about the quirks of IIS (configuring users with the right access to network shares comes to mind, configuring app pools to not restart etc) isn't required. I've worked with IIS for 10 years and it's not something this project needs, kestrel is plenty secure and edge hardened and uses a fraction of the resources.

kimboslice99:
@Xhanti I dont need you to tell me that I can fork it, dont understand your input there. As already stated, there is nothing to really maintain, this is a hilariously simple modification that has zero effect on any other system, just adding the possibility of running on IIS. 

If a few clicks of a mouse is too much to disable app pool recycling then I dont know what to tell you.

Jellyfin.Server/Program.cs
```diff
@@ -139,7 +140,14 @@ private static async Task StartServer(IServerApplicationPaths appPaths, StartupO
                 host = Host.CreateDefaultBuilder()
                     .UseConsoleLifetime()
                     .ConfigureServices(services => appHost.Init(services))
-                    .ConfigureWebHostDefaults(webHostBuilder => webHostBuilder.ConfigureWebHostBuilder(appHost, startupConfig, appPaths, _logger))
+                    .ConfigureWebHostDefaults(webHostBuilder =>
+                    {
+                        webHostBuilder.ConfigureWebHostBuilder(appHost, startupConfig, appPaths, _logger);
+                        if (Environment.GetEnvironmentVariable("ASPNETCORE_IIS_PHYSICAL_PATH") is not null)
```
crobibero, line 146: Lets change this ENV to `JELLYFIN_ENABLE_IIS` so the user needs to explicitly opt-in to enabling IIS inside of Jellyfin.
```suggestion
                        if (bool.TryParse(Environment.GetEnvironmentVariable("JELLYFIN_ENABLE_IIS"), out var iisEnabled) && iisEnabled)
```

oddstr13:
If this change doesn't affect how Jellyfin is normally run with kestrel, I see no reason to block this PR.

Documentation about this feature should, however, note that; ***IIS is unsupported*** even if the functionality exists, and that issues with it are likely to be ignored by the team, as none of us are using it.

Shadowghost:
I'm not against merging it but only if we ensure that using IIS us unsupported explicitly because this allows the triage team to close issues with IIS.

kimboslice99:
When I write the docs I will note that this is an unsupported configuration, however issues that may arise I would appreciate a mention as im happy to help

Jellyfin.Server/Program.cs
```diff
@@ -139,7 +140,14 @@ private static async Task StartServer(IServerApplicationPaths appPaths, StartupO
                 host = Host.CreateDefaultBuilder()
                     .UseConsoleLifetime()
                     .ConfigureServices(services => appHost.Init(services))
-                    .ConfigureWebHostDefaults(webHostBuilder => webHostBuilder.ConfigureWebHostBuilder(appHost, startupConfig, appPaths, _logger))
+                    .ConfigureWebHostDefaults(webHostBuilder =>
+                    {
+                        webHostBuilder.ConfigureWebHostBuilder(appHost, startupConfig, appPaths, _logger);
+                        if (bool.TryParse(Environment.GetEnvironmentVariable("JELLYFIN_ENABLE_IIS"), out var iisEnabled) && iisEnabled)
+                        {
```
JPVenson, line 147:
```suggestion
                        {
                        _logger.LogCritical("UNSUPPORTED HOSTING ENVIORMENT Microsoft Internet Information Services. The option to run Jellyfin within an IIS is an unsupported and untested feature. Only use at your own digression.");
```

Plus a very clear and very explicit description in the docs and i am ok with it.

JPVenson (changes requested):

Jellyfin.Server/Program.cs
kimboslice99: I take it English is not your first language :) Should read more like

`UNSUPPORTED HOSTING ENVIRONMENT Microsoft Internet Information Services. The option to run Jellyfin within IIS is an unsupported and untested feature. Only use at your own digression.`

kimboslice99:
I just wanted to note here a couple of reasons why I prefer this configuration, also so that the benefits of this PR are realized

- RPS - 3x faster at serving static files on ANCM than over ARR - tested with bombardier. IIS comes _very_ close to kestrels performance here.

- Modules
    - [Geoblocking](https://github.com/RvdHout/IIS-GeoIP2block-Module)
    - [DDoS protection - DIPR](https://learn.microsoft.com/en-us/iis/manage/configuring-security/using-dynamic-ip-restrictions)
    
- Overlapped recycling works well, so Jellyfin can be restarted with minimal downtime
- Response caching

JPVenson:
> I just wanted to note here a couple of reasons why I prefer this configuration, also so that the benefits of this PR are realized
> 
> * RPS - 3x faster at serving static files on ANCM than over ARR - tested with bombardier. IIS comes _very_ close to kestrels performance here.

You can already, without any changes to the server just serve the default WebUI over any WebServer you want. Its completely separate.

> * Modules
>   
>   * [Geoblocking](https://github.com/RvdHout/IIS-GeoIP2block-Module)
>   * [DDoS protection - DIPR](https://learn.microsoft.com/en-us/iis/manage/configuring-security/using-dynamic-ip-restrictions)
> * Response caching

Are all features we usually see at the ReverseProxy responsibility.

> * Overlapped recycling works well, so Jellyfin can be restarted with minimal downtime

Its a home media server, also just a docker restart will have the whole server (starting) in less then a sec. We even support in-process restarting with the next version.

gnattu:
> I just wanted to note here a couple of reasons why I prefer this configuration, also so that the benefits of this PR are realized

I absolutely agree with you that there are situations where leveraging Microsoft IIS can bring notable benefits, particularly in certain environments or configurations. 

However, as @JPVenson mentioned, Jellyfin already offers a mode where the web client can be hosted separately. This flexibility allows users to choose whichever web server suits their preferences or requirements. Many of our users are already taking advantage of this feature, opting to host the web client separately while running Jellyfin in API-only mode.

However, in this mode, the power user understands that they are on their own when issues occur on the web server they have chosen, and the Jellyfin team is not responsible for setting up their own web client or providing a configuration guide. We believe that the IIS mode should fall into the similar support status, but to have an alternate web server built-in might confuse our user to make them think everything would work out of the box like the default kestrel, but this might not always be the case, that's why we mandates scary warning messages and proper documentation of this option.

kimboslice99:
@JPVenson

 There are more files to be served than just the webui, such as covers/posters. These will also be faster served if not over ARR.

> Its a home media server

Ya, so what? Still nice to keep it up as best I can for my users.

>  just a docker restart will have the whole server (starting) in less then a sec.

Interesting to know docker restarts fast

If you correct that grammar and spelling I will commit your suggestion

Jellyfin.Server/Program.cs
JPVenson: After some deliberation with some other team members, it was decided that "on" would be better

kimboslice99: `ENVIORMENT` should be `ENVIRONMENT` :)

JPVenson:
> @JPVenson
> 
> There are more files to be served than just the webui, such as covers/posters. These will also be faster served if not over ARR.
> 
> > Its a home media server
> 
> Ya, so what? Still nice to keep it up as best I can for my users.
> 
> > just a docker restart will have the whole server (starting) in less then a sec.
> 
> Interesting to know docker restarts fast
> 
> If you correct that grammar and spelling I will commit your suggestion

Tbo, browser caching and a fast underlying storage medium or even a reverse proxy with an cache are all techniques that are working absolutely fine. It would be better to improve jellyfins image caching instead of just saying that an IIS "works better". I run my instance over 3 reverse proxies, a VPN though, a 30mbit upload and all docker hosted and the images load in an instant.  Apart from "I want that" i cannot see any benefit the IIS provides towards the existing recommended methods.

Jellyfin.Server/Program.cs
JPVenson: You know you can just push the change yourself right?

sevenrats: friendly reminder that this space is for code-focused, results-oriented discussion.

kimboslice99: I'm headed to work, lol.... I could later

sevenrats: @kimboslice99 i now feel like need to warn you specifically about maintaining a professional tone. My interpretation of this thread is that you are not showing Venson a reasonable level of respect here, and I encourage you to address that shortcoming so that we can all maintain a professional relationship.

sevenrats: one way that you could submit this exact same feedback without coming off as flippant or condescending is to actually push the required changes in. that makes it pretty incontrovertable that you have good intentions -- something that is missing here.

kimboslice99: @sevenrats Did not see your comment. 

Not sure how we've gotten so off track here. I simply did not have time, I thought perhaps if he felt like it he could correct it while I was preoccupied.

Anyways, I will accept the change. Thanks for your input.

kimboslice99:
The point I had was the raw performance difference of ARR vs ANCM. Sure you could add all forms of caching (kernel, user-mode, client, proxy, etc) but I am comparing raw performance with no caching anywhere.

Jellyfin running kestrel with ARR to proxy requests
```
./bombardier -c 2 -d 10s http://192.168.1.249/web/runtime.bundle.js?4e29c36da5d445b8fc84
Bombarding http://192.168.1.249:80/web/runtime.bundle.js?4e29c36da5d445b8fc84 for 10s using 2 connection(s)
[=====================================================================================================================================================] 10s
Done!
Statistics        Avg      Stdev        Max
  Reqs/sec      1099.93      54.28    1700.49
  Latency        1.81ms   197.84us     4.31ms
  HTTP codes:
    1xx - 0, 2xx - 10991, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:    48.33MB/s
```
Jellyfin running with ANCM
```
./bombardier -c 2 -d 10s http://192.168.1.249/web/runtime.bundle.js?4e29c36da5d445b8fc84
Bombarding http://192.168.1.249:80/web/runtime.bundle.js?4e29c36da5d445b8fc84 for 10s using 2 connection(s)
[=====================================================================================================================================================] 10s
Done!
Statistics        Avg      Stdev        Max
  Reqs/sec      1794.15     195.76    2249.11
  Latency        1.11ms   427.27us    36.09ms
  HTTP codes:
    1xx - 0, 2xx - 17937, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:    78.89MB/s
```

I am not attempting to be unprofessional or rude so please do not take it in this manner. I just wish to present my case... I chose a random static resource but the difference is pretty similar for most content

crobibero (approved):

Bond-009 (approved):

Issue #11112, opened 2024-03-05
[Issue]: BDMV playback with subtitles results in crash


adam-kulagowski:
### Please describe your bug

If I attempt to watch BDMV folder without subtitles then everything is fine. If I attempt to watch selected `.m2ts`  directly with subtitles - subtitles appear (no transcoding is done, browser reports `direct play`). If I attempt to watch the same movie but as BDMV folder with subtitles - there is a crash in jellyfin server

### Reproduction Steps

1. download and extract Sintel blu ray ISO from http://download.blender.org/durian/movies/
2. copy 0000.m2ts file outside of BDMV folder structure
3. attempt to watch extrated ISO w/ and w/o subtitles [Eng]
4. attempt to watch `.m2ts` file w/ and w/o subtitles  [Eng]

Note: Subtitles appears at 2:10 in the movie (to avoid searching for them).

### Jellyfin Version

Unstable (master branch)

### Environment

```markdown
- OS: Linux
- Linux Kernel:
- Virtualization: `docker run -d -e  JELLYFIN_FFmpeg__analyzeduration=10000000  -v /tmp/media:/media -p 8096:8096 $USER/jellyfin`
- Clients: Chromium w/ Mp4 play enabled
- Browser: Chromium 121.0.6167.85 
- FFmpeg Version:
- Playback Method: Direct
- Hardware Acceleration: None
- GPU Model: AMD
- Plugins: None
- Reverse Proxy: None
- Base URL: 127.0.0.1:8096
- Networking: Host
- Storage: local mount over linux cifs
```

### Jellyfin logs

```shell
Logs from the attempt to watch BDMV w/ subtitles.

[09:59:34] [INF] [52] Jellyfin.Api.Helpers.MediaInfoHelper: User policy for root. EnablePlaybackRemuxing: True EnableVideoPlaybackTranscoding: True EnableAudioPlaybackTranscoding: True
[09:59:35] [INF] [73] Jellyfin.Api.Controllers.DynamicHlsController: Current HLS implementation doesn't support non-keyframe breaks but one is requested, ignoring that request
[09:59:35] [INF] [73] MediaBrowser.MediaEncoding.Encoder.MediaEncoder: Starting /usr/lib/jellyfin-ffmpeg/ffprobe with args -analyzeduration 10000000 -probesize 1G -i file:"/media/Sintel-a/BDMV/STREAM/00000.m2ts" -threads 0 -v warning -print_format json -show_streams -show_format
[...]

PR #11129, opened 2024-03-09
Specify file as concat when extracting attachments

crobibero:
Fixes https://github.com/jellyfin/jellyfin/issues/11112


Shadowghost (approved):
I was so sure I did not miss this the last time....

nyanmisaka (approved):

Bond-009 (approved):

PR #11136, opened 2024-03-10
Fixes  'replace missing metadata' overwrites existing metadata that does exist

### LINKED ISSUE #9760: [Issue]: Feature 'replace missing metadata' overwrites existing metadata that does exist
### Please describe your bug

I'm running the latest version of Jellyfin on Ubuntu Server, which works great.
The problem is that I have 2 libraries, and when I click on 'Replace missing metadata' in the Web UI, that starts overwriting data that I have manually entered.

This is how it reads the files from the disk without any user modications:
wrong

I then manually edit the name (field 'Title') of all episodes to display the correct episode title, like:
correct

When I click on 'Replace missing metadata' in the Web UI after the manual modifications, it happens again.. So every time I have to manually update all episodes.
The data is being saved correctly, I have rebooted the VM to check and it stores the manual entries as it should. It dissapears as soon as the 'missing' metadata is added. 

### Jellyfin Version

Other

### if other:

10.8.10

### Environment

```markdown
- OS: Ubuntu Server 
- Linux Kernel: Linux
- Virtualization: Hyper-V VM
- Clients: Browser
- Browser: Safari 16.4.1
- FFmpeg Version: 5.1.3
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins: OpenSubtitles
- Reverse Proxy:
- Base URL:
- Networking:
- Storage: OS, media on shared VHDX files using Hyper-V. Mounted under /media/{Series,Movies}
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

scampower3:
Fixes  'replace missing metadata' overwrites existing metadata that does exist

**Changes**
https://github.com/jellyfin/jellyfin/blob/9dbef98a157ca316deb2d634d288d8fe4aeda427/MediaBrowser.Providers/Manager/MetadataService.cs#L776
The above code will always cause new metadata to overwrite the existing metadata. Changed from true to options.ReplaceAllMetadata.
**Issues**
Fixes #9760

### DIFF
--- a/MediaBrowser.Providers/Manager/MetadataService.cs
+++ b/MediaBrowser.Providers/Manager/MetadataService.cs
@@ -773,7 +773,7 @@ namespace MediaBrowser.Providers.Manager
                             MergeData(metadata, temp, Array.Empty<MetadataField>(), false, false);
                         }
 
-                        MergeData(temp, metadata, item.LockedFields, true, false);
+                        MergeData(temp, metadata, item.LockedFields, options.ReplaceAllMetadata, false);
                     }
                 }
             }


Shadowghost (approved):

IDisposable (approved):

MediaBrowser.Providers/Manager/MetadataService.cs
```diff
@@ -773,7 +773,15 @@ public bool CanRefreshPrimary(Type type)
                             MergeData(metadata, temp, Array.Empty<MetadataField>(), false, false);
                         }
 
-                        MergeData(temp, metadata, item.LockedFields, options.ReplaceAllMetadata, false);
+                        // For when new and updated files are scanned, metadata is always replaced since the data at this point is new and there is no prior data.
```
IDisposable, line 776: How about this entire block being the logically equivalent: 
```
MergeData(temp, 
    metadata,
    item.LockedFields, 
    options.MetadataRefreshMode == MetadataRefreshMode.Default || options.ReplaceAllMetadata,
    false);
```

crobibero (approved):

PR #7514, opened 2022-03-28
Extend music parsing

Shadowghost:
Our music parser is notoriously inflexible, so I tried making it better.

**Changes**
* use TagLibSharp to properly read tags from files (preference in this order):
  1. ID3v2
  2. APE
  3. FLAC
  4. Audible
  5. ID3v1
* properly read multiple artists and album artist tags (ffprobe did not provide this)
* properly create and associate persons with tracks
* enable people association on albums
* add all artists from tracks as persons to album
* add album artists from tracks to the album
* add composers to tracks and albums
* add MusicBrainz releasegroup and album Ids to album
* allow subfolders named as the plural of [MusicBrain ReleaseGroups](https://musicbrainz.org/doc/Release_Group/Type)  to be present under an artist
* add proper replace logic to services (some may be duplicate)
* add xmldocs for modified files

**ToDo**
* write tests

**Examples**
Now valid structure:
```
Artist
|---- Compilations
       |---- Album 1
              |---- Track 1
              |---- Track 2
              |---- Track 3
       |---- Album 2
              |---- Track 1
       |---- Album 3
              |---- Track 1
|---- Singles
       |---- Single1
              |---- Track 1
              |---- Track 2
```

**Issues**

PR #11155, opened 2024-03-14
Fix parsing of audio PremiereDate property

### LINKED ISSUE #7667: [Issue] jellyfin still in 10.8.9 overwrites correct <releasedate> nfo value with bogus "01-01-<correctyear>"-value
### Please describe your bug

[respecified after comment]

I ordered my library with nfo files inside the album folders. Inside the nfo is a tag 

`<releasedate>2017-06-14</releasedate>`

which is apparently not pulled out correctly, but instead inside jellyfin always 1st of January is used, e.g.

`2017-01-01`

Update:

I now installed v10.8.9 and experienced the same again - after triggering "Scan for New and updated files" for my library, everything has now 01-01 of the correct releaseyear. i wrote a script to pull out the correct dates of my second jellyfin instance and inject them into the faulty nfo folders - after the rescan everything is back on 01-01-xxxx ...

I'll provide any proof, which is needed - happy to help fixing this, although i have no experience with NET Core.

Edit2: Retested this with the AUR-version of jellyfin 10.8.3 and with a new docker container. Both had this behavior for me. latest 10.7 no issues

#### To reproduce:
 1. Get Album Folder with correct releasedate in album.nfo
 2. Setup new jellyfin server with v10.8.9 and add the folder (docker or not)
 3. watch the nfo file change the correct date to 01.01.xxxx
 
#### (Alternatively:)
 
1. mount volume with music albums and correct nfo as :ro
2. load the source into a fresh installation of jellyfin 10.8.{1...9}
3. The nfo is not changed anymore obviously, but inside jellyfin still 01.01.yyyy is chosen as date.

This makes 10.8.x unusable for me currently...

### Jellyfin Version

10.8.9 retested, issue still there 

### Environment

```markdown
- OS:Manjaro, debian, raspbian... all show the same problem
- Virtualization: Docker
```
### Code of Conduct

mft5ln9una4a1
kkv15x9una4a1

sel10ut:
**Changes**

- Retrieve the release date from a ffmpeg probe, instead of hard-coding it.

While parsing the music library, PremiereDate was written incorrectly, only getting the year properly, which resulted in it having a format of YYYY-01-01. This commit #7514 introduced `TagLibSharp` library, which, looks like, despite reading full raw metadata from a file, unfortunately, doesn't provide a method to retrieve a full date from it. That behavior is incorrect for some tag formats.

**Issues**

fixes #7667

### DIFF
--- a/MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
+++ b/MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
@@ -228,6 +228,7 @@ namespace MediaBrowser.Providers.MediaInfo
 
             audio.RunTimeTicks = mediaInfo.RunTimeTicks;
             audio.Size = mediaInfo.Size;
+            audio.PremiereDate = mediaInfo.PremiereDate;
 
             if (!audio.IsLocked)
             {
@@ -370,7 +371,6 @@ namespace MediaBrowser.Providers.MediaInfo
                 {
                     var year = Convert.ToInt32(tags.Year);
                     audio.ProductionYear = year;
-                    audio.PremiereDate = new DateTime(year, 01, 01);
                 }
 
                 if (!audio.LockedFields.Contains(MetadataField.Genres))
--- a/MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
+++ b/MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
@@ -371,6 +371,11 @@ namespace MediaBrowser.Providers.MediaInfo
                 {
                     var year = Convert.ToInt32(tags.Year);
                     audio.ProductionYear = year;
+
+                    if (!audio.PremiereDate.HasValue)
+                    {
+                        audio.PremiereDate = new DateTime(year, 01, 01);
+                    }
                 }
 
                 if (!audio.LockedFields.Contains(MetadataField.Genres))


Bond-009 (approved):

Shadowghost (approved):

sel10ut:
Looks like Jellyfin's ffprobe can fail to parse the date on some specific ID3v2.x tag version. Just compared Jellyfin's ffprobe to system's ffprobe on the same file and date was interpreted differently: 
- Jellyfin's ffprobe `"TYER": "2021;2021"`
- system's ffprobe `"date": "2021"`

So having a hard-coded fallback for safety is good, in this case TagLib might parse correctly if ffprobe fails. Need more investigation

sel10ut:
Found some more mp3 files in my database, though both probes output same `TYER` tag with the same value. The value is actually a full date inside a `TYER` tag, which is non-standard.

> Looks like Jellyfin's ffprobe can fail to parse the date on some specific ID3v2.x tag version. Just compared Jellyfin's ffprobe to system's ffprobe on the same file and date was interpreted differently:
> 
>     * Jellyfin's ffprobe `"TYER": "2021;2021"`
> 
>     * system's ffprobe `"date": "2021"`
> 
> 
> So having a hard-coded fallback for safety is good, in this case TagLib might parse correctly if ffprobe fails. Need more investigation

This one was probably just a broken tag with the year repeated twice. So most likely nothing wrong with ffprobe. TagLib is indeed can parse a correct year from a non-standard `TYER` tag. But what bothered me initially is that on fresh container it parsed date correctly, but on old server it was always null after running "Replace all metadata". Cannot reproduce that anymore

JPVenson (approved):

PR #11161, opened 2024-03-15
Fix the fragile segment deletion

nyanmisaka:
**Changes**
- Add `TranscodingSegmentCleaner` to replace ffmpeg's hlsenc deletion

FFmpeg deletes segments based on its own transcoding progress,
but we need to delete segments based on client download progress.
Since disk and GPU speeds vary, using hlsenc's built-in deletion
will result in premature deletion of some segments. As a consequence,
the server has to constantly respin new ffmpeg instances, resulting
in choppy video playback.

- Add sanity check for `ThrottleDelaySeconds`
- Set input `readrate` for using SegmentDeletion with stream-copy

**Issues**
- The [old implementation](https://github.com/jellyfin/jellyfin/pull/8744) just happened to work. It's actually very fragile and can easily break playback by just entering some smaller `SegmentKeepSeconds` value.

### DIFF
--- a/tests/Jellyfin.MediaEncoding.Tests/EncoderValidatorTests.cs
+++ b/tests/Jellyfin.MediaEncoding.Tests/EncoderValidatorTests.cs
@@ -17,16 +17,11 @@ namespace Jellyfin.MediaEncoding.Tests
         }
 
         [Theory]
+        [InlineData(EncoderValidatorTestsData.FFmpegV611Output, true)]
         [InlineData(EncoderValidatorTestsData.FFmpegV60Output, true)]
         [InlineData(EncoderValidatorTestsData.FFmpegV512Output, true)]
         [InlineData(EncoderValidatorTestsData.FFmpegV44Output, true)]
-        [InlineData(EncoderValidatorTestsData.FFmpegV432Output, true)]
-        [InlineData(EncoderValidatorTestsData.FFmpegV431Output, true)]
-        [InlineData(EncoderValidatorTestsData.FFmpegV43Output, true)]
-        [InlineData(EncoderValidatorTestsData.FFmpegV421Output, true)]
-        [InlineData(EncoderValidatorTestsData.FFmpegV42Output, true)]
-        [InlineData(EncoderValidatorTestsData.FFmpegV414Output, true)]
-        [InlineData(EncoderValidatorTestsData.FFmpegV404Output, true)]
+        [InlineData(EncoderValidatorTestsData.FFmpegV432Output, false)]
         [InlineData(EncoderValidatorTestsData.FFmpegGitUnknownOutput2, true)]
         [InlineData(EncoderValidatorTestsData.FFmpegGitUnknownOutput, false)]
         public void ValidateVersionInternalTest(string versionOutput, bool valid)
@@ -38,17 +33,12 @@ namespace Jellyfin.MediaEncoding.Tests
         {
             public GetFFmpegVersionTestData()
             {
+                Add(EncoderValidatorTestsData.FFmpegV611Output, new Version(6, 1, 1));
                 Add(EncoderValidatorTestsData.FFmpegV60Output, new Version(6, 0));
                 Add(EncoderValidatorTestsData.FFmpegV512Output, new Version(5, 1, 2));
                 Add(EncoderValidatorTestsData.FFmpegV44Output, new Version(4, 4));
                 Add(EncoderValidatorTestsData.FFmpegV432Output, new Version(4, 3, 2));
-                Add(EncoderValidatorTestsData.FFmpegV431Output, new Version(4, 3, 1));
-                Add(EncoderValidatorTestsData.FFmpegV43Output, new Version(4, 3));
-                Add(EncoderValidatorTestsData.FFmpegV421Output, new Version(4, 2, 1));
-                Add(EncoderValidatorTestsData.FFmpegV42Output, new Version(4, 2));
-                Add(EncoderValidatorTestsData.FFmpegV414Output, new Version(4, 1, 4));
-                Add(EncoderValidatorTestsData.FFmpegV404Output, new Version(4, 0, 4));
-                Add(EncoderValidatorTestsData.FFmpegGitUnknownOutput2, new Version(4, 0));
+                Add(EncoderValidatorTestsData.FFmpegGitUnknownOutput2, new Version(4, 4));
                 Add(EncoderValidatorTestsData.FFmpegGitUnknownOutput, null);
             }
         }
--- a/tests/Jellyfin.MediaEncoding.Tests/EncoderValidatorTestsData.cs
+++ b/tests/Jellyfin.MediaEncoding.Tests/EncoderValidatorTestsData.cs
@@ -2,6 +2,18 @@ namespace Jellyfin.MediaEncoding.Tests
 {
     internal static class EncoderValidatorTestsData
     {
+        public const string FFmpegV611Output = @"ffmpeg version n6.1.1-16-g33efa50fa4-20240317 Copyright (c) 2000-2023 the FFmpeg developers
+built with gcc 13.2.0 (crosstool-NG 1.26.0.65_ecc5e41)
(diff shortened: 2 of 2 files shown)


Shadowghost (approved):

MediaBrowser.Controller/MediaEncoding/TranscodingSegmentCleaner.cs
```diff
@@ -0,0 +1,188 @@
...
+            if (job.Type == TranscodingJobType.Hls)
+            {
+                DeleteHlsSegmentFiles(path, idxMin, idxMax);
+            }
+        }
+        catch (IOException ex)
+        {
+            _logger.LogError(ex, "Error deleting segment file(s) {Path}", path);
+
+            await DeleteSegmentFiles(job, idxMin, idxMax, retryCount + 1, 500).ConfigureAwait(false);
```
JPVenson, line 141: If this throws an exception, and a file does not get deleted it will be enumerated the next timer tick and will fail again

nyanmisaka: Any suggestions? The only reason I can think of for this to fail is that the user has set incorrect permissions for the transcodes folder.

JPVenson: Yes but its the filesystem so stuffs bound to happen.
I mean as this whole class is already scoped just remember all segments you already tried to delete in a list and ignore them on any subsequent loops.

cvium: I don't see a scenario where one segment deletion failing due to non-transient errors (transient would be ok to retry) doesn't mean all future deletions for any segment won't fail as well. Wouldn't it make more sense to look at the IO error and simply stop the cleaner if it's a perm-related error?

nyanmisaka: If it's a perm-related issue then the transcoding won't happen at all, or it should break immediately (due to ffmpeg failing).

MediaBrowser.Controller/MediaEncoding/TranscodingSegmentCleaner.cs
```diff
@@ -0,0 +1,182 @@
...
+    {
+        var directory = Path.GetDirectoryName(outputFilePath)
+                        ?? throw new ArgumentException("Path can't be a root directory.", nameof(outputFilePath));
+
+        var name = Path.GetFileNameWithoutExtension(outputFilePath);
+
+        var filesToDelete = _fileSystem.GetFilePaths(directory)
+            .Where(f => (!_excludeFilePaths?.Contains(f) ?? true)
+                        && long.TryParse(Path.GetFileNameWithoutExtension(f).Replace(name, string.Empty, StringComparison.Ordinal), out var idx)
+                        && (idx >= idxMin && idx <= idxMax));
```
JPVenson, line 150:
```suggestion
            .Except(_excludeFilePaths).Where(f => long.TryParse(Path.GetFileNameWithoutExtension(f).Replace(name, string.Empty, StringComparison.Ordinal), out var idx)
                        && (idx >= idxMin && idx <= idxMax));
```

JPVenson (approved):

MediaBrowser.Controller/MediaEncoding/TranscodingSegmentCleaner.cs
cvium: Then I'm not sure I see the point of keeping track of failures. Either it can be retried later or none of the deletions will ever succeed. Or am I missing something?

nyanmisaka: It may be locked by system programs, which is more likely to happen on Windows, such as virus scanning. If the system does not unlock some segment files, Jellyfin will continue to throw IOException and record it in the log. I'm okay with both since it's more of a cosmetic change to the log.

@JPVenson Any more comments on this?

JPVenson: Yea the logbloat is my biggest concern. If you would just not log the subsequent errors i am fine with it.

JPVenson: I mean, if we want to make it sofisticated you could use a FileSystemWatcher and a Channel with a delay. So you will only ever process an item once without a loop.

nyanmisaka: That one seems a bit overkill. Maybe lowering the log level to debug is also feasible? This should avoid spamming in the log file in general use cases.

JPVenson:
Overall question:
If we delete the transcode segments while the stream is still running, would that make seeking more annoying because we would need to rebuild that segment?

nyanmisaka:
> Overall question: If we delete the transcode segments while the stream is still running, would that make seeking more annoying because we would need to rebuild that segment?

Similar to the transcoding throttler, this feature is only recommended for users who do not seek frequently. If the user rewinds beyond the set `SegmentKeepSeconds`, these deleted segments will be re-transcoded. Or set a larger `SegmentKeepSeconds` to increase tolerance.

MediaBrowser.Controller/MediaEncoding/TranscodingSegmentCleaner.cs
nyanmisaka: I drop the excludeFilePaths and lower the log level to debug to avoid spamming in the log file. Now even if individual segments fail to be deleted, they will not be recorded in the normal log unless the user adjusts the log level to debug.

Jellyfin.Api/Controllers/DynamicHlsController.cs
```diff
@@ -1604,7 +1604,7 @@ private string GetCommandLineArguments(string outputPath, StreamState state, boo
                 Path.GetFileNameWithoutExtension(outputPath));
         }
 
-        var hlsArguments = GetHlsArguments(isEventPlaylist, state.SegmentLength);
+        var hlsArguments = string.Format(CultureInfo.InvariantCulture, "-hls_playlist_type {0} -hls_list_size 0", isEventPlaylist ? "event" : "vod");
```
cvium, line 1607: string interpolation is preferred (I don't remember if the parentheses are required for ternarys inside string interpolation)
```suggestion
        var hlsArguments = $"-hls_playlist_type {(isEventPlaylist ? "event" : "vod")} -hls_list_size 0"; 
```

MediaBrowser.Controller/MediaEncoding/TranscodingSegmentCleaner.cs
```diff
@@ -0,0 +1,178 @@
...
+        foreach (var file in filesToDelete)
+        {
+            try
+            {
+                _logger.LogDebug("Deleting HLS segment file {0}", file);
+                _fileSystem.DeleteFile(file);
+            }
+            catch (IOException ex)
+            {
+                (exs ??= new List<Exception>(4)).Add(ex);
```
cvium, line 159: Why 4?

cvium (approved):
LGTM, just a few suggestions/questions

MediaBrowser.Controller/MediaEncoding/TranscodingSegmentCleaner.cs
nyanmisaka: I was also wondering why 4. Because the code here is basically taken from TranscodeManager. I can remove it later.

https://github.com/jellyfin/jellyfin/blob/b6d130ae2d337294904f6d46d943e9f82c162978/MediaBrowser.MediaEncoding/Transcoding/TranscodeManager.cs#L324

cvium: Strange. I would just remove it.

Jellyfin.Api/Controllers/DynamicHlsController.cs
nyanmisaka: Tested and it works. Parentheses is necessary here.

MediaBrowser.Controller/MediaEncoding/TranscodingSegmentCleaner.cs
nyanmisaka: https://github.com/search?q=repo%3Ajellyfin%2Fjellyfin+%3E%284&type=code

Similar situations occur elsewhere, I just don't get it. Do I need to clean them up in this PR?

Bond-009: Microoptimization by me, by default list gets created with an empty internal array, when the add method gets called the fast path is designed when there is enough space for the item, otherwise another method is called (not inlined) that grows the array. All of this can be skipped by creating the list with the first default capacity of 4.
Not necessary at all, just something I added when I wrote the code is is copied from.

nyanmisaka: I also speculate that this may be an optimization for the CLR. So i didn't touch it.

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -51,6 +51,8 @@ public partial class EncodingHelper
         private readonly Version _minFFmpegOclCuTonemapMode = new Version(5, 1, 3);
         private readonly Version _minFFmpegSvtAv1Params = new Version(5, 1);
         private readonly Version _minFFmpegVaapiH26xEncA53CcSei = new Version(6, 0);
+        private readonly Version _minFFmpegAutoscaleOption = new Version(4, 4);
```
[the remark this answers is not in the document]
nyanmisaka: There seems to be some distro on ff 4.4, and ff 4.4 is an LTS version last released on 2023-04-12. For the most basic software transcoding, there seems to be no need to increase the minimum ffmpeg version.

I'd like to hold off on doing this as long as possible unless our core functionality really requires it.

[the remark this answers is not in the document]
nyanmisaka: When we get to ff 7.0, I will try to increase the minimum version to ff 5.x.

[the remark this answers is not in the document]
Bond-009: I meant updating from 4.0 (2018?) as minimum to 4.4 so we don't need these checks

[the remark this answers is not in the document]
nyanmisaka: Done in ae7c0c83e9f7b5e35ac067f09a92d10d33407219

MediaBrowser.Controller/MediaEncoding/TranscodingSegmentCleaner.cs
JPVenson: @nyanmisaka this _used_ to be a known optimization for Dotnet Framework but ever since Framework 4.7.2 the clr got very good at predicating the initial size of list arrays. Also we are nowhere near the performance or amount of arrays where this would be justified.

Bond-009 (approved):

PR #11162, opened 2024-03-15
Remove legacy build utilities

crobibero:
All non-rpm building has been moved to https://github.com/jellyfin/jellyfin-packaging/
We now point to RPM Fusion for RPM Builds (https://github.com/rpmfusion/jellyfin)

Clean this up to prevent users from using legacy scripts and tools


nielsvanvelzen (approved):

Bond-009 (approved):

Shadowghost (approved):

joshuaboniface (approved):
Ship it :rocket:

Erotemic:
I'm confused about what the proper way to be building now is. It seems jellyfin-packaging is now how the code is built, but all packagers are still supported? And the RPM-fusion is for fedora specific packages? A README on that repo would be helpful. Also, would it help to have the jellyfin-packaging README reference the RPM repo? Its also unclear what the legacy build system that was removed was.

Does it make sense to include some instructions in the jellyfin-packaging repo to point at how to build for development as well? I.e. "this repo is for building deployable packages, for building in a development environment see...".

Lastly, in the jellyfin-packaging repo, it might be useful to start off with the line: "This repository contains packaging for Jellyfin 10.9.0 and above, for use by manual builders or the CI system with GitHub workflows.", then have the blurb about what jellyfin is, and then get into the list of components (which I found very helpful).

These are just my $0.02 as someone interested in working on the project, who hasn't done any major server development before.

Erotemic:
Yeah, the blurb threw me off. At first I thought I was looking at the main repo, then I thought maybe the wrong README was coppied over, and then I realized the information I wanted was there, it was just under what seemed to be some boilerplate text. It probably will be helpful to keep it (or a revised version there) for people new to the repo, but it would me nice if the README confirmed what I think the code is for, and then expanded on it. Perhaps these could even be as small of a change as changing the header from "Jellyfin" to "Jellyfin Packaging". Then I would be cued into the fact that the wall of text is more likely to explain information about the repo rather than just be generic text about jellyfin.

joshuaboniface:
I've pushed up some changes both to the README (shorten the boilerplate header, expand the second paragraph a bit, add a Quickstart section with explanations) and the `build.py` script (better argument parsing for improved reliability), so please check that out and let me know!

Erotemic:
I don't see a README in rpmfusion/jellyfin, but the updates to jellyfin-packaging look great. I briefly looked over it and it was much easier to follow.

vessd:
https://jellyfin.org/docs/general/installation/source/ is very confusing right now.

PR #11164, opened 2024-03-16
Introduce scrutor for automagic dependency registration

crobibero:
This allows the LDAP plugin to be registered without the plugin needing to manually register it.

I also tried hooking up the `IWebSocketListener` registration but it looks like not all assemblies are loaded yet? Further investigation is required.


Shadowghost (approved):

JPVenson:
closed due to inactivity

PR #11167, opened 2024-03-16
Bugfix/fix permission check on trickplay

JPVenson:
**Changes**
 - Fixed Trickplay api not checking for Visibility for requested items

**Issues**
Fixes #11166

PR #11168, opened 2024-03-16
Added user permission check for LyricsApi endpoints

JPVenson:
**Changes**
- Added permission checks for Lyrics api methods


Shadowghost (approved):

cvium:
Why is all this not just part of the auth policy?

JPVenson:
@cvium because we are a few weeks away from the release and i am already working on a generalized method to handle that. In the meantime this and #11167 should be fixed anyway.

cvium:
You are duplicating the auth stuff from DefaultAuthorizationHandler. The only check you need is the IsVisible check, the rest is already handled.

JPVenson:
@cvium ok yes fair point

crobibero:
https://github.com/jellyfin/jellyfin/pull/11171

PR #11170, opened 2024-03-16
Add filter to validate item access

crobibero:
Should catch 99% of the endpoints, ones with id in body will have to be manually updated

### DIFF
--- a/Jellyfin.Api/Filters/ItemAccessFilter.cs
+++ b/Jellyfin.Api/Filters/ItemAccessFilter.cs
@@ -1,0 +1,110 @@
+using System;
+using Jellyfin.Api.Extensions;
+using Jellyfin.Api.Helpers;
+using Jellyfin.Extensions;
+using MediaBrowser.Controller.Entities;
+using MediaBrowser.Controller.Library;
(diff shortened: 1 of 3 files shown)


JPVenson:
i thought about doing something similar but this loads and discards the user and Item entity for each request its handling and each endpoint then loads the same entity again which is not great

crobibero:
UserManager.GetUserById, LibraryManager.GetItemById, and LibraryManager.UserRootFolder all return cached objects, so I don't think loading and discarding makes a difference

Jellyfin.Api/Filters/ItemAccessFilter.cs
```diff
@@ -0,0 +1,115 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Jellyfin.Api.Filters;
+
+/// <summary>
+/// Action filter to assess whether a user can access an item.
+/// </summary>
+public sealed class ItemAccessFilter : IActionFilter
+{
+    private const string ItemIdArgument = "itemId";
```
JPVenson, line 18: i am not too stoked for this to be name based. To much "magic" involved if you just read the controller side.

PR #11171, opened 2024-03-17
Validate item access

crobibero:
~Still ~15 direct usages of `LibraryManager.GetItemById` but they were more involved.~


Jellyfin.Api/Controllers/InstantMixController.cs
```diff
@@ -259,11 +279,16 @@ public class InstantMixController : BaseJellyfinApiController
...
     {
-        var item = _libraryManager.GetItemById(itemId);
         userId = RequestHelpers.GetUserId(User, userId);
         var user = userId.IsNullOrEmpty()
             ? null
             : _userManager.GetUserById(userId.Value);
+        var item = _libraryManager.GetItemById(itemId, user);
+        if (item is null)
+        {
+            return NotFound();
```
Shadowghost, line 289: Should be added to the XMLdoc (same for other changed endpoints)

Jellyfin.Api/Controllers/LibraryController.cs
```diff
@@ -335,17 +333,11 @@ public async Task<ActionResult> RefreshLibrary()
     [ProducesResponseType(StatusCodes.Status404NotFound)]
     public ActionResult DeleteItem(Guid itemId)
     {
-        var isApiKey = User.GetIsApiKey();
```
Shadowghost, line 338: Is this properly respected in you changes? If an API key is used Items should always be returned

Emby.Server.Implementations/Library/LibraryManager.cs
```diff
@@ -1232,6 +1228,20 @@ public BaseItem GetItemById(Guid id)
...
 
+        /// <inheritdoc />
+        public BaseItem GetItemById(Guid id, Guid userId)
+        {
+            var user = userId.IsEmpty() ? null : _userManager.GetUserById(userId);
+            return GetItemById(id, user);
+        }
+
+        /// <inheritdoc />
+        public BaseItem GetItemById(Guid id, User user)
```
cvium, line 1239: are these overloads needed?

crobibero: Not needed, but I figured since we sometimes have the user object we might as well use it and not attempt to retrieve it again

cvium: I am talking about the generic and BaseItem specific overloads. Isn't the generic sufficient?

crobibero: Ah- the BaseItem specific already existed and the generic was recently added.
I continued adding the BaseItem specific to stay the pattern but it would be simple enough to remove and update the callers

cvium (approved):

Bond-009 (approved):

PR #11181, opened 2024-03-18
Add default auth policy to generated openapi spec

crobibero:


Shadowghost (approved):

Jellyfin.Server/Filters/SecurityRequirementsOperationFilter.cs
```diff
@@ -6,86 +6,72 @@
...
+/// <summary>
+/// Security requirement operation filter.
+/// </summary>
+public class SecurityRequirementsOperationFilter : IOperationFilter
 {
-    /// <summary>
-    /// Security requirement operation filter.
-    /// </summary>
-    public class SecurityRequirementsOperationFilter : IOperationFilter
+    private const string DefaultAuthPolicy = "Default";
```
nielsvanvelzen, line 16: This was previously called `DefaultAuthorization`, I can change it easily in the SDK but perhaps we want to stay consistent?

Jellyfin.Server/Filters/SecurityRequirementsOperationFilter.cs
```diff
@@ -6,86 +6,72 @@
...
-        {
-            var requiredScopes = new List<string>();
+        var requiredScopes = new List<string>();
 
-            var requiresAuth = false;
-            // Add all method scopes.
-            foreach (var attribute in context.MethodInfo.GetCustomAttributes(true))
+        var requiresAuth = false;
+        // Add all method scopes.
+        foreach (var authorizeAttribute in context.MethodInfo.GetCustomAttributes(_attributeType, true).Cast<AuthorizeAttribute>())
```
nielsvanvelzen, line 26: Some controllers, like the CollectionController, specify the attribute on the class instead of the method. This check fails for those.

nielsvanvelzen (changes requested):

Jellyfin.Server/Filters/SecurityRequirementsOperationFilter.cs
crobibero: L40-53 add the controller level scopes

nielsvanvelzen: Ok then something else is causing some issue. The `CreateCollection` method only contains `CollectionManagement`, not `DefaultAuthorization`.

crobibero: That makes sense, the endpoint only has that requirement since it is a subset of DefaultAuthorization

nielsvanvelzen: How should I know which ones are subsets of which ones? The OpenAPI specification doesn't provide any data about that.
I think it will be better to include all parents.

nielsvanvelzen (approved):
Turns out we've now even improved the output compared to 10.8!

Bond-009 (approved):

PR #11188, opened 2024-03-19
Fix Contains for Sets

goremykin:
Our [Contains](https://github.com/jellyfin/jellyfin/blob/f49de51225b2206609df6a89f3cbb5fd7459ff68/src/Jellyfin.Extensions/EnumerableExtensions.cs#L19) extension method from EnumerableExtensions.cs is used for HashSets and actually kills their performance.
So I decided to fix these places and limit the extension method by IReadOnlyList to prevent this in the future.

Please write if you have better ideas.

### DIFF
--- a/Emby.Server.Implementations/Library/Resolvers/PhotoResolver.cs
+++ b/Emby.Server.Implementations/Library/Resolvers/PhotoResolver.cs
@@ -4,13 +4,11 @@ using System.Linq;
 using Emby.Naming.Common;
 using Emby.Naming.Video;
 using Jellyfin.Data.Enums;
-using Jellyfin.Extensions;
 using MediaBrowser.Controller.Drawing;
 using MediaBrowser.Controller.Entities;
 using MediaBrowser.Controller.Library;
 using MediaBrowser.Controller.Providers;
 using MediaBrowser.Controller.Resolvers;
-using MediaBrowser.Model.Entities;
 
 namespace Emby.Server.Implementations.Library.Resolvers
 {
@@ -103,8 +101,8 @@ namespace Emby.Server.Implementations.Library.Resolvers
         {
             ArgumentNullException.ThrowIfNull(path);
 
-            var extension = Path.GetExtension(path.AsSpan()).TrimStart('.');
-            if (!imageProcessor.SupportedInputFormats.Contains(extension, StringComparison.OrdinalIgnoreCase))
+            var extension = Path.GetExtension(path.AsSpan()).TrimStart('.').ToString();
+            if (!imageProcessor.SupportedInputFormats.Contains(extension))
             {
                 return false;
             }
--- a/src/Jellyfin.Drawing.Skia/SkiaEncoder.cs
+++ b/src/Jellyfin.Drawing.Skia/SkiaEncoder.cs
@@ -1,7 +1,9 @@
 using System;
+using System.Collections.Frozen;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
+using System.Linq;
 using BlurHashSharp.SkiaSharp;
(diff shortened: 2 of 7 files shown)


Emby.Server.Implementations/Library/Resolvers/PhotoResolver.cs
```diff
@@ -103,8 +101,8 @@ internal static bool IsImageFile(string path, IImageProcessor imageProcessor)
         {
             ArgumentNullException.ThrowIfNull(path);
 
-            var extension = Path.GetExtension(path.AsSpan()).TrimStart('.');
-            if (!imageProcessor.SupportedInputFormats.Contains(extension, StringComparison.OrdinalIgnoreCase))
+            var extension = Path.GetExtension(path.AsSpan()).TrimStart('.').ToString();
```
goremykin, line 104: Unfortunately, hash sets don't have an overload for spans, so we have to allocate a string

goremykin: Actually, I would expose `bool IsInputFormatSupported(string inputFormat)` method instead of `IReadOnlyCollection<string> SupportedInputFormats { get; }` that is a set under the hood.
So we don't have to think about what is under the hood and use what Contains overload should we use.

src/Jellyfin.Drawing.Skia/SkiaEncoder.cs
```diff
@@ -69,8 +71,8 @@ public SkiaEncoder(ILogger<SkiaEncoder> logger, IApplicationPaths appPaths)
     public bool SupportsImageEncoding => true;
 
     /// <inheritdoc/>
-    public IReadOnlyCollection<string> SupportedInputFormats =>
-        new HashSet<string>(StringComparer.OrdinalIgnoreCase)
+    public IReadOnlyCollection<string> SupportedInputFormats { get; } =
```
goremykin, line 74: Now it's immutable, so we can create it once instead of creating on every call

src/Jellyfin.Extensions/EnumerableExtensions.cs
```diff
@@ -16,7 +16,7 @@ public static class EnumerableExtensions
     /// <param name="stringComparison">The string comparison.</param>
     /// <returns>A value indicating whether the value is contained in the collection.</returns>
     /// <exception cref="ArgumentNullException">The source is null.</exception>
-    public static bool Contains(this IEnumerable<string> source, ReadOnlySpan<char> value, StringComparison stringComparison)
+    public static bool Contains(this IReadOnlyList<string> source, ReadOnlySpan<char> value, StringComparison stringComparison)
```
goremykin, line 19: Another option - just rename it. So we will be able to use it with IEnumerable

Shadowghost:
@goremykin could you please rebase this PR so we can get it reviewed and merged? 😃

goremykin:
@Shadowghost, done

Shadowghost (approved):

crobibero (commented):
I think it would be better to just add a case for in our `Contains` extension

```c#
if (source is HashSet<string> set)
{
    return set.Contains(value.ToString());
}
```

goremykin:
@crobibero, Don't forget about `StringComparison`. Of course we can map it to `StringComparer` but usually it is specified in the constructor when you create a set, not on each `Contains` call.

goremykin:
I still think the `Contains` extension shouldn't be used with any kind of Sets and should be renamed to prevent it's accidental use. But taking into account new [collection lookups with spans](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-9/libraries#collection-lookups-with-spans) all these changes should be re-implemented from scratch. So I am closing this PR.

PR #11198, opened 2024-03-23
fix: use Metal tonemap instead of OpenCL

gnattu:
**Changes**

As discussed in jellyfin/jellyfin-ffmpeg#357, the current OpenCL kernel runs extremely slow on Apple Silicon. This change updates the filter chain to use the Metal implementation of the tone map filter to handle Dolby Vision Profile 5 videos introduced by jellyfin/jellyfin-ffmpeg#369.

For Intel Macs, the OpenCL kernel is not compatible with certain GPUs and we will have the following error:

```
[AVHWDeviceContext @ 0x7faa2510e200] OpenCL error: [CL_IMAGE_FORMAT_NOT_SUPPORTED] : OpenCL Error : clCreateImage failed: The specified format is not supported for the specified image type.
[Parsed_tonemap_opencl_5 @ 0x7faa2510b940] Failed to create image for dither matrix: -10.
```

We need the Metal kernel to properly support Intel Macs.

**Issues**

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -1377,6 +1377,14 @@ namespace MediaBrowser.Controller.MediaEncoding
                 return FormattableString.Invariant($" -rc_mode VBR -b:v {bitrate} -maxrate {bitrate} -bufsize {bufsize}");
             }
 
+            if (string.Equals(videoCodec, "h264_videotoolbox", StringComparison.OrdinalIgnoreCase)
+                || string.Equals(videoCodec, "hevc_videotoolbox", StringComparison.OrdinalIgnoreCase))
+            {
+                // The `maxrate` and `bufsize` options can potentially lead to performance regression
+                // and even encoder hangs, especially when the value is very high.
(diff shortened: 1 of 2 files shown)


gnattu:
Mark as draft until decisions are made about whether this will be included in 10.9.

nyanmisaka:
It's probably worth mentioning that `tonemap_opencl` doesn't work on Intel based Macs. Also change the title feat->fix: xxx.

nyanmisaka (approved):
Then we will have to tag a jf-ffmpeg 6.0.1-5 before RC1 to match this changes.

crobibero (approved):

Bond-009 (approved):

nyanmisaka:
@gnattu You may also want to modify this string and their translations. (to OpenCL/CUDA/Metal or just GPGPU)

https://github.com/jellyfin/jellyfin-web/blob/fa6981cb7fe1f1e976b0989363a54654df83d2f5/src/strings/en-us.json#L42
https://github.com/jellyfin/jellyfin-web/blob/fa6981cb7fe1f1e976b0989363a54654df83d2f5/src/strings/en-us.json#L1553-L1555

PR #11204, opened 2024-03-24
Don't expect `BaseItem` to be a movie/video file.

revam:
**Changes**

This fix is mainly so I can mass-add series _and_ movie entries using a `IMultiItemResolver` without having to resort to complicated logic using _both_ a `IItemResolver` and a `IMultiItemResolver` by splitting up what gets added where.

I've also added three new interface methods to the `IDirectoryService`, one of which is used in the modified
`ResolverHelper.SetInitialItemValues(…)` to get the file system entry info for the item regardless of which type the file system entry is.

In my local testing so far I haven't found any issues introduced by this change.

**Issues**

### DIFF
--- a/MediaBrowser.Controller/Providers/DirectoryService.cs
+++ b/MediaBrowser.Controller/Providers/DirectoryService.cs
@@ -63,13 +63,13 @@ namespace MediaBrowser.Controller.Providers
         public FileSystemMetadata? GetFile(string path)
         {
             var entry = GetFileSystemEntry(path);
-            return entry != null && !entry.IsDirectory ? entry : null;
+            return entry is not null && !entry.IsDirectory ? entry : null;
         }
 
         public FileSystemMetadata? GetDirectory(string path)
         {
             var entry = GetFileSystemEntry(path);
-            return entry != null && entry.IsDirectory ? entry : null;
+            return entry is not null && entry.IsDirectory ? entry : null;
         }
 
         public FileSystemMetadata? GetFileSystemEntry(string path)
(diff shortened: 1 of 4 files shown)


revam:
Also it should be noted that if I were to do the workaround of using both a `IMultiItemResolver` and `IItemResolver` then it will fail to properly create the Show/Series entries for media libraries without a collection type when the collection doesn't contain any movies (yet), because of how the built-in resolving logic and how the built-in movie resolver works, in addition to how we need to defer the creation of shows/series onto the next set of single item resolvers because the `Items` field on the `MultiItemResolverResult` contains an empty array (because we don't have any movies added to the collection yet) and thus the internals will skip the `ExtraFiles` even if they're set because the `Items` is empty, and move on to the built-in movie resolver which will mark all the files within the folders as movies and it won't ever reach the single item resolvers.

But with this code in place I can place my resolver before the movie resolver and add both the series and movies at once without any issues.

Gauvino:
Failing build action with your change

revam:
I fixed up the tests that were failing and added two new tests for directories.

nielsvanvelzen (approved):

Shadowghost (approved):

MediaBrowser.Controller/Providers/DirectoryService.cs
```diff
@@ -61,11 +61,23 @@ public List<FileSystemMetadata> GetFiles(string path)
...
         public FileSystemMetadata? GetFile(string path)
+        {
+            var entry = GetFileSystemEntry(path);
+            return entry != null && !entry.IsDirectory ? entry : null;
+        }
+
+        public FileSystemMetadata? GetDirectory(string path)
+        {
+            var entry = GetFileSystemEntry(path);
+            return entry != null && entry.IsDirectory ? entry : null;
```
Bond-009, line 72:
```suggestion
            return entry is not null && !entry.IsDirectory ? entry : null;
        }

        public FileSystemMetadata? GetDirectory(string path)
        {
            var entry = GetFileSystemEntry(path);
            return entry is not null && entry.IsDirectory ? entry : null;
```

Bond-009 (changes requested):

Bond-009 (approved):

crobibero (approved):

PR #11214, opened 2024-03-25
Add openapi.json upload to new repo server

joshuaboniface:
**Changes**
N/A

**Issues**
N/A

### DIFF
--- a/.github/workflows/ci-openapi.yml
+++ b/.github/workflows/ci-openapi.yml
@@ -140,9 +140,9 @@ jobs:
 
   publish:
     name: OpenAPI - Publish Unstable Spec
-    if:
-      - github.event_name != 'pull_request'
-      - contains(github.repository_owner, 'jellyfin')
+    if: |
+      github.event_name != 'pull_request' && 
+      contains(github.repository_owner, 'jellyfin')
     runs-on: ubuntu-latest
     needs:
       - openapi-head
--- a/.github/workflows/ci-openapi.yml
+++ b/.github/workflows/ci-openapi.yml
@@ -176,7 +176,7 @@ jobs:
           script: |
             TGT_DIR="/srv/repository/main/openapi"
             LAST_SPEC="$( ls -lt ${TGT_DIR} | grep 'jellyfin-openapi' | head -1 | awk '{ print $NF }' )"
-            sudo mv /srv/incoming/openapi/unstable/jellyfin-openapi-${{ env.JELLYFIN_VERSION }}.json ${TGT_DIR}/unstable/ jellyfin-openapi-${{ env.JELLYFIN_VERSION }}.json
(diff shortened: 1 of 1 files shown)


nielsvanvelzen (commented):
We already have a workflow that generates the OpenAPI specification (ci-openapi.yml) so it's probably better to re-use that one by adding an additional "publish" job to that workflow. The artifact from the `openapi-head` job can be used.

The stable specification should also be added, I guess that can be done on tag? Or do you plan to move that part into the packaging CI?

nielsvanvelzen:
You can add a condition to the deploy/publish job, example for the jellyfin.org repository: https://github.com/jellyfin/jellyfin.org/blob/master/.github/workflows/build.yml#L35
I'd also suggest adding a condition that checks for the organization, so it won't try to run on forks. Example for that: (second part of the condition) https://github.com/jellyfin/jellyfin-sdk-kotlin/blob/master/.github/workflows/sdk-publish.yaml#L15

joshuaboniface:
That should do it.

Shadowghost (approved):

.github/workflows/ci-openapi.yml
```diff
@@ -137,3 +137,54 @@ jobs:
             <!--openapi-diff-workflow-comment-->
 
             No changes to OpenAPI specification found. See history of this comment for previous changes.
+
+  publish:
+    name: OpenAPI - Publish Unstable Spec
+    if:
+      - github.event_name != 'pull_request'
+      - contains(github.repository_owner, 'jellyfin')
```
nielsvanvelzen, line 145: I don't think you are allowed to use an array here

.github/workflows/ci-openapi.yml
```diff
@@ -137,3 +137,54 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs:
+      - openapi-head
+    steps:
+      - name: Set unstable dated version
+        id: version
+        run: |-
+          echo "JELLYFIN_VERSION=$(date +'%Y%m%d%H')" >> $GITHUB_ENV
```
nielsvanvelzen, line 153: Won't this overwrite the existing spec if there are two merges on the same day?

.github/workflows/ci-openapi.yml
joshuaboniface: Aah yes you are correct, needs to be a YAML block.

.github/workflows/ci-openapi.yml
joshuaboniface: Within one hour, yes. I could make it granular to the second if we wish, I was trying to keep versioning similar to the current unstables but this might cause a problem yes.

nielsvanvelzen (commented):
This should work, one additional change I'd like is to only add the new unstable specification if it actually changed. I think this could be implemented in the last step (Move...into place) by comparing the old json with the new json. Probably need to edit out the version (in `info.version` and `info.x-jellyfin-version`) as those will always be different.

If we don't do this we will generate a lot of unnecessary SDK releases, which clutters our branches/pull requests due to all the automatons around the API spec.

joshuaboniface:
Easy fix @nielsvanvelzen, will implement.

nielsvanvelzen (approved):

PR #11220, opened 2024-03-26
Add playlist ACL endpoints

Shadowghost:
Missed adding this in the initial PR...

**Changes**
* Add API to add/remove playlist permissions
* Add API to update playlists


Jellyfin.Api/Models/PlaylistDtos/CreatePlaylistDto.cs
```diff
@@ -31,4 +32,14 @@ public class CreatePlaylistDto
     /// Gets or sets the media type.
     /// </summary>
     public MediaType? MediaType { get; set; }
+
+    /// <summary>
+    /// Gets or sets the shares.
+    /// </summary>
+    public IReadOnlyList<Share> Shares { get; set; } = [];
```
nielsvanvelzen, line 39: Use a proper DTO for Share with a descriptive name.

nielsvanvelzen (changes requested):

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -92,12 +92,136 @@ public class PlaylistsController : BaseJellyfinApiController
...
+    public IReadOnlyList<Share> GetPlaylistShares(
+        [FromRoute, Required] Guid playlistId)
+    {
+        var userId = RequestHelpers.GetUserId(User, default);
+
+        var playlist = _playlistManager.GetPlaylist(userId, playlistId);
+        var isPermitted = playlist.OwnerUserId.Equals(userId)
+            || playlist.Shares.Any(s => s.CanEdit && (s.UserId?.Equals(userId) ?? false));
+
+        return isPermitted ? playlist.Shares : new List<Share>();
```
crobibero, line 121:
```suggestion
        return isPermitted ? playlist.Shares : [];
```
or `Array.Empty<Share>`

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -92,12 +92,136 @@ public class PlaylistsController : BaseJellyfinApiController
...
+    /// Toggles OpenAccess of a playlist.
+    /// </summary>
+    /// <param name="playlistId">The playlist id.</param>
+    /// <returns>
+    /// A <see cref="Task" /> that represents the asynchronous operation to toggle OpenAccess of a playlist.
+    /// The task result contains an <see cref="OkResult"/> indicating success.
+    /// </returns>
+    [HttpPost("{playlistId}/ToggleOpenAccess")]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    public async Task<ActionResult> ToggleopenAccess(
```
nielsvanvelzen, line 134:
```suggestion
    public async Task<ActionResult> ToggleOpenAccess(
```

nielsvanvelzen, line 134: It might be better to add a more generic "UpdatePlaylist" endpoint so we can change other stuff (like the name) in the future.

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -92,12 +92,136 @@ public class PlaylistsController : BaseJellyfinApiController
...
+    /// <summary>
+    /// Remove a user from a playlist's shares.
+    /// </summary>
+    /// <param name="playlistId">The playlist id.</param>
+    /// <param name="userId">The user id.</param>
+    /// <returns>
+    /// A <see cref="Task" /> that represents the asynchronous operation to delete a user from a playlist's shares.
+    /// The task result contains an <see cref="OkResult"/> indicating success.
+    /// </returns>
+    [HttpDelete("{playlistId}/Shares")]
```
nielsvanvelzen, line 196:
```suggestion
    [HttpDelete("{playlistId}/User/{userId}")]
```

Instead of sending the userId in the body

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -92,12 +92,136 @@ public class PlaylistsController : BaseJellyfinApiController
...
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    public async Task<ActionResult> RemoveUserFromPlaylistShares(
```
nielsvanvelzen, line 198:
```suggestion
    public async Task<ActionResult> RemoveUserFromPlaylist(
```

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -92,12 +92,136 @@ public class PlaylistsController : BaseJellyfinApiController
...
+    /// </summary>
+    /// <param name="playlistId">The playlist id.</param>
+    /// <param name="shares">The shares.</param>
+    /// <returns>
+    /// A <see cref="Task" /> that represents the asynchronous operation to add shares to a playlist.
+    /// The task result contains an <see cref="OkResult"/> indicating success.
+    /// </returns>
+    [HttpPost("{playlistId}/Shares")]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    public async Task<ActionResult> AddUserToPlaylistShares(
```
nielsvanvelzen, line 164:
```suggestion
    public async Task<ActionResult> AddUserToPlaylist(
```

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -92,12 +92,136 @@ public class PlaylistsController : BaseJellyfinApiController
...
+    /// <summary>
+    /// Adds shares to a playlist's shares.
+    /// </summary>
+    /// <param name="playlistId">The playlist id.</param>
+    /// <param name="shares">The shares.</param>
+    /// <returns>
+    /// A <see cref="Task" /> that represents the asynchronous operation to add shares to a playlist.
+    /// The task result contains an <see cref="OkResult"/> indicating success.
+    /// </returns>
+    [HttpPost("{playlistId}/Shares")]
```
nielsvanvelzen, line 162:
```suggestion
    [HttpPost("{playlistId}/User/{userId}")]
```

The "canEdit" bool could be in the body, and this would effectively become an upsert

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -92,12 +92,136 @@ public class PlaylistsController : BaseJellyfinApiController
...
     }
 
+    /// <summary>
+    /// Get a playlist's shares.
+    /// </summary>
+    /// <param name="playlistId">The playlist id.</param>
+    /// <returns>
+    /// A list of <see cref="Share"/> objects.
+    /// </returns>
+    [HttpGet("{playlistId}/Shares")]
```
nielsvanvelzen, line 110:
```suggestion
    [HttpGet("{playlistId}/User")]
```

Makes more sense because I suggested that path for the add/remove operations

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -92,12 +92,136 @@ public class PlaylistsController : BaseJellyfinApiController
...
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    public IReadOnlyList<Share> GetPlaylistShares(
```
nielsvanvelzen, line 112:
```suggestion
    public IReadOnlyList<Share> GetPlaylistUsers(
```

nielsvanvelzen (commented):
After thinking about this some more I think we could replace the term "share" with "user" so it describes the relation (playlist+user).

MediaBrowser.Model/Playlists/PlaylistCreationRequest.cs
```diff
@@ -33,5 +33,10 @@ public class PlaylistCreationRequest
     /// <summary>
     /// Gets or sets the shares.
     /// </summary>
-    public Share[]? Shares { get; set; }
+    public IReadOnlyList<Share>? Shares { get; set; }
+
+    /// <summary>
+    /// Gets or sets a value indicating whether open access is enabled.
+    /// </summary>
+    public bool? OpenAccess { get; set; }
```
nielsvanvelzen, line 41: Does this just mean whether it is publicly visible or not?

```suggestion
    public bool? Public { get; set; } = true;
```

MediaBrowser.Model/Playlists/PlaylistCreationRequest.cs
```diff
@@ -33,5 +33,10 @@ public class PlaylistCreationRequest
     /// <summary>
     /// Gets or sets the shares.
     /// </summary>
-    public Share[]? Shares { get; set; }
+    public IReadOnlyList<Share>? Shares { get; set; }
```
nielsvanvelzen, line 36:
```suggestion
    public IReadOnlyList<Share>? Users { get; set; }
```

MediaBrowser.Model/Entities/UserPermissions.cs
```diff
@@ -0,0 +1,19 @@
+namespace MediaBrowser.Model.Entities;
+
+/// <summary>
+/// Class to hold data on user permissions for lists.
+/// </summary>
+/// <param name="userId">The user id.</param>
+/// <param name="canEdit">Edit permission.</param>
+public class UserPermissions(string userId, bool canEdit = false)
```
crobibero, line 8: I don't believe we want to introduce primary constructors unless it serves a real purpose

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -92,12 +92,135 @@ public class PlaylistsController : BaseJellyfinApiController
...
+    /// <param name="playlistId">The playlist id.</param>
+    /// <returns>
+    /// A list of <see cref="UserPermissions"/> objects.
+    /// </returns>
+    [HttpGet("{playlistId}/User")]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    public IReadOnlyList<UserPermissions> GetPlaylistUsers(
+        [FromRoute, Required] Guid playlistId)
+    {
+        var userId = RequestHelpers.GetUserId(User, default);
```
crobibero, line 115: This doesn't add any value over `User.GetUserId();`

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -92,12 +92,135 @@ public class PlaylistsController : BaseJellyfinApiController
...
+
+        var playlist = _playlistManager.GetPlaylist(userId, playlistId);
+        var isPermitted = playlist.OwnerUserId.Equals(userId)
+            || playlist.Shares.Any(s => s.CanEdit && s.UserId.Equals(userId));
```
crobibero, line 119: This will throw a nullref if the playlist isn't found

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -92,12 +92,135 @@ public class PlaylistsController : BaseJellyfinApiController
...
+
+        return isPermitted ? playlist.Shares : [];
```
crobibero, line 121: I think instead of returning an empty list we should just return a 401 (or 404)?

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -92,12 +92,135 @@ public class PlaylistsController : BaseJellyfinApiController
...
+            || playlist.Shares.Any(s => s.CanEdit && s.UserId.Equals(callingUserId));
+
+        if (!isPermitted)
+        {
+            return Unauthorized("Unauthorized access");
+        }
+
+        await _playlistManager.ToggleOpenAccess(playlistId, callingUserId).ConfigureAwait(false);
+
+        return NoContent();
```
crobibero, line 150: Please make sure the attribute aligns with the status codes the endpoint returns

Jellyfin.Api/Controllers/PlaylistsController.cs
thornbill: I was wondering about this also. For the frontend it probably makes sense for this all to be on the same page, so it would likely be easier to implement via a single API call.

MediaBrowser.Model/Entities/UserPermissions.cs
```diff
@@ -0,0 +1,28 @@
...
+    public UserPermissions(string userId, bool canEdit = false)
+    {
+        UserId = userId;
+        CanEdit = canEdit;
+    }
+
+    /// <summary>
+    /// Gets or sets the user id.
+    /// </summary>
+    public string UserId { get; set; }
```
nielsvanvelzen, line 22: UserId must be a Guid

MediaBrowser.Model/Entities/UserPermissions.cs
```diff
@@ -0,0 +1,28 @@
+namespace MediaBrowser.Model.Entities;
+
+/// <summary>
+/// Class to hold data on user permissions for lists.
+/// </summary>
+public class UserPermissions
```
nielsvanvelzen, line 6: Still not a fan of the name, this is to vague because we have user permissions on other objects too.

```suggestion
public class PlaylistUser
```

```suggestion
public class PlaylistUserPermissions
```

Shadowghost: I do agree, but the problem is still that the `IHasShare` interface depends on this and that one is generic.
It's currently only used by playlists but I'd rather not rename the whole thing because that would require another migration for existing playlists if we want to do it properly.

Shadowghost: Maybe collections or the idea with dynamic lists?
I've no problem with renaming and moving it and writing another migrations, jsut want to be sure this is not too big of a change for 10.9 since we are in the freeze

Jellyfin.Api/Models/PlaylistDtos/UpdatePlaylistDto.cs
```diff
@@ -0,0 +1,34 @@
+using System;
+using System.Collections.Generic;
+using System.Text.Json.Serialization;
+using Jellyfin.Extensions.Json.Converters;
+using MediaBrowser.Model.Entities;
+
+namespace Jellyfin.Api.Models.PlaylistDtos;
+
+/// <summary>
+/// Updateexisting playlist dto.
```
nielsvanvelzen, line 10:
```suggestion
/// Updat eexisting playlist dto. Fields set to `null` will not be updated and keep their current values.
```

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -92,12 +92,182 @@ public class PlaylistsController : BaseJellyfinApiController
...
+    /// <response code="401">Unauthorized access.</response>
+    /// <response code="404">Playlist not found.</response>
+    /// <returns>
+    /// A <see cref="Task" /> that represents the asynchronous operation to modify an user's playlist permissions.
+    /// The task result contains an <see cref="OkResult"/> indicating success.
+    /// </returns>
+    [HttpPost("{playlistId}/User/{userId}")]
+    [ProducesResponseType(StatusCodes.Status204NoContent)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    public async Task<ActionResult> ModifyPlaylistUserPermissions(
```
nielsvanvelzen, line 198:
```suggestion
    public async Task<ActionResult> UpdatePlaylistUser(
```

I'd also suggest using a request body (UpdatePlaylistUserDto), even though it will contain only one property for now, that way it's future proof if we add additional options.

nielsvanvelzen (changes requested):
Can you also rename the existing "addToPlaylist" and "removeFromPlaylist" to "addItemToPlaylist" and "removeItemFromPlaylist". That would make it more descriptive now that we also have removeUserFromPlaylist

nielsvanvelzen (approved):
✅ API changes

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -92,12 +92,187 @@ public class PlaylistsController : BaseJellyfinApiController
...
+    /// <returns>
+    /// A <see cref="Task" /> that represents the asynchronous operation to update a playlist.
+    /// The task result contains an <see cref="OkResult"/> indicating success.
+    /// </returns>
+    [HttpPost("{playlistId}")]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    public async Task<ActionResult> UpdatePlaylist(
+        [FromRoute, Required] Guid playlistId,
+        [FromBody(EmptyBodyBehavior = EmptyBodyBehavior.Disallow)] UpdatePlaylistDto updatePlaylistRequest)
```
crobibero, line 120: Empty body is already disallowed by default

MediaBrowser.Controller/Playlists/IPlaylistManager.cs
```diff
@@ -4,12 +4,35 @@
...
+        /// <param name="playlistId">The playlist identifier.</param>
+        /// <param name="userId">The user identifier.</param>
+        /// <returns>Playlist.</returns>
+        Playlist GetPlaylistForUser(Guid playlistId, Guid userId);
+
+        /// <summary>
+        /// Creates the playlist.
+        /// </summary>
+        /// <param name="request">The <see cref="PlaylistCreationRequest"/>.</param>
+        /// <returns>Task&lt;Playlist&gt;.</returns>
```
crobibero, line 26:
```suggestion
        /// <returns>The created playlist.</returns>
```

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -92,12 +92,187 @@ public class PlaylistsController : BaseJellyfinApiController
...
+        {
+            return NotFound("Playlist not found");
+        }
+
+        var isPermitted = playlist.OwnerUserId.Equals(callingUserId)
+            || playlist.Shares.Any(s => s.CanEdit && s.UserId.Equals(callingUserId));
+
+        if (!isPermitted)
+        {
+            return Unauthorized("Unauthorized access");
```
crobibero, line 135: Should be forbidden (and would drop the message)

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -92,12 +92,187 @@ public class PlaylistsController : BaseJellyfinApiController
...
+        var playlist = _playlistManager.GetPlaylistForUser(playlistId, userId);
+        if (playlist is null)
+        {
+            return NotFound("Playlist not found");
+        }
+
+        var isPermitted = playlist.OwnerUserId.Equals(userId)
+            || playlist.Shares.Any(s => s.CanEdit && s.UserId.Equals(userId));
+
+        return isPermitted ? playlist.Shares.ToList() : Unauthorized("Unauthorized Access");
```
crobibero, line 179: Forbidden

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -92,29 +92,256 @@ public class PlaylistsController : BaseJellyfinApiController
...
+    {
+        var callingUserId = User.GetUserId();
+
+        var playlist = _playlistManager.GetPlaylistForUser(playlistId, callingUserId);
+        if (playlist is null)
+        {
+            return NotFound("Playlist not found");
+        }
+
+        var isPermitted = playlist.OwnerUserId.Equals(userId)
```
nielsvanvelzen, line 211: Shouldn't this be

```suggestion
        var isPermitted = playlist.OwnerUserId.Equals(callingUserId)
```

nielsvanvelzen (changes requested):

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -92,29 +92,259 @@ public class PlaylistsController : BaseJellyfinApiController
...
+    /// Get a playlist's users.
+    /// </summary>
+    /// <param name="playlistId">The playlist id.</param>
+    /// <response code="200">Found shares.</response>
+    /// <response code="403">Access forbidden.</response>
+    /// <response code="404">Playlist not found.</response>
+    /// <returns>
+    /// A list of <see cref="PlaylistUserPermissions"/> objects.
+    /// </returns>
+    [HttpGet("{playlistId}/User")]
```
crobibero, line 162:
```suggestion
    [HttpGet("{playlistId}/Users")]
```

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -92,29 +92,259 @@ public class PlaylistsController : BaseJellyfinApiController
...
+    /// <param name="playlistId">The playlist id.</param>
+    /// <param name="userId">The user id.</param>
+    /// <response code="200">User permission found.</response>
+    /// <response code="200">No user permission found but open access.</response>
+    /// <response code="403">Access forbidden.</response>
+    /// <response code="404">Playlist not found.</response>
+    /// <returns>
+    /// <see cref="PlaylistUserPermissions"/>.
+    /// </returns>
+    [HttpGet("{playlistId}/User/{userId}")]
```
crobibero, line 194:
```suggestion
    [HttpGet("{playlistId}/Users/{userId}")]
```

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -92,29 +92,259 @@ public class PlaylistsController : BaseJellyfinApiController
...
+    /// <param name="userId">The user id.</param>
+    /// <param name="updatePlaylistUserRequest">The <see cref="UpdatePlaylistUserDto"/>.</param>
+    /// <response code="204">User's permissions modified.</response>
+    /// <response code="403">Access forbidden.</response>
+    /// <response code="404">Playlist not found.</response>
+    /// <returns>
+    /// A <see cref="Task" /> that represents the asynchronous operation to modify an user's playlist permissions.
+    /// The task result contains an <see cref="OkResult"/> indicating success.
+    /// </returns>
+    [HttpPost("{playlistId}/User/{userId}")]
```
crobibero, line 232:
```suggestion
    [HttpPost("{playlistId}/Users/{userId}")]
```

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -92,29 +92,259 @@ public class PlaylistsController : BaseJellyfinApiController
...
+    /// <param name="playlistId">The playlist id.</param>
+    /// <param name="userId">The user id.</param>
+    /// <response code="204">User permissions removed from playlist.</response>
+    /// <response code="401">Unauthorized access.</response>
+    /// <response code="404">No playlist or user permissions found.</response>
+    /// <returns>
+    /// A <see cref="Task" /> that represents the asynchronous operation to delete a user from a playlist's shares.
+    /// The task result contains an <see cref="OkResult"/> indicating success.
+    /// </returns>
+    [HttpDelete("{playlistId}/User/{userId}")]
```
crobibero, line 278:
```suggestion
    [HttpDelete("{playlistId}/Users/{userId}")]
```

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -92,29 +92,259 @@ public class PlaylistsController : BaseJellyfinApiController
...
+    /// <param name="updatePlaylistRequest">The <see cref="UpdatePlaylistDto"/> id.</param>
+    /// <response code="204">Playlist updated.</response>
+    /// <response code="403">Access forbidden.</response>
+    /// <response code="404">Playlist not found.</response>
+    /// <returns>
+    /// A <see cref="Task" /> that represents the asynchronous operation to update a playlist.
+    /// The task result contains an <see cref="OkResult"/> indicating success.
+    /// </returns>
+    [HttpPost("{playlistId}")]
+    [ProducesResponseType(StatusCodes.Status200OK)]
```
nielsvanvelzen, line 116: 204

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -92,29 +92,259 @@ public class PlaylistsController : BaseJellyfinApiController
...
+        {
+            return NotFound("Playlist not found");
+        }
+
+        var userPermission = playlist.Shares.FirstOrDefault(s => s.UserId.Equals(userId));
+        var isPermitted = playlist.OwnerUserId.Equals(callingUserId)
+            || playlist.Shares.Any(s => s.CanEdit && s.UserId.Equals(callingUserId))
+            || userId.Equals(callingUserId);
+
+        return isPermitted ? userPermission is not null ? userPermission : NotFound("User permissions not found") : playlist.OpenAccess ? NoContent() : Forbid();
```
nielsvanvelzen, line 216: this is messy, just use normal if-else statements.

nielsvanvelzen, line 216: I'm not sure about the "public playlist = 204, otherwise 404" response behavior. I think it is the client responsibility to deal with that, and no specific permissions for a user means it should just 404.

nielsvanvelzen (changes requested):

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -92,29 +92,260 @@ public class PlaylistsController : BaseJellyfinApiController
...
+            return NotFound("Playlist not found");
+        }
+
+        var isPermitted = playlist.OwnerUserId.Equals(userId);
+
+        return isPermitted ? playlist.Shares.ToList() : Forbid();
+    }
+
+    /// <summary>
+    /// Get a playlist users.
```
nielsvanvelzen, line 183: ``
```suggestion
    /// Get a playlist user.
```

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -92,29 +92,260 @@ public class PlaylistsController : BaseJellyfinApiController
...
+        {
+            return NotFound("Playlist not found");
+        }
+
+        var userPermission = playlist.Shares.FirstOrDefault(s => s.UserId.Equals(userId));
+        var isPermitted = playlist.OwnerUserId.Equals(callingUserId)
+            || playlist.Shares.Any(s => s.CanEdit && s.UserId.Equals(callingUserId))
+            || userId.Equals(callingUserId);
+
+        if (isPermitted && userPermission is not null)
```
nielsvanvelzen, line 212: The other endpoints already `Forbid()` when not permitted so we could mirror that behavior here.

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -92,29 +92,260 @@ public class PlaylistsController : BaseJellyfinApiController
...
+    /// The task result contains an <see cref="OkResult"/> indicating success.
+    /// </returns>
+    [HttpPost("{playlistId}/Users/{userId}")]
+    [ProducesResponseType(StatusCodes.Status204NoContent)]
+    [ProducesResponseType(StatusCodes.Status403Forbidden)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    public async Task<ActionResult> UpdatePlaylistUser(
+        [FromRoute, Required] Guid playlistId,
+        [FromRoute, Required] Guid userId,
+        [FromBody(EmptyBodyBehavior = EmptyBodyBehavior.Allow)] UpdatePlaylistUserDto updatePlaylistUserRequest)
```
nielsvanvelzen, line 240: Shouldn't this be required?

Jellyfin.Api/Models/PlaylistDtos/CreatePlaylistDto.cs
```diff
@@ -31,4 +32,14 @@ public class CreatePlaylistDto
...
+
+    /// <summary>
+    /// Gets or sets the playlist users.
+    /// </summary>
+    public IReadOnlyList<PlaylistUserPermissions> Users { get; set; } = [];
+
+    /// <summary>
+    /// Gets or sets a value indicating whether the playlist is public.
+    /// </summary>
+    public bool Public { get; set; } = true;
```
nielsvanvelzen, line 44: I should've caught this one earlier, but we might want to rename this to "IsPublic because most of our other models do the same (look at BaseItemDto for example, with isLive/isMovie/isFolder etc.).

Personally I'd prefer just "Public" but consistency might be the better way to go.

nielsvanvelzen: (same for the UpdatePlaylistDto model)

nielsvanvelzen (changes requested):

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -180,17 +180,19 @@ public class PlaylistsController : BaseJellyfinApiController
     }
 
     /// <summary>
-    /// Get a playlist users.
+    /// Get a playlist user.
     /// </summary>
     /// <param name="playlistId">The playlist id.</param>
     /// <param name="userId">The user id.</param>
     /// <response code="200">User permission found.</response>
+    /// <response code="200">Access forbidden.</response>
```
nielsvanvelzen, line 188: wrong code

crobibero (approved):

nielsvanvelzen (approved):
✅ API changes

PR #11225, opened 2024-03-26
fix: disable api endpoint for ffmpeg path for security

gnattu:
**Changes**

This PR ports the security fix for [GHSA-rr9h-w522-cvmr](https://github.com/jellyfin/jellyfin/security/advisories/GHSA-rr9h-w522-cvmr) and disabled the ffmpeg path setting API.

**Issues**


Shadowghost (approved):

crobibero (approved):

Bond-009 (approved):

nielsvanvelzen:
Alternatively to this change, we could also remove that entire endpoint as we'll never make it work again.

nielsvanvelzen (approved):

PR #11227, opened 2024-03-26
Fix incorrect artwork displayed when playing audio

### LINKED ISSUE #8570: [Issue]:Not useing correct music cover when there is mutiple covers used in an album
### Please describe your bug

When I Add an album with multiple covers used in different songs, jellyfin picks a random cover in the song metadata and makes all songs that cover artwork. 

for example, I have 6 disks in an album, and each disk has a different cover used, in the metadata, they are all correct with 6 other covers associated with the 6 disks. jellyfin will only pick 1 cover and assign all songs to have that selected cover.

I have tried to use different metadata fetchers MusicBrainz and TheAudioDB but the issue still persists.

I have noticed that if you give the album a cover then jellyfin will assign all songs that the album artwork ignoring the actual metadata and using the wrong image when you play the song.

file structure
AllMusic
---Half-Life Alyx Soundtrack (Artist)
        ---Half-Life Alyx Soundtrack (Album)
                --Disk1
                --Disk2
                ...
                --....Disk11the 
                
                cover.jpeg

### Jellyfin Version

10.8.0

### if other:

_No response_

### Environment

```markdown
- OS: Windows
- Virtualization:
- Clients:browser, UWP app
- Browser: Firefox
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- Plugins: none
- Reverse Proxy: none
- Base URL: none
- Networking:
- Storage: Local
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

Metadata cover image

Here you can see it's playing the song "Only One" in the Half-life Alyx soundtrack.
I'm showing that windows' default music player is doing it correctly (Yellow-green image), showing the correct image in the metadata. 
I'm also using mp3 tag to display the actual metadata for this song proving that jellyfin has the wrong metadata for this song (Blueish image that's for disk 10, not disk 2)

This album has no album artwork I set up it and found the metadata using the Image Extractor, the other methods like MuisicBrainz are turned off for this example

Screenshot 2022-10-18 053803

when I added a cover artwork in the files and refreshed the metadata the currently playing song metadata was still wrong

Screenshot 2022-10-18 054230
File structure of my songs

### Code of Conduct

MisterGenius0:
**Changes**

During media scan, images will be compared by size + album and will only cache unique images

Added: MisterGenius0 to COMTRIBUTORS.md

**Issues**

Fixes: #8570


MediaBrowser.Providers/MediaInfo/AudioImageProvider.cs
```diff
@@ -108,21 +107,19 @@ private async Task<DynamicImageResponse> GetImage(Audio item, List<MediaStream>
             };
         }
 
-        private string GetAudioImagePath(Audio item)
+        private string GetAudioImagePath(Audio item, string newImagePath)
```
MisterGenius0, line 110: Get the incoming image that was scanned, Look at the file length+Album and use that for the unique identifier

Shadowghost (commented):
Will this re-extract all images and re-save them? What about cleaning up existing images?

MisterGenius0:
This will only extract unique images to AppData\Local\jellyfin\cache\extracted-audio-images (Windows) when you do a library scan,

if you had images before you will have to re-scan the library for it to grab the new images.

The images created in the Temp folder \AppData\Local\jellyfin\cache\temp will get deleted after the scan on lines 95-100.

if you have existing images from before, it will not clear the cache folder (the code before didn't clear it either),
The images are only cleared when you clear the cache (not sure if jellyfin ever clears the cache on its own somewhere else in the codebase)

Before, I did notice it did extract duplicate images, it would extract duplicate images before if it was the same image but in the same album but with a different artist.

Shadowghost:
This would mean we do re-extract all images and have duplicates. I don't think this is cleaned-up anywhere else, so we don't really change behaviour except creating the duplicates. Ideally we'd do a cleanup though.

MisterGenius0:
Do you want the cleanup added to this PR or this should be another issue/PR?
The current version of Jelly-fin doesn't empty the metadata folder.
What would be the criteria for deleting "old" files?

Since we can't just delete the entire metadata folder when scanning a library, that might remove some photos that were saved from another library scan.

sparklecatdev (approved):

TT-cyber-sys (approved):

TT-cyber-sys:
merge when?

PR #11231, opened 2024-03-27
Move rename command permission check to script

felix920506:
Removes the permission check from the GHA and let the script handle it instead.
This command was intended to allow triage team members to rename issue, but the GHA permission check makes it so that it can only be ran by people who already have rename permissions.

The change to the script:
https://github.com/jellyfin/jellyfin-triage-scripts/pull/14/

**Issues**
None

### DIFF
--- a/.github/workflows/commands.yml
+++ b/.github/workflows/commands.yml
@@ -124,7 +124,7 @@ jobs:
 
   rename:
     name: Rename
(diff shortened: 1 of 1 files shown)


Shadowghost (approved):

joshuaboniface (approved):
Reviewed both the script change and this one, and seems OK as long as it works!

PR #11232, opened 2024-03-27
Fix default values missing in ClientCapabilities(Dto)

nielsvanvelzen:
We emit the DefaultValue in OpenAPI spec so a 10.9 based SDK can automatically fill in a default, but we didn't actually send a default value causing a 10.8 based SDK to fail.

**Changes**

Updated the ClientCapabilities to send default values for SupportsContentUploading and SupportsSync to fix sign in in the ATV app.

**Issues**

### DIFF
--- a/MediaBrowser.Model/Session/ClientCapabilities.cs
+++ b/MediaBrowser.Model/Session/ClientCapabilities.cs
@@ -35,11 +35,11 @@ namespace MediaBrowser.Model.Session
         // TODO: Remove after 10.9
         [Obsolete("Unused")]
         [DefaultValue(false)]
-        public bool? SupportsContentUploading { get; set; }
+        public bool? SupportsContentUploading { get; set; } = false;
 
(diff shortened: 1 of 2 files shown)


MediaBrowser.Model/Session/ClientCapabilities.cs
```diff
@@ -35,11 +35,12 @@ public ClientCapabilities()
...
         [Obsolete("Unused")]
         [DefaultValue(false)]
-        public bool? SupportsContentUploading { get; set; }
+        public bool? SupportsContentUploading { get; set; } = false;
 
         // TODO: Remove after 10.9
         [Obsolete("Unused")]
         [DefaultValue(false)]
-        public bool? SupportsSync { get; set; }
+
```
dmitrylyzo, line 43:
```suggestion
```

crobibero (approved):

Shadowghost (approved):

PR #11235, opened 2024-03-28
 fix: don't do empty hwupload for VT

gnattu:
**Changes**

A hwupload without any filter to actually consume the frame would cause VideoToolbox's encoder to produce frames with artifacts. Check if the filter chain actually has a valid filter before adding hwupload.

**Issues**

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -5096,14 +5096,6 @@ namespace MediaBrowser.Controller.MediaEncoding
                 return (null, null, null);
             }
 
-            // INPUT videotoolbox/memory surface(vram/uma)
-            // this will pass-through automatically if in/out format matches.
-            var hwuploadFilters = new List<string>
-            {
-                "format=nv12|p010le|videotoolbox_vld",
-                "hwupload=derive_device=videotoolbox"
-            };
-
             /* Make main filters for video stream */
             var mainFilters = new List<string>();
 
@@ -5171,8 +5163,10 @@ namespace MediaBrowser.Controller.MediaEncoding
             // will cause the encoder to produce incorrect frames.
             if (needFiltering)
             {
-                hwuploadFilters.AddRange(mainFilters);
-                mainFilters = hwuploadFilters;
+                // INPUT videotoolbox/memory surface(vram/uma)
+                // this will pass-through automatically if in/out format matches.
+                mainFilters.Insert(0, "format=nv12|p010le|videotoolbox_vld");
+                mainFilters.Insert(0, "hwupload=derive_device=videotoolbox");
             }
 
             return (mainFilters, subFilters, overlayFilters);
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -5096,13 +5096,16 @@ namespace MediaBrowser.Controller.MediaEncoding
                 return (null, null, null);
             }
 
-            /* Make main filters for video stream */
-            var mainFilters = new List<string>();
-
             // INPUT videotoolbox/memory surface(vram/uma)
             // this will pass-through automatically if in/out format matches.
-            mainFilters.Add("format=nv12|p010le|videotoolbox_vld");
-            mainFilters.Add("hwupload=derive_device=videotoolbox");
+            var hwuploadFilters = new List<string>
+            {
+                "format=nv12|p010le|videotoolbox_vld",
+                "hwupload=derive_device=videotoolbox"
+            };
+
+            /* Make main filters for video stream */
+            var mainFilters = new List<string>();
 
             // hw deint
             if (doDeintH2645)
@@ -5157,6 +5160,16 @@ namespace MediaBrowser.Controller.MediaEncoding
 
                 subFilters.Add("hwupload=derive_device=videotoolbox");
                 overlayFilters.Add("overlay_videotoolbox=eof_action=pass:repeatlast=0");
+            }
+
+            var needFiltering = mainFilters.Any(f => !string.IsNullOrEmpty(f)) ||
+                                subFilters.Any(f => !string.IsNullOrEmpty(f)) ||
+                                overlayFilters.Any(f => !string.IsNullOrEmpty(f));
+
+            if (needFiltering)
+            {
+                hwuploadFilters.AddRange(mainFilters);
+                mainFilters = hwuploadFilters;
             }
 
             return (mainFilters, subFilters, overlayFilters);
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -5166,6 +5166,9 @@ namespace MediaBrowser.Controller.MediaEncoding
                                 subFilters.Any(f => !string.IsNullOrEmpty(f)) ||
                                 overlayFilters.Any(f => !string.IsNullOrEmpty(f));
 
+            // This is a workaround for ffmpeg's hwupload implementation
+            // For VideoToolbox encoders, a hwupload without a valid filter actually consuming its frame
+            // will cause the encoder to produce incorrect frames.
             if (needFiltering)
             {
                 hwuploadFilters.AddRange(mainFilters);


Shadowghost (approved):

nyanmisaka (approved):
Another de-sync issue similar to when videotoolbox encodes BGRA?

Both of these functions should block until the target data is well prepared.
- hwcontext_videotoolbox.c -> [vt_transfer_data_to()](https://github.com/FFmpeg/FFmpeg/blob/b9297128f5f66bb9e6d2e755e4b82818bc1ada32/libavutil/hwcontext_videotoolbox.c#L696)
- videotoolboxenc.c -> [create_cv_pixel_buffer()](https://github.com/FFmpeg/FFmpeg/blob/b9297128f5f66bb9e6d2e755e4b82818bc1ada32/libavcodec/videotoolboxenc.c#L2450C12-L2450C34)

This is sufficient for the current use case. Best to fix it in FFmpeg in the future.

gnattu:
> Another de-sync issue similar to when videotoolbox encodes BGRA?
> 
> Both of these functions should block until the target data is well prepared.
> 
> * hwcontext_videotoolbox.c -> [vt_transfer_data_to()](https://github.com/FFmpeg/FFmpeg/blob/b9297128f5f66bb9e6d2e755e4b82818bc1ada32/libavutil/hwcontext_videotoolbox.c#L696)
> * videotoolboxenc.c -> [create_cv_pixel_buffer()](https://github.com/FFmpeg/FFmpeg/blob/b9297128f5f66bb9e6d2e755e4b82818bc1ada32/libavcodec/videotoolboxenc.c#L2450C12-L2450C34)
> 
> This is sufficient for the current use case. Best to fix it in FFmpeg in the future.

The BGRA case is an upstream implementation issue. The upstream will initialize HEVC encoder with `kCMVideoCodecType_HEVCWithAlpha` when feed with BGRA and that encoder does not handle frame reordering properly, but `kVTCompressionPropertyKey_AllowFrameReordering` is still set to true.

This `hwhpload` one is harder to debug and I have not noticed any problematic code so far. It is a different kind of problem as the frame order is kind of OK but you will see the frames been corrupted like two or more frames are blend together.

nyanmisaka:
> The BGRA case is an upstream implementation issue. The upstream will initialize HEVC encoder with kCMVideoCodecType_HEVCWithAlpha when feed with BGRA and that encoder does not handle frame reordering properly, but kVTCompressionPropertyKey_AllowFrameReordering is still set to true.

Is it allowed to use `kCMVideoCodecType_HEVC` with BGRA? If so, just add BGR0 to bypass encoding alpha channel.

> This hwhpload one is harder to debug and I have not noticed any problematic code so far. Just checked with more files. It is indeed a frame ordering issue. It seems like b-frame related settings is very broken.

So is disabling b-frames *fixing* it?

I haven't looked at `videotoolboxenc.c` carefully, you might know if it supports async encoding? That is, queue some frames in videotoolboxenc.c or VT runtime, and use callbacks to release these in-queue/reference frames and retrieve the encoded h26x data, which may cause artifacts if these refs frames are released prematurely.

gnattu:
> So is disabling b-frames _fixing_ it?

No.

> I haven't looked at `videotoolboxenc.c` carefully, you might know if it supports async encoding? That is, queue some frames in videotoolboxenc.c or VT runtime, and use callbacks to release these in-queue/reference frames and retrieve the encoded h26x data, which may cause artifacts if these refs frames are released prematurely.

I set `max_ref_frames` to 0 and it does not fix the problem.

nyanmisaka:
> Just look at the code further, when using `hwupload`, the `videotoolboxenc` will use the `hwcontext` managed `CVPixelBufferPool`, when used without, `videotoolboxenc` will create its own `CVPixelBufferPool`. Probably this is where things start to break.

But this still does not explain why the hw frames copied by `scale_vt` are not affected.

nyanmisaka:
Apart from this,
```
// hwupload path:
CVPixelBufferPoolCreate() -> CVPixelBufferPoolCreatePixelBuffer() -> Lock -> copy -> Unlock

// encoder internal path:
VTCompressionSessionGetPixelBufferPool() -> CVPixelBufferPoolCreatePixelBuffer() -> Lock -> copy -> Unlock
```

Another difference is that the encoder internal copy path [does not use](https://github.com/FFmpeg/FFmpeg/blob/b9297128f5f66bb9e6d2e755e4b82818bc1ada32/libavcodec/videotoolboxenc.c#L2384-L2413) `av_frame_copy`.

gnattu:
> But this still does not explain why the hw frames copied by `scale_vt` are not affected.

Yes, it's indeed quite strange. If I use my brand new `mjpeg_videotoolbox`, then this won't happen, but it does occur with `h264_videotoolbox`, `hevc_videotoolbox`, and `prores_videotoolbox`. Very hard to debug, as the code look correct everywhere.

gnattu:
> Apart from this,
> 
> ```
> // hwupload path:
> CVPixelBufferPoolCreate() -> CVPixelBufferPoolCreatePixelBuffer() -> Lock -> copy -> Unlock
> 
> // encoder internal path:
> VTCompressionSessionGetPixelBufferPool() -> CVPixelBufferPoolCreatePixelBuffer() -> Lock -> copy -> Unlock
> ```
> 
> Another difference is that the encoder internal copy path [does not use](https://github.com/FFmpeg/FFmpeg/blob/b9297128f5f66bb9e6d2e755e4b82818bc1ada32/libavcodec/videotoolboxenc.c#L2384-L2413) `av_frame_copy`.

If this is the case then this is unfixable unless we make a not upstream-able `hwupload_videotoolbox`. I would prefer just don't hwupload without filters instead.

nyanmisaka:
> If this is the case then this is unfixable unless we make a not upstream-able `hwupload_videotoolbox`. I would prefer just don't hwupload without filters instead.

If it's just de-sync and not distorted luma/chroma, it should have nothing to do with how to copy planes.

gnattu:
> > If this is the case then this is unfixable unless we make a not upstream-able `hwupload_videotoolbox`. I would prefer just don't hwupload without filters instead.
> 
> If it's just de-sync and not distorted luma/chroma, it should have nothing to do with how to copy planes.

I'm not very sure:

https://github.com/jellyfin/jellyfin/assets/12995396/cdc40d51-30db-471e-a4df-1aca2727a481

It seems like desync, but if you pause you can see some frames been distorted.

nyanmisaka:
This is still [tearing](https://en.wikipedia.org/wiki/Screen_tearing) caused by desync, no twisted chroma.

image

gnattu:
Reported to upstream: https://trac.ffmpeg.org/ticket/10939

Let's don't think about it too much for now. There is no performance diff, and this workaround is not that ugly.

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -5159,6 +5162,19 @@ public string GetHwTonemapFilter(EncodingOptions options, string hwTonemapSuffix
...
+                                subFilters.Any(f => !string.IsNullOrEmpty(f)) ||
+                                overlayFilters.Any(f => !string.IsNullOrEmpty(f));
+
+            // This is a workaround for ffmpeg's hwupload implementation
+            // For VideoToolbox encoders, a hwupload without a valid filter actually consuming its frame
+            // will cause the encoder to produce incorrect frames.
+            if (needFiltering)
+            {
+                hwuploadFilters.AddRange(mainFilters);
+                mainFilters = hwuploadFilters;
```
crobibero, line 5175:
```suggestion
                mainFilters.Insert(0, "hwupload=derive_device=videotoolbox");
                mainFilters.Insert(0, "format=nv12|p010le|videotoolbox_vld");
```

If order doesn't matter just use `Add()`, and remove `hwuploadFilters`

nyanmisaka: It matters. `hwupload*` should be the first.

gnattu: Yes, hwupload has to be first.

crobibero: Ok, my suggestion is to not create a list just to maybe use it- but to just insert the required filters at the start of the list when needed

gnattu: Done.

crobibero (approved):

Bond-009 (approved):

PR #11246, opened 2024-03-29
fix: unset qmin and qmax for vt

gnattu:
**Changes**

We are using bitrate-based control, but VT will target qmax first, potentially making the video bitrate much higher than the limit when the limit is very low and the video resolution is very high. Unset it so that the bitrate is close to the target.

In this case, even the `-maxrate` setting will have no effect on the bitrate. It only reduces the performance, but the result bitrate is still much higher than the limit. 

Other encoders potentially have same issue, but most of them is already patched in some other way, like very high qmax or using CBR.

Ideally, we should perform a video resolution downscale when the bitrate limit is too low for the resolution to preserve better visual quality. However, this use case could be very edgey, such as attempting to stream 4K60 video over a 5Mbps network, so probably not worth the effort.

**Issues**

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -1402,7 +1402,7 @@ namespace MediaBrowser.Controller.MediaEncoding
             {
                 // The `maxrate` and `bufsize` options can potentially lead to performance regression
                 // and even encoder hangs, especially when the value is very high.
-                return FormattableString.Invariant($" -b:v {bitrate}  -qmin -1  -qmax -1");
+                return FormattableString.Invariant($" -b:v {bitrate} -qmin -1 -qmax -1");
             }
 
             return FormattableString.Invariant($" -b:v {bitrate} -maxrate {bitrate} -bufsize {bufsize}");
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -1899,6 +1899,9 @@ namespace MediaBrowser.Controller.MediaEncoding
                         param += " -prio_speed 1";
                         break;
                 }
+
+                param += " -qmin -1";
+                param += " -qmax -1";
             }
             else if (string.Equals(videoEncoder, "libvpx", StringComparison.OrdinalIgnoreCase)) // vp8
             {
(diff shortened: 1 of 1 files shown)


Bond-009 (approved):

Shadowghost (approved):

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -1899,6 +1899,9 @@ public string GetVideoQualityParam(EncodingJobInfo state, string videoEncoder, E
                         param += " -prio_speed 1";
                         break;
                 }
+
+                param += " -qmin -1";
+                param += " -qmax -1";
```
nyanmisaka, line 1904: How about moving these to `GetVideoBitrateParam()` ? They are highly correlated.
https://github.com/jellyfin/jellyfin/blob/a49f552d6ec81c68a5d11f4ce585f824bf5f3150/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs#L1400-L1406

gnattu: Other encoders are setting the q value here so I chose to do the same as others. If move this to `GetVideoBitrateParam()` is really much better I could do that.

nyanmisaka: https://github.com/jellyfin/jellyfin/blob/a49f552d6ec81c68a5d11f4ce585f824bf5f3150/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs#L1991

`GetVideoBitrateParam()` is called in `GetVideoQualityParam()`.

In fact, we don't expect them to be called separately, resulting in video quality being partially controlled. It makes more sense to change `GetVideoBitrateParam()` to private.

gnattu: Done.

crobibero (approved):

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -1402,7 +1402,7 @@ public string GetVideoBitrateParam(EncodingJobInfo state, string videoCodec)
             {
                 // The `maxrate` and `bufsize` options can potentially lead to performance regression
                 // and even encoder hangs, especially when the value is very high.
-                return FormattableString.Invariant($" -b:v {bitrate}");
+                return FormattableString.Invariant($" -b:v {bitrate}  -qmin -1  -qmax -1");
```
nyanmisaka, line 1405: Redundant spaces.
```suggestion
                return FormattableString.Invariant($" -b:v {bitrate} -qmin -1 -qmax -1");
```

nyanmisaka (approved):

PR #11249, opened 2024-03-29
Support "extra" folder for extras content

nielsvanvelzen:
In my library structure I've always used the folder "extra" for extra content, when I migrated to Jellyfin (years ago) I never imported all this media. Figured I should do something about that so I can develop support for theme songs/video in the ATV app. 

**Changes**
- Add an extra rule for the "extra" folder

**Issues**

### DIFF
--- a/Emby.Naming/Common/NamingOptions.cs
+++ b/Emby.Naming/Common/NamingOptions.cs
@@ -540,6 +540,12 @@ namespace Emby.Naming.Common
                 new ExtraRule(
                     ExtraType.Unknown,
                     ExtraRuleType.DirectoryName,
+                    "extra",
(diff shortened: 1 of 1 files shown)


crobibero (approved):

Bond-009 (approved):

Shadowghost (approved):

dathbe:
How is this different from the way Jellyfin handled "extras" folders previously?  is it just adding another name option to the same type of folder, or does it do something different?  I have several movies with an "extras" folder already, containing things like trailers and behind-the-scenes, and it seems do show up in Jellyfin.

nielsvanvelzen:
It just added an extra folder name to the list

Issue #8687, opened 2022-11-03
[Issue]: Server Crash "An invalid IP address was specified. (Parameter 'address')"


shortN0te:
### Please describe your bug

Jellyfin version: 10.8.7
Installed on an Debian host with docker managed by Portainer.
docker-compose.yml:
version: "3.5"
services:
  jellyfin:
    image: jellyfin/jellyfin
    container_name: jellyfin
    user: 1000:1000
    network_mode: "host"
    volumes:
      - /opt/docker/jellyfin/media:/media
      - /opt/docker/jellyfin/config:/config
      - /opt/docker/jellyfin/cache:/cache
      - /expanse/Media/Movies:/mediaGERMovies:ro
      - /expanse/Media/Movies-English:/mediaENMovies:ro
      - /expanse/Media/Series:/mediaShows:ro
      - /expanse/Media/Music:/mediaMusic:ro
      - /expanse/Media/OST:/mediaOST:ro
      - /expanse/Media/Youtube:/mediaYT:ro
    restart: "no"
   environment:
      - JELLYFIN_PublishedServerUrl=https://removed
 
Disabled all Plugins and disabled DLNA Server. This happens randomly when clients are connected or not and the time frame between those crashes is random as well. Extended Log here: https://pastebin.com/AUYhU9T8 but not much happening.

It coincided with the 10.8.6 release, the setup was running fine for months before that.

### Jellyfin Version

Other

### if other:

10.8.7

### Environment

```markdown
- OS:Debian
- Virtualization:Docker
- Clients:Jellyfin Media Player, Web Client
- Plugins: TheTVDB, Webhook, YoutubeMetadata
- Reverse Proxy: nginx
- Networking: Host
```

### Jellyfin logs

```shell
[15:22:16] [FTL] [26] Main: Unhandled Exception

System.ArgumentException: An invalid IP address was specified. (Parameter 'address')

   at System.Net.IPAddress..ctor(ReadOnlySpan`1 address)

   at System.Net.NetworkInformation.LinuxNetworkInterface.GetLinuxNetworkInterfaces()

   at System.Net.NetworkInformation.NetworkInterfacePal.GetIsNetworkAvailable()

   at System.Net.NetworkInformation.NetworkChange.OnAvailabilityTimerFired(Object state)

   at System.Threading.TimerQueueTimer.Fire(Boolean isThreadPool)

   at System.Threading.TimerQueue.FireNextTimers()

[...]

PR #11253, opened 2024-03-30
cli: add option to disable network change detection

gnattu:
**Changes**

The NetworkChange API implemented in dotnet is problematic on a lot of platforms. It does not reliably report networking changes and can even cause unhandled exceptions to fail the main thread in extreme cases. Add an option to disable it to allow users to work around the issues caused by this API.

**Issues**

Potentially workarounds #8687, need user testing.

### DIFF
--- a/src/Jellyfin.Networking/Manager/NetworkManager.cs
+++ b/src/Jellyfin.Networking/Manager/NetworkManager.cs
@@ -97,7 +97,7 @@ public class NetworkManager : INetworkManager, IDisposable
         _networkEventLock = new object();
         _remoteAddressFilter = new List<IPNetwork>();
 
-        var detectNetworkChange = startupConfig[DetectNetworkChangeKey] == bool.TrueString;
+        _ = bool.TryParse(startupConfig[DetectNetworkChangeKey], out var detectNetworkChange);
 
         UpdateSettings(_configurationManager.GetNetworkConfiguration());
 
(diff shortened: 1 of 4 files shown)


gnattu:
I'm not sure if this is still good enough for 10.9. If it does not I will turn this into a draft.

Shadowghost (approved):

Bond-009:
Personally I think the env var is enough, this doesn't have to be a CLI arg

gnattu:
> @gnattu Do you have any upstream references for the broken behaviour for future reference?

Well it is from the implementation, they are not properly caching the exceptions for the unsafe get interface method and the exception is just thrown to the main thread, and in the worst case it crashed the server for some users. If we want to implement things like this, we'd better to use our own timer instead of the broken upstream ones.

gnattu:
Rebased to the latest master as this still seems necessary since the network change API remains unreliable in dotnet 8.

src/Jellyfin.Networking/Manager/NetworkManager.cs
```diff
@@ -97,10 +97,15 @@ public NetworkManager(IConfigurationManager configurationManager, IConfiguration
         _networkEventLock = new object();
         _remoteAddressFilter = new List<IPNetwork>();
 
+        var detectNetworkChange = startupConfig[DetectNetworkChangeKey] == bool.TrueString;
```
crobibero, line 100: Should use `bool.TryParse` here (or at least use case-insensitive since the value is "True"`)

crobibero (approved):

PR #11261, opened 2024-03-30
Use enum for BaseItemDto.ExtraType

nielsvanvelzen:
Easy change, too late for 10.9 though :p

**Changes**
- Use enum for BaseItemDto.ExtraType

**Issues**

### DIFF
--- a/Emby.Server.Implementations/Dto/DtoService.cs
+++ b/Emby.Server.Implementations/Dto/DtoService.cs
@@ -903,10 +903,7 @@ namespace Emby.Server.Implementations.Dto
             if (item is Audio audio)
             {
                 dto.Album = audio.Album;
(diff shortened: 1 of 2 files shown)


Bond-009 (approved):

crobibero (approved):
I'm fine with making the API change for this since we still need the playlist endpoints

Shadowghost (approved):

nielsvanvelzen:
> I'm fine with making the API change for this since we still need the playlist endpoints

It's probably fine to get into 10.9, I don't expect it to break anything. But as we've seen with moving stuff over to enums: clients do weird stuff with these values and case sensitivity may break some of them. So if we merge this and we get reports of clients breaking because of it, we should revert it for 10.9.

PR #11270, opened 2024-03-31
Notifications: use original title if req.Name is null

knocte:
**Changes**
Fixes problem about notifications receiving only the description of the movie instead of the title.

**Issues**
Related: https://github.com/MindFlavor/Jellyfin.Plugins.Telegram/pull/14

### DIFF
--- a/Emby.Notifications/NotificationManager.cs
+++ b/Emby.Notifications/NotificationManager.cs
@@ -72,6 +72,10 @@ namespace Emby.Notifications
                 .ToArray();
 
             var title = request.Name;
+            if (string.IsNullOrEmpty(title) && relatedItem != null)
+            {
+                title = relatedItem.OriginalTitle;
+            }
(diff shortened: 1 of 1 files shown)


Shadowghost:
Please do not PR against the 10.8 branch, PR against master.

knocte:
> Please do not PR against the 10.8 branch, PR against master.

Thanks for your reply @Shadowghost , but how can I PR this fix against master if the Emby.Notifications project is not there? Where did it go? I also tried to look for a file or type called NotificationManager, in case it was moved to some other location, but I still don't find it.

crobibero:
Emby.Notifications was deprecated in favor of listening to each event. See the webhook plugin for an example as it listens to every event.

https://github.com/jellyfin/jellyfin-plugin-webhook/tree/master/Jellyfin.Plugin.Webhook/Notifiers

PR #11272, opened 2024-03-31
Lowercase CollectionTypeOptions to match legacy experience

crobibero:
Fixes: https://github.com/jellyfin/jellyfin-sdk-swift/issues/34

Swashbuckle doesn't utilize the JsonConverters when writing the spec.


nielsvanvelzen:
This enum was added before 10.9 though:
https://github.com/jellyfin/jellyfin-sdk-kotlin/blob/release-1.4.z/jellyfin-model/src/commonMain/kotlin-generated/org/jellyfin/sdk/model/api/CollectionTypeOptions.kt

crobibero:
It was added, but added incorrectly. The API would return the value lowercase, but the spec has it listed as PascalCase

nielsvanvelzen (approved):

LePips:
Oops I just created https://github.com/jellyfin/jellyfin/issues/11273 without seeing this. `CollectionTypeOptions` is only used for a single property, how difficult would it be to migrate it to just use `CollectionType`?

Bond-009 (approved):

crobibero:
> Oops I just created https://github.com/jellyfin/jellyfin/issues/11273 without seeing this. `CollectionTypeOptions` is only used for a single property, how difficult would it be to migrate it to just use `CollectionType`?

I considered moving to CollectionType but that enum has many more values and would be an actual API change

PR #11280, opened 2024-04-02
fix: av1 codecs string

gnattu:
**Changes**

Browsers are very picky about this string. The profile level has to be padded to double digits otherwise the stream is going to be rejected and the playback would not continue.

This is confirmed to work with latest Chromium 122 and Firefox 124, but Chromium used to be more picky about this string so additional hacks to lying about the level to it may be needed.

**Issues**

Fixes jellyfin/jellyfin-web#5364

### DIFF
--- a/Jellyfin.Api/Helpers/HlsCodecStringHelpers.cs
+++ b/Jellyfin.Api/Helpers/HlsCodecStringHelpers.cs
@@ -220,9 +220,6 @@ public static class HlsCodecStringHelpers
             level = 19;
         }
 
-        // Needed to pad it two double digits; otherwise, browsers will reject the stream.
-        var levelString = level < 10 ? $"0{level}" : $"{level}";
-
         if (bitDepth != 8
             && bitDepth != 10
             && bitDepth != 12)
@@ -232,7 +229,8 @@ public static class HlsCodecStringHelpers
         }
 
         result.Append('.')
-            .Append(levelString)
+            // Needed to pad it double digits; otherwise, browsers will reject the stream.
+            .Append(CultureInfo.InvariantCulture, $"{level:D2}")
             .Append(tierFlag ? 'H' : 'M');
 
         string bitDepthD2 = bitDepth.ToString("D2", CultureInfo.InvariantCulture);
--- a/Jellyfin.Api/Helpers/HlsCodecStringHelpers.cs
+++ b/Jellyfin.Api/Helpers/HlsCodecStringHelpers.cs
@@ -192,7 +192,7 @@ public static class HlsCodecStringHelpers
     /// <returns>The AV1 codec string.</returns>
     public static string GetAv1String(string? profile, int level, bool tierFlag, int bitDepth)
     {
-        // https://aomedia.org/av1/specification/annex-a/
+        // https://aomediacodec.github.io/av1-isobmff/#codecsparam
         // FORMAT: [codecTag].[profile].[level][tier].[bitDepth]
         StringBuilder result = new StringBuilder("av01", 13);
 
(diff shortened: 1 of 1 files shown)


Shadowghost (approved):

Jellyfin.Api/Helpers/HlsCodecStringHelpers.cs
```diff
@@ -214,13 +214,15 @@ public static string GetAv1String(string? profile, int level, bool tierFlag, int
...
-        if (level <= 0
-            || level > 31)
+        if (level is <= 0 or > 31)
         {
             // Default to the maximum defined level 6.3
             level = 19;
         }
 
+        // Needed to pad it two double digits; otherwise, browsers will reject the stream.
+        var levelString = level < 10 ? $"0{level}" : $"level";
```
dmitrylyzo, line 224: Isn't `{}` missing?
```suggestion
        var levelString = level < 10 ? $"0{level}" : $"{level}";
```

Jellyfin.Api/Helpers/HlsCodecStringHelpers.cs
```diff
@@ -214,13 +214,15 @@ public static string GetAv1String(string? profile, int level, bool tierFlag, int
...
-        if (level <= 0
-            || level > 31)
+        if (level is <= 0 or > 31)
         {
             // Default to the maximum defined level 6.3
             level = 19;
         }
 
+        // Needed to pad it two double digits; otherwise, browsers will reject the stream.
+        var levelString = level < 10 ? $"0{level}" : $"{level}";
```
JPVenson, line 224: `level.ToString().PadLeft(2, '0')` ?

Jellyfin.Api/Helpers/HlsCodecStringHelpers.cs
```diff
@@ -230,7 +232,7 @@ public static string GetAv1String(string? profile, int level, bool tierFlag, int
         }
 
         result.Append('.')
-            .Append(level)
+            .Append(levelString)
```
Bond-009, line 235:
```suggestion
            .AppendFormat("{0:D2}", level)
```

Bond-009 (changes requested):

Jellyfin.Api/Helpers/HlsCodecStringHelpers.cs
JPVenson: also a good way

nyanmisaka (approved):
Good catch. The link in line 195 is outdated. Here’s the latest https://aomediacodec.github.io/av1-isobmff/#codecsparam

The level and bitdepth strings can be updated to use AppendFormat.

Jellyfin.Api/Helpers/HlsCodecStringHelpers.cs
gnattu: Done

Jellyfin.Api/Helpers/HlsCodecStringHelpers.cs
```diff
@@ -230,7 +229,8 @@ public static string GetAv1String(string? profile, int level, bool tierFlag, int
         }
 
         result.Append('.')
-            .Append(level)
+            // Needed to pad it double digits; otherwise, browsers will reject the stream.
+            .Append(CultureInfo.InvariantCulture, $"{level:D2}")
```
Bond-009, line 233: Use AppendFormat instead of formatting a string and appending that

Bond-009 (changes requested):

Bond-009 (approved):

PR #11296, opened 2024-04-04
Reduce complexity in video controller

lackolaxton:
Noticed in K8s its holding onto memory. Reduced complexity on the controller in hope I can start to narrow down where.


lackolaxton:
Broke parameter bindings. Will fix and re PR

PR #11300, opened 2024-04-05
Reduce Video Controller Complexity

lackolaxton:
Noticed when hosted in a k8s env it was holding onto memory. Cleaned up the video controller so i can dig into why it is doing so.

Looks to potentially be not disposing state/transcoding files but this will help me find out.


Jellyfin.Api/Helpers/StreamingHelper.cs
```diff
@@ -25,33 +24,21 @@ namespace Jellyfin.Api.Helpers;
 /// <summary>
 /// The streaming helpers.
 /// </summary>
-public static class StreamingHelpers
+public class StreamingHelper(IMediaSourceManager mediaSourceManager, IUserManager userManager, ILibraryManager libraryManager, IServerConfigurationManager serverConfigurationManager, IMediaEncoder mediaEncoder, ITranscodeManager transcodeManager) : IStreamingHelper
```
Shadowghost, line 27: Usually we use a dedicated constructor

nielsvanvelzen:
Please revert the API changes

lackolaxton:
> Please revert the API changes

The actual API has not changed, only the auto API definition. I was hoping to bring the changes in because it has improved readability due to there only being 2 parameters rather than 10's.

nielsvanvelzen:
The casing of the query parameters has changed.

lackolaxton:
> The casing of the query parameters has changed.

Ill add the property name field to the model? I believe DOTNET should handle converting camelcase to Proper but leave it with me

Jellyfin.Api/Controllers/VideosController.cs
```diff
@@ -253,225 +251,71 @@ public async Task<ActionResult> MergeVersions([FromQuery, Required, ModelBinder(
...
-        [FromQuery] string? liveStreamId,
-        [FromQuery] bool? enableMpegtsM2TsMode,
-        [FromQuery] [RegularExpression(EncodingHelper.ValidationRegex)] string? videoCodec,
-        [FromQuery] [RegularExpression(EncodingHelper.ValidationRegex)] string? subtitleCodec,
-        [FromQuery] string? transcodeReasons,
-        [FromQuery] int? audioStreamIndex,
-        [FromQuery] int? videoStreamIndex,
-        [FromQuery] EncodingContext? context,
-        [FromQuery] Dictionary<string, string> streamOptions)
+    public async Task<ActionResult> GetVideoStream([FromRoute][Required] Guid itemId, [FromQuery] VideoStreamRequest request)
```
crobibero, line 261: We moved away from using a single object for query parameters. Please revert this change.

crobibero (changes requested):

Jellyfin.Api/Controllers/VideosController.cs
lackolaxton: Reverted.

Jellyfin.Api/Helpers/StreamingHelper.cs
lackolaxton: Reverted.

lackolaxton:
How do we get people to re review this?

nielsvanvelzen:
Just wait. This PR is extremely low priority and our focus is on 10.9 related changes at the moment.

lackolaxton:
Hi,
I have updated this.

Ideally I would like to get this in, as it will make it easier for me to investigate 2 issues i am seeing.

- High memory usage in K8s.
- Throttling transcoding is causing freezing

Issue #11297, opened 2024-04-05
[Issue]: /Library/VirtualFolders endpoint returns 403


joshuaboniface:
### Please describe your bug

GET requests to the /Library/VirtualFolders endpoint return a 403 even with a properly configured authentication token.

### Reproduction Steps

1. Create an API key (`xxxxxx` here for security).
2. Hit the endpoint as follows:
```
$ curl -v -X GET -d "" -H "X-MediaBrowser-Token: ${JELLYFIN_API_KEY}" "${JELLYFIN_ADDRESS}/Library/VirtualFolders"                                                                               
*   Trying x.x.x.x:8096...
* Connected to jf1 (x.x.x.x) port 8096 (#0)
> GET /Library/VirtualFolders HTTP/1.1
> Host: jf1:8096
> User-Agent: curl/7.74.0
> Accept: */*
> X-MediaBrowser-Token: xxxxxx
> Content-Length: 0
> Content-Type: application/x-www-form-urlencoded
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 403 Forbidden
< Content-Length: 0
< Date: Fri, 05 Apr 2024 03:53:40 GMT
< Server: Kestrel
< X-Response-Time-ms: 2.4509
< 
* Connection #0 to host jf1 left intact
```

### Jellyfin Version

Unstable (master branch)

### Environment

```markdown
- OS: Debian
- Linux Kernel: 5.10
- Virtualization: KVM
- Clients: JMP, Browser
- Browser: Firefox 115
- FFmpeg Version: Latest
- Playback Method: N/A
- Hardware Acceleration: N/A
- GPU Model: N/A
- Plugins: N/A
- Reverse Proxy: Haproxy
- Base URL: /
- Networking: Default
- Storage: CephFS
```

### Jellyfin logs

```shell
Apr 04 23:54:48 jf1 jellyfin[480]: [DBG] AuthenticationScheme: CustomAuthentication was successfully authenticated.
Apr 04 23:54:48 jf1 jellyfin[480]: [DBG] AuthenticationScheme: CustomAuthentication was successfully authenticated.
Apr 04 23:54:48 jf1 jellyfin[480]: [INF] AuthenticationScheme: CustomAuthentication was forbidden.
```

### FFmpeg logs

```shell
N/A
```

### Please attach any browser or client logs here

N

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

joshuaboniface:
[...]

PR #11338, opened 2024-04-11
Allow ApiKey to authorize against the FirstTimeSetupOrElevated policy

crobibero:
Fixes https://github.com/jellyfin/jellyfin/issues/11297

### DIFF
--- a/Jellyfin.Api/Auth/FirstTimeSetupPolicy/FirstTimeSetupHandler.cs
+++ b/Jellyfin.Api/Auth/FirstTimeSetupPolicy/FirstTimeSetupHandler.cs
@@ -1,10 +1,6 @@
 using System.Threading.Tasks;
 using Jellyfin.Api.Constants;
-using Jellyfin.Api.Extensions;
-using Jellyfin.Extensions;
 using MediaBrowser.Common.Configuration;
-using MediaBrowser.Common.Extensions;
-using MediaBrowser.Controller.Library;
 using Microsoft.AspNetCore.Authorization;
 
 namespace Jellyfin.Api.Auth.FirstTimeSetupPolicy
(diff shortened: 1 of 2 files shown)


Bond-009 (approved):

nielsvanvelzen (approved):

Jellyfin.Api/Auth/FirstTimeSetupPolicy/FirstTimeSetupHandler.cs
```diff
@@ -40,6 +40,13 @@ protected override Task HandleRequirementAsync(AuthorizationHandlerContext conte
             }
 
             var contextUser = context.User;
+            if (contextUser.GetIsApiKey())
```
cvium, line 43: Shouldn't this just check for Admin role? We have too much authz logic tied to API Keys tbh. API keys and real admin users should both have the Administrator role.

Jellyfin.Api/Auth/FirstTimeSetupPolicy/FirstTimeSetupHandler.cs
```diff
@@ -36,37 +27,14 @@ protected override Task HandleRequirementAsync(AuthorizationHandlerContext conte
...
-            var user = _userManager.GetUserById(userId);
-            if (user is null)
-            {
-                throw new ResourceNotFoundException();
-            }
-
-            if (user.IsParentalScheduleAllowed())
+            else
             {
+                // Any user-specific checks are handled in the DefaultAuthorizationHandler.
```
cvium, line 37: Are those triggered when this one succeeds?

crobibero: The DefaultAuthorizationHandler is called before the FirstTimeSetupHandler

Even if the FirstTimeSetupHandler calls Succeed on the requirement the call will still result in a 403

cvium (approved):

Bond-009 (approved):

Issue #11264, opened 2024-03-30
[Issue]: Cannot update refresh progress of item 'Folder' (...) because a refresh for this item is not running


sjorge:
### Please describe your bug

After upgrading from 10.8.13 -> 10.9.0 while running a `Scan all libraries` (After updating all plugins) it seems this never really finishes and I get a lot of:

```
Mar 29 15:03:08 artemis.acheron.be jellyfin[236568]: [15:03:08] [ERR] Error refreshing folder
Mar 29 15:03:08 artemis.acheron.be jellyfin[236568]: System.InvalidOperationException: Cannot update refresh progress of item 'Folder' (ed4b05e7fae6991b5d3a95b1bf5bb510) because a refresh for this item is not running
Mar 29 15:03:08 artemis.acheron.be jellyfin[236568]:    at MediaBrowser.Providers.Manager.ProviderManager.<>c__DisplayClass69_0.<OnRefreshProgress>b__0(Guid _)
Mar 29 15:03:08 artemis.acheron.be jellyfin[236568]:    at MediaBrowser.Providers.Manager.ProviderManager.OnRefreshProgress(BaseItem item, Double progress)
Mar 29 15:03:08 artemis.acheron.be jellyfin[236568]:    at MediaBrowser.Controller.Entities.Folder.<>c__DisplayClass64_2.<ValidateChildrenInternal2>b__1(Double innerPercent)
```

I left it running for a nigh and it still never finished. Deleting the library to start fresh also doesn't help as the library never seems to delete.

However nuking `/var/lib/jellyfin`, /var/cache/jellyfin` and `/etc/jellyfin` and going through the setup wizard again on 10.9.0 and adding a new library does seem to work and I do not get these errors. (I only did a partial movie library)

### Reproduction Steps

1. install unstable (20240325) over an existing stable
2. Swap the plugin repo to the unstable one
3. trigger the plugin update schedule task
4. restart to load the updated plugins
5. trigger a scan all libraries

### Jellyfin Version

Unstable (master branch)

### if other:

10.9.0

### Environment

```markdown
- OS: Ubuntu 22.04.4 LTS
- Linux Kernel: GNU/Linux 6.5.0-26-generic x86_64 (for ARC support)
- Virtualization: BHYVE
- Clients: Browser
- Browser: Firefox 124.0b9 (64-bit)
- FFmpeg Version: jellyfin-ffmpeg6 - 6.0.1-5-jammy
- Playback Method: n/a
[...]

PR #11341, opened 2024-04-11
Only update if actively refreshing

crobibero:
Fixes https://github.com/jellyfin/jellyfin/issues/11264

`.AddOrUpdate()` is doing essentially the same thing internally, this just removes the exception

### DIFF
--- a/MediaBrowser.Providers/Manager/ProviderManager.cs
+++ b/MediaBrowser.Providers/Manager/ProviderManager.cs
@@ -969,7 +969,8 @@ namespace MediaBrowser.Providers.Manager
             _logger.LogDebug("OnRefreshProgress {Id:N} {Progress}", id, progress);
 
             if (!_activeRefreshes.TryGetValue(id, out var current)
-                || !_activeRefreshes.TryUpdate(id, progress, current))
+                || progress <= current
+                || (progress > current && !_activeRefreshes.TryUpdate(id, progress, current)))
             {
                 // Item isn't currently refreshing so don't trigger event.
                 return;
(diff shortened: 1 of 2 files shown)


mark-monteiro (approved):

Shadowghost (approved):

nielsvanvelzen:
Even with the exception gone, the actual issue where it tries to update out of order still exists. I believe this affects the scanner progress in the dashboard. Is that fine or do we want to fix it in a different way?

crobibero:
I think the issue is that the refresh progress is called after refresh completes. When the refresh completes and the item is removed from the dictionary a progress of `100` is sent, so I don't think this will have any negative impact on user experience

MediaBrowser.Providers/Manager/ProviderManager.cs
```diff
@@ -969,7 +969,8 @@ public void OnRefreshProgress(BaseItem item, double progress)
             _logger.LogDebug("OnRefreshProgress {Id:N} {Progress}", id, progress);
 
             if (!_activeRefreshes.TryGetValue(id, out var current)
-                || !_activeRefreshes.TryUpdate(id, progress, current))
+                || progress <= current
```
mark-monteiro, line 972:
```suggestion
```
No need to check this condition twice?

MediaBrowser.Providers/Manager/ProviderManager.cs
```diff
@@ -969,7 +969,8 @@ public void OnRefreshProgress(BaseItem item, double progress)
...
+                || (progress > current && !_activeRefreshes.TryUpdate(id, progress, current)))
             {
                 // Item isn't currently refreshing so don't trigger event.
```
mark-monteiro, line 975:
```suggestion
                // Item isn't currently refreshing, or update was received out-of-order, so don't trigger event.
```

mark-monteiro (approved):

barronpm:
There should be a try-catch somewhere in the callers that can be removed now, if I understand the changes correctly, I added it when I was replacing our custom `IProgress` impl with `System.Progress`

nielsvanvelzen (approved):

PR #11351, opened 2024-04-12
fix: fix off-by-one error in `GetAttributeValue`

revam:
**Changes**
The `GetAttributeValue` extension will currently fail with ids in the single digit, because it's using a length in a place where it should be an index.

@fearnlj01 helped me debug an issue with a copy of this extension used in the Shokofin plugin, where it failed to recognize single digit ids, which in turn has lead to this PR to fix it upstream.

**Issues**
\-


crobibero:
This looks like a great fix for a unit test 😉

Shadowghost (approved):

Bond-009 (approved):

crobibero (approved):

Issue #11349, opened 2024-04-12
[Issue]: 10.9 restarting Jellyfin via Dashboard causes Plugin error


jumoog:
### Please describe your bug

Restarting the server from the dashboard causes problems with external plugins. I can reproduce this using `LDAP-Auth` and my plugin.

### Reproduction Steps

- Install Jellyfins `LDAP-Auth` plugin
- Restart the Server to have a complelet install
- Restart the Server via Dashboard
- Visit the `LDAP-Auth` plugin config Page
- Settings wont load and you get a error in the logs

### Jellyfin Version

Unstable (master branch)

### if other:

jellyfin/jellyfin:2024040805-amd64

### Environment

```markdown
- Virtualization: Docker
```

### Jellyfin logs

```shell
[2024-04-12 15:32:01.847 +02:00] [ERR] [17] Jellyfin.Api.Middleware.ExceptionMiddleware: Error processing request. URL "GET" "/Plugins/958aad66-3784-4d2a-b89a-a7b6fab6e25c/Configuration".
System.InvalidOperationException: There was an error generating the XML document.
 ---> System.InvalidCastException: [A]Jellyfin.Plugin.LDAP_Auth.Config.PluginConfiguration cannot be cast to [B]Jellyfin.Plugin.LDAP_Auth.Config.PluginConfiguration. Type A originates from 'LDAP-Auth, Version=18.0.0.0, Culture=neutral, PublicKeyToken=null' in the context '"" Emby.Server.Implementations.Plugins.PluginLoadContext #4' at location '/config/plugins/LDAP Authentication_18.0.0.0/LDAP-Auth.dll'. Type B originates from 'LDAP-Auth, Version=18.0.0.0, Culture=neutral, PublicKeyToken=null' in the context '"" Emby.Server.Implementations.Plugins.PluginLoadContext #3' at location '/config/plugins/LDAP Authentication_18.0.0.0/LDAP-Auth.dll'.
   at Microsoft.Xml.Serialization.GeneratedAssembly.XmlSerializationWriterPluginConfiguration.Write5_PluginConfiguration(Object o)
   at InvokeStub_XmlSerializationWriterPluginConfiguration.Write5_PluginConfiguration(Object, Span`1)
   at System.Reflection.MethodBaseInvoker.InvokeWithOneArg(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   --- End of inner exception stack trace ---
[...]

PR #11356, opened 2024-04-14
fix: use new serializer cache per IXmlSerializer

crobibero:
Fixes https://github.com/jellyfin/jellyfin/issues/11349

Other option is to clear the dictionary when a new `MyXmlSerializer` is created, but this is cleaner and essentially just as performant


Bond-009 (approved):

jumoog (approved):
works!

nielsvanvelzen (approved):

PR #11361, opened 2024-04-14
 Properly await Task.Delay() 

Bond-009:
`.Wait()` bad, `Task.Delay().Wait()` special kind of bad


crobibero (approved):

nielsvanvelzen:
Please add a description, what is the benefit of this change?

MarkCiliaVincenti:
Why not go async all the way and then await Task.Delay instead?

Shadowghost (approved):

Issue #11366, opened 2024-04-16
[Issue]: SessionController permissions


bin101:
### Please describe your bug

A user can see every session of every user if they do not provide their own user ID as a query parameter. Instead, it is assumed that the user has administrator rights.

Noticed this as a friend of mine said he can see every session via his homarr dashboard.

### Reproduction Steps

1. Create a user without admin rights
2. Ask for all sessions via curl
3. Profit 

### Jellyfin Version

10.8.13

### Environment

```markdown
n/a
```

### Jellyfin logs

```shell
n/a
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #11370, opened 2024-04-16
Limit sessions per user

crobibero:
Partially fixes https://github.com/jellyfin/jellyfin/issues/11366

I tried adding a test for this but having the logic directly in the controller make it very difficult


bin101 (commented):
This will not fix the bug as the surrounding if block will not be entered when controllableByUserId is not provided inside the request.

crobibero:
> This will not fix the bug as the surrounding if block will not be entered when controllableByUserId is not provided inside the request.

I believe L109-113 addresses your concern. If not- please be more specific

bin101:
It does. Don't know how I missed it this morning, that why I already marked it as solved.

Shadowghost (approved):

Bond-009 (approved):

PR #11377, opened 2024-04-17
fix: don't apply whitelist tags to libraries

### LINKED ISSUE #11375: Allowed Tags in Parental Controls - Not Working?
Is this working?

I am on version 10.9.0

Screen Recording 2024-04-17 at 9 14 59 AM

_Originally posted by @adoreparler in https://github.com/jellyfin/jellyfin/issues/9139#issuecomment-2061242953_

gnattu:
**Changes**

The way the whitelist tag currently works is weird. It requires the entire library to have that tag to be shown, and if the library has the tag, all items inside it inherit that tag, making all items visible. This change aims to make it more sensible by allowing the whitelist to work on a per-item basis, ensuring that the allowed libraries are always shown and the whitelist tag only applies to the items inside the library, not the library itself.

**Issues**

Fixes #11375

### DIFF
--- a/MediaBrowser.Controller/Entities/BaseItem.cs
+++ b/MediaBrowser.Controller/Entities/BaseItem.cs
@@ -1602,6 +1602,12 @@ namespace MediaBrowser.Controller.Entities
                 return false;
             }
 
+            var parent = GetParents().FirstOrDefault() ?? this;
+            if (parent is UserRootFolder)
+            {
+                return true;
+            }
+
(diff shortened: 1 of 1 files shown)


Shadowghost (approved):

adoreparler:
Working how I expected
My Testing:

`-> jellyfin git:(better-whitelist-tag) dotnet run --project Jellyfin.Server --webdir /home/seion/jellyfin-web/dist`

Screen Recording 2024-04-18 at 9 17 07 AM

crobibero (approved):

PR #11386, opened 2024-04-18
Add Open Telemetry Exporter

lackolaxton:
Add OTEL exporter options. For anyone that self hosts and has a collector, it would be useful for thme to get the metrics and tracing to help monitor performance.


rasmustwh:
Wouldn't auto-instrumentation be more useful and flexible for everyone? The hardcoded instrumentation seems rather overkill for the most use cases.

lackolaxton:
> Wouldn't auto-instrumentation be more useful and flexible for everyone? The hardcoded instrumentation seems rather overkill for most use cases.

Never really thought of auto instrumentation. If someone is hosting in containers though it'll make it hard to do, so having this option at least gives the Hoster a way to get decent metrics etc.

I plan to try and add some tracing and better metrics so this is just a start

MediaBrowser.Model/Configuration/OtelConfiguration.cs
```diff
@@ -0,0 +1,27 @@
...
+
+    /// <summary>
+    /// Gets or Sets the endpoint to exporter to.
+    /// </summary>
+    public string Endpoint { get; set; } = string.Empty;
+
+    /// <summary>
+    /// Gets or sets the meters to listen to.
+    /// </summary>
+    public string Meters { get; set; } = string.Empty;
```
crobibero, line 21: Looks like this should be an array

MediaBrowser.Model/Configuration/OtelConfiguration.cs
```diff
@@ -0,0 +1,27 @@
...
+
+    /// <summary>
+    /// Gets or sets the activity sources to listen to.
+    /// </summary>
+    public string ActivitySources { get; set; } = string.Empty;
```
crobibero, line 26: Looks like this should be an array

Jellyfin.Server/Extensions/ApiServiceCollectionExtensions.cs
```diff
@@ -255,6 +259,48 @@ public static IServiceCollection AddJellyfinApiSwagger(this IServiceCollection s
...
+                serviceCollection
+                    .AddOpenTelemetry()
+                    .ConfigureResource(resourceBuilder => resourceBuilder.AddService("JellyFin", appVersion))
+                    .WithMetrics(metrics =>
+                    {
+                        metrics
+                        .AddMeter(config.Meters)
+                        .AddAspNetCoreInstrumentation()
+                        .AddRuntimeInstrumentation()
+                        .AddHttpClientInstrumentation();
```
crobibero, line 282:
```suggestion
                            .AddMeter(config.Meters)
                            .AddAspNetCoreInstrumentation()
                            .AddRuntimeInstrumentation()
                            .AddHttpClientInstrumentation();
```

Jellyfin.Server/Extensions/ApiServiceCollectionExtensions.cs
```diff
@@ -255,6 +259,48 @@ public static IServiceCollection AddJellyfinApiSwagger(this IServiceCollection s
...
+                        {
+                            exporter.Endpoint = new Uri(config.Endpoint);
+                        });
+                    })
+                    .WithTracing(tracing =>
+                    {
+                        tracing
+                        .AddSource(config.ActivitySources)
+                        .AddAspNetCoreInstrumentation()
+                        .AddHttpClientInstrumentation();
```
crobibero, line 293:
```suggestion
                            .AddSource(config.ActivitySources)
                            .AddAspNetCoreInstrumentation()
                            .AddHttpClientInstrumentation();
```

Jellyfin.Server/Extensions/ApiServiceCollectionExtensions.cs
```diff
@@ -255,6 +259,48 @@ public static IServiceCollection AddJellyfinApiSwagger(this IServiceCollection s
...
+        /// <param name="config">Otel Config.</param>
+        /// <param name="appVersion">Application Version.</param>
+        /// <returns>Updated Service Collection.</returns>
+        public static IServiceCollection AddOTel(this IServiceCollection serviceCollection, OtelConfiguration config, string appVersion)
+        {
+            if (config.Enabled)
+            {
+                serviceCollection
+                    .AddOpenTelemetry()
+                    .ConfigureResource(resourceBuilder => resourceBuilder.AddService("JellyFin", appVersion))
```
crobibero, line 275:
```suggestion
                    .ConfigureResource(resourceBuilder => resourceBuilder.AddService("Jellyfin", appVersion))
```

crobibero (changes requested):
if any of the properties are improperly specified it will cause the server to fail to start- I think we should just log the exception and continue.

PR #11388, opened 2024-04-19
fix: restore os name in GetPublicSystemInfo endpoint

### LINKED ISSUE #11382: [Issue]: OS name endpoint was removed and the VPP tonemap options disappears
### Please describe your bug

The OS name endpoint was removed in [#9175](https://github.com/jellyfin/jellyfin/pull/9175) since it was considered useless, but the VPP tonemap options in the [web client relies on this](https://github.com/jellyfin/jellyfin-web/blob/17b1a757ddcfd317f34881628c7ca79b14481db7/src/controllers/dashboard/encodingsettings.js#L234) to choose to display on Linux and hide on Windows.

```js
        if (systemInfo.OperatingSystem.toLowerCase() === 'linux' && (this.value == 'qsv' || this.value == 'vaapi')) {
            page.querySelector('.vppTonemappingOptions').classList.remove('hide');
        } else {
            page.querySelector('.vppTonemappingOptions').classList.add('hide');
        }
```

In the future, cross-platform Vulkan and VA-API HWA may also need OS name to display the corresponding web options. So I'm not sure if it's necessary to revert this change.

### Reproduction Steps

1. In Jellyfin server Linux/Docker builds
2. Go to Dashboard -> Playback
3. Select Intel QuickSync HWA
4. You can see the VPP tonemap options are missing

### Jellyfin Version

Unstable (master branch)

### if other:

_No response_

### Environment

```markdown
- OS: Linux
- Linux Kernel: Any
- Virtualization: None and Docker
- Clients: Web client
- Browser: Any
- FFmpeg Version: Not used
- Playback Method: Not used
- Hardware Acceleration: Not used
- GPU Model: Not used
- Plugins: Not used
- Reverse Proxy: Not used
- Base URL: Not used
- Networking: Localhost
- Storage: Local
```

### Jellyfin logs

```shell
This error is not shown in the server log
```

### FFmpeg logs

```shell
This error is not shown in the ffmpeg log
```

This error is not shown in the browser log

image

### Code of Conduct

gnattu:
**Changes**

This is still used by clients to show platform-specific hardware acceleration options, restore the os name for now.

A better approach would be to create another endpoint that returns the underlying hardware capabilities from the server instead of checking from the client side. Hardware capabilities can be very complicated and cannot be easily tested by clients with limited information. With this API, we can remove the OS name again from GetPublicSystemInfo.

**Issues**

Fixes #11382

### DIFF
--- a/Emby.Server.Implementations/SystemManager.cs
+++ b/Emby.Server.Implementations/SystemManager.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Linq;
 using System.Threading.Tasks;
 using MediaBrowser.Common.Configuration;
@@ -78,6 +79,7 @@ public class SystemManager : ISystemManager
         {
             Version = _applicationHost.ApplicationVersionString,
             ProductName = _applicationHost.Name,
+            OperatingSystem = GetServerOs(),
             Id = _applicationHost.SystemId,
             ServerName = _applicationHost.FriendlyName,
             LocalAddress = _applicationHost.GetSmartApiUrl(request),
@@ -99,5 +101,25 @@ public class SystemManager : ISystemManager
             _applicationHost.ShouldRestart = restart;
             _applicationLifetime.StopApplication();
         });
+    }
+
+    private static string GetServerOs()
+    {
+        if (OperatingSystem.IsLinux())
+        {
+            return "Linux";
+        }
+
+        if (OperatingSystem.IsWindows())
+        {
+            return "Windows";
+        }
+
+        if (OperatingSystem.IsMacOS())
+        {
+            return "macOS";
+        }
+
+        return "Others";
     }
 }
--- a/MediaBrowser.Model/System/PublicSystemInfo.cs
+++ b/MediaBrowser.Model/System/PublicSystemInfo.cs
@@ -34,7 +34,6 @@ namespace MediaBrowser.Model.System
         /// Gets or sets the operating system.
         /// </summary>
         /// <value>The operating system.</value>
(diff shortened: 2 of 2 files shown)


nielsvanvelzen:
I don't think we should add this back like this. If we truly need to know the operating system in the dashboard it should be added to an endpoint only accessible by authenticated user or preferably administrators.
Although, if it is only used to know which hardware acceleration options are available we should probably just add those options to the API instead.

gnattu:
Our current function is broken, and a new API is not suitable for 10.9, and this is the only way to restore the intended behavior for now, otherwise we will use a more hacky way to workaround this in the web, and you will not like that approach.

nielsvanvelzen:
If we're not going to add a proper fix now I'd say we just remove the OS check in web (always show that option) instead of changing the API now.

Bond-009:
> Although, if it is only used to know which hardware acceleration options are available we should probably just add those options to the API instead.

I agree with this, we should add 'CanVppTonemap' or something like that somewhere in the API behind an authenticated endpoint

nyanmisaka:
> If we're not going to add a proper fix now I'd say we just remove the OS check in web (always show that option) instead of changing the API now.

I think it would be acceptable to remove the Linux check for VppTonemap option in the web client, because currently even if this option is forcibly checked on Windows, nothing will happen. We can add VppTonemap support on Windows when we migrate to FFmpeg 6.1+, or we backport manually.

As for new features in the future, we can add new endpoints after refactoring the transcoder to better handle this.

PR #11390, opened 2024-04-20
Enable nullable for UserItemData

Bond-009:
MetadataResult.GetOrAddUserData doesn't ever get used and is probably broken since the migration to .NET Core as it still expects a Guid for userId

### DIFF
--- a/MediaBrowser.Controller/Entities/UserItemData.cs
+++ b/MediaBrowser.Controller/Entities/UserItemData.cs
@@ -1,5 +1,3 @@
-#nullable disable
-
 #pragma warning disable CS1591
 
 using System;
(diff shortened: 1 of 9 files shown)


crobibero (approved):

MediaBrowser.Controller/Entities/UserItemData.cs
```diff
@@ -19,17 +17,11 @@ public class UserItemData
...
-        /// </summary>
-        /// <value>The user id.</value>
-        public Guid UserId { get; set; }
-
         /// <summary>
         /// Gets or sets the key.
         /// </summary>
         /// <value>The key.</value>
-        public string Key { get; set; }
+        public required string Key { get; set; }
```
nielsvanvelzen, line 24: Slightly unrelated but it might be worth updating this summary because I have absolutely no idea what "the key" means

nielsvanvelzen (approved):
✅ API changes

Issue #11113, opened 2024-03-05
[Issue]: Wrong play time shown


stefan1983:
### Please describe your bug

Hey there,

I am using Jellyfin 10.8.13 on a Linux server 5.10.0-0.deb10.16-amd64 Debian 5.10.127-2~bpo10+1 (2022-07-28) x86_64.
When using the web browser or any client to play my music, I always get a wrong play time shown. Especially when using gapless tracks (live albums) those are completely off:

Web player:
![308967832-20901e66-b8b3-436a-8eac-807f9bbf3e07](https://github.com/jellyfin/jellyfin/assets/8542806/55d498fa-c4a2-4e3a-8753-d3f692ec1e9b)

Finamp (counts further):
![308661745-136cd2cd-dae6-4ce7-9f2e-adc985eeeca9](https://github.com/jellyfin/jellyfin/assets/8542806/531a347d-1877-437c-aac3-10545ecbd15f)

Is there anything I can do on server side?
The files don't have any issues when playing somewhere else (foobar2000, IINA, VLC...).
That are just standard MP3 or FLAC files.

### Reproduction Steps

Just play music in the web or any available jellyfin compatible app (iOS / macOS).

### Jellyfin Version

10.8.13

### Environment

```markdown
- OS: Debian
- Linux Kernel: -
- Virtualization: - 
- Clients: WEB, Jellyfin iOS App, Finamp
- Browser: Safari / Chrome
- FFmpeg Version: -  
- Playback Method: -
- Hardware Acceleration: none
- GPU Model: Intel Onboard
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

```shell
Attached
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

stefan1983:
[Log.txt](https://github.com/jellyfin/jellyfin/files/14500578/Log.txt)

stefan1983:
I would be able to deliver sample files if needed.

stefan1983:
I just found out that it is specific to macOS Safari. When using another browser, there is no such issue. 
Currently I am using Safari Version 17.3.1:

<img width="396" alt="image" src="https://github.com/jellyfin/jellyfin/assets/8542806/226820da-aaf6-40d8-8034-b5acb4168a78">

gnattu:
[...]

Issue #3731, opened 2020-07-28
Can't read certain FLAC files


Jerroder:
**Describe the bug**
I can play all of my FLAC music but I have one album for which I get :
````
Playback Error
This client isn't compatible with the media and the server isn't sending a compatible media format.
````

**System (please complete the following information):**
 - OS: Debian 10
 - Virtualization: KVM
 - Clients: Browser
 - Browser: Firefox 78.1.0esr
 - Jellyfin Version: 10.6.1 but also present in 10.5
 - Playback: Direct Play I believe
 - Installed Plugins: none,
 - Reverse Proxy: none
 - Base URL: none
 - Networking: LAN
 - Storage: local

**Logs**
````
[...]

PR #11399, opened 2024-04-21
Add audio remux to UniversalAudioController

gnattu:
**Changes**

This PR adds audio remuxing (copying audio into another container without transcoding) support to the UniversalAudioController, allowing the client to request the audio to be in a specific container. When remuxing with HLS, the HLS container's constraint applies, and it will fallback to transcoding when the original audio codec is not supported by the chosen HLS container. When remuxing with HTTP, this assumes that the client-provided container supports the audio codec. If it does not, ffmpeg would fail.

To workaround the complex codec+container support situation in HLS, an HLS container override feature has been added. The client can explicitly require a container+codec combination supported by HLS. For example, `ts|mp3` and `mp4|flac`. When combinations like these are present, the HLS stream will use the container specified in this profile, ignoring the one set by the `TranscodingContainer` query string. The most common use case for this mode is to force mp3 to be remuxed in mpegts container because most browsers do not support mp3 in mp4 container, although it is technically supported by HLS. When specifying such overrides, the client should be responsible to ensure the container and codec compatibility. 

This would allow clients with strict container requirements to playback certain audio codecs. For example, Opus on Safari can be played with HLS + fmp4.

This can also workarounds a lot of container related issues, as listed below in the issues section.

**Issues**

#11113 This one is due to the flac/mp3 container containing too little information, the client's estimation of the "ticks" might not be accurate enough. Allowing the client to force an HLS remuxing with fmp4 can mitigate this issue as the mp4 container has more information about the timing. Most browsers do not support mp3 in mp4 container, so probably need to force to remux into mpegts to workaround the timing issue for mp3.

#3731 This issue arises because the flac container has non-standard metadata tags, causing the client to refuse to play the file. This can be mitigated by falling back to HLS remuxing when direct play fails, or simply by forcing HLS remuxing anyway.

### DIFF
--- a/Jellyfin.Api/Controllers/UniversalAudioController.cs
+++ b/Jellyfin.Api/Controllers/UniversalAudioController.cs
@@ -17,6 +17,7 @@ using MediaBrowser.Controller.MediaEncoding;
 using MediaBrowser.Controller.Streaming;
 using MediaBrowser.Model.Dlna;
 using MediaBrowser.Model.MediaInfo;
+using MediaBrowser.Model.Session;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
@@ -137,6 +138,8 @@ public class UniversalAudioController : BaseJellyfinApiController
         // set device specific data
         foreach (var sourceInfo in info.MediaSources)
         {
+            sourceInfo.TranscodingContainer = transcodingContainer;
+            sourceInfo.TranscodingSubProtocol = transcodingProtocol ?? sourceInfo.TranscodingSubProtocol;
             _mediaInfoHelper.SetDeviceSpecificData(
                 item,
                 sourceInfo,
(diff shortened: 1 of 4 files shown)


nyanmisaka (approved):

Jellyfin.Api/Controllers/UniversalAudioController.cs
```diff
@@ -171,27 +174,30 @@ public class UniversalAudioController : BaseJellyfinApiController
...
         {
             // hls segment container can only be mpegts or fmp4 per ffmpeg documentation
             // ffmpeg option -> file extension
             //        mpegts -> ts
             //          fmp4 -> mp4
-            // TODO: remove this when we switch back to the segment muxer
             var supportedHlsContainers = new[] { "ts", "mp4" };
 
+            // fallback to mpegts if device reports some weird value unsupported by hls
+            var requestedSegmentContainer = Array.Exists(supportedHlsContainers, element => element == transcodingContainer) ? transcodingContainer : "ts";
```
Bond-009, line 189:
```suggestion
            var requestedSegmentContainer = Array.Exists(
                supportedHlsContainers, 
                element => string.Equals(element, transcodingContainer, StringComparison.OrdinalIgnoreCase)) ? transcodingContainer : "ts";
```

Jellyfin.Api/Controllers/UniversalAudioController.cs
```diff
@@ -171,27 +174,30 @@ public class UniversalAudioController : BaseJellyfinApiController
...
+            var segmentContainer = Array.Exists(supportedHlsContainers, element => element == mediaSource.TranscodingContainer) ? mediaSource.TranscodingContainer : requestedSegmentContainer;
```
Bond-009, line 190:
```suggestion
            var segmentContainer = Array.Exists(
                supportedHlsContainers,
                element => string.Equals(element, mediaSource.TranscodingContainer, StringComparison.OrdinalIgnoreCase)) ? mediaSource.TranscodingContainer : requestedSegmentContainer;
```

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -124,6 +124,46 @@ public StreamBuilder(ITranscoderSupport transcoderSupport, ILogger logger)
                 }
             }
 
+            if (directPlayMethod is PlayMethod.DirectStream)
+            {
+                var remuxContainer = item.TranscodingContainer ?? "ts";
+                var supportedHlsContainers = new[] { "ts", "mp4" };
+                // If the container specified for the profile is an HLS supported container, use that container instead, overriding the preference
+                // The client should be responsible to ensure this container is compatible
+                remuxContainer = Array.Exists(supportedHlsContainers, element => element == directPlayInfo.Profile?.Container) ? directPlayInfo.Profile?.Container : remuxContainer;
```
Bond-009, line 133:
```suggestion
                remuxContainer = Array.Exists(supportedHlsContainers, element => string.Equals(element, directPlayInfo.Profile?.Container, StringComparison.OrdinalIgnoreCase)) ? directPlayInfo.Profile?.Container : remuxContainer;
```

Bond-009 (changes requested):

Bond-009 (approved):

Issue #11413, opened 2024-04-23
[Issue]: Jellyfin starts 2 ffprobe instances per file when scanning


felix920506:
### Please describe your bug

Jellyfin unstable starts 2 ffprobe instances for each file with the exact same parameters according to the logs.

### Reproduction Steps

1. Rescan a music album with "Replace All Metadata"
2. See Error in logs

### Jellyfin Version

Unstable (master branch)

### Environment

```markdown
- OS: Windows 11
- Linux Kernel: N/A
- Virtualization: None
- Clients: Web
- Browser: Firefox
- FFmpeg Version: 6.0.1-jellyfin
- Playback Method: N/A
- Hardware Acceleration: N/A
- GPU Model: N/A
- Plugins: Default
- Reverse Proxy: Caddy
- Base URL:jftest
- Networking: Reverse Proxy
- Storage: SMB
```

### Jellyfin logs

```shell
[2024-04-23 10:46:12.381 +08:00] [INF] [26] MediaBrowser.MediaEncoding.Encoder.MediaEncoder: Starting "ffprobe" with args "-analyzeduration 200M -probesize 1G -i file:\"Y:\acg-music\僕は今夜、出て行く\01 僕は今夜、出て行く.flac\" -threads 0 -v warning -print_format json -show_streams -show_format"
[2024-04-23 10:46:12.514 +08:00] [INF] [19] MediaBrowser.MediaEncoding.Encoder.MediaEncoder: Starting "ffprobe" with args "-analyzeduration 200M -probesize 1G -i file:\"Y:\acg-music\僕は今夜、出て行く\01 僕は今夜、出て行く.flac\" -threads 0 -v warning -print_format json -show_streams -show_format"
[2024-04-23 10:46:13.296 +08:00] [INF] [26] MediaBrowser.MediaEncoding.Encoder.MediaEncoder: Starting "ffprobe" with args "-analyzeduration 200M -probesize 1G -i file:\"Y:\acg-music\僕は今夜、出て行く\02 あやふやな世界観.flac\" -threads 0 -v warning -print_format json -show_streams -show_format"
[...]

PR #11418, opened 2024-04-23
Always scan ReplayGain tag

Bond-009:
* Removes the UseReplayGainTags option from the LibraryOptions
* Removes extra ffrobe instance when option was enabled
* Makes LUFS value nullable

cc @TelepathicWalrus

### DIFF
--- a/MediaBrowser.Controller/Entities/BaseItem.cs
+++ b/MediaBrowser.Controller/Entities/BaseItem.cs
@@ -135,7 +135,7 @@ namespace MediaBrowser.Controller.Entities
         /// </summary>
         /// <value>The LUFS Value.</value>
         [JsonIgnore]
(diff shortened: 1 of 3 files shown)


Shadowghost (approved):

gnattu:
This will also Fixes #11413 as the additional ffprobe for LUFS tags is eliminated.

Bond-009:
@gnattu I'm not sure it does, this didn't spawn ffprobe with the same args, only with `-hide_banner -i`

nielsvanvelzen (approved):
✅ API changes

crobibero (approved):

nielsvanvelzen:
@Bond-009 will you also be making the required changes to remove the `UseReplayGainTags` option in jellyfin-web?

PR #11436, opened 2024-04-25
Require elevation for plugin related endpoints

nielsvanvelzen:
This came up in chat yesterday. We allow reading of installed/available plugins and repositories in addition to allowing both read and write for plugin configuration. This is **unsafe** and something we never encouraged (plugins should only be available in the dashboard).
In case this breaks a plugin (although that is unlikely) it should be fixed in the plugin by adding additional endpoints.

The only plugin related APIs left as open-access are:
- The endpoint to get a plugin image.
- The endpoint to get the HTML of a plugin page

**Changes**

- Require elevation for plugin related endpoints

**Issues**

### DIFF
--- a/Jellyfin.Api/Controllers/DashboardController.cs
+++ b/Jellyfin.Api/Controllers/DashboardController.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using System.Net.Mime;
 using Jellyfin.Api.Attributes;
 using Jellyfin.Api.Models;
+using MediaBrowser.Common.Api;
 using MediaBrowser.Common.Plugins;
 using MediaBrowser.Model.Net;
 using MediaBrowser.Model.Plugins;
@@ -45,9 +46,9 @@ public class DashboardController : BaseJellyfinApiController
     /// <response code="404">Server still loading.</response>
     /// <returns>An <see cref="IEnumerable{ConfigurationPageInfo}"/> with infos about the plugins.</returns>
     [HttpGet("web/ConfigurationPages")]
+    [Authorize(Policy = Policies.RequiresElevation)]
     [ProducesResponseType(StatusCodes.Status200OK)]
     [ProducesResponseType(StatusCodes.Status404NotFound)]
-    [Authorize]
     public ActionResult<IEnumerable<ConfigurationPageInfo>> GetConfigurationPages(
         [FromQuery] bool? enableInMainMenu)
     {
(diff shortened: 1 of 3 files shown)


nielsvanvelzen:
The PR itself is ready, I've only marked it as draft to make sure we discuss this PR (internally) first.

joshuaboniface (approved):

Jellyfin.Api/Controllers/DashboardController.cs
```diff
@@ -45,6 +46,7 @@ public class DashboardController : BaseJellyfinApiController
     /// <response code="404">Server still loading.</response>
     /// <returns>An <see cref="IEnumerable{ConfigurationPageInfo}"/> with infos about the plugins.</returns>
     [HttpGet("web/ConfigurationPages")]
+    [Authorize(Policy = Policies.RequiresElevation)]
     [ProducesResponseType(StatusCodes.Status200OK)]
     [ProducesResponseType(StatusCodes.Status404NotFound)]
     [Authorize]
```
oddstr13, line 52: I'm not familiar with the C# annotation function annotations, but this line looks redundant at a very least?

cvium:
The policy for plugin and package controller should be at the class level

Jellyfin.Api/Controllers/DashboardController.cs
nielsvanvelzen: Good call, it doesn't hurt adding it twice but it is not necessary. Will fix.

Bond-009 (approved):

Jellyfin.Api/Controllers/DashboardController.cs
```diff
@@ -45,9 +46,9 @@ public class DashboardController : BaseJellyfinApiController
     /// <response code="404">Server still loading.</response>
     /// <returns>An <see cref="IEnumerable{ConfigurationPageInfo}"/> with infos about the plugins.</returns>
     [HttpGet("web/ConfigurationPages")]
+    [Authorize(Policy = Policies.RequiresElevation)]
```
cvium, line 49: Does the other en dpoint need to be public?

nielsvanvelzen: Right now, yes. That endpoint is loaded without any authentication in an iframe by the web client. So changing that would need additional work.

cvium: Should we add a note or something?

crobibero (approved):

Jellyfin.Api/Controllers/DashboardController.cs
cvium: Sure

PR #11457, opened 2024-04-28
Improve audio normalization

Bond-009:
* Move calculation of LUFS to a scheduled task as it's pretty slow
* Correctly calculate album LUFS
* Don't try to convert replaygain tags to LUFS values

TODO:
* Add progress to task
* Scan album gain tags
* Make more configurable


MediaBrowser.Controller/Entities/BaseItem.cs
```diff
@@ -137,6 +137,13 @@ protected BaseItem()
         [JsonIgnore]
         public float? LUFS { get; set; }
 
+        /// <summary>
+        /// Gets or sets the gain required for audio normalization.
+        /// </summary>
+        /// <value>The gain required for audio normalization..</value>
```
dmitrylyzo, line 143:
```suggestion
        /// <value>The gain required for audio normalization.</value>
```

MediaBrowser.Model/Dto/BaseItemDto.cs
```diff
@@ -782,10 +782,10 @@ public class BaseItemDto : IHasProviderIds, IItemDto, IHasServerId
         public string TimerId { get; set; }
 
         /// <summary>
-        /// Gets or sets the LUFS value.
+        /// Gets or sets the gain required for audio normalization.
         /// </summary>
-        /// <value>The LUFS Value.</value>
-        public float? LUFS { get; set; }
+        /// <value>The gain required for audio normalization..</value>
```
dmitrylyzo, line 787:
```suggestion
        /// <value>The gain required for audio normalization.</value>
```

Emby.Server.Implementations/Dto/DtoService.cs
```diff
@@ -898,7 +898,14 @@ private void AttachBasicFields(BaseItemDto dto, BaseItem item, BaseItem owner, D
                 dto.IsPlaceHolder = supportsPlaceHolders.IsPlaceHolder;
             }
 
-            dto.LUFS = item.LUFS;
+            if (item.LUFS.HasValue)
+            {
+                dto.NormalizationGain = -18f - item.LUFS;
```
dmitrylyzo, line 903: Maybe it worth to move `DefaultLUFSValue` const you have removed here? Or leave a comment about magic value (`-18f`)?

nielsvanvelzen (approved):
✅ API changes look fine, but please make sure they are merged (together with the required web changes!!) before the next unstable build so we have at least 1 week to find any left-over API mismatches between server and web.

Chaphasilor:
I've tested this with @Bond-009's help and it seems to be working (on Windows at least). We found one bug related to albums with only a single track but that is now fixed.

I've confirmed this works properly and already integrated the changes into Finamp without issues 😁

Shadowghost (approved):

Emby.Server.Implementations/ScheduledTasks/Tasks/AudioNormalizationPostScanTask.cs
```diff
@@ -0,0 +1,198 @@
...
+using MediaBrowser.Controller.MediaEncoding;
+using MediaBrowser.Controller.Persistence;
+using MediaBrowser.Model.Globalization;
+using MediaBrowser.Model.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Emby.Server.Implementations.ScheduledTasks.Tasks;
+
+/// <summary>
+/// The splashscreen post scan task.
```
crobibero, line 24: is it now

Emby.Server.Implementations/ScheduledTasks/Tasks/AudioNormalizationPostScanTask.cs
```diff
@@ -0,0 +1,198 @@
...
+        foreach (var library in _libraryManager.RootFolder.Children)
+        {
+            var libraryOptions = _libraryManager.GetLibraryOptions(library);
+            if (!libraryOptions.EnableLUFSScan)
+            {
+                continue;
+            }
+
+            // Album gain
+            var albums = _libraryManager.GetItemList(new InternalItemsQuery
```
crobibero, line 87: I'm never a fan of pulling all of an item type from the database.

Emby.Server.Implementations/ScheduledTasks/Tasks/AudioNormalizationPostScanTask.cs
```diff
@@ -0,0 +1,198 @@
...
+            {
+                IncludeItemTypes = [BaseItemKind.MusicAlbum],
+                Parent = library,
+                Recursive = true
+            });
+
+            foreach (var a in albums)
+            {
+                if (a.NormalizationGain.HasValue || a.LUFS.HasValue)
```
crobibero, line 96: TODO allow replacing the normalization gain?

Emby.Server.Implementations/ScheduledTasks/Tasks/AudioNormalizationPostScanTask.cs
```diff
@@ -0,0 +1,198 @@
...
+            new TaskTriggerInfo
+            {
+                Type = TaskTriggerInfo.TriggerInterval,
+                IntervalTicks = TimeSpan.FromHours(24).Ticks
+            }
+        ];
+    }
+
+    private string EscapeFilename(string filename)
+        => filename;
```
crobibero, line 156: Why have this function?

Emby.Server.Implementations/ScheduledTasks/Tasks/AudioNormalizationPostScanTask.cs
Bond-009: Yeah, but this can be changed when progress reporting is added

Emby.Server.Implementations/ScheduledTasks/Tasks/AudioNormalizationPostScanTask.cs
Bond-009: Believe it or not, but it actually used to do something :p
But since escaping for albums and tracks is different this isn't used anymore

Emby.Server.Implementations/ScheduledTasks/Tasks/AudioNormalizationPostScanTask.cs
Bond-009: it is now

Emby.Server.Implementations/ScheduledTasks/Tasks/AudioNormalizationTask.cs
```diff
@@ -0,0 +1,195 @@
...
+
+            // Album gain
+            var albums = _libraryManager.GetItemList(new InternalItemsQuery
+            {
+                IncludeItemTypes = [BaseItemKind.MusicAlbum],
+                Parent = library,
+                Recursive = true
+            });
+
+            foreach (var a in albums)
```
crobibero, line 94: Should add a `try..catch` so we don't stop processing on first (or any) failure

Bond-009: There is a try catch inside of `CalculateLUFSAsync` where I'd expect these failures to take place.
Seems unlikely anything would go wrong here, and it it does (like failing to create the concat file) it would likely happen to the other items as well

Emby.Server.Implementations/ScheduledTasks/Tasks/AudioNormalizationPostScanTask.cs
Chaphasilor: it is now?

crobibero (approved):

PR #11472, opened 2024-05-01
Fix missing season Index

### LINKED ISSUE #11470: [Issue]: TV Seasons duplicated with differing settings
### Please describe your bug

Upgrading from 10.8.13 to unstable (2024042905). Some, possibly all, shows are showing duplicate season entries.

Example screenshot: 
image

Curiously, the two season entries for each season are slightly differing. Clicking "Edit Metadata", the first shows:

image

This being especially curious since the add date is the date I upgraded to Unstable.

The second shows a more expected value:

image

### Reproduction Steps

Unknown, upgrade from 10.8.13 to unstable.

### Jellyfin Version

Unstable (master branch)

### if other:

_No response_

### Environment

```markdown
- OS: Debian 12
- Linux Kernel: 6.1
- Virtualization: KVM
- Clients: Web, JMP
- Browser: Firefox
- FFmpeg Version: 6.0.1
- Playback Method: Direct
- Hardware Acceleration: nVidia
- GPU Model: Tesla T4
- Plugins: Multiple
- Reverse Proxy: HAProxy
- Base URL: `/`
- Networking: Default
- Storage: CephFS
```

### Jellyfin logs

```shell
Nothing relevant
```

### FFmpeg logs

```shell
Nothing relevant
```

Nothing relevant

Posted above.

### Code of Conduct

gnattu:
**Changes**

This will add missing season index when it can be guessed from the season name, useful for very old libraries that may contain null index

**Issues**

Fixes #11470

### DIFF
--- a/MediaBrowser.Providers/TV/SeriesMetadataService.cs
+++ b/MediaBrowser.Providers/TV/SeriesMetadataService.cs
@@ -224,7 +224,6 @@ namespace MediaBrowser.Providers.TV
             var seasonNames = series.SeasonNames;
             var seriesChildren = series.GetRecursiveChildren(i => i is Episode || i is Season);
             var seasons = seriesChildren.OfType<Season>().ToList();
-
             var uniqueSeasonNumbers = seriesChildren
                 .OfType<Episode>()
                 .Select(e => e.ParentIndexNumber >= 0 ? e.ParentIndexNumber : null)
--- a/MediaBrowser.Providers/TV/SeriesMetadataService.cs
+++ b/MediaBrowser.Providers/TV/SeriesMetadataService.cs
@@ -208,7 +208,7 @@ namespace MediaBrowser.Providers.TV
 
             foreach (var season in seasons)
             {
-                var hasUpdate = season.BeforeMetadataRefresh(false);
+                var hasUpdate = season.BeforeMetadataRefresh(replaceAllMetadata);
                 if (hasUpdate)
                 {
                     await season.UpdateToRepositoryAsync(ItemUpdateType.MetadataEdit, cancellationToken).ConfigureAwait(false);
--- a/MediaBrowser.Providers/TV/SeriesMetadataService.cs
+++ b/MediaBrowser.Providers/TV/SeriesMetadataService.cs
@@ -40,8 +40,7 @@ namespace MediaBrowser.Providers.TV
         {
             if (item is Series series)
             {
-                var seriesChildren = series.GetRecursiveChildren(i => i is Episode || i is Season);
-                var seasons = seriesChildren.OfType<Season>().ToList();
+                var seasons = series.GetRecursiveChildren(i => i is Season).ToList();
 
                 foreach (var season in seasons)
                 {
--- a/MediaBrowser.Providers/TV/SeriesMetadataService.cs
+++ b/MediaBrowser.Providers/TV/SeriesMetadataService.cs
@@ -8,7 +8,6 @@ using System.Threading;
 using System.Threading.Tasks;
 using MediaBrowser.Controller.Configuration;
 using MediaBrowser.Controller.Dto;
-using MediaBrowser.Controller.Entities;
 using MediaBrowser.Controller.Entities.TV;
 using MediaBrowser.Controller.Library;
 using MediaBrowser.Controller.Providers;
--- a/MediaBrowser.Providers/TV/SeriesMetadataService.cs
+++ b/MediaBrowser.Providers/TV/SeriesMetadataService.cs
@@ -204,6 +204,12 @@ namespace MediaBrowser.Providers.TV
             var seasonNames = series.SeasonNames;
             var seriesChildren = series.GetRecursiveChildren(i => i is Episode || i is Season);
             var seasons = seriesChildren.OfType<Season>().ToList();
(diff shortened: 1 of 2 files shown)


MediaBrowser.Providers/TV/SeriesMetadataService.cs
```diff
@@ -204,6 +204,14 @@ private async Task UpdateAndCreateSeasonsAsync(Series series, CancellationToken
             var seasonNames = series.SeasonNames;
             var seriesChildren = series.GetRecursiveChildren(i => i is Episode || i is Season);
             var seasons = seriesChildren.OfType<Season>().ToList();
+
+            // Very old Jellyfin library will have a cursed DB that a season may have NULL index number
```
TimGels, line 208: Are we not able to add a migration step to change the null index number?

joshuaboniface: At this point, I don't know if we can. It's far too late in the release process, and anyone who's already tested unstable is affected (it's how I found this). Plus, I get the feeling fixing it in scanning is the right move anyways, but I'm not married to that opinion.

TimGels: I am not familiar with the scanner so I am fine with any solution that, someone who is familiar with it, sees as best fitting. I trust that decision. However, I am always a bit worried about implementing bug fixes with work around instead of fixing a root cause. We should avoid adding bug fix/ work around complexity to scanners, when it can be avoided. I also have a feeling that we might be jumping the gun here a bit (just a hunch I do not have a good grasp of the situation).

As for affected users. I personally see unstable as unstable and we should not be too reluctant to have breaking changes inside unstable versions. If something is discovered in the unstable builds, then it should be fixed accordingly, without being _too_ reluctant to break unstable setups. People should know the risk of running unstable and should treat it like so.

gnattu: > fixing a root cause

This is already impossible, can you invent Time machine so that we can go back to 2020 and fix that cursed scanner in Jellyfin 10.4?

TimGels: Then, in my opinion it should be part of a migration step. I am not familiar with this part of the code, but if all this is doing is making sure values are inserted during a scan, why not do that in a migration step so it runs only one time? As after that it wont be needed.

MediaBrowser.Providers/TV/SeriesMetadataService.cs
```diff
@@ -197,13 +197,20 @@ private void DeleteEpisode(Episode episode)
...
         /// <param name="cancellationToken">The cancellation token.</param>
         /// <returns>The async task.</returns>
-        private async Task UpdateAndCreateSeasonsAsync(Series series, CancellationToken cancellationToken)
+        private async Task UpdateAndCreateSeasonsAsync(Series series, bool replaceAllMetadata, CancellationToken cancellationToken)
         {
             var seasonNames = series.SeasonNames;
             var seriesChildren = series.GetRecursiveChildren(i => i is Episode || i is Season);
             var seasons = seriesChildren.OfType<Season>().ToList();
+
+            foreach (var season in from season in seasons let hasUpdate = season.BeforeMetadataRefresh(replaceAllMetadata) where hasUpdate select season)
```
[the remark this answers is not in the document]
gnattu: I can make it a plain if statement, if that is better for readability.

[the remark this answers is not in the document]
Bond-009: Yes please, I don't like mixing LINQ with foreach like this

MediaBrowser.Providers/TV/SeriesMetadataService.cs
```diff
@@ -197,13 +197,24 @@ private void DeleteEpisode(Episode episode)
...
-        private async Task UpdateAndCreateSeasonsAsync(Series series, CancellationToken cancellationToken)
+        private async Task UpdateAndCreateSeasonsAsync(Series series, bool replaceAllMetadata, CancellationToken cancellationToken)
         {
             var seasonNames = series.SeasonNames;
             var seriesChildren = series.GetRecursiveChildren(i => i is Episode || i is Season);
             var seasons = seriesChildren.OfType<Season>().ToList();
+
+            foreach (var season in seasons)
+            {
+                var hasUpdate = season.BeforeMetadataRefresh(replaceAllMetadata);
```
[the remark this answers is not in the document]
gnattu: Addressed. Now SeriesMetadataService will override the `RefreshMetadata` method of `MetadataService` and call `BeforeMetadataRefresh` of series before the base class's `RefreshMetadata`.

Shadowghost (approved):

MediaBrowser.Providers/TV/SeriesMetadataService.cs
```diff
@@ -35,6 +36,25 @@ public class SeriesMetadataService : MetadataService<Series, SeriesInfo>
             _localizationManager = localizationManager;
         }
 
+        public override async Task<ItemUpdateType> RefreshMetadata(BaseItem item, MetadataRefreshOptions refreshOptions, CancellationToken cancellationToken)
+        {
+            if (item is Series series)
+            {
+                var seasons = series.GetRecursiveChildren(i => i is Season).ToList();
```
cvium, line 43: FYI, this is an extra db lookup for every series

cvium (approved):

crobibero (approved):

PR #11489, opened 2024-05-05
Add remuxing support for VP9

gnattu:
**Changes**

Add VP9 as a valid HLS format to enable remuxing. This is useful when audio processing is required, but the VP9 video can be passed as-is to avoid unnecessary video transcoding.

No VP9 encoder is enabled and should not be enabled. AV1 and HEVC should be preferred over VP9 if video transcoding is required.

**Issues**

### DIFF
--- a/Jellyfin.Api/Helpers/DynamicHlsHelper.cs
+++ b/Jellyfin.Api/Helpers/DynamicHlsHelper.cs
@@ -714,6 +714,21 @@ public class DynamicHlsHelper
             return HlsCodecStringHelpers.GetAv1String(profile, level, false, bitDepth);
         }
 
+        // VP9 HLS is for video remuxing only, everything is probed from the original video
+        if (string.Equals(codec, "vp9", StringComparison.OrdinalIgnoreCase))
+        {
+            var width = state.VideoStream.Width ?? 0;
+            var height = state.VideoStream.Height ?? 0;
+            var framerate = state.VideoStream.AverageFrameRate ?? 30;
+            var bitDepth = state.VideoStream.BitDepth ?? 8;
+            return HlsCodecStringHelpers.GetVp9String(
+                width,
+                height,
+                state.VideoStream.PixelFormat,
+                framerate,
+                bitDepth);
+        }
+
         return string.Empty;
     }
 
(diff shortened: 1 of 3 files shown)


nyanmisaka:
How does it perform on HLS.js and Safari?

gnattu:
> How does it perform on HLS.js and Safari?

HLS.js on chrome works fine, firefox untested yet.

Safari is a little bit tricky. Only iOS Safari understands VP9 over HLS with the native HLS player, and the desktop Safari does not play.

nyanmisaka:
> > How does it perform on HLS.js and Safari?
> 
> HLS.js on both Firefox and Chrome works with fmp4 container.
> 
> Firefox can play with TS container as well, but Chrome cannot.
> 
> Safari is a little bit tricky. Only iOS Safari understands VP9 over HLS with the native HLS player through fmp4 container, and the desktop Safari does not play.

Safari 17 has added [AV1 over fMP4](https://bitmovin.com/apple-av1-support) support, it's hard to understand why VP9 has been left behind on desktop Safari. Maybe there's some kind of experimental option that's not enabled by default yet?

gnattu:
> Safari 17 has added [AV1 over fMP4](https://bitmovin.com/apple-av1-support) support, it's hard to understand why VP9 has been left behind on desktop Safari. Maybe there's some kind of experimental option that's not enabled by default yet?

No, not experimental option. It is more likely intentional or not bother to add it. iOS Safari supports this for years if I remembered it correctly, and macOS Safari just never supported this. HLS.js works on desktop Safari though.

nyanmisaka:
> > Safari 17 has added [AV1 over fMP4](https://bitmovin.com/apple-av1-support) support, it's hard to understand why VP9 has been left behind on desktop Safari. Maybe there's some kind of experimental option that's not enabled by default yet?
> 
> No, not experimental option. It is more likely intentional or not bother to add it. iOS Safari supports this for years if I remembered it correctly, and macOS Safari just never supported this. HLS.js works on desktop Safari though.
> 
> Apple just don't want you to use VP9 and they are making this more or less only for Google's YouTube. They did not even announce publicly when VP9 is available in their SoC and only quietly added the support in iOS 14 update, unlike AV1 that they publicly announced the availability when it is added. The VP9 support policy for Apple is probably "not an issue as long as YouTube does not break".

This makes sense. Youtube should have enabled its own fMP4/Dash player for desktop Safari through MSE, but this is still not supported on iOS due to power saving considerations.

crobibero (approved):

nyanmisaka (approved):

Bond-009 (approved):

PR #11492, opened 2024-05-05
Add better audio VBR settings

gnattu:
LAME's VBR mode only has advantage over a certain bitrate range. For very low and very high bitrate, use the ABR mode instead.

aac_at's CVBR mode produces very good quality and is not worse than its TVBR mode in blind testing. Use this mode for convenience.

The ffmpeg native aac encoder will have quality regression with its VBR mode. Always use CBR mode for ffmpeg's native aac encoder.

This also adds `EnableAudioVbrEncoding` to TranscodingProfile, enables the client to selectively disable VBR transcoding when it causes playback problems.

**Changes**

- Use ABR for libmp3lame when VBR mode has its limitation
- Supports aac_at's CVBR mode
- Adds `EnableAudioVbrEncoding` to TranscodingProfile

**Issues**

### DIFF
--- a/Jellyfin.Api/Controllers/AudioController.cs
+++ b/Jellyfin.Api/Controllers/AudioController.cs
@@ -83,6 +83,7 @@ public class AudioController : BaseJellyfinApiController
     /// <param name="videoStreamIndex">Optional. The index of the video stream to use. If omitted the first video stream will be used.</param>
     /// <param name="context">Optional. The <see cref="EncodingContext"/>.</param>
     /// <param name="streamOptions">Optional. The streaming options.</param>
+    /// <param name="enableAudioVbrEncoding">Optional. Whether to enable Audio Encoding.</param>
     /// <response code="200">Audio stream returned.</response>
     /// <returns>A <see cref="FileResult"/> containing the audio file.</returns>
     [HttpGet("{itemId}/stream", Name = "GetAudioStream")]
@@ -138,7 +139,8 @@ public class AudioController : BaseJellyfinApiController
         [FromQuery] int? audioStreamIndex,
         [FromQuery] int? videoStreamIndex,
         [FromQuery] EncodingContext? context,
-        [FromQuery] Dictionary<string, string>? streamOptions)
+        [FromQuery] Dictionary<string, string>? streamOptions,
+        [FromQuery] bool? enableAudioVbrEncoding)
     {
         StreamingRequestDto streamingRequest = new StreamingRequestDto
         {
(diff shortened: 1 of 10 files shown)


MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -2576,8 +2576,9 @@ private static int ScaleBitrate(int bitrate, string inputVideoCodec, string outp
             return 128000 * (outputAudioChannels ?? audioStream.Channels ?? 2);
         }
 
-        public string GetAudioVbrModeParam(string encoder, int bitratePerChannel)
+        public string GetAudioVbrModeParam(string encoder, int bitrate, int channels)
         {
+            var bitratePerChannel = bitrate / channels;
```
crobibero, line 2581:
```suggestion
            var bitratePerChannel = bitrate / Math.Max(channels, 1);
```

Would probably be caught upstream but if it is preventable might as well

Jellyfin.Api/Controllers/AudioController.cs
```diff
@@ -138,7 +139,8 @@ public AudioController(AudioHelper audioHelper)
         [FromQuery] int? audioStreamIndex,
         [FromQuery] int? videoStreamIndex,
         [FromQuery] EncodingContext? context,
-        [FromQuery] Dictionary<string, string>? streamOptions)
+        [FromQuery] Dictionary<string, string>? streamOptions,
+        [FromQuery] bool? enableAudioVbrEncoding)
```
crobibero, line 143:
```suggestion
        [FromQuery] bool enableAudioVbrEncoding = true)
```

I would specify the default value here so it gets pulled into the api spec

crobibero (approved):

nyanmisaka (approved):

MediaBrowser.Controller/MediaEncoding/EncodingJobInfo.cs
```diff
@@ -508,6 +508,14 @@ public string ActualOutputAudioCodec
             }
         }
 
+        public bool EnableAudioVbrEncoding
+        {
+            get
+            {
+                return BaseRequest.EnableAudioVbrEncoding;
+            }
+        }
```
Bond-009, line 517:
```suggestion
        public bool EnableAudioVbrEncoding => BaseRequest.EnableAudioVbrEncoding;
```

Bond-009 (approved):

PR #11500, opened 2024-05-06
Restore 255.255.255.255 discovery for Linux and Windows

### LINKED ISSUE #11498: Auto-discovery causes server to stop responding

```shell
```

### Jellyfin Version

Unstable (master branch)

gnattu:
Actually, this address should be removed because we cannot enforce the listening interface with this address at all. However, it seems that all the clients are not yet ready for this change. For now, restore the ability to respond auto discovery message with 255.255.255.255, and perhaps remove it later when clients are ready for the change.

**Changes**

**Issues**

Fixes #11498


crobibero (approved):

nielsvanvelzen:
This does not solve the issue for me. Same exception as in the issue.

gnattu:
> This does not solve the issue for me. Same exception as in the issue.

Windows only quirk. You have no way to restrict the interface on Windows at all. Added the workaround.

PR #11503, opened 2024-05-06
Avoid creating multiple UdpClient instances for the same IPAddress in AutoDiscoveryHost

### LINKED ISSUE #11498: Auto-discovery causes server to stop responding

nielsvanvelzen:
**Changes**
- Avoid creating multiple UdpClient instances for the same IPAddress in AutoDiscoveryHost

**Issues**

Fixes #11498
Supersedes #11500

### DIFF
--- a/src/Jellyfin.Networking/AutoDiscoveryHost.cs
+++ b/src/Jellyfin.Networking/AutoDiscoveryHost.cs
@@ -110,7 +110,15 @@ public sealed class AutoDiscoveryHost : BackgroundService
                     var text = Encoding.UTF8.GetString(result.Buffer);
                     if (text.Contains("who is JellyfinServer?", StringComparison.OrdinalIgnoreCase))
                     {
-                        await RespondToV2Message(respondAddress, result.RemoteEndPoint, cancellationToken).ConfigureAwait(false);
+                        if (respondAddress.Equals(listenAddress))
+                        {
+                            await RespondToV2Message(udpClient, result.RemoteEndPoint, cancellationToken).ConfigureAwait(false);
+                        }
+                        else
+                        {
+                            using var responseClient = new UdpClient(new IPEndPoint(respondAddress, PortNumber));
+                            await RespondToV2Message(responseClient, result.RemoteEndPoint, cancellationToken).ConfigureAwait(false);
+                        }
                     }
                 }
                 catch (SocketException ex)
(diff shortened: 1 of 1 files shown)


Shadowghost (approved):

crobibero (approved):

gnattu:
No, this won’t work on all OSes, please don’t

src/Jellyfin.Networking/AutoDiscoveryHost.cs
```diff
@@ -141,11 +149,10 @@ private async Task RespondToV2Message(IPAddress responderIp, IPEndPoint endpoint
 
         var response = new ServerDiscoveryInfo(localUrl, _appHost.SystemId, _appHost.FriendlyName);
 
-        using var responder = new UdpClient(new IPEndPoint(responderIp, PortNumber));
         try
         {
             _logger.LogDebug("Sending AutoDiscovery response");
-            await responder
+            await udpClient
```
gnattu, line 155: Won't work on all platforms. At least macOS will prohibit such usage.

gnattu (changes requested):

src/Jellyfin.Networking/AutoDiscoveryHost.cs
cvium: Why?

gnattu: No why. macOS prohibits using udpClient bind to broadcast addresses as a responder to send to unicast addresses because the responder IP is too ambiguous in this case. You have to be explicit the responder IP and specify it.

cvium: I would suggest adding OS specific code then. I don't think going back to 0.0.0.0 is the right choice

gnattu: Why? The interface restriction **never** worked as intended and I personally don't like all the "half-working" implementation that only makes the logic more complicated and actually does nothing because we are still binding to 255.255.255.255 and providing support to respond this address, which inherently binds to all interfaces and does not follow the server interface configuration anyway. OS specific quirks are everywhere:

- You cannot recieve interface scoped broadcasts without binding to 255.255.255.255 first on Linux for some old kernel on Linux
- You cannot prevent Windows to bind to 255.255.255.255

To actually enforce the interface binding, we need to implement a workaround that drop everthing recieved on 255.255.255.255, but this will break almost all current clients.

macOS will work with binding with interface-scoped broadcast addresses, but if we want to keep current behavior and provide support for responding 255.255.255.255, we actually have to bind to 0.0.0.0 to receive broadcasts to 255.255.255.255 on macOS because you cannot bind to 255.255.255.255 along, you have to bind to all addresses to actually receive packet send to "all addresses".

But on all platforms, 255.255.255.255 will listen on all interfaces which already breaks our intended behavior to listen only on configured interfaces, which I don't see any benefit over just listen to all addresses.

PR #11504, opened 2024-05-06
Let AutoDiscoveryHost bind to all addresses

### LINKED ISSUE #11309: [Issue]: Auto Discovery listening address changed in unstable from 10.8.13
### Please describe your bug

I recently moved from running 10.8.13 to unstable to get my AMD hardware acceleration working because I needed newer drivers.  I run Jellyfin in kubernetes, and have some forwarding setup on my LAN to send the magic "Who is JellyfinServer?" packets to the service running in kubernetes on port 7359.  This stopped working in unstable, and I suspect it's due to #10793 which looks to be far more restrictive in the addresses it binds to than the old code it replaced.  This can be verified by running `lsof`:

On 10.8.13, `lsof -nP -iUDP` shows these addresses and ports bound:
```
COMMAND  PID USER   FD   TYPE    DEVICE SIZE/OFF NODE NAME
jellyfin   1 root  318u  IPv4 150056711      0t0  UDP *:7359
jellyfin   1 root  326u  IPv4 150060688      0t0  UDP *:1900
jellyfin   1 root  327u  IPv4 150060689      0t0  UDP *:48770
jellyfin   1 root  328u  IPv4 150060690      0t0  UDP 172.16.0.101:40704
jellyfin   1 root  329u  IPv4 150060691      0t0  UDP 127.0.0.1:45939
```

On unstable, `lsof -nP -iUDP` shows these addresses and ports bound:
```
COMMAND  PID USER   FD   TYPE    DEVICE SIZE/OFF NODE NAME
jellyfin   1 root  447u  IPv4 187993370      0t0  UDP 255.255.255.255:7359
jellyfin   1 root  448u  IPv4 187993371      0t0  UDP 127.255.255.255:7359
jellyfin   1 root  449u  IPv4 187993372      0t0  UDP 172.16.3.255:7359
```

I can run `tcpdump -i any -A 'port 7359' in both instances as well and see the UDP packet come into the Pod, but only see the response go out with 10.8.13.

### Reproduction Steps

1. Launch a Pod of `jellyfin/jellyfin:10.8.13` and define a Service with an IP so you can talk to it from a host machine
2. `echo -n 'Who is JellyfinServer?' | socat - udp-datagram:{service address}:7359,broadcast`
3. See JSON response
4. Repeat (1) and (2) with `jellyfin/jellyfin:unstable` and get no JSON response

### Jellyfin Version

Unstable (master branch)

### if other:

_No response_

### Environment

```markdown
- OS: Host is running Ubuntu 22.04
- Linux Kernel: Host is running `6.5.0-26-generic`
- Virtualization: Kubernetes
- FFmpeg Version: whatever is in `jellyfin/jellyfin` container image
- Plugins: default only
- Reverse Proxy: traefik, but I don't think it matters for this
```

### Jellyfin logs

```shell
n/a
```

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

### LINKED ISSUE #11498: Auto-discovery causes server to stop responding

gnattu:
An alternative to #11500 and #11503

Because we never let auto discovery actually respect the server settings, and most of our clients don't want us to make it more restrictive and only bind to interface-bounding addresses. Just bind to all addresses to make everyone happy.

**Changes**

**Issues**

Fixes #11498
Fixes #11309

### DIFF
--- a/src/Jellyfin.Networking/AutoDiscoveryHost.cs
+++ b/src/Jellyfin.Networking/AutoDiscoveryHost.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Net;
 using System.Net.Sockets;
 using System.Text;
@@ -11,7 +10,6 @@ using MediaBrowser.Common.Configuration;
 using MediaBrowser.Common.Net;
 using MediaBrowser.Controller;
 using MediaBrowser.Model.ApiClient;
-using MediaBrowser.Model.Net;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
 
@@ -51,25 +49,6 @@ public sealed class AutoDiscoveryHost : BackgroundService
         _networkManager = networkManager;
     }
 
-    private static IPAddress GetBindAddress(IPData intf)
-    {
-        if (intf.AddressFamily == AddressFamily.InterNetwork)
-        {
-            if (OperatingSystem.IsLinux())
-            {
-                return NetworkUtils.GetBroadcastAddress(intf.Subnet);
-            }
-
-            if (OperatingSystem.IsMacOS())
-            {
-                // macOS kernel does not allow bind to 127:255:255:255
-                return IPAddress.IsLoopback(intf.Address) ? intf.Address : NetworkUtils.GetBroadcastAddress(intf.Subnet);
-            }
-        }
-
-        return intf.Address;
-    }
-
     /// <inheritdoc />
     protected override async Task ExecuteAsync(CancellationToken stoppingToken)
     {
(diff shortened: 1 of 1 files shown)


nielsvanvelzen:
This "fix" works for me. The exception is gone and client receives discovery replies again. However, that is because you're now re-using the udpClient just like in #11503. And listening on `0.0.0.0` is likely not the behavior @Shadowghost intended.

Shadowghost:
The problem with listening on any is that we now handle broadcast from any subnet too, contrary to just global broadcast and bind subnet specific ones

gnattu:
> However, that is because you're now re-using the udpClient just like in https://github.com/jellyfin/jellyfin/pull/11503.

No, not quite. These are two entirely two different things. Listen to 0.0.0.0 allows the udpClient to be shared on all platforms. On Windows, previously we bind to a unicast address and that one automatically receives broadcast, on non-windows, we have to manually bind the .255 address, we were binding to different addresses, and the former can be shared but the latter cannot be shared on all platforms. Only on windows the udpclient will always use the same responder ip and has to share the client.

> And listening on 0.0.0.0 is likely not the behavior @Shadowghost intended.
> we now handle broadcast from any subnet too, contrary to just global broadcast and bind subnet specific ones

What's the problem? Clients from any subnet can always use 255.255.255.255 and get a response, which is what most clients currently do. Let's not pretend that we're limiting subnet access to configured ones when we aren't. The reality is, we'd have to break almost all clients to enforce this, and if we don't, we're no better off than just listening to all addresses.

cvium (approved):

Shadowghost (approved):
Let's do it like this and replace it with MDNS in the future.

nielsvanvelzen (approved):

PR #11506, opened 2024-05-07
Fix broken hardware encoder and filter for trickplay

gnattu:
The hardware processing pipeline never worked for trickplay, because the required encoder was not added to the validator and the software encoder is always passed to the ProcessingFilter builder, which returns the software filter+encoder.

**Changes**

**Issues**


Shadowghost (commented):
Shouldn't we also add `mjpeg_videotoolbox` and `mjpeg_rkmpp`?

gnattu:
> Shouldn't we also add `mjpeg_videotoolbox` and `mjpeg_rkmpp`?

These additional platforms should go with 10.10 in my opinion, they will require additional code to work.

nyanmisaka (approved):

Shadowghost (approved):

Bond-009 (approved):

PR #11510, opened 2024-05-08
Enable hardware Trickplay processing pipeline for VideoToolbox

gnattu:
**Changes**

- Enable `mjpeg_videotoolbox` encoder
- Adapt the trickplay processing pipeline for VideoToolbox

**Issues**

Requires jellyfin/jellyfin-ffmpeg#388

### DIFF
--- a/MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
+++ b/MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
@@ -929,11 +929,10 @@ namespace MediaBrowser.MediaEncoding.Encoder
 
             if (vidEncoder.Contains("videotoolbox", StringComparison.OrdinalIgnoreCase))
             {
-                // videotoolbox's mjpeg encoder uses jpeg quality instead of ffmpeg defined qscale
+                // videotoolbox's mjpeg encoder uses jpeg quality scaled to QP2LAMBDA (118) instead of ffmpeg defined qscale
                 // ffmpeg qscale is a value from 1-31, with 1 being best quality and 31 being worst
                 // jpeg quality is a value from 0-100, with 0 being worst quality and 100 being best
-                // This mapping is just an estimation, with bias towards higher quality
-                encoderQuality = 100 - ((qualityScale - 1) * 3);
+                encoderQuality = 118 - ((qualityScale - 1) * (118 / 30));
             }
 
             // Output arguments
--- a/MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
+++ b/MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
@@ -142,7 +142,13 @@ namespace MediaBrowser.MediaEncoding.Encoder
         /// <inheritdoc />
         public bool IsVaapiDeviceSupportVulkanDrmInterop => _isVaapiDeviceSupportVulkanDrmInterop;
 
-        public bool IsJellyfinFfmpeg => SupportsFilter("alphasrc");
+        /// <summary>
+        /// Gets a value indicating whether the ffmpeg in use is likely Jellyfin FFmpeg.
+        /// Assume the FFmpeg currently using is Jellyfin ffmpeg if it supports alphasrc filter.
+        /// The alphasrc filter is only included in Jellyfin's own fork and is almost useless
+        /// on its own.
+        /// </summary>
+        public bool LikelyToBeJellyfinFfmpeg => SupportsFilter("alphasrc");
 
         [GeneratedRegex(@"[^\/\\]+?(\.[^\/\\\n.]+)?$")]
         private static partial Regex FfprobePathRegex();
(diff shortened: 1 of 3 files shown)


MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
```diff
@@ -144,6 +144,8 @@ public partial class MediaEncoder : IMediaEncoder, IDisposable
         /// <inheritdoc />
         public bool IsVaapiDeviceSupportVulkanDrmInterop => _isVaapiDeviceSupportVulkanDrmInterop;
 
+        public bool IsJellyfinFfmpeg => SupportsFilter("alphasrc");
```
crobibero, line 147: I don't think `IsJellyfinFfmpeg` is a good name for this since it only tests for `alphasrc`

I don't know what `alphasrc` is, but I assume other ffmpeg builds can provide it

gnattu: >  but I assume other ffmpeg builds can provide it

Very very unlikely because this is our own custom filter. We use this filter with our subtitle burn-in pipeline but this filter is almost useless with its own. If other people decide to port this filter and not port all of our patches then this is totally on them and not us.

gnattu: I changed the name of it and added a lengthy document to this value. The assumption of having `alphasrc` means is Jellyfin ffmpeg is already widely used in the whole transcoding pipeline and there is not bug report regarding this assumption. We should be fine to keep doing it this way.

crobibero (approved):

MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
```diff
@@ -862,6 +870,12 @@ private async Task<string> ExtractImageInternal(string inputPath, string contain
                 inputArg = "-threads " + threads + " " + inputArg; // HW accel args set a different input thread count, only set if disabled
             }
 
+            if (options.HardwareAccelerationType.Contains("videotoolbox", StringComparison.OrdinalIgnoreCase) && LikelyToBeJellyfinFfmpeg)
+            {
+                // VideoToolbox supports low priority decoding, which is useful for trickplay
+                inputArg = "-hwaccel_flags +low_priority " + inputArg;
```
nyanmisaka, line 876: How about we use `./ffmpeg -hide_banner -h full` to query? This only needs to be checked once at startup.
```
  -hwaccel_flags     <flags>      .D.V....... (default ignore_level)
     ignore_level                 .D.V....... ignore level even if the codec level used is unknown or higher than the maximum supported level reported by the hardware driver
     allow_high_depth              .D.V....... allow to output YUV pixel formats with a different chroma sampling than 4:2:0 and/or other than 8 bits per component
     allow_profile_mismatch              .D.V....... attempt to decode anyway if HW accelerated decoder's supported profiles do not exactly match the stream
     unsafe_output                .D.V....... allow potentially unsafe hwaccel frame output that might require special care to process successfully
     low_priority                 .D.V....... attempt to run decode pipeline at a lower priority than is used for realtime decoding
```

nyanmisaka: That's strange. I can find `low_priority` in both linux(gcc) and win(gcc+clang) builds.

These are fixed strings and are usually independent of the host type.
https://github.com/jellyfin/jellyfin-ffmpeg/blob/443d842d1e13723f78d5089280f9f2c589d37fbb/debian/patches/0076-vt-low-priority-keyframe-decoding.patch#L37

gnattu: The whole `-hwaccel_flags` helper text  is missing on macOS

gnattu: It seems like running with `ffmpeg -hwaccel videotoolbox -hwaccel_flags +low_priority -hide_banner -f lavfi -i nullsrc=s=1x1:d=500 -f null -` and check if the command exit with 0 will do it. I prefer this way rather to parse an 1MB long text output.

nyanmisaka: This is a good trick. All it takes is a fake input to let ffmpeg parse whether the input options are correct.

nyanmisaka: Changing to `-hwaccel auto` can make it more versatile, especially if other hwaccels can adopt this flag in the future.

gnattu: I think the `-hwaccel` can be omitted entirely as ffmpeg will not really setup the hardware but only parse the input option.

MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
```diff
@@ -142,6 +143,14 @@ public partial class MediaEncoder : IMediaEncoder, IDisposable
...
         public bool IsVaapiDeviceSupportVulkanDrmInterop => _isVaapiDeviceSupportVulkanDrmInterop;
 
+        /// <summary>
+        /// Gets a value indicating whether the ffmpeg in use is likely Jellyfin FFmpeg.
+        /// Assume the FFmpeg currently using is Jellyfin ffmpeg if it supports alphasrc filter.
+        /// The alphasrc filter is only included in Jellyfin's own fork and is almost useless
+        /// on its own.
+        /// </summary>
+        public bool LikelyToBeJellyfinFfmpeg => SupportsFilter("alphasrc");
+
```
nyanmisaka, line 153:
```suggestion
```

nyanmisaka (approved):

crobibero (approved):

Bond-009 (approved):

PR #11512, opened 2024-05-08
Adds sorting by premier dates to the /Items/Latest api endpoint

scampower3:
**Changes**

Adds sorting by premier dates to the /Items/Latest api endpoint
Also adds NewReleases enum to HomeSectionType and SortBy to LatestItemsQuery
For: https://github.com/jellyfin/jellyfin-web/pull/5464
Supersedes: https://github.com/jellyfin/jellyfin/pull/10443
**Issues**


Jellyfin.Data/Enums/HomeSectionType.cs
```diff
@@ -53,6 +53,11 @@ public enum HomeSectionType
         /// <summary>
         /// Continue Reading.
         /// </summary>
-        ResumeBook = 9
+        ResumeBook = 9,
+
+        /// <summary>
+        /// Latest Premier.
+        /// </summary>
+        LatestPremier = 10
```
thornbill, line 61: Maybe "NewReleases"? `LatestMedia` was a really terrible name considering how ambiguous "latest" is...

thornbill:
I'm curious if we really need to use the `Latest` API vs just using the `Items` API... I believe the only real difference is how it handles episodes (returning individual episodes if only one is new and the show if more than one is new) and I'm not sure that is functionality we want to continue using. :man_shrugging:

scampower3:
Tbh, I rather keep that functionality. IMO, it looks kinda bad when a show has a multiple episodes premiering on the same day and the homepage displays each episode contiguously. To me, it looks better when the episodes are grouped together

thornbill:
Right. I was proposing always showing the season even when there is a single new episode... but I was probably too brief. 🙂

scampower3:
Personally, I prefer showing single episode when there is only one new episode.

JPVenson:
closed due to inactivity

Issue #11502, opened 2024-05-04
Chromecast no longer appears in casting list(possible due to missing options in user Playback settings)


soultaco83:
**Describe The Bug**
Chromecast does not appear in list of devices when attempting to cast

**Steps To Reproduce**

1. Go to top right
2. Click on Cast to Device button
3. Only device showing is the browser window

**Expected Behavior**

**Logs**
Chrome Console: 
![image](https://github.com/jellyfin/jellyfin-web/assets/76927195/6ab0f201-549a-41a5-b3a6-1a46b0ec4842)
No jellyfin logs. 

**Screenshots**
![image](https://github.com/jellyfin/jellyfin-web/assets/76927195/1064663b-69c2-4f34-b181-466d01414727)
Option missing in user playback setting

**System (please complete the following information):**
 - Platform: Windows
 - Browser: Chrome and Edge
 - Jellyfin Version: 10.9

**Additional Context**

davidfdezalcoba:
I can confirm I had this same issue after upgrading to unstable and had to manually add the lines in `system.xml` file

crobibero:
My unstable server does not have this issue.

For those with the issue, can you check `migrations.xml` to see if the following migration was applied?
```xml
<ValueTupleOfGuidString>
  <Item1>34a1a1c4-5572-418e-a2f8-32cdfe2668e8</Item1>
  <Item2>AddDefaultCastReceivers</Item2>
</ValueTupleOfGuidString>
```

thornbill:
I do have that entry in my migrations file.

soultaco83:
Sorry took a minute. I can confirm I have this in my migrations.xml file as well

nielsvanvelzen:
My best guess is that the server was downgraded to an earlier 10.9 (or 10.8) version after the migration was applied. That would've removed the new configuration options and upgrading afterwards would never execute the migration again.

joshuaboniface:
Can confirm here at least that mine are properly applied, having upgraded about a month-and-a-half ago from 10.8.13 to 10.9.0 unstables, so I'd definitely agree it's a weird edge case. I think we're OK unless the issue becomes more widespread.

ziogref:
I was missing all entries for   <CastReceiverApplications>  in my system xml file and added

My system.xml file now looks like

`  <ParallelImageEncodingLimit>0</ParallelImageEncodingLimit>
  <CastReceiverApplications>
   <CastReceiverApplication>
[...]

PR #11516, opened 2024-05-09
Always set cast receivers during migration

crobibero:
Might prevent https://github.com/jellyfin/jellyfin/issues/11502 though I haven't been able to reproduce the issue

### DIFF
--- a/Jellyfin.Server/Migrations/Routines/AddDefaultCastReceivers.cs
+++ b/Jellyfin.Server/Migrations/Routines/AddDefaultCastReceivers.cs
@@ -32,24 +32,20 @@ public class AddDefaultCastReceivers : IMigrationRoutine
     /// <inheritdoc />
     public void Perform()
     {
-        // Only add if receiver list is empty.
-        if (_serverConfigurationManager.Configuration.CastReceiverApplications.Length == 0)
-        {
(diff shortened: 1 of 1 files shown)


gnattu:
A dirtier but maybe more reliable workaround might be just set this as the default value, instead of setting it during migration. I think it is very likely that clients can overwrite this during configuration update and accidentally reset it to the default (empty).

crobibero:
> A dirtier but maybe more reliable workaround might be just set this as the default value, instead of setting it during migration. I think it is very likely that clients can overwrite this during configuration update and accidentally reset it to the default (empty).

This would prevent a user from completely removing reference to our cast receivers

gnattu:
> > A dirtier but maybe more reliable workaround might be just set this as the default value, instead of setting it during migration. I think it is very likely that clients can overwrite this during configuration update and accidentally reset it to the default (empty).
> 
> This would prevent a user from completely removing reference to our cast receivers

Would adding an option to the web like "reset cast devices" which restore the default receivers list on the web be a good workaround then? Such accidents do occur and the user might be unaware of the actual reason, and only want to make it working again. Manually edit the xml file in this case is very undesirable.

Bond-009 (approved):

nielsvanvelzen (approved):

PR #11520, opened 2024-05-09
Allow absolute path mismatch when getting virtual folder info

gnattu:
We have historically required the path stored in the database to exactly match the actual path on the filesystem to retrieve virtual folder information, dating back to the Emby era. However, this approach can cause issues when symbolic links are involved, as the link path may differ from the actual pointed path. To address this, allow matching based solely on the folder name when absolute path checking fails, mitigating such problems.

I have no idea why this works previously for 10.8 as @joshuaboniface said. The mismatching path should never worked as the code is from emby. 

**Changes**

- allow matching based solely on the folder name when absolute path checking fails for `GetVirtualFolders`

**Issues**

Fixes jellyfin/jellyfin-web#4945 Although I don't understand why it worked for 10.8

### DIFF
--- a/Emby.Server.Implementations/Library/LibraryManager.cs
+++ b/Emby.Server.Implementations/Library/LibraryManager.cs
@@ -1152,11 +1152,11 @@ namespace Emby.Server.Implementations.Library
             var refreshQueue = includeRefreshState ? ProviderManager.GetRefreshQueue() : null;
 
             return _fileSystem.GetDirectoryPaths(_configurationManager.ApplicationPaths.DefaultUserViewsPath)
-                .Select(dir => GetVirtualFolderInfo(dir, topLibraryFolders, refreshQueue))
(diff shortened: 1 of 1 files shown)


Shadowghost (approved):

Emby.Server.Implementations/Library/LibraryManager.cs
```diff
@@ -1184,6 +1184,12 @@ private VirtualFolderInfo GetVirtualFolderInfo(string dir, List<BaseItem> allCol
             };
 
             var libraryFolder = allCollectionFolders.FirstOrDefault(i => string.Equals(i.Path, dir, StringComparison.OrdinalIgnoreCase));
+            if (libraryFolder is null && allowAbsolutePathMismatch)
+            {
+                _logger.LogWarning("Absolute path does not match any virtual folder, guessing from folder name");
```
TimGels, line 1189: Does it show somewhere in the log which path this is warning about?

gnattu:
Closing for now as it seems not needed.

PR #11523, opened 2024-05-09
Default to not extract lyrics

gnattu:
Such operations require filesystem write permissions and should not be set as default. Users should be required to explicitly enable them.

**Changes**

Default value for  `SaveLyricsWithMedia` is now`false`

The `SaveSubtitlesWithMedia` should also default to `false`, but this change is postponed to 10.10

**Issues**

### DIFF
--- a/MediaBrowser.Model/Configuration/LibraryOptions.cs
+++ b/MediaBrowser.Model/Configuration/LibraryOptions.cs
@@ -20,8 +20,8 @@ namespace MediaBrowser.Model.Configuration
 
             AutomaticallyAddToCollection = false;
             EnablePhotos = true;
-            SaveSubtitlesWithMedia = true;
-            SaveLyricsWithMedia = true;
+            SaveSubtitlesWithMedia = false;
+            SaveLyricsWithMedia = false;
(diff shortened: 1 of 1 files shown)


crobibero (approved):

Shadowghost (approved):

nielsvanvelzen:
The `SaveSubtitlesWithMedia` option has been set to `true` since we forked. Changing this now is not a "regression" fix and should therefor be postponed to 10.10.

The `SaveLyricsWithMedia` option is new in 10.9 and the default behavior was mirrored from the subtitle one. While I understand that it will automatically enable the option for existing libraries, it's not that difficult to manually disable. I'd agree with changing this option for 10.9 though.

However, I don't see a pull request for jellyfin-web. Which must also be updated with the new defaults.

jumoog:
Maybe leave it `true` and disable it when write errors occur?

gnattu:
> Maybe leave it `true` and disable it when write errors occur?

This is not for error handling. Enabling such option by default should never happen in the first place. Functions that alter the media folder should not be enabled by default, and it was a mistake to do so for the past 5 years.

PR #11542, opened 2024-05-11
Add Systemd Socket Activation

rektide:
**Changes**

_What:_ Support Systemd Socket Activation. Listen on all sockets passed in at startup whenever `SYSTEMD_FDS` is specified.

_How:_ When setting up Kestrel, make use of [Microsoft.AspNetCore.Hosting.KestrelServerOptionsSystemdExtensions#UseSystemd](https://github.com/dotnet/aspnetcore/blob/main/src/Servers/Kestrel/Core/src/Systemd/KestrelServerOptionsSystemdExtensions.cs#L37) which seems safe to run anywhere & will only try to listen on passed-in sockets if the appropriate env vars are set.

_Why:_ Having Jellyfin be able to start & stop on demand would lower resource and power consumption for many users when not running Jellyfin. I'm not 100% but I suspect that, for example, having Jellyfin open keeps [systemd automounts](https://www.freedesktop.org/software/systemd/man/latest/systemd.automount.html) from operating as desired, keeping drives mounted & keeping them from fully powering down. This consumes significant power and adds noise to my home that I would rather not have. Supporting socket activation would be the easiest way to have Jellyfin start only as needed, for me.

_Compatibility:_ Should work on any ASP.NET Core since 2.0 (August 2017). References:
* https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.hosting.kestrelserveroptionssystemdextensions.usesystemd?view=aspnetcore-2.0#microsoft-aspnetcore-hosting-kestrelserveroptionssystemdextensions-usesystemd(microsoft-aspnetcore-server-kestrel-core-kestrelserveroptions)
* https://github.com/dotnet/aspnetcore/commit/2351c1b558fc6f4312af262aa866fd2c5164761f

_Links:_

For more information on Systemd Socket Activation:
* http://0pointer.de/blog/projects/socket-activation.html
* https://mgdm.net/weblog/systemd-socket-activation/

This covers a bunch of other ground, but I was happy to see a solid writeup on Dotnet and Systemd here: https://devblogs.microsoft.com/dotnet/net-core-and-systemd/

**Issues**

Satisfies https://features.jellyfin.org/posts/2201/systemd-socket-activation-support

**Gotchas**

I tend-to-think/guess-that if systemd socket activation happens on the port that Jellyfin expects to use, Jellyfin is likely to fail to start (the code above this addition will fail). The theory is that Systemd will have already opened the port, to pass it in, and then Jellyfin will try and fail to open the same port. This gotcha needs validation/testing, but seems likely to me.

My plan is to run systemd socket activation on a separate port than what Jellyfin is configured for, either telling Jellyfin to use a different port & binding on 8096 or just using a different port for socket activation & directing clients to that port (ex: 8095). 

Ideally/in-the-future we might also have a way to tell Jellyfin not to try to listen on any TCP/IP ports, using only whatever unix sockets and/or socket activation sockets are available. (This maybe somehow is possible with the current `NetManager` in this file, but I am not well acquainted & have not thoroughly investigated.)

Another (weirder) possible option is  that both sides (systemd and jellyfin) could try to set `SO_REUSEPORT` on the sockets, so that _both_ can open the same port. It doesn't really matter which of the two sockets sees the incoming connection; the same Jellyfin handlers run on both. This seems like a pretty weird option to juggle to workaround this problem! There's a quirky but academically interesting upside to this route, which is that it could help make non-disruptive in-place Jellyfin upgrades possible by allowing the running two separate instances at once; existing connections could keep being serviced by the old Jellyfin instance while a new upgraded instance could be started. To really be useful, the old instance would need a way to be told to stop listening for connections. There's probably not a ton of user demand for this feature set, but it's some pretty neat system/sockets lore that is pretty neat, and it could well be a very low-code path to working around this gotcha.

### DIFF
--- a/Jellyfin.Server/Extensions/WebHostBuilderExtensions.cs
+++ b/Jellyfin.Server/Extensions/WebHostBuilderExtensions.cs
@@ -83,6 +83,9 @@ public static class WebHostBuilderExtensions
                     options.ListenUnixSocket(socketPath);
                     logger.LogInformation("Kestrel listening to unix socket {SocketPath}", socketPath);
                 }
+
+                // look for LISTEN_FDS and listen on those sockets
+                KestrelServerOptionsSystemdExtensions.UseSystemd(options);
             })
             .UseStartup(_ => new Startup(appHost));
     }


rektide:
Currently WIP because I haven't done any .NET dev in well over a decade & have no idea what I'm doing anymore, haven't tested.

joshuaboniface:
I concur with @JPVenson, I'd like to get more detail on the "why" of this, i.e. what problems does it actually solve, and why would socket activation inclusion in the main code be the best solution to those problems.

Also worth remembering that there would need to be packaging changes to possibly enable this, which is also more maintenance burden.

eriksjolund:
Podman supports [_socket activation of containers_](https://github.com/containers/podman/blob/main/docs/tutorials/socket_activation.md#socket-activation-of-containers).
Docker does not yet have support for it. 

I think main advantages of adding _socket activation_ support to Jellyfin are 

* possibilty to use `--network=none`.  This improves security. For details, see my blog post https://www.redhat.com/sysadmin/socket-activation-podman
* source address is preserved in incoming connections when using a custom network (i.e. `podman create mynet`) and Pasta (network driver) and rootless Podman. I wrote about it here: https://github.com/eriksjolund/podman-networking-docs?tab=readme-ov-file#source-address-preserved
* When using rootless Podman, network traffic is normally passed through _slirp4netns_ or _pasta_. This comes with a performance penalty. Fortunately, communication over the socket-activated socket does not pass through slirp4netns/pasta so it has the same performance characteristics as the normal network on the host.


A possible future  advantage could be that Jellyfin would serve the internet on-the-demand. But for that to work  Jellyfin would need to exit when there is no activity.

JPVenson:
>  (Here I assume that Jellyfin does not need to create outgoing connections)

Well yes, that is my point. It does for Metadata/Image/Plugin lookups.

> source address is preserved in incoming connections when using a custom network (i.e. podman create mynet) and Pasta (network driver) and rootless Podman

Usually thats solved by using a reverse proxy, which we heighly recommend for network safety anyway.

>  performance penalty

Could you provide some data on that? Jellyfin is not relying on realtime data transfer so latency is not really that much of a concern if something arives 10-20 ms later then it could be, it will have no noticable impact on the experience.

eriksjolund:
> 
> > performance penalty
> 
> Could you provide some data on that? Jellyfin is not relying on realtime data transfer so latency is not really that much of a concern if something arives 10-20 ms later then it could be, it will have no noticable impact on the experience.

I don't have any specific numbers but I would like to do some general measurements (not related to jellyfin) that would compare "socket activation" with pasta. I have an idea to use iperf3 that is built with this PR

* https://github.com/esnet/iperf/pull/1171

to do the measurements. This is on my todo-list.

theacodes:
@JPVenson thank you for responding, I appreciate it!

To answer the "but why" it's largely to reduce the amount of system resources used when Jellyfin isn't actually being used by anyone. Lots of us have relatively weak hosts that are juggling a handful of self-hosted services, so this kind of optimization is useful, though not critical.

In terms of how socket activation would impact Jellyfin's scheduled tasks and structure - that is totally valid. I'm not super familiar with Jellyfin's codebase so I can't comment specifically for Jellyfin's case, but I can speak to the general case.

The cleanest approach would be to have Jellyfin work as two processes - one for handling user interaction and streaming, one for handing all the background work. The first one ("server") can be socket-activated and can linger around a bit after disconnection to keep things warmed up. The second one ("worker") could be triggered by [systemd.timer](https://www.freedesktop.org/software/systemd/man/latest/systemd.timer.html) for periodic tasks and [systemd.path](https://www.freedesktop.org/software/systemd/man/latest/systemd.path.html) for reacting to media library changes.

A less invasive approach might be to keep the single jellyfin process, but have `systemd.timer` & `systemd.path` run a small script that sends a request to jellyfin asking it to run some task, waits for it to finish, and then exits. Socket activation would cause the jellyfin service to be started if it wasn't already running, and having the script wait for the task to finish will keep the socket (& service) alive until jellyfin is done.

I hope that helps! Totally understand if this is more than y'all wanna deal with right now or if it's just not a priority. Thanks for all your work. :)

theacodes:
Got it, thanks for considering it and explaining. :)

JPVenson:
> jellyfin is "online" before it actually finishes starting up

This will be solved after https://github.com/jellyfin/jellyfin/pull/12880

We still dont see why we should enable a feature, that has no benefits and adds additional maintaince and support effort from our side, without having a single reason that does not conflict with JFs core functions.

rektide:
> Also worth remembering that there would need to be packaging changes to possibly enable this, which is also more maintenance burden.

My hope was that using standard Microsoft libraries wouldn't add any imposition, wouldn't have any packaging complexities. The library here comes from `Microsoft.AspNetCore.Hosting`. Does C# have have any relatively straightforward/canonical/accepted way to load a library only if present, to allow not needing ASP.NET Core if it's not on the system?

I agree that Jellyfin's loading is not ideal for on-demand socket-activation flows right now. But letting the OS or the container system manage my sockets rather than leaving it to programs to fully configure listening sockets is too powerful and convenient for me to give up. Just the benefit of not needing to run Jellyfin 95% of the time makes it totally worthwhile for me to accept listening sockets when provided.

Bond-009:
Closing as not planned

PR #11569, opened 2024-05-12
Default to processor count concurrent scan instead of 2 * processor count

crobibero:


thornbill:
We will also need to update the placeholder value in the web settings.

MediaBrowser.Controller/Entities/Folder.cs
```diff
@@ -593,8 +593,7 @@ void UpdateProgress()
                 progress.Report(childrenProgress.Average());
             }
 
-            var fanoutConcurrency = ConfigurationManager.Configuration.LibraryScanFanoutConcurrency;
-            var parallelism = fanoutConcurrency > 0 ? fanoutConcurrency : 2 * Environment.ProcessorCount;
+            var parallelism = Math.Max(ConfigurationManager.Configuration.LibraryScanFanoutConcurrency, 1);
```
thornbill, line 596: This completely removes the ability to set the value to be based on processor count... is that desired? The validation and descriptions in web will need updated if so. 🙃

cvium:
This isn't a 10.9 feature iirc. Isn't defaulting to 1 a step backwards?

MediaBrowser.Controller/Entities/Folder.cs
crobibero: Our database can't handle that much concurrency, so I think having it based on processor count won't actually work

crobibero:
> This isn't a 10.9 feature iirc. Isn't defaulting to 1 a step backwards?

Before 10.9 it was defaulted to just ProcessorCount (https://github.com/jellyfin/jellyfin/commit/3982b0e057cc692875ba35573197b86abf8d7adb#diff-41e6711494c9c6ceb7d6aab39d16245177564b0711e402cb2c757d37c3b6815eL589)

We can change the default back to that if desired

Bond-009:
> > This isn't a 10.9 feature iirc. Isn't defaulting to 1 a step backwards?
> 
> 
> 
> Before 10.9 it was defaulted to just ProcessorCount (https://github.com/jellyfin/jellyfin/commit/3982b0e057cc692875ba35573197b86abf8d7adb#diff-41e6711494c9c6ceb7d6aab39d16245177564b0711e402cb2c757d37c3b6815eL589)
> 
> 
> 
> We can change the default back to that if desired

Yes I think that's the correct solution here.
I changed it as it worked fine for me, but I only have 6 cores in my server and no SMT.

Shadowghost (approved):

Bond-009 (approved):

nielsvanvelzen (approved):

PR #11570, opened 2024-05-12
Fix absolute path checking on windows

gnattu:
The fragile path matching no longer works with dotnet 8, causing Windows users to lose collection items because the item path is incorrectly validated, then the server will remove all items due to file not found.

**Changes**

**Issues**

See forum https://forum.jellyfin.org/t-items-in-collections-all-gone-after-10-9-0-upgrade?pid=22471#pid22471

### DIFF
--- a/tests/Jellyfin.Server.Implementations.Tests/IO/ManagedFileSystemTests.cs
+++ b/tests/Jellyfin.Server.Implementations.Tests/IO/ManagedFileSystemTests.cs
@@ -20,7 +20,7 @@ namespace Jellyfin.Server.Implementations.Tests.IO
             _sut = _fixture.Create<ManagedFileSystem>();
         }
 
-        [Theory]
+        [SkippableTheory]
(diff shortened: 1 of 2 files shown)


crobibero:
Should add tests for these cases

Shadowghost (approved):

tests/Jellyfin.Server.Implementations.Tests/IO/ManagedFileSystemTests.cs
```diff
@@ -24,22 +24,39 @@ public ManagedFileSystemTests()
...
-                Assert.Equal(expectedWindowsPath, generatedPath.Split(':')[1]);
+                return;
             }
-            else
+
+            var generatedPath = _sut.MakeAbsolutePath(folderPath, filePath);
+            Assert.Equal(expectedAbsolutePath, generatedPath);
+        }
+
+        [Theory]
```
Bond-009, line 42: Use `SkippableTheoryAttribute`

See `GetFileInfo_DanglingSymlink_ExistsFalse` test below on how it's used

Bond-009 (changes requested):

Bond-009 (approved):

crobibero (approved):

nielsvanvelzen (approved):

zotabee:
Thank you so much!!!

Collections are back, without restoring a backup. Great job!

PR #11103, opened 2024-03-04
Fix spelling

jsoref:
**Changes**
Fixes misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

The misspellings have been reported at https://github.com/jsoref/jellyfin/actions/runs/8136096545#summary-22231815791

The action will report that the changes in this PR would make it happy: https://github.com/jsoref/jellyfin/actions/runs/8136096761#summary-22231816255

**Issues**

PR #11600, opened 2024-05-13
Sp/gr

systemcrash:
https://www.dictionary.com/browse/supersede

Latin root: sedere - to sit (compare: sedentary)

Consider: cede: to give up or yield; supercede; to super give up?

We're also not in the car business, I hope.


oddstr13 (changes requested):
This at a minimum requires testing whether migrations are required for the plugin manifests, and depending on how a unknown value is handled, I think it might be required.

I am not sure of how and if clients are using this enum, but it is likely that at least jellyfin-web needs to be updated for this change.

Further, I would like to point out that supercede seems to be fairly widely used, even if often considered a misspelling, and has according to [Wikitionary][1] been in use in English since the 1400s. [Merriam-Webster][2] lists it as a spelling variant, albeit disputed. [Dictionary.com][3] also has a entry, if only to point it out as *«a frequent misspelling of supersede»*. It is also listed as a synonym in other online dictionaries, such as [Vocabulary.com][4].

Due to the above points, I question the usefulness of this change

[1]: https://en.wiktionary.org/wiki/supercede
[2]: https://www.merriam-webster.com/dictionary/supercede
[3]: https://www.dictionary.com/browse/supercede
[4]: https://www.vocabulary.com/dictionary/supersede

holow29:
`UpdatePluginSuperceedStatus` is clearly wrong and should match the actual enum value name which is currently `PluginStatus.Superceded`. I'm not sure if supercede vs. supsersede matters too much, as you point out, but I would think the function should at least match the enum value name.

holow29:
I think at the least that the function name `UpdatePluginSuperceedStatus` should at least match the enum value name `PluginStatus.Superceded`. Superceed is wrong no matter what one thinks of supercede vs. supersede and it is not uniform in the code.

MediaBrowser.Model/Plugins/PluginStatus.cs
```diff
@@ -34,9 +34,9 @@ public enum PluginStatus
         Malfunctioned = -3,
 
         /// <summary>
-        /// This plugin has been superceded by another version.
+        /// This plugin has been superseded by another version.
         /// </summary>
-        Superceded = -4,
+        Superseded = -4,
```
Bond-009, line 39: This breaks the API

Bond-009 (changes requested):

Bond-009:
Superseded by #11103

PR #11603, opened 2024-05-13
Add OpenAPI spec generator for stable releases

joshuaboniface:
**Changes**
Adds a stable publish action which runs on new tags and pushes the spec to the repository server. Uses all the same logic as Unstable specs but with the correct paths.

Now, I'm not 100% sure if this being on master will let it run on the stable branch. I think it does but not 100%. If not we can forward-port cherry-pick it into the release branch.

**Issues**
N/A

### DIFF
--- a/.github/workflows/ci-openapi.yml
+++ b/.github/workflows/ci-openapi.yml
@@ -140,7 +140,7 @@ jobs:
 
             No changes to OpenAPI specification found. See history of this comment for previous changes.
 
-  publish_unstable:
+  publish-unstable:
     name: OpenAPI - Publish Unstable Spec
     if: |
       github.event_name != 'pull_request_target' && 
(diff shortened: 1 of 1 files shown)


crobibero (approved):

Shadowghost (approved):

.github/workflows/ci-openapi.yml
```diff
@@ -138,10 +140,11 @@ jobs:
 
             No changes to OpenAPI specification found. See history of this comment for previous changes.
 
-  publish:
+  publish_unstable:
```
nielsvanvelzen, line 143:
```suggestion
  publish-unstable:
```

our workflows use dashes for job names

nielsvanvelzen (approved):

PR #11610, opened 2024-05-13
Add helm chart for k8s deployment

jlpedrosa:
**Changes**
* Add helm chart to deploy Jellyfin in k8s clusters
  - Supports volumes for config and media (statically or dynamically provisioned)
  - Based on deployment (not statefulset)
  - Supports ingress

* Release process needs to be addressed in a separate PR, as there are many ways to do this process.  (I commit to do so)
* Documentation to Jellyfin.org (I commit to do so)

**Issues**
Discussion ref about this "feature"/deployment option: 
https://github.com/jellyfin/jellyfin/discussions/11180

cc @joshuaboniface

### DIFF
--- a/charts/jellyfin/Chart.yaml
+++ b/charts/jellyfin/Chart.yaml
@@ -1,0 +1,12 @@
+apiVersion: v2
+appVersion: "10.9.0"
+description: Jellyfin Media Server
+name: jellyfin
+version: 10.9.0
+type: application
+home: https://jellyfin.org/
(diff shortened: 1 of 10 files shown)


joshuaboniface:
I am OK with this feature, but it shouldn't be in the main code repo. It should go into https://github.com/jellyfin/jellyfin-packaging or, possibly, its own repo.

TimGels:
@joshuaboniface I assume this PR can be closed?

Issue #11640, opened 2024-05-14
[Issue]: Transcoding job killed before playback starts.


kevincox:
### Please describe your bug

When playing a transcoded video on a chromecast the ffmpeg process is killed before the video starts playing.

### Reproduction Steps

1. Cast a h265 video from the Android app.
2. Wait for it to start playing.

### Jellyfin Version

10.9.0

### if other:

10.9.1

### Environment

```markdown
- OS: Linux
- Linux Kernel: 6.6.30
- Virtualization: None
- Clients: Jellyfin Android
- Browser: N/A
- FFmpeg Version: jellyfin-ffmpeg 6.0.1
- Playback Method: Transcode
- Hardware Acceleration: AMD AMF
- GPU Model: N/A
- Plugins: None
- Reverse Proxy: nginx
- Base URL: /
- Networking: ?
- Storage: N/A
```

### Jellyfin logs

```shell
nginx: {"cache_status":"","connection_id":3813,"connection_request_count":9,"duration_s":0.001,"protocol":"HTTP/1.1","remote_addr":"::ffff:192.168.2.69","req_bytes":609,"req_content_length":"","req_content_type":"","req_host":"jellyfin.example","req_method":"OPTIONS","req_pipelined":".","req_referrer":"https://apps.jellyfin.org/","req_uri":"/Sessions/Playing/Ping?playSessionId=4122ccf6510b40f79ceb27509411005d","req_via":"","req_user":"","req_user_agent":"Mozilla/5.0 (X11; Linux armv7l) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/90.0.4430.225 Safari/537.36 CrKey/1.56.500000","res_bytes":264,"res_bytes_body":0,"res_content_length":"","res_content_type":"","res_status":204,"res_gzip_ratio":"","ssl_session_reused":"r","time":"2024-05-14T18:59:23+00:00","http2":"","uri":"https://jellyfin.example/Sessions/Playing/Ping?playSessionId=4122ccf6510b40f79ceb27509411005d"}
[...]

PR #11643, opened 2024-05-14
Raise transcoding kill timeout.

kevincox:
**Changes**
The Chromecast client only pings every 10 seconds, so having a 10 second timeout will result in spurious killing any time there is slight deviation in the polling. Set to a much safer value so that transcoding will only get killed if there is a serious hang somewhere.

**Issues**
Fixes: https://github.com/jellyfin/jellyfin/issues/11640

### DIFF
--- a/MediaBrowser.MediaEncoding/Transcoding/TranscodeManager.cs
+++ b/MediaBrowser.MediaEncoding/Transcoding/TranscodeManager.cs
@@ -147,14 +147,7 @@ public sealed class TranscodeManager : ITranscodeManager, IDisposable
             return;
         }
 
-        var timerDuration = 10000;
-
-        if (job.Type != TranscodingJobType.Progressive)
-        {
-            timerDuration = 60000;
-        }
-
-        job.PingTimeout = timerDuration;
+        job.PingTimeout = 60_000;
         job.LastPingDate = DateTime.UtcNow;
 
         // Don't start the timer for playback checkins with progressive streaming


kevincox:
This timeout appears to be introduced by @crobibero in https://github.com/jellyfin/jellyfin/pull/3460.

gnattu:
I believe this is not the problem if your playback does not even start

kevincox:
It isn't the main problem, but it is a problem. The fact is that 10s pings with a 10s timeout isn't going to work reliably.

gnattu:
> It isn't the main problem, but it is a problem. The fact is that 10s pings with a 10s timeout isn't going to work reliably.

Except we are not:

https://github.com/jellyfin/jellyfin-chromecast/blob/b8f75fcf5c1a21633932273ed64eef67b32faac4/src/components/jellyfinActions.ts#L158

On chromecast we are pinging at 5s.

kevincox:
That isn't what I am seeing in the nginx access logs: https://github.com/jellyfin/jellyfin/files/15312949/jellyfin.txt

kevincox:
Trying the Unstable channel the pings seem to be usually a bit over 5s, but just copying the last few I did notice a missed one at 11s. Even then a bit more buffer would be good.

- 23:54:07
- 23:54:13 (6s)
- 23:54:18  (5s)
- 23:54:23 (5s)
- 23:54:28 (5s)
- 23:54:39  (11s)
- 23:54:44 (5s)
- 23:54:49 (5s)

kevincox:
Ok, in that trace `Sessions/Playing/Progress` was hit in the gap which appears to be wired up to the ping endpoint. 5s with 10s timeout still seems too tight for me as one missed request can result in a timeout (What if the request hangs?), but it seems that in good conditions the Unstable Chromecast client shouldn't have issues with this.

cvium:
Does Chromecast even use Progressive streams? I don't think so, but not sure.

gnattu:
> Does Chromecast even use Progressive streams? I don't think so, but not sure.

From the Chromecast Client source code it does, but that pings the server at 5 second interval which should be good enough.

kevincox:
I don't use Chromecast anymore so am not going to pursue this. Someone else is welcome to take the code and continue.

Issue #11656, opened 2024-05-14
Jellyfin v10.9.1 isn't respecting the metadata lock for Season names in Show libraries causing custom season names to be lost.


maru801:
**Describe The Bug**
For a Show library that contains series with seasons, you're allowed to manually overwrite the title names for each season. You can then lock the metadata for the name so that its title will not change upon a future scan. This is how it worked before v10.9.0.

However, since upgrading to v10.9.0 (and v10.9.1), Jellyfin doesn't respect the metadata lock for the season name/title. With v10.9, a rescan of any season will prompt the server to change the name of any season to be "Season x" where x is the season number in the metadata.

**Steps To Reproduce**
1. Go to a show that has at least one season.
2. Edit the season's metadata and change its title to be anything. After changing the name, make sure to scroll to the bottom of the metadata edit popup to lock the name. Either uncheck the "Name" box, or select the "Lock this item to prevent future changes" box.
3. Reload the client, then navigate to the same show to see your change to the season's name.
4. Refresh the metadata for the whole show (to capture the season(s) as well), and select the "Scan for new and updated files" option. You can also go to the dashboard and use the "Scan All Libraries" button to get this same result.
5. Wait a short while to let the scan do its work.
6. Reload the client again, then navigate to the same show. You'll notice the client changed the custom name you gave the season back to "Season x" despite you locking it.

**Expected Behavior**
Jellyfin should respect the metadata lock and not change any season's name if you set a custom name for the season.

**System (please complete the following information):**
 - Platform: Linux
 - Browser: Any
 - Jellyfin Version: v10.9.1

**Additional Context**
The names seem to be pulled from the .nfo files I have saved next to the media files. If the .nfo has a season number stored with it, doing the above will show the error as I describe. I found that if I remove the season number from the season's .nfo files, this bug doesn't happen.

gnattu:
Should be fixed by #11647

maru801:
This issue was fixed with the server version v10.9.2.

maru801:
Reopening this as I found out that v10.9.2 only fixed the season names from being changed when using the dashboard's "Scan All Libraries" or the "Scan for new and updated files" option for all else. Still, If you scan a season with either the "Search for missing metadata" or "Replace all metadata" options, it will still have its title name changed despite being locked.

And what's more, some seasons are now getting their season name changed to that of another season of another series when scanned. I don't understand what is going on here.

maru801:
[...]

Issue #11655, opened 2024-05-15
[Issue]: Season titles keep reverting to "Season 1", "Season Unknown" etc after being edited manually.


BrianCrowe3:
### Please describe your bug

After updating from 10.8.13 to 10.9.0 and then again to 10.9.1, I noticed all of the manual edits I've made to season titles reverted either to "Season 1", "Season 2" etc in the case that the season was numbered, or "Season Unknown" is the case that it wasn't numbered.

After making the edits again, I noticed that they reverted back to "Season 1" "Season Unknown" etc again, regardless of whether the "Lock this item to prevent future changes" box is ticked or not. 

I have "Metadata savers\Nfo" enabled in library settings, whatever is happening is also editing the "title" part of the nfo files themselves to be "Season 1" "Season Unknown" etc.

### Reproduction Steps

1. Install Jellyfin server on fully updated Windows 10
2. Edit the title of a season
3. At some point within ~4 hours or so it will revert the edit back to "Season 1" etc or "Season Unknown"

### Jellyfin Version

10.9.0

### if other:

10.9.1

### Environment

```markdown
- OS: Windows 10 Pro Build 19045
- Linux Kernel: none
- Virtualization: none
- Clients: Browser
- Browser: Firefox 115.11.0esr
- FFmpeg Version: 6.0.1-Jellyfin
- Playback Method: Direct Play
- Hardware Acceleration: NVENC
- GPU Model: GTX 960
- Plugins: AniDB, AniList, AniSearch, Fanart, TMDB, Studio Images, OMDb, MusicBrainz, AudioDB
- Reverse Proxy: Caddy
- Base URL: none
- Networking: Host
- Storage: local
```

### Jellyfin logs

```shell
Changes to season titles don't appear to be logged.
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Shadowghost:
Likely fixed by #11647

PR #11647, opened 2024-05-14
Fix season names

Shadowghost:
Season names got lost when not scanning the base NFO, so they got reset on any subsequent scan.

**Changes**
Fetch them from the internal dict if it was filled while scanning, otherwise fetch from child seasons.

PR #11674, opened 2024-05-16
fix: keep current season name if no override exists

revam:
**Changes**
TL;DR: Keep the season name if the no override exists for the season.

Longer version:
So my plugin uses custom season names — which have worked fine throughout all of 10.8's lifecycle — but after the new NFO named season parsing was added in 10.9 (21 months ago no less), then the support for custom season names provided by remote metadata plugins partially broke as a result. This simple change fixes the issue by keeping the current season name for the existing seasons but allows the NFO season names to override the name if needed/desired, thus **not** breaking the support for custom season names provided by remote metadata plugins that we've loved Jellyfin for so far.

Also, I did the digging and debugging to fix this _before_ checking for existing issues, and also doesn't think the other PR addressed the root issue here, so I decided to submit this PR  since I believe it fixes it.

**Issues**
https://github.com/jellyfin/jellyfin/issues/11655 ← Fixes this.
https://github.com/jellyfin/jellyfin/issues/11656 ← Fixes this.
https://github.com/jellyfin/jellyfin/pull/11647 ← I felt this didn't address the issue I was having, as I've been debugging Jellyfin to find out _why_ it changed it back, and saw that PR didn't address that issue.

### DIFF
--- a/MediaBrowser.Providers/TV/SeriesMetadataService.cs
+++ b/MediaBrowser.Providers/TV/SeriesMetadataService.cs
@@ -237,7 +237,7 @@ namespace MediaBrowser.Providers.TV
 
                 if (!seasonNumber.HasValue || !seasonNames.TryGetValue(seasonNumber.Value, out var seasonName))
                 {
-                    seasonName = GetValidSeasonNameForSeries(series, null, seasonNumber);
+                    seasonName = GetValidSeasonNameForSeries(series, existingSeason?.Name, seasonNumber);
                 }
 
                 if (existingSeason is null)


gnattu:
I don't think your change will fix the problem here either, as you are *not* addressing the root cause. #11647 is doing the correct thing.

Your code will pass in a null or empty string and still get the season name reset because you are not addressing the root cause: the current season names are not present in the `series.SeasonNames` at the time of scanning. 

In fact, the line you changed won't be executed if that season has a custom season name in the first place. If that line is ever executed, it means the season name is lost and needs to be fixed by #11647.

revam:
image
image

revam:
The fact that they're not in the dict is the issue. it's resetting it because they're not in the dict.

revam:
> In fact, the line you changed won't be executed if that season has a custom season name in the first place. If that line is ever executed, it means the season name is lost and needs to be fixed by https://github.com/jellyfin/jellyfin/pull/11647.

I don't know if this is intended behavior or not, but it will be reset to "Season X" if a series NFO doesn't contain a name for the season but a name was previously set by a remote metadata provider. If that is the intended behavior, then I will leave my case here and close this PR.

Shadowghost:
@revam since I can't debug it with the shoko plugin, would you mind checking my updated PR?

revam:
I'll test the other pr later today when i have some free time. @Shadowghost

revam:
@Shadowghost That seems to have done the trick for my use case at least. So should I close this PR then?

image

revam:
I'll just close it.

PR #11680, opened 2024-05-16
Secure local playlist path handling

Shadowghost:
Currently we resolved any valid paths denoted in the playlist, possibly also to files outside of libraries

**Changes**
* Restrict referenced items in playlists to be within any library

### DIFF
--- a/MediaBrowser.Providers/Playlists/PlaylistItemsProvider.cs
+++ b/MediaBrowser.Providers/Playlists/PlaylistItemsProvider.cs
@@ -78,7 +78,7 @@ namespace MediaBrowser.Providers.Playlists
 
                 if (string.Equals(".m3u8", extension, StringComparison.OrdinalIgnoreCase))
                 {
-                    return GetM3u8Items(stream, path);
+                    return GetM3uItems(stream, path);
                 }
 
                 if (string.Equals(".pls", extension, StringComparison.OrdinalIgnoreCase))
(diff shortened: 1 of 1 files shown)


gnattu:
Are playlist designed to only include files within the same folder though? For example for a playlist referencing multiple albums, which is very likely to go across boundary.

cvium:
IMO, paths should be relative to a library root but not be constrained to a single library

MediaBrowser.Providers/Playlists/PlaylistItemsProvider.cs
```diff
@@ -25,10 +26,13 @@ public class PlaylistItemsProvider : ICustomMetadataProvider<Playlist>,
...
     {
         private readonly ILogger<PlaylistItemsProvider> _logger;
+        private readonly ILibraryManager _libraryManager;
+        private readonly CollectionType[] _ignoredCollections = [CollectionType.livetv, CollectionType.boxsets, CollectionType.playlists];
 
-        public PlaylistItemsProvider(ILogger<PlaylistItemsProvider> logger)
+        public PlaylistItemsProvider(ILogger<PlaylistItemsProvider> logger, ILibraryManager libraryManager)
         {
             _logger = logger;
+            _libraryManager = libraryManager;
```
gnattu, line 35: I *think* we can use the static `BaseItem.LibraryManager`

MediaBrowser.Providers/Playlists/PlaylistItemsProvider.cs
```diff
@@ -59,109 +63,127 @@ public Task<ItemUpdateType> FetchAsync(Playlist item, MetadataRefreshOptions opt
...
+            else
+            {
+                var baseFolder = Path.GetDirectoryName(playlistPath);
+                var basePath = Path.Combine(baseFolder, itemPath);
+                pathToCheck = Path.GetFullPath(basePath);
+                if (!File.Exists(pathToCheck))
+                {
+                    return false;
+                }
+            }
```
gnattu, line 175: The whole logic may be simplified by `BaseItem.FileSystem.MakeAbsolutePath()`. That method has some extensive check of relative path, and returns the absolute path. You pass in a containing folder and a path to be checked, if the path to be checked is absolute, it will return it directly, otherwise it will combine both paths to make it absolute.

crobibero (approved):

gnattu (approved):

PR #11743, opened 2024-05-19
Fix replace logic

### LINKED ISSUE #8532: [Issue]: Metadata edits are being reverted
### Please describe your bug

I have some kids series that are screen recorded from TV long time ago. It does get correctly identified but not fully. Problem is mostly names of episodes are episode numbers. I have edited this to be correct, but the metadata is reverted back to the detected wrong values after some time (few hours?). I have set my library to not refresh metadata (which is the default anyway) and I even tried to remove the metadata fetchers, but even this change gets reverted. It is driving me nuts to be honest.
I am new to Jellyfin, so not sure, maybe I am doing some stupid mistake, but this is not rocket science and I should be able to figure that, right? 

### Jellyfin Version

10.8.0

### if other:

_No response_

### Environment

```markdown
- OS: Linux
- Virtualization: Docker
- Clients: Chromecast, Kodi, Android
- Browser: Firefox
- FFmpeg Version: Whatever you ship in the Docker image
- Playback Method: I guess all of the above options, but there is no problem with playback for me
- Hardware Acceleration: No
- Plugins: Kodi Queue
- Reverse Proxy: Caddy 2
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

```shell
There was no playback of any sort between my metadata edit and the reversion. The only thing I see in logs is old playback from 2 days ago. 
I am happy to provide additional info but at this point I don't see there is anything more to add.
```

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

### LINKED ISSUE #11294: [Issue]: Lock this item to prevent future changes - does not work
### Please describe your bug

Lock this Item does not lock it completely. Critic and Community reviews that have been deleted by User are repopulated by a System Scan or moving the file to a new directory. I have read that is by design and that is how it works. This is NOT the way it should work. If a user enters a zero entry in a data field, Jellyfin should recognize that null entry and not repopulate it with new data.

### Reproduction Steps

Locking item and seeing it change.

### Jellyfin Version

10.8.13

### if other:

_No response_

### Environment

```markdown
- OS:10.8.8
```

### Jellyfin logs

```shell
N/A
```

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

### LINKED ISSUE #11656: Jellyfin v10.9.1 isn't respecting the metadata lock for Season names in Show libraries causing custom season names to be lost.

### LINKED ISSUE #11717: [Issue]: "Replace all metadata" does not work
### Please describe your bug

When I want to refresh metadata, the "Replace all metadata" does not work. I have to remove the old data manually, then use "Search for missing metadata" to get the new metadata.

This bug starts from 10.9.0

### Reproduction Steps

1. upgrade Jellyfin to 10.9.0+
2. choose an item with old metadata
3. refresh metadata using "Replace all metadata"
4. nothing will be updated

### Jellyfin Version

10.9.0

### if other:

10.9.2

### Environment

```markdown
- OS: Alpine
- Linux Kernel: 6.8.4
- Virtualization: Docker
- Clients: Jellyfin-Web
- Browser: Firefox 126.0
- FFmpeg Version: 6.0.1-Jellyfin
- Playback Method: Direct
- Hardware Acceleration: Intel QSV
- GPU Model: Intel UHD 630
- Plugins: TMDB / jellyfin-bangumi
- Reverse Proxy: Caddy / none
- Base URL: /jellyfin
- Networking: Host
- Storage: NFS
```

### Jellyfin logs

```shell
[2024-05-18 16:51:37.092 +00:00] [INF] [86] MediaBrowser.MediaEncoding.Encoder.MediaEncoder: Starting "/usr/lib/jellyfin-ffmpeg/ffprobe" with args "-analyzeduration 200M -probesize 1G -i file:\"/media/I Was Reincarnated as the 7th Prince so I Can Take My Time Perfecting My Magical Ability (2024)/Season 1/I Was Reincarnated as the 7th Prince so I Can Take My Time Perfecting My Magical Ability - S01E06.mp4\" -threads 0 -v warning -print_format json -show_streams -show_chapters -show_format"
[2024-05-18 16:51:37.690 +00:00] [INF] [86] Jellyfin.Plugin.Bangumi.Providers.EpisodeProvider: metadata for "I Was Reincarnated as the 7th Prince so I Can Take My Time Perfecting My Magical Ability - S01E06.mp4": "<Bangumi Episode #1310124: 姉と魔獣と魔剣と兄と>"
```

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

### LINKED ISSUE #11773: [Issue]: Jellyfin changes back metadata on refresh despite manual identification
### Please describe your bug

Jellyfin keeps changing back metadata for media even if I use the "Identify..." functionality to tell Jellyfin what the content actually is. Mainly happens with foreign content such as Korean movies. Examples include:

- Burning (2018)
- The Phone (2015)
- Mother (2009)
- Deranged (2012)
- The Call (2020)

The movie files follow the naming syntax outlined by the documentation, but I'm guessing Jellyfin is still misidentifying them because there are multiple movies with the same name and release year.

Ideally, Jellyfin should remember that the user/admin has properly identified a piece of content and remember the unique ID from the manual identification, so that the next metadata refresh gets updated metadata from the correct content. Locking metadata is a workaround that will prevent future updates to the metadata should they become available.

### Reproduction Steps

1. See Jellyfin has misidentified media
2. Correctly tag media with "Identify"
3. Wait until next metadata refresh
4. Jellyfin sometimes overwrites correct metadata with that from a different movie/show

### Jellyfin Version

10.9.0

### if other:

10.9.2

### Environment

```markdown
- OS: Unraid
- Linux Kernel: 6.1.79-Unraid
- Virtualization: Docker
- Clients: Browser
- Browser: 124.0.6367.208 (Official Build) (64-bit) (cohort: Control) 
- FFmpeg Version: ffmpeg version 6.0.1-Jellyfin 
- Playback Method: N/A
- Hardware Acceleration: VAAPI
- GPU Model: AMD Ryzen 5 PRO 4650G with Radeon Graphics
- Plugins: AniDB, AudioDB, MusicBrainz, OMDb, Open Subtitles, Playback Reporting, Reports, Studio Images, Subtitle Extract, TMDb, TMDb Trailers, Trakt, Webhook
- Reverse Proxy: nginx
- Base URL: none
- Networking: Bridge/NAT
- Storage: local
```

### Jellyfin logs

```shell
N/A
```

### FFmpeg logs

```shell
N/A
```

N/A

_No response_

### Code of Conduct

### LINKED ISSUE #11831: [Issue]: Editing parental rating in series overwrites locked parental rating in seasons and episodes
### Please describe your bug

Editing parental rating in series overwrites parental rating in seasons and episodes, even if parental rating is locked in season or episodes.
It is okay if the parental rating value is writen to seasons and episodes, but it schould not overide locked values.

### Reproduction Steps

1. edit metadata of series
2. set new parental rating
3. save changes
4. edit metadata of season or episode of this series
5. parental rating is set to what was set in series, even if it was locked

### Jellyfin Version

10.9.0

### if other:

10.9.2

### Environment

```markdown
- OS: Ubuntu 22.04.4 LTS
- Linux Kernel:
- Virtualization: Docker
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins: Fanart, Open Subtitles, Playback Reporting, Reports, Subtitle Extract, TheTVDB, Webhook
- Reverse Proxy: none + nginx
- Base URL: none
- Networking: host
- Storage: local, NFS
```

### Jellyfin logs

```shell
I have no log output if I edit parental rating
```

### FFmpeg logs

```shell
I have no log output if I edit parental rating
```

_No response_

parental_rating_locked

### Code of Conduct

Shadowghost:
We never passed the flag for metadata replacement to the remote provider logic, hence the remote result would never replace what's already there.
Additionally, the logic did not repect if an item was locked or not

**Changes**
* Pass replace flag to remote providers
* Fix merge logic to only override empty fields, except when merging with the actual item
  * this will gradually build up metadata beginning with local providers and going through remote providers later on

**Issues**
Fixes #8532 (potentially)
Fixes #11294 (the part where it is repopulated)
Fixes #11656
Fixes #11717
Fixes #11773 (potentially)
Fixes #11831

**Testing**
* Add series, manually remove some metadata and replace all metadata -> should refill
* Add series, manually remove some metadata and fetch missing metadata -> should refill
* Add series, manually identify, remove some metadata and replace metadata -> should keep identified series
* Any other combination of the above and with replace images

### DIFF
--- a/MediaBrowser.Providers/Manager/MetadataService.cs
+++ b/MediaBrowser.Providers/Manager/MetadataService.cs
@@ -654,139 +654,123 @@ namespace MediaBrowser.Providers.Manager
                 await RunCustomProvider(provider, item, logName, options, refreshResult, cancellationToken).ConfigureAwait(false);
             }
 
-            var temp = new MetadataResult<TItemType>
+            if (!item.IsLocked)
             {
-                Item = CreateNew()
-            };
-            temp.Item.Path = item.Path;
-
-            // If replacing all metadata, run internet providers first
-            if (options.ReplaceAllMetadata)
-            {
-                var remoteResult = await ExecuteRemoteProviders(temp, logName, true, id, providers.OfType<IRemoteMetadataProvider<TItemType, TIdType>>(), cancellationToken)
-                    .ConfigureAwait(false);
-
-                refreshResult.UpdateType |= remoteResult.UpdateType;
-                refreshResult.ErrorMessage = remoteResult.ErrorMessage;
-                refreshResult.Failures += remoteResult.Failures;
-            }
-
-            var hasLocalMetadata = false;
-            var foundImageTypes = new List<ImageType>();
-
-            foreach (var provider in providers.OfType<ILocalMetadataProvider<TItemType>>())
-            {
-                var providerName = provider.GetType().Name;
-                Logger.LogDebug("Running {Provider} for {Item}", providerName, logName);
-
-                var itemInfo = new ItemInfo(item);
-
-                try
+                var temp = new MetadataResult<TItemType>
                 {
-                    var localItem = await provider.GetMetadata(itemInfo, options.DirectoryService, cancellationToken).ConfigureAwait(false);
+                    Item = CreateNew()
+                };
+                temp.Item.Path = item.Path;
+                temp.Item.Id = item.Id;
 
-                    if (localItem.HasMetadata)
+                var hasLocalBaseMetadataOrLocked = false;
+                var foundImageTypes = new List<ImageType>();
+
+                foreach (var provider in providers.OfType<ILocalMetadataProvider<TItemType>>())
+                {
+                    var providerName = provider.GetType().Name;
+                    Logger.LogDebug("Running {Provider} for {Item}", providerName, logName);
+
+                    var itemInfo = new ItemInfo(item);
+
+                    try
                     {
-                        foreach (var remoteImage in localItem.RemoteImages)
+                        var localItem = await provider.GetMetadata(itemInfo, options.DirectoryService, cancellationToken).ConfigureAwait(false);
+
+                        if (localItem.HasMetadata)
                         {
-                            try
+                            foreach (var remoteImage in localItem.RemoteImages)
                             {
-                                if (item.ImageInfos.Any(x => x.Type == remoteImage.Type)
-                                    && !options.IsReplacingImage(remoteImage.Type))
+                                try
                                 {
-                                    continue;
+                                    if (item.ImageInfos.Any(x => x.Type == remoteImage.Type)
+                                        && !options.IsReplacingImage(remoteImage.Type))
+                                    {
+                                        continue;
+                                    }
+
+                                    await ProviderManager.SaveImage(item, remoteImage.Url, remoteImage.Type, null, cancellationToken).ConfigureAwait(false);
+                                    refreshResult.UpdateType |= ItemUpdateType.ImageUpdate;
+
+                                    // remember imagetype that has just been downloaded
+                                    foundImageTypes.Add(remoteImage.Type);
                                 }
-
-                                await ProviderManager.SaveImage(item, remoteImage.Url, remoteImage.Type, null, cancellationToken).ConfigureAwait(false);
-                                refreshResult.UpdateType |= ItemUpdateType.ImageUpdate;
-
-                                // remember imagetype that has just been downloaded
-                                foundImageTypes.Add(remoteImage.Type);
+                                catch (HttpRequestException ex)
+                                {
+                                    Logger.LogError(ex, "Could not save {ImageType} image: {Url}", Enum.GetName(remoteImage.Type), remoteImage.Url);
+                                }
                             }
-                            catch (HttpRequestException ex)
+
+                            if (foundImageTypes.Count > 0)
                             {
-                                Logger.LogError(ex, "Could not save {ImageType} image: {Url}", Enum.GetName(remoteImage.Type), remoteImage.Url);
+                                imageService.UpdateReplaceImages(options, foundImageTypes);
                             }
+
+                            if (imageService.MergeImages(item, localItem.Images, options))
+                            {
+                                refreshResult.UpdateType |= ItemUpdateType.ImageUpdate;
+                            }
+
+                            MergeData(localItem, temp, Array.Empty<MetadataField>(), false, true);
+                            refreshResult.UpdateType |= ItemUpdateType.MetadataImport;
+
+                            // Only one local provider allowed per item
+                            if (item.IsLocked || localItem.Item.IsLocked || HasBaseMetadata(localItem.Item))
+                            {
+                                hasLocalBaseMetadataOrLocked = true;
+                            }
+
+                            break;
                         }
 
-                        if (foundImageTypes.Count > 0)
-                        {
-                            imageService.UpdateReplaceImages(options, foundImageTypes);
-                        }
-
-                        if (imageService.MergeImages(item, localItem.Images, options))
-                        {
-                            refreshResult.UpdateType |= ItemUpdateType.ImageUpdate;
-                        }
-
-                        MergeData(localItem, temp, Array.Empty<MetadataField>(), options.ReplaceAllMetadata, true);
-                        refreshResult.UpdateType |= ItemUpdateType.MetadataImport;
-
-                        // Only one local provider allowed per item
-                        if (item.IsLocked || localItem.Item.IsLocked || IsFullLocalMetadata(localItem.Item))
-                        {
-                            hasLocalMetadata = true;
-                        }
-
-                        break;
+                        Logger.LogDebug("{Provider} returned no metadata for {Item}", providerName, logName);
                     }
-
-                    Logger.LogDebug("{Provider} returned no metadata for {Item}", providerName, logName);
-                }
-                catch (OperationCanceledException)
-                {
-                    throw;
-                }
-                catch (Exception ex)
-                {
-                    Logger.LogError(ex, "Error in {Provider}", provider.Name);
-
-                    // If a local provider fails, consider that a failure
-                    refreshResult.ErrorMessage = ex.Message;
-                }
-            }
-
-            // Local metadata is king - if any is found don't run remote providers
-            if (!options.ReplaceAllMetadata && (!hasLocalMetadata || options.MetadataRefreshMode == MetadataRefreshMode.FullRefresh || !item.StopRefreshIfLocalMetadataFound))
-            {
-                var remoteResult = await ExecuteRemoteProviders(temp, logName, false, id, providers.OfType<IRemoteMetadataProvider<TItemType, TIdType>>(), cancellationToken)
-                    .ConfigureAwait(false);
-
-                refreshResult.UpdateType |= remoteResult.UpdateType;
-                refreshResult.ErrorMessage = remoteResult.ErrorMessage;
-                refreshResult.Failures += remoteResult.Failures;
-            }
-
-            if (providers.Any(i => i is not ICustomMetadataProvider))
-            {
-                if (refreshResult.UpdateType > ItemUpdateType.None)
-                {
-                    if (hasLocalMetadata)
+                    catch (OperationCanceledException)
                     {
-                        MergeData(temp, metadata, item.LockedFields, true, true);
+                        throw;
                     }
-                    else
+                    catch (Exception ex)
                     {
-                        if (!options.RemoveOldMetadata)
-                        {
-                            MergeData(metadata, temp, Array.Empty<MetadataField>(), false, false);
-                        }
+                        Logger.LogError(ex, "Error in {Provider}", provider.Name);
 
-                        // Will always replace all metadata when Scan for new and updated files is used. Else, follow the options.
-                        MergeData(temp, metadata, item.LockedFields, options.MetadataRefreshMode == MetadataRefreshMode.Default || options.ReplaceAllMetadata, false);
+                        // If a local provider fails, consider that a failure
+                        refreshResult.ErrorMessage = ex.Message;
                     }
                 }
-            }
 
-            foreach (var provider in customProviders.Where(i => i is not IPreRefreshProvider))
-            {
-                await RunCustomProvider(provider, item, logName, options, refreshResult, cancellationToken).ConfigureAwait(false);
+                if (options.ReplaceAllMetadata || !(hasLocalBaseMetadataOrLocked && item.StopRefreshIfLocalMetadataFound) || options.MetadataRefreshMode == MetadataRefreshMode.FullRefresh)
+                {
+                    var remoteResult = await ExecuteRemoteProviders(temp, logName, false, id, providers.OfType<IRemoteMetadataProvider<TItemType, TIdType>>(), cancellationToken)
+                        .ConfigureAwait(false);
+
+                    refreshResult.UpdateType |= remoteResult.UpdateType;
+                    refreshResult.ErrorMessage = remoteResult.ErrorMessage;
+                    refreshResult.Failures += remoteResult.Failures;
+                }
+
+                if (providers.Any(i => i is not ICustomMetadataProvider))
+                {
+                    if (refreshResult.UpdateType > ItemUpdateType.None)
+                    {
+                        if (options.RemoveOldMetadata)
+                        {
+                            MergeData(metadata, temp, Array.Empty<MetadataField>(), true, true);
+                        }
+
+                        MergeData(temp, metadata, item.LockedFields, options.MetadataRefreshMode >= MetadataRefreshMode.Default || options.ReplaceAllMetadata, true);
+                    }
+                }
+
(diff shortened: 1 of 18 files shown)


Bond-009 (approved):

oddstr13:
I haven't studied the code, but I have the following questions;
How does this interact with multiple enabled metadata providers and the ordering of those?
What about fields that contain multiple values, such as Genres, People, Tags and such?

sjorge:
> What about fields that contain multiple values, such as Genres, People, Tags and such?

First provider wins I think? (Otherwise you get duplicates if there are small variations in spelling, which I've seen happen with actor names)

Shadowghost:
> How does this interact with multiple enabled metadata providers and the ordering of those?

NFO/Local provider metadata is always preferred, after that providers only add metadata not already there

> What about fields that contain multiple values, such as Genres, People, Tags and such?

The new logic merges them from the different providers while removing duplicates

> Would be nice if you could disable those in the provider configuration

Purely provider dependant

TimGels:
# Test Results:
I pulled your branch and started testing. I executed 4 different test cases. The first three were based on your description tests. Number four was made up by me. 1 - 3 seem to have passed. the 4th one I had a surprising finding, but that finding could be unrelated to your changes.

To start off, the following screenshot displays the information of the show right after the library scan. This will be the starting point for the test cases 1 and 2.
afbeelding

## Executing test 1 (passed):
I started by manually emptying the fields: "Original title", "Community rating" and "Overview". This is shown in the following screenshot:
afbeelding

I saved the changes and proceeded by executing a "Refresh Metadata"-> "Replace all metadata" on this show.

I refreshed the page and the manually deleted information is now present again:
afbeelding

## Executing test 2 (passed):
I manually emptied the fields: "Original title", "Community rating" and "Overview".
afbeelding

I then proceeded by executing a "Refresh Metadata"-> "Search for missing metadata" on this show.

I refreshed the page and the manually deleted information is now present again:
afbeelding

## Executing test 3 (passed):
I manually identified the series as: "A Certain Magical Index". The metadata fields are updated:
afbeelding

I started by manually emptying the fields: "Original title", "Community rating" and "Overview".
afbeelding

I then proceeded by executing a "Refresh Metadata"-> "Replace all metadata" on the **library** level.

I refreshed the page and the manually deleted information is now present again:
afbeelding

## Executing test 4 (failed, possibly unrelated to this PR): 
I manually emptied the fields: "Original title", "Community rating" and "Overview".
afbeelding

You can see the fields getting emptied also show nothing on the shows page anymore:
afbeelding

I proceed by enabling "Lock this item to prevent future changes".
afbeelding

I initiate a "Replace all metadata" on this show level. 

I refreshed the page and the manually deleted information is still empty (as it should due to it being locked):
afbeelding

However, the actors do not show up anymore. This could be unrelated to this PR, but I wanted to share this finding regardless.
afbeelding

MediaBrowser.Providers/Manager/MetadataService.cs
```diff
@@ -164,7 +164,7 @@ public virtual async Task<ItemUpdateType> RefreshMetadata(BaseItem item, Metadat
             }
 
             // Next run remote image providers, but only if local image providers didn't throw an exception
-            if (!localImagesFailed && refreshOptions.ImageRefreshMode != MetadataRefreshMode.ValidationOnly)
+            if (!localImagesFailed && refreshOptions.ImageRefreshMode > MetadataRefreshMode.ValidationOnly)
```
crobibero, line 167: I'm not a fan of using `>` when comparing enums since it can have unintended consequences when adding a new value

crobibero (approved):

MediaBrowser.Providers/Manager/MetadataService.cs
Shadowghost: I do agree but we currently use that kind of comparison for this enum quite often, so it's fine I think.

Shadowghost:
@TimGels Test case 4 should now work too.

TimGels:
@Shadowghost 
Test case 4 indeed seems to pass now. I locked it, re-scanned everything, refreshed the page and it still showed people. Basically I re executed step 4. Another thing we discussed testing was playing around with removing some actors and combining that with locking and not locking in different test scenario's. I also tested that and there were no duplicated and the actors worked as expected.

xd003:
Can confirm that everything is working as mentioned in the PR

Bond-009 (approved):

MediaBrowser.Providers/Manager/MetadataService.cs
```diff
@@ -654,148 +655,124 @@ public bool CanRefreshPrimary(Type type)
                 await RunCustomProvider(provider, item, logName, options, refreshResult, cancellationToken).ConfigureAwait(false);
             }
 
-            var temp = new MetadataResult<TItemType>
+            if (!item.IsLocked)
```
cvium, line 658: This should be inverted afaict. It gets very deep already.

MediaBrowser.Providers/Manager/MetadataService.cs
```diff
@@ -1048,6 +1021,10 @@ private bool HasChanged(BaseItem item, IHasItemChangeMonitor changeMonitor, IDir
                 {
                     target.Studios = source.Studios;
                 }
+                else if (source.Studios.Length >= 0)
```
cvium, line 1024: If it's equal to 0, it doesn't change anything?

Shadowghost: Since we're oly in this check if we either not replace or the target already has studios this is intended. If souce has studios, we append, if it does not we ignore.

cvium: I was just wondering why the check was `>= 0` and not `> 0`

Shadowghost: I think removing the checks to delete any duplicates is sensible regardless

nodje:
Was this merged into 10.9.4?

I still have the issue

DrMcCoy:
For me, replacing works now with 10.9.4...mostly. I does grab description and ratings again now. But it doesn't get season names, despite that tick being on in the settings. So that's still a bug.

However, it also now doesn't break season names anymore when they've been manually changed in the nfo file. Which is good, because I kinda only want fix season names in my library and keep the other info as-is. I wrote myself a small Python script that goes over my library, gets the season names from TMDB and patches the nfo files. 

In case that's useful for anyone else, it's here: https://github.com/DrMcCoy/fix-season-names . No warranties, use at your own peril, etc., though, of course.

nodje:
OK, to be very specific, here's the test I perform:
- go to a movie, Edit Metadata
- In Metadata Settings->Preferred download language change language to a specific one to override the Library's default
- then Refresh Metadata->mode: Replace All Metadata

and it doesn't change the metadata like it used to in 10.8.x

Logs from the operation
```
[16:01:44] [INF] [31] MediaBrowser.MediaEncoding.Encoder.MediaEncoder: Starting /usr/lib/jellyfin-ffmpeg/ffprobe with args -analyzeduration 200M -probesize 1G -i file:"/media/Directors/France/Dupieux, Quentin - 1974-/2022 - Dupieux, Quentin - Fumer fait tousser [imdbid-tt15471560]/2022 - Dupieux, Quentin - Fumer fait tousser - BD1080pHEVC.mp4" -threads 0 -v warning -print_format json -show_streams -show_chapters -show_format
[mov,mp4,m4a,3gp,3g2,mj2 @ 0x55abeff24ec0] stream 0, timescale not set
[16:01:45] [WRN] [31] MediaBrowser.XbmcMetadata.Providers.MovieNfoProvider: Trailer URL uses a deprecated format : plugin://plugin.video.youtube/?action=play_video&videoid=u80Bh3sMrzU. Using plugin://plugin.video.youtube/play/?video_id=u80Bh3sMrzU instead is advised.
[16:01:45] [WRN] [31] MediaBrowser.XbmcMetadata.Providers.MovieNfoProvider: Trailer URL uses a deprecated format : plugin://plugin.video.youtube/?action=play_video&videoid=Q0X8LXlE9pA. Using plugin://plugin.video.youtube/play/?video_id=Q0X8LXlE9pA instead is advised.
[16:01:45] [WRN] [31] MediaBrowser.XbmcMetadata.Providers.MovieNfoProvider: Trailer URL uses a deprecated format : plugin://plugin.video.youtube/?action=play_video&videoid=fxc6TyTdW50. Using plugin://plugin.video.youtube/play/?video_id=fxc6TyTdW50 instead is advised.
```
Incidently, I'm not sure why the msg " Trailer URL uses a deprecated format" shows up as I never enabled/used Trailers.

Shadowghost:
@nodje if you remove the NFO, does it work then? I think the issue is that NFOs take precedence over anything else and since they are language-independent, they'll fill up the metadata before the remote providers do.
Another option would be to identify the item with the correct language.

NiklasBr:
But should the NFO really take precedence when *Replace All Metadata* is checked? Previous behaviour was that the NFO content was replaced in that case.

Shadowghost:
I think this kind of is an apple and egg problem since technically NFOs are metadata sources to query for metadata and nothing more. They are the ultimate source of truth in my opinion. I agree and know (since I changed it) that before remote provider results overrode the NFOs if you replaced all metadata, but I'm unsure what to do about it.
Maybe we need a switch for ignoring local metadata all together - but that would be a 10.10 thing. You should be able to fix it by removing the NFO and force refreshing afterwards.

Regardless, I think this conversation should be moved to an issue.

DrMcCoy:
You're right, that doesn't work for me either.

I add a new TV show, it defaults to English (because most of what I have is English, so that's okay). But this show is a German one and I want German metadata. So I change it to German and Germany, then replace all metadata. That used to work, but now it doesn't.

I can delete all the nfos now, but that's a bit cumbersome? Also, isn't the language selection saved to the tvshow.nfo in the first place?

DrMcCoy:
> Maybe we need a switch for ignoring local metadata all together - but that would be a 10.10 thing. You should be able to fix it by removing the NFO and force refreshing afterwards

I disagree that this is a 10.10 thing or that deleting the NFO is a "good" solution, even in the meantime.

You basically broke adding items to mixed-language metadata libraries. Since there is no way to signal "I want this new stuff to be in $LANGUAGE instead of the default", "replace all metadata" basically the only way to do this. You shouldn't need fuzz around with the generated file in the filesystem afterwards, that's just...bad.

michaelkrieger:
There 100% needs to be a “fetch metadata from remote sources” vs “refresh data from local NFO”. A checkbox will do. One looks past the NFO. One looks only to the NFO.

NiklasBr:
> They are the ultimate source of truth in my opinion

This was not the case before 10.9, and before 10.9 it was intuitive and worked. This new behaviour should have been a separate feature.

And on top of that, they are never the ultimate source of truth in my opinion since the the remote source is likely to have better, more up-to-date data in > 99% of the cases because at most one person contributes to the NFO file, while potentially thousands contribute to the remote source.

Shadowghost:
#11921 should take care of it - skipping local NFOs when replacing and saving is enabled. Please test if you have the time.

nodje:
It does work to remove the NFO.
Identifying the movie with the Preferred download language set doesn't seem to work unfortunately. It would have been a good work around.

Unrelated to the case but I'd appreciate a pointer to an explanation: what puzzles me with NFO, is that I get two of them for existing movies after the 10.9 upgrade. The new movie.nfo one an the old one that bears the name of the movie file.
E.G.
- movie.nfo
- 2011 - Winding Refn, Nicholas - Drive - 2160pHEVC.HDR10.nfo

Shall I delete all NFO with the old format?

PR #11754, opened 2024-05-20
Fix BD/DVD folder chapter image extraction

Shadowghost:
**Changes**
* Generate concat file and pass it to the chapter extraction command


Bond-009 (approved):
Partial fix, but better than nothing

nyanmisaka (approved):

PR #11762, opened 2024-05-20
Mark Audio as RequiresDeserialization and backfill data

Shadowghost:
The provider checks for lyric changes rely on the content of `LyricFiles` within `Audio`. But `Audio` did not require serialization into the database, dropping any non-`BaseItem` data on save. Therefore on every scan the check would fail and the provider would run.

**Changes**
* Declare `Audio` as `RequiresDeserialization`
* Backfill serializable data for all `Audio` objects

### DIFF
--- a/Jellyfin.Server/Migrations/Routines/FixAudioData.cs
+++ b/Jellyfin.Server/Migrations/Routines/FixAudioData.cs
@@ -1,0 +1,104 @@
+using System;
+using System.Globalization;
+using System.IO;
+using System.Linq;
+using System.Threading;
+using Jellyfin.Data.Enums;
+using MediaBrowser.Controller;
+using MediaBrowser.Controller.Entities;
+using MediaBrowser.Controller.Entities.Audio;
+using MediaBrowser.Controller.Persistence;
+using MediaBrowser.Model.Entities;
+using Microsoft.Extensions.Logging;
+
+namespace Jellyfin.Server.Migrations.Routines
+{
+    /// <summary>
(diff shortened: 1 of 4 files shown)


crobibero (approved):

Jellyfin.Server/Migrations/Routines/FixAudioData.cs
```diff
@@ -0,0 +1,104 @@
...
+            });
+
+            while (startIndex < records)
+            {
+                var results = _itemRepository.GetItemList(new InternalItemsQuery
+                {
+                    IncludeItemTypes = [BaseItemKind.Audio],
+                    StartIndex = startIndex,
+                    Limit = 100,
+                    SkipDeserialization = true
```
1337joe, line 84: What's the impact of not skipping deserialization given that before this change nothing would be serialized for Audio anyway?

All the changes to pass the skip flag just seem like dead code once this migration has run, so how valuable is it to add?

Shadowghost: Allowing deserialization means that any non-BaseItem field not annotated with @JsonIgnore is being saved as one big JSON blob into the database.
In this case it only includes the LyricFiles.
The migration alone is not enough because the database field would never get updated.

Shadowghost: Otherwise a simple rescan after enabling serialization would have been enough to fix the issue.

1337joe (approved):

PR #11766, opened 2024-05-21
Filter invalid IPs on external interface matching

gnattu:
Currently, we are assuming all IPs not in user-set subnet are external IPs, but this is not accurate enough:

- IPv4 addresses are not in IPv6 subnets and vice versa, and we will be returning IPv4 binding address as IPv6 external IP

- Link Local Addresses are not considered as LAN address in most cases, but we are treating them as external address if available and that is wrong because such addresses are not route-able.

**Changes**

**Issues**

### DIFF
--- a/src/Jellyfin.Networking/Manager/NetworkManager.cs
+++ b/src/Jellyfin.Networking/Manager/NetworkManager.cs
@@ -903,6 +903,17 @@ public class NetworkManager : INetworkManager, IDisposable
         return false;
     }
 
+    /// <summary>
+    ///  Get if the IPAddress is Link-local.
+    /// </summary>
+    /// <param name="address">The IP Address.</param>
+    /// <returns>Bool indicates if the address is link-local.</returns>
+    public bool IsLinkLocalAddress(IPAddress address)
+    {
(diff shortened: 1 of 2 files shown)


src/Jellyfin.Networking/Manager/NetworkManager.cs
```diff
@@ -903,6 +903,17 @@ public bool IsInLocalNetwork(string address)
...
 
+    /// <summary>
+    ///  Get if the IPAddress is Link-local.
+    /// </summary>
+    /// <param name="address">The IP Address.</param>
+    /// <returns>Bool indicates if the address is link-local.</returns>
+    public bool IsLinkLocalAddress(IPAddress address)
+    {
+        ArgumentNullException.ThrowIfNull(address);
+        return IPNetwork.Parse("169.254.0.0/16").Contains(address) || address.IsIPv6LinkLocal;
```
Shadowghost, line 914: Please add this to our NetworkConstants and check against it

Shadowghost (approved):

crobibero (approved):

PR #11781, opened 2024-05-21
Retain order blu-ray segments

Bond-009:
This fixes the order, but playback and chapter image extraction still don't work properly for me

### DIFF
--- a/MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
+++ b/MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
@@ -1137,13 +1137,11 @@ namespace MediaBrowser.MediaEncoding.Encoder
             var validPlaybackFiles = _blurayExaminer.GetDiscInfo(path).Files;
 
             // Get all files from the BDMV/STREAMING directory
-            var directoryFiles = _fileSystem.GetFiles(Path.Join(path, "BDMV", "STREAM"));
-
             // Only return playable local .m2ts files
-            return directoryFiles
-                .Where(f => validPlaybackFiles.Contains(f.Name, StringComparer.OrdinalIgnoreCase))
+            return validPlaybackFiles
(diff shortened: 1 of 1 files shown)


MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
```diff
@@ -1137,13 +1137,11 @@ public IReadOnlyList<string> GetPrimaryPlaylistM2tsFiles(string path)
             var validPlaybackFiles = _blurayExaminer.GetDiscInfo(path).Files;
 
             // Get all files from the BDMV/STREAMING directory
-            var directoryFiles = _fileSystem.GetFiles(Path.Join(path, "BDMV", "STREAM"));
-
             // Only return playable local .m2ts files
-            return directoryFiles
-                .Where(f => validPlaybackFiles.Contains(f.Name, StringComparer.OrdinalIgnoreCase))
+            return validPlaybackFiles
+                .Select(f => _fileSystem.GetFileInfo(Path.Join(path, "BDMV", "STREAM", f)))
```
cvium, line 1142: Won't this read every file individually?

Bond-009: One fstat per file, yes
Compared to one opendir and a readdir for each file in that directory

crobibero (approved):

Shadowghost (approved):

MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
cvium: Is that how EnumerateFiles works?

Bond-009: https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/IO/Enumeration/FileSystemEnumerator.Unix.cs,180

cvium (approved):

Issue #11783, opened 2024-05-21
[Issue]: Sidecar subtitles are incorrectly scanned, sometimes doubled


Znuff:
### Please describe your bug

Since the upgrade to 10.9, I have noticed multiple random issues with subtitles (sidecar).

I store 2 EN (`en.srt`) and RO (`.ro.srt`) subtitles next to the video file, which I maintain with Bazarr.

So far I've encountered:

- Disappearing `RO` subs, on episodes I haven't touched in at least 2 months, Jellyfin was only reporting the `en.srt` file in the JSON result
- Duplicating Subs listed in Jellyfin, like having the same file listed twice.

For example, I recently refreshed/changed `Stargate Universe - S01E04 - Darkness Bluray`, and after the initial scan, Jellyfin lists the English SRT twice:

![image](https://github.com/jellyfin/jellyfin/assets/496793/f852d67f-2c22-4df7-81c0-d9ae144171f2)

This is not just a visual issue, the JSON data from the server shows the same:

```json
[...]
            "MediaStreams": [
                {
                    "Codec": "subrip",
                    "Language": "eng",
                    "TimeBase": "1/1000",
                    "VideoRange": "Unknown",
                    "VideoRangeType": "Unknown",
                    "AudioSpatialFormat": "None",
                    "LocalizedUndefined": "Undefined",
                    "LocalizedDefault": "Default",
                    "LocalizedForced": "Forced",
                    "LocalizedExternal": "External",
                    "LocalizedHearingImpaired": "Hearing Impaired",
                    "DisplayTitle": "English - SUBRIP - External",
                    "IsInterlaced": false,
                    "IsAVC": false,
                    "IsDefault": false,
                    "IsForced": false,
                    "IsHearingImpaired": false,
                    "Height": 0,
                    "Width": 0,
                    "Type": "Subtitle",
                    "Index": 0,
                    "IsExternal": true,
                    "IsTextSubtitleStream": true,
                    "SupportsExternalStream": true,
                    "Path": "/vod/shows/Stargate Universe (2009)/Season 1/Stargate Universe - S01E04 - Darkness Bluray-1080p.en.srt",
                    "Level": 0
                },
                {
                    "Codec": "subrip",
                    "Language": "eng",
                    "TimeBase": "1/1000",
                    "VideoRange": "Unknown",
                    "VideoRangeType": "Unknown",
                    "AudioSpatialFormat": "None",
                    "LocalizedUndefined": "Undefined",
[...]

PR #11792, opened 2024-05-22
Improve reliability of HasChanged check

Shadowghost:
Before we write the Paths to the item we check for distinction. This should be done on the comparison too. I threw in ordering to prevent issues of potential FS inconsistent return values.

**Issues**
Potentially #11783

### DIFF
--- a/MediaBrowser.Providers/MediaInfo/ProbeProvider.cs
+++ b/MediaBrowser.Providers/MediaInfo/ProbeProvider.cs
@@ -143,14 +143,14 @@ namespace MediaBrowser.Providers.MediaInfo
                 && !video.IsPlaceHolder)
             {
                 var externalFiles = new HashSet<string>(_subtitleResolver.GetExternalFiles(video, directoryService, false).Select(info => info.Path), StringComparer.OrdinalIgnoreCase);
-                if (!new HashSet<string>(video.SubtitleFiles, StringComparer.OrdinalIgnoreCase).SetEquals(externalFiles))
+                if (!new HashSet<string>(video.SubtitleFiles, StringComparer.Ordinal).SetEquals(externalFiles))
                 {
                     _logger.LogDebug("Refreshing {ItemPath} due to external subtitles change.", item.Path);
                     return true;
                 }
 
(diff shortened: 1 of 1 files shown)


crobibero (approved):

MediaBrowser.Providers/MediaInfo/ProbeProvider.cs
```diff
@@ -141,18 +141,18 @@ public bool HasChanged(BaseItem item, IDirectoryService directoryService)
                 && item.SupportsLocalMetadata
                 && !video.IsPlaceHolder)
             {
-                if (!video.SubtitleFiles.SequenceEqual(
+                if (!video.SubtitleFiles.Order().SequenceEqual(
```
cvium, line 144: Rather than ordering and calling distinct, it might be more efficient to use Set operations

```C#
                var set = new HashSet(video.SubtitleFiles);
                set.SymmetricExceptWith( _subtitleResolver.GetExternalFiles(video, directoryService, false).Select(info => info.Path));
                if (set.Count > 0)
```

MediaBrowser.Providers/MediaInfo/ProbeProvider.cs
```diff
@@ -141,34 +142,30 @@ public bool HasChanged(BaseItem item, IDirectoryService directoryService)
                 && item.SupportsLocalMetadata
                 && !video.IsPlaceHolder)
             {
-                if (!video.SubtitleFiles.Order().SequenceEqual(
-                        _subtitleResolver.GetExternalFiles(video, directoryService, false)
-                            .Select(info => info.Path).Distinct().Order().ToList(),
-                        StringComparer.Ordinal))
+                var externalFiles = new HashSet<string>(_subtitleResolver.GetExternalFiles(video, directoryService, false).Select(info => info.Path), StringComparer.OrdinalIgnoreCase);
```
cvium, line 145: Should we be using ignore case for file paths?

Shadowghost: Actually a good point. Depending on OS this might not be correct. What's your preference?

PR #11798, opened 2024-05-23
Recalculate trickplay image height for anamorphic videos

### LINKED ISSUE #11785: [Issue]: Trickplays ignore display aspect ratio
### Please describe your bug

When viewing a generated trick play for a video with a DAR of anything but 1:1, the displayed trickplay ignores it. It seems ffmpeg does correctly embed the DAR in the output jpg, and viewing it directly with MPV looks right, but the display in the browser is stretched nonetheless.

The solution could be implemented in either the client side (correctly displaying trickplay aspect ratio) or on the server side (pre-scaling trickplays, which would require regenerating for all videos without 1:1 DAR)

### Reproduction Steps

1. Find a video with a non-square DAR
1. Generate a trickplay for it
1. Hover over the progress bar and witness the stretched thumbnail

### Jellyfin Version

10.9.0

### if other:

_No response_

### Environment

```markdown
- OS: Debian 12.5
- Linux Kernel: 6.5
- Virtualization: Docker
- Clients: Browser, Desktop (JMP)
- Browser: Firefox
- FFmpeg Version: 6.0.1-jellyfin
- Playback Method: Direct play
- Hardware Acceleration: none
- GPU Model: none
- Plugins: Ani-Sync, AniDB, AniList, Artwork, AudioDB, Discogs, Fanart, InfuseSync, Intro Skipper, Kitsu, Kodi Sync Queue, Media Analyzer, Merge Versions, MusicBrainz, OMDb, One Pace, Open Subtitles, Playback Reporting, Reports, Session Cleaner, Studio Images, Subtitle Extract, TMDb, TheTVDB, Transcode Killer, VGMdb, Webhook
- Reverse Proxy: Caddy
- Base URL: none
- Networking: Docker bridge
- Storage: local
```

### Jellyfin logs

```shell
[10:45:33] [INF] [56] Jellyfin.Server.Implementations.Trickplay.TrickplayManager: Creating trickplay files at 320 width, for /media/shows/The IT Crowd/Season 2/It.Crowd.s02e03.mkv [ID: 41568ed4-03b2-9f36-fc9c-640e5c2088c4]
[10:45:33] [INF] [56] MediaBrowser.MediaEncoding.Encoder.MediaEncoder: Trickplay generation: /usr/lib/jellyfin-ffmpeg/ffmpeg -loglevel error -threads 0 -noautorotate -i file:"/media/shows/The IT Crowd/Season 2/It.Crowd.s02e03.mkv" -an -sn -vf "fps=0.20000000298023224,setparams=color_primaries=bt709:color_trc=bt709:colorspace=bt709,scale=trunc(min(max(iw\,ih*a)\,320)/2)*2:trunc(ow/a/2)*2,format=yuv420p" -threads 0 -c:v mjpeg -qscale:v 4 -f image2 "/cache/temp/a5aea42026674e97802b46255957997d/%08d.jpg"
[10:45:45] [INF] [101] Jellyfin.Server.Implementations.Trickplay.TrickplayManager: Finished creation of trickplay files for /media/shows/The IT Crowd/Season 2/It.Crowd.s02e03.mkv
```

### FFmpeg logs

_No response_

_No response_

Screenshot_2024_05_21_4_27_24_PM

### Code of Conduct

gnattu:
This might still be not exactly accurate but should not be tell from human eyes on most videos.

The most problematic ones would be the 3D Videos where the aspect ratio stored in the container might be per eye instead of per frame, which will lead to an incorrect height being calculated. But we can really do little about that for current pipeline.

**Changes**

- Re-calculate frame height for hardware mjpeg encoders
- Use SAR to recalculate aspect ratio for software mjpeg encoder

**Issues**

Fixes #11785

### DIFF
--- a/MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
+++ b/MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
@@ -827,7 +827,7 @@ namespace MediaBrowser.MediaEncoding.Encoder
             if (imageStream.Width is not null && imageStream.Height is not null)
             {
                 // For hardware trickplay encoders, we need to re-calculate the size because they used fixed scale dimensions
-                var darParts = imageStream.AspectRatio.Split(":");
+                var darParts = imageStream.AspectRatio.Split(':');
                 var (wa, ha) = (double.Parse(darParts[0], CultureInfo.InvariantCulture), double.Parse(darParts[1], CultureInfo.InvariantCulture));
                 // When dimension / DAR does not equal to 1:1, then the frames are most likely stored stretched.
                 // Note: this might be incorrect for 3D videos as the SAR stored might be per eye instead of per video, but we really can do little about it.
--- a/MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
+++ b/MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
@@ -824,6 +824,22 @@ namespace MediaBrowser.MediaEncoding.Encoder
                 options.EnableTonemapping = false;
             }
 
+            if (imageStream.Width is not null && imageStream.Height is not null)
+            {
+                // For hardware trickplay encoders, we need to re-calculate the size because they used fixed scale dimensions
+                var darParts = imageStream.AspectRatio.Split(":");
+                var (wa, ha) = (int.Parse(darParts[0], CultureInfo.InvariantCulture), int.Parse(darParts[1], CultureInfo.InvariantCulture));
+                // When dimension / DAR does not equal to 1:1, then the frames are most likely stored stretched.
+                // Note: this might be incorrect for 3D videos as the SAR stored might be per eye instead of per video, but we really can do little about it.
+                var shouldResetHeight = imageStream.Width * ha != imageStream.Height * wa;
+                if (shouldResetHeight)
+                {
+                    // SAR = DAR * Height / Width
+                    // RealHeight = Height / SAR = Height / (DAR * Height / Width) = Width / DAR
+                    imageStream.Height = Convert.ToInt32(imageStream.Width.Value * (double)ha / wa);
+                }
+            }
+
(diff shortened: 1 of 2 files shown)


Shadowghost (approved):

nyanmisaka (approved):

MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
```diff
@@ -824,6 +824,22 @@ private async Task<string> ExtractImageInternal(string inputPath, string contain
                 options.EnableTonemapping = false;
             }
 
+            if (imageStream.Width is not null && imageStream.Height is not null)
+            {
+                // For hardware trickplay encoders, we need to re-calculate the size because they used fixed scale dimensions
+                var darParts = imageStream.AspectRatio.Split(":");
```
crobibero, line 830:
```suggestion
                var darParts = imageStream.AspectRatio.Split(':');
```

MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
```diff
@@ -824,6 +824,22 @@ private async Task<string> ExtractImageInternal(string inputPath, string contain
...
+                var (wa, ha) = (int.Parse(darParts[0], CultureInfo.InvariantCulture), int.Parse(darParts[1], CultureInfo.InvariantCulture));
```
crobibero, line 831: This assumes both parts of the aspect ratio exist, and that they are both integers.
What happens if the aspect ratio is something like `2.40:1`?

gnattu: As long as this string is not empty it should contain both parts as the probe result is already sanitized in most cases. Let's see if there is really weird videos like this.

crobibero (approved):

Bond-009 (approved):

PR #11816, opened 2024-05-24
Add file hash to images and remove duplicates after scan

### LINKED ISSUE #9456: [Issue]: NFO defined backdrops result in addition duplicates after each refresh all images

Shadowghost:
We are allowing multiple backdrops but never clean up duplicates after scan - these might appear if we have an image in the media folder and download the image from the NFO but they are the same.

**Changes**
* Use `XxHash3` to calcultate image file hash on scan if we haven't already
* Save hash to the database
* Remove duplicate backdrops on scan based on hash, preferring external images over ones in the JF metadata folder

**ToDo**
* Potentially write a migration, though it is fixed by a simple rescan

**Issues**
Fixes #9456
Replaces #9551

### DIFF
--- a/Emby.Server.Implementations/Library/LibraryManager.cs
+++ b/Emby.Server.Implementations/Library/LibraryManager.cs
@@ -1892,7 +1892,7 @@ namespace Emby.Server.Implementations.Library
         {
             if (image.Path is not null && image.IsLocalFile)
             {
-                if (image.Width == 0 || image.Height == 0 || string.IsNullOrEmpty(image.BlurHash))
+                if (image.Width == 0 || image.Height == 0 || string.IsNullOrEmpty(image.BlurHash) || string.IsNullOrEmpty(image.FileHash))
                 {
                     return true;
                 }
@@ -1916,9 +1916,11 @@ namespace Emby.Server.Implementations.Library
         {
             ArgumentNullException.ThrowIfNull(item);
 
+            var imagesWithPaths = item.ImageInfos.Where(i => i.Path is not null).ToArray();
             var outdated = forceUpdate
-                ? item.ImageInfos.Where(i => i.Path is not null).ToArray()
+                ? imagesWithPaths
                 : item.ImageInfos.Where(ImageNeedsRefresh).ToArray();
+
             // Skip image processing if current or live tv source
             if (outdated.Length == 0 || item.SourceType != SourceType.Library)
             {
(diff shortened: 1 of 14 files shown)


Emby.Server.Implementations/Data/SqliteItemRepository.cs
```diff
@@ -1224,10 +1233,14 @@ internal ItemImageInfo ItemImageInfoFromValueString(ReadOnlySpan<char> value)
...
-                    var length = value.Length;
+                    nextSegment = value.IndexOf(Delimiter);
+                    if (nextSegment == -1)
+                    {
+                        nextSegment = value.Length;
+                    }
 
-                    Span<char> blurHashSpan = stackalloc char[length];
-                    for (int i = 0; i < length; i++)
+                    Span<char> blurHashSpan = stackalloc char[nextSegment];
```
JPVenson, line 1242: i know its not your code but is a stackalloc really nesseary here?

Emby.Server.Implementations/Library/LibraryManager.cs
```diff
@@ -1938,6 +1950,26 @@ public async Task UpdateImagesAsync(BaseItem item, bool forceUpdate = false)
...
+            var metadataPath = item.GetInternalMetadataPath();
+            var distinctImages = imagesWithPaths
+                                    // Prefer extermal images
+                                    .OrderBy(i => !i.Path.StartsWith(metadataPath, StringComparison.Ordinal))
+                                    .DistinctBy(i => i.FileHash);
+            var imagesToRemove = imagesWithPaths.Where(i => !string.IsNullOrEmpty(i.FileHash)).Except(distinctImages).ToList();
+            if (imagesToRemove.Count > 0)
+            {
+                foreach (var image in imagesToRemove)
+                {
```
JPVenson, line 1963: i would put that in a try catch and aggregate all exceptions to delete as much as possible in the even of an error

gnattu:
This now requires a rebase to make CI happy, otherwise LGTM

Emby.Server.Implementations/Data/SqliteItemRepository.cs
gnattu: For such small fixed size byte arrays stackalloc is theoretically faster. I have no problem with its usage.

crobibero (approved):

gnattu (approved):

Emby.Server.Implementations/Library/LibraryManager.cs
```diff
@@ -1948,6 +1960,42 @@ public async Task UpdateImagesAsync(BaseItem item, bool forceUpdate = false)
                 }
             }
 
+            // Remove duplicate images
+            var metadataPath = item.GetInternalMetadataPath();
+            var distinctImages = imagesWithPaths
+                                    // Prefer extermal images
```
Bond-009, line 1966:
```suggestion
                                    // Prefer external images
```

Bond-009 (commented):
Is this worth it hashing every image added to the library? Why can't we compare filesize of the item images and only calculate the hash if the size matches? That doesn't complicate the db code even more.

Shadowghost:
Tbh I'm not sure anymore, has been some time since I proposed this.
Quite sure we discussed this on Element though and the result was to use hashing.

crobibero:
What is the status of this? Have we decided if this is still needed?

JPVenson (approved):

Shadowghost:
> Have we decided if this is still needed?

No we haven't.

theguymadmax:
The linked issue can no longer be reproduced, closing as stale.

Issue #11813, opened 2024-05-24
[Issue]: Database Becomes Locked On Startup


michaelharrigan:
### Please describe your bug

After updating to the latest server version, on startup, the database seems to getting locked preventing the server to finish its setup.

### Reproduction Steps

1. Start up jellyfin server on Mac
2. Choose to launch web UI
3. WebUI locks and nothing proceeds successfully

### Jellyfin Version

10.9.0

### if other:

10.9.2

### Environment

```markdown
- OS: MacOS 14.5
- Linux Kernel:
- Virtualization:
- Clients: Browser, AppleTV, iOS
- Browser: Safari
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking: Host
- Storage: Local
```

### Jellyfin logs

```shell
[2024-05-24 07:08:54.499 -04:00] [FTL] [1] Main: Error while starting server
Microsoft.Data.Sqlite.SqliteException (0x80004005): SQLite Error 5: 'database is locked'.
   at Microsoft.Data.Sqlite.SqliteException.ThrowExceptionForRC(Int32 rc, sqlite3 db)
   at Microsoft.Data.Sqlite.SqliteDataReader.NextResult()
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteReader(CommandBehavior behavior)
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteReader()
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteNonQuery()
   at Emby.Server.Implementations.Data.SqliteExtensions.Execute(SqliteConnection sqliteConnection, String commandText)
   at Emby.Server.Implementations.Data.BaseSqliteRepository.GetConnection()
   at Emby.Server.Implementations.Data.BaseSqliteRepository.Initialize()
   at Emby.Server.Implementations.Data.SqliteItemRepository.Initialize()
   at Emby.Server.Implementations.ApplicationHost.InitializeServices()
   at Jellyfin.Server.Program.StartServer(IServerApplicationPaths appPaths, StartupOptions options, IConfiguration startupConfig)
[2024-05-24 07:08:54.508 -04:00] [INF] [1] Main: Running query planner optimizations in the database... This might take a while
[2024-05-24 07:08:54.522 -04:00] [INF] [1] Emby.Server.Implementations.ApplicationHost: Disposing "CoreAppHost"
[...]

PR #11823, opened 2024-05-24
Add Env Var to disable second level cache

gnattu:
**Changes**

This is an attempt to track down possible causes of remaining database lockups. Add an environment variable to disable the second-level cache entirely to see if it works better on systems that still experience lockups.

For example (note the double underscore `__`  between `SQLITE` and `disableSecondLevelCache` is required):

```
export  JELLYFIN_SQLITE__disableSecondLevelCache=true
```

This is not a guaranteed fix, just an option for (advanced) users to test if it improves the situation. So it may be removed in future version.

**Issues**

See #11589 #11813 #11834

Probably also others I'm not aware of.

### DIFF
--- a/MediaBrowser.Controller/Extensions/ConfigurationExtensions.cs
+++ b/MediaBrowser.Controller/Extensions/ConfigurationExtensions.cs
@@ -141,18 +141,7 @@ namespace MediaBrowser.Controller.Extensions
         /// <returns>Whether second level cache disabled.</returns>
         public static bool GetSqliteSecondLevelCacheDisabled(this IConfiguration configuration)
         {
-            var disableSecondLevelCacheConfig = configuration.GetValue<string?>(SqliteDisableSecondLevelCacheKey);
-            var disableSecondLevelCache = false;
-            if (disableSecondLevelCacheConfig is not null)
-            {
-                disableSecondLevelCache = disableSecondLevelCacheConfig.ToUpperInvariant() switch
-                {
-                    "FALSE" or "NO" or "0" => false,
-                    _ => true
-                };
-            }
-
-            return disableSecondLevelCache;
+            return configuration.GetValue<bool>(SqliteDisableSecondLevelCacheKey);
         }
     }
 }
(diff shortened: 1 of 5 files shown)


MediaBrowser.Controller/Extensions/ConfigurationExtensions.cs
```diff
@@ -128,5 +133,26 @@ public static bool UseUnixSocket(this IConfiguration configuration)
...
             => configuration.GetValue<int?>(SqliteCacheSizeKey);
+
+        /// <summary>
+        /// Gets whether second level cache disabled from the <see cref="IConfiguration" />.
+        /// </summary>
+        /// <param name="configuration">The configuration to read the setting from.</param>
+        /// <returns>Whether second level cache disabled.</returns>
+        public static bool GetSqliteSecondLevelCacheDisabled(this IConfiguration configuration)
+        {
+            var disableSecondLevelCacheConfig = configuration.GetValue<string?>(SqliteDisableSecondLevelCacheKey);
```
cvium, line 144: Can't we use `GetValue<bool>(SqliteDisableSecondLevelCacheKey, false)`?

gnattu: We can, but that will make this mush less user-friendly, as the user will have to type in a valid bool string (TRUE or FALSE, case insensitive) and all other value will make dotnet complain and error out. This is more user friendly.

crobibero: This shouldn't be a user-friendly config option, if they change the config then they must conform to the accepted values.

gnattu: But not even supporting 1 and 0 which is super common for env vars is a little bit too restrictive.

Shadowghost (approved):

MediaBrowser.Controller/Extensions/ConfigurationExtensions.cs
joshuaboniface: I'd agree with gnattu here, sure it's not *common* to use but making it deliberately obtuse to use is not good design. If this is needed to get sensible boolean-like flags (e.g. `true`, `TRUE`, `1`, etc.) then I think it's worthwhile.

1337joe (approved):

crobibero (approved):

PR #11833, opened 2024-05-25
Use bluray protocol to play BDMVs

Bond-009:
This is way faster than using BDInfo to create concat file with all the correct segments. FFmpeg picks the longest playlist by default, same as Jellyfin.

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -1250,7 +1250,7 @@ namespace MediaBrowser.Controller.MediaEncoding
                 arg.Append(canvasArgs);
             }
 
-            if (state.MediaSource.VideoType == VideoType.Dvd || state.MediaSource.VideoType == VideoType.BluRay)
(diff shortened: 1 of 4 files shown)


Shadowghost:
Before I implemented the concat filter usage this was how we used to play BDMVs.
But from the testing and issues back then FFmpeg often picked incomplete playlist, resulting in only parts of the movie being played. This was especially problematic when you've used a BDMV with multiple eisodes for TV series.

If the probing is causing significant delays, we can save the list of m2ts files on scan (as we do for external audio and subs) and just use that.

Bond-009:
Did you use the skip estimate duration from PTS? This calculated a wrong duration in my case.
Also you can specify the exact playlist instead of letting ffmpeg pick the longest.

Shadowghost:
I just tested what we had in the code back then. It didn't use any special options, just changed the input handler

stanionascu:
Using the ffmpeg Blu-ray protocol is the only way to play the iso and bdmv folders correctly, especially with branched playlists. For BDJ discs, libbluray needs Java in the system.

This is the way kodi + vlc and others can correctly play them.

Issue #11476, opened 2024-05-01
[Issue]: Cleanup the trickplay DB table


enter-a-random-username:
### Please describe your bug

Changing setting, rename a file or anything else that qualify for a regeneration and the images will be deleted in file system, but the entries in the DB will remain. I run also cleanup task.
`Scan media library` handles the clean up of the images, it should also clean the DB or one of the `Maintenance` scheduled tasks

### Reproduction Steps

1. Generate a trickplay
2. Change tiling or/and rename a file
3. Run trickplay generation
4. Verify with `find metadata/library -name trickplay -type d | wc -l ` vs DB rows

### Jellyfin Version

Unstable (master branch)

### Environment

```markdown
- OS:
- Linux Kernel:
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

```shell
It is not a exception or error, it leftovers/missing clean up
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Shadowghost:
Not cleaning up different solutions is intentional afaik, but we should cleanup the entries if the file changes (I think I only implemented deleting the trickplay files, will take a look)

enter-a-random-username:
I think the current state is a massive improvement. I would love to see cleanup for removed folder like test of 1280x720 or so. The folder is gone after generating new image, I cleaned the old folders. Finding a good size for your devices takes a few tries.

I will close the issue. I can clean up the remaining entries by myself.

kodifan:
I’m facing a similar issue. I initially created trickplay images at a 320 resolution. Later, I changed the setting to 640, but no matter what I do, I can’t get it to work.

I’ve tried refreshing the library, re-adding it, and regenerating trickplay images multiple times. I even deleted the all trickplay folders that contained the 320-resolution images.

Now, I don’t have any trickplay images at all—just a gray 320 box instead of actual images.

It seems the problem is that the 320-resolution entries are still stored in the database.

kodifan:
ok i managed to delete those 320 entries in database and everything works now  - but it is diffidently  a bug i trickplay management process.

PR #11883, opened 2024-05-30
Enhance Trickplay

### LINKED ISSUE #11747: [Issue]: trickplay files are stored in /config/metadata/library and not with media
### Please describe your bug

The change from the original trickplay plugin to the "in box" trickplay plugin brought a change that the trickplay files are only stored in /config/metadata/library and do not have on option for storing with media.

As the trickplay files are stored in /config/metadata/library they have to be created on each jellyfin instance, for a clean deployment of jellyfin this can a significant time to complete, and has to be done for each new instance deployed.

As the trickplay files are stored in /config/metadata/library this also significantly increases the size of a backup of the /config

I deployed a clean instance of 10.9.1 on dedicated hardware with an i5-12500 CPU & QSV (UHD770) enabled for trickplay. With dedicated 1G connection to NFS storage (which can saturate the link) after 7 days I am ~40% complete with a library size of ~40G, so expected library size will be ~100GB when completed after an expected ~21 days.

### Reproduction Steps

1. deploy Jellyfin (in this case docker)
2. scan library with trickplay enabled
3. watch /config/metadata/library grow

### Jellyfin Version

10.9.0

### if other:

10.9.1 (not in dropdown)

### Environment

```markdown
- OS: Linux (Ubuntu 22.04)
- Linux Kernel: 6.5.0-35-generic
- Virtualization: Docker
- Clients: Browser
- Browser: Edge 124 
- FFmpeg Version: as per Docker
- Playback Method: n/a
- Hardware Acceleration: QSV
- GPU Model:  Intel Alderlake_s (Gen12) (UHD Graphics 770)
- Plugins: only built in
- Reverse Proxy: none
- Base URL: none
- Networking: Host
- Storage: local for config, NFS for media
```

### Jellyfin logs

```shell
n/a
```

### FFmpeg logs

```shell
n/a
```

n/a

n/a

### Code of Conduct

Shadowghost:
Companion to https://github.com/jellyfin/jellyfin-web/pull/5635

**Changes**
* Allow trickplay images to be saved in the media folder
* Add library option for the above
* Add dedicated option on refresh to replace trickplay images instead of relying on `replace all images`
* Add task to move existing trickplay images after library option was changed
* Migrate to better folder schema (`{Width} - {TileWidth}x{TileHeight}` to be portable

ToDo / Questions:
* Should the trickplay controller try the internal/media dir if the other returns no trickplay?

**Issues**
Fixes #11747


Simsala91:
´´´
Allow trickplay images to be saved in the media folder
´´´

How is this implemented? Are the trickplay images stored along side each episode/movie like this:

´´´
\\media\movies\movie1\movie1.mkv
\\media\movies\movie1\trickplay\0.jpg
´´´

or can you specify a folder that is located on the media drive where every trickplay image will be saved?

gnattu:
> ´´´ Allow trickplay images to be saved in the media folder ´´´
> 
> How is this implemented? Are the trickplay images stored along side each episode/movie like this:

Current implementation uses a suffixed folder with the same name as the original video file.

For you example: 

```
\media\movies\movie1\movie1.mkv
```

will have its own trickplay folder:

```
\media\movies\movie1\movie1.trickplay
```

And inside that folder there will be all the trickplay images for that video:

```
\media\movies\movie1\movie1.trickplay\{resolution}-{tilesize}\0.jpg
```

Shadowghost:
That's up for question right now. I'd say we should check both locations and only respect the setting when it comes to generating and saving/replacing trickplay images.

thermionic:
I will be very happy to see this completed.

michaelcurry:
Was just searching for this functionality within Jellyfin 10.9.6, and came across this PR.  Super cool! 
Thanks for making this happen @Shadowghost 🎉

sjorge:
> When can we expect a merge on this? I'm also very excited to use this feature.

It doesn't have the backport label, so I don't think it will end up in the next 10.9.x release.

RandomNamer:
Thank you for implementing this feature🎉 As someone who has multiple servers that access the same media library storage, I've been waited for a long time for this. One little question here: It seems currently Jellyfin determines tile size and interval from the server options, does that mean users have to keep trickplay configs of multiple servers the same when using the same trickplay copy? Or can we just save `localTrickPlayInfo` to local folder to help other servers read these trickplay images properly?

Shadowghost:
I was thinking about this too and for at least the first implementation the prerequisite will be having the same config. This is because the config is global and if we have multiple different resolutions inconsistently over different movies/series it is impossible to auto-import them.

thermionic:
> I was thinking about this too and for at least the first implementation the prerequisite will be having the same config. This is because the config is global and if we have multiple different resolutions inconsistently over different movies/series it is impossible to auto-import them.

If stored with media, could the size and "interval" be part of the path or name of the trickplay files, so that if a different size is "expected" from the config if would create (without overwriting) if not pre-existing?

From my perspective, the advantage of having them stored with media far outweighs what I would consider to be a minor inconvenience of having to have the trickplay config the same on all hosts that access the media.

Shadowghost:
We do not remove existing trickplay files, if the resolution does not exists, the files should be created on next scan/run of the scheduled task.

Emby.Server.Implementations/IO/ManagedFileSystem.cs
```diff
@@ -149,6 +149,26 @@ public virtual void CreateShortcut(string shortcutPath, string target)
...
+            {
+                Directory.Move(source, destination);
+            }
+            catch (IOException)
+            {
+                // Cross device move requires a copy
+                Directory.CreateDirectory(destination);
+                foreach (string file in Directory.GetFiles(source))
+                {
+                    File.Copy(file, Path.Combine(destination, Path.GetFileName(file)), true);
```
gnattu, line 165: I think we need to handle errors better here because various common user errors could happen. For example:

- The destination drive is full and only part of the files are moved (a common user error).
- The destination path is too long (common with network mounts).
- A trickplay file already existing in the destination folder (also common as some users share media between multiple instances).

Some of these issues, such as file already exists, should not be considered failures, and we can just delete the source folder. The problem here is that both "file exists" and "destination drive is full" will throw an `IOException`. If the performance allows, maybe we should checking if the file exists before performing the copy?

Shadowghost: I don't think we should be the ones responsible for handling `drive is full` but I do agree that the other two cases should be handled better. Will look into it.

sjorge: Do we handle destination is ro? This is also common.

gnattu: This will fail with a permission error and we will bail out. Which is totally fine to me as we can do nothing anyway.

Shadowghost: Just realized we are calling copy with `overrde = true` so case C is already handled.

soultaco83:
I built this PR and the web PR
 but I can not get the server to load. I get the error: 



[16:43:34] [FTL] [1] Main: Error while starting server
System.InvalidOperationException: Unable to resolve service for type 'Jellyfin.Server.Implementations.Trickplay.TrickplayManager' while attempting to activate 'Jellyfin.Server.Migrations.Routines.MoveTrickplayFiles'.
   at Microsoft.Extensions.DependencyInjection.ActivatorUtilities.ConstructorMatcher.CreateInstance(IServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ActivatorUtilities.CreateInstance(IServiceProvider provider, Type instanceType, Object[] parameters)
   at Jellyfin.Server.Migrations.MigrationRunner.<>c__DisplayClass2_0.<Run>b__0(Type m) in /mnt/e/Jellyfin/jellyfin-server/Jellyfin.Server/Migrations/MigrationRunner.cs:line 61
   at System.Linq.Enumerable.SelectArrayIterator`2.MoveNext()
   at System.Linq.Enumerable.OfTypeIterator[TResult](IEnumerable source)+MoveNext()
   at System.Collections.Generic.LargeArrayBuilder`1.AddRange(IEnumerable`1 items)
   at System.Collections.Generic.EnumerableHelpers.ToArray[T](IEnumerable`1 source)
   at Jellyfin.Server.Migrations.MigrationRunner.Run(CoreAppHost host, ILoggerFactory loggerFactory) in /mnt/e/Jellyfin/jellyfin-server/Jellyfin.Server/Migrations/MigrationRunner.cs:line 60
   at Jellyfin.Server.Program.StartServer(IServerApplicationPaths appPaths, StartupOptions options, IConfiguration startupConfig) in /mnt/e/Jellyfin/jellyfin-server/Jellyfin.Server/Program.cs:line 160
[16:43:34] [INF] [1] Main: Running query planner optimizations in the database... This might take a while

Shadowghost:
@soultaco83 should be fixed now.

JPVenson:
> Should the trickplay controller try the internal/media dir if the other returns no trickplay?

I would be against any sort of retry logic. If we have a migration strategy to move the files after the option has changed, its enough. Dont overcomplicate stuff here, also that can get tricky to debug and unintuitive.

Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
```diff
@@ -144,14 +204,53 @@ private async Task RefreshTrickplayDataInternal(
...
-                if (!replace && Directory.Exists(outputDir) && (await GetTrickplayResolutions(video.Id).ConfigureAwait(false)).ContainsKey(actualWidth))
+                // Import existing trickplay tiles
+                if (!replace && Directory.Exists(outputDir))
                 {
-                    _logger.LogDebug("Found existing trickplay files for {ItemId}. Exiting", video.Id);
-                    return;
+                    var existingFiles = Directory.GetFiles(outputDir);
+                    if (existingFiles.Length > 0)
+                    {
+                        var existingTrickplayResolution = (await GetTrickplayResolutions(video.Id).ConfigureAwait(false)).ContainsKey(actualWidth);
```
JPVenson, line 218: i dont like quering the whole dataset just to check for one key. Maybe build a dedicated `Exist` method

Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
```diff
@@ -76,7 +76,65 @@ public TrickplayManager(
...
+    private void MoveContent(string sourceFolder, string destinationFolder)
+    {
+        _fileSystem.MoveDirectory(sourceFolder, destinationFolder);
+        var parent = Directory.GetParent(sourceFolder);
+        if (parent is not null)
+        {
+            var parentContent = Directory.GetDirectories(parent.FullName);
+            if (parentContent.Length == 0)
+            {
+                Directory.Delete(parent.FullName);
```
JPVenson, line 131:
```suggestion
                Directory.Delete(parent.FullName, true);
```
To delete subdirectories too

Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
```diff
@@ -392,9 +509,15 @@ public async Task<Dictionary<string, Dictionary<int, TrickplayInfo>>> GetTrickpl
     }
 
     /// <inheritdoc />
-    public string GetTrickplayTilePath(BaseItem item, int width, int index)
+    public async Task<string> GetTrickplayTilePath(BaseItem item, int width, int index, bool saveWithMedia)
```
JPVenson, line 512:
```suggestion
    public async Task<string> GetTrickplayTilePathAsync(BaseItem item, int width, int index, bool saveWithMedia)
```

Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
```diff
@@ -355,6 +454,24 @@ public async Task<Dictionary<int, TrickplayInfo>> GetTrickplayResolutions(Guid i
         return trickplayResolutions;
     }
 
+    /// <inheritdoc />
+    public async Task<List<Guid>> GetTrickplayItems()
```
JPVenson, line 458:
```suggestion
    public async Task<List<Guid>> GetTrickplayItemsAsync()
```

Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
Shadowghost: We only delete the dir if we don't have subdirs, so not necessaray.

Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
```diff
@@ -612,4 +612,19 @@ public string GetTrickplayDirectory(BaseItem item, int tileWidth, int tileHeight
 
         return Path.Combine(path, subdirectory);
     }
+
+    private async Task<bool> HasTrickplayResolution(Guid itemId, int width)
```
JPVenson, line 616:
```suggestion
    private async Task<bool> HasTrickplayResolutionAsync(Guid itemId, int width)
```

JPVenson (approved):

Jellyfin.Api/Controllers/TrickplayController.cs
```diff
@@ -92,8 +92,18 @@ public ActionResult GetTrickplayTileImage(
...
 
-        var path = _trickplayManager.GetTrickplayTilePath(item, width, index);
-        if (System.IO.File.Exists(path))
+        var saveWithMedia = _libraryManager.GetLibraryOptions(item).SaveTrickplayWithMedia;
+        if (saveWithMedia)
+        {
+            var withMediaPath = await _trickplayManager.GetTrickplayTilePathAsync(item, width, index, saveWithMedia).ConfigureAwait(false);
+            if (!string.IsNullOrEmpty(withMediaPath) && System.IO.File.Exists(withMediaPath))
+            {
+                return PhysicalFile(withMediaPath, MediaTypeNames.Image.Jpeg);
```
crobibero, line 101:
```suggestion
				Response.Headers.ContentDisposition = "attachment";
                return PhysicalFile(withMediaPath, MediaTypeNames.Image.Jpeg);
```

MediaBrowser.Controller/Trickplay/ITrickplayManager.cs
```diff
@@ -42,6 +43,12 @@ public interface ITrickplayManager
     /// <returns>Map of width resolutions to trickplay tiles info.</returns>
     Task<Dictionary<int, TrickplayInfo>> GetTrickplayResolutions(Guid itemId);
 
+    /// <summary>
+    /// Gets the item ids of all items with trickplay info.
+    /// </summary>
+    /// <returns>Map of width resolutions to trickplay tiles info.</returns>
```
crobibero, line 49:
```suggestion
    /// <returns>The list of item ids that have trickplay info.</returns>
```

Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
```diff
@@ -355,6 +454,24 @@ public async Task<Dictionary<int, TrickplayInfo>> GetTrickplayResolutions(Guid i
         return trickplayResolutions;
     }
 
+    /// <inheritdoc />
+    public async Task<List<Guid>> GetTrickplayItemsAsync()
```
crobibero, line 458:
```suggestion
    public async Task<IReadOnlyList<Guid>> GetTrickplayItemsAsync()
```

Also update interface

Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
```diff
@@ -470,29 +593,38 @@ public string GetTrickplayTilePath(BaseItem item, int width, int index)
         return null;
     }
 
-    private string GetTrickplayDirectory(BaseItem item, int? width = null)
+    /// <inheritdoc />
+    public string GetTrickplayDirectory(BaseItem item, int tileWidth, int tileHeight, int width, bool saveWithMedia = false)
     {
         var path = Path.Combine(item.GetInternalMetadataPath(), "trickplay");
+        if (saveWithMedia)
```
crobibero, line 600: 599-604 can be changed to

```c#
var path = saveWithMedia
            ? Path.Combine(item.ContainingFolderPath, Path.ChangeExtension(item.Path, ".trickplay")) 
            : Path.Combine(item.GetInternalMetadataPath(), "trickplay");
```

Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
```diff
@@ -470,29 +593,38 @@ public string GetTrickplayTilePath(BaseItem item, int width, int index)
...
-        }
-        catch (IOException)
-        {
-            // Cross device move requires a copy
-            Directory.CreateDirectory(destination);
-            foreach (string file in Directory.GetFiles(source))
-            {
-                File.Copy(file, Path.Join(destination, Path.GetFileName(file)), true);
-            }
+        var trickplayResolutions = new Dictionary<int, TrickplayInfo>();
```
crobibero, line 618:
```suggestion
```

unused

gnattu (changes requested):

Shadowghost:
@crobibero we need the migration because for the internal path the layout changed too, so we need to migrate the existing files.

crobibero (approved):

gnattu (approved):
We might also want to have a scheduled task to handle #11476 which will be more common after this change

sjorge:
> We might also want to have a scheduled task to handle #11476 which will be more common after this change

Maybe something for one big cleanup task https://github.com/jellyfin/jellyfin/issues/12427

Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
```diff
@@ -144,14 +204,53 @@ private async Task RefreshTrickplayDataInternal(
                     actualWidth = 2 * ((int)mediaSource.VideoStream.Width / 2);
                 }
 
-                var outputDir = GetTrickplayDirectory(video, actualWidth);
+                var tileWidth = options.TileWidth;
+                var tileHeight = options.TileHeight;
+                var saveWithMedia = libraryOptions is not null ? libraryOptions.SaveTrickplayWithMedia : false;
```
Bond-009, line 209:
```suggestion
                var saveWithMedia = libraryOptions is null ? false : libraryOptions.SaveTrickplayWithMedia;
```

Jellyfin.Server/Migrations/Routines/MoveTrickplayFiles.cs
```diff
@@ -0,0 +1,71 @@
...
+
+    /// <inheritdoc />
+    public void Perform()
+    {
+        var trickplayItems = _trickplayManager.GetTrickplayItemsAsync().GetAwaiter().GetResult();
+        foreach (var itemId in trickplayItems)
+        {
+            var resolutions = _trickplayManager.GetTrickplayResolutions(itemId).GetAwaiter().GetResult();
+            var item = _libraryManager.GetItemById(itemId);
+            if (item is not null)
```
Bond-009, line 50: continue on null instead, less indentation

Bond-009 (changes requested):

Jellyfin.Api/Controllers/TrickplayController.cs
```diff
@@ -92,8 +92,19 @@ public ActionResult GetTrickplayTileImage(
...
+            var withMediaPath = await _trickplayManager.GetTrickplayTilePathAsync(item, width, index, saveWithMedia).ConfigureAwait(false);
+            if (!string.IsNullOrEmpty(withMediaPath) && System.IO.File.Exists(withMediaPath))
+            {
+                Response.Headers.ContentDisposition = "attachment";
+                return PhysicalFile(withMediaPath, MediaTypeNames.Image.Jpeg);
+            }
+        }
+
+        var path = await _trickplayManager.GetTrickplayTilePathAsync(item, width, index, false).ConfigureAwait(false);
+        if (!string.IsNullOrEmpty(path) && System.IO.File.Exists(path))
```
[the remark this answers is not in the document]
Shadowghost: Currently we fall back to the itnernal path if the external return nothing. But as Venson mentioned above, we should not do this anyways.

Bond-009 (approved):

SabinVI:
So this feature is complete now? It’s ready and will be available to use in Jellyfin 10.10? It won’t be available in any 10.9.x releases?

Does anyone have a timeline for when we can expect 10.10?

crobibero:
No, this PR will not be available in 10.9.x.
No, there is no timeline in which you can expect 10.10.

SabinVI:
> No, this PR will not be available in 10.9.x. No, there is no timeline in which you can expect 10.10.

Damn. I'm really excited to use this feature on my server. I just can't use it until I have this update because of my limited storage on the drive Jellyfin is installed on.

Thanks for the update though! I guess its another thing I'll look forward to when 10.10 does get released one day. :)

soultaco83:
Just got done building this and the Webui(left a note there about a change). This is a perfect addon. Thank you.

eomanis:
The way I understand this enhancement is that, once finished, it will be possible to store the Trickplay images in
 - The "metadata" path, subdirectory "library" (as is currently the case)
 - The media libraries' file trees themselves

I neither want the "metadata" directory to blow up in size on my SSD, nor do I want Jellyfin cluttering up the media directories with Trickplay images. It does not have write permissions there anyway, principle of least privilege and all.

Having Trickplay files sprayed into the media directories would at the very least give me the ick, not to mention that it would also complicate directory listings for my FTP users.

What I would like is to have a designated Trickplay images directory at a custom path "somewhere off to the side" on the large media file system. Would that be possible?

If there is no dedicated option for a custom Trickplay images path, do you think a symbolic link from within the "metadata" directory would work?

Shadowghost:
Symlinks should always work. No reason why they aren't as long as your permissions are set correctly.

eomanis:
> Symlinks should always work. No reason why they aren't as long as your permissions are set correctly.

Within "metadata/library" there is what looks like a file-hash-based directory tree though. Assuming Jellyfin puts the Trickplay images into that tree and not into a dedicated subdirectory, symlinking the Trickplay images away will not be possible.

brkr1:
So, is it working on 10.10.0? I'm only waiting for it to update from 10.8.13.

coreykovacs1:
This release is great. Definitely prefer the trickplay files stored within the media folders. One issue I am having though is that the trickplay files in the media folders aren't portable between servers. The new server on the trickplay scan is capable of detecting the trickplay files but doesn't get the metadata stored in the trickplayinfos table correct. In my experience it doesn't change the Height value by the number of rows of tiles in the output JPG files.

Would it be possible to have the correct metadata added to the trickplayinfos folder easily?

flashlab:
> This release is great. Definitely prefer the trickplay files stored within the media folders. One issue I am having though is that the trickplay files in the media folders aren't portable between servers. The new server on the trickplay scan is capable of detecting the trickplay files but doesn't get the metadata stored in the trickplayinfos table correct. In my experience it doesn't change the Height value by the number of rows of tiles in the output JPG files.
> 
> Would it be possible to have the correct metadata added to the trickplayinfos folder easily?

As my opinion, you should create an issue which point here instead of comment under a merged commit, to make sure more people can see it.

Bryan792:
I do have  #12887 up

thegranddesign:
@Shadowghost thank you SO MUCH for doing this!! Especially the ability to move trickplay images next to media!

I had one quick question. How difficult would it be to have trickplay images be segregated under a `.trickplay`  folder?

So the structure would be more like:

```
<library_directory>/<movie_or_show_directory>/.trickplay/<media_item_base_filename>.trickplay/<trickplay_resolution> <trickplay_tile_height>x<trickplay_tile_width>
```

Why would I want this?

For something like a movie, where there is only one media file in a directory, it doesn't really matter, but some "shows" can contain dozens of "episodes". Having all of those folders in the same directory kind of clutters everything up. At the end of the day, the thing that I care about is that the trickplay images are in the same directory as the media file, but I don't really need to see them 99% of the time.

thegranddesign:
Would anyone be opposed to making this the default?

Shadowghost:
Yes. A lot of people mount their media read only for consumption, so saving anything next to it is impossible.

kevoc:
@Simsala91 if your Jellyfin instance is running on Linux, you can already achieve what you're looking for using `overlayfs`. Your media folder with the movie files will be the lower dir, and you can have a different folder for your trickplay files as upper dir. Any writes or changes made by Jellyfin only change the upper dir. 

Disclaimer: I haven't done this yet, but intend to. It works in theory. As Shadowghost pointed out, I'm one of those people whose media collection is mounted read-only, so I'll be using `overlayfs` to put a writable layer on top.

eomanis:
> @Simsala91 if your Jellyfin instance is running on Linux, you can already achieve what you're looking for using `overlayfs`. (...)

Requiring overlayfs for a Jellyfin feature that could conceivably be implemented within Jellyfin seems excessive though.

kevoc:
@eomanis I agree, in the universe where Jellyfin has infinite funding and infinite developers. 

On Planet Earth, highly custom requests should be approached with pragmatism.

eomanis:
> @eomanis I agree, in the universe where Jellyfin has infinite funding and infinite developers.
> 
> On Planet Earth, highly custom requests should be approached with pragmatism.

Also true.

Please do not read the following as greedy; *who writes the code gets to decide*.

I am mentally hung up about the fact that someone would prefer a solution that 
 - Requires entirely new logic for calculating file paths, which better not have any glitches because it dangerously operates among the precious media files
 - Pollutes the media directories with ephemeral files (yuck)
 - Requires security concessions (read-write media) or complex and non-trivial operating system specific workarounds (overlayfs) to function if security of the media files should not be compromised

to a solution that just re-uses the Library directory's logic of path calculation below a user-customizable root directory. I. e. an "extra Library directory for Trickplay files" kind of thing.

Of those two ways to solve the request "have the Trickplay images in a custom location" I subjectively view the first one as "highly custom" and the latter as "pragmatic".

thegranddesign:
> Yes. A lot of people mount their media read only for consumption, so saving anything next to it is impossible.

That's not what I meant. What I meant was that _for people who choose to place trickplay images next to their media,_ could we not make it the default so that all trickplay images are within a single `trickplay` subdirectory?

And yes, I agree with others that _if the trickplay images were a single file,_ placing it next to the media wouldn't be that bad, but that's not what's happening. Having them be within a directory means that, for almost everyone using it, your `ls` and desktop file browser is going to sort all of those folders at the top, followed by all of the files. That's what makes it cluttered.

Doing this for trickplay images is similar to Jellyfin looking for trailers in `trailers` directory as well as next to the media.

GlassedSilver:
> > Yes. A lot of people mount their media read only for consumption, so saving anything next to it is impossible.
> 
> That's not what I meant. What I meant was that _for people who choose to place trickplay images next to their media,_ could we not make it the default so that all trickplay images are within a single `trickplay` subdirectory?
> 
> And yes, I agree with others that _if the trickplay images were a single file,_ placing it next to the media wouldn't be that bad, but that's not what's happening. Having them be within a directory means that, for almost everyone using it, your `ls` and desktop file browser is going to sort all of those folders at the top, followed by all of the files. That's what makes it cluttered.
> 
> Doing this for trickplay images is similar to Jellyfin looking for trailers in `trailers` directory as well as next to the media.

Beyond me how others don't get that...

NiklasBr:
@Shadowghost how do I clear all trickplay images in my library using this?

NiklasBr:
I probably misread "Add dedicated option on refresh to replace trickplay images instead" then.

Shadowghost:
That option forces a regeneration of trickplay files on scan

NiklasBr:
So, it's impossible to regenerate all tricklplay images which were generated using the old method?

PR #11886, opened 2024-05-30
Fix multi-part album folder being detected as artist folder

gnattu:
In rare cases, such as when an artist only has multi-part albums, the multi-part album folder might be incorrectly detected as an artist folder, causing the album discs to be split.

Caveat 1: For some artists with names indistinguishable from an album part (for example, "Volume 10"), the folder name for the artist needs to be escaped, like `[Volume 10]`.

Caveat 2: Users already affected by this might see multiple duplicated albums, both the wrongly split ones and the correctly grouped ones, after a library-specific scan. This can be fixed by performing a global validation in the scheduled task -> scan media library. Only the valid albums will remain after that is completed.

**Changes**

- Don't mark folder as artist folder when the folder appears to be a multi-part album.

**Issues**

### DIFF
--- a/Emby.Server.Implementations/Library/Resolvers/Audio/MusicArtistResolver.cs
+++ b/Emby.Server.Implementations/Library/Resolvers/Audio/MusicArtistResolver.cs
@@ -3,6 +3,7 @@
 using System;
 using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


Shadowghost (approved):

SenorSmartyPants:
Should update test cases.

PR #11896, opened 2024-05-31
Trim leading and trailing whitespace from item search term.

### LINKED ISSUE #11868: [Issue]: Search in webUI does not trim leading or trailing spaces, returning no results; Does not search year
### Please describe your bug

When entering a search term in the search box, ie: " Sleeping Beauty" and "Sleeping Beauty " result in no search results.  Only when removing the trailing and leading space do we get the result "Sleeping Beauty".  These should be trimmed for leading whitespace (spaces, tabs, etc) prior to being searched.  This occurs for searching people or in titles.

_Aside:_ Also noticed, "Sleeping Beauty 1959" and "Sleeping Beauty (1959)" will also return no results. I expected this to work in my testing and was surprised when it didn't.

### Reproduction Steps

1. Click on the magnifying glass in the WebUI
2. Enter a movie name with a leading or trailing space
3. See no results
4. Remove the space to find the correct results.

### Jellyfin Version

10.9.0

### if other:

_No response_

### Environment

```markdown
- OS: Debian Linux 12
- Linux Kernel: 6.2.x
- Virtualization: Docker
- Clients: WebUI
- Browser: Chrome
- FFmpeg Version: N/A
- Playback Method: N/A
- Hardware Acceleration: N/A
- GPU Model: N/A
- Plugins: N/A
- Reverse Proxy: Traefik
- Base URL: /
- Networking: Docker IPv4
- Storage: Local ZFS
```

### Jellyfin logs

```shell
N/A
```

### FFmpeg logs

```shell
N/A
```

N/A

N/A

### Code of Conduct

riashg:
**Changes**
Trim the leading and trailing whitespace from the searchTerm used in the ItemsController GetItems() method.

**Issues**
Fixes #11868

### DIFF
--- a/Jellyfin.Api/Controllers/ItemsController.cs
+++ b/Jellyfin.Api/Controllers/ItemsController.cs
@@ -370,7 +370,7 @@ public class ItemsController : BaseJellyfinApiController
                 EnableTotalRecordCount = enableTotalRecordCount,
                 ExcludeItemIds = excludeItemIds,
                 DtoOptions = dtoOptions,
-                SearchTerm = searchTerm,
+                SearchTerm = !string.IsNullOrEmpty(searchTerm) ? searchTerm.Trim() : searchTerm,
                 MinDateLastSaved = minDateLastSaved?.ToUniversalTime(),
                 MinDateLastSavedForUser = minDateLastSavedForUser?.ToUniversalTime(),
                 MinPremiereDate = minPremiereDate?.ToUniversalTime(),


gnattu:
I think this should be implemented by clients and not to put it on the server like this.

JPVenson:
> I think this should be implemented by clients and not to put it on the server like this.

I dont think it hurts to do it on the server to keep a more generalised resultset. It costs next to nothing and only helps because if some client forgets it we have it covered.

JPVenson (approved):

crobibero (changes requested):
I think a proper place to make this change would be in `SqliteItemRepository` in the `BindSearchParams` function.

theguymadmax:
This issue has been resolved: #14879

PR #11901, opened 2024-06-01
Implement Device Cache to replace EFCoreSecondLevelCacheInterceptor

gnattu:
The EFCoreSecondLevelCacheInterceptor will place a huge lock even for reading. Implement a ConcurrentDictionary cache to replace it.

**Changes**

**Issues**

Closes #11823

### DIFF
--- a/Jellyfin.Server.Implementations/Devices/DeviceManager.cs
+++ b/Jellyfin.Server.Implementations/Devices/DeviceManager.cs
@@ -123,7 +123,7 @@ namespace Jellyfin.Server.Implementations.Devices
         /// <inheritdoc />
         public DeviceInfo? GetDevice(string id)
         {
-            var device = _devices.Values.OrderByDescending(d => d.DateLastActivity).FirstOrDefault(d => d.DeviceId == id);
+            var device = _devices.Values.Where(d => d.DeviceId == id).OrderByDescending(d => d.DateLastActivity).FirstOrDefault();
             _deviceOptions.TryGetValue(id, out var deviceOption);
 
             var deviceInfo = device is null ? null : ToDeviceInfo(device, deviceOption);
--- a/Jellyfin.Server.Implementations/Devices/DeviceManager.cs
+++ b/Jellyfin.Server.Implementations/Devices/DeviceManager.cs
@@ -172,10 +172,9 @@ namespace Jellyfin.Server.Implementations.Devices
         /// <inheritdoc />
         public QueryResult<DeviceInfo> GetDevicesForUser(Guid? userId)
         {
-            var devices = _devices.Values
+            IEnumerable<Device> devices = _devices.Values
                 .OrderByDescending(d => d.DateLastActivity)
-                .ThenBy(d => d.DeviceId)
-                .AsEnumerable();
+                .ThenBy(d => d.DeviceId);
 
             if (!userId.IsNullOrEmpty())
             {
--- a/Jellyfin.Server.Implementations/Devices/DeviceManager.cs
+++ b/Jellyfin.Server.Implementations/Devices/DeviceManager.cs
@@ -199,8 +199,7 @@ namespace Jellyfin.Server.Implementations.Devices
         /// <inheritdoc />
         public async Task DeleteDevice(Device device)
         {
-            var id = _devices.FirstOrDefault(x => x.Value.Equals(device)).Key;
-            _devices.TryRemove(id, out _);
+            _devices.TryRemove(device.Id, out _);
             var dbContext = await _dbProvider.CreateDbContextAsync().ConfigureAwait(false);
             await using (dbContext.ConfigureAwait(false))
             {
--- a/Jellyfin.Server.Implementations/Devices/DeviceManager.cs
+++ b/Jellyfin.Server.Implementations/Devices/DeviceManager.cs
@@ -97,12 +97,8 @@ namespace Jellyfin.Server.Implementations.Devices
             await using (dbContext.ConfigureAwait(false))
             {
                 dbContext.Devices.Add(device);
-
                 await dbContext.SaveChangesAsync().ConfigureAwait(false);
-                var newDevice = await dbContext.Devices
-                    .FirstAsync(d => d.Id == device.Id)
-                    .ConfigureAwait(false);
-                _devices.TryAdd(device.Id, newDevice);
+                _devices.TryAdd(device.Id, device);
             }
 
             return device;
--- a/Jellyfin.Server.Implementations/Devices/DeviceManager.cs
+++ b/Jellyfin.Server.Implementations/Devices/DeviceManager.cs
@@ -133,10 +133,11 @@ namespace Jellyfin.Server.Implementations.Devices
         /// <inheritdoc />
         public QueryResult<Device> GetDevices(DeviceQuery query)
         {
-            IEnumerable<Device> devices = _devices.Values.OrderBy(d => d.Id)
+            IEnumerable<Device> devices = _devices.Values
                 .Where(device => !query.UserId.HasValue || device.UserId.Equals(query.UserId.Value))
                 .Where(device => query.DeviceId == null || device.DeviceId == query.DeviceId)
                 .Where(device => query.AccessToken == null || device.AccessToken == query.AccessToken)
+                .OrderBy(d => d.Id)
                 .ToList();
             var count = devices.Count();
 
--- a/Jellyfin.Server.Implementations/Devices/DeviceManager.cs
+++ b/Jellyfin.Server.Implementations/Devices/DeviceManager.cs
@@ -100,9 +100,12 @@ namespace Jellyfin.Server.Implementations.Devices
                 dbContext.Devices.Add(device);
 
                 await dbContext.SaveChangesAsync().ConfigureAwait(false);
+                var newDevice = await dbContext.Devices
+                    .Include(d => d.User)
+                    .FirstOrDefaultAsync(d => d.Id == device.Id)
+                    .ConfigureAwait(false);
+                _devices.Add(device.Id, newDevice!);
             }
-
-            _devices.Add(device.Id, device);
 
             return device;
         }
--- a/Jellyfin.Server.Implementations/Devices/DeviceManager.cs
+++ b/Jellyfin.Server.Implementations/Devices/DeviceManager.cs
@@ -27,6 +27,8 @@ namespace Jellyfin.Server.Implementations.Devices
         private readonly IDbContextFactory<JellyfinDbContext> _dbProvider;
         private readonly IUserManager _userManager;
         private readonly ConcurrentDictionary<string, ClientCapabilities> _capabilitiesMap = new();
+        private readonly IDictionary<int, Device> _devices;
+        private readonly IDictionary<string, DeviceOptions> _deviceOptions;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="DeviceManager"/> class.
@@ -37,6 +39,24 @@ namespace Jellyfin.Server.Implementations.Devices
         {
             _dbProvider = dbProvider;
             _userManager = userManager;
+            _devices = new ConcurrentDictionary<int, Device>();
+            _deviceOptions = new ConcurrentDictionary<string, DeviceOptions>();
+
+            using var dbContext = _dbProvider.CreateDbContext();
+            foreach (var device in dbContext.Devices
+                         .Include(d => d.User)
+                         .OrderBy(d => d.Id)
+                         .AsEnumerable())
+            {
+                _devices.Add(device.Id, device);
+            }
+
+            foreach (var deviceOption in dbContext.DeviceOptions
+                         .OrderBy(d => d.Id)
+                         .AsEnumerable())
+            {
+                _deviceOptions.Add(deviceOption.DeviceId, deviceOption);
+            }
         }
 
         /// <inheritdoc />
@@ -66,6 +86,8 @@ namespace Jellyfin.Server.Implementations.Devices
                 await dbContext.SaveChangesAsync().ConfigureAwait(false);
             }
 
+            _deviceOptions[deviceId] = deviceOptions;
+
             DeviceOptionsUpdated?.Invoke(this, new GenericEventArgs<Tuple<string, DeviceOptions>>(new Tuple<string, DeviceOptions>(deviceId, deviceOptions)));
         }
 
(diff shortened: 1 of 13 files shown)


Jellyfin.Server.Implementations/Devices/DeviceManager.cs
```diff
@@ -108,57 +127,45 @@ public ClientCapabilities GetCapabilities(string deviceId)
...
-        public async Task<QueryResult<Device>> GetDevices(DeviceQuery query)
+        public QueryResult<Device> GetDevices(DeviceQuery query)
         {
-            var dbContext = await _dbProvider.CreateDbContextAsync().ConfigureAwait(false);
-            await using (dbContext.ConfigureAwait(false))
+            var devices = _devices.Values.OrderBy(d => d.Id)
+                .Where(device => !query.UserId.HasValue || device.UserId.Equals(query.UserId.Value))
+                .Where(device => query.DeviceId == null || device.DeviceId == query.DeviceId)
+                .Where(device => query.AccessToken == null || device.AccessToken == query.AccessToken);
+            var canGetCountDirectly = devices.TryGetNonEnumeratedCount(out var count);
```
cvium, line 146: This will always return false since `devices` is a LINQ query.

Jellyfin.Server.Implementations/Devices/DeviceManager.cs
```diff
@@ -167,38 +174,38 @@ public async Task<QueryResult<DeviceInfo>> GetDeviceInfos(DeviceQuery query)
...
-                var sessions = dbContext.Devices
-                    .Include(d => d.User)
-                    .OrderByDescending(d => d.DateLastActivity)
-                    .ThenBy(d => d.DeviceId)
-                    .SelectMany(d => dbContext.DeviceOptions.Where(o => o.DeviceId == d.DeviceId).DefaultIfEmpty(), (d, o) => new { Device = d, Options = o })
-                    .AsAsyncEnumerable();
+            var devices = _devices.Values
+                .OrderByDescending(d => d.DateLastActivity)
+                .ThenBy(d => d.DeviceId)
+                .AsEnumerable();
```
cvium, line 182: Redundant, it already is an IEnumerable

Jellyfin.Server.Implementations/Devices/DeviceManager.cs
```diff
@@ -37,6 +39,24 @@ public DeviceManager(IDbContextFactory<JellyfinDbContext> dbProvider, IUserManag
         {
             _dbProvider = dbProvider;
             _userManager = userManager;
+            _devices = new ConcurrentDictionary<int, Device>();
+            _deviceOptions = new ConcurrentDictionary<string, DeviceOptions>();
```
cvium, line 43: Would be better with a tuple i.e. `ConcurrentDictionary<int, (Device, DeviceOptions?)>()` as they are very tightly coupled

Jellyfin.Server.Implementations/Devices/DeviceManager.cs
```diff
@@ -27,6 +27,8 @@ public class DeviceManager : IDeviceManager
         private readonly IDbContextFactory<JellyfinDbContext> _dbProvider;
         private readonly IUserManager _userManager;
         private readonly ConcurrentDictionary<string, ClientCapabilities> _capabilitiesMap = new();
+        private readonly IDictionary<int, Device> _devices;
```
cvium, line 30:
```suggestion
        private readonly ConcurrentDictionary<int, Device> _devices;
```

Jellyfin.Server.Implementations/Devices/DeviceManager.cs
```diff
@@ -27,6 +27,8 @@ public class DeviceManager : IDeviceManager
...
+        private readonly IDictionary<string, DeviceOptions> _deviceOptions;
```
cvium, line 31:
```suggestion
        private readonly ConcurrentDictionary<string, DeviceOptions> _deviceOptions;
```

Jellyfin.Server.Implementations/Devices/DeviceManager.cs
```diff
@@ -167,38 +174,38 @@ public async Task<QueryResult<DeviceInfo>> GetDeviceInfos(DeviceQuery query)
...
+                }).ToArray();
+
+            return new QueryResult<DeviceInfo>(array);
         }
 
         /// <inheritdoc />
         public async Task DeleteDevice(Device device)
         {
+            var id = _devices.FirstOrDefault(x => x.Value.Equals(device)).Key;
+            _devices.Remove(id);
```
cvium, line 208:
```suggestion
            _devices.TryRemove(device.Id, out _);
```

Jellyfin.Server.Implementations/Devices/DeviceManager.cs
gnattu: It is an `IOrderedEncumerable` without this which cannot be implicitly converted to  `IEnumerable`

Jellyfin.Server.Implementations/Devices/DeviceManager.cs
gnattu: Well now it turns out it is not, as different devices could have same DeviceId for whatever reason and store them separately is easier for such many to one relationship.

Jellyfin.Server.Implementations/Devices/DeviceManager.cs
cvium: But an IOrderedEnumerable implements IEnumerable and the implementation for AsEnumerable just returns it as-is.

Jellyfin.Server.Implementations/Devices/DeviceManager.cs
```diff
@@ -167,38 +170,38 @@ public async Task<QueryResult<DeviceInfo>> GetDeviceInfos(DeviceQuery query)
...
+                    return ToDeviceInfo(device, option);
+                }).ToArray();
+
+            return new QueryResult<DeviceInfo>(array);
         }
 
         /// <inheritdoc />
         public async Task DeleteDevice(Device device)
         {
+            var id = _devices.FirstOrDefault(x => x.Value.Equals(device)).Key;
```
cvium, line 203: Is the Id not available in the input arg?

Jellyfin.Server.Implementations/Devices/DeviceManager.cs
gnattu: Other methods like `.Where` and `.Select` does not return `IOrderedEnumerable` so we need to explicitly declare the type of `devices` as `IEnumerable<Device>`, but should work

Jellyfin.Server.Implementations/Devices/DeviceManager.cs
gnattu: It is. Just use the id should work

DaRealHansForSure:
Thats a fucking Awesome Pull Request

I manually merged it into my Branch and my Server is Running since 25h without Issues, less Performance issues and MUCH MUCH less Memory.

Always when the DB Lock kicked in, a chain reaction happened and Memory increased huge.

Thank you

Jellyfin.Server.Implementations/Devices/DeviceManager.cs
```diff
@@ -78,23 +100,20 @@ await using (dbContext.ConfigureAwait(false))
                 dbContext.Devices.Add(device);
 
                 await dbContext.SaveChangesAsync().ConfigureAwait(false);
+                var newDevice = await dbContext.Devices
+                    .Include(d => d.User)
+                    .FirstOrDefaultAsync(d => d.Id == device.Id)
+                    .ConfigureAwait(false);
+                _devices.TryAdd(device.Id, newDevice!);
```
crobibero, line 107: Since we know an item is added, this can be `.First()` and the null-suppression can be removed

Jellyfin.Server.Implementations/Devices/DeviceManager.cs
```diff
@@ -37,6 +39,24 @@ public DeviceManager(IDbContextFactory<JellyfinDbContext> dbProvider, IUserManag
         {
             _dbProvider = dbProvider;
             _userManager = userManager;
+            _devices = new ConcurrentDictionary<int, Device>();
+            _deviceOptions = new ConcurrentDictionary<string, DeviceOptions>();
+
+            using var dbContext = _dbProvider.CreateDbContext();
+            foreach (var device in dbContext.Devices
+                         .Include(d => d.User)
```
crobibero, line 47: I think this has a chance to get out of sync with the User cache. Looks like only the username is actually used?

User might also get updated when the devices is updated, I haven't checked.

Jellyfin.Server.Implementations/Devices/DeviceManager.cs
```diff
@@ -108,57 +127,41 @@ public ClientCapabilities GetCapabilities(string deviceId)
...
-                    .Where(device => query.DeviceId == null || device.DeviceId == query.DeviceId)
-                    .Where(device => query.AccessToken == null || device.AccessToken == query.AccessToken);
-
-                var count = await devices.CountAsync().ConfigureAwait(false);
-
-                if (query.Skip.HasValue)
-                {
-                    devices = devices.Skip(query.Skip.Value);
-                }
+            var devices = _devices.Values.OrderBy(d => d.Id)
```
crobibero, line 142:
```suggestion
            IEnumerable<Device> devices = _devices.Values.OrderBy(d => d.Id)
```

and add a `.ToList()` to the end

No need to filter the full list of devices twice

Jellyfin.Server.Implementations/Devices/DeviceManager.cs
gnattu: This is a required relationship, but most of our logic does not depend on this yet. For current use case I think it would be good enough to not include this and query the user from user cache directly, as the userId should not change.

Jellyfin.Server.Implementations/Devices/DeviceManager.cs
```diff
@@ -239,14 +238,19 @@ public bool CanAccessDevice(User user, string deviceId)
         private DeviceInfo ToDeviceInfo(Device authInfo, DeviceOptions? options = null)
         {
             var caps = GetCapabilities(authInfo.DeviceId);
+            var user = _userManager.GetUserById(authInfo.UserId);
+            if (user is null)
+            {
+                throw new ResourceNotFoundException();
```
crobibero, line 244: Please add a message here

Jellyfin.Server.Implementations/Devices/DeviceManager.cs
```diff
@@ -108,57 +125,42 @@ public ClientCapabilities GetCapabilities(string deviceId)
...
+        public QueryResult<Device> GetDevices(DeviceQuery query)
         {
-            var dbContext = await _dbProvider.CreateDbContextAsync().ConfigureAwait(false);
-            await using (dbContext.ConfigureAwait(false))
+            IEnumerable<Device> devices = _devices.Values.OrderBy(d => d.Id)
+                .Where(device => !query.UserId.HasValue || device.UserId.Equals(query.UserId.Value))
+                .Where(device => query.DeviceId == null || device.DeviceId == query.DeviceId)
+                .Where(device => query.AccessToken == null || device.AccessToken == query.AccessToken)
+                .ToList();
+            var count = devices.Count();
```
crobibero, line 145:
```suggestion
            var count = devices.Count;
```

crobibero (approved):

Jellyfin.Server.Implementations/Devices/DeviceManager.cs
gnattu: We cannot do this as devices is a generic delegate, maybe `var count = ((List<Device>)devices).Count;` will do the trick?

crobibero: Ah right, I just remembered the `.ToList()`. Updating the `.Count()` is unneeded then

Jellyfin.Server.Implementations/Devices/DeviceManager.cs
```diff
@@ -78,23 +99,19 @@ await using (dbContext.ConfigureAwait(false))
                 dbContext.Devices.Add(device);
 
                 await dbContext.SaveChangesAsync().ConfigureAwait(false);
+                var newDevice = await dbContext.Devices
```
cvium, line 102: Why is it being retrieved here? `device` is already tracked by EF

gnattu: This is leftover code. We were caching user in device cache as well and the device passed in will not have a user attached to it. But now we are querying user directly from the user cache so this can be safely removed and add the device to the cache directly

Shadowghost (approved):

Jellyfin.Server.Implementations/Devices/DeviceManager.cs
```diff
@@ -108,57 +121,42 @@ public ClientCapabilities GetCapabilities(string deviceId)
...
-            await using (dbContext.ConfigureAwait(false))
-            {
-                var device = await dbContext.Devices
-                    .Where(d => d.DeviceId == id)
-                    .OrderByDescending(d => d.DateLastActivity)
-                    .Include(d => d.User)
-                    .SelectMany(d => dbContext.DeviceOptions.Where(o => o.DeviceId == d.DeviceId).DefaultIfEmpty(), (d, o) => new { Device = d, Options = o })
-                    .FirstOrDefaultAsync()
-                    .ConfigureAwait(false);
+            var device = _devices.Values.OrderByDescending(d => d.DateLastActivity).FirstOrDefault(d => d.DeviceId == id);
```
Bond-009, line 126: This should be faster in theory, no?
```suggestion
            var device = _devices.Values.Where(d => d.DeviceId == id).OrderByDescending(d => d.DateLastActivity).FirstOrDefault();
```

Jellyfin.Server.Implementations/Devices/DeviceManager.cs
```diff
@@ -108,57 +121,42 @@ public ClientCapabilities GetCapabilities(string deviceId)
...
-        public async Task<QueryResult<Device>> GetDevices(DeviceQuery query)
+        public QueryResult<Device> GetDevices(DeviceQuery query)
         {
-            var dbContext = await _dbProvider.CreateDbContextAsync().ConfigureAwait(false);
-            await using (dbContext.ConfigureAwait(false))
+            IEnumerable<Device> devices = _devices.Values.OrderBy(d => d.Id)
+                .Where(device => !query.UserId.HasValue || device.UserId.Equals(query.UserId.Value))
+                .Where(device => query.DeviceId == null || device.DeviceId == query.DeviceId)
+                .Where(device => query.AccessToken == null || device.AccessToken == query.AccessToken)
+                .ToList();
```
Bond-009, line 140: Same here
```suggestion
            IEnumerable<Device> devices = _devices.Values
                .Where(device => !query.UserId.HasValue || device.UserId.Equals(query.UserId.Value))
                .Where(device => query.DeviceId == null || device.DeviceId == query.DeviceId)
                .Where(device => query.AccessToken == null || device.AccessToken == query.AccessToken)
                .OrderBy(d => d.Id)
                .ToList();
```

Bond-009 (approved):
I'd filter before sorting

joshuaboniface:
This did miss 10.9.8 due to me not noticing it - sorry. It will land in 10.9.9, likely some time this week or next weekend.

cvium:
There's a workaround in the linked issue

JamsRepos:
> There's a workaround in the linked issue

Apologies, must've missed that. Much obliged 😀

Issue #11909, opened 2024-06-01
[Issue]: Audio Normalization fails on Radio Show/Podcast collections due to `OutOfMemoryException`


mihawk90:
### Please describe your bug

I noticed immediately after the upgrade that the Audio Normalization would eventually fail for me. In the logs I could see that it eventually hit an `OutOfMemoryException`:
```log
[2024-06-01 11:57:53.532 +00:00] [ERR] [10] Emby.Server.Implementations.ScheduledTasks.TaskManager: Error executing Scheduled Task
System.OutOfMemoryException: Insufficient memory to continue the execution of the program.
   at System.Text.StringBuilder.ExpandByABlock(Int32 minBlockCharCount)
   at System.Text.StringBuilder.AppendWithExpansion(Char& value, Int32 valueCount)
   at System.IO.StreamReader.ReadToEndAsyncInternal(CancellationToken cancellationToken)
   at Emby.Server.Implementations.ScheduledTasks.Tasks.AudioNormalizationTask.CalculateLUFSAsync(String inputArgs, CancellationToken cancellationToken)
   at Emby.Server.Implementations.ScheduledTasks.Tasks.AudioNormalizationTask.ExecuteAsync(IProgress`1 progress, CancellationToken cancellationToken)
   at Emby.Server.Implementations.ScheduledTasks.ScheduledTaskWorker.ExecuteInternal(TaskOptions options)
[2024-06-01 11:57:53.563 +00:00] [INF] [10] Emby.Server.Implementations.ScheduledTasks.TaskManager: "Audio Normalization" Failed after 53 minute(s) and 49 seconds
```

But because the Normalization process logs absolutely nothing it took me a while to get down to the cause.

With Debug logging I figured that it passes a `.concat` text file to ffmpeg's `concat` filter. The files seem to be written on a per-album basis (which in itself makes sense of course).
However the problem with that comes when one organises Radio Shows or Podcasts into a single Album. Luckily JF doesn't immediately clean up the text file after the task fails, so I was able to pull the offending album fairly easily:
[625f488c-7c25-4d0e-97ca-dcf0e93def36.concat.txt](https://github.com/user-attachments/files/15522079/625f488c-7c25-4d0e-97ca-dcf0e93def36.concat.txt)

As you can tell this list is quite long and it should be noted every episode is 2h minimum (more for specials). So whenever JF encounters this album it will eventually just run out of memory.

[...]

PR #11911, opened 2024-06-01
Log album name and id in normalization task

Bond-009:
Filename of the concat file is now the same as the album id.
Temp file gets deleted even if LUFS calculation failed


Shadowghost (approved):
Should this be targeting 10.9?

mihawk90:
IMHO it should. It took me hours of re-runs to find the reason for my Normalizations failing (see #11909), and that was mostly due to there being no logging on the task whatsoever.
I'm sure this is not going to be the last issue cropping up with Normalization so some basic logging would be nice to have ASAP. I'd wager not everyone has the inclination or patience to sit through 10 hours of retries just to find a potential issue.

oddstr13 (approved):

PR #11943, opened 2024-06-03
Increase lyrics migration batch size to 5000

Shadowghost:
100 was a bit too low, 5000 should speedup the migration.

**Changes**
* Increase batch size to 5000


Jellyfin.Server/Migrations/Routines/FixAudioData.cs
```diff
@@ -80,7 +80,7 @@ public void Perform()
                 {
                     IncludeItemTypes = [BaseItemKind.Audio],
                     StartIndex = startIndex,
-                    Limit = 100,
+                    Limit = 5000,
```
cvium, line 83: Two things:
1. startIndex must be updated as well and it should be updated with results.Length
2. Might be nice with some logging on execution time on the DB backup so we can see if that is the issue

crobibero (approved):

Bond-009 (approved):

Issue #11888, opened 2024-05-30
[Issue]: Very slow next up


FML128:
### Please describe your bug

Everything loads relatively slow, especially Next Up. (between ~9s - 60s until the whole page is loaded)

On the Android TV App and sometime on web it actually isn't even able to finish load next up.

On the exact same setup I also tried Emby, there everything loads within 2 seconds (same user data)

This is tested with my user with ~ 10k watched statuses.

Playback itself starts within a second or so

### Reproduction Steps

1. Load the page
2. Wait

### Jellyfin Version

10.9.0

### if other:

10.9.3

### Environment

```markdown
- OS: Arch Linux
- Linux Kernel: 6.9.2-arch1-1
- Virtualization: Docker
- Clients: everywhere
- Browser: everywhere
- FFmpeg Version: latest docker (official)
- Playback Method: N/A
- Hardware Acceleration: QSV
- GPU Model: Arc 770
- Plugins: AniDB, AudioDB, Cover Art Archive, Fanart, InfuseSync, Kitsu, Kodi Sync Queue, LDAP-Auth, MusicBrainz, OMDb, Playback Reporting, Reports, SSO-Auth, Session Cleaner, Studio Images, TMDb, TMDb Box Sets, TV Maze, TheTVDB, VGMdb
- Reverse Proxy: traefik
- Base URL: /
- Networking: Reverse proxy non host
- Storage: local (zfs ssd + hdd ca 150tb media)
```

### Jellyfin logs

```shell
[2024-05-30 22:02:11.652 +00:00] [INF] [70] Emby.Server.Implementations.HttpServer.WebSocketManager: WS "172.18.0.4" request
```

### Please attach any browser or client logs here

![image](https://github.com/jellyfin/jellyfin/assets/33214722/bde01cf8-d8ff-4493-9010-7b33dc9935dd)

### Please attach any screenshots here

![image](https://github.com/jellyfin/jellyfin/assets/33214722/09d96336-652c-4210-b9da-5460bfe5d7ee)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

jedthe3rd:
@FML128 I am curious if #11951 is related. 
Try disabling Group movies into collections in libraries > display settings:
![image](https://github.com/jellyfin/jellyfin/assets/7817335/677ef397-a455-4cd6-b26d-b05256709f5a)

FML128:
@jedthe3rd this didn't seem to make any difference.

[...]

Issue #11316, opened 2024-04-08
[Issue]: Next Up on home is extremely slow


pdf:
### Please describe your bug

I don't recall if this has always been a problem, I suspect it has but maybe I haven't noticed in the past, since I don't use the web UI often.

Upon loading the home page on the web interface, all widgets load on my dataset within a second or two, however the Next Up widget takes 15-20 seconds to load. During this time, the loading spinner remains overlaid on the display, which is an irritation.

As a quick and dirty test I added the following index on the library DB:

```sql
CREATE INDEX idx_test_UserDataLastPlayed on UserDatas (key, userId, lastPlayedDate);
```

With this index, query times dropped to about 600ms, a pretty drastic improvement. I'm not familiar enough with C# or this project to contribute a fix for this, but this seems like a pretty easy win.

### Reproduction Steps

1. With a moderately sized library, load the home page of the web interface
2. Note response time is very long for Next Up HTTP request.

### Jellyfin Version

10.8.13

### Environment

```markdown
- OS: Ubuntu 20.04
- Linux Kernel: 5.4.0-148-generic
- Virtualization: LXC
- Clients: Browser
- Browser: Chromium
- FFmpeg Version: 5.1.4-3-focal
- Playback Method: N/A
- Hardware Acceleration: N/A
- GPU Model: N/A
- Plugins: N/A
- Reverse Proxy: traefik
- Base URL: /
- Networking: macvlan
- Storage: local
```

### Jellyfin logs

```shell
[2024-04-08 18:04:36.432 +10:00] [WRN] Slow HTTP Response from "https://<hostname>/Shows/NextUp?Limit=24&Fields=PrimaryImageAspectRatio%2CDateCreated%2CBasicSyncInfo%2CPath%2CMediaSourceCount&UserId=<userid>&ImageTypeLimit=1&EnableImageTypes=Primary%2CBackdrop%2CBanner%2CThumb&EnableTotalRecordCount=false&DisableFirstEpisode=false&NextUpDateCutoff=2023-04-09T08%3A04%3A15.879Z&EnableRewatching=false" to "<client-ip>" in 0:00:20.5455497 with Status Code 200
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

felix920506:
before anyone wants to reopen this, this is a duplicate of #11888 , please track over there instead

PR #11955, opened 2024-06-04
NextUp query respects Limit

cptn-x:
**Changes**
Currently Nextup query does not use the limit defined in the query. It just gets ignored halfway through and defaults to "null". This means that no limit is placed and the sqlite query just pulls everything. This fix uses the query to define a limit X. This should also massively speed up nextup from pulling everything to only pulling X number of items. 

**Issues**
Resolves slow homescreen loading due to slow nextup query. - #11888 #11316


Shadowghost:
This should be PRed against 10.9 branch.

cptn-x:
tried changing the branch directly but seems to be making a mess. will close this PR and create a new one against 10.9

PR #11959, opened 2024-06-04
Do not stop validation if folder was removed

Shadowghost:
Got an error when an item was moved and even after a library validation it reappeared, even though the folder was removed. Checked the code and found this.

**Changes**
* do not stop validation if no children are returned, instead continue and cleanup all children

**Issues**


Bond-009 (approved):

1337joe (approved):

revam:
This change successfully broke the library and marked all the series as removed when it threw on a library root folder.

revam:
I'll also add that the series were still fine in the file system, and they just got removed from the database/UI. :slightly_smiling_face:

PR #11966, opened 2024-06-05
add full-width comma as _nameDelimiters

### LINKED ISSUE #11965: [Feature Request] add full-width comma as _nameDelimiters
Currently, jellyfin splits artist name by these characters: ``` '/', '|', ';', '\\' ```

I would like to add ``` ，``` as _nameDelimiters. It's used in some Chinese and Japanese CDs and seldom appear in an artist name.

yuygfgg:
**Changes**

**Issues**
Fix #11965

### DIFF
--- a/MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
+++ b/MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
@@ -29,7 +29,7 @@ namespace MediaBrowser.MediaEncoding.Probing
 
         private const string ArtistReplaceValue = " | ";
 
-        private readonly char[] _nameDelimiters = { '/', '|', ';', '\\' };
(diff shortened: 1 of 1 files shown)


Shadowghost (approved):

gnattu:
This currently does nothing though, because this only affects ffprobe and we are defaulting to taglib and will use ATL in 10.10. I think the delimiters need a rework to be user-configurable and users can add whitelist for artist name to not be split.

yuygfgg:
> This currently does nothing though, because this only affects ffprobe and we are defaulting to taglib and will use ATL in 10.10. I think the delimiters need a rework to be user-configurable and users can add whitelist for artist name to not be split.

yeah I found that. So currently we can't change the artist splitting settings?

gnattu:
> yeah I found that. So currently we can't change the artist splitting settings?

Not for 10.9. Ideally this should be possible for 10.10

Issue #11957, opened 2024-06-04
[Issue]: database table locked when refreshing library


bleriaf:
### Please describe your bug

i'm trying to refresh my library however it keeps failing, after checking the console there is a sqlite error 6 error saying database table is locked

### Reproduction Steps

in unraid, open jellyfin and refresh your library

### Jellyfin Version

10.9.0

### Environment

```markdown
- OS: Unraid
- Linux Kernel:
- Virtualization: Docker
- Clients: 
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

```shell
[08:30:40] [ERR] [200] Emby.Server.Implementations.ScheduledTasks.TaskManager: Error executing Scheduled Task
Microsoft.Data.Sqlite.SqliteException (0x80004005): SQLite Error 6: 'database table is locked'.
   at Microsoft.Data.Sqlite.SqliteDataReader.NextResult()
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteReader(CommandBehavior behavior)
   at Microsoft.Data.Sqlite.SqliteConnectionExtensions.ExecuteNonQuery(SqliteConnection connection, String commandText, SqliteParameter[] parameters)
   at Microsoft.Data.Sqlite.SqliteTransaction..ctor(SqliteConnection connection, IsolationLevel isolationLevel, Boolean deferred)
   at Emby.Server.Implementations.Data.SqliteItemRepository.SaveImages(BaseItem item)
   at Emby.Server.Implementations.Library.LibraryManager.UpdateImagesAsync(BaseItem item, Boolean forceUpdate)
   at Emby.Server.Implementations.Library.LibraryManager.RunMetadataSavers(BaseItem item, ItemUpdateType updateReason)
   at Emby.Server.Implementations.Library.LibraryManager.UpdateItemsAsync(IReadOnlyList`1 items, BaseItem parent, ItemUpdateType updateReason, CancellationToken cancellationToken)
   at Emby.Server.Implementations.Library.LibraryManager.ConvertImageToLocal(BaseItem item, ItemImageInfo image, Int32 imageIndex, Boolean removeOnFailure)
   at Emby.Server.Implementations.Library.LibraryManager.UpdateImagesAsync(BaseItem item, Boolean forceUpdate)
[...]

PR #11969, opened 2024-06-05
Create readonly DB connections when possible

Bond-009:
I'm just hoping SQLite does some internal magic
#11589 #11957


Shadowghost (approved):

gnattu:
Question: is this option really used?

crobibero (approved):

PR #11979, opened 2024-06-06
Fix warning CA1819 in TaskManager

kesslern:
**Changes**
Fix [CA1819: Properties should not return arrays](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1819) by changing the type of `ScheduledTasks` to an `ImmutableList`. Other improvements related to newer syntax and structured logging were made while fixing this class up.

**Issues**
Related to #2149

### DIFF
--- a/Emby.Server.Implementations/ScheduledTasks/TaskManager.cs
+++ b/Emby.Server.Implementations/ScheduledTasks/TaskManager.cs
@@ -3,6 +3,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Linq;
 using System.Threading.Tasks;
 using Jellyfin.Data.Events;
@@ -20,8 +21,7 @@ namespace Emby.Server.Implementations.ScheduledTasks
         /// <summary>
         /// The _task queue.
         /// </summary>
-        private readonly ConcurrentQueue<Tuple<Type, TaskOptions>> _taskQueue =
-            new ConcurrentQueue<Tuple<Type, TaskOptions>>();
+        private readonly ConcurrentQueue<Tuple<Type, TaskOptions>> _taskQueue = new();
 
         private readonly IApplicationPaths _applicationPaths;
         private readonly ILogger<TaskManager> _logger;
@@ -38,7 +38,7 @@ namespace Emby.Server.Implementations.ScheduledTasks
             _applicationPaths = applicationPaths;
             _logger = logger;
 
-            ScheduledTasks = Array.Empty<IScheduledTaskWorker>();
+            ScheduledTasks = [];
         }
 
         public event EventHandler<GenericEventArgs<IScheduledTaskWorker>>? TaskExecuting;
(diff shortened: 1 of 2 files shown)


Emby.Server.Implementations/ScheduledTasks/TaskManager.cs
```diff
@@ -38,7 +38,7 @@ public TaskManager(
             _applicationPaths = applicationPaths;
             _logger = logger;
 
-            ScheduledTasks = Array.Empty<IScheduledTaskWorker>();
+            ScheduledTasks = [];
```
JPVenson, line 41: this will create a new instance each time, please use the Array.Empty which reuses the instances instead.

Emby.Server.Implementations/ScheduledTasks/TaskManager.cs
```diff
@@ -49,7 +49,7 @@ public TaskManager(
         /// Gets the list of Scheduled Tasks.
         /// </summary>
         /// <value>The scheduled tasks.</value>
-        public IScheduledTaskWorker[] ScheduledTasks { get; private set; }
+        public ImmutableList<IScheduledTaskWorker> ScheduledTasks { get; private set; }
```
JPVenson, line 52: We prefer `IReadOnlyList` as its more dynamic with its implementation
```suggestion
        public IReadOnlyList<IScheduledTaskWorker> ScheduledTasks { get; private set; }
```

MediaBrowser.Model/Tasks/ITaskManager.cs
```diff
@@ -17,7 +18,7 @@ public interface ITaskManager : IDisposable
         /// Gets the list of Scheduled Tasks.
         /// </summary>
         /// <value>The scheduled tasks.</value>
-        IScheduledTaskWorker[] ScheduledTasks { get; }
+        ImmutableList<IScheduledTaskWorker> ScheduledTasks { get; }
```
JPVenson, line 21:
```suggestion
        IReadOnlyList<IScheduledTaskWorker> ScheduledTasks { get; }
```

JPVenson (changes requested):

Emby.Server.Implementations/ScheduledTasks/TaskManager.cs
Bond-009: This is not true, `[]` gets optimized to the same assembly as Array.Empty
https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIGYACMhgYQYG8aHunHjyADAG0AugwCyACgCUHLjwXEA7AwB2MAO5NBQgSIDc87gF8jDM/W3Cx40jLnUFilQEEoUbAE8AdAFF8AA4YngA8/AIAfDKGjjymsdwWfDo2dPacCU7KDKIxCvHGQA===

JPVenson: In that case, disregard

PR #12011, opened 2024-06-07
Add UploadHelper to validate true MimeTypes

Shadowghost:
Our current MIME type checks solely rely on file extension. Use https://github.com/MediatedCommunications/Mime-Detective for file-based MIME recogition

**Changes**
* Implement file based MIME type recognition

**Issues**

### DIFF
--- a/Emby.Server.Implementations/ApplicationHost.cs
+++ b/Emby.Server.Implementations/ApplicationHost.cs
@@ -630,6 +630,7 @@ namespace Emby.Server.Implementations
             serviceCollection.AddSingleton<ITranscodeManager, TranscodeManager>();
             serviceCollection.AddScoped<MediaInfoHelper>();
             serviceCollection.AddScoped<AudioHelper>();
+            serviceCollection.AddSingleton<UploadHelper>();
             serviceCollection.AddScoped<DynamicHlsHelper>();
             serviceCollection.AddScoped<IClientEventLogger, ClientEventLogger>();
             serviceCollection.AddSingleton<IDirectoryService, DirectoryService>();
--- a/Emby.Server.Implementations/ApplicationHost.cs
+++ b/Emby.Server.Implementations/ApplicationHost.cs
@@ -630,7 +630,7 @@ namespace Emby.Server.Implementations
             serviceCollection.AddSingleton<ITranscodeManager, TranscodeManager>();
             serviceCollection.AddScoped<MediaInfoHelper>();
             serviceCollection.AddScoped<AudioHelper>();
-            serviceCollection.AddScoped<UploadHelper>();
+            serviceCollection.AddSingleton<UploadHelper>();
             serviceCollection.AddScoped<DynamicHlsHelper>();
             serviceCollection.AddScoped<IClientEventLogger, ClientEventLogger>();
             serviceCollection.AddSingleton<IDirectoryService, DirectoryService>();
--- a/Emby.Server.Implementations/ApplicationHost.cs
+++ b/Emby.Server.Implementations/ApplicationHost.cs
@@ -630,6 +630,7 @@ namespace Emby.Server.Implementations
             serviceCollection.AddSingleton<ITranscodeManager, TranscodeManager>();
             serviceCollection.AddScoped<MediaInfoHelper>();
             serviceCollection.AddScoped<AudioHelper>();
+            serviceCollection.AddScoped<UploadHelper>();
             serviceCollection.AddScoped<DynamicHlsHelper>();
             serviceCollection.AddScoped<IClientEventLogger, ClientEventLogger>();
             serviceCollection.AddSingleton<IDirectoryService, DirectoryService>();
--- a/Jellyfin.Api/Helpers/UploadHelper.cs
+++ b/Jellyfin.Api/Helpers/UploadHelper.cs
@@ -1,0 +1,124 @@
+using System;
+using System.Collections.Frozen;
+using System.IO;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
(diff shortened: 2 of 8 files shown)


Emby.Server.Implementations/ApplicationHost.cs
```diff
@@ -555,6 +555,7 @@ protected virtual void RegisterServices(IServiceCollection serviceCollection)
             serviceCollection.AddSingleton<ITranscodeManager, TranscodeManager>();
             serviceCollection.AddScoped<MediaInfoHelper>();
             serviceCollection.AddScoped<AudioHelper>();
+            serviceCollection.AddScoped<UploadHelper>();
```
crobibero, line 558: Should probably be registered as a Singleton

Jellyfin.Api/Helpers/UploadHelper.cs
```diff
@@ -0,0 +1,131 @@
...
+namespace Jellyfin.Api.Helpers;
+
+/// <summary>
+/// Utitlity class providing upload helper functions.
+/// </summary>
+public class UploadHelper
+{
+    private readonly ImmutableArray<Definition> _videoDefinitions;
+    private readonly ImmutableArray<Definition> _audioDefinitions;
+    private readonly ImmutableArray<Definition> _imageDefinitions;
```
crobibero, line 25: Use FrozenSet instead of ImmutableArray

Emby.Server.Implementations/ApplicationHost.cs
```diff
@@ -575,6 +575,7 @@ protected virtual void RegisterServices(IServiceCollection serviceCollection)
...
```
crobibero: The UploadHelper is expensive to create with all of the FrozenSets so this should be added as Singleton

Jellyfin.Api/Helpers/UploadHelper.cs
```diff
@@ -0,0 +1,124 @@
...
+namespace Jellyfin.Api.Helpers;
+
+/// <summary>
+/// Utitlity class providing upload helper functions.
+/// </summary>
+public class UploadHelper
+{
+    private readonly FrozenSet<Definition> _videoDefinitions;
+    private readonly FrozenSet<Definition> _audioDefinitions;
+    private readonly FrozenSet<Definition> _imageDefinitions;
```
crobibero, line 54: There isn't a benefit in using FrozenSet here since we aren't using it for lookup but just converting it to a list
```suggestion
    private readonly List<Definition> _videoDefinitions;
    private readonly List<Definition> _audioDefinitions;
    private readonly List<Definition> _imageDefinitions;
```

Jellyfin.Api/Helpers/UploadHelper.cs
```diff
@@ -0,0 +1,124 @@
...
+    /// <param name="namingOptions">The naming options.</param>
+    public UploadHelper(
+        NamingOptions namingOptions)
+    {
+        var allDefinitions = new ExhaustiveBuilder()
+            {
+                UsageType = UsageType.PersonalNonCommercial
+            }.Build();
+
+        var extensions = namingOptions.AudioFileExtensions.Select(x => x.Replace(".", string.Empty, StringComparison.OrdinalIgnoreCase)).ToArray();
```
crobibero, line 69:
```suggestion
        var extensions = namingOptions.AudioFileExtensions.Select(x => x.Replace(".", string.Empty, StringComparison.Ordinal)).ToArray();
```

Jellyfin.Api/Helpers/UploadHelper.cs
```diff
@@ -0,0 +1,124 @@
...
+        _audioDefinitions = allDefinitions
+            .ScopeExtensions(extensions)
+            .TrimMeta()
+            .TrimDescription()
+            .ToFrozenSet();
+
+        extensions = namingOptions.VideoFileExtensions.Select(x => x.Replace(".", string.Empty, StringComparison.OrdinalIgnoreCase)).ToArray();
```
crobibero, line 76:
```suggestion
        extensions = namingOptions.VideoFileExtensions.Select(x => x.Replace(".", string.Empty, StringComparison.Ordinal)).ToArray();
```

Jellyfin.Api/Helpers/UploadHelper.cs
```diff
@@ -0,0 +1,124 @@
...
+    /// <returns>MIME type information.</returns>
+    public DefinitionMatch? GetMimeInfo(Stream? stream, string? contentType)
+    {
+        ArgumentNullException.ThrowIfNull(stream);
+        ArgumentNullException.ThrowIfNull(contentType);
+
+        var definitions = GetDefinitionsForType(contentType.Split('/')[0]);
+        var inspector = new ContentInspectorBuilder()
+        {
+            Definitions = definitions.ToList(),
```
crobibero, line 121: How does the inspector react when an empty list is set?
Can we just return null if it isn't a known definition type?

Jellyfin.Api/Helpers/UploadHelper.cs
```diff
@@ -0,0 +1,124 @@
...
+        return null;
+    }
+
+    /// <summary>
+    /// Writes the stream content to a file.
+    /// </summary>
+    /// <param name="stream">The data stream.</param>
+    /// <param name="filePath">The file path to write the data to.</param>
+    /// <param name="cancellationToken">The cancellation token.</param>
+    public static async void WriteStreamToFile(Stream stream, string filePath, CancellationToken cancellationToken)
```
crobibero, line 190: `async void` is an anti-pattern and should be avoided unless it is absolutely required

crobibero (changes requested):

PR #12020, opened 2024-06-07
Improve album artwork filename hashing

### LINKED ISSUE #11942: [Issue]: Incorrect album art for albums with the same name
### Please describe your bug

Running the latest stable version (`10.9.4`) of Jellyfin in Docker and I currently use MusicBrainz to name my albums, with the rough schema of [album artist]/[album](original release year)/[song].

Examples:
`//Days of the New/Days of the New (1997)/Track 01 - Shelf in the Room.mp3`
`//Days of the New/Days of the New (1999)/Track 01 - Flight Response`
`//Days of the New/Days of the New (2001)/Track 01 - Hang on to This`

Each album has embedded album art respectively.

The problem is that when there are multiple albums with the same title as is the case with all of the Days of the New albums, Jellyfin seemingly chooses at random a single album cover and uses it for all albums with the same name when it should be pulling the correct cover from each album.

image

### Reproduction Steps

1. Deploy a fresh install of Jellyfin
2. Configure a music library that contains an artist that has multiple albums with the same name such as Days of the New or Weezer
3. Review cover art.

### Jellyfin Version

10.9.0

### if other:

_No response_

### Environment

```markdown
- OS: Ubuntu 22.04
- Linux Kernel:
- Virtualization: Docker
- Clients: iOS, Manet, Swiftfin
- Browser: Safari
- FFmpeg Version:
- Playback Method: Direct
- Hardware Acceleration: none
- GPU Model: none
- Plugins: Fanart, Theme Songs
- Reverse Proxy: Caddy
- Base URL: none
- Networking: Bridge
- Storage: local
```

### Jellyfin logs

```shell
none
```

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

PleaseStopAsking:
**Changes**

This PR adds an additional field (`.PremiereDate`) to the hashing that determines what an albums artwork should be.
Also added a check to `if` to ensure the `.PremiereDate` exists alongside the exists checks for `.Album` and `.AlbumArtists[0]`.

**Issues**

Fixes #11942 
<!-- Tag any issues that this PR solves here.
ex. Fixes

### DIFF
--- a/MediaBrowser.Providers/MediaInfo/AudioImageProvider.cs
+++ b/MediaBrowser.Providers/MediaInfo/AudioImageProvider.cs
@@ -116,9 +116,10 @@ namespace MediaBrowser.Providers.MediaInfo
             {
                 if (item.AlbumArtists.Count > 0
                     && !string.IsNullOrWhiteSpace(item.Album)
-                    && !string.IsNullOrWhiteSpace(item.AlbumArtists[0]))
+                    && !string.IsNullOrWhiteSpace(item.AlbumArtists[0])
+                    && item.PremiereDate.HasValue)
                 {
-                    filename = (item.Album + "-" + item.AlbumArtists[0]).GetMD5().ToString("N", CultureInfo.InvariantCulture);
+                    filename = (item.Album + "-" + item.PremiereDate + "-" + item.AlbumArtists[0]).GetMD5().ToString("N", CultureInfo.InvariantCulture);
                 }
                 else
                 {


Shadowghost:
Might be worth PRing against 10.9.z

gnattu:
This is a behavior change and is not always correct. Complication albums will have songs with different PremiereDate because that song might be released earlier as a single, and the PremiereDate of such song will be different from other songs.

PleaseStopAsking:
@gnattu In my review of dozens of compilation albums, the premiere data is always the same across songs in the same album as its set to the album release date and not on a per song basis.

This is unless something else is defining the `.PremiereDate` that I am missing.

gnattu:
We are reading per song specific tag here. There is no guarantee that all songs in the same album will get tagged with the same date. All of your collections happen to have them tagged with same date does not mean it is also the case for others. For example, albums bought from iTunes Store will have songs tagged with a different year.

PleaseStopAsking:
The fact that iTunes (possibly others) is tagging individual songs in an album with their release date and not the album's release date, really speaks more toward using a service such as MusicBrainz before bringing music into a library. With that said, I can see the performance concern when this is the case.

gnattu:
We can make it a per-library option, like "Allow Albums with same name to have different covers", and add a note to the user that enabling this would correctly search for the album cover if enabled, but will slow down library scan. With that option we can use the `AlbumEntity` of Audio to actually find the real album and use its guid instead. So users will have a choice between performance and correctness, which would work better than this.

PleaseStopAsking:
That sounds like a viable option. I will close this PR and look into your suggestion as a new PR but it's very likely outside my current level of experience with Jellyfin.

Issue #10637, opened 2023-11-28
[Issue]: Embedded Thumnail is not displayed (Regression - worked in 10.8.10)


MattBlackOnly:
### Please describe your bug

Embedded Thumbnails were previously displayed in Version 10.8.10, however (since Version 10.8.11) they are no longer displayed.
Library is set to "Mixed Movies and Shows" and it's using Embedded Image Extractor.

https://forum.jellyfin.org/t-embedded-thumbnails-are-not-being-displayed

### Jellyfin Version

Other

### if other:

10.8.12

### Please attach any screenshots here

![NoThumbnail](https://github.com/jellyfin/jellyfin/assets/152306888/f9393069-292c-4b83-9d47-f1412d48e5ed)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

MattBlackOnly:
The logs just say:
`MediaBrowser.Providers.Movies.MovieMetadataService: Running "EmbeddedImageProvider" for "video.mp4"`
and there is no error message.

The issue appears to be associated with the changes introduced in PR #10173 but I don't see where exactly...

TheMelmacian:
With #10173 the image types bmp, gif, png and webp are no longer checked if the media type is not audio.
Only mjpeg is checked for videos.

So if a video can have other formats embedded than mjpeg the code of [ProbeResultNormalizer.cs](https://github.com/jellyfin/jellyfin/pull/10173/files#diff-2bad799e56914e76976398c1a62474393b3950cd6ad6cd5d9acda67005295032R764) should probably look something like that:
```csharp
                if (isAudio
                    || string.Equals(stream.Codec, "bmp", StringComparison.OrdinalIgnoreCase)
                    || string.Equals(stream.Codec, "gif", StringComparison.OrdinalIgnoreCase)
                    || string.Equals(stream.Codec, "png", StringComparison.OrdinalIgnoreCase)
                    || string.Equals(stream.Codec, "webp", StringComparison.OrdinalIgnoreCase)))
                {
                    stream.Type = MediaStreamType.EmbeddedImage;
                }
                // mjpeg can be both: video or embedded image
                else if (string.Equals(stream.Codec, "mjpeg", StringComparison.OrdinalIgnoreCase))
                {
                    // if the media type is audio we can be sure that it is an embedded image
                    if (isAudio)
                    {
                        stream.Type = MediaStreamType.EmbeddedImage;
                    }
                    // How to differentiate between video and embedded image?
[...]

PR #12031, opened 2024-06-08
Fix video embedded image detection

nyanmisaka:
**Changes**
- Fix video embedded image detection.

**Issues**
Fixes https://github.com/jellyfin/jellyfin/commit/debbfaa502cae91212f2386d237616ad2a671a88. Embedded images also exist in video.

Fixes [#10637](https://github.com/jellyfin/jellyfin/issues/10637)


gnattu:
One thing to note that mjpeg sequences passed in format `jpeg_pipe` will also have no codec tags and can be converted to a normal video stream. Our current detection will flag those as embedded images as well. But this use case is super edgy as most of the mjpeg videos will be contained in avi or mov containers and the `jpeg_pipe` is super rare.

gnattu (approved):
I'd say let's ignore the `jpeg_pipe` for now. That format is not common and is badly supported everywhere as VLC does not even play it (mpv does play).

Bond-009 (approved):

crobibero (approved):

PR #12039, opened 2024-06-08
Fix local episode image thumb recognition

Shadowghost:
Function already checks for `-thumb`. Regression from #11934

**Changes**
* Remove `-thumb` from initial input to episode thumb check
* Simplify `metadata` subdir check

**Issues**


MediaBrowser.LocalMetadata/Images/EpisodeLocalImageProvider.cs
```diff
@@ -69,6 +66,7 @@ private List<LocalImageInfo> GetImageFilesFromFolder(ReadOnlySpan<char> filename
                 if (BaseItem.SupportedImageExtensions.Contains(i.Extension.AsSpan(), StringComparison.OrdinalIgnoreCase))
                 {
                     var currentNameWithoutExtension = Path.GetFileNameWithoutExtension(i.FullName.AsSpan());
+                    var thumbName = string.Concat(filenameWithoutExtension, "-thumb");
```
cvium, line 69: It should be outside the loop

cvium (approved):

crobibero (approved):

Bond-009 (approved):

PR #12043, opened 2024-06-09
Fix the Australian PG rating

### LINKED ISSUE #11650: [Issue]: Australian Classification Ratings not correct
### Please describe your bug

I believe this file needs editing: 

https://github.com/jellyfin/jellyfin/blob/fd250e4fe1c1a35e8e40c008a7d25675729e3dac/Emby.Server.Implementations/Localization/Ratings/au.csv

The Australian Classification has 6 ratings:

https://www.classification.gov.au/classification-ratings/what-do-ratings-mean

- G: no age restriction
- PG: PG-rated content is not recommended for viewing by people under the age of 15 without guidance from parents, teachers or guardians.
- M: M-rated films and computer games are not recommended for children under the age of 15. However, children under the age of 15 may legally access this material as the classification rating is an advisory category.
- MA 15+: legally restricted to people over the age of 15
- R 18+: legally restricted to people over the age of 18
- X 18+: legally restricted to people over the age of 18

in Jellyfin 1.0.9.0+ the list currently looks like the attached screenshot.

PG and M should be before MA 15+, probably at age 14. I'm not sure if the file needs all the other age ratings that do not appear in the Australian classification system, I don't think they appeared in 1.0.8.0 and below.

### Reproduction Steps

Jellfin > Dashboard > Users > Parental Control > Maximum Allowed Parental Rating dropdown

### Jellyfin Version

10.9.0

### if other:

_No response_

### Environment

```markdown
any
```

### Jellyfin logs

```shell
none needed
```

### FFmpeg logs

_No response_

_No response_

image

### Code of Conduct

oddstr13:
As per https://www.classification.gov.au/classification-ratings/what-are-ratings

**Changes**
Move PG to the top of the list for 15+

**Issues**
Fixes #11650

Well, sort of. I don't think it is possible to differentiate between them, as we'd be comparing the integer values, not the position in the list?

I also think it's not really appropriate for us to set it to a lower age either, as that'd be factually wrong.


gnattu:
I'm fine with this as PG with 16 is incorrect anyway. However, this still leaves a problem for us because PG and M now behave the same as MA, but they actually have different meanings in Australian law.

For 10.9, I guess this is the best we can do. For 10.10, maybe we can extend the rating from int to float and use decimals to indicate different levels at the same age restriction. For example, PG in Australia could be 15.0, M could be 15.1, and MA could be 15.2. This way, we don't need to change all of the comparison logic, and Australian users can still set the children with a PG limit to block access to M and MA content.

Shadowghost (approved):
That's all we can do for 10.9

cvium (approved):

Bond-009 (approved):

Rygle:
Please see my comments here
https://github.com/jellyfin/jellyfin/issues/11650#issuecomment-2335010190

It should be that setting a profile to parental rating AU-PG will show U, G, AU-G, 7, and AU-PG, but not above.

AU-PG should incorporate only ratings that are clearly below it such as U, G, AU-G, 7, and AU-PG, but not ratings where there might be any ambiguity including PG, 13, 14, 15, or anything else that is open to interpretation. The onus is then on the parent to set the ratings in the metadata before something shows up. It is better to err toward caution rather than have a parental rating of AU-PG show AU-MA15+.

Rygle:
This whole thing seems have gone wrong in an effort to make things more granular.

Perhaps a better approach would be to make things more granular in the options available in the settings rather than in the interpretation of the settings that the user has no control over (apart from re-compiling Jellyfin). So all profiles have a check list of all possible ratings, and the parent/administrator has to explicitly tick every rating they want to be visible. Or to speed this up, it could have umbrella categories with sub-categories that get selected if the umbrella category is selected, but only if the umbrella category is selected. Alternately, there could be a tick box for "more granular ratings selections" that does something like the following (if un-ticked it can remain as it was)...

So, in the Jellyfin settings, I will see something like;

NR [ ]
-- NR [ ]
-- AU-NC [ ]
etc for other countries

U [ ]
-- U [ ]
-- AU-U [ ]
etc for other countries

G [ ]
-- MPAA-G [ ]
-- AU-G [ ]
-- 7 [ ]
etc for other countries

PG [ ]
-- 13
-- PG-13 [ ]
-- AU-PG [ ]
-- 14 [ ]
etc for other countries

M [ ]
-- MPAA-M [ ]
-- 15 [ ]
-- AU-M [ ]
-- AU-MA15 [ ]
etc for other countries

etc, etc.

That way a parent/administrator can tick either the larger category that will select all under it, or just tick the ratings they want.

Additionally, if there are ratings/classifications that are not available in Jellyfin that match the metadata for your region/location, Jellyfin should allow the parent/administrator to type in their own custom ratings that match the metadata that their scraper is getting for their location, and that these get incorporated into the longer list of tick-boxes for ratings once entered.

PR #12046, opened 2024-06-09
Fix mpeg-ts detection

gnattu:
When the container name is `mpeg`, it means it is MPEG-PS, while the TS container should have the explicit name `mpegts`.

**Changes**

**Issues**

Fixes https://forum.jellyfin.org/t-transcoding-fails-for-mpeg1-mpeg2-videos-with-mp2-audio?pid=26210#pid26210

### DIFF
--- a/MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
+++ b/MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
@@ -280,8 +280,8 @@ namespace MediaBrowser.MediaEncoding.Probing
                     splitFormat[i] = "mpeg";
                 }
 
-                // Handle MPEG-2 container
-                else if (string.Equals(splitFormat[i], "mpeg", StringComparison.OrdinalIgnoreCase))
+                // Handle MPEG-TS container
+                else if (string.Equals(splitFormat[i], "mpegts", StringComparison.OrdinalIgnoreCase))
                 {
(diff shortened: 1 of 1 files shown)


Shadowghost (approved):

Bond-009 (approved):

crobibero (approved):

MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
```diff
@@ -280,8 +280,8 @@ private string NormalizeFormat(string format, IReadOnlyList<MediaStream> mediaSt
                     splitFormat[i] = "mpeg";
                 }
 
-                // Handle MPEG-2 container
-                else if (string.Equals(splitFormat[i], "mpeg", StringComparison.OrdinalIgnoreCase))
+                // Handle MPEG-TS container
+                else if (string.Equals(splitFormat[i], "mpegts", StringComparison.OrdinalIgnoreCase))
```
cvium, line 284: The description and PR title mentions `mpeg-ts`. I just want to confirm that is `mpegts`?

gnattu: Yes. They are the same thing but the ffmpeg short name is mpegts.

cvium (approved):

jumoog (approved):

PR #12056, opened 2024-06-09
Allow streaming of raw PGS subtitles without transcoding

Arcus92:
**Changes**
This PR adds support for streaming raw PGS subtitles (graphical subtitles) if requested by the client. PGS subtitles can be loaded from external `.sub` / `.sup` files or extracted from an `.mkv` container.
This removed the need for video transcoding just for the PGS subtitle.

**Related Pull Request (web):** https://github.com/jellyfin/jellyfin-web/pull/5688
**Related Pull Request (vue):** https://github.com/jellyfin/jellyfin-vue/pull/2404

**Discussion points:**
- Extracting subtitles from large `.mkv` can take multiple minutes for ffmpeg to process the whole file. This is only be done once per media file. The subtitles are cached afterwards.
- PGS subtitles are large (multiple megabytes) and may take a while to load.

### DIFF
--- a/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
+++ b/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
@@ -215,6 +215,20 @@ namespace MediaBrowser.MediaEncoding.Subtitles
             var currentFormat = (Path.GetExtension(subtitleStream.Path) ?? subtitleStream.Codec)
                 .TrimStart('.');
 
+            // Handle PGS subtitles as raw streams for the client to render
+            if (string.Equals(currentFormat, "sub", StringComparison.OrdinalIgnoreCase)
+               || string.Equals(currentFormat, "sup", StringComparison.OrdinalIgnoreCase)
+               || string.Equals(currentFormat, "pgs", StringComparison.OrdinalIgnoreCase))
+            {
+                return new SubtitleInfo()
+                {
+                    Path = subtitleStream.Path,
+                    Protocol = _mediaSourceManager.GetPathProtocol(subtitleStream.Path),
+                    Format = "pgssub",
+                    IsExternal = true
+                };
+            }
+
             // Fallback to ffmpeg conversion
             if (!_subtitleParser.SupportsFileExtension(currentFormat))
             {
--- a/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
+++ b/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
@@ -198,10 +198,11 @@ namespace MediaBrowser.MediaEncoding.Subtitles
         {
             if (!subtitleStream.IsExternal || subtitleStream.Path.EndsWith(".mks", StringComparison.OrdinalIgnoreCase))
             {
-                await ExtractAllTextSubtitles(mediaSource, cancellationToken).ConfigureAwait(false);
+                await ExtractAllExtractableSubtitles(mediaSource, cancellationToken).ConfigureAwait(false);
 
-                var outputFormat = GetTextSubtitleFormat(subtitleStream);
-                var outputPath = GetSubtitleCachePath(mediaSource, subtitleStream.Index, "." + outputFormat);
+                var outputFileExtension = GetExtractableSubtitleFileExtension(subtitleStream);
+                var outputFormat = GetExtractableSubtitleFormat(subtitleStream);
+                var outputPath = GetSubtitleCachePath(mediaSource, subtitleStream.Index, "." + outputFileExtension);
 
                 return new SubtitleInfo()
                 {
@@ -216,9 +217,7 @@ namespace MediaBrowser.MediaEncoding.Subtitles
                 .TrimStart('.');
 
             // Handle PGS subtitles as raw streams for the client to render
-            if (string.Equals(currentFormat, "sub", StringComparison.OrdinalIgnoreCase)
-               || string.Equals(currentFormat, "sup", StringComparison.OrdinalIgnoreCase)
-               || string.Equals(currentFormat, "pgs", StringComparison.OrdinalIgnoreCase))
+            if (MediaStream.IsPgsFormat(currentFormat))
             {
                 return new SubtitleInfo()
                 {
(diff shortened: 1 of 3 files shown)


MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
```diff
@@ -441,33 +455,47 @@ private string GetTextSubtitleFormat(MediaStream subtitleStream)
             }
         }
 
+        private string GetExportableSubtitleFileExtension(MediaStream subtitleStream)
+        {
+            // Using .pgssub as file extension is not allowed by ffmpeg. The file extension for pgs subtitles is .sup (or .sub).
```
nyanmisaka, line 460: ~(or .sub)~

MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
```diff
@@ -441,33 +455,47 @@ private string GetTextSubtitleFormat(MediaStream subtitleStream)
...
 
         /// <summary>
-        /// Extracts all text subtitles.
+        /// Extracts all extractable subtitles (text and pgs).
         /// </summary>
         /// <param name="mediaSource">The mediaSource.</param>
         /// <param name="cancellationToken">The cancellation token.</param>
         /// <returns>Task.</returns>
-        private async Task ExtractAllTextSubtitles(MediaSourceInfo mediaSource, CancellationToken cancellationToken)
+        private async Task ExtractAllExtractableSubtitles(MediaSourceInfo mediaSource, CancellationToken cancellationToken)
```
nyanmisaka, line 486: Also rename the `ExtractAllTextSubtitlesInternal` to match `ExtractAllExtractableSubtitles`.

MediaBrowser.Model/Entities/MediaStream.cs
```diff
@@ -585,6 +585,31 @@ public bool IsTextSubtitleStream
...
+                    return false;
+                }
+
+                return IsPgsFormat(Codec);
+            }
+        }
+
+        /// <summary>
+        /// Gets a value indicating whether this is a subtitle steam that is extractable by ffmpeg.
+        /// All text-based pgs subtitles can be extracted.
```
nyanmisaka, line 608: text-based pgs? (Presentation Graphic Stream)

MediaBrowser.Model/Entities/MediaStream.cs
```diff
@@ -666,6 +691,15 @@ public static bool IsTextFormat(string format)
                    && !string.Equals(codec, "dvb_subtitle", StringComparison.OrdinalIgnoreCase);
         }
 
+        public static bool IsPgsFormat(string format)
+        {
+            string codec = format ?? string.Empty;
+
+            return codec.Contains("pgs", StringComparison.OrdinalIgnoreCase)
+                   || string.Equals(codec, "sub", StringComparison.OrdinalIgnoreCase)
+                   || string.Equals(codec, "sup", StringComparison.OrdinalIgnoreCase);
```
nyanmisaka, line 700: I think the file extension used by PGSSUB is [`.SUP`](https://github.com/FFmpeg/FFmpeg/blob/8d6014dbc6fb7c324a114030761901220ebc3540/libavformat/supdec.c#L104). While the [`.SUB/.IDX`](https://github.com/FFmpeg/FFmpeg/blob/8d6014dbc6fb7c324a114030761901220ebc3540/libavformat/mpeg.c#L1041) pairs is used by VOBSUB/DVDSUB.

ffmpeg can infer the codec from the file content, even if the file extension is wrong. For example:
```
Input #0, sup, from '1.sub':
  Duration: N/A, start: 0.000000, bitrate: N/A
  Stream #0:0: Subtitle: hdmv_pgs_subtitle, 1920x1080
```

```
Input #0, sup, from '1.file':
  Duration: N/A, start: 0.000000, bitrate: N/A
  Stream #0:0: Subtitle: hdmv_pgs_subtitle, 1920x1080
```

MediaBrowser.Model/Entities/MediaStream.cs
Arcus92: Right ... I missed an "and" there!

nyanmisaka (approved):

Shadowghost (approved):

gnattu (approved):
I'm not a big fan of this approach because:

- PGS compression only uses RLE compression which could results in extremely huge file for a multi-hour movie like >100MB, and to load this much data is not ideal for internet streaming
- Colorspace conversion and decompression in browser with plain javascript is slow, and probably won't work with low-end devices and Smart TVs.

But we can try this until we have better approaches (like server-side conversion + streaming)

nyanmisaka:
> * PGS compression only uses RLE compression which could results in extremely huge file for a multi-hour movie like >100MB, and to load this much data is not ideal for internet streaming
> * Colorspace conversion and decompression in browser with plain javascript is slow, and probably won't work with low-end devices and Smart TVs.

It would be nice if there would be an experimental option here that could be disabled by default until it is stable enough. I mentioned the feasibility of using a compression algorithm. Preferably through wasm. https://forum.jellyfin.org/t-native-pgs-support

Arcus92:
> It would be nice if there would be an experimental option in the web client that could be disabled by default until it is stable enough. I mentioned the feasibility of using a compression algorithm. Preferably through wasm. https://forum.jellyfin.org/t-native-pgs-support

You'll probably right. I'll looked into wasm a bit and eventually I'll port the rendering code to C. Currently I'm using the 'burn-in' option in the client to enable the feature. However, it is set to auto by default which makes this opt-out and not opt-in. I'll add an 'Enable experimental PGSSUB rendering' option or something for now.

crobibero (approved):

MediaBrowser.Model/Entities/MediaStream.cs
```diff
@@ -585,6 +585,31 @@ public bool IsTextSubtitleStream
...
+                return IsPgsFormat(Codec);
+            }
+        }
+
+        /// <summary>
+        /// Gets a value indicating whether this is a subtitle steam that is extractable by ffmpeg.
+        /// All text-based and pgs subtitles can be extracted.
+        /// </summary>
+        /// <value><c>true</c> if this is a extractable subtitle steam otherwise, <c>false</c>.</value>
+        public bool IsExtractableSubtitleStream => IsTextSubtitleStream || IsPgsSubtitleStream;
```
nielsvanvelzen, line 611: I'm a bit late to the party but hopefully the celebration isn't over yet. Both `IsPgsSubtitleStream` and `IsExtractableSubtitleStream` are now part of the HTTP API. Is that intentional or were those properties only meant for the server internally?

If they are meant for the API they should be marked as nullable for backwards compatibility in the generated SDK's. (can be made not-null afterwards for 10.11).

PR #12074, opened 2024-06-10
Resolve remote images using providers

revam:
Currently the resolution of remote images provided by plugins are handled differently in two places;

- If a remote image is acquired internally in Jellyfin it will ask the provider to fetch the image for us. As seen [here](https://github.com/jellyfin/jellyfin/blob/18dd6b2875b05d6b9c8249476ae509bfd20e8314/MediaBrowser.Providers/Manager/ItemImageProvider.cs#L535) and [here](https://github.com/jellyfin/jellyfin/blob/18dd6b2875b05d6b9c8249476ae509bfd20e8314/MediaBrowser.Providers/Manager/ItemImageProvider.cs#L644).

- If a remote image is asked by a client to be added to an item, it will always be downloaded using a standard HTTP client. As seen [here](https://github.com/jellyfin/jellyfin/blob/6fb6b5f1766a1f37a61b9faaa40209bab995bf30/Jellyfin.Api/Controllers/RemoteImageController.cs#L166). I'll also note that the client is sending the image url, provider name, and image type with each request. As seen [here](https://github.com/jellyfin/jellyfin-web/blob/20381bd3ec2f23907ed53d7da568cae921b05506/src/components/imageDownloader/imageDownloader.js#L140-L154).

**Changes**

This PR changes the behavior of the second scenario to better align with the first scenario, where it will ask the provider to fetch the image if the client tells us which provider to use. Of course, the provider needs to be enabled for the item, otherwise it will result in a `400 Bad Request`. I feel this change is necessary to accommodate when it's not possible for the Jellyfin server itself to directly fetch the images, but a provider can fetch it for the server, all the while the clients can also fetch the images. This can arise when the server and clients use different DNS server and live on different LANs, e.g. a docker network and normal LAN outside the docker network on a different sub-net, both with different DNS servers, where nether have the other DNS server in their DNS stack.

**Issues**
- None yet, because I never made any.

### DIFF
--- a/Jellyfin.Api/Controllers/RemoteImageController.cs
+++ b/Jellyfin.Api/Controllers/RemoteImageController.cs
@@ -180,6 +180,12 @@ public class RemoteImageController : BaseJellyfinApiController
                 return BadRequest();
             }
 
+            var images = await provider.GetImages(item, CancellationToken.None);
+            if (!images.Any(image => string.Equals(image.Url, imageUrl, StringComparison.Ordinal)))
+            {
+                return BadRequest();
+            }
+
             await _providerManager.SaveImage(item, provider, imageUrl, type, null, CancellationToken.None)
                 .ConfigureAwait(false);
         }
--- a/Jellyfin.Api/Controllers/RemoteImageController.cs
+++ b/Jellyfin.Api/Controllers/RemoteImageController.cs
@@ -161,7 +161,7 @@ public class RemoteImageController : BaseJellyfinApiController
     public async Task<ActionResult> DownloadRemoteImage(
         [FromRoute, Required] Guid itemId,
         [FromQuery, Required] ImageType type,
-        [FromQuery] string? providerName,
+        [FromQuery, Required] string providerName,
         [FromQuery] string? imageUrl)
     {
         var item = _libraryManager.GetItemById<BaseItem>(itemId, User.GetUserId());
@@ -170,30 +170,22 @@ public class RemoteImageController : BaseJellyfinApiController
             return NotFound();
         }
 
-        if (!string.IsNullOrEmpty(providerName))
+        var provider = _providerManager.GetImageProviders(item, new(new DirectoryService(_fileSystem)))
+            .OfType<IRemoteImageProvider>()
+            .FirstOrDefault(provider => string.Equals(provider.Name, providerName, StringComparison.Ordinal));
+        if (provider is null)
         {
-            var provider = _providerManager.GetImageProviders(item, new(new DirectoryService(_fileSystem)))
-                .OfType<IRemoteImageProvider>()
-                .FirstOrDefault(provider => string.Equals(provider.Name, providerName, StringComparison.Ordinal));
-            if (provider is null)
-            {
-                return BadRequest();
-            }
-
-            var images = await provider.GetImages(item, CancellationToken.None);
-            if (!images.Any(image => string.Equals(image.Url, imageUrl, StringComparison.Ordinal)))
-            {
-                return BadRequest();
-            }
-
-            await _providerManager.SaveImage(item, provider, imageUrl, type, null, CancellationToken.None)
-                .ConfigureAwait(false);
+            return BadRequest();
         }
-        else
+
+        var images = await provider.GetImages(item, CancellationToken.None);
+        if (!images.Any(image => string.Equals(image.Url, imageUrl, StringComparison.Ordinal)))
         {
-            await _providerManager.SaveImage(item, imageUrl, type, null, CancellationToken.None)
-                .ConfigureAwait(false);
+            return BadRequest();
         }
+
+        await _providerManager.SaveImage(item, provider, imageUrl, type, null, CancellationToken.None)
+            .ConfigureAwait(false);
 
         await item.UpdateToRepositoryAsync(ItemUpdateType.ImageUpdate, CancellationToken.None).ConfigureAwait(false);
         return NoContent();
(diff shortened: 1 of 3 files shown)


Jellyfin.Api/Controllers/RemoteImageController.cs
```diff
@@ -163,8 +170,24 @@ public ActionResult<IEnumerable<ImageProviderInfo>> GetRemoteImageProviders([Fro
             return NotFound();
         }
 
-        await _providerManager.SaveImage(item, imageUrl, type, null, CancellationToken.None)
-            .ConfigureAwait(false);
+        if (!string.IsNullOrEmpty(providerName))
+        {
+            var provider = _providerManager.GetImageProviders(item, new(new DirectoryService(_fileSystem)))
+                .OfType<IRemoteImageProvider>()
+                .FirstOrDefault(provider => string.Equals(provider.Name, providerName!, StringComparison.Ordinal));
```
revam, line 177: I forgot to remove the `!`…

MediaBrowser.Providers/Manager/ProviderManager.cs
```diff
@@ -182,8 +189,10 @@ await using (cacheStream.ConfigureAwait(false))
                     }
                 }
 
-                var httpClient = _httpClientFactory.CreateClient(NamedClient.Default);
-                using var response = await httpClient.GetAsync(url, cancellationToken).ConfigureAwait(false);
+                using var httpClient = _httpClientFactory.CreateClient(NamedClient.Default);
```
revam, line 192: I'll note that the compiler was complaining about an ignored `IDisposable` if i in-lined the HTTP client in the ternary, and without the `using` when it was used in the ternary, so I added it as a separate line above. Feel free to tell me a better way to do it if any.

Haxxor247 (approved):

MediaBrowser.Providers/Manager/ProviderManager.cs
cvium: HttpClients shouldn't be disposed regardless of any warnings you may see. It is effectively a no-op when the client is created using the factory

Jellyfin.Api/Controllers/RemoteImageController.cs
```diff
@@ -163,7 +170,21 @@ public ActionResult<IEnumerable<ImageProviderInfo>> GetRemoteImageProviders([Fro
             return NotFound();
         }
 
-        await _providerManager.SaveImage(item, imageUrl, type, null, CancellationToken.None)
+        var provider = _providerManager.GetImageProviders(item, new(new DirectoryService(_fileSystem)))
```
cvium, line 173: Way too much responsibility for the controller. The controller should not care about providers. The manager should manage it.

cvium:
> The URL validation was added because @thornbill pointed out that [GHSA-rgjw-4fwc-9v96](https://github.com/jellyfin/jellyfin/security/advisories/GHSA-rgjw-4fwc-9v96) existed on Matrix. But thinking about it some more, then I feel like making the validation _optional_ is not enough, because an actor just needs to omit the provider to be able to send any URL they want.
> 
> I can make the `providerName` query parameter required for all requests, thus forcing the validation of input URLs on all requests, and making sure the clients/actors can only request a small set of valid URLs, but that change will require that all clients send the provider name, and not just the Web UI client (which already does, as stated in the original post).
> 
> The potential minus of doing the extra validation the way I did that I can foresee is that it may lead to additional outgoing requests in the provider, depending on how it's wired up internally. I'll also note that it's not an issue for my provider, but it may be an issue for other providers if they don't cache data internally.

I don't think it makes sense to use the provider name in the query. Clients shouldn't really care too much about providers. Ideally, and this is just my opinion:
1. All images are proxied through the Jellyfin backend
2. Identify (and other remote download flows?) stores an ID per image
3. Remote image requests always use the ID

revam:
> > The URL validation was added because @thornbill pointed out that [GHSA-rgjw-4fwc-9v96](https://github.com/jellyfin/jellyfin/security/advisories/GHSA-rgjw-4fwc-9v96) existed on Matrix. But thinking about it some more, then I feel like making the validation _optional_ is not enough, because an actor just needs to omit the provider to be able to send any URL they want.
> > I can make the `providerName` query parameter required for all requests, thus forcing the validation of input URLs on all requests, and making sure the clients/actors can only request a small set of valid URLs, but that change will require that all clients send the provider name, and not just the Web UI client (which already does, as stated in the original post).
> > The potential minus of doing the extra validation the way I did that I can foresee is that it may lead to additional outgoing requests in the provider, depending on how it's wired up internally. I'll also note that it's not an issue for my provider, but it may be an issue for other providers if they don't cache data internally.
> 
> I don't think it makes sense to use the provider name in the query. Clients shouldn't really care too much about providers. Ideally, and this is just my opinion:
> 
>     1. All images are proxied through the Jellyfin backend
> 
>     2. Identify (and other remote download flows?) stores an ID per image
> 
>     3. Remote image requests always use the ID and based on the ID, the server knows what to download and store
> 
> 
> EDIT: This is a pretty big API breaking change, but ultimately the only reasonable (imo) fix for the security advisory. To ease clients into it, we could create a new set of endpoints.

I sadly enough do not have time to sit down and research how the entire image infrastructure is currently implemented in Jellyfin and change it on the server side and in _all_ clients. But if we're strictly keeping the scope to changes in this class then I can try to add an in-memory cache for images served through the Identify endpoint so we know where and how to forward the request to get back the image data to the client, then change the URLs returned by the Identity endpoint to point to proxied URLs for a new endpoint that just looks up the id in the cache and forwards the request to the remote image location using the provider, then modify the Download endpoint to first check if the URL sent to it is a proxied image URL, and if it is check the in-memory cache if the id exists to download the image through the provider.

If this sounds fine to you then I can either edit this PR or create a new one with these changes.

revam:
I'll close this PR for now, and instead file an issue describing the problem at hand, linking the security issue and this PR for historical context, to get more input on how to best solve it, or for some other person better qualified than me to submit a PR.

PR #12120, opened 2024-06-17
Do not override <year> if <releasedate> is set

### LINKED ISSUE #12111: [Issue]: When <premiered> or <releasedate> is present in .nfo, it overrides <year>
### Please describe your bug

When an .nfo is present with both `<premiered>` (or `<releasedate>`) and `<year>` populated, the "Year" value Jellyfin uses will be taken from `<premiered>`, rather than `<year>`. Editing the metadata using Jellyfin's editor will confirm this.

This is problematic in cases such as:
- (Assuming German metadata) an American movie from late 2021 sees its release in Germany in early 2022, making it show up as Movie (2022). This is very common.
- (Assuming American metadata) a Japanese movie from 1967 is first released on DVD in the US in 2002, making it show up as Movie (2002).

Since I assume "Year" and "Release date" are separate fields for a reason, it does not really make sense that it is not possible to import them from .nfo with separate values. Therefore, I would consider this a bug.

Also, no information would be lost taking "Year" from the `<year>` entry, since "Release date" is still available.

### Reproduction Steps

1. Have an .nfo with both `<year>` and `<premiered>` populated (but with different years).
2. The movie will show up in Jellyfin with the year from `<premiered>`. This can be confirmed by checking the "Year" field in the metadata editor.
3. Removing `<premiered>` from the .nfo correctly sets "Year" to the value of `<year>`, meaning Jellyfin is capable of interpreting `<year>`.

### Jellyfin Version

10.9.0

### if other:

10.9.6

### Environment

```markdown
- OS: Fedora 40
- Linux Kernel: 6.8.11
- Virtualization: Docker
- Clients: jellyfin-web
- Browser: Firefox
```

### Jellyfin logs

```shell
N/A. I don't believe .nfo readouts and metadata updates are logged.
```

### FFmpeg logs

_No response_

_No response_

An .nfo with the following content (note the year 2001):
```xml
  <title>Movie</title>
  <originaltitle>Motion picture</originaltitle>
  <plot>This is a movie.</plot>
  <year>2001</year>
  <premiered>2005-06-07</premiered>
```

will have this effect on the movie listing:
image

and this is the metadata editor:
image

### Code of Conduct

Shadowghost:
`<year>` is the  tag for production year. The date in `<releasedate>` can be different due to locale, so we should only use it if it is unset.

**Changes**
* Do not override already set `ProductionYear`

**Issues**
Fixes #12111

### DIFF
--- a/MediaBrowser.XbmcMetadata/Parsers/BaseNfoParser.cs
+++ b/MediaBrowser.XbmcMetadata/Parsers/BaseNfoParser.cs
@@ -518,13 +518,10 @@ namespace MediaBrowser.XbmcMetadata.Parsers
                 case "releasedate":
                     if (reader.TryReadDateTimeExact(nfoConfiguration.ReleaseDateFormat, out var releaseDate))
                     {
-                        // Production year could already be set by the year tag
-                        if (item.ProductionYear is null)
-                        {
-                            item.ProductionYear = releaseDate.Year;
-                        }
-
                         item.PremiereDate = releaseDate;
+
+                        // Production year can already be set by the year tag
+                        item.ProductionYear ??= releaseDate.Year;
                     }
 
                     break;
--- a/MediaBrowser.XbmcMetadata/Parsers/BaseNfoParser.cs
+++ b/MediaBrowser.XbmcMetadata/Parsers/BaseNfoParser.cs
@@ -518,8 +518,13 @@ namespace MediaBrowser.XbmcMetadata.Parsers
                 case "releasedate":
                     if (reader.TryReadDateTimeExact(nfoConfiguration.ReleaseDateFormat, out var releaseDate))
                     {
+                        // Production year could already be set by the year tag
+                        if (item.ProductionYear is null)
+                        {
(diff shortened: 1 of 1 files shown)


MediaBrowser.XbmcMetadata/Parsers/BaseNfoParser.cs
```diff
@@ -518,8 +518,13 @@ protected virtual void FetchDataFromXmlNode(XmlReader reader, MetadataResult<T>
                 case "releasedate":
                     if (reader.TryReadDateTimeExact(nfoConfiguration.ReleaseDateFormat, out var releaseDate))
                     {
+                        // Production year could already be set by the year tag
+                        if (item.ProductionYear is null)
+                        {
+                            item.ProductionYear = releaseDate.Year;
```
cvium, line 524: Could save 3 lines with the null coalescing assignment operator

Bond-009 (approved):

crobibero (approved):

Issue #12155, opened 2024-06-22
[Issue]: Items in collections disappear after refreshing metadata


joshuaavalon:
### Please describe your bug

I found that no matter if you add item through Web UI or plugins. The items in collections will be removed after refreshing metadata.

### Reproduction Steps

1. Start a new Jellyfin 10.9.6 
2. Create a movie library with at least two movies that exists the same TMDb Box Sets
3. Install [Jellyfin TMDb Box Sets Plugin](https://github.com/jellyfin/jellyfin-plugin-tmdbboxsets)
4. Run the plugin the add movies to collections
5. Go the the collection and run `Refresh Metadata` > `Replace all metadata`
6. All the items in the collection disappear

I can confirm it works before 10.9

### Jellyfin Version

10.9.0

### Environment

```markdown
- OS: Debian  11
- Linux Kernel: N/A
- Virtualization: Docker
- Clients: Web
- Browser: Chrome
- FFmpeg Version: N/A
- Playback Method: Web
- Hardware Acceleration: Yes
- GPU Model: RTX 4060
- Plugins: Jellyfin TMDb Box Sets Plugin
- Reverse Proxy: Traefik
- Base URL: N/A
- Networking: N/A
- Storage: Local
```

### Jellyfin logs

```shell
Not related
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Liquoriced:
I wonder if this is connected to #12032 which seems to have been forgotten.

joshuaavalon:
I though it may be caused by merging data during refresh . So, I create #12156.

However, it may be caused by validation changes in `/MediaBrowser.Controller/Entities/Folder.cs` after I look around more in the code. I am not familiar with the code flow. Can maintainer take a look at it?

maximo1984:
> I wonder if this is connected to #12032 which seems to have been forgotten.

Yes, sad to see that my report is falling by the way side for such an important thing that needs remedy.

Shadowghost:
If you're using the boxset plugin this likely needs to be fixed over there. Issue should be that the LinkedChildren are discarded on merge instead of replaced.

joshuaavalon:
@Shadowghost I believe it does not belongs to the plugin because the plugin is only a metadata provider.
I have checked the API that metadata provider cannot access `LinkedChildren`. A metadata provider can only provide a metadata result (for user to choose), the merging is handled by Jellyfin

Shadowghost:
[...]

PR #12156, opened 2024-06-22
Fix refresh data does not retain LinkedChildren

joshuaavalon:
I think the issue is caused by new `LinkedChildren` introduced in 10.9. 
When refreshing metadata, a new item is created instead of updating. but `LinkedChildren` is not copied.

**Changes**
When refreshing metadata, the `LinkedChildren` should also be copied if the item is `Folder`.
Otherwise, all the added items will be removed on refreshing metadata.

**Issues**
Fixes: #12155


joshuaavalon:
Actually, this may not be the cause. Let discuss on the issue.

PR #12165, opened 2024-06-24
Improve LibraryManager.Sort when using "Random" order

### LINKED ISSUE #12162: [Issue]: `LibraryManager.Sort` does not sort "Random" with a uniform distribution
### Please describe your bug

When using the [LibraryManager.Sort](https://github.com/jellyfin/jellyfin/blob/d37c5d8921488c4d3602f7ed84e2dc0ae694b06c/Emby.Server.Implementations/Library/LibraryManager.cs#L1711-L1735) function with an `OrderBy` set to "Random", the algorithm uses a [RandomComparer](https://github.com/jellyfin/jellyfin/blob/d37c5d8921488c4d3602f7ed84e2dc0ae694b06c/Emby.Server.Implementations/Sorting/RandomComparer.cs) interface to compare two items at a time.
When using the RandomComparer, the "uniforme distribution" of the sort is poor as  50% of the time each element will remain in the same index it was previously, while 25% of the the element will be only be shuffled one index from it's initial position.

It would be better in this case to use the [Fisher-Yates shuffle](http://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle#The_modern_algorithm) algorithm to provide better distribution to the randomization.

As the project is using .NET 8, the `LibraryManager.Sort` function can be updated to use the [Random.Shuffle](https://learn.microsoft.com/en-us/dotnet/api/system.random.shuffle?view=net-8.0) method to shuffle the enumerablewith a uniform distribution. This can be done if comparer type is `RandomComparer`.

### Reproduction Steps

1. Set up server to contain multiple series that start with a different letter in the alphabet
   i.e
```
Shows/
    <Show That Starts with letter 'A'>
    <Show That Starts with letter 'B'>
    <Show That Starts with letter 'C'>
    <Show That Starts with letter 'D'>
    ...
```

2. Use the /Items api to query the list of "Series" using a `sortBy` option of "Random" and an `includeItemTypes` option of "Series".
    * The required options in this case are as follows
```json
{
    "includeItemTypes": ["'Series" ]
    "sortBy": [ "Random" ],
    "recursive": true
}
```
3. Notice that about 50% of the time, each show doesn't get randomly shuffled and remains in it's initial position

### Jellyfin Version

Master branch

### if other:

_No response_

### Environment

```markdown
- OS: ArchLinux
- Linux Kernel: 6.9.5
- Virtualization: None
- Clients: Server
- Browser: Firefox
- FFmpeg Version: 7.0.1
- Playback Method: None
- Hardware Acceleration: None
- GPU Model: GeForce RTX 3080
- Plugins: None
- Reverse Proxy: None
- Base URL: localhost:8096/api-docs/swagger/index.html
- Networking: None
- Storage: None
```

### Jellyfin logs

```shell
This is an algorithm issue with the `LibraryManager.Sort` method, the server isn't technically failing in anyway.
```

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

ItsAllAboutTheCode:
**Changes**

The `LibraryManager.Sort` function, now uses the `Random.Shared.Shuffle` method to randomly shuffle Enumerable items if the orderBy value is "Random". This fixes the issue with non-uniform distribution of items sorted at random.

What occurs is that when using the "Random" SortBy option, the RandomComparer is used by the  `IEnumerable` sorting logic to sort each element one at a time by generating two Random GUIDS and comparing them.

What this means is that if a list of items is as follows
`A B C D E`.

Then the first iteration of sorting would first "compare" `A` and `B` by generating a random GUID for each a comparing them.
*  `A < B` - 50% of the time the random guid generated for `A` would be less than `B` and therefore `A` would remain in it's current slot.
* Assuming `A` guid is greater than `B` (50% chance), then the next comparison is `A < C`. At which point it would be 50% that the random guid for `A` is less than the random guid for `C`

Extrapolating the iterations over the entire set of items, the probabilities for the slot  item `A` would reside in as follows.
`A` remains in first slot = 50%
`A` moves to second slot = 25%
`A` moves to third slot = 12.5%
`A` moves to nth slot = 1 / (2^n)

For `B` it would also have a 50% chance to reside in the first slot. Actually it has the exact same slot probabilities as `A`
`B` remains in first slot = 50%
`B` moves to second slot = 25%
`B` moves to third slot = 12.5%
`B` moves to nth slot = 1 / (2^n)

Because `C` element is first compared in the second iteration, it has a 50% chance of being in the _second slot_ with no  chance to be in the first slot.
`C` remains in first slot = 0%
`C` moves to second slot = 50%
`C` moves to third slot = 25%
`C` moves to fourth slot = 12.5%
`C` moves to nth slot = 1 / (2^(n - 1))

So the current `Random` sort algorithm bias elements to remain close to their original index.

**Issues**

fixes #12162


Emby.Server.Implementations/Library/LibraryManager.cs
```diff
@@ -1710,13 +1712,20 @@ private async Task<IEnumerable<IntroInfo>> GetIntros(IIntroProvider provider, Ba
...
-
             IOrderedEnumerable<BaseItem>? orderedItems = null;
 
             foreach (var orderBy in sortBy.Select(o => GetComparer(o, user)).Where(c => c is not null))
             {
-                if (isFirst)
+                if (orderBy?.GetType() == typeof(RandomComparer))
+                {
+                    BaseItem[]? randomItems = null;
+                    randomItems = items.ToArray();
```
crobibero, line 1722:
```suggestion
                    var randomItems = items.ToArray();
```

Emby.Server.Implementations/Library/LibraryManager.cs
```diff
@@ -1710,13 +1712,20 @@ private async Task<IEnumerable<IntroInfo>> GetIntros(IIntroProvider provider, Ba
...
         public IEnumerable<BaseItem> Sort(IEnumerable<BaseItem> items, User? user, IEnumerable<ItemSortBy> sortBy, SortOrder sortOrder)
         {
-            var isFirst = true;
-
             IOrderedEnumerable<BaseItem>? orderedItems = null;
 
             foreach (var orderBy in sortBy.Select(o => GetComparer(o, user)).Where(c => c is not null))
             {
-                if (isFirst)
+                if (orderBy?.GetType() == typeof(RandomComparer))
```
crobibero, line 1719:
```suggestion
                if (orderBy is RandomComparer)
```

Untested but should work

ItsAllAboutTheCode:
> Sorry I should have been clear, please apply the suggestions to both parts of the change

That is my mistake, I should have changed both functions.
It has been fixed now.

crobibero (approved):

Shadowghost (approved):

PR #12167, opened 2024-06-24
Changed `GetThemeMedia` to support SortBy/Order options

ItsAllAboutTheCode:
**Changes**

The `GetThemeMedia, `GetThemeVideos` and `GetThemeSongs` functions can optionally sort the results based based on passing an ItemSortBy type and a SortOrder.

This is intended to be used by jellyfin-web in order to allow users to control the order of theme playback. See PR: https://github.com/jellyfin/jellyfin-web/pull/5714

**Issues**

Relates to feature request: https://features.jellyfin.org/posts/2740/select-audio-from-the-theme-music-folder-at-random

There is a soft dependendcy on #12165 , which improves the random distribution of items when using the "Random" `ItemSortBy` type.


MediaBrowser.Controller/Entities/BaseItem.cs
```diff
@@ -2548,14 +2548,24 @@ public bool UpdateRatingToItems(IList<BaseItem> children)
...
-            return GetExtras().Where(e => e.ExtraType == Model.Entities.ExtraType.ThemeVideo).ToArray();
+            return LibraryManager.Sort(GetExtras().Where(e => e.ExtraType == Model.Entities.ExtraType.ThemeSong), user, orderBy).ToArray();
+        }
+
+        public IReadOnlyList<BaseItem> GetThemeVideos(User user = null)
+        {
+            return GetThemeVideos(user, Array.Empty<(ItemSortBy, SortOrder)>());
+        }
+
+        public IReadOnlyList<BaseItem> GetThemeVideos(User user = null, IEnumerable<(ItemSortBy SortBy, SortOrder SortOrder)> orderBy = null)
```
crobibero, line 2566:
```suggestion
        public IReadOnlyList<BaseItem> GetThemeVideos(User user = null, IEnumerable<(ItemSortBy SortBy, SortOrder SortOrder)> orderBy)
```

ItsAllAboutTheCode: Nice catch. Actually I'll update the suggestion as the `user` parameter in the two argument overload also shouldn't be defaulted in this case either.

The calling patterns should be
* `GetThemeVideos()` -> 1 argument overload
* `GetThemeVideos(user)` -> 1 argument overload
* `GetThemeVideos(user, orderBy)` -> 2 argument overload

```suggestion
        public IReadOnlyList<BaseItem> GetThemeVideos(User user, IEnumerable<(ItemSortBy SortBy, SortOrder SortOrder)> orderBy)
```

crobibero (approved):

Bond-009 (approved):

PR #11879, opened 2024-05-30
Migrate to z440.atl from TagLib-Sharp

gnattu:
The ATL lib provides a lot of advantages to the TagLib we are currently using.

Notably:

- auto-detect the format of the audio data, even if the file extension has the wrong label, and provides unified API for different file types.

- supports more audio formats than TagLib

- supports lyrics natively

- supports playlists and cuesheets

- provides relatively simple and controllable way for non-standard fields, enable us to implement compatibility features instead of waiting for lib updates

- is actually maintained

**Changes**

**Issues**

PR #12173, opened 2024-06-25
Add support for reading and storing Recording MBIDs from file metadata

lyarenei:
**Changes**

Add support for reading and storing MusicBrainz Recording IDs if they are present in the file metadata.

image

**Issues**

I'm not aware of any. It'll make the recording MBID conveniently accessible within Jellyfin (and editable from Jellyfin UI).


gnattu:
We will not be using this for reading tags. See #11879 

You need to utilize z440.atl to read that tag.

lyarenei:
> We will not be using this for reading tags. See #11879
> 
> You need to utilize z440.atl to read that tag.

I see, I'll wait for the changes then. Thanks for the info. :)

MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
```diff
@@ -1338,6 +1338,10 @@ private void SetAudioInfoFromTags(MediaInfo audio, Dictionary<string, string> ta
             mb = GetMultipleMusicBrainzId(tags.GetValueOrDefault("MusicBrainz Release Track Id"))
                  ?? GetMultipleMusicBrainzId(tags.GetValueOrDefault("MUSICBRAINZ_RELEASETRACKID"));
             audio.TrySetProviderId(MetadataProvider.MusicBrainzTrack, mb);
+
+            mb = GetMultipleMusicBrainzId(tags.GetValueOrDefault("MusicBrainz Track Id"))
```
gnattu, line 1342: I'm not a big fan of adding more tag parsing ability to the now (deprecated) ffprobe based parser, but I'm also fine with this if it does not have practical downsides.

lyarenei: This has been written before I knew this probing will be deprecated, so now it's kind of just being left here. I don't have any strong opinions either. 

As for usefulness... If we assume most users are using Picard for tagging, then ffmpeg can't even read the tag, so it's not going to be any better here. In that case I'd be for removing it. But I have no statistics on that, so I don't know.

lyarenei:
@gnattu Do you know if there is anything I can do to have this merged? I'd like to have this in the next major release if possible (10.11). Perhaps ask for a review on Matrix or somewhere?

MediaBrowser.Providers/Plugins/MusicBrainz/MusicBrainzRecordingId.cs
```diff
@@ -0,0 +1,27 @@
...
+    public string ProviderName => "MusicBrainz";
+
+    /// <inheritdoc />
+    public string Key => MetadataProvider.MusicBrainzRecording.ToString();
+
+    /// <inheritdoc />
+    public ExternalIdMediaType? Type => ExternalIdMediaType.Recording;
+
+    /// <inheritdoc />
+    public string UrlFormatString => Plugin.Instance!.Configuration.Server + "/recording/{0}";
```
gnattu, line 23: Will the metadata fetcher work correctly with this URL?

gnattu:
This generally looks okay, but I think it should be merged to master after the 10.10 branch has been cut.

MediaBrowser.Providers/Plugins/MusicBrainz/MusicBrainzRecordingId.cs
lyarenei: The URL is similar to all other MusicBrainz URLs, just with `recording` instead of for example `track`.

For example, `All the Small Things` recording and one of the specific track:
https://musicbrainz.org/recording/f36d9818-019b-4379-ad13-5080feb9ad8a
https://musicbrainz.org/track/27681a20-3aab-30b8-ab45-7fd274e6dfd5

Also, I have not implemented an `IRemoteMetadataProvider` for this as I don't think there is any value in that. But if implemented, it'd probably behave similarly to the `MusicBrainzAlbumProvider`.

Or are you maybe asking about something else?

gnattu: I'm just curious if our current metadata fetcher can handle this, but it looks like it does not for now, but that is fine.

lyarenei:
> This generally looks okay, but I think it should be merged to master after the 10.10 branch has been cut.

@gnattu Since 10.10 is already out, can we proceed with this?

Don't know what `a team` is (probably a deleted one?),
image
so tagging @Shadowghost as well.

lyarenei:
@gnattu @Shadowghost A gentle reminder on this, thank you.

gnattu (approved):

Bond-009 (approved):

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
```diff
@@ -355,6 +355,28 @@ private async Task FetchDataFromTags(Audio audio, Model.MediaInfo.MediaInfo medi
...
+                else if (track.AdditionalFields.TryGetValue("UFID", out var ufIdValue) && !string.IsNullOrEmpty(ufIdValue))
+                {
+                    // If tagged with MB Picard, the format is 'http://musicbrainz.org\0<recording MBID>
+                    if (ufIdValue.Contains("musicbrainz.org", StringComparison.OrdinalIgnoreCase))
+                    {
+                        var ufidParts = ufIdValue.Split('\0');
+                        if (ufidParts.Length > 1)
+                        {
+                            audio.TrySetProviderId(MetadataProvider.MusicBrainzRecording, ufidParts[1]);
+                        }
```
crobibero, line 375:
```suggestion
                        audio.TrySetProviderId(MetadataProvider.MusicBrainzRecording, ufIdValue.AsSpan().RightPart('\0').ToString());
```

lyarenei: Thanks, added in https://github.com/jellyfin/jellyfin/pull/12173/commits/33fe27be8c4f0d2226a3ad4d05670a4d0592d89d

lyarenei:
Failing tests seem to be the same as the ones failing in `master`.

JPVenson (approved):

JPVenson:
Yes, but there are tests that are failing for an actual reason
GetExternalStreams_MergeMetadata_HandlesOverridesCorrectly

lyarenei:
Yeah, I just wanted to point out for any reviewer looking at this that the changes in this pull request are fine and did not break any of these tests (as far as I can tell at least).

crobibero (approved):

PR #12196, opened 2024-06-30
fix: use release date from album metadata provider instead of year

joshuaavalon:
**Changes**
Disable updating `PremiereDate` from audio

**Issues**
If `EnableUpdatingPremiereDateFromChildren` is `true`, `AudioFileProber` override `PremiereDate` to `{year}-01-01` , even if provider like `MusicBrainzAlbumProvider` provide the exact date.

https://github.com/jellyfin/jellyfin/blob/478d8b07bf73d07c577ea4e39cbccab561d27b6a/MediaBrowser.Providers/MediaInfo/AudioFileProber.cs#L298-L314

https://github.com/jellyfin/jellyfin/blob/478d8b07bf73d07c577ea4e39cbccab561d27b6a/MediaBrowser.Providers/Plugins/MusicBrainz/MusicBrainzAlbumProvider.cs#L148-L154

### DIFF
--- a/MediaBrowser.Providers/Music/AlbumMetadataService.cs
+++ b/MediaBrowser.Providers/Music/AlbumMetadataService.cs
@@ -38,9 +38,6 @@ namespace MediaBrowser.Providers.Music
         }
 
         /// <inheritdoc />
-        protected override bool EnableUpdatingPremiereDateFromChildren => true;
-
-        /// <inheritdoc />
         protected override bool EnableUpdatingGenresFromChildren => true;
 
         /// <inheritdoc />


Shadowghost:
Maybe the better idea would be to only update it from children if it is not set

joshuaavalon:
@Shadowghost Currently, the children (track) does not set `PremiereDate`. There are several causes to it.

1. There is no standard format for date in id3 tags. While many agrees on the year, the are many format to save the date in tags. Even the library (`TagLib`) used by Jellyfin, only try to parse the year.
2. There is no track level metadata provider. It is current relies on file metadata, so a track does not have a  `PremiereDate`. because of point 1.
3. Current way of Jellyfin is just putting January 1st + year on `PremiereDate`. It is not the correct metadata anyway.
4. There is no way to fix on metadata provider level. Even `MusicBrainzAlbumProvider` provides `PremiereDate`, it is not used by `MetadataService`. Unless you change how `MetadataService` works, which affects collection and season,

In my opinion, disable `EnableUpdatingPremiereDateFromChildren` is the best option.

On the meta level, I believe `PremiereDate` of album should belong the album not individual track. It should be the other way around. We are trying to use `PremiereDate` from track is because track is a file but album is not.

gnattu:
>  Currently, the children (track) does not set PremiereDate

Wrong. It does.

joshuaavalon:
@gaam24 Sorry, I missed the code. I just found it.

https://github.com/jellyfin/jellyfin/blob/6225b92da092b378988f497ab8091c4820c49be8/MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs#L161-L168

Shadowghost:
@joshuaavalon is this PR still needed?

gaam24:
> @gaam24 Sorry, I missed the code. I just found it.
> 
> https://github.com/jellyfin/jellyfin/blob/6225b92da092b378988f497ab8091c4820c49be8/MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs#L161-L168

???
did you want to mention me, misspell or something?

nielsvanvelzen:
Closing as this targets an old, unsupported, release branch

PR #12201, opened 2024-06-30
Use complete paths in BD info

Bond-009:
This way we don't need to find the complete path later


Shadowghost (commented):
The check for actual existence of the files was intentional since the Blu-ray examiner result doesn't check for actual file existences.

Bond-009:
> The check for actual existence of the files was intentional since the Blu-ray examiner result doesn't check for actual file existences.

I can add the check back, but I don't understand when silently dropping segments is desirable. At best the runtime will be off and you'll miss some openings/credits, at worst you'll skip whole scenes.

Shadowghost (approved):
If files are missing, failing should be fine.

crobibero (approved):

PR #12226, opened 2024-07-06
Use real temp dir instead of cache dir for temp files

Bond-009:
This way we take advantage of systems where the temp directory is backed by RAM.

* Clean up temp files right after use where possible
* Prefer temp dir over cache/trancode dir for short lived one time use files
* Add more identifiable information to some temp files/directories
* Cache DVD/Blu-Ray concat file to improve subsequent playback start times

### DIFF
--- a/Emby.Server.Implementations/Images/BaseDynamicImageProvider.cs
+++ b/Emby.Server.Implementations/Images/BaseDynamicImageProvider.cs
@@ -122,6 +122,7 @@ namespace Emby.Server.Implementations.Images
             }
 
             await ProviderManager.SaveImage(item, outputPath, mimeType, imageType, null, false, cancellationToken).ConfigureAwait(false);
+            File.Delete(outputPath);
 
             return ItemUpdateType.ImageUpdate;
         }
--- a/Emby.Server.Implementations/AppBase/BaseApplicationPaths.cs
+++ b/Emby.Server.Implementations/AppBase/BaseApplicationPaths.cs
@@ -104,6 +104,6 @@ namespace Emby.Server.Implementations.AppBase
         /// Gets the folder path to the temp directory within the cache folder.
         /// </summary>
         /// <value>The temp directory.</value>
-        public string TempDirectory => Path.Combine(CachePath, "temp");
+        public string TempDirectory => Path.Join(Path.GetTempPath(), "jellyfin");
     }
 }
--- a/Emby.Server.Implementations/IO/ManagedFileSystem.cs
+++ b/Emby.Server.Implementations/IO/ManagedFileSystem.cs
@@ -466,7 +466,7 @@ namespace Emby.Server.Implementations.IO
             File.Copy(file1, temp1, true);
 
             File.Copy(file2, file1, true);
-            File.Copy(temp1, file2, true);
+            File.Move(temp1, file2, true);
         }
 
         /// <inheritdoc />
--- a/Emby.Server.Implementations/ScheduledTasks/Tasks/AudioNormalizationTask.cs
+++ b/Emby.Server.Implementations/ScheduledTasks/Tasks/AudioNormalizationTask.cs
@@ -29,7 +29,7 @@ public partial class AudioNormalizationTask : IScheduledTask
     private readonly IItemRepository _itemRepository;
     private readonly ILibraryManager _libraryManager;
     private readonly IMediaEncoder _mediaEncoder;
-    private readonly IConfigurationManager _configurationManager;
+    private readonly IApplicationPaths _applicationPaths;
     private readonly ILocalizationManager _localization;
     private readonly ILogger<AudioNormalizationTask> _logger;
 
@@ -39,21 +39,21 @@ public partial class AudioNormalizationTask : IScheduledTask
     /// <param name="itemRepository">Instance of the <see cref="IItemRepository"/> interface.</param>
     /// <param name="libraryManager">Instance of the <see cref="ILibraryManager"/> interface.</param>
     /// <param name="mediaEncoder">Instance of the <see cref="IMediaEncoder"/> interface.</param>
-    /// <param name="configurationManager">Instance of the <see cref="IConfigurationManager"/> interface.</param>
+    /// <param name="applicationPaths">Instance of the <see cref="IApplicationPaths"/> interface.</param>
     /// <param name="localizationManager">Instance of the <see cref="ILocalizationManager"/> interface.</param>
     /// <param name="logger">Instance of the <see cref="ILogger{AudioNormalizationTask}"/> interface.</param>
     public AudioNormalizationTask(
         IItemRepository itemRepository,
         ILibraryManager libraryManager,
         IMediaEncoder mediaEncoder,
-        IConfigurationManager configurationManager,
+        IApplicationPaths applicationPaths,
         ILocalizationManager localizationManager,
         ILogger<AudioNormalizationTask> logger)
     {
         _itemRepository = itemRepository;
         _libraryManager = libraryManager;
         _mediaEncoder = mediaEncoder;
-        _configurationManager = configurationManager;
+        _applicationPaths = applicationPaths;
         _localization = localizationManager;
         _logger = logger;
     }
@@ -107,7 +107,9 @@ public partial class AudioNormalizationTask : IScheduledTask
                 }
 
                 _logger.LogInformation("Calculating LUFS for album: {Album} with id: {Id}", a.Name, a.Id);
-                var tempFile = Path.Join(_configurationManager.GetTranscodePath(), a.Id + ".concat");
+                var tempDir = _applicationPaths.TempDirectory;
+                Directory.CreateDirectory(tempDir);
+                var tempFile = Path.Join(tempDir, a.Id + ".concat");
                 var inputLines = albumTracks.Select(x => string.Format(CultureInfo.InvariantCulture, "file '{0}'", x.Path.Replace("'", @"'\''", StringComparison.Ordinal)));
                 await File.WriteAllLinesAsync(tempFile, inputLines, cancellationToken).ConfigureAwait(false);
                 try
--- a/Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
+++ b/Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
@@ -230,7 +230,7 @@ public class TrickplayManager : ITrickplayManager
             throw new ArgumentException("Can't create trickplay from 0 images.");
         }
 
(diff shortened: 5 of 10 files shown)


gnattu:
Please note that this is not entirely safe on all OS. Some OS will automatically remove files in tmp folder to reclaim space which is not desired. I would want to have this behind an option and disable by default.

Bond-009:
@gnattu That's that's fine, we do that too. By default we delete everything in the temp folder that hasn't been written to in 1 day. 
As I've said, this is mostly for single use files.

gnattu:
Maybe this path to the start-up logging in `StartupHelper.cs`? This path returned by `Path.GetTempPath()` is not always that trivial and could be very hard to find if it is not printed. For example on my system it returns:
`/var/folders/0f/fnxhvc5d5k3f2src8_qkq8vw0000gn/T/`

Bond-009:
If they want to change it they can, using the TMPDIR env var (on linux)

Shadowghost (approved):

gnattu:
I'm still a bit concerned about this change. At least this change should be documented very clearly with instructions on how to disable this. A lot of side effects would be made by this and I could see that a lot of users will have accidentally put temp files in to the wrong drive he intended to use because the location returned by `Path.GetTempPath()` is not always a tmpfs, and on OSes like Windows this path in controlled by a lot of factors that is probably out of our user's knowledge. Also, this API does not guarantee the folder returned is writeable for the current user which I could see that a lot of them will encounter permission issue and struggles to deal with that. What's more, for container based installs, this option is not the ideal option by default because it directly writes data to the container instead of a mounted folder, as /tmp is not mounted as tmpfs by default.

Bond-009:
> A lot of side effects would be made by this and I could see that a lot of users will have accidentally put temp files in to the wrong drive he intended to use because the location returned by `Path.GetTempPath()` is not always a tmpfs, and on OSes like Windows this path in controlled by a lot of factors that is probably out of our user's knowledge.

Users shouldn't care that much, this PR just provides a better default for the few temp files Jellyfin has. (most of which get deleted instantly after use)

> Also, this API does not guarantee the folder returned is writeable for the current user which I could see that a lot of them will encounter permission issue and struggles to deal with that.

If the returned folder is not writeable the system is misconfigured.

> What's more, for container based installs, this option is not the ideal option by default because it directly writes data to the container instead of a mounted folder, as /tmp is not mounted as tmpfs by default.

Again this is just a better default. Why would it matter if the files are on a mountpoint? They are temp files, countless docker containers do it like this. And if it is a problem this can be fixed in the image by setting $TMPDIR.

gnattu:
> Users shouldn't care that much, this PR just provides a better default for the few temp files Jellyfin has. (most of which get deleted instantly after use)

It is NOT always a better default, which is the problem here.

> If the returned folder is not writeable the system is misconfigured.

Or write to rootfs is prevented by certain security mechanisms unless explicitly approved.

> Again this is just a better default. Why would it matter if the files are on a mountpoint? They are temp files, countless docker containers do it like this. And if it is a problem this can be fixed in the image by setting $TMPDIR.

Consider this use case: a user is running jellyfin on RK3588 based SBC, an officially supported platform, have its OS installed to internal emmc or an sdcard, and mounted an nvme drive for jellyfin. This new default will default write to the slower sdcard/emmc storage instead of the nvme drive unless the user explicitly configured the env var to overwrite it. This is not a better default, this is a regression.

Consider another use case: The user deployed this on a k8s cluster with `securityContext: readOnlyRootFilesystem: true`, this new default will break such install totally unless the user explicitly define the TMPDIR env var.

Consider another use case: The user deployed jellyfin on Windows and find out that Jellyfin continuously writes to his precious C drive and he is unable to find out an easy way to stop this, and we have to modify our tray app to load an env var to support change temp folder.

All of above are just some quick use cases I could think of and probably there are more. This new default is better if and only if the system implements tmpfs and the `/tmp` folder or `TMPDIR` points to a tmpfs and the system has sufficient ram to not to write to swap when the video files are writing to the tmpfs. You could still argue that the user could always overwrite this behavior by defining an additional env var, but at least such behavior change needs to be well documented to inform our users. They will get confused by the new behavior and will create github issues claiming this is a bug.

Bond-009:
> It is NOT always a better default, which is the problem here.

How is honoring the system temp path not a better default than putting tmp files in our own cache dir. Following conventions is what most users would expect in the first place.

> Or write to rootfs is prevented by certain security mechanisms unless explicitly approved.

Again, if the temp dir isn't writeable that's a system misconfiguration.

> Consider this use case: a user is running jellyfin on RK3588 based SBC, an officially supported platform, have its OS installed to internal emmc or an sdcard, and mounted an nvme drive for jellyfin. This new default will default write to the slower sdcard/emmc storage instead of the nvme drive unless the user explicitly configured the env var to overwrite it. This is not a better default, this is a regression.

Well all other applications running on that SBC presumably also use the default temp dir, if the user wishes to change the temp dir he/she can using the STANDARDIZED tools/options (change /tmp mount or $TMPDIR). They can configure it system wide or application specific.

> This is not a better default, this is a regression.

I'd argue that following standard practice is an improvement even if the resulting tmp dir is less performant.

> Consider another use case: The user deployed jellyfin on Windows and find out that Jellyfin continuously writes to his precious C drive and he is unable to find out an easy way to stop this, and we have to modify our tray app to load an env var to support change temp folder.

We can always document how to change the location of the temp dir for the different platforms, at least it will be (mostly) the same as for all other applications on that platform.

> and the system has sufficient ram to not to write to swap when the video files are writing to the tmpfs.

We don't write video files to the tmp dir, we have a separately configurable transcode dir that defaults to inside our cache dir.

nielsvanvelzen (approved):

Bond-009:
Extra context:
* Currently there is no way to configure the temp path of Jellyfin, it's always inside of it's cache path (so no option to use tmpfs if you want)
* The system tmp dir is already used by dotnet and ASP.NET, so "Jellyfin" is using it regardless of this PR, just not for files explicitly created by us.
https://source.dot.net/#Microsoft.AspNetCore.Http/src/Http/WebUtilities/src/AspNetCoreTempDirectory.cs
https://source.dot.net/#System.Formats.Tar/System/Formats/Tar/TarHelpers.Unix.cs,25

Bond-009:
@Gylesie You can already configure the transcode directory in the GUI
dashboard > playback > transcoding > transcode path

joshuaboniface:
I think that as long as the files in question are small and transient, this is file. While I can see some of @gnattu's points here, especially if these were larger files, for a few small files it's far more likely that the user will already have `/tmp` or similar mounted as a `tmpfs` which will always be faster, and the small size and quick turnaround for removal would prevent too much RAM usage. However if this is getting used for anything more than a few dozen files of a few kB each, it will get really hairy really quick. So if we're going to do this we have to be sure that it's actually just a few small files here or there.

Bond-009:
```sh
~/.cache/jellyfin/temp$ du -sh .
8.7M    .
~/.cache/jellyfin/temp$ find . -type f | wc -l
872
```

We should be fine

joshuaboniface:
> ```shell
> ~/.cache/jellyfin/temp$ du -sh .
> 8.7M    .
> ~/.cache/jellyfin/temp$ find . -type f | wc -l
> 872
> ```
> 
> We should be fine

Agreed, just checked mine and it's 1.2MB in 6 files. This is fine for a tmpfs.

gnattu:
The biggest transient files are probably the trickplay intermediate frames. They could be larger at ~20-30MB per video but not that huge, and they are going to be removed after the generation is done for current video.

joshuaboniface:
Yea I think 20MB or so is still nicely under the limit where I'd get concerned about tempfiles (I said `kB` when I meant `MB` haha). Even in a really constrained 1GB system with a 512MB tmpfs that should be fine for transient files.

lukaszsobala:
I'm not sure if it's due to this change, but I think it caused my Rock 5B to stop booting after generating the trickplay images (over 300 GB of them). It seems to be stuck at "Job systemd-tmpfiles-setup.service/start running" - this has no limit. Should I make a new issue?

crobibero:
> I'm not sure if it's due to this change, but I think it caused my Rock 5B to stop booting after generating the trickplay images (over 300 GB of them). It seems to be stuck at "Job systemd-tmpfiles-setup.service/start running" - this has no limit. Should I make a new issue? 

No, 10.10.1 was just released with a fix for this issue.

gnattu:
> No, 10.10.1 was just released with a fix for this issue.

I'm not sure if it will fix his issue because trickplay temp files are cleaned up immediately before 10.10.1.

Anyway, I highly recommend RK3588 users to use `TMPDIR` env var to set the temp file to an ssd if the tmpfs implementation is not reliable enough for large amount of temp files.

lukaszsobala:
Yup, the rootfs is on an SSD.

gnattu:
OK let me explain this more clearly, even if your rootfs is an SSD, the default value that uses the system temp will use some mechanisms that is more than a simple file system to handle those temp files. For most OS that is just a volatile fs which is cleared on reboot, but on some OS it will actually makes the boot up process to struggle when the system temp folder has tons of temp files. This is why you may want to set the env variable to point to an alternate folder that is not the system one.

Emby.Server.Implementations/Images/BaseDynamicImageProvider.cs
```diff
@@ -122,6 +122,7 @@ protected async Task<ItemUpdateType> FetchToFileInternal(
             }
 
             await ProviderManager.SaveImage(item, outputPath, mimeType, imageType, null, false, cancellationToken).ConfigureAwait(false);
+            File.Delete(outputPath);
```
IDisposable, line 125: Seems a bit odd to save the file and then immediately delete it...

gnattu: > Seems a bit odd to save the file and then immediately delete it...

Because output path here is always a temp folder and the save function is what to save it into the metadata folder

lukaszsobala:
@gnattu that's the problem... I've been waiting for 1 day now and nothing. I'll wait up to 3 days.

gnattu:
> @gnattu that's the problem... I've been waiting for 1 day now and nothing. I'll wait up to 3 days.

If it is that slow maybe just remove the files yourself would be faster

lukaszsobala:
> > @gnattu that's the problem... I've been waiting for 1 day now and nothing. I'll wait up to 3 days.
> 
> If it is that slow maybe just remove the files yourself would be faster 

If it was this easy I would, but it requires dismantling the device and putting the SSD into an enclosure (I've flashed the SPI with nvme boot image).

gnattu:
If they were really trickplay files then my theory is that RK3588 generates those files too slow so the temp files just stack up. But I could be wrong, anyway point to a folder that is not the system temp is recommended on such platform.

Issue #12238, opened 2024-07-08
[Issue]: Crashes running on low disk space


lugrinder:
### Please describe your bug

When the app is running with low free space, it crashes and can't restart until manual cleaning is done. It's normal for an app to malfunction under low disk space conditions, but it should at least manage its generated data to prevent such issues and auto-clean if it can't start. In my case, the problem was with a growing transcode cache during playback; cache and running app uses the same drive. I fixed it by changing this folder to another disk with more space. However, if that disk also runs low on free space, it stops the player but doesn't crash the app. This way, the app still has enough space to run, but not enough to transcode.

### Reproduction Steps

1.Run the app in a small drive/partition, 8 or 16Gb, is enough to run Jellyfin.
2.Ensure that all caches and app uses the same drive/partition.
3.Play any content that uses transcode.
4.Wait until transcode fills the cache.
5.Player will stop and Jellyfin service can't start again.
6.Clean transcode cache manually (rm) and service will start without any problem.

### Jellyfin Version

10.9.0

### if other:

10.9.7

### Environment

```markdown
- OS: Ubuntu 22.04.4 LTS
- Linux Kernel: Linux 6.8.8-1-pve
- Virtualization: LXC (Proxmox VE 8.2.4)
- Clients: Jellyfin Tizen Build (web client)
- Browser: Not applicable
- FFmpeg Version: 6.0.1-Jellyfin
- Playback: Direct Stream and Remux
- Hardware Acceleration: VAAPI
- GPU Model:Intel UHD Graphics (Intel N100)
- Installed Plugins: Intro Skipper, LrcLib, Merge Versions, Subtitle Extract, TMDb, Studio Images, OMDb, Music Brainz, AudioDB
- Reverse Proxy: nginx
- Base URL: none
- Networking: Bridge
- Storage: local (16Gb SSD)
```

### Jellyfin logs

```shell
Comment is too long (maximum is 65536 characters)
```

### FFmpeg logs

```shell
[...]

PR #12239, opened 2024-07-08
Run DeleteTranscodeFileTask on startup

Bond-009:
Possible fix for #12238

Also removed `TriggerSystemEvent` as it doesn't exist AFAIK


Shadowghost (approved):
Might be worth a backport

crobibero (approved):

PR #12050, opened 2024-06-09
Fix season handling

Shadowghost:
See linked issue for details. For some reason `DisplaySpecialsWithinSeasons` breaks season creation if there are no season folders. I still don't really understand how, but this fixes it.

**Changes**
* Remove optimization which breaks season checks
* Re-add saving existing seasons forcefully as non-virtual if they are virtual

**Issues**
Fixes #11575
Fixes #12045

PR #12240, opened 2024-07-08
Fix season handling ("Season Unknown" / unneccesary empty seasons)

### LINKED ISSUE #12208: [Issue]: Seasons and Specials showing when files not present
### Please describe your bug

Since updating to 10.9.7 from 10.9.6, I have noticed a bug where some shows will now show the "Specials" season when there are no specials available for that show. Please see example below:
image

Some shows, for example "Love Island", is showing me ALL seasons when I only have the most recent downloaded. All other seasons are empty and there are no files to suggest that Jellyfin needs to show and populate them. Please see example below:
image

### Reproduction Steps

When running 10.9.7, check shows for specials if enabled.

When also running 10.9.7, check shows which have multiple seasons, but only a few downloaded locally.

### Jellyfin Version

10.9.0

### if other:

_No response_

### Environment

```markdown
- OS: Slackware 15.0
- Linux Kernel: 6.1.79-Unraid
- Virtualization: Docker (LinuxServer Image)
- Clients: Andriod, WebOS / Client, Swiftfin.
- Browser: Firefox & Chrome
- FFmpeg Version: latest
- Playback Method: N/A
- Hardware Acceleration: N/A
- GPU Model: N/A
- Plugins: Default + TMDB, Intro Skipper, Infuse Sync
- Reverse Proxy: Yes
- Base URL: N/A
- Networking: N/a
- Storage: N/A
```

### Jellyfin logs

```shell
No logs can be found to connect to this bug.
```

### FFmpeg logs

```shell
No logs can be found to connect to this bug.
```

_No response_

_No response_

### Code of Conduct

nielsvanvelzen:
Reverts a part of #12050.

**Changes**

Starting with 10.9.7, when refreshing metadata or scanning a series it would reset the virtual season to "Season Unknown". Reverting a part of the 10.9.7 related to season lookups fixed this.

~~Probably needs more testing for non-virtual season usages.~~

**Issues**

Fixes #12208 (confirmed via forum)

### DIFF
--- a/MediaBrowser.Controller/Entities/TV/Series.cs
+++ b/MediaBrowser.Controller/Entities/TV/Series.cs
@@ -350,10 +350,17 @@ namespace MediaBrowser.Controller.Entities.TV
 
         public List<BaseItem> GetSeasonEpisodes(Season parentSeason, User user, DtoOptions options, bool shouldIncludeMissingEpisodes)
         {
+            var queryFromSeries = ConfigurationManager.Configuration.DisplaySpecialsWithinSeasons;
+
+            // add optimization when this setting is not enabled
+            var seriesKey = queryFromSeries ?
+                GetUniqueSeriesKey(this) :
+                GetUniqueSeriesKey(parentSeason);
+
             var query = new InternalItemsQuery(user)
             {
-                AncestorWithPresentationUniqueKey = null,
-                SeriesPresentationUniqueKey = GetUniqueSeriesKey(this),
+                AncestorWithPresentationUniqueKey = queryFromSeries ? null : seriesKey,
+                SeriesPresentationUniqueKey = queryFromSeries ? seriesKey : null,
                 IncludeItemTypes = new[] { BaseItemKind.Episode },
                 OrderBy = new[] { (ItemSortBy.SortName, SortOrder.Ascending) },
                 DtoOptions = options


Bond-009 (approved):

Shadowghost:
@sjorge please take a look

sjorge:
@Shadowghost still havent figured out how to build jellyfin myself. I can certainly retest once thus hits the next release.

fernvenue:
Hi, I just tested this version, it works perfectly fine on my device, and the unneccesary empty seasons doesn't show anymore :)

X012C:
I deleted the NFO files and it caused an unknown season to appear. Now I can't delete these non-existent error messages.

crobibero (approved):

JamsRepos:
> Hi, I just tested this version, it works perfectly fine on my device, and the unneccesary empty seasons doesn't show anymore :)

How did you manage to test it no longer appearing? I'm on the latest release which this seems to be included in and I don't see any indication of it being fixed. I still have blank seasons and specials, after refreshing metadata to make them disappear, then they re-appear on a library scan.

sjorge:
@Shadowghost sadly https://github.com/jellyfin/jellyfin/issues/11575 is broken again, seasons are no longer created when 'Display specials within seasons they are aired' is unchecked.

sjorge:
Even worse, with the work around of enabling the option a lot of other weirdness is happening now.

All shows that are completed, now show 1 empty blue dot indicator.

image
image
image

I'm just going to run the risk of a downgrade breaking everything and try and roll back.

fernvenue:
> How did you manage to test it no longer appearing?

Hi @JamsRepos, I manually ran the branch `nielsvanvelzen:season-unknown`, and it performed well during testing. I have also upgraded to the latest version on my server, restarted the service, and after rescanning the library, there is no longer an issue with unknown seasons.

> EDIT: I've restarted the server since upgrading and tried again, seems to be working. I'll test further.

Great, I think it would be better to have more people test it :)

> Exit: is it safe to downgrade? There are no new migrations in the .8 release right?

Hi @sjorge, not sure about that, give it a snapshot before you downgrade will be better maybe?

Shadowghost:
@sjorge it's safe to downgrade, no database migrations in 10.9.8. We only had one in 10.9.4
@nielsvanvelzen seems like #12240 did break this logic again (as mentioned in the original ticket there is a bug in the ancestor calculation somewhere)

sjorge:
Ok downgrade went fine, am traveling so was way harder to do than usual as I did not make a vm backup before upgrading.

10.9.8: regressed to https://github.com/jellyfin/jellyfin/issues/11575 behavior, but the workaround of having the option checked no longer works. the partial revert seems to have made worse to what we had pre-10.9.7.
10.9.7: creates seasons fine with the option checked or unchecked. I'm not getting additional ghost seasons, what the partial revert tried to address.

I did the testing by dropping a .ignore in a show folder, triggering 2 full rescans, removing the .ignore and triggering 1 full rescan.

Bit hard to provide full logs on my phone though. If need be I can make fresh ones in ~1 week when I get home.

sjorge:
> @nielsvanvelzen seems like this PR did break the logic again (as mentioned in the original ticket there is a bug in the ancestor calculation somewhere), so we need to find another way to fix this.

Broken ancestry would also explain the weird new unwatched status issue I think.

Shadowghost:
If I remember correctly from my debugging, ancestry is borked only for virtual seasons because those are created on-demand after scanning. In this case we need to recalculate the episode ancestors but that is somewhat broken.

KimPig:
Even after the update, virtual seasons still exist. Scanning all the libraries and restarting the server both didn't help me.

p37307:
> Even after the update, virtual seasons still exist. Scanning all the libraries and restarting the server both didn't help me.

Same. I rescanned, restarted, rescanned. restarted. Windows 10 JF 10.9.8

I even did a couple series on their own. 


Anything special we need to do? I added a reply to the forum post. Not sure where people want to discuss this issue. https://forum.jellyfin.org/t-how-to-disable-empty-seasons-missing-episode-fetcher?pid=30445#pid30445

p37307:
> You need to scan your library and select replace all metadata to remove the virtual seasons from your library.

Thanks for your response, but I don't want to replace all my metadata. 

After reading your response, I created a batch file that created "Specials" folders in the root of all my series folders, manually ran a scan for new and updated. That created an actual path to the folder in the system instead of it being virtual. Afterwards, I then reversed my batch script and deleted the "Specials" folders to get rid of them. Luckily, all my real specials were in "Season 00" folders so I could get away with this. Looks like the issue is resolved. 

As a side product of doing this, it appears a few library issues I was having has been solved as it picked up several items it should have had in the library but for whatever reason, weren't there.

Suzaru:
> You need to scan your library and select replace all metadata to remove the virtual seasons from your library.

Hey, sadly this didn't fix the virutal seasons for me. I'm trying to fix this, but they re-appear everytime. I tried full replace, deleting images, restarting, everything. Any ideas?

KimPig:
Is there any other solution? I don't want to receive the metadata again because I have modified it.

sjorge:
Probably wait until the ancestry is properly fixed, I would expect a normal scan to then notice there is a virtual season with no episodes and removes it.

Suzaru:
So I tried everything and this is definitely not fixed fully, sadly. Future seasons for example will show up, when sonarr has them too. I deactivated everything I could, but there is just nothing else left to do. Even rebuilding series fully isn't working. The empty seasons will reappear eventually.

Guess we have to wait. Don't waste ~18h trying to fix it, like I did. Not worth it.

sjorge:
There is still an issue with item ancestry.The original PR fixed it incompletely, the partial revert made it worse IMHO.~ sjorgeOn 24 Jul 2024, at 18:07, jotpunktopunkt ***@***.***> wrote:﻿
i can only add that the issue persists - but haven't completely rebuilt my library, as i would like to avoid that work
the question i have is: is there actually still an issue (and was it acknowledged by the dev team?) or is the (lasting) solution to rebuild the library?

nielsvanvelzen:
This PR reverted the part that made my set-up work again and it was then confirmed by multiple other users that it also fixed their problems. It's important to know that this change does not magically fix the extra seasons. You'll need to go to the individual series and use the "search for missing metadata" option (or replace all). It does **not** fix anything when doing a library scan.

From the reports it's obvious there are still more issues to be solved. I'd ask everyone to please make a proper issue for that instead of commenting on this pull request.

nielsvanvelzen:
There were different kinds of issues related to seasons so it is possible the issue you're having is not solved yet.

Aydonger:
Sorry, but the Jellyfin Version .8 doesnt fix the Problem.
Did a Refresh / Rescan / Search for Missing MetaData - it seems to be solved, but come back after a couple of hours.

What should we do?

RandomNamer:
It seems the fix is to make sure episode count (including all virtual episodes) is correct based on "display specials in season" setting. I think it's fine to have completely virtual seasons with  only virtual episodes, however from user's perspective, seasons that doesn't include any physical episodes appearing in series page seems buggy. Probably we can improve `"{seriesId}/Seasons"` request handling by adding the same `DisplayMissingEpisodes` logic as `"{seriesId}/Episodes"` does, and all virtual seasons should be filtered at the server side according to user settings. This should be a more consistent experience and does not require rebuilding the metadata.

Aydonger:
So, i updated to the actual Version (10.09.08). I do the "scan for missing metadata" - no luck. The other Scan-Methods also. Nothing.

I delete the whole TV-Series-Library and rebuild it. And yes: there where only the Episodes, which i have on my HDD.

Fine. Not.

Because the next day, the bug was back (or to be true, it was never gone): missing Seasons populates my TV-Shows again.

So i can confirm, that the "bugfix" or whatever sadly doesnt work. The Problem seems much deeper, than just a roll back...

Issue #12143, opened 2024-06-20
[Issue]: Images resolved from the "Edit Images" modal aren't using the provider


revam:
### Please describe your bug

Follow up for #12074.

Currently the resolution of remote images provided by plugins are handled differently in two places;

- If a remote image is acquired internally in Jellyfin it will ask the provider to fetch the image for us. As seen [here](https://github.com/jellyfin/jellyfin/blob/18dd6b2875b05d6b9c8249476ae509bfd20e8314/MediaBrowser.Providers/Manager/ItemImageProvider.cs#L535) and [here](https://github.com/jellyfin/jellyfin/blob/18dd6b2875b05d6b9c8249476ae509bfd20e8314/MediaBrowser.Providers/Manager/ItemImageProvider.cs#L644).

- If a remote image is asked by a client to be added to an item, it will always be downloaded using a standard HTTP client. As seen [here](https://github.com/jellyfin/jellyfin/blob/6fb6b5f1766a1f37a61b9faaa40209bab995bf30/Jellyfin.Api/Controllers/RemoteImageController.cs#L166). I'll also note that the client is sending the image url, provider name, and image type with each request. As seen [here](https://github.com/jellyfin/jellyfin-web/blob/20381bd3ec2f23907ed53d7da568cae921b05506/src/components/imageDownloader/imageDownloader.js#L140-L154).

@thornbill pointed out that https://github.com/jellyfin/jellyfin/security/advisories/GHSA-rgjw-4fwc-9v96 existed in the matrix dev. channel, so the new solution must be resistant to this attack vector.

@cvium mentioned in https://github.com/jellyfin/jellyfin/pull/12074#issuecomment-2174448293
> […] Ideally, and this is just my opinion:
>
>  1. All images are proxied through the Jellyfin backend
>
>  2. Identify (and other remote download flows?) stores an ID per image
>
>  3. Remote image requests always use the ID and based on the ID, the server knows what to download and store
>
>
> EDIT: This is a pretty big API breaking change, but ultimately the only reasonable (imo) fix for the security advisory. To ease clients into it, we could create a new set of endpoints.

This issue exists to find the best way to solve it if cvium's solution is not enough, or for somebody qualified to undertake the "pretty big API breakage change" to have a reference and/or reminder that it needs fixing at some point.

### Reproduction Steps

[...]

PR #12245, opened 2024-07-09
Broken Image Downloads for Azure Storage files.

mjohl:
**Changes**
- Improved error handling on image download failures to prevent log flooding and provide more info about the status code from the source.
- Addition to the SaveImage method to handle application/octet-stream from Azure Blob Storage when providers do not set the content type.

**Issues**
#12143 

Pre-Change
Screenshot 2024-07-07 164602

Post-Change
Screenshot 2024-07-07 164921

### DIFF
--- a/Emby.Server.Implementations/Library/LibraryManager.cs
+++ b/Emby.Server.Implementations/Library/LibraryManager.cs
@@ -2834,27 +2834,22 @@ namespace Emby.Server.Implementations.Library
                 }
                 catch (HttpRequestException ex)
                 {
-                    if (ex.StatusCode is not (HttpStatusCode.NotFound or HttpStatusCode.Forbidden))
-                    {
-                        throw;
-                    }
-
-                    switch (ex.StatusCode.Value)
+                    switch (ex.StatusCode)
                     {
                         case HttpStatusCode.NotFound:
-                            _logger.LogDebug(ex, "Image {Url} not found: {HttpStatusCode}", url, ex.StatusCode.Value);
+                            _logger.LogWarning(ex, "Image not found response received: {HttpStausCode} {URL}", ex.StatusCode, url);
                             break;
                         case HttpStatusCode.Forbidden:
-                            _logger.LogDebug(ex, "Forbidden to access image: {HttpStausCode} {URL}", ex.StatusCode.Value, url);
+                            _logger.LogWarning(ex, "Forbidden response received for image: {HttpStausCode} {URL}", ex.StatusCode, url);
                             break;
                         default:
-                            _logger.LogDebug(ex, "The requestes to get the image failed: {HttpStausCode} {URL}", ex.StatusCode.Value, url);
-                            throw;
+                            _logger.LogWarning(ex, "Image download failed: {HttpStausCode} {URL}", ex.StatusCode, url);
+                            break;
                     }
                 }
                 catch (Exception ex)
                 {
-                    _logger.LogError(ex, "Error converting image to local: {Url}", url);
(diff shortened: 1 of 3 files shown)


Emby.Server.Implementations/Library/LibraryManager.cs
```diff
@@ -2834,19 +2834,34 @@ public async Task<ItemImageInfo> ConvertImageToLocal(BaseItem item, ItemImageInf
...
+                    {
+                        case HttpStatusCode.NotFound:
+                            _logger.LogDebug(ex, "Image {Url} not found: {HttpStatusCode}", url, ex.StatusCode.Value);
+                            break;
+                        case HttpStatusCode.Forbidden:
+                            _logger.LogDebug(ex, "Forbidden to access image: {HttpStausCode} {URL}", ex.StatusCode.Value, url);
+                            break;
+                        default:
+                            _logger.LogDebug(ex, "The requestes to get the image failed: {HttpStausCode} {URL}", ex.StatusCode.Value, url);
+                            throw;
```
Shadowghost, line 2852: This will never be called since you already filter before entering the switch

Jellyfin.Api/Controllers/ImageController.cs
```diff
@@ -1961,7 +1961,19 @@ public ActionResult DeleteCustomSplashscreen()
...
+            {
+                imageInfo = await _libraryManager.ConvertImageToLocal(item, imageInfo, imageIndex ?? 0).ConfigureAwait(false);
+            }
+            catch (Exception ex)
+            {
+                // Log the exception. Adjust the logging mechanism based on your application's logging setup.
+                _logger.LogError(ex, "Failed to convert image to local for item {ItemId} and imageIndex {ImageIndex}", item.Id, imageIndex);
+
+                // Fallback action: return a default image or an error response.
+                // This example returns a NotFound result. Replace this with your chosen fallback action.
```
Shadowghost, line 1974: What's up with those comments? Seems like boilerplate?

mjohl: @Shadowghost  Happy to remove - Old habits from when I did training and copilot doing it's thing.

mjohl (commented):
Changes have been made off of the comments from the first PR

MediaBrowser.Providers/Manager/ProviderManager.cs
```diff
@@ -216,6 +216,27 @@ await using (cacheStream.ConfigureAwait(false))
                     contentType = MimeTypes.GetMimeType(new Uri(url).GetLeftPart(UriPartial.Path));
                 }
 
+                // Need to handle application/octet-stream as well as some providers, especially MENA providers use Blob storage like Azure
+                // and the content type is not set correctly on the blob properties. So you end up with the default application/octet-stream
+                // https://learn.microsoft.com/en-us/rest/api/storageservices/put-blob?tabs=microsoft-entra-id#request-headers-all-blob-types
+                if (contentType.StartsWith("application/octet-stream", StringComparison.OrdinalIgnoreCase))
+                {
+                    var extension = Path.GetExtension(url).ToLowerInvariant();
```
crobibero, line 224: Can we use `MimeTypes.GetMimeType` here instead of creating a new mapping?

mjohl: @crobibero  The danger of stepping into a large code base wanting to bug squash without exploring first. Thanks for pointing out the method. Commit is in.

mjohl:
@crobibero @Shadowghost  Will look back into this at some point, maybe. Sadly no time to waste on this any further.

PR #12270, opened 2024-07-14
Add software tonemap filter support

gnattu:
This adds a fast SIMD optimized software tonemap filter support to the server. This filter currently does not support dolby vision reshaping and only support HDR10 and HLG inputs. With current veryfast libx264 settings we can expect 4K60 transcoding+tone mapping on a modern processor with 8 cores equivalent to Zen3/Golden Cove/Firestorm level performance. 

**Changes**

- Add tonemapx filter support to software transcode pipeline
- Add tonemapx filter support to thumbnail generation pipeline

**Issues**

Depends on ffmpeg pr https://github.com/jellyfin/jellyfin-ffmpeg/pull/407

The perf figures in the original post only reflects the initial implementation. The finalized perf figures are here: https://github.com/jellyfin/jellyfin-ffmpeg/pull/407#issuecomment-2198303178

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -276,6 +276,21 @@ namespace MediaBrowser.Controller.MediaEncoding
                 && _mediaEncoder.SupportsFilter("scale_vt");
         }
 
+        private bool IsSwTonemapAvailable(EncodingJobInfo state, EncodingOptions options)
+        {
+            if (state.VideoStream is null
+                || !options.EnableTonemapping
+                || GetVideoColorBitDepth(state) != 10
+                || !_mediaEncoder.SupportsFilter("tonemapx")
+                || !(string.Equals(state.VideoStream?.ColorTransfer, "smpte2084", StringComparison.OrdinalIgnoreCase) || string.Equals(state.VideoStream?.ColorTransfer, "arib-std-b67", StringComparison.OrdinalIgnoreCase)))
+            {
+                return false;
+            }
+
+            return state.VideoStream.VideoRange == VideoRange.HDR
+                   && state.VideoStream.VideoRangeType is VideoRangeType.HDR10 or VideoRangeType.HLG or VideoRangeType.DOVIWithHDR10 or VideoRangeType.DOVIWithHLG;
+        }
+
         private bool IsHwTonemapAvailable(EncodingJobInfo state, EncodingOptions options)
         {
             if (state.VideoStream is null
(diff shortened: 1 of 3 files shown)


Bond-009 (approved):

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -276,6 +276,21 @@ private bool IsVideoToolboxFullSupported()
                 && _mediaEncoder.SupportsFilter("scale_vt");
         }
 
+        private bool IsSwTonemapAvailable(EncodingJobInfo state, EncodingOptions options)
+        {
+            if (state.VideoStream is null
+                || !options.EnableTonemapping
+                || GetVideoColorBitDepth(state) != 10
+                || !_mediaEncoder.SupportsFilter("tonemapx")
+                || !(string.Equals(state.VideoStream?.ColorTransfer, "smpte2084", StringComparison.OrdinalIgnoreCase) || string.Equals(state.VideoStream?.ColorTransfer, "arib-std-b67", StringComparison.OrdinalIgnoreCase)))
```
nyanmisaka, line 285: Are these checks duplicated?
```
|| !(string.Equals(state.VideoStream?.ColorTransfer, "smpte2084", StringComparison.OrdinalIgnoreCase) || string.Equals(state.VideoStream?.ColorTransfer, "arib-std-b67", StringComparison.OrdinalIgnoreCase)))
```
```
            return state.VideoStream.VideoRange == VideoRange.HDR
                   && state.VideoStream.VideoRangeType is VideoRangeType.HDR10 or VideoRangeType.HLG or VideoRangeType.DOVIWithHDR10 or VideoRangeType.DOVIWithHLG;
```

nyanmisaka (approved):
Speaking of Golden Cove, let's hope that the flagship models in Raptor Lake don't crash in our AVX2 code.

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
gnattu: The former checks for all supported HDR curves and the latter filters out Dolby vision videos without fallback profiles. Only checks for the latter should be enough but we are doing this double check for hardware tone mapping as well.

nyanmisaka: https://github.com/jellyfin/jellyfin/blob/eb45a36af0ff3ef540a4a936ea1329139f0523c1/MediaBrowser.Model/Entities/MediaStream.cs#L703-L755

`ColorTransfer` is normalized to `VideoRange` and `VideoRangeType`. We should eventually avoid using `ColorTransfer` directly.

Shadowghost (approved):

PR #12274, opened 2024-07-14
Add check for ProviderIds to prevent '=' from appearing in keys, also support '=' in the values.

EraYaN:
**Changes**

So this prevents the deserialization code from breaking if there are `=` characters in the key or value of the provider id.
Also adds some documentation to the functions themselves.

This also add a set of `Remove*` functions and made the input checking stricter for the `Set*` functions.

**Issues**

Reference: https://github.com/ShokoAnime/Shokofin/issues/62
Custom ProviderIds were disappearing on a normal (non "replace") scan due to them not being deserialized (quietly)

When this is in, I'll manually backport it to the 10.9 branch but without the breaking behavior changes.

### DIFF
--- a/Emby.Server.Implementations/Data/SqliteItemRepository.cs
+++ b/Emby.Server.Implementations/Data/SqliteItemRepository.cs
@@ -1046,9 +1046,10 @@ namespace Emby.Server.Implementations.Data
             foreach (var part in value.SpanSplit('|'))
             {
                 var providerDelimiterIndex = part.IndexOf('=');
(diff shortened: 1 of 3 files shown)


Bond-009 (approved):

Shadowghost (approved):
do we want to backport this?

crobibero (approved):

EraYaN:
> do we want to backport this?

I'll give it a go somewhere this week. At least the documentation side of things.

PR #12279, opened 2024-07-15
Move external url listing to provider for plugin use

crobibero:
**Changes**

Moves the `GetRelatedUrls()` out from `BaseItem` to a dedicated provider so plugins can add dynamic urls.

See linked plugin PRs for sample implementations.

### DIFF
--- a/Emby.Server.Implementations/ApplicationHost.cs
+++ b/Emby.Server.Implementations/ApplicationHost.cs
@@ -664,7 +664,8 @@ namespace Emby.Server.Implementations
                 GetExports<IMetadataService>(),
                 GetExports<IMetadataProvider>(),
                 GetExports<IMetadataSaver>(),
-                GetExports<IExternalId>());
+                GetExports<IExternalId>(),
+                GetExports<IExternalUrlProvider>());
 
             Resolve<IMediaSourceManager>().AddParts(GetExports<IMediaSourceProvider>());
         }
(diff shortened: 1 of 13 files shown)


Bond-009 (approved):

nielsvanvelzen (approved):
Is there much difference between external id's and external URL's, and would it make sense to merge those two?

crobibero:
> Is there much difference between external id's and external URL's, and would it make sense to merge those two?

If we wanted to simplify this further I could remove the url generation from `IExternalId` since the new interface is more generic (and then Simkl and Trakt don't need to be external id providers)

nielsvanvelzen:
> If we wanted to simplify this further I could remove the url generation from IExternalId since the new interface is more generic (and then Simkl and Trakt don't need to be external id providers)

So that would mean removing the "UrlFormatString" and plugins that currently provide that will add an additional IExternalUrlProvider to replace that part?

crobibero:
> > If we wanted to simplify this further I could remove the url generation from IExternalId since the new interface is more generic (and then Simkl and Trakt don't need to be external id providers)
> 
> So that would mean removing the "UrlFormatString" and plugins that currently provide that will add an additional IExternalUrlProvider to replace that part?

Correct- I currently left that property since that's good bit of work to push through each plugin

nielsvanvelzen (approved):

Shadowghost (approved):

PR #12288, opened 2024-07-16
Don't include new internal MediaStream properties in api spec

crobibero:
https://github.com/jellyfin/jellyfin/pull/12056#pullrequestreview-2177950757

### DIFF
--- a/MediaBrowser.Model/Entities/MediaStream.cs
+++ b/MediaBrowser.Model/Entities/MediaStream.cs
@@ -7,6 +7,7 @@ using System.ComponentModel;
 using System.Globalization;
 using System.Linq;
 using System.Text;
+using System.Text.Json.Serialization;
 using Jellyfin.Data.Enums;
 using Jellyfin.Extensions;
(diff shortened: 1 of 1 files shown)


nielsvanvelzen (approved):
Changes look good but it would be nice if @Arcus92 could confirm the intention of these properties.

Bond-009 (approved):

Arcus92:
Hi @nielsvanvelzen,
To be honest, I wasn't aware of the impact to the API. The original `IsTextSubtitleStream` was also used as a condition whether the subtitle can be extracted from the source. But this was just a coincidence and as soon I added a non-text but extract-able subtitle the property name was misleading. That's why I added `IsExtractableSubtitleStream`.

Since this is only be used in the private extract method, there is no need to make it accessible through the API. Same for `IsPgsSubtitleStream`. **So ignoring both in the JSON is fine from my pov.**

nielsvanvelzen:
Awesome, thanks for confirming @Arcus92!

Issue #12256, opened 2024-07-11
[Issue]: Tracks with multiple artists


HanaO00:
### Please describe your bug

I was using ";" to separate artists on tracks, for example in the "artist" field: Elvis Presley; Celine Dion

But since the last updates it no longer works and it takes it as a single artist

### Reproduction Steps

1. Add multiple artist with ";" to separate them
2. The two are recognized as one artist

### Jellyfin Version

10.9.0

### Environment

```markdown
- OS: Debian 12
- Linux Kernel: ?
- Virtualization: ?
- Clients: PC (jellyfin media player)
- Browser: Firefox
- FFmpeg Version: ?
- Playback Method: ?
- Hardware Acceleration: ?
- GPU Model: ?
- Plugins: ? 
- Reverse Proxy: ?
- Base URL: ?
- Networking: ? 
- Storage: ?
```

### Jellyfin logs

```shell
No logs because not a playback problem
```

### Please attach any screenshots here

![Capture d’écran_2024-07-11_16-23-46](https://github.com/jellyfin/jellyfin/assets/17810985/f2df3b74-f90a-4700-a29d-bafebee6b9f1)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Issue #11871, opened 2024-05-29
[Issue]: Music Artist Didn't Split by Semicolon


hjkcai:
### Please describe your bug

I have a semicolon-separated artist "Adam Lambert;Nile Rodgers;Sam Sparro" in my ID3 metadata. However Jellyfin thinks there is only one artist called "Adam Lambert;Nile Rodgers;Sam Sparro". There should be 3 artists.

<img width="577" alt="image" src="https://github.com/jellyfin/jellyfin/assets/731326/6beb0573-6144-414c-96cb-ae02bccd567c">

The more interesting thing is that if I click "Edit Metadata" on the web client, don't modify anything and click "Save Changes", the artist are correct now.

Here is the screen record:

https://github.com/jellyfin/jellyfin/assets/731326/e44e326e-c8ed-4cdb-be2c-b2b4edac3f8b

### Reproduction Steps

As above

### Jellyfin Version

10.9.0

### if other:

10.9.3

### Environment

```markdown
- OS: macOS 14.3.1
- Clients: Browser
- Browser: Chrome 122
- FFmpeg Version: 6.0.1
```

### Jellyfin logs

```shell
[2024-05-29 10:20:21.411 +08:00] [INF] [31] MediaBrowser.MediaEncoding.Encoder.MediaEncoder: Starting "/Applications/Jellyfin.app/Contents/MacOS/ffprobe" with args "-analyzeduration 200M -probesize 1G -i file:\"/mnt/media/MusicDLs/Adam Lambert;Nile Rodgers;Sam Sparro - Shady.mp3\" -threads 0 -v warning -print_format json -show_streams -show_format"
```

### FFmpeg logs

```shell
{
    "streams": [
        {
            "index": 0,
            "codec_name": "mp3",
            "codec_long_name": "MP3 (MPEG audio layer 3)",
            "codec_type": "audio",
            "codec_tag_string": "[0][0][0][0]",
            "codec_tag": "0x0000",
            "sample_fmt": "fltp",
            "sample_rate": "44100",
            "channels": 2,
            "channel_layout": "stereo",
            "bits_per_sample": 0,
            "initial_padding": 0,
            "r_frame_rate": "0/0",
            "avg_frame_rate": "0/0",
            "time_base": "1/14112000",
            "start_pts": 353600,
            "start_time": "0.025057",
            "duration_ts": 2525552640,
            "duration": "178.964898",
            "bit_rate": "261193",
            "disposition": {
                "default": 0,
[...]

Issue #10794, opened 2024-01-01
[Issue]: Treats artist names with slashes in them as multiple artists


ieure:
### Please describe your bug

I recently added a rip of "[Black Science](https://www.discogs.com/master/122329-Geezer-Black-Science)," an album by "[G//Z/R](https://www.discogs.com/artist/11609-GZR)" (a side project of Geezer Butler).  The relevant FLAC tags for the files in this album are:

```
    comment[7]: ALBUM=Black Science
    comment[8]: ALBUM ARTIST=G//Z/R
    comment[9]: ALBUMARTIST=G//Z/R
    comment[10]: ALBUMARTIST_CREDIT=geezer
    comment[11]: ALBUMARTISTSORT=G//Z/R
    comment[16]: ARTIST=G//Z/R
    comment[17]: ARTIST_CREDIT=geezer
    comment[18]: ARTISTSORT=G//Z/R
    comment[24]: COMPILATION=0
```

The album is held in `/star/library/audio/flac/G__Z_R/Black Science`.

When Jellyfin shows the album in the album list, it says the artist is "G":

![image](https://github.com/jellyfin/jellyfin/assets/35252/3aad925e-9270-4cbe-9d13-a7ef12bd15e4)

In the album detail, it says the artist is "G / R / Z":

![image](https://github.com/jellyfin/jellyfin/assets/35252/5106b40d-25a0-495b-b589-25ba007b55d2)

I believe what JF is doing is treating this like an album by three artists, then sorting those names alphabetically for display, which is incorrect.  FLAC files which credit multiple artists will have multiple `ARTIST` (for multi-artist tracks) or `ALBUMARTIST` (for multi-artist albums) tags.

I'm not sure **why** it's doing this — maybe the double slashes is throwing it off?  Other artists I'd expect to see the same behavior with (ex. "AC/DC") are unaffected.

### Jellyfin Version

Other

### if other:

10.8.13

### Environment

```markdown
- OS: Debian 11
- Linux Kernel: Linux neutron 5.10.0-23-amd64 #1 SMP Debian 5.10.179-1 (2023-05-12) x86_64 GNU/Linux
- Virtualization: Docker
- Clients: Web
- Browser: [LibreWolf](https://librewolf.net)
- FFmpeg Version: n/a
- Playback Method: n/a
- Hardware Acceleration: n/a
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

[...]

PR #12293, opened 2024-07-16
Add env var to always use legacy audio prober

gnattu:
Our current TagLib-based audio prober only processes spec-defined multi-value tags. For non-standard methods, like using a special character as a separator, TagLib simply does not support them. Properly implementing such mechanisms is beyond the scope of version 10.9, as it would introduce API changes for additional user settings. Add an environment variable so users can choose to always use the legacy ffprobe results, which process some separators, when re-tagging the library with the spec defined method is not an option.

**Changes**

**Issues**

Workaround for #11871 #12256 #10730 #10794


gnattu:
Don't backport this to master as we are going to implement the real solution, and even if this is needed for something else, it should be implemented separately as taglib is no longer used in 10.10.

Shadowghost (approved):

crobibero (approved):

gnattu:
Closing as we will not do more 10.9.z releases. Users with this need should check #12385 when 10.10 is out.

PR #12295, opened 2024-07-17
Add TrySetProviderId extension

Bond-009:
Diff is best viewed while ignoring whitespace changes


Shadowghost (approved):

crobibero (approved):
This doesn't actually break the plugin abi, but marking it as such for visibility so I remember to update plugins to use this instead

Issue #11397, opened 2024-04-21
Unstable: Max audio channel setting no honored when audio is 5.0


solidsnake1298:
### Please describe your bug

When playing a 5.0 (not 5.1) channel video in a browser, the max audio channel setting is not honored.  The server will output 7.1 AAC, which is not playable in Chrome.

Media Info:

From Up On Poppy Hill
Container: mkv
Path: /media/extended/Anime Movies/From Up On Poppy Hill/From Up On Poppy Hill.mkv
Size: 7337 MB

Video
Title: 1080p HEVC SDR
Codec: HEVC
AVC: No
Profile: Main
Level: 120
Resolution: 1920x1036
Aspect ratio: 1.85:1
Anamorphic: No
Interlaced: No
Framerate: 23.976025
Bitrate: 11210 kbps
Bit depth: 8 bit
Video range: SDR
Video range type: SDR
Color space: bt709
Color transfer: bt709
Color primaries: bt709
Pixel format: yuv420p
Ref frames: 1

Audio
Title: Surround 5.0 - English - DTS-HD MA - Default
Language: eng
Codec: DTS
AVC: No
Profile: DTS-HD MA
Layout: 5.0
Channels: 5 ch
Sample rate: 48000 Hz
Bit depth: 16 bit
Default: Yes
Forced: No
External: No

Audio
Title: Surround 5.0 - English - DTS
Language: eng
Codec: DTS
AVC: No
Profile: DTS
Layout: 5.0
Channels: 5 ch
Bitrate: 1536 kbps
Sample rate: 48000 Hz
Default: No
Forced: No
External: No

Audio
Title: Surround 5.0 - Japanese - DTS-HD MA
Language: jpn
Codec: DTS
AVC: No
Profile: DTS-HD MA
Layout: 5.0
Channels: 5 ch
Sample rate: 48000 Hz
Bit depth: 16 bit
Default: No
Forced: No
External: No

Audio
Title: Surround 5.0 - Japanese - DTS
Language: jpn
Codec: DTS
AVC: No
Profile: DTS
Layout: 5.0
Channels: 5 ch
Bitrate: 1536 kbps
Sample rate: 48000 Hz
Default: No
Forced: No
External: No

Subtitle
Title: English - PGSSUB
Language: eng
Codec: PGSSUB
AVC: No
Default: No
Forced: No
External: No

Subtitle
Title: English - PGSSUB
Language: eng
Codec: PGSSUB
AVC: No
Default: No
[...]

PR #12300, opened 2024-07-18
Add RFC7845 downmix algorithm

gnattu:
This algorithm is defined in RFC7845 Section 5.1.1.5, supports 2.1, 4.0, 5.0, 5.1, 6.1 and 7.1 channel layouts

Refer: https://datatracker.ietf.org/doc/html/rfc7845#section-5.1.1.5

**Changes**

**Issues**


crobibero:
Is there a friendlier name for the downmix algorithm?

EraYaN:
> Is there a friendlier name for the downmix algorithm?

It's from the Ogg/Opus RFC, but I doubt naming either of those two will be more helpful.

gnattu:
There is no "user friendly name" I'm afraid as the RFC document is not a "user friendly" document anyway. I think just use the name RFC7845 is fine as we are using similar name like bt2390 for tonemap algorithms.

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -2653,26 +2653,121 @@ public string GetAudioFilterParam(EncodingJobInfo state, EncodingOptions encodin
...
+                switch (state.AudioStream.Channels.Value)
                 {
-                    case DownMixStereoAlgorithms.Dave750:
-                        filters.Add("pan=stereo|c0=0.5*c2+0.707*c0+0.707*c4+0.5*c3|c1=0.5*c2+0.707*c1+0.707*c5+0.5*c3");
+                    case 3:
+                    {
+                        switch (encodingOptions.DownMixStereoAlgorithm)
+                        {
+                            case DownMixStereoAlgorithms.Rfc7845:
+                                filters.Add("pan=stereo|FL=0.414214*FC+0.585786*FL|FR =0.414214*FC+0.585786*FR");
```
crobibero, line 2670: Don't know if it matters but there is extra whitespace in the filter

crobibero (approved):

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
gnattu: Should not affect functionality but removed for consistency

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -2653,26 +2653,121 @@ public string GetAudioFilterParam(EncodingJobInfo state, EncodingOptions encodin
...
                 {
                     filters.Add("volume=" + encodingOptions.DownMixAudioBoost.ToString(CultureInfo.InvariantCulture));
                 }
 
-                switch (encodingOptions.DownMixStereoAlgorithm)
+                switch (state.AudioStream.Channels.Value)
                 {
-                    case DownMixStereoAlgorithms.Dave750:
-                        filters.Add("pan=stereo|c0=0.5*c2+0.707*c0+0.707*c4+0.5*c3|c1=0.5*c2+0.707*c1+0.707*c5+0.5*c3");
+                    case 3:
```
cvium, line 2665: Just thinking out loud here, but maybe this would be preferable to nested switches. Thoughts?
```suggestion
                    case 3 when encodingOptions.DownMixStereoAlgorithm == ownMixStereoAlgorithms.Rfc7845:
```

gnattu: I plan to add more algorithms later and will modify this in that one.

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -2653,26 +2653,121 @@ public string GetAudioFilterParam(EncodingJobInfo state, EncodingOptions encodin
...
+                            case DownMixStereoAlgorithms.Rfc7845:
+                                filters.Add("pan=stereo|c0=0.374107*c2+0.529067*c0+0.458186*c4+0.264534*c5+0.374107*c3|c1=0.374107*c2+0.529067*c1+0.458186*c5+0.264534*c4+0.374107*c3");
+                                break;
+                            case DownMixStereoAlgorithms.None:
+                            default:
+                                break;
+                        }
+
+                        break;
+                    }
```
cvium, line 2735: A longer example here. It reduces the nesting but I can't tell if readability suffers slightly.
```suggestion
                    case 6 when encodingOptions.DownMixStereoAlgorithm == DownMixStereoAlgorithms.Dave750:
                        filters.Add("pan=stereo|c0=0.5*c2+0.707*c0+0.707*c4+0.5*c3|c1=0.5*c2+0.707*c1+0.707*c5+0.5*c3");
                        break;
                    case 6 when encodingOptions.DownMixStereoAlgorithm == DownMixStereoAlgorithms.NightmodeDialogue:
                        filters.Add("pan=stereo|c0=c2+0.30*c0+0.30*c4|c1=c2+0.30*c1+0.30*c5");
                        break;
                    case 6 when encodingOptions.DownMixStereoAlgorithm == DownMixStereoAlgorithms.Rfc7845:
                        filters.Add("pan=stereo|c0=0.374107*c2+0.529067*c0+0.458186*c4+0.264534*c5+0.374107*c3|c1=0.374107*c2+0.529067*c1+0.458186*c5+0.264534*c4+0.374107*c3");
                        break;
                    default:
                        break;
```

gnattu:
I refactored the downmix filter related code and split the downmix filter related helpers into a dedicated static class. This makes the algorithms list more extensible and allows for more precise checks to determine if the current settings should fall back to ffmpeg's internal downmix system.

MediaBrowser.Controller/MediaEncoding/DownMixAlgorithmsHelper.cs
```diff
@@ -0,0 +1,61 @@
...
+        var inferredLayout = audioStream.Channels.Value switch
+        {
+            1 => "mono",
+            2 => "stereo",
+            3 => "2.1", // Could also be 3.0, prefer 2.1
+            4 => "quad", // Could also be 4.0 (with front center and rear center) and 3.1 with LFE. prefer quad with back left and back right
+            5 => "5.0",
+            6 => "5.1", // Could also be 6.0 or hexagonal, prefer 5.1
+            7 => "6.1", // Could also be 7.0, prefer 6.1
+            8 => "7.1", // Could also be 8.0, prefer 6.1
```
crobibero, line 56:
```suggestion
            8 => "7.1", // Could also be 8.0, prefer 7.1
```

Jellyfin.Api/Controllers/DynamicHlsController.cs
```diff
@@ -1733,7 +1733,7 @@ private string GetAudioArguments(StreamState state)
 
         var channels = state.OutputAudioChannels;
 
-        var useDownMixAlgorithm = state.AudioStream.Channels is 6 && _encodingOptions.DownMixStereoAlgorithm != DownMixStereoAlgorithms.None;
+        var useDownMixAlgorithm = DownMixAlgorithmsHelper.AlgorithmFilterStrings.TryGetValue((_encodingOptions.DownMixStereoAlgorithm, DownMixAlgorithmsHelper.InferChannelLayout(state.AudioStream)), out _);
```
crobibero, line 1736:
```suggestion
        var useDownMixAlgorithm = DownMixAlgorithmsHelper.AlgorithmFilterStrings.ContainsKey((_encodingOptions.DownMixStereoAlgorithm, DownMixAlgorithmsHelper.InferChannelLayout(state.AudioStream)));
```

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -7008,7 +6997,10 @@ public string GetProgressiveVideoAudioArguments(EncodingJobInfo state, EncodingO
 
             var channels = state.OutputAudioChannels;
 
-            if (channels.HasValue && ((channels.Value != 2 && state.AudioStream?.Channels != 6) || encodingOptions.DownMixStereoAlgorithm == DownMixStereoAlgorithms.None))
+            var useDownMixAlgorithm = state.AudioStream is not null
+                                      && DownMixAlgorithmsHelper.AlgorithmFilterStrings.TryGetValue((encodingOptions.DownMixStereoAlgorithm, DownMixAlgorithmsHelper.InferChannelLayout(state.AudioStream)), out _);
```
crobibero, line 7001:
```suggestion
                                      && DownMixAlgorithmsHelper.AlgorithmFilterStrings.ContainsKey((encodingOptions.DownMixStereoAlgorithm, DownMixAlgorithmsHelper.InferChannelLayout(state.AudioStream)));
```

MediaBrowser.Controller/MediaEncoding/DownMixAlgorithmsHelper.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        // When we don't have definitive channel layout, we have to guess from the channel count
+        // Guessing is not always correct, but for most videos we don't have to guess like this as the definitive layout is recorded during scan
+        var inferredLayout = audioStream.Channels.Value switch
+        {
+            1 => "mono",
+            2 => "stereo",
+            3 => "2.1", // Could also be 3.0, prefer 2.1
+            4 => "quad", // Could also be 4.0 (with front center and rear center) and 3.1 with LFE. prefer quad with back left and back right
+            5 => "5.0",
```
Bond-009, line 53: Why not prefer 4.1?

MediaBrowser.Controller/MediaEncoding/DownMixAlgorithmsHelper.cs
```diff
@@ -0,0 +1,61 @@
...
+
+    /// <summary>
+    /// Get the audio channel layout string from the audio stream
+    /// If the input audio string does not have a valid layout string, guess from channel count.
+    /// </summary>
+    /// <param name="audioStream">The audio stream to get layout.</param>
+    /// <returns>Channel Layout string.</returns>
+    public static string InferChannelLayout(MediaStream audioStream)
+    {
+        if (!string.IsNullOrWhiteSpace(audioStream.ChannelLayout))
```
Bond-009, line 35: `ChannelLayout` should be normalized before being used like this.
For BDMVs it will always be "$CHANNELS.$LFE" and never "stereo" or "quad"

Bond-009 (changes requested):

MediaBrowser.Controller/MediaEncoding/DownMixAlgorithmsHelper.cs
gnattu: > For BDMVs it will always be "$CHANNELS.$LFE" and never "stereo" or "quad"

ffmpeg never outputs 2.0 as "2.0", it will always output stereo

gnattu: And "quad" and "4.0" are actually two different channel layouts. 4.0 means front left, front right ,front center and back center, but quad means front left, front right, back left and back right. All names have their own layout so it is not a normalization problem, it is a correctness problem.

gnattu: If our BDMV parsing code has such normalization, we should revert them. I hope we don't.

Bond-009: > ffmpeg never outputs 2.0 as "2.0", it will always output stereo

That's bc we don't get that info from ffmpeg for BDMVs, we get it from the playlist info

MediaBrowser.Controller/MediaEncoding/DownMixAlgorithmsHelper.cs
gnattu: Because 5.0 is more common? But this is purely subjective and such guessing could be wrong anyway.

Bond-009: I've personally never seen 5.0, but have seen 4.1
https://www.blu-ray.com/movies/Alien-4K-Blu-ray/230308/

MediaBrowser.Controller/MediaEncoding/DownMixAlgorithmsHelper.cs
gnattu: Just looked at the BDMV parser, now we have the problem of unable to identify the correct layout due to lack of information. I personally *want* to cowardly reject such ambiguous layouts and let ffmpeg's internal downmixing system to handle it as ffmpeg knows the layout better than us. The good news is that, we don't need to worry about the stereo case and the quad case is relatively rare compared to more common 5.1 or 7.1.

MediaBrowser.Controller/MediaEncoding/DownMixAlgorithmsHelper.cs
gnattu: We just fixed an issued related to 5.0 in 10.9: https://github.com/jellyfin/jellyfin/issues/11397

MediaBrowser.Controller/MediaEncoding/DownMixAlgorithmsHelper.cs
gnattu: I think we are safe for this case after investigation the current BD implementations. Although dolby digital technically supports 2/2 quad configuration but it is extremely rare and that layout is usually seen on legacy DVDs and music albums. So the 4.0 for BDs would almost always be the 4.0 in ffmpeg definition. I think we can just add a comment to the code just in case there is really an exception.

MediaBrowser.Controller/MediaEncoding/DownMixAlgorithmsHelper.cs
gnattu: Search with [4.1 channel resulted in 37 matches](https://www.blu-ray.com/movies/search.php?keyword=&studioid=&videocodec=&disc=&yearfrom=&yearto=&regioncoding=&aspectratio=&aspectratio_original=&releaseyear=&synopsis=&retailerexclusive=&mpaa=&runtimemin=&runtimemax=&videoresolutionid=&sourceid=&subtitles=&audio=4.1&upc=&ean=&asin=&casingid=&slipcoverfront=&slipcoverback=&submit=Search&action=search) and [5.0 resulted in 303 matches](https://www.blu-ray.com/movies/search.php?keyword=&studioid=&videocodec=&disc=&yearfrom=&yearto=&regioncoding=&aspectratio=&aspectratio_original=&releaseyear=&synopsis=&retailerexclusive=&mpaa=&runtimemin=&runtimemax=&videoresolutionid=&sourceid=&subtitles=&audio=5.0&upc=&ean=&asin=&casingid=&slipcoverfront=&slipcoverback=&submit=Search&action=search). I think we should stay with 5.0.

MediaBrowser.Controller/MediaEncoding/DownMixAlgorithmsHelper.cs
gnattu: The RFC7845 explicitly requires the quad (2 front 2 rear) layout to use that downmix algorithm, so for 4 channel BDMVs, we will always use ffmpeg's internal system which will be safe enough.

Bond-009 (approved):
It isn't perfect, but it should be an improvement :+1:

PR #12303, opened 2024-07-19
feat: Added user creation when no user is found

rizqtalal:
**Changes**

**Issues**


gnattu:
Won't this create arbitrary user when the user is not in the database? I don't think this is a correct behavior for the authentication endpoint. What do you want to archive?

JPVenson (changes requested):
this is a security nightmare

PR #12319, opened 2024-07-21
Check MaxAudioChannels for directAudioStream candidates

gnattu:
The current stream builder logic does not check the channel limit when determining if the audio stream can be directly used, and this can cause some undesired effects:

- A high channel count surround sound stream might be picked even if a stereo one exists when the user requires stereo audio.

- The user's preferred audio codec might not be respected during the downmix because the requested codec is now forced to be the same as the original source.

**Changes**

- Check TranscodingMaxAudioChannels for directAudioStream candidates
- Update test cases to reflect this change

**Issues**

### DIFF
--- a/MediaBrowser.Model/Dlna/StreamBuilder.cs
+++ b/MediaBrowser.Model/Dlna/StreamBuilder.cs
@@ -908,12 +908,11 @@ namespace MediaBrowser.Model.Dlna
                 }
             }
 
-            var audioStreamWithSupportedCodec = candidateAudioStreams.Where(stream => ContainerProfile.ContainsContainer(audioCodecs, stream.Codec)).ToList();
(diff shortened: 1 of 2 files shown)


Shadowghost (approved):

Bond-009:
> A high channel count surround sound stream might be picked even if a stereo one exists when the user requires stereo audio.

I think this is intended, main reason being that blu-rays always have surround audio and occasionally stereo directors commentary. You wouldn't want to pick the directors commentary by default. If you want to add this behaviors please put it behind an option.

gnattu:
I updated to match the old stream selection behavior, the only difference now is that if the selected stream exceeds channel limit it will not be marked as direct playable. The unit tests are passing so I assume the stream selection is working as intended.

Bond-009 (approved):

crobibero (approved):

PR #12327, opened 2024-07-23
Fix up getting livetv programs api docs

crobibero:
Functionally no different on the API side, but the api spec doesn't correctly reflect that the property is optional.

Requested here: https://github.com/jellyfin/jellyfin-sdk-kotlin/pull/960#issuecomment-2176198249


Shadowghost (approved):

Bond-009 (approved):

nielsvanvelzen (changes requested):
These should also be optional or have a default value in the OpenAPI spec:
- EnableTotalRecordCount
- LibrarySeriesId

nielsvanvelzen (approved):

PR #10530, opened 2023-11-04
Feature: Media Segments

endrl:
I want to propose my take for the "Intro Outro" feature. This is considered a basic implementation that can be advanced.

Meta Ref: https://github.com/jellyfin/jellyfin-meta/discussions/30

### Thoughts
* Chapter integration
  * They are still in the wrong db. Is there a code example how i can migrate them to ef core?
  * How should the final "MediaSegment" look like?
* MediaSegmentAction
  * Usually it will be set  to "Auto" where user settings will eval the final action.
  * A MediaSegment provider (plugin) can also change this through the config page. This might be interesting when you use youtube or other video sources where the default behaviour for "all" content is not desired.
  * As the REST api allow edits you can manage this also on your own for each single entry (media management enthusiasts?)
* ItemId
  * Is also missing in db, migration required

### Testing
To explore the ecosystem i created some implementations. A docker compose [file](https://github.com/endrl/segment-editor#installation) is available.

### Related projects

* Jellyfin Plugin: [Media Analyzer](https://github.com/endrl/jellyfin-plugin-media-analyzer)
* Jellyfin Plugin: [.EDL Creator](https://github.com/endrl/jellyfin-plugin-edl)
* Tool: [Jellyfin Segment Editor](https://github.com/endrl/segment-editor)
* Player: [Jellyfin Vue Fork](https://github.com/endrl/jellyfin-vue)

Happy testing! Feedback welcome.

PR #12345, opened 2024-07-26
Add media segments API

JPVenson:
Implements media segments api

**Issues**
https://github.com/jellyfin/jellyfin-meta/discussions/30#discussioncomment-9629257

supersedes https://github.com/jellyfin/jellyfin/pull/10530

Please Squash


Jellyfin.Api/Controllers/MediaSegmentController.cs
```diff
@@ -0,0 +1,58 @@
...
+
+    /// <summary>
+    /// Gets all media segments based on an itemId.
+    /// </summary>
+    /// <param name="itemId">The ItemId.</param>
+    /// <returns>A list of media segement objects related to the requested itemId.</returns>
+    [HttpGet("MediaSegments/{itemId}")]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    public ActionResult<IAsyncEnumerable<MediaSegment>> GetSegmentsAsync([FromRoute, Required] Guid itemId)
```
crobibero, line 47: We should return an api-specific model instead of reusing the db model

Jellyfin.Data/Entities/MediaSegment.cs
```diff
@@ -0,0 +1,52 @@
...
+///     An entity representing the metadata for a group of trickplay tiles.
+/// </summary>
+public class MediaSegment
+{
+    /// <summary>
+    ///     Gets or sets the id of the media segment.
+    /// </summary>
+    /// <remarks>
+    /// Required.
+    /// </remarks>
```
crobibero, line 16:
```suggestion
```

These remarks are unneeded

Jellyfin.Data/Entities/MediaSegmentType.cs
```diff
@@ -0,0 +1,37 @@
+namespace Jellyfin.Data.Entities;
+
+/// <summary>
+///     Defines the types of content a individial <see cref="MediaSegment"/> represents.
+/// </summary>
+public enum MediaSegmentType
```
crobibero, line 6: Enum values should have their int representation set

Jellyfin.Server.Implementations/MediaSegments/MediaSegementManager.cs
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /// <inheritdoc />
+    public async IAsyncEnumerable<MediaSegment> GetSegmentsAsync(Guid itemId)
+    {
+        using var db = await _dbProvider.CreateDbContextAsync().ConfigureAwait(false);
+        await foreach (var segment in db.MediaSegments
+            .Where(e => e.ItemId.Equals(itemId))
+            .OrderBy(e => e.StartTick)
+            .AsAsyncEnumerable())
```
crobibero, line 56: I don't think an AsyncEnumerable has any real benefit here since the list is going to be < 10 items?

Jellyfin.Server.Implementations/MediaSegments/MediaSegementManager.cs
```diff
@@ -0,0 +1,68 @@
...
+        _dbProvider = dbProvider;
+    }
+
+    /// <inheritdoc />
+    public async Task<MediaSegment> CreateSegmentAsync(MediaSegment mediaSegment)
+    {
+        if (mediaSegment.EndTick < mediaSegment.StartTick)
+        {
+            throw new InvalidOperationException($"A segments {nameof(MediaSegment.EndTick)} cannot be before its {nameof(MediaSegment.StartTick)}");
+        }
```
crobibero, line 33:
```suggestion
        ArgumentOutOfRangeException.ThrowIfLessThan(mediaSegment.EndTick, mediaSegment.StartTick);
```

Might as well use the new features

Jellyfin.Data/Entities/MediaSegment.cs
JPVenson: Note: might as well remove them on the trickplay comments too

Jellyfin.Server.Implementations/MediaSegments/MediaSegementManager.cs
JPVenson: i think it can be way more but thats not really the point as using IAsyncEnumerable allows consumers to be most flexiable in their way to work with media segments.

MediaBrowser.Model/MediaSegments/MediaSegmentModel.cs
```diff
@@ -0,0 +1,34 @@
...
+
+    /// <summary>
+    ///     Gets or sets the end of the segment.
+    /// </summary>
+    public int EndTick { get; set; }
+
+    /// <summary>
+    ///     Gets or sets the Type of content this segment defines.
+    /// </summary>
+    public MediaSegmentTypeModel Type { get; set; }
```
nielsvanvelzen, line 33: Bit of a nitpick but I would specify type after item id, doesn't matter for behavior at all though

Jellyfin.Api/Controllers/MediaSegmentController.cs
```diff
@@ -0,0 +1,54 @@
...
+        _mediaSegmentManager = mediaSegmentManager;
+        _libraryManager = libraryManager;
+    }
+
+    /// <summary>
+    /// Gets all media segments based on an itemId.
+    /// </summary>
+    /// <param name="itemId">The ItemId.</param>
+    /// <returns>A list of media segement objects related to the requested itemId.</returns>
+    [HttpGet("MediaSegments/{itemId}")]
```
nielsvanvelzen, line 40:
```suggestion
    [HttpGet("{itemId}")]
```
To get `/MediaSegment/{itemId}` as path, otherwise the URL would become `/MediaSegment/MediaSegments/{itemId}`.

I would also change the controller name to `MediaSegmentsController`

Jellyfin.Server.Implementations/MediaSegments/MediaSegementManager.cs
```diff
@@ -0,0 +1,90 @@
...
+using MediaBrowser.Controller;
+using MediaBrowser.Model.MediaSegments;
+using Microsoft.EntityFrameworkCore;
+
+namespace Jellyfin.Server.Implementations;
+
+/// <summary>
+///     Manages media segments retrival and storage.
+/// </summary>
+public class MediaSegementManager : IMediaSegmentManager
```
nielsvanvelzen, line 15:
```suggestion
public class MediaSegmentManager : IMediaSegmentManager
```

MediaBrowser.Controller/Entities/BaseItem.cs
```diff
@@ -1116,7 +1118,8 @@ private MediaSourceInfo GetVersionInfo(bool enablePathSubstitution, BaseItem ite
                 RunTimeTicks = item.RunTimeTicks,
                 Container = item.Container,
                 Size = item.Size,
-                Type = type
+                Type = type,
+                HasSegments = item.MediaType is MediaType.Video && (protocol is null or MediaProtocol.File) && MediaSegmentManager.HasSegments(item.Id)
```
nielsvanvelzen, line 1122: media segment could work for audio types too, e.g. podcasts/audiobooks. Perhaps it makes sense to move the MediaType check to MediaSegmentManager so we can manage the "supported types" over there

Jellyfin.Api/Controllers/MediaSegmentController.cs
```diff
@@ -0,0 +1,54 @@
...
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    public ActionResult<IAsyncEnumerable<MediaSegmentModel>> GetSegmentsAsync([FromRoute, Required] Guid itemId)
```
nielsvanvelzen, line 43: Didn't think of this in the meta issue but it would be good to add an (optional) filter `includeSegmentTypes` that, if specified, filters to only the requested segment types. Otherwise return all.

Jellyfin.Api/Controllers/MediaSegmentController.cs
```diff
@@ -0,0 +1,54 @@
...
+    {
+        var item = _libraryManager.GetItemById<BaseItem>(itemId, User.GetUserId());
+        if (item is null)
```
nielsvanvelzen, line 46: ^ does this part validate item access? (not sure how library manager works)

Jellyfin.Api/Controllers/MediaSegmentController.cs
```diff
@@ -0,0 +1,54 @@
...
+        {
+            return NotFound();
+        }
+
+        var items = _mediaSegmentManager.GetSegmentsAsync(item.Id);
+        return Ok(items);
```
nielsvanvelzen, line 52: I would recommend using an object in the response with a single "Segments" property that contains the result. That way, if we ever want to add more data to the response we can without having a breaking API change.

We already have the QueryResult that could potentially be re-used for that

Jellyfin.Data/Entities/MediaSegment.cs
```diff
@@ -0,0 +1,37 @@
...
+    ///     Gets or sets the id of the media segment.
+    /// </summary>
+    [JsonIgnore]
+    public Guid Id { get; set; }
+
+    /// <summary>
+    ///     Gets or sets the id of the associated item.
+    /// </summary>
+    [JsonIgnore]
+    public Guid ItemId { get; set; }
```
nielsvanvelzen, line 21: Those are no longer needed now that you have a separate DTO right?

```suggestion
    public Guid Id { get; set; }

    /// <summary>
    ///     Gets or sets the id of the associated item.
    /// </summary>
    public Guid ItemId { get; set; }
```

nielsvanvelzen (changes requested):

Jellyfin.Api/Controllers/MediaSegmentController.cs
JPVenson: hmm what would be the usecase for that?

Jellyfin.Api/Controllers/MediaSegmentController.cs
JPVenson: Yes its done like that for Trickplay to validate. If i can retrive the item with that userId, i can see it-

Jellyfin.Api/Controllers/MediaSegmentController.cs
nielsvanvelzen: - A client may not support all segment types, for example by not having an implementation (e.g. the annotation type from the possible extensions)
- A client may want to list the segment for an item and have a filter for the user to only show specific types

Jellyfin.Api/Controllers/MediaSegmentsController.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        _mediaSegmentManager = mediaSegmentManager;
+        _libraryManager = libraryManager;
+    }
+
+    /// <summary>
+    /// Gets all media segments based on an itemId.
+    /// </summary>
+    /// <param name="itemId">The ItemId.</param>
+    /// <param name="includeSegmentTypes">Optional filter of requested segmeent types.</param>
```
nielsvanvelzen, line 43:
```suggestion
    /// <param name="includeSegmentTypes">Optional filter of requested segment types.</param>
```

Jellyfin.Api/Controllers/MediaSegmentsController.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// <returns>A list of media segement objects related to the requested itemId.</returns>
```
nielsvanvelzen, line 44:
```suggestion
    /// <returns>A list of media segment objects related to the requested itemId.</returns>
```

MediaBrowser.Model/MediaSegments/MediaSegmentDto.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// </summary>
+    public Guid Id { get; set; }
+
+    /// <summary>
+    ///     Gets or sets the id of the associated item.
+    /// </summary>
+    public Guid ItemId { get; set; }
+
+    /// <summary>
+    ///     Gets or sets the Type of content this segment defines.
```
nielsvanvelzen, line 22:
```suggestion
    ///     Gets or sets the type of content this segment defines.
```

Jellyfin.Data/Entities/MediaSegmentType.cs
```diff
@@ -0,0 +1,37 @@
+namespace Jellyfin.Data.Entities;
+
+/// <summary>
+///     Defines the types of content a individial <see cref="MediaSegment"/> represents.
```
nielsvanvelzen, line 4:
```suggestion
///     Defines the types of content an individual <see cref="MediaSegment"/> represents.
```

Jellyfin.Data/Entities/MediaSegmentType.cs
```diff
@@ -0,0 +1,37 @@
+namespace Jellyfin.Data.Entities;
```
nielsvanvelzen, line 1: I think this should be in Jellyfin.Data.Enums.

nielsvanvelzen (approved):
Approved from an API perspective. Few typos to fix.

MediaBrowser.Controller/MediaSegements/IMediaSegmentManager.cs
RePhilipp, line 1: How about cancellation token support so plugin creators can already make use if they want to.
CancellationToken cancellationToken = default) would be the best of both worlds i think.

crobibero: Where do you think a cancellation token fits into the functions of this interface?

JPVenson: in general i would agree, but that feature isn't used anywhere else so until the decision comes to implement that for all other parts of the code i would keep it very simple here.

JPVenson: @crobibero the common case is to get a cancelation token of the request from the api

```csharp
    public async Task<ActionResult<QueryResult<MediaSegmentDto>>> GetSegmentsAsync(
        [FromRoute, Required] Guid itemId,
        [FromQuery] IEnumerable<MediaSegmentType>? includeSegmentTypes = null,
        CancellationToken requestToken = default)
```

and hand it all the way down to EF to allow query to be aborted if the connection is closed

```csharp

    public async Task<IEnumerable<MediaSegmentDto>> GetSegmentsAsync(Guid itemId, IEnumerable<MediaSegmentType> typeFilter, CancellationToken stopRequest = default)
    {
        using var db = await _dbProvider.CreateDbContextAsync(stopRequest).ConfigureAwait(false);

        return (await db.MediaSegments
            .Where(e => e.ItemId.Equals(itemId))
            .Where(e => typeFilter.Contains(e.Type))
            .OrderBy(e => e.StartTicks)
            .ToArrayAsync(stopRequest)
            .ConfigureAwait(false))
            .Select(Map);
    }
```

This helps to clean up a lot of junk if connections are aborted when there are long running queries happening and some other advantages. But as i said, that should be done, everywhere not just here and jumps the scope of the PR

crobibero: I understand what cancellation tokens are for- I was asking specifically about which function a plugin would use and benefit from having it

Jellyfin.Api/Controllers/MediaSegmentsController.cs
```diff
@@ -0,0 +1,61 @@
...
+        [FromRoute, Required] Guid itemId,
+        [FromQuery] IEnumerable<MediaSegmentType>? includeSegmentTypes = null)
+    {
+        var item = _libraryManager.GetItemById<BaseItem>(itemId, User.GetUserId());
+        if (item is null)
+        {
+            return NotFound();
+        }
+
+        var items = await _mediaSegmentManager.GetSegmentsAsync(item.Id, includeSegmentTypes ?? Enum.GetValues<MediaSegmentType>()).ConfigureAwait(false);
```
crobibero, line 58: I would prefer just passing `null` instead of `Enum.GetValues`. The query can just be updated to ignore the parameter if `null` is sent

Jellyfin.Api/Controllers/MediaSegmentsController.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// </summary>
+    /// <param name="itemId">The ItemId.</param>
+    /// <param name="includeSegmentTypes">Optional filter of requested segment types.</param>
+    /// <returns>A list of media segment objects related to the requested itemId.</returns>
+    [HttpGet("{itemId}")]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    public async Task<ActionResult<QueryResult<MediaSegmentDto>>> GetSegmentsAsync(
+        [FromRoute, Required] Guid itemId,
+        [FromQuery] IEnumerable<MediaSegmentType>? includeSegmentTypes = null)
```
crobibero, line 50:
```suggestion
        [FromQuery] MediaSegmentType[]? includeSegmentTypes = null)
```

I don't see a benefit for using IEnumerable here

Jellyfin.Server.Implementations/MediaSegments/MediaSegmentManager.cs
```diff
@@ -0,0 +1,99 @@
...
+        _dbProvider = dbProvider;
+    }
+
+    /// <inheritdoc />
+    public async Task<MediaSegmentDto> CreateSegmentAsync(MediaSegmentDto mediaSegment)
+    {
+        ArgumentOutOfRangeException.ThrowIfLessThan(mediaSegment.EndTicks, mediaSegment.StartTicks);
+
+        using var db = await _dbProvider.CreateDbContextAsync().ConfigureAwait(false);
+        mediaSegment.Id = Guid.NewGuid();
```
crobibero, line 37: instead of doing `NewGuid` here, we should utilize `[DatabaseGenerated(DatabaseGeneratedOption.Identity)]` like our other models

Jellyfin.Api/Controllers/MediaSegmentsController.cs
JPVenson: Usually its considered best practice for webAPI as thats what the interface is good for.

Jellyfin.Server.Implementations/MediaSegments/MediaSegmentManager.cs
JPVenson: in that case you might want to look into trickplay too as it does not define that too.

Jellyfin.Data/Entities/MediaSegment.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    /// <summary>
+    ///     Gets or sets the id of the media segment.
+    /// </summary>
+    public Guid Id { get; set; }
+
+    /// <summary>
+    ///     Gets or sets the id of the associated item.
+    /// </summary>
+    public Guid ItemId { get; set; }
```
Shadowghost, line 21: To make it more obvious we might want to rename this to `MediaSourceId`, wdyt?

Shadowghost (commented):
`SegmentProvider` interface for plugins and its integration is missing

JPVenson:
> `SegmentProvider` interface for plugins and its integration is missing

Not part of this PR. Will follow later.

Jellyfin.Data/Entities/MediaSegment.cs
JPVenson: i named that in line with other entities "itemid" like trickplay. I would keep it in line with that or refactor the other existing ones too.

Shadowghost: IMHO `ItemId` migth be misleading. Is trickplay even media source based (if not, this should be changed...)?

Jellyfin.Server.Implementations/MediaSegments/MediaSegmentManager.cs
```diff
@@ -0,0 +1,104 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using Jellyfin.Data.Entities;
+using Jellyfin.Data.Enums;
+using MediaBrowser.Controller;
+using MediaBrowser.Controller.Entities;
+using MediaBrowser.Model.MediaSegments;
+using Microsoft.EntityFrameworkCore;
+
+namespace Jellyfin.Server.Implementations;
```
crobibero, line 13:
```suggestion
namespace Jellyfin.Server.Implementations.MediaSegments;
```

Jellyfin.Server.Implementations/MediaSegments/MediaSegmentManager.cs
```diff
@@ -0,0 +1,104 @@
...
+            .Where(e => e.ItemId.Equals(itemId));
+
+        if (typeFilter is not null)
+        {
+            query = query.Where(e => typeFilter.Contains(e.Type));
+        }
+
+        return query
+            .OrderBy(e => e.StartTicks)
+            .ToImmutableList()
```
crobibero, line 64: Why `ToImmutableList`?

JPVenson: List because the original size of the source is unknown and there is a small performance and memory advantage if you use the Immutable variant if there is no need for modifications which isn't here as the result will be discarded as soon as the mapping happens

crobibero (approved):

JPVenson:
part 2 is in #12359

crobibero (approved):

Shadowghost (approved):

Jellyfin.Server.Implementations/MediaSegments/MediaSegmentManager.cs
```diff
@@ -0,0 +1,105 @@
...
+        using var db = await _dbProvider.CreateDbContextAsync().ConfigureAwait(false);
+        await db.MediaSegments.Where(e => e.Id.Equals(segmentId)).ExecuteDeleteAsync().ConfigureAwait(false);
+    }
+
+    /// <inheritdoc />
+    public async Task<IEnumerable<MediaSegmentDto>> GetSegmentsAsync(Guid itemId, IEnumerable<MediaSegmentType>? typeFilter)
+    {
+        using var db = await _dbProvider.CreateDbContextAsync().ConfigureAwait(false);
+
+        var query = db.MediaSegments
```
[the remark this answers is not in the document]
JPVenson: Well we _could_ do that but as the context gets disposed pretty much immediatly after there should not be _that_ big of an impact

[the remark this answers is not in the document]
EraYaN: `AsNoTracking` can also just speed up the queries themself.

Bond-009 (approved):

JPVenson:
https://github.com/jellyfin/jellyfin/pull/12392

PR #12348, opened 2024-07-28
Implement VSCode task for openapi generation

### LINKED ISSUE #12347: VSCode tasks.json ease of use additions
For development purposes, I have a suggestion of updating the VSCode [tasks.json](https://github.com/jellyfin/jellyfin/blob/master/.vscode/tasks.json) file to have a task that can generate an updated `openapi.json` that can be used by developers with to generate an updated typescript API using the  [jellyfin-sdk-typescript](https://github.com/jellyfin/jellyfin-sdk-typescript) repo.

Currently there is not many documented steps for how a contributor can make an openapi change in the `jellyfin` server repo and then use that change in the `jellyfin-web` repo without first waiting an update to the jellyfin SDK package and then an update to the `jellyfin-web` package.json to use that updated package.

Looking through the three repos of `jellyfin`, `jellyfin-sdk-typescript` and `jellyfin-web` I found that the process involves
1. Running the `Jellyfin.Server.Integration.Tests` test from the `jellyfin` server repo
2. Copying the `openapi.json` that the test outputs in the `<jellyfin-server-repo-folder>//tests/Jellyfin.Server.Integration.Tests/bin/Release/net8.0/` to the `<jellyfin-sdk-typescript-repo-folder>`
3. In the `jellyfin-sdk-typescript` repo to update the generated client code, the `build:generated-client` and `build:sdk` task need to run using npm or yarn.
4. Creating a package .tar.gz for the jellyfin sdk using the `npm pack` or `yarn pack` command.
5. From the `jellyfin-web` repo root the package.json needs to be updated to use the locally built package using the `npm install --save <path-to-local-jellyfin-sdk-pkg>` or `yarn add file:<path-to-local-jellyfin-sdk-pkg>`.

At this point the local  `jellyfin-web` client project can now run against the local `jellyfin` server project openapi modifications.
To make it easier for VS Code users to go through this process, I propose adding an `openapi-generation` command that can perform steps 1 and 2 above to generate an `openapi.json` file using the current code and to copy it to a developer's local `jellyfin-sdk-typescript` repo.

Below is a video showing how the task would run in VS Code

https://github.com/user-attachments/assets/d801f6b2-fd1a-4b1c-8b46-8602278cd0b7

ItsAllAboutTheCode:
Adds a Visual Studio Code task to generate the openapi.json file for the Jellyfin typescript API.
The `dotnet swagger` tool is used to generate the API.

Furthermore expanded the build task to support both Release and Debug configurations.

Finally added a launch task that can launch the jellyfin server without the need to run through the debugger.

Side Note: The configurations inside of the `launch.json` file, will all start the debugger even when using the "Run without debugging" menu option VS Code due to this ticket: https://github.com/microsoft/vscode-cpptools/issues/1201

resolves #12347


.vscode/tasks.json
```diff
@@ -20,6 +99,81 @@
                 "${workspaceFolder}/tests/Jellyfin.Api.Tests/Jellyfin.Api.Tests.csproj"
             ],
             "problemMatcher": "$msCompile"
+        },
+        {
+            "label": "openapi test",
```
JPVenson, line 104: i am not happy with this. Referencing a particular test just to get the openapi.json sounds like way to many moving parts.
I would recommend you use the openapitools dotnet tool instead
https://learn.microsoft.com/en-us/aspnet/core/web-api/microsoft.dotnet-openapi?view=aspnetcore-8.0

JPVenson (changes requested):

.vscode/tasks.json
ItsAllAboutTheCode: I'll look into using the .NET openapitools. 
I found the logic for the Jellyfin openapi generation in as part of this repo Github workflow which invokes a unit test to generate the openapi.json here: https://github.com/jellyfin/jellyfin/blob/master/.github/workflows/ci-openapi.yml#L12-L35

However I do think it would be good for the project to use the standalone tool command if possible.

I also think it would be good if there was some method to auto-install the `openapitools` tool when the Jellyfin.sln file is open for the first time in vscode.

Would you have any ideas regarding method to make it easier on developers to installing the OpenAPI tool automatically?

So far the idea I am thinking of following is replacing the task to run the OpenAPI test with a task that runs the `dotnet install` command if openapitools command isn't installed.

ItsAllAboutTheCode: Actually it looks like the [dotnet-tools.json](https://github.com/jellyfin/jellyfin/blob/master/.config/dotnet-tools.json) file in the repo can be used to install that dependency.

ItsAllAboutTheCode: @JPVenson I have updated the openapi generation logic to use the [Swashbuckle CLI tool](https://github.com/domaindrivendev/Swashbuckle.AspNetCore?tab=readme-ov-file#swashbuckleaspnetcorecli) to generate the `openapi.json` file.

I have left the Github CI workflows as is and those still build and run the Server intergration test to generate the `openapi.json`.

Jellyfin.Server/Program.cs
```diff
@@ -59,6 +59,18 @@ static Task ErrorParsingArguments(IEnumerable<Error> errors)
                 return Task.CompletedTask;
             }
 
+            // The dotnet swagger tool passes in an --applicationName that isn't allowed by the CLI parser
+            // Skip over it and add the "nowebclient" option to the server to allow it to start successfully
+            // Swagger is used to generate the OpenAPI document using the startup assembly and then exit the app
+            if (Assembly.GetEntryAssembly()?.GetName().Name == "dotnet-swagger")
```
JPVenson, line 65: if i remember correctly, all arguments you pass to swagger that are not parsed by it, will be handed down to the application itself. So if running the app without webclient, you can just put the cmd argument for that in the swagger cmd instead

JPVenson (changes requested):

Jellyfin.Server/Program.cs
ItsAllAboutTheCode: Well the `dotnet swagger` tool doesn't actually support forwarding additional arguments to the application.
Trying to pass in and additional argument of  "--nowebclient" has the swagger command fail with exit code 1.
 ```
dotnet swagger tofile --help
Usage: dotnet swagger tofile [options] [startupassembly] [swaggerdoc]

startupassembly:
  relative path to the application's startup assembly

swaggerdoc:
  name of the swagger doc you want to retrieve, as configured in your startup class

options:
  --output:  relative path where the Swagger will be output, defaults to stdout
  --host:  a specific host to include in the Swagger output
  --basepath:  a specific basePath to include in the Swagger output
  --serializeasv2:  output Swagger in the V2 format rather than V3
  --yaml:  exports swagger in a yaml format
```

JPVenson (approved):

prachi218200:
done

PR #12354, opened 2024-07-29
Add AC-4 downmix algorithm

gnattu:
This adds the AC-4 spec-defined downmix algorithm, which supports 3.0, 5.0, 5.1, 7.0 and 7.1 inputs. The downmix method is defined in [ETSI TS 103 190-1](https://www.etsi.org/deliver/etsi_ts/103100_103199/10319001/01.03.01_60/ts_10319001v010301p.pdf), Section 6.2.17.

This implementation uses the LoRo downmix mode (`preferred_dmx_method == 1`) as we do not need LtRt to support Dolby Pro Logic systems.

The channel mode for 7.1 inputs is always `3/4/0.x` because the channel layout of this mode matches the ffmpeg definition of 7.1.

All AC-4 stream-specific fields used in the downmixing formula are initialized with the spec-defined default values, which means:

- -3dB gain for `loro_cmg` and `loro_smg`, as defined in Table 148.
- -∞dB gain for `lfe_mg` (discard LFE), as defined in Section 4.3.12.2.18.

This PR also extends Dave750 and NightmodeDialogue to use the same cascaded downmix operations as AC-4 to support 7.1 inputs. These algorithms will now first downmix 7.1 inputs to 5.1 using the AC-4 downmix method, and then perform their own operations to downmix to stereo. 
  
**Changes**

- Adds AC-4 downmix algorithm
- Extends Dave750 and NightmodeDialogue to support 7.1 inputs.

**Issues**

~~Requires #12300, mark as draft until that one is merged~~

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/DownMixAlgorithmsHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/DownMixAlgorithmsHelper.cs
@@ -25,7 +25,9 @@ public static class DownMixAlgorithmsHelper
         { (DownMixStereoAlgorithms.Rfc7845, "6.1"), "pan=stereo|c0=0.321953*c2+0.455310*c0+0.394310*c5+0.227655*c6+0.278819*c4+0.321953*c3|c1=0.321953*c2+0.455310*c1+0.394310*c6+0.227655*c5+0.278819*c4+0.321953*c3" },
         { (DownMixStereoAlgorithms.Rfc7845, "7.1"), "pan=stereo|c0=0.274804*c2+0.388631*c0+0.336565*c6+0.194316*c7+0.336565*c4+0.194316*c5+0.274804*c3|c1=0.274804*c2+0.388631*c1+0.336565*c7+0.194316*c6+0.336565*c5+0.194316*c4+0.274804*c3" },
         { (DownMixStereoAlgorithms.Ac4, "3.0"), "pan=stereo|c0=c0+0.707*c2|c1=c1+0.707*c2"},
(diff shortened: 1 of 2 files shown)


crobibero (approved):

Shadowghost (approved):

MediaBrowser.Controller/MediaEncoding/DownMixAlgorithmsHelper.cs
```diff
@@ -15,13 +15,20 @@ public static class DownMixAlgorithmsHelper
     public static readonly Dictionary<(DownMixStereoAlgorithms, string), string> AlgorithmFilterStrings = new()
     {
         { (DownMixStereoAlgorithms.Dave750, "5.1"), "pan=stereo|c0=0.5*c2+0.707*c0+0.707*c4+0.5*c3|c1=0.5*c2+0.707*c1+0.707*c5+0.5*c3" },
+        { (DownMixStereoAlgorithms.Dave750, "7.1"), "pan=5.1(side)|c0=c0|c1=c1|c2=c2|c3=c3|c4=0.707*c4+0.707*c6|c5=0.707*c5+0.707*c7,pan=stereo|c0=0.5*c2+0.707*c0+0.707*c4+0.5*c3|c1=0.5*c2+0.707*c1+0.707*c5+0.5*c3" },
```
Bond-009, line 18: Maybe add a comment to these 2 that you're using the AC-4 7.1 to 5.1 downmix algo

Bond-009 (approved):

PR #12373, opened 2024-08-02
Use subtitle cache when burning-in subs

heartles:
The server does not currently use the subtitle cache (the one used for serving subtitles to clients to render) when burning-in subtitles at transcode time for clients that do not support client-side rendering. This can cause playback to stall when trying to play large files from slow media sources such as HDDs. 

**Changes**

This PR adds a method to `MediaBrowser.Controller.MediaEncoding.ISubtitleEncoder` to get the path of the extracted subtitle. `MediaBrowser.Controller.MediaEncoding.EncodingHelper` is modified to call this method instead of getting the subtitle directly from the media file itself.

**Issues**

This provides a partial solution to #11938, but the first attempt to play a file might still time out due to not already having the subtitles cached. This can be mitigated with a plugin like [Subtitle Extract](https://github.com/jellyfin/jellyfin-plugin-subtitleextract).

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -187,6 +187,7 @@
  - [HonestlyWhoKnows](https://github.com/honestlywhoknows)
  - [TheMelmacian](https://github.com/TheMelmacian)
  - [ItsAllAboutTheCode](https://github.com/ItsAllAboutTheCode)
+ - [jaina heartles](https://github.com/heartles)
 
 # Emby Contributors
 
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -1525,13 +1525,15 @@ namespace MediaBrowser.Controller.MediaEncoding
                     setPtsParam);
             }
 
-            var mediaPath = state.MediaPath ?? string.Empty;
+            var subtitlePath = _subtitleEncoder.GetSubtitleFilePath(
+                    state.SubtitleStream,
+                    state.MediaSource,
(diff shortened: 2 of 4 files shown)


gnattu:
What will happen if the subtitles are not extracted?

nyanmisaka:
Should fallback to old behavior when subtitle cache is not available.

heartles:
I'll make those changes this weekend.

sjorge:
After some debugging it seems I'm also hit by this :(

> What will happen if the subtitles are not extracted?

If `Allow subtitle extraction on the fly` is enabled, should they not be extracted first and then used?

That should help if the users seeks outside of the already buffered transcode or when switching to a different subtitle language since we have multi extract and it seems to take about the same time to extract them than it takes before the transcode starts with the current implementation.

oxixes:
Is this PR still active? I'd be willing to take over if it isn't

sjorge:
> Is this PR still active? I'd be willing to take over if it isn't

I don't think there are any rules for it, but personally after 2 or 3 months it would be fair game to pick up as long as you leave leave their commits as is and not rebase them.

JustinLex:
I would really like to see a PR like this get merged! #11938 is pretty frustrating watching anything with complex subs on Android.

Even just a solution that skips the cache when subtitles aren't extracted yet would be a game changer for subtitled content.

gnattu:
OP is not responding to our questions and not making requested changes, so I *think* anyone want to take it over can begin now to not miss the feature freeze window for 10.10

heartles:
yeah sorry go for it, I've been really busy with some personal stuff and haven't had a chance to finish it

nyanmisaka:
Closing in #12627

PR #12374, opened 2024-08-02
Fix compatibility between TranscodingThrottler and FFmpeg 7.0

nyanmisaka:
Preliminarily tested with https://github.com/jellyfin/jellyfin-ffmpeg/pull/424
- With multi-threaded cli support, FFmpeg 7 is less sensitive to keyboard input
- Vanilla FFmpeg 7 no longer supports using the `c` key to pause

**Changes**
- Fix compatibility between TranscodingThrottler and FFmpeg 7.0
- Add tests for FFmpeg 7.0.1

### DIFF
--- a/MediaBrowser.MediaEncoding/Encoder/EncoderValidator.cs
+++ b/MediaBrowser.MediaEncoding/Encoder/EncoderValidator.cs
@@ -477,7 +477,9 @@ namespace MediaBrowser.MediaEncoding.Encoder
             return false;
         }
 
-        public bool CheckSupportedRuntimeKey(string keyDesc)
+        private readonly Version _minFFmpegMultiThreadedCli = new Version(7, 0);
+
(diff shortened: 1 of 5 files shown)


crobibero (approved):

Shadowghost (approved):

gnattu (approved):

joshuaboniface:
Do we really want/need this in 10.9? I don't see us using FFmpeg 7 with the current stable, it would wait until 10.10.

nyanmisaka:
> Do we really want/need this in 10.9? I don't see us using FFmpeg 7 with the current stable, it would wait until 10.10.

If user does not use our jellyfin-ffmpeg but vanilla ffmpeg 7.0+, then transcoding progress will be corrupted when transcoding throttler is turned on.

PR #12375, opened 2024-08-02
Update the default repository URL for the StudioImages plugin

### LINKED ISSUE #10538: [Issue]: Studio Images plugin comes with default URL that 404's
### Please describe your bug

I just stumbled on this but I'm not sure this is the right repo, however I also [couldn't find a separate repository](https://github.com/orgs/jellyfin/repositories?q=studio&type=all&language=&sort=) for the Studio Images plugin, so here goes.

The Studio Images default plugin comes with a default URL: [`https://raw.github.com/jellyfin/emby-artwork/master/studios`](https://raw.github.com/jellyfin/emby-artwork/master/studios) however, as you can tell when trying to open this, it just returns `404: Not Found`

That subdomain has been redirecting to `raw.githubusercontent.com` for a while, e.g.: https://raw.githubusercontent.com/jellyfin/emby-artwork/master/studios/thumbs.txt

I don't believe it is planned to terminate the old subdomain, however I do believe it is a good idea to keep URLs current, if nothing else to avoid unnecessary DNS roundtrips.

Now, going to [`https://raw.githubusercontent.com/jellyfin/emby-artwork/master/studios/`](https://raw.githubusercontent.com/jellyfin/emby-artwork/master/studios/posters.txt) returns a `400: Invalid request`, however I assume this base-url is never called directly, but rather the files inside them so it would be fine.

On a side note: The hint shown on the field says "This can be any Jellyfin-compatible artwork repository. Leave blank to use default repository.", but, well... this field is filled by default, so why not just leave it blank for the default repository?
I noticed that when I delete the value and save, go back to plugins and back into the Plugin settings, it stays empty. However the field is filled in again after a restart of Jellyfin. Now I'm wondering what happens if the URL ever changes? Since the field is filled it would take precedence over the default, no?

### Jellyfin Version

10.8.z

### if other:

10.8.11

### Environment

```markdown
- OS:TrueNAS Scale
- Linux Kernel: 5.15.131
- Virtualization: Linuxserver.io Docker image (through TrueNAS Apps)
- Clients: Web
- Browser: Vivaldi/Chromium
- FFmpeg Version: 5.1.3-Jellyfin
- Playback Method: -
- Hardware Acceleration: none
- GPU Model: none
- Plugins: Studio Images
- Reverse Proxy: none
- Base URL: none
- Networking: Bridge
- Storage: local
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

tobias-varden:
**Changes**
Point the default repository server to the `raw.githubusercontent.com` domain instead of the old `raw.github.com` domain.

**Issues**
Fixes #10538

### DIFF
--- a/MediaBrowser.Providers/Plugins/StudioImages/Plugin.cs
+++ b/MediaBrowser.Providers/Plugins/StudioImages/Plugin.cs
@@ -18,7 +18,7 @@ namespace MediaBrowser.Providers.Plugins.StudioImages
         /// <summary>
         /// Artwork repository URL.
         /// </summary>
-        public const string DefaultServer = "https://raw.github.com/jellyfin/emby-artwork/master/studios";
+        public const string DefaultServer = "https://raw.githubusercontent.com/jellyfin/emby-artwork/master/studios";
 
         /// <summary>
         /// Initializes a new instance of the <see cref="Plugin"/> class.
--- a/MediaBrowser.Providers/Plugins/StudioImages/StudiosImageProvider.cs
+++ b/MediaBrowser.Providers/Plugins/StudioImages/StudiosImageProvider.cs
@@ -53,10 +53,7 @@ namespace MediaBrowser.Providers.Plugins.StudioImages
         /// <inheritdoc />
         public IEnumerable<ImageType> GetSupportedImages(BaseItem item)
         {
-            return new ImageType[]
-            {
-                ImageType.Thumb
-            };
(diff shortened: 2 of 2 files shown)


Bond-009 (approved):

Shadowghost (approved):
Should this target 10.9.z?

nielsvanvelzen (approved):

Issue #11880, opened 2024-05-30
[Issue]: Wrong content sorting in RU language.


borozenetsvv:
### Please describe your bug

Wrong content sorting in RU language. 
![image](https://github.com/jellyfin/jellyfin/assets/12553723/7bee1594-385e-48b2-8aff-2fb3cb2742b8)

### Reproduction Steps

Simply add files.

### Jellyfin Version

10.9.0

### Environment

```markdown
- OS: Ubuntu 22.04
- Linux Kernel: 6.5.0-27
- Virtualization:
- Clients: Any
- Browser: Any
- FFmpeg Version: 
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

```shell
Not need
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

geriko2000:
Same issue. I'm also find this problem after commit with Transliterator in unstable 9 version

geriko2000:
Name sort broken after this commit https://github.com/jellyfin/jellyfin/commit/c16135800c53873746321d0c9ea2fc36550511c1

melod1n:
Same issue

felix920506:
Marking confirmed from multiple reports

mirlex1986:
Jellyfin Version
10.9.11

`result = result.Transliterated()`

Same issue. Movies sorts like Transliterated name.

gnattu:
You will be able to disable the Transliterator in 10.10: https://github.com/jellyfin/jellyfin/pull/12378

trustin:
`JELLYFIN_TRANSLITERATOR_ID=Any-Hex` environment variable will make Jellyfin transliterate everything into hexadecimal representation of unicode value, which largely fixes my issue.

CHerSun:
Using `JELLYFIN_TRANSLITERATOR_ID=Any-Hex` content is sorted differently depending on if there's Sort title present or not. Like, setting sorting title to `Квартет И 2018` I expect the movie to be somewhere near other titles, which have no Sort title and start with `К` too, but instead it gets sorted way higher, right after numbers (`К` is in the middle of alphabet):

<img width="1340" height="591" alt="Image" src="https://github.com/user-attachments/assets/60fc5a70-e5ff-42fd-b8e6-eb3b6ec4521b" />

It looks that transliterator is applied differently to title and sort title

PR #12378, opened 2024-08-02
feat(i18n): able to finetune transliterator

BLumia:
**Changes**
After https://github.com/jellyfin/jellyfin/pull/11172, this gives user the ability to finetune the behavior of the existing transliterator to fit their own need. All they need to do is set the `JELLYFIN_TRANSLITERATOR_ID` environment variable to a preferred value, which is explained in [this page](https://unicode-org.github.io/icu/userguide/transforms/general/rules.html#script-transliterators) and [this page](https://unicode-org.github.io/icu/userguide/transforms/general/).

This approach is more flexible than only offers an option to disable the ICU transliterator, users can then:

1. Only transliterate specified curtain languages instead of "Any", for example, use `Hangul-Latin; Hiragana-Latin;` instead of `Any-Latin;` so only Hangul and Hiragana characters will get transliterated. Other non-ascii characters (e.g. Kanji and Chinese characters) will be untouched.
2. Finetune transliterator behavior (since you can set any valid Transliterator ID to that environment variable).
3. Simply disable it. Set this environment variable with an empty value, the transliterator will simply do nothing.

Some side notes (some of the notes have nothing to do with the changes made in this PR though):

1. Users (and plugins as well) can always override the sorting title by setting the sort title field manually. If a sort title is already manually set, Jellyfin won't attempt to override it.
2. If you choose to set sort title manually or by plugin, make sure at least the first character of the sort title is an ascii character, or you'll not able to find that item by using the alphabet filter on the right-hand side of the page.
3. When generating the fallback sort title, Jellyfin already attempt to remove diacritics in `ModifySortChunks` Before the ICU transliterator gets used. Keep that in mind when choosing the right transliterator rule (Transliterator ID).
4. You need to trigger a re-scan to make Jellyfin update the existing sort title cache. Choosing "Update Missing Metadata" will be fine.

**Issues**

- Fixes https://github.com/jellyfin/jellyfin/issues/11880
- See also: https://github.com/jellyfin/jellyfin/pull/11172#issuecomment-2249687663

### DIFF
--- a/src/Jellyfin.Extensions/StringExtensions.cs
+++ b/src/Jellyfin.Extensions/StringExtensions.cs
@@ -9,8 +9,21 @@ namespace Jellyfin.Extensions
     /// </summary>
     public static partial class StringExtensions
     {
-        private static readonly Lazy<Transliterator> _transliterator = new(() => Transliterator.GetInstance(
-            "Any-Latin; Latin-Ascii; Lower; NFD; [:Nonspacing Mark:] Remove; [:Punctuation:] Remove;"));
+        private static readonly Lazy<string> _transliteratorId = new(() =>
+            Environment.GetEnvironmentVariable("JELLYFIN_TRANSLITERATOR_ID")
+            ?? "Any-Latin; Latin-Ascii; Lower; NFD; [:Nonspacing Mark:] Remove; [:Punctuation:] Remove;");
+
+        private static readonly Lazy<Transliterator?> _transliterator = new(() =>
+        {
+            try
+            {
+                return Transliterator.GetInstance(_transliteratorId.Value);
+            }
+            catch (ArgumentException)
+            {
+                return null;
+            }
+        });
 
         // Matches non-conforming unicode chars
         // https://mnaoumov.wordpress.com/2014/06/14/stripping-invalid-characters-from-utf-16-strings/
@@ -108,7 +121,7 @@ namespace Jellyfin.Extensions
         /// <returns>The transliterated string.</returns>
         public static string Transliterated(this string text)
         {
-            return _transliterator.Value.Transliterate(text);
+            return (_transliterator.Value is null) ? text : _transliterator.Value.Transliterate(text);
         }
     }
 }


gnattu:
Related: https://github.com/jellyfin/jellyfin/issues/11880

src/Jellyfin.Extensions/StringExtensions.cs
```diff
@@ -9,8 +9,21 @@ namespace Jellyfin.Extensions
     /// </summary>
     public static partial class StringExtensions
     {
-        private static readonly Lazy<Transliterator> _transliterator = new(() => Transliterator.GetInstance(
-            "Any-Latin; Latin-Ascii; Lower; NFD; [:Nonspacing Mark:] Remove; [:Punctuation:] Remove;"));
+        private static readonly Lazy<string> _transliteratorId = new(() =>
+            Environment.GetEnvironmentVariable("JELLYFIN_TRANSLITERATOR_ID")
```
gnattu, line 13: I'm personally fine with reading env var directly but I'm not sure if there is any better way to integrate the configuration for string extensions

gnattu (approved):
Need better documentation, ideally on jellyfin.org. Probably also should mention this is in changelog as user actions needed to fix undesired behaviors. Otherwise LGTM

BLumia:
> Need better documentation, ideally on jellyfin.org.

Agree.

Related to this, maybe we should simply change `Sort title` to something else (e.g. "Transliterated title"), too? By doing this, user could understand the `sorttitle` metadata is supposed to be ascii-only, and maybe more plugin developer can make use of that metadata.

gnattu:
> sortname metadata is supposed to be ascii-only

Actually it is not. Computers do sort non ascii characters but may or may not in the human acceptable way and using ascii only characters is a easier way to control sorting behavior. But if users really want to input a non-ascii character it is also OK.

BLumia:
> Computers do sort non ascii characters but may or may not in the human acceptable way

Is there any real-life usage example to help me understand better about this usage? If it's intended to allow `sortname` contain non-ascii character, then why user cannot use the filter on the right-hand side to find there programs which their `sortname`s are started with non-ascii characters?

gnattu:
> > Computers do sort non ascii characters but may or may not in the human acceptable way
> 
> Is there any real-life usage example to help me understand better about this usage? If it's intended to allow `sortname` contain non-ascii character, then why user cannot use the filter on the right-hand side to find there programs which their `sortname`s are started with non-ascii characters?

> If it's intended to allow `sortname` contain non-ascii character, then why user cannot use the filter on the right-hand side to find there programs which their `sortname`s are started with non-ascii characters?

Because there are too many of them

> Is there any real-life usage example to help me understand better about this usage?

For example the ascii order of romaji of hiragana and katakanas are not in the same order as the Japanese traditional あいうえお order. Language is just complicated and you cannot assume the ASCII order is the most correct order.

BLumia:
> Because there are too many of them

If a user has too many items which starts with a non-ascii character, then I'd say the right-hand side filter is completely useless for such user. Things like [Alphabetic Index](https://icu.unicode.org/design/alphabetic-index) *might* be helpful to improve this, while I'm not 100% sure about it. Whatever if that can be improved, it will be a completely new discussion topic which is not related to this PR.

Anyway, related to this PR, as a conclusion, we won't change the copywriting of "Sort Title" to something else :)

crobibero (approved):

harry71019:
I tried setting `JELLYFIN_TRANSLITERATOR_ID` to nothing but that doesn't seem to change/disable the original settings. So I set it to `Hangul-Latin` since I don't have any hangul.

gnattu:
> I tried setting `JELLYFIN_TRANSLITERATOR_ID` to nothing but that doesn't seem to change/disable the original settings. So I set it to `Hangul-Latin` since I don't have any hangul.

It needs to be explicitly empty string, not really "nothing". Really nothing is the same as use default.

harry71019:
> > I tried setting `JELLYFIN_TRANSLITERATOR_ID` to nothing but that doesn't seem to change/disable the original settings. So I set it to `Hangul-Latin` since I don't have any hangul.
> 
> It needs to be explicitly empty string, not really "nothing". Really nothing is the same as use default.

Thanks for the explanation!

trustin:
`JELLYFIN_TRANSLITERATOR_ID=Any-Hex` transliterates everything into hexadecimal representation of unicode value, which largely fixes my issue.

PR #12384, opened 2024-08-03
Update QSV device args  to use the renderNodePath

evanjarrett:
When using QSV, there was no way to tell it to prioritize a different render device.
This allows passing in the renderNodePath into the device args so it is prioritized correctly.

**Changes**
 - Update QSV device args  to use the renderNodePath

**Issues**

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -903,7 +903,7 @@ namespace MediaBrowser.Controller.MediaEncoding
             if (OperatingSystem.IsWindows())
             {
                 // on Windows, the deviceIndex is an int
-                if (int.TryParse(renderNodePath, NumberStyles.Any, CultureInfo.InvariantCulture, out int deviceIndex))
+                if (int.TryParse(renderNodePath, NumberStyles.Integer, CultureInfo.InvariantCulture, out int deviceIndex))
                 {
                     return GetD3d11vaDeviceArgs(deviceIndex, string.Empty, D3d11vaAlias) + arg + "@" + D3d11vaAlias;
                 }
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -902,6 +902,12 @@ namespace MediaBrowser.Controller.MediaEncoding
 
             if (OperatingSystem.IsWindows())
             {
+                // on Windows, the deviceIndex is an int
+                if (int.TryParse(renderNodePath, NumberStyles.Any, CultureInfo.InvariantCulture, out int deviceIndex))
+                {
+                    return GetD3d11vaDeviceArgs(deviceIndex, string.Empty, D3d11vaAlias) + arg + "@" + D3d11vaAlias;
+                }
+
                 // derive qsv from d3d11va device
                 return GetD3d11vaDeviceArgs(0, "0x8086", D3d11vaAlias) + arg + "@" + D3d11vaAlias;
             }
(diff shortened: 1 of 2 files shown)


gnattu:
The problem I'm seeing this is that we have no way to guarantee that the path passed in is indeed an Intel device, and unfortunately the vappi render-device does have a default value that may or may not pointing to an Intel device, which means this change will break current install and will require user action to specify a different path or just remove that default value to be an empty string to restore the old behavior. 

If we are merging this as is we need to be clear in the changelog to inform multi-gpu users.

evanjarrett:
I think it would be better to create a new option for QSVDevice instead of re-using VaapiDevice, 
I was just having trouble getting that implemented and wanted to get something pushed up for feedback.

nyanmisaka:
Eventually this will be replaced by a drop-down menu showing a list of available GPUs.

MediaBrowser.Model/Configuration/EncodingOptions.cs
```diff
@@ -26,6 +26,7 @@ public EncodingOptions()
         // This is a DRM device that is almost guaranteed to be there on every intel platform,
         // plus it's the default one in ffmpeg if you don't specify anything
         VaapiDevice = "/dev/dri/renderD128";
+        QsvDevice = "";
```
Shadowghost, line 29: You can also use String.Empty

MediaBrowser.Model/Configuration/EncodingOptions.cs
```diff
@@ -136,6 +137,11 @@ public EncodingOptions()
     /// </summary>
     public string VaapiDevice { get; set; }
 
+    /// <summary>
+    /// Gets or sets the QSV device.
+    /// </summary>
+    public string QsvDevice { get; set; }
```
nyanmisaka, line 143: This will confuse QSV users on Windows. Please add support for it as well.

Users who modify this in settings should be aware of what they are doing, as this skips the vendor id check.

gnattu: This is not passed on Windows at all so there will be no issues, just document it that on Windows this option will be ignored.

nyanmisaka: > This is not passed on Windows at all so there will be no issues

So this only solves half of the problem - unable to select the index of QSV sub-device on both Linux & Windows.

evanjarrett: This has taken me awhile to get to, I don't actively have a windows device to be able to test this.
I'm having trouble targeting the correct device in windows using d3d11va
From my experience, it seems to split the load between the CPU and GPU.
When I try and set deviceIndex=1 I get an ffmpeg error and the stream crashes.
Testing with deviceIndex=0 yields the same results as the current implementation which just uses the vendor id.

I'm not sure how to modify the windows path in order to only target my intel GPU

nyanmisaka: For Windows, just allow `QsvDevice` to be a number >= 0.
If you don't have dual Intel graphics cards on a system, setting this to 1 will crash qsv.
e.g.
0 -> UHD770 iGPU
1 -> Arc A770 dGPU

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -891,17 +891,23 @@ private string GetDrmDeviceArgs(string renderNodePath, string alias)
...
             {
                 // derive qsv from vaapi device
-                return GetVaapiDeviceArgs(null, "iHD", "i915", null, VaapiAlias) + arg + "@" + VaapiAlias;
+                return GetVaapiDeviceArgs(renderNodePath, "iHD", "i915", null, VaapiAlias) + arg + "@" + VaapiAlias;
             }
 
             if (OperatingSystem.IsWindows())
             {
+                // on Windows, the deviceIndex is an int
+                if (int.TryParse(renderNodePath, NumberStyles.Any, CultureInfo.InvariantCulture, out int deviceIndex))
```
nyanmisaka, line 906:
```suggestion
                if (int.TryParse(renderNodePath, NumberStyles.Integer, CultureInfo.InvariantCulture, out int deviceIndex))
```

crobibero (approved):

Bond-009 (approved):

PR #12386, opened 2024-08-04
Fix Issue #12145: Audio Normalization Scheduled Task temp file race condition causing failure

### LINKED ISSUE #12145: [Issue]: Audio Normalization Scheduled Task Errors
### Please describe your bug

I posted this in the forum first, and it was suggested that I create this post here, as the logging is showing a locked concat file, during the audio normalization scheduled task routine, that appears to be an under the hood problem, possibly a minor bug. Or it could be a plugin, maybe. The main log is sparse on the topic. 

From the Forum (edited for brevity):

I recently reset my Jellyfin to a clean install on a new box. 

The Audio Normalization scheduled task continues to fail. Log snip of the failure event and surrounding events is below.

    [2024-06-17 20:27:51.680 -04:00] [ERR] [57] Emby.Server.Implementations.ScheduledTasks.TaskManager: Error executing Scheduled Task System.IO.IOException: The process cannot access the file 'C:\ProgramData\Jellyfin\Server\cache\transcodes\6bab25e2-5614-4b60-a3d7-735ccb7f6501.concat' because it is being used by another process.

      at System.IO.FileSystem.DeleteFile(String fullPath)

      at Emby.Server.Implementations.ScheduledTasks.Tasks.AudioNormalizationTask.ExecuteAsync(IProgress`1 progress, CancellationToken cancellationToken)

      at Emby.Server.Implementations.ScheduledTasks.ScheduledTaskWorker.ExecuteInternal(TaskOptions options)

    [2024-06-17 20:27:51.974 -04:00] [INF] [57] Emby.Server.Implementations.ScheduledTasks.TaskManager: "Audio Normalization" Failed after 0 minute(s) and 10 seconds

Jellyfin Server and PC Specs:

Jellyfin 10.9.6
Windows Server (Win 11 Pro) running Jellyfin service
AMD Ryzen 9 7950X
64GB DDR5-4800 RAM
NVIDIA RTX 3060 12GB
OS Drive: Samsung 990 Pro 2TB
Media Drive (all libraries): QNAP TS-932PX NAS with 5 x Seagate Ironwolf Pro 16TB in RAID5 connected with 10Gbps iSCSI link 

Logging output with debugging on, including the contents of the concat file, which is displayed with the same data in every concat file. It never changes. My audio library is thousands of songs, including audiobooks, and the scheduled task never runs for more than 10 seconds before bombing out. 

https://pastebin.com/a53KDFnG

### Reproduction Steps

Run Audio Normalization task. 

### Jellyfin Version

10.9.0

### if other:

10.9.6

### Environment

```markdown
- OS: Windows 11 Pro 64 23H2
- Linux Kernel: None
- Virtualization: None
- Clients: Browser and latest JMP 1.11.1
- Browser: Chrome, Edge, Firefox, Opera GX
- FFmpeg Version: Whatever the default is with 10.9.6
- Playback Method:
- Hardware Acceleration: NVIDIA NVENC
- GPU Model: RTX 3060 12GB
- Plugins: Artwork, AudioDB, Bookshelf, Covert Art Archive, Discogs, Fanart, IMVDb, Intro Skipper, MusicBrainz, OMDb, OPDS Feed, Open Subtitles, Session Cleaner, Studio Images, TMDb, TMDb Box Sets, TMDb Trailers, TVmaze, TheTVDB, VGMdb, Webhook
- Reverse Proxy: None
- Base URL: http://192.168.86.3
- Networking: local only. basic /24 subnet
- Storage: Noted in main description above.
```

### Jellyfin logs

```shell
On pastebin: https://pastebin.com/a53KDFnG
```

### FFmpeg logs

```shell
Please let me know where to find these if they're needed.
```

Let me know if these are needed and where to grab them from. 

_No response_

### Code of Conduct

AIsaac8573:
**Changes**

Making the CalculateLUFSAsync method wait for the external encoder process to end, conditionally before deleting the temp file.

The audio normalization task runs the encoder as a process.
When the output matches the regex value the CalculateLUFSAsync returned.
That is however could happen before the encoder external process ends, which is still locking the temp file.
Therefore the whole task failed for trying to delete the temp file before process ends. 
So I had to make CalculateLUFSAsync wait always for the external encoder process to end

I tested this fix on Windows 11 64 bit, and it's working as expected

**Issues**

Fixes #12145

### DIFF
--- a/Emby.Server.Implementations/ScheduledTasks/Tasks/AudioNormalizationTask.cs
+++ b/Emby.Server.Implementations/ScheduledTasks/Tasks/AudioNormalizationTask.cs
@@ -116,6 +116,7 @@ public partial class AudioNormalizationTask : IScheduledTask
                 {
                     a.LUFS = await CalculateLUFSAsync(
                         string.Format(CultureInfo.InvariantCulture, "-f concat -safe 0 -i \"{0}\"", tempFile),
+                        true,
                         cancellationToken).ConfigureAwait(false);
                 }
                 finally
@@ -142,7 +143,7 @@ public partial class AudioNormalizationTask : IScheduledTask
                     continue;
                 }
 
-                t.LUFS = await CalculateLUFSAsync(string.Format(CultureInfo.InvariantCulture, "-i \"{0}\"", t.Path.Replace("\"", "\\\"", StringComparison.Ordinal)), cancellationToken);
+                t.LUFS = await CalculateLUFSAsync(string.Format(CultureInfo.InvariantCulture, "-i \"{0}\"", t.Path.Replace("\"", "\\\"", StringComparison.Ordinal)), false, cancellationToken);
             }
 
             _itemRepository.SaveItems(tracks, cancellationToken);
@@ -162,7 +163,7 @@ public partial class AudioNormalizationTask : IScheduledTask
         ];
     }
 
-    private async Task<float?> CalculateLUFSAsync(string inputArgs, CancellationToken cancellationToken)
+    private async Task<float?> CalculateLUFSAsync(string inputArgs, bool waitForProcessToFinish, CancellationToken cancellationToken)
     {
         var args = $"-hide_banner {inputArgs} -af ebur128=framelog=verbose -f null -";
 
@@ -195,7 +196,11 @@ public partial class AudioNormalizationTask : IScheduledTask
 
                 if (match.Success)
                 {
-                    await process.WaitForExitAsync(cancellationToken);
+                    if (waitForProcessToFinish)
+                    {
+                        await process.WaitForExitAsync(cancellationToken);
+                    }
+
                     return float.Parse(match.Groups[1].ValueSpan, CultureInfo.InvariantCulture.NumberFormat);
                 }
             }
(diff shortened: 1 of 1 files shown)


Emby.Server.Implementations/ScheduledTasks/Tasks/AudioNormalizationTask.cs
```diff
@@ -195,7 +195,8 @@ await foreach (var line in reader.ReadAllLinesAsync(cancellationToken))
 
                 if (match.Success)
                 {
-                    return float.Parse(match.Groups[1].ValueSpan, CultureInfo.InvariantCulture.NumberFormat);
+                        await process.WaitForExitAsync(cancellationToken);
```
Shadowghost, line 198:
```suggestion
                        await process.WaitForExitAsync(cancellationToken).ConfigureAwait(false);
```

AIsaac8573: I tried to make that change, but now the debug logs(calculating....) are not shown in console. Can you check?

Bond-009:
Any reason you went from only waiting for the ffmpeg processes that use the concat files to waiting for every process to exit?
Also maybe it's worth it only waiting for exit on Windows as that's the only OS that has this problem.

Emby.Server.Implementations/ScheduledTasks/Tasks/AudioNormalizationTask.cs
JPVenson, line 198: A timeout for waiting for the process to stop should be included here. I am thinking of 5 sec

Bond-009:
Superseded by #13593

PR #12397, opened 2024-08-05
Add lyrics library options, add download scheduled task

crobibero:


MediaBrowser.Providers/Lyric/LyricScheduledTask.cs
```diff
@@ -0,0 +1,171 @@
...
+            int previousCount;
+            var startIndex = 0;
+            do
+            {
+                itemQuery.StartIndex = startIndex;
+                var audioItems = _libraryManager.GetItemList(itemQuery);
+
+                foreach (var audioItem in audioItems.OfType<Audio>())
+                {
+                    cancellationToken.ThrowIfCancellationRequested();
```
JPVenson, line 107: i would prefer to abort gracefully here instead of throwing

MediaBrowser.Providers/Lyric/LyricScheduledTask.cs
```diff
@@ -0,0 +1,171 @@
...
+        _lyricManager = lyricManager;
+        _logger = logger;
+        _localizationManager = localizationManager;
+    }
+
+    /// <inheritdoc />
+    public string Name => _localizationManager.GetLocalizedString("TaskDownloadMissingLyrics");
+
+    /// <inheritdoc />
+    public string Key => "DownloadLrcLibLyrics";
```
nielsvanvelzen, line 60:
```suggestion
    public string Key => "DownloadLyrics";
```

MediaBrowser.Providers/Lyric/LyricScheduledTask.cs
crobibero: The `ScheduledTaskWorker` sets the status based on how the scheduled task completes so right now this is how we specify that the task has been cancelled instead of completed successfully.

JPVenson (approved):

Bond-009 (approved):

gnattu:
This caused 

>LyricScheduledTask.cs(153,13): Error SA1508 : A closing brace should not be preceded by a blank line (https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1508.md)

PR #12398, opened 2024-08-05
Update XmlTvListingsProvider.cs to fix the record series button missing on many programs

bpauquette:
Fixes a bug where certain programs like the "Good Morning America" no longer have the record series button.

**Changes**
Assume that if the program has an EpisodeNumber then it is part of a series. 

**Issues**
The record series button is not displaying on many programs that are generated by zaptoxml.pl perl script.    

https://github.com/jef/zap2xml

### DIFF
--- a/src/Jellyfin.LiveTv/Listings/XmlTvListingsProvider.cs
+++ b/src/Jellyfin.LiveTv/Listings/XmlTvListingsProvider.cs
@@ -167,7 +167,7 @@ namespace Jellyfin.LiveTv.Listings
                 Overview = program.Description,
                 ProductionYear = program.CopyrightDate?.Year,
                 SeasonNumber = program.Episode.Series,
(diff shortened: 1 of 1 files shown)


bpauquette:
Fixes issue #11976

bpauquette:
For reference the xmltv.dtd is here https://github.com/XMLTV/xmltv/blob/master/xmltv.dtd  .   I am not sure if it really speaks about an isSeries boolean.  Seems like it must be derived.

Bond-009 (approved):

Shadowghost (approved):

theguymadmax:
This probably should have targeted the 10.9.z branch. Otherwise, you'll have some grumpy guy commenting every week what's taking so long ☺️.

nielsvanvelzen (commented):
Please add a descriptive title to the pull request

PR #12403, opened 2024-08-06
Feature/item aware tasks

JPVenson:
Added Library triggered Scheduled task to run when changes to items where made.


MediaBrowser.Controller/Tasks/IBaseItemScheduledTask.cs
```diff
@@ -0,0 +1,23 @@
...
+public interface IBaseItemScheduledTask : IScheduledTask
+{
+    /// <summary>
+    /// Runs this task in the context of one or many base items.
+    /// </summary>
+    /// <param name="progress">The progress.</param>
+    /// <param name="baseItems">The list of items that where changed.</param>
+    /// <param name="cancellationToken">The cancellation token.</param>
+    /// <returns>A task.</returns>
+    Task ExecuteAsync(IProgress<double> progress, IReadOnlyList<(BaseItem Item, TaskActionTypes ChangeType)> baseItems, CancellationToken cancellationToken);
```
nielsvanvelzen, line 22: Might be more efficient to send a list of baseitems instead of calling the execute function multiple times.

JPVenson: Please elaborate, there is already a list of baseItems with thier associated action as a parameter.

nielsvanvelzen: Whoops, I did't read the list part of the signature. nvm my comment

MediaBrowser.Model/Tasks/ITaskManager.cs
```diff
@@ -17,7 +17,7 @@ public interface ITaskManager : IDisposable
         /// Gets the list of Scheduled Tasks.
         /// </summary>
         /// <value>The scheduled tasks.</value>
-        IScheduledTaskWorker[] ScheduledTasks { get; }
+        IReadOnlyCollection<IScheduledTaskWorker> ScheduledTasks { get; }
```
Shadowghost, line 20: Any reason to use `IReadOnlyCollection` over `IReadOnlyList`?

MediaBrowser.Model/Tasks/ITaskManager.cs
```diff
@@ -70,7 +70,7 @@ void QueueIfNotRunning<T>()
         /// Adds the tasks.
         /// </summary>
         /// <param name="tasks">The tasks.</param>
-        void AddTasks(IEnumerable<IScheduledTask> tasks);
+        void AddTasks(IReadOnlyCollection<IScheduledTask> tasks);
```
Shadowghost, line 73: Same as above

Emby.Server.Implementations/ScheduledTasks/TaskManager.cs
```diff
@@ -49,7 +55,7 @@ public TaskManager(
         /// Gets the list of Scheduled Tasks.
         /// </summary>
         /// <value>The scheduled tasks.</value>
-        public IScheduledTaskWorker[] ScheduledTasks { get; private set; }
+        public IReadOnlyCollection<IScheduledTaskWorker> ScheduledTasks { get; private set; }
```
Shadowghost, line 58: Any reason to use `IReadOnlyCollection` over `IReadOnlyList`? PR currently uses both

Emby.Server.Implementations/ScheduledTasks/TaskManager.cs
```diff
@@ -190,9 +196,12 @@ private void QueueScheduledTask(IScheduledTaskWorker task, TaskOptions options)
         /// Adds the tasks.
         /// </summary>
         /// <param name="tasks">The tasks.</param>
-        public void AddTasks(IEnumerable<IScheduledTask> tasks)
+        public void AddTasks(IReadOnlyCollection<IScheduledTask> tasks)
```
Shadowghost, line 199: Same as above

Emby.Server.Implementations/ScheduledTasks/TaskManager.cs
```diff
@@ -25,20 +28,23 @@ public class TaskManager : ITaskManager
...
         /// Initializes a new instance of the <see cref="TaskManager" /> class.
         /// </summary>
         /// <param name="applicationPaths">The application paths.</param>
         /// <param name="logger">The logger.</param>
+        /// <param name="serviceProvider">The service provider.</param>
         public TaskManager(
             IApplicationPaths applicationPaths,
-            ILogger<TaskManager> logger)
+            ILogger<TaskManager> logger,
+            IServiceProvider serviceProvider)
```
Shadowghost, line 42: Do we really need to pull in the full service provider just to later get the lilbrary manager?

Emby.Server.Implementations/ScheduledTasks/ScheduledTaskWorker.cs
```diff
@@ -3,6 +3,9 @@
 #pragma warning disable CS1591
```
Shadowghost, line 3: Can we fix that too? Class was heavily modified anyways

Shadowghost:
I'm a bit concerned about what happens if a lot of items change. Ideally we should have something different than a scheduled task for handling item-specific tasks.

MediaBrowser.Model/Tasks/ITaskManager.cs
JPVenson: no particular

Emby.Server.Implementations/ScheduledTasks/TaskManager.cs
JPVenson: ILibraryManager injects ITaskManager creating a ciruclar dependecy otherwise. Without factories i dont know how to get around that.

EraYaN:
> I'm a bit concerned about what happens if a lot of items change. Ideally we should have something different than a scheduled task for handling item-specific tasks.

Or at least a backoff or something. But the biggest problem is if this task edits items again, or there are bunch of them order might be important. You get reconciliation loops like you get in k8s controllers that go bonkers. k8s has an API to return "requeue: false" if it's "done" and it detects actual fields changed to decide if it should requeue if it's not specified.

So the Changed event should probably do a deep compare to avoid loops and the docs should document that that is how the reconciliation loop works for items.

JPVenson:
Well we have. There is a delay of 30 sec in the `BaseItemScheduledTaskWorker` that aggregates changes and will only trigger if there is for a rolling 30 sec window no change detected so to not hinder any other active task. The Task will then get handed the *list* of changed tasks and not invoked for each task individually.

JPVenson:
@Shadowghost yes i would also prefer having something more purpose build for this case but without building something from the ground up again i dont see that happening in JF anytime soon.

Emby.Server.Implementations/ScheduledTasks/TaskManager.cs
JPVenson: Correct, but that would as i said just create a factory and move the usage of the serviceprovider somewhere else ^^

mark-monteiro: Yes, but it makes the dependencies of this class more clear and is also makes a cleaner implementation if you need to mock the dependency for testing

JPVenson:
closed due to inactivity

PR #12428, opened 2024-08-11
Remove unused encoders

gnattu:
**Changes**

Removes the old code paths for the encoders that we are not really supporting. This includes:

- libvpx based encoders, like vp8 and vp9
- mpeg4 encoders
- WMA and WMV endcoders
- libtheora encoder

None of these encoders have ever implemented a hardware acceleration path and are not truly supported by us. The current parameters for these encoders might also be incorrect, which could cause confusion for users who craft special API requests using these encoders. Removing them to confusion and send a clear message that these encoders are not supported.

**Issues**

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -107,7 +107,6 @@ namespace MediaBrowser.Controller.MediaEncoding
         // AAC, FLAC, ALAC, libopus, libvorbis encoders all support at least 8 channels
         private static readonly Dictionary<string, int> _audioTranscodeChannelLookup = new(StringComparer.OrdinalIgnoreCase)
         {
-            { "wmav2", 2 },
             { "libmp3lame", 2 },
             { "libfdk_aac", 6 },
             { "ac3", 6 },
@@ -398,16 +397,6 @@ namespace MediaBrowser.Controller.MediaEncoding
                 if (string.Equals(codec, "mjpeg", StringComparison.OrdinalIgnoreCase))
                 {
                     return GetMjpegEncoder(state, encodingOptions);
-                }
-
-                if (string.Equals(codec, "wmv", StringComparison.OrdinalIgnoreCase))
-                {
-                    return "wmv2";
-                }
-
-                if (string.Equals(codec, "theora", StringComparison.OrdinalIgnoreCase))
-                {
-                    return "libtheora";
                 }
 
                 if (_validationRegex.IsMatch(codec))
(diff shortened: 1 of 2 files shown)


nyanmisaka:
There are still a few leftovers. https://github.com/search?q=repo%3Ajellyfin%2Fjellyfin+Options%3A+h265&type=code

gnattu:
> There are still a few leftovers. https://github.com/search?q=repo%3Ajellyfin%2Fjellyfin+Options%3A+h265&type=code

All of these API docs are outdated and the descriptions do not really reflects to what the server actually behaves. I just removed the valid options list as that field may or may not be used as encoding targets.

Jellyfin.Api/Controllers/DynamicHlsController.cs
```diff
@@ -902,7 +902,7 @@ public class DynamicHlsController : BaseJellyfinApiController
     /// <param name="cpuCoreLimit">Optional. The limit of how many cpu cores to use.</param>
     /// <param name="liveStreamId">The live stream id.</param>
     /// <param name="enableMpegtsM2TsMode">Optional. Whether to enable the MpegtsM2Ts mode.</param>
-    /// <param name="videoCodec">Optional. Specify a video codec to encode to, e.g. h264. If omitted the server will auto-select using the url's extension. Options: h265, h264, mpeg4, theora, vpx, wmv.</param>
+    /// <param name="videoCodec">Optional. Specify a video codec to encode to, e.g. h264. Options: h265, h264, mpeg4, theora, vpx, wmv.</param>
```
nyanmisaka, line 905: > Options: h265, h264, mpeg4, theora, vpx, wmv.

Some outdated options.

Jellyfin.Api/Controllers/DynamicHlsController.cs
```diff
@@ -1258,7 +1258,7 @@ public class DynamicHlsController : BaseJellyfinApiController
...
```
nyanmisaka, line 1261: > Options: h265, h264, mpeg4, theora, vpx, wmv.

Some outdated options.

nyanmisaka (approved):

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -6428,24 +6300,15 @@ public string GetRkmppVidDecoder(EncodingJobInfo state, EncodingOptions options,
...
-            }
-
-            if (threads >= Environment.ProcessorCount)
-            {
-                return Environment.ProcessorCount;
+                return 0;
             }
 
-            return threads;
+            return threads >= Environment.ProcessorCount ? Environment.ProcessorCount : threads;
```
Bond-009, line 6311:
```suggestion
            return Math.Min(threads, Environment.ProcessorCount);
```

Bond-009 (approved):

Shadowghost (approved):

crobibero (approved):

PR #12430, opened 2024-08-12
Add option to always burn in subtitles if transcoding is triggered

gnattu:
**Changes**

Subtitles often have synchronization issues after video transcoding. This option allows the client to opt-in to always burn in all subtitles, preventing out-of-sync subtitles at the cost of transcoding performance.

To support this feature, the `TranscodeReason` has been extended to accurately determine if a stream can be remuxed.

Notably:

- `VideoCodecTag` now has its own transcode reason instead of sharing one with `VideoCodec`. The former can be fixed by remuxing, while the latter always triggers a full transcode.
- The `VideoReasons` have been updated to check for profile mismatches, as these will also require transcoding.
- `DirectStreamingReasons` now includes the new `VideoCodecTagNotSupported` reason.

**Issues**

### DIFF
--- a/Jellyfin.Api/Controllers/MediaInfoController.cs
+++ b/Jellyfin.Api/Controllers/MediaInfoController.cs
@@ -109,7 +109,6 @@ public class MediaInfoController : BaseJellyfinApiController
     /// <param name="enableTranscoding">Whether to enable transcoding. Default: true.</param>
     /// <param name="allowVideoStreamCopy">Whether to allow to copy the video stream. Default: true.</param>
     /// <param name="allowAudioStreamCopy">Whether to allow to copy the audio stream. Default: true.</param>
-    /// <param name="alwaysBurnInSubtitleWhenTranscoding">Always burn-in subtitle when transcoding.</param>
     /// <param name="playbackInfoDto">The playback info.</param>
     /// <response code="200">Playback info returned.</response>
     /// <response code="404">Item not found.</response>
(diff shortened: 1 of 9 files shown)


Jellyfin.Api/Controllers/MediaInfoController.cs
```diff
@@ -132,6 +133,7 @@ public async Task<ActionResult<PlaybackInfoResponse>> GetPlaybackInfo([FromRoute
         [FromQuery, ParameterObsolete] bool? enableTranscoding,
         [FromQuery, ParameterObsolete] bool? allowVideoStreamCopy,
         [FromQuery, ParameterObsolete] bool? allowAudioStreamCopy,
+        [FromQuery, ParameterObsolete] bool? alwaysBurnInSubtitleWhenTranscoding,
```
nielsvanvelzen, line 136: Don't introduce new parameters that are immediately obsolete. This option should only come from the body.

gnattu: The problem: web is using deprecated apiclient for this endpoint

nielsvanvelzen: Then web should use the newer non-deprecated endpoints via the SDK first. We're strictly not supporting legacy endpoints unless it's needed to keep a client working. In this case it's a new feature.

crobibero (approved):

nyanmisaka (approved):

Issue #11854, opened 2024-05-27
[Issue]: Cannot delete users


sypion:
### Please describe your bug

When attempting to delete a user from my Jellyfin server it does not proceed and continuously shows the loading animation.

### Reproduction Steps

1. Visit the `Users` tab in your admin dashboard.
2. Find the user you'd like to delete from your Jellyfin server
3. Click on `Delete`

### Jellyfin Version

10.9.0

### if other:

10.9.3

### Environment

```markdown
- OS: Linux (Ubuntu) x64
- Linux Kernel: 5.15.0-107-generic
- Virtualization: KVM
- Clients: Browser, Windows Client
- Browser: Arc (Chromium)
- FFmpeg Version: 6
- Reverse Proxy: No
- Base URL: localhost
- Networking: Host
- Storage: local
```

### Jellyfin logs

```shell
[2024-05-27 11:20:54.192 -07:00] [ERR] Error processing request. URL "DELETE" "/Users/a8d22bd83f9840c7b3e263d6b3f51e8f".
System.NullReferenceException: Object reference not set to an instance of an object.
   at Emby.Server.Implementations.Playlists.PlaylistManager.GetUserPlaylists(Guid userId)
   at Emby.Server.Implementations.Playlists.PlaylistManager.RemovePlaylistsAsync(Guid userId)
   at Jellyfin.Api.Controllers.UserController.DeleteUser(Guid userId)
   at lambda_method2405(Closure, Object)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ActionMethodExecutor.TaskOfActionResultExecutor.Execute(ActionContext actionContext, IActionResultTypeMapper mapper, ObjectMethodExecutor executor, Object controller, Object[] arguments)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.<InvokeActionMethodAsync>g__Awaited|12_0(ControllerActionInvoker invoker, ValueTask`1 actionResultValueTask)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.<InvokeNextActionFilterAsync>g__Awaited|10_0(ControllerActionInvoker invoker, Task lastTask, State next, Scope scope, Object state, Boolean isCompleted)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.Rethrow(ActionExecutedContextSealed context)
[...]

PR #12442, opened 2024-08-13
Properly handle null values in PlaylistManager.cs

phito:
`PlaylistManager.cs` had a #nullable disable declaration, which lead to NullReferenceException  when dependencies are return null values.

**Changes**
Removed `#nullable disable` and properly handling null values in `PlaylistManager.cs`.

**Issues**
529d24ec308645e92839b216fd0b52acd4ad62a2

#11854 was definitely cause by this, and I believe it has already been fixed by [this commit](https://github.com/jellyfin/jellyfin/commit/529d24ec308645e92839b216fd0b52acd4ad62a2), but it did not address the root cause.

(this is also practice for my first contribution, thanks all of you for this awesome software, I am looking forward to contribute more)

### DIFF
--- a/Emby.Server.Implementations/Playlists/PlaylistManager.cs
+++ b/Emby.Server.Implementations/Playlists/PlaylistManager.cs
@@ -590,7 +590,12 @@ namespace Emby.Server.Implementations.Playlists
                     EnableImages = true
                 }).ConfigureAwait(false);
 
-                playlist = GetPlaylistForUser(request.Id, request.UserId)!;
+                playlist = GetPlaylistForUser(request.Id, request.UserId);
+            }
+
+            if (playlist is null)
+            {
+                throw new ArgumentException($"No playlist exists with Id {request.Id}");
             }
 
             if (request.Name is not null)
--- a/Emby.Server.Implementations/Playlists/PlaylistManager.cs
+++ b/Emby.Server.Implementations/Playlists/PlaylistManager.cs
@@ -1,5 +1,3 @@
-#nullable disable
-
 #pragma warning disable CS1591
 
 using System;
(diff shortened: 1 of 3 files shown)


gnattu:
I'm sorry but I don't think just yolo change all return types to be nullable without looking at the actual usage is proper.

The current handling can definitely be improved by checking all the null cases but just return null would not work well because all the usages would assume the return value is not null.

Emby.Server.Implementations/Playlists/PlaylistManager.cs
```diff
@@ -566,19 +569,28 @@ public async Task RemovePlaylistsAsync(Guid userId)
...
                 await UpdatePlaylistInternal(playlist).ConfigureAwait(false);
 
-                var user = _userManager.GetUserById(request.UserId);
                 await AddToPlaylistInternal(request.Id, request.Ids, user, new DtoOptions(false)
                 {
                     EnableImages = true
                 }).ConfigureAwait(false);
 
-                playlist = GetPlaylistForUser(request.Id, request.UserId);
+                playlist = GetPlaylistForUser(request.Id, request.UserId)!;
```
gnattu, line 593: Avoid this because this is not properly handling the null case

phito: It has already been checked by the `GetPlaylistForUser` call above, so it is safe to assume that UserId is valid. But I understand if you want to make it more explicit, I can handle the null value there too.

phito: We're getting a playlist we've just updated, so it is safe so assume it will not be null. But I can make it more explicit and handle it again if it is more compliant to standards.

PR #12446, opened 2024-08-14
Remove left-over network path references

nielsvanvelzen:
Although removed in 10.9 we forgot to remove some references to the old network path support. This caused the server to still send these paths in specific cases.

**Changes**
- Remove left-over network path references

**Issues**

### DIFF
--- a/Emby.Server.Implementations/Library/LibraryManager.cs
+++ b/Emby.Server.Implementations/Library/LibraryManager.cs
@@ -2725,33 +2725,9 @@ namespace Emby.Server.Implementations.Library
 
         public string GetPathAfterNetworkSubstitution(string path, BaseItem? ownerItem)
         {
-            string? newPath;
-            if (ownerItem is not null)
-            {
-                var libraryOptions = GetLibraryOptions(ownerItem);
-                if (libraryOptions is not null)
-                {
-                    foreach (var pathInfo in libraryOptions.PathInfos)
-                    {
-                        if (path.TryReplaceSubPath(pathInfo.Path, pathInfo.NetworkPath, out newPath))
-                        {
-                            return newPath;
-                        }
-                    }
-                }
-            }
-
-            var metadataPath = _configurationManager.Configuration.MetadataPath;
-            var metadataNetworkPath = _configurationManager.Configuration.MetadataNetworkPath;
-
-            if (path.TryReplaceSubPath(metadataPath, metadataNetworkPath, out newPath))
-            {
-                return newPath;
-            }
-
             foreach (var map in _configurationManager.Configuration.PathSubstitutions)
             {
-                if (path.TryReplaceSubPath(map.From, map.To, out newPath))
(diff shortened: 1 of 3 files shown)


Shadowghost (approved):

oddstr13 (commented):
I was under the impression that this was going to stay in for some time still, as quite a few people use the «native» mode in jf4kodi, which depends on this.

There's some discussion after the removal from the UI here https://github.com/jellyfin/jellyfin/discussions/11654

I do not think this is a good idea.

gnattu:
Well it seems like this path is only used for kodi clients and nowhere else, which means we can have some mitigations if we still want to support the native mode:

- Move the network path to library mapping into a plugin and provide an api endpoint in that plugin,  which translate the physical paths on the server mount point to the network path, and let the kodi clients request that api if native mode is preferred.

- Make this mapping entirely client-side. Users set network path mapping to mount point mapping in the client side entirely.

To have this path in the server as part of the library setting is hard to maintain and make it hard to implement proper path handling for both the server and clients other than kodi.

mcarlton00:
> Well it seems like this path is only used for kodi clients and nowhere else

Kodi native mode, mpv shim, and iirc at least one third party client, but I don't remember which one.

We all hate native mode and it's basically unsupported on the kodi side, but there are unfortunately some circumstances where it does work better (BDMV playback and vobsub subtitles both immediately spring to mind).  However, removing it basically just brings it in line with other clients instead of having special oddball treatment.

Bond-009 (approved):

it5c0z1mG:
Is there any plan to make native mode an addon feature or some other way to manually undo your removal of this? in the very near future hopefully, native worked flawlessly, tried to switch to addon mode and it is ridiculously useless in everyway and jellycon is slow aswell as useless, everything is now broken, happily reverted and block auto updates but just want to know if i should wait and hope or whether to just remove jellyfin in its entirety now and go back to a sole kodi usage. Thanks.

mcarlton00:
https://github.com/jellyfin/jellyfin-kodi/pull/926

it5c0z1mG:
Missed that! Amazing news, will test it tommorrow

omelettedufromagee:
> > Well it seems like this path is only used for kodi clients and nowhere else
> 
> Kodi native mode, mpv shim, and iirc at least one third party client, but I don't remember which one.
> 
> We all hate native mode and it's basically unsupported on the kodi side, but there are unfortunately some circumstances where it does work better (BDMV playback and vobsub subtitles both immediately spring to mind). However, removing it basically just brings it in line with other clients instead of having special oddball treatment.

I confirm. On a remote server, Native mode using WebDAV works wayyyyyyy better than Addon-mode on Kodi for large file (like a Remux). It puts the strain on the client rather than the server.
When using addon-mode, the Jellyfin server always drops the ball at some point (resuming, forwarding etc) and it is way less responsive overall.

Issue #11331, opened 2024-04-10
[Issue]:  newer Abi shows/installs older targetAbi even if it is not compatible


jumoog:
### Please describe your bug

I have my plugin for several Abi versions. So people don't have to change their url and there is only one url. In `10.9` the `10.8.4` versions are offered to me and also automatically updated when there is a new `10.8.4` version.

Jellyfin should only allow minor releases. For example, `10.8.13` allows `"targetAbi": "10.8.4.0"`. 10.9.x should ignore all `10.8.x` versions.

### Reproduction Steps

- Add `https://raw.githubusercontent.com/jumoog/jellyfin-plugin-repo/main/manifest.json` as Plugin source.
- It shows all Versions
- Install the newest Version
- It successfully installs
- After reboot it shows `Status NotSupported`

### Jellyfin Version

Unstable (master branch)

### if other:

jellyfin/jellyfin:2024040805-amd64

### Environment

```markdown
- Virtualization: Docker
```

### Jellyfin logs

```shell
[10:36:59] [INF] [54] Emby.Server.Implementations.Plugins.PluginManager: No local manifest exists for plugin Intro Skipper. Skipping manifest reconciliation.
[10:36:59] [INF] [54] Emby.Server.Implementations.Updates.InstallationManager: Plugin installed: Intro Skipper 0.1.16.5
[10:36:59] [INF] [54] Emby.Server.Implementations.ApplicationHost: App needs to be restarted.
[10:37:05] [INF] [57] Emby.Server.Implementations.Session.SessionManager: Sending shutdown notifications
[10:37:05] [INF] [51] Emby.Server.Implementations.HttpServer.WebSocketManager: WS 172.20.0.1 closed
[10:37:05] [INF] [55] Jellyfin.Networking.PortForwardingHost: Stopping NAT discovery
[10:37:05] [INF] [51] Main: Running query planner optimizations in the database... This might take a while
[10:37:05] [INF] [58] Emby.Server.Implementations.IO.LibraryMonitor: Stopping directory watching for path /test
[10:37:05] [INF] [51] Emby.Server.Implementations.ApplicationHost: Disposing CoreAppHost
[10:37:05] [INF] [51] Emby.Server.Implementations.ApplicationHost: Disposing PluginManager
[10:37:05] [INF] [51] Emby.Server.Implementations.ApplicationHost: Disposing MusicBrainzAlbumProvider
[...]

PR #12460, opened 2024-08-16
plugins: add optional maximumAbi option

jumoog:
In 10.9 there have been some breakable changes. This change prevents users from installing plugins that do not work after installation. It also removes plugins from the catalogue if there is no 10.9 Plugin version. I do not expect breakable changes between versions very often. This can be adjusted in the future if another breakable change happens.

**Changes**
Filter plugins targeting version 10.8 or older.

**Issues**
#11331


crobibero:
I think a proper solution would be to add a `maximumAbi` field that a plugin manifest can populate.

jumoog:
todo tests

jumoog:
Not sure if i ping someone for a review?

Emby.Server.Implementations/Updates/InstallationManager.cs
```diff
@@ -123,6 +123,12 @@ public async Task<PackageInfo[]> GetPackages(string manifestName, string manifes
                             continue;
                         }
 
+                        // Only show plugins that are less than or equal to maximumAbi.
+                        if (Version.TryParse(ver.MaximumAbi, out var maximumAbi) && _applicationHost.ApplicationVersion <= maximumAbi)
```
Maxr1998, line 127: With this change, if the `maximumAbi` matches, the `targetAbi` isn't checked anymore. The logic of this whole loop should probably be inverted to fix it, and handle both Abi fields properly.

Maxr1998 (changes requested):

Emby.Server.Implementations/Updates/InstallationManager.cs
AbandonedCart: Don't you mean to swap the order of the checks?

As it stands, you can install a plugin that is made for an older server version. You can also install a version meant for a newer server version than you are running. The problem is that when a plugin updates to the higher version before the server, the user is left in an awkward situation where they have to prevent updates. By placing a cap on the version, you prevent that from happening. Setting a max ABI acts as that cap, so you want the server version to be **lower** than the cap.

jumoog: I think i tested all use cases but i'll do my tests again

jumoog: the continue skips the `if (_applicationHost.ApplicationVersion >= targetAbi)` check?

Issue #12458, opened 2024-08-16
Playback issue with new video files added 


Sarxzer:
### This issue respects the following points:

- [X] This is a **bug**, not a question or a configuration issue; Please visit our forum or chat rooms first to troubleshoot with volunteers, before creating a report. The links can be found [here](https://jellyfin.org/contact/).
- [X] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [X] I'm using an up to date version of Jellyfin Server stable, unstable or master; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.
- [X] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [X] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Description of the bug

Since recently, when I add a video file and scan the folder, Jellyfin is able to find it, name it an get every metadata, but when I try to launch the video, I get a "Playback failed due to a fatal player error.". Using vlc, I check the codec, resolution etc but they are the same as other working video files.

### Reproduction steps

I don't know how to reproduce this bug. 

### What is the current _bug_ behavior?

I get an error when I try to launch a recently added video, and it never did that. 

### What is the expected _correct_ behavior?

Launching the video without error. 

### Jellyfin Server version

10.9.9+

### Specify the build version

10.9.9

### Environment

```markdown
- OS:Windows 
- Linux Kernel:none
- Virtualization:
- Clients:All
- Browser:All
- FFmpeg Version:
- Playback Method:I don't know 
- Hardware Acceleration:none, nvenc 
- GPU Model:RTX 4060
- Plugins:Anilist, Open Subtitles
- Reverse Proxy:none
- Base URL:none
- Networking:host
- Storage:local
```

### Jellyfin logs

```shell
[log_20240816.txt](https://github.com/user-attachments/files/16641368/log_20240816.txt)
```

### Client / Browser logs

[localhost-1723839874793.log](https://github.com/user-attachments/files/16641456/localhost-1723839874793.log)

gnattu:
Jellyfin is unable to locate ffmpeg in your install. Did you set env var or the config to overwrite the location, and point to an invalid path?

Sarxzer:
[...]

Issue #12411, opened 2024-08-08
Playback error for some files


realtofuine:
### This issue respects the following points:

- [X] This is a **bug**, not a question or a configuration issue; Please visit our forum or chat rooms first to troubleshoot with volunteers, before creating a report. The links can be found [here](https://jellyfin.org/contact/).
- [X] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [X] I'm using an up to date version of Jellyfin Server stable, unstable or master; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.
- [X] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [X] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Description of the bug

I upgraded to 10.9.9 on Windows 11 using the Windows installer. I have now started to get Playback Error: The client isn't compatible with the media and the server isn't sending a compatible media format but only on SOME of the files.

Interestingly, the files play just fine through Infuse.

### Reproduction steps

1. Sign in to jellyfin.
2. Play a video.
3. Playback error for some videos.

### What is the current _bug_ behavior?

Playback error

### What is the expected _correct_ behavior?

The video plays

### Jellyfin Server version

10.9.9+

### Specify the build version

10.9.9

### Environment

```markdown
- OS: Windows 11
- Linux Kernel:
- Virtualization:
- Clients: Browser
- Browser: Chrome 127
- FFmpeg Version:
- Playback Method: Not sure
- Hardware Acceleration: NVIDIA NVENC
- GPU Model: RTX 3060
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

```shell
[2024-08-08 10:51:49.151 -05:00] [ERR] [21] Jellyfin.Api.Middleware.ExceptionMiddleware: Error processing request. URL "GET" "/videos/10a275fe-aa0b-8570-f6bc-0005367884b9/hls1/main/0.ts".
System.ArgumentNullException: Value cannot be null. (Parameter '_mediaEncoder.EncoderPath')
   at System.ArgumentNullException.Throw(String paramName)
   at System.ArgumentNullException.ThrowIfNull(Object argument, String paramName)
[...]

Issue #11696, opened 2024-05-17
[Question]: New media files are not being played.


rohitghali:
### Please describe your bug

New media doesn't play. Old media works just fine. New ones show up in the server, but won't play.

### Reproduction Steps

Upgraded from 10.8 to 10.9.1. New media files are not being played. They're being scanned and updated fine, but playback is not working. 

### Jellyfin Version

10.9.0

### Environment

```markdown
- OS: Windows 11
- Linux Kernel:
- Virtualization:
- Clients: Android app, web browser, findroid.
- Browser:
- FFmpeg Version:
- Playback Method: Any method. Jellyfin web, android app, findroid etc.
- Hardware Acceleration: Yes
- GPU Model: 
- Plugins: Basic ones enabled
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

```shell
[2024-05-17 10:42:33.497 -04:00] [INF] [17] MediaBrowser.MediaEncoding.Encoder.MediaEncoder: Starting null with args "-analyzeduration 200M -probesize 1G -i file:\"MediaFilePathName.mkv\" -threads 0 -v warning -print_format json -show_streams -show_chapters -show_format"
[2024-05-17 10:42:33.501 -04:00] [ERR] [17] MediaBrowser.Providers.Movies.MovieMetadataService: Error in "Probe Provider"
System.InvalidOperationException: Cannot start process because a file name has not been provided.
   at System.Diagnostics.Process.Start()
   at MediaBrowser.MediaEncoding.Encoder.MediaEncoder.StartProcess(ProcessWrapper process)
   at MediaBrowser.MediaEncoding.Encoder.MediaEncoder.GetMediaInfoInternal(String inputPath, String primaryPath, MediaProtocol protocol, Boolean extractChapters, String probeSizeArgument, Boolean isAudio, Nullable`1 videoType, CancellationToken cancellationToken)
   at MediaBrowser.MediaEncoding.Encoder.MediaEncoder.GetMediaInfoInternal(String inputPath, String primaryPath, MediaProtocol protocol, Boolean extractChapters, String probeSizeArgument, Boolean isAudio, Nullable`1 videoType, CancellationToken cancellationToken)
   at MediaBrowser.Providers.MediaInfo.FFProbeVideoInfo.ProbeVideo[T](T item, MetadataRefreshOptions options, CancellationToken cancellationToken)
[...]

Issue #11577, opened 2024-05-12
[Issue]: Cant play new Media after updating Jellyfin Server to 10.9


stumpyofpain:
### Please describe your bug

After Updating Jellyfin to Version 10.9 on Ubuntu, old media can be played without any problems, but new Media cant be played.
The Jellyfin log shows :

```
"Error in "Probe Provider"
System.InvalidOperationException: Cannot start process because a file name has not been provided."
```

There are no ffmpeg logs, as these playbacks didnt even start.

After downgrading back to 10.8.13 the media can be played without any problems.

### Reproduction Steps

1. Update to 10.9
2. Add new Media
3. Start the new Media (tested with Jellyfin Web and Jellyfin Android TV)
4. The players both cant play the media

### Jellyfin Version

10.9.0

### Environment

```markdown
- OS: Ubuntu 22.04.4 LTS
- Linux Kernel: GNU/Linux 6.2.16-15-pve x86_64
- Virtualization: Proxmox LXC
- Clients: Jellyfin Web & Jellyfin Android TV
- Browser: Brave 1.65.132
- FFmpeg Version: 6?
- Playback Method: Transcode
- Hardware Acceleration: no
- GPU Model: none
- Plugins: aniDB, AniList, Anime, AudioDB, Fanart, Merge Versions, MusicBrainz, OMDb, Open Subtitles, Playback Reporting, Reports, SkinManager, Studio Images, TMDb, TMDb Box Sets, TVmaze, TheTVDB, Theme Songs, Trakt
- Reverse Proxy: NGINX, but it was played Local
- Base URL: /jellyfin
- Networking: vnic
- Storage: local (config) + NFS (media)
```

### Jellyfin logs

```shell
[2024-05-12 16:55:57.293 +00:00] [INF] Processing playback tracker : "TW96aWxsYS81LjAgKFdpbmRvd3MgTlQgMTAuMDsgV2luNjQ7IHg2NCkgQXBwbGVXZWJLaXQvNTM3LjM2IChLSFRNTCwgbGlrZSBHZWNrbykgQ2hyb21lLzEyMy4wLjAuMCBTYWZhcmkvNTM3LjM2fDE3MTEzNDQwOTU4MTc1-d355f8afcab14db18d1c8a9c8ffd1361-60a8dac61b6c997695690b295727f7ab"
[...]

PR #12463, opened 2024-08-17
Prevent server from starting if the ffmpeg path is invalid

gnattu:
We've received numerous reports from users that Jellyfin is failing to play certain files, and it turns out that many of these issues are due to invalid FFmpeg configurations, such as incorrect installation, stale configuration values, or an unsupported FFmpeg version. The server currently still starts in this situation, leading users to mistakenly believe it's a server bug rather than a configuration error.

This PR makes the server be prevented from starting and print a helper text to help the user fix the installation if the FFmpeg path is not found, not supported, or not executable.

This also removed the internal logic to update the config file's ffmpeg path as we are no longer using that.

For debug purposes that want to run jellyfin without ffmpeg, set the env var `JELLYFIN_FFMPEG__NOVALIDATION=true` to bypass this check.

**Changes**

**Issues**

Some of the reports that turns out to be a config/installation issue. These might not be all of them:

#11543
#11577
#11696
#12042
#12411
#12458

### DIFF
--- a/Emby.Server.Implementations/ApplicationHost.cs
+++ b/Emby.Server.Implementations/ApplicationHost.cs
@@ -402,7 +402,12 @@ namespace Emby.Server.Implementations
             ConfigurationManager.ConfigurationUpdated += OnConfigurationUpdated;
             ConfigurationManager.NamedConfigurationUpdated += OnConfigurationUpdated;
 
-            Resolve<IMediaEncoder>().SetFFmpegPath();
+            var ffmpegValid = Resolve<IMediaEncoder>().SetFFmpegPath();
+
+            if (!ffmpegValid)
+            {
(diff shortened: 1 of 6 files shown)


Shadowghost (approved):

jumoog (approved):

crobibero (approved):

cvium (approved):

Bond-009:
I'd prefer using `Microsoft.Extensions.Configuration` and use the same key structure as the other ffmpeg related options.
`ffmpeg:validate` or `ffmpeg:novalidation` or something else.

gnattu:
> `Microsoft.Extensions.Configuration`

Changed to use that, and now the env var requires `JELLYFIN_FFMPEG__NOVALIDATION=true` due to the requirements of `Microsoft.Extensions.Configuration`

PR #12482, opened 2024-08-20
Set segmentFormat for mpegts container

lomion0815:
Fixes the need to disable "Prefer fMP4-HLS Media Container" when using the TVHeadend plugin (and playback of a TS recording) as part of the issue https://github.com/jellyfin/jellyfin-plugin-tvheadend/issues/89. 

**Changes**
The current code sets the segmentFormat to mpegts for container "ts". This PR also includes container "mpegts".

**Issues**
Backend part of https://github.com/jellyfin/jellyfin-plugin-tvheadend/issues/89


lomion0815:
Sorry have to use release-10.9.x as a base.

PR #12483, opened 2024-08-20
Set segmentFormat for mpegts container

lomion0815:
Fixes the need to disable "Prefer fMP4-HLS Media Container" when using the TVHeadend plugin (and playback of a TS recording) as part of the issue https://github.com/jellyfin/jellyfin-plugin-tvheadend/issues/89. 

**Changes**
The current code sets the segmentFormat to mpegts for container "ts". This PR also includes container "mpegts".

**Issues**
Backend part of https://github.com/jellyfin/jellyfin-plugin-tvheadend/issues/89

### DIFF
--- a/Jellyfin.Api/Controllers/DynamicHlsController.cs
+++ b/Jellyfin.Api/Controllers/DynamicHlsController.cs
@@ -1569,7 +1569,7 @@ public class DynamicHlsController : BaseJellyfinApiController
         var segmentContainer = outputExtension.TrimStart('.');
         var inputModifier = _encodingHelper.GetInputModifier(state, _encodingOptions, segmentContainer);
 
-        if (string.Equals(segmentContainer, "ts", StringComparison.OrdinalIgnoreCase))
+        if (string.Equals(segmentContainer, "ts", StringComparison.OrdinalIgnoreCase) || string.Equals(segmentContainer, "mpegts", StringComparison.OrdinalIgnoreCase))
         {
             segmentFormat = "mpegts";
         }


Bond-009 (approved):

Shadowghost (approved):

gnattu:
Can you provide a more specific reason that why this plugin cannot set the segment container to be ts like everyone else? Generally we use ts and mp4 exclusively as two supported containers for hls and I personally don't like to have something that makes exceptions like this.

lomion0815:
I need to set the hls_segment_type to mpegts (ffmpeg only accepts fmp4 and mpegts). If container = ts at the API does the trick I will change the plugin accordingly.

gnattu:
It *should* do the trick as all of the api requesting mpegts is using ts as segment container.

lomion0815:
Will test it later today.

lomion0815:
No ... setting the field "Container" in MediaSourceInfo to 'ts' results in hls_segment_type fmp4!

```
"/usr/lib/jellyfin-ffmpeg/ffmpeg" "-analyzeduration 2000000 -probesize 1G -init_hw_device vaapi=va:,kernel_driver=i915,driver=iHD -init_hw_device qsv=qs@va -filter_hw_device qs -noautorotate -i \"http://192.168.0.8:9981/stream/channelid/578199168?ticket=c636a4d4a4671e5d763898858c1897158327dfee\" -map_metadata -1 -map_chapters -1 -threads 2 -sn -codec:v:0 h264_qsv -preset veryfast -look_ahead 0 -b:v 14616000 -maxrate 14616000 -bufsize 29232000 -g:v:0 150 -keyint_min:v:0 150 -vf \"setparams=color_primaries=bt709:color_trc=bt709:colorspace=bt709,yadif=0:-1:0,scale=trunc(min(max(iw\,ih*a)\,1920)/2)*2:trunc(ow/a/2)*2,format=nv12\" -codec:a:0 libfdk_aac -ab 384000 -copyts -avoid_negative_ts disabled -max_muxing_queue_size 2048 -f hls -max_delay 5000000 -hls_time 3 -hls_segment_type fmp4 -hls_fmp4_init_filename \"e08d58c47ed9984bc4b5e597fbffb9b7-1.mp4\" -start_number 0 -hls_base_url \"hls/e08d58c47ed9984bc4b5e597fbffb9b7/\" -hls_segment_filename \"/transcodes/e08d58c47ed9984bc4b5e597fbffb9b7%d.mp4\" -hls_playlist_type event -hls_list_size 0 -y \"/transcodes/e08d58c47ed9984bc4b5e597fbffb9b7.m3u8\""
```

This causes not working playback in Firefox at least.

lomion0815:
The plugin uses the interface ILiveTvService with which I can provide a MediaSourceInfo class but can not provide nor control the streaming request directly.

gnattu:
But how will adding a check for `mpegts` make it work? This endpoint always defaults to setting `segmentFormat` to `mpegts` when the query parameter is not recognized. So in your Firefox case, where it still uses `mp4`, it's very likely that `mp4` is being requested explicitly. This change will not be helpful.

gnattu:
Can you check that is your LiveStreamId null or empty in the MediaSourceInfo? If that field is not null or empty the server should inject compatibility profile automatically so that you don't have to do extra hacking here

lomion0815:
Just now I was also looking at "LiveStreamId" which would trigger 
```
// Remove all fmp4 transcoding profiles, because it causes playback error and/or A/V sync issues
// Notably: Some channels won't play on FireFox and LG webOS
// Some channels from HDHomerun will experience A/V sync issues
```
quote from StreamingHelpers.cs line 152 ... which is exactly what I need. But I don't know what LiveStreamId is intended for?

I will give it a try.

lomion0815:
Not working
```
Jellyfin.Api.Middleware.ExceptionMiddleware: Error processing request. URL "POST" "/Sessions/Playing".
MediaBrowser.Common.Extensions.ResourceNotFoundException: Exception of type 'MediaBrowser.Common.Extensions.ResourceNotFoundException' was thrown.
```
LiveStreamId should point to an existing stream.
```
            var liveStreamInfo = _mediaSourceManager.GetLiveStreamInfo(streamingRequest.LiveStreamId);
            if (liveStreamInfo is null)
            {
                throw new FileNotFoundException();
            }
```

lomion0815:
What about changing this PR to
```
        if (string.Equals(segmentContainer, "ts", StringComparison.OrdinalIgnoreCase) || string.Equals(state.MediaSource.Container, "ts", StringComparison.OrdinalIgnoreCase) )
```
(not tested yet).
This way it is also controllable by the plugin ... monkey patch?

Other than that we could only try to move the logic into the client, like the comment mentioned above says the problem only occur in certain setups "FireFox and LG webOS" ... the "Some channels from HDHomerun will experience A/V sync issues" will be more difficult to tackle.

gnattu:
The current server is expecting that the mediasource is an opened livestream to think it is a livetv session and injecting compatibility profiles. The `GetPostedPlaybackInfo` endpoint will open automatically if `autoOpenLiveStream` is true. For requests not using this endpoint, the client should manually call `LiveStreams/Open` to open one if `mediaSource.RequiresOpening` is true, and the provider like the plugin should maintain `mediaSource.OpenToken` and `mediaSource.LiveStreamId`. I'm not quite sure which part of these are currently implemented in the plugin or the clients, and it seems like the plugin needs a little bit more work to make it work as our intended. 

> This way it is also controllable by the plugin ... monkey patch?

If we want it to be solely controllable with plugins without changing the API or the core logic of how the plugin is implemented, our only option is to use a monkey patch, in this case, we need to define a streamtype that does not requires opening but can be reliably treat as a livestream, and your approach of solely defining it with container is way too hacky.

I'm currently thinking for `IsRemote==true`, `RunTimeTicks is null` and any of `IgnoreDts`, `IgnoreIndex`, `GenPtsInput` is true. I hope this will not cause more regressions.

lomion0815:
Quote: I'm currently thinking for IsRemote==true, RunTimeTicks is null and any of IgnoreDts, IgnoreIndex, GenPtsInput is true. I hope this will not cause more regressions. => Did not work.

I now looked into setting TranscodingContainer to "ts" but this is overwritten by the "trancodingProfile" at StreamBuilder.cs line 587. Very frustrating ...

I can not decide this on the client side because if decoding works depends on the input type (if ts input, the segment type must be ts otherwise fmp4 is fine as well). This is not related to jellyfin but can be shown by a native ffmpeg transcoding.

gnattu:
> Did not work

This is because we *need* to implement a new mechanisms for that if we can agree on this and use that to replace your PR because this PR would not work anyway

lomion0815:
You know the backends innerworkings the best. Whatever you see best for supporting this kind of controlling behaviour from within a plugin.
In the meantime I would add a remark to the README of jellyfin-plugin-tvheadend to mention that you need to disable "Prefer fMP4-HLS Media Container".

gnattu:
Closing as we are not making more 10.9 releases. For this use case, plugins should use the new field `UseMostCompatibleTranscodingProfile` introduced by #12529 in 10.10.

PR #12490, opened 2024-08-21
Set Content-Disposition header to attachment for image endpoints

nielsvanvelzen:
The server serves images in various endpoints, those are meant to be displayed in the various clients and not for sharing. Setting the content disposition header to `attachment` will force a browser to download the images when opened directly instead of displaying them in the browser.

**Changes**
- Update item image endpoints
- Update plugin image endpoint
- Update user image endpoints
- Update splashscreen image endpoint

**Issues**

### DIFF
--- a/Jellyfin.Api/Controllers/ImageController.cs
+++ b/Jellyfin.Api/Controllers/ImageController.cs
@@ -2089,6 +2089,8 @@ public class ImageController : BaseJellyfinApiController
         Response.Headers.Append(HeaderNames.Age, Convert.ToInt64((DateTime.UtcNow - dateImageModified).TotalSeconds).ToString(CultureInfo.InvariantCulture));
         Response.Headers.Append(HeaderNames.Vary, HeaderNames.Accept);
 
+        Response.Headers.ContentDisposition = "attachment";
+
         if (disableCaching)
         {
             Response.Headers.Append(HeaderNames.CacheControl, "no-cache, no-store, must-revalidate");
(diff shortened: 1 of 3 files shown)


ferferga:
No idea how it's done in C# atm, but I'd refactor this into an annotation so it can be reused more easily across controllers?

ferferga (commented):
No idea how it's done in C# atm, but I'd refactor this into an annotation so it can be reused more easily across controllers?

nielsvanvelzen:
All image responses are already going through the `GetImageResult` method in ImageController, only exceptions being the plugin/trickplay images right now. We could improve this a bit so we also add caching headers to trickplay/plugin but that would be an enhancement for a non-patch release.

crobibero (approved):

HimbeersaftLP:
What exactly is the advantage of this? I see Content-Disposition: Attachment used less and less on the web today, because most users find it annoying if they open an image in a new tab and it doesn't open in a new tab as expected but downloads instead.

nielsvanvelzen:
As mentioned in the description:

> The server serves images in various endpoints, those are meant to be displayed in the various clients and not for sharing.

We don't want images to be opened directly in the browser.

Techman:
I understand the reasoning behind this, but it does make saving full-size thumbnails of stuff a little more inconvenient.

Sometimes, Jellyfin picks a show I like to generate a library thumbnail and I save it manually in case I want to keep that image or go back to it in the future.

Perhaps we should have an editor in the UI that allows us to click a button to generate thumbnails and then keep one that we like as a static option.

PR #12496, opened 2024-08-21
Added external search provider

JPVenson:
Added Plugin interface and implementation for externally provided search providers

### DIFF
--- a/Emby.Server.Implementations/ApplicationHost.cs
+++ b/Emby.Server.Implementations/ApplicationHost.cs
@@ -66,6 +66,7 @@ using MediaBrowser.Controller.Playlists;
 using MediaBrowser.Controller.Providers;
 using MediaBrowser.Controller.QuickConnect;
 using MediaBrowser.Controller.Resolvers;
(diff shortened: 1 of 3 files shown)


cvium (dismissed):

DomiStyle:
I'm not entirely sure how the plugin system works in Jellyfin but I took a quick look at the interface here:

https://github.com/jellyfin/jellyfin/blob/260a9baeb117de864196e71802ea21d0cc342fc6/MediaBrowser.Controller/SearchExtension/IBaseItemSearchProvider.cs#L9-L18

It looks like there's no `Start` and `Stop` method for when Jellyfin starts/stops, which would be needed for some search providers since they may need to open and close database connections.

Another thing that comes to mind is handling reindexing when media is added or removed or when scans are done. Is there an event bus that can be subscribed to, to catch these events or should these events be added to the interface?

JPVenson:
Pr gets postponed until the EFCore refactor is done and then gets implemented as a Decorator for the IItemRepository

PR #12499, opened 2024-08-22
Apply all codec conditions

dmitrylyzo:
We use all codec conditions to test for DirectPlay, but only apply those from the first codec profile.

To maintain backward compatibility with clients, we reverse the codec profiles so that the last (actually the first) codec profile sets the final values for non-numeric attributes.

**Changes**
Apply all codec conditions.

**Issues**
Fixes https://github.com/jellyfin/jellyfin-androidtv/issues/2516
Fixes https://github.com/jellyfin/jellyfin-androidtv/issues/3388


nielsvanvelzen (approved):
✅ This fixes the "max 1080p" conditions for the Android TV app

crobibero (approved):

Shadowghost (approved):

Bond-009 (approved):

PR #12509, opened 2024-08-24
Remove "try out" functionality from built-in swagger-ui

nielsvanvelzen:
**Changes**

- Remove "try out" functionality from built-in swagger-ui
- Display operation ids
- Hide authorize button and server selection

**Issues**

### DIFF
--- a/Jellyfin.Server/Extensions/ApiApplicationBuilderExtensions.cs
+++ b/Jellyfin.Server/Extensions/ApiApplicationBuilderExtensions.cs
@@ -1,9 +1,11 @@
+using System;
 using System.Collections.Generic;
 using Jellyfin.Api.Middleware;
 using MediaBrowser.Common.Net;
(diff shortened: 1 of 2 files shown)


crobibero:
While I understand why this change is being proposed, I regularly use the “try it out” functionality

nielsvanvelzen:
When you use it, do you then manually input `MediaBrowser token="xxx"` into the authorize field?

crobibero:
Correct, that is exactly what I do

sjorge:
Can we perhaps use the custom.css file to fill in the correct value (sans token ofcourse) ?

```css
.somefieldclass:before {
  content: `MediaBrowser token="<tokenhere>"`;
}
```

Should theoretically work.

soultaco83:
I too have used this multiple times. Manually inputting Mediabrowser Token="" in the auth field

PR #1136, opened 2019-03-19
Require access type to be included in bug report

JustAMan:
**Changes**
Inspired by https://github.com/jellyfin/jellyfin/issues/1085#issuecomment-473833591

It seems that the issue with "setup wizard" described there is only prominent when reverse-proxied, not when accessed directly. So this type of information should be gathered in the bug report as well.

Thus I'm adding another field to the bug report template to gather such information.

PR #12511, opened 2024-08-25
Add FamilyName field to FontFile class and /FallbackFont/Fonts API call

p0358:
**Changes**

This adds the FamilyName field to the fallback fonts API, which is required to be able to load fonts on-demand in client.

Currently the way fallback fonts work is that they load **all** the fonts into browser before starting playback (up to the arbitrary 20 MB limit), which for obvious reasons sucks. Subtitle Octopus library has a feature where you can define a map of font family to URL to have it only load the required fonts used by particular .ASS file on-demand instead of preloading them all and then figuring out what's gonna be used, which is infinitely better.

However I currently create this as draft PR, because the current naive simple approach requires opening all the font files, which takes 10+ seconds with 1000 font files and on HDD (which is my real use-case that I have right now...). At the very least probably thus the font file path to font family name should be cached somewhere, and possibly precached on both server startup and when the font preloading preference is turned on in admin settings during runtime, if it was off before. I would welcome feedback on what would be the best way of going about this...

Btw it adds a dependency on new library, according to my research it's the only pain-free and cross-platform library for fonts that also supports all the required formats. It's dual-licensed, but Jellyfin as open-source projects falls into the allowance of using Apache2 license of it. So hopefully that's okay?

Also future work TODO: should also make fallback fonts work with transcoding (afaik currently they're completely ignored for burn-in? but that's a matter for separate issue and PR; related to #7275 that currently only supports subtitles from MKV attachments afaik)

**Issues**

didn't create one for this :/

related jellyfin-web PR that follows this: jellyfin/jellyfin-web#5960

I also have hopes to add support for this to some other clients afterwards in some way...

### DIFF
--- a/Emby.Server.Implementations/Localization/Core/en-US.json
+++ b/Emby.Server.Implementations/Localization/Core/en-US.json
@@ -80,6 +80,8 @@
     "SubtitleDownloadFailureFromForItem": "Subtitles failed to download from {0} for {1}",
     "Sync": "Sync",
     "System": "System",
+    "TaskSubtitleFallbackFontListPrecache": "Preload fallback subtitles font names",
+    "TaskSubtitleFallbackFontListPrecacheDescription": "Preloads the font family names for fallback subtitles fonts",
     "TvShows": "TV Shows",
     "Undefined": "Undefined",
     "User": "User",
--- a/MediaBrowser.Providers/Subtitles/SubtitleManager.cs
+++ b/MediaBrowser.Providers/Subtitles/SubtitleManager.cs
@@ -1,14 +1,19 @@
 #pragma warning disable CS1591
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
+using System.IO.Hashing;
 using System.Linq;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Jellyfin.Extensions;
+using MediaBrowser.Common.Configuration;
 using MediaBrowser.Common.Extensions;
+using MediaBrowser.Controller.Configuration;
 using MediaBrowser.Controller.Entities;
 using MediaBrowser.Controller.Entities.Movies;
 using MediaBrowser.Controller.Entities.TV;
@@ -21,7 +26,9 @@ using MediaBrowser.Model.Entities;
 using MediaBrowser.Model.Globalization;
 using MediaBrowser.Model.IO;
 using MediaBrowser.Model.Providers;
+using MediaBrowser.Model.Subtitles;
 using Microsoft.Extensions.Logging;
+using SixLabors.Fonts;
 
 namespace MediaBrowser.Providers.Subtitles
 {
@@ -31,15 +38,19 @@ namespace MediaBrowser.Providers.Subtitles
         private readonly IFileSystem _fileSystem;
         private readonly ILibraryMonitor _monitor;
         private readonly IMediaSourceManager _mediaSourceManager;
+        private readonly IServerConfigurationManager _serverConfigurationManager;
         private readonly ILocalizationManager _localization;
 
         private readonly ISubtitleProvider[] _subtitleProviders;
+
+        private readonly ConcurrentDictionary<ulong, string> _cachedFontFamilyNames = new();
 
         public SubtitleManager(
             ILogger<SubtitleManager> logger,
             IFileSystem fileSystem,
             ILibraryMonitor monitor,
             IMediaSourceManager mediaSourceManager,
+            IServerConfigurationManager serverConfigurationManager,
             ILocalizationManager localizationManager,
             IEnumerable<ISubtitleProvider> subtitleProviders)
         {
@@ -47,6 +58,7 @@ namespace MediaBrowser.Providers.Subtitles
             _fileSystem = fileSystem;
             _monitor = monitor;
             _mediaSourceManager = mediaSourceManager;
+            _serverConfigurationManager = serverConfigurationManager;
             _localization = localizationManager;
             _subtitleProviders = subtitleProviders
                 .OrderBy(i => i is IHasOrder hasOrder ? hasOrder.Order : 0)
@@ -409,6 +421,73 @@ namespace MediaBrowser.Providers.Subtitles
                     Name = i.Name,
                     Id = GetProviderId(i.Name)
                 }).ToArray();
+        }
+
+        public IEnumerable<FontFile> GetFallbackFontList(IProgress<double>? progress, CancellationToken cancellationToken)
+        {
+            var encodingOptions = _serverConfigurationManager.GetEncodingOptions();
+            var fallbackFontPath = encodingOptions.FallbackFontPath;
+
+            if (!string.IsNullOrEmpty(fallbackFontPath))
+            {
+                var files = _fileSystem.GetFiles(fallbackFontPath, new[] { ".woff", ".woff2", ".ttf", ".otf" }, false, false);
+                var fontFiles = files
+                    .Select(i => new FontFile
+                    {
+                        Name = i.Name,
+                        Size = i.Length,
+                        DateCreated = _fileSystem.GetCreationTimeUtc(i),
+                        DateModified = _fileSystem.GetLastWriteTimeUtc(i)
+                    })
+                    .OrderBy(i => i.Size)
+                    .ThenBy(i => i.Name)
+                    .ThenByDescending(i => i.DateModified)
+                    .ThenByDescending(i => i.DateCreated)
+                    .ToList();
+                _logger.LogDebug("We have {Count} fallback font files", fontFiles.Count);
+
+                var index = 0;
+
+                foreach (var fontFile in fontFiles)
+                {
+                    cancellationToken.ThrowIfCancellationRequested();
+
+                    if (progress is not null)
+                    {
+                        progress.Report(100D * (index + 1) / fontFiles.Count);
+                    }
+
+                    var crc = new Crc64();
+                    crc.Append(Encoding.UTF8.GetBytes(fontFile.Name ?? string.Empty));
+                    crc.Append(BitConverter.GetBytes(fontFile.Size));
+                    crc.Append(BitConverter.GetBytes(fontFile.DateCreated.ToBinary()));
+                    crc.Append(BitConverter.GetBytes(fontFile.DateModified.ToBinary()));
+                    var hash = crc.GetCurrentHashAsUInt64();
+
+                    string? familyName;
+                    if (_cachedFontFamilyNames.TryGetValue(hash, out familyName))
+                    {
+                        fontFile.FamilyName = familyName;
+                    }
+                    else
+                    {
+                        FontCollection collection = new();
+                        FontFamily family = collection.Add(fallbackFontPath + "/" + fontFile.Name, CultureInfo.InvariantCulture);
+                        fontFile.FamilyName = family.Name;
+                        _cachedFontFamilyNames.TryAdd(hash, fontFile.FamilyName);
+                        _logger.LogDebug("Fallback subtitles font file \"{FileName}\" has family name \"{FamilyName}\"", fontFile.Name, fontFile.FamilyName);
+                    }
+
+                    index++;
+
+                    yield return fontFile;
+                }
+            }
+            else
+            {
+                _logger.LogWarning("The path of fallback font folder has not been set");
+                encodingOptions.EnableFallbackFont = false;
+            }
         }
     }
 }
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -66,6 +66,7 @@
     <PackageVersion Include="Serilog.Sinks.Graylog" Version="3.1.1" />
     <PackageVersion Include="SerilogAnalyzer" Version="0.15.0" />
     <PackageVersion Include="SharpFuzz" Version="2.1.1" />
+    <PackageVersion Include="SixLabors.Fonts" Version="2.0.4" />
     <PackageVersion Include="SkiaSharp" Version="2.88.8" />
     <PackageVersion Include="SkiaSharp.HarfBuzz" Version="2.88.8" />
     <PackageVersion Include="SkiaSharp.NativeAssets.Linux" Version="2.88.8" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -76,6 +76,7 @@
     <PackageVersion Include="Swashbuckle.AspNetCore.ReDoc" Version="6.5.0" />
     <PackageVersion Include="Swashbuckle.AspNetCore" Version="6.2.3" />
     <PackageVersion Include="System.Globalization" Version="4.3.0" />
+    <PackageVersion Include="System.IO.Hashing" Version="8.0.0" />
     <PackageVersion Include="System.Linq.Async" Version="6.0.1" />
     <PackageVersion Include="System.Text.Encoding.CodePages" Version="8.0.0" />
     <PackageVersion Include="System.Text.Json" Version="8.0.4" />
--- a/Emby.Server.Implementations/ScheduledTasks/Tasks/SubtitleFallbackFontListPrecacheTask.cs
+++ b/Emby.Server.Implementations/ScheduledTasks/Tasks/SubtitleFallbackFontListPrecacheTask.cs
@@ -1,0 +1,83 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using MediaBrowser.Controller.Configuration;
+using MediaBrowser.Controller.Subtitles;
+using MediaBrowser.Model.Globalization;
+using MediaBrowser.Model.IO;
+using MediaBrowser.Model.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Emby.Server.Implementations.ScheduledTasks.Tasks;
+
+/// <summary>
+/// Precaches the mapping of fallback font file to font family name.
+/// </summary>
+public class SubtitleFallbackFontListPrecacheTask : IScheduledTask, IConfigurableScheduledTask
+{
+    private readonly ILocalizationManager _localization;
+    private readonly IServerConfigurationManager _serverConfigurationManager;
+    private readonly ISubtitleManager _subtitleManager;
+    private readonly IFileSystem _fileSystem;
+    private readonly ILogger<SubtitleFallbackFontListPrecacheTask> _logger;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="SubtitleFallbackFontListPrecacheTask"/> class.
+    /// </summary>
+    /// <param name="localization">Instance of the <see cref="ILocalizationManager"/> interface.</param>
+    /// <param name="serverConfigurationManager">Instance of the <see cref="IServerConfigurationManager"/> interface.</param>
+    /// <param name="subtitleManager">Instance of the <see cref="ISubtitleManager"/> interface.</param>
+    /// <param name="logger">The logger.</param>
+    /// <param name="fileSystem">The filesystem.</param>
+    public SubtitleFallbackFontListPrecacheTask(
+        ILocalizationManager localization,
+        IServerConfigurationManager serverConfigurationManager,
+        ISubtitleManager subtitleManager,
+        IFileSystem fileSystem,
+        ILogger<SubtitleFallbackFontListPrecacheTask> logger)
+    {
+        _localization = localization;
+        _serverConfigurationManager = serverConfigurationManager;
+        _subtitleManager = subtitleManager;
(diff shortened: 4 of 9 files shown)


dmitrylyzo:
Maybe do it during startup?

p0358:
I hope adding a whole new task for this isn't an overkill, but it's maybe justified since it could take between 8 to 48 seconds to load the filenames up on my system. Added the task as hidden, since imo there isn't really any justification to run it again once the server has started up, as any new/changed font files will be dynamically picked up on the API request (because listing directory is cheap and I made the cache take into account size and file dates), so hiding it will at least declutter the UI a bit (and hey, the IsHidden property exists for a reason I guess).

Also moved the func to get fallback font list to SubtitleManager, so that it can be accessed from both task and API, in the future maybe also from transcoding tasks too, so the move was probably justified anyways.

I didn't end up adding task fire upon settings save, since I figured that it'd be only ran for the first time when admin sets it up, probably with not that many fonts at once, and thus they can eat the cost of it during the first API request where they'll probably test it immediately afterwards.

MediaBrowser.Providers/Subtitles/SubtitleManager.cs
```diff
@@ -31,22 +37,27 @@ public class SubtitleManager : ISubtitleManager
         private readonly IFileSystem _fileSystem;
         private readonly ILibraryMonitor _monitor;
         private readonly IMediaSourceManager _mediaSourceManager;
+        private readonly IServerConfigurationManager _serverConfigurationManager;
         private readonly ILocalizationManager _localization;
 
         private readonly ISubtitleProvider[] _subtitleProviders;
 
+        private readonly Dictionary<ulong, string> _cachedFontFamilyNames = new();
```
crobibero, line 45: This cache isn't thread safe

crobibero (commented):
Please revert changes to `en-GB.json` and `pl.json` as they are managed by Weblate

Emby.Server.Implementations/Localization/Core/en-US.json
```diff
@@ -131,5 +131,7 @@
     "TaskKeyframeExtractor": "Keyframe Extractor",
     "TaskKeyframeExtractorDescription": "Extracts keyframes from video files to create more precise HLS playlists. This task may run for a long time.",
     "TaskCleanCollectionsAndPlaylists": "Clean up collections and playlists",
-    "TaskCleanCollectionsAndPlaylistsDescription": "Removes items from collections and playlists that no longer exist."
+    "TaskCleanCollectionsAndPlaylistsDescription": "Removes items from collections and playlists that no longer exist.",
```
dmitrylyzo, line 134: I think new keys should be placed in alphabetical order to minimise conflicts.
Since the first part is in order, I suggest placing them after `"System"`.

p0358: Hmm, wouldn't this make it look pretty out of place compared to existing keys? I mean I can do it

nyanmisaka:
> Also future work TODO: should also make fallback fonts work with transcoding (afaik currently they're completely ignored for burn-in? but that's a matter for separate issue and PR; related to https://github.com/jellyfin/jellyfin/pull/7275 that currently only supports subtitles from MKV attachments afaik

FFmpeg's subtitles filter can only set one `fontsdir`. So you have to choose between attachments fonts (MKV) and fallback fonts. In addition, it does not support the mapping of font families and font paths, but scans all files in `fontsdir` during initialization. Not sure how much 1000 font files would slow down FFmpeg startup.

https://github.com/jellyfin/jellyfin/blob/ed572467aab96d6149c85dc0d0b406bc6c5106e2/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs#L1452-L1456

gnattu:
I've had discussions with someone with the ideas of downloading a lot of fonts and I'm personally against such idea as the fonts can be very huge and you can easily download fonts larger than you video this way because you can never underestimate how crazy those ass subtitles will use the fonts especially for the CJK ones. The CJK fonts can easily exceed the 20MB limit with only one font and if you want to load multiple of them it is a very easy way to crash your browser tab due to memory limit exceeded especially on mobile phone/TV. The 20 MB limit sucks but it is a safe value so that the browser on those memory limited devices don't crash. 

For server-side rendering we generally don't use the fallback fonts, instead we install the fonts into the system and use the system font provider to render it. It is faster this way and you don't have to do all the dirty things inside Jellyfin because the system will generate a font cache for you.

So basically I really think the scope of this API should be limited and don't make it too fancy due to the complex situation. I'm not really a big fan of letting fallback fonts becoming the primary font source because that against the original intention of making it as a fallback. In my opinion those should be provided in a way similar to the attachment fonts rather than fallback fonts

p0358:
> The 20 MB limit sucks but it is a safe value so that the browser on those memory limited devices don't crash.

The limit could be re-implemented client-side in JS then as a safety measure. But it should be removed from the API, as the client would need a full list of fonts to be able to decide which ones to download (based on which ones it needs in current ASS file), instead of trying to load them all and then hope it contains the needed font faces later.

To be clear about the use-case: it's combining playback of existing MKV files from some general source, that might contain some subtitles and some corresponding fonts for *other languages* (not of interest for the viewer), with *external* ASS files next to it that came with *external* font files. Those external font files are put into fallback fonts directory, allowing the Jellyfin web client to use them properly, all without baking everything into a new MKV file (which would need to be done manually and would prevent the user from ability to e.g. seed the original MKV file).

And that use-case is filled perfectly fine with fallback fonts, apart from the fact that it'd try to load them all into browser, regardless of whether they'd be meant to be used, which in my case is like 1200 fonts. While particular translated series would probably never use more than 5 custom fonts.

> For server-side rendering we generally don't use the fallback fonts, instead we install the fonts into the system and use the system font provider to render it. It is faster this way and you don't have to do all the dirty things inside Jellyfin because the system will generate a font cache for you.

Hmm, I guess that might be an alternate solution for server-side rendering side of the problem. But I think then adding some env var in Docker container to install all fonts from specified directory on startup would be in order. And outside of container it wouldn't be ideal, since fonts installed on the system impact all the programs, but I guess that's something that someone using Jellyfin outside of containerization would have to accept (and/or not be using this many custom fonts either).

p0358:
> For server-side rendering we generally don't use the fallback fonts, instead we install the fonts into the system and use the system font provider to render it. It is faster this way and you don't have to do all the dirty things inside Jellyfin because the system will generate a font cache for you.

So as far as this point itself goes, I poke around and actually if one is using Docker, apparently it's enough to mount a custom font directory into for example `/usr/local/share/fonts/custom` (the `custom` part is arbitrary name, but the primary `/usr/local/share/fonts` folder shouldn't be mounted itself as it also has `.uuid` file there, so a subdir needs to be created.

And once that's done, the fontcache will apparently be automatically updated on container restart. So that just works out of the box with no changes to Jellyfin required at all, and thus that'd at least cover the server subtitle burn-in side without complications.

I might make some additional PR to docs to maybe include something like this in the Docker Compose example:
```yaml
    volumes:
      - type: bind
        source: /path/to/customfonts
        target: /usr/local/share/fonts/custom
        read_only: true
```

p0358:
> So basically I really think the scope of this API should be limited and don't make it too fancy due to the complex situation. I'm not really a big fan of letting fallback fonts becoming the primary font source because that against the original intention of making it as a fallback.

I think it's by all means still a fallback in the sense that MKV attachments or client-side installed fonts will still take precedence if available. These fonts would only be used if neither has the needed font available, and then only such fonts would be streamed.

> In my opinion those should be provided in a way similar to the attachment fonts rather than fallback fonts

So in other words it should be the server deciding which extra fonts are needed for particular subtitles/playback configuration, and then client could download them similarly to attachment fonts? This might not be a bad idea as the target solution if I understand this correctly then. Because then there'd be only a single code and place for parsing the subtitles, instead of the clients having to figure out what's needed.

With that said, I think providing these fonts in a way that can be turned into a map of `font family => URL` for Subtitle Octopus is still beneficial: it can avoid streaming the fonts from server at all if they're currently installed on the client and available to the browser. In fact the exact same problem currently applies to how font attachments are currently handled in the web client: always preloaded from server. So I think ideally there'd be the same approach applied for both in regards to whether client-side fonts should take precedence. This requires answering the question: do we assume that the font is the same if it has the same family name? (imo might be relatively safe to assume so, since eg. localized fonts usually have their family name changed, eg. `A Year Without Rain` to `A Year Without Rain PL` when translators create them)

But I think that at least for now, it'd be good to merge these PRs and at least fix the existing feature with the way it's currently thought out, and I'd appreciate if that could happen.

Btw I'm thinking already how desktop player could implement custom fonts. And while it shouldn't use attachments per se (as I think MPV just plays the MKV directly, thus extracts fonts itself), it could download fallback fonts to some kind of temporary directory and provide it as subtitle font fallback dir to MKV (it has such an option). But for that indeed it'd be nicer if the server eventually provided what fonts are needed for given video instead of it having to download .ass and parse it to figure it out... Different case from Subtitle Octopus that already has the .ass file and list of fonts therein available either way.

p0358:
> The limit could be re-implemented client-side in JS then

And I take it back, it wouldn't actually make sense. We need to pass the full map of subtitles to Subtitle Octopus library in order for it to find out what it needs. We cannot limit this in any way (or we'd end up with missing subtitle entries) if the library doesn't support it after deciding on what subtitles it needs, as only then the total size to be used would be known. Can only count on that it's extremely unlikely for any particular single ASS file to require too many big subtitle files...

The only thing we could reasonably do is to impose some limit on max size of a single subtitle file. Which I did now in jellyfin/jellyfin-web#5960 at https://github.com/jellyfin/jellyfin-web/pull/5960/commits/d821ad3562a55c028fbe3342e821ee31cfc09f83

gnattu:
> Can only count on that it's extremely unlikely for any particular single ASS file to require too many big subtitle files...

This is unfortunately very easy to do for CJK fonts. Just require Noto Sans CJK and it immediately requires > 100MB font, and having a "fallback" font to crash a browser tab is not something we want to do.

We are having different definitions about fallbacks here. At least my definition of a fallback, is the fonts that contains the basic character set so that the characters at least displays instead of just a tofu block. This is the intention of a fallback font, which is not to provide a "font family class" so that preferred fonts can be picked.

The API providing "preferred fonts" should be implemented in a way like the embed fonts not the fallback fonts. And that is my whole point.

dmitrylyzo:
Tbf, embedded fonts can also crash the browser - we don't limit their size. We set some memory limits for libass, but they don't seem to help (sometimes).

> The API providing "preferred fonts" should be implemented in a way like the embed fonts not the fallback fonts. And that is my whole point.

To clarify. Do you mean that the server should parse the subs and generate a set of fonts that could/should be loaded (like a subset of the available fonts from the server)?

gnattu:
> To clarify. Do you mean that the server should parse the subs and generate a set of fonts that could/should be loaded (like a subset of the available fonts from the server)?

*If* we want to implement something like this we should do it this way, instead of passing 1000 fonts to the client.

p0358:
Okay so I looked into SubtitlesOctopus and it seems they have a new option now (since 2022, but that's *after* fallback fonts in Jellyfin were implemented):

```js
self.fallbackFont = options.fallbackFont || 'default.woff2'; // URL to override fallback font, for example, with a CJK one. Default fallback font is Liberation Sans (Optional)
```

This then gets passed to native function `ass_set_fonts` as param `default_font`, making it an explicit fallback. They by default fall to Liberation Sans, which afaik doesn't include CJK.

Now here's my proposal for consideration: what about removing the existing fallback fonts feature in Jellyfin and either replacing it with a single "fallback font" feature, or just supplying a default fallback font that supports CJK (or both, supply and allow changing it)?

The reasoning is that a feature of loading arbitrary number of fonts turns out to be confusing to users about the purpose and original intent of the feature and it predates the designated fallback feature in the underlying library. It currently relies, the way I understand, on libass just scanning all available fonts and picking any random one that contains the needed glyphs instead of displaying a tofu. I'd guess it does that if specified explicit fallback doesn't contain the needed glyphs? The only reason I'd see a single font problematic is if someone wanted to put in a font that has both CJK and emojis, but is that even a really desirable/important use-case? Not sure how popular emojis in ASS outside of MKV are out there, I didn't see any in my life. That'd be the only argument against I see, apart from the fact that it's a breaking change alone.

Then for loading preferred fonts for loose ASS files and streaming them to the clients, we could just use server-side system fonts (with undergoing PR #1136 that aims to explain better how to install them). That would create just a single source of truth and single font bank for both burn-in, thumbnail rendering and font streaming. Meaning users just need to install a font dir just once, in one place and in one way, to eventually get their fonts working everywhere (apart from clients still being sometimes able to install fonts client-side if server doesn't provide fonts, as that depends on the system admin).

@gnattu @dmitrylyzo What do you think, sounds good?

Of course we may proceed with just adding the preferred font streaming with system fonts for now, without touching fallback fonts fro the time being. But seeing that there's a different option now in the underlying library, I can't help but suggest it, seeing that it might make the font situation the least confusing for end-users in the end. I mean can't get it wrong what the "fallback" feature intent is, if there's just one font possible to specify there, right?

(I searched some random threads on the internet out there and it seems that a bunch of people misunderstood the original purpose of fallback fonts and would treat them as a bank of fonts for external ASS like I did at first too)

p0358:
Also to be sure: would using the library `SixLabors.Fonts` be fine? If so, then I'd probably want to contribute usage of fontcache on Linux there to speed up the font operations and avoid having to have a scan task on startup, plus have consistency with something like ffmpeg looking for fonts.

Otherwise we might have to implement some kind of "in-house" font scanning on all three systems. Which isn't to say it's undoable, given the library mentioned just does a scan of known font paths of the systems, optionally using native cache functions only on macOS, and we don't use any of the actual rendering functionality there. But still it'd be less low-level code to maintain in Jellyfin...

I imagine it should be perfectly fine since nobody said no so far, but don't want to find out only after I put in the work to improve it already 🤷

p0358:
Or actually it might turn out to be the best to just wrap libass for both file parsing and to use the exact same font matching logic as the actual library on client would, to deliver the exact file it needs...

gnattu:
> Now here's my proposal for consideration: what about removing the existing fallback fonts feature in Jellyfin and either replacing it with a single "fallback font" feature, or just supplying a default fallback font that supports CJK (or both, supply and allow changing it)?

Single font is bad because "super fonts" containing all CJK glyphs are super huge so we need to allow multi-font fallback.

> The reasoning is that a feature of loading arbitrary number of fonts turns out to be confusing to users about the purpose and original intent of the feature and it predates the designated fallback feature in the underlying library. It currently relies, the way I understand, on libass just scanning all available fonts and picking any random one that contains the needed glyphs instead of displaying a tofu.

This is how it should be to handle multi-font fallback. We moved back to SubtitlesOctopus instead of another library during the development of 10.9 for this exact reason. That one requires the fallback font to contain each and every glyph required by the subtitle to display correctly and that sucks. Because you need a 150MB font to be used as a fallback.

> Also to be sure: would using the library SixLabors.Fonts be fine? 

It looks like their license is not compatible with GPLv2 but I'm not a lawyer, but better not using it before we are certain.

p0358:
Okay I see the point now, someone could want to e.g. load up two fallback fonts for just Japanese and Korean, but not one for Chinese (given Chinese is much bigger), without having to create a custom merged font file. Dropping my single font proposal then.

> It looks like their license is not compatible with GPLv2 but I'm not a lawyer, but better not using it before we are certain.

Hmm, technically their dual-license says that if it's used within an open source project then it's licensed under Apache license, which is compatible with GPL. And GPL means the whole project is then licensed under GPL, which means that the condition of the original dual-license notice that permitted Apache licensing will always be kept, if I understand correctly.

But yeah I'll probably focus on trying to wrap libass first then, should be a superior solution if it works for all these reasons.

gnattu:
> Hmm, technically their dual-license says that if it's used within an open source project then it's licensed under Apache license, which is compatible with GPL.

That part is the problem. The dual license they use is apache 2.0 which is only compatible with GPLv3 but Jellyfin is licensed under GPLv2 which at least GNU does not consider it is compatible:

https://www.gnu.org/licenses/license-list.html#apache2

p0358:
> That part is the problem. The dual license they use is apache 2.0 which is only compatible with GPLv3 but Jellyfin is licensed under GPLv2 which at least GNU does not consider it is compatible:

Ah, didn't notice the version thing, sorry.

---

Btw I now noticed a certain naivety of both me and Subtitle Octopus with its `availableFonts` thing. Basically it won't work for fonts that contain different styles under one common family name... Say `Fontname` in ass was `Times New Roman`, the Octopus would parse the ass file itself, extract that name, and try to preload whatever file is in map for `Times New Roman`.

Meanwhile...

```ini
[V4+ Styles]
Format: Name, Fontname, Fontsize, PrimaryColour, SecondaryColour, OutlineColour, BackColour, Bold, Italic, Underline, StrikeOut, ScaleX, ScaleY, Spacing, Angle, BorderStyle, Outline, Shadow, Alignment, MarginL, MarginR, MarginV, Encoding
Style: Default,Arial,20,&H00FFFFFF,&H000000FF,&H00000000,&H00000000,0,0,0,0,100,100,0,0,1,2,2,2,10,10,10,1
Style: TNR Test 1,Times New Roman,20,&H00FFFFFF,&H000000FF,&H00000000,&H00000000,0,0,0,0,100,100,0,0,1,2,2,2,10,10,10,1
Style: TNR Test 2,Times New Roman,20,&H00FFFFFF,&H000000FF,&H00000000,&H00000000,-1,0,0,0,100,100,0,0,1,2,2,2,10,10,10,1
Style: TNR Test 3,Times New Roman,24,&H00FFFFFF,&H000000FF,&H00000000,&H00000000,-1,-1,-1,-1,103,103,0.1,1,3,2.2,2.2,3,11,11,11,1
```

```
libass: fontselect: (Times New Roman, 400, 0) -> /usr/share/fonts/truetype/msttcorefonts/Times_New_Roman.ttf, 0, TimesNewRomanPSMT
libass: fontselect: (Times New Roman, 700, 0) -> /usr/share/fonts/truetype/msttcorefonts/Times_New_Roman_Bold.ttf, 0, TimesNewRomanPS-BoldMT
libass: fontselect: (Times New Roman, 700, 100) -> /usr/share/fonts/truetype/msttcorefonts/Times_New_Roman_Bold_Italic.ttf, 0, TimesNewRomanPS-BoldItalicMT
```
(3x same `Fontname`, but resolving to three different actual font files)

Actual libass uses a logic where it matches the style (boldness, italicness) together with the family name, resulting in a different file being resolved.

https://github.com/libass/libass/blob/6a759836e5e76bb7b69b0ac244eea76b0d290512/libass/ass_fontselect.c#L735-L746

Basically if Fontname was to equal font's full name or postscript name, such as `Times New Roman Bold Italic` or `TimesNewRomanPS-BoldItalicMT`, it would be selected with maximum score and simply used. But I tested popular software Aegisub to actually insert just `Times New Roman` family name instead with the properties specified.

And so it seems that specifying either of the three names is valid in the file format. But a simple map won't do, so server will probably have to generate a font **array** for given external ASS with what's available on the system rather than a map, and just append it to `fonts` just like existing MKV attachment fonts are.

Annoyingly it seems that on Windows there's not underlying font file path available at least with the current implementation:
```
libass: fontselect: (Times New Roman, 700, 100) -> TimesNewRomanPS-BoldItalicMT, 0, TimesNewRomanPS-BoldItalicMT
```

p0358:
Okay, so I got prototype code within libass working to get ass file -> font paths list working. Unfortunately I had to modify libass itself a bit (or rather dig into unexported functions)... So this would probably mean that a fork should be made and then put under jellyfin org like a bunch of other libraries already are? I presume I'd also put some kind of C# binding/wrapping functions there so that it can be imported as a library and used from server with a simple API?

p0358:
All right, I have a local prototype:
obraz

```
> GET /Fonts/Arial,0,400,0 HTTP/1.1
> Host: 127.0.0.1:8096
> User-Agent: curl/8.7.1
> Accept: */*
>
< HTTP/1.1 200 OK
< Content-Length: 1036584
< Content-Type: font/ttf
< Date: Thu, 24 Oct 2024 02:28:27 GMT
< Server: Kestrel
< Last-Modified: Sat, 07 Dec 2019 09:08:26 GMT
< X-Selected-Font-Path: C:\WINDOWS\FONTS\ARIAL.TTF
< X-Selected-Font-Postscript-Name: ArialMT
< X-Response-Time-ms: 1.5946
<
```

Now I should "only" figure out packaging my helper package with modified libass in a cross-platform way...

Basic request to `/Fonts/` endpoint should be composed of FontName + Vertical + Bold + Italic delimeted by a `,` to reproduce the same parameters that the libass would use when parsing ass. I'm not sure whether the clients should glue that together or whether some DeliveryUrl param should be added, or something like LookupKey for just the glued full name? Idk.

But also the font can be looked up to download by just name, and it can be both family name, postscript name or full name, the latter will then take precedence over just family name, as they're more specific (since in theory the last two should not be repeated between different fonts).

Bond-009:
Closing as stale

PR #12516, opened 2024-08-26
Single audio track selection policy

dmitrylyzo:
Depends on #10756

**Changes**
- Add single audio track selection policy.
- Add `IsDefaultTrack` codec condition.

**Issues**
It looks like we have a bug with detecting secondary audio: browsers skip unsupported streams, and the second audio stream (supported) becomes the first.
Moreover, Chrome 115+ detects only default audio tracks (Chrome bug?).
This can be fixed after complete stream analysis (#10756).


MediaBrowser.Model/Dlna/SingleAudioPolicy.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        /// <summary>
+        /// Selects the track marked 'Default'.
+        /// </summary>
+        Default = 2,
+
+        /// <summary>
+        /// Selects the supported track marked 'Default'.
+        /// </summary>
+        DefaultSupported = 3
```
dmitrylyzo, line 26: These modes can probably be omitted since we can combine `First`/`FirstSupported` with `IsDefaultTrack` condition.

Bond-009:
@dmitrylyzo What's the status of this?

PR #12521, opened 2024-08-26
Fix CodecProfiles and video encoder profiles

nyanmisaka:
**Changes**
- Fix CodecProfiles not taking effect during transcoding

  Fixed the issue where only `CodecProfiles` **matching the source** video codec were applied during transcoding.
  The target codec of the transcoding is not necessarily the same as the source codec.

  Partically reverts #9723,
  because [this line](https://github.com/jellyfin/jellyfin/pull/9723/files#diff-bde118dfa6110f699ead01698c41be0bb9b4b7690a4a0adaf62ad0697ef61d69R749) already fixed the issue described in #9411

---

- Only apply known video encoder profiles

  To avoid passing invalid profiles to ffmpeg.

**Issues**
- https://github.com/jellyfin/jellyfin-androidtv/pull/3933

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -2005,7 +2005,26 @@ namespace MediaBrowser.Controller.MediaEncoding
             }
 
             var profile = state.GetRequestedProfiles(targetVideoCodec).FirstOrDefault() ?? string.Empty;
-            profile = WhiteSpaceRegex().Replace(profile, string.Empty);
+            profile = WhiteSpaceRegex().Replace(profile, string.Empty).ToLowerInvariant();
+
+            var videoProfiles = Array.Empty<string>();
+            if (string.Equals("h264", targetVideoCodec, StringComparison.OrdinalIgnoreCase))
+            {
+                videoProfiles = _videoProfilesH264;
+            }
+            else if (string.Equals("hevc", targetVideoCodec, StringComparison.OrdinalIgnoreCase))
+            {
+                videoProfiles = _videoProfilesH265;
+            }
+            else if (string.Equals("av1", targetVideoCodec, StringComparison.OrdinalIgnoreCase))
+            {
+                videoProfiles = _videoProfilesAv1;
+            }
+
+            if (!videoProfiles.Contains(profile, StringComparison.OrdinalIgnoreCase))
+            {
+                profile = string.Empty;
+            }
 
             // We only transcode to HEVC 8-bit for now, force Main Profile.
             if (profile.Contains("main10", StringComparison.OrdinalIgnoreCase)
--- a/MediaBrowser.Model/Dlna/StreamBuilder.cs
+++ b/MediaBrowser.Model/Dlna/StreamBuilder.cs
@@ -897,7 +897,7 @@ namespace MediaBrowser.Model.Dlna
 
             var appliedVideoConditions = options.Profile.CodecProfiles
                 .Where(i => i.Type == CodecType.Video &&
-                    i.ContainsAnyCodec(videoStream?.Codec, container) &&
+                    i.ContainsAnyCodec(videoCodec, container) &&
                     i.ApplyConditions.All(applyCondition => ConditionProcessor.IsVideoConditionSatisfied(applyCondition, width, height, bitDepth, videoBitrate, videoProfile, videoRangeType, videoLevel, videoFramerate, packetLength, timestamp, isAnamorphic, isInterlaced, refFrames, numVideoStreams, numAudioStreams, videoCodecTag, isAvc)))
                 // Reverse codec profiles for backward compatibility - first codec profile has higher priority
                 .Reverse();
@@ -930,7 +930,7 @@ namespace MediaBrowser.Model.Dlna
 
             var appliedAudioConditions = options.Profile.CodecProfiles
                 .Where(i => i.Type == CodecType.VideoAudio &&
-                    i.ContainsAnyCodec(audioStream?.Codec, container) &&
+                    i.ContainsAnyCodec(audioCodec, container) &&
                     i.ApplyConditions.All(applyCondition => ConditionProcessor.IsVideoAudioConditionSatisfied(applyCondition, audioChannels, inputAudioBitrate, inputAudioSampleRate, inputAudioBitDepth, audioProfile, isSecondaryAudio)))
                 // Reverse codec profiles for backward compatibility - first codec profile has higher priority
                 .Reverse();


MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -897,7 +897,7 @@ private void BuildStreamVideoItem(
 
             var appliedVideoConditions = options.Profile.CodecProfiles
                 .Where(i => i.Type == CodecType.Video &&
-                    i.ContainsAnyCodec(videoStream?.Codec, container) &&
+                    i.ContainsAnyCodec(videoCodec, container) &&
```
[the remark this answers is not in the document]
nyanmisaka: Maybe, but with your [one-line fix](https://github.com/jellyfin/jellyfin/pull/9723/files#diff-bde118dfa6110f699ead01698c41be0bb9b4b7690a4a0adaf62ad0697ef61d69R749), I can no longer reproduce the issue.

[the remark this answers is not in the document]
dmitrylyzo: Oh. So the problem was only in `GetVideoTranscodeProfile`: using `transcodingProfile.VideoCodec` there (testing with its own codecs instead of the source codec) seems wrong.

And it looks like `BuildStreamVideoItem` should definitely use profile codecs to generate the correct qualifiers for attributes.

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -2005,7 +2005,26 @@ public string GetVideoQualityParam(EncodingJobInfo state, string videoEncoder, E
...
             var profile = state.GetRequestedProfiles(targetVideoCodec).FirstOrDefault() ?? string.Empty;
-            profile = WhiteSpaceRegex().Replace(profile, string.Empty);
+            profile = WhiteSpaceRegex().Replace(profile, string.Empty).ToLowerInvariant();
+
+            if (string.Equals("h264", targetVideoCodec, StringComparison.OrdinalIgnoreCase)
+                || string.Equals("avc", targetVideoCodec, StringComparison.OrdinalIgnoreCase))
+            {
+                profile = _videoProfilesH264.Contains(profile, StringComparison.OrdinalIgnoreCase) ? profile : string.Empty;
+            }
+            else if (string.Equals("h265", targetVideoCodec, StringComparison.OrdinalIgnoreCase)
```
dmitrylyzo, line 2015: I believe it will never equal `h265` because we assign `hevc` ourselves: https://github.com/jellyfin/jellyfin/blob/a1c7dd68f9bb9c79d969795e866acd664c029ab5/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs#L2004

Possible rewrite (pseudo-code):
```cs
var videoProfiles;

if (h264) videoProfiles = _videoProfilesH264;
else if (hevc) videoProfiles = _videoProfilesH265;
else if (av1) videoProfiles = _videoProfilesAv1;
else videoProfiles = new string[]; // or static dummy array

if (!videoProfiles.Contains(profile, StringComparison.OrdinalIgnoreCase)
{
    profile = string.Empty;
}
```

MediaBrowser.Model/Dlna/StreamBuilder.cs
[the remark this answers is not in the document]
nyanmisaka: That's it. Without this fix, qualifiers will only be applied to the transcoder in h264-to-h264 or hevc-to-hevc and similar cases.

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
nyanmisaka: Done. IMHO `avc` is not needed too.

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -2005,7 +2005,26 @@ public string GetVideoQualityParam(EncodingJobInfo state, string videoEncoder, E
...
             var profile = state.GetRequestedProfiles(targetVideoCodec).FirstOrDefault() ?? string.Empty;
-            profile = WhiteSpaceRegex().Replace(profile, string.Empty);
+            profile = WhiteSpaceRegex().Replace(profile, string.Empty).ToLowerInvariant();
+
+            var videoProfiles = Array.Empty<string>();
+            if (string.Equals("h264", targetVideoCodec, StringComparison.OrdinalIgnoreCase))
+            {
+                videoProfiles = _videoProfilesH264;
+            }
+            else if (string.Equals("h265", targetVideoCodec, StringComparison.OrdinalIgnoreCase))
```
dmitrylyzo, line 2015:
```suggestion
            else if (string.Equals("hevc", targetVideoCodec, StringComparison.OrdinalIgnoreCase))
```

nyanmisaka: Oops

dmitrylyzo (approved):
I would squash the last 3 commits (video profiles fix).

Bond-009 (approved):

nielsvanvelzen (approved):

crobibero (approved):

MediaBrowser.Model/Dlna/StreamBuilder.cs
dmitrylyzo, line 900: Reviewing #9374, now I think we probably need to use `playlistItem.VideoCodecs` in place of `videoCodec` because the latter remains unchanged - without HLS restrictions (https://github.com/nyanmisaka/jellyfin/blob/7bd971ffefe983a1f32836b0f2fcbe602f92efa0/MediaBrowser.Model/Dlna/StreamBuilder.cs#L806-L809).

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -930,7 +930,7 @@ private void BuildStreamVideoItem(
 
             var appliedAudioConditions = options.Profile.CodecProfiles
                 .Where(i => i.Type == CodecType.VideoAudio &&
-                    i.ContainsAnyCodec(audioStream?.Codec, container) &&
+                    i.ContainsAnyCodec(audioCodec, container) &&
```
dmitrylyzo, line 933: Reviewing #9374, now I think we probably need to use `playlistItem.AudioCodecs` in place of `audioCodec` because the latter remains unchanged - without HLS restrictions (https://github.com/nyanmisaka/jellyfin/blob/7bd971ffefe983a1f32836b0f2fcbe602f92efa0/MediaBrowser.Model/Dlna/StreamBuilder.cs#L843-L853).

MediaBrowser.Model/Dlna/StreamBuilder.cs
nyanmisaka: That makes sense. I'll start a new one.

dmitrylyzo: Also, here https://github.com/jellyfin/jellyfin/blob/2fe13f54eaf87eefefd27f4ccb2ace1371f5e886/MediaBrowser.Model/Dlna/StreamBuilder.cs#L907

Or maybe just assign it back?
```cs
videoCodec = playlistItem.VideoCodecs;
```

MediaBrowser.Model/Dlna/StreamBuilder.cs
dmitrylyzo: Also, here https://github.com/jellyfin/jellyfin/blob/2fe13f54eaf87eefefd27f4ccb2ace1371f5e886/MediaBrowser.Model/Dlna/StreamBuilder.cs#L940

Or maybe just assign it back?
```cs
audioCodec = playlistItem.AudioCodecs;
```

PR #12524, opened 2024-08-26
Adjust filters to adapt FFmpeg 7.0

nyanmisaka:
**Changes**
- Adjust filters to adapt FFmpeg 7.0

Adjusted the order and parameters of some filters to adapt to FFmpeg 7.0. While maintaining backward compatibility.

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -3010,7 +3010,7 @@ namespace MediaBrowser.Controller.MediaEncoding
             {
                 return string.Format(
                     CultureInfo.InvariantCulture,
-                    @"scale=-1:{1}:fast_bilinear,scale,crop,pad=max({0}\,iw):max({1}\,ih):(ow-iw)/2:(oh-ih)/2:black@0,crop={0}:{1}",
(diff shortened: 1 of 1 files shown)


Bond-009 (approved):

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -3771,7 +3771,7 @@ public string GetLibplaceboFilter(
                 // sw => hw
                 if (doOclTonemap)
                 {
-                    mainFilters.Add("hwupload=derive_device=d3d11va:extra_hw_frames=16");
+                    mainFilters.Add("hwupload=derive_device=d3d11va:extra_hw_frames=24");
```
sjorge, line 3774: You mentioned changing the order but this seems to change the frame pool size to be allocated, 24 still seems to be very conservative with some of the QSV examples I found. But not sure this was intentional or this happened to slip into this pr.

sjorge: nvm, re-reading the commit messages it also mentioned adapting fro ffmpeg 7 separately from changing the order, so this is probably intended.

gnattu (approved):

Shadowghost (approved):

PR #12534, opened 2024-08-29
Add new tonemap mode options

gnattu:
**Changes**

This adds the new tonemap modes introduced in jellyfin-ffmpeg to be selectable, and currently available in CUDA, OpenCL and Metal implementation. One implementation detail to be discussed is how we deal with the vulkan pipeline for AMGPUs on Linux now. Currently if the user select the new modes LUM or ITP, it will do nothing for vulkan pipeline using libplacebo as those modes does not exists in libplacbo. Should we also implement the OpenCL pipeline for AMDGPU under linux and use that when these modes are selected?

**Issues**

See the following PRs:
https://github.com/jellyfin/jellyfin-ffmpeg/pull/430
https://github.com/jellyfin/jellyfin-ffmpeg/pull/438

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -66,6 +66,7 @@ namespace MediaBrowser.Controller.MediaEncoding
         private readonly Version _minFFmpegReadrateOption = new Version(5, 0);
         private readonly Version _minFFmpegWorkingVtHwSurface = new Version(7, 0, 1);
         private readonly Version _minFFmpegDisplayRotationOption = new Version(6, 0);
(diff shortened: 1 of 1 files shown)


nyanmisaka (approved):
> One implementation detail to be discussed is how we deal with the vulkan pipeline for AMGPUs on Linux now.

I'm thinking of making a `vf_tonemap_vulkan` to replace `vf_libplacebo`. The latter's behavior is inconsistent with the native vulkan filter in ffmpeg, and the API is always changing out of our control.

> Should we also implement the OpenCL pipeline for AMDGPU under linux and use that when these modes are selected?

We just recently moved away from ROCm/ROCr OpenCL, which is always breaking and does not support any form of zero-copy with video codec API.

Shadowghost (approved):

Bond-009 (approved):

PR #12540, opened 2024-08-29
Tune x264/5 encoding params for realtime playback

nyanmisaka:
Some options that are too performance-intensive are disabled, while others are left enabled to trade-off between real-time playback and quality. Now the encoding fps of `libx265` under the veryfast preset is about half that of `libx264` when tonemapping 4k video in software.

**Changes**
- Tune x264/5 encoding params for realtime playback

### DIFF
--- a/Jellyfin.Api/Controllers/DynamicHlsController.cs
+++ b/Jellyfin.Api/Controllers/DynamicHlsController.cs
@@ -45,6 +45,7 @@ public class DynamicHlsController : BaseJellyfinApiController
     private const TranscodingJobType TranscodingJobType = MediaBrowser.Controller.MediaEncoding.TranscodingJobType.Hls;
 
     private readonly Version _minFFmpegFlacInMp4 = new Version(6, 0);
+    private readonly Version _minFFmpegX265BframeInFmp4 = new Version(7, 0, 1);
 
     private readonly ILibraryManager _libraryManager;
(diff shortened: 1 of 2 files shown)


Shadowghost (approved):

crobibero (approved):

gnattu (approved):
It is still way too CPU intensive and I recommend against using it for 4K videos without downscaling to 1080p as most CPUs will struggle with libx265 even with those params.

nyanmisaka:
> It is still way too CPU intensive and I recommend against using it for 4K videos without downscaling to 1080p as most CPUs will struggle with libx265 even with those params.

That's right. Unfortunately, if we want to add HDR2HDR or 10-bit transcoding, libx265 is a must. Btw how is x265 performing on Apple Silicon recently? Are there any ongoing optimizations?

gnattu:
> > It is still way too CPU intensive and I recommend against using it for 4K videos without downscaling to 1080p as most CPUs will struggle with libx265 even with those params.
> 
> That's right. Unfortunately, if we want to add HDR2HDR or 10-bit transcoding, libx265 is a must. Btw how is x265 performing on Apple Silicon recently? Are there any ongoing optimizations?

arm optimization is always ongoing as the arm based server are getting adoption but the performance is still only okayish (barely does 4k 30fps end to end on M1 Max, for some inputs even lower).

PR #12546, opened 2024-08-30
Add audio ranking for transcoding profiles

dmitrylyzo:
It looks like `DirectStream` is currently disabled with [`!ForceDirectStream`](https://github.com/jellyfin/jellyfin/blob/9644e894f0f6e1206f68a2b23aeb8065a565f9bf/Jellyfin.Api/Helpers/MediaInfoHelper.cs#L237-L241) (https://github.com/jellyfin/jellyfin/pull/7325).
So [GetVideoDirectPlayProfile](https://github.com/jellyfin/jellyfin/blob/9644e894f0f6e1206f68a2b23aeb8065a565f9bf/MediaBrowser.Model/Dlna/StreamBuilder.cs#L1178) only looks for `DirectPlay`, leaving `DirectStream` for [GetVideoTranscodeProfile](https://github.com/jellyfin/jellyfin/blob/9644e894f0f6e1206f68a2b23aeb8065a565f9bf/MediaBrowser.Model/Dlna/StreamBuilder.cs#L486). But the latter does not currently rank audio, which may cause problems if there are multiple TranscodingProfiles - choosing a suboptimal profile.

I originally made these changes to force `DirectStream` to use the "transcoding" flow/pipeline by removing it from `GetVideoDirectPlayProfile`. That's why there is unused [`playMethod`](https://github.com/jellyfin/jellyfin/blob/c52fddeede84cea9c1b6fce6d21e185665697876/MediaBrowser.Model/Dlna/StreamBuilder.cs#L754).

**Changes**
Add audio ranking for transcoding profiles.

**Issues**
N/A


gnattu:
I had similar observation when I implement audio remuxing. It seems like the meaning of DirectStream changed during the development so that that playmethod is not really used in the pipeline. The original meaning of this play method seems to be closer to direct play than direct streaming that we are referring now (copy audio and transcode audio).

dmitrylyzo:
Users seem to be confused as well.

At some point I started thinking that we should only have 2 pipelines: `DirectPlay` and `NonDirectPlay` (Remux / TranscodeAudio / TranscodeVideo / TranscodeBoth); and should only remux to what is specified in TranscodingProfiles.

Shadowghost (approved):

gnattu (approved):

crobibero (approved):

PR #12561, opened 2024-09-01
Use enums for encoding options

Shadowghost:
Add enums for:
* DeinterlaceMethod
* EncoderPreset
* HardwareAccelerationType
* TonemappingAlgorithm
* TonemappingMode
* TonemappingRange

A migration for the existing config was also added for completeness' sake. Since the string values are the same, it should not be necessary.

I also made sure they are backwards compatible with our current values by enforcing lowercase.


JPVenson:
are the enum keys only used for display representation or are they used as actual values? i dont seem to find a mapping layer to convert "superfast" to its actual representation. If there is no translation layer, it has to be made *cristal clear* in the description, with 3 exclamation marks that changing an enum key has significate impact on other parts of the code.

Shadowghost:
Some of them are used as actual values within the FFmpeg command builder. Depending on used HWA there are mappings implemented. If the enum needs to be converted to a string, I used `toString().toLowerInvariant()` to make sure we don't have issues if we haver use proper casing on the Enum values.

nielsvanvelzen (changes requested):
Before this PR the value of hardwareAccelerationType was of type HardwareEncodingType, which used uppercase values (e.g. "AMF", "VAAPI"). In this PR the type was renamed to HardwareAccelerationType and uses lowercase values (e.g. "amf", "vaapi"). This is an incompatible API change as enums are case sensitive.

Shadowghost:
@nielsvanvelzen doesn't jf-web use lowercase though (see [here](https://github.com/jellyfin/jellyfin-web/blob/3ecfda52921d9f73416442e068218caeafc412ce/src/controllers/dashboard/encodingsettings.html#L13))

MediaBrowser.Model/Configuration/EncodingOptions.cs
```diff
@@ -275,10 +276,10 @@ public EncodingOptions()
     /// <summary>
     /// Gets or sets the codecs hardware encoding is used for.
     /// </summary>
-    public string[] HardwareDecodingCodecs { get; set; }
+    public IReadOnlyList<string> HardwareDecodingCodecs { get; set; }
```
crobibero, line 279: Last I checked `IReadOnlyList` doesn't play well with xml serialization

Shadowghost: According to this you seem to be correct: https://stackoverflow.com/questions/31552724/how-why-does-xmlserializer-treat-a-class-differently-when-it-implements-ilistt

Will change it back later.

Shadowghost: AFAIK we tried this multiple times and it always lead to issues.

Jellyfin.Server/Migrations/PreStartupRoutines/MigrateEncodingOptions.cs
```diff
@@ -0,0 +1,246 @@
...
+        catch (InvalidOperationException ex)
+        {
+            _logger.LogError(ex, "Migrate EncodingOptions deserialize Invalid Operation error");
+        }
+        catch (Exception ex)
+        {
+            _logger.LogError(ex, "Migrate EncodingOptions deserialize error");
+        }
+
+        if (oldConfig is not null)
```
Bond-009, line 59: Just do an early return if it is null

Jellyfin.Server/Migrations/PreStartupRoutines/MigrateEncodingOptions.cs
```diff
@@ -0,0 +1,246 @@
...
+
+        public double VppTonemappingBrightness { get; set; }
+
+        public double VppTonemappingContrast { get; set; }
+
+        public int H264Crf { get; set; }
+
+        public int H265Crf { get; set; }
+
+        public string EncoderPreset { get; set; }
```
Bond-009, line 216:
```suggestion
        public string EncoderPreset { get; set; }

```

Jellyfin.Server/Migrations/PreStartupRoutines/MigrateEncodingOptions.cs
```diff
@@ -0,0 +1,246 @@
...
+        public bool EnableVideoToolboxTonemapping { get; set; }
+
+
+        public string TonemappingAlgorithm { get; set; }
+
+
+        public string TonemappingMode { get; set; }
+
+        public string TonemappingRange { get; set; }
+
```
Bond-009, line 200:
```suggestion
        public string TonemappingRange { get; set; }
```

Jellyfin.Server/Migrations/PreStartupRoutines/MigrateEncodingOptions.cs
```diff
@@ -0,0 +1,246 @@
...
+        public bool EnableTonemapping { get; set; }
+
+        public bool EnableVppTonemapping { get; set; }
+
+        public bool EnableVideoToolboxTonemapping { get; set; }
+
+
+        public string TonemappingAlgorithm { get; set; }
+
```
Bond-009, line 195:
```suggestion
        public string TonemappingAlgorithm { get; set; }
```

Jellyfin.Server/Migrations/PreStartupRoutines/MigrateEncodingOptions.cs
```diff
@@ -0,0 +1,246 @@
...
+        public string EncoderAppPathDisplay { get; set; }
+
+        public string VaapiDevice { get; set; }
+
+        public bool EnableTonemapping { get; set; }
+
+        public bool EnableVppTonemapping { get; set; }
+
+        public bool EnableVideoToolboxTonemapping { get; set; }
+
```
Bond-009, line 192:
```suggestion
        public bool EnableVideoToolboxTonemapping { get; set; }
```

Jellyfin.Server/Migrations/PreStartupRoutines/MigrateEncodingOptions.cs
```diff
@@ -0,0 +1,246 @@
...
+        public bool AllowAv1Encoding { get; set; }
+
+        public bool EnableSubtitleExtraction { get; set; }
+
+        public string[] HardwareDecodingCodecs { get; set; }
+
+        public string[] AllowOnDemandMetadataBasedKeyframeExtractionForExtensions { get; set; }
+    }
+#pragma warning restore
```
Bond-009, line 245: No need to restore at the end of a file

Bond-009 (approved):
Just style things

MediaBrowser.Model/Entities/EncoderPreset.cs
```diff
@@ -0,0 +1,59 @@
+#pragma warning disable SA1300 // Lowercase required for backwards compat.
+
+namespace MediaBrowser.Model.Entities;
+
+/// <summary>
+/// Enum containing encoder presets.
+/// </summary>
+public enum EncoderPreset
+{
```
gnattu, line 9: We do have an "auto" preset which is basically an empty string so that it will be assigned to the per-encoder default. Now this breaks all auto cases

gnattu (changes requested):

MediaBrowser.Model/Entities/EncoderPreset.cs
Shadowghost: Where is this used?

gnattu: When we map the preset to encoder specific qualities, we do have a case that is not any of those value which is the real default "auto" value

gnattu (approved):
LGTM on the server side now. Web needs some slight changes for this PR though

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -176,18 +177,18 @@ private string GetH26xOrAv1Encoder(string defaultEncoder, string hwEncoder, Enco
...
-                    { "vaapi",                hwEncoder + "_vaapi" },
-                    { "videotoolbox",         hwEncoder + "_videotoolbox" },
-                    { "v4l2m2m",              hwEncoder + "_v4l2m2m" },
-                    { "rkmpp",                hwEncoder + "_rkmpp" },
+                    { HardwareAccelerationType.amf,                  hwEncoder + "_amf" },
+                    { HardwareAccelerationType.nvenc,                hwEncoder + "_nvenc" },
+                    { HardwareAccelerationType.qsv,                  hwEncoder + "_qsv" },
+                    { HardwareAccelerationType.vaapi,                hwEncoder + "_vaapi" },
+                    { HardwareAccelerationType.videotoolbox,         hwEncoder + "_videotoolbox" },
+                    { HardwareAccelerationType.v4l2m2,              hwEncoder + "_v4l2m2m" },
```
crobibero, line 187: This was previously `v4l2m2m`

nyanmisaka (approved):

crobibero (approved):

PR #12574, opened 2024-09-03
Add catch to remove cached user if creation fails

JPVenson:
Added catch block to clear newly created item from cache if database update fails


Shadowghost (approved):

Bond-009:
Another possibility is changing `CreateUserInternalAsync` so it doesn't add the user to the dict but just returns it and move the code that does that after the DB code. That way it isn't reached if the DB throws and you don't have these try-catches

Bond-009 (approved):

crobibero (approved):

PR #12585, opened 2024-09-05
Migrate rating level to double

### LINKED ISSUE #11650: [Issue]: Australian Classification Ratings not correct

Shadowghost:
We need to have more granular control if we have rating strings which map to the same age.

**Changes**
* Switch all rating level related logic to double
* Re-enable migration to re-create and re-populate changed column

**Issues**
Prerequisite to fix #11650

### DIFF
--- a/Emby.Server.Implementations/Data/SqliteItemRepository.cs
+++ b/Emby.Server.Implementations/Data/SqliteItemRepository.cs
@@ -467,7 +467,7 @@ namespace Emby.Server.Implementations.Data
                 AddColumn(connection, "TypedBaseItems", "ExternalServiceId", "Text", existingColumnNames);
                 AddColumn(connection, "TypedBaseItems", "Tags", "Text", existingColumnNames);
                 AddColumn(connection, "TypedBaseItems", "IsFolder", "BIT", existingColumnNames);
-                AddColumn(connection, "TypedBaseItems", "InheritedParentalRatingValue", "INT", existingColumnNames);
+                AddColumn(connection, "TypedBaseItems", "InheritedParentalRatingValue", "Float", existingColumnNames);
                 AddColumn(connection, "TypedBaseItems", "UnratedType", "Text", existingColumnNames);
                 AddColumn(connection, "TypedBaseItems", "TopParentId", "Text", existingColumnNames);
                 AddColumn(connection, "TypedBaseItems", "TrailerTypes", "Text", existingColumnNames);
@@ -1746,7 +1746,7 @@ namespace Emby.Server.Implementations.Data
 
             if (HasField(query, ItemFields.InheritedParentalRatingValue))
             {
-                if (reader.TryGetInt32(index++, out var parentalRating))
+                if (reader.TryGetDouble(index++, out var parentalRating))
                 {
                     item.InheritedParentalRatingValue = parentalRating;
                 }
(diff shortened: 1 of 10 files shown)


Shadowghost:
This won't work as we thought it would work out. To fix this properly, a more invasive approach is required for 10.10 or 10.11.

Rygle:
This whole thing seems have gone wrong in an effort to make things more granular.

Perhaps a better approach would be to make things more granular in the options available in the settings rather than in the interpretation of the settings that the user has no control over (apart from re-compiling Jellyfin). So all profiles have a check list of all possible ratings, and the parent/administrator has to explicitly tick every rating they want to be visible. Or to speed this up, it could have umbrella categories with sub-categories that get selected if the umbrella category is selected, but only if the umbrella category is selected. Alternately, there could be a tick box for "more granular ratings selections" that does something like the following (if un-ticked it can remain as it was)...

So, in the Jellyfin settings, I will see something like;

NR [ ]
-- NR [ ]
-- AU-NC [ ]
etc for other countries

U [ ]
-- U [ ]
-- AU-U [ ]
etc for other countries

G [ ]
-- MPAA-G [ ]
-- AU-G [ ]
-- 7 [ ]
etc for other countries

PG [ ]
-- 13
-- PG-13 [ ]
-- AU-PG [ ]
-- 14 [ ]
etc for other countries

M [ ]
-- MPAA-M [ ]
-- 15 [ ]
-- AU-M [ ]
-- AU-MA15 [ ]
etc for other countries

etc, etc.

That way a parent/administrator can tick either the larger category that will select all under it, or just tick the ratings they want.

Additionally, if there are ratings/classifications that are not available in Jellyfin that match the metadata for your region/location, Jellyfin should allow the parent/administrator to type in their own custom ratings that match the metadata that their scraper is getting for their location, and that these get incorporated into the longer list of tick-boxes for ratings once entered.

Shadowghost:
This would be way more effort and would still be kind of a hack. Another problem with this approach is just the pure amount of different ratings we have.
Additionally, some ratings can mean different things in different ratings systems which would make it super intransparent.
I appreciate the input, but the whole idea of mapping the ratings is to NOT force the user to granularity configure any kind of limitations.

PR #12587, opened 2024-09-05
Add more DTOs to API

### LINKED ISSUE #11244: API should use a new "SessionInfoDto" type to avoid duplicate types for ClientCapabilities and BaseItem
Currently the OpenAPI specification contains a few duplicate types because we're not using a data transfer object for SessionInfo.

The SessionInfo is "leaked" from the API in a few places:
- SessionController.getSessions
- SessionsMessage (in a property)
- AuthenticationResult (in a property)

These places need to be changed to use a (new) SessionInfoDto type, that uses the proper Dto types for its members. An alternative fix would be to use the dto types in the existing SessionInfo.

Shadowghost:
**Changes**
*  Add `SessionInfoDto`

**Issues**
Fixes #11244


MediaBrowser.Model/Dto/SessionInfoDto.cs
```diff
@@ -0,0 +1,186 @@
...
+    /// Gets or sets the additional users.
+    /// </summary>
+    /// <value>The additional users.</value>
+    public IReadOnlyList<SessionUserInfo>? AdditionalUsers { get; set; }
+
+    /// <summary>
+    /// Gets or sets the client capabilities.
+    /// </summary>
+    /// <value>The client capabilities.</value>
+    public ClientCapabilities? Capabilities { get; set; }
```
nielsvanvelzen, line 29: Should use the DTO variant

nielsvanvelzen (changes requested):

nielsvanvelzen (approved):
✅ It does look like `ClientCapabilities` (the non-dto type) is also used in `DeviceInfo` so we may need to add a `DeviceInfoDto` as-well. Probably better to do that in a separate pull request.

src/Jellyfin.Extensions/Json/Converters/JsonDelimitedArrayConverter.cs
```diff
@@ -75,7 +59,22 @@ protected JsonDelimitedArrayConverter()
         /// <inheritdoc />
         public override void Write(Utf8JsonWriter writer, T[]? value, JsonSerializerOptions options)
         {
-            throw new NotImplementedException();
+            writer.WriteStartArray();
+            if (value is not null && value.Length > 0)
```
crobibero, line 63: If `value` is null we shouldn't write an empty array

Jellyfin.Api/Controllers/SessionController.cs
```diff
@@ -56,66 +47,25 @@ public SessionController(
...
+            User.GetUserId(),
+            deviceId,
+            activeWithinSeconds,
+            controllableUserToCheck);
+
+        if (result.Count == 0)
         {
-            var minActiveDate = DateTime.UtcNow.AddSeconds(0 - activeWithinSeconds.Value);
-            result = result.Where(i => i.LastActivityDate >= minActiveDate);
+            return NotFound();
```
crobibero, line 68: Since this endpoint is searching for sessions an `Ok` is correct here and not `NotFound`

Shadowghost: Agreed, but this just replicates the current behaviour. Should I change it?

crobibero: Yes, I would change it

crobibero (approved):

src/Jellyfin.Extensions/Json/Converters/JsonDelimitedArrayConverter.cs
```diff
@@ -75,7 +65,29 @@ protected JsonDelimitedArrayConverter()
...
-            throw new NotImplementedException();
+            if (value is not null)
+            {
+                writer.WriteStartArray();
+                if (value.Length > 0)
+                {
+                    var toWrite = value.Length - 1;
+                    foreach (var it in value)
+                    {
+                        writer.WriteStringValue(it!.ToString());
```
Bond-009, line 76: Shouldn't you properly handle null values here?

Shadowghost: How do you think they should be handled? `"null"` or `null` or dropped?

crobibero: I think `null` would typically be dropped

Jellyfin.Api/Controllers/SessionController.cs
crobibero: @Shadowghost I think this was missed?

src/Jellyfin.Extensions/Json/Converters/JsonDelimitedArrayConverter.cs
```diff
@@ -75,7 +66,39 @@ protected JsonDelimitedArrayConverter()
...
+                writer.WriteStartArray();
+                if (value.Length > 0)
+                {
+                    var toWrite = value.Length - 1;
+                    foreach (var it in value)
+                    {
+                        var wrote = false;
+                        if (it is not null)
+                        {
+                            writer.WriteStringValue(it!.ToString());
```
crobibero, line 80:
```suggestion
                            writer.WriteStringValue(it.ToString());
```

Shadowghost:
@crobibero the whole logic for selecting sessions including this codepiece was moved into `public IReadOnlyList<SessionInfoDto> GetSessions`

crobibero:
> @crobibero the whole logic for selecting sessions including this codepiece was moved into `public IReadOnlyList<SessionInfoDto> GetSessions`

I'm referring to the endpoint returning NotFound if no results are in the query instead of just returning an empty list

Jellyfin.Api/Controllers/SessionController.cs
Shadowghost: It was intentionally reverted because of comments on Matrix.

crobibero (approved):

src/Jellyfin.Extensions/Json/Converters/JsonDelimitedArrayConverter.cs
```diff
@@ -35,38 +37,27 @@ protected JsonDelimitedArrayConverter()
...
-                {
-                    if (parsedValues[i] is not null)
-                    {
-                        typedValues.SetValue(parsedValues[i], typedValueIndex);
-                        typedValueIndex++;
-                    }
-                }
-
-                return typedValues;
+                return [.. typedValues];
```
Bond-009, line 60:
```suggestion
                return typedValues.ToArray();
```

src/Jellyfin.Extensions/Json/Converters/JsonDelimitedArrayConverter.cs
```diff
@@ -75,7 +66,39 @@ protected JsonDelimitedArrayConverter()
...
+                        {
+                            writer.WriteStringValue(it.ToString());
+                            wrote = true;
+                        }
+
+                        if (toWrite > 0)
+                        {
+                            if (wrote)
+                            {
+                                writer.WriteStringValue(Delimiter.ToString());
```
Bond-009, line 88: Not sure about the perf impact of doing `ToString` everytime :thinking: 
For now this should be fine but I'll do some benchmarks when I've got time.

Bond-009 (approved):
Just 1 small comment

PR #12562, opened 2024-09-01
Use filtered codecs to build appliedConditions

nyanmisaka:
Based on the new suggestions here. https://github.com/jellyfin/jellyfin/pull/12521#discussion_r1740054405

**Changes**
- Use filtered codecs to build appliedConditions

PR #12602, opened 2024-09-07
Backport #12562 and #12521

gnattu:
**Changes**

This is a manual back-port of #12562 and #12521

**Issues**

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -1915,7 +1915,26 @@ namespace MediaBrowser.Controller.MediaEncoding
             }
 
             var profile = state.GetRequestedProfiles(targetVideoCodec).FirstOrDefault() ?? string.Empty;
-            profile = WhiteSpaceRegex().Replace(profile, string.Empty);
+            profile = WhiteSpaceRegex().Replace(profile, string.Empty).ToLowerInvariant();
+
+            var videoProfiles = Array.Empty<string>();
+            if (string.Equals("h264", targetVideoCodec, StringComparison.OrdinalIgnoreCase))
+            {
+                videoProfiles = _videoProfilesH264;
+            }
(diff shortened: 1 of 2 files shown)


dmitrylyzo:
`videoStream?.Codec` -> `videoCodecs`.
https://github.com/jellyfin/jellyfin/blob/d25addaf7e3b9d03e4b55224fd071f53bad46f7c/MediaBrowser.Model/Dlna/StreamBuilder.cs#L969

I would do a backport, keeping Nyan as the author, but I'm not sure which author would be taken when squashing.

gnattu:
> `videoStream?.Codec` -> `videoCodecs`.
> 
> https://github.com/jellyfin/jellyfin/blob/d25addaf7e3b9d03e4b55224fd071f53bad46f7c/MediaBrowser.Model/Dlna/StreamBuilder.cs#L969
> 
> I would do a backport, keeping Nyan as the author, but I'm not sure which author would be taken when squashing.

This change is intentional as that is what we did to 10.9, or something has changed since then.

> I would do a backport, keeping Nyan as the author, but I'm not sure which author would be taken when squashing.

Can this be done by PR though? I think this needs to commit to master directly?

dmitrylyzo:
> This change is intentional as that is what we did to 10.9, or something has changed since then.

https://github.com/jellyfin/jellyfin/pull/12562/files#diff-bde118dfa6110f699ead01698c41be0bb9b4b7690a4a0adaf62ad0697ef61d69R900
It just requires the change from the previous PR: https://github.com/jellyfin/jellyfin/pull/12521/files#diff-bde118dfa6110f699ead01698c41be0bb9b4b7690a4a0adaf62ad0697ef61d69R900

> Can this be done by PR though? I think this needs to commit to master directly?

I would cherry-pick Nyan's commit and fix conflicts. Or you can just copy stats from the commit using `-C` flag for Git CLI.

Shadowghost (approved):

nyanmisaka (approved):

dmitrylyzo (approved):

crobibero (approved):

PR #11559, opened 2024-05-12
Add support DoVi Profile 10

llutic:
Added compatibility for AV1 DoVi Profile 10 based on dolby documentation

**Changes**

Dolby Vision profile 10 detection added for movies and series encoded with AV1 codec with Dolby Vision HDR layer.

Sources : https://professionalsupport.dolby.com/s/article/What-is-Dolby-Vision-Profile

**Issues**

I'm starting to get movies and series encoded in AV1 with DV but I can't see whether or not they are Dolby Vision compatible in the media information.

PR #12604, opened 2024-09-07
Enable tone-mapping and HLS remuxing for DoVi Profile 10 in AV1

nyanmisaka:
Together with #11559 and jellyfin-ffmpeg7, we can enable DoVi Profile 10.

**Changes**
- Enable hardware tone-mapping for DoVi Profile 10
- Add HLS remuxing support for DoVi Profile 10


gnattu (approved):
Theoretically it will work but can't check dav1 remuxing compatibility at the moment.

Shadowghost (approved):

crobibero (approved):

PR #12605, opened 2024-09-07
Add SUPPLEMENTAL-CODECS for Dolby Vision video with fallbacks

gnattu:
This is a more correct way to supply Dolby Vision videos with PQ/HLG fallbacks. The `CODECS` is used to describe the base layer and this `SUPPLEMENTAL-CODECS` attribute will be used to describe the enhanced codec. Supported clients will use this field to check if the enhancement layer is available, and this attribute will simply be ignored when the client does not support that field.

One thing to note is that for Dolby Vision videos without a fallback layer, the HLS attributes generated by our current implementation are not entirely correct. If the Dolby Vision video lacks a fallback layer, the `CODECS` attribute should directly describe the Dolby Vision profile, such as `dvh1.05.09`, rather than the HEVC codec profile. However, I'm a bit concerned about compatibility if we decide to make this change. I’ve tested both HLS.js and Safari’s native HLS, and they can handle the Dolby Vision fourCC correctly, but I’m unsure about the support status on other clients. We can change that later if that becomes a problem to support `dav1` or Dolby Vision Profile 10 videos.

**Changes**

**Issues**

### DIFF
--- a/Jellyfin.Api/Helpers/DynamicHlsHelper.cs
+++ b/Jellyfin.Api/Helpers/DynamicHlsHelper.cs
@@ -304,6 +304,8 @@ public class DynamicHlsHelper
 
         AppendPlaylistCodecsField(playlistBuilder, state);
 
+        AppendPlaylistSupplementalCodecsField(playlistBuilder, state);
+
         AppendPlaylistResolutionField(playlistBuilder, state);
 
         AppendPlaylistFramerateField(playlistBuilder, state);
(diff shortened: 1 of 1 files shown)


nyanmisaka:
> One thing to note is that for Dolby Vision videos without a fallback layer, the HLS attributes generated by our current implementation are not entirely correct. If the Dolby Vision video lacks a fallback layer, the `CODECS` attribute should directly describe the Dolby Vision profile, such as `dvh1.05.09`, rather than the HEVC codec profile. However, I'm a bit concerned about compatibility if we decide to make this change. I’ve tested both HLS.js and Safari’s native HLS, and they can handle the Dolby Vision fourCC correctly, but I’m unsure about the support status on other clients. We can change that later if that becomes a problem to support `dav1` or Dolby Vision Profile 10 videos.

This has been left unchanged for compatibility reasons. Not all clients can falling back from `dvh1` to `hvc1` entry.
But clients that are known to support `dvh1` are able to tolerate mismatched fourccs between playlist and fmp4 stream.

nyanmisaka (approved):

crobibero (approved):

holow29:
Is there any thought to adding HDR10+ supplemental-codecs information?
From Apple:

> Signaling HDR10+ content
Within a Multivariant Playlist the codec is signaled via a format in the CODECS attribute. There is no way to signal HDR10+ in a format. To indicate that your content includes HDR10+ video, supply a SUPPLEMENTAL-CODECS attribute containing the appropriate format and brand (separated by a slash character). The format should match the HDR10 format. The correct brand for HDR10+ is 'cdm4'. For example, SUPPLEMENTAL-CODECS="hvc1.2.20000000.L123.B0/cdm4".

gnattu:
There is no clean way to detect HDR10+ metadata presence in ffprobe now because unlike Dolby Vision which has a configuration record in container, you have to actually decode a frame to see if there is HDR10+ side data to tell if the Video is HDR10+, and we has to modify ffprobe to have a fast enough way to probe that I'm afraid.

PR #12610, opened 2024-09-08
Enable Rockchip MJPEG encoder for Trickplay

nyanmisaka:
**Changes**
- Enable Rockchip MJPEG encoder for Trickplay
- Use `-global_quality` for QSV and VA-API MJPEG encoder (as suggested by Intel)

**Issues**
- https://features.jellyfin.org/posts/2783/trickplay-hardware-accelerated-mjpeg-encoding-on-rockchip-devices


MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -5578,12 +5582,6 @@ public string GetVideoTransposeDirection(EncodingJobInfo state)
             if (doOclTonemap)
             {
                 var tonemapFilter = GetHwTonemapFilter(options, "opencl", "nv12");
-                // enable tradeoffs for performance
-                if (!string.IsNullOrEmpty(tonemapFilter))
-                {
-                    tonemapFilter += ":tradeoff=1";
```
gnattu, line 5584: Is this because the dovi tone mapping not available or the filter can automatically set tradeoff flag?

gnattu (approved):
A question: does RKMPP support key frame only decoding?

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
nyanmisaka: The filter can do it automatically, so it is redundant.

nyanmisaka:
> A question: does RKMPP support key frame only decoding?

I tried implementing but it wasn't very stable, especially with some mpegts or interlaced videos where feeding the decoder only I-frames could cause errors. So I'm not going to add it until I figure it out.

Shadowghost (approved):

gnattu:
> or interlaced videos

This reminds me one thing: should we disable deinterlance filter for trickplay image generation? YADIF and BWDIF filters are temporal filters and probably will not work correctly without all frames being decoded.

nyanmisaka:
> > or interlaced videos
> 
> This reminds me one thing: should we disable deinterlance filter for trickplay image generation? YADIF and BWDIF filters are temporal filters and probably will not work correctly without all frames being decoded.

Yes, deint algorithms need consecutive frames.

I'm not sure if the various deint filters handled the progressive video, but disabling it should give a slight fps boost?
`state.DeInterlace`

gnattu:
> but disabling it should give a slight fps boost?

Not that measurable but it is slightly faster (about 5%), and the results look similar in trickplay size.

nyanmisaka:
> > but disabling it should give a slight fps boost?
> 
> Not that measurable but it is slightly faster (about 5%), and the results look similar in trickplay size.

Not sure if it's worth it then, if it requires a lot of changes to disable it.

crobibero (approved):

PR #12613, opened 2024-09-08
Update issue template version from 10.9.10 to 10.9.11

TimGels:
**Changes**
Issue template from 10.9.10 -> 10.9.11

**Issues**

N/A


gnattu (approved):
Maybe having this integrated into our release CI so that we don't have to do this manually every time?

TimGels:
> Maybe having this integrated into our release CI so that we don't have to do this manually every time?

Yea that is the plan. But I currently don't have direct access to my pc as I am troubleshooting a problem with it. But it is on the radar to get done eventually.

nielsvanvelzen (approved):

PR #12626, opened 2024-09-11
Only sort item by width when they have the same path

### LINKED ISSUE #12129: Multiple versions sorts by resolution regardless of naming structure
### Please describe your bug

According to the docs, multiple versions should be sorted by _filenames_ in alphabetic order except for the case of resolution filenames.

"Order of Versions
Movie versions are presented in an alphabetically sorted list. An exception applies to resolution names, which are sorted in descending order from highest to lowest resolution. A version name qualifies as a resolution name when ending with either a p or an i. "

However, the WebUI always sorts by resolution first regardless of filenames. 

### Reproduction Steps

Beta
Path: V: \Movies\Big Buck Bunny (2008)\Big Buck Bunny (2008) - Beta.mp4
Resolution: 3840x2160

Alpha
Path: V: \Movies\Big Buck Bunny (2008)\Big Buck Bunny (2008) - Alpha.mp4
Resolution: 1920x1080

Gamma
Path: V: \Movies\Big Buck Bunny (2008)\Big Buck Bunny (2008) - Gamma.mp4
Resolution: 1920x1080

The higher resolution gets sorted first, even though the filename doesn't include any resolutions. 

versions

### Jellyfin Version

10.9.0

### if other:

10.9.6

### Environment

```markdown
- OS: Windows
- Linux Kernel: N/A
- Virtualization: No
- Clients: Web
- Browser: Firefox
- FFmpeg Version: N/A
- Playback Method: N/A
- Hardware Acceleration: N/A
- GPU Model: N/A
- Plugins: N/A
- Reverse Proxy: N/A
- Base URL: N/A
- Networking: N/A
- Storage: N/A
- Server: 10.9.6
```

### Jellyfin logs

```shell
N/A
```

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

gnattu:
Since emby era we are re-ordering all media sources by its width when reading from database. However, that would break the sorting made by the media prober when resolving multiple versions from different files. Only do such sorting when media streams share the same path to respect the filename based sorting done by the multi version resolver.

This should have little to no impact for most cases are most of the files will only have strictly one video stream.

**Changes**

**Issues**

Fixes #12129

### DIFF
--- a/MediaBrowser.Controller/Entities/MediaSourceWidthComparator.cs
+++ b/MediaBrowser.Controller/Entities/MediaSourceWidthComparator.cs
@@ -6,7 +6,7 @@ using MediaBrowser.Model.Dto;
 namespace MediaBrowser.Controller.Entities;
 
 /// <summary>
-/// Alphanumeric <see cref="IComparer{T}" />.
+/// Compare MediaSource of the same file by Video width <see cref="IComparer{T}" />.
 /// </summary>
 public class MediaSourceWidthComparator : IComparer<MediaSourceInfo>
 {
--- a/MediaBrowser.Controller/Entities/MediaSourceWidthComparator.cs
+++ b/MediaBrowser.Controller/Entities/MediaSourceWidthComparator.cs
@@ -1,0 +1,56 @@
+using System;
+using System.Collections.Generic;
+using System.Runtime.Intrinsics.X86;
+using MediaBrowser.Model.Dto;
+
+namespace MediaBrowser.Controller.Entities;
+
+/// <summary>
+/// Alphanumeric <see cref="IComparer{T}" />.
+/// </summary>
+public class MediaSourceWidthComparator : IComparer<MediaSourceInfo>
+{
+    /// <inheritdoc />
+    public int Compare(MediaSourceInfo? x, MediaSourceInfo? y)
+    {
+        if (x is null && y is null)
+        {
+            return 0;
+        }
+
+        if (x is null)
+        {
+            return -1;
+        }
+
+        if (y is null)
+        {
+            return 1;
+        }
+
+        if (string.Equals(x.Path, y.Path, StringComparison.OrdinalIgnoreCase))
+        {
+            if (x.VideoStream is null && y.VideoStream is null)
+            {
+                return 0;
+            }
+
+            if (x.VideoStream is null)
+            {
+                return -1;
+            }
+
+            if (y.VideoStream is null)
+            {
+                return 1;
+            }
+
+            var xWidth = x.VideoStream.Width ?? 0;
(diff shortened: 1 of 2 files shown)


MediaBrowser.Controller/Entities/MediaSourceWidthComparator.cs
```diff
@@ -0,0 +1,56 @@
...
+                return 1;
+            }
+
+            var xWidth = x.VideoStream.Width ?? 0;
+            var yWidth = y.VideoStream.Width ?? 0;
+
+            return xWidth - yWidth;
+        }
+
+        return 0;
```
dmitrylyzo, line 54: Shouldn't we compare paths here?

gnattu: No. We strictly want to keep the order when paths mismatch to keep original ordering made by the prober.

Shadowghost (approved):

theguymadmax:
Results: 
Screenshot from 2024-09-11 02-58-26

Works as documented based on this test! The results of X-Men Apocalypse (2016).mp4 is probably a not valid test for sorting multiple versions, as files should follow the pattern: "filename - label".

MediaBrowser.Controller/Entities/MediaSourceWidthComparator.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            return -1;
+        }
+
+        if (y is null)
+        {
+            return 1;
+        }
+
+        if (string.Equals(x.Path, y.Path, StringComparison.OrdinalIgnoreCase))
```
dmitrylyzo, line 31: If `Path` is an absolute path, we will never get the same paths for 2 different items. 🤷‍♂️

gnattu: Some containers like Matroska can have multiple video streams in the same file with the same path, and I’m unsure what would happen if we simply remove that part, as it’s been in place since the Emby era. This sorting wasn’t initially added for multi-version handling, but I’m unsure of its original intent. I’m not confident in removing it entirely, as it could potentially break things.

gnattu: Also I could imagine that some LiveTV plugins can supply HLS streams with multiple resolution in one MediaSource and having the same URL, which probably relies on this currently.

dmitrylyzo: > Some containers like Matroska can have multiple video streams in the same file with the same path

Oh, those. AFAIK, we don't support multiple video streams. But if it will be implemented in the future, that's fine.

crobibero (approved):

PR #12628, opened 2024-09-11
Fix some analyzer IDisposable warnings

fice-t:
I haven't had much experience with C#, so I decided to start with fixing some of what the IDE analyzers were yelling about.

There are still remaining CA2000 violations, which fall into several categories:

* Required touching larger portions of code (such as `SKBitMap`)
* Not sure if a simple `(await) using` is appropriate (such as `CancellationTokenSource` and the log `Stream` for FFmpeg)
* Fixing the "recommended way" introduces an IDISP003 warning, though I believe it's an analyzer bug (`FromBase64Transform`, `Socket`)
* False positives (analyzer bugs for CA2000) (such as not recognizing that `ControllerBase.File` disposes its `Stream` argument, and something weird with `GetBitmapFromSvg` being part of a conditional expression).

**Changes**

These bring down the number of CA2000 violations that VS reports from 47 to 22, as well as remove the only CA2213 violation and add CA2213+CA2216 as build errors.

### DIFF
--- a/Emby.Server.Implementations/ApplicationHost.cs
+++ b/Emby.Server.Implementations/ApplicationHost.cs
@@ -608,6 +608,7 @@ namespace Emby.Server.Implementations
                 var localCert = new X509Certificate2(path, password, X509KeyStorageFlags.UserKeySet);
                 if (!localCert.HasPrivateKey)
                 {
+                    localCert.Dispose();
                     Logger.LogError("No private key included in SSL cert {CertificateLocation}.", path);
                     return null;
                 }
--- a/Emby.Server.Implementations/ApplicationHost.cs
+++ b/Emby.Server.Implementations/ApplicationHost.cs
@@ -161,8 +161,6 @@ namespace Emby.Server.Implementations
                 ConfigurationManager.Configuration,
                 ApplicationPaths.PluginsPath,
                 ApplicationVersion);
-
-            _disposableParts.Add(_pluginManager);
         }
 
         /// <summary>
@@ -971,6 +969,7 @@ namespace Emby.Server.Implementations
 
             if (dispose)
             {
+                _pluginManager?.Dispose();
                 var type = GetType();
 
                 Logger.LogInformation("Disposing {Type}", type.Name);
--- a/Jellyfin.Api/Controllers/VideosController.cs
+++ b/Jellyfin.Api/Controllers/VideosController.cs
@@ -425,7 +425,7 @@ public class VideosController : BaseJellyfinApiController
             EnableAudioVbrEncoding = enableAudioVbrEncoding
         };
 
-        var state = await StreamingHelpers.GetStreamingState(
+        using var state = await StreamingHelpers.GetStreamingState(
                 streamingRequest,
                 HttpContext,
                 _mediaSourceManager,
--- a/MediaBrowser.MediaEncoding/Attachments/AttachmentExtractor.cs
+++ b/MediaBrowser.MediaEncoding/Attachments/AttachmentExtractor.cs
@@ -124,7 +124,8 @@ namespace MediaBrowser.MediaEncoding.Attachments
         {
             using (await _semaphoreLocks.LockAsync(outputPath, cancellationToken).ConfigureAwait(false))
             {
-                if (!File.Exists(Path.Join(outputPath, id)))
+                var outputIdPath = Path.Join(outputPath, id);
+                if (!File.Exists(outputIdPath))
                 {
                     await ExtractAllAttachmentsInternal(
                         inputArgument,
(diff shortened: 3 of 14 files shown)


Jellyfin.Api/Controllers/VideosController.cs
```diff
@@ -425,7 +425,7 @@ [FromQuery] [RegularExpression(EncodingHelper.ValidationRegex)] string? subtitle
             EnableAudioVbrEncoding = enableAudioVbrEncoding
         };
 
-        var state = await StreamingHelpers.GetStreamingState(
+        using var state = await StreamingHelpers.GetStreamingState(
```
JPVenson, line 428: this will cause the stream to be closed if disposed here resulting in only the first chunk beeing transmitted

fice-t: I was concerned about that, but similar code in [`AudioHelper.GetAudioStream`](https://github.com/jellyfin/jellyfin/blob/81aca67745d541e276cfa05efe57e52e9f5a5d0e/Jellyfin.Api/Helpers/AudioHelper.cs#L88) uses `using`, which appears to be the audio counterpart to `GetVideoStream`. Is that a bug, or is it only an issue for the video case?

gnattu: Videos have transcode throttler which will have the need to manage the state and read from this state during the whole transcoding and this should only be disposed after the transcoding process has being killed or exited.

- StartFfMpeg assign the current TranscodingJob to StreamState
- Once StreamState is disposed, StreamState.TranscodingJob = null
- StreamState.ReportTranscodingProgress() cannot update TranscodingPositionTicks in current TranscodingJob
- job.TranscodingPositionTicks stuck at the initial value
- TranscodingThrottler query job.TranscodingPositionTicks -> IsThrottleAllowed == false

Basically I'm very against modifying code without understanding the consequence and just make "less warnings". Yes warnings sucks but cleanups like this should be made with certain level of confidence of the codebase.

Emby.Server.Implementations/ApplicationHost.cs
```diff
@@ -971,6 +970,7 @@ protected virtual void Dispose(bool dispose)
 
             if (dispose)
             {
+                _pluginManager?.Dispose();
```
gnattu, line 973: I don't think this change is needed though. We could just suppress this warning because it is disposed with other disposable parts.

MediaBrowser.MediaEncoding/Attachments/AttachmentExtractor.cs
```diff
@@ -134,7 +135,7 @@ await ExtractAllAttachmentsInternal(
 
                     if (Directory.Exists(outputPath))
                     {
-                        File.Create(Path.Join(outputPath, id));
+                        await File.Create(outputIdPath).DisposeAsync().ConfigureAwait(false);
```
gnattu, line 138: I think you can just close the file opened.

Jellyfin.Api/Controllers/VideosController.cs
fice-t: I understand that one shouldn't blindly modify stuff for the sole purpose of less warning noise, but I thought looking at some of the more straightforward warnings would be a small start to getting familiarity with the codebase for future work. I chose warnings regarding `IDisposable` as I read that they could perhaps be related to potential memory leaks in some scenarios.

I did leave half of the CA2000 warnings alone in part due to complexity; I was going to leave this one as well, but I saw **extremely** similar code in `GetAudioStream` and naively thought it would apply here as well. I'm sorry about that.

What is the simplest way to test the `GetVideoStream` path for transcodes? I've been playing around with transcode/network settings and transcoding appears to use the `DynamicHlsController` path.

MediaBrowser.MediaEncoding/Attachments/AttachmentExtractor.cs
fice-t: Probably, but the [documentation ](https://learn.microsoft.com/en-us/dotnet/api/system.io.filestream?view=net-8.0#methods) suggests that disposing is preferred over `Close`. Switching to a plain `Dispose` yields a [CA1849](https://learn.microsoft.com/en-ca/dotnet/fundamentals/code-analysis/quality-rules/ca1849) warning.

Emby.Server.Implementations/ApplicationHost.cs
fice-t: True, it would be disposed anyway. Still, `_disposableParts` is a mutable collection, so putting the dispose explicitly is a clearer guarantee that it is disposed, which is why the analyzer flagged it. I think a simple fix makes sense over an unnecessary suppression (which could be considered noise).

Bond-009:
Closing as stale

Issue #12630, opened 2024-09-12
10.10.0 latest build. QSV transcoding fails


soultaco83:
### This issue respects the following points:

- [X] This is a **bug**, not a question or a configuration issue; Please visit our forum or chat rooms first to troubleshoot with volunteers, before creating a report. The links can be found [here](https://jellyfin.org/contact/).
- [X] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [X] I'm using an up to date version of Jellyfin Server stable, unstable or master; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.
- [X] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [X] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Description of the bug

I have just completed building the lastest version of jellyfin for version 10.10.0.
Since this lastest pull I am unable to use the QSV transcoding option. All video files that require transcoding fail.
To note: Remux and direct are fine. Just the transcodes
I am also unable to select None as my Hardware acceleration choice as I get the following message on save: 
There was an error processing the request. Please try again later.

As of the build I have that last commit was
https://github.com/jellyfin/jellyfin/commit/751e12e5b5c59f9df4494251166f47c8cf9ebfe1

I am able to proceed without issue via Vappi and selecting /dev/dri/renderD128 as my render device

### Reproduction steps

1: Navigate to Dashboard
2: select Playback>Transcoding
3: If already on Intel Quicksync press save
4: Attempt a video transcode. 

### What is the current _bug_ behavior?

Transcode fails with 
Playback Error
Playback failed due to a fatal player error.

### What is the expected _correct_ behavior?

Video to transcode without failure

### Jellyfin Server version

Master

### Specify commit id

751e12e5b5c59f9df4494251166f47c8cf9ebfe1

### Specify the build version

Build version Release

### Environment

```markdown
- OS: Unraid
- Linux Kernel: Linux 6.8.12-Unraid x86_64
- Virtualization: Docker
- Clients: Desktop
- Browser: Chrome, Edge, firefox
- FFmpeg Version: 6.0.1-Jellyfin
- Playback Method: 
- Hardware Acceleration: QSV
- GPU Model: Intel Arc A770
- Plugins: Disabled other than base plugins
[...]

PR #12633, opened 2024-09-12
Fix QSV presets may be empty

nyanmisaka:
**Changes**
- Fix QSV presets may be empty

**Issues**
- #12630

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -1491,14 +1491,7 @@ namespace MediaBrowser.Controller.MediaEncoding
             {
                 EncoderPreset[] valid_presets = [EncoderPreset.veryslow, EncoderPreset.slower, EncoderPreset.slow, EncoderPreset.medium, EncoderPreset.fast, EncoderPreset.faster, EncoderPreset.veryfast];
 
-                if (valid_presets.Contains(encoderPreset))
-                {
-                    param += " -preset " + encodingOptions.EncoderPreset;
-                }
-                else
-                {
-                    param += " -preset " + EncoderPreset.veryfast.ToString().ToLowerInvariant();
-                }
+                param += " -preset " + (valid_presets.Contains(encoderPreset) ? encoderPreset : EncoderPreset.veryfast).ToString().ToLowerInvariant();
             }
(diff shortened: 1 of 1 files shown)


gnattu:
I don't think it is necessary after the web part is merged though.

cc @Shadowghost What do you think. Should we do explicit mapping or?

Shadowghost:
I don't think we should do explicit mappings if not necessary. Whole reason to use an enum is to not do this kind of thjngs

nyanmisaka:
I just don't get why the preset can be empty...

EraYaN:
> I don't think we should do explicit mappings if not necessary. Whole reason to use an enum is to not do this kind of things

I feel like explicit mapping is always a better idea, analyzers can alert you to missing cases and you never pass anything that is not explicitly supported.

gnattu:
> I just don't get why the preset can be empty...

It is due to the mistake made in input checking. It checks validity for encoding preset passed in as the function parameter which will always fallback to a valid value `superfast` which always passes the validity check, and then it passes the user config value which could be empty due to the client sent incorrect value which is not a valid enum value, and then passed to ffmpeg as empty string.

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -1489,16 +1489,16 @@ private string GetEncoderParam(EncoderPreset? preset, EncoderPreset defaultPrese
                      || string.Equals(videoEncoder, "hevc_qsv", StringComparison.OrdinalIgnoreCase) // hevc (hevc_qsv)
                      || string.Equals(videoEncoder, "av1_qsv", StringComparison.OrdinalIgnoreCase)) // av1 (av1_qsv)
             {
-                EncoderPreset[] valid_presets = [EncoderPreset.veryslow, EncoderPreset.slower, EncoderPreset.slow, EncoderPreset.medium, EncoderPreset.fast, EncoderPreset.faster, EncoderPreset.veryfast];
-
-                if (valid_presets.Contains(encoderPreset))
-                {
-                    param += " -preset " + encodingOptions.EncoderPreset;
```
gnattu, line 1496: Just change this line to pass the encoderPreset as is should be fine if I understand it correctly

Shadowghost: This should likely fix it, yes.

gnattu (approved):

Shadowghost (approved):

crobibero (approved):

PR #12639, opened 2024-09-13
Test StreamBuilder more precisely & improve the transcode reason

gnattu:
The unit tests for `StreamBuilder` is a bit outdated and does not reflect the actual results of all clients. This PR updates both  device profiles and the test cases to make the unit tests reflects what the actual browsers may expecting. During the test case updating, I also noticed that our current transcode reason is not accurate enough to write a reasonable test case, and I updated how we derive transcode reasons from the direct play profiles, the details are explained below. Additional test cases using Hi10P video that utilizes the newly added `SubContainer` field introduced in #12420 for remux and an mkv file with broken fps detection described in  #12603 is also added.

For all major browsers (Chrome, Firefox, Safari), I updated their device profile to be the same as what a 10.10 client would look like. For other clients' device profiles, I modified the test cases so that their current profile and the server behavior is respected.

We previously was using the first direct play profile's transcoding reason as the conclusive transcoding reason, but that is somewhat arbitrary, as many clients' first profile may not be the most compatible one. For instance, browsers often set WebM as the first profile, which doesn’t support common codecs like H.264 and AAC by design. This causes `VideoCodecNotSupported` and `AudioCodecNotSupported` to be returned, even if the browser supports those codecs. Only use those reasons when all profiles indicate that the codec is not supported.

**Changes**

- Update StreamBuilder to return more precise transcode reasons
- Add more StreamBuilder test cases
- Update StreamBuilder test cases 

**Issues**


dmitrylyzo:
> We previously was using the first direct play profile's transcoding reason as the conclusive transcoding reason, but that is somewhat arbitrary, as many clients' first profile may not be the most compatible one.

Not just the first, but the first of the ordered compatibility. Ideally, minimal. But yeah, it probably can't be done in one pass.

Reason doesn't really mean why it's transcoding, but why it's not direct playing.
It will still be "wrong" because of combining all the reasons from different containers. For example, DoVi in MP4 vs MKV or maximum audio channels in FLAC vs MP4.

If we need the actual reason for transcoding, we can get it from this function:
https://github.com/jellyfin/jellyfin/blob/edc15c8e923610c7074328f1950d18f01926d552/MediaBrowser.Model/Dlna/StreamBuilder.cs#L806

But I think we still need a reason why it is not direct playing.

gnattu:
> Reason doesn't really mean why it's transcoding, but why it's not direct playing.

Then the unit tests are even more wrong because it is tested as why it's transcoding

tests/Jellyfin.Model.Tests/Dlna/StreamBuilderTests.cs
```diff
@@ -86,18 +92,6 @@ public class StreamBuilderTests
         [InlineData("JellyfinMediaPlayer", "mkv-vp9-aac-srt-2600k", PlayMethod.DirectPlay)] // #6450
         [InlineData("JellyfinMediaPlayer", "mkv-vp9-ac3-srt-2600k", PlayMethod.DirectPlay)] // #6450
         [InlineData("JellyfinMediaPlayer", "mkv-vp9-vorbis-vtt-2600k", PlayMethod.DirectPlay)] // #6450
-        // Chrome-NoHLS
```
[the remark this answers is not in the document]
gnattu: > MKV

I don't think it is suitable and if any clients are using that should stop using that. ffmpeg does not treat mkv as a streamable container and will write back to previously written location (meaning, change already written sections) and that is impossible to do the same thing with streaming and may cause playback issues.

dmitrylyzo:
> Then the unit tests are even more wrong because it is tested as why it's transcoding

You just fixed one (`EnableDirectStream` bug): https://github.com/jellyfin/jellyfin/blob/edc15c8e923610c7074328f1950d18f01926d552/tests/Jellyfin.Model.Tests/Dlna/StreamBuilderTests.cs#L144
Others are looking fine, as "why it is not direct playing".

> Why it is not direct playing cannot be summarized by a list of reasons because it is a result of testing all possible container and codec and other profile combinations and all of those failed and each combination will have a different reason especially when the original container is not supported at all and all direct play profiles are failing. Yes combining all profiles could still result in incorrect reasons but it avoids the most obvious ones.

In theory, we could just do:
- container is not supported
  - audio codec is not supported
    - or audio codec has incompatible attributes
  - video codec is not supported
    - or video codec has incompatible attributes

Actually, codecs are currently checked against the original container.
https://github.com/jellyfin/jellyfin/blob/7d178aad0ba3bdfbebb61e40e1cb02895188a9a9/MediaBrowser.Model/Dlna/StreamBuilder.cs#L1296
So, the reason combining will probably not overfill the final reason.

tests/Jellyfin.Model.Tests/Dlna/StreamBuilderTests.cs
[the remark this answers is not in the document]
dmitrylyzo: AFAIK, there is currently no client that uses progressive streaming.
NoHLS profile uses MP4 and WebM (_same issue as MKV?_).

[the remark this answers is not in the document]
gnattu: The most "safe" option would be mpegts if you want me to say, mp4 works better than mkv and I'm not very clear if ffmpeg's webm implementation is stream friendly enough or not

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -1415,16 +1415,49 @@ private static int GetMaxAudioBitrateForTotalBitrate(long totalBitrate)
...
+
+            if (failureReasons.FirstOrDefault() == 0)
+            {
+                inferredReason = TranscodeReason.DirectPlayError;
+            }
+            else
             {
-                failureReasons = TranscodeReason.DirectPlayError;
+                var videoCodecNotSupportedCount = failureReasons.Count(r => (r & TranscodeReason.VideoCodecNotSupported) != 0);
+                var audioCodecNotSupportedCount = failureReasons.Count(r => (r & TranscodeReason.AudioCodecNotSupported) != 0);
```
crobibero, line 1433:
```suggestion
                var videoCodecNotSupportedCount = failureReasons.Count(r => r.HasFlag(TranscodeReason.VideoCodecNotSupported));
                var audioCodecNotSupportedCount = failureReasons.Count(r => r.HasFlag(TranscodeReason.AudioCodecNotSupported));
```

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -1415,16 +1415,49 @@ private static int GetMaxAudioBitrateForTotalBitrate(long totalBitrate)
                 return profileMatch;
             }
 
+            TranscodeReason inferredReason = 0;
+
             var failureReasons = analyzedProfiles[false]
                 .Select(analysis => analysis.Result)
                 .Where(result => !containerSupported || (result.TranscodeReason & TranscodeReason.ContainerNotSupported) == 0)
```
crobibero, line 1422:
```suggestion
                .Where(result => !containerSupported || result.TranscodeReason.HasFlag(TranscodeReason.ContainerNotSupported))
```

Wasn't changed in this PR but this should also be `HasFlag` for readability

crobibero, line 1422: Wasn't changed in this PR but this line could be the following for readability 

```
.Where(result => !containerSupported || !result.TranscodeReason.HasFlag(TranscodeReason.ContainerNotSupported))
```

gnattu:
> Btw, I suppose the WebM bug you mentioned only happens for MKV labeled as mkv,webm before https://github.com/jellyfin/jellyfin/pull/10772

No, it happens for all those test cases which does not have `webm` marked alongside `mkv` as well.

> It might be a good idea to drop the webm label when fetching from DB. I was afraid that this could harm the performance. But this would eliminate the need for rescanning.

It is worse than you think it is. Our "smart" pipeline will try to add the `mkv` to client profiles when they claim they do support `webm`, and if all codecs are webm supported codec it will return as if the file can be direct played. For most files it seems to work in Firefox and Chrome, but I do not like such behavior.

dmitrylyzo:
> No, it happens for all those test cases which does not have `webm` marked alongside `mkv` as well.

Could you add a test for such case?
Because I only saw this for MKV+H264+\<any audio\>: `containerSupported: true` (because also WebM).

> Our "smart" pipeline will try to add the `mkv` to client profiles when they claim they do support `webm`, and if all codecs are webm supported codec in the requested video file

In the server? :eyes: 
Iirc, this was in the web client, but we decided not to play with fire (specs) and removed MKV.

gnattu:
> Could you add a test for such case?

The `"Firefox", "mp4-h264-hi10p-aac-5000k"` case already has described problematic behavior despite the original video is in mp4. The mkv hi10p case will also have similar result for safari despite the video file only has `mkv` in container info and does not have `webm` there.

> In the server? 👀

Yes, and I noticed this in the unit test and that is why the "Chrome", "mkv-vp9-vorbis-vtt-2600k"` case is direct playing instead of direct streaming (fmp4 does not support vorbis, so audio has to be transcoded)

tests/Jellyfin.Model.Tests/Dlna/StreamBuilderTests.cs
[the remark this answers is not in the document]
gnattu: Added back the tests as-is, only specifying the play method and explicitly set http protocol. It should be good enough for now.

tests/Jellyfin.Model.Tests/Dlna/StreamBuilderTests.cs
```diff
@@ -147,29 +156,46 @@ public class StreamBuilderTests
...
+        [InlineData("Tizen4-4K-5.1", "mp4-hevc-truehd-srt-15200k", PlayMethod.Transcode, TranscodeReason.AudioCodecNotSupported)]
         [InlineData("Tizen4-4K-5.1", "mkv-vp9-aac-srt-2600k", PlayMethod.DirectPlay)]
         [InlineData("Tizen4-4K-5.1", "mkv-vp9-ac3-srt-2600k", PlayMethod.DirectPlay)]
         [InlineData("Tizen4-4K-5.1", "mkv-vp9-vorbis-vtt-2600k", PlayMethod.DirectPlay)]
+        // WebOS 23
+        [InlineData("WebOS-23", "mkv-dvhe.08-eac3-15200k", PlayMethod.Transcode, TranscodeReason.VideoRangeTypeNotSupported, "Remux")]
+        [InlineData("WebOS-23", "mp4-dvh1.05-eac3-15200k", PlayMethod.DirectPlay)]
+        [InlineData("WebOS-23", "mp4-dvhe.08-eac3-15200k", PlayMethod.DirectPlay)]
+        [InlineData("WebOS-23", "mkv-dvhe.05-eac3-28000k", PlayMethod.Transcode, TranscodeReason.VideoRangeTypeNotSupported, "Remux")]
+        // Non-HLS Progressive transcoding
```
dmitrylyzo, line 187: Could you move `Non-HLS Progressive transcoding` block before `TranscodeMedia` - to where it was before? Less change.

tests/Jellyfin.Model.Tests/Dlna/StreamBuilderTests.cs
```diff
@@ -412,7 +439,7 @@ public async Task BuildVideoItemWithDirectPlayExplicitStreams(string deviceName,
                 // Full transcode
                 if (transcodeMode.Equals("Transcode", StringComparison.Ordinal))
                 {
-                    if ((streamInfo.TranscodeReasons & (StreamBuilder.ContainerReasons | TranscodeReason.DirectPlayError)) == 0)
+                    if ((streamInfo.TranscodeReasons & (StreamBuilder.ContainerReasons | TranscodeReason.DirectPlayError | TranscodeReason.VideoRangeTypeNotSupported)) == 0)
```
dmitrylyzo, line 442: Tbf, this block looks weird.
We are going to transcode. Why can't we use the same codec?

gnattu: It was an old test though, where the transcoding usually happens when input codec is not supported and the transcoding reason video codec not supported is not conclusive enough to determine if it should use another codec. Do we want to change this?

dmitrylyzo: I tried:
```cs
if (why.HasFlag(TranscodeReason.VideoCodecNotSupported))
{
    Assert.All(
        videoStreams,
        stream => Assert.DoesNotContain(stream.Codec, streamInfo.VideoCodecs));
}
else
{
    Assert.Contains(targetVideoStream?.Codec, streamInfo.VideoCodecs);
}
```
But `SecondaryAudioNotSupported` and `Full transcode because profile only has ts which does not allow hevc`.
To fix this block, we need to pass the real/full reason or transcoding codec (like `transcodeProtocol`).

gnattu: > To fix this block, we need to pass the real/full reason or transcoding codec (like transcodeProtocol).

We don't have to make it that complicated I think. If the transcode reason that failed the direct play back does make sense we can give a list of reasons that the original codec must be avoided and check those, and don't assume the new codec is same as the original if transcode reason is not any of those due to complex compatibility reasons.

dmitrylyzo: Or we can pass `transcodeVideoCodec` and test with it directly.

We already have `transcodeProtocol` (`http`, `HLS.ts`, `HLS.mp4`).

Hmm, we can also introduce a simple struct with protocol, video and audio codecs. This way we can fill it partially.
```cs
new OutputStream { .Protocol = "HLS.mp4", .AudioCodec = "aac" }
```

Maybe left it for the next PR? 🤔

gnattu: Probably yes. Let's limit the scope of this PR to test for the current things, and then move on to use explicit output codecs instead of guessing from transcoding reasons in following PRs.

dmitrylyzo (approved):
Technically, this is also an improvement of the transcode reason. I mean the PR title.

Shadowghost (approved):

crobibero (approved):

Issue #11343, opened 2024-04-11
[Issue]: Jellyfin rclone mount high bandwidth issue


maulik9898:
### Please describe your bug

When playback stops, Jellyfin continues to read the file from the rclone mount, resulting in high bandwidth usage. The Jellyfin process keeps the file open, which I verified using the lsof command. This behavior leads to very high bandwidth consumption.

### Reproduction Steps

- Start playing any movie or show
- check bandwidth usage using `cbm`
- stop playback 
- check bandwidth usage. still fetching data from rclone. chekc for open file using `lsof | grep filename_or_jellyfin`

### Jellyfin Version

10.8.13

### Environment

```markdown
- OS: ubuntu
- Linux Kernel:
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method: direct
- Hardware Acceleration: none
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

```shell
[14:53:00] [INF] [1] Main: Jellyfin version: 10.8.13
[14:53:00] [INF] [1] Main: Environment Variables: ["[JELLYFIN_PublishedServerUrl, https://jelly.maulik.tech]", "[JELLYFIN_LOG_DIR, /config/log]", "[JELLYFIN_DATA_DIR, /config]", "[JELLYFIN_FFMPEG, /usr/lib/jellyfin-ffmpeg/ffmpeg]", "[JELLYFIN_CONFIG_DIR, /config/config]", "[JELLYFIN_CACHE_DIR, /cache]", "[JELLYFIN_WEB_DIR, /jellyfin/jellyfin-web]", "[DOTNET_SYSTEM_GLOBALIZATION_INVARIANT, 1]"]
[14:53:00] [INF] [1] Main: Arguments: ["/jellyfin/jellyfin.dll"]
[14:53:00] [INF] [1] Main: Operating system: Linux
[14:53:00] [INF] [1] Main: Architecture: Arm64
[14:53:00] [INF] [1] Main: 64-Bit Process: True
[14:53:00] [INF] [1] Main: User Interactive: True
[14:53:00] [INF] [1] Main: Processor count: 4
[14:53:00] [INF] [1] Main: Program data path: /config
[14:53:00] [INF] [1] Main: Web resources path: /jellyfin/jellyfin-web
[14:53:00] [INF] [1] Main: Application directory: /jellyfin/
[...]

PR #12641, opened 2024-09-13
Use FileStream to read static media files

3flex:
**Changes**
Use a `FileStream` to read static media files.

This solves three issues when reading from slow I/O sources:
1. Current implementation uses maximum available bandwidth to download files from slow I/O sources even if the media doesn't require it. For example using DirectPlay on a file with ~15Mbit average bitrate would cause the read with network I/O to saturate the 100Mbps link. With this change the bandwidth used is appropriate for the bitrate required.
2. Current implementation would not stop downloading from source when streaming from client was stopped, using data unnecessarily.
3. File handles were not released when streaming stopped.

I don't have .NET experience so apologies if this causes unintended issues but in local testing it solves the issues with current implementation.

**Issues**
#11343

### DIFF
--- a/Jellyfin.Api/Helpers/FileStreamResponseHelpers.cs
+++ b/Jellyfin.Api/Helpers/FileStreamResponseHelpers.cs
@@ -56,7 +56,8 @@ public static class FileStreamResponseHelpers
         string path,
         string contentType)
     {
-        return new PhysicalFileResult(path, contentType) { EnableRangeProcessing = true };
+        FileStream fs = File.OpenRead(path);
+        return new FileStreamResult(fs, contentType) { EnableRangeProcessing = true };
     }
 
     /// <summary>


JPVenson:
> Current implementation uses maximum available bandwidth to download files from slow I/O sources even if the media doesn't require it. For example using DirectPlay on a file with ~15Mbit average bitrate would cause the read with network I/O to saturate the 100Mbps link. With this change the bandwidth used is appropriate for the bitrate required.

Using a filestream instead does not change that a bit.

> Current implementation would not stop downloading from source when streaming from client was stopped, using data unnecessarily.

`PhysicalFileResult` uses a stream internally anyway, https://github.com/dotnet/aspnetcore/blob/main/src/Http/Http.Extensions/src/SendFileResponseExtensions.cs#L25 that does not change the fact that rclone will load the whole source file anyway.

> File handles were not released when streaming stopped.
File handles are closed as soon as the request is and the data has been send. If that would not be the case, millions of apps currently running in production would face the same issues.

jumoog:
>  However, it seems like it is either the FUSE implementation of rclone has a big compatibility issue with that call, or dotnet's implementation using that call has some quirks that file descriptor is never closed.

Okay, now it makes sense that I had the same problem with `GocryptFS`

gnattu:
> Okay, now it makes sense that I had the same problem with `GocryptFS`

Can you try this PR and see if that workaround fixed your issue as well? If so I have no objection to make this a user-configurable workaround, just don't enable it by default due to possible performance regressions.

3flex:
@JPVenson thanks for reviewing. I should have said that those three points describe my experience and the contrast before and after this change - I'm not sure what else to tell you except that this solves my issues nicely.

Like I said I'm not a .NET dev but without this it makes jellyfin practically unusable for me when using DirectPlay.

@gnattu putting this behind an env var flag would be fine for me but beyond my capabilities to implement. As an alternative, is there a way to detect that the file is on a FUSE mount and use a FileStream automatically? Then no config would be required.

jumoog:
I'll do the Tests over the weekend. But I think we can can call `statfs` (asuming we have `libc`) to find out if the FS is a FUSE

gnattu:
FUSE might not always cause this issue though as we have tons of users using mergerfs and does not have such issue, so a manual flag is probably not avoidable.

jumoog:
I did some tests with the current version of jellyfin. When I use DirectPlay, everything works as intended. As soon as I stop playback. I can see that the huge bandwidth to my NAS is gone. When I noticed this behaviour, it may have been because I was playing a video with DTS audio in the web player and Jellyfin was rencoding the audio in the background.

RunAway189:
It has asked me to provide my rclone mount flags to allow the team investigating more into this issue. Here they are, if you need anything else just ask me and thanks for your precious time.

[rclone_mount_flags.txt](https://github.com/user-attachments/files/17003514/rclone_mount_flags.txt)

RunAway189:
@3flex Hey man, I've built the latest 10.10.1 with your fix and it seems it doesn't work anymore. Could you please verify on your side as well? Thanks in advance

3flex:
Not sure, I've dropped this patch from my 10.10  setup for now. Need to retest the original issue and see if I'm still seeing it since the update & other local config changes but not sure when that will happen.

RunAway189:
> Our team think this approach might be too hacky and want to implement a more proper fix. However @JPVenson failed to reproduce to make deeper investigation on this problem. He wants full recline config and mount config, basically everything other than the credentials. I personally think hacks like this is fine as long as it is controlled by an env var and disabled by default and actually solves or at least, workarounds problems.

I've already shared my mount flags. That's how I mount my rclone, it worked perfectly for a lot of time but right now I've just realised it doesn't work anymore on my side.

PR #12643, opened 2024-09-13
Tweak Trickplay migration for speed

crobibero:
The `MoveTrickplayFiles` migration took 25 minutes on my server, and 35s on my development machine.

This PR reduces the time on my development machine to 11s which should make a large difference on my server but I'm not able to easily deploy 1-off builds


Shadowghost (approved):

Jellyfin.Server/Migrations/Routines/MoveTrickplayFiles.cs
```diff
@@ -42,26 +50,50 @@ public MoveTrickplayFiles(ITrickplayManager trickplayManager, IFileSystem fileSy
...
                 if (_fileSystem.DirectoryExists(oldPath))
                 {
                     _fileSystem.MoveDirectory(oldPath, newPath);
                 }
             }
-        }
+        } while (previousCount != 0);
+
+        _logger.LogInformation("Moved {Count} items in {Time}", itemCount, sw.Elapsed);
+        throw new FfmpegException("Don't save me!");
```
nielsvanvelzen, line 96: I guess this is debugging code?

crobibero: :cold_sweat: no idea what you're talking about

EraYaN:
If you want to speed up a full enumeration in chunks, instead of using skip which is quite slow at later chunks at it does evaluate the whole query for the skipped items. Keyset pagination is much better (using `>= <last value form previous chunk>`) especially if you order on an index.

Something like https://github.com/mrahhal/MR.EntityFrameworkCore.KeysetPagination

We use it in production and it's quite nice.

crobibero:
> If you want to speed up a full enumeration in chunks, instead of using skip which is quite slow at later chunks at it does evaluate the whole query for the skipped items. Keyset pagination is much better (using `>= <last value form previous chunk>`) especially if you order on an index.
> 
> Something like [mrahhal/MR.EntityFrameworkCore.KeysetPagination](https://github.com/mrahhal/MR.EntityFrameworkCore.KeysetPagination)
> 
> We use it in production at work and it's quite nice and way way faster.

This looks like a good library to have in our back pocket but unnecessary for the ~2 times this code will be run

EraYaN:
> This looks like a good library to have in our back pocket but unnecessary for the ~2 times this code will be run

Ooh sure, just what we are using for more complicated keysets) but for a single Id that are sortable you can just do it with a `.Where` and passing in the last seen id.

crobibero:
> > This looks like a good library to have in our back pocket but unnecessary for the ~2 times this code will be run
> 
> Ooh sure, (just what we are using for more complicated keysets) but for a single Id that are sortable you can just do it with a `.Where` and passing in the last seen id. If you want to do perf optimizations than you might as well, it doesn't add much complexity (arguably even having fewer parameters in your function).

Just tested changing the query to use the last seen id as an offset and the result for 50k items on my dev machine was within 100ms, so I don't think the change is needed

JPVenson (approved):

PR #12644, opened 2024-09-14
Better bitrate and resolution normalization

gnattu:
Previously we don't cap how much we want to scale the bitrate when transcoding to a less efficient codec like h264, resulting in extremely undesired high bitrate from high bitrate original videos, for example 80M AV1 will produce 120M H.264 file which overloads the server encoder, the client decoder and probably also the network. This set a scaling cap at 30Mbit, which means we don't scale up bitrate after 30M anymore, so that the bitrate will not be significantly larger than original if the original video has a high enough bitrate. For videos at such high bitrate the result is hardly noticeable (or not noticeable at all) with our current prefer speed encoding presets.

The previous resolution normalization is also outdated and not well-suited for modern advanced codecs and UHD videos. It is overly conservative with the use of high resolutions. This normalizes resolution while taking into account codec efficiency, frame rate, and dynamic range, providing in a more appropriate resolution config. All resolution configurations are now normalized based on the bitrate for H.264 in SDR at 30fps. The reference bitrate used to determine the resolution will be adjusted accordingly for codec, frame rate, and dynamic range (HDR reserved for future use, we are not doing HDR transcoding now). As a result, users will see higher resolutions more frequently when bitrate caps are applied.

The old logic of "using the previous configuration if the max bitrate is reached," inherited from Emby also didn't make much sense to me. I've changed it to "if the reference bitrate is below this configuration's max value, then use this configuration" which makes more sense for a "maxBitrate".

We probably also want to change the web texts based on this change as the resolution texts on web will now make almost no sense and even more wrong. In extreme condition now the text can say 720p 8Mbps on web and the video is streamed in 1440p which is a 2x resolution difference.

The new resolution config's max bitrate is taken from [Apple's HLS document](https://developer.apple.com/documentation/http-live-streaming/hls-authoring-specification-for-apple-devices#Video) with modifications to fit our own scaling. 

**Changes**

- Don't scale bitrate >= 30Mbps
- Normalize Resolution better

**Issues**

See recent post on forum: https://forum.jellyfin.org/t-solved-jellyfin-bitrate-is-way-to-high-on-firefox

### DIFF
--- a/MediaBrowser.Model/Dlna/ResolutionNormalizer.cs
+++ b/MediaBrowser.Model/Dlna/ResolutionNormalizer.cs
@@ -47,7 +47,7 @@ namespace MediaBrowser.Model.Dlna
 
             if (isHdr)
             {
-                referenceBitrate *= 1.2f;
+                referenceBitrate *= 0.8f;
             }
 
             var resolutionConfig = GetResolutionConfiguration(Convert.ToInt32(referenceBitrate));
--- a/MediaBrowser.Model/Dlna/ResolutionNormalizer.cs
+++ b/MediaBrowser.Model/Dlna/ResolutionNormalizer.cs
@@ -2,28 +2,33 @@
 #pragma warning disable CS1591
 
 using System;
+using System.Linq;
 
 namespace MediaBrowser.Model.Dlna
 {
     public static class ResolutionNormalizer
     {
-        private static readonly ResolutionConfiguration[] Configurations =
-            new[]
-            {
-                new ResolutionConfiguration(426, 320000),
-                new ResolutionConfiguration(640, 400000),
-                new ResolutionConfiguration(720, 950000),
-                new ResolutionConfiguration(1280, 2500000),
-                new ResolutionConfiguration(1920, 4000000),
-                new ResolutionConfiguration(2560, 20000000),
-                new ResolutionConfiguration(3840, 35000000)
-            };
+        // Please note: all bitrate here are in the scale of SDR h264 bitrate at 30fps
+        private static readonly ResolutionConfiguration[] _configurations =
+        [
+            new ResolutionConfiguration(416, 365000),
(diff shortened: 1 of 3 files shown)


Shadowghost (approved):

systemcrash:
This should probably be an independent setting, just as we have max LAN bitrate. How about an e.g. +15% ceiling when using an older generation codec, with a max of the LAN bandwidth (or some nominal factor less than that)?

Capping at 30Mbit isn't unreasonable, but as bandwidth (requirements/availability) grows over time, this static value will become a limiting factor.

gnattu:
You are not understanding what this PR is doing. This PR only prevents the original 80Mbps file to be transcoded as 120Mbps file, and now it will transcode as 80Mbps as is. **It does not place hard cap on transcoding bitrate**

systemcrash:
Ah, of course - scale factor 🤦 

Thanks.

MediaBrowser.Model/Dlna/ResolutionNormalizer.cs
```diff
@@ -38,16 +43,26 @@ public static ResolutionOptions Normalize(
                 }
             }
 
-            var resolutionConfig = GetResolutionConfiguration(outputBitrate);
-            if (resolutionConfig is not null)
+            var referenceBitrate = h264EquivalentOutputBitrate * (30.0f / (targetFps ?? 30.0f));
```
dmitrylyzo, line 46: Can `targetFps` be `0` (due to a bug)?

gnattu: I would say just let it throw and fail in this case. This either means:

- the client is requesting 0fps explicitly

Or

- the source video is probed as 0fps

And either way the pipeline will break more than just resolution normalization.

crobibero (approved):

nyanmisaka (approved):
Tested it a few days ago.

Issue #12657, opened 2024-09-16
Jellyfin not hardware decoding yuv444p10


1000thSummer:
### This issue respects the following points:

- [X] This is a **bug**, not a question or a configuration issue; Please visit our forum or chat rooms first to troubleshoot with volunteers, before creating a report. The links can be found [here](https://jellyfin.org/contact/).
- [X] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [X] I'm using an up to date version of Jellyfin Server stable, unstable or master; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.
- [X] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [X] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Description of the bug

Jellyfin version 10.9.11
CPU: intel n100
Platform: win10

While playing HEVC yuv444p10le encoded videos, transcoding was invoked as my client (edge browser) does not support the codec.

Jellyfin seems to go below process:
- QSV hardware decoding
- use CPU to do ~color space conversion~ chroma downsampling (yuv444p10 to yuv420)
- QSV hardware encoding

The second step uses a lot of CPU whereas QSV already provide `vpp_qsv` filters that does the GPU hardware processing

i.e., when transcoding locally, I use below cmd that barely uses CPU:
```sh
.\ffmpeg.exe -hwaccel qsv -i file:"<input_file_name>" -vf 'hwupload=extra_hw_frames=64,vpp_qsv=format=p010le' -c:a aac -b:a 128k -c:v hevc_qsv -preset slow -profile:v main10 -global_quality 23 "<output_file_name>"
```

Expected behavior:
The ffmpeg cmd that jellyfin uses should be smart enough to try `vpp_qsv` filters when possible.

### Reproduction steps

1. transcoding any YUV444 encoded video.
2. observe the extreme CPU usage

### What is the current _bug_ behavior?

very high CPU usage because of ~color space conversion~ chroma downsampling

### What is the expected _correct_ behavior?

very little CPU usage during ~conversion~ chroma downsampling

### Jellyfin Server version

10.9.11+

### Specify the build version

10.9.11

### Environment

```markdown
- OS:win10
- Linux Kernel:N/A
- Virtualization:No
[...]

PR #12664, opened 2024-09-17
Enable HEVC RExt HW decoding for 4:2:2/4:4:4 content

nyanmisaka:
**Changes**
- Enable HEVC RExt HW decoding for 4:2:2/4:4:4 content

**Issues**
- Fixes https://github.com/jellyfin/jellyfin/issues/12657

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -377,7 +377,7 @@ namespace MediaBrowser.Controller.MediaEncoding
         {
             if (state.VideoStream is null
                 || !options.EnableVideoToolboxTonemapping
-                || GetVideoColorBitDepth(state) != 10)
+                || GetVideoColorBitDepth(state) < 10)
             {
                 return false;
             }
@@ -6496,7 +6496,10 @@ namespace MediaBrowser.Controller.MediaEncoding
                 || string.Equals("yuv422p", videoStream.PixelFormat, StringComparison.OrdinalIgnoreCase)
                 || string.Equals("yuv444p", videoStream.PixelFormat, StringComparison.OrdinalIgnoreCase)
                 || string.Equals("yuv422p10le", videoStream.PixelFormat, StringComparison.OrdinalIgnoreCase)
-                || string.Equals("yuv444p10le", videoStream.PixelFormat, StringComparison.OrdinalIgnoreCase);
+                || string.Equals("yuv444p10le", videoStream.PixelFormat, StringComparison.OrdinalIgnoreCase)
+                || string.Equals("yuv420p12le", videoStream.PixelFormat, StringComparison.OrdinalIgnoreCase)
+                || string.Equals("yuv422p12le", videoStream.PixelFormat, StringComparison.OrdinalIgnoreCase)
+                || string.Equals("yuv444p12le", videoStream.PixelFormat, StringComparison.OrdinalIgnoreCase);
 
             // The related patches make videotoolbox hardware surface working is only available in jellyfin-ffmpeg 7.0.1 at the moment.
             bool useHwSurface = (_mediaEncoder.EncoderVersion >= _minFFmpegWorkingVtHwSurface) && IsVideoToolboxFullSupported();
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -6001,7 +6001,11 @@ namespace MediaBrowser.Controller.MediaEncoding
             var decoderName = decoderPrefix + '_' + decoderSuffix;
 
             var isCodecAvailable = _mediaEncoder.SupportsDecoder(decoderName) && options.HardwareDecodingCodecs.Contains(videoCodec, StringComparison.OrdinalIgnoreCase);
-            if (bitDepth == 10 && isCodecAvailable)
+
+            // VideoToolbox decoders have built-in SW fallback
+            if (bitDepth == 10
+                && isCodecAvailable
+                && (options.HardwareAccelerationType != HardwareAccelerationType.videotoolbox))
             {
                 if (string.Equals(videoCodec, "hevc", StringComparison.OrdinalIgnoreCase)
                     && options.HardwareDecodingCodecs.Contains("hevc", StringComparison.OrdinalIgnoreCase)
@@ -6077,7 +6081,9 @@ namespace MediaBrowser.Controller.MediaEncoding
                 && ffmpegVersion >= _minFFmpegDisplayRotationOption;
             var stripRotationDataArgs = stripRotationData ? " -display_rotation 0" : string.Empty;
 
-            if (isCodecAvailable)
+            // VideoToolbox decoders have built-in SW fallback
+            if (isCodecAvailable
+                && (options.HardwareAccelerationType != HardwareAccelerationType.videotoolbox))
             {
                 if (string.Equals(videoCodec, "hevc", StringComparison.OrdinalIgnoreCase)
                     && options.HardwareDecodingCodecs.Contains("hevc", StringComparison.OrdinalIgnoreCase))
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -299,7 +299,7 @@ namespace MediaBrowser.Controller.MediaEncoding
         {
             if (state.VideoStream is null
                 || !options.EnableTonemapping
-                || GetVideoColorBitDepth(state) != 10
+                || GetVideoColorBitDepth(state) < 10
                 || !_mediaEncoder.SupportsFilter("tonemapx"))
             {
                 return false;
@@ -312,7 +312,7 @@ namespace MediaBrowser.Controller.MediaEncoding
         {
             if (state.VideoStream is null
                 || !options.EnableTonemapping
-                || GetVideoColorBitDepth(state) != 10)
+                || GetVideoColorBitDepth(state) < 10)
             {
                 return false;
             }
@@ -354,7 +354,7 @@ namespace MediaBrowser.Controller.MediaEncoding
         {
             if (state.VideoStream is null
                 || !options.EnableVppTonemapping
-                || GetVideoColorBitDepth(state) != 10)
+                || GetVideoColorBitDepth(state) < 10)
             {
                 return false;
             }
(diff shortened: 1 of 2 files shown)


Shadowghost (approved):

nyanmisaka:
@gnattu It seems that VideoToolbox [in FFmpeg](https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/hevc/hevcdec.c#L432-L545) does not support 12-bit HEVC decode, right?

1000thSummer:
hi @nyanmisaka that was so fast lol.
Not familiar with dotnet stuff, am I suppose to build jellyfin binary off this branch and replace the existing one?

nyanmisaka:
> hi @nyanmisaka that was so fast lol. Not familiar with dotnet stuff, am I suppose to build jellyfin binary off this branch and replace the existing one?

You can refer to [jellyfin-packaging](https://github.com/jellyfin/jellyfin-packaging). Just make sure to checkout the jellyfin and jellyfin-web commits for these two PRs.

gnattu:
That may not make it into 10.10 though because technically it has missed the 10.10 feature freeze window, so probably another 6 month if we decide not to include that in 10.10.

1000thSummer:
> > hi @nyanmisaka that was so fast lol. Not familiar with dotnet stuff, am I suppose to build jellyfin binary off this branch and replace the existing one?
> 
> You can refer to [jellyfin-packaging](https://github.com/jellyfin/jellyfin-packaging). Just make sure to checkout the jellyfin and jellyfin-web commits for these two PRs.

Good to learn, thanks!

1000thSummer:
> That may not make it into 10.10 though because technically it has missed the 10.10 feature freeze window, so probably another 6 month if we decide not to include that in 10.10.

hopefully it catches the train :) . If not will there be nightly build (not stable release) that contains this feature?

nyanmisaka:
> > That may not make it into 10.10 though because technically it has missed the 10.10 feature freeze window, so probably another 6 month if we decide not to include that in 10.10.
> 
> hopefully it catches the train :) . If not will there be nightly build (not stable release) that contains this feature?

There are weekly builds in the unstable channel. https://jellyfin.org/downloads/server

gnattu:
https://github.com/jellyfin/jellyfin/blob/41ac5f8d76ce11a852e4dafbf20ad57d63d55f96/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs#L376-L383

VideoToolbox native tonemap accepts 12bit inputs, need change it to `< 10bit` as well

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -6414,6 +6492,11 @@ public string GetVideotoolboxVidDecoder(EncodingJobInfo state, EncodingOptions o
             var is8bitSwFormatsVt = string.Equals("yuv420p", videoStream.PixelFormat, StringComparison.OrdinalIgnoreCase)
                                     || string.Equals("yuvj420p", videoStream.PixelFormat, StringComparison.OrdinalIgnoreCase);
             var is8_10bitSwFormatsVt = is8bitSwFormatsVt || string.Equals("yuv420p10le", videoStream.PixelFormat, StringComparison.OrdinalIgnoreCase);
+            var is8_10_12bitSwFormatsVt = is8_10bitSwFormatsVt
+                || string.Equals("yuv422p", videoStream.PixelFormat, StringComparison.OrdinalIgnoreCase)
+                || string.Equals("yuv444p", videoStream.PixelFormat, StringComparison.OrdinalIgnoreCase)
+                || string.Equals("yuv422p10le", videoStream.PixelFormat, StringComparison.OrdinalIgnoreCase)
+                || string.Equals("yuv444p10le", videoStream.PixelFormat, StringComparison.OrdinalIgnoreCase);
```
gnattu, line 6499: We can add 12bit formats for vt after https://github.com/jellyfin/jellyfin-ffmpeg/pull/459 merged

gnattu:
I'm still a bit concerned about the complexity of such options because they are:

- very niche and most of our users probably do not even now what HEVC `Rext` profile is
- very verbose to have such profile options where there is already a lot of codecs to choose from
- very limited hardware has actual support but lots of users will just blindly check every ticks they can see, this already caused issues that a lot of users just check the av1 encoding blindly without noticing the hardware has to be very new to support that, though this might cause less problems due to rext videos are rare

Maybe we can hide those options for platforms have grace software fallback (like VideoToolbox) and assumes rext is always enabled when the base HEVC profile is enabled for now, and/or make a dedicated section for such advanced profiles so that users without actual need never has to look at it?

nyanmisaka:
> I'm still a bit concerned about the complexity of such options because they are:
> 
> * very niche and most of our users probably do not even now what HEVC `Rext` profile is
> * very verbose to have such profile options where there is already a lot of codecs to choose from
> * very limited hardware has actual support but lots of users will just blindly check every ticks they can see, this already caused issues that a lot of users just check the av1 encoding blindly without noticing the hardware has to be very new to support that, though this might cause less problems due to rext videos are rare
> 
> Maybe we can hide those options for platforms have grace software fallback (like VideoToolbox) and assumes rext is always enabled when the base HEVC profile is enabled for now, and/or make a dedicated section for such advanced profiles so that users without actual need never has to look at it?

This is why I have been hesitant to add new options. It will always be a problem unless we can check hardware caps and hide unsupported codecs to avoid confusing the user.

Another approach is to add some comments for advanced features like AV1 encoder and HEVC RExt, telling users which generation of GPU is required for this feature. This saves us from redesigning a new advanced page for the time being.

For enabling HEVC RExt by default for VideoToolbox that supports software rollback, is this enough?

```cs
EnableDecodingColorDepth10HevcRext = (HardwareAccelerationType == HardwareAccelerationType.videotoolbox);
EnableDecodingColorDepth12HevcRext = (HardwareAccelerationType == HardwareAccelerationType.videotoolbox);
```

gnattu:
Just don't check `EnableDecodingColorDepth10HevcRext` and `EnableDecodingColorDepth12HevcRext` value when videotoolbox is the active hardware acceleration in `GetHwaccelType ` so that it will always go to hevc check.

nyanmisaka:
> Just don't check `EnableDecodingColorDepth10HevcRext` and `EnableDecodingColorDepth12HevcRext` value when videotoolbox is the active hardware acceleration in `GetHwaccelType` so that it will always go to hevc check.

Well, let's make it behave the same as High10P. Then we can hide it for VT in web.

nyanmisaka:
> Just don't check `EnableDecodingColorDepth10HevcRext` and `EnableDecodingColorDepth12HevcRext` value when videotoolbox is the active hardware acceleration in `GetHwaccelType` so that it will always go to hevc check.

So why not do the same for HEVC/VP9 10bit? VT should have fallback for them too, right?

gnattu:
> > Just don't check `EnableDecodingColorDepth10HevcRext` and `EnableDecodingColorDepth12HevcRext` value when videotoolbox is the active hardware acceleration in `GetHwaccelType` so that it will always go to hevc check.
> 
> So why not do the same for HEVC/VP9 10bit? VT should have fallback for them too, right?

It should, and the devices without real hardware support for those is a bit too old and probably too slow to decode those in high resolution anyway. It should be fine to simplify the process a bit.

nyanmisaka:
> > So why not do the same for HEVC/VP9 10bit? VT should have fallback for them too, right?
> 
> It should, and the devices without real hardware support for those is a bit too old and probably too slow to decode those in high resolution anyway. It should be fine to simplify the process a bit.

Done.

gnattu (approved):
This is okay for transcoding only support to me. But to have better RExt support there is a lot of more works to do:

- Check browser support and conditionally add RExt profile (that would be more complex than you think)
- Fix our HLS string builder to properly build the RExt codec string, currently it is semi-hardcoded for Main and Main 10

One challenge for proper RExt direct play and/or remuxing is that all the browsers reports this overly confident. Chrome even does not check level at all and blindly report it can play it, and the video may not be played correctly or good enough for consuming (like significant frame drops). Mobile devices like smartphones and tablets have reporting support for rext may also experience serious frame drops. This probably deserves a dedicated advanced option like Hi10P for Safari.

nyanmisaka:
> This is okay for transcoding only support to me. But to have better RExt support there is a lot of more works to do:
> 
> * Check browser support and conditionally add RExt profile (that would be more complex than you think)
> * Fix our HLS string builder to properly build the RExt codec string, currently it is semi-hardcoded for Main and Main 10

Back in the early days we didn't even think about streaming 422/444 content via HLS because there was no browser support. But in recent years Chromium has added RExt support for Apple and Intel so it's starting to become a thing.

> One challenge for proper RExt direct play and/or remuxing is that all the browsers reports this overly confident. Chrome even does not check level at all and blindly report it can play it, and the video may not be played correctly or good enough for consuming (like significant frame drops). Mobile devices like smartphones and tablets have reporting support for rext may also experience serious frame drops. This probably deserves a dedicated advanced option like Hi10P for Safari.

Did they only add decoding support and not update `HTMLMediaElement: canPlayType()`?

crobibero:
@nyanmisaka is there an accompanying web PR for these changes?

nyanmisaka:
> @nyanmisaka is there an accompanying web PR for these changes?

Sure. https://github.com/jellyfin/jellyfin-web/pull/6081

crobibero (approved):

EraYaN:
> Maybe we can hide those options for platforms have grace software fallback (like VideoToolbox) and assumes rext is always enabled when the base HEVC profile is enabled for now, and/or make a dedicated section for such advanced profiles so that users without actual need never has to look at it?

I've been toying with capability detection in the past, but it needs full detection/verification of video output, since some hardware just generates green video when input in unsupported. But having the options there is valuable cause then you can set them.

I'll see if I can find the (python) code I once wrote for this. Essentially it's a lot harder than it seems sadly, lot of hardware is really crap/weird especially on the weird non-RPi SBCs.

gnattu:
>  but it needs full detection/verification of video output, since some hardware just generates green video when input in unsupported.

This is the problem of capability detection. Some drivers just report support for a specific codec when it actually does not and does not even fail/reject when you feed in an unsupported video stream. We will keep the manual config just in case

EraYaN:
> > but it needs full detection/verification of video output, since some hardware just generates green video when input in unsupported.
> 
> This is the problem of capability detection. Some drivers just report support for a specific codec when it actually does not and does not even fail/reject when you feed in an unsupported video stream. We will keep the manual config just in case

You'd essentially would need to show screenshots to the user in a wizard like form. Cause HDR support is also all over the place. Getting the tests files is also tricky though.

Bond-009 (approved):

1000thSummer:
thanks very much on the quick movement, did not expect it to be so fast.
Q - Does the [webui PR]( https://github.com/jellyfin/jellyfin-web/pull/6081) also need to be merge in order to enable this RExt feature? thanks

nyanmisaka:
> thanks very much on the quick movement, did not expect it to be so fast. Q - Does the [webui PR](https://github.com/jellyfin/jellyfin-web/pull/6081) also need to be merge in order to enable this RExt feature? Or is it by default enabled?
> 
> Thanks.

It is disabled by default. You can change it vim `config/encoding.xml` before web changes are merged.

PR #12670, opened 2024-09-19
Add Dolby Vision tests for Tizen

dmitrylyzo:
**Changes**
- Fix Tizen H264 profiles.
- Add Dolby Vision tests for Tizen.
- Allow Dolby Vision fallback layer on Tizen 3+.

**Issues**
N/A


gnattu:
If you add `DOVIwithHLG`, `DOVIWithHDR10` and `DOVIWithSDR` to the supported range types, will profile 8 videos being directly played back? If so maybe we can just add those to prevent remuxing at all as Samsung TVs does not support Dolby Vision anyway and direct play the fallback layer is nothing to be worried about.

dmitrylyzo:
> If you add `DOVIwithHLG`, `DOVIWithHDR10` and `DOVIWithSDR` to the supported range types, will profile 8 videos being directly played back? If so maybe we can just add those to prevent remuxing at all as Samsung TVs does not support Dolby Vision anyway and direct play the fallback layer is nothing to be worried about.

Yes, it seems to work.
https://github.com/jellyfin/jellyfin-web/pull/6090

crobibero (approved):

gnattu (approved):

gnattu (changes requested):
The old tizen3 stereo needs to updated to either apply max channel limit to direct play profile or we expect direct play. Current profile only applies stereo to transcoding profile.

dmitrylyzo:
> The old tizen3 stereo needs to updated to either apply max channel limit to direct play profile or we expect direct play. Current profile only applies stereo to transcoding profile.

Yeah, I am aware of that. I pushed changes, but it was late. 🛌 

I wanted to limit DirectPlay to Stereo, but I had to remove some tests. The same problem with Safari.

We can adjust Tizen 3 input data (set DirectPlay) to just make tests to pass.

Bond-009 (approved):

gnattu (approved):

PR #12676, opened 2024-09-21
Let HLS Controller decide if subtitle should be burn in

gnattu:
Previously, we predicted whether the subtitle should be burned in with transcode reasons, but that was not accurate because the actual transcoding codec is only determined after the client has requested the stream. This pass through the option to the `DynamicHlsController` to handle the subtitle burn-in during the actual transcoding process. Now the client should be responsible to conditionally load the subtitle when this option is enabled.

**Changes**

**Issues**

### DIFF
--- a/Jellyfin.Api/Controllers/DynamicHlsController.cs
+++ b/Jellyfin.Api/Controllers/DynamicHlsController.cs
@@ -578,7 +578,6 @@ public class DynamicHlsController : BaseJellyfinApiController
     /// <param name="streamOptions">Optional. The streaming options.</param>
     /// <param name="enableAdaptiveBitrateStreaming">Enable adaptive bitrate streaming.</param>
     /// <param name="enableAudioVbrEncoding">Optional. Whether to enable Audio Encoding.</param>
-    /// <param name="alwaysBurnInSubtitleWhenTranscoding">Whether to always burn in subtitles when transcoding.</param>
     /// <response code="200">Audio stream returned.</response>
     /// <returns>A <see cref="FileResult"/> containing the playlist file.</returns>
     [HttpGet("Audio/{itemId}/master.m3u8")]
@@ -636,8 +635,7 @@ public class DynamicHlsController : BaseJellyfinApiController
         [FromQuery] EncodingContext? context,
         [FromQuery] Dictionary<string, string> streamOptions,
         [FromQuery] bool enableAdaptiveBitrateStreaming = true,
-        [FromQuery] bool enableAudioVbrEncoding = true,
-        [FromQuery] bool alwaysBurnInSubtitleWhenTranscoding = false)
+        [FromQuery] bool enableAudioVbrEncoding = true)
     {
         var streamingRequest = new HlsAudioRequestDto
         {
@@ -690,7 +688,7 @@ public class DynamicHlsController : BaseJellyfinApiController
             StreamOptions = streamOptions,
             EnableAdaptiveBitrateStreaming = enableAdaptiveBitrateStreaming,
             EnableAudioVbrEncoding = enableAudioVbrEncoding,
-            AlwaysBurnInSubtitleWhenTranscoding = alwaysBurnInSubtitleWhenTranscoding
+            AlwaysBurnInSubtitleWhenTranscoding = false
         };
 
         return await _dynamicHlsHelper.GetMasterHlsPlaylist(TranscodingJobType, streamingRequest, enableAdaptiveBitrateStreaming).ConfigureAwait(false);
--- a/Jellyfin.Api/Controllers/DynamicHlsController.cs
+++ b/Jellyfin.Api/Controllers/DynamicHlsController.cs
@@ -158,6 +158,7 @@ public class DynamicHlsController : BaseJellyfinApiController
     /// <param name="maxHeight">Optional. The max height.</param>
     /// <param name="enableSubtitlesInManifest">Optional. Whether to enable subtitles in the manifest.</param>
     /// <param name="enableAudioVbrEncoding">Optional. Whether to enable Audio Encoding.</param>
+    /// <param name="alwaysBurnInSubtitleWhenTranscoding">Whether to always burn in subtitles when transcoding.</param>
     /// <response code="200">Hls live stream retrieved.</response>
     /// <returns>A <see cref="FileResult"/> containing the hls file.</returns>
     [HttpGet("Videos/{itemId}/live.m3u8")]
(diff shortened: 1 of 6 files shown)


Jellyfin.Api/Controllers/DynamicHlsController.cs
```diff
@@ -251,7 +253,7 @@ [FromQuery] [RegularExpression(EncodingHelper.ValidationRegex)] string? subtitle
             Height = height,
             VideoBitRate = videoBitRate,
             SubtitleStreamIndex = subtitleStreamIndex,
-            SubtitleMethod = subtitleMethod ?? SubtitleDeliveryMethod.Encode,
+            SubtitleMethod = subtitleMethod ?? SubtitleDeliveryMethod.External,
```
dmitrylyzo, line 256: I think the idea is that `Encode` will always work.
But `SubtitleMethod` must be specified for stable results, though.

Shadowghost (approved):

Jellyfin.Api/Controllers/DynamicHlsController.cs
```diff
@@ -629,7 +636,8 @@ [FromQuery] [RegularExpression(EncodingHelper.ValidationRegex)] string? subtitle
         [FromQuery] EncodingContext? context,
         [FromQuery] Dictionary<string, string> streamOptions,
         [FromQuery] bool enableAdaptiveBitrateStreaming = true,
-        [FromQuery] bool enableAudioVbrEncoding = true)
+        [FromQuery] bool enableAudioVbrEncoding = true,
+        [FromQuery] bool alwaysBurnInSubtitleWhenTranscoding = false)
```
dmitrylyzo, line 640: Do we need `alwaysBurnInSubtitleWhenTranscoding` in the audio enpoints?

gnattu: Probably not, but there are also other video options in audio endpoints

crobibero (approved):

Jellyfin.Api/Controllers/DynamicHlsController.cs
gnattu: The problem is that StreamInfo's URL builder always specifies Encode and will leave it blank for External. We should make the behavior consistent.

Jellyfin.Api/Controllers/DynamicHlsController.cs
```diff
@@ -911,6 +921,7 @@ [FromQuery] [RegularExpression(EncodingHelper.ValidationRegex)] string? subtitle
     /// <param name="context">Optional. The <see cref="EncodingContext"/>.</param>
     /// <param name="streamOptions">Optional. The streaming options.</param>
     /// <param name="enableAudioVbrEncoding">Optional. Whether to enable Audio Encoding.</param>
+    /// <param name="alwaysBurnInSubtitleWhenTranscoding">Whether to always burn in subtitles when transcoding.</param>
```
dmitrylyzo, line 924: Then remove this one as well.
_I expected you to remove them all._ :sweat_smile: 
```suggestion
```

Jellyfin.Api/Controllers/DynamicHlsController.cs
```diff
@@ -966,7 +977,8 @@ [FromQuery] [RegularExpression(EncodingHelper.ValidationRegex)] string? subtitle
         [FromQuery] int? videoStreamIndex,
         [FromQuery] EncodingContext? context,
         [FromQuery] Dictionary<string, string> streamOptions,
-        [FromQuery] bool enableAudioVbrEncoding = true)
+        [FromQuery] bool enableAudioVbrEncoding = true,
+        [FromQuery] bool alwaysBurnInSubtitleWhenTranscoding = false)
```
dmitrylyzo, line 981:
```suggestion
        [FromQuery] bool enableAudioVbrEncoding = true)
```

Jellyfin.Api/Controllers/DynamicHlsController.cs
```diff
@@ -1018,7 +1030,8 @@ [FromQuery] [RegularExpression(EncodingHelper.ValidationRegex)] string? subtitle
             VideoStreamIndex = videoStreamIndex,
             Context = context ?? EncodingContext.Streaming,
             StreamOptions = streamOptions,
-            EnableAudioVbrEncoding = enableAudioVbrEncoding
+            EnableAudioVbrEncoding = enableAudioVbrEncoding,
+            AlwaysBurnInSubtitleWhenTranscoding = alwaysBurnInSubtitleWhenTranscoding
```
dmitrylyzo, line 1034:
```suggestion
            AlwaysBurnInSubtitleWhenTranscoding = false
```

Jellyfin.Api/Controllers/DynamicHlsController.cs
```diff
@@ -1267,6 +1283,7 @@ [FromQuery] [RegularExpression(EncodingHelper.ValidationRegex)] string? subtitle
     /// <param name="context">Optional. The <see cref="EncodingContext"/>.</param>
     /// <param name="streamOptions">Optional. The streaming options.</param>
     /// <param name="enableAudioVbrEncoding">Optional. Whether to enable Audio Encoding.</param>
+    /// <param name="alwaysBurnInSubtitleWhenTranscoding">Whether to always burn in subtitles when transcoding.</param>
```
dmitrylyzo, line 1286:
```suggestion
```

Jellyfin.Api/Controllers/DynamicHlsController.cs
```diff
@@ -1328,7 +1345,8 @@ [FromQuery] [RegularExpression(EncodingHelper.ValidationRegex)] string? subtitle
         [FromQuery] int? videoStreamIndex,
         [FromQuery] EncodingContext? context,
         [FromQuery] Dictionary<string, string> streamOptions,
-        [FromQuery] bool enableAudioVbrEncoding = true)
+        [FromQuery] bool enableAudioVbrEncoding = true,
+        [FromQuery] bool alwaysBurnInSubtitleWhenTranscoding = false)
```
dmitrylyzo, line 1349:
```suggestion
        [FromQuery] bool enableAudioVbrEncoding = true)
```

Jellyfin.Api/Controllers/DynamicHlsController.cs
```diff
@@ -1382,7 +1400,8 @@ [FromQuery] [RegularExpression(EncodingHelper.ValidationRegex)] string? subtitle
             VideoStreamIndex = videoStreamIndex,
             Context = context ?? EncodingContext.Streaming,
             StreamOptions = streamOptions,
-            EnableAudioVbrEncoding = enableAudioVbrEncoding
+            EnableAudioVbrEncoding = enableAudioVbrEncoding,
+            AlwaysBurnInSubtitleWhenTranscoding = alwaysBurnInSubtitleWhenTranscoding
```
dmitrylyzo, line 1404:
```suggestion
            AlwaysBurnInSubtitleWhenTranscoding = false
```

dmitrylyzo (approved):

rehan678p (approved):

PR #12679, opened 2024-09-22
Fix DeviceProfile.Id should be nullable

nielsvanvelzen:
Previously (10.9<=) the id of a device profile was a nullable string. This was changed to a GUID during the 10.10 development but with this change the nullability was removed. A lot of clients (Android TV included) do not set a id so this is was a breaking change. The only 2 usages of the id still accept null so I've made the id nullable to keep compatibility.

**Changes**
- Fix DeviceProfile.Id should be nullable

**Issues**

### DIFF
--- a/MediaBrowser.Model/Dlna/DeviceProfile.cs
+++ b/MediaBrowser.Model/Dlna/DeviceProfile.cs
@@ -22,7 +22,7 @@ public class DeviceProfile
     /// <summary>
     /// Gets or sets the unique internal identifier.
     /// </summary>
-    public Guid Id { get; set; }
+    public Guid? Id { get; set; }
 
     /// <summary>
     /// Gets or sets the maximum allowed bitrate for all streamed content.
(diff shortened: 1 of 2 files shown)


nielsvanvelzen:
I'm not entirely sure what the id is used for but from my understanding it doesn't have to be a Guid, additionally the DeviceProfileId in stream info is still a string so shouldn't we change that as well?

gnattu:
From the original PR:

> Changed Profile.Id to Guid - in prep for DLNA plugin.

But that DLNA change never took place unless I missed something, so it should be fine for us to make it nullable

Bond-009 (approved):

crobibero (approved):

Shadowghost (approved):

rehan678p:
Scammer attack all wallet withdrawal money

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

PR #12678, opened 2024-09-22
Extract container, video and audio compatibility checks

dmitrylyzo:
This is kind of a start point for https://github.com/jellyfin/jellyfin/pull/10756

Later, we can properly [filter the direct audio stream](https://github.com/jellyfin/jellyfin/blob/38f80edc8011783d6544a9c1bd6401e3fe13ff57/MediaBrowser.Model/Dlna/StreamBuilder.cs#L967-L977):
```cs
var directAudioStream = candidateAudioStreams.FirstOrDefault(stream => ContainerHelper.ContainsContainer(audioCodecs, false, stream.Codec)
    && (transcodingProfile is null || GetCompatibilityAudioCodec(playlistItem.MediaSource!, container!, stream, options.Profile, true, false) == 0));
```

**Changes**
Extract container, video and audio compatibility checks.

**Issues**
N/A

PR #12689, opened 2024-09-23
Check all conditions for direct audio stream

dmitrylyzo:
_Depends on https://github.com/jellyfin/jellyfin/pull/12678_

**Changes**
Check all conditions for direct audio stream.

**Issues**
N/A

### DIFF
--- a/tests/Jellyfin.Model.Tests/Test Data/DeviceProfile-SafariNext.json
+++ b/tests/Jellyfin.Model.Tests/Test Data/DeviceProfile-SafariNext.json
@@ -191,6 +191,28 @@
     "ContainerProfiles": [],
     "CodecProfiles": [
         {
+            "Type": "Audio",
+            "Conditions": [
+                {
+                    "Condition": "LessThanEqual",
+                    "Property": "AudioChannels",
+                    "Value": "2",
+                    "IsRequired": false
+                }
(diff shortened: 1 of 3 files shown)


tests/Jellyfin.Model.Tests/Test Data/DeviceProfile-SafariNext.json
```diff
@@ -190,6 +190,28 @@
...
+            "Type": "VideoAudio",
+            "Conditions": [
+                {
+                    "Condition": "LessThanEqual",
+                    "Property": "AudioChannels",
+                    "Value": "2",
+                    "IsRequired": false
+                }
+            ]
+        },
```
gnattu, line 214: Is this how web going to supply global channel limit in the future or is it already the case?

dmitrylyzo: It is currently only set when `Settings -> Playback -> Maximum Audio Channels` is explicitly selected (not `Auto`).

Tbh, I think we should probably move this to `options.MaxAudioChannels` (since we don't have a special field in DeviceProfile) because it is not a codec limits.

Bond-009:
What's the status of this?

PR #12696, opened 2024-09-23
Fix get sessions with api key

gnattu:
The API key will not be associated with any user and has to be handled separately.

**Changes**

**Issues**

### DIFF
--- a/Emby.Server.Implementations/Session/SessionManager.cs
+++ b/Emby.Server.Implementations/Session/SessionManager.cs
@@ -1858,13 +1858,36 @@ namespace Emby.Server.Implementations.Session
             Guid userId,
             string deviceId,
             int? activeWithinSeconds,
-            Guid? controllableUserToCheck)
+            Guid? controllableUserToCheck,
+            bool isApiKey)
         {
             var result = Sessions;
-            var user = _userManager.GetUserById(userId);
             if (!string.IsNullOrEmpty(deviceId))
             {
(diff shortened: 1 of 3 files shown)


gnattu:
> ABI Difference
> ```
> 
> MediaBrowser.Controller.dll
> API compatibility errors between './abi-base/MediaBrowser.Controller.dll' (left) and './abi-head/MediaBrowser.Controller.dll' (right):
> CP0002: Member 'System.Collections.Generic.IReadOnlyList<MediaBrowser.Model.Dto.SessionInfoDto> MediaBrowser.Controller.Session.ISessionManager.GetSessions(System.Guid, string, System.Nullable<int>, System.Nullable<System.Guid>)' exists on ./abi-base/MediaBrowser.Controller.dll but not on ./abi-head/MediaBrowser.Controller.dll
> CP0006: Cannot add interface member 'System.Collections.Generic.IReadOnlyList<MediaBrowser.Model.Dto.SessionInfoDto> MediaBrowser.Controller.Session.ISessionManager.GetSessions(System.Guid, string, System.Nullable<int>, System.Nullable<System.Guid>, bool)' to ./abi-head/MediaBrowser.Controller.dll because it does not exist on ./abi-base/MediaBrowser.Controller.dll
> API breaking changes found. If those are intentional, the APICompat suppression file can be updated by specifying the '--generate-suppression-file' parameter.
> ```

This should be fine as those methods are for controllers and not intended for plugin use

crobibero (approved):

PR #12697, opened 2024-09-23
Add endpoint for getting playlists by id

Shadowghost:
This is needed to properly implement editing playlists.

**Changes**
* Add endpoint to get the full playlist object by id

### DIFF
--- a/Jellyfin.Api/Controllers/PlaylistsController.cs
+++ b/Jellyfin.Api/Controllers/PlaylistsController.cs
@@ -163,7 +163,7 @@ public class PlaylistsController : BaseJellyfinApiController
     [ProducesResponseType(StatusCodes.Status200OK)]
     [ProducesResponseType(StatusCodes.Status403Forbidden)]
     [ProducesResponseType(StatusCodes.Status404NotFound)]
-    public ActionResult<Playlist> GetPlaylist(
+    public ActionResult<PlaylistDto> GetPlaylist(
         [FromRoute, Required] Guid playlistId)
     {
         var userId = User.GetUserId();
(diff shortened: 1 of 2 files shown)


MediaBrowser.Model/Dto/PlaylistDto.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    /// <summary>
+    /// Gets or sets a value indicating whether the playlist is publicly readable.
+    /// </summary>
+    public bool OpenAccess { get; set; }
+
+    /// <summary>
+    /// Gets or sets the share permissions.
+    /// </summary>
+    public IReadOnlyList<PlaylistUserPermissions>? Shares { get; set; }
```
nielsvanvelzen, line 20: What is the difference between null and empty list here? (Same question for the ItemIds)

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -149,6 +149,39 @@ await _playlistManager.UpdatePlaylist(new PlaylistUpdateRequest
...
+    /// <param name="playlistId">The playlist id.</param>
+    /// <response code="200">The playlist.</response>
+    /// <response code="403">Access forbidden.</response>
+    /// <response code="404">Playlist not found.</response>
+    /// <returns>
+    /// A <see cref="Playlist"/> objects.
+    /// </returns>
+    [HttpGet("{playlistId}")]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status403Forbidden)]
```
crobibero, line 164: remove, we don't return 403 here

MediaBrowser.Model/Dto/PlaylistDto.cs
crobibero: I think the properties should just be marked as `public required` and nullability removed

nielsvanvelzen (approved):

crobibero (approved):

Bond-009 (approved):

PR #12698, opened 2024-09-23
Sanitize CustomTagDelimiters server side

gnattu:
The API requires an array type and does not support runtime generated default value. Use server side helper function to sanitize it into char.

**Changes**

**Issues**

### DIFF
--- a/MediaBrowser.Model/Configuration/LibraryOptions.cs
+++ b/MediaBrowser.Model/Configuration/LibraryOptions.cs
@@ -149,19 +149,5 @@ namespace MediaBrowser.Model.Configuration
 
             return null;
         }
-
-        public char[] GetCustomTagDelimiters()
-        {
-            return CustomTagDelimiters.Select<string, char?>(x =>
-            {
-                var isChar = char.TryParse(x, out var c);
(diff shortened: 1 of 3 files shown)


Shadowghost (approved):

MediaBrowser.Model/Configuration/LibraryOptions.cs
```diff
@@ -149,5 +149,19 @@ public LibraryOptions()
 
             return null;
         }
+
+        public char[] GetCustomTagDelimiters()
```
JPVenson, line 153: We should stop adding buissniss logic in our DTOs.
Use a extension method instead

JPVenson (changes requested):

MediaBrowser.Model/Configuration/LibraryOptions.cs
gnattu: The Emby time getter is kept in its old place to not break the code too much. This new one has being moved.

JPVenson (approved):

nielsvanvelzen (approved):

PR #12705, opened 2024-09-24
Workaround ATL tag parsing

gnattu:
Current version of ATL will separate some ID3v2.3/2.2 tag fields with its internally defined separator `/`, but this is not desired as we are providing a more customizable separating behavior.  Because the spec of ID3v2.3/2.2 does not have official separator, the separated string will have no ambiguity, which means we can replace the `InternalValueSeparator` back to `/` as a workaround. Ideally we can wait for ATL to add an option to disable such behavior: https://github.com/Zeugma440/atldotnet/issues/281

This also disabled the ATL automatic filename fallback with options instead of checking metadata type, which is a less hacky way to prefer the ffprobe result.

**Changes**

**Issues**

### DIFF
--- a/MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
+++ b/MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
@@ -65,6 +65,7 @@ namespace MediaBrowser.Providers.MediaInfo
             _lyricResolver = lyricResolver;
             _lyricManager = lyricManager;
             ATL.Settings.DisplayValueSeparator = InternalValueSeparator;
+            ATL.Settings.UseFileNameWhenNoTitle = false;
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
```diff
@@ -437,5 +440,15 @@ private List<string> SplitWithCustomDelimiter(string val, char[] tagDelimiters,
 
             return items;
         }
+
+        private static string[] SplitWithWorkaround(string val, char separator, bool useWorkaround)
+        {
+            if (useWorkaround)
+            {
+                val = val.Replace(InternalValueSeparator.ToString(), AtlId3V22Separator.ToString(), StringComparison.OrdinalIgnoreCase);
```
dmitrylyzo, line 448: Why `ToString`?
Shouldn't we use exact match?
```suggestion
                val = val.Replace(InternalValueSeparator, AtlId3V22Separator);
```

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
```diff
@@ -161,12 +164,12 @@ private async Task FetchDataFromTags(Audio audio, Model.MediaInfo.MediaInfo medi
             var libraryOptions = _libraryManager.GetLibraryOptions(audio);
             Track track = new Track(audio.Path);
 
-            // ATL will fall back to filename as title when it does not understand the metadata
-            if (track.MetadataFormats.All(mf => mf.Equals(ATL.Factory.UNKNOWN_FORMAT)))
-            {
-                track.Title = mediaInfo.Name;
-            }
+            // Current version of ATL split some ID3v2.2/2.3 tags by '/' which is not desired.
+            var shouldWorkaroundAtlId3V22Split = track.MetadataFormats
```
dmitrylyzo, line 168: We can also check `libraryOptions.UseCustomTagDelimiters` if we want to preserve the old behavior without custom tag delimiters.

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
gnattu: because val is string and replace does not take char as input

crobibero: We don't need to set string comparison since we know what both characters are and neither have case issues, so we can do the char replacement as suggested.

crobibero (approved):

Shadowghost (approved):

dmitrylyzo (approved):

PR #12718, opened 2024-09-25
Extract subtitles of MKS aux files if they're extractable.

IceStormNG:
The current implementation only extracts from the main container. This makes SRT/ASS subtitles from MKS files unusable (except when burning them in), because the server expects to retrieve them from the subtitles directory.

This change also extracts subtitles from MKS files that are associated with the video file and stores them in their appropriate location, so the client can request and render them.

**Changes**
Adds additional subtitle extractor logic for MKS file(s).

**Issues**
https://github.com/jellyfin/jellyfin/issues/12236

### DIFF
--- a/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
+++ b/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
@@ -612,7 +612,13 @@ namespace MediaBrowser.MediaEncoding.Subtitles
                 var outputCodec = IsCodecCopyable(subtitleStream.Codec) ? "copy" : "srt";
                 var streamIndex = EncodingHelper.FindIndex(mediaSource.MediaStreams, subtitleStream);
 
-                if (streamIndex == -1 || subtitleStream.Path.EndsWith(".mks", StringComparison.OrdinalIgnoreCase))
+                if (subtitleStream.Path.EndsWith(".mks", StringComparison.OrdinalIgnoreCase))
+                {
+                    _logger.LogDebug("Subtitle {Index} for file {InputPath} is part in an MKS file. Skipping", inputPath, subtitleStream.Index);
+                    continue;
+                }
+
+                if (streamIndex == -1)
                 {
                     _logger.LogError("Cannot find subtitle stream index for {InputPath} ({Index}), skipping this stream", inputPath, subtitleStream.Index);
                     continue;
--- a/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
+++ b/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
@@ -489,10 +489,15 @@ namespace MediaBrowser.MediaEncoding.Subtitles
             try
             {
                 var subtitleStreams = mediaSource.MediaStreams
-                    .Where(stream => stream is { IsExtractableSubtitleStream: true, SupportsExternalStream: true, IsExternal: false });
+                    .Where(stream => stream is { IsExtractableSubtitleStream: true, SupportsExternalStream: true });
 
                 foreach (var subtitleStream in subtitleStreams)
                 {
+                    if (subtitleStream.IsExternal && !subtitleStream.Path.EndsWith(".mks", StringComparison.OrdinalIgnoreCase))
+                    {
+                        continue;
+                    }
+
                     var outputPath = GetSubtitleCachePath(mediaSource, subtitleStream.Index, "." + GetExtractableSubtitleFileExtension(subtitleStream));
 
                     var releaser = await _semaphoreLocks.LockAsync(outputPath, cancellationToken).ConfigureAwait(false);
@@ -510,6 +515,7 @@ namespace MediaBrowser.MediaEncoding.Subtitles
                 if (extractableStreams.Count > 0)
                 {
                     await ExtractAllExtractableSubtitlesInternal(mediaSource, extractableStreams, cancellationToken).ConfigureAwait(false);
+                    await ExtractAllExtractableSubtitlesMKS(mediaSource, extractableStreams, cancellationToken).ConfigureAwait(false);
                 }
             }
             catch (Exception ex)
@@ -519,6 +525,72 @@ namespace MediaBrowser.MediaEncoding.Subtitles
             finally
             {
                 locks.ForEach(x => x.Dispose());
+            }
+        }
+
+        private async Task ExtractAllExtractableSubtitlesMKS(
+           MediaSourceInfo mediaSource,
+           List<MediaStream> subtitleStreams,
+           CancellationToken cancellationToken)
+        {
+            var mksFiles = new List<string>();
+
+            foreach (var subtitleStream in subtitleStreams)
+            {
+                if (!subtitleStream.Path.EndsWith(".mks", StringComparison.OrdinalIgnoreCase))
+                {
+                    continue;
+                }
+
+                if (!mksFiles.Contains(subtitleStream.Path))
+                {
+                    mksFiles.Add(subtitleStream.Path);
+                }
+            }
+
+            if (mksFiles.Count == 0)
+            {
+                return;
+            }
+
+            foreach (string mksFile in mksFiles)
+            {
+                var inputPath = _mediaEncoder.GetInputArgument(mksFile, mediaSource);
+                var outputPaths = new List<string>();
+                var args = string.Format(
+                CultureInfo.InvariantCulture,
+                "-i {0} -copyts",
+                inputPath);
+
+                foreach (var subtitleStream in subtitleStreams)
+                {
+                    if (!subtitleStream.Path.Equals(mksFile, StringComparison.OrdinalIgnoreCase))
+                    {
+                        continue;
+                    }
+
+                    var outputPath = GetSubtitleCachePath(mediaSource, subtitleStream.Index, "." + GetExtractableSubtitleFileExtension(subtitleStream));
+                    var outputCodec = IsCodecCopyable(subtitleStream.Codec) ? "copy" : "srt";
+                    var streamIndex = EncodingHelper.FindIndex(mediaSource.MediaStreams, subtitleStream);
+
+                    if (streamIndex == -1)
+                    {
+                        _logger.LogError("Cannot find subtitle stream index for {InputPath} ({Index}), skipping this stream", inputPath, subtitleStream.Index);
+                        continue;
+                    }
+
+                    Directory.CreateDirectory(Path.GetDirectoryName(outputPath) ?? throw new FileNotFoundException($"Calculated path ({outputPath}) is not valid."));
+
+                    outputPaths.Add(outputPath);
+                    args += string.Format(
+                        CultureInfo.InvariantCulture,
+                        " -map 0:{0} -an -vn -c:s {1} \"{2}\"",
+                        streamIndex,
+                        outputCodec,
+                        outputPath);
+                }
+
+                await ExtractSubtitlesForFile(inputPath, args, outputPaths, cancellationToken).ConfigureAwait(false);
             }
         }
 
@@ -540,7 +612,7 @@ namespace MediaBrowser.MediaEncoding.Subtitles
                 var outputCodec = IsCodecCopyable(subtitleStream.Codec) ? "copy" : "srt";
                 var streamIndex = EncodingHelper.FindIndex(mediaSource.MediaStreams, subtitleStream);
 
-                if (streamIndex == -1)
+                if (streamIndex == -1 || subtitleStream.Path.EndsWith(".mks", StringComparison.OrdinalIgnoreCase))
                 {
                     _logger.LogError("Cannot find subtitle stream index for {InputPath} ({Index}), skipping this stream", inputPath, subtitleStream.Index);
                     continue;
@@ -557,6 +629,15 @@ namespace MediaBrowser.MediaEncoding.Subtitles
                     outputPath);
             }
 
+            await ExtractSubtitlesForFile(inputPath, args, outputPaths, cancellationToken).ConfigureAwait(false);
+        }
+
+        private async Task ExtractSubtitlesForFile(
+            string inputPath,
+            string args,
+            List<string> outputPaths,
+            CancellationToken cancellationToken)
+        {
             int exitCode;
 
             using (var process = new Process
--- a/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
+++ b/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
@@ -635,6 +635,11 @@ namespace MediaBrowser.MediaEncoding.Subtitles
                     outputPath);
             }
 
+            if (outputPaths.Count == 0)
+            {
+                return;
+            }
+
             await ExtractSubtitlesForFile(inputPath, args, outputPaths, cancellationToken).ConfigureAwait(false);
         }
 
--- a/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
+++ b/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
@@ -608,15 +608,15 @@ namespace MediaBrowser.MediaEncoding.Subtitles
 
             foreach (var subtitleStream in subtitleStreams)
             {
-                var outputPath = GetSubtitleCachePath(mediaSource, subtitleStream.Index, "." + GetExtractableSubtitleFileExtension(subtitleStream));
-                var outputCodec = IsCodecCopyable(subtitleStream.Codec) ? "copy" : "srt";
-                var streamIndex = EncodingHelper.FindIndex(mediaSource.MediaStreams, subtitleStream);
-
                 if (subtitleStream.Path.EndsWith(".mks", StringComparison.OrdinalIgnoreCase))
                 {
                     _logger.LogDebug("Subtitle {Index} for file {InputPath} is part in an MKS file. Skipping", inputPath, subtitleStream.Index);
                     continue;
                 }
+
+                var outputPath = GetSubtitleCachePath(mediaSource, subtitleStream.Index, "." + GetExtractableSubtitleFileExtension(subtitleStream));
+                var outputCodec = IsCodecCopyable(subtitleStream.Codec) ? "copy" : "srt";
+                var streamIndex = EncodingHelper.FindIndex(mediaSource.MediaStreams, subtitleStream);
 
                 if (streamIndex == -1)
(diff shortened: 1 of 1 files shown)


MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
```diff
@@ -524,6 +530,72 @@ private async Task ExtractAllExtractableSubtitles(MediaSourceInfo mediaSource, C
...
+            }
+
+            foreach (string mksFile in mksFiles)
+            {
+                var inputPath = _mediaEncoder.GetInputArgument(mksFile, mediaSource);
+                var outputPaths = new List<string>();
+                var args = string.Format(
+                CultureInfo.InvariantCulture,
+                "-i {0} -copyts",
+                inputPath);
```
Shadowghost, line 565: Formatting is off

MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
```diff
@@ -524,6 +530,72 @@ private async Task ExtractAllExtractableSubtitles(MediaSourceInfo mediaSource, C
...
+                    var outputCodec = IsCodecCopyable(subtitleStream.Codec) ? "copy" : "srt";
+                    var streamIndex = EncodingHelper.FindIndex(mediaSource.MediaStreams, subtitleStream);
+
+                    if (streamIndex == -1)
+                    {
+                        _logger.LogError("Cannot find subtitle stream index for {InputPath} ({Index}), skipping this stream", inputPath, subtitleStream.Index);
+                        continue;
+                    }
+
+                    Directory.CreateDirectory(Path.GetDirectoryName(outputPath) ?? throw new FileNotFoundException($"Calculated path ({outputPath}) is not valid."));
```
Shadowghost, line 584: I don't think this type of exception is the correct one if we were unable to create the directory

IceStormNG: The already existing function for extracting the subtitles from the main container throws the same exception. Therefore, I did the same for the MKS extractor.

MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
IceStormNG: Should be fixed now.

MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
```diff
@@ -542,7 +614,7 @@ private async Task ExtractAllExtractableSubtitlesInternal(
                 var outputCodec = IsCodecCopyable(subtitleStream.Codec) ? "copy" : "srt";
                 var streamIndex = EncodingHelper.FindIndex(mediaSource.MediaStreams, subtitleStream);
 
-                if (streamIndex == -1)
+                if (streamIndex == -1 || subtitleStream.Path.EndsWith(".mks", StringComparison.OrdinalIgnoreCase))
```
dmitrylyzo, line 617: I'd split this `if` and use a different message for `.mks`.

IceStormNG: You're right. I separated those conditions and changed added a different debug message for the case that the sub is part of an MKS file.

MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
```diff
@@ -542,6 +614,12 @@ private async Task ExtractAllExtractableSubtitlesInternal(
                 var outputCodec = IsCodecCopyable(subtitleStream.Codec) ? "copy" : "srt";
                 var streamIndex = EncodingHelper.FindIndex(mediaSource.MediaStreams, subtitleStream);
 
+                if (subtitleStream.Path.EndsWith(".mks", StringComparison.OrdinalIgnoreCase))
+                {
+                    _logger.LogDebug("Subtitle {Index} for file {InputPath} is part in an MKS file. Skipping", inputPath, subtitleStream.Index);
+                    continue;
+                }
+
```
dmitrylyzo, line 622: Move it to the beginning of the loop to skip redundant `outputPath`, `outputCodec` and `streamIndex`.

MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
```diff
@@ -559,6 +637,15 @@ private async Task ExtractAllExtractableSubtitlesInternal(
                     outputPath);
             }
 
+            await ExtractSubtitlesForFile(inputPath, args, outputPaths, cancellationToken).ConfigureAwait(false);
```
dmitrylyzo, line 640: I think we should exit if `outputPaths.Count == 0` - MKSs only subs.

IceStormNG: Agreed. I added an early abort there.

MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
IceStormNG: Done.

gnattu:
I'm a bit concerned about this file format TBH because our external subtitle handling is not really designed for such file and still has hard to reproducible bugs, and I do think this file format that allows a file with multiple subtitle tracks could lead to some subtitle refreshing issues because the files are mutable and our scanner is not entirely prepared for this, especially the language coding part is pure filename based now iirc.

Shadowghost:
We're using the file metadata for mks, otherwise we couldn't differentiate the tracks in any way except id.

IceStormNG:
The language coding works fine with MKS either by filename (which would only makes sense if all streams inside the MKS are of the same language) or by meta tags of the Matroska Container.

Problem is: Right now, text-based subtitles inside MKS files are selectable and detected, but they're unplayable as soft sub, because they're not extracted and Jellyfin cannot stream subs directly from the MKS, only extracted or when burning them in.

As for the refresh. Can't we use the "last modified" date from the MKS and the extracted sub files to check whether the MKS was changed after we extracted the files last time and then extract the files again?

If MKS shouldn't be extracted due to the issues you noted, then maybe text-based subs shouldn't be supported at all from MKS, or at least not as soft-subs because they simply don't work right now.

Shadowghost (approved):

Gladtbam:
```
[2024-12-21 09:14:10.290 +00:00] [INF] User policy for "My Name". EnablePlaybackRemuxing: True EnableVideoPlaybackTranscoding: True EnableAudioPlaybackTranscoding: True
[2024-12-21 09:14:11.451 +00:00] [INF] Adding playback tracker : a80c6a531bf2618984f15d6ba4324cf6bbcb3a66e20ab811-84f15d6ba4324cf6bbcb3a66e20ab811-698c380618a6056fb1f9c3bccf653850
[2024-12-21 09:14:11.453 +00:00] [INF] PlaybackTracker : Adding Start Event : 12/21/2024 09:14:11
[2024-12-21 09:14:11.453 +00:00] [INF] Creating StartPlaybackTimer Task
[2024-12-21 09:14:11.459 +00:00] [INF] StartPlaybackTimer : Entered
[2024-12-21 09:14:11.685 +00:00] [INF] Processing playback tracker : "a80c6a531bf2618984f15d6ba4324cf6bbcb3a66e20ab811-84f15d6ba4324cf6bbcb3a66e20ab811-698c380618a6056fb1f9c3bccf653850"
[2024-12-21 09:14:11.907 +00:00] [INF] Started user data sync
[2024-12-21 09:14:11.931 +00:00] [ERR] An error has occurred in UserUpdateTimerCallback
System.InvalidCastException: Unable to cast object of type 'System.Guid' to type 'System.String'.
   at lambda_method1341(Closure, Object, Object)
   at LiteDB.BsonMapper.DeserializeObject(EntityMapper entity, Object obj, BsonDocument value)
   at LiteDB.BsonMapper.Deserialize(Type type, BsonValue value)
   at LiteDB.LiteQueryable`1.<ToEnumerable>b__27_2(BsonDocument x)
   at System.Linq.Enumerable.SelectEnumerableIterator`2.ToList()
   at Jellyfin.Plugin.KodiSyncQueue.Data.DbRepo.SetUserInfoSync(List`1 dtos, List`1 itemRefs, String userId)
   at Jellyfin.Plugin.KodiSyncQueue.EntryPoints.UserSyncNotification.SaveUserChanges(List`1 dtos, List`1 itemRefs, String userName, String userId)
   at Jellyfin.Plugin.KodiSyncQueue.EntryPoints.UserSyncNotification.SendNotifications(IEnumerable`1 changes, List`1 itemRefs, CancellationToken cancellationToken)
   at Jellyfin.Plugin.KodiSyncQueue.EntryPoints.UserSyncNotification.UpdateTimerCallback(Object state)
[2024-12-21 09:14:16.525 +00:00] [INF] "/usr/lib/jellyfin-ffmpeg/ffmpeg" "-i file:\"/mnt/remote/Anime/VCB/AG/BanG Dream! It’s MyGO!!!!! (2023)/Season 01/S01E12 - [Nekomoe kissaten&VCB-Studio] BanG Dream! It’s MyGO!!!!! [12][Ma10p_1080p][x265_flac].zh.mks\" -copyts -map 0:0 -an -vn -c:s copy \"/var/lib/jellyfin/data/subtitles/4/41ca9444-691c-9d17-05d3-587fc4c6c5cd.ass\" -map 0:1 -an -vn -c:s copy \"/var/lib/jellyfin/data/subtitles/7/7a776776-73b8-dba8-0a03-83e1e1cd11f4.ass\""
[2024-12-21 09:14:17.316 +00:00] [INF] Setting ass font within "/var/lib/jellyfin/data/subtitles/4/41ca9444-691c-9d17-05d3-587fc4c6c5cd.ass"
[2024-12-21 09:14:17.356 +00:00] [INF] ffmpeg subtitle extraction completed for "file:\"/mnt/remote/Anime/VCB/AG/BanG Dream! It’s MyGO!!!!! (2023)/Season 01/S01E12 - [Nekomoe kissaten&VCB-Studio] BanG Dream! It’s MyGO!!!!! [12][Ma10p_1080p][x265_flac].mks\"" to "/var/lib/jellyfin/data/subtitles/4/41ca9444-691c-9d17-05d3-587fc4c6c5cd.ass"
[2024-12-21 09:14:17.356 +00:00] [INF] Setting ass font within "/var/lib/jellyfin/data/subtitles/7/7a776776-73b8-dba8-0a03-83e1e1cd11f4.ass"
[2024-12-21 09:14:17.405 +00:00] [INF] ffmpeg subtitle extraction completed for "file:\"/mnt/remote/Anime/VCB/AG/BanG Dream! It’s MyGO!!!!! (2023)/Season 01/S01E12 - [Nekomoe kissaten&VCB-Studio] BanG Dream! It’s MyGO!!!!! [12][Ma10p_1080p][x265_flac].mks\"" to "/var/lib/jellyfin/data/subtitles/7/7a776776-73b8-dba8-0a03-83e1e1cd11f4.ass"
[2024-12-21 09:14:31.460 +00:00] [INF] session.RemoteEndPoint : ""
[2024-12-21 09:14:31.462 +00:00] [INF] StartPlaybackTimer : event_playing_id     = ""
[2024-12-21 09:14:31.462 +00:00] [INF] StartPlaybackTimer : event_user_id        = ""
[2024-12-21 09:14:31.462 +00:00] [INF] StartPlaybackTimer : event_user_id_int    = 1
[2024-12-21 09:14:31.462 +00:00] [INF] StartPlaybackTimer : session_playing_id   = ""
[2024-12-21 09:14:31.462 +00:00] [INF] StartPlaybackTimer : session_user_id      = ""
[2024-12-21 09:14:31.462 +00:00] [INF] StartPlaybackTimer : play_method          = "DirectPlay"
[2024-12-21 09:14:31.462 +00:00] [INF] StartPlaybackTimer : e.ClientName         = "Jellyfin Android"
[2024-12-21 09:14:31.462 +00:00] [INF] StartPlaybackTimer : e.DeviceName         = "S21"
[2024-12-21 09:14:31.462 +00:00] [INF] StartPlaybackTimer : ItemName             = "BanG Dream! It's MyGO!!!!! - s01e12 - It’s my go!!!!!"
[2024-12-21 09:14:31.462 +00:00] [INF] StartPlaybackTimer : ItemId               = "698c380618a6056fb1f9c3bccf653850"
[2024-12-21 09:14:31.462 +00:00] [INF] StartPlaybackTimer : ItemType             = "Episode"
[2024-12-21 09:14:31.462 +00:00] [INF] StartPlaybackTimer : All matches, playback registered
[2024-12-21 09:14:31.462 +00:00] [INF] Playback tracker found, adding playback info : "a80c6a531bf2618984f15d6ba4324cf6bbcb3a66e20ab811-84f15d6ba4324cf6bbcb3a66e20ab811-698c380618a6056fb1f9c3bccf653850"
[2024-12-21 09:14:31.462 +00:00] [INF] Saving playback tracking activity in DB
[2024-12-21 09:14:31.581 +00:00] [INF] StartPlaybackTimer : Exited
[2024-12-21 09:14:32.409 +00:00] [INF] Processing playback tracker : "a80c6a531bf2618984f15d6ba4324cf6bbcb3a66e20ab811-84f15d6ba4324cf6bbcb3a66e20ab811-698c380618a6056fb1f9c3bccf653850"
[2024-12-21 09:14:53.032 +00:00] [INF] Processing playback tracker : "a80c6a531bf2618984f15d6ba4324cf6bbcb3a66e20ab811-84f15d6ba4324cf6bbcb3a66e20ab811-698c380618a6056fb1f9c3bccf653850"
[2024-12-21 09:15:13.558 +00:00] [INF] Processing playback tracker : "a80c6a531bf2618984f15d6ba4324cf6bbcb3a66e20ab811-84f15d6ba4324cf6bbcb3a66e20ab811-698c380618a6056fb1f9c3bccf653850"
[2024-12-21 09:15:25.006 +00:00] [INF] "Webhook Item Added Notifier" Completed after 0 minute(s) and 0 seconds
[2024-12-21 09:15:34.301 +00:00] [INF] Processing playback tracker : "a80c6a531bf2618984f15d6ba4324cf6bbcb3a66e20ab811-84f15d6ba4324cf6bbcb3a66e20ab811-698c380618a6056fb1f9c3bccf653850"
[2024-12-21 09:15:54.758 +00:00] [INF] Watching directory "/mnt/remote/Anime/TV"
[2024-12-21 09:15:54.997 +00:00] [INF] Processing playback tracker : "a80c6a531bf2618984f15d6ba4324cf6bbcb3a66e20ab811-84f15d6ba4324cf6bbcb3a66e20ab811-698c380618a6056fb1f9c3bccf653850"
[2024-12-21 09:16:15.541 +00:00] [INF] Processing playback tracker : "a80c6a531bf2618984f15d6ba4324cf6bbcb3a66e20ab811-84f15d6ba4324cf6bbcb3a66e20ab811-698c380618a6056fb1f9c3bccf653850"
[2024-12-21 09:16:36.298 +00:00] [INF] Processing playback tracker :  #"a80c6a531bf2618984f15d6ba4324cf6bbcb3a66e20ab811-84f15d6ba4324cf6bbcb3a66e20ab811-698c380618a6056fb1f9c3bccf653850"
```


The PR successfully extracts subtitle files from MKS containers during playback. However, I noticed that it doesn't extract embedded fonts from the MKS. 
Jellyfin uses the local system fonts to render subtitles, which causes ASS-specific styling and animations to be lost.

Gladtbam:
I found that ASS subtitles, even those not in MKS containers, are also being processed as SRT.

gnattu:
> I found that ASS subtitles, even those not in MKS containers, are also being processed as SRT.

Such behavior is client-specific and is a workaround for clients did not implement a native ass renderer.

Gladtbam:
> > I found that ASS subtitles, even those not in MKS containers, are also being processed as SRT.
> 
> Such behavior is client-specific and is a workaround for clients did not implement a native ass renderer.

Oh, I see—it’s handled differently on the TV side. I found a bug where, if the subtitles are bilingual, the fonts of the two lines overlap with each other.

Shadowghost (changes requested):
I tested this and it seems like you need to add a check for `string.IsNullOrEmpty(subtitleStream.Path)` otherwise you'll get null  pointers on internal files (see https://github.com/Shadowghost/jellyfin/commit/d7f8d318f1b021c81407164e5bcb979565944574)

IceStormNG:
> I tested this and it seems like you need to add a check for `string.IsNullOrEmpty(subtitleStream.Path)` otherwise you'll get null pointers on internal files.

Yeah. I found that out a few days ago. I have only very few files where the subs are in the main video file. I fixed it on my install, but haven't pushed it yet. Will upload the fix for that tomorrow or so.

Shadowghost (commented):
You should use `string.IsNullOrEmpty` to also immediately ignore empty strings

IceStormNG:
> You should use `string.IsNullOrEmpty` to also immediately ignore empty strings

Alright. Changed. I used regular null check because the string comparison would handle empty strings. But I can also use this method.

Shadowghost (approved):

JPVenson (approved):
can't speak to the impact of the code just the code-quallity which is fine.

crobibero (approved):

mihawk90:
Before this gets merged I just want to note I left a longer comment on the original issue as well, which brings up some additional considerations.

Shadowghost:
@IceStormNG can you please rebase on current master?

IceStormNG:
> @IceStormNG can you please rebase on current master?

Sure. The branch is rebased.

mihawk90:
I still don't think this solution makes sense in the context of ASS subtitles, see my previous comment above.

This is fine for SRT and other unformated subtitles, but this *will* break ASS scripts using fonts as already noted.

Shadowghost:
@mihawk90 this PR at least fixe the issues where we are burning in the subs and we don't have the required fonts available locally (because up until now the MKS fonts were not extracted)
That aside, in one of my PRs which moved the extracted files I also changed them so the extracted files are named properly - this could possibly also fix delivery of required fonts since the lookup by name now works.

mihawk90:
> this PR at least fixe the issues where we are burning in the subs and we don't have the required fonts available locally

Not sure what you're referring to? Burning the subtitles has no issues with fonts, see my testing and screenshot in the original issue.

That aside, I don't think it should be the goal to introduce new bugs while fixing another just to get it merged. One solution was already proposed (namely remuxing in case of ASS), which should work with any client across the board.

It's not like this concern is new either. It was first mentioned in comment 4 in December last year, that's half a year ago. The fact that noone even reacted to that concern boggles me but besides that I don't understand why it has to be rushed now instead of addressing it.

PR #12736, opened 2024-09-28
Add ImageSource to MediaSegment

Shadowghost:
**Changes**
* Add `ImageSource` to `MediaSegment`

Implements https://github.com/orgs/jellyfin/projects/63/views/1?pane=issue&itemId=80829926

**What is this for?**
`ImageSource`  onsists of the URI path to an image representing the segment.


JPVenson (approved):

nielsvanvelzen:
I think we should discuss this a bit more first. What would the use-case be for an image in the context of a segment? I'd imagine it would be different per segment type (for all current types it would probably be similar to chapter images).

Then if we add those images, why would the server store an URL to the image, wouldn't it be better to integrate it with our existing image functionality (e.g. image API to allow resizing/format conversion/blurhashing)?

Aside: this PR also does not add anything to the API yet, so retrieving images isn't possible at all.

PR #12737, opened 2024-09-28
Never set RemoveOldMetadata when replacing all metadata.

nielsvanvelzen:
When RemoveOldMetadata is set it will clear all existing metadata from items before applying the new information. This breaks virtual seasons as episodes will never have a ParentIndexNumber set unless the season is part of the episode file name (which in a lot of cases, it is not). The fallback to "season 1" is only available in EpisodeResolver which is not called during metadata refresh. When the ParentIndexId is reset to null the RemoveObsoleteSeasons method in SeriesMetadataService will remove the virtual season(s) as it thinks there are no episodes in them.

This change only affects the behavior when using a full metadata refresh, other cases of the season unknown issue have been fixed in my previous fixes (#12240 and #12356) in 10.9.z releases.

**Changes**

- Never set RemoveOldMetadata when replacing all metadata.

**Issues**

### DIFF
--- a/Jellyfin.Api/Controllers/ItemRefreshController.cs
+++ b/Jellyfin.Api/Controllers/ItemRefreshController.cs
@@ -80,8 +80,7 @@ public class ItemRefreshController : BaseJellyfinApiController
                 || imageRefreshMode == MetadataRefreshMode.FullRefresh
                 || replaceAllImages
                 || replaceAllMetadata,
-            IsAutomated = false,
-            RemoveOldMetadata = replaceAllMetadata
+            IsAutomated = false
         };
 
         _providerManager.QueueRefresh(item.Id, refreshOptions, RefreshPriority.High);


gnattu:
I'm afraid this is not a proper fix because this will break replace logic once again especially when NFO files are involved...

nielsvanvelzen:
This restores the behavior like it was in 10.8 and earlier so it is in fact a proper fix.

gnattu:
> This restores the behavior like it was in 10.8 and earlier

You cannot say that without reverting ALL 10.9 changes, especially we have a totally different metadata source preference order when nfo files are involved. You would end up in a situation that it is neither 10.8 (because you still allow the NFO to overwrite database) and broken (replace does not really replace all fields).

nielsvanvelzen:
Closing in favor of #12739

Issue #12045, opened 2024-06-09
[Issue]: TV show not listing any seasons because season is flagged as virtual


jellouser:
### Please describe your bug

I'm using locally generated NFO files for TV recordings: when the recording is completed then NFO file is created for the episode.

The problem is that when first episode for the show is recorded Jellyfin should create 1) show 2) season and 3) episode which it does according to database items, but the season is flagged as "IsVirtualItem" and because of this not listed in the user interface.

I am able "fix" this by forcing full metadata update for the show but it is annoying workaround and should work without it. After the season is "fixed" all episodes for this season work fine, until next new season which also needs "fix".

Looking at the code SeasonMetadataService SaveIsVirtualItem has some logic related to episode count, maybe episode does yet not "exist" at this point which would explain metadata update "fixing" it later? This problem also existed in 10.8.x.

EDIT: It seems that for master I need to refresh metadata twice until episode count is correct (1) and virtual flag is set to false for season.

### Reproduction Steps

1. Create a show folder including episode video file and NFO file for episode with title, season, seasonnumber and episode tags.
2. After automatic scan database information is correct for show, season and episode but season is marked as virtual
3. Do full metadata fresh for the show, now season is no longer marked as virtual and displayed in UI

### Jellyfin Version

Master branch

### Environment

```markdown
- OS:
- Linux Kernel:
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

```shell
<recording started>

[12:50:30] [DBG] [18] Emby.Server.Implementations.IO.LibraryMonitor: New file refresher created for /mnt/Recordings/Show Name
...
[13:02:01] [DBG] [39] Emby.Server.Implementations.IO.LibraryMonitor: Timer stopped.
[13:02:01] [INF] [39] Emby.Server.Implementations.IO.LibraryMonitor: Recordings (/mnt/Recordings) will be refreshed.
[13:02:01] [DBG] [40] MediaBrowser.Providers.TV.SeriesMetadataService: Running SeriesNfoProvider for /mnt/Recordings/Show Name
[13:02:01] [DBG] [40] MediaBrowser.Providers.TV.SeriesMetadataService: SeriesNfoProvider returned no metadata for /mnt/Recordings/Show Name
[...]

PR #12739, opened 2024-09-28
Pre-fill ParentIndexNumber to preserve it between scans

Shadowghost:
Alternative to #12737

**Changes**
* Preserve `ParentIndexNumber`

### DIFF
--- a/MediaBrowser.Providers/Manager/MetadataService.cs
+++ b/MediaBrowser.Providers/Manager/MetadataService.cs
@@ -675,6 +675,7 @@ namespace MediaBrowser.Providers.Manager
             };
             temp.Item.Path = item.Path;
             temp.Item.Id = item.Id;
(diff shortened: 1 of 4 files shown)


nielsvanvelzen (approved):
I can confirm this fixes the issue where virtual seasons would be removed when using "Replace all metadata".

nyanmisaka (approved):

crobibero (approved):

jellouser:
Unfortunately this breaks my setup with "Display specials within seasons they aired in" checked as workaround to get proper seasons created using NFOs - see https://github.com/jellyfin/jellyfin/issues/12045

Reverting this pull request correct seasons are again created with  "Display specials within seasons they aired in" _checked_.

DarkFighterLuke:
This PR is also the cause of https://github.com/jellyfin/jellyfin/issues/13358.

PR #12741, opened 2024-09-28
Do not consider tags in search

Shadowghost:
The extended tag search has apparently slowed down search responses significantly.

**Changes**
* Reverts #8914 and #8842
* Fixes CS1591 in SqliteItemRepository


Bond-009 (approved):

crobibero (approved):

VigibotDev:
Great job. Search is now much faster and we can search for a tag by clicking on it in a movie page.

Issue #12740, opened 2024-09-28
[Issue]: Server spawns many ffmpeg/ffprobe processes no matter the settings, exhausting IO, RAM and CPU


detly:
### This issue respects the following points:

- [X] This is a **bug**, not a question or a configuration issue; Please visit our forum or chat rooms first to troubleshoot with volunteers, before creating a report. The links can be found [here](https://jellyfin.org/contact/).
- [X] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [X] I'm using an up to date version of Jellyfin Server stable, unstable or master; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.
- [X] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [X] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Description of the bug

I'm using Jellyfin 10.9.11 on a Raspberry Pi 2B+ running Raspbian 12. It's installed from the apt repo at `https://repo.jellyfin.org/debian`, not running in Docker.

I am aware that running Jellyfin on an RPi is not supported, but:

1. this issue does not _seem_ like it depends on the platform
2. you probably don't want it to happen on a higher spec machine either
3. it was running absolutely fine on 10.8.13 or lower for _years_ before this
4. I don't have another (supported) machine to try to repro this on
5. if you have suggestions for how I could debug this I'm happy to try them
 
Shortly after starting, Jellyfin appears to launch `ffmpeg` processes with no limit on number. This completely maxes out memory, CPU and IO to the point where the service is killed by eg. systemd, or the device itself needs to be hard power cycled.

So far I have:

- set `ParallelImageEncodingLimit` to `1` in `/etc/jellyfin/system.xml`
- set `LibraryMetadataRefreshConcurrency` to `1` in `/etc/jellyfin/system.xml`
- set `LibraryScanFanoutConcurrency` to `1` in `/etc/jellyfin/system.xml`
- set `EncodingThreadCount` to `1` in `/etc/jellyfin/encoding.xml`
- set `JELLYFIN_FFmpeg__analyzeduration="100M"` in `/etc/default/jellyfin`
- set `JELLYFIN_FFmpeg__probesize="50M"` in `/etc/default/jellyfin`

[...]

PR #12744, opened 2024-09-29
Add perf tradeoff mode to image extractor

gnattu:
On low-end CPUs, image extraction can be quite slow because the CPU has to decode and analyze 24 frames to pick the best one. If the input is an HDR image, it also needs to perform tone mapping. On a Raspberry Pi 4 level of performance, this usually means more than 10 seconds per image extracted, and for 4K HDR inputs, it may even timeout. This PR adds a tradeoff mode where the extracted image is the single frame that was picked. Ideally it would come from one of the key frame which is very fast to extract, but if that failed it will fallback to extract the slower normal frame. The frame picked by this mode is worse and less representative, but the process is significantly faster.

Another benefit of this mode is that there is no longer the need to pool and cache lots of frames in the memory, which might also be helpful in low-memory scenarios.

Users can activate this mode by export env var:

```
export JELLYFIN_FFMPEG__ImgExtractPerfTradeoff=true
```

**Changes**

**Issues**

Might make users like #12740 happier.

### DIFF
--- a/Emby.Server.Implementations/ConfigurationOptions.cs
+++ b/Emby.Server.Implementations/ConfigurationOptions.cs
@@ -21,6 +21,7 @@ namespace Emby.Server.Implementations
             { BindToUnixSocketKey, bool.FalseString },
             { SqliteCacheSizeKey, "20000" },
             { FfmpegSkipValidationKey, bool.FalseString },
+            { FfmpegImgExtractPerfTradeoffKey, bool.FalseString },
             { DetectNetworkChangeKey, bool.TrueString }
         };
     }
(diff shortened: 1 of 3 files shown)


nyanmisaka (approved):

Shadowghost (approved):

crobibero (approved):

Bond-009 (approved):

rieje:
This worked for my Raspberry Pi 4 (I did not need to configure any other options) where none of the (embedded) images were shown despite scanning the library and waiting indefinitely, but only when I deleted all libraries. Refreshing metadata or adding a new library did not show the images. Not sure if this is expected behavior because I would think refreshing metadata would be enough and I would not have deleted the libraries if it wasn't a small one for testing purposes.

PR #12772, opened 2024-10-04
Include Channel Groups when responding to LiveTV channel requests

joshjryan:
LiveTV channels can optionally be grouped together. This is typically the case with M3U sources.

Jellyfin was already parsing the ChannelGroup out of the m3u and storing it in the db, but the value was never returned to any API clients.  This change adds the ChannelGroup to Channel Items.

This should allow clients to address https://features.jellyfin.org/posts/186/iptv-channel-groupings, which will allow client applications to improve their UI by grouping channels and/or allowing users to filter by group.


Shadowghost (approved):

topiga:
Up

Bond-009 (approved):

nielsvanvelzen (changes requested):
From a client developer perspective I can't do anything with this. At the very least the id of the channel should be used instead of the name. Additionally, I should be able to get all channel groups from the API, filter channels by group etc.

Otherwise this is just useless bloat on de baseitem model.

joshjryan:
Initially I submitted this because "if it's available, we might as well return it, just in a case a client wants it"...but @nielsvanvelzen is correct. There should also be an API endpoint to get the Channel Groups, sort by group, and filter by group.
I'll see if I can make time to re-do this PR to include those API additions.

Bond-009:
Closing as stale, if you do end up reworking this PR please open a new one

PR #12775, opened 2024-10-05
Fix Use GetAsync to get headers for stream type check to eliminate timeout on Live TV streams

joshjryan:
Some IPTV proxy applications (e.g. https://github.com/xteve-project/xTeVe) do not support HTTP HEAD requests. This would cause a timeout the first time that a connection to a new channel was initiated. Clients would have to wait for 100s timeout to elapse before initiating the stream.

M3UTunerHost was making a HEAD request to check the mimetype of the stream.  This change updates M3UTunerHost to perform a GetAsync, with `HttpCompletionOption.ResponseHeadersRead` in order to quickly get the headers for the mimetype check.

### DIFF
--- a/src/Jellyfin.LiveTv/TunerHosts/M3UTunerHost.cs
+++ b/src/Jellyfin.LiveTv/TunerHosts/M3UTunerHost.cs
@@ -102,9 +102,8 @@ namespace Jellyfin.LiveTv.TunerHosts
                 {
                     try
                     {
-                        using var message = new HttpRequestMessage(HttpMethod.Head, mediaSource.Path);
                         using var response = await _httpClientFactory.CreateClient(NamedClient.Default)
-                            .SendAsync(message, cancellationToken)
+                            .GetAsync(mediaSource.Path, HttpCompletionOption.ResponseHeadersRead, cancellationToken)
                             .ConfigureAwait(false);
 
                         if (response.IsSuccessStatusCode)


gnattu:
I'm personally fine with this approach, but I'm a bit concerned about whether this will introduce compatibility for another sets of TV Tuners that implements http server badly. Technically it is already workaround-able by disabling "allow shared stream" in the Tuner settings in current unstable, or at least it is supposed to be workaround-able by that. Can you check if that is the case? If disabling that option cannot workaround your problem then we need to fix this there as well.

joshjryan:
> I'm personally fine with this approach, but I'm a bit concerned about whether this will introduce compatibility for another sets of TV Tuners that implements http server badly. Technically it is already workaround-able by disabling "allow shared stream" in the Tuner settings in current unstable, or at least it is supposed to be workaround-able by that. Can you check if that is the case? If disabling that option cannot workaround your problem then we need to fix this there as well.

Thanks for the quick discussion @gnattu !

You make a great point.  Until you just mentioned that, I had forgotten that I still had my 10.10 dev server using the 10.9 UI (which doesn't have the Allow Stream Sharing checkbox)....I can't believe I didn't notice that.

So: that'll give a reasonable work-around for now, if you feel like this change is risky for 10.10.   If that is the case, I think I'd still ask for this change to be merged for a future release, so that xteve users can leverage the stream sharing feature.

crobibero:
You may want to consider moving to one of the xteve forks since xteve hasn't been updated in 4 years. The "threadfin" fork is actively maintained and supports http head requests.

joshjryan:
> You may want to consider moving to one of the xteve forks since xteve hasn't been updated in 4 years. The "threadfin" fork is actively maintained and supports http head requests.

I'm not sure what it says about me, but this morning I asked myself "would I rather figure this out and submit a PR, or finally move to Threadfin?"....and my answer was to submit a PR :)

ortonomy:
> The "threadfin" fork is actively maintained and supports http head requests.

No it doesn't @crobibero  - [or if it does, it's not properly responding with them...](https://github.com/jellyfin/jellyfin/issues/9813#issuecomment-2869922608)

crobibero:
> > The "threadfin" fork is actively maintained and supports http head requests.
> 
> Are you sure @crobibero - [if it does, it's not properly responding to them...](https://github.com/jellyfin/jellyfin/issues/9813#issuecomment-2869922608)

Yes. https://github.com/Threadfin/Threadfin/issues/244
If it isn't working (again), you should report the issue in the Threadfin repo.

ortonomy:
> > > The "threadfin" fork is actively maintained and supports http head requests.
> 
> > 
> 
> > Are you sure @crobibero - [if it does, it's not properly responding to them...](https://github.com/jellyfin/jellyfin/issues/9813#issuecomment-2869922608)
> 
> 
> 
> Yes. https://github.com/Threadfin/Threadfin/issues/244
> 
> If it isn't working (again), you should report the issue in the Threadfin repo.

Nope. Haha. What a comedy of errors, if you follow the breadcrumbs on that issue you shared:

1. It got patched in v1.12.2
2. It got unpatched in a new version 
3. Someone opened a PR to repatch it
4. The PR got merged but not into main so it's not in release at all 😂: https://github.com/Threadfin/Threadfin/pull/499#issuecomment-2588430030

Bond-009:
Closing as working as intended

PR #12776, opened 2024-10-05
Clean up logging in MediaSegmentManager

crobibero:
The current log statements balloon the log files with redundant information, so only log at information if we are actually attempting to extract segments


Shadowghost (approved):

nielsvanvelzen (approved):
How to save disk space with one simple trick they don't want you to know about

Issue #12738, opened 2024-09-28
How to use external subtitle in bdmv format


aSoloist:
### This issue respects the following points:

- [X] This is a **bug**, not a question or a configuration issue; Please visit our forum or chat rooms first to troubleshoot with volunteers, before creating a report. The links can be found [here](https://jellyfin.org/contact/).
- [X] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [X] I'm using an up to date version of Jellyfin Server stable, unstable or master; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.
- [X] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [X] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Description of the bug

When I play a BDMV movie, because the embedded subtitles do not have the subtitles I want, I download external subtitles, but I cannot find the subtitles in the subtitle selection.
![image](https://github.com/user-attachments/assets/347e8373-57c5-4fc6-afbb-c965076c1e77)
![image](https://github.com/user-attachments/assets/8faf6064-504b-4d17-b18c-f7c4e70facd3)

### Reproduction steps

1.sign in on the jellyfin web client
2.choice one BDMV movie
3.change the subtitle

### What is the current _bug_ behavior?

I cannot find the subtitles in the BDMV movie subtitle selection.

### What is the expected _correct_ behavior?

the subtitle selection can recognize external subtitle in BDMV movie

### Jellyfin Server version

Older*

### Specify version number

10.8.13

### Specify the build version

10.8.13

### Environment

```markdown
- OS:Debian
- Linux Kernel:4.4.302+
- Virtualization:Docker
- Clients:Browser
- Browser:Chrome 129.0.6668.60
- FFmpeg Version:5.1.4-Jellyfin
- Playback Method:Transcode
- Hardware Acceleration:QSV
- GPU Model:HD610
- Plugins:AudioDB,MusicBrainz,OMDb,Open Subtitles,Studio Images,Subtitle Extract,TMDb
- Reverse Proxy:none
- Base URL:none
- Networking:Bridge
- Storage:local
```

### Jellyfin logs

```shell
[...]

PR #12779, opened 2024-10-06
Fixed the issue that Blu-ray movies cannot get external subtitles

aSoloist:
resolve the issues https://github.com/jellyfin/jellyfin/issues/12738, Fixed the issue that Blu-ray movies cannot get external subtitles.

**Changes**
Keep the external stream, add them after adding the blurayInfo stream.

**Issues**
https://github.com/jellyfin/jellyfin/issues/12738

### DIFF
--- a/MediaBrowser.Providers/MediaInfo/FFProbeVideoInfo.cs
+++ b/MediaBrowser.Providers/MediaInfo/FFProbeVideoInfo.cs
@@ -323,9 +323,17 @@ namespace MediaBrowser.Providers.MediaInfo
 
             var ffmpegVideoStream = mediaStreams.FirstOrDefault(s => s.Type == MediaStreamType.Video);
 
+            List<MediaStream> externalStreams = mediaStreams.FindAll(s => s.IsExternal);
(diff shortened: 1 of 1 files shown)


MediaBrowser.Providers/MediaInfo/FFProbeVideoInfo.cs
```diff
@@ -323,9 +323,17 @@ private void FetchBdInfo(Video video, ref ChapterInfo[] chapters, List<MediaStre
...
             // Fill video properties from the BDInfo result
             mediaStreams.Clear();
             mediaStreams.AddRange(blurayInfo.MediaStreams);
+            // Fill video properties from external stream
+            var startIndex = mediaStreams.Count;
+            foreach (var externalStream in externalStreams)
+            {
+                externalStream.Index = startIndex++;
+                mediaStreams.Add(externalStream);
+            }
```
Shadowghost, line 336: Usually external streams are added first into the List. Otherwise nice catch!

Shadowghost:
@aSoloist this should be fine once the stream order is fixed

PR #12783, opened 2024-10-06
Implement TaskTriggerInfoType enum

LePips:
Just something I noticed while we're implementing tasks in Swiftfin. Not necessary for 10.10.

**Changes**
Implements `TaskTriggerInfoType` for the `TaskTriggerInfo.Type` field.

### DIFF
--- a/Emby.Server.Implementations/ScheduledTasks/ScheduledTaskWorker.cs
+++ b/Emby.Server.Implementations/ScheduledTasks/ScheduledTaskWorker.cs
@@ -471,7 +471,7 @@ namespace Emby.Server.Implementations.ScheduledTasks
                     new()
                     {
                         IntervalTicks = TimeSpan.FromDays(1).Ticks,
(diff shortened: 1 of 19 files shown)


Shadowghost (approved):

Bond-009 (approved):

JPVenson:
that would likely also break the webui right?

LePips:
No, as these are the only used trigger values. There used to be a `StartupTrigger` [that was removed recently](https://github.com/jellyfin/jellyfin/commit/f68038f2cff4ac94ac5afe97b5bfb5c9e172e6ff) and the [web hasn't been updated to accommodate that](https://github.com/jellyfin/jellyfin-web/blob/c3e253d98d81dd3c1963108b0b98f20d534f7d7f/src/controllers/dashboard/scheduledtasks/scheduledtask.js#L90), but that value will now never be returned.

JPVenson (approved):

crobibero (approved):

PR #12824, opened 2024-10-15
Don't check remote sources for trickplay

gnattu:
These sources will not have trickplay generated anyway.

**Changes**

**Issues**

### DIFF
--- a/Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
+++ b/Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
@@ -498,12 +498,10 @@ public class TrickplayManager : ITrickplayManager
         var trickplayManifest = new Dictionary<string, Dictionary<int, TrickplayInfo>>();
         foreach (var mediaSource in item.GetMediaSources(false))
         {
-            if (mediaSource.IsRemote)
+            if (mediaSource.IsRemote || !Guid.TryParse(mediaSource.Id, out var mediaSourceId)
             {
                 continue;
             }
-
-            var mediaSourceId = Guid.Parse(mediaSource.Id);
             var trickplayResolutions = await GetTrickplayResolutions(mediaSourceId).ConfigureAwait(false);
 
             if (trickplayResolutions.Count > 0)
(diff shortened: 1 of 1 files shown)


JPVenson (approved):

nyanmisaka (approved):

Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
```diff
@@ -498,6 +498,11 @@ public async Task<Dictionary<string, Dictionary<int, TrickplayInfo>>> GetTrickpl
         var trickplayManifest = new Dictionary<string, Dictionary<int, TrickplayInfo>>();
         foreach (var mediaSource in item.GetMediaSources(false))
         {
+            if (mediaSource.IsRemote)
```
JPVenson, line 501:
```suggestion
            if (mediaSource.IsRemote || !Guid.TryParse(mediaSource.Id, out var mediaSourceId)
            {
                continue;
            }
```

Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
```diff
@@ -498,6 +498,11 @@ public async Task<Dictionary<string, Dictionary<int, TrickplayInfo>>> GetTrickpl
...
+            {
+                continue;
+            }
+
             var mediaSourceId = Guid.Parse(mediaSource.Id);
```
JPVenson, line 506:
```suggestion
            if (mediaSource.IsRemote || !Guid.TryParse(mediaSource.Id, out var mediaSourceId)
            {
                continue;
            }
```

gnattu: Do we really want to do this?

JPVenson: dont see harm in just checking for it.
Could add a log debug entry as well but its better then just failing.

JPVenson: better would be to make the Id field non nullable/required but thats a bit outside the scope

Bond-009 (approved):

crobibero (approved):

PR #12832, opened 2024-10-16
Check for path overlaps

JPVenson:
Added marker file to identify most build in folders and check for cross mounting folders to then stop server.


Jellyfin.Server/StartupOptions.cs
```diff
@@ -73,6 +73,12 @@ public class StartupOptions : IStartupOptions
         [Option("nonetchange", Required = false, HelpText = "Indicates that the server should not detect network status change.")]
         public bool NoDetectNetworkChange { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether the server should not detect folders that are merged by configuration.
+        /// </summary>
+        [Option("nocrossaccesscheck", Required = false, HelpText = "Indicates that the server should not detect overlapping folder.")]
+        public bool NoFolderCrossCheck { get; set; }
```
nielsvanvelzen, line 80: I don't really see a good use for this option. We know that the server will (horribly) break with the overlapping paths so why even allow a user to ignore that?

JPVenson: i _can_ see usecases where though some docker magic you _could_ non-destructivly put for example your config and metadata path under the same root. Its just a fallback for people that do crazy shit (or also LSIO containers i have not tested).

nielsvanvelzen: But do we really want to support that niche?

JPVenson: well i see no harm in offering a simple solution if we break something with that. Its not complicated nor hard to maintain. But if the concenses is against that, i have no issue with removing the toggle.

EraYaN: IMO we should break/discourage setups that are asking for trouble, but maybe it's not super user friendly now but better in the long run. And in this case I don't think we can prevent some of the scheduled tasks for temp files clobbering say ones config, right? Better to just have people not be stupid.

JPVenson: I would still argue to leave this in __for now__. Its fine if we break those users installs (if they are working) but not without warning and time to do it right for as long as we can.

Emby.Server.Implementations/AppBase/BaseApplicationPaths.cs
```diff
@@ -105,5 +106,47 @@ protected BaseApplicationPaths(
...
+            CreateAndCheckMarker(LogDirectoryPath, "log");
+            CreateAndCheckMarker(PluginsPath, "plugin");
+            CreateAndCheckMarker(ProgramDataPath, "data");
+            CreateAndCheckMarker(CachePath, "cache", true);
+            CreateAndCheckMarker(DataPath, "data");
+        }
+
+        /// <inheritdoc cref="IApplicationPaths"/>
+        public void CreateAndCheckMarker(string path, string markerName, bool recursive = false)
+        {
```
jhparker88, line 123: This may seem irrelevant but I had a hard time to understand what Marker means in this particular context and only after reading the signature comment that it was something about a file as placeholder. Maybe for clarity rename it something more specific such as `CreateAndCheckFilePlaceholder`.

JPVenson: those files are not placeholders.
Think of them as differently colored permanent markers 
71yyAy2jS7L _AC_SL1500_-265399118

MediaBrowser.Common/Configuration/IApplicationPaths.cs
```diff
@@ -84,11 +84,23 @@ public interface IApplicationPaths
         /// </summary>
         /// <value>The magic string used for virtual path manipulation.</value>
         string VirtualDataPath { get; }
-
-        /// <summary>
```
Shadowghost, line 88: This should not be removed I think

MediaBrowser.Controller/IServerApplicationPaths.cs
Shadowghost, line 1: Why are these needed now without changes to the actual file?

Emby.Server.Implementations/AppBase/BaseApplicationPaths.cs
```diff
@@ -75,5 +76,47 @@ protected BaseApplicationPaths(
 
         /// <inheritdoc />
         public string TrickplayPath => Path.Combine(DataPath, "trickplay");
+        
+        /// <inheritdoc cref="IApplicationPaths"/>
```
Shadowghost, line 80:
```suggestion
        /// <inheritdoc />
```

Emby.Server.Implementations/AppBase/BaseApplicationPaths.cs
```diff
@@ -75,5 +76,47 @@ protected BaseApplicationPaths(
...
+        {
+            CreateAndCheckMarker(ConfigurationDirectoryPath, "config");
+            CreateAndCheckMarker(LogDirectoryPath, "log");
+            CreateAndCheckMarker(PluginsPath, "plugin");
+            CreateAndCheckMarker(ProgramDataPath, "data");
+            CreateAndCheckMarker(CachePath, "cache", true);
+            CreateAndCheckMarker(DataPath, "data");
+        }
+
+        /// <inheritdoc cref="IApplicationPaths"/>
```
Shadowghost, line 91:
```suggestion
        /// <inheritdoc />
```

MediaBrowser.Common/Configuration/IApplicationPaths.cs
JPVenson: github merge ui sucks

MediaBrowser.Common/Configuration/IApplicationPaths.cs
Shadowghost, line 1: Comments in this file seem to be off in general 🤔

JPVenson:
@Shadowghost whitespace nonsense should be fixed now

Shadowghost (approved):

theguymadmax:
I tested by creating a library in the cache folder and then restarting, which failed to start. How does one fix the error without being able to start the server? 

[Logs](https://gist.github.com/theguymadmax/c6a5ce9217456e8bc156499bc8eacdfc)

theguymadmax:
Created a new instance and it wouldn't start.
`System.InvalidOperationException: Exepected to find only .jfcache but found marker for C:\Users\Max\AppData\Local\jellyfin\cache\transcodes\.jftranscode.`

[Logs](https://gist.github.com/theguymadmax/9956e7db1517b71d0f38df5f7f41317c)

Emby.Server.Implementations/AppBase/BaseApplicationPaths.cs
```diff
@@ -75,5 +76,47 @@ protected BaseApplicationPaths(
...
+
+        /// <inheritdoc />
+        public void CreateAndCheckMarker(string path, string markerName, bool recursive = false)
+        {
+            if (!Directory.Exists(path))
+            {
+                Directory.CreateDirectory(path);
+            }
+
+            CheckOrCreateMarker(path, $".jf{markerName}", recursive);
```
crobibero, line 99:
```suggestion
            CheckOrCreateMarker(path, $".jellyfin-{markerName}", recursive);
```

crobibero (approved):

PR #12833, opened 2024-10-16
Add filter for near same segment

JPVenson:
Added query parameter defaulting to true to filter out same typed segments with near same (within 1 sec) segments on get request.


fubar-coder:
I have some (dumb?) questions, because I'm curious:

* What does this PR tries to solve? Does this cause problems/bugs?
* Why are there "duplicates"? In other words: Why aren't those duplicates merged automatically when they're saved?
* Why is a media segment considered as duplicate if they're too close to each other? Why aren't those media segments merged instead?
* Maybe implementing a background job that merges those "duplicates" would be the better solution?
* Can there be overlapping media segment items?

JPVenson:
> What does this PR tries to solve? Does this cause problems/bugs?

Yes this can cause issues from player clients when there are multiple times the same typed segment. Ofc they could just filter on their own but for the sake of given clients an easier time the server should optionally provide this filtering.

>  Why are there "duplicates"? In other words: Why aren't those duplicates merged automatically when they're saved?
> Why aren't those media segments merged instead?
> Why is a media segment considered as duplicate if they're too close to each other? 

In contrast to the old plugins that existed in a vaccum, we can now run multiple providers that produce segments. For example our chapter segment plugin and intro skipper can both identify a segment as an Intro. Those 2 segment will not be saved or aggregated because we believe that we should always store exactly what comes from a provide, because this data can be used for other purposes too in the future and we will also provide more extensive management of ones segments to a plugin. Having them aggregated would destry that ownership and makes advanced mangement hard.

>  Maybe implementing a background job that merges those "duplicates" would be the better solution?

No. Only playback clients really care for that flat list of segments, other clients like a potential segment editor will like to get the _real_ unmodified list of segments so we have to keep them.

>  Can there be overlapping media segment items?
Yes. Only on request segments that have the same type and where their start and end timespan is within 1sec will be considered near-same. everything else will be delivered as-is.

fubar-coder:
Hi,

just 2 more questions:

> > Why are there "duplicates"? In other words: Why aren't those duplicates merged automatically when they're saved?
> > Why aren't those media segments merged instead?
> > Why is a media segment considered as duplicate if they're too close to each other?
> 
> In contrast to the old plugins that existed in a vaccum, we can now run multiple providers that produce segments. For example our chapter segment plugin and intro skipper can both identify a segment as an Intro. Those 2 segment will not be saved or aggregated because we believe that we should always store exactly what comes from a provide, because this data can be used for other purposes too in the future and we will also provide more extensive management of ones segments to a plugin. Having them aggregated would destry that ownership and makes advanced mangement hard.

1. Wouldn't merging be the better experience for a (streaming) client/user, instead of dropping the duplicates?
2. Is there a way to distinguish between segments (of the same type) created by different plugins?

JPVenson:
>Wouldn't merging be the better experience for a (streaming) client/user, instead of dropping the duplicates?

Maybe, but we do not aim for pixel perfect precision and we only drop dups if within 1 sec of each other so its good "enough" for now.

> Is there a way to distinguish between segments (of the same type) created by different plugins? IOW: Can we always determine the plugin that created a given segment?

Yes

Shadowghost (approved):

PR #12834, opened 2024-10-16
Have device deletion take list of ids

LePips:
While we're implementing administrative tasks in Swiftfin, I noticed that the endpoint for device deletion only takes a single id. The web performs "delete all" by iteratively calling the endpoint. This could be problematic for larger device counts and was just easy an easy optimization.

**Changes**

Changes the `id` parameter to accept a list of items, which occur through parameter repetition. This is not a breaking change.

My only concern is that trying to delete too many devices will create long URLs that may cause other issues. So, an implementation by falling back to the body if parameters weren't present may be warranted.

### DIFF
--- a/Jellyfin.Api/Controllers/DevicesController.cs
+++ b/Jellyfin.Api/Controllers/DevicesController.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
 using System.Linq;
 using System.Threading.Tasks;
@@ -116,7 +117,6 @@ public class DevicesController : BaseJellyfinApiController
     /// <summary>
     /// Deletes devices.
     /// </summary>
-    /// <param name="id">Optional. Device Id.</param>
     /// <param name="ids">Device Ids.</param>
     /// <response code="204">Device deleted.</response>
     /// <response code="404">Device not found.</response>
@@ -124,19 +124,24 @@ public class DevicesController : BaseJellyfinApiController
     [HttpDelete]
     [ProducesResponseType(StatusCodes.Status204NoContent)]
     [ProducesResponseType(StatusCodes.Status404NotFound)]
-    public async Task<ActionResult> DeleteDevice([FromQuery, ParameterObsolete] string? id, [FromQuery, ModelBinder(typeof(CommaDelimitedArrayModelBinder))] string[] ids)
+    public async Task<ActionResult> DeleteDevice([FromQuery, ModelBinder(typeof(CommaDelimitedArrayModelBinder))] string[] ids)
     {
-        string[] allIDs = ids.Concat(id != null ? [id] : Array.Empty<string>()).ToArray();
+        List<DeviceInfoDto> devices = new List<DeviceInfoDto>();
 
-        foreach (var i in allIDs)
+        foreach (var id in ids)
         {
-            var existingDevice = _deviceManager.GetDevice(i);
+            var existingDevice = _deviceManager.GetDevice(id);
             if (existingDevice is null)
             {
                 return NotFound();
             }
 
-            var sessions = _deviceManager.GetDevices(new DeviceQuery { DeviceId = i });
+            devices.Append(existingDevice);
+        }
+
+        foreach (var device in devices)
+        {
+            var sessions = _deviceManager.GetDevices(new DeviceQuery { DeviceId = device.Id });
 
             foreach (var session in sessions.Items)
             {
--- a/Jellyfin.Api/Controllers/DevicesController.cs
+++ b/Jellyfin.Api/Controllers/DevicesController.cs
@@ -1,7 +1,10 @@
 using System;
 using System.ComponentModel.DataAnnotations;
+using System.Linq;
 using System.Threading.Tasks;
+using Jellyfin.Api.Attributes;
 using Jellyfin.Api.Helpers;
+using Jellyfin.Api.ModelBinders;
 using Jellyfin.Data.Dtos;
 using Jellyfin.Data.Queries;
 using MediaBrowser.Common.Api;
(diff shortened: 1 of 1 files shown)


Jellyfin.Api/Controllers/DevicesController.cs
```diff
@@ -111,28 +114,34 @@ public async Task<ActionResult> UpdateDeviceOptions(
...
+        string[] allIDs = ids.Concat(id != null ? [id] : Array.Empty<string>()).ToArray();
+
+        foreach (var i in allIDs)
         {
-            return NotFound();
-        }
+            var existingDevice = _deviceManager.GetDevice(i);
+            if (existingDevice is null)
+            {
+                return NotFound();
```
nielsvanvelzen, line 136: This really is a no-go. You should first check all devices for existence before proceeding with the delete operation. Otherwise the client can never know which devices were deleted.

Jellyfin.Api/Controllers/DevicesController.cs
```diff
@@ -111,28 +114,34 @@ public async Task<ActionResult> UpdateDeviceOptions(
...
+    /// <param name="ids">Device Ids.</param>
     /// <response code="204">Device deleted.</response>
     /// <response code="404">Device not found.</response>
-    /// <returns>A <see cref="NoContentResult"/> on success, or a <see cref="NotFoundResult"/> if the device could not be found.</returns>
+    /// <returns>A <see cref="NoContentResult"/> on success, or a <see cref="NotFoundResult"/> if a device could not be found.</returns>
     [HttpDelete]
     [ProducesResponseType(StatusCodes.Status204NoContent)]
     [ProducesResponseType(StatusCodes.Status404NotFound)]
-    public async Task<ActionResult> DeleteDevice([FromQuery, Required] string id)
+    public async Task<ActionResult> DeleteDevice([FromQuery, ParameterObsolete] string? id, [FromQuery, ModelBinder(typeof(CommaDelimitedArrayModelBinder))] string[] ids)
```
nielsvanvelzen, line 127:
```suggestion
    public async Task<ActionResult> DeleteDevice([FromQuery, ModelBinder(typeof(CommaDelimitedArrayModelBinder))] string[] ids)
```

Dashboard API's don't follow our "no-break" rules so we can just remove the old parameter.

nielsvanvelzen, line 127: IIRC we don't use `CommaDelimitedArrayModelBinder` for new parameters, that's just for old API's

Jellyfin.Api/Controllers/DevicesController.cs
```diff
@@ -111,28 +114,34 @@ public async Task<ActionResult> UpdateDeviceOptions(
...
     {
-        var existingDevice = _deviceManager.GetDevice(id);
-        if (existingDevice is null)
+        string[] allIDs = ids.Concat(id != null ? [id] : Array.Empty<string>()).ToArray();
+
+        foreach (var i in allIDs)
```
nielsvanvelzen, line 131:
```suggestion
        foreach (var id in allIDs)
```

There is no rule that you must use `i` for variable names in a for(each) loop

nielsvanvelzen (changes requested):
Can you link the companion jellyfin-web PR for these changes in the PR description?

Jellyfin.Api/Controllers/DevicesController.cs
JPVenson: i for index in for loops is acceptable but not in foreach where the variable should be a self explanetory name

Jellyfin.Api/Controllers/DevicesController.cs
LePips: What should be used instead?

crobibero: You don't need to specify a model binder, the comma-delimited binder is for legacy support

LePips: So instead, we should delimit the ids manually? Searching for `] string[]`, all instances have some kind of model binder. Testing with a comma delimited string without the model binder returns the entire parameter for me.

Or, am I missing something in .NET that does this for us?

Jellyfin.Api/Controllers/DevicesController.cs
```diff
@@ -111,28 +115,38 @@ public async Task<ActionResult> UpdateDeviceOptions(
...
+    /// <returns>A <see cref="NoContentResult"/> on success, or a <see cref="NotFoundResult"/> if a device could not be found.</returns>
     [HttpDelete]
     [ProducesResponseType(StatusCodes.Status204NoContent)]
     [ProducesResponseType(StatusCodes.Status404NotFound)]
-    public async Task<ActionResult> DeleteDevice([FromQuery, Required] string id)
+    public async Task<ActionResult> DeleteDevice([FromQuery, ModelBinder(typeof(CommaDelimitedArrayModelBinder))] string[] ids)
     {
-        var existingDevice = _deviceManager.GetDevice(id);
-        if (existingDevice is null)
+        List<DeviceInfoDto> devices = new List<DeviceInfoDto>();
```
JPVenson, line 129:
```suggestion
        var devices = ids.Select(_deviceManager.GetDevice).ToArray();
        if (devices.Any(f => f is null))
        {
        	return NotFound();
        }
```

Jellyfin.Api/Controllers/DevicesController.cs
nielsvanvelzen: You would change it from `[FromQuery, ModelBinder(typeof(CommaDelimitedArrayModelBinder))] string[] ids` to just `[FromQuery] string[] ids`. Then to supply ids you use `?ids=id1&ids=id2&ids=id3` etc.

Jellyfin.Api/Controllers/DevicesController.cs
```diff
@@ -111,28 +115,42 @@ public async Task<ActionResult> UpdateDeviceOptions(
...
+        foreach (var id in ids)
+        {
+            var existingDevice = _deviceManager.GetDevice(id);
+            if (existingDevice is null)
+            {
+                return NotFound();
+            }
+
+            devices.Add(existingDevice);
+        }
```
JPVenson, line 144:
```suggestion
```

Jellyfin.Api/Controllers/DevicesController.cs
```diff
@@ -111,28 +115,31 @@ public async Task<ActionResult> UpdateDeviceOptions(
...
+    /// <param name="ids">Device Ids.</param>
     /// <response code="204">Device deleted.</response>
     /// <response code="404">Device not found.</response>
-    /// <returns>A <see cref="NoContentResult"/> on success, or a <see cref="NotFoundResult"/> if the device could not be found.</returns>
+    /// <returns>A <see cref="NoContentResult"/> on success, or a <see cref="NotFoundResult"/> if a device could not be found.</returns>
     [HttpDelete]
     [ProducesResponseType(StatusCodes.Status204NoContent)]
     [ProducesResponseType(StatusCodes.Status404NotFound)]
-    public async Task<ActionResult> DeleteDevice([FromQuery, Required] string id)
+    public async Task<ActionResult> DeleteDevice([FromQuery] string[] ids)
```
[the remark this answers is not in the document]
LePips: Done in [`8e33f2a` (#12834)](https://github.com/jellyfin/jellyfin/pull/12834/commits/8e33f2ad1d948e0096db447c88aa2da5898bd610)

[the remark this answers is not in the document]
LePips: Since this is backwards compatible, should changes to the web be _necessary_ for this then?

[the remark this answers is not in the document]
nielsvanvelzen: No that should no longer be necessary as the existing requests will just work. Although it would be nice to eventually migrate that delete all button for performance.

LePips:
Pinging because we (Swiftfin) were just in this area again and reminded me of this PR. I've updated the PR description to better explain the motivations and state of this change.

JPVenson (approved):

nielsvanvelzen (changes requested):
API for 10.11 is set in stone. Marking as changes requested to avoid accidental merging.

Jellyfin.Api/Controllers/DevicesController.cs
```diff
@@ -111,28 +115,31 @@ public async Task<ActionResult> UpdateDeviceOptions(
...
     [ProducesResponseType(StatusCodes.Status404NotFound)]
-    public async Task<ActionResult> DeleteDevice([FromQuery, Required] string id)
+    public async Task<ActionResult> DeleteDevice([FromQuery] string[] id)
     {
-        var existingDevice = _deviceManager.GetDevice(id);
-        if (existingDevice is null)
+        var devices = id.Select(_deviceManager.GetDevice).ToArray();
+        if (devices.Any(f => f is null))
         {
             return NotFound();
```
nielsvanvelzen: Within a batching endpoint this would be a bad request (requesting one or multiple incorrect devices). Only if the id was a part of the path would we use a 404.

nielsvanvelzen (changes requested):

JPVenson (approved):

JPVenson:
@LePips you really have the worst of worst timings. I spoke with the team about the PR and apart from already 12.0 planned features we have entered the api freeze two days ago. I will put that on my 13.0 list.

nielsvanvelzen (approved):
Was convinced in chat to still review this (even though API reviews froze yesterday). This is fine **for now**. But in the longer term I want to make some changes that will break this behavior again;

1. Change the device API routes to have the device id in URL
2. Add a different approach for batch update/delete within the whole API

There is a high change point 1 will happen before point 2. And especially since the device API is meant for the dashboard we do not guarantee device safety so that might happen in 13.0 without notice.

LePips:
lol thanks, I've just been catching up on my PRs. Do what you wish with this.

Bond-009 (approved):

Issue #12808, opened 2024-10-12
The connection error message is not clear when the maximum number of simultaneous user sessions has been reached.


Nsbx:
### This issue respects the following points:

- [X] This is a **bug**, not a question or a configuration issue; Please visit our forum or chat rooms first to troubleshoot with volunteers, before creating a report. The links can be found [here](https://jellyfin.org/contact/).
- [X] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [X] I'm using an up to date version of Jellyfin Server stable, unstable or master; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.
- [X] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [X] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Description of the bug

When the “Maximum number of simultaneous user sessions” option is set to anything other than 0, and you try to connect n times + 1.
The resulting error message is unclear.

When the login/password pair is used, we get a message telling us that the login or password is incorrect.
And when we use a quick login code, the system tells us that the code doesn't exist.

Beyond that, I think this option is problematic.
Limited, the number of simultaneous streams is understandable, but the number of connections... it makes no sense.
If you “love tech”, you're likely to have a PC, a phone, a tablet and maybe even a connected TV or streaming key (apple tv, firestick).

In my case, it's a friend's server, and I sometimes switch between the Infuse, Jellyfin and Switftfin applications.
I didn't know he'd activated this option, and I soon found myself with the messages mentioned above.
And because of that (and the non-existent logs) we were looking in the wrong places.

### Reproduction steps

Set up a Maximum number of simultaneous user sessions.
Use an iphone and an ipad and try to log in on multiple app like Jellyfin, Swiftfin and Infuse.

### What is the current _bug_ behavior?

The server return an error message that that my couple login password is incorrect OR my login code is incorrect.

### What is the expected _correct_ behavior?

Instead, the server should return a message indicating that the maximum number of sessions has been reached.

### Jellyfin Server version

10.9.11+

### Specify the build version

10.9.11

### Environment

```markdown
Not needed here
```

### Jellyfin logs

```shell
No logs needed
```

### Additional information
Oh and in all this trouble, I was confronted with infuse loading ad infinitum (although I had been connected without any problems before).

[...]

PR #12842, opened 2024-10-18
Fix authenticateWithQuickConnect rethrow

wh1tewolfxx:
The exception being rethrown when the token was not found was a different exception type than the underlying exception.

**Changes**
 - Change rethrown exception type from SecurityException to ResourceNotFoundException.
 - Add new constructor for ResourceNotFoundException to support rethrowing caught exception. Copied from SecurityException Class.
 - Add additional response StatusCode to summary for 404NotFound.
 
**Fixes**
 - The exception is now caught and rethrown.
 - 404 Response Message for /Users/authenticateWithQuickConnect
Before: Unable to find request
After: [127.0.0.1] Unable to find request

### DIFF
--- a/Jellyfin.Api/Controllers/UserController.cs
+++ b/Jellyfin.Api/Controllers/UserController.cs
@@ -239,6 +239,7 @@ public class UserController : BaseJellyfinApiController
     /// <param name="request">The <see cref="QuickConnectDto"/> request.</param>
     /// <response code="200">User authenticated.</response>
     /// <response code="400">Missing token.</response>
+    /// <response code="404">Token not found.</response>
     /// <returns>A <see cref="Task"/> containing an <see cref="AuthenticationRequest"/> with information about the new session.</returns>
     [HttpPost("AuthenticateWithQuickConnect")]
     [ProducesResponseType(StatusCodes.Status200OK)]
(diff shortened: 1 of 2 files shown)


JPVenson:
To be frank using an exception here is not good in the first place. I would rather see the endpoint just returning the StatusCodeResult instead with the data.

wh1tewolfxx:
@JPVenson That makes sense. I notice that some of controllers returned status codes, while others were throwing exceptions. I was looking into Issue [#12808](https://github.com/jellyfin/jellyfin/issues/12808) and noticed the mismatch in the exception type. I think some modification to the AuthenticationResult to store some error information might help, when the result bubbles back up to the controller. That way logic can be used to return the correct StatusCode on the endpoint. Should I close this PR, and record my findings here on the issue mentioned above?

JPVenson:
Thats a question thats really not that easy to awnser. Part of the API is old and very convoluted and yes we are very aware of that. However there are a couple of factors that mix in here. Firstly we _try_ to keep existing api methods that are not "newer" as pristine as possible to not keep clients reinventing the weel each release. The "grand" api rewrite is the 11.00 version but there are some other things that need to be done before.

However in my opinion this satisfies a grander discussion about that topic which i would kindly ask you to open here(https://github.com/jellyfin/jellyfin-meta/discussions). Please describe the changes you want to make and why you want to make them so the whole team can discuss it. But i have to warn you, that is not a quick process as we really have to keep tabs on this old code which many clients rely on and we need to think about changing this kind of "basic" function, how it will effect all clients.

Also in addition we are currently in feature freeze until the next release so that would be perfect for you to get your proposed change anyway.

Bond-009:
Closing as stale

PR #12843, opened 2024-10-18
Add MaxActiveStreams functionality

Migs3:
Feature Add: Adding the functionality to limit Active Streams per user by a configurable number.

**Changes**
This adds a configurable number that, if a user exceeds by starting a new stream, will issue a stop command to another active stream, which will then effectively limit the number of active streams per user.
This piggybacks onto OnPlaybackStarted in SessionManager to check Sessions for active sessions for the user (NowPlayingItem) and then Issue a Stop command to as many Sessions as needed to be within the users' configured Max allowed Active Streams count.

NOTES:
1.) This is not "fully" tested. (I have only tested using jellyfin-web, not against other clients)
2.) The .First() doesn't seem to be working as I originally intended.  I think the desire would be to stop the "oldest" Session, but this doesn't seem to work as expected, not sure if this is important or not at this juncture.
3.) This adds MaxActiveStreams to the User Policy.  Requiring a db update (included) and client/sdk updates (below)
4.) This depends on the ability to update the value through a UI/client which I've done for [jellyfin-web here](https://github.com/Migs3/jellyfin-web/commit/50a00aea169f0a45ec2b2b40b0554c1b4d79bd7a):  but other languages will also need updates (Sorry, I only know English) and other clients will need updated as well to show the properly display the setting in their UI.
5.) This also then depends on [updating the sdk](https://github.com/Migs3/jellyfin-sdk-typescript/commit/4173bc44fe2966625d5b67cbf6610daac4b2be95).
6.) I would assume that this number should not exceed the MaxActiveSessions number, but I haven't implemented any checks to ensure it doesn't.  Not sure if this is needed?

**Issues**
This addresses functionality request here: https://features.jellyfin.org/posts/1444/limit-the-number-of-simultaneous-streams-per-user

### DIFF
--- a/Emby.Server.Implementations/Session/SessionManager.cs
+++ b/Emby.Server.Implementations/Session/SessionManager.cs
@@ -721,6 +721,36 @@ namespace Emby.Server.Implementations.Session
             {
                 foreach (var user in users)
                 {
+                    var activeStreams = Sessions
+                        .Where(i => i.UserId.Equals(user.Id)
+                            && i.DeviceId != session.DeviceId
+                            && i.NowPlayingItem is not null)
+                        .ToList();
+                    int maxActiveStreams = user.MaxActiveStreams;
+                    _logger.LogInformation("Current/Max streams for user {User}: {Streams}/{Max}", user.Username, activeStreams.Count(), maxActiveStreams);
+                    while (maxActiveStreams >= 1 && activeStreams.Count() >= maxActiveStreams)
+                    {
+                        try
+                        {
+                            var activeStream = activeStreams.First();
+                            await SendPlaystateCommand(
+                                activeStream.Id,
+                                activeStream.Id,
+                                new PlaystateRequest()
+                                {
+                                    Command = PlaystateCommand.Stop,
+                                    ControllingUserId = activeStream.UserId.ToString(),
+                                    SeekPositionTicks = activeStream.PlayState?.PositionTicks
+                                },
+                                CancellationToken.None).ConfigureAwait(true);
+                            activeStreams.RemoveAt(0);
+                        }
+                        catch (Exception ex)
+                        {
+                            _logger.LogDebug(ex, "Error calling SendPlaystateCommand for stopping max active session {Session}.", session.Id);
+                        }
+                    }
+
                     OnPlaybackStart(user, libraryItem);
                 }
             }
--- a/Emby.Server.Implementations/Session/SessionManager.cs
+++ b/Emby.Server.Implementations/Session/SessionManager.cs
@@ -726,15 +726,19 @@ namespace Emby.Server.Implementations.Session
                             && i.DeviceId != session.DeviceId
                             && i.NowPlayingItem is not null)
                         .ToList();
+
                     int maxActiveStreams = user.MaxActiveStreams;
+
                     _logger.LogInformation("Current/Max streams for user {User}: {Streams}/{Max}", user.Username, activeStreams.Count(), maxActiveStreams);
+
                     while (maxActiveStreams >= 1 && activeStreams.Count() >= maxActiveStreams)
                     {
+                        var activeStream = activeStreams.First();
+
                         try
                         {
-                            var activeStream = activeStreams.First();
                             await SendPlaystateCommand(
-                                activeStream.Id,
+                                session.Id,
                                 activeStream.Id,
                                 new PlaystateRequest()
                                 {
@@ -743,12 +747,13 @@ namespace Emby.Server.Implementations.Session
                                     SeekPositionTicks = activeStream.PlayState?.PositionTicks
                                 },
                                 CancellationToken.None).ConfigureAwait(true);
-                            activeStreams.RemoveAt(0);
                         }
                         catch (Exception ex)
                         {
-                            _logger.LogDebug(ex, "Error calling SendPlaystateCommand for stopping max active session {Session}.", session.Id);
+                            _logger.LogDebug(ex, "Error calling SendPlaystateCommand for stopping max active session {Session}.", activeStream.Id);
                         }
+
+                        activeStreams.RemoveAt(0);
                     }
 
                     OnPlaybackStart(user, libraryItem);
--- a/Jellyfin.Data/Entities/User.cs
+++ b/Jellyfin.Data/Entities/User.cs
@@ -158,6 +158,11 @@ namespace Jellyfin.Data.Entities
         public int MaxActiveSessions { get; set; }
 
         /// <summary>
+        /// Gets or sets the maximum number of active streams the user can have at once.
+        /// </summary>
+        public int MaxActiveStreams { get; set; }
+
+        /// <summary>
         /// Gets or sets the subtitle mode.
         /// </summary>
         /// <remarks>
--- a/Jellyfin.Server.Implementations/Migrations/20241017013200_AddMaxActiveStreams.Designer.cs
+++ b/Jellyfin.Server.Implementations/Migrations/20241017013200_AddMaxActiveStreams.Designer.cs
@@ -3,6 +3,7 @@ using System;
 using Jellyfin.Server.Implementations;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Migrations;
 using Microsoft.EntityFrameworkCore.Storage.ValueConversion;
 
 #nullable disable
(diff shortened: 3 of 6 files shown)


Migs3:
A couple of other notes:
I'm not sure if this is the "best" or "right" way to do this, I just figured this is a good place to do it where all the required pieces already exist (declarations of Sessions, Users, etc.) and Session Manager kind of manages sessions, so yeah...
Maybe it should have it's own async task to be called, instead of all this code being included directly inside of OnPlaybackStarted, Didn't really consider the "cleanliness" of the code at all.  Sorry, I'm a bit of an amateur when it comes to that kind of stuff.

Just figured I'd get the ball rolling on this feature.

JPVenson:
In general we already have that kind of limit with the max user sessions. Why limiting streams specificly?

Migs3:
> In general we already have that kind of limit with the max user sessions. Why limiting streams specificly?

The session limit is fine but this adds additional functionality.  This reduces administrative overhead of having to re-enable accounts when people try to use too many devices or they burn up all your bandwidth because they shared their account with everyone and their dog. You don't need to "up their limit" of sessions just to allow them to use another device, they have a set limit on streams and they can do so on whatever device(s) they want, but the server will limit their stream count no matter how many devices they are using.

A sessions limit only stops them from logging into more than X amount of devices at a time.  Which, yes, effectively limits their ability to stream to that same number... However, a streams limit let's them log into as many devices as they want (up to the sessions limit, if set) but then have a set limit on how many streams they can have running at any given time.

This further limits the ability of people sharing accounts as they won't want to share the account if they can't stream their own content if the person/people they share it with is also streaming.

It also allows for someone that has X number of devices to use them all as you can increase the session limit without the concern of enabling them to have additional streams. 
They then can use all devices at once without locking out their user account if you set a limit and they try to log into another device.

This also allows for a user to use quick connect functionality without enabling them to stream more than one stream at a time.  Currently if you set them to 1 session, they effectively cannot use quick connect to setup a second device.  Attempting to do so and you get a can't log in error because they are limited to just a single session, this will also lock their account if they attempt it too many times.  And given some people's inability to understand why this happens this can easily be avoided by setting a stream limit instead of a session limit.  Not to mention, the quick connect error message is not clear when you attempt it on a new device and you're at your limit.

Emby.Server.Implementations/Session/SessionManager.cs
```diff
@@ -721,6 +721,36 @@ public async Task OnPlaybackStart(PlaybackStartInfo info)
...
+                                    Command = PlaystateCommand.Stop,
+                                    ControllingUserId = activeStream.UserId.ToString(),
+                                    SeekPositionTicks = activeStream.PlayState?.PositionTicks
+                                },
+                                CancellationToken.None).ConfigureAwait(true);
+                            activeStreams.RemoveAt(0);
+                        }
+                        catch (Exception ex)
+                        {
+                            _logger.LogDebug(ex, "Error calling SendPlaystateCommand for stopping max active session {Session}.", session.Id);
```
Migs3, line 750: Just realized this is wrong (copy/pasta got me here), for two reasons:
1.) This is using the sessionId of the new session instead of the one it's attempting stopping
2.) Two this will not work if I update it to use activeStream.Id, which is the correct one, as currently activeStream is currently declared inside the try-block, so I don't think the catch-block will be able to access it.

To capture this and have it flow correctly, I'd likely have to move the declaration of activeStream to right before the try-block and possibly move the RemoveAt() to after the end of catch-block.

Not completely sure about the latter, but I think from a while-looping perspective, this might create the possibility of an infinite loop as it might not be able to ever remove any items from activeStreams if the try-block fails before it gets to the RemoveAt().

Migs3: Fixed this and other minor issues.

gnattu:
> You don't need to "up their limit" of sessions just to allow them to use another device, they have a set limit on streams and they can do so on whatever device(s) they want, but the server will limit their stream count no matter how many devices they are using.

If this is the intended use case then what you implemented cannot do that. The client can just ignore the stop playback request and just keep playing current media just fine

Migs3:
> > You don't need to "up their limit" of sessions just to allow them to use another device, they have a set limit on streams and they can do so on whatever device(s) they want, but the server will limit their stream count no matter how many devices they are using.
> 
> If this is the intended use case then what you implemented cannot do that. The client can just ignore the stop playback request and just keep playing current media just fine

Not entirely sure what you're getting at here, the same thing could be said for anything you expose to the internet.  There are ways around everything.  If someone really wants to get in, they will get in, nothing is fail proof from a security perspective, breaches happen all the time.  I'm hardly concerned with someone maliciously writing clients to ignore server issued commands.  If that happens, then I remove them from having access to my server(s) and that's that.

For the purpose of adding this feature to function, the clients I have tested with (jellyfin-web and Android), it does work as intended.  I was unable to play a stream on more than 1 device when the limit was set to 1.  (2 devices when set to 2, etc.)

bonswouar:
I don't know if the limits of this PR can be fixed _(ideally it shouldn't impact the api/sdk I guess - but I'm not sure it's doable in Jellyfin's current state)_, but this feature would be really useful to limit server overload even if it's not perfect!
As @Migs3 pointed out, most users will want to have more than 1 session (one per device basically), but more than 1 (or X) stream per user can be problematic.

nebb00:
Really need this functionality to migrate from Plex

Migs3:
> I don't know if the limits of this PR can be fixed _(ideally it ~shouldn't impact~ should have minimum impact on the api/sdk I guess - just a new error code maybe - but I'm not sure it's doable in Jellyfin's current state)_, but this feature would be really useful to limit server overload even if it's not perfect! As @Migs3 pointed out, most users will want to have more than 1 session (one per device basically - I'm not even sure why would anyone want to limit this), but more than 1 (or X) stream per user can be problematic.

To limit sdk changes, we could just repurpose the current max active session value.  Just remove the current implementation where it blocks the login session and then just use the value for this PR implementation instead, this would allow any number of login sessions but limit the streams possible still.  This isn't an ideal solution though as the hole pointed out above does still exist. (the ability for the client to disregard the stop command issued)

In my opinion, there should be 2 separate configurable options.  The first would limit the devices per user (new) and then the second would limit the number of active streams. (this PR) 

What can be added in is the ability for the user to see their own devices that are attached to their accounts.  Then also allow them to manage their own devices.

This way it's easier to understand for the user no matter their technical level of understanding.  It then becomes very much like other streaming services or cable services even..  You(as the administrator) set a certain limit on their device count, and the user can then go into their account settings/dashboard and click devices and manage their devices and see the total/limit count.

This would not only impose the device limit, but also offload the management of the devices they authorize.  It would also give them the ability to add/remove devices that they no longer own (like when switching phones or tvs or whatever) which again is much like some other major streaming services.

Just my two cents on how it currently works vs how it should work.

sambartik:
> Not entirely sure what you're getting at here, the same thing could be said for anything you expose to the internet. There are ways around everything. If someone really wants to get in, they will get in, nothing is fail proof from a security perspective, breaches happen all the time.

This argument is just a logical fallacy. Just because everything is vulnerable doesn't mean security should be dismissed altogether.

Client-side security is no security at all.

In my own opinion, Jellyfin tries to move from the messy ways of Emby to a more decent codebase. This would just add to the clutter. This needs to be implemented properly.

sjorge:
Perhaps this needs a proper discussion thread over at https://github.com/jellyfin/jellyfin-meta/discussions

Migs3:
> This is one of the most requested feature.
> 
> https://features.jellyfin.org/posts/59/account-based-stream-limit
> 
> One of the comments below. Please read the comments in the above link.
> 
> ```
> I think this still needs improvement, as this adds unnecessary hassle for the user. I don't particularly care how many devices the user is logged into at the same time, only how many streams the user is running.
> Consider the following scenario with session limit set to 1: two TVs (amazon fire sticks in my case), the user is currently watching something on TV 1. To move to TV 2, the user currently has to sign out of the app on TV 1, then log into TV 2 before starting to watch on TV 2.
> This is an unnecessary hassle for the user. Much better would be if the user only has to stop the stream on TV 1.
> ```
> 
> This is the reason why the below feature was created to speak explicitly. The is also another duplicate with 12-15 upvotes.
> 
> https://features.jellyfin.org/posts/1444/limit-the-number-of-simultaneous-streams-per-user

I stopped looking at this based on the negative feedback on what I've proposed.  Instead of providing any positive/constructive feedback, it's "this won't work", "we already have this" (which we don't) and seemingly an overall "you're doing it wrong".  Didn't exactly provide the warm fuzzy that made me want to bother to look into it any further.

JPVenson:
@Migs3 Please dont missunderstand this, i dont think anyone would oppose the feature its just the way its currently implemented that does not sit well.

Usually for features such as this its always a good idea to first propose a design at the meta repository.

PR #12847, opened 2024-10-19
Fix seeking beyond EOF in HWA transcoding

nyanmisaka:
**Changes**
- Fix seeking beyond EOF in HWA transcoding

**Issues**
- Seeking beyond EOF makes no sense in transcoding. Clamp the seekTick value to
  [0, RuntimeTicks - 0.5s], so that the muxer gets packets and avoid error codes.


gnattu (approved):

Shadowghost (approved):

nielsvanvelzen (approved):
Can confirm this fixes my playback issues

crobibero (approved):

Bond-009 (approved):

Issue #12835, opened 2024-10-16
New created playlist are public by default


MaverickPOL:
### This issue respects the following points:

- [X] This is a **bug**, not a question or a configuration issue; Please visit our forum or chat rooms first to troubleshoot with volunteers, before creating a report. The links can be found [here](https://jellyfin.org/contact/).
- [X] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [X] I'm using an up to date version of Jellyfin Server stable, unstable or master; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.
- [X] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [X] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Description of the bug

In new version of the Jellyfin i saw that new created playlists are public by default. User must choose to make playlist private.  And in description of 10.9.x version there was information that those playlist should be private by default (https://jellyfin.org/posts/jellyfin-release-10.9.0/)

This is not that issue when i use WEB GUI or Jellyfin app, but when user use for example Finamp, he dont have option to choose that playlist should be private or not, and if playlist by default are public, all created playlist become public.

Could we have option to choose how playlist should be create by default (without user interaction), or just have private by default?
So right now i believe this is bug, becouse in patchnotes was information that playlist should be private by default.

It will be very nice also if there will be option to change public/private for playlist. Maby even from admin page for all users?

### Reproduction steps

Click Create New Playlist for audio file

### What is the current _bug_ behavior?

New created playlist are by default public.

### What is the expected _correct_ behavior?

Correct behaviour should be "private playlist by default without user interaction"

### Jellyfin Server version

10.9.11+

### Specify the build version

10.9.11

### Environment

```markdown
- OS: Debian
- Linux Kernel:  n/a
- Virtualization: LXC
- Clients: jellyfin, finamp
- Browser: chrome
- FFmpeg Version: n/a
- Playback Method: n/a
- Hardware Acceleration: n/a
- GPU Model: n/a
- Plugins: none
- Reverse Proxy: nginx
```

### Jellyfin logs

```shell
[2024-10-17 00:17:20.445 +02:00] [INF] Creating image collage and saving to "/var/cache/jellyfin/temp/3537978bedb841f8a521bf83584d14df.png"
[...]

PR #12853, opened 2024-10-20
make playlist creation private by default

dkanada:
@nielsvanvelzen you requested the current behavior but it would change the status quo for existing clients.

**Changes**

Invert public boolean for backwards compatibility on clients that don't currently support this attribute. Currently untested so don't merge this unless it looks correct.

**Issues**

#12835


nielsvanvelzen (approved):
Pre 10.9 behavior would always create public playlists. To allow clients to adapt we kept that behavior with 10.9.
It should be fine to change it to private by default with 10.11 as that would've given client developers 2 releases / 1 year to migrate.

Shadowghost (approved):

Bond-009 (approved):

Issue #12101, opened 2024-06-15
[Issue]: Sort by year bug


Liquoriced:
### Please describe your bug

When sorting movies by year, I noticed that not all movies are sorted.
This does not happen with tv series, at least not in my experience.

As you can see in the image below, when sorting by release year in ascending order, the first 11 movies are from later years. The same happens when sorting in descending order (the incorrect movies are at the end, instead of the beginning).
The metadata is correct, and no spaces are present.

### Reproduction Steps

1. Sort movies by year
2. Not all movies are sorted correctly

### Jellyfin Version

10.9.0

### Environment

```markdown
OS: Linux
Virtualization: Docker
Clients: Browser and Android
Browser: Firefox and Chrome
Jellyfin Version: 10.9.6
Playback: Direct Play
Reverse Proxy: none
Storage: local NAS
```

### Jellyfin logs

```shell
nothing relevant
```

### Please attach any screenshots here

![image](https://github.com/jellyfin/jellyfin/assets/80312124/04a69aa1-7673-4fe8-97a3-87bd74cb4519)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

BrianCArnold:
I was able to replicate this, the Release Date sort actually doesn't sort on "Year", but "Release Date".

This movie showed up at the beginning of my list of movies when I sorted by "Release Date", and after I added the specific release date of the movie to the metadata, it was sorted correctly: 
![image](https://github.com/jellyfin/jellyfin/assets/38593283/c7d50f45-c864-470f-b277-78577c558692)

felix920506:
Multiple reports, marking as confirmed.

Liquoriced:
Maybe there should be a fallback process when the specific release date is blank, like going by year.

honestlywhoknows:
Wrote some unit tests to evaluate the sorting function, it seems as though it already has the functionality @Liquoriced is suggesting. In the case of a movie not having a Premiere Date, it will use 01-01-{ProductionYear} to compare, if available. If neither date is available, it will return the minimum date (meaning the movie would be listed as if it came out the earliest). 
<img width="659" alt="screenshot" src="https://github.com/jellyfin/jellyfin/assets/55208637/b815e68f-0aac-4c57-a073-d206de4b9885">

@Liquoriced can you please confirm whether the Release Date and Year values are populated in the metadata for the movies that are not sorting correctly?

Liquoriced:
[...]

PR #12798, opened 2024-10-09
Refactored library.db to be moved into jellyfin.db

JPVenson:
Migrated all data from old library.db into new tables inside jellyfin.db

**Changes**
- Migrated all data into new schema
- Refactored IItemRepository and extracted IPeopleRepository, IMediaStreamRepository, IUserDataRepository, IChapterRepository
- Created new Table BaseItemProvider to expand the aggregate column of ItemProvider in TypedBaseItem
- Added forginKeys where applicable

**Issues**
All of them

PR #12859, opened 2024-10-21
Order using production year when premiere date is not available

yutongluo:
**Changes**

Add coalesce clause to use production year to sort when premiere date is not available. Behavior is as follows
- Use Premiere Date if not null
- else use `date(ProductionYear||'-01-01')`
- if `ProductionYear` is also null, date will return null since '-01-01' is invalid, and sort will be using NULL (expected behavior)

Could not find where to add a test, if folks more familiar could help point it out that'd be great. 

**Issues**

#12101


JPVenson:
hey this is blocked by our EFCore rewrite
https://github.com/jellyfin/jellyfin/pull/12798
Please wait until that is finished.

theguymadmax:
Superseded by #13733

PR #12615, opened 2024-09-08
Rework parental ratings

Shadowghost:
**Changes**
* Migrate rating system definitions to JSON:
  * SupportsSubscores
  * Scores
    * RatingStrings
    * Score
    * Sub score
* Add `InheritedParentalRatingSubValue` to `TypedBaseItem` table
* Adopt existing migration and re-run (changed GUID)

**ToDo**
* [x] Migrate `MaxParentalAgeRating` in user table
* [x] Adopt `BaseItem.IsParentalAllowed` afterwards
* [x] Adopt SQL queries to only check sub rating if server rating system uses sub ratings too

**Comments**
* Please re-save you user's parental rating limits to properly populate them in the database (we can't infer this automatically)

**Issues**
Fixes #11650
Fixes #13077
Fixes #13507

PR #12867, opened 2024-10-23
Move TV-PG ratings to be in line with PG rating.

Dessyreqt:
I'd like to revert the TV-PG ratings to be in line with the PG ratings. I ran into this issue earlier while trying to manage parental controls for my 10-year-old; there are several TV-PG shows I'm fine with him watching, and he would like to watch, but I don't want to enable some of the PG-13 movies at the same time. 

Under the current version, TV-PG and all of its subcategories are lumped together with PG-13, which doesn't make much sense considering there is a TV-14 rating, the rest of the rating scale isn't so granular as to provide ratings in one-year increments, and according to the [Rating System Wiki](https://rating-system.fandom.com/wiki/TV-PG), the rating is designed to be equivalent to the MPA PG rating.

Hopefully this is an easy change that makes sense!


Shadowghost (approved):
Fine with me. Keep in mind though that just changing the mapping here will not change the actual filtered value in Jellyfin. You'd need to re-enable the corresponding migration again (just change the UUID)

Dessyreqt:
I've updated the UUID, please let me know if there is anything else that needs to be done.

1337joe (approved):

Bond-009 (approved):

Shadowghost:
If we don't merge this to the release branch, it'll likley be superseded by https://github.com/jellyfin/jellyfin/pull/12615

PR #12877, opened 2024-10-25
[WIP] Smart font streaming for external ASS subtitles

p0358:
This is new feature based on discussion under #12511 that proposed this general idea instead.

**Background/reasoning** (as short and concise as I could :sob:)

Currently there is no way to stream fonts from server for loose .ass files next to the video files. The fonts either need to be (a) bundled inside of .mkv or (b) pre-installed on the clients. Some people (c) misused the fallback fonts feature to make the web client load up the fonts they wanted, abusing that it'd preload them all to the browser.

These existing options sucked because:
- (a) it's not desirable to re-bake the existing MKV file, as that both requires work + you don't keep original for example seed such original file anymore, or for archival reasons
- (b) impossible on TVs, phones, consoles, even on web client, so basically possible just for PC clients
- (c) fallback fonts preload ALL of the fonts there up to 20 MB limit, this causes huge unnecessary strain on the browser if used this way, plus delays subtitle viewing by a good couple of seconds on LAN!

I attempted a PR #12511, but it'd break existing original use-case of fallback fonts, plus it wasn't desirable by the contributors to send full list of fonts (could be many of them), plus the naive approach of mapping just font name from .ass to font file could NOT actually work reliably due to more complicated logic of font-picking, as used by libass for example (most popular library).

It was concluded that server should decide and filter what fonts should be loaded for a given subtitles file. That approach makes perfect sense, since the server is then just one place that has to parse the file and run selection logic, while clients can then just load the fonts. Then the fonts should all be preloaded into video backend and the embedded library such as libass will match them up with its extended logic.

The source of fonts will be server-side fonts installed on the system. This creates a common single source of extra fonts for videos, since those fonts are already used for server-side subtitle burn-in automatically by ffmpeg now, plus there's now instructions how to install extra fonts in Docker in docs now. Alternative of adding 4th source of fonts as a streamable folder similar to fallback fonts would be bad on the other hand.

**Implementation**

My implementation idea boils down to just using libass (modified and wrapped into C# module) to parse and select a list of needed fonts.

Why libass:
- guarantees proper and consistent parsing of .ass file
- exact same consistent logic for matching up fonts as on the client, meaning the client-side will be happy with the fonts it gets (and the logic isn't as simple as name -> file turns out)
- libass happens to already have integration of native system font provider APIs of three major systems: Linux (fontconfig), Windows (directwrite), macOS (CoreText), this means no need to reinvent the wheel for us all while taking advantage of system's existing font cache to make the picking faster

I spent actually a lot of time and effort to get a minimal version of libass built (without rendering features and thus without having to link some unneeded libs). It works pretty well now. Two levels of caching (fontselect caches font info, I cache params->font info mapping) make the thing very reasonably fast. The advantage of this approach is that it also doesn't require advanced changes in Jellyfin codebase.

Also getting just a list of needed fonts instead of a map (that was flawed anyways, as Subtitle Octopus would just simply download all fonts it saw in Styles with no further checks!) will allow many clients to very easily support it. If you have clients that expect a dir with fonts and you already create a temp dir and download MediaAttachments fonts there, it should be trivial for you to also download these fonts there too. Peace of cake, and this will allow the adoption to be more than just the web client...

⚠️ [**WIP part**]: I still need to figure out CI building of the modified libass and packaging it into NuGet... I named my prototype package **`Jellyfin.LibAssTools`**, code-wise it works already but I still need to figure out some things like logging redirection maybe. It's not a full libass wrapper by any means, wraps just the interface created for Jellyfin's needed functionality to remain minimalistic.

And so I think we should move it under Jellyfin org once it's ready, similarly to a helper module like: https://github.com/jellyfin/Jellyfin.XmlTv

**Changes**

1. Adds following API endpoints:
- `Videos/{routeItemId}/{routeMediaSourceId}/Subtitles/{routeIndex}/{routeStartPositionTicks}/Fonts` -> get FontStreamInfo[]
- `Videos/{routeItemId}/{routeMediaSourceId}/Subtitles/{routeIndex}/Fonts` -> get FontStreamInfo[]
- `Fonts/{family},{bold},{italic}` -> get font file (the key to use as params will be returned in items from above^)
- `Fonts/{family}` -> get font file (easy version of above, less precise)

on `{routeStartPositionTicks}`:
The first two routes have two variants exactly like the one ending with `/Stream.ass`. Right now the variant with `{routeStartPositionTicks}` is kinda useless, but: for one, it can be used more consistently since for font /Stream.ass always the longer version is used; secondly, if someone ever implements it properly in the future that events before start tick are filtered out even when subtitles are not converted between formats (right now they're streamed directly if same format, converted with dropping otherwise). And if that was to happen, the library would already filter out the fonts that are only used by the dropped events. So that's kinda future-proofing it, since the intended interface is already there I guess?

2. Adds following schema:
- `MediaBrowser.Model/Subtitles/FontStreamInfo`

⚠️ **TODO**

Todo:
- module `Jellyfin.LibAssTools` is missing currently!!! (see above)
- PR to support it in jellyfin-web at least
- PRs to support it in some other clients (I'll look into some of them after we merge the baseline of server+web I think)

Questions:
- should the link to /Fonts list be next to `DeliveryUrl` that's consumed by clients, or should the client take DeliveryUrl and figure out path to /Fonts endpoint on its own (with regex?)? clearly at least having it there would make it less error-prone for clients plus easier to feature-detect it's not yet supported on old server; but that'd require adding a new field to class MediaStream if that's okay (like `FontsUrl`)
- is API endpoint prefix `Fonts/...` for the last two endpoints ok? Fallback fonts have `FallbackFont/Fonts/`, and my logic was "okay, and these are *just* fonts". But some alternatives could include something like `StreamableFont/Fonts/`, `ServerFont/Fonts/`, `ServerSideFont/Fonts/` idk!!
- should enablement of this be configurable (on server?)? if yes then enabled or disabled by default?
- should some API feedback also be returned for fonts that couldn't be found installed on the server?

Possible enhancements:
- if the above with `FontsUrl` on MediaStream is answered "yes", then server could maybe list fonts in advance and only add the `FontsUrl` or something if there's >0 fonts available, so that the client wouldn't have to bother with round-trip to `.../Fonts`
- currently fonts existing in MediaAttachments of MKV aren't filtered out. It's TBD whether clients should attempt loading external fonts if they stream non-external subtitles. Simple option for now is to not do that and assume that .ass in MKV will have corresponding font files attached all in the MKV. While also assuming that an .ass outside will likely not and hence the new feature.
- maybe the last two endpoints could be combined into one with `bold` and `italic` as optional query params...

Problems:
- ideally we'd want a way on the SubtitleEncoder interface to grab either ssa/ass if available, else bail out; currently something like "ssa" will probably be converted to "ass" and throw out Styles; and something like .srt would be converted to generic .ass and then it'd stream Default font from converter (Arial or whatever)
- I need to wait for libass contribs to answer me point 2 in libass/libass#843 so that I know if I can simplify Bold and Italic parameters into booleans or not... (I already realized I don't need Vertical right before making the PR)

**Issues**

Supersedes these old PRs that were based on the old idea:
Closes #12511
Closes jellyfin/jellyfin-web#5960

So yeah this isn't finished yet, but I wanna start gathering some feedback already if people have some. Working on it slowly while Jellyfin is in its lock phase before release. Maybe we could get this in after lifting the lock and at least there'd be plenty of time for interested clients to start supporting it then. Hope I didn't type out too much text :sob:

### DIFF
--- a/Jellyfin.Api/Controllers/SubtitleController.cs
+++ b/Jellyfin.Api/Controllers/SubtitleController.cs
@@ -14,6 +14,7 @@ using Jellyfin.Api.Attributes;
 using Jellyfin.Api.Extensions;
 using Jellyfin.Api.Helpers;
 using Jellyfin.Api.Models.SubtitleDtos;
+using Jellyfin.LibAssUtils;
 using MediaBrowser.Common.Api;
 using MediaBrowser.Common.Configuration;
 using MediaBrowser.Controller.Configuration;
@@ -326,6 +327,74 @@ public class SubtitleController : BaseJellyfinApiController
     }
 
     /// <summary>
+    /// Gets a list of fonts for given ASS subtitles file.
+    /// </summary>
+    /// <param name="routeItemId">The (route) item id.</param>
+    /// <param name="routeMediaSourceId">The (route) media source id.</param>
+    /// <param name="routeIndex">The (route) subtitle stream index.</param>
+    /// <response code="200">File returned.</response>
+    /// <returns>A <see cref="FileContentResult"/> with the subtitle file.</returns>
+    [HttpGet("Videos/{routeItemId}/{routeMediaSourceId}/Subtitles/{routeIndex}/Fonts")]
+    [Authorize]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    public IAsyncEnumerable<FontStreamInfo> GetSubtitleFonts(
+        [FromRoute, Required] Guid routeItemId,
+        [FromRoute, Required] string routeMediaSourceId,
+        [FromRoute, Required] int routeIndex)
+    {
+        return GetSubtitleFontsWithTicks(routeItemId, routeMediaSourceId, routeIndex, 0);
+    }
+
+    /// <summary>
+    /// Gets a list of fonts for given ASS subtitles file.
+    /// </summary>
+    /// <param name="routeItemId">The (route) item id.</param>
+    /// <param name="routeMediaSourceId">The (route) media source id.</param>
+    /// <param name="routeIndex">The (route) subtitle stream index.</param>
+    /// <param name="routeStartPositionTicks">The (route) start position of the subtitle in ticks.</param>
+    /// <response code="200">File returned.</response>
+    /// <returns>A <see cref="FileContentResult"/> with the subtitle file.</returns>
+    [HttpGet("Videos/{routeItemId}/{routeMediaSourceId}/Subtitles/{routeIndex}/{routeStartPositionTicks}/Fonts")]
+    [Authorize]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    public async IAsyncEnumerable<FontStreamInfo> GetSubtitleFontsWithTicks(
+        [FromRoute, Required] Guid routeItemId,
+        [FromRoute, Required] string routeMediaSourceId,
+        [FromRoute, Required] int routeIndex,
+        [FromRoute, Required] long routeStartPositionTicks)
+    {
+        // return File(,MimeTypes.GetMimeType("file." + format));
+        var stream = await EncodeSubtitles(
+                routeItemId,
+                routeMediaSourceId,
+                routeIndex,
+                "ass", // TODO: problem: "ssa" subtitles will probably be converted instead of grabbed as-is, and that'll throw out Styles
+                routeStartPositionTicks,
+                null,
+                false).ConfigureAwait(false);
+
+        var fontInfos = LibAssUtils.Fonts.GetFontsForSubtitlesBuffer(stream);
+
+        foreach (var fontInfo in fontInfos)
+        {
+            if (fontInfo.path is null)
+            {
+                continue;
+            }
+
+            yield return new FontStreamInfo
+            {
+                Path = fontInfo.path,
+                Key = string.Format(CultureInfo.InvariantCulture, "{0},{1},{2}", fontInfo.font_name, fontInfo.bold, fontInfo.italic),
+                FontName = fontInfo.font_name,
+                PostScriptName = fontInfo.postscript_name,
+                Bold = fontInfo.bold,
+                Italic = fontInfo.italic,
+            };
+        }
+    }
+
+    /// <summary>
     /// Gets an HLS subtitle playlist.
     /// </summary>
     /// <param name="itemId">The item id.</param>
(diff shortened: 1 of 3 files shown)


p0358:
Slowly realizing the extra scale of actual complications to parse ASS correctly, but being glad I took this approach. So I didn't so far consider that Events in ASS can contain override tags, which can override font style (name, boldness, italics, verticality), which should trigger new font lookup. Means I'll have to rewrite my whole pickup logic after getting the initial prototype fully working.

Technically every next tag in Event adds a new combination for font lookup, unless there's no text between a series of tags. Including "no change from base Style" combination if there's some text before any font style override tags were used. This is gonna be interesting :)

vuhe:
If the font size is too large, can a subset font function be added on the server side, even if CJK fonts usually have a file size of no more than 2M after processing. Embed fonts into ASS files when needed by the client, and the ASS subtitle files with embedded fonts can handle most scenarios that require fonts after processing. I also tested Jellyfin's web player and they worked very well.

p0358:
Not sure how hard would it be to compute the needed subset, but the idea of embedding the font into the ASS file on the fly is interesting, technically this could mean easier ability of streaming the fonts without the given client needing to download the fonts separately from the API and then providing them explicitly to the player. On the other hand being able to cache such fonts (at least if full font was streamed instead of subset!) could allow less network bandwidth, as I imagine usually one show/season would have the same/similar set of fonts if subtitles are from the same source, which would still speak in favor of the separate font download path being kept in I suppose...

I should definitely resume work on this PR eh (big chunk of the libass part was already done, but I should still do a dry run through the file to properly figure out all needed fonts, plus to get it to compile on all platforms to create a NuGet package that works on all platforms). I think whether the fonts are streamed separately or embedded into ASS should be a rather small change in all of that.

One risk of the subset approach is that if we get it any wrong in any capacity, then there will be missing glyphs or glyphs loaded from fallback font, which might be ugly. How likely that could be, I don't know at this point.

Would have to look whether the library that libass uses to check the detailed characteristics of fonts could also be used to compose subset of fonts based only on the queried glyphs, idk. Could be potentially a big win in the edge case where an ASS file uses a lot of different CJK fonts, but potentially not that much of a win for the work in average scenario. I would probably want to put it behind the initial feature set to not scope creep before it all works.

vuhe:
I agree with your point of view that calculating the font required for ASS seems to be a very tedious task.

I proposed the idea of subsets because I downloaded fonts over 58GB to prevent missing fonts, and individual files of some fonts can easily exceed 20MB.

I think Jellyfin developers probably don't want clients to save large font files as cache.

Although the fonts used in each season may be similar, if there are many seasons in the library and the fonts used in each season are different, it may take up a lot of space.

Anyway, calculating the required font and sending it is a very good improvement, thank you for your effort.

jwood00:
It's been a while since the last update, but has any further work been done on this? Handling external .ASS subtitles has been one of my biggest pain points, so I find this method of streaming them to the client instead of abusing the backup fonts directory to be quite interesting.

p0358:
I've burned out on it a bit once the novelty wore off, I should try to revisit it after my holidays. I don't believe anything has changed since I worked on it, and it's a shame I didn't finish back then since I was almost done with it.

PR #12880, opened 2024-10-26
Added Setup overlay app to communicate status of startup

JPVenson:
Adds a temporary aspcore pipeline to ensure the server will respond to requests even while a migration or a long running setup is still in progress.

The temp service will be stopped and removed as soon as the application server is ready to be started.

This fixes the issue of long running migrations triggering a watchdog and stopping the server externally.

Provides an local network only accessable alternative endpoint to obtain the current logfile to debug and progress.


Shadowghost (approved):

Jellyfin.Server/ServerSetupApp/SetupServer.cs
```diff
@@ -0,0 +1,171 @@
...
+                                            while (jfApplicationHost is null && retryCounter < 5)
+                                            {
+                                                await Task.Delay(500).ConfigureAwait(false);
+                                                jfApplicationHost = serverApplicationHost();
+                                                retryCounter++;
+                                            }
+
+                                            if (jfApplicationHost is null)
+                                            {
+                                                context.Response.StatusCode = (int)HttpStatusCode.ServiceUnavailable;
```
crobibero, line 94: Most SDKs / clients won't parse the response if the status isn't success

Jellyfin.Server/Program.cs
```diff
@@ -42,6 +45,9 @@ public static class Program
         public const string LoggingConfigFileSystem = "logging.json";
 
         private static readonly SerilogLoggerFactory _loggerFactory = new SerilogLoggerFactory();
+        private static SetupServer? _setupServer = new();
```
crobibero, line 48: No reason to have this be nullable

Jellyfin.Server/Program.cs
```diff
@@ -42,6 +45,9 @@ public static class Program
...
+        private static CoreAppHost? _appHost;
+        private static IHost? _jfHost = null;
```
crobibero, line 50:
```suggestion
        private static IHost? _jellyfinHost = null;
```

Jellyfin >> jf

Jellyfin.Server/ServerSetupApp/SetupServer.cs
```diff
@@ -0,0 +1,171 @@
...
+using MediaBrowser.Controller;
+using MediaBrowser.Model.System;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Diagnostics.HealthChecks;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+using Microsoft.Extensions.Hosting;
+using SQLitePCL;
```
crobibero, line 19: Likely some unused using statements here

Jellyfin.Server/ServerSetupApp/SetupServer.cs
```diff
@@ -0,0 +1,171 @@
...
+                                        loggerRoute.Run(async context =>
+                                        {
+                                            var networkManager = networkManagerFactory();
+                                            if (context.Connection.RemoteIpAddress is null || networkManager is null || !networkManager.IsInLocalNetwork(context.Connection.RemoteIpAddress))
+                                            {
+                                                context.Response.StatusCode = (int)HttpStatusCode.Unauthorized;
+                                                return;
+                                            }
+
+                                            var logfilePath = Directory.EnumerateFiles(applicationPaths.LogDirectoryPath).Select(e => new FileInfo(e)).OrderBy(f => f.CreationTimeUtc).FirstOrDefault()?.FullName;
```
crobibero, line 70: Could simplify this as

```suggestion
                                            var logFilePath = new DirectoryInfo(applicationPaths.LogDirectoryPath)
                                                .EnumerateFiles()
                                                .OrderBy(f => f.CreationTimeUtc)
                                                .FirstOrDefault()
                                                ?.FullName;
```

crobibero (commented):
Untested, just reading thought the code

Jellyfin.Server/ServerSetupApp/SetupServer.cs
JPVenson: which is fine? i guess. 
Clients can ignore it for as long as the server is starting, that should all in all result in the same behavior as before for now until they choose to implement that specific waiting behavior.

This is mainly intended for automatic/external systems like watchdogs which will give the _option_ to handle a 503 differently then a 404

crobibero (approved):

PR #12893, opened 2024-10-27
Auto update issue template version on new release

HadrienPatte:
**Changes**

Extend the existing `bump_version` script to also update the jellyfin version in the github issue template file so if is no longer necessary to manually update it on every new release.

I tested that the script does behave as expected with the issue template file that contains a space character in its filename.

**Issues**

Context/motivation: https://github.com/jellyfin/jellyfin/pull/12613#pullrequestreview-2288536359

### DIFF
--- a/bump_version
+++ b/bump_version
@@ -28,6 +28,7 @@ jellyfin_subprojects=(
   Emby.Naming/Emby.Naming.csproj
   src/Jellyfin.Extensions/Jellyfin.Extensions.csproj
 )
+issue_template_file="./.github/ISSUE_TEMPLATE/issue report.yml"
 
 new_version="$1"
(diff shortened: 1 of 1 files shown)


Bond-009 (approved):

Shadowghost (approved):

joshuaboniface (approved):

joshuaboniface:
>I tested that the script does behave as expected with the issue template file that contains a space character in its filename.

For posterity this was not, in fact, true :sweat_smile: The filename in the `sed` command needed quotes too.

Fixed in https://github.com/jellyfin/jellyfin/commit/d5fd5dfe6a04a829306da0e19efb08492ce189da

HadrienPatte:
ah 😅 sorry about that, thanks for the fix!

PR #12904, opened 2024-10-29
Fixed possible infinite recursion on MediaSourceCount

JPVenson:
Checks for infinite recursion

**Changes**
added stack for tracing callstack

**Issues**
#12902

PR #12907, opened 2024-10-29
Fix stackoverflow in MediaSourceCount

JPVenson:
unfucked version of #12904

### DIFF
--- a/MediaBrowser.Controller/Entities/Video.cs
+++ b/MediaBrowser.Controller/Entities/Video.cs
@@ -152,16 +152,7 @@ namespace MediaBrowser.Controller.Entities
         {
             get
             {
-                if (!string.IsNullOrEmpty(PrimaryVersionId))
-                {
-                    var item = LibraryManager.GetItemById(PrimaryVersionId);
-                    if (item is Video video)
-                    {
-                        return video.MediaSourceCount;
-                    }
-                }
-
-                return LinkedAlternateVersions.Length + LocalAlternateVersions.Length + 1;
+                return GetMediaSourceCount(new Stack<Guid>());
             }
         }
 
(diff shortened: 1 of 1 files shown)


MediaBrowser.Controller/Entities/Video.cs
```diff
@@ -550,5 +541,25 @@ public virtual MediaStream GetDefaultVideoStream()
 
             return list;
         }
+
+        private int GetMediaSourceCount(Stack<Guid> callstack)
```
cvium, line 545: I somewhat see how `Stack` might be semantically suitable, but `.Contains` is a linear search. A `HashSet` or similar is a better data structure for this.

cvium (approved):

Shadowghost (approved):

crobibero (approved):

Bond-009 (changes requested):
I don't like that the incorrect data just gets ignored.
IMO a better solution would be throwing in the PrimaryVersionId setter when the Video it points to also has a PrimaryVersionId. This way the culprit (probably a plugin) can get blamed and it can be fixed instead of covered up.

JPVenson:
> Video it points to also has a PrimaryVersionId. 

Isnt the issue here, thats why its recursive, the issue is that most likely there is a video that PrimaryId is itself. I agree that this should be better handled on setter and never create bad data in the first place, but then we also need to migrate bad data as well which is more then just this quickfix.

Bond-009:
> Isnt the issue here, thats why its recursive, the issue is that most likely there is a video that PrimaryId is itself.

Even easier to check, don't even need to interact with the DB, it's basically free.

>  but then we also need to migrate bad data

If it's like you say, PrimaryId == Id a migration isn't too hard

> which is more then just this quickfix.

If this is just a "quick fix" it shouldn't be backported but fixed properly in master.

PR #12908, opened 2024-10-29
Fix search for missing metadata overriding existing metadata

scampower3:
Fixes a regression where search for missing metadata overrides existing metadata
**Changes**

**Issues**


Shadowghost (approved):
I think as it is after the last commit it should work fine. But I can't test it right now.

Shadowghost:
@scampower3 might make sense to PR this against master now

PR #12916, opened 2024-10-30
Added query filter to disregard disabled Providers

JPVenson:
MediaSegment's are now always filtered based on the providers they exist in.

**Changes**

**Issues**

### DIFF
--- a/Jellyfin.Server.Implementations/MediaSegments/MediaSegmentManager.cs
+++ b/Jellyfin.Server.Implementations/MediaSegments/MediaSegmentManager.cs
@@ -141,6 +141,19 @@ public class MediaSegmentManager : IMediaSegmentManager
     /// <inheritdoc />
     public async Task<IEnumerable<MediaSegmentDto>> GetSegmentsAsync(Guid itemId, IEnumerable<MediaSegmentType>? typeFilter)
     {
+        var baseItem = _libraryManager.GetItemById(itemId);
+
+        if (baseItem is null)
+        {
+            _logger.LogError("Tried to request segments for an invalid item");
+            return [];
+        }
+
+        var libraryOptions = _libraryManager.GetLibraryOptions(baseItem);
+        var providers = _segmentProviders
+            .Where(e => !libraryOptions.DisabledMediaSegmentProviders.Contains(GetProviderId(e.Name)))
+            .ToArray();
+
         using var db = await _dbProvider.CreateDbContextAsync().ConfigureAwait(false);
 
         var query = db.MediaSegments
@@ -150,6 +163,9 @@ public class MediaSegmentManager : IMediaSegmentManager
         {
             query = query.Where(e => typeFilter.Contains(e.Type));
         }
+
+        var providerIds = providers.Select(f => GetProviderId(f.Name)).ToArray();
+        query = query.Where(e => providerIds.Contains(e.SegmentProviderId));
 
         return query
             .OrderBy(e => e.StartTicks)
--- a/Jellyfin.Server.Implementations/MediaSegments/MediaSegmentManager.cs
+++ b/Jellyfin.Server.Implementations/MediaSegments/MediaSegmentManager.cs
@@ -139,17 +139,17 @@ public class MediaSegmentManager : IMediaSegmentManager
     }
 
     /// <inheritdoc />
-    public Task<IEnumerable<MediaSegmentDto>> GetSegmentsAsync(Guid itemId, IEnumerable<MediaSegmentType>? typeFilter, bool filterByProvider = true)
+    public async Task<IEnumerable<MediaSegmentDto>> GetSegmentsAsync(Guid itemId, IEnumerable<MediaSegmentType>? typeFilter, bool filterByProvider = true)
     {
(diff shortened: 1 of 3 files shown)


Jellyfin.Server.Implementations/MediaSegments/MediaSegmentManager.cs
```diff
@@ -139,8 +139,16 @@ public async Task DeleteSegmentAsync(Guid segmentId)
     }
 
     /// <inheritdoc />
-    public async Task<IEnumerable<MediaSegmentDto>> GetSegmentsAsync(Guid itemId, IEnumerable<MediaSegmentType>? typeFilter)
+    public async Task<IEnumerable<MediaSegmentDto>> GetSegmentsAsync(Guid itemId, IEnumerable<MediaSegmentType>? typeFilter, bool filterByProvider = true)
     {
+        var baseItem = _libraryManager.GetItemById(itemId);
```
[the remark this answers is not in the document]
JPVenson: Added an overload for ID only

Shadowghost (approved):

Jellyfin.Server.Implementations/MediaSegments/MediaSegmentManager.cs
```diff
@@ -139,18 +139,47 @@ public async Task DeleteSegmentAsync(Guid segmentId)
...
+                return [];
+            }
+
+            var providerIds = providers.Select(f => GetProviderId(f.Name)).ToArray();
+            query = query.Where(e => providerIds.Contains(e.SegmentProviderId));
+        }
+
         return query
             .OrderBy(e => e.StartTicks)
             .AsNoTracking()
```
cvium, line 185: Unrelated to the PR: Why is there a `ToImmutableList()` followed by a `Select` below this line? The immutability isn't necessary here and it is basically discarded by the `Select`

Jellyfin.Server.Implementations/MediaSegments/MediaSegmentManager.cs
```diff
@@ -139,18 +139,47 @@ public async Task DeleteSegmentAsync(Guid segmentId)
     }
 
     /// <inheritdoc />
-    public async Task<IEnumerable<MediaSegmentDto>> GetSegmentsAsync(Guid itemId, IEnumerable<MediaSegmentType>? typeFilter)
+    public Task<IEnumerable<MediaSegmentDto>> GetSegmentsAsync(Guid itemId, IEnumerable<MediaSegmentType>? typeFilter, bool filterByProvider = true)
```
cvium, line 142: David Fowler recommends using async-await over returning the Task directly: https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#prefer-asyncawait-over-directly-returning-task

I don't think the slight performance gain (premature optimization) by not using async-await is worth mangling the stacktrace etc.

JPVenson: I disagree that the performance difference is negaligable esp in our case where we usually use ConfigureAwait(false)

Jellyfin.Server.Implementations/MediaSegments/MediaSegmentManager.cs
JPVenson: Fixed by using AsEnumerable to enumerate the EfCore query.

Jellyfin.Server.Implementations/MediaSegments/MediaSegmentManager.cs
cvium: Can you elaborate? I don't see how ConfigureAwait has any impact on whether we should use async-await or return tasks directly.

Jellyfin.Server.Implementations/MediaSegments/MediaSegmentManager.cs
```diff
@@ -139,23 +139,53 @@ public async Task DeleteSegmentAsync(Guid segmentId)
...
+            var libraryOptions = _libraryManager.GetLibraryOptions(item);
+            var providers = _segmentProviders
+                .Where(e => !libraryOptions.DisabledMediaSegmentProviders.Contains(GetProviderId(e.Name)))
+                .ToArray();
+            if (providers.Length == 0)
+            {
+                return [];
+            }
+
+            var providerIds = providers.Select(f => GetProviderId(f.Name)).ToArray();
```
crobibero, line 179: This could be combined into the `providers` statement to not create two arrays

Jellyfin.Server.Implementations/MediaSegments/MediaSegmentManager.cs
```diff
@@ -139,23 +139,53 @@ public async Task DeleteSegmentAsync(Guid segmentId)
...
+        }
+
         return query
             .OrderBy(e => e.StartTicks)
             .AsNoTracking()
-            .ToImmutableList()
-            .Select(Map);
+            .AsEnumerable()
+            .Select(Map)
+            .ToImmutableArray();
```
cvium, line 188:
```suggestion
            .ToArray();
```

cvium (approved):

PR #12925, opened 2024-10-31
Always await instead of directly returning Task

Bond-009:
https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#prefer-asyncawait-over-directly-returning-task

The performance impact is negligible (and it's me saying that!)

### DIFF
--- a/Emby.Server.Implementations/HttpServer/WebSocketConnection.cs
+++ b/Emby.Server.Implementations/HttpServer/WebSocketConnection.cs
@@ -82,17 +82,17 @@ namespace Emby.Server.Implementations.HttpServer
         public WebSocketState State => _socket.State;
 
         /// <inheritdoc />
-        public Task SendAsync(OutboundWebSocketMessage message, CancellationToken cancellationToken)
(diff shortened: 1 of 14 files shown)


cvium (approved):

Shadowghost (approved):

DaRealHansForSure:
Are there any downsides of this?
Why is nobody give some feedback from the Jellyfin team?

When not, i will manually apply the patch :)

gnattu:
Claiming "Nobody gives feedback" when there are two approvals is a bit weird.

DaRealHansForSure:
“Thumbs up” or ‘Approval’ is not a sign of ‘Yes it is noticeable, but it makes sense or yes we can stand behind it’.

“Approval” here can primarily mean ‘Yep, the code doesn't seem to cause any problems, let's go for it’.

I wanted to hear if anyone has effectively observed this from the devs etc. and possibly why it didn't make it into the releases as it still has a big performance impact?

PR #12932, opened 2024-11-01
Add support for external chapter files

tharunprabhu:
**Changes**
Adds support for external chapter files (XML format). If embedded chapters are already present, they will be ignored and external chapters will be used. 

Chapter file should be named as `<videofilename>.chapters.xml`. XML file must follow the Matroska Media Container Format Specification as specified in [RFC9559](https://www.rfc-editor.org/rfc/rfc9559#name-chapters)

Sample XML structure that is supported by this PR (I've added comments explaining what chapters will be read and what will be ignored):

```
<?xml version="1.0" encoding="UTF-8"?>
<!-- <!DOCTYPE Chapters SYSTEM "matroskachapters.dtd"> -->
<Chapters>

  <!-- The EditionEntry to be read will be selected based on EditionFlagDefault option. If this option is not specified, then the very first EditionEntry will be read-->
  <EditionEntry>
    <EditionUID>111</EditionUID>

    <!-- ✅ This chapter will be read (top-level, not hidden, valid time) -->
    <ChapterAtom>
      <ChapterUID>1</ChapterUID>
      <ChapterTimeStart>00:00:00.000000000</ChapterTimeStart>
      <ChapterDisplay>
        <ChapString>Intro</ChapString>
      </ChapterDisplay>
    </ChapterAtom>

    <!-- ❌ This chapter will be skipped due to ChapterFlagHidden=1 -->
    <ChapterAtom>
      <ChapterUID>2</ChapterUID>
      <ChapterTimeStart>00:01:00.000000000</ChapterTimeStart>
      <ChapterFlagHidden>1</ChapterFlagHidden>
      <ChapterDisplay>
        <ChapString>Hidden Chapter</ChapString>
      </ChapterDisplay>
    </ChapterAtom>

    <!-- ✅ This will be read as a top-level chapter -->
    <ChapterAtom>
      <ChapterUID>3</ChapterUID>
      <ChapterTimeStart>00:02:00.000000000</ChapterTimeStart>
      <ChapterDisplay>
        <ChapString>Main Chapter</ChapString>
      </ChapterDisplay>

      <!-- ❌ These nested subchapters will be ignored -->
      <ChapterAtom>
        <ChapterUID>4</ChapterUID>
        <ChapterTimeStart>00:02:10.000000000</ChapterTimeStart>
        <ChapterDisplay>
          <ChapString>Subchapter A</ChapString>
        </ChapterDisplay>
      </ChapterAtom>

      <ChapterAtom>
        <ChapterUID>5</ChapterUID>
        <ChapterTimeStart>00:02:30.000000000</ChapterTimeStart>
        <ChapterDisplay>
          <ChapString>Subchapter B</ChapString>
        </ChapterDisplay>
      </ChapterAtom>
    </ChapterAtom>

    <!-- ❌ This chapter will be skipped due to missing time -->
    <ChapterAtom>
      <ChapterUID>6</ChapterUID>
      <!-- Missing ChapterTimeStart -->
      <ChapterDisplay>
        <ChapString>Missing Time</ChapString>
      </ChapterDisplay>
    </ChapterAtom>

    <!-- ✅ This chapter will be read -->
    <ChapterAtom>
      <ChapterUID>3</ChapterUID>
      <ChapterTimeStart>00:02:00.000000000</ChapterTimeStart>
      <ChapterDisplay>
        <ChapString>Second Chapter</ChapString>
      </ChapterDisplay>
    </ChapterAtom>

  </EditionEntry>

  <!-- ❌ This entire EditionEntry will be ignored (not the first one) -->
  <EditionEntry>
    <EditionUID>222</EditionUID>
    <ChapterAtom>
      <ChapterUID>7</ChapterUID>
      <ChapterTimeStart>00:10:00.000000000</ChapterTimeStart>
      <ChapterDisplay>
        <ChapString>Secondary Edition Chapter</ChapString>
      </ChapterDisplay>
    </ChapterAtom>
  </EditionEntry>

</Chapters>
```

**Issues**
https://features.jellyfin.org/posts/1782/support-side-by-side-external-chapter-metadata


MediaBrowser.Providers/Chapters/ChapterManager.cs
```diff
@@ -22,5 +26,57 @@ public void SaveChapters(Guid itemId, IReadOnlyList<ChapterInfo> chapters)
...
+        /// <param name="video">The video item.</param>
+        /// <param name="chapters">The set of chapters.</param>
+        /// <returns>A list of ChapterInfo objects created using data read from the local XML file.</returns>
+        public ChapterInfo[] AddExternalChapters(Video video, IReadOnlyList<ChapterInfo> chapters)
+        {
+            string? xmlFilePath = null;
+
+            // Look for the XML Chapter file.
+            string[] matchingFiles = Directory.GetFiles(video.ContainingFolderPath, "*chapters.xml");
+            xmlFilePath = matchingFiles.FirstOrDefault(f => f.EndsWith("chapters.xml", StringComparison.OrdinalIgnoreCase));
```
mihawk90, line 42: Just a question for clarification (also for future documentation), am I reading lines 41-42 right that the exact filename doesn't matter?

So for example for a video file `video.mkv` would both `video.mkv.chapters.xml` as well as `video.chapters.xml` work?
What about things like `video.mkv.somerandomstring.chapters.xml` (since it's a wildcard)?

tharunprabhu: Hi, as long as the filename ends with chapters.xml, the file will be picked. So yes, video.mkv.chapters.xml, video.chapters.xml, and video.mkv.somerandomstring.chapters.xml are all valid filenames.

mihawk90: Alright thanks, perfect :)

Shadowghost: I wonder if it would make more sense to be less lenient and just allow `<filename.chapters.xml`

tharunprabhu: Yeah, that makes much more sense. I’ve updated it so that only files named `<videofilename>.chapters.xml` will match. 

For example, if the video file is `Sintel (2010).mkv`, only `Sintel (2010).chapters.xml` will be selected.

Marth64x:
Hi,
Does this support external ffmetadata file too?
Thanks!

tharunprabhu:
Hi, the current PR doesn't support it. But I do have plans to add support for ffmetadata files too.

gitproxyid:
> Hi, the current PR doesn't support it. But I do have plans to add support for ffmetadata files too.

Thank you for doing this even with XML!

Marth64x:
> Hi, the current PR doesn't support it. But I do have plans to add support for ffmetadata files too.

Thank you for doing this even with XML!

MediaBrowser.Providers/MediaInfo/ProbeProvider.cs
```diff
@@ -162,6 +164,20 @@ public bool HasChanged(BaseItem item, IDirectoryService directoryService)
                     _logger.LogDebug("Refreshing {ItemPath} due to external audio change.", item.Path);
                     return true;
                 }
+
+                var xmlFilePath = MediaInfoResolver.GetExternalChapterFile(video);
+                externalFiles = new HashSet<string>([xmlFilePath], StringComparer.OrdinalIgnoreCase);
+                if (!new HashSet<string>(video.ChapterFiles, StringComparer.Ordinal).SetEquals(externalFiles))
```
JPVenson, line 170: this is a comparison of one element to a set. Creating a new hashset for this feels overkill.

MediaBrowser.Providers/MediaInfo/MediaInfoResolver.cs
```diff
@@ -304,6 +304,29 @@ public IReadOnlyList<ExternalPathParserResult> GetExternalFiles(
...
+            string? xmlFilePath = null;
+
+            // Look for the XML Chapter file.
+            string[] matchingFiles = Directory.GetFiles(video.ContainingFolderPath, "*.chapters.xml");
+            xmlFilePath = matchingFiles.FirstOrDefault(f =>
+            {
+                ReadOnlySpan<char> fileName = Path.GetFileName(f.AsSpan());
+                ReadOnlySpan<char> expectedFileName = video.FileNameWithoutExtension;
+                return fileName.StartsWith(expectedFileName, StringComparison.Ordinal)
+                    && fileName.Slice(expectedFileName.Length).SequenceEqual(".chapters.xml");
```
JPVenson, line 324:
```suggestion
                    && fileName.EndsWith(".chapters.xml", StringComparison.Ordinal);
```

MediaBrowser.Providers/MediaInfo/FFProbeVideoInfo.cs
```diff
@@ -625,6 +629,57 @@ private async Task AddExternalAudioAsync(
             currentStreams.AddRange(externalAudioStreams);
         }
 
+        /// <summary>
+        /// Reads chapter information from external XML file.
+        /// </summary>
+        /// <param name="video">The video item.</param>
+        /// <param name="chapters">The set of chapters.</param>
+        /// <returns>A list of ChapterInfo objects read from the local XML file.</returns>
+        private static Task<ChapterInfo[]> AddExternalChapters(Video video, IReadOnlyList<ChapterInfo> chapters)
```
JPVenson, line 638: should this really be the source of truth if exists?

tharunprabhu: Hi, Thanks for the review. I think external chapters should be prioritized over embedded ones because users generally wouldn't have external chapters if they wanted to use the embedded ones.

MediaBrowser.Providers/MediaInfo/MediaInfoResolver.cs
tharunprabhu: Based on @Shadowghost 's input (see https://github.com/jellyfin/jellyfin/pull/12932#discussion_r1912239609) we decided it would be better to enforce stricter filename checks. The proposed change ensures that filenames like `<moviefilename> somerandomstring.chapters.xml` are recognized, whereas the original code only matches `<moviefilename>.chapters.xml`. Do you still recommend proceeding with this change?

tharunprabhu: Yes, it’s overkill for now. I just wanted to future-proof the code a bit, as it could be helpful if someone wants to add support for other types of chapter files (like ffmetadata files) in the future. In such cases, where GetExternalChapterFile(s) might return more than one element, hashsets could prove useful. I can switch to a simpler check if you’d prefer.

MediaBrowser.Providers/MediaInfo/ProbeProvider.cs
tharunprabhu: Yes, it’s overkill for now. I just wanted to future-proof the code a bit, as it could be helpful if someone wants to add support for other types of chapter files (like ffmetadata files) in the future. In such cases, where GetExternalChapterFile(s) might return more than one element, hashsets could prove useful. I can switch to a simpler check if you’d prefer.

MediaBrowser.Providers/MediaInfo/MediaInfoResolver.cs
soultaco83: Hope it is appropriate for me to comment on this.
I was looking to make my own chapter files and came across this pr. I agree with tharunorabhu and shadowghost. I believe it would be better to keep it filename.chapters.xml instead of allowing wildcard. Allows for less accidental associations

MediaBrowser.Providers/MediaInfo/FFProbeVideoInfo.cs
soultaco83: I agree with @tharunprabhu
If a external chapter files exists it would be because I made one.
A toggle in the library might be nice just in case though. Just to say yes prioritize or not.

mihawk90: Should also consider the fact that it isn't even possible any other way.

Since JF does not support multiple different chapters there would be no way to choose between the two. Which in turn means this entire PR would be moot if the external file were not prioritised.

Multiple different chapter support may be nice in the future (although... usecases are probably limited at best), but that would require a whole lot more work then this (relatively) simple implementation.

MediaBrowser.Providers/MediaInfo/FFProbeVideoInfo.cs
```diff
@@ -625,6 +629,57 @@ private async Task AddExternalAudioAsync(
...
+
+            // Navigate to the "chapters" node
+            XmlNodeList chapterNodes = xmlDoc.GetElementsByTagName("chapter");
+
+            foreach (XmlNode chapterNode in chapterNodes)
+            {
+                // Parse the "time" attribute as the chapter start time
+                var startTime = chapterNode.Attributes?["time"]?.InnerText;
+                // Parse the "name" attribute as the chapter name
+                var name = chapterNode.Attributes?["name"]?.InnerText;
```
mihawk90: I have two notes on this:

1. Can you provide a sample of the XML structure that would work with this PR?
    * In the PR's description you're referencing [a feature request](https://features.jellyfin.org/posts/1782/support-side-by-side-external-chapter-metadata), but that in turn is explicitly referencing Matroska-formatted Chapter files.
    * However, the [Matroska Chapter spec](https://www.matroska.org/technical/chapters.html) doesn't specify a tag by the name of `<chapter>`. They use `<ChapterAtom>` instead.
    * From what I can tell this code would only work with a structure like this:
    ```xml
    <?xml version="1.0"?>
    <chapters>
      <chapter time="00:00:00.000" name="Chapter 01" />
      <chapter time="00:01:00.000" name="Chapter 02" />
    </chapters>
    ```
    However, that is not valid for Matroska and in fact can't be loaded by MKVToolNix at all.
    * edit months later: I just happened to stumble on an old file in my library that is using this format, and this seems to be Kodi's [Episode Bookmarks](https://kodi.wiki/view/Bookmarks_and_chapters), altho I have no recollection on whether Kodi wrote this file or if I created it manually (the file is from 2014, go figure). Regardless it's not a standardised format and Kodi doesn't even seem to document this, and it wasn't what the Feature request asked for anyway.
    * Neither PR nor code specify the allowed format for the timestamps (although I guess `TryParse` below just tries to guess with whatever we throw at it), which is also not ideal.
    * Given the above, I don't think it's ideal for Jellyfin to introduce a new chapter format for really no reason I can think of. There is already existing tooling to easily create Matroska-compliant chapter XMLs (MKVToolNixGUI, [chapters_for_mpv](https://github.com/mar04/chapters_for_mpv), and - if we're staying in the Jellyfin ecosystem - [jellyfin-plugin-ms-chapter](https://github.com/intro-skipper/jellyfin-plugin-ms-chapter), just to name 3 examples). With a custom structure it would mean that these tools couldn't be used and users would have to write their Chapter files manually from scratch, which IMO cannot be the goal here. I don't think anyone would even use this feature if they had to manually write XMLs for however many episodes of TV shows they have.
2. I guess this doesn't really apply right now given 1., but I'll mention it anyways: *If* this is rewritten to use Matroska-compliant XMLs, simply selecting by `ChapterAtom` is going to be problematic when users drop in XML files containing multiple Editions. You'd have to select by `EditionEntry` first and go from there.
    * granted this could also just be documented, but given how enthusiastically the average user is reading documentation, it's probably best to just deal with it in code to begin with.
    * Example XML with 2 editions:
    ```xml
    <?xml version="1.0"?>
    <!-- <!DOCTYPE Chapters SYSTEM "matroskachapters.dtd"> -->
    <Chapters>
      <EditionEntry>
        <EditionUID>14834319457174143500</EditionUID>
        <ChapterAtom>
          <ChapterTimeStart>00:00:00.000000000</ChapterTimeStart>
          <ChapterFlagHidden>0</ChapterFlagHidden>
          <ChapterFlagEnabled>1</ChapterFlagEnabled>
          <ChapterDisplay>
            <ChapterString>Chapter 01</ChapterString>
            <ChapterLanguage>und</ChapterLanguage>
            <ChapLanguageIETF>und</ChapLanguageIETF>
          </ChapterDisplay>
          <ChapterUID>4100189026629290061</ChapterUID>
        </ChapterAtom>
        <ChapterAtom>
          <ChapterTimeStart>00:05:00.000000000</ChapterTimeStart>
          <ChapterFlagHidden>0</ChapterFlagHidden>
          <ChapterFlagEnabled>1</ChapterFlagEnabled>
          <ChapterDisplay>
            <ChapterString>Chapter 02</ChapterString>
            <ChapterLanguage>und</ChapterLanguage>
            <ChapLanguageIETF>und</ChapLanguageIETF>
          </ChapterDisplay>
          <ChapterUID>11659832165557252992</ChapterUID>
        </ChapterAtom>
      </EditionEntry>
      <EditionEntry>
        <EditionUID>495346149353424957</EditionUID>
        <ChapterAtom>
          <ChapterTimeStart>00:01:00.000000000</ChapterTimeStart>
          <ChapterFlagHidden>0</ChapterFlagHidden>
          <ChapterFlagEnabled>1</ChapterFlagEnabled>
          <ChapterDisplay>
            <ChapterString>Chapter 01</ChapterString>
            <ChapterLanguage>und</ChapterLanguage>
            <ChapLanguageIETF>und</ChapLanguageIETF>
          </ChapterDisplay>
          <ChapterUID>10975746129216566163</ChapterUID>
        </ChapterAtom>
        <ChapterAtom>
          <ChapterTimeStart>00:06:00.000000000</ChapterTimeStart>
          <ChapterFlagHidden>0</ChapterFlagHidden>
          <ChapterFlagEnabled>1</ChapterFlagEnabled>
          <ChapterDisplay>
            <ChapterString>Chapter 02</ChapterString>
            <ChapterLanguage>und</ChapterLanguage>
            <ChapLanguageIETF>und</ChapLanguageIETF>
          </ChapterDisplay>
          <ChapterUID>13593348617991404047</ChapterUID>
        </ChapterAtom>
      </EditionEntry>
    </Chapters>
    ```
    If you were to select only by `ChapterAtom` users would end up with 4 Chapters (at 0, 1, 5, and 6 minutes, respectively) instead of the intended 2.

Zofware:
This PR is a thing of beauty. I love it. I want it. I'm running it. The merge conflict is trivial. @tharunprabhu please consider getting this PR across the finish line. I'm testing it with a Home Movies collection on Windows and it's working really well. Thank you.

tharunprabhu:
@Zofware Thank you! Glad to hear it’s working well. I’ll resolve the merge conflict by this weekend and get it ready to merge.

mihawk90:
Would probably be a good idea to clarify with JF devs what file format is supposed to be supported, see my review comment above.

jumoog:
> Would probably be a good idea to clarify with JF devs what file format is supposed to be supported, see my review comment above.

I fully agree! It would make sense to support a standardised format, such as [MKV Chapters](https://www.matroska.org/technical/chapters.html).

tharunprabhu:
@jumoog To the best of my knowledge, external chapter files are typically in XML format. This is the format used by both (the now defunct) chapterDB and chapterEditor tool. This current PR would add external chapter support for video files irrespective of the container they use. However, MKV chapters are usually embedded within the MKV container itself and are specific to that format.

jumoog:
Your implementation does not follow any current/used standards and cannot be validated. While the MKV Chapters XML is specified in [RFC9559](https://www.rfc-editor.org/rfc/rfc9559#section-20.5.1-3). From my point of view, we should choose a format that can easily be imported and exported. Also the feature request mentions MKV Chapters.

mihawk90:
> in XML format

This says nothing.

We are not talking about the file format, we are talking about the structure used *in* that format.

As I'm sure you are aware, MKV chapters are also XML, yet they are not compatible with your PR, hence my review comment above.

> This is the format used by both (the now defunct) chapterDB and chapterEditor tool

If you are talking about this tool here:
https://www.videohelp.com/software/chapterEditor

Then that can also write MKV chapters, so this point is moot (i.e. users of that tool wouldn't even have to change the tool they are using).

edit:
One more point about this that just came to mind, since I just woke up:
Is there any *other* software that writes this format? I don't think it is appropriate for Jellyfin to force users into using a specific tool if they don't want to be typing out XML by hand. A tool mind you that seems to be barely maintained and *not* OpenSource.

That it was used by chapterDB is irrelevant since - as you say - it is defunct (and maybe the nonstandard format was part of why it wasn't being used?).

> However, MKV chapters are usually embedded within the MKV container itself and are specific to that format.

This is all the more reason to use it.

Just because it is "usually" embedded doesn't disqualify it as a candidate, IMO quite the contrary.

One usecase was already mentioned in the Feature Request: Extracting Chapters from an existing file that is to be replaced, without having to edit the new file replacing the old.

With your current PR, this would not work. Users would have to not only extract the chapters from the existing file (a task which takes a whole 2 seconds to do), but also then convert the standardised format into this unknown nonstandard format, but for what benefit?

Furthermore, using MKV chapters format also allows reusing that same file for muxing later on if the user wants to, which likewise would not be possible with this custom format. As already mentioned, there are also tools within the Jellyfin ecosystem (e.g. [jellyfin-plugin-ms-chapter](https://github.com/intro-skipper/jellyfin-plugin-ms-chapter)) that make use of this very fact. They export the file in MKV chapter format precisely *because* it is so widely used.

As it currently stands this PR could not be used to utilise such files, which leaves 2 options:
1. The plugin to write this non-standard format, which could then not be used to mux into an MKV later on
2. The user manually converting the files yet again

Both of which provide 0 benefit over just using a standardised format in the first place. If this PR were to use the MKV chapters, then users could just use the file right away, or mux later on, whatever they choose.

I fail to see any kind of justification for using a non-standard format over arguably *the* most widely used chapter format in the average user's media, simply because most content is provided as an MKV.

tharunprabhu:
@mihawk90 Thanks for the clarification—those are solid points. My code currently parses the "ChapterGrabber XML" format, which, while not obscure, is non-standard and likely no longer maintained. I agree it’s better to stick to a standardized format that's already defined in an RFC. I’ll update the XML parsing code accordingly.

mihawk90:
> @mihawk90 I’ll update the XML parsing code accordingly.

Thank you.
When you're doing that, please keep in mind the second bullet point in my original review comment (about the multiple editions).

tharunprabhu:
> please keep in mind the second bullet point in my original review comment (about the multiple editions).

This is about how Jellyfin has to prioritize external chapters over embedded ones?

mihawk90:
No. As explained there this is about MKV chapter files being able to contain mutliple editions and how (and why) to deal with this.

mihawk90:
Actually I just noticed something...
IDK WTF is wrong with GitHub, but when I open this page in an Incognito window, Review is nowhere to be found ... ???

I'm just gonna copy and paste it here as a comment, I guess (note this was written at the end of February).

I guess that would explain why noone ever addressed it :melting_face: 

---

I have two notes on this:

1. Can you provide a sample of the XML structure that would work with this PR?
    * In the PR's description you're referencing [a feature request](https://features.jellyfin.org/posts/1782/support-side-by-side-external-chapter-metadata), but that in turn is explicitly referencing Matroska-formatted Chapter files.
    * However, the [Matroska Chapter spec](https://www.matroska.org/technical/chapters.html) doesn't specify a tag by the name of `<chapter>`. They use `<ChapterAtom>` instead.
    * From what I can tell this code would only work with a structure like this:
    ```xml
    <?xml version="1.0"?>
    <chapters>
      <chapter time="00:00:00.000" name="Chapter 01" />
      <chapter time="00:01:00.000" name="Chapter 02" />
    </chapters>
    ```
    However, that is not valid for Matroska and in fact can't be loaded by MKVToolNix at all.
    * edit months later: I just happened to stumble on an old file in my library that is using this format, and this seems to be Kodi's [Episode Bookmarks](https://kodi.wiki/view/Bookmarks_and_chapters), altho I have no recollection on whether Kodi wrote this file or if I created it manually (the file is from 2014, go figure). Regardless it's not a standardised format and Kodi doesn't even seem to document this, and it wasn't what the Feature request asked for anyway.
    * Neither PR nor code specify the allowed format for the timestamps (although I guess `TryParse` below just tries to guess with whatever we throw at it), which is also not ideal.
    * Given the above, I don't think it's ideal for Jellyfin to introduce a new chapter format for really no reason I can think of. There is already existing tooling to easily create Matroska-compliant chapter XMLs (MKVToolNixGUI, [chapters_for_mpv](https://github.com/mar04/chapters_for_mpv), and - if we're staying in the Jellyfin ecosystem - [jellyfin-plugin-ms-chapter](https://github.com/intro-skipper/jellyfin-plugin-ms-chapter), just to name 3 examples). With a custom structure it would mean that these tools couldn't be used and users would have to write their Chapter files manually from scratch, which IMO cannot be the goal here. I don't think anyone would even use this feature if they had to manually write XMLs for however many episodes of TV shows they have.
2. I guess this doesn't really apply right now given 1., but I'll mention it anyways: *If* this is rewritten to use Matroska-compliant XMLs, simply selecting by `ChapterAtom` is going to be problematic when users drop in XML files containing multiple Editions. You'd have to select by `EditionEntry` first and go from there.
    * granted this could also just be documented, but given how enthusiastically the average user is reading documentation, it's probably best to just deal with it in code to begin with.
    * Example XML with 2 editions:
    ```xml
    <?xml version="1.0"?>
    <!-- <!DOCTYPE Chapters SYSTEM "matroskachapters.dtd"> -->
    <Chapters>
      <EditionEntry>
        <EditionUID>14834319457174143500</EditionUID>
        <ChapterAtom>
          <ChapterTimeStart>00:00:00.000000000</ChapterTimeStart>
          <ChapterFlagHidden>0</ChapterFlagHidden>
          <ChapterFlagEnabled>1</ChapterFlagEnabled>
          <ChapterDisplay>
            <ChapterString>Chapter 01</ChapterString>
            <ChapterLanguage>und</ChapterLanguage>
            <ChapLanguageIETF>und</ChapLanguageIETF>
          </ChapterDisplay>
          <ChapterUID>4100189026629290061</ChapterUID>
        </ChapterAtom>
        <ChapterAtom>
          <ChapterTimeStart>00:05:00.000000000</ChapterTimeStart>
          <ChapterFlagHidden>0</ChapterFlagHidden>
          <ChapterFlagEnabled>1</ChapterFlagEnabled>
          <ChapterDisplay>
            <ChapterString>Chapter 02</ChapterString>
            <ChapterLanguage>und</ChapterLanguage>
            <ChapLanguageIETF>und</ChapLanguageIETF>
          </ChapterDisplay>
          <ChapterUID>11659832165557252992</ChapterUID>
        </ChapterAtom>
      </EditionEntry>
      <EditionEntry>
        <EditionUID>495346149353424957</EditionUID>
        <ChapterAtom>
          <ChapterTimeStart>00:01:00.000000000</ChapterTimeStart>
          <ChapterFlagHidden>0</ChapterFlagHidden>
          <ChapterFlagEnabled>1</ChapterFlagEnabled>
          <ChapterDisplay>
            <ChapterString>Chapter 01</ChapterString>
            <ChapterLanguage>und</ChapterLanguage>
            <ChapLanguageIETF>und</ChapLanguageIETF>
          </ChapterDisplay>
          <ChapterUID>10975746129216566163</ChapterUID>
        </ChapterAtom>
        <ChapterAtom>
          <ChapterTimeStart>00:06:00.000000000</ChapterTimeStart>
          <ChapterFlagHidden>0</ChapterFlagHidden>
          <ChapterFlagEnabled>1</ChapterFlagEnabled>
          <ChapterDisplay>
            <ChapterString>Chapter 02</ChapterString>
            <ChapterLanguage>und</ChapterLanguage>
            <ChapLanguageIETF>und</ChapLanguageIETF>
          </ChapterDisplay>
          <ChapterUID>13593348617991404047</ChapterUID>
        </ChapterAtom>
      </EditionEntry>
    </Chapters>
    ```
    If you were to select only by `ChapterAtom` users would end up with 4 Chapters (at 0, 1, 5, and 6 minutes, respectively) instead of the intended 2.

mihawk90 (commented):
Actually I just noticed something...
IDK WTF is wrong with GitHub, but when I open this page in an Incognito window, my Review Comment is nowhere to be found ... ???

~~I'm just gonna copy and paste it here as a comment, I guess (note this was written at the end of February).~~

I guess that would explain why noone ever addressed it 🫠

---

edit:
GitHub has the literal worst UX in the galaxy for this, courtesy of this comment here:

https://github.com/orgs/community/discussions/30638#discussioncomment-4574199

Only 2 years old I guess, so we got plenty of time!

tharunprabhu:
@mihawk90 I was staring at the page wondering if I was missing something or losing my mind 😅 All good now though. I can see your review comments. I've got a basic implementation of external Matroska XML chapter support working now. It’ll take some time to add support for multiple editions, nested chapters, and other possible edge cases. I’ll update this PR once everything is fully working.

mihawk90:
> It’ll take some time to add support for multiple editions

I don't think it makes sense to (truly) implement multiple editions.

Jellyfin does not support Editions (or rather, multiple different versions of chapter menus in general), so even if you were to implement reading all Editions, they could not be used. This is why I mentioned just getting the first Edition and be done with it.
That is also what Jellyfin does when it encounters an MKV with multiple embedded Editions (might actually be what ffmpeg does, not Jellyfin specifically), so mimicking that same behaviour just makes sense.

I have a movie somewhere that has this, but I'll need to find which it was.

tharunprabhu:
> > It’ll take some time to add support for multiple editions
> 
> I don't think it makes sense to (truly) implement multiple editions.
> 
> Jellyfin does not support Editions (or rather, multiple different versions of chapter menus in general), so even if you were to implement reading all Editions, they could not be used. This is why I mentioned just getting the first Edition and be done with it. That is also what Jellyfin does when it encounters an MKV with multiple embedded Editions (might actually be what ffmpeg does, not Jellyfin specifically), so mimicking that same behaviour just makes sense.
> 
> I have a movie somewhere that has this, but I'll need to find which it was.

Yeah, this is what I was thinking too. By "support" I simply meant that the code needs to do what it's expected to do when it encounters multiple editions which you have already explained. Maybe "support" wasn't the right word to use.

Also, do you know how JF handles nested chapters? Does it ignore the parent chapter or does it ignore the sub-chapters?

mihawk90:
> I have a movie somewhere that has this, but I'll need to find which it was.

Ugh, took a bit, but I found it:
```
❯ mediainfo *mkv
General
Complete name                            : Vantage Point (2008) [tmdbid-7461] - 1080p.mkv
…

Menu #1
00:00:00.000                             : en:Chapter 01
00:05:33.624                             : en:Chapter 02
00:11:52.837                             : en:Chapter 03
00:18:31.610                             : en:Chapter 04
00:22:29.848                             : en:Chapter 05
00:26:03.478                             : en:Chapter 06
00:30:36.042                             : en:Chapter 07
00:37:00.051                             : en:Chapter 08
00:42:50.985                             : en:Chapter 09
00:48:51.094                             : en:Chapter 10
00:51:50.398                             : en:Chapter 11
00:54:42.195                             : en:Chapter 12
00:59:53.214                             : en:Chapter 13
01:04:56.475                             : en:Chapter 14
01:12:05.446                             : en:Chapter 15
01:18:19.528                             : en:Chapter 16

Menu #2
00:00:00.000                             : en:Chapter 01
00:01:23.750                             : en:Chapter 02
00:03:04.017                             : en:Chapter 03
00:05:18.693                             : en:Chapter 04
00:06:15.750                             : en:Chapter 05
00:06:50.952                             : en:Chapter 06
00:07:24.235                             : en:Chapter 07
00:09:19.600                             : en:Chapter 08
00:11:47.998                             : en:Chapter 09
00:13:16.545                             : en:Chapter 10
00:14:32.997                             : en:Chapter 11
00:15:49.531                             : en:Chapter 12
00:17:19.955                             : en:Chapter 13
00:18:32.444                             : en:Chapter 14
00:19:41.096                             : en:Chapter 15
00:20:33.065                             : en:Chapter 16
00:22:38.899                             : en:Chapter 17
00:23:58.645                             : en:Chapter 18
00:25:21.686                             : en:Chapter 19
00:26:30.297                             : en:Chapter 20
00:27:34.736                             : en:Chapter 21
00:28:40.760                             : en:Chapter 22
00:30:46.052                             : en:Chapter 23
00:31:58.833                             : en:Chapter 24
00:32:49.509                             : en:Chapter 25
00:34:33.780                             : en:Chapter 26
00:36:07.707                             : en:Chapter 27
00:37:38.839                             : en:Chapter 28
00:39:13.642                             : en:Chapter 29
00:40:51.240                             : en:Chapter 30
00:41:38.454                             : en:Chapter 31
00:43:24.727                             : en:Chapter 32
00:45:56.128                             : en:Chapter 33
00:47:58.333                             : en:Chapter 34
00:49:46.733                             : en:Chapter 35
00:50:50.422                             : en:Chapter 36
00:52:41.992                             : en:Chapter 37
00:54:35.647                             : en:Chapter 38
00:56:07.364                             : en:Chapter 39
00:57:07.715                             : en:Chapter 40
00:58:07.400                             : en:Chapter 41
00:59:05.291                             : en:Chapter 42
01:01:07.789                             : en:Chapter 43
01:02:01.843                             : en:Chapter 44
01:03:18.336                             : en:Chapter 45
01:04:53.014                             : en:Chapter 46
01:06:17.807                             : en:Chapter 47
01:09:47.808                             : en:Chapter 48
01:12:21.295                             : en:Chapter 49
01:13:37.413                             : en:Chapter 50
01:14:48.275                             : en:Chapter 51
```

image

> Also, do you know how JF handles nested chapters? Does it ignore the parent chapter or does it ignore the sub-chapters?

I don't know, I have never encountered a file with nested chapters and haven't gone out of my way to test it.

I also don't know the exact usecase of having this, from reading the MKV spec it sounds like this is more intended for technical specification of subsegments instead of "real" (human-interactable) chapters, in which case it would make sense to drop the Nested chapters.

tharunprabhu:
@mihawk90 My current code (testing locally) ignores Menu 2 (i.e. the 2nd EditionEntry) and only parses the 16 chapters in Menu 1 . This is your expectation, right? Thanks for the clarity on nested chapters.

mihawk90:
> Only parses the 16 chapters in Menu 1 . This is your expectation, right?

Yes.

On that note, this is also what ffmpeg reports (and I assume why JF does not know about the 2nd Edition):
```
❯ ffprobe *mkv
…
Input #0, matroska,webm, from 'Vantage Point (2008) [tmdbid-7461] - 1080p.mkv':
…
  Chapters:
    Chapter #0:0: start 0.000000, end 333.624958
      Metadata:
        title           : Chapter 01
    Chapter #0:1: start 333.624958, end 712.837125
      Metadata:
        title           : Chapter 02
    Chapter #0:2: start 712.837125, end 1111.610500
      Metadata:
        title           : Chapter 03
    Chapter #0:3: start 1111.610500, end 1349.848500
      Metadata:
        title           : Chapter 04
    Chapter #0:4: start 1349.848500, end 1563.478583
      Metadata:
        title           : Chapter 05
    Chapter #0:5: start 1563.478583, end 1836.042542
      Metadata:
        title           : Chapter 06
    Chapter #0:6: start 1836.042542, end 2220.051167
      Metadata:
        title           : Chapter 07
    Chapter #0:7: start 2220.051167, end 2570.985083
      Metadata:
        title           : Chapter 08
    Chapter #0:8: start 2570.985083, end 2931.094833
      Metadata:
        title           : Chapter 09
    Chapter #0:9: start 2931.094833, end 3110.398958
      Metadata:
        title           : Chapter 10
    Chapter #0:10: start 3110.398958, end 3282.195583
      Metadata:
        title           : Chapter 11
    Chapter #0:11: start 3282.195583, end 3593.214625
      Metadata:
        title           : Chapter 12
    Chapter #0:12: start 3593.214625, end 3896.475917
      Metadata:
        title           : Chapter 13
    Chapter #0:13: start 3896.475917, end 4325.446125
      Metadata:
        title           : Chapter 14
    Chapter #0:14: start 4325.446125, end 4699.528167
      Metadata:
        title           : Chapter 15
    Chapter #0:15: start 4699.528167, end 5405.400000
      Metadata:
        title           : Chapter 16
  Stream #0:0(eng): Video: h264 (High) …
…
```

tharunprabhu:
@mihawk90 @jumoog I've pushed my changes. Also updated my original PR message with the relevant details (including sample XML file). Please test it out if possible and let me know if anything needs improvement. Thanks!

PS: If you're testing with multiple XML files, please do a manual "Replace All Metadata" refresh on the video file to force Jellyfin to read the new XML file immediately.

mihawk90:
The only thing I'll say is that the Edition selection could also follow the spec's "Default" selection outlined here:

https://www.rfc-editor.org/rfc/rfc9559#name-default-edition

Though whether that's desired is up to the JF team's review.

MediaBrowser.Providers/MediaInfo/ProbeProvider.cs
Shadowghost: @JPVenson wdyt?

Shadowghost (commented):
> Though whether that's desired is up to the JF team's review.

If we're at it, let's do it properly and implement it like the spec proposes

MediaBrowser.Providers/MediaInfo/ProbeProvider.cs
JPVenson: if we want to "futureproof" this, we should go the full road and make the whole system provider based. Meaning you have an interface "IChapterProvider" that gets called from an "IChapterService" and those providers will read the chapter information for different sources. Overall just like we do it with media segments.

Zofware:
I understand the rationale for switching to the Matroska chapter format. Still, the simplicity and easy readability of the XML format as originally implemented was nice. I will miss it.

Shadowghost (approved):

MediaBrowser.Controller/Entities/Video.cs
```diff
@@ -106,6 +107,12 @@ public override bool SupportsPositionTicksResume
         /// <value>The audio paths.</value>
         public string[] AudioFiles { get; set; }
 
+        /// <summary>
+        /// Gets or sets the chapter paths.
+        /// </summary>
+        /// <value>The chapter paths.</value>
+        public string[] ChapterFiles { get; set; }
```
JPVenson, line 114: that should really not be another json serialized value. We have a proper DB scheme now where stuff like this belongs in

MediaBrowser.Providers/MediaInfo/FFProbeVideoInfo.cs
```diff
@@ -628,6 +632,120 @@ private async Task AddExternalAudioAsync(
...
+        /// <summary>
+        /// Reads chapter information from external XML file (Matroska Media Container Format).
+        /// </summary>
+        /// <param name="video">The video item.</param>
+        /// <param name="chapters">The set of chapters.</param>
+        /// <returns>A list of ChapterInfo objects read from the local XML file.</returns>
+        private static Task<ChapterInfo[]> AddExternalChapters(Video video, IReadOnlyList<ChapterInfo> chapters)
+        {
+            var xmlFilePath = MediaInfoResolver.GetExternalChapterFile(video);
+            video.ChapterFiles = new[] { xmlFilePath };
```
JPVenson, line 644: should this really set it always even if the file does not exist?

MediaBrowser.Providers/MediaInfo/FFProbeVideoInfo.cs
```diff
@@ -628,6 +632,120 @@ private async Task AddExternalAudioAsync(
             currentStreams.AddRange(externalAudioStreams);
         }
 
+        /// <summary>
+        /// Reads chapter information from external XML file (Matroska Media Container Format).
+        /// </summary>
+        /// <param name="video">The video item.</param>
+        /// <param name="chapters">The set of chapters.</param>
+        /// <returns>A list of ChapterInfo objects read from the local XML file.</returns>
+        private static Task<ChapterInfo[]> AddExternalChapters(Video video, IReadOnlyList<ChapterInfo> chapters)
```
JPVenson, line 641: Doesn't use await so why async?

JPVenson (changes requested):

PR #12935, opened 2024-11-01
Require password being set if user is admin

### LINKED ISSUE #12927: Able to create Admin user with empty password

### Description of the bug

The server has moved to forbidding admin users with empty passwords when changing user passwords, however you can create a user that isn't an admin with an empty password and _then_ make them an admin.

Tested on 10.9 and 10.10.

### Reproduction steps

1 - Create normal user with empty password
2 - Update user policy `IsAdministrator` to true, or `Allow this user to manage the server` in web
3 - Now user with empty password is admin

An admin with empty password is allowed.

Probably an error, but I am honestly unsure of how to properly check this condition without looking at the password on server.

### Jellyfin Server version

10.10.0+

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.10.0

### Environment

N/A, is API issue.

### Jellyfin logs

Get 200 :shrug:

### FFmpeg logs

_No response_

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

Shadowghost:
**Changes**
* Require password being set if user is admin

**Issues**
Fixes #12927

### DIFF
--- a/Jellyfin.Api/Controllers/UserController.cs
+++ b/Jellyfin.Api/Controllers/UserController.cs
@@ -454,6 +454,12 @@ public class UserController : BaseJellyfinApiController
             }
         }
 
+        // If granting admin access, password needs to be set
+        if (newPolicy.IsAdministrator && string.IsNullOrEmpty(user.Password))
+        {
+            return StatusCode(StatusCodes.Status403Forbidden, "Administrator accounts always need to have a password.");
+        }
+
         // If disabling
         if (newPolicy.IsDisabled && user.HasPermission(PermissionKind.IsAdministrator))
         {


nielsvanvelzen (changes requested):
This must also be checked on the web client before merging

prahal:
Won't that lock install with only an admin without a password ?

nielsvanvelzen:
This doesn't change login so any admin users that currently have no password keep working. This change blocks the option to mark someone as admin when they don't have a password.

JPVenson (approved):

joshuaboniface:
@nielsvanvelzen What needs to be done in web for this?

nielsvanvelzen:
@joshuaboniface  The web client should check whether the user is admin or not and show an appropriate message, if this change is merged without a web client change it will just fail to save without feedback as to why.

viown:
Web seems to already enforce passwords for admins.

- https://github.com/jellyfin/jellyfin-web/blob/2ab28266f71dbc80256f9f3ac2c7177ae08c4009/src/components/dashboard/users/UserPasswordForm.tsx#L81
- https://github.com/jellyfin/jellyfin-web/blob/2ab28266f71dbc80256f9f3ac2c7177ae08c4009/src/controllers/wizard/user/index.html#L14

Jellyfin.Api/Controllers/UserController.cs
```diff
@@ -453,6 +453,12 @@ public async Task<ActionResult> UpdateUserPolicy(
             }
         }
 
+        // If granting admin access, password needs to be set
+        if (newPolicy.IsAdministrator && string.IsNullOrEmpty(user.Password))
```
viown, line 457: Should we check if the user is already an admin to still allow them to save if they don't have a password?

Shadowghost: Hm, good question. @nielsvanvelzen wdyt?

PR #12940, opened 2024-11-02
Remove DynamicImageResponse local image after saved to metadata folder

### LINKED ISSUE #12937: jellyfin make milions of temp files .jpg and fill inode linux

### Description of the bug

with the latest jellyfin updates, the server creates millions of temporary .jpg files by filling the ext4 inode
when I reboot the system it takes hours to delete all the temporary files, and I have to delete them before the reboot otherwise the server stays booting for hours.
my next test is to put a memory disk in fstab for the /tmp folder
the logs show no errors on these files

### Reproduction steps

open folder /tmp/jellyfin/nnnnnnnnnn/*

everyday

everyday

### Jellyfin Server version

10.10.0+

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

main

### Environment

- OS: arm64 ubuntu 22.04
- Linux Kernel: 5.10
- Virtualization: no
- Clients: jellyfinweb
- Browser: firefox
- FFmpeg Version: 
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:

### Jellyfin logs

irrelevant

### FFmpeg logs

_No response_

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

gnattu:
**Changes**

Previously, local images provided by DynamicImageResponse were never cleaned up until the server was restarted. This issue has become more severe in 10.10, as the default is now set to use the system's native temp folder, which might be a RAM backed tmpfs. This behavior could lead to resource starvation for long-running servers performing multiple library scans.

Metadata plugins prefer the old behavior should do its own backup.

**Issues**

Fixes #12937

### DIFF
--- a/MediaBrowser.Providers/Manager/ProviderManager.cs
+++ b/MediaBrowser.Providers/Manager/ProviderManager.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Net.Mime;
+using System.Runtime.ExceptionServices;
 using System.Threading;
 using System.Threading.Tasks;
 using AsyncKeyedLock;
@@ -258,7 +259,7 @@ namespace MediaBrowser.Providers.Manager
                 throw new ArgumentNullException(nameof(source));
             }
 
-            Exception? saveException = null;
+            ExceptionDispatchInfo? saveException = null;
 
             try
             {
@@ -267,7 +268,7 @@ namespace MediaBrowser.Providers.Manager
             }
             catch (Exception ex)
             {
-                saveException = ex;
+                saveException = ExceptionDispatchInfo.Capture(ex);
                 _logger.LogError(ex, "Unable to save image {Source}", source);
             }
             finally
@@ -282,14 +283,11 @@ namespace MediaBrowser.Providers.Manager
                 }
                 catch (Exception ex)
                 {
-                    saveException ??= ex;
+                    saveException ??= ExceptionDispatchInfo.Capture(ex);
                 }
             }
 
-            if (saveException is not null)
-            {
-                throw saveException;
-            }
+            saveException?.Throw();
         }
 
         /// <inheritdoc/>
--- a/MediaBrowser.Providers/Manager/ProviderManager.cs
+++ b/MediaBrowser.Providers/Manager/ProviderManager.cs
@@ -270,7 +270,7 @@ namespace MediaBrowser.Providers.Manager
                 {
                     File.Delete(source);
                 }
-                catch (IOException ex)
+                catch (Exception ex)
                 {
                     _logger.LogError(ex, "Source file {Source} not found or in use, skip removing", source);
                 }
--- a/MediaBrowser.Providers/Manager/ProviderManager.cs
+++ b/MediaBrowser.Providers/Manager/ProviderManager.cs
@@ -258,15 +258,37 @@ namespace MediaBrowser.Providers.Manager
                 throw new ArgumentNullException(nameof(source));
             }
 
-            var fileStream = AsyncFile.OpenRead(source);
-            await new ImageSaver(_configurationManager, _libraryMonitor, _fileSystem, _logger).SaveImage(item, fileStream, mimeType, type, imageIndex, saveLocallyWithMedia, cancellationToken);
+            Exception? saveException = null;
+
             try
             {
-                File.Delete(source);
+                var fileStream = AsyncFile.OpenRead(source);
+                await new ImageSaver(_configurationManager, _libraryMonitor, _fileSystem, _logger).SaveImage(item, fileStream, mimeType, type, imageIndex, saveLocallyWithMedia, cancellationToken);
             }
-            catch (IOException ex)
+            catch (Exception ex)
             {
-                _logger.LogError(ex, "Source file {Source} not found or in use, skip removing", source);
+                saveException = ex;
+                _logger.LogError(ex, "Unable to save image {Source}", source);
+            }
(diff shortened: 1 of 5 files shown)


Shadowghost (approved):

MediaBrowser.Providers/Manager/ProviderManager.cs
```diff
@@ -251,15 +251,23 @@ public Task SaveImage(BaseItem item, Stream source, string mimeType, ImageType t
...
             if (string.IsNullOrWhiteSpace(source))
             {
                 throw new ArgumentNullException(nameof(source));
             }
 
             var fileStream = AsyncFile.OpenRead(source);
-            return new ImageSaver(_configurationManager, _libraryMonitor, _fileSystem, _logger).SaveImage(item, fileStream, mimeType, type, imageIndex, saveLocallyWithMedia, cancellationToken);
+            await new ImageSaver(_configurationManager, _libraryMonitor, _fileSystem, _logger).SaveImage(item, fileStream, mimeType, type, imageIndex, saveLocallyWithMedia, cancellationToken);
+            try
+            {
```
JPVenson, line 264:
```suggestion
         try
         {
            await new ImageSaver(_configurationManager, _libraryMonitor, _fileSystem, _logger).SaveImage(item, fileStream, mimeType, type, imageIndex, saveLocallyWithMedia, cancellationToken);
   
```

JPVenson (approved):

MediaBrowser.Providers/Manager/ProviderManager.cs
gnattu: Isn't throwing the exception more close to the old behavior?

JPVenson: I figure there are other things that might go wrong in the SaveImage method (external connection and stuff) and we would still cleanup even in the even of an error or?

JPVenson: oh i see the confusion

gnattu: The old behavior would be remove the temp file after saving to metadata folder. If the save failed just throw, at least that was the old implementation in `BaseDynamicImageProvider`. This PR made that also available for `ItemImageProvider`. From current implementation in both use cases the source files are temp files so should be safe enough to remove, but we still want to throw the exception so that the task will properly fail.

JPVenson: Ok i see that, but still in the case the SaveImage method throws, we still would want to delete the temp file or? So it should rather be:

```csharp
 	 try
	 {
	    await new ImageSaver(_configurationManager, _libraryMonitor, _fileSystem, _logger).SaveImage(item, fileStream, mimeType, type, imageIndex, saveLocallyWithMedia, cancellationToken);
            
      }
      catch (IOException ex)
      {
          _logger.LogError(ex, "Source file {Source} not found or in use, skip removing", source);
      }
      finally
      {
      	  if (File.Exists(source))
      	  {
      	  	      File.Delete(source);
      	  }
      }
```

If the delete fails, there aint much we can do anyway.

MediaBrowser.Providers/Manager/ProviderManager.cs
```diff
@@ -251,15 +251,45 @@ public Task SaveImage(BaseItem item, Stream source, string mimeType, ImageType t
...
+            Exception? saveException = null;
+
+            try
+            {
+                var fileStream = AsyncFile.OpenRead(source);
+                await new ImageSaver(_configurationManager, _libraryMonitor, _fileSystem, _logger).SaveImage(item, fileStream, mimeType, type, imageIndex, saveLocallyWithMedia, cancellationToken);
+            }
+            catch (Exception ex)
+            {
+                saveException = ex;
```
JPVenson, line 270: Use [ExceptionDispatchInfo](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.exceptionservices.exceptiondispatchinfo?view=net-8.0).Capture to preserve stacktrace

JPVenson (changes requested):

JPVenson (approved):

MediaBrowser.Providers/Manager/ProviderManager.cs
```diff
@@ -251,15 +252,42 @@ public Task SaveImage(BaseItem item, Stream source, string mimeType, ImageType t
...
+            ExceptionDispatchInfo? saveException = null;
+
+            try
+            {
+                var fileStream = AsyncFile.OpenRead(source);
+                await new ImageSaver(_configurationManager, _libraryMonitor, _fileSystem, _logger).SaveImage(item, fileStream, mimeType, type, imageIndex, saveLocallyWithMedia, cancellationToken);
+            }
+            catch (Exception ex)
+            {
+                saveException = ExceptionDispatchInfo.Capture(ex);
```
cvium, line 271: Why not just let it throw? The finally block will get executed regardless. If `File.Delete` throws a non-IOException, the exception by ImageSaver will get swallowed

gnattu: I was not quite confident with the behavior of just throw without catching for such nested try catch. So we can still see if it is from the `ImageSaver` or from the `File.Delete` right? Because if the `ImageSaver` failed for some reason the `File.Delete` would also have a high chance to fail as well, we need to tell which one caused the failure.

cvium (approved):

PR #12941, opened 2024-11-02
Add API support for ELRC word-based lyrics

AlexDalas:
Adds support for word-based timestamps from within ELRC files. This adds to the lyrics support, as the current lyrics support does not support word-by-word lyrics in an ELRC file Instead, it just ignores it. Adding this would mean that a client could read the TimeTags and get word-by-word timings for lyrics.

I've created [a fork of Jellyfin-web](https://github.com/AlexDalas/jellyfin-web-elrc-lyrics) that works with the lyrics from the backend, but it is a bit buggy at the moment because of where the code for searching the word is ran.

**Changes**
- Adds new parameter to LyricLine > timetags - "Gets the Enhanced LRC timestamps for the song."
- Adds new entry to "Audio/id/Lyrics" page with a dictionary that has "Section (letters): time". 
          - For example "4: 6000" would be 6 seconds in and 4 letters into the lyrics.

**Issues**
No issues, but related: Feishin (which works with Jellyfin) shows [this](https://github.com/jeffvli/feishin/issues/555) when you use ELRC files with Navidrome. It does not show this on Jellyfin, but it doesn't time the word lyrics either.


Shadowghost:
I wonder if it would make more sense to add this to the upstream library

AlexDalas:
The data is somewhat there already, but it isn't sent to the client. I think it should be sent to the client so that you can have word-timed lyrics like you do in other apps

crobibero:
Instead of using a generic dictionary I would prefer creating a new `TimeTag` object with the properties specified

AlexDalas:
I've just updated the code
- It uses a new object - TimeTag, which holds Position and Timestamp.
- It is also in the same time format as the media player did, earlier it was returning values shorter than the media player reports in the API (ex. timestamp 11987 goes to 119870000, or whatever is in line with the media player)

Shadowghost (approved):

MediaBrowser.Model/Lyrics/LyricLine.cs
```diff
@@ -25,4 +30,9 @@ public LyricLine(string text, long? start = null)
     /// Gets the start time in ticks.
     /// </summary>
     public long? Start { get; }
+
+    /// <summary>
+    /// Gets the Enhanced LRC timestamps for the song.
+    /// </summary>
+    public List<TimeTag>? TimeTags { get; }
```
crobibero, line 37:
```suggestion
    public IReadOnlyList<TimeTag>? TimeTags { get; }
```

MediaBrowser.Model/Lyrics/TimeTag.cs
```diff
@@ -0,0 +1,28 @@
...
+
+    /// <summary>
+    /// Gets the character index of the lyric.
+    /// </summary>
+    public int Position { get; }
+
+    /// <summary>
+    /// Gets the timestamp the lyric is synced to.
+    /// </summary>
+    public int? Timestamp { get; }
```
crobibero, line 27: When could `Timestamp` be null?

AlexDalas: That was a complete oversight, I'll fix that now

MediaBrowser.Model/Lyrics/LyricLine.cs
```diff
@@ -1,19 +1,24 @@
...
 public class LyricLine
 {
     /// <summary>
     /// Initializes a new instance of the <see cref="LyricLine"/> class.
     /// </summary>
     /// <param name="text">The lyric text.</param>
     /// <param name="start">The lyric start time in ticks.</param>
-    public LyricLine(string text, long? start = null)
+    /// <param name="timetags">The Enhanced LRC timestamps for the song.</param>
+    public LyricLine(string text, long? start = null, List<TimeTag>? timetags = null)
```
crobibero, line 17:
```suggestion
    public LyricLine(string text, long? start = null, IReadOnlyList<TimeTag>? timetags = null)
```

MediaBrowser.Model/Lyrics/LyricLine.cs
```diff
@@ -1,19 +1,24 @@
+using System.Collections.Generic;
+
 namespace MediaBrowser.Model.Lyrics;
 
 /// <summary>
 /// Lyric model.
 /// </summary>
+///
```
crobibero, line 8:
```suggestion
```

MediaBrowser.Model/Lyrics/TimeTag.cs
```diff
@@ -0,0 +1,28 @@
...
+        Timestamp = timestamp;
+    }
+
+    /// <summary>
+    /// Gets the character index of the lyric.
+    /// </summary>
+    public int Position { get; }
+
+    /// <summary>
+    /// Gets the timestamp the lyric is synced to.
```
crobibero, line 25: What unit is the the timestamp in? (looks like it might be ticks?)

MediaBrowser.Model/Lyrics/TimeTag.cs
```diff
@@ -0,0 +1,28 @@
...
+/// TimeTag model.
+/// </summary>
+public class TimeTag
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="TimeTag"/> class.
+    /// </summary>
+    /// <param name="position">The character index of the lyric.</param>
+    /// <param name="timestamp">The timestamp the lyric is synced to.</param>
+    public TimeTag(int position, int timestamp = null)
```
crobibero, line 13:
```suggestion
    public TimeTag(int position, int timestamp)
```

MediaBrowser.Providers/Lyric/LrcLyricParser.cs
```diff
@@ -68,8 +68,11 @@ public LrcLyricParser()
 
         for (int i = 0; i < sortedLyricData.Count; i++)
         {
+            List<TimeTag> timeTags = sortedLyricData[i].TimeTags
+                .Select(kvp => new TimeTag(position: kvp.Key.Index + (kvp.Key.Index == -1 ? 1 : 0), timestamp: kvp.Value * 10000)).ToList();
```
crobibero, line 72: I find this hard to parse, maybe change it to be 

```c#
position: Math.Max(kvp.Key.Index, 0)
```

Would also change this to only create the list of there are any TimeTags on the parsed data

```c#
List<TimeTag>? timeTags = null;
if(sortedLyricData[i].TimeTags.Count != 0)
{
	// TODO build list here
}
```

MediaBrowser.Model/Lyrics/TimeTag.cs
AlexDalas: Yeah, it's in ticks.

AlexDalas: I have changed the timestamp object, so it's now a _long_ type instead of an _int_, as this is what the other ticks do.

MediaBrowser.Providers/Lyric/LrcLyricParser.cs
AlexDalas: I've changed it to use that instead

MediaBrowser.Model/Lyrics/TimeTag.cs
AlexDalas: I have also updated the description to say "in ticks'

crobibero (approved):

MediaBrowser.Model/Lyrics/TimeTag.cs
```diff
@@ -0,0 +1,28 @@
+namespace MediaBrowser.Model.Lyrics;
+
+/// <summary>
+/// TimeTag model.
+/// </summary>
+public class TimeTag
```
nielsvanvelzen, line 6: Should be more descriptive, at least add the "lyric" keyword to it.

```suggestion
public class LyricLineTimeTag
```

MediaBrowser.Model/Lyrics/TimeTag.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="TimeTag"/> class.
+    /// </summary>
+    /// <param name="position">The character index of the lyric.</param>
+    /// <param name="timestamp">The timestamp the lyric is synced to in ticks.</param>
+    public TimeTag(int position, long timestamp)
```
nielsvanvelzen, line 13: The lyric line uses "Start" so why do we use "Timestamp" here? That's not really consistent

nielsvanvelzen, line 13: The Position refers to the start index but there is no end index or size/width

MediaBrowser.Model/Lyrics/LyricLine.cs
```diff
@@ -25,4 +29,9 @@ public LyricLine(string text, long? start = null)
     /// Gets the start time in ticks.
     /// </summary>
     public long? Start { get; }
+
+    /// <summary>
+    /// Gets the Enhanced LRC timestamps for the song.
+    /// </summary>
+    public IReadOnlyList<TimeTag>? TimeTags { get; }
```
nielsvanvelzen, line 36: Maybe call it Cues instead?

MediaBrowser.Model/Lyrics/LyricLine.cs
```diff
@@ -25,4 +29,9 @@ public LyricLine(string text, long? start = null)
     /// Gets the start time in ticks.
     /// </summary>
     public long? Start { get; }
+
+    /// <summary>
+    /// Gets the Enhanced LRC timestamps for the song.
```
nielsvanvelzen, line 34: The lyric data structure is format independent and should **not** be made for one specific format.

nielsvanvelzen (changes requested):

MediaBrowser.Model/Lyrics/TimeTag.cs
AlexDalas: Would it be best for me to add an end index or a size/width entry? The LRC library only reports times that the TimeTags are placed in there, so I'm pretty sure I would need to get the start of the next index to get either the size or the end index.

AlexDalas:
I've just pushed changes based on the comments that were made. I'm not sure about a couple of things:

- Details about the end index of a line aren't in these files, so I'm not exactly sure how to go about adding the "end" index (My only idea is to use the start of the next line as the end of the current line). 
- I think there is an "end" timestamp in addition to the "start" timestamp. Should I include that as well in the LyricLineTimeTag?

MediaBrowser.Model/Lyrics/LyricLineTimeTag.cs
```diff
@@ -0,0 +1,28 @@
+namespace MediaBrowser.Model.Lyrics;
+
+/// <summary>
+/// LyricLineTimeTag model.
+/// </summary>
+public class LyricLineTimeTag
```
nielsvanvelzen, line 6: Would make sense to use "Cue" instead of "Tag" for consistency with the property name in LyricLine

```suggestion
public class LyricLineCue
```

MediaBrowser.Model/Lyrics/LyricLineTimeTag.cs
```diff
@@ -0,0 +1,28 @@
+namespace MediaBrowser.Model.Lyrics;
+
+/// <summary>
+/// LyricLineTimeTag model.
```
nielsvanvelzen, line 4: Not really descriptive. Should at least give some idea of what this is used for.

MediaBrowser.Model/Lyrics/TimeTag.cs
nielsvanvelzen: From a client perspective I want to know the start & end position of each cue to apply styling to it. So an ""Width"/"Length"/"Size" property would be useful to have, that way if we ever add support for lyric formats that have cues for some sections but not all in a line we don't need to change the API again.

nielsvanvelzen (changes requested):

AlexDalas:
I've made changes to the code:
- Includes data about "end" and "width"
- LyricLineTimeTag changed -> LyricLineCue

MediaBrowser.Model/Lyrics/LyricLine.cs
```diff
@@ -25,4 +29,9 @@ public LyricLine(string text, long? start = null)
     /// Gets the start time in ticks.
     /// </summary>
     public long? Start { get; }
+
+    /// <summary>
+    /// Gets the time-aligned cues for the song's lyrics.
+    /// </summary>
+    public IReadOnlyList<LyricLineCue>? Cues { get; }
```
nielsvanvelzen, line 36: Make sure to always set this value (so empty list by default), but keep the nullability for backwards compatibility. Then we can make this not-null in a later API version.

MediaBrowser.Model/Lyrics/LyricLineCue.cs
```diff
@@ -0,0 +1,42 @@
...
+public class LyricLineCue
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="LyricLineCue"/> class.
+    /// </summary>
+    /// <param name="position">The start of the character index of the lyric.</param>
+    /// <param name="width">The count of the characters in the lyric.</param>
+    /// <param name="start">The start of the timestamp the lyric is synced to in ticks.</param>
+    /// <param name="end">The end of the timestamp the lyric is synced to in ticks.</param>
+    public LyricLineCue(int position, int width, long start, long end)
```
nielsvanvelzen, line 15: I think adding both width and end may be a bit redundant, let's choose one and stick with that. I'm not sure if we have other places in the API that we use as reference.

nielsvanvelzen (changes requested):

MediaBrowser.Model/Lyrics/LyricLineCue.cs
AlexDalas: I've decided to add "end" because it's more clear on exactly where the end point is, and is consistent with "start". However, if we need to use width anyway, the code for it is in the repo history anyway

MediaBrowser.Model/Lyrics/LyricLine.cs
AlexDalas: I've changed the code later to not include "null" for this object type

nielsvanvelzen (approved):
✅  API changes look good

crobibero (dismissed):
This throws an exception when attempting to parse file contents.
Since this is adding actual logic to the LrcLyricParser I think we should have a test to make sure it works as intended

```
System.ArgumentOutOfRangeException
Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
   at System.Collections.Generic.List`1.get_Item(Int32 index)
   at MediaBrowser.Providers.Lyric.LrcLyricParser.ParseLyrics(LyricFile lyrics) in /home/crobibero/Code/jellyfin/MediaBrowser.Providers/Lyric/LrcLyricParser.cs:line 82
   at Jellyfin.Providers.Tests.Lyrics.LrcLyricParserTests.ParseElrc() in /home/crobibero/Code/jellyfin/tests/Jellyfin.Providers.Tests/Lyrics/LrcLyricParserTests.cs:line 15
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```

crobibero:
I changed the API so that the last lyric cue will have a `null` end, hopefully that is acceptable to the group

Shadowghost (approved):

crobibero:
While this does add functionality to the API I think it should still be merged for 10.11 because it also fixes the text display after parsing word-based lyrics

Shadowghost:
I do agree, especially because it is backwards compatible too.

nielsvanvelzen (approved):
API changes are still fine. Since this is only additions to a response model it is not a breaking change and thus fine to include in 10.11. 

Clients can choose to ignore the new fields and keep working as expected.

Chaphasilor:
I'll try to add support for this in Finamp once this PR is merged :)

MediaBrowser.Model/Lyrics/LyricLineCue.cs
Maxr1998: @nielsvanvelzen I just started working on word-synced lyrics for Finamp, and honestly missed that width field. I think there was some confusion because position and width would have been string indices, whereas start and end are timestamps, so they aren't actually redundant.

PR #12970, opened 2024-11-05
Add ability to remove a ChannelMapping

Giermann:
Remove a ChannelMapping by selecting the same mapping again. This should be an intuitive way to de-select a mapping which currently requires the manual editing of a config file: https://forum.jellyfin.org/t-how-to-unmap-livetv-channels

**Changes**

**Issues**


Shadowghost (approved):
Seems like a reasonable change to me. But since I don't have any LiveTV to test with, I'd appreciate someone testing with a actual working setup 😄

theguymadmax:
I tested this, and it successfully removes the channel mapping as described.

src/Jellyfin.LiveTv/Listings/ListingsManager.cs
```diff
@@ -230,10 +230,15 @@ public async Task<TunerChannelMapping> SetChannelMapping(string providerId, stri
         var listingsProviderInfo = config.ListingProviders
             .First(info => string.Equals(providerId, info.Id, StringComparison.OrdinalIgnoreCase));
 
+        var existingChannelMapping = listingsProviderInfo.ChannelMappings
+            .Count(pair => string.Equals(pair.Name, tunerChannelNumber, StringComparison.OrdinalIgnoreCase)
```
Bond-009, line 234:
```suggestion
        var channelMappingExists = listingsProviderInfo.ChannelMappings
            .Any(pair => string.Equals(pair.Name, tunerChannelNumber, StringComparison.OrdinalIgnoreCase)
```

src/Jellyfin.LiveTv/Listings/ListingsManager.cs
```diff
@@ -230,10 +230,15 @@ public async Task<TunerChannelMapping> SetChannelMapping(string providerId, stri
...
+                        && string.Equals(pair.Value, providerChannelNumber, StringComparison.OrdinalIgnoreCase));
+
         listingsProviderInfo.ChannelMappings = listingsProviderInfo.ChannelMappings
             .Where(pair => !string.Equals(pair.Name, tunerChannelNumber, StringComparison.OrdinalIgnoreCase)).ToArray();
 
-        if (!string.Equals(tunerChannelNumber, providerChannelNumber, StringComparison.OrdinalIgnoreCase))
+        if (!string.Equals(tunerChannelNumber, providerChannelNumber, StringComparison.OrdinalIgnoreCase)
+            && existingChannelMapping < 1)
```
Bond-009, line 241:
```suggestion
            && !channelMappingExists)
```

Bond-009 (changes requested):
The exact count doesn't matter, just check if it exists

Giermann:
Is there anything else for me to do?

Shadowghost:
The comments from Bond are not addressed, so at least those you should take care of 😄

Bond-009 (approved):

PR #12973, opened 2024-11-06
Fix pixel format in HEVC RExt SDR transcoding

nyanmisaka:
**Changes**
- Fix pixel format in HEVC RExt SDR transcoding

**Issues**
- Found in https://github.com/jellyfin/jellyfin/issues/12657#issuecomment-2458228878

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -4131,7 +4131,7 @@ namespace MediaBrowser.Controller.MediaEncoding
             else if (isD3d11vaDecoder || isQsvDecoder)
             {
                 var isRext = IsVideoStreamHevcRext(state);
-                var twoPassVppTonemap = isRext;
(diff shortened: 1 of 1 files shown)


Shadowghost (approved):

gnattu:
I'm not sure if that will fix that user's problem if at all because the overlay filter does accept the p010 inputs and the encoder also accepts the output and finishes the encoding, although the result video is in 8bit using the command provided.

If the N100 capability is cut down compared to UHD770 it really looks like a driver issue to me.

nyanmisaka:
> I'm not sure if that will fix that user's problem if at all because the overlay filter does accept the p010 inputs and the encoder also accepts the output and finishes the encoding, although the result video is in 8bit using the command provided.
> 
> If the N100 capability is cut down compared to UHD770 it really looks like a driver issue to me.

This is just an unrelated issue found in that log. The actual error in the log should be caused by the driver. We need report to Intel if updating the driver does not solve the problem.

crobibero (approved):

gnattu (approved):

PR #12975, opened 2024-11-06
FreeBSD behave the same as Linux concerning hardware acceleration

bapt:
**Changes**
Add support for FreeBSD (which the same as Linux) when dealing with how to setup hardware acceleration for the encoder.

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -894,7 +894,7 @@ namespace MediaBrowser.Controller.MediaEncoding
         private string GetQsvDeviceArgs(string renderNodePath, string alias)
         {
             var arg = " -init_hw_device qsv=" + (alias ?? QsvAlias);
-            if (OperatingSystem.IsLinux())
+            if (OperatingSystem.IsLinux() || OperatingSystem.IsFreeBSD())
             {
                 // derive qsv from vaapi device
                 return GetVaapiDeviceArgs(renderNodePath, "iHD", "i915", "0x8086", null, VaapiAlias) + arg + "@" + VaapiAlias;
@@ -970,7 +970,7 @@ namespace MediaBrowser.Controller.MediaEncoding
 
             var args = new StringBuilder();
             var isWindows = OperatingSystem.IsWindows();
-            var isLinux = OperatingSystem.IsLinux();
+            var isLinux = OperatingSystem.IsLinux() || OperatingSystem.IsFreeBSD();
             var isMacOS = OperatingSystem.IsMacOS();
             var optHwaccelType = options.HardwareAccelerationType;
             var vidDecoder = GetHardwareVideoDecoder(state, options) ?? string.Empty;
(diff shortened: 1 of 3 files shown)


MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -1807,13 +1807,13 @@ public string GetVideoQualityParam(EncodingJobInfo state, string videoEncoder, E
...
             {
-                if (OperatingSystem.IsLinux())
+                if (OperatingSystem.IsLinux() || OperatingSystem.IsFreeBSD())
                 {
                     var ver = Environment.OSVersion.Version;
                     var isFixedKernel60 = ver.Major == 6 && ver.Minor == 0 && ver >= _minFixedKernel60i915Hang;
                     var isUnaffectedKernel = ver < _minKerneli915Hang || ver > _maxKerneli915Hang;
 
-                    if (!(isUnaffectedKernel || isFixedKernel60))
+                    if (!(isUnaffectedKernel || isFixedKernel60 || OperatingSystem.IsFreeBSD()))
```
JPVenson, line 1816: can you explain this change please?

nyanmisaka:
I'm not sure if all the components needed for hardware transcoding have been ported to FreeBSD and verified. For example:

- [i915](https://github.com/torvalds/linux/tree/master/drivers/gpu/drm/i915) kernel module
- i965/[intel-vaapi-driver](https://github.com/intel/intel-vaapi-driver)
- iHD/[media-driver](https://github.com/intel/media-driver)
- MSDK/[MediaSDK](https://github.com/Intel-Media-SDK/MediaSDK)
- VPL/[vpl-gpu-rt](https://github.com/intel/vpl-gpu-rt)
- Mesa [radeonsi](https://gitlab.freedesktop.org/mesa/mesa/-/tree/main/src/gallium/drivers/radeonsi) and [RADV](https://gitlab.freedesktop.org/mesa/mesa/-/tree/main/src/amd/vulkan)
- closed-source NVIDIA cuvid/nvdec/nvenc/cuda drivers

Assuming they work, you'll also need to build [jellyfin-ffmpeg](https://github.com/jellyfin/jellyfin-ffmpeg) for FreeBSD that includes [patches for our custom functionality](https://github.com/jellyfin/jellyfin-ffmpeg/tree/jellyfin/debian/patches) to get full hardware acceleration working.

gnattu:
> I'm not sure if all the components needed for hardware transcoding have been ported to FreeBSD and verified. For example:
> 
> * [i915](https://github.com/torvalds/linux/tree/master/drivers/gpu/drm/i915) kernel module
> * i965/[intel-vaapi-driver](https://github.com/intel/intel-vaapi-driver)
> * iHD/[media-driver](https://github.com/intel/media-driver)
> * MSDK/[MediaSDK](https://github.com/Intel-Media-SDK/MediaSDK)
> * VPL/[vpl-gpu-rt](https://github.com/intel/vpl-gpu-rt)
> * Intel OpenCL/[compute-runtime](https://github.com/intel/compute-runtime)
> * Mesa [radeonsi](https://gitlab.freedesktop.org/mesa/mesa/-/tree/main/src/gallium/drivers/radeonsi) and [RADV](https://gitlab.freedesktop.org/mesa/mesa/-/tree/main/src/amd/vulkan)
> * closed-source NVIDIA cuvid/nvdec/nvenc/cuda drivers
> 
> Assuming they work, you'll also need to build [jellyfin-ffmpeg](https://github.com/jellyfin/jellyfin-ffmpeg) for FreeBSD that includes [patches for our custom functionality](https://github.com/jellyfin/jellyfin-ffmpeg/tree/jellyfin/debian/patches) to get full hardware acceleration working.

FreeBSD has libva dummy driver which exposes vaapi compatible APIs and a `libva-intel-media-driver` so that the qsv decoder/encoders can be used. Anything other than these two are not verified yet or at least there is no public source indicating where it would work or not. 

I do think the FreeBSD should have its own code path if we want it to be upstreamed because I don't even know if the freebsd libdrm could interoperate with the filters like on Linux and probably it will need frame copies or even perform the filter on CPU. Just marking it work in the same way as Linux is a bit too hacky to me.

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
bapt: my understanding of this part is this is workarounding an issue with the linux kernel driver at some point concerning the drm drivers, the FreeBSD version of the drm drivers are the drm drivers from 6.1-LTS so they should not be affected

Thefrank:
Looking back at my own very-hacky patch, just adding FreeBSD to the `isLinux` is a much easier approach than my building of a whole `var isFreeBSD` case. Very unnecessary in retrospect.

This comes at a perfect time as I was getting frustrated trying to work around `sass-embedded` not working for FreeBSD

So after almost 4 years, I can now drop my jellyfin repo :)

@bapt Thank you!

bapt:
I will just keep this patch locally to the FreeBSD package system.

PR #12989, opened 2024-11-08
Fix InvariantCulture in VPP tonemap options

nyanmisaka:
**Changes**
- Fix InvariantCulture in VPP tonemap options

**Issues**
- Fixes #12988

PR #12991, opened 2024-11-08
Use invariant culture for tonemap options

gnattu:
Those values that can have decimals need to explicitly use invariant culture for stringification, otherwise the user locale may use the decimal point that ffmpeg does not support.

**Changes**

**Issues**

Related: #12989


Bond-009 (dismissed):
Use `string.Format(CultureInfo.InvariantCulture, ...`
See https://github.com/jellyfin/jellyfin/pull/12989#pullrequestreview-2424671798

crobibero (approved):

PR #13023, opened 2024-11-12
Update projects to .NET 9

Bond-009:


.devcontainer/devcontainer.json
```diff
@@ -1,15 +1,15 @@
 {
     "name": "Development Jellyfin Server",
-    "image":"mcr.microsoft.com/devcontainers/dotnet:8.0-jammy",
+    "image":"mcr.microsoft.com/devcontainers/dotnet:9.0-jammy",
```
Gauvino, line 3: Same as above

Jellyfin.Server/Helpers/StartupHelpers.cs
```diff
@@ -292,13 +292,5 @@ public static void PerformStaticInitialization()
...
         // Ref: https://docs.microsoft.com/en-us/dotnet/api/system.text.codepagesencodingprovider.instance?view=netcore-3.0#remarks
         Encoding.RegisterProvider(CodePagesEncodingProvider.Instance);
-
-        // Increase the max http request limit
-        // The default connection limit is 10 for ASP.NET hosted applications and 2 for all others.
-        ServicePointManager.DefaultConnectionLimit = Math.Max(96, ServicePointManager.DefaultConnectionLimit);
-
-        // Disable the "Expect: 100-Continue" header by default
-        // http://stackoverflow.com/questions/566437/http-post-returns-the-error-417-expectation-failed-c
-        ServicePointManager.Expect100Continue = false;
```
crobibero, line 302: Are these just no longer needed?

JPVenson (approved):

Jellyfin.Server/Helpers/StartupHelpers.cs
Bond-009: > In modern .NET, HttpClient does not take into account any configurations set on ServicePointManager.

Not sure how long this has been the case

Shadowghost (approved):

crobibero (approved):

zent1n0:
Latest build still uses dotnet 8.0. Is this by design or just an accident that the commit is not merged into release?

crobibero:
> Latest build still uses dotnet 8.0. Is this by design or just an accident that the commit is not merged into release?

10.10.x is net8.0, 10.11.x will be net9.0.

PR #13051, opened 2024-11-17
Fixed provider filter enabled

JPVenson:
Disabled provider disabled check 

ONLY MERGE FOR HOTFIX


JPVenson:
@joshuaboniface potential hotfix candidate for 10.10.3

nielsvanvelzen:
Isn't `DisabledMediaSegmentProviders` always an empty array because there is no way to set it in jellyfin-web?

Issue #13052, opened 2024-11-17
Some h264 videos (encoding dependent) when streamed by remuxing play poorly: wrong framerate, dropped frames; ffmpeg logs warnings "Non-monotonic DTS ... This may result in incorrect timestamps in the output file."


tsuereth:
### This issue respects the following points:

- [x] This is a **bug**, not a question or a configuration issue; Please visit our forum or chat rooms first to troubleshoot with volunteers, before creating a report. The links can be found [here](https://jellyfin.org/contact/).
- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [x] I'm using an up to date version of Jellyfin Server stable, unstable or master; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Description of the bug

Select video files, which are encoded in h264 - and stream to the `jellyfin-web` player by remuxing (not transcoding) - play back poorly, at a visibly lower framerate than the original video.  In some cases so many frames are dropped that the video may even appear to freeze, although the stream is still running and other channels (audio and subtitles) are fine.

When this occurs, the Jellyfin server's log file for the stream, `FFmpeg.Remux-*` will include frequent warnings from the ffmpeg command performing the remux:

> [vost#0:0/copy @ 0x56517bc84400] Non-monotonic DTS; previous: -1001, current: -1001; changing to -1000. This may result in incorrect timestamps in the output file.
[vost#0:0/copy @ 0x56517bc84400] Non-monotonic DTS; previous: 2002, current: 2002; changing to 2003. This may result in incorrect timestamps in the output file.
[vost#0:0/copy @ 0x56517bc84400] Non-monotonic DTS; previous: 6006, current: 6006; changing to 6007. This may result in incorrect timestamps in the output file.
[vost#0:0/copy @ 0x56517bc84400] Non-monotonic DTS; previous: 9009, current: 9009; changing to 9010. This may result in incorrect timestamps in the output file.
[vost#0:0/copy @ 0x56517bc84400] Non-monotonic DTS; previous: 13013, current: 13013; changing to 13014. This may result in incorrect timestamps in the output file.
...

[...]

PR #13053, opened 2024-11-17
Set `-copytb 1` in ffmpeg when creating an HLS stream, to fix "non-monotonic DTS" time breakage when remuxing.

tsuereth:
Set `-copytb 1` which "is sometimes required to avoid non monotonically increasing timestamps when copying video streams with variable frame rate."

**Does this adversely affect non-copy (transcode) streams?** ... is an excellent question. :)  In my limited testing, I don't think it does, though admittedly I don't have a ton of domain knowledge or relevant testing material.

**Changes**
This change simply adds `-copytb 1` to ffmpeg's command-line arguments when creating an HLS stream.  As ffmpeg documentation mentions: https://ffmpeg.org/ffmpeg.html

> -copytb mode
Specify how to set the encoder timebase when stream copying. mode is an integer numeric value, and can assume one of the following values:

> 1
Use the demuxer timebase.

> The time base is copied to the output encoder from the corresponding input demuxer. This is sometimes required to avoid non monotonically increasing timestamps when copying video streams with variable frame rate.

> 0
Use the decoder timebase.

> The time base is copied to the output encoder from the corresponding input decoder.

> -1
Try to make the choice automatically, in order to generate a sane output.

> Default value is -1.

Like the doc says, when a source video has "non monotonically increasing timestamps," this setting appears necessary to ensure that a remux operation outputs frames with the source's intended timings.

Otherwise an HLS remux stream will adjust those non-monotonic timestamps, and produce a video which appears to miss or skip frames.  This results in warnings logged by ffmpeg such as:

> [vost#0:0/copy @ 0x56517bc84400] Non-monotonic DTS; previous: -1001, current: -1001; changing to -1000. This may result in incorrect timestamps in the output file.
[vost#0:0/copy @ 0x56517bc84400] Non-monotonic DTS; previous: 2002, current: 2002; changing to 2003. This may result in incorrect timestamps in the output file.
[vost#0:0/copy @ 0x56517bc84400] Non-monotonic DTS; previous: 6006, current: 6006; changing to 6007. This may result in incorrect timestamps in the output file.
[vost#0:0/copy @ 0x56517bc84400] Non-monotonic DTS; previous: 9009, current: 9009; changing to 9010. This may result in incorrect timestamps in the output file.
[vost#0:0/copy @ 0x56517bc84400] Non-monotonic DTS; previous: 13013, current: 13013; changing to 13014. This may result in incorrect timestamps in the output file.
...

**Issues**
* https://github.com/jellyfin/jellyfin/issues/13052

### DIFF
--- a/Jellyfin.Api/Controllers/DynamicHlsController.cs
+++ b/Jellyfin.Api/Controllers/DynamicHlsController.cs
@@ -1645,7 +1645,7 @@ public class DynamicHlsController : BaseJellyfinApiController
 
         return string.Format(
             CultureInfo.InvariantCulture,
-            "{0} {1} -map_metadata -1 -map_chapters -1 -threads {2} {3} {4} {5} -copyts -avoid_negative_ts disabled -max_muxing_queue_size {6} -f hls -max_delay 5000000 -hls_time {7} -hls_segment_type {8} -start_number {9}{10} -hls_segment_filename \"{11}\" {12} -y \"{13}\"",
+            "{0} {1} -map_metadata -1 -map_chapters -1 -threads {2} {3} {4} {5} -copyts -avoid_negative_ts disabled -copytb 1 -max_muxing_queue_size {6} -f hls -max_delay 5000000 -hls_time {7} -hls_segment_type {8} -start_number {9}{10} -hls_segment_filename \"{11}\" {12} -y \"{13}\"",
             inputModifier,
             _encodingHelper.GetInputArgument(state, _encodingOptions, segmentContainer),
             threads,


gnattu:
In my opinion, it is not feasible for Jellyfin to address every single broken video. Furthermore, I am confident that blindly changing this flag will result in breaking other videos that require the use of mode 0. Maybe we can document this and ask users to fix the files instead.

tsuereth:
> blindly changing this flag will result in breaking other videos

Absolutely valid concern, the last thing I'd want to do is break playback for existing videos and users.

Is there any automated streaming test suite for checking Jellyfin's ffmpeg results?  Or a reference video library?  I don't know how to check for regressions aside from spot-checking my own collection.

Short of that, as a server admin and user I'd like a way to override the ffmpeg default in my own installation -- since finding and re-encoding all affected files would be non-trivial, as far as I currently know.

(I haven't yet discovered a way to identify affected files with ffprobe or other tools, before getting all the way to the ffmpeg remux command used by Jellyfin streaming.  Some updated info on affected files in https://github.com/jellyfin/jellyfin/issues/13052#issuecomment-2520523897 it seems like AMDVCE, possibly NVENC as well, is likely to produce videos with such non-monotonic decoding timestamps.)

I just updated this PR with a config setting to override ffmpeg's default `-copyts` option, confirmed with a local quick-and-dirty jellyfin-web `<select>`:

image

At a high level though, I still don't know if there's a better approach for "experimental" ffmpeg streaming changes.  What can be done to test ffmpeg option changes against known-good content?

gnattu:
> At a high level though, I still don't know if there's a better approach for "experimental" ffmpeg streaming changes. What can be done to test ffmpeg option changes against known-good content?

In the future I think we have to implement file based compatibility flags so that such workarounds can be applied to individual files. We are having a lot of user reports that some specific files wont work with remux and a lot of the time a forced transcoding is required. But I still don’t have a concrete plan on how to implement that especially to implement that in a user friendly way.

For current change, I don’t think we need to expose that to the API and make it configurable in the GUI. You can refer to what I did in #12744 to add a server-side env var flag to turn on this workaround mode, where we already have a few for FFmpeg behavior tweaking that we think most of our users should not touch but configurable for super advanced users, and we can remove this flag in the future if the per-file compatibility feature lands.

Issue #13048, opened 2024-11-16
Attempts to write file in read-only library during upgrade migration


kg333:
### This issue respects the following points:

- [x] This is a **bug**, not a question or a configuration issue; Please visit our forum or chat rooms first to troubleshoot with volunteers, before creating a report. The links can be found [here](https://jellyfin.org/contact/).
- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [x] I'm using an up to date version of Jellyfin Server stable, unstable or master; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Description of the bug

After installing the latest 10.10.2 package of Jellyfin on Ubuntu 22.04.5 LTS (10.10.2+ubu2204), Jellyfin fails to start with a bug attempting to access an m3u playlist in a read-only library.  Based on the migration description, it appears to be trying to delete it during migration `RemoveDuplicatePlaylistChildren`  - however, this is a read-only library and Jellyfin shouldn't be attempting to modify any files.  Jellyfin is able to read the file via `sudo -u jellyfin cat <path to m3u>`, so it does have correct read-only access.

### Reproduction steps

1. Set up Jellyfin 10.9.11.0 instance with read-only music library containing m3u playlists. (ex. /home/matthew/Music)
2. Install the updated Jellyfin package on Ubuntu 22.04.5 for version 10.10.2.
3. Attempt to start Jellyfin and observe failure to start when it attempts to access m3u files.

### What is the current _bug_ behavior?

Jellyfin fails to complete migration 'RemoveDuplicatePlaylistChildren" while attempting to delete an m3u it doesn't have write access to, and fails to start.

### What is the expected _correct_ behavior?

Jellyfin should not attempt cleanup or deletion of files in read-only libraries and should start.

### Jellyfin Server version

10.10.0+

### Specify commit id

Not known

### Specify unstable release number

Not known

### Specify version number

10.10.2+ubu2204

### Specify the build version

Unable to access dashboard, failure to start

### Environment

```markdown
- OS: Ubuntu 22.04.5 LTS Jammy
[...]

Issue #13057, opened 2024-11-17
Jellyfin stopped launching after 10.10.2 update when using read only library directories


lukaszsobala:
### This issue respects the following points:

- [x] This is a **bug**, not a question or a configuration issue; Please visit our forum or chat rooms first to troubleshoot with volunteers, before creating a report. The links can be found [here](https://jellyfin.org/contact/).
- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [x] I'm using an up to date version of Jellyfin Server stable, unstable or master; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Description of the bug

After upgrading from 10.10.1 to 10.10.2, jellyfin stopped launching because of an "exception" - read only directory. When I changed the mount point from `ro` to `rw` in `/etc/fstab`, it started working again.
But I do not think it should be necessary to have the libraries in a folder with read/write permissions for jellyfin. 

### Reproduction steps

1. Have libraries in a read only directory
2. `apt upgrade` from 10.10.1 to 10.10.2
3. Try starting jellyfin

### What is the current _bug_ behavior?

Jellyfin service fails to start

### What is the expected _correct_ behavior?

Jellyfin service starts

### Jellyfin Server version

10.10.0+

### Specify the build version

10.10.2

### Environment

```markdown
- OS: Ubuntu 24.04
- Linux Kernel: 6.1.0-1025-rockchip
- Hardware Acceleration: rkmpp
- GPU Model: Mali-G610
- Networking: ethernet
- Storage: library is on a samba share
```

### Jellyfin logs

```shell
Nov 17 10:09:17 rock5b jellyfin[7269]: System.IO.IOException: Read-only file system : '/mnt/samba_jellyfin/path/redacted/file.m3u'
```

### FFmpeg logs

```shell

```

Err0r0815:
Same Problem here but with Docker.
I have mounted multiple volumes mounted and only the volumes with the m3u Files generating the error.
All other volumes could be RO and the server is starting after RW the "m3u volume".

hope this helps.

Greeting

lukaszsobala:
[...]

PR #13059, opened 2024-11-17
Exclude file system based library playlists from migration

Shadowghost:
We can't update file system sourced playlists because they might be read only.

**Changes**
* Exclude file system based library playlists from migration
  * They are `OpenAccess` and have the default GUID as owner ID

**Issues**
#13057
#13048


sibbi77 (approved):
I verified, it fixes #13048

nielsvanvelzen (approved):

crobibero (approved):

PR #13038, opened 2024-11-15
Update dependencies

Bond-009:
Unsure about the z440.atl.core updates, I can drop them if needed.

PR #13063, opened 2024-11-18
Downgrade minimum sdk version

crobibero:
https://github.com/jellyfin/jellyfin/pull/13038#issuecomment-2481505391

We avoided updating global.json in the past for exactly this reason but attempted to fix codeql ci issues.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
     "sdk": {
-        "version": "8.0.404",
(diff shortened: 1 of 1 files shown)


mooninite:
Thank you! I'll patch Fedora for now so we can ship it.

Shadowghost (approved):

Issue #7427, opened 2022-03-07
[Issue]: MKV track languages are not displayed properly


shirt-dev:
### Please describe your bug

MKV track languages are not properly displayed in some cases, showing only a 3 letter code instead of the full language name.

Ideally, the localized language name should be shown instead of the ISO code.

### Jellyfin Version

10.7.7

### Environment

```markdown
- OS: Windows
- Virtualization: Docker
- Clients: Browser
- Browser: Chrome 99
- FFmpeg Version: 4.4.1-Jellyfin
- Playback Method: Remux
- Hardware Acceleration: NVENC
- Plugins: none
- Reverse Proxy: nginx
- Base URL: none
- Networking: Bridge
- Storage: local
```

### Please attach any screenshots here

![image](https://user-images.githubusercontent.com/2660574/157134705-ace2a9de-0733-447e-ae3c-a2b3b07e6c24.png)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

michaelkrieger:
Maybe related to #7330  ?

mikewesten:
I was wondering why English and Japanese and others are displaying properly, whereas French and Chinese and some others are not. That explains it. English and Japanese, for example, use the same code for all 3 ISO instances of 639-2/T, 639-2/B, and 639-3, whereas French and Chinese do not. Pic related [https://i.imgur.com/JOqfden.jpg](url)

shirt-dev:
Nuh uh
![nuh-uh.gif](https://github.com/jellyfin/jellyfin/assets/2660574/6c8397c1-03b1-44ea-89b1-1bce9a71c786)

guiweber:
This issue is not limited to MKVs, but is also present with external subtitle files (e.g. srt) and other video files such as mp4. 

Should another more general issue be opened? Or maybe this one should be updated to reflect this (and avoid duplicates)?

guiweber:
Probably also related to https://github.com/jellyfin/jellyfin/issues/6302

baka0815:
``Cze``, ``Fre`` and ``Ger`` are language codes from the ISO-639-2/B list, but the languages are read using ISO-639-2/T (see https://en.wikipedia.org/wiki/List_of_ISO_639-2_codes).
#13068 should fix this.

SThomasP:
I think the issue is here 
[...]

Issue #6302, opened 2021-07-15
Language code in subtitles problem with Chinese


mhqschen:
**Describe the bug**
There are two problems related with the language code for Chinese.

1. The language code in subtitles for Chinese can't be recognized.
XXX.chi.srt, won't be recognized as "Chinese" but "Chi". While XXX.spa.srt will be successfully phased as "Spanish" and XXX.en.srt will be successfully phased as "English" 
And after tried several common language code for Chinese, I found chi, zh, zho, zh-hk will all be recognized as "Chi"
On the other hand, ch, chn, chs, cht, cn, zh-cn, zh-cht are not able to be recognized but only shows them as subtitle name.

And this lead to the "Preferred subtitle language" setup not functional.

2. There are 3 option in language choose dropdown for Chinese.
"Chinese", "Chinese traditional" and "Chinese HongKong"
When I used the upload subtitle function in browser. No matter what I option choose, the uploaded subtitles will all be renamed to XXX.chi.srt.

**System (please complete the following information):**
 - OS: [e.g. Debian, Windows]
 - Virtualization: Docker
 - Clients: Browser
 - Browser: Microsoft Edge 91 (chromium)
 - Jellyfin Version: 10.7.6

**Expected behavior**
Although Chinese has many branches, the most well known and used ones are "Chinese Simplified" and "Chinese Traditional".
They are used in most mutil-language scenarios and used by Netflix as well.
Better to change from "Chinese", "Chinese traditional" and "Chinese HongKong" to "Chinese Simplified" and "Chinese Traditional".

And hope to have a clear list of language code on wiki page instructing people how to rename their subtitles. And Jellyfin can be able to recognize "Chinese Simplified" and "Chinese Traditional". 

If I can expect more, it would be wonderful to be able to fallback to other subtitles with "Chinese" in name when I choose "Preferred subtitle language" to"Chinese Simplified" but there isn't a subtitle in "Chinese Simplified" (same for "Chinese Traditional")

**Server Logs**
No error

**FFmpeg Logs**
No error

**Browser Console Logs**
No error

**Screenshots**
![image](https://user-images.githubusercontent.com/9138710/125746629-67bc895f-2e03-4f97-9661-4a9175cd82ae.png)
![image](https://user-images.githubusercontent.com/9138710/125746666-ea1082b3-a83c-4b51-833b-59b28d4fe5f1.png)
[...]

PR #13068, opened 2024-11-19
Translate the ISO-639-2/B codes to ISO-639-2/T.

### LINKED ISSUE #12821: Translation of stream language
The language of streams is - as far as I can tell - not properly translated.

I have subtitle streams which are displayed as "Englisch" (correct translation of en/eng/English) and some which are displayed as "Ger" which should be "Deutsch" (German).

Trying to find the source I found this line:
https://github.com/jellyfin/jellyfin/blob/0bc142de9acd01b8b9191b8e99da41fdebabe0a7/MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs#L716

Is the translation missing or am I looking at the wrong place?

baka0815:
Some languages differ in ISO-639-2/B and ISO-639-2/T. Those languages get not display properly, at least in the web UI (for example "German" is displayed as "Ger" because it's read as "ger" (ISO-639-2/B) and not "deu" (ISO-639-2/T)). With the addition of a small lookup table, those /B language codes get converted to /T language codes.
This enables 19 additional languages to be displayed correctly.

Before:
image
image

After:
image
image

**Changes**
Translate the ISO-639-2/B codes to ISO-639-2/T.

**Issues**
Fixes #12821


Shadowghost (approved):

crobibero:
Please use a FrozenDictionary and place it in LocalizationManger.

baka0815:
@crobibero I didn't use a dictionary because the usage of ``r.ThreeLetterISOLanguageName.Equals(Language, StringComparison.OrdinalIgnoreCase)`` suggests that the case is not guaranteed to be upper- or lowercase. If ``Language`` is always lower case, then I can use a dictionary for a quicker lookup.

crobibero:
Just specify the comparer when creating the dictionary like so: `.ToFrozenDictionary(StringComparer.OrdinalIgnoreCase)`

baka0815:
The only place where the transformation from /B to /T would be used is inside the ``MediaStream`` class, which currently has no dependancy on ``LocalizationManager``.
The translation itself is done via the .NET ``CultureInfo`` class and not the ``LocalizationManager``, so I'm not sure moving the dictionary to the ``LocalizationManger`` is reasonable.

JPVenson (approved):

photonconvergence:
This should also resolve #7427  and partially resolve #6302

MediaBrowser.Model/Entities/MediaStream.cs
```diff
@@ -33,6 +34,30 @@ public class MediaStream
...
+        // Transformation table from ISO-639-2/B to ISO-639-2/T
+        private static readonly FrozenDictionary<string, string> _iso6392BtoT = new KeyValuePair<string, string>[]
+        {
+            new("alb", "sqi"), // Albanian
+            new("arm", "hye"), // Armenian
+            new("baq", "eus"), // Basque
+            new("bur", "mya"), // Burmese
+            new("chi", "zho"), // Chinese
+            new("cze", "ces"), // Czech
+            new("dut", "nld"), // Dutch; Flemish
```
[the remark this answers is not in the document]
baka0815: Thanks, changed.

crobibero (dismissed):

crobibero:
The test `GetStreamScore_MediaStream_CorrectScore` is failing 

With the update that includes French I think we would need to also update iso6392.txt, and every user that has their preferred language set to French would need to resave (or write a migration for it) since it is saved as `fre` instead of `fra`

baka0815:
Ok, I got it now, sorry for the confusion.

I didn't want to change the property, just the display language. So using a local variable does the trick...

nielsvanvelzen:
Why do we need these values hardcoded in a dictionary when all this data is already in iso6392.txt?

baka0815:
I think the list is small enough, and I think there won't be many (any?) additional values. Reading from the text file would add much complexity for this. Or is the list already in some kind of structure in code?

cvium:
> I think the list is small enough, and I think there won't be many (any?) additional values. Reading from the text file would add much complexity for this. Or is the list already in some kind of structure in code?

https://github.com/jellyfin/jellyfin/blob/06c603428bceccdb793eb34cc0340e25552d6c1d/Emby.Server.Implementations/Localization/LocalizationManager.cs#L109

baka0815:
Ok, then I would need to get the list of cultures, iterate through the complete list (as it's not a dictionary) until I find an entry where ThreeLetterISOLanguageNames.Count > 1 and ThreeLetterISOLanguageNames[1] equals the language I'm searching.
That sounds not well optimized.

Can I add the functionality to the LocalizationManager then? As while loading the cultures would be the point where it's about free to check for the /B and /T variants and I could build the dictionary there to include all cases.
Then I would just need to know how to pass the LocalizationManager to the MediaStream.

baka0815:
DisplayTitle is called from DynamicHlsHelper and StreamInfo, both don't have a LocalizationManager.

MediaStream is initialized from BlurayDiscInfo, BdInfoExaminer, LiveTvMediaSourceProvider and other classes which also don't have a LocalizationManager.

So providing it in the constructor or converting DisplayTitle to a method with the LocalizationManager as a parameter would both mean a bit of a refactoring. Which way would be better or is there a third alternative I'm missing?

Changing the iso6932.txt from
```ger|deu|de|German|allemand```
to
```deu|ger|de|German|allemand```
seems to do nothing.

**Comment by baka0815** [comment]
Ok, so what to do with this? I see three options:

1. Merge as is. The list is small and there shouldn't be many (any) new additions
2. Change the LocalizationManager and make DisplayTitle a method with the manager as parameter. Change DynamicHlsHelper and StreamInfo on the way
3. Change the LocalizationManager and add it as a parameter to the constructor of MediaStream. Change BlurayDiscInfo, BdInfoExaminer, LiveTvMediaSourceProvider and others on the way

The change of the iso txt-file should also be done, I think. Currently the format is 0|1|2|3|4
```
where 0 and 1 are the three letter ISO 639 codes.
If 0 *and* 1 are set then 0 is the ISO 639-2/B code and 1 is the ISO 639-2/T (or ISO 639-3) code.
If only 0 is set then 1 is the ISO 639-2/T  (or ISO 639-3) code.
3 is the 2-letter ISO 639-1 code.
4 is the full language name in English.
5 is the full language name in French.
Examples:
fre|fra|fr|French|français
eng||en|English|anglais
```

@cvium @nielsvanvelzen @crobibero

**Comment by baka0815** [comment]
See #13314 for the changes in the column order of the iso6392.txt file.

**Comment by jhparker88** [comment]
> See https://github.com/jellyfin/jellyfin/pull/13314 for the changes in the column order of the iso6392.txt file.

Now that the mentioned changes got merged, @baka0815 does this PR require any update given that I had similar issues in the past with the subtitle language names.

**Comment by baka0815** [comment]
Now, as the changes of the ISO file are merged, it's guaranteed that ``ThreeLetterISOLanguageNames[0]`` is the ISO 639-2/T code we're needing, so it's possible to use the localization manager to get the language info and set the language to the correct code in the MediaStreamRepository.

This works in my case and has the same result as the previous implementation, however I'm unsure if this one location is the only location in the code where a change is needed.

Please check and advice.

File: Jellyfin.Server.Implementations/Item/MediaStreamRepository.cs [hunk 2]
```diff
@@ -100,7 +100,22 @@ private MediaStream Map(MediaStreamInfo entity)
...
+        // if yes choose the first as that is the ISO 639-2/T code we're needing
+        if (language != null)
+        {
+            var langInfo = _localization.FindLanguageInfo(language);
+            if (langInfo != null && langInfo.ThreeLetterISOLanguageNames.Count > 1)
+            {
+                language = langInfo.ThreeLetterISOLanguageNames[0];
+            }
+        }
+
```
jhparker88, line 116: Would it make more sense to move this into the LocalizationManager class and add a specific method (e.g. `GetFirstThreeLetterISOLanguageName ( language )`) and maybe a test for it?

baka0815: That's what I meant with the possible optimization.
I did the change to use a dictionary and load the info while loading the ISO txt file. This way removes the need to traverse the list and should be faster.

baka0815:
I'm waiting for feedback (@cvium @nielsvanvelzen @crobibero ?) before fixing the ABI/API difference.

crobibero:
The changes are looking much cleaner!
Only thing I'm not a fan of is calling the function `GetISO6392TFromB` because it will return the input if the ISO 639-2/T isn't available. Here's a patch that changes it to `TryGetIso6392TFromB` and also changes the `_iso6392BtoT` to a `FrozenDictionary`

I think it would still be beneficial to add some tests to prevent breakage in the future

```patch
Index: Jellyfin.Server.Implementations/Item/MediaStreamRepository.cs
IDEA additional info:
<+>UTF-8
diff --git a/Jellyfin.Server.Implementations/Item/MediaStreamRepository.cs b/Jellyfin.Server.Implementations/Item/MediaStreamRepository.cs
--- a/Jellyfin.Server.Implementations/Item/MediaStreamRepository.cs	(revision 4c2557e9eccd0225e154fee070f0c6199183616d)
+++ b/Jellyfin.Server.Implementations/Item/MediaStreamRepository.cs	(date 1743726060813)
@@ -105,9 +105,9 @@
 
         // Check if the language has multiple three letter ISO codes
         // if yes choose the first as that is the ISO 639-2/T code we're needing
-        if (language != null)
+        if (language != null && _localization.TryGetIso6392TFromB(language, out var isoT))
         {
-            language = _localization.GetISO6392TFromB(language);
+            language = isoT;
         }
 
         dto.Language = language;
Index: Emby.Server.Implementations/Localization/LocalizationManager.cs
IDEA additional info:
<+>UTF-8
diff --git a/Emby.Server.Implementations/Localization/LocalizationManager.cs b/Emby.Server.Implementations/Localization/LocalizationManager.cs
--- a/Emby.Server.Implementations/Localization/LocalizationManager.cs	(revision 4c2557e9eccd0225e154fee070f0c6199183616d)
+++ b/Emby.Server.Implementations/Localization/LocalizationManager.cs	(date 1743725942647)
@@ -1,6 +1,8 @@
 using System;
 using System.Collections.Concurrent;
+using System.Collections.Frozen;
+using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Linq;
 using System.Reflection;
@@ -38,7 +40,7 @@
 
         private List<CultureDto> _cultures = [];
 
-        private Dictionary<string, string> _iso6392BtoT = [];
+        private FrozenDictionary<string, string> _iso6392BtoT = null!;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="LocalizationManager" /> class.
@@ -155,7 +157,7 @@
                 }
 
                 _cultures = list;
-                _iso6392BtoT = iso6392BtoTdict;
+                _iso6392BtoT = iso6392BtoTdict.ToFrozenDictionary();
             }
         }
 
@@ -514,19 +516,17 @@
             yield return new LocalizationOption("廣東話 (香港)", "zh-HK");
         }
 
-        /// <summary>
-        /// Returns the language in ISO 639-2/T when the input is ISO 639-2/B.
-        /// </summary>
-        /// <param name="isoB">The language in ISO 639-2/B.</param>
-        /// <returns>The language in ISO 639-2/T.</returns>
-        public string GetISO6392TFromB(string isoB)
+        /// <inheritdoc />
+        public bool TryGetIso6392TFromB(string isoB, [NotNullWhen(true)] out string? isoT)
         {
             if (_iso6392BtoT.TryGetValue(isoB, out string? result) && !string.IsNullOrEmpty(result))
             {
-                return result;
+                isoT = result;
+                return true;
             }
 
-            return isoB;
+            isoT = null;
+            return false;
         }
     }
 }
Index: MediaBrowser.Model/Globalization/ILocalizationManager.cs
IDEA additional info:
<+>UTF-8
diff --git a/MediaBrowser.Model/Globalization/ILocalizationManager.cs b/MediaBrowser.Model/Globalization/ILocalizationManager.cs
--- a/MediaBrowser.Model/Globalization/ILocalizationManager.cs	(revision 4c2557e9eccd0225e154fee070f0c6199183616d)
+++ b/MediaBrowser.Model/Globalization/ILocalizationManager.cs	(date 1743725940812)
@@ -1,4 +1,5 @@
+using System.Diagnostics.CodeAnalysis;
 using MediaBrowser.Model.Entities;
 
 namespace MediaBrowser.Model.Globalization;
@@ -66,6 +67,7 @@
     /// Returns the language in ISO 639-2/T when the input is ISO 639-2/B.
     /// </summary>
     /// <param name="isoB">The language in ISO 639-2/B.</param>
-    /// <returns>The language in ISO 639-2/T.</returns>
-    public string GetISO6392TFromB(string isoB);
+    /// <param name="isoT">The language in ISO 639-2/T.</param>
+    /// <returns>Whether the language could be converted.</returns>
+    bool TryGetIso6392TFromB(string isoB, [NotNullWhen(true)] out string? isoT);
 }
```

baka0815:
@crobibero 
```
        public bool TryGetIso6392TFromB(string isoB, [NotNullWhen(true)] out string? isoT)
         {
             if (_iso6392BtoT.TryGetValue(isoB, out string? result) && !string.IsNullOrEmpty(result))
             {
                isoT = result;
                return true;
             }
 
            isoT = null;
            return false;
         }
```

Can't this be simplified as
```
        public bool TryGetIso6392TFromB(string isoB, [NotNullWhen(true)] out string? isoT)
         {
             isoT = null;
             return _iso6392BtoT.TryGetValue(isoB, isoT);
         }
```

baka0815:
> I think it would still be beneficial to add some tests to prevent breakage in the future

And I hope I'm not appearing ignorant, but where would I add those tests?

crobibero:
> @crobibero 
> 
> ```c#
> 
>         public bool TryGetIso6392TFromB(string isoB, [NotNullWhen(true)] out string? isoT)
> 
>          {
> 
>              if (_iso6392BtoT.TryGetValue(isoB, out string? result) && !string.IsNullOrEmpty(result))
> 
>              {
> 
>                 isoT = result;
> 
>                 return true;
> 
>              }
> 
>  
> 
>             isoT = null;
> 
>             return false;
> 
>          }
> 
> ```
> 
> 
> 
> Can't this be simplified as
> 
> ```c#
> 
>         public bool TryGetIso6392TFromB(string isoB, [NotNullWhen(true)] out string? isoT)
> 
>          {
> 
>              isoT = null;
> 
>              return _iso6392BtoT.TryGetValue(isoB, isoT);
> 
>          }
> 
> ```
> 
> 

Likely wouldn't even need to set isoT to null before- but then you lose the IsNullOrEmpty check

crobibero:
> > I think it would still be beneficial to add some tests to prevent breakage in the future
> 
> 
> 
> And I hope I'm not appearing ignorant, but where would I add those tests?

https://github.com/jellyfin/jellyfin/blob/master/tests/Jellyfin.Server.Implementations.Tests/Localization/LocalizationManagerTests.cs

crobibero (approved):

webysther:
If this PR is present in build 2025041517 don't work for `pt-BR` and `pob` or `pt-br`.

This table in txt sounds strange and too complex to me, for instance, the values don't translate for the regional name. I think this table need to be a localization string for every language and 1:1 string for improved and definitive working solution. Any ideias about this?

PR #13092, opened 2024-11-23
Fix: handling of <set> elements in NfoParser

### LINKED ISSUE #10767: [Issue]: xml special characters in collection names result in multiple escape sequences added to the name over time
### Please describe your bug

XML special characters like `&`, `<` and `>` will be escaped when saved in the collection.xml. So every `&` will become `&amp;`.
The `Lilo & Stitch Collection` will be saved as `Lilo &amp; Stitch Collection` for example. And the collection folder is also named accordingly.
Loading and saving from XML should un-escape/escape these characters correctly.
At first everything seems fine. But after some time something in the loading or saving process goes wrong and the ampersand from the escape sequence is escaped again, which will result in `&amp;amp;` -> `Lilo &amp;amp; Stitch Collection`.
At this time the problem should also be visible in the UI, because even if un-escaped correctly the name now contains `&amp;` instead of a single `&`

Eventually this will repeat itself. `&amp;amp;` -> `&amp;amp;amp;` -> `&amp;amp;amp;amp;` -> ...

The bug was initially reported in the tmdbboxsets plugin repo:
https://github.com/jellyfin/jellyfin-plugin-tmdbboxsets/issues/73 (includes logs and screenshots)
https://github.com/jellyfin/jellyfin-plugin-tmdbboxsets/issues/37

I couldn't find any code in the boxset plugin that could trigger this behaviour since the boxset name is extracted from the movie metadata and boxset creation is handled by the server code.
So as far as I can tell it's not a problem with the plugin.

### Jellyfin Version

10.8.z

### if other:

10.8.13

### Environment

```markdown
- OS: Debian 11
- Linux Kernel: 6.1.0-0.deb11.7
- Virtualization: Docker
- Storage: ext4 HDD
```

### Code of Conduct

TheMelmacian:
**Changes**
- moved parsing of `<set>` elements without sub elements to `ParseSetXml`method
- use method to retrieve value that uses proper unescaping of special characters
- added new test to check unescaping works properly

**Issues**
Fixes #10767

### DIFF
--- a/MediaBrowser.XbmcMetadata/Parsers/MovieNfoParser.cs
+++ b/MediaBrowser.XbmcMetadata/Parsers/MovieNfoParser.cs
@@ -82,21 +82,13 @@ namespace MediaBrowser.XbmcMetadata.Parsers
 
                         if (!string.IsNullOrWhiteSpace(val) && movie is not null)
                         {
-                            // TODO Handle this better later
-                            if (!val.Contains('<', StringComparison.Ordinal))
+                            try
                             {
-                                movie.CollectionName = val;
+                                ParseSetXml(val, movie);
                             }
-                            else
+                            catch (Exception ex)
                             {
-                                try
-                                {
-                                    ParseSetXml(val, movie);
-                                }
-                                catch (Exception ex)
-                                {
-                                    Logger.LogError(ex, "Error parsing set node");
-                                }
+                                Logger.LogError(ex, "Error parsing set node");
                             }
                         }
 
@@ -139,7 +131,12 @@ namespace MediaBrowser.XbmcMetadata.Parsers
                     // Loop through each element
                     while (!reader.EOF && reader.ReadState == ReadState.Interactive)
                     {
-                        if (reader.NodeType == XmlNodeType.Element)
+                        if (reader.NodeType == XmlNodeType.Text && reader.Depth == 1)
+                        {
+                            movie.CollectionName = reader.Value;
+                            break;
+                        }
+                        else if (reader.NodeType == XmlNodeType.Element)
                         {
                             switch (reader.Name)
                             {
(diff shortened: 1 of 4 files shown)


Shadowghost (changes requested):
I don't think that this works as intended. The code needs to properly parse two schemas:
```
<set>Test Collection</set>
```
and the Kodi NFO schema:
```
<set>
   <name></name>
   <overview></overview>
</set>
```

The writer should also write the latter one to adher to Kodi standards.

gnattu (changes requested):
This indeed changes behavior and failed unit tests:

```

Jellyfin.XbmcMetadata.Tests.Parsers.MovieNfoParserTests.Fetch_Valid_Success [FAIL]
  Failed Jellyfin.XbmcMetadata.Tests.Parsers.MovieNfoParserTests.Fetch_Valid_Success [65 ms]
  Error Message:
   Assert.Equal() Failure: Values differ
Expected: 2017-11-15T00:00:00.0000000
Actual:   null
```

TheMelmacian:
Sorry, was late and I didn't check the other tests. I pushed a fix, that should work with both possible schemas.

Shadowghost (approved):

TheMelmacian:
> The writer should also write the latter one to adher to Kodi standards.

@Shadowghost The MovieNfoSaver currently writes the non Kodi variant to the nfo file: https://github.com/jellyfin/jellyfin/blob/9ae1ac2513e88a0897652c94abca712df18751a6/MediaBrowser.XbmcMetadata/Savers/MovieNfoSaver.cs#L116-L119

Shadowghost:
We should change that. If you want, you can do it in this PR, otherwise I'll take care of it later on another one.

TheMelmacian:
@gnattu If you have time, could you please review the changes?

gnattu (approved):

Bond-009 (approved):

PR #12385, opened 2024-08-04
Add non-standard multi-value audio tag support

gnattu:
This adds support for widely used non-standard methods of recording multiple values in audio tags. Notably:

- The non-standard ARTISTS tag used by Kodi and Picard
- The use of a custom character (e.g., `;`) as a delimiter for tag values.

A user-customizable whitelist to exclude certain items from tag splitting is also added, as some artists may have special characters in their names.

Enabling this feature is a per-library setting and is disabled by default.

**Changes**

- Adds customizable tag splitting in library options

**Issues**

Fixes #11871
Fixes #12256
Fixes #10730
Fixes #10794
Fixes #11411

PR #13098, opened 2024-11-24
Fix subtitle selection to respect preferred language for forced subtitles

### LINKED ISSUE #11776: Forced (Foreign)  Subtitles ignores the language preference on android tv 0.16.10 since version 10.9.2
### Please describe your bug

The foreign subtitles ignores the langage preferences under SMART mode. I have 2 forced subs, One english, one french. If there's a Foreign english, it will trun on which is fine, if I set the preferred language to French, the French subs will activate. So far it works. But if there's only a French forced foreign and the langage is set to English, the french subs will turn on even though it is marked to NONE on the android app. 

### Reproduction Steps

Set the Subtitles mode under user to SMART and the Preferred langage to english. 

Get a file who has foreign subs other than english, french (mkvtoolnix) French: default off, forced On 

Play the movie and you'll noticed that the french subs will be activated even though in CC (On android device, Firetv or Chromecast) ) it is set to NONE. 

You can also check on The Jellyfin server any movies that has only french forced, Subtitles should be set to Off if the Preferred langage is set to English under USER - Subtitles - Preferred subtitle language

### Jellyfin Version

10.9.0

### if other:

10.9.2

### Environment

```markdown
- OS:
- Linux Kernel:
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

[2024-05-21 14:43:18.438 +00:00] [INF] [20] MediaBrowser.MediaEncoding.Subtitles.SubtitleEncoder: "/usr/bin/ffmpeg" "-i \"/home/alliyance/files/Series/The.walking.dead.the.ones.who.live (2024)/Season.01/S01E05.mkv\" -copyts -map 0:3 -an -vn -c:s copy \"/home/alliyance/.config/jellyfin/data/data/subtitles/1/1b8e9e5d-90a9-9c9f-71ba-9358dfe09685.srt\" -map 0:4 -an -vn -c:s copy \"/home/alliyance/.config/jellyfin/data/data/subtitles/b/b56e498a-929b-6a4c-a1f6-0410f8784ac5.srt\" -map 0:5 -an -vn -c:s copy \"/home/alliyance/.config/jellyfin/data/data/subtitles/4/465ddbbf-55ab-9ce0-cf26-a5bf272efcb1.srt\""
[2024-05-21 14:43:24.889 +00:00] [INF] [56] MediaBrowser.MediaEncoding.Subtitles.SubtitleEncoder: ffmpeg subtitle extraction completed for "/home/alliyance/files/Series/The.walking.dead.the.ones.who.live (2024)/Season.01/S01E05.mkv" to "/home/alliyance/.config/jellyfin/data/data/subtitles/1/1b8e9e5d-90a9-9c9f-71ba-9358dfe09685.srt"
[2024-05-21 14:43:24.889 +00:00] [INF] [56] MediaBrowser.MediaEncoding.Subtitles.SubtitleEncoder: ffmpeg subtitle extraction completed for "/home/alliyance/files/Series/The.walking.dead.the.ones.who.live (2024)/Season.01/S01E05.mkv" to "/home/alliyance/.config/jellyfin/data/data/subtitles/b/b56e498a-929b-6a4c-a1f6-0410f8784ac5.srt"
[2024-05-21 14:43:24.889 +00:00] [INF] [56] MediaBrowser.MediaEncoding.Subtitles.SubtitleEncoder: ffmpeg subtitle extraction completed for "/home/alliyance/files/Series/The.walking.dead.the.ones.who.live (2024)/Season.01/S01E05.mkv" to "/home/alliyance/.config/jellyfin/data/data/subtitles/4/465ddbbf-55ab-9ce0-cf26-a5bf272efcb1.srt"

### FFmpeg logs

```shell
None
```

_No response_

_No response_

### Code of Conduct

timminator:
Changes:
This PR addresses an issue where forced subtitles in non-preferred languages were being selected. I adjusted the subtitle selection logic for the SubtitlePlaybackMode "Smart" and "OnlyForced" to ensure that forced subtitles are only chosen if they match the user's preferred language.

Issues:
Fixes one of two issues mentioned in #10829
Fixes #11776

### DIFF
--- a/Emby.Server.Implementations/Library/MediaStreamSelector.cs
+++ b/Emby.Server.Implementations/Library/MediaStreamSelector.cs
@@ -39,46 +39,48 @@ namespace Emby.Server.Implementations.Library
                 return null;
             }
 
+            // Sort in the following order: Default > No tag > Forced
             var sortedStreams = streams
                 .Where(i => i.Type == MediaStreamType.Subtitle)
                 .OrderByDescending(x => x.IsExternal)
-                .ThenByDescending(x => x.IsForced && string.Equals(x.Language, audioTrackLanguage, StringComparison.OrdinalIgnoreCase))
-                .ThenByDescending(x => x.IsForced)
                 .ThenByDescending(x => x.IsDefault)
-                .ThenByDescending(x => preferredLanguages.Contains(x.Language, StringComparison.OrdinalIgnoreCase))
+                .ThenByDescending(x => !x.IsForced && MatchesPreferredLanguage(x.Language, preferredLanguages))
+                .ThenByDescending(x => x.IsForced && MatchesPreferredLanguage(x.Language, preferredLanguages))
+                .ThenByDescending(x => x.IsForced && IsLanguageUndefined(x.Language))
+                .ThenByDescending(x => x.IsForced)
                 .ToList();
 
             MediaStream? stream = null;
+
             if (mode == SubtitlePlaybackMode.Default)
             {
-                // Load subtitles according to external, forced and default flags.
-                stream = sortedStreams.FirstOrDefault(x => x.IsExternal || x.IsForced || x.IsDefault);
+                // Load subtitles according to external, default and forced flags.
+                stream = sortedStreams.FirstOrDefault(x => x.IsExternal || x.IsDefault || x.IsForced);
             }
             else if (mode == SubtitlePlaybackMode.Smart)
             {
                 // Only attempt to load subtitles if the audio language is not one of the user's preferred subtitle languages.
-                // If no subtitles of preferred language available, use default behaviour.
+                // If no subtitles of preferred language available, use none.
+                // If the audio language is one of the user's preferred subtitle languages behave like OnlyForced.
                 if (!preferredLanguages.Contains(audioTrackLanguage, StringComparison.OrdinalIgnoreCase))
                 {
-                    stream = sortedStreams.FirstOrDefault(x => preferredLanguages.Contains(x.Language, StringComparison.OrdinalIgnoreCase)) ??
-                        sortedStreams.FirstOrDefault(x => x.IsExternal || x.IsForced || x.IsDefault);
+                    stream = sortedStreams.FirstOrDefault(x => MatchesPreferredLanguage(x.Language, preferredLanguages));
                 }
                 else
                 {
-                    // Respect forced flag.
-                    stream = sortedStreams.FirstOrDefault(x => x.IsForced);
+                    stream = BehaviorOnlyForced(sortedStreams, preferredLanguages).FirstOrDefault();
                 }
             }
             else if (mode == SubtitlePlaybackMode.Always)
             {
-                // Always load (full/non-forced) subtitles of the user's preferred subtitle language if possible, otherwise default behaviour.
-                stream = sortedStreams.FirstOrDefault(x => !x.IsForced && preferredLanguages.Contains(x.Language, StringComparison.OrdinalIgnoreCase)) ??
-                    sortedStreams.FirstOrDefault(x => x.IsExternal || x.IsForced || x.IsDefault);
+                // Always load (full/non-forced) subtitles of the user's preferred subtitle language if possible, otherwise OnlyForced behaviour.
+                stream = sortedStreams.FirstOrDefault(x => !x.IsForced && MatchesPreferredLanguage(x.Language, preferredLanguages)) ??
+                    BehaviorOnlyForced(sortedStreams, preferredLanguages).FirstOrDefault();
             }
             else if (mode == SubtitlePlaybackMode.OnlyForced)
             {
-                // Only load subtitles that are flagged forced.
-                stream = sortedStreams.FirstOrDefault(x => x.IsForced);
+                // Load subtitles that are flagged forced of the user's preferred subtitle language or with an undefined language
+                stream = BehaviorOnlyForced(sortedStreams, preferredLanguages).FirstOrDefault();
             }
 
             return stream?.Index;
@@ -110,38 +112,70 @@ namespace Emby.Server.Implementations.Library
             if (mode == SubtitlePlaybackMode.Default)
             {
                 // Prefer embedded metadata over smart logic
-                filteredStreams = sortedStreams.Where(s => s.IsForced || s.IsDefault)
+                // Load subtitles according to external, default, and forced flags.
+                filteredStreams = sortedStreams.Where(s => s.IsExternal || s.IsDefault || s.IsForced)
                     .ToList();
             }
             else if (mode == SubtitlePlaybackMode.Smart)
             {
                 // Prefer smart logic over embedded metadata
+                // Only attempt to load subtitles if the audio language is not one of the user's preferred subtitle languages, otherwise OnlyForced behavior.
                 if (!preferredLanguages.Contains(audioTrackLanguage, StringComparison.OrdinalIgnoreCase))
                 {
-                    filteredStreams = sortedStreams.Where(s => !s.IsForced && preferredLanguages.Contains(s.Language, StringComparison.OrdinalIgnoreCase))
+                    filteredStreams = sortedStreams.Where(s => MatchesPreferredLanguage(s.Language, preferredLanguages))
                         .ToList();
+                }
+                else
+                {
+                    filteredStreams = BehaviorOnlyForced(sortedStreams, preferredLanguages);
                 }
             }
             else if (mode == SubtitlePlaybackMode.Always)
             {
-                // Always load the most suitable full subtitles
-                filteredStreams = sortedStreams.Where(s => !s.IsForced).ToList();
+                // Always load (full/non-forced) subtitles of the user's preferred subtitle language if possible, otherwise OnlyForced behavior.
+                filteredStreams = sortedStreams.Where(s => !s.IsForced && MatchesPreferredLanguage(s.Language, preferredLanguages))
+                    .ToList() ?? BehaviorOnlyForced(sortedStreams, preferredLanguages);
             }
             else if (mode == SubtitlePlaybackMode.OnlyForced)
             {
-                // Always load the most suitable full subtitles
-                filteredStreams = sortedStreams.Where(s => s.IsForced).ToList();
+                // Load subtitles that are flagged forced of the user's preferred subtitle language or with an undefined language
+                filteredStreams = BehaviorOnlyForced(sortedStreams, preferredLanguages);
             }
 
-            // Load forced subs if we have found no suitable full subtitles
-            var iterStreams = filteredStreams is null || filteredStreams.Count == 0
-                ? sortedStreams.Where(s => s.IsForced && string.Equals(s.Language, audioTrackLanguage, StringComparison.OrdinalIgnoreCase))
-                : filteredStreams;
+            // If filteredStreams is null, initialize it as an empty list to avoid null reference errors
+            filteredStreams ??= new List<MediaStream>();
 
-            foreach (var stream in iterStreams)
+            foreach (var stream in filteredStreams)
             {
                 stream.Score = GetStreamScore(stream, preferredLanguages);
             }
+        }
+
+        private static bool MatchesPreferredLanguage(string language, IReadOnlyList<string> preferredLanguages)
+        {
+            // If preferredLanguages is empty, treat it as "any language" (wildcard)
+            return preferredLanguages.Count == 0 ||
+                preferredLanguages.Contains(language, StringComparison.OrdinalIgnoreCase);
+        }
+
+        private static bool IsLanguageUndefined(string language)
+        {
+            // Check for null, empty, or known placeholders
+            return string.IsNullOrEmpty(language) ||
+                language.Equals("und", StringComparison.OrdinalIgnoreCase) ||
+                language.Equals("unknown", StringComparison.OrdinalIgnoreCase) ||
+                language.Equals("undetermined", StringComparison.OrdinalIgnoreCase) ||
+                language.Equals("mul", StringComparison.OrdinalIgnoreCase) ||
+                language.Equals("zxx", StringComparison.OrdinalIgnoreCase);
+        }
+
+        private static List<MediaStream> BehaviorOnlyForced(IEnumerable<MediaStream> sortedStreams, IReadOnlyList<string> preferredLanguages)
+        {
+            return sortedStreams
+                .Where(s => s.IsForced && (MatchesPreferredLanguage(s.Language, preferredLanguages) || IsLanguageUndefined(s.Language)))
+                .OrderByDescending(s => MatchesPreferredLanguage(s.Language, preferredLanguages))
+                .ThenByDescending(s => IsLanguageUndefined(s.Language))
+                .ToList();
         }
 
         internal static int GetStreamScore(MediaStream stream, IReadOnlyList<string> languagePreferences)


gnattu:
I don't think simply filter out all forced subtitles with language preferences is a good enough behavior change to this. The whole reason for force subtitles to exist is to provide extra narrative contents. The current behavior will definitely cause some problems but a bigger issue is that a lot of users use this tag wrong. The forced subtitles are not meant to be used as the "default" subtitle, but in a lot of files, they do. 

In my opinion, a more practical implementation would be to load the “forced subtitles” in a “non-forced” manner. This means that we always attempt to load the subtitles based on the user’s preferences first. If the user’s preferences do not match any subtitles, or if the user has no explicit preferences, we then load the forced subtitles. This way, we can still support the intended use of forced subtitles while allowing the user to override the selection if it is being used incorrectly.

timminator:
Hi! Thank you for taking a look at my commit.
The categories Default, Smart, OnlyForced and so on are the users preferences at the moment.
If we have a video for example with three subtitle tracks (eng, ger, ger forced) and my language preference is set to english and my subtitle mode is set to OnlyForced I expect jellyfin as a user to default to subtitles off, because forced german subtitles do not provide extra narrative contents for me in this case. This is also not a misuse of the forced flag inside the video file.
And this case is currently not handled by jellyfin with the current preferences the user can select from.
This would be fixed by my commit.
My commit could only cause problems if a track has a forced flag and not a language flag right? I think this is an oversight in my behaviour change.
What do you think of adding another if statement to filter forced subtitles by language if a language flag exists, and if not just load the subtitles that are flagged forced?

gnattu:
My *draft* implementation would be like this:

- Default: no changes, users need to explicitly pick one
- Smart: check if the audio is in the preferred language, if so the same behavior **as only forced**, if not, pick one subtitle in the preferred language if available, load nothing if there is none in the preferred language, pick the marked as default one if there are multiple ones in the same language, only loads marked as forced one if there are no other options in the preferred language.
- Always: always use user settings, if no matched subtitles same **as only forced**. If there are multiple, load order should be similar to smart, which is default > no tag > forced.
- OnlyForced: Only load the one in the preferred language and marked as forced, load the first one if there are multiple. If there is none in the current language, try load the one with undefined language, if there is still none, load nothing.

As you can see this behaves differently from what we are having here. This will try to load forced subtitles based on language preferences and will try to load such subtitles as much as possible, but it will not be preferred over other subtitles like it currently is because it is not designed to be preferred in all cases. For users misuses this as the "default" tag, this will still try to load the forced one if there is no other subtitles in the selected language, but if there is multiple the user has to re-tag the file.

This is what I *think* that would work better than current implementation, I could missed some use case though.

Alliyance:
Finally. A big thank you! Can't wait for the update. If everything works, I'll close my report #11776

timminator:
Thank you for writing out your draft implementation. Your new implementation seems completely logical and also includes the changes I would like to make.
Now I've pretty much redone the whole subtitle selection logic to reflect your draft implementation.
I tested all cases I could think of and could not notice any issues with your approach.
Could you take another look at it?

Alliyance:
Question, Roku use mostly the default mode. Smart or Forced only never did anything for Roku. Will that script change Roku behavior?

holow29:
> For users misuses this as the "default" tag, this will still try to load the forced one if there is no other subtitles in the selected language, but if there is multiple the user has to re-tag the file.
This seems to imply that a correctly-tagged 'force' track will not work if there are multiple subs in the preferred language. It does not seem logical to me to have the system geared towards people with mistagged subs rather than those who have it tagged correctly. That seems backwards.

gnattu:
> It is my understanding that "forced" is supposed to be more meaningful than 'default,'

Your understanding is wrong and you are using forced wrong. Forced is not meant to be used as a stronger default.

holow29:
> > It is my understanding that "forced" is supposed to be more meaningful than 'default,'
> 
> Your understanding is wrong and you are using forced wrong. Forced is not meant to be used as a stronger default.

I think you are wrong. Forced is supposed to be used as a substitute to burned-in subs - those for foreign/alien text or language so the narrative makes sense. How does your flow take this into account? Default would be for actual subtitles, but what if you don’t want subs for regular dialogue but do for the parts that are supposed to be forced?

gnattu:
> > > It is my understanding that "forced" is supposed to be more meaningful than 'default,'
> > 
> > 
> > Your understanding is wrong and you are using forced wrong. Forced is not meant to be used as a stronger default.
> 
> I think you are wrong. Forced is supposed to be used as a substitute to burned-in subs - those for foreign/alien text or language so the narrative makes sense. How does your flow take this into account? Default would be for actual subtitles, but what if you don’t want subs for regular dialogue but do for the parts that are supposed to be forced?

Read the logic again. It is already handled.

holow29:
Sorry I was unclear, and I haven't looked at the updated code. I am concerned with following situation: mulilingual viewer, preferred language: english, forced spanish subs. Is there a situation where these forced subs work properly?

gnattu:
> Sorry I was unclear, and I haven't looked at the updated code. I am concerned with following situation: mulilingual viewer, preferred language: english, forced spanish subs. Is there a situation where these forced subs work properly?

Preferred language is a list, you can just add Spanish after English

holow29:
> Preferred language is a list, you can just add Spanish after English

IMO users shouldn't have to do this to have their media shown as it was intended (i.e. with forced subs). I think OnlyForced should also try to load forced subs in the audioTrackLanguage as a last resort - because it isn't guaranteed without user intervention that audioTrackLanguage will be in preferred language list. (Basically I think OnlyForced should first try to load forced subs in preferred language, then undefined, then audioTrackLanguage.) In many cases, this will probably be the same as audioTrackLanguage will be in preferred language list to begin with. However, for cases it is not, at least users will get the forced subs that should be there.

For example, if you have preferred language as English, no English subs, a Spanish movie with Spanish audio and Spanish forced subs, the forced subs should be loaded even if not in my preferred language list IMO; it isn't my preferred language but at least the forced subs that should be there will be loaded.

timminator:
That looks like a valid edge case you brought up @holow29 .
I think that would also be pretty easy to add to my commit.
I will look into it.

Alliyance:
I vote to not change back the new script. If a client is having issues because the subs are not well programmed, it's their problem. Loading the next available forced subs regardless of the language preferences isn't a good idea.

gnattu:
> However, for cases it is not, at least users will get the forced subs that should be there.

That would essentially remove the option to get rid of the forced subtitles in undesired languages without manually disable all subtitles. Please, don’t think the problem in the way that you or the small group of people like you is more privileged than others and doing things to remove the choice of other users just to make your use case more convenient. There are tons of posts asking for how to get rid of forced subtitles in other media solution communities like Kodi. 

> IMO users shouldn't have to do this to have their media shown as it was intended (i.e. with forced subs).

User preferences aside, the problem we have with the forced subtitles still persists. People labeling forced subtitles as incorrect and even very common software used for encoding, such as Handbrake, abuses the forced subtitles tag (at least some versions of Handbrake do). The subtitles that are not intended to be forced are automatically labeled as forced by software, leaving users with little to no understanding of the situation.  Loading a subtitle not in preferred language is a easy way to load subtitles the user does not want to see and confuse our user because Jellyfin is loading a subtitle in the language that the user does not understand. Yes you can tell your user to fix the files, but that would be harder if the software the user is used to is doing thing wrong which is very common, unfortunately. This situation is similar to the one we’re facing with #12385, where there are standard ways to add multi-value audio metadata tags, but almost everyone is doing it incorrectly and the popular software is defaulting to the non-standard way. We cannot just say everyone is wrong and force everyone to retag the files.

holow29:
Language preference is supposed to be a **preference**, not deterministic - otherwise it should be renamed and other behavior modified. If there is no forced sub for your _preferred_ language, but a forced sub exists for the audio track language, I think it should be displayed.

For example, if you are watching a foreign film that doesn't have forced subs for your language but does for the audio language. The proper way to do that would be to display the forced subs since they are supposed to be forced and then you won't miss potentially important parts of the narrative. Forced subs are supposed to act like burned-in subs, but it is better when they aren't burned in so that they can easily be translated. In a situation where there is no translation to your preferred language, I think it is more proper to display them in the audioTrackLanguage (which is probably always going to be the original audio language or at least one you can understand since you are watching the media in it). I'm actually erring on the side of the subs being well programmed; I don't think we should have to kowtow to people with badly-tagged media. (On that note, I actually think the default should be OnlyForced because the forced subs are supposed to play even when no other subs are selected....but that is likely not an issue for this PR.)

@gnattu Forced subs are there for a reason - to refuse to load them because they aren't in the preferred audio language doesn't make sense unless there are forced subs in the preferred audio language. I am not the one trying to please a minority of users - you are the one trying to develop this program to appease those with badly-tagged media. Logically, it should work the way I suggested; you are saying it should not because it would be wrong to tell users that it is their fault for having bad media; to call it privileged to pay attention to the forced subtitle flag is odd. If you pop a Blu-ray in, you expect the forced subs to be on automatically, even if there are none in your preferred language. That is how it should work and that should be the expectation. If you play media that came with forced subs in their original language and they don't play, that is an error and you are likely missing out on narrative - unless there are forced subs in your preferred language.

> That would essentially remove the option to get rid of the forced subtitles in undesired languages without manually disable all subtitles.

As far as removing user choice, I don't understand your point; can't you just choose before you start playback to disable subtitles for that play? If you don't want the forced subs which are supposed to be there, that is your choice, but that shouldn't be the default (and IIRC isn't it the default already if you don't choose OnlyForced mode?).

gnattu:
> can't you just choose before you start playback to disable subtitles for that play?

This is a weird suggestion. We can just close this PR because you can always just choose the subtitles you want to use for that play and we don’t need to have auto subtitle loading at all if manual pick is a feasible solution.

> Meanwhile where is the choice for someone who watches a lot of foreign films, potentially without preferred language subs and doesn't want to miss narrative but might not necessarily know there are forced subs they would have to manually enable? The whole point of forced subs is that they are automatically enabled so as not to miss anything.

Users do know which languages they do understand isn’t it? Adding those languages and they would not miss a thing. Plus, an empty list essentially means “All languages” and subtitles do have its own language preference setting so the audio preference is irrelevant. Just turn use OnlyForced and leave the language list empty. It is that easy.

> Logically, it should work the way I suggested; you are saying it should not because it would be wrong to tell users that it is their fault for having bad media; to call it privileged to pay attention to the forced subtitle flag is odd.

It is odd, but unfortunately this is what we have to do. If you have seen files that half of the subtitles are forced and things like that is very common you would understand the situation we are having here. It is easy to tell users that they have bad media, but it is much harder to tell them to fix the files and teaching them for how. It is even harder to let users be willing to fix their collections because that would be huge and manually identify the real forced subtitles is very very time consuming unless someone invented some sort of AI powered tools. 

Your intention might not to be privileged, but the suggestion you made could actually force tens of thousands of our users to spend hours modifying their files because you don’t want to just to change a small custom setting that you could easily do in a few seconds. That’s not very fair, and it does sounds very privileged.

holow29:
> This is a weird suggestion. We can just close this PR because you can always just choose the subtitles you want to use for that play and we don’t need to have auto subtitle loading at all if manual pick is a feasible solution.

I think if media is supposed to have forced subs and you don't want them, you should need to manually disable them since that is the whole reason the forced flag exists in the first place. That doesn't seem weird to me. Maybe there should be another subtitle mode that is "DontRespectForcedSubs" or "MyMediaIsMistagged" in that case so people will only have the subs they want loaded - meanwhile potentially missing out on narrative because forced subs which should be loaded are not.

> Just turn use OnlyForced and leave the language list empty.

This does seem like the best suggestion, but it doesn't really follow logically that this would be the way to do it. If I want to prefer English subs when available but go back to the audio track's forced subs when it isn't available, I don't have that option. I would rather have forced subs that I can't understand than no forced subs at all because at least I would know there is somethign I am missing there. At the end of the day, maybe this comes down to what "preferred" means in terms of subtitle preference. If the language preference for subtitles is actually not a preference but telling that ONLY subs in that language are allowed, then fine but that was not my understanding of how it should work - and I think it should be renamed in that case.

> Your intention might not to be privileged, but the suggestion you made could actually force tens of thousands of our users to spend hours modifying their files because you don’t want to just to change a small custom setting that you could easily do in a few seconds. That’s not very fair, and it does sounds very privileged.

I certainly don't mean to do that. I would prefer to have a system that works the way it is supposed to by respecting forced subtitles; it really isn't about my effort, but if I have correctly tagged media then I would expect it to play correctly. Jellyfin already has a lot of hacks in it server-side because they don't seem to "trust" client devs to use the APIs or HLS builder or device profiles correctly. That already leaves a bad taste in my mouth because the server devs seem to want to make the system idiot-proof. Unfortunately, in doing so, they make it harder to develop the clients because expected behavior changes with these hacks - and by going against the purpose of forced subtitles, the same is done to users as well. Anyway, I understand the desire to balance real-world use vs. correct implementation; we just happen to want to be on opposite sides of that I suppose.

Alliyance:
"I would rather have forced subs that I can't understand than no forced subs at all because at least I would know there is somethign I am missing there."


Hell no. All of my media are bilingual. If someone watch a movie in French, you will have a lot of forced French subs that will explained for example, if the actor is texting, it will show up in the French forced subs so the viewer can understand what he's writing. But for someone who is watching it in its original language, it can be very annoying having to see all those French forced subs showing up. 

I also have Plex and it doesn't react that way. It plays the forced subs only for your preferred language. It will not turn on the next available forced subs.

holow29:
> Hell no. All of my media are bilingual. If someone watch a movie in French, you will have a lot of forced French subs that will explained for example, if the actor is texting, it will show up in the French forced subs so the viewer can understand what he's writing. But for someone who is watching it in its original language, it can be very annoying having to see all those French forced subs showing up.

I don't understand...is French the original language or not? If you are watching it in the original language, the forced subs will only show up if they are in either the preferred language or the original language (or unknown) in my proposal vs. preferred or unknown in Gnattu's logic. If you are saying you are saying its original language is French but you are watching it in a different language and there are forced French subs for some parts, that is exactly right....because unless there are forced subs for your preferred language you are missing out on narrative content and you might not know that it is content you should be expected to understand.

Alliyance:
I completely understand your point of view. What I meant is, if the original movie is in English and someone watch it in French, they will end up with a lot of forced French subs. Those French subs will show up when there any written text embedded in the movie. So there's no need for a person who watch it in English to end up having those French subs showing up. 

If that option will only effect "Only forced mode", that's fine with me. I just wanna make sure that in Smart mode, the preferred language is not ignore. Means, if there is no forced English in Smart mode with preferred language set to English, then, subs must stay off. 

Now Roku device works with the default mode. If you have a forced default set to yes in mkvtool, it will show up, with the preferred language, English. If there is no forced English but a French forced is present with the default set to No, nothing will show up.

holow29:
> I completely understand your point of view. What I meant is, if the original movie is in English and someone watch it in French, they will end up with a lot of forced French subs. Those French subs will show up when there any written text embedded in the movie. So there's no need for a person who watch it in English to end up having those French subs showing up.

This raises another interesting case I think. Presumably you would have both English and French in your preferred language list since you are choosing to watch it in French without English subs. In that case, there might not be forced subs in English because 1) the forced subs are for parts that would need to be translated for on-screen graphics/text as you said and 2) the movie is originally in English. However, there would be forced subs in French. In this case, under Gnattu's plan, both OnlyForced (& Smart) would result in forced French subs. I'm not sure if Jellyfin can work around this because that would involve recognizing the original language of the movie and the fact that there are no forced subs in the original language as well. I would assume if this were a Blu-ray in a player, you would just get the forced french subs because it is assuming that you would want them as you are choosing to watch the movie in French vs. English.

Alliyance:
Jellyfin was working fine a few months ago. 

It's simple, if under Smart mode your preferred language is English, it should turn on any forced English that is set with mkvtool (forced display on & default track on or off) if there's no forced english, then it should stay to off. 

If your preferred language is set to French and your French subs is set to (forced display on, default track off) it should turn French forced on. I think the default track doesn't matter if it's on or off. Jellyfin should just match the preferred language for the subs who is marked "forced display on". That's how it was before and that's how Plex works. Of course under SMART mode.

holow29:
> It's simple, if under Smart mode your preferred language is English, it should turn on any forced English that is set with mkvtool (forced display on & default track on or off) if there's no forced english, then it should stay to off.

it is not that simple. Let’s take your example of movies, split between French and English. Let’s say you just have one preferred subtitle language (even though it is a list and you should have both French and English probably). Here are the scenarios:

1) preferred sub language: English, audio track: French, original language: English (so no forced English subs, but there are forced French subs.) In Gnattu’s logic, no forced subs would appear. (This is a scenario I think is more complicated because you could miss narrative in this instance, which is what my first comments in this PR are regarding.)

2) preferred sub language: French, audio track: French, original language: English (so no forced English subs, but there are forced French subs.) In Gnattu’s logic, French forced subs would appear.

However this can also be turned around; for example what if it were the opposite and you were watching a film originally in French but you want to watch with English audio. If your preferred subtitle language is English, you will get English forced subs if they exist, and this would be the same as the example you gave of subs that you don’t want for an English film in French, assuming the French forced subs don’t exist for this film because the text is already in French on-screen.

Alliyance:
No one will set the preferred subs language to English and choose French audio. It just doesn't make sense. If I have a friend who watch movies in French, both audio and preferred subs will be set to French. Same for English, audio and subs to English.

holow29:
> No one will set the preferred subs language to English and choose French audio.

Okay then I don't understand your objection to this PR or my suggestion...

> if the original movie is in English and someone watch it in French, they will end up with a lot of forced French subs. Those French subs will show up when there any written text embedded in the movie. So there's no need for a person who watch it in English to end up having those French subs showing up.

I understood this as you describing a different situation. If your situation is:

- Preferred subtitle language: English
- Audio track: English
- Forced subs only available in: French

No one is suggesting to display the forced French subs in this instance.

However your example that I just quoted seemed like it was:

- Preferred subtitle language: English
- Audio track: French
- Forced subs only available in: French

In this case, I think forced French subs should appear. (But won't under Gnattu's logic); you seem to think no one would watch this way but sometimes there is no choice because the film might not have an English audio track.

Also, in a situation where you are watching a movie that was originally in English with the following:

- Preferred subtitle language: French
- Audio track: French
- Forced subs only available in: French

That is also a case where the forced French subs should appear (and would under gnattu's logic as well).

gnattu:
> In this case, I think forced French subs should appear. (But won't under Gnattu's logic); you seem to think no one would watch this way but sometimes there is no choice because the film might not have an English audio track.

Expressing your opinion repeatedly without considering alternative use cases is not helpful. Criticizing others for incorrect tag usage and blaming the software maintainer for tolerating such irregularities does not make your opinion more acceptable. We already have enough reasons to oppose forced languages ignoring the preferred language setting, especially when there is no option to opt out. In my opinion, your use case is already archive-able by modifying current server settings which means we don’t really need an extra weird “ignore language for forced subs” setting to increase the complexity.

> For example, imagine this situation:
>
>Preferred subtitle language: English, French
>Audio track: English
>Forced subs only available in: French
>With the current logic if I am following the code correctly, the French forced subs will appear.

Adding this check to the only forced mode could make things more consistent, but then we’d have the issue of forced tags being abused, and those subtitles wouldn’t load. Users don’t care about file correctness; they just want their files to work. By making their files work, we’ll have to break some standards because the files can’t be handled in the standard way. I don’t want to keep repeating these realistic points over and over in this discussion. It’s not productive and won’t lead us anywhere. If you really care about your use case that much, you should take over this PR and implement a ‘forced subs strict mode’ that makes the forced subs to behave however you think is most correct by adding a feature switch. Instead of writing endless walls of text asking for some change that we’ve already explained why we can’t do, let’s focus on finding a solution that works for everyone.

Alliyance:
Like timminator mentioned 

"I adjusted the subtitle selection logic for the SubtitlePlaybackMode "Smart" and "OnlyForced" to ensure that forced subtitles are only chosen if they match the user's preferred language."

Simple and fixed everything.

Alliyance:
Anyone knows when will the update be release?

Alliyance:
Why the new update 10.10.4 did not include that fix?

thornbill:
Well it’s not approved or merged so that is why.

Alliyance:
> Well it’s not approved or merged so that is why. 

Oh I thought it was approved. My bad.

timminator:
Could somebody review my PR please? I would really appreciate it.

Alliyance (approved):
PR seem in perfect order.

gnattu (approved):
This looks OK for current use case where the user are seeing subtitles in languages they do not understand. For even stricter forced subtitles behavior we can modify the `BehaviorOnlyForced` method to allow forced subtitles not within the language preferences and make that a user option(which requires additional API changes). But that is for another PR.

crobibero (approved):

timminator:
Thank you crobibero and gnattu for approving my PR! Would it be possible to get it merged soon?

solbu:
First: I have no intention of rehashing this PR.
I have an interest in subtitles, and this PR came up as one the changes on the recent RC2 test release, so I clicked here to see what this was about.
I just have a comment on _this single comment,_ for future reference only.


> No one will set the preferred subs language to English and choose French audio. It just doesn't make sense.

I do exactly this, only with norwegian as the prefered language.
I am norwegian. I speak norwegian, and are fluent in English. I also understand swedish and danish very well since the Scandinavian languages are related and very similar.

So I have set prefered language to norwegian, but if the _original_ audio language is french and it is also dubbed into norwegian, I will watch with the prefered norwegian subtitles and _french_ audio – despite the fact that I do not understand french at all.

In Scandinavia it is normal to see all movies and TV series in their original audio language, regardless of whether or not we understand the spoken language. So we watch all foreign language media with subtitles. This is true for absolutely every norwegian TV channel. They broadcast content in the original audio language with norwegian subtitles. We even do this on the TV news. Everything that is not spoken in norwegian are _subtitled,_ not dubbed.
In my 52 years of living on this planet, I have yet to see any norwegian or swedish broadcast send anything that is dubbed into norwegian.

Dubbing into another language is frowned upon here, to the point that most people really do hate it. This also has the effect that absolutely nothing is dubbed into norwegian, because we do not want it.
The _only_ exception to this is childrens content, which is regularly dubbed. Everything else is watched in the original audio language with subtitles.

PR #13103, opened 2024-11-24
Add check to prevent downgrade from future EFCore refactor

JPVenson:
Added check to prevent downgrade and produce a speaking error message

Dont ever backport this!

### DIFF
--- a/Jellyfin.Server/Migrations/MigrationRunner.cs
+++ b/Jellyfin.Server/Migrations/MigrationRunner.cs
@@ -88,6 +88,12 @@ namespace Jellyfin.Server.Migrations
                  ? (MigrationOptions)xmlSerializer.DeserializeFromFile(typeof(MigrationOptions), migrationConfigPath)!
                  : new MigrationOptions();
 
+            // 10.10 specific EFCore migration check.
+            if (migrationOptions.Applied.Any(f => f.Id.ToString("D") == "5C4B82A2-F053-4009-BD05-B6FCAD82F14C" || f.Name == "MigrateUserDatabase"))
+            {
+                throw new InvalidOperationException("You cannot downgrade your jellyfin install from the library.db migration.");
+            }
+
             // We have to deserialize it manually since the configuration manager may overwrite it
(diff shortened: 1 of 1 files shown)


Jellyfin.Server/Migrations/MigrationRunner.cs
```diff
@@ -88,6 +88,12 @@ public static void RunPreStartup(ServerApplicationPaths appPaths, ILoggerFactory
                  ? (MigrationOptions)xmlSerializer.DeserializeFromFile(typeof(MigrationOptions), migrationConfigPath)!
                  : new MigrationOptions();
 
+            // 10.10 specific EFCore migration check.
+            if (migrationOptions.Applied.Any(f => f.Id.ToString("D") == "5C4B82A2-F053-4009-BD05-B6FCAD82F14C" || f.Name == "MigrateUserDatabase"))
```
cvium, line 92: MigrateUserDatabase was applied in 2020. Won't you be bricking all existing 10.10 installs with this?

Shadowghost (approved):
Maybe do an ordinal string comparison instead of `==`

Jellyfin.Server/Migrations/MigrationRunner.cs
```diff
@@ -88,6 +88,12 @@ public static void RunPreStartup(ServerApplicationPaths appPaths, ILoggerFactory
                  ? (MigrationOptions)xmlSerializer.DeserializeFromFile(typeof(MigrationOptions), migrationConfigPath)!
                  : new MigrationOptions();
 
+            // 10.10 specific EFCore migration check.
+            if (migrationOptions.Applied.Any(f => f.Id.ToString("D").Equals("36445464-849f-429f-9ad0-bb130efa0664", StringComparison.InvariantCulture) || f.Name == "MigrateLibraryDbData"))
```
crobibero, line 92: Should save the guid above and do a guid comparison instead
Also this comparison is failing build ci

cvium (approved):

crobibero (approved):

PR #13102, opened 2024-11-24
Removed RemoveOldPlugins configuration flag

RealGreenDragon:
Fixes #12959

PR #13106, opened 2024-11-25
Enable RemoveOldPlugins by default (10.10.z backport)

RealGreenDragon:
Backport of PR #13102 to 10.10.z branch.

### DIFF
--- a/MediaBrowser.Model/Configuration/ServerConfiguration.cs
+++ b/MediaBrowser.Model/Configuration/ServerConfiguration.cs
@@ -246,7 +246,7 @@ public class ServerConfiguration : BaseApplicationConfiguration
     /// <summary>
     /// Gets or sets a value indicating whether older plugins should automatically be deleted from the plugin folder.
     /// </summary>
-    public bool RemoveOldPlugins { get; set; }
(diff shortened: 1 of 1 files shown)


JPVenson (approved):

nyanmisaka (approved):

Shadowghost (approved):
Even though this is a behaviour change, I think this is what is expected to happen when you upgrade because otherwise the plugin is never properly uninstalled

Bond-009 (approved):

GlassedSilver:
> Even though this is a behaviour change, I think this is what is expected to happen when you upgrade because otherwise the plugin is never properly uninstalled

Two questions, because I just noticed this in the current change log.

1) I swear I was able to uninstall an incompatible plugin before, am I hallucinating that or did something change at some point? :S 
2) Say you upgrade JF before a plugin got adapted to the new release and is marked as incompatible and hence removed... Now eventually - after having the incompatible plugin not dangle in your setup but be uninstalled - a new plugin version releases. I take it the new release will not automatically re-install, and the settings will be gone as well, right? Or is there a sort of smart treatment about such chronologically unmatched conditions?

PR #13113, opened 2024-11-26
Only do DoVi remux when the client supports profiles without fallbacks

gnattu:
In 10.10 clients that can only play the fallback layer like the Samsung TVs will report `DOVIWithHDR10` as supported video range, but the server should not do remux in DoVi as the client can only play the fallback layer. This changes the server to only do DoVi remux when the client can play DoVi videos without a fallback layer.

**Changes**

**Issues**

Forum report: https://forum.jellyfin.org/t-issue-with-playback-of-hevc-files-in-jellyfin-dolby-vision-tag-misconfiguration?pid=43992

### DIFF
--- a/Jellyfin.Api/Controllers/DynamicHlsController.cs
+++ b/Jellyfin.Api/Controllers/DynamicHlsController.cs
@@ -1819,16 +1819,13 @@ public class DynamicHlsController : BaseJellyfinApiController
         if (isActualOutputVideoCodecHevc || isActualOutputVideoCodecAv1)
         {
             var requestedRange = state.GetRequestedRangeTypes(state.ActualOutputVideoCodec);
-            var requestHasDOVI = requestedRange.Contains(VideoRangeType.DOVI.ToString(), StringComparison.OrdinalIgnoreCase);
-            var requestHasDOVIWithHDR10 = requestedRange.Contains(VideoRangeType.DOVIWithHDR10.ToString(), StringComparison.OrdinalIgnoreCase);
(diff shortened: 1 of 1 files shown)


nyanmisaka (approved):

Bond-009 (approved):

gnattu:
> The `supportsDolbyVision` function in browserDeviceProfile.js in the web client should have returned false, shouldn't it have? Even if it did somehow return true, shouldn't the canPlayType functions have also returned false? I do not understand how the capabilities were reported as such in the first place.

LG TV and Samsung TV has different logics because for Samsung TVs there are no DoVi capabilities at all so those color ranges are reported purely for the fallback layer. On LG TVs the real DoVi without fallback is always reported on supported TVs.

gnattu:
Clients can inject the profiles by design.

GlassedSilver:
Does this trigger a re-encode then or is the fallback layer remuxed then?

gnattu:
Only the fallback layer is remuxed

bm867002:
That modification breaks DoVi playback on my LG C2. Just tested Jellyfin with and without this code modification. The new code lead to HDR fallback of all my DoVi (profile 8.1) files muxed in MKV. Other containers were not tested since not available. With the old code, everything is working fine.

I did not really proceed far with debugging, but variable "videoIsDoVi" seems to be set properly, but "clientSupportsDoVi" not. I am using the recent Jellyfin player from LG webstore. Maybe the player is not reporting DoVi modes without fallbacks?!

PR #13125, opened 2024-11-29
Migrate keyframes to configured cache directory

tcely:
I'm not sure why these were put into the data directory in the first place.

The TODO comment about making it configurable is satisfied, for me, by using the already configurable cache directory instead.

**Changes**

Check for old keyframes json files in the data directory.
Only write new files to the cache directory.

**Issues**

### DIFF
--- a/src/Jellyfin.MediaEncoding.Hls/Cache/CacheDecorator.cs
+++ b/src/Jellyfin.MediaEncoding.Hls/Cache/CacheDecorator.cs
@@ -51,6 +51,11 @@ public class CacheDecorator : IKeyframeExtractor
         var cachePath = GetCachePath(_keyframeCachePath, filePath);
         if (TryReadFromCache(cachePath, out var cachedResult))
         {
+            // touch the used files to avoid cleanup
+            // unreferenced files should be cleaned up eventually
+            var fi = new FileInfo(cachePath);
+            fi.LastWriteTime = DateTime.Now;
+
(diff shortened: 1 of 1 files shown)


gnattu (changes requested):
These are technically metadata and not temporary data which should not be placed in cache.

tcely:
Alright. So, integrating these files into the also configurable metadata directory would be better?

gnattu:
> Alright. So, integrating these files into the also configurable metadata directory would be better?

Just keep where it is at the moment is good enough. Is there any specific reason you want to put those json file somewhere else that much?

tcely:
It's data that is rarely used that is uncompressed and unreferenced files aren't being cleaned up, so I'm trying to clear it out of the same location as the database file.

Attachments and subtitles seem to share the same condition. Having them on the same filesystem as my database just means there is less room for that file to grow.

gnattu:
> It's data that is rarely used

Except it is used for each and every remux playback

tcely:
Remix playback is happening much less often than browsing the library (database access) in my usage.

tcely:
> Keyframe data is meant to go into the db.

Hopefully, an external database is an option before that happens.

Needing a specific filesystem to support sqlite3 is why I'm not already moving my Jellyfin configuration into a larger shared filesystem.

Shadowghost:
This is obsolet once #13771 is merged

PR #13132, opened 2024-11-30
Fix the issue where the external audio track always defaults.

choyakawa:
When a file contains only one track, it seems the default status should always be determined by the filename. Otherwise, with multiple single-track external audio files, each one will always be identified as the default.


Bond-009 (approved):

crobibero (approved):

gnattu:
This breaks subtitle selection for containers have self-contained metadata. Unlike audio track where it is usually marked as default on each and every file, some container (like matroska based mks) can have a single subtitle stream which is metadata taggeg explicitly as non-default. This change caused a regression to make mks can never be default subtitle unless there are multiple subtitles within the same container, or explicitly named to be the default.

IDisposable:
> This change caused a regression to make subtitles in mks can never be default subtitle unless there are multiple subtitles within the same container, or explicitly named to be the default.

This is, in fact breaking unit tests on PR #9560 (with unrelated changes).

Do we want to change the test data in [_MediaInfoResolverTests.cs_ lines 242-253](https://github.com/jellyfin/jellyfin/blob/master/tests/Jellyfin.Providers.Tests/MediaInfo/MediaInfoResolverTests.cs#L242-L253) or do we want to revert this logic/tweak the resolver logic to set `IsDefault` to true when the container has only one non-default subtitle stream?

PR #13151, opened 2024-12-03
Always do tone-mapping for HDR transcoding when software pipeline is used

nyanmisaka:
**Changes**
- Enable software tone-mapping by default
  (Web https://github.com/jellyfin/jellyfin-web/pull/6362)

**Issues**
- Transcoding HDR video without tonemapping results in an unacceptable viewing experience.
  Many users are not even aware of the option and therefore we should always enable the software tonemapx filter.

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -309,7 +309,6 @@ namespace MediaBrowser.Controller.MediaEncoding
         private bool IsSwTonemapAvailable(EncodingJobInfo state, EncodingOptions options)
         {
             if (state.VideoStream is null
-                || !options.EnableTonemapping
                 || GetVideoColorBitDepth(state) < 10
                 || !_mediaEncoder.SupportsFilter("tonemapx"))
             {


Shadowghost (approved):
Won't this kill low-power machines though?

nyanmisaka:
> Won't this kill low-power machines though?

Low-power devices should set up hardware transcoding or upgrade their hardware. Jellyfin should not provide incorrect transcoded video with pale colors, otherwise people will blame Jellyfin.

nyanmisaka:
In many cases we design software from the programmer's perspective, but this is not always advisable. I bet there are people who have never even heard of the term *tonemapping*.

gnattu:
This is more likely “always do tone-mapping for HDR transcoding when software pipeline is used”. I’m personally fine with this change because systems not using hardware pipeline should be powerful enough already, but having such behavior change in bugfix releases should be better documented or delay this change to next major release

Bond-009:
I agree that this doesn't seem like something that should be introduced in a bug-fix release.

joshuaboniface:
I'm on the fence here, I agree with @Bond-009 100%, but also, I agree with the idea that HDR shouldn't be washed out by default. I think this is fine to enable and just deal with the fallout.

nyanmisaka:
> I'm on the fence here, I agree with @Bond-009 100%, but also, I agree with the idea that HDR shouldn't be washed out by default. I think this is fine to enable and just deal with the fallout.

IMHO the video with washed out color is broken. Some people may be able to live with it, but I think most people can't.

The reason we don't enable tonemap by default for hardware transcoding is that hardware detection is not yet complete, and enabling it by default may cause transcoding to fail.

When introducing the software tonemap filter, we reused the same checkbox, but the software tonemap filter will not fail anyway. After this change, software tonemap will be automatically applied just like software filters such as deinterlace. Unless the user is doing 4k -> 4k pure software transcoding on a low-power computer, our highly optimized software tonemapx filter should perform well.

PR #13175, opened 2024-12-07
Migrate to IExternalUrlProvider

Shadowghost:
**Changes**
* Migrate external URL generation to own providers

### DIFF
--- a/MediaBrowser.Providers/Movies/ImdbExternalUrlProvider.cs
+++ b/MediaBrowser.Providers/Movies/ImdbExternalUrlProvider.cs
@@ -1,0 +1,27 @@
+using System.Collections.Generic;
+using MediaBrowser.Controller.Entities;
+using MediaBrowser.Controller.Providers;
+using MediaBrowser.Model.Entities;
+
+namespace MediaBrowser.Providers.Movies;
+
+/// <summary>
+/// External URLs for IMDb.
+/// </summary>
+public class ImdbExternalUrlProvider : IExternalUrlProvider
+{
+    /// <inheritdoc/>
+    public string Name => "IMDb";
+
+    /// <inheritdoc/>
+    public IEnumerable<string> GetExternalUrls(BaseItem item)
+    {
+        var baseUrl = "https://www.imdb.com/";
+        var externalId = item.GetProviderId(MetadataProvider.Imdb);
+
+        if (!string.IsNullOrEmpty(externalId))
+        {
+            yield return baseUrl + $"title/{externalId}";
+        }
+    }
+}
--- a/MediaBrowser.Providers/Movies/ImdbExternalUrlProvider.cs
+++ b/MediaBrowser.Providers/Movies/ImdbExternalUrlProvider.cs
@@ -17,9 +17,7 @@ public class ImdbExternalUrlProvider : IExternalUrlProvider
     public IEnumerable<string> GetExternalUrls(BaseItem item)
     {
         var baseUrl = "https://www.imdb.com/";
-        var externalId = item.GetProviderId(MetadataProvider.Imdb);
-
-        if (!string.IsNullOrEmpty(externalId))
+        if (item.TryGetProviderId(MetadataProvider.Imdb, out var externalId))
         {
             yield return baseUrl + $"title/{externalId}";
         }
--- a/MediaBrowser.Providers/Plugins/MusicBrainz/MusicBrainzAlbumArtistExternalUrlProvider.cs
+++ b/MediaBrowser.Providers/Plugins/MusicBrainz/MusicBrainzAlbumArtistExternalUrlProvider.cs
@@ -1,0 +1,29 @@
+using System.Collections.Generic;
+using MediaBrowser.Controller.Entities;
+using MediaBrowser.Controller.Entities.Audio;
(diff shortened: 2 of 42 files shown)


MediaBrowser.Providers/Plugins/Tmdb/TmdbExternalUrlProvider.cs
```diff
@@ -0,0 +1,73 @@
...
+                    yield return TmdbUtils.BaseTmdbUrl + $"tv/{seriesExternalId}/season/{season.IndexNumber}";
+                }
+
+                break;
+            case Episode episode:
+                seriesExternalId = episode.Series.GetProviderId(MetadataProvider.Tmdb);
+                if (!string.IsNullOrEmpty(seriesExternalId))
+                {
+                    yield return TmdbUtils.BaseTmdbUrl + $"tv/{seriesExternalId}/season/{episode.Season.IndexNumber}/episode/{episode.IndexNumber}";
+                }
```
scampower3, line 44: Maybe restrict this to default order?

MediaBrowser.Providers/Plugins/Tmdb/TmdbExternalUrlProvider.cs
```diff
@@ -0,0 +1,73 @@
...
+                    yield return TmdbUtils.BaseTmdbUrl + $"tv/{externalId}";
+                }
+
+                break;
+            case Season season:
+                var seriesExternalId = season.Series.GetProviderId(MetadataProvider.Tmdb);
+                if (!string.IsNullOrEmpty(seriesExternalId))
+                {
+                    yield return TmdbUtils.BaseTmdbUrl + $"tv/{seriesExternalId}/season/{season.IndexNumber}";
+                }
```
scampower3, line 36: Maybe restrict this to default order?
Also, for alternate orders, maybe we can use episode groups provided from TMDB?

Shadowghost: I need to check if we can easily fetch them. If we do, we should be able to use them.

MediaBrowser.Providers/Plugins/Tmdb/TmdbExternalUrlProvider.cs
Shadowghost: I checked it and we are currently not saving any orders in the provider...

MediaBrowser.Providers/Plugins/Tmdb/TmdbExternalUrlProvider.cs
Shadowghost: We don't save those groups anywhere so we can't do this.

MediaBrowser.Providers/Movies/ImdbExternalUrlProvider.cs
```diff
@@ -0,0 +1,27 @@
...
+public class ImdbExternalUrlProvider : IExternalUrlProvider
+{
+    /// <inheritdoc/>
+    public string Name => "IMDb";
+
+    /// <inheritdoc/>
+    public IEnumerable<string> GetExternalUrls(BaseItem item)
+    {
+        var baseUrl = "https://www.imdb.com/";
+        var externalId = item.GetProviderId(MetadataProvider.Imdb);
```
Bond-009, line 20: Can't we use `TryGetProviderId` instead?
(same for all the following instances)

MediaBrowser.Providers/Plugins/MusicBrainz/MusicBrainzAlbumArtistExternalUrlProvider.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// <inheritdoc/>
+    public string Name => "MusicBrainz Album Artist";
+
+    /// <inheritdoc/>
+    public IEnumerable<string> GetExternalUrls(BaseItem item)
+    {
+        if (item is MusicAlbum)
+        {
+        if (item.TryGetProviderId(MetadataProvider.MusicBrainzAlbumArtist, out var externalId))
+            {
```
crobibero, line 23:
```suggestion
            if (item.TryGetProviderId(MetadataProvider.MusicBrainzAlbumArtist, out var externalId))
            {
```

MediaBrowser.Providers/Plugins/MusicBrainz/MusicBrainzAlbumExternalUrlProvider.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// <inheritdoc/>
+    public string Name => "MusicBrainz Album";
+
+    /// <inheritdoc/>
+    public IEnumerable<string> GetExternalUrls(BaseItem item)
+    {
+        if (item is MusicAlbum)
+        {
+        if (item.TryGetProviderId(MetadataProvider.MusicBrainzAlbum, out var externalId))
+            {
```
crobibero, line 23:
```suggestion
            if (item.TryGetProviderId(MetadataProvider.MusicBrainzAlbum, out var externalId))
            {
```

JPVenson (approved):

crobibero (approved):

PR #13194, opened 2024-12-09
Enable VideoToolbox AV1 decode

gnattu:
This decoder differs from others provided by VideoToolbox in that it lacks any software fallback. To achieve consistent behavior with other VideoToolbox decoders, this PR implemented additional checking on the server to simulate the software fallback provided by VideoToolbox.

The current fallback checking mechanism is a temporary solution. In the long term, it should be replaced with a more capable hardware capability checking system.

**Changes**

**Issues**

Required ffmpeg change is already backported here: https://github.com/jellyfin/jellyfin-ffmpeg/pull/500

The upstream has not released tagged version with VideoToolbox AV1 yet. Jellyfin-ffmpeg is required for now.

### DIFF
--- a/MediaBrowser.MediaEncoding/Encoder/ApplePlatformHelper.cs
+++ b/MediaBrowser.MediaEncoding/Encoder/ApplePlatformHelper.cs
@@ -1,0 +1,85 @@
+#pragma warning disable CA1031
+
+using System;
+using System.Linq;
+using System.Runtime.InteropServices;
+using Microsoft.Extensions.Logging;
+
+namespace MediaBrowser.MediaEncoding.Encoder;
+
+/// <summary>
+/// Helper class for Apple platform specific operations.
+/// </summary>
+public static class ApplePlatformHelper
+{
+    private static readonly string[] _av1DecodeBlacklistedCpuClass = ["M1", "M2"];
+
+    private static string GetSysctlValue(string name)
+    {
+        IntPtr length = IntPtr.Zero;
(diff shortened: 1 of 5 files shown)


Shadowghost (approved):

JPVenson (approved):

nyanmisaka (approved):

crobibero (approved):

MediaBrowser.MediaEncoding/Encoder/ApplePlatformHelper.cs
```diff
@@ -0,0 +1,85 @@
...
+    {
+        return NativeMethods.SysctlByName(System.Text.Encoding.ASCII.GetBytes(name), oldp, ref oldlenp, newp, newlen);
+    }
+
+    /// <summary>
+    /// Check if the current system has hardware acceleration for AV1 decoding.
+    /// </summary>
+    /// <param name="logger">The logger used for error logging.</param>
+    /// <returns>Boolean indicates the hwaccel support.</returns>
+    public static bool HasAv1HardwareAccel(ILogger logger)
```
Bond-009, line 55: Make sure we don't call this on an unsupported platform
```suggestion
    [SupportedOSPlatform("macos")]
    public static bool HasAv1HardwareAccel(ILogger logger)
```

Bond-009 (changes requested):

Bond-009 (approved):

PR #13208, opened 2024-12-11
Update DayOfWeekHelper.cs

yuvashrikarunakaran:
I've included comments for clarity and adjusted syntax where necessary for improved performance and maintainability

**Changes**

**Issues**


micdev42:
The `GetDaysOfWeek()` method does not appear to be referenced in the codebase - or am I missing something?
Even if the method is needed, why is it returning a `new` array? Surely it would be better to define the arrays as static readonly and return a reference to the requested one?
I may be completely misunderstanding what's going on as I am new to Jellyfin.

PR #13209, opened 2024-12-11
Transcode to audio codec satisfied other conditions when copy check failed.

gnattu:
Some clients may have specific requirements for audio codec support. For instance, WebOS only supports flac in stereo, while Safari only supports opus in stereo. When attempting to check the validity for remuxing fails for such clients, we should instead fallback to alternative codecs that support other transcoding conditions, instead of forcing a downmix in the source codec.

If this change is considered too much for a bug fix release, I will rebase this to target master.

**Changes**

**Issues**

### DIFF
--- a/MediaBrowser.Model/Dlna/StreamBuilder.cs
+++ b/MediaBrowser.Model/Dlna/StreamBuilder.cs
@@ -872,8 +872,8 @@ namespace MediaBrowser.Model.Dlna
                         {
                             var appliedVideoConditions = options.Profile.CodecProfiles
                                 .Where(i => i.Type == CodecType.VideoAudio &&
-                                            i.ContainsAnyCodec(transcodingAudioCodec, container) &&
-                                            i.ApplyConditions.All(applyCondition => ConditionProcessor.IsVideoAudioConditionSatisfied(applyCondition, audioChannels, audioBitrate, audioSampleRate, audioBitDepth, audioProfile, false)))
+                                    i.ContainsAnyCodec(transcodingAudioCodec, container) &&
+                                    i.ApplyConditions.All(applyCondition => ConditionProcessor.IsVideoAudioConditionSatisfied(applyCondition, audioChannels, audioBitrate, audioSampleRate, audioBitDepth, audioProfile, false)))
                                 .Select(i =>
                                     i.Conditions.All(condition => ConditionProcessor.IsVideoAudioConditionSatisfied(condition, audioChannels, audioBitrate, audioSampleRate, audioBitDepth, audioProfile, false)));
 
--- a/MediaBrowser.Model/Dlna/StreamBuilder.cs
+++ b/MediaBrowser.Model/Dlna/StreamBuilder.cs
@@ -846,6 +846,12 @@ namespace MediaBrowser.Model.Dlna
                         return transcodingProfile;
                     }
 
+                    // Directly return the transcoding profile if the audio codec is already on its own
+                    if (string.Equals(transcodingProfile.AudioCodec, audioStream.Codec, StringComparison.OrdinalIgnoreCase))
+                    {
+                        return transcodingProfile;
+                    }
+
                     var targetAudioCodecTranscodingProfile = new TranscodingProfile(transcodingProfile)
                     {
                         AudioCodec = audioStream.Codec
@@ -2297,6 +2303,11 @@ namespace MediaBrowser.Model.Dlna
                 if (string.Equals(profile.AudioCodec, audioCodec, StringComparison.OrdinalIgnoreCase) ||
                     string.Equals(profile.Container, audioCodec, StringComparison.OrdinalIgnoreCase))
                 {
+                    if (string.Equals("opus", audioCodec, StringComparison.OrdinalIgnoreCase) && audioStream?.Channels > 2)
+                    {
+                        return false;
+                    }
+
                     return true;
                 }
             }
--- a/MediaBrowser.Model/Dlna/StreamBuilder.cs
+++ b/MediaBrowser.Model/Dlna/StreamBuilder.cs
@@ -816,54 +816,11 @@ namespace MediaBrowser.Model.Dlna
             }
 
             var transcodingProfiles = options.Profile.TranscodingProfiles
-                .Where(i => i.Type == playlistItem.MediaType && i.Context == options.Context).ToList();
+                .Where(i => i.Type == playlistItem.MediaType && i.Context == options.Context);
 
             if (item.UseMostCompatibleTranscodingProfile)
             {
-                transcodingProfiles = transcodingProfiles.Where(i => string.Equals(i.Container, "ts", StringComparison.OrdinalIgnoreCase)).ToList();
-            }
-
-            // Pre-process the transcoding profiles to split the audio codec if the audio stream is going to be checked for direct streaming.
-            // This is to ensure that the audio codec conditions are checked against the transcoding profile that can directly stream the audio codec,
-            // instead of other codecs carried within the same transcoding profile.
-            // For example: the transcoding profile has aac and flac, and the audio stream is flac. The transcoding profile should be split into an aac profile and a flac profile.
-            // So that the conditions for flac will not be applied to aac.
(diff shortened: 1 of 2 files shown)


Shadowghost (approved):
IMHO this is a bugfix

gnattu:
> Changes in OpenAPI specification found. Expand to see details.

For some reason the documentation changes are counted as API changes... This does not change the API in any way.

dmitrylyzo:
@gnattu Here is a draft of what I was talking about (in chat).
```cs
var transcodingAudioCodecs = ContainerHelper.Split(transcodingProfile.AudioCodec);

for (var transcodingAudioCodec in transcodingAudioCodecs)
{
            i.ContainsAnyCodec(transcodingAudioCodec, container) &&
            i.ApplyConditions.All(applyCondition => ConditionProcessor.IsVideoAudioConditionSatisfied(applyCondition, audioChannels, audioBitrate, audioSampleRate, audioBitDepth, audioProfile, false)))
            .Select(i =>
                i.Conditions.All(condition => ConditionProcessor.IsVideoAudioConditionSatisfied(condition, audioChannels, audioBitrate, audioSampleRate, audioBitDepth, audioProfile, false)));

    // An empty appliedVideoConditions means that the codec has no conditions for the current audio stream
    var conditionsSatisfied = appliedVideoConditions.All(satisfied => satisfied);

    var rankAudio = 3;

    if (conditionsSatisfied)
    {
        rankAudio = string.Equals(transcodingAudioCodec, audioCodec, StringComparison.OrdinalIgnoreCase) ? 1 : 2;
    }

    rank.Audio = Math.Min(rank.Audio, rankAudio);

    if (rank.Audio == 1)
    {
        break;
    }
}
```
We test each audio codec individually and take the best rank as the transcoding profile rank. The output codec will be selected later in the `DynamicHlsController`.
I've added a check for the current audio codec, so we can probably remove the top block.

gnattu:
> We test each audio codec individually and take the best rank as the transcoding profile rank. The output codec will be selected later in the DynamicHlsController.

This would require changing even more logic because the transcoding conditions are applied after we picked the candidate transcoding codec and we always prefer the one that is the same as the origin because that one has the possibility to be copied. I'm not really sure if changing the behavior all over the place is better than just split the profiles and check them individually, especially for a bugfix release.

dmitrylyzo:
> To make your theory work we need to do better to check if the selected stream can be remuxed here:

Looks familiar 😅 
https://github.com/jellyfin/jellyfin/pull/12689

> This would require changing even more logic because the transcoding conditions are applied after we picked the candidate transcoding codec and we always prefer the one that is the same as the origin because that one has the possibility to be copied. I'm not really sure if changing the behavior all over the place is better than just split the profiles and check them individually, especially for a bugfix release.

Tbf, I thought `DynamicHlsController` was doing something with the codec list, but looking at [this](https://github.com/jellyfin/jellyfin/blob/b89877554cf1a655251958e19439926ac99c91f7/Jellyfin.Api/Helpers/StreamingHelpers.cs#L101), I think we can (_have to_) just rearrange the codecs in the list (from best to worst). But we must take into account the user's preferred codec. 🤷‍♂️

As for splitting, shouldn't we split by any audio codec with constraints?
From my example:
video `MKV+DTS5.1`, profile `TS+AAC,FLAC` and FLAC is Stereo only. 
We have DTS, but should split by FLAC: `TS+AAC` and `TS+FLAC`.

Ideally, we should also not change the order of audio codecs (codec priority).

gnattu:
> Ideally, we should also not change the order of audio codecs (codec priority).

I managed to achieve this without significantly altering the existing logic, which merges your logic to test all codec profiles and re-check all conditions before selecting the direct stream audio. This approach works for both client-split profiles and combined profiles, and the user-preference is still respected.

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -963,9 +982,36 @@ private void BuildStreamVideoItem(
...
+                                             options.Profile.CodecProfiles
+                                                 .Where(i => i.Type == CodecType.VideoAudio &&
+                                                             i.ContainsAnyCodec(audioStreamWithSupportedCodec.Codec, container) &&
+                                                             i.ApplyConditions.All(applyCondition =>
+                                                             {
+                                                                 var satisfied = ConditionProcessor.IsVideoAudioConditionSatisfied(applyCondition, audioStreamWithSupportedCodec.Channels, audioStreamWithSupportedCodec.BitRate, audioStreamWithSupportedCodec.SampleRate, audioStreamWithSupportedCodec.BitDepth, audioStreamWithSupportedCodec.Profile, false);
+                                                                 if (!satisfied)
+                                                                 {
+                                                                     playlistItem.TranscodeReasons |= GetTranscodeReasonForFailedCondition(applyCondition);
```
dmitrylyzo, line 996: Don't build `TranscodeReasons` from `ApplyConditions` failures: when `ApplyConditions` fails, we simply shouldn't use that codec profile.

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -963,9 +982,36 @@ private void BuildStreamVideoItem(
...
-            var directAudioStream = audioStreamWithSupportedCodec?.Channels is not null && audioStreamWithSupportedCodec.Channels.Value <= (playlistItem.TranscodingMaxAudioChannels ?? int.MaxValue) ? audioStreamWithSupportedCodec : null;
-
-            var channelsExceedsLimit = audioStreamWithSupportedCodec is not null && directAudioStream is null;
+            var channelsExceedsLimit = audioStreamWithSupportedCodec is not null && audioStreamWithSupportedCodec.Channels > (playlistItem.TranscodingMaxAudioChannels ?? int.MaxValue);
+
+            var directAudioStreamSatisfied = audioStreamWithSupportedCodec is not null && !channelsExceedsLimit &&
+                                             options.Profile.CodecProfiles
+                                                 .Where(i => i.Type == CodecType.VideoAudio &&
+                                                             i.ContainsAnyCodec(audioStreamWithSupportedCodec.Codec, container) &&
+                                                             i.ApplyConditions.All(applyCondition =>
```
dmitrylyzo, line 991: What happened to the indentation? Here and above. Are we going to align at the end of `=`, `=>`?

gnattu: This is just IDE auto indent. Where should it indent with?

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -862,18 +862,37 @@ private StreamInfo BuildVideoItem(MediaSourceInfo item, MediaOptions options)
...
+                        var transcodingAudioCodecs = ContainerHelper.Split(transcodingProfile.AudioCodec);
+
+                        foreach (var transcodingAudioCodec in transcodingAudioCodecs)
                         {
                             var appliedVideoConditions = options.Profile.CodecProfiles
                                 .Where(i => i.Type == CodecType.VideoAudio &&
-                                    i.ContainsAnyCodec(audioCodec, container) &&
-                                    i.ApplyConditions.All(applyCondition => ConditionProcessor.IsVideoAudioConditionSatisfied(applyCondition, audioChannels, audioBitrate, audioSampleRate, audioBitDepth, audioProfile, false)))
+                                            i.ContainsAnyCodec(transcodingAudioCodec, container) &&
+                                            i.ApplyConditions.All(applyCondition => ConditionProcessor.IsVideoAudioConditionSatisfied(applyCondition, audioChannels, audioBitrate, audioSampleRate, audioBitDepth, audioProfile, false)))
```
Bond-009, line 876:
```suggestion
                                    i.ContainsAnyCodec(transcodingAudioCodec, container) &&
                                    i.ApplyConditions.All(applyCondition => ConditionProcessor.IsVideoAudioConditionSatisfied(applyCondition, audioChannels, audioBitrate, audioSampleRate, audioBitDepth, audioProfile, false)))
```

Bond-009 (approved):

crobibero (approved):

PR #13218, opened 2024-12-13
Fix missing episode removal

Shadowghost:
Missing episodes should be removed if the actual episode was added.

**Changes**
* Extend virtual episode removal to also take missing episodes into account

### DIFF
--- a/MediaBrowser.Providers/TV/SeriesMetadataService.cs
+++ b/MediaBrowser.Providers/TV/SeriesMetadataService.cs
@@ -147,23 +147,32 @@ namespace MediaBrowser.Providers.TV
 
             foreach (var seasonEpisodes in episodesBySeason)
             {
-                // Only consider non-physical episodes
-                var nonPhysicalEpisodes = seasonEpisodes.Where(e => e.IsVirtualItem || e.IsMissingEpisode).ToList();
-                var physicalEpisodes = seasonEpisodes.Except(nonPhysicalEpisodes);
-                foreach (var episode in nonPhysicalEpisodes)
+                List<Episode> nonPhysicalEpisodes = [];
+                List<Episode> physicalEpisodes = [];
+                foreach (var episode in seasonEpisodes)
                 {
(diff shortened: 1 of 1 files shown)


MediaBrowser.Providers/TV/SeriesMetadataService.cs
```diff
@@ -140,38 +140,30 @@ private void RemoveObsoleteSeasons(Series series)
...
-                    DeleteEpisode(currentEpisode);
-                    continue;
-                }
-
-                for (var j = i + 1; j < numberOfEpisodes; j++)
+            foreach (var seasonEpisodes in episodesBySeason)
+            {
+                // Only consider non-physical episodes
+                var nonPhysicalEpisodes = seasonEpisodes.Where(e => e.IsVirtualItem || e.IsMissingEpisode).ToList();
+                var physicalEpisodes = seasonEpisodes.Except(nonPhysicalEpisodes);
```
Bond-009, line 152: I feel like there should exist a better way of doing this...
Worst case we write something ourselves.

MediaBrowser.Providers/TV/SeriesMetadataService.cs
```diff
@@ -140,38 +140,30 @@ private void RemoveObsoleteSeasons(Series series)
...
                         continue;
                     }
 
-                    DeleteEpisode(currentEpisode);
-                    break;
+                    var physicalEpisodeFound = physicalEpisodes.Any(e => e.ContainsEpisodeNumber(episode.IndexNumber.Value));
+                    if (physicalEpisodeFound)
+                    {
+                        DeleteEpisode(episode);
+                    }
```
Bond-009, line 166: I liked the previous control flow of this loop better, a single `DeleteEpisode` at the end and early exists if one of the criteria doesn't match

Bond-009 (changes requested):

Bond-009 (approved):

crobibero (approved):

PR #13231, opened 2024-12-15
Improve SkiaEncoder's font handling

### LINKED ISSUE #10848: [Issue]: Broken Unicode Emoji in library view
### Please describe your bug

Unicode emojis are not properly handled, when they are placed in Name of the library.

Reproduction steps:
1. Create new library
2. Give it a name containing Unicode character (like `Movies🎥`)
3. See in library miniature (check screenshot)

### Jellyfin Version

10.8.z

### if other:

_No response_

### Environment

```markdown
- OS: Linux
- Linux Kernel: 6.2.0
- Virtualization: Docker
- Clients: Android, Browser(firefox), Desktop client (linux)
(I don't think other data are relevant for this case, I will provide them if needed)
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

image

### Code of Conduct

### LINKED ISSUE #13281: Chinese characters not readable in generated picture

### Description of the bug

I created a library with a Chinese name (中国电视剧). The picture for the library generated shows the first two characters, others are not readable.

### Reproduction steps

1. Create a new library
2. Give it a Chinese name (e.g. 中国电视剧)
3. Generate metadata
4. Look at the picture generated.

Chinese characters in generated library picture not shown correctly.
Image
Image

All characters are shown.

### Jellyfin Server version

10.10.0+

### Specify commit id

_No response_

2024122305

### Specify version number

_No response_

### Specify the build version

10.11.0

### Environment

```markdown
- OS: Microsoft Windows [Version 10.0.26120.2705]
- FFmpeg Version: 7.0.2
- Playback Method: DLNA
- Hardware Acceleration: NVIDIA
- GPU Model: RTX 2060 Q-Max
- Plugins: DLNA
```

### Jellyfin logs

```shell
N/A
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

gnattu:
Our previous approach didn’t work with some complex library names, even when the required fonts were present, because the font handling logic was too simplistic. Modern Unicode and the fonts have become quite complex, making it challenging to implement it correctly. This improved implementation still isn’t the most correct way, but it’s better than it used to be. It now falls back to multiple fonts to find the best one and also handles extended grapheme clusters that were incorrectly processed before.

**Changes**

**Issues**

Fixes #10848
Fixes #13281

### DIFF
--- a/src/Jellyfin.Drawing.Skia/SkiaEncoder.cs
+++ b/src/Jellyfin.Drawing.Skia/SkiaEncoder.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
+using System.Linq;
 using BlurHashSharp.SkiaSharp;
 using Jellyfin.Extensions;
 using MediaBrowser.Common.Configuration;
@@ -24,6 +25,7 @@ public class SkiaEncoder : IImageEncoder
     private readonly ILogger<SkiaEncoder> _logger;
     private readonly IApplicationPaths _appPaths;
     private static readonly SKImageFilter _imageFilter;
+    private static readonly List<SKTypeface> _typefaces;
 
 #pragma warning disable CA1810
     static SkiaEncoder()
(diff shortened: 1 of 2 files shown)


src/Jellyfin.Drawing.Skia/SkiaEncoder.cs
```diff
@@ -24,6 +25,7 @@ public class SkiaEncoder : IImageEncoder
     private readonly ILogger<SkiaEncoder> _logger;
     private readonly IApplicationPaths _appPaths;
     private static readonly SKImageFilter _imageFilter;
+    private static readonly List<SKTypeface> _typefaces;
```
Bond-009, line 28: Any reason this is a List instead of an array?

src/Jellyfin.Drawing.Skia/SkiaEncoder.cs
```diff
@@ -97,6 +114,11 @@ public SkiaEncoder(ILogger<SkiaEncoder> logger, IApplicationPaths appPaths)
     public IReadOnlyCollection<ImageFormat> SupportedOutputFormats
         => new HashSet<ImageFormat> { ImageFormat.Webp, ImageFormat.Jpg, ImageFormat.Png, ImageFormat.Svg };
 
+    /// <summary>
+    /// Gets the default typeface to use.
+    /// </summary>
+    public static SKTypeface DefaultTypeFace => _typefaces.Last();
```
Bond-009, line 120: Wouldn't it make more sense to put the default first?

gnattu: When we iterate to find best fonts, we start from the beginning of the typefaces array. However, since we begin with the default font, having it at the beginning of the list would always result in a lookup miss for the first fallback try, which I want to avoid.

src/Jellyfin.Drawing.Skia/SkiaEncoder.cs
gnattu: No, I just picked it with early draft but now it can be an array as it is now intended to be readonly

src/Jellyfin.Drawing.Skia/SkiaEncoder.cs
gnattu: The thing is, we never had a “preferred font” like what most websites do. What we used to do was simply use the system’s default sans-serif font. If the library title contained a CJK character, we would select the system’s default fonts for CJK characters. This way, we were always using the system’s default fonts and if the title is pure ANSI characters we were explicitly picking the default sans-serif font. I’m keeping this logic in place to ensure that the first CJK fonts don’t override the default sans-serif preference.

src/Jellyfin.Drawing.Skia/StripCollageBuilder.cs
```diff
@@ -200,4 +201,110 @@ private SKBitmap BuildSquareCollageBitmap(IReadOnlyList<string> paths, int width
...
+    /// <summary>
+    /// Draw shaped text with given SKPaint.
+    /// </summary>
+    /// <param name="canvas">If not null, draw text to this canvas, otherwise only measure the text width.</param>
+    /// <param name="x">x position of the canvas to draw text.</param>
+    /// <param name="y">y position of the canvas to draw text.</param>
+    /// <param name="text">The text to draw.</param>
+    /// <param name="textPaint">The SKPaint to style the text.</param>
+    /// <returns>The width of the text.</returns>
+    private static float DrawTextWithoutFallback(SKCanvas? canvas, float x, float y, string text, SKPaint textPaint)
```
JPVenson, line 214:
```suggestion
    private static float MeasureAndDrawText(SKCanvas? canvas, float x, float y, string text, SKPaint textPaint)
```

JPVenson (approved):

Shadowghost (approved):

crobibero (approved):

PR #13236, opened 2024-12-15
feat: allow grouping shows into collections

jheuel:
Group shows into collections

**Changes**
Allows grouping shows into collections similar to how it works for movies
See feature request: https://features.jellyfin.org/posts/333/group-shows-into-collections

**Issues**

Goes with https://github.com/jellyfin/jellyfin-web/pull/6389

### DIFF
--- a/Jellyfin.Server/Migrations/PreStartupRoutines/RenameEnableGroupingIntoCollections.cs
+++ b/Jellyfin.Server/Migrations/PreStartupRoutines/RenameEnableGroupingIntoCollections.cs
@@ -1,0 +1,63 @@
+using System;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Xml.Linq;
+using Emby.Server.Implementations;
+using Microsoft.Extensions.Logging;
+
+namespace Jellyfin.Server.Migrations.PreStartupRoutines;
+
+/// <inheritdoc />
+public class RenameEnableGroupingIntoCollections : IMigrationRoutine
+{
+    private readonly ServerApplicationPaths _applicationPaths;
+    private readonly ILogger<RenameEnableGroupingIntoCollections> _logger;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="RenameEnableGroupingIntoCollections"/> class.
+    /// </summary>
+    /// <param name="applicationPaths">An instance of <see cref="ServerApplicationPaths"/>.</param>
+    /// <param name="loggerFactory">An instance of the <see cref="ILoggerFactory"/> interface.</param>
+    public RenameEnableGroupingIntoCollections(ServerApplicationPaths applicationPaths, ILoggerFactory loggerFactory)
+    {
+        _applicationPaths = applicationPaths;
+        _logger = loggerFactory.CreateLogger<RenameEnableGroupingIntoCollections>();
+    }
+
+    /// <inheritdoc />
+    public Guid Id => Guid.Parse("E73B777D-CD5C-4E71-957A-B86B3660B7CF");
+
+    /// <inheritdoc />
+    public string Name => nameof(RenameEnableGroupingIntoCollections);
+
+    /// <inheritdoc />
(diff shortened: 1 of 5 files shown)


MediaBrowser.Model/Configuration/ServerConfiguration.cs
```diff
@@ -199,7 +199,9 @@ public ServerConfiguration()
 
     public bool EnableFolderView { get; set; } = false;
 
-    public bool EnableGroupingIntoCollections { get; set; } = false;
+    public bool EnableGroupingMoviesIntoCollections { get; set; } = false;
```
sjorge, line 202: Is there some sort of migration in place to update EnableGroupingIntoCollections into EnableGroupingMoviesIntoCollections ? (Assuming these are saved in the db)

jheuel: Happy to implement some migration logic if you point me to a good place for it. I am not very familiar with the project and you are looking at my first lines of C#.

sjorge: This PR just caught my interest so I am not 100% sure how the migration all hook together, but they are located here https://github.com/jellyfin/jellyfin/tree/master/Jellyfin.Server/Migrations

If this is indeed required someone from core can probably point you to a more specific spot.

jheuel: I may have figured out a nice way to migrate the configuration.

Shadowghost (approved):

Jellyfin.Server/Migrations/PreStartupRoutines/RenameEnableGroupingIntoCollections.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+            _logger.LogWarning("Configuration file not found: {Path}", path);
+            return;
+        }
+
+        try
+        {
+            XDocument xmlDocument = XDocument.Load(path);
+            var element = xmlDocument.Descendants("EnableGroupingIntoCollections").FirstOrDefault();
+            if (element != null)
```
Bond-009, line 51:
```suggestion
            if (element is not null)
```

Jellyfin.Server/Migrations/PreStartupRoutines/RenameEnableGroupingIntoCollections.cs
```diff
@@ -0,0 +1,63 @@
...
+    public string Name => nameof(RenameEnableGroupingIntoCollections);
+
+    /// <inheritdoc />
+    public bool PerformOnNewInstall => false;
+
+    /// <inheritdoc />
+    public void Perform()
+    {
+        string path = Path.Combine(_applicationPaths.ConfigurationDirectoryPath, "system.xml");
+        if (!File.Exists(path))
```
Bond-009, line 41: No need to check this IMO, the try catch block will catch it if the file doesn't exist.

Bond-009 (changes requested):

Jellyfin.Server/Migrations/PreStartupRoutines/RenameEnableGroupingIntoCollections.cs
jheuel: Yes, but having the clear error message does not hurt.

Bond-009 (approved):

crobibero (approved):

cmcraes:
This branch has been merged and the latest update was after this, but the option still is not present on my docker image (10.10.7). Will it be in the next update?

jheuel:
> This branch has been merged and the latest update was after this, but the option still is not present on my docker image (10.10.7). Will it be in the next update?

It is listed in the [10.11.0 pre-release changelog](https://github.com/jellyfin/jellyfin/releases/tag/v10.11.0-rc2).

PR #13262, opened 2024-12-21
Don't use custom params on ultrafast x265 preset

gnattu:
Our custom parameters are slower than the ultrafast preset, but users would expect encoding to be as fast as possible when selecting ultrafast. Only apply those parameters to superfast and slower presets.

**Changes**

**Issues**

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -2056,13 +2056,18 @@ namespace MediaBrowser.Controller.MediaEncoding
                 param += " -x264opts:0 subme=0:me_range=16:rc_lookahead=10:me=hex:open_gop=0";
             }
 
-            if (string.Equals(videoEncoder, "libx265", StringComparison.OrdinalIgnoreCase) && encodingOptions.EncoderPreset < EncoderPreset.ultrafast)
+            if (string.Equals(videoEncoder, "libx265", StringComparison.OrdinalIgnoreCase))
             {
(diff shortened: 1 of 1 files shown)


MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -2056,11 +2056,12 @@ public string GetVideoQualityParam(EncodingJobInfo state, string videoEncoder, E
...
             }
 
-            if (string.Equals(videoEncoder, "libx265", StringComparison.OrdinalIgnoreCase))
+            if (string.Equals(videoEncoder, "libx265", StringComparison.OrdinalIgnoreCase) && encodingOptions.EncoderPreset < EncoderPreset.ultrafast)
             {
                 // libx265 only accept level option in -x265-params.
                 // level option may cause libx265 to fail.
                 // libx265 cannot adjust the given level, just throw an error.
+                // The following params are slower than the ultrafast preset, don't use when ultrafast is selected.
                 param += " -x265-params:0 subme=3:merange=25:rc-lookahead=10:me=star:ctu=32:max-tu-size=32:min-cu-size=16:rskip=2:rskip-edge-threshold=2:no-sao=1:no-strong-intra-smoothing=1:no-scenecut=1:no-open-gop=1:no-info=1";
```
nyanmisaka, line 2065: These three params are always required

```
:no-scenecut=1:no-open-gop=1:no-info=1
```

nyanmisaka (approved):

Bond-009 (approved):

Shadowghost (approved):

PR #13274, opened 2024-12-22
Refactoring of DirectoryService: change all direct creation to Depend…

purehaTime:
Hi friends!

I did a little refactoring for removing direct creation of DirectoryService, like 
`new DirectoryService(FileSystem)`

I guess it help reduce strong dependencies and help jellyfin code more flexibility
Also I tried to create plugin whats implement my own DirectoryService and it has problem what it can be injected

please review code

### DIFF
--- a/Emby.Server.Implementations/ApplicationHost.cs
+++ b/Emby.Server.Implementations/ApplicationHost.cs
@@ -636,6 +636,7 @@ namespace Emby.Server.Implementations
             BaseItem.LocalizationManager = Resolve<ILocalizationManager>();
             BaseItem.ItemRepository = Resolve<IItemRepository>();
             BaseItem.FileSystem = Resolve<IFileSystem>();
+            BaseItem.DirectoryService = Resolve<IDirectoryService>();
             BaseItem.UserDataManager = Resolve<IUserDataManager>();
             BaseItem.ChannelManager = Resolve<IChannelManager>();
             Video.RecordingsManager = Resolve<IRecordingsManager>();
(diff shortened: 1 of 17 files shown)


gnattu (changes requested):
I’m afraid we cannot simply replace the current implementation of DirectoryService for multiple reasons:

- The underlying `IFileSystem` may differ, and we cannot assume that all DirectoryServices share the same `IFileSystem` instance.
- DirectoryService instances maintain their own caches, and most cached data is not intended to have an unlimited lifetime. The change to the caching behavior may cause unexpected side effects due to stale cached data being passed to the logics that expect empty cache.
- The current design, where everything is a `BaseItem`, is an Emby legacy that has been proven to be flawed. Therefore, the changes made here are not future-proof, although the future is still quite distant.

Also, I am uncertain whether DirectoryService was originally designed to be implemented by plugins because a lot of the logics assumes the implementation should behave like the server’s original code and it would be very likely to cause unwanted side effects if the behavior is altered.

purehaTime:
Ok, thanks!

Bond-009:
> The underlying IFileSystem may differ, and we cannot assume that all DirectoryServices share the same IFileSystem instance.

IMO that it's pretty safe to assume that

> DirectoryService instances maintain their own caches, and most cached data is not intended to have an unlimited lifetime. The change to the caching behavior may cause unexpected side effects due to stale cached data being passed to the logics that expect empty cache.

And that's why we should just remove `DirectoryService` completely. I already did a simple benchmark and didn't notice any worse scanning perf.

purehaTime:
> And that's why we should just remove DirectoryService completely
So, I guess it is better to remove field and constructor form
`public ImageRefreshOptions(IDirectoryService directoryService)`
didn't see any usages, only direct parameters to other methods, that can be change by DI
(usages only ProbeProvider)

I can make new PR to remove it as field for first iteration to remove it, if it is okay 

@Bond-009 @gnattu fyi

PR #13288, opened 2024-12-27
Fix DTS in HLS

Shadowghost:
`DTS` should usually have codec name `dts` and encoder `dca`. We're currently only adding `dca` to the codec list which prevents remuxing of DTS when streaming via fMP4.

**Changes**
* Replace `dca` with `dts` in supported HLS audio codecs and shift logic

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -7070,7 +7070,7 @@ namespace MediaBrowser.Controller.MediaEncoding
             {
                 // DTS and TrueHD are not supported by HLS
                 // Keep them in the supported codecs list, but shift them to the end of the list so that if transcoding happens, another codec is used
-                shiftAudioCodecs.Add("dca");
+                shiftAudioCodecs.Add("dts");
                 shiftAudioCodecs.Add("truehd");
             }
             else
--- a/MediaBrowser.Model/Dlna/StreamBuilder.cs
+++ b/MediaBrowser.Model/Dlna/StreamBuilder.cs
@@ -30,7 +30,7 @@ namespace MediaBrowser.Model.Dlna
         private readonly ITranscoderSupport _transcoderSupport;
         private static readonly string[] _supportedHlsVideoCodecs = ["h264", "hevc", "vp9", "av1"];
         private static readonly string[] _supportedHlsAudioCodecsTs = ["aac", "ac3", "eac3", "mp3"];
-        private static readonly string[] _supportedHlsAudioCodecsMp4 = ["aac", "ac3", "eac3", "mp3", "alac", "flac", "opus", "dca", "truehd"];
+        private static readonly string[] _supportedHlsAudioCodecsMp4 = ["aac", "ac3", "eac3", "mp3", "alac", "flac", "opus", "dts", "truehd"];
 
         /// <summary>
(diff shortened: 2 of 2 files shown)


Bond-009 (approved):

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -30,7 +30,7 @@ public class StreamBuilder
         private readonly ITranscoderSupport _transcoderSupport;
         private static readonly string[] _supportedHlsVideoCodecs = ["h264", "hevc", "vp9", "av1"];
         private static readonly string[] _supportedHlsAudioCodecsTs = ["aac", "ac3", "eac3", "mp3"];
-        private static readonly string[] _supportedHlsAudioCodecsMp4 = ["aac", "ac3", "eac3", "mp3", "alac", "flac", "opus", "dca", "truehd"];
+        private static readonly string[] _supportedHlsAudioCodecsMp4 = ["aac", "ac3", "eac3", "mp3", "alac", "flac", "opus", "dca", "dts", "truehd"];
```
gnattu, line 33:
```suggestion
        private static readonly string[] _supportedHlsAudioCodecsMp4 = ["aac", "ac3", "eac3", "mp3", "alac", "flac", "opus", "dts", "truehd"];
```

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -7070,6 +7070,7 @@ private void ShiftAudioCodecsIfNeeded(List<string> audioCodecs, MediaStream audi
             {
                 // DTS and TrueHD are not supported by HLS
                 // Keep them in the supported codecs list, but shift them to the end of the list so that if transcoding happens, another codec is used
+                shiftAudioCodecs.Add("dts");
                 shiftAudioCodecs.Add("dca");
```
gnattu, line 7074:
```suggestion
```

MediaBrowser.Model/Dlna/StreamBuilder.cs
```diff
@@ -1721,12 +1721,6 @@ private void ApplyTranscodingConditions(StreamInfo item, IEnumerable<ProfileCond
                     continue;
                 }
 
-                // No way to express this
-                if (condition.Condition == ProfileConditionType.GreaterThanEqual)
-                {
-                    continue;
-                }
-
```
gnattu, line 1729: Don't change such behavior in bugfix releases as it might break some weird clients.

gnattu (changes requested):
I think we can just drop dca from the list as it should always be dts.

Jellyfin.Api/Controllers/DynamicHlsController.cs
gnattu, line 1: I'm not a big fan of having so many cosmetic changes on unrelated files. This will cause numerous merge conflicts for backporting and would require rebasing for current PRs.

gnattu (changes requested):

Jellyfin.Api/Controllers/DynamicHlsController.cs
Shadowghost: I totally agree. Currently not on my main rif and apparently auto-fromat was enabled. Fixed it.

gnattu (approved):

Bond-009 (approved):

Bond-009 (approved):

crobibero (approved):

PR #13306, opened 2024-12-31
Add option to disable deprecated legacy authorization options

nielsvanvelzen:
**Changes**

Adds a new `EnableLegacyAuthorization` configuration option to `system.xml`. It defaults to `true` (keep legacy methods working) but we will add a migration to change it to `false` in the future before entirely removing the old authorization methods.

For now the option can be used by (client-)developers to find network requests using a wrong authorization method.

The only method we'll allow is the `Authorization` header with `MediaBrowser` scheme and the `ApiKey` query parameter. The other headers (`X-Emby-Authorization`, `X-Emby-Token`, `X-MediaBrowser-Token`), query parameter (`api_key`) and authorization scheme (`Emby`) are all deprecated.

We've warned for these changes for almost 2 years using [this gist](https://gist.github.com/nielsvanvelzen/ea047d9028f676185832e51ffaf12a6f) we share almost weekly to people developing for Jellyfin.

Note: It is currently not possible to use jellyfin-web without legacy authorization (unable to sign in due to [this line](https://github.com/jellyfin-archive/jellyfin-apiclient-javascript/blob/master/src/connectionManager.js#L380) in the archived JS apiclient).

**Issues**

### DIFF
--- a/Jellyfin.Server.Implementations/Security/AuthorizationContext.cs
+++ b/Jellyfin.Server.Implementations/Security/AuthorizationContext.cs
@@ -7,6 +7,7 @@ using System.Threading.Tasks;
 using Jellyfin.Data.Queries;
 using Jellyfin.Extensions;
 using MediaBrowser.Controller;
+using MediaBrowser.Controller.Configuration;
 using MediaBrowser.Controller.Devices;
 using MediaBrowser.Controller.Library;
 using MediaBrowser.Controller.Net;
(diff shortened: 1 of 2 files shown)


crobibero (approved):

JPVenson (approved):

crobibero:
Marking as blocked until the web client moves from the legacy auth

PR #13307, opened 2025-01-01
always sort season by index number

qiqian:
**Changes**
always sort seasons by their index number so that seasons are in correct order

**Issues**
seasons from fairy tails are in wrong order 
https://www.themoviedb.org/tv/46261-fairy-tail/seasons


Shadowghost (approved):

theguymadmax:
I don't have any issues sorting that series (or any other) with the current code. If you're encountering problems, I suspect it might be a naming issue. This PR breaks the sorting-by-name feature for series, which should take priority.

JPVenson (approved):

JPVenson (approved):

PR #13313, opened 2025-01-02
Use nv15 as intermediate format for 2-pass rkrga scaling

gnattu:
It seems like RKRGA is expecting no zero padding bits during 2 pass scaling which distorts P010 frames after scale. This breaks trickplay image generation for HDR videos. Workaround this by specifying nv15 frame format.

This might cause issues with other ffmpeg implementations, as NV15 is not supported everywhere. However, rockchip users should use jellyfin-ffmpeg for proper hardware acceleration anyway.

**Changes**

**Issues**

Fixes this: 
image

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -5695,7 +5695,8 @@ namespace MediaBrowser.Controller.MediaEncoding
                     if (!string.IsNullOrEmpty(doScaling)
                         && !IsScaleRatioSupported(inW, inH, reqW, reqH, reqMaxW, reqMaxH, 8.0f))
                     {
-                        // RKRGA is expecting NV15(P010 with no zero padding bits) when doing 2-pass scaling. Not setting this will cause distorted frames.
+                        // Vendor provided BSP kernel has an RGA driver bug that causes the output to be corrupted for P010 format.
+                        // Use NV15 instead of P010 to avoid the issue.
                         // SDR inputs are using BGRA formats already which is not affected.
                         var intermediateFormat = string.Equals(outFormat, "p010", StringComparison.OrdinalIgnoreCase) ? "nv15" : outFormat;
                         var hwScaleFilterFirstPass = $"scale_rkrga=w=iw/7.9:h=ih/7.9:format={intermediateFormat}:force_divisible_by=4:afbc=1";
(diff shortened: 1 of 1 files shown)


Shadowghost (approved):

Bond-009 (approved):

nyanmisaka:
RGA3 does support nv15 and p010 inputs, but there is a kernel issue that needs to be fixed via a patch. Most common distros include it. Without it RGA3 will treat p010 as nv15.

https://github.com/armbian/linux-rockchip/commit/03384d158994c62c4378677d54d6d6b69f1e34d0

https://github.com/search?q=repo%3Aarmbian%2Flinux-rockchip+nyanmisaka&type=commits)

gnattu:
> RGA3 does support nv15 and p010 inputs, but there is a kernel issue that needs to be fixed via a patch. Most common distros include it. Without it RGA3 will treat p010 as nv15.

Now the million dollar question: how can we tell from the server? Because most if not all kernels shipped by board vendors are affected and buggy and those are the kernels being used most.

nyanmisaka:
> > RGA3 does support nv15 and p010 inputs, but there is a kernel issue that needs to be fixed via a patch. Most common distros include it. Without it RGA3 will treat p010 as nv15.
> 
> Now the million dollar question: how can we tell from the server? Because most if not all kernels shipped by board vendors are affected and buggy and those are the kernels being used most.

Just memcmp one rawvideo frame of p010 testsrc with the p010->p010 output after the rkrga filter.

Their checksums are the same when there is no glitches.

nyanmisaka (approved):
But that's a bit of overkill. To mitigate the unpatched vendor kernel, this PR should be enough.

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -5695,7 +5695,10 @@ public string GetVideoTransposeDirection(EncodingJobInfo state)
                     if (!string.IsNullOrEmpty(doScaling)
                         && !IsScaleRatioSupported(inW, inH, reqW, reqH, reqMaxW, reqMaxH, 8.0f))
                     {
-                        var hwScaleFilterFirstPass = $"scale_rkrga=w=iw/7.9:h=ih/7.9:format={outFormat}:afbc=1";
+                        // RKRGA is expecting NV15(P010 with no zero padding bits) when doing 2-pass scaling. Not setting this will cause distorted frames.
```
nyanmisaka, line 5698: And the comments should be updated to reflect this.

Issue #12395, opened 2024-08-05
Jellyfin keeps trying to scan libraries/directories that no longer exist.


erichuecker:
### This issue respects the following points:

- [X] This is a **bug**, not a question or a configuration issue; Please visit our forum or chat rooms first to troubleshoot with volunteers, before creating a report. The links can be found [here](https://jellyfin.org/contact/).
- [X] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [X] I'm using an up to date version of Jellyfin Server stable, unstable or master; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.
- [X] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [X] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Description of the bug

I removed a library and now whenever I initiate a scan from the Scheduled Tasks screen the scan fails. I checked the logs and it's still trying to Scan the library that I deleted. I have to go and manually add this directory back to the config/root/default directory to get the scanning to work again.

[ERR] [30] MediaBrowser.Controller.Entities.BaseItem: Error refreshing owned items for "/config/root/default/DELETEDLIBRARY"
System.IO.DirectoryNotFoundException: Could not find a part of the path '/config/root/default/DELETEDLIBRARY'

I've tried rolling back to previous server versions with no luck. Anyone have any ideas? Is there some sort of cache I can clear so Jellyfin will know those directories/libraries no longer exist?

### Reproduction steps

remove a library, execute a scan, scan fails after trying to update non-existent library.

### What is the current _bug_ behavior?

can't scan library because jellyfin is trying to scan libraries that no long exist.

### What is the expected _correct_ behavior?

jellyfin should only scan directories that I have configured in Jellyfin.

### Jellyfin Server version

10.9.9+

### Specify the build version

10.9.9

### Environment

```markdown
- OS: Ubuntu Server running docker
```

### Jellyfin logs

```shell
[2024-08-05 15:33:36.777 -05:00] [ERR] [30] MediaBrowser.Controller.Entities.BaseItem: Error refreshing owned items for "/config/root/default/---"
System.IO.DirectoryNotFoundException: Could not find a part of the path '/config/root/default/---'.
[...]

PR #13315, opened 2025-01-02
Removing CollectionFolders from cache when they are deleted on disk.

sinterdev:
If the user has somehow deleted a collection on disk, it can prevent other collections from being scanned properly.

**Changes**
Added a few lines to synchronize the state between the cached collections and the collections in the root Jellyfin directory.

**Issues**
#12395

**Testing**

You can test this change by...
- Creating a new collection.
- Deleting the collection from `%LOCALAPPDATA%\jellyfin\root\default`.
- Executing the "Scan Media Library" task in "Dashboard > Scheduled Tasks".

### DIFF
--- a/Emby.Server.Implementations/Library/LibraryManager.cs
+++ b/Emby.Server.Implementations/Library/LibraryManager.cs
@@ -1053,19 +1053,16 @@ namespace Emby.Server.Implementations.Library
                 cancellationToken: cancellationToken).ConfigureAwait(false);
 
             // Quickly scan CollectionFolders for changes
-            foreach (var child in GetUserRootFolder().Children)
+            foreach (var child in GetUserRootFolder().Children.OfType<Folder>())
             {
-                if (child is Folder folder)
+                // If the user has somehow deleted the collection directory, remove the metadata from the database.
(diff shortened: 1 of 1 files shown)


Emby.Server.Implementations/Library/LibraryManager.cs
```diff
@@ -1053,9 +1053,20 @@ await GetUserRootFolder().ValidateChildren(
                 cancellationToken: cancellationToken).ConfigureAwait(false);
 
             // Quickly scan CollectionFolders for changes
-            foreach (var folder in GetUserRootFolder().Children.OfType<Folder>())
+            foreach (var child in GetUserRootFolder().Children)
             {
-                await folder.RefreshMetadata(cancellationToken).ConfigureAwait(false);
+                if (child is Folder folder)
```
baka0815, line 1058: Why do you change the code from ``.OfType<Folder>`` to ``if (child is Folder folder)``?

sinterdev: Ha! Yeah, good point. I changed it back.

Shadowghost (approved):

Bond-009 (approved):
Not 100% sure this is the best place for it, but we can always move it somewhere else later.

Issue #13119, opened 2024-11-28
Backdrops not visible for albums


przemek54:
### This issue respects the following points:

- [x] This is a **bug**, not a question or a configuration issue; Please visit our forum or chat rooms first to troubleshoot with volunteers, before creating a report. The links can be found [here](https://jellyfin.org/contact/).
- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [x] I'm using an up to date version of Jellyfin Server stable, unstable or master; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Description of the bug

Selecting a backdrop image for music albums specifically results in no backdrop image being displayed, despite the "enable backdrops" option being on. In the "edit images" tab, the backdrop appears as a blank rectangle that is impossible to delete (except through refreshing the album's metadata with the replace images option). Notably, this only seems to affect music albums. All other media types (movies, artists) function as expected.

There is a way to get a backdrop on an album's page, and that is to add a backdrop to its album artist. However, it means that it's currently impossible to assign different backdrops to different albums from the same artist. Interestingly, when any album artist in the library has a backdrop image, albums that I have uploaded backdrops to also share that same image, even if they don't belong to the same artist, and even when the image is totally different.

The uploaded backdrop does appear in the appropriate folder of the album as "backdrop.png" (or .jpg), so it doesn't seem to be an issue of sending the image to the correct location.

NOTE: Not sure how relevant it is, but refreshing the album's metadata (with the "replace all metadata" option and "replace existing images" checked on) also removes the backdrop of its album artist. Maybe it's an intended feature.

### Reproduction steps

1. Sign in to Jellyfin
2. Navigate to a library that contains music
3. Make sure that no artist has a backdrop image
4. Select the "Albums" tab on the top bar
5. Click any album
6. Under the three-dots menu, click "Edit images"
7. Click the "+" icon and select an image
8. Set the "image type" to "backdrop"
9. Click "upload"
10. Click off the "Edit images" menu and refresh. There is likely no backdrop visible
11. Open the "Edit images" menu again
[...]

PR #13317, opened 2025-01-03
Fixing backdrop image saving for music albums.

sinterdev:
Currently, music albums force their backdrop images to be derived from their parent artist instead of allowing users to upload their own backdrop images.

**Changes**
Changed `BaseItem.GetImageInfo` to prefer existing images over derived images if they are available.

**Issues**
#13119

### DIFF
--- a/MediaBrowser.Controller/Entities/BaseItem.cs
+++ b/MediaBrowser.Controller/Entities/BaseItem.cs
@@ -1915,7 +1915,7 @@ namespace MediaBrowser.Controller.Entities
                 throw new ArgumentException("Cannot set chapter images using SetImagePath");
             }
 
-            var image = GetImageInfo(type, index);
+            var image = GetImageInfo(type, index, false);
 
             if (image is null)
             {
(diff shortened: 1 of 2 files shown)


Shadowghost (approved):

MediaBrowser.Controller/Entities/BaseItem.cs
```diff
@@ -2053,8 +2054,10 @@ public ItemImageInfo GetImageInfo(ImageType imageType, int imageIndex)
                 };
             }
 
+            var result = GetImages(imageType).ElementAtOrDefault(imageIndex);
+
             // Music albums usually don't have dedicated backdrops, so return one from the artist instead
-            if (GetType() == typeof(MusicAlbum) && imageType == ImageType.Backdrop)
+            if (result == null && includeDerived && GetType() == typeof(MusicAlbum) && imageType == ImageType.Backdrop)
```
Bond-009, line 2060: Isn't the null check enough? If so I'd prefer that simple change over adding another param to the function.
Also use `result is null`

Bond-009 (changes requested):

MediaBrowser.Controller/Entities/BaseItem.cs
sinterdev: We want to retain the current behavior of returning the parent image in most cases (~37 callers of `GetImageInfo`), but we also want to be able to distinguish between computed / derived / inherited images and user-specified images in the cases where the user is trying to save a new image manually.

If we just check `result is null`, then we need some other way to determine whether the returned image is derived or not derived. I suppose we could add a flag / enum to `ItemImageInfo` denoting the type of image it is (derived or not derived). Then, the callers could use that to make a decision when saving the image.

PR #13332, opened 2025-01-07
Improve video resolution filtering and classification logic

### LINKED ISSUE #13413: 4K Movies not Filtered when "HD" Video Type is Selected
**Describe The Bug**
When using the Jellyfin web interface or the desktop media player on macOS, I noticed a problem with the "Video Types" filter. My library is made up of rips of my personal DVDs, Blu-rays and UHD Blu-rays which have been remuxed into MKV files. This is what happens when I select the "SD", "HD" and "4K" filters:
- SD: 7 / 101 movies are visible, all of which are my DVD rips
- 4K: 70 / 101 movies are visible, all of which are my UHD Blu-ray rips
- HD: 94 / 101 movies are visible, both my UHD Blu-ray and FHD Blu-ray rips are shown

This is unintuitive to me because I have a means of only seeing my DVDs, and only seeing my UHD Blu-rays, but I don't have a way of only seeing my 1080p Blu-rays.

**Steps To Reproduce**
1. Create a movie library
2. Add an MKV file with a resolution of 3840x2160
3. Add an MKV file with a resolution of 1920x1080
4. Add an MKV file with a resolution of 720x480
5. Go to the movie library
6. Click on the filter button
7. Click on "Video Types"
8. Click on "HD"
9. Notice how both the 1080p and 2160p videos are visible

**Expected Behavior**
Given my library of movies I would expect the following outcomes:
- SD: 7 / 101 movies are visible, same as the current behaviour
- 4K: 70 / 101 movies are visible, same as the current behaviour
- HD: 24 / 101 movies are visible, all of which are my FHD Blu-ray rips

**Logs**
N/A

**Screenshots**
image

 - Platform: macOS 14.6.1
 - Browser: Brave (latest version)
 - Jellyfin Version: 10.9.10

**Additional Context**
I hope you agree with me that the current behaviour is incorrect, I can see why people might want 4K movies to display when they select a "HD" video type because 4K is often referred to as a high definition format but if the current behaviour is maintained then I would appreciate a "FHD" video type as that would be very helpful.

### LINKED ISSUE #13792: SD filter not working on 10.11
### Description of the bug

Applying the SD filter on 10.11 does not filter correctly and returns the same results as the HD filter. 

### Reproduction steps

1. Run an instance of a server 10.11
2. Apply an SD filter to a library

Filter returns the same query has a HD filter

Filter should only return SD media

### Jellyfin Server version

Master

### Specify commit id

869d884e54cdce7340377064007c513a672ec314

_No response_

### Specify version number

_No response_

### Specify the build version

Dev Server

### Environment

```markdown
- OS: Windows 11
- Browser: Firefox
- Plugins: None
```

### Jellyfin logs

```shell
N/A
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

theguymadmax:
**Changes**
- Add height-based detection for UHD/4K content (2100px threshold)
- Prevent HD queries from including UHD/4K content
- Logic now supports multiple resolution selections simultaneously

**Issues**
 Fixes #13413 
Fixes #13792

### DIFF
--- a/Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
+++ b/Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
@@ -1304,34 +1304,39 @@ public sealed class BaseItemRepository
         JellyfinDbContext context,
         InternalItemsQuery filter)
     {
+        const int HDWidth = 1200;
+        const int UHDWidth = 3800;
+        const int UHDHeight = 2100;
+
         var minWidth = filter.MinWidth;
         var maxWidth = filter.MaxWidth;
         var now = DateTime.UtcNow;
 
-        if (filter.IsHD.HasValue)
+        if (filter.IsHD.HasValue || filter.Is4K.HasValue)
         {
-            const int Threshold = 1200;
-            if (filter.IsHD.Value)
-            {
-                minWidth = Threshold;
-            }
-            else
-            {
-                maxWidth = Threshold - 1;
-            }
-        }
+            bool includeSD = false;
+            bool includeHD = false;
+            bool include4K = false;
 
-        if (filter.Is4K.HasValue)
-        {
-            const int Threshold = 3800;
-            if (filter.Is4K.Value)
+            if (filter.IsHD.HasValue && !filter.IsHD.Value)
             {
-                minWidth = Threshold;
+                includeSD = true;
             }
-            else
+
+            if (filter.IsHD.HasValue && filter.IsHD.Value)
             {
-                maxWidth = Threshold - 1;
+                includeHD = true;
             }
+
+            if (filter.Is4K.HasValue && filter.Is4K.Value)
+            {
+                include4K = true;
+            }
+
+            baseQuery = baseQuery.Where(e =>
+                (includeSD && e.Width < HDWidth) ||
+                (includeHD && e.Width >= HDWidth && !(e.Width >= UHDWidth || e.Height >= UHDHeight)) ||
+                (include4K && (e.Width >= UHDWidth || e.Height >= UHDHeight)));
         }
 
         if (minWidth.HasValue)


gnattu (changes requested):
This change needs to be discussed further. Not from the code themselves but from the behavior.

The term "4K" carries different meanings in various contexts. However, in the context of commercial cinema and TV video systems, it strictly refers to specific resolution classes with a horizontal pixel count of approximately 4000. Within the DCI standard, the DCI 4K resolution can be expressed as 4096 × 2160, 3996 × 2160, or 4096 × 1716, depending on the frame crop. The SMPTE and ITU-R UHD standards strictly define 4K as the resolution 3840 × 2160. The CEA Ultra HD standard does not use the term "4K" but enforces a minimum requirement of 3840 × 2160 resolution. Additionally, the adoption of ATSC and DVB systems also employs the term "4K" to denote the resolution 3840 × 2160.

On the other hand, stream media platforms and hardware vendors loosely use the term "4K" to refer to "2160p" content. For instance, on YouTube, "4K UHD" means a video has a vertical resolution of 2160 to 3840 lines. Some camera vendors also market their products as "4K cameras", but only capable of shooting 4:3 2880x2160 footages.

The behavior change this PR brings actually switches the 4K definition we are using here. This might not be expected by our users and some of them might disagree with this change because now the 4K query shows "non-4K" contents.

theguymadmax:
I created a discussion thread [here](https://github.com/jellyfin/jellyfin-meta/issues/85).

Shadowghost (approved):
I think this is an overall improvement

crobibero (approved):

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -1304,33 +1304,47 @@ private IQueryable<BaseItemEntity> TranslateQuery(
...
+            {
+                baseQuery = baseQuery.Where(e => e.Width < UHDWidth && e.Height < UHDHeight);
+            }
+        }
+
+        if (filter.IsSD.HasValue)
+        {
+            if (filter.IsSD.Value)
+            {
+                baseQuery = baseQuery.Where(e => e.Width <= (HDWidth - 1));
```
Bond-009, line 1343:
```suggestion
                baseQuery = baseQuery.Where(e => e.Width < HDWidth);
```

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -1304,33 +1304,47 @@ private IQueryable<BaseItemEntity> TranslateQuery(
...
-            const int Threshold = 1200;
             if (filter.IsHD.Value)
             {
-                minWidth = Threshold;
+                baseQuery = baseQuery.Where(e => e.Width >= HDWidth && !(e.Width >= UHDWidth || e.Height >= UHDHeight));
             }
             else
             {
-                maxWidth = Threshold - 1;
+                baseQuery = baseQuery.Where(e => e.Width <= (HDWidth - 1));
```
Bond-009, line 1323:
```suggestion
                baseQuery = baseQuery.Where(e => e.Width < HDWidth );
```

Bond-009 (commented):
I'm not convinced we need an IsSD filter. How is it useful when we can get the same result by combining IsUHD and IsHD?

theguymadmax:
> I'm not convinced we need an IsSD filter. How is it useful when we can get the same result by combining IsUHD and IsHD?

There is a small consistency issue in the web UI. Currently, using the IsHD filter (true or false) for HD/SD limits you to selecting either HD or SD at a time, which creates an inconsistency since other filters allow multiple selections simultaneously. Also, refactoring the web UI with IsSD filter makes the code more consistent overall.

Bond-009:
> > I'm not convinced we need an IsSD filter. How is it useful when we can get the same result by combining IsUHD and IsHD?
> 
> There is a small consistency issue in the web UI. Currently, using the IsHD filter (true or false) for HD/SD limits you to selecting either HD or SD at a time, which creates an inconsistency since other filters allow multiple selections simultaneously. Also, refactoring the web UI with IsSD filter makes the code more consistent overall.

Sounds like a web issue. IMO there is no reason to add an extra filter option to the server (one which makes it possible to contradict oneself I might add!) when the clients just need this bit of logic with the same end result:

```
    if (filterSD) {
        if (!filterHD)
        {
            isHd = false;
        }

        if (!filter4K)
        {
            is4K = false;
        }
    }
    else
    {
        isHd = filterHD;
        is4K = filter4K;
    }
```

theguymadmax:
Ok, IsSD filter has been removed.

gnattu (approved):
Ideally this should also consider portrait videos like the ones shotted on mobile phones. This is fine for now though. We can detect video orientation later and ideally make a dedicated filter for that.

Bond-009:
Does this work correctly when both IsHD and Is4K are true? Or does that return nothing?

theguymadmax:
> Does this work correctly when both IsHD and Is4K are true? Or does that return nothing?

If both IsHD and Is4K are selected, it will return nothing. This is the same behavior as selecting both the SD and 4K filter. Each resolution filter is mutually exclusive and cannot be combined.

JPVenson (approved):

crobibero (approved):

PR #13342, opened 2025-01-09
Prefer ApiKey over api_key in generated URL's

nielsvanvelzen:
See #13306 and [this gist](https://gist.github.com/nielsvanvelzen/ea047d9028f676185832e51ffaf12a6f) for context.

**Changes**

- Use `ApiKey` instead of `api_key` in all URL's returned by the server to clients. Ideally we wouldn't include an API key at all and require the client to set the Authorization header but that's a breaking change for another time.

**Issues**


Bond-009 (approved):

JPVenson (approved):

JPVenson:
while you are at it maybe `HttpUtility.UrlEncode` as a safety precauation?

crobibero (approved):

PR #13352, opened 2025-01-11
Add self to CONTRIBUTORS.md

michaelkrieger:
Add self to contributors. A number of PRs to jellyfin official plug-ins and web.


Bond-009:
This file is for contributors to the server, web has it's own file in their repo: https://github.com/jellyfin/jellyfin-web/blob/master/CONTRIBUTORS.md

PR #13356, opened 2025-01-12
Add channel queries to series

Kevinjil:
Currently, the IChannel interface can deliver channel result folders which are interpreted as series and seasons. However, Jellyfin does not query for the contents of these folders when viewing said serie of season. This results in empty series in the API.

The parsing of series and seasons is implemented in the [ChannelManager](https://github.com/jellyfin/jellyfin/blob/v10.10.3/src/Jellyfin.LiveTv/Channels/ChannelManager.cs#L973). The `IChannel` implementation is never queried for the contents of said series and seasons.
Looking at the implementation of channel [Folders](https://github.com/jellyfin/jellyfin/blob/v10.10.3/MediaBrowser.Controller/Entities/Folder.cs#L951), this implementation for series and seasons seems to fit in the structure of Jellyfin and works with my testing.

**Changes**
- Add query for seasons in series in case the series originates from a channel.
- Add query for episodes in season in case the series originates from a channel.

**Issues**

### DIFF
--- a/MediaBrowser.Controller/Entities/BaseItem.cs
+++ b/MediaBrowser.Controller/Entities/BaseItem.cs
@@ -23,6 +23,7 @@ using MediaBrowser.Controller.Chapters;
 using MediaBrowser.Controller.Configuration;
 using MediaBrowser.Controller.Dto;
 using MediaBrowser.Controller.Entities.Audio;
+using MediaBrowser.Controller.Entities.TV;
 using MediaBrowser.Controller.Library;
 using MediaBrowser.Controller.Persistence;
 using MediaBrowser.Controller.Providers;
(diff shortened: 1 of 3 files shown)


JPVenson:
new features may not target the stable branch. Please target the master instead.

Kevinjil:
Although I see this as a bugfix: these queries were added to the normal folder but not to the other folder types, I'll retarget main.

Bond-009 (approved):

JPVenson (approved):

Shadowghost (approved):
@Kevinjil since the other large database change got merged, can you please rebase again

crobibero (approved):

PR #13382, opened 2025-01-17
Fix interface selection

Shadowghost:
**Changes**
* Use overrides matching the catchall addresses as a last resort (they come last because we first oder descending by subnet size)
* Fix ordering in bind interface selection

### DIFF
--- a/src/Jellyfin.Networking/Manager/NetworkManager.cs
+++ b/src/Jellyfin.Networking/Manager/NetworkManager.cs
@@ -997,7 +997,9 @@ public class NetworkManager : INetworkManager, IDisposable
             // Get interface matching override subnet
             var intf = _interfaces.OrderBy(x => x.Index).FirstOrDefault(x => data.Data.Subnet.Contains(x.Address));
 
-            if (intf?.Address is not null)
+            if (intf?.Address is not null
+                || (data.Data.AddressFamily == AddressFamily.InterNetwork && data.Data.Address.Equals(IPAddress.Any))
+                || (data.Data.AddressFamily == AddressFamily.InterNetworkV6 && data.Data.Address.Equals(IPAddress.IPv6Any)))
             {
                 // If matching interface is found, use override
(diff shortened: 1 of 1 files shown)


JPVenson (approved):

Bond-009 (approved):

gnattu:
The order change breaks local discovery because localhost address is now always sorted at the beginning of the list. What problem this PR was intended to resolve in real world?

src/Jellyfin.Networking/Manager/NetworkManager.cs
```diff
@@ -1062,7 +1065,7 @@ private bool MatchesBindInterface(IPAddress source, bool isInExternalSubnet, out
                 // Check to see if any of the external bind interfaces are in the same subnet as the source.
                 // If none exists, this will select the first external interface if there is one.
                 bindAddress = externalInterfaces
-                    .OrderByDescending(x => x.Subnet.Contains(source))
+                    .OrderBy(x => x.Subnet.Contains(source))
```
gnattu, line 1068: Also, what this does is to put interfaces which having an IP Subnet that does not contains the source IP before the one that does. I don't think this is really the intended behavior here.

Shadowghost: Are you really sure? I tried the sorting in a c# fiddle because I thought this to be the case too, but if you order by a Boolean condition the ones matching the condition comes first.

gnattu: Yes I am really sure: https://dotnetfiddle.net/gKKkJI

A common concept is that false is 0 and true is 1, and dotnet's OrderBy is sorting with this which places false before true.

Shadowghost: Damn, I'm quite sure I tested this exact case. Seems like I butchered something when rebasing before submitting the PR.

PR #13384, opened 2025-01-17
Fix parallel use of not thread-safe SubtitleFormat instance

### LINKED ISSUE #12113: [Issue]: Broken subtitles: Multiple languages and timestamps mixing up on screen when using chromecast
### Please describe your bug

The issue started happening when upgrading from 10.8.13-1 to 10.9.x (I've confirmed 10.9.3 and 10.9.6 are affected).

Most movies with multiple subtitle tracks are affected, even if there is a bit of randomness to the issue and it does seem to depend, to some extent, on the size of the subtitle files (more details below).

Affected movies work just fine in a web browser or in the Android Client (web player). When playing on these platforms the players display the shiny new "Secondary Subtitle" option introduced in https://github.com/jellyfin/jellyfin-web/pull/3906 ; I have not tested the secondary subtitle feature extensively but it seems to work overall.

*However* when casting from my Android phone to my chromecast, the subtitles are completly broken. It generally starts fine for the first couple of subtitles, then quickly subtitles from multiple language tracks (and generally completely different timestamps) are overlayed directly on top of each other making them unreadable.

Note than when casting I have to select *one* language, and the "Secondary Subtitle" option is not available in the remote control. I suspect this is by design as the secondary subtitle feature was only introduced in the HTML player. However the fact that the bug started appearing on 10.9.x and that it only happens when the secondary subtitle feature is unavailable makes me wonder if this bug could be a side effect of introducing secondary subtitles?

Things I have tried to help gather more information:

- I have tried enabling and disabling the "Allow subtitle extraction on the fly" option on the server. This seems to have no effect on the bug.
- I have ran some experiments by modifying the subtitle files of one particular affected movie. I have not found any particular character or sequence of characters causing the bug. However it seems the issue does not happen (or very rarely) if I almost empty the subtitle files to contain only about 10 subtitles each, it happens very often (though not every time) with the complete subtitle files (about 1100 to 1300 subtitles depending on the language), and it seems somewhat random with intermediary sizes.
- The bug seems very strongly correlated (99 if not 100% of occurrences) with the display of a particular error in the logs: `[ERR] 163 errors encountered while parsing '"srt"' subtitle using the "SubRip" format parser`. I am pasting more complete logs below for the same movie, both when it works and when it does not work.

### Reproduction Steps

1. Environment and config: see below
2. Choose a movie with at least two subtitle tracks, for instance two SRT external subtitle files. Make sure those subtitle files have at least 1500 subtitles each.
3. Open the Android client, connect to a google chrome cast
4. Go to the movie on the Android client, select the first subtitle track
5. Hit play
6. Wait for 1-2 minutes after the first dialogues. Subtitles should break rapidly (randomly not showing, multiple languages and timestamps being overlayed).

### Jellyfin Version

10.9.0

### if other:

_No response_

### Environment

```markdown
- OS:Debian Bookworm
- Linux Kernel: Linux 6.1.0-21-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.90-1 (2024-05-03) x86_64 GNU/Linux
- Virtualization: None, running directly on Intel NUC
- Clients: Android
- Browser: N/A
- FFmpeg Version: jellyfin-ffmpeg6 6.0.1-7-bookworm
- Hardware Acceleration: Intel QSV
- GPU Model: 
- Plugins: None
- Reverse Proxy: None
- Base URL: N/A
- Networking: Local home network (IPv6)
- Storage: Local NAS
```

### Jellyfin logs

```shell
*** When it works: ***

[2024-06-15 22:59:31.300 +01:00] [DBG] AuthenticationScheme: "CustomAuthentication" was successfully authenticated.
[2024-06-15 22:59:31.311 +01:00] [DBG] charset "utf-8" detected for "/path/to/library/MovieName.720p.x264.en.srt"
[2024-06-15 22:59:31.316 +01:00] [DBG] Trying to parse '"srt"' subtitle using the "SubRip" format parser
[2024-06-15 22:59:31.336 +01:00] [DBG] AuthenticationScheme: "CustomAuthentication" was successfully authenticated.
[2024-06-15 22:59:31.347 +01:00] [DBG] charset "utf-8" detected for "/path/to/library/MovieName.720p.x264.fr.srt"
[2024-06-15 22:59:31.359 +01:00] [DBG] Trying to parse '"srt"' subtitle using the "SubRip" format parser


*** When the bug happens: ***

[2024-06-15 23:03:00.361 +01:00] [DBG] charset "utf-8" detected for "/path/to/library/MovieName.720p.x264.en.srt"
[2024-06-15 23:03:00.364 +01:00] [DBG] charset "utf-8" detected for "/path/to/library/MovieName.720p.x264.fr.srt"
[2024-06-15 23:03:00.411 +01:00] [DBG] Trying to parse '"srt"' subtitle using the "SubRip" format parser
[2024-06-15 23:03:00.412 +01:00] [DBG] Trying to parse '"srt"' subtitle using the "SubRip" format parser
[2024-06-15 23:03:00.449 +01:00] [ERR] 163 errors encountered while parsing '"srt"' subtitle using the "SubRip" format parser
[2024-06-15 23:03:00.453 +01:00] [ERR] 163 errors encountered while parsing '"srt"' subtitle using the "SubRip" format parser
```

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

alltilla:
**Changes**

`SubtitleFormat`'s `LoadSubtitle()` function is not thread-safe.

A `SubtitleEditParser` instance's `Parse()` function can be called from multiple threads at the same time.

`SubtitleFormat`s are cached in the constructor of each `SubtitleEditParser`, and the same instances are used for each possibly parallel `Parse()` function call, which causes subtitle parse problems.

This patch modifies the code, so we only cache the extension -> `SubtitleFormat` type/class mapping and create a new `SubtitleFormat` instance in each `Parse()` call, so no `SubtitleFormat` instance is accessed from multiple threads.

**Issues**

Fixes #12113

Kudos for everyone investigating the issue there, most notably @RenV123 for PoC-ing the solution.

### DIFF
--- a/MediaBrowser.MediaEncoding/Subtitles/SubtitleEditParser.cs
+++ b/MediaBrowser.MediaEncoding/Subtitles/SubtitleEditParser.cs
@@ -17,7 +17,7 @@ namespace MediaBrowser.MediaEncoding.Subtitles
     public class SubtitleEditParser : ISubtitleParser
     {
         private readonly ILogger<SubtitleEditParser> _logger;
-        private readonly Dictionary<string, SubtitleFormat[]> _subtitleFormats;
+        private readonly Dictionary<string, List<Type>> _subtitleFormatTypes;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="SubtitleEditParser"/> class.
@@ -26,10 +26,7 @@ namespace MediaBrowser.MediaEncoding.Subtitles
         public SubtitleEditParser(ILogger<SubtitleEditParser> logger)
         {
             _logger = logger;
-            _subtitleFormats = GetSubtitleFormats()
-                .Where(subtitleFormat => !string.IsNullOrEmpty(subtitleFormat.Extension))
-                .GroupBy(subtitleFormat => subtitleFormat.Extension.TrimStart('.'), StringComparer.OrdinalIgnoreCase)
-                .ToDictionary(g => g.Key, g => g.ToArray(), StringComparer.OrdinalIgnoreCase);
+            _subtitleFormatTypes = GetSubtitleFormatTypes();
         }
 
         /// <inheritdoc />
@@ -38,13 +35,14 @@ namespace MediaBrowser.MediaEncoding.Subtitles
             var subtitle = new Subtitle();
             var lines = stream.ReadAllLines().ToList();
 
-            if (!_subtitleFormats.TryGetValue(fileExtension, out var subtitleFormats))
+            if (!_subtitleFormatTypes.TryGetValue(fileExtension, out var subtitleFormatTypesForExtension))
             {
                 throw new ArgumentException($"Unsupported file extension: {fileExtension}", nameof(fileExtension));
             }
 
-            foreach (var subtitleFormat in subtitleFormats)
+            foreach (var subtitleFormatType in subtitleFormatTypesForExtension)
             {
+                var subtitleFormat = (SubtitleFormat)Activator.CreateInstance(subtitleFormatType, true)!;
                 _logger.LogDebug(
                     "Trying to parse '{FileExtension}' subtitle using the {SubtitleFormatParser} format parser",
                     fileExtension,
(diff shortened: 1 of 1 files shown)


RenV123:
I've tested this using the  reproduction steps (with media containing 2 external srt subtitle files) and using chromecast on a wide range of content (that had problems in the past) and so far I haven't found any issues with subtitles. 
It all looks good to me!

gnattu (approved):

gnattu:
Probably targeting 10.10 and backport to master would be better

BarbUk:
I confirm that it solves the issues with several media that had the problem before, mainly mkv with several sub tracks.

alltilla:
@gnattu Changed the base branch to release-10.10.z.

Shadowghost (approved):

PR #13393, opened 2025-01-19
(feat) Add external JWT auth to Jellyfin

wrmedford:
**Changes**

This adds the ability to use external JWT auth in Jellyfin. This allows Jellyfin to be used in conjunction with Identity Aware Proxies, like [Pomerium](https://github.com/pomerium/pomerium). 

**Issues**

None found.

### DIFF
--- a/Jellyfin.Api/Auth/CustomAuthenticationHandler.cs
+++ b/Jellyfin.Api/Auth/CustomAuthenticationHandler.cs
@@ -1,7 +1,9 @@
+using System;
 using System.Globalization;
 using System.Security.Claims;
 using System.Text.Encodings.Web;
 using System.Threading.Tasks;
+using Jellyfin.Api.Auth.JwtAuth;
 using Jellyfin.Api.Constants;
 using Jellyfin.Data.Enums;
 using MediaBrowser.Controller.Authentication;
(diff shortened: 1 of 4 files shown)


Mohamed1242012:
This will be useful.

sjorge:
How would this work for the existing clients? I'm I missing something obvious? Or would the only way for them to work be quick connect and a browser somewhere else logged to generate it?

nielsvanvelzen:
I don't think we will accept this change. None of our official clients will use it and we've already reserved the bearer scheme for future usage when we redo authorization in Jellyfin. Additionally, an significant change like this should first be discussed with the Jellyfin team which did not happen.

MathRdt:
Hi @nielsvanvelzen,
Do you have an estimate on the authentication refactoring in jellyfin ? I tried to look for discussions on the matter but did not find any active one on github.

I know this is a open source project and every one is doing their best !
This is just a selfish question on how to handle it on my side in the short term, i.e : using third party LDAP/OIDC plugin or waiting for native solution.

nielsvanvelzen:
> Do you have an estimate on the authentication/authorization refactoring in jellyfin ? I tried to look for discussions on the matter but did not find any active one on github.

There is no active development related to renewed authorization. The only thing related is the deprecation of old authorization methods (see everything marked as deprecated in [this document](https://gist.github.com/nielsvanvelzen/ea047d9028f676185832e51ffaf12a6f))

PR #13405, opened 2025-01-22
Remove useless checks and dead code

Bond-009:
Mostly old Emby stuff

Enables adaptive bitrate streaming, this is a separate commit so we can easily revert if it causes any issues.

### DIFF
--- a/Jellyfin.Api/Helpers/DynamicHlsHelper.cs
+++ b/Jellyfin.Api/Helpers/DynamicHlsHelper.cs
@@ -267,7 +267,7 @@ public class DynamicHlsHelper
 
         if (EnableAdaptiveBitrateStreaming(state, isLiveStream, enableAdaptiveBitrateStreaming, _httpContextAccessor.HttpContext.GetNormalizedRemoteIP()))
         {
-            var requestedVideoBitrate = state.VideoRequest is null ? 0 : state.VideoRequest.VideoBitRate ?? 0;
+            var requestedVideoBitrate = state.VideoRequest?.VideoBitRate ?? 0;
 
             // By default, vary by just 200k
             var variation = GetBitrateVariation(totalBitrate);
@@ -526,9 +526,7 @@ public class DynamicHlsHelper
             return false;
         }
 
-        // Having problems in android
-        return false;
-        // return state.VideoRequest.VideoBitRate.HasValue;
+        return state.VideoRequest?.VideoBitRate.HasValue ?? false;
     }
 
     private void AddSubtitles(StreamState state, IEnumerable<MediaStream> subtitles, StringBuilder builder, ClaimsPrincipal user)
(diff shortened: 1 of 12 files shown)


MediaBrowser.Providers/MediaInfo/FFProbeVideoInfo.cs
```diff
@@ -354,7 +354,7 @@ private void FetchBdInfo(Video video, ref ChapterInfo[] chapters, List<MediaStre
                 blurayVideoStream.Codec = ffmpegVideoStream.Codec;
                 blurayVideoStream.BitRate = blurayVideoStream.BitRate.GetValueOrDefault() == 0 ? ffmpegVideoStream.BitRate : blurayVideoStream.BitRate;
                 blurayVideoStream.Width = blurayVideoStream.Width.GetValueOrDefault() == 0 ? ffmpegVideoStream.Width : blurayVideoStream.Width;
-                blurayVideoStream.Height = blurayVideoStream.Height.GetValueOrDefault() == 0 ? ffmpegVideoStream.Width : blurayVideoStream.Height;
```
gnattu, line 357: This breaks current code and DB isn't it? Now it is designed to have wrong value stored

Jellyfin.Api/Helpers/DynamicHlsHelper.cs
```diff
@@ -526,9 +526,7 @@ private bool EnableAdaptiveBitrateStreaming(StreamState state, bool isLiveStream
             return false;
         }
 
-        // Having problems in android
-        return false;
```
gnattu, line 530: Don't just make such magic decisions. Add a server option and default to false, mark as experimental instead. I looked at its implementation and I can say that this implementation is guaranteed not to work well.

MediaBrowser.Providers/MediaInfo/FFProbeVideoInfo.cs
Bond-009: Not sure what you're asking here. Yes in some rare cases the wrong video height could have been saved to the DB. Personally I haven't seen any such instances. With this change a user can simply rescan any media where the wrong resolution was saved to fix them.

Jellyfin.Api/Helpers/DynamicHlsHelper.cs
Bond-009: > Don't just make such magic decisions.

One could see this as a bugfix. `EnableAdaptiveBitrateStreaming` is a parameter with the request that defaults to `true`, currently we don't honor that option.

>  Add a server option and default to false, mark as experimental instead.

I'm not adding an option for every little thing, an alternative solution would be defaulting to false for the request parameter which makes more sense to me.

> I looked at its implementation and I can say that this implementation is guaranteed not to work well. Even if we can fix this implementation in the future we have to still provide a way disable it in case some clients really do not play well with adaptive bitrate.

Again, the client can set the parameter in the request. If this doesn't play nice with a client the client can disable it.

gnattu: > I'm not adding an option for every little thing, an alternative solution would be defaulting to false for the request parameter which makes more sense to me.

That makes sense. I'm all for change it to default to false so that it does not immediately break current clients.  Clients have the responsibility to validate the compatibility so enable this on demand.

gnattu (approved):

crobibero (approved):

Issue #12593, opened 2024-09-05
loop in "Saving image": Generating a "movie file" from a movie file


rienafairefr:
### This issue respects the following points:

- [X] This is a **bug**, not a question or a configuration issue; Please visit our forum or chat rooms first to troubleshoot with volunteers, before creating a report. The links can be found [here](https://jellyfin.org/contact/).
- [X] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [X] I'm using an up to date version of Jellyfin Server stable, unstable or master; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.
- [X] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [X] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Description of the bug

I have no idea how it got into this:
Suddenly a movie started to 'replicate itself' in the library. 
![image](https://github.com/user-attachments/assets/c572785e-5f4e-4c1e-adf0-af415c32b9fc)
Looking at the logs it seems that it's in a loop: Jellyfin find the movie file, generates an image in the folder for it , thinks that image is a movie file, generates an image in the folder for it, etc. etc.
I don't know why the landscape image is generated with a .bin extension, and this is probably why (a .bin file is considered a CD/DVD movie). 
The loop is not really deterministic (maybe there is a tempo to it through the times the library is synchronized, manually synchronized or when other files are found or whatever.

### Reproduction steps

There is probably something in the original video file that tells Jellyfin to save an image with .bin, maybe the codec/container or something else

### What is the current _bug_ behavior?

a movie 'replicates itself' in the library. 

### What is the expected _correct_ behavior?

no replication

### Jellyfin Server version

Older*

### Specify version number

10.9.7

### Specify the build version

10.9.7

### Environment

```markdown
- OS: Linux
- Linux Kernel: 6.1.0
- Virtualization: Docker (official image)
```

### Jellyfin logs

```shell
2024-09-03T19:50:18.981309313Z [21:50:18] [DBG] [40] MediaBrowser.Providers.Manager.ProviderManager: Saving image to /movies/Pan's Labyrinth (2006)/landscape.bin
[...]

PR #13439, opened 2025-01-26
Fall back to calculating mime type from path when needed

Bond-009:
Should fix #12593 and #7816

PR #13441, opened 2025-01-26
Fallback to URL extension for image type when content header is invalid

theguymadmax:
**Changes**

This PR modifies the `DownloadImage` method in` ItemImageProvider` to handle cases where the image's content header does not indicate a valid image type. In these cases, the extension of the image URL is used to infer the content type. If the content type still cannot be determined by either the content header or the URL extension, it defaults to `image/jpeg`. This ensures that images are correctly downloaded with the appropriate extension and prevents them from being mistakenly downloaded as binary files.

**Issues**
#12593


gnattu:
duplicate of #13439

PR #13443, opened 2025-01-26
chore(ci): Add permissions grant

jsoref:
**Changes**
Add permissions grant to ci-codeql-analysis.yml

**Issues**

Without this fix, repositories/organizations that are configured to with restrictive permissions (and all repositories/orgs should be) fail like this:
https://github.com/jsoref/jellyfin/actions/runs/12976527234

### DIFF
--- a/.github/workflows/ci-codeql-analysis.yml
+++ b/.github/workflows/ci-codeql-analysis.yml
@@ -8,6 +8,10 @@ on:
   schedule:
     - cron: '24 2 * * 4'
 
+permissions:
+  contents: read
+  security-events: write
+
 jobs:
   analyze:
     name: Analyze


gnattu:
The CI is not designed to be run in forked repositories other than making pull requests iirc, which means this is not needed.

jsoref:
1. Permissions should really be assigned to specify least privilege and your organization/repository should be using restrictive permissions regardless of whether you allow them to run in forks.
2. If you don't want it running in forks, I'm happy to make a PR to add an `if` to restrict it to the main repository (many of the other workflows have such a restriction).

Bond-009 (approved):

nielsvanvelzen (approved):

PR #13444, opened 2025-01-26
Fix spelling

jsoref:
**Changes**
Fixes misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

The misspellings have been reported at https://github.com/jsoref/jellyfin/actions/runs/12979401882#summary-36195047679

The action will report that the changes in this PR would make it happy: https://github.com/jsoref/jellyfin/actions/runs/12979402012#summary-36195047936 -- Note that there are a number of tests that appear to be consistently failing. In order to get CI not not fail miserably, I included 0c236225278edacc5f9c93467d628ced8a960660 in my test PR (but they aren't included in this PR).

---

As before, if people are happy with the changes, I'll squash them.

There's one change that I omitted which was `BluRay` to `Bluray` (for everything except file names) because that upset the API checker. Ignoring files with `BluRay` in their names, there's an exactly equal number of `BluRay` and `Bluray`, I looked around and checked with people and do believe that the latter would be more correct (although it'd obviously require adding a new set of APIs and deprecating the old ones). If people are interested, I'm happy to make a PR for that separately.

### DIFF
--- a/Emby.Naming/TV/SeriesResolver.cs
+++ b/Emby.Naming/TV/SeriesResolver.cs
@@ -12,7 +12,7 @@ namespace Emby.Naming.TV
         /// <summary>
         /// Regex that matches strings of at least 2 characters separated by a dot or underscore.
         /// Used for removing separators between words, i.e turns "The_show" into "The show" while
-        /// preserving namings like "S.H.O.W".
+        /// preserving names like "S.H.O.W".
         /// </summary>
         [GeneratedRegex(@"((?<a>[^\._]{2,})[\._]*)|([\._](?<b>[^\._]{2,}))")]
         private static partial Regex SeriesNameRegex();
--- a/Jellyfin.Api/Helpers/StreamingHelpers.cs
+++ b/Jellyfin.Api/Helpers/StreamingHelpers.cs
@@ -210,7 +210,7 @@ public static class StreamingHelpers
                     && state.VideoRequest.VideoBitRate.Value >= state.VideoStream.BitRate.Value)
                 {
                     // Don't downscale the resolution if the width/height/MaxWidth/MaxHeight is not requested,
-                    // and the requested video bitrate is higher than source video bitrate.
+                    // and the requested video bitrate is greater than source video bitrate.
                     if (state.VideoStream.Width.HasValue || state.VideoStream.Height.HasValue)
                     {
                         state.VideoRequest.MaxWidth = state.VideoStream?.Width;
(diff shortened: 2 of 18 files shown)


Emby.Naming/TV/SeriesResolver.cs
```diff
@@ -12,7 +12,7 @@ public static partial class SeriesResolver
         /// <summary>
         /// Regex that matches strings of at least 2 characters separated by a dot or underscore.
         /// Used for removing separators between words, i.e turns "The_show" into "The show" while
-        /// preserving namings like "S.H.O.W".
+        /// preserving names like "S.H.O.W".
```
jsoref, line 15: I think I skipped over this in my prior PR because I was being conservative. This time around, I'm more confident that this change is correct.

Jellyfin.Api/Helpers/StreamingHelpers.cs
```diff
@@ -210,7 +210,7 @@ public static async Task<StreamState> GetStreamingState(
                     && state.VideoRequest.VideoBitRate.Value >= state.VideoStream.BitRate.Value)
                 {
                     // Don't downscale the resolution if the width/height/MaxWidth/MaxHeight is not requested,
-                    // and the requested video bitrate is higher than source video bitrate.
+                    // and the requested video bitrate is greater than source video bitrate.
```
jsoref, line 213: This is a comparison between two actual quantities, so `greater than` is appropriate.

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -289,7 +289,7 @@ private IQueryable<BaseItemEntity> ApplyGroupingFilter(IQueryable<BaseItemEntity
         return dbQuery;
     }
 
-    private IQueryable<BaseItemEntity> ApplyQueryPageing(IQueryable<BaseItemEntity> dbQuery, InternalItemsQuery filter)
+    private IQueryable<BaseItemEntity> ApplyQueryPaging(IQueryable<BaseItemEntity> dbQuery, InternalItemsQuery filter)
```
jsoref, line 292: As this is private, it seems reasonable.

MediaBrowser.Controller/Session/SessionInfo.cs
```diff
@@ -286,7 +286,7 @@ public bool SupportsRemoteControl
         /// <summary>
         /// Gets or sets the playlist item id.
         /// </summary>
-        /// <value>The splaylist item id.</value>
+        /// <value>The playlist item id.</value>
```
jsoref, line 289: It's possible it should be a `session playlist`, but...

jsoref:
Um, the check in https://github.com/jellyfin/jellyfin/pull/13444#issuecomment-2614639392 is buggy.
It should mention `anamporphic` from 863e321e10e4ef4ef763591766c3c18f333387a8 but doesn't.

Shadowghost (approved):

Bond-009 (approved):

PR #13451, opened 2025-01-27
Add Postgres SQL support

JPVenson:
Refactors code architecture for database usage.
Adds support for postgres SQL for new installs.

Changes:
- Moved database related code to own Project
- Refactored User entity to not use collation for username for compataibility reasons
- Refactored user and role entities to no longer implement buissniss logic
- Added SqLite adapter
- Added pgsql adapter
- Added Pgsql ready devcontainer
- Added optional database.xml configuration file
- Added document for explaining migration path
-


Jellyfin.Server.Implementations/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -11,17 +17,59 @@ namespace Jellyfin.Server.Implementations.Extensions;
...
+        var efCoreConfiguration = configurationManager.GetConfiguration<DatabaseConfigurationOptions>("database");
+        var providers = GetSupportedDbProviders();
+        JellyfinDbProviderFactory? providerFactory = null;
+
+        if (efCoreConfiguration is null)
+        {
+            // when nothing is setup via new Database configuration, fallback to SqLite with default settings.
+            efCoreConfiguration = new DatabaseConfigurationOptions()
+            {
+                DatabaseType = "SqLite",
```
EraYaN: Small nit: SQLite is the right capitalization but we might want to make those strings all lowercase anyway.

EraYaN:
And in addition, maybe just write PgSql as PostgreSQL everywhere to match SQLite so that we use correct names as seen on their websites.

tests/Jellyfin.Server.Implementations.Tests/EfMigrations/EfMigrationTests.cs
```diff
@@ -9,10 +10,18 @@ namespace Jellyfin.Server.Implementations.Tests.EfMigrations;
...
 {
     [Fact]
-    public void CheckForUnappliedMigrations()
+    public void CheckForUnappliedMigrations_PgSQL()
     {
-        var dbDesignContext = new DesignTimeJellyfinDbFactory();
+        var dbDesignContext = new PgSqlDesignTimeJellyfinDbFactory();
         var context = dbDesignContext.CreateDbContext([]);
-        Assert.False(context.Database.HasPendingModelChanges(), "There are unapplied changes to the EfCore model. Please create a Migration.");
+        Assert.False(context.Database.HasPendingModelChanges(), "There are unapplied changes to the EfCore model for PgSQL. Please create a Migration.");
```
EraYaN, line 17: Same here with EfCore -> EFCore.

Jellyfin.Database/Jellyfin.Database.Providers.PgSql/PgSqlDatabaseProvider.cs
```diff
@@ -0,0 +1,75 @@
...
+    {
+        var dbSettings = _configurationManager.GetConfiguration<DatabaseConfigurationOptions>("database");
+
+        if (dbSettings.PostgreSql is null)
+        {
+            throw new InvalidOperationException("Selected PgSQL as database provider but did not provide required configuration. Please see docs.");
+        }
+
+        var connectionBuilder = new NpgsqlConnectionStringBuilder();
+        connectionBuilder.ApplicationName = "jellyfin";
```
EraYaN, line 45: Maybe include the version in the application name.

Jellyfin.Server.Implementations/Extensions/ServiceCollectionExtensions.cs
JPVenson: i opted to make the lookup case insensitive instead

EraYaN (commented):
Postgres is still abbreviated in some spots (namespaces especially) but not sure what we want to do there. EFCore always writes it in full, but this code base is a bit of a mix for writing things out and using abbreviations.

JPVenson:
Yea pgsql should be really enough i dont want to plaster it all over the place, it adds 0 benefit. Its a widely known term, esp in that context so i would say its fine?

EraYaN:
It's mostly that I just think it's ugly, they use pg, psql or PostgreSQL (sometimes even just Postgres) themselves. We could also name the providers after the actual implementation (as in `Npgsql` if someone wants to add some alternative for example), at least in the Implementations namespace.

EraYaN:
When this gets merged: https://github.com/jellyfin/jellyfin/pull/13492 this need to not apply for Postgres.

JPVenson:
> When this gets merged: https://github.com/jellyfin/jellyfin/pull/13492 this need to not apply for Postgres.

Yes, depending on what gets merged first either i will do it here or do it in the other branch.

Implementation wise, its pretty straight forward. Just do it the same way as the ModelCreation hook.

Jellyfin.Api/Controllers/UserController.cs
```diff
@@ -389,7 +390,7 @@ public async Task<ActionResult> UpdateUser(
             return StatusCode(StatusCodes.Status403Forbidden, "User update not allowed.");
         }
 
-        if (!string.Equals(user.Username, updateUser.Name, StringComparison.Ordinal))
+        if (!string.Equals(user.Username, updateUser.Name, StringComparison.OrdinalIgnoreCase))
```
crobibero, line 393: This would prevent users from changing the case of their username

crobibero (commented):
I would really prefer this PR be split into setting up the codebase to allow alternate db providers, and then a follow up to actually add the alternate provider

JPVenson:
FYI
This has __NOT__ been merged actually but only the ground work PR. as they share the same git history GH assumes this has also been merged which is not correct.

Postgres Support will not hit 10.11 but maybe later versions.

JPVenson:
@PrivatePuffin nothing relating to pgsql support within jellyfin has been merged so far. Right now it only exists as a plugin:
https://github.com/JPVenson/Jellyfin.Pgsql

JPVenson:
this PR was not merged. Github thinks it was because the https://github.com/jellyfin/jellyfin/pull/13589 PR was branched from this one. Ignore anything that is in this PR look only at 13589

0xFEEDC0DE64:
oh come on guys can we please get this started again?

PR #13459, opened 2025-01-29
Fixed Websocket not locking state correctly

### LINKED ISSUE #13457: Server crash when handling a closed WebSocket connection

### Description of the bug

When closing a browser window with Jellyfin open, the server unexpectedly crashed. I was logged in locally with an admin account, but I had taken no action aside from checking the dashboard. I had enabled Trickplay images on my library for the first time, and was checking the status of the (very long-running) task, and it was nearly complete (~98%). I later discovered that there was an error logged at the time I closed the browser window which crashed the server completely, and it needed to be restarted manually. No errors were logged after restarting.

I've checked for comparable issue reports, and didn't find much. This appears distinct from #12090 , as the message is different and there was no other issue logged related to the client disconnecting. However, it could be running in the same location which is missing exception handling, allowing for the server to crash.

### Reproduction steps

See description - however, this has only happened one time while taking the same actions. Seems intermittent - but given the severity, I thought it was best to report it.

An error is logged to the Jellyfin server log file, then the server crashes with the same error message saved to the event log. 

An unhandled exception related to a client disconnecting should not take down the entire server.

### Jellyfin Server version

10.10.0+

### Specify commit id

_No response_

_No response_

### Specify version number

10.10.5

### Specify the build version

10.10.5

### Environment

```markdown
- OS: Windows Server 2025
- Linux Kernel: none
- Virtualization: none
- Clients: Browser
- Browser: Firefox 134
- FFmpeg Version: 7.0.2-Jellyfin
- Playback Method: N/A
- GPU Model: RTX 3070
- Plugins: none
- Reverse Proxy: Caddy
- Base URL: none
- Networking: host
- Storage: local
```

### Jellyfin logs

```shell
(IP address has been redacted

[2025-01-28 23:54:17.699 -04:00] [INF] [89] Emby.Server.Implementations.HttpServer.WebSocketManager: WS "[client local IP]" closed
[2025-01-28 23:54:17.701 -04:00] [INF] [125] Emby.Server.Implementations.HttpServer.WebSocketManager: WS "[client local IP]" closed
[2025-01-28 23:54:17.707 -04:00] [FTL] [123] Main: Unhandled Exception
System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
   at System.Collections.Generic.List`1.Remove(T item)
   at Emby.Server.Implementations.Session.WebSocketController.OnConnectionClosed(Object sender, EventArgs e)
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_1(Object state)
   at System.Threading.QueueUserWorkItemCallback.Execute()
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
   at System.Threading.PortableThreadPool.WorkerThread.WorkerThreadStart()
[2025-01-28 23:54:17.713 -04:00] [INF] [121] Emby.Server.Implementations.HttpServer.WebSocketManager: WS "[client local IP]" closed
```

### FFmpeg logs

```shell

```

### Client / Browser logs

The following logs are saved to the Windows event log, about one second apart and starting on the same second as the Jellyfin server logs:

Error - .NET Runtime
```
Application: jellyfin.exe
CoreCLR Version: 8.0.1224.60305
.NET Version: 8.0.12
Description: The process was terminated due to an unhandled exception.
Exception Info: System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
   at System.Collections.Generic.List`1.Remove(T item)
   at Emby.Server.Implementations.Session.WebSocketController.OnConnectionClosed(Object sender, EventArgs e)
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_1(Object state)
   at System.Threading.QueueUserWorkItemCallback.Execute()
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
   at System.Threading.PortableThreadPool.WorkerThread.WorkerThreadStart()
```

Error - Application Error
```
Faulting application name: jellyfin.exe, version: 10.10.5.0, time stamp: 0x674f0000
Faulting module name: KERNELBASE.dll, version: 10.0.26100.2454, time stamp: 0x398a1cce
Exception code: 0xe0434352
Fault offset: 0x00000000000c837a
Faulting process id: 0x1924
Faulting application start time: 0x1DB703A8920118C
Faulting application path: C:\Program Files\Jellyfin\Server\jellyfin.exe
Faulting module path: C:\WINDOWS\System32\KERNELBASE.dll
Report Id: a046d559-96b5-40bc-be20-22a87dc037cd
Faulting package full name: 
Faulting package-relative application ID: 
```

_No response_

### Additional information

Trickplay image generation was running, though no errors had been logged pertaining to that. It had also been running for over two days, as it was enabled for the first time recently. Hardware decoding was enabled for image generation.

JPVenson:
Ensures websockets collection are properly locked and disposed

**Issues**
Fixes #13457

### DIFF
--- a/Emby.Server.Implementations/Session/WebSocketController.cs
+++ b/Emby.Server.Implementations/Session/WebSocketController.cs
@@ -21,6 +21,7 @@ namespace Emby.Server.Implementations.Session
         private readonly SessionInfo _session;
 
         private readonly List<IWebSocketConnection> _sockets;
+        private readonly ReaderWriterLockSlim _socketsLock;
         private bool _disposed = false;
 
         public WebSocketController(
@@ -31,10 +32,26 @@ namespace Emby.Server.Implementations.Session
             _logger = logger;
             _session = session;
             _sessionManager = sessionManager;
-            _sockets = new List<IWebSocketConnection>();
+            _sockets = new();
+            _socketsLock = new();
         }
 
-        private bool HasOpenSockets => GetActiveSockets().Any();
+        private bool HasOpenSockets
+        {
+            get
+            {
+                ObjectDisposedException.ThrowIf(_disposed, this);
+                try
+                {
+                    _socketsLock.EnterReadLock();
+                    return _sockets.Where(i => i.State == WebSocketState.Open).Any();
+                }
+                finally
+                {
+                    _socketsLock.ExitReadLock();
+                }
+            }
+        }
 
         /// <inheritdoc />
         public bool SupportsMediaControl => HasOpenSockets;
@@ -42,13 +59,19 @@ namespace Emby.Server.Implementations.Session
         /// <inheritdoc />
         public bool IsSessionActive => HasOpenSockets;
 
-        private IEnumerable<IWebSocketConnection> GetActiveSockets()
-            => _sockets.Where(i => i.State == WebSocketState.Open);
-
         public void AddWebSocket(IWebSocketConnection connection)
         {
             _logger.LogDebug("Adding websocket to session {Session}", _session.Id);
-            _sockets.Add(connection);
+            ObjectDisposedException.ThrowIf(_disposed, this);
+            try
+            {
+                _socketsLock.EnterWriteLock();
+                _sockets.Add(connection);
+            }
+            finally
+            {
+                _socketsLock.ExitWriteLock();
+            }
 
             connection.Closed += OnConnectionClosed;
         }
@@ -57,7 +80,17 @@ namespace Emby.Server.Implementations.Session
         {
             var connection = sender as IWebSocketConnection ?? throw new ArgumentException($"{nameof(sender)} is not of type {nameof(IWebSocketConnection)}", nameof(sender));
             _logger.LogDebug("Removing websocket from session {Session}", _session.Id);
-            _sockets.Remove(connection);
+            ObjectDisposedException.ThrowIf(_disposed, this);
+            try
+            {
+                _socketsLock.EnterWriteLock();
+                _sockets.Remove(connection);
+            }
+            finally
+            {
+                _socketsLock.ExitWriteLock();
+            }
+
             connection.Closed -= OnConnectionClosed;
             await _sessionManager.CloseIfNeededAsync(_session).ConfigureAwait(false);
         }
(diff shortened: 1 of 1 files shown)


gnattu:
Can't we just use `ConcurrentBag`?  This list does not need to be ordered if I understand its use case correctly.

JPVenson:
ConcurrentBag is not ordered and thats not needed here, thats correct. But ConcurrentBag does not allow random access and we need to remove elements out of order so its no option here.

gnattu:
I'd still prefer using a native thread safe collection instead of implementing locks unless there is no other choice for easier for maintenance as it is easier to make mistakes. I think we can just use a `ConcurrentDictionary<IWebSocketConnection, null>` to replace all the operations implemented here which also supports removal by object reference. Its `.Keys` provides thread-safe snapshots for read-only operations.

Emby.Server.Implementations/Session/WebSocketController.cs
```diff
@@ -57,7 +80,17 @@ private async void OnConnectionClosed(object? sender, EventArgs e)
...
+            {
+                _socketsLock.EnterWriteLock();
+                _sockets.Remove(connection);
+            }
+            finally
+            {
+                _socketsLock.ExitWriteLock();
+            }
+
             connection.Closed -= OnConnectionClosed;
```
gnattu, line 94: I might lock the mutations of each connections as well because the handler might be called multiple times. But this is probably fine for this handler.

JPVenson: i mean we have the scope, why not include it anway.

EraYaN:
Is there a reason for not using the new `Lock` type plus the `lock` keyword? The code base uses it in other spots. And it lowers into almost the same code while looking a lot nicer IMO. It just uses the `Scope` type internally.

JPVenson:
same reason as above.
A ReaderWriterLock is far more efficient then an blank locking.
Using a lock blocks reading and writing concurrently, a ReaderWriterLock only blocks when a Write is requested but does not block concurrent reads which is the majority operation here.

gnattu (approved):

EraYaN (approved):

Bond-009 (approved):

PR #13469, opened 2025-02-01
Fix SchedulesDirect image prefetching

Shadowghost:
**Changes**
* Only pre-cache images for new and updated programs.
* Fix `forceUpdate` calculation to always fetch images if necessary

**Issues**
* Currently we are hammering the SD API because we always re-fetch all images.

As always, I do not use any kind of LiveTV, so someone else needs to test this.

### DIFF
--- a/src/Jellyfin.LiveTv/Guide/GuideManager.cs
+++ b/src/Jellyfin.LiveTv/Guide/GuideManager.cs
@@ -3,7 +3,6 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
-using Jellyfin.Data.Entities.Libraries;
 using Jellyfin.Data.Enums;
 using Jellyfin.Extensions;
 using Jellyfin.LiveTv.Configuration;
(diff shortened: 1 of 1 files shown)


theguymadmax:
I can't speak for how it works with SD data. But now it doesn't cache images on the first import.


I only get images after refreshing guide data after an initial import.

Bond-009 (approved):

src/Jellyfin.LiveTv/Guide/GuideManager.cs
```diff
@@ -632,37 +615,35 @@ private async Task<LiveTvChannel> GetChannel(
         item.IndexNumber = info.EpisodeNumber;
         item.ParentIndexNumber = info.SeasonNumber;
 
-        forceUpdate = forceUpdate || UpdateImages(item, info);
+        forceUpdate = UpdateImages(item, info) || forceUpdate;
```
IDisposable, line 618: This is a good fix, but it is not the same fix for Issue #13496 as PR #13497.  This one fixes the issue where _all_ images are not updated.

crobibero:
```suggestion
        forceUpdate |= UpdateImages(item, info);
```

I think for clarity this should be updated to match the other pattern

crobibero (approved):

PR #13475, opened 2025-02-02
Fix transcoding if audio is unsupported

ACertainCoder:
**Changes**
When using DLNA to play a video on another device that doesn't support the audio stream of that video, Jellyfin doesn't transcode it (typical log message: `PlayMethod=DirectStream, TranscodeReason=ContainerNotSupported, AudioCodecNotSupported`). The line responsible for filtering direct stream issues *excludes* these reasons rather than making them *exclusive*.

Changing the negated bitwise-anding to a regular one fixes this problem. I could verify that the correct audio stream was then provided to my DLNA renderer and that playback via Jellyfin web was still functioning as expected (without the need for transcoding).

**Issues**
https://github.com/jellyfin/jellyfin-plugin-dlna/issues/23

Please correct me if this change would have unwanted consequences.

### DIFF
--- a/MediaBrowser.Model/Dlna/StreamBuilder.cs
+++ b/MediaBrowser.Model/Dlna/StreamBuilder.cs
@@ -1367,7 +1367,7 @@ namespace MediaBrowser.Model.Dlna
                         failureReasons |= audioCodecProfileReasons;
                     }
 
-                    var directStreamFailureReasons = failureReasons & (~DirectStreamReasons);
+                    var directStreamFailureReasons = failureReasons & ~DirectStreamReasons;
 
                     PlayMethod? playMethod = null;
                     if (failureReasons == 0 && isEligibleForDirectPlay && mediaSource.SupportsDirectPlay)
--- a/MediaBrowser.Model/Dlna/StreamBuilder.cs
+++ b/MediaBrowser.Model/Dlna/StreamBuilder.cs
@@ -1367,7 +1367,7 @@ namespace MediaBrowser.Model.Dlna
                         failureReasons |= audioCodecProfileReasons;
                     }
 
-                    var directStreamFailureReasons = failureReasons & ~DirectStreamReasons;
+                    var directStreamFailureReasons = failureReasons & DirectStreamReasons;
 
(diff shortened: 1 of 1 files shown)


gnattu (changes requested):
This is not the intended behavior. The `directStreamFailureReasons` want to get all reasons that are not `DirectStreamReasons` and if all reasons are `DirectStreamReasons`, the result would be 0, so that we know that this is indeed detected as a direct stream.

The `PlayMethod.DirectStream` is really fragile anyway because that play method had different meanings during development. Now the `DirectStream` is treated more as a special case of transcoding instead of having its own play method, and the problem that you are having is probably due to some other errors in the pipeline that does not transcode for directstream, so where you are changing is not the core problem.

ACertainCoder:
Ok, thank you for clarifying that.
I'll take another look at this in the following days and see what else could be affecting this.

gnattu:
Please don't. The `StreamBuilder` is more nasty than you could ever think. The problem you are having can be workaround by setting `EnableDirectStream = false` on all `MediaOptions` instances it passed to the server in the DLNA plugin. For the server itself, that option is disabled for years:
https://github.com/jellyfin/jellyfin/blob/ce64dbc034b2176d863fb598cf7079aa6a1fce55/Jellyfin.Api/Helpers/MediaInfoHelper.cs#L243

The `PlayMethod.DirectStream` only has one correct usage: that is the pure audio(for music) pipeline used by our new API `UniversalAudioController`, where that play method would mean copy stream into another container and the streambuilder is doing jobs based on that play method. However, for video and/or video with audio usage, I believe `PlayMethod.DirectStream` should be banned. This is because all code related to this method is likely to malfunction, as `PlayMethod.Transcode` handles all non-direct playback, including Remux (full stream copy) and DirectStream (copying video with audio transcoding) for server main logic.

gnattu:
Closing as this PR's change is not desirable and the fix for original issue should be implemented by alternative approaches.

ACertainCoder:
Alright, thanks for your further explanation. I'll leave the `StreamBuilder` alone and check if the DLNA plugin can handle things differently based on your feedback.

PR #13486, opened 2025-02-03
Add Waveform bar api endpoint

FrenchGithubUser:
**Changes**
Added an API endpoint Audio/itemId/audiowaveform to generate a .dat file to create audiowaveforms. Following this library's standard : https://github.com/bbc/audiowaveform

This is my first PR and first experience with C#. I tried my best at following guidelines and existing code. Please tell me anything I should change, I don't expect this PR from passing.

### DIFF
--- a/Jellyfin.Api/Controllers/AudioController.cs
+++ b/Jellyfin.Api/Controllers/AudioController.cs
@@ -92,18 +92,18 @@ public class AudioController : BaseJellyfinApiController
     [ProducesAudioFile]
     public async Task<ActionResult> GetAudioStream(
         [FromRoute, Required] Guid itemId,
-        [FromQuery] [RegularExpression(EncodingHelper.ValidationRegex)] string? container,
+        [FromQuery][RegularExpression(EncodingHelper.ValidationRegex)] string? container,
         [FromQuery] bool? @static,
         [FromQuery] string? @params,
         [FromQuery] string? tag,
         [FromQuery, ParameterObsolete] string? deviceProfileId,
         [FromQuery] string? playSessionId,
-        [FromQuery] [RegularExpression(EncodingHelper.ValidationRegex)] string? segmentContainer,
+        [FromQuery][RegularExpression(EncodingHelper.ValidationRegex)] string? segmentContainer,
         [FromQuery] int? segmentLength,
         [FromQuery] int? minSegments,
         [FromQuery] string? mediaSourceId,
         [FromQuery] string? deviceId,
-        [FromQuery] [RegularExpression(EncodingHelper.ValidationRegex)] string? audioCodec,
+        [FromQuery][RegularExpression(EncodingHelper.ValidationRegex)] string? audioCodec,
         [FromQuery] bool? enableAutoStreamCopy,
         [FromQuery] bool? allowVideoStreamCopy,
         [FromQuery] bool? allowAudioStreamCopy,
@@ -133,8 +133,8 @@ public class AudioController : BaseJellyfinApiController
         [FromQuery] int? cpuCoreLimit,
         [FromQuery] string? liveStreamId,
         [FromQuery] bool? enableMpegtsM2TsMode,
-        [FromQuery] [RegularExpression(EncodingHelper.ValidationRegex)] string? videoCodec,
-        [FromQuery] [RegularExpression(EncodingHelper.ValidationRegex)] string? subtitleCodec,
+        [FromQuery][RegularExpression(EncodingHelper.ValidationRegex)] string? videoCodec,
+        [FromQuery][RegularExpression(EncodingHelper.ValidationRegex)] string? subtitleCodec,
         [FromQuery] string? transcodeReasons,
         [FromQuery] int? audioStreamIndex,
         [FromQuery] int? videoStreamIndex,
@@ -265,12 +265,12 @@ public class AudioController : BaseJellyfinApiController
         [FromQuery] string? tag,
         [FromQuery, ParameterObsolete] string? deviceProfileId,
         [FromQuery] string? playSessionId,
-        [FromQuery] [RegularExpression(EncodingHelper.ValidationRegex)] string? segmentContainer,
+        [FromQuery][RegularExpression(EncodingHelper.ValidationRegex)] string? segmentContainer,
         [FromQuery] int? segmentLength,
         [FromQuery] int? minSegments,
         [FromQuery] string? mediaSourceId,
         [FromQuery] string? deviceId,
-        [FromQuery] [RegularExpression(EncodingHelper.ValidationRegex)] string? audioCodec,
+        [FromQuery][RegularExpression(EncodingHelper.ValidationRegex)] string? audioCodec,
         [FromQuery] bool? enableAutoStreamCopy,
         [FromQuery] bool? allowVideoStreamCopy,
         [FromQuery] bool? allowAudioStreamCopy,
@@ -300,8 +300,8 @@ public class AudioController : BaseJellyfinApiController
         [FromQuery] int? cpuCoreLimit,
         [FromQuery] string? liveStreamId,
         [FromQuery] bool? enableMpegtsM2TsMode,
-        [FromQuery] [RegularExpression(EncodingHelper.ValidationRegex)] string? videoCodec,
-        [FromQuery] [RegularExpression(EncodingHelper.ValidationRegex)] string? subtitleCodec,
+        [FromQuery][RegularExpression(EncodingHelper.ValidationRegex)] string? videoCodec,
+        [FromQuery][RegularExpression(EncodingHelper.ValidationRegex)] string? subtitleCodec,
         [FromQuery] string? transcodeReasons,
         [FromQuery] int? audioStreamIndex,
         [FromQuery] int? videoStreamIndex,
@@ -363,5 +363,19 @@ public class AudioController : BaseJellyfinApiController
         };
 
         return await _audioHelper.GetAudioStream(_transcodingJobType, streamingRequest).ConfigureAwait(false);
+    }
+
+    /// <summary>
+    /// Gets audio data to draw a waveform bar.
+    /// </summary>
+    /// <param name="itemId">The item id.</param>
+    /// <response code="200">AudioWaveForm returned.</response>
+    /// <returns>A <see cref="FileResult"/> containing the audiowaveform file.</returns>
+    [HttpGet("{itemId}/audiowaveform")]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    public async Task GetAudioWaveForm([FromRoute, Required] Guid itemId)
+    {
+        await _audioHelper.GetAudioWaveForm(itemId).ConfigureAwait(false);
+        // return string.Empty;
     }
 }
--- a/Jellyfin.Api/Controllers/AudioController.cs
+++ b/Jellyfin.Api/Controllers/AudioController.cs
@@ -7,6 +7,7 @@ using Jellyfin.Api.Helpers;
 using Jellyfin.Api.Models.StreamingDtos;
 using MediaBrowser.Controller.MediaEncoding;
 using MediaBrowser.Controller.Streaming;
+using MediaBrowser.Model;
 using MediaBrowser.Model.Dlna;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
@@ -373,9 +374,10 @@ public class AudioController : BaseJellyfinApiController
     /// <returns>A <see cref="FileResult"/> containing the audiowaveform file.</returns>
     [HttpGet("{itemId}/audiowaveform")]
     [ProducesResponseType(StatusCodes.Status200OK)]
-    public async Task GetAudioWaveForm([FromRoute, Required] Guid itemId)
+    public async Task<ActionResult> GetAudioWaveForm([FromRoute, Required] Guid itemId)
     {
-        await _audioHelper.GetAudioWaveForm(itemId).ConfigureAwait(false);
-        // return string.Empty;
+        var fileStream = await _audioHelper.GetAudioWaveForm(itemId).ConfigureAwait(false);
+
+        return File(fileStream, MimeTypes.GetMimeType("file.dat"));
     }
 }
--- a/Jellyfin.Api/Helpers/AudioHelper.cs
+++ b/Jellyfin.Api/Helpers/AudioHelper.cs
@@ -1,11 +1,16 @@
+using System;
+using System.Diagnostics;
 using System.IO;
+using System.Linq;
 using System.Net.Http;
+using System.Text.Json;
 using System.Threading;
 using System.Threading.Tasks;
 using MediaBrowser.Common.Configuration;
 using MediaBrowser.Common.Extensions;
 using MediaBrowser.Common.Net;
 using MediaBrowser.Controller.Configuration;
+using MediaBrowser.Controller.Entities;
 using MediaBrowser.Controller.Library;
 using MediaBrowser.Controller.MediaEncoding;
 using MediaBrowser.Controller.Streaming;
@@ -153,5 +158,48 @@ public class AudioHelper
             ffmpegCommandLineArguments,
             transcodingJobType,
             cancellationTokenSource).ConfigureAwait(false);
+    }
+
+    /// <summary>
+    /// Gets audio data to draw a waveform bar.
+    /// </summary>
+    /// <param name="itemId">The item id.</param>
+    /// <returns>A <see cref="BaseItem"/> containing the resulting <see cref="ActionResult"/>.</returns>
+    public async Task GetAudioWaveForm(Guid itemId)
+    {
+        var item = _libraryManager.GetItemById<BaseItem>(itemId)
+        ?? throw new ResourceNotFoundException();
+
+        var libraryOptions = BaseItem.LibraryManager.GetLibraryOptions(item);
+        // considering waveform data as metadata
+        var saveInMediaFolder = libraryOptions.SaveLocalMetadata;
+        var saveFileName = Path.GetFileNameWithoutExtension(item.Path) + "." + "pcm";
+        string outputPath;
+        if (saveInMediaFolder)
+        {
+            outputPath = Path.GetFullPath(Path.Combine(item.ContainingFolderPath, saveFileName));
+        }
+        else
+        {
+            outputPath = Path.GetFullPath(Path.Combine(item.GetInternalMetadataPath(), saveFileName));
+        }
+
+        Console.WriteLine(item.Path);
+        Console.WriteLine(outputPath);
+
+        if (File.Exists(outputPath))
+        {
+        }
+        else
+        {
+            await Task.Run(() => Process.Start("ffmpeg", $"-y -i \"{item.Path}\" -acodec pcm_s16le -f s16le -ac 1 -ar 1000 \"{outputPath}\"").WaitForExit()).ConfigureAwait(false);
+
+            byte[] pcmFile = await File.ReadAllBytesAsync(outputPath).ConfigureAwait(false);
+
+            short[] pcmData = Enumerable.Range(0, pcmFile.Length / 2)
+                .Select(i => BitConverter.ToInt16(pcmFile, i * 2)).ToArray();
+        }
+
+        // return item;
     }
 }
--- a/Jellyfin.Api/Helpers/AudioHelper.cs
+++ b/Jellyfin.Api/Helpers/AudioHelper.cs
@@ -164,42 +164,26 @@ public class AudioHelper
     /// Gets audio data to draw a waveform bar.
     /// </summary>
     /// <param name="itemId">The item id.</param>
-    /// <returns>A <see cref="BaseItem"/> containing the resulting <see cref="ActionResult"/>.</returns>
-    public async Task GetAudioWaveForm(Guid itemId)
+    /// <returns>A <see cref="FileStream"/> containing the resulting .dat file.</returns>
+    public async Task<FileStream> GetAudioWaveForm(Guid itemId)
     {
         var item = _libraryManager.GetItemById<BaseItem>(itemId)
-        ?? throw new ResourceNotFoundException();
+            ?? throw new ResourceNotFoundException();
 
         var libraryOptions = BaseItem.LibraryManager.GetLibraryOptions(item);
-        // considering waveform data as metadata
         var saveInMediaFolder = libraryOptions.SaveLocalMetadata;
-        var saveFileName = Path.GetFileNameWithoutExtension(item.Path) + "." + "pcm";
-        string outputPath;
-        if (saveInMediaFolder)
-        {
-            outputPath = Path.GetFullPath(Path.Combine(item.ContainingFolderPath, saveFileName));
-        }
-        else
-        {
-            outputPath = Path.GetFullPath(Path.Combine(item.GetInternalMetadataPath(), saveFileName));
-        }
+        var saveFileName = Path.GetFileNameWithoutExtension(item.Path) + ".dat";
+        string outputPath = saveInMediaFolder
+            ? Path.Combine(item.ContainingFolderPath, saveFileName)
+            : Path.Combine(item.GetInternalMetadataPath(), saveFileName);
 
-        Console.WriteLine(item.Path);
-        Console.WriteLine(outputPath);
-
-        if (File.Exists(outputPath))
+        if (!File.Exists(outputPath))
         {
-        }
-        else
-        {
-            await Task.Run(() => Process.Start("ffmpeg", $"-y -i \"{item.Path}\" -acodec pcm_s16le -f s16le -ac 1 -ar 1000 \"{outputPath}\"").WaitForExit()).ConfigureAwait(false);
-
-            byte[] pcmFile = await File.ReadAllBytesAsync(outputPath).ConfigureAwait(false);
-
-            short[] pcmData = Enumerable.Range(0, pcmFile.Length / 2)
-                .Select(i => BitConverter.ToInt16(pcmFile, i * 2)).ToArray();
+            await Task.Run(() => Process.Start("audiowaveform", $"-i \"{item.Path}\" -o \"{outputPath}\"").WaitForExit()).ConfigureAwait(false);
         }
 
-        // return item;
+        var fileStream = new FileStream(outputPath, FileMode.Open, FileAccess.Read, FileShare.Read);
+
+        return fileStream;
     }
 }
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -195,6 +195,7 @@
  - [Kenneth Cochran](https://github.com/kennethcochran)
  - [benedikt257](https://github.com/benedikt257)
  - [revam](https://github.com/revam)
+ - [FrenchGithubUser](https://github.com/FrenchGithubUser)
 
 # Emby Contributors
 


gnattu (changes requested):
This requires there is a separate binary installed on the server which is not quite desirable, and the binary you are using supports less audio formats than Jellyfin supports which will also be a problem.

FrenchGithubUser:
> This requires there is a separate binary installed on the server which is not quite desirable, and the binary you are using supports less audio formats than Jellyfin supports which will also be a problem.

After [discussing](https://github.com/jellyfin/jellyfin-meta/discussions/89), I couldn't find a better solution. The best would have been to use ffmpeg (imo, as it is already a dependency of jellyfin and supports many audio formats), but the only outputs possible are raw binary or pcm. Both generate files that are around 3 times as big as the bbc's library does.

The lightest seems to be pcm files. If a few hundred KB files are fine, then I could go this way instead.

Which is what is done in this commit : https://github.com/jellyfin/jellyfin/pull/13486/commits/234eb00d52931a572e52cef7b60f51c580610cd9

Let me know what would be the best.

gnattu:
The first problem is that it does not look like there are many renderers of the audiowaveform format that BBC is providing, other than the renderer they provided in their own release which is not easy for integration, makes the use of this format a questionable choice. 

A second problem is that even if we want to take this approach we need to write our own implementation that generate jsons compliant to that spec and integrate it in server, because the original binaries included too many features that we don't need. Including a new binary is also very very problematic because we need to make our own pipeline that build for all the targets we want to support and trust me that is not easy, especially when the upstream has no intention to build on macOS and Windows at all.

FrenchGithubUser:
I found a way to use ffprobe instead, creating a csv (simple list of values) which weights less than 5KiB for most tracks.

Let me know what you think about this approach.

Jellyfin.Api/Helpers/AudioHelper.cs
```diff
@@ -154,4 +159,31 @@ public async Task<ActionResult> GetAudioStream(
...
+        var libraryOptions = BaseItem.LibraryManager.GetLibraryOptions(item);
+        var saveInMediaFolder = libraryOptions.SaveLocalMetadata;
+        var saveFileName = Path.GetFileNameWithoutExtension(item.Path) + ".dat";
+        string outputPath = saveInMediaFolder
+            ? Path.Combine(item.ContainingFolderPath, saveFileName)
+            : Path.Combine(item.GetInternalMetadataPath(), saveFileName);
+
+        if (!File.Exists(outputPath))
+        {
+            await Task.Run(() => Process.Start("ffprobe", $"-v error -f lavfi -i \"amovie={item.Path},asetnsamples=44100,astats=metadata=1:reset=1\" -show_entries frame_tags=lavfi.astats.Overall.RMS_level -of csv=p=0 -o \"{outputPath}\"").WaitForExit()).ConfigureAwait(false);
```
gnattu, line 182: - 44100 sample batch is only about 1 sample per second and that is too low to generate a good waveform. This value should be dynamic and being adjusted accordingly from the input source audio's sample rate to make each sample has a constant time.
- RMS_level is a bad choice for waveforms. The effective energy level from root mean squares are vary "flat" when plotted, and the squared value is always positive which means this will look differently from the waveforms you see from most audio software, where the wave is from the raw signal which has both positive and negative amplitudes.

gnattu (changes requested):

Jellyfin.Api/Helpers/AudioHelper.cs
gnattu: This is how your waveform is going to look like, as you can see it has a quite boring shape:

image

FrenchGithubUser:
Thank you for your feedback, indeed that doesn't look good. I edited the code to have a dynamic sample rate (2 measurements per second seem enough to me and also avoids creating a file that would be too heavy). Also, I used the RMS_peak measurement, which seemed to give the best results.

Comparison with soundcloud for a title : (https://soundcloud.com/soulji/soulji-elf)

image

Soundcloud actually draws one wave and the bottom is just the mirror of what's above. Their server returns a single list of values : https://wave.sndcdn.com/hMbPJeotBAlV_m.json

Let me know if those modifications fit the requirements better. And what's left to change to make this PR valid.

Thanks again for the assistance !

Jellyfin.Api/Helpers/AudioHelper.cs
```diff
@@ -154,4 +159,44 @@ public async Task<ActionResult> GetAudioStream(
...
+                if (process != null)
+                {
+                    string output = await process.StandardOutput.ReadToEndAsync().ConfigureAwait(false);
+                    // This value sets the length of a frame, the amount of measurements per second.
+                    int samplesPerSecond = 2;
+                    sampleRate = Math.Max(int.Parse(output, System.Globalization.CultureInfo.InvariantCulture) / samplesPerSecond, 1);
+                }
+            }
+
+            await Task.Run(() => Process.Start("ffprobe", $"-v error -f lavfi -i \"amovie={item.Path},asetnsamples={sampleRate},astats=metadata=1:reset=1\" -show_entries frame_tags=lavfi.astats.Overall.RMS_peak -of csv=p=0 -o \"{outputPath}\"").WaitForExit()).ConfigureAwait(false);
```
gnattu, line 195: The path should be escaped carefully because all kind of special characters that will crash the cli may exist

Jellyfin.Api/Helpers/AudioHelper.cs
```diff
@@ -154,4 +159,44 @@ public async Task<ActionResult> GetAudioStream(
...
+        }
+
+        var fileStream = new FileStream(outputPath, FileMode.Open, FileAccess.Read, FileShare.Read);
+
+        return fileStream;
```
gnattu, line 200: To make this production ready I really don't think serve the csv as is a good enough. We need some sort of server side process to make the client developer's life easier. For example make it a json response to match the behavior of may other apis, this would also allow us to extend this api without breaking all clients. The JSON should at least the sample duration and the sample array consists all samples, so that the client doesn't have to hard code an early development decision about the duration because that is subject to change. The actual values being measured will be discussed separately because what ffmpeg returns may not be the best to plot a soundwave.

gnattu (changes requested):
So we are getting closer and wee need to do more work to make this actually useable. 

I don't really think SoundCloud generates the wave based on RMS_Peak, but instead they just doing their own sample on max/min peak of the sampled duration. But we are jellyfin, not SoundCloud, so we don't have to match their behavior.

The value returned from ffmpeg is measured in dBFS but I found that value is not visually appealing when plotted as it flattens the amplitude too much. When I restored the value to (approximate) its raw value, the shape looks much better:

image

This is also from the song you linked to. It does look differently from what SoundCloud does but it still looks good, especially when we downsample it further to create a bar chart:

image

The formula for raw value restoration is `2^(value/6)`.

Jellyfin.Api/Helpers/AudioHelper.cs
```diff
@@ -154,4 +159,44 @@ public async Task<ActionResult> GetAudioStream(
...
+            ?? throw new ResourceNotFoundException();
+
+        var libraryOptions = BaseItem.LibraryManager.GetLibraryOptions(item);
+        var saveInMediaFolder = libraryOptions.SaveLocalMetadata;
+        var saveFileName = Path.GetFileNameWithoutExtension(item.Path) + ".csv";
+        string outputPath = saveInMediaFolder
+            ? Path.Combine(item.ContainingFolderPath, saveFileName)
+            : Path.Combine(item.GetInternalMetadataPath(), saveFileName);
+
+        if (!File.Exists(outputPath))
```
gnattu, line 180: Also although on-demand generation is okay for this use case (ffmpeg does so in reasonable time), but we would still need a proper metadata manager for it and have proper db entry to record this. Relying solely on file path is very fragile as the files can be moved, deleted or updated. In this case the sample file also needs to be regenerated and the old invalid file should be removed.

FrenchGithubUser: yes that's true ! could you guide me a bit on how I could add a metadata manager and a db entry ?

FrenchGithubUser:
I did all the changes requested, except for the metadata manager/db entry, I started looking around but couldn't figure it out.

Also, a json is generated but a csv file is also generated (and removed) as an intermediary file. It would be better to redirect the output of ffprobe instead but I also could not figure it out properly.

Could I get some guidance on those things and feedback on potential other changes ?

Thanks again !

Jellyfin.Api/Helpers/AudioHelper.cs
```diff
@@ -154,4 +160,61 @@ public async Task<ActionResult> GetAudioStream(
...
+            var samples = csvLines
+                    .Select(v =>
+                {
+                    if (v.Trim() == "-inf")
+                    {
+                        return 0.0;  // sometimes ffprobe returns "-inf" instead of a number
+                    }
+                    else
+                    {
+                        return Math.Pow(2, double.Parse(v, CultureInfo.InvariantCulture) / 6);
```
gnattu, line 206: `double.Parse` will throw. Either provide a fallback here or just catch and fail the metadata generation and respond with empty samples.

Jellyfin.Api/Helpers/AudioHelper.cs
```diff
@@ -154,4 +160,61 @@ public async Task<ActionResult> GetAudioStream(
...
+        {
+            var processStartInfo = new ProcessStartInfo("ffprobe", $"-v error -select_streams a:0 -show_entries stream=sample_rate -of csv=p=0 \"{item.Path}\"") { RedirectStandardOutput = true, UseShellExecute = false };
+            int sampleRate = 0;
+            int fileSampleRate = item.GetMediaStreams()[0].SampleRate ?? 44100;
+            // This value sets the length of a frame, the amount of measurements per second.
+            int samplesPerSecond = 2;
+            sampleRate = Math.Max(fileSampleRate / samplesPerSecond, 1);
+
+            // escape double quotes and backslashes
+            string itemPathEscaped = item.Path.Replace("\\", "\\\\", StringComparison.Ordinal).Replace("\"", "\\\"", StringComparison.Ordinal);
```
gnattu, line 192: I'm not sure if this escape is valid under windows where backslashes have its own use. Wouldn't this just escaping each and every valid windows path?

gnattu (changes requested):
> Probably rounding the samples as well (less digits for each value)

I don't think rounding is a good idea because some clients may want higher precision data to scale their waveform. But if we choose that path, my recommendation is to establish a set of valid levels, such as the maximum number of heights a “bar” in the waveform can have. We multiply this floating-point number by the maximum valid height and discard all decimal parts. For instance, we could define that there can only be 128 levels of height. We multiply the raw value by 128 and remove the decimal parts, retaining only the integer value. 

Another approach to save storage space without affecting the precision is to store samples in native format of float so that each values only consumes 4 bytes, instead of be proportional to the length of its string representation. But this approach needs a very careful design of the binary format and it could be harder to extend in the future. An easy implementation would be storing the external metadata like sample duration in the database, and the reference this binary file in a column. In the file we first use an integer followed by a null char to indicate how may samples this file has, and then we read 4 bytes at a time, try to parse the binary into a float till the end of the file. If any of the 4 byte cannot be interpreted as a valid float or the sample amount is not right we discard that file and return empty samples. This approach would require some careful handling of the binary file and depending on the file size it may be an overkill. My take on this is, if the json size is not larger than the lyric file size of the song (too much) we don't have to worry about the size.

Jellyfin.Api/Helpers/AudioHelper.cs
```diff
@@ -154,4 +160,61 @@ public async Task<ActionResult> GetAudioStream(
...
+            string outputPathCsvEscaped = outputPathCsv.Replace("\\", "\\\\", StringComparison.Ordinal).Replace("\"", "\\\"", StringComparison.Ordinal);
+            await Task.Run(() => Process.Start("ffprobe", $"-v error -f lavfi -i \"amovie={itemPathEscaped},asetnsamples={sampleRate},astats=metadata=1:reset=1\" -show_entries frame_tags=lavfi.astats.Overall.RMS_peak -of csv=p=0 -o \"{outputPathCsvEscaped}\"").WaitForExitAsync()).ConfigureAwait(false);
```
JPVenson, line 194: This task serves no real purpose. You spin up a task, for the sole purpose of awaiting another task.

Jellyfin.Api/Helpers/AudioHelper.cs
```diff
@@ -154,4 +160,61 @@ public async Task<ActionResult> GetAudioStream(
...
+                        return 0.0;  // sometimes ffprobe returns "-inf" instead of a number
+                    }
+                    else
+                    {
+                        return Math.Pow(2, double.Parse(v, CultureInfo.InvariantCulture) / 6);
+                    }
+                })
+                .ToArray();
+
+            var json = JsonSerializer.Serialize(new { samples, sampleRate });
```
JPVenson, line 211: dont use anoymous types for serialization.

JPVenson (changes requested):

Jellyfin.Api/Helpers/AudioHelper.cs
FrenchGithubUser: This was a mistake, forgot to remove it

FrenchGithubUser:
> > Probably rounding the samples as well (less digits for each value)
> 
> I don't think rounding is a good idea because some clients may want higher precision data to scale their waveform. But if we choose that path, my recommendation is to establish a set of valid levels, such as the maximum number of heights a “bar” in the waveform can have. We multiply this floating-point number by the maximum valid height and discard all decimal parts. For instance, we could define that there can only be 128 levels of height. We multiply the raw value by 128 and remove the decimal parts, retaining only the integer value.
> 
> Another approach to save storage space without affecting the precision is to store samples in native format of float so that each values only consumes 4 bytes, instead of be proportional to the length of its string representation. But this approach needs a very careful design of the binary format and it could be harder to extend in the future. An easy implementation would be storing the external metadata like sample duration in the database, and the reference this binary file in a column. In the file we first use an integer followed by a null char to indicate how may samples this file has, and then we read 4 bytes at a time, try to parse the binary into a float till the end of the file. If any of the 4 byte cannot be interpreted as a valid float or the sample amount is not right we discard that file and return empty samples. This approach would require some careful handling of the binary file and depending on the file size it may be an overkill. My take on this is, if the json size is not larger than the lyric file size of the song (too much) we don't have to worry about the size.

Yes I see, the result file is a bit less than twice the size of the .lrc files, would you consider that fine ? (a few KiB)

Jellyfin.Api/Helpers/AudioHelper.cs
```diff
@@ -154,4 +160,61 @@ public async Task<ActionResult> GetAudioStream(
...
+            await File.WriteAllTextAsync(outputPathJson, json).ConfigureAwait(false);
+            File.Delete(outputPathCsv);
+        }
+
+        var fileStream = new FileStream(outputPathJson, FileMode.Open, FileAccess.Read, FileShare.Read);
+
+        return fileStream;
```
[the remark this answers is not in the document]
FrenchGithubUser: I created a class `AudioWaveformInfo` to store some metadata in the database (but not the actual wave data itself). Should I also create a dto only for sending back the data to the client ?

FrenchGithubUser:
I implemented all the required changes, except for the Dto return (see my comment on it : https://github.com/jellyfin/jellyfin/pull/13486#discussion_r1948083592)

I could not test the code as the AudioWaveformInfos table was not created automatically, is there a migration to run or something similar ?

Jellyfin.Server.Implementations/AudioWaveform/AudioWaveformManager.cs
```diff
@@ -0,0 +1,170 @@
...
+            int fileSampleRate = item.GetMediaStreams()[0].SampleRate ?? 44100;
+            // This value sets the length of a frame, the amount of measurements per second.
+            int samplesPerSecond = 2;
+            sampleRate = Math.Max(fileSampleRate / samplesPerSecond, 1);
+
+            // escape double quotes
+            string itemPathEscaped = item.Path.Replace("\"", "\\\"", StringComparison.Ordinal);
+            string outputPathCsvEscaped = outputPathCsv.Replace("\"", "\\\"", StringComparison.Ordinal);
+
+            var process = Process.Start("ffprobe", $"-v error -f lavfi -i \"amovie={itemPathEscaped},asetnsamples={sampleRate},astats=metadata=1:reset=1\" " + $"-show_entries frame_tags=lavfi.astats.Overall.RMS_peak -of csv=p=0 -o \"{outputPathCsvEscaped}\"");
```
JPVenson, line 98: `reset=1\" " + $"`
Why concatinate that string? Use a raw string instead 
```csharp

"""
this is a raw string
"""
```

Jellyfin.Server.Implementations/AudioWaveform/AudioWaveformManager.cs
```diff
@@ -0,0 +1,170 @@
...
+
+            var waitTask = process.WaitForExitAsync();
```
JPVenson, line 100:
```suggestion
            await process.WaitForExitAsync().WaitAsync(Timespan.FromSecounds(10));
```

https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.waitasync?view=net-9.0#system-threading-tasks-task-waitasync(system-timespan)

Jellyfin.Server.Implementations/AudioWaveform/AudioWaveformManager.cs
```diff
@@ -0,0 +1,170 @@
...
+                    else
+                    {
+                        return double.TryParse(v, NumberStyles.Float, CultureInfo.InvariantCulture, out double parsedValue)
+                            ? Math.Pow(2, parsedValue / 6)
+                            : 0.0;
+                    }
+                })
+                .ToArray();
+
+            var data = new Dictionary<string, object>
```
JPVenson, line 129: please use an actual c# class instead.
The Json serializer caches type information and it makes working with those informations much more structured

Jellyfin.Server.Implementations/AudioWaveform/AudioWaveformManager.cs
```diff
@@ -0,0 +1,170 @@
...
+                { "samples", samples },
+                { "sampleRate", sampleRate }
+            };
+
+            var json = JsonSerializer.Serialize(data);
+            await File.WriteAllTextAsync(outputPathJson, json).ConfigureAwait(false);
+            File.Delete(outputPathCsv);
+        }
+
+        var fileStream = new FileStream(outputPathJson, FileMode.Open, FileAccess.Read, FileShare.Read);
```
JPVenson, line 140: move the creation of the FileStream as late as possible. If something crashes in the SaveAudioWaveform method, this might lead to memory leaks

JPVenson (changes requested):
You are still missing a migration and configuration file (which we prefer over attribute based configuration) to setup your database changes.

Jellyfin.Api/Controllers/AudioController.cs
```diff
@@ -300,8 +305,8 @@ [FromQuery] [RegularExpression(EncodingHelper.ValidationRegex)] string? audioCod
         [FromQuery] int? cpuCoreLimit,
         [FromQuery] string? liveStreamId,
         [FromQuery] bool? enableMpegtsM2TsMode,
-        [FromQuery] [RegularExpression(EncodingHelper.ValidationRegex)] string? videoCodec,
-        [FromQuery] [RegularExpression(EncodingHelper.ValidationRegex)] string? subtitleCodec,
+        [FromQuery][RegularExpression(EncodingHelper.ValidationRegex)] string? videoCodec,
```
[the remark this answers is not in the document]
FrenchGithubUser: that's vscode's autoformatter somehow

Jellyfin.Server.Implementations/AudioWaveform/AudioWaveformManager.cs
```diff
@@ -0,0 +1,170 @@
...
+        _mediaEncoder = mediaEncoder;
+        _fileSystem = fileSystem;
+        _encodingHelper = encodingHelper;
+        _libraryManager = libraryManager;
+        _dbProvider = dbProvider;
+        _appPaths = appPaths;
+    }
+
+    /// <inheritdoc />
+    public async Task<FileStream> GetAudioWaveformAnsyc(Guid itemId)
```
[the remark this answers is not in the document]
FrenchGithubUser: it should take a few seconds (for less than 5min audio tracks, which is most of the time). But I haven't tried edge cases

Jellyfin.Server.Implementations/AudioWaveform/AudioWaveformManager.cs
```diff
@@ -0,0 +1,170 @@
...
+    {
+        var tempDir = Path.Combine(_appPaths.TempDirectory, "audiowaveform_" + itemId);
+        Directory.CreateDirectory(tempDir);
+
+        var item = _libraryManager.GetItemById<BaseItem>(itemId)
+            ?? throw new ResourceNotFoundException();
```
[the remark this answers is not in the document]
FrenchGithubUser: the thing is that if I remove the catch, I get a possible null reference. I checked other managers and many do that. I'm not sure what I should do, pass I the info I need (item paths, etc.) to the manager ? or keep it this way ?

Jellyfin.Server.Implementations/AudioWaveform/AudioWaveformManager.cs
```diff
@@ -0,0 +1,170 @@
...
+        _encodingHelper = encodingHelper;
+        _libraryManager = libraryManager;
+        _dbProvider = dbProvider;
+        _appPaths = appPaths;
+    }
+
+    /// <inheritdoc />
+    public async Task<FileStream> GetAudioWaveformAnsyc(Guid itemId)
+    {
+        var tempDir = Path.Combine(_appPaths.TempDirectory, "audiowaveform_" + itemId);
```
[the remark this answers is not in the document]
FrenchGithubUser: right, this is a problem indeed

Jellyfin.Server.Implementations/AudioWaveform/AudioWaveformManager.cs
```diff
@@ -85,6 +85,7 @@ public async Task<FileStream> GetAudioWaveformAnsyc(Guid itemId)
             string outputPathCsv = Path.Combine(tempDir, Path.ChangeExtension(saveFileName, "csv"));
             int sampleRate = 0;
             var mediaStreams = item.GetMediaStreams();
+            // error : mediaStreams[0].SampleRate can be null ? but why ? it is checked just before with .HasValue.
             int fileSampleRate = (mediaStreams.Count > 0 && mediaStreams[0].SampleRate.HasValue) ? mediaStreams[0].SampleRate.Value : 44100;
```
JPVenson, line 89:
```suggestion
            int fileSampleRate = mediaStreams.FirstOrDefault()?.SampleRate ?? 44100;
```

Jellyfin.Server.Implementations/AudioWaveform/AudioWaveformManager.cs
FrenchGithubUser: Where should I create this class ? In a dedicated file, in a specific folder ? I'm sorry but I'm a bit lost, there seems to be multiple ways to do it and I'm not sure which one would be the most appropriate

FrenchGithubUser:
Hi, are there any news on this ? @gnattu @JPVenson 

But no pressure ofc, just wondering if it got at the bottom of the inbox :)

gnattu:
Well, this API won’t be included in the next release because the EFCore migration is already extensive (and there are still bugs after that migration). So, after the next major version is released, more team members will review this API and provide more advice on its actual design, data storage, and so on. If you’re unsure if you’ll still be available then, one of our team members may pick up your work, but it would be best if you’re still available by then.

FrenchGithubUser:
> Well, this API won’t be included in the next release because the EFCore migration is already extensive (and there are still bugs after that migration). So, after the next major version is released, more team members will review this API and provide more advice on its actual design, data storage, and so on. If you’re unsure if you’ll still be available then, one of our team members may pick up your work, but it would be best if you’re still available by then.

Are there any estimated dates on when this will happen ? But I think I should still be available.

Thanks for the quick reply anyways !

FrenchGithubUser:
Hi, just wondering if there is any ETA on when this will be reviewed/discussed 

@gnattu

JPVenson:
@FrenchGithubUser this will only get some serious attention from the server devs if its out of Draft status.

Architecture or in depth planning of features should always done beforehand in our meta repo.

thornygravy:
hoo-ha gnattu has saved the day lets get this ball rolling folks

Jellyfin.Api/Controllers/AudioController.cs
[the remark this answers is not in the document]
JPVenson: other attributes are all using a shared [] please follow that style or maintain the spaces as before.

Jellyfin.Api/Controllers/AudioController.cs
```diff
@@ -364,4 +376,25 @@ [FromQuery] [RegularExpression(EncodingHelper.ValidationRegex)] string? subtitle
...
     }
+
+    /// <summary>
+    /// Gets audio data to draw a waveform bar.
+    /// </summary>
+    /// <param name="itemId">The item id.</param>
+    /// <response code="200">AudioWaveForm returned.</response>
+    /// <returns>A <see cref="FileResult"/> containing the audiowaveform file.</returns>
+    [HttpGet("{itemId}/audiowaveform")]
+    [ProducesResponseType(StatusCodes.Status200OK)]
```
JPVenson, line 387: missing response code 404

Jellyfin.Data/Entities/AudioWaveformInfo.cs
```diff
@@ -0,0 +1,29 @@
...
+public class AudioWaveformInfo
+{
+    /// <summary>
+    /// Gets or sets the ID of the associated item.
+    /// </summary>
+    /// <remarks>
+    /// Required. Acts as the primary key.
+    /// </remarks>
+    [Key]
+    [JsonIgnore]
```
JPVenson, line 19: Why is that nessesary?

Jellyfin.Data/Entities/AudioWaveformInfo.cs
```diff
@@ -0,0 +1,29 @@
...
+    public Guid ItemId { get; set; }
```
JPVenson, line 20: You are missing the navigation property

```
public BaseItemEntity Item { get; set; }
```

if not added efcore does not know this is a forgin key

Jellyfin.Data/Entities/AudioWaveformInfo.cs
```diff
@@ -0,0 +1,29 @@
...
+/// </summary>
+public class AudioWaveformInfo
+{
+    /// <summary>
+    /// Gets or sets the ID of the associated item.
+    /// </summary>
+    /// <remarks>
+    /// Required. Acts as the primary key.
+    /// </remarks>
+    [Key]
```
JPVenson, line 18: Use configuration class instead of annotations

Jellyfin.Server.Implementations/AudioWaveform/AudioWaveformManager.cs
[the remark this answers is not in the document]
JPVenson: You should hand in a CancelationToken from the original request to allow aborting the operation

Jellyfin.Server.Implementations/AudioWaveform/AudioWaveformManager.cs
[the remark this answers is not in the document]
JPVenson: Then you have to check if you actually get an item and return a value that the controller can interpret for succes. e.g return null and check in the controller for null.

Jellyfin.Server.Implementations/AudioWaveform/AudioWaveformManager.cs
```diff
@@ -0,0 +1,160 @@
...
+        var saveInMediaFolder = BaseItem.LibraryManager.GetLibraryOptions(item).SaveLocalMetadata;
+        var saveFileName = Path.ChangeExtension(item.Path, "json");
+        string outputPathJson = saveInMediaFolder
+            ? Path.Combine(item.ContainingFolderPath, saveFileName)
+            : Path.Combine(item.GetInternalMetadataPath(), saveFileName);
+
+        if (!File.Exists(outputPathJson))
+        {
+            var tempDir = Path.Combine(_appPaths.TempDirectory, "audiowaveform_" + itemId);
+            Directory.CreateDirectory(tempDir);
```
JPVenson, line 84: this can fail if the directory already exists

Jellyfin.Server.Implementations/AudioWaveform/AudioWaveformManager.cs
```diff
@@ -0,0 +1,160 @@
...
+            string outputPathCsv = Path.Combine(tempDir, Path.ChangeExtension(saveFileName, "csv"));
```
JPVenson, line 85: use `var`

Jellyfin.Server.Implementations/AudioWaveform/AudioWaveformManager.cs
```diff
@@ -0,0 +1,160 @@
...
+            int samplesPerSecond = 2;
+            sampleRate = Math.Max(fileSampleRate / samplesPerSecond, 1);
+
+            // escape double quotes
+            string itemPathEscaped = item.Path.Replace("\"", "\\\"", StringComparison.Ordinal);
+            string outputPathCsvEscaped = outputPathCsv.Replace("\"", "\\\"", StringComparison.Ordinal);
+
+            var process = Process.Start("ffprobe", $"-v error -f lavfi -i \"amovie={itemPathEscaped},asetnsamples={sampleRate},astats=metadata=1:reset=1\" -show_entries frame_tags=lavfi.astats.Overall.RMS_peak -of csv=p=0 -o \"{outputPathCsvEscaped}\"");
+
+            int timeout = 10; // seconds, this is quite an arbitrary value
```
JPVenson, line 100: this should be a `const` on class level

Jellyfin.Server.Implementations/AudioWaveform/AudioWaveformManager.cs
```diff
@@ -0,0 +1,160 @@
...
+        {
+            var tempDir = Path.Combine(_appPaths.TempDirectory, "audiowaveform_" + itemId);
+            Directory.CreateDirectory(tempDir);
+            string outputPathCsv = Path.Combine(tempDir, Path.ChangeExtension(saveFileName, "csv"));
+            int sampleRate = 0;
+            var mediaStreams = item.GetMediaStreams();
+            // error : mediaStreams[0].SampleRate can be null ? but why ? it is checked just before with .HasValue.
+            int fileSampleRate = (mediaStreams.Count > 0 && mediaStreams[0].SampleRate.HasValue) ? mediaStreams[0].SampleRate.Value : 44100;
+            // This value sets the length of a frame, the amount of measurements per second.
+            int samplesPerSecond = 2;
```
JPVenson, line 91: this should be a `const` on class level

Jellyfin.Server.Implementations/AudioWaveform/AudioWaveformManager.cs
JPVenson: classes always go into dedicated files.
Next to this one should be fine.

Jellyfin.Server.Implementations/AudioWaveform/AudioWaveformManager.cs
```diff
@@ -0,0 +1,160 @@
...
+                })
+                .ToArray();
+
+            var data = new Dictionary<string, object>
+            {
+                { "samples", samples },
+                { "sampleRate", sampleRate }
+            };
+
+            var json = JsonSerializer.Serialize(data);
```
JPVenson, line 126:
```suggestion
            var json = JsonSerializer.Serialize(data, JsonDefaults.Options);
```
Always set the JsonSerializerOptions as this is where the typecache lives

JPVenson (changes requested):

FrenchGithubUser:
Considering the comment on this [discussion](https://github.com/jellyfin/jellyfin-meta/discussions/89#discussioncomment-12665592), I think that it makes more sense for me to wait until there is a clearer description of how this api endpoint should be before I make any more changes

Chaphasilor:
I think what was shown in https://github.com/jellyfin/jellyfin/pull/13486#pullrequestreview-2602759484 is already easily usable by clients. Rendering a sort of bar chart based on a bunch of values should be relatively trivial in any client that isn't template-based.  
But I'd be interested if more restrictive platforms such as Roku could work with this. What do you think @1hitsong?

1hitsong:
> But I'd be interested if more restrictive platforms such as Roku could work with this. What do you think @1hitsong?

Is there a sample of the output from the endpoint anywhere?

Roku can draw rectangles and lines if we give it coordinates. No matter what, we'll have to play around to get it working, if we can. 🤞

FrenchGithubUser:
> > But I'd be interested if more restrictive platforms such as Roku could work with this. What do you think @1hitsong?
> 
> Is there a sample of the output from the endpoint anywhere?
> 
> Roku can draw rectangles and lines if we give it coordinates. No matter what, we'll have to play around to get it working, if we can. 🤞

[01 - Par ici.json](https://github.com/user-attachments/files/19674555/01.-.Par.ici.json)

Here is the output of this song : https://www.youtube.com/watch?v=VpPjL_7HDqY&pp=ygUOcGFyIGljaSA3IGphd3M%3D

FrenchGithubUser:
are there some updates on this ? could someone give more insights on how the code should be restructured if the current implementation isn't okay ?

PR #13489, opened 2025-02-03
Fix subnet contains check

gnattu:
We are still using `Subnet.Contains` a lot but that does not handle IPv4 mapped to IPv6 addresses at all. It was partially fixed by #12094 in local network checking, but it may not always happen on LAN.

Also make all local network checking to use IsInLocalNetwork method instead of just performing `Subnet.Contains` which is not accurate.

Filter out all link-local addresses for external interface matching.

**Changes**

**Issues**

### DIFF
--- a/src/Jellyfin.Networking/Manager/NetworkManager.cs
+++ b/src/Jellyfin.Networking/Manager/NetworkManager.cs
@@ -692,7 +692,7 @@ public class NetworkManager : INetworkManager, IDisposable
             if (_remoteAddressFilter.Any() && !IsInLocalNetwork(remoteIP))
             {
                 // remoteAddressFilter is a whitelist or blacklist.
-                var matches = _remoteAddressFilter.Count(remoteNetwork => SubNetContainsAddress(remoteNetwork, remoteIP));
+                var matches = _remoteAddressFilter.Count(remoteNetwork => SubnetContainsAddress(remoteNetwork, remoteIP));
                 if ((!config.IsRemoteIPFilterBlacklist && matches > 0)
                     || (config.IsRemoteIPFilterBlacklist && matches == 0))
                 {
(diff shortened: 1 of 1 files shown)


src/Jellyfin.Networking/Manager/NetworkManager.cs
```diff
@@ -919,6 +909,19 @@ public bool IsLinkLocalAddress(IPAddress address)
...
+        ArgumentNullException.ThrowIfNull(address);
+        ArgumentNullException.ThrowIfNull(network);
+
+        if (address.IsIPv4MappedToIPv6)
+        {
+            address = address.MapToIPv4();
+        }
+
+        return network.Contains(address);
+    }
```
Shadowghost, line 923: Since this is universal, it should be moved to `NetworkUtils`

Shadowghost (approved):

src/Jellyfin.Networking/Manager/NetworkManager.cs
gnattu: Lets break ABI for next release

Shadowghost: Since we only add something new I don't see why it should be a problem.

src/Jellyfin.Networking/Manager/NetworkManager.cs
```diff
@@ -919,6 +909,19 @@ public bool IsLinkLocalAddress(IPAddress address)
         return NetworkConstants.IPv4RFC3927LinkLocal.Contains(address) || address.IsIPv6LinkLocal;
     }
 
+    private static bool SubNetContainsAddress(IPNetwork network, IPAddress address)
```
crobibero, line 912:
```suggestion
    private static bool SubnetContainsAddress(IPNetwork network, IPAddress address)
```

crobibero:
@gnattu The typo correction needs to be applied in more places :)

PR #13491, opened 2025-02-04
Don't allow library name with leading or trailing space

### LINKED ISSUE #13429: Adding a space after Display Name in library throws error
### Description of the bug

In the master branch (post EFCore merge) a space following the display name when adding a library throws an error. 

### Reproduction steps

1. On the master branch, add a new movie library, and name it "Movies " (note the space after "Movies").
2. Observe errors

Doesn't allow folder to be created, throws errors in logs

Folder is added

### Jellyfin Server version

Master

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

b318f335991167102a5fa8d65030d200bbec898d

### Environment

```markdown
- OS: Windows 11
- Linux Kernel:
- Virtualization: Debug on VSCode
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

```shell
[15:42:53] [INF] [41] Emby.Server.Implementations.IO.LibraryMonitor: Watching directory F:\Movies DVHDR
[15:42:54] [INF] [12] Emby.Server.Implementations.IO.LibraryMonitor: Watching directory F:\TV
Exception thrown: 'System.IO.DirectoryNotFoundException' in System.Private.CoreLib.dll
[15:42:54] [ERR] [67] Jellyfin.Api.Middleware.ExceptionMiddleware: Error processing request: Could not find a part of the path 'C:\Users\max\AppData\Local\jellyfin\root\default\Movies test space after \movies.collection'. URL POST /Library/VirtualFolders.
[15:42:54] [INF] [26] Emby.Server.Implementations.IO.LibraryMonitor: Watching directory I:\Movies DVHDR



Exception has occurred: CLR/System.IO.DirectoryNotFoundException
An exception of type 'System.IO.DirectoryNotFoundException' occurred in System.Private.CoreLib.dll but was not handled in user code: 'Could not find a part of the path 'C:\Users\max\AppData\Local\jellyfin\root\default\Movies test space after \movies.collection'.'
   at Microsoft.Win32.SafeHandles.SafeFileHandle.CreateFile(String fullPath, FileMode mode, FileAccess access, FileShare share, FileOptions options)
   at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(String fullPath, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallocationSize, Nullable`1 unixCreateMode)
   at System.IO.File.OpenHandle(String path, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallocationSize)
   at System.IO.File.<<WriteAllBytesAsync>g__Core|103_0>d.MoveNext()
   at Emby.Server.Implementations.Library.LibraryManager.<AddVirtualFolder>d__181.MoveNext() in \\wsl.localhost\Ubuntu-24.04\home\max\jellyfin\Emby.Server.Implementations\Library\LibraryManager.cs:line 2889
   at Emby.Server.Implementations.Library.LibraryManager.<AddVirtualFolder>d__181.MoveNext() in \\wsl.localhost\Ubuntu-24.04\home\max\jellyfin\Emby.Server.Implementations\Library\LibraryManager.cs:line 2916
   at Jellyfin.Api.Controllers.LibraryStructureController.<AddVirtualFolder>d__5.MoveNext() in \\wsl.localhost\Ubuntu-24.04\home\max\jellyfin\Jellyfin.Api\Controllers\LibraryStructureController.cs:line 91
   at Microsoft.AspNetCore.Mvc.Infrastructure.ActionMethodExecutor.TaskOfActionResultExecutor.<Execute>d__0.MoveNext()
   at System.Runtime.CompilerServices.ValueTaskAwaiter`1.GetResult()
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.<<InvokeActionMethodAsync>g__Awaited|12_0>d.MoveNext()
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.<<InvokeNextActionFilterAsync>g__Awaited|10_0>d.MoveNext()
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

gnattu:
Windows has a specific requirement on filenames that cannot have leading or trailing spaces and the folder creation would fail. Reject such request in the api controller.

**Changes**

**Issues**

Fixes #13429

### DIFF
--- a/Jellyfin.Api/Controllers/LibraryStructureController.cs
+++ b/Jellyfin.Api/Controllers/LibraryStructureController.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
+using System.Diagnostics;
 using System.Globalization;
 using System.IO;
 using System.Linq;
@@ -81,6 +82,12 @@ public class LibraryStructureController : BaseJellyfinApiController
         [FromBody] AddVirtualFolderDto? libraryOptionsDto,
         [FromQuery] bool refreshLibrary = false)
     {
+        // Windows does not allow files or folders with names that has leading or trailing spaces
+        if (name.Length != name.Trim().Length)
+        {
+            return BadRequest();
+        }
+
         var libraryOptions = libraryOptionsDto?.LibraryOptions ?? new LibraryOptions();
 
         if (paths is not null && paths.Length > 0)
--- a/Jellyfin.Api/Controllers/LibraryStructureController.cs
+++ b/Jellyfin.Api/Controllers/LibraryStructureController.cs
@@ -1,7 +1,6 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
-using System.Diagnostics;
 using System.Globalization;
 using System.IO;
 using System.Linq;
--- a/Jellyfin.Api/Controllers/LibraryStructureController.cs
+++ b/Jellyfin.Api/Controllers/LibraryStructureController.cs
@@ -75,18 +75,14 @@ public class LibraryStructureController : BaseJellyfinApiController
     [HttpPost]
     [ProducesResponseType(StatusCodes.Status204NoContent)]
     public async Task<ActionResult> AddVirtualFolder(
-        [FromQuery] string name,
+        [FromQuery]
+        [RegularExpression(@"^(?:\S(?:.*\S)?)$", ErrorMessage = "Library name cannot be empty or have leading/trailing spaces.")]
+        string name,
         [FromQuery] CollectionTypeOptions? collectionType,
         [FromQuery, ModelBinder(typeof(CommaDelimitedArrayModelBinder))] string[] paths,
         [FromBody] AddVirtualFolderDto? libraryOptionsDto,
         [FromQuery] bool refreshLibrary = false)
     {
-        // Windows does not allow files or folders with names that has leading or trailing spaces
-        if (name.Length != name.Trim().Length)
-        {
-            return BadRequest();
-        }
-
         var libraryOptions = libraryOptionsDto?.LibraryOptions ?? new LibraryOptions();
 
         if (paths is not null && paths.Length > 0)


Jellyfin.Api/Controllers/LibraryStructureController.cs
```diff
@@ -81,6 +81,12 @@ public async Task<ActionResult> AddVirtualFolder(
         [FromBody] AddVirtualFolderDto? libraryOptionsDto,
         [FromQuery] bool refreshLibrary = false)
     {
+        // Windows does not allow files or folders with names that has leading or trailing spaces
+        if (name.Length != name.Trim().Length)
```
nielsvanvelzen, line 85: This kind of validation should preferably be done with a regex attribute so it is added to the open api spec. Additionally web changes need to be made to implement this same validation locally to hint a user about invalid values.

nielsvanvelzen (changes requested):

Jellyfin.Api/Controllers/LibraryStructureController.cs
```diff
@@ -81,6 +81,12 @@ public async Task<ActionResult> AddVirtualFolder(
...
+        {
+            return BadRequest();
```
JPVenson, line 87: i totally get why we reject that, but it would not be obvious at all for a user _why_ it failed which in this instance can be quite frustraiting as its not obvious.

Either we should just trim and use the processed input or this requires a webPR too.

gnattu: > Either we should just trim

Some user may have fancy ideas to pick such name and we should warn them instead of doing silent corrections and get "bug reports" that the spaces are ignored

JPVenson: i dont think so really. If i put spaces in an JF ignores them/trims them out its better to just fail and dont tell the user why.

Either like so or we rewrite the server code to trim spaces for paths always anyway and allow the user to set the name to whatever.

gnattu: > i dont think so really. If i put spaces in an JF ignores them/trims them out its better to just fail and dont tell the user why.

Then make a web change to tell user why

Jellyfin.Api/Controllers/LibraryStructureController.cs
```diff
@@ -75,7 +75,9 @@ public ActionResult<IEnumerable<VirtualFolderInfo>> GetVirtualFolders()
     [HttpPost]
     [ProducesResponseType(StatusCodes.Status204NoContent)]
     public async Task<ActionResult> AddVirtualFolder(
-        [FromQuery] string name,
+        [FromQuery]
+        [RegularExpression(@"^(?:\S(?:.*\S)?)$", ErrorMessage = "Library name cannot be empty or have leading/trailing spaces.")]
```
crobibero, line 79: Can this be simplified to `^\b.+\b$`?

gnattu: Would't this also reject some valid library names that has punctuation marks at the beginning or the end of the name? Like `My Favourites!` will fail this regex

crobibero (approved):

Bond-009 (approved):

EraYaN:
As an evolution of this for later, we should maybe just slugify the library name and use that internally for filenames etc. Would also allow emojis and all other random nonsense to be in library names since it would just end up being the "Display Name".

gnattu:
Actually emoji is already possible because emojis are indeed valid file name on modern filesystem. It is just the font required for render it is not properly loaded but that will be fixed by #13231

EraYaN:
Maybe something like this: https://richjenks.com/filename-regex/ to stop con etc from being used as well, if we don't already filter for this.

gnattu:
That regex also reject non-latin characters which is not we want. If we really want to sanitize the file path we just hash the library name and use that as the actual folder name instead of using the library name as is. The problem of that would be the current data migration (we have absolute path stored in too many places) and the extra complexity for manual management.

EraYaN:
We can grandfather old libraries with their slug being the current name. And instead of hashing using slugify (.Core from nuget) is much nicer imo, to not have random hashes in file paths. Much better to browse manually. But that is for another time. Maybe after everything if in EF Core we can start migrations to remove path prefixes etc. But that is for another time.

One the regex, it's actually quite hard to get Javascript to do unicode correctly in a regex, .NET and JS have very different ideas of what is the right way. So while choosing it we need to be cognizant of that.

JPVenson (approved):

nielsvanvelzen (approved):

Issue #13496, opened 2025-02-04
LiveTV Guide Backdrop images not updating.


IDisposable:
If the guide data changes an any other image, then the _Backdrop_ image will not be updated.

The changed line that causes the issue is: 
```csharp
// Backdrop
return updated || UpdateImage(ImageType.Backdrop, item, info);
```

This occurs because the `||` operator is a short-circuit so  if `updated` was already `true` then the call to `UpdateImage` would never execute, thus the _Backdrop_ image would not be updated.

This was unintentionally introduced in PR #13227 [here](https://github.com/jellyfin/jellyfin/pull/13227/files#diff-68cbfbf495bc9462524e5e837822194c426c5008491dfccdbc8773ef784aea50R682) and subsequently backported to main branch.

Note that if we made another guide data update, and none of the other images changed in this pass, then the _Backdrop_ would eventually be updated.

Shadowghost:
Already fixed in #13469

IDisposable:
No, that fixes an entirely different issue. The erroneous line is still present in #13469 [here](https://github.com/jellyfin/jellyfin/pull/13469/files#diff-68cbfbf495bc9462524e5e837822194c426c5008491dfccdbc8773ef784aea50R663)

PR #13227, opened 2024-12-14
Fix EPG image caching

Shadowghost:
**Changes**
* Only return cachable images to the `ListingManager` so we don't save any image URLs with API key in the database
* Run pre-caching on all existing programs, not only updated/new ones

**Issues**
Fixes #13147

**ToDo**
* Potentially write migration to remove all existing images referencing schedules direct, otherwise it will take 12 days for them to be removed from the database and the issue to be actually fixed.

**Note**
This is a workaround since the proper fix would require quite some database changes which we currently can't do in a point release due to the ongoing EFCore migration. The proper fix would entail something along these lines:
* Extend `ItemImageInfo` with new fields `ProviderName` and `RequiresToken`
* Write migration for existing data
* Add interface for external ImageURL resolution
* Implement Interface for `SchedulesDirect`
* When getting images of an item, run the path through the providers if `RequiresToken=true`, provider with matching name handles resolution
* Return that URL

PR #13497, opened 2025-02-04
Fix LiveTV Guide Backdrop image not updating

IDisposable:
**Changes**
Replace the short-circuiting `||` operator with a normal `|` OR operator to ensure that the call for `UpdateImage` for the _Backdrop_ image still happens even if other images were *also* updated.

```csharp
// Backdrop
return updated || UpdateImage(ImageType.Backdrop, item, info);
```

**Issues**
Fixes #13496

PR #13504, opened 2025-02-05
Fix LiveTV Guide Backdrop image not updating

IDisposable:
If the guide data changes any other image, the Backdrop image would not be updated.

This error was unintentionally introduced in PR https://github.com/jellyfin/jellyfin/pull/13227 [here](https://github.com/jellyfin/jellyfin/pull/13227/files#diff-68cbfbf495bc9462524e5e837822194c426c5008491dfccdbc8773ef784aea50R682)

## Changes
Replace the short-circuiting || operator with a normal | OR operator to ensure that the call for UpdateImage for the Backdrop image still happens even if other images were also updated.

```csharp
// Backdrop
return updated || UpdateImage(ImageType.Backdrop, item, info);
```
## Issues
Fixes https://github.com/jellyfin/jellyfin/issues/13496

### DIFF
--- a/src/Jellyfin.LiveTv/Guide/GuideManager.cs
+++ b/src/Jellyfin.LiveTv/Guide/GuideManager.cs
@@ -679,7 +679,9 @@ public class GuideManager : IGuideManager
         updated |= UpdateImage(ImageType.Logo, item, info);
 
         // Backdrop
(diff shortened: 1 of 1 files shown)


IDisposable:
Retarget of previous PR #13497

Shadowghost (approved):

crobibero (approved):

PR #13516, opened 2025-02-07
Fix Search results are case-sensitive for people

### LINKED ISSUE #13510: Search results are case-sensitive for people
### Description of the bug

On the master branch, search for a person using all lowercase. Observe no results. 

### Reproduction steps

1. Search for a person using all lowercase

Person will not show up

Person results should not be case-sensitive

### Jellyfin Server version

Master

### Specify commit id

00b66a06eac4d1331d59a6a9b7de3e99088ebaa9

_No response_

### Specify version number

_No response_

### Specify the build version

2025020500

### Environment

```markdown
- OS: Windows 11
- Virtualization: No
```

### Jellyfin logs

```shell
none
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

tkloy24:
The search for people is currently case-sensitive (#13510).

**Changes**
The SQL query has been modified to make the search case-insensitive.

**Issues**
Fixes #13510

### DIFF
--- a/Jellyfin.Server.Implementations/Item/PeopleRepository.cs
+++ b/Jellyfin.Server.Implementations/Item/PeopleRepository.cs
@@ -155,7 +155,7 @@ public class PeopleRepository(IDbContextFactory<JellyfinDbContext> dbProvider, I
 
         if (!string.IsNullOrWhiteSpace(filter.NameContains))
         {
-            query = query.Where(e => EF.Functions.Like(e.Name, $"%{filter.NameContains}%"));
+            query = query.Where(e => e.Name.ToUpper().Contains(filter.NameContains.ToUpper()));
         }
 
         return query;
--- a/Jellyfin.Server.Implementations/Item/PeopleRepository.cs
+++ b/Jellyfin.Server.Implementations/Item/PeopleRepository.cs
@@ -155,7 +155,7 @@ public class PeopleRepository(IDbContextFactory<JellyfinDbContext> dbProvider, I
 
         if (!string.IsNullOrWhiteSpace(filter.NameContains))
         {
-            query = query.Where(e => e.Name.Contains(filter.NameContains));
+            query = query.Where(e => EF.Functions.Like(e.Name, $"%{filter.NameContains}%"));
         }
 
         return query;
--- a/Jellyfin.Server.Implementations/Item/PeopleRepository.cs
+++ b/Jellyfin.Server.Implementations/Item/PeopleRepository.cs
@@ -158,7 +158,8 @@ public class PeopleRepository(IDbContextFactory<JellyfinDbContext> dbProvider, I
 
         if (!string.IsNullOrWhiteSpace(filter.NameContains))
         {
-            query = query.Where(e => e.Name.ToUpper().Contains(filter.NameContains.ToUpper()));
+            var nameContainsUpper = filter.NameContains.ToUpper();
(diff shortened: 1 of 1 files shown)


Jellyfin.Server.Implementations/Item/PeopleRepository.cs
```diff
@@ -155,7 +155,7 @@ private IQueryable<People> TranslateQuery(IQueryable<People> query, JellyfinDbCo
 
         if (!string.IsNullOrWhiteSpace(filter.NameContains))
         {
-            query = query.Where(e => e.Name.Contains(filter.NameContains));
+            query = query.Where(e => EF.Functions.Like(e.Name, $"%{filter.NameContains}%"));
```
JPVenson, line 158: `Like` isn't nessesarly case insensitive.
```suggestion
            query = query.Where(e => e.Name.ToUpper() == filter.NameContains.ToUpper());
```

JPVenson (changes requested):

Jellyfin.Server.Implementations/Item/PeopleRepository.cs
```diff
@@ -155,7 +155,7 @@ private IQueryable<People> TranslateQuery(IQueryable<People> query, JellyfinDbCo
 
         if (!string.IsNullOrWhiteSpace(filter.NameContains))
         {
-            query = query.Where(e => e.Name.Contains(filter.NameContains));
+            query = query.Where(e => e.Name.ToUpper().Contains(filter.NameContains.ToUpper()));
```
JPVenson, line 158: please check for warnings and errors generated by our editorconfig file. Same as with ItemRepository you have to disable them.

https://github.com/jellyfin/jellyfin/actions/runs/13206101876/job/36869455854?pr=13516#step:4:46

JPVenson (changes requested):

Jellyfin.Server.Implementations/Item/PeopleRepository.cs
IkeVoodoo: Apologies, shouldn't you calculate `filter.NameContains` outside of `Where` to make sure you're not converting the string to uppercase each time you check?

tkloy24: Fair point, that way it will be more efficient.

JPVenson: No that is not how entity framework works. In this instance, it makes no difference.

jtholen001: Butting in here, but why not use `StringComparison.OrdinalIgnoreCase` as in `query = query.Where(e => e.Name.Contains(filter.NameContains, StringComparison.OrdinalIgnoreCase));` 

This avoids creating a new string each time, and the super edge case of the turkish i problem, and just generally being culture invariant.

JPVenson (approved):

Jellyfin.Server.Implementations/Item/PeopleRepository.cs
JPVenson: @jtholen001 
https://learn.microsoft.com/en-us/ef/core/providers/sqlite/functions#string-functions
That would efcore require to have knowlage about non-standard collations that could be used for case insensetive comparisons which is not a feature they provide.

Bond-009 (approved):

Issue #1269, opened 2019-04-21
[JF >= 10.3] thumbnail extraction now pegs CPU at 100%


greyvv:
New version maxes cpu and memory usage climbs to 100%.  Deleted existing docker and recreated new one from scratch.  same result

reverted docker to 10.2.2.  scans directories fine.  cpu and memory increase by expected amount

mikey242:
Yeah I have this problem too. Crashed my pi3 several times now due to intense cpu and mem usage.

pedro-araujo:
I believe this might be related to this FFMPEG issue:
- https://www.reddit.com/r/jellyfin/comments/bfe2c5/ffmpeg_consuming_vast_cpu_with_no_playback_since/

Can you guys check your log files and paste them here?

ferferga:
Same here. My first diagnose started by running a ssh session and seeing where the memory usage was going (as I thought it was related to rclone instead) using top. Saw this:

![Lots of ffprobe](https://user-images.githubusercontent.com/10274099/56472215-fd8a1f80-645b-11e9-8e13-b8cbd3caaa07.png)

I have many logs (from ffmpeg and jellyfin, different dates, captured exactly after a reboot and before my Pi dies). Which ones are necessary?

ferferga:
@pedro-araujo About the thread you mentioned: I also had same errors in 10.2 version, but it seems that it's getting worse in 10.3, as Jellyfin isn't capable of kill ffmpeg as the old version did.

joshuaboniface:
First question for everyone here: is "Enable chapter image extraction" enabled for your TV libraries? This is my first though: the upgrade triggered a scan, and with this option enabled JF spawns numerous ffmpeg processes to perform the image extraction. The issue seems to be that these are all started at once, rather than sequentially for a whole season. I'm doing some more testing to reproduce but confirmation of that setting will help.

joshuaboniface:
OK I think there's actually two issues going on here that are being conflated.

@greyvv, your output indicates that the processes causing issues are `ffprobe` - does the actual `ffmpeg` command ever start and use CPU?

The `ffprobe` calls are a standard part of the library scan that can't be disabled, since they're extracting the chapters from the media. The processes do look to be triggered asynchronously which is why you see dozens of processes. If this is really intensive on the system, it might be a storage speed problem.

Then there's the issue with chapter *image* extraction, which is what the Reddit post is about. I'm going to make a separate issue for that.

Nosirus:
[...]

PR #13519, opened 2025-02-07
Removed unneccessary array creation

IkeVoodoo:
**Changes**
1. Removed unnecessary array creation after `Where()`.
2. Replaced LINQ usage with a `foreach` loop as to avoid allocating a new `IEnumerable` instance.

**Issues**
Related to #1269


IkeVoodoo:
It seems as if I was on an old version of the project. My bad.

PR #13532, opened 2025-02-10
Fix image encoding concurrency limit

### LINKED ISSUE #13531: Parallel task limits not being honored.
### Description of the bug

I read in another bug report that you can limit the number of threads during library scans with some configs (in the screenshot), but it doesn't appear to work.  the parallelism is set to 10, but as you can see in this screenshot there are waaaaaaaaay more than 10

Image

### Reproduction steps

New jellyfin library with thousands of high quality high frame rate videos, set the parallel library scan task limit and parallel image encoding limit to a low number, restart jellyfin for good measure, then trigger a library scan.  

parallel task count isn't honored.

parallel task count config is used to limit the number of files being scanned concurrently.

### Jellyfin Server version

10.10.0+

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.10.5

### Environment

```markdown
- OS: win server 2k22
- Linux Kernel:
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method: 
- Hardware Acceleration:
- GPU Model: quatro p600
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking: 
- Storage: drive being used atm has about 12 12tb drives in a raid 6. speeds are high.
```

### Jellyfin logs

```shell
too much censoring is required...
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

Image

### Additional information

This server isn't supposed to be just for jellyfin to run on... In order to play with your software I needed to shut down all my VMs and anything that might use memory/CPU.  It's likely going to crash due to the high memory usage...

gnattu:
The current FFmpeg image extractor is configured to use a resource pool size that always equals 2 times the number of CPU cores, which is somewhat excessive. Make the default equal to the core count instead of twice, and respect the `ParallelImageEncodingLimit` option.

**Changes**

**Issues**

Fixes #13531

### DIFF
--- a/MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
+++ b/MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
@@ -122,7 +122,12 @@ namespace MediaBrowser.MediaEncoding.Encoder
             _jsonSerializerOptions = new JsonSerializerOptions(JsonDefaults.Options);
             _jsonSerializerOptions.Converters.Add(new JsonBoolStringConverter());
 
-            var semaphoreCount = 2 * Environment.ProcessorCount;
+            var semaphoreCount = serverConfig.Configuration.ParallelImageEncodingLimit;
+            if (semaphoreCount < 1)
+            {
+                semaphoreCount =  Environment.ProcessorCount;
+            }
+
             _thumbnailResourcePool = new(semaphoreCount);
         }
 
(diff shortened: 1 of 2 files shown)


JPVenson (approved):

Bond-009 (approved):

crobibero (changes requested):
This is causing tests to fail with a null ref

crobibero (approved):

Issue #13250, opened 2024-12-18
Attempting to delete a single 'extras' file deletes all extras in that folder (and the folder)


TimF0:
**Describe The Bug**
Deleting a single 'extras' file via the web interface appears to delete all 'extras' file

**Steps To Reproduce**
1. Find a TV show with > 1 extras against a season
2. Choose one of the extras and use the three dots option to bring up the options menu
3. Select 'delete media'
4. Accept warning prompt
5. Web interface appears to only delete the selected file, but in fact all extras files and extras library are deleted (show files remain untouched), and any attempt to play the remaining extras files results in a fatal error message

**Expected Behavior**
Where multiple extras files exist in the extras directory, deleting a single file to delete should only delete that file and leave the others unchanged

**Logs**
n/a

**Screenshots**
![Image](https://github.com/user-attachments/assets/f34ba15f-275b-479f-8630-44e9061fa40b)
![Image](https://github.com/user-attachments/assets/c4424b36-499b-4e4d-beb3-9c6374f87fe4)
![Image](https://github.com/user-attachments/assets/7c1776b4-3756-453c-8a21-bfcd7b8ec4e7)
![Image](https://github.com/user-attachments/assets/b9a7372e-9428-4f0d-98ae-1b2777e48713)
![Image](https://github.com/user-attachments/assets/48c83593-e0e7-42c3-87b0-70f8cee963f1)
![Image](https://github.com/user-attachments/assets/b502015a-45e6-4d1f-a3a9-12d53e15b8a3)

**System (please complete the following information):**
 - Platform: Unraid 6.12.13 (Jellyfin docker)
 - Browser: Firefox 133.0.3
 - Jellyfin Version: 10.10.3

**Additional Context**

theguymadmax:
I just tested this, and I can confirm that deleting an extra deletes the entire folder, not just the extra file.

```
[2024-12-18 19:46:10.445 +00:00] [INF] [28] Emby.Server.Implementations.Library.LibraryManager: Removing item, Type: "Video", Name: "Attack of the Hackers", Path: "/Media/Test/WarGames (1983)/Attack of the Hackers 2008-featurette.mp4", Id: 28884252-e169-b5a1-5ff3-c5fa707f892b
[...]

PR #13536, opened 2025-02-10
Fix IsInMixedFolder not being set for Extras

tkloy24:
**Changes**
Changed FindExtras() and GetExtra() to identify if there are more than one file in the directory. If this is the case, the files will receive "IsInMixedFolder=true" to avoid issues for subsequent steps like deleting a single Extra.

**Issues**
#13250


Shadowghost (approved):
@tkloy24 can you please rebase once on current master if you have some time.

crobibero (approved):

Bond-009 (approved):

PR #13547, opened 2025-02-11
Don't rewrite unchanged NFO files

### LINKED ISSUE #12978: nfo saver makes unneccesary disk writes during library scan

### Description of the bug

When re-scanning a library using the .nfo saver, all .nfo files are updated even if there were no changes.

While mostly benign, this does result in unnecessary disk writes and cause some backup software and software-based raid systems (like SnapRAID) to treat the file as new because of the changed timestamp. I am not sure if this is a bug for feature request, so apologies if it is misfiled.

I believe this operation is being handled by SaveToFileAsync () on [line 204 of   jellyfin/MediaBrowser.XbmcMetadata/Savers/BaseNfoSaver.cs](https://github.com/jellyfin/jellyfin/blob/bf00899f92881c987f019ad7d20f0cef42d4e3e7/MediaBrowser.XbmcMetadata/Savers/BaseNfoSaver.cs#L204)

I am not a C# programmer, nor am I familiar with the codebase and am therefore uncomfortable submitting a pull request on my own, but believe that this could be improved without noticeably impacting scan performance by adding a simple CRC32 hash to compare the stream with the existing file before writing.

Here is my attempt at a patch if helpful, I am not confident that it will work as written, but hate pointing out problems without at least suggesting a solution:

```diff
         private async Task SaveToFileAsync(Stream stream, string path)
         {
             var directory = Path.GetDirectoryName(path) ?? throw new ArgumentException($"Provided path ({path}) is not valid.", nameof(path));
             Directory.CreateDirectory(directory);
+		// Compare metadata hashes before proceeding
+		if (File.Exists(path))
+		{
+			byte[] existingFileHash;
+			using (var existingFileStream = File.OpenRead(path))
+			{
+				existingFileHash = new Crc32().ComputeHash(existingFileStream);
+			}
+
+			byte[] newContentHash;
+			using (var crc32 = new Crc32())
+			{
+				newContentHash = crc32.ComputeHash(stream);
+			}
+
+			stream.Position = 0;
+
+			if (existingFileHash.SequenceEqual(newContentHash))
+			{
+				return; // Don't save since .nfo is unchanged
+			}
+		}
+
            // On Windows, saving the file will fail if the file is hidden or readonly
            FileSystem.SetAttributes(path, false, false);

            var fileStreamOptions = new FileStreamOptions()
            {
                Mode = FileMode.Create,
                Access = FileAccess.Write,
                Share = FileShare.None,
                PreallocationSize = stream.Length,
                Options = FileOptions.Asynchronous
            };

            var filestream = new FileStream(path, fileStreamOptions);
            await using (filestream.ConfigureAwait(false))
            {
                await stream.CopyToAsync(filestream).ConfigureAwait(false);
            }

            if (ConfigurationManager.Configuration.SaveMetadataHidden)
            {
                SetHidden(path, true);
            }
        }
```

### Reproduction steps

1.	Monitor the filesystem for changes (for example using auditd or inotifywait on Linux). Alternatively, just look at the timestamps for .nfo files before and after a library scan.
2.	Ensure that .nfo saver is for the library and that it has previously saved files
3.	Wait for a scheduled library scan or manually kickoff the job
4.	Observe that all .nfo get re-saved with new timestamps

After each library scan with nfo saver on, all .nfo files are recreated regardless of changes to the metadata.

That only new and changed .nfo files are added to the filesystem.

### Jellyfin Server version

10.10.0+

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

 10.10.1

### Environment

- OS: Debian bookworm
- Linux Kernel: 6.1.0-20
- Virtualization: Docker (lscr.io/linuxserver/jellyfin:latest)
- Clients: All
- Browser: All
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration: No
- GPU Model:
- Plugins:
- Reverse Proxy: nginx
- Base URL:
- Networking: Docker internal network
- Storage: EXT4 using Docker bind

### Jellyfin logs

None relevant

### FFmpeg logs

None relevant

### Client / Browser logs

None relevant

None relevant

IDisposable:
When computing the NFO contents, we will rewrite the file even if everything is unchanged. This causes the file updated time to change and increases load on the filesystem (and underlying drives).

Add `ComputeHash` to compute the  _XxHash3_ hash of the file-on-disk and the stream we've just created and if the same just returns.

Also now passes the `cancellationToken` down to the stream copy so we abort faster if cancelled.

Fixes #12978

### DIFF
--- a/MediaBrowser.XbmcMetadata/Savers/BaseNfoSaver.cs
+++ b/MediaBrowser.XbmcMetadata/Savers/BaseNfoSaver.cs
@@ -5,6 +5,7 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Security.Cryptography;
 using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading;
@@ -197,14 +198,31 @@ namespace MediaBrowser.XbmcMetadata.Savers
 
                 cancellationToken.ThrowIfCancellationRequested();
 
-                await SaveToFileAsync(memoryStream, path).ConfigureAwait(false);
+                await SaveToFileAsync(memoryStream, path, cancellationToken).ConfigureAwait(false);
             }
         }
 
-        private async Task SaveToFileAsync(Stream stream, string path)
+        private async Task SaveToFileAsync(Stream stream, string path, CancellationToken cancellationToken)
         {
             var directory = Path.GetDirectoryName(path) ?? throw new ArgumentException($"Provided path ({path}) is not valid.", nameof(path));
             Directory.CreateDirectory(directory);
+
+            // Compare metadata hashes before proceeding
+            if (File.Exists(path))
+            {
+                byte[] existingFileHash;
+                using (var existingFileStream = File.OpenRead(path))
+                {
+                    existingFileHash = await ComputeHash(existingFileStream, cancellationToken).ConfigureAwait(false);
+                }
+
+                byte[] newContentHash = await ComputeHash(stream, cancellationToken).ConfigureAwait(false);
+
+                if (existingFileHash.SequenceEqual(newContentHash))
+                {
+                    return; // Don't save since .nfo is unchanged
+                }
+            }
 
             // On Windows, saving the file will fail if the file is hidden or readonly
             FileSystem.SetAttributes(path, false, false);
@@ -221,13 +239,25 @@ namespace MediaBrowser.XbmcMetadata.Savers
             var filestream = new FileStream(path, fileStreamOptions);
             await using (filestream.ConfigureAwait(false))
             {
-                await stream.CopyToAsync(filestream).ConfigureAwait(false);
+                await stream.CopyToAsync(filestream, cancellationToken).ConfigureAwait(false);
             }
 
             if (ConfigurationManager.Configuration.SaveMetadataHidden)
             {
                 SetHidden(path, true);
             }
+        }
+
+        private static async Task<byte[]> ComputeHash(Stream stream, CancellationToken cancellationToken)
+        {
+            byte[] hash;
+            using (SHA256 hasher = SHA256.Create())
+            {
+                hash = await hasher.ComputeHashAsync(stream, cancellationToken).ConfigureAwait(false);
+                stream.Position = 0;    // reseek to the stream beginning for the upcoming save
+            }
+
+            return hash;
         }
 
         private void SetHidden(string path, bool hidden)
--- a/MediaBrowser.XbmcMetadata/Savers/BaseNfoSaver.cs
+++ b/MediaBrowser.XbmcMetadata/Savers/BaseNfoSaver.cs
@@ -4,6 +4,7 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
+using System.IO.Hashing;
 using System.Linq;
 using System.Security.Cryptography;
 using System.Text;
@@ -250,14 +251,12 @@ namespace MediaBrowser.XbmcMetadata.Savers
 
         private static async Task<byte[]> ComputeHash(Stream stream, CancellationToken cancellationToken)
         {
-            byte[] hash;
-            using (SHA256 hasher = SHA256.Create())
-            {
-                hash = await hasher.ComputeHashAsync(stream, cancellationToken).ConfigureAwait(false);
-                stream.Position = 0;    // reseek to the stream beginning for the upcoming save
-            }
+            stream.Position = 0;
+            var hasher = new XxHash3();
+            await hasher.AppendAsync(stream, cancellationToken).ConfigureAwait(false);
+            stream.Position = 0;
 
-            return hash;
+            return hasher.GetCurrentHash();
         }
 
         private void SetHidden(string path, bool hidden)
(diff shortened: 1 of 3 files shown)


MediaBrowser.XbmcMetadata/Savers/BaseNfoSaver.cs
```diff
@@ -230,6 +248,18 @@ private async Task SaveToFileAsync(Stream stream, string path)
             }
         }
 
+        private static async Task<byte[]> ComputeHash(Stream stream, CancellationToken cancellationToken)
+        {
+            byte[] hash;
+            using (SHA256 hasher = SHA256.Create())
```
IDisposable, line 254: I opted for SHA256 here, but could easily be any other hasher. This method also could be extracted to some common library.

IDisposable:
Let me know if I should target a different branch... I chose master because this won't be fully useful until we emit stable NFO files 

(I have a pending PR #9560 that will ensure we always emit the same NFO instead of the current behavior where we swap albums and genre and such around from run-to-run due to the collections not being sorted)

MediaBrowser.XbmcMetadata/Savers/BaseNfoSaver.cs
Shadowghost: Maybe we can use XxHash3 as in #11816 since it is way faster and still good enough

gnattu:
I’m still not entirely convinced about implementing such function using a hashing approach. Now, the I/O becomes read and write, and an additional computational overhead has been introduced without any apparent benefits. The read + compute time might already cost about the same time just write the nfo to the disk, and we are very likely to have a performance regression just by checksumming the NFO. Do you have any benchmarks of the performance impact of this PR?

IDisposable:
It's not just the read/write + hashing time (though that is very small indeed, especially because NFO files are almost always one logical block on-disk since they are tiny), we're concerned about the wear-leveling of storage and having a stable NFO so our backup tools (offsite or other servers) aren't constantly updating the file for no reason.

MediaBrowser.XbmcMetadata/Savers/BaseNfoSaver.cs
IDisposable: Happy to swap to that. One moment.

IDisposable: Now using XxHash3

gnattu:
> I'm confused, is this not the server's code I modified?

No. What your implemented is a hack at best to hide a more complicated defect of the metadata saver. The metadata saver does have an enum of `ItemUpdateType` passed in and for NFO savers it will never trigger the saving action when the update type is `None`. The reason you’re seeing those unnecessary writing is because some metadata services are flagging updates without checking if the actual value has changed. For example, this one (not the complete list, just one case that is obvious):

https://github.com/jellyfin/jellyfin/blob/dd7a804cfbf43198e84a36f4f00eaf43e01c5cbc/MediaBrowser.Providers/Manager/MetadataService.cs#L839-L847

If we want this to work properly we need to fix each and every case of the actual update logic instead of implement a hack that deal with the things on the filesystem. By doing those update correctly we can also reduce unnecessary DB writes because the DB save commands will not be called either if the update type is `None`.

IDisposable:
Yes, it certainly would be great to _also_ make changes upstream, but that's a MUCH bigger project.

The one advantage of doing this now is that it gets measurable progress, and builds motivation to make those upstream changes.

gnattu:
> The one advantage of doing this now is that it gets measurable progress, and builds motivation to make those upstream changes.

I’m afraid this will actually do the opposite because people will lose motivation to improve it since a hack exists that makes the issue disappear from their filesystem metadata. They can then call it a day and pretend this issue never existed. And a portion of code that will get removed after proper implementation is far from "measurable progress". A measurable progress is that we fix the metadata updater one case by one case until all updates are properly tracked and flagged.

I'm still not convinced that we need to workaround the NFO writer's behavior that much so that we have to implement a hack right now. Out of so many metadata savers behave like this, why only the NFO gets special treatment? Other savers will also increase `wear-leveling of storage` and make your backup tools unhappy. 

Please don't get me wrong. I'm aware that the current behavior is suboptimal and needs improvement, but the way this PR is doing it is far from optimal IMO. If enough team members are fine with this approach and approve this PR, then I'm fine with that. This is just my personal opinions after all.

IDisposable:
> Out of so many metadata savers behave like this, why only the NFO gets special treatment?

I all transparency, I store the NFO files on the storage server and all the other services I use (Kodi / ChannelsDVR / MediaMonkey / etc.) ALL reread those files any time the NFO is changed... so that one useless write hammers all my other clients/servers with null updates.

In other words, I noticed the problem myself and went looking for other NFO issues that I might also be able to fix.

IDisposable:
> A measurable progress is that we fix the metadata updater one case by one case until all updates are properly tracked and flagged.

Totally agreed, and I would love to participate, assuming the changes are accepted one-by-one, because I certainly can't do all of them in sync.

IDisposable:
Rebased

JPVenson:
closed due to inactivity

IDisposable:
@JPVenson so I can assume that we aren't interested in this change? I don't think it actually has any merge conflicts.

Bond-009:
Is computing a hash here really faster than a simple compare where you can do an early exit?

IDisposable:
Probably not, I could certainly just code as a binary compare as these are generally quite small. I guess I leaned toward hashing due to FTP servers where you can ask it to compute the CRC and only download those 4 bytes 😁.

Happy to switch to byte-wise compare

Bond-009:
@IDisposable Sounds good :+1:

IDisposable:
Switched to byte-by-byte compare, but I can't reopen this... want a new PR @Bond-009 ?

Issue #13164, opened 2024-12-05
[EfJellyfin.db] Live TV Guide issues


theguymadmax:
### This issue respects the following points:

- [x] This is a **bug**, not a question or a configuration issue; Please visit our forum or chat rooms first to troubleshoot with volunteers, before creating a report. The links can be found [here](https://jellyfin.org/contact/).
- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [x] I'm using an up to date version of Jellyfin Server stable, unstable or master; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Description of the bug

After migrating to EfJellyfin.db, the live TV guide will appear blank after performing a refresh guide data.

### Reproduction steps

1. Migrate to EFCore branch
2. Refresh Guide Data (XMl)

### What is the current _bug_ behavior?

There are no shows listed in the guide. 

### What is the expected _correct_ behavior?

The guide is populated with shows.

### Jellyfin Server version

Master

### Specify the build version

EFCore Branch

### Environment

```markdown
Any
```

### Jellyfin logs

```shell
Nothing in logs.
```

### FFmpeg logs

```shell

```

### Relevant screenshots or videos

10.10.3
![Image](https://github.com/user-attachments/assets/01b997ff-f3be-46a9-9a44-e4640e598872)

10.10.11 EFCore
![Image](https://github.com/user-attachments/assets/b4d58291-b9d5-402f-b6cd-585ea7b67299)

fishcharlie:
Maybe a duplicate of #13147?

theguymadmax:
> Maybe a duplicate of [#13147](https://github.com/jellyfin/jellyfin/issues/13147)?

Thank you for pointing that out. However, I can confirm that this is not a duplicate of that issue.

crobibero:
https://github.com/jellyfin/jellyfin/issues/13047#issuecomment-2495726330

Issue #13047, opened 2024-11-16
[Group] Jellyfin.db EFCore refactor issue


JPVenson:
This is the group issue for Bugs releated to the refactor of library.db into the jellyfin.db.

crobibero:
- [ ] LiveTV program ChannelIds are saved as `Guid.ToString("D")` but then queried as `Guid.ToString("N")`. Consider changing the `BaseItemEntity.ChannelId` to be a `Guid` instead of `string`

gnattu:
> * [ ]  LiveTV program ChannelIds are saved as `Guid.ToString("D")` but then queried as `Guid.ToString("N")`. Consider changing the `BaseItemEntity.ChannelId` to be a `Guid` instead of `string`[ ]  Unable to view Episodes when going Show -> Season for migrated shows. New shows seem to be linked correctly. If you go to an episode view you can navigate back to its parent season

Can’t really use guid for channel id for LiveTV because some LiveTV plugins will not generate valid channel ID because it was not required to be a guid, and this field may be very hard to migrate for those plugins.

joshuaboniface:
Can't replicate @crobibero 's issue 2, at least for a few shows on my migrated library. It does take 2-3 seconds to load the episodes, but then they load.

TheMelmacian:
At the moment the OwnerId of BaseItemEntity is of type `string` while all other fields that reference the Id (Id, ParentId, TopParentId, SeriesId, SeasonId) are of type `Guid`:

https://github.com/jellyfin/jellyfin/blob/7684986fa16cd6246d4929097f76d379f55fc1fe/Jellyfin.Data/Entities/BaseItemEntity.cs#L137

While both Id and OwnerId are defined as text in the database, due to the different types in the code the former is saved as upper case and the latter is saved as lower case. This makes any comparison between these two fields inconvenient and inefficient.
So maybe it makes sense to convert OwnerId to a Guid?

FerronN:
I read about the refactoring in the RC release notes. Does this also enable us to use other EF Core database providers like postgresql or mysql?

PR #13553, opened 2025-02-14
Change BaseItemEntity ChannelId to nullable Guid

crobibero:
Fixes: https://github.com/jellyfin/jellyfin/issues/13164

BaseItem.ChannelId has been a Guid since Jellyfin's inception, should be safe to change the db model to reflect that.

### DIFF
--- a/Jellyfin.Data/Entities/BaseItemEntity.cs
+++ b/Jellyfin.Data/Entities/BaseItemEntity.cs
@@ -22,7 +22,7 @@ public class BaseItemEntity
 
     public DateTime? EndDate { get; set; }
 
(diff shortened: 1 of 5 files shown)


gnattu:
This might break a lot of third party livetv plugins though as the channelid was being written as arbitrary stuff, but major version breakage is acceptable?

crobibero:
I don't think this actually is a breaking change as everywhere we parse that value from the database as a Guid. 
I also don't know of any LiveTv plugins so that would de difficult to verify

Bond-009 (approved):

theguymadmax:
Fixes: #13164

PR #13556, opened 2025-02-14
Don't allow usernames to have leading or trailing spaces

Jxiced:
**Changes**
Perform a trim on the entered username and throw a different exception if the name exists but with a whitespace.

### DIFF
--- a/Jellyfin.Server.Implementations/Users/UserManager.cs
+++ b/Jellyfin.Server.Implementations/Users/UserManager.cs
@@ -113,7 +113,7 @@ namespace Jellyfin.Server.Implementations.Users
         // This is some regex that matches only on unicode "word" characters, as well as -, _ and @
         // In theory this will cut out most if not all 'control' characters which should help minimize any weirdness
         // Usernames can contain letters (a-z + whatever else unicode is cool with), numbers (0-9), at-signs (@), dashes (-), underscores (_), apostrophes ('), periods (.) and spaces ( )
-        [GeneratedRegex(@"^[\w\ \-'._@+]+$")]
+        [GeneratedRegex(@"^(?!\s)[\w \-'._@]+(?<!\s)$")]
         private static partial Regex ValidUsernameRegex();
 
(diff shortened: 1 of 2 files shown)


gnattu:
Yes. Don't do the trim. If we want to do validation, we just reject the request when it has trailing whitespaces

Jxiced:
So updating this change to make ThrowIfInvalidUsername check for whitespaces, like in this new change, then adding validation to the front-end on jellyfin-web would be a good idea?

gnattu (changes requested):
Just update the ValidUsernameRegex. I think `^(?!\s)[\w \-'._@]+(?<!\s)$` would do the trick.

Jxiced:
Thanks for the suggestion! Please see the revised code.

gnattu (changes requested):

Jellyfin.Server.Implementations/Users/UserManager.cs
```diff
@@ -113,7 +113,7 @@ public UserManager(
         // This is some regex that matches only on unicode "word" characters, as well as -, _ and @
         // In theory this will cut out most if not all 'control' characters which should help minimize any weirdness
         // Usernames can contain letters (a-z + whatever else unicode is cool with), numbers (0-9), at-signs (@), dashes (-), underscores (_), apostrophes ('), periods (.) and spaces ( )
-        [GeneratedRegex(@"^[\w\ \-'._@+]+$")]
+        [GeneratedRegex(@"^(?!\s)[\w \-'._@]+(?<!\s)$")]
```
gnattu, line 116:
```suggestion
        [GeneratedRegex(@"^(?!\s)[\w\ \-'._@+]+(?<!\s)$")]
```

gnattu (approved):
LGTM on server side. Need to implement similar check on the web to notice the user that this name is invalid.

Bond-009 (approved):

cvium (approved):

PR #13561, opened 2025-02-14
Patch for WebOS issue 248 (Hanging on DV files)

patrickd77-eng:
**Changes**

The change aims to only impact the WebOS client by limiting segment length to 1 instead of 6. The author, fhelland,  described the issue it fixes as this:

“Using a shorter hls_time value of 1 second ensures that the playlist generated by FFmpeg matches the one created ahead of playback. When playback on Jellyfin WebOS reached a mismatched segment, playback would stop."

Multiple users have tested this change with success, most are now running their own custom build with it added in. Credit to @fhelland for the investigation and reasoning. This will help to prevent an extremely frustrating issue, at least until a better alternative is found.

**Issues**

https://github.com/jellyfin/jellyfin-webos/issues/248

### DIFF
--- a/MediaBrowser.Controller/Streaming/StreamState.cs
+++ b/MediaBrowser.Controller/Streaming/StreamState.cs
@@ -93,6 +93,12 @@ public class StreamState : EncodingJobInfo, IDisposable
                     return 6;
                 }
 
+                //Prevent freezes on WebOS when skipping ahead during Dolby Vision playback. Issue 248 jellyfin/jellyfin-webos.
+                if (userAgent.Contains("Web0S", StringComparison.OrdinalIgnoreCase))
+                {
+                      return 1;
+                }
+
                 if (IsSegmentedLiveStream)
                 {
                     return 3;


gnattu (changes requested):
Sorry, but we no longer perform client-specific hacks like this on the server. The `SegmentLength` field is a parameter in the `TranscodingProfile`, which is specified by the client. Affected clients should explicitly specify this value instead of adding more hacks to the default settings on the server side. The current user agent hacks are implemented in the Emby time, and it’s too late to remove them as they would break existing clients. For new workarounds, specify `SegmentLength: 1` in the `TranscodingProfile` of the client and the server will respect that. Making this purely client side is also easier to control its behavior like implementing a client side feature switch in case it causes regression.

gnattu:
If you cannot find where to add, for webos's ts container use case, it is here:

https://github.com/jellyfin/jellyfin-web/blob/badd1dff41b76caf5b805b56e512a999e078fad9/src/scripts/browserDeviceProfile.js#L900-L910

You can add one line with `SegmentLength: browser.web0s ? 1 : undefined,` into the profile object which will have the same effect. If you want to do it better, you can also implement a client side option so that the users can turn this off if problems evolves without rebuilding the client and/or the server.

patrickd77-eng:
> If you cannot find where to add, for webos's ts container use case, it is here:
> 
> 
> 
> https://github.com/jellyfin/jellyfin-web/blob/badd1dff41b76caf5b805b56e512a999e078fad9/src/scripts/browserDeviceProfile.js#L900-L910
> 
> 
> 
> You can add one line with `SegmentLength: browser.web0s ? 1 : undefined,` into the profile object which will have the same effect and you can inject this for fmp4 profile above as well using the same method. If you want to do it better, you can also implement a client side option so that the users can turn this off if problems evolves without rebuilding the client and/or the server.


I completely understand, thanks for the detailed information, the alternative is definitely a better idea. I had been thinking turning it into a client side preference is a better idea, but wasn't sure on the repo's stance for this sort of thing. I'll take a look when I get some time and see how it can fit in.

lsvergau:
> > If you cannot find where to add, for webos's ts container use case, it is here:
> > https://github.com/jellyfin/jellyfin-web/blob/badd1dff41b76caf5b805b56e512a999e078fad9/src/scripts/browserDeviceProfile.js#L900-L910
> > You can add one line with `SegmentLength: browser.web0s ? 1 : undefined,` into the profile object which will have the same effect and you can inject this for fmp4 profile above as well using the same method. If you want to do it better, you can also implement a client side option so that the users can turn this off if problems evolves without rebuilding the client and/or the server.
> 
> I completely understand, thanks for the detailed information, the alternative is definitely a better idea. I had been thinking turning it into a client side preference is a better idea, but wasn't sure on the repo's stance for this sort of thing. I'll take a look when I get some time and see how it can fit in.

Hi, I'm experiencing the same issue, and would like to implement what gnattu is suggesting, however, I'm not experienced at all in this subject. I've been able to clone the WebOS repository but cannot find the file TranscodingProfile that is being referenced. I know the reference is made to the Web repository, which I don't fully understand as the issue is related to WebOS, I thought.

Thanks for any tips and help

patrickd77-eng:
I have made[ a new PR](https://github.com/jellyfin/jellyfin-web/pull/6530) on the relevant project based on your suggestions, so I think this PR can be closed for now. 

> If you cannot find where to add, for webos's ts container use case, it is here:
> 
> https://github.com/jellyfin/jellyfin-web/blob/badd1dff41b76caf5b805b56e512a999e078fad9/src/scripts/browserDeviceProfile.js#L900-L910
> 
> You can add one line with `SegmentLength: browser.web0s ? 1 : undefined,` into the profile object which will have the same effect and you can inject this for fmp4 profile above as well using the same method. If you want to do it better, you can also implement a client side option so that the users can turn this off if problems evolves without rebuilding the client and/or the server.

Issue #9636, opened 2023-04-13
[Issue]: Artists with Semicolon in name treated as multiple artists


FairlyBored:
### Please describe your bug

Artists with names containing a semicolon are treated as two artists. An example of an artist with a semicolon in the name is [We;Na](https://musicbrainz.org/artist/532996ca-2cf6-4c70-8b4e-7f8a99a2ac62)

For reference, my files(FLAC) are tagged with [Picard[(https://picard.musicbrainz.org/) and matched to their corresponding entries on MusicBrainz, and contain the appropriate MBIDs in addition to standard tags.

### Jellyfin Version

10.8.0

### if other:

10.8.9

### Environment

```markdown
- OS:Linux
- Virtualization:VirtualBox
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

AmbulantRex:
The `;`, `'|'`, `,`, and `/` characters are very common to use as delimiters. In the case of Artists, I believe this is the relevant code:
https://github.com/jellyfin/jellyfin/blob/c2b88f1852b16c2dde69b13c89feb34f78268057/MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs#L152-L162

Obviously, some artists are going to use our delimiters (e.g. AC/DC) so a whitelist appears to be available in this same file here:

https://github.com/jellyfin/jellyfin/blob/c2b88f1852b16c2dde69b13c89feb34f78268057/MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs#L45-L71

If this is the right spot, then it could be as simple as adding another artist to the list. A more robust solution may be warranted though.

AverageHelper:
I'm facing this issue as well with artist [リア;クオ](https://www.youtube.com/@liaquo) (lia;quo). Is there some escape character I should enter manually in the metadata editor i.e. `lia\;quo`? Or is the current solution to file a PR to add affected artist's names to [ProbeResultNormalizer.cs](https://github.com/1337joe/jellyfin/blob/927fe33d3a0ec7f9e0fb568cfd423c6e8b966c9d/MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs#L44) (as in #10329)?

gnattu:
[...]

PR #13562, opened 2025-02-15
Add リア;クオ or lia;quo to artist split whitelist

AverageHelper:
**Changes**
Adds another artist ([リア;クオ or lia;quo](https://www.youtube.com/watch?v=bfOOkjcKlw4)) to the artist split whitelist.

**Issues**
Related to #9636


gnattu:
This list is only kept for fallback and legacy reasons and should not be updated.

AverageHelper:
Ah, noted. Thanks!

Issue #13508, opened 2025-02-06
People show up multiple times in search results


theguymadmax:
### Description of the bug

On the master branch, search for a person. Observe that person is listed multiple times.

### Reproduction steps

1. On the master branch set up a new shows library. (default settings)
2. Add a shows library (replace all metadata after initial scan)
3. Search for an actor in the series

### What is the current _bug_ behavior?

The person displays multiple times in the search results

### What is the expected _correct_ behavior?

The person is displayed only once

### Jellyfin Server version

Master

### Specify commit id

00b66a06eac4d1331d59a6a9b7de3e99088ebaa9

### Specify the build version

2025020500

### Environment

```markdown
- OS: Windows 11
- Virtualization: No
```

### Jellyfin logs

```shell
none
```

### FFmpeg logs

```shell

```

### Relevant screenshots or videos

![Image](https://github.com/user-attachments/assets/ad3a70a8-0d48-4bb0-bd85-d92a01642ae6)

theguymadmax:
This issue is still present on the Master.

PR #13563, opened 2025-02-15
Prevent Duplicate Persons when Searching

Jxiced:
**Changes**
Use DistinctBy with IDs when displaying persons after searching.

**Issues**
#13508


gnattu:
I think this just hides the underlying query problem and is not a proper fix. The query should not return duplicated person.

PR #13583, opened 2025-02-19
Add profile condition to limit the number of streams

dmitrylyzo:
_Depends on https://github.com/jellyfin/jellyfin/pull/12678_

Some clients cannot play directly media with a large number of streams.

**Changes**
Add profile condition to limit the number of streams.

**Issues**
https://github.com/jellyfin/jellyfin-tizen/issues/289


Shadowghost (approved):

Bond-009 (approved):

MediaBrowser.Model/Dlna/ProfileConditionValue.cs
```diff
@@ -27,6 +27,7 @@ public enum ProfileConditionValue
         IsInterlaced = 21,
         AudioSampleRate = 22,
         AudioBitDepth = 23,
-        VideoRangeType = 24
+        VideoRangeType = 24,
+        NumStreams = 25
```
crobibero, line 31: I think `StreamCount` is better than `NumStreams`, especially since the error is `StreamCountExceedsLimit`

dmitrylyzo: We have `Num*Streams` for audio and video:
https://github.com/jellyfin/jellyfin/blob/869d884e54cdce7340377064007c513a672ec314/MediaBrowser.Model/Dlna/ProfileConditionValue.cs#L22-L23

crobibero (approved):

PR #13599, opened 2025-02-21
Provide MusicBrainz Recording Id.

yshui:
**Changes**
Add MusicBrainz Recording Id to ProviderIds.

Reference: https://picard-docs.musicbrainz.org/en/appendices/tag_mapping.html#id21

### DIFF
--- a/MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
+++ b/MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
@@ -400,6 +400,30 @@ namespace MediaBrowser.Providers.MediaInfo
                 }
             }
 
+            if (options.ReplaceAllMetadata || !audio.TryGetProviderId(MetadataProvider.MusicBrainzRecording, out _))
+            {
+                if (!track.AdditionalFields.TryGetValue("MUSICBRAINZ_TRACKID", out var musicBrainzRecordingIdTag)
+                     && !track.AdditionalFields.TryGetValue("MusicBrainz Recording Id", out musicBrainzRecordingIdTag)
+                     && track.AdditionalFields.TryGetValue("UFID", out musicBrainzRecordingIdTag))
+                {
+                    string[] parts = musicBrainzRecordingIdTag.Split("\0");
+                    if (parts[0] != "http://musicbrainz.org")
+                    {
+                        musicBrainzRecordingIdTag = null;
+                    }
+                    else
+                    {
+                        musicBrainzRecordingIdTag = parts[1];
+                    }
+                }
+
+                if (!string.IsNullOrEmpty(musicBrainzRecordingIdTag))
+                {
+                    var id = GetFirstMusicBrainzId(musicBrainzRecordingIdTag, libraryOptions.UseCustomTagDelimiters, libraryOptions.GetCustomTagDelimiters(), libraryOptions.DelimiterWhitelist);
+                    audio.TrySetProviderId(MetadataProvider.MusicBrainzRecording, id);
+                }
+            }
+
             // Save extracted lyrics if they exist,
             // and if the audio doesn't yet have lyrics.
             var lyrics = track.Lyrics.SynchronizedLyrics.Count > 0 ? track.Lyrics.FormatSynchToLRC() : track.Lyrics.UnsynchronizedLyrics;
--- a/MediaBrowser.Model/Entities/MetadataProvider.cs
+++ b/MediaBrowser.Model/Entities/MetadataProvider.cs
@@ -84,6 +84,11 @@ namespace MediaBrowser.Model.Entities
         /// <summary>
         /// The TvMaze provider.
         /// </summary>
-        TvMaze = 19
+        TvMaze = 19,
+
+        /// <summary>
+        /// The MusicBrainz recording provider.
+        /// </summary>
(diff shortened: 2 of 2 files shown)


MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
```diff
@@ -1342,6 +1342,10 @@ private void SetAudioInfoFromTags(MediaInfo audio, Dictionary<string, string> ta
             mb = GetMultipleMusicBrainzId(tags.GetValueOrDefault("MusicBrainz Release Track Id"))
                  ?? GetMultipleMusicBrainzId(tags.GetValueOrDefault("MUSICBRAINZ_RELEASETRACKID"));
             audio.TrySetProviderId(MetadataProvider.MusicBrainzTrack, mb);
+
+            mb = GetMultipleMusicBrainzId(tags.GetValueOrDefault("MusicBrainz Recording Id"))
```
IDisposable, line 1346: The preference order doesn't match between here and [the prober](https://github.com/jellyfin/jellyfin/pull/13599/files#diff-81f5576e4c1b35f0541e41ee4a0bda47b62121e3c2d2735f448bb6b7c12dcf88R405-R406)

yshui: This is how all the other tags are ordered (idk why).

gnattu: The order does not matter and they will not typically coexist. Unless you forcefully want to create a file to look for trouble. There are two maapping only because different file container has different tagging system and some of them uses the all capitals with underscore(like vorbis tags) but some use natural writing(like mp4 tags).

gnattu, line 1346: But, we don't add more functions to the ProbeResultNormalizer for music. This one is considered legacy and only being used for fallback for a minimal set of tags. The changes to this file should be removed.

gnattu (changes requested):

MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
yshui: Done!

yshui:
Removed changes in ProbeResultNormalizer, also add check for `UFID`. hopefully I am doing it right, C# isn't my native language.

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
```diff
@@ -400,6 +400,30 @@ private async Task FetchDataFromTags(Audio audio, Model.MediaInfo.MediaInfo medi
...
+            if (options.ReplaceAllMetadata || !audio.TryGetProviderId(MetadataProvider.MusicBrainzRecording, out _))
+            {
+                if (!track.AdditionalFields.TryGetValue("MUSICBRAINZ_TRACKID", out var musicBrainzRecordingIdTag)
+                     && !track.AdditionalFields.TryGetValue("MusicBrainz Recording Id", out musicBrainzRecordingIdTag)
+                     && track.AdditionalFields.TryGetValue("UFID", out musicBrainzRecordingIdTag))
+                {
+                    string[] parts = musicBrainzRecordingIdTag.Split("\0");
+                    if (parts[0] != "http://musicbrainz.org")
+                    {
+                        musicBrainzRecordingIdTag = null;
```
gnattu, line 412: I don't think this is entirely correct? It looks like UFID split by `\0` with the first part being musicbrainz url is only the ID3 specific value and on Vorbis tags and MP4 tags they are stored as-is without the url part, if I understand its mapping correctly. Can you check that if you have such files? All common FLAC files should be using Vorbis tags.

yshui: My understanding is `UFID` is id3v2 only, vorbis would be using `MUSICBRAINZ_TRACKID`. I don't know if ATL let you know if the tag is coming from id3v2?

gnattu: > My understanding is that UFID is id3v2 only

That is exactly the problem. Only for id3v2 UFID tag you have to perform the split, and for other tagging systems the tagging value is written as is. Your current logic will simply put everything not tagged like UFID to null, isn't it?

> I don't know if ATL let you know if the tag is coming from id3v2?

It does but we should not care in this case. We just check the UFID case separately from the Voribs/MP4 tag so that they don't overwrite their values each other.

To simplify code we first check UFID, pre-process it by doing the splitting and get the second part, and then use the Vorbis/MP4 logic, if that UFID extracted value is empty.

gnattu:
Wait, isn't this duplicate of https://github.com/jellyfin/jellyfin/pull/12173?

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
yshui: > not tagged like UFID to null, isn't it?

I don't think it would? `MUSICBRAINZ_TRACKID` should be handled fine?

Bond-009:
Closing as duplicate of #12173

PR #13605, opened 2025-02-22
Implement limiting caches

crobibero:
Simple POC to see what shakes out

I looked at [FusionCache](https://github.com/ZiggyCreatures/FusionCache) also but that library seems to be way more functionality then we can use at the moment.


MediaBrowser.Model/Configuration/ServerConfiguration.cs
```diff
@@ -177,6 +177,11 @@ public ServerConfiguration()
     /// <value>The library update duration.</value>
     public int LibraryUpdateDuration { get; set; } = 30;
 
+    /// <summary>
+    /// Gets or sets the amount of items to cache.
+    /// </summary>
+    public int LibraryCacheSize { get; set; } = 10_000;
```
crobibero, line 183: I have 161k in my "TypedBaseItems" table, so I just picked 10k for now.

Emby.Server.Implementations/Library/LibraryManager.cs
```diff
@@ -145,14 +146,18 @@ public LibraryManager(
             _mediaEncoder = mediaEncoder;
             _itemRepository = itemRepository;
             _imageProcessor = imageProcessor;
-            _cache = new ConcurrentDictionary<Guid, BaseItem>();
+            _cache = new ConcurrentLruBuilder<Guid, BaseItem>()
```
JPVenson, line 149: Please provide that via DI instead

JPVenson (changes requested):

Emby.Server.Implementations/Library/LibraryManager.cs
cvium: I disagree. Not everything needs to be resolved from DI. This is a local cache

JPVenson: i see no reasons why this could not be shared between other components and/or maybe even used for a 2nd level database cache

cvium: What is the benefit of sharing it? If another component needs a cache they can just new up their own?

IDisposable:
Since we're targeting .NetCore 9.x, perhaps we should consider using [HybridCache](https://learn.microsoft.com/en-us/aspnet/core/performance/caching/hybrid?view=aspnetcore-9.0) with or without a `IDistributedCache` injected to services.

crobibero:
> Since we're targeting .NetCore 9.x, perhaps we should consider using [HybridCache](https://learn.microsoft.com/en-us/aspnet/core/performance/caching/hybrid?view=aspnetcore-9.0) with or without a `IDistributedCache` injected to services.

There is no benefit for Jellyfin to use a HybridCache since we don't need L1/L2 caching or any of the other advanced features. As mentioned in the PR body I also looked into FusionCache but that also has way more functionality then we need

Emby.Server.Implementations/Library/LibraryManager.cs
crobibero: I tried adding the second level cache interceptor back in and performance was much worse. I don't see any place that we could use a shared cache right now

IDisposable:
Understood, the one nice feature that HybridCache brings is stampede avoidance. That said, simpler is good

crobibero:
> Understood, the one nice feature that HybridCache brings is stampede avoidance. That said, simpler is good

If we end up needing to mitigate cache stampede the library that was added in this PR has https://github.com/bitfaster/BitFaster.Caching/wiki/Atomic-GetOrAdd

Though we've been running with just a ConcurrentDictionary for a while now so I don't think we will run into that issue

MediaBrowser.Model/Configuration/ServerConfiguration.cs
IDisposable: Correct my assumption if wrong, but this gets used in the `.WithCapacity` in _LibraryManager.cs_ and in _UserLibraryManager.cs_ so we could have 10K of each, correct? Also assuming that the caches in _DirectoryService.cs_ are not capacity-constrained... how many will they cache?

Emby.Server.Implementations/Library/LibraryManager.cs
JPVenson: Fair enough.
Thought it might just give us _the option_ to reuse it later if its already there.

JPVenson (approved):

MediaBrowser.Model/Configuration/ServerConfiguration.cs
crobibero: 🤷 as many as it currently does. The only change there was to make the cache entries expire after 10 minutes of no access

My thought process is that the DirectoryService cache is hit the hardest during a library scan so increased memory is expected, but after scan it should drop back down

crobibero: If a cache isn't initialized with a size it is set to 128, which is much too small for our use case.

crobibero:
I did some further reading and discovered that the caching library doesn't eagerly expire cached items, only when items are added. We could manually call `.TrimExpired()` but I think it would just be best to set a cap on the cache size.

I went back and forth on ConcurrentLru vs ConcurrentLfu (https://github.com/bitfaster/BitFaster.Caching/wiki/Performance) and ended up with ConcurrentLru because Jellyfin is targeted towards non production server hardware, and also not millions of items.

I also went with FastConcurrentLru instead of ConcurrentLru because we aren't monitoring hit count (https://github.com/bitfaster/BitFaster.Caching/wiki/What-are-the-'fast'-LRU-variants%3F)

crobibero:
I navigated around my library using the webui and it took quite a bit of clicking around to grow the cache to 200.. so 10k would be for many concurrent users

crobibero:
Tests are currently failing because `DirectoryService.Initialize()` isn't being called, but that's another day problem

JPVenson (approved):

Bond-009 (approved):

PR #13613, opened 2025-02-24
Change BaseItemEntity.PrimaryVersionId type to Guid?

Bond-009:
This is just another item's id, so always a valid Guid.


Jellyfin.Server.Implementations/Migrations/20250224184813_PrimaryVersionIdGuid.cs
```diff
@@ -0,0 +1,22 @@
...
+#nullable disable
+
+namespace Jellyfin.Server.Implementations.Migrations
+{
+    /// <inheritdoc />
+    public partial class PrimaryVersionIdGuid : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
```
JPVenson, line 12: if that is actually supposed to be a reference to another BaseItem, please introduce an forginKey

JPVenson: can you elaborate on that a bit? So PrimaryVersionId will still be stored as text in the DB and it references another BaseItem where the key it references is stored in the Data column?

If thats the case, doing it now like you did is fine with me, but i then would like to see an extensive documentation in the code comment nex to the PrimaryVersionId property https://github.com/jellyfin/jellyfin/pull/13613/files#diff-cecd04310e3c734033ce9d1f777919a2c1ffdd45ca321a3c32d0386110a17b47R99 explaining why our Entity does not match the DB scheme exactly

Bond-009:
Superseded by #16062

PR #13615, opened 2025-02-25
Support custom resolvers. 

Aeyk:
**Changes**
Allow additional scanners to read metadata from the file system structure of folders during a library scan. This should help support [custom resolvers](https://github.com/jellyfin/jellyfin/discussions/5732) as plugins. 

**Issues**
[5732](https://github.com/jellyfin/jellyfin/discussions/5732)

### DIFF
--- a/MediaBrowser.Controller/Resolvers/Scanner.cs
+++ b/MediaBrowser.Controller/Resolvers/Scanner.cs
@@ -22,7 +22,7 @@ namespace MediaBrowser.Controller.Resolvers
 
         protected BaseItem ApplyMetadata(BaseItem t)
         {
-            throw new NotImplementedException();
+            protected abstract BaseItem ApplyMetadata(BaseItem t);
         }
 
         public ICollection<BaseItem> ApplyMetadata(ICollection<BaseItem> ts)
--- a/MediaBrowser.Controller/Resolvers/Scanner.cs
+++ b/MediaBrowser.Controller/Resolvers/Scanner.cs
@@ -1,0 +1,38 @@
+#pragma warning disable CS1591
+
+using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Linq;
+using Jellyfin.Data.Enums;
+using MediaBrowser.Controller.Entities;
+using MediaBrowser.Controller.Library;
+using MediaBrowser.Controller.Providers;
+using MediaBrowser.Model.IO;
+
+namespace MediaBrowser.Controller.Resolvers
+{
+    public abstract class Scanner : IScanner
+    {
+        public bool Enabled { get; set; } = false;
+
+        public bool Default { get; set; } = false;
+
+        public ResolverPriority Priority => ResolverPriority.Plugin;
+
+        protected BaseItem ApplyMetadata(BaseItem t)
+        {
+            throw new NotImplementedException();
+        }
+
+        public ICollection<BaseItem> ApplyMetadata(ICollection<BaseItem> ts)
+        {
+            foreach (var t in ts)
+            {
+                ApplyMetadata(t);
+            }
+
+            return ts;
+        }
+    }
+}
--- a/MediaBrowser.Providers/Manager/ProviderManager.cs
+++ b/MediaBrowser.Providers/Manager/ProviderManager.cs
@@ -591,6 +591,14 @@ namespace MediaBrowser.Providers.Manager
                 Type = MetadataPluginType.MediaSegmentProvider
             }));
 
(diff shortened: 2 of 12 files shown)


MediaBrowser.Controller/Resolvers/Scanner.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        public bool Enabled { get; set; } = false;
+
+        public bool Default { get; set; } = false;
+
+        public ResolverPriority Priority => ResolverPriority.Plugin;
+
+        protected BaseItem ApplyMetadata(BaseItem t)
+        {
+            throw new NotImplementedException();
```
IDisposable, line 25: This would prevent safe use of calling `t = base.ApplyMetadata(t);` in multiple-layer derived resolver classes. Do we explicitly want to discourage that sort of coding?

Perhaps better to just make it this 
```c#
return t;
```

MediaBrowser.Providers/Manager/ProviderManager.cs
```diff
@@ -611,10 +619,10 @@ private void AddMetadataPlugins<T>(List<MetadataPlugin> list, T item, LibraryOpt
             var providers = GetMetadataProvidersInternal<T>(item, libraryOptions, options, true, true).ToList();
 
             // Locals
-            list.AddRange(providers.Where(i => i is ILocalMetadataProvider).Select(i => new MetadataPlugin
+            list.AddRange(providers.Where(i => i is IMetadataProvider && i is not IRemoteMetadataProvider && i is not ILocalMetadataProvider && i is not ICustomMetadataProvider).Select(i => new MetadataPlugin
```
IDisposable, line 622: Doesn't this kill off the `ILocalMetadataProvider` derived classes that the old code explicitly included?

Also seems very fragile to test for a (limited) set of `is not` instead of explicitly listing the marker interfaces we _want_ to include.

MediaBrowser.Providers/Manager/ProviderManager.cs
```diff
@@ -591,6 +591,14 @@ private MetadataPluginSummary GetPluginSummary<T>()
                 Type = MetadataPluginType.MediaSegmentProvider
             }));
 
+            // Filesystem to metadata resolvers
+            var metadataResolverProviders = summary.Plugins.Select(p => p.Type == MetadataPluginType.MetadataResolver && p.Name != null ? p : null);
+            pluginList.AddRange(metadataResolverProviders.Select(i => new MetadataPlugin
+            {
+                Name = i?.Name,
+                Type = MetadataPluginType.MetadataResolver
+            }));
```
IDisposable, line 600: ... or here add a  `.Where(i => i.Name != null);` (inside last parathesis)?

MediaBrowser.Providers/Manager/ProviderManager.cs
```diff
@@ -591,6 +591,14 @@ private MetadataPluginSummary GetPluginSummary<T>()
                 Type = MetadataPluginType.MediaSegmentProvider
             }));
 
+            // Filesystem to metadata resolvers
+            var metadataResolverProviders = summary.Plugins.Select(p => p.Type == MetadataPluginType.MetadataResolver && p.Name != null ? p : null);
```
IDisposable, line 595: Do we want to tack on a `.Where(i != null)` here?

MediaBrowser.Controller/Resolvers/Scanner.cs
JPVenson: yes.

this is the base function/only function that actually does something here. There is no other reason to implement this base class, so why not just make this function abstract itself?

```suggestion
            protected abstract BaseItem ApplyMetadata(BaseItem t);
```

MediaBrowser.Providers/Manager/ProviderManager.cs
JPVenson: how _could_ they ever be null?

IDisposable: I was referring to the adding onto the block `.Select(p => p.Type == MetadataPluginType.MetadataResolver && p.Name != null ? p : null)`, where what comes out of the `.Select` could be null if the `.Type` is not `MetadataPluginType.MetadataResolver` OR if it has a `null` setting on the `.Name`... in those cases we would get a `null` out of the ternary... so maybe we should filter/skip those?

`.Select(p => p.Type == MetadataPluginType.MetadataResolver && p.Name != null ? p : null).Where(n => n != null);`

Bond-009:
Closing as stale, no response to the review for 1+ year.

PR #13660, opened 2025-03-04
Use package name to replace "jellyfin" in temp directory if set

Inuyasha-cxk:
When deploying dual Jellyfin instances on the same machine with different user accounts, the hardcoded temporary directory path (/tmp/jellyfin) causes file permission conflicts. Existing configuration options do not expose temp directory customization.

Introduce dynamic temp directory resolution:
- If package-name(command line arguments --package-name) is specified in configuration, use /tmp/{package-name}
- Fallback to /tmp/jellyfin when unconfigured
The temp directory like:
/tmp/jellyfin-main  #  --package-name=jellyfin-main
/tmp/jellyfin-test  #  --package-name=jellyfin-test
It works on my debian 12


Inuyasha-cxk:
> No, `Path.GetTempPath` method is not hardcoded to `/tmp` and the actual path is platform dependent. You can use `TMPDIR` env var to overwrite the system default which means we don't need a complicated solution like this.

Ah, I overlooked that user-specific environment variables allow setting a temp directory individually for each user. I focused solely on modifying the 'jellyfin' part and missed this critical point. My approach was indeed overcomplicated. 
I’ll close this pull request.

PR #13277, opened 2024-12-23
Improve dynamic HDR metadata handling

gnattu:
**Changes**

This is a relatively big change to the way advanced HDR formats are handled. Previously, we were only checking HDR10, HLG, and Dolby Vision files, with a naive mapping of fallbacks. While this approach was somewhat effective, we continue to encounter compatibility issues. This PR addresses these issues by:

- Properly detecting HDR10+ videos and assigning the corresponding `VideoRangeType`.
- Introducing additional Dolby Vision `VideoRangeType` variants to address common compatibility issues, such as Profile 7 videos, videos with invalid or out-of-specification profile values, and Dolby Vision HDR10+ hybrid videos.
- Implementing a functionality to remove dynamic HDR metadata (Dolby Vision or HDR10+) when necessary to mitigate compatibility issues from the client side.

Metadata rescan on the already indexed problematic files would be necessary.

~~Current implementation is abusing the EL flag field to avoid changing the database schema. Should add a proper field once #12798 is merged.~~ Done

**Issues**

Depends on https://github.com/jellyfin/jellyfin-ffmpeg/pull/519

Fixes jellyfin/jellyfin-androidtv#4021
Fixes jellyfin/jellyfin-androidtv#2700
Fixes jellyfin/jellyfin-androidtv#2630
Fixes #11275

Closes feature request: https://features.jellyfin.org/posts/2557/support-for-playing-dv-profile-7-hybrid-dv-hdr-files-on-non-dv-hdr-device

PR #13665, opened 2025-03-05
DoVi Prof5 is only one that should use dvh1 tag

timitt:
Only DoVi Profile 5 should have dvh1 tag as for example Profile 8 can fallback to HDR10 and therefore does not need to understand dvh1 tag.

Idea is that others have fallbacks to non-DoVi mode and for example Profile 8 fallbacks to HDR10 video correctly if DoVi is not supported. However it might not do that if it can't recognize tag.

As an example hls.js fails to play Profile 8 video as HDR if tag is dvh1, but plays it as HDR if tah is hvc1.

Also the change does not seem to break native hls on WebOS. It might be the case that most implementations do not read the tag at all but find hvcC, dvcC or dvvC configuration boxes and choose by them.

I can't confirm if this breaks playback on some hardware (as I could only test WebOS native hls and hls.js on WebOS), but this should be the correct way to use tag.

Without other fixes this only helps WebOS with hls.js play videos that has DoVi Profile 8 and aac audio. I will make more pull requests later to jellyfin-web. I believe this pull request is valid alone too.

**Changes**
Check if DoVi video is Profile 5 and then only change that video tag to dvh1 and let other types have hvc1.
We still need -strict -2 for both of them. I also changed code a bit to not repeat that -strict -2 line part three times.

**Issues**
Relates, but does not directly fix: https://github.com/jellyfin/jellyfin-webos/issues/275

### DIFF
--- a/Jellyfin.Api/Controllers/DynamicHlsController.cs
+++ b/Jellyfin.Api/Controllers/DynamicHlsController.cs
@@ -1823,19 +1823,33 @@ public class DynamicHlsController : BaseJellyfinApiController
             // Only enable Dolby Vision remuxing if the client explicitly declares support for profiles without fallbacks.
             var clientSupportsDoVi = requestedRange.Contains(VideoRangeType.DOVI.ToString(), StringComparison.OrdinalIgnoreCase);
             var videoIsDoVi = state.VideoStream.VideoRangeType is VideoRangeType.DOVI or VideoRangeType.DOVIWithHDR10 or VideoRangeType.DOVIWithHLG or VideoRangeType.DOVIWithSDR;
+            var videoIsDoviProfile5 = state.VideoStream.DvProfile == 5;
 
             if (EncodingHelper.IsCopyCodec(codec)
                 && (videoIsDoVi && clientSupportsDoVi))
             {
                 if (isActualOutputVideoCodecHevc)
                 {
-                    // Prefer dvh1 to dvhe
-                    args += " -tag:v:0 dvh1 -strict -2";
+                    // Profile 5 does not have any fallbacks so it should have dvh1 tag.
+                    // Other DoVi videos generally have fallbacks so they should have generic hvc1 tag.
+                    if (videoIsDoviProfile5)
+                    {
+                        // Prefer dvh1 to dvhe
+                        args += " -tag:v:0 dvh1";
+                    }
+                    else
+                    {
+                        // Prefer hvc1 to hev1
+                        args += " -tag:v:0 hvc1";
+                    }
                 }
                 else if (isActualOutputVideoCodecAv1)
                 {
-                    args += " -tag:v:0 dav1 -strict -2";
+                    args += " -tag:v:0 dav1";
                 }
+
+                // All DoVi videos should have strict -2 (experimental) flag.
+                args += " -strict -2";
             }
             else if (isActualOutputVideoCodecHevc)
             {


Jellyfin.Api/Controllers/DynamicHlsController.cs
```diff
@@ -1823,19 +1823,33 @@ private string GetVideoArguments(StreamState state, int startNumber, bool isEven
             // Only enable Dolby Vision remuxing if the client explicitly declares support for profiles without fallbacks.
             var clientSupportsDoVi = requestedRange.Contains(VideoRangeType.DOVI.ToString(), StringComparison.OrdinalIgnoreCase);
             var videoIsDoVi = state.VideoStream.VideoRangeType is VideoRangeType.DOVI or VideoRangeType.DOVIWithHDR10 or VideoRangeType.DOVIWithHLG or VideoRangeType.DOVIWithSDR;
+            var videoIsDoviProfile5 = state.VideoStream.DvProfile == 5;
```
gnattu, line 1826: Use `VideoRangeType.DOVI` with the HEVC codec check. Don’t do profile check directly.

Jellyfin.Api/Controllers/DynamicHlsController.cs
```diff
@@ -1823,19 +1823,33 @@ private string GetVideoArguments(StreamState state, int startNumber, bool isEven
...
+                    else
+                    {
+                        // Prefer hvc1 to hev1
+                        args += " -tag:v:0 hvc1";
+                    }
                 }
                 else if (isActualOutputVideoCodecAv1)
                 {
-                    args += " -tag:v:0 dav1 -strict -2";
+                    args += " -tag:v:0 dav1";
```
gnattu, line 1848: Actually av1 outputs should do the same. We only need to apply dav1 with range type without any kind of fallbacks.

gnattu (changes requested):
This will conflict with #13277. If we want to make this change we need to apply such change there because the dynamic HDR metadata handling is changed dramatically in that PR and the checks done here will be not enough. The old force dvh1 tag is to force the Dolby Vision native HLS players to play in Dolby Vision mode. But I *think* we should be fine as we already supplied `SUPPLEMENTAL-CODECS` in the hls playlist which should make the complaint players be aware of the Dolby Vision metadata. But I don’t know the webOS implementation there as its implementation is always problematic. The implementation on Apple devices supports either way, as long as the metadata is valid.

Jellyfin.Api/Controllers/DynamicHlsController.cs
timitt: I am not sure what you mean. I did not change av1 logic as -strict -2 is added to it too. Or do you mean that it should not have -strict -2?

Jellyfin.Api/Controllers/DynamicHlsController.cs
timitt: Yes, I only now realised the full logic about those VideoRangeType values.

timitt:
I am not sure if putting dvh1 tag there would force native hls there anyway. If it does then it should only be used with the certain hls implementation as using it will break other cases (like for example hls.js implementation).

Should I modify this pull request to use VideoRangeType.DOVI or is it pointless because of that other pull request?

Jellyfin.Api/Controllers/DynamicHlsController.cs
gnattu: Dolby Vision Profile 10 (for av1) has both non-fallback profile level and with fallback profile level which means the file should be tagged as av01 instead of dav1 for the profile levels have fallbacks.

gnattu:
> I am not sure if putting dvh1 tag there would force native hls there anyway. If it does then it should only be used with the certain hls implementation as using it will break other cases (like for example hls.js implementation).
> 
> Should I modify this pull request to use VideoRangeType.DOVI or is it pointless because of that other pull request?

This is a bit pointless if targeting master as the logic here will be replaced by that PR anyway.

Jellyfin.Api/Controllers/DynamicHlsController.cs
timitt: Oh yes. I understand. I had no test videos for that so did not want to touch it.

timitt:
Not sure if there is point to offer it to releae branch either if it is intentional to use "wrong" tag. But if you think it makes sense I can make pull request to release branch.

timitt:
I won't do pull request to jellyfin-web side either for now. Especially when I actually found a way for hls.js and WebOS pair to play ac-3 and aec-3 too. That just needs more work on server side that also don't make sense to be made before https://github.com/jellyfin/jellyfin/pull/13277 is pulled.

PR #13679, opened 2025-03-08
Add fast-path to getting just the SeriesPresentationUniqueKey for NextUp

crobibero:

This doesn't actually fix the issue, but reduces the execution time with my prod database (~8-12s to ~3-5s) since we don't need to hydrate every single item

### DIFF
--- a/Emby.Server.Implementations/TV/TVSeriesManager.cs
+++ b/Emby.Server.Implementations/TV/TVSeriesManager.cs
@@ -66,7 +66,7 @@ namespace Emby.Server.Implementations.TV
             else
             {
                 parents = _libraryManager.GetUserRootFolder().GetChildren(user, true)
-                   .Where(i => i is Folder)
+                   .Where(i => i is CollectionFolder f && f.CollectionType == CollectionType.tvshows)
                    .Where(i => !user.GetPreferenceValues<Guid>(PreferenceKind.LatestItemExcludes).Contains(i.Id))
                    .ToArray();
             }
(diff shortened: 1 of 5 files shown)


Emby.Server.Implementations/TV/TVSeriesManager.cs
```diff
@@ -66,7 +66,7 @@ public QueryResult<BaseItem> GetNextUp(NextUpQuery query, DtoOptions options)
             else
             {
                 parents = _libraryManager.GetUserRootFolder().GetChildren(user, true)
-                   .Where(i => i is Folder)
+                   .Where(i => i is CollectionFolder f && f.CollectionType == CollectionType.tvshows)
```
crobibero, line 69: Can "mixed movies and tv" show up in next up?
If so this collection type filter should include those libraries

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -255,6 +255,23 @@ public IReadOnlyList<BaseItemDto> GetItemList(InternalItemsQuery filter)
...
+
+        using var context = _dbProvider.CreateDbContext();
+        IQueryable<BaseItemEntity> dbQuery = PrepareItemQuery(context, filter);
+
+        dbQuery = TranslateQuery(dbQuery, context, filter);
+
+        dbQuery = ApplyGroupingFilter(dbQuery, filter);
+        dbQuery = ApplyQueryPaging(dbQuery, filter);
+
+        return dbQuery.Where(q => q.SeriesPresentationUniqueKey != null).Select(q => q.SeriesPresentationUniqueKey!).AsEnumerable().Distinct().ToArray();
```
cvium, line 272: Won't `AsEnumerable().Distinct()` cause it to do the filtering client-side rather than in SQL? It shouldn't be necessary to even have AsEnumerable, I think.
```suggestion
        return dbQuery.Where(q => q.SeriesPresentationUniqueKey != null).Select(q => q.SeriesPresentationUniqueKey!).Distinct().ToArray();
```

crobibero: Doing the distinct call in the db breaks ordering (see ApplyGroupingFiltering)

PR #5026, opened 2021-01-15
Revert "Don't return first episodes in next up"

nielsvanvelzen:
Reverts jellyfin/jellyfin#4595

Unfortunately the change caused issues in the Android TV app. The "play next up" button stopped working since it used an endpoint that called this function and relied on it returning first episodes.

A proper fix is to add a new query parameter to the controllers so we can change the behavior from the client. It should default to the old behavior.

PR #4595, opened 2020-11-27
Don't return first episodes in next up

MrTimscampi:
**Changes**

This removes the weird behavior where GetNextUp will return first episodes of shows if there is no series in progress.

The rationale is that we shouldn't assume something is next up in queue if it hasn't been started yet.

No changes needed in jellyfin-web, since the section will hide if empty.

**Issues**

PR #13687, opened 2025-03-10
Add fast-path to getting just the SeriesPresentationUniqueKey for NextUp

crobibero:

Alternate to https://github.com/jellyfin/jellyfin/pull/13679

This change makes the next up call take ~1.5s (0.1s in get series keys)

It's possible that this query is missing some of the functionality but it returns the correct set of next up series for the account I tested with

### DIFF
--- a/Emby.Server.Implementations/Library/LibraryManager.cs
+++ b/Emby.Server.Implementations/Library/LibraryManager.cs
@@ -1344,7 +1344,7 @@ namespace Emby.Server.Implementations.Library
             return _itemRepository.GetItemList(query);
         }
 
-        public IReadOnlyList<string> GetNextUpSeriesKeys(InternalItemsQuery query, IReadOnlyCollection<BaseItem> parents)
+        public IReadOnlyList<string> GetNextUpSeriesKeys(InternalItemsQuery query, IReadOnlyCollection<BaseItem> parents, DateTime dateCutoff)
         {
             SetTopParentIdsOrAncestors(query, parents);
 
@@ -1356,7 +1356,7 @@ namespace Emby.Server.Implementations.Library
                 }
             }
 
-            return _itemRepository.GetNextUpSeriesKeys(query);
+            return _itemRepository.GetNextUpSeriesKeys(query, dateCutoff);
         }
 
         public QueryResult<BaseItem> QueryItems(InternalItemsQuery query)
--- a/Emby.Server.Implementations/Library/LibraryManager.cs
+++ b/Emby.Server.Implementations/Library/LibraryManager.cs
@@ -1344,6 +1344,21 @@ namespace Emby.Server.Implementations.Library
             return _itemRepository.GetItemList(query);
         }
 
+        public IReadOnlyList<string> GetNextUpSeriesKeys(InternalItemsQuery query, IReadOnlyCollection<BaseItem> parents)
+        {
+            SetTopParentIdsOrAncestors(query, parents);
+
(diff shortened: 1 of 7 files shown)


JPVenson (approved):
i hate this query.
Approved

Bond-009 (approved):

crobibero:
This can be improved further by filtering series by `NextUpDateCutoff`
Will update this PR with that change

crobibero:
With a date filter of `2024-03-11` the request time goes from 1.7s to ~500ms
This has a side-effect of not returning random first episodes for shows, which I always thought was weird anyway.

I recommend clients start supporting `nextUpDateCutoff` to make the NextUp call nice and quick

JPVenson (approved):

cvium:
> This has a side-effect of not returning random first episodes for shows, which I always thought was weird anyway.

What do you mean by this? Does this affect the "Show all Episode 1's if nothing is watched"?

crobibero:
> > This has a side-effect of not returning random first episodes for shows, which I always thought was weird anyway.
> 
> What do you mean by this? Does this affect the "Show all Episode 1's if nothing is watched"?

That's the exact thing I removed. I looked through the web client and didn't see an option to actually change this functionality?

If it is still desired I guess we can add a path where if nothing is watched, return x random episode 1s? Showing "all" is what makes this request 10+ seconds

cvium:
> > > This has a side-effect of not returning random first episodes for shows, which I always thought was weird anyway.
> > 
> > 
> > What do you mean by this? Does this affect the "Show all Episode 1's if nothing is watched"?
> 
> That's the exact thing I removed. I looked through the web client and didn't see an option to actually change this functionality?
> 
> If it is still desired I guess we can add a path where if nothing is watched, return x random episode 1s? Showing "all" is what makes this request 10+ seconds

Regression lifecycles are circular. Are we sure that removing this [again](https://github.com/jellyfin/jellyfin/pull/4595) won't cause [this](https://github.com/jellyfin/jellyfin/pull/5026) again?

crobibero:
> > > > This has a side-effect of not returning random first episodes for shows, which I always thought was weird anyway.
> > > 
> > > 
> > > What do you mean by this? Does this affect the "Show all Episode 1's if nothing is watched"?
> > 
> > 
> > That's the exact thing I removed. I looked through the web client and didn't see an option to actually change this functionality?
> > If it is still desired I guess we can add a path where if nothing is watched, return x random episode 1s? Showing "all" is what makes this request 10+ seconds
> 
> Regression lifecycles are circular. Are we sure that removing this [again](https://github.com/jellyfin/jellyfin/pull/4595) won't cause [this again](https://github.com/jellyfin/jellyfin/pull/5026)?

The AndroidTV client no longer uses that endpoint so I'm not worried about that client breaking

I've confirmed that NextUp with a series id specified returns S01E01 if no episodes have been watched, so I think this is as good as it will get for now

Bond-009 (approved):

soultaco83:
I installed this pr to see if this will fix the next up being slow. Which it does 
But I am seeing a lot of season 1 episode 1 videos appearing in my list now
I see some talks about it. But just wanted to note what I see as well.

markup_3460

chenks:
hi

will these changes allow "next up" to be updated without having to perform a "scan all libraries"?
at the moment new episodes are picked by real time monitoring and added to "recently added", but they won't show up in "next up" until a full library scan is completed.

PR #12401, opened 2024-08-06
Bugfix/gh issue refactoring

JPVenson:
Updated several instances of array usage to IReadOnlyList 

**Changes**
Functionally the code should behave the same apart from one functional change in the `GuideManager.PrecacheImages` method

PR #13693, opened 2025-03-10
Rebase #12401 refactoring and resolve merge conflicts

richlander:
I wanted to help by moving some old PRs along. Opening a new PR with a fresh rebase seemed to be the fastest path and will produce the cleanest history. So far, this is <30mins of effort. I wanted to validate that this approach is acceptable before continuing. I don't know what the acceptable norms are for this project.

Done:

- Rebase of #12401 with `master`
- Resolved merge conflicts
- Fixed build breaks

Remaining:

- One test is failing, at the following line.

https://github.com/jellyfin/jellyfin/blob/de5b6470beef2497d268de15c9c91f57f6203d0f/tests/Jellyfin.Providers.Tests/MediaInfo/MediaInfoResolverTests.cs#L325

@JPVenson

### DIFF
--- a/Emby.Naming/AudioBook/AudioBookNameParserResult.cs
+++ b/Emby.Naming/AudioBook/AudioBookNameParserResult.cs
@@ -17,17 +17,27 @@ namespace Emby.Naming.AudioBook
         /// </summary>
         public int? Year { get; set; }
 
-#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
+        /// <summary>
+        /// Comparison equallity operators for <see cref="AudioBookNameParserResult"/>.
+        /// </summary>
+        /// <param name="left">Source comparison object.</param>
+        /// <param name="right">Target comparision object.</param>
+        /// <returns>If both objects are the equal.</returns>
         public static bool operator ==(AudioBookNameParserResult left, AudioBookNameParserResult right)
         {
             return left.Equals(right);
         }
 
+        /// <summary>
+        /// Comparison unequallity operators for <see cref="AudioBookNameParserResult"/>.
+        /// </summary>
+        /// <param name="left">Source comparison object.</param>
+        /// <param name="right">Target comparision object.</param>
+        /// <returns>If both objects are the equal.</returns>
         public static bool operator !=(AudioBookNameParserResult left, AudioBookNameParserResult right)
         {
             return !(left == right);
         }
-#pragma warning restore CS1591 // Missing XML comment for publicly visible type or member
 
         /// <inheritdoc/>
         public override bool Equals(object? obj)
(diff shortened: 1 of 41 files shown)


Emby.Naming/AudioBook/AudioBookNameParserResult.cs
```diff
@@ -14,5 +16,50 @@ public struct AudioBookNameParserResult
         /// Gets or sets optional year of release.
         /// </summary>
         public int? Year { get; set; }
+
+        /// <summary>
```
JPVenson, line 20: i am generally not a huge fan of overloading operators if you dont have a very good reason for it

MediaBrowser.Controller/Entities/TV/Series.cs
```diff
@@ -30,7 +30,7 @@ public Series()
             AirDays = Array.Empty<DayOfWeek>();
         }
 
-        public DayOfWeek[] AirDays { get; set; }
+        public IReadOnlyList<DayOfWeek> AirDays { get; set; }
```
JPVenson, line 33: this could potentially cause issues with serialisation

MediaBrowser.Model/Dto/BaseItemDto.cs
```diff
@@ -111,21 +111,21 @@ public class BaseItemDto : IHasProviderIds, IItemDto, IHasServerId
         /// Gets or sets the external urls.
         /// </summary>
         /// <value>The external urls.</value>
-        public ExternalUrl[] ExternalUrls { get; set; }
+        public IReadOnlyList<ExternalUrl> ExternalUrls { get; set; }
```
JPVenson, line 114: same here. Those fields get serialised in the backend and AFAIK the json serializer cannot handle those interfaces well

JPVenson (changes requested):

Emby.Naming/AudioBook/AudioBookNameParserResult.cs
richlander: I'm confused. This is from your commit: https://github.com/jellyfin/jellyfin/pull/12401/files#diff-ae9b100c09870f12039b99d5fa71e295a6ba55ac78e363b54270bdd58f8c9f51. I have no opinion on it. I was solely trying to help to move your PR forward. If you no longer want to move forward with those changes, we can certainly close both. I'll look for another stale PR to try to land.

Here are the changes I made (to fix build breaks, after resolving the merge conflicts): https://github.com/jellyfin/jellyfin/pull/13693/commits/099e994cae1d8127fbe30a0889dc8223b01f8335

JPVenson: you are not wrong.
But that Pr was quite a long time ago and was in all honesty a bit low effort to begin with.

PR #13726, opened 2025-03-17
Enable segment deletion for livestreams that are transcoded or remuxed

### LINKED ISSUE #13725: Delete Segments not working for livestreams that are transcoded or remuxed
### Description of the bug

Transcoded cache files are stored for as long as the stream is running even though delete segments is enabled and the time to keep segments is set for example to 300 seconds.
My cache easily grows to more than 40GB if several people are watching.

### Reproduction steps

1. Watch a livestream that needs remuxing or transcoding
2. Monitor the cache folder size

Segments are not deleted.

Segments should be deleted.

### Jellyfin Server version

10.10.0+

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.10.3

### Environment

```markdown
- OS: Windows 11
- Linux Kernel:
- Virtualization:
- Clients: Browser, Android, Fire Stick
- Browser: Chrome 131
- FFmpeg Version: 7.0
- Playback Method: Transcode, Remux
- Hardware Acceleration: Nvenc
- GPU Model: Nvidia Geforce GTX 1660 Ti
- Plugins:
- Reverse Proxy:
- Base URL: 
- Networking:
- Storage:
```

### Jellyfin logs

```shell
No relevant logs.
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

timminator:
This PR enables segment deletion for livestreams that are transcoded or remuxed.

The problem I ran into was that livestreams do not report the value downloadpositionticks to reflect the current watch position.
I came up with the idea to use the positionticks value reported in the OnPlaybackProgressinfo class. This worked perfectly and segments are now correctly deleted after the specified time by "Time to keep segments".

Fixes #13725

### DIFF
--- a/Emby.Server.Implementations/Session/SessionManager.cs
+++ b/Emby.Server.Implementations/Session/SessionManager.cs
@@ -62,6 +62,9 @@ namespace Emby.Server.Implementations.Session
         private readonly ConcurrentDictionary<string, SessionInfo> _activeConnections
             = new(StringComparer.OrdinalIgnoreCase);
 
+        private static readonly ConcurrentDictionary<string, long> _liveStreamPlaybackPositions
+            = new(StringComparer.OrdinalIgnoreCase);
+
         private Timer _idleTimer;
         private Timer _inactiveTimer;
 
@@ -818,6 +821,12 @@ namespace Emby.Server.Implementations.Session
                 }
             }
 
+            if (!string.IsNullOrEmpty(info.LiveStreamId) && !string.IsNullOrEmpty(info.PlaySessionId) && info.PlayMethod == PlayMethod.Transcode)
+            {
+                var playbackPosition = info.PositionTicks.HasValue ? info.PositionTicks.Value / 10000000 : 0;
+                _liveStreamPlaybackPositions.AddOrUpdate(info.PlaySessionId, playbackPosition, (key, oldValue) => playbackPosition);
+            }
+
             var eventArgs = new PlaybackProgressEventArgs
             {
                 Item = libraryItem,
@@ -911,6 +920,12 @@ namespace Emby.Server.Implementations.Session
             }
 
             return changed;
+        }
+
+        /// <inheritdoc />
+        public Task<long> GetLiveStreamPlaybackPositionAsync(string playSessionId)
+        {
+            return Task.FromResult(_liveStreamPlaybackPositions.TryGetValue(playSessionId, out var position) ? position : 0L);
         }
 
         /// <summary>
@@ -1007,6 +1022,11 @@ namespace Emby.Server.Implementations.Session
                 catch (Exception ex)
                 {
                     _logger.LogError(ex, "Error closing live stream");
+                }
+
+                if (!string.IsNullOrEmpty(info.PlaySessionId))
+                {
+                _liveStreamPlaybackPositions.TryRemove(info.PlaySessionId, out _);
                 }
             }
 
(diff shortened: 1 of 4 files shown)


gnattu (changes requested):
I don't like this design. I don't like the design of the segment removal feature from the beginning. It is designed to work unreliably and it was almost being removed in 10.9, and the only reason it is still kept is that it seems to work in most cases and a lot of people want to use super small cache disk or even use a ram disk. The whole reason to have such a complicated and unreliable mechanism is to allow seek-able videos to remove the  "probably no longer needed" segments. For non-seekable livestreams we should not use playback position based removal at all. This is just asking for troubles. If we really want to implement this, we modify the HLS parameters passed to ffmpeg by specifying `hls_list_size` and `hls_flags delete_segments`. This will make ffmpeg to rotate the playlist and remove the old segments.

timminator:
I worked on what was there. A complete redesign is not something that I can do.
In a transcoded livestream I can go back and forth like in a video that keeps getting extended. That was the reason I used the playback position similar to how it's currently handled for movies. I cannot see how this should cause more issues then the current implementation for movies? From my testing it worked exactly as expected.
This issue is really bad because a cache folder of easily more than 40GB is not feasible and this PR could fix this without a complete redesign that will probably take a long time to implement.

gnattu:
> In a transcoded livestream I can go back and forth like in a video that keeps getting extended.

This behavior is problematic enough because the source video(the livestream) is not able to "go back and forth". It is a livestream, with only a few seconds served each time you probe it. You can only seek back to the cached place and now you are removing those cache. Such playback should not be seekable at all in the first place, I don't know which client you are using but its player should prevent seeking on livestream to prevent playback issues when the cache will be going away. The client should also prevent pause/resume or handle pause/resume with its own client side caching mechanism for livestreams because it cannot expect the server side will always cache the data since the playback has started. 

>  I cannot see how this should cause more issues then the current implementation for movies?

The current implementation for movies is already unreliable enough that cause a few playback/seeking issues in some cases. We receive such reports from time to time. The design to use playback time for an infinite livestream is a hack at best, and is a hack based on a known to be unreliable system.

> This issue is really bad because a cache folder of easily more than 40GB is not feasible

This really depends on the use case. For high-bitrate BD remuxes, a 40GB transcoding cache is actually quite common. However, for your specific use case, this might be excessive. In my opinion, your implementation is not feasible for merging at least because it is hacky and there is at least one known better way to do it. If this works for you, good, you can use your version on your own deployment.

timminator:
> I  don't know which client you are using but its player should prevent seeking on livestream to prevent playback issues when the cache will be going away. 

This happens on all clients I am using: Jellyfin via Browser, Jellyfin on Android and on AndroidTV and Jellyfin WebOS. 

> The client should also prevent pause/resume or handle pause/resume with its own client side caching mechanism for livestreams because it cannot expect the server side will always cache the data since the playback has started.

This issue would also occur with your redesign. That is not specifically to my solution.

> The design to use playback time for an infinite livestream is a hack at best, and is a hack based on a known to be unreliable system.

I feel like my PR is the best solution to enable it for livestreams with the current underlying system, because as it is right now it is broken and segment deletion is not working as expected for livestreams.

> This really depends on the use case. For high-bitrate BD remuxes, a 40GB transcoding cache is actually quite common. 

But this is currently solved by enabling segment deletion as expected. And this feature is not applying for livestreams as it should.

I understand everything you said - the clients should handle this way better and the segment deletion as it is right now is unreliable, but maybe my PR could be a solution until a new better system implemented?
Changes to all clients and to the server seem to be needed for what you want to achieve. I cannot see this being done anytime soon...

JPVenson:
closed do to inactivity

Issue #13744, opened 2025-03-21
Custom splash screen images are removed on branding updates


thornbill:
The server determines that a custom splash screen exists by saving the file path in `BrandingOptions`. However the path is [ignored](https://github.com/jellyfin/jellyfin/blob/master/MediaBrowser.Model/Branding/BrandingOptions.cs#L34-L35) in the API so the path is removed when making any updates to `BrandingOptions`. This makes the custom image feature fundamentally broken.

https://github.com/jellyfin/jellyfin-web/pull/6616#pullrequestreview-2691652017

PR #13745, opened 2025-03-21
Preserve SplashscreenLocation when updating branding config

KGT1:
**Changes**
This PR adds handling for BrandingOptions to preserve SplashscreenLocation when updating configuration. 

**Issues**
#13744


crobibero (changes requested):
Please retarget on `master`, and move this logic into the `BaseConfigurationManager`

PR #13748, opened 2025-03-22
Remove textless poster

### LINKED ISSUE #9878: [Issue]: Jellyfin pulls textless posters from TMDB when language-matching posters exist
### Please describe your bug

For awhile now, occasionally Jellyfin pulls posters from TMDB with no text despite there being language-matching posters with text with more votes. These are annoying and clearly wrong, requiring manual intervention. Perhaps posters should be picked by most votes first, I feel it would produce more accurate results more often.

For example, my server instead of picking the primary image on the TMDB page for _The Great Dictator_, instead picked the textless one under "No Language" https://www.themoviedb.org/movie/914-the-great-dictator/images/posters?language=en-US

I found this PR that might have something to do with it: https://github.com/jellyfin/jellyfin/pull/8946

I thought this PR might have fixed it, but I can confirm with the attached image below (automatically sourced image is the same as the first one shown below) that this still reproduces

### Jellyfin Version

10.8.z

### if other:

_No response_

### Environment

```markdown
- OS: Unraid
- Linux Kernel:
- Virtualization: docker
- Clients: Any
- Browser: Any
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

Another active example is _Harry Potter and the Order of the Phoenix_, which currently shows this priority
image

_No response_

### Code of Conduct

TOomaAh:
**Changes**
Removed textless posters by enhancing image language filtering. For images like posters, logos, and thumbnails, this PR enforces language requirements to filter out images without text. Images are now retrieved with a default language preference, eliminating textless versions while maintaining the same behavior when the "All languages" option is selected.

I am not a C# developer, so any feedback or suggestions for improvement would be greatly appreciated. Please feel free to point out any issues or better approaches in the code - I'm looking to learn and improve.

**Issues**
Fixes #9878 
[forum post](https://forum.jellyfin.org/t-textless-posters)

### DIFF
--- a/MediaBrowser.Providers/Manager/ProviderManager.cs
+++ b/MediaBrowser.Providers/Manager/ProviderManager.cs
@@ -388,16 +388,14 @@ namespace MediaBrowser.Providers.Manager
         private bool IsValidImage(RemoteImageInfo image, string preferredLanguage, string fallbackLanguage = "en")
         {
             // TODO: should exception case of "en" (English) eventually be removed?
+            bool matchesLanguage = MatchesLanguage(preferredLanguage, image.Language) || MatchesLanguage(image.Language, fallbackLanguage);
+
             if (RequiresLanguage(image.Type))
             {
-                return image.Language is not null &&
-                       (MatchesLanguage(preferredLanguage, image.Language) ||
-                        MatchesLanguage(image.Language, fallbackLanguage));
+                return image.Language is not null && matchesLanguage;
             }
 
-            return string.IsNullOrEmpty(image.Language) ||
-                   MatchesLanguage(preferredLanguage, image.Language) ||
-                   MatchesLanguage(image.Language, fallbackLanguage);
+            return string.IsNullOrEmpty(image.Language) || matchesLanguage;
         }
 
         /// <summary>
--- a/MediaBrowser.Providers/Manager/ProviderManager.cs
+++ b/MediaBrowser.Providers/Manager/ProviderManager.cs
@@ -354,9 +354,9 @@ namespace MediaBrowser.Providers.Manager
                     // Filter out languages that do not match the preferred languages.
                     //
                     // TODO: should exception case of "en" (English) eventually be removed?
-                    result = result.Where(i => string.IsNullOrWhiteSpace(i.Language) ||
-                                               string.Equals(preferredLanguage, i.Language, StringComparison.OrdinalIgnoreCase) ||
-                                               string.Equals(i.Language, "en", StringComparison.OrdinalIgnoreCase));
+                    result = result.Where(i => i.Language is not null &&
+                                               (string.Equals(preferredLanguage, i.Language, StringComparison.OrdinalIgnoreCase) ||
+                                               string.Equals(i.Language, "en", StringComparison.OrdinalIgnoreCase)));
                 }
 
                 return result.OrderByLanguageDescending(preferredLanguage);
--- a/MediaBrowser.Providers/Manager/ProviderManager.cs
+++ b/MediaBrowser.Providers/Manager/ProviderManager.cs
@@ -354,9 +354,7 @@ namespace MediaBrowser.Providers.Manager
                     // Filter out languages that do not match the preferred languages.
                     //
                     // TODO: should exception case of "en" (English) eventually be removed?
-                    result = result.Where(i => i.Language is not null &&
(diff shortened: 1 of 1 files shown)


TOomaAh:
> It seems like personal preference.
> [...]
> ... or a personal patch.

I'm just resolving a github issue that's also causing me problems (like a lot of PR).

> Should be constant for all (major) media providers and not all provide the language information.
> [...]
> You skipped other content that sometimes use languages like banners and clearart

You are right

> Smaller releases sometime have only low res 640px or text cluttered posters, and you hide high-res posters

Do posters with text sometimes have a higher resolution than posters without text? I may have misunderstood this part.

> Smaller releases sometime have only low res 640px or text cluttered posters, and you hide high-res posters
> Some users prefer a textless (depending on the client logo is standalone and could be overlayed or so)

Posters without text can still be retrieved.

> Seems something that could be an option

I agree that this should be an option. But the code for the tmdb plugin, for example, has a comment: ‘TODO use image languages if All Languages isn't toggled, but there's currently no way to get that value in here’.

So this is something that's normally intended, not just something I want to do (which I get the impression you don't like).

It's something I'm trying to do in parallel, but as I said in my initial post, I'm not a C# developer.

TOomaAh:
> Yeah, but there are a lot of providers and the behavior should be consistent. If you go with TVDB because it forces a resolution and ratio and TMDB as fallback, you will always get Fallback content. TVDB has no such info.

I agree, the treatment must be consistent for all providers (TVDB, TMDB or Fanart have this language information but not TVMaze for example) but also not exclude any provider that might not have this information.

> TMDB is user content, so smaller res is sometimes the case (older things seem to be scrapped from other DBs). If you want a non-English image, the selection is often tiny, only logo or a season name included and not the series. New blockbuster have normally no issues, but smaller/older releases will suffer. If you want a textless, because all the text options aren't that good, you need to look between unknown languages.

I haven't removed the search for posters without text or in other languages. I just put a poster with text rather than without text in the automatic search. If you ever want posters without text or even in another language, my code still allows it.

I'm going to see if this can easily become an option. Although I'm not sure where to put it.

TOomaAh:
Maybe just to give an example. The default search does not display items without text.

image

But the full search still exists. I haven't removed it at all, and I haven't removed the possibility of retrieving posters in another language or without text. But by default, all posters have text.

image

TOomaAh:
> I mean in your case for major US release it works, but if you have a lot of non-English artwork, it will be mixed in, and you can get a lot of additional pages with unfamiliar languages. If you watch anything not made in US it will be a huge mix.
> 
> Look at all languages added with All languages. And also `All Languages` indicate text and not `textless` basically the opposite.
> 
> e.g. https://www.themoviedb.org/movie/12405-slumdog-millionaire/images/posters https://www.themoviedb.org/tv/82452-avatar-the-last-airbender/images/posters
> 
> An option per library (that adds/sorts) the providers should be added and not force everybody to have covers with a logo and finding textless in big language text mix. Usually only a few textless exist anyway (especially highly voted) and they are foreign to anybody.

Whatever happens, the current default sorting order shows posters in your language first, posters with no language and finally posters in English.

For your examples, let's choose ‘Slumdog Millionaire’ (but it's the same for Avatar), you'll normally find (because there are lots of languages for the posters, unless you have an unusual language) your posters with text first, then the posters without text and then the posters in other languages. Posters without text won't be drowned out by a mixture of posters in other languages, as you say. This is the current order for all filters (based on provider, language or other). If, in your example, you want to choose an image without text, you have to choose it manually.

Now, here are some examples (I only put 3 but I have a lot if you want):
https://www.themoviedb.org/movie/1029235-azrael/images/posters
https://www.themoviedb.org/movie/726429-xtremo/images/posters
https://www.themoviedb.org/movie/2109-rush-hour/images/posters

There is no poster in my language. The title of the film in my language is the same as the one in English (apart from Russian and Chinese) and yet, by default, I have a poster without text (I'm talking about the default display). For many media (because I'm not American or even English) I have to click on each media to choose a poster, even though a poster with text exists with the same title.

Anyway, I'm in the process of creating an option in the library settings.

I'd love to hear the Jellyfin team's opinion on this. I've put this option in the ‘fetcher settings’ in the libraries, with the default behaviour being to retrieve posters without text.

Issue #12427, opened 2024-08-11
Extracted subtitles never get removed if media gets (re)moved.


sjorge:
### This issue respects the following points:

- [X] This is a **bug**, not a question or a configuration issue; Please visit our forum or chat rooms first to troubleshoot with volunteers, before creating a report. The links can be found [here](https://jellyfin.org/contact/).
- [X] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [X] I'm using an up to date version of Jellyfin Server stable, unstable or master; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.
- [X] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [X] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Description of the bug

Subtitles extracted during playback or by `jellyfin-plugin-subtitleextract` never get cleaned up if the media they belong to is (re)moved.

I would have expected either a full scan to trigger removal once the media is detected as removed or there to be a task to clean those up.

### Reproduction steps

1. setup a new jellyfin instance
2. create a library (show or movie)
3. add media to newly created library that has embedded subtitles
4. (a) play media with subtitles so they are extracted on demand || (b) install and run the `jellyfin-plugin-subtitleextract` 
5. check /var/lib/jellyfin/data/subtitles for the newly extracted files
6. remove the media the subtitles belong to
7. trigger a few `scan all libraries` and a few `Clean Cache Directory` tasks runs (although technically they subtitles are not in the cache paths, but it was worth a shot)
8. files are still in  /var/lib/jellyfin/data/subtitles 

### What is the current _bug_ behavior?

Subtitles once extracted remain in /var/lib/jellyfin/data/subtitles forever.

### What is the expected _correct_ behavior?

Subtitles should be cleaned up from /var/lib/jellyfin/data/subtitles once the media they belong to gets removed.

### Jellyfin Server version

10.9.9+

### Specify version number

10.9.9

### Specify the build version

10.9.9

### Environment

```markdown
- OS: Ubuntu 22.04 LTS
[...]

PR #13753, opened 2025-03-23
Remove all DB data on item removal, delete internal trickplay files

Shadowghost:
**Changes**
* Cleanup additional DB tables on item removal

**Issues**
* Images and subtitles can't reliably be cleaned up because their save folder depends on the media files's modifcaition date

**Still not cleaned up**
- keyframe JSONs
- subtitles (we are not able to clean them up because the path depends on the modification date of the already not existing media file...)
- attachments

### DIFF
--- a/Emby.Server.Implementations/Library/LibraryManager.cs
+++ b/Emby.Server.Implementations/Library/LibraryManager.cs
@@ -78,6 +78,7 @@ namespace Emby.Server.Implementations.Library
         private readonly NamingOptions _namingOptions;
         private readonly IPeopleRepository _peopleRepository;
         private readonly ExtraResolver _extraResolver;
+        private readonly IPathManager _pathManager;
 
         /// <summary>
         /// The _root folder sync lock.
@@ -113,7 +114,8 @@ namespace Emby.Server.Implementations.Library
         /// <param name="imageProcessor">The image processor.</param>
         /// <param name="namingOptions">The naming options.</param>
         /// <param name="directoryService">The directory service.</param>
-        /// <param name="peopleRepository">The People Repository.</param>
+        /// <param name="peopleRepository">The people repository.</param>
+        /// <param name="pathManager">The path manager.</param>
         public LibraryManager(
             IServerApplicationHost appHost,
             ILoggerFactory loggerFactory,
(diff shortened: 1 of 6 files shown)


JPVenson (approved):

sjorge:
> * subtitles (we are not able to clean them up because the path depends on the modification date of the already not existing media file...)

https://github.com/jellyfin/jellyfin/issues/12427#issue-2459528219 still seems to work on current stable at least.

It does a bit of the reverse, it basically generates a list of all currently valid subtitle files... then walks the data dir for them and remove them if they are no longer valid. I think the same approach could also work for attachments but most of the wasted space for me is always in the subtitles.

Bond-009 (approved):

Shadowghost:
@sjorge it seems like we are saving the date in the baseitem, so we should be able to properly calculate the path. I'd still rather fix this properly by not using this kind of hackery in 10.12. Same issue goes for attachments.

sjorge:
> @sjorge it seems like we are saving the date in the baseitem, so we should be able to properly calculate the path. I'd still rather fix this properly by not using this kind of hackery in 10.12. Same issue goes for attachments.

Absolutely agree that the way it's done there is very hacky. Was more of a if you couldn't think of a way to do it, it's possible. If there are better ways to fix it eventually, that is of course preferred.

crobibero (approved):

Emby.Server.Implementations/Library/LibraryManager.cs
```diff
@@ -78,6 +78,7 @@ public class LibraryManager : ILibraryManager
         private readonly NamingOptions _namingOptions;
         private readonly IPeopleRepository _peopleRepository;
         private readonly ExtraResolver _extraResolver;
+        private readonly IPathManager _pathManager;
```
jhparker88, line 82: Tests failing with `   Failed Jellyfin.Server.Integration.Tests.Controllers.MusicGenreControllerTests.MusicGenres_FakeMusicGenre_NotFound [380 ms]
  Error Message:
   System.InvalidOperationException : Unable to resolve service for type 'MediaBrowser.Controller.IO.IPathManager' while attempting to activate 'Emby.Server.Implementations.Library.LibraryManager'.
  Stack Trace:`

@Shadowghost I'm guessing the `MediaBrowser.Controller.IO;` import is missing here!

PR #13589, opened 2025-02-19
[Feature] Database code refactor

JPVenson:
Refactors code architecture for database usage.

Changes:

    Moved database related code to own Project
    Refactored User entity to not use collation for username for compataibility reasons
    Refactored user and role entities to no longer implement buissniss logic
    Added SqLite adapter
    Added optional database.xml configuration file
    Added document for explaining migration path

This PR supersedes the groundwork done for #13451

PR #13754, opened 2025-03-23
Feature/backup on migration

JPVenson:
Simplifies migration backups for all applicable migrations.

**Changes**
Adds MigrationRunner migration when a migration declares the be an `IDatabaseMigrationRoutine`

This builds upon #13589 and needs to be merged afterwards


Shadowghost (approved):

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteDatabaseProvider.cs
```diff
@@ -84,4 +85,36 @@ public void ConfigureConventions(ModelConfigurationBuilder configurationBuilder)
     {
         configurationBuilder.Conventions.Add(_ => new DoNotUseReturningClauseConvention());
     }
+
+    /// <inheritdoc />
+    public Task<string> MigrationBackupFast(CancellationToken cancellationToken)
+    {
+        var key = Guid.NewGuid().ToString("D");
+        var path = Path.Combine(_applicationPaths.DataPath, "jellyfin.db");
```
crobibero, line 93: I think it would be good to append the timestamp that the backup was created here
I know that file create/change stamps are a thing but names are also good :)

JPVenson: good point, in that case i would exchange the guid with just the timestamp as they _key_ is just arbitrary anyway

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteDatabaseProvider.cs
```diff
@@ -16,6 +17,7 @@ namespace Jellyfin.Database.Providers.Sqlite;
 [JellyfinDatabaseProviderKey("Jellyfin-SQLite")]
 public sealed class SqliteDatabaseProvider : IJellyfinDatabaseProvider
 {
+    private const string BACKUPFOLDERNAME = "SQLiteBackups";
```
crobibero, line 20:
```suggestion
    private const string BackupFolderName= "SQLiteBackups";
```

Surprised the analyzer didn't flag the full caps 😆

Jellyfin.Server/Migrations/MigrationRunner.cs
```diff
@@ -111,38 +119,60 @@ private static void HandleStartupWizardCondition(IEnumerable<IMigrationRoutine>
...
                 {
-                    migrationRoutine.Perform();
+                    migrationKey = await jellyfinDatabaseProvider.MigrationBackupFast(CancellationToken.None).ConfigureAwait(false);
                 }
-                catch (Exception ex)
+                catch (NotImplementedException)
                 {
-                    logger.LogError(ex, "Could not apply migration '{Name}'", migrationRoutine.Name);
-                    throw;
+                    logger.LogWarning("Dont perform backup of database before migration because provider does not support it");
```
jhparker88, line 144: "Could not perform ..." instead of "Dont perform ..."

Jellyfin.Server/Migrations/MigrationRunner.cs
```diff
@@ -111,38 +119,60 @@ private static void HandleStartupWizardCondition(IEnumerable<IMigrationRoutine>
...
-                var migrationRoutine = migrations[i];
-                if (appliedMigrationIds.Contains(migrationRoutine.Id))
-                {
-                    logger.LogDebug("Skipping migration '{Name}' since it is already applied", migrationRoutine.Name);
-                    continue;
-                }
-
-                logger.LogInformation("Applying migration '{Name}'", migrationRoutine.Name);
-
+                logger.LogInformation("Database backup is performed");
```
Bond-009, line 137: Make it clearer that it hasn't been done yet
```suggestion
                logger.LogInformation("Performing database backup");
```

Jellyfin.Server/Migrations/MigrationRunner.cs
```diff
@@ -111,38 +119,60 @@ private static void HandleStartupWizardCondition(IEnumerable<IMigrationRoutine>
...
+            IJellyfinDatabaseProvider? jellyfinDatabaseProvider)
         {
             // save already applied migrations, and skip them thereafter
             saveConfiguration(migrationOptions);
             var appliedMigrationIds = migrationOptions.Applied.Select(m => m.Id).ToHashSet();
+            var migrationsToBeApplied = migrations.Where(e => !appliedMigrationIds.Contains(e.Id)).ToArray();
 
-            for (var i = 0; i < migrations.Length; i++)
+            string? migrationKey = null;
+            if (migrationsToBeApplied.Any(f => f is IDatabaseMigrationRoutine) && jellyfinDatabaseProvider is not null)
```
Bond-009, line 135: Lets do the cheaper check first
```suggestion
            if (jellyfinDatabaseProvider is not null && migrationsToBeApplied.Any(f => f is IDatabaseMigrationRoutine))
```

Jellyfin.Server/Migrations/MigrationRunner.cs
```diff
@@ -111,38 +119,60 @@ private static void HandleStartupWizardCondition(IEnumerable<IMigrationRoutine>
...
-                saveConfiguration(migrationOptions);
-                logger.LogDebug("Migration '{Name}' marked as applied in configuration.", migrationRoutine.Name);
+                    // Mark the migration as completed
+                    logger.LogInformation("Migration '{Name}' applied successfully", migrationRoutine.Name);
+                    migrationOptions.Applied.Add((migrationRoutine.Id, migrationRoutine.Name));
+                    saveConfiguration(migrationOptions);
+                    logger.LogDebug("Migration '{Name}' marked as applied in configuration.", migrationRoutine.Name);
+                }
+            }
+            catch (System.Exception) when (migrationKey is not null && jellyfinDatabaseProvider is not null)
```
Bond-009, line 171: It's impossible for jellyfinDatabaseProvider to be null, no?
```suggestion
            catch (System.Exception) when (migrationKey is not null)
```

Bond-009 (approved):
Looks good otherwis

src/Jellyfin.Database/Jellyfin.Database.Implementations/IJellyfinDatabaseProvider.cs
```diff
@@ -45,4 +46,20 @@ public interface IJellyfinDatabaseProvider
...
+    /// <summary>
+    /// Runs a full Database backup that can later be restored to.
+    /// </summary>
+    /// <param name="cancellationToken">A cancelation token.</param>
+    /// <returns>A key to identify the backup.</returns>
+    /// <exception cref="NotImplementedException">May throw an NotImplementException if this operation is not supported for this database.</exception>
+    Task<string> MigrationBackupFast(CancellationToken cancellationToken);
+
+    /// <summary>
+    /// Runs a full Database backup that can later be restored to.
```
jhparker88, line 59: I'm guessing this method restores a backup for a given key instead of " Runs a full Database backup ..."

Jellyfin.Server/Migrations/MigrationRunner.cs
JPVenson: Technically this should never happen, only if somebody were to try to perform a database related migration on the pre-migration stage but i guess you are right

crobibero (approved):

PR #13756, opened 2025-03-23
Preserve SplashscreenLocation when updating branding config

KGT1:
Changes
This PR adds handling for BrandingOptions to preserve SplashscreenLocation when updating configuration.

_sorry accidentally closed the last_

Moved logic from to the ConfigurationController to BaseConfigurationManager.cs and now target master

Issues
Fixes: #13744

### DIFF
--- a/Jellyfin.Api/Controllers/ConfigurationController.cs
+++ b/Jellyfin.Api/Controllers/ConfigurationController.cs
@@ -9,6 +9,7 @@ using Jellyfin.Extensions.Json;
 using MediaBrowser.Common.Api;
 using MediaBrowser.Controller.Configuration;
 using MediaBrowser.Controller.MediaEncoding;
+using MediaBrowser.Model.Branding;
 using MediaBrowser.Model.Configuration;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Http;
@@ -117,6 +118,51 @@ public class ConfigurationController : BaseJellyfinApiController
     public ActionResult<MetadataOptions> GetDefaultMetadataOptions()
     {
         return new MetadataOptions();
+    }
+
+    /// <summary>
+    /// Gets branding configuration.
+    /// </summary>
+    /// <response code="200">Branding configuration returned.</response>
+    /// <returns>Branding configuration.</returns>
+    [HttpGet("Configuration/Branding")]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    public ActionResult<BrandOptionsDto> GetBrandingConfiguration()
+    {
+        var brandingOptions = (BrandingOptions)_configurationManager.GetConfiguration("branding");
+
+        var brandOptionsDto = new BrandOptionsDto
+        {
+            LoginDisclaimer = brandingOptions.LoginDisclaimer,
+            CustomCss = brandingOptions.CustomCss,
+            SplashscreenEnabled = brandingOptions.SplashscreenEnabled
+        };
+
+        return brandOptionsDto;
+    }
+
+    /// <summary>
+    /// Updates branding configuration.
+    /// </summary>
+    /// <param name="configuration">Branding configuration.</param>
+    /// <response code="204">Branding configuration updated.</response>
+    /// <returns>Update status.</returns>
+    [HttpPost("Configuration/Branding")]
+    [Authorize(Policy = Policies.RequiresElevation)]
+    [ProducesResponseType(StatusCodes.Status204NoContent)]
+    public ActionResult UpdateBrandingConfiguration([FromBody, Required] BrandOptionsDto configuration)
+    {
+        // Get the current branding configuration to preserve SplashscreenLocation
+        var currentBranding = (BrandingOptions)_configurationManager.GetConfiguration("branding");
+
+        // Update only the properties from BrandOptionsDto
+        currentBranding.LoginDisclaimer = configuration.LoginDisclaimer;
+        currentBranding.CustomCss = configuration.CustomCss;
+        currentBranding.SplashscreenEnabled = configuration.SplashscreenEnabled;
+
+        _configurationManager.SaveConfiguration("branding", currentBranding);
+
+        return NoContent();
     }
 
     /// <summary>
--- a/Jellyfin.Api/Controllers/ConfigurationController.cs
+++ b/Jellyfin.Api/Controllers/ConfigurationController.cs
@@ -121,27 +121,6 @@ public class ConfigurationController : BaseJellyfinApiController
     }
 
     /// <summary>
-    /// Gets branding configuration.
-    /// </summary>
-    /// <response code="200">Branding configuration returned.</response>
-    /// <returns>Branding configuration.</returns>
-    [HttpGet("Configuration/Branding")]
-    [ProducesResponseType(StatusCodes.Status200OK)]
-    public ActionResult<BrandOptionsDto> GetBrandingConfiguration()
-    {
-        var brandingOptions = (BrandingOptions)_configurationManager.GetConfiguration("branding");
-
-        var brandOptionsDto = new BrandOptionsDto
-        {
-            LoginDisclaimer = brandingOptions.LoginDisclaimer,
-            CustomCss = brandingOptions.CustomCss,
-            SplashscreenEnabled = brandingOptions.SplashscreenEnabled
-        };
-
-        return brandOptionsDto;
-    }
-
-    /// <summary>
     /// Updates branding configuration.
     /// </summary>
     /// <param name="configuration">Branding configuration.</param>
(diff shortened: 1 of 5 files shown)


KGT1:
I dont think tests failing is on my end? they pass locally and in the dev container just fine.

crobibero:
Bond suggested in chat that we use `IValidatingConfiguration` to handle the logic for us, I think it ends up cleaner this way and we don't pollute the BaseConfigurationManager.

```patch
Index: Emby.Server.Implementations/Branding/BrandingConfigurationFactory.cs
IDEA additional info:
<+>UTF-8
diff --git a/Emby.Server.Implementations/Branding/BrandingConfigurationFactory.cs b/Emby.Server.Implementations/Branding/BrandingConfigurationFactory.cs
--- a/Emby.Server.Implementations/Branding/BrandingConfigurationFactory.cs	(revision a7c34b2be2f5eb98fd4555d8e987773b8b485052)
+++ b/Emby.Server.Implementations/Branding/BrandingConfigurationFactory.cs	(date 1743127895980)
@@ -2,24 +2,16 @@
 using MediaBrowser.Common.Configuration;
 using MediaBrowser.Model.Branding;
 
-namespace Emby.Server.Implementations.Branding
-{
-    /// <summary>
-    /// A configuration factory for <see cref="BrandingOptions"/>.
-    /// </summary>
-    public class BrandingConfigurationFactory : IConfigurationFactory
-    {
-        /// <inheritdoc />
-        public IEnumerable<ConfigurationStore> GetConfigurations()
-        {
-            return new[]
-            {
-                new ConfigurationStore
-                {
-                     ConfigurationType = typeof(BrandingOptions),
-                     Key = "branding"
-                }
-            };
-        }
+namespace Emby.Server.Implementations.Branding;
+
+/// <summary>
+/// A configuration factory for <see cref="BrandingOptions"/>.
+/// </summary>
+public class BrandingConfigurationFactory : IConfigurationFactory
+{
+    /// <inheritdoc />
+    public IEnumerable<ConfigurationStore> GetConfigurations()
+    {
+        return [new BrandingConfigurationStore()];
     }
 }
Index: Emby.Server.Implementations/AppBase/BaseConfigurationManager.cs
IDEA additional info:
<+>UTF-8
diff --git a/Emby.Server.Implementations/AppBase/BaseConfigurationManager.cs b/Emby.Server.Implementations/AppBase/BaseConfigurationManager.cs
--- a/Emby.Server.Implementations/AppBase/BaseConfigurationManager.cs	(revision a7c34b2be2f5eb98fd4555d8e987773b8b485052)
+++ b/Emby.Server.Implementations/AppBase/BaseConfigurationManager.cs	(date 1743127547768)
@@ -335,17 +335,6 @@
                 validatingStore.Validate(currentConfiguration, configuration);
             }
 
-            // Special handling for BrandingOptions to preserve SplashscreenLocation
-            if (key.Equals("branding", StringComparison.OrdinalIgnoreCase) &&
-                configurationType == typeof(MediaBrowser.Model.Branding.BrandingOptions))
-            {
-                var existingOptions = (MediaBrowser.Model.Branding.BrandingOptions)GetConfiguration(key);
-                var newOptions = (MediaBrowser.Model.Branding.BrandingOptions)configuration;
-
-                // Preserve SplashscreenLocation
-                newOptions.SplashscreenLocation = existingOptions.SplashscreenLocation;
-            }
-
             NamedConfigurationUpdating?.Invoke(this, new ConfigurationUpdateEventArgs(key, configuration));
 
             _configurations.AddOrUpdate(key, configuration, (_, _) => configuration);
Index: Emby.Server.Implementations/Branding/BrandingConfigurationStore.cs
IDEA additional info:
<+>UTF-8
diff --git a/Emby.Server.Implementations/Branding/BrandingConfigurationStore.cs b/Emby.Server.Implementations/Branding/BrandingConfigurationStore.cs
new file mode 100644
--- /dev/null	(date 1743127908687)
+++ b/Emby.Server.Implementations/Branding/BrandingConfigurationStore.cs	(date 1743127908687)
@@ -0,0 +1,28 @@
+using MediaBrowser.Common.Configuration;
+using MediaBrowser.Model.Branding;
+
+namespace Emby.Server.Implementations.Branding;
+
+/// <summary>
+/// Configuration store for branding.
+/// </summary>
+public class BrandingConfigurationStore : ConfigurationStore, IValidatingConfiguration
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="BrandingConfigurationStore"/> class.
+    /// </summary>
+    public BrandingConfigurationStore()
+    {
+        ConfigurationType = typeof(BrandingOptions);
+        Key = "branding";
+    }
+
+    /// <inheritdoc />
+    public void Validate(object oldConfig, object newConfig)
+    {
+        var newBranding = (BrandingOptions)newConfig;
+        var oldBranding = (BrandingOptions)oldConfig;
+
+        newBranding.SplashscreenLocation = oldBranding?.SplashscreenLocation;
+    }
+}
```

KGT1:
Thanks for the patch, ive applied and tested it. It works just as fine.

Generally i think having a separation with dto's would be better. Your change seems more maintainable even tho its not really validation.

Sorry, i dont think ive given proper credit in the commit? but since none of my changes are remaining if you want to take this over, be my guest.

JPVenson (approved):

nielsvanvelzen:
Would it make sense to add unit tests for this?

KGT1:
Guys sorry the more i look at this code the more i dont like it.
This currently only works and updates the image to a new one because in ImageController.cs the configuration is updated by reference in memory before "validating" it. So when saving and validating it, it allready has the new path in oldBranding. 
```
var brandingOptions = _serverConfigurationManager.GetConfiguration<BrandingOptions>("branding");
brandingOptions.SplashscreenLocation = filePath;
_serverConfigurationManager.SaveConfiguration("branding", brandingOptions);
```

i think a proper fix would refactor BrandingOptions.cs into an internal one and a dto. the real issue is, that:
```
var deserializedConfiguration = configuration.Deserialize(configurationType, _serializerOptions)
```
adds `splashscreenLocation: null`, since its not passed in the json via the api and then updates it.

Ill of course let you guys decide whats best but if you dont mind i would just forcepush this in here?:
https://github.com/jellyfin/jellyfin/compare/master...KGT1:jellyfin:fixSplashscreenDto

crobibero:
I was never a big fan of the generic configuration endpoint, so I think making it specific makes sense.
Just make the route `Configuration/Branding` to follow our naming convention

Jellyfin.Api/Controllers/ConfigurationController.cs
```diff
@@ -119,6 +120,51 @@ public ActionResult<MetadataOptions> GetDefaultMetadataOptions()
...
     }
 
+    /// <summary>
+    /// Gets branding configuration.
+    /// </summary>
+    /// <response code="200">Branding configuration returned.</response>
+    /// <returns>Branding configuration.</returns>
+    [HttpGet("Configuration/Branding")]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    public ActionResult<BrandOptionsDto> GetBrandingConfiguration()
```
nielsvanvelzen, line 130: There already is a getBrandingOptions in BrandingController. That one needs to be updated to use the DTO and this operation removed.

nielsvanvelzen (changes requested):

Jellyfin.Api/Controllers/ConfigurationController.cs
KGT1: yeah wanted to do that. shouldnt break the api now👀

KGT1:
> The BrandingOptions type still ends up in our OpenAPI specification, not sure how that is added now though as it seems unused. Otherwise the API changes look fine.

i think i know why, it should still be returned by
[HttpGet("Configuration/{key}")]

want me to prevent that with an if, or should i maybe just add the 
[HttpGet("Configuration/Branding")] 

again but just with the same method as /Branding/Configuration?
I think i would prefer the 2nd just to also have it in the same place.

nielsvanvelzen:
I don't think that's the reason it adds the model to the specification, I think we manually add it to Swashbuckle somewhere. @crobibero probably knows what is doing that.

crobibero:
🪄 https://github.com/jellyfin/jellyfin/blob/master/Jellyfin.Server/Filters/AdditionalModelFilter.cs#L180

KGT1:
sooo.. want me to add the BrandingOptions to _ignoredConfigurations in that file?

crobibero:
👍 that makes sense to me

MediaBrowser.Model/Branding/BrandOptionsDto.cs
```diff
@@ -0,0 +1,25 @@
+namespace MediaBrowser.Model.Branding;
+
+/// <summary>
+/// The branding options DTO for API use.
+/// This DTO excludes SplashscreenLocation to prevent it from being updated via API.
+/// </summary>
+public class BrandOptionsDto
```
nielsvanvelzen, line 7: For naming consistency we'd probably want to name this "BrandingOptionsDto". No strong preference on either option though.

nielsvanvelzen (approved):
API changes look fine now. Would be nice if we changes all current configuration files to this approach in the long term!

crobibero (approved):

PR #13764, opened 2025-03-24
Added OpenAPI spec for #12880

JPVenson:
Added open api spec for new 503 response on all endpoints

### DIFF
--- a/Jellyfin.Server/Extensions/ApiServiceCollectionExtensions.cs
+++ b/Jellyfin.Server/Extensions/ApiServiceCollectionExtensions.cs
@@ -247,6 +247,7 @@ namespace Jellyfin.Server.Extensions
                 c.AddSwaggerTypeMappings();
 
                 c.SchemaFilter<IgnoreEnumSchemaFilter>();
+                c.OperationFilter<RetryOnTemporarlyUnavailableFilter>();
                 c.OperationFilter<SecurityRequirementsOperationFilter>();
                 c.OperationFilter<FileResponseFilter>();
                 c.OperationFilter<FileRequestFilter>();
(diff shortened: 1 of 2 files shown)


Shadowghost (approved):

Bond-009 (approved):

oddstr13:
Might be desirable to include a hint about the retry-after header in the spec?

    The server is currently starting or is temporarly not available.

Specifically that string, if it isn't otherwise clearly visible in the documentation interface.

JPVenson:
where would you propose?
I am totally open to it.

oddstr13:
Should be all good then :)

oddstr13:
For the API paths, I think it would make sense to duplicate the information in a JSON response (getting HTML when JSON is expected is generally annoying).

That is out of scope for this PR tho, and I haven't checked the API how this kind of thing would be handled in different parts and error conditions (consistency is king).

nielsvanvelzen (changes requested):

nielsvanvelzen:
> For the API paths, I think it would make sense to duplicate the information in a JSON response (getting HTML when JSON is expected is generally annoying).
> 
> That is out of scope for this PR tho, and I haven't checked the API how this kind of thing would be handled in different parts and error conditions (consistency is king).

Currently the response body should never be trusted whenever the Jellyfin API does not respond with HTTP 200 Ok. Sometimes it will respond with JSON, sometimes with plaintext and sometimes with HTML.
For this specific PR we chose to not trust the body as reverse-proxies and tunneling services may respond with captcha pages etc.

nielsvanvelzen (approved):

Issue #13709, opened 2025-03-06
Player Remote UI only working for Admin


BotBlake:
### This issue respects the following points:

- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-web/issues?q=is%3Aissue) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Describe the bug

When trying to use Jellyfins builtin remote, the UI for it only works when you have the "manage Server" permissions.
Here is a screenshot of an admin streaming to the AndroidTV client:

![Image](https://github.com/user-attachments/assets/43e2c2c1-e9e5-4acd-9366-edb222ceafae)

And here is a screenshot of a normal user doing the same:

![Image](https://github.com/user-attachments/assets/03d48e9a-1d27-4860-af1e-020be58923da)

This issue does not only happen with the

### Reproduction Steps

1. Open a Session on a Client that can be Remote Controlled (ATV, Web, ...) as a User (non admin)
2. Open a Session on Jellyfin-Web with the same non admin user
3. "stream" from the jellyfin web session to the other Client.
4. Click on the "now playing" view

### Expected/Actual behaviour

I expected the RemoteUI to work for normal Users as well.

### Server version

10.10.6

### Web version

10.10.6

### Build version

10.10.6

### Platform

Linux, Windows

### Browser

Firefox, Chrome

PR #13767, opened 2025-03-24
Allow non-admin users to subscribe to their own Sessions

KGT1:
Changes:
This PR fixes an issue where non-admin users were unable to subscribe to their own sessions via WebSockets.

Question: should i rather then creating the method `GetDataToSendForConnection` refactor `GetDataToSend` to always require a socket connection? or do you think the current approach is fine? 

 
Issues:
#13709

### DIFF
--- a/Jellyfin.Api/WebSocketListeners/SessionInfoWebSocketListener.cs
+++ b/Jellyfin.Api/WebSocketListeners/SessionInfoWebSocketListener.cs
@@ -1,4 +1,5 @@
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using Jellyfin.Data.Enums;
 using MediaBrowser.Controller.Authentication;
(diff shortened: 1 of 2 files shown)


theguymadmax:
@KGT1 For issues, please use the format "Fixes: #itemnumber." This will ensure that GitHub automatically links the issue.

Shadowghost (approved):

crobibero (approved):

Jellyfin.Api/WebSocketListeners/SessionInfoWebSocketListener.cs
```diff
@@ -55,6 +56,25 @@ protected override Task<IEnumerable<SessionInfo>> GetDataToSend()
...
 
+    /// <summary>
+    /// Gets the data to send for a specific connection.
+    /// </summary>
+    /// <param name="connection">The connection.</param>
+    /// <returns>Task{IEnumerable{SessionInfo}}.</returns>
+    protected override Task<IEnumerable<SessionInfo>> GetDataToSendForConnection(IWebSocketConnection connection)
+    {
+        // For non-admin users, filter the sessions to only include their own sessions
+        if (connection.AuthorizationInfo?.User != null &&
```
Bond-009:
```suggestion
        if (connection.AuthorizationInfo?.User is not null &&
```

Jellyfin.Api/WebSocketListeners/SessionInfoWebSocketListener.cs
```diff
@@ -55,6 +56,25 @@ protected override Task<IEnumerable<SessionInfo>> GetDataToSend()
         return Task.FromResult(_sessionManager.Sessions);
     }
 
+    /// <summary>
+    /// Gets the data to send for a specific connection.
+    /// </summary>
+    /// <param name="connection">The connection.</param>
+    /// <returns>Task{IEnumerable{SessionInfo}}.</returns>
```
Bond-009:
```suggestion
    /// <inheritdoc />
```

Bond-009 (approved):

Shadowghost:
@KGT1 if you could please address Bond's comment, then we can get this merged. Sorry for the delays.

PR #13771, opened 2025-03-25
Import Keyframes into database

Shadowghost:
Keyframe data is currently saved in JSON files. This PR migrates the data into the database and properly cleans it up when the item is removed.

A mediaSource is always an item.


src/Jellyfin.Database/Jellyfin.Database.Implementations/Entities/KeyframeData.cs
```diff
@@ -0,0 +1,42 @@
...
+public class KeyframeData
+{
+    /// <summary>
+    /// Gets the id.
+    /// </summary>
+    /// <remarks>
+    /// Required.
+    /// </remarks>
+    [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
+    public int Id { get; private set; }
```
crobibero, line 21: Why have an integer id instead of a primary key based on ItemId and MediaSourceId?
Doing this would also add an index to made querying quicker

Shadowghost: I'll need to check how to do that in efcore, but makes totally sense

src/Jellyfin.Database/Jellyfin.Database.Implementations/Entities/KeyframeData.cs
```diff
@@ -0,0 +1,42 @@
...
+
+    /// <summary>
+    /// Gets or Sets the ItemId.
+    /// </summary>
+    public required Guid ItemId { get; set; }
```
JPVenson, line 26: Also add the navigation properties so EFCore can properly add Forgin keys:

```
public BaseItem Item { get; set; }
public MediaSource MediaSource { get; set; }
```

Jellyfin.Server.Implementations/Item/KeyframeRepository.cs
```diff
@@ -0,0 +1,63 @@
...
+    /// <inheritdoc />
+    public IReadOnlyList<MediaEncoding.Keyframes.KeyframeData> GetKeyframeData(Guid itemId)
+    {
+        using var context = _dbProvider.CreateDbContext();
+
+        return context.KeyframeData.AsNoTracking().Where(e => e.ItemId.Equals(itemId)).Select(e => Map(e)).ToList();
+    }
+
+    /// <inheritdoc />
+    public void SaveKeyframeData(Guid itemId, MediaEncoding.Keyframes.KeyframeData data, CancellationToken cancellationToken)
```
JPVenson, line 54: New DB methods should always be async

Jellyfin.Server/Migrations/Routines/MigrateKeyframeData.cs
```diff
@@ -0,0 +1,172 @@
...
+using MediaBrowser.Controller.Library;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.Logging;
+
+namespace Jellyfin.Server.Migrations.Routines;
+
+/// <summary>
+/// Migration to move extracted files to the new directories.
+/// </summary>
+public class MigrateKeyframeData : IMigrationRoutine
```
JPVenson, line 25:
```suggestion
public class MigrateKeyframeData : IDatabaseMigrationRoutine
```

JPVenson (changes requested):

JPVenson (approved):

Jellyfin.Server/Migrations/Routines/MigrateKeyframeData.cs
```diff
@@ -0,0 +1,172 @@
...
+                    context.KeyframeData.AddRange(keyframes);
+                    keyframes.Clear();
+                    _logger.LogInformation("Imported keyframes for {Count} items in {Time}", itemCount, sw.Elapsed);
+                }
+            }
+        } while (previousCount == Limit);
+
+        context.KeyframeData.AddRange(keyframes);
+        context.SaveChanges();
+        transaction.Commit();
```
ferferga, line 108: It's true that in BaseItemRepository's DeleteItem everything is wrapped in a transaction as well, but afaik this is unnecessary: SaveChanges does already everything in a transaction, transaction only make sense if you want to call SaveChanges multiple times and make all of them succeed or fail together.

Correct me if I'm wrong.

Shadowghost: You're correct, the initial implementation called save multiple times, so this was necessary. It doesn't hurt to be there though

src/Jellyfin.MediaEncoding.Hls/ScheduledTasks/KeyframeExtractionScheduledTask.cs
```diff
@@ -74,19 +73,16 @@ public Task ExecuteAsync(IProgress<double> progress, CancellationToken cancellat
...
-            var currentPageCount = videos.Count;
-            // TODO parallelize with Parallel.ForEach?
-            for (var i = 0; i < currentPageCount; i++)
+            foreach (var video in videos)
             {
-                var video = videos[i];
                 // Only local files supported
-                if (video.IsFileProtocol && File.Exists(video.Path))
+                var path = video.Path;
+                if (!string.IsNullOrEmpty(path) && File.Exists(path))
```
Bond-009, line 80: File.Exists already checks for null and empty
https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/IO/File.cs,3360368484a9f131
```suggestion
                if (File.Exists(path))
```

MediaBrowser.Controller/Persistence/IKeyframeRepository.cs
```diff
@@ -0,0 +1,31 @@
+#nullable disable
```
Bond-009, line 1: Why is this needed?

Jellyfin.Server/Migrations/Routines/MigrateKeyframeData.cs
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    private bool TryGetKeyframeData(BaseItem item, [NotNullWhen(true)] out KeyframeData? data)
+    {
+        data = null;
+        var path = item.Path;
+        if (!string.IsNullOrEmpty(path))
+        {
+            var cachePath = GetCachePath(KeyframeCachePath, path);
+            if (!string.IsNullOrEmpty(cachePath) && TryReadFromCache(cachePath, out var keyframeData))
```
Bond-009, line 122:
```suggestion
            if (TryReadFromCache(cachePath, out var keyframeData))
```

Bond-009 (changes requested):

MediaBrowser.Controller/Persistence/IKeyframeRepository.cs
```diff
@@ -0,0 +1,29 @@
...
+/// <summary>
+/// Provides methods for accessing keyframe data.
+/// </summary>
+public interface IKeyframeRepository
+{
+    /// <summary>
+    /// Gets the keyframe data.
+    /// </summary>
+    /// <param name="itemId">The item id.</param>
+    /// <returns>IEnumerable{KeyframeData}.</returns>
```
Bond-009, line 18: This should never just contain the type.
```suggestion
    /// <returns>The keyframe data.</returns>
```

Bond-009 (approved):

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/Migrations/20250327101120_AddKeyframeData.cs
```diff
@@ -0,0 +1,41 @@
...
+                },
+                constraints: table =>
+                {
+                    table.PrimaryKey("PK_KeyframeData", x => x.ItemId);
+                    table.ForeignKey(
+                        name: "FK_KeyframeData_BaseItems_ItemId",
+                        column: x => x.ItemId,
+                        principalTable: "BaseItems",
+                        principalColumn: "Id",
+                        onDelete: ReferentialAction.Cascade);
```
crobibero, line 30: I think our fk don't cascade delete?

crobibero (approved):

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/Migrations/20250327101120_AddKeyframeData.cs
Shadowghost: This is what efcore generated. It's the same for most of the other table too

PR #13776, opened 2025-03-26
Restrict GetPublicUsers to local IPs only

mateuszdrab:
**Changes**

This small PR blocks listing of Jellyfin users on public networks by returning an empty list of users. 
An alternative option would be to return 401 Unauthorized or 403 Forbidden.
This is done by adding a local IP check in GetPublicUsers which would change the behavior of Jellyfin and should be considered a breaking change. 
It is not wise to disclose any user information to the internet due to security concerns and similar features in other software such as Home Assistant got pulled due to the same exact reason.
Please let me know if this would be sufficient or if extra functionality needs to be added to provide a setting option to

**Issues**

Addresses https://github.com/jellyfin/jellyfin-web/issues/2042

### DIFF
--- a/MediaBrowser.Common/Net/NetworkConfiguration.cs
+++ b/MediaBrowser.Common/Net/NetworkConfiguration.cs
@@ -173,4 +173,14 @@ public class NetworkConfiguration
     /// Gets or sets a value indicating whether <seealso cref="RemoteIPFilter"/> contains a blacklist or a whitelist. Default is a whitelist.
     /// </summary>
     public bool IsRemoteIPFilterBlacklist { get; set; }
+
+    /// <summary>
+    /// Gets or sets a value indicating whether public user listing is enabled.
+    /// </summary>
+    public bool PublicUserListing { get; set; } = true;
+
+    /// <summary>
+    /// Gets or sets a value indicating whether public user listing is scoped to local network only.
+    /// </summary>
+    public bool PublicUserListingLocalOnly { get; set; } = true;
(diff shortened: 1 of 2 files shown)


Shadowghost:
I'd also prefer an option for it. But I'd argue it's sensible to set it to block access by default for remote connections.

mateuszdrab:
I've come up with an alternative idea where NetworkConfiguration is extended with two booleans PublicUserListing and PublicUserListingLocalOnly which are controlled by the UI through a drop down shown below.

image

Happy to push this into this branch, will also need to open a PR for the web client.

MediaBrowser.Common/Net/NetworkConfiguration.cs
```diff
@@ -173,4 +173,14 @@ public string BaseUrl
     /// Gets or sets a value indicating whether <seealso cref="RemoteIPFilter"/> contains a blacklist or a whitelist. Default is a whitelist.
     /// </summary>
     public bool IsRemoteIPFilterBlacklist { get; set; }
+
+    /// <summary>
+    /// Gets or sets a value indicating whether public user listing is enabled.
+    /// </summary>
+    public bool PublicUserListing { get; set; } = true;
```
gnattu, line 180: I prefer to have a single enum instead of two booleans for this option. `Disabled` `Local only` and `Always` describes the behavior easier than two booleans.

Bond-009:
Closing as stale 1+ year without a response

PR #13791, opened 2025-03-28
Parallelize library.db migration steps, minor fixes

ferferga:
This PR further reduces the migration time from 04:26 (achieved through #13749) to **03:45** in my Raspberry Pi by parallelizing everything that can be executed in parallel.

The downside is that this increase a little the RAM usage because all the tasks are performed as early as possible and flushing changes to the DB is deferred to the end. The biggest one though (BaseItems) is flushed to the DB as soon as possible. I don't think this will ever be a problem though, since my bottleneck now is I/O bound, and  we can assume a correlation between the library size and better I/O speeds for flushing to the database (and, as soon as changes are flushed, space is freed). My 4 GB Pi 4 could handle it with a lot of headroom.

Also took the chance to fix the logger messages [(for people entries, ``MediaStreamInfos`` count was shown instead, for example)](https://github.com/jellyfin/jellyfin/blob/master/Jellyfin.Server/Migrations/Routines/MigrateLibraryDb.cs#L275) and improve the wording (in my opinion) of the logged messages.

### DIFF
--- a/Jellyfin.Server/Migrations/Routines/MigrateLibraryDb.cs
+++ b/Jellyfin.Server/Migrations/Routines/MigrateLibraryDb.cs
@@ -82,7 +82,7 @@ internal class MigrateLibraryDb : IDatabaseMigrationRoutine
 
         // Perform a cleanup first to remove stale data from previously failed migrations.
         _logger.LogInformation("Cleaning leftovers to ensure no data from previous failed migrations exist...");
-        using var dbContext = _provider.CreateDbContext();
+        var dbContext = CreateDbContext();
         dbContext.BaseItems.ExecuteDelete();
         dbContext.ItemValues.ExecuteDelete();
         dbContext.UserData.ExecuteDelete();
@@ -131,7 +131,7 @@ internal class MigrateLibraryDb : IDatabaseMigrationRoutine
         var taskTimes = Task.WhenAll(
             Task.Run(() =>
             {
-                using var dbContext = _provider.CreateDbContext();
+                var dbContext = CreateDbContext();
                 using var connection = new SqliteConnection(dbPath);
                 connection.Open();
                 var stopwatch = new Stopwatch();
@@ -175,7 +175,7 @@ internal class MigrateLibraryDb : IDatabaseMigrationRoutine
             }),
             Task.Run(() =>
             {
-                using var dbContext = _provider.CreateDbContext();
+                var dbContext = CreateDbContext();
                 using var connection = new SqliteConnection(dbPath);
                 connection.Open();
                 var stopwatch = new Stopwatch();
@@ -226,7 +226,7 @@ internal class MigrateLibraryDb : IDatabaseMigrationRoutine
             }),
             Task.Run(() =>
             {
-                using var dbContext = _provider.CreateDbContext();
+                var dbContext = CreateDbContext();
                 using var connection = new SqliteConnection(dbPath);
                 connection.Open();
                 var stopwatch = new Stopwatch();
@@ -256,7 +256,7 @@ internal class MigrateLibraryDb : IDatabaseMigrationRoutine
             }),
             Task.Run(() =>
             {
-                using var dbContext = _provider.CreateDbContext();
+                var dbContext = CreateDbContext();
                 using var connection = new SqliteConnection(dbPath);
                 connection.Open();
                 var stopwatch = new Stopwatch();
@@ -320,7 +320,7 @@ internal class MigrateLibraryDb : IDatabaseMigrationRoutine
             }),
             Task.Run(() =>
             {
-                using var dbContext = _provider.CreateDbContext();
+                var dbContext = CreateDbContext();
                 using var connection = new SqliteConnection(dbPath);
                 connection.Open();
                 var stopwatch = new Stopwatch();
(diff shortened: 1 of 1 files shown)


Shadowghost (approved):

Jellyfin.Server/Migrations/Routines/MigrateLibraryDb.cs
```diff
@@ -97,242 +107,321 @@ public void Perform()
...
+                commitTasks.Add(() => CommitChanges(dbContext, "AncestorIds"));
 
-        _logger.LogInformation("Start moving Chapters.");
-        const string chapterQuery = """
-        SELECT ItemId,StartPositionTicks,Name,ImagePath,ImageDateModified,ChapterIndex from Chapters2
-        WHERE EXISTS(SELECT 1 FROM TypedBaseItems WHERE TypedBaseItems.guid = Chapters2.ItemId)
-        """;
-        dbContext.Chapters.ExecuteDelete();
+                return stopwatch.Elapsed;
+            })).GetAwaiter().GetResult();
```
ferferga, line 375:
```suggestion
            })).ConfigureAwait(false).GetAwaiter().GetResult();
```

Shall I use ConfigureAwait here as well?

Jellyfin.Server/Migrations/Routines/MigrateLibraryDb.cs
```diff
@@ -69,23 +70,32 @@ public MigrateLibraryDb(
...
         var migrationTotalTime = TimeSpan.Zero;
 
-        var stopwatch = new Stopwatch();
-        stopwatch.Start();
-
-        connection.Open();
-        using var dbContext = _provider.CreateDbContext();
-
+        // Perform a cleanup first to remove stale data from previously failed migrations.
+        _logger.LogInformation("Cleaning leftovers to ensure no data from previous failed migrations exist...");
```
jhparker88, line 84: "leftovers" sounds like someone left a burger on a plate instead I would state just "Cleaning data from previous migrations ..."

Jellyfin.Server/Migrations/Routines/MigrateLibraryDb.cs
```diff
@@ -69,23 +70,32 @@ public MigrateLibraryDb(
...
+        dbContext.MediaStreamInfos.ExecuteDelete();
+        dbContext.Peoples.ExecuteDelete();
+        dbContext.PeopleBaseItemMap.ExecuteDelete();
+        dbContext.Chapters.ExecuteDelete();
         migrationTotalTime += stopwatch.Elapsed;
-        _logger.LogInformation("Saving UserData entries took {0}.", stopwatch.Elapsed);
+        _logger.LogInformation("Cleaning took {0}.", stopwatch.Elapsed);
         stopwatch.Restart();
 
+        using var connection = new SqliteConnection(dbPath);
```
jhparker88, line 97: Is this the correct approach to call `SqliteConnection` directly instead of the `CreateDbContext` ?

ferferga: This is for connecting to the old database, not to the new one.

ferferga:
@JPVenson The constraints are enforced only on save. You can try to remove that `Task.Run` and add the lambda to the `commitTasks` list instead: every other foreach will go through successfully.

The only reason to do BaseItems before all the other tasks is so `baseItemIds` and `legacyBaseItemWithUserKeys` collections are ready, but if the other tasks didn't depend on those, it could be parallelizable as well (with the difference that the lambda call to ``CommitChanges`` needs to be prepended instead of appended, so it gets executed first and other ``CommitChanges`` don't get constraint violations). Another thing you can try to see this in action is to call `CommitChanges` inside any of the (People for instance), you will see how the FK constraints are violated then. But again, this is not the case here since everything is ensured to run in the correct order.

So, to sum up:

* Everything is memory-only until calling SaveChanges, so constraints apply only there. When those constraints need to be verified, they will succeed.
* Writing is safe, because we're just doing insertions, and in a sequentially manner. Another history would be to do queries in the context.
* Only one writing happens at once (I first tried to do it async as well because I was not sure if EFCore has mechanisms to gracefully handle database-locked scenarios, but I got `database locked instantly).

Essentially, what we're doing it's flushing to disk while iterating through the rest of the entries. **`Task.WhenAll` ensures the writes from the other contexts execute when the 'BaseItems' one has finished**.

As for the logging, that's a good concern and something I thought about, but didn't do anything special with it. Feedback is welcome in that regard, but imo that's not a big enough concern to drop this altogether: when a scanning is taking place, playback, transcoding events, slow API responses... Are all logged and easily distinguishable, so this is something more related with "presentation" than anything else.

Jellyfin.Server/Migrations/Routines/MigrateLibraryDb.cs
JPVenson: that should not be nessesary here as no context switching is happening

crobibero:
current master (d75216cf3a9bb3f869415bed829dca996f4ee7c9): `Migrating Library db took 00:15:51.2031214.`
this PR: `Migrating Library DB took 00:11:02.3628840.`

for ~160k base items
Tested with 5800X3D and 16GB ram

ferferga:
@crobibero @Shadowghost Can you test with this branch: https://github.com/JPVenson/jellyfin/tree/feature/fasterMigrationWithoutCache? You're the ones that are yielding the biggest differences in execution time from all the tests we've performed against this branch.

We've been discussing and it's clear that the performance gains from this PR is wildly system and library dependent (unlike the changes from #13749), but we're not sure if the gains are mostly from the removal of EFCore change tracking detection + threading (this PR) or just the former (that branch) and threading has just a placebo effect.

If @JPVenson's branch is faster in your case too, I would say we merge that branch instead of mine, since running things sequentally will still be slighty better because it's better debuggeable.

Shadowghost:
@ferferga will check later today and report back.

ferferga:
Superseded by #13809

PR #13812, opened 2025-03-30
Improve search with partial term matching

### LINKED ISSUE #1674: Search for series expects exact punctuation between words in title

TOomaAh:
**Changes**
This commit enhances the search functionality by implementing fuzzy matching for partial terms. It modifies how special characters are processed, adds support for shorter tokens in search queries, and improves the handling of numeric parts in search terms. These changes allow users to find content even with incomplete or partially typed search terms, making the search experience more forgiving and user-friendly. This is a test implementation to gather feedback from users and determine the best approach for fuzzy searching in the long term.

I've only been able to test this implementation on a small library with limited content, and the performance appears acceptable with no noticeable delays in search results. However, additional testing on larger libraries would be beneficial to fully understand the performance characteristics of these changes. Users with extensive collections may want to monitor search response times when testing this PR.

image

**Issues**
Fixes #1674

### DIFF
--- a/src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteDatabaseProvider.cs
+++ b/src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteDatabaseProvider.cs
@@ -134,6 +134,7 @@ public sealed class SqliteDatabaseProvider : IJellyfinDatabaseProvider
             using var command = connection.CreateCommand();
 
             // Drop the existing FTS5 table if it exists
+            // In the long term, these lines will need to be removed
             command.CommandText = "DROP TABLE IF EXISTS item_search_fts;";
             _ = command.ExecuteNonQuery();
             command.CommandText = "DROP TRIGGER IF EXISTS baseitem_ai_fts;";
@@ -143,7 +144,6 @@ public sealed class SqliteDatabaseProvider : IJellyfinDatabaseProvider
             command.CommandText = "DROP TRIGGER IF EXISTS baseitem_ad_fts;";
             _ = command.ExecuteNonQuery();
 
-            // Créer une table virtuelle FTS5 pour les titres et autres champs de recherche
             command.CommandText = @"
                 -- Create a virtual FTS5 table for searching
                 CREATE VIRTUAL TABLE IF NOT EXISTS item_search_fts USING FTS5(
@@ -158,6 +158,7 @@ public sealed class SqliteDatabaseProvider : IJellyfinDatabaseProvider
                 );
 
                 -- Create triggers to keep the FTS5 table in sync with the BaseItems table
+
                 CREATE TRIGGER IF NOT EXISTS baseitem_ai_fts AFTER INSERT ON BaseItems BEGIN
                     INSERT INTO item_search_fts(id, clean_name, original_title, tags, genres)
                     VALUES (
@@ -171,7 +172,6 @@ public sealed class SqliteDatabaseProvider : IJellyfinDatabaseProvider
                     );
                 END;
 
-                -- Create triggers to keep the FTS5 table in sync with the BaseItems table
                 CREATE TRIGGER IF NOT EXISTS baseitem_au_fts AFTER UPDATE ON BaseItems BEGIN
                     UPDATE item_search_fts
                     SET
@@ -191,7 +191,7 @@ public sealed class SqliteDatabaseProvider : IJellyfinDatabaseProvider
 
             _ = command.ExecuteNonQuery();
 
-            // Initialiser la table avec les données existantes
+            // Initialize the FTS5 table with existing data
             command.CommandText = @"
                 INSERT INTO item_search_fts(id, clean_name, original_title, episode_title, path, tags, genres)
                 SELECT
--- a/src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteDatabaseProvider.cs
+++ b/src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteDatabaseProvider.cs
@@ -127,112 +127,92 @@ public sealed class SqliteDatabaseProvider : IJellyfinDatabaseProvider
     {
         try
         {
-            // Create a connection to register our custom functions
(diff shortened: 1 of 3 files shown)


jhparker88:
Instead of generating Ngrams in a subroutine as basis for the fuzzy search, creating additional index fields, and inject methods via `CreateFunction` I would rather see that for string matches on title and original title (+ tags and genres) the FTS-5 is used instead which comes with its own native "trigram tokenizer" [1] and bm25 raking. Doing a MATCH on a FTS-5 virtual table will always be faster than a LIKE select on a standard table especially for larger libraries.

[1] https://sqlite.org/fts5.html

TOomaAh:
> Instead of generating Ngrams in a subroutine as basis for the fuzzy search, creating additional index fields, and inject methods via `CreateFunction` I would rather see that for string matches on title and original title (+ tags and genres) the FTS-5 is used instead which comes with its own native "trigram tokenizer" [1] and bm25 raking. Doing a MATCH on a FTS-5 virtual table will always be faster than a LIKE select on a standard table especially for larger libraries.
> 
> [1] https://sqlite.org/fts5.html

Thank you very much. I'll try to do that.

TOomaAh:
It's clearly not perfect, but the search works at least to a minimum.

image

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteDatabaseProvider.cs
```diff
@@ -118,4 +121,98 @@ public Task RestoreBackupFast(string key, CancellationToken cancellationToken)
...
+                -- Create a virtual FTS5 table for searching
+                CREATE VIRTUAL TABLE IF NOT EXISTS item_search_fts USING FTS5(
+                    id UNINDEXED,
+                    clean_name,
+                    original_title,
+                    episode_title,
+                    path,
+                    tags,
+                    genres,
+                    tokenize='unicode61 remove_diacritics 0 tokenchars ''0123456789'''
```
jhparker88, line 157: If you wanted to rely on the trigram [1] you would need to use tokenize="trigram" which is part of SQLite since 2020-12-01 (3.34.0) [2].

[1] https://sqlite.org/fts5.html#the_trigram_tokenizer
[1] https://sqlite.org/changes.html

TOomaAh: Yes, that's what I started to do, but the searches that only contained numbers didn't work. That's why I did it. I'll have to keep looking

jhparker88:
>     [1] https://sqlite.org/fts5.html
> Thank you very much. I'll try to do that.

A couple of quick reads on the topic include [1], [2], [3], and [4].

[1] https://andrewmara.com/blog/faster-sqlite-like-queries-using-fts5-trigram-indexes
[2] https://stackoverflow.com/questions/68282278/sqllite-fts5-substring-matching-using-trigram-tokenizer
[3] https://www.bricelam.net/2020/08/08/sqlite-fts-and-efcore.html
[4] https://github.com/VahidN/EFCoreSQLiteFTS

jhparker88:
The general good thing (in my opinion) about using FTS-5 and `MATCH` is that you could rely on the column filter [1] to extent the query condition without complex query parsing compared to using `LIKE`. It would allow for things like `title:test movie AND year:1988 AND tag:test1` (for this example one would require the title to be aggregated from clearName and original title etc., year would need to be added as a column) to work nearly out of the box and when no explicit column is mentioned it would use the entire index as search body.

[1] https://sqlite.org/fts5.html#fts5_column_filters

crobibero:
We just had a very large effort to move away from hand-built sql to using EF, so I don't believe we would want to immediately add back hand rolled SQLite specific queries.

jhparker88:
> We just had a very large effort to move away from hand-built sql to using EF, so I don't believe we would want to immediately add back hand rolled SQLite specific queries.

As a proof of concept given the draft status that should be fine in order to pinpoint directional intentions.

Given @JPVenson previous attempt in #12496 to provide a search interface, he (or others) may envision something else but as starting point instead of modifying the query directly in the `BaseItemRepository` create a new interface `ItemSearch` as decorator and inject it into the `BaseItemRepository` where `ItemSearch::GetIdsThatMatchTerms` returns IDs matched to a search term.

The `ItemSearchSqliteProvider` provides that actual SQLite implementation and give that things like  `EF.Functions.Match(fts, query)` aren't available according to [2, 3] the small raw SQL should be negligible here.

I'm not really familiar with EFCore but it should be possible to add the update trigger [1, 3] in the migration builder.


[1] https://stackoverflow.com/questions/55336035/is-it-possible-and-how-create-a-trigger-with-entity-framework-core
[2] https://github.com/dotnet/efcore/issues/4823
[3] https://www.bricelam.net/2020/08/08/sqlite-fts-and-efcore.html

JPVenson:
I am fully on side with cody here.
The way you are currently doing it, will not be accepted by us. I spend too much time cleaning all sqlite specific stuff to just add some more now. The way i see it is there are 3 possible solutions how full text search can be implemented:
1. Custom EF query extensions
2. Our own Builder-esk framework involving `IJellyfinDatabaseProvider`
3. A full plugin based infastructure that either extends or replaces our search

All of them have up and downsides but raw sql directly in the repositories is a no-go.

TOomaAh:
> I am fully on side with cody here.
> 
> The way you are currently doing it, will not be accepted by us. I spend too much time cleaning all sqlite specific stuff to just add some more now. The way i see it is there are 3 possible solutions how full text search can be implemented:
> 
> 1. Custom EF query extensions
> 
> 2. Our own Builder-esk framework involving `IJellyfinDatabaseProvider`
> 
> 3. A full plugin based infastructure that either extends or replaces our search
> 
> 
> 
> All of them have up and downsides but raw sql directly in the repositories is a no-go.

I may have opened the pull request prematurely but I need feedback as I'm not comfortable with efcore.

I actually intend to delete all the sql rows and transform them into efcore queries, just give me some time.

My next commit will transform all that, it's currently just a test to make sure it works.

PR #13813, opened 2025-03-31
Fix wrong audio track selected when using instant play

### LINKED ISSUE #13774: Multiple audio default tracks will play always first track
### Description of the bug

According to the specs of MKV, all language tracks that aren't special (e.g. commentary) are default for specified language.

https://www.matroska.org/technical/notes.html

> Default flag
>
> The “default track” flag is a hint for a Matroska Player indicating that a given track SHOULD be eligible to be automatically selected as the default track for a given language. If no tracks in a given language have the default track flag set, then all tracks in that language are eligible for automatic selection. This can be used to indicate that a track provides “regular service” suitable for users with default settings, as opposed to specialized services, such as commentary, hearing-impaired captions, or descriptive audio.

Unsure if this is a client thing, but based on my observation this happens with Web/JMP and instant play (not using detail view with select in between). So I thought this is default logic from Jellyfin server-side that is applied.
_Please move to web, if it is JS and URL building._

### File 1

| TrackID | Track | Lang | Default |
| -------- | ------- | -------- | ------- |
| 1 | video | `und` | `x` |
| 2 | audio | `AA` | `x` |
| 3 | audio | `BB` | `x` |
| 4 | audio commentary | `AA` | |

My preferred language is `BB`, but I will get AA

### File 2

| TrackID | Track | lang | Default |
| -------- | ------- | -------- | ------- |
| 1 | video | `und` | `x` |
| 2 | audio | `AA` | `x` |
| 3 | audio | `BB` | |
| 4 | audio commentary | `AA` | |

I will get `BB` without default flag, that should be set based on MKV specs

### Reproduction steps

1. Create a file like described
2. Instant playback (so no detail view to select the track). `Continue Watching` will also play the first audio default track

Starting any instant playback e.g. `Continue Watching`, `recently added`, from list or `Next Up` will play the first default track and not preferred language. (No matter the additional flag in settings)

Play my wanted language `BB` if it has a default flag. Multiple use can have different preferred languages.

### Jellyfin Server version

10.10.0+

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

Release (had main/master before new EFCore)

### Environment

```markdown
- OS: arch
- Linux Kernel: 6.14
- Virtualization: -
- Clients: Web or JMP
- Browser: Firefox/Chromium
- FFmpeg Version: 5,6,7 (tried mutliple)
- Playback Method: direct
- Hardware Acceleration: 
- GPU Model:
- Plugins:
- Reverse Proxy: yes
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

```shell
No error - it is wrong behavior
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

timminator:
When using "instant play" described in #13774 Jellyfin Web/JMP do not specify an audioindex in their request. This causes Jellyfin to decide on an audiotrack by the candidateaudiostreams logic in StreamBuilder.cs, which is not taking any user or last played preferences into account. The issue is also more widespread than described in Issue #13774. It also selects the wrong track when you have for example 2 english tracks. If you played it the last time for example with track 2 it will switch to track 1 when using instant play. This is not specific to the default track flag.

**Changes**
The candidateAudioStreams logic is no longer applied when the playlistitem has already an AudioStreamIndex.

Fixes #13774

### DIFF
--- a/MediaBrowser.Model/Dlna/StreamBuilder.cs
+++ b/MediaBrowser.Model/Dlna/StreamBuilder.cs
@@ -669,11 +669,17 @@ namespace MediaBrowser.Model.Dlna
             {
                 if (audioStream?.IsDefault == true)
                 {
-                    candidateAudioStreams = item.MediaStreams.Where(stream => stream.Type == MediaStreamType.Audio && stream.IsDefault).ToArray();
+                    candidateAudioStreams = item.MediaStreams
+                        .Where(stream => stream.Type == MediaStreamType.Audio && stream.IsDefault)
+                        .OrderByDescending(stream => stream.Index == audioStream?.Index)
+                        .ToArray();
                 }
                 else
                 {
-                    candidateAudioStreams = item.MediaStreams.Where(stream => stream.Type == MediaStreamType.Audio && stream.Language == audioStream?.Language).ToArray();
+                    candidateAudioStreams = item.MediaStreams
+                        .Where(stream => stream.Type == MediaStreamType.Audio && stream.Language == audioStream?.Language)
+                        .OrderByDescending(stream => stream.Index == audioStream?.Index)
+                        .ToArray();
                 }
             }
 


Shadowghost:
@timminator can you please check the failing tests

Shadowghost (approved):

timminator:
I have to say I could not fully understand why these three demo tests were failing. I think what is happening is that the old logic with candidatestreams returned the external stream as the one to be played resulting in the TranscodeReason "AudioIsExternal". With the new logic it seems like the ac3 stream is selected resulting in the TranscodeReason "AudioCodecNotSupported".
I tried way too long to replicate the demo tests with real files, but that did not really work out.
I hope it's fine like this.

tests/Jellyfin.Model.Tests/Dlna/StreamBuilderTests.cs
```diff
@@ -45,7 +45,7 @@ public class StreamBuilderTests
         [InlineData("Firefox", "mp4-h264-aac-vtt-2600k", PlayMethod.DirectPlay)] // #6450
         [InlineData("Firefox", "mp4-h264-ac3-aac-srt-2600k", PlayMethod.Transcode, TranscodeReason.AudioCodecNotSupported, "DirectStream", "HLS.mp4")] // #6450
         [InlineData("Firefox", "mp4-h264-ac3-aacDef-srt-2600k", PlayMethod.Transcode, TranscodeReason.SecondaryAudioNotSupported, "Remux", "HLS.mp4")] // #6450
-        [InlineData("Firefox", "mp4-h264-ac3-aacExt-srt-2600k", PlayMethod.Transcode, TranscodeReason.AudioIsExternal, "DirectStream", "HLS.mp4")] // #6450
```
gnattu, line 48: If you have to do something like this the streambuilder is selecting a wrong audio track. And no this is not fine.

gnattu (changes requested):
Please don't pretend the unit test is nothing. If you change the track selection logic you have to make sure the old ones still selects the intended tracks and obviously your code is now selecting a wrong track.

timminator:
I did not pretend that unit tests are nothing. I tried for hours to replicate these demo tests but I couldn't. In my testing I could not create the case "mp4-h264-ac3-aacExt-srt-2600k" in real life. That's were all three tests failed.
From my understanding, this requires an mp4 video with one soundtrack in ac3 format and one external soundtrack in aac.
The issue is, that you cannot create such an mp4 file. FFmpeg always flags at least one audio or video track in a .mp4 as default. That means I can only create "mp4-h264-ac3Def-aacExt-srt-2600k" and not the required test.
I also tried disposition and passthrough flags but I cannot create this demo test. Without that I could not debug this further.
If one of you has an idea or some thoughts on this I would appreciate them.

timminator:
@Shadowghost @gnattu I finally got it! Now I know why the tests were failing and it actually exactly reflects the behaviour change that I wanted to achieve with this PR. The json files with the example files were the missing piece, they contain also a DefaultStreamIndex:

https://github.com/jellyfin/jellyfin/blob/3fc3b04daf929d1d3a9533fc410cb77885eb2e8a/tests/Jellyfin.Model.Tests/Test%20Data/MediaSourceInfo-mp4-h264-ac3-aacExt-srt-2600k.json#L87

And with my PR this is taken into account when the web request specifies no index, as its logical to play the file with the best match determined or the last played match. This results into the AC3 file being chosen.
Before this PR this was not the case and only done based on the score and that was higher for the external stream:

Score AC3:
https://github.com/jellyfin/jellyfin/blob/3fc3b04daf929d1d3a9533fc410cb77885eb2e8a/tests/Jellyfin.Model.Tests/Test%20Data/MediaSourceInfo-mp4-h264-ac3-aacExt-srt-2600k.json#L48
Score external Stream:
https://github.com/jellyfin/jellyfin/blob/3fc3b04daf929d1d3a9533fc410cb77885eb2e8a/tests/Jellyfin.Model.Tests/Test%20Data/MediaSourceInfo-mp4-h264-ac3-aacExt-srt-2600k.json#L64

So there are two options: If we want to test if the externalstream is being chosen we could also change the defaultaudiostreamindex to 2 in the three json files or we change the tests I did.
I am looking forward to your feedback! :-)

gnattu:
> I did not pretend that unit tests are nothing. I tried for hours to replicate these demo tests but I couldn't.

Sorry if I misunderstood your intention. From my perspective it seems like you:

1. knows your change breaks unit test
2. you don't know why
3. so you just changed the unit test to make it accept the modified results

This *does* look like ignoring the unit tests to me. Modifying a unit test without knowing what it is for is something I really hate and it really triggered me. Sorry again if that sounds offensive and that might not be your intention. But please, if you don't know, just ask before you make something you are not confident of.

> The json files with the example files were the missing piece, they contain also a DefaultStreamIndex: And with my PR this is taken into account when the web request specifies no index

Sorry, but this is wrong, or at least not entirely correct. This is much more complicated than you think and this is not only about external vs internal. Your change would overwrite the complex stream selection based on preferred codec later which would cause more transcoding. And no, the AAC is not selected based on the score. Instead, it’s because those browsers can play AAC without transcoding, but it would require AC3 to be transcoded. This logic implies that when the client has no explicit stream selection, choosing a stream that doesn’t require transcoding is preferred. However, this approach doesn’t work well when the user consistently requests a stream that needs transcoding and would want the stream to be selected based on something else. The same test case for Safari would result in direct play because Safari can play AC3 (so your change is now irrelevant).

The problem here is more likely the stream selection should be explicit is now not explicit and the implicit selection result would be overwrite by the stream selector based on something else. Your preferred stream with the default index was selected already at the beginning, but the logic later overwritten this selection because the server thinks the client has no explicit stream preference.

timminator:
> Sorry, but this is wrong, or at least not entirely correct. This is much more complicated than you think and this is not only about external vs internal. Your change would overwrite the complex stream selection based on preferred codec later which would cause more transcoding.

That is actually my intended case! When there is no web request explicitly stating an index, it should fall back to the defaultstreamindex that is saved from last playing the item or determined by the mediastreamselector logic which decided which stream is most suited by the language and settings chosen by the user. Without that we have the issue I described here in my post or in Issue #13774. I think nobody wants their stream changed without asking, only to avoid transcoding...

> And no, the AAC is not selected based on the score. Instead, it’s because those browsers can play AAC without transcoding, but it would require AC3 to be transcoded. 

Yeah, after writing my essage I also got my doubts. I've looked at the GetStreamScore function before and somehow got this wrong. Good to know!

>  The same test case for Safari would result in direct play because Safari can play AC3 (so your change is now irrelevant).

I'm not sure what you mean, when I change the defaultstreamindex in the .json to 2 it now plays the aacExt as I intended (for this change that test needs to be updated).

> The problem here is more likely the stream selection should be explicit is now not explicit and the implicit selection result would be overwritten by the stream selector based on something else. Your preferred stream with the default index was selected already at the beginning, but the logic later overwritten this selection because the server thinks the client has no explicit stream preference.

Exactly! That was my long troubleshooting session until I finally found the place, where its being overwritten! But as I said in the beginning of my post I don't think this is the right thing to do. I understand the point of avoiding transcoding completely, but not when it results in wrong audiostreams being selected that do not match the previous selection or the user settings.

So, where do we go from here? What are your thoughts to solve this problem?

gnattu:
> I'm not sure what you mean, when I change the defaultstreamindex in the .json to 2 it now plays the aacExt as I intended (for this change that test needs to be updated).

What I meant is that the fact that you don't have to change the unit test to let safari pick AC3 stream describes the current behavior to avoid transcoding.

> When there is no web request explicitly stating an index, it should fall back to the defaultstreamindex that is saved from last playing the item or determined by the mediastreamselector logic which decided which stream is most suited by the language and settings chosen by the user.

We cannot simplify things like this because the index left in the default does not always make sense especially when we consider that there are more than one client implementations and such behavior change would just pick the wrong stream that will enforce transcoding. 

> I think nobody wants their stream changed without asking, only to avoid transcoding...

The thing is, the client told the server that there is no explicit audio track which means, there is no "changed without asking" because there is nothing to be changed. Which one the server will pick this is totally up to the server, and we have surprisingly amount of users that hate transcoding. You change would pick the track that needs to be transcoded even if there do exist another track that is in the exactly the same language that does not need to be transcoded like the ac3 vs aac case in our unit test. Your change would also change the behavior when selecting based on channel limit as well, which also requires stream reselect. Your change is broken because it basically disabled stream reselect for every single file as each and every file would have a default audio track index and that may or may not be the best one for the client.

> So, where do we go from here? What are your thoughts to solve this problem?

The StreamBuilder has always been a pain to maintain because any modification to it would potentially cause unintended consequences. We cannot just say the default index is always the best choice for the client because that one is not. In a lot of cases it will just select the first stream and that is not always the best. The problem we are having here is that the default index set by user and the default index calculated by the server is passed to the StreamBuilder in the same field, which makes it impossible for StreamBuilder to tell should it really perform the stream reselect and to fix that we need to:

- Add a new field which is a flag to tell the StreamBuilder that whether the currently picked index is set by user, or calculated based on language preferences, or is just a fallback value when there is no user input
- Use that field, to enable stream reselect conditionally. This means, when the index is set by user, we do nothing, wen the index is calculated based on language preferences, we only allow reselect in the preferred language, when the index is only a fallback, we do what we used to do here to perform a full reselect.

This is much more involved but that should break the current behavior less and that would not require to change any of the current unit tests. Ideally new unit tests should be added to test the user preference cases as they are currently missing now.

timminator:
I understand all points that you have made. But your solution would not work, if a user starts a movie for the first time via instant play, which I presume will occur quite more often than resuming it, because in that case the index set by the user will be empty.

I had a different idea, that does not require a lot of changes, requires no changes to any unit tests whatsoever and still solves #13774 and the issues I mentioned. I updated my PR again. All we are doing now is moving the audiostream we determined before from defaultaudiostreamindex to the top of the candidateAudioStreams list. That means if it is fullfilling the requirements, it will be chosen. The StreamSelectionLogic for Transcoding is still applied. So if no audiostream is specified via web, the defaultaudiostreamindex is now considered favourably but it is not fixed as can be seen by the three unit tests I modified yesterday, where with this logic in place it is still switching to the aacExt track.

What do you think of this @gnattu and @Shadowghost?

gnattu:
> But your solution would not work

No it will work. It works on my setup. See #13832

The index is not only calculated base on previous item, it will also be calculated based on user preferences. If the `StreamBuilder` get enough information it can make much better decisions without using stream ordering hacks.

timminator:
I see, you added language filtering for that situation aswell. Looks great!

PR #13828, opened 2025-04-01
Trim library names

JPVenson:
Remove leading and tailing spaces from library names as they are used for on-disk storage which can cause issues.

#13429

### DIFF
--- a/Emby.Server.Implementations/Library/LibraryManager.cs
+++ b/Emby.Server.Implementations/Library/LibraryManager.cs
@@ -2899,7 +2899,7 @@ namespace Emby.Server.Implementations.Library
                 throw new ArgumentNullException(nameof(name));
             }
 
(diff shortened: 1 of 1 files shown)


Bond-009 (approved):

gnattu:
Wouldn't this make the already affected folder path invalid? Leading or trailing spaces are not always invalid on all filesystems and some filesystems on Linux does allow that, and our server used to allow that which means there could already be some library with name has a trailing or leading space, and a folder named with a leading or trailing space is already created on the filesystem. If we really want to get rid of this we need a migration to rename the already existing folder as well I think.

gnattu (approved):
I really don't think there are many people with fancy idea to put a space though, so such behavior could be fine.

crobibero (approved):

PR #13832, opened 2025-04-02
Only reselect audio streams when user preference is respected

### LINKED ISSUE #13774: Multiple audio default tracks will play always first track

gnattu:
The old logic will prefer the stream don't need transcoding with a very high priority. That behavior will mistakenly pick an audio stream in a language that is not preferred. We need to respect user preference in the StreamBuilder to limit the audio stream candidates.

**Changes**

~~A new `DefaultAudioIndexSource` field is added to `MediaSourceInfo`. This is exposed to the API as the server-side clone is done by json serialization and deserialization which will reset all `JsonIgnore` fields. If there are any better way to survive a deep clone and keep this field server side only, please let me know.~~

This field is now copied explicitly and is marked with `JsonIgnore`.

**Issues**

Closes #13813

Fixes #13774

### DIFF
--- a/Emby.Server.Implementations/Library/MediaSourceManager.cs
+++ b/Emby.Server.Implementations/Library/MediaSourceManager.cs
@@ -427,6 +427,7 @@ namespace Emby.Server.Implementations.Library
                 if (source.MediaStreams.Any(i => i.Type == MediaStreamType.Audio && i.Index == index))
                 {
                     source.DefaultAudioStreamIndex = index;
+                    source.DefaultAudioIndexSource = AudioIndexSource.User;
                     return;
                 }
             }
@@ -434,6 +435,15 @@ namespace Emby.Server.Implementations.Library
             var preferredAudio = NormalizeLanguage(user.AudioLanguagePreference);
 
             source.DefaultAudioStreamIndex = MediaStreamSelector.GetDefaultAudioStreamIndex(source.MediaStreams, preferredAudio, user.PlayDefaultAudioTrack);
+            if (user.PlayDefaultAudioTrack)
+            {
+                source.DefaultAudioIndexSource |= AudioIndexSource.Default;
+            }
+
+            if (preferredAudio.Count > 0)
+            {
+                source.DefaultAudioIndexSource |= AudioIndexSource.Language;
+            }
         }
 
         public void SetDefaultAudioAndSubtitleStreamIndices(BaseItem item, MediaSourceInfo source, User user)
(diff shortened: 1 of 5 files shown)


timminator:
Tested it! I could not produce any situations where it did not behave as intended. Thanks gnattu!

Shadowghost (approved):

crobibero (approved):

timminator:
@enter-a-random-username I think the second issue you are describing is already mentioned in #10829 starting from this [comment](https://github.com/jellyfin/jellyfin/issues/10829#issuecomment-2082850977).

PR #13835, opened 2025-04-03
Backport 10.11 API enum changes

nielsvanvelzen:
**Changes**

This PR backports #13821 and part of #13277 to use the "Unknown" enum member as default value for all non-nullable properties in the API using an enum that has an Unknown member.

This is specifically needed in a 10.10 backport so a client build with the 10.10.7 API can still read responses from a 10.11 server now with the breaking API changes from #13277.

**Issues**

### DIFF
--- a/MediaBrowser.Model/Dto/BaseItemDto.cs
+++ b/MediaBrowser.Model/Dto/BaseItemDto.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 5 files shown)


nielsvanvelzen:
Do **not** apply stable backport label to this PR. These changes are already in master!

crobibero (approved):

JPVenson (approved):

Bond-009 (approved):

PR #13839, opened 2025-04-04
Recognize file changes and remove data on change

Shadowghost:
Up until now you could replace a file with a file named the same and Jellyfin would only refresh it under certain circumstances without pruning already extracted data. But this file could (and likey does) have different subtitles and attachments and potentially even has a different resolution or runtime, so trickplay files might be off too.

**Changes**
* Properly update and check `DateModified`
* Add migration to set `BaseItem.DateModified` and check `BaseItem.Size` and fix them if necessary
* Add validation logic to `MetadataManager.BeforeSave` to check for filesystem differences:
  * Always update `DateModified`, `Size` and potentially `DateCreated` (depending on server setting) on scan
  * Remove extracted trickplay images, subtitles and attachments if file changed
  * DVD and BD directories use the sizes returned by the probers while other items use the ones reported by the file system

**Remarks**
We use the `LastWriteTime` which corresponds with `mtime` on Linux and is reliable on Windows. Contrary to that creation time  [is not reliable on Windows due to filesystem tunneling ](https://superuser.com/questions/1804174/when-the-storage-device-is-formatted-with-ntfs-the-file-tunneling-system-does-n),


Emby.Server.Implementations/Library/PathManager.cs
```diff
@@ -64,10 +65,29 @@ public string GetSubtitlePath(string mediaSourceId, int streamIndex, string exte
     /// <inheritdoc />
     public string GetTrickplayDirectory(BaseItem item, bool saveWithMedia = false)
     {
-        var id = item.Id.ToString("D", CultureInfo.InvariantCulture).AsSpan();
+        var id = item.Id.ToString("D", CultureInfo.InvariantCulture);
```
crobibero, line 68: Looks like this was reverted :)

MediaBrowser.Providers/Manager/MetadataService.cs
```diff
@@ -304,6 +316,43 @@ protected virtual ItemUpdateType BeforeSaveInternal(TItemType item, bool isFullR
...
+            {
+                var creationDate = FileSystem.GetCreationTimeUtc(itemPath);
+                var itemFileCreationDate = item.DateCreatedFilesystem;
+                if (itemFileCreationDate != creationDate)
+                {
+                    if (item is Video)
+                    {
+                        Logger.LogDebug("File creation date changed from {Then} to {Now}: {Path}", itemFileCreationDate, creationDate, itemPath);
+
+                        var itemId = item.Id.ToString("N", CultureInfo.InvariantCulture);
```
crobibero, line 331:
```suggestion
```

Unused

MediaBrowser.Providers/Manager/MetadataService.cs
```diff
@@ -304,6 +316,43 @@ protected virtual ItemUpdateType BeforeSaveInternal(TItemType item, bool isFullR
...
+                        var validPaths = PathManager.GetExtractedDataPaths(item).Where(Directory.Exists).ToList();
```
crobibero, line 332: Whether the directory exists is also being checked in the loop below

MediaBrowser.Controller/Entities/BaseItem.cs
```diff
@@ -1408,7 +1415,13 @@ public string GetPresentationUniqueKey()
 
         public virtual bool RequiresRefresh()
         {
-            return false;
+            if (string.IsNullOrEmpty(Path) || DateLastModifiedFilesystem is null)
+            {
+                return false;
+            }
+
+            var creationTimeUtc = File.GetCreationTimeUtc(Path);
```
crobibero, line 1423: Should this be `GetLastWriteTimeUtc`?

tam1m:
Just to give some user input. I'd be a bit hesitant about relying on filesystem timestamps to detect actual changes. These can be pretty inconsistent across environments, especially with things like network shares or users simply copying files around without actually modifying them.

Have you considered using filesize instead? For mediafiles, the filesize alone should give pretty high confidence about whether a file has changed or not.

I'd argue the risk of two different files (different versions of a movie or episode) having the exact same size is extremely low, and this would help avoid a lot of unnecessary re-processing, particularly with trickplay generation.

Would love to hear your thoughts on this.

Shadowghost:
@tam1m should now be taken care of.

crobibero (approved):

crobibero (approved):

Emby.Server.Implementations/Library/LibraryManager.cs
```diff
@@ -472,6 +472,33 @@ public void DeleteItem(BaseItem item, DeleteOptions options, BaseItem parent, bo
...
+                Person => _configurationManager.ApplicationPaths.PeoplePath,
+                Studio => _configurationManager.ApplicationPaths.StudioPath,
+                Year => _configurationManager.ApplicationPaths.YearPath,
+                _ => null
+            };
+
+            var itemPath = item.Path;
+            if (!string.IsNullOrEmpty(pathToCheck) && !string.IsNullOrEmpty(itemPath))
+            {
+                return itemPath.StartsWith(pathToCheck, StringComparison.OrdinalIgnoreCase);
```
cvium, line 496: I don't know if I would trust this comparison. I believe we already have this implemented somewhat correctly somewhere. Unix is also case sensitive.

MediaBrowser.Controller/Entities/BaseItem.cs
```diff
@@ -1408,6 +1408,22 @@ public string GetPresentationUniqueKey()
...
+
+            var info = FileSystem.GetFileSystemInfo(Path);
+            if (info.Exists)
+            {
+                if (info.IsDirectory)
+                {
+                    return info.LastWriteTimeUtc != DateModified;
+                }
+
+                return info.LastWriteTimeUtc != DateModified && info.Length != (Size ?? 0);
```
cvium, line 1424: Can we be sure that LastWriteTime is always "correct" i.e. will all filesystems report it?

MediaBrowser.Providers/Manager/MetadataService.cs
```diff
@@ -304,6 +315,52 @@ protected virtual ItemUpdateType BeforeSaveInternal(TItemType item, bool isFullR
                 updateType |= ItemUpdateType.MetadataImport;
             }
 
+            // Cleanup extracted files if source file was modified
+            var itemPath = item.Path;
+            if (!string.IsNullOrEmpty(itemPath))
+            {
+                var info = FileSystem.GetFileSystemInfo(itemPath);
```
cvium, line 322: Can't item path be a folder if it's a dvd or bluray?

MediaBrowser.Controller/Entities/BaseItem.cs
Shadowghost: We checked it on NTFS, ext4, ZFS, samba and NFS and it worked correctly on all of them

MediaBrowser.Providers/Manager/MetadataService.cs
Shadowghost: Yes, but that's handled properly by `FileSystem.GetFileSystemInfo` and later on we only do the size check if it is not a DVD or BD (those are always folders).

MediaBrowser.Providers/MediaInfo/ProbeProvider.cs
```diff
@@ -137,9 +137,9 @@ public bool HasChanged(BaseItem item, IDirectoryService directoryService)
                 if (!string.IsNullOrWhiteSpace(path) && item.IsFileProtocol)
                 {
                     var file = directoryService.GetFile(path);
-                    if (file is not null && file.LastWriteTimeUtc != item.DateModified)
+                    if (file is not null && file.LastWriteTimeUtc != item.DateModified && file.Length != item.Size)
```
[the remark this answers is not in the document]
Shadowghost: That won't affect filesize, metadata is saved separately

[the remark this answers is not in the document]
Bond-009: I'm talking about ID3 tags or MKV metadata.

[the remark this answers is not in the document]
Shadowghost: Since those are file modifications which do change the size in most cases, we should be good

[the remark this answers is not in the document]
Bond-009: What's the point of adding this condition? IMO it just increases the chance of false negatives.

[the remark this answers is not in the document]
Shadowghost: I don't have a preference, this was mainly based on @tam1m input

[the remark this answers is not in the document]
tam1m: TBH, I forgot about file metadata changes when I made the suggestion. That's a good point. I guess the question is how often do minimal changes like that (single character or similar) actually happen in practice. I don’t have a strong opinion on that as I personally rarely ever change file metadata. 

That said my concern is still that removing the filesize condition entirely and relying solely on timestamps brings back the original issue I described above. This would lead to a lot of unnecessary rescans, with trickplay generation of course being the biggest problem since it costs so much time and resources.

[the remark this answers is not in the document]
tam1m: What about a middleground? If the timestamp changes but the filesize stays the same, skip regenerating expensive stuff like trickplay, chapter images, and maybe subtitle extraction by default, and just refresh the lightweight parts. You can still force a full refresh via the Refresh Metadata dialog if necessary, for whatever reason.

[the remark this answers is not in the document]
Shadowghost: That's what I'm currently doing, we're always setting the `DateModified` and trigger a metadata update, but only on size change we actually remove extracted data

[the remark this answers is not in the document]
Bond-009: Sounds like a great middle ground :+1:

theguymadmax:
Tested on Windows 11: Extracted attachments (subtitles and fonts) are removed on a re-scan when the original media is replaced with a different source. If the same media file is copied over, the attachments are retained.

Emby.Server.Implementations/Library/KeyframeManager.cs
```diff
@@ -0,0 +1,44 @@
...
+    /// <param name="repository">The keyframe repository.</param>
+    public KeyframeManager(IKeyframeRepository repository)
+    {
+        _repository = repository;
+    }
+
+    /// <inheritdoc />
+    public IReadOnlyList<KeyframeData> GetKeyframeData(Guid itemId)
+    {
+        return _repository.GetKeyframeData(itemId);
```
JPVenson, line 30: maybe a good idea to work with DTOs here?

Jellyfin.Server/Migrations/Routines/RefreshInternalDateModified.cs
```diff
@@ -0,0 +1,139 @@
...
+using MediaBrowser.Model.IO;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.Logging;
+
+namespace Jellyfin.Server.Migrations.Routines;
+
+/// <summary>
+/// Migration to re-read creation dates for library items with internal metadata paths.
+/// </summary>
+public class RefreshInternalDateModified : IDatabaseMigrationRoutine
```
JPVenson, line 22: with unified migration now merged, use the attribute instead

JPVenson (changes requested):

Emby.Server.Implementations/Library/KeyframeManager.cs
Shadowghost: This is already a DTO, not the Entity

MediaBrowser.Providers/Manager/MetadataService.cs
```diff
@@ -303,6 +323,55 @@ protected virtual ItemUpdateType BeforeSaveInternal(TItemType item, bool isFullR
...
+                var modificationDate = info.LastWriteTimeUtc;
+                var itemLastModifiedFileSystem = item.DateModified;
+                if (info.Exists && itemLastModifiedFileSystem != modificationDate)
```
filipini, line 333: Why are we doing this? There's already an option to replace Metadata for specific folders/libraries/content.
This way there is no control over it and it just prunes/purges entire libraries on every scan in certain configurations (like mine)

Shadowghost: Because if a file is modified, certain data used by jellyfin is not potentially valid anymore:
- chapters
- trick play images
- key frames
- and some more

And since we don't have an easy way to validate this data, we opted for pruning it and regenerating it to make sure it is accurate.

If your data setup doesn't properly expose creation and modification date, that's on you. It's the smallest single denominator we can operate on.

filipini: But that's exactly what I mean, if a file is modified it's usually done pro-actively and then most people will already know they should Refresh/Replace existing Metadata, and well this way there is no control over what happens with the metadata. All I am saying that behavior that might trigger a refresh of 2+ days of trickplay image generation might benefit from a on/off switch.

PR #13855, opened 2025-04-06
Optimize migrations

Shadowghost:
Directly access the database to fetch the necessary data. This speeds up the migrations significantly.

**Changes**
* Optimize Directory.CreateDirectory usage
* Directly access database within migrations

### DIFF
--- a/Jellyfin.Server/Migrations/Routines/MigrateKeyframeData.cs
+++ b/Jellyfin.Server/Migrations/Routines/MigrateKeyframeData.cs
@@ -62,14 +62,15 @@ public class MigrateKeyframeData : IDatabaseMigrationRoutine
         var sw = Stopwatch.StartNew();
 
         using var context = _dbProvider.CreateDbContext();
-        var records = context.BaseItems.Where(b => b.MediaType == MediaType.Video.ToString() && !b.IsVirtualItem && !b.IsFolder).Count();
+        var baseQuery = context.BaseItems.Where(b => b.MediaType == MediaType.Video.ToString() && !b.IsVirtualItem && !b.IsFolder).OrderBy(e => e.Id);
+        var records = baseQuery.Count();
         _logger.LogInformation("Checking {Count} items for importable keyframe data.", records);
 
         context.KeyframeData.ExecuteDelete();
         using var transaction = context.Database.BeginTransaction();
         do
         {
-            var results = context.BaseItems.Where(b => b.MediaType == MediaType.Video.ToString() && !b.IsVirtualItem && !b.IsFolder).OrderBy(e => e.Id).Skip(offset).Take(Limit).Select(b => new Tuple<Guid, string?>(b.Id, b.Path)).ToList();
+            var results = baseQuery.Skip(offset).Take(Limit).Select(b => new Tuple<Guid, string?>(b.Id, b.Path)).ToList();
             foreach (var result in results)
             {
                 if (TryGetKeyframeData(result.Item1, result.Item2, out var data))
--- a/Jellyfin.Server/Migrations/Routines/MigrateKeyframeData.cs
+++ b/Jellyfin.Server/Migrations/Routines/MigrateKeyframeData.cs
@@ -62,17 +62,17 @@ public class MigrateKeyframeData : IDatabaseMigrationRoutine
         var sw = Stopwatch.StartNew();
 
         using var context = _dbProvider.CreateDbContext();
-        var records = context.BaseItems.Where(b => b.MediaType == MediaType.Video.ToString() && !b.IsVirtualItem).Count();
+        var records = context.BaseItems.Where(b => b.MediaType == MediaType.Video.ToString() && !b.IsVirtualItem && !b.IsFolder).Count();
         _logger.LogInformation("Checking {Count} items for importable keyframe data.", records);
 
         context.KeyframeData.ExecuteDelete();
         using var transaction = context.Database.BeginTransaction();
         do
         {
-            var items = context.BaseItems.Where(b => b.MediaType == MediaType.Video.ToString() && !b.IsVirtualItem).OrderBy(e => e.Id).Skip(offset).Take(Limit).ToList();
-            foreach (var item in items)
+            var results = context.BaseItems.Where(b => b.MediaType == MediaType.Video.ToString() && !b.IsVirtualItem && !b.IsFolder).OrderBy(e => e.Id).Skip(offset).Take(Limit).Select(b => new Tuple<Guid, string?>(b.Id, b.Path)).ToList();
+            foreach (var result in results)
             {
-                if (TryGetKeyframeData(item, out var data))
+                if (TryGetKeyframeData(result.Item1, result.Item2, out var data))
                 {
                     itemCount++;
                     context.KeyframeData.Add(data);
@@ -94,10 +94,9 @@ public class MigrateKeyframeData : IDatabaseMigrationRoutine
         }
     }
 
-    private bool TryGetKeyframeData(BaseItemEntity item, [NotNullWhen(true)] out KeyframeData? data)
+    private bool TryGetKeyframeData(Guid id, string? path, [NotNullWhen(true)] out KeyframeData? data)
     {
         data = null;
-        var path = item.Path;
         if (!string.IsNullOrEmpty(path))
         {
             var cachePath = GetCachePath(KeyframeCachePath, path);
@@ -105,7 +104,7 @@ public class MigrateKeyframeData : IDatabaseMigrationRoutine
             {
                 data = new()
                 {
-                    ItemId = item.Id,
+                    ItemId = id,
                     KeyframeTicks = keyframeData.KeyframeTicks.ToList(),
                     TotalDuration = keyframeData.TotalDuration
                 };
--- a/Jellyfin.Server/Migrations/Routines/MigrateKeyframeData.cs
+++ b/Jellyfin.Server/Migrations/Routines/MigrateKeyframeData.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
@@ -12,8 +11,6 @@ using Jellyfin.Database.Implementations.Entities;
 using Jellyfin.Extensions.Json;
 using MediaBrowser.Common.Configuration;
 using MediaBrowser.Common.Extensions;
-using MediaBrowser.Controller.Entities;
-using MediaBrowser.Controller.Library;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Logging;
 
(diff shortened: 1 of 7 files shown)


Jellyfin.Server/Migrations/Routines/MigrateKeyframeData.cs
```diff
@@ -63,48 +56,36 @@ public MigrateKeyframeData(
...
-
         do
         {
-            var result = _libraryManager.GetItemsResult(itemsQuery);
-            _logger.LogInformation("Importing keyframes for {Count} items", result.TotalRecordCount);
-
-            var items = result.Items;
-            previousCount = items.Count;
-            offset += Limit;
+            var items = context.BaseItems.FromSql($"SELECT * FROM [BaseItems] WHERE MediaType = 'Video' AND IsVirtualItem = 0 ORDER BY Id LIMIT {Limit} OFFSET {offset}").ToList();
```
crobibero, line 72: Is raw sql that much faster then using linq to query?

```
var items = context.BaseItems
                .Where(i => i.MediaType == "Video" && i.IsVirtualItem == false)
                .OrderBy(i => i.Id)
                .Skip(offset)
                .Take(Limit)
                .ToList();
```

Shadowghost: I forgot to push, this is already changed.

Jellyfin.Server/Migrations/Routines/MigrateKeyframeData.cs
```diff
@@ -63,48 +56,36 @@ public MigrateKeyframeData(
...
+            }
 
-                if (++itemCount % 10_000 == 0)
-                {
-                    context.KeyframeData.AddRange(keyframes);
-                    keyframes.Clear();
-                    _logger.LogInformation("Imported keyframes for {Count} items in {Time}", itemCount, sw.Elapsed);
-                }
+            offset += items.Count;
+            if (offset % 5_000 == 0)
```
crobibero, line 83: This check can be removed to always log since the limit is also 5k (repeat for other routines updated)

crobibero (approved):

Jellyfin.Server/Migrations/Routines/MigrateKeyframeData.cs
```diff
@@ -63,48 +56,33 @@ public MigrateKeyframeData(
...
-        var itemsQuery = new InternalItemsQuery
-        {
-            MediaTypes = [MediaType.Video],
-            SourceTypes = [SourceType.Library],
-            IsVirtualItem = false,
-            IsFolder = false
-        };
 
         using var context = _dbProvider.CreateDbContext();
+        var records = context.BaseItems.Where(b => b.MediaType == MediaType.Video.ToString() && !b.IsVirtualItem).Count();
```
jhparker88, line 65: Does this require `&& !b.IsFolder` given the original query contained `IsFolder = false`?

Jellyfin.Server/Migrations/Routines/MoveExtractedFiles.cs
```diff
@@ -68,51 +68,34 @@ public MoveExtractedFiles(
...
         Directory.CreateDirectory(AttachmentCachePath);
-
-        itemsQuery.EnableTotalRecordCount = false;
         do
         {
-            itemsQuery.StartIndex = offset;
-            var result = _libraryManager.GetItemsResult(itemsQuery);
-
-            var items = result.Items;
+            var items = context.BaseItems.Where(b => b.MediaType == MediaType.Video.ToString() && !b.IsVirtualItem && !b.IsFolder).OrderBy(e => e.Id).Skip(offset).Take(Limit).ToList();
```
cvium, line 85: Wouldn't it be more efficient to join MediaStreamInfos instead of querying once per baseitem?

Jellyfin.Server/Migrations/Routines/MigrateKeyframeData.cs
```diff
@@ -63,48 +56,33 @@ public MigrateKeyframeData(
...
-
         do
         {
-            var result = _libraryManager.GetItemsResult(itemsQuery);
-            _logger.LogInformation("Importing keyframes for {Count} items", result.TotalRecordCount);
-
-            var items = result.Items;
-            previousCount = items.Count;
-            offset += Limit;
+            var items = context.BaseItems.Where(b => b.MediaType == MediaType.Video.ToString() && !b.IsVirtualItem).OrderBy(e => e.Id).Skip(offset).Take(Limit).ToList();
```
cvium, line 72: Instead of checking all BaseItems, couldn't we just read all files in the keyframe directory and then look up the respective items? This could even be batched i.e. grab X keyframe files, do `select itemid from baseitems where itemid in (insert list of ids)`, then AddRange on the keyframe data that has a corresponding baseitem

Shadowghost: We can't use the keyframe stuff as source because the filename is the MD4 of a combo out of file modification date and name ...

cvium: Right. You can just select item id and path then. There's no need to fetch all columns.

Jellyfin.Server/Migrations/Routines/MigrateKeyframeData.cs
Shadowghost: True, not sure if it limits the results though. I'll test it and add it back if necessary.

Jellyfin.Server/Migrations/Routines/MoveExtractedFiles.cs
Shadowghost: What do you mean by this? We just filter by columns which are in the BaseItemEntity?

Jellyfin.Server/Migrations/Routines/MoveExtractedFiles.cs
```diff
@@ -68,51 +68,34 @@ public MoveExtractedFiles(
...
-
-        itemsQuery.EnableTotalRecordCount = false;
         do
         {
-            itemsQuery.StartIndex = offset;
-            var result = _libraryManager.GetItemsResult(itemsQuery);
-
-            var items = result.Items;
-            foreach (var item in items)
+            var results = context.BaseItems.Where(b => b.MediaType == MediaType.Video.ToString() && !b.IsVirtualItem && !b.IsFolder).OrderBy(e => e.Id).Skip(offset).Take(Limit).Select(b => new Tuple<Guid, string?>(b.Id, b.Path)).ToList();
```
JPVenson, line 85: I think that relates to L124 where you requery the media streams.


```suggestion
            var results = context.BaseItems.Include(e => e.MediaStreamInfos.Where(s=> s.StreamType == MediaStreamTypeEntity.Subtitle && !s.IsExternal)).Where(b => b.MediaType == MediaType.Video.ToString() && !b.IsVirtualItem && !b.IsFolder).OrderBy(e => e.Id).Skip(offset).Take(Limit).Select(b => new Tuple<Guid, string?>(b.Id, b.Path)).ToList();
```

JPVenson: And then you can just access them though the BaseItemEntity directly which does save you some DB roundtrips

crobibero: @Shadowghost in case you missed this

Jellyfin.Server/Migrations/Routines/MigrateKeyframeData.cs
```diff
@@ -63,48 +56,33 @@ public MigrateKeyframeData(
...
-        var itemsQuery = new InternalItemsQuery
-        {
-            MediaTypes = [MediaType.Video],
-            SourceTypes = [SourceType.Library],
-            IsVirtualItem = false,
-            IsFolder = false
-        };
 
         using var context = _dbProvider.CreateDbContext();
+        var records = context.BaseItems.Where(b => b.MediaType == MediaType.Video.ToString() && !b.IsVirtualItem && !b.IsFolder).Count();
```
JPVenson, line 65: minor
```suggestion
        var records = context.BaseItems.Count(b => b.MediaType == MediaType.Video.ToString() && !b.IsVirtualItem && !b.IsFolder);
```

Jellyfin.Server/Migrations/Routines/MoveExtractedFiles.cs
```diff
@@ -68,51 +70,41 @@ public MoveExtractedFiles(
...
-            IsFolder = false,
-            Limit = Limit,
-            StartIndex = offset,
-            EnableTotalRecordCount = true,
-        };
-
-        var records = _libraryManager.GetItemsResult(itemsQuery).TotalRecordCount;
+
+        using var context = _dbProvider.CreateDbContext();
+        var records = context.BaseItems.Where(b => b.MediaType == MediaType.Video.ToString() && !b.IsVirtualItem && !b.IsFolder).Count();
```
JPVenson, line 79: minor
```suggestion
        var records = context.BaseItems.Count(b => b.MediaType == MediaType.Video.ToString() && !b.IsVirtualItem && !b.IsFolder);
```

JPVenson (approved):

Jellyfin.Server/Migrations/Routines/MigrateKeyframeData.cs
```diff
@@ -63,48 +56,33 @@ public MigrateKeyframeData(
...
         do
         {
-            var result = _libraryManager.GetItemsResult(itemsQuery);
-            _logger.LogInformation("Importing keyframes for {Count} items", result.TotalRecordCount);
-
-            var items = result.Items;
-            previousCount = items.Count;
-            offset += Limit;
-            foreach (var item in items)
+            var results = context.BaseItems.Where(b => b.MediaType == MediaType.Video.ToString() && !b.IsVirtualItem && !b.IsFolder).OrderBy(e => e.Id).Skip(offset).Take(Limit).Select(b => new Tuple<Guid, string?>(b.Id, b.Path)).ToList();
```
cvium, line 72: Minor thing, but you should store the "base" IQueryable in a variable so any query you make uses the same filter. As it is now, the Count and the Select filters can drift.

Bond-009 (approved):

ferferga (commented):
Why there are limits? It's faster this way or it's to reduce memory usage?

ferferga:
@Shadowghost Isn't IAsyncEnumerables better? That way we iterate over a generator (like with yield functions) instead of a full collection.

JPVenson:
IAsyncEnumerable while beeing the most memory efficient, is not the fastest though

cvium:
Using OFFSET and LIMIT for pagination isn't very fast either.

As for IAsyncEnumerable: SQLite has _no_ support for async and EFCore will not paginate for you.

JPVenson:
yes that is correct however if you do use a query that does not require the engine to seek the whole table (like using an unindexed ordering) it will only load the current dataset into memory

PR #13857, opened 2025-04-06
Add support for downloading playlists

flip-dots:
**Changes**

This PR adds support for downloading playlists. When downloading a playlist items are put into a .zip file and the structure is preserved (i.e if I have a track from /media/music/artist/album/track_1 in a playlist it will generate a structure inside the playlist.zip of media/music/artist/album/track_1). It will also generate a .m3u8 from the playlist with the correct relative paths for these files to be played in the order they are defined in the playlist and this is put in the .zip file as well.

I suspect this could be fairly easily extended to also work for albums but havent tried that yet and it might make more sense for that to be in a seperate PR.

P.S - Please dont be too harsh, I havent written any C# in years :P

**Issues**

Fixes: 
- https://github.com/jellyfin/jellyfin-media-player/issues/604
- https://github.com/jellyfin/jellyfin-android/issues/498
- https://forum.jellyfin.org/t-playlist-download-all
- https://github.com/jellyfin/jellyfin-android/issues/162
- https://features.jellyfin.org/posts/655/download-a-batch-not-only-single-media-file

### DIFF
--- a/Jellyfin.Api/Controllers/LibraryController.cs
+++ b/Jellyfin.Api/Controllers/LibraryController.cs
@@ -705,6 +705,67 @@ public class LibraryController : BaseJellyfinApiController
             await LogDownloadAsync(item, user).ConfigureAwait(false);
         }
 
+        // Quotes are valid in linux. They'll possibly cause issues here.
+        var filename = Path.GetFileName(item.Path)?.Replace("\"", string.Empty, StringComparison.Ordinal);
+        var filePath = item.Path;
+        if (item.IsFileProtocol)
+        {
+            // PhysicalFile does not work well with symlinks at the moment.
+            var resolved = FileSystemHelper.ResolveLinkTarget(filePath, returnFinalTarget: true);
+            if (resolved is not null && resolved.Exists)
+            {
+                filePath = resolved.FullName;
+            }
+        }
+
+        return PhysicalFile(filePath, MimeTypes.GetMimeType(filePath), filename, true);
+    }
+
+    /// <summary>
+    /// Exports item media.
+    /// </summary>
+    /// <param name="itemId">The item id.</param>
+    /// <response code="200">Media downloaded.</response>
+    /// <response code="404">Item not found.</response>
+    /// <returns>A <see cref="FileResult"/> containing the media stream.</returns>
+    /// <exception cref="ArgumentException">User can't download or item can't be downloaded.</exception>
+    [HttpGet("Items/{itemId}/Export")]
+    [Authorize(Policy = Policies.Download)]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    [ProducesFile("video/*", "audio/*")]
+    public async Task<ActionResult> GetExport([FromRoute, Required] Guid itemId)
+    {
+        var userId = User.GetUserId();
+        var user = userId.IsEmpty()
+            ? null
+            : _userManager.GetUserById(userId);
+        var item = _libraryManager.GetItemById<BaseItem>(itemId, user);
+        if (item is null)
+        {
+            return NotFound();
+        }
+
+        if (user is not null)
+        {
+            if (!item.CanExport(user))
+            {
+                throw new ArgumentException("Item does not support exporting");
+            }
+        }
+        else
+        {
+            if (!item.CanExport())
+            {
+                throw new ArgumentException("Item does not support exporting");
+            }
+        }
+
+        if (user is not null)
+        {
+            await LogDownloadAsync(item, user).ConfigureAwait(false);
+        }
+
         if (item.GetType() == typeof(Playlist))
         {
             Response.Headers.Append("Content-Disposition", "attachment; filename=\"Playlist.zip\"");
@@ -736,20 +797,7 @@ public class LibraryController : BaseJellyfinApiController
         }
         else
         {
-            // Quotes are valid in linux. They'll possibly cause issues here.
-            var filename = Path.GetFileName(item.Path)?.Replace("\"", string.Empty, StringComparison.Ordinal);
-            var filePath = item.Path;
-            if (item.IsFileProtocol)
-            {
-                // PhysicalFile does not work well with symlinks at the moment.
-                var resolved = FileSystemHelper.ResolveLinkTarget(filePath, returnFinalTarget: true);
-                if (resolved is not null && resolved.Exists)
-                {
-                    filePath = resolved.FullName;
-                }
-            }
-
-            return PhysicalFile(filePath, MimeTypes.GetMimeType(filePath), filename, true);
+            throw new ArgumentException("Item does not support exporting");
         }
     }
 
--- a/Jellyfin.Api/Controllers/LibraryController.cs
+++ b/Jellyfin.Api/Controllers/LibraryController.cs
@@ -713,11 +713,9 @@ public class LibraryController : BaseJellyfinApiController
             var playlistFile = playlist.GetPlaylistAsFileBytes(user);
 
             List<(string, byte[])> filesVirtual = [("Playlist.m3u8", playlistFile)];
-            List<(string, string)> filesDisk = new List<(string, string)>();
-            foreach (var playlistItem in playlist.GetChildPaths(user))
-            {
-                filesDisk.Add((playlistItem.TrimStart('/'), playlistItem));
-            }
+            List<(string, string)> filesDisk = playlist.GetChildPaths(user)
+                .Select(path => (path.TrimStart('/'), path))
+                .ToList();
 
             await WriteZipAsStream(Response.Body, filesVirtual, filesDisk).ConfigureAwait(false);
             return new EmptyResult();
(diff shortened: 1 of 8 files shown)


KGT1 (commented):
wouldnt every client just timeout the request? it should be possible to take hours creating the zip in this pr. loading entire media files into memory could easily crash the server for large playlists.

the current implementation keeps everything in memory instead of using temp files, which is really risky for server stability.

if a solution like this is really wanted, i dont really think the zip should be in the playlist folder, but rather in the cache folder and cleaned up regularly.

i think a solution that would be a little better would be something like a Queue manager, which cues up the compression and after that finishes you should be able to download the zip. this would avoid timeout issues and be more reliable.

But that would require more substansive changes + api changes. Im not even sure that this is generally a good idea, since the zip could still be multiple TiB(could be mitigable with a warning if it gets too large)

jhparker88:
When I read the title "downloading playlists" I first thought you just generate a m3u file containing the URL references of the playlist and make this available for download which could then be played by VLC etc. but actually downloading the entire content of the playlist surely needs a different approach given the potential amount of data to be handled. 

```
#EXTM3U

#EXTINF:-1, Video 1
http://jellyfin.example.com/Items/XX1/Download?api_key=YY1

#EXTINF:-1, Video 2
http://jellyfin.example.com/Items/XX2/Download?api_key=YY1

```

PS: There is a "Download All" functionality on the series/episode which may already address issues like size and/or queuing.

flip-dots:
I originally only indended for this to allow for the downloading of music playlists where the content being buffered in memory wouldnt be too bad, but when I was working on the original implementation I learned that it was possible to avoid keeping the zip file in memory, i've had a go at implementing it using the example I found [here](https://swimburger.net/blog/dotnet/create-zip-files-on-http-request-without-intermediate-files-using-aspdotnet-mvc-framework) and it seems to work.

We569:
جميل

Jellyfin.Api/Controllers/LibraryController.cs
```diff
@@ -697,10 +699,79 @@ public async Task<ActionResult> GetDownload([FromRoute, Required] Guid itemId)
...
+                var filePath = file.Item2;
+
+                var entry = archive.CreateEntry(fileName);
+                using (var entryStream = entry.Open())
+                using (var fileStream = System.IO.File.OpenRead(filePath))
+                {
+                    await fileStream.CopyToAsync(entryStream).ConfigureAwait(false);
+                }
+            }
+        }
```
JPVenson, line 774: This whole block should be in a service not in the controller directly

Jellyfin.Api/Controllers/LibraryController.cs
```diff
@@ -697,10 +699,79 @@ public async Task<ActionResult> GetDownload([FromRoute, Required] Guid itemId)
             await LogDownloadAsync(item, user).ConfigureAwait(false);
         }
 
-        // Quotes are valid in linux. They'll possibly cause issues here.
-        var filename = Path.GetFileName(item.Path)?.Replace("\"", string.Empty, StringComparison.Ordinal);
+        if (item.GetType() == typeof(Playlist))
+        {
+            Response.ContentType = "application/octet-stream";
```
JPVenson, line 704: Use a proper ActionResult type instead of manipulating the Response

Jellyfin.Api/Controllers/LibraryController.cs
```diff
@@ -697,10 +699,79 @@ public async Task<ActionResult> GetDownload([FromRoute, Required] Guid itemId)
...
+            Response.Headers.Append("Content-Disposition", "attachment; filename=\"Playlist.zip\"");
+
+            var playlist = (Playlist)item;
+            var playlistFile = playlist.GetPlaylistAsFileBytes(user);
+
+            List<Tuple<string, byte[]>> filesVirtual = [new Tuple<string, byte[]>("Playlist.m3u8", playlistFile)];
```
JPVenson, line 710: Use value tuples instead

JPVenson (changes requested):

Jellyfin.Api/Controllers/LibraryController.cs
flip-dots: I have not been able to find a way to do that. The magic of WriteZipAsStream is that it sends the data without storing it all in memory and to do that it uses the response stream, but the problem is that as soon as it is called it starts sending the data, meaning we can no longer set the headers, so if you return a File or a FileStreamResult you get an exception as data is already being sent and headers cant be edited in this state, the only thing we can return is an empty response. If you try to get around this by creating a MemoryStream and then giving that to WriteZipAsStream then finally returning that with File or FileStreamResult then you end up buffering the entire file in memory defeating the whole point of WriteZipAsStream. Im open to suggestions if anyone knows a way around this.

Jellyfin.Api/Controllers/LibraryController.cs
flip-dots: Do you have any suggestions for where in the folder structure I should put it? There are a bunch of service like things I found in Emby.Server.Implementations but since this zip stream functionality is for both albums and playlists it doesnt really fit in any of the existing catagories and its to do with the API rather than core server functionality so im not sure if it makes any sense for it to be in that folder.

nielsvanvelzen (changes requested):
Zipping up the entire playlist on the server before sending it to the client is not an efficient way that we'd support in the Android apps. We'd rather see an API that returns a list of downloads so we the app can show download progress and download multiple files simultaneously.

flip-dots:
> Zipping up the entire playlist on the server before sending it to the client is not an efficient way that we'd support in the Android apps. We'd rather see an API that returns a list of downloads so we the app can show download progress and download multiple files simultaneously.

I agree that a download manager is a much better way of handling multiple downloads in an app but that is a whole other can of worms. This PR is targeted at just the web client (though it does actually work for the Android client other than the .zip filetype being missing for some reason) and I think this approach is the best possible in a web browser. The zipping of the files is for convinience more than anything, browsers absolutly hate downloading multiple things and it allows for the folder structue to be preserved which means the playlist file will work, I have disabled compression so the performance impact on the server should be minimal (basically just as if it was downloading the files one by one), so its not that inefficent, the only inefficency I can see is that if your download gets interupted you will have to start again which would kind of suck, but its not like there is a better way to do this in a web browser.

flip-dots:
> Zipping up the entire playlist on the server before sending it to the client is not an efficient way that we'd support in the Android apps. We'd rather see an API that returns a list of downloads so we the app can show download progress and download multiple files simultaneously.

In case there was a misunderstanding here I would like to clarify that after the changes I made it no longer zips the entire playlist before sending it, the zip is sent as a stream, live, so it shouldn't be buffering anything in memory or on disk.

flip-dots:
Well that rebase didn’t seem to work...

flip-dots:
> Well that rebase didn’t seem to work...

It seems to be fixed now and the conflict has gone away.

JPVenson:
another flawless rebase :-D

nielsvanvelzen (changes requested):
My previous review still stands. We shouldn't zip up stuff server-side but instead provide the list of items to download to clients.

JPVenson:
generally i dont have anything against a zip endpoint from a pure download perspective but i am with niels that a download manifest would be much more useful from a client perspective.

flip-dots:
@nielsvanvelzen For the purposes of downloading music, movies, and shows to watch offline using an app (like you do with Netflix or YouTube), I agree that a download manager based approach would be better, but that's not what this is or what I would like implemented.

I think this PR might have been better described as exporting playlists. I want to be able to download music albums and playlists created in Jellyfin into a format which can be played on anything, not Jellyfin clients.

This functionality is nice to have as it means you don't need to re-create playlists for every platform (it gets really tedious) since pretty much all of them understand .m3u.

Here are some example devices/use cases:
- Using a phone's built-in music app for offline listening
- Using VLC on a computer for offline listening
- Using a smart watches built-in music app
- Using an iPod or other MP3 player
- Burning music/playlists to CDs 
- Using swimming headphones (Bluetooth does not work underwater)
- Using music players apps which don't support Jellyfin (e.g Sonos, WiiM)

Using some form of container (zip, tar, etc) is pretty much a requirement for this as web browsers don't like downloading many files at once and it dirties up the downloads directory (I have like 300 songs in just one playlist and downloading multiple playlists would make it hard to figure out what track belongs to which and probably crash my browser).

The way I implemented this means it also works on mobile and for things which aren't music which isn't intended and if that might cause too much confusion I am happy to implement something to disable it for that (if someone can point me in the right direction). 

The way I see it is that these are two separate complementary things of which it would be good to have both, but for the moment I just want to be able to load up my iPod with music without needing to resort to spinning up my custom fork or going through the pain of manually re-creating my playlists every time I want to put music on a new device or re-order some tracks, and I suspect this functionality would be useful for other people who want to put their music on non-supported devices.

nielsvanvelzen:
If we want to add an export feature it should use a dedicated export api, not hijack the existing download one.

flip-dots:
It took me a little while to figure out how to do it but there is now a dedicated export API and button in the web UI and it works.

PR #13874, opened 2025-04-08
Use Guid for parentPrimaryImageItemId

nielsvanvelzen:
**Changes**

Use Guid for parentPrimaryImageItemId in BaseItemDto and SeriesTimerInfoDto

**Issues**


Bond-009 (approved):

JPVenson (approved):

nielsvanvelzen:
> Shouldn't these be nullable Guids?

Yes they should, got a bit confused because the string didn't appear to be nullable (but it was because nullability is disabled for both files)

crobibero (approved):

PR #13888, opened 2025-04-09
Add Api and startup check for sufficient storage capacity

JPVenson:
Add Device check to test for sensible free space on startup.
Add api to get storage data on JF and library paths


Jellyfin.Server/Program.cs
```diff
@@ -218,6 +220,43 @@ private static async Task StartServer(IServerApplicationPaths appPaths, StartupO
...
         }
 
+        private static void TestDataDirectoryAccess(IServerApplicationPaths appPaths, ILogger logger)
+        {
+            try
+            {
+                var configTestFilePath = Path.Combine(appPaths.DataPath, "storagetest.bin");
+                if (File.Exists(configTestFilePath))
+                {
+                    logger.LogInformation("Testfile {TestFilename} seem to already exist, try deleting it.", configTestFilePath);
```
sjorge, line 230: >  try deleting it

Feels more like an instruction for whoever is reading the log.

Shadowghost (commented):
Not sure if info logging is the right level here

JPVenson:
open to ideas

TimGels (changes requested):
Logic looks alright to me. Aside from the spelling mistakes, please correct those. Kinda curious if something like: `DriveInfo.AvailableFreeSpace` has been considered.

Shadowghost (approved):

crobibero (changes requested):
This is a breaking API change.
Add new properties and mark the old ones as obsolete.

JPVenson:
@crobibero done

Jellyfin.Server.Implementations/StorageHelpers/StorageHelper.cs
```diff
@@ -0,0 +1,101 @@
...
+    /// Formats a size in bytes into a common human readable form.
+    /// </summary>
+    /// <remarks>
+    /// Taken and slightly modified from https://stackoverflow.com/a/4975942/1786007 .
+    /// </remarks>
+    /// <param name="byteCount">The size in bytes.</param>
+    /// <returns>A human readable approximate representation of the argument.</returns>
+    public static string BytesToString(long byteCount)
+    {
+        string[] suf = ["B", "KB", "MB", "GB", "TB", "PB", "EB"];
```
crobibero, line 89: this should be extracted

nielsvanvelzen (changes requested):
Considering the system info endpoint can be used by any user I propose a slightly bigger change:

1. Deprecate all current path properties in SystemInfo (make nullable and add obsolete attribute)
2. Add new GetSystemResources operation at /System/Resources to the SystemController. This is admin only and returns a SystemResourcesDto.
3. Add directories to the new SystemResourcesDto

Then in a future Jellyfin version we'll remove the old properties from SystemInfo and we can safely extend SystemResourcesDto with additional information.

Jellyfin.Server.Implementations/StorageHelpers/StorageHelper.cs
JPVenson: In my opinion this is actually the right place for that. StorageHelper should contain all the helper methods to get data about the underlying storage system and a humanizer for FileSize does fit here. But where to otherwise?

crobibero: Sorry, I meant to take the array out of the function and make it static :)

Jellyfin.Server.Implementations/StorageHelpers/StorageHelper.cs
```diff
@@ -0,0 +1,100 @@
...
+
+namespace Jellyfin.Server.Implementations.StorageHelpers;
+
+/// <summary>
+/// Contains methods to help with checking for storage and returning storage data for jellyfin folders.
+/// </summary>
+public static class StorageHelper
+{
+    private const long TwoGigabyte = 2147483647L;
+    private const long FiveHundredAndTwelveMegaByte = 536_870_911L;
```
Shadowghost, line 16: Please use the same format for both

Jellyfin.Server.Implementations/StorageHelpers/StorageHelper.cs
```diff
@@ -0,0 +1,100 @@
...
+    private static string[] _byteHumanizedSuffixes = ["B", "KB", "MB", "GB", "TB", "PB", "EB"];
+
+    /// <summary>
+    /// Tests the available storage capacity on the jellyfin paths with estimated minimum values.
+    /// </summary>
+    /// <param name="applicationPaths">The application paths.</param>
+    /// <param name="logger">Logger.</param>
+    public static void TestCommonJfPathsForStorageCapacity(IApplicationPaths applicationPaths, ILogger logger)
```
Shadowghost, line 24:
```suggestion
    public static void TestCommonPathsForStorageCapacity(IApplicationPaths applicationPaths, ILogger logger)
```

Jellyfin.Server.Implementations/StorageHelpers/StorageHelper.cs
```diff
@@ -0,0 +1,100 @@
...
+    {
+        logger.LogDebug("Check path {TestPath} for storage capacity", path);
+        var drive = new DriveInfo(path);
+        if (threshold != -1 && drive.AvailableFreeSpace < threshold)
+        {
+            throw new InvalidOperationException($"The path `{path}` exceeds the minimum required free capacity of {HumanizeStorageSize(threshold)}");
+        }
+
+        logger.LogInformation(
+            "Storage path `{TestPath}` ({StorageType}) successfully tested with {FreeSpace} left free which is over the minimum of {MinFree}.",
```
TimGels, line 73: Similar as previous comment. Consider changing tested to checked. 

Also, "{FreeSpace} left free" doesnt read well. Consider something along the lines of: "Storage path `{TestPath}` ({StorageType}) was successfully checked. Free space: {FreeSpace}, which is above the required minimum of {MinFree}."?

Jellyfin.Server.Implementations/StorageHelpers/StorageHelper.cs
```diff
@@ -0,0 +1,107 @@
...
+    /// <param name="logger">The logger.</param>
+    /// <param name="threshold">The threshold to check for or -1 to just log the data.</param>
+    /// <exception cref="InvalidOperationException">Thrown when the threshold is not available on the underlying storage.</exception>
+    private static void TestDataDirectorySize(string path, ILogger logger, long threshold = -1)
+    {
+        logger.LogDebug("Check path {TestPath} for storage capacity", path);
+        var drive = new DriveInfo(path);
+        if (threshold != -1 && drive.AvailableFreeSpace < threshold)
+        {
+            throw new InvalidOperationException($"The path `{path}` exceeds the minimum required free capacity of {HumanizeStorageSize(threshold)}");
```
TimGels, line 76: This exception message feels like it is a good thing that it exceeds a minimum something. Something like this is clearer:
```suggestion
            throw new InvalidOperationException($"The path `{path}` has insufficient free space. Required: at least {HumanizeStorageSize(threshold)}.");
```

Jellyfin.Api/Controllers/SystemController.cs
```diff
@@ -71,6 +71,19 @@ public SystemController(
...
         => _systemManager.GetSystemInfo(Request);
 
+    /// <summary>
+    /// Gets information about the server.
+    /// </summary>
+    /// <response code="200">Information retrieved.</response>
+    /// <response code="403">User does not have permission to retrieve information.</response>
+    /// <returns>A <see cref="SystemInfo"/> with info about the system.</returns>
+    [HttpGet("StorageResources")]
+    [Authorize(Policy = Policies.FirstTimeSetupOrElevated)]
```
crobibero, line 81: What is the reasoning for allowing first time setup to access this endpoint?

JPVenson: The setup UI could also benefit from displaying this information as it also is used to add libraries

nielsvanvelzen: Let's keep it admin only for now, we can always enable it for setup once jf-web is going to use it.

Jellyfin.Api/Controllers/SystemController.cs
```diff
@@ -71,6 +71,19 @@ public SystemController(
     public ActionResult<SystemInfo> GetSystemInfo()
         => _systemManager.GetSystemInfo(Request);
 
+    /// <summary>
+    /// Gets information about the server.
+    /// </summary>
+    /// <response code="200">Information retrieved.</response>
+    /// <response code="403">User does not have permission to retrieve information.</response>
+    /// <returns>A <see cref="SystemInfo"/> with info about the system.</returns>
+    [HttpGet("StorageResources")]
```
nielsvanvelzen, line 80: I would remove the "storage" part from the URL so we can also add other system information later like CPU/RAM usage (as an example, not saying we should). Maybe `/System/Info/Resources` is a good path, as we already have the /System/Info part for 2 other endpoints?

```suggestion
    [HttpGet("Info/Resources")]
```

Jellyfin.Api/Controllers/SystemController.cs
```diff
@@ -71,6 +71,19 @@ public SystemController(
...
+    [Authorize(Policy = Policies.FirstTimeSetupOrElevated)]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status403Forbidden)]
+    public ActionResult<SystemStorageInfo> GetSystemStorageInfo()
```
nielsvanvelzen, line 84:
```suggestion
    public ActionResult<SystemStorageInfo> GetSystemResources()
```
(if we rename the path)

MediaBrowser.Model/System/SystemStorageInfo.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.Collections.Generic;
+using System.ComponentModel;
+using MediaBrowser.Model.Updates;
+
+namespace MediaBrowser.Model.System;
+
+/// <summary>
+/// Contains informations about the systems storage.
+/// </summary>
+public class SystemStorageInfo
```
nielsvanvelzen, line 14: I would suggest adding a separate DTO class in case we ever add stuff here that should NOT be in API responses.

MediaBrowser.Model/System/SystemStorageInfo.cs
```diff
@@ -0,0 +1,62 @@
...
+{
+    /// <summary>
+    /// Gets or sets the program data path.
+    /// </summary>
+    /// <value>The program data path.</value>
+    public FolderStorageInfo ProgramDataDirectoryInfo { get; set; }
```
nielsvanvelzen, line 20: If all the variables in here are just information about paths the "DirectoryInfo" suffix is unneeded. (Same for all other variables too)

```suggestion
    public FolderStorageInfo ProgramData { get; set; }
```

If we chose to generalize this class (E.g. `SystemResourcesDto`) then we do want some suffix, but I think DirectoryInfo is quite long. Maybe just `Folder` or `Storage` works, also consistent with the "FolderStorageInfo" class name it uses.

```suggestion
    public FolderStorageInfo ProgramDataFolder { get; set; }
```

```suggestion
    public FolderStorageInfo ProgramDataStorage{ get; set; }
```

MediaBrowser.Model/System/SystemStorageInfo.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// <summary>
+    /// Gets or sets the transcode path.
+    /// </summary>
+    /// <value>The transcode path.</value>
+    public FolderStorageInfo TranscodingTempDirectoryInfo { get; set; }
+
+    /// <summary>
+    /// Gets or sets the storage informations of all libraries.
+    /// </summary>
+    public IReadOnlyCollection<LibraryStorageInfo> Libraries { get; set; }
```
nielsvanvelzen, line 61:
```suggestion
    public IReadOnlyCollection<LibraryStorageInfo> LibrariesDirectoryInfo { get; set; }
```

```suggestion
    public IReadOnlyCollection<LibraryStorageInfo> LibraryDirectoryInfos { get; set; }
```

```suggestion
    public IReadOnlyCollection<LibraryStorageInfo> LibraryFolders { get; set; }
```

MediaBrowser.Model/System/LibraryStorageInfo.cs
```diff
@@ -0,0 +1,30 @@
...
+
+    /// <summary>
+    /// Gets or sets the name of the library.
+    /// </summary>
+    public string LibraryName { get; set; }
+
+    /// <summary>
+    /// Gets or sets the storage informations about the folders used in a library.
+    /// </summary>
+    public IReadOnlyCollection<FolderStorageInfo> FolderDirectoryInfos { get; set; }
```
nielsvanvelzen, line 29:
```suggestion
    /// <summary>
    /// Gets or sets the Library Id.
    /// </summary>
    public Guid Id { get; set; }

    /// <summary>
    /// Gets or sets the name of the library.
    /// </summary>
    public string Name { get; set; }

    /// <summary>
    /// Gets or sets the storage informations about the folders used in a library.
    /// </summary>
    public IReadOnlyCollection<FolderStorageInfo> Folders { get; set; }
```

MediaBrowser.Model/System/FolderStorageInfo.cs
```diff
@@ -0,0 +1,32 @@
...
+
+    /// <summary>
+    /// Gets the used space of the underlying storage device of the <see cref="Path"/>.
+    /// </summary>
+    public long UsedSpace { get; init; }
+
+    /// <summary>
+    /// Gets the kind of storage device of the <see cref="Path"/>.
+    /// </summary>
+    public string? StorageType { get; init; }
```
nielsvanvelzen, line 26: Would it make sense to add an enum for this or is that too much maintenance?

nielsvanvelzen (changes requested):
Rough API review, I'm mostly thinking about future usages of this API so we can avoid breaking changes if we want to add additional resource information.

Jellyfin.Api/Controllers/SystemController.cs
JPVenson: hmm i really dont know. I would like to see those as seperate endpoints but then thinking this further this is not the most optimal api for other system resources like if you want CPU/RAM resources you would likely pull that every secound or so which is total overkill for the storage resources which will not change meaningfully in that timeframe but cost quite a bit of resources to evaluate.

I would be on board with `/System/Info/Storage` though

MediaBrowser.Model/System/FolderStorageInfo.cs
JPVenson: the source is already an string, so god knows what values we might get there and for now its really just a nice to know info.

Jellyfin.Api/Controllers/SystemController.cs
nielsvanvelzen: Sounds good, let's do `GetSystemStorage` at `/System/Info/Storage` then 👍

MediaBrowser.Model/System/FolderStorageInfo.cs
nielsvanvelzen: Ok so string it is then

JPVenson: OK slight correction, its __supposed__ to be a fixed number of values, but those also differ from OS to OS

MediaBrowser.Model/System/SystemStorageInfo.cs
```diff
@@ -0,0 +1,56 @@
...
+
+namespace MediaBrowser.Model.System;
+
+/// <summary>
+/// Contains informations about the systems storage.
+/// </summary>
+public class SystemStorageInfo
+{
+    /// <summary>
+    /// Gets or sets the program data path.
```
nielsvanvelzen, line 11: All these summaries _could_ do with some better descriptions. Some are obvious what they do but I have no clue what "ItemsByName" does for example. The descriptions are directly copied over to the OpenAPI specification and thus the ApiClients so they **are** the API documentation.

MediaBrowser.Model/System/LibraryStorageInfo.cs
```diff
@@ -0,0 +1,25 @@
...
+
+    /// <summary>
+    /// Gets or sets the name of the library.
+    /// </summary>
+    public required string Name { get; set; }
+
+    /// <summary>
+    /// Gets or sets the storage informations about the folders used in a library.
+    /// </summary>
+    public required IReadOnlyCollection<FolderStorageInfo> Folders { get; set; }
```
nielsvanvelzen, line 24: Don't have a better place to comment this, can we also add the FolderStorageInfo collection to VirtualFolderInfo? (could be in a separate PR). It would probably deprecate the existing Locations array.

MediaBrowser.Model/System/SystemStorageInfo.cs
```diff
@@ -0,0 +1,56 @@
...
+    /// </summary>
+    /// <value>The program data path.</value>
+    public required FolderStorageInfo ProgramDataStorage { get; set; }
```
nielsvanvelzen, line 14: I think these "Storage" suffixes are redundant when the class itself already implies these properties contain storage info.

nielsvanvelzen (changes requested):
Few small nitpicks but this API design looks good overall 👍

MediaBrowser.Model/System/SystemStorageInfo.cs
JPVenson: then your previous comment on the topic was very missleading^^

nielsvanvelzen: No no no read it again:

> If all the variables in here are just information about paths the "DirectoryInfo" suffix is unneeded.

I meant that a suffix would only be necessary if we also added different types of info, but we opted make this storage specific :p

MediaBrowser.Model/System/LibraryStorageInfo.cs
JPVenson: i would really like to keep that seperate as much as possible. But yes if we decide to do it, that would be a seperate PR.

MediaBrowser.Model/System/SystemStorageInfo.cs
JPVenson: Renamed to `*Folder`

JPVenson:
Does no longer __require__ web changes. Though to take advantage of the new api, web could utilise the new endpoint

JPVenson:
@crobibero api breaking changes have been removed

Jellyfin.Server.Implementations/StorageHelpers/StorageHelper.cs
```diff
@@ -0,0 +1,109 @@
...
+namespace Jellyfin.Server.Implementations.StorageHelpers;
+
+/// <summary>
+/// Contains methods to help with checking for storage and returning storage data for jellyfin folders.
+/// </summary>
+public static class StorageHelper
+{
+    private const long TwoGigabyte = 2_147_483_647L;
+    private const long FiveHundredAndTwelveMegaByte = 536_870_911L;
+    private static string[] _byteHumanizedSuffixes = ["B", "KB", "MB", "GB", "TB", "PB", "EB"];
```
crobibero, line 17:
```suggestion
    private static readonly string[] _byteHumanizedSuffixes = ["B", "KB", "MB", "GB", "TB", "PB", "EB"];
```

crobibero (approved):

PR #13892, opened 2025-04-10
Fix outdated xmltv cache

### LINKED ISSUE #6103: Changes in xmltv requires manual purge of caches
**Describe the bug**
In order to get Jellyfin to update channels and livetv guides properly, manual removal of /cache/xmltv/* and /cache/*_channels is required for the change to take effect. Honestly not sure if it's only /cache/xmltv/*, but it seems like both are necessary during my testing.

 - OS: CentOS
 - Virtualization: Docker
 - Clients: All
 - Jellyfin Version: 10.7.5
 - Reverse Proxy: nginx
 - Base URL: none
 - Networking: bridge
 - Storage: nfs

**To Reproduce**
1. Set up some channels, add tuner using m3u and xmltv guide, save and load xml.
2. Add/remove available channels (in ErsatzTV for example), change channel contents
3. Reload xml
4. Changes won't take place until caches are manually deleted, and xml reloaded.

**Expected behavior**
All data should be refreshed on xml reload

**Server Logs**
No error is logged

timminator:
If a xmltv program is added and updated afterwards changes are not showing up in Jellyfin. This is due to the parameter _maxCacheAge. Even if the guide data is manually refreshed the logic returns immediately with the old cache file instead of using the new one if the cache is not older than _maxCacheAge.
I updated this logic to compare thelastmodified time of the cache and the potentially new xml file, if the cache is not older than _maxCacheAge.
This ensures that changes are showing up in Jellyfin properly.

Changes:
Add Last-Modified check to XMLTV cache logic if cache is not older than 1 hour.

Fixes #6103

### DIFF
--- a/src/Jellyfin.LiveTv/Listings/XmlTvListingsProvider.cs
+++ b/src/Jellyfin.LiveTv/Listings/XmlTvListingsProvider.cs
@@ -66,11 +66,39 @@ namespace Jellyfin.LiveTv.Listings
 
             if (File.Exists(cacheFile) && File.GetLastWriteTimeUtc(cacheFile) >= DateTime.UtcNow.Subtract(_maxCacheAge))
             {
-                return cacheFile;
-            }
+                try
+                {
+                    if (info.Path.StartsWith("http", StringComparison.OrdinalIgnoreCase))
+                    {
+                        using var headRequest = new HttpRequestMessage(HttpMethod.Head, info.Path);
+                        using var response = await _httpClientFactory.CreateClient(NamedClient.Default)
+                            .SendAsync(headRequest, HttpCompletionOption.ResponseHeadersRead, cancellationToken)
+                            .ConfigureAwait(false);
 
-            // Must check if file exists as parent directory may not exist.
-            if (File.Exists(cacheFile))
+                        if (response.Content.Headers.LastModified.HasValue &&
+                            response.Content.Headers.LastModified.Value.UtcDateTime <= File.GetLastWriteTimeUtc(cacheFile))
+                        {
+                            return cacheFile;
+                        }
+
+                        File.Delete(cacheFile);
+                    }
+                    else if (File.Exists(info.Path))
+                    {
+                        if (File.GetLastWriteTimeUtc(info.Path) <= File.GetLastWriteTimeUtc(cacheFile))
+                        {
+                            return cacheFile;
+                        }
+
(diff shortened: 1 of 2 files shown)


src/Jellyfin.LiveTv/Listings/XmlTvListingsProvider.cs
```diff
@@ -66,11 +66,39 @@ private async Task<string> GetXml(ListingsProviderInfo info, CancellationToken c
...
 
-            // Must check if file exists as parent directory may not exist.
-            if (File.Exists(cacheFile))
+                        File.Delete(cacheFile);
+                    }
+                }
+                catch (Exception)
+                {
+                    File.Delete(cacheFile);
+                }
```
crobibero, line 99:
```suggestion
                catch (Exception ex)
                {
                    _logger.LogWarning(ex, "Unable to get last modified time for {Path}", info.Path);
                    File.Delete(cacheFile);
                }
```

Shouldn't just silently ignore the exception or we'll never know if there was an issue with this code

crobibero:
I'm not sure if this is the correct fix, the `GetXml` function is called for every channel so we could potentially be making an unneeded HEAD request 100s of times

timminator:
Hm, thats a valid concern. I've thought about it for a while, but right now I can't think of a good solution...

crobibero:
A proper solution might be to check the new file validity before starting import? Or always clearing the cache before starting import?

timminator:
> A proper solution might be to check the new file validity before starting import?

That was also my first thought. But from what I can see the refresh guide data task is triggered by `RefreshChannelsInternal()` inside the GuideManager. At this stage the ListingsProviderInfo with the path for the xmltv file is not available. But even when we had this info I cannot see an easy way to pass the information about which file contains several channels and does not need to be validated again to xmltvlistingsprovider. That seems kinda difficult to implement. Maybe I'm missing something?

> Or always clearing the cache before starting import?

Do you mean to just remove the _maxCacheAge parameter and check? If that's what you meant I thought about that aswell. That would actually be the easiest fix. If a refresh is triggered it _is_ then actually refreshed. The current _maxCacheAge implementation is far too error-prone in my opinion. I would prefer this solution if that's acceptable. Preventing unneeded requests is a good idea but as it is right now it just causes problems.

JPVenson:
closed due to inactivity

PR #13906, opened 2025-04-12
Add .gitignore style ignoring

Shadowghost:
Integrate https://github.com/arition/jellyfin-plugin-ignore as core functionality

**Changes**
* Keep the current logic as is and if the file has text in it, we parse it as .gitignore exclusions

### DIFF
--- a/Emby.Server.Implementations/Library/DotIgnoreIgnoreRule.cs
+++ b/Emby.Server.Implementations/Library/DotIgnoreIgnoreRule.cs
@@ -50,7 +50,7 @@ public class DotIgnoreIgnoreRule : IResolverIgnoreRule
 
         var folder = new DirectoryInfo(parentDirPath);
         var ignoreFile = FindIgnoreFile(folder);
-        if (ignoreFile == null)
+        if (ignoreFile is null)
         {
             return false;
         }
@@ -68,7 +68,7 @@ public class DotIgnoreIgnoreRule : IResolverIgnoreRule
         }
 
         // If file has content, base ignoring off the content .gitignore-style rules
-        var ignoreRules = ignoreFileString.Split("\n", StringSplitOptions.RemoveEmptyEntries);
+        var ignoreRules = ignoreFileString.Split('\n', StringSplitOptions.RemoveEmptyEntries);
         var ignore = new Ignore.Ignore();
         ignore.Add(ignoreRules);
 
--- a/Emby.Server.Implementations/Library/DotIgnoreIgnoreRule.cs
+++ b/Emby.Server.Implementations/Library/DotIgnoreIgnoreRule.cs
@@ -1,0 +1,77 @@
+using System;
+using System.IO;
+using MediaBrowser.Controller.Entities;
+using MediaBrowser.Controller.Resolvers;
+using MediaBrowser.Model.IO;
+
+namespace Emby.Server.Implementations.Library;
+
+/// <summary>
+/// Resolver rule class for ignoring files via .ignore.
+/// </summary>
+public class DotIgnoreIgnoreRule : IResolverIgnoreRule
+{
+    private static FileInfo? FindIgnoreFile(DirectoryInfo directory)
+    {
+        var ignoreFile = new FileInfo(Path.Join(directory.FullName, ".ignore"));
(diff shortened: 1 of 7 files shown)


jhparker88:
> Integrate https://github.com/arition/jellyfin-plugin-ignore as core functionality

Given that the repository includes some tests maybe it is a good idea to add them as well.

JPVenson (approved):

Emby.Server.Implementations/Library/DotIgnoreIgnoreRule.cs
```diff
@@ -0,0 +1,77 @@
...
+    {
+        var parentDirPath = Path.GetDirectoryName(fileInfo.FullName);
+        if (string.IsNullOrEmpty(parentDirPath))
+        {
+            return false;
+        }
+
+        var folder = new DirectoryInfo(parentDirPath);
+        var ignoreFile = FindIgnoreFile(folder);
+        if (ignoreFile == null)
```
crobibero, line 53:
```suggestion
        if (ignoreFile is null)
```

Emby.Server.Implementations/Library/DotIgnoreIgnoreRule.cs
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        if (string.IsNullOrEmpty(ignoreFileString))
+        {
+            // Ignore directory if we just have the file
+            return true;
+        }
+
+        // If file has content, base ignoring off the content .gitignore-style rules
+        var ignoreRules = ignoreFileString.Split("\n", StringSplitOptions.RemoveEmptyEntries);
```
crobibero, line 71:
```suggestion
        var ignoreRules = ignoreFileString.Split('\n', StringSplitOptions.RemoveEmptyEntries);
```

Emby.Server.Implementations/Library/DotIgnoreIgnoreRule.cs
```diff
@@ -0,0 +1,77 @@
...
+        return IsIgnored(fileInfo, parent);
+    }
+
+    /// <summary>
+    /// Checks whether or not the file is ignored.
+    /// </summary>
+    /// <param name="fileInfo">The file information.</param>
+    /// <param name="parent">The parent BaseItem.</param>
+    /// <returns>True if the file should be ignored.</returns>
+    public static bool IsIgnored(FileSystemMetadata fileInfo, BaseItem? parent)
```
crobibero, line 43: I think this should be cached, right now it will attempt to locate and read the `.ignore` file for every single file

crobibero (changes requested):

Emby.Server.Implementations/Library/DotIgnoreIgnoreRule.cs
Shadowghost: Yeah I was thinking about this too. Though there wasn't a big impact on my scan time  with the PR merged.

crobibero: Probably fine to not cache for now but we may need to revisit based on user reports
The bigger thing would be ensuring that the cache is invalidated when library scan starts so we don't keep ignoring after the .ignore file is changed

crobibero (approved):

Emby.Server.Implementations/Library/LibraryManager.cs
```diff
@@ -649,10 +649,16 @@ private Guid GetNewItemIdInternal(string key, Type type, bool forceCaseInsensiti
...
-            // Check to see if we should resolve based on our contents
-            if (args.IsDirectory && !ShouldResolvePathContents(args))
+            // Filter content based on ignore rules
+            if (args.IsDirectory)
             {
-                return null;
+                var filtered = args.GetActualFileSystemChildren().ToArray();
+                if (filtered.Length == 0)
+                {
+                    return null;
```
JPVenson, line 658: this causes the tests to fail. When there are no children present, an empty Folder type BaseItem will get returned as Null instead, breaking other parts of the code

JPVenson (changes requested):

JPVenson (approved):

Issue #13893, opened 2025-04-10
Scanning theme file causes error


theguymadmax:
### Description of the bug

On the master branch, scan a TV series with a theme.mp3 file in the series folder.  The error does not occur on the 10.10.z branch. 

### Reproduction steps

1. Scan in a Series library that contains a theme.mp3 file on the master branch/unstable build

### What is the current _bug_ behavior?

Error in logs

### What is the expected _correct_ behavior?

No errors

### Jellyfin Server version

Master

### Specify commit id

b31c09409296b6e11f1c9e8347e55d5a25060dba

### Specify the build version

Dev Server

### Environment

```markdown
- OS: Windows 11
- FFmpeg Version: 7.0.2-9 
- Plugins: None
```

### Jellyfin logs

```shell
[2025-04-09 21:51:11.105 -04:00] [INF] [6] MediaBrowser.MediaEncoding.Encoder.MediaEncoder: Starting "ffprobe" with args "-analyzeduration 200M -probesize 1G -i file:\"I:\TV\The Orville (2017)\theme.mp3\" -threads 0 -v warning -print_format json -show_streams -show_format"
[2025-04-09 21:51:11.242 -04:00] [ERR] [52] MediaBrowser.Providers.Music.AudioMetadataService: Error in "Probe Provider"
System.NullReferenceException: Object reference not set to an instance of an object.
   at MediaBrowser.Providers.MediaInfo.AudioFileProber.FetchDataFromTags(Audio audio, MediaInfo mediaInfo, MetadataRefreshOptions options, Boolean tryExtractEmbeddedLyrics) in C:\git\jellyfin\MediaBrowser.Providers\MediaInfo\AudioFileProber.cs:line 182
   at MediaBrowser.Providers.MediaInfo.AudioFileProber.FetchAsync(Audio audio, MediaInfo mediaInfo, MetadataRefreshOptions options, CancellationToken cancellationToken) in C:\git\jellyfin\MediaBrowser.Providers\MediaInfo\AudioFileProber.cs:line 148
   at MediaBrowser.Providers.MediaInfo.AudioFileProber.Probe[T](T item, MetadataRefreshOptions options, CancellationToken cancellationToken) in C:\git\jellyfin\MediaBrowser.Providers\MediaInfo\AudioFileProber.cs:line 116
[...]

PR #13908, opened 2025-04-12
Safeguard against null value trimming in tag results

Shadowghost:
MediaInfo values can be null and then the `Trim()` throws

**Changes**
* Safeguard against trimming `null`


CloCkWeRX (approved):

JPVenson (approved):

crobibero (approved):

JPVenson:
fixes #13893

PR #13934, opened 2025-04-16
Add  Filter MatchAllTags

geanmark:
Currently, filtering by tags only returns results that match any of the selected tags (i.e., OR logic).

**Changes**
- This change adds an optional parameter to enable filtering by all selected tags (i.e., AND logic), allowing for more precise results.

A follow-up pull request will be submitted to update the UI accordingly to support this new filtering option.

### DIFF
--- a/Jellyfin.Api/Controllers/ItemsController.cs
+++ b/Jellyfin.Api/Controllers/ItemsController.cs
@@ -154,6 +154,7 @@ public class ItemsController : BaseJellyfinApiController
     /// <param name="nameLessThan">Optional filter by items whose name is equally or lesser than a given input string.</param>
     /// <param name="studioIds">Optional. If specified, results will be filtered based on studio id. This allows multiple, pipe delimited.</param>
     /// <param name="genreIds">Optional. If specified, results will be filtered based on genre id. This allows multiple, pipe delimited.</param>
+    /// <param name="matchAllTags">Optional. If specified, results will be filtered based on matchAllTags. This allows multiple, pipe delimited.</param>
     /// <param name="enableTotalRecordCount">Optional. Enable the total record count.</param>
     /// <param name="enableImages">Optional, include image information in output.</param>
     /// <returns>A <see cref="QueryResult{BaseItemDto}"/> with the items.</returns>
@@ -244,6 +245,7 @@ public class ItemsController : BaseJellyfinApiController
         [FromQuery] string? nameLessThan,
         [FromQuery, ModelBinder(typeof(CommaDelimitedCollectionModelBinder))] Guid[] studioIds,
         [FromQuery, ModelBinder(typeof(CommaDelimitedCollectionModelBinder))] Guid[] genreIds,
+        [FromQuery] bool matchAllTags = false,
         [FromQuery] bool enableTotalRecordCount = true,
         [FromQuery] bool? enableImages = true)
     {
@@ -352,6 +354,7 @@ public class ItemsController : BaseJellyfinApiController
                 IsHD = isHd,
                 Is4K = is4K,
                 Tags = tags,
+                MatchAllTags = matchAllTags,
                 OfficialRatings = officialRatings,
                 Genres = genres,
                 ArtistIds = artistIds,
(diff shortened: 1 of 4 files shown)


JPVenson (approved):

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -1778,8 +1778,16 @@ private IQueryable<BaseItemEntity> TranslateQuery(
         if (tags.Count > 0)
         {
             var cleanValues = tags.Select(e => GetCleanValue(e)).ToArray();
-            baseQuery = baseQuery
-                    .Where(e => e.ItemValues!.Any(f => f.ItemValue.Type == ItemValueType.Tags && cleanValues.Contains(f.ItemValue.CleanValue)));
+            if (filter.MatchAllTags)
+            {
+                baseQuery = baseQuery
+                .Where(e => cleanValues.All(tag => e.ItemValues!.Any(f => f.ItemValue.Type == ItemValueType.Tags && f.ItemValue.CleanValue == tag)));
```
Bond-009, line 1784:
```suggestion
                    .Where(e => cleanValues.All(tag => e.ItemValues!.Any(f => f.ItemValue.Type == ItemValueType.Tags && f.ItemValue.CleanValue == tag)));
```

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -1778,8 +1778,16 @@ private IQueryable<BaseItemEntity> TranslateQuery(
...
+            }
+            else
+            {
+                baseQuery = baseQuery
+                .Where(e => e.ItemValues!.Any(f => f.ItemValue.Type == ItemValueType.Tags && cleanValues.Contains(f.ItemValue.CleanValue)));
```
Bond-009, line 1789:
```suggestion
                    .Where(e => e.ItemValues!.Any(f => f.ItemValue.Type == ItemValueType.Tags && cleanValues.Contains(f.ItemValue.CleanValue)));
```

Bond-009 (approved):

nielsvanvelzen (changes requested):
IMO a better API design would be to consider the existing `tags` parameter as `anyTag` and add a new `allTags` property. That way a client can also filter for items with any tag in list 1 and all tags in list 2 and there wouldn't be a need to set 2 parameters for a single filter.

geanmark:
@nielsvanvelzen  Thank you for the review! I’m not quite sure I understand the reasoning behind this suggestion—could you please explain your intention a bit more? If possible, I’d really appreciate some example sites that use this approach. Once I fully understand the purpose, I’ll be happy to implement it. Thanks again!

nielsvanvelzen:
Using multiple query parameters for a single behavior is bad API design. That's why I suggest to add am entirely new parameter that works on its own.

Jellyfin.Api/Controllers/ItemsController.cs
```diff
@@ -678,6 +682,7 @@ public ActionResult<QueryResult<BaseItemDto>> GetItemsByUserIdLegacy(
         [FromQuery, ModelBinder(typeof(PipeDelimitedCollectionModelBinder))] string[] genres,
         [FromQuery, ModelBinder(typeof(PipeDelimitedCollectionModelBinder))] string[] officialRatings,
         [FromQuery, ModelBinder(typeof(PipeDelimitedCollectionModelBinder))] string[] tags,
+        [FromQuery, ModelBinder(typeof(PipeDelimitedCollectionModelBinder))] string[] allTags,
```
nielsvanvelzen, line 685: The legacy API operations should not receive new parameters. They are called legacy for a reason

geanmark:
I'm having a bit of trouble finding the non-legacy API—any chance you could point me in the right direction?

nielsvanvelzen:
All operations ending with "Legacy" in the name (and have `ApiExplorerSettings(IgnoreApi = true)` set) are the old ones.

Bond-009:
Closing as stale, no response for 1+ year

PR #13946, opened 2025-04-18
Fix SyncPlay WebSocket OpenAPI schemas

nielsvanvelzen:
For better or worse, this _should_ allow SyncPlay to be implemented in Kotlin-SDK based clients. The way this is implemented will keep existing implementations working as there is no actual change in messages. This is purely for the OpenAPI specification.

**Changes**
- Redo the way SyncPlay GroupUpdate messages are defined so a correct polymorphic schema is generated in the OpenAPI document. The way it works is similar to how websocket messages work.
- Remove two unused GroupUpdateType members: CreateGroupDenied & JoinGroupDenied

**Issues**


MediaBrowser.Controller/Session/ISessionManager.cs
```diff
@@ -161,9 +161,8 @@ public interface ISessionManager
         /// <param name="sessionId">The identifier of the session.</param>
         /// <param name="command">The group update.</param>
         /// <param name="cancellationToken">The cancellation token.</param>
-        /// <typeparam name="T">Type of group.</typeparam>
         /// <returns>Task.</returns>
-        Task SendSyncPlayGroupUpdate<T>(string sessionId, GroupUpdate<T> command, CancellationToken cancellationToken);
+        Task SendSyncPlayGroupUpdate(string sessionId, GroupUpdate command, CancellationToken cancellationToken);
```
jhparker88, line 165: Given that a bit a rework has been done in this PR, I was wondering whether to rename `SendSyncPlayGroupUpdate` to `SendSyncPlayGroupUpdateAsync` in order to match the general name schema for those await/task methods?

nielsvanvelzen: The goal is to fix the OpenAPI schema with as little changes as possible. Anything else can be done by others in separate pull requests.

Shadowghost (approved):

nielsvanvelzen:
Looks like there is a small issue where half of the JSON is missing now when serializing, need to look into that before merging.

Bond-009 (approved):

crobibero (approved):

PR #13959, opened 2025-04-21
Add implementation for moving items

### LINKED ISSUE #13958: Support for moving items to allow for automatic organisation of libraries
I've been developing a plugin that automatically organises files on the file system according the recommendations [in the documentation](https://jellyfin.org/docs/general/server/media/shows). I would ultimately like to release this soon, but I keep running into the same issue with path mapping on items.

I've tried multiple approaches, but every time the moment a file is moved on the file system, Jellyfin considers the item to no longer exist and adds a new item for the file at the new location while deleting the item that referenced the old location. This is a challenge as it means that any attempts to automatically organise a library with Jellyfin or otherwise will cause a user to lose all configuration on the items within it (custom images, ratings, favourites, other metadata changes, etc.).

I have developed some functionality which allows a user to move an item via the API. Ultimately, I would like to have this functionality implemented within the UI too so that an admin can move an item in the same way that they may delete the item too. I will link the PR to this issue here.

I can use this functionality in conjunction with my auto-organiser plugin to organise a library and update the reference to the path in Jellyfin's database. This functionality effectively works, but I am still facing the same issue where the item is replaced with a brand new item for the moved file each time.

From looking around at some previous issues and some comments there, I understand there is an issue in which the item ID is coupled to the path and so any updates to the path generates a new ID for an item hence Jellyfin assumes the file is deleted and adds a new item with the new ID. I also understand that there migrations to the database in development currently that may address this. Will these changes help support items with updated paths? If help is needed to support, I will happily contribute, though I may need some more context if any one could support.

Thanks and looking forward to contributing!

geo-martino:
**Changes**

Adds API endpoint and implementation for moving items to new locations on the file system while updating the reference to this new location on the item in the repository. The implementation closely follows the same patterns used for `DeleteItem` including setting per-user permissions for moving which ultimately can be configured in the UI with later changes.

Also expands the functionality of the `MoveDirectory` method in `ManagedFileSystem` to better integrate with the new `MoveFile` method and related methods.

I have conducted regression tests on a development build for the API endpoint to confirm the changes work as expected. The major caveat being that the item is replaced in the database when the item is moved to a new location. Happy to either expand this PR or open a new PR to address the remapping of IDs on items if required. (I've added a more expanded explanation of this issue in the linked issue).

**Issues**

Partially addresses #13958


Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -566,6 +566,42 @@ public void UpdateOrInsertItems(IReadOnlyList<BaseItemDto> items, CancellationTo
...
+
+        var entity = Map(item);
+        // TODO: refactor this "inconsistency"
+        entity.TopParentId = item.GetTopParent()?.Id;
+
+        using var context = _dbProvider.CreateDbContext();
+        using var transaction = context.Database.BeginTransaction();
+
+        context.BaseItems.Update(entity);
+        context.BaseItems
```
gnattu, line 594: I’m afraid simply updating the path for the `BaseItem` entry is not enough. Playlists and Collections might still reference the absolute path of the item in their respective columns (and that is a json in a column due to emby's design), making it challenging to perform a reverse lookup to determine which playlist or collection requires an update. In the worst case we need to loop through all the playlists and collections to just check for updates.

Emby.Server.Implementations/Library/LibraryManager.cs
```diff
@@ -311,6 +311,134 @@ public void RegisterItem(BaseItem item)
...
+
+            var sourceParent = Path.GetDirectoryName(item.Path) ?? string.Empty;
+            var destinationParent = Path.GetDirectoryName(destination) ?? string.Empty;
+
+            _itemRepository.MoveItem(item, destination);
+            _cache.TryRemove(item.Id, out _);
+
+            foreach (var child in children!)
+            {
+                // children are moved recursively when item is moved on file system, just update paths in repository
```
gnattu, line 422: Sometimes, items needs to be moved and not everything related to it on the filesystems are its children in the filesystem schema. For example, external subtitles, external audio tracks, trickplay files, nfo files and cover images.

MediaBrowser.Controller/Entities/BaseItem.cs
```diff
@@ -801,6 +801,43 @@ public virtual string CreatePresentationUniqueKey()
...
         }
 
+        public virtual bool CanMove()
+        {
+            if (SourceType == SourceType.Channel)
+            {
+                return ChannelManager.CanMove(this);
+            }
+
+            return IsFileProtocol;
```
gnattu, line 811: Correct me if I'm wrong. This would means that the user is allowed to move a single item in a library out of the library root path, and move a children like an episode out of the season path. If that is the case, this would not work well with our current library scanning code and those items may be deleted on next scan. Our current library scanning logic relies on the file hierarchy heavily and cannot support the flexible item sources like such movement could create.

gnattu (commented):
I appreciate the effort you’ve put into implementing this feature. However, our current database schema and library management logic are inadequate to support a flexible feature like this. My suggestion is to restrict the scope of this feature to only allow the relocation of the library root path. To ensure that all playlists and collections are updated, we have to check all of the entries in db which is an expensive operation, but if it is only executed with another already expensive operation like relocating a library, its cost is not that bad.

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
geo-martino: Understood. Then it seems like there are 2 options to getting this to work as expected:

- Implement logic to do the reverse lookup on playlists and collections (performance concerns)
- Refactor the references for playlists and collections to use the IDs of items instead of their absolute paths (potential major refactoring)

Given that the former option would be more of a workaround, I would be happy to look into the latter option in a separate PR to allow for the safe moving of child items. Is this something that is worth looking into or is this already in scope with the EF Core changes? I had seen some mention of improvements to path/ID handling mentioned elsewhere so just confirming. If so, any points in the right direction would be greatly appreciated to get me started. Would love to see this feature supported and very happy to contribute to make it work!

Emby.Server.Implementations/Library/LibraryManager.cs
geo-martino: Noted. I can take another look into this with that assumption in mind.

geo-martino:
> I appreciate the effort you’ve put into implementing this feature. However, our current database schema and library management logic are inadequate to support a flexible feature like this. My suggestion is to restrict the scope of this feature to only allow the relocation of the library root path. To ensure that all playlists and collections are updated, we have to check all of the entries in db which is an expensive operation, but if it is only executed with another already expensive operation like relocating a library, its cost is not that bad.

Understood. Then to reduce the scope of this PR, I can throw an error on any item types coming in to the `MoveItem` method on the `LibraryManager` that aren't `Library`. I can then come back to remove this throw if/when other item types can be supported.

JPVenson:
closed due to inactivity

PR #13964, opened 2025-04-22
Remove internal TLS support

JPVenson:
Removes jellyfins internal handling of TLS certificates and https protocol support.
Adds loading of application settings files to substitute configuration via kestrels internal configuration.

https://learn.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel/endpoints?view=aspnetcore-9.0#configure-https-in-appsettingsjson


JPVenson:
@enter-a-random-username can you please provide a logfile?

KGT1:
If the larger goal is to move at least startup-required config into appsettings.json, wouldnt some kind of migration / tool be necessary anyway for 10.12. to keep most installs working? Would it then really make sense to not have the tls stuff in that?

mitzsch:
So, when this is merged, a user could still enable tls support with the appsettings.json? Where does the JSON need to be placed? /etc/jellyfin?

JPVenson:
yes. In the config dir alongside the rest of the configs

mitzsch:
Thanks!

I just tried it on a test install of Jellyfin 10.11.8, and it works perfectly fine (and was very easy to set up)

For anyone interested, here is my config of the appsettings.json created under /etc/jellyfin:

Minimal config:
```
{
  "Kestrel": {
    "Endpoints": {
      "JellyfinSecure": {
        "Url": "https://*:8920",
        "Certificate": {
          "Path": "/opt/ssl_cert/JellyfinTest_Cert.pfx",
          "Password": "<cert-pass>"
        }
      }
    }
  }
}
```

Config for altering the HTTP port as well:

```
{
  "Kestrel": {
    "Endpoints": {
      "JellyfinUnSecure": {
        "Url": "http://*:5000"
      },
      "JellyfinSecure": {
        "Url": "https://*:8920",
        "Certificate": {
          "Path": "/opt/ssl_cert/JellyfinTest_Cert.pfx",
          "Password": "<cert-pass>"
        }
      }
    }
  }
}
```


Config for using .cert and .pem instead of pfx (and without password):
```
{
  "Kestrel": {
    "Endpoints": {
      "JellyfinUnSecure": {
        "Url": "http://*:5000"
      },
      "JellyfinSecure": {
        "Url": "https://*:8922",
        "Certificate": {
          "Path": "/opt/ssl_cert/JellyfinTest_Cert.crt",
          "KeyPath": "/opt/ssl_cert/JellyfinTest_Cert.pem",
        }
      }
    }
  }
}
```

JPVenson:
Yes thats the way it will work.

However keep in mind that currently
```json
      "JellyfinUnSecure": {
        "Url": "http://*:5000"
      },
```

will add a _new_ binding and not override the existing 8096 one. That is still configured on code and _might_ be changed at a later stage.

mitzsch:
Ah yeah, noticed that as well, although nice to see that with the appsettings.json it's possible to "_change_" (= add new binding) the http port as well.

Shadowghost (approved):

Issue #14008, opened 2025-04-26
Inadequate logging of blocked connections due to remote IP filter and other similar configuration


lf-:
### Description of the bug

There's been many reports of this type of issue before, which are all user error in the end, but the common thread among them is that very little is logged, the server returns 503 to `/system/info/public` or some other endpoint, and the administrator is confused.

It's my second or third time debugging this issue several months apart (I dunno why I keep messing it up) and Jellyfin doesn't make it easy to debug. I *know* my configuration is broken, but I don't know what part of it is causing the failure, and the logging doesn't help.

Examples: https://forum.jellyfin.org/t-solved-can-no-longer-reach-jellyfin-after-reboot, https://github.com/jellyfin/jellyfin/issues/13954, https://github.com/jellyfin/jellyfin/issues/12595, and anything that looks similar.

### Reproduction steps

Mess up the known proxies, allowed remote IPs or other similar settings so that the client is not allowed to connect. `<RemoteIPFilter><string>127.0.0.1</string></RemoteIPFilter>` should do.

Try to connect.

Look at server logs.

### What is the current _bug_ behavior?

There is no helpful logging of the problem even when setting MinimumLevel to Debug: https://jellyfin.org/docs/general/administration/troubleshooting/#debug-logging.

This is the shape of logs that I do get, which are unhelpful since they don't identify that a request came in, nor that the request hit any authorization machinery and failed:

```
Apr 26 23:18:22 cube jellyfin[5885]: [23:18:22] [DBG] [11] Jellyfin.Api.Auth.CustomAuthenticationHandler: AuthenticationS
cheme: CustomAuthentication was not authenticated.
Apr 26 23:18:22 cube jellyfin[5885]: [23:18:22] [DBG] [11] Jellyfin.Networking.Manager.NetworkManager: Trying to get bind
 address for source 127.0.0.1 - External: False
Apr 26 23:18:22 cube jellyfin[5885]: [23:18:22] [DBG] [11] Jellyfin.Networking.Manager.NetworkManager: 127.0.0.1: Matchin
g bind address override found: https://REDACTED
Apr 26 23:18:48 cube jellyfin[5885]: [23:18:48] [DBG] [11] Jellyfin.Api.Auth.CustomAuthenticationHandler: AuthenticationS
cheme: CustomAuthentication was not authenticated.
[...]

PR #14011, opened 2025-04-27
Fix missing logging of connections by disallowed IPs

lf-:
**Changes**
I rewrote the remote access policy code (intended as no functional change) to give reasons for rejecting/allowing connections as well as restructuring it to be more in line with best practices in authorization code of having the fall-through case be to reject. This then allows for fixing #14008 in a satisfying way: actually reporting *why* the connection was rejected.

Secondly I deleted the LAN checking middleware, since it is a subset of the functionality of the IP checking middleware anyway and has worse diagnostics.

I have tested this locally (which actually is why I found the redundant LAN middleware).

**Issues**

Fixes: https://github.com/jellyfin/jellyfin/issues/14008


crobibero:
> Hrm. I spot an API breaking change. I will push a commit that fixes the API break with a wrapper function marked deprecated, and then y'all can decide if you would actually like to change the API.

You had an ABI change which in this case is fine, you can remove the commit to add that property

MediaBrowser.Common/Net/INetworkManager.cs
```diff
@@ -18,6 +18,38 @@ public interface INetworkManager
         /// </summary>
         event EventHandler NetworkChanged;
 
+        /// <summary>
+        /// Result of <see cref="INetworkManager.ShouldAllowServerAccess" />.
+        /// </summary>
+        public enum RemoteAccessPolicyResult
```
crobibero, line 24: Move this enum to own file

src/Jellyfin.Networking/Manager/NetworkManager.cs
```diff
@@ -690,33 +691,42 @@ public bool TryParseInterface(string intf, [NotNullWhen(true)] out IReadOnlyList
...
-                var matches = _remoteAddressFilter.Count(remoteNetwork => NetworkUtils.SubnetContainsAddress(remoteNetwork, remoteIP));
-                if ((!config.IsRemoteIPFilterBlacklist && matches > 0)
-                    || (config.IsRemoteIPFilterBlacklist && matches == 0))
-                {
-                    return true;
-                }
 
-                return false;
+            // remoteAddressFilter is a whitelist or blacklist.
+            var matches = _remoteAddressFilter.Count(remoteNetwork => NetworkUtils.SubnetContainsAddress(remoteNetwork, remoteIP));
```
crobibero, line 714: Since we only care about 0/1 we can do `.Any()` here instead for a small fast-path

crobibero (approved):

Shadowghost (approved):

tests/Jellyfin.Networking.Tests/NetworkParseTests.cs
```diff
@@ -299,15 +299,38 @@ public void HasRemoteAccess_GivenWhitelist_AllowsOnlyIPsInWhitelist(string addre
             var startupConf = new Mock<IConfiguration>();
             using var nm = new NetworkManager(NetworkParseTests.GetMockConfig(conf), startupConf.Object, new NullLogger<NetworkManager>());
 
-            Assert.NotEqual(nm.HasRemoteAccess(IPAddress.Parse(remoteIP)), denied);
+            Assert.Equal(nm.ShouldAllowServerAccess(IPAddress.Parse(remoteIP)), expectedResult);
```
Bond-009, line 302: First argument is the expected result
```suggestion
            Assert.Equal(expectedResult, nm.ShouldAllowServerAccess(IPAddress.Parse(remoteIP)));
```

tests/Jellyfin.Networking.Tests/NetworkParseTests.cs
```diff
@@ -299,15 +299,38 @@ public void HasRemoteAccess_GivenWhitelist_AllowsOnlyIPsInWhitelist(string addre
...
+                EnableIPv4 = true,
+                EnableRemoteAccess = false,
+                RemoteIPFilter = addresses.Split(','),
+                IsRemoteIPFilterBlacklist = false
+            };
+
+            var startupConf = new Mock<IConfiguration>();
+            using var nm = new NetworkManager(NetworkParseTests.GetMockConfig(conf), startupConf.Object, new NullLogger<NetworkManager>());
+
+            Assert.Equal(nm.ShouldAllowServerAccess(IPAddress.Parse(remoteIP)), expectedResult);
```
Bond-009, line 325:
```suggestion
            Assert.Equal(expectedResult, nm.ShouldAllowServerAccess(IPAddress.Parse(remoteIP)));
```

tests/Jellyfin.Networking.Tests/NetworkParseTests.cs
```diff
@@ -321,7 +344,7 @@ public void HasRemoteAccess_GivenBlacklist_BlacklistTheIPs(string addresses, str
             var startupConf = new Mock<IConfiguration>();
             using var nm = new NetworkManager(NetworkParseTests.GetMockConfig(conf), startupConf.Object, new NullLogger<NetworkManager>());
 
-            Assert.NotEqual(nm.HasRemoteAccess(IPAddress.Parse(remoteIP)), denied);
+            Assert.Equal(nm.ShouldAllowServerAccess(IPAddress.Parse(remoteIP)), expectedResult);
```
Bond-009, line 347:
```suggestion
            Assert.Equal(expectedResult, nm.ShouldAllowServerAccess(IPAddress.Parse(remoteIP)));
```

src/Jellyfin.Networking/Manager/NetworkManager.cs
```diff
@@ -690,33 +690,42 @@ public bool TryParseInterface(string intf, [NotNullWhen(true)] out IReadOnlyList
...
-    public bool HasRemoteAccess(IPAddress remoteIP)
+    public RemoteAccessPolicyResult ShouldAllowServerAccess(IPAddress remoteIP)
     {
         var config = _configurationManager.GetNetworkConfiguration();
+        if (IsInLocalNetwork(remoteIP))
+        {
+            return RemoteAccessPolicyResult.Allow;
+        }
+
         if (config.EnableRemoteAccess)
```
Bond-009, line 701: Do an early return here just like you did for the LAN check, makes the code easier to follow.

Bond-009 (changes requested):

Bond-009 (approved):
Not 100% sure about the `ShouldAllowServerAccess` method, with the current design the `Allow` value of the enum is only for internal use AFAIK. Maybe a better design would be `bool ShouldAllowServerAccess((IPAddress remoteIP, out RemoteAccessPolicyResult failureReason)`

PR #14023, opened 2025-04-29
Cleanup file related code

Bond-009:
* Add a helper function to efficiently create empty files
* Use enumeration methods to gets files where possible
* Don't check for directory existence before calling Directory.Create
* Add AsyncFile.Create helper function

Benchmark results for the different file creation options:
```
| Method                | Mean     | Error     | StdDev    | Gen0   | Allocated |
|---------------------- |---------:|----------:|----------:|-------:|----------:|
| FileCreate            | 2.544 us | 0.0026 us | 0.0021 us | 0.0038 |     232 B |
| FileCreateEmptyBuffer | 2.511 us | 0.0050 us | 0.0046 us |      - |     160 B |
| FileWriteAllBytes     | 2.744 us | 0.0030 us | 0.0028 us |      - |      64 B |
| FileOpenHandle        | 2.448 us | 0.0071 us | 0.0066 us |      - |      64 B |
```

### DIFF
--- a/MediaBrowser.Model/IO/AsyncFile.cs
+++ b/MediaBrowser.Model/IO/AsyncFile.cs
@@ -27,6 +27,14 @@ namespace MediaBrowser.Model.IO
         };
 
         /// <summary>
+        /// Creates, or truncates and overwrites, a file in the specified path.
+        /// </summary>
+        /// <param name="path">The path and name of the file to create.</param>
+        /// <returns>A <see cref="FileStream" /> that provides read/write access to the file specified in path.</returns>
+        public static FileStream Create(string path)
+            => new FileStream(path, FileMode.Create, FileAccess.ReadWrite, FileShare.None, IODefaults.FileStreamBufferSize, FileOptions.Asynchronous);
+
+        /// <summary>
         /// Opens an existing file for reading.
         /// </summary>
         /// <param name="path">The file to be opened for reading.</param>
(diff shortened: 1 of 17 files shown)


Shadowghost (approved):

MediaBrowser.Model/IO/AsyncFile.cs
```diff
@@ -26,6 +26,14 @@ public static class AsyncFile
             Options = FileOptions.Asynchronous
         };
 
+        /// <summary>
+        /// Creates, or truncates and overwrites, a file in the specified path.
+        /// </summary>
+        /// <param name="path">The path and name of the file to create.</param>
+        /// <returns>A <see cref="FileStream" /> that provides read/write access to the file specified in path.</returns>
+        public static FileStream Create(string path)
```
JPVenson, line 34:
```suggestion
        public static FileStream CreateOpen(string path)
```
just an idea to signal that it returns a filestream?

JPVenson (approved):

MediaBrowser.Model/IO/AsyncFile.cs
Bond-009: This class is intended to mirror the System.IO.File class but with async filestreams by default.

https://learn.microsoft.com/en-us/dotnet/api/system.io.file.create?view=netframework-4.7#system-io-file-create(system-string)

PR #14030, opened 2025-04-30
Localization/iso6392.txt: Correct the Portuguese entries

baka0815:
``pob`` and ``pop`` are not valid iso 639 codes for Portuguese. Wikipedia says ``pop`` is Pwapwâ (https://en.wikipedia.org/w/index.php?title=ISO_639:pop) and ``pob`` is Poqomchiʼ (https://en.wikipedia.org/w/index.php?title=ISO_639:pob).

For testing, it was needed to rescan the specific files. Just renaming/adding subtitle files was not enough, but the movie had to be new.

**Changes**
Changed the values to ``por`` which is the valid ISO 639-2 code for Portuguese.
Before this PR subtitle files like ``movie.pt-br.srt`` would be scanned as ``pob``:
image
After this PR they are read as ``por``:
image

**Issues**
Should fix the problem referenced by @webysther in #6302.

### DIFF
--- a/Emby.Server.Implementations/Localization/iso6392.txt
+++ b/Emby.Server.Implementations/Localization/iso6392.txt
@@ -347,8 +347,8 @@ pli||pi|Pali|pali
 pol||pl|Polish|polonais
 pon|||Pohnpeian|pohnpei
 por||pt|Portuguese|portugais
-pop||pt-pt|Portuguese (Portugal)|portugais (pt-pt)
-pob||pt-br|Portuguese (Brazil)|portugais (pt-br)
+por||pt-pt|Portuguese (Portugal)|portugais (pt-pt)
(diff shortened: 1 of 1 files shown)


webysther:
This look correct but I want to expand this topic, I will open a discussion about this.

webysther:
https://github.com/jellyfin/jellyfin/discussions/14032

baka0815:
@Shadowghost @crobibero @JPVenson @nielsvanvelzen 
Is there anything preventing this to be merged? If so, then please let me know.

JPVenson:
please dont ping team members out of the blue without very good reason.
Stuff get merged when it gets merged.

Bond-009 (approved):

Shadowghost (approved):

Issue #14035, opened 2025-05-01
Incorrect path used for Trickplay images saved next to media


theguymadmax:
### Description of the bug

Trickplay images saved next to media are no longer displayed on the Master/unstable build due to an incorrect path. 

 Regression from #13847,

### Reproduction steps

1. Run a 10.11 server instance
2. Generate trickplay images saved next to media.

### What is the current _bug_ behavior?

The path is duplicated, causing tickplay actions not to work when they are saved next to media.

### What is the expected _correct_ behavior?

The correct path is used, causing tickplay related items to work when they are saved next to media. 

### Jellyfin Server version

Master

### Specify commit id

055d466d16fd61b89c4b7c88a0e57da615e86ec1

### Specify the build version

10.11 Master (2025042805)

### Environment

```markdown
- OS: Windows 11
- FFmpeg Version: 7.1.1-2 
- Plugins: None
```

### Jellyfin logs

```shell
[22:11:56] [INF] [10] Jellyfin.Server.Implementations.Trickplay.TrickplayManager: Creating trickplay files at 320 width, for I:\Test Libraries\Movies Trickplays\Big Buck Bunny\Big Buck Bunny.mp4 [ID: de6c4a31-9e9a-8403-91e5-18b0059f3d0b]
[22:11:56] [INF] [10] MediaBrowser.MediaEncoding.Encoder.MediaEncoder: Trickplay generation: c:/usr/lib/jellyfin-ffmpeg/ffmpeg -loglevel error -threads 1 -i file:"I:\Test Libraries\Movies Trickplay\Big Buck Bunny\Big Buck Bunny.mp4" -an -sn -vf "fps=0.10000000149011612,setparams=color_primaries=bt709:color_trc=bt709:colorspace=bt709,scale=trunc(min(max(iw\,ih*(a*sar))\,320)/2)*2:trunc(ow/(a*sar)/2)*2,format=yuv420p" -threads 1 -c:v mjpeg -qscale:v 4 -fps_mode passthrough -f image2 "C:\Users\Max\AppData\Local\Temp\jellyfin\393f04d18fe44a40bd44c2e67fcbf5e8\%08d.jpg"
6
Exception thrown: 'System.Threading.Tasks.TaskCanceledException' in System.Private.CoreLib.dll
Exception thrown: 'System.IO.IOException' in System.Private.CoreLib.dll
[...]

PR #14036, opened 2025-05-01
Fix trickplay directory path construction

theguymadmax:
**Changes**
Get filename using `Path.GetFileName()` before changing extension. 
Reverted to using `Path.Combine()` instead of` Path.Join()`

**Issues**
 Fixes: #14035


gnattu:
This is probably not the correct way to fix this. All `Path.Combine` which can resolve two absolute paths are changed to `Path.Join` without clear reasons. I think we need to revert most of the usages back to `Path.Combine`.

cc @Shadowghost 
What was the intention to use `Path.Join` instead? Was `Path.Combine` causing any problems?

Shadowghost (approved):

Shadowghost:
Yeah I think I did a logic error there, migth want to revert all of them back.

Bond-009 (approved):

PR #14039, opened 2025-05-01
Add partition helper

JPVenson:
Adds helper for enumerating paged EFCore queries


EraYaN:
Instead of Skip + Take maybe consider KeySet (like [this](https://github.com/mrahhal/MR.EntityFrameworkCore.KeysetPagination)) Pagination, so the last couple of queries don't have to process the full result set. Can really really speed up your Partitioning code since ordering on an index can be much quicker than retrieving a full result set.

JPVenson:
in general looks quite nice, but this is not compatiable with our current infastructure without heavy modification or workaround (sometimes we really only have an iqueryable and the ordering info is detaced added) i guess its really only worth if build with from the ground up.

crobibero:
@JPVenson build is failing

JPVenson:
@crobibero no its not :-P

Jellyfin.Server/Migrations/Routines/MoveExtractedFiles.cs
```diff
@@ -76,27 +75,21 @@ public void Perform()
         // Make sure directories exist
         Directory.CreateDirectory(SubtitleCachePath);
         Directory.CreateDirectory(AttachmentCachePath);
-        do
+
+        await foreach (var result in context.BaseItems
```
crobibero, line 79: This query removes the limited select which was a significant performance gain for the migration

crobibero (changes requested):

Jellyfin.Server/Migrations/Routines/MoveExtractedFiles.cs
JPVenson: readded

src/Jellyfin.Database/Jellyfin.Database.Implementations/ProgressablePartitionReporting.cs
```diff
@@ -0,0 +1,55 @@
...
+
+    internal Action<int>? OnBeginPartition { get; set; }
+
+    internal Action<TEntity, int, int, TimeSpan>? OnEndItem { get; set; }
+
+    internal Action<int, TimeSpan>? OnEndPartition { get; set; }
+
+    internal IOrderedQueryable<TEntity> Source => _source;
+
+    internal void BeginItem(TEntity entity, int itteration, int itemIndex)
```
crobibero, line 34:
```suggestion
    internal void BeginItem(TEntity entity, int iteration, int itemIndex)
```

Someone's IDE doesn't have spellcheck enabled :P (also in QueryPartitionHelpers)

Jellyfin.Server.Implementations/Jellyfin.Server.Implementations.csproj
```diff
@@ -39,4 +39,8 @@
     <ProjectReference Include="..\src\Jellyfin.Database\Jellyfin.Database.Providers.Sqlite\Jellyfin.Database.Providers.Sqlite.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Folder Include="Helpers/" />
+  </ItemGroup>
+
```
crobibero, line 45:
```suggestion
```

src/Jellyfin.Database/Jellyfin.Database.Implementations/ProgressablePartitionReporting.cs
JPVenson: vscode has a spellchecker? xD

crobibero (approved):

Shadowghost (approved):

PR #14051, opened 2025-05-03
EFCore/#13049: update database to 10.10.z before EFCoe migration is applied

### LINKED ISSUE #13049: [EfJellyfin.db] Migration from 10.8.13 fails

### Description of the bug

When trying to migrate from a database thats still on 10.8.13 up to 10.11 the migration fails

### Reproduction steps

Setup a 10.8.13 instance.
Skip 10.9 and install 10.11

Migration fails with exception

does not fail

### Jellyfin Server version

Unstable

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

latest

### Environment

```markdown
- OS:
- Linux Kernel:
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

```shell
[2024-11-15 09:08:47.609 -05:00] [ERR] [1] Jellyfin.Server.Migrations.MigrationRunner: Could not apply migration '"MigrateLibraryDbData"'
Microsoft.Data.Sqlite.SqliteException (0x80004005): SQLite Error 1: 'no such column: LUFS'.
   at Microsoft.Data.Sqlite.SqliteException.ThrowExceptionForRC(Int32 rc, sqlite3 db)
   at Microsoft.Data.Sqlite.SqliteCommand.PrepareAndEnumerateStatements()+MoveNext()
   at Microsoft.Data.Sqlite.SqliteCommand.GetStatements()+MoveNext()
   at Microsoft.Data.Sqlite.SqliteDataReader.NextResult()
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteReader(CommandBehavior behavior)
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteReader()
   at Emby.Server.Implementations.Data.SqliteExtensions.Query(SqliteConnection sqliteConnection, String commandText)+MoveNext() in \\wsl.localhost\Ubuntu-24.04\home\Max\jellyfin\Emby.Server.Implementations\Data\SqliteExtensions.cs:line 63
   at Jellyfin.Server.Migrations.Routines.MigrateLibraryDb.Perform() in \\wsl.localhost\Ubuntu-24.04\home\Max\jellyfin\Jellyfin.Server\Migrations\Routines\MigrateLibraryDb.cs:line 94
   at Jellyfin.Server.Migrations.MigrationRunner.PerformMigrations(IMigrationRoutine[] migrations, MigrationOptions migrationOptions, Action`1 saveConfiguration, ILogger logger) in \\wsl.localhost\Ubuntu-24.04\home\Max\jellyfin\Jellyfin.Server\Migrations\MigrationRunner.cs:line 132
   at Jellyfin.Server.Migrations.MigrationRunner.Run(CoreAppHost host, ILoggerFactory loggerFactory) in \\wsl.localhost\Ubuntu-24.04\home\Max\jellyfin\Jellyfin.Server\Migrations\MigrationRunner.cs:line 69
[2024-11-15 09:08:47.652 -05:00] [FTL] [1] Main: Error while starting server
Microsoft.Data.Sqlite.SqliteException (0x80004005): SQLite Error 1: 'no such column: LUFS'.
   at Microsoft.Data.Sqlite.SqliteException.ThrowExceptionForRC(Int32 rc, sqlite3 db)
   at Microsoft.Data.Sqlite.SqliteCommand.PrepareAndEnumerateStatements()+MoveNext()
   at Microsoft.Data.Sqlite.SqliteCommand.GetStatements()+MoveNext()
   at Microsoft.Data.Sqlite.SqliteDataReader.NextResult()
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteReader(CommandBehavior behavior)
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteReader()
   at Emby.Server.Implementations.Data.SqliteExtensions.Query(SqliteConnection sqliteConnection, String commandText)+MoveNext() in \\wsl.localhost\Ubuntu-24.04\home\Max\jellyfin\Emby.Server.Implementations\Data\SqliteExtensions.cs:line 63
   at Jellyfin.Server.Migrations.Routines.MigrateLibraryDb.Perform() in \\wsl.localhost\Ubuntu-24.04\home\Max\jellyfin\Jellyfin.Server\Migrations\Routines\MigrateLibraryDb.cs:line 94
   at Jellyfin.Server.Migrations.MigrationRunner.PerformMigrations(IMigrationRoutine[] migrations, MigrationOptions migrationOptions, Action`1 saveConfiguration, ILogger logger) in \\wsl.localhost\Ubuntu-24.04\home\Max\jellyfin\Jellyfin.Server\Migrations\MigrationRunner.cs:line 132
   at Jellyfin.Server.Migrations.MigrationRunner.Run(CoreAppHost host, ILoggerFactory loggerFactory) in \\wsl.localhost\Ubuntu-24.04\home\Max\jellyfin\Jellyfin.Server\Migrations\MigrationRunner.cs:line 69
   at Jellyfin.Server.Program.StartServer(IServerApplicationPaths appPaths, StartupOptions options, IConfiguration startupConfig) in \\wsl.localhost\Ubuntu-24.04\home\Max\jellyfin\Jellyfin.Server\Program.cs:line 161
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

TheMelmacian:
**Changes**
- added a migration routine that contains the migration code from the old `SqliteItemRepository.cs` and `SqliteUserDataRepository.cs`
- migration of existing `migration.xml` files is executed regardless of the `IsStartupWizardCompleted` flag to prevent execution of already applied migrations

**Issues**
Fixes #13049

### DIFF
--- a/Jellyfin.Server/Migrations/JellyfinMigrationService.cs
+++ b/Jellyfin.Server/Migrations/JellyfinMigrationService.cs
@@ -1,3 +1,5 @@
+#pragma warning disable RS0030 // Do not use banned APIs
+
 using System;
 using System.Collections.Generic;
 using System.IO;
@@ -93,35 +95,33 @@ internal class JellyfinMigrationService
 
             logger.LogInformation("Migration system initialisation completed.");
         }
-        else
-        {
-            // migrate any existing migration.xml files
-            var migrationConfigPath = Path.Join(appPaths.ConfigurationDirectoryPath, "migrations.xml");
-            var migrationOptions = File.Exists(migrationConfigPath)
-                 ? (MigrationOptions)xmlSerializer.DeserializeFromFile(typeof(MigrationOptions), migrationConfigPath)!
-                 : null;
-            if (migrationOptions != null && migrationOptions.Applied.Count > 0)
-            {
-                logger.LogInformation("Old migration style migration.xml detected. Migrate now.");
-                var dbContext = await _dbContextFactory.CreateDbContextAsync().ConfigureAwait(false);
-                await using (dbContext.ConfigureAwait(false))
-                {
-                    var historyRepository = dbContext.GetService<IHistoryRepository>();
-                    var appliedMigrations = await dbContext.Database.GetAppliedMigrationsAsync().ConfigureAwait(false);
-                    var oldMigrations = Migrations.SelectMany(e => e)
-                        .Where(e => migrationOptions.Applied.Any(f => f.Id.Equals(e.Metadata.Key!.Value))) // this is a legacy migration that will always have its own ID.
-                        .Where(e => !appliedMigrations.Contains(e.BuildCodeMigrationId()))
-                        .ToArray();
-                    var startupScripts = oldMigrations.Select(e => (Migration: e.Metadata, Script: historyRepository.GetInsertScript(new HistoryRow(e.BuildCodeMigrationId(), GetJellyfinVersion()))));
-                    foreach (var item in startupScripts)
-                    {
-                        logger.LogInformation("Migrate migration {Key}-{Name}.", item.Migration.Key, item.Migration.Name);
-                        await dbContext.Database.ExecuteSqlRawAsync(item.Script).ConfigureAwait(false);
-                    }
 
-                    logger.LogInformation("Rename old migration.xml to migration.xml.backup");
-                    File.Move(migrationConfigPath, Path.ChangeExtension(migrationConfigPath, ".xml.backup"), true);
+        // migrate any existing migration.xml files
+        var migrationConfigPath = Path.Join(appPaths.ConfigurationDirectoryPath, "migrations.xml");
+        var migrationOptions = File.Exists(migrationConfigPath)
+            ? (MigrationOptions)xmlSerializer.DeserializeFromFile(typeof(MigrationOptions), migrationConfigPath)!
+            : null;
+        if (migrationOptions != null && migrationOptions.Applied.Count > 0)
+        {
+            logger.LogInformation("Old migration style migration.xml detected. Migrate now.");
+            var dbContext = await _dbContextFactory.CreateDbContextAsync().ConfigureAwait(false);
+            await using (dbContext.ConfigureAwait(false))
+            {
+                var historyRepository = dbContext.GetService<IHistoryRepository>();
+                var appliedMigrations = await dbContext.Database.GetAppliedMigrationsAsync().ConfigureAwait(false);
+                var oldMigrations = Migrations.SelectMany(e => e)
+                    .Where(e => migrationOptions.Applied.Any(f => f.Id.Equals(e.Metadata.Key!.Value))) // this is a legacy migration that will always have its own ID.
+                    .Where(e => !appliedMigrations.Contains(e.BuildCodeMigrationId()))
+                    .ToArray();
+                var startupScripts = oldMigrations.Select(e => (Migration: e.Metadata, Script: historyRepository.GetInsertScript(new HistoryRow(e.BuildCodeMigrationId(), GetJellyfinVersion()))));
+                foreach (var item in startupScripts)
+                {
+                    logger.LogInformation("Migrate migration {Key}-{Name}.", item.Migration.Key, item.Migration.Name);
+                    await dbContext.Database.ExecuteSqlRawAsync(item.Script).ConfigureAwait(false);
                 }
+
+                logger.LogInformation("Rename old migration.xml to migration.xml.backup");
+                File.Move(migrationConfigPath, Path.ChangeExtension(migrationConfigPath, ".xml.backup"), true);
             }
         }
     }
--- a/Jellyfin.Server/Migrations/Routines/UpdateDatabaseToTenDotTen.cs
+++ b/Jellyfin.Server/Migrations/Routines/UpdateDatabaseToTenDotTen.cs
@@ -1,0 +1,378 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Emby.Server.Implementations.Data;
+using Jellyfin.Database.Implementations.Entities;
+using Jellyfin.Extensions;
+using MediaBrowser.Controller;
+using MediaBrowser.Controller.Library;
+using Microsoft.Data.Sqlite;
+using Microsoft.Extensions.Logging;
+
+namespace Jellyfin.Server.Migrations.Routines;
+
+/// <summary>
+/// Updates the library.db to version 10.10.z; replaces the migration code from the old SqliteItemRepository and SqliteUserDataRepository.
+/// </summary>
+[JellyfinMigration("2025-04-19T00:00:00", nameof(UpdateDatabaseToTenDotTen), "462ED27A-8822-4115-9089-F060CB6BE32C")]
+#pragma warning disable CS0618 // Type or member is obsolete
+internal class UpdateDatabaseToTenDotTen : IMigrationRoutine
+#pragma warning restore CS0618 // Type or member is obsolete
+{
+    private const string DbFilename = "library.db";
+    private readonly ILogger<UpdateDatabaseToTenDotTen> _logger;
+    private readonly IServerApplicationPaths _paths;
+    private readonly IUserManager _userManager;
+
+    public UpdateDatabaseToTenDotTen(
+        ILogger<UpdateDatabaseToTenDotTen> logger,
+        IServerApplicationPaths paths,
+        IUserManager userManager)
+    {
+        _logger = logger;
+        _paths = paths;
+        _userManager = userManager;
+    }
+
+    /// <inheritdoc/>
+    public void Perform()
+    {
+        const string ChaptersTableName = "Chapters2";
+
+        string[] queries =
+        [
+            "create table if not exists TypedBaseItems (guid GUID primary key NOT NULL, type TEXT NOT NULL, data BLOB NULL, ParentId GUID NULL, Path TEXT NULL)",
+
+            "create table if not exists AncestorIds (ItemId GUID NOT NULL, AncestorId GUID NOT NULL, AncestorIdText TEXT NOT NULL, PRIMARY KEY (ItemId, AncestorId))",
+            "create index if not exists idx_AncestorIds1 on AncestorIds(AncestorId)",
+            "create index if not exists idx_AncestorIds5 on AncestorIds(AncestorIdText,ItemId)",
+
+            "create table if not exists ItemValues (ItemId GUID NOT NULL, Type INT NOT NULL, Value TEXT NOT NULL, CleanValue TEXT NOT NULL)",
+
+            "create table if not exists People (ItemId GUID, Name TEXT NOT NULL, Role TEXT, PersonType TEXT, SortOrder int, ListOrder int)",
+
+            "drop index if exists idxPeopleItemId",
+            "create index if not exists idxPeopleItemId1 on People(ItemId,ListOrder)",
+            "create index if not exists idxPeopleName on People(Name)",
+
+            "create table if not exists " + ChaptersTableName + " (ItemId GUID, ChapterIndex INT NOT NULL, StartPositionTicks BIGINT NOT NULL, Name TEXT, ImagePath TEXT, PRIMARY KEY (ItemId, ChapterIndex))",
+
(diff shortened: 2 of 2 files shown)


Jellyfin.Server/Migrations/Routines/UpdateDatabaseToTenDotTen.cs
```diff
@@ -0,0 +1,378 @@
...
+                {
+                    continue;
+                }
+
+                statement.TryBind("@UserId", user.Id);
+                statement.TryBind("@InternalUserId", user.InternalId);
+
+#pragma warning disable IDE0058 // Expression value is never used
+                statement.ExecuteNonQuery();
+#pragma warning restore IDE0058 // Expression value is never used
```
Bond-009, line 310:
```suggestion
                _ = statement.ExecuteNonQuery();
```

Jellyfin.Server/Migrations/JellyfinMigrationService.cs
```diff
@@ -93,35 +95,33 @@ public async Task CheckFirstTimeRunOrMigration(IApplicationPaths appPaths)
 
             logger.LogInformation("Migration system initialisation completed.");
         }
-        else
+
```
JPVenson, line 98: When the startup is not completed yet, we dont need to run this, please revert

Jellyfin.Server/Migrations/Routines/UpdateDatabaseToTenDotTen.cs
```diff
@@ -0,0 +1,376 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Jellyfin.Server.Migrations.Routines;
+
+/// <summary>
+/// Updates the library.db to version 10.10.z; replaces the migration code from the old SqliteItemRepository and SqliteUserDataRepository.
+/// </summary>
+[JellyfinMigration("2025-04-19T00:00:00", nameof(UpdateDatabaseToTenDotTen), "462ED27A-8822-4115-9089-F060CB6BE32C")]
+#pragma warning disable CS0618 // Type or member is obsolete
+internal class UpdateDatabaseToTenDotTen : IMigrationRoutine
```
JPVenson, line 20: The old IMigrationRoutine is obsolete, see message

Jellyfin.Server/Migrations/Routines/UpdateDatabaseToTenDotTen.cs
```diff
@@ -0,0 +1,376 @@
...
+using MediaBrowser.Controller.Library;
+using Microsoft.Data.Sqlite;
+using Microsoft.Extensions.Logging;
+
+namespace Jellyfin.Server.Migrations.Routines;
+
+/// <summary>
+/// Updates the library.db to version 10.10.z; replaces the migration code from the old SqliteItemRepository and SqliteUserDataRepository.
+/// </summary>
+[JellyfinMigration("2025-04-19T00:00:00", nameof(UpdateDatabaseToTenDotTen), "462ED27A-8822-4115-9089-F060CB6BE32C")]
```
JPVenson, line 18: Guid is obsolete see ctor comment

Jellyfin.Server/Migrations/Routines/UpdateDatabaseToTenDotTen.cs
```diff
@@ -0,0 +1,376 @@
...
+            "create index if not exists idx_ItemValues8 on ItemValues(Type, ItemId, Value)",
+
+            "CREATE INDEX IF NOT EXISTS idx_TypedBaseItemsUserDataKeyType ON TypedBaseItems(UserDataKey, Type)",
+            "CREATE INDEX IF NOT EXISTS idx_PeopleNameListOrder ON People(Name, ListOrder)"
+        ];
+
+        string[] userDataQUeries =
+        [
+            "create table if not exists UserDatas (key nvarchar not null, userId INT not null, rating float null, played bit not null, playCount int not null, isFavorite bit not null, playbackPositionTicks bigint not null, lastPlayedDate datetime null, AudioStreamIndex INT, SubtitleStreamIndex INT)",
+            "drop index if exists idx_userdata",
```
JPVenson, line 117: No need to drop those, the database will never be used again.

Jellyfin.Server/Migrations/Routines/UpdateDatabaseToTenDotTen.cs
```diff
@@ -0,0 +1,376 @@
...
+            "create index if not exists UserDatasIndex2 on UserDatas (key, userId, played)",
+            "create index if not exists UserDatasIndex3 on UserDatas (key, userId, playbackPositionTicks)",
+            "create index if not exists UserDatasIndex4 on UserDatas (key, userId, isFavorite)",
+            "create index if not exists UserDatasIndex5 on UserDatas (key, userId, lastPlayedDate)"
+        ];
+
+        _logger.LogInformation("Prepare database for EFCore migration: update to version 10.10.z");
+
+        var dataPath = _paths.DataPath;
+
```
JPVenson, line 134: it is common for library.db migrations to backup the db before modifications are done

Jellyfin.Server/Migrations/Routines/UpdateDatabaseToTenDotTen.cs
```diff
@@ -0,0 +1,376 @@
...
+            "create table if not exists mediaattachments (ItemId GUID, AttachmentIndex INT, Codec TEXT, CodecTag TEXT NULL, Comment TEXT NULL, Filename TEXT NULL, MIMEType TEXT NULL, PRIMARY KEY (ItemId, AttachmentIndex))",
+
+            "pragma shrink_memory"
+        ];
+
+        string[] postQueries =
+        [
+            "create index if not exists idx_PathTypedBaseItems on TypedBaseItems(Path)",
```
JPVenson, line 69: adding indexes is not nessesary, the database will not be used afterwards

JPVenson (changes requested):
first review

Jellyfin.Server/Migrations/JellyfinMigrationService.cs
TheMelmacian: To test the migration, I checked out 10,8.z and started it once to create a new database.
After that, I checked out the master branch and every time the server started it failed with `SQLite Error 1: 'no such table: ActivityLog'` cause the `MigrateActivityLogDb` routine was running, although it was marked as applied in the existing `migration.xml`.

Since we can't be sure if the instance that is updated was already set up, or is a clean install, we can't rely on the `IsStartupWizardCompleted` flag. This is maybe not very common, but we shouldn't force users to set up the server before they can install a new version. And the migration should be robust enough to take all possible server states into account.

Jellyfin.Server/Migrations/Routines/UpdateDatabaseToTenDotTen.cs
TheMelmacian: All existing migration routines, even the new ones like `MigrateLibraryDB` have a GUID applied ¯\_(ツ)_/¯
But I will remove it, no problem.

Jellyfin.Server/Migrations/Routines/UpdateDatabaseToTenDotTen.cs
TheMelmacian: I just applied the complete code from the old SqliteRepositories without thinking about it much.
But they could speed up EFCore migration on large libraries since many selects with EXISTS and sub selects are executed in the routine.

JPVenson: Not really. The only other migration that happens after this is effectively the library.db -> jellyfin.db migration so we can really disregard anything thats not just to bring the library.db up to standard.

Jellyfin.Server/Migrations/Routines/UpdateDatabaseToTenDotTen.cs
```diff
@@ -0,0 +1,324 @@
...
+    }
+
+    /// <inheritdoc/>
+    public void Perform()
+    {
+        const string ChaptersTableName = "Chapters2";
+
+        string[] queries =
+        [
+            "create table if not exists TypedBaseItems (guid GUID primary key NOT NULL, type TEXT NOT NULL, data BLOB NULL, ParentId GUID NULL, Path TEXT NULL)",
```
JPVenson, line 44: at this stage the tables should already exist. We are migrating, so that means its not a new install and on new installs this migration should not run either. If its an actual migration, those tables have to already exist in the first place.

JPVenson (changes requested):

JPVenson:
As we just discussed this internally, i wanted to give you overall feedback. The reason i did not just copy-paste all of the old migration code when i did the refactor was that nobody wants to support that code because its ugly, hard to understand and overall a bit too complicated to our liking. In its current form, it contains a lot of dead code (as you already noticed from my other comments) and if we want to achieve the goal of allowing <10.10 versions to migrate it has to be done in a way any developer of the team can possibly manage to work on it which the current code does not allow (therefor its complete removal in the first place).

I am totally not against the idea overall i just hope that gives you context what is expected from this.

oddstr13:
My production install is ~10.3 era, currently running 10.9.11.
I'm intending to test this once it is deemed mergeable (or close to)

joshuaboniface:
I actually think this is a major release blocker as I mentioned in chat previously. That said, this can also just be a one-time thing that we rip out after 10.11 so I'm not entirely sure the maintenance burden concern is a big one. Once this is in, it should never need to be changed again because the entire goal is just to get the DB to a state where the "Real" new migrations can run.

JPVenson:
@joshuaboniface no we can never rip this one out again otherwise migrations would then not work in the future anyway.

sjorge:
I would think covering this in the release notes should be sufficient that you can't skip some versions in the upgrade path. However the recent security notice in the release notes and the subsequent questions about why things weren't working makes be doubt that would be sufficient :(

What currently happens if you skip versions? Do you end up with a broken install or can you move to the intermediate version, then to the latest to get out of the situation?

JPVenson:
@sjorge yes the update is non-destructive and can be reattempted infinite

sjorge:
> @sjorge yes the update is non-destructive and can be reattempted infinite

Great, if that happens is the message clear on what they need to do? I personally think having it covered in the release notes/blog announcement should be sufficient.

JPVenson:
It is _distinct_ but not so clear for most users.

JPVenson:
@TheMelmacian just a quick question, are you willing to keep working on this PR given you are now aware of the scope and requirements?

TheMelmacian:
Yes, I will continue to work on this.
I already made most requested changes and I found out that the UserDatas code didn't change since 10.1.0 so it's probably safe to remove it completely.
Not sure how the ADD COLUMN part can be simplified. If it were just the table names we could use a loop instead of one row for each column. But with the different types a dictionary would be necessary to map types and column names and I'm not sure if the code would be easier to understand than now. I will look into it.

JPVenson:
merge conflicts still

TheMelmacian:
> merge conflicts still
Sorry, didn't recognize there were new conflicts when I re-requested your review. Should be fixed now.

Any reason why you added the using directives for SystemBackupService in the JellyfinMigrationService? My IDE shows them as unused.

JPVenson:
No i think the using is just obsolete. But also the fix you made in the migration service has been applied by me already, so you can rollback your changes to the migration service

TheMelmacian:
I already adapted your Key null check and reverted mine.
The change that the migration of an existing `migration.xml` always runs is for a completely different error, as described above.
I tested it again before I pushed the latest commit. Without that change the server start fails on a new install, since it tries to rerun already applied migrations.

oddstr13:
@coderabbitai review

Jellyfin.Server/Migrations/Routines/UpdateDatabaseToTenDotTen.cs
oddstr13: Is this index not used at all during the library.db -> jellyfin.db migration?

oddstr13:
> I actually think this is a major release blocker as I mentioned in chat previously. That said, this can also just be a one-time thing that we rip out after 10.11 so I'm not entirely sure the maintenance burden concern is a big one. Once this is in, it should never need to be changed again because the entire goal is just to get the DB to a state where the "Real" new migrations can run.

@joshuaboniface I don't think this should be ripped out again anytime soon, but I also don't see much that likely would be in need of maintenance here.
Removal of this migration should only be done (1) when we wish to force admins to step thru a specific release before continuing updates – the current state without this PR is «You MUST upgrade to 10.10 and start the server to let it complete migrations before upgrading to 10.11», (or 2) when we're fairly certain 99.9% of existing installs have already been upgraded.


Presumably, this migration code wouldn't even be ran unless we are upgrading from an older version?
That is, on a new install from master/10.11 RC with this patch applied, the migration added here wouldn't even run?

jumoog:
I can see that this affects roughly 450 - 500 intro-skipper users.

TheMelmacian:
It seem the jellyfin-bot doesn't recognize that the merge conflicts have been resolved?
At least for me GitHub is no longer showing the `This branch has conflicts that must be resolved` message, but the label wasn't removed.

Is there an update on what is planned going forward?
Do you plan to merge this PR (in any form), or did you decide to force the users to first update to 10.10.7?

JPVenson:
we were still internally debating heavily on this, but i would say currently we tend towards not adding pre 10.9.X support.

theguymadmax:
@TheMelmacian , thanks for putting the effort into this PR. Ultimately, we decided not to add support for systems older than 10.9.x to the migration process. 

When you first started the PR, migration from 10.8.x systems would brick existing installs when it failed. Now that the database is properly backed up, users can roll back and perform an incremental upgrade before attempting a 10.11.x upgrade.

Given the already existing migration issues we are currently facing, the added complexity of not supporting even older systems was probably a good idea.

PR #14052, opened 2025-05-03
Enable metrics via an environmental variable, optionally separate Port

### LINKED ISSUE #6247: Possible to set EnableMetrics via environment variable?

cbcoutinho:
**Changes**
This change builds on #10503, enabling exposing prometheus metrics via an environmental variable rather than updating the system configuration file. This is necessary in containerized environments where the configuration file is not available at startup and requires some workarounds when deploying to Kubernetes. Additionally, users can specify a separate HTTP port to host the `/metrics` endpoint. If set to `0` - the default - the `/metrics` endpoint will be exposed on the default server port.

Users can set the following environmental variables to configure metrics at startup:
- `JELLYFIN_EnableMetrics`
- `JELLYFIN_MetricsListenPort`

As requested in the original PR, the changes have been moved to `ServerConfigurationManager`.

**Issues**
Fixes #6247


Shadowghost:
I wonder if it wouldn't make more sense to also move the metrics to their own port

JPVenson:
@Shadowghost what would be the benefit of that?

cbcoutinho:
Hi @Shadowghost - I was curious about that as well. I've updated my PR to include an optional configuration parameter that allows you to set a different port for the `/metrics` endpoint. By default (or if set to 0) it will use the default port of 8096.

You can set a different port in the config file or via the environmental variable `JELLYFIN_MetricsListenPort`

Shadowghost:
@JPVenson usually you don't want you metrics exposed to the public, same with health endpoints.

JPVenson:
@Shadowghost then lets put that behind auth instead

Shadowghost:
That would break any automated system, e.g. in docker or k8s

JPVenson:
i thought we dont even have metrics yet as i dont see anywhere in current master where its enabled.
Can't break stuff thats not yet working anyway^^

cbcoutinho:
I was able to test the metrics endpoint without a client and both the default/non-default ports work. Please let know if there's anything else you would like, otherwise IMHO it's ready for review

Emby.Server.Implementations/Configuration/ServerConfigurationManager.cs
```diff
@@ -17,18 +18,54 @@ namespace Emby.Server.Implementations.Configuration
...
+        /// <param name="startupConfig">The startup configuration containing environment variables.</param>
         public ServerConfigurationManager(
             IApplicationPaths applicationPaths,
             ILoggerFactory loggerFactory,
-            IXmlSerializer xmlSerializer)
+            IXmlSerializer xmlSerializer,
+            IConfiguration startupConfig) // Add IConfiguration parameter
             : base(applicationPaths, loggerFactory, xmlSerializer)
         {
+            _startupConfig = startupConfig; // Store the configuration
```
Shadowghost, line 37:
```suggestion
            IConfiguration startupConfig)
            : base(applicationPaths, loggerFactory, xmlSerializer)
        {
            _startupConfig = startupConfig;
```

Emby.Server.Implementations/Configuration/ServerConfigurationManager.cs
```diff
@@ -17,18 +18,54 @@ namespace Emby.Server.Implementations.Configuration
...
+
+            // Check for environment variable override AFTER base constructor loads XML
+            // You can enable metrics by setting the JELLYFIN_EnableMetrics environmental variable
+            if (bool.TryParse(_startupConfig["EnableMetrics"], out var enableMetricsEnv) && enableMetricsEnv)
+            {
+                if (!Configuration.EnableMetrics) // Only override if not already true from XML
```
Shadowghost, line 43: I'd argue that a CLI argument should always override any configuration, even if it diables

Jellyfin.Server/Program.cs
```diff
@@ -162,6 +163,24 @@ private static async Task StartServer(IServerApplicationPaths appPaths, StartupO
                     .ConfigureServices(services => appHost.Init(services))
                     .ConfigureWebHostDefaults(webHostBuilder =>
                     {
+                        // Configure Kestrel to listen on the dedicated metrics port if specified
+                        var serverConfig = appHost.ConfigurationManager.Configuration;
+                        if (serverConfig.EnableMetrics && serverConfig.MetricsListenPort > 0)
```
Shadowghost, line 168: This would expose the whole Jellyfin instance on the metrics port, no?

cbcoutinho: Yes, configuring a separate `MetricsListenerPort` would result in the entire backend listening on both ports, with the additional `/metrics` endpoint available on the new additional port. For what it's worth, I was able to connect the server on the metrics port from the web client and configure the server via the web ui. I'm not sure what the downsides of this are in practice, and I didn't test thoroughly to notice a difference.

I don't plan on exposing the metrics port beyond an internal prometheus instance, so I'm not concerned with the implicit behavior of jellyfin being exposed on an additional non-standard port as well.

Emby.Server.Implementations/Configuration/ServerConfigurationManager.cs
cbcoutinho: I agree, I've updated this section to make it possible to disable metrics via environmental variables as well

moritz31:
anything we can do to support this issue ? Would exactly need this behaviour

Emby.Server.Implementations/Configuration/ServerConfigurationManager.cs
```diff
@@ -17,18 +18,55 @@ namespace Emby.Server.Implementations.Configuration
...
             ILoggerFactory loggerFactory,
-            IXmlSerializer xmlSerializer)
+            IXmlSerializer xmlSerializer,
+            IConfiguration startupConfig)
             : base(applicationPaths, loggerFactory, xmlSerializer)
         {
+            _startupConfig = startupConfig;
+
+            // Check for environment variable override AFTER base constructor loads XML
+            // You can enable metrics by setting the JELLYFIN_EnableMetrics environmental variable
```
jpds, line 40: I think the convention, at least on Linux, is that environment variables are `ALL_CAPS_ALL_THE_WAY=true`

cbcoutinho: Hi @jpds - I agree for env vars on Linux this is typically the convention; however, for this project I noticed that the env vars use `CamelCase`. I'm assuming it to be a Windows/C# convention

If any project maintainers can chime in, I'd be happy to make any required changes.

Skaronator: > Environment variable names are case-sensitive on Unix-like systems but aren't case-sensitive on Windows.

Since Windows does not care they should be, in my opionon, uppercase only.

[Source](
https://learn.microsoft.com/en-us/dotnet/fundamentals/runtime-libraries/system-environment-getenvironmentvariable#:~:text=Environment%20variable%20names%20are%20case%2Dsensitive%20on%20Unix%2Dlike%20systems%20but%20aren%27t%20case%2Dsensitive%20on%20Windows.)

PR #14053, opened 2025-05-03
Addressing minor code optimisations

JellyCoding1:
**Changes**

- Started with a cleanup of the DictionaryExtension method 'GetFirstNotNullNorWhiteSpaceValue' + included tests
- Continued with small improvements and minor changes across some files to better understand how the components tie together.

### DIFF
--- a/Emby.Server.Implementations/EntryPoints/LibraryChangedNotifier.cs
+++ b/Emby.Server.Implementations/EntryPoints/LibraryChangedNotifier.cs
@@ -118,35 +118,34 @@ public sealed class LibraryChangedNotifier : IHostedService, IDisposable
 
         _lastProgressMessageTimes.AddOrUpdate(item.Id, _ => DateTime.UtcNow, (_, _) => DateTime.UtcNow);
 
-        var dict = new Dictionary<string, string>();
-        dict["ItemId"] = item.Id.ToString("N", CultureInfo.InvariantCulture);
-        dict["Progress"] = progress.ToString(CultureInfo.InvariantCulture);
-
-        try
-        {
-            _sessionManager.SendMessageToAdminSessions(SessionMessageType.RefreshProgress, dict, CancellationToken.None);
-        }
-        catch
-        {
-        }
-
-        var collectionFolders = _libraryManager.GetCollectionFolders(item);
-
-        foreach (var collectionFolder in collectionFolders)
-        {
-            var collectionFolderDict = new Dictionary<string, string>
+        // Code can be simplified if calling item.GetRefreshProgress() returns the same progress value as the argument provided
+        // Push item & collectionFolders into a List<BaseItem>, iterate over them and create Dictionary when calling SendMessageToAdminSessions
+        List<Dictionary<string, string>> progressMessages = [
+            new Dictionary<string, string>
+            {
+                ["ItemId"] = item.Id.ToString("N", CultureInfo.InvariantCulture),
+                ["Progress"] = progress.ToString(CultureInfo.InvariantCulture)
+            },
+            .. _libraryManager.GetCollectionFolders(item).Select(collectionFolder => new Dictionary<string, string>()
             {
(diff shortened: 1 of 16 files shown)


Emby.Server.Implementations/EntryPoints/LibraryChangedNotifier.cs
```diff
@@ -118,35 +118,34 @@ private void OnProviderRefreshProgress(object? sender, GenericEventArgs<Tuple<Ba
...
 
-            try
-            {
-                _sessionManager.SendMessageToAdminSessions(SessionMessageType.RefreshProgress, collectionFolderDict, CancellationToken.None);
-            }
-            catch
+        foreach (var progressMessage in progressMessages)
+        {
+            // Explicit Fire & Forget
+            _ = Task.Run(async () =>
```
JPVenson, line 139: Never run a task like this as it will hide all exceptions and displaces the callstack making any errors untraceable

JPVenson (changes requested):

Emby.Server.Implementations/EntryPoints/LibraryChangedNotifier.cs
JellyCoding1: I was hesitant of making this change but the original caller is synchronously;
" An exception thrown in an unawaited Task won't be caught by the caller. "

That's why I wrapped it in the Task.Run, if in the future, there is a need to add logging to this call and to avoid having the exception being thrown at some random thread.
https://dev.to/souvikk27/fire-and-forget-strategy-for-non-blocking-long-running-tasks-5b1o

Some additional googling gave me:
[HandleSafeFireAndForget](https://github.com/TheCodeTraveler/AsyncAwaitBestPractices/blob/main/src/AsyncAwaitBestPractices/SafeFireAndForgetExtensions.shared.cs)

which is another possibility of wrapping it in an async void method.

src/Jellyfin.Extensions/DictionaryExtensions.cs
```diff
@@ -41,21 +35,23 @@ public static class DictionaryExtensions
...
-
-            if (!string.IsNullOrEmpty(key3) && dictionary.TryGetValue(key3, out val) && !string.IsNullOrWhiteSpace(val))
+        /// <summary>
+        /// Gets a string from a string dictionary, checking all keys sequentially,
+        /// stopping at the first key that returns a result that's neither null nor blank.
+        /// </summary>
+        /// <param name="dictionary">The dictionary.</param>
+        /// <param name="keys">Keys to check, starting at first parameter.</param>
+        /// <returns>System.String.</returns>
+        public static string? GetFirstNotNullNorWhiteSpaceValue(this IReadOnlyDictionary<string, string> dictionary, params string[] keys)
```
Bond-009, line 47: Pretty sure `params` was not used intentionally in the original code to avoid the extra allocation

Emby.Server.Implementations/EntryPoints/LibraryChangedNotifier.cs
JellyCoding1: Exactly, there is no await in the calling code.
Following the last part of the [docs](https://learn.microsoft.com/en-us/dotnet/standard/parallel-programming/exception-handling-task-parallel-library#observing-exceptions-by-using-the-taskexception-property), they also use the ```_ = Task.Run()``` but instead of the try catch, they use ```ContinueWith``` .

If you are certain, I'll revert it since it hasn't given a problem so far

src/Jellyfin.Extensions/DictionaryExtensions.cs
JellyCoding1: True but isn't the additional allocation negligible in favor for limiting additional overloads & easier to maintain tests?

Additional question: for future PR's, is performance/allocation prioritised? Example: I am an avid LINQ user over foreach due to the readability.

Issue #13830, opened 2025-04-02
.m3u8 and .m3u files are skipped during library scan.


anilsoni85:
### Description of the bug

Jellyfin version 10.10.6 library scan doesn't recognize and add (*.m3u8, *.m3u) files to library

The best workaround is to create a .strm file with .m3u8 url in it

### Reproduction steps

1. create a new library
2. Make sure that folder has few .m3u8 files
3. After scan completes m3u8 aren't available in library

### What is the current _bug_ behavior?

m3u8 and .m3u are skipped because these extension are `NamingOPtions.VideoFileExtensions`

### What is the expected _correct_ behavior?

m3u8 and .m3u should be added into library after scan without using ,strm container file

### Jellyfin Server version

10.10.0+

### Specify version number

10.10.6

### Specify the build version

Jellyfin.Server 10.10.6.0

### Environment

```markdown
- OS: Debian 11
- Linux Kernel: 6.1.21-v8+
- Virtualization: N/A
- Clients: Browser
- Browser: Edge
- FFmpeg Version: 
- Playback Method: N/A
- Hardware Acceleration: N/A
- GPU Model: N/A
- Plugins: Standard
- Reverse Proxy: None
- Base URL: none
- Networking: NAT
- Storage: local
```

### Jellyfin logs

```shell
N/A
```

### FFmpeg logs

```shell
N/A
```

PR #14054, opened 2025-05-03
add m3u8 files support

anilsoni85:
Local m3u8 file are not detected and not loaded in library, there is workaround to use m3u8 container file .strm.
 
**Changes**
1. Adding m3u and m3u8 in array of supporting extension.
2. Local m3u, m3u8 file with http segments requires protocol whitelist.

**Issues**
#13830

### DIFF
--- a/MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
+++ b/MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
@@ -513,8 +513,8 @@ namespace MediaBrowser.MediaEncoding.Encoder
             CancellationToken cancellationToken)
         {
             var args = extractChapters
-                ? "{0} -i {1} -threads {2} -v warning -print_format json -show_streams -show_chapters -show_format"
-                : "{0} -i {1} -threads {2} -v warning -print_format json -show_streams -show_format";
+                ? "{0} -i {1} -threads {2} -v warning -print_format json -show_streams -show_chapters -show_format -protocol_whitelist \"file,https,tls,crypto,http,tcp\""
+                : "{0} -i {1} -threads {2} -v warning -print_format json -show_streams -show_format -protocol_whitelist \"file,https,tls,crypto,http,tcp\"";
 
             if (_proberSupportsFirstVideoFrame)
             {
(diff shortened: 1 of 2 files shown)


MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
```diff
@@ -513,8 +513,8 @@ private async Task<MediaInfo> GetMediaInfoInternal(
             CancellationToken cancellationToken)
         {
             var args = extractChapters
-                ? "{0} -i {1} -threads {2} -v warning -print_format json -show_streams -show_chapters -show_format"
-                : "{0} -i {1} -threads {2} -v warning -print_format json -show_streams -show_format";
+                ? "{0} -i {1} -threads {2} -v warning -print_format json -show_streams -show_chapters -show_format -protocol_whitelist \"file,https,tls,crypto,http,tcp\""
```
sjorge, line 516: What are the security implications of allow file, and some of the others? Perhaps this should be opt-in behind a checkbox somewhere.

gnattu:
Well sorry, but m3u/m3u8 as video files are not intended to be supported and the changes in this PR is also far from a complete support. First of all, m3u files are not necessarily video files, they can also just be a playlist and our library manager will also scan the m3u files for playlists, just include a file extension that can be both playlist and a video file is not the way to make the library to work properly. What's more, the validity of the m3u8 files, and the handling of such files for our clients is also a big problem and that might need to be implemented separately. It is more than "just allow a bit more protocols so that ffprobe can get the video source info". The track selection, audio/video muxing and subtitles delivery needs to work with our clients and currently it does not for a lot of the case. What is PR does is basically just blindly include those files and hope for ffmpeg and/or the clients can handle those files with its default handling logic. This may or may not work, but I'm not a big fan of accepting "may or may not work" feature like this.

Shadowghost:
I agree with what @gnattu said. This leaves us with an indefinite amount of options playback could break and I am not comfortable with that kind of "should work but if it does not, we can't really do much about it" approach.

anilsoni85:
You are right that m3u8 files are playlist and this change lets you play m3u8 file just like .strm, without .strm wrapper file. 

audio/video remux is being taken care by hls.js which is already used in client. I have tested this in browser (edge, safari) on (windows, mac, ios).

I can add checkbox somewhere in dashboard to enable it as experimental features if that helps.

PR #14056, opened 2025-05-03
Refactor Display preference manager

JPVenson:


Jellyfin.Api/Controllers/DisplayPreferencesController.cs
```diff
@@ -96,9 +96,7 @@ public ActionResult<DisplayPreferencesDto> GetDisplayPreferences(
             dto.CustomPrefs.TryAdd(key, value);
         }
 
-        // This will essentially be a noop if no changes have been made, but new prefs must be saved at least.
-        _displayPreferencesManager.SaveChanges();
-
+        _displayPreferencesManager.SetCustomItemDisplayPreferences(userId.Value, itemId, displayPreferences.Client, dto.CustomPrefs);
```
crobibero: This will cause the item's display preferences to be removed and inserted every time they are pulled even if they haven't changed

Reading through the code in master the only thing the SaveChanges is doing is saving the addition so I think this line can be removed

PR #14059, opened 2025-05-04
Stop choosing stream with "Latin" in its title as preferred language stream

a-collado:
**Changes**


Croman12:
I just installed Jellyfin and I want to select default audio language for Spanish, Castillan for a profile.. When I refresh the page the default language that is actually selected is Spanish, Latin. This is about that same behaviour or should I open a new issue?

PR #14060, opened 2025-05-04
Only selecting a forced subtitle in smart mode if it matches audio language

a-collado:
**Changes**
This PR addresses an issue where, in "Smart" mode, if a subtitle track is marked as default and as forced and is in your preferred language, it will be selected independently of the audio language of the file. I adjusted the subtitle selection in "Smart" mode to only select your preferred language forced subtitle track if it's language it's the same as the audio track language, or if the language is undefined (for example: silent movies).

### DIFF
--- a/Emby.Server.Implementations/Library/MediaStreamSelector.cs
+++ b/Emby.Server.Implementations/Library/MediaStreamSelector.cs
@@ -64,7 +64,7 @@ namespace Emby.Server.Implementations.Library
                 // If the audio language is one of the user's preferred subtitle languages behave like OnlyForced.
                 if (!preferredLanguages.Contains(audioTrackLanguage, StringComparison.OrdinalIgnoreCase))
                 {
-                    stream = sortedStreams.FirstOrDefault(x => MatchesPreferredLanguage(x.Language, preferredLanguages) && (x.Language == audioTrackLanguage || !x.IsForced));
+                    stream = sortedStreams.FirstOrDefault(x => MatchesPreferredLanguage(x.Language, preferredLanguages) && (x.Language == audioTrackLanguage || !x.IsForced || IsLanguageUndefined(audioTrackLanguage)));
                 }
                 else
                 {
--- a/Emby.Server.Implementations/Library/MediaStreamSelector.cs
+++ b/Emby.Server.Implementations/Library/MediaStreamSelector.cs
@@ -64,7 +64,7 @@ namespace Emby.Server.Implementations.Library
                 // If the audio language is one of the user's preferred subtitle languages behave like OnlyForced.
                 if (!preferredLanguages.Contains(audioTrackLanguage, StringComparison.OrdinalIgnoreCase))
                 {
-                    stream = sortedStreams.FirstOrDefault(x => MatchesPreferredLanguage(x.Language, preferredLanguages));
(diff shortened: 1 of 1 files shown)


gnattu:
>  it will be selected independently of the language of the audio track

It will not if the audio language is in your preferred language list after #13098

a-collado:
Sorry, I should have clarified I'm on the master branch, not on 10.10. 

The changes I made only affects when your preferred audio language and preferred subtitle language are different. 

Right now, if your preferred subtitle language contains the language of the audio track, it falls to `BehaviorOnlyForced`, but if not, it will try to select the `FirstOrDefault` that matches your preferred subtitle language.

In my case, my preferred audio language is Japanese and my preferred subtitle language is Spanish, and when a Spanish subtitle track is marked as forced and default, it selects it:
Captura de pantalla_20250504_210802

With the changes in this PR, when selecting the `FirstOrDefault`  subtitle track, it checks if this track is forced, and in that case, only selects it if the audio language is different from the language of the subtitle track or if the audio track language is undefined.

After the changes, in the example of the picture, when I set my audio language preference as Japanese, it selects the Japanese audio and the complete Spanish subtitle track, and when I set the audio preference to Spanish, it selects the Spanish audio and the forced Spanish subtitle track.

gnattu:
Oh I do understand now. Actually this is something that we need to be done with the sorting:
https://github.com/jellyfin/jellyfin/blob/a7bb3ea21489a3a4c9a3e9f60c9521eaa84bad31/Emby.Server.Implementations/Library/MediaStreamSelector.cs#L42-L51

Before we are doing the selection, all subtitle streams are sorted. It should be in the order of Default > No tag > Forced. As you can see, we don't want forced to be selected when there is some choices.

The problem now is that, the subtitle is being tagged as both forced and default in which, could be arguably "more preferable" as it is marked as "default". We have to be very careful here as the change we are going to make will overwrite the "default" behavior.

As an alternative to your “very specific matching,” we can also move every subtitle with a forced tag to the end of the list, regardless of whether it’s default or not. Alternatively, we could place default forced tags before other forced subtitles, but still behind no tags. My only concern is whether this approach will break another user’s use case where they know both languages and want the forced subtitles to be displayed. What may be convenient for one user might not be for another, and it’s challenging to find a solution that works for everyone in this situation. We’ll have to inform a group of users to retag their files anyway, which could be difficult due to the sheer number of files.

a-collado:
Thank you for your response!

This solution works great for my use case, but I get that it could break a lot of people's use case.

JeWe37:
I worked on this and similar changes in #11072 before(though that has been stalled for a while). There it was also pointed out that only automatically selecting forced subtitles that match the audio language is the [more standards-compliant behavior](https://github.com/jellyfin/jellyfin/pull/11072#issuecomment-1979317589).

gnattu:
> I worked on this and similar changes in #11072 before(though that has been stalled for a while). There it was also pointed out that only automatically selecting forced subtitles that match the audio language is the [more standards-compliant behavior](https://github.com/jellyfin/jellyfin/pull/11072#issuecomment-1979317589).

The spec did not specify that:
http://www.matroska.org/technical/specs/index.html, search for FlagForced.

> Applies only to subtitles. Set if that track is eligible for automatic selection by the player if it matches the user's language preference, even if the user's preferences would normally not enable subtitles with the selected audio track; this can be used for tracks containing only translations of foreign-language audio or onscreen text.

PR #14063, opened 2025-05-05
Fix cross filesystem dir moving

### LINKED ISSUE #14062: Trickplay image saving to permanent storage is broken
### Description of the bug

Images are generated and in `/tmp/jellyfin/trickpla_*` and won't move to the trickplay folder. No exception thrown.

My guess for cause would be https://github.com/jellyfin/jellyfin/pull/14023 because the build before worked and I see no other file handling/trickplay related changes.

### Reproduction steps

1. Generated trickplay
2. look into tmp folder
3. Look into data folder or try the frontend (no trickplay is loaded, but the box appears)

Images are generated but not moved to permanent storage

Store the images

### Jellyfin Server version

Master

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

-

### Environment

```markdown
- OS:
- Linux Kernel:
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

```shell
Nothing related to file handling
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

gnattu:
Regression from #14023

**Changes**

**Issues**

Fixes #14062

### DIFF
--- a/Emby.Server.Implementations/IO/ManagedFileSystem.cs
+++ b/Emby.Server.Implementations/IO/ManagedFileSystem.cs
@@ -159,13 +159,14 @@ namespace Emby.Server.Implementations.IO
             catch (IOException)
             {
                 // Cross device move requires a copy
-                var directory = Directory.CreateDirectory(destination);
-                foreach (var file in directory.EnumerateFiles())
+                Directory.CreateDirectory(destination);
+                var sourceDir = new DirectoryInfo(source);
+                foreach (var file in sourceDir.EnumerateFiles())
                 {
                     file.CopyTo(Path.Combine(destination, file.Name), true);
                 }
 
-                directory.Delete(true);
+                sourceDir.Delete(true);
             }
         }
 


JPVenson (approved):

Shadowghost (approved):

Emby.Server.Implementations/IO/ManagedFileSystem.cs
```diff
@@ -159,13 +159,14 @@ public void MoveDirectory(string source, string destination)
             catch (IOException)
             {
                 // Cross device move requires a copy
-                var directory = Directory.CreateDirectory(destination);
-                foreach (var file in directory.EnumerateFiles())
+                Directory.CreateDirectory(destination);
```
cvium, line 162: Why is this necessary? Isn't that what it already does?

gnattu: `Directory.Move` does not handle the case where the source and destination are not on the same file system (like moving cross disks) and will throw an IOException.

Bond-009 (approved):

Bond-009:
Good catch, not sure if we should just revert to the old code now as we don't use the result of `Directory.CreateDirectory` and have to create the `DirectoryInfo` ourselves.

cvium (approved):

crobibero (approved):

PR #9765, opened 2023-05-13
Convert ItemSortBy to enum

crobibero:
Requested by client devs

Draft until fully tested

PR #14064, opened 2025-05-05
Adds User Ratings

RomeoAlphaTango:
Continues work done by @TheItalianGame from this pull request: https://github.com/jellyfin/jellyfin/pull/8933

**Changes**

- Utilizes existing Rating property from UserItemData for user ratings.
- The Likes property no longer affects Rating value.
- The Rating value sets Likes property based on MinLikeValue.
- Update API to set/remove Rating value.
- Filter by MinUserRating.
- Sort by UserRating.

### DIFF
--- a/Emby.Server.Implementations/Sorting/UserRatingComparer.cs
+++ b/Emby.Server.Implementations/Sorting/UserRatingComparer.cs
@@ -1,0 +1,64 @@
+#nullable disable
+#pragma warning disable CS1591
+
+using Jellyfin.Data.Enums;
+using Jellyfin.Database.Implementations.Entities;
+using MediaBrowser.Controller.Entities;
+using MediaBrowser.Controller.Library;
+using MediaBrowser.Controller.Sorting;
+
+namespace Emby.Server.Implementations.Sorting
+{
+    /// <summary>
+    /// Class UserRatingComparer.
+    /// </summary>
+    public class UserRatingComparer : IUserBaseItemComparer
+    {
+        /// <summary>
+        /// Gets or sets the user.
+        /// </summary>
+        /// <value>The user.</value>
+        public User User { get; set; }
+
+        /// <summary>
+        /// Gets the name.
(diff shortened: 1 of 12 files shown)


Emby.Server.Implementations/Sorting/UserRatingComparer.cs
```diff
@@ -0,0 +1,69 @@
...
+using MediaBrowser.Controller.Entities;
+using MediaBrowser.Controller.Library;
+using MediaBrowser.Controller.Sorting;
+
+namespace Emby.Server.Implementations.Sorting
+{
+    /// <summary>
+    /// Class UserRatingComparer.
+    /// </summary>
+    public class UserRatingComparer : IUserBaseItemComparer
```
JPVenson, line 15: class is never used?

RomeoAlphaTango: Please let me know if this is wrong - I'm still learning the code, but since the sortBy enum of UserRating (added in this pull request) isn't a default sortBy field like ItemSortBy.ProductionYear it isn't explicitly called in the code. So I believe this class will only be used if a client sets the ItemsortBy[] sortBy field in an API call to include "UserRating". Similar to the CriticRatingComparer class.

jhparker88: > ItemSortBy.ProductionYear it isn't explicitly called in the code. So I believe this class will only be used if a client sets the ItemsortBy[] sortBy field in an API call to include "UserRating". Similar to the CriticRatingComparer class.

Yes, see #9765.

Maybe extend the OrderMapperTest that was added with #12101 to check for the added `UserRating` sortBy.

JPVenson:
closed due to inactivity

jhparker88:
@RomeoAlphaTango Do you have any intentions to bring this PR over the finish line (given that you still pushed changes three month ago)?

PR #14078, opened 2025-05-07
Fix incorrect language for Logo images

TheBosZ:
**Changes**
Modify the image provider to skip any Logos that don't have the correct (requested) language.

Problem: When fetching images for media (movies in particular), Jellyfin requests images in all languages and not just the requested language. This is fine except for Logos. If a user specifies English as the language, they most likely don't want to see Spanish Logos. 

Investigation: While tracing the request path, I found what I thought could be a solution. In [TmdbClientManager.cs](https://github.com/jellyfin/jellyfin/blob/a7bb3ea21489a3a4c9a3e9f60c9521eaa84bad31/MediaBrowser.Providers/Plugins/Tmdb/TmdbClientManager.cs#L597-L601), Backdrops with any language specified get coerced to be Thumbs instead. I figured that that would be a viable way to fix the issue.

Unfortunately, this was a dead-end as it only fixed the problem for a single provider instead of all providers. After a bit more investigation I figured out that [all languages were being fetched](https://github.com/jellyfin/jellyfin/blob/a7bb3ea21489a3a4c9a3e9f60c9521eaa84bad31/MediaBrowser.Providers/Manager/ItemImageProvider.cs#L296) for some reason. I don't know why the change was to fetch images for all languages instead of just the preferred language so I didn't feel comfortable changing it.

Solution: In ItemImageProvider, after fetching images, just remove all logos that have a language that is different than the preferred language. I'm not sure of any adverse effects of this change, but I can't imagine that it would be a problem. 

**Issues**
https://features.jellyfin.org/posts/3192/logos-should-select-the-correct-language


Shadowghost:
I wonder if it wouldn't make more sense to just sort the results accordingly with English as second language so we have a proper fallback

TheBosZ:
> I wonder if it wouldn't make more sense to just sort the results accordingly with English as second language so we have a proper fallback

I'm not sure how that would help. The problem is that there's no English logos at all, not that it's picking a different language over English.

jester-xbmc:
> **This has a lot of problems for me:**
> 
>     * If you watch anime, you also want a logo. Some anime have no English logo. I want the KO?JP logos fetched, because they are used in the episodes as well.
> 
>     * Same the other way around. If you are in a country with less maintained metadata, you will get no logos anymore.
> 
>     * Some shows have a global name and only one logo and TMDB or so set it to English and you will get the logo if your preferred language is English.
> 
> 
> For a nice solution, you need a list of acceptable languages and/or allow original language as fallback. Indie movies often have maximum one logo.
> 
> You could also mark autofetched logos as "not preferred" and schedule a retry so it can be improved.

Sorry, but the point is when you select English as your preferred language you now get a Spanish logo for certain movies (yes I had this bug several times), while you should either get an English logo or none if there aren't any logo's available in English.

Still want a logo ? that's fine, either change your preferred language (KO/JP in your case) or (better) create one and upload it so everybody wins.

In reality this is more a meta data provider issue mainly, however Jellyfin should really just use the logo for the selected preferred language not just one that happens to be available (and for the incorrect language)

TheBosZ:
To go along with what @jester-xbmc said, the problem is **unwanted** logo selections. If I **want** English logos, I don't want to see a logo in a different language. As a user, I would consider that a bug.

If there's an instance where you want a different language logo, there's a couple of solutions. Two that I think are very easy:
1. Set the preferred language for an individual media item
2. Manually download the logo with the language you want

Either of these solutions require very little effort and don't require massive code changes.

anthonyryan1:
A list of languages seems rather complex, since people might be consuming content in many different languages. They might not even remember all the languages their content is in to select it.

There seems to be two camps here, the ones who only want to see their selected language logos (and nothing wrong), and the ones who don't want missing logos.

This change fixes the first one, could it be implemented as a preference?

"Selected language only" (what this pull request achieves)
"Selected language preferred" (try the selected language, then fall back to the content language)

I think by using the content language as the fallback, it won't matter if the content is Korean or Japanese or Spanish. Because the second choice will be original production language, it's basically guaranteed to have a logo.

I feel like that could appease both camps here reasonably well, with just a two option toggle.

solbu:
To give a perspective from a Norwegian user who's native language isn't english.
I prefer Norwegian posters, failing that I prefer the english posters or the native language posters.

So if there is no norwegian poster for something, it should fetch the english poster instead.
Currently it does this, and it usually works just fine.

jester-xbmc:
> It also does this with the metadata it fetches. In my case if there is no norwegian metadata available, it fetches the english metadata. It seems that the poster fetching does something similar.
> 
> For most users it's more important to _Have_ a poster rather than nothing, just because their prefered language didn't have any.
> 
> Maybe one could implement a check button in the media library setting to enable image fetching of only the prefered language. (No, I have no idea how easy this would be to implement, It's just an idea)

Sorry, disagree, and lets not mix the 2.
For users is it important to get the CORRECT meta data of the language they have actively specified.

Put differently: new user adds 1700 movies with media library set to English, with the current bug it will pull in logo's for whatever other language is available if English isn't available, ie the new user ends up with an X amount of movies with the logo set to the wrong language, new user confused, user thought the right language was selected, also, enjoy manually cleaning up the wrong language logo's... 

It is maybe a personal preference that you don't mind if it pulls in another language, but it is certainly not expected from a user point of view.

As said, you want a fall-back language ? no problem, open a PR for that, however this PR is addressing a bug, let's fix it and not muddy the waters.

JPVenson:
closed due to inactivity

PR #14079, opened 2025-05-08
Use VBR and MBBRC in QSV encoders for better quality

nyanmisaka:
**Changes**
- Use VBR and MBBRC in QSV encoders for better quality

Use VBR for better bitrate allocation. Enable MacroBlock level bitrate control for better subjective visual quality.

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -1572,6 +1572,26 @@ namespace MediaBrowser.Controller.MediaEncoding
                 return FormattableString.Invariant($" -maxrate {bitrate} -bufsize {bufsize}");
             }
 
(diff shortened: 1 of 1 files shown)


Bond-009 (approved):

Shadowghost (approved):

crobibero (approved):

oysstu:
Does VBR work with low power encoding? I thought at least older generations of Intel required CQP bitrate control (gen7 ice lake and prior).

nyanmisaka:
VBR always works. Low-power mode switches the rate control algorithm from using shaders to fixed-function hardware.

Jellyfin doesn't use CQP. The 7th generation is Kaby Lake with gen9 graphics. The 10th generation mobile is Ice Lake with gen11 graphics.

oysstu:
I see, thanks for the reply. I think I mixed up VBR with the ICQ bitrate control mode.

PR #14085, opened 2025-05-09
Prune trickplay data on regenerate and scan

### LINKED ISSUE #13918: Unable to Disable Trickplay After Generating Thumbnails
### Description of the bug

Once Trickplay is enabled and thumbnails are generated for a movie, there is no way to revert the process through the Jellyfin interface. Even if the "Enable Trickplay thumbnail extraction" option is disabled and metadata is refreshed or replaced, the thumbnails still appear on the progress bar.

After disabling Trickplay and refreshing the library, the progress bar should return to its original state, showing only the time indicator without thumbnails.

Trickplay thumbnails continue to appear, or a grey box with a timestamp is shown (if the images were deleted manually). There is no way to revert to the original display without thumbnails, making the change effectively irreversible through the UI.

Conclusion:

Once Trickplay is enabled and thumbnails are generated, it is not possible to fully revert the behavior.
It would be ideal to have an option to restore the classic progress bar display.

### Reproduction steps

1- Enable Trickplay and allow thumbnails to be generated.

2- Later, disable the Trickplay option in the settings.

3- Force a library scan or metadata replacement.

4- Check the progress bar of an affected movie.

After enabling and generating Trickplay thumbnails for a movie, it is not possible to revert to the original progress bar behavior. Disabling the "Enable video preview thumbnails" option and refreshing the library, replacing metadata, or even manually deleting the generated Trickplay images does not restore the original minimal progress bar that shows only the current timestamp.

Instead, even after removing the thumbnails, a grey Trickplay box with the time appears in the progress bar, and the user is unable to restore the classic display. There appears to be no way within the UI or via library refresh to fully disable or remove Trickplay once it has been generated.

After disabling the "Enable video preview thumbnails" option and refreshing or replacing the metadata of the library, the movie’s progress bar should return to its original state — showing only the current timestamp without displaying Trickplay previews. Additionally, if the Trickplay images are manually deleted, Jellyfin should detect their absence and revert to the classic progress bar automatically.

### Jellyfin Server version

10.10.0+

### Specify commit id

_No response_

_No response_

### Specify version number

10.10.7

### Specify the build version

10.10.7

### Environment

```markdown
- OS:
- Linux Kernel:
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

```shell
no logs
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

Shadowghost:
**Changes**
* If trickplay is disabled for library all trickplay data is pruned
* If we refresh, all trickplay data will be removed and then for the current widths recreated
* If we just run, all trickplay data which is not matching the current configuration is removed
* When running the scheduled task we also prune any data for libraries where trickplay was disabled

**Issues**
Fixes #13918

### DIFF
--- a/Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
+++ b/Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
@@ -7,6 +7,7 @@ using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using AsyncKeyedLock;
+using J2N.Collections.Generic.Extensions;
 using Jellyfin.Database.Implementations;
 using Jellyfin.Database.Implementations.Entities;
 using MediaBrowser.Common.Configuration;
@@ -80,7 +81,12 @@ public class TrickplayManager : ITrickplayManager
     public async Task MoveGeneratedTrickplayDataAsync(Video video, LibraryOptions libraryOptions, CancellationToken cancellationToken)
     {
         var options = _config.Configuration.TrickplayOptions;
-        if (!CanGenerateTrickplay(video, options.Interval, libraryOptions))
+        if (!CanGenerateTrickplay(video, options.Interval))
+        {
+            return;
+        }
+
+        if (libraryOptions is null || !libraryOptions.EnableTrickplayImageExtraction)
         {
             return;
         }
@@ -137,9 +143,40 @@ public class TrickplayManager : ITrickplayManager
     /// <inheritdoc />
     public async Task RefreshTrickplayDataAsync(Video video, bool replace, LibraryOptions libraryOptions, CancellationToken cancellationToken)
     {
+        var options = _config.Configuration.TrickplayOptions;
+        if (!CanGenerateTrickplay(video, options.Interval) || libraryOptions is null)
+        {
+            return;
+        }
+
+        using var dbContext = await _dbProvider.CreateDbContextAsync(cancellationToken).ConfigureAwait(false);
+        var saveWithMedia = libraryOptions.SaveTrickplayWithMedia;
+        var trickplayDirectory = _pathManager.GetTrickplayDirectory(video, saveWithMedia);
+        if (!libraryOptions.EnableTrickplayImageExtraction || replace)
+        {
+            // Prune existing data
+            try
+            {
+                Directory.Delete(trickplayDirectory, true);
+            }
+            catch (Exception ex)
+            {
+                _logger.LogWarning("Unable to clear trickplay directory: {Directory}: {Exception}", trickplayDirectory, ex);
+            }
+
+            await dbContext.TrickplayInfos
+                    .Where(i => i.ItemId.Equals(video.Id))
+                    .ExecuteDeleteAsync(cancellationToken)
+                    .ConfigureAwait(false);
+
+            if (!replace)
+            {
+                return;
+            }
+        }
+
         _logger.LogDebug("Trickplay refresh for {ItemId} (replace existing: {Replace})", video.Id, replace);
 
-        var options = _config.Configuration.TrickplayOptions;
         if (options.Interval < 1000)
         {
             _logger.LogWarning("Trickplay image interval {Interval} is too small, reset to the minimum valid value of 1000", options.Interval);
@@ -154,8 +191,29 @@ public class TrickplayManager : ITrickplayManager
                 replace,
                 width,
                 options,
-                libraryOptions,
+                saveWithMedia,
                 cancellationToken).ConfigureAwait(false);
+        }
+
+        // Cleanup old trickplay files
+        var existingFolders = Directory.GetDirectories(trickplayDirectory).ToList();
+        var trickplayInfos = await dbContext.TrickplayInfos
+                .AsNoTracking()
+                .Where(i => i.ItemId.Equals(video.Id))
+                .ToListAsync(cancellationToken)
+                .ConfigureAwait(false);
+        var expectedFolders = trickplayInfos.Select(i => GetTrickplayDirectory(video, i.TileWidth, i.TileHeight, i.Width, saveWithMedia)).ToList();
+        var foldersToRemove = existingFolders.Except(expectedFolders);
+        foreach (var folder in foldersToRemove)
+        {
+            try
+            {
+                Directory.Delete(folder, true);
+            }
+            catch (Exception ex)
+            {
+                _logger.LogWarning("Unable to remove trickplay directory: {Directory}: {Exception}", folder, ex);
+            }
         }
     }
 
@@ -164,14 +222,9 @@ public class TrickplayManager : ITrickplayManager
         bool replace,
         int width,
         TrickplayOptions options,
-        LibraryOptions libraryOptions,
+        bool saveWithMedia,
         CancellationToken cancellationToken)
     {
-        if (!CanGenerateTrickplay(video, options.Interval, libraryOptions))
-        {
-            return;
-        }
-
         var imgTempDir = string.Empty;
 
         using (await _resourcePool.LockAsync(cancellationToken).ConfigureAwait(false))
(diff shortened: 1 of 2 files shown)


theguymadmax:
Scenario 1:
1. Set up media library
2. Save trickplay next to media
3. Forget to enable trickplay for library
4. Run trickplay generation 
Results:
  All trickplays next to media are deleted, even though the task is called "Generate Trickplay Images"

Possible solution:
Keep the existing task behavior as is and create a new task "Remove Trickplay Images"

Shadowghost:
I'd rather rename the task and its descriptin than creating yet another task 🤔

theguymadmax:
Maybe add it to "Migrate Trickplay Image Location" and rename to "Migrate & Clean Trickplay Image Location"

Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
```diff
@@ -80,7 +81,12 @@ public TrickplayManager(
     public async Task MoveGeneratedTrickplayDataAsync(Video video, LibraryOptions libraryOptions, CancellationToken cancellationToken)
     {
         var options = _config.Configuration.TrickplayOptions;
-        if (!CanGenerateTrickplay(video, options.Interval, libraryOptions))
+        if (!CanGenerateTrickplay(video, options.Interval))
+        {
+            return;
+        }
+
+        if (libraryOptions is null || !libraryOptions.EnableTrickplayImageExtraction)
```
JPVenson, line 89: shouldn't we check for that before anything else to avoid any unnessesary operation?

Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
```diff
@@ -137,9 +143,40 @@ private void MoveContent(string sourceFolder, string destinationFolder)
     /// <inheritdoc />
     public async Task RefreshTrickplayDataAsync(Video video, bool replace, LibraryOptions libraryOptions, CancellationToken cancellationToken)
     {
+        var options = _config.Configuration.TrickplayOptions;
+        if (!CanGenerateTrickplay(video, options.Interval) || libraryOptions is null)
+        {
+            return;
+        }
+
+        using var dbContext = await _dbProvider.CreateDbContextAsync(cancellationToken).ConfigureAwait(false);
```
JPVenson, line 152: async using

JPVenson (approved):

theguymadmax:
### Scenario 1
Trickplay extraction not enabled 
Trickplay saved next to media enabled 
Generate trickplays 

Results: Trickplay files next to media delted  -- Results not expected

Logs: 
```
[15:56:54] [DBG] [12] Emby.Server.Implementations.ScheduledTasks.TaskManager: Executing Generate Trickplay Images
[15:56:55] [INF] [12] Emby.Server.Implementations.ScheduledTasks.TaskManager: Generate Trickplay Images Completed after 0 minute(s) and 0 seconds
```

### Scenario 2
Trickplay extraction not enabled 
Trickplay saved next to media not enabled 
Generate trickplays 

Results: No files deleted -- Expected

Logs: 
```
[15:47:37] [WRN] [18] Jellyfin.Server.Implementations.Trickplay.TrickplayManager: Unable to clear trickplay directory: C:\Users\Mac\AppData\Local\jellyfin trick\data\trickplay\fb\fb5adda7-9d6a-5d41-b2ea-d3632d30ad1b: System.IO.DirectoryNotFoundException: Could not find a part of the path 'C:\Users\Max\AppData\Local\jellyfin trick\data\trickplay\fb\fb5adda7-9d6a-5d41-b2ea-d3632d30ad1b'.
   at System.IO.FileSystem.GetFindData(String fullPath, Boolean isDirectory, Boolean ignoreAccessDenied, WIN32_FIND_DATA& findData)
   at System.IO.FileSystem.RemoveDirectory(String fullPath, Boolean recursive)
   at Jellyfin.Server.Implementations.Trickplay.TrickplayManager.RefreshTrickplayDataAsync(Video video, Boolean replace, LibraryOptions libraryOptions, CancellationToken cancellationToken) in C:\git\jellyfin\Jellyfin.Server.Implementations\Trickplay\TrickplayManager.cs:line 157
[15:47:37] [INF] [18] Emby.Server.Implementations.ScheduledTasks.TaskManager: Generate Trickplay Images Completed after 0 minute(s) and 0 second
```

### Scenario 3
Trickplay extraction enabled
Generated trickplay in metadata folder from the scheduled task
Disable trickplay image extraction
Generated trickplay from scheduled task

Results: Trickplays files and db entries are deleted  -- Expected

Logs:
```
[16:14:50] [DBG] [17] Emby.Server.Implementations.ScheduledTasks.TaskManager: Executing Generate Trickplay Images
[16:14:50] [DBG] [10] Jellyfin.Api.Auth.CustomAuthenticationHandler: AuthenticationScheme: CustomAuthentication was successfully authenticated.
[16:14:50] [INF] [17] Emby.Server.Implementations.ScheduledTasks.TaskManager: Generate Trickplay Images Completed after 0 minute(s) and 0 seconds
```



### Scenario 4
Trickplay extraction enabled
Generated trickplay in metadata folder from the scheduled task
Disable trickplay image extraction
Manually delete trickplay folder from the metadata folder
Generated trickplay from scheduled task

Results: Db entries are removed  -- Expected

Logs:
```
[16:23:46] [DBG] [3] Emby.Server.Implementations.ScheduledTasks.TaskManager: Executing Generate Trickplay Images
[16:23:47] [WRN] [3] Jellyfin.Server.Implementations.Trickplay.TrickplayManager: Unable to clear trickplay directory: C:\Users\Max\AppData\Local\jellyfin trick\data\trickplay\fb\fb5adda7-9d6a-5d41-b2ea-d3632d30ad1b: System.IO.DirectoryNotFoundException: Could not find a part of the path 'C:\Users\Max\AppData\Local\jellyfin trick\data\trickplay\fb\fb5adda7-9d6a-5d41-b2ea-d3632d30ad1b'.
   at System.IO.FileSystem.GetFindData(String fullPath, Boolean isDirectory, Boolean ignoreAccessDenied, WIN32_FIND_DATA& findData)
   at System.IO.FileSystem.RemoveDirectory(String fullPath, Boolean recursive)
   at Jellyfin.Server.Implementations.Trickplay.TrickplayManager.RefreshTrickplayDataAsync(Video video, Boolean replace, LibraryOptions libraryOptions, CancellationToken cancellationToken) in C:\git\jellyfin\Jellyfin.Server.Implementations\Trickplay\TrickplayManager.cs:line 157
[16:23:47] [INF] [3] Emby.Server.Implementations.ScheduledTasks.TaskManager: Generate Trickplay Images Completed after 0 minute(s) and 0 seconds
```

In Scenario 1, the fact that not enabling an option (Trickplay image extraction) results in file deletion is wrong. If a user enables "trickplays saved next to media" but forgets to also enable image extraction, any existing trickplay files are deleted when the "Generate Trickplays" task runs, either automatically or by the user. Configuration settings should never lead to destructive actions by omission. File deletion should always be explicit, intentional, and well-communicated.

Also, there should be a log entry when files are being removed.

crobibero:
I like the idea of having the "Move Trickplay files" task being updated to purge any files that shouldn't be there since that isn't an automatic action by default. We for sure shouldn't automatically remove the Trickplay files on scan if not enabled at the library level since they can take a very long time to generate

Shadowghost:
@enter-a-random-username this is intentional, we don't know what changed in the file, we just know that it changed.

ferferga:
> I like the idea of having the "Move Trickplay files" task being updated to purge any files that shouldn't be there since that isn't an automatic action by default. We for sure shouldn't automatically remove the Trickplay files on scan if not enabled at the library level since they can take a very long time to generate

In my opinion, on scan we should remove all stale files. I think users assume that scan makes all the "internal state" of Jellyfin match their filesystem.

Then, the create trickplay task will generate whatever new ones are missing (if the user doesn't have it enabled to be part of the scanning process).

In my opinion we have way too many tasks right now, we should simplify as much as possible.

crobibero (approved):

theguymadmax:
> @Shadowghost / @theguymadmax Is scenario 1 still an issue?

Yes, that's still the current behavior of this PR.

Shadowghost:
@crobibero what would be your proposed fix for scenario 1? If we don't clean it up we also don't import

crobibero:
Upon second review I think that scenario 1 is ok. If the user disables trickplay extraction and then runs generate I think it is ok to remove the files.

This may be revisited in the future based on user feedback

Issue #13965, opened 2025-04-22
Transcoding Media sometimes results in the video being delayed by a few seconds


PeachesMLG:
### Description of the bug

When watching some media, sometimes the video will be delayed by about 5 seconds or so

You can see watching this on Jellyfin I get this
![Image](https://github.com/user-attachments/assets/431c7456-2db7-463f-84ac-20af5735258d)

However opening the raw file on VLC Media player for example, at the same timestamp I get this

![Image](https://github.com/user-attachments/assets/2290402b-c9d9-4555-9531-1ccb60814388)

### Reproduction steps

No idea, seems to happen at random

### What is the current _bug_ behavior?

How is this different from Description of the bug

### What is the expected _correct_ behavior?

For the Video and subtitles to be in sync

### Jellyfin Server version

10.10.0+

### Specify the build version

10.10.7

### Environment

```markdown
- OS: Ubuntu
- Linux Kernel:
- Virtualization: Docker
- Clients: Browser
- Browser: Firefox
- FFmpeg Version: Whatever is shipped with Docker
- Playback Method: Transcoding
- Hardware Acceleration: Nope
- GPU Model:
- Plugins: 
- Reverse Proxy: Caddy
- Base URL:
- Networking:
- Storage:


If useful here is the Playback Information

Playback Info
Player
Html Video Player
Play method
Transcoding
Protocol
https
Stream type
HLS
Video Info
Player dimensions
1720x1281
Video resolution
1920x1080
Dropped frames
0
Corrupted frames
0
Transcoding Info
Video codec
H264
Audio codec
AAC (direct)
Audio channels
2
Bitrate
6.1 Mbps
Transcoding progress
78.1%
Transcoding framerate
203 fps (8.47x)
Reason for transcoding
The container is not supported
Original Media Info
Container
mkv
Size
636.9 MiB
Bitrate
3.9 Mbps
Video codec
HEVC Main
Video bitrate
3.8 Mbps
Video range type
SDR
Audio codec
AAC LC
Audio bitrate
128 kbps
Audio channels
2
Audio sample rate
44100 Hz
```

### Jellyfin logs

```shell
[14:49:27] [INF] [31] Emby.Server.Implementations.Session.SessionManager: Playback stopped reported by app Jellyfin Web 10.10.7 playing New Friends in a New Atelier. Stopped at 1004355 ms

[...]

PR #14092, opened 2025-05-11
Added an option to Disable "noaccurate_seek"

PeachesMLG:
**Changes**
Adds a configuration option to disable "noaccurate_seek" when transcoding ffmpeg

**Issues**
Fixes https://github.com/jellyfin/jellyfin/issues/13965


gnattu:
This is required to have a/v in sync for fmp4. If it troubles you that much you need to disable fmp4 to use MPEGTS. With mpegts server will not request non-accurate seek.

PeachesMLG:
> This is required to have a/

Well its optional? So just dont disable it surely?

PR #14100, opened 2025-05-13
Terminate at null char for audio tags

### LINKED ISSUE #13674: Null Characters Being Displayed in Web App
### Description of the bug

Important context to this issue: https://github.com/jmshrv/finamp/issues/1040

Some of my songs with a particular codec have null characters at the end of some tags. How they got there, I'm not sure, but they are appearing in the Jellyfin web app, as shown below:

They also appear in the Finamp app and appear to be affecting things such as the ability for the server to get lyrics among other things. 

I also run [Lyrion Music Server ](https://lyrion.org/) with the same music collection, which does not seem to have the same problem (at least, the null characters aren't displayed in the web interface anywhere that I can see).

### Reproduction steps

1. Have some music in your collection with a null character at the end of one of the tags (the `ARTIST`/`ALBUMARTIST` tags are what I suspect was causing my problem)
2. Refresh your Jellyfin music library.
3. Go to that album and hover of the name of the album,
4. See the null character being displayed (as shown above).

The null characters are displayed in both the web app and third-party Jellyfin clients.

Jellyfin should just ignore the null characters in the tags, or there should be an option to have them ignored. I don't see any reason that you would need the null characters displayed, but maybe there is some use case I am unaware of.

### Jellyfin Server version

10.10.0+

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.10.6

### Environment

```markdown
- OS: Ubuntu Server
- Linux Kernel: 
- Virtualization: Docker
- Clients: 
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins: 
- Reverse Proxy: 
- Base URL: 
- Networking: 
- Storage: local
```

### Jellyfin logs

I'm not sure which logs are relevant in this case.

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

**Finamp app**

image

**Jellyfin Web App** 

### Additional information

There may be more relevant information on this issue: https://github.com/jmshrv/finamp/issues/1040

I am still not quite sure on the causes of this issue, so please let me know if you need any more information.

gnattu:
Some users may use a misbehaved tag editor that writes a null char to the tag when not allowed by the tagging system. Terminates all string value from the tag at null char and ignore everything after.

Having a null char in the metadata fields not only affects value display, it will actually crash some job like nfo saving where the writer cannot handle null char gracefully.

**Changes**

**Issues**

Fixes #13674

### DIFF
--- a/src/Jellyfin.Extensions/StringExtensions.cs
+++ b/src/Jellyfin.Extensions/StringExtensions.cs
@@ -135,5 +135,24 @@ namespace Jellyfin.Extensions
         {
             return values.Select(i => (i ?? string.Empty).Trim());
         }
+
+        /// <summary>
+        /// Truncates a string at the first null character ('\0').
+        /// </summary>
+        /// <param name="text">The input string.</param>
+        /// <returns>
+        /// The substring up to (but not including) the first null character,
+        /// or the original string if no null character is present.
+        /// </returns>
+        public static string TruncateAtNull(this string text)
+        {
+            if (string.IsNullOrEmpty(text))
+            {
+                return text;
+            }
+
+            var idx = text.IndexOf('\0', StringComparison.Ordinal);
+            return idx < 0 ? text : text[..idx];
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


src/Jellyfin.Extensions/StringExtensions.cs
```diff
@@ -135,5 +135,24 @@ public static IEnumerable<string> Trimmed(this IEnumerable<string> values)
...
+        /// </returns>
+        public static string TruncateAtNull(this string text)
+        {
+            if (string.IsNullOrEmpty(text))
+            {
+                return text;
+            }
+
+            var idx = text.IndexOf('\0', StringComparison.Ordinal);
+            return idx < 0 ? text : text[..idx];
```
crobibero, line 155: Should be able to use the LeftPart extension here

Shadowghost (approved):

crobibero (approved):

PR #14131, opened 2025-05-18
Fix race condition in SessionManager.GetSessionInfo

### LINKED ISSUE #14132: Race condition in SessionManager.GetSessionInfo
### Description of the bug

# Description:
The issue happens when the user opens jellyfin, where a websocket connection is initiated and an http request is also initiated from `SessionController` and `PlaystateController`

In this case, the two threads start serving different requests, but both of them tries to get the current session info (mainly via `RequestHelpers.GetSessionId`)

The `GetSessionInfo` method name is misleading, because if the session doesn't exist, it will try to create it. This creation step is not thread safe, and causes the `_activeConnections` to overwrite one session with another. In some races, this causes the session with a websocket controller to get overwritten, causing weird bugs with SyncPlay and other functions that depend on the web sockets..

# Solution 
Using `_activeConnections.GetOrAdd` instead of `_activeConnections.TryGetValue` for an atomic operation.

### Reproduction steps

You have to trigger the race condition by opening the web client for example, on load, the client will initiate a web socket connection and any http request from the above mentioned controllers.

An example stack trace of the code path from the "PostFullCapabilities" controller:

```
Emby.Server.Implementations.Session.SessionManager.GetSessionInfo
Emby.Server.Implementations.Session.SessionManager.LogSessionActivity
Jellyfin.Api.Helpers.RequestHelpers.GetSession
Jellyfin.Api.Helpers.RequestHelpers.GetSessionId
Jellyfin.Api.Controllers.SessionController.PostFullCapabilities
```

An example stack trace of the code path from web socket controller:

```
Emby.Server.Implementations.Session.SessionManager.GetSessionInfo
Emby.Server.Implementations.Session.SessionManager.LogSessionActivity
Emby.Server.Implementations.Session.SessionManager.GetSessionByAuthenticationToken
Emby.Server.Implementations.Session.SessionWebSocketListener.GetSession
Emby.Server.Implementations.Session.SessionWebSocketListener.ProcessWebSocketConnectedAsync
Emby.Server.Implementations.HttpServer.WebSocketManager.WebSocketRequestHandler
```

The sessions created from each "GetSessionInfo" invocations can overwrite each other. 

A single session should be created, and no multi sessions that can overwrite the other shall exist.

### Jellyfin Server version

10.10.0+

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

Release

### Environment

```markdown
- OS: Fedora 42
- Linux Kernel: 6.14.6-300.fc42.x86_64
- Virtualization: None
- Clients: Browser
- Browser: 138.0.3 (64-bit)
- FFmpeg Version: ffmpeg version 7.1.1
- Hardware Acceleration: none
- GPU Model: none
- Plugins: none
- Reverse Proxy: none
- Base URL: none
- Networking: Host
- Storage:local
```

### Jellyfin logs

```shell
N/A as this is a hidden bug with no traces of failure from the system.
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

moalhaddar:
This PR addresses a thread safety issue in SessionManager.GetSessionInfo where concurrent access could lead to a race condition.

Fixes #14132

### DIFF
--- a/Emby.Server.Implementations/Session/SessionManager.cs
+++ b/Emby.Server.Implementations/Session/SessionManager.cs
@@ -508,13 +508,11 @@ namespace Emby.Server.Implementations.Session
             ArgumentException.ThrowIfNullOrEmpty(deviceId);
 
             var key = GetSessionKey(appName, deviceId);
-
-            CheckDisposed();
-
-            if (!_activeConnections.TryGetValue(key, out var sessionInfo))
+            SessionInfo newSession = CreateSessionInfo(key, appName, appVersion, deviceId, deviceName, remoteEndPoint, user);
+            SessionInfo sessionInfo = _activeConnections.GetOrAdd(key, newSession);
+            if (ReferenceEquals(newSession, sessionInfo))
             {
-                sessionInfo = CreateSession(key, appName, appVersion, deviceId, deviceName, remoteEndPoint, user);
-                _activeConnections[key] = sessionInfo;
+                OnSessionStarted(newSession);
             }
 
             sessionInfo.UserId = user?.Id ?? Guid.Empty;
@@ -538,7 +536,7 @@ namespace Emby.Server.Implementations.Session
             return sessionInfo;
         }
 
-        private SessionInfo CreateSession(
+        private SessionInfo CreateSessionInfo(
             string key,
             string appName,
             string appVersion,
@@ -582,7 +580,6 @@ namespace Emby.Server.Implementations.Session
                 sessionInfo.HasCustomDeviceName = true;
             }
 
-            OnSessionStarted(sessionInfo);
             return sessionInfo;
         }
 


JPVenson (changes requested):
Use GetOrAdd instead

https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.getoradd?view=net-9.0#system-collections-concurrent-concurrentdictionary-2-getoradd(-0-1)

moalhaddar:
@JPVenson Done.

Emby.Server.Implementations/Session/SessionManager.cs
```diff
@@ -509,13 +509,8 @@ private SessionInfo GetSessionInfo(
 
             var key = GetSessionKey(appName, deviceId);
 
-            CheckDisposed();
```
JPVenson, line 512: why remove the CheckDisposed?

moalhaddar: It was a duplicate, there are two check disposed being called back to back. The first one is on line 506.

JPVenson (approved):

cvium:
The change to GetOrAdd now causes OnSessionStarted to sometimes be called. GetOrAdd may run its delegates multiple times

moalhaddar:
After reading the documentation, you're correct. GetOrAdd will run the create session multiple times, creating extra events that might cause unknown side effects.

Is rolling back to the previous lock mechanism preferred in this case?

Emby.Server.Implementations/Session/SessionManager.cs
```diff
@@ -508,13 +510,20 @@ private SessionInfo GetSessionInfo(
...
 
             var key = GetSessionKey(appName, deviceId);
+            SessionInfo sessionInfo;
 
-            CheckDisposed();
-
-            if (!_activeConnections.TryGetValue(key, out var sessionInfo))
+            try
+            {
+                _activeConnectionsLock.EnterWriteLock();
```
JPVenson, line 517: There is no benefit of using a ReaderWriterLock when you dont also lock read operations and only use it to lock write ones

JPVenson (changes requested):

Emby.Server.Implementations/Session/SessionManager.cs
```diff
@@ -508,13 +510,20 @@ private SessionInfo GetSessionInfo(
...
+                if (!_activeConnections.TryGetValue(key, out sessionInfo))
+                {
+                    sessionInfo = CreateSession(key, appName, appVersion, deviceId, deviceName, remoteEndPoint, user);
+                    _activeConnections[key] = sessionInfo;
+                }
+            }
```
cvium, line 523: You are not using the Reader part of the lock. I would suggest this pattern instead with [Lock](https://learn.microsoft.com/en-us/dotnet/api/system.threading.lock?view=net-9.0)

```C#
private readonly Lock _activeConnectionsLock = new();
```

```C#
if (!_activeConnections.TryGetValue(key, out sessionInfo))
{
    using (_activeConnectionsLock.EnterScope())
    {
        if (!_activeConnections.TryGetValue(key, out sessionInfo))
        {
            sessionInfo = CreateSession(key, appName, appVersion, deviceId, deviceName, remoteEndPoint, user);
            _activeConnections[key] = sessionInfo;
        }
    }
}
```

moalhaddar:
@cvium @JPVenson Thank you for the comments. Indeed, the read part if unnecessary. The new code now uses "Lock" instead.

Emby.Server.Implementations/Session/SessionManager.cs
```diff
@@ -508,13 +510,15 @@ private SessionInfo GetSessionInfo(
             ArgumentException.ThrowIfNullOrEmpty(deviceId);
 
             var key = GetSessionKey(appName, deviceId);
+            SessionInfo sessionInfo;
 
-            CheckDisposed();
-
-            if (!_activeConnections.TryGetValue(key, out var sessionInfo))
+            using (_activeConnectionsLock.EnterScope())
```
cvium, line 515: I think GetSessionInfo gets called a lot. This might be a lot of unnecessary locking

moalhaddar: When i traced the method calls, the GetSessionInfo is only called in this scenarios:
- Creating a new websocket connection (ProcessWebSocketConnectedAsync in SessionWebSocketListener)
- AuthorizeQuickConnect in the QuickConnectController
- AuthenticateUserByName in the UserController

From that, i can see the GetSessionInfo does not fall into a hot path of being repeatedly called from the clients, it's only called once from the above controllers/websocket

moalhaddar: @cvium  After going through the code again, i believe you're correct that this is a hot path for some controllers. However, this lock is necessary with the current design of the session creation/retrieval.

Let me know if you have other suggestions.

gnattu:
What's my idea on fixing this:

1. Separate the `CreateSession` function with the pure function part and the side effect part. So that we can have one `CreateSessionInfo` function that only returns a new session Info, but don't start this session yet.
2. Then, we can use the `GetOrAdd` to add this created `SessionInfo` to `_activeConnections`.
3. Each thread can do a reference comparison to determine if the object currently in the dictionary is the one current thread created. If so, it means current thread win the race, and we invoke the `OnSessionStarted` callback on current thread with the `SessionInfo` created by current thread.

This approach uses slightly more memory as it may create multiple `SessionInfo` objects on multiple threads but it should perform better than a huge lock for everything.

moalhaddar:
Great idea! @gnattu 
I've went and implemented that change in the PR, I've also tested the change locally with no side effects (other than the minimum memory for the unused sessioninfos).

Emby.Server.Implementations/Session/SessionManager.cs
```diff
@@ -508,13 +508,17 @@ private SessionInfo GetSessionInfo(
...
+            if (ReferenceEquals(newSession, existingOrNewSession))
+            {
+                OnSessionStarted(newSession);
+                sessionInfo = newSession;
+            }
+            else
             {
-                sessionInfo = CreateSession(key, appName, appVersion, deviceId, deviceName, remoteEndPoint, user);
-                _activeConnections[key] = sessionInfo;
+                sessionInfo = existingOrNewSession;
```
gnattu, line 521: This does not really need an else branch. It uses existingOrNewSession in both cases.

gnattu (approved):

moalhaddar:
@JPVenson  @cvium The new solution does not introduce any locks, please let me know if you have other comments.

JPVenson (approved):

crobibero (approved):

PR #14138, opened 2025-05-19
Add authentication for 'Get subtitles in a specified format' endpoints

### LINKED ISSUE #13985: "Gets subtitles in a specified format" and "Gets subtitles in a specified format" (SubtitleController) endpoints are unauthenticated
Split from #5415 

Allows anyone who is able to retrieve the item id to download subtitles, even if not logged in
Potential fix: Authenticate both endpoints

jezeniel:
Tested locally, if not provided any token, the endpoints return `401` HTTP code. 

The `?ApiKey` query parameter is already being provided by the web client.

**Changes**
- Add `Authorize` attribute for the Get subtitles in a specified format endpoints.

**Issues**
Fixes #13985

### DIFF
--- a/Jellyfin.Api/Controllers/SubtitleController.cs
+++ b/Jellyfin.Api/Controllers/SubtitleController.cs
@@ -206,6 +206,7 @@ public class SubtitleController : BaseJellyfinApiController
     /// <response code="200">File returned.</response>
     /// <returns>A <see cref="FileContentResult"/> with the subtitle file.</returns>
     [HttpGet("Videos/{routeItemId}/{routeMediaSourceId}/Subtitles/{routeIndex}/Stream.{routeFormat}")]
+    [Authorize]
     [ProducesResponseType(StatusCodes.Status200OK)]
     [ProducesFile("text/*")]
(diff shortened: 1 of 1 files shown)


JPVenson:
i guess this will break all external players. We first have to divse a method to produce one-time tokens that clients can hand over to extenal players before we introduce auth on those endpoints

jezeniel:
@JPVenson, hello, aside from kodi are there any external players to watch out for?

nielsvanvelzen:
As mentioned before, we first need a way to create temporary scoped access tokens for playback before we can enforce authorization on all playback related endpoints that don't currently have them. Clients **will** break with changes like this.

JPVenson:
@jezeniel no worries. But this would be more or less warrent a full fleched discussion and proposal because its a non-complex topic. So i would ask you to write an implementation proposal as detailed as possible and put it up for discussion on our meta repo https://github.com/jellyfin/jellyfin-meta/discussions

Bond-009:
See comments above, it's on the TODO list.
Closing for now.

PR #14155, opened 2025-05-22
feat(syncplay): add playbackRate support

Aztup:
**Changes**
Adds backend support for `playbackRate`.


JPVenson (approved):

Aztup:
I had to make the create group request use your saved playback rate, otherwise it would feel very inconsistent. I also added sanitize to the set playback rate request to be extra safe

Apologies for the extra review 😅

Issue #8230, opened 2022-08-08
[Issue]: Folder view: Only one video in a folder results in the file having the name of the folder


Biepa:
### Please describe your bug

Hey,

I have library in which I want to display the folders as they are on my disk.
I know of the feature request for a "real" folder view.
The current implementation works quite good for me. But as an example on my disk I have:

Folder1 with Video1 and Video2
Folder2 with Video3

Then in jellyfin it is shown as:

Folder1 with Video1 and Video2
Video3 with the Name of Folder2 (without any folder)

The Folder2 is gone and only Video3 is shown, but it has the name of Folder2.

Happy to help figuring it out, when told how.

### Jellyfin Version

Other

### if other:

10.8.3

### Environment

```markdown
- OS: Ubuntu
- Virtualization: Docker
- Clients: Brwoser
- Browser: Firefox
- Storage: via mounted share
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Biepa:
Bug still exists.

SjoerdV:
maybe add an option that will **show single video files in their actual folder** in contrast to the present functionality of 'lifting up' the singular video to the parent level and letting it assume the name of the folder? 

A 'true' folder view would be nice to have IMHO

nonobio:
Hi,
I have the same issue: https://github.com/jellyfin/jellyfin/issues/9221#issuecomment-1636713205

SjoerdV:
Still an issue, please re-open

nonobio:
Hi,
Just tried with 10.8.12 server version: the issue is still here

nonobio:
Unless I'm mistaken, the problem is still present.

Biepa:
Big is still there.

felix920506:
what type of library are you using?

nonobio:
https://github.com/jellyfin/jellyfin/issues/8230#issuecomment-1636713691

Biepa:
The one I saw it for shows type "Other". If saw is the info you are looking for?

nonobio:
> [#8230 (comment)](https://github.com/jellyfin/jellyfin/issues/8230#issuecomment-1636713691)

PR #14162, opened 2025-05-23
Use filename for single videos (non-movie/null collections) in MovieResolver

theguymadmax:
**Changes**
Fixes a bug where single-video folders used the folder name as the title. Now, only movies and mixed (null) collections retain that behavior; all others use the actual filename.

**Issues**
Fixes: #8230


crobibero:
Since this is a change to the resolver I think tests should be added

theguymadmax:
The logic for movies didn’t change, it still names the video after the folder like before. The only difference is that I added a check to skip folder-based naming for home videos and music videos when there’s only a single file.  I can rewrite that condition to be more explicit:

`if (collectionType != CollectionType.homevideos && collectionType != CollectionType.musicvideos)`

Would you like me to add a test for the home videos case to confirm it works as expected, or update the code to use this clearer condition?

Shadowghost (approved):

Bond-009 (approved):

PR #14167, opened 2025-05-25
Add coderabbit config

Shadowghost:
**Changes**
* disable automatic status comment
* collapse reviews by default
* don't suggest labels and reviewsers
* don't answer in poem


JPVenson (approved):

nielsvanvelzen:
We can configure these settings on the UI so no need to add a config to the codebase for a specific tool we only use on the upstream repository at GitHub.

Issue #14177, opened 2025-05-26
Live TV stream closes prematurely for all users when one concurrent viewer stops


Fahmula:
### Description of the bug

When multiple users are concurrently watching the same Live TV channel, if one user stops their stream, the underlying M3U stream is prematurely closed for all remaining active viewers. This results in an interruption of service for users who were still actively watching.
This issue occurs regardless of how the first user's stream stops – whether they explicitly close their player/client, their client's WebSocket connection to the Jellyfin server is lost, or their specific HLS transcoding session times out due to inactivity. The core problem appears to be in how the server tracks active client sessions for a shared Live TV stream. When one client's session ends, the server incorrectly determines that no active consumers are left for the shared stream, even if other clients are still watching.

### Reproduction steps

1. Have at least two clients (Client A and Client B) start streaming the same Live TV channel.
2. Client A stops watching (e.g., closes the browser tab, stops playback in the app, or their connection is otherwise terminated).
3. Observe that after a short delay (often related to HLS segment timeouts or session cleanup), the stream also stops for Client B, even though Client B was still actively watching.
4. Logs from the M3U proxy (e.g., xTeve/Threadfin) confirm that the stream client count drops from 2 to 1 when Client A stops, and then incorrectly drops from 1 to 0 shortly thereafter, causing the stream to terminate.

### What is the current _bug_ behavior?

Live tv stream is prematurely closed for all remaining active viewers.

### What is the expected _correct_ behavior?

When Client A stops watching the shared Live TV channel, Client B (and any other concurrent viewers of the same channel) should be able to continue watching without interruption. The underlying M3U stream should only be closed by Jellyfin when the last active client for that specific channel stops watching.

### Jellyfin Server version

10.10.0+

### Specify the build version

10.10.7

### Environment

```markdown
- OS: Linux/Docker
- Clients: Jellyfin Web, Android App
- Browser: Chrome, Firefox
- FFmpeg Version: 7.0.2-Jellyfin
- Playback Method: Live TV
- Hardware Acceleration: QSV
- GPU Model: HD630
- Plugins: None
- Reverse Proxy: None
- Base URL: None
- Networking: Host
- Storage: Local
```

### Jellyfin logs

```shell
[2025-05-26 09:23:08.588 -04:00] [INF] [11] Jellyfin.LiveTv.TunerHosts.M3UTunerHost: Parsed channel: "US USA Network East (A)"
[...]

PR #13220, opened 2024-12-13
Fix consumer count off by one when closing a browser tab with a livestream that is transcoding

timminator:
This PR solves an issue where the consumer count is off by one when closing a browser tab with a livestream that is transcoding as described in Issue #13219. The task "CloseLivestream" is called twice in this case.

I resolved this issue by modifying the if-statement in the "CloseIfNeededAsync" task that determines if the "CloseLivestream" task is run or not. In my updated version it is not run if the livestream is transcoded, which resolves the aforementioned issue.
The consumer ount is then reduced by one when PlaybackStopped is reported, which is the expected behavior that is also seen by quitting the stream via the back button instead of closing the browser tab.

Issues:
Fixes #13219 
Fixes #13721 
Fixes #13684

PR #14178, opened 2025-05-26
Fix: Prevent premature Live TV stream closure for concurrent viewers

Fahmula:
When multiple clients are watching the same Live TV channel, if one client stops their stream (e.g., closes the player, their HLS session times out, or WebSocket disconnects), the underlying M3U stream was being incorrectly closed for all remaining active viewers.

This occurred because SessionManager.CloseLiveStreamIfNeededAsync would flag the stream for closure if any associated session identifier was removed, without correctly verifying if other active client sessions (and their associated identifiers) for that same liveStreamId still existed.

This change modifies SessionManager.CloseLiveStreamIfNeededAsync to only flag the underlying live stream for closure if the activeSessionMappings collection (tracking active session identifiers for that liveStreamId) becomes completely empty after removing the current session's identifiers.

This ensures that the shared Live TV stream remains active as long as at least one client session is still actively associated with it, resolving the premature closure issue for concurrent viewers.

Issue
Fixes: #14177

### DIFF
--- a/Emby.Server.Implementations/Session/SessionManager.cs
+++ b/Emby.Server.Implementations/Session/SessionManager.cs
@@ -326,33 +326,83 @@ namespace Emby.Server.Implementations.Session
         {
             bool liveStreamNeedsToBeClosed = false;
 
+            // Try to get the active session mappings for the given live stream ID.
             if (_activeLiveStreamSessions.TryGetValue(liveStreamId, out var activeSessionMappings))
             {
-                if (activeSessionMappings.TryRemove(sessionIdOrPlaySessionId, out var correspondingId))
-                {
-                    if (!string.IsNullOrEmpty(correspondingId))
-                    {
-                        activeSessionMappings.TryRemove(correspondingId, out _);
-                    }
+                // Attempt to remove the primary session/playSession ID.
+                bool mainIdRemoved = activeSessionMappings.TryRemove(sessionIdOrPlaySessionId, out var correspondingId);
 
+                if (mainIdRemoved && !string.IsNullOrEmpty(correspondingId))
+                {
+                    // If a corresponding ID was found (e.g., linking a PlaySessionId to a SessionId or vice-versa),
+                    // attempt to remove that as well to fully clear this client's association.
+                    activeSessionMappings.TryRemove(correspondingId, out _);
+                    // We don't strictly need to check the return of the second TryRemove here for the core logic,
+                    // as the goal is to ensure both are attempted if a link exists.
+                }
+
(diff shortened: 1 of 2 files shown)


timminator:
Did you recreate the bug #14177 on the current master branch or only on 10.10.7?
Because I resolved this issue in #13220 by adding the CloseLiveStreamIfNeededAsync () function in the first place.
But this is not yet implemented in 10.10.7 and will be available starting from 10.11.0.
I ran into this issue aswell in  November last year and implemented that fix.
10.10.7 has no CloseLiveStreamIfNeededAsync () and ActiveSessionMappings () yet, thats the reason your issue occured.

Emby.Server.Implementations/Session/SessionManager.cs
```diff
@@ -326,33 +326,83 @@ public async Task CloseLiveStreamIfNeededAsync(string liveStreamId, string sessi
...
+                    activeSessionMappings.TryRemove(correspondingId, out _);
+                    // We don't strictly need to check the return of the second TryRemove here for the core logic,
+                    // as the goal is to ensure both are attempted if a link exists.
+                }
 
+                // Only if we successfully removed an identifier AND the map for this live stream is now empty,
+                // should we consider closing the underlying stream.
+                if (mainIdRemoved && activeSessionMappings.IsEmpty)
+                {
                     liveStreamNeedsToBeClosed = true;
```
timminator, line 348: Theres also a misunderstanding what the previous logic did and your logic does.

LiveStreamNeedsToBeClosed does not actually mean that the stream is actually closed after this. This just determines that the CloseLivestream function is called. If you look at the code of this function:

https://github.com/jellyfin/jellyfin/blob/4ba95b9ac4d1da87b2ed81ddf46eb6be66507705/Emby.Server.Implementations/Library/MediaSourceManager.cs#L856-L879

You will see that this function only closes the livestream if the consumer count reaches zero. This means that we need to set closeLivestreamNeedsToBeClosed to true if an identifier is removed and NOT only if the map is empty aswell, because then the consumer count would never reach zero and the stream would never be closed if there is more than one person watching.

Fahmula:
> Did you recreate the bug #14177 on the current master branch or only on 10.10.7? Because I resolved this issue in #13220 by adding the CloseLiveStreamIfNeededAsync () function in the first place. But this is not yet implemented in 10.10.7 and will be available starting from 10.11.0. I ran into this issue aswell in November last year and implemented that fix. 10.10.7 has no CloseLiveStreamIfNeededAsync () and ActiveSessionMappings () yet, thats the reason your issue occured.

Just tested your fix and it worked thanks! I'll close my pr.

PR #14190, opened 2025-05-28
Fix using subnets from interfaces as lan networks

### LINKED ISSUE #14189: Can't reach jellyfin via IPv6 when restricted to local network only
### Description of the bug

I’ve set up Jellyfin in my local network but since I want to have it accessible via remote, I use a DynDNS domain. Via that, I can also reach it in my home network without the traffic having to leave the LAN because I enter the IPv6-GUA in the DynDNS.

I have separated the admin account from all other accounts, and now I want to restrict the login for the admin only to the local network, to improve security. If I do that, however, I will no longer be able to log in — all my devices are primarily using IPv6.

I’ve thought about entering the IPv6 subnet into the settings in Jellyfin, but because the prefix changes somewhat regularly by my ISP (that’s why the DynDNS), I might lose access at some point again. I could enter the whole subnet owned by my ISP, but that seems excessive and would not be that much of a security enhancement I originally wanted.

The cause is the method `InitializeLan` in the `NetworkManager` class, where all networks which are considered LAN by RFCs are added as subnets if the user has none manually configured. (RFC 1918, Loopback, IPv6-LLA and IPv6-ULA)
The debug log entry even mentions the correct behavior:
`_logger.LogDebug("Using LAN interface addresses as user provided no LAN details.");`
but the LAN interface addresses are then completely disregarded by the code.

### Reproduction steps

1. Disallow remote access for a user of your choice
2. Add a line in your `hosts` file with the IPv6-GUA of your server in the first column and then a domain name like `jellyfin-gua`
3. Open `jellyfin-gua` in your browser.
4. Try to log in as said user

The user is disallowed from logging in.

The user should be allowed to log in, because jellyfin still runs in the same IPv6-subnet (/64, probably).

### Jellyfin Server version

10.10.0+

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.10.7

### Environment

```markdown
- OS: Debian 12
- Linux Kernel: 6.1.0-37
- Virtualization: none
- Clients: Browser, Android
- Browser: Firefox 138.0.4
- Reverse Proxy: caddy
- Base URL: none
- Networking: Host
```

### Jellyfin logs

```shell
[INF] Defined LAN subnets: ["::1/128", "fe80::/10", "fc00::/7", "127.0.0.1/8", "10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16"]
[INF] Defined LAN exclusions: []
[INF] Used LAN subnets: ["::1/128", "fe80::/10", "fc00::/7", "127.0.0.1/8", "10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16"]
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

Elsensee:
**Changes**
Instead of looking at the spec-RFC LAN addresses, the LAN subnets are now derived from the unicast addresses assigned to the network interfaces. Now, all addresses are considered LAN for which a router is technically not required, so even IPv6-GUAs within the same subnet (mostly /64). As a side effect, RFC1918-subnets which are not on link, are no more included — so a host on a 192.168.x.0/24 will no longer consider 10.0.0.0/8 local, if there exists no interface with that address. (And I inverted the `if` for better code readability)
Also, I have applied the default configuration to the `detectNetworkChange` variable in case `TryParse` fails, because in the case of dynamically changing GUAs, this would lead to the wrong behavior which would only be fixable by a restart of Jellyfin.

**Issues**
Fixes #14189

### DIFF
--- a/src/Jellyfin.Networking/Manager/NetworkManager.cs
+++ b/src/Jellyfin.Networking/Manager/NetworkManager.cs
@@ -91,7 +91,10 @@ public class NetworkManager : INetworkManager, IDisposable
         _networkEventLock = new();
         _remoteAddressFilter = new List<IPNetwork>();
 
-        _ = bool.TryParse(startupConfig[DetectNetworkChangeKey], out var detectNetworkChange);
+        if (!bool.TryParse(startupConfig[DetectNetworkChangeKey], out var detectNetworkChange))
+        {
+            detectNetworkChange = true; // Set to default value
+        }
 
         UpdateSettings(_configurationManager.GetNetworkConfiguration());
 
@@ -316,32 +319,14 @@ public class NetworkManager : INetworkManager, IDisposable
             // Get configuration options
             var subnets = config.LocalNetworkSubnets;
 
-            // If no LAN addresses are specified, all private subnets and Loopback are deemed to be the LAN
-            if (!NetworkUtils.TryParseToSubnets(subnets, out var lanSubnets, false) || lanSubnets.Count == 0)
+            if (NetworkUtils.TryParseToSubnets(subnets, out var lanSubnets, false) && lanSubnets.Count != 0)
             {
-                _logger.LogDebug("Using LAN interface addresses as user provided no LAN details.");
-
-                var fallbackLanSubnets = new List<IPNetwork>();
-                if (IsIPv6Enabled)
-                {
-                    fallbackLanSubnets.Add(NetworkConstants.IPv6RFC4291Loopback); // RFC 4291 (Loopback)
-                    fallbackLanSubnets.Add(NetworkConstants.IPv6RFC4291SiteLocal); // RFC 4291 (Site local)
-                    fallbackLanSubnets.Add(NetworkConstants.IPv6RFC4193UniqueLocal); // RFC 4193 (Unique local)
-                }
-
-                if (IsIPv4Enabled)
-                {
-                    fallbackLanSubnets.Add(NetworkConstants.IPv4RFC5735Loopback); // RFC 5735 (Loopback)
-                    fallbackLanSubnets.Add(NetworkConstants.IPv4RFC1918PrivateClassA); // RFC 1918 (private Class A)
-                    fallbackLanSubnets.Add(NetworkConstants.IPv4RFC1918PrivateClassB); // RFC 1918 (private Class B)
-                    fallbackLanSubnets.Add(NetworkConstants.IPv4RFC1918PrivateClassC); // RFC 1918 (private Class C)
-                }
-
-                _lanSubnets = fallbackLanSubnets;
+                _lanSubnets = lanSubnets;
             }
             else
             {
-                _lanSubnets = lanSubnets;
+                _logger.LogDebug("Using LAN interface addresses as user provided no LAN details.");
+                _lanSubnets = _interfaces.Select(iface => iface.Subnet).ToList();
             }
 
             _excludedSubnets = NetworkUtils.TryParseToSubnets(subnets, out var excludedSubnets, true)
@@ -606,9 +591,6 @@ public class NetworkManager : INetworkManager, IDisposable
         var config = (NetworkConfiguration)configuration;
         HappyEyeballs.HttpClientExtension.UseIPv6 = config.EnableIPv6;
 
-        InitializeLan(config);
-        InitializeRemote(config);
-
         if (string.IsNullOrEmpty(MockNetworkSettings))
         {
             InitializeInterfaces();
@@ -642,6 +624,9 @@ public class NetworkManager : INetworkManager, IDisposable
 
             _interfaces = interfaces;
         }
+
+        InitializeLan(config);
+        InitializeRemote(config);
 
         EnforceBindSettings(config);
         InitializeOverrides(config);


gnattu (changes requested):
This won't work for more complex networks and only resolves the problem for your specific use case where you don't even have control of your own ipv6 prefix, and Jellyfin is primarily designed for users can at least make a static IP for its server. And no, public IPv6 addresses under the same IPv6 prefix are still publicly WAN routable IP address so your assumption is not even technically right. Having multiple subnets in a LAN is a very common use case, docker containers have its own network, for example, and you don't want to consider all containers "WAN access". For your specific use case, you should make your Jellyfin instance run within a private network instead of solely relying on ISP provided IPv6 subnets.

Elsensee:
> This won't work for more complex networks

IMHO, that's what the setting is for: complex networks — not the default.

> And no, public IPv6 addresses under the same IPv6 prefix are still publicly WAN routable IP address so your assumption is not even technically right.

They are publicly routable, but if source and destination are within the same subnet, they are not routed. That's what being in the same subnet means.

> Having multiple subnets in a LAN is a very common use case, docker containers have its own network, for example, and you don't want to consider all containers "WAN access".

Hm, that is true, I guess. The RFC1918 address ranges should all be allowed, no matter the interface addresses.

gnattu:
> Also, I have applied the default configuration to the detectNetworkChange variable in case TryParse fails, because in the case of dynamically changing GUAs, this would lead to the wrong behavior which would only be fixable by a restart of Jellyfin.

Also this part is (almost) useless. No Jellyfin's internal services detects network changes and that is kept solely for potential plugin usages and is actually subject to be removed if there is really nobody using it. Kestrel does not directly support rebind its network interface at runtime which means this detection is not useful many ways, for example, not useful for your use case if your ISP just changed your Jellyfin's IP address and your Jellyfin server is bind to that specific IP address instead of the wildcard [::]. DotNet's network change implementation on Linux is also unreliable that could even lead to main thread crash as getting interface info from the OS is not a safe operation: #8687. The said detection mechanism on Linux is just an internal timer to read the interface information and compares them, this can also cause a lot of false positives if you are running a very busy virtual bridge for containers, and the networks may be signaled as changed when it is actually not. 

Your code is only useful in the following very specific case:

- The server's network interface directly being assigned a public IP address but that address could change at any time, making static subnet setting impossible
- The server and its LAN client does not require a router to communicate with each other
- The server binds to the [::] wildcard interface instead of a specific interface so that kestrel does not have to rebind its interface
- Your OS/system setup behaves nicely with dotnet's unreliable network change detection implementation so that it won't crash your server or doing false positives.

sjorge:
> In my opinion, using dynamic IPv6 prefixes directly from your ISP is non-standard and not typical for "home users".

It's rather common setup were you track your ISP's IPv6 single delegated range when they don't delegate a prefix. One would hope ISP would just always do the later but alas.

e.g. https://docs.opnsense.org/manual/ipv6.html#track-interface

You can technically use IPv6 with local-scoped addresses, or even setup a ULA with NAT66 but the later is very not standard. 

It's quite common for all your devices to have a GUA, usually more than one :| (because some use short lived temporary ones, private one, ... depending on the SLAAC and/or DHCPv6 config)

Overall IPv6 is a big headache most of the time as not all OS'es (and you can change the behavior too) will pick different 'flavors' of there GUA depending on if it needs to each in the subnet, outside of it but within the prefix, ... 

That aside like gnattu said it's not a great idea to just implicitly trust all of them on an interface you find.

gnattu:
> Again, Jellyfin is a software targeted at home users to create their own media server. There exists a wide range of possible network configurations, dynamic IPv6-prefixes being one of them. Because of that, I would have considered Jellyfin to be able to support such cases (or at least want to support it)

Yes, Jellyfin is primarily designed for home users. However, we must acknowledge that the common technologies Jellyfin relies on are primarily intended for classic server use cases. For instance, many of those technologies lack support for dynamically changing listening endpoints at runtime, such as the DotNet’s Kestrel web server and common reverse proxies like Nginx. I understand that assigning a static public IP to home users is impractical, so we need to implement mechanisms to address this issue. One common solution is to use a spec-defined private network with a stable IP for the server, and let the router to handle the rest. This is a guaranteed to work solution and in my opinion, most practical solution to the instability of IPs and subnets. This does not require you to bind to a wildcard interface and the system IP remains unchanged for the server application.

Of course, for your specific use case, dynamically updating the trusted subnet would also work. However, as I mentioned earlier, current implementation has its own limitations. If we truly want to proceed with this route, I would prefer a more stable and predictable method for such configuration updates. The internal mechanism of .NET could cause issues as mentioned earlier, and it would fetch every interface on the system, including those not used for public internet access, such as veth for containers and bridges for VMs. Instead, I prefer to modify our network config manager so that that specific network configurations can be updated without requiring a server restart, like the trusted LAN subnets. We can then use the already-preset admin-only API to update these configurations at runtime. With this, you simply create an API key and call that API when updating the dynamic DNS entry. This approach eliminates any implicit trust in public IP ranges and avoids relying on a potentially problematic .NET mechanism. This is also useful for more deployments other than only your use case.

Elsensee:
I get that you’d prefer another approach to solve more than just this one issue. I will close this PR then. :)

PR #14207, opened 2025-06-02
Add Configuration/UpdateTrustedNetwork api to change trusted network at runtime

### LINKED ISSUE #14189: Can't reach jellyfin via IPv6 when restricted to local network only

gnattu:
Previously, updating trusted IPs (LocalNetworkSubnets, RemoteIPFilter, IsRemoteIPFilterBlacklist) required a server restart. This change introduces a new API endpoint that allows these network configuration fields to be updated at runtime, as their values are only read when needed. This improves flexibility for some special network environments.

Example:

```
curl 'http://localhost:8096/System/Configuration/UpdateTrustedNetwork' \
-X 'POST' \
-H 'Content-Type: application/json' \
-H 'Accept: */*' \
-H 'Authorization: MediaBrowser Client="Jellyfin Test", Device="Test", DeviceId="abcdefg", Version="10.11.0", Token="mytoken"' \
--data-raw '{"LocalNetworkSubnets":[],"RemoteIPFilter":[],"IsRemoteIPFilterBlacklist":false}'
```

Any of the LocalNetworkSubnets, RemoteIPFilter, IsRemoteIPFilterBlacklist can be null in the request, which means "no change".

**Changes**

**Issues**

Fixes #14189 

An alternative approach to #14190

### DIFF
--- a/MediaBrowser.Common/Net/INetworkManager.cs
+++ b/MediaBrowser.Common/Net/INetworkManager.cs
@@ -129,5 +129,11 @@ namespace MediaBrowser.Common.Net
         /// <param name="remoteIP">IP address of the client.</param>
         /// <returns><b>True</b> if it has access, otherwise <b>false</b>.</returns>
         bool HasRemoteAccess(IPAddress remoteIP);
+
+        /// <summary>
+        /// Reloads all settings and re-Initializes the instance.
+        /// </summary>
+        /// <param name="configuration">The <see cref="NetworkConfiguration"/> to use.</param>
+        public void UpdateSettings(object configuration);
     }
 }
--- a/MediaBrowser.Common/Net/INetworkManager.cs
+++ b/MediaBrowser.Common/Net/INetworkManager.cs
@@ -134,6 +134,6 @@ namespace MediaBrowser.Common.Net
         /// Reloads all settings and re-Initializes the instance.
         /// </summary>
         /// <param name="configuration">The <see cref="NetworkConfiguration"/> to use.</param>
-        public void UpdateSettings(object configuration);
+        void UpdateSettings(object configuration);
     }
 }
--- a/Jellyfin.Api/Controllers/ConfigurationController.cs
+++ b/Jellyfin.Api/Controllers/ConfigurationController.cs
@@ -1,14 +1,14 @@
 using System;
 using System.ComponentModel.DataAnnotations;
+using System.Linq;
 using System.Net.Mime;
 using System.Text.Json;
 using Jellyfin.Api.Attributes;
-using Jellyfin.Api.Constants;
 using Jellyfin.Api.Models.ConfigurationDtos;
 using Jellyfin.Extensions.Json;
 using MediaBrowser.Common.Api;
+using MediaBrowser.Common.Net;
 using MediaBrowser.Controller.Configuration;
-using MediaBrowser.Controller.MediaEncoding;
 using MediaBrowser.Model.Branding;
 using MediaBrowser.Model.Configuration;
 using Microsoft.AspNetCore.Authorization;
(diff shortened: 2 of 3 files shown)


Shadowghost (approved):

MediaBrowser.Common/Net/INetworkManager.cs
```diff
@@ -129,5 +129,11 @@ public interface INetworkManager
         /// <param name="remoteIP">IP address of the client.</param>
         /// <returns><b>True</b> if it has access, otherwise <b>false</b>.</returns>
         bool HasRemoteAccess(IPAddress remoteIP);
+
+        /// <summary>
+        /// Reloads all settings and re-Initializes the instance.
+        /// </summary>
+        /// <param name="configuration">The <see cref="NetworkConfiguration"/> to use.</param>
+        public void UpdateSettings(object configuration);
```
Bond-009:
```suggestion
        void UpdateSettings(object configuration);
```

Bond-009 (approved):

nielsvanvelzen (changes requested):
(marked as request changes so this won't accidentally be merged for 10.11)

KGT1:
shouldnt 'POST /System/Configuration/network' already support hot reload via 
https://github.com/jellyfin/jellyfin/blob/f693c9d39f00317e33153061a0c406aa8a607555/src/Jellyfin.Networking/Manager/NetworkManager.cs#L589-L596
?

nielsvanvelzen (changes requested):
We already have the UpdateNamedConfiguration operation that can be used to update network config. If special handling is required you should make a change similar to #13756.

PR #14221, opened 2025-06-04
Change GetNormalizedRemoteIP to not pull proxy IP

supitsmike:
**Changes**
Get the Header "X-Forwarded-For" and check if it has a proxy, if it does get the first IP which will be the client and return that.  If it does not have that header do original code.

### DIFF
--- a/MediaBrowser.Common/Extensions/HttpContextExtensions.cs
+++ b/MediaBrowser.Common/Extensions/HttpContextExtensions.cs
@@ -1,3 +1,4 @@
+using System.Linq;
 using System.Net;
 using Microsoft.AspNetCore.Http;
 
--- a/MediaBrowser.Common/Extensions/HttpContextExtensions.cs
+++ b/MediaBrowser.Common/Extensions/HttpContextExtensions.cs
@@ -21,12 +21,30 @@ namespace MediaBrowser.Common.Extensions
         }
 
         /// <summary>
-        /// Extracts the remote IP address of the caller of the HTTP context.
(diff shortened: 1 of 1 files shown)


supitsmike:
I recently setup Jellyfin behind a reverse proxy and then Cloudflare proxied DNS and noticed a failed IP attempt would always be the Cloudflare proxied IP. this will now check if it has the "X-Forwarded-For" header and if it does check to see if it has a proxy, if so pull the first IP in the list that will be the client's IP.

I am not sure if this is the best way to do this, but I think this should fix the issues!

crobibero:
We already check that header, you just need to add the proxy IPs to your configured known proxies.

supitsmike:
@crobibero That's impossible if you are running it through cloudflare proxied dns as the ip changes constantly. I also see no references in the code for that specific header unless I missed it.

gnattu:
> That's impossible if you are running it through cloudflare proxied dns as the ip changes constantly.

No it is possible. You just add one more reverse proxy at localhost and forward that header to the server, then you can use localhost or 127.0.0.1 as the trusted reverse proxy.

Trusting this header from arbitrary remote IP is a security flaw as any malicious client can fake this header to make the server to believe it is from an IP that it is actually not. See https://github.com/jellyfin/jellyfin/security/advisories/GHSA-qcmf-gmhm-rfv9 for more info.

> I also see no references in the code for that specific header unless I missed it.

You did miss it:
https://github.com/jellyfin/jellyfin/blob/916e897ed25f74b0112e51758e673f18ad13af1d/Jellyfin.Server/Extensions/ApiServiceCollectionExtensions.cs#L117-L139

PR #14228, opened 2025-06-06
Fix source directory for setup template

JPVenson:
R solves here.
ex. Fixes # -->


Shadowghost (approved):

gnattu:
If this is not meant to be user configurable we should handle it using the similar way we are embedding html for plugins by using `EmbeddedResource` and get the content of the html with `Assembly.GetManifestResourceStream` instead of searching a path relative to a specific dll, no?

soultaco83:
Can confirm my test server starts with this patch.

joshuaboniface (approved):

PR #14236, opened 2025-06-07
Change image language priority to prefer English over images with no language

rasko-dev:
**Changes**
Changes the ordering of remote images to prioritize English over images with no language. This was the intended behaviour prior to #8946, which changed this in order to prevent English from being equally ranked with No language when English was the requested language. However to me it looks like the change of ordering priority was due to an oversight, as the fix could easily have kept the intended priority order.

Image ordering is now:

1. Requested language
2. English 
3. No language
4. Any language

Which means that when English is the requested language, it is still prioritized higher than No language.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -272,3 +272,4 @@
  - [Martin Reuter](https://github.com/reuterma24)
  - [Michael McElroy](https://github.com/mcmcelro)
  - [Soumyadip Auddy](https://github.com/SoumyadipAuddy)
+ - [rasko](https://github.com/rasko-dev)
--- a/MediaBrowser.Model/Extensions/EnumerableExtensions.cs
+++ b/MediaBrowser.Model/Extensions/EnumerableExtensions.cs
@@ -28,9 +28,9 @@ namespace MediaBrowser.Model.Extensions
                 {
                     // Image priority ordering:
                     //  - Images that match the requested language
+                    //  - Images in English
                     //  - Images with no language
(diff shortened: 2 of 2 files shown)


gnattu:
This is still quite debatable if English posters are always better than no language one, as some users may prefer not to see a poster with a foreign language to him if possible. IMO the best way to deal with this is to not using English as a fallback but to implement the “original language” poster logic so that we don’t have to use English as the fallback.

rasko-dev:
> This is still quite debatable if English posters are always better than no language one, as some users may prefer not to see a poster with a foreign language to him if possible. IMO the best way to deal with this is to not using English as a fallback but to implement the “original language” poster logic so that we don’t have to use English as the fallback.

I agree that the ideal solution would be to (and this should apply to anything that has to do with language imo; metadata for server, audio and subtitles for clients) allow ranking in the settings, such that every user (or admin) can specify their own order of which languages in which order should be prioritized (including original and no language), however that would presumably be a large project, far beyond my skill level.

I also agree that it is debatable whether or not English should be prioritized and it's hard for me to really make any arguments other than I think it makes more sense this way, although I do see why not everyone would agree with that. Personally the current behaviour makes it so I need to have my library metadata set to English instead of my language because it fills up with posters with no language if I keep it in my own language. In my experience these posters are not very good compared to the English ones, which I would prefer instead of no language. 

This sorting behaviour would hopefully change in the future anyways to implement customizable sorting, so this would in a way be a temporary change as well.

PR #14243, opened 2025-06-08
Rework startup topic handling and reenable output to logging framework

JPVenson:
Startup log entries are now ensured to be written to logfile again


Jellyfin.Server/ServerSetupApp/StartupLogger{TCategory}.cs
crobibero, line 1: Shouldn't have brackets in the filename, use `StartupLoggerOfCategory.cs`
Then just disable the filename warning

JPVenson: lol did not see that

Shadowghost (approved):

PR #14244, opened 2025-06-08
fix(Session): don't query DB if queue hasn't changed

### LINKED ISSUE #13628: High CPU usage due to MediaSegments and TrickplayInfos SELECT DbCommands
### Description of the bug

I recently noticed extremely high CPU usage by the jellyfin process during playback (paused or resumed) of any content. It looks like the following when playback is paused:

Image

and like the following during playback (completely uses up ~200% of CPU (2 cores)):

Image

This happens regardless of the client (tried on jellyfin-media-player, jellyfin mpv shim and the jellyfin android app). This is on direct streaming, since the above pictures are the CPU usage of the jellyfin process and not any ffmpeg processes.

Since initially I figured it might be an issue with some of the plugins (namely, Playback Reporting and Kodi Sync Queue), I uninstalled both and checked but the issue still persisted.

I have already tried debugging this on the troubleshooting channel on the jellyfin server, but a solution wasn't found.

What seemed odd to me was the no. `Executed DbCommand` logs that seemed to match with the exact timing of the CPU usage spiking while paused and were constantly thrown during resumed playback. They were the following repeated over and over:
```
[2025-02-26 05:51:36.907 +05:30] [INF] [19] Microsoft.EntityFrameworkCore.Database.Command: Executed DbCommand ("0"ms) [Parameters=["@__itemId_0='?' (DbType = Guid)"], CommandType='Text', CommandTimeout='30']"
""SELECT EXISTS (
    SELECT 1
    FROM \"MediaSegments\" AS \"m\"
    WHERE \"m\".\"ItemId\" = @__itemId_0)"
[2025-02-26 05:51:36.909 +05:30] [INF] [19] Microsoft.EntityFrameworkCore.Database.Command: Executed DbCommand ("0"ms) [Parameters=["@__itemId_0='?' (DbType = Guid)"], CommandType='Text', CommandTimeout='30']"
""SELECT EXISTS (
    SELECT 1
    FROM \"MediaSegments\" AS \"m\"
    WHERE \"m\".\"ItemId\" = @__itemId_0)"
[2025-02-26 05:51:36.910 +05:30] [INF] [19] Microsoft.EntityFrameworkCore.Database.Command: Executed DbCommand ("0"ms) [Parameters=["@__itemId_0='?' (DbType = Guid)"], CommandType='Text', CommandTimeout='30']"
""SELECT \"t\".\"ItemId\", \"t\".\"Width\", \"t\".\"Bandwidth\", \"t\".\"Height\", \"t\".\"Interval\", \"t\".\"ThumbnailCount\", \"t\".\"TileHeight\", \"t\".\"TileWidth\"
FROM \"TrickplayInfos\" AS \"t\"
WHERE \"t\".\"ItemId\" = @__itemId_0"
```

These are clearly running every few milliseconds and as I mentioned, matched exactly with the CPU usage spikes. What's odd to me is that when I ran the following tasks, it took very little time to execute:

Image

Image

My Trickplay settings are set to default (shown down below in additional information).

### Reproduction steps

1. Start watching something
2. Observe CPU usage

Extremely high CPU usage for no reason (since there is no transcoding happening).

Normal CPU usage (~10-20%).

### Jellyfin Server version

10.10.0+

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.10.6

### Environment

```markdown
- OS: Void Linux
- Linux Kernel: 6.6.78_1
- Virtualization: None (bare metal)
- Clients: Jellyfin Media Player 1.11.1, Jellyfin MPV Shim 2.8.0, Latest Jellyfin Android Application (as of 27/02/2025)
- Browser: Firefox 135.0.1
- FFmpeg Version: 6.1.2
- Playback Method: Direct
- Hardware Acceleration: None
- GPU Model: None
- Plugins: None and with Playback Reporting 15 and Kodi Sync Queue 14
- Reverse Proxy: nginx
- Base URL: https://media.skaytacium.com
- Networking: Direct Ethernet connection
- Storage: Local HDDs
```

### Jellyfin logs

```shell
Posted below in additional files (I've provided debug logs, since the basic logging doesn't provide any useful information at all. I will send those too if required.)
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

[Debug logs for the server](https://github.com/user-attachments/files/18993689/log_20250226.log)

Image

Image

Image

Skaytacium:
Background and additional debugging details are given in the mentioned issue (jellyfin/jellyfin#13628), and jellyfin/jellyfin-mpv-shim#265

**Changes**
Checks if the session's `nowPlayingQueue` has changed since the last `ProgressUpdate` event, and if so, then queries the database for data transfer objects.

**Issues**
fixes #13628

**Note**
The last time I used C# was 3 years ago, so I don't know how to exactly get around the following. `SequenceEqual` uses the equality operator to check each item in an `IEnumerable<T>`, which compares the references and not the actual values. In this case, the values are more important, not the memory location. Changing `MediaBrowser.Model/Session/QueueItem.cs` to the following would be ideal:
```c#
#nullable disable
#pragma warning disable CS1591

using System;

namespace MediaBrowser.Model.Session
{
    public class QueueItem : IEquatable<QueueItem>
    {
        public Guid Id { get; set; }

        public string PlaylistItemId { get; set; }

        public bool Equals(QueueItem? other) => Id == other.Id && PlaylistItemId == other.PlaylistItemId;
    }
}
```
But the project's settings do not allow this. If there is a better way to do this, then it would be ideal, because currently the queues remain unequal even when the value is the same, due to memory location changes probably.

### DIFF
--- a/Emby.Server.Implementations/Session/SessionManager.cs
+++ b/Emby.Server.Implementations/Session/SessionManager.cs
@@ -456,7 +456,7 @@ namespace Emby.Server.Implementations.Session
 
             var nowPlayingQueue = info.NowPlayingQueue;
 
-            if (nowPlayingQueue?.Length > 0)
+            if (nowPlayingQueue?.Length > 0 && !nowPlayingQueue.SequenceEqual(session.NowPlayingQueue) && updateLastCheckInTime)
             {
                 session.NowPlayingQueue = nowPlayingQueue;
 


crobibero:
The `Equals` method can be implemented as the following, we just disallow `==` comparison for Guid as `.Equals` is more efficient 

```c#
#nullable disable
#pragma warning disable CS1591

using System;

namespace MediaBrowser.Model.Session
{
    public class QueueItem : IEquatable<QueueItem>
    {
        public Guid Id { get; set; }

        public string PlaylistItemId { get; set; }

        public bool Equals(QueueItem other) => other != null && Id.Equals(other.Id) && string.Equals(PlaylistItemId, other.PlaylistItemId, StringComparison.Ordinal);
    }
}
```

Skaytacium:
Perfect, compiles without errors and works as intended. Thanks for the tip!

JPVenson:
why not just a record?

Issue #14251, opened 2025-06-08
Jellyfin 10.11.0-rc1 upgrade hang


systemcrash:
### Description of the bug

Started the upgrade as per instructions. First attempt produced crash. Second attempt worked out (had to stop and start again). Post upgrade has a few bugs.

### Reproduction steps

1. Upgrade from 10.10.7

### What is the current _bug_ behavior?

Everything in the GUI seems to work but the storage indicators in the dashboard are 'busy'

### What is the expected _correct_ behavior?

Upgrade completes. JF goes  idle.

### Jellyfin Server version

10.11.0-rc1

### Specify version number

10.11.0-rc1

### Specify the build version

10.11.0-rc1

### Environment

```markdown
- OS:
- Linux Kernel:
- Virtualization: docker
- Clients: browser
- Browser: FF
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Jellyfin Data Storage:
- Media Storage:
- External Integrations:
```

### Jellyfin logs

relevant log section

```shell
[2025-06-08 21:40:32.249 +00:00] [ERR] [3] Jellyfin.Api.Middleware.ExceptionMiddleware: Error processing request. URL "GET" "/System/Info/Storage".
System.ArgumentNullException: Value cannot be null. (Parameter 'input')
   at System.ArgumentNullException.Throw(String paramName)
   at System.Guid.Parse(String input)
   at Emby.Server.Implementations.SystemManager.<>c.<GetSystemStorageInfo>b__9_0(VirtualFolderInfo e)
   at System.Linq.Enumerable.ListSelectIterator`2.ToArray()
   at Emby.Server.Implementations.SystemManager.GetSystemStorageInfo()
   at Jellyfin.Api.Controllers.SystemController.GetSystemStorage()
   at lambda_method6517(Closure, Object, Object[])
   at Microsoft.AspNetCore.Mvc.Infrastructure.ActionMethodExecutor.SyncObjectResultExecutor.Execute(ActionContext actionContext, IActionResultTypeMapper mapper, ObjectMethodExecutor executor, Object controller, Object[] arguments)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.InvokeActionMethodAsync()
[...]

PR #14253, opened 2025-06-08
Ignore null key virtual folders

JPVenson:
fixed: #14251

### DIFF
--- a/Emby.Server.Implementations/SystemManager.cs
+++ b/Emby.Server.Implementations/SystemManager.cs
@@ -87,7 +87,7 @@ public class SystemManager : ISystemManager
     {
         var virtualFolderInfos = _libraryManager
             .GetVirtualFolders()
-            .Where(e => !e.ItemId.IsNullOrEmpty()) // this should not be null but for some users it is.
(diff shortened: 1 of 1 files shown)


Emby.Server.Implementations/SystemManager.cs
```diff
@@ -85,7 +85,10 @@ public SystemInfo GetSystemInfo(HttpRequest request)
     /// <inheritdoc/>
     public SystemStorageInfo GetSystemStorageInfo()
     {
-        var virtualFolderInfos = _libraryManager.GetVirtualFolders().Select(e => new LibraryStorageInfo()
+        var virtualFolderInfos = _libraryManager
+            .GetVirtualFolders()
+            .Where(e => e.ItemId is not null) // this should not be null but for some users it is.
```
crobibero:
```suggestion
            .Where(e => !e.ItemId.IsNullOrEmpty()) // this should not be null but for some users it is.
```

Unless we use `Guid.Empty` for filtering but that would be weird

Shadowghost (approved):

crobibero (approved):

PR #14262, opened 2025-06-09
Feature/persistent watch data

JPVenson:
Adds placeholder base item for detach user watch data

### DIFF
--- a/Emby.Server.Implementations/Localization/Core/en-US.json
+++ b/Emby.Server.Implementations/Localization/Core/en-US.json
@@ -137,5 +137,5 @@
     "TaskMoveTrickplayImages": "Migrate Trickplay Image Location",
     "TaskMoveTrickplayImagesDescription": "Moves existing trickplay files according to the library settings.",
     "CleanupUserDataTask": "User data cleanup task",
-    "CleanupUserDataTaskDescription": "Cleans all user data (Watch state, faviorite status etc) from media that is no longer present for at least 90 days."
+    "CleanupUserDataTaskDescription": "Cleans all user data (Watch state, favorite status etc) from media that is no longer present for at least 90 days."
 }


src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/Migrations/20250609115616_DetachUserDataInsteadOfDelete.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<DateTimeOffset>(
+                name: "RetentionDate",
+                table: "UserData",
+                type: "TEXT",
+                nullable: true);
+
+            migrationBuilder.InsertData(
```
crobibero, line 20: Do we really need to specify all of these columns?

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -511,6 +524,16 @@ public void UpdateOrInsertItems(IReadOnlyList<BaseItemDto> items, CancellationTo
...
         context.SaveChanges();
 
+        foreach (var item in newItems)
+        {
+            // reattach old userData entries
+            var userKeys = item.UserDataKey.ToArray();
+            context.UserData
+                .Where(e => e.ItemId == PlaceholderId)
+                .Where(e => userKeys.Contains(e.CustomDataKey))
+                .ExecuteUpdate(e => e.SetProperty(f => f.ItemId, item.Item.Id));
```
crobibero: Should also null out RetentionDate?

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -53,6 +53,11 @@ This is your only warning/message regarding this topic.
 public sealed class BaseItemRepository
     : IItemRepository
 {
+    /// <summary>
+    /// Gets the placeholder id for UserData detached items.
+    /// </summary>
+    public static readonly Guid PlaceholderId = Guid.Parse("00000000-0000-0000-0000-000000000001");
```
crobibero, line 59:
```suggestion
    public static readonly Guid PlaceholderId = Guid.Empty;
```

Jellyfin.sln
```diff
@@ -88,6 +88,9 @@ EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Jellyfin.LiveTv", "src\Jellyfin.LiveTv\Jellyfin.LiveTv.csproj", "{8C6B2B13-58A4-4506-9DAB-1F882A093FE0}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Jellyfin.Database", "Jellyfin.Database", "{4C54CE05-69C8-48FA-8785-39F7F6DB1CAD}"
+	ProjectSection(SolutionItems) = preProject
+		src\Jellyfin.Database\readme.md = src\Jellyfin.Database\readme.md
+	EndProjectSection
```
crobibero, line 93: Was this added on purpose? We don't typically add readme to the solution

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/Migrations/20250609115616_DetachUserDataInsteadOfDelete.cs
JPVenson: its autogenerated via efcore ... i would just leave it in tbo

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
JPVenson: Nope that does not work

```
Unable to create a 'DbContext' of type 'JellyfinDbContext'. The exception 'The seed entity for entity type 'BaseItemEntity' cannot be added because a default value was provided for the required property 'Id'. Please provide a value different from '00000000-0000-0000-0000-000000000000'.' was thrown while attempting to create an instance. For the different patterns supported at design time, see https://go.microsoft.com/fwlink/?linkid=851728
```

Jellyfin.sln
JPVenson: Yes it was, otherwise its hidden in the solution and it contains important infos about how to create migrations that otherwise will get overlooked

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
crobibero: Oh duh I didn't see you used `1` as the guid
Reading is hard

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
JPVenson: eh prob a good idea

Emby.Server.Implementations/ScheduledTasks/Tasks/CleanupUserDataTask.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// <param name="logger">A logger.</param>
+    public CleanupUserDataTask(ILocalizationManager localization, IDbContextFactory<JellyfinDbContext> dbProvider, ILogger<CleanupUserDataTask> logger)
+    {
+        _localization = localization;
+        _dbProvider = dbProvider;
+        _logger = logger;
+    }
+
+    /// <inheritdoc />
+    public string Name => _localization.GetLocalizedString("CleanupUserDataTask");
```
Shadowghost, line 40: You need to add those translation keys too

Emby.Server.Implementations/Localization/Core/en-US.json
```diff
@@ -135,5 +135,7 @@
     "TaskExtractMediaSegments": "Media Segment Scan",
     "TaskExtractMediaSegmentsDescription": "Extracts or obtains media segments from MediaSegment enabled plugins.",
     "TaskMoveTrickplayImages": "Migrate Trickplay Image Location",
-    "TaskMoveTrickplayImagesDescription": "Moves existing trickplay files according to the library settings."
+    "TaskMoveTrickplayImagesDescription": "Moves existing trickplay files according to the library settings.",
+    "CleanupUserDataTask": "User data cleanup task",
+    "CleanupUserDataTaskDescription": "Cleans all user data (Watch state, faviorite status etc) from media that is no longer present for at least 90 days."
```
crobibero:
```suggestion
    "CleanupUserDataTaskDescription": "Cleans all user data (Watch state, favorite status etc) from media that is no longer present for at least 90 days."
```

Shadowghost (approved):

soleous:
I've just been exploring this on 10.11.0 RC3 and it works when removing and readding files, however on rename its not persevering watched status.  Is this a bug or normal from this change?

PR #14263, opened 2025-06-09
Fix incorrectly parsed positions in ELRC cues

### LINKED ISSUE #14254: LrcLyricParser returns mismatched cues / whitespace removal breaks cue positions
### Description of the bug

Currently, [the LrcLyricParser removes duplicate whitespace and trims the lyric lines](https://github.com/AlexDalas/jellyfin-elrc-lyrics/blob/32742fec3758c74c69518d1a34fe926ef286cedb/MediaBrowser.Providers/Lyric/LrcLyricParser.cs#L102-L103). As a result, the positions returned as part of the lyric cues in the API don't match the actual lyric line positions. This breaks support for word-synced lyrics on clients.

From how I understand the underlying issue, the (E)LRC spec isn't very standardized. On [Wikipedia](https://en.wikipedia.org/wiki/LRC_(file_format)#A2_extension_(Enhanced_LRC_format)), the tags and words in ELRC lyrics are generally surrounded by spaces. However, [the parser](https://github.com/karaoke-dev/LrcParser) Jellyfin currently uses [doesn't specifically handle whitespace](https://github.com/karaoke-dev/LrcParser/blob/2239a40366c5785c0711b9c46feeaf15414579aa/LrcParser.Tests/Parser/Lrc/Utils/LrcTimedTextUtilsTest.cs#L15), and simply sees it as part of the lyrics line.

As a result, it parses a line `[00:06.84] <00:06.84> Every <00:07.20>   <00:07.56> night <00:07.87>   <00:08.19> that <00:08.46>   <00:08.79> goes <00:09.19>   <00:09.59> between` as ` 
 Every     night     that     goes     between`. THis issue is already tracked in karaoke-dev/LrcParser#83. However, Jellyfin [tries to work around this issue by removing duplicate whitespace and trimming the string](https://github.com/AlexDalas/jellyfin-elrc-lyrics/blob/32742fec3758c74c69518d1a34fe926ef286cedb/MediaBrowser.Providers/Lyric/LrcLyricParser.cs#L102-L103), which breaks cue positions as a side effect.

### Reproduction steps

1. Add word-synced lyrics in the Enhanced LRC format to the server, which have tags and words surrounded by whitespace (see above)
2. Query those lyrics through the Jellyfin API

The API returns lyrics where the line text and cue positions don't match, specifically, the positions are usually larger than the actual index in the string.

The cue positions should be able to be used as substring indices, and correctly select the relevant segments from the lyrics. They also shouldn't go out of bounds of the returned string.

### Jellyfin Server version

Master

### Specify commit id

bdb3adeb302a6e7a3d9f3cefe85650e8ec63ad6a

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.0-rc1

### Environment

```markdown
n/a
```

### Jellyfin logs

```shell
n/a
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

Maxr1998:
As described in #14254, the positions returned with the ELRC cues are currently incorrect due to whitespace removal. This PR modifies the positions to correctly map to the lyric lines again.

**Changes**
If `TimeTags` are returned by the lrc parser, the lyric line is now assembled manually, by first extracting each substring from the original line, trimming it, and joining it with a single space. To ensure the positions are still correct, the new slice length is compared to the previous length, and added to an offset variable that impacts are consecutive slice positions. That way, all positions remain valid. Blank/empty slices (which only contain whitespace) are dropped silently, since they aren't relevant for clients.
I also updated the tests to match the newly returned data, and added checks for the positions (which weren't included before).

**Issues**
Fixes #14254.

### DIFF
--- a/MediaBrowser.Providers/Lyric/LrcLyricParser.cs
+++ b/MediaBrowser.Providers/Lyric/LrcLyricParser.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text;
 using System.Text.RegularExpressions;
 using Jellyfin.Extensions;
 using LrcParser.Model;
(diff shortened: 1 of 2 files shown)


Maxr1998:
As an aside: while working with the new API on a client, I honestly missed an end position/width from the cues. This was originally part of the PR adding ELRC support, [but removed due to a potential misunderstanding](https://github.com/jellyfin/jellyfin/pull/12941#discussion_r2134842861). Without this field, clients always have to query both the current and the next cue to extract the current slice, which creates unnecessary boilerplate. Thus, I suggest adding back a `width` or `endPosition` field to the API. Preferably, this will still happen for 10.11 since the current usability is severely impacted and adding fields shouldn't cause huge issues with the API — the original motivation presented in the ELRC PR still applies.

If desirable, I can handle this change as part of the current PR.

Maxr1998:
There still seems to be a little issue with the returned positions, I'll give it another look.

Maxr1998:
Ok, I did some more tests and modifications, and it looks like the upstream parser already returns incorrect indices for specific formats.
I tried some variants and let that parser tokenize the lyric line.

### Format currently used in Jellyfin's test, with very verbose time tags, surrounding each token

`[00:06.84] <00:06.84> Every <00:07.20>   <00:07.56> night <00:07.87>   <00:08.19> that <00:08.46>   <00:08.79> goes <00:09.19>   <00:09.59> between`
` Every|   |  night|   |  that|   |  goes|   |  between`

This variant would require trimming each token on both sides and dropping blank tokens, as described above. Seems uncommon, but should work.
The first parsed time tag has an index of `-1`, which seems like an upstream bug.

### Alternating tokens and time tags, space separated
`[00:06.84] Every <00:07.20> night <00:08.19> that <00:08.46> goes <00:09.19> between`
`Every|  night|  that|  goes|  between`

This is the most common format, and what's given as an example on Wikipedia. Parses relatively fine, but requires trimming of a duplicate space at the start. The upstream parser removes the initial space after the line tag (`[xx:xx.xx]`), however.

### Alternating tokens and time tags with space prefix 
`[00:06.84] Every<00:07.20> night<00:08.19> that<00:08.46> goes<00:09.19> between`
`Ever|y nigh|t tha|t goe|s between`

This is an obvious bug in the upstream parser. Apparently, the initial space completely confuses its position calculation.

### Alternating tokens and time tags with space suffix
`[00:06.84]Every <00:07.20>night <00:08.19>that <00:08.46>goes <00:09.19>between`
`Every |night |that |goes |between`

This gets parsed correctly, doesn't require any trimming or further modification on our end.

### Alternating tokens and time tags, no spaces 
`[00:06.84]Every<00:07.20>night<00:08.19>that<00:08.46>goes<00:09.19>between`
`Every|night|that|goes|between`

That seems to be the format primarily supported by the upstream parser, which is also used in their tests.
We need to insert spaces ourselves here, but otherwise it seems ok.

With these results, I'm not sure which formats we want to support, and how I should proceed with this PR.
It does unfortunately look like the upstream parser needs to be improved before we can fix this in Jellyfin itself.

Maxr1998:
PR has been updated to use the new library and adapt the current code to work with it. Might need a bit more testing, but is ready for an initial review.

Shadowghost (approved):

MediaBrowser.Providers/Lyric/LrcLyricParser.cs
```diff
@@ -68,45 +69,53 @@ public LrcLyricParser()
...
         {
-            var cues = new List<LyricLineCue>();
             var lyric = sortedLyricData[l];
 
-            if (lyric.TimeTags.Count != 0)
+            // Extract cues from time tags
+            var cues = new List<LyricLineCue>();
+            var keys = lyric.TimeTags.Keys.ToList();
+            int current = 0, next = 1;
+            while (next < keys.Count)
```
JPVenson: for loop instead please

Maxr1998: Fair enough. I just kept it as is, also because of the two iteration indices, but I agree that this code is quite awkward. Gonna fix that 👍

JPVenson (approved):

crobibero (approved):

Issue #14271, opened 2025-06-10
User cleanup SQL exception


enter-a-random-username:
### Description of the bug

I started it manually and it crashes instant. Log is in the log field with SQL exception.

### Reproduction steps

1. Start it - sorry not a lot more info, new  *awesome* function
2. crashed

### What is the current _bug_ behavior?

Stops instant

### What is the expected _correct_ behavior?

run

### Jellyfin Server version

Master

### Specify commit id

latest

### Specify the build version

10.11.0 (44c434556-dirty)

### Environment

```markdown
- OS:
- Linux Kernel:
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Jellyfin Data Storage:
- Media Storage:
- External Integrations:
```

### Jellyfin logs

```shell
[ERR] [52] Emby.Server.Implementations.ScheduledTasks.TaskManager: Error executing Scheduled Task
System.InvalidOperationException: The LINQ expression 'DbSet<UserData>()
    .Where(u => u.ItemId == 00000000-0000-0000-0000-000000000001)
    .Where(u => u.RetentionDate < __userDataDate_0)' could not be translated. Either rewrite the query in a form that can be translated, or switch to client evaluation explicitly by inserting a call to 'AsEnumerable', 'AsAsyncEnumerable', 'ToList', or 'ToListAsync'. See https://go.microsoft.com/fwlink/?linkid=2101038 for more information.
   at Microsoft.EntityFrameworkCore.Query.QueryableMethodTranslatingExpressionVisitor.Translate(Expression expression)
   at Microsoft.EntityFrameworkCore.Query.QueryCompilationContext.CreateQueryExecutorExpression[TResult](Expression query)
   at Microsoft.EntityFrameworkCore.Query.QueryCompilationContext.CreateQueryExecutor[TResult](Expression query)
   at Microsoft.EntityFrameworkCore.Storage.Database.CompileQuery[TResult](Expression query, Boolean async)
   at Microsoft.EntityFrameworkCore.Query.Internal.QueryCompiler.CompileQueryCore[TResult](IDatabase database, Expression query, IModel model, Boolean async)
   at Microsoft.EntityFrameworkCore.Query.Internal.QueryCompiler.<>c__DisplayClass11_0`1.<ExecuteCore>b__0()
[...]

PR #14280, opened 2025-06-11
Fix UserData cleanup task and queries

JPVenson:
fixes: #14271


src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/Migrations/JellyfinDbModelSnapshot.cs
```diff
@@ -1388,7 +1388,7 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                     b.Property<double?>("Rating")
                         .HasColumnType("REAL");
 
-                    b.Property<DateTimeOffset?>("RetentionDate")
+                    b.Property<DateTime?>("RetentionDate")
```
JPVenson, line 1391: we should not do stuff like this when we support providers that actually understand the difference but for sqlite its _fine_

nielsvanvelzen:
Please add a proper title

crobibero (approved):

PR #14281, opened 2025-06-11
Use proper scheduler that honors the parallel task limit

JPVenson:
Fixes several issues with system overloads when there are multiple folders present in a library as each would spawn its own queue limited to the task limit.

The new scheduler uses a global limit and cleansup tasks afterwards.


crobibero:
I haven't tested this but it looks like this has a cts leak (+ a few other tweaks)

```patch
Index: MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
IDEA additional info:
<+>UTF-8
diff --git a/MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs b/MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
--- a/MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs	(revision a2fb1f63d9d30e09de4244e2b6085bb6ec0e75b6)
+++ b/MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs	(date 1749650016869)
@@ -13,19 +13,16 @@
 /// <summary>
 /// Provides Parallel action interface to process tasks with a set concurrency level.
 /// </summary>
-public class LimitedConcurrencyLibraryScheduler : ILimitedConcurrencyLibraryScheduler
+public sealed class LimitedConcurrencyLibraryScheduler : ILimitedConcurrencyLibraryScheduler, IDisposable
 {
     private readonly IHostApplicationLifetime _hostApplicationLifetime;
     private readonly ILogger<LimitedConcurrencyLibraryScheduler> _logger;
     private readonly Dictionary<CancellationTokenSource, Task> _taskRunners = new();
-
-    private static AsyncLocal<CancellationTokenSource> _deadlockDetector = new();
-
-    /// <summary>
-    /// Gets used to lock all operations on the Tasks queue and creating workers.
-    /// </summary>
+    private readonly BlockingCollection<TaskQueueItem> _tasks = new();
     private readonly Lock _taskLock = new();
 
+    private static readonly AsyncLocal<CancellationTokenSource> _deadlockDetector = new();
+
     private Task? _cleanupTask;
 
     /// <summary>
@@ -39,8 +36,6 @@
         _logger = logger;
     }
 
-    private BlockingCollection<TaskQueueItem> Tasks { get; set; } = new();
-
     private void ScheduleTaskCleanup()
     {
         lock (_taskLock)
@@ -57,13 +52,13 @@
         void RunCleanupTask()
         {
             _cleanupTask = Task.Delay(TimeSpan.FromSeconds(10)).ContinueWith(
-                t =>
+                _ =>
                 {
                     lock (_taskLock)
                     {
-                        if (Tasks.Count > 0)
+                        if (_tasks.Count > 0)
                         {
-                            // tasks are still there so its still in use. Reschedule cleanup task.
+                            // tasks are still there so it is still in use. Reschedule cleanup task.
                             // we cannot just exit here and rely on the other invoker because there is a considerable timeframe where it could have already ended.
                             RunCleanupTask();
                             return;
@@ -88,8 +83,7 @@
             var parallelism = (fanoutConcurrency > 0 ? fanoutConcurrency : Environment.ProcessorCount) - _taskRunners.Count;
             for (int i = 0; i < parallelism; i++)
             {
-                var stopToken = new CancellationTokenSource();
-                var combinedSource = CancellationTokenSource.CreateLinkedTokenSource(stopToken.Token, _hostApplicationLifetime.ApplicationStopping);
+                var combinedSource = CancellationTokenSource.CreateLinkedTokenSource(new CancellationTokenSource().Token, _hostApplicationLifetime.ApplicationStopping);
                 _taskRunners.Add(
                     combinedSource,
                     Task.Factory.StartNew(
@@ -108,19 +102,20 @@
         _deadlockDetector.Value = stopToken;
         try
         {
-            foreach (var item in Tasks.GetConsumingEnumerable(stopToken.Token))
+            foreach (var item in _tasks.GetConsumingEnumerable(stopToken.Token))
             {
                 await ProcessItem(item).ConfigureAwait(false);
             }
         }
         catch (OperationCanceledException) when (stopToken.IsCancellationRequested)
         {
-            // thats how you do it, interupt the waiter thread. There is nothing to do here when it was on purpose.
+            // Task was cancelled, nothing to do.
         }
         finally
         {
-            _deadlockDetector.Value = default!;
+            _deadlockDetector.Value = null!;
             _taskRunners.Remove(stopToken);
+            stopToken.Dispose();
         }
     }
 
@@ -130,15 +125,15 @@
         {
             if (item.CancellationToken.IsCancellationRequested)
             {
-                // if item is cancled, just skip it
+                // if item is cancelled, just skip it
                 return;
             }
 
             await item.Worker(item.Data).ConfigureAwait(true);
         }
-        catch (System.Exception ex)
+        catch (Exception ex)
         {
-            _logger.LogError(ex, "Error while performing a library operation.");
+            _logger.LogError(ex, "Error while performing a library operation");
         }
         finally
         {
@@ -167,7 +162,7 @@
                     {
                         // round the percent and only update progress if it changed to prevent excessive UpdateProgress calls
                         var innerPercentRounded = Math.Round(innerPercent);
-                        if (queueItem!.ProgressValue != innerPercentRounded)
+                        if (queueItem.ProgressValue != innerPercentRounded)
                         {
                             queueItem.ProgressValue = innerPercentRounded;
                             UpdateProgress();
@@ -181,13 +176,13 @@
         for (var i = 0; i < workItems.Length; i++)
         {
             var item = workItems[i]!;
-            Tasks.Add(item, CancellationToken.None);
+            _tasks.Add(item, CancellationToken.None);
         }
 
         if (_deadlockDetector.Value is not null)
         {
-            // we are in a nested loop. There is no reason to spawn a task here as that would just lead to deadlocks and no additional concurrency is achived
-            while (workItems.Any(e => !e.Done.Task.IsCompleted) && Tasks.TryTake(out var item, 0, _deadlockDetector.Value.Token))
+            // we are in a nested loop. There is no reason to spawn a task here as that would just lead to deadlocks and no additional concurrency is achieved
+            while (workItems.Any(e => !e.Done.Task.IsCompleted) && _tasks.TryTake(out var item, 0, _deadlockDetector.Value.Token))
             {
                 await ProcessItem(item).ConfigureAwait(false);
             }
@@ -200,6 +195,13 @@
         }
     }
 
+    /// <inheritdoc />
+    public void Dispose()
+    {
+        _tasks.Dispose();
+        _cleanupTask?.Dispose();
+    }
+
     private class TaskQueueItem
     {
         public required object Data { get; init; }
```

MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
```diff
@@ -0,0 +1,217 @@
...
+            {
+                var stopToken = new CancellationTokenSource();
+                var combinedSource = CancellationTokenSource.CreateLinkedTokenSource(stopToken.Token, _hostApplicationLifetime.ApplicationStopping);
+                _taskRunners.Add(
+                    combinedSource,
+                    Task.Factory.StartNew(
+                        ItemWorker,
+                        combinedSource,
+                        combinedSource.Token,
+                        TaskCreationOptions.LongRunning | TaskCreationOptions.PreferFairness,
```
cvium: Are you sure they are long running? The "thread" that you are spawning will be released at the first `await` and  the work will be continued on the threadpool.

MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
```diff
@@ -0,0 +1,217 @@
...
+    private void Worker()
+    {
+        lock (_taskLock)
+        {
+            var fanoutConcurrency = BaseItem.ConfigurationManager.Configuration.LibraryScanFanoutConcurrency;
+            var parallelism = (fanoutConcurrency > 0 ? fanoutConcurrency : Environment.ProcessorCount) - _taskRunners.Count;
+            for (int i = 0; i < parallelism; i++)
+            {
+                var stopToken = new CancellationTokenSource();
+                var combinedSource = CancellationTokenSource.CreateLinkedTokenSource(stopToken.Token, _hostApplicationLifetime.ApplicationStopping);
```
cvium, line 109: As Cody mentioned, the CTS are leaking here. You must always dispose them when finished.

MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
```diff
@@ -0,0 +1,217 @@
...
+
+            RunCleanupTask();
+        }
+
+        void RunCleanupTask()
+        {
+            _cleanupTask = Task.Delay(TimeSpan.FromSeconds(10)).ContinueWith(
+                t =>
+                {
+                    lock (_taskLock)
```
cvium: Lock does not support tasks. If the continuation of `Task.Delay` is run on a separate thread from the one holding the lock, this won't work at all.

MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
```diff
@@ -0,0 +1,217 @@
...
+    private readonly IHostApplicationLifetime _hostApplicationLifetime;
+    private readonly ILogger<LimitedConcurrencyLibraryScheduler> _logger;
+    private readonly Dictionary<CancellationTokenSource, Task> _taskRunners = new();
+
+    private static AsyncLocal<CancellationTokenSource> _deadlockDetector = new();
+
+    /// <summary>
+    /// Gets used to lock all operations on the Tasks queue and creating workers.
+    /// </summary>
+    private readonly Lock _taskLock = new();
```
cvium, line 30: Lock is not for async workloads. Why not a semaphoreslim?

MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
```diff
@@ -0,0 +1,217 @@
...
+            var item = workItems[i]!;
+            Tasks.Add(item, CancellationToken.None);
+        }
+
+        if (_deadlockDetector.Value is not null)
+        {
+            // we are in a nested loop. There is no reason to spawn a task here as that would just lead to deadlocks and no additional concurrency is achived
+            while (workItems.Any(e => !e.Done.Task.IsCompleted) && Tasks.TryTake(out var item, 0, _deadlockDetector.Value.Token))
+            {
+                await ProcessItem(item).ConfigureAwait(false);
```
cvium: Why is Enqueue processing items directly? If there is a deadlock, shouldn't it just return and let the caller continue?

MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
```diff
@@ -0,0 +1,217 @@
...
+                // cleanup task is already running.
+                return;
+            }
+
+            RunCleanupTask();
+        }
+
+        void RunCleanupTask()
+        {
+            _cleanupTask = Task.Delay(TimeSpan.FromSeconds(10)).ContinueWith(
```
cvium: ContinueWith is from before async-await. If you want to create a fire-and-forget task, just don't await it (if you want error handling etc. it becomes slightly more involved but not much), e.g.:

```C#
_ = RunCleanupTask();

**Review by cvium — DISMISSED** [review]

File: MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs (hunk 3 above)
Comment (JPVenson) [reply MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs]: ofc it does lock internally ... this is exactly as it supposed to be

File: MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs (hunk 4 above)
Comment (JPVenson) [reply MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs:30]: this lock is just for accessing the task list nothing more. There is no async locking taking place

File: MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs (hunk 5 above)
Comment (JPVenson) [reply MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs]: continue with what?
The contract of that method is that the list of items get processed and it waits until all tasks are processed

File: MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs (hunk 3 above)
Comment (cvium) [reply MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs]: My bad, you are right, the lock is released before the continuation occurs, but it does introduce the potential for developer error down the line.

File: MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs (hunk 1 above)
Comment (JPVenson) [reply MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs]: yea they can run pretty long with extraction tasks in them, hogging the TaskScheduler for this does not seem like a good idea

File: MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs (hunk 6 above)
Comment (JPVenson) [reply MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs]: please elaborate, i dont await anything here and i dont want the caller to wait for the resolve of the task

File: MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs (hunk 5 above)
Comment (cvium) [reply MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs]: I think I get it now, but I am wondering if it is not possible to just add the workItems to Tasks and let the outer loop handle it?

**Comment by JPVenson** [comment]
yea good catch on the CTS, totally overlooked that one

File: MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs (hunk 1 above)
Comment (cvium) [reply MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs]: But any time you use `await`, it will use the thread pool for that work

File: MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs (hunk 3 above)
Comment (JPVenson) [reply MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs]: you are not _wrong_ with that ... but i cannot come up with an easy and simple solution for that right away. Ideas?

File: MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs (hunk 5 above)
Comment (JPVenson) [reply MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs]: no because if you have for example the limit set to 1, the outer task already awaits this current method therefor this exactly would produce a deadlock where the outer task waits on the inner task and the inner tasks waits for its sub tasks to be scheduled which will never happen because its already processing the outer task.

There is a pattern i would rather do but thats not possible without massive refactoring in the caller

File: MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs (hunk 1 above)
Comment (JPVenson) [reply MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs]: i dont think it should in this case as the awaiting task should inherit the task options of the caller. Though thinking about it for longer i dont think thats very nessesary in this case.

Just the fact i dont take up space in the ThreadPool for this might be an upside. The more i think about it, the less i think it might be beneficial ... but i dont think it hurts much either.

File: MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs [hunk 7]
```diff
@@ -0,0 +1,232 @@
...
+    /// <inheritdoc/>
+    public void Dispose()
+    {
+        if (_disposed)
+        {
+            return;
+        }
+
+        _disposed = true;
+        _tasks.Dispose();
```
crobibero, line 294:
```suggestion
        _tasks.Dispose();
        _cleanupTask?.Dispose();
```

MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
cvium: That is how async-await works though. Unless you use a custom scheduler, it will schedule the work on the threadpool. You are therefore wasting a lot (relatively) of resources in creating a new thread and then immediately releasing it.

image

JPVenson: yea try with ConfigureAwait(true) as i did

cvium: What do you mean? `true` is the default value and `false` tells it to allow continuations on a different thread. **But** we are not using a custom SynchronizationContext (ASP.NET Core doesn't have one) or custom task scheduler, so `ConfigureAwait` is a no-op anyway.

MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
cvium: This behaves identically, but uses async-await

```csharp
    private void ScheduleTaskCleanup()
    {
        lock (_taskLock)
        {
            if (_cleanupTask is not null)
            {
                // cleanup task is already running.
                return;
            }

            _cleanupTask = RunCleanupTask();
        }
    }

    private async Task RunCleanupTask()
    {
        await Task.Delay(TimeSpan.FromSeconds(10));
        
        // do stuff
    }

File: MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs [hunk 8]
```diff
@@ -0,0 +1,277 @@
...
+        if (BaseItem.ConfigurationManager.Configuration.LibraryScanFanoutConcurrency == 1)
+        {
+            try
+            {
+                foreach (var item in workItems)
+                {
+                    await ProcessItem(item).ConfigureAwait(false);
+                }
+            }
+            catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
```
cvium, line 236: Will this ever happen? The cancellationtoken is, as far as I can see, not used downstream.

JPVenson: Yes but it is used upstream and the caller is generally responsible for canceling the operation anyway even in all other cases

cvium: My point is that `cancellationToken` will never be responsible for the cancellation if it is not used downstream.

JPVenson: The scheduler does not cancel it so you are right in assuming that however the token is handed to the method from the caller and the caller is responsible for using it downstream in the worker function not us.

MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
```diff
@@ -60,7 +60,7 @@ private void ScheduleTaskCleanup()
         async Task RunCleanupTask()
         {
             _logger.LogDebug("Schedule cleanup task in 10 sec.");
```
gnattu:
```suggestion
            _logger.LogDebug("Schedule cleanup task in 120 sec.");
```

MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
```diff
@@ -0,0 +1,310 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace MediaBrowser.Controller.LibraryTaskScheduler;
+
+/// <summary>
+/// Provides Parallel action interface to process tasks with a set concurrency level.
+/// </summary>
+public sealed class LimitedConcurrencyLibraryScheduler : ILimitedConcurrencyLibraryScheduler, IAsyncDisposable
+{
+    private const int _cleanupGracePerioid = 60;
```
crobibero:
```suggestion
    private const int CleanupGracePeriod = 60;
```

wrong style and typo

MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
```diff
@@ -0,0 +1,310 @@
...
+        }
+    }
+
+    private async Task ProcessItem(TaskQueueItem item)
+    {
+        try
+        {
+            if (item.CancellationToken.IsCancellationRequested)
+            {
+                // if item is cancled, just skip it
```
crobibero:
```suggestion
                // if item is cancelled, just skip it
```

MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
```diff
@@ -0,0 +1,310 @@
...
+            try
+            {
+                foreach (var item in workItems)
+                {
+                    await ProcessItem(item).ConfigureAwait(false);
+                }
+            }
+            catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
+            {
+                // operation is cancled. Do nothing.
```
crobibero:
```suggestion
                // operation is cancelled. Do nothing.
```

MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
```diff
@@ -0,0 +1,310 @@
...
+            {
+                // we are in a nested loop. There is no reason to spawn a task here as that would just lead to deadlocks and no additional concurrency is achieved
+                while (workItems.Any(e => !e.Done.Task.IsCompleted) && _tasks.TryTake(out var item, 200, _deadlockDetector.Value.Token))
+                {
+                    await ProcessItem(item).ConfigureAwait(false);
+                }
+            }
+            catch (OperationCanceledException) when (_deadlockDetector.Value.IsCancellationRequested)
+            {
+                // operation is cancled. Do nothing.
```
crobibero:
```suggestion
                // operation is cancelled. Do nothing.
```

crobibero (commented):
Just a spelling pass, testing is later

MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
cvium: I see that the current callers of `RunTasks` capture the CT in the delegates, but it should be an argument to the delegate instead and be called like `await worker(item.Data, item.CancellationToken)`. That gives more control over where the tasks can be cancelled from.

As this is old, I won't require the change, but I would really recommend it

JPVenson: yea i know and i dont like it either but i dont want to modify any more code then this for fixing the actual bug

crobibero:
Seems to be working fine for me, but we shouldn't add more usages to the BaseItem.ConfigurationManager if we don't absolutely have to


```patch
Index: MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
IDEA additional info:
<+>UTF-8
diff --git a/MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs b/MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
--- a/MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs	(revision 3ea69ff0f219af32f41c507710679f38bac4dd1c)
+++ b/MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs	(date 1749917894574)
@@ -5,7 +5,7 @@
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
-using MediaBrowser.Controller.Entities;
+using MediaBrowser.Controller.Configuration;
 using Microsoft.Extensions.Hosting;
 
@@ -19,6 +19,7 @@
     private const int CleanupGracePeriod = 60;
     private readonly IHostApplicationLifetime _hostApplicationLifetime;
     private readonly ILogger<LimitedConcurrencyLibraryScheduler> _logger;
+    private readonly IServerConfigurationManager _serverConfigurationManager;
     private readonly Dictionary<CancellationTokenSource, Task> _taskRunners = new();
 
     private static readonly AsyncLocal<CancellationTokenSource> _deadlockDetector = new();
@@ -39,10 +40,15 @@
     /// </summary>
     /// <param name="hostApplicationLifetime">The hosting lifetime.</param>
     /// <param name="logger">The logger.</param>
-    public LimitedConcurrencyLibraryScheduler(IHostApplicationLifetime hostApplicationLifetime, ILogger<LimitedConcurrencyLibraryScheduler> logger)
+    /// <param name="serverConfigurationManager">The server configuration manager.</param>
+    public LimitedConcurrencyLibraryScheduler(
+        IHostApplicationLifetime hostApplicationLifetime,
+        ILogger<LimitedConcurrencyLibraryScheduler> logger,
+        IServerConfigurationManager serverConfigurationManager)
     {
         _hostApplicationLifetime = hostApplicationLifetime;
         _logger = logger;
+        _serverConfigurationManager = serverConfigurationManager;
     }
 
     private void ScheduleTaskCleanup()
@@ -94,7 +100,7 @@
     {
         lock (_taskLock)
         {
-            var fanoutConcurrency = BaseItem.ConfigurationManager.Configuration.LibraryScanFanoutConcurrency;
+            var fanoutConcurrency = _serverConfigurationManager.Configuration.LibraryScanFanoutConcurrency;
             var parallelism = (fanoutConcurrency > 0 ? fanoutConcurrency : Environment.ProcessorCount) - _taskRunners.Count;
             _logger.LogDebug("Spawn {NumberRunners} new runners.", parallelism);
             for (int i = 0; i < parallelism; i++)
@@ -217,7 +223,7 @@
             };
         }).ToArray();
 
-        if (BaseItem.ConfigurationManager.Configuration.LibraryScanFanoutConcurrency == 1)
+        if (_serverConfigurationManager.Configuration.LibraryScanFanoutConcurrency == 1)
         {
             _logger.LogDebug("Process sequentially.");
             try
```

crobibero:
An alternate approach might be to use Channels to manage the queue and just have the concurrent number of workers as sleeping threads

JPVenson:
> An alternate approach might be to use Channels to manage the queue and just have the concurrent number of workers as sleeping threads

i thought about that too, even with the current code that would made things __waaaay__ easier because i dont need to cleanup everything after running but it also hogs resources and i dont see why we should reserve those threads for something that only runs once a day.
There are many better patterns, like having scoped scheduler runners for operations or more interactive sharing of resources which i all want to utilise but they require much more effort and intigration into the code and i am not willing to touch more code in the RC then absolutly nessesary.

Channels are also a very nice pattern but IMO they are more suited for static usage. I think a channel is overkill here for this internal usage.

crobibero (approved):

Shadowghost (approved):
Depending on user testing we might need to change the default concurrency

crobibero:
This can be improved for future rc/release, but merging for now

PR #14296, opened 2025-06-13
Enhance and fix live tv recording file names and retry logic

theshoeshiner:
### Changes

This PR fixes (and enhances) the livetv recording logic. The high level points are:

- Adjusts retry logic to lose less content while still preventing swamping the server
- Changes the file names to use a more sane pattern
- Adds a concatenation process at the end of the recording to ensure the result is a single file

For a more detailed summary and justification...
(Apologies for the length, first PR here, trying to be thorough, perhaps to a fault)

### Detailed summary of changes
1) Keeps track of all recording parts for given timer in `TimerInfo.RecordingPartPaths` and file currently being written in `TimerInfo.CurrentRecordingPath`.
1) Tracks the total retry count in `TimerInfo.RetryCount` for posterity, and the _true_ failed retries (retries that produced no data) in `TimerInfo.FailedRetryCount`.
1) Scales the retry delay up to 60s based on the number of _true_ failures so that a single short failure will only lose a minimal amount of stream content.
1) Fails the recording only when the `TimerInfo.FailedRetryCount` reaches 20 i.e. about ~15m of total retrying _with no data_.
1) Fixes issue with how recording retry files are named:
   - Current logic creates files like this:
      - [Prefix] - 1.ts
      - [Prefix] - 1 - 2.ts
      - [Prefix] - 1 - 2 - 3.ts
   - New logic will create:
      - [Prefix]_0.ts
      - [Prefix]_1.ts
      - [Prefix]_2.ts
   - Final file will be named: [Prefix].ts
1) Runs a simple FFmpeg concatenation process on the list of files in `TimerInfo.RecordingPartPaths` when the recording is either complete or failed and has multiple files listed in `TimerInfo.RecordingPartPaths`.
   - Makes use of some existing logic in `MediaEncoder.GenerateConcatConfig()` to generate the .concat config file, and adds `MediaEncoder.ConcatenateMedia()` to handle execution.
   - Removes the intermediate recording files *only* if concatenation is successful.

### Justification

1) Regarding concatenation - Given that retry logic already exists, it's clear the aim is to record the entire time slot, regardless of failures. With this in mind, it seems reasonable to have the recording produce a single file. Discontinuities will be present, but they exist regardless, and the separate files will remain if for some reason it cannot concatenate. Currently, the only way to do this would be to write a somewhat complex, implementation dependent, post-processing script (which means for the average user it is essentially impossible). I think that *maybe* this is a candidate for being configurable, but I have struggled to come up with a good reason for preferring multiple files (are users actually going to replace the gaps with the real content in the future??). If it were to be configurable then I would think concatenation should at least default to true, given that retrying is also the default. Note: While it is theoretically possible to setup an FFmpeg process that appends to a current file, it involves (per my understanding) stdout redirection, and overall seems more brittle than post-stream concatenation, which is the more standard method (Unless the custom jellyfin FFmpeg can handle appending in ways that core cannot).
1) Regarding the retry count and timer - I have personally encountered IPTV streams that generate intermittent FFmpeg failures every 5-10 minutes, but are otherwise high quality and stable. With the current logic, the recording would lose 1m each failure, max out at ~60m total recording time, and produce 10 different files. Obviously less than ideal. It's a much better experience to continue trying to record as long as the stream is returning data, and attempt to retain as much data as possible, while also avoiding swamping the server with useless retries, by starting with a short delay and scaling it up.
1) Regarding the file naming - I consider that to be a likely bug. I cannot fathom a reason that one would want files named like they are currently.

### Backwards Compatibility

I do not believe this will introduce any breaking changes to current features.
- No properties are removed from `TimerInfo`.
- No existing functionality is altered in `MediaEncoder`.
- During recording, files are kept separate so that users can still preview them as they can now.
- The only things that *might* break would be user post-processing scripts, if they are relying on the implementation details of the file naming. But honestly that functionality is already sort of broken since the script only gets a reference to the *latest* file. This enhancement would actually fix that issue since the script would now receive the path to the final merged file.

### Scenarios Tested

All of these scenarios worked as expected, producing a single file for the recording that contained all the content captured when the recording was in progress.
1) Streams with 0 up to ~30 failures / files to concatenate. The enhancement imposes no real limit (see concern # 1 below), but I only tested up to ~30)
1) Restarting the server mid-recording - Recording resumed automatically after startup and concatenated new chunk to existing chunk.
1) Manually stopping the recording and restarting - Recording resumed and concatenated new chunk to existing chunk.

I could maybe see an argument for not concatenating when the user manually stops/starts the recording, if we take that to be a sign that they *don't* want a single recording? My gut still tells me that a single time slot =  a single file, but could go either way.

### Concerns

1) I would be concerned about a theoretical stream (not sure if its even possible) that continuously produced a miniscule amount of data for FFmpeg to write to a file, and then failed. This would cause our logic to assume that the stream is "valid" and continue to quickly retry, until the timer expired. This shouldn't overload the server, as each prior process would be terminated, but could produce lots of files. If this is a valid possibility then I could introduce a cap on the number of files for a single recording. I would think something high like 1000 should reasonably cover most scenarios, and this could also be made configurable if necessary.
2) I have also considered that it may be nice to notify the user that "Hey, this recording was not perfect". Currently that is apparent by the presence of multiple (ridiculously named :-)) files. But if this change were implemented the user would not have much of an idea after the recording, aside from looking at the duration - which is not all that bad. Notifying would be a cooler UX, but live recording already has its imperfections, and there will always be tons of scenarios where the recordings are imperfect with no way whatsoever to notify the user. So I think any improvements to that would be reasonable to handle in follow up work.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -198,6 +198,7 @@
  - [revam](https://github.com/revam)
  - [allesmi](https://github.com/allesmi)
  - [ThunderClapLP](https://github.com/ThunderClapLP)
+ - [theshoeshiner](https://github.com/theshoeshiner)
 
 # Emby Contributors
 
--- a/MediaBrowser.Controller/LiveTv/TimerInfo.cs
+++ b/MediaBrowser.Controller/LiveTv/TimerInfo.cs
@@ -19,6 +19,7 @@ namespace MediaBrowser.Controller.LiveTv
             ProviderIds = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
             SeriesProviderIds = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
             Tags = Array.Empty<string>();
+            RecordingPartPaths = Array.Empty<string>();
         }
 
         public Dictionary<string, string> ProviderIds { get; set; }
@@ -112,6 +113,8 @@ namespace MediaBrowser.Controller.LiveTv
 
         public int RetryCount { get; set; }
 
+        public int FailedRetryCount { get; set; }
+
         // Program properties
         public int? SeasonNumber { get; set; }
 
@@ -160,6 +163,10 @@ namespace MediaBrowser.Controller.LiveTv
         public string[] Genres { get; set; }
 
         public string RecordingPath { get; set; }
+
+        public string CurrentRecordingPath { get; set; }
+
+        public string[] RecordingPartPaths { get; set; }
 
         public KeepUntil KeepUntil { get; set; }
     }
--- a/MediaBrowser.Controller/MediaEncoding/IMediaEncoder.cs
+++ b/MediaBrowser.Controller/MediaEncoding/IMediaEncoder.cs
@@ -7,7 +7,6 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Threading;
 using System.Threading.Tasks;
-using MediaBrowser.Model.Configuration;
 using MediaBrowser.Model.Dlna;
 using MediaBrowser.Model.Drawing;
 using MediaBrowser.Model.Dto;
@@ -281,5 +280,14 @@ namespace MediaBrowser.Controller.MediaEncoding
         /// <param name="source">The <see cref="MediaSourceInfo"/>.</param>
         /// <param name="concatFilePath">The path the config should be written to.</param>
         void GenerateConcatConfig(MediaSourceInfo source, string concatFilePath);
+
+        /// <summary>
+        /// Runs a simple FFmpeg concatenation job on a list of files and writes them to the output file.
+        /// </summary>
+        /// <param name="filePaths">The file paths to concatenate.</param>
+        /// <param name="outputFile">The output file.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <returns>Task.</returns>
+        Task ConcatenateMedia(IReadOnlyCollection<string> filePaths, string outputFile, CancellationToken cancellationToken);
     }
 }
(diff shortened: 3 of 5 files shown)


jhparker88:
> Runs a simple FFmpeg concatenation process on the list of files in TimerInfo.RecordingPartPaths when the recording is either complete or failed and has multiple files listed in TimerInfo.RecordingPartPaths

I'm looking for same clarification about the concatenation.

Currently, when recording a program (Day 1) the file is named `recording_example.ts`, several days later the same program (Day 5) is recorded (the first recording had some issues etc.) which would use the same folder and name plus `- 1` such as `recording_example - 1.ts`  and on Day 11 same program is recorded as `recording_example - 2.ts` . Given that `recording_example.ts`, `recording_example - 1.ts`, and `recording_example - 2.ts` are completely different entities for the same program, I assume this PR doesn't try to merge those files into a single entity names `recording_example.ts`.

theshoeshiner:
@jhparker88 

> Given that recording_example.ts, recording_example - 1.ts, and recording_example - 2.ts are completely different entities for the same program, I assume this PR doesn't try to merge those files into a single entity names recording_example.ts.

No. The logic doesn't attempt to search for files by a pattern. It only concatenates "part" files that have been directly created for the given timer, and have been added into `TimerInfo.RecordingPartPaths`.

When a recording for a new time slot for the same program is kicked off, the prefix for the recordings will include the start date/time of the time slot. For two recordings of the same show, this PR will leave files like this in the recording folder for that show:

 `[name][yyyy_mm_dd_hh_mm_ss].ts`

It drops the "- #' suffix, as that was only necessary to prevent collisions between error-retries, not between separate time slots.

_During_ recording, there may be files like `[name][yyyy_mm_dd_hh_mm_ss]_[part#].ts`. These are the "parts", which may be caused by server restarts, or stream failures. The unique part names are saved explicitly to `TimerInfo.RecordingPartPaths`. So that if a file with a clashing name already existed it will not get caught up in the recording/concatenation process. The logic will simply jump to the next index, and the clashing file will remain in the directory.

PR #14304, opened 2025-06-13
Always set update action when item does not exist

JPVenson:


Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -2353,4 +2354,14 @@ private IQueryable<BaseItemEntity> TranslateQuery(
...
         return baseQuery;
     }
+
+    /// <inheritdoc/>
+    public async Task<bool> ItemExistsAsync(Guid id)
+    {
+        var dbContext = await _dbProvider.CreateDbContextAsync().ConfigureAwait(false);
+        await using (dbContext.ConfigureAwait(false))
+        {
+            return dbContext.BaseItems.Any(f => f.Id == id);
```
crobibero:
```suggestion
            return await dbContext.BaseItems.AnyAsync(f => f.Id == id).ConfigureAwait(false);
```

Since the function is already async

MediaBrowser.Providers/Manager/MetadataService.cs
```diff
@@ -284,12 +291,20 @@ protected virtual Task AfterMetadataRefresh(TItemType item, MetadataRefreshOptio
...
         {
             var updateType = BeforeSaveInternal(item, isFullRefresh, currentUpdateType);
 
             updateType |= item.OnMetadataChanged();
 
+            if (updateType == ItemUpdateType.None)
+            {
+                if (!await ItemRepository.ItemExistsAsync(item.Id).ConfigureAwait(false))
+                {
+                    return ItemUpdateType.MetadataEdit;
```
crobibero:
```suggestion
                    return ItemUpdateType.MetadataImport;
```
Mentioned in chat, I also think this makes more sense

crobibero (approved):

PR #14306, opened 2025-06-15
Add progress reporting to AudioNormalizationTask.

shanepowell:
Add missing progress reporting to AudioNormalizationTask

### DIFF
--- a/Emby.Server.Implementations/ScheduledTasks/Tasks/AudioNormalizationTask.cs
+++ b/Emby.Server.Implementations/ScheduledTasks/Tasks/AudioNormalizationTask.cs
@@ -87,6 +87,8 @@ public partial class AudioNormalizationTask : IScheduledTask
             double nextPercent = numComplete + 1;
             nextPercent /= libraries.Length;
             nextPercent -= percent;
+            // Split the progress for this single library into two halves: album gain and track gain.
+            // The first half will be for album gain, the second half for track gain.
             nextPercent /= 2;
             var albumComplete = 0;
 
@@ -120,7 +122,7 @@ public partial class AudioNormalizationTask : IScheduledTask
                     }
                 }
 
-                // Update progress
+                // Update sub-progress for album gain
                 albumComplete++;
                 double albumPercent = albumComplete;
                 albumPercent /= albums.Count;
@@ -128,6 +130,7 @@ public partial class AudioNormalizationTask : IScheduledTask
                 progress.Report(100 * (percent + (albumPercent * nextPercent)));
             }
 
+            // Update progress to start at the track gain percent calculation
             percent += nextPercent;
 
             _itemRepository.SaveItems(albums, cancellationToken);
@@ -146,7 +149,7 @@ public partial class AudioNormalizationTask : IScheduledTask
                         cancellationToken).ConfigureAwait(false);
                 }
 
-                // Update progress
+                // Update sub-progress for track gain
                 tracksComplete++;
                 double trackPercent = tracksComplete;
                 trackPercent /= tracks.Count;


Emby.Server.Implementations/ScheduledTasks/Tasks/AudioNormalizationTask.cs
```diff
@@ -76,6 +76,10 @@ public AudioNormalizationTask(
     /// <inheritdoc />
     public async Task ExecuteAsync(IProgress<double> progress, CancellationToken cancellationToken)
     {
+        var numComplete = 0;
+        var numlibraries = _libraryManager.RootFolder.Children.Count();
```
gnattu: This won't work if not all libraries are Music libraries and/or not all music libraries have enabled LUFS scan. You have to either always report the progress later(in line 88) when we breaks the loop early, or only count the libraries that have enabled library scan.

shanepowell:
updated to always report progress

gnattu:
You are not resolving what I'm taking about. `_libraryManager.RootFolder.Children.Count();` will return the library count for all kinds of libraries, not only the ones that the LUFS scan should be run on. Your global percentage is still calculated based on this and not properly handles the libraries that the LUFS scan will not run on.

shanepowell:
So you want me to change it to not include the EnableLUFSScan libraries?

shanepowell:
Changed as I think you requested.

Emby.Server.Implementations/ScheduledTasks/Tasks/AudioNormalizationTask.cs
```diff
@@ -76,81 +76,94 @@ public AudioNormalizationTask(
...
+            var albumComplete = 0;
 
             foreach (var a in albums)
             {
-                if (a.NormalizationGain.HasValue || a.LUFS.HasValue)
-                {
-                    continue;
-                }
-
                 // Skip albums that don't have multiple tracks, album gain is useless here
```
gnattu: This comment is now in a wrong place. It should now be put on the `if (albumTracks.Count > 1)` line.

Emby.Server.Implementations/ScheduledTasks/Tasks/AudioNormalizationTask.cs
```diff
@@ -76,81 +76,95 @@ public AudioNormalizationTask(
...
-            var albums = _libraryManager.GetItemList(new InternalItemsQuery
-            {
-                IncludeItemTypes = [BaseItemKind.MusicAlbum],
-                Parent = library,
-                Recursive = true
-            });
+            double nextPercent = numComplete + 1;
+            nextPercent /= libraries.Length;
+            nextPercent -= percent;
+            nextPercent /= 2;
```
IDisposable, line 92: Would be nice to have a comment here about why we're dividing by two?

ghost:
In the current code and this PR the LUFS progress is only saved to database after all albums or all tracks have been completed.  If the normalisation is interrupted nothing is written to the database.  This negatively impacts large libraries that can take days to process.  Suggest doing _itemRepository.SaveItems([album or track], cancellationToken); incrementally for each album/track rather than saving all at once.

Shadowghost (changes requested):
We should save LUFS after each album was processed

entropicdrifter:
I think calling for every single track/album might be excessive and could cause this to be more I/O bound. Perhaps writing to DB on a set interval, say every 200 tracks or so would make more sense?

shanepowell:
Maybe time based interval would be better? e.g. 5 min interval?

shanepowell:
Also, shouldn't changing the saving to the db change be a separate push request as it's not really related to progress reporting.

crobibero (approved):

PR #14307, opened 2025-06-15
Fixed a hang the randomly happens running LUFS scan

shanepowell:
Fixed a hang the randomly happens (at least on windows) running ffmpeg and reading the StandardError output.

Switch to the reading the StandardError using the callback method.
Add TaskExtensions.WithTimeout to limit the time spent waiting for ffmpeg to complete (hardcoded to 30 mins).
Add more error handling around the running of ffmpeg so it continues on when running ffmpeg fails.


gnattu:
> Add TaskExtensions.WithTimeout to limit the time spent waiting for ffmpeg to complete (hardcoded to 30 mins).

This might not be enough for lufs scan especially for that concat file.

gnattu:
I don't really understand this problem as why it is "hang" and how your change could "fix" it. To me you are just killing the ffmpeg when it has took too long to process and you don't really care if it is really hung or actually doing things.

shanepowell:
The hang is in the call to ReadAllLinesAsync reading the std err output.  It hangs the ffmpeg process from exiting as well. 
It happens a alot on my machine.
If you google ReadAllLines/ReadToEnd hang when reading stdout/stderr you will find many examples of this happening to people.

gnattu:
How could you even google for this `ReadAllLinesAsync` implementation when it’s a Jellyfin-specific extension and finding "examples of this happening to people"?

https://github.com/jellyfin/jellyfin/blob/d1d9c8ed06c2f21607626d7f89fdf5e98be6e984/src/Jellyfin.Extensions/StreamExtensions.cs#L49-L63

It is a stream reader. It drains the buffer as we reads from it unlike `ReadToEnd`, which requires the buffer to be big enough to hold all output text, and that will hang the process because when the stdout/stderr buffer is full, any further writes will be blocked.

If Windows is experiencing a deadlock while processing the stderr, we need to handle it. However, I don’t believe a deadlock is the cause here. It’s more likely that the stderr is not properly line-terminated in Windows, causing the line reader to wait indefinitely.

src/Jellyfin.Extensions/TaskExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+    {
+        if (await TimeoutTask(task, timeout).ConfigureAwait(false))
+        {
+            throw new TimeoutException($"Timeout {timeout} exceeded");
+        }
+
+        return await task.ConfigureAwait(false);
+    }
+
+    private static async Task<bool> TimeoutTask(Task task, TimeSpan timeout)
```
JPVenson: There is already a WaitAsync method on the task
https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.waitasync?view=net-9.0

JPVenson (changes requested):

src/Jellyfin.Extensions/TaskExtensions.cs
shanepowell: Switched to WaitFor.

JPVenson:
closed due to inactivity

PR #14308, opened 2025-06-15
fix exception selecting the Content Type "Mixed Films and Programmes"

shanepowell:
fix exception selecting the Content Type "Mixed Films and Programmes" in the "Add Media Library" web client dialog.

CollectionType is minning the "mixed" type witct is what the web client sends to the server when you select the "Mixed Films and Programmes" type.


gnattu:
This library type is deprecated:

https://jellyfin.org/docs/general/server/media/mixed-movies-and-shows

shanepowell:
So should it sot be removed in the web client as a option then?

PR #14309, opened 2025-06-15
Fix PeopleHelper.AddPerson Exceptions due to bad meta-data extracts.

shanepowell:
Fix PeopleHelper.AddPerson ArgumentException.ThrowIfNullOrEmpty(person.Name) due to bad meta-data extracts.

Ignore metadata extracts with blank names.

### DIFF
--- a/MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
+++ b/MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
@@ -201,17 +201,14 @@ namespace MediaBrowser.Providers.MediaInfo
 
                 foreach (var albumArtist in albumArtists)
                 {
-                    var name = albumArtist.Trim();
-                    if (string.IsNullOrEmpty(name))
+                    if (!string.IsNullOrWhiteSpace(albumArtist))
                     {
-                        continue;
+                        PeopleHelper.AddPerson(people, new PersonInfo
+                        {
+                            Name = albumArtist,
+                            Type = PersonKind.AlbumArtist
+                        });
                     }
-
-                    PeopleHelper.AddPerson(people, new PersonInfo
-                    {
-                        Name = name,
-                        Type = PersonKind.AlbumArtist
-                    });
                 }
 
                 string[]? performers = null;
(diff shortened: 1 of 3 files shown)


MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
```diff
@@ -201,14 +201,17 @@ private async Task FetchDataFromTags(Audio audio, Model.MediaInfo.MediaInfo medi
 
                 foreach (var albumArtist in albumArtists)
                 {
-                    if (!string.IsNullOrWhiteSpace(albumArtist))
```
gnattu: `string.IsNullOrWhiteSpace()` already handles the case where the string will be null or empty after trimming. No need to do extra check here.

MediaBrowser.Providers/MediaInfo/FFProbeVideoInfo.cs
```diff
@@ -516,9 +516,15 @@ private void FetchPeople(Video video, Model.MediaInfo.MediaInfo data, MetadataRe
 
                 foreach (var person in data.People)
                 {
+                    var name = person.Name.Trim();
```
gnattu: Actually we don't really have to do the Trim if this trim is only for AddPerson. The AddPerson implementation already has a trim inside, and we only have to do the `IsNullOrWhiteSpace` check here.

MediaBrowser.Providers/Music/AlbumMetadataService.cs
```diff
@@ -197,18 +197,30 @@ private void SetPeople(MusicAlbum item)
 
             foreach (var albumArtist in item.AlbumArtists)
             {
+                var name = albumArtist.Trim();
```
gnattu: Same as ffprobe. Only do `IsNullOrWhiteSpace` here, no Trimming needed.

MediaBrowser.Providers/MediaInfo/FFProbeVideoInfo.cs
shanepowell: In my example the data is "    ".  This would get past the ArgumentException.ThrowIfNullOrEmpty(person.Name) check and you end up with a entry with a empty name (after trim), which I assume the code was trying to avoid.

MediaBrowser.Providers/Music/AlbumMetadataService.cs
shanepowell: I was trying to be constant in how PeopleHelper.AddPerson was called.  I can undo those changes if you like.

MediaBrowser.Providers/MediaInfo/FFProbeVideoInfo.cs
gnattu: This will be detected by `IsNullOrWhiteSpace`

Bond-009 (changes requested):
Don't remove the .Trims, these were intentionally added in #9560

gnattu:
> Don't remove the .Trims, these were intentionally added in #9560

No, the none of the trims being removed here are relevant as those are used solely for add person and nothing else.

And AddPerson does its own trim:

https://github.com/jellyfin/jellyfin/blob/d1d9c8ed06c2f21607626d7f89fdf5e98be6e984/MediaBrowser.Controller/Entities/PeopleHelper.cs#L18

PR #14047, opened 2025-05-02
Add multiple options for internal locking

JPVenson:
Add options for multiple query locking behaviors and query congestion logging for pessimistic mode

Fixes locking up the server when viewing faviorites

fixes https://github.com/jellyfin/jellyfin/issues/13708 
fixes https://github.com/jellyfin/jellyfin/issues/13873
fixes #13508

PR #14310, opened 2025-06-15
Fix SQLite Error 5: 'database is locked' and other db errors.

shanepowell:
The main problem that is happening is that jelly fin is breaking the sqlite Multi-thread mode rules and EF is not helping in any way. https://www.sqlite.org/threadsafe.html
EF uses the Multi-thread mode of sqlite.

These basically break down to:
* Reads can be multi-threaded
* Only one write operation can happen at any one time (this includes overlapping reads, i.e. a read during a write operation/transaction can cause the 'database is locked')
* Only one transaction scope at one time (no overlapping writes or transactions)

In sqlite you can set a retry timeout but that doesn't help if you transactions.

My fix is to use i reader writer lock primitive (multiple readers but only one writer) at i coarse-grained lock level. This means it's not automatically added to JellyFin but has to be added to db usage areas manually and maintained. I have done this for the current db usage and tried to limit the db lock scope as much as makes sense.

The implementation also supports recursive locking, including upgrading/downgrading locks.

- Added `DotNext.Threading` package reference for AsyncReaderWriterLock.
- Introduced `Jellyfin.Database.Implementations.Locking` for database locking.
- Updated classes to include `IEntityFrameworkDatabaseLockingBehavior` for managing locks.
- Refactored methods in `CleanDatabaseScheduledTask`, `UserDataManager`, and others to use reader/writer locks.
- Implemented new locking strategies with `ConcurrentDatabaseLockingBehavior` and `SerializedDatabaseLockingBehavior`.
- Improved error handling and logging for better insights during database operations.

Also fixed
Microsoft.EntityFrameworkCore.DbUpdateException: An error occurred while saving the entity changes. See the inner exception for details.
 ---> Microsoft.Data.Sqlite.SqliteException (0x80004005): SQLite Error 19: 'FOREIGN KEY constraint failed'.
...
   at Jellyfin.Database.Implementations.JellyfinDbContext.SaveChanges(Boolean acceptAllChangesOnSuccess) in D:\depot\github\jellyfin\src\Jellyfin.Database\Jellyfin.Database.Implementations\JellyfinDbContext.cs:line 286
   at Jellyfin.Server.Implementations.Item.PeopleRepository.UpdatePeople(Guid itemId, IReadOnlyList`1 people) in D:\depot\github\jellyfin\Jellyfin.Server.Implementations\Item\PeopleRepository.cs:line 106
   at Emby.Server.Implementations.Library.LibraryManager.UpdatePeopleAsync(BaseItem item, IReadOnlyList`1 people, CancellationToken cancellationToken) in D:\depot\github\jellyfin\Emby.Server.Implementations\Library\LibraryManager.cs:line 2864
   at Emby.Server.Implementations.Library.LibraryManager.UpdatePeople(BaseItem item, List`1 people) in D:\depot\github\jellyfin\Emby.Server.Implementations\Library\LibraryManager.cs:line 2850

With the multi-threaded nature of JellyFin, multiple overlapping calls to PeopleRepository.UpdatePeople can cause this problem. The write locking scope fixes this sort of problem, I also added a transaction boundary if JellyFin is gonig to support other db types where locking behavior can be turned off.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -17,6 +17,7 @@
     <PackageVersion Include="Diacritics" Version="3.3.29" />
     <PackageVersion Include="DiscUtils.Udf" Version="0.16.13" />
     <PackageVersion Include="DotNet.Glob" Version="3.1.3" />
+    <PackageVersion Include="DotNext.Threading" Version="5.22.0" />
     <PackageVersion Include="FsCheck.Xunit" Version="3.3.0" />
     <PackageVersion Include="HarfBuzzSharp.NativeAssets.Linux" Version="8.3.1.1" />
     <PackageVersion Include="ICU4N.Transliterator" Version="60.1.0-alpha.356" />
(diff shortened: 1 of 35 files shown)


gnattu:
Your mechanism is basically the same as the pessimistic locking already implemented here: https://github.com/jellyfin/jellyfin/pull/14047

where it will lock the whole db access when anything is attempting to write to db. This is safe and this is what we did for 10.10, but this could also hurt the performance.

PR #14337, opened 2025-06-18
Fix metadata manager infinite loading when Live TV access is revoked

### LINKED ISSUE #9908: [Issue]: Cannot edit metadata from metadata manager
### Please describe your bug

Since version 10.8.9 I cannot edit metadata from the Metadata Manager. Whenever I try to load it up I get an infinite loading screen:

image

In my browser's DevTools I also see the following errors:

```
main.jellyfin.bundle.js?dafe991ea4b0f4a6e0a5:2     GET http://jellyfin.lan:8096/LiveTv/Channels?limit=0 net::ERR_ABORTED 403 (Forbidden)

Response {type: 'basic', url: 'http://jellyfin.lan:8096/LiveTv/Channels?limit=0', redirected: false, status: 403, ok: false, …}

index.html#!/dashboard.html:1 Uncaught (in promise) Response {type: 'basic', url: 'http://jellyfin.lan:8096/LiveTv/Recordings?UserId=…false&EnableImageTypes=Primary%2CThumb%2CBackdrop', redirected: false, status: 403, ok: false, …}
```

I haven't configured Live TV and manually loading up that URL also results in a 403. Downgrading to 10.8.8 fixes the issue.

### Jellyfin Version

10.8.9

### if other:

10.8.10

### Environment

```markdown
- OS: unRAID 6.12
- Linux Kernel: n/a
- Virtualization: Docker
- Clients: Browser
- Browser: Chrome 114
- FFmpeg Version: none
- Playback Method: none
- Hardware Acceleration: none
- GPU Model: none
- Plugins: none
- Reverse Proxy: Nginx Proxy Manager (but also directly)
- Base URL: none
- Networking: Host (Docker macvlan)
- Storage: local
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

image

_No response_

### Code of Conduct

evanqhuang:
Resolves issue where the metadata manager would show infinite loading screen when users don't have Live TV access permissions. The problem was caused by 403 Forbidden errors from LiveTv/Channels and LiveTv/Recordings endpoints.

**Changes**- Remove [Authorize(Policy = Policies.LiveTvAccess)] from GetLiveTvChannels() and GetRecordings()
- Add manual permission checks that return empty QueryResult instead of 403 errors
- Check authenticated user permissions rather than target user permissions
- Preserve security by validating PermissionKind.EnableLiveTvAccess

This allows the metadata manager to receive empty results gracefully instead of failing with 403 errors, preventing the infinite loading issue while maintaining proper authorization.

**Issues**
Fixes #9908

### DIFF
--- a/Jellyfin.Api/Controllers/LiveTvController.cs
+++ b/Jellyfin.Api/Controllers/LiveTvController.cs
@@ -144,7 +144,6 @@ public class LiveTvController : BaseJellyfinApiController
     /// </returns>
     [HttpGet("Channels")]
     [ProducesResponseType(StatusCodes.Status200OK)]
-    [Authorize(Policy = Policies.LiveTvAccess)]
     public ActionResult<QueryResult<BaseItemDto>> GetLiveTvChannels(
         [FromQuery] ChannelType? type,
         [FromQuery] Guid? userId,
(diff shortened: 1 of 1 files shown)


crobibero:
The correct way to fix this would be to update the client.

Bond-009:
See comment from @crobibero

PR #14342, opened 2025-06-18
LibraryManager is doing too many image refresh operations

coderbone:
because it thinks that all images with no BlurHash need to be refreshed.

However, the blur hash computation code rejects svg files, so folders with svg files end up getting refreshed on every library scan.  This slows things down and adds a lot of write operations to the database which in turn aggravates the db locking problems.

**Changes**
LibraryManager ImageNeedsRefresh has been updated to excluded svg files in a manner similar  to SkiaEncoder/GetImageBlurHash.


Shadowghost (approved):

Bond-009 (approved):

Emby.Server.Implementations/Library/LibraryManager.cs
```diff
@@ -1947,11 +1947,24 @@ private bool ImageNeedsRefresh(ItemImageInfo image)
...
+                if (image.Width == 0 || image.Height == 0)
                 {
                     return true;
                 }
 
+                if (string.IsNullOrEmpty(image.BlurHash))
+                {
+                    // this is only refreshable if a blurhash can be computed
+                    var extension = Path.GetExtension(image.Path.AsSpan()).TrimStart('.');
+                    if (extension.Equals("svg", StringComparison.OrdinalIgnoreCase))
```
cvium: This is a bit leaky. It's a bigger change, but I would prefer having this logic in `ImageProcessor`. The library manager shouldn't need to know if blurhash can be computed.

cvium (changes requested):

Emby.Server.Implementations/Library/LibraryManager.cs
coderbone: certainly reasonable to hide the detail about what images can't have a blurhash computed moved somewhere else as there may be other places that should guard their call to GetBlurHash, however, LibraryManager is examining the blur hash value and deciding some amount of refresh should be triggered, without a check at that level it's going to go around and around with far more refreshing and database updates than are needed

JPVenson (approved):

Shadowghost (approved):

Bond-009 (approved):

Emby.Server.Implementations/Library/LibraryManager.cs
cvium: Sorry for the delay. It is fine that LibraryManager makes the check, but the logic is not for the manager to know. Your change is a step in the right direction, but the logic is now "duplicated" (it's slightly different actually): https://github.com/jellyfin/jellyfin/blob/master/src/Jellyfin.Drawing.Skia/SkiaEncoder.cs#L228

I think a proper solution is to put the logic inside IImageEncoder and use it in both places.

Emby.Server.Implementations/Library/LibraryManager.cs
```diff
@@ -2023,14 +2030,21 @@ public async Task UpdateImagesAsync(BaseItem item, bool forceUpdate = false)
...
                 try
                 {
-                    image.BlurHash = _imageProcessor.GetImageBlurHash(image.Path, size);
+                    if (_imageProcessor.CanComputeBlurHash(image.Path))
+                    {
+                        image.BlurHash = _imageProcessor.GetImageBlurHash(image.Path, size);
+                    }
+                    else
+                    {
+                        _logger.LogWarning("Cannot compute blurhash for {ImagePath}", image.Path);
```
IDisposable, line 2048: I doubt we want this warning every time an `.SVG` file is scanned. This is a normal case, and I know I wouldn't want to have to _keep_ wading through (just a different) message about BlurHashes on my many SVGs.

IDisposable (changes requested):

Emby.Server.Implementations/Library/LibraryManager.cs
crobibero: I agree, this log level should be changed to `Debug` so it can still be viewed

Shadowghost:
@coderbone can you please resolve the merge conflicts and the comment so we can go ahead with merging this?

coderbone:
> @coderbone can you please resolve the merge conflicts and the comment so we can go ahead with merging this?

I dropped the ball on that, been busy packing up house for a big move.

Merge conflict resolved but now that I look at the other "anyChange" logic that was introduced near my changes I'm wondering if my changes are needed at all.... I will put this PR back to draft mode and rethink the whole thing.

xuhao61:
不搞我  就没啥。


최준호


coderbone ***@***.***>于2025年7月20日 周日上午4:06写道：

JPVenson:
@xuhao61 this is a primarly English community, please only use English
这是一个主要英语社区，请只使用英语

cvium:
@coderbone Will you pick this up again? Otherwise I might

Emby.Server.Implementations/Library/LibraryManager.cs
```diff
@@ -2027,11 +2034,19 @@ public async Task UpdateImagesAsync(BaseItem item, bool forceUpdate = false)
                     image.Height = 0;
                 }
 
+                image.BlurHash = string.Empty;
```
IDisposable, line 2037: This will force line 2043 to always indicate a change...

enter-a-random-username:
@cvium would be awesome also to resolve issue with PNG logos failing if they are high-res and “extreme” ratio. Adds minutes to every scan, same as SVG. If I use a different for blur hashing it can be generated.
Workaround is only to add some default blur hash for empty or use an external script to write into Jellyfin DB (no API endpoint)

Issue #14317, opened 2025-06-16
[10.11.0-RC2] ArtistMetadataService refreshed on each scan due to file modified date


theguymadmax:
### This issue respects the following points:

- [x] This is a **bug**, not a question or a configuration issue; Please visit our [forum or chat rooms](https://jellyfin.org/contact/) first to troubleshoot with volunteers, before creating a report.
- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [x] I'm using an up to date version of Jellyfin Server stable, unstable or master; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Description of the bug

The `artist.nfo` file is regenerated for each artist in your music library every time the library is scanned, because its file modification date is not preserved.

Full logs: https://gist.github.com/theguymadmax/0e9e194b2eecc7e94530475a171ec92f

### Reproduction steps

1. On 10.11 - Scan in a music library
2. Turn on debugging logging.
3. Scan the library again

### What is the current _bug_ behavior?

The artist.nfo file is regenerated for each artist in your music library every time the library is scanned.

### What is the expected _correct_ behavior?

Music artist should not be refreshed. 

### Jellyfin Server version

Master

### Specify the build version

10.11 RC2

### Environment

```markdown
- OS: Windows 11
- Virtualization: No
- Plugins: None
```

### Jellyfin logs

```shell
1st scan after initial scan into library. 
[2025-06-15 22:48:53.244 -04:00] [DBG] [29] MediaBrowser.Providers.Music.ArtistMetadataService: File modification time changed from 01/01/0001 05:00:00 to 06/16/2025 02:47:51: "C:\Users\Max\AppData\Local\jellyfin\metadata\artists\David Bowie"
[2025-06-15 22:48:53.265 -04:00] [DBG] [29] MediaBrowser.Providers.Music.ArtistMetadataService: Running "ArtistNfoProvider" for "C:\Users\Max\AppData\Local\jellyfin\metadata\artists\David Bowie"
[...]

Issue #14303, opened 2025-06-13
[10.11.0-RC1] BDMV files get prunned on each scan


theguymadmax:
### This issue respects the following points:

- [x] This is a **bug**, not a question or a configuration issue; Please visit our [forum or chat rooms](https://jellyfin.org/contact/) first to troubleshoot with volunteers, before creating a report.
- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [x] I'm using an up to date version of Jellyfin Server stable, unstable or master; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Description of the bug

BDMV files are pruned on each scan. 

https://gist.github.com/theguymadmax/3ffd8e6581ace765fd834f6e405daa65

### Reproduction steps

1. On 10.11 - Add a movie library with a BDMV folder
2. After the file has been scanned, run the "Scan All Libraries" task

### What is the current _bug_ behavior?

File gets pruned

### What is the expected _correct_ behavior?

The file is not pruned

### Jellyfin Server version

Master

### Specify commit id

23c25289dae40f292f0ffcf8c83f8ea2f9eebd61

### Specify the build version

2025061219 (Master)

### Environment

```markdown
- OS: Windows 11
- Virtualization: No
- Plugins: None
- Media are on a network smb share 
- Jellyfin system is on NTFS
```

### Jellyfin logs

```shell
See link, entry was too long.
```

### FFmpeg logs

```shell

```

PR #14347, opened 2025-06-19
Fix modification checks and make sure to use UTC

Shadowghost:
**Changes**
* Ensure we always save UTC time for `DateModified` and `DateCreated`
* Simplify some checks

**Issues**
Maybe #14303 and #14317 if they were timezone related. I was unable to reproduce at least #14317

### DIFF
--- a/Emby.Photos/PhotoProvider.cs
+++ b/Emby.Photos/PhotoProvider.cs
@@ -108,7 +108,7 @@ public class PhotoProvider : ICustomMetadataProvider<Photo>, IForcedProvider, IH
                     var dateTaken = image.ImageTag.DateTime;
                     if (dateTaken.HasValue)
                     {
-                        item.DateCreated = dateTaken.Value;
+                        item.DateCreated = dateTaken.Value.ToUniversalTime();
                         item.PremiereDate = dateTaken.Value;
                         item.ProductionYear = dateTaken.Value.Year;
                     }
--- a/Emby.Server.Implementations/HttpServer/WebSocketConnection.cs
+++ b/Emby.Server.Implementations/HttpServer/WebSocketConnection.cs
@@ -57,7 +57,7 @@ namespace Emby.Server.Implementations.HttpServer
             RemoteEndPoint = remoteEndPoint;
 
             _jsonOptions = JsonDefaults.Options;
-            LastActivityDate = DateTime.Now;
+            LastActivityDate = DateTime.UtcNow;
         }
 
         /// <inheritdoc />
--- a/Emby.Server.Implementations/Images/BaseDynamicImageProvider.cs
+++ b/Emby.Server.Implementations/Images/BaseDynamicImageProvider.cs
@@ -43,7 +43,7 @@ namespace Emby.Server.Implementations.Images
         protected IImageProcessor ImageProcessor { get; set; }
 
         protected virtual IReadOnlyCollection<ImageType> SupportedImages { get; }
-            = new ImageType[] { ImageType.Primary };
+            = [ImageType.Primary];
 
         /// <inheritdoc />
         public string Name => "Dynamic Image Provider";
@@ -292,8 +292,14 @@ namespace Emby.Server.Implementations.Images
 
         protected virtual bool HasChangedByDate(BaseItem item, ItemImageInfo image)
         {
-            var age = DateTime.UtcNow - image.DateModified;
-            return age.TotalDays > MaxImageAgeDays;
+            var path = image.Path;
+            if (!string.IsNullOrEmpty(path))
+            {
+                var modificationDate = FileSystem.GetLastWriteTimeUtc(path);
+                return image.DateModified != modificationDate;
+            }
+
+            return false;
         }
 
         protected string CreateSingleImage(IEnumerable<BaseItem> itemsWithImages, string outputPathWithoutExtension, ImageType imageType)
--- a/Emby.Server.Implementations/Library/LibraryManager.cs
+++ b/Emby.Server.Implementations/Library/LibraryManager.cs
@@ -2991,13 +2991,12 @@ namespace Emby.Server.Implementations.Library
                     {
                         var path = Person.GetPath(person.Name);
                         var info = Directory.CreateDirectory(path);
-                        var lastWriteTime = info.LastWriteTimeUtc;
                         personEntity = new Person()
                         {
                             Name = person.Name,
                             Id = GetItemByNameId<Person>(path),
                             DateCreated = info.CreationTimeUtc,
-                            DateModified = lastWriteTime,
+                            DateModified = info.LastWriteTimeUtc,
                             Path = path
                         };
 
--- a/MediaBrowser.Controller/Entities/BaseItem.cs
+++ b/MediaBrowser.Controller/Entities/BaseItem.cs
@@ -1429,17 +1429,10 @@ namespace MediaBrowser.Controller.Entities
             }
 
             var info = FileSystem.GetFileSystemInfo(Path);
-            if (info.Exists)
-            {
-                if (info.IsDirectory)
-                {
-                    return info.LastWriteTimeUtc != DateModified;
-                }
 
-                return info.LastWriteTimeUtc != DateModified;
-            }
-
-            return false;
+            return info.Exists
+                ? info.LastWriteTimeUtc != DateModified
+                : false;
         }
 
         public virtual List<string> GetUserDataKeys()
--- a/MediaBrowser.Controller/Entities/Folder.cs
+++ b/MediaBrowser.Controller/Entities/Folder.cs
@@ -210,12 +210,12 @@ namespace MediaBrowser.Controller.Entities
                 item.Id = LibraryManager.GetNewItemId(item.Path, item.GetType());
             }
 
-            if (item.DateCreated == DateTime.MinValue)
+            if (item.DateCreated == default)
             {
                 item.DateCreated = DateTime.UtcNow;
             }
 
-            if (item.DateModified == DateTime.MinValue)
+            if (item.DateModified == default)
             {
                 item.DateModified = DateTime.UtcNow;
             }
--- a/MediaBrowser.Providers/Manager/MetadataService.cs
+++ b/MediaBrowser.Providers/Manager/MetadataService.cs
@@ -73,11 +73,11 @@ namespace MediaBrowser.Providers.Manager
 
         public virtual int Order => 0;
 
-        private FileSystemMetadata TryGetFile(string path, IDirectoryService directoryService)
+        private FileSystemMetadata TryGetFileSystemMetadata(string path, IDirectoryService directoryService)
         {
             try
             {
-                return directoryService.GetFile(path);
+                return directoryService.GetFileSystemEntry(path);
             }
             catch (Exception ex)
             {
@@ -225,7 +225,7 @@ namespace MediaBrowser.Providers.Manager
                 {
                     if (item.IsFileProtocol)
                     {
-                        var file = TryGetFile(item.Path, refreshOptions.DirectoryService);
+                        var file = TryGetFileSystemMetadata(item.Path, refreshOptions.DirectoryService);
                         if (file is not null)
                         {
                             item.DateModified = file.LastWriteTimeUtc;
--- a/tests/Jellyfin.Providers.Tests/Manager/MetadataServiceTests.cs
+++ b/tests/Jellyfin.Providers.Tests/Manager/MetadataServiceTests.cs
@@ -22,7 +22,7 @@ namespace Jellyfin.Providers.Tests.Manager
         {
             var newLocked = new[] { MetadataField.Genres, MetadataField.Cast };
             var newString = "new";
-            var newDate = DateTime.Now;
+            var newDate = DateTime.UtcNow;
 
             var oldLocked = new[] { MetadataField.Genres };
             var oldString = "old";
@@ -39,6 +39,7 @@ namespace Jellyfin.Providers.Tests.Manager
                     DateCreated = newDate
                 }
             };
+
             if (defaultDate)
             {
                 source.Item.DateCreated = default;
@@ -141,8 +142,8 @@ namespace Jellyfin.Providers.Tests.Manager
                 { "ProductionYear", 1, 2 },
                 { "CommunityRating", 1.0f, 2.0f },
                 { "CriticRating", 1.0f, 2.0f },
-                { "EndDate", DateTime.UnixEpoch, DateTime.Now },
-                { "PremiereDate", DateTime.UnixEpoch, DateTime.Now },
+                { "EndDate", DateTime.UnixEpoch, DateTime.UtcNow },
+                { "PremiereDate", DateTime.UnixEpoch, DateTime.UtcNow },
                 { "Video3DFormat", Video3DFormat.HalfSideBySide, Video3DFormat.FullSideBySide }
             };
 


JPVenson (approved):

coderbone:
I was also looking at those fields as part of a "why does some movies get refreshed over and over" problem and can offer a few more observations to consider while you are looking at this stuff.

In BaseItemRepository.cs, UpdateOrInsertItems I put this loop near the top of the function as it makes sense to update DateLastSaved at the point where items are actually being saved... otherwise we seemed to be saving a lot of items with DateLastSaved as default.

```
        foreach (var item in items)
        {
            item.DateLastSaved = DateTime.UtcNow;
        }
```

And in MetadataService.cs MergeBaseItemData the DateModified field is getting clobbered when replaceData is true and the "source" has a default DateModified but the target has a valid DateModified... unclear what to do there but this is working for me on my testing server:

```
                if (replaceData || source.DateModified != default)
                {
                    if (source.DateModified == default &&
                        target.DateModified != default)
                    {
                        Logger.LogInformation("hmmmm: " + source.Path);
                    }
                    else
                    {
                        target.DateModified = source.DateModified;
                    }
                }
```

Shadowghost:
@coderbone  - I agree with the 2nd point but the 1st point is not necessarily correct. Afaik the property is meant to denote when the external savers were run the last time, not when the item was saved last

coderbone:
Fair enough, still learning my way around the code base.

Jellyfin.Server/Migrations/Routines/FixDates.cs
```diff
@@ -0,0 +1,176 @@
...
+        var baseQuery = context.BaseItemImageInfos;
+        var records = baseQuery.Count();
+        _logger.LogInformation("Modifying dates for {Count} BaseItemImageInfos.", records);
+
+        var transaction = await context.Database.BeginTransactionAsync(cancellationToken).ConfigureAwait(false);
+        await using (transaction.ConfigureAwait(false))
+        {
+            do
+            {
+                var results = baseQuery.Skip(offset).Take(PageSize).ToList();
```
JPVenson: use the pagination helper

Jellyfin.Server/Migrations/Routines/FixDates.cs
```diff
@@ -0,0 +1,176 @@
...
+    private async Task FixBaseItemImageInfos(JellyfinDbContext context, Stopwatch sw, CancellationToken cancellationToken)
+    {
+        int offset = 0;
+
+        var baseQuery = context.BaseItemImageInfos;
+        var records = baseQuery.Count();
+        _logger.LogInformation("Modifying dates for {Count} BaseItemImageInfos.", records);
+
+        var transaction = await context.Database.BeginTransactionAsync(cancellationToken).ConfigureAwait(false);
+        await using (transaction.ConfigureAwait(false))
```
JPVenson: No need for a dedicated transaction, those modifications are all within the same SaveChanges which itself uses a transaction

JPVenson (changes requested):

Jellyfin.Server/Migrations/Routines/FixDates.cs
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        if (dateTime.Value.Year == 1 && dateTime.Value.Month == 1 && dateTime.Value.Day == 1)
+        {
+            return DateTime.MinValue;
+        }
+
+        return isUTC
+            ? dateTime.Value
+            : DateTime.SpecifyKind(dateTime.Value, DateTimeKind.Local).ToUniversalTime();
```
theguymadmax:
```suggestion
            return DateTime.SpecifyKind(DateTime.MinValue, DateTimeKind.Utc);
        }

        if (dateTime.Value.Kind == DateTimeKind.Utc || isUTC)
        {
            return dateTime.Value;
        }

        return DateTime.SpecifyKind(dateTime.Value, DateTimeKind.Local).ToUniversalTime();
```

Jellyfin.Server/Migrations/Routines/FixDates.cs
```diff
@@ -0,0 +1,134 @@
...
+    private async Task FixBaseItemsAsync(JellyfinDbContext context, Stopwatch sw, CancellationToken cancellationToken)
+    {
+        int itemCount = 0;
+
+        var baseQuery = context.BaseItems.OrderBy(e => e.Id);
+        var records = baseQuery.Count();
+        _logger.LogInformation("Fixing dates for {Count} BaseItems.", records);
+
+        await foreach (var result in context.BaseItems.OrderBy(e => e.Id)
+                        .WithPartitionProgress((partition) => _logger.LogInformation("Processed: {Count} - Time: {Time}", itemCount, sw.Elapsed))
```
theguymadmax:
```suggestion

        sw.Start();
        await foreach (var result in context.BaseItems.OrderBy(e => e.Id)
                        .WithPartitionProgress((partition) => _logger.LogInformation("Processing BaseItems batch {BatchNumber} ({ProcessedSoFar}/{TotalRecords}) - Time: {ElapsedTime}", partition + 1, partition * PageSize, records, sw.Elapsed))
```

Jellyfin.Server/Migrations/Routines/FixDates.cs
```diff
@@ -0,0 +1,134 @@
...
+                        .WithCancellation(cancellationToken)
+                        .ConfigureAwait(false))
+        {
+            result.DateCreated = ToUniversalTime(result.DateCreated);
+            result.DateLastMediaAdded = ToUniversalTime(result.DateLastMediaAdded);
+            result.DateLastRefreshed = ToUniversalTime(result.DateLastRefreshed);
+            result.DateLastSaved = ToUniversalTime(result.DateLastSaved);
+            result.DateModified = ToUniversalTime(result.DateModified);
+            itemCount++;
+        }
```
theguymadmax, line 83:
```suggestion
        }

        var saveCount = await context.SaveChangesAsync(cancellationToken).ConfigureAwait(false);
        _logger.LogInformation("BaseItems: Processed {ItemCount} items, saved {SaveCount} changes in {ElapsedTime}", itemCount, saveCount, sw.Elapsed);
```

Jellyfin.Server/Migrations/Routines/FixDates.cs
```diff
@@ -0,0 +1,134 @@
...
+    {
+        int itemCount = 0;
+
+        var baseQuery = context.Chapters;
+        var records = baseQuery.Count();
+        _logger.LogInformation("Fixing dates for {Count} Chapters.", records);
+
+        sw.Start();
+        await foreach (var result in context.Chapters.OrderBy(e => e.ItemId)
+                        .WithPartitionProgress((partition) => _logger.LogInformation("Processed: {Count} - Time: {Time}", itemCount, sw.Elapsed))
```
theguymadmax:
```suggestion
                        .WithPartitionProgress((partition) => _logger.LogInformation("Processing Chapters batch {BatchNumber} ({ProcessedSoFar}/{TotalRecords}) - Time: {ElapsedTime}", partition + 1, partition * PageSize, records, sw.Elapsed))
```

Jellyfin.Server/Migrations/Routines/FixDates.cs
```diff
@@ -0,0 +1,134 @@
...
+                        .PartitionEagerAsync(PageSize, cancellationToken)
+                        .WithCancellation(cancellationToken)
+                        .ConfigureAwait(false))
+        {
+            result.ImageDateModified = ToUniversalTime(result.ImageDateModified);
+            itemCount++;
+        }
```
theguymadmax, line 113:
```suggestion
        }

        var saveCount = await context.SaveChangesAsync(cancellationToken).ConfigureAwait(false);
        _logger.LogInformation("Chapters: Processed {ItemCount} items, saved {SaveCount} changes in {ElapsedTime}", itemCount, saveCount, sw.Elapsed);
```

Jellyfin.Server/Migrations/Routines/FixDates.cs
```diff
@@ -0,0 +1,134 @@
...
+    {
+        int itemCount = 0;
+
+        var baseQuery = context.BaseItemImageInfos;
+        var records = baseQuery.Count();
+        _logger.LogInformation("Fixing dates for {Count} BaseItemImageInfos.", records);
+
+        sw.Start();
+        await foreach (var result in context.BaseItemImageInfos.OrderBy(e => e.Id)
+                        .WithPartitionProgress((partition) => _logger.LogInformation("Processed: {Count} - Time: {Time}", itemCount, sw.Elapsed))
```
theguymadmax:
```suggestion
                        .WithPartitionProgress((partition) => _logger.LogInformation("Processing BaseItemImageInfos batch {BatchNumber} ({ProcessedSoFar}/{TotalRecords}) - Time: {ElapsedTime}", partition + 1, partition * PageSize, records, sw.Elapsed))
```

Jellyfin.Server/Migrations/Routines/FixDates.cs
```diff
@@ -0,0 +1,134 @@
...
+                        .PartitionEagerAsync(PageSize, cancellationToken)
+                        .WithCancellation(cancellationToken)
+                        .ConfigureAwait(false))
+        {
+            result.DateModified = ToUniversalTime(result.DateModified) ?? DateTime.MinValue;
+            itemCount++;
+        }
```
theguymadmax, line 143:
```suggestion
        }

        var saveCount = await context.SaveChangesAsync(cancellationToken).ConfigureAwait(false);
        _logger.LogInformation("BaseItemImageInfos: Processed {ItemCount} items, saved {SaveCount} changes in {ElapsedTime}", itemCount, saveCount, sw.Elapsed);
```

theguymadmax (changes requested):

JPVenson (approved):

Emby.Server.Implementations/Plugins/PluginManager.cs
```diff
@@ -423,7 +423,7 @@ public async Task<bool> PopulateManifest(PackageInfo packageInfo, Version versio
                 Overview = packageInfo.Overview,
                 Owner = packageInfo.Owner,
                 TargetAbi = versionInfo.TargetAbi ?? string.Empty,
-                Timestamp = string.IsNullOrEmpty(versionInfo.Timestamp) ? DateTime.MinValue : DateTime.Parse(versionInfo.Timestamp, CultureInfo.InvariantCulture),
+                Timestamp = string.IsNullOrEmpty(versionInfo.Timestamp) ? DateTime.MinValue : DateTimeOffset.Parse(versionInfo.Timestamp, CultureInfo.InvariantCulture).UtcDateTime,
```
cvium: Use `DateTimeStyles.AdjustToUniversal` instead of .UtcDateTime

Jellyfin.Server/Migrations/Routines/FixDates.cs
```diff
@@ -0,0 +1,147 @@
...
+    private DateTime? ToUniversalTime(DateTime? dateTime, bool isUTC = false)
+    {
+        if (dateTime is null)
+        {
+            return null;
+        }
+
+        if (dateTime.Value.Year == 1 && dateTime.Value.Month == 1 && dateTime.Value.Day == 1)
+        {
+            return DateTime.SpecifyKind(DateTime.MinValue, DateTimeKind.Utc);
```
cvium: Why not return null? Year 1 is not a valid date anyway

Jellyfin.Server/Migrations/Routines/FixDates.cs
```diff
@@ -0,0 +1,147 @@
...
+        {
+            result.DateModified = ToUniversalTime(result.DateModified) ?? DateTime.MinValue;
+            itemCount++;
+        }
+
+        var saveCount = await context.SaveChangesAsync(cancellationToken).ConfigureAwait(false);
+        _logger.LogInformation("BaseItemImageInfos: Processed {ItemCount} items, saved {SaveCount} changes in {ElapsedTime}", itemCount, saveCount, sw.Elapsed);
+    }
+
+    private DateTime? ToUniversalTime(DateTime? dateTime, bool isUTC = false)
```
cvium, line 149: IsUTC seems to always be false? (I'm on mobile though)

Jellyfin.Server/Migrations/Routines/FixDates.cs
```diff
@@ -0,0 +1,147 @@
...
+        {
+            return DateTime.SpecifyKind(DateTime.MinValue, DateTimeKind.Utc);
+        }
+
+        if (dateTime.Value.Kind == DateTimeKind.Utc || isUTC)
+        {
+            return dateTime.Value;
+        }
+
+        return DateTime.SpecifyKind(dateTime.Value, DateTimeKind.Local).ToUniversalTime();
```
cvium: ToUniversalTime already assumes Local if Kind is Unspecified

Shadowghost: Wow, `ToUniversalTime` assumes Local and `ToLocalTime` assume Universal...

Jellyfin.Server/Migrations/Routines/FixDates.cs
Shadowghost: DateTime.MinValue is `1/1/0001 12:00:00 AM` so year one is valid? This is a catchall check for getting all saved LocalTime-asjusted MinValue entries from the DB and re-save them as UTC MinValue

Jellyfin.Server/Migrations/Routines/FixDates.cs
```diff
@@ -0,0 +1,147 @@
...
+        _logger.LogInformation("Fixing dates for {Count} BaseItemImageInfos.", records);
+
+        sw.Start();
+        await foreach (var result in context.BaseItemImageInfos.OrderBy(e => e.Id)
+                        .WithPartitionProgress((partition) => _logger.LogInformation("Processing BaseItemImageInfos batch {BatchNumber} ({ProcessedSoFar}/{TotalRecords}) - Time: {ElapsedTime}", partition + 1, partition * PageSize, records, sw.Elapsed))
+                        .PartitionEagerAsync(PageSize, cancellationToken)
+                        .WithCancellation(cancellationToken)
+                        .ConfigureAwait(false))
+        {
+            result.DateModified = ToUniversalTime(result.DateModified) ?? DateTime.MinValue;
```
cvium: Should this not be null?

MediaBrowser.XbmcMetadata/Providers/BaseNfoProvider.cs
```diff
@@ -72,7 +73,7 @@ public bool HasChanged(BaseItem item, IDirectoryService directoryService)
                 return false;
             }
 
-            return file.Exists && _fileSystem.GetLastWriteTimeUtc(file) > item.DateLastSaved;
+            return file.Exists && item.DateLastSaved != DateTimeOffset.MinValue && _fileSystem.GetLastWriteTimeUtc(file) > item.DateLastSaved;
```
cvium: Is this still relevant?

Shadowghost: Since this happens after we transfer to a DTO, it should still be relevane. @theguymadmax maybe you can check again?

Jellyfin.Server/Migrations/Routines/FixDates.cs
Shadowghost: This is currently non-nullable. I think it makes sense to make it nullable

Jellyfin.Server/Migrations/Routines/FixDates.cs
```diff
@@ -0,0 +1,147 @@
...
+    {
+        int itemCount = 0;
+
+        var baseQuery = context.BaseItems.OrderBy(e => e.Id);
+        var records = baseQuery.Count();
+        _logger.LogInformation("Fixing dates for {Count} BaseItems.", records);
+
+        sw.Start();
+        await foreach (var result in context.BaseItems.OrderBy(e => e.Id)
+                        .WithPartitionProgress((partition) => _logger.LogInformation("Processing BaseItems batch {BatchNumber} ({ProcessedSoFar}/{TotalRecords}) - Time: {ElapsedTime}", partition + 1, partition * PageSize, records, sw.Elapsed))
```
crobibero: I think we should call `context.SaveChanges()` inside of the `endPartition` progress func so we don't keep all of the tracked changes in memory until the full list completes

Repeat for all fix functions

PR #14357, opened 2025-06-20
Fix issue with EncodedRecorder streams not showing up under "Active Recordings"

theshoeshiner:
**Changes**

Due to ffmpeg internals and delays associated with IPTV streams, the `EncodedRecorder` often does not cause a file to be created when a recording first starts. This causes the library refresh in `RecordingsManager`  to not detect the recording as present / active. Creating an empty file (if the recording file does not exist) fixes this. Even if ffmpeg doesnt flush data before the library refresh, the empty file allows the recording to show up under "**Active Recordings**" in the live tv "**Schedule**" tab.

### DIFF
--- a/src/Jellyfin.LiveTv/IO/EncodedRecorder.cs
+++ b/src/Jellyfin.LiveTv/IO/EncodedRecorder.cs
@@ -75,7 +75,7 @@ namespace Jellyfin.LiveTv.IO
             Directory.CreateDirectory(Path.GetDirectoryName(targetFile));
             if (!File.Exists(targetFile))
             {
-                await File.Create(targetFile).DisposeAsync().ConfigureAwait(false);
+                FileHelper.CreateEmpty(targetFile);
             }
 
             var processStartInfo = new ProcessStartInfo


src/Jellyfin.LiveTv/IO/EncodedRecorder.cs
```diff
@@ -73,6 +73,10 @@ private async Task RecordFromFile(MediaSourceInfo mediaSource, string inputFile,
         {
             _targetPath = targetFile;
             Directory.CreateDirectory(Path.GetDirectoryName(targetFile));
+            if (!File.Exists(targetFile))
+            {
+                await File.Create(targetFile).DisposeAsync().ConfigureAwait(false);
```
Bond-009: We have a helper to efficiently create new empty files https://github.com/jellyfin/jellyfin/blob/master/src/Jellyfin.Extensions/FileHelper.cs

Bond-009 (changes requested):

Bond-009 (approved):

crobibero (approved):

PR #14361, opened 2025-06-21
Add endpoint to fetch a random movie

SkyLostTR:
**Changes**

* Introduced a new `GET /Random` endpoint in the Movies controller.
* Accepts optional `userId` query parameter to filter results and attach user-specific data.
* Utilizes internal query to fetch a random movie with proper `DtoOptions`.
* Returns `200 OK` with the movie DTO if found; otherwise, returns `404 Not Found`.

**Issues**

* Improves API support for personalized or randomized movie suggestions.


nielsvanvelzen:
You can already use the existing item endpoints with sort by random to do this. Why add a new endpoint? Also, why modify the README file to explain this specific functionality?

Issue #14380, opened 2025-06-26
Slow performance of Music Album item


binarymelon:
### Description of the bug

When retrieving data for a music album the API call is very slow (6s+).  This causes a slowdown in a few places of the UI (context menu/actionsheet, album view, metadata manager, etc.)

### Reproduction steps

1. Sign into Jellyfin with a large library
2. Open context menu for an album

### What is the current _bug_ behavior?

Very slow response time /Users/{userId}/Item/{itemId} endpoint

### What is the expected _correct_ behavior?

Gotta go fast

### Jellyfin Server version

Unstable

### Specify unstable release number

10.11.0

### Specify the build version

2025062305

### Environment

```markdown
- OS: Unraid 7.0.1
- Linux Kernel: Linux version 6.6.78-Unraid (root@Develop-70) (gcc (GCC) 14.2.0, GNU ld version 2.43.1-slack151)
- Virtualization: Docker
- Clients: Any
- Browser: Any
- FFmpeg Version: 7.1.1-Jellyfin
- Playback Method: n/a
- Hardware Acceleration: n/a
- GPU Model: n/a
- Plugins: AniDB 10.2505.9.0 | AudioDB 10.11.0.0 | Cover Art Archive 8.2505.7.0 | LrcLib Lyrics 2.2505.1.0 | MusicBrainz 10.11.0.0 | OMDb 10.11.0.0 | Open Subtitles 20.2505.12.0 | Simkl 6.2505.8.0 | Studio Images 10.11.0.0 | TheTVDB 19.2505.13.0 | TMDb 10.11.0.0 | Trakt 26.2505.10.0
- Reverse Proxy: Nginx Proxy Manager
- Base URL: n/a
- Networking: n/a
- Jellyfin Data Storage: local nvme
- Media Storage: local hdd
- External Integrations: jellyseer, jellystat
```

### Jellyfin logs

```shell
[2025-06-25 10:05:41.025 -03:00] [INF] [51] Microsoft.EntityFrameworkCore.Database.Command: Executed DbCommand ("6,042"ms) [Parameters=["@__filter_IsFolder_0='?' (DbType = Boolean), @__isVirtualItem_Value_1='?' (DbType = Boolean), @__filter_AncestorIds_2='?' (Size = 40)"], CommandType='Text', CommandTimeout='30']"
""SELECT COUNT(*)
FROM \"BaseItems\" AS \"b\"
[...]

Issue #14336, opened 2025-06-17
API endpoint for Items significantly worsed in 10.11.0 RC2


bradbeattie:
### Description of the bug

After upgrading from 10.10.7 to 10.11.0 RC2, `/Users/<userid>/Items` as on the Collections page shows a significant drop in performance.

### Reproduction steps

1. Create a 30+ collections each with 100+ titles
2. Run Jellyfin 10.10.7. Load the Collections page with network inspection active, noting that response time is roughly 200ms
3. Run Jellyfin 10.11.0 RC 2. Try loading the Collections page again

### What is the current _bug_ behavior?

Response time to `/Users/<userid>/Items` is in the ballpark of 20s, roughly 100x slower than before.

### What is the expected _correct_ behavior?

Response time should be roughly in the same sub-1s ballpark.

### Jellyfin Server version

10.10.0+

### Specify commit id

f60281d8fd36af6cdb3357a1bbd3317e2724afe9

### Specify the build version

10.11.0 RC2

### Environment

```markdown
- OS: Linux, official docker compose image `2c70e3a6686355828ddb252183b3649d023b89d6b54cd1af798214fbc6736c9d`
```

### Jellyfin logs

```shell
Not applicable (no errors).
```

### FFmpeg logs

```shell

```

bradbeattie:
Possible relation to #14278.

JPVenson:
whats your library page size?

bradbeattie:
100.

bradbeattie:
Following https://jellyfin.org/docs/general/administration/troubleshooting/#debug-logging, I was able to generate 24MB of logs (10k+ select statements) from the single curl command to `/Users/<user_id>/Items?StartIndex=0&Limit=100&Fields=PrimaryImageAspectRatio%2CSortName%2CPath%2CChildCount%2CMediaSourceCount%2CPrimaryImageAspectRatio&ImageTypeLimit=1&ParentId=<collections_id>&SortBy=IsFolder%2CSortName&SortOrder=Ascending`. Happy to provide any specifics.

JPVenson:
Thanks for your help, the issue is already known and well understood due to leagcy code issues

theguymadmax:
Fixed by https://github.com/jellyfin/jellyfin/pull/14795

Issue #14285, opened 2025-06-11
[10.11.0-RC1] Library Filters are extremely slow


PhantomPhoton:
### Description of the bug

After upgrading to 10.11.0-RC1, when I go to my Show library, it loads up fine, but if I add any filtering (Unwatched, tag), it sits and spins and never loads up the results. I have reset my page size to be 100 per the release notes.

### Reproduction steps

1. Upgrade to 10.11.0-RC1
2. Navigate to Shows
3. Select Unwatched filter

### What is the current _bug_ behavior?

The loading spinner runs forever, the backend is doing work, but it appears that it takes too long to respond. The frontend javascript console shows a `Request timed out`, then it attempts to reconnect with the same filter, times out again, `Reporting request failure` and then a `Uncaught (in promise)` and then the interface halts with the spinner going until I force a refresh.

### What is the expected _correct_ behavior?

For the library to load with the filtered data

### Jellyfin Server version

Unstable

### Specify the build version

10.11.0-rc1

### Environment

```markdown
- OS: Debian
- Linux Kernel: 6.8.0-57
- Virtualization: Docker
- Clients: Browser
- Browser: Chrome
- FFmpeg Version: Whatever is bundled with 10.11.0-rc1
- Playback Method: N/A
- Hardware Acceleration: No
- GPU Model: N/A
- Plugins: None
- Reverse Proxy: Traefik
- Base URL: None
- Networking: Docker
- Jellyfin Data Storage: M2 SSD
- Media Storage: Local HDD
- External Integrations: None
```

### Jellyfin logs

```shell
Nothing specific looks relevant
```

### FFmpeg logs

```shell
N/A
```

Eyt-Lev:
was unable to replicate. did you run the server on localhost?

PhantomPhoton:
Yes, the server is running locally to the client.

felix920506:
what cpu and how much ram do you have?

Eyt-Lev:
How big is your library? I tried with 46 tv shows and 251 movies and can't replicate

PhantomPhoton:
I have a pair of xeon E5-2699A's and 768 gigabytes of ram. My library is very large. I ran 10.10 on a backup of the library and the same requests take under 10 seconds (usually around 6 seconds). I used the python api to bypass the timeout issues and 10.11 is taking around 70 seconds, so a little over 11x decrease in speed.

PhantomPhoton:
[...]

Issue #14278, opened 2025-06-10
10.11.0-RC1 Performance


markster1700:
### Description of the bug

Upgraded 10.10.7 to latest RC1. It took less than 10 minutes. Performed library scans and refreshed TV guide.
Let the container run stable for about 30 minutes then restarted. Tested mainly for performance with local network and Firefox client.

Comparing to 10.10.7 its noticeable performance degradation in some areas.
Photos takes up to 10-15 seconds every 3 pictures to show on the screen. Every 3rd picture would show black screen and wait for the fetch.

Live TV m3u streams takes longer to load.
Music and Movies performance is the same.

When saving new setting it takes up to 10 seconds to receive confirmation.

These are just few areas I think we may need to focus at this point.

### Reproduction steps

1. sign in
2. navigate to photos
3. browse photos

### What is the current _bug_ behavior?

1.After 3 photos I get black screen (waiting) for 10-15 seconds then the photo shows.
2. Streaming Live TV takes longer than 10.10.7
3. Saving config chnages takes ~5 seconds.

### What is the expected _correct_ behavior?

1.Should show photo within 1 second. Its only a small jpg file on local network.
2. 10.10.7 it is much faster in some cases about 1 seconds and starts streaming
3. Configuration saves in 10.10.7 takes 1 second

### Jellyfin Server version

Unstable

### Specify the build version

10.11.0 RC1

### Environment

```markdown
Official jellyfin docker 10.11.0RC1
```

### Jellyfin logs

```shell
NA
```

### FFmpeg logs

```shell

```

ClappTV:
I've also noticed slightly worse performance when loading images and in the interface overall, especially as the libraries were being scanned and got larger. When you have little content, the performance seems good, but as I kept adding the entire library, the performance got worse.

Even so, in my opinion, even in version 10.10 it takes too long to load images. It should be something really fast, especially since Jellyfin is supposed to compress the images — although I’m not sure if it’s enough to make the interface smoother. So I wouldn’t rule out the possibility that all this loading slows down the interface overall.

theguymadmax:
Are you getting any errors in your logs? Also how big is the photo library? I added a small photo library, 200 pictures, and the pictures in UI loaded up the same as the 10.10.z branch.  Also, it's recommend to perform a "Scan All Libraries" from the Dashboard after upgrade. Can you try that to see if it makes a difference.

markster1700:
[...]

PR #14382, opened 2025-06-26
Make server side evaluation of DTO parallel

JPVenson:
evaluates all properties added by the DTO creation in parallel to all requested items.

This is only a bandaid and will only help with performance on systems with adequate performance headroom but the alternative of rewriting all of the DTO creation so it can util all properties from the source query is a fundemental refactoring which cannot be done for 10.11.

Mitigates: #14336 , #14278 #14285 #14380


crobibero (approved):

Emby.Server.Implementations/Dto/DtoService.cs
```diff
@@ -88,7 +88,7 @@ public IReadOnlyList<BaseItemDto> GetBaseItemDtos(IReadOnlyList<BaseItem> items,
             List<(BaseItem, BaseItemDto)>? programTuples = null;
             List<(BaseItemDto, LiveTvChannel)>? channelTuples = null;
 
-            for (int index = 0; index < accessibleItems.Count; index++)
+            Enumerable.Range(0, accessibleItems.Count).AsParallel().ForAll(index =>
```
cvium, line 92: List.Add is not thread safe.

If item is `IItemByName` it does concurrent DB lookups, I don't know if that is a good idea.

enter-a-random-username:
This is awful performance wise.
Library paging with some items goes from 150-300 ms (without patch)  to 3-5 seconds (with this PR).

EraYaN:
> This is awful performance wise. Library paging with some items goes from 150-300 ms (without patch) to 3-5 seconds (with this PR).

What type of system are you running the server on? That might be relevant in finding the performance regression

binarymelon:
> > This is awful performance wise. Library paging with some items goes from 150-300 ms (without patch) to 3-5 seconds (with this PR).
> 
> What type of system are you running the server on? That might be relevant in finding the performance regression

Are there directions for generating a docker image?  I can also try running against the branch on my setup to compare.

enter-a-random-username:
> > This is awful performance wise. Library paging with some items goes from 150-300 ms (without patch) to 3-5 seconds (with this PR).
> 
> What type of system are you running the server on? That might be relevant in finding the performance regression

- I have around 8 P-Cores at 5+ Ghz + Hyperthreading (doubles them) enabled (no `taskset` is used to avoid E cores)
- Installed via package with my own PKGBUILD that allows for PR list and also a revert list. Similar to stable. Web is served static via nginx and not via jellyfin server
- All information (SQLite and assets -video) on a NVMe that reaches PCIe 4 bandwidth limit and few GB per sec are no issue at all.
- Removing this PR, and it is back to normal - I think I had no error or warning in the logs.
- Meilisearch disabled

How can this degrade so much for a simple listing so it needs 30 times the runtime or so

binarymelon:
I'm not seeing any additional slowdown over RC2 for library paging but I'm also not seeing any improvements for #14380.

Mine is running in docker container on Unraid.  Ryzen 3900x and sqlite is running from local nvme.

crobibero:
Since the parallelization can have a poor impact on some systems I think the query updates to directly pull count should be moved to a fresh PR

JPVenson:
@crobibero since this also introduces bugs related to watch count indicators, i am thinking of scrapping the PR in its entirety

JPVenson:
Closed as irrelevant now

PR #14387, opened 2025-06-27
Cleanup compiler warnings

Shadowghost:
Most of the property should not be array warnings need to be ignored due to being API responses.

### DIFF
--- a/Emby.Naming/Common/EpisodeExpression.cs
+++ b/Emby.Naming/Common/EpisodeExpression.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.Text.RegularExpressions;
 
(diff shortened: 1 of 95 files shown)


Shadowghost:
Likely blocked until after 10.11 release due to ABI changes

gnattu (approved):
Looks like no functional changes to me

EraYaN:
I believe ICollection<> would return the same openapi.yaml in swagger as an array.

gabparrot (commented):
This pr should go further if we're going in that direction. Why are some property arrays replaced by nullable arrays instead of getting rid of the arrays altogether? (ex ChannelQuery.cs)

JPVenson:
closed due to inactivity

Issue #14350, opened 2025-06-19
[Group] Jellyfin 10.11.0 RC bugs & issues


theguymadmax:
This issue tracks bugs related to the 10.11.0 Release Candidates.

nyanmisaka:
https://github.com/jellyfin/jellyfin-web/issues/6881 Is this intentional or a regression?

thornbill:
It is a regression

MiTvSpain:
When adding a series with more than 80 episodes, it deletes the history of recently added episodes and only the most recently added one appears. I think it should be possible to change the number of episodes to display. Thanks.

Shadowghost:
It has always been like this, so no new issue related to the RC.

MiTvSpain:
Ok, pero podrían modificar esa opción para que deje de ser siempre asi y se mas visible. gracias

Ch3ck3rM0n:
TVdB plugin is not working after upgrade to RC3 and so my library is build up again and now looking aweful in parts. I make a rollback till the plugin is supported.

crobibero:
> TVdB plugin is not working after upgrade to RC3 and so my library is build up again and now looking aweful in parts. I make a rollback till the plugin is supported.

Please use the unstable plugin repo linked in the release notes.

Ch3ck3rM0n:
> > TVdB plugin is not working after upgrade to RC3 and so my library is build up again and now looking aweful in parts. I make a rollback till the plugin is supported.
> 
> Please use the unstable plugin repo linked in the release notes.

i just can find the link to the stable once https://repo.jellyfin.org/?path=/plugin but not to unstable-plugins.. can you share it?

PR #14396, opened 2025-06-29
Fix movie recommendations

### LINKED ISSUE #14088: Movie Suggestion on populating on 10.11

### Description of the bug

Movie Suggestions do not populate on Master/unstable

### Reproduction steps

1. Setup an 10.11 server instance 
2. Scan in movie library
3. Mark movie items as watched
4. Navigate to the movies suggestions page

No movie suggestions are currently available. Start watching and rating your movies, and then come back to view your recommendations.

Movie suggestions should be populated.

### Jellyfin Server version

Master

### Specify commit id

https://github.com/jellyfin/jellyfin/commit/a7bb3ea21489a3a4c9a3e9f60c9521eaa84bad31

_No response_

### Specify version number

_No response_

### Specify the build version

10.11 (Master)

### Environment

```markdown
- OS:Windows 11
- FFmpeg Version: 7.1.1-3 (GH Artifact)
- Plugins: None
```

### Jellyfin logs

```shell
None
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

Image

### Additional information

_No response_

Eyt-Lev:
**Changes**
Fix movie recommendations by ditching `IsMovie` (I would really like to know what is its use over `BaseItemKind.Movie`, anyway it just wasn't working so I removed it) and using `Tags` and `Genres` instead.
Also changed the categories to be more diverse and biased towards liked movies.

**Issues**
fixes #14088 (Parent #14350)

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -199,6 +199,7 @@
  - [allesmi](https://github.com/allesmi)
  - [ThunderClapLP](https://github.com/ThunderClapLP)
  - [Shoham Peller](https://github.com/spellr)
+ - [Eytam Levine](https://github.com/Eyt-Lev)
 
 # Emby Contributors
 
--- a/Jellyfin.Api/Controllers/LibraryController.cs
+++ b/Jellyfin.Api/Controllers/LibraryController.cs
@@ -779,6 +779,7 @@ public class LibraryController : BaseJellyfinApiController
         var query = new InternalItemsQuery(user)
         {
             Genres = item.Genres,
+            Tags = item.Tags,
             Limit = limit,
             IncludeItemTypes = includeItemTypes.ToArray(),
             DtoOptions = dtoOptions,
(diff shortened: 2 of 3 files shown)


gnattu:
Ahh now I know the problem.

The `IsMovie` is an emby era design decision to differentiate the LiveTV movie programs from the rest of other LiveTV programs, as the LiveTV programs will not have `BaseItemKind.Movie`, but instead having `BaseItemKind.LiveTvProgram`.

And there could be some nasty emby era workarounds that only applies the `IsMovie` query to non-movie baseitems and such logic is not present in the refactored new db code.

I prefer to not use this anymore too, but if we want to support LiveTV movies in the recommendations we might want to think of a better way of doing this.

Eyt-Lev:
So the solution would be to add `BaseItemKind.LiveTvMovieProgram` and get entirely rid of `IsMovie`? I can try it if that's what's necessary.
But I Assume that the team don't want this stuff right before the release so I guess we can merge this now and Deal With It Later™.

gnattu:
Adding new base item kind would be too much for 10.11 now.

Eyt-Lev:
So you're saying that we should find a way to make it work with LiveTV movie programs without major rewriting for now, or is it good to go just to patch this up or now?

PR #14404, opened 2025-06-30
Fix duplicate media entries when sorting

### LINKED ISSUE #14403: Improper Handling of Sort Words
### Description of the bug

Jellyfin is returning duplicated titles that have sort words stripped from them.

Example, if there are movies:

1. The Foo
2. The Bar
3. The Baz

The API appears to be returning:

- Foo
- Bar
- Baz
- The Foo
- The Bar
- The Baz

Based on the fact that this is showing up on both the Android TV App and the Jellyfin Web UI.

### Reproduction steps

1. Login to the Jellyfin Web App.
2. Go to a Library
3. See duplicate media showing up. (For instance, if "The Foo" is an entry, it will be seen both under "F" and "T").

The API appears to be returning the same movie twice if there are sort words from the SortRemoveWords list.

Any media should only be returned once and only once from the API when a call is made, we should not see dups in the return data.

### Jellyfin Server version

10.10.0+

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11-rc2

### Environment

```markdown
OS: Fedora Workstation 42
Linux Kernel: 6.15.3
Clients: Browser, Android
```

### Jellyfin logs

```shell
N/A
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

theguymadmax:
**Changes**
Updated filtering logic to prevent duplicate media items caused by SortName and Name comparisons.

**Issues**
Fixes #14403

### DIFF
--- a/Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
+++ b/Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
@@ -1798,7 +1798,7 @@ public sealed class BaseItemRepository
 
         if (!string.IsNullOrWhiteSpace(filter.NameStartsWith))
         {
-            baseQuery = baseQuery.Where(e => e.SortName!.StartsWith(filter.NameStartsWith) || e.Name!.StartsWith(filter.NameStartsWith));
+            baseQuery = baseQuery.Where(e => e.SortName!.StartsWith(filter.NameStartsWith));
         }
 
         if (!string.IsNullOrWhiteSpace(filter.NameStartsWithOrGreater))


crobibero:
This doesn't seem like a correct fix, the original issue implies that the item is stored twice in the database

theguymadmax:
Technically, name is stored in at least three different variations in the database. In this case, we should only be using `SortName`. When a user clicks on "M" in the alphapicker, `SortName`, which uses `SortRemoveWords,` returns "matrix". But if they click on "T", `Name` returns "The Matrix", which is why we’re seeing duplicate results.

CleanName:
the matrix

Name:
The Matrix

SortName
matrix

EraYaN:
So you mean they are duplicated across letters, not within one list? It might be desirable to have them be returned for both letters to make finding them easier no?

theguymadmax:
That's correct, it's only duplicated when sorting by letter using the alphapicker. This happens because `SortName` applies `SortRemoveWords`("A", "An", and "The") to the title, which doesn't match the value in `Name`.

Leaving it as-is would be a design change from how it has been implemented, which only applied `SortName`:
https://github.com/jellyfin/jellyfin/blob/55f5eaf0e313783a8e45fa3f3d63a8aea5d91f48/Emby.Server.Implementations/Data/SqliteItemRepository.cs#L3450

 In fact, the other comparisons between `SortName` and `Name` should probably be updated too, but I’m not sure where those queries are actually applied, so I haven’t been able to test them.

I think we should aim for consistency and stick to the current behavior. Since the library is arranged by `SortName` when unsorted, we should also filter by that value.  Also, I’d say it's actually harder to find things when both comparisons are used. For example, if I use `Name` and apply the "T" filter, I get around 500 entries because "The" is such a common word, compared to about 100 when using `SortName`.

PR #14407, opened 2025-07-01
Add additional chinese languages

dyphire:
**Changes**
Add Chinese Simplified as additional metadata languages.

**Issues**
Fixes #9285

PR #14410, opened 2025-07-01
Further refinement of BCP 47 language labeling support

### LINKED ISSUE #6302: Language code in subtitles problem with Chinese

dyphire:
Current support for BCP 47 language tags is still very much lacking, with some languages not being recognized and displayed correctly.
They are correctly supported through regionalized language code recognition in `.NET`.
Refer https://learn.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.getcultures?view=net-9.0

Before:
image
image

After:
image
image

**Changes**
Based on #14407
Further refinement of BCP 47 language labeling support

**Issues**
Fixes #6302

### DIFF
--- a/Emby.Naming/ExternalFiles/ExternalPathParser.cs
+++ b/Emby.Naming/ExternalFiles/ExternalPathParser.cs
@@ -97,14 +97,18 @@ namespace Emby.Naming.ExternalFiles
 
                     if (culture is not null && pathInfo.Language is null)
                     {
-                        pathInfo.Language = culture.ThreeLetterISOLanguageName;
+                        pathInfo.Language = culture.Name.Contains('-', StringComparison.OrdinalIgnoreCase)
+                                          ? culture.Name
+                                          : culture.ThreeLetterISOLanguageName;
                         extraString = extraString.Replace(currentSlice, string.Empty, StringComparison.OrdinalIgnoreCase);
                     }
                     else if (culture is not null && pathInfo.Language == "hin")
                     {
                         // Hindi language code "hi" collides with a hearing impaired flag - use as Hindi only if no other language is set
                         pathInfo.IsHearingImpaired = true;
-                        pathInfo.Language = culture.ThreeLetterISOLanguageName;
+                        pathInfo.Language = culture.Name.Contains('-', StringComparison.OrdinalIgnoreCase)
+                                          ? culture.Name
+                                          : culture.ThreeLetterISOLanguageName;
                         extraString = extraString.Replace(currentSlice, string.Empty, StringComparison.OrdinalIgnoreCase);
                     }
                     else if (_namingOptions.MediaHearingImpairedFlags.Any(s => currentSliceWithoutSeparator.Equals(s, StringComparison.OrdinalIgnoreCase)))
--- a/Emby.Server.Implementations/Library/MediaSourceManager.cs
+++ b/Emby.Server.Implementations/Library/MediaSourceManager.cs
@@ -379,7 +379,7 @@ namespace Emby.Server.Implementations.Library
             var culture = _localizationManager.FindLanguageInfo(language);
             if (culture is not null)
             {
-                return culture.ThreeLetterISOLanguageNames;
+                return culture.Name.Contains('-', StringComparison.OrdinalIgnoreCase) ? [culture.Name] : culture.ThreeLetterISOLanguageNames;
             }
 
             return [language];
--- a/Jellyfin.Api/Controllers/ItemUpdateController.cs
+++ b/Jellyfin.Api/Controllers/ItemUpdateController.cs
@@ -158,7 +158,10 @@ public class ItemUpdateController : BaseJellyfinApiController
             ParentalRatingOptions = _localizationManager.GetParentalRatings().ToList(),
             ExternalIdInfos = _providerManager.GetExternalIdInfos(item).ToArray(),
             Countries = _localizationManager.GetCountries().ToArray(),
-            Cultures = _localizationManager.GetCultures().ToArray()
+            Cultures = _localizationManager.GetCultures()
+                .DistinctBy(c => c.DisplayName, StringComparer.OrdinalIgnoreCase)
+                .OrderBy(c => c.DisplayName)
+                .ToArray()
         };
 
         if (!item.IsVirtualItem
--- a/Jellyfin.Api/Controllers/LocalizationController.cs
+++ b/Jellyfin.Api/Controllers/LocalizationController.cs
@@ -1,4 +1,6 @@
+using System;
 using System.Collections.Generic;
+using System.Linq;
 using MediaBrowser.Common.Api;
 using MediaBrowser.Model.Entities;
 using MediaBrowser.Model.Globalization;
@@ -34,7 +36,14 @@ public class LocalizationController : BaseJellyfinApiController
     [ProducesResponseType(StatusCodes.Status200OK)]
     public ActionResult<IEnumerable<CultureDto>> GetCultures()
     {
-        return Ok(_localization.GetCultures());
+        var allCultures = _localization.GetCultures();
+
+        var distinctCultures = allCultures
+            .DistinctBy(c => c.DisplayName, StringComparer.OrdinalIgnoreCase)
+            .OrderBy(c => c.DisplayName)
+            .AsEnumerable();
+
+        return Ok(distinctCultures);
     }
 
     /// <summary>
--- a/MediaBrowser.Model/Entities/MediaStream.cs
+++ b/MediaBrowser.Model/Entities/MediaStream.cs
@@ -273,11 +273,28 @@ namespace MediaBrowser.Model.Entities
                         // Do not display the language code in display titles if unset or set to a special code. Show it in all other cases (possibly expanded).
                         if (!string.IsNullOrEmpty(Language) && !_specialCodes.Contains(Language, StringComparison.OrdinalIgnoreCase))
                         {
-                            // Get full language string i.e. eng -> English.
-                            string fullLanguage = CultureInfo
-                                .GetCultures(CultureTypes.NeutralCultures)
-                                .FirstOrDefault(r => r.ThreeLetterISOLanguageName.Equals(Language, StringComparison.OrdinalIgnoreCase))
-                                ?.DisplayName;
+                            // Get full language string i.e. eng -> English, zh-Hans -> Chinese (Simplified).
+                            var cultures = CultureInfo.GetCultures(CultureTypes.NeutralCultures);
+                            CultureInfo match = null;
+                            if (Language.Contains('-', StringComparison.OrdinalIgnoreCase))
+                            {
+                                match = cultures.FirstOrDefault(r =>
+                                    r.Name.Equals(Language, StringComparison.OrdinalIgnoreCase));
+
+                                if (match is null)
+                                {
+                                    string baseLang = Language.AsSpan().LeftPart('-').ToString();
+                                    match = cultures.FirstOrDefault(r =>
+                                        r.TwoLetterISOLanguageName.Equals(baseLang, StringComparison.OrdinalIgnoreCase));
+                                }
+                            }
+                            else
+                            {
+                                match = cultures.FirstOrDefault(r =>
+                                    r.ThreeLetterISOLanguageName.Equals(Language, StringComparison.OrdinalIgnoreCase));
+                            }
+
+                            string fullLanguage = match?.DisplayName;
                             attributes.Add(StringHelper.FirstToUpper(fullLanguage ?? Language));
                         }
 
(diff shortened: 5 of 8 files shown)


Emby.Naming/ExternalFiles/ExternalPathParser.cs
```diff
@@ -1,3 +1,5 @@
+#pragma warning disable CA1307
```
Shadowghost: Is there a reason you disable this? If not, making it compliant is fairly simple (just check out any other comparison in the code base)

dyphire: I just followed what was done in the existing code:
https://github.com/jellyfin/jellyfin/blob/38d0367c428fdce7e42c8e70da11327ecbba2c6c/Jellyfin.Server.Implementations/Item/BaseItemRepository.cs#L446

But apparently the code base is not consistent in its behavior of using `Contains`, and I will implement your suggestion instead.

dyphire: Done.

dyphire:
Any suggestions and progress?

Emby.Naming/ExternalFiles/ExternalPathParser.cs
Bond-009: It's disabled only for EFCore interactions

Emby.Server.Implementations/Library/MediaSourceManager.cs
```diff
@@ -379,7 +379,7 @@ private IReadOnlyList<string> NormalizeLanguage(string language)
             var culture = _localizationManager.FindLanguageInfo(language);
             if (culture is not null)
             {
-                return culture.ThreeLetterISOLanguageNames;
+                return [culture.Name.Contains('-', StringComparison.OrdinalIgnoreCase) ? culture.Name : culture.ThreeLetterISOLanguageName];
```
Bond-009: Why don't you return all `ThreeLetterISOLanguageNames`?
```suggestion
                return culture.Name.Contains('-', StringComparison.OrdinalIgnoreCase) ? [culture.Name] : culture.ThreeLetterISOLanguageNames;
```

dyphire: Ah, here is a mistake. fixed.

Jellyfin.Api/Controllers/ItemUpdateController.cs
```diff
@@ -158,7 +158,11 @@ public ActionResult<MetadataEditorInfo> GetMetadataEditorInfo([FromRoute, Requir
             ParentalRatingOptions = _localizationManager.GetParentalRatings().ToList(),
             ExternalIdInfos = _providerManager.GetExternalIdInfos(item).ToArray(),
             Countries = _localizationManager.GetCountries().ToArray(),
-            Cultures = _localizationManager.GetCultures().ToArray()
+            Cultures = _localizationManager.GetCultures()
+                .GroupBy(c => c.DisplayName, StringComparer.OrdinalIgnoreCase)
+                .Select(g => g.First())
```
Bond-009:
```suggestion
                .DistinctBy(c => c.DisplayName, StringComparer.OrdinalIgnoreCase)
```

Jellyfin.Api/Controllers/LocalizationController.cs
```diff
@@ -34,7 +36,15 @@ public LocalizationController(ILocalizationManager localization)
     [ProducesResponseType(StatusCodes.Status200OK)]
     public ActionResult<IEnumerable<CultureDto>> GetCultures()
     {
-        return Ok(_localization.GetCultures());
+        var allCultures = _localization.GetCultures();
+
+        var distinctCultures = allCultures
+            .GroupBy(c => c.DisplayName, StringComparer.OrdinalIgnoreCase)
+            .Select(g => g.First())
```
Bond-009:
```suggestion
            .DistinctBy(c => c.DisplayName, StringComparer.OrdinalIgnoreCase)
```

MediaBrowser.Model/Entities/MediaStream.cs
```diff
@@ -273,11 +273,28 @@ public string DisplayTitle
...
-                                ?.DisplayName;
+                            // Get full language string i.e. eng -> English, zh-Hans -> Chinese (Simplified).
+                            var cultures = CultureInfo.GetCultures(CultureTypes.NeutralCultures);
+                            CultureInfo match = null;
+                            if (Language.Contains("-", StringComparison.OrdinalIgnoreCase))
+                            {
+                                match = cultures.FirstOrDefault(r =>
+                                    r.Name.Equals(Language, StringComparison.OrdinalIgnoreCase));
+
+                                if (match == null)
```
Bond-009:
```suggestion
                                if (match is null)
```

MediaBrowser.Model/Entities/MediaStream.cs
```diff
@@ -273,11 +273,28 @@ public string DisplayTitle
...
+                                {
+                                    string baseLang = Language.Split('-')[0];
```
Bond-009: Use our `LeftPart` extension method instead

MediaBrowser.Model/Entities/MediaStream.cs
```diff
@@ -273,11 +273,28 @@ public string DisplayTitle
...
                         {
-                            // Get full language string i.e. eng -> English.
-                            string fullLanguage = CultureInfo
-                                .GetCultures(CultureTypes.NeutralCultures)
-                                .FirstOrDefault(r => r.ThreeLetterISOLanguageName.Equals(Language, StringComparison.OrdinalIgnoreCase))
-                                ?.DisplayName;
+                            // Get full language string i.e. eng -> English, zh-Hans -> Chinese (Simplified).
+                            var cultures = CultureInfo.GetCultures(CultureTypes.NeutralCultures);
+                            CultureInfo match = null;
+                            if (Language.Contains("-", StringComparison.OrdinalIgnoreCase))
```
Bond-009: Use the char overload instead

MediaBrowser.Model/Entities/MediaStream.cs
```diff
@@ -376,11 +393,28 @@ public string DisplayTitle
...
+                            {
+                                match = cultures.FirstOrDefault(r =>
+                                    r.Name.Equals(Language, StringComparison.OrdinalIgnoreCase));
+
+                                if (match == null)
```
Bond-009:
```suggestion
                                if (match is null)
```

MediaBrowser.Model/Entities/MediaStream.cs
```diff
@@ -376,11 +393,28 @@ public string DisplayTitle
...
                         {
-                            // Get full language string i.e. eng -> English.
-                            string fullLanguage = CultureInfo
-                                .GetCultures(CultureTypes.NeutralCultures)
-                                .FirstOrDefault(r => r.ThreeLetterISOLanguageName.Equals(Language, StringComparison.OrdinalIgnoreCase))
-                                ?.DisplayName;
+                            // Get full language string i.e. eng -> English, zh-Hans -> Chinese (Simplified).
+                            var cultures = CultureInfo.GetCultures(CultureTypes.NeutralCultures);
+                            CultureInfo match = null;
+                            if (Language.Contains("-", StringComparison.OrdinalIgnoreCase))
```
Bond-009: Same as above

MediaBrowser.Model/Entities/MediaStream.cs
```diff
@@ -376,11 +393,28 @@ public string DisplayTitle
...
+                            {
+                                match = cultures.FirstOrDefault(r =>
+                                    r.Name.Equals(Language, StringComparison.OrdinalIgnoreCase));
+
+                                if (match == null)
+                                {
+                                    string baseLang = Language.Split('-')[0];
```
Bond-009: Use our LeftPart extension method instead

MediaBrowser.Model/Entities/MediaStream.cs
dyphire: Done.

MediaBrowser.Model/Entities/MediaStream.cs
dyphire: Done.

MediaBrowser.Model/Entities/MediaStream.cs
```diff
@@ -273,11 +273,28 @@ public string DisplayTitle
...
                         {
-                            // Get full language string i.e. eng -> English.
-                            string fullLanguage = CultureInfo
-                                .GetCultures(CultureTypes.NeutralCultures)
-                                .FirstOrDefault(r => r.ThreeLetterISOLanguageName.Equals(Language, StringComparison.OrdinalIgnoreCase))
-                                ?.DisplayName;
+                            // Get full language string i.e. eng -> English, zh-Hans -> Chinese (Simplified).
+                            var cultures = CultureInfo.GetCultures(CultureTypes.NeutralCultures);
+                            CultureInfo match = null;
+                            if (Language.Contains('-'))
```
Shadowghost: Should use `StringComparison.OrdinalIgnoreCase`

MediaBrowser.Model/Entities/MediaStream.cs
```diff
@@ -376,11 +393,28 @@ public string DisplayTitle
...
```
Shadowghost: Same as above

MediaBrowser.Model/Entities/MediaStream.cs
dyphire: This is exactly the previous behavior, but @Bond-009  suggested using char overload instead.

Bond-009: Sorry, I should have been more specific. I meant the `public bool Contains(char value, StringComparison comparisonType)` overload.

dyphire: Looks like I misunderstood, fix.

Bond-009 (approved):

Jellyfin.Api/Controllers/ItemUpdateController.cs
```diff
@@ -158,7 +158,10 @@ public ActionResult<MetadataEditorInfo> GetMetadataEditorInfo([FromRoute, Requir
             ParentalRatingOptions = _localizationManager.GetParentalRatings().ToList(),
             ExternalIdInfos = _providerManager.GetExternalIdInfos(item).ToArray(),
             Countries = _localizationManager.GetCountries().ToArray(),
-            Cultures = _localizationManager.GetCultures().ToArray()
+            Cultures = _localizationManager.GetCultures()
```
crobibero, line 161: Since both usages of `GetCultures()` filter by distinct and then order I think `_cultures` should just store this list

dyphire: The original list is still required in `FindLanguageInfo()` to work.

Also used in `tests/Jellyfin.Server.Implementations.Tests/Localization/LocalizationManagerTests.cs`:
```cs
var cultures = localizationManager.GetCultures().ToList();
```

crobibero: Ok that makes sense, then we can add an additional call or something to avoid recreating the distinct list every time. I haven't looked closely but I think that is a hot path

PR #14420, opened 2025-07-04
Add 2nd level cache for DB queries

Shadowghost:
Since we're now fully on EFcore, let's try this again

**Changes**
* Add 2nd level cache
* Use CacheManager Bond Serializer ([Benchmark](https://github.com/MichaCo/CacheManager/blob/dev/benchmarks/CacheManager.Benchmarks/SerializationBenchmark.cs))

**Comments**
* This increased UI responsiveness immensly for me
* So far I haven't encountered any issues

### DIFF
--- a/Jellyfin.Server.Implementations/Extensions/ServiceCollectionExtensions.cs
+++ b/Jellyfin.Server.Implementations/Extensions/ServiceCollectionExtensions.cs
@@ -88,7 +88,7 @@ public static class ServiceCollectionExtensions
 
         serviceCollection.AddSingleton(typeof(ICacheManager<>), typeof(BaseCacheManager<>));
         serviceCollection.AddSingleton(new CacheConfigurationBuilder()
-                                        .WithJsonSerializer()
+                                        .WithBondCompactBinarySerializer()
                                         .WithMicrosoftMemoryCacheHandle(instanceName: "MemoryCache")
                                         .Build());
 
--- a/Jellyfin.Server.Implementations/Extensions/ServiceCollectionExtensions.cs
+++ b/Jellyfin.Server.Implementations/Extensions/ServiceCollectionExtensions.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using CacheManager.Core;
 using EFCoreSecondLevelCacheInterceptor;
 using Jellyfin.Database.Implementations;
 using Jellyfin.Database.Implementations.DbConfiguration;
@@ -80,10 +81,16 @@ public static class ServiceCollectionExtensions
         IConfiguration configuration)
     {
         serviceCollection.AddEFSecondLevelCache(options =>
-            options.UseMemoryCacheProvider()
+            options.UseCacheManagerCoreProvider()
                 .ConfigureLogging(true)
                 .UseCacheKeyPrefix("EF_")
                 .UseDbCallsIfCachingProviderIsDown(TimeSpan.FromMinutes(1)));
+
+        serviceCollection.AddSingleton(typeof(ICacheManager<>), typeof(BaseCacheManager<>));
+        serviceCollection.AddSingleton(new CacheConfigurationBuilder()
+                                        .WithJsonSerializer()
+                                        .WithMicrosoftMemoryCacheHandle(instanceName: "MemoryCache")
+                                        .Build());
 
         var efCoreConfiguration = configurationManager.GetConfiguration<DatabaseConfigurationOptions>("database");
         JellyfinDbProviderFactory? providerFactory = null;
--- a/Jellyfin.Server.Implementations/Extensions/ServiceCollectionExtensions.cs
+++ b/Jellyfin.Server.Implementations/Extensions/ServiceCollectionExtensions.cs
@@ -159,7 +159,6 @@ public static class ServiceCollectionExtensions
             opt.AddInterceptors(serviceProvider.GetRequiredService<JellyfinSecondLevelCacheInterceptor>());
             var lockingBehavior = serviceProvider.GetRequiredService<IEntityFrameworkCoreLockingBehavior>();
             lockingBehavior.Initialise(opt);
-            opt.AddInterceptors(serviceProvider.GetRequiredService<SecondLevelCacheInterceptor>());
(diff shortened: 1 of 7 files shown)


Jellyfin.Server.Implementations/Jellyfin.Server.Implementations.csproj
```diff
@@ -31,6 +31,14 @@
     <PackageReference Include="Microsoft.EntityFrameworkCore.Relational" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="CacheManager.Core" />
+    <PackageReference Include="CacheManager.Microsoft.Extensions.Caching.Memory" />
+    <PackageReference Include="CacheManager.Serialization.Bond" />
+    <PackageReference Include="EFCoreSecondLevelCacheInterceptor.CacheManager.Core" />
+    <PackageReference Include="Newtonsoft.Json" />
```
EraYaN: Why does this add Newtonsoft.Json? Shouldn't that show up as a transient package? I don't see any new using statements using it specifically.

Shadowghost: We need to add it explicitly because otherwise a version is pulled in which in turn pulls in incompatible versions of packages we are using elsewhere

EraYaN (approved):

cvium (changes requested):
The issue wasn't that we weren't fully in EFCore. The issue is the locking pattern in [EFCoreSecondLevelCacheInterceptor](https://github.com/VahidN/EFCoreSecondLevelCacheInterceptor) is super aggressive for reads _and_ writes, so ramping up the concurrency is bound to cause threadpool starvation, which becomes a _big_ problem for bad hardware

EraYaN:
Now that I read into it the dev seems to be rather averse to making any changes in general, also like when cache providers are unavailable, not transparently falling back to db which would also be a problem for Jellyfin. 

Also is there a reason why "UseMemoryCache" on the DbContextBuilder is not enough? We are single process anyway, so no need for redis or something similar. A shame there is no built in for the new HybridCache but alas. But maybe I'm misunderstanding what that extension method sets up exactly, seem to just set the internal _memoryCache variable in the options.

crobibero:
Instead of using MemoryCache it might make sense to implement our own IEFCacheServiceProvider based on Bitfaster.Caching

Shadowghost:
@crobibero possible sure, but doesn't fix the issue @cvium pointed out. Not sure though how what to do now.

EraYaN:
I wonder if they removed the `ReaderWriterLock ` based locking because it was slower in non-contested scenarios? Granted, most of Jellyfin is mostly readonly, so `ReaderWriterLock` might be sufficient. Although, some cache implementations are thread safe. Removes most of the locking and "many connections" problems. Then again a single ReaderWriterLock will do 600k/s locks on a single 10th gen intel core. (some 32 writers writing 10 times a second the rest readers). Which is probably just fine for most Jellyfin installs.

Note the critical section here was writing to a Dictionary so MemoryCache is probably much slower. We might just need to test it though.

crobibero:
We should probably disable caching during startup / migrations, and we also need to handle purging the cache when `ExecuteDelete` is called as it doesn't flow through the change tracker

Jellyfin.Server.Implementations/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -78,6 +80,18 @@ public static IServiceCollection AddJellyfinDbContext(
         IServerConfigurationManager configurationManager,
         IConfiguration configuration)
     {
+        serviceCollection.AddEFSecondLevelCache(options =>
+            options.UseCacheManagerCoreProvider()
+                .ConfigureLogging(true)
+                .UseCacheKeyPrefix("EF_")
+                .UseDbCallsIfCachingProviderIsDown(TimeSpan.FromMinutes(1)));
```
IDisposable: One minute seems really long to me...

src/Jellyfin.Database/Jellyfin.Database.Implementations/Cache/JellyfinSecondLevelCacheInterceptor.cs
```diff
@@ -0,0 +1,290 @@
...
+    {
+        using (DbLock.EnterRead(_logger))
+        {
+            return ValueTask.FromResult(_processor.ProcessExecutingCommands(command, eventData?.Context, result));
+        }
+    }
+
+#pragma warning disable MT1013 // Releasing lock without guarantee of execution
+#pragma warning disable MT1012 // Acquiring lock without guarantee of releasing
+    private sealed class DbLock : IDisposable
```
JPVenson, line 185: maybe its better to refactor the existing one so it can be reused instead

crobibero: I copied it with the assumption that the locking behavior would be removed since it's locking above, but of course this would need full testing 😃

JPVenson: hmm ... just as a thought, how about we make this optional via a parameter? Then i would be happier to add it now, otherwise as you said, full-testing required but if its optional and non-default i would rest easier

sjorge: +1 to making this opt-in, this can always be changed to opt-out based on real world feedback / more testing very much like a bunch of the tweaks mentioned in https://github.com/jellyfin/jellyfin/pull/14450 the user could then experiment or try things based on feedback.

webysther:
Some points about the cache:
- It will be helpful even if the provider writes to disk, as it reduces the database I/O after caching. I haven’t seen a disk-based provider, but it’s worth mentioning that even if the system starts using disk through SWAP, there will still be a benefit.
- Invalidation needs to be possible, even if only via command line—ideally, it should also be available as a manual trigger in the UI.
- I think a factory pattern could be used for the providers, allowing the choice to be made via an environment variable.
- It should be possible to configure the cache as read-only.

PR #14421, opened 2025-07-04
Avoid creating duplicate folder in a library

gzsombor:
**Changes**
Currently it is possible that the same folder added multiple times to the same library, as there are no verification before adding a path. With this change when a path is already in the library, adding that new one is skipped.
   
**Issues**
A folder can be added multiple times to a library, which is highly confusing.


crobibero:
This is a good start, but nested folders is also unsupported 

Library 1:
`/path/to/media`

Library 2:
`/path/to/media/more`

gzsombor:
> This is a good start, but nested folders is also unsupported
> 
> Library 1: `/path/to/media`
> 
> Library 2: `/path/to/media/more`

Thanks, that is true, I've just fixed this problem.

crobibero:
I just tested this and this fails when creating a brand new library since the library is already added in `LibraryManager.AddVirtualFolder`

It is also possible to add nested paths all at once, which this PR doesn't prevent

Bond-009:
Closing as stale

theguymadmax:
Note: the function of the original PR was also implemented in the web repo: https://github.com/jellyfin/jellyfin-web/pull/7469

PR #14422, opened 2025-07-04
Add jellyfin cli to handle at least initial wizard setup

ibizaman:
**Changes**
This PR adds a command-line tool which allows the user to go through the initial setup (the wizard) using the command line. This is useful because it allows declarative setups to work with Jellyfin.

The implementation directly taps into Jellyfin server libraries which makes it less brittle than existing CLI tools. Also, for going through the initial setup, the API is not used since it requires creating an API key which is not possible without a user.

One caveat: the Jellyfin server must be started before running the cli and must also be restarted afterwards. I do not plan to fix the restarting issue in this PR. A solution could be sending a `SIGUSR1` to the server.

I plan to add commands in subsequent PRs to manager users, trigger a library scan, manage plugins and others I don't think about now. I can add those in this PR if required.

**Usage**

```bash
$ jellyfin-cli --help

  wizard     Run the server's initial setup.

  help       Display more information on a specific command.

  version    Display version information.
```

```bash
$ jellyfin-cli wizard --help

  --write                                (Default: false) Write configuration.

  --datadir                              Path to use for the data folder (database files, etc.).

  --configdir                            Path to use for configuration data (user settings and pictures).

  -C, --cachedir                         Path to use for caching.

  -l, --logdir                           Path to use for writing log files.

  --username                             Username of the initial admin user.

  --password-file                        Path to a file containing the initial admin user password.

  --preferred-display-language           Preferred display language.

  --preferred-metadata-language          Preferred metadata language.

  --preferred-metadata-country-region    Preferred metadata country/region.

  --enable-remote-access                 Enable remote connections to the server. Not needed if behind a reverse proxy.

  --enable-automatic-port-mapping        Enable automatic port mapping.

  --help                                 Display this help screen.

  --version                              Display version information.
```

```
$ mkdir $PWD/tmp

$ jellyfin --nowebclient --datadir $PWD/tmp
[ ... ]

$ echo admin > ./tmp/mypassword
```

```
$ jellyfin-cli wizard --datadir $PWD/tmp/ --password-file ./tmp/mypassword
Current Configuration:

  Username:                        jellyfin
  Password:                        
  PreferredDisplayLanguage:        en-US
  PreferredMetadataCountryRegion:  US
  PreferredMetadataLanguage:       en
  EnableRemoteAccess:              True
  EnableAutomaticPortMapping:      False

New Configuration:

  Username:                        jellyfin
  Password:                        <new password redacted>
  PreferredDisplayLanguage:        en-US
  PreferredMetadataCountryRegion:  US
  PreferredMetadataLanguage:       en
  EnableRemoteAccess:              True
  EnableAutomaticPortMapping:      False

Use --write to save config.
```

```
$ jellyfin-cli wizard --datadir $PWD/tmp/ --password-file ./tmp/mypassword --write

Current Configuration:

  Username:                        jellyfin
  Password:                        
  PreferredDisplayLanguage:        en-US
  PreferredMetadataCountryRegion:  US
  PreferredMetadataLanguage:       en
  EnableRemoteAccess:              True
  EnableAutomaticPortMapping:      False

New Configuration:

  Username:                        jellyfin
  Password:                        <new password redacted>
  PreferredDisplayLanguage:        en-US
  PreferredMetadataCountryRegion:  US
  PreferredMetadataLanguage:       en
  EnableRemoteAccess:              True
  EnableAutomaticPortMapping:      False

The new configuration has been written. Jellyfin server must be restarted to take it into account.
```

```
$ jellyfin-cli wizard --datadir $PWD/tmp/ --password-file ./tmp/mypassword

Current Configuration:

  Username:                        timi
  Password:                        $PBKDF2-SHA512$iterations=210000$5788376199C0F7F533E1F93E55DAB253$B54D9484DBA1081860BF16E2BDCB662CE90BE96EC9E8AC26A7D0E128A07668CCE45ECD842E1660279AE9E281DD61AB4E9D7F7676513A1E0CF2B6B6EBE5425950
  PreferredDisplayLanguage:        en-US
  PreferredMetadataCountryRegion:  US
  PreferredMetadataLanguage:       en
  EnableRemoteAccess:              True
  EnableAutomaticPortMapping:      False

New Configuration:

  Username:                        timi
  Password:                        <new password redacted>
  PreferredDisplayLanguage:        en-US
  PreferredMetadataCountryRegion:  US
  PreferredMetadataLanguage:       en
  EnableRemoteAccess:              True
  EnableAutomaticPortMapping:      False

Use --write to save config.
```

**Issues**

No issues but other cli implementations exist. They all suffer from the same issue in my opinion which is they are implemented outside of the Jellyfin codebase which makes them brittle and require reverse engineering what the Jellyfin server is doing.

This implementation relies directly on the Jellyfin server libraries which makes it much less brittle and easier to extend.

Prior work:
- https://github.com/LSchallot/JellyRoller (Rust)
- https://github.com/Smiley-McSmiles/jellyman (Shell)
- https://pypi.org/project/Jellyfin-CLI (Python)
- https://gist.github.com/nwithan8/34c36c615a347dd576e0a8d157f69fc0 (Shell)
- https://github.com/Sveske-Juice/declarative-jellyfin (Nix/Shell)

Questions about cli:
- https://www.reddit.com/r/jellyfin/comments/gy27i6/how_do_i_trigger_a_library_scan_from_the_command/
- https://forum.jellyfin.org/t-how-to-enable-ipv6-on-command-line?highlight=command%2Bline

Btw I tried to search the forum for "cli" but that returns 200 pages because it includes results like "client". I can't find a way to just search for a word.

### DIFF
--- a/Emby.Server.Implementations/ApplicationHost.cs
+++ b/Emby.Server.Implementations/ApplicationHost.cs
@@ -692,7 +692,7 @@ namespace Emby.Server.Implementations
         /// <summary>
         /// Discovers the types.
         /// </summary>
-        protected void DiscoverTypes()
+        public void DiscoverTypes()
         {
             Logger.LogInformation("Loading assemblies");
 
--- a/Jellyfin.Cli/Jellyfin.Cli.csproj
+++ b/Jellyfin.Cli/Jellyfin.Cli.csproj
@@ -1,0 +1,40 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <AssemblyName>jellyfin-cli</AssemblyName>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net9.0</TargetFramework>
+    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="..\SharedVersion.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="CommandLineParser" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Emby.Server.Implementations\Emby.Server.Implementations.csproj" />
+    <ProjectReference Include="..\Jellyfin.Api\Jellyfin.Api.csproj" />
+    <ProjectReference Include="..\Jellyfin.Server\Jellyfin.Server.csproj" />
+  </ItemGroup>
+
+  <!-- Code Analyzers -->
+  <ItemGroup Condition=" '$(Configuration)' == 'Debug' ">
+    <PackageReference Include="IDisposableAnalyzers">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
+    </PackageReference>
+    <PackageReference Include="Microsoft.CodeAnalysis.BannedApiAnalyzers">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
+    </PackageReference>
+    <PackageReference Include="SerilogAnalyzer" PrivateAssets="All" />
+    <PackageReference Include="StyleCop.Analyzers" PrivateAssets="All" />
+    <PackageReference Include="SmartAnalyzers.MultithreadingAnalyzer" PrivateAssets="All" />
+  </ItemGroup>
+
+</Project>
--- a/Jellyfin.Cli/Options/Wizard.cs
+++ b/Jellyfin.Cli/Options/Wizard.cs
@@ -1,0 +1,120 @@
+using System.IO;
+using CommandLine;
+
+namespace Jellyfin.Cli.Options
+{
+    /// <summary>
+    /// Class used by CommandLine package when parsing the wizard setup command line arguments.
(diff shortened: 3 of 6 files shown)


ferferga:
Cool, thank you very much for doing this, it's much needed imo!

Before server people chime in, I think it would be best if this is integrated into Jellyfin itself, so it doesn't depend on checking if the server is up and running.

ibizaman:
Thanks! I’ve definitely been scratching my own itch with this one :)

I’m not sure of what you mean with not being separate from the server. Do you mean not having a separate binary and instead having a “config” subcommand where this could go into?

ferferga:
@ibizaman Yes, exactly

ibizaman:
I’m totally fine making the change but should I wait on someone else to confirm ? I just don’t want to put in the work just to revert it 😁 nor do I have the background in jellyfin dev to make my own opinion

nielsvanvelzen:
Yes, changing the current binary (jellyfin.exe/jellyfin.dll) to be a CLI would be great, if there is no subcommand specified it should default to current behavior (start server). I think it would be best to separate the changes into multiple pull requests. One to add the CLI infrastructure without any subcommands and one PR for each subcommand group (e.g. commands related to each other).

I also believe we should use [System.CommandLine](https://github.com/dotnet/command-line-api) which will become stable in .NET 10 (which Jellyfin 10.12 is pretty much guaranteed to use).

ibizaman:
@nielsvanvelzen that works for me! I like splitting the PR that way.

JPVenson:
that 10.11.6-1 branch was created by accident and deleting that branch also for some reason closed your Pr. Please target the correct branch 10.11.X and reopen

PR #14427, opened 2025-07-05
make jellyfin accept more video file extensions

Fjox:
**Changes**
Jellyfin should now recognise ".dat",".m1v",".m2p",".m4r",".m4p",".mka",".mks",".movie",".mp1",".mp2",".mj2",".mj2p",".mpe",".mpv",".mts",".ps",".qt",".tsa",".tsv" as valid fileformats for video.


gnattu:
Most suffixes you added are not intended for video but for something else like audio and subtitles. P

PR #14428, opened 2025-07-05
Make jellyfin accept mpeg-1(.mp1), video cd(.dat) and quicktime(.qt) file extensions as videos

Fjox:
**Changes**
Jellyfin should now recognise files with ".dat",".m1v",".movie",".mp1",".mp2",".mpe",".qt" as videos.


gnattu:
mp1 and mp2 are audio files, and dat is not necessarily a video file. I know some old VCD file would have files in dat suffix but such dat are usually use mpeg files which is the same thing as .mpg. .dat is not a safe extension to be assumed as video files as it is too generic.

PR #14437, opened 2025-07-08
Added advanced parallel process calculation

JPVenson:
Adds new calculation to always leave some headroom for other system resources on scan when the auto setting is set

requires adaption of the translation resources.

### DIFF
--- a/MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
+++ b/MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
@@ -96,20 +96,20 @@ public sealed class LimitedConcurrencyLibraryScheduler : ILimitedConcurrencyLibr
         }
     }
 
-    private bool ShouldForceSequencialOperation()
+    private bool ShouldForceSequentialOperation()
     {
-        // if the user either set the setting to 1 or its unset and we we have less then 4 cores its better to run sequencially.
+        // if the user either set the setting to 1 or it's unset and we have fewer than 4 cores it's better to run sequentially.
         var fanoutSetting = _serverConfigurationManager.Configuration.LibraryScanFanoutConcurrency;
-        return fanoutSetting is 1 || (fanoutSetting <= 0 && Environment.ProcessorCount <= 3);
+        return fanoutSetting == 1 || (fanoutSetting <= 0 && Environment.ProcessorCount <= 3);
     }
 
     private int CalculateScanConcurrencyLimit()
     {
-        // when this is invoked, we already checked ShouldForceSequencialOperation for the sequencial check.
+        // when this is invoked, we already checked ShouldForceSequentialOperation for the sequential check.
         var fanoutConcurrency = _serverConfigurationManager.Configuration.LibraryScanFanoutConcurrency;
         if (fanoutConcurrency <= 0)
         {
-            // in case the user did not set a limit manually, we can assume he has 3 or more cores as already checked by ShouldForceSequencialOperation.
+            // in case the user did not set a limit manually, we can assume he has 3 or more cores as already checked by ShouldForceSequentialOperation.
             return Environment.ProcessorCount - 3;
         }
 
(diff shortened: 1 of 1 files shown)


MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
```diff
@@ -96,14 +96,33 @@ async Task RunCleanupTask()
         }
     }
 
+    private bool ShouldForceSequencialOperation()
+    {
+        // if the user either set the setting to 1 or its unset and we we have less then 4 cores its better to run sequencially.
```
crobibero:
```suggestion
        // if the user either set the setting to 1 or it's unset and we have fewer than 4 cores it's better to run sequentially.
```

MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
```diff
@@ -96,14 +96,33 @@ async Task RunCleanupTask()
         }
     }
 
+    private bool ShouldForceSequencialOperation()
```
crobibero:
```suggestion
    private bool ShouldForceSequentialOperation()
```

MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
```diff
@@ -96,14 +96,33 @@ async Task RunCleanupTask()
...
+    {
+        // if the user either set the setting to 1 or its unset and we we have less then 4 cores its better to run sequencially.
+        var fanoutSetting = _serverConfigurationManager.Configuration.LibraryScanFanoutConcurrency;
+        return fanoutSetting is 1 || (fanoutSetting <= 0 && Environment.ProcessorCount <= 3);
```
crobibero:
```suggestion
        return fanoutSetting == 1 || (fanoutSetting <= 0 && Environment.ProcessorCount <= 3);
```

Compiles to the same thing but I think explicit is better

MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
```diff
@@ -96,14 +96,33 @@ async Task RunCleanupTask()
...
+    }
+
+    private int CalculateScanConcurrencyLimit()
+    {
+        // when this is invoked, we already checked ShouldForceSequencialOperation for the sequencial check.
```
crobibero:
```suggestion
        // when this is invoked, we already checked ShouldForceSequentialOperation for the sequential check.
```

MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
```diff
@@ -96,14 +96,33 @@ async Task RunCleanupTask()
...
+        var fanoutConcurrency = _serverConfigurationManager.Configuration.LibraryScanFanoutConcurrency;
+        if (fanoutConcurrency <= 0)
+        {
+            // in case the user did not set a limit manually, we can assume he has 3 or more cores as already checked by ShouldForceSequencialOperation.
```
crobibero:
```suggestion
            // in case the user did not set a limit manually, we can assume he has 3 or more cores as already checked by ShouldForceSequentialOperation.
```

MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
```diff
@@ -223,7 +242,7 @@ void UpdateProgress()
             };
         }).ToArray();
 
-        if (_serverConfigurationManager.Configuration.LibraryScanFanoutConcurrency == 1)
+        if (ShouldForceSequencialOperation())
```
crobibero:
```suggestion
        if (ShouldForceSequentialOperation())
```

Issue #14434, opened 2025-07-08
Jellyfin 10.11/master regression on SQLite database: does not enable WAL mode


evan314159:
### Description of the bug

Jellyfin 10.11/master regression on SQLite database configuration: does enable WAL mode.

Consequences:
- poorer performance -- Jellyfin without WAL performs very visibly poorer.
- if WAL is enabled by the user, the journal_size_limit is left set to default -1 (unlimited) leading to unbounded WAL -- I stopped my test container at 80GB WAL (I think there is another issue that prevents WAL truncation, maybe a read transaction that never closes, but I have not found it yet to raise an issue).

Evidence: from brand new deployment of -master from a few days ago:

$ sqlite3 jellyfin.db
sqlite> PRAGMA journal_mode;
delete

### Reproduction steps

See above + diff the database code...

### What is the current _bug_ behavior?

journal_mode=WAL not set in master, and WAL grew to >80GB in testing.   Both are regressions from 10.10.

### What is the expected _correct_ behavior?

Jellyfin should enable WAL as per 10.10:
- PRAGMA journal_mode=WAL
- PRAGMA journal_size_limit=128MB, or some other reasonable value.

### Jellyfin Server version

Master

### Specify the build version

6d4169a4495a6206a4b56a484843fd87b0926c48

### Environment

```markdown
OS: TrueNAS 25.04.1 (Linux)
Linux kernel: 6.12.15
Docker.io
Container built using build.py from master
```

### Jellyfin logs

```shell
N/A
```

### FFmpeg logs

```shell

```

bruceharrison1984:
I verified this on the official RC2 docker image that was a completely "fresh" install.

```bash
docker exec -it jellyfin bash
sqlite3 jellyfin.db
sqlite> PRAGMA journal_mode;
delete
```

WAL mode doesn't appear to be enabled. I actually started looking into this earlier this week, but stopped once I saw that EF *should* enable WAL by default now. Glad you kept digging. 

In my research I did find similar discussions around pooling not unlocking the DB, however it looked like a slightly custom DbContext was able to circumvent the issue by closing all connections once they were deemed "dead". ~Though surely I'll never find the thread again now that I need it.~ [Found a very simple example](https://stackoverflow.com/a/75517185)

The loss of pooling has caused significant performance issues for me as well, with a QSV server frequently hovering around 1% utilization coming to a screeching halt whenever a scan takes place. this includes when using either optimistic or pessimistic settings for concurrency.

[...]

Issue #14366, opened 2025-06-22
Playback and UI Lag During Full Library Scan


iPleyGit:
### Description of the bug

When a full library scan is in progress, playback stutters on any device while the scan is active. Once it’s completed or stopped, content playback returns to normal.
I’ve also noticed that navigating the interface—especially in the web browser—is significantly slower than usual.

### Reproduction steps

```
[2025-06-22 07:26:17.353 +00:00] [ERR] [30] MediaBrowser.Controller.LibraryTaskScheduler.LimitedConcurrencyLibraryScheduler: Error while performing a library operation
Microsoft.Data.Sqlite.SqliteException (0x80004005): SQLite Error 5: 'database is locked'.
   at Microsoft.Data.Sqlite.SqliteException.ThrowExceptionForRC(Int32 rc, sqlite3 db)
   at Microsoft.Data.Sqlite.SqliteDataReader.NextResult()
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteReader(CommandBehavior behavior)
   at Microsoft.Data.Sqlite.SqliteConnectionExtensions.ExecuteNonQuery(SqliteConnection connection, String commandText, SqliteParameter[] parameters)
   at Microsoft.Data.Sqlite.SqliteConnection.BeginTransaction(IsolationLevel isolationLevel, Boolean deferred)
   at Microsoft.EntityFrameworkCore.Storage.RelationalConnection.BeginTransaction(IsolationLevel isolationLevel)
   at Jellyfin.Server.Implementations.Item.BaseItemRepository.UpdateOrInsertItems(IReadOnlyList`1 items, CancellationToken cancellationToken)
   at Jellyfin.Server.Implementations.Item.BaseItemRepository.SaveItems(IReadOnlyList`1 items, CancellationToken cancellationToken)
   at Emby.Server.Implementations.Library.LibraryManager.CreateItems(IReadOnlyList`1 items, BaseItem parent, CancellationToken cancellationToken)
   at Emby.Server.Implementations.Library.LibraryManager.CreateItem(BaseItem item, BaseItem parent)
   at MediaBrowser.Controller.Entities.Folder.AddChild(BaseItem item)
   at MediaBrowser.Providers.TV.SeriesMetadataService.CreateSeasonAsync(Series series, String seasonName, Nullable`1 seasonNumber, CancellationToken cancellationToken)
[...]

Issue #3372, opened 2020-06-18
Playback image size for some some video files through FireTV app is ~25% of TV resolution so image is very small


gregvolk:
**Screenshots**
Attaching two screenshots that show the problem.
![20200618_103736_lr](https://user-images.githubusercontent.com/13316371/85045469-9d246a00-b154-11ea-9dbe-0fedb04cd473.jpg)
![20200618_103740_lr](https://user-images.githubusercontent.com/13316371/85045485-a1e91e00-b154-11ea-9fbd-1c3022737ba7.jpg)

**Media Info of the file**
General
Complete name                            : s01e01.SD.DVD.avi
Format                                   : AVI
Format/Info                              : Audio Video Interleave
File size                                : 25.8 MiB
Duration                                 : 5 min 5 s
Overall bit rate mode                    : Variable
Overall bit rate                         : 708 kb/s
Writing application                      : ***
Writing library                          : ***

Video
ID                                       : 0
Format                                   : MPEG-4 Visual
Format settings                          : BVOP1
Format settings, BVOP                    : 1
Format settings, QPel                    : No
Format settings, GMC                     : No warppoints
Format settings, Matrix                  : Default (H.263)
Muxing mode                              : Packed bitstream
Codec ID                                 : DX50
Codec ID/Hint                            : DivX 5
Duration                                 : 5 min 5 s
Bit rate                                 : 596 kb/s
Nominal bit rate                         : 4 000 kb/s
Width                                    : 512 pixels
Height                                   : 416 pixels
Display aspect ratio                     : 5:4
Frame rate                               : 25.000 FPS
Color space                              : YUV
Chroma subsampling                       : 4:2:0
Bit depth                                : 8 bits
Scan type                                : Progressive
Compression mode                         : Lossy
Bits/(Pixel*Frame)                       : 0.112
Stream size                              : 21.7 MiB (84%)
Writing library                          : DivX 5.2.0 (DrDivX 105) (UTC 2004-07-17)

Audio
ID                                       : 1
Format                                   : MPEG Audio
Format version                           : Version 1
Format profile                           : Layer 3
Format settings                          : Joint stereo / MS Stereo
Codec ID                                 : 55
Codec ID/Hint                            : MP3
Duration                                 : 5 min 4 s
Bit rate mode                            : Variable
Bit rate                                 : 96.0 kb/s
Channel(s)                               : 2 channels
Sampling rate                            : 44.1 kHz
Compression mode                         : Lossy
[...]

PR #14450, opened 2025-07-10
SQLite set database options and PRAGMAs via database.xml and apply on every database connection

evan314159:
**Changes**
Enable SQLite database options and PRAGMAs to be configured via database.xml, and apply on every connection.  Most SQLite pragmas are per-connection.

Will enable Jellyfin team and advanced users to tune performance further without recompiling Jellyfin.

New configurable database options:
- Database Directory: defaults to DataPath, enables SQLite databases to be on separate storage including SSD or zfs datasets with database-tuned settings

New configurable SQLite options:
- Connection pooling: defaults to false (from 10.11)
- Journal mode: defaults to WAL (from 10.10, and what was expected of 10.11 but in practice does not work)
- Journal size limit: 128MB (from 10.10) -- NOTE: this is not the WAL maximum size, it is the size at which SQLite will allow an inactive WAL to remain in preparation for further writes
- Cache size: SQLite default (2MB, from 10.10), uses SQLite semantics that a positive number is expressed in pages, negative number in kB (-65536 = 64MB).
- MMAP size: SQLite default (0, not previously available)

Also sets PRAGMA temp_store = 2 (MEMORY).  Temporary tables are used for operations like ORDER BY, GROUP BY, DISTINCT not satisfied by an index.

Settings to be applied are logged once at start at information level, and on every connection at debug level.

Example complete database.xml (note: these are not good settings!  do not use these! read the SQLite documentation, and understand what the settings are likely to do in your environment!):
```
<?xml version="1.0" encoding="utf-8"?>
<DatabaseConfigurationOptions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XM LSchema">
  <DatabaseType>Jellyfin-SQLite</DatabaseType>
  <LockingBehavior>NoLock</LockingBehavior>
  <DatabaseDirectory>/sqlite_data</DatabaseDirectory>
  <SqliteOptions>
    <Pooling>false</Pooling>
    <JournalMode>WAL</JournalMode>
    <JournalSizeLimit>1073741824</JournalSizeLimit>
    <CacheSize>-65536</CacheSize>
    <MmapSize>2147483648</MmapSize>
  </SqliteOptions>
</DatabaseConfigurationOptions>
```
Testing completed:
Test 1: 10.11-master new library without this change: poor performance during and after as per #14366

Test 2: 10.11-master new library with this change: scan completed in ~7 hours, excellent performance during and after

Test 3: 10.10 upgrade to 10.11 with change: upgrade completed in ~1 minute, library scan ~5 min, excellent performance during and after

Library consists of over 40,000 tracks, almost 2500 albums, and a few hundred videos.  Server is: Dell T330, 4CPU/8 thread, 64GB RAM, media and database on 2x spinning disk in ZFS mirror.

Notes:
- AI assisted change. I designed the change and reveiewed every line but I am not an expert in Jellyfin or EFCore. I appreciate the time reviewers will take to review the change and welcome feedback.
- On first start after replacing the 10.10 container with 10.11-master+changes the database was marked read-only. I found it had journal_mode = DELETE, when 10.10 was set to WAL.  After resetting to WAL Jellyfin restarted and the migration proceeded as expected.
- If you make an error in database.xml, rather than fail EFCore will replace it with a clean default. Keep a copy of your database.xml if making changes!

**Issues**
Addresses performance/usability issues post-upgrade to 10.11-RC raised in #14366 and #14434, and feature #3372.

### DIFF
--- a/src/Jellyfin.Database/Jellyfin.Database.Implementations/DbConfiguration/DatabaseConfigurationOptions.cs
+++ b/src/Jellyfin.Database/Jellyfin.Database.Implementations/DbConfiguration/DatabaseConfigurationOptions.cs
@@ -22,4 +22,14 @@ public class DatabaseConfigurationOptions
     /// Defaults to "NoLock".
     /// </summary>
     public DatabaseLockingBehaviorTypes LockingBehavior { get; set; }
+
+    /// <summary>
+    /// Gets or Sets the database directory path. If not set, defaults to DataPath.
+    /// </summary>
+    public string? DatabaseDirectory { get; set; }
+
+    /// <summary>
+    /// Gets or Sets the SQLite-specific options.
+    /// </summary>
+    public SqliteOptions? SqliteOptions { get; set; }
 }
--- a/src/Jellyfin.Database/Jellyfin.Database.Implementations/DbConfiguration/DatabaseConfigurationOptions.cs
+++ b/src/Jellyfin.Database/Jellyfin.Database.Implementations/DbConfiguration/DatabaseConfigurationOptions.cs
@@ -24,11 +24,6 @@ public class DatabaseConfigurationOptions
     public DatabaseLockingBehaviorTypes LockingBehavior { get; set; }
 
     /// <summary>
-    /// Gets or Sets the database directory path. If not set, defaults to DataPath.
-    /// </summary>
-    public string? DatabaseDirectory { get; set; }
-
-    /// <summary>
     /// Gets or Sets the SQLite-specific options.
     /// </summary>
     public SqliteOptions? SqliteOptions { get; set; }
--- a/src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteDatabaseProvider.cs
+++ b/src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteDatabaseProvider.cs
@@ -82,10 +82,27 @@ public sealed class SqliteDatabaseProvider : IJellyfinDatabaseProvider
 
         // Log per-connection SQLite PRAGMAs set through SqlitePragmaInterceptor once at startup
         _logger.LogInformation("Set SQLite PRAGMA journal_mode = {JournalMode}", _databaseConfig.SqliteOptions.JournalMode);
-        _logger.LogInformation("Set SQLite PRAGMA journal_size_limit = {JournalSizeLimit}", _databaseConfig.SqliteOptions.JournalSizeLimit);
-        _logger.LogInformation("Set SQLite PRAGMA synchronous = {Synchronous}", _databaseConfig.SqliteOptions.Synchronous);
-        _logger.LogInformation("Set SQLite PRAGMA cache_size = {CacheSize}", _databaseConfig.SqliteOptions.CacheSize);
-        _logger.LogInformation("Set SQLite PRAGMA mmap_size = {MmapSize}", _databaseConfig.SqliteOptions.MmapSize);
+
+        if (_databaseConfig.SqliteOptions.JournalSizeLimit != SqliteOptions.DefaultJournalSizeLimit)
+        {
+            _logger.LogInformation("Set SQLite PRAGMA journal_size_limit = {JournalSizeLimit}", _databaseConfig.SqliteOptions.JournalSizeLimit);
+        }
+
+        if (_databaseConfig.SqliteOptions.Synchronous != SqliteOptions.DefaultSynchronous)
+        {
+            _logger.LogInformation("Set SQLite PRAGMA synchronous = {Synchronous}", _databaseConfig.SqliteOptions.Synchronous);
+        }
+
+        if (_databaseConfig.SqliteOptions.CacheSize != SqliteOptions.DefaultCacheSize)
+        {
+            _logger.LogInformation("Set SQLite PRAGMA cache_size = {CacheSize}", _databaseConfig.SqliteOptions.CacheSize);
+        }
+
+        if (_databaseConfig.SqliteOptions.MmapSize != SqliteOptions.DefaultMmapSize)
+        {
+            _logger.LogInformation("Set SQLite PRAGMA mmap_size = {MmapSize}", _databaseConfig.SqliteOptions.MmapSize);
+        }
+
         _logger.LogInformation("Set SQLite PRAGMA temp_store = MEMORY");
 
         options
--- a/src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteDatabaseProvider.cs
+++ b/src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteDatabaseProvider.cs
@@ -52,25 +52,25 @@ public sealed class SqliteDatabaseProvider : IJellyfinDatabaseProvider
         var connectionString = $"Filename={databasePath};Pooling={pooling}";
 
         _logger.LogInformation("Opening database {DatabasePath} with pooling={Pooling}", databasePath, pooling);
-        
+
         // Log SQLite configuration once at startup
         var sqliteOptions = _databaseConfig.SqliteOptions;
         var journalMode = sqliteOptions?.JournalMode ?? "WAL";
         var journalSizeLimit = sqliteOptions?.JournalSizeLimit ?? (128 * 1024 * 1024);
-        
+
         _logger.LogInformation("Set SQLite PRAGMA journal_mode = {JournalMode}", journalMode);
         _logger.LogInformation("Set SQLite PRAGMA journal_size_limit = {JournalSizeLimit}", journalSizeLimit);
-        
+
         if (sqliteOptions?.CacheSize.HasValue == true)
         {
             _logger.LogInformation("Set SQLite PRAGMA cache_size = {CacheSize}", sqliteOptions.CacheSize.Value);
         }
-        
+
         if (sqliteOptions?.MmapSize.HasValue == true)
         {
             _logger.LogInformation("Set SQLite PRAGMA mmap_size = {MmapSize}", sqliteOptions.MmapSize.Value);
         }
-        
+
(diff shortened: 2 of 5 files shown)


src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteDatabaseProvider.cs
```diff
@@ -116,7 +151,8 @@ public Task RestoreBackupFast(string key, CancellationToken cancellationToken)
     {
         // ensure there are absolutly no dangling Sqlite connections.
         SqliteConnection.ClearAllPools();
-        var path = Path.Combine(_applicationPaths.DataPath, "jellyfin.db");
+        var databaseDirectory = _databaseConfig.DatabaseDirectory ?? _applicationPaths.DataPath;
```
IDisposable: What if we're doing a migration from 10.10.7 (and before) where the file is known to be on the Data directory (because it couldn't have been elsewhere) and the new config expects it elsewhere? We need to document (at least) that the upgrade to 10.11.x should allow all migrations to complete _before_ allowing the changing of the database directory

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteDatabaseProvider.cs
```diff
@@ -164,4 +200,122 @@ public async Task PurgeDatabase(JellyfinDbContext dbContext, IEnumerable<string>
...
+        private void ApplyPragmaSettings(SqliteConnection connection)
+        {
+            var sqliteOptions = _databaseConfig.SqliteOptions;
+
+            using var command = connection.CreateCommand();
+
+#pragma warning disable CA2100 // Review if the query string passed to 'string SqliteCommand.CommandText' accepts any user input
+            // SQLite PRAGMA processor is not susceptible to SQL injection. It only processes one PRAGMA command and does not allow multiple commands separated by ;.
+            // Set journal mode (default WAL)
+            var journalMode = sqliteOptions?.JournalMode ?? "WAL";
```
IDisposable: For ease of documentation and code modification, we might want to switch to a default singleton of the `sqliteOptions` that has all the _old_ defaults for values set. That way all the configuration defaults are in one place.

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteDatabaseProvider.cs
```diff
@@ -164,4 +200,122 @@ public async Task PurgeDatabase(JellyfinDbContext dbContext, IEnumerable<string>
...
+            command.CommandText = $"PRAGMA journal_mode = {journalMode}";
+            command.ExecuteNonQuery();
+            _logger.LogDebug("Set SQLite PRAGMA journal_mode = {JournalMode}", journalMode);
+
+            // Set journal size limit (default 128MB)
+            var journalSizeLimit = sqliteOptions?.JournalSizeLimit ?? (128 * 1024 * 1024);
```
IDisposable: Alternatively, maybe the absence of a `sqliteOptions` setting should _just not sent the `PRAGMA`?_

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteDatabaseProvider.cs
```diff
@@ -164,4 +200,122 @@ public async Task PurgeDatabase(JellyfinDbContext dbContext, IEnumerable<string>
...
+            // Set temp_store to MEMORY
+            command.CommandText = "PRAGMA temp_store = 2";
+            command.ExecuteNonQuery();
+            _logger.LogDebug("Set SQLite PRAGMA temp_store = 2 (MEMORY)");
+#pragma warning restore CA2100
+        }
+
+        private async Task ApplyPragmaSettingsAsync(SqliteConnection connection, CancellationToken cancellationToken)
+        {
+            var sqliteOptions = _databaseConfig.SqliteOptions;
```
IDisposable: Rather than duplicate all the logic, could we just make the _non `async`_ version be 
```csharp
private void ApplyPragmaSettings(SqliteConnection connection)
{
   await ApplyPragmaSettings(connection);
}
```

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteDatabaseProvider.cs
ghost: database.xml is created as part of the migration.  The situation you are describing could potentially occur if someone went of their way to create a file that doesn't exist prior to upgrading from a version that doesn't support that file...

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteDatabaseProvider.cs
ghost: I can do that.  How would you like to handle options with no default like cache_size and mmap_size -- set to SQLite default (-2000 and 0 respectively)?

I will also move the database directory into SQLite Options.  I had been thinking about generalising it but realised afterwards that the only other database that I hope Jellyfin will adopt soon, Postgres, doesn't need it.

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteDatabaseProvider.cs
ghost: Setting the pragmas is unfortunately important as they don't persist and make a big difference to performance.

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteDatabaseProvider.cs
ghost: I can do that.

I will wait a couple days to see if there is any other feedback and as long as the direction is not conflicting will make the changes and update the PR.

ghost: @IDisposable Thank you very much for your time to review!

crobibero: I think it is fine to point all of the pragma statements to the synchronous implementation. I wouldn't do what was suggested as `async void` is an anti-pattern.

src/Jellyfin.Database/Jellyfin.Database.Implementations/DbConfiguration/DatabaseConfigurationOptions.cs
```diff
@@ -22,4 +22,14 @@ public class DatabaseConfigurationOptions
...
+
+    /// <summary>
+    /// Gets or Sets the database directory path. If not set, defaults to DataPath.
+    /// </summary>
+    public string? DatabaseDirectory { get; set; }
+
+    /// <summary>
+    /// Gets or Sets the SQLite-specific options.
+    /// </summary>
+    public SqliteOptions? SqliteOptions { get; set; }
```
crobibero, line 29: Ideally this would be set using the `CustomDatabaseOptions` that already exists instead of building out a new property

DatabaseDirectory could also be a part of the `CustomDatabaseOptions` collection since it is likely only to be used for SQLite

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteDatabaseProvider.cs
crobibero: I don't think this is a case that we need to consider

src/Jellyfin.Database/Jellyfin.Database.Implementations/DbConfiguration/DatabaseConfigurationOptions.cs
ghost: If it is not decided to merge this into the mainline then I will make available a version that uses CustomDatabaseOptions instead for those who need it.  If it is going into the mainline however I would think it should be part of the main SQLite code.

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteDatabaseProvider.cs
ghost: The code will only send PRAGMAs for SQLite non-default values based on a standard SQLite install, with journal_mode and temp_store always sent.

src/Jellyfin.Database/Jellyfin.Database.Implementations/DbConfiguration/DatabaseConfigurationOptions.cs
joshuaboniface: I would actually second permitting databaseDirectory to be a discrete directory. Storing the SQLite on one filesystem while storing the rest of `data` on another might be a viable usecase for some. Definitely a niche usecase though, esp if we get PGSql into 10.12.

ghost: For a certain scale one would put databases on a separate filesystem (even if on the same storage) so they have their own filesystem metadata locking.  With Postgres the postgres_data and postgres_wal can be put on separate filesystem to have separate metadata locking between them.

With ZFS, I get 3.75:1 compression on the Jellyfin database with zstd-3.  That is 1/3.75 the disk I/O required to read the database -- nice for spinning disk.  However, most of /data is incompressible and compressing all of /data with zstd actually makes the net result larger.

JPVenson:
The underlying issue will be correctly fixed with #14465 and additional options are introduced in the correct way using existing infastructure with #14466

ghost:
Addressed through another change, closing.

PR #14461, opened 2025-07-13
Fix refreshing the library cannot delete old attachments

nyanmisaka:
**Changes**
- Fix refreshing the library cannot delete old attachments

**Issues**
- A user replaced an MKV video from one that included embedded fonts to one that did not, causing SSA/ASS subtitle rendering on the web to fail because the old attachment information was not deleted from the database.

image

### DIFF
--- a/Jellyfin.Server.Implementations/Item/MediaAttachmentRepository.cs
+++ b/Jellyfin.Server.Implementations/Item/MediaAttachmentRepository.cs
@@ -25,8 +25,16 @@ public class MediaAttachmentRepository(IDbContextFactory<JellyfinDbContext> dbPr
     {
         using var context = dbProvider.CreateDbContext();
         using var transaction = context.Database.BeginTransaction();
+
+        // Users may replace a media with a version that includes attachments to one without them.
+        // So when saving attachments is triggered by a library scan, we always unconditionally
+        // clear the old ones, and then add the new ones if given.
         context.AttachmentStreamInfos.Where(e => e.ItemId.Equals(id)).ExecuteDelete();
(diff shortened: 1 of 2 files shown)


Shadowghost (approved):

sjorge:
How does this work with the recently added attachment/subtitle cleanup? Does that not rely on the db having the old paths? (I'm not sure and I can't figure out how it works by looking at the code (not even sure i was looking at the right spot))

Give the error, it seems to me the old attachments have either already be cleared from disk, or not yet. (I think the former, so removing them from the DB is fine?)

nyanmisaka:
> How does this work with the recently added attachment/subtitle cleanup? Does that not rely on the db having the old paths? (I'm not sure and I can't figure out how it works by looking at the code (not even sure i was looking at the right spot))
> 
> Give the error, it seems to me the old attachments have either already be cleared from disk, or not yet. (I think the former, so removing them from the DB is fine?)

`[INF] [79] MediaBrowser.Providers.Movies.MovieMetadataService: File changed, pruning extracted data: /path/to/attachments`

When Jellyfin detects that the video file has changed, or the user manually triggers a media library scan, the old attachment will be deleted, so the scheduled task should not have a chance to delete the attachment in this case.

In this case the attachment had been deleted, but the database still had the old information, which led to the failure of real-time extraction of attachment fonts and web rendering of subtitles. This fixes that mismatch.

JPVenson (approved):

crobibero (approved):

aa1171842415:
最新版本jellyfin ，pve lxc docker部署，扫描媒体库会出现内存泄漏的问题。

PR #14465, opened 2025-07-13
Use IDatabaseCreator instead of relying on History repository

JPVenson:


crobibero (approved):

Jellyfin.Server/Migrations/JellyfinMigrationService.cs
```diff
@@ -105,6 +106,13 @@ public async Task CheckFirstTimeRunOrMigration(IApplicationPaths appPaths)
             var dbContext = await _dbContextFactory.CreateDbContextAsync().ConfigureAwait(false);
             await using (dbContext.ConfigureAwait(false))
             {
+                var databaseCreator = dbContext.Database.GetService<IDatabaseCreator>() as IRelationalDatabaseCreator
```
cvium, line 109: Why this interface over simply calling EnsureCreated?

JPVenson: image

crobibero (approved):

PR #14466, opened 2025-07-13
Allow loading of Database options for DB provider

JPVenson:
Allow DatabaseProvider access to the DatabaseConfigurationOptions from the settings file to better handle custom configuration

For Sqlite its now supported to set the folowing options via the database.xml

```xml
 <CustomProviderOptions>
    <Options>
      <CustomDatabaseOption>
        <Key>Cache</Key>
        <Value>Shared</Value>
      </CustomDatabaseOption>
      <CustomDatabaseOption>
        <Key>Pooling</Key>
        <Value>True</Value>
      </CustomDatabaseOption>
    </Options>
  </CustomProviderOptions>
```


Shadowghost (approved):

crobibero:
Consider allowing the user to set PRAGMAs from config for fine tuning?

Maybe parsing all keys of "PRAGMA:$key"

JPVenson:
i did indeed consider it

crobibero (approved):

PR #14476, opened 2025-07-15
Fix Serbian ISO 639-2 codes

### LINKED ISSUE #14473: 10.11 RC3 Subtitle languange not recognized
### Description of the bug

I went from 10.10.7 to 10.11 RC3, did a full scan as recommended in documentation. It all went well, but there are issues with subtitles. 
I use external subtites following naming scheme movifilename.sr.srt for Serbian subtitles, and they were recognized normally on stable version 10.10.7, but not on 10.11 RC3. I will attach a screenshot. I also use Croatian subtitles, moviefilename.hrv.srt,  in case Serbian are missing, and on 10.10.7 it was enough to set subtitle default language to Serbian and Smart mode for Jellyfin to default to ether Serbian or Croatian external subtitles, depending on what is available. However, on 10.11 RC3, with the same settings it only defaults to Serbian subtitles, so in instances where there is only Croatian external subtitle available, it defaults to off.

### Reproduction steps

I updated to 10.11 RC3, did a full library scan, as recommended.

Subtitles named movie.sr.srt are not recognized as Serbian subtitles.

Subtitles named movie.sr.srt should be recognized as Serbian subtitles, and they were back on version 10.10.7.

### Jellyfin Server version

Unstable

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11 RC3

### Environment

```markdown
- OS: Linux Mint 
- Virtualization: Off
- Plugins: None
```

### Jellyfin logs

```shell
none
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

Image

### Additional information

_No response_

theguymadmax:
The correct ISO 639-2/T for Serbia is srp, scc was deprecated in 2008: https://www.loc.gov/standards/iso639-2/php/code_changes.php

**Issues**
 Fixes #14473

### DIFF
--- a/Emby.Server.Implementations/Localization/iso6392.txt
+++ b/Emby.Server.Implementations/Localization/iso6392.txt
@@ -373,7 +373,7 @@ sam|||Samaritan Aramaic|samaritain
 san||sa|Sanskrit|sanskrit
 sas|||Sasak|sasak
 sat|||Santali|santal
-scc|srp|sr|Serbian|serbe
+srp||sr|Serbian|serbe
 scn|||Sicilian|sicilien
 sco|||Scots|écossais
 sel|||Selkup|selkoupe


crobibero:
I think the order should be swapped instead of removing `scc` since it is in the ISO 639-2/B list

theguymadmax:
All the sources I found had the scc as depreciated, so I don't think we need to list it anymore. 


https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes
<br><br>
"There were originally 22 B codes; scc and scr are now deprecated."<br>
https://en.wikipedia.org/wiki/ISO_639-2
<br><br>
 "ISO 639-2/B code deprecated in favor of ISO 639-2/T code"<br>
https://www.loc.gov/standards/iso639-2/php/code_changes.php

Shadowghost (approved):

PR #14481, opened 2025-07-15
﻿Drawing: make SkiaEncoder more robust when reading image dimensions

queukat:
* NormalizePath is now used inside GetImageSize to handle Unicode paths
* Accept SKCodecResult.InternalError if codec.Info is available
* Skip zero‑byte files early
* Fallback to SKBitmap.DecodeBounds as a last resort
* Downgrade noisy errors to warnings/debug, reducing log spam
* Net effect: no more "Unable to determine image dimensions" flood and ~10‑15 % faster scans on large libraries

**Changes**

- Unified file path handling by applying `NormalizePath` in `GetImageSize` to support Unicode and diacritics across platforms.
- Allowed fallback to `codec.Info` even if `SKCodec.Create` returns `InternalError`, increasing resilience for partially corrupted files.
- Skipped processing of zero-byte files early to avoid unnecessary work and exceptions.
- Added a final fallback using `SKBitmap.DecodeBounds` to estimate image dimensions when all other methods fail.
- Downgraded frequent image loading errors to warnings or debug messages to reduce log noise.

**Issues**


src/Jellyfin.Drawing.Skia/SkiaEncoder.cs
```diff
@@ -202,18 +202,41 @@ public ImageDimensions GetImageSize(string path)
...
+            _logger.LogDebug("Skip zero‑byte image {FilePath}", path);
+            return default;
+        }
+
+        using var codec = SKCodec.Create(safePath, out var result);
+
         switch (result)
         {
             case SKCodecResult.Success:
+            case SKCodecResult.InternalError when codec is not null:
```
JPVenson, line 223: how can that happen?

queukat: Skia/SkiaSharp can return a *non‑null* `SKCodec` **and** `SKCodecResult.InternalError`
when header parsing succeeds but the pixel decode fails.  
Typical triggers:

* truncated JPEG/PNG (missing scan lines or IEND),
* malformed / exotic ICC or EXIF blocks,
* CMYK JPEGs without color‐transform (Skia bug #4139),
* fuzz cases where Skia aborts decode part‑way but has already filled `codec.Info`.

In these situations we only need the *dimensions*, and `codec.Info`
is already populated, so falling through here lets us avoid a hard failure
and the noisy “Unable to determine image dimensions” log‑spam.

Hence the guard `when codec is not null`: we accept **InternalError** only
if Skia still gave us a valid codec/header.

(Happy to add an inline code comment if you feel it would help.)

JPVenson: yea that feels like a workaround for some skia shananigans. Please add an comment for that

queukat: Done

crobibero (approved):

Issue #14472, opened 2025-07-14
[10.11.0] Parental Controls - Allow tags not working


theguymadmax:
### Description of the bug

The allow tag option in parental controls is not functioning correctly. 

### Reproduction steps

1. On 10.11.0 server, set up a Movie and a Series library.
2. Verify that items show up in the recently added section
3. Now add a tag (test) to items that are in the recently added section for movies and series
4. Now set your user account to only be able to view the allowed (test) tag in parental controls.

### What is the current _bug_ behavior?

Libraries:
Allowed tags not working in Movies (library is empty)
Allowed tags not working in Series (library is empty)

Recently added sections:
Allowed tags not working in Movies (section is empty)
Allowed tags work in Series (item shows up and is playable)

### What is the expected _correct_ behavior?

It works

### Jellyfin Server version

Master

### Specify the build version

10.11.0 RC3

### Environment

```markdown
- OS: Windows 11
- Virtualization: NO
- Plugins: None
```

### Jellyfin logs

```shell
None
```

### FFmpeg logs

```shell

```

Bandit:
This is broken again, as of `10.11.2`

Issue #14468, opened 2025-07-14
[10.11] Parental Controls - Block Tags only work partially resulting in media appearing in recently added section.


justunscored:
### Description of the bug

In all 3 10.11 RCs of Jellyfin:
Blocked tags do not seem to function properly at least with the recently added home screen or the expanded list when clicking on the text to go to the suggestions at home screen.

At first I thought it was an issue with capitalization being used but now I believe the tags just don't work in certain parts of jellyfin (see additional info)

### Reproduction steps

1. Open jellyfin web or android(tv) (this is where I tested)
2. Have media blocked (the specific tags I used were Blocked-Movies and Second-Shows, Third-Shows in my tests)
3. Login with a user that has the block tag applied
4. Observe that the blocked media is accessible on the home screen within the recently added list but is not visible within the show menu, potentially is visible under suggestions.

### What is the current _bug_ behavior?

The media with the tag applied will appear in recently added even when blocked but correctly be inaccessible if the user clicks on the image or name of the media.

additionally only on android tv it will disappear from the recently added list till the home page is refreshed.

### What is the expected _correct_ behavior?

The media with the tag applied should disappear from the user so they are "unaware it exists" and should never be seen even in recently added

### Specify the build version

10.11.0-rc3

### Jellyfin logs
Errors from user trying to access media they cannot when clicking on the half-blocked media
```shell
[2025-07-14 02:38:01.647 +00:00] [INF] [30] Emby.Server.Implementations.HttpServer.WebSocketManager: WS "127.0.0.1" request
[2025-07-14 02:38:13.921 +00:00] [INF] [30] Emby.Server.Implementations.HttpServer.WebSocketManager: WS "127.0.0.1" closed
[2025-07-14 02:38:14.100 +00:00] [INF] [30] Emby.Server.Implementations.HttpServer.WebSocketManager: WS "127.0.0.1" request
```

### Relevant screenshots or videos

<img width="382" height="375" alt="Image" src="https://github.com/user-attachments/assets/0d0fade9-1143-43e6-bdb9-2a4c5faf24ae" /> (with only Blocked-Movie applied to user blocklist)

<img width="1013" height="688" alt="Image" src="https://github.com/user-attachments/assets/d3e0918e-7f8f-472e-aff4-88fde504d286" /> (both Blocked-Movie and blocked-movie applied to user blocklist)

### Additional information

[...]

PR #14482, opened 2025-07-16
Fix allow and block queries

theguymadmax:
**Changes**
- Use` ItemValueType.InheritedTags`  || `ItemValueType.Tags `
- Use `e.ItemValues!` directly instead of `e.Parents!`
- Filter tags returned from `GetPreference()`

**Issues**
Fixes: #14472
Fixes: #14468

### DIFF
--- a/MediaBrowser.Controller/Entities/InternalItemsQuery.cs
+++ b/MediaBrowser.Controller/Entities/InternalItemsQuery.cs
@@ -3,6 +3,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using Diacritics.Extensions;
 using Jellyfin.Data;
 using Jellyfin.Data.Enums;
 using Jellyfin.Database.Implementations.Entities;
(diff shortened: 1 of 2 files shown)


MediaBrowser.Controller/Entities/InternalItemsQuery.cs
```diff
@@ -373,8 +374,15 @@ public void SetUser(User user)
                 .Where(i => i != other)
                 .Select(e => Enum.Parse<UnratedItem>(e, true)).ToArray();
 
-            ExcludeInheritedTags = user.GetPreference(PreferenceKind.BlockedTags);
-            IncludeInheritedTags = user.GetPreference(PreferenceKind.AllowedTags);
+            ExcludeInheritedTags = user.GetPreference(PreferenceKind.BlockedTags)
+                .Where(tag => !string.IsNullOrWhiteSpace(tag))
+                .Select(tag => tag.RemoveDiacritics().ToLowerInvariant())
```
Shadowghost, line 379: If we remove diacritics here, don't we also need to remove them before we compare strings in the filter?

theguymadmax: The filter uses CleanValue, which also removes diacritics.

JPVenson:
i want to depricate the inherited tags kind in the long run, i think when creating new BaseItems they are not even created anymore even

Shadowghost (approved):

theguymadmax:
When migrating from 10.10.7 `InheritedTags` is removed, leaving only the `tags` ItemValuesID in the db. But when creating new tags both types (`InheritedTags` & `tags`) are created, so we should check for both.

Bandit:
This is broken again in `10.11.2`

PR #14488, opened 2025-07-17
improve userdata migration

JPVenson:
improves user data migration

Removes existing UserData entries that might be present for 10.11 migrations and duplicate entries from the old db and also filteres duplicate entries for newly added user data entries


crobibero:
Please add a description for what this is doing :)

Shadowghost (approved):

crobibero (approved):

Issue #14484, opened 2025-07-16
[10.11.0-RC3] Jellyfin backup file has unexpected contents


evan314159:
### Description of the bug

I looked inside a backup file from 10.11.0-rc3.  The contents include:

- Database files literally named like "Database\file.json" with the "\\" being part of the filename and not a directory (Windows path separator used on Linux)
- 2 files named /config/config/ScheduledTasks, overwrite each other on unzip
- 16 files named /config/data/ScheduledTasks, overwrite each other on unzip
- 1828 files named /config/data/trickplay, overwrite each other on unzip
- 9704 files named /config/metadata, overwrite each other on unzip

Also:
- use of absolute paths - paths beginning with /
- config/config (e.g. config/config/system.xml) is backed-up in the zip as /config, so if unzipped manually the files will not be put in to the right location

### Reproduction steps

1. Login to Jellyfin web with admin account
2. Use Dashboard->Backups->Create Backup to create a backup
3. Use unzip to look at tool

### What is the current _bug_ behavior?

Excerpt from backup:
```
$ unzip -l jellyfin-backup-20250716162912.zip
Archive:  jellyfin-backup-20250716162912.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
        2  2025-07-16 16:29   Database\AccessSchedules.json ["\" embedded in filename]
    38176  2025-07-16 16:29   Database\ActivityLogs.json
        2  2025-07-16 16:29   Database\ApiKeys.json
     5305  2025-07-16 16:29   Database\Devices.json
        2  2025-07-16 16:29   Database\DeviceOptions.json
      723  2025-07-16 16:29   Database\DisplayPreferences.json
        2  2025-07-16 16:29   Database\ImageInfos.json
      465  2025-07-16 16:29   Database\ItemDisplayPreferences.json
      901  2025-07-16 16:29   Database\CustomItemDisplayPreferences.json
     4619  2025-07-16 16:29   Database\Permissions.json
     2613  2025-07-16 16:29   Database\Preferences.json
     2718  2025-07-16 16:29   Database\Users.json
    82578  2025-07-16 16:29   Database\TrickplayInfos.json
        2  2025-07-16 16:29   Database\MediaSegments.json
     4904  2025-07-16 16:29   Database\UserData.json
 41996137  2025-07-16 16:29   Database\AncestorIds.json
     3702  2025-07-16 16:29   Database\AttachmentStreamInfos.json
[...]

PR #14489, opened 2025-07-18
Clean up and fix backup/restore

crobibero:
Fixes https://github.com/jellyfin/jellyfin/issues/14484

Tested backing up on linux and restoring on windows, and reverse.
Could use more testing with larger datasets.

### DIFF
--- a/Jellyfin.Server.Implementations/FullSystemBackup/BackupService.cs
+++ b/Jellyfin.Server.Implementations/FullSystemBackup/BackupService.cs
@@ -39,7 +39,7 @@ public class BackupService : IBackupService
         ReferenceHandler = ReferenceHandler.IgnoreCycles,
     };
 
-    private readonly Version _backupEngineVersion = Version.Parse("0.1.0");
+    private readonly Version _backupEngineVersion = Version.Parse("0.2.0");
 
     /// <summary>
     /// Initializes a new instance of the <see cref="BackupService"/> class.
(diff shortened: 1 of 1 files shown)


nielsvanvelzen (commented):
Do backups made before this PR still restore or do we need a version bump?

Jellyfin.Server.Implementations/FullSystemBackup/BackupService.cs
```diff
@@ -193,7 +193,7 @@ void CopyDirectory(string source, string target)
                     {
                         _logger.LogInformation("Read backup of {Table}", entityType.Type.Name);
 
-                        var zipEntry = zipArchive.GetEntry($"Database\\{entityType.Type.Name}.json");
+                        var zipEntry = zipArchive.GetEntry(Path.Join("Database", $"{entityType.Type.Name}.json"));
```
sjorge: There is a mix of Path.Join and Path.Combine and It's not clear why sometimes one is used over the other.

Jellyfin.Server.Implementations/FullSystemBackup/BackupService.cs
```diff
@@ -193,7 +196,7 @@ void CopyDirectory(string source, string target)
                     {
                         _logger.LogInformation("Read backup of {Table}", entityType.Type.Name);
 
-                        var zipEntry = zipArchive.GetEntry($"Database\\{entityType.Type.Name}.json");
+                        var zipEntry = zipArchive.GetEntry(NormalizePathSeparator(Path.Join("Database", $"{entityType.Type.Name}.json")));
```
sjorge: Why does this one use Path.Join? Other places use Path.Combine.

crobibero:
> Do backups made before this PR still restore or do we need a version bump?

Good point. Backups from previous RCs don't work so I've bumped the version.

Shadowghost (approved):

JPVenson (approved):

PR #14493, opened 2025-07-18
Fix trickplay extraction ffmpeg error-handling

jkhsjdhjs:
If ffmpeg runs to completion, i.e. isn't killed due to being unresponsive, its exit code is >0 in case of an error. However, the current logic only throws an exception in case the exit code is -1, which is the virtual exit code assigned in case ffmpeg has been killed prematurely.

This commit simplifies the logic to throw an exception if ffmpeg didn't run to completion (as before) or exited with a non-zero exit code (new).

This change is especially important for #14467, as the fallback will not be triggered if ffmpeg errors are not recognized properly.

### DIFF
--- a/MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
+++ b/MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
@@ -1071,9 +1071,7 @@ namespace MediaBrowser.MediaEncoding.Encoder
                     }
                 }
 
-                var exitCode = ranToCompletion ? processWrapper.ExitCode ?? 0 : -1;
-
-                if (exitCode == -1)
+                if (!ranToCompletion || processWrapper.ExitCode != 0)
                 {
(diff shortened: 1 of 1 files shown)


JPVenson (approved):

jkhsjdhjs:
@JPVenson I added one commit, please re-review, sorry. I decided to add it here since it's only one line and somewhat related to the previous change.

As for the commit:
In case trickplay extraction with ffmpeg fails, an error is logged and an exception is raised. Logging the error here is unnecessary, since it is already logged by the exception handler.

The exception is already logged by this handler:
https://github.com/jellyfin/jellyfin/blob/ba0f61ef2defdd961a4fb874b0b0e0daac74d9c7/Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs#L377-L380

This change makes sense by itself, but is also relevant for #14467, which also handles this exception. Without this change, jellyfin will first log an error, saying the extraction failed, followed by a warning, again stating that the extraction failed and that the fallback will be attempted.

crobibero (approved):

PR #14503, opened 2025-07-21
MetadataService: fix file modification date comparison

evan314159:
**Changes**
On filesystems like ZFS with higher modification time accuracy than the database, the comparison itemLastModifiedFileSystem != modificationDate fails.  Allow a tolerance of 1 second.

**Issues**
No issue raised.  On ZFS causes every file to be rescanned on every scan run.


Shadowghost:
I wouldn't mind merging this but I use ZFS too and never had any issue with this while testing before and after we released the RCs.

ghost:
I never noticed this until I put 40k music tracks into Jellyfin and found scans took all day and reprocessed all media.  With debugging enabled I recorded logs like:

[08:42:41] [DBG] [31] [MediaBrowser.Providers.TV](http://mediabrowser.providers.tv/).EpisodeMetadataService: File modification time changed from 03/12/2025 13:14:21 to 03/12/2025 13:14:21:

With the changes a library scan takes a few minutes.

Shadowghost:
But `EpisodeMetadataService` should not trigger on music library?

ghost:
The problem was on all my media and that episode happened to be the one I chose to reference in the jellyfin developer chat, so I had a record of it in the chat to give here.

gnattu (approved):

Emby.Photos/PhotoProvider.cs
```diff
@@ -49,7 +49,7 @@ public bool HasChanged(BaseItem item, IDirectoryService directoryService)
         if (item.IsFileProtocol)
         {
             var file = directoryService.GetFile(item.Path);
-            return file is not null && file.LastWriteTimeUtc != item.DateModified;
+            return file is not null && item.DateModified.Subtract(file.LastWriteTimeUtc).Duration().TotalSeconds > 1;
```
cvium: This is copied in so many places now that I think it warrants an extension/utility method.

ghost:
> We have to find a better way in the long term to do those comparisions like truncating the timestamps to ms accuracy in the DirectoryService and FileService but for now that will do it

FileService will return whatever with whatever accuracy on a given platform, the database will return with whatever accuracy on a given platform, and Jelly should not care as long as the file modification date is sufficiently different for its purposes.  

For this change "sufficiently different" was chosen to be 1 second, but there is existing code in Jelly that works the same way for "sufficiently different" being 1 day instead (for once per day updates).  It just can't directly compare fileModified == databaseModified, just like floating point values can't be directly compared with == because floating point accuracy can vary.

The code is simple and robust; if we are worried about it being ugly, it could easily be done as a helper function (e.g. if fileChanged(databaseModified, fileModified, accuracySeconds = 1) { } and then is also usable in those cases using a 1 day resolution.

PR #14536, opened 2025-07-26
Ignore directory if empty .ignore file is present

### LINKED ISSUE #14502: Folders with empty .ignore file are added to the library
### Description of the bug

After setting up a 10.11.x test environment, I scanned my current library and all Folders with an empty .ignore file are included. I can see in the debugger that the DotIgnoreIgnoreRule code is executed for all files inside one of the folders, returning true for IsIgnored.
If a folder is ignored, the scanner should simply abort instead of checking each element in the folder (and sub folders). And the scanner should mark the folder itself as ignored and don't add it to the library.

### Reproduction steps

The bug is also present in the movie library, but since each folder in a season folder is normally directly added to the show overview, it's easier to directly see the problem.
 
1. Create a new folder inside a season main directory in a tv show library
2. Create an empty `.ignore` file inside that folder
3. refresh metadata -> scan for new and updated files
4. ignored folder shows up in the tv show overview

Scanner adds ignored folders to the library and checks each sub element individually.

1. If a folder is ignored, the scanner should simply abort instead of checking each element in the folder (and sub folders)
2. The scanner should mark the folder itself as ignored and don't add it to the library.

### Jellyfin Server version

Master

### Specify commit id

6cee66119e02d7afc9bd9ce24c25a33ba8ba71d1

_No response_

### Specify version number

_No response_

### Specify the build version

Dev Server

### Environment

```markdown
- OS:Windows 10
- Clients: browser
- Browser: Firefox / Edge
- FFmpeg Version: jellyfin-ffmpeg_7.0.2-7-portable_win64
- Pliugins: AudioDB 10.11.0.0, MusicBrainz 10.11.0.0. OMDb 10.11.0.0, Studio Images 10.11.0.0, TMDb 10.11.0.0, TMDb Box Sets 11.0.0.0
- Jellyfin Data Storage: SSD; Windows standard folder
- Media Storage: SMB NAS
```

### Jellyfin logs

```shell
Renaming and rescanning the show folder:

[21:26:50] [INF] [8] Emby.Server.Implementations.Library.LibraryManager: Removing item, Type: Season, Name: Covers and Logos, Path: V:\SHOWS\Farscape\Covers and Logos, Id: 58788f57-5e53-2ac4-2635-4ed0292c9936
[21:27:31] [INF] [11] Emby.Server.Implementations.IO.LibraryMonitor: Covers and Logos 1 (V:\SHOWS\Farscape\Covers and Logos 1) will be refreshed.
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

Image

Folder structure:
```
V/
├─ SHOWS/
│  ├─ Farscape/
│  │  ├─ Covers and Logos/
│  │  │  ├─ .ignore
│  │  │  ├─ ...
│  │  ├─ Season 01/
│  │  │  ├─ ...
│  │  ├─ Season 02/
│  │  │  ├─ ...
│  │  ├─ Season 03/
│  │  │  ├─ ...
│  │  ├─ Season 04/
│  │  │  ├─ ...
│  │  ├─ Season 05 - Peacekeeper Wars/
│  │  │  ├─ ...
│  │  ├─ tvshow.nfo
│  │  ├─ ...
```

### Additional information

_No response_

piyushrungta25:
Fixes #14502

### DIFF
--- a/Emby.Server.Implementations/Library/DotIgnoreIgnoreRule.cs
+++ b/Emby.Server.Implementations/Library/DotIgnoreIgnoreRule.cs
@@ -42,6 +42,19 @@ public class DotIgnoreIgnoreRule : IResolverIgnoreRule
     /// <returns>True if the file should be ignored.</returns>
     public static bool IsIgnored(FileSystemMetadata fileInfo, BaseItem? parent)
     {
+        if (fileInfo.IsDirectory)
+        {
+            var dirIgnoreFile = FindIgnoreFile(new DirectoryInfo(fileInfo.FullName));
+            if (dirIgnoreFile is null)
+            {
+                return false;
+            }
+
+            // ignore the directory only if the .ignore file is empty
+            // evaluate individual files otherwise
+            return string.IsNullOrWhiteSpace(GetFileContent(dirIgnoreFile));
+        }
+
         var parentDirPath = Path.GetDirectoryName(fileInfo.FullName);
         if (string.IsNullOrEmpty(parentDirPath))
         {
@@ -55,13 +68,9 @@ public class DotIgnoreIgnoreRule : IResolverIgnoreRule
             return false;
         }
 
-        string ignoreFileString;
-        using (var reader = ignoreFile.OpenText())
-        {
-            ignoreFileString = reader.ReadToEnd();
-        }
+        string ignoreFileString = GetFileContent(ignoreFile);
 
-        if (string.IsNullOrEmpty(ignoreFileString))
+        if (string.IsNullOrWhiteSpace(ignoreFileString))
         {
             // Ignore directory if we just have the file
             return true;
@@ -73,5 +82,13 @@ public class DotIgnoreIgnoreRule : IResolverIgnoreRule
         ignore.Add(ignoreRules);
 
         return ignore.IsIgnored(fileInfo.FullName);
+    }
+
+    private static string GetFileContent(FileInfo dirIgnoreFile)
+    {
+        using (var reader = dirIgnoreFile.OpenText())
+        {
+            return reader.ReadToEnd();
+        }
     }
 }


jumoog:
I think checking whether the `.ignore` file is empty can be skipped. Reading the entire file seems like a lot of unnecessary work.

piyushrungta25:
> I think checking whether the `.ignore` file is empty can be skipped. Reading the entire file seems like a lot of unnecessary work.

@jumoog If we skip it entirely, the pattern based ignore feature of #13906 will never get executed? If any directory has a ignore file with pattern to ignore few files only, the directory itself will get ignored and the child items of the directory will never be processed.

I can alternatively check if the file size == 0 instead of reading the contents. But I like the IsNullOrWhiteSpace check better as it handles the cases where some utilities create an "empty" file with a trailing new line (nemo file manager on linux does it for instance).

jumoog:
Oh i see! It's also like .gitignore. I didn't knew. And from what i'm reading `file size == 0` doesn't work in some edge cases.

jumoog:
hmm wouldn't it be better to stay with `.jellyfinignore` (parse that) and for `.ignore` (no parsing) skip?

piyushrungta25:
> hmm wouldn't it be better to stay with `.jellyfinignore` (parse that) and for `.ignore` (no parsing) skip?

I believe the design decision to use single .ignore file was already made in the last PR. Will let @Shadowghost comment if we want to revisit that decision.

Shadowghost:
Rationale was to extend the support of existing `.ignore` logic and not have two things. Sadly there is no standard for such a "config"

Shadowghost (approved):

crobibero (approved):

pando85:
Sadly, I hit this breaking change. Documentation would help in this case.

Shadowghost:
This is not a breaking change. We've been supporting .ignore for ages, the difference being that we now support complex ignore patterns. There were some fixes to explosions of extras though because there the .ignore wasn't respected in certain circumstances

pando85:
I'm just telling that this change broke my setup ( my ignore file was with a comment and not empty) and I had to dig into the PRs to figure out what happens there.

PR #14540, opened 2025-07-26
Improve extra rule resolution and file handling

TokerX:
**NamingOptions.cs**

- Updated suffix-based extra rules to prevent false positives on
legitimate file names ending in trailer or sample.

**ExtraRuleResolver.cs**

- Refactored if/else if into a switch expression for clarity.
- Moved repeated result assignment outside conditional blocks to
reduce duplication.
- Lifted invariant calculations outside the loop to improve
efficiency.

**CoreResolutionIgnoreRule.cs**

- Added support for checking top parent.
- Reduced nesting for readability.
- Removed redundant  "&& parent is not AggregateFolder" check
already evaluated upstream.

**Tests**

**ExtraTests.cs** and **VideoListResolverTests.cs**: updated to reflect
new logic.

**CoreResolutionIgnoreRuleTest.cs**: new tests added

**PluginManagerTests.cs**: fixed test failure by formatting
versionInfo.Timestamp using the "o" format to ensure consistent
UTC parsing.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -200,6 +200,7 @@
  - [ThunderClapLP](https://github.com/ThunderClapLP)
  - [Shoham Peller](https://github.com/spellr)
  - [theshoeshiner](https://github.com/theshoeshiner)
+ - [TokerX](https://github.com/TokerX)
 
 # Emby Contributors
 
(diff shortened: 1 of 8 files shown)


Shadowghost (approved):
Looks fine to me, though the `sample` -> `- sample` changes is technically breaking.

PR #14543, opened 2025-07-27
Preserve interpunct (·)

### LINKED ISSUE #14541: Jellyfin does not recognise "·" character in a filename
### Description of the bug

In Jellyfin 10.11.0 RC3 & 2025072105 Unstable, Jellyfin is not able to recognise the "·" character. https://en.wikipedia.org/wiki/Interpunct

This character is used in Catalan language in normal words. It means you need to write it in the filename so the Catalan series "V.I.E. Vida Intel·ligent a l'Espai" can be detected in TMDB by Jellyfin. This character is even used in the word "Pel·lícula" (film), so is need in 56 titles in Jellyfin with this word in the name (example: The Super Mario Bros. Movie > Super Mario Bros.: La pel·lícula)

### Reproduction steps

Jellyfin is not finding metadata if the filename is for example "Angry Birds La pel·lícula" but it does if the filename is "The Angry Birds Movie". The same happens withe any filename with the character "·".

If there is a filename with the character "·", Jellyfin doesn't obtain metadata from TMDB.

If there is a filename with the character "·", Jellyfin obtains metadata from TMDB.

### Jellyfin Server version

10.10.0+

### Specify commit id

_No response_

2025072105

### Specify version number

10.11.0 RC3

### Specify the build version

2025071405

### Environment

```markdown
- OS: Windows 11 Home
- Browser: Firefox
```

### Jellyfin logs

```shell
Don't know how to do that
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

theguymadmax:
**Changes**
Change regex to exclude interpunct

**Issues**

 Fixes #14541

### DIFF
--- a/MediaBrowser.Providers/Plugins/Tmdb/TmdbUtils.cs
+++ b/MediaBrowser.Providers/Plugins/Tmdb/TmdbUtils.cs
@@ -48,7 +48,7 @@ namespace MediaBrowser.Providers.Plugins.Tmdb
             PersonKind.Producer
         };
 
-        [GeneratedRegex(@"[\W_]+")]
+        [GeneratedRegex(@"[\W_-[·]]+")]
         private static partial Regex NonWordRegex();
 
         /// <summary>


Shadowghost (approved):

crobibero (approved):
~Could probably add this to the test suite 🤷~

nvm this is just tmdb specific

PR #14550, opened 2025-07-28
AudioNormalizationTask db progress saving

shanepowell:
Change db updating in AudioNormalizationTask to save in 5 min intervals of work to make sure progress is saved regularly so if the task is cancelled progress is not lost.

As requested by https://github.com/jellyfin/jellyfin/pull/14306.

**Changes**
Break up the SaveItems into roughly 5 min intervals of work.

**Issues**
https://github.com/jellyfin/jellyfin/pull/14306

### DIFF
--- a/Emby.Server.Implementations/ScheduledTasks/Tasks/AudioNormalizationTask.cs
+++ b/Emby.Server.Implementations/ScheduledTasks/Tasks/AudioNormalizationTask.cs
@@ -33,6 +33,8 @@ public partial class AudioNormalizationTask : IScheduledTask
     private readonly ILocalizationManager _localization;
     private readonly ILogger<AudioNormalizationTask> _logger;
 
+    private static readonly TimeSpan _dbSaveInterval = TimeSpan.FromMinutes(5);
+
     /// <summary>
     /// Initializes a new instance of the <see cref="AudioNormalizationTask"/> class.
     /// </summary>
@@ -82,7 +84,9 @@ public partial class AudioNormalizationTask : IScheduledTask
 
         foreach (var library in libraries)
         {
+            var startDbSaveInterval = Stopwatch.GetTimestamp();
             var albums = _libraryManager.GetItemList(new InternalItemsQuery { IncludeItemTypes = [BaseItemKind.MusicAlbum], Parent = library, Recursive = true });
+            var toSaveDbItems = new List<BaseItem>();
 
             double nextPercent = numComplete + 1;
             nextPercent /= libraries.Length;
@@ -114,12 +118,24 @@ public partial class AudioNormalizationTask : IScheduledTask
                                 string.Format(CultureInfo.InvariantCulture, "-f concat -safe 0 -i \"{0}\"", tempFile),
                                 OperatingSystem.IsWindows(), // Wait for process to exit on Windows before we try deleting the concat file
                                 cancellationToken).ConfigureAwait(false);
+                            toSaveDbItems.Add(a);
                         }
                         finally
                         {
                             File.Delete(tempFile);
                         }
                     }
+                }
+
+                if (Stopwatch.GetElapsedTime(startDbSaveInterval) > _dbSaveInterval)
+                {
+                    if (toSaveDbItems.Count > 1)
+                    {
+                        _itemRepository.SaveItems(toSaveDbItems, cancellationToken);
+                        toSaveDbItems.Clear();
+                    }
+
+                    startDbSaveInterval = Stopwatch.GetTimestamp();
                 }
 
                 // Update sub-progress for album gain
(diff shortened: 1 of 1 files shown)


Shadowghost:
I wonder how bad the performance impact would be if we save after each album

nielsvanvelzen:
Considering there may be albums with just 1-3 tracks it could be bad. Maybe we should just save after every 100 calculations or so? Most of our scheduled tasks work like that.

shanepowell:
From my point of view, time based batch saving is a good compromise between performance and safety.

ghost:
I suggest to allow Shane's PR as-is so we can have the progress indicator.  I will be happy to raise a PR once it is merged to commit after every item so we can continue this discussion.

This is a very strange place to be concerned about Jellyfin's efficiency.  I estimate that re-reading 1 average song (in FLAC) is about equivalent I/O to checkpointing 1000 items.

If we are really worried about I/O during one-off operations maybe we should replace ffmpeg with rsgain which does parallel single-pass album scanning, an opportunity for 50% reduction in I/O and large reduction in time to complete.  Jellyfin should also run ffmpeg (or rsgain) at a higher nice / lower priority during scanning and normalisation so as not to interfere with other tasks.

HammyHavoc:
> I suggest to allow Shane's PR as-is so we can have the progress indicator. I will be happy to raise a PR once it is merged to commit after every item so we can continue this discussion.
> 
> This is a very strange place to be concerned about Jellyfin's efficiency. I estimate that re-reading 1 average song (in FLAC) is about equivalent I/O to checkpointing 1000 items.
> 
> If we are really worried about I/O during one-off operations maybe we should replace ffmpeg with rsgain which does parallel single-pass album scanning, an opportunity for 50% reduction in I/O and large reduction in time to complete. Jellyfin should also run ffmpeg (or rsgain) at a higher nice / lower priority during scanning and normalisation so as not to interfere with other tasks.

Completely agreed re avoiding interference with other tasks as it can be very heavy on low-spec (read "energy efficient") hardware that otherwise had no problem with Plex.

ghost:
> Completely agreed re avoiding interference with other tasks as it can be very heavy on low-spec (read "energy efficient") hardware that otherwise had no problem with Plex.

PR #14651

crobibero (commented):
I would prefer using the Stopwatch object and letting it handle elapsed calculation. It's basically the same just easier to parse.

```patch
Index: Emby.Server.Implementations/ScheduledTasks/Tasks/AudioNormalizationTask.cs
IDEA additional info:
<+>UTF-8
diff --git a/Emby.Server.Implementations/ScheduledTasks/Tasks/AudioNormalizationTask.cs b/Emby.Server.Implementations/ScheduledTasks/Tasks/AudioNormalizationTask.cs
--- a/Emby.Server.Implementations/ScheduledTasks/Tasks/AudioNormalizationTask.cs	(revision 15799a0e8ce8aab198ee1b45a6a2c2d3e35f3b03)
+++ b/Emby.Server.Implementations/ScheduledTasks/Tasks/AudioNormalizationTask.cs	(date 1755359264182)
@@ -84,7 +84,7 @@
 
         foreach (var library in libraries)
         {
-            var startDbSaveInterval = Stopwatch.GetTimestamp();
+            var stopwatch = Stopwatch.StartNew();
             var albums = _libraryManager.GetItemList(new InternalItemsQuery { IncludeItemTypes = [BaseItemKind.MusicAlbum], Parent = library, Recursive = true });
             var toSaveDbItems = new List<BaseItem>();
 
@@ -127,7 +127,7 @@
                     }
                 }
 
-                if (Stopwatch.GetElapsedTime(startDbSaveInterval) > _dbSaveInterval)
+                if (stopwatch.Elapsed > _dbSaveInterval)
                 {
                     if (toSaveDbItems.Count > 1)
                     {
@@ -135,7 +135,7 @@
                         toSaveDbItems.Clear();
                     }
 
-                    startDbSaveInterval = Stopwatch.GetTimestamp();
+                    stopwatch.Restart();
                 }
 
                 // Update sub-progress for album gain
@@ -155,7 +155,7 @@
                 toSaveDbItems.Clear();
             }
 
-            startDbSaveInterval = Stopwatch.GetTimestamp();
+            stopwatch.Restart();
 
             // Track gain
             var tracks = _libraryManager.GetItemList(new InternalItemsQuery { MediaTypes = [MediaType.Audio], IncludeItemTypes = [BaseItemKind.Audio], Parent = library, Recursive = true });
@@ -172,7 +172,7 @@
                     toSaveDbItems.Add(t);
                 }
 
-                if (Stopwatch.GetElapsedTime(startDbSaveInterval) > _dbSaveInterval)
+                if (stopwatch.Elapsed > _dbSaveInterval)
                 {
                     if (toSaveDbItems.Count > 1)
                     {
@@ -180,7 +180,7 @@
                         toSaveDbItems.Clear();
                     }
 
-                    startDbSaveInterval = Stopwatch.GetTimestamp();
+                    stopwatch.Restart();
                 }
 
                 // Update sub-progress for track gain
```

shanepowell:
> I would prefer using the Stopwatch object and letting it handle elapsed calculation. It's basically the same just easier to parse.

I disagree.  The Stopwatch.GetTimestamp/Stopwatch.GetElapsedTime usage is a more efficient Stopwatch usage.
See: [Efficient Stopwatch usage](https://www.alwaysdeveloping.net/dailydrop/2023/01/05-new-stopwatch/) or [Are you using the Stopwatch efficiently in .NET?](https://www.youtube.com/watch?v=NTz99yN2urc)

crobibero (approved):

PR #14556, opened 2025-07-31
Add support MsSql database

Tony-Starling:
Add support MsSql database

**Changes**

 Create new database provider to mssql database with initial database migration.

**Issues**


crobibero:
While we appreciate the eagerness to support alternative database engines only SQLite will be included in the core server, any other engines would be built as plugins.

PR #14565, opened 2025-08-02
Improve RecordFromMediaSource resilience.

FerronN:
Hi,

I recently added LiveTV functionality to my Jellyfin installation. When recording programs, I noticed that my recordings were much shorter then then they should be. For example, if I the recording should be 1,5 hours, it only recorded 1 to 30 minutes, but I couldn't find any errors. Probably because the streams I watch aren't as stable as you want to be. Although you don't notice this when watching live (buffering in the Android TV client).

I noticed that the current code doesn't have any error handling if the stream for whatever reasons stops before the durationToken cancellation is requested. So, I updated the code to handle this. For me this works perfectly now.

I think the recording should always try to reconnect as long as the linkedCancellationToken cancellation hasn't been requested.

Please let me know if you have feedback on my implementation. 

**Changes**
- Added retry logic for restarting the stream for method `RecordFromMediaSource` in `EncodedRecorder.cs`.
- Added a bufferstream to smooth out minor hiccups (shows less drops then the retry logic) for `CopyToAsyncInternal` in `StreamHelper.cs`.

### DIFF
--- a/MediaBrowser.Model/IO/IODefaults.cs
+++ b/MediaBrowser.Model/IO/IODefaults.cs
@@ -21,5 +21,10 @@ namespace MediaBrowser.Model.IO
         /// The default <see cref="StreamWriter" /> buffer size.
         /// </summary>
         public const int StreamWriterBufferSize = 1024;
+
+        /// <summary>
+        /// The default buffer stream buffer size.
+        /// </summary>
+        public const int BufferStreamBufferSize = 131072;
     }
 }
--- a/src/Jellyfin.LiveTv/IO/DirectRecorder.cs
+++ b/src/Jellyfin.LiveTv/IO/DirectRecorder.cs
@@ -81,32 +81,59 @@ namespace Jellyfin.LiveTv.IO
 
         private async Task RecordFromMediaSource(MediaSourceInfo mediaSource, string targetFile, TimeSpan duration, Action onStarted, CancellationToken cancellationToken)
         {
-            using var response = await _httpClientFactory.CreateClient(NamedClient.Default)
-                .GetAsync(mediaSource.Path, HttpCompletionOption.ResponseHeadersRead, cancellationToken).ConfigureAwait(false);
+            using var httpClient = _httpClientFactory.CreateClient(NamedClient.Default);
+            int maxRetries = 5;
+            int retry = 0;
 
-            _logger.LogInformation("Opened recording stream from tuner provider");
+            // The media source is infinite so we need to handle stopping ourselves
+            using var durationToken = new CancellationTokenSource(duration);
+            using var linkedCancellationToken = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, durationToken.Token);
+            cancellationToken = linkedCancellationToken.Token;
 
-            Directory.CreateDirectory(Path.GetDirectoryName(targetFile) ?? throw new ArgumentException("Path can't be a root directory.", nameof(targetFile)));
-
-            var output = new FileStream(targetFile, FileMode.CreateNew, FileAccess.Write, FileShare.Read, IODefaults.CopyToBufferSize, FileOptions.Asynchronous);
-            await using (output.ConfigureAwait(false))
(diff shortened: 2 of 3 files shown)


src/Jellyfin.LiveTv/IO/StreamHelper.cs
```diff
@@ -106,8 +131,9 @@ private static async Task<int> CopyToAsyncInternal(Stream source, Stream destina
         {
             int bytesRead;
             int totalBytesRead = 0;
+            using var bufferedStream = new BufferedStream(source, IODefaults.BufferStreamBufferSize);
```
Bond-009, line 134: Maybe move the creation of the `BufferedStream` up the call stack so we don't have to allocate its buffer every time.

MediaBrowser.Model/IO/IODefaults.cs
```diff
@@ -21,5 +21,10 @@ public static class IODefaults
         /// The default <see cref="StreamWriter" /> buffer size.
         /// </summary>
         public const int StreamWriterBufferSize = 1024;
+
+        /// <summary>
+        /// The default buffer stream buffer size.
+        /// </summary>
+        public const int BufferStreamBufferSize = 131072;
```
Bond-009, line 28: The other defaults are copied from the .NET source code, while this one isn't.

Bond-009 (changes requested):

src/Jellyfin.LiveTv/IO/StreamHelper.cs
FerronN: I will change that

MediaBrowser.Model/IO/IODefaults.cs
FerronN: Ah I didn't know that. Should I put it in a different constant class or put the value directly into the bufferstream constructor?

FerronN:
Everything should be okay now :)

src/Jellyfin.LiveTv/IO/DirectRecorder.cs
```diff
@@ -81,32 +81,63 @@ await _streamHelper.CopyToAsync(
...
+                    _logger.LogInformation("Recording completed to file {0}", targetFile);
+                }
+                catch (Exception ex)
+                {
+                    retry++;
+                    if (retry >= maxRetries)
+                    {
+                        _logger.LogError("Stream failed permanently after retries. {0}", ex.Message);
+                        return;
+                    }
```
JPVenson, line 131: I would rather see polly being used for resiliancy

JPVenson (changes requested):

FerronN:
@JPVenson @Bond-009  I've changed it to use Polly. However, I moved the `BufferedStream` instantiation back to `CopyToAsyncInternal`.

I can't really explain it, but my streams are sometimes cut short when the `BufferedStream` instantiation is done in `CopyUntilCancelled` and passed to `CopyToAsyncInternal`.

The logs don't show any logs when this happens. The timer even fires when the recording should be stopped.
For example, I had a recording that should take 1,5 hours, but the duration of the recording was only 2 and once 8 minutes, before I moved the `BufferedStream` back.

I will validate the lasted commit in the coming few days to see if I experience any issues.

FerronN:
Sorry for the mess, but I think my fist implementation is still the best except for some details. 
I've been so focused on making the feedback work that I forgot the main reason I started this.

The while loop:

https://github.com/jellyfin/jellyfin/pull/14565/files/dcaa7a5932fd58adc0323a09e18be28a37592b55 

Line 93

which keeps retrying even when the http request/stream was successful but stopped before cancellation was requested. To make this work I still have to keep some retry count, with currently al lot of additional code for the usage of Polly. Because Polly only catches exceptions.

Or am I missing something?

theguymadmax:
@FerronN Please re-target this to the master branch; otherwise, this will be closed.

Bond-009:
Closing as it still targets the 10.10 release branch

PR #14591, opened 2025-08-06
Fix Peoples to remove duplicates, Artists and Artists-like types, and treat Conductors and Composers like Artists

ghost:
**Changes**
This change fixes:
- Duplicate inserts into Peoples using an INSERT ON CONFLICT DO NOTHING-type approach
- Stop creating Person records for Artists
- Treat Composers and Conductors as Artists

Includes migration to:
- Remap Conductors and Composers into Artists before removing from People
- Remove duplicate and unlinked People
- Add unique constraint index on Peoples.(Name, PersonType)
- Remove unnecessary index on Peoples.Name

Conductors and Composers show up right away as Artists after the migration but requires a scan before they will appear in the web UI with their Artist picture.

Web UI changes in jellyfin/jellyfin-web#7069

**Issues**

I had 300k people in Peoples because on every run through it was inserting duplicates.
Fixes artists showing up as people with incorrect links to TMDB etc., no links to their albums.

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -3,7 +3,7 @@
   "isRoot": true,
   "tools": {
     "dotnet-ef": {
-      "version": "9.0.7",
+      "version": "9.0.8",
       "commands": [
         "dotnet-ef"
       ]
--- a/.github/workflows/ci-codeql-analysis.yml
+++ b/.github/workflows/ci-codeql-analysis.yml
@@ -20,18 +20,18 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
     - name: Setup .NET
       uses: actions/setup-dotnet@67a3573c9a986a3f9c594539f4ab511d57bb3ce9 # v4.3.1
       with:
         dotnet-version: '9.0.x'
 
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@51f77329afa6477de8c49fc9c7046c15b9a4e79d # v3.29.5
+      uses: github/codeql-action/init@df559355d593797519d70b90fc8edd5db049e7a2 # v3.29.9
       with:
         languages: ${{ matrix.language }}
         queries: +security-extended
     - name: Autobuild
-      uses: github/codeql-action/autobuild@51f77329afa6477de8c49fc9c7046c15b9a4e79d # v3.29.5
+      uses: github/codeql-action/autobuild@df559355d593797519d70b90fc8edd5db049e7a2 # v3.29.9
     - name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@51f77329afa6477de8c49fc9c7046c15b9a4e79d # v3.29.5
+      uses: github/codeql-action/analyze@df559355d593797519d70b90fc8edd5db049e7a2 # v3.29.9
--- a/.github/workflows/ci-compat.yml
+++ b/.github/workflows/ci-compat.yml
@@ -11,7 +11,7 @@ jobs:
     permissions: read-all
     steps:
       - name: Checkout repository
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
           ref: ${{ github.event.pull_request.head.sha }}
           repository: ${{ github.event.pull_request.head.repo.full_name }}
@@ -40,7 +40,7 @@ jobs:
     permissions: read-all
     steps:
       - name: Checkout repository
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
           ref: ${{ github.event.pull_request.head.sha }}
           repository: ${{ github.event.pull_request.head.repo.full_name }}
@@ -85,13 +85,13 @@ jobs:
 
     steps:
       - name: Download abi-head
-        uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4.3.0
+        uses: actions/download-artifact@634f93cb2916e3fdff6788551b99b062d0335ce0 # v5.0.0
         with:
           name: abi-head
           path: abi-head
 
       - name: Download abi-base
-        uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4.3.0
+        uses: actions/download-artifact@634f93cb2916e3fdff6788551b99b062d0335ce0 # v5.0.0
         with:
           name: abi-base
           path: abi-base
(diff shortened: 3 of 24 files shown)


Shadowghost:
Why do we change composer and Conductor to artist? IMO this change should be done independently with a proper reasoning.

Jellyfin.Server.Implementations/Item/PeopleRepository.cs
```diff
@@ -94,16 +160,41 @@ public void UpdatePeople(Guid itemId, IReadOnlyList<PersonInfo> people)
...
+        {
+            // Use reflection to check SqlState property
+            var sqlStateProperty = ex.InnerException.GetType().GetProperty("SqlState");
+            if (sqlStateProperty?.GetValue(ex.InnerException) is string sqlState)
             {
-                // person mapping already exists so remove from list
-                maps.Remove(existingMap);
+                return sqlState == "23505"; // unique_violation
             }
         }
```
Shadowghost, line 191: Since we don't really supoprt postgres right now, we might not want to add any extra handling for it

cc @JPVenson

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
```diff
@@ -244,48 +232,61 @@ private async Task FetchDataFromTags(Audio audio, Model.MediaInfo.MediaInfo medi
...
-                        {
-                            Name = performer,
-                            Type = PersonKind.Artist
-                        });
+                        composers = composers.SelectMany(c => SplitWithCustomDelimiter(c, libraryOptions.GetCustomTagDelimiters(), libraryOptions.DelimiterWhitelist)).ToArray();
                     }
                 }
+                else
+                {
+                    composers = Array.Empty<string>();
```
Shadowghost:
```suggestion
                    composers = [];
```

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
```diff
@@ -244,48 +232,61 @@ private async Task FetchDataFromTags(Audio audio, Model.MediaInfo.MediaInfo medi
...
-                                Name = composer,
-                                Type = PersonKind.Composer
-                            });
-                        }
+                        conductors = conductors.SelectMany(c => SplitWithCustomDelimiter(c, libraryOptions.GetCustomTagDelimiters(), libraryOptions.DelimiterWhitelist)).ToArray();
                     }
                 }
+                else
+                {
+                    conductors = Array.Empty<string>();
```
Shadowghost, line 266:
```suggestion
                    conductors = [];
```

src/Jellyfin.Database/Jellyfin.Database.Implementations/Entities/ItemValueType.cs
```diff
@@ -16,6 +16,11 @@ public enum ItemValueType
     /// </summary>
     AlbumArtist = 1,
 
+    /// <summary>
+    /// Composer.
+    /// </summary>
+    Composer = 5,
```
Shadowghost: Move it down so the numbering is in order. What's the reasoning behind moving composer to ItemValueType?

ghost:
> Why do we change composer and Conductor to artist? IMO this change should be done independently with a proper reasoning.

When composer or conductor is added as a Person, the metadata comes from Movie sources like an Actor or Director, and they are not linked back to their albums.

I had separate originally but then I needed to do a separate migration for that when I was already having to fix up Peoples.

src/Jellyfin.Database/Jellyfin.Database.Implementations/Entities/ItemValueType.cs
ghost: Sorry this was a mistake, fixed.

Jellyfin.Server.Implementations/Item/PeopleRepository.cs
ghost: EF Core does not have a portable way to either to UPSERT or to catch duplicate key exceptions, and Postgres is planned for next version.  There are no additional dependencies.

A more ideal approach would be to have the database interface offer an UPSERT capability, but that seemed a more significant functionality to implement.

ghost:
> Why do we change composer and Conductor to artist? IMO this change should be done independently with a proper reasoning.

By being in People, Composers and Conductors are treated like Actors and Directors with metadata from movie/TV metadata sources and missing their linkage back to their albums which is by Artist/Album Artist.

This was the logic I posted in the developer chat:

- MusicArtist records are appropriate for Artists, AlbumArtists, Composers, Conductors and link to their albums and music videos
- Person records are appropriate for Actors, Directors, etc., and link to their records in IMDB etc. and their movies/TV shows
- A human should not have both a Person and MusicArtist unless they are both a musician and something else, which happens but only infrequently.
- As MusicArtists won't need Person records they will not appear in Peoples and their PeopleTypes can be removed.
- Peoples still needs dedup logic for the Actors etc.

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
ghost: Fixed style

Jellyfin.Server.Implementations/Item/PeopleRepository.cs
ghost: If the preference is to have the database interface offer an UPSERT capability, if someone could design how it should work I will implement it.  If the logic was moved into the database plugin / SQLite code then please also consider if we could implement using an extension like EFCore.BulkExtensions rather than DIY.

Shadowghost:
I wonder if we could just unify those types. Seems a bit random to differentiate people between music and video.
But regardless, that's a way bigger change.

ghost:
Yes, I was surprised to find Artists and People are different but it is understandable given they use different metadata sources that do not have a consistent reference.  There are other artist metadata issues as well, like John Williams the composer and conductor, and John Williams the guitar player, are two different people.

Shadowghost:
We can't really do much about that because the metadata providers usually don't differentiate (MB does, bnut others do not). We have the same issue with any other person too.

Jellyfin.Server.Implementations/Item/PeopleRepository.cs
JPVenson: Yes having provider specific code is a no go here, all core code needs to be db agnostic.

Adding the BulkExtensions was planned for 10.12, but for now we need to seperate optimisations and features as much as possible. I am not willing to add new features here because every time we did that over the last RCs it caused a whole lot of bugs again.

ghost: > Adding the BulkExtensions was planned for 10.12, but for now we need to seperate optimisations and features as much as possible.

Unless you really want the fix portion of this my suggestion is to wait for 10.11 to be released and then look at it as a 10.11.1+ or 10.12 change.  I suspect Peoples has been broken for some time, and a non-essential and non-trivial migration is not a good candidate for merging in the middle of a release candidate cycle.

I looked at BulkExtensions.  We can’t import all of BulkExtensions as it has x86 and amd64 native code for SQL Server.  It caused all my libraries to go 80386…  The more specific modules BulkExtensions.SQLite and .PostgreSQL are platform-independent but the Postgres one has dependencies on EFCore’s Postgres, so also cannot be imported in core.  I need some help to design a platform-independent upsert interface.

There is no way to do platform independent upsert otherwise.  It’s bizarre but EF Core doesn’t have a platform-independent exception for duplicate key insert.

ghost:
Resubmitted as #14654

PR #14608, opened 2025-08-08
avoid Take(0) when Limit == 0

evan314159:
**Changes**
avoid Take(0) when Limit == 0

**Issues**
Limit == 0 is defined to mean "return all items".  In statements like:
```
if (query.Limit.HasValue) { items = items.Take(query.Limit.Value); }
```
if Limit != null then HasValue == true a call to Take(0).

### DIFF
--- a/Emby.Server.Implementations/Library/SearchEngine.cs
+++ b/Emby.Server.Implementations/Library/SearchEngine.cs
@@ -42,7 +42,7 @@ namespace Emby.Server.Implementations.Library
                 results = results.GetRange(query.StartIndex.Value, totalRecordCount - query.StartIndex.Value);
             }
 
-            if (query.Limit.HasValue)
+            if (query.Limit.HasValue && query.Limit.Value > 0)
             {
                 results = results.GetRange(0, Math.Min(query.Limit.Value, results.Count));
             }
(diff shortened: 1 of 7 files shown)


JPVenson (approved):

crobibero (approved):

nielsvanvelzen:
Is this a new behavior difference or regression compared to 10.10?

ghost:
> Is this a new behavior difference or regression compared to 10.10?

I did not check.  It was something I noticed while going through the code and decided to fix rather than leaving.

JPVenson:
@nielsvanvelzen even if this is "new" behavior it should always have been like this. Its the correct and expected behavior from that property

Bond-009 (approved):

nielsvanvelzen (dismissed):
Seems to be an API behavior difference. Marking as changes requested so it isn't accidentally merged into 10.11.

ghost:
From sampling uses of Take in 10.10.7, it does not seem to take care to avoid Take(0), so the API could behave differently in outcome if limit was set to 0.

ghost:
If the view is that this is an API change (even if correcting the result to match the documented behaviour) isn't 10.11 a better time to make API change than maybe 10.11.1?

nielsvanvelzen:
API changes are not added to patch releases either, so this would be 10.12 at the earliest.

PR #14609, opened 2025-08-08
Refactor Type in BaseItem from String to integer with FK support

icegood:
1. Replace internal mapping from types to classes from code to DB (less DB space and better search)
2. Remove obsolete types (program, tvchannel) and map appropriate classes to other already existing item kinds. 
3. Remove some extra indexes (too much of them)
4. REmoved one obsolete tests (we don't need to convert BaseItemKind directly from class name anymore, instead GetBaseItemKind should take in place)


nielsvanvelzen (changes requested):
Removing item types is something that should first be discussed in a meta issue and if we decide to remove them we'll need to keep them in the API for at the minimum 1 major release for compatibility.

Additionally, I personally believe we should never use integer enums and use explicit names instead (no need to add the entire namespace though).

icegood:
Hello @nielsvanvelzen . I didn't remove enum, so API with client part is rather not changed in this regard. My  concern was to remove string values from DB, therefore make queries faster (i believe), since comparison of strings in indexes is much slower than integers.  Instead, i removed ItemLookup as this job should be done by dictionary table in DB.

icegood:
Seems now i understand what's going on. My intention was to change DB side only. Since,
**Jellyfin.Database.Implementations.Entities.BaseItemEntity**
is a part of EF and it is for sure for DB side changes only, api should never ever use it in any if them calls.

Instead, 
**MediaBrowser.Controller.Entities.BaseItem**
should be used instead. And intepretation of it as data tranfer object  of it is correct: 

using BaseItemDto = MediaBrowser.Controller.Entities.BaseItem;

Without proper architectural changes given PR is not possible

PR #14617, opened 2025-08-09
Detect Forced/Default/SDH from embedded subtitle titles

yslpn:
**Changes**
- Reuse Emby.Naming NamingOptions flags to infer Forced/Default/SDH from embedded subtitle stream titles, matching the external subtitle filename logic.
- Match flags with word boundaries and keep explicit disposition flags authoritative; only complement missing flags.
- Handle the “hi” vs Hindi language edge case to avoid false HI detection when the stream language is hi/hin/hi-*/hindi.
- Add unit tests covering case-insensitivity, word boundaries, Hindi exception, and combinations with existing dispositions.
- Add project reference to Emby.Naming for shared flag definitions.

**Issues**
My feature request with detailed description:
https://features.jellyfin.org/posts/3404/intelligent-subtitle-type-detection-forced-full-sdh-from-embedded-track-titles

### DIFF
--- a/tests/Jellyfin.MediaEncoding.Tests/Probing/ProbeResultNormalizerTests.cs
+++ b/tests/Jellyfin.MediaEncoding.Tests/Probing/ProbeResultNormalizerTests.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Globalization;
 using System.IO;
+using System.Linq;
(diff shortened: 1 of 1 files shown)


yslpn:
@gnattu hello, could you please help me with this? For me it's killer feature

PR #14639, opened 2025-08-12
Change Spanish; Latin language code according to BCP 47 specification

a-collado:
Changed the language code of the Spanish; Latin to better accommodate the BCP 47 standard (`es-419` – Spanish appropriate for the Latin America and Caribbean region, using the [UN M.49 region code](https://en.wikipedia.org/wiki/UN_M.49)). `es-mx` only refers to the region of Mexico, while `es-419` refers to the Latin America regions as a whole.

**Changes**
Changed the language code in `Emby.Server.Implementations/Localization/iso6392.txt`

### DIFF
--- a/Emby.Server.Implementations/Localization/iso6392.txt
+++ b/Emby.Server.Implementations/Localization/iso6392.txt
@@ -402,7 +402,7 @@ sog|||Sogdian|sogdien
 som||so|Somali|somali
 son|||Songhai languages|songhai, langues
 sot||st|Sotho, Southern|sotho du Sud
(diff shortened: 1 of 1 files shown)


Shadowghost:
@dyphire since you did work on this, I'd appreciate your input.

dyphire:
> @dyphire since you did work on this, I'd appreciate your input.

In the language code supported by `NET`, `es-419` stands for Latin America, so LGTM.

PS: I missed some unfamiliar languages in that PR.

crobibero (approved):

PR #14648, opened 2025-08-14
Fix anamorphic video detection (#14640)

### LINKED ISSUE #14640: Anamorphic detection is inverted
### Description of the bug

Jellyfin's anamorphic flag is computed by:

```
// http://stackoverflow.com/questions/17353387/how-to-detect-anamorphic-video-with-ffprobe
stream.IsAnamorphic = string.Equals(streamInfo.SampleAspectRatio, "0:1", StringComparison.OrdinalIgnoreCase);
```

FFprobe reports sample_aspect_ratio=0:1 only when the ratio is unknown/unspecified, not when it's anamorphic. As a result, IsAnamorphic is false for actual anamorphic content (e.g. SAR 8:9) and true only when SAR is unknown.

### Reproduction steps

1. Add an NTSC DVD-sourced MKV with MPEG-2 video flagged:
- size 720x480
- SAR 8:9
- DAR 4:3

2. Let Jellyfin analyze the file.

In the item's Media Info, observe "Anamorphic: No."

Because Anamorphic is incorrectly marked "No," clients that use this info display the content in an incorrect aspect ratio.

The item's Media Info should read "Anamorphic: Yes."

Anamorphic should be true whenever SAR is known and not equal to 1:1.

### Jellyfin Server version

10.10.0+

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.10.7

### Environment

```markdown
- OS: Debian 12
- Virtualization: Docker in Proxmox LXC
```

### Jellyfin logs

```shell
n/a
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

GeneMarks:
**Changes**
Previously, `IsAnamorphic` was set to `true` only when Sample Aspect Ratio (SAR) was `"0:1"`, causing incorrect detection for valid anamorphic videos. Updated the detection logic to handle common SAR/DAR edge cases, following the approach described https://stackoverflow.com/questions/17353387/how-to-detect-anamorphic-video-with-ffprobe.

**Issues**
Fixes #14640

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -202,6 +202,7 @@
  - [Shoham Peller](https://github.com/spellr)
  - [theshoeshiner](https://github.com/theshoeshiner)
  - [TokerX](https://github.com/TokerX)
+ - [GeneMarks](https://github.com/GeneMarks)
 
 # Emby Contributors
 
--- a/MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
+++ b/MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
@@ -850,12 +850,36 @@ namespace MediaBrowser.MediaEncoding.Probing
                     }
                 }
 
-                // stream.IsAnamorphic = string.Equals(streamInfo.sample_aspect_ratio, "0:1", StringComparison.OrdinalIgnoreCase) ||
-                //    string.Equals(stream.AspectRatio, "2.35:1", StringComparison.OrdinalIgnoreCase) ||
-                //    string.Equals(stream.AspectRatio, "2.40:1", StringComparison.OrdinalIgnoreCase);
-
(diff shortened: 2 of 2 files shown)


Shadowghost (approved):

ghost:
After the change does metadata need to be replaced?

crobibero:
> After the change does metadata need to be replaced?

Yes, this would require a scan to fix

gnattu:
Late to the party but this logic is generally OK for most videos but still does not cover one tricky case: 3D video.  Currently the very popular(but actually not standardized) practice is to encode 3D frames side by side, which may store the SAR as per eye instead of per frame, which means for some videos like HSBS, the 1:1 SAR could also be a technically, anamorphic video. This is out of scope of this PR though as that case is extremely tricky as there is no way to tell such video from video metadata solely, and common 3D players require very special naming conventions for such videos to be correctly detected.

PR #14651, opened 2025-08-15
Run background ffmpeg tasks as ProcessPriorityClass.BelowNormal

ghost:
**Changes**
Run background ffmpeg/ffprobe tasks as ProcessPriorityClass.BelowNormal:

- Keyframe extraction
- Media info probing during library scans
- Audio normalization
- Image extraction

following example from TrickPlay.

**Issues**
None


Shadowghost (approved):

Bond-009 (approved):

gnattu:
This causes regression for #14901

Not every system does have the permission required to set the process priority to below normal and it would simply fail all background tasks there. 

For trickplay we do have a user-configurable option so that the affected users can fix this themselves but a hardcoded value like this is not user-fixable.

PR #14654, opened 2025-08-16
Fix duplicate people insertion

evan314159:
Prevent duplicate records in Peoples by checking for existing records before bulk-inserting into Peoples and PeopleBaseItemMap.
    
Implements migration to cleanup existing duplicate records:
- Consolidates duplicate People records by (Name, PersonType)
- Updates PeopleBaseItemMap to point to canonical People records
- Removes duplicate mappings and orphaned records
- Makes PersonType column NOT NULL with unique constraint
- Creates unique index on (Name, PersonType)


ghost:
I don't understand the CI/CD error re: unapplied migrations, would appreciate if one of the team could assist with what it is looking for.  The changes are to the indexes and making the PersonType column NOT NULL, which are covered by the migration.

ghost:
I have worked out the CI/CD error and will test and push an update.

Jellyfin.Server.Implementations/Item/PeopleRepository.cs
```diff
@@ -70,18 +73,107 @@ public void UpdatePeople(Guid itemId, IReadOnlyList<PersonInfo> people)
...
+            var nameTypePairs = peopleWithoutIds
+                .Select(x => new { x.Name, PersonType = x.Type.ToString() })
+                .ToArray();
+
+            var names = nameTypePairs.Select(x => x.Name).ToArray();
+            var existingPeople = context.Peoples
+                .Where(p => names.Contains(p.Name))
+                .Select(p => new { p.Id, p.Name, p.PersonType })
+                .ToList()
+                .Where(p => nameTypePairs.Any(pair => p.Name == pair.Name && p.PersonType == pair.PersonType))
```
JPVenson, line 90: Create a string key and compare that. Thats currently the only way to do complex comparisions until dotnet10 is released. Yes it does not use indexes but its still faster then doing it serverside.

See: https://github.com/jellyfin/jellyfin/blob/master/Jellyfin.Server.Implementations/Item/BaseItemRepository.cs#L630

Jellyfin.Server.Implementations/Item/PeopleRepository.cs
```diff
@@ -70,18 +73,107 @@ public void UpdatePeople(Guid itemId, IReadOnlyList<PersonInfo> people)
...
+            {
+                context.Peoples.AddRange(stillNeedIds.Select(person => new People
+                {
+                    Id = person.Id,
+                    Name = person.Name,
+                    PersonType = person.Type.ToString()
+                }));
+                context.SaveChanges();
+            }
+            catch (DbUpdateException)
```
JPVenson: i would prefer if you use polly for that to make this a bit more readable

Jellyfin.Server.Implementations/Item/PeopleRepository.cs
ghost: I wonder if this comment may have belonged on the ItemValues change.

This code in Peoples is a simple SQL WHERE Name IN (Names) AND PersonType IN (PersonTypes) query that would be indexed.  The number of search results from a query like this is nearly zero so whatever the server is doing is very minimal.  I do not understand what changing this is meant to achieve.

Jellyfin.Server.Implementations/Item/PeopleRepository.cs
```diff
@@ -69,20 +73,130 @@ public IReadOnlyList<string> GetPeopleNames(InternalPeopleQuery filter)
...
+            var nameTypePairs = peopleWithoutIds
+                .Select(x => new { x.Name, PersonType = x.Type.ToString() })
+                .ToArray();
+
+            var names = nameTypePairs.Select(x => x.Name).ToArray();
+            var personTypes = nameTypePairs.Select(x => x.PersonType).ToArray();
+            var existingPeople = context.Peoples.AsNoTracking()
+                .Where(p => names.Contains(p.Name) && personTypes.Contains(p.PersonType))
+                .Select(p => new { p.Id, p.Name, p.PersonType })
+                .ToList()
```
crobibero:
```suggestion
                .AsEnumerable()
```

No reason to create a full list to just filter further

Jellyfin.Server.Implementations/Item/PeopleRepository.cs
```diff
@@ -69,20 +73,130 @@ public IReadOnlyList<string> GetPeopleNames(InternalPeopleQuery filter)
...
+                if (existing != null)
+                {
+                    person.Id = existing.Id;
+                }
+            }
+        }
 
-        var maps = context.PeopleBaseItemMap.Where(e => e.ItemId == itemId).ToList();
+        // Step 3: For people that are truly missing, assign new GUIDs and insert with Polly retry
+        var missingPeople = people.Where(p => p.Id == Guid.Empty).ToList();
```
crobibero:
```suggestion
        var missingPeople = people.Where(p => p.Id.IsEmpty()).ToList();
```

Jellyfin.Server.Implementations/Item/PeopleRepository.cs
```diff
@@ -69,20 +73,130 @@ public IReadOnlyList<string> GetPeopleNames(InternalPeopleQuery filter)
...
+                    context.ChangeTracker.Clear();
+
+                    // Re-check what already exists (handles concurrent inserts during retries)
+                    // Use AsNoTracking to avoid tracking conflicts from existence queries
+                    var namesToCheck = missingPeople.Select(p => p.Name).ToArray();
+                    var personTypesToCheck = missingPeople.Select(p => p.Type.ToString()).ToArray();
+                    var existingPeople = context.Peoples.AsNoTracking()
+                        .Where(p => namesToCheck.Contains(p.Name) && personTypesToCheck.Contains(p.PersonType))
+                        .Select(p => new { p.Id, p.Name, p.PersonType })
+                        .ToList()
```
crobibero:
```suggestion
                        .AsEnumerable()
```

Jellyfin.Server/Migrations/Routines/MigrateLibraryDb.cs
```diff
@@ -504,13 +504,9 @@ private People GetPerson(SqliteDataReader reader)
         {
             Id = Guid.NewGuid(),
             Name = reader.GetString(1),
+            PersonType = reader.GetString(3)
```
crobibero: I think this should be reverted and set PersonType to Unknown if we aren't able to pull the string out

Jellyfin.Server.Implementations/Item/PeopleRepository.cs
```diff
@@ -69,20 +73,130 @@ public IReadOnlyList<string> GetPeopleNames(InternalPeopleQuery filter)
...
+        using var transaction = context.Database.BeginTransaction();
 
-        // TODO: yes for __SOME__ reason there can be duplicates.
-        people = people.DistinctBy(e => e.Id).ToArray();
-        var personids = people.Select(f => f.Id);
-        var existingPersons = context.Peoples.Where(p => personids.Contains(p.Id)).Select(f => f.Id).ToArray();
-        context.Peoples.AddRange(people.Where(e => !existingPersons.Contains(e.Id)).Select(Map));
-        context.SaveChanges();
+        // Step 1: Get people that need IDs
+        var peopleWithoutIds = people.Where(p => p.Id == Guid.Empty).ToArray();
```
crobibero:
```suggestion
        var peopleWithoutIds = people.Where(p => p.Id.IsEmpty()).ToArray();
```

Jellyfin.Server/Migrations/Routines/20250814000002_RemoveDuplicatePeopleStage2.cs
crobibero, line 1: I may be mistaken but I don't think we have a way to run a code migration in between EF migrations

We also don't name routines with a timestamp in the name

JPVenson: The new Migration service allows mixing code and EFCore migrations.
For the naming you are correct.

Jellyfin.Server/Migrations/Routines/MigrateLibraryDb.cs
crobibero: This was missed on your last pass.

JPVenson:
this also get fixed in #14795 by reintroducing a simple transaction over the People update operation. I honestly would prefer doing it that way as its more idomatic for handing this issue (and should also handle it way more performant with pgsql ;-) as that dbms can introduce a rangelock instead)

ghost:
PR #14795 is huge, it is not obvious to me how it fixes the issues addressed here, and it is not reasonable for me to review it all to understand which parts of this change are fixed from that change.  If you want me to continue on this change please provide guidance.

There are different parts to this change:

1. Prevent duplicates by changing index to unique constraint and PersonType to non-nullable
2. Cleanup the duplicates
3. Protect Peoples update with a transaction
4. Check before inserting people, and then insert in bulk -- see // TODO: yes for __SOME__ reason there can be duplicates, because the existing code is avoiding duplicates in the wrong place
5. Retry inserts on unique constraint failures

Again without determining what #14795 does, 1-4 are desirable notwithstanding anything else.

5 is required for Postgres due to different consistency properties vs. Sqlite.  Notwithstanding the transaction, Postgres will create a few unique constraint violations on every library import that need to be retried.  According to Postgres documentation, even with a serialised transaction we still need to account for transaction errors due to duplicate insert attempts using some kind of retry.

I see these options:

1. Polly Retry: The code is already there.  It doesn't bother Sqlite and helps Postgres.  Polly is used elsewhere in the code.
2. INSERT ON CONFLICT UPDATE RETURNING and the equivalent Sqlite command: EF Core does not support INSERT ON CONFLICT.  If we don't want Polly retries, we can implement an extension on the database provider and JellyfinDbContext to support INSERT ON CONFLICT.  That may be more efficient but for the number of times that this is actually required, I think it is overkill.
3. Do not support Postgres: Realising that Postgres is not supported today, but is tracked to be mainline-supported in future, this would be disappointing when the code is already there and doesn't bother Sqlite.

Maybe you see other options as well.  The same issue and options would also apply to the duplicate ItemValues PR #14656.  It also generates PK or unique constraint violations on Postgres.

Please advise which approach you want.

JPVenson:
The pr just reintroduces a transaction for handling people operations, that should be the correct fix for now. We do not accept PRs that are only tailored towards pgsql for now when the solution to the issue can be a 2 line fix and it only works for sqlite.

ghost:
I'm not worried about the Postgres support but that was only one issue that this PR fixed i.e. #1-4 were issues on Sqlite too.  If your PR now doesn't generate duplicates and fixes the duplicates that were already created then this PR is not required and can be closed.

ghost:
#14795 still results in duplicate insertions into Peoples -- on a fresh install I had almost 260,000 duplicate entries in Peoples when it should only have around 9000 total records -- and does not cleanup previous duplicate insertions, so the PR is still relevant.

Rebased the change against #14795.  UpdatePeople in particular is identical with the Polly retry removed.  Added two new cases to the Stage 2 migration code:
- Step 2 added mappingsToCreate to create missing PeopleBaseItemMaps
- Step 5 remove orphaned People records (those with no BaseItem mappings)

These were picked up with the test cases that I had previously uploaded into one of the above comment -- attaching again [test_people_baseitems_consistency_sqlite.sql.txt](https://github.com/user-attachments/files/22448588/test_people_baseitems_consistency_sqlite.sql.txt).

Removing Polly does result in duplicate insert errors on Postgres however the data integrity on Postgres seems to be preserved i.e. the result passes the test cases.

Shadowghost:
> https://github.com/jellyfin/jellyfin/pull/14795 still results in duplicate insertions into Peoples -- on a fresh install I had almost 260,000 duplicate entries in Peoples when it should only have around 9000 total records -- and does not cleanup previous duplicate insertions, so the PR is still relevant.

Did you check on sqlite and/or pgsql?

ghost:
> > #14795 still results in duplicate insertions into Peoples -- on a fresh install I had almost 260,000 duplicate entries in Peoples when it should only have around 9000 total records -- and does not cleanup previous duplicate insertions, so the PR is still relevant.
> 
> Did you check on sqlite and/or pgsql?

Everything here was tested on SQLite, and the test script is for SQLite, which you can run on your own system before/after.  I also did the same testing with Postgres with the same result, other than that the number of database objects can vary between deployments.

Jellyfin.Server.Implementations/Item/PeopleRepository.cs
```diff
@@ -68,22 +68,89 @@ public IReadOnlyList<string> GetPeopleNames(InternalPeopleQuery filter)
...
+            .Where(m => !allPeopleIds.Contains(m.PeopleId))
+            .ToList();
+        context.PeopleBaseItemMap.RemoveRange(obsoleteMappings);
+
+        // Upsert mappings for all people
         foreach (var person in people)
         {
-            var existingMap = maps.FirstOrDefault(e => e.PeopleId == person.Id);
-            if (existingMap is null)
+            if (existingMappings.TryGetValue(person.Id, out var existing))
```
crobibero, line 144: I'm on mobile so I may have missed it, but I don't see where existingMappings is used below? Is there an entity update call missing?

JPVenson:
So after an extensive debugging session with @Shadowghost we came to the conclusion that the approach of relying on the Id for mapping is fundamentially flawed. I opened another Pr #14848 that handles deduplication and person reference mapping solely based on the Name/Role/Type values instead.

crobibero:
People de-duplication is handled in https://github.com/jellyfin/jellyfin/pull/14848

This doesn't strictly prevent it but cleans up during validation.

PR #14655, opened 2025-08-16
Correct Album Artists merge logic

evan314159:
**Changes**
Correct Album Artists merge logic in MetadataService that causes empty
metadata sources to overwrite populated Album Artists arrays. This impacted
People-to-BaseItem relationships and caused orphaned records in Peoples.

Correct equality checks to be case-sensitive so Jelly metadata exactly
matches file metadata.

**Issues**
This impacted People-to-BaseItem relationships and caused orphaned records in Peoples.

File metadata changes that varied only in case were not updated in Jellyfin, causing Firstname Lastname and firstname lastname to be 2 separate records because the metadata update was case-insensitive but people record names are case-sensitive.

### DIFF
--- a/MediaBrowser.Providers/Manager/MetadataService.cs
+++ b/MediaBrowser.Providers/Manager/MetadataService.cs
@@ -1268,7 +1268,7 @@ namespace MediaBrowser.Providers.Manager
                 {
                     targetHasAlbumArtist.AlbumArtists = sourceHasAlbumArtist.AlbumArtists;
                 }
-                else if (sourceHasAlbumArtist.AlbumArtists.Count > 0)
+                else
                 {
                     targetHasAlbumArtist.AlbumArtists = targetHasAlbumArtist.AlbumArtists.Concat(sourceHasAlbumArtist.AlbumArtists).Distinct(StringComparer.OrdinalIgnoreCase).ToArray();
                 }
--- a/MediaBrowser.Providers/Manager/MetadataService.cs
+++ b/MediaBrowser.Providers/Manager/MetadataService.cs
@@ -1268,7 +1268,7 @@ namespace MediaBrowser.Providers.Manager
                 {
                     targetHasAlbumArtist.AlbumArtists = sourceHasAlbumArtist.AlbumArtists;
                 }
-                else
+                else if (sourceHasAlbumArtist.AlbumArtists.Count > 0)
                 {
                     targetHasAlbumArtist.AlbumArtists = targetHasAlbumArtist.AlbumArtists.Concat(sourceHasAlbumArtist.AlbumArtists).Distinct(StringComparer.OrdinalIgnoreCase).ToArray();
                 }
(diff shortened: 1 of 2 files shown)


Shadowghost (approved):

ghost:
I found & removed some existing code relating to merging Album Artist.  The condition in the code is very similar but not the same, maybe enough to not work as intended.

crobibero:
The latest changes are causing tests to fail.

ghost:
I reviewed the tests, it is because it is expecting AlbumArtist is handled by the base MergeBaseItemData method instead of in the derived class.  This is inconsistent one way or another, because Artist is not handled there.  For clarity I think the right answer should be artist and album artist handled the same.  I am guessing that the path of least resistance is to move the code into the base, but it's awful OOP style having the base class take an interest in the subclass's data.

ghost:
Tests are successful now.

MediaBrowser.Providers/Manager/MetadataService.cs
```diff
@@ -1268,7 +1268,7 @@ private static void MergeAlbumArtist(BaseItem source, BaseItem target, bool repl
                 {
                     targetHasAlbumArtist.AlbumArtists = sourceHasAlbumArtist.AlbumArtists;
                 }
-                else if (sourceHasAlbumArtist.AlbumArtists.Count > 0)
+                else
```
crobibero: If the goal here is to call Distinct on the list of target album artists then a third branch should be added to avoid the Concat call

MediaBrowser.Providers/Music/AlbumMetadataService.cs
```diff
@@ -116,7 +116,7 @@ private ItemUpdateType SetAlbumArtistFromSongs(MusicAlbum item, IReadOnlyList<Au
 
         updateType |= SetProviderIdFromSongs(item, songs, MetadataProvider.MusicBrainzAlbumArtist);
 
-        if (!item.AlbumArtists.SequenceEqual(albumArtists, StringComparer.OrdinalIgnoreCase))
+        if (!item.AlbumArtists.SequenceEqual(albumArtists))
```
crobibero: Should specify `Ordinal` instead of removing the comparer

ghost:
Corrected to use Ordinal.

For the merge issue, it had been so long that I had to go back and find what the original problem is.  The fix I made resolved duplicate problems by using Distinct, but as far as I can tell the problem actually started earlier with unexpected duplicates at the audio file level.  I restored the original merge behaviour and corrected the artist GroupBy to be case-insensitive.  Now JF is case-preserving but case-insensitive which I think is as desired.  The end result remains that BaseItems and People are aligned even with some metadata differences in case.

Shadowghost (approved):

whyman:
@evan314159 just wanted to comment to say thanks for your efforts, I was looking at similar issues with album artists and you've already solved them 🙌🙌

crobibero (approved):

PR #14657, opened 2025-08-16
Avoid duplicate BaseItem & BaseItemProviders

evan314159:
**Changes**
Avoid duplicate BaseItems using UPSERT, similar to Peoples.

Avoid duplicate BaseItemProviders using by reordering operations to move provider deletion after Map() call to ensure delete and insert happen in same transaction.

Refactor BaseItem query that used string concatenation for indexability.
    
**Issues**
Addresses occasional duplicate BaseItem / BaseItemProviders.

Note: EF Core will log duplicate key errors (x3 for command error retry), however they are followed by a log showing that Jellyfin rematched the records.

### DIFF
--- a/Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
+++ b/Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
@@ -600,8 +600,8 @@ public sealed class BaseItemRepository
             }
             else
             {
-                context.BaseItemProviders.Where(e => e.ItemId == entity.Id).ExecuteDelete();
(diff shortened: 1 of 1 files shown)


crobibero (changes requested):
Changing to a model that saves for every item would absolutely kill scan performance, so we can't do that.

The optimization to the query can be extracted to a new PR

ghost:
Updated to convert to bulk insert with individual retry.

ghost:
This is now creating merge errors with the ItemValues change.  Closing this one and will include both in the ItemValues change.

PR #14674, opened 2025-08-19
fix ProbeProvider.HasChanged: if file date OR file size changed

ghost:
**Changes**
Media has changed if file date OR file size has changed

**Issues**
Fixes metadata not updating when file changed but size the same

### DIFF
--- a/MediaBrowser.Providers/MediaInfo/ProbeProvider.cs
+++ b/MediaBrowser.Providers/MediaInfo/ProbeProvider.cs
@@ -130,7 +130,7 @@ namespace MediaBrowser.Providers.MediaInfo
                 if (!string.IsNullOrWhiteSpace(path) && item.IsFileProtocol)
                 {
                     var file = directoryService.GetFile(path);
-                    if (file is not null && (item.HasChanged(file.LastWriteTimeUtc) || file.Length != item.Size))
+                    if (file is not null && item.HasChanged(file.LastWriteTimeUtc))
                     {
                         _logger.LogDebug("Refreshing {ItemPath} due to file system modification.", path);
                         return true;


Shadowghost:
We put his into place as a safeguard against miniscule differences. I don't mind changing it, wdyt @theguymadmax ?

ghost:
> We put his into place as a safeguard against miniscule differences. I don't mind changing it, wdyt @theguymadmax ?

I can understand where you are going with this; maybe the user did something like copied the media somewhere and back and now file modified dates are different.

I don’t think we should replace logical if undesired jellyfin behaviours (jellyfin rescans items because they look changed) with not logical behaviours (jellyfin doesn’t rescan items even though they look changed because you maybe didn’t really mean it, or maybe you did…).  For something like FLAC there are many kilobytes of padding around the metadata so the file would need to be repacked to get a file size change even with major metadata changes.

The check could be simplified to remove the file size comparison.

theguymadmax:
Here is the original discussion on this: https://github.com/jellyfin/jellyfin/pull/13839/files#r2036993988

I don't think the file size comparison is needed. Based on the conversation, it was added to address some user concerns.

ghost:
Removed file size check

Shadowghost (approved):

Bond-009 (approved):

tam1m:
Sorry for being late to this. I'm the one that brought up the point about the filesize in #13839.

I still hold on to the opinion that relying solely on timestamps is problematic, as this will nuke especially the trickplay files for users, which can take a long time to regenerate. I get that this is kind of a hacky workaround, but I feel like it's worth it in this case.

How do you all feel about a system setting (in settings.xml) that determines the behavior, so users can mitigate the problem if needed? I opend #14716 which implements this.

PR #14683, opened 2025-08-21
Fix false positive matching in specific regex expression.

arabcoders:
I considered removing the regex because it failed two cases it was originally meant to handle:

* `/server/anything_102.mp4` → Series: `anything`, Season: `1`, Episode: `02` (**intended match**)
* `/server/james.corden.2017.04.20.anne.hathaway.720p.hdtv.x264-crooks.mkv` → Series: `james.corden`, Season: `20`, Episode: `17` (**should not match, but does**)
* `/server/anything_1996.11.14.mp4` → Series: `anything`, Season: `19`, Episode: `96` (**should not match, but does**)

Removing the expression caused other tests to fail due to side effects. To maintain backward compatibility, I replaced it with a stricter pattern. The new regex still matches all formats it was originally intended to support (e.g., `anything_102`) and rejects unintended patterns such as `YYMMDD anything`.

### DIFF
--- a/Emby.Naming/Common/NamingOptions.cs
+++ b/Emby.Naming/Common/NamingOptions.cs
@@ -380,7 +380,7 @@ namespace Emby.Naming.Common
                 // /server/anything_102.mp4
                 // /server/james.corden.2017.04.20.anne.hathaway.720p.hdtv.x264-crooks.mkv
                 // /server/anything_1996.11.14.mp4
-                new EpisodeExpression(@"[\\/._ -](?<seriesname>(?![0-9]+[0-9][0-9])([^\\\/_])*)[\\\/._ -](?<seasonnumber>[0-9]+)(?<epnumber>[0-9][0-9](?:(?:[a-i]|\.[1-9])(?![0-9]))?)([._ -][^\\\/]*)$")
+                new EpisodeExpression(@"(?:^|[\\/])(?<seriesname>(?=[^\\/]*[A-Za-z])[^\\/_]+?(?:[._ -][^\\/_]+?)*)[._ -](?<!\d)(?<seasonnumber>\d{1,3})(?<epnumber>(?!00)\d{2})(?:(?:[a-i]|\.[1-9])(?!\d))?(?:[._ -][^\\/]*)*\.(?<ext>\w{3,4})$")
                 {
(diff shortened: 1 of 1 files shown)


crobibero:
We have tests specifically for this https://github.com/jellyfin/jellyfin/blob/master/tests/Jellyfin.Naming.Tests/TV/DailyEpisodeTests.cs

If we are updating a regex for NamingOptions then tests should be added to show why

Also, we use `[0-9]` instead of `\d` because it [matches any decimal digit](https://learn.microsoft.com/en-us/dotnet/standard/base-types/character-classes-in-regular-expressions#decimal-digit-character-d)

arabcoders:
> We have tests specifically for this https://github.com/jellyfin/jellyfin/blob/master/tests/Jellyfin.Naming.Tests/TV/DailyEpisodeTests.cs
> 
> If we are updating a regex for NamingOptions then tests should be added to show why

The functionality still the same and will match what the regex implied that it should be used for.

```
// /server/anything_102.mp4
// /server/james.corden.2017.04.20.anne.hathaway.720p.hdtv.x264-crooks.mkv
// /server/anything_1996.11.14.mp4
```

These test cases the regex was added for will still match however i just made it stricter so it doesn't eagerly match more than what it intended for. 

There is also [SimpleEpisodeTests.cs](https://github.com/jellyfin/jellyfin/blob/master/tests/Jellyfin.Naming.Tests/TV/SimpleEpisodeTests.cs) which i discovered also rely on the same behavior.

Just so i understand, do you want me to add a test case to assert that the regex will not match `/server/YYMMDD anything`?  i.e.

```C#
  [Theory]
        [InlineData("/server/250821 anything.mp4")]
        [InlineData("/server/20250821 anything.mp4")]
        public void ValidateNotMatch(string path)
        {
            var result = _resolver.Resolve(path, false);
            Assert.Null(result?.SeasonNumber);
            Assert.Null(result?.EpisodeNumber);
        }
```

> Also, we use `[0-9]` instead of `\d` because it [matches any decimal digit](https://learn.microsoft.com/en-us/dotnet/standard/base-types/character-classes-in-regular-expressions#decimal-digit-character-d)

Sure i'll update it right now.

PR #14685, opened 2025-08-22
Update artists and album artists during library scan if changed

evan314159:
**Changes**
Update artists and album artists during library scan if changed

**Issues**
Previously artists and album artists would not be updated during scans unless Replace All Metadata was set.


ghost:
Duplicates #14655

PR #14709, opened 2025-08-25
Fix: Add 'Kairon; IRSE!' to artist whitelist

loop95:
## Problem
Jellyfin incorrectly splits the artist **"Kairon; IRSE!"** into two separate artists:
- "Kairon"
- "IRSE!"

This happens because the semicolon is treated as a delimiter.

## Solution
Added `"Kairon; IRSE!"` to the whitelist in `ProbeResultNormalizer.cs`, so it is preserved as a single artist name.

### DIFF
--- a/MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
+++ b/MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
@@ -81,6 +81,7 @@ namespace MediaBrowser.MediaEncoding.Probing
             "Smith/Kotzen",
             "We;Na",
             "LSR/CITY",
+            "Kairon; IRSE!",
         };
 
         /// <summary>


gnattu:
I'm sorry but is this really needed for our new scanner where you can set the custom delimiter and a whitelist yourself?
What this PR changes is the old ffprobe based scanner which is only being used as a fallback for now. If a specific media file requires this, and this change is confirmed to fix the splitting, this usually means our main prober (the z440.atl) failed to read metadata from that specific file which is more preferably to be fixed by correcting the that prober instead.

loop95:
I tried changing the custom delimiter settings (adding the semicolon to delimiter whitelist) but it's not working on my end

gnattu:
> I tried changing the custom delimiter settings (adding the semicolon to delimiter whitelist) but it's not working on my end

No it does not work this way, you have to add full name `Kairon; IRSE!` in whitelist.

PR #14716, opened 2025-08-28
feat: add option to require file size change for media refreshes

tam1m:
This adds new system setting `FileChangeRequireSizeChange` that allows users to control how Jellyfin detects changes in media files. When enabled, the server will require both the modification time and the file size to have changed before triggering a metadata refresh.

This is based on the discussions in #13839 and #14674 and makes the behavior reintroduced in #14674 configurable.

With #14674, Jellyfin only checks the modification time of a file to determine if it has changed, which can lead to unnecessary metadata refreshes when a file's timestamp is updated without any actual change to its content. This is problematic for users with remote filesystems that may not provide stable timestamps, aswell as just copying files around without changing anything. 

I do recognize that this feels a bit hacky and would be fixed properly by hashing the files instead of relying on their timestsamps and size, but it's worth it in my opinion as this workaroud will prevent a lot of unnecessary rescans and especially uninented trickplay regeneration for a lot of users.

### DIFF
--- a/MediaBrowser.Controller/Extensions/ConfigurationExtensions.cs
+++ b/MediaBrowser.Controller/Extensions/ConfigurationExtensions.cs
@@ -75,6 +75,11 @@ namespace MediaBrowser.Controller.Extensions
         public const string DetectNetworkChangeKey = "DetectNetworkChange";
 
         /// <summary>
+        /// The key for the file change require size change option.
+        /// </summary>
+        public const string FileChangeRequireSizeChangeKey = "FileChangeRequireSizeChange";
+
+        /// <summary>
         /// Gets a value indicating whether the application should host static web content from the <see cref="IConfiguration"/>.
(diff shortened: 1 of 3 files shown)


ghost:
The Trickplay default could also be different (use keyframes only, which is much faster and could be the default on new deployments, or even could be changed during 10.11 upgrade for new files), or now that we have keyframe data maybe Trickplay could be generated on-the-fly.  Jellyfin could also not invalidate Trickplay on file changes by default (there are already scan settings to replace Trickplay), but that runs the risk that the Trickplay is out of sync.  Any of those options can be done by default.

An advanced setting to prevent Jellyfin from seeing file changes is unintuitive, and is not going to be enabled by most people, so does not seem helpful.

nielsvanvelzen:
API changes will not be merged to 10.11.

PR #14724, opened 2025-08-31
Skip non-media folders in movie resolver

### LINKED ISSUE #14646: Movies removed on intial scan after 10.11 upgrade
### Description of the bug

In an upgrade to 10.11 RC4/RC5 from 10.10.7, movies are removed on the initial scan, then re-added.  In the linked logs, I  created a new instance of 10.10.7 and added a single movie with trickplay enabled (trickplay enabled doesn't matter, was testing to see if trickplays would be removed).  

After upgrading to 10.11 (both RC4/RC5 had the same results), the movie was there, and I was able to view the movie with the trickplay. After performing an initial scan, the movie is removed from the library, then re-added, trickplays were also deleted because of this action. 

Full log from 10.10.7 creation to 10.11 RC5 upgrade:
https://gist.github.com/theguymadmax/66f479b0135e236b736f682f0008760a

### Reproduction steps

1. Upgrade from 10.10.7 to 10.11
2. Scan library 

Movies are removed after an initial scan after upgrading to 10.11 from 10.10.7, then re-added.

Movies are not removed after an initial scan after upgrading to 10.11 from 10.10.7

### Jellyfin Server version

Master

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

Tested on 10.11 RC4 & RC5

### Environment

```markdown
- OS: Windows & Ubuntu
- Virtualization: Docker on Ubuntu, exe on Windows
- Plugins: None
```

### Jellyfin logs

```shell
Linked in description
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

theguymadmax:
**Changes**
Skip processing of non-media directories during movie resolution:

- Extras folders 
- Trickplay folders
- Hidden folders
- Metadata folders

**Issues**
Fixes #14646


theguymadmax:
> Maybe we should put this into a more generic place and also apply it to e.g. Series/Seasons?

If I had looked a little more, I would have found that such a place already existed. Thanks!

Shadowghost (approved):

Bond-009:
Shouldn't we use `LibraryManager.IgnoreFile` instead?

theguymadmax:
> Shouldn't we use `LibraryManager.IgnoreFile` instead?

IgnorePatterns should have been correct. I ended up finding some additional issues and came up with a different solution.

crobibero (approved):

Issue #1215, opened 2019-04-08
Add support for two factor authentication


ConfusedPolarBear:
**Describe the feature you'd like**
Add support for two factor authentication. Ideally, both time based rolling codes (using the TOTP algorithm, compatible with Google Authenticator, andOTP, etc.) and a hardware based solution (possibly FIDO2 or WebAuthn) would be supported.

PR #14729, opened 2025-09-01
Refactor authentication

stenlan:
Draft pull request for maintainers to already see the structure of the changes that I want to make. 
Will need to fix tests, ~a circular dependency issue~ (done), ~model migrations~ (done), ~and in-code documentation~ (done). More documentation to follow. 

**Changes**
- Completely refactors authentication in a way that more cleanly separates the responsibilities of the session manager, user manager and the new `IUserAuthenticationManager`, and decouples it from the username/password type of authentication and instead makes it fully generic. Added some easy to use abstract building blocks for authentication providers;
  - `AbstractAuthenticationProvider` (used by username password auth)
  - `AbstractExternallyTriggeredAuthenticationProvider` (used by quick connect auth)
  - `AbstractChallengeResponseAuthenticationProvider` (WIP; can be used by general challenge-response type authentication providers).
- Adds TOTP MFA support to the back end
  -  Includes legacy client authentication fallback by *appending* TOTP token to password for compatibility with old clients. This is not unheard of, apple has done this too: https://discussions.apple.com/thread/254488819?sortBy=rank
  - Front end/web client support pending PR (TODO: add web client PR reference)).
- Changed `IAuthenticationProvider` interface to a generic `IAuthenticationProvider<T>` interface, where T is the type of payload data that an authentication provider takes upon authentication, so that multiple types of authentication can be supported.

**Issues**
Fixes [#1215](https://github.com/jellyfin/jellyfin/issues/1215).
Completes [this feature request](https://features.jellyfin.org/posts/26/add-support-for-two-factor-authentication-2fa)


v3DJG6GL:
Great to see this PR :)
Could this also be the basis for native integration for something like what this plugin currently does?
https://github.com/9p4/jellyfin-plugin-sso

stenlan:
> Great to see this PR :) Could this also be the basis for native integration for something like what this plugin currently does? https://github.com/9p4/jellyfin-plugin-sso

Yeah, I have taken special care to make the framework such that LDAP auth and that specific plugin will be easily implemented. It will be plug-and-play with the `AbstractAuthenticationProvider` and will in fact be so simple to implement that I will be making a PR to that repo myself once this is all merged.

JPVenson:
just as an FYI:

As i already told you in chat and others too:
1. A new Auth implementation should probably build upon ASP.NET Core Identity:
https://learn.microsoft.com/en-us/aspnet/core/security/authentication/?view=aspnetcore-9.0
2. We should really discuss this first in a MetaIssue before you start coding to avoid you spending time and it then just gets rejected because it does not align with the teams vision.
https://github.com/jellyfin/jellyfin-meta

stenlan:
> just as an FYI:
> 
> As i already told you in chat and others too:
> 
>   1. A new Auth implementation should probably build upon ASP.NET Core Identity:
>      https://learn.microsoft.com/en-us/aspnet/core/security/authentication/?view=aspnetcore-9.0

I've looked at ASP.NET Core Identity but I don't believe it's feasible to implement it since it's not just authentication but also authorization, which is deeply rooted throughout this entire application. I also don't think it's necessarily the way to go even if it were easier to switch to it, since it's quite opinionated and mostly geared towards web clients, while the Jellyfin backend serves not just web clients but all kinds of clients. My implementation provides a lot more flexibility and is not as opinionated.

>   2. We should really discuss this first in a MetaIssue before you start coding to avoid you spending time and it then just gets rejected because it does not align with the teams vision.
>   https://github.com/jellyfin/jellyfin-meta

I see, I was not aware of this meta issue board... I asked before if I should create an issue for it and you explicitly told me "thats not an issue as those are reserved for bugs". I will create an issue there anyway, but I have already spent the time...

stenlan:
Unfortunately I don't agree with your assessment so I can not contribute these changes to the project. Best of luck!

VslVictor7:
Indeed a shame, such a feature would be a massive + for jellyfin in general. I understand the team's vision on this, but jellyfin today can be hard to secure because of the risk of brute forcing in the login page. This recently happened with me.

I appreciate your effort @stenlan with this PR and implementation, since this is something a lot of people is asking for quite sometime now.

VslVictor7:
I think this PR should be at least disclosed for other people to see and possibly continue to work on this feature. Letting it closed would make this feel empty/not-taken-care-of

stenlan:
@VslVictor7 I'll probably release this as a plug-in instead not too far into the future.

stenlan:
You can follow the progress here https://github.com/stenlan/jellyguard

VslVictor7:
Very nice, thank you

cookie050:
Back to plex :P

TOTP Support is a must (for me)

PR #14736, opened 2025-09-02
Fix latest items grouping by collection type

### LINKED ISSUE #14735: [10.11.0] Grouped movies folders only display one recently added item
### Description of the bug

In 10.11, recent items display as expected in an ungrouped movie library.  However, when two or more movie library folders are grouped, only one item appears in the recent items section. This issue is specific to Movie libraries with groupedfolders.

Note: This is a 10.11.0-specific issue. It is not related to the previous issue in 10.10.x, where scanning bulk episodes in a Series library caused only one episode to appear. That issue has already been resolved in 10.11.

### Reproduction steps

1. On 10.11.0 have two movie libraries, then group them in the Home menu where it says: Automatically group content from the following folders..  

Only 1 item is displayed in the recently added section

There should be more than one item displayed

### Jellyfin Server version

Master

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.0 RC5

### Environment

```markdown
Irrelevant
```

### Jellyfin logs

```shell
None
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

theguymadmax:
**Changes**
Update `UserViewManager` to correctly detect collection types by checking both ICollectionFolder and UserView objects. This replaces the previous `All()`-based checks, which caused grouped movies to be misclassified as a tvshows collection type.

**Issues**
Fixes #14735

### DIFF
--- a/Emby.Server.Implementations/Library/UserViewManager.cs
+++ b/Emby.Server.Implementations/Library/UserViewManager.cs
@@ -381,7 +381,7 @@ namespace Emby.Server.Implementations.Library
                         UserView userView => userView.CollectionType,
                         _ => null
                     })
-                    .FirstOrDefault(type => type != null);
+                    .FirstOrDefault(type => type is not null);
 
                 if (collectionType == CollectionType.tvshows)
                 {


Emby.Server.Implementations/Library/UserViewManager.cs
```diff
@@ -374,13 +374,22 @@ private IReadOnlyList<BaseItem> GetItemsForLatestItems(User user, LatestItemsQue
...
             {
-                if (parents.OfType<ICollectionFolder>().All(i => i.CollectionType == CollectionType.tvshows))
+                var collectionType = parents
+                    .Select(parent => parent switch
+                    {
+                        ICollectionFolder collectionFolder => collectionFolder.CollectionType,
+                        UserView userView => userView.CollectionType,
+                        _ => null
+                    })
+                    .FirstOrDefault(type => type != null);
```
Bond-009:
```suggestion
                    .FirstOrDefault(type => type is not null);
```

JPVenson (approved):

Issue #14346, opened 2025-06-19
`NormalizationGain` ignores the `Track`/`Album` preference in Audio playback settings


nagisa:
### Description of the bug

As of recently the Jellyfin API has started including a `NormalizationGain` field in its responses about audio.

However this field appears to always return the value of `REPLAYGAIN_TRACK_GAIN` field, which the [implementation confirms](https://github.com/jellyfin/jellyfin/blob/cc7915c2e6a45cd0c040f8d38e0620594bb6a5a4/MediaBrowser.Providers/MediaInfo/AudioFileProber.cs#L348-L361), regardless of the setting in Playback section where a choice between track and album normalization can be made.

### Reproduction steps

1. Add a audio file with replay gain (different for track and album) tags to the library;
2. Go to settings and pick album normalization;
3. Make an `/Items/{id}` request for the audio file. 

### What is the current _bug_ behavior?

The `AudioNormalization` field unconditionally returns value for the track gain.

### What is the expected _correct_ behavior?

As there are options for both track and album normalization during playback, I'd expect this API field to somehow handle both types of normalization. Possibly by returning normalization values for all types of normalization supported so that clients can pick out the appropriate one, or by returning the value for the normalization type chosen in the settings.

### Jellyfin Server version

10.10.0+

### Specify version number

v10.10.7

### Specify the build version

Release

### Environment

```markdown
- OS: NixOS
- Linux Kernel: 6.12.26
- Virtualization: N/A
- Clients: Web, Tauon
- Browser: Firefox
- FFmpeg Version: 7.1.1
- Playback Method: N/A
- Hardware Acceleration: N/A
- GPU Model: N/A
- Plugins: N/A
- Reverse Proxy: Nginx
- Base URL: N/A
- Networking: N/A
- Jellyfin Data Storage: local SATA SSD
- Media Storage: local SATA SSD
- External Integrations: N/A
```

### Jellyfin logs

```shell
See Client/Browser logs
```

### FFmpeg logs

```shell
N/A
```

### Client / Browser logs

![Image](https://github.com/user-attachments/assets/8e79d5a8-27aa-4616-a7ea-e5272fc3f15a)

gnattu:
The album gain in the db is calculated by averaging all track gains currently.

nagisa:
That is a reasonable behaviour. Can this average be exposed in the API in some manner?

Another idea I had just now: maybe expose it as a NormalizationGain field on the `MusicAlbum` Item?

[...]

PR #14745, opened 2025-09-04
Add AlbumNormalizationGain field to BaseItemDto

HeroBrine1st:
**Changes**

This PR adds the `AlbumNormalizationGain` field to BaseItemDto which is inherited from music albums and available only on tracks with album:

```json
{
  "Name":"...",
  ...
  "NormalizationGain":-1.5,
  "AlbumNormalizationGain":-0.70000076
}
```

No new tests are written since there's no tests for existing track NormalizationGain. The change follows existing example and is tiny enough to be expected not to break.

**Issues**

Sorry I did that as a way to avoid waiting on a feature request (and there's nothing in contributing guidelines on doing feature requests). This does not fix anything in jellyfin itself but:

- #14346 can be fixed using this PR (that's actually jellyfin-web issue)
- Generally, any client wanting to use album gain must fetch album separately like we did at https://github.com/jmshrv/finamp/pull/1302 (and jellyfin-web does that too - see previous item). This allows for less round trips here

**Testing**

The tests were done manually with one caveat: I need to restart server before album normalization gain is returned to clients, both on rc5 and PR base with both this change applied and not (i.e. in total 4 test configurations). This means that `AlbumNormalizationGain` was not returned for tracks with this change applied as well as `NormalizationGain` was not returned for albums with or without this change applied before server restart. Note that I tested that on rootless docker and album count in library was 2, your mileage may wary.

To test, I simply looked at developer console (network tab) in browser while opening album or track.

Since there's no official packaging for jellyfin 10.11 provided, I used existing jellyfin-packaging repository and overridden dotnet version to 9.0:

<details>
<summary>Part of docker-compose.yml</summary>

```yaml
    build:
      context: ./jellyfin-packaging
      dockerfile: docker/Dockerfile
      args:
        DOTNET_ARCH: x64
        IMAGE_ARCH: amd64
        PACKAGE_ARCH: amd64
        QEMU_ARCH: x86_64
        TARGET_ARCH: amd64
        JELLYFIN_VERSION: "v10.11.0"
        CONFIG: Release
        DOTNET_VERSION: "9.0"
```

</details>

### DIFF
--- a/Emby.Server.Implementations/Dto/DtoService.cs
+++ b/Emby.Server.Implementations/Dto/DtoService.cs
@@ -1019,6 +1019,15 @@ namespace Emby.Server.Implementations.Dto
                 {
                     dto.AlbumId = albumParent.Id;
                     dto.AlbumPrimaryImageTag = GetTagAndFillBlurhash(dto, albumParent, ImageType.Primary);
+                    if (albumParent.LUFS.HasValue)
+                    {
+                        // -18 LUFS reference, same as ReplayGain 2.0, compatible with ReplayGain 1.0
+                        dto.AlbumNormalizationGain = -18f - albumParent.LUFS;
+                    }
+                    else if (albumParent.NormalizationGain.HasValue)
+                    {
+                        dto.AlbumNormalizationGain = albumParent.NormalizationGain;
+                    }
                 }
 
                 // if (options.ContainsField(ItemFields.MediaSourceCount))
--- a/MediaBrowser.Model/Dto/BaseItemDto.cs
+++ b/MediaBrowser.Model/Dto/BaseItemDto.cs
@@ -790,6 +790,12 @@ namespace MediaBrowser.Model.Dto
         public float? NormalizationGain { get; set; }
 
         /// <summary>
+        /// Gets or sets the gain required for audio normalization. This field is inherited from music album normalization gain.
+        /// </summary>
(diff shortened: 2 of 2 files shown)


HeroBrine1st:
I have not tested this rebase. There was no conflicts, I just updated it. I doubt it will break anyway

HeroBrine1st:
It's actually the only field that requires fetching album separately. It also reduces traffic usage and, most importantly, avoids waiting on a second request. Take Finamp as an example: if Jellyfin is slow to respond album request ("slow metadata fetcher" as I called it), user will hear wrong volume first, then it will be normalized to album. There's no mitigation aside of fetching albums before playing tracks, which was considered and dropped.

This is more important in case of big playlists. Instead of fetching a playlist and then all albums when you get two requests which turn out to two database queries where second one is a part of first (because fetching tracks already joins them to albums), Jellyfin can do db query just once, and respond with complete response suitable for playing right away.

So it's not a matter of bloating, it's a matter of the fact BaseItemDto already gives album info to application and there's no reason to require fetching album again so that it has one field. In the end, it's just more convenient to application like currently applications do not query for album cover and name.

HeroBrine1st:
Yeah, I'd prefer a polymorphic class, but hey - it can at least be made polymorphic in clients since available fields are determined by mediaType field.

Thank you guys too :-) Gonna land this into finamp in near time

Issue #14752, opened 2025-09-05
isPlayed filter is always applied


enter-a-random-username:
### Description of the bug

Building the current master and go into library I noticed all played items are gone. If I filter for unplayed the count stays the same and filtering for played is empty,

### Reproduction steps

1.  Build
2. go to a library
3. look for played

### What is the current _bug_ behavior?

Missing played entities

### What is the expected _correct_ behavior?

Show played and filter them only if unpklayed is selected.

### Jellyfin Server version

Master

### Specify commit id

20f7ddbf8fe9e06e366ad80484dd386542245c61

### Specify the build version

Release

### Environment

```markdown
- OS: linux
- Linux Kernel: 6.16
```

### Jellyfin logs

```shell
nothing, no new entries
```

### FFmpeg logs

```shell

```

### Additional information

Reverting da3f3b09d963b729c2d50002db035304a531efaf fixes the issue

theguymadmax:
@enter-a-random-username What type of library are you having this issue with, Series or Movies? And how many libraries do you have, and what are the types?

enter-a-random-username:
I have 8 library (all types, mixed, movies, music, series, collection) and the issue was with series based library (need to rebuild jellyfin to test the others types - reverted the commit). Compared the Parameters from GET in dev tools and isPlayed is only difference.

enter-a-random-username:
OK, tested it again

- series fails
- movies work
- mixed only shows movies and no series type
- Music has  no played filter

theguymadmax:
Thanks, that confirms the same behavior I'm seeing. 

It only happens when you have more than one series library set up in Jellyfin, ex: 
Anime
Tv Shows
Cartoons

If there is only one series library, the played filter, when applied, will display items correctly (although slow). 
When there are multiple series libraries, the played filter, when applied, will not return any items. 

If you group all the series libraries from the home menu, where it says "Automatically group content from the following folders into views such as 'Movies', 'Music' and 'TV'" then the is played filter will work. 

This is a regression from #14703

PR #14765, opened 2025-09-09
Get user data in bulk, fix IsPlayed filter

crobibero:
Fixes https://github.com/jellyfin/jellyfin/issues/14752

This unfortunately brings the IsPlayed/IsUnplayed query time back up to unacceptable times :/


src/Jellyfin.Extensions/GuidExtensions.cs
```diff
@@ -23,4 +25,28 @@ public static bool IsEmpty(this Guid guid)
...
         => guid is null || guid.Value.IsEmpty();
+
+    /// <summary>
+    /// Convert a guid to an upper case string.
+    /// </summary>
+    /// <param name="guid">The guid.</param>
+    /// <param name="format">The format to convert the guid to.</param>
+    /// <param name="formatProvider">The format provider.</param>
+    /// <returns>The uppercase string.</returns>
+    public static string ToUpper(this Guid guid, string format = "N", IFormatProvider? formatProvider = null)
```
Bond-009, line 36: This could use some tests

crobibero:
Getting UserItemData in bulk doesn't help on this level, closing

PR #14786, opened 2025-09-12
Fix IsPlayed

crobibero:
Fixes https://github.com/jellyfin/jellyfin/issues/14752


JPVenson:
this fixes the issue and #14795 fixes the new introduced performace issues.

JPVenson (approved):

PR #14794, opened 2025-09-13
Fix playlist move from smaller to larger index

### LINKED ISSUE #14792: When moving an item in a playlist to a position larger than the current one, the new position is off by one
### Description of the bug

Using the `/Playlists/{playlistId}/Items/{itemId}/Move/{newIndex}` endpoint when `newIndex` is nonzero results in the actual index being one greater. So, if `newIndex` would be 1, the actual result (0-based) would be 2.

### Reproduction steps

1. Have a playlist with at least 3 items.
2. In the web UI, attempt to drag the first item to the second position (call `/move/...`
3. Refresh the page. Note that the item is in fact in the third position

The problem is that the index that the song _appears_ to be in after moving is actually off by one. For example (using 0-indexed counting) dragging from 0->1 should swap 0 and 1, but it instead swaps 0 and 2.

Moving to `{newIndex}` should move the item currently in the playlist to right after what is **currently** at `{newIndex}`, and not what is at `{newIndex}` when the current item is removed.

### Jellyfin Server version

Unstable

### Specify commit id

_No response_

2025090805

### Specify version number

_No response_

### Specify the build version

2025090805

### Environment

```markdown
- OS: Arch
- Linux Kernel: 6.12.46-1-lts
- Virtualization: Docker
- Clients: Web UI
- Browser: Firefox
- FFmpeg Version:
- Playback Method: 
- Hardware Acceleration:
- GPU Model: 
- Plugins: none
- Reverse Proxy: None
- Base URL: http://localhost:8096
- Networking: `127.0.0.1:8096:8096`
- Jellyfin Data Storage: local ssd
- Media Storage: local ssd
- External Integrations: none
```

### Jellyfin logs

```shell
None relevant (just WS requests)
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

https://github.com/user-attachments/assets/b1a26cb2-22e7-4cbe-9cc9-1e4b21556789

In this example, I try to move cygnus (first item) after Orbit (second item). Instead, it moves after the third item (Galaxy2Galaxy)

### Additional information

The fix for https://github.com/jellyfin/jellyfin/issues/12892 appears to have an off by one when moving an item from a lower index to a higher one (e.g., 0 -> 1). The reverse path (e.g., 1 -> 0) appears to be fine.

kgarner7:
**Changes**
When attempting to use the `/move` API to move from a smaller index to a larger one, the actual value is one greater.
The fix for this is to always remove one from the index (except where it is already zero).
Before: `insertBeforeIndex = newIndex + 1 (from DetermineAdjustedIndex) + 1 (the item is removed from the list) = newIndex + 2`
After: `insertBeforeIndex = newINdex + 1 (from DetermineAdjustedIndex) - 1 (from this PR) + 1 (item is removed from list, things are offset by 1) = newIndex + 1`

**Issues**
Fixes #14792

### DIFF
--- a/Emby.Server.Implementations/Playlists/PlaylistManager.cs
+++ b/Emby.Server.Implementations/Playlists/PlaylistManager.cs
@@ -314,9 +314,7 @@ namespace Emby.Server.Implementations.Playlists
                 return;
             }
 
-            // Subtract when (newIndex > oldIndexAccessible), as the existing item is removed from the list.
-            // Otherwise, the index is now one greater than expected
-            var newPriorItemIndex = newIndex - 1 < 0 ? 0 : newIndex - 1;
+            var newPriorItemIndex = Math.Max(newIndex - 1, 0);
             var newPriorItemId = accessibleChildren[newPriorItemIndex].Item1.ItemId;
             var newPriorItemIndexOnAllChildren = children.FindIndex(c => c.Item1.ItemId.Equals(newPriorItemId));
             var adjustedNewIndex = DetermineAdjustedIndex(newPriorItemIndexOnAllChildren, newIndex);


Shadowghost (approved):

theguymadmax (approved):

JPVenson (approved):

Emby.Server.Implementations/Playlists/PlaylistManager.cs
```diff
@@ -314,7 +314,9 @@ public async Task MoveItemAsync(string playlistId, string entryId, int newIndex,
                 return;
             }
 
-            var newPriorItemIndex = newIndex > oldIndexAccessible ? newIndex : newIndex - 1 < 0 ? 0 : newIndex - 1;
+            // Subtract when (newIndex > oldIndexAccessible), as the existing item is removed from the list.
+            // Otherwise, the index is now one greater than expected
+            var newPriorItemIndex = newIndex - 1 < 0 ? 0 : newIndex - 1;
```
crobibero:
```suggestion
            var newPriorItemIndex = Math.Max(newIndex - 1, 0);
```

crobibero (approved):

PR #14791, opened 2025-09-13
Reenable common PRAGMA setters

JPVenson:

PR #14795, opened 2025-09-13
Temp/options and isplayed

JPVenson:
It somehow works like this

Builds upon #14791 and #14786 

- Cache a Folders "Children" instead of requerying them every time
- Improve internal querying to respect "Limit" parameter on querytime instead of applying it server side
- Query UserData with each BaseItem instead of requesting it individually
- Introduce dedicated "IsPlayed" query to recusively check a folder
- Reintroduce Transaction on Peoples to prevent FK collision
- Introduce FK for ParentId relation and cascade delete on it to ensure proper cleanup after library deletion

Squash pls


Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -434,7 +434,17 @@ private IQueryable<BaseItemEntity> PrepareItemQuery(JellyfinDbContext context, I
...
+            .Include(e => e.UserData);
+
+        // if (filter.Recursive)
+        // {
+        //     dbQuery = dbQuery.Include(e => e.DirectChildren)!
+        //         .ThenInclude(e => e.TrailerTypes)
+        //         .Include(e => e.Provider)
+        //         .Include(e => e.LockedFields)
+        //         .Include(e => e.UserData);
+        // }
```
Shadowghost: If unsued, remove

JPVenson: tell that to the original author of emby's code

MediaBrowser.Controller/Entities/Folder.cs
```diff
@@ -1700,16 +1700,18 @@ public override void MarkUnplayed(User user)
...
+            // var itemsResult = GetItemList(new InternalItemsQuery(user)
+            // {
+            //     Recursive = true,
+            //     IsFolder = false,
+            //     IsVirtualItem = false,
+            //     EnableTotalRecordCount = false
+            // });
+
+            // return itemsResult
+            //     .All(i => i.IsPlayed(user, userItemData: null));
```
Shadowghost: Remove if unneeded

Emby.Server.Implementations/Library/UserDataManager.cs
```diff
@@ -159,7 +160,12 @@ private UserData Map(UserItemData dto, Guid userId, Guid itemId)
             };
         }
 
-        private UserItemData Map(UserData dto)
+        /// <summary>
+        /// Maps a UserData entity to an UserItemData dto.
+        /// </summary>
+        /// <param name="dto">Entity.</param>
+        /// <returns>Dto.</returns>
```
crobibero:
```suggestion
```

Emby.Server.Implementations/Library/UserDataManager.cs
```diff
@@ -237,7 +243,10 @@ private static string GetCacheKey(long internalUserId, Guid itemId)
         /// <inheritdoc />
         public UserItemData? GetUserData(User user, BaseItem item)
         {
-            return GetUserData(user, item.Id, item.GetUserDataKeys());
+            return item.UserData.Where(e => e.UserId.Equals(user.Id)).Select(Map).FirstOrDefault() ?? new UserItemData()
```
crobibero, line 241: The original query pulls the UserData that matches the ItemId first and then falls back to a random key. It probably doesn't matter but thought I would point it out

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -434,7 +434,8 @@ private IQueryable<BaseItemEntity> PrepareItemQuery(JellyfinDbContext context, I
         dbQuery = dbQuery.AsSingleQuery()
             .Include(e => e.TrailerTypes)
             .Include(e => e.Provider)
-            .Include(e => e.LockedFields);
+            .Include(e => e.LockedFields)
+            .Include(e => e.UserData);
```
crobibero, line 442: Would it be possible to pull out the UserData grouped by UserId? We only ever care about UserData by a specific id so might be worth the optimization

MediaBrowser.Controller/Entities/Folder.cs
```diff
@@ -108,11 +110,18 @@ public override string FileNameWithoutExtension
...
         /// </summary>
         /// <value>The actual children.</value>
         [JsonIgnore]
-        public virtual IEnumerable<BaseItem> Children => LoadChildren();
+        public virtual IEnumerable<BaseItem> Children
+        {
+            get
+            {
+                return _children ?? (_children = LoadChildren());
+            }
```
crobibero:
```suggestion
            get => _children ??= LoadChildren();
```

crobibero:
Need to make sure to add to the release notes that watch history will be removed if a library is removed

Emby.Server.Implementations/Library/UserDataManager.cs
JPVenson: yea i dont think that matters. This is already matched by id first by being loaded as a foreignkey. In my testing that did not make any difference but i agree we should keep that in mind if issues arise.

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
JPVenson: Well not entirely, this is also loaded into the LibraryManagers cache where all users load their data from so we need to have all provided. I will see if thats a worthwhile optimisation but from my testing loading everything does not have any noticable performance impact.

JPVenson:
When the PR gets merged i will add it to the release docs

Emby.Server.Implementations/Library/UserDataManager.cs
JPVenson: yea so i double checked, the CustomDataKey is still a relict that was important in the pre-foregin key times where the connection between UserData and Item was not relational. Now its only important if we want to reconnect detached UserData. For the actual act of reading the data it makes little difference as long as we always provide _the same_ which we do via FirstOrDefault which coincidentially should always be the one thats the ID one anyway which mimics the behavior as before.

crobibero (approved):

JPVenson:
@crobibero note should no longer be nessesary, proper cleanup of dependend BaseItems is now added

crobibero (approved):

PR #14823, opened 2025-09-20
Skip removed images when updating images

Shadowghost:
It can happen that `item.ValidateImage()` which removes inaccessible images was not run before this method is run, so non-existing images could be probed here, leading to a lot of unneccessary log spam about not found images.

**Changes**
* check for file existence and skip if unavailable

**Remarks**
It might make sense to additionally call `item.ValidateImage()` at the end of the method just to be sure that images are in a clean state when returning.

### DIFF
--- a/Emby.Server.Implementations/Library/LibraryManager.cs
+++ b/Emby.Server.Implementations/Library/LibraryManager.cs
@@ -2032,6 +2032,12 @@ namespace Emby.Server.Implementations.Library
                     }
                 }
 
+                if (!File.Exists(image.Path))
+                {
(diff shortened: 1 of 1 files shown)


ghost:
I think this has broken image generation on root folders.  With -master on an existing database, deleting root folder images never re-creates the image.  On a new database, root folder images are never generated.

ghost:
Or could also be jellyfin/jellyfin#14962

ghost:
I found the issue -- apologies, it is in my own code but from long ago that somehow is now causing an issue.

Issue #14108, opened 2025-05-14
EF error when querying for HasAnyProviderId


crobibero:
TheTVDB plugin sees this issue
https://github.com/jellyfin/jellyfin-plugin-tvdb/blob/a8fbc5fd03e8742bd5d5c4ee86a5ad3729bf9820/Jellyfin.Plugin.Tvdb/ScheduledTasks/UpdateTask.cs#L155

Full stacktrace:
```
[ERR] [19] Emby.Server.Implementations.ScheduledTasks.TaskManager: Error executing Scheduled Task
System.InvalidOperationException: The LINQ expression '__filter_HasAnyProviderId_0
    .Any(e => StructuralTypeShaperExpression(
        StructuralType: Jellyfin.Database.Implementations.Entities.BaseItemProvider
        ValueBufferExpression: ProjectionBindingExpression: EmptyProjectionMember
        IsNullable: False).ProviderId == e.Key && StructuralTypeShaperExpression(
        StructuralType: Jellyfin.Database.Implementations.Entities.BaseItemProvider
        ValueBufferExpression: ProjectionBindingExpression: EmptyProjectionMember
        IsNullable: False).ProviderValue == e.Value)' could not be translated. Either rewrite the query in a form that can be translated, or switch to client evaluation explicitly by inserting a call to 'AsEnumerable', 'AsAsyncEnumerable', 'ToList', or 'ToListAsync'. See https://go.microsoft.com/fwlink/?linkid=2101038 for more information.
   at Microsoft.EntityFrameworkCore.Query.QueryableMethodTranslatingExpressionVisitor.Translate(Expression expression)
   at Microsoft.EntityFrameworkCore.Query.QueryableMethodTranslatingExpressionVisitor.TranslateSubquery(Expression expression)
   at Microsoft.EntityFrameworkCore.Query.RelationalSqlTranslatingExpressionVisitor.<VisitMethodCall>g__TranslateAsSubquery|33_0(Expression expression, <>c__DisplayClass33_0&)
   at Microsoft.EntityFrameworkCore.Query.RelationalSqlTranslatingExpressionVisitor.VisitMethodCall(MethodCallExpression methodCallExpression)
   at Microsoft.EntityFrameworkCore.Sqlite.Query.Internal.SqliteSqlTranslatingExpressionVisitor.VisitMethodCall(MethodCallExpression methodCallExpression)
[...]

PR #14831, opened 2025-09-21
fix: add back missing behavior for HasAnyProviderId

revam:
**Changes**

Follow up for #14249, adding back the missing behavior to query items with any value for a provider id set.

I noticed while testing the 10.11 RC 5 that the `HasAnyProviderId` usage with an empty value was broken compared to 10.10, where it previously allowed you to query for all base items with the the given providers set regardless of the actual ID. Tested that it works locally.

**Issues / PRs**
https://github.com/jellyfin/jellyfin/issues/14108
https://github.com/jellyfin/jellyfin/pull/14249

### DIFF
--- a/Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
+++ b/Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
@@ -2278,6 +2278,7 @@ public sealed class BaseItemRepository
 
         if (filter.HasAnyProviderId is not null && filter.HasAnyProviderId.Count > 0)
         {
+            // Allow setting a null or empty value to get all items that have the specified provider set.
             var includeAny = filter.HasAnyProviderId.Where(e => string.IsNullOrEmpty(e.Value)).Select(e => e.Key).ToArray();
             if (includeAny.Length > 0)
             {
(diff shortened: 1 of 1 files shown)


Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -2278,8 +2278,17 @@ private IQueryable<BaseItemEntity> TranslateQuery(
 
         if (filter.HasAnyProviderId is not null && filter.HasAnyProviderId.Count > 0)
         {
-            var include = filter.HasAnyProviderId.Select(e => $"{e.Key}:{e.Value}").ToArray();
-            baseQuery = baseQuery.Where(e => e.Provider!.Select(f => f.ProviderId + ":" + f.ProviderValue)!.Any(f => include.Contains(f)));
+            var includeAny = filter.HasAnyProviderId.Where(e => string.IsNullOrEmpty(e.Value)).Select(e => e.Key).ToArray();
```
JPVenson, line 2282: A comment with a description of why please

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -2278,8 +2278,17 @@ private IQueryable<BaseItemEntity> TranslateQuery(
...
+            if (includeAny.Length > 0)
+            {
+                baseQuery = baseQuery.Where(e => e.Provider!.Select(f => f.ProviderId)!.Any(f => includeAny.Contains(f)));
```
JPVenson: shorten to

```suggestion
                baseQuery = baseQuery.Where(e => e.Provider!.Any(f => includeAny.Contains(f.ProviderId)));
```

JPVenson (approved):

PR #14863, opened 2025-09-25
Fixes FK on unconnected base items

JPVenson:


Shadowghost (approved):

JPVenson:
If you have, your DB deserves to die anyway

PR #14864, opened 2025-09-25
Fix people deduplication lookup

JPVenson:


crobibero:
Doesn't this also need to happen in PeopleValidationTask?

JPVenson:
Nah i took it wrong too. The "GuestStar" is a PersonType not a Role and we handle the correct distinction there already

theguymadmax (approved):
Tested and upgrading to this version prevents the different PersonType from being lost during migration.  

A scan for missing metadata is needed if the server was already migrated to a prior RC build, before this PR.

Shadowghost (approved):

crobibero (approved):

PR #14878, opened 2025-09-26
Restore playlists after #14795

evan314159:
**Changes**
Restore playlists after #14795

**Issues**
After #14795 playlists return zero children


ghost:
Cannot duplicate on a clean SQLite database, closing.

PR #14879, opened 2025-09-27
Fix: normalize punctuation when computing CleanName so searches without punctuation match (closes #1674)

### LINKED ISSUE #1674: Search for series expects exact punctuation between words in title

audrey-inglish:
**Summary**
Normalize punctuation/symbols to spaces when computing CleanName (storage-time). This makes searches like "mr robot" match a video titled "Mr. Robot," as described in issue #1674. 

**Changes**
GetCleanValue now delegates to the new NormalizeNameForSearch helper method. This method treats punctuation and symbols (anything that isn't a letter or digit) as whitespace. Then, it strips out excess whitespace. 

I was able to reproduce the issue locally and successfully fail unit tests to demonstrate the behavior. After implementing the change and rescanning the library, the tests pass locally and the search behaves as expected.

**Issues**
Fixes #1674

### DIFF
--- a/Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
+++ b/Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
@@ -1362,6 +1362,16 @@ public sealed class BaseItemRepository
             return value;
         }
 
+        return NormalizeNameForSearch(value);
+    }
+
+    private static string NormalizeNameForSearch(string value)
+    {
+        if (string.IsNullOrWhiteSpace(value))
+        {
+            return value;
+        }
+
         var noDiacritics = value.RemoveDiacritics();
 
         // Build a string where any punctuation or symbol is treated as a separator (space).
--- a/Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
+++ b/Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
@@ -1362,7 +1362,27 @@ public sealed class BaseItemRepository
             return value;
         }
 
-        return value.RemoveDiacritics().ToLowerInvariant();
+        var noDiacritics = value.RemoveDiacritics();
+
+        // Build a string where any punctuation or symbol is treated as a separator (space).
+        var sb = new StringBuilder(noDiacritics.Length);
+        foreach (var ch in noDiacritics)
+        {
+            // Keep letters, digits and whitespace. Treat anything else as a space.
+            if (char.IsLetterOrDigit(ch) || char.IsWhiteSpace(ch))
+            {
+                sb.Append(ch);
+            }
+            else
+            {
+                sb.Append(' ');
+            }
+        }
+
+        // Collapse multiple spaces into single space and trim.
+        var collapsed = System.Text.RegularExpressions.Regex.Replace(sb.ToString(), "\\s+", " ").Trim();
+
+        return collapsed.ToLowerInvariant();
     }
 
     private List<(ItemValueType MagicNumber, string Value)> GetItemValuesToSave(BaseItemDto item, List<string> inheritedTags)
--- a/tests/Jellyfin.Server.Implementations.Tests/Data/SearchPunctuationTests.cs
+++ b/tests/Jellyfin.Server.Implementations.Tests/Data/SearchPunctuationTests.cs
@@ -39,7 +39,7 @@ namespace Jellyfin.Server.Implementations.Tests.Data
         }
 
         [Fact]
-        public void CleanName_keeps_punctuation_and_search_without_punctuation_fails()
+        public void CleanName_keeps_punctuation_and_search_without_punctuation_passes()
         {
             var series = new Series
             {
@@ -47,16 +47,41 @@ namespace Jellyfin.Server.Implementations.Tests.Data
                 Name = "Mr. Robot"
             };
 
-       series.SortName = "Mr. Robot";
+            series.SortName = "Mr. Robot";
 
             var entity = _repo.Map(series);
-
-            // Map sets CleanName using GetCleanValue (lowercases, removes diacritics but keeps punctuation)
-            Assert.Equal("mr. robot", entity.CleanName);
+            Assert.Equal("mr robot", entity.CleanName);
 
             var searchTerm = "Mr Robot".ToLowerInvariant();
 
-         Assert.Contains(searchTerm, entity.CleanName ?? string.Empty, StringComparison.OrdinalIgnoreCase);
+            Assert.Contains(searchTerm, entity.CleanName ?? string.Empty, StringComparison.OrdinalIgnoreCase);
+        }
+
+        [Theory]
+        [InlineData("Spider-Man: Homecoming", "spider man homecoming")]
+        [InlineData("Beyoncé — Live!", "beyonce live")]
+        [InlineData("Hello, World!", "hello world")]
+        [InlineData("(The) Good, the Bad & the Ugly", "the good the bad the ugly")]
+        [InlineData("Wall-E", "wall e")]
+        [InlineData("No. 1: The Beginning", "no 1 the beginning")]
+        [InlineData("Café-au-lait", "cafe au lait")]
+        public void CleanName_normalizes_various_punctuation(string title, string expectedClean)
+        {
+            var series = new Series
+            {
+                Id = Guid.NewGuid(),
+                Name = title
+            };
+
+            series.SortName = title;
+
+            var entity = _repo.Map(series);
+
+            Assert.Equal(expectedClean, entity.CleanName);
+
+            // Ensure a search term without punctuation would match
+            var searchTerm = expectedClean;
+            Assert.Contains(searchTerm, entity.CleanName ?? string.Empty, StringComparison.OrdinalIgnoreCase);
         }
     }
 }
--- a/tests/Jellyfin.Server.Implementations.Tests/Data/SearchPunctuationTests.cs
+++ b/tests/Jellyfin.Server.Implementations.Tests/Data/SearchPunctuationTests.cs
@@ -1,0 +1,62 @@
+using System;
+using AutoFixture;
+using AutoFixture.AutoMoq;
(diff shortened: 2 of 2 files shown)


Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -1362,7 +1362,37 @@ private string GetCleanValue(string value)
             return value;
         }
 
-        return value.RemoveDiacritics().ToLowerInvariant();
+        return NormalizeNameForSearch(value);
+    }
+
+    private static string NormalizeNameForSearch(string value)
```
JPVenson: i dont see a reason for adding this as its own dedicated method.
Just include that in the normalise method

JPVenson (changes requested):

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
audrey-inglish: Thanks for the feedback! I made the requested change.

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -1362,7 +1362,37 @@ private string GetCleanValue(string value)
...
+                sb.Append(ch);
+            }
+            else
+            {
+                sb.Append(' ');
+            }
+        }
+
+        // Collapse multiple spaces into single space and trim.
+        var collapsed = System.Text.RegularExpressions.Regex.Replace(sb.ToString(), "\\s+", " ").Trim();
```
theguymadmax: You could drop the Regex.Replace and handle space collapsing directly in the loop using a flag.

audrey-inglish: Good idea, thanks! I updated the loop to match what I think you're describing.

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -1362,7 +1362,42 @@ private string GetCleanValue(string value)
             return value;
         }
 
-        return value.RemoveDiacritics().ToLowerInvariant();
+        var noDiacritics = value.RemoveDiacritics();
+
+        // Build a string where any punctuation or symbol is treated as a separator (space).
+        var sb = new StringBuilder(noDiacritics.Length);
+        var previousWasSpace = false;
+        foreach (var ch in noDiacritics)
```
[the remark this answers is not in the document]
audrey-inglish: This is a lot cleaner. Thanks for the suggestion! I have made the change and verified the behavior.

theguymadmax:
@

theguymadmax:
@audrey-inglish Your PR has already been approved, so any changes would need to go through re-approval. I believe those suggestions came from an AI bot. I ran a comparison test on a small library, and the updated version takes about 4 minutes longer than the approved one. When scaled, this could result in a noticeable performance impact. I recommend reverting to the approved version.

Also, please note that there is currently a freeze on any new enhancements until the 10.11 release. As such, it's unlikely that this will be merged before then.

Shadowghost:
One question: Does this require a scan with replace all metadata? If that's the case a proper migration might be warranted to do this for existing items once
Just out of interest since I ran into it recently, how would V/H/S be searchable?

theguymadmax:
> One question: Does this require a scan with replace all metadata? If that's the case a proper migration might be warranted to do this for existing items once Just out of interest since I ran into it recently, how would V/H/S be searchable?

I still had this PR open so i tested it. A scan for missing metadata is required to generate an updated cleanname. Also, no, this doesn't handle "/" in titles. Another example, "Face/Off"

Shadowghost:
Ideally we'd handle this case too and then have a migration to fix it.

audrey-inglish:
@theguymadmax Thanks for explaining that, and thanks for testing out the performance. I have reverted back to the approved commit.

As for the notes from @Shadowghost , I've started looking into those changes. Handling slashes is already covered in the approved commit I reverted to. I added tests to verify the behavior. Searching for "face off" will yield a show titled "Face/Off."

As far as running a proper migration, I'm unsure on how to handle that correctly. Any direction you can give on this would be appreciated.

Thanks!

Jellyfin.Server/Migrations/Routines/RefreshCleanNames.cs
```diff
@@ -0,0 +1,151 @@
...
+
+        var sw = Stopwatch.StartNew();
+
+        using var context = _dbProvider.CreateDbContext();
+        var totalRecords = context.BaseItems.Count(b => !string.IsNullOrEmpty(b.Name));
+        _logger.LogInformation("Refreshing CleanName for {Count} library items", totalRecords);
+
+        do
+        {
+            var batch = context.BaseItems
```
JPVenson: you can use the pagination extensions for that

Jellyfin.Server/Migrations/Routines/RefreshCleanNames.cs
```diff
@@ -0,0 +1,151 @@
...
+            sw.Elapsed);
+    }
+
+    /// <summary>
+    /// Gets the clean value for search and sorting purposes.
+    /// This is a copy of the GetCleanValue logic from BaseItemRepository.
+    /// </summary>
+    /// <param name="value">The value to clean.</param>
+    /// <returns>The cleaned value.</returns>
+    private static string? GetCleanValue(string? value)
```
JPVenson: can't we reuse the method for that?

Jellyfin.Server/Migrations/Routines/RefreshCleanNames.cs
```diff
@@ -0,0 +1,151 @@
...
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.Logging;
+
+namespace Jellyfin.Server.Migrations.Routines;
+
+/// <summary>
+/// Migration to refresh CleanName values for all library items.
+/// </summary>
+[JellyfinMigration("2025-10-08T12:00:00", nameof(RefreshCleanNames))]
+public class RefreshCleanNames : IDatabaseMigrationRoutine
```
JPVenson: Use the IAsyncMigration interface instead and use the Backup attribute with the JellyfinDb set to true

Jellyfin.Server/Migrations/Routines/RefreshCleanNames.cs
audrey-inglish: I updated the GetCleanValue method to be `public static` in order to use it in the migration. Is this appropriate?

Jellyfin.Server/Migrations/Routines/RefreshCleanNames.cs
audrey-inglish: I'm a bit unclear on this, so I'd appreciate it if you could take a look at the change I made for this.

JPVenson: you can look the pattern up in
https://github.com/jellyfin/jellyfin/blob/master/Jellyfin.Server/Migrations/Routines/MoveExtractedFiles.cs#L82

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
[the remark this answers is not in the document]
JPVenson: that is no bot affiliated or enabled by us, i have no idea what its doing here commenting on this PR please ignore

Jellyfin.Server/Migrations/Routines/RefreshCleanNames.cs
audrey-inglish: @JPVenson Thank you for the reference. I've done my best to replicate this pattern. Please let me know if any other changes need to be made!

Jellyfin.Server/Migrations/Routines/RefreshCleanNames.cs
```diff
@@ -0,0 +1,105 @@
...
+                          .OrderBy(e => e.Id)
+                          .WithPartitionProgress((partition) => _logger.LogInformation("Processed: {Offset}/{Total} - Updated: {UpdatedCount} - Time: {Elapsed}", partition * Limit, records, itemCount, sw.Elapsed))
+                          .PartitionEagerAsync(Limit, cancellationToken)
+                          .WithCancellation(cancellationToken)
+                          .ConfigureAwait(false))
+        {
+            try
+            {
+                var newCleanName = BaseItemRepository.GetCleanValue(item.Name ?? string.Empty);
+                if (newCleanName != item.CleanName)
```
JPVenson:
```suggestion
                if (!string.Equals(newCleanName, item.CleanName, StringComparison.Ordinal))
```

Jellyfin.Server/Migrations/Routines/RefreshCleanNames.cs
```diff
@@ -0,0 +1,105 @@
...
+                          .Where(b => !string.IsNullOrEmpty(b.Name))
+                          .OrderBy(e => e.Id)
+                          .WithPartitionProgress((partition) => _logger.LogInformation("Processed: {Offset}/{Total} - Updated: {UpdatedCount} - Time: {Elapsed}", partition * Limit, records, itemCount, sw.Elapsed))
+                          .PartitionEagerAsync(Limit, cancellationToken)
+                          .WithCancellation(cancellationToken)
+                          .ConfigureAwait(false))
+        {
+            try
+            {
+                var newCleanName = BaseItemRepository.GetCleanValue(item.Name ?? string.Empty);
```
JPVenson:
```suggestion
                var newCleanName = string.IsNullOrWhiteSpace(item.Name) ? string.Empty : BaseItemRepository.GetCleanValue(item.Name);
```

JPVenson (approved):

soultaco83:
I've had this pr installed for awhile. I have noticed no issues with this that I can tell. So far clean names and standard names work well. 
Prime example of Mr. Robot was always a pain point for one of my friends.

PR #14882, opened 2025-09-27
Comment code that intentionally generates MultipleCollectionIncludeWarning

evan314159:
This code generates MultipleCollectionIncludeWarning.  Verified with JPVenson that it is intended to be a single query.

Warnings can only be disabled globally so adding a comment for future developers.


nomener:
According to https://learn.microsoft.com/ef/core/querying/single-split-queries it should also be possible to silence the warning by adding `AsSingleQuery()` to the call (which is the default behavior if no other option in given).

PR #14907, opened 2025-09-29
Prometheus metrics

ysf3rdm:
**Changes**

This only implements user account metrics for now as outlined in #3016. This PR adds a new `UserMetrics` class that tracks user accounts, permissions, authentication providers, activity, and security metrics.

**Implemented Metrics:**

- [x] `jellyfin_user_accounts_total` - Total number of user accounts
- [x] `jellyfin_user_accounts_active` - Number of active (non-disabled) accounts
- [x] `jellyfin_user_accounts_admin` - Number of administrator accounts
- [x] `jellyfin_users_by_permission` - Users with specific permissions (labeled by permission type)
- [x] `jellyfin_users_by_auth_provider` - Users by authentication provider (labeled by provider)
- [x] `jellyfin_users_external_auth_total` - Total users using external authentication
- [x] `jellyfin_users_recently_active` - Users active within last 30 days
- [x] `jellyfin_users_failed_logins` - Users with failed login attempts (labeled by attempt range)

**Planned Metrics (from #3016):**

- [ ] Session metrics (active sessions, device types, client types)
- [ ] Stream metrics (play methods, transcoding, codecs)
- [ ] Library metrics (library counts, media types, codec distribution)

**Questions for Reviewers**

1. **Label Cardinality**: Currently using `AuthenticationProviderId` as a dynamic label for the auth provider metric. In typical Jellyfin installations, how many distinct auth providers exist? Should this be categorized into "Default" vs "External" instead to prevent unbounded cardinality growth?

2. **Performance**: Currently loads all users with `ToList()` for metric calculations. For large instances (10K+ users), is this acceptable or should it use projections/streaming?

3. **Architecture**: Should the remaining metrics (sessions, streams, libraries) be in separate classes following the same pattern, or would you prefer a different organization?

4. **Additional Metrics**: Are there any other user-related or Jellyfin-specific metrics you'd like to see added? Please suggest in the comments.

**Issues**

#3016

### DIFF
--- a/Jellyfin.Server.Implementations/Metrics/UserMetrics.cs
+++ b/Jellyfin.Server.Implementations/Metrics/UserMetrics.cs
@@ -1,7 +1,6 @@
 #pragma warning disable CA1307
 
 using System;
-using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using Jellyfin.Data.Entities;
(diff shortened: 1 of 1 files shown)


JPVenson:
please retarget master, no more features are accepted in 10.10

PR #14916, opened 2025-09-30
Podman Compose File

Marchosias140:
docker-compose.yml file suitable to run with podman-compose.

Fixes out-of-the-box incompatibilities between the original Docker-oriented compose file with Podman.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,13 @@
 </a>
 </p>
 
+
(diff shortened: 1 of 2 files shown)


JPVenson:
Deployment help documentation is handled by our docs
https://github.com/jellyfin/jellyfin.org

PR #14919, opened 2025-10-01
Fix BoxSet sorting

theguymadmax:
**Changes**
Sort BoxSet items by production year in chronological order

**Issues**
Fixes sorting regression introduced in #14795

### DIFF
--- a/MediaBrowser.Controller/Entities/Folder.cs
+++ b/MediaBrowser.Controller/Entities/Folder.cs
@@ -1346,6 +1346,14 @@ namespace MediaBrowser.Controller.Entities
             var realChildren = visibleChildren
                 .Where(e => query is null || UserViewBuilder.FilterItem(e, query))
                 .ToArray();
+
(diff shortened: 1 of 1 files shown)


Shadowghost (approved):

MediaBrowser.Controller/Entities/Folder.cs
```diff
@@ -1346,6 +1346,14 @@ private int AddChildrenFromCollection(IEnumerable<BaseItem> children, User user,
             var realChildren = visibleChildren
                 .Where(e => query is null || UserViewBuilder.FilterItem(e, query))
                 .ToArray();
+
+            if (this is BoxSet && (query.OrderBy == null || query.OrderBy.Count == 0))
```
crobibero:
```suggestion
            if (this is BoxSet && (query.OrderBy is null || query.OrderBy.Count == 0))
```

PR #14920, opened 2025-10-01
Prometheus metrics

ysf3rdm:
**Changes**

This only implements user account metrics for now as outlined in #3016. This PR adds a new `UserMetrics` class that tracks user accounts, permissions, authentication providers, activity, and security metrics.

**Implemented Metrics:**

- [x] `jellyfin_user_accounts_total` - Total number of user accounts
- [x] `jellyfin_user_accounts_active` - Number of active (non-disabled) accounts
- [x] `jellyfin_user_accounts_admin` - Number of administrator accounts
- [x] `jellyfin_users_by_permission` - Users with specific permissions (labeled by permission type)
- [x] `jellyfin_users_by_auth_provider` - Users by authentication provider (labeled by provider)
- [x] `jellyfin_users_external_auth_total` - Total users using external authentication
- [x] `jellyfin_users_recently_active` - Users active within last 30 days
- [x] `jellyfin_users_failed_logins` - Users with failed login attempts (labeled by attempt range)

**Planned Metrics (from #3016):**

- [x] Session metrics (active sessions, device types, client types)
- [x] Stream metrics (play methods, transcoding, codecs)
- [ ] Library metrics (library counts, media types, codec distribution)

**Questions for Reviewers**

1. **Label Cardinality**: Currently using `AuthenticationProviderId` as a dynamic label for the auth provider metric. In typical Jellyfin installations, how many distinct auth providers exist? Should this be categorized into "Default" vs "External" instead to prevent unbounded cardinality growth?

2. **Architecture**: Should the remaining metrics (sessions, streams, libraries) be in separate classes following the same pattern, or would you prefer a different organization?

3. **Additional Metrics**: Are there any other user-related or Jellyfin-specific metrics you'd like to see added? Please suggest in the comments.

**Issues**

#3016


sambartik:
Targeting master is OK, but just a note: this wont be accepted to 10.11 AFAIK as this version is currently in the RC phase and no new features are accepted anymore.

ysf3rdm:
When new features are accepted, should I wait for version 10.11.1 or 10.12?

theguymadmax:
> When new features are accepted, should I wait for version 10.11.1 or 10.12?

There's nothing more you need to do; you're already correctly targeting the Master Branch. There is a feature freeze for 10.11, any new features will be reviewed/merged after 10.11 has been released. I would just change your PR title; you don't need to reference the version number you're targeting.

Jellyfin.Server/CoreAppHost.cs
```diff
@@ -88,6 +89,8 @@ protected override void RegisterServices(IServiceCollection serviceCollection)
             serviceCollection.AddSingleton<IDeviceManager, DeviceManager>();
             serviceCollection.AddSingleton<ITrickplayManager, TrickplayManager>();
 
+            serviceCollection.AddSingleton<UserMetrics>();
```
JPVenson: this should be interface based

Jellyfin.Server.Implementations/Users/UserManager.cs
```diff
@@ -73,7 +76,8 @@ public UserManager(
             ILogger<UserManager> logger,
             IServerConfigurationManager serverConfigurationManager,
             IEnumerable<IPasswordResetProvider> passwordResetProviders,
-            IEnumerable<IAuthenticationProvider> authenticationProviders)
+            IEnumerable<IAuthenticationProvider> authenticationProviders,
+            UserMetrics? userMetrics = null)
```
JPVenson: i am not too happy with this dependecy. I think it would be far better if we invert the responsibilities here or at the very least make it far more generic so that other code could also benefit from it

JPVenson (changes requested):
Aren't you missing the actual pipeline registration to enable the prometheus endpoint?
https://github.com/prometheus-net/prometheus-net#aspnet-core-exporter-middleware

Jellyfin.Server.Implementations/Users/UserManager.cs
ysf3rdm: I can create a generic class with an updateMetrics function, which would make it more suitable for other metric types such as session, stream, and library. Do you think this approach is appropriate? Also it resolve your other comment.

ysf3rdm:
> Aren't you missing the actual pipeline registration to enable the prometheus endpoint? https://github.com/prometheus-net/prometheus-net#aspnet-core-exporter-middleware

Jellyfin already exposes Prometheus metrics and includes the necessary middleware/pipeline registration, so no additional registration is required.

Jellyfin.Server.Implementations/Metrics/UserMetrics.cs
```diff
@@ -0,0 +1,249 @@
...
+using Jellyfin.Data;
+using Jellyfin.Database.Implementations.Entities;
+using Jellyfin.Database.Implementations.Enums;
+using MediaBrowser.Controller.Authentication;
+using MediaBrowser.Controller.Library;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using Prometheus;
+
+namespace Jellyfin.Server.Implementations.Metrics
```
Bond-009: use file scoped namespaces for new files

oddstr13: Would that be `Jellyfin.Server.Implementations.Metrics.UserMetrics` in this case?

How is the general structure of the metrics part of this PR, is this how you'd want it structured?

I've been wanting to try my hand at adding some metrics too, but figuring out how and where to place it has been daunting enough that I've just kicked that can down the road. If this structure looks fine other than a few nits, then it looks like a easy starting point for adding more metrics.

ysf3rdm: Thanks for the feedback. I’ve updated the structure based on the comments I received and I think the current version is much better, but I haven’t had the time to push the commits yet. 

To remove the dependencies of the metric classes from other classes — for example, removing UserMetrics’ dependency inside UserManager — using events to update the metrics seemed like the most reasonable approach, and that’s the direction I’m planning to take. 

I also think categorizing the metrics into separate classes makes the overall structure clearer and makes it easier to know exactly where a new metric should go. Once the main structure is finalized, adding new metrics should be as easy as you mentioned.

Jellyfin.Server/CoreAppHost.cs
ysf3rdm: I am not sure it is a best way to work but I don't know where to call UpdateMetrics. I changed it to a HostedService and connected related events to metrics services.

Jellyfin.Server.Implementations/Metrics/SessionMetrics.cs
```diff
@@ -0,0 +1,184 @@
...
+    public class SessionMetrics : IHostedService
+    {
+        private readonly ILogger<SessionMetrics> _logger;
+        private readonly ISessionManager _sessionManager;
+
+        private static readonly Gauge _activeSessions = Prometheus.Metrics
+            .CreateGauge("jellyfin_sessions_active_total", "Number of currently active sessions");
+
+        private static readonly Gauge _totalSessions = Prometheus.Metrics
+            .CreateGauge("jellyfin_sessions_total", "Total number of sessions, active and inactive");
```
jpds, line 26: I don't think there's a need for two different metric names when you can consolidate this into `jellyfin_sessions_total` with a `state={active,inactive}` label.

Jellyfin.Server.Implementations/Metrics/SessionMetrics.cs
```diff
@@ -0,0 +1,184 @@
...
+            .CreateGauge("jellyfin_sessions_by_client", "Number of sessions by client type", new[] { "client" });
+
+        private static readonly Gauge _sessionsByDevice = Prometheus.Metrics
+            .CreateGauge("jellyfin_sessions_by_device", "Number of sessions by device model", new[] { "device" });
+
+        private static readonly Gauge _streamsByMethod = Prometheus.Metrics
+            .CreateGauge("jellyfin_streams_by_method", "Active streams grouped by playback method", new[] { "method" });
+
+        private static readonly Gauge _transcodesByMode = Prometheus.Metrics
+            .CreateGauge("jellyfin_transcodes_by_mode", "Transcodes grouped by software/hardware mode", new[] { "mode" });
```
jpds, line 38: `by_mode` isn't needed here.

Jellyfin.Server.Implementations/Metrics/UserMetrics.cs
```diff
@@ -0,0 +1,281 @@
...
+            .CreateGauge("jellyfin_users_by_permission", "Number of users with specific permissions", new[] { "permission" });
+
+        private static readonly Gauge _usersWithExternalAuth = Prometheus.Metrics
+            .CreateGauge("jellyfin_users_external_auth_total", "Total number of users using external authentication");
+
+        private static readonly Gauge _recentlyActiveUsers = Prometheus.Metrics
+            .CreateGauge("jellyfin_users_recently_active", "Number of users active within the last 30 days");
+
+        private static readonly Gauge _usersWithFailedLogins = Prometheus.Metrics
+            .CreateGauge("jellyfin_users_failed_logins", "Number of users with failed login attempts", new[] { "attempt_range" });
```
jpds, line 52: You can drop the range label here. As an admin - I'm just going to have a graph of `rate(jellyfin_users_failed_logins[$__rate_interval])` and investigate the logs on any abnormalities.

Jellyfin.Server.Implementations/Metrics/UserMetrics.cs
```diff
@@ -0,0 +1,281 @@
...
+        private readonly ILogger<UserMetrics> _logger;
+        private readonly IEnumerable<IAuthenticationProvider> _authenticationProviders;
+        private readonly IUserManager _userManager;
+
+        // Prometheus Gauges for user metrics
+        private static readonly Gauge _totalUserAccounts = Prometheus.Metrics
+            .CreateGauge("jellyfin_user_accounts_total", "Total number of user accounts in Jellyfin");
+
+        private static readonly Gauge _activeUserAccounts = Prometheus.Metrics
+            .CreateGauge("jellyfin_user_accounts_active", "Number of active (non-disabled) user accounts");
```
jpds, line 34: Similar to above - these could be one.

jpds (commented):
Awesome PR. Not a Jellyfin developer - but long-time Prometheus user.

Jellyfin.Server.Implementations/Metrics/UserMetrics.cs
```diff
@@ -0,0 +1,281 @@
...
+            .CreateGauge("jellyfin_users_by_auth_provider", "Number of users by authentication provider", new[] { "provider" });
+
+        private static readonly Gauge _usersByPermission = Prometheus.Metrics
+            .CreateGauge("jellyfin_users_by_permission", "Number of users with specific permissions", new[] { "permission" });
+
+        private static readonly Gauge _usersWithExternalAuth = Prometheus.Metrics
+            .CreateGauge("jellyfin_users_external_auth_total", "Total number of users using external authentication");
+
+        private static readonly Gauge _recentlyActiveUsers = Prometheus.Metrics
+            .CreateGauge("jellyfin_users_recently_active", "Number of users active within the last 30 days");
```
jpds, line 49: This is also not needed as one can just do: `sum(jellyfin_sessions_total{state="active"})`

Issue #14906, opened 2025-09-29
Deleting media that is still being watched in SyncPlay results in errors


kmarius:
### This issue respects the following points:

- [x] This is a **bug**, not a question or a configuration issue; Please visit our [forum or chat rooms](https://jellyfin.org/contact/) first to troubleshoot with volunteers, before creating a report.
- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [x] I'm using an up to date version of Jellyfin Server stable, unstable or master; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Description of the bug

After deleting media that is still being watched in a SyncPlay group it is not possible to for any user that is not in the group to view the list of SyncPlay groups (by clicking the SyncPlay button on the top right). 
An error is shown on the UI: `An error occurred while accessing groups list.` and it is not possible to view or create groups, because the menu doesn't open. Server logs show a `NullReferenceException`.
Everything starts working again once everyone leaves and the group is dissolved.

### Reproduction steps

1. Prepare or backup a Movie from your library; it will be deleted from disk.
2. Log in User1 via browser, create a SyncPlay group
3. Log in User2 via browser, join User1's SyncPlay group (different browser/container tab)
4. Begin watching the movie via SyncPlay
5. Return from playback with User1 (via the Back button)
6. Delete the movie being watched via the UI (three dots menu > Delete media)
7. Leave the SyncPlay group on User1 (playback on User2 continues)
8. With User1, attempt to view SyncPlay groups by clicking the SyncPlay button

### What is the current _bug_ behavior?

The following error is shown and no SyncPlay groups are displayed (the menu doesn't even open):
`An error occurred while accessing groups list.`

### What is the expected _correct_ behavior?

The list of active SyncPlay groups should be displayed.

### Jellyfin Server version

10.10.0+

### Specify the build version

10.10.7

### Environment

```markdown
- OS: Debian 12
- Linux Kernel: 6.1.0-26-amd64
- Virtualization: Docker
- Browser: Firefox, Brave
- Playback Method: Browser
```

### Jellyfin logs

```shell
[...]

PR #14935, opened 2025-10-02
Fixed "Deleting media that is still being watched in SyncPlay results in errors"

pokreman06:
Fixed a problem where if there was a null item in the queue it threw an uncaught exception. 

**Changes**
We added a null check in the has access to queue and returned false if the item was null.
We also added a couple tests using moq to test this functionality.
**Issues**
solving this issue
https://github.com/jellyfin/jellyfin/issues/14906
@ChristopherTYoung @pokreman06


Emby.Server.Implementations/SyncPlay/Group.cs
```diff
@@ -206,7 +206,8 @@ private bool HasAccessToQueue(User user, IReadOnlyList<Guid> queue)
             foreach (var itemId in queue)
             {
                 var item = _libraryManager.GetItemById(itemId);
-                if (!item.IsVisibleStandalone(user))
+
+                if (!item?.IsVisibleStandalone(user) ?? true)
```
Bond-009:
```suggestion
                if (item is null || !item.IsVisibleStandalone(user))
```
Is more readable IMO

tests/Jellyfin.Server.Implementations.Tests/SyncPlay/GroupTests.cs
```diff
@@ -0,0 +1,85 @@
...
+using Jellyfin.Database.Implementations.Entities;
+using MediaBrowser.Controller.Entities;
+using MediaBrowser.Controller.Library;
+using MediaBrowser.Controller.Session;
+using Microsoft.Extensions.Logging;
+using Moq;
+using Xunit;
+
+namespace Jellyfin.Server.Implementations.Tests.SyncPlay
+{
```
Bond-009: Use file scoped namespaces for new files.

tests/Jellyfin.Server.Implementations.Tests/SyncPlay/GroupTests.cs
```diff
@@ -0,0 +1,85 @@
...
+    public class GroupTests
+    {
+        [Fact]
+        public void HasAccessToPlayQueue_ReturnsTrue_WhenItemsAreVisible()
+        {
+            // Arrange
```
Bond-009:
```suggestion
```

tests/Jellyfin.Server.Implementations.Tests/SyncPlay/GroupTests.cs
```diff
@@ -0,0 +1,85 @@
...
+
+            var result = group.HasAccessToPlayQueue(user);
+
+            Assert.True(result);
+        }
+
+        [Fact]
+        public void HasAccessToPlayQueue_ReturnsFalse_WhenLibraryReturnsNullForItem()
+        {
+            // Arrange
```
Bond-009:
```suggestion
```

tests/Jellyfin.Server.Implementations.Tests/SyncPlay/GroupTests.cs
pokreman06: Sorry I might be confused as to how the namespace declarations should be, I reviewed the other tests in similar locations i.e. Jellyfin.Server.Implementations.Tests.Localization
and it was a similar style to how we had the namespace

pokreman06:
Hey just asking for a follow up on this

tests/Jellyfin.Server.Implementations.Tests/SyncPlay/GroupTests.cs
Bond-009: Old files created before file scoped namespaces haven't been changed yet, but all new files should use it.

ChristopherTYoung: Sorry for the confusion. The GroupTests file is now using a file scoped namespace. Thanks for taking the time to review this!

Bond-009 (approved):
Change itself looks fine
Not sure about the tests, lots of duplicated code

ChristopherTYoung:
> Change itself looks fine Not sure about the tests, lots of duplicated code

Thank you! I will spend some time refactoring today

soultaco83:
appears to be working
If file is deleted while active playback
The playback continues for user 2 until point of transcode is reached
user 1 no longer gets error opening groups. sync group just shows closed and am presented an option to create a new one.

JPVenson (approved):

Shadowghost (approved):

Issue #14922, opened 2025-10-01
TMDB Backdrop Scraping is Broken


mikewesten:
### Description of the bug

TMDB must have updated something on their end.
Backdrop images marked "No Language (xx-XX)" are now being scraped as Thumbs by Jellyfin instead of as Backdrops.
This may be something for the plugin developer to look into versus the Jellyfin Server developers, I'm not really sure.
Unless I'm mistaken this bug should be easily reproducible and is fairly self-explanatory without the need of a lengthy, detailed report.

### Reproduction steps

See above

### What is the current _bug_ behavior?

See above

### What is the expected _correct_ behavior?

Scrape TMDB backdrop images marked "No Language (xx-XX)" as Backdrops instead of as Thumbs.

### Jellyfin Server version

10.10.0+

### Specify the build version

10.10.7

### Environment

```markdown
- OS:
- Linux Kernel:
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Jellyfin Data Storage:
- Media Storage:
- External Integrations:
```

### Jellyfin logs

```shell
...
```

### FFmpeg logs

```shell

```

mikewesten:
Nevermind. Server restart seemed to fix it.

mikewesten:
Okay this is likely related to an API change at TMDB that just went live today
See here: https://www.themoviedb.org/talk/68c797b755b9cafed2a0f53f

Imgkl:
I am able to reproduce this on my end too.  Any tentative timeline on when the fix will land for this?

kandykarter:
Seems like all the backdrop images are loading into the Thumbnail image area, visible if you select "All Languages". Decent workaround until a fix lands.

Imgkl:
Oh. That's good to know. Thanks @kandykarter 

But the problem is many 3rd party apps and even swiftfin picks up only the backdrop and if it's not available, it falls back to primary image as backdrop or in some cases empty image icon. Hope this gets fixed soon.

kandykarter:
Oh yeah I meant you can still search from the interface, you just have to manually save the images and set them as backdrops, still easier than a google search

Shadowghost:
This would likely need fixing in TMDbLib directly.

Kockah:
I also encountered the same issue.

[...]

PR #14941, opened 2025-10-04
Handle xx as TMDb no language for backdrops

crobibero:
**Changes**

TMDb now returns `xx` for images that don't have a language, so this change only sets the image type to Thumb if the language is something other than that.

**Issues**

Fixes https://github.com/jellyfin/jellyfin/issues/14922


theguymadmax:
Background images are now downloaded automatically when scanning into the library. However, the selection appears empty when you manually search for images. You need to select "All Languages" for results to populate.

crobibero:
Thanks for testing, now we will just set "xx" to an empty string to make the pipeline the same as before

theguymadmax:
Images now show correctly for the default languages in the metadata image fetcher. However, choosing "All languages" for backdrops works the same as leaving it unselected; you'll only see results in your default language.

That said, it should be good enough for most users.

crobibero:
It's hard to know now but I think that was the previous behavior

JPVenson (approved):

Imgkl:
Thanks @crobibero.

PR #14950, opened 2025-10-05
Deprecate HasPassword property on UserDto

nielsvanvelzen:
**Changes**

Deprecate the HasPassword and HasConfiguredPassword properties on UserDto. They're now nullable and a default value is always set to true. In addition, IAuthenticationProvider no longer provides a `HasPassword` method.

This fixes a long-standing security issue where users of a Jellyfin server could see if a different user has an empty password and public users were exposed to be passwordless to the entire internet.

Additionally fixes the deprecation of HasConfiguredEasyPassword (make nullable and provide default) so it can be removed in a future API version.

Impacted projects:

- HasConfiguredPassword is used by jellyfin-web
- HasPassword is used by jellyfin-web, jellyfin-kodi, jellyfin-vue and jellycon
- LDAP plugin needs changes to remove the `HasPassword` method from IAuthenticationProvider.

**Issues**

### DIFF
--- a/tests/Jellyfin.Server.Integration.Tests/Controllers/UserControllerTests.cs
+++ b/tests/Jellyfin.Server.Integration.Tests/Controllers/UserControllerTests.cs
@@ -147,5 +147,22 @@ namespace Jellyfin.Server.Integration.Tests.Controllers
             using var response = await UpdateUserPassword(client, _testUserId, createRequest);
             Assert.Equal(HttpStatusCode.NoContent, response.StatusCode);
         }
+
+        [Fact]
+        [Priority(2)]
+        public async Task UpdateUserPassword_Empty_RemoveSetPassword()
+        {
+            var client = _factory.CreateClient();
+
+            client.DefaultRequestHeaders.AddAuthHeader(_accessToken!);
+
+            var createRequest = new UpdateUserPassword()
+            {
+                CurrentPw = "4randomPa$$word",
+            };
+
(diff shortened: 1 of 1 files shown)


Shadowghost (approved):

tests/Jellyfin.Server.Integration.Tests/Controllers/UserControllerTests.cs
```diff
@@ -149,35 +146,6 @@ public async Task UpdateUserPassword_Valid_Success()
...
-            var users = await JsonSerializer.DeserializeAsync<UserDto[]>(
-                await client.GetStreamAsync("Users"), _jsonOptions);
-            var user = users!.First(x => x.Id.Equals(_testUserId));
-            Assert.True(user.HasPassword);
-            Assert.True(user.HasConfiguredPassword);
-        }
-
-        [Fact]
-        [Priority(2)]
-        public async Task UpdateUserPassword_Empty_RemoveSetPassword()
```
cvium, line 162: We still support setting an empty password, right? If so, this test should remain, but the assertions should be changed

nielsvanvelzen: The API no longer exposes whether a user has configured a password or not, so there is no way to change the assertions to check for empty password. I added the test back now and only added the response status check.

cvium: That's fine

Bond-009 (approved):

PR #14956, opened 2025-10-07
Add support for inverting episode order in Series

need4swede:
Added InvertEpisodeOrder, allowing episode lists to be displayed in descending order

**Changes**
Updates DtoService and ItemUpdateController to handle this property, and adds corresponding support in BaseItemDto. Includes unit tests to verify correct sorting behavior based on the new property.

**Issues**
Some shows have hundreds of episodes per season. Scrolling to the latest episode sucks. This fixes that experience.

### DIFF
--- a/Emby.Server.Implementations/Dto/DtoService.cs
+++ b/Emby.Server.Implementations/Dto/DtoService.cs
@@ -1256,13 +1256,11 @@ namespace Emby.Server.Implementations.Dto
             }
 
             // Add SeriesInfo
-            Series? series;
-            if (item is Series tmp)
+            if (item is Series seriesItem)
             {
-                series = tmp;
-                dto.AirDays = series.AirDays;
-                dto.AirTime = series.AirTime;
-                dto.Status = series.Status?.ToString();
+                dto.AirDays = seriesItem.AirDays;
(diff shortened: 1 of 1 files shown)


JPVenson:
as an FYI, as this is not a critical bugfix it will not land in 10.11


ref: https://github.com/jellyfin/jellyfin-web/pull/7173

need4swede:
Understandable - thanks!

thornbill:
Tagging as "discussion needed" since **every** client would require updating to support this.

HammyHavoc:
Just wanted to throw this out there:

I've worked doing audio post on a few ongoing documentary programmes for years on end, and this would be hugely useful to have the most recent episodes as being the priority in how they're displayed.

What I would suggest is a toggle for automatically enabling this particular sort-order on libraries that use year-based naming on their `season` folders, e.g., `2005`, `2006`, `2007`, `2008`, `2022, 2023` et cetera, I'm sure the idea will be understood.

However, where I can see this perhaps annoying some users is with old-skool cartoons that are frequently sorted by year rather than the typical ascending season numbers beginning at `1`. Perhaps genre tags could be utilised for rational/reasonable automatic sort order?

PR #14957, opened 2025-10-07
Remove global subtitle configuration

KGT1:
because it doesnt seem to be used.

OpenSubtitles settings are in the plugin anyway,
all other settings are allready settable via library options(LibraryOptions.cs).

This was previously accessible via `POST/GET /System/Configuration/subtitles` but the webclient doesnt use it, and i found no other usage(if you find something do let me know and post the subtitles.xml file. then i can write a migration)

**Changes**
- removes the SubtitleConfigurationFactory and its usages.
- removes SubtitleOptions Model (also from OpenAPI)

**Issues**


Shadowghost (approved):

Shadowghost:
Since it is an API change, this won't be merged for 10.11

KGT1:
sorry, didnt mean to close:)

MBR-0001 (approved):
I like this 👍
 
I wanted to do this because this legacy crap made me waste an hour first time I encountered it

nielsvanvelzen (approved):

Bond-009 (approved):

PR #14963, opened 2025-10-08
Improve performance on people query

theguymadmax:
**Changes**

Replaced nested subqueries with a direct join between `BaseItems` and` Peoples` on `Name`, then filtered `PeopleBaseItemMap` using the resulting `People.Id`.

Individual cast and crew queries now take less than half a second compared to 2-5 seconds. 

**Issues**

### DIFF
--- a/Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
+++ b/Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
@@ -1875,10 +1875,17 @@ public sealed class BaseItemRepository
 
         if (filter.PersonIds.Length > 0)
         {
+            var peopleEntityIds = context.BaseItems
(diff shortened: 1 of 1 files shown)


Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -1875,10 +1875,17 @@ private IQueryable<BaseItemEntity> TranslateQuery(
 
         if (filter.PersonIds.Length > 0)
         {
+            var peopleEntityIds = context.BaseItems
+                .Where(b => filter.PersonIds.Contains(b.Id))
```
crobibero:
```suggestion
                .WhereOneOrMany(filter.PersonIds, b => b.Id)
```

We should probably start using this in more places

crobibero (approved):

PR #6934, opened 2021-12-01
HWA pipeline refactor, AMD/Intel/Nvidia full hardware filtering support, AV1 hwdec

nyanmisaka:
**FFmpeg Changes**
- jellyfin-ffmpeg version 4.4.1
- https://github.com/jellyfin/jellyfin-ffmpeg/pull/76
- Many patches regarding HWA have been applied into `jellyfin-ffmpeg` version 4.4.1.
**IMPORTANT: Using old or original FFmpeg may disable the hardware filtering improvements down below.**

**Changes**
- HWA pipeline refactor, separate the HW pipeline according to HWA method for maintainability.
- AMD/Intel/Nvidia full hardware filtering support. (such as QSV tone-mapping is extended to Windows)
- Experimental AV1 hardware decoding. (I do not have latest gen AMD and Nvidia graphic card for the time being)
- Intel Low-Power encoding. (Reduce overhead in 4k transcoding and tone-mapping, pre-Gen11 only support LP H264)
- Improved tone-mapping and subtitle burn-in performance.
---
- Fix the issue that QSV may fail on Windows if no display is connected.
- Fix green/corrupted output when transcoding HDR content on QSV.
- Fix pixelated output when encoding 4k content on AMD VAAPI.

**Web Changes**
- https://github.com/jellyfin/jellyfin-web/pull/3206

**Issues**
Fixes #6835 #6809 #6725 #6524 #5535

**Win64 build for testing**
https://drive.google.com/drive/folders/1qO2l5Z12nMC7-GHLNEw299N16YdJVdON?usp=sharing

**Docker image for testing**
`docker pull nyanmisaka/jellyfin:latest`

PR #14977, opened 2025-10-10
Adding CudaDevice encoding option to allow users to select Cuda deviceIndex

thomasbrooks4:
My Jellyfin instance on my Unraid server is unable to properly use the second GPU on my machine. If I pass `NVIDIA_VISIBLE_DEVICES=<GPU_ID>` the container picks up my GPU but when running ffmpeg I get a CUDA_ERROR_NO_DEVICE error since it shows as GPU 0 in the container but is GPU 1 on my machine. If I expose `NVIDIA_VISIBLE_DEVICES=all` then it will always select my first GPU. Adding the encoding option to set which index to use will solve my issue. I have tried to separate the GPU's with IOMMU but without any luck. With testing I found that setting `NVIDIA_VISIBLE_DEVICES=all` and using `-init_hw_device cuda=cu:1` manually it uses the correct GPU without issues.

**Changes**

- Add `CudaDevice` to `EncodingOptions.cs`
- Pass `options.CudaDevice` into `GetCudaDeviceArgs` instead of hardcoded `0`
- Default `options.CudaDevice` to `0` to remain existing functionality

**Links**
jellyfin-web PR: https://github.com/jellyfin/jellyfin-web/pull/7184

I followed the pattern from a previously added encoding option: https://github.com/jellyfin/jellyfin/pull/12384
Reference to initial hardcode in 2021 and unchanged since then: https://github.com/jellyfin/jellyfin/pull/6934#discussion_r771781193

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -1126,7 +1126,7 @@ namespace MediaBrowser.Controller.MediaEncoding
                     return string.Empty;
                 }
 
-                args.Append(GetCudaDeviceArgs(0, CudaAlias))
+                args.Append(GetCudaDeviceArgs(options.CudaDevice, CudaAlias))
                      .Append(GetFilterHwDeviceArgs(CudaAlias));
             }
             else if (optHwaccelType == HardwareAccelerationType.amf)
--- a/MediaBrowser.Model/Configuration/EncodingOptions.cs
+++ b/MediaBrowser.Model/Configuration/EncodingOptions.cs
@@ -29,6 +29,7 @@ public class EncodingOptions
         // plus it's the default one in ffmpeg if you don't specify anything
         VaapiDevice = "/dev/dri/renderD128";
         QsvDevice = string.Empty;
+        CudaDevice = 0;
         EnableTonemapping = false;
         EnableVppTonemapping = false;
         EnableVideoToolboxTonemapping = false;
@@ -145,6 +146,11 @@ public class EncodingOptions
     /// Gets or sets the QSV device.
     /// </summary>
     public string QsvDevice { get; set; }
+
+    /// <summary>
+    /// Gets or sets the Cuda device.
+    /// </summary>
+    public int CudaDevice { get; set; }
 
     /// <summary>
     /// Gets or sets a value indicating whether tonemapping is enabled.


nyanmisaka:
Does using `CUDA_VISIBLE_DEVICES` instead of `NVIDIA_VISIBLE_DEVICES` help?

https://developer.nvidia.com/blog/cuda-pro-tip-control-gpu-visibility-cuda_visible_devices/

thomasbrooks4:
> Does using `CUDA_VISIBLE_DEVICES` instead of `NVIDIA_VISIBLE_DEVICES` help?
> 
> https://developer.nvidia.com/blog/cuda-pro-tip-control-gpu-visibility-cuda_visible_devices/

I can try that, but as I mentioned using `NVIDIA_VISIBLE_DEVICES` and GPU ID still works and the device is visible to the container. I can use `nvidia--smi` and see the correct GPU. Although I think since they are stuck in the same IOMMU group (I think due to pcie bridging) indexing it as `cu:0` doesn't work even though it is index 0 in the container. So it can't hurt to try, but I'm not sure how it might be different since it was already visible to the container anyway.

EDIT: Read further on the article and can see how this might differ, I do appreciate the suggestion and will let anyone know if this helps. Although I will say regardless it would be nice to have some control on this device index as well.

thomasbrooks4:
> Does using `CUDA_VISIBLE_DEVICES` instead of `NVIDIA_VISIBLE_DEVICES` help?
> 
> https://developer.nvidia.com/blog/cuda-pro-tip-control-gpu-visibility-cuda_visible_devices/

Man where was this comment when I was first searching. Searching up more about `CUDA_VISIBLE_DEVICES` led me to more people with my issue as you knew. These were buried I guess in all my other searches. Well it took several iterations but it did end up working. Most people said to use `CUDA_VISIBLE_DEVICES` instead of `NVIDIA_VISIBLE_DEVICES` but that never worked.

Ultimately I needed to keep my existing settings: `--runtime=nvidia` and `NVIDIA_DRIVER_CAPBILITIES`. Then I had to change `NVIDIA_VISIBLE_DEVICES` from GPU ID to "all". From there I had to add ANOTHER variable `CUDA_VISIBLE_DEVICES` to 1 so that we can address the correct one. All other configurations never allowed the container to still use `nvidia-smi`. So thank you!

I do think this would be good functionality though, hopefully this can go in.

nyanmisaka:
> > Does using `CUDA_VISIBLE_DEVICES` instead of `NVIDIA_VISIBLE_DEVICES` help?
> > [developer.nvidia.com/blog/cuda-pro-tip-control-gpu-visibility-cuda_visible_devices](https://developer.nvidia.com/blog/cuda-pro-tip-control-gpu-visibility-cuda_visible_devices/)
> 
> Man where was this comment when I was first searching. Searching up more about `CUDA_VISIBLE_DEVICES` led me to more people with my issue as you knew. These were buried I guess in all my other searches. Well it took several iterations but it did end up working. Most people said to use `CUDA_VISIBLE_DEVICES` instead of `NVIDIA_VISIBLE_DEVICES` but that never worked.
> 
> Ultimately I needed to keep my existing settings: `--runtime=nvidia` and `NVIDIA_DRIVER_CAPBILITIES`. Then I had to change `NVIDIA_VISIBLE_DEVICES` from GPU ID to "all". From there I had to add ANOTHER variable `CUDA_VISIBLE_DEVICES` to 1 so that we can address the correct one. All other configurations never allowed the container to still use `nvidia-smi`. So thank you!
> 
> I do think this would be good functionality though, hopefully this can go in.

These options were previously added for VAAPI and QSV because they lacked corresponding environment variables.

The next step is for the server/ffmpeg to detect available GPU options and capabilities and present them to the user in a list. Therefore, these options will be deprecated or removed.

Bond-009:
Closing as not planned

PR #14994, opened 2025-10-13
Remove database Endpoint to restore pre 10.11 functionality

KGT1:
this restores pre 10.11 handling of this endpoint.
as per chat with @JPVenson it seems like this was unintentionally added anyway in [aa811eb](https://github.com/jellyfin/jellyfin/commit/aa811eb1e3c78bdf8f4a751311c1bb6d639e851e).

Also im not so sure if its a good idea to let every authenticated user see the database connection details. with a provider like JPVenson/Jellyfin.Pgsql, this would leak the connection string.

**Changes**
- Removes the `POST/GET /System/Configuration/database` endpoint

this is kind of a hack, in the medium term i think the generic endpoint should be removed anyway.

### DIFF
--- a/Jellyfin.Api/Controllers/ConfigurationController.cs
+++ b/Jellyfin.Api/Controllers/ConfigurationController.cs
@@ -80,6 +80,11 @@ public class ConfigurationController : BaseJellyfinApiController
     [ProducesFile(MediaTypeNames.Application.Json)]
     public ActionResult<object> GetNamedConfiguration([FromRoute, Required] string key)
     {
+        if (string.Equals(key, "database", StringComparison.OrdinalIgnoreCase))
+        {
+            return NotFound();
+        }
+
         return _configurationManager.GetConfiguration(key);
     }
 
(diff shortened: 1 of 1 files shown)


KGT1:
Obviously understandable, i just dont think its very wise nor intentional to have a new endpoint in 10.11 that leaks database connection info to all authenticated users, otherwise i would have waited with this.

nielsvanvelzen:
In 10.11 this configuration isn't really used to store sensitive data. We don't use any authentication for the local sqlite file. This abstract configuration API leaks a lot more data to users, therefor the plan is for 10.12 to replace it with individual endpoints per configuration so we have better control over which options are exposed and to who.

KGT1:
Yes i am aware. Since its technically possible for plugins to register anything with auth and then unintentionally leaking stuff, it thought this might be of concern, but if its not because its not supported for 10.11, feel free to close this. Its a hack for this Problem anyway.

KGT1:
since this doesnt seem to be of concern, or at least not to warrent any api change, ill close this in favour of a the bigger configuration refactor for 10.12.

PR #15048, opened 2025-10-20
Fix stale UserData cache

### LINKED ISSUE #14981: [Bug] 10.11: Favorites get lost during playback
### Description of the bug

I set up jellyfin 10.11.0-rc8 with docker compose behind nginx. I used the recommended settings with no modifications, except taking out the ssl config of the nginx file and having it generated by certbot instead.

When playing music and favoriting, sometimes new changes get reset. This seems to be tied to the progress update the client sends to the server.

### Reproduction steps

1. Sign in to Jellyfin with Firefox or the Android app.
2. Start playing music.
3. Favorite the song currently playing.
4. Wait a bit (by watching the network trace, the moment can be more-or-less pinpointed to the POST request to the /progress endpoint.)
5. When reloading, or loading the favorite list, the song is no longer favorited.
6. Alternatively, favorite and reload *before* the progress POST; the song is now correctly favorited.
7. This works analogous for unfavoriting.

The favoriting status of songs does not 'stick', without further investigation the action seems to fail at random.

When a song is favorited or unfavorited, this persists regardless of other actions.

### Jellyfin Server version

10.10.0+

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.0-RC8

### Environment

```markdown
- OS: Debian 12
- Linux Kernel: 6.1.0-38-amd64 x86_64
- Virtualization: Docker [compose]
- Clients: Browser, Android
- Browser: Firefox 136.0.3
- FFmpeg Version: default
- Playback Method: default
- Hardware Acceleration: default
- GPU Model: none
- Plugins: none
- Reverse Proxy: nginx
- Base URL: none
- Networking: unknown
- Jellyfin Data Storage: Docker Volume
- Media Storage: bind mount sftp into docker
- External Integrations: none
```

### Jellyfin logs

```shell
It's 4 at night and I just got 10.10.7 running successfully I'll get you the logs tomorrow
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

### LINKED ISSUE #15615: Items appear in "Unplayed" folder listing despite being recently played
### Description of the bug

I'm using `10.11.3+ubu2404`. I have a folder with a large number of "Home Videos and Photos" (335 currently). I have a web client and Android TV client that both have the same folder set with the "Unplayed" filter, but after watching a video, that video will still sometimes appear in both clients.

I've confirmed in the SQLITE backend the content in question is correctly marked as watched/played:

```
sqlite3 jellyfin.db "SELECT UserData.Played, UserData.PlayCount, UserData.PlaybackPositionTicks, UserData.LastPlayedDate FROM UserData, BaseItems WHERE UserData.ItemId = BaseItems.Id AND UserData.UserId = '3C78E926-89D6-4BFC-9133-657ADCC33D13' AND BaseItems.Id = 'E0D424CD-3257-5297-8F14-999D654FD8AF';"
1|1|0|2025-11-23 15:21:30.2429039
```
If I run "Scan all libraries" from the Dashboard the issues no longer presents itself.

### Reproduction steps

1. Have a large number of items in a "Home Videos and Photos" media library.
2. Set the filter to "Unplayed".
3. Watch a video and confirm it's marked as "Watched".
4. See that it still shows in the media library, but when you navigate to the item detail page the "watched" check mark is checked.
5. Confirm via SQL that it has been watched (as above).

The item is still visible in an "Unplayed" filter, when it shouldn't be.

The item should be hidden from an "Unplayed" filter.

### Jellyfin Server version

10.11.3

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.3

### Environment

```markdown
- OS: Ubuntu 24.04
- Linux Kernel: 6.8.0-86-generic
- Virtualization: N/A
- Clients: Web and Android TV
- Browser: Chrome
- FFmpeg Version: jellyfin-ffmpeg7 7.1.2-4-noble          
- Hardware Acceleration: none
- GPU Model: none
- Plugins: none
- Reverse Proxy: none
- Base URL: none
- Networking: N/A
- Media Storage: external USB drive, mounted to `/srv/jellyfin` via `/etc/fstab` using ext4 filesystem
- External Integrations: None
```

### Jellyfin logs

```shell
N/A
```

### FFmpeg logs

```shell
N/A
```

### Client / Browser logs

N/A

N/A

### Additional information

N/A

theguymadmax:
**Changes**

Check cache before reading `item.UserData` to fix stale data. Also caches empty results and removes nullable return type.

**Issues**
 Fixes #14981
Fixes #15615

### DIFF
--- a/Emby.Server.Implementations/Library/UserDataManager.cs
+++ b/Emby.Server.Implementations/Library/UserDataManager.cs
@@ -236,12 +236,20 @@ namespace Emby.Server.Implementations.Library
         }
 
         /// <inheritdoc />
-        public UserItemData? GetUserData(User user, BaseItem item)
+        public UserItemData GetUserData(User user, BaseItem item)
         {
-            return item.UserData?.Where(e => e.UserId.Equals(user.Id)).Select(Map).FirstOrDefault() ?? new UserItemData()
+            var cacheKey = GetCacheKey(user.InternalId, item.Id);
+            if (_cache.TryGet(cacheKey, out var cachedData))
             {
-                Key = item.GetUserDataKeys()[0],
-            };
+                return cachedData;
+            }
+
+            var data = item.UserData?.Where(e => e.UserId.Equals(user.Id)).Select(Map).FirstOrDefault() ?? new UserItemData()
+                {
+                    Key = item.GetUserDataKeys()[0],
+                };
+
+            return _cache.GetOrAdd(cacheKey, _ => data);
         }
 
         /// <inheritdoc />
--- a/Emby.Server.Implementations/Library/UserDataManager.cs
+++ b/Emby.Server.Implementations/Library/UserDataManager.cs
@@ -239,17 +239,13 @@ namespace Emby.Server.Implementations.Library
         public UserItemData GetUserData(User user, BaseItem item)
         {
             var cacheKey = GetCacheKey(user.InternalId, item.Id);
-            if (_cache.TryGet(cacheKey, out var cachedData))
-            {
-                return cachedData;
-            }
-
-            var data = item.UserData?.Where(e => e.UserId.Equals(user.Id)).Select(Map).FirstOrDefault() ?? new UserItemData()
+            return _cache.GetOrAdd(
+                cacheKey,
+                (k, i) => i.UserData?.Where(e => e.UserId.Equals(user.Id)).Select(Map).FirstOrDefault() ?? new UserItemData()
                 {
-                    Key = item.GetUserDataKeys()[0],
-                };
-
-            return _cache.GetOrAdd(cacheKey, _ => data);
+                    Key = i.GetUserDataKeys()[0],
+                },
+                item);
         }
 
         /// <inheritdoc />


JPVenson:
i would rather see the UserData entries on the BaseItem get invalidated or Rehydrated instead of the global cache.

theguymadmax:
I can't get item.UserData to be invalidated or rehydrated when there are multiple instances of the same item, e.g., when on an album page and then having the items loaded in the player queue. It works fine when just dealing with the album page. Is there a way to get it to invalidate UserData across all instances of the same item? 

It's not an issue with the Cache-first approach.

mthaddon:
> I can't get item.UserData to be invalidated or rehydrated when there are multiple instances of the same item, e.g., when on an album page and then having the items loaded in the player queue. It works fine when just dealing with the album page. Is there a way to get it to invalidate UserData across all instances of the same item?

Lines 80-83 of this same file seems to be the ones that are updating and rehydrating the cache:
```
var userId = user.InternalId;
var cacheKey = GetCacheKey(userId, item.Id);
_cache.AddOrUpdate(cacheKey, userData);
item.UserData = dbContext.UserData.Where(e => e.ItemId == item.Id).AsNoTracking().ToArray(); // rehydrate the cached userdata
```
Is there something about how this is done that means it's not applied as expected, or the cache isn't populated with expected data?

JPVenson:
@theguymadmax sry for slacking off with the response.

The [UserDataManager](https://github.com/jellyfin/jellyfin/pull/15048/changes#diff-6851c1ed0d58b35503b08d426721e8d59fa451341160bda88f4b388c91c83680) is supposed to be on its way out, cached items should always come from the LibraryManager which caches BaseItems and we always now query BaseItems with their UserData attached directly from the database instead of relying on the UserDataManager.

Emby.Server.Implementations/Library/UserDataManager.cs
```diff
@@ -236,12 +236,20 @@ private static string GetCacheKey(long internalUserId, Guid itemId)
...
-            };
+                return cachedData;
+            }
+
+            var data = item.UserData?.Where(e => e.UserId.Equals(user.Id)).Select(Map).FirstOrDefault() ?? new UserItemData()
+                {
+                    Key = item.GetUserDataKeys()[0],
+                };
+
+            return _cache.GetOrAdd(cacheKey, _ => data);
```
cvium: Why wasn't it done like this?
```suggestion
            return _cache.GetOrAdd(cacheKey, (k, i) => i.UserData?.Where(e => e.UserId.Equals(user.Id)).Select(Map).FirstOrDefault() ?? new UserItemData()
                {
                    Key = i.GetUserDataKeys()[0],
                }, item);
```

JPVenson: also why would the BaseItem.UserData not be preferred? Its was more likely to be fresher then the cache

theguymadmax: I tried `BaseItem.UserData` but it doesn't work, Audio items don't appear to be cached by LibraryManager, so when you open up the player queue on the album page there are duplicate copies of the same track in memory. Updating UserData on one copy doesn't update the others, so we need to go through the `UserDataManager` cache. Well, that's my best guess, why it's not working.

JPVenson (approved):
as discussed internally.

As a fix for 10.11.10 this is "_fine_" but it may never be backported and instead needs to be fixed proper on master

Shadowghost (approved):

Issue #12734, opened 2024-09-28
Playlists are randomly empty.


viasux:
### This issue respects the following points:

- [X] This is a **bug**, not a question or a configuration issue; Please visit our forum or chat rooms first to troubleshoot with volunteers, before creating a report. The links can be found [here](https://jellyfin.org/contact/).
- [X] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [X] I'm using an up to date version of Jellyfin Server stable, unstable or master; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.
- [X] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [X] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Description of the bug

Earlier today, my playlists were perfectly intact. Now they are completely empty.

### Reproduction steps

This has happened twice. I've updated since the first time it happened. I run the docker version, it appears to happen when my rclone mount fails, and persists even when I restore the directory.

### What is the current _bug_ behavior?

The playlists remain completely empty, but should restore once the drive is reconnected.

### What is the expected _correct_ behavior?

My playlists persisting.

### Jellyfin Server version

Older*

### Specify version number

10.9.7

### Specify the build version

10.9.7

### Environment

```markdown
- OS:Linux Debian
- Linux Kernel: Stable
- Virtualization: Yes
- Clients: Finamp & Web
- Browser: Thorium
- FFmpeg Version: ?
- Playback Method: ?
- Hardware Acceleration: No
- GPU Model: N/A
- Plugins: AniList, AudioDB, Cover Art Archive, Last.fm, ListenBrainz, MusicBrainz, OMDB, Open Subtitles, Studio Images, TMDb
- Reverse Proxy: Yes
- Base URL: Do not wish to provide
- Networking: ? 
- Storage: B2 Bucket
```

### Jellyfin logs

```shell
[2024-09-28 02:00:00.168 +00:00] [INF] [22] Emby.Server.Implementations.ScheduledTasks.TaskManager: "Extract Chapter Images" Completed after 0 minute(s) and 0 seconds
[...]

PR #15049, opened 2025-10-20
Fix issue 12734 Rm StartupTrigger for CleanupCollectionAndPlaylistpathsTask

h4p:
**Changes**
The "Clean up collection and playlist" task is not triggered on startup anymore because that can lead to data loss quickly if the data storage is not mounted.

**Issues**
issue [12734](https://github.com/jellyfin/jellyfin/issues/12734)


theguymadmax:
This task is being removing in https://github.com/jellyfin/jellyfin/pull/16062, thus closing this PR.

PR #15062, opened 2025-10-21
Add media source version prioritization

mlaustin44:
**Changes**
Adds server configuration options to control the ordering of media sources when multiple video versions are available (e.g., 4K, 1080p, 720p). Administrators can set a preferred resolution (e.g., 1080p) to be selected first, and configure whether remaining versions sort by ascending or descending quality. This allows control of which version is used by default during playback, which is highly useful for libraries that have a 4k and a 720p/1080p version of content - as right now the default behavior would be to transcode the 4k version!

This PR only covers the server, but I intend to work on a matching PR to the web app once this is merged.

**Issues**

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -87,6 +87,7 @@
  - [Maxr1998](https://github.com/Maxr1998)
  - [mcarlton00](https://github.com/mcarlton00)
  - [mitchfizz05](https://github.com/mitchfizz05)
+ - [mlaustin44](https://github.com/mlaustin44)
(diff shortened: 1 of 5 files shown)


lostb1t:
just my 2 cents. Id rather see clients send supported resolutions to the server so the server can make an educated guess about which version to play

sjorge:
> just my 2 cente. Id rather see clients send supported resolutions to the server so the server can make an educated guess about which version to play

And maybe also HDR or not. As that can also help picking the best version for the client

PR #15071, opened 2025-10-21
 Fix backdrop image ordering regression in EF Core migration

### LINKED ISSUE #15105: [10.11.0] Backdrop image order not being saved
### Description of the bug

On a large number of backdrops for a library item (4+ but sometimes also less), the order of bacldrop images are getting shuffled despite rearranging them repeatedly. This does not seem to happen with library items with only 2 or 3 images.

I haven't been able to ascertain whether it's being shuffled or simply returning to a position undefined by user (e.g. filename or order that images were downloadet)

### Reproduction steps

Upgrade to 10.11

Upon arranging backdrop image order by using the left and right arrows. Images will appear to shuffle position when a set amount of time has passed.

When rearranging backdrop images for their order, this is saved upon exit.

### Jellyfin Server version

10.10.0+

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11

### Environment

```markdown
- OS:
- Linux Kernel:
- Virtualization: Docker
- Clients: Browser
- Browser: Brave
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Jellyfin Data Storage:
- Media Storage:
- External Integrations:
```

### Jellyfin logs

```shell
Bug does nor show in log (no exceptions or similar)
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

Rearranging backdrop images:

Image

After 5 to 10 minutes has passed, backdrop images are shuffled. I haven't tested if it's truly random or the same (i.e., according to filename or similar):

Image

### Additional information

_No response_

seeyabye:
## Issue

After migrating from SQLite direct queries to Entity Framework Core, backdrop images are no longer displayed in the correct order. Specifically, `fanart.jpg` (which should have priority) is often displayed after images from the `extrafanart/` folder.

This affects the display when "Select view" is set to "Thumb" or "Thumb Card", the backdrop image shown becomes unpredictable.

  ### Root Cause

  **In v10.10.7:**
  - Images were stored as a serialized string in a single database column
  - Order was preserved during serialization/deserialization

  **In current main (EF Core):**
  - Images are stored in the `BaseItemImageInfos` table without an explicit order column
  - When loading images, no `OrderBy` clause was specified
  - EF Core returns images in undefined order (typically database row order)
  - Simple alphabetical sorting doesn't work because `extrafanart/fanart1.jpg` sorts before `fanart.jpg` alphabetically

  ### Solution

  Implements priority-based sorting that matches the `LocalImageProvider` discovery order:

  1. `{mediaFileName}-fanart.jpg` (highest priority)
  2. `fanart.jpg` (not in extrafanart folder)
  3. `fanart-N.jpg` (numbered backdrops)
  4. `background.jpg` / `background-N.jpg`
  5. `art.jpg` / `art-N.jpg`
  6. `extrafanart/*` folder contents (alphabetical)
  7. `backdrop.jpg` / `backdropN.jpg`
  8. Other images (lowest priority)

  Within each priority group, images are sorted by numeric index (if present) and then alphabetically by path for stability.

  ### Changes

  - **File**: `Jellyfin.Server.Implementations/Item/BaseItemRepository.cs`
  - Added `GetImageOrderPriority()` helper method to determine priority based on filename patterns
  - Added `GetNumericImageIndex()` helper method to extract numeric indices from filenames
  - Updated image mapping to apply priority-based ordering

  ### Testing

  With a file structure like:
```
  ├── fanart.jpg
  ├── extrafanart/
  │   ├── fanart1.jpg
  │   ├── fanart2.jpg
  │   └── ...
  └── video.mp4
```

  **Before**: `extrafanart/fanart1.jpg` often displayed first
  **After**: `fanart.jpg` correctly displays first


Shadowghost:
Bugfixes should target the release branch for 10.11

JPVenson:
i really dont see this as a bugfix though.

AFAIK we never had an explicit order and it was just ordered in that particular way accidentally

Shadowghost:
It might have been implicitly ordered by how we scanned and saved images though, so I'd argue that this is a bugfix if it restores 10.10 behaviour (even if explicit now)

JPVenson:
that order was never explicitly defined or should have been expected. As there was no implicit order before, i argue you should have exepected a random order before.

seeyabye:
Although it has never been explicitly defined, one could argue that the implementation for adding/populating images is somewhat implicitly ordered. For instance, in `LocalImageProvider.cs`, methods like `PopulateImages` have comments saying which is first preferred. This means ordering is implied somewhat.

https://github.com/jellyfin/jellyfin/blob/aefb9b2cffca3ccb7a95d89af0d9fb2040de6ed3/MediaBrowser.LocalMetadata/Images/LocalImageProvider.cs#L210-L219

And then you have dedicated methods like populating `extrafanarts` which was added at the end in `PopulateBackdrops`

https://github.com/jellyfin/jellyfin/blob/aefb9b2cffca3ccb7a95d89af0d9fb2040de6ed3/MediaBrowser.LocalMetadata/Images/LocalImageProvider.cs#L313-L344

Last but not least, there is a swapping functionality that is supported for the backdrop images when we choose "Edit Images". Why have swaps if ordering is random?

https://github.com/jellyfin/jellyfin/blob/aefb9b2cffca3ccb7a95d89af0d9fb2040de6ed3/Emby.Server.Implementations/IO/ManagedFileSystem.cs#L469-L485

I think ordering is implicitly implied, but the only issue is *which* backdrop gets chosen to be shown as a "Thumb" / "Thumb Card" isn't clearly defined.

----

Anyhow if the decision is that this is not a bugfix, then I will make a new feature PR. Would that be acceptable?

This PR has issues as I just realized that I hardcoded it for "jpg" only and it should've supported different extensions as well. Also, I think it might be better to implement this by introducing a new `SortOrder` column in the `BaseItemImageInfos` instead.

theguymadmax:
A user could define the image order through the webUI, which is now impossible ( #15105)

This is a bugfix to me.

JPVenson:
@theguymadmax i cannot find any place in the webUI that allows explict ordering of images, can you specify?

theguymadmax:
You can order the images, with the < > icons,  when you select edit images.

JPVenson:
i just tried that on 10.11 and it works fine?

image

image
---
image
image

seeyabye:
@JPVenson you might want to take a look at this issue https://github.com/jellyfin/jellyfin/issues/15105

The images will eventually go out of order.

The retrieval order now is random because the order of query isn't guaranteed. Previously, the order or retrieval is preserved during serialization/deserialization.

theguymadmax:
In 10.10.7 the order is persevered in the Images column in the TypeBaseItems:

`T:\Movies\Movies Artwork\The Big Short\fanart.jpg*638325570898431373*Backdrop*1920*1080*W^JsaX%0IpxCNIRl~9t6M\slRlWCx@X8V[aeWCoLxtf+aeWBays.|T:\Movies\Movies Artwork\The Big Short\extrafanart\fanart1.jpg*638325570700013131*Backdrop*1920*1080*WfGlI/-;ofoft7xv~qR/Rkj?axtRt7M{ayt7WBf7xut7j[t7WBV@|T:\Movies\Movies Artwork\The Big Short\extrafanart\fanart2.jpg*638146726381976439*Backdrop*1920*1080*WmOq4i019G.7%M?b_MxukCWCayt7%Mxut7afRjRj_3ayfkt7oft6|T:\Movies\Movies Artwork\The Big Short\extrafanart\fanart3.jpg*638325570737919531*Backdrop*1920*1080*WqJ@wDxrV?t7f,oN~ooajEM\R-a$-;oeRji_xuxa-;R+WAnht8oe|T:\Movies\Movies Artwork\The Big Short\extrafanart\fanart4.jpg*638325570627205554*Backdrop*2997*1686*WnD,_;x]DORjWAjE?^t8IBR/oIaytPe-RlX9ofay%3s.WAWCaLe.|T:\Movies\Movies Artwork\The Big Short\extrafanart\fanart5.jpg*638325570879109516*Backdrop*1920*1080*WFBYS=%qRkp-pVx=$UL;M#XhRVZ-MOWTt$Q]VetMpopmtRn8x.kT|%MetadataPath%\library\fa\fad805fd2b8f90b3b9227416fd311054\poster.jpg*638966909849402649*Primary*640*480*VA7nXg?]?cXQRQ?b-;xuWWRjozxbX7ofWUozt7offjbF`

After swapping the order of the fist two items. 
`T:\Movies\Movies Artwork\The Big Short\fanart.jpg*638325570700013131*Backdrop*1920*1080*WfGlI/-;ofoft7xv~qR/Rkj?axtRt7M{ayt7WBf7xut7j[t7WBV@|T:\Movies\Movies Artwork\The Big Short\extrafanart\fanart1.jpg*638325570898431373*Backdrop*1920*1080*W^JsaX%0IpxCNIRl~9t6M\slRlWCx@X8V[aeWCoLxtf+aeWBays.|T:\Movies\Movies Artwork\The Big Short\extrafanart\fanart2.jpg*638146726381976439*Backdrop*1920*1080*WmOq4i019G.7%M?b_MxukCWCayt7%Mxut7afRjRj_3ayfkt7oft6|T:\Movies\Movies Artwork\The Big Short\extrafanart\fanart3.jpg*638325570737919531*Backdrop*1920*1080*WqJ@wDxrV?t7f,oN~ooajEM\R-a$-;oeRji_xuxa-;R+WAnht8oe|T:\Movies\Movies Artwork\The Big Short\extrafanart\fanart4.jpg*638325570627205554*Backdrop*2997*1686*WnD,_;x]DORjWAjE?^t8IBR/oIaytPe-RlX9ofay%3s.WAWCaLe.|T:\Movies\Movies Artwork\The Big Short\extrafanart\fanart5.jpg*638325570879109516*Backdrop*1920*1080*WFBYS=%qRkp-pVx=$UL;M#XhRVZ-MOWTt$Q]VetMpopmtRn8x.kT|%MetadataPath%\library\fa\fad805fd2b8f90b3b9227416fd311054\poster.jpg*638966909849402649*Primary*640*480*VA7nXg?]?cXQRQ?b-;xuWWRjozxbX7ofWUozt7offjbF`

10.11 in the it will change in the UI. But once there's a refresh, it will revert back since the ordering is not defined.

Shadowghost (approved):
IMO this is a sensible solution to fix it

ferferga:
Given the order can be changed in the WebUI, the "ordering by path" solution is not enough. Why we don't add an extra column from the beginning to store the order?

Shadowghost:
@ferferga the ordering is (believe it or not) decided by the name. If you order in the webui, it changes the names 😓

ferferga:
@Shadowghost Funnily enough this doesn't surprise me at all 😅😂. Still, why not change that and do it properly? It requires a migration and scanning the filesystem. All in all, probably it's faster a single roundtrip to the DB than 2 roundtrips (DB + filesystem).

This way it's also predicatble and we avoid further missteps in the future.

Shadowghost:
Don't get me wrong, I fully support handling it in the DB

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -561,9 +566,127 @@ public void SaveImages(BaseItemDto item)
...
+        var sort1 = images[index1].SortOrder;
+        var sort2 = images[index2].SortOrder;
+
+        // Use EF metadata to get the correct table name with schema (provider-aware)
+        var entityType = context.Model.FindEntityType(typeof(BaseItemImageInfo));
+        var tableName = entityType!.GetSchemaQualifiedTableName();
+
+        // Use ExecuteSqlRaw with schema-qualified table name (from trusted EF metadata) and parameterized values
+        // Table name is safe to concatenate as it comes from EF Core model, not user input
+        var sql = $@"
```
JPVenson: Raw SQL will not be accepted

JPVenson (changes requested):

Jellyfin.Server/Migrations/Routines/PopulateImageSortOrder.cs
```diff
@@ -0,0 +1,239 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Jellyfin.Server.Migrations.Routines;
+
+/// <summary>
+/// Populate SortOrder for existing images after the AddSortOrderToBaseItemImageInfo migration.
+/// </summary>
+#pragma warning disable CS0618 // Type or member is obsolete
+[JellyfinMigration("2025-10-23T02:30:00", nameof(PopulateImageSortOrder), "A7B3F8E9-4C2D-4E1A-9B5F-6D8E3F2A1C9B")]
+internal class PopulateImageSortOrder : IMigrationRoutine
```
JPVenson: Manipulates database -> Needs backup attribute 
Use `IAsyncMigrationRoutine` instead

Jellyfin.Server/Migrations/Routines/PopulateImageSortOrder.cs
```diff
@@ -0,0 +1,239 @@
...
+#pragma warning restore CS0618 // Type or member is obsolete
+{
+    private const int UnknownImagePriority = 999;
+
+    private readonly IDbContextFactory<JellyfinDbContext> _dbProvider;
+    private readonly ILogger<PopulateImageSortOrder> _logger;
+
+    public PopulateImageSortOrder(
+        IDbContextFactory<JellyfinDbContext> dbProvider,
+        ILogger<PopulateImageSortOrder> logger)
```
JPVenson: Use `IStartupLogger` instead to provide UI Feedback for StartupUI

Jellyfin.Server/Migrations/Routines/PopulateImageSortOrder.cs
```diff
@@ -0,0 +1,239 @@
...
+
+    /// <inheritdoc/>
+    public void Perform()
+    {
+        _logger.LogInformation("Starting SortOrder population for existing images");
+
+        using var context = _dbProvider.CreateDbContext();
+
+        // Get all distinct ItemIds that have images
+        var itemIdsWithImages = context.BaseItemImageInfos
```
JPVenson: No need to materialise the list here just to have it as a parameter in the next query. Use the IQueryable, EFCore is perfectly capable of translating that

Jellyfin.Server/Migrations/Routines/PopulateImageSortOrder.cs
```diff
@@ -0,0 +1,239 @@
...
+
+        _logger.LogInformation("Processing {Count} items with images", itemIdsWithImages.Count);
+
+        // Batch processing configuration
+        const int batchSize = 500;
+        int processedCount = 0;
+        int errorCount = 0;
+
+        // Load all base item paths into memory for fast lookup
+        var itemPathLookup = context.BaseItems
```
JPVenson: Loading all BaseItems into memory also pushes a lot of ram which isn't nessesary here.

Jellyfin.Server/Migrations/Routines/PopulateImageSortOrder.cs
```diff
@@ -0,0 +1,239 @@
...
+            .ToList();
+
+        _logger.LogInformation("Loaded {Count} images into memory for processing", allImages.Count);
+
+        // Group images by ItemId
+        var imagesByItem = allImages.GroupBy(i => i.ItemId).ToList();
+
+        for (int batchStart = 0; batchStart < imagesByItem.Count; batchStart += batchSize)
+        {
+            var batch = imagesByItem.Skip(batchStart).Take(batchSize).ToList();
```
JPVenson: Use the PartitionHelper instead and join in the navigation properties instead

Jellyfin.Server/Migrations/Routines/PopulateImageSortOrder.cs
```diff
@@ -0,0 +1,239 @@
...
+        }
+
+        _logger.LogInformation(
+            "SortOrder population completed. Processed: {Processed}, Errors: {Errors}, Total: {Total}",
+            processedCount,
+            errorCount,
+            imagesByItem.Count);
+    }
+
+    private static int GetImageOrderPriority(string? path, string? mediaFileName)
```
JPVenson, line 196: duplicated code with BaseItemRepository

MediaBrowser.Controller/Persistence/IItemRepository.cs
```diff
@@ -35,6 +36,20 @@ public interface IItemRepository
...
+    /// <param name="itemId">The item id.</param>
+    /// <param name="imageType">The image type.</param>
+    /// <param name="index1">The first image index.</param>
+    /// <param name="index2">The second image index.</param>
+    void SwapImageSortOrder(Guid itemId, ImageType imageType, int index1, int index2)
+    {
+        // Default implementation for compatibility with external implementations
+        // that may not have been updated to support this method
+        throw new NotSupportedException("SwapImageSortOrder is not implemented for this repository.");
+    }
```
JPVenson: nowhere else we use interface implementations

JPVenson (changes requested):

Emby.Server.Implementations/Localization/Core/mr.json
```diff
@@ -132,5 +132,10 @@
     "TaskDownloadMissingLyrics": "उपलब्ध नसलेली गीतपट्टी (Lyrics) डाउनलोड करा",
     "TaskAudioNormalization": "ऑडिओ सामान्यीकरण",
     "TaskAudioNormalizationDescription": "ऑडिओ सामान्यीकरणाचा डाटा स्कॅन करतो.",
-    "TaskDownloadMissingLyricsDescription": "गाण्यांची गीतपट्टी (Lyrics) डाउनलोड करतो"
+    "TaskDownloadMissingLyricsDescription": "गाण्यांची गीतपट्टी (Lyrics) डाउनलोड करतो",
+    "TaskExtractMediaSegmentsDescription": "सक्रिय असलेल्या प्लगिनमधून मीडिया विभाग प्राप्त करते.",
+    "TaskMoveTrickplayImagesDescription": "लायब्ररीच्या सेटिंग्जप्रमाणे आधीपासून अस्तित्वात असलेल्या ट्रिकप्ले फाइल्सचे स्थान बदलते.",
+    "TaskCleanCollectionsAndPlaylistsDescription": "जे संग्रह आणि प्लेलिस्ट आता अस्तित्वात नाहीत, त्यांमधील घटक हटवते.",
+    "CleanupUserDataTask": "वापरकर्ता डेटाची स्वच्छता प्रक्रिया",
+    "CleanupUserDataTaskDescription": "९० दिवसांहून अधिक काळ अनुपस्थित असलेल्या माध्यमांवरील सर्व वापरकर्ता माहिती (जसे पाहण्याची स्थिती, आवडी इ.) हटवते."
```
nielsvanvelzen: translations should never be added via pull requests, only via translate.jellyfin.org.

seeyabye: @nielsvanvelzen thanks for the comments!
I am actually not sure why this is getting added. I made no changes to `mr.json` and from the commits, it's being automatically prepended to my fix... am I missing something?

seeyabye:
@JPVenson thanks for the reviews!

Emby.Server.Implementations/Localization/Core/mr.json
nielsvanvelzen: Looking at the PR commits you may have accidentally pulled in some commits to the master branch, probably when switching the target to the release branch. You can fix it by dropping the first two commits from your branch with an interactive rebase in git.

seeyabye: @nielsvanvelzen noted with thanks!

src/Jellyfin.Database/Jellyfin.Database.Implementations/ModelConfiguration/BaseItemImageInfoConfiguration.cs
```diff
@@ -0,0 +1,21 @@
...
+public class BaseItemImageInfoConfiguration : IEntityTypeConfiguration<BaseItemImageInfo>
+{
+    /// <inheritdoc/>
+    public void Configure(EntityTypeBuilder<BaseItemImageInfo> builder)
+    {
+        builder.HasKey(e => e.Id);
+        builder.HasOne(e => e.Item).WithMany(e => e.Images).HasForeignKey(e => e.ItemId);
+
+        // Index on SortOrder for efficient ordering queries
+        builder.HasIndex(e => new { e.ItemId, e.SortOrder });
```
JPVenson: should't that fail the moment its migrated as there would be tons and tons of duplicate index entries?

ferferga (approved):

src/Jellyfin.Database/Jellyfin.Database.Implementations/ModelConfiguration/BaseItemImageInfoConfiguration.cs
zepx: Thanks pointing this out. Indeed with only ItemId and SortOrder, this would cause a ton of duplicate index entries. I have now included ImageType as part of the index and that should resolve the duplicate index issues.

alexkahler:
How's this pr coming along? I really hope to be able to order my backdrops :)

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/Migrations/20251022214259_AddSortOrderToBaseItemImageInfo.cs
```diff
@@ -0,0 +1,47 @@
...
+                table: "BaseItemImageInfos");
+
+            migrationBuilder.AddColumn<int>(
+                name: "SortOrder",
+                table: "BaseItemImageInfos",
+                type: "INTEGER",
+                nullable: false,
+                defaultValue: 0);
+
+            migrationBuilder.CreateIndex(
```
JPVenson, line 41: this would still break no?
The SortOrder is by default 0 for all rows at this point and when you add the index then and have more then 2 images for the same type, they still have the same index.

seeyabye: You are right, they would have the same index.

I have gave it some thought and I think it's quite tricky to implement the indexing, so your advice would be most appreciated. I have currently moved the creation of the index to app migration stage rather than the ef core. This would ensure that there are no duplicates on the indexes. 

I am not sure if this is an acceptable thing, so please correct me if I am wrong.

The other alternatives that I have given some thought are:

1. First, create the 3 column indexes in the ef migration stage. This would mean tons of duplicates.
2. Perform the backfill
3. Redo the indexes in the app migration stage.

The downside of this second approach is that it's more computationally expensive to perform the re-indexing.

Given that Jellyfin performs ef and app migration separately, I couldn't introduce the indexes in the most optimal way. If you have any better ideas, it would also be much appreciated.

JPVenson: you can just run smth like this inbetween the statement to pre-populate the SortOrder correctly:

```sql
WITH ImageInfos AS 
(
    SELECT Id, ROW_NUMBER() OVER (PARTITION BY ItemId, ImageType ORDER BY DateModified) OrderId FROM BaseItemImageInfos 
)

UPDATE BaseItemImageInfos 
    SET SortOrder = (SELECT OrderId FROM ImageInfos WHERE BaseItemImageInfos.Id = ImageInfos.Id)
```

seeyabye:
I have made the following changes:

1. The ef core migration will first populate `SortOrder` using `DateModified` and `Path` (for tiebreakers)
2. The app migration will then backfill, by applying the applying priority settings
3. The outcome is that there will be **no** duplicates on the `(ItemId, ImageType, SortOrder)` index tuples

I have also verified that the following operations after migrations worked:

1. Rescan & Update. No duplicates after performing this operation. All `BaseItemImageInfos` has a `SortOrder` assigned.
2. Swapping between different backdrops.
3. Addition of backdrops through the WebUI. `SortOrder` is correctly assigned by appending the appropriate `SortOrder`.
4. Deletion of backdrops through the WebUI. `SortOrder` is correctly reassigned so that there are no gaps.

theguymadmax:
@seeyabye. It's going to be too late to merge this on the release branch, as we are preparing for a 12.0 release.  Once #16062 is merged, please rebase and target the master branch. Thanks.

theguymadmax:
@seeyabye  the above-mentioned PR has been merged. Please rebase and target the master branch.

seeyabye:
Moved to PR #16990

PR #15113, opened 2025-10-22
Add season number fallback for OMDB and TMDB plugins

### LINKED ISSUE #13997: New episodes in TV shows with no seasons don't get grouped with existing episodes
### Description of the bug

For TV series which are organized without season information, prior to #13611 episodes would be assigned to season index 1. After upgrading to 10.10.7, newly added episodes will get assigned to null season ("Unknown Season") and won't appear along with new episodes.

Refreshing metadata does not fix season assignments.

### Reproduction steps

1.  Have a TV Show without season information indexed before upgrading Jellyfin server to 10.10.7. E.g.
    ```
    Show/Show E1.mkv
    Show/Show E2.mkv
    ```
    These would've been assigned to Season 1
2.  Upgrade Jellyfin server to 10.10.7
3.  Add a new episode. It will be assigned to Season Unknown.

Episodes added before upgraded stay assigned to "Season 1", while new episodes get assigned to "Season Unknown"

New episodes should be grouped together. Either
-  New episodes keep being assigned to season 1
-  Already existing episodes are moved to season null when refreshing.

### Jellyfin Server version

10.10.0+

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.10.7

### Environment

```markdown
- OS: Linux
- Linux Kernel:
- Virtualization: 
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

```shell
N/A
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

ivanjx:
**Changes**
Add fallback value of 1 for season number when it is null.

Newer version of jellyfin does not seem to assign season number in case it is not indicated in the folder/file name and will group the episodes into "Season Unknown". This PR forces season 1 in case that happens.

**Issues**
Fixes #13997

### DIFF
--- a/MediaBrowser.Providers/Plugins/Tmdb/TV/TmdbEpisodeProvider.cs
+++ b/MediaBrowser.Providers/Plugins/Tmdb/TV/TmdbEpisodeProvider.cs
@@ -177,8 +177,8 @@ namespace MediaBrowser.Providers.Plugins.Tmdb.TV
 
             var item = new Episode
             {
-                IndexNumber = info.IndexNumber,
-                ParentIndexNumber = info.ParentIndexNumber,
+                IndexNumber = episodeNumber,
+                ParentIndexNumber = seasonNumber,
                 IndexNumberEnd = info.IndexNumberEnd,
                 Name = episodeResult.Name,
                 PremiereDate = episodeResult.AirDate,
--- a/MediaBrowser.Providers/Plugins/Omdb/OmdbProvider.cs
+++ b/MediaBrowser.Providers/Plugins/Omdb/OmdbProvider.cs
@@ -138,6 +138,8 @@ namespace MediaBrowser.Providers.Plugins.Omdb
             }
 
             var item = itemResult.Item;
+            item.IndexNumber = episodeNumber;
+            item.ParentIndexNumber = seasonNumber;
 
             var seasonResult = await GetSeasonRootObject(seriesImdbId, seasonNumber, cancellationToken).ConfigureAwait(false);
 


MediaBrowser.Providers/Plugins/Tmdb/TV/TmdbEpisodeProvider.cs
```diff
@@ -177,8 +177,8 @@ public async Task<MetadataResult<Episode>> GetMetadata(EpisodeInfo info, Cancell
 
             var item = new Episode
             {
-                IndexNumber = info.IndexNumber,
-                ParentIndexNumber = info.ParentIndexNumber,
+                IndexNumber = episodeNumber,
+                ParentIndexNumber = seasonNumber,
```
Shadowghost, line 181: I wonder if we should only fallback to this if there is none provided by the info

ivanjx: those variables are already from the info itself
https://github.com/jellyfin/jellyfin/blob/64e4dc65b1c573db943ce6cb604543b60534dc78/MediaBrowser.Providers/Plugins/Tmdb/TV/TmdbEpisodeProvider.cs#L100-L101

Shadowghost:
@ivanjx This should target the release branch of 10.11

ivanjx:
@Shadowghost is it good now?

JPVenson (approved):

PR #15126, opened 2025-10-22
Fix Has(Imdb/Tmdb/Tvdb)Id checks

MBR-0001:
**Changes**
Seems like provider ids are not all lowercase, so this fixes those checks. I've only tested with `hasTmdbId` and `HasImdbId`, tvdb is untested

**Issues**
Fixes https://github.com/jellyfin/jellyfin-plugin-tmdbboxsets/issues/110

### DIFF
--- a/Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
+++ b/Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
@@ -2354,22 +2354,22 @@ public sealed class BaseItemRepository
         if (filter.HasImdbId.HasValue)
         {
             baseQuery = filter.HasImdbId.Value
-                ? baseQuery.Where(e => e.Provider!.Any(f => string.Equals(f.ProviderId, MetadataProvider.Imdb.ToString(), StringComparison.OrdinalIgnoreCase)))
-                : baseQuery.Where(e => e.Provider!.All(f => !string.Equals(f.ProviderId, MetadataProvider.Imdb.ToString(), StringComparison.OrdinalIgnoreCase)));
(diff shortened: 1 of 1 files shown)


MBR-0001:
While looking at 10.10 code I noticed that the behavior differs
```csharp
if (query.HasTmdbId.HasValue)
{
    whereClauses.Add(GetProviderIdClause(query.HasTmdbId.Value, "tmdb"));
}

private string GetProviderIdClause(bool includeResults, string provider)
{
    return string.Format(
        CultureInfo.InvariantCulture,
        "ProviderIds {0} like '%{1}=%'",
        includeResults ? string.Empty : "not",
        provider);
}
```

seems like current implementation doesn't handle correctly the case when HasTmdbId = false so I added handling for that too

Shadowghost:
EFcore LINQ can't handle StringComparison in Where clauses, so you'd need to use toLowercase and normal equality check

JPVenson:
did you test this?
https://learn.microsoft.com/en-us/ef/core/miscellaneous/collations-and-case-sensitivity

> in addition, .NET provides overloads of [string.Equals](https://learn.microsoft.com/en-us/dotnet/api/system.string.equals) accepting a [StringComparison](https://learn.microsoft.com/en-us/dotnet/api/system.stringcomparison) enum, which allows specifying case-sensitivity and a culture for the comparison. By design, EF Core refrains from translating these overloads to SQL, and attempting to use them will result in an exception.

collation is not supported by us for sqlite. So you have to use ToLowercase == ToLowercase like ghost said

MBR-0001:
Yeah, I tested it but forgot to restart the app after making changes, sorry 😃

JPVenson (approved):

Shadowghost (approved):

crobibero (approved):

PR #15138, opened 2025-10-23
Add moveToTop option to IPlaylistManager.AddItemToPlaylistAsync

### LINKED ISSUE #15137: Add moveToTop option to IPlaylistManager.AddItemToPlaylistAsync
I have some really big playlists and sometimes when I add a new item to a playlist, I want to move it to the top immediately; otherwise, I have to add it to the playlist, then go to the playlist and then scroll all the way down and then click the Move To Top option.

I'd like to add a feature that adds an "Add to top of playlist" option in the "Add to playlist" dialog, so the first step for that is to add an option into the API.

kevgrig:
**Changes**

When adding an item to a playlist, this provides an option to move the item to the top of the playlist immediately.

**Issues**

Fixes #15137


Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -371,6 +372,7 @@ public async Task<ActionResult> RemoveUserFromPlaylist(
     public async Task<ActionResult> AddItemToPlaylist(
         [FromRoute, Required] Guid playlistId,
         [FromQuery, ModelBinder(typeof(CommaDelimitedCollectionModelBinder))] Guid[] ids,
+        [FromQuery] bool? moveToTop,
```
nielsvanvelzen: For the API design it would be better to add a "position" parameter instead of a boolean. That way a newly added item can be placed anywhere in the playlist. With 0 being the first item, 1 after the current first etc.

kevgrig: Updated; please check again when you have a chance. Thanks

JPVenson (approved):

Jellyfin.Api/Controllers/PlaylistsController.cs
```diff
@@ -359,6 +359,7 @@ public async Task<ActionResult> RemoveUserFromPlaylist(
     /// </summary>
     /// <param name="playlistId">The playlist id.</param>
     /// <param name="ids">Item id, comma delimited.</param>
+    /// <param name="position">Optional. 0-based index where to place the items or at the end if null.</param>
```
Bond-009:
```suggestion
    /// <param name="position">Optional. 0-based index where to place the items or at the end if <c>null</c>.</param>
```

PR #15159, opened 2025-10-23
Added Endpoint to Get the Items on a SyncPlay Queue

PeachesMLG:
**Changes**
The current syncplay implementation has endpoints to Create a Queue and add/remove items from the queue, aswell as alerting other users through websockets when the queue is modified. However there is no way to get the current state of the queue through REST. (Especially useful if a new User just joined a group and wants to know what is on the queue)

This change simply adds in that GET SyncPlay/Queue endpoint

The method names here are fairly inconsistent between Queue and Playlist so im not sure which one is preferred, so went with Queue

It also Enhances the QueueItem model to include the BaseItemDto model so clients dont have to make extra api requests to see details of each item in the queue

### DIFF
--- a/Jellyfin.Api/Controllers/SyncPlayController.cs
+++ b/Jellyfin.Api/Controllers/SyncPlayController.cs
@@ -222,6 +222,21 @@ public class SyncPlayController : BaseJellyfinApiController
     }
 
     /// <summary>
+    /// Gets the current SyncPlay queue.
+    /// </summary>
+    /// <response code="200">Queue returned.</response>
+    /// <returns>An <see cref="IEnumerable{SyncPlayQueueItem}"/> containing the current queue items.</returns>
+    [HttpGet("Queue")]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    [Authorize(Policy = Policies.SyncPlayIsInGroup)]
+    public async Task<ActionResult<IEnumerable<SyncPlayQueueItem>>> SyncPlayGetQueue()
+    {
+        var currentSession = await RequestHelpers.GetSession(_sessionManager, _userManager, HttpContext).ConfigureAwait(false);
+        var queue = _syncPlayManager.GetQueue(currentSession) ?? [];
+        return Ok(queue.AsEnumerable());
+    }
+
+    /// <summary>
     /// Request unpause in SyncPlay group.
     /// </summary>
     /// <response code="204">Unpause update sent to all group members.</response>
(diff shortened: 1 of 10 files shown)


PeachesMLG:
Tested locally and working

image


Same with the websockets

image


All changes are backwards compatable and dont break any pre-existing schemas, only enhancing them with new information if a client wants to use this information

Jellyfin.Api/Controllers/SyncPlayController.cs
```diff
@@ -221,6 +221,21 @@ public async Task<ActionResult> SyncPlayQueue(
...
 
+    /// <summary>
+    /// Gets the current SyncPlay queue.
+    /// </summary>
+    /// <response code="200">Queue returned.</response>
+    /// <returns>An <see cref="IEnumerable{SyncPlayQueueItem}"/> containing the current queue items.</returns>
+    [HttpGet("Queue")]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    [Authorize(Policy = Policies.SyncPlayIsInGroup)]
+    public async Task<ActionResult<IEnumerable<SyncPlayQueueItem>>> SyncPlayGetQueue()
```
nielsvanvelzen: Avoid returning arrays as top-level in API responses. Wrap them in an object instead for potential extensibility in the future.

In this case I believe we already have a generic class for lists that would make it consistent with other endpoints returning lists.

PeachesMLG: Very good point, I was also wondering if we may want to extend this object and get the actual Item Object from ItemId.

Means if a Client actually wants to implement some sort of Interface around a Playlist/Queue they dont have to then do multiple API calls for each Id.


Will have a bit more of a think around this later

Jellyfin.Api/Controllers/SyncPlayController.cs
```diff
@@ -221,6 +222,21 @@ public async Task<ActionResult> SyncPlayQueue(
...
+    /// <returns>A <see cref="QueryResult{SyncPlayQueueItem}"/> containing the current queue items.</returns>
+    [HttpGet("Queue")]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    [Authorize(Policy = Policies.SyncPlayIsInGroup)]
+    public async Task<ActionResult<QueryResult<SyncPlayQueueItem>>> SyncPlayGetQueue()
+    {
+        var currentSession = await RequestHelpers.GetSession(_sessionManager, _userManager, HttpContext).ConfigureAwait(false);
+        var queue = _syncPlayManager.GetQueue(currentSession);
+        return Ok(new QueryResult<SyncPlayQueueItem>(queue));
+    }
```
PeachesMLG: I want to try enhance this with a BaseItemDto instead of the ID. So if a client wants to utilise this endpoint they dont have to do extra API queries to get the Item from the Id

IDisposable (approved):

PR #15187, opened 2025-10-24
Fix pagination and sorting for folders

### LINKED ISSUE #15142: 10.11.0 - Library sort-by option sorts each page individually instead of the library as a whole
### Description of the bug

After upgrading to 10.11.0, sorting a library that is large enough to span multiple pages sorts only the current page.

What looks like is happening (with sort by Date Added as the example) is that pagination and sorting is happening out of order, so the library is paginated based on the original library order (alphabetical) and the sort is then being applied to the current page. The result is that each page has an alphabetically correct set of content i.e. page 1 has videos all starting with the letter "A", but the order is based on the Date Added for each item on that page instead of across the entire library.

Choosing Ascending or Descending reorders the current page, but doesn't bring in any items from the rest of the library.

My library has a few thousand videos, I can't confirm if this happens on smaller libraries.

### Reproduction steps

1. Have a library with multiple pages of content, added over a period of time.
2. Change sort-by to "Date Added", ascending or descending both have this issue.

Each individual page is sorted instead of the entire paginated library.

Sort should sort the entire library, not just the current page.

### Jellyfin Server version

10.10.0+

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.0

### Environment

```markdown
- OS: Synology DSM
- Virtualization: Docker
- Clients: Browser, iOS
- Browser: Firefox, Chrome
```

### Jellyfin logs

```shell
no relevant logs available for the sort option
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

### LINKED ISSUE #15201: Grouping of "Movies" and "TV Shows" not working properly
### Description of the bug

I have roughly 3600 Movies distributed in 3 different movie type libraries. Two of them i grouped on the "front page view" setup. When accessing this combined library i only can see the first 100 entries (it only lists 100 entries in total). For the single libraries (not combined) it's working normal. Please see attached pictures for what i mean, can't explain it better in english.
The same happens for  TV Shows type libraries. Uncombined all tV shows are visible, combined only 100 TV Shows are visible.

### Reproduction steps

1. Have at least 2 different libraries of the same content type (Movies or TV Shows) with more the 100 entries
2. Combine these two libraries in the "front page setup"

In combined librariy views only 100 entries are visible, althoug, in my case, there should be around 2000 entries.

You should be able to see all Movies / TV Show entries in the combined view.

### Jellyfin Server version

10.10.0+

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.0

### Environment

```markdown
- OS:Windows 11
- Linux Kernel:
- Virtualization:
- Clients:
- Browser:Firefox
- FFmpeg Version:bundled with Jellyfin
- Playback Method:
- Hardware Acceleration:NVENC
- GPU Model:RTX 3070 M
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Jellyfin Data Storage:
- Media Storage:
- External Integrations:
```

### Jellyfin logs

```shell
no specific entry in the log
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

Uncombined:

Image

Image

Image

Combined:

Image

Image

Image

### Additional information

_No response_

Shadowghost:
**Changes**
* Make sure `PostFilterAndSort` gets the correct pagination params applied

**Issues**
Fixes #15142
Fixes #15201

### DIFF
--- a/MediaBrowser.Controller/Entities/Folder.cs
+++ b/MediaBrowser.Controller/Entities/Folder.cs
@@ -715,9 +715,18 @@ namespace MediaBrowser.Controller.Entities
                 }
                 else
                 {
-                    items = GetRecursiveChildren(user, query, out totalCount);
+                    // Save pagination params before clearing them to prevent pagination from happening
+                    // before sorting. PostFilterAndSort will apply pagination after sorting.
+                    var limit = query.Limit;
+                    var startIndex = query.StartIndex;
                     query.Limit = null;
-                    query.StartIndex = null; // override these here as they have already been applied
+                    query.StartIndex = null;
+
+                    items = GetRecursiveChildren(user, query, out totalCount);
+
+                    // Restore pagination params so PostFilterAndSort can apply them after sorting
+                    query.Limit = limit;
+                    query.StartIndex = startIndex;
                 }
 
                 var result = PostFilterAndSort(items, query);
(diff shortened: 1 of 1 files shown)


crobibero (approved):
Not ideal to need to load all children but a much larger refactor would be needed to properly fix it.

Users with large/many collections will likely be the only ones to experience any slowdown

Shadowghost:
In my tests only the first time the collections lib was loaded it really slow, afterwards it's rather fast due to server-side caching. 20 secs on first access vs <5s on following

PR #15209, opened 2025-10-25
Improve symlink handling

### LINKED ISSUE #14899: Symlinked artwork does not refresh when target file changes
### Description of the bug

In a TV show folder, if `folder.jpg` is a symlink pointing to `season1-poster.jpg`, Jellyfin does not update the show’s main poster when the target file is changed.

For instance, after setting up the symlink, both the show poster and the Season 1 poster initially display correctly. If I then replace the contents of `season1-poster.jpg` with a new image and run a metadata refresh, Jellyfin correctly updates the Season 1 poster, but the show poster remains stuck on the old image even though it points to the same updated file. The only way to force an update is to recreate or touch the symlink itself, which is not practical.

Although the issue is easy to reproduce with a `folder.jpg` pointing to `season1-poster.jpg`, it is likely more general. Jellyfin seems to detect changes to the symlink itself but not to the file that the symlink points to, which causes inconsistent updates whenever artwork relies on symbolic links.  

### Reproduction steps

Create a symlink `folder.jpg` pointing to `season1-poster.jpg` in a series folder. Let Jellyfin scan the library so both posters are loaded. Update the `season1-poster.jpg` file by overwriting it with new content while keeping the same filename. Run a `Refresh Metadata/Scan for new and updated files` on the series. 

The Season 1 poster updates, but the main show poster remains stale, despite the symlink pointing to the updated file.  

Both poster images are updated.

### Jellyfin Server version

10.10.0+

### Specify commit id

_No response_

_No response_

### Specify version number

10.10.7

### Specify the build version

Release

### Environment

```markdown
- OS: Arch Linux
- Linux Kernel: 6.16-seblu
- Virtualization: None
- Browser: Firefox
```

### Jellyfin logs

```shell
There is no log related to refreshing metadata.
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

Shadowghost:
**Changes**
* Use `fileInfo.LinkTarget` for symlink checks
* Use `fileInfoe.ResolveLinkTarget` to get the actual file

**Issues**
Fixes #14899

### DIFF
--- a/Emby.Server.Implementations/IO/ManagedFileSystem.cs
+++ b/Emby.Server.Implementations/IO/ManagedFileSystem.cs
@@ -254,18 +254,22 @@ namespace Emby.Server.Implementations.IO
 
                 if (info is FileInfo fileInfo)
                 {
+                    result.CreationTimeUtc = GetCreationTimeUtc(info);
+                    result.LastWriteTimeUtc = GetLastWriteTimeUtc(info);
                     if (fileInfo.LinkTarget is not null)
                     {
-                        result.IsSymlink = true;
                         try
                         {
                             var targetFileInfo = (FileInfo?)fileInfo.ResolveLinkTarget(returnFinalTarget: true);
                             if (targetFileInfo is not null)
                             {
                                 result.Exists = targetFileInfo.Exists;
-                                result.Length = targetFileInfo.Length;
-                                result.CreationTimeUtc = GetCreationTimeUtc(targetFileInfo);
-                                result.LastWriteTimeUtc = GetLastWriteTimeUtc(targetFileInfo);
+                                if (result.Exists)
+                                {
+                                    result.Length = targetFileInfo.Length;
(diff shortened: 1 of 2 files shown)


Jellyfin.Server/Startup.cs
```diff
@@ -69,8 +66,6 @@ public void ConfigureServices(IServiceCollection services)
                 options.HttpsPort = _serverApplicationHost.HttpsPort;
             });
 
-            // TODO remove once this is fixed upstream https://github.com/dotnet/aspnetcore/issues/34371
```
JPVenson, line 72: was this fixed upstream? issue is still open

crobibero: My understanding from the change is that the issue is mitigated since we now return the destination link instead of the origin link

Shadowghost: correct, we're now directly using the link target path

JPVenson (approved):

PR #15212, opened 2025-10-25
Skip invalid database migration

crobibero:
The migration creates the file and then attempts to read which doesn't work if the table doesn't exist


martimarkov:
I was looking at the different Routines and I think `Jellyfin.Server/Migrations/Routines/MigrateUserDb.cs` is doing the exact same thing, no?

JPVenson (approved):
i still think we should make this a noop

Issue #15174, opened 2025-10-24
[10.11.0] "Air before season" and "Air before episode" are not respected


alexkahler:
### Description of the bug

Specials do not play as defined by the metadata. E.g., if metadata designates an episode to play "Before Season 1, Episode 8" the special is queued before entire season 1. Simiarly, if the metadata states that it should play "After season 1" and "Before Season 2", special episode will still play before season 1.

- "Release date" does not matter or have an impact on the play queue.
- Changing the metadata to "Airs before season 2" does not affect the play queue. 
- When playing a series with the "Next Up", it skips any episodes found in the "Specials" season.
- Changing the series "Display order" does not have any effect.

### Reproduction steps

1. Designate a special episode to play in the middle of an unwatched season, e.g., Season 1, Episode 8, with the "airs before episode" metadata field, and "Season 1" should have more than 8 episodes.
2. Click "Play" icon for the series (not the "Next Up" that designates S1:E1 as the starting episode)
3. The Season 0 special episode will play as the first episode.

### What is the current _bug_ behavior?

Season 0 specials always play as the first episode in the queue, despite having defined "Airs before episode 8". Specifying "Air before season", "Air after season", or a later "Release date" does not have any effect.

### What is the expected _correct_ behavior?

Episodes in the "Specials" season should play as defined by "Airs before season" and "Air before episode" in the middle of a season, if e.g., "Air before episode" is in the middle of a season.

### Jellyfin Server version

10.10.0+

### Specify the build version

10.11.0

### Environment

```markdown
- OS:
- Linux Kernel:
- Virtualization: Docker
- Clients: Browser
- Browser: Brave
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Jellyfin Data Storage:
- Media Storage:
- External Integrations:
```

### Jellyfin logs

```shell
No relevant logs for play queue.
```

### FFmpeg logs

```shell

```

### Relevant screenshots or videos

Special episode set to air before Season 1, Episode 8:

<img width="840" height="165" alt="Image" src="https://github.com/user-attachments/assets/152cf376-44d8-4a5c-b6b7-9913bed99ad9" />

Playing the entire series from the "Play" icon on the Series details page (note all items in the series are unwatched):

[...]

PR #15251, opened 2025-10-27
Fixed missing sort?

JPVenson:
Readded missing sort arg

fixes: https://github.com/jellyfin/jellyfin/issues/15174

### DIFF
--- a/tests/Jellyfin.Server.Implementations.Tests/Item/OrderMapperTests.cs
+++ b/tests/Jellyfin.Server.Implementations.Tests/Item/OrderMapperTests.cs
@@ -12,7 +12,7 @@ public class OrderMapperTests
     [Fact]
     public void ShouldReturnMappedOrderForSortingByPremierDate()
     {
-        var orderFunc = OrderMapper.MapOrderByField(ItemSortBy.PremiereDate, new InternalItemsQuery()).Compile();
+        var orderFunc = OrderMapper.MapOrderByField(ItemSortBy.PremiereDate, new InternalItemsQuery(), null!).Compile();
 
         var expectedDate = new DateTime(1, 2, 3);
         var expectedProductionYearDate = new DateTime(4, 1, 1);


ItsOnlyBinary:
I've tested this against my library and it returns the same order as 10.10.7. Greatly appreciated @JPVenson :)

Regarding performance mentioned in discord. There is a big different between 10.10.7 and this branch. I measured <1s~ on my 10.10.7 instance (running in docker on old i5 2600k) and 9.64s~ for this pr (running as dev on i9 9900k).

I'm happy to pay the small waiting tax to not loose this feature. But maybe for a long term solution an entry for the show could be added in `UserData` to store the `LastPlayedDate` and updated when an episode’s `LastPlayedDate` is updated. Not having to evaluate every episode to get LastPlayedDate will surely solve the performance issue.

ItsOnlyBinary:
I'm not sure if the current implementation will have issues with multiple users (I only have 1 user), in the old query the `LEFT JOIN` had a `UserId` filter. The resultant query from this does not contain that.

I've also been digging into the performance issues with this commit. It seems that the issue is cause by the resultant query using nested `SELECT` which breaks the advantages created by the Indexes. I did recreate the old query against the new db structure to confirm that there is no performance loss caused by the new structure and indexes.

I have also noticed that the current query returns multiple duplicate id's for each show (I'm assuming one for each episode) due to the outer `SELECT` and `LEFT JOIN`'s not having a `GROUP BY` but in direct database testing this does not appear to affect performance or cause any noticeable issues with the functionality in the client.


Here are the queries used by 10.10 & 10.11 jellyfin: https://gist.github.com/ItsOnlyBinary/e118693c7dd58e9fa54f623cd470f38d

JPVenson:
The performance issues come from an O(N) lookup for each baseitem we search though every other BaseItem to find matching episodes. To fix that we would need to construct a syntetic column in the outer query which would be much much faster. On the user related lookup, thats an somewhat easy fix i will do that later

ItsOnlyBinary:
As I don't know C# but wanted to try and help anyway I can, I threw the performance issue at Claude-Sonnet-4.5 LLM based on your last reply. Hopefully the response will help you more quickly implement the changes needed when you have time. I have tested the resultant code and it does return performance to 10.10 levels and maintains the sort order. I don't expect any recognition for this code as it is not my own. 

https://github.com/JPVenson/jellyfin/compare/bugfix/SeriesDatePlayed...ItsOnlyBinary:jellyfin:Bugfix/SeriesDatePlayed-Performance

JPVenson:
i already tested it like this but we decided internally against that as its too much of a rewrite rn. maybe for 10.12

JPVenson:
Closed for https://github.com/jellyfin/jellyfin/pull/15413

PR #15254, opened 2025-10-27
Update password reset to always return the same response structure

thornbill:
**Changes**
Fixes user enumeration via password reset. The password reset process would return different responses based on whether a user account exists or not allowing enumeration of all existing user accounts.

Credit for discovery to @maru801 and Carlos García-Olalla Olivera 

**Issues**
Security issue reported via email

### DIFF
--- a/Jellyfin.Server.Implementations/Users/DefaultPasswordResetProvider.cs
+++ b/Jellyfin.Server.Implementations/Users/DefaultPasswordResetProvider.cs
@@ -96,45 +96,35 @@ namespace Jellyfin.Server.Implementations.Users
         public async Task<ForgotPasswordResult> StartForgotPasswordProcess(User? user, string enteredUsername, bool isInNetwork)
         {
             DateTime expireTime = DateTime.UtcNow.AddMinutes(30);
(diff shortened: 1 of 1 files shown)


Jellyfin.Server.Implementations/Users/DefaultPasswordResetProvider.cs
```diff
@@ -92,13 +92,26 @@ public async Task<PinRedeemResult> RedeemPasswordResetPin(string pin)
...
+
+            if (user is null || !isInNetwork)
+            {
+                // If an invalid user is requested or the request is made outside of local network, return a response
+                // using a random GUID in place of the user ID.
+                return new ForgotPasswordResult
+                {
+                    Action = ForgotPasswordAction.PinCode,
+                    PinExpirationDate = expireTime,
+                    PinFile = _passwordResetFileBase + Guid.NewGuid() + ".json"
```
crobibero: Since this returns a random guid it is still possible to enumerate users by seeing if the file path changes

chicheng:
Still has timing vulnerability? System will respond faster if user not exists?

PR #15263, opened 2025-10-28
Resolve symlinks for static media source infos

revam:
**Changes**
This change will prevent issues like _trying to Direct Play a symbolic link resulting in an out of range exception_ at the deepest possible level instead of in the API layer — in which it was handled before #15209 removed the symbolic link handler from the API — to ensure both the API and any internal processes using the media sources will always use a real file and never a symbolic link.

As part of this change, I swept through the code base to look for potential issues this may introduce, and found that the trickplay manager is checking the parent of the media source when checking if the video is in a backdrop folder, so i changed that to instead use the video path. I did not find any other issues in regards to this change.

**Issues / PRs**
Follow up for https://github.com/jellyfin/jellyfin/pull/15209

Player errs out when you try to play a symlinked file on 10.11.1;
```
[00:17:47] [ERR] [043] Jellyfin.Api.Middleware.ExceptionMiddleware: Error processing request. URL GET /Videos/1e295cae3782876cf0ac5cccb8ec1d85/stream.mkv.
System.ArgumentOutOfRangeException:  (Parameter 'count')
Actual value was 965864780.
   at Microsoft.AspNetCore.Http.SendFileFallback.SendFileAsync(Stream destination, String filePath, Int64 offset, Nullable`1 count, CancellationToken cancellationToken)
   at Microsoft.AspNetCore.Http.SendFileResponseExtensions.SendFileAsyncCore(HttpResponse response, String fileName, Int64 offset, Nullable`1 count, CancellationToken cancellationToken)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.<InvokeNextResultFilterAsync>g__Awaited|30_0[TFilter,TFilterAsync](ResourceInvoker invoker, Task lastTask, State next, Scope scope, Object state, Boolean isCompleted)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.Rethrow(ResultExecutedContextSealed context)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.ResultNext[TFilter,TFilterAsync](State& next, Scope& scope, Object& state, Boolean& isCompleted)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.InvokeResultFilters()
--- End of stack trace from previous location ---
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.<InvokeNextResourceFilter>g__Awaited|25_0(ResourceInvoker invoker, Task lastTask, State next, Scope scope, Object state, Boolean isCompleted)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.Rethrow(ResourceExecutedContextSealed context)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.Next(State& next, Scope& scope, Object& state, Boolean& isCompleted)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.InvokeFilterPipelineAsync()
--- End of stack trace from previous location ---
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.<InvokeAsync>g__Awaited|17_0(ResourceInvoker invoker, Task task, IDisposable scope)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.<InvokeAsync>g__Awaited|17_0(ResourceInvoker invoker, Task task, IDisposable scope)
   at Jellyfin.Api.Middleware.ServerStartupMessageMiddleware.Invoke(HttpContext httpContext, IServerApplicationHost serverApplicationHost, ILocalizationManager localizationManager) in /home/revam/Projects/Jellyfin/Jellyfin.Api/Middleware/ServerStartupMessageMiddleware.cs:line 41
   at Jellyfin.Api.Middleware.WebSocketHandlerMiddleware.Invoke(HttpContext httpContext, IWebSocketManager webSocketManager) in /home/revam/Projects/Jellyfin/Jellyfin.Api/Middleware/WebSocketHandlerMiddleware.cs:line 33
   at Jellyfin.Api.Middleware.IPBasedAccessValidationMiddleware.Invoke(HttpContext httpContext, INetworkManager networkManager) in /home/revam/Projects/Jellyfin/Jellyfin.Api/Middleware/IpBasedAccessValidationMiddleware.cs:line 41
   at Microsoft.AspNetCore.Authorization.AuthorizationMiddleware.Invoke(HttpContext context)
   at Jellyfin.Api.Middleware.QueryStringDecodingMiddleware.Invoke(HttpContext httpContext) in /home/revam/Projects/Jellyfin/Jellyfin.Api/Middleware/QueryStringDecodingMiddleware.cs:line 36
   at Swashbuckle.AspNetCore.ReDoc.ReDocMiddleware.Invoke(HttpContext httpContext)
   at Swashbuckle.AspNetCore.SwaggerUI.SwaggerUIMiddleware.Invoke(HttpContext httpContext)
   at Swashbuckle.AspNetCore.Swagger.SwaggerMiddleware.Invoke(HttpContext httpContext, ISwaggerProvider swaggerProvider)
   at Microsoft.AspNetCore.Authentication.AuthenticationMiddleware.Invoke(HttpContext context)
   at Jellyfin.Api.Middleware.RobotsRedirectionMiddleware.Invoke(HttpContext httpContext) in /home/revam/Projects/Jellyfin/Jellyfin.Api/Middleware/RobotsRedirectionMiddleware.cs:line 43
   at Jellyfin.Api.Middleware.LegacyEmbyRouteRewriteMiddleware.Invoke(HttpContext httpContext) in /home/revam/Projects/Jellyfin/Jellyfin.Api/Middleware/LegacyEmbyRouteRewriteMiddleware.cs:line 51
   at Jellyfin.Api.Middleware.ResponseTimeMiddleware.Invoke(HttpContext context, IServerConfigurationManager serverConfigurationManager) in /home/revam/Projects/Jellyfin/Jellyfin.Api/Middleware/ResponseTimeMiddleware.cs:line 66
   at Jellyfin.Api.Middleware.ExceptionMiddleware.Invoke(HttpContext context) in /home/revam/Projects/Jellyfin/Jellyfin.Api/Middleware/ExceptionMiddleware.cs:line 55
```
image

### DIFF
--- a/MediaBrowser.Controller/IO/FileSystemHelper.cs
+++ b/MediaBrowser.Controller/IO/FileSystemHelper.cs
@@ -87,9 +87,13 @@ public static class FileSystemHelper
             return File.ResolveLinkTarget(linkPath, returnFinalTarget: false) as FileInfo;
         }
 
-        FileInfo? targetInfo = null;
-        var currentPath = linkPath;
-        var visited = new HashSet<string>(StringComparer.Ordinal) { linkPath };
+        if (File.ResolveLinkTarget(linkPath, returnFinalTarget: false) is not FileInfo targetInfo)
+        {
+            return null;
+        }
+
+        var currentPath = targetInfo.FullName;
+        var visited = new HashSet<string>(StringComparer.Ordinal) { linkPath, currentPath };
         while (File.ResolveLinkTarget(currentPath, returnFinalTarget: false) is FileInfo linkInfo)
         {
             var targetPath = linkInfo.FullName;


Shadowghost (commented):
Using video.Path instead of mediaPath is a behavior change which might be incorrect (e.g. on multiple versions). We're trying to use the dedicated first media source here.

revam:
In my setup, which uses a symbolic-linked structure, checking the media source in the underlying file system causes the backdrop check in the trickplay manager to fail for actual backdrop videos within the symbolic-linked structure. This happens because my underlying structure doesn’t follow a Jellyfin‑sanctioned layout — which is why I provide a symbolic‑linked structure to Jellyfin in the first place.

That’s why I changed the logic to check the video’s path rather than the media source’s. I believe the decision on whether a video is a backdrop should be based on its location within the structure Jellyfin sees, not on the media source — which, in my case, lies outside that structure.

If this change introduces too much confusion, or you think it’s better to ensure the media source always stays in sync with the base item, I can close this PR and open a new one to revert the removal of the symbolic link handling logic in the API layer. That would also restore correct direct‑play handling for symbolic links (which was my original reason for opening this PR), without introducing potentially breaking internal changes.

Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
```diff
@@ -254,10 +254,10 @@ private async Task RefreshTrickplayDataInternal(
                 }
 
                 // We support video backdrops, but we should not generate trickplay images for them
-                var parentDirectory = Directory.GetParent(mediaPath);
+                var parentDirectory = Directory.GetParent(video.Path);
```
Shadowghost, line 257: Can you check if this works if you have a manually grouped video where the two versions are in different folders?

revam: One is in a backdrops directory and one isn't? Or both either in or not in two separate backdrop directories? Even without testing I know the former case will be a gamble depending on which version is being checked, while the latter case should work since both versions are either in or not in a backdrop directory.

revam: I've also noticed the final destination for the trickplay directory depends entirely on the video and not the media source (as shown on line 276 below). So checking if the trickplay directory should exist based on the media source and later determining it's destination based on the video seems weird to me. But then again, I'm not an expert on the matter, so I don't ultimately know why the first check was done on the media source, and I may have broken something by changing it.

revam:
Thinking about it more, I'll close this PR and instead open an issue about "Direct Play not working on symbolic links."

I probably went too deep into the core to try and fix it, and there might still be something else relying on the media source info's and video base item's path being in sync for all I know. If somebody wants to repoen this then feel free to, but for now I'm closing it.

I already appied a (hopefully only) temp. fix for the plugin to add back resolution of said symbolic links in the API layer until the API layer is (hopefully) fixed in the core, so it is not an as urgent matter now as it was when 10.11.1 was released.

revam:
I'll re-open this and let the team decide. Didn't notice @Shadowghost had already linked an issue to this PR, and only saw it when I went searching to open a new one. 😅 Next time I will try not to think too deeply about complex stuff on my own before letting others share their thoughts.

---

I also just tested out displaying the multiple versions in the UI, since I was concerned it would show the source files and not symbolic links, but it seems to not be the case and it's properly showing the symbolic link names in the multiple versions selector in UI as intended. I was going to test the trickplay generation on them, but still not quite sure how to trigger it for a single video, so figuring that out right now to properly test what Tim asked for earlier.

Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
revam: > Can you check if this works if you have a manually grouped video where the two versions are in different folders?

While trying to debug generating trickplay for a video with multiple sources I noticed it will always get the media source that have the same id as the video (as shown on line 241), so it doesn't matter how many you have grouped and where they're located as it will always pick the media source for that particular video base item.

Shadowghost:
Makes sense: video and mediaSource are interchargable in the code you modified, see the code comment:
```
// Note: Media sources under parent items exist as their own video/item as well. Only use this video stream for trickplay.
var mediaSource = video.GetMediaSources(false).FirstOrDefault(source => Guid.Parse(source.Id).Equals(video.Id));
````

So your fix should be fine since any non-extraction logic relies on file structure (so location without symlink resolution) and only extraction logic would need the full path.

**Review by Shadowghost — APPROVED** [review]

**Comment by revam** [comment]
Forgot it could be attempting to resolve other types of media, so I added a guard to only attempt to resolve file paths.

File: MediaBrowser.Controller/Entities/BaseItem.cs [hunk 2]
```diff
@@ -1127,14 +1127,23 @@ private MediaSourceInfo GetVersionInfo(bool enablePathSubstitution, BaseItem ite
...
             var protocol = item.PathProtocol;
 
+            // Resolve the item path so everywhere we use the media source it will always point to
+            // the correct path even if symlinks are in use. Calling ResolveLinkTarget on a non-link
+            // path will return null, so it's safe to check for all paths.
+            var itemPath = item.Path;
+            if (protocol is MediaProtocol.File && File.ResolveLinkTarget(itemPath, returnFinalTarget: true) is { } linkInfo)
+            {
+                itemPath = linkInfo.FullName;
+            }
```
Shadowghost: This can now throw IO Exceptions, not sure what to do about it though. maybe something like this?
```  
            if (protocol == MediaProtocol.File)
            {
                try
                {
                    var linkFileInfo = File.ResolveLinkTarget(itemPath, returnFinalTarget: true);
                    if (linkFileInfo is not null && linkFileInfo.Exists)
                    {
                        itemPath = linkFileInfo.FullName;
                    }
                }
                catch (IOException)
                {
                    // Ignore, so we don't fail the removal process
                }
            }
```

Zofware:
I'm still having trouble with symlinks, even with this PR applied.

With the original ["Improve symlink handling" change](https://github.com/jellyfin/jellyfin/pull/15209), I had these problems with videos referenced by symlinks to UNC paths:

1. Video files could no longer be opened directly, resulting in playback failures or, at best, playback via transcoding when direct playback was previously available.
2. The Scan Library scheduled task failed to resolve symlinks to UNC paths, resulting in all of those videos being nuked from the library.

With this PR applied, problem 1 above still happens. The errors in the log make it seem like the path resolution is failing in a strange way by adding `\UNC\` to the middle of the path for unknown reasons. For example, with a video file at `\\myserver\myshare\myfolder\myvideo.mp4` and referenced in the library via symlink, the error message looks like this:

```
[14:25:23] [ERR] [25] Jellyfin.Api.Middleware.ExceptionMiddleware: Error processing request: Could not find file 'X:\jellyfin\jellyfin\Jellyfin.Server\UNC\myserver\myshare\myfolder\myvideo.mp4'. URL GET /Videos/<some-uuid>/stream.mp4.
```

Problem 2 might still happen, but I still need to confirm.

Zofware:
Ah, yeah. So [this issue](https://github.com/dotnet/runtime/issues/92128) still seems to be present in .NET. That's where the weird `UNC` prefix is coming from. I think we still need `SymlinkFollowingPhysicalFileResultExecutor.cs` and `FileSystemInfo.ResolveLinkTarget` still isn't ready for prime time.

revam:
Reading the linked issue, it seems like doing the traversal of the links ourselves might work — since it seems to be doing a different path depending on if the return final argument is set — though that might become slightly more expensive for nested links compared to the presumably faster path when the argument is set to true, but I'm not sure.

@Zofware can you test with the below loop to traverse the links and see if it works for UNC paths? I don't have any windows PCs at home atm to test with, but I can try to set up a VM test bench this weekend if you can't test.

Simple loop code;
```cs
            if (protocol is MediaProtocol.File)
            {
                try
                {                    
                    while (File.ResolveLinkTarget(itemPath, returnFinalTarget: false) is { Exists: true } linkInfo)
                    {
                        itemPath = linkInfo.FullName;
                    }
                }
                catch (IOException)
                {
                    // Ignore, so we don't fail the removal process and don't interfere with any API requests.
                    itemPath = item.Path;
                }
            }
```

Zofware:
@revam Thanks for that code fragment. I'll give it a try and report my findings. 👍

Zofware:
@revam, I applied your code fragment to the `ResolveLinkTarget()` call in `BaseItem.cs` in the branch for this PR. As you predicted, the iterative approach prevents the resulting path from erroneously containing any `"UNC"` strings. It fixes my first issue (videos being transcoded unnecessarily or failing to open).

However, my second issue (library scan nuking videos) still happens, presumably because we also need to apply the iterative pattern to the `ResolveLinkTarget()` call in `ManagedFileSystem.cs`. I'll try that.

Zofware:
Yes, applying the iterative pattern to the `ResolveLinkTarget()` call in `ManagedFileSystem.cs` solves the problem of library scanning nuking the library contents. Cool.

There's one other call to `ResolveLinkTarget()` and it's in `DotIgnoreIgnoreRule.cs`. I haven't tried modifying it since I don't have a ready test case for ignore rules.

revam:
@Zofware thanks for the help.

If it's fine by @Shadowghost (and the rest of the team) then I can modify the other places resolving symlinks with a new helper with some context for why said helper is used by adding the runtime issue link and a short explanation.

Shadowghost:
Feel free to do so. I'm just annoyed that .NET keeps having this weird, unreliable implementations. Like why do you expose this if it doesn't work reliably...?

Zofware:
@revam, I had one other thought about the library scan's failure to resolve symlinks causing videos to be removed from the library. What about the scenario where a temporary network outage causes the symlink resolution to fail? In that case, we probably don't want to permanently remove the item from the library. Can this somehow be accounted for in the logic? Thanks.

revam:
> @revam, I had one other thought about the library scan's failure to resolve symlinks causing videos to be removed from the library. What about the scenario where a temporary network outage causes the symlink resolution to fail? In that case, we probably don't want to permanently remove the item from the library. Can this somehow be accounted for in the logic? Thanks.

Jellyfin won't scan a media folder if it's not mounted afaik.

Zofware:
That makes sense. My scenario is a little different though. The media folder and all the metadata files are local, but the video files themselves are referenced individually via symlinks to physical files on a NAS. What happens in that case if the network goes down momentarily during a scan?

Shadowghost:
That's another issue I would not fix in this PR. And IMO this is out of Jellyfin's responsibility

Zofware:
Fair. Thanks.

JPVenson (approved):

revam:
> That makes sense. My scenario is a little different though. The media folder and all the metadata files are local, but the video files themselves are referenced individually via symlinks to physical files on a NAS. What happens in that case if the network goes down momentarily during a scan?

If i were to guess, then Jellyfin would momentarily mark the file(s) as not existing, without removing the actual link(s), and then they would appear again during the next scan the NAS is available. And I'm with @Shadowghost that this seems like it lies outside Jellyfin's responsibilities to handle.

Shadowghost:
@revam I think you still need to switch to the iterative approach int he code. That aside, I think this PR should be ready to go.

revam:
@Shadowghost should be good to go.

MediaBrowser.Controller/IO/FileSystemHelper.cs
```diff
@@ -61,4 +61,77 @@ public static void DeleteEmptyFolders(IFileSystem fileSystem, string path, ILogg
...
+    /// </returns>
+    public static FileInfo? ResolveLinkTarget(string linkPath, bool returnFinalTarget = false)
+    {
+        // Check if the file exists so the native resolve handler won't throw at us.
+        if (!File.Exists(linkPath))
+        {
+            return null;
+        }
+
+        var targetInfo = (FileInfo?)null;
```
Shadowghost: FileInfo? targetInfo = null;

MediaBrowser.Controller/IO/FileSystemHelper.cs
```diff
@@ -61,4 +61,77 @@ public static void DeleteEmptyFolders(IFileSystem fileSystem, string path, ILogg
...
         }
     }
+
+    /// <summary>
+    /// Gets the target of the specified file link.
+    /// </summary>
+    /// <remarks>
+    /// This helper exists because of this upstream runtime issue; https://github.com/dotnet/runtime/issues/92128.
+    ///
+    /// It is adapted to suit the use case of the consumers of this helper.
```
Shadowghost: Can be removed

MediaBrowser.Controller/IO/FileSystemHelper.cs
```diff
@@ -61,4 +61,77 @@ public static void DeleteEmptyFolders(IFileSystem fileSystem, string path, ILogg
...
+        if (!File.Exists(linkPath))
+        {
+            return null;
+        }
+
+        var targetInfo = (FileInfo?)null;
+        if (returnFinalTarget)
+        {
+            // We're running the loop ourselves because of the upstream runtime issue (see remarks section of method).
+            // Because of this, we can't incur a nested IOException, but we do need to look out for infinite loops.
```
Shadowghost: I don't think we need these remarks, should be obvious from the mention in the method description and the comments within the loop

crobibero:
I think the current implementation has the potential to fall into an infinite loop still.
I did a little refactor to avoid it

```c#
    /// <summary>
    /// Gets the target of the specified file link.
    /// </summary>
    /// <remarks>
    /// This helper exists because of this upstream runtime issue; https://github.com/dotnet/runtime/issues/92128.
    /// </remarks>
    /// <param name="linkPath">The path of the file link.</param>
    /// <param name="returnFinalTarget">true to follow links to the final target; false to return the immediate next link.</param>
    /// <returns>
    /// A <see cref="FileInfo"/> if the <paramref name="linkPath"/> is a link, regardless of if the target exists; otherwise, <c>null</c>.
    /// </returns>
    public static FileInfo? ResolveLinkTarget(string linkPath, bool returnFinalTarget = false)
    {
        // Check if the file exists so the native resolve handler won't throw at us.
        if (!File.Exists(linkPath))
        {
            return null;
        }

        if (!returnFinalTarget)
        {
            return File.ResolveLinkTarget(linkPath, returnFinalTarget: false) as FileInfo;
        }

        var visited = new HashSet<string>(StringComparer.Ordinal);
        FileInfo? targetInfo = null;
        string currentPath = linkPath;

        while (File.ResolveLinkTarget(currentPath, returnFinalTarget: false) is FileInfo linkInfo)
        {
            string targetPath = linkInfo.FullName;

            if (!visited.Add(targetPath))
            {
                return targetInfo ?? new FileInfo(linkPath);
            }

            targetInfo = linkInfo;
            currentPath = targetPath;

            // Exit if the target doesn't exist, so the native resolve handler won't throw at us
            if (!targetInfo.Exists)
            {
                break;
            }
        }
```

revam:
Will apply it when I get back home. 👍

crobibero (approved):

Shadowghost (approved):

MediaBrowser.Controller/IO/FileSystemHelper.cs
```diff
@@ -61,4 +62,73 @@ public static void DeleteEmptyFolders(IFileSystem fileSystem, string path, ILogg
...
+        }
+
+        if (!returnFinalTarget)
+        {
+            return File.ResolveLinkTarget(linkPath, returnFinalTarget: false) as FileInfo;
+        }
+
+        FileInfo? targetInfo = null;
+        var currentPath = linkPath;
+        var visited = new HashSet<string>(StringComparer.Ordinal) { linkPath };
```
crobibero:
```suggestion
        if (File.ResolveLinkTarget(linkPath, returnFinalTarget: false) is not FileInfo targetInfo)
        {
            return null;
        }

        var currentPath = targetInfo.FullName;
        var visited = new HashSet<string>(StringComparer.Ordinal) { linkPath, currentPath };
```

agrif:
I want to add a note here that, after upgrading, I also had to re-scan my libraries with "replace all metadata" in order to get reliable playback of symlinked files. For some reason, all the music I had previously tried and failed to play on an earlier version *still failed* until I refreshed the library.

asanoplus:
Great job, everyone thank you! It’s finally working, although only partially.
There’s still something not quite right on Windows.

Movies are recognized perfectly.
TV shows, however, are not detected when creating a library with the TV Shows category selected. They are recognized if the Movies & TV Shows option is selected.

I need to be more precise.
When a library is set to TV Shows, it recognizes the TV series, but it does not recognize the seasons or the related episodes.
However, when a library is set to Movies & TV Shows, it recognizes the TV series and correctly detects the seasons and their episodes.
I’m going crazy I really want to fix this issue as soon as possible.

There’s nothing written in the log. I start the scan, but nothing shows up it doesn’t give any output at all.
So frustrating… it’s a fresh installation, and I wanted to start this new setup properly organized with symlinks, but I’ve already been waiting for almost a month…

Please, I’m asking for this issue to be fixed as soon as possible.
Thank you.

Famaku:
@revam From what I can tell this merge is the cause of issues with HDD spindown. Drives get woken up simply by looking at detail pages as the underlying file gets hit. Could you check to confirm?

PR #15302, opened 2025-10-29
Add PDF to CBZ conversion feature

### LINKED ISSUE #15301: Add PDF to CBZ conversion for book (comics) libraries
### Description
Add automatic and manual conversion of PDF books to CBZ format for better compatibility with comic/book readers.

### Motivation
PDF files are far less comfortably supported by Jellyfin than CBZ files (zoom, switch pages...). Converting them to CBZ (Comic Book ZIP) format would improve user experience (instead of modify the way for a PDF to be displayed)

### Proposed Solution
- Automatic conversion after library scan (configurable)
- Manual conversion via API endpoint
- Scheduled task for batch conversion
- Configuration options for conversion behavior

### Additional Context
I have implemented the feature for a small PR, but I'm not very experimented so it should be necessary to test more.

DeMASKe:
## Description
This PR adds automatic and manual PDF to CBZ conversion functionality for book libraries. I'm still working on it because I'm learning on the job, sorry for the inconvenience if my method is not very clear...

Fixes #15301 

## Changes
- **PdfToCbzConverter**: Core conversion logic from PDF to CBZ format
- **PdfToCbzConversionTask**: Scheduled task for batch conversion
- **PdfConversionPostScanTask**: Auto-convert PDFs after library scan
- **BookConversionController**: REST API endpoint for manual conversion
- **BookConversionSettingsDto**: Configuration model for conversion settings
- **BookXmlSaver**: Metadata persistence for converted books

### DIFF
--- a/Emby.Server.Implementations/Library/BookConversion/PdfConversionPostScanTask.cs
+++ b/Emby.Server.Implementations/Library/BookConversion/PdfConversionPostScanTask.cs
@@ -1,0 +1,110 @@
+using System;
+using System.IO;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Jellyfin.Data.Enums;
+using MediaBrowser.Controller.Configuration;
+using MediaBrowser.Controller.Entities;
+using MediaBrowser.Controller.Library;
+using MediaBrowser.Model.Configuration;
+using MediaBrowser.Model.IO;
+using MediaBrowser.Model.Tasks;
+using Microsoft.Extensions.Logging;
+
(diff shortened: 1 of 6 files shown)


crobibero:
I don't think this is something that we will want to add to the server.
It should be fairly easy to make a plugin to run this instead, and then you would be able to make fixes without waiting for the server to update.

DeMASKe:
Thank you for you reply. I know that the "Books" part of Jellyfin is not very active (and I understand that). I'll try to learn more about plug-ins on this app so. ;)

PR #15305, opened 2025-10-29
Fix missing movie filter.

emveepee:
**Changes**

Add the BaseItemKind.LiveTvProgram  to the filters for database types that present movies.  PVR apps will identify  movies on TV based on genre Movies but there type is BaseItemKind.LiveTvProgram .

As as result rows in Movies category from NextPVR and probably other PVR  pass the first condition but not the filter so the result is random TV items from TV are return to PVR Programs unfiltered.

**Issues**


emveepee:
How does one add the reviewers.  This really improves the PVR experience.

theguymadmax:
Fixed by #15594

PR #15325, opened 2025-10-31
Update file size when refreshing metadata

vinnyspb:
**Changes**
Currently, whenever Jellyfin server refreshes metadata of files in a library, it does correctly refresh information like codecs, etc. But if the filename stays the same, it never refreshes file size. A use case may be: replacing a file with another file having the same name but e.g. different codec.
This change fixes this behavior.

**How did I test**
Compiled and installed on a local server verifying the behavior is fixed.


Shadowghost:
This should target the release branch (you need to rebase your changes after doing so)

PR #15328, opened 2025-10-31
fix: in optimistic locking, key off table is locked

evanreichard:
**Changes**
Optimistic locking logic only keys off of "database is locked". Another possible error is "table is locked". This adds a check for "table is locked" so retries are attempted.

**Notes**
I validated this on my local instance. Before I would just see "table is locked" stack traces, now I see (e.g.):

```
[14:04:06] [WRN] [12] Jellyfin.Database.Implementations.Locking.OptimisticLockBehavior: Operation failed retry 1
```


Shadowghost:
Should likely be targeting the release branch

evanreichard:
> Should likely be targeting the release branch

Is this something I should do? This is my first PR here so not sure what the process in general is.

Shadowghost:
Ideally yes, you can change it in the PR at the top, when you click edit. You then need to rebase your commit onto the release branch and force push

crobibero (approved):

Issue #5415, opened 2021-03-08
Collection of potential security issues in Jellyfin


GermanCoding:
# Collection of *potential* security issues in Jellyfin

This is a non exhaustive list of *potential* security issues found in Jellyfin. Some of these might cause controversy. Some of these are design flaws and require breaking changes to correct them. I tried to categorize the issues in general topics. Expansions of this list are welcome. Most of the items on this list are not meant for immediate fixing, but rather for long term planning with a 11.0+ future release in mind, where significant (breaking) changes are allowed. My intention is not to highlight why these issues are here, but what could be done to improve the situation.

## Unauthenticated endpoints
  - ~~#5689 WebSocket connection doesn't require authentication~~
  - #1501 Video streams completely unauthenticated (VideosController, HlsSegmentController, ~~LiveTVController~~)
     - Summary: The current implementation allows direct play without authentication. Transcoded media is partially authenticated (old HLS only has metadata protection, new HLS protects all). The Jellyfin web client *does* provide an api_key parameter in the URL for ~~all~~ some stream types<sup>1</sup>, but it isn't checked.
     - HlsSegmentController has a comment saying "Can't require authentication just yet due to seeing some requests come from Chrome without full query string"
     -  ~~This also applied to streams in LiveTvController endpoints. Those are also not authenticated.~~ Update: Appears to be resolved in 10.8.9
     - Potential fix: Require authentication on all video streams, both direct streams as well as any transcoded media.
  - "Gets subtitles in a specified format" and "Gets subtitles in a specified format" (SubtitleController) endpoints are unauthenticated
     - Allows anyone who is able to retrieve the item id to download subtitles, even if not logged in
     - Potential fix: Authenticate both endpoints
  - All endpoints in AudioController are unauthenticated
    - Has a comment saying "TODO: In order to authenticate this in the future, Dlna playback will require updating"
    - Currently affects "GetAudioStream()" and "GetAudioStreamByContainer()" endpoints. Anyone with knowledge about item ID's can retrieve audio from the server, even if not logged in.
    - Potential fix: Authenticate endpoints
  - ~~All~~ Most endpoints in DashboardController are unauthenticated [Update: Partially authenticated since 10.8b1]
[...]

PR #15339, opened 2025-11-01
Fix: GetUsers endpoint now requires Admin privileges

### LINKED ISSUE #13990: GetUsers() endpoint (UserController) only requires user privileges
split from #5415 

Allows listing of any user, id included, on the server, not limited to public users. Can leverage massively more invasive attacks due to further issues with user id authentication, see below.
Potential fix: Require admin privileges on this endpoint

Cody-Howell-Snow:
**Summary**
GetUsers endpoint was identified as a security vulnerability in #5415, which the breakoff issue #13990 suggests requiring Admin privileges to access. 

**Changes**
GetUsers endpoint now requires admin access. 

UserControllerTests now includes a test to validate the created user cannot access that endpoint. 
AuthHelper had a method split to allow getting the access token of a particular user without changing the signature or function of `CompleteStartupAsync`. The new method is `Task<string> AuthenticateUserByNameAsync(HttpClient client, string username, string password)`.

**Issues**
Fixes https://github.com/jellyfin/jellyfin/issues/13990

### DIFF
--- a/Jellyfin.Api/Controllers/UserController.cs
+++ b/Jellyfin.Api/Controllers/UserController.cs
@@ -89,7 +89,7 @@ public class UserController : BaseJellyfinApiController
     /// <response code="200">Users returned.</response>
     /// <returns>An <see cref="IEnumerable{UserDto}"/> containing the users.</returns>
     [HttpGet]
-    [Authorize]
+    [Authorize(Roles = UserRoles.Administrator)]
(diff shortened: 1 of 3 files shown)


cvium:
I would prefer just returning the authenticated user if they are not admin.

Bond-009:
Closing as stale, no response for 6+ months

PR #12766, opened 2024-10-03
Add watchlist feature

yuvve:
Added [watchlist feature](https://features.jellyfin.org/posts/576/watchlist-like-netflix).

Added IsWatchlisted state with sorting, filtering, persistence and endpoints. IsWatchlisted is automatically set to false when the item is played to completion.

PR #15360, opened 2025-11-03
Add Feature: Watchlist

denj15:
Added [watchlist feature](https://features.jellyfin.org/posts/576/watchlist-like-netflix) to 10.11.1
Used changes made by [@yuvve](https://github.com/yuvve) from #12766 

Added IsWatchlisted state with sorting, filtering, persistence and endpoints. IsWatchlisted is automatically set to false when the item is played to completion.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -281,4 +281,5 @@
  - [Martin Reuter](https://github.com/reuterma24)
  - [Michael McElroy](https://github.com/mcmcelro)
  - [Soumyadip Auddy](https://github.com/SoumyadipAuddy)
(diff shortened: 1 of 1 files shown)


JPVenson:
the code itself looks fine to me but i think it would be better if watchlists would be its own table so the user could add multiple watchlists, like "Action movies to watch". This does also tie very nicely into the planned migration of playlists and collections into the db as they are all more or less the same and could be just a shared table with a type. 

```
BaseItems <-N-M-> ItemListBaseItemMap <-N-M-> ItemList (UserId, ListType[Collection, Playlist, Watchlist], Name)
```

PR #15366, opened 2025-11-03
Extract embedded images from mkv files using -dump_attachment

### LINKED ISSUE #15365: Extract file attachments from mkv files
When extracting embedded images from mkv files, while the extractor correctly identifies the stream, the ffmpeg command the current implementation runs is incorrect if the file is an mkv file. Jellyfin should use the -dump_attachment option to do so

varkokonyi:
**Changes**
This change adds a function to MediaEncoder that dumps file attachments. It also checks in EmbeddedImageProvider if the container is mkv, and if so, uses the new function to get the image. Without this, extracting embedded thumbnails from mkv files simply doesn't work

**Issues**
Fixes #15365


varkokonyi:
The two additional commits:
- Save mkv attachment stream info to the AttachmentStreamInfos table
- Use that table to extract the embedded thumbnail

Now I consider this PR done on my part, let me know if you want me to make any adjustments. I put some effort into mapping the relevant functions and putting the branches in a logical place

|
| Force push because a test failed :)
|
V

Issue #15153, opened 2025-10-23
Cannot login through Jellyseerr after upgrading to 10.11


jyourstone:
### Description of the bug

After upgrading to Jellyfin 10.11.0.0, I cannot login to Jellyseerr using my Jellyfin account, only getting the error "Something went wrong while trying to sign in."

### Reproduction steps

1. Setup Jellyfin 10.11 and Jellyseerr
2. Configure to login with your Jellyfin account
3. Try to login with your Jellyfin account to Jellyseerr

### What is the current _bug_ behavior?

When trying to login in with my Jellyfin account to my Jellyseerr instance, I get an error.

### What is the expected _correct_ behavior?

Logging in with my Jellyfin to Jellyseerr account should work.

### Jellyfin Server version

10.10.0+

### Specify the build version

10.11.0

### Environment

```markdown
- OS: Unraid
- Linux Kernel: Unraid
- Virtualization: Docker
- Clients: Any
- Browser: Any
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Jellyfin Data Storage:
- Media Storage:
- External Integrations:
```

### Jellyfin logs

```shell
[09:38:46] [WRN] [22] Microsoft.AspNetCore.HttpOverrides.ForwardedHeadersMiddleware: Unknown proxy: 192.168.51.25:60090
[09:38:46] [ERR] [22] Jellyfin.Database.Implementations.JellyfinDbContext: Error trying to save changes.
Microsoft.EntityFrameworkCore.DbUpdateConcurrencyException: The database operation was expected to affect 1 row(s), but actually affected 0 row(s); data may have been modified or deleted since entities were loaded. See https://go.microsoft.com/fwlink/?LinkId=527962 for information on understanding and handling optimistic concurrency exceptions.
   at Microsoft.EntityFrameworkCore.Update.AffectedCountModificationCommandBatch.ThrowAggregateUpdateConcurrencyExceptionAsync(RelationalDataReader reader, Int32 commandIndex, Int32 expectedRowsAffected, Int32 rowsAffected, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Update.AffectedCountModificationCommandBatch.ConsumeResultSetWithRowsAffectedOnlyAsync(Int32 commandIndex, RelationalDataReader reader, CancellationToken cancellationToken)
[...]

PR #15368, opened 2025-11-03
Fix UserManager after EFcore refactor

### LINKED ISSUE #15714: User logins are being corrupted intermittently
### Description of the bug

On version 10.11.3 and 10.11.4, I’ve noticed multiple users on my server being unable to logon with their credentials intermittently. This happens across all connected apps as well (e.g. jellyseerr). I was baffled for a while, and determined if I log in as an admin and update any setting on the user (library access, permissions, etc.) the issue resolves itself. I have also verified it’s not a password issue, because I have been able to regain access for an account without modifying the password at all. The bigger issue is i’ve been locked out of my admin account twice and was almost unable to recover. 

I’m running Jellyfin via docker container and restarting the container didn’t help at all.

### Reproduction steps

1. Attempt to log into jellyfin
2. you will see a “Unable to connect to server” error message.
3. log in as an admin and view the user account to verify it exists

Unable to login with correct credentials

Login works.

### Jellyfin Server version

10.11.3

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.4

### Environment

```markdown
- OS:Unraid
- Linux Kernel:
- Virtualization: Docker
- Clients:
- Browser: Safari, Firefox, chrome
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins:
- Reverse Proxy: Nginx Proxy manager
- Base URL:
- Networking: bridge
- Jellyfin Data Storage:
- Media Storage:
- External Integrations: Jellystat, Jellyseerr, Jellysweep, jellyplex-watched
```

### Jellyfin logs

```shell
logs are not loading at the moment… will try to update the ticket once I get them
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

JPVenson:
Changes:
- Removes cache from UserManager
- Fixes RowVersion mismatch lookup
- Adds locking for UserManager operations based on UserId to avoid cross-operational access

fixes https://github.com/jellyfin/jellyfin/issues/15153
fixes https://github.com/jellyfin/jellyfin/issues/16353
fixes #15714


Shadowghost (approved):

nagisa:
> Call for testing is pointless as next to impossible to reproduce reliably as it currently stands.

I get `DbUpdateConcurrencyException`s quite consistently with `supersonic` restarts. I'll apply this patch to my jellyfin sometime later.

nagisa:
Login stops working entirely for me with this patch applied. The login request eventually 504 timeouts.

cvium:
A deadlock occurs. I've mentioned it internally but I forgot to comment on the PR. This patch is not the correct fix either way.

Shadowghost (approved):

Jellyfin.Server.Implementations/Users/UserManager.cs
```diff
@@ -139,42 +160,62 @@ public UserManager(
                 throw new ArgumentException("Invalid username", nameof(name));
             }
 
-            return _users.Values.FirstOrDefault(u => string.Equals(u.Username, name, StringComparison.OrdinalIgnoreCase));
+            using var dbContext = _dbProvider.CreateDbContext();
+#pragma warning disable CA1862 // Use the 'StringComparison' method overloads to perform case-insensitive string comparisons
+#pragma warning disable CA1311 // Specify a culture or use an invariant version to avoid implicit dependency on current culture
+#pragma warning disable CA1304 // The behavior of 'string.ToUpper()' could vary based on the current user's locale settings
+            return UserQuery(dbContext)
+                .FirstOrDefault(u => u.Username.ToUpper() == name.ToUpper());
```
IDisposable, line 168: Wouldn't it be better to use [`.ToUpperInvariant()`](https://learn.microsoft.com/en-us/dotnet/api/system.string.toupperinvariant?view=net-10.0) here?

JPVenson: i need to check if that is available in an EFQuery but i dont think so

Jellyfin.Server.Implementations/Users/UserManager.cs
```diff
@@ -139,42 +160,62 @@ public UserManager(
...
         /// <inheritdoc/>
-        public async Task RenameUser(User user, string newName)
+        public async Task RenameUser(Guid userId, string oldName, string newName)
         {
-            ArgumentNullException.ThrowIfNull(user);
-
             ThrowIfInvalidUsername(newName);
 
-            if (user.Username.Equals(newName, StringComparison.Ordinal))
+            if (oldName.Equals(newName, StringComparison.Ordinal))
```
IDisposable: Since we're using `.ToUpper()` above, shouldn't this be `StringComparison.OrdinalIgnoreCase`?

Jellyfin.Server.Implementations/Users/UserManager.cs
```diff
@@ -139,42 +160,62 @@ public UserManager(
...
+                            newName));
+                    }
 #pragma warning restore CA1304 // The behavior of 'string.ToUpper()' could vary based on the current user's locale settings
 #pragma warning restore CA1311 // Specify a culture or use an invariant version to avoid implicit dependency on current culture
 #pragma warning restore CA1862 // Use the 'StringComparison' method overloads to perform case-insensitive string comparisons
 
-                user.Username = newName;
-                await UpdateUserInternalAsync(dbContext, user).ConfigureAwait(false);
+                    // Load a fresh tracked entity inside the lock so the RowVersion is current.
+                    user = await dbContext.Users
```
IDisposable: Could use `UserQuery(dbContext)` here, except it forces `.AsNoTracking()` might be cleaner to make that an optional flag so we only have the `.Include` train in one place.

Jellyfin.Server.Implementations/Users/UserManager.cs
```diff
@@ -218,23 +258,30 @@ public async Task<User> CreateUserAsync(string name)
...
             await using (dbContext.ConfigureAwait(false))
             {
+#pragma warning disable CA1862 // Use the 'StringComparison' method overloads to perform case-insensitive string comparisons
+#pragma warning disable CA1311 // Specify a culture or use an invariant version to avoid implicit dependency on current culture
+#pragma warning disable CA1304 // The behavior of 'string.ToUpper()' could vary based on the current user's locale settings
+                if (await dbContext.Users
+                        .AnyAsync(u => u.Username.ToUpper() == name.ToUpper())
+                        .ConfigureAwait(false))
+                {
+                    throw new ArgumentException(string.Format(
```
IDisposable, line 268: Since we're not actually transactional... this condition could still occur unless we have a unique constraint on the username column, in which case all this could simply be done by catching the exception (and ensuring/detecting it's a unique-constraint violation)

Jellyfin.Server.Implementations/Users/UserManager.cs
```diff
@@ -245,39 +292,48 @@ public async Task<User> CreateUserAsync(string name)
...
-                    string.Format(
-                        CultureInfo.InvariantCulture,
-                        "The user '{0}' cannot be deleted because there must be at least one admin user in the system.",
-                        user.Username),
-                    nameof(userId));
-            }
+                    var userCount = await dbContext.Users.CountAsync().ConfigureAwait(false);
+                    if (userCount == 1)
+                    {
+                        throw new InvalidOperationException(string.Format(
```
IDisposable, line 310: Technically, we don't need to check this, because the very next condition will fail if there won't be a remaining administrator... which is really what we're trying to ensure, right?

IDisposable (changes requested):
Comments aren't blocking, but should be considered

Jellyfin.Server.Implementations/Users/UserManager.cs
IDisposable: It doesn't, sorry... because the column's collation determines if that matters. Did we create the username with a case-insensitive collation... in which case the database will do the right thing on a simple `.Equals()`... for EFCore, I think best-practice is to use `u.Username.Equals(name, StringComparison.OrdinalIgnoreCase)` because EFCore can recognize it.

Jellyfin.Server.Implementations/Users/UserManager.cs
JPVenson: yes technically correct but that would be using an exception for control flow which is bad design

Jellyfin.Server.Implementations/Users/UserManager.cs
JPVenson: Yes but that code was there before and i dont really want to start touching that. There _might_ be a reason for it i dont see rn. I remove it if you insist.

Jellyfin.Server.Implementations/Users/UserManager.cs
JPVenson: Are you sure?

> In addition, .NET provides overloads of [string.Equals](https://learn.microsoft.com/en-us/dotnet/api/system.string.equals) accepting a [StringComparison](https://learn.microsoft.com/en-us/dotnet/api/system.stringcomparison) enum, which allows specifying case-sensitivity and a culture for the comparison. By design, EF Core refrains from translating these overloads to SQL, and attempting to use them will result in an exception.

https://learn.microsoft.com/en-us/ef/core/miscellaneous/collations-and-case-sensitivity

Jellyfin.Server.Implementations/Users/UserManager.cs
```diff
@@ -403,102 +481,114 @@ public UserDto GetUserDto(User user, string? remoteEndPoint = null)
...
+            bool success;
+            var user = GetUserByName(username);
+            using (await _userLock.LockAsync(user?.Id ?? Guid.Empty).ConfigureAwait(false))
             {
-                string updatedUsername = authResult.Username;
-
-                if (success
-                    && authenticationProvider is not null
-                    && authenticationProvider is not DefaultAuthenticationProvider)
+                // Reload the user now that we hold the lock so the RowVersion is current.
```
cvium, line 475: Doesn't it defeat the purpose of the concurrency check?

JPVenson:
@IDisposable if your comments aren't blocking dont submit a "request changes" because that is blocking xD

Jellyfin.Server.Implementations/Users/UserManager.cs
IDisposable: No, I'm not sure. We used to have the opposite guidance... looks like they came to the same realization as I did... the database's collation on the column(s) and index really should drive this.

> To force a query to use case-sensitive or case-insensitive comparison, specify a collation explicitly via EF.Functions.Collate as [detailed above](https://learn.microsoft.com/en-us/ef/core/miscellaneous/collations-and-case-sensitivity#explicit-collations-and-indexes).

Looking briefly at the EFCore code, it looks like it does support some case-insensitivity with EFCore.Cosmos CONTAINS, but not for Equals.

IDisposable (approved):
Looks good, not sure how much we care about this comment https://github.com/jellyfin/jellyfin/pull/15368/changes/2ac0edc052ed5d22cee3d46fb933bb12a2b36283..586fa01e4685125ef3a9e71f1070a40e61ef7620#diff-cff9cf8fca144299e27f39f7f6625130687e75c435c23aae1fc35126fd175a1fR695

IDisposable:
> @IDisposable if your comments aren't blocking dont submit a "request changes" because that is blocking xD

That was a miss-click... mea culpa.

Jellyfin.Server.Implementations/Users/UserManager.cs
JPVenson: Yea lets not deal with that in this PR. Other parts of the code do that the same and that is a baseline discussion for another time.

Jellyfin.Server.Implementations/Users/UserManager.cs
JPVenson: No not really well at least not in this particular case. 

The issue with re-using the original dto would be here that it has the old auth state but there is no real reason for us to use that. The authentication is valid or not based on the input parameters. 

The concurrency token should never __override__ data with an older state but as we are only want to modify the object inside the lock anyway we requery it.
What we modify here is essentially only the LastActivityDate and InvalidLoginAttemptCount and those are either static values or increments so its even better if we get the most current values here when we enter the lock again so that if the user has been locked out by a previous locking attempt we DO NOT override it with an previous InvalidLoginAttemptCount +1 but always get the most up to date data.

Works as intended

cvium: But the fact it's even needed to refresh it points to an incorrect usage pattern somewhere. You're ignoring any concurrency which defeats the purpose.

It's not the only place where it's done, it's just where my comment landed. I'm talking more general and not this specific case necessarily

JPVenson: i mean i do not?

If you look at the code the base issue is that User does not have a DTO. That is why we cannot "reuse" the DTO as an entity here. Dont think of the object that comes via the parameter as an entity but as what should be the DTO and if it would be an DTO we would need to requery anyway to get tracking.

I dont disagree there is an issue with an incorrect usage but that is because we are not using Dtos here. Also doing it this way is the absolute correct way of doing it even if that would be an DTO instead of the actual entity.

But this elevates to a bigger issue we have with the legacy code overall esp with users. 

Generally without massive 3 layer architecture refactoring this is the easiest solution to fix it and even if we do it, the code inside the lock stays the same.

Jellyfin.Server.Implementations/Users/UserManager.cs
```diff
@@ -139,42 +160,59 @@ public UserManager(
...
-                user.Username = newName;
-                await UpdateUserInternalAsync(dbContext, user).ConfigureAwait(false);
+                    // Load a fresh tracked entity inside the lock so the RowVersion is current.
+                    user = await UserQuery(dbContext)
+                        .AsTracking()
+                        .FirstOrDefaultAsync(u => u.Id == userId)
+                        .ConfigureAwait(false)
+                        ?? throw new ResourceNotFoundException(nameof(userId));
+
+                    user.Username = newName;
```
cvium, line 212: This is the place I was looking for. A fresh entity is loaded but another process could have changed the username before the lock was acquired, but it will be discarded. Same thing happens in changepassword.

JPVenson: No i am sory to dissapoint you, that comment is just wrong. If you look at the code this is the first time we load the entity here.

JPVenson: Removed the comment

Jellyfin.Server.Implementations/Users/UserManager.cs
JPVenson: I rewrote the ChangePassword and ResetPassword methods so this no longer applies either

PR #15369, opened 2025-11-03
feat(sqlite): add timeout config

evanreichard:
**Changes**
Adds "Command-Timeout" tunable ability in `database.xml`. By default, this [is set to 30s](https://github.com/dotnet/efcore/blob/60524c9b11cdadb0d4be96adbe8d0954f9c7ed0a/src/Microsoft.Data.Sqlite.Core/SqliteConnectionStringBuilder.cs#L56) by EFCore. This allows us to override the timeout.

Bumping this up reduces the amount of "database table is locked" occurrences on my instance. Since SQLite can only have a single writer at a time, if you have a lot queued up they can timeout if your storage can't keep up. You _can_ address this with e.g. Optimistic or Pessimistic locking modes, but this just provides a different tunable.

### Usage

```xml
<?xml version="1.0" encoding="utf-8"?>
<DatabaseConfigurationOptions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
  <DatabaseType>Jellyfin-SQLite</DatabaseType>
  <LockingBehavior>Optimistic</LockingBehavior>
  <CustomProviderOptions>
    <Options>
      <CustomDatabaseOption>
        <Key>Command-Timeout</Key>
        <Value>60</Value>
      </CustomDatabaseOption>
    </Options>
  </CustomProviderOptions>
</DatabaseConfigurationOptions>
```

### DIFF
--- a/src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteDatabaseProvider.cs
+++ b/src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteDatabaseProvider.cs
@@ -64,6 +64,7 @@ public sealed class SqliteDatabaseProvider : IJellyfinDatabaseProvider
         sqliteConnectionBuilder.DataSource = Path.Combine(_applicationPaths.DataPath, "jellyfin.db");
         sqliteConnectionBuilder.Cache = GetOption(customOptions, "cache", Enum.Parse<SqliteCacheMode>, () => SqliteCacheMode.Default);
         sqliteConnectionBuilder.Pooling = GetOption(customOptions, "pooling", e => e.Equals(bool.TrueString, StringComparison.OrdinalIgnoreCase), () => true);
+        sqliteConnectionBuilder.DefaultTimeout = GetOption(customOptions, "command-timeout", int.Parse, () => 30);
 
         var connectionString = sqliteConnectionBuilder.ToString();
 


src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteDatabaseProvider.cs
```diff
@@ -64,6 +64,7 @@ public void Initialise(DbContextOptionsBuilder options, DatabaseConfigurationOpt
         sqliteConnectionBuilder.DataSource = Path.Combine(_applicationPaths.DataPath, "jellyfin.db");
         sqliteConnectionBuilder.Cache = GetOption(customOptions, "cache", Enum.Parse<SqliteCacheMode>, () => SqliteCacheMode.Default);
         sqliteConnectionBuilder.Pooling = GetOption(customOptions, "pooling", e => e.Equals(bool.TrueString, StringComparison.OrdinalIgnoreCase), () => true);
+        sqliteConnectionBuilder.DefaultTimeout = GetOption(customOptions, "timeout", int.Parse, () => 30);
```
JPVenson:
```suggestion
        sqliteConnectionBuilder.DefaultTimeout = GetOption(customOptions, "command-timeout", int.Parse, () => 30);
```

JPVenson (changes requested):

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteDatabaseProvider.cs
JPVenson: there are multiple timeout parameters, this one sets the command timeout default

EraYaN: Maybe also reflect the "CommandTimeout" name in the XML

evanreichard: Set to `command-timeout`. Is reflected in the XML as well. (updated description)

PR #15372, opened 2025-11-03
Fix 4 compiler warnings in Emby.Server.Implementations by renaming th…

sarah-strawberries:
**Changes**

We resolved 4 different compiler warnings as described below:
- In `Emby.Server.Implementations/Session/SessionManager.cs`: Changed the name of GetSessions to GetAllSessions to remove warning CA1721; also removed the comment suppressing the warning as it was no longer needed
- In `Emby.Server.Implementations/Configuration/ServerConfigurationManager.cs`: Renamed Configuration to ServerConfig to remove warning CA1721 and reduce ambiguity
- In `Emby.Server.Implementations/Images/ArtistImageProvider.cs: Removed #pragma warning that disables CS1591 directives and added proper XML documentation comments to improve code documentation coverage
- In `Emby.Server.Implementations/AppBase/BaseConfigurationManager.cs‎: Renamed ConfigurationType to CommonConfigurationType to better reflect its purpose and resolved warning CA1721 

**Issues**

Contributes to #2149 - Eliminate All Compiler Warnings In Solution

### DIFF
--- a/Emby.Server.Implementations/AppBase/BaseConfigurationManager.cs
+++ b/Emby.Server.Implementations/AppBase/BaseConfigurationManager.cs
@@ -64,10 +64,10 @@ namespace Emby.Server.Implementations.AppBase
         public event EventHandler<ConfigurationUpdateEventArgs>? NamedConfigurationUpdated;
 
         /// <summary>
-        /// Gets the type of the configuration.
+        /// Gets the type of the common configuration.
         /// </summary>
-        /// <value>The type of the configuration.</value>
-        protected abstract Type ConfigurationType { get; }
+        /// <value>The type of the common configuration.</value>
+        protected abstract Type CommonConfigurationType { get; }
 
         /// <summary>
         /// Gets the logger.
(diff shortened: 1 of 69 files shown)


sjorge:
> * In `Emby.Server.Implementations/Configuration/ServerConfigurationManager.cs`: Renamed Configuration to ServerConfig to remove warning CA1721 and reduce ambiguity

This should probably be ServerConfiguration then to keep things consistant.

sarah-strawberries:
> > * In `Emby.Server.Implementations/Configuration/ServerConfigurationManager.cs`: Renamed Configuration to ServerConfig to remove warning CA1721 and reduce ambiguity
> 
> This should probably be ServerConfiguration then to keep things consistant.

Thanks for the input! That's actually what I had initially thought, too; and then I noticed that the class/data type was also named ServerConfiguration, so I named the instance ServerConfig instead. But would you advise doing otherwise in this case?

Bond-009:
Way to big a change for a single commit/PR
Maybe start with a PR that just fixes some CS1591 warnings

PR #15415, opened 2025-11-05
Fix System.NullReferenceException when people's role is null

IceStormNG:
If "role" is `null` for a person in a movie record (TMDB Metadata), the movie will fail to import properly as it crashes here.
In this case the movie entry will be created and it will also load pictures, but it won't show any streams as the import process stops in the middle, therefore, the movie cannot be played.

**Changes**
Check whether role is null and only trim if it has a value.

**Issues**
I haven't seen someone else report this issue yet. It happened mostly with some German movies. This issue happend since the 10.11 release.

The log showed this:
```
[2025-11-04 11:43:15.293 +01:00] [ERR] [25] MediaBrowser.Providers.Movies.MovieMetadataService: Error in "Probe Provider"
System.NullReferenceException: Object reference not set to an instance of an object.
   at MediaBrowser.Providers.MediaInfo.FFProbeVideoInfo.FetchPeople(Video video, MediaInfo data, MetadataRefreshOptions options)
   at MediaBrowser.Providers.MediaInfo.FFProbeVideoInfo.Fetch(Video video, CancellationToken cancellationToken, MediaInfo mediaInfo, BlurayDiscInfo blurayInfo, MetadataRefreshOptions options)
   at MediaBrowser.Providers.MediaInfo.FFProbeVideoInfo.ProbeVideo[T](T item, MetadataRefreshOptions options, CancellationToken cancellationToken)
   at MediaBrowser.Providers.Manager.MetadataService`2.RunCustomProvider(ICustomMetadataProvider`1 provider, TItemType item, String logName, MetadataRefreshOptions options, RefreshResult refreshResult, CancellationToken cancellationToken)
```

### DIFF
--- a/Emby.Server.Implementations/Localization/Core/pt-PT.json
+++ b/Emby.Server.Implementations/Localization/Core/pt-PT.json
@@ -5,7 +5,7 @@
     "Artists": "Artistas",
     "AuthenticationSucceededWithUserName": "{0} autenticado com sucesso",
     "Books": "Livros",
-    "CameraImageUploadedFrom": "Uma nova imagem de câmara foi enviada a partir de {0}",
+    "CameraImageUploadedFrom": "Uma nova imagem da câmara foi enviada a partir de {0}",
     "Channels": "Canais",
     "ChapterNameValue": "Capítulo {0}",
     "Collections": "Coleções",


IceStormNG:
> Would it not be simpler to just do: `person.Role == person.Role?.Trim()`?

Yeah, you're right. I have to admit that I'm not that used to the special C# features like the ? with its integrated null-check as I'm mostly a C/C++ developer. 

But I can change it if your solution is preferred for readability.

MediaBrowser.Providers/MediaInfo/FFProbeVideoInfo.cs
```diff
@@ -520,7 +520,7 @@ private void FetchPeople(Video video, Model.MediaInfo.MediaInfo data, MetadataRe
                         {
                             Name = person.Name,
                             Type = person.Type,
-                            Role = person.Role.Trim()
+                            Role = person.Role = person.Role?.Trim()
```
KGT1, line 523:
```suggestion
                            Role = person.Role?.Trim() ?? string.Empty
```

KGT1 (commented):
also if this is a bugfix it should target the 10.11. branch

MediaBrowser.Providers/MediaInfo/FFProbeVideoInfo.cs
IceStormNG: This is also fine with me. I tried to restore previous behavior for null values, which was simply without the trim. That would've passed through null values. The remaining server and clients should be able to handle null values as it was like that before, right?

JPVenson:
can you retarget to the 10.11.x branch?

IceStormNG:
> can you retarget to the 10.11.x branch?

Sure. Now targeting 10.11.z branch

JPVenson:
i think you did something wrong while rebasing

IceStormNG:
> i think you did something wrong while rebasing

Yeah... this doesn't look right at all. I might just abandon this PR and make a new one targeting the 10.11.z branch directly.

PR #15433, opened 2025-11-07
Fix episodes showing up on recently added shows

### LINKED ISSUE #15232: Sometimes "Recently Added" directs to a specific episode instead of an entire series
### Description of the bug

Sometimes, "Recently Added" directs to a specific episode instead of an entire series. This makes it somewhat confusing to navigate, and it might be the cause of issue https://github.com/jellyfin/jellyfin-roku/issues/495 for Jellyfin Roku. To clarify, it is not showing me the next episode I'm supposed to watch; it's showing the latest episode, even for series I've never watched before. This may lead to spoilers and confusion when navigating.

I noticed this before 10.11.0.

### Reproduction steps

1. Add new episode

Jellyfin sometimes links/directs to recently added episode instead of whole series.

Direct to whole series.

### Jellyfin Server version

10.10.0+

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.1

### Environment

```markdown
- OS: MacOS Tahoe
- Linux Kernel:
- Virtualization: Native
- Clients: Android TV, Roku, Web
- Browser: Chrome
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Jellyfin Data Storage:
- Media Storage:
- External Integrations:
```

### Jellyfin logs

```shell
n/a
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

See underneath the posters.

Image

### Additional information

_No response_

theguymadmax:
**Changes**
Fix episodes showing up on recently added shows

**Issues**
Fixes #15232

### DIFF
--- a/Jellyfin.Api/Controllers/UserLibraryController.cs
+++ b/Jellyfin.Api/Controllers/UserLibraryController.cs
@@ -13,6 +13,7 @@ using Jellyfin.Extensions;
 using MediaBrowser.Controller.Dto;
 using MediaBrowser.Controller.Entities;
 using MediaBrowser.Controller.Entities.Audio;
+using MediaBrowser.Controller.Entities.TV;
 using MediaBrowser.Controller.Library;
 using MediaBrowser.Controller.Providers;
 using MediaBrowser.Model.Dto;
@@ -569,7 +570,7 @@ public class UserLibraryController : BaseJellyfinApiController
             var item = i.Item2[0];
             var childCount = 0;
 
-            if (i.Item1 is not null && (i.Item2.Count > 1 || i.Item1 is MusicAlbum))
+            if (i.Item1 is not null && (i.Item2.Count > 1 || i.Item1 is MusicAlbum || i.Item1 is Series ))
             {
                 item = i.Item1;
                 childCount = i.Item2.Count;


JPVenson (approved):

Bond-009 (approved):

soultaco83:
I can confirm that with this PR installed that the Recently added TV and anime now have series instead of episodes.

image

nothing2obvi:
Did I read somewhere that we will not see this until 10.12.x or am I making that up? Thanks.

nickhomelab:
> Did I read somewhere that we will not see this until 10.12.x or am I making that up? Thanks.

You're probably thinking of this comment from a related issue: https://github.com/jellyfin/jellyfin/issues/15749#issuecomment-3628310850

nickhomelab:
After updating to 12.0, I'm still seeing the behavior where the last show in the "Recently Added in Shows" list links to the final episode of the season instead of the show itself. I removed the problematic show's directory, scanned libraries so it was fully removed from Jellyfin, added the show's directory back, then scanned libraries again. Same behavior.

Anyone else?

theguymadmax:
The changes in this PR were reverted in #16967, so episodes will show up in 12.0.

nickhomelab:
I guess I'm confused about the intended behavior then. Why would we want to link to the final episode in a season when earlier episodes in that season are unwatched? Is the idea that the "Recently Added in Shows" always links to the most recently added episode? I guess I can see the logic due to the latest episode being the most "Recently Added," but that section becomes effectively useless to users that aren't watching new episodes every week.

pafade89:
I actually see it the same way, and I’d also be surprised if the view in 12.0 were the same as it was before.

PR #15442, opened 2025-11-08
Allow resolvers to set creation/modified dates

revam:
**Changes**
Allow resolvers to set the creation and modified dates for items. Before https://github.com/jellyfin/jellyfin/pull/15209 I could use the symbolic links to alter the creation dates of items at a file system level — because it was reading the created/modified at dates of the symbolic link and not the source file — but after this PR landed it was no longer possible. This change will allow the resolver(s) to set the created/modified dates of items, which will allow my plugin to resolve the items with the dates even with the symbolic link fix applied.

I've tested that the `DateCreated` and `DateModified` are both set to `DateTime.MinValue` upon initialization if not altered/set by the resolver.

**Issues**
Related to https://github.com/jellyfin/jellyfin/pull/15209

### DIFF
--- a/Emby.Server.Implementations/Library/ResolverHelper.cs
+++ b/Emby.Server.Implementations/Library/ResolverHelper.cs
@@ -132,25 +132,28 @@ namespace Emby.Server.Implementations.Library
 
         private static void SetDateCreated(BaseItem item, FileSystemMetadata? info)
         {
-            var config = BaseItem.ConfigurationManager.GetMetadataConfiguration();
-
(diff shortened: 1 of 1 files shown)


Bond-009:
Please target master for features

revam:
So two things;

- this PR was flawed in the fact that it allowed setting the modified date of items, but Jellyfin core has checks for the file's modified date which bypasses the code where it may be set by resolvers. Which in turn may lead to infinite ffprobes on the items upon refresh. The creation date is not affected by this it seems.
- I've lost interest in upstreaming this and will be closing this now.

PR #15455, opened 2025-11-09
Add FTS5 full-text search support for BaseItems

TOomaAh:
**Changes**
Implements SQLite FTS5 virtual table to improve search performance on large media libraries. Uses pure LINQ queries via BaseItemFtsEntity with the Match column approach. Includes migration with automatic sync triggers for INSERT/UPDATE/DELETE operations. Configures Porter stemming tokenizer with prefix matching for more flexible search results.

Current implementation:

image

With FTS5:

image


Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -1701,16 +1693,7 @@ private IQueryable<BaseItemEntity> TranslateQuery(
...
-            if (SearchWildcardTerms.Any(f => searchTerm.Contains(f)))
-            {
-                searchTerm = $"%{searchTerm.Trim('%')}%";
-                baseQuery = baseQuery.Where(e => EF.Functions.Like(e.CleanName!.ToLower(), searchTerm) || (e.OriginalTitle != null && EF.Functions.Like(e.OriginalTitle.ToLower(), searchTerm)));
-            }
-            else
-            {
-                baseQuery = baseQuery.Where(e => e.CleanName!.ToLower().Contains(searchTerm) || (e.OriginalTitle != null && e.OriginalTitle.ToLower().Contains(searchTerm)));
-            }
+            baseQuery = baseQuery.SearchByTextCaseInsensitive(context, filter.SearchTerm);
```
JPVenson: this essentially throws us back to a state where sqlite dependency is hardcoded requirement. We dont want to do that to support other dbms in the future like pgsql.

It would be better if you open a ticket on our meta repo first laying out how this could be implemented adaptable and futerproof to support multiple dbms in the future without the code being aware of the implementation detail.

JPVenson (changes requested):

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteFtsProvider.cs
```diff
@@ -0,0 +1,88 @@
...
+            ftsQuery = ownedContext.BaseItemFts;
+        }
+        else
+        {
+            return null;
+        }
+
+        try
+        {
+            var ftsIds = ftsQuery
```
JPVenson: do we really need to materialise the query here?

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteJellyfinDbContext.cs
```diff
@@ -0,0 +1,37 @@
...
+namespace Jellyfin.Database.Providers.Sqlite;
+
+/// <summary>
+/// SQLite-specific DbContext that extends JellyfinDbContext with FTS5 entities.
+/// </summary>
+/// <param name="options">The database context options.</param>
+/// <param name="logger">Logger.</param>
+/// <param name="jellyfinDatabaseProvider">The provider for the database engine specific operations.</param>
+/// <param name="entityFrameworkCoreLocking">The locking behavior.</param>
+public class SqliteJellyfinDbContext(
```
JPVenson, line 17: would it make more sense if this extends the JellyfinDbContext and replaces it instead? That way we also have migrations and everything

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteFtsProvider.cs
TOomaAh: You're right.

TOomaAh: It should be fine
I still have some tests left to do.

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteJellyfinDbContext.cs
TOomaAh: I'm not sure if I'm taking the right approach. I had to change the type for each SQLite migration.

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteFtsProvider.cs
TOomaAh: That looks good. I tried again with a JOIN, but the performance was mixe, probably because of the "ORDER BY rank". I switched back to an IN clause without materializing the IDs.

PR #15466, opened 2025-11-10
Don't error out when searching for marker files fails

### LINKED ISSUE #15445: Transcoding fails if /var/cache/jellyfin/transcodes/ is a separate ext[234] filesystem
### Description of the bug

If you mount a separate filesystem formatted with ext[234] on `/var/cache/jellyfin/transcodes/` transcoding fails with:

```
Nov 09 02:17:24 jellyfin[1066879]: System.UnauthorizedAccessException: Access to the path '/var/cache/jellyfin/transcodes/lost+found' is denied.                                    
Nov 09 02:17:24 jellyfin[1066879]:  ---> System.IO.IOException: Permission denied                                                                                                   
Nov 09 02:17:24 jellyfin[1066879]:    --- End of inner exception stack trace ---                                                                                                    
```

I could supply the rest of stack trace if necessary but I suspect it's not really needed.

### Reproduction steps

1. Create a block device (partition, LVM LV, etc.)
2. Format it with an ext4 filesystem
3. Mount that filesystem on /var/cache/jellyfin/transcodes/
4. Sleep better at night knowing that transcoding is not going to fill up your disk and break other things when they start hitting a full disk

When you try to play something that needs transcoding you hit the error in the description.

Transcoding should not break simply because the `lost+found` directory that ext[234] wants for filesystem recovery exists.

### Jellyfin Server version

10.11.0+

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

Release

### Environment

```markdown
- OS: EL9
- Linux Kernel: 5.14.0-570.49.1.el9_6.x86_64
- Virtualization: No
- Clients: Web
- Browser: Chrome
- FFmpeg Version: jellyfin-ffmpeg-7.1.2-3
- Playback Method: transcode
- Hardware Acceleration: VAAPI
- GPU Model: Intel
- Plugins: N/A
- Reverse Proxy: No
- Base URL: none
- Networking: Native
- Media Storage: local HDD
- External Integrations: N/A
```

### Jellyfin logs

```shell
See description.
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

Bond-009:
Fixes #15445

### DIFF
--- a/Emby.Server.Implementations/AppBase/BaseApplicationPaths.cs
+++ b/Emby.Server.Implementations/AppBase/BaseApplicationPaths.cs
@@ -107,10 +107,20 @@ namespace Emby.Server.Implementations.AppBase
 
         private void CheckOrCreateMarker(string path, string markerName, bool recursive = false)
         {
-            var otherMarkers = GetMarkers(path, recursive).FirstOrDefault(e => Path.GetFileName(e) != markerName);
+            string? otherMarkers = null;
+            try
+            {
+                otherMarkers = GetMarkers(path, recursive).FirstOrDefault(e => !Path.GetFileName(e.AsSpan()).Equals(markerName, StringComparison.OrdinalIgnoreCase));
+            }
+            catch
+            {
+                // Error while checking for marker files, assume none exist and keep going
+                // TODO: add some logging
+            }
+
             if (otherMarkers is not null)
             {
-                throw new InvalidOperationException($"Exepected to find only {markerName} but found marker for {otherMarkers}.");
+                throw new InvalidOperationException($"Expected to find only {markerName} but found marker for {otherMarkers}.");
             }
 
             var markerPath = Path.Combine(path, markerName);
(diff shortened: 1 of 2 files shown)


JPVenson:
in what case could the operation error out that allows us to continue here? Its an enumeration, if there are none, the operation does not fail but its a noop.

Bond-009:
> in what case could the operation error out that allows us to continue here? Its an enumeration, if there are none, the operation does not fail but its a noop.

The search is recursive, if there are subfolders we can't access it throws an exception. We can continue since the marker files are just a nice to have to prevent PEBKAC errors. We still create the marker marker file, thus testing if we have write access.

JPVenson:
IIRC only the transcode and cache folder are search recursive and thats by design to fail if there are some folders that cannot be accessed because for the cleanup routines would crash otherwise. 

All other marker searches are toplevel only and if that fails too, we cannot access those folders and this is a nice early fail

Bond-009:
> IIRC only the transcode and cache folder are search recursive and thats by design to fail if there are some folders that cannot be accessed because for the cleanup routines would crash otherwise.

That's stupid, this prevents people from using some filesystems roots as transcode dirs. The cleanup tasks are non-essential, user configurable and can be made more robust. This isn't a valid reason to prevent people from using these directories.

> All other marker searches are toplevel only and if that fails too, we cannot access those folders and this is a nice early fail

We still fail only a few lines later when trying to create the marker file.

PR #15490, opened 2025-11-12
Avoid concurrent dictionary snapshot

grzonu:
**Changes**
When you using ConcurrentDictionary .Values there is lock, memory snapshot then unlock which has negative performance impact and negative memory impact.
Better to use GetEnumerator which is thread-safe but not make dictionary snapshot

### DIFF
--- a/Emby.Server.Implementations/Library/MediaSourceManager.cs
+++ b/Emby.Server.Implementations/Library/MediaSourceManager.cs
@@ -491,15 +491,6 @@ namespace Emby.Server.Implementations.Library
             .Where(i => i.Type != MediaSourceType.Placeholder);
         }
 
-        private IEnumerable<ILiveStream> GetCurrentLiveStreams()
-        {
-            using var streams = _openStreams.GetEnumerator();
-            while (streams.MoveNext())
-            {
-                yield return streams.Current.Value;
-            }
-        }
-
         public async Task<Tuple<LiveStreamResponse, IDirectStreamProvider>> OpenLiveStreamInternal(LiveStreamRequest request, CancellationToken cancellationToken)
         {
             MediaSourceInfo mediaSource;
@@ -509,7 +500,7 @@ namespace Emby.Server.Implementations.Library
             {
                 var (provider, keyId) = GetProvider(request.OpenToken);
 
-                var currentLiveStreams = GetCurrentLiveStreams().ToList();
+                var currentLiveStreams = _openStreams.Select(x => x.Value).ToList();
 
                 liveStream = await provider.OpenMediaSource(keyId, currentLiveStreams, cancellationToken).ConfigureAwait(false);
 
--- a/Emby.Server.Implementations/Library/MediaSourceManager.cs
+++ b/Emby.Server.Implementations/Library/MediaSourceManager.cs
@@ -491,6 +491,15 @@ namespace Emby.Server.Implementations.Library
             .Where(i => i.Type != MediaSourceType.Placeholder);
         }
 
+        private IEnumerable<ILiveStream> GetCurrentLiveStreams()
+        {
+            using var streams = _openStreams.GetEnumerator();
+            while (streams.MoveNext())
+            {
+                yield return streams.Current.Value;
+            }
+        }
+
         public async Task<Tuple<LiveStreamResponse, IDirectStreamProvider>> OpenLiveStreamInternal(LiveStreamRequest request, CancellationToken cancellationToken)
         {
             MediaSourceInfo mediaSource;
@@ -500,7 +509,7 @@ namespace Emby.Server.Implementations.Library
             {
                 var (provider, keyId) = GetProvider(request.OpenToken);
 
-                var currentLiveStreams = _openStreams.Values.ToList();
+                var currentLiveStreams = GetCurrentLiveStreams().ToList();
 
                 liveStream = await provider.OpenMediaSource(keyId, currentLiveStreams, cancellationToken).ConfigureAwait(false);
 
--- a/Emby.Server.Implementations/Session/SessionManager.cs
+++ b/Emby.Server.Implementations/Session/SessionManager.cs
@@ -155,7 +155,19 @@ namespace Emby.Server.Implementations.Session
         /// Gets all connections.
         /// </summary>
         /// <value>All connections.</value>
-        public IEnumerable<SessionInfo> Sessions => _activeConnections.Values.OrderByDescending(c => c.LastActivityDate);
+        public IEnumerable<SessionInfo> Sessions
+        {
+            get
+            {
+                using var enumerator = _activeConnections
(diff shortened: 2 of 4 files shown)


Brokolis:
Could just use `.Select(x = x.Value)` instead of writing a custom enumerator, less code, but same amount of allocations. Also, I wonder if the performance improvement here outweighs the loss of code readability.

Jellyfin.Server.Implementations/Devices/DeviceManager.cs
```diff
@@ -135,7 +135,19 @@ public ClientCapabilities GetCapabilities(string? deviceId)
         /// <inheritdoc />
         public DeviceInfoDto? GetDevice(string id)
         {
-            var device = _devices.Values.Where(d => d.DeviceId == id).OrderByDescending(d => d.DateLastActivity).FirstOrDefault();
+            using var enumerator = _devices
+                .OrderByDescending(d => d.Value.DateLastActivity)
```
Brokolis: If we're trying to improve performance, we should do the filtering before ordering.

grzonu: rebuilded that to assign only when LastActivity is newer - can be done in one loop

grzonu:
> Could just use `.Select(x = x.Value)` instead of writing a custom enumerator, less code, but same amount of allocations. Also, I wonder if the performance improvement here outweighs the loss of code readability.

I made benchmark to prove it's not pointless: https://github.com/grzonu/concurrency-dict-benchmark/tree/main/ConcurrentDictionaryBench


| Method                            | Mean     | Error   | StdDev  | Gen0   | Gen1   | Allocated |
|---------------------------------- |---------:|--------:|--------:|-------:|-------:|----------:|
| IterateViaValues                  | 113.0 us | 1.11 us | 0.99 us | 2.4414 |      - |   40136 B |
| IterateWithWhereViaValues         | 151.5 us | 0.77 us | 0.60 us |      - |      - |     176 B |
| IterateViaEnumerator              | 103.8 us | 1.09 us | 0.97 us |      - |      - |         - |
| IterateParallelViaValues          | 246.2 us | 1.77 us | 1.57 us | 5.3711 | 0.4883 |   81119 B |
| IterateParallelWithWhereViaValues | 194.7 us | 0.93 us | 0.78 us |      - |      - |    1199 B |
| IterateParallelViaEnumerator      | 137.9 us | 0.65 us | 0.61 us |      - |      - |     845 B |


There is not much time difference when we using values / enumerator single thread but it's huge difference in memory allocation
but .Where() is the slowest but memory efficient too

But look on Parallel run - same cases with 3 threads and .Values is 2 times slower when Enumerator is scaling much better. And still memory efficient - almost made "in-place"

Brokolis:
What I meant with allocations, was using the `yield` keyword, it generates a class enumerable, so compared to just a `foreach` or straight `.Select` should have the same amount of allocations, but should be more readable.

Your benchmarks do show that memory allocations can be significantly reduced though, by not using `.Values`. I wonder if `.Keys` is the same?

Jellyfin.Server.Implementations/Devices/DeviceManager.cs
```diff
@@ -135,7 +135,20 @@ public ClientCapabilities GetCapabilities(string? deviceId)
         /// <inheritdoc />
         public DeviceInfoDto? GetDevice(string id)
         {
-            var device = _devices.Values.Where(d => d.DeviceId == id).OrderByDescending(d => d.DateLastActivity).FirstOrDefault();
+            using var enumerator = _devices
+                .GetEnumerator();
+            Device? device = null;
+            DateTime? deviceLastActivity = null;
```
Brokolis: Having both `device` and `deviceLastActivity` seems a bit redundant, could just use `device.DateLastActivity` directly.

Jellyfin.Server.Implementations/Devices/DeviceManager.cs
```diff
@@ -135,7 +135,20 @@ public ClientCapabilities GetCapabilities(string? deviceId)
...
+            while (enumerator.MoveNext())
```
Brokolis: A regular foreach should be fine here.

Jellyfin.Server.Implementations/Devices/DeviceManager.cs
```diff
@@ -135,7 +135,20 @@ public ClientCapabilities GetCapabilities(string? deviceId)
...
+            {
+                if (enumerator.Current.Value.DeviceId == id && (deviceLastActivity == null || deviceLastActivity < enumerator.Current.Value.DateLastActivity))
+                {
+                    device = enumerator.Current.Value;
+                    deviceLastActivity = enumerator.Current.Value.DateLastActivity;
+                    break;
```
Brokolis: Should remove this, since we don't want to break out of the loop on the first match.

Emby.Server.Implementations/Session/SessionManager.cs
```diff
@@ -155,7 +155,19 @@ public SessionManager(
         /// Gets all connections.
         /// </summary>
         /// <value>All connections.</value>
-        public IEnumerable<SessionInfo> Sessions => _activeConnections.Values.OrderByDescending(c => c.LastActivityDate);
+        public IEnumerable<SessionInfo> Sessions
+        {
+            get
```
Brokolis, line 160: I would think `return _activeConnections.OrderByDescending(x => x.Value.LastActivityDate).Select(x => x.Value)` would have the same performance and allocations.

Emby.Server.Implementations/Library/MediaSourceManager.cs
```diff
@@ -500,7 +509,7 @@ public async Task<Tuple<LiveStreamResponse, IDirectStreamProvider>> OpenLiveStre
             {
                 var (provider, keyId) = GetProvider(request.OpenToken);
 
-                var currentLiveStreams = _openStreams.Values.ToList();
+                var currentLiveStreams = GetCurrentLiveStreams().ToList();
```
Brokolis: I think `_openStreams.Select(x => x.Value).ToList()` would have the same performance and allocations.

grzonu:
I changed to more readability code to simmilar performance solutions

Emby.Server.Implementations/Session/SessionManager.cs
```diff
@@ -155,7 +155,16 @@ public SessionManager(
...
+            {
+                return _activeConnections
+                    .OrderByDescending(c => c.Value.LastActivityDate)
+                    .Select(x => x.Value)
+                    .ToList();
```
cvium, line 165: As far as I know, ToList makes this operation not thread-safe. The list size is read from the Dictionary's Count property and the contents are then copied over, but the number of items can change, which can lead to out of bounds errors.

PR #15514, opened 2025-11-15
Add 1 minute tolerance for NFO change detection

### LINKED ISSUE #14841: DateCreated of existing items gets updated if feature UseFileCreationTimeForDateAdded is activated
### Description of the bug

When the feature UseFileCreationTimeForDateAdded is activated, the value of DateCreated of already existing items gets updated to the current system time on library rescan for some item types (seasons, albums, artists – seems related to "folder" types).

The update happens in line https://github.com/jellyfin/jellyfin/blob/a2c07994893e8bf7fd9d41b46872a42911830901/MediaBrowser.Providers/Manager/MetadataService.cs#L342

Maybe only on specific file systems (I'm using Linux) as the update uses a value from FileSystem.GetFileSystemInfo.

### Reproduction steps

1. Create a new library for shows.
2. Enable UseFileCreationTimeForDateAdded and nfo saver for this library.
3. Add a show to the library and scan for new and updated items.
4. Check "dateadded" in the nfo file season.nfo.
5. Rescan the new library (search for new and updated items).
6. Check "dateadded" in the nfo file season.nfo again.
7. The value shouldn't have changed but it will have been updated to the time of the second scan.

A library rescan causes an update of the value of DateAdded of some items even if they already existed in Jellyfin before the rescan. These new values are stored in the databass and in nfo files (if nfo saver is enabled).

The value of DateCreated should only be set on initial item creation and not get updated on library rescans.

### Jellyfin Server version

Master

### Specify commit id

a2c07994893e8bf7fd9d41b46872a42911830901

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.0-rc5

### Environment

```markdown
- OS: Synology DSM 7.2.2-72806 Update 4
- Virtualization: Docker Container (Synology Container Manager)
- Clients: Windows 11 Pro Desktop and Android Tablet
- Browser: Firefox
- Storage Access: Docker Bind Mounts
- Networking: local LAN/WLAN (client to server)
- Jellyfin Data Storage: Synology NAS (SSD volume)
- Media Storage: same Synology NAS (HDD volume)
```

### Jellyfin logs

```shell
no log available
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

theguymadmax:
The NFO saver was writing the files 1-2 seconds after the `DateLastSaved` time.  On the next library scan, `HasChanged` detected the NFO file as newer and triggered a metadata refresh, which caused the file to be rewritten again.

**Changes**
Added a 1-minute tolerance to `BaseNfoProvider.HasChanged()`

**Issues**
 Fixes #14841

### DIFF
--- a/MediaBrowser.XbmcMetadata/Providers/BaseNfoProvider.cs
+++ b/MediaBrowser.XbmcMetadata/Providers/BaseNfoProvider.cs
@@ -75,8 +75,8 @@ namespace MediaBrowser.XbmcMetadata.Providers
 
             var fileTime = _fileSystem.GetLastWriteTimeUtc(file);
 
-            // 5s tolerance to avoid detecting our own file writes
-            return (fileTime - item.DateLastSaved).TotalSeconds > 5;
+            // 1 minute tolerance to avoid detecting our own file writes
+            return (fileTime - item.DateLastSaved) > TimeSpan.FromMinutes(1);
         }
 
         protected abstract void Fetch(MetadataResult<T> result, string path, CancellationToken cancellationToken);


crobibero:
I think we can safely increase this tolerance to 1 minute

The only situation I can think of where we would hit this is if someone manually updates the nfo. At that point they can trigger a scan if we didn't pick up the change

crobibero (approved):

Issue #15361, opened 2025-11-03
Cannot edit media metadata


wolf7689:
### Description of the bug

After attempting to edit the metadata for an unknown season, I can no longer edit the metadata for any items. you can see the logs.

### Reproduction steps

1. edit the metadata for an unknown season
2. Change inheritance to episodes
3. save it

### What is the current _bug_ behavior?

When opening the metadata editing interface, no content loads and it keeps spinning indefinitely.

### What is the expected _correct_ behavior?

metadata can be edit

### Jellyfin Server version

10.10.0+

### Specify the build version

10.11.0

### Environment

```markdown
- OS:docker jellyfin/jellyfin
- Linux Kernel:
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Jellyfin Data Storage:
- Media Storage:
- External Integrations:
```

### Jellyfin logs

```shell
[2025-11-03 14:17:49.594 +08:00] [ERR] [16] Jellyfin.Api.Middleware.ExceptionMiddleware: Error processing request. URL "GET" "/Items/0e0d9129afd1cab57ea9fba085dbbcf2/MetadataEditor".
System.NullReferenceException: Object reference not set to an instance of an object.
   at Emby.Server.Implementations.IO.ManagedFileSystem.AreEqual(String path1, String path2)
   at Emby.Server.Implementations.Library.LibraryManager.<>c__DisplayClass168_0.<GetContentTypeOverride>b__0(NameValuePair i)
   at System.Linq.Enumerable.TryGetFirst[TSource](IEnumerable`1 source, Func`2 predicate, Boolean& found)
   at Jellyfin.Api.Controllers.ItemUpdateController.GetMetadataEditorInfo(Guid itemId)
   at lambda_method101473(Closure, Object, Object[])
   at Microsoft.AspNetCore.Mvc.Infrastructure.ActionMethodExecutor.SyncObjectResultExecutor.Execute(ActionContext actionContext, IActionResultTypeMapper mapper, ObjectMethodExecutor executor, Object controller, Object[] arguments)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.InvokeActionMethodAsync()
[...]

PR #15548, opened 2025-11-18
Fix NullReferenceException in filesystem path comparison

theguymadmax:
**Changes**
Add null check to the `AreEqual` method

**Issues**
#15361

### DIFF
--- a/Emby.Server.Implementations/IO/ManagedFileSystem.cs
+++ b/Emby.Server.Implementations/IO/ManagedFileSystem.cs
@@ -497,8 +497,17 @@ namespace Emby.Server.Implementations.IO
         /// <inheritdoc />
         public virtual bool AreEqual(string path1, string path2)
         {
-            return Path.TrimEndingDirectorySeparator(path1).Equals(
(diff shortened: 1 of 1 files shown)


Shadowghost (approved):

Emby.Server.Implementations/IO/ManagedFileSystem.cs
```diff
@@ -497,8 +497,17 @@ public virtual bool ContainsSubPath(string parentPath, string path)
...
         {
-            return Path.TrimEndingDirectorySeparator(path1).Equals(
-                Path.TrimEndingDirectorySeparator(path2),
+            if (string.IsNullOrWhiteSpace(path1) || string.IsNullOrWhiteSpace(path2))
+            {
+                return false;
+            }
+
+            var normalized1 = Path.TrimEndingDirectorySeparator(path1);
+            var normalized2 = Path.TrimEndingDirectorySeparator(path2);
```
Bond-009, line 506: Using the Span overload would have been better

PR #15559, opened 2025-11-18
Disable legacy authorization methods by default

nielsvanvelzen:
This is part 2 of 3 for the removal of legacy authorization method. Previously we added an configuration options that is disabled by default in #13306 which released with 10.11. This PR will change the EnableLegacyAuthorization option to false for all installs, allowing a user to (temporarily) keep using the legacy method if they use an client that is not updated yet.

Getting this into 10.12 unstable builds early so clients that haven't migrated yet will know in time.

We'll remove this configuration option (and the authorization methods) in a future release, likely 10.13.

**Changes**
- Disable legacy authorization methods by default

**Issues**

### DIFF
--- a/Jellyfin.Server/Migrations/Routines/DisableLegacyAuthorization.cs
+++ b/Jellyfin.Server/Migrations/Routines/DisableLegacyAuthorization.cs
@@ -1,3 +1,5 @@
+using System.Threading;
+using System.Threading.Tasks;
 using MediaBrowser.Controller.Configuration;
 
 namespace Jellyfin.Server.Migrations.Routines;
@@ -5,10 +7,8 @@ namespace Jellyfin.Server.Migrations.Routines;
 /// <summary>
 /// Migration to disable legacy authorization in the system config.
 /// </summary>
-#pragma warning disable CS0618 // Type or member is obsolete
-[JellyfinMigration("2025-11-18T16:00:00", nameof(DisableLegacyAuthorization), "F020F843-E079-4061-99E0-F43D145F2557")]
-public class DisableLegacyAuthorization : IMigrationRoutine
-#pragma warning restore CS0618 // Type or member is obsolete
+[JellyfinMigration("2025-11-18T16:00:00", nameof(DisableLegacyAuthorization))]
+public class DisableLegacyAuthorization : IAsyncMigrationRoutine
 {
     private readonly IServerConfigurationManager _serverConfigurationManager;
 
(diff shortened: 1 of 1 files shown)


crobibero (approved):
Should probably use the non-obsolete migrator for new migrations

nielsvanvelzen:
I copied an existing migration as base and build on top of that as I have never worked with our migrations before. Can you link to a migration with the new system I can use as a basis instead?

crobibero:
https://github.com/jellyfin/jellyfin/blob/master/Jellyfin.Server/Migrations/Routines/CleanMusicArtist.cs

tldr; implement `IAsyncMigrationRoutine`

nielsvanvelzen:
Ok that was rather easy

ghost:
This breaks Infuse, Manet, Fintunes.  I let Firecore customer support and André@Manet know, and created an issue for Fintunes.

nielsvanvelzen:
It is expected that some clients will break as they're not using our best practices. The deprecation has been communicated for a few years now and we've also explicitly mentioned it in the 10.11 release. Fixing it is easy and the recommended method is also compatible with older (unsupported) Jellyfin versions though.

oddstr13:
This PR breaks a rather large part of the Jellyfin ecosystem, including many official clients.

When I've asked in chat for the reasoning behind why this is done, I'm only referenced to this gist, which does NOT sufficiently warn against using the api_key, and the 10.11 release notes which simply point to the gist.
https://gist.github.com/nielsvanvelzen/ea047d9028f676185832e51ffaf12a6f

The errors in the log are also insufficient as evident in the seerr-team/seerr#2249 issue report.

The `api_key` parameter is still used in many of our own clients and plugins, including all the Python based clients, and by extension, any client using the api-client we publish.

Non-exhaustive list of clients still using `api_key`:
- Jellyfin for Kodi
- Mopidy
- Jellycon
- The Python apiclient and anything using it
- Chromecast
- MPV shim
- Vue
- Roku
- iOS
- Several of our plugins, but hard to say which as the parameter name is also used for external APIs

The "new" URL parameter name was added in #7080 which was first released with v10.8.0.
This will be a hard-break in API backwards compatibility.

I was made aware of the new key back in 2023 when the gist was published, but there was made no indication that it was going to get removed, and as such the decision was to keep api_key which is more pythonic and more in line with what is used across the web.

The Jellyfin documentation only mentions api_key in relation to censoring logs, and even then api_key is mentioned more often than ApiKey. A personal gist does not count as project documentation, even if posted in chat and otherwise several times.

---

1. What is the benefit of this change?
2. Why is this change being made now?
3. Logging must be improved.
4. This change will be release-blocking until clients are updated and released.
5. This change will break plugins in a non-obvious way (no compile errors).
6. This change will break countless external clients and tools.

PR #15583, opened 2025-11-20
Add ability to sort and filter activity log entries

crobibero:
Part of https://github.com/jellyfin/jellyfin-web/issues/6221

### DIFF
--- a/Jellyfin.Api/Controllers/ActivityLogController.cs
+++ b/Jellyfin.Api/Controllers/ActivityLogController.cs
@@ -44,7 +44,7 @@ public class ActivityLogController : BaseJellyfinApiController
     /// <param name="shortOverview">Filter by short overview.</param>
     /// <param name="type">Filter by type.</param>
     /// <param name="itemId">Filter by item id.</param>
-    /// <param name="userId">Filter by user id.</param>
+    /// <param name="username">Filter by username.</param>
     /// <param name="severity">Filter by log severity.</param>
     /// <param name="sortBy">Specify one or more sort orders. Format: SortBy=Name,Type.</param>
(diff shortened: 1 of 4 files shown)


Jellyfin.Api/Controllers/ActivityLogController.cs
```diff
@@ -44,14 +56,60 @@ public async Task<ActionResult<QueryResult<ActivityLogEntry>>> GetLogEntries(
         [FromQuery] int? startIndex,
         [FromQuery] int? limit,
         [FromQuery] DateTime? minDate,
-        [FromQuery] bool? hasUserId)
+        [FromQuery] bool? hasUserId,
+        [FromQuery] string? name,
+        [FromQuery] string? overview,
+        [FromQuery] string? shortOverview,
+        [FromQuery] string? type,
+        [FromQuery] string? itemId,
```
nielsvanvelzen: Shouldn't this be a Guid?

nielsvanvelzen (changes requested):

Jellyfin.Data/Enums/ActivityLogSortBy.cs
```diff
@@ -0,0 +1,47 @@
...
+
+    /// <summary>
+    /// Sort by date.
+    /// </summary>
+    DateCreated = 5,
+
+    /// <summary>
+    /// Sort by user id.
+    /// </summary>
+    UserId = 6,
```
nielsvanvelzen: I don't think this one makes that much sense, shouldn't it be user names instead?

Bond-009 (approved):

Shadowghost (approved):

PR #15593, opened 2025-11-21
include external IDs and URLs for book providers

dkanada:
**Changes**

Part one of several that will extract the local metadata providers from the bookshelf plugin and include them in the server to improve book support out of the box.

**Issues**

None

### DIFF
--- a/MediaBrowser.Providers/Books/Isbn/ISBNExternalId.cs
+++ b/MediaBrowser.Providers/Books/Isbn/ISBNExternalId.cs
@@ -1,0 +1,23 @@
+using MediaBrowser.Controller.Entities;
+using MediaBrowser.Controller.Providers;
+using MediaBrowser.Model.Entities;
+using MediaBrowser.Model.Providers;
+
+namespace MediaBrowser.Providers.Books.Isbn
+{
+    /// <inheritdoc />
+    public class IsbnExternalId : IExternalId
+    {
+        /// <inheritdoc />
+        public string ProviderName => "ISBN";
+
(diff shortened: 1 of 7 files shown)


crobibero:
Only the providers that will be moved to server should be included in the server.
The others should be in their respective plugins

dkanada:
@crobibero the idea here is that similar to how NFO files can have TMDB or OMDb links, OPF files can have ISBN or even Google Book IDs and I want to display those in the UI regardless.

I could see a path where Google Books and Comic Vine IDs are ignored unless their providers are installed (don't agree with it but if we plan on pulling out *all* plugin IDs it would make sense) but how about ISBN? Or a possible future Open Library plugin which would be more equivalent to MusicBrainz in terms of a FOSS data archive?

This is kind of a larger question about whether or not we still want to display links from local metadata (trailers from YouTube are another example) even if the user isn't actively pulling data from that provider.

crobibero (approved):

MediaBrowser.Providers/Books/Isbn/ISBNExternalId.cs
```diff
@@ -0,0 +1,23 @@
+using MediaBrowser.Controller.Entities;
+using MediaBrowser.Controller.Providers;
+using MediaBrowser.Model.Entities;
+using MediaBrowser.Model.Providers;
+
+namespace MediaBrowser.Providers.Books.Isbn
```
Bond-009, line 6: Use file scoped namespaces

PR #15604, opened 2025-11-22
extract local metadata from OPF and EPUB files

dkanada:
**Changes**

I have pulled these files from the Bookshelf plugin with minor changes as a PoC for the local book providers. Once everything checks out another pull request will follow for comic books. I'll also have a PR for the parser to extract basic information from the directory structure.

**Issues**

None


nielsvanvelzen:
Looks like this one is causing compile issues now after merging to master.

PR #15623, opened 2025-11-24
feat: Add Live TV channel grouping and filtering capabilities.

Nicell:
**Changes**

- Add a `/LiveTv/ChannelGroups` endpoint so clients can discover available channel groups and their IDs.
- Extend the existing channel/program/guide endpoints to filter by `channelGroupIds` and surface each item’s group membership in the return.
- Update the M3U ingestion pipeline to track multiple group names per channel, normalize/merge them, and expose them consistently. As of #10323 the parser always replaces `tvg-id` on `channel.Id` with the stream URL hash, so only the first duplicate survives. This change should preserve that behavior while merging the extra group metadata into that single entry.
- Introduce the `LiveTvChannelGroupDto` type and supporting `NameGuidPair` mapping so clients receive deterministic IDs alongside display names.

**Issues**

No issues, but this is the API implementation and first step towards implementing [this highly requested feature](https://features.jellyfin.org/posts/186/iptv-channel-groupings).

### DIFF
--- a/Jellyfin.Api/Controllers/LiveTvController.cs
+++ b/Jellyfin.Api/Controllers/LiveTvController.cs
@@ -19,6 +19,7 @@ using Jellyfin.Database.Implementations.Enums;
 using Jellyfin.Extensions;
 using MediaBrowser.Common.Api;
 using MediaBrowser.Common.Configuration;
+using MediaBrowser.Common.Extensions;
 using MediaBrowser.Common.Net;
 using MediaBrowser.Controller.Dto;
 using MediaBrowser.Controller.Entities;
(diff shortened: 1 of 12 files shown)


Nicell:
Disappointed no one took the time to read this even with a message in discord/matrix.

theguymadmax:
@Nicell Apologies for the lack of communication on your PR, this happens more often than it should. There's a small team that does PR reviews, and like you this is done in their free time. Lately, most effort has been dedicated to bug fixes / performance issues with 10.11. And we are currently in the process of fast tracking a performance PR for our next major release (12.0). Due to this we are holding off on any API changes at the moment.

PR #15633, opened 2025-11-26
Add ProgressState field for precise EPUB location tracking

bluesam1:
## Summary

In order to make the 'Books' feature useful, this adds a generic string-based `ProgressState` field to support complex media position tracking that cannot be represented as simple ticks. Primary use case is EPUB CFI (Canonical Fragment Identifier) strings for precise ebook location resumption.

## Problem

EPUB files use CFI strings (e.g., `epubcfi(/6/4[chap01]!/4[body]/10[para05]/3:10)`) to point to exact locations regardless of font size, screen width, or layout changes. The current `PlaybackPositionTicks` field only supports linear percentage values, causing users to lose their reading position when syncing across devices with different screen sizes.

## Solution

- Add `ProgressState` nullable string field to `UserData` database entity
- Create EF Core migration for the new column
- Extend DTOs (`UserItemDataDto`, `UpdateUserItemDataDto`, `PlaybackProgressInfo`) to include `ProgressState`
- Update `UserDataManager` to map and persist `ProgressState`
- Update `SessionManager.OnPlaybackProgress` to save `ProgressState` when provided

## Changes

- **Database:** Added `ProgressState` column to `UserData` table via EF Core migration
- **Models:** Added `ProgressState` property to `UserItemData`, `UserItemDataDto`, `UpdateUserItemDataDto`, and `PlaybackProgressInfo`
- **Business Logic:** Updated mapping and persistence logic in `UserDataManager` and `SessionManager`

## Testing

- Manual testing with EPUB files
- Verified backward compatibility (existing items without `ProgressState` continue to work)
- Database migration tested successfully

## Related

Complements frontend PR: jellyfin/jellyfin-web#7353 (Use ProgressState for precise EPUB location tracking)
Addresses the issue described in the feature proposal for native EPUB CFI support. jellyfin/jellyfin-web/issues/3791


dkanada:
I would almost say this could be `int? | string? Progress` or `string? Cfi` depending on future plans. Ticks aren't exactly a universal format and I don't think FFmpeg expects them - only the C# language.

If the goal is to migrate away from ticks and use something else for FFmpeg streams we could add it to this field in the future, or just continue with the status quo but make this field more targeted with a `Cfi` label.

dkanada:
cc @nielsvanvelzen since this could be a useful API change if we want to change the existing progress fields.

dkanada:
In that case, here is a new proposal in TypeScript limited only to changed fields. We'd leave PlaybackPositionTicks alone indefinitely (everywhere) and mark it as deprecated to steer new developers towards the objects. I added Page for better PDF support in the future and the Duration type you suggested on Matrix.

An alternative name for Progress could be Position to reframe the meaning as a generic "saved location" which could allow for bookmarks or other interesting features in the future.

```
type Progress = Tick | Page | Cfi | Duration;

/** @deprecated use the Duration type for video and audio */
interface Tick {
  Type: 'Tick';
  Data: number;
}

interface Page {
  Type: 'Page';
  Data: number;
}

interface Cfi {
  Type: 'Cfi';
  Data: string;
}

// ISO 8601
interface Duration {
  Type: 'Duration';
  Data: string;
}

interface UserData {
  /** @deprecated use the Progress array in all future endeavors */
  PlaybackPositionTicks: number;
  Progress: Progress[];
}
```

dkanada:
Realized that we also need a `Percentage` type to support progress in UserData for EPUB files.

PR #15638, opened 2025-11-26
Remove special client behavior for API response fields

nielsvanvelzen:
Clients are responsible for collection the item fields they need, the server shouldn't decide that for them.

Note that Android TV hasn't used the name "androidtv" (without space) for quite some time now. The only clients this might affect are Kodi/Roku, in which case we'll need to update them before 10.12.

**Changes**
- Remove DtoExtensions.AddClientFields

**Issues**

### DIFF
--- a/Jellyfin.Api/Controllers/ArtistsController.cs
+++ b/Jellyfin.Api/Controllers/ArtistsController.cs
@@ -122,7 +122,6 @@ public class ArtistsController : BaseJellyfinApiController
     {
         userId = RequestHelpers.GetUserId(User, userId);
         var dtoOptions = new DtoOptions { Fields = fields }
-            .AddClientFields(User)
             .AddAdditionalDtoOptions(enableImages, enableUserData, imageTypeLimit, enableImageTypes);
 
         User? user = null;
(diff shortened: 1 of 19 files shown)


nielsvanvelzen:
Roku devs (@1hitsong): if you use the "ChildCount" property on BaseItemDto's somewhere, make sure it is requested from the server too. With this PR it won't be set unless you ask for it.

Kodi devs: same as above for the Kodi client, but this also affects the "RecursiveItemCount" property.

oddstr13 (approved):
Approving this on behaf of jf4kodi, the parameters are explicitly included, in more places than needed, and where it is needed is only for decorating watched/total episodes of series in the dynamic (jellycon style) views. 

Explicitly falls back to a default of 0 should it ever be missing for any reason.

No impact expected.

1hitsong:
> Roku devs (@1hitsong): if you use the "ChildCount" property on BaseItemDto's somewhere, make sure it is requested from the server too. With this PR it won't be set unless you ask for it.

When you say "make sure it is requested from the server too" do you mean including ChildCount in the fields param.

To put this another way to ensure I understand this change, does PR mean ChildCount is automatically added for some clients regardless of if they ask for it or not and this code change removes that functionality so you only get ChildCount if you include it in the API request?

crobibero:
> and this code change removes that functionality so you only get ChildCount if you include it in the API request?

Yes, ChildCount (and RecursiveChildCount) was automatically added for some clients. This PR removes that super legacy behavior

https://github.com/jellyfin/jellyfin/pull/15638/files#diff-b46ce85d5a39f1eaff39d61b71c02f6d9a89c268de19c45e9816da30e93b65d5

PR #15655, opened 2025-11-28
parse more information from book filenames

dkanada:
**Changes**

This code was also extracted from the bookshelf plugin with minor changes. Once it gets merged I'll be implementing a book series resolver that groups items into collections under the hood.

**Issues**

None


crobibero (approved):
This can certainly be cleaned up/modernized but that can come in further PRs

PR #15662, opened 2025-11-28
Fix blocking in async context in LimitedConcurrencyLibraryScheduler

### LINKED ISSUE #15661: [10.11.3] High bitrate movies causing server freezes during transcoding
### Description of the bug

When playing high bitrate movies, regardless of the device (tested on mobile, desktop browser, and jellyfin-tizen on my S95C), the server freezes intermittently during transcoding. Transcoding is necessary because the movie file is Dolby Vision which is not supported on any of my tested devices. Ffmpeg still proceeds in the background but the entire jellyfin server and webui is unresponsive when the freezes occur. Initially I thought it was a corrupted file, but I tested with other similar versions and the same problem occurred. Next, I thought it was just the video stream buffer under-running, but it isn't, the transcoding continues at a normal pace and my CPU/GPU is not pegged at 100%. The actual jellyfin server process appears to be blocked. Process prioritization and affinity has no impact on the problem. This is all running on a 12-gen i7 processor with 64GB DDR5 and an Arc 770 GPU, so it should be able to handle all of this with no problems. I observe no anomalies or error messages in the server logs and ffmpeg logs before, during, and after the freezing behavior.

### Reproduction steps

1. Start playback of a high-bitrate movie on Jellyfin using any device (e.g., mobile browser, desktop browser, or Jellyfin-Tizen on an S95C TV).
2. Ensure the movie requires transcoding due to containing a Dolby Vision video stream unsupported by the playback device(s).
3. Observe that FFmpeg begins transcoding normally in the background on the server.
4. During playback, note that the Jellyfin server and WebUI intermittently freeze while transcoding continues in the background (observe that ffmpeg log files continue to be generated as well).
5. Verify that the issue is not file-specific by testing multiple similar Dolby Vision movie files and reproducing the same freeze behavior.
6. Confirm that video buffer underrun is not the cause by checking that transcoding proceeds at normal speed without CPU or GPU saturation.
7. Observe that the Jellyfin server process itself becomes blocked or unresponsive during these freezes, testing by using multiple browsers and the Android app, all on the same network segment as the server to ensure network performance is not the problem.
8. Attempt adjusting process prioritization or CPU/GPU affinity and note that these changes have no effect on the issue.

The jellyfin server and webui are completely unresponsive during the transcoding operation, intermittently. Video playback freezes on client devices during this intermittent unresponsiveness, but playback and responsiveness returns after a short time. The freezing behavior occurs seemingly at random, for several seconds. The timing of the freezing behavior and the duration of the freeze is not consistent.

The server should be responsive, and video playback should proceed normally.

### Jellyfin Server version

10.11.3

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.3

### Environment

```markdown
- OS: Windows 11
- Virtualization: None, bare metal
- Clients: Jellyfin Android (latest), Jellyfin-Web, jellyfin-tizen (latest)
- Browser: Mozilla Firefox
- FFmpeg Version: ffmpeg version 7.1.2-Jellyfin
- Playback Method:Transcode
- Hardware Acceleration: QSV
- GPU Model: Arc A770
- Plugins: AudioDB, Chapter Segments Provider, Intro Skipper, Local Intros, MusicBrainz, OMDb, Playback Reporting, Reports, Session Cleaner, Studio Images, Theme Songs, TMDb, TMDb Box Sets
- Reverse Proxy: nginx
- Base URL: none
- Networking: Host
- Jellyfin Data Storage: Local NVMe SSD
- Media Storage: Local HDD array transcoding to local NVMe SSD
- External Integrations: None
```

### Jellyfin logs

```shell
There are no errors in the logs, log activity appears completely normal.
```

### FFmpeg logs

```shell
Uploaded separately.
```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

SapientGuardian:
**Changes**
* Replace `BlockingCollection` with `Channel` in `LimitedConcurrencyLibraryScheduler` to avoid blocking in an async context
* Add SapientGuardian to `CONTRIBUTORS.MD`

**Issues**
Fixes #15661

### DIFF
--- a/MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
+++ b/MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Threading;
+using System.Threading.Channels;
 using System.Threading.Tasks;
 using MediaBrowser.Controller.Configuration;
 using Microsoft.Extensions.Hosting;
@@ -29,7 +30,7 @@ public sealed class LimitedConcurrencyLibraryScheduler : ILimitedConcurrencyLibr
     /// </summary>
     private readonly Lock _taskLock = new();
 
-    private readonly BlockingCollection<TaskQueueItem> _tasks = new();
+    private readonly Channel<TaskQueueItem> _tasks = Channel.CreateUnbounded<TaskQueueItem>();
 
     private volatile int _workCounter;
     private Task? _cleanupTask;
@@ -77,7 +78,7 @@ public sealed class LimitedConcurrencyLibraryScheduler : ILimitedConcurrencyLibr
 
             lock (_taskLock)
             {
-                if (_tasks.Count > 0 || _workCounter > 0)
+                if (_tasks.Reader.Count > 0 || _workCounter > 0)
                 {
                     _logger.LogDebug("Delay cleanup task, operations still running.");
                     // tasks are still there so its still in use. Reschedule cleanup task.
@@ -144,9 +145,9 @@ public sealed class LimitedConcurrencyLibraryScheduler : ILimitedConcurrencyLibr
         _deadlockDetector.Value = stopToken.TaskStop;
         try
         {
-            foreach (var item in _tasks.GetConsumingEnumerable(stopToken.GlobalStop.Token))
+            while (!stopToken.GlobalStop.Token.IsCancellationRequested)
             {
-                stopToken.GlobalStop.Token.ThrowIfCancellationRequested();
+                var item = await _tasks.Reader.ReadAsync(stopToken.GlobalStop.Token).ConfigureAwait(false);
                 try
                 {
                     var newWorkerLimit = Interlocked.Increment(ref _workCounter) > 0;
@@ -264,7 +265,7 @@ public sealed class LimitedConcurrencyLibraryScheduler : ILimitedConcurrencyLibr
         for (var i = 0; i < workItems.Length; i++)
         {
             var item = workItems[i]!;
-            _tasks.Add(item, CancellationToken.None);
+            await _tasks.Writer.WriteAsync(item, CancellationToken.None).ConfigureAwait(false);
         }
 
         Worker();
@@ -283,13 +284,12 @@ public sealed class LimitedConcurrencyLibraryScheduler : ILimitedConcurrencyLibr
         }
 
         _disposed = true;
-        _tasks.CompleteAdding();
+        _tasks.Writer.Complete();
         foreach (var item in _taskRunners)
         {
             await item.Key.CancelAsync().ConfigureAwait(false);
         }
 
-        _tasks.Dispose();
         if (_cleanupTask is not null)
         {
             await _cleanupTask.ConfigureAwait(false);


drspangle:
I can confirm that the updated MediaBrowser.Controller.dll fixes #15661. Tested with parallel library scan tasks limit and parallel image encoding limit set to 1, 2, 4, and null (default). The freezing behavior is gone.

MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
```diff
@@ -295,6 +296,28 @@ void UpdateProgress()
...
 
+    /// <summary>
+    /// Waits for for the specified interval OR CancellationToken to retrieve an item from the channel reader. Returns null if either condition elapses.
+    /// </summary>
+    /// <typeparam name="T">Type of item.</typeparam>
+    /// <param name="channelReader">Channel reader from which to read.</param>
+    /// <param name="interval">Interval to wait before aborting.</param>
+    /// <param name="linkedToken">CancellationToken to trigger abort.</param>
+    /// <returns>Retrieved item, or null if one of the supplied conditions elapses.</returns>
+    private static async Task<T?> ChannelTryTake<T>(ChannelReader<T> channelReader, TimeSpan interval, CancellationToken linkedToken)
```
cvium: This is pretty much what WaitAsync does

SapientGuardian: `WaitAsync` doesn't return the value if available, it only tells you if/when a value becomes available. In a multi-threaded context like this, one of the other readers could take the item between `WaitAsync` telling you one was available and you attempting to retrieve it with `WaitAsync`, so it doesn't really get the job done here. You'd have to implement some sort of double-checked locking pattern, which would only increase the complexity.

cvium: I think you may be thinking of [`WaitToReadAsync`](https://learn.microsoft.com/en-us/dotnet/api/system.threading.channels.channelreader-1.waittoreadasync?view=net-9.0#system-threading-channels-channelreader-1-waittoreadasync(system-threading-cancellationtoken)). I am talking about [Task.WaitAsync](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.waitasync?view=net-9.0#system-threading-tasks-task-waitasync(system-timespan-system-threading-cancellationtoken)).

Image

SapientGuardian: Sorry, I misread your comment. I don't believe we can use `WaitAsync` here because the time-based cancellation is not forwarded to the inner `ReadAsync` call, which would result in the value still getting read from the channel, but then discarded.

image

theguymadmax:
This should target the release branch. Please rebase and drop all commits that are not part of this PR.

SapientGuardian:
> This should target the release branch. Please rebase and drop all commits that are not part of this PR.

Done. It would be helpful if https://jellyfin.org/docs/general/contributing/development#official-branches reflected that as standard procedure (unless of course you're considering this an emergency hotfix).

crobibero:
@SapientGuardian There are now merge conflicts after merging a different PR

MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
SapientGuardian: With the latest changes in the release branch, this method is no longer necessary and has been removed.

SapientGuardian:
> @SapientGuardian There are now merge conflicts after merging a different PR

I have re-rebased.

drspangle:
Any updates on the progress of this PR being reviewed?

cvium (approved):

Issue #15643, opened 2025-11-27
System.InvalidOperationException thrown when playing Stream without File-Extension


martenumberto:
### Describe the Bug
When attempting to play a Live TV stream (via M3U Tuner) where the source URL does not contain a file extension (e.g., http://provider/stream/12345?token=abc), the server throws a System.InvalidOperationException: Failed to find an appropriate file extension and returns a 500 error to the client.
This happens in Jellyfin.Api.Helpers.StreamingHelpers.GetOutputFileExtension. The method fails to return a valid extension when:
- Path.GetExtension(state.RequestedUrl) is empty/null.
- The requested VideoCodec is not in the explicit list of checked codecs (e.g., mpeg2video is missing from the list).
- mediaSource.Container is not yet populated (likely due to probing race conditions on initial playback).

### Steps to Reproduce
Add an M3U Tuner provider.
Ensure the M3U contains stream URLs that do not have a file extension (e.g., http://192.168.1.10/stream/1001).
Attempt to play the channel (e.g., on Android TV or Web Client).
Playback fails immediately.
Check server logs for InvalidOperationException.

### Expected Behavior
The server should default to a safe fallback extension (like .ts) for video requests if the specific extension cannot be determined, allowing ffmpeg to handle the stream probing/remuxing instead of crashing the request.

### Additional Context
Upon further analysis, the stream in question utilized the mpeg2video codec. The current implementation of GetOutputFileExtension checks for h264, hevc, vp8, etc., but is missing a check for mpeg2video. Furthermore, relying solely on the URL structure is brittle for IPTV providers that use dynamic URLs without extensions.
I have prepared a fix that adds a fallback to .ts for state.IsVideoRequest at the end of GetOutputFileExtension, which resolves the crash and allows playback to start successfully (verified locally).
I will be submitting a PR for this shortly.

### Logs
```
[2025-11-27 10:54:58.037 +01:00] [ERR] [10] Jellyfin.Api.Middleware.ExceptionMiddleware: Error processing request. URL "GET" "/Videos/f705a814-7be3-ec88-18ec-7391510fde6a/stream".
System.InvalidOperationException: Failed to find an appropriate file extension
   at Jellyfin.Api.Helpers.StreamingHelpers.GetOutputFileExtension(StreamState state, MediaSourceInfo mediaSource) in C:\Users\user1\Documents\GitHub\jellyfin\Jellyfin.Api\Helpers\StreamingHelpers.cs:line 356
[...]

PR #15644, opened 2025-11-27
Fix: Prevent crash in GetOutputFileExtension for streams without URL extension

martenumberto:
### Describe the changes
This PR adds a defensive fallback to `Jellyfin.Api.Helpers.StreamingHelpers.GetOutputFileExtension`. 

When handling a video request where:
1. The URL does not contain a file extension (common in IPTV/M3U streams with tokens),
2. The video codec is not in the hardcoded list (e.g. `mpeg2video`),
3. And the `MediaSource.Container` is not yet populated (due to race conditions during initial probing),

...the server currently throws an `InvalidOperationException`, resulting in an HTTP 500 error and playback failure.

This change adds a default fallback to `.ts` (Transport Stream) for video requests when no other extension can be determined. This allows the stream to be passed to ffmpeg for remuxing/transcoding instead of crashing the request.

### Fixes
Fixes https://github.com/jellyfin/jellyfin/issues/15643
We should considering backport to 10.11.x

### How to test
1. Add an M3U Tuner with streams that do not have a file extension in the URL (e.g., `http://provider/stream/12345?token=abc`).
2. Ensure the stream uses a codec not explicitly handled in `StreamingHelpers` (e.g., `mpeg2video`) or trigger a cold start where probing data is missing.
3. Attempt to play the stream.
4. **Before this fix:** Server logs show `System.InvalidOperationException: Failed to find an appropriate file extension` and playback fails.
5. **After this fix:** Playback starts successfully (typically remuxing to TS).

### Checklist
- [x] I have read the contributing guidelines.
- [x] I have verified that the change works locally.
- [x] The code follows the project's coding standards.

PR #15690, opened 2025-12-01
Fix: Add .ts fallback for video streams to prevent crash

martenumberto:
### Describe the changes
This PR adds a defensive fallback to `Jellyfin.Api.Helpers.StreamingHelpers.GetOutputFileExtension`. 

When handling a video request where:
1. The URL does not contain a file extension (common in IPTV/M3U streams with tokens),
2. The video codec is not in the hardcoded list (e.g. `mpeg2video`),
3. And the `MediaSource.Container` is not yet populated (due to race conditions during initial probing),

...the server currently throws an `InvalidOperationException`, resulting in an HTTP 500 error and playback failure.

This change adds a default fallback to `.ts` (Transport Stream) for video requests when no other extension can be determined. This allows the stream to be passed to ffmpeg for remuxing/transcoding instead of crashing the request.

### Fixes
Fixes https://github.com/jellyfin/jellyfin/issues/15643
Duplicate of https://github.com/jellyfin/jellyfin/pull/15644 cause of rebase trouble.

### How to test
1. Add an M3U Tuner with streams that do not have a file extension in the URL (e.g., `http://provider/stream/12345?token=abc`).
2. Ensure the stream uses a codec not explicitly handled in `StreamingHelpers` (e.g., `mpeg2video`) or trigger a cold start where probing data is missing.
3. Attempt to play the stream.
4. **Before this fix:** Server logs show `System.InvalidOperationException: Failed to find an appropriate file extension` and playback fails.
5. **After this fix:** Playback starts successfully (typically remuxing to TS).

### Checklist
- [x] I have read the contributing guidelines.
- [x] I have verified that the change works locally.
- [x] The code follows the project's coding standards.

### DIFF
--- a/Jellyfin.Api/Helpers/StreamingHelpers.cs
+++ b/Jellyfin.Api/Helpers/StreamingHelpers.cs
@@ -159,7 +159,9 @@ public static class StreamingHelpers
 
         string? containerInternal = Path.GetExtension(state.RequestedUrl);
 
-        if (string.IsNullOrEmpty(containerInternal) && !string.IsNullOrWhiteSpace(streamingRequest.LiveStreamId))
+        if (string.IsNullOrEmpty(containerInternal)
+            && (!string.IsNullOrWhiteSpace(streamingRequest.LiveStreamId)
+                || (mediaSource != null && mediaSource.IsInfiniteStream)))
         {
             containerInternal = ".ts";
         }


gnattu:
Adding fallback for liveTV is okay as we already have a lot of fallbacks (LiveTV never works in ideal ways), but this is probably not the best place to add the fallback. Because this will alter the behavior for all kind of playbacks and we have far more use cases that is not for livestreams. I *think* it should be better added in StreamingHelper when `streamingRequest.LiveStreamId` has a value and directly assign containerInternal with a fallback value when `state.RequestedUrl` has no extention.

martenumberto:
I've change that with keeping your advice in mind.
Since "streamingRequest.LiveStreamId" can be null, we should also check if mediaSource.IsInfiniteStream==true.

crobibero (approved):

PR #15698, opened 2025-12-02
Fix stack overflow during scan (#15000)

myzhysz:
In function MediaBrowser.Controller.LibraryTaskScheduler.Enqueue, if we enter the "we are in a nested loop" block, the logic try get a item from _tasks, making function Enqueue not process the file in subfolder, but start from a new folder which increase Call Stack. This make the length of Call Stack is not depends on the depth of filesystem, but the count of subfolders. Therefore, if there are lots of subfolders, the Call Stack will exceed a certain limit and cause Stack Overflow.

**Changes**
When _deadlockDetector.Value is not null, using existed ForceSequentialOperation.

**Issues**
Fixes https://github.com/jellyfin/jellyfin/issues/15000

### DIFF
--- a/MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
+++ b/MediaBrowser.Controller/LibraryTaskScheduler/LimitedConcurrencyLibraryScheduler.cs
@@ -242,7 +242,7 @@ public sealed class LimitedConcurrencyLibraryScheduler : ILimitedConcurrencyLibr
             };
         }).ToArray();
 
-        if (ShouldForceSequentialOperation())
+        if (ShouldForceSequentialOperation() || _deadlockDetector.Value is not null)
         {
             _logger.LogDebug("Process sequentially.");
             try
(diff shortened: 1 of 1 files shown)


JPVenson:
Can you give an example of what kind of folder structure causes this?

The new code has the same issue as the previous scheduler that single folders get processed only sequentially so if you have one folder with a lot of episodes, only a single worker will work on that which is a very bad use of resources

myzhysz:
My folder is like this.
D:\PIC\#
D:\PIC\#\Folder1-4000
D:\PIC\L1\Folder1-300
D:\PIC\L1\L2\Folder1-300
D:\PIC\L1\L2\L3\Folder1-300
...
There are total 17K+ subfolders and 600K+ picture files in PIC folder.

phmg701:
> The new code has the same issue as the previous scheduler that single folders get processed only sequentially so if you have one folder with a lot of episodes, only a single worker will work on that which is a very bad use of resources

For what it's worth, I tested the compiled DLL on 10.11.4 with a directory containing 13.882 files and 4.759 folders. It took 69 seconds to complete, for the first time in months without errors.

Bond-009 (approved):

PR #15713, opened 2025-12-04
Locking cleaning

MarkCiliaVincenti:


Shadowghost (approved):

crobibero (changes requested):
If a plugin wants to consume AsyncKeyedLock it should reference the dependency itself.

Please revert all unneeded changes to csproj

MarkCiliaVincenti:
> If a plugin wants to consume AsyncKeyedLock it should reference the dependency itself.
> 
> Please revert all unneeded changes to csproj

Would it be able to do so reliably though? What if it's referencing a different version to what is applied on jellyfin?

crobibero:
> > If a plugin wants to consume AsyncKeyedLock it should reference the dependency itself.
> 
> > 
> 
> > Please revert all unneeded changes to csproj
> 
> 
> 
> Would it be able to do so reliably though? What if it's referencing a different version to what is applied on jellyfin?

If the plugin needs to include a library that the server doesn't load it can include it in the zip. If the server does include it then it uses the server's version 

As long as there aren't breaking changes to the ABI between the version included in plugin and server then there won't be any issue

MarkCiliaVincenti:
@crobibero 
And what if the plugin and the server both include it? Will there be a conflict?

crobibero (approved):

PR #15718, opened 2025-12-05
Library: Async the SaveImages function

luigi311:
**Changes**
Change the SameImages database function to be async. Added a cancellation token but it is unused. Gonna work through converting the other functions to be async aswell and if its called at a higher level see if that can be converted to async, if not create a wrapper that can exist until the top layers are converted to async calls.

**Issues**

### DIFF
--- a/Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
+++ b/Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
@@ -559,25 +559,27 @@ public sealed class BaseItemRepository
         var images = item.ImageInfos.Select(e => Map(item.Id, e)).ToArray();
 
         var context = await _dbProvider.CreateDbContextAsync(cancellationToken).ConfigureAwait(false);
-
-        if (!await context.BaseItems
+        await using (context.ConfigureAwait(false))
+        {
+            if (!await context.BaseItems
                 .AnyAsync(bi => bi.Id == item.Id, cancellationToken)
                 .ConfigureAwait(false))
-        {
-            _logger.LogWarning("Unable to save ImageInfo for non existing BaseItem");
-            return;
+            {
+                _logger.LogWarning("Unable to save ImageInfo for non existing BaseItem");
+                return;
+            }
+
+            await context.BaseItemImageInfos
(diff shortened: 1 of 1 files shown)


Bond-009 (approved):

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -552,22 +552,32 @@ public ItemCounts GetItemCounts(InternalItemsQuery filter)
...
-    public void SaveImages(BaseItemDto item)
+    public async Task SaveImagesAsync(BaseItemDto item, CancellationToken cancellationToken = default)
     {
         ArgumentNullException.ThrowIfNull(item);
 
-        var images = item.ImageInfos.Select(e => Map(item.Id, e));
-        using var context = _dbProvider.CreateDbContext();
+        var images = item.ImageInfos.Select(e => Map(item.Id, e)).ToArray();
+
+        var context = await _dbProvider.CreateDbContextAsync(cancellationToken).ConfigureAwait(false);
```
crobibero, line 561: Looks like this is missing a dispose

luigi311: Looks like when i try using, using, it gives me an error asking for me to ConfigureAwait and it seems to come down to the compiler expanding out for that Dispose and there doesnt seem to be a way for me to tell it to expand with a configure await on it so i manually expanded it out. How does that look?

luigi311: if this looks good i can squash

crobibero: Here's an example of the `await using` pattern
https://github.com/jellyfin/jellyfin/blob/45cb5a0008b27e326b8e14071ea03039dc6040c9/Jellyfin.Server.Implementations/Devices/DeviceManager.cs#L78-L79

luigi311: Ohh thank you for that, i swapped over to this method now

PR #15722, opened 2025-12-05
Library/BaseItemRepository: Async the DeleteItems function

luigi311:
**Changes**
Move the DeleteItems to async calls in the database as much as possible. Some of the functions that call it are blocking functions so i left some wrappers for the original calls so those continue to function.

**Issues**

### DIFF
--- a/Emby.Server.Implementations/Library/LibraryManager.cs
+++ b/Emby.Server.Implementations/Library/LibraryManager.cs
@@ -327,18 +327,6 @@ namespace Emby.Server.Implementations.Library
             await DeleteItemAsync(item, options, parent, notifyParentItem, cancellationToken).ConfigureAwait(false);
         }
 
-        // Legacy sync wrappers:
-        public void DeleteItem(BaseItem item, DeleteOptions options)
-            => DeleteItem(item, options, false);
-
-        public void DeleteItem(BaseItem item, DeleteOptions options, bool notifyParentItem)
-            => DeleteItemAsync(item, options, notifyParentItem, CancellationToken.None)
-                .GetAwaiter().GetResult();
-
-        public void DeleteItem(BaseItem item, DeleteOptions options, BaseItem parent, bool notifyParentItem)
-            => DeleteItemAsync(item, options, parent, notifyParentItem, CancellationToken.None)
-                .GetAwaiter().GetResult();
-
         public async Task DeleteItemsUnsafeFastAsync(IEnumerable<BaseItem> items, CancellationToken cancellationToken = default)
         {
             var pathMaps = items.Select(e => (Item: e, InternalPath: GetInternalMetadataPaths(e), DeletePaths: e.GetDeletePaths())).ToArray();
--- a/Emby.Server.Implementations/Library/LibraryManager.cs
+++ b/Emby.Server.Implementations/Library/LibraryManager.cs
@@ -1185,7 +1185,7 @@ namespace Emby.Server.Implementations.Library
 
             if (toDelete.Count > 0)
             {
-                _itemRepository.DeleteItem(toDelete.ToArray());
+                await _itemRepository.DeleteItemsAsync(toDelete.ToArray()).ConfigureAwait(false);
             }
         }
 
--- a/Emby.Server.Implementations/Library/LibraryManager.cs
+++ b/Emby.Server.Implementations/Library/LibraryManager.cs
@@ -313,21 +313,33 @@ namespace Emby.Server.Implementations.Library
             _cache.AddOrUpdate(item.Id, item);
         }
 
-        public void DeleteItem(BaseItem item, DeleteOptions options)
+        public async Task DeleteItemAsync(BaseItem item, DeleteOptions options, CancellationToken cancellationToken = default)
         {
-            DeleteItem(item, options, false);
+            await DeleteItemAsync(item, options, false, cancellationToken: cancellationToken).ConfigureAwait(false);
         }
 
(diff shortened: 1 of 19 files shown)


luigi311:
setting as draft for now. The code is pretty much done but i need to make sure its actually working correctly.

luigi311:
This is ready to review, sorry had some logging on my side for some other functions to test something and it was spamming the logs so i thought something was wrong. 

Tested by adding in all my non music libraries, removed one of the movie libraries and the count did go down, added it back in and it went back correctly to the right amount. This should be good to review

Shadowghost (commented):
Since this targets master, we can also remove the sync wrappers I guess. Not sure which way we want to go

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -101,14 +101,22 @@ public BaseItemRepository(
...
+        CancellationToken cancellationToken = default)
     {
         if (ids is null || ids.Count == 0 || ids.Any(f => f.Equals(PlaceholderId)))
         {
             throw new ArgumentException("Guid can't be empty or the placeholder id.", nameof(ids));
         }
 
-        using var context = _dbProvider.CreateDbContext();
-        using var transaction = context.Database.BeginTransaction();
+        var context = await _dbProvider.CreateDbContextAsync(cancellationToken).ConfigureAwait(false);
```
JPVenson:
```suggestion
               var dbContext = await _dbContextFactory.CreateDbContextAsync(cancellationToken).ConfigureAwait(false);
        await using (dbContext.ConfigureAwait(false))
        {
        
```

Emby.Server.Implementations/Library/LibraryManager.cs
```diff
@@ -313,21 +313,33 @@ public void RegisterItem(BaseItem item)
...
         }
 
-        public void DeleteItemsUnsafeFast(IEnumerable<BaseItem> items)
+        // Legacy sync wrappers:
+        public void DeleteItem(BaseItem item, DeleteOptions options)
+            => DeleteItem(item, options, false);
+
+        public void DeleteItem(BaseItem item, DeleteOptions options, bool notifyParentItem)
+            => DeleteItemAsync(item, options, notifyParentItem, CancellationToken.None)
+                .GetAwaiter().GetResult();
```
JPVenson: not happy about this

JPVenson (changes requested):

Emby.Server.Implementations/Library/LibraryManager.cs
luigi311: yeah not happy about them either but those are just wrappers to hold over the external calls that are blocking until they are made to async calls. The alternative would be to go and modify those scattered calls and put that GetAwaiter().GetResult() on them there and then down the line have to do a search for all of those instances to clean them out. This lets us search specifically for DeleteItem which is a function for LibraryManager that should never exist anymore. Up to you guys which method you prefer. I can make all those GetAwaiter swaps in a second commit to keep them easy to find.

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
luigi311: Ive renamed the context to dbContext. I see you changed from _db_provider to _dvContextFactory, that doesnt exist, the BaseItemReposity only has _dbProvider of IDbContextFactory<JellyfinDbContext>, are you asking me to rename class variable from _dbProvider to _dvContextFactory and update all the references?

Emby.Server.Implementations/Library/LibraryManager.cs
cvium: Searching for a method or searching for GetAwaiter is kinda the same, so I don't see the point in keeping the legacy methods.

Also, just FYI, SQLite does not support async and it is the only DB we currently support, so I'm not sure if now is the correct time to make these changes considering we have a lot of issues with our current query patterns (postgres would not automatically fix it).

luigi311: After sleeping on it and based on cvium comment i agree, we can drop the blocking calls and instead add in the getawaiter getresults. Down the line we can do a search for all those references and start phasing them out.

luigi311: > Also, just FYI, SQLite does not support async and it is the only DB we currently support, so I'm not sure if now is the correct time to make these changes considering we have a lot of issues with our current query patterns (postgres would not automatically fix it).

Not sure what you mean, this is a change on the jellyfin side of the equation not for anything on the sqlite side, making this async along with the entire upper calls, at some point down the line, has the benefit of prevent a thread lockup in jellyfin while this function is being called and removing the need to work around that. 

I'm assuming what you talking about on the sqlite side is the fact that you can not do multiple writes at the same time, that has no affect on this because this being async because other functions doing select statements should continue to function unimpeded without having to wait for the database lock to be lifted. This have the benefit of making sure for example the UI continues to function for navigation atleast.

JPVenson: the underlying provider for sqlite does not support IO driven async methods, so even if we call the async varients of efcore, they are never truely async but are only wrapped by EFCore. 

That was more of an FYI

soultaco83:
There is a deleteitem in the video.cs. Line 517
would this need to be updated as well to async as well?
```
LibraryManager.DeleteItem(orphanedVideo, new DeleteOptions { DeleteFileLocation = false });
```

Issue #14095, opened 2025-05-12
[Group] Jellyfin Memory Usage / Out Of Memory related issues


felix920506:
Issue group for everything related to Jellyfin using up a lot of ram and/or being oom killed

gnattu:
> The unmerged PR that changes the locking reduces the ram usage down from 6 GB or so to < 1GB. Major difference while scanning and using some endpoints was the removal from duplicates.
> 
> [fba36eb](https://github.com/jellyfin/jellyfin/commit/fba36ebf23d280a4353b371d0fca0a848ab5e0f8)

This can also be explained by reduced concurrency and therefore lower memory fragmentation comes from concurrency.

If you have time, can you test the docker image build upon 10.10.7 in this comment: https://github.com/jellyfin/jellyfin/issues/11588#issuecomment-2869367328

To see if the memory footprint goes down compared to original 10.10?

gnattu:
> * 10.11 master - I had up to 8 GB ram used (or sometimes even more). Stuck queries via search bar for people was a big performance issue

Well what I meant is how my modified 10.10 image performs because my image is not supposed to be "leaky" on 10.10.

thraxarious:
Adding my $0.02 here. 
Server version 10.10.7 
on Debian 6.1.137-1 (2025-05-07)
Running as a basic service not in Docker. 

               total        used        free      shared  buff/cache   available
Mem:            19Gi        18Gi       171Mi        85Mi       940Mi       705Mi
Swap:          975Mi       975Mi          0B

-rw-r--r--  1 jellyfin jellyfin 1.2G May 23 21:42 library.db-wal
-rw-r--r--  1 jellyfin jellyfin  33M May 23 21:42 library.db-shm
-rw-r--r--  1 jellyfin jellyfin  32K May 23 21:41 jellyfin.db-shm
-rw-r--r--  1 jellyfin jellyfin 157K May 23 21:41 jellyfin.db-wal
-rw-r--r--  1 jellyfin jellyfin  17G May 23 21:39 library.db
drwxr-xr-x  6 jellyfin jellyfin 4.0K May 23 21:32 .
-rw-r--r--  1 jellyfin jellyfin 1.4M May 23 04:14 jellyfin.db
drwxr-xr-x 14 jellyfin jellyfin 4.0K May 21 23:59 keyframes
drwxr-xr-x 18 jellyfin jellyfin 4.0K May  8 23:26 subtitles
[...]

PR #15724, opened 2025-12-05
FFProbe: Reduce probesize to 500M

luigi311:
**Changes**
Reduce the ffprobe probesize down from 1G -> 500M. With 500M i havent seen any warnings asking to increase the probsize or duration in the logs after doing a full scan on my movies. This brings down the average memory usage for my full movies scan down from 3.38 GB -> 2.75 GB. Should have a bigger impact for servers that are using a higher parallel scan count.

Memory usage before

probe_1g

Memory usage after

probe_500m

**Issues**
Relates to #14095

### DIFF
--- a/Emby.Server.Implementations/ConfigurationOptions.cs
+++ b/Emby.Server.Implementations/ConfigurationOptions.cs
@@ -15,7 +15,7 @@ namespace Emby.Server.Implementations
         {
             { HostWebClientKey, bool.TrueString },
             { DefaultRedirectKey, "web/" },
-            { FfmpegProbeSizeKey, "1G" },
+            { FfmpegProbeSizeKey, "500M" },
             { FfmpegAnalyzeDurationKey, "200M" },
             { BindToUnixSocketKey, bool.FalseString },
             { SqliteCacheSizeKey, "20000" },


nyanmisaka:
> With 500M i havent seen any warnings asking to increase the probsize or duration in the logs after doing a full scan on my movies.

This is not really true for a wider range of videos. Based on my testing with a 68.9GB MKV video, that requires 1GB to allow some metadata to be detected.

luigi311:
I tested with up to 61 gb mkv video and didnt get any issues. I dont have anything higher i can test with

luigi311:
Actually it looks like the warnings just dont show up in the logs i think. I manually called ffprobe on that 61gb mkv file and i do see the warning but its the same issue even with 1G

[matroska,webm @ 0x7cd6baaed600] Could not find codec parameters for stream 6 (Subtitle: hdmv_pgs_subtitle (pgssub)): unspecified size
Consider increasing the value for the 'analyzeduration' (200000000) and 'probesize' (1000000000) options



Looks like no reasonable amount will fix that issue, 10G analyze duration and 10G probesize still gives the warning

[matroska,webm @ 0x769a685ff600] Could not find codec parameters for stream 6 (Subtitle: hdmv_pgs_subtitle (pgssub)): unspecified size
Consider increasing the value for the 'analyzeduration' (10000000000) and 'probesize' (10000000000) options

gnattu:
This is already configurable via env var. instead of changing the default value (that is designed to be safe), maybe you can try to expose this via api settings so that this is more accessible to users,

nyanmisaka:
Closing this is because it can already be set via environment variables. Future improvements could focus on setting these parameters via the API.

PR #15734, opened 2025-12-06
Some small optimizations

MarkCiliaVincenti:


Jellyfin.Api/Helpers/StreamingHelpers.cs
```diff
@@ -319,17 +319,13 @@ private static string GetOutputFileExtension(StreamState state, MediaSourceInfo?
...
-                    return ".mp3";
-
                 case "vorbis":
                     return ".ogg";
 
+                case "aac":
+                case "mp3":
                 case "wma":
-                    return ".wma";
+                    return $".{audioCodec}";
```
EraYaN: Not sure this is actually faster, since the string interpolation might be slower than the actual switch.

Jellyfin.Api/Helpers/StreamingHelpers.cs
```diff
@@ -287,62 +287,49 @@ private static string GetOutputFileExtension(StreamState state, MediaSourceInfo?
             return ext;
         }
 
-        // Try to infer based on the desired video codec
         if (state.IsVideoRequest)
         {
-            var videoCodec = state.Request.VideoCodec;
+            // Try to infer based on the desired video codec
+            var videoCodec = state.Request.VideoCodec.ToLowerInvariant();
```
EraYaN: You should normalize this to uppercase as a best practice. And it might still be slower since it allocates a new string.

Jellyfin.Api/Helpers/StreamingHelpers.cs
MarkCiliaVincenti: Fixed.

Jellyfin.Api/Helpers/StreamingHelpers.cs
MarkCiliaVincenti: Normalized as requested. It won't be slower than a multiple case invariant search.

Jellyfin.Api/Helpers/StreamingHelpers.cs
```diff
@@ -287,62 +287,50 @@ private static string GetOutputFileExtension(StreamState state, MediaSourceInfo?
...
-                return ".ogg";
-            }
+                case "MP3":
+                    return ".mp3";
 
-            if (string.Equals("wma", audioCodec, StringComparison.OrdinalIgnoreCase))
-            {
-                return ".wma";
+                case "WMA":
+                    return $".wma";
```
NiklasBr:
```suggestion
                    return ".wma";
```

Maybe? I understand that it is an interpolated string when the dollar sign is there, and because I am a newbie here I might not understand the reason it was added, I guess it was by mistake because it was not in the previous version?

MarkCiliaVincenti: Good call. It should be compiled to the same result but it's unnecessary there.

NiklasBr (approved):

cvium:
What's the motivation here exactly? And if this is optimizing something, I would like to see some numbers.

MarkCiliaVincenti:
> What's the motivation here exactly? And if this is optimizing something, I would like to see some numbers.

You probably asked this half-seriously, but I got intrigued and ran some numbers haha.

Seems like for the first entries, the existing method is faster, but then somewhere halfway through the proposed method becomes faster. More of a fun experiment than anything.

BenchmarkDotNet v0.15.8, Windows 10 (10.0.19045.6466/22H2/2022Update)
Intel Core i7-3632QM CPU 2.20GHz (Ivy Bridge), 1 CPU, 8 logical and 4 physical cores
.NET SDK 10.0.101
  [Host]     : .NET 10.0.1 (10.0.1, 10.0.125.57005), X64 RyuJIT x86-64-v2 [AttachedDebugger]
  DefaultJob : .NET 10.0.1 (10.0.1, 10.0.125.57005), X64 RyuJIT x86-64-v2


| Method          | VideoCodec | Mean      | Error     | StdDev    | Median    | Ratio | RatioSD | Allocated | Alloc Ratio |
|---------------- |----------- |----------:|----------:|----------:|----------:|------:|--------:|----------:|------------:|
| OriginalMethod  | AV1        |  4.698 ns | 0.1866 ns | 0.1996 ns |  4.634 ns |  1.00 |    0.06 |         - |          NA |
| OptimizedMethod | AV1        |  8.094 ns | 0.1908 ns | 0.1692 ns |  8.036 ns |  1.73 |    0.08 |         - |          NA |
|                 |            |           |           |           |           |       |         |           |             |
| OriginalMethod  | H264       |  1.762 ns | 0.0603 ns | 0.0535 ns |  1.750 ns |  1.00 |    0.04 |         - |          NA |
| OptimizedMethod | H264       | 10.037 ns | 0.1734 ns | 0.1537 ns | 10.066 ns |  5.70 |    0.19 |         - |          NA |
|                 |            |           |           |           |           |       |         |           |             |
| OriginalMethod  | HEVC       |  3.539 ns | 0.1636 ns | 0.1530 ns |  3.468 ns |  1.00 |    0.06 |         - |          NA |
| OptimizedMethod | HEVC       |  9.461 ns | 0.1867 ns | 0.1655 ns |  9.430 ns |  2.68 |    0.12 |         - |          NA |
|                 |            |           |           |           |           |       |         |           |             |
| OriginalMethod  | THEORA     |  6.234 ns | 0.1658 ns | 0.1470 ns |  6.261 ns |  1.00 |    0.03 |         - |          NA |
| OptimizedMethod | THEORA     | 11.041 ns | 0.3180 ns | 0.4857 ns | 10.843 ns |  1.77 |    0.09 |         - |          NA |
|                 |            |           |           |           |           |       |         |           |             |
| OriginalMethod  | VP8        |  8.141 ns | 0.1665 ns | 0.1557 ns |  8.115 ns |  1.00 |    0.03 |         - |          NA |
| OptimizedMethod | VP8        |  7.981 ns | 0.1619 ns | 0.1864 ns |  7.910 ns |  0.98 |    0.03 |         - |          NA |
|                 |            |           |           |           |           |       |         |           |             |
| OriginalMethod  | VP9        | 10.519 ns | 0.2733 ns | 0.2422 ns | 10.475 ns |  1.00 |    0.03 |         - |          NA |
| OptimizedMethod | VP9        |  8.206 ns | 0.2600 ns | 0.4883 ns |  7.989 ns |  0.78 |    0.05 |         - |          NA |
|                 |            |           |           |           |           |       |         |           |             |
| OriginalMethod  | VPX        | 12.516 ns | 0.2093 ns | 0.1855 ns | 12.438 ns |  1.00 |    0.02 |         - |          NA |
| OptimizedMethod | VPX        |  8.707 ns | 0.1744 ns | 0.1546 ns |  8.662 ns |  0.70 |    0.02 |         - |          NA |
|                 |            |           |           |           |           |       |         |           |             |
| OriginalMethod  | WMV        | 13.616 ns | 0.3644 ns | 0.3742 ns | 13.558 ns |  1.00 |    0.04 |         - |          NA |
| OptimizedMethod | WMV        |  8.424 ns | 0.2218 ns | 0.2465 ns |  8.342 ns |  0.62 |    0.02 |         - |          NA |


````
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Configs;
using BenchmarkDotNet.Diagnosers;
using System;
using System.Text;

namespace BenchmarkTest;

[Config(typeof(MemoryConfig))]
[MemoryDiagnoser]
[JsonExporterAttribute.Full]
[JsonExporterAttribute.FullCompressed]
public class Benchmarks
{
    private class MemoryConfig : ManualConfig
    {
        public MemoryConfig()
        {
            AddDiagnoser(MemoryDiagnoser.Default);
        }
    }

    [Params("H264", "HEVC", "AV1", "THEORA", "VP8", "VP9", "VPX", "WMV")] public string VideoCodec { get; set; }

    [Benchmark(Baseline = true)]
    public string OriginalMethod()
    {
        var videoCodec = VideoCodec;

        if (string.Equals(videoCodec, "h264", StringComparison.OrdinalIgnoreCase))
        {
            return ".ts";
        }

        if (string.Equals(videoCodec, "hevc", StringComparison.OrdinalIgnoreCase)
            || string.Equals(videoCodec, "av1", StringComparison.OrdinalIgnoreCase))
        {
            return ".mp4";
        }

        if (string.Equals(videoCodec, "theora", StringComparison.OrdinalIgnoreCase))
        {
            return ".ogv";
        }

        if (string.Equals(videoCodec, "vp8", StringComparison.OrdinalIgnoreCase)
            || string.Equals(videoCodec, "vp9", StringComparison.OrdinalIgnoreCase)
            || string.Equals(videoCodec, "vpx", StringComparison.OrdinalIgnoreCase))
        {
            return ".webm";
        }

        if (string.Equals(videoCodec, "wmv", StringComparison.OrdinalIgnoreCase))
        {
            return ".asf";
        }

        return null;
    }

    [Benchmark]
    public string OptimizedMethod()
    {
        var videoCodec = VideoCodec.ToUpperInvariant();

        switch (videoCodec)
        {
            case "H264":
                return ".ts";

            case "HEVC":
            case "AV1":
                return ".mp4";

            case "THEORA":
                return ".ogv";

            case "VP8":
            case "VP9":
            case "VPX":
                return ".webm";

            case "WMV":
                return ".asf";
        }

        return null;
    }
}
````

cvium:
I was full serious. We're not interested in unproven "optimizations".

Your benchmark is not correct. Your params are all uppercase so the call to ToUpperInvariant doesn't actually create a new string.

MarkCiliaVincenti:
You're right @cvium, I'll close this.

PR #15735, opened 2025-12-06
OGG mime/extension lookups

MarkCiliaVincenti:
I believe that this was entered unintentionally. Please confirm. I also have some doubts about others, like audio/webm.

Also what's this?
new(".tbn", "image/jpeg"),


sjorge:
> new(".tbn", "image/jpeg"),

.tbn are thumbnail files from kodi, they are just jpegs with a different extension.

MarkCiliaVincenti:
> > new(".tbn", "image/jpeg"),
> 
> .tbn are thumbnail files from kodi, they are just jpegs with a different extension.

OK, but then the 'reverse' is this:

`new("image/jpeg", ".jpg"),`

Shouldn't they be exactly the same, or not?

gnattu:
> > > new(".tbn", "image/jpeg"),
> > 
> > 
> > .tbn are thumbnail files from kodi, they are just jpegs with a different extension.
> 
> OK, but then the 'reverse' is this:
> 
> `new("image/jpeg", ".jpg"),`
> 
> Shouldn't they be exactly the same, or not?

Why? Jellyfin should not save files with kodi specific extension, but jellyfin should recognize jpeg with kodi flavor extension, two different use cases.

MarkCiliaVincenti:
OK, so I should close this PR @gnattu ?

gnattu:
I think you can close this.

PR #15737, opened 2025-12-06
Flip default ConfigureAwait value

MarkCiliaVincenti:
Massive PR, I know.

But reading all the .ConfigureAwait(false) is so annoying.

This PR sets ConfigureAwait as default through a source generator.
It will also show errors if someone sets .ConfigureAwait(false) but no errors on .ConfigureAwait(true) in case this needs to be set to true.

Fully expecting this to be rejected because, I understand, such a big change and so many merge conflicts will happen, but I took a chance. The code looks much cleaner now.


MarkCiliaVincenti:
Not exactly sure why the checks are failing.

"It works on my machine".

Let's see if you're interested before we figure out a fix.

luigi311:
it might be easier to read through this if you split this into two commits, one that enables this functionality and a second one where its literally just a removal of all the .ConfigureAwait(false)

MarkCiliaVincenti:
> it might be easier to read through this if you split this into two commits, one that enables this functionality and a second one where its literally just a removal of all the .ConfigureAwait(false)

In hindsight I agree but too much work now.

cvium:
Fody has an annoying license model. We will not be using it.

MarkCiliaVincenti:
> Fody has an annoying license model. We will not be using it.

Ah, I see. What if I made something similar fully free and under the MIT licence?

cvium:
I can only speak for myself: I think the downside to having such a massive change dwarfs the very small upside.

MarkCiliaVincenti:
> I can only speak for myself: I think the downside to having such a massive change dwarfs the very small upside.

What downsides do you see, apart from the obvious short-term mess and merge conflicts?

cvium:
> > I can only speak for myself: I think the downside to having such a massive change dwarfs the very small upside.
> 
> What downsides do you see, apart from the obvious short-term mess and merge conflicts?

A source generator can also contain bugs. Writing `.ConfigureAwait(false)` is not a big problem/pain. It is also a no-op for pretty much all deployments we support. The amount of work testing, reviewing, bug fixing etc. is a lot compared to just including `.ConfigureAwait(false)`.

MarkCiliaVincenti:
I understand. Would be nice if .NET came out with a means to set the default ConfigureAwait value in the csproj, although I can see how it can lead to a lot of undesirable outcomes.

PR #15751, opened 2025-12-08
Update Authorization Controller endpoints to be more explicit

PeachesMLG:
Change the AuthenticateUserByName endpoint to return a 401 with `Invalid Authorization Cookie` instead of a generic 400 Error

Add 404 onto the AuthenticateWithQuickConnect so the docs get [updated here
](https://api.jellyfin.org/#tag/User/operation/AuthenticateWithQuickConnect)

### DIFF
--- a/Jellyfin.Api/Controllers/UserController.cs
+++ b/Jellyfin.Api/Controllers/UserController.cs
@@ -251,6 +251,7 @@ public class UserController : BaseJellyfinApiController
     /// <returns>A <see cref="Task"/> containing an <see cref="AuthenticationRequest"/> with information about the new session.</returns>
     [HttpPost("AuthenticateWithQuickConnect")]
     [ProducesResponseType(StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
     public ActionResult<AuthenticationResult> AuthenticateWithQuickConnect([FromBody, Required] QuickConnectDto request)
     {
         try


Jellyfin.Api/Controllers/UserController.cs
```diff
@@ -240,9 +247,11 @@ public async Task<ActionResult<AuthenticationResult>> AuthenticateUserByName([Fr
     /// <param name="request">The <see cref="QuickConnectDto"/> request.</param>
     /// <response code="200">User authenticated.</response>
     /// <response code="400">Missing token.</response>
+    /// <response code="404">Quick Connect code not found.</response>
     /// <returns>A <see cref="Task"/> containing an <see cref="AuthenticationRequest"/> with information about the new session.</returns>
     [HttpPost("AuthenticateWithQuickConnect")]
     [ProducesResponseType(StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
```
PeachesMLG, line 254: Im assuming these then generate the Swagger contracts and be automatically updated to the api website?

https://api.jellyfin.org/#tag/User/operation/AuthenticateWithQuickConnect

Jellyfin.Api/Controllers/UserController.cs
```diff
@@ -205,13 +205,20 @@ public async Task<ActionResult<AuthenticationResult>> AuthenticateUser(
     /// </summary>
     /// <param name="request">The <see cref="AuthenticateUserByName"/> request.</param>
     /// <response code="200">User authenticated.</response>
+    /// <response code="401">Invalid authorization cookie.</response>
     /// <returns>A <see cref="Task"/> containing an <see cref="AuthenticationRequest"/> with information about the new session.</returns>
     [HttpPost("AuthenticateByName")]
     [ProducesResponseType(StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
```
crobibero, line 212: 401 is already added by https://github.com/jellyfin/jellyfin/blob/master/Jellyfin.Server/Filters/SecurityRequirementsOperationFilter.cs

PeachesMLG: But it doesnt show up on the api docs https://api.jellyfin.org/#tag/User/operation/AuthenticateUserByName
Which I assume is auto generated based off these fields.

PeachesMLG:
> * The Auth header is not a cookie, so the error message doesn't make sense
> 
>     * The exception is thrown in SessionManager:AuthenticateNewSessionInternal
> 
> 
> ```
> [20:27:32.414] [ERR] [34] Jellyfin.Api.Middleware.ExceptionMiddleware: Error processing request. URL POST /Users/authenticatebyname.
> System.ArgumentNullException: Value cannot be null. (Parameter 'request.App')
>    at System.ArgumentNullException.Throw(String paramName)
>    at System.ArgumentNullException.ThrowIfNull(Object argument, String paramName)
>    at System.ArgumentException.ThrowNullOrEmptyException(String argument, String paramName)
>    at System.ArgumentException.ThrowIfNullOrEmpty(String argument, String paramName)
>    at Emby.Server.Implementations.Session.SessionManager.AuthenticateNewSessionInternal(AuthenticationRequest request, Boolean enforcePassword) in /home/crobibero/Code/jellyfin/Emby.Server.Implementations/Session/SessionManager.cs:line 1588
> ```
> 
> To make this error clearer we would need to update the exception logging
> 
> In SessionManager:AuthenticateNewSessionInternal change the
> 
> ```cs
> ArgumentException.ThrowIfNullOrEmpty(request.App);
> ```
> 
> with
> 
> ```cs
> if (string.IsNullOrEmpty(request.App))
> {
>     throw new ArgumentException("Authorization header not complete", nameof(request.App));
> }
> ```
> 
> This should probably be extracted to a static method to de-duplicate code

Im less worried about the logging personally, Im more talking about what the API returns. I (personally) dont think a developer should have to go looking through the logs to see what went wrong in this scenario. I think that should be returned in the response code / body

Open to thoughts on this though

cvium:
If login methods are too verbose, someone will open a CVE claiming a user enumeration vulnerability

PeachesMLG:
> If login methods are too verbose, someone will open a CVE claiming a user enumeration vulnerability

For sure, I dont believe the level of information we are providing here warrents that though

Jellyfin.Api/Controllers/UserController.cs
nielsvanvelzen, line 254: Why does a 404 need to be added here? Does this endpoint actually return 404 responses?

Jellyfin.Api/Controllers/UserController.cs
```diff
@@ -205,13 +205,20 @@ public async Task<ActionResult<AuthenticationResult>> AuthenticateUser(
...
     public async Task<ActionResult<AuthenticationResult>> AuthenticateUserByName([FromBody, Required] AuthenticateUserByName request)
     {
         var auth = await _authContext.GetAuthorizationInfo(Request).ConfigureAwait(false);
 
+        if (auth.Client == null || auth.Version == null || auth.DeviceId == null || auth.Device == null)
+        {
+            return Unauthorized("Invalid Authorization Cookie");
```
nielsvanvelzen, line 219: This is wrong, the input is incorrect so it should be a bad request

Jellyfin.Api/Controllers/UserController.cs
PeachesMLG: yep if no quickconnect code is found - hoping adding this will update https://api.jellyfin.org/#tag/User/operation/AuthenticateUserByName

Jellyfin.Api/Controllers/UserController.cs
PeachesMLG: Yeah, I debated on this a lil bit, cuz technically its an Authorization header, but its also not actually used to authorize anything. Happy to change it

Jellyfin.Api/Controllers/UserController.cs
nielsvanvelzen: Yes these attributes are for the OpenAPI metadata, if the endpoint is able to respond with 404 already then it should be added here.

Jellyfin.Api/Controllers/UserController.cs
nielsvanvelzen: If the values are missing (null) the client didn't send a correct request. If the values are all provided but invalid (e.g. token is set to some nonsense string) THEN it is unauthorized.

PR #15760, opened 2025-12-10
fix: watched status resetting on re-watch

MSalman5230:
Preservation of Watched Status on Re-watch

**Changes**
 This change modifies the playback start logic to preserve the "Watched" status of media items when they are re-watched. Previously, starting playback on any item that supports resume would immediately mark the item as "Unwatched". This update ensures that Played status is only set to false if the item was not previously marked as played, allowing completed items to remain marked as completed during a re-watch session


xelu:
why was this closed? :(

PR #15762, opened 2025-12-10
Fix watched status resetting on re-watch

### LINKED ISSUE #15761: watched status resetting on re-watch
### Description of the bug

Watched media changes to unwatch if we open/rewatch the media again

### Reproduction steps

1. Completely watch a media.
2. It will be marked as watched.
3. Play the media again. For example, you want to rewatch a part in the middle.
4. rewatch a scene and close it.
5. Media status changed to unwatched, and it shows continue from X time

Media status changed to unwatched, and it shows continue from X time. Even though I have already watched the media once

It should stay marked as watched.

### Jellyfin Server version

10.11.4

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.4

### Environment

```markdown
- OS:Debian 12
- Linux Kernel:
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:QSV
- CPU Model:Intel N100
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Media Storage & Filesystem:
- External Integrations:
```

### Jellyfin logs

```shell
NA
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

MSalman5230:
Preservation of Watched Status on Re-watch

**Changes**
This change modifies the playback start logic to preserve the "Watched" status of media items when they are re-watched. Previously, starting playback on any item that supports resume would immediately mark the item as "Unwatched". This update ensures that Played status is only set to false if the item was not previously marked as played, allowing completed items to remain marked as completed during a re-watch session.

**Issues**
 Fixes #15761


Emby.Server.Implementations/Session/SessionManager.cs
```diff
@@ -822,7 +822,7 @@ private void OnPlaybackStart(User user, BaseItem item)
             {
                 data.Played = true;
             }
-            else
+            else if (!data.Played)
```
nielsvanvelzen: You can just remove the else clause if it's not going to change the value at al

nielsvanvelzen (changes requested):

Emby.Server.Implementations/Session/SessionManager.cs
MSalman5230: Yeah, that's correct.

CONTRIBUTORS.md
```diff
@@ -281,3 +281,4 @@
  - [Martin Reuter](https://github.com/reuterma24)
  - [Michael McElroy](https://github.com/mcmcelro)
  - [Soumyadip Auddy](https://github.com/SoumyadipAuddy)
+ - [MSalman5230](https://github.com/MSalman5230)
```
nielsvanvelzen: Why did you add yourself as Emby contributor?

MSalman5230: Yeah fixed it.

nielsvanvelzen (approved):

Shadowghost (approved):

joshuaboniface (changes requested):
How does this handle the - very real - situation of rewatching an already watched show? Are the "next up" episodes properly handled with this change?

joshuaboniface (approved):
My question was answered to my satisfaction out-of-band. Approved.

thornbill (dismissed):
Is the goal just to keep watched indicators active? That could easily be managed by a client side option that just checks `PlayCount`. 

I think we need a very clear documented list of what all this will impact. You mention in the [corresponding issue](https://github.com/jellyfin/jellyfin/issues/15761#issuecomment-3637037577) this prevents items from showing in the Continue Watching section on the Home Screen. What else does this impact? Are you still able to resume partially played content?

MSalman5230:
PlayCount does not check whether the media was completely watched (i.e., marked as watched). It only tracks whether the media was played and how many times.

Now, regarding the next part, you raise a good question.
With this change, you can still resume from the Continue Watching section even if the media is marked as watched. So this does not fix the messiness of the “Continue Watching” section on the home screen.

Now we need to decide what we want:

Should we keep this behavior, or should we change the logic used to fetch items for the Continue Watching section?

The current behavior still works. At least you can see the red checkmark if you finished watching something. It also helps you know which episode or movie you’ve already watched, and it preserves the resume/last position if you’re rewatching something.

I think the change I made is the best solution.

If we change the retrieval logic, we’ll lose the ability to resume already watched media (which could be a dealbreaker for some users)

Bond-009 (approved):

Issue #15141, opened 2025-10-23
Performance - Folders make library loading drastically slower in 10.11.0


User-PDB3D:
### Description of the bug

After successfully upgrading to version 10.11.0, me and others users have noticed that when you have to load folders in a library, the loading of the library page becomes drastically slower. 
this is confirmed to happen in book libraries, folder view, and mixed libraries that utilize folders.
if you make a page in a library have less entries than the default 100, and some pages have folders and some don't, you can see a drastic difference between loading times of those pages.

in cases of manga libraries that utilize a lot of folders to put individual volumes of the same franchise together, the load times can be as drastic as needing to take multiple minutes, until the library is loaded.

going back to 10.10.7 showed this is a 10.11.0 issue.

for clarification, when i mean "Folders", i'm talking about an entry in a library with a folder icon at the top right.

### Reproduction steps

1. Update to 10.11.0
2. make a library of the relevant category (for example, a books library)
3. make a file structure that causes a lot of folders to appear
4. try to load the library

### What is the current _bug_ behavior?

The library loads very slowly

### What is the expected _correct_ behavior?

the library loads at an acceptable speed

### Jellyfin Server version

10.10.0+

### Specify the build version

10.11.0

### Environment

```markdown
- OS: Server: Ubuntu in VM on proxmox, 
Client that tries to load library: Windows 10, Android TV

- Linux Kernel:
- Clients: Android TV, Jellyfin Media Player for windows
- Browser: Librewolf
- Jellyfin Data Storage: SSD
- Media Storage: HDD with passthrough to VM
```

### Jellyfin logs

```shell
n/a
```

### FFmpeg logs

```shell

```

### Relevant screenshots or videos

I have provided a video comparing load times between 10.10.7 and 10.11.0.
the videos are named after their version number. the video that shows 10.11.0 is the longer one

https://github.com/user-attachments/assets/d1e02b7c-1424-4077-ac28-a376f2849dbd
https://github.com/user-attachments/assets/1c17c376-055e-478b-832d-4db3970d1088

chrissi55:
Maybe additionally to this topic is that the book library crash or only show a blanc page during entering the library from startpage when books are organized in Sub/Sub/Sub/Sub/... directories.

Figured out that jellyfin 10.11.0 seems to handle maximum a /Sub/Sub/PDF-file tree not deeper. Otherwise the library stops working and scans are slowed down enormously.

[...]

PR #15772, opened 2025-12-12
Reimplement caching for GetCachedChildren()

alexanderwise:
If my spelunking is correct, back in 0bd1f36ece `GetCachedChildren()` would actually use a cache, hence the method name.
Then in 95250b9f22 & f4d61ddcc6, `GetCachedChildren()` would no longer use the cache, but would simply return GUIDs (`ActualChildren` -> `GetItemByID`) instead of full objects.
Then b09a41ad1f (EF Core Migration) changed the method back to returning full objects (`ItemRepository.GetItemList()`), but now without a cache.
I think this contributes to the performance issues when opening folder-based library views in 10.11.0

This PR adds the cache lookup back into GetCachedChildren(), batching fetches for misses into a single query.
It might be more correct to add some kind of `ActualChildren` method back into the EF Core changes. Happy to close this in favor of that if that is what the project wants for the EF Core migration, but would be a much bigger change.

I chose not to reimplement the old `LibraryManager.GetMemoryItemById(guid)` method because it's less efficient to recurse over each item rather than batch.

In local testing this showed improvement in performance for #15141 and related issues.

**Changes**
- Add GetItemsByIds() to ILibraryManager for batch item retrieval
- Implement cache-aware batch fetching in LibraryManager that checks the LRU cache first and batch-fetches misses in a single query
- Update GetCachedChildren() to get IDs first then resolve items through the cache.

**Issues**
Partial fix for https://github.com/jellyfin/jellyfin/issues/15141

### DIFF
--- a/MediaBrowser.Controller/Library/ILibraryManager.cs
+++ b/MediaBrowser.Controller/Library/ILibraryManager.cs
@@ -211,7 +211,8 @@ namespace MediaBrowser.Controller.Library
         /// </summary>
         /// <param name="ids">The item ids to retrieve.</param>
         /// <returns>The items found, in the same order as the input ids. Missing items are excluded.</returns>
-        IReadOnlyList<BaseItem> GetItemsByIds(IReadOnlyList<Guid> ids);
+        IReadOnlyList<BaseItem> GetItemsByIds(IReadOnlyList<Guid> ids)
+            => throw new NotImplementedException();
 
         /// <summary>
         /// Gets the intros.


CLAUDE.md
```diff
@@ -0,0 +1,36 @@
+# Claude Code Guidelines for This Repository
```
JPVenson: nope not gonna happen

alexanderwise: oh yeah sorry didn't mean to commit that.

PR #15776, opened 2025-12-12
Fix AV1 decoding hang regression on RK3588

nyanmisaka:
**Changes**
- Fix AV1 decoding hang regression on RK3588

**Issues**
- The AFBC image output by AV1 decoder is invalid if the video width is not a multiple of 64 and this will reset the RGA3 hardware. Awaiting feedback from upstream. Disable it for now. https://forum.jellyfin.org/t-rk3588-do-not-hw-transcode-anymore

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -7039,8 +7039,8 @@ namespace MediaBrowser.Controller.MediaEncoding
 
                 if (string.Equals(videoStream.Codec, "av1", StringComparison.OrdinalIgnoreCase))
                 {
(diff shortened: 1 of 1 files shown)


Shadowghost (approved):

gnattu (approved):

gnattu:
> Why was something like this needed?

Welcome to the world of imperfect hardware :)

In fact hardware quirks like this is very common but most vendors would implement their own workaround in the driver/acceleration library. But such workaround is not present in rockchip driver(yet).

nyanmisaka:
It hasn't actually been confirmed that this is a hardware defect. It's entirely possible that it's a software bug caused by incorrectly programmed hardware registers. But that's beyond my capabilities and needs to be handled by the chip manufacturer themselves.

Issue #15742, opened 2025-12-06
10.11.4 is slow to load


Ondjultomte:
### Description of the bug

Upgraded 10.11.4 now the dashboard /controlpanel and general webb gui is really slow.

### Reproduction steps

Upgraded 10.11.4 now the dashboard /controlpanel and general webb gui is really slow.

### What is the current _bug_ behavior?

Upgraded 10.11.4 now the dashboard /congtrolpannel and general webb gui is really slow, some dasgboard nanes just get stuck. see below screenshots.

what do you need to help track down these issues?

<img width="1258" height="1217" alt="Image" src="https://github.com/user-attachments/assets/f973c22c-6e23-461b-bf10-2c6d81550671" />

### What is the expected _correct_ behavior?

fast , responsive web gui

### Jellyfin Server version

10.11.4

### Specify the build version

10.11.4

### Environment

```markdown
- OS:
- Linux Kernel:
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- CPU Model:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Jellyfin Data Storage & Filesystem:
- Media Storage & Filesystem:
- External Integrations:
```

### Jellyfin logs

```shell
at Jellyfin.Api.Middleware.ExceptionMiddleware.Invoke(HttpContext context)
[2025-12-06 20:19:15.427 +00:00] [ERR] Error reporting playback progress
System.MissingMethodException: Method not found: 'System.Collections.Generic.List`1<Jellyfin.Data.Entities.User> MediaBrowser.Controller.Library.PlaybackProgressEventArgs.get_Users()'.
   at Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint.SessionManager_PlaybackProgress(Object sender, PlaybackProgressEventArgs e)
   at Emby.Server.Implementations.Session.SessionManager.OnPlaybackProgress(PlaybackProgressInfo info, Boolean isAutomated)
   at MediaBrowser.Controller.Session.SessionInfo.OnProgressTimerCallback(Object state)
[2025-12-06 20:19:16.427 +00:00] [ERR] Error reporting playback progress
System.MissingMethodException: Method not found: 'System.Collections.Generic.List`1<Jellyfin.Data.Entities.User> MediaBrowser.Controller.Library.PlaybackProgressEventArgs.get_Users()'.
[...]

PR #15803, opened 2025-12-16
Removing N+1 queries from the /items/Latest endpoint

JackDanger:
This PR address the N+1 queries on the /Latest endpoint that loads libraries into the home screen.

Prior to this change a sizable library (thousands of movies, hundreds of thousands of audio tracks) would not load faster than 30s on even very fast hardware.
After this change the endpoint returns even very large collections within 10 seconds.

This may address https://github.com/jellyfin/jellyfin/issues/15742

### DIFF
--- a/Emby.Server.Implementations/Dto/DtoService.cs
+++ b/Emby.Server.Implementations/Dto/DtoService.cs
@@ -160,10 +160,17 @@ namespace Emby.Server.Implementations.Dto
             List<(BaseItem, BaseItemDto)>? programTuples = null;
             List<(BaseItemDto, LiveTvChannel)>? channelTuples = null;
 
+            // Batch-fetch user data for all items to avoid N+1 queries
+            Dictionary<Guid, UserItemData>? userDataBatch = null;
+            if (user is not null && options.EnableUserData)
+            {
+                userDataBatch = _userDataRepository.GetUserDataBatch(accessibleItems, user);
+            }
+
             for (int index = 0; index < accessibleItems.Count; index++)
             {
                 var item = accessibleItems[index];
-                var dto = GetBaseItemDtoInternal(item, options, user, owner);
+                var dto = GetBaseItemDtoInternal(item, options, user, owner, userDataBatch);
 
                 if (item is LiveTvChannel tvChannel)
                 {
@@ -197,7 +204,7 @@ namespace Emby.Server.Implementations.Dto
 
         public BaseItemDto GetBaseItemDto(BaseItem item, DtoOptions options, User? user = null, BaseItem? owner = null)
         {
-            var dto = GetBaseItemDtoInternal(item, options, user, owner);
+            var dto = GetBaseItemDtoInternal(item, options, user, owner, null);
             if (item is LiveTvChannel tvChannel)
             {
                 LivetvManager.AddChannelInfo(new[] { (dto, tvChannel) }, options, user);
@@ -215,7 +222,7 @@ namespace Emby.Server.Implementations.Dto
             return dto;
         }
 
-        private BaseItemDto GetBaseItemDtoInternal(BaseItem item, DtoOptions options, User? user = null, BaseItem? owner = null)
+        private BaseItemDto GetBaseItemDtoInternal(BaseItem item, DtoOptions options, User? user = null, BaseItem? owner = null, Dictionary<Guid, UserItemData>? userDataBatch = null)
         {
             var dto = new BaseItemDto
             {
@@ -252,7 +259,7 @@ namespace Emby.Server.Implementations.Dto
 
             if (user is not null)
             {
-                AttachUserSpecificInfo(dto, item, user, options);
+                AttachUserSpecificInfo(dto, item, user, options, userDataBatch);
             }
 
             if (item is IHasMediaSources
@@ -458,7 +465,7 @@ namespace Emby.Server.Implementations.Dto
         /// <summary>
         /// Attaches the user specific info.
         /// </summary>
-        private void AttachUserSpecificInfo(BaseItemDto dto, BaseItem item, User user, DtoOptions options)
+        private void AttachUserSpecificInfo(BaseItemDto dto, BaseItem item, User user, DtoOptions options, Dictionary<Guid, UserItemData>? userDataBatch = null)
         {
             if (item.IsFolder)
             {
@@ -466,7 +473,17 @@ namespace Emby.Server.Implementations.Dto
 
                 if (options.EnableUserData)
                 {
-                    dto.UserData = _userDataRepository.GetUserDataDto(item, dto, user, options);
+                    if (userDataBatch is not null && userDataBatch.TryGetValue(item.Id, out var batchedUserData))
+                    {
+                        // Use pre-fetched user data
+                        dto.UserData = GetUserItemDataDto(batchedUserData, item.Id);
+                        item.FillUserDataDtoValues(dto.UserData, batchedUserData, dto, user, options);
+                    }
+                    else
+                    {
+                        // Fall back to individual fetch
+                        dto.UserData = _userDataRepository.GetUserDataDto(item, dto, user, options);
+                    }
                 }
 
                 if (!dto.ChildCount.HasValue && item.SourceType == SourceType.Library)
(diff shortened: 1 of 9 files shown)


Emby.Server.Implementations/Dto/DtoService.cs
```diff
@@ -215,7 +222,7 @@ public BaseItemDto GetBaseItemDto(BaseItem item, DtoOptions options, User? user
             return dto;
         }
 
-        private BaseItemDto GetBaseItemDtoInternal(BaseItem item, DtoOptions options, User? user = null, BaseItem? owner = null)
+        private BaseItemDto GetBaseItemDtoInternal(BaseItem item, DtoOptions options, User? user = null, BaseItem? owner = null, Dictionary<Guid, UserItemData>? userDataBatch = null)
```
cvium: As far as I can tell you don't actually use the batch anywhere. The signature could just be `UserItemData? userData`, i.e.,

```C#
                var dto = GetBaseItemDtoInternal(item, options, user, owner, userDataBatch?.GetValueOrDefault(item.Id));

File: Emby.Server.Implementations/Library/UserDataManager.cs [hunk 2]
```diff
@@ -244,6 +244,83 @@ private static string GetCacheKey(long internalUserId, Guid itemId)
...
+            var allItemIds = itemsNeedingQuery.Select(x => x.Item.Id).ToList();
+            var allKeys = itemsNeedingQuery.SelectMany(x => x.Keys).Distinct().ToList();
+
+            if (allKeys.Count > 0)
+            {
+                using var context = _repository.CreateDbContext();
+                var userDataArray = context.UserData
+                    .AsNoTracking()
+                    .Where(e => allItemIds.Contains(e.ItemId) && allKeys.Contains(e.CustomDataKey) && e.UserId.Equals(user.Id))
+                    .ToArray();
```
MarkCiliaVincenti, line 294: Consider switching to async here, as painful as it sounds with the references

MarkCiliaVincenti (commented):
Great PR, have a nitpick.

cvium:
@JackDanger Can you fix my comment and retarget the release-10.11.z branch?

cedricwalter:
yes! this is an urgent PR, the system is so slow it crash sometimes with lots of audio tracks.

kiawizard:
@JackDanger please return to us with your mega-important PR! :-) Or, if you don't have time, give me permission to do what cvium asks for

JackDanger:
I'm back from holiday vacation and never before have I been so popular. I'm returning to this PR!

Emby.Server.Implementations/Dto/DtoService.cs
JackDanger: The batch gets plumbed through `AttachUserSpecificInfo` and then used by `GetUserItemDataDto` to fetch already-loaded user data.

This is kinda clunky though, since I really only need the item id. I've added a new commit that simplifies this.

JackDanger:
Okay, updated! There was a little simplification to be done but the logic was sound.

I've now exhaustively tested this against some large libraries and even a half-million tracks return in less than 10 seconds. TV and Movies are much faster than that.

If we want to add a database index on (album, track, created) for quickly selecting recent tracks this can get faster, but this improvement is a lot better than nothing.

I think this is ready!

crobibero:
Please also re-target this change onto the release-10.11.z branch

JackDanger:
I think this is now ready to ship. 

* Preserves the exact existing ordering logic
* Avoids N+1 queries
* Attempts to land the code in a more maintainable, legible state
* Works for even very large libraries (thousands of TV episodes, hundreds of thousands of music tracks)

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -306,44 +306,210 @@ public IReadOnlyList<BaseItem> GetLatestItemList(InternalItemsQuery filter, Coll
...
+        }
+        else
+        {
+            mainquery = mainquery.Where(e => topParentIds.Contains(e.SeriesId!.Value));
+        }
+
         mainquery = ApplyGroupingFilter(context, mainquery, filter);
         mainquery = ApplyQueryPaging(mainquery, filter);
-
+        mainquery = mainquery.AsSplitQuery();
```
JPVenson, line 379: generally we do not want to use split query. This has been attempted by a lot of people before you in PRs (including me) and in my testing all query performance on average was slower. The additional queries may not be of so much effect with other DBMS but with sqlite in particular and on devices with already constrained filesystem backends, this is not an option. It might be better for you, but it will be worse for most sadly.

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -306,44 +306,210 @@ public IReadOnlyList<BaseItem> GetLatestItemList(InternalItemsQuery filter, Coll
...
         mainquery = ApplyNavigations(mainquery, filter);
 
         return mainquery.AsEnumerable().Where(e => e is not null).Select(w => DeserializeBaseItem(w, filter.SkipDeserialization)).ToArray();
     }
 
+    /// <inheritdoc />
+    public IReadOnlyList<(BaseItem Container, IReadOnlyList<BaseItem> Items)> GetLatestItemsGrouped(InternalItemsQuery filter, CollectionType? collectionType)
```
JPVenson, line 386: i am not a huge fan of having public methods with named tuples. Better use an actual poco here

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -306,44 +306,210 @@ public IReadOnlyList<BaseItem> GetLatestItemList(InternalItemsQuery filter, Coll
...
+                .AsNoTracking()
+                .Where(e => e.Type == episodeTypeName)
+                .Where(e => e.SeriesId.HasValue)
+                .Where(e => filter.TopParentIds.Contains(e.TopParentId!.Value))
+                .GroupBy(e => e.SeriesId!.Value)
+                .Select(g => new { SeriesId = g.Key, MaxDate = g.Max(e => e.DateCreated) })
+                .OrderByDescending(g => g.MaxDate)
+                .Take(limit)
+                .Select(g => g.SeriesId)
+                .ToList();
```
JPVenson, line 446: this should be one parameterised query not two

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -306,44 +306,210 @@ public IReadOnlyList<BaseItem> GetLatestItemList(InternalItemsQuery filter, Coll
...
+        }
+
+        if (topParentIds.Count == 0)
+        {
+            return Array.Empty<(BaseItem, IReadOnlyList<BaseItem>)>();
```
JPVenson, line 451:
```suggestion
            return [];
```

MediaBrowser.Controller/Library/IUserDataManager.cs
```diff
@@ -64,6 +64,28 @@ public interface IUserDataManager
         /// <returns>User data dto.</returns>
         UserItemDataDto? GetUserDataDto(BaseItem item, BaseItemDto? itemDto, User user, DtoOptions options);
 
+        /// <summary>
+        /// Gets user data for multiple items in a batch.
+        /// </summary>
+        /// <param name="items">Items to get user data for.</param>
+        /// <param name="user">User to use.</param>
+        /// <returns>Dictionary mapping item ID to user data.</returns>
+        Dictionary<Guid, UserItemData> GetUserDataBatch(IReadOnlyList<BaseItem> items, User user)
```
JPVenson, line 73: i dont see an adventage of having this as an interface method

JPVenson (changes requested):
first review

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
JackDanger: lol, good idea

JackDanger:
@JPVenson thanks for the careful review! I’ll address all of it once I get a chance 🙇

theguymadmax:
@JackDanger Thanks for your contribution. I commented in the other N+1 PR (#15986), but forgot to address you as well. Apologies for that. 

The team is prioritizing PR #16062, which also addresses the performance issues in 10.11.x. Once that PR is merged, it will serve as the foundation for the 12.0 release. As a result, this PR will no longer be reviewed/considered for the current release. We can revisit additional improvements once #16062 is merged.

theguymadmax:
Superseded by https://github.com/jellyfin/jellyfin/pull/16062

Issue #15283, opened 2025-10-28
Music Artists Not Correct on Tracks with Multiple 'artist' Entries


lexitronic:
### Description of the bug

On Jellyfin 10.11+, and when adding a song with multiple 'artist' metadata entries, Jellyfin seems to randomise the artists is adds to its database, often excluding the artist that is also in the 'albumartist' field if they exist. This happens regardless of if I use the "Replace existing metadata", or "Search for missing metadata" options, or remove and re-add the library.

### Reproduction steps

0. Fresh Jellyfin configuration, with no changes to default settings.
1. Add Music library.
2. Add song with multiple 'artist' metadata entries.
3. Scan for new songs.
4. Go to the album that was added.
5. Look at the metadata that Jellyfin stored for the song using "Edit Metadata".
6. its wrong

### What is the current _bug_ behavior?

The "artists" field in the "Edit Metadata" menu has different data to the file itself.

<img width="1154" height="755" alt="Image" src="https://github.com/user-attachments/assets/f0b4a16d-ca33-4a30-bb7d-d74932f1aae7" />

### What is the expected _correct_ behavior?

The "artists" field in the "Edit Metadata" menu matches the metadata in the file.

<img width="1145" height="764" alt="Image" src="https://github.com/user-attachments/assets/76e5cb05-23a3-49bb-b43d-b8010034c714" />

### Jellyfin Server version

10.10.0+

### Specify version number

10.11.1

### Specify the build version

10.11.1

### Environment

```markdown
- OS: Debian 13
- Linux Kernel: 6.12.48+deb13-amd64
- Virtualization: Docker in KVM
- Clients: Browser
- Browser: Firefox 143.0.1
- FFmpeg Version: 7.1.2-Jellyfin
- Playback Method: None
- Hardware Acceleration: None
- GPU Model: None
- Plugins: None
- Reverse Proxy: None
- Base URL: None
- Networking: Bridge
- Jellyfin Data Storage: Local SATA SSD
- Media Storage: Local SATA SSD
- External Integrations: None
```

### Jellyfin logs

```shell
See attachments
```

### FFmpeg logs

```shell

```

### Relevant screenshots or videos

Library setup settings on creation, 10.10.7:
<img width="1201" height="3417" alt="Image" src="https://github.com/user-attachments/assets/a457e131-cc48-4757-95d9-daaec8aa2f71" />

Library setup settings on creation, 10.11.1:
[...]

PR #15808, opened 2025-12-16
Trim music artist names

theguymadmax:
When multiple artists are saved with extra whitespace, matching fails. 

**Changes**
Trim whitespace from artist names on save

**Issues**
Fixes https://github.com/jellyfin/jellyfin/issues/15283#issuecomment-3658541044


gnattu:
I don’t think this is the correct way of fixing though? We are still saving entires with white spaces and that is the root problem. We should fix this on save time, not on find time. 

Also I’m not a big fan of doing trimming in base item queries as such things would cause unexpected surprises.

Shadowghost:
Can't remove my review, but agree with gnattu.

gnattu (approved):

crobibero (approved):

PR #15809, opened 2025-12-16
Migration to the newer .slnx format

MarkCiliaVincenti:


Jellyfin.slnx
```diff
@@ -0,0 +1,77 @@
...
+    <Property Name="AssemblyInfoVersionSettings" Value="None.None.None.None" />
+    <Property Name="AssemblyInfoVersionType" Value="SettingsVersion" />
+    <Property Name="AssemblyVersionSettings" Value="None.None.None.None" />
+    <Property Name="InheritWinAppVersionFrom" Value="None" />
+    <Property Name="UpdateAssemblyFileVersion" Value="True" />
+    <Property Name="UpdateAssemblyInfoVersion" Value="True" />
+    <Property Name="UpdateAssemblyVersion" Value="True" />
+    <Property Name="UpdatePackageVersion" Value="False" />
+  </Properties>
+  <Properties Name="MonoDevelopProperties">
```
MarkCiliaVincenti, line 70: Are these still needed @joshuaboniface or are they leftovers from the Mono days?

Jellyfin.slnx
```diff
@@ -0,0 +1,77 @@
...
+    <File Path="SharedVersion.cs" />
+  </Folder>
+  <Folder Name="/src/">
+    <Project Path="src/Jellyfin.CodeAnalysis/Jellyfin.CodeAnalysis.csproj" />
+    <Project Path="src/Jellyfin.Drawing.Skia/Jellyfin.Drawing.Skia.csproj" />
+    <Project Path="src/Jellyfin.Drawing/Jellyfin.Drawing.csproj" />
+    <Project Path="src/Jellyfin.Extensions/Jellyfin.Extensions.csproj" />
+    <Project Path="src/Jellyfin.LiveTv/Jellyfin.LiveTv.csproj" />
+    <Project Path="src/Jellyfin.MediaEncoding.Hls/Jellyfin.MediaEncoding.Hls.csproj">
+      <BuildType Project="Debug" />
```
MarkCiliaVincenti, line 15: Is this on purpose?

Jellyfin.slnx
```diff
@@ -0,0 +1,77 @@
...
+    </Project>
+    <Project Path="src/Jellyfin.MediaEncoding.Keyframes/Jellyfin.MediaEncoding.Keyframes.csproj">
+      <BuildType Project="Debug" />
```
MarkCiliaVincenti, line 18: Is this on purpose?

MarkCiliaVincenti:
Not quite sure why CodeQL analysis is broken.

gnattu:
I’m sorry but I don’t understand the benefit of doing such migration, at least not for now. To me, this just breaks toolchain compatibility for no obvious reason. And you already noticed that our ci is broken after this migration because CodeQL does not support slnx at all(yet). We have tons of downstream packagers using different of toolings and I’m sure this change would just make their life harder. I don’t really think we are at correct timing do make such change as the world still needs sln and we have no immediate need for slnx

PR #15816, opened 2025-12-17
Fix artist display order

theguymadmax:
**Changes**
Reconstruct the artist dictionary using the original input order so artists are returned in the same order as stored in database.

**Issues**
Fixes: https://github.com/jellyfin/jellyfin/issues/15283#issuecomment-3663664230


crobibero:
I don't think this is a fully correct fix. Dictionaries don't have deterministic ordering so the order of insert isn't always the order of iteration.

theguymadmax:
I ended up fixing this in DtoService by iterating over the original `hasArtist.Artists` array instead of the dictionary, so the artist order now stays correct.

We don’t need the changes  in the BaseItemRepository anymore, but it’s nice to have it preserve the order there too.

alanpemex77-glitch:
> 
> 
> **Changes**
> Reconstruct the artist dictionary using the original input order so artists are returned in the same order as stored in database.
> 
> 
> **Issues**
> Fixes: https://github.com/jellyfin/jellyfin/issues/15283#issuecomment-3663664230
>

PR #15820, opened 2025-12-17
Fixed Multi Sort in New ActivityManager

bjorntp:
**Changes**
Fixed Multi Sort by adding condition that sets `THEN BY` to the queries following the first one instead of `ORDER BY` on every one as `ORDER BY` overwrites.
Added MaxDate endpoint to allow for range selector among dates as that might come useful among logs and the MRT-table used in the frontend has support for it.

**Issues**
Multi sort did not work as intended #15583.
jellyfin/jellyfin-web/issues/6221

### DIFF
--- a/Jellyfin.Server.Implementations/Activity/ActivityManager.cs
+++ b/Jellyfin.Server.Implementations/Activity/ActivityManager.cs
@@ -174,15 +174,15 @@ public class ActivityManager : IActivityManager
 
             if (ordered == null)
             {
-            ordered = sortOrder == SortOrder.Ascending
-                ? (ordered ?? query).OrderBy(orderBy)
-                : (ordered ?? query).OrderByDescending(orderBy);
+                ordered = sortOrder == SortOrder.Ascending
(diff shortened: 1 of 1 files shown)


Jellyfin.Server.Implementations/Activity/ActivityManager.cs
```diff
@@ -166,9 +171,19 @@ private IOrderedQueryable<ExpandedActivityLog> ApplyOrdering(IQueryable<Expanded
         foreach (var (sortBy, sortOrder) in sorting)
         {
             var orderBy = MapOrderBy(sortBy);
+
+            if (ordered == null)
+            {
             ordered = sortOrder == SortOrder.Ascending
```
cvium: This appears to have incorrect indentation now?

bjorntp:
> Not a huge fan of how API changes were sneaked into a bugfix PR, but the API changes seem fine.

I get it. I was contemplating whether to split it up or not, but next time I will. I guess there’s no such thing as too small of a PR

PR #15839, opened 2025-12-20
AV1 RKMPP Decoding - still use when possible

blueacid:
**Changes**
This patch: https://github.com/jellyfin/jellyfin/pull/15776 removes hardware decoding on RKMPP for AV1 altogether, due to issues with decoding files where the width is not a multiple of 64.

Therefore, we should only disable HW Decoding in the event that the video isn't a multiple of 64 pixels wide.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -13,6 +13,7 @@
  - [barronpm](https://github.com/barronpm)
  - [bilde2910](https://github.com/bilde2910)
  - [bfayers](https://github.com/bfayers)
+ - [Blueacid](https://github.com/blueacid)
  - [BnMcG](https://github.com/BnMcG)
  - [Bond-009](https://github.com/Bond-009)
  - [brianjmurrell](https://github.com/brianjmurrell)


MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -7039,8 +7039,18 @@ public string GetRkmppVidDecoder(EncodingJobInfo state, EncodingOptions options,
 
                 if (string.Equals(videoStream.Codec, "av1", StringComparison.OrdinalIgnoreCase))
                 {
-                    var accelType = GetHwaccelType(state, options, "av1", bitDepth, hwSurface);
-                    return accelType + ((!string.IsNullOrEmpty(accelType) && isAfbcSupported) ? " -afbc rga" : string.Empty);
```
nyanmisaka, line 7043: It disabled AFBC (ARM Frame Buffer Compression) instead of disabling AV1 hardware decoding.

blueacid: Ah my apologies, sorry!

PR #15837, opened 2025-12-19
Enable listening on IPv6 by default

der-eismann:
**Changes**
This change enables listening on IPv6 addresses by default. Right now the server is only listening on IPv4 and can't be easily configured to change that which makes it hard to use in an IPv6-only environment.

**Issues**
Fixes #13930

PR #15841, opened 2025-12-20
fix crashes on devices that don't support ipv6

JanzenJohn:
**Changes**
This PR fixes crashes on devices that do not support IPv6, needed to realize #15837


JanzenJohn:
Yeah I tested using a Kernel compiled with  CONFIG_IPV6 set to false, I'll add a warning if you want.

JanzenJohn:
I wrote a simple nixos test so you can verify the result without much effort (but perhaps a bit of compute time as this recompiles a kernel without ipv6 support)

i took my pr + enabling IPv6 per default
[jelly.patch](https://github.com/user-attachments/files/24279743/jelly.patch)
and a test config which checks if jellyfin crashed after 5 seconds
[test.txt](https://github.com/user-attachments/files/24279745/test.txt)

`nix-build test.txt`

results in the expected behavior (a notice that IPv6 has been ignored)

my output is below if you don't want to run it yourself

[test.log](https://github.com/user-attachments/files/24279794/test.log)

JanzenJohn:
Is there something left for me to do, to get this included in an upcoming version? Or is it just a matter or time

PR #15869, opened 2025-12-26
Fix race conditions in AudioDb and Omdb provider file system caches

georgeto:
With unfortunate timing multiple async tasks might try to write the
same file concurrently, resulting in the IOException below.

```
[ERR] Error in "The Open Movie Database"
System.IO.IOException: The process cannot access the file '/var/cache/jellyfin/omdb/tt4564123.json' because it is being used by another process.
   at Microsoft.Win32.SafeHandles.SafeFileHandle.Init
   at Microsoft.Win32.SafeHandles.SafeFileHandle.Open
   at System.IO.Strategies.OSFileStreamStrategy..ctor
   at System.IO.FileStream..ctor
   at MediaBrowser.Providers.Plugins.Omdb.OmdbProvider.EnsureItemInfo
   at MediaBrowser.Providers.Plugins.Omdb.OmdbProvider.GetRootObject
   at MediaBrowser.Providers.Plugins.Omdb.OmdbProvider.Fetch[T]
   at MediaBrowser.Providers.Plugins.Omdb.OmdbItemProvider.GetResult[T]
   at MediaBrowser.Providers.Manager.MetadataService`2.ExecuteRemoteProviders
```

### DIFF
--- a/MediaBrowser.Providers/Plugins/AudioDb/AudioDbAlbumProvider.cs
+++ b/MediaBrowser.Providers/Plugins/AudioDb/AudioDbAlbumProvider.cs
@@ -179,12 +179,19 @@ namespace MediaBrowser.Providers.Plugins.AudioDb
             Directory.CreateDirectory(Path.GetDirectoryName(path));
 
             using var response = await _httpClientFactory.CreateClient(NamedClient.Default).GetAsync(url, cancellationToken).ConfigureAwait(false);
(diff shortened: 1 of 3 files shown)


georgeto:
An alternative could be to use a lock-based approach with something like AsyncKeyedLocker, which would actually prevent downloading the resource multiple times.

PR #15886, opened 2025-12-29
Upgrade Swashbuckle and fix OpenAPI spec

Shadowghost:
**Changes**
* Upgrade swashbuckle
* Fix FlagEnum serialization

**Comments**
I tested it with the TS and Kotlin SDK and the generated code is the same (on release branch) except some nullability changes


nielsvanvelzen (approved):
Schema changes seem to be fine. For the Kotlin SDK the only change is some nullability improvements in dictionaries which is good.

PR #15888, opened 2025-12-29
Fix API diff workflow

Shadowghost:
Fixes the API diff output by using a custom GH action I created. Example output can be seen here: https://github.com/Shadowghost/jellyfin/pull/4 and on the associated action artifacts

### DIFF
--- a/.github/workflows/ci-openapi.yml
+++ b/.github/workflows/ci-openapi.yml
@@ -100,7 +100,7 @@ jobs:
 
       - name: Detect OpenAPI changes
         id: openapi-diff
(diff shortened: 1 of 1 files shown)


crobibero:
Note that changes will be triplicated in the changelog due to https://github.com/oasdiff/oasdiff/issues/594

.. Can we remove the json outputter? Is anyone actually using it?

crobibero (approved):

nielsvanvelzen (approved):

cvium (approved):

PR #15905, opened 2025-12-31
Add async overloads to IPeopleRepository

joshwkearney:
## Changes

- Add async overloads of `GetPeopleAsync`, `UpdatePeopleAsync`, and `GetPeopleNamesAsync` to `IPeopleRepository`. 
- Modify the existing implementation of `PeopleRepository` to use async EF Core calls.
- Modify the sync overloads to call the new async implementations with `.GetAwaiter().GetResult()`.

## Motivation

Long term I would like to see Jellyfin support and run well on Postgres, and a big part of that is not blocking on network database calls. Moving the blocking EF Core calls to their async versions is the first step in that direction.

## Notes

- I agree that in general, `.GetAwaiter().GetResult()` should not be used because it blocks the calling thread and can lead to deadlocks. However, in this case the existing sync implementation already blocks the calling thread, so this matches existing behavior. The use of `.ConfigureAwait(false)` will also help to mitigate deadlocks.
- Should I create an issue for this? I wasn't sure what the policy here was for architecture / tech debt improvements that didn't immediately relate to a user issue

## AI Disclosure

No AI was used to make this PR.

## PS

This is my first PR here so feel free to close this / roast me / etc

### DIFF
--- a/Jellyfin.Server.Implementations/Item/PeopleRepository.cs
+++ b/Jellyfin.Server.Implementations/Item/PeopleRepository.cs
@@ -2,10 +2,11 @@ using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
 using Jellyfin.Data.Enums;
 using Jellyfin.Database.Implementations;
 using Jellyfin.Database.Implementations.Entities;
-using Jellyfin.Database.Implementations.Entities.Libraries;
 using Jellyfin.Extensions;
 using MediaBrowser.Controller.Entities;
 using MediaBrowser.Controller.Persistence;
@@ -30,7 +31,7 @@ public class PeopleRepository(IDbContextFactory<JellyfinDbContext> dbProvider, I
     private readonly IDbContextFactory<JellyfinDbContext> _dbProvider = dbProvider;
 
     /// <inheritdoc/>
-    public IReadOnlyList<PersonInfo> GetPeople(InternalPeopleQuery filter)
+    public async Task<IReadOnlyList<PersonInfo>> GetPeopleAsync(InternalPeopleQuery filter, CancellationToken token = default)
     {
         using var context = _dbProvider.CreateDbContext();
         var dbQuery = TranslateQuery(context.Peoples.AsNoTracking(), context, filter);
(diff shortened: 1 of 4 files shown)


Jellyfin.Server.Implementations/Item/PeopleRepository.cs
```diff
@@ -31,6 +32,12 @@ public class PeopleRepository(IDbContextFactory<JellyfinDbContext> dbProvider, I
...
     /// <inheritdoc/>
     public IReadOnlyList<PersonInfo> GetPeople(InternalPeopleQuery filter)
+    {
+        return GetPeopleAsync(filter, CancellationToken.None).GetAwaiter().GetResult();
+    }
+
+    /// <inheritdoc/>
+    public async Task<IReadOnlyList<PersonInfo>> GetPeopleAsync(InternalPeopleQuery filter, CancellationToken token = default)
     {
         using var context = _dbProvider.CreateDbContext();
```
JPVenson, line 42:
```suggestion
        var dbContext = await _dbProvider.CreateDbContextAsync().ConfigureAwait(false);
        await using (dbContext.ConfigureAwait(false))
        {
```

and everywhere else too

Jellyfin.Server.Implementations/Item/PeopleRepository.cs
```diff
@@ -31,6 +32,12 @@ public class PeopleRepository(IDbContextFactory<JellyfinDbContext> dbProvider, I
 
     /// <inheritdoc/>
     public IReadOnlyList<PersonInfo> GetPeople(InternalPeopleQuery filter)
+    {
+        return GetPeopleAsync(filter, CancellationToken.None).GetAwaiter().GetResult();
```
JPVenson, line 36: this should be avoided at any cost. We have to go up the call chain and use proper async/await to make this a meaningful change.

Jellyfin.Server.Implementations/Item/PeopleRepository.cs
```diff
@@ -53,11 +60,18 @@ public IReadOnlyList<PersonInfo> GetPeople(InternalPeopleQuery filter)
             dbQuery = dbQuery.Take(filter.Limit);
         }
 
-        return dbQuery.AsEnumerable().Select(Map).ToArray();
+        var results = await dbQuery.ToListAsync(token).ConfigureAwait(false);
+        return results.Select(Map).ToList();
```
JPVenson, line 64:
```suggestion
        return await dbQuery.AsAsyncEnumerable().Select(Map).ToArrayAsync().ConfigureAwait(false);
```

lets not overcomplicate things. Also why the changes from Array to list? array was specificly chosen here as the return should be read-only

Jellyfin.Server.Implementations/Item/PeopleRepository.cs
```diff
@@ -84,29 +104,35 @@ public void UpdatePeople(Guid itemId, IReadOnlyList<PersonInfo> people)
         var personKeys = people.Select(e => e.Name + "-" + e.Type).ToArray();
 
         using var context = _dbProvider.CreateDbContext();
-        using var transaction = context.Database.BeginTransaction();
-        var existingPersons = context.Peoples.Select(e => new
+        using var transaction = await context.Database.BeginTransactionAsync(token).ConfigureAwait(false);
+
+        var existingPersons = await context.Peoples.Select(e => new
```
JPVenson, line 109: this is a change in behavior!

JPVenson (changes requested):

Bond-009:
Closing as stale

PR #15906, opened 2025-12-31
Fix infinite metadata pruning loop on rclone/network mounts

### LINKED ISSUE #15904: Library scan causes infinite "File changed, pruning extracted data" loop with rclone network mounts (10.11.x)
### Description of the bug

During library scans on Jellyfin 10.11.x, files on rclone network mounts are repeatedly flagged as "changed" causing Jellyfin to continuously prune and re-extract metadata in an infinite loop. This prevents library scans from completing and causes severe performance degradation.
This appears related to issue #15861 but is specifically about the pruning behavior rather than crashes.

### Reproduction steps

1. Mount a WebDAV backend using rclone
2. Configure Jellyfin library pointing to rclone mount path
3. Start a full library scan
4. Observe logs showing repeated "File changed, pruning extracted data" messages for the same files
5. Scan never completes, CPU drops to 0%, files continuously re-processed

Files are repeatedly detected as "changed" even though mount timestamps are stable. Same files processed dozens of times. Library scans hang or never complete.

Files should be scanned once, metadata extracted, and scan should complete normally as it did in Jellyfin 10.10.3 and earlier.

### Jellyfin Server version

10.11.5

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.5

### Environment

```markdown
- OS: Ubuntu 24 (Docker)
- Jellyfin: Docker (jellyfin/jellyfin:latest)
- Mount: rclone with WebDAV backend
- Storage: Network mount via rclone FUSE
```

### Jellyfin logs

```shell
[17:08:43] [INF] [26] MediaBrowser.Providers.TV.EpisodeMetadataService: File changed, pruning extracted data: /mnt/media/shows/Show Name (2022)/Season 03/Episode.S03E02.1080p.WEB-DL.mkv
[17:08:43] [INF] [26] MediaBrowser.Providers.TV.EpisodeMetadataService: File changed, pruning extracted data: /mnt/media/shows/Show Name (2022)/Season 03/Episode.S03E03.1080p.WEB.mkv
[17:08:43] [INF] [26] MediaBrowser.Providers.TV.EpisodeMetadataService: File changed, pruning extracted data: /mnt/media/shows/Show Name (2022)/Season 03/Episode.S03E08.1080p.WEB.mkv

(These same files repeat continuously throughout the scan)
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

ZeusCraft10:
## Summary

Fixes #15904

During library scans on Jellyfin 10.11.x, files on rclone network mounts were repeatedly flagged as "changed" causing Jellyfin to continuously prune and re-extract metadata in an infinite loop. The scan never completed.

## Root Cause

The original code used `item.HasChanged(info.LastWriteTimeUtc)` which checked if the stored `DateModified` differed from the file system's `LastWriteTimeUtc` by more than 1 second. On rclone/FUSE/network mounts, timestamps can have precision issues or jitter that caused this check to repeatedly return `true`, even when the file hadn't actually changed.

## Solution

Changed the file modification detection to require **BOTH**:
1. Timestamp differs by >1 second
2. File size differs

This prevents false positives from timestamp jitter on network/FUSE file systems while still correctly detecting legitimate file changes (which will always change the file size).

## Changes

- `MediaBrowser.Providers/Manager/MetadataService.cs`: Updated `BeforeSaveInternal` to use combined timestamp+size check
- `CONTRIBUTORS.md`: Added myself per contribution guidelines

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -164,6 +164,7 @@
  - [XVicarious](https://github.com/XVicarious)
  - [YouKnowBlom](https://github.com/YouKnowBlom)
  - [ZachPhelan](https://github.com/ZachPhelan)
+ - [ZeusCraft10](https://github.com/ZeusCraft10)
  - [KristupasSavickas](https://github.com/KristupasSavickas)
  - [Pusta](https://github.com/pusta)
  - [nielsvanvelzen](https://github.com/nielsvanvelzen)
--- a/MediaBrowser.Providers/Manager/MetadataService.cs
+++ b/MediaBrowser.Providers/Manager/MetadataService.cs
@@ -336,26 +336,32 @@ namespace MediaBrowser.Providers.Manager
             if (!string.IsNullOrEmpty(itemPath))
             {
                 var info = FileSystem.GetFileSystemInfo(itemPath);
-                if (info.Exists && item.HasChanged(info.LastWriteTimeUtc))
+                if (info.Exists)
                 {
-                    Logger.LogDebug("File modification time changed from {Then} to {Now}: {Path}", item.DateModified, info.LastWriteTimeUtc, itemPath);
+                    var timestampChanged = Math.Abs((item.DateModified - info.LastWriteTimeUtc).TotalSeconds) > 1;
+                    var sizeChanged = !info.IsDirectory && item.Size.HasValue && item.Size.Value != info.Length;
 
-                    item.DateModified = info.LastWriteTimeUtc;
-                    if (ServerConfigurationManager.GetMetadataConfiguration().UseFileCreationTimeForDateAdded)
+                    if (timestampChanged && sizeChanged)
                     {
-                        if (info.CreationTimeUtc > DateTime.MinValue)
+                        Logger.LogDebug("File modification time changed from {Then} to {Now}: {Path}", item.DateModified, info.LastWriteTimeUtc, itemPath);
+
+                        item.DateModified = info.LastWriteTimeUtc;
+                        if (ServerConfigurationManager.GetMetadataConfiguration().UseFileCreationTimeForDateAdded)
                         {
-                            item.DateCreated = info.CreationTimeUtc;
+                            if (info.CreationTimeUtc > DateTime.MinValue)
+                            {
+                                item.DateCreated = info.CreationTimeUtc;
+                            }
                         }
-                    }
 
-                    if (item is Video video)
-                    {
-                        Logger.LogInformation("File changed, pruning extracted data: {Path}", item.Path);
-                        ExternalDataManager.DeleteExternalItemDataAsync(video, CancellationToken.None).GetAwaiter().GetResult();
-                    }
+                        if (item is Video video)
+                        {
+                            Logger.LogInformation("File changed, pruning extracted data: {Path}", item.Path);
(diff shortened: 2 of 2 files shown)


sjorge:
There is already a PR open dor this https://github.com/jellyfin/jellyfin/pull/14716 I believe there was another one too, but can't find it on mobile.

I believe there were concerns that the file size would not always differ if for example correct a name in a tag if the change was small.

Shadowghost:
We initially had the file size check added to the logic but dropped it because technically you can modify a file without the file size changing.
I know people don't like it when I keep saying this, but I still think that it's not Jellyfin's job to take care of issues caused by your file system of choice if it doesn't concern basic stuff like Database write issues.

ZeusCraft10:
> There is already a PR open dor this #14716 I believe there was another one too, but can't find it on mobile.
> 
> I believe there were concerns that the file size would not always differ if for example correct a name in a tag if the change was small.

The PR does not mention the issue specifically.

beanbaganalytica:
Hey, I just wanted to share that I tested this patch and it actually works! This has been driving me crazy.

**My setup:**
- Jellyfin 10.11.5 (cherry-picked your commit to the release-10.11.z branch since it wasn't there yet)
- rclone mount with `--vfs-cache-mode full`
- Running in Docker

**The problem I was having:**
Library scans would just... never finish. The logs were getting absolutely spammed with "Pruning provider from..." messages for the same files over and over. New content wouldn't show up properly because the scan would get stuck in this infinite loop. It made the whole setup basically unusable.

**After applying your patch:**
No more pruning spam in the logs. Full library scans actually complete now. The fix makes sense too checking both timestamp AND size change instead of just timestamp means rclone's little timestamp variations don't trigger false positives anymore.

This fix is honestly a lifesaver for anyone running Jellyfin with rclone mounts. Would be amazing if this could get merged into the stable 10.11.x branch so people don't have to do custom builds like I did.

Thanks for figuring this out!

ZeusCraft10:
> Hey, I just wanted to share that I tested this patch and it actually works! This has been driving me crazy.
> 
> **My setup:**
> 
> * Jellyfin 10.11.5 (cherry-picked your commit to the release-10.11.z branch since it wasn't there yet)
> * rclone mount with `--vfs-cache-mode full`
> * Running in Docker
> 
> **The problem I was having:** Library scans would just... never finish. The logs were getting absolutely spammed with "Pruning provider from..." messages for the same files over and over. New content wouldn't show up properly because the scan would get stuck in this infinite loop. It made the whole setup basically unusable.
> 
> **After applying your patch:** No more pruning spam in the logs. Full library scans actually complete now. The fix makes sense too checking both timestamp AND size change instead of just timestamp means rclone's little timestamp variations don't trigger false positives anymore.
> 
> This fix is honestly a lifesaver for anyone running Jellyfin with rclone mounts. Would be amazing if this could get merged into the stable 10.11.x branch so people don't have to do custom builds like I did.
> 
> Thanks for figuring this out!

No problem. hopefully it gets pushed.

beanbaganalytica:
Hey all,

Upgraded to 10.11.6 and now seems like I can refresh. Still get the constant pruning errors but it is not breaking anything and the scan completes.

Still would be a good merge.

PR #15921, opened 2026-01-01
Added PlaybackSpeed option to SyncPlay

### LINKED ISSUE #7185: Syncplay with different speed will force back to 1x

- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-web/issues?q=is%3Aissue) _(I've searched it)_.

### Describe the bug

I’m not sure if this is considered a bug or feature request. 

When I use different speed and try to use Syncplay, it will automatically change back to 1x and the other machines wont follow. 

### Reproduction Steps

1. turn on sync play
2. Change playback speed
3. Play the video

### Expected/Actual behaviour

Expect: the change of speed remains unchanged 
Actually: automatically change back to 1x

### Logs

_No response_

### Server version

10.10.7

### Web version

10.10.7

### Build version

10.10.7

### Platform

Windows

### Browser

Chrome 

### Additional information

_No response_

PeachesMLG:
Adds the ability to set playback speed via syncplay, should be backwards compatable with clients that havent implemented this

**Changes**
Adds PlaybackSpeed field on SyncPlay websocket messages
Adds new POST Syncplay/SetPlaybackSpeed method

**Issues**
Fixes https://github.com/jellyfin/jellyfin-web/issues/7185

### DIFF
--- a/Jellyfin.Api/Models/SyncPlayDtos/SetPlaybackSpeedRequestDto.cs
+++ b/Jellyfin.Api/Models/SyncPlayDtos/SetPlaybackSpeedRequestDto.cs
@@ -6,7 +6,7 @@ namespace Jellyfin.Api.Models.SyncPlayDtos;
 public class SetPlaybackSpeedRequestDto
 {
     /// <summary>
-    /// Gets or sets the playback speed.
+    /// Gets or sets the playback speed, should be between 0.25 and 5.
     /// </summary>
     /// <value>The playback speed.</value>
     public double PlaybackSpeed { get; set; } = 1.0;
--- a/Jellyfin.Api/Models/SyncPlayDtos/SetPlaybackSpeedRequestDto.cs
+++ b/Jellyfin.Api/Models/SyncPlayDtos/SetPlaybackSpeedRequestDto.cs
@@ -1,3 +1,5 @@
+using System.ComponentModel.DataAnnotations;
+
 namespace Jellyfin.Api.Models.SyncPlayDtos;
 
 /// <summary>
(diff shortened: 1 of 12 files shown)


Jellyfin.Api/Models/SyncPlayDtos/SetPlaybackSpeedRequestDto.cs
```diff
@@ -0,0 +1,13 @@
...
+/// <summary>
+/// Class SetPlaybackSpeedRequestDto.
+/// </summary>
+public class SetPlaybackSpeedRequestDto
+{
+    /// <summary>
+    /// Gets or sets the playback speed.
+    /// </summary>
+    /// <value>The playback speed.</value>
+    public double PlaybackSpeed { get; set; } = 1.0;
```
nielsvanvelzen, line 15: Could add the min/max flags here so the openapi spec shows the value needs to be between 0.25 and 5

nielsvanvelzen (approved):
API changes look fine (for what that is worth in this awful syncplay API design), just got one small suggestion

Jellyfin.Api/Models/SyncPlayDtos/SetPlaybackSpeedRequestDto.cs
nielsvanvelzen: I actually meant with a range attribute 😅 

https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.rangeattribute

PeachesMLG: ah, I was wondering if the xml docs were even put on the openapi schema, this would make way more sense

entropicdrifter:
Just putting it out there that the PR title has a typo, it says PaybackSpeed not PlaybackSpeed

PeachesMLG:
> Just putting it out there that the PR title has a typo, it says PaybackSpeed not PlaybackSpeed

I would never make a typo. I am extreamly well known for my perfect grammer

Nol22815 (approved):

PR #11689, opened 2024-05-17
Workaround ffmpeg keyframe seeking for external subtitles

gnattu:
We seek to the exact position of the keyframe for direct stream/remuxing, but FFmpeg seeks to the previous keyframe when the exact time is provided as input. To work around this, add a 0.5 second offset to the seeking time so that the correct keyframe is returned for most videos.

This is probably not the most correct way to handle this, but I found that it might be the only way to include this in the 10.9 branch. Other approaches might require either significant changes to the server or an API change.

This does not solve desyncing after transcoding for some video files, but I haven't found a good way to fix those yet due to the complexity of different encoders' outputs. Maybe adding an option to force burn-in subtitles during transcoding could be a good workaround? Currently, client settings enforce burn-in in all situations when set.

**Changes**

- add 0.5s offset to seek position for direct stream/remuxing

**Issues**

Fixes #11567

Partially Fixes #2547

PR #1659, opened 2019-08-22
Fix segment_time_delta value for ffmpeg 4.1

cvium:
**Changes**
See https://github.com/jellyfin/jellyfin/pull/1434#issuecomment-522082156

Since @Verequies seems busy, I am opening this PR. He is free to update his PR (#1434) and we'll close this one 🙂 

**Issues**
Fixes #555

PR #15926, opened 2026-01-02
Refactor HLS transcode seeking

### LINKED ISSUE #15845: noaccurate_seek breaks subtitle synchronisation
### Description of the bug

Jellyfin [passes `-noaccurate_seek` to FFmpeg when transcoding to HLS](https://github.com/jellyfin/jellyfin/blob/master/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs#L2947). This means that when seeking ahead in a video and causing a new transcode to start (which therefore passes the `-ss` parameter), the HLS segments are not precisely aligned with the video timestamp - the timestamp in the new HLS stream may be a few seconds offset from the timestamp in the original video.

For clients that render their own subtitles based on the video timestamp, this means subtitles are offset, and the offset changes whenever a seek occurs.

I can't find the original reason why this parameter was added. It seems that since it has been added there has been one instance of the parameter being required (#7435), but that PR doesn't actually provide details on how to reproduce the issue it is fixing. On the other hand, there have also been bugs that result from this flag being added, including this subtitle bug, which may be the cause of several issues raised on the jellyfin repos (#7338, https://github.com/jellyfin/jellyfin-web/issues/7413, https://github.com/jellyfin/jellyfin-web/issues/6210, https://github.com/jellyfin/jellyfin-web/issues/6210, https://github.com/jellyfin/jellyfin-web/issues/4346, https://github.com/jellyfin/jellyfin/issues/2547).

Is there any reason to keep this flag? I can see why it needs to be kept in the case of #7435, but the check added in that PR seems to be too large and is causing quite a few problems elsewhere. I was going to submit a PR that simply narrows the check to only add `-noaccurate_seek` when dealing with the issue encountered in that PR, but without repro instructions, I can't.

I'd want to remove the flag entirely myself but I don't want to risk breaking jellyfin for others. I have patched my server and the subtitle sync issues I was having on iOS, web, and Android have all disappeared since.

### Reproduction steps

1. Play any media with subtitles rendered by the client and HLS transcoding
2. Seek beyond the buffered range
3. Subtitles are desynced.

Subtitles are desynced.

More accurately, the transcoded video is not aligned with the player timestamp, which causes subtitle desync (the transcoded video is the one with the wrong timestamp, not the player).

Subtitles should be synced.

### Jellyfin Server version

10.11.5

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.5

### Environment

```markdown
- OS: Ubuntu 24.04.3 LTS (containerized)
- Linux Kernel: Linux 6.17.8-arch1-1
- Virtualization: Docker
- Clients: Jellyin Web on Chrome, Jellyfin for Android, Jellyfin for iOS (iPad)
- Browser: Chrome
- FFmpeg Version: 7.1.3-Jellyfin
- Playback Method: Transcode
- Hardware Acceleration: QSV, vaapi, none
- CPU Model: Intel(R) Celeron(R) N4120 CPU @ 1.10GHz
- GPU Model: Intel UHD Graphics 600
- Plugins: None
- Reverse Proxy: nginx
- Base URL: (none)
- Networking: Docker bridge
- Jellyfin Data Storage & Filesystem: ext4 SSD
- Media Storage & Filesystem: btrfs HDD
- External Integrations: none
```

### Jellyfin logs

```shell
None
```

### FFmpeg logs

```shell
None
```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

abitofevrything:
**Changes**
This PR refactors the way in which `-noaccurate_seek` is passed to FFmpeg when transcoding HLS, to ensure the transcoded video has more accurate timestamps.

The major change made here is that `-noaccurate_seek` is no longer passed when performing a transcode (as opposed to a stream copy/remux). In this case, the seeking can be performed accurately since the stream is not copied, so we should do so in order to ensure the transcoded video timestamps correspond to the requested timestamp (important for client side subtitle rendering and SyncPlay).

#15845 also discusses the potential issues of "partial transcodes" where the audio stream is transcoded but the video stream is copied. It turns out that such a situation is a non-issue, since in this case the segment starts are aligned with key frames in the source media, so non-accurate seeking has the same result as accurate seeking (thanks to #11689).

This PR also removes the `BreakOnNonKeyFrames` option that has been meaningless since #1659 (and is now no longer used). Clients should therefore stop sending this option.

**Issues**
Closes #15845

### DIFF
--- a/MediaBrowser.Model/Configuration/EncodingOptions.cs
+++ b/MediaBrowser.Model/Configuration/EncodingOptions.cs
@@ -1,6 +1,7 @@
 #pragma warning disable CA1819 // XML serialization handles collections improperly, so we need to use arrays
 
 #nullable disable
+using System.ComponentModel;
 using MediaBrowser.Model.Entities;
 
 namespace MediaBrowser.Model.Configuration;
@@ -306,5 +307,6 @@ public class EncodingOptions
     /// <summary>
     /// Gets or sets the method used for audio seeking in HLS.
     /// </summary>
+    [DefaultValue(HlsAudioSeekStrategy.DisableAccurateSeek)]
     public HlsAudioSeekStrategy HlsAudioSeekStrategy { get; set; }
 }


abitofevrything:
I see that I have made breaking changes to the API. How should I address this?

Should I leave it as-is, or add the fields back with a deprecation marker? Maybe something else?

gnattu:
You don’t remove parameters generally. That option can be a no-op but remove it from server will break clients which is not something we want.

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -2932,17 +2932,15 @@ public string GetFastSeekCommandLineParameter(EncodingJobInfo state, EncodingOpt
...
-                        && state.TranscodingType != TranscodingJobType.Progressive
-                        && !state.EnableBreakOnNonKeyFrames(outputVideoCodec)
-                        && (state.BaseRequest.StartTimeTicks ?? 0) > 0)
+                    // If we are remuxing, then the copied stream cannot be seeked accurately (it will seek to the nearest
+                    // keyframe). If we are using fMP4, then force all other streams to use the same inaccurate seeking to
+                    // avoid A/V sync issues which cause playback issues on some devices.
+                    // In any case, when remuxing the segment start times correspond to key frames in the source stream,
+                    // so this option shouldn't change the seeked point that much.
+                    // Important: make sure not to use it with wtv because it breaks seeking
+                    if (isHlsRemuxing
```
gnattu: This is not safe enough because we also have video transcode + audio copy. I think we should check in the reverse way, adding noaccurate_seek for most cases and only skip when both audio and video are not copy

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -2932,17 +2932,15 @@ public string GetFastSeekCommandLineParameter(EncodingJobInfo state, EncodingOpt
...
+                        && !string.Equals(state.InputContainer, "wtv", StringComparison.OrdinalIgnoreCase)
+                        && string.Equals(segmentContainer, "mp4", StringComparison.OrdinalIgnoreCase))
```
gnattu: use segmentFormat, not segmentContainer

Jellyfin.Api/Controllers/AudioController.cs
```diff
@@ -107,7 +106,6 @@ [FromQuery] [RegularExpression(EncodingHelper.ContainerValidationRegex)] string?
         [FromQuery] bool? enableAutoStreamCopy,
         [FromQuery] bool? allowVideoStreamCopy,
         [FromQuery] bool? allowAudioStreamCopy,
-        [FromQuery] bool? breakOnNonKeyFrames,
```
gnattu, line 110: Removing query parameters could potentially break clients and we generally don't directly do that. I *think* the server would not throw errors even if clients still passes this though so removing this *maybe* fine for 10.12 where breaking API changes are allowed.

nielsvanvelzen: Removing query parameters is not breaking change for clients

PeachesMLG: Should have probably been in a separate pr though since (I dont believe) this is relevant to the actual bug fix

abitofevrything: To be honest, this PR could have been titled "Remove breakOnNonKeyFrames", the changes regarding `-noaccurate_seek` can just be seen as refactoring the logic around the removal of this option. I didn't do that since the original motive for the change was the `-noaccurate_seek` option.

PeachesMLG:
FYI, I fixed the failing github actions in a different pr, if you merge from main they should pass
(Id assume we cant merge this until its green)

abitofevrything:
> You don’t remove parameters generally. That option can be a no-op but remove it from server will break clients which is not something we want.

The parameter removal isn't an issue. The breaking change is on removing the field from a couple of classes returned by the API.

I'll add them back with a deprecation warning.

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
abitofevrything: segmentFormat is equal to segmentContainer, apart from in the case where `segmentContainer` is `null`. This never happens for HLS, and even if it does, the current check has the same result (checking `!= "ts"` vs `== "mp4"`).

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
sjorge: I think an option somewhere to also transcode audio if the video is being transcoded regardless if a audio copy would suffice, generally the processing overhead for for audio seems lowish. I wouldn't make it the default though as it seems people still run jellyfin on way underpowered systems like RPI4's (would they even handle video transcoding?)

Although I'm not sure how much more complicated this will make things as I think the whole transcode code is already rather complicated :(

abitofevrything: I've just implemented this. I haven't yet figured out how to expose the option to clients, so I'll need a big longer to work on that (any pointers would be appreciated).

I set the option to be to transcode audio by default. I feel that if a system is already transcoding video, then it can also transcode audio (especially if it's just a decode-encode round trip). The option doesn't trigger if video is being copied and not transcoded, as in that case the HLS segments are aligned with keyframes anyway so non accurate seeking is sufficient.

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -2557,6 +2557,13 @@ public bool CanStreamCopyAudio(EncodingJobInfo state, MediaStream audioStream, I
                 return false;
             }
 
+            if (state.TranscodingType is TranscodingJobType.Hls
+                && !IsCopyCodec(state.OutputVideoCodec)
+                && options.AudioSyncType is AudioSyncType.TranscodeAudio)
```
gnattu: Doing the check here is not the best way IMO because this helper will be called by pure audio endopints like music playback as well and this option will essentially diable audio remux for music as well.

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
gnattu: What I think would be better is, instead of leaking the encoding options in the `CanStreamCopyAudio` method, we do such check in the places calls it where we can have better insights of the transcoding job, like whether it is for video or for audio, and if for video, we can simply disable `AllowAudioStreamCopy` to force transcode audio.

MediaBrowser.Model/Configuration/EncodingOptions.cs
```diff
@@ -60,6 +60,7 @@ public EncodingOptions()
         SubtitleExtractionTimeoutMinutes = 30;
         AllowOnDemandMetadataBasedKeyframeExtractionForExtensions = ["mkv"];
         HardwareDecodingCodecs = ["h264", "vc1"];
+        AudioSyncType = AudioSyncType.TranscodeAudio;
```
gnattu: Please don't make new behavior the default. Introducing surprising for users and users flood in github issue with "unnecessary transcoding" bug report is not fun

MediaBrowser.Model/Configuration/AudioSyncType.cs
```diff
@@ -0,0 +1,23 @@
+namespace MediaBrowser.Model.Configuration
+{
+    /// <summary>
+    /// An enum representing the options to synchronise the audio and video streams
+    /// in HLS.
+    /// </summary>
+    public enum AudioSyncType
```
gnattu: `AudioSyncType` looks like a too generic name as we are only applies this when video stream is transcoded and that behavior is better reflected in the enum type name.

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -2557,6 +2557,13 @@ public bool CanStreamCopyAudio(EncodingJobInfo state, MediaStream audioStream, I
                 return false;
             }
 
+            if (state.TranscodingType is TranscodingJobType.Hls
+                && !IsCopyCodec(state.OutputVideoCodec)
```
gnattu: Having this depending on video codec for a helper for audio stream is not something I personally think the best way. How about moving this check to the caller, and just disable `request.AllowAudioStreamCopy` if the check result is to force transcoding?

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -7102,7 +7108,7 @@ public void TryStreamCopy(EncodingJobInfo state)
             }
 
             if (state.AudioStream is not null
-                && CanStreamCopyAudio(state, state.AudioStream, state.SupportedAudioCodecs))
+                && CanStreamCopyAudio(state, options, state.AudioStream, state.SupportedAudioCodecs))
```
gnattu: So basically we move the check here where we have better insights of the transcoding job. First we should be sure that there is indeed an video stream and the video is indeed transcoding here.

MediaBrowser.Model/Configuration/EncodingOptions.cs
sjorge: Still think it should be a server side toggle somewhere for the admin/owner to decide if extra resources should be spend or not.

No client is going to set this option as they all strongly seem to favor direct play if possible. The only exception here seems to be web and it’s option to always burn subtitles for example.

gnattu: > Still think it should be a server side toggle somewhere for the admin/owner to decide if extra resources should be spend or not.

This is already a server side toggle from the way it is implemented

abitofevrything: > Please don't make new behavior the default.

I don't see why not. Technically this is a bug fix; the current behaviour is buggy. It's also not as if it's a hugely expensive to perform this - from my testing transcoding audio results in a < 1% increase in transcode time when also transcoding video, and this PR doesn't prevent audio from being copied if video is not transcoded.

Also, if we don't make it the default _now_, then I'd want to make it the default _later_ anyway. Otherwise we stay stuck with buggy behaviour as the default.

gnattu: > I don't see why not. 

Make decisions for other users is one of the worst thing we could do. Transcoding is not only about processing power, it is also about information lost during the transcoding, especially when you are transcoding from a lossless format or patented format (like dolby digital). Maybe you don't care about such use cases but a lot of our users do. You see inaccurate timing as a bug and other users sees unnecessary transcoding as a bug. It is *that* simple.

abitofevrything: Fair enough. I don't think lossless formats would lose information in a decode-encode round trip, but lossy formats could. I've switched the default.

MediaBrowser.Model/Dlna/TranscodingProfile.cs
```diff
@@ -139,11 +138,13 @@ public TranscodingProfile(TranscodingProfile other)
...
-    /// Gets or sets a value indicating whether breaking the video stream on non-keyframes is supported.
+    /// Gets a value indicating whether breaking the video stream on non-keyframes is supported.
     /// </summary>
-    [DefaultValue(false)]
+    /// <remarks>
+    /// Deprecated.
+    /// </remarks>
     [XmlAttribute("breakOnNonKeyFrames")]
-    public bool BreakOnNonKeyFrames { get; set; }
+    public bool BreakOnNonKeyFrames => false;
```
nielsvanvelzen: This is **not** how to deprecate a property in a response model. You need to make it nullable and add the obsolete and default value attributes (the latter was already there and you removed it). Then in a future release it can be removed entirely.

Consdering it already has a default (falsy) value it should be safe to remove entirely already.

abitofevrything: Thanks, wasn't sure how this is usually handled in C#. I won't remove it just yet since it hasn't been marked as obsolete, and some 3p clients may depend on the field being sent to properly deserialise, but I have marked it as obsolete.

nielsvanvelzen: A response field that is nullable or has a default value is safe to remove as SDK's built on top of our OpenAPI specification have a fallback value.

Also removing the query parameters was completely fine, I didn't comment about those at all.

abitofevrything: However removing the field will remove it from the spec, which means clients that just generate the bindings from that spec will no longer have the field either.

Should I leave it as is or just completely remove the parameters and this field?

nielsvanvelzen: Keeping it is fine for now, as long as it is properly marked as deprecated:
- Add obsolete attribute
- Make nullable
- Set default value

abitofevrything: I think it's done properly now.

Jellyfin.Api/Controllers/UniversalAudioController.cs
```diff
@@ -114,7 +114,7 @@ [FromQuery] [RegularExpression(EncodingHelper.ContainerValidationRegex)] string?
         [FromQuery] int? maxAudioBitDepth,
         [FromQuery] bool? enableRemoteMedia,
         [FromQuery] bool enableAudioVbrEncoding = true,
-        [FromQuery] bool breakOnNonKeyFrames = false,
+        [FromQuery, ParameterObsolete] bool breakOnNonKeyFrames = false,
```
nielsvanvelzen: Why did you keep only this one??

abitofevrything: This parameter wasn't nullable, unlike the others, so my find+replace didn't catch it. Removed now.

MediaBrowser.Model/Configuration/EncodingOptions.cs
```diff
@@ -301,4 +302,9 @@ public EncodingOptions()
     /// Gets or sets the file extensions on-demand metadata based keyframe extraction is enabled for.
     /// </summary>
     public string[] AllowOnDemandMetadataBasedKeyframeExtractionForExtensions { get; set; }
+
+    /// <summary>
+    /// Gets or sets the method used for audio seeking in HLS.
+    /// </summary>
+    public HlsAudioSeekStrategy HlsAudioSeekStrategy { get; set; }
```
nielsvanvelzen, line 311: A default value should be set for the OpenAPI spec

abitofevrything:
What's the process for getting this merged? AFAICS most PRs are merged by the last approving reviewer, but it's been a week with no activity on this one.

Issue #15917, opened 2025-12-31
Shows with many episodes never load on browser or take extremely large amounts of time to load.


ButtercupKitten:
### Description of the bug

After upgrading to the latest version from 10.10.07 and following all steps correctly I am unable to load seasons for shows with many episodes. Detective Conan has over 1000+ episodes and trying to load the entry never loads on desktop and takes 5+ minutes on a client like Roku. It used to only take a few seconds to load the entire show, now it takes an extremely long time . I have done two full scans of all the files, and it still does not fix the issue. I have even tried many different fixes floating around forums to no avail. Other shows have major issues loading seasons and episodes, but Detective Conan is the one that outright never works.

### Reproduction steps

upgrade from 10.10.07 to 10.11.05 and try to load a show with many episodes 1,000+ and watch as it does not load in the browser on the pc the server is being ran on. 

### What is the current _bug_ behavior?

The seasons and episodes never load.

### What is the expected _correct_ behavior?

Everything loads in a few seconds like it used to.

### Jellyfin Server version

10.11.5

### Specify the build version

10.11.5

### Environment

```markdown
- OS: Windows 10
- Linux Kernel: N/A
- Virtualization: Windows system tray
- Clients: Roku
- Browser: Firefox
- FFmpeg Version:Default with Jellyfin install
- Playback Method: Direct
- Hardware Acceleration: Nvenc
- CPU Model: I7 7600K
- GPU Model: Rtx 3060
- Plugins: Air Times, AniDB, AniList, AnimeThemes, AniSearch, Artwork, AudioDB, Chapter Segments Provider, Fanart, Intro Skipper, Kitsu, MusicBrainz, OMDb, Open Subtitles, SkinManager, Studio Images, Subtitle Extract, Theme Songs, TheTVDB, TMDb, TMDb Box Sets, TVmaze
- Reverse Proxy: none
- Base URL: none
- Networking: local
- Jellyfin Data Storage & Filesystem:
- Media Storage & Filesystem: Local on SSD
- External Integrations:none
```

### Jellyfin logs

```shell
[...]

PR #15927, opened 2026-01-02
Fix EF Core Cartesian Explosion and SQLite migration bugs

ZeusCraft10:
## Problem Description
This PR addresses Issue #15917 where loading TV shows with many episodes caused an EF Core Cartesian Explosion due to the default QuerySplittingBehavior.SingleQuery. It also resolves the SQLite schema error 
o such column: u.MaxParentalAgeRating by fixing a migration bug and refreshing stale designer snapshots.

## Proposed Changes
- **BaseItemRepository**: Removed global .AsSingleQuery() and implemented selective .AsSplitQuery() in ApplyNavigations when 2+ collections are loaded.
- **Migrations**: Fixed bug in 20250326065026_AddInheritedParentalRatingSubValue.cs (wrong column dropped in Down()).
- **Snapshots**: Corrected stale designer snapshots in 20250327101120_AddKeyframeData.Designer.cs and 20250327171413_AddHdr10PlusFlag.Designer.cs.

## Verification
- Verified build success for SQLite provider and Implementations project.
- Logic confirms selective splitting only for complex multi-collection queries, preserving performance for simple ones.

### DIFF
--- a/Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
+++ b/Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
@@ -408,31 +408,45 @@ public sealed class BaseItemRepository
         return dbQuery;
     }
 
-    private static IQueryable<BaseItemEntity> ApplyNavigations(IQueryable<BaseItemEntity> dbQuery, InternalItemsQuery filter)
+    private IQueryable<BaseItemEntity> ApplyNavigations(IQueryable<BaseItemEntity> dbQuery, InternalItemsQuery filter)
     {
+        var collectionIncludes = 0;
         if (filter.TrailerTypes.Length > 0 || filter.IncludeItemTypes.Contains(BaseItemKind.Trailer))
         {
             dbQuery = dbQuery.Include(e => e.TrailerTypes);
+            collectionIncludes++;
         }
 
         if (filter.DtoOptions.ContainsField(ItemFields.ProviderIds))
         {
             dbQuery = dbQuery.Include(e => e.Provider);
+            collectionIncludes++;
         }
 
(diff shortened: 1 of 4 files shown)


JPVenson:
this is the wrong approach to the issue and the code is somewhat nonsensical. This has the sour taste of "works better on my pc" but its not like sql is build to work with joins. The Cartesian explosion isn't really a problem for database engines those days and with our dataset but for sqlite specifically its much more adventurous to do everything in one go as the overhead of multiple seperate queries is also something you have to consider as this query gets executed multiple times per sec on scan. So this might be faster for some sure but it will be slower for most in the end according to my testing as sqlite is much worse in creating connections (yes even with pooling) then to return the query in one go. 

We are somewhat at the limit of what we can do on a base query standpoint and the better way forward is to rework the schema itself then this.

ZeusCraft10:
Thanks for the detailed explanation @JPVenson! You're completely right - I was focused on fixing the warning instead of understanding why it's happening in the first place.

After digging into this more, I see what you mean:
- Split queries add more overhead than they save (especially with SQLite)
- The real issue is that we're loading way too much data at once (Images, UserData, ProviderIds, etc. for 1000+ episodes)
- The database can handle the joins just fine - it's just a lot of data to pull

I'm going to close this PR since it doesn't actually solve the problem. I'll comment on the original issue with some thoughts on the schema side of things and see what direction makes sense.

Thanks for taking the time to explain this - really appreciate the guidance!

ZeusCraft10:
Closing this PR as the split query approach doesn't address the root cause. Will follow up on the original issue with schema redesign analysis.

PR #15942, opened 2026-01-04
Add plugin based backup support

JPVenson:
This code allows plugins to interact with the backup system to provide their own dataset that will be backed up into the general JF backup archive. Plugins must decorate the implementation of the `IPlugin` interface with the new `PluginBackupAttribute<TDataLoader>` attribute to provide a seperated class for those purposes. 

The `TDataLoader` generic must point to a class that implements the `IPluginBackupService`. This service has the explicit contract that it must be able to run fully standalone without any assurances of the plugin being initialised as the backup procedure runs before the server has been initialised. 

Implementation description:
The Jellyfin backup archive already contains a manifest that stores the metadata of its contents. This pr extends this manifest and adds a list of plugins that are contained in the backup and a mapping of the contents a plugin has requested to backup. The plugins are strictly restricted in the types of data they can backup or how they are able to interact with the backup archive itself which ensures compatability in the future as well as for plugins between versions as this is also the goal of the backup system as a whole.


cvium:
This feels overly complicated. I still don't get what's wrong with the HA model of just assigning a folder to every plugin and then backing that up.

JPVenson:
i checked that too. Firstly the path the plugin uses for data can be overwritten by that plugin so its non deterministic at restore time which makes this unusable without changes to __all__ plugins and the IPlugin interface. Next as described in the description, this aims for the same level of support for plugins as we have for our own data in terms of support and feature level which can enable plugins to have more functions and new ways to interact with JF as a whole while still putting strict guard rails on. Plugins that for example save trickplay that is stored next to media or external databases can be supported which a simple folder backup would never been able to.

cvium:
- Plugin devs shouldn't overwrite the path if they want to allow easy backups. I don't think that's an unfair ask.
- When we store stuff next to media, I expect the admin to handle backups (I imagine that is part of the motivation to store it next to media in the first place).
- Why wouldn't plugins use our DB? Or a file-based one at least.

I get what you're going for, but without a real example this becomes too hypothetical to me and seems like over-engineering.

JPVenson:
> Plugin devs shouldn't overwrite the path if they want to allow easy backups. I don't think that's an unfair ask.

Well we say that _now_ but its never been an issue and always possible.

> When we store stuff next to media, I expect the admin to handle backups (I imagine that is part of the motivation to store it next to media in the first place).

Currently thats our standpoint on this yes, but i see no reason for not wanting that included if somebody wants to write a plugin for it.

> Why wouldn't plugins use our DB? Or a file-based one at least.

Because they cannot. We 100% do not want plugins meddling with our DB and we currently have no official way for them to do so, that is a story for another day and even file-based ones currently face the same issue. Today with sqlite, while it technically works, its totally inefficient to just store the DB file and that DB file may contain data that is absolutly not backup worthy but how would you handle external Dbs in the very near future?

jumoog:
> We 100% do not want plugins meddling with our DB and we currently have no official way for them to do so, that is a story for another day and even file-based ones currently face the same issue.

Yes, a plugin shouldn't interfere with the Jellyfin database. We've had enough of dealing with locks and other problems. We don't need any more. 

Currently, there are no specifications detailing what plugins should or shouldn't do. The only recommendation I'm aware of is that plugins shouldn't modify index.html. In the absence of any specifications or best practices, this is the best way forward.

jhparker88:
>     Why wouldn't plugins use our DB? Or a file-based one at least.
>  I get what you're going for, but without a real example this becomes too hypothetical to me and seems like over-engineering.

https://github.com/Shadowghost/jellyfin-plugin-segment-recognition is a specific example for how a plugin writes its own DB without interfering with the Jellyfin DB.

PR #15944, opened 2026-01-04
Add support for TTML subtitles

CatNoir2006:
**Changes**

- Added support for .ttml subtitles by recognizing the extension and mapping its MIME type.
- Implemented manual resolution for external TTML subtitle files to ensure they are correctly detected as media streams.
- Added a conversion path from TTML to SRT within the SubtitleEncoder to support subtitle burning via FFmpeg.
- Improved the subtitle parser to handle alternate file extensions for various subtitle formats.
- Updated the encoding helper to utilize the effective subtitle path when applying filters and added a unit test for TTML recognition.

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -1831,16 +1831,18 @@ namespace MediaBrowser.Controller.MediaEncoding
                 ":fontsdir='{0}'",
                 _mediaEncoder.EscapeSubtitleFilterPath(fontPath));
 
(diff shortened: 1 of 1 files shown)


CalmSeoul:
I'm assuming this would not be relevant to ttml being used for lyrics for songs on jellyfin?

CatNoir2006:
@CalmSeoul Im not sure if I understand it correctly but. (My English is not that good) But I dont think that TTML is just at any part of the Jellyfin Projekt as I Only found `public partial class TtmlWriter : ISubtitleWriter` but I do not know what exactly Its used for as it does not allow reading ttml.

PR #15954, opened 2026-01-06
Fix Book collections speed issues

### LINKED ISSUE #15378: book collection completly broken !
### Description of the bug

With version 10.10.7, the entire book section works perfectly, even with almost 2000 ebooks! Since the second-to-last update, my library loads, but the loading wheel (of death) just keeps spinning endlessly... After the last two updates, I must have performed 10 complete scans of my libraries, but it hasn't changed anything. I haven't found anything in the logs that could shed any light on the problem.

### Reproduction steps

1.install lastest jellyfin server's version
2/ make ebooks library
3/ scan and wait
4/ go to ebooks library and wait...

yes, it's a bug because, it works fine before !!

At least display the ebooks!!

### Jellyfin Server version

10.10.0+

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.2

### Environment

```markdown
- OS: kubuntu 24.04.3
- Linux Kernel: 6.8.0-86-generic
- Virtualization: no
- Clients: web page
- Browser: all
- FFmpeg Version: lastest
- Playback Method: I can read anything !
- Hardware Acceleration: no
- GPU Model: Nvidia
- Plugins: yes
- Reverse Proxy: no
- Base URL:
- Networking:
- Jellyfin Data Storage: local
- Media Storage: local
- External Integrations:
```

### Jellyfin logs

```shell
no log !
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

IDisposable:
Fix Books collection access that is really slow/broken when an entire media library is huge.

**Changes**
The root cause is that there's no handling in `GetUserItems` for books, nor is the `BaseItemKind`s being set for folders with `ContentType.books`

- Added `static` method `GetBaseItemKindsForCollectionType` in _DtoExtensions.cs_ (moved from _ContentFolderImageProvider.cs_ to be shared)
- Updated `GetItemsWithImages` in _ContentFolderImageProvider.cs_ to use new method in _DtoExtensions.cs_
- Added `AudioBook` to `GetRepresentativeItemTypes` for `CollectionType.books` for consistency with other uses
- Added `GetBooks` to `GetUserItems` for `viewType == CollectionType.books` , which gets `BaseItemKind.Book` and `BaseItemKind.AudioBook` items.
- Associate `Person` to `Book` and `AudioBook` for related items in _DtoService.cs_ since `SupportsPeople is true` in  `MediaBrowser.Controller.Entities`
- Cleaned up missing null checks and used new collection expressions.

**Issues**
Fixes #15378

### DIFF
--- a/Emby.Server.Implementations/Images/CollectionFolderImageProvider.cs
+++ b/Emby.Server.Implementations/Images/CollectionFolderImageProvider.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using Jellyfin.Api.Extensions;
 using Jellyfin.Data.Enums;
 using Jellyfin.Database.Implementations.Enums;
 using MediaBrowser.Common.Configuration;
@@ -14,7 +15,6 @@ using MediaBrowser.Controller.Entities;
 using MediaBrowser.Controller.Providers;
 using MediaBrowser.Model.Entities;
 using MediaBrowser.Model.IO;
-using MediaBrowser.Model.Querying;
 
 namespace Emby.Server.Implementations.Images
 {
@@ -28,38 +28,7 @@ namespace Emby.Server.Implementations.Images
         {
             var view = (CollectionFolder)item;
             var viewType = view.CollectionType;
-
-            BaseItemKind[] includeItemTypes;
-
-            switch (viewType)
-            {
-                case CollectionType.movies:
-                    includeItemTypes = new[] { BaseItemKind.Movie };
-                    break;
-                case CollectionType.tvshows:
-                    includeItemTypes = new[] { BaseItemKind.Series };
-                    break;
-                case CollectionType.music:
-                    includeItemTypes = new[] { BaseItemKind.MusicArtist };  // Music albums usually don't have dedicated backdrops, so use artist instead
-                    break;
-                case CollectionType.musicvideos:
-                    includeItemTypes = new[] { BaseItemKind.MusicVideo };
-                    break;
-                case CollectionType.books:
-                    includeItemTypes = new[] { BaseItemKind.Book, BaseItemKind.AudioBook };
-                    break;
-                case CollectionType.boxsets:
-                    includeItemTypes = new[] { BaseItemKind.BoxSet };
-                    break;
-                case CollectionType.homevideos:
-                case CollectionType.photos:
-                    includeItemTypes = new[] { BaseItemKind.Video, BaseItemKind.Photo };
-                    break;
-                default:
-                    includeItemTypes = new[] { BaseItemKind.Video, BaseItemKind.Audio, BaseItemKind.Photo, BaseItemKind.Movie, BaseItemKind.Series };
-                    break;
-            }
-
+            var includeItemTypes = DtoExtensions.GetBaseItemKindsForCollectionType(viewType);
             var recursive = viewType != CollectionType.playlists;
 
             return view.GetItemList(new InternalItemsQuery
@@ -67,12 +36,9 @@ namespace Emby.Server.Implementations.Images
                 CollapseBoxSetItems = false,
                 Recursive = recursive,
                 DtoOptions = new DtoOptions(false),
-                ImageTypes = new[] { ImageType.Primary },
+                ImageTypes = [ImageType.Primary],
                 Limit = 8,
-                OrderBy = new[]
-                {
-                    (ItemSortBy.Random, SortOrder.Ascending)
-                },
+                OrderBy = [(ItemSortBy.Random, SortOrder.Ascending)],
                 IncludeItemTypes = includeItemTypes
             });
         }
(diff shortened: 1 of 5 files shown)


Emby.Server.Implementations/Images/CollectionFolderImageProvider.cs
```diff
@@ -28,38 +29,7 @@ protected override IReadOnlyList<BaseItem> GetItemsWithImages(BaseItem item)
...
-                case CollectionType.homevideos:
-                case CollectionType.photos:
-                    includeItemTypes = new[] { BaseItemKind.Video, BaseItemKind.Photo };
-                    break;
-                default:
-                    includeItemTypes = new[] { BaseItemKind.Video, BaseItemKind.Audio, BaseItemKind.Photo, BaseItemKind.Movie, BaseItemKind.Series };
-                    break;
-            }
-
+            var includeItemTypes = ItemsController.GetBaseItemKindsForCollectionType(viewType);
```
IDisposable: Is there a better place for this to be moved to?

Jellyfin.Api/Controllers/ItemsController.cs
```diff
@@ -288,12 +290,13 @@ public ActionResult<QueryResult<BaseItemDto>> GetItems(
         if (folder is IHasCollectionType hasCollectionType)
         {
             collectionType = hasCollectionType.CollectionType;
+            includeItemTypes = [.. includeItemTypes.Union(DtoExtensions.GetBaseItemKindsForCollectionType(collectionType))];
```
IDisposable, line 319: Open question here: Should we **not** get the defaults if someone actually specified item types?

With this PR's changes, we'll _add_ the expected base types to the inclusions. For most collection types, this will likely not add anything, but for Books and AudioBooks, this is essential to keep the query from attempting to read **all** items.

Jellyfin.Api/Controllers/LibraryController.cs
```diff
@@ -990,7 +990,7 @@ private static string[] GetRepresentativeItemTypes(CollectionType? contentType)
             CollectionType.playlists => new[] { "Playlist" },
             CollectionType.movies => new[] { "Movie" },
             CollectionType.tvshows => new[] { "Series", "Season", "Episode" },
-            CollectionType.books => new[] { "Book" },
+            CollectionType.books => new[] { "Book", "AudioBook" },
```
IDisposable, line 1024: Everywhere else, Books and AudioBooks are treated the same, so added for consistency

Jellyfin.Api/Extensions/DtoExtensions.cs
```diff
@@ -9,6 +10,35 @@ namespace Jellyfin.Api.Extensions;
 /// </summary>
 public static class DtoExtensions
 {
+    /// <summary>
```
IDisposable, line 13: This is essentially what was originally in _CollectionFolderImageProvider.cs_ with null protection and collection expression syntax.

MediaBrowser.Controller/Entities/UserViewBuilder.cs
```diff
@@ -63,6 +63,9 @@ public QueryResult<BaseItem> GetUserItems(Folder queryParent, Folder displayPare
                 case CollectionType.folders:
                     return GetResult(_libraryManager.GetUserRootFolder().GetChildren(user, true), query);
 
+                case CollectionType.books:
```
IDisposable, line 64: This ensures we query books correctly (include parent, honor user-restrictions, specify item types).

dkanada:
Do audiobook items currently exist under the music library? I know the media type exists but I've never used them in any libraries.

I just tried creating a collection with two books and it shows up fine on the latest version so the title might need some explanation - what exactly is empty? You referenced #15378 which is about performance and seems unrelated.

IDisposable:
> what exactly is empty... is about performance

When a client doesn't specify the `includedItemTypes` the result set is built from entire corpus of item types (e.g. EVERYTHING), so on a larger library it essentially never returns (or throws). This ensures that when we know the base item type from the context (books/audiobooks/etc) we always filter on those types.

IDisposable:
(rebased)

IDisposable:
Rebased on master (again)

Shadowghost (approved):

Bond-009 (approved):

Jellyfin.Api/Controllers/ItemsController.cs
theguymadmax: This breaks the folder library view (dashboard->libraries-display).

IDisposable: I can certainly pull that change.

theguymadmax: Submitted a PR to fix.

theguymadmax: I'm still seeing other regressions from this. I'll open up a PR to revert this change.

IDisposable: That sucks, so sorry. What's broken, I can surgical strike.

theguymadmax: See: https://github.com/jellyfin/jellyfin/pull/17348

PR #15959, opened 2026-01-06
Add TrueHD and DTS codes string for HLS

nyanmisaka:
Since there are no Apple devices that support these two audio codecs in their HLS player, fourccs are taken and crosschecked from gpac/mp4box, androidx/media, libavcodec and [ts_103285v010401p.pdf](https://www.etsi.org/deliver/etsi_ts/103200_103299/103285/01.04.01_60/ts_103285v010401p.pdf).

**Changes**
- Add TrueHD and DTS codes string for HLS

### DIFF
--- a/Jellyfin.Api/Helpers/HlsCodecStringHelpers.cs
+++ b/Jellyfin.Api/Helpers/HlsCodecStringHelpers.cs
@@ -42,6 +42,11 @@ public static class HlsCodecStringHelpers
     public const string OPUS = "Opus";
 
     /// <summary>
+    /// Codec name for TRUEHD.
+    /// </summary>
+    public const string TRUEHD = "mlpa";
+
+    /// <summary>
     /// Gets a MP3 codec string.
     /// </summary>
     /// <returns>MP3 codec string.</returns>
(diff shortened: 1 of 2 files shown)


Jellyfin.Api/Helpers/HlsCodecStringHelpers.cs
```diff
@@ -59,7 +64,7 @@ public static string GetAACString(string? profile)
     {
         StringBuilder result = new StringBuilder("mp4a", 9);
 
-        if (string.Equals(profile, "HE", StringComparison.OrdinalIgnoreCase))
+        if (string.Equals(profile, "HE-AAC", StringComparison.OrdinalIgnoreCase))
```
homer-simpson, line 67: Why you add "-AAC"? Is the profile "HE" not a valid one?

nyanmisaka: Nope
- https://github.com/jellyfin/jellyfin-web/blob/fe5f2af4b41baa68acc0cdf68c76033bfba2dd7f/src/scripts/browserDeviceProfile.js#L956
- https://github.com/jellyfin/jellyfin-ffmpeg/blob/fedbcdeee255372d80229df4343a8d71763bbb6b/libavcodec/profiles.c#L29

homer-simpson: Thanks - looks good

PR #15970, opened 2026-01-07
Implement similarity providers

Shadowghost:
Similarity and recommendations are worse due to DB performance issues in 10.11.
This PR extracts similarity calculation to a provider based approach:
* Local providers based on genre/tags as it is implemented right now
* TMDB providers for TV series and Movies
* ListenBrainz provider for artists based on [Labs API](https://labs.api.listenbrainz.org/similar-artists)


MediaBrowser.Providers/Manager/ProviderManager.cs
```diff
@@ -137,17 +138,234 @@ public void AddParts(
...
+        {
+            var requestedLimit = limit ?? 50;
+
+            var localProviders = _similarItemsProviders.OfType<ILocalSimilarItemsProvider<T>>().Cast<ISimilarItemsProvider>();
+            var remoteProviders = _similarItemsProviders.OfType<IRemoteSimilarItemsProvider<T>>().Cast<ISimilarItemsProvider>();
+            var matchingProviders = localProviders.Concat(remoteProviders).ToList();
+            var typeOptions = libraryOptions?.GetTypeOptions(typeof(T).Name);
+
+            // Filter by enabled providers if TypeOptions is configured for this item type
+            if (typeOptions is not null)
```
theguymadmax:
```suggestion
            if (typeOptions?.SimilarItemProviders?.Length > 0)
```

Shadowghost: There are also other issues I need to fix after the refactor. Will take are of this too.

MediaBrowser.Controller/Providers/IProviderManager.cs
```diff
@@ -101,13 +103,43 @@ public interface IProviderManager
...
-            IEnumerable<IExternalUrlProvider> externalUrlProviders);
+            IEnumerable<IExternalUrlProvider> externalUrlProviders,
+            IEnumerable<ISimilarItemsProvider> similarItemsProviders);
+
+        /// <summary>
+        /// Gets the similar items providers for a specific item type.
+        /// </summary>
+        /// <typeparam name="T">The item type.</typeparam>
+        /// <returns>The list of similar items providers for that type.</returns>
+        List<ISimilarItemsProvider> GetSimilarItemsProviders<T>()
```
JPVenson:
```suggestion
        IReadOnlyList<ISimilarItemsProvider> GetSimilarItemsProviders<T>()
```

Emby.Server.Implementations/Library/SimilarItems/SimilarItemsManager.cs
```diff
@@ -0,0 +1,400 @@
...
+        int? limit,
+        LibraryOptions? libraryOptions,
+        CancellationToken cancellationToken)
+    {
+        ArgumentNullException.ThrowIfNull(item);
+        ArgumentNullException.ThrowIfNull(excludeArtistIds);
+
+        var method = typeof(SimilarItemsManager)
+            .GetMethod(nameof(GetSimilarItemsInternalAsync), System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance)!
+            .MakeGenericMethod(item.GetType());
```
JPVenson: you should either precompile that list based on possible implementations of BaseItem or use a ConcurrentDictionary for reflection cache

Emby.Server.Implementations/Library/SimilarItems/SimilarItemsManager.cs
```diff
@@ -0,0 +1,400 @@
...
+                    }
+
+                    continue;
+                }
+
+                var responses = new List<SimilarItemProviderResponse>();
+                TimeSpan? cacheDuration = null;
+                int? nextPage = 0;
+
+                while (allResults.Count < requestedLimit && nextPage is not null && !cancellationToken.IsCancellationRequested)
```
JPVenson: Why do we need to read the pages until we have our desired result set should't the provider give us a conclusive set of pagesize we desire? Maybe the GetSimilarItemsAsync method can return an IAsyncEnumerable instead and if thats paged the implementation should handle that

Shadowghost: Because we need to resolve the returned items, remote providers only give us external IDs, but we need to check if those exist in our internal DB. If they do not, we need to fetch more pages.

Emby.Server.Implementations/Library/SimilarItems/SimilarItemsManager.cs
```diff
@@ -0,0 +1,405 @@
...
+        DtoOptions dtoOptions,
+        int? limit,
+        LibraryOptions? libraryOptions,
+        CancellationToken cancellationToken)
+    {
+        ArgumentNullException.ThrowIfNull(item);
+        ArgumentNullException.ThrowIfNull(excludeArtistIds);
+
+        var itemType = item.GetType();
+        var method = _genericMethodCache.GetOrAdd(itemType, static type => _getSimilarItemsInternalMethod.MakeGenericMethod(type));
```
JPVenson:
```suggestion
        var method = _genericMethodCache.GetOrAdd(itemType, static type => _getSimilarItemsInternalMethod.MakeGenericMethod(type).CreateDelegate(typeof(Func<...>)));
```

Shadowghost: Can't be used here because the method has a type parameter

theguymadmax:
Error when starting up this branch:
<details>
  <summary>Log Error</summary>

```
15:07:03.087] [FTL] [9] Main: Error while starting server
System.AggregateException: Some services are not able to be constructed (Error while validating the service descriptor 'ServiceType: Jellyfin.Api.Controllers.SearchController Lifetime: Transient ImplementationType: Jellyfin.Api.Controllers.SearchController': Unable to resolve service for type 'MediaBrowser.Controller.Library.ISearchEngine' while attempting to activate 'Jellyfin.Api.Controllers.SearchController'.)
 ---> System.InvalidOperationException: Error while validating the service descriptor 'ServiceType: Jellyfin.Api.Controllers.SearchController Lifetime: Transient ImplementationType: Jellyfin.Api.Controllers.SearchController': Unable to resolve service for type 'MediaBrowser.Controller.Library.ISearchEngine' while attempting to activate 'Jellyfin.Api.Controllers.SearchController'.
 ---> System.InvalidOperationException: Unable to resolve service for type 'MediaBrowser.Controller.Library.ISearchEngine' while attempting to activate 'Jellyfin.Api.Controllers.SearchController'.
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteFactory.CreateArgumentCallSites(ServiceIdentifier serviceIdentifier, Type implementationType, CallSiteChain callSiteChain, ParameterInfo[] parameters, Boolean throwIfCallSiteNotFound)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteFactory.CreateConstructorCallSite(ResultCache lifetime, ServiceIdentifier serviceIdentifier, Type implementationType, CallSiteChain callSiteChain)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteFactory.TryCreateExact(ServiceDescriptor descriptor, ServiceIdentifier serviceIdentifier, CallSiteChain callSiteChain, Int32 slot)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteFactory.GetCallSite(ServiceDescriptor serviceDescriptor, CallSiteChain callSiteChain)
   at Microsoft.Extensions.DependencyInjection.ServiceProvider.ValidateService(ServiceDescriptor descriptor)
   at Microsoft.Extensions.DependencyInjection.ServiceProvider.ValidateService(ServiceDescriptor descriptor)
   at Microsoft.Extensions.DependencyInjection.ServiceProvider..ctor(ICollection`1 serviceDescriptors, ServiceProviderOptions options)
   at Microsoft.Extensions.DependencyInjection.ServiceProvider..ctor(ICollection`1 serviceDescriptors, ServiceProviderOptions options)
   at Microsoft.Extensions.DependencyInjection.ServiceCollectionContainerBuilderExtensions.BuildServiceProvider(IServiceCollection services, ServiceProviderOptions options)
   at Microsoft.Extensions.Hosting.HostBuilder.InitializeServiceProvider()
   at Microsoft.Extensions.Hosting.HostBuilder.Build()
   at Jellyfin.Server.Program.StartServer(IServerApplicationPaths appPaths, StartupOptions options, IConfiguration startupConfig) in C:\Gits\jellyfin\Jellyfin.Server\Program.cs:line 168
```

</details>

JPVenson (approved):

enter-a-random-username:
This will not affect `more like this` or?
- Added the PR to the build + FE selection of a provider (activated local for now, sounded like fastest for initial test)
- If it impacts `more like this`, the issue is maybe a provider adds a generic tag like Animation. Perhaps drop genres that is common (e.g. 50% of all entries) like other searches do.
(Added comment here, because I saw no impact after running library scan and so on)

Shadowghost:
It should affect "more like this", but the sql provider should mirror what we currently do.

enter-a-random-username:
Works with TMD - I guess local had too much overlap, so it seems just random

Emby.Server.Implementations/Library/SimilarItems/SimilarItemsManager.cs
```diff
@@ -0,0 +1,423 @@
...
+namespace Emby.Server.Implementations.Library.SimilarItems;
+
+/// <summary>
+/// Manages similar items providers and orchestrates similar items operations.
+/// </summary>
+public class SimilarItemsManager : ISimilarItemsManager
+{
+    private static readonly ConcurrentDictionary<Type, MethodInfo> _genericMethodCache = new();
+    private static readonly MethodInfo _getSimilarItemsInternalMethod = typeof(SimilarItemsManager)
+        .GetMethod(nameof(GetSimilarItemsInternalAsync), BindingFlags.NonPublic | BindingFlags.Instance)!;
```
cvium: Why reflection?

Emby.Server.Implementations/Library/SimilarItems/SeriesSimilarItemsProvider.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// Initializes a new instance of the <see cref="SeriesSimilarItemsProvider"/> class.
+    /// </summary>
+    /// <param name="libraryManager">The library manager.</param>
+    public SeriesSimilarItemsProvider(ILibraryManager libraryManager)
+    {
+        _libraryManager = libraryManager;
+    }
+
+    /// <inheritdoc/>
+    public string Name => "Local Genre/Tag";
```
cvium, line 31: What is the significance of Name and why is the same for all of them?

Emby.Server.Implementations/Library/SimilarItems/SimilarItemsManager.cs
```diff
@@ -0,0 +1,423 @@
...
+                Directory.CreateDirectory(directory);
+            }
+
+            var cache = new SimilarItemsCache
+            {
+                References = references,
+                ExpiresAt = DateTime.UtcNow.Add(cacheDuration)
+            };
+
+            var stream = File.Create(cachePath);
```
cvium, line 374: Why cache it on the file system and not in a database or in-memory?

MediaBrowser.Providers/Plugins/ListenBrainz/Api/ListenBrainzLabsClient.cs
```diff
@@ -0,0 +1,104 @@
...
+    private void EnforceRateLimit(double rateLimitSeconds)
+    {
+        lock (_rateLimitLock)
+        {
+            var timeSinceLastRequest = DateTime.UtcNow - _lastRequestTime;
+            var requiredDelay = TimeSpan.FromSeconds(rateLimitSeconds) - timeSinceLastRequest;
+
+            if (requiredDelay > TimeSpan.Zero)
+            {
+                Thread.Sleep(requiredDelay);
```
cvium: Thread.Sleep inside an async method is bad. You're locking a thread pool thread. Also why not use Polly for this?

Emby.Server.Implementations/Library/SimilarItems/SimilarItemsManager.cs
Shadowghost: Didn't want to create a table for kind of provider specific data since this can get quite large. But I don't have a strong preference here.

Emby.Server.Implementations/Library/SimilarItems/SeriesSimilarItemsProvider.cs
Shadowghost: That's the SQL based provider - currently it is majorly based on Genre/Tag but I'm happy to rename it to something more usfule. I doN't like the name either

MediaBrowser.Providers/Plugins/ListenBrainz/Api/ListenBrainzLabsClient.cs
Shadowghost: No reason, but I think a SemaphoreSlim should be good enough

Emby.Server.Implementations/Library/SimilarItems/SimilarItemsManager.cs
cvium: How large and how long is it cached?

Emby.Server.Implementations/Library/SimilarItems/SimilarItemsManager.cs
```diff
@@ -0,0 +1,406 @@
...
+                        User = user,
+                        Limit = requestedLimit - allResults.Count,
+                        DtoOptions = dtoOptions,
+                        ExcludeItemIds = [.. excludeIds],
+                        ExcludeArtistIds = excludeArtistIds
+                    };
+
+                    // Collect references in batches and resolve against local library.
+                    // Stop fetching once we have enough resolved local items.
+                    const int BatchSize = 20;
```
cvium, line 174: Why this batch size?

Shadowghost: To not over-fetch. But with all the perf improvements we can likely increase it if we want to

Emby.Server.Implementations/Library/SimilarItems/SimilarItemsManager.cs
Shadowghost: Large in the sense of items pulled from tmdb, it heavily depends on how many items you have in your local DB. The less items you have the more we need to fetch. When we stop is decided by the limit set by the caller or 50 if none is set,
CacheDuration is decided by provider, currently I've configure it with 14 days for Listenbrainz and 7 Days for Tmdb.

cvium: It's worth considering

cvium (approved):

Emby.Server.Implementations/Library/SimilarItems/SimilarItemsManager.cs
Shadowghost: What is worth considerin? Raiisng the cache duration or the size?

crobibero:
Since the goal is to move external plugins to standalone plugins, I think the ListenBrainz implementation should be moved to a plugin. TMDb is still bundled with server so it should stay for now

Emby.Server.Implementations/Library/SimilarItems/SimilarItemsManager.cs
cvium: Using the database

Emby.Server.Implementations/Library/SimilarItems/SimilarItemsManager.cs
JPVenson: its a generic method we need to call without having the generic known at runtime.

crobibero (approved):

Bond-009 (approved):

PR #15971, opened 2026-01-07
Add bitrate ladder support for automatic media source selection

hmac2222:
When multiple versions of media exist (4K HDR, 1080p SDR, 720p), Jellyfin selects the highest resolution first—even when it requires transcoding. This PR adds intelligent source selection that prioritizes direct-playable versions and enables true adaptive bitrate streaming between pre-encoded files.

**Changes**
- Add `MediaSourceSelectionMode` enum with `HighestQuality`, `PreferDirectPlay` (default), and `NetworkAware` modes
- Add `OptimalMediaSourceId` to `PlaybackInfoResponse` for client auto-selection
- Prioritize direct-playable sources over higher-quality sources that require transcoding
- Push unplayable sources (e.g., needs transcoding but user lacks permission) to end of selection
- Extend HLS master playlist to include variants for all pre-encoded alternate versions, enabling mid-stream quality switching without server transcoding

**Issues**
Addresses feature request for "bitrate ladder" support similar to Netflix, where clients automatically receive the most appropriate version based on device capabilities and network conditions. See https://features.jellyfin.org/posts/3680/bitrate-ladder-support

### DIFF
--- a/Jellyfin.Api/Controllers/MediaInfoController.cs
+++ b/Jellyfin.Api/Controllers/MediaInfoController.cs
@@ -213,7 +213,11 @@ public class MediaInfoController : BaseJellyfinApiController
                     Request.HttpContext.GetNormalizedRemoteIP());
             }
 
-            _mediaInfoHelper.SortMediaSources(info, maxStreamingBitrate);
+            _mediaInfoHelper.SortMediaSources(
+                info,
+                maxStreamingBitrate,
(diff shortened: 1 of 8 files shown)


sjorge:
How will this deal with minor differences in runtime length? Say between a BR and DVD ripped media.

Switching between those will probably have the playback end up in slightly different spots. 

I suppose the same issue exists now too when manually switching.

PR #15974, opened 2026-01-08
fix: artist metadata not saving properly

Almamu:
**Changes**
Partially reverted commit 5a6d9180fed81a30cb91ef3fed30176cd4402116 which introduced a regression in artist creation. 

**Issues**
Fixes metadata edition from #15283

**Some questions regarding this metadata**
The way artist creation is handled is really confusing. As it is right now, artists are added to the BaseItems table when an Item containing unknown Artists is returned by the API. This was done for both Album Artists and Artists, but on 5a6d9180fed81a30cb91ef3fed30176cd4402116 the code that was handling that was changed to prevent duplication. Judging by the changes made in that commit, this change wasn't actually necessary and should come back, although in the future I feel like the artist should be added to the database on creation instead of when it's returned by any API endpoint? Maybe I can work on that next if no one is working on it?

During the investigation of this issue I also came across three ways of saving Artist and Album Artists, which sounds a bit backwards to me. One is in the BaseItems table, another in each BaseItems record (there's an Artist and AlbumArtist field that contains a list of them separated by ```|```) and another table ItemValues (joined to ItemValuesMap) that has the artist info again, maybe we should look into consolidating this into just one?

### DIFF
--- a/Emby.Server.Implementations/Dto/DtoService.cs
+++ b/Emby.Server.Implementations/Dto/DtoService.cs
@@ -1031,26 +1031,6 @@ namespace Emby.Server.Implementations.Dto
             {
                 dto.Artists = hasArtist.Artists;
 
-                // var artistItems = _libraryManager.GetArtists(new InternalItemsQuery
-                // {
-                //    EnableTotalRecordCount = false,
-                //    ItemIds = new[] { item.Id.ToString("N", CultureInfo.InvariantCulture) }
-                // });
-
-                // dto.ArtistItems = artistItems.Items
-                //    .Select(i =>
-                //    {
-                //        var artist = i.Item1;
-                //        return new NameIdPair
-                //        {
-                //            Name = artist.Name,
-                //            Id = artist.Id.ToString("N", CultureInfo.InvariantCulture)
-                //        };
-                //    })
-                //    .ToList();
-
-                // Include artists that are not in the database yet, e.g., just added via metadata editor
-                // var foundArtists = artistItems.Items.Select(i => i.Item1.Name).ToList();
                 var artistsLookup = _libraryManager.GetArtists([.. hasArtist.Artists.Where(e => !string.IsNullOrWhiteSpace(e))]);
 
                 dto.ArtistItems = hasArtist.Artists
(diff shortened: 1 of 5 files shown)


Emby.Server.Implementations/Dto/DtoService.cs
```diff
@@ -1051,16 +1051,21 @@ private void AttachBasicFields(BaseItemDto dto, BaseItem item, BaseItem? owner,
 
                 // Include artists that are not in the database yet, e.g., just added via metadata editor
                 // var foundArtists = artistItems.Items.Select(i => i.Item1.Name).ToList();
-                dto.ArtistItems = _libraryManager.GetArtists([.. hasArtist.Artists.Where(e => !string.IsNullOrWhiteSpace(e))])
```
JPVenson: IIRC we introduced the batch operation for performance reasons so i dont see how this fixes anything

JPVenson (changes requested):

Emby.Server.Implementations/Dto/DtoService.cs
Almamu: GetArtists does not create any record in the database, but GetArtist does. In fact, GetArtist is still used a few lines below for AlbumArtists.

Almamu: @JPVenson just to confirm, do you still think this should be changed? Should the same be applied to AlbumArtists a few lines below? If so, should the artist creation be moved to item saving instead?

JPVenson: yea this isn't the right place and this implicit artist creation sucks. Also the batch loading is important for performance so it would be best to do that when we save the item instead

Almamu: I've changed the logic to work only on item save, so this item should be solved.

Jellyfin.Api/Controllers/ItemsController.cs
```diff
@@ -459,17 +459,9 @@ public ActionResult<QueryResult<BaseItemDto>> GetItems(
...
-                    try
-                    {
-                        return _libraryManager.GetArtist(i, new DtoOptions(false));
-                    }
-                    catch
-                    {
-                        return null;
-                    }
-                }).Where(i => i is not null).Select(i => i!.Id).ToArray();
+                query.ArtistIds = _libraryManager.GetArtists(artists)
```
Almamu, line 462: This was creating new artists in the database for each search if they did not exist.

Almamu:
@JPVenson pinging for visibility.

Almamu:
I've fixed the merge conflict that was caused by a recent merge and cleaned up the branch so it's easier to review.

Emby.Server.Implementations/Dto/DtoService.cs
```diff
@@ -1067,24 +1047,6 @@ private void AttachBasicFields(BaseItemDto dto, BaseItem item, BaseItem? owner,
...
-                //    {
-                //        var artist = i.Item1;
-                //        return new NameIdPair
-                //        {
-                //            Name = artist.Name,
-                //            Id = artist.Id.ToString("N", CultureInfo.InvariantCulture)
-                //        };
-                //    })
-                //    .ToList();
-
```
JPVenson, line 1087: if you gonna remove that, remove it dont comment it out

Almamu: You're replying to the actual removal of those comments, they were left there by someone else before.

Bond-009 (approved):

PR #15977, opened 2026-01-08
Fix duplicate paths in PhysicalLocationsList

### LINKED ISSUE #15966: Collections library takes 5 seconds to load after upgrading to 10.11 - duplicate paths in PhysicalLocationsList
After upgrading from 10.10.7 to 10.11 (Docker Compose on Debian), my Collections library takes about 5 seconds to load. It was instant before the upgrade. I have 355 manually created collections. Once loaded, everything works fine. The issue persists on the current version 10.11.5.

I investigated the SQLite database and found that the root and Explorer (my Collections library) entries in the BaseItems table contain duplicate paths in their PhysicalLocationsList JSON field:

**root entry:**
```json
"PhysicalLocationsList":["/config/data/root","/home/user/medias/series","/home/user/medias/docs","/config/data/data/collections","/home/user/medias/films"]
```

**Explorer (Collections library) entry:**
```json
"PhysicalLocationsList":["/config/data/root/default/Collections","/config/data/data/collections"]
```

The path /config/data/data/collections appears in both entries, and the Explorer entry has two paths pointing to essentially the same content. This duplication seems to have been introduced during the 10.11 migration.

I initially had warnings in my logs like:
```
[WRN] MediaBrowser.Controller.Entities.BaseItem: Unable to find linked item at path /config/data/data/collections/[Collection Name]
[WRN] MediaBrowser.Controller.Entities.BaseItem: Unable to find linked item at path /config/data/metadata/People/[Person Name]
```

These were caused by collections containing People and other collections as "linked children" with hardcoded absolute paths instead of %AppDataPath% or %MetadataPath% variables. I removed these linked items through the UI and the warnings stopped, but the 5-second delay persists.

My current theory is that the duplicate paths in PhysicalLocationsList cause Jellyfin to scan the collections folder twice on every page load.

Has anyone encountered this? Is there a safe way to clean up these duplicate paths, or should this be addressed in a future migration fix?

ZeusCraft10:
**Changes**

Saw the issue report about Collections taking ~5 seconds to load after upgrading to 10.11. I was able to replicate it and traced the problem to duplicate paths in PhysicalLocationsList - the same folders were being scanned multiple times during library loads.

The fix adds a .Distinct() call when setting PhysicalLocationsList in both CollectionFolder and AggregateFolder. Also updated the SequenceEqual() check for _requiresRefresh to use the same case-insensitive comparer, so the refresh detection and deduplication logic are consistent. Existing databases will self-correct on the next library refresh, so no migration is needed.

**Issues**
Fixes #15966

### DIFF
--- a/MediaBrowser.Controller/Entities/AggregateFolder.cs
+++ b/MediaBrowser.Controller/Entities/AggregateFolder.cs
@@ -141,7 +141,7 @@ namespace MediaBrowser.Controller.Entities
                 args.FileSystemChildren = files;
             }
 
-            _requiresRefresh = _requiresRefresh || !args.PhysicalLocations.SequenceEqual(PhysicalLocations);
+            _requiresRefresh = _requiresRefresh || !args.PhysicalLocations.SequenceEqual(PhysicalLocations, StringComparer.OrdinalIgnoreCase);
             if (setPhysicalLocations)
             {
                 PhysicalLocationsList = args.PhysicalLocations
--- a/MediaBrowser.Controller/Entities/CollectionFolder.cs
+++ b/MediaBrowser.Controller/Entities/CollectionFolder.cs
@@ -299,7 +299,7 @@ namespace MediaBrowser.Controller.Entities
                 args.FileSystemChildren = files;
             }
 
-            _requiresRefresh = _requiresRefresh || !args.PhysicalLocations.SequenceEqual(PhysicalLocations);
+            _requiresRefresh = _requiresRefresh || !args.PhysicalLocations.SequenceEqual(PhysicalLocations, StringComparer.OrdinalIgnoreCase);
 
             if (setPhysicalLocations)
             {


Debouelesmorts:
Hi @ZeusCraft10, I'm the one who opened issue #15966. I was curious why you closed this PR - did you find an issue with the approach, or is there another reason? I've been investigating this problem and your fix looked like it addressed the root cause.

ZeusCraft10:
> Hi @ZeusCraft10, I'm the one who opened issue #15966. I was curious why you closed this PR - did you find an issue with the approach, or is there another reason? I've been investigating this problem and your fix looked like it addressed the root cause.

Hi there @Debouelesmorts. I orignally wrote the PR, and then found some issues with the code, but I wrote up a new PR with the better changes and have already submitted it.

Debouelesmorts:
Thanks for the quick reply @ZeusCraft10 ! I'll keep an eye on the new PR.

PR #15979, opened 2026-01-08
Fix duplicate paths in PhysicalLocationsList causing slow library loads

### LINKED ISSUE #15966: Collections library takes 5 seconds to load after upgrading to 10.11 - duplicate paths in PhysicalLocationsList

ZeusCraft10:
**Changes**

Saw #15966 about Collections libraries taking ~5 seconds to load after upgrading to 10.11. I was able to reproduce the issue and found that duplicate paths were being stored in PhysicalLocationsList, causing the same folders to get scanned multiple times.

The fix deduplicates paths when storing and comparing PhysicalLocationsList in both CollectionFolder and AggregateFolder. Also made the comparison logic use case-insensitive matching throughout so paths like \/Media/Movies\ and \/media/movies\ are treated as the same on case-insensitive filesystems.

Existing databases should fix themselves on the next library refresh - no migration needed.

**Issues**
Fixes #15966

### DIFF
--- a/MediaBrowser.Controller/Entities/AggregateFolder.cs
+++ b/MediaBrowser.Controller/Entities/AggregateFolder.cs
@@ -93,11 +93,15 @@ namespace MediaBrowser.Controller.Entities
 
             if (!changed)
             {
-                var locations = PhysicalLocations;
+                var locations = PhysicalLocations
+                    .Distinct(StringComparer.OrdinalIgnoreCase)
+                    .ToArray();
 
-                var newLocations = CreateResolveArgs(new DirectoryService(FileSystem), false).PhysicalLocations;
+                var newLocations = CreateResolveArgs(new DirectoryService(FileSystem), false).PhysicalLocations
+                    .Distinct(StringComparer.OrdinalIgnoreCase)
+                    .ToArray();
 
-                if (!locations.SequenceEqual(newLocations))
+                if (!locations.SequenceEqual(newLocations, StringComparer.OrdinalIgnoreCase))
                 {
                     changed = true;
                 }
@@ -141,10 +145,16 @@ namespace MediaBrowser.Controller.Entities
                 args.FileSystemChildren = files;
             }
 
-            _requiresRefresh = _requiresRefresh || !args.PhysicalLocations.SequenceEqual(PhysicalLocations);
+            var physicalLocations = args.PhysicalLocations
+                .Distinct(StringComparer.OrdinalIgnoreCase)
+                .ToArray();
+            var currentLocations = PhysicalLocations
+                .Distinct(StringComparer.OrdinalIgnoreCase)
+                .ToArray();
+            _requiresRefresh = _requiresRefresh || !physicalLocations.SequenceEqual(currentLocations, StringComparer.OrdinalIgnoreCase);
             if (setPhysicalLocations)
             {
-                PhysicalLocationsList = args.PhysicalLocations;
+                PhysicalLocationsList = physicalLocations;
             }
 
             return args;
--- a/MediaBrowser.Controller/Entities/CollectionFolder.cs
+++ b/MediaBrowser.Controller/Entities/CollectionFolder.cs
@@ -189,11 +189,15 @@ namespace MediaBrowser.Controller.Entities
 
             if (!changed)
             {
-                var locations = PhysicalLocations;
+                var locations = PhysicalLocations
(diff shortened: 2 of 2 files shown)


Debouelesmorts:
This fix should resolve an issue impacting Collections usage since 10.11.0. Hoping it gets merged soon in upcoming releases.

MediaBrowser.Controller/Entities/AggregateFolder.cs
```diff
@@ -93,11 +93,15 @@ public override bool RequiresRefresh()
 
             if (!changed)
             {
-                var locations = PhysicalLocations;
+                var locations = PhysicalLocations
+                    .Distinct(StringComparer.OrdinalIgnoreCase)
```
dfederm, line 97: Linux is case-sensitve so you should use the right comparer depending on OS.

MediaBrowser.Controller/Entities/AggregateFolder.cs
```diff
@@ -93,11 +93,15 @@ public override bool RequiresRefresh()
...
+                    .ToArray();
 
-                var newLocations = CreateResolveArgs(new DirectoryService(FileSystem), false).PhysicalLocations;
+                var newLocations = CreateResolveArgs(new DirectoryService(FileSystem), false).PhysicalLocations
+                    .Distinct(StringComparer.OrdinalIgnoreCase)
+                    .ToArray();
 
-                if (!locations.SequenceEqual(newLocations))
+                if (!locations.SequenceEqual(newLocations, StringComparer.OrdinalIgnoreCase))
```
dfederm, line 104: Similar comment here.

MediaBrowser.Controller/Entities/AggregateFolder.cs
entropicdrifter: It's even worse than that. Linux supports some case-insensitive filesystems and some of the Linux filesystems have case-insensitive optional modes, like ext4. On top of this, filesystems mounted through possibly multiple layers of abstraction might not allow for a way to detect whether they are case sensitive or not. 

In fact, it looks to still be an open issue for dotnet for them to implement a platform-neutral way to detect if a filesystem is case sensitive.

satmandu: Lots of suggestions around https://stackoverflow.com/a/38559406

I'd also note that you could have a folder hierarchy with a case sensitive filesystem mounted into a subfolder of a case-insensitive filesystem, so I think you need to check every location to determine case-sensitivity, regardless of OS.

Many operating systems can support a simultaneous mix of case-sensitive filesystems. Linux & macOS support both (e.g., both `zfs` and `apfs` can be either.)

theguymadmax:
The assumptions in this PR and the linked issue are incorrect. Closing as invalid.

PR #15983, opened 2026-01-09
Prioritize better matches on search

Shadowghost:
This makes search results with full and prefix matches appear before contains matches to improve results.

**Changes**
Prioritize search results by `exact match > prefix match with word boundary > prefix match > contains`

**Issues**
If you search for the series `FROM`, you'll first get any episodes that includes `from` in their name right now because their `CleanName` is prefixed with numbers. If you have more than 100 of such episodes, you'll not see the series, even though the name is a full match. Same thing happens for movies like `From Dusk Till Dawn`.

### DIFF
--- a/Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
+++ b/Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
@@ -1558,29 +1558,36 @@ public sealed class BaseItemRepository
 
         IOrderedQueryable<BaseItemEntity>? orderedQuery = null;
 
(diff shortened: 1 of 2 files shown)


crobibero (approved):

ServeurpersoCom:
The truly efficient way to access all the titles in a massive collection is simply to sort the results from shortest to longest. This gives access to all the generic (and therefore exact) fragments first, allowing the user to refine their search by lengthening their query string. (I have over 28,000 titles and I still have to use my own local TMDb fuzzy search engine who naturally does this)

PR #15995, opened 2026-01-11
Only scale chapter images based on resolution height to preserve aspect ratio and don't upscale them

MarcoCoreDuo:
Currently, chapter image extraction uses a fixed 16:9 resolution when not set to "Match Source" . This causes images from videos with other aspect ratios to be stretched to 16:9 when a fixed resolution is applied. In addition, if a video’s resolution is lower than the configured chapter image resolution, ffmpeg will upscale the image, which seems unnecessary in this case.

**Changes**
- chapter images are now scaled only based on height
- chapter images are not upscaled if they are lower resolution than the configured resolution

### DIFF
--- a/MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
+++ b/MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
@@ -659,7 +659,7 @@ namespace MediaBrowser.MediaEncoding.Encoder
                 _ => string.Empty
             };
 
-            return string.IsNullOrEmpty(target) ? string.Empty : $",scale=-2:'min(ih,{target})'";
+            return string.IsNullOrEmpty(target) ? string.Empty : $",scale=-2:min(ih\\,{target})";
         }
 
         private async Task<string> ExtractImageInternal(
--- a/MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
+++ b/MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
@@ -646,25 +646,20 @@ namespace MediaBrowser.MediaEncoding.Encoder
 
         private string GetImageResolutionParameter()
         {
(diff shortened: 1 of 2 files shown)


MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
```diff
@@ -646,25 +646,20 @@ private async Task<string> ExtractImage(
...
                 _ => string.Empty
             };
 
-            if (!string.IsNullOrEmpty(imageResolutionParameter))
-            {
-                imageResolutionParameter = " -s " + imageResolutionParameter;
-            }
-
-            return imageResolutionParameter;
+            return string.IsNullOrEmpty(target) ? string.Empty : $",scale=-2:'min(ih,{target})'";
```
dfederm: I believe this deviates from the established pattern in the codebase of escaping commas with backslashes.

Consider:
```cs
return string.IsNullOrEmpty(target) ? string.Empty : $",scale=-2:min(ih\\,{target})";
```

MarcoCoreDuo: You're right. I applied your suggestion, thanks.

Shadowghost:
Does this work as intended if we have portrait videos?

MarcoCoreDuo:
> Does this work as intended if we have portrait videos?

It would scale them to a fixed height but preserve the aspect ratio. However, they would be lower resolution compared to landscape chapter images. I'm not 100% sure what the best solution for this would be.

We could use this as the ffmpeg parameter:
`
$",scale='if(gt(ih\\,iw)\\,min({target}\\,iw)\\,-2)':'if(gt(ih\\,iw)\\,-2\\,min({target}\\,ih))'"
`
This would scale chapter images based on height if they are landscape and based on width if they are portrait.

PR #15996, opened 2026-01-11
Swagger: load xml comments from plugins

jon4hz:
**Changes**

This PR allows plugins to include xml comments from custom rest endpoints in the swagger docs.
With the comments included, you can then test your plugin's endpoints on the swagger page and also generate openapi client implementations for the plugin.

For this to work, the plugin must enable docs generation in the `.csproj` file:
```xml
  <PropertyGroup>
    <GenerateDocumentationFile>true</GenerateDocumentationFile>
  </PropertyGroup>
```

The resulting `My.Plugin.Name.xml` file must also be included in the release zip of the plugin.

### DIFF
--- a/Jellyfin.Server/Extensions/ApiServiceCollectionExtensions.cs
+++ b/Jellyfin.Server/Extensions/ApiServiceCollectionExtensions.cs
@@ -194,8 +194,9 @@ namespace Jellyfin.Server.Extensions
         /// Adds Swagger to the service collection.
         /// </summary>
         /// <param name="serviceCollection">The service collection.</param>
+        /// <param name="pluginAssemblies">An IEnumerable containing all plugin assemblies.</param>
         /// <returns>The updated service collection.</returns>
-        public static IServiceCollection AddJellyfinApiSwagger(this IServiceCollection serviceCollection)
+        public static IServiceCollection AddJellyfinApiSwagger(this IServiceCollection serviceCollection, IEnumerable<Assembly> pluginAssemblies)
         {
             return serviceCollection.AddSwaggerGen(c =>
             {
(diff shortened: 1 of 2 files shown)


Jellyfin.Server/Extensions/ApiServiceCollectionExtensions.cs
```diff
@@ -232,6 +233,26 @@ public static IServiceCollection AddJellyfinApiSwagger(this IServiceCollection s
...
+                // Add xml doc files from plugins.
+                foreach (var pluginAssembly in pluginAssemblies)
+                {
+                    var pluginDir = Path.GetDirectoryName(pluginAssembly.Location);
+                    if (string.IsNullOrEmpty(pluginDir))
+                    {
+                        continue;
+                    }
+
+                    var pluginXmlFiles = Directory.EnumerateFiles(
```
dfederm, line 245: This loads all xml files from plugin directories, not just doc files, which could lead to errors or unexpected behavior.

Additionally, this does not have any error handling. I suggest wrapping `IncludeXmlComments` in a try/catch.

Consider something more like...
```cs
   // Add xml doc files from plugins.
   foreach (var pluginAssembly in pluginAssemblies)
   {
       var assemblyLocation = pluginAssembly.Location;
       if (string.IsNullOrEmpty(assemblyLocation))
       {
           continue;
       }

       // Only load the XML documentation file that matches the assembly name
       var pluginXmlFile = Path.ChangeExtension(assemblyLocation, ".xml");
       if (File.Exists(pluginXmlFile))
       {
           try
           {
               c.IncludeXmlComments(pluginXmlFile);
           }
           catch (Exception)
           {
               // Silently skip malformed documentation files
               // Could add logging if logger is available
           }
       }
   }
```

PR #15997, opened 2026-01-11
Fix typos in XML documentation comments

vignesh1507:
Fixed few typos in the code file, Filteres --> Filters & IP4 --> IPV4

### DIFF
--- a/src/Jellyfin.Networking/Manager/NetworkManager.cs
+++ b/src/Jellyfin.Networking/Manager/NetworkManager.cs
@@ -115,7 +115,7 @@ public class NetworkManager : INetworkManager, IDisposable
     public static string MockNetworkSettings { get; set; } = string.Empty;
 
     /// <summary>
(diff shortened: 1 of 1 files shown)


src/Jellyfin.Networking/Manager/NetworkManager.cs
```diff
@@ -115,7 +115,7 @@ public NetworkManager(IConfigurationManager configurationManager, IConfiguration
     public static string MockNetworkSettings { get; set; } = string.Empty;
 
     /// <summary>
-    /// Gets a value indicating whether IP4 is enabled.
+    /// Gets a value indicating whether IPV4 is enabled.
```
nielsvanvelzen: Should be IPv4 then

Bond-009:
```suggestion
    /// Gets a value indicating whether IPv4 is enabled.
```

PR #16001, opened 2026-01-12
Add language priority list and original language support for images

MarcoCoreDuo:
**Changes**
Implements a new field PreferredImageLanguages in libraryOptions that accepts an array of image language options. These options can be one of three types:
- A specific language
- No language
- The original language of the media

If the array has values, the server will sort images based on this explicit order. It will only fall back to no-language images and not to other languages, to avoid adding images in non-preferred languages. When no preferred image languages are set, it will fall back to the old default behavior.

The PR for jellyfin-web: https://github.com/jellyfin/jellyfin-web/pull/7492

**Features implemented**
https://features.jellyfin.org/posts/2173/image-fetcher-preferred-language
https://features.jellyfin.org/posts/2734/fallback-language-for-posters
https://features.jellyfin.org/posts/3192/logos-should-select-the-correct-language

Closes https://github.com/jellyfin/jellyfin/pull/14078

### DIFF
--- a/MediaBrowser.Model/Extensions/EnumerableExtensions.cs
+++ b/MediaBrowser.Model/Extensions/EnumerableExtensions.cs
@@ -16,7 +16,7 @@ namespace MediaBrowser.Model.Extensions
         /// <param name="remoteImageInfos">The remote image infos.</param>
         /// <param name="requestedLanguage">The requested language for the images.</param>
         /// <returns>The ordered remote image infos.</returns>
-        public static IEnumerable<RemoteImageInfo> OrderByLanguageDescending(this IEnumerable<RemoteImageInfo> remoteImageInfos, string requestedLanguage)
+        private static IEnumerable<RemoteImageInfo> OrderByLanguageDescending(this IEnumerable<RemoteImageInfo> remoteImageInfos, string requestedLanguage)
         {
             if (string.IsNullOrWhiteSpace(requestedLanguage))
             {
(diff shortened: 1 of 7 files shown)


Shadowghost:
@MarcoCoreDuo could you please rebase onto master if you have some time

I'm also not sure if having type string for the language options is a good idea if we want this to be used universally by plugins

MarcoCoreDuo:
> I'm also not sure if having type string for the language options is a good idea if we want this to be used universally by plugins

Would it be a better approach to use an ImageLanguageOption object array instead of the string array? The object could contain the language code itself and an explicit IsNoLanguage boolean property. That would make it easier for plugins, I think.

```csharp
public class ImageLanguageOption
{
    /// <summary>
    /// Gets or sets the language.
    /// </summary>
    /// <value>The language.</value>
    public string? Language { get; set; }

    /// <summary>
    /// Gets or sets a value indicating whether this option represents no language.
    /// </summary>
    /// <value><c>true</c> if is no language; otherwise, <c>false</c>.</value>
    public bool IsNoLanguage { get; set; }
}
```

MediaBrowser.Model/Extensions/EnumerableExtensions.cs
```diff
@@ -33,24 +33,70 @@ public static IEnumerable<RemoteImageInfo> OrderByLanguageDescending(this IEnume
...
+        /// <param name="remoteImageInfos">The remote image infos.</param>
+        /// <param name="requestedMetadataLanguage">The requested metadata language for fallback if no preferred languages are specified.</param>
+        /// <param name="preferredImageLanguages">Array of preferred image languages (e.g., ["en", "de", "fr", "nolang"]). If null or empty, uses requestedLanguage.</param>
+        /// <returns>The ordered remote image infos.</returns>
+        public static IEnumerable<RemoteImageInfo> OrderByLanguageDescending(
+            this IEnumerable<RemoteImageInfo> remoteImageInfos,
+            string requestedMetadataLanguage,
+            string[]? preferredImageLanguages)
+        {
+            // If no preferred image languages are configured, fall back to the original behavior
```
IDisposable: If the specification is that we use the `preferredImageLanguages` and fallback to `requestedMetadataLanguage` then `"en"` (like the old code did) we can just do something like 

```csharp
string[] languages = [requestedMetadataLanguage, .. preferredImageLanguages ?? [], "en"]; 
```

then fall through to the rest of this code... swapping in `languages` for `preferredImageLanguages`... then we can just delete the old (now `private`) version.

MarcoCoreDuo: Thank you for your feedback. I simplified the sorting logic and removed the private method like you suggested

MarcoCoreDuo:
Hello @Lampan-git, I’m currently implementing support for original language, but I’ve encountered an issue with the OriginalLanguage field. When the data source is TMDB, the field contains a single ISO 639-1 language code. However, when the source is OMDB, it returns a comma-separated string of language names featured in a movie, show, or episode.

Is there a reason why these values are not normalized before being assigned to the OriginalLanguage field? This inconsistency makes the field significantly more complex to use.

I’m also unsure whether using the OMDB language data is appropriate here, since it does not represent the original language of the content, but rather a list of all languages featured in a movie, show, or episode. The original language is often the first entry, but this is not guaranteed.

Lampan-git:
I didn't really think it would be used elsewhere, so I didn't really think of normalize at the metadata level. But I agree that it is not optimal. You can see how I handle OMDb in my PR with the commas. I'm not at my computer right now, so can't link easily. It's up to you if you want to correct my mistake and normalize properly where we fetch OMDb, or just copy how I do it

I do however believe that grabbing the first language is the correct choice.

I can maybe put up a PR to fix OMDb tomorrow, or the day after, if you want. It's up to you.

MarcoCoreDuo:
> I didn't really think it would be used elsewhere, so I didn't really think of normalize at the metadata level. But I agree that it is not optimal. You can see how I handle OMDb in my PR with the commas. I'm not at my computer right now, so can't link easily. It's up to you if you want to correct my mistake and normalize properly where we fetch OMDb, or just copy how I do it
> 
> I do however believe that grabbing the first language is the correct choice.
> 
> I can maybe put up a PR to fix OMDb tomorrow, or the day after, if you want. It's up to you.

Thanks for the quick response. I saw how you handled it in your PR, but I thought it would make more sense to normalize it at the metadata level instead. I’d really appreciate it if you could fix it when you have the time

Lampan-git:
> Thanks for the quick response. I saw how you handled it in your PR, but I thought it would make more sense to normalize it at the metadata level instead. I’d really appreciate it if you could fix it when you have the time

I have opened https://github.com/jellyfin/jellyfin/pull/16829. It normalizes OriginalLanguage to 3 letter, and `GetInheritedOriginalLanguage` should resolve the OriginalLanguage easily.

MarcoCoreDuo:
Waiting for #16829 to be merged.

MarcoCoreDuo:
Original Language for images is now supported thanks to @Lampan-git, and preferredImageLanguages now uses an ImageLanguageOption array instead of a string array.

jhparker88:
> The work on the https://github.com/jellyfin/jellyfin-meta/discussions/125 is set to start once we release 12.0 RC1, which we are trying to get out this weekend

While I would love to see this feature being shipped with 12.0, I fear given the remark from @theguymadmax this seems rather challenging hence I would urge you (@MarcoCoreDuo) to keep the PR fresh for the next release.  @theguymadmax any input?

PR #16029, opened 2026-01-15
Skip hidden directories and .ignore paths in library monitoring

### LINKED ISSUE #16021: Changes to directories with an `.ignore` file cause a library refresh and non-trivial CPU usage
### Description of the bug

I have a couple of libraries which are refreshing too frequently (i.e. logs include entries like *Emby.Server.Implementations.IO.LibraryMonitor: "music" ("/music") will be refreshed.*), and I determined this was because of hourly snapshots being created by Snapper's default configuration. These libraries are mapped directly to a BTRFS subvolume root. I've since solved the issue with snapshots being created hourly by means of a custom timeline script, however it seems the other side of my problem is that Jellyfin isn't ignoring these directories, or at least the ignore feature does not work as I expect.

I first tried using the new Git style rules in an `.ignore` file in the root of the subvolumes (The rule being `/.snapshots/`). This caused performance problems: it took many hours to complete a rescan of the affected libraries every time a snapshot was taken. I've now put a 0 byte `.ignore` file directly in the `.snapshots` subdirectory. The performance of this is drastically better, however I still have it refreshing the whole library every time a snapshot is taken. I'm inclined to think the FileSystemWatcher component of the ignore feature simply ignores files _found_ in these directories, but does not ignore the changes to these directories and subdirectories altogether, causing the refresh. After enabling debug logging, my suspicions seem confirmed:

```
[2026-01-10 14:29:15.270 +11:00] [DBG] [26] Emby.Server.Implementations.IO.LibraryMonitor: New file refresher created for "/music/.snapshots/3"
```

This line was entered immediately upon taking a manual snapshot. I think it's a bug: I think it should not be necessary to watch subdirectories of ignored directories, but perhaps somebody with a better grasp of the design could elaborate? I’m running Jellyfin 10.11.5 in Docker. Ideally we would not waste CPU cycles processing a directory tree we have no intention of loading files from.

### Reproduction steps

1. Create a new directory in the root directory of a Jellyfin library
2. `touch .ignore` in that directory to ignore it.
3. Refresh the library to ensure the ignore file is known (might not be a necessary step - simply adding the directory and `.ignore` file were enough to trigger a refresh.
4. Wait until the refresh is complete. I used `top` and waited for the jellyfin process to drop off.
5. Add a new file to the directory.
6. After a about a minute, another refresh is triggered and the CPU spikes.

The library is refreshed when contents of ignored directories are modified: the logs will contain an entry like:

```
[2026-01-14 22:08:02.293 +11:00] [INF] [71] Emby.Server.Implementations.IO.LibraryMonitor: "music" ("/music") will be refreshed.
```

I'm not sure the amount of time spent scanning would suggest it's rescanning the whole library, but it certainly seems excessive for the number of actual changes. It occupied two cores for about 30 seconds. When I was using a Git style ignore file in my library root, it took hours.

I expected that if the directory is ignored, then files added to or removed from the directory will not cause a refresh to occur, and I don't anticipate any noticeable spike in CPU usage.

### Jellyfin Server version

10.11.5

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.5

### Environment

```markdown
- OS: Debian 13
- Linux Kernel: 6.12.57+deb13-amd6
- Virtualization: Docker
- Clients: N/A
- Browser: N/A
- FFmpeg Version: N/A
- Playback Method: N/A
- Hardware Acceleration: N/A
- CPU Model: Intel Celeron N5105 (Jasper Lake)
- GPU Model: N/A
- Plugins: AudioDB, DLNA, Musicbrainz, OMDb, Studion Images, TMDb. 
- Reverse Proxy: N/A
- Base URL: N/A
- Networking: Host
- Jellyfin Data Storage & Filesystem: nvme - btrfs
- Media Storage & Filesystem: sata - btrfs
- External Integrations: N/A
```

### Jellyfin logs

```shell
Snippets included inline - debug is too verbose to include verbatim.
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

theguymadmax:
Prevents unnecessary library refreshes when files change in ignored directories.

**Changes**
- Check `.ignore` files during file system monitoring
- Add `**/.*/**` pattern to ignore contents of hidden directories

**Issues**
 Fixes #16021

### DIFF
--- a/Emby.Server.Implementations/Library/IgnorePatterns.cs
+++ b/Emby.Server.Implementations/Library/IgnorePatterns.cs
@@ -83,6 +83,7 @@ namespace Emby.Server.Implementations.Library
 
             // Unix hidden files
             "**/.*",
+            "**/.*/**",
 
             // Mac - if you ever remove the above.
             // "**/._*",
--- a/Emby.Server.Implementations/IO/LibraryMonitor.cs
+++ b/Emby.Server.Implementations/IO/LibraryMonitor.cs
@@ -352,6 +352,12 @@ namespace Emby.Server.Implementations.IO
                 return;
             }
 
+            var fileInfo = _fileSystem.GetFileSystemInfo(path);
+            if (DotIgnoreIgnoreRule.IsIgnored(fileInfo, null))
+            {
+                return;
+            }
+
             // Ignore certain files, If the parent of an ignored path has a change event, ignore that too
             foreach (var i in _tempIgnoredPaths.Keys)
             {
--- a/tests/Jellyfin.Server.Implementations.Tests/Library/IgnorePatternsTests.cs
+++ b/tests/Jellyfin.Server.Implementations.Tests/Library/IgnorePatternsTests.cs
@@ -19,7 +19,7 @@ namespace Jellyfin.Server.Implementations.Tests.Library
         [InlineData("/media/movies/#recycle", true)]
         [InlineData("thumbs.db", true)]
         [InlineData(@"C:\media\movies\movie.avi", false)]
-        [InlineData("/media/.hiddendir/file.mp4", false)]
+        [InlineData("/media/.hiddendir/file.mp4", true)]
         [InlineData("/media/dir/.hiddenfile.mp4", true)]
         [InlineData("/media/dir/._macjunk.mp4", true)]
         [InlineData("/volume1/video/Series/@eaDir", true)]
@@ -32,7 +32,7 @@ namespace Jellyfin.Server.Implementations.Tests.Library
         [InlineData("/media/music/Foo B.A.R", false)]
         [InlineData("/media/music/Foo B.A.R.", false)]
         [InlineData("/movies/.zfs/snapshot/AutoM-2023-09", true)]
-        public void PathIgnored(string path, bool expected)
+        public void PathIgnored(string path,  bool expected)
         {
             Assert.Equal(expected, IgnorePatterns.ShouldIgnore(path));
         }


Emby.Server.Implementations/Library/IgnorePatterns.cs
```diff
@@ -83,6 +83,7 @@ public static class IgnorePatterns
 
             // Unix hidden files
             "**/.*",
+            "**/.*/**",
```
cvium, line 86: Do we want this behaviour change in a patch release? Is this actually a bug?

theguymadmax: I was thinking along the same lines, but I ran a few tests, and in 10.11, you  can add a hidden folder as a library path: 
```
.hiddenfolder\
--Moive1.mkv
```
But nothing gets scanned into the library. 

Also, individual hidden directories do not get scanned in 
```
Movies\
--Movie 1\
--.hiddenmovie\
```
In 10.10.x, both of these scenarios were possible. I believe #14540 changed that behavior. So this change would just prevent library refreshes.  

If preferred, I could make this part a separate PR targeting the master, keeping just .ignore files, and also adding snapshots folders to the ignore pattern.

cvium: Keep it in then

Bond-009 (approved):

PR #16033, opened 2026-01-15
[10.11] Fix querying media with language filters

### LINKED ISSUE #9606: [Issue]: Subtitle download settings ignore library setting "Skip if the video already contains embedded subtitles"
### Please describe your bug

I have the setting in the title checked for my Movies library. When I refresh metadata for the Movies library and choose "search for missing metadata", Jellyfin downloads subtitles for hundreds of movies that already have embedded subtitles and saves them to my library.

### Jellyfin Version

10.8.0

### if other:

_No response_

### Environment

```markdown
- OS: Unraid 6.11.5 [Slackware 15+]
- Virtualization: docker
- Clients: Browser, Jellyfin Media Player 1.9.0
- Browser: Firefox 112.0b9 
- FFmpeg Version: ffmpeg version 5.1.2-Jellyfin
- Playback Method: not relevant
- Hardware Acceleration: not relevant
- Plugins: not relevant
- Reverse Proxy: traefik
- Base URL: none
- Networking: not relevant
- Storage: local
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

#### Subtitle settings:
image

#### Sample movie with embedded subtitles that Jellyfin downloaded subs for anyway:
image

### Code of Conduct

### LINKED ISSUE #12434: Opensubtitles keeps downloading same subtitles.
Copying from another forum that told me to add it here.

I have every day if you look at my pictures. It is downloading the subtitles for ones I already have.  See picture. Every day it just grabs the same subtitles for things I already have subtitles for since you can see in my second picture every single one ends with .eng.#.  The number indicates that there is already a subtitle and so it just adds a .0 or .1.  Look at the third picture called Austin.  It is the EXACT SAME FILE that it has downloaded, 2 on the same day. Is there a specific setting I am missing that needs to be set so it doesn't keep downloading the same subtitles for ones already with a set of subtitles since it keeps downloading something I already have? It also doesn't even notice new movies and adds subtitles for them since it seems to think that it has no subtitles.

I forgot to add but for a long time there would be no subtitles added since I had added nothing. but May 25th (probably an update) it started adding in repeated downloaded subtitles.

Also do I just have a temp "downloading library" that I then send over to the main once I know they have all the subtitles?
Capture1 Dodgeball
Capture 2
Austin

### LINKED ISSUE #15513: The Subtitle Downloader automatically downloads all languages selected when only 1 of those languages is missing
### Description of the bug

I have configured that subtitles must be downloaded automatically for English and Spanish, but skip the download if the video has subtitles with those languages already embedded.

If a movie only has Spanish subtitles, it downloads English and Spanish subtitles. So now I have duplicated Spanish subtitles.

### Reproduction steps

1. Go to the settings of a Movie/TV Show library and enable a subtitle downloader.
2. Select at least 2 languages.
3. Check the checkbox of the option "Skip if the video already contains embedded subtitles".
4. Run the subtitle downloader in a newly added movie/episode or run the scheduled task. The movie/episode needs to have only 1 subtitle of the selected languages.
5. The movie will have the new downloaded subtitles added with all the languages selected in the subtitle downloader settings, instead of only the missing ones from that movie/episode.

The subtitle downloader downloads all the languages selected, instead of only the ones with the missing language subtitle of the movie/episode. Like if a movie has only English subtitles and the subtitle downloader has selected English and Spanish, it will download the Spanish and the English subtitles. So now the movie will have duplicated English subtitles.

I believe the subtitle downloader should only download the languages that are missing instead of all the languages selected. If the movie is only missing the English subtitles, it should only download an English subtitle.

### Jellyfin Server version

10.11.0+

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.2

### Environment

```markdown
- OS: Debian 13
- Linux Kernel: 
- Virtualization: Docker
- Clients: Android, Kodi, Android TV
- Plugins: Open Subtitles
- Reverse Proxy: none
- Base URL: none
- Networking: Host
- Media Storage: Local SSD
```

### Jellyfin logs

```shell
[2025-11-14 23:53:31.228 -06:00] [INF] [12] MediaBrowser.Providers.Subtitles.SubtitleManager: Saving subtitles to "/media/Movies/Barbie (2023)/Barbie (2023).eng.srt"
[2025-11-14 23:53:31.858 -06:00] [INF] [3] Jellyfin.Plugin.OpenSubtitles.OpenSubtitleDownloader: Remaining subtitle downloads: 16
[2025-11-14 23:53:32.572 -06:00] [INF] [12] MediaBrowser.Providers.Subtitles.SubtitleManager: Saving subtitles to "/media/Movies/Barbie (2023)/Barbie (2023).spa.srt"
[2025-11-14 23:53:32.817 -06:00] [INF] [3] Jellyfin.Plugin.OpenSubtitles.OpenSubtitleDownloader: Remaining subtitle downloads: 16
[2025-11-14 23:53:32.869 -06:00] [INF] [3] MediaBrowser.Providers.Subtitles.SubtitleManager: Saving subtitles to "/media/Movies/Barbie (2023)/Barbie (2023).spa.0.srt"
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

Example of a movie with only Spanish subtitles (missing English subtitles)

After the subtitles are downloaded, the movie appears with duplicated Spanish subtitles.

### Additional information

_No response_

MBR-0001:
After looking into subtitle handing in jellyfin I've realized the situation is not great, some tables hold the language in one format, code queries it in another and all sorts of issues come up.

**Changes**
For all queries that include languages: localizationManager is used to get all (both?) language name variants for the wanted language.
Example: query is for german (_deu_) so via localizationManager we get both _deu_ and _ger_, and then check for both of those in the query. (From my understanding of EF `Language = 'deu'` -> `Language IN ('deu', 'ger')` .

Any input is welcome, I hope this fixes the listed issues and doesn't break anything :)

**Issues**
_Maybe (hopefully)_
fixes #9606
fixes #15513
fixes #12434
fixes https://github.com/jellyfin/jellyfin-plugin-opensubtitles/issues/185

### DIFF
--- a/Jellyfin.Api/Helpers/StreamingHelpers.cs
+++ b/Jellyfin.Api/Helpers/StreamingHelpers.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using Jellyfin.Api.Extensions;
@@ -485,7 +486,7 @@ public static class StreamingHelpers
                     request.StartTimeTicks = long.Parse(val, CultureInfo.InvariantCulture);
                     break;
                 case 15:
-                    if (videoRequest is not null && EncodingHelper.LevelValidationRegexStr().IsMatch(val))
+                    if (videoRequest is not null && Regex.IsMatch(val, EncodingHelper.LevelValidationRegexStr))
                     {
                         videoRequest.Level = val;
                     }


Shadowghost (approved):

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -2248,26 +2253,46 @@ private IQueryable<BaseItemEntity> TranslateQuery(
 
         if (!string.IsNullOrWhiteSpace(filter.HasNoAudioTrackWithLanguage))
         {
-            baseQuery = baseQuery
-                .Where(e => !e.MediaStreams!.Any(f => f.StreamType == MediaStreamTypeEntity.Audio && f.Language == filter.HasNoAudioTrackWithLanguage));
+            var lang = _localizationManager.FindLanguageInfo(filter.HasNoAudioTrackWithLanguage);
+            if (lang is not null)
+            {
+                var langSet = new HashSet<string?>(lang.ThreeLetterISOLanguageNames);
```
JPVenson: why are you creating a hashset here? that does not translate in EF queries

JPVenson (changes requested):

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
MBR-0001: `lang.ThreeLetterISOLanguageNames.Contains(f.Language)` produced
```sql
m.Language IN (
                SELECT l.value
                FROM json_each(@__lang_ThreeLetterISOLanguageNames_0) AS l
            )
```
hashset seems to produce
```sql
AND EXISTS (
    SELECT 1
    FROM json_each(@__langSet_0) AS l
    WHERE
        l.value = m.Language
        OR (
            l.value IS NULL
            AND m.Language IS NULL
        )
)
```

_allegedly_ EXISTS is faster, so that's why I went with it

does that make any sense? its very likely possible that my _source_ is hallucinating :thinking:

michnovka:
I came to the same conclusion. I think that the multiple sub downloads stems from this

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
JPVenson: Can you have a look at the `WhereOneOrMany` extension method? it handles this usecase far better by not relying on json_each if not nessesary. Also can you link the resource saying that EXIST is faster then IN? that sounds very counterintertuitive to me as IN should yield a result on the first matching item where EXIST will need to enumerate all items in the subquery (which is known to be a big performance hog overall)

MBR-0001: @JPVenson I can't figure out how to use WhereOneOrMany here, I don't have the EF skills for that :/
I can only offer one of the queries above, which one would you prefer? :D

as for source, we were taught at college that subqueries slow things down (in this case `m.Language IN (SELECT ...`)... I never encountered EXISTS until now so I just took what chatgpt and stackoverflow (can't find the post anymore) said for granted

cvium: I think Venson has it the other way around. EXISTS is a short circuit and IN executes the full subquery. If there's an index it's kinda moot though.

Either way though, it's kinda pointless arguing one way or the other without looking at the query plan.

Sidenote: I believe the openjson default is going away in EF Core 10 due to perf regressions.

MBR-0001: after the latest changes (dropping unnecessary null checks) the query is now
```sql
.param set @__lang_ThreeLetterISOLanguageNames_0 

SELECT allthecolumns
FROM "BaseItems" AS "b"
WHERE "b"."Type" IN ('MediaBrowser.Controller.Entities.TV.Episode', 'MediaBrowser.Controller.Entities.Movies.Movie') AND NOT EXISTS (
    SELECT 1
    FROM "MediaStreamInfos" AS "m"
    WHERE "b"."Id" = "m"."ItemId" AND "m"."StreamType" = 2 AND "m"."IsExternal" AND "m"."Language" IN (
        SELECT "l"."value"
        FROM json_each(@__lang_ThreeLetterISOLanguageNames_0) AS "l"
    ))
```

the query plan is
id | parent | notused | detail 
-- | -- | -- | --
3 | 0 | 0 | SEARCH b USING INDEX IX_BaseItems_Type_TopParentId_StartDate (Type=?)
23 | 0 | 0 | CORRELATED SCALAR SUBQUERY 2
28 | 23 | 0 | SEARCH m USING INDEX sqlite_autoindex_MediaStreamInfos_1 (ItemId=?)
40 | 23 | 0 | LIST SUBQUERY 1
42 | 40 | 0 | SCAN l VIRTUAL TABLE INDEX 1:

MBR-0001:
accidentally deleted my commit and can't reopen, new PR: https://github.com/jellyfin/jellyfin/pull/16538

PR #16037, opened 2026-01-16
feat: link-share playlists.

pantierra:
Allows playlist owners to share playlists via link with users who don't have an account (a.k.a. anonymous users).

Implementation approach:

- Share tokens: 64-char hex strings via `RandomNumberGenerator`
- Authorization: `PlaylistShareAccessPolicy` for token-based anonymous access
- Backend: `ShareToken` property on `Playlist`, new `PlaylistManager` methods, custom authorization handler

### API Endpoints

**Owner operations:**
- `POST /Playlists/{playlistId}/Share` - Generate share token
- `DELETE /Playlists/{playlistId}/Share` - Revoke share token
- `GET /Playlists/{playlistId}/Share` - Get share link (returns 404 if none exists)

**Anonymous access (token-based):**
- `GET /Playlists/Share/{shareToken}` - Get playlist metadata
- `GET /Playlists/Share/{shareToken}/Items` - Get playlist items
- `GET /Audio/{itemId}/stream?shareToken={token}` - Stream audio
- `GET /Videos/{itemId}/stream?shareToken={token}` - Stream video

### DIFF
--- a/Emby.Server.Implementations/Playlists/PlaylistManager.cs
+++ b/Emby.Server.Implementations/Playlists/PlaylistManager.cs
@@ -7,11 +7,16 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Security.Cryptography;
+using System.Text;
+using System.Text.Json;
 using System.Threading;
 using System.Threading.Tasks;
 using Jellyfin.Data.Enums;
+using Jellyfin.Database.Implementations;
 using Jellyfin.Database.Implementations.Entities;
 using Jellyfin.Extensions;
+using Jellyfin.Extensions.Json;
 using MediaBrowser.Controller.Dto;
 using MediaBrowser.Controller.Entities;
 using MediaBrowser.Controller.Entities.Audio;
@@ -22,6 +27,7 @@ using MediaBrowser.Controller.Providers;
 using MediaBrowser.Model.Entities;
 using MediaBrowser.Model.IO;
 using MediaBrowser.Model.Playlists;
+using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
 using PlaylistsNET.Content;
@@ -40,6 +46,7 @@ namespace Emby.Server.Implementations.Playlists
         private readonly IUserManager _userManager;
         private readonly IProviderManager _providerManager;
         private readonly IConfiguration _appConfig;
+        private readonly IDbContextFactory<JellyfinDbContext> _dbProvider;
 
         public PlaylistManager(
             ILibraryManager libraryManager,
@@ -48,7 +55,8 @@ namespace Emby.Server.Implementations.Playlists
             ILogger<PlaylistManager> logger,
             IUserManager userManager,
             IProviderManager providerManager,
-            IConfiguration appConfig)
+            IConfiguration appConfig,
+            IDbContextFactory<JellyfinDbContext> dbProvider)
         {
             _libraryManager = libraryManager;
             _fileSystem = fileSystem;
@@ -57,6 +65,7 @@ namespace Emby.Server.Implementations.Playlists
             _userManager = userManager;
             _providerManager = providerManager;
             _appConfig = appConfig;
+            _dbProvider = dbProvider;
         }
 
         public Playlist GetPlaylistForUser(Guid playlistId, Guid userId)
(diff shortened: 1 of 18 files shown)


pantierra:
These compatibility warnings seem to be false positives. The changes are backward compatible:

**OpenAPI Diff Report**: The `shareToken` query parameter added to stream endpoints is optional (`string?`), so existing clients continue to work without modification. The new playlist share endpoints are additions, not modifications.

**ABI difference**: `IPlaylistManager` is an internal implementation interface used only by the `PlaylistManager` class. It's not part of the public plugin API, so adding methods doesn't break external code, as this interface isn't exposed to plugins or external implementations.

Emby.Server.Implementations/Playlists/PlaylistManager.cs
```diff
@@ -656,5 +665,95 @@ private async Task UpdatePlaylistInternal(Playlist playlist)
...
+            var playlistEntity = dbContext.BaseItems
+                .AsNoTracking()
+                .Where(e => e.Type == playlistType && e.Data != null)
+                .AsEnumerable()
+                .FirstOrDefault(e =>
+                {
+                    try
+                    {
+                        var playlist = JsonSerializer.Deserialize<Playlist>(e.Data, JsonDefaults.Options);
+                        return playlist != null && string.Equals(playlist.ShareToken, shareToken, StringComparison.OrdinalIgnoreCase);
```
dfederm, line 735: There is a timing attack vulnerability here. `string.Equals` checks character-by-character, so an attacker could potentially determine valid tokens character-by-character by measuring response time.

The industry standard is to use constant-time comparisons.

MediaBrowser.Controller/Playlists/Playlist.cs
```diff
@@ -44,6 +44,8 @@ public Playlist()
 
         public IReadOnlyList<PlaylistUserPermissions> Shares { get; set; }
 
+        public string ShareToken { get; set; }
```
dfederm, line 47: This is stored in plain text in the database. If the database is compromised, all share tokens are exposed, and unlike passwords, these tokens grant direct access to content.

Consider storing a hash of the token instead.

Jellyfin.Api/Controllers/AudioController.cs
```diff
@@ -85,13 +101,19 @@ public AudioController(AudioHelper audioHelper)
     /// <param name="streamOptions">Optional. The streaming options.</param>
     /// <param name="enableAudioVbrEncoding">Optional. Whether to enable Audio Encoding.</param>
     /// <response code="200">Audio stream returned.</response>
+    /// <response code="403">Access forbidden.</response>
+    /// <response code="404">Item not found or not accessible via share token.</response>
     /// <returns>A <see cref="FileResult"/> containing the audio file.</returns>
     [HttpGet("{itemId}/stream", Name = "GetAudioStream")]
     [HttpHead("{itemId}/stream", Name = "HeadAudioStream")]
+    [AllowAnonymous]
```
dfederm, line 109: This is a bit scary. It removes all authorization from the endpoint. The token validation does happen manually i the method body, but any bug could lead to the entire library having unfettered anonymous access.

Jellyfin.Api/Controllers/AudioController.cs
```diff
@@ -85,13 +101,19 @@ public AudioController(AudioHelper audioHelper)
...
     [ProducesResponseType(StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status403Forbidden)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
     [ProducesAudioFile]
     public async Task<ActionResult> GetAudioStream(
```
dfederm, line 114: There's no rate limiting on this endpoint, so an attacker could brute-force this, especially combined with the timing attack I metnioned in another comment.

Emby.Server.Implementations/Playlists/PlaylistManager.cs
```diff
@@ -656,5 +665,95 @@ private async Task UpdatePlaylistInternal(Playlist playlist)
...
+            {
+                return null;
+            }
+
+            using var dbContext = _dbProvider.CreateDbContext();
+            var playlistType = typeof(Playlist).ToString();
+            var playlistEntity = dbContext.BaseItems
+                .AsNoTracking()
+                .Where(e => e.Type == playlistType && e.Data != null)
+                .AsEnumerable()
```
dfederm, line 729: This `AsEnumerable` will load ALL playlists into memory and then deserialize each one to find a matching token. That's `O(n)` for each and every token validation.

And remember that token validation happens in an anonymous endpoint, so this could easily lead to a DOS attack.

I strongly suggest storing the token (well, a hash of it as mentioned in another comment) as a separate indexed column in the database

Emby.Server.Implementations/Playlists/PlaylistManager.cs
```diff
@@ -656,5 +665,95 @@ private async Task UpdatePlaylistInternal(Playlist playlist)
...
+            var shareToken = Convert.ToHexString(tokenBytes).ToLowerInvariant();
+            playlist.ShareToken = shareToken;
+
+            await UpdatePlaylistInternal(playlist).ConfigureAwait(false);
+
+            return shareToken;
+        }
+
+        /// <inheritdoc />
+        public async Task RevokeShareToken(Guid playlistId, Guid userId)
```
dfederm, line 699: There does not appear to be any expiration to these share tokens so once generated they're valid forever unless manually revoked. This matters since shared links can be effectively forwarded forever to many many people and expose access to the resource until the owner notices.

Emby.Server.Implementations/Playlists/PlaylistManager.cs
```diff
@@ -656,5 +665,95 @@ private async Task UpdatePlaylistInternal(Playlist playlist)
                 SavePlaylistFile(playlist);
             }
         }
+
+        /// <inheritdoc />
+        public async Task<string> GenerateShareToken(Guid playlistId, Guid userId)
```
dfederm, line 670: There should be an audit log for all token operations. Currently there is no logging at all.

This matters as playlist owners cannot see who has accessed their shared content. Admins cannot audit shared content access. And in general there's no way to investigate what if any content has leaked due to this feature.

dfederm:
Although well-intentioned, I have some pretty strong security and privacy concerns with this feature.

PR #16066, opened 2026-01-19
Fix TMDB crew department mapping

### LINKED ISSUE #16064: No director displayed on newly added movies since 10.11.6
### Description of the bug

After upgrading to 10.11.6, newly added movies don't show director info. This is likely related to https://github.com/jellyfin/jellyfin/pull/15872

Image

### Reproduction steps

Add a new movie and look at the details page (alternately Replace All Metadata for an existing movie)

The item details page shows no Director

Should show a director

### Jellyfin Server version

10.11.6

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.6

### Environment

```markdown
Debian up to date, using docker
```

### Jellyfin logs

```shell
No logging required, nothing in logs about this issue
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

theguymadmax:
Regression from #15872

**Changes**
1. Add correct mapping for director to the Directing department.
2. Add Screenplay role to the Writing department

**Issues**
Fixes #16064

### DIFF
--- a/MediaBrowser.Providers/Plugins/Tmdb/TmdbUtils.cs
+++ b/MediaBrowser.Providers/Plugins/Tmdb/TmdbUtils.cs
@@ -69,7 +69,7 @@ namespace MediaBrowser.Providers.Plugins.Tmdb
         /// <returns>The Jellyfin person type.</returns>
         public static PersonKind MapCrewToPersonType(Crew crew)
         {
-            if (crew.Department.Equals("production", StringComparison.OrdinalIgnoreCase)
+            if (crew.Department.Equals("directing", StringComparison.OrdinalIgnoreCase)
                 && crew.Job.Equals("director", StringComparison.OrdinalIgnoreCase))
             {
                 return PersonKind.Director;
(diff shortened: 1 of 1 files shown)


Shadowghost (approved):
are we sure it can't be both?

theguymadmax:
Directing looks to be the only one it would be in. https://www.themoviedb.org/talk/598c3a70925141080100e601. Mostly likely it worked before because of the fallback logic.

IDisposable (approved):

booxter:
Will users have to refresh metadata manually to fix affected entries or will it eventually be picked up by some periodic job?

theguymadmax:
> Will users have to refresh metadata manually to fix affected entries or will it eventually be picked up by some periodic job?

A manual refresh of the library or individual items with the option search for missing metadata is needed to populate the missing entries.

tpolakov:
When is this fix going to be available?

Rand0mB0y:
I have just completed a long and painful migration from Windows (JF v10.11.5) to Proxmox (JF v10.11.6), just to realise AT THE VERY END that libraries now have no directors and very few writers. I use directors all the time, so this is unusable for me, ATM.
I'd highly appreciate a patch to fix the issue, as I do not have a downgradable version I can go to, unless I rebuild the whole system again with 10.11.5, but this is a large multi-language  library, a major PITA to set it all up.
Thanks so much as always for your work, guys... not demanding anything, just asking if this could be fixed sooner than later, or if there is a quirk I can try on my own.

theguymadmax:
The fix will be included in the next bugfix release (10.11.7).  There's no real timeline for those point releases. It happens when there are usually enough changes to warrant a release. That being said, it's been a while since our last point release (10.11.6), so I'd expect a release sooner rather than later.

PR #16071, opened 2026-01-20
Rehydrate cached UserData after reattachment

MarcoCoreDuo:
After renaming media items, the GetItem API returns cached items without the reattached UserData. This occurs because ReattachUserDataAsync updates the database but does not rehydrate the in-memory item instance. As a result, navigating to a movie or episode page incorrectly shows the item as not played.

**Changes**
- Rehydrate item.UserData after database update in ReattachUserDataAsync by reloading UserData from the database
- Wrap the operation in a transaction

**Issues**

Fixes https://forum.jellyfin.org/t-new-jellyfin-server-web-release-10-11-6?pid=69525#pid69525

### DIFF
--- a/Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
+++ b/Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
@@ -753,10 +753,10 @@ public sealed class BaseItemRepository
 
         var dbContext = await _dbProvider.CreateDbContextAsync(cancellationToken).ConfigureAwait(false);
 
-        await using (dbContext)
+        await using (dbContext.ConfigureAwait(false))
         {
             var transaction = await dbContext.Database.BeginTransactionAsync(cancellationToken).ConfigureAwait(false);
-            await using (transaction)
+            await using (transaction.ConfigureAwait(false))
             {
                 var userKeys = item.GetUserDataKeys().ToArray();
                 var retentionDate = (DateTime?)null;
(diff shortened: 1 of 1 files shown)


Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -753,18 +753,33 @@ public async Task ReattachUserDataAsync(BaseItemDto item, CancellationToken canc
 
         var dbContext = await _dbProvider.CreateDbContextAsync(cancellationToken).ConfigureAwait(false);
 
-        await using (dbContext.ConfigureAwait(false))
+        await using (dbContext)
```
JPVenson: why this change?

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -753,18 +753,33 @@ public async Task ReattachUserDataAsync(BaseItemDto item, CancellationToken canc
...
+                        e => e
+                            .SetProperty(f => f.ItemId, item.Id)
+                            .SetProperty(f => f.RetentionDate, retentionDate),
+                        cancellationToken).ConfigureAwait(false);
+
+                // Rehydrate the cached userdata
+                item.UserData = await dbContext.UserData
+                    .AsNoTracking()
+                    .Where(e => e.ItemId == item.Id)
+                    .Where(e => userKeys.Contains(e.CustomDataKey))
```
JPVenson: why this check? it seems a bit useless

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
MarcoCoreDuo: Sorry, forgot to add it while changing code. I'll add it back.

Issue #16075, opened 2026-01-21
Overzealous ignoring of directories during library scan


thatsatoffee:
### Description of the bug

After upgrading to 10.11.6, my library at a location like `/home/jellyfin/.media/video/movies` is no longer scanned.

### Reproduction steps

1. Add a library at a location that is nested in a unix-style hidden directory (like `/path/to/.hidden/library/location`)
2. Scan the library

### What is the current _bug_ behavior?

No files in the library are indexed; the directory is ignored entirely.

### What is the expected _correct_ behavior?

The library should be scanned at all points where there is no `.ignore` file.

### Jellyfin Server version

10.11.6

### Specify the build version

10.11.6

### Environment

```markdown
- OS: Alpine Linux (server)
- Linux Kernel: 6.8.12-4-pve
- Virtualization: Docker
- Jellyfin Data Storage & Filesystem: RAID on local SATA HDDs bind-mounted at specific locations in Docker
- Media Storage & Filesystem: See above
```

### Jellyfin logs

```shell
[2026-01-21 08:47:21.857 +00:00] [INF] [44] Emby.Server.Implementations.IO.LibraryMonitor: Stopping directory watching for path "/home/jellyfin/.media/video/movies"
[2026-01-21 08:47:21.857 +00:00] [INF] [44] Emby.Server.Implementations.Library.LibraryManager: Validating media library
[2026-01-21 08:47:21.894 +00:00] [WRN] [44] MediaBrowser.Controller.Entities.BaseItem: Library folder "/config/data/playlists" is inaccessible or empty, skipping
[2026-01-21 08:47:21.955 +00:00] [WRN] [44] MediaBrowser.Controller.Entities.BaseItem: Library folder "/config/data/playlists" is inaccessible or empty, skipping
[2026-01-21 08:47:30.002 +00:00] [INF] [44] Emby.Server.Implementations.ScheduledTasks.TaskManager: "Scan Media Library" Completed after 0 minute(s) and 8 seconds
[2026-01-21 08:47:30.141 +00:00] [INF] [40] Emby.Server.Implementations.IO.LibraryMonitor: Watching directory "/home/jellyfin/.media/video/movies"
```

### FFmpeg logs

```shell

```

### Additional information

I believe to have located the source of this behaviour in PR #16029.

[...]

PR #16077, opened 2026-01-21
Revert hidden directory ignore pattern

theguymadmax:
Reverts the hidden directory ignore pattern introduced in #16029, due to undesired behavioral changes for library paths:

```
10.10.x
\media\.Movies\  Works
\.media\Movies\  Works

10.11.5
\media\.Movies\  Library ignored
\.media\Movies\  Works

10.11.6
\media\.Movies\  Library ignored
\.media\Movies\  Library ignored
```
**Changes**

Remove `**/.*/**` from ignore patterns
Add `.snapshots` patterns to ignore snapshot directories and their contents

**Issues**
Fixes: #16075


sjorge:
Probably also want to include .snapshot (no s), i see both at work on various filesystems and it absolutely sucks nobody seems to agree on this.

PR #16079, opened 2026-01-21
Fix: Sync HLS playlist sequence with FFmpeg segment numbers

ZeusCraft10:
Hey team, I finally cracked that nasty HLS infinite buffering bug we've been seeing with Dolby Vision content! 🐛

The root cause was pretty tricky: when you seek, FFmpeg restarts and creates segment files starting from a new index (like 1187.mp4), but our server was still generating playlists starting at 0.mp4. The client would ask for segment 0, get a 404, and just spin forever.

This PR fixes it by:
1. Checking for existing segment files on disk when generating the playlist.
2. Using that file's index as the starting point for the playlist sequence.

Now the playlist URLs match the actual files on disk, so playback resumes instantly after seeking. 🚀

I've added a unit test to cover this case. Let me know what you think!

### DIFF
--- a/Jellyfin.Api/Controllers/DynamicHlsController.cs
+++ b/Jellyfin.Api/Controllers/DynamicHlsController.cs
@@ -1399,6 +1399,18 @@ public class DynamicHlsController : BaseJellyfinApiController
                 TranscodingJobType,
                 cancellationTokenSource.Token)
             .ConfigureAwait(false);
+
+        // Calculate the starting segment index from current transcoding state
+        var playlistPath = Path.ChangeExtension(state.OutputFilePath, ".m3u8");
+        var segmentExtension = EncodingHelper.GetSegmentFileExtension(state.Request.SegmentContainer);
+        var startSegmentIndex = 0;
+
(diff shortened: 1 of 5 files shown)


ZeusCraft10:
> > nasty HLS infinite buffering bug we've been seeing with Dolby Vision content
> 
> I haven't encountered the issue you mentioned with various Dolby Vision videos in my collection. It would be best to provide a video sample that reproduces the problem in HLS.js or Apple's HLS player.
> 
> > but our server was still generating playlists starting at 0.mp4
> 
> This is expected. The server generates a complete, fixed-length playlist to inform the HLS player of the total video length so you can seek forward or backward, rather than editing the playlist after a seek.
> 
> When the client requests the Nth segment, the server terminates the previous transcoding process and seeks to the position corresponding to the Nth segment to begin transcoding. Therefore, if the client requests the 0th segment, it shouldn't get a 404 error.

sorry for the late response, Looks like I understood the issue incorrectly. I did try to reproduce the issue, but it did not reproduce. Im not sure what even caused it in the first place. let me take a look, but Im closing this PR for now

PR #16084, opened 2026-01-23
Fix random sort inflated item count

### LINKED ISSUE #16035: Random sort produces inflated item count when pagination is disabled
### Description of the bug

When sorting a library by Random, Jellyfin reports an incorrect item count that is much higher than the actual number of items (around 4× in my case).

This happens for both Movies and TV Shows. The issue shows up in both the UI and the API. On the API side, it occurs when using the user-scoped Items endpoint (/Users/{userId}/Items) with SortBy=Random. The non-user-scoped Items endpoint (/Items) returns the correct count under the same conditions.

In the UI, disabling pagination (library page size set to 0) makes the problem obvious, with an inflated count and apparent duplicate entries. Turning pagination back on can make the UI look correct, but the API still returns the wrong TotalRecordCount.

### Reproduction steps

1. Create or use an existing Movies or TV Shows library with a known number of items.
2. Set the library page size to 0 (disable pagination) and save the setting.
3. Open the library in the web UI.
4. Change the sort order to Random.
5. Observe the reported item count and list contents.

API reproduction:

1. Call the user-scoped Items endpoint with Random sorting:
   /Users/{userId}/Items?Recursive=true&SortBy=Random
2. Observe the TotalRecordCount in the response.

The item count shown is much higher than the actual number of items in the library. When pagination is disabled, the UI shows an inflated count and duplicate entries. The user-scoped Items API endpoint also returns an incorrect TotalRecordCount when using Random sorting.

The item count should match the actual number of items in the library. Sorting by Random should not change the total count, and the UI and API should report the same correct value without showing duplicate entries.

### Jellyfin Server version

10.11.5

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.5

### Environment

```markdown
- OS: Pop!_OS 22.04 LTS
- Linux Kernel: 6.12.10-76061203-generic
- Virtualization: Docker
- Clients: Android, iOS, Desktop
- Browser: Chrome
- Plugins:
Installed Plugins:
Artwork
AudioDB
Cover Art Archive
Intro Skipper
MusicBrainz
OMDb
Open Subtitles
Reports
Studio Images
Subtitle Extract
TMDb
TMDb Box Sets
Webhook
- Reverse Proxy:
nginx
- Jellyfin Data Storage & Filesystem: SSD
- Media Storage & Filesystem: HDD
- External Integrations:
Ersatz TV, Jellyseerr, Jellystat, Tracearr, son/rad/baz/prowlarr
```

### Jellyfin logs

```shell
No relevant logs
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

ZeusCraft10:
Fixes #16035

I noticed that when sorting by Random, the total item count was coming back way too high - around 4 times what it should be. After digging into it, I found that the ORDER BY Random() clause was being applied before the COUNT query ran, which caused Entity Framework to generate a query where the random ordering interfered with the count calculation.

The fix is pretty straightforward - I moved the ordering step to happen after we calculate the count in GetItems(). I also had to add explicit ordering calls in a few other methods (GetItemList, GetLatestItemList, and ApplyQueryFilter) to make sure they still sort things correctly.

**Testing:**
- Build passes with no warnings or errors
- All 414 unit tests pass
- Manual testing with the reproduction steps from the issue would be helpful to fully verify the fix

### DIFF
--- a/Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
+++ b/Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
@@ -274,6 +274,8 @@ public sealed class BaseItemRepository
             result.TotalRecordCount = dbQuery.Count();
         }
 
+        // Apply ordering after count to prevent EF.Functions.Random() from affecting COUNT queries
+        dbQuery = ApplyOrder(dbQuery, filter, context);
         dbQuery = ApplyQueryPaging(dbQuery, filter);
         dbQuery = ApplyNavigations(dbQuery, filter);
 
@@ -294,6 +296,7 @@ public sealed class BaseItemRepository
         dbQuery = TranslateQuery(dbQuery, context, filter);
 
         dbQuery = ApplyGroupingFilter(context, dbQuery, filter);
+        dbQuery = ApplyOrder(dbQuery, filter, context);
         dbQuery = ApplyQueryPaging(dbQuery, filter);
         dbQuery = ApplyNavigations(dbQuery, filter);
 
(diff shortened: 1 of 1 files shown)


theguymadmax:
This PR did not fix the linked issue.  Closing since #16098 was merged, which does.

PR #16103, opened 2026-01-25
Add PreferOriginalReleaseDate option for music

arminius-smh:
**Changes**

Jellyfin right now uses only `track.Date`, which is trying to mainly use the recording date for the metadata: https://github.com/Zeugma440/atldotnet/blob/5b1966200449df6d3cb8d2208b9adfeaad5e69ef/ATL/Entities/MetadataHolder.cs#L130.

My changes tries to get the actual original release date: https://github.com/Zeugma440/atldotnet/blob/5b1966200449df6d3cb8d2208b9adfeaad5e69ef/ATL/Entities/MetadataHolder.cs#L222
or the publishing date: https://github.com/Zeugma440/atldotnet/blob/5b1966200449df6d3cb8d2208b9adfeaad5e69ef/ATL/Entities/MetadataHolder.cs#L262 before then trying the normal date at the end again

It also sets the production year from the premierdate, since that is also done for albums but is kept as a last fallback as well.

So I believe overall no metadata should get lost, only the earlier dates filled for the production

**Issues**

The issue is that if releases have specific version info tagged, jellyfin tries to use that instead of the original date, its mostly a non issue for newer music, where there is the digital release and thats it, but especially for older music with different versions that may be released much later than the original release, it leads to these issues:

Before (basically random numbers of when the specific issue was released, which leads to odd sortings):
image

After (original release dates):
image

### DIFF
--- a/MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
+++ b/MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
@@ -177,7 +177,6 @@ namespace MediaBrowser.Providers.MediaInfo
             // To properly handle fallback values, we make a clone of those fields when valid.
             var trackTitle = (string.IsNullOrEmpty(track.Title) ? mediaInfo.Name : track.Title)?.Trim();
             var trackAlbum = (string.IsNullOrEmpty(track.Album) ? mediaInfo.Album : track.Album)?.Trim();
-            var trackYear = track.Year is null or 0 ? mediaInfo.ProductionYear : track.Year;
             var trackTrackNumber = track.TrackNumber is null or 0 ? mediaInfo.IndexNumber : track.TrackNumber;
             var trackDiscNumber = track.DiscNumber is null or 0 ? mediaInfo.ParentIndexNumber : track.DiscNumber;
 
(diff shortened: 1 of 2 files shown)


MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
```diff
@@ -318,28 +317,40 @@ private async Task FetchDataFromTags(Audio audio, Model.MediaInfo.MediaInfo medi
                 audio.ParentIndexNumber ??= trackDiscNumber;
             }
 
-            if (track.Date.HasValue)
+            if (track.OriginalReleaseDate.HasValue && track.OriginalReleaseDate != DateTime.MinValue)
             {
-                audio.PremiereDate = track.Date;
+                audio.PremiereDate = track.OriginalReleaseDate.Value.ToLocalTime();
```
dfederm: The use of `ToLocalTime` is a change in behavior.

Also, I believe that these DateTimes are `DateTimeKind.Unspecified`, so this could shift the date by a whole day depending on the server timezone.

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
```diff
@@ -318,28 +317,40 @@ private async Task FetchDataFromTags(Audio audio, Model.MediaInfo.MediaInfo medi
...
+            }
+            else if (track.PublishingDate.HasValue && track.PublishingDate != DateTime.MinValue)
+            {
+                audio.PremiereDate = track.PublishingDate.Value.ToLocalTime();
+            }
+            else if (track.Date.HasValue)
```
dfederm: Here you check for `HasValue` but you should also check for `MinValue` as you do above with the publishing date. Although I suppose this matches existing behavior...

dfederm:
I wonder if this should be a user setting because I could see it both ways. The current behavior might be desirable as it's when _this_ file was produced, eg a newer remaster of an older song.

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
arminius-smh: Ah right, sorry, i forgot to mention, that was actually a bug fix for the that it stops shifting it by a whole day depending on the server timezone

I am in Timezone Europe/Berlin, if the metadata of the song is 1985-01-01, for example track.Date would be 1985-01-01 00:00:00, but after assigning it to audio.PremierDate in the web view it would displayed as 1984-12-31 23:00:00, and uhh this fixed it for me, now with ToLocalTime then it would be saved and displayed as 1985-01-01 00:00:00

This wasn't really noticeable before, since the Year option was just set with Track.Year instead of being referred from the PremierDate

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
arminius-smh: Okay I just added that check as well. i think it makes it more easy to understand whats being tried as well

arminius-smh:
> I wonder if this should be a user setting because I could see it both ways. The current behavior might be desirable as it's when _this_ file was produced, eg a newer remaster of an older song.

Okay, sure I made an option out of it and made we pr in the web repo for it: https://github.com/jellyfin/jellyfin-web/pull/7550

Oh and I suppose this feature request can be closed if this would be merged: https://features.jellyfin.org/posts/1711/option-to-use-original-release-date-instead-of-date-in-audio-metadata


I just thought that having that as an default made more sense for "premierDate" in general, but ig so its more clear to as to what is happening and doesnt change existing behaviour

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
dfederm: The server time and client time could be different though. So "local" time here is the server time, not the client's (user's) time.

The bug you're describing seems like something (client or server) isn't treating the time input properly. As this is a "date only", the timezone really should be unspecified and/or UTC. I believe there is actually a DateOnly structure in C# to deal with this, although admittedly I dont know if it works with EFCore nor do I suggest making a schema change in this PR as thats much larger scope probably.

In any case though,  I think this is the wrong place to be shifting the timezone to correct for the bug you noticed. I'd guess it should be on the client input side for that data.

dfederm:
> > I wonder if this should be a user setting because I could see it both ways. The current behavior might be desirable as it's when _this_ file was produced, eg a newer remaster of an older song.
> 
> Okay, sure I made an option out of it and made we pr in the web repo for it: [jellyfin/jellyfin-web#7550](https://github.com/jellyfin/jellyfin-web/pull/7550)
> 
> Oh and I suppose this feature request can be closed if this would be merged: https://features.jellyfin.org/posts/1711/option-to-use-original-release-date-instead-of-date-in-audio-metadata
> 
> I just thought that having that as an default made more sense for "premierDate" in general, but ig so its more clear to as to what is happening and doesnt change existing behaviour

To be clear, this is just an observation I had since I could see both behaviors being desirable for certain scenarios. I'll defer to the Jellyfin devs as to what the behavior should actually be.

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
arminius-smh: Okay, sure I removed it.

rharish101:
Hi, is this still being worked on? I'm willing to test this out, if that's keeping the maintainers from merging it.

arminius-smh:
> Hi, is this still being worked on? I'm willing to test this out, if that's keeping the maintainers from merging it.

The work is finished I'd say
I used/tested this pr for a couple of months just fine myself without any problems

Issue #16102, opened 2026-01-25
TMDB image fetching constructs incorrect URL (missing size parameter)


Gladtbam:
### Description of the bug

I encountered this issue specifically when using the "Edit Images" interface to search for remote images.

When I open the "Edit Images" modal for a media item and attempt to search/download an image from TMDB, the image fails to load. Inspecting the request shows that the constructed URL is malformed because it is missing the image size parameter.

Correct format: `Base URL` + `Size` + `File Path`
e.g: `https://image.tmdb.org/t/p/original/qrGeDTOW6OdoqJ3IQc8ffq2pNRm.jpg`

Current issue: The system constructs the URL without the Size parameter (e.g., w500 or original) between the base URL and the filename.
e.g: `https://image.tmdb.org/t/p//qrGeDTOW6OdoqJ3IQc8ffq2pNRm.jpg`

### Reproduction steps

1. Open the Jellyfin Web Client and log in as an admin.
2. Navigate to a movie or show.
3. Click the three-dot menu "..." and select "Edit Images".
4. Click the Search (magnifying glass) icon to look for new images.
5. (Ensure TMDB is selected as a provider).
6. The search results (or the preview images) fail to display.
7. Checking the browser console (Network tab) reveals an  error for the TMDB image request due to the malformed URL (missing size segment).

### What is the current _bug_ behavior?

The system requests the image using a URL that lacks the size parameter. Example of the incorrect URL: https://image.tmdb.org/t/p//qrGeDTOW6OdoqJ3IQc8ffq2pNRm.jpg (Note the missing size between /p/ and the filename).

### What is the expected _correct_ behavior?

The system should construct the URL with a valid size parameter as per TMDB API standards.

The URL should look like: https://image.tmdb.org/t/p/original/qrGeDTOW6OdoqJ3IQc8ffq2pNRm.jpg

### Jellyfin Server version

10.11.6

### Specify the build version

10.11.6

### Environment

```markdown
- OS: Debian GNU/Linux 11 (bullseye)
- Linux Kernel: 5.10.0-26
- Virtualization: KVM
- Clients: Browser
- Browser: Microsoft Edge 43.0.3650.96
- FFmpeg Version: jjellyfin-ffmpeg7/unknown,now 7.1.3-1-bullseye
- Playback Method:
- Hardware Acceleration:
- CPU Model: AMD Ryzen 5 3600 6-Core Processor
- GPU Model: none
- Plugins: 
[...]

PR #16106, opened 2026-01-25
Fix TMDB image URLs missing size parameter

saltpi:
**Changes**

When TMDB plugin configuration size values (PosterSize, BackdropSize, etc.) 
are null or empty, the GetUrl() method now defaults to "original" instead of 
passing null/empty to the TMDb client. This prevents malformed URLs like 
"https://image.tmdb.org/t/p//filename.jpg" and ensures proper format 
"https://image.tmdb.org/t/p/original/filename.jpg".

Fixes remote image search failing to display TMDB images in Edit Images modal.

**Issues**

Fix TMDB image URLs missing size parameter #16102


Shadowghost:
IMO this should be targeting the release branch and be backported later automatically

crobibero:
Agreed, this can be rebased onto release-10.11.z

PR #16110, opened 2026-01-25
Remove some deprecated API members

nielsvanvelzen:
**Changes**

Remove deprecated endpoints and model properties that have been non-functional for at least 1 major release.

**Issues**

### DIFF
--- a/MediaBrowser.Model/Session/ClientCapabilities.cs
+++ b/MediaBrowser.Model/Session/ClientCapabilities.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.ComponentModel;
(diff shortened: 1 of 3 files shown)


MediaBrowser.Model/Session/ClientCapabilities.cs
dfederm, line 1: Nit, remove `using System.ComponentModel;` from this file. You may want to audit the other files as well.

nielsvanvelzen: Good catch! I thought my IDE was configured to automatically clean those but I guess it doesn't for my clone of the jellyfin server.

PR #16121, opened 2026-01-26
Implement search providers

Shadowghost:
Extracts searching into internal and external search providers.

Example plugin for external search: https://github.com/Shadowghost/jellyfin-plugin-meilisearch

**Changes**
* Extract existing search into SqlSearchprovider


JPVenson (approved):

Emby.Server.Implementations/Library/Search/SearchManager.cs
```diff
@@ -0,0 +1,454 @@
...
+            results = await CollectFromProvidersAsync(_internalProviders, query, searchTerm, cancellationToken).ConfigureAwait(false);
+        }
+
+        // External providers don't know about user permissions, so they may return IDs from
+        // hidden libraries or items the user is otherwise blocked from. Filter the candidate
+        // set to only items this user can access (top-parent libraries, parental rating,
+        // blocked/allowed tags, owned-item rules) before returning. The Items controller's
+        // second roundtrip via folder.GetItems applies most of these again, but it does not
+        // restrict by TopParentIds when ItemIds is set, leaving a gap that this closes.
+        if (results.Count > 0 && query.UserId.HasValue && !query.UserId.Value.IsEmpty())
```
JPVenson: if that is primarily needed for external providers only i assume the internal ones should do that by themself, then it should only run for external provided results? that would reduce the impact

Emby.Server.Implementations/Library/Search/SearchManager.cs
```diff
@@ -0,0 +1,454 @@
...
+        // TopParentIds for the user's accessible libraries — we call it before assigning ItemIds
+        // because LibraryManager.AddUserToQuery skips TopParentIds when ItemIds is non-empty.
+        var accessFilter = new InternalItemsQuery(user);
+        _libraryManager.ConfigureUserAccess(accessFilter, user);
+
+        var candidateIds = new Guid[candidates.Count];
+        for (var i = 0; i < candidates.Count; i++)
+        {
+            candidateIds[i] = candidates[i].ItemId;
+        }
```
JPVenson:
```suggestion
        Guid[] candidateIds = [.. candidates.Select(e => e.Itemid)];
```

Emby.Server.Implementations/Library/Search/SearchManager.cs
```diff
@@ -0,0 +1,454 @@
...
+
+        var dbContext = await _dbProvider.CreateDbContextAsync(cancellationToken).ConfigureAwait(false);
+        await using (dbContext.ConfigureAwait(false))
+        {
+            var baseQuery = dbContext.BaseItems
+                .AsNoTracking()
+                .Where(e => candidateIds.Contains(e.Id));
```
JPVenson:
```suggestion
                .WhereOneInMany(candidateIds, e => e.Id);
```

JPVenson: We might wanna also consider either limiting the result for external providers or doing this in batches to avoid creating queries that are too big to parse. I would say we should rather limit the external providers results first.

Emby.Server.Implementations/Library/Search/SearchManager.cs
```diff
@@ -0,0 +1,454 @@
...
+
+            baseQuery = _queryHelpers.ApplyAccessFiltering(dbContext, baseQuery, accessFilter);
+
+            var allowedIds = await baseQuery
```
JPVenson, line 156: Instead of materializing the list here, i would rather use the Count() extension to get the elements and then use it as an inline query down below

Emby.Server.Implementations/Library/Search/SearchManager.cs
```diff
@@ -0,0 +1,454 @@
...
+
+        if (query.StartIndex.HasValue)
+        {
+            orderedResults = orderedResults.Skip(query.StartIndex.Value).ToList();
+        }
+
+        if (query.Limit.HasValue)
+        {
+            orderedResults = orderedResults.Take(query.Limit.Value).ToList();
+        }
```
JPVenson, line 258: Shouldnt the pagination rather be handled via the internalQuery?

Emby.Server.Implementations/Library/Search/SearchManager.cs
```diff
@@ -0,0 +1,454 @@
...
+        var requestedLimit = providerQuery.Limit ?? 100;
+
+        foreach (var provider in providers.Where(p => p.CanSearch(providerQuery)))
+        {
+            if (bestScores.Count >= requestedLimit || cancellationToken.IsCancellationRequested)
+            {
+                break;
+            }
+
+            await CollectFromProviderAsync(provider, providerQuery, searchTerm, bestScores, requestedLimit, cancellationToken).ConfigureAwait(false);
```
JPVenson: i am still not sure if we not should rather run this in parallel instead. Its not so bad if we overproduce items but this could quickly go n(O) instead we could have Sigma time.

Emby.Server.Implementations/Library/Search/SqlSearchProvider.cs
```diff
@@ -0,0 +1,200 @@
...
+            });
+
+            var top = await scored
+                .OrderByDescending(x => x.Score)
+                .ThenBy(x => x.Id)
+                .Take(limit)
+                .ToListAsync(cancellationToken)
+                .ConfigureAwait(false);
+
+            var results = new List<SearchResult>(top.Count);
```
JPVenson: i mean if you go so far to create this list with a fixed capacity, just use an array

Or just bake it into the linq as a clientside part

Emby.Server.Implementations/Library/Search/SqlSearchProvider.cs
```diff
@@ -0,0 +1,200 @@
...
+            foreach (var row in top)
+            {
+                results.Add(new SearchResult(row.Id, row.Score));
+            }
+
+            return results;
```
JPVenson:
```suggestion
            return await scored
                .OrderByDescending(x => x.Score)
                .ThenBy(x => x.Id)
                .Take(limit)
                .AsEnumerableAsync(cancellationToken)
                .Select(e => new Searchresult(e.Id, e.Score)
                .ToArrayAsync()
                .ConfigureAwait(false);
```

src/Jellyfin.Database/Jellyfin.Database.Implementations/JellyfinQueryHelperExtensions.cs
```diff
@@ -138,13 +224,13 @@ public static Expression<Func<TEntity, bool>> OneOrManyExpressionBuilder<TEntity
...
         var containsMethodInfo = _containsQueryCache.GetOrAdd(typeof(TProperty), static (key) => _containsMethodGenericCache.MakeGenericMethod(key));
 
-        if (oneOf.Count < 4) // arbitrary value choosen.
-        {
-            // if we have 3 or fewer values to check against its faster to do a IN(const,const,const) lookup
-            return Expression.Lambda<Func<TEntity, bool>>(Expression.Call(null, containsMethodInfo, Expression.Constant(oneOf), property.Body), parameter);
-        }
-
-        return Expression.Lambda<Func<TEntity, bool>>(Expression.Call(null, containsMethodInfo, Expression.Call(null, _efParameterInstruction.MakeGenericMethod(oneOf.GetType()), Expression.Constant(oneOf)), property.Body), parameter);
+        return Expression.Lambda<Func<TEntity, bool>>(
```
JPVenson: why this change?

Emby.Server.Implementations/Library/Search/SearchManager.cs
```diff
@@ -0,0 +1,454 @@
...
+
+    private static async Task<int> CollectFromExternalProviderAsync(
+        IExternalSearchProvider provider,
+        SearchProviderQuery providerQuery,
+        Dictionary<Guid, float> bestScores,
+        int requestedLimit,
+        CancellationToken cancellationToken)
+    {
+        var count = 0;
+        await foreach (var result in provider.SearchAsync(providerQuery, cancellationToken).ConfigureAwait(false))
```
JPVenson, line 330: nit: we _could_ also check here if we already gathered enough results according to the limiter in the calling methods to save time.

JPVenson (changes requested):

Emby.Server.Implementations/Library/Search/SearchManager.cs
Shadowghost: Web fetches 800 items, I had no issues so far with that. And the problem with paging through remote results is that this paging up to the requested size is quite slow because we need to run full item resolution for every page.

JPVenson: can you elaborate?

Shadowghost: If web requests 800 entries and we paginate on the provider side, we need to do the item fetching and post-filtering based on permissions for every batch until we either don't get any results form the provider or reach 800 entries. This means we run the expensive evaluation and DTO building for potentially hundreds of batches, increasing the response time

Emby.Server.Implementations/Library/Search/SearchManager.cs
Shadowghost: we can't because we do the ranking in memory afterwards

Emby.Server.Implementations/Library/Search/SearchManager.cs
Shadowghost: Fair point.

Emby.Server.Implementations/Library/Search/SqlSearchProvider.cs
Shadowghost: `AsEnumerableAsync` does not exist for this input

JPVenson: `AsAsyncEnumerable` https://learn.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.entityframeworkqueryableextensions.asasyncenumerable?view=efcore-10.0

Emby.Server.Implementations/Library/Search/SearchManager.cs
```diff
@@ -0,0 +1,443 @@
...
+    public async Task<IReadOnlyList<SearchResult>> GetSearchResultsAsync(
+        SearchProviderQuery query,
+        CancellationToken cancellationToken = default)
+    {
+        ArgumentNullException.ThrowIfNull(query);
+        ArgumentException.ThrowIfNullOrWhiteSpace(query.SearchTerm);
+
+        var searchTerm = query.SearchTerm.Trim().RemoveDiacritics();
+
+        var results = await CollectFromProvidersAsync(_externalProviders, query, searchTerm, cancellationToken).ConfigureAwait(false);
```
gallyamb: **question (blocking):** why do we allow _multiple_ search providers to be used at the same time? Don't we should configure exactly one search provider to be used? As a Jellyfin instance admin, I can't imagine a case where I can benefit from using multiple search providers simultaneously

But I see the drawbacks:
- the scores of the search providers are out of my (as an admin) control, so I cannot rely on ranking the search results from multiple providers => I'll configure exactly one preferred provider (but intention here will be implicit, instead of explicit configuration)
- some providers _may_ abuse this behavior to rank their search results above other providers (and not only from the start, but after a while with the next plugin update, which is more likely to be unnoticed)
- it's hard to implement search results' merging with user-based restrictions. If there is only one active search provider, then this became just a simple pagination until the whole search results' page is completed

JPVenson: If you just want one provider, install one.
If you want multiple providers, for example one that Searches the internal DB and one that searches Seer you can do that too.

Plugins may never be fully trusted, modifying the rank of search should be your least concerns.

gallyamb: > If you just want one provider, install one.

The internal DB search provider is always installed. There is no way currently in this PR to disable it, unless installing at least one external search provider

> If you want multiple providers

What is the case, where I may want multiple search providers? The described "internal + Serr" case does not related to the search, AFAIU: this PR is about searching the items that _already_ present in the Jellyfin, isn't it?

And even if current implementation allows to configure "internal + Seer" search to be configured, imo, this specific case should be handled in the other way: the Seer integration should either be native, or via specific SPI related to the discovering new media to enable new media requests specific actions

> modifying the rank of search should be your least concerns

Well, it's already not the only point 🙂 The main one is that there is no any way to make different plugins' results to be comparable between each other

JPVenson (approved):
Still strongly advice to parallelize providers

enter-a-random-username:
Could you resolve the conflict? Would love to test meilsearch as one of possible providers.

jhparker88:
I have a general question, now that you introduce a new search interface here I was wondering whether you allow for a plugin (`IExternalSearchProvider`, `SearchResult`) the provide more information other than the item ID and score.

For example, filtering via facets [1] would elevate the user search experience and given that meilisearch (and others like manticoresearch do as well) excel at providing **facet buckets** [2,3,4] for a search query (you don't want to retrieve facets via SQL from a DB) that the web interface could access via the API as filtering input.

My question is not about a specific implementation at this point but whether the interface design would allow for such additional information to be brought "back" in order for the API to provide those information in future?

[1] https://moma.meilisearch.com/
[2] https://www.meilisearch.com/docs/capabilities/filtering_sorting_faceting/how_to/filter_with_facets#search-with-facets
[3] https://manual.manticoresearch.com/Searching/Faceted_search#Faceted-search
[4] https://typesense.org/docs/30.2/api/search.html#facet-results

Shadowghost:
@jhparker88 in theory yes, the main issue for external providers is that they do not (and can't really) apply any access filtering, so I don't think facets, etc would provide that much value.

Emby.Server.Implementations/Library/Search/SearchManager.cs
gallyamb: Guys, lets go back to this thread, please. I believe it's a huge mistake to allow to configure multiple search providers. If after a while the community finds out that this is really needed, then adding this feature won't be difficult. But it'll be a challenge to restrict to only one search provider, if we allow this from the beginning

JPVenson (approved):

nyanmisaka (approved):

Emby.Server.Implementations/Library/Search/SearchManager.cs
```diff
@@ -0,0 +1,458 @@
...
+
+        var providerQuery = BuildProviderQuery(query);
+        var candidates = await GetSearchResultsAsync(providerQuery, cancellationToken).ConfigureAwait(false);
+        if (candidates.Count == 0)
+        {
+            return new QueryResult<SearchHintInfo>();
+        }
+
+        var candidateScores = BuildScoreLookup(candidates);
+        var user = !query.UserId.IsEmpty() ? _userManager.GetUserById(query.UserId) : null;
```
Bond-009:
```suggestion
        var user = query.UserId.IsEmpty() ? null : _userManager.GetUserById(query.UserId);
```

Jellyfin.Api/Controllers/ItemsController.cs
```diff
@@ -353,8 +386,8 @@ public async Task<ActionResult<QueryResult<BaseItemDto>>> GetItems(
                 Recursive = recursive ?? false,
                 OrderBy = RequestHelpers.GetOrderBy(sortBy, sortOrder),
                 IsFavorite = isFavorite,
-                Limit = limit,
-                StartIndex = startIndex,
+                Limit = searchResultScores is not null ? null : limit,
+                StartIndex = searchResultScores is not null ? null : startIndex,
```
Bond-009:
```suggestion
                Limit = searchResultScores is null ? limit : null,
                StartIndex = searchResultScores is null ? startIndex : null,
```

Bond-009 (approved):

PR #16123, opened 2026-01-27
Fix migration batching and folder refresh recursion

ZeusCraft10:
## Summary
- Batch library.db migration inserts to keep memory use steady and avoid OOMs on large libraries.
- Reload child lists safely, guard recursion, and skip parent/self cycles during folder metadata refresh to prevent stack overflows and redundant work.
- Let OOM/StackOverflow exceptions bubble up for accurate crash reporting.

## Related Issues
- #15101

## Test Plan
- dotnet build Jellyfin.Server/Jellyfin.Server.csproj


nielsvanvelzen:
Closing per https://jellyfin.org/docs/general/contributing/llm-policies/

PR #16124, opened 2026-01-27
Fix: Resolve CA1849 synchronous IO in EpubImageProvider

Gladtbam:
Fixes build errors related to CA1849 in .NET 10.0 builds.

**Changes**

- Replaced synchronous ZipFile.OpenRead with ZipFile.OpenReadAsync.

- Replaced synchronous ZipArchiveEntry.Open with ZipArchiveEntry.OpenAsync.

- Implemented await using for IAsyncDisposable resources (Streams and ZipArchive) to prevent leaks.

- Added ConfigureAwait(false) to async disposables to avoid capturing the sync context.

**Issues**

none


nielsvanvelzen:
Nice, seems to fix the issues introduced by #15604

crobibero (approved):

PR #16128, opened 2026-01-28
Fix: Prioritize existing filename.nfo over movie.nfo when saving metadata

### LINKED ISSUE #16105: [Regression] Manual .nfo files are ignored during scan (Re-occurs in 10.11.6)
### Description of the bug

I am experiencing the same issue described in the closed issue #15376. Although that issue is closed, the behavior has persisted or returned in version 10.11.6.

**Specific behavior**: Jellyfin fails to recognize my manually created .nfo files that are named identically to the movie files. Instead of reading the existing NFO, it seems to ignore it, momentarily generates a generic movie.nfo, and ultimately leaves the item with default/scraped metadata rather than the data provided in my NFO.

**Reference:**

Related closed issue: #15376

### Reproduction steps

1. Prepare a movie file in a folder (e.g., `MovieName (Year).mkv`).

2. Manually create an NFO file with the exact same name (e.g., `MovieName (Year).nfo`) containing custom metadata.

3. Add this folder to a Jellyfin library (ensure "Nfo" reader is enabled).

4. Wait for the Jellyfin server to finish the initial library scan.

5. **Observation 1**: The manual NFO is ignored. Surprisingly, a new `movie.nfo` file is generated in the folder (ignoring the sidecar NFO).

6。 Wait for the scheduled "Scan Media Library" task to run (or trigger it manually).

7. **Observation 2**: The generated `movie.nfo` disappears.

8. **Final Result**: The manual `MovieName (Year).nfo` is still ignored, and the movie metadata in Jellyfin remains as it was during the initial add (likely from online scrapers), failing to read the custom NFO.

The system ignores the valid, existing sidecar .nfo file. It temporarily creates a `movie.nfo` file which is later deleted, but the metadata from the original `.nfo` file is never applied to the media item.

Jellyfin should detect the existing `MovieName (Year).nfo`, parse it, and apply the metadata contained within it to the movie. It should not generate a duplicate `movie.nfo` or ignore the user-provided NFO file.

### Jellyfin Server version

10.11.6

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.6

### Environment

```markdown
- OS: Debian GNU/Linux 11 (bullseye)
- Linux Kernel: 5.10.0-26
- Virtualization: KVM
- Clients: Browser
- Browser: Microsoft Edge 43.0.3650.96
- FFmpeg Version: jjellyfin-ffmpeg7/unknown,now 7.1.3-1-bullseye
- Playback Method:
- Hardware Acceleration:
- CPU Model: AMD Ryzen 5 3600 6-Core Processor
- GPU Model: none
- Plugins: 
- Reverse Proxy: nginx/1.18.0
- Base URL: none
- Networking: Host
- Jellyfin Data Storage & Filesystem: local HDD - ext4
- Media Storage & Filesystem: Local HDD
- External Integrations:
```

### Jellyfin logs

```shell
none
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

Gladtbam:
**Changes**

Currently, saving metadata defaults to creating `movie.nfo` in non-mixed folders, ignoring existing file-specific NFOs. This change adds a check to prioritize the file-specific NFO (e.g., `filename.nfo`) if it already exists on disk. This prevents the creation of duplicate NFO files during metadata updates.

**Issues**

Fixes #16105

### DIFF
--- a/MediaBrowser.XbmcMetadata/Savers/MovieNfoSaver.cs
+++ b/MediaBrowser.XbmcMetadata/Savers/MovieNfoSaver.cs
@@ -52,6 +52,11 @@ namespace MediaBrowser.XbmcMetadata.Savers
                 yield return Path.Combine(path, "VIDEO_TS", "VIDEO_TS.nfo");
             }
 
+            if (File.Exists(Path.ChangeExtension(item.Path, ".nfo")))
+            {
+                yield return Path.ChangeExtension(item.Path, ".nfo");
+            }
+
             // only allow movie object to read movie.nfo, not owned videos (which will be itemtype video, not movie)
             if (!item.IsInMixedFolder && item.ItemType == typeof(Movie))
             {


Gladtbam:
> I don't see this as an issue that needs to be fixed, as this is the desired behavior approved in #10339

I understand the PR's point that some private torrent repositories contain custom NFO files, but I don't think that's a sufficient reason to lower the priority of `<filename>.nfo` files. In fact, many torrent filenames are quite messy and unsuitable for use in Jellyfin repositories. Many beginner guides also recommend using hard links and renaming files. If we ignore user-defined handling simply to avoid junk files in some cluttered folders, we're actually creating more problems for users who follow best practices and keep their repositories clean.

PR #16131, opened 2026-01-28
Allow setting http/https port via environment variable and cli flag

stavros-k:
Allows setting http/https port via environment variable and cli flag.
Makes it easy to adjust basic network settings, so you can always get back in the webui.
ie someone uses docker with hostnetwork but another service uses the port.

**Changes**

**Issues**

### DIFF
--- a/Emby.Server.Implementations/ApplicationHost.cs
+++ b/Emby.Server.Implementations/ApplicationHost.cs
@@ -444,13 +444,31 @@ namespace Emby.Server.Implementations
             }
 
             var networkConfiguration = ConfigurationManager.GetNetworkConfiguration();
-            HttpPort = networkConfiguration.InternalHttpPort;
-            HttpsPort = networkConfiguration.InternalHttpsPort;
+
+            // Check for port overrides from CLI/environment variables
+            var httpPortOverride = _startupConfig.GetValue<int?>(HttpPortKey);
+            var httpsPortOverride = _startupConfig.GetValue<int?>(HttpsPortKey);
+
+            HttpPort = httpPortOverride ?? networkConfiguration.InternalHttpPort;
+            HttpsPort = httpsPortOverride ?? networkConfiguration.InternalHttpsPort;
 
             // Safeguard against invalid configuration
             if (HttpPort == HttpsPort)
             {
                 HttpPort = NetworkConfiguration.DefaultHttpPort;
+                HttpsPort = NetworkConfiguration.DefaultHttpsPort;
+            }
+
+            // Validate port ranges
+            if (HttpPort < 1 || HttpPort > 65535)
+            {
+                Logger.LogWarning("HTTP port {Port} is outside valid range (1-65535), using default {Default}", HttpPort, NetworkConfiguration.DefaultHttpPort);
+                HttpPort = NetworkConfiguration.DefaultHttpPort;
+            }
+
+            if (HttpsPort < 1 || HttpsPort > 65535)
+            {
+                Logger.LogWarning("HTTPS port {Port} is outside valid range (1-65535), using default {Default}", HttpsPort, NetworkConfiguration.DefaultHttpsPort);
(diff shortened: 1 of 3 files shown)


Emby.Server.Implementations/ApplicationHost.cs
```diff
@@ -444,8 +444,13 @@ public void Init(IServiceCollection serviceCollection)
...
             var networkConfiguration = ConfigurationManager.GetNetworkConfiguration();
-            HttpPort = networkConfiguration.InternalHttpPort;
-            HttpsPort = networkConfiguration.InternalHttpsPort;
+
+            // Check for port overrides from CLI/environment variables
+            var httpPortOverride = _startupConfig.GetValue<int?>(HttpPortKey);
+            var httpsPortOverride = _startupConfig.GetValue<int?>(HttpsPortKey);
+
+            HttpPort = httpPortOverride ?? networkConfiguration.InternalHttpPort;
+            HttpsPort = httpsPortOverride ?? networkConfiguration.InternalHttpsPort;
```
dfederm, line 454: The safeguard below is a bit awkward and may lead to unexpected behavior, especially in this case with overrides as you may lose the override.

Example:
* User add `--http-port=9000`
* Config file has HTTPs port configure to 9000
* They're the same, so both revert to the defaults of 8096 and 8920

So even though the user was explicit about `--http-port=9000`, it doesn't happen.

Perhaps this should go after the check below? And then another check for equality which will just hard error or something?

One idea:

```cs
   if (HttpPort == HttpsPort)
   {
       if (httpPortOverride.HasValue && !httpsPortOverride.HasValue)
       {
           // User explicitly set HTTP, adjust HTTPS
           Logger.LogWarning("HTTPS port {Port} conflicts with HTTP port override, using default {Default}", HttpsPort, NetworkConfiguration.DefaultHttpsPort);
           HttpsPort = NetworkConfiguration.DefaultHttpsPort;
       }
       else if (httpsPortOverride.HasValue && !httpPortOverride.HasValue)
       {
           // User explicitly set HTTPS, adjust HTTP
           Logger.LogWarning("HTTP port {Port} conflicts with HTTPS port override, using default {Default}", HttpPort, NetworkConfiguration.DefaultHttpPort);
           HttpPort = NetworkConfiguration.DefaultHttpPort;
       }
       else
       {
           // Both overridden to same value or both from config - reset both
           Logger.LogWarning("HTTP and HTTPS ports are both {Port}, resetting to defaults", HttpPort);
           HttpPort = NetworkConfiguration.DefaultHttpPort;
           HttpsPort = NetworkConfiguration.DefaultHttpsPort;
       }
   }
```

Emby.Server.Implementations/ApplicationHost.cs
```diff
@@ -454,6 +459,19 @@ public void Init(IServiceCollection serviceCollection)
                 HttpsPort = NetworkConfiguration.DefaultHttpsPort;
             }
 
+            // Validate port ranges
```
dfederm: The validation happening after the equality check is also a bit confusing.

eg what if the user says `--http-port=0 --https-port=0`. The checks above will end up forcing the defaults of 8096/8920, which this will then validate and find no issue with.

Emby.Server.Implementations/ApplicationHost.cs
```diff
@@ -444,8 +444,13 @@ public void Init(IServiceCollection serviceCollection)
             }
 
             var networkConfiguration = ConfigurationManager.GetNetworkConfiguration();
-            HttpPort = networkConfiguration.InternalHttpPort;
-            HttpsPort = networkConfiguration.InternalHttpsPort;
+
+            // Check for port overrides from CLI/environment variables
+            var httpPortOverride = _startupConfig.GetValue<int?>(HttpPortKey);
```
dfederm: What's the behavior if the user sets the argument to something that's not an int? A proper error should be displayed.

PR #16062, opened 2026-01-19
Query Performance Improvements

Shadowghost:
This PR implements multiple performance and correctness improvements across the item query pipeline, library management, and database layer.

### Repository Split & Architecture
- **Split `BaseItemRepository`** into focused services: `ItemPersistenceService`, `NextUpService`, `ItemCountService`, `LinkedChildrenService` - each injected via `IItemQueryHelpers`
- **Introduce `DescendantQueryHelper`** to efficiently handle recursive ancestor/descendant lookups in DB
- **Add `FolderAwareFilterExtensions`** to push more filters (extras, trailers, subtitles, etc.) down to the DB query level instead of in-memory `PostFiltering`

### LinkedChildren Migration
- **Migrate `LinkedChildren` to its own database table** (`LinkedChildEntity`) with proper foreign keys
  - Allows safe management of boxsets, playlists, and video alternate versions
  - Deduplicate LinkedChildren entries
  - Remove the `CleanupCollectionAndPlaylistPathsTask` since this is now dealt with when the deletion happens

### OwnerId / Extras Normalization
- **Normalize `OwnerId` to GUID** with proper FK to `BaseItems`
  - Add navigation properties for `Owner` and `Extras`
  - Extras are now handled via this FK - `ExtraIds` column is dropped
  - Migration cleans up dangling/orphaned extras

### Alternative Versions Overhaul
- **Move alternative version handling into `LibraryManager`** with robust promotion/demotion logic
  - Migrate `PrimaryVersionId` from string to GUID
  - Proper rerouting of LinkedChildren on primary version change or removal
  - Handle stacking, deferred primary deletion, and metadata preservation on primary swap
  - Cleanup orphaned alternate versions and duplicate video items
  - Fix trickplay generation for parts and alt versions

### Query & DTO Optimizations
- **Batch user data fetching** in DtoService (`GetUserDataBatch`, `GetChildCountBatch`, `GetPlayedAndTotalCountBatch`) - prevents N+1 DB roundtrips
- **Optimize NextUp queries** with `GetNextUpEpisodesBatch` - batched episode lookups instead of per-series queries
- **Optimize latest items grouping** in `UserViewManager`, also expose Seasons (needs web changes)
- **Optimize search and collection grouping queries**
- **Optimize `SeriesDatePlayed` ordering**
- **Simplify `UserDataManager`** and remove unused private methods
- Skip metadata refresh for already-indexed items during validation (reduces post-scan time)
[...]

PR #16145, opened 2026-01-31
Consider user's library access permissions when querying genre for collections

### LINKED ISSUE #16135: Privacy/Security Issue , Collections library leaking genre
ssue Title: Metadata Leak: Restricted Library Genres Visible in Collections Filter for Unauthorized Users
Description There is a metadata leak in the "Collections" view. When a user navigates to the Collections library, the "Genre" filter dropdown populates with genres from libraries that the user does not have access to. This exposes metadata (genre tags) from private content to unauthorized users.

Environment :

Jellyfin Server Version: 10.11.6

Client used: Web Browser , Swiftfin

Steps to Reproduce

Phase 1: Setup (Admin Actions)

Create Two Libraries:

Library Public: Add a movie (e.g., Toy Story) and tag it with the genre "Animation".

Library Private: Add a movie (e.g., Saw) and tag it with a unique genre that exists only in this library, e.g., "Gorefest".

Create Two Users:

User A (Admin/Owner): Has access to All Libraries (Public + Private).

User B (Restricted): Uncheck access to "Library Private". Ensure they only have access to "Library Public".

Create a Collection (The Variable):

Log in as User A.

Create a Collection named "Test Collection" or any TMDB collection can be used for reference

Add items from both "Library Public" and "Library Private" to this collection. (This ensures the Collection entity is linked to the restricted metadata).

Phase 2: Reproduction (The Exploit)

Log out of User A.

Log in as User B (The restricted user).

Navigate to the Collections Library (usually found on the Home Screen or via the sidebar).

User B will see "Test Collection" (likely showing only the Toy Story cover art, as Saw is hidden).

The Trigger: Click on the Filter button (often a funnel icon) at the top of the screen.

Look at the "Genre" dropdown list.

Expected Behavior
The Genre dropdown for User B should only display "Animation." Since User B has zero access to "Library Private," the system should sanitize all metadata derived from that library. "Gorefest" should not exist in User B's world.

Actual Behavior
The Genre dropdown displays both "Animation" and "Gorefest". Even though User B cannot see the movie Saw, the system is leaking the existence of the "Gorefest" genre through the Collections filter logic.

dfederm:
Fixes #16135

In `FilterController.GetQueryFilters()`, when querying for global item types (BoxSet, Playlist, Trailer, Program), the code set `parentItem = null`. This resulted in `genreQuery.AncestorIds =  Array.Empty<Guid>()`, which queried ALL libraries regardless of user permissions.

This change fixes the issue by making it so that when `parentItem is null` but the user exists, restrict the query to only the user's accessible library folders.

### DIFF
--- a/Jellyfin.Api/Controllers/FilterController.cs
+++ b/Jellyfin.Api/Controllers/FilterController.cs
@@ -151,16 +151,16 @@ public class FilterController : BaseJellyfinApiController
             ? null
             : _userManager.GetUserById(userId.Value);
 
-        BaseItem? parentItem = null;
-        if (includeItemTypes.Length == 1
+        // For global item types (BoxSet, Playlist, Trailer, Program), we don't use a parent item
+        // but we still need to respect user library access permissions
+        var isGlobalItemType = includeItemTypes.Length == 1
             && (includeItemTypes[0] == BaseItemKind.BoxSet
                 || includeItemTypes[0] == BaseItemKind.Playlist
                 || includeItemTypes[0] == BaseItemKind.Trailer
-                || includeItemTypes[0] == BaseItemKind.Program))
-        {
-            parentItem = null;
-        }
-        else if (parentId.HasValue)
+                || includeItemTypes[0] == BaseItemKind.Program);
+
+        BaseItem? parentItem = null;
+        if (!isGlobalItemType && parentId.HasValue)
         {
             parentItem = _libraryManager.GetItemById<BaseItem>(parentId.Value);
         }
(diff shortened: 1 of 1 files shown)


dfederm:
I'm actually having difficulty reproing the issue now. Putting this on hold until I have more confidence in the fix.

theguymadmax:
I agree, I tried to reproduce the issue I had confirmed, but couldn’t 🙂. That said, the issue I had in mind when I confirmed it was that filters were still showing genres/tags for blocked media (via parental ratings or blocked tags) in the library, even though the media itself was blocked. However, that issue will be fixed in #16062.

PR #16164, opened 2026-02-01
Cache rules of .ignore files

TheMelmacian:
Added a simple cache for ignore rules to prevent unnecessary re-creations of the same rule set over and over on each IsIgnored check.
This has massively reduced the scan times of one of my libraries that makes heavy use of a large .ignore rule set.
(see linked issue for more info).

**Changes**
- added a dictionary that functions as a simple cache for ignore rules
- after the content of an .ignore file is checked for ignore-rules, the resulting rule list is saved together with the path and change date of the .ignore file
- on subsequent checks the change date of the .ignore file is compared to the one in the cache and if no there are no changes the rule set is loaded from the cache
- if the file changed, the rules in the cache are updated

**Issues**
Fixes #16163

PR #16166, opened 2026-02-01
Implement ignore rule caching

### LINKED ISSUE #16163: Scan of large folder trees in combination with .ignore file(s) with a very long rule set is extremely slow
### Description of the bug

I have a very large music library that is currently not in an optimal state for jellyfin. So i put a single .ignore file in the root folder that contains all folders (~600) that are not ready for scanning (folder structure, metadata etc.).
The initial scan took a very long time until all not ignored artists and albums were added to the library. Updates are also very slow.

Problem is that ignore rules are currently not cached. Each IsIgnored check needs to read the content of the .ignore file again and create the list of ignore rules. Adding a logging output with the needed time for an ignore check reveals most checks need up to 1000ms.

`info: Emby.Server.Implementations.Library.DotIgnoreIgnoreRule[0] --->>> IsIgnored took 00:00:00.9593885`

Adding a simple caching mechanism for the rules reduced the time after the first read significantly:

`info: Emby.Server.Implementations.Library.DotIgnoreIgnoreRule[0] --->>> IsIgnored took 00:00:00.0004388`

### Reproduction steps

Have a large folder tree and put a single .ignore file in the root folder that includes rules for a large number of folders.

Each IsIgnored check needs to read the content of the .ignore file and creates a new list of ignore rules.

Ignore rules are cached and re-used as long as the .ignore file hasn't changed.

### Jellyfin Server version

10.11.6

### Specify the build version

10.11.6

### Environment

```markdown
- OS: Debian 13
- Linux Kernel: Linux 6.12.63
- Virtualization: Docker
- CPU Model: AMD Ryzen 5 PRO 4650G
- GPU Model: AMD Ryzen 5 PRO 4650G
- Jellyfin Data Storage & Filesystem: NVMe M.2 SSD / BTRFS
- Media Storage & Filesystem: HDD / EXT4
```

Shadowghost:
Implements Ignore rule caching.

Alternative to #16164

**Changes**
* Implement file and ignore rule caching with `FastConcurrentLru`

**Issues**
Fixes #16163

### DIFF
--- a/Emby.Server.Implementations/Library/DotIgnoreIgnoreRule.cs
+++ b/Emby.Server.Implementations/Library/DotIgnoreIgnoreRule.cs
@@ -1,6 +1,7 @@
 using System;
 using System.IO;
 using System.Text.RegularExpressions;
+using BitFaster.Caching.Lru;
 using MediaBrowser.Controller.Entities;
 using MediaBrowser.Controller.IO;
 using MediaBrowser.Controller.Resolvers;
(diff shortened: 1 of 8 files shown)


Emby.Server.Implementations/Library/DotIgnoreIgnoreRule.cs
```diff
@@ -130,11 +134,203 @@ internal static bool CheckIgnoreRules(string path, string[] rules, bool isDirect
         return ignore.IsIgnored(pathToCheck);
     }
 
-    private static string GetFileContent(FileInfo ignoreFile)
+    /// <summary>
+    /// Clears the internal caches.
+    /// </summary>
+    public void ClearCache()
```
JPVenson: never called?

JPVenson (commented):
Should't there be _some_ kind of eviction mechanism? I would understand if we aggressively cache the value for a scan or longer when we are actually doing the stupid shit like re and re and rechecking directory and such but fire-and-forget can get very annoying if somebody would actually use the ignore files and tries to edit them no?

Emby.Server.Implementations/Library/DotIgnoreIgnoreRule.cs
```diff
@@ -15,63 +16,66 @@ public class DotIgnoreIgnoreRule : IResolverIgnoreRule
...
+    {
+        var cacheSize = Math.Max(100, Environment.ProcessorCount * 100);
+        _directoryCache = new FastConcurrentLru<string, IgnoreFileCacheEntry>(
+            Environment.ProcessorCount,
+            cacheSize,
+            StringComparer.OrdinalIgnoreCase);
+        _rulesCache = new FastConcurrentLru<string, ParsedIgnoreCacheEntry>(
+            Environment.ProcessorCount,
+            Math.Max(32, cacheSize / 4),
+            StringComparer.OrdinalIgnoreCase);
```
TheMelmacian: On unix-like operating systems file names and paths are case sensitive:

```suggestion
        _directoryCache = new FastConcurrentLru<string, IgnoreFileCacheEntry>(
            Environment.ProcessorCount,
            cacheSize,
            StringComparer.Ordinal);
        _rulesCache = new FastConcurrentLru<string, ParsedIgnoreCacheEntry>(
            Environment.ProcessorCount,
            Math.Max(32, cacheSize / 4),
            StringComparer.Ordinal);
```

JPVenson (approved):

crobibero (approved):

Issue #16190, opened 2026-02-06
AndroidTV client overlay fails to start recording when clicking record button.


WizardOfYendor1:
### Description of the bug

I encountered this issue in the course of developing a plugin of my own and chased down what I believe is the underlying problem(s). The clients using the API are more directly at fault but the server side may also be. 

I've created this issue here instead of the androidtv because I have a PR that mitigates (some would say mask) the problem. Not only for AndroidTV but probably other clients also. 

I need to say I am a total noob to this entire codebase ecosystem. So it's very hard for me to say where the "fault(s)" lie. I think there a mix of fixes in different places but I do not have enough understanding to decide what is the best. i.e. alot of this below is my noob understanding could definitely be wrong.

Scenario: on an AndroidTV client if you are watching LiveTV, hit the up arrow on the remote, in the display/playback overlay menu (or whatever it's called), hit the record button. I get an "unable to create recording" error. It sounds alot like this reported previously: https://github.com/jellyfin/jellyfin-androidtv/issues/4371

So I added some "temporary" extra logging in the server code locally to see what the POST body was. Saw the dates with 0000 years - which was odd. 
{
  "Type": "SeriesTimer",
  "StartDate": "0000-12-31T19:03:58-04:56:02",
  "EndDate": "0000-12-31T19:03:58-04:56:02",

  Blah blah. 
}

On the server, who's now consuming the POST message, the function JsonDateTimeConverter calls DatetTime.TryParse() and fails with a year of 0000. It has to be a minimum of year 0001. This is the "victim" but I believe the origin is in the SDK codebase.

There's a WHOLE lot more to this. It's been a helluva journey to figure out and reinforces my hatred of date implementations in languages :-),  - so  I'm just going to cut to the chase:

I BELIEVE the root  problem is deep in the SDK. 
Looking at this code: https://github.com/jellyfin/jellyfin-sdk-kotlin/blame/master/jellyfin-model/src/jvmMain/kotlin/org/jellyfin/sdk/model/serializer/DateTimeSerializer.kt

And this recent commit(I DO NOT believe this introduced the problem): https://github.com/jellyfin/jellyfin-sdk-kotlin/commit/2477a5a823beb046718508c80ef358eba8c7d84b

@nielsvanvelzen 
I think this is the underlying issue:
encoder.encodeString(value.atZone(ZoneId.systemDefault()).format(DateTimeFormatter.ISO_OFFSET_DATE_TIME))

[...]

PR #16191, opened 2026-02-06
Handle year 0000 dates in JsonDateTimeConverter

WizardOfYendor1:
Clients, like AndroidTV in particular, are submitting date strings with a year of 0000 which .Net will not parse (0001 is the min). This is occuring due to timezone offsets or other date math issues elsewhere that submit a year of 0000 in JSON payloads.,

It would seem ISO-8601 permits 0000 years,  Java.time (thus Kotlin) permits 0000 years, C#.Net uses gregorian calendar which only allows dates starting at year 0001. 

**Changes**
Put regex pattern matching into C#.net date parsing to make an assumption that if the year is 0000, what you really want is the minimum date possible from .Net. 

**Issues**
https://github.com/jellyfin/jellyfin/issues/16190
https://github.com/jellyfin/jellyfin-androidtv/issues/4371

### DIFF
--- a/src/Jellyfin.Extensions/Json/Converters/JsonDateTimeConverter.cs
+++ b/src/Jellyfin.Extensions/Json/Converters/JsonDateTimeConverter.cs
@@ -14,7 +14,22 @@ namespace Jellyfin.Extensions.Json.Converters
         /// <inheritdoc />
         public override DateTime Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
         {
-            return reader.GetDateTime();
+            if (reader.TryGetDateTime(out var result) )
+            {
+                return result;
+            }
+            else
+            {
+                // it failed, let's look at the string value, and if it looks like a zero date, return DateTime.MinValue
+                var text = reader.GetString();
+                if (text != null && System.Text.RegularExpressions.Regex.IsMatch(text, @"^0000-\d{2}-\d{2}"))
+                {
+                    return DateTime.MinValue;
+                }
+
+                // We tried. Let's just throw the original exception for backwards compatiblity.
+                return reader.GetDateTime();
+            }
         }
 
         /// <inheritdoc />


crobibero:
I don't see a valid reason why the server should accept `0000` as a year. This isn't a usable date so either way it will need to be fixed on the client.

WizardOfYendor1:
> I don't see a valid reason why the server should accept `0000` as a year. This isn't a usable date so either way it will need to be fixed on the client.


You have a valid point. I actually wrestled with this change quite a bit because it is "hacky". Part of my angst is that it only traps for one weird scenario where the year is 0000 but callers could be sending all kinds of perfectly valid, crazy date formats that C# won't accept (min year = 0001). Adding to that, this is almost like an abstracted validator. The proper place would be "higher up" in the API stack.

On the pro side, this is a minimally invasive change even though it's only catching one particular scenario. Namely, **when negative UTC offsets are applied to the .Net minimum date.** When that happens, they "rollback" to year 0000. e.g.if you're in Europe somewhere - you're good; Americas not so much :-). 

At least this is how I understand this world... and I have to admit my head is spinning crawling across the code base with different languages datetime implementations/usages in the mix. 

Here are some of my related findings: 
In basetimerinfodto, there are getter and setter for Start/End DateTime. Those are NON NULLABLE... that, as I understand it, means it automatically gets set to the C# DateTime.MinValue (the 0001-01-01 date). So if a client sends in a request with missing dates - that's what your get. Like if you were create a defaultimer or something for a recording with no datetimes. 


So in this particular case, this is what I **think** is happening. The Android TV client created a default timer.... because *I'm* in a -5 offset, the SDK applied that offset to the DateTime.MinValue - "creating" a date with a year of 0000 - which .Net can't handle. 

The "flow" of this problem basically looks like this:

- The androidTV client creates a timer without datetimes. The API/server sticks in C# min dates because it didn't receive any.
- The Server sends "0001-01-01T00:00:00.0000000Z" (MinValue with Z).
- Java successfully parses it as a ZonedDateTime in UTC
- The client calls .withZoneSameInstant(systemDefault) converts from UTC midnight year 0001 to the local timezone
- **For any timezone west of UTC, this rolls back into year 0000** (December 31st)
- The historical offset for that timezone at year 0001 produces bizarre values like -04:56:02 (don't know why)
- Client sends back "0000-12-31T19:03:58-04:56:02" — which is technically correct math, but .NET can't represent year 0000 nor will it understand the seconds of the offset anyways - I believe
- Server returns 400
- The catch statement (DateTimeParseException) in the SDK is a red herring — it's never reached for this case. The SDK serialize path needs to check for year 0000 or year 0001 in the converted LocalDateTime, not for LocalDateTime.MIN. 
_That is to say, it needs to KNOW the server has the C# minimum date and guard against it._

I'll also add, I believe the server will always send back a string like 0001-01-01T00:00:00.0000000Z for MinDate so the check/test in the SDK wouldn't be hit. But I could be wrong. 

As an aside, I started going down another rabbit hole looking at the DefaultLiveTvService. There are places like at like 96 where if there's the "default" 0001-01-01 set on timer.EndDate - it will just quit. But I feel way in over my head in this rabbit hole so I'll probably stop :-)

I asked Claude to search for the default date places. I'll include them here in case they are valuable to anyone:

### High Risk — Frequently Hit, Null DB Values → MinValue

| Property | Source | Code |
|----------|--------|------|
| `BaseItemDto.DateCreated` | `BaseItemRepository.cs:887` | `entity.DateCreated ?? DateTime.MinValue` |
| `BaseItemDto.DateModified` | `BaseItemRepository.cs:888` | `entity.DateModified ?? DateTime.MinValue` |
| `BaseItemDto.DateLastRefreshed` | `BaseItemRepository.cs:890` | `entity.DateLastRefreshed ?? DateTime.MinValue` |
| `BaseItemDto.DateLastSaved` | `BaseItemRepository.cs:891` | `entity.DateLastSaved ?? DateTime.MinValue` |
| `BaseItemDto.DateLastMediaAdded` | `BaseItemRepository.cs:994` | `entity.DateLastMediaAdded ?? DateTime.MinValue` |

### Medium Risk — LiveTV Specific

| Property | Source | Code |
|----------|--------|------|
| `ProgramInfo.EndDate` | `LiveTvManager.cs:1028` | `program.EndDate ?? DateTime.MinValue` |
| `BaseTimerInfoDto.StartDate/EndDate` | `BaseTimerInfoDto.cs:73-78` | Non-nullable, defaults to `default(DateTime)` |
| `NextUpQuery.NextUpDateCutoff` | `NextUpQuery.cs:15` | Explicitly set to `DateTime.MinValue` in constructor |


In summary, the only reason I suggest this narrow "fix" is to "keep the peace" with the ecosystem. This narrowly infers some poor client somewhere has applied a UTC offset to the default min datetime the server sent back. I can see that being a rather innocent assumption by client code - take the server's UTC date and "make it mine". 

From a software architecture approach - I fully understand why this PR would be rejected. A bigger picture approach would be warranted. e.g. Perhaps the API should allow those datetimes to be nullable? e.g. DateTime? EndDate etc.. Perhaps the API guards what it's receiving and throwing a meaningful error back? Etc....

I'll leave it to the folks here to make the call. I hope this helps anyways.

Bond-009:
Sorry, bit this "solution" is too hacky. We should improve/add validation to prevent creating timers with nonsensical datetimes instead.

PR #16193, opened 2026-02-06
Add tmdb id filter on Item api endpoint

### LINKED ISSUE #16192: Add the possibility to query item based on tmdbid in API
Allow api users to query a specific tmdbid in the get Item endpoint.
The query should return only one Item, or nothing.

Noustaa:
Add a tmdb id filter in Item API endpoint.
You can search for an item by tmdbid.

**Changes**
Added a new field "tmdbId" in GetItems function.

**Issues or Discussions**
#10156
Resolves #16192

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -211,6 +211,7 @@
  - [martenumberto](https://github.com/martenumberto)
  - [ZeusCraft10](https://github.com/ZeusCraft10)
  - [MarcoCoreDuo](https://github.com/MarcoCoreDuo)
+ - [Noustaa](https://github.com/Noustaa)
 
 # Emby Contributors
 
--- a/Emby.Server.Implementations/Localization/LocalizationManager.cs
+++ b/Emby.Server.Implementations/Localization/LocalizationManager.cs
@@ -262,6 +262,24 @@ namespace Emby.Server.Implementations.Localization
         }
 
         /// <inheritdoc />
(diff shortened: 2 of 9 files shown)


nielsvanvelzen (changes requested):
We should stop adding individual parameters for each external provider of and come up with a solution that works for all of them.

Noustaa:
> We should stop adding individual parameters for each external provider of and come up with a solution that works for all of them.

Good point ! I decided to pair it with the existing fields "Has<Provider>Id" to avoid adding a new field "Provider".

sjorge:
Would be nicer if you could look up any providerId not just those three. Although I am not sure how core-team feels about other plugin provided providers IDs like anidb, anilist, ...

Noustaa:
I updated to support all providers defined in MediaBrowser.Model.Entities.MetadataProvider.
Regarding plugin providers, i don't really know how to handle that.

What i have tested myself so far: Tmdb, Tvdb, Imdb and TmdbCollection. I can't test the others as i do not have them in my library.

Noustaa:
@nielsvanvelzen Are you okay with latest changes ?

nielsvanvelzen (changes requested):
The current approach is confusing (combination of 2 parameters to do a single thing) does not allow for using multiple provider ids. From a client-dev perspective I would probably expect to set my query parameters like `providerId[tmdb]=123&providerId[imdb]=567`. I believe you can achieve that with a dictionary in asp.net.

PR #16204, opened 2026-02-08
Fix broken library subtitle download settings

MBR-0001:
It seems somewhere between 10.9 and 10.10 the change was made to use 639-2/T language names but the options.xml kept old values. This resulted in following behavior:
if options.xml contains **dut** it will download subtitles, save them under filename.**dut**.srt but in MediaStreamInfos table it will save the language as **nld**, result -> subsequent scans will not find dut in MediaStreamInfos and will download subs again and again and again

Also, some users have duplicated entries https://github.com/jellyfin/jellyfin-plugin-opensubtitles/issues/185#issuecomment-3849609675

**Changes**
Added a migration to remove duplicate entries and convert all 639-2/B languages to 639-2/T in options.xml of every library. 

**Issues**
Maybe (hopefully)
fixes https://github.com/jellyfin/jellyfin/issues/12434
fixes https://github.com/jellyfin/jellyfin-plugin-opensubtitles/issues/185

### DIFF
--- a/Jellyfin.Server/Migrations/Routines/FixLibrarySubtitleDownloadLanguages.cs
+++ b/Jellyfin.Server/Migrations/Routines/FixLibrarySubtitleDownloadLanguages.cs
@@ -1,0 +1,106 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Jellyfin.Server.ServerSetupApp;
+using MediaBrowser.Controller.Entities;
+using MediaBrowser.Controller.Library;
+using MediaBrowser.Model.Globalization;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.Logging;
+
+namespace Jellyfin.Server.Migrations.Routines;
+
+/// <summary>
+/// Migration to fix broken library subtitle download languages.
+/// </summary>
+[JellyfinMigration("2026-02-06T20:00:00", nameof(FixLibrarySubtitleDownloadLanguages))]
+internal class FixLibrarySubtitleDownloadLanguages : IAsyncMigrationRoutine
+{
+    private readonly ILocalizationManager _localizationManager;
+    private readonly ILibraryManager _libraryManager;
+    private readonly ILogger _logger;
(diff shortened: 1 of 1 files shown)


crobibero (approved):

michnovka:
I confirm this fix resolves the issue for me locally :)

Jellyfin.Server/Migrations/Routines/FixLibrarySubtitleDownloadLanguages.cs
```diff
@@ -0,0 +1,101 @@
...
+
+            var fixedLanguages = new List<string>();
+
+            foreach (var language in options.SubtitleDownloadLanguages)
+            {
+                var foundLanguage = _localizationManager.FindLanguageInfo(language)?.ThreeLetterISOLanguageName;
+                if (foundLanguage is not null)
+                {
+                    // Converted ISO 639-2/B to T (ger to deu)
+                    if (foundLanguage != language)
```
Bond-009: use string.Equals

Jellyfin.Server/Migrations/Routines/FixLibrarySubtitleDownloadLanguages.cs
```diff
@@ -0,0 +1,101 @@
...
+
+                    fixedLanguages.Add(foundLanguage);
+                }
+                else
+                {
+                    _logger.LogInformation("Could not resolve language '{Language}' in library '{LibraryName}'. Skipping.", language, virtualFolder.Name);
+                }
+            }
+
+            // Distinct removes duplicates (spa, pob etc.)
```
Bond-009: I'd prefer checking this while adding them instead of at the end with distinct

Bond-009 (changes requested):

Jellyfin.Server/Migrations/Routines/FixLibrarySubtitleDownloadLanguages.cs
MBR-0001: done

Shadowghost (approved):

PR #16211, opened 2026-02-09
Add in new Audiobook Provider

bfordham:
Add provider to provider richer metadata for audio books.

**Changes**
New Provider to give better metadata for audio books.

### DIFF
--- a/MediaBrowser.Providers/Audiobooks/AudiobookImageProvider.cs
+++ b/MediaBrowser.Providers/Audiobooks/AudiobookImageProvider.cs
@@ -8,116 +8,115 @@ using MediaBrowser.Controller.Providers;
 using MediaBrowser.Model.Entities;
 using Microsoft.Extensions.Logging;
 
-namespace MediaBrowser.Providers.Audiobooks
+namespace MediaBrowser.Providers.Audiobooks;
+
+/// <summary>
+/// Provides the primary image for audiobook items that have embedded covers.
+/// </summary>
+public class AudiobookImageProvider : IDynamicImageProvider
 {
+    private readonly ILogger<AudiobookImageProvider> _logger;
+
     /// <summary>
-    /// Provides the primary image for audiobook items that have embedded covers.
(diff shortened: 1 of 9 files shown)


MediaBrowser.Providers/Audiobooks/AudiobookImageProvider.cs
```diff
@@ -0,0 +1,123 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using MediaBrowser.Controller.Entities;
+using MediaBrowser.Controller.Providers;
+using MediaBrowser.Model.Entities;
+using Microsoft.Extensions.Logging;
+
+namespace MediaBrowser.Providers.Audiobooks
```
JPVenson: please use file scoped namespaces and everywhere else too

bfordham: Thanks, done

MediaBrowser.Providers/Audiobooks/AudiobookTagReader.cs
```diff
@@ -0,0 +1,317 @@
...
+
+    private string CleanTitle(string title)
+    {
+        if (string.IsNullOrEmpty(title))
+        {
+            return "Unknown Title";
+        }
+
+        // Remove common patterns like [Unabridged], (Unabridged), etc.
+        var cleanTitle = System.Text.RegularExpressions.Regex.Replace(
```
JPVenson: i dont think its good to have those regexes being rebuild every time esp for larger lookups. better to have them precompiled

JPVenson (changes requested):
apart from the regexes LGTM

MediaBrowser.Providers/Audiobooks/AudiobookTagReader.cs
bfordham: Great call out, updated

JPVenson (approved):

PR #16226, opened 2026-02-12
Deduplicate provider IDs during MigrateLibraryDb migration

dfederm:
Fixes #16134

The `MigrateLibraryDb` migration parses the legacy `ProviderIds` column (a pipe-delimited string like `Tmdb=12345|Imdb=tt0000123`) into individual `BaseItemProvider` entities keyed by `(ItemId, ProviderId)`.

If the legacy string contains duplicate provider keys within the same row (e.g. `Tmdb=12345|Imdb=tt0000123|Tmdb=67890`), the parser produces duplicate entities that violate the primary key constraint, causing the migration to fail with "UNIQUE constraint failed: BaseItemProviders.ItemId, BaseItemProviders.ProviderId" and preventing server startup.

This adds `.DistinctBy(e => e.ProviderId)` to the parsing pipeline, keeping the first occurrence of each provider and discarding duplicates. This is defensive migration code that runs once during upgrade.

PR #16227, opened 2026-02-12
Reattach user data after item removal during library scan

### LINKED ISSUE #16149: Watch state lost when replacing episode file if both versions exist temporarily
### Description of the bug

After the fix for file rename/replacement in 10.11.6, watch state is preserved when an episode file is deleted before the new file is added. #15001 #15899

However, there is still a scenario where watch state is lost:

If the new episode file is added while the old file still exists (even temporarily), Jellyfin creates two media items for the same episode, and the newer file shows as unwatched, even if the old one was previously watched. When the old (watched) file is later deleted, the episode shows as unwatched.

This indicates that user data (watched status) remains attached to the original media item identity and is not consolidated when multiple versions coexist and one version is later removed.

The previously fixed rename/replace scenario works correctly when the replacement occurs via deletion before processing a new file.

### Reproduction steps

1. Sign into Jellyfin as a user.
2. Mark an episode as watched
3. Add a new version of the same episode without deleting the old one.
4. Trigger library scan/wait for automatic detection
5. Observe Jellyfin now detects two media items:
- Old file played = True
- New file played = False
6. Delete the original watched file
7. Trigger another library scan/wait for automatic detection

After deleting the original watched file, the episode becomes unwatched. The unwatched status does not transfer to the remaining file, even though it represents the same episode. 

However, if the old file is deleted before the new file is added, the watched state is preserved correctly

Replacing an episode file should preserve watch status regardless of if the old episode is deleted after the new file has been added.

If the old version was watched and the new version replaces it, the resulting episode should remain watched.

### Jellyfin Server version

10.11.6

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.6

### Environment

```markdown
- OS: Ubuntu 24.04.3 LTS
- Linux Kernel: 6.14.0-37-generic
- Virtualization: Docker
- Clients: All
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- CPU Model: 
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL: 
- Networking:
- Media Storage & Filesystem: local HDD - ext4
- External Integrations:
```

### Jellyfin logs

```shell
No relevant log output
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

dfederm:
Fixes #16149

When an episode file is replaced and both versions temporarily coexist, watch state is lost. This occurs because:
1. When the new file is first scanned, `ReattachUserDataAsync` runs during its first metadata refresh but finds nothing - the old file's user data hasn't been detached yet.
2. When the old file is later deleted, `DeleteItem` detaches its user data to the placeholder. But the replacement file is now an existing item, so `ReattachUserDataAsync` never runs again.

This fix adds a reattachment step in `ValidateChildrenInternal2` immediately after the item removal loop. It collects the user data keys of deleted items and checks if any remaining valid children share those keys (eg. same episode identified by series provider ID + season/episode number). If a match is found, `ReattachUserDataAsync` is called to transfer the detached user data to the replacement item.

The reattachment only runs when items were actually removed, minimizing performance impact during normal scans.

### DIFF
--- a/MediaBrowser.Controller/Entities/Folder.cs
+++ b/MediaBrowser.Controller/Entities/Folder.cs
@@ -452,6 +452,7 @@ namespace MediaBrowser.Controller.Entities
                 // That's all the new and changed ones - now see if any have been removed and need cleanup
                 var itemsRemoved = currentChildren.Values.Except(validChildren).ToList();
                 var shouldRemove = !IsRoot || allowRemoveRoot;
+                var actuallyRemoved = new List<BaseItem>();
                 // If it's an AggregateFolder, don't remove
                 if (shouldRemove && itemsRemoved.Count > 0)
                 {
@@ -467,6 +468,7 @@ namespace MediaBrowser.Controller.Entities
                         {
                             Logger.LogDebug("Removed item: {Path}", item.Path);
 
+                            actuallyRemoved.Add(item);
                             item.SetParent(null);
                             LibraryManager.DeleteItem(item, new DeleteOptions { DeleteFileLocation = false }, this, false);
                         }
@@ -476,6 +478,20 @@ namespace MediaBrowser.Controller.Entities
                 if (newItems.Count > 0)
                 {
                     LibraryManager.CreateItems(newItems, this, cancellationToken);
+                }
+
+                // After removing items, reattach any detached user data to remaining children
+                // that share the same user data keys (eg. same episode replaced with a new file).
+                if (actuallyRemoved.Count > 0)
+                {
+                    var removedKeys = actuallyRemoved.SelectMany(i => i.GetUserDataKeys()).ToHashSet();
+                    foreach (var child in validChildren)
+                    {
+                        if (child.GetUserDataKeys().Any(removedKeys.Contains))
+                        {
+                            await child.ReattachUserDataAsync(cancellationToken).ConfigureAwait(false);
+                        }
+                    }
                 }
             }
             else


JPVenson:
should't that target 10.11.z as a bugfix?

dfederm:
I'm a somewhat new contributor to the repo, but I assumed based on experience in other repos that fixes go to mainline and then are backported if deemed necessary.

If the preferred workflow is to do the inverse (PR in release branch, then forward-port thr change) I can retarget, although that does put the onus on the PR author whether it should be a patch fix/hotfix or not, and at least for an external contributor I'm not sure I'm really the one to make that decision definitively.

Unless that question wasn't directed at me, in which case disregard.

JPVenson:
ah no thats alr.

We backport changes if nessesary from the release branches to master. Target the current release branch "10.11.z" and add the "stable-backport" label

theguymadmax:
The workflow is somewhat reversed here, we backport from the release branch to the master.  As a rule,  new features / behavioral changes should always target the master and bugfixes should target the release branch, which then gets backported to the master.  Since this is unclear to most new contributors, we generally advise to always target the master and then if asked re-target .  With that please re-target this PR and https://github.com/jellyfin/jellyfin/pull/16226 to the release branch. Thanks.

dfederm:
Done. Doesn't look like I have permissions to add the `stable-backport` label though, so I assume that's something y'all do.

I would suggest considering changing your process to backporting from master -> release. That way changes make it into the next version at least in case there's some debate as to risk for a patch fix. And there's also less risk of the fix being lost and regressing if the port from release -> master doesn't end up happening. Ofc, y'all have the experience with managing the releases for Jellyfin and I don't, so take this just for what it is; just a suggestion.

PR #16228, opened 2026-02-12
Prevent full library scan when real-time monitoring detects movie folder changes

### LINKED ISSUE #16172: Individual movie directories are not tracked as separate DB entries, breaking real-time monitoring for movie libraries.
### Description of the bug

Currently Jellyfin tracks the following 2 entities for movies in the DB.

- `MediaBrowser.Controller.Entities.Movies.Movie` which includes the absolute path to a movie file
- `MediaBrowser.Controller.Entities.Movies.BoxSet` which includes the path to the collection

https://github.com/jellyfin/jellyfin/tree/master/MediaBrowser.Controller/Entities/Movies

This is different behaviour compared to what is tracked for shows in the DB with distinct DB entries for a show being:

- `MediaBrowser.Controller.Entities.TV.Series` which tracks the path to the show folder
- `MediaBrowser.Controller.Entities.TV.Season` which tracks the path to each season folder for a show
- `MediaBrowser.Controller.Entities.TV.Episode` which tracks the absolute path to the file for each episode of a show

https://github.com/jellyfin/jellyfin/tree/master/MediaBrowser.Controller/Entities/TV

The information tracked in the DB for movies breaks the logic of [FileRefresher.cs](https://github.com/jellyfin/jellyfin/blob/eaf33f01e1fce023491fe94d69f28de2f2c9666f/Emby.Server.Implementations/IO/FileRefresher.cs#L163) as part of the real-time monitoring function, as when a new or changed file is detected Jellyfin performs the following:

1. checks for an entry in the DB of an absolute path to the changed/new file (e.g. /Movies/The Movie (2024)/movie.mkv).
2. If an entry does not exists (which is going to be true for new files) it checks the DB for an entry to the parent folder of the file. (e.g. /Movies/The Movie (2024)/ ).
3. **A DB entry for the movie folder does not exist by default as such an entity does not exist in the DB (this is where the problem arises)**
4.  Jellyfin the move up on directory checking for an entry on the root directory of the library (e.g. /Movies/) for which it finds an entry in the DB as `MediaBrowser.Controller.Entities.Folder`
5. Jellyfin then starts a scan of the whole library root directory, unnecessarily.

### Reproduction steps

1. Have a movies library with real-time monitoring enabled
2. Add a new file in a movie folder in the movies library (e.g. new subtitle or better version of the movie)
3. check logs for `Emby.Server.Implementations.IO.LibraryMonitor: "movies" ("/movies") will be refreshed.` which indicates a scan of the root folder of the library

The information tracked in the DB for movies breaks the logic of [FileRefresher.cs](https://github.com/jellyfin/jellyfin/blob/eaf33f01e1fce023491fe94d69f28de2f2c9666f/Emby.Server.Implementations/IO/FileRefresher.cs#L163) as part of the real-time monitoring function, as when a new or changed file is detected Jellyfin performs the following:

1. checks for an entry in the DB of an absolute path to the changed/new file (e.g. /Movies/The Movie (2024)/movie.mkv).
2. If an entry does not exists (which is going to be true for new files) it checks the DB for an entry to the parent folder of the file. (e.g. /Movies/The Movie (2024)/ ).
3. **A DB entry for the movie folder does not exist by default as such an entity does not exist in the DB (this is where the problem arises)**
4.  Jellyfin the move up on directory checking for an entry on the root directory of the library (e.g. /Movies/) for which it finds an entry in the DB as `MediaBrowser.Controller.Entities.Folder`
5. Jellyfin then starts a scan of the whole library root directory, unnecessarily.

The solution to this would be to have another entity in the DB for the parent folder of each movie so that real-monitoring will only scan the specific folder where the file was changed/added and not the whole library directory.

Workflow then would look as so:

1. a new subtitle is added to an existing movie, or a movie is replaced with a better version.
2. Jellyfin detects the new file, and checks for a DB entry of the absolute path to the file, which is not found since it's a new file.
3. Jellyfin then checks for a DB entry to the parent folder of the file which is the /Movies/movie (2024) folder.
4. An entry is found so Jellyfin only scans the individual movie folder and not the whole movies library.

### Jellyfin Server version

10.11.6

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.6

### Environment

```markdown
using Unraid latest version.
```

### Jellyfin logs

```shell
[2026-02-02 15:09:03.760 +00:00] [INF] [26] Emby.Server.Implementations.IO.LibraryMonitor: "movies" ("/movies") will be refreshed.
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

dfederm:
Fixes #16172

Movie directories are not tracked as separate DB entities unlike TV show/season folders. When real-time file monitoring detects a change inside a movie folder (eg. a new subtitle or replaced movie file), `FileRefresher.GetAffectedBaseItem` walks up the directory tree looking for a DB entry. Since the movie folder itself isn't tracked, it walks all the way up to the library root, triggering an unnecessary full library scan.

This change modifies `GetAffectedBaseItem` to check for known media items inside untracked directories before walking further up. When a directory isn't found in the DB but exists on disk, the method enumerates its files and looks for any that are tracked (eg. the movie file). If found, that item is returned, scoping the refresh to just the affected movie instead of the entire library.

For completely new movies (no files in the directory are tracked yet), the behavior is unchanged - the walk continues up to the library root so the new movie folder can be discovered.

For files directly in the library root, like adding subtitles to an existing movie, the behavior is unchanged and there will be a full library rescan. However, this is not the recommended structure anyway.

### DIFF
--- a/Emby.Server.Implementations/IO/FileRefresher.cs
+++ b/Emby.Server.Implementations/IO/FileRefresher.cs
@@ -168,7 +168,36 @@ namespace Emby.Server.Implementations.IO
             {
                 item = _libraryManager.FindByPath(path, null);
 
-                path = System.IO.Path.GetDirectoryName(path) ?? string.Empty;
+                if (item is null)
+                {
+                    var parentDir = System.IO.Path.GetDirectoryName(path) ?? string.Empty;
+
+                    // If this path is an untracked directory, look for a known media
+                    // item inside it (eg. a movie file in a movie folder). This prevents
+                    // walking all the way up to the library root when intermediate
+                    // directories like movie folders are not tracked in the DB.
+                    if (!string.IsNullOrEmpty(parentDir) && Directory.Exists(path))
+                    {
+                        try
+                        {
+                            foreach (var file in Directory.EnumerateFiles(path))
+                            {
+                                var childItem = _libraryManager.FindByPath(file, false);
+                                if (childItem is not null)
+                                {
+                                    item = childItem;
+                                    break;
+                                }
+                            }
+                        }
+                        catch (Exception ex)
+                        {
+                            _logger.LogDebug(ex, "Error enumerating files in {Path}", path);
+                        }
+                    }
+
+                    path = parentDir;
+                }
             }
 
             if (item is not null)
--- a/Emby.Server.Implementations/IO/FileRefresher.cs
+++ b/Emby.Server.Implementations/IO/FileRefresher.cs
@@ -5,6 +5,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Threading;
+using Emby.Server.Implementations.Library;
 using MediaBrowser.Controller.Configuration;
 using MediaBrowser.Controller.Entities;
 using MediaBrowser.Controller.Library;
@@ -182,6 +183,11 @@ namespace Emby.Server.Implementations.IO
                         {
                             foreach (var file in Directory.EnumerateFiles(path))
                             {
+                                if (IgnorePatterns.ShouldIgnore(file))
+                                {
+                                    continue;
+                                }
+
                                 var childItem = _libraryManager.FindByPath(file, false);
                                 if (childItem is not null)
                                 {


Emby.Server.Implementations/IO/FileRefresher.cs
```diff
@@ -168,7 +168,36 @@ private void ProcessPathChanges(List<string> paths)
...
+                    // item inside it (eg. a movie file in a movie folder). This prevents
+                    // walking all the way up to the library root when intermediate
+                    // directories like movie folders are not tracked in the DB.
+                    if (!string.IsNullOrEmpty(parentDir) && Directory.Exists(path))
+                    {
+                        try
+                        {
+                            foreach (var file in Directory.EnumerateFiles(path))
+                            {
+                                var childItem = _libraryManager.FindByPath(file, false);
```
JPVenson, line 191: @Shadowghost should this be checking our ignore system?

dfederm: @JPVenson I am now filtering through `IgnorePatterns.ShouldIgnore`. I'm still new to contributing to the repo so wasn't aware of ignored patterns :)

dfederm: Hmm, but maybe I need `DotIgnoreIgnoreRule` too? Looking...

dfederm: Making a judgement call here but let me know if you'd prefer a different option.

Option 1 (my recommendation): Just use `IgnorePatterns.ShouldIgnore`. It's lightweight and worst case there are some extraneous `_libraryManager.FindByPath` calls made. The library manager itself filters, so there would be not items by ignored paths to begin with, ie this won't find anything that's supposed to be ignored.

Option 2: Use `_libraryManager.IgnoreFile`. This would be the most consistent, but also pretty expensive as it looks up ignore files for each of the files in the dir.

Option 3: Go back to my original change where there is no `IgnorePatterns.ShouldIgnore` call either. Same reasoning as Option 1, worst case it's extra `_libraryManager.FindByPath` calls. IMHO `IgnorePatterns.ShouldIgnore` is lightweight enough to keep though.

morkyy:
Thanks for looking into this and creating a fix! Really appreciate it!

That's said I still think there is room for improvement here with library scanning when a new movie is added.

Why scan the whole library when RTM detects the specific folder and file added to the library? I get that if the folder is not in the DB it goes up the tree but all the other files are in the DB and RTM didn't detect a change in those. Is there a need to ffprobe every single file? 

Wouldn't it be simpler to just do the folder that RTM detected plus it's contents? Or do a full library scan but not ffprobe unless the file matches what RTM detected?

Not sure what the exact logic is for the current method but just pointing out some things here.

I'd also suggest having folder entries for movies in the DB similar to how we do for shows as it makes it more consistent.

dfederm:
> Wouldn't it be simpler to just do the folder that RTM detected plus it's contents?

Yes, that's what this change does. It enumerates and untracked dir before walking up.

morkyy:
> > Wouldn't it be simpler to just do the folder that RTM detected plus it's contents?
> 
> Yes, that's what this change does. It enumerates and untracked dir before walking up.

Yeah but it does this only for updated movies. If I understand correctly from your description, if a new movie is added in a new folder then it will still scan the entire library.

Emby.Server.Implementations/IO/FileRefresher.cs
JPVenson: No worries, i wasn't sure myself thats why i pinged ghost for it (he rebuild the system recently).

He is also currently cooking on a cached lookup so to make the library managers `IgnoreFile` method more performant but i want his opinion on that @Shadowghost

dfederm:
Seems like scope creep. The original issue described a new file (eg subtitle) in an existing movie folder. This change addresses that scenario.

Sure other scenarios exist which could also be improved, but let's not let perfect be the enemy of good.

dfederm:
@Shadowghost gentle poke for this PR.

dfederm:
@Shadowghost nudge on this PR

Shadowghost (approved):
I think this is fine as-is, there should not be a correctness issue.

IDisposable (approved):

dfederm:
Nudge @JPVenson

Emby.Server.Implementations/IO/FileRefresher.cs
```diff
@@ -168,7 +169,41 @@ private void ProcessPathChanges(List<string> paths)
             {
                 item = _libraryManager.FindByPath(path, null);
 
-                path = System.IO.Path.GetDirectoryName(path) ?? string.Empty;
+                if (item is null)
+                {
+                    var parentDir = System.IO.Path.GetDirectoryName(path) ?? string.Empty;
+
+                    // If this path is an untracked directory, look for a known media
+                    // item inside it (eg. a movie file in a movie folder). This prevents
```
JPVenson:
```suggestion
                    // If this path is an untracked directory, look for a known media
                    // file inside it (eg. a movie file in a movie folder). This prevents
```

JPVenson (commented):
how does this handle multiple versions? Say i request a change for `media/movies/Stargate Arc Of Truth/` and that has `Stargate Arc of Truth - 1080p.mkv` and `Stargate Arc of Truth - 4K.mkv` in it, only the first ordered file would be scheduled for refresh when i changed/added the later one, no?

dfederm:
Good catch. Pushed a fix.

Issue #16215, opened 2026-02-10
[10.11.6] Folder watch behaviour fails to see newly added tv shows with existing season content


marlon-tucker:
### Description of the bug

I have a staging area for my media, where I make sure the folder structure is correct as well as the naming standards.

If I copy a newly created tv show into the tv show media folder Jellyfin is watching, it will add a media entry for the tv show, but have no seasons found. This also marks the show as watched - I guess that's due to it seeing 0 out of 0 epeisodes have been watched and treating that as true. 

Regardless, if I initiate a manual media scan, it then sees the season content beneath the show and everything works as I would expect.

### Reproduction steps

Have a complete tv show, with the structure Jellyfin expects. in a staging location, not where Jellyfin is actively watching for folder content changes. Make sure Jellyfin has no history of the show. 

For example 

```
> Some Show (2026) 
--> Season 01
----> Some.Show.S01E01.mp4
----> Some.Show.S01E02.mp4
...
...
```

Copy the complete folder structure into your tv library.. Wait for Jellyfin to auto refresh, usually this is a few minutes in my experience.

### What is the current _bug_ behavior?

If the bug is reproduced, the TV show will appear, but have no episodes inside it.

### What is the expected _correct_ behavior?

TV show appears with all the seasons in the folder structure populated.

### Jellyfin Server version

10.11.6

### Specify the build version

10.11.6

### Environment

```markdown
- OS: Unraid 7.0.1 host, ubuntu 24.04.3 LTS docker image
Docker image:
https://github.com/linuxserver/docker-jellyfin/pkgs/container/jellyfin

Media mounted inside the docker image.
```

### Jellyfin logs

```shell
[2026-02-10 18:51:55.184 +00:00] [INF] [56] Emby.Server.Implementations.IO.LibraryMonitor: "tvshows" ("/data/tvshows") will be refreshed.
[2026-02-10 18:52:14.017 +00:00] [INF] [56] MediaBrowser.Providers.TV.SeriesMetadataService: Creating Season "Season Unknown" entry for "[Season name redacted]"
[2026-02-10 18:52:18.857 +00:00] [INF] [15] Emby.Server.Implementations.IO.LibraryMonitor: "movies" ("/data/movies") will be refreshed.
[2026-02-10 18:52:20.654 +00:00] [INF] [9] MediaBrowser.Providers.TV.SeriesMetadataService: Removing virtual season null in series "[Season name redacted]"
[...]

Issue #15327, opened 2025-10-31
'Season Unknown' metadata created for new TV Series with jellyfin server 10.11.1


cohortcw179:
### Description of the bug

A "Seasons Unknown" metadata was created when I added a new TV Series into my library. When I tried editing the metadata, the Content type is listed as "Inherit" and most of the fields are empty. 

### Reproduction steps

This can be recreated with another new TV Series folder and scan for new and updated files.

### What is the current _bug_ behavior?

An additional "Seasons Unknown" is created when refresh metadata is triggered. Triggering refresh all metadata does not help to remove this additional metadata.

### What is the expected _correct_ behavior?

No additional folder "Season Unknown" should be created.

### Jellyfin Server version

10.10.0+

### Specify the build version

10.11.1

### Environment

```markdown
- OS: Ubuntu 22.04.5 LTS
- Linux Kernel: Linux 6.14.11-4-pve x86_64)
- Virtualization: Proxmox 9.0.11
- Clients:[Brave Jellyfin Web 10.11.1]
- Browser:Brave
- FFmpeg Version: ffmpeg 7:4.4.2-0ubuntu0.22.04.1
- Playback Method: Brave + AndroidTV
- Hardware Acceleration: QSV
- GPU Model: iGPU (Intel Gen9)
- Plugins: default plugins, Meilisearch, Reports, TheTVDB, Kodi Sync Queue, TVMaze, TMDb Box Sets, Trakt
- Reverse Proxy: None
- Base URL:
- Networking: Bridge/NAT
- Jellyfin Data Storage: local SSD
- Media Storage: OneDrive via rclone
- External Integrations: none
```

### Jellyfin logs

```shell
[2025-10-30 21:59:58.020 +08:00] [INF] Started library sync
[2025-10-30 22:00:00.062 +08:00] [INF] Library Sync: Updated 3 items
[2025-10-30 22:00:00.062 +08:00] [INF] Finished library sync, taking "00:00:02.0417074"
[...]

PR #16236, opened 2026-02-13
Fix episodes appearing in Season Unknown incorrectly and prevent unnecessary virtual season creation

theguymadmax:
Episodes without a parseable filename that were physically inside a season folder were being incorrectly added to both their physical season and Season Unknown folders.

This was caused by `FilterEpisodesBySeason` in `Series.cs`, which has three conditions to match an episode to a season. These episodes would match both the second and third conditions, causing the duplication.

**Changes:** 
Added a guard to the second condition so it only matches Season Unknown for episodes that don't already have a real physical season parent.

**Issues:**
 N/A

---
Virtual seasons unknown would unnecessarily be created and deleted on every scan.

This was caused by `CreateSeasonsAsync` in `SeriesMetadataService.cs` treating a null ParentIndexNumber as a valid season number requiring a dummy season, regardless of whether the episode already had a physical home or hadn't been fully processed yet.

**Changes:** 
Added checks to skip episodes from season creation if they already belong to a physical season folder (matched by ID and path) or haven't been fully processed yet (empty SeasonId and null ParentIndexNumber). This covers both season subfolders and flat series structures.

**Issues:**
 N/A

### DIFF
--- a/MediaBrowser.Controller/Entities/TV/Series.cs
+++ b/MediaBrowser.Controller/Entities/TV/Series.cs
@@ -451,8 +451,7 @@ namespace MediaBrowser.Controller.Entities.TV
 
                 if (!currentSeasonNumber.HasValue && !seasonNumber.HasValue && parentSeason.LocationType == LocationType.Virtual)
                 {
-                    var episodeSeason = episodeItem.Season;
-                    return episodeSeason is null || episodeSeason.LocationType == LocationType.Virtual;
+                    return episodeItem.Season is null or { LocationType: LocationType.Virtual };
                 }
 
                 var season = episodeItem.Season;


MediaBrowser.Controller/Entities/TV/Series.cs
```diff
@@ -451,7 +451,8 @@ public static IEnumerable<BaseItem> FilterEpisodesBySeason(IEnumerable<BaseItem>
 
                 if (!currentSeasonNumber.HasValue && !seasonNumber.HasValue && parentSeason.LocationType == LocationType.Virtual)
                 {
-                    return true;
+                    var episodeSeason = episodeItem.Season;
+                    return episodeSeason is null || episodeSeason.LocationType == LocationType.Virtual;
```
JPVenson:
```suggestion
                    return episodeItem.Season is null or { LocationType == LocationType.Virtual };
```

Bond-009 (approved):

Pentaphon:
Why not mention that this fixes https://github.com/jellyfin/jellyfin/issues/15327 ?

tom-pryor:
Does this potentially also fix #16215 ?

PR #16256, opened 2026-02-18
Upgrade Swashbuckle to v10

Shadowghost:
**Changes**
* Upgrade to latest swashbuckle v10
* Fix Filters
* Add Filter for SecurityScheme


nielsvanvelzen (approved):
My expectation was that this upgrade across multiple major versions would result in a lot of changes to the OpenAPI output.. well I'm surprised to say it did not make any changes to the definitions, only changes are the version bump and the addition of the tags object, both non-breaking.

So ✅ for the API "changes"!

Issue #2710, opened 2020-03-29
Subtitles don't appear, "Fetching additional data" indefinitely


CutterXYZ:
Sometimes Jellyfin doesn't display the subtitles and the message "Fetching additionnal data" is displayed indefinitely in the player. In this example the subtitles are encoded in UTF8 and stored in the mkv file being played. I've tried restarting the server and refreshing the library. I've had this bug in previous versions already. I couldn't find an error in the logs.

**Media Info of the file**
the.kominsky.method.s01e01.720p.webrip.x264-strife
Containermkv
PathD:\Downloads\BitTorrent\Séries\The.Kominsky.Method.S01.720p.WEBRip.x264-STRiFE\the.kominsky.method.s01e01.720p.webrip.x264-strife.mkv
Size724 MB
Video
Title720P H264
CodecH264
AVCYes
ProfileHigh
Level41
Resolution1280x678
Aspect ratio640:339
InterlacedNo
Framerate23.976025
Bitrate2992 kbps
Bit depth8 bit
Pixel formatyuv420p
Ref frames1
NAL4
Audio
TitleEng Dolby Digital+ 6 ch Default
Languageeng
CodecEAC3
Channels6 ch
Sample rate48000 Hz
DefaultYes
Subtitle
TitleEng
Languageeng
CodecSUBRIP
DefaultNo
ForcedNo
ExternalNo

**Logs**
no error found in logs

**Stats for Nerds Screenshots**

**Server System (please complete the following information):**
 - OS: Windows 10 64-bit
 - Jellyfin Version: 10.5.2
 - Hardware settings & device: i7 860, no hardware acceleration
 - Reverse proxy: no
 - Other hardware notes: media stored locally
 
**Client System (please complete the following information):**
 - Device: the server
 - OS: Windows 10
 - Client: Web browser
 - Browser: Firefox
 - Client and Browser Version: 75.0b10

eltiel:
I believe it is a known problem, similar to this https://github.com/jellyfin/jellyfin/issues/2650

I've tested that it is fixed in the latest nightly. Give it a try.

CutterXYZ:
I confirm it's fixed in today's nightly. Thanks!

mrkylegp:
I don't believe this is fixed entirely. I'm on 10.5.5 and this is still happening to me.

Scripty:
This is still an issue with 10.10.7

e-v-o-l-v-e:
[...]

PR #16257, opened 2026-02-19
Fix subtitle extraction caching empty files

lowbit:
**Changes**
When FFMPEG subtitle extraction fails or times out (for example file still being downloaded via torrent and is incomplete during sequential or non sequential download) but you start playing via jellyfin, it leaves behind 0 byte files in subtitle cache. Since the cache only checks if file exists `File.Exists()` without validating content, retries are skipped which causes the client to show "Fetching additional data" indefinitely.

**Issues**
https://github.com/jellyfin/jellyfin/issues/6672
https://github.com/jellyfin/jellyfin/issues/2710
I have found by searching google similar issues people had, not sure if this will 100% fix their issues as well but is a good start (solves my issue with web player and subtitles never showing up due to extraction) on webos html5 player.
Kodi for example works because it doesnt need extraction like html 5 players.

### DIFF
--- a/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
+++ b/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
@@ -321,7 +321,7 @@ namespace MediaBrowser.MediaEncoding.Subtitles
         {
             using (await _semaphoreLocks.LockAsync(outputPath, cancellationToken).ConfigureAwait(false))
             {
-                if (!File.Exists(outputPath))
+                if (!File.Exists(outputPath) || _fileSystem.GetFileInfo(outputPath).Length == 0)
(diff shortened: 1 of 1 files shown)


nyanmisaka:
Since this is a minor fix, could you retarget it to the **release-10.11.z** branch so that it can be backported?

lowbit:
changed target to `jellyfin/jellyfin:release-10.11.z`

nyanmisaka:
> changed target to `jellyfin/jellyfin:release-10.11.z`

You will also need to rebase, which requires a force-push.

lowbit:
Sorry I realized as soon as i clicked, should be fine now.
Note: I am swallowing `FileNotFoundException` if there is a need to special handle or throw it up let me know.

PR #16259, opened 2026-02-19
Fix folders being identified as seasons in mixed libraries

### LINKED ISSUE #15676: Mixed movies and TV shows Folder mistakenly recognized as a TV series
### Description of the bug

**Overview**
After upgrading from the old version 10.9.3 to the current 10.11.3, when using a mix of movies and TV shows, if the first character of a folder name (after removing special symbols) is a number, the folder will be incorrectly identified as a TV series, and the files/subfolders under it will have their hierarchical structure removed.
**Process**
After upgrading from the old version 10.9.3 to the current 10.11.3, some folders that were originally recognized as regular folders suddenly started being identified as TV series. Initially, I assumed this was due to changes in file recognition logic. To verify, I copied the entire directory, deleted all video and other files while retaining only the folder structure—but the folders were still incorrectly identified as TV series. I then compared the differences between these folders and others that worked normally, and discovered the root cause: if the first character of a folder name (after removing special symbols) is a number, the system now recognizes it as a TV series (instead of a regular folder) and flattens the subfolder hierarchy underneath.
**Thank you to the developers for their selfless sharing.**
The content is AI-translated. Please forgive any ambiguities.

### Reproduction steps

In version 10.11.3, when using the "Mixed movies and TV shows" mode, if you create folders such as [7sht123]665-C, [123]abcC, or 7kill under a directory, these folders will be incorrectly identified as TV series instead of regular folders. Additionally, if such folders are added under a directory that was previously recognized as a file, the parent directory will also be reclassified as a TV series.

Incorrectly identifies folders as TV series

Identify a directory as a TV series/movie if it contains video files; if a directory only has subfolders (with no video files), identify it as a regular folder.

### Jellyfin Server version

10.11.3

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.3

### Environment

```markdown
System: Ubuntu 24.04.3
Kernel: 6.14.0-36-generic
Client: Browser
```

### Jellyfin logs

```shell
No error logs are available.
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

theguymadmax:
**Changes**
Added a guard so that in mixed libraries, numeric folders without explicit season keywords are not misidentified as TV seasons

**Issues**
Fixes: #15676

### DIFF
--- a/Emby.Naming/TV/SeasonPathParser.cs
+++ b/Emby.Naming/TV/SeasonPathParser.cs
@@ -21,6 +21,9 @@ namespace Emby.Naming.TV
         [GeneratedRegex(@"[sS](\d{1,4})(?!\d|[eE]\d)(?=\.|_|-|\[|\]|\s|$)", RegexOptions.None)]
         private static partial Regex SeasonPrefix();
 
+        [GeneratedRegex(@"(season|staffel|stagione|sæson|temporada|series|kausi|säsong|seizoen|seasong|sezon|sezona|sezóna|sezonul|시즌|シーズン|сезон)", RegexOptions.IgnoreCase)]
+        private static partial Regex SeasonKeyword();
+
         /// <summary>
         /// Attempts to parse season number from path.
         /// </summary>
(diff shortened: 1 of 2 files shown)


Emby.Naming/TV/SeasonPathParser.cs
```diff
@@ -21,6 +21,9 @@ public static partial class SeasonPathParser
         [GeneratedRegex(@"[sS](\d{1,4})(?!\d|[eE]\d)(?=\.|_|-|\[|\]|\s|$)", RegexOptions.None)]
         private static partial Regex SeasonPrefix();
 
+        [GeneratedRegex(@"(season|staffel|stagione|sæson|temporada|series|kausi|säsong|seizoen|seasong|sezon|sezona|sezóna|sezonul|시즌|シーズン|сезон)", RegexOptions.IgnoreCase)]
+        private static partial Regex SeasonKeyword();
```
Shadowghost: We have a more complex regex fort his somewhere, we should use that one

theguymadmax:
Github has lost its mind, I did no such thing.

Emby.Naming/TV/SeasonPathParser.cs
theguymadmax: Hmm, didn't see one. But we can consolidate them so we don't have to maintain three different ones anytime we need to make an update.

Shadowghost (approved):

JPVenson (approved):

Bond-009 (approved):

PR #16263, opened 2026-02-20
Mitigate pull_request_target privilege escalation

andrewrabert:
Hotfix: replace `pull_request_target` with `pull_request` to stop granting write permissions and secret access to fork PRs.

Some workflows will break (PR comments, labeling, deploy previews). Can be restored later with `workflow_run`.

https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/

See also: https://github.com/jellyfin/jellyfin/pull/16162

### DIFF
--- a/.github/workflows/ci-compat.yml
+++ b/.github/workflows/ci-compat.yml
@@ -1,6 +1,6 @@
 ﻿name: ABI Compatibility
 on:
-  pull_request_target:
+  pull_request:
 
 permissions: {}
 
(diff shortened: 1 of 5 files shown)


andrewrabert:
This was a hole that needed to be addressed. I took the blind `sed s/pull_request_target/pull_request/g` approach to just get it done now rather than finding the time to properly fix the pipelines.

joshuaboniface (approved):

PR #16267, opened 2026-02-20
Add environment variable to set the server name

### LINKED ISSUE #16261: Support setting JELLYFIN_ServerName via environment variable
Image

Hello Jellyfin team,

I would like to be able to set the **Server Name** of Jellyfin using an **environment variable**.

Currently, the only way to set the server name in Docker is by using the `hostname` in the Docker Compose file, but I prefer to configure it via environment variables for consistency with other settings.

This would allow automated deployments and easier configuration in containerized environments.

Thank you!

etaichoset:
**Changes**
Add logic in `Emby.Server.Implementations.ApplicationHost` that reads from environment variable `JELLYFIN_SERVER_NAME` to set the server name. This take precedence over the machine name and the configuration.

**Issues**
Fixes [16261](https://github.com/jellyfin/jellyfin/issues/16261)

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -288,3 +288,4 @@
  - [Michael McElroy](https://github.com/mcmcelro)
  - [Soumyadip Auddy](https://github.com/SoumyadipAuddy)
  - [DerMaddis](https://github.com/dermaddis)
+ - [Etai Choset](https://github.com/etaichoset)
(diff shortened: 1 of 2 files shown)


Shadowghost (approved):
@etaichoset this would need a rebase but otherwise it's fine

JPVenson:
Technically this is superseded by https://github.com/jellyfin/jellyfin/pull/16750 which allows all config to be modified via env/arg/config or yet to be determined methods in a global way.

Bond-009:
I agree with @JPVenson here IConfiguration is the preferred way to solve this.

PR #16276, opened 2026-02-22
change password reset behavior to more closely match other services

dkanada:
**Changes**

This pull request deprecates the ResetPassword feature as it sets the user password to an empty string. I'm a bit undecided on *when* authentication should be required for password updates, but I don't think it's good practice to set empty passwords - we even reject them for administrators outside this specific case.

Old Behavior

- all users can change their own password if they know their current one
- admins can change **any** password without authentication except their own

New Behavior

- password resets to empty strings are marked as deprecated
- all users can change their own password if they know their current one
- admins can change user passwords without authentication
- admins can change other admin passwords with **their own** password
- the API should be fully backwards compatible

All bullets are open to change - the goal was simply an improvement over the previous situation. Perhaps requiring admin reauthentication in all situations would be good. I'll open a PR for the web UI if this gains any traction.

**Issues**

None

### DIFF
--- a/Jellyfin.Api/Controllers/VideosController.cs
+++ b/Jellyfin.Api/Controllers/VideosController.cs
@@ -435,6 +435,7 @@ public class VideosController : BaseJellyfinApiController
                 cancellationTokenSource.Token)
             .ConfigureAwait(false);
 
+#pragma warning disable CS0618
         if (@static.HasValue && @static.Value && state.DirectStreamProvider is not null)
         {
             var liveStreamInfo = _mediaSourceManager.GetLiveStreamInfo(streamingRequest.LiveStreamId);
@@ -447,6 +448,7 @@ public class VideosController : BaseJellyfinApiController
             // TODO (moved from MediaBrowser.Api): Don't hardcode contentType
             return File(liveStream, MimeTypes.GetMimeType("file.ts"));
         }
+#pragma warning restore CS0618
 
         // Static remote stream
         if (@static.HasValue && @static.Value && state.InputProtocol == MediaProtocol.Http)
--- a/Jellyfin.Api/Helpers/AudioHelper.cs
+++ b/Jellyfin.Api/Helpers/AudioHelper.cs
@@ -99,6 +99,7 @@ public class AudioHelper
                 cancellationTokenSource.Token)
             .ConfigureAwait(false);
 
+#pragma warning disable CS0618
         if (streamingRequest.Static && state.DirectStreamProvider is not null)
         {
             var liveStreamInfo = _mediaSourceManager.GetLiveStreamInfo(streamingRequest.LiveStreamId);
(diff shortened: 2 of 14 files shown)


MediaBrowser.Model/Plugins/PluginStatus.cs
```diff
@@ -39,8 +41,9 @@ public enum PluginStatus
         Superseded = -4,
 
         /// <summary>
-        /// [DEPRECATED] See Superseded.
+        /// This plugin has been superseded by another version.
```
jumoog, line 44: why? it's not even related to your change

MediaBrowser.Controller/Streaming/StreamState.cs
```diff
@@ -53,9 +53,7 @@ public StreamingRequestDto Request
     /// <summary>
     /// Gets or sets the direct stream provider.
     /// </summary>
-    /// <remarks>
-    /// Deprecated.
-    /// </remarks>
+    [Obsolete("Deprecated Feature")]
```
jumoog, line 56: why? it's not even related to your password change

Jellyfin.Api/Helpers/AudioHelper.cs
```diff
@@ -99,6 +99,7 @@ public async Task<ActionResult> GetAudioStream(
                 cancellationTokenSource.Token)
             .ConfigureAwait(false);
 
+#pragma warning disable CS0618
```
jumoog, line 102: why? it's not even related to your password change

Jellyfin.Api/Controllers/VideosController.cs
```diff
@@ -435,6 +435,7 @@ [FromQuery] [RegularExpression(EncodingHelper.ContainerValidationRegex)] string?
                 cancellationTokenSource.Token)
             .ConfigureAwait(false);
 
+#pragma warning disable CS0618
```
jumoog, line 438: why? it's not even related to your password change

Jellyfin.Api/Models/UserDtos/UpdateUserPassword.cs
```diff
@@ -8,20 +11,36 @@ public class UpdateUserPassword
     /// <summary>
     /// Gets or sets the current sha1-hashed password.
     /// </summary>
+    [Obsolete("Use SessionPassword")]
     public string? CurrentPassword { get; set; }
 
+    /// <summary>
+    /// Gets or sets a new plain text password for the requested user.
+    /// </summary>
+    [Required]
```
jumoog, line 20: why did you choose required?

Jellyfin.Api/Helpers/StreamingHelpers.cs
```diff
@@ -144,7 +144,10 @@ public static async Task<StreamState> GetStreamingState(
         {
             var liveStreamInfo = await mediaSourceManager.GetLiveStreamWithDirectStreamProvider(streamingRequest.LiveStreamId, cancellationToken).ConfigureAwait(false);
             mediaSource = liveStreamInfo.Item1;
+
+#pragma warning disable CS0618
```
jumoog, line 148: why? it's not even related to your password change

PR #16283, opened 2026-02-23
Fix cached thumbnails spinning up HDDs by deferring source image File.Exists to cache-miss path

glic3rinu:
`ProcessImage` calls `File.Exists(originalImagePath)` unconditionally before checking if a cached resize exists. This stats the source filesystem on every image request even when the cache is warm.

On systems with HDD media storage and spindown enabled, this wakes drives on every page load. The `dateModified` used in the cache key comes from the database (set during library scans), so the stat provides no cache-invalidation value.

Setup:
- media on large HDD array, spun down on idle to reduce power/noise
- cache on NVMe

User Issue:
- browsing jellyfin media always spin up drives, even if everything is cached.

**Changes**

Move the `File.Exists` check inside the cache-miss branch so it only runs when the image actually needs to be read and encoded. Also removes the no-op `GetSupportedImage` call whose both branches return their input unchanged.

**Issues**

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -211,6 +211,7 @@
  - [martenumberto](https://github.com/martenumberto)
  - [ZeusCraft10](https://github.com/ZeusCraft10)
  - [MarcoCoreDuo](https://github.com/MarcoCoreDuo)
+ - [glic3rinu](https://github.com/glic3rinu)
 
 # Emby Contributors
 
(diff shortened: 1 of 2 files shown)


glic3rinu:
initial test didn't spin up the drives, but i tested again after 1 hour and now i see drives waking up when browsing library, so there is something more going on.

glic3rinu:
I did more digging and I can confirm this patch actually works. The problem I saw was because some of my thumbnails were not yet cached, causing fs reads to populate cache.

glic3rinu:
dropping this PR since after more testing I don't think this patch makes any real difference.

PR #16286, opened 2026-02-23
Allow direct play for M3U with a simultaneous stream limit set

WizardOfYendor1:
_(Updated/Revised) This is now only about the TunerCount == 0 condition in M3UTunerHost._

For M3U Live TV sources, setting a simultaneous stream limit greater than zero causes the server to report that direct play is unsupported, even when the source could otherwise be direct played. This causes playback to fall back to remuxing or transcoding.

Basically, I don't think a configured max stream limit should also mean the source cannot advertise direct play. The tuner count should limit concurrent streams;  it should not change whether direct play is supported. The Live TV stream lifecycle is managed elsewhere, and an over-limit stream should result in a LiveTvConflictException.

**Changes**
Remove the TunerCount == 0 restriction for direct play in M3UTunerHost, since the server already manages tuner lifecycle through the LiveStreamFiles. A LiveTvConflictException would be thrown if the limit is exceeded. 

**Issues**
I believe this is partially related to: 
https://github.com/jellyfin/jellyfin/issues/9813

### DIFF
--- a/src/Jellyfin.LiveTv/TunerHosts/M3UTunerHost.cs
+++ b/src/Jellyfin.LiveTv/TunerHosts/M3UTunerHost.cs
@@ -145,7 +145,7 @@ namespace Jellyfin.LiveTv.TunerHosts
         {
             var path = channel.Path;
 
-            var supportsDirectPlay = !info.EnableStreamLooping && info.TunerCount == 0;
+            var supportsDirectPlay = !info.EnableStreamLooping;
             var supportsDirectStream = !info.EnableStreamLooping;
 
             var protocol = _mediaSourceManager.GetPathProtocol(path);


john-pierre:
That’s great 🙂 Have you tested this with TVHeadend? And how are the TV stream startup times in general with the changes? For me, they are between 5 and 10 seconds at the moment

WizardOfYendor1:
> That’s great 🙂 Have you tested this with TVHeadend? And how are the TV stream startup times in general with the changes? For me, they are between 5 and 10 seconds at the moment

All I have is Tvheadend, so yes. Extensively :-). I do NOT use the plugin though, I use M3U (IPTV) tuner my tvheadend server. My tuning times for "direct play" (mpeg-ts) is 3 seconds or less. Fast enough to not be annoying. 
With the current parameters for FFMPEG to to remux/transcode - ~ 8 seconds is about he best you can hope for.  And it's really is difficult to tune up FFMPEG to start streaming much fast than that. I tried, I even have a branch for my server that lets me put my own custom FFMPEG parameters in - overriding the jellyfin defaults.  In the end you'll deal with annoying stability problems to shave off maybe 3 seconds (being optimistic).

WizardOfYendor1:
To add a bit more to this. I've tested this change locally with an older Roku-TV client..... and it doesn't like it :-).  Perhaps the origin of the universal disablement.

It will eventually tune a direct mpeg-ts channel after 2 minutes or something - but basically it's unusable.  
The Roku client is advertising to server "EnableDirectPlay" for ts containers (config setting) so the server delivers it (with this change) but Roku doesn't handle it. It seems to want it's streams remuxed into HLS (especially "infinite" ones). This gets even more colorful with h.264 content in ts containers (versus MPEG2 in my case/testing). There are issues/discussions about this in different places.

Setting my Roku client to force transcode (mux) mitigates this issue.

To add to that, once it DOES finally starts streaming, if you try to view the stream details on the overlay (movie camera icon) they player crashes out and returns the user to the home screen. I SUSPECT the is related an issue with the server assigning -1 indexes to the stream and when the Roku client attempts to find details on that - boom.  I have related PRs in the AndroidTV client . Different issue(s) though. 

As I understand all of this, and I could be wrong - it's complex,  this change is correct in that it's giving the clients what they are asking for.  

I could see an argument to enhance the server API to say - give me direct playback of ts containers for content fully on disk (i.e. recorded live tv content) but NOT for "infinite streams".  i.e.  Clients would say, I can handle ts but only "finite" ones. Remux infinite ones.

MoDDe2025:
I can confirm this issue. I'm running Jellyfin with TVHeadend via M3U tuner (?profile=pass) on a Ugreen NAS (Intel N100). Every Live TV stream gets transcoded despite H264 video and AAC audio, with the bitrate incorrectly reported as ~140 Mbps.
Kodi with the TVHeadend HTSP plugin on the same setup switches channels in 2-3 seconds with no transcoding. Jellyfin takes 10-30 seconds due to unnecessary transcoding.
I would be happy to test this fix. Is a test build available?

gw8484:
Any idea when the changes will be included in a release?  I can understand concerns about the changes breaking some things in old/buggy clients.  Perhaps a config setting could be added to enable/disable the changes to support such clients?

WizardOfYendor1:
> Any idea when the changes will be included in a release? I can understand concerns about the changes breaking some things in old/buggy clients. Perhaps a config setting could be added to enable/disable the changes to support such clients?

It's not up to me but certainly willing to change/test and answer questions by the maintainers. This is higher risk change as noted with Roku testing that I've done locally. This change and the related androidtv client issue(s), was certainly not "vibe coded" in a few hours. I spent several weekends and many hours drilling, debugging and coding these fixes to the best of my ability/knowledge (which admitedly may not be correct or good enough).  

FWIW, I've been using this, plus a fix in the androidtv client (that will not be merged) for many months now on my devices/server and I can't live without it  (along with a few other fixes/enhancements of mine in the androidtv client :-) ).  No issues thus far and I have nearly instantanous live tv channel changing.  Given all the other issues, I sometimes feel I'm one of the very few people who watch live tv on androidtv.

gw8484:
> FWIW, I've been using this, plus a fix in the androidtv client (that will not be merged) for many months now on my devices/server and I can't live without it (along with a few other fixes/enhancements of mine in the androidtv client :-) ). No issues thus far and I have nearly instantanous live tv channel changing. Given all the other issues, I sometimes feel I'm one of the very few people who watch live tv on androidtv.

It's good to know the changes work for you.   It certainly looks like the maintainers do not see live TV as a priority.  It makes no sense to do unnecessary transcoding.  I guess I'll have to continue to use Channels DVR.

BTW, if I am understanding the changes correctly, the existing server code will prevent any client from direct play for live TV MPEG2 TS?  I've wasted a lot of time trying different clients.

WizardOfYendor1:
@gw8484 
> BTW, if I am understanding the changes correctly, the existing server code will prevent any client from direct play for live TV MPEG2 TS? I've wasted a lot of time trying different clients.

Yes.  I too have spent alot of frustrating time trying to tune/stream live tv channels faster. Interestingly, I started this journey on Emby but came to Jellyfin because Emby is closed source. With Jellyfin I could dive into the code to see if I could figure out what was happening and code some changes to fix it - and give back in the process like this PR (software dev is my day job).... but I digress....

Do note the difference of transcoding and remuxing into different containers. The best you can get with without this change is a remux, which is still too slow for my tastes (5-8 seconds to tune channel). 
With "direct stream" (very confusing terminology) it is basically true passthrough of MPEG-TS, I can change channels in 1-2 seconds - practically instant.  If an FFMPEG process get spun up for any reason, it's a minimum 5 seconds  just to remux it because in addition to spawning the process, FFMPEG needs to buffer some of the stream to analyze it and start generating segments for the client.

Also,  add the Moonfin web client to the list of clients that don't like this change (Android TV is fine though, oddly).  Still early  looking into this but Moonfin appears to be doing the same thing at Roku - telling Jellyfin they can support containers they can't. 

I can add a parameter to the server config that will allow toggling it (off by default) - called something like: "Allow direct playback of live TV (experimental)".  That's probably wise give the blast radius of clients who just rely on the current state. If there is a chance this will get merged someday, I'll put the extra effort in to enhance/polish it up.

Jellyfin.Api/Helpers/MediaInfoHelper.cs
```diff
@@ -246,9 +246,9 @@ public void SetDeviceSpecificData(
 
         options.MaxBitrate = GetMaxBitrate(maxBitrate, user, ipAddress);
 
-        if (!options.ForceDirectStream)
+        if (!options.ForceDirectStream && !mediaSource.IsInfiniteStream)
         {
-            // direct-stream http streaming is currently broken
+            // this was purposefully disabled in PR 7325 in 2022. It's unclear as to why other than "its currently broken".
```
gnattu: This does not work the same way as you think it is. It should not be enabled in any cases because both the meaning and handling of this direct-stream is different.

gw8484:
> @gw8484
> 
> Do note the difference of transcoding and remuxing into different containers. The best you can get with without this change is a remux, which is still too slow for my tastes (5-8 seconds to tune channel). 

I can see this clearly in my tests.  The HDHomeRun and Channels DVR apps (on Android/Google TV) are much faster switching channels and clearly do not transcode or remux.

> I can add a parameter to the server config that will allow toggling it (off by default) - called something like: "Allow direct playback of live TV (experimental)". That's probably wise give the blast radius of clients who just rely on the current state. If there is a chance this will get merged someday, I'll put the extra effort in to enhance/polish it up.

That would be great.

Jellyfin.Api/Helpers/MediaInfoHelper.cs
gnattu: Plus that code path is not tested and maintained for years, even it appears to work for your use case NOW, it might break any time.

If we want to provide an http proxy-like behavior for liveTV I prefer to explicitly implement it instead of relying on not maintained code as a workaround.

gnattu (changes requested):

Jellyfin.Api/Helpers/MediaInfoHelper.cs
WizardOfYendor1: Thanks for responding and looking at this. I've been focused nearly exclusively on understanding how IPTV/LiveTV works, so this helps. 

Getting my head back into this, in short, I can see why I should revert this if condition, thus revert back to the original state (options.EnableDirectStream=false) - aka the direct stream kill switch.  Directplay vs directstream is what confused me here. I now see that this "path" has been abandoned, I definitely don't want to re-enable that abandoned code path. 

So, IF I understand this correctly - the clients are NOT directly *requesting* a proxied http stream, they merely request the stream with their supported profile (container, codecs, etc). This is part of the POST request in the device profile that says here's what I can "direct play". In the event the container and the video/audio codecs don't match that profile, it falls into to the ffmpeg path (transcode). Which, depending on the failure "reasons", it will attempt to remux into a different container or transcode the audio and/or video stream to something the client says it can support in it's transcoding profiles.

That is what I understand to be true of M3U tuners..... HDHomerun tuners appears to be different. Directplay is hardcoded to false, DirectStream is blocked (without this change) thus is closed for everyone. So even though HDHomerun says "SupportsDirectStream=true", it's never evaluated. I presume this is left overs of this abandoned approach. In a nutshell, clients can never evade ffmpeg remux/transcode for HDhomerun tuners... perhaps by design (don't have one, thus never tested with it)

I'll definitely revert this 1st infinite change. Is there anything else I'm missing, is removing the tunercount==0 condition legit? That seems like it should be removed. You can "work around" it by merely setting the tuner count to zero in the tuner configs. Seems odd to have that behave differently if zero or > zero. If so, I can re-title the PR and change the description with all of this in mind.

WizardOfYendor1:
> I can confirm this issue. I'm running Jellyfin with TVHeadend via M3U tuner (?profile=pass) on a Ugreen NAS (Intel N100). Every Live TV stream gets transcoded despite H264 video and AAC audio, with the bitrate incorrectly reported as ~140 Mbps. Kodi with the TVHeadend HTSP plugin on the same setup switches channels in 2-3 seconds with no transcoding. Jellyfin takes 10-30 seconds due to unnecessary transcoding.

I was looking through the codebase on how transcoding/remux gets invoked for live tv streams I noted the "path" that if the max bitrate is exceeded it immediately gets transcoded  (which is totally expected of course). That reminded me of your comment. 
I think your problem might be due to the ffprobe of the stream's bitrate is wildly inaccurate. At 140 Mbit, it would generally exceed normal configs.... unless....perhaps... you manually set that in the client to something crazy like 200 MBit (android tv et al).  The analyze duration is 3 seconds by default, which I believe you can tweak somewhere with extra parameters.
I believe I've evaded this problem because my tvheadend server's tuner card streams out mpeg-ts. And doing a quick search suggests that standard requires a certain field value to be populated that ffprobe relies on versus trying to math it out on a 3 second burst of streaming data.  This issue is logged in 15337,  which goes more in depth (and is soon to auto-close). 

Hope that helps some. TL:DR version: set your tuner's "simultaneous stream limit" to zero then  set the max bit rate on your client to > 140 and see what happens.

Jellyfin.Api/Helpers/MediaInfoHelper.cs
WizardOfYendor1: I've reverted the direct stream change and left the tuner count change.... and updated the PR accordingly.

PR #5121, opened 2021-01-28
Prepare OpenApi specification for multiple API versions

daullmer:
**Changes**

- Add OpenApi support for multiple API versions. Each API version generates its own specification file (see URLs below)
- Add demo v2 API controller. I'm not proposing the current place or route as the future place and route. This is simply meant to remind us in the future how we need to implement the new API controllers to show up in the spec files correctly.
- Change test URL according to the new specification location

**Breaking URL changes for OpenApi specification**
The OpenApi specification file moved, so in theory this is a breaking change:
|                   | Current                | New                   | Notes |
|-------------------|------------------------|---------------------------|-------------------------|
| v1 OpenApi specification | /api-docs/openapi.json | /api-docs/v1/openapi.json |
| v2 OpenApi specification |                    | /api-docs/v2/openapi.json |
| Swagger UI       | /api-docs/swagger      | /api-docs/swagger         | Stayed the same. Support for multiple API versions (v1 and v2) |
| ReDoc UI          | /api-docs/redoc        | /api-docs/swagger         | Stayed the same. Doesn't support multiple API versions. Only shows v1 API |

I tried to leave the old specification at its place but that didn't work.

PR #16292, opened 2026-02-24
basic api versioning on all endpoints

dkanada:
**Changes**

I didn't like my own solution for API backwards compatibility in #16276 so I reimplemented #5121 with some changes.

- version is read from a custom header with PascalCase
- unversioned endpoints are made available on all API versions (feels a bit hacky but automation is nice)
- redoc is served from multiple endpoints for version support

I think the best versioning would just be tracking major versions on the server, but I implemented a suffix in case a hotfix release is required. I don't think it's *strictly* necessary though, and would love to just use v10 et cetera.

**Issues**

None

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -4,6 +4,7 @@
   </PropertyGroup>
   <!-- Run "dotnet list package (dash,dash)outdated" to see the latest versions of each package.-->
   <ItemGroup Label="Package Dependencies">
+    <PackageVersion Include="Asp.Versioning.Mvc.ApiExplorer" Version="8.1.1" />
     <PackageVersion Include="AsyncKeyedLock" Version="8.0.2" />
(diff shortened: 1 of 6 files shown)


dkanada:
I realized we can't release a new API version without a server release, which means we can just track the current server version directly. I also removed v10 since the OpenAPI schema has changed more than I expected since the last release and it's unlikely we'll reach compatibility now.

nielsvanvelzen:
I don't think versioning our API is something we'd want to do. It will be a lot of effort to maintain multiple versions at once and we'd still have to break compatibility sometimes. I'd rather keep our current deprecation approach, which so far has worked out fine.

dkanada:
@Bond-009 I had discussed with @nielsvanvelzen and #16664 is a direct replacement for API versioning - just hadn't got around to closing this pull request yet.

PR #16301, opened 2026-02-25
Fix case-insensitive search by original title for non-ASCII scripts

tdimov93:
**Changes**
Fix case-insensitive search by OriginalTitle for non-ASCII scripts such as Cyrillic, where searching for "Неудържим" would not match "неудържим".

SQLite's `lower()` and `LIKE` only handle ASCII case folding, returning non-ASCII strings unchanged.
Register a `CleanValue` user-defined function using C#'s `ToLowerInvariant()` to perform Unicode-aware case folding when searching by `OriginalTitle`.

No schema changes or data migration required.

**Issues**


Shadowghost:
@tdimov93 since a ton of DB stuff was changed, could you please check if this is still necessary and rebase if it is?

Shadowghost:
I think we also need to fix this in SqlSearchProvider

PR #16314, opened 2026-02-28
Fix contradictory IsFolder/IsNotFolder filters overwriting each other

### LINKED ISSUE #16299: Shuffle/Play All Fails on musicvideos Library Type due to contradictory isFolder logic
### Description of the bug

Component: jellyfin-web -> src/components/playback/playbackmanager.js
Related Issue: [#16260](https://github.com/jellyfin/jellyfin/issues/16297)

Switching the library type to musicvideos (as suggested as a workaround in https://github.com/jellyfin/jellyfin/issues/11667#issuecomment-3856139291) does not resolve the Shuffle/Play All failure. The MediaTypes=Photo bug is gone, but a new bug is introduced: the shuffle request sends the logically contradictory Filters=IsFolder,IsNotFolder in the same query, which also leads to the "Playback Error - Unable to find a valid media source to play"

Observed Request:
```
GET /Users/{userId}/Items?
  ParentId=6c821de4ee8bb273c4f3c63f84fbc3e9
  &Filters=IsFolder%2CIsNotFolder        **contradictory**
  &Recursive=true
  &SortBy=Random
  &MediaTypes=Video
  &Fields=Chapters%2CTrickplay
  &ExcludeLocationTypes=Virtual
  &EnableTotalRecordCount=false
  &CollapseBoxSetItems=false
```

Potential root cause:

It seems the function mergePlaybackQueries() in playbackmanager.js enforces **IsNotFolder** on every playback query:
https://github.com/jellyfin/jellyfin-web/blob/7ab3f5f9ba13f660f281fe8558ce204ef4ec90bd/src/components/playback/playbackmanager.js#L168

```
function mergePlaybackQueries(obj1, obj2) {
    const query = merge({}, obj1, obj2);
    const filters = query.Filters?.split(',');
    if (!filters.includes(ItemFilter.IsNotFolder)) {
        filters.push(ItemFilter.IsNotFolder);
    }
    query.Filters = filters.join(',');
    return query;
}
```

However, in my case the upstream query object passed for the musicvideos library type already contains Filters: 'IsFolder'. After the join in mergePlaybackQueries, both values survive, which results in **Filters=IsFolder,IsNotFolder**

On the server (ItemsController.cs), the foreach filter loop then processes these sequentially with no mutual-exclusion logic (similar to https://github.com/jellyfin/jellyfin/issues/16297)

https://github.com/jellyfin/jellyfin/blob/56a469d8c3c0eed0d809d42a60d847e67ab61671/Jellyfin.Api/Controllers/ItemsController.cs#L389

```
case ItemFilter.IsFolder:
    query.IsFolder = true;   // set first
    break;
case ItemFilter.IsNotFolder:
    query.IsFolder = false;  // overwrites immediately
    break;
```

Potential Fix with guards:
```
case ItemFilter.IsFolder when !filters.Contains(ItemFilter.IsNotFolder):
    query.IsFolder = true;
    break;
case ItemFilter.IsNotFolder when !filters.Contains(ItemFilter.IsFolder):
    query.IsFolder = false;
    break;

```

### Reproduction steps

1. Create library of type Music Videos
2. Add content source that contains both videos in root and subfolders
3. navigate to library, Videos tab
4. press shuffle play
5. playback fails

GET Request contains both IsFolder+IsNotFolder leading to contradictary logic in ItemsController.cs

Shuffle GET request is handled correctly, no injection of IsNotFolder or sanitizing via when guards so that the query actually returns items to play

### Jellyfin Server version

10.11.6

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.6

### Environment

```markdown
- OS:Fedora 43
- Linux Kernel:
- Virtualization:docker
- Clients:Browser
- Browser:Brave
- FFmpeg Version:
- Playback Method:transcode
- Hardware Acceleration:Nvidia NVENC
- CPU Model:Intel(R) Core(TM) i7-7700K CPU @ 4.20GHz
- GPU Model:Nvidia 1080ti
- Plugins:opensubtitles
- Reverse Proxy:none
- Base URL:none, direct access to port on docker host
- Networking:LAN
- Jellyfin Data Storage & Filesystem:container on SSD, media libraries on NAS mounts
- Media Storage & Filesystem:
- External Integrations:
```

### Jellyfin logs

```shell
error is not logged, no other errors observerd in logs
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

robinw0410:
When both IsFolder and IsNotFolder filters are present in a request,
  the sequential switch statement would let IsNotFolder silently overwrite
  IsFolder. Add 'when' guards so contradictory filters cancel out, leaving
  query.IsFolder as null (no folder filtering) instead of applying the
  wrong constraint.

  Fixes #16299

### DIFF
--- a/Jellyfin.Api/Controllers/ArtistsController.cs
+++ b/Jellyfin.Api/Controllers/ArtistsController.cs
@@ -200,10 +200,10 @@ public class ArtistsController : BaseJellyfinApiController
                 case ItemFilter.IsFavoriteOrLikes:
                     query.IsFavoriteOrLiked = true;
                     break;
-                case ItemFilter.IsFolder:
+                case ItemFilter.IsFolder when !filters.Contains(ItemFilter.IsNotFolder):
                     query.IsFolder = true;
                     break;
-                case ItemFilter.IsNotFolder:
+                case ItemFilter.IsNotFolder when !filters.Contains(ItemFilter.IsFolder):
                     query.IsFolder = false;
                     break;
                 case ItemFilter.IsPlayed:
@@ -403,10 +403,10 @@ public class ArtistsController : BaseJellyfinApiController
                 case ItemFilter.IsFavoriteOrLikes:
                     query.IsFavoriteOrLiked = true;
                     break;
-                case ItemFilter.IsFolder:
+                case ItemFilter.IsFolder when !filters.Contains(ItemFilter.IsNotFolder):
                     query.IsFolder = true;
                     break;
-                case ItemFilter.IsNotFolder:
+                case ItemFilter.IsNotFolder when !filters.Contains(ItemFilter.IsFolder):
                     query.IsFolder = false;
                     break;
                 case ItemFilter.IsPlayed:
--- a/Jellyfin.Api/Controllers/ChannelsController.cs
+++ b/Jellyfin.Api/Controllers/ChannelsController.cs
@@ -146,10 +146,10 @@ public class ChannelsController : BaseJellyfinApiController
         {
             switch (filter)
             {
-                case ItemFilter.IsFolder:
+                case ItemFilter.IsFolder when !filters.Contains(ItemFilter.IsNotFolder):
                     query.IsFolder = true;
                     break;
-                case ItemFilter.IsNotFolder:
+                case ItemFilter.IsNotFolder when !filters.Contains(ItemFilter.IsFolder):
                     query.IsFolder = false;
                     break;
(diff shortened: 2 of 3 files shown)


SP3NGL3R:
Interestingly I just discovered that if you add a single library as a playlist (ex: Home Videos), you can select Shuffle All from within the Playlists and it works.

Bond-009:
Superseded by #16398

PR #16315, opened 2026-02-28
Fix HLS playback stalls caused by throttler re-pausing FFmpeg

jonodavis:
## The Problem

Playback periodically stalls and takes a long time to recover during HLS transcoded streaming, even when both client and server connections are healthy.

## Root Cause

A race condition between HLS segment delivery and the `TranscodingThrottler`:

1. FFmpeg transcodes ahead of the client by >60s → throttler pauses FFmpeg
2. Client catches up and requests a segment near the transcoding edge
3. `GetDynamicSegment` calls `UnpauseTranscoding()` to resume FFmpeg
4. The segment wait loop polls every 100ms for the segment to appear on disk
5. **But**: the throttler's 5-second timer fires and checks `transcodingPositionTicks - downloadPositionTicks`
6. `DownloadPositionTicks` only updates in `Response.OnCompleted` (after the segment is fully served), so the gap still appears large
7. Throttler **re-pauses** FFmpeg → the segment is never produced → client stalls indefinitely

## Fix

- **Update `DownloadPositionTicks` eagerly** at segment request time (before the wait loop), so the throttler sees the client's actual demand position and won't re-pause FFmpeg while we're waiting for the segment
- **Add a 60-second timeout** to the segment wait loop to prevent indefinite stalls if FFmpeg hangs for any reason
- **Make the throttler timer callback non-reentrant** using `Interlocked.CompareExchange` to prevent overlapping async callbacks from issuing conflicting pause/unpause commands to FFmpeg's stdin

### DIFF
--- a/Jellyfin.Api/Controllers/DynamicHlsController.cs
+++ b/Jellyfin.Api/Controllers/DynamicHlsController.cs
@@ -1915,13 +1915,9 @@ public class DynamicHlsController : BaseJellyfinApiController
         TranscodingJob? transcodingJob,
         CancellationToken cancellationToken)
     {
-        // Update the download position to reflect the client's current demand.
-        // This prevents the TranscodingThrottler from re-pausing FFmpeg while
-        // we are actively waiting for this segment to be produced.
         if (transcodingJob is not null)
         {
-            var requestedPositionTicks = state.Request.CurrentRuntimeTicks;
-            transcodingJob.DownloadPositionTicks = Math.Max(transcodingJob.DownloadPositionTicks ?? requestedPositionTicks, requestedPositionTicks);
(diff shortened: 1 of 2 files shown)


LTe:
@jonodavis can you reproduce this in specs somehow?

jonodavis:
@LTe Added tests to validate the fix

LTe:
I do not think this test is testing anything useful. I guess it is hard to test race condition in reliable way. Maybe this part of the code should be changes to more event based than timer based.

Do you even can reproduce this with real jellyfin server somehow?

jonodavis:
@LTe I've realised this might not be the right fix. Will create an issue for further investigation.

Issue #327, opened 2018-12-30
jellyfin-theater-pi asking for premium (if fixed or issue already raised apologises) 


bwaddington:

bwaddington:
i should of said jellyfin-theater-pi

JustAMan:
Can you please move this issue to repo of theater-pi?

PR #16341, opened 2026-03-04
Fix hls segment length adjustment for remuxed content

crimsonspecter:
**Changes**
PR #16053 introduced a slight adjustment to the segment length for 23.976 fps content. That makes sense for transcoded content.  However, when video is remuxed, Jellyfin generates the playlist based on key-frames, so the segment length used for generating the manifest must match what FFmpeg is using. Otherwise, we can end up with a playlist whose advertised durations don’t correspond to the segments generated by FFmpeg. Also we can end up with different number of segments generated by FFmpeg compared to the Jellyfin manifest. Some native HLS clients may stall during playback if the actual segment durations differ significantly from what the manifest announces.

**Issues**
Fixes [#327](https://github.com/jellyfin/jellyfin-webos/issues/327)

### DIFF
--- a/Jellyfin.Api/Controllers/DynamicHlsController.cs
+++ b/Jellyfin.Api/Controllers/DynamicHlsController.cs
@@ -1424,8 +1424,8 @@ public class DynamicHlsController : BaseJellyfinApiController
         double fps = state.TargetFramerate ?? 0.0f;
         int segmentLength = state.SegmentLength * 1000;
 
-        // If framerate is fractional (i.e. 23.976), we need to slightly adjust segment length
-        if (Math.Abs(fps - Math.Floor(fps + 0.001f)) > 0.001)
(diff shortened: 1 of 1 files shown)


nyanmisaka:
I haven't actually triggered the issue in the original PR, but it might happen.

I changed your target branch to 10.11 for a stable backport. Please rebase and force push.

crimsonspecter:
> I haven't actually triggered the issue in the original PR, but it might happen.
> 
> I changed your target branch to 10.11 for a stable backport. Please rebase and force push.

Done.

The issue only occurs with certain files, i.e., those with an FPS of 23.975 and a non-constant key-frame interval. Hls.js will probably handle it fine, while native HLS players will sometimes get playback issues because of this.

nyanmisaka (approved):

PR #16346, opened 2026-03-05
Add power inhibitor during playback

ananthb:
# Inhibit Power Transition During Playback

New setting to inhibit power transitions during playback. Supported on Windows, Linux, and MacOS.

**Changes**

Adds a new service that can inhibit power transitions (suspend, sleep, & shutdown) on Windows, MacOS, and Linux.
On Linux it uses systemd-inhibit, so it does not work on non-systemd installations.


nielsvanvelzen:
This is not a feature we want in the server, as mentioned many times in previous issues/feature requests/chat. Instead, this behavior could be implemented by a (third-party) plugin.

ananthb:
I'm sorry to hear that.

Issue #16295, opened 2026-02-24
.avif not automatically detected in movie folders simular issue as #12412


StevMaHo:
### Description of the bug

.avif images are not automatically detected for backdrop and landscape artwork.
However, the images can be added manually without issues.
If the same files are renamed to .jpg or .gif for animated picture they are detected and displayed correctly.

Same issue as #12412 but now for the movie library.

### Reproduction steps

1. Place backdrop and/or landscape images in .avif format in a movie folder
2. Scan the movie for new or updated metadata
3. See that .avif pictures are not detected automatically
4. Rename the same .avif file/s to .jpg or .gif
5. Rescan the movie again
6. picture/s are automatically detected

### What is the current _bug_ behavior?

.avif images are not automatically detected.
They can only be seen if added manually, or if renamed to .jpg or for animated .gif

### What is the expected _correct_ behavior?

.avif images should be automatically detected and displayed for backdrop and landscape artwork, just like other extensions

### Jellyfin Server version

10.11.6

### Specify the build version

10.11.6

### Environment

```markdown
- OS: Windows 11
```

### Jellyfin logs

```shell
-
```

### FFmpeg logs

```shell

```

### Relevant screenshots or videos

![Image](https://github.com/user-attachments/assets/4567e8b6-3eb9-4b97-9269-8b277b425767)

Brady3035:
Added this functionality and created a pull request.

Issue #12412, opened 2024-08-08
Jellyfin-10.9.9 amd64 build on the repo doesn't support AVIF


ikelos:
### This issue respects the following points:

- [X] This is a **bug**, not a question or a configuration issue; Please visit our forum or chat rooms first to troubleshoot with volunteers, before creating a report. The links can be found [here](https://jellyfin.org/contact/).
- [X] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [X] I'm using an up to date version of Jellyfin Server stable, unstable or master; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.
- [X] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [X] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Description of the bug

Using jellyfin 10.9.8 (and retesting with 10.9.9) installed through Gentoo using the repo version of jellyfin, adding a library of type "Home videos and photos" containing AVIF files, does not show any AVIF files, despite the 10.9.0 release notes suggesting that AVIF is now supported.

### Reproduction steps

1. Sign in to Jellyfin
2. Dashboard -> Libraries -> Libraries -> Add Media Library
3. Select content type "Home Videos and Photos"
4. Select a foldering containing valid AVIF files.
5. Rescan the metadata for the folder to be sure.

### What is the current _bug_ behavior?

No AVIF files are displayed or indexed, even when logging is set to "debug", the AVIF files do not appear to be even scanned or mentioned in the logs.

### What is the expected _correct_ behavior?

AVIF files are found, indexed and displayed.

### Jellyfin Server version

10.9.9+

### Specify version number

10.9.9

### Specify the build version

10.9.9

### Environment

```markdown
- OS: Gentoo
- Linux Kernel: 6.10.0-x86_64
- Virtualization: None
- Clients: Browser
- Browser: Firefox, Chrome
- FFmpeg Version: 6.1.1 (Gentoo -r8)
- Playback Method: N/A
- Hardware Acceleration: None
- GPU Model: None
- Plugins: AudioDB, MusicBrainz, OMDb, Studio Images, TMDb
- Reverse Proxy: None
- Base URL: None
- Networking: Host
- Storage: SSHFS
```

### Jellyfin logs

```shell
[...]

PR #16349, opened 2026-03-05
Added .avif support to image types

Brady3035:
Added .avif file type to supported image types. Issue #16295

Added .avif files to SupportedImageExtensions in BaseItem.cs to allow avif images to be found in scans.

**Changes**
Added support for .avif image files, in case users want to replace or convert files to .avif for space saving.

**Issues**
Fixes Issue #16295 and #12412

### DIFF
--- a/MediaBrowser.Controller/Entities/BaseItem.cs
+++ b/MediaBrowser.Controller/Entities/BaseItem.cs
@@ -53,7 +53,7 @@ namespace MediaBrowser.Controller.Entities
         /// The supported image extensions.
         /// </summary>
         public static readonly string[] SupportedImageExtensions
-            = [".png", ".jpg", ".jpeg", ".webp", ".tbn", ".gif", ".svg"];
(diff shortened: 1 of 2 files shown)


Brady3035:
Any reason this isn't being reviewed? The images load when swapped with avif. Not sure what else needs to be done here

Shadowghost (approved):

Shadowghost:
Not sure if we also need to add mappings to `ImageFormatExtension`, `MimeTypes` or `EmbeddedImageProvider` too.
Would also be nice to add a test case for the new file type.

gnattu:
This should be wait till skiasharp is updated to a version that can actually process avif iirc, and only until very recently they fixed the regression on debian arm64 so that it can finally be updated. After the main has the latest skiasharp in, this is safe to merge

Current version still targets 3.119.2: https://github.com/jellyfin/jellyfin/pull/16455

The fix we need is in 3.119.3. We still have to wait for a while.

StevMaHo:
Is it now possible? Skiasharp got updated to 3.119.4 if i see it correctly.

Issue #16248, opened 2026-02-17
Bitrate is incorrectly reported to frontend for video files.


Renari:
### Description of the bug

On the front end for a video file when you check the media info, it will report the bitrate under the `video` category however this is misleading because this reports the overall bitrate, not the video bitrate.

I appears this stems from this block in `ProbeResultNormalizer.cs`, which will cause the video bitrate to be reported as the overall bitrate.
```csharp

            // The bitrate info of FLAC musics and some videos is included in formatInfo.
            if (bitrate == 0
                && formatInfo is not null
                && (stream.Type == MediaStream.Type.Video || (isAudio && stream.Type == MediaStreamType.Audio)))
            {
                // If the stream info doesn't have a bitrate get the value from the media format info
                if (int.TryParse(formatInfo.BitRate, CultureInfo.InvariantCulture, out value))
                {
                    bitrate = value;
                }
            }
```
I looked into fixing this by inferring the video bitrate via taking the total bitrate (obtained above) and subtracting the bitrate of everything else but this only would get me close due to container overhead.

### Reproduction steps

1. Find any video file in Jellyfin.
2. Click the additional menu (three dots) option on that video and select media info.
3. Open the same video file in something like mediainfo
4. See that the video bitrate shown is actually the overall bitrate in mediainfo

### What is the current _bug_ behavior?

The frontend shows a misleading video bitrate that is actually the total bitrate because the video bitrate is not obtained from ffprobe.

### What is the expected _correct_ behavior?

The frontend correctly shows the video bitrate, which is currently being misreported.

### Jellyfin Server version

10.11.6

### Specify the build version

10.11.6

### Environment

```markdown
- OS: CachyOS Linux
- Linux Kernel: 6.19.2-2-cachyos
- Virtualization: podman
- Clients: Browser
- Browser: Firefox 147.0.3
- FFmpeg Version: 8.0.1
- Playback Method: Direct Play
- Hardware Acceleration: NVENC
- CPU Model: AMD Ryzen 7 9800X3D
- GPU Model: NVIDIA GeForce RTX 4070 Ti SUPER
- Plugins: AudioDB, MusicBrainz, OMDb, Studio Images, TMDb
- Reverse Proxy: none
- Base URL: none
- Networking: host
- Jellyfin Data Storage & Filesystem: local ssd
- Media Storage & Filesystem: local ssd
- External Integrations: none
```

### Relevant screenshots or videos

Jellyfin:
[...]

PR #16351, opened 2026-03-05
UpdatedBitrate Detection Order

Brady3035:
Summary
I’ve restructured how we determine the bitrate for media streams. Previously, the logic preffered the Bitrate that the file provided instead of the per-stream variable returned by FFPROBE.  This PR re-orders the system so that ensures we grab the most granular data (the stream-level BPS tag) before falling back to rougher estimates like container-level info or manual calculations based on file size and length. All previous logic is preserved, just reordered.

What’s changing?
Prioritizing Metadata: The code now looks for BPS-eng or BPS tags first. These are provided by FFprobe and are generally the most reliable for individual streams.

Cleaner Fallback Logic: I’ve organized the logic into a clear hierarchy. If the BPS tag is missing, it cascades down through:

A manual calculation using (Total Bytes * 8) / Duration.

The standard streamInfo.BitRate field.

The global formatInfo.BitRate (useful for single-stream files like FLAC), this was the existing logic. 

Simplified Types: I updated GetBPSFromTags to return a standard int instead of int?. This cleans up the calling code since we can now just check if bitrate > 0 rather than dealing with null checks. If the ffprobe returns null, it gets assigned as 0 so that the fallback calculations can take place.

Why?
The previous logic wasn't accurately checking video bitrate, it was reporting the entire container's bitrate as the audio. Now it reports as expected. 

Before: as per Issue #16248
Screenshot 2026-03-05 at 2 49 48 PM
 
After: 
Screenshot 2026-03-05 at 2 12 01 PM

This solves issue #16248

### DIFF
--- a/MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
+++ b/MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
@@ -968,20 +968,13 @@ namespace MediaBrowser.MediaEncoding.Probing
                 return null;
             }
 
+            // Get stream bitrate
             int bitrate = 0;
 
-            // Extract bitrate from "BPS" tag (FFPROBE per-stream data rate), only for Matroska.
-            if ((streamInfo.CodecType == CodecType.Audio || streamInfo.CodecType == CodecType.Video)
-                && formatInfo?.FormatName is not null
-                && formatInfo.FormatName.Contains("matroska", StringComparison.OrdinalIgnoreCase))
+            // Extract bitrate from "BPS" tag (FFPROBE per-stream data rate)
+            if (streamInfo.CodecType == CodecType.Audio || streamInfo.CodecType == CodecType.Video)
             {
                 bitrate = GetBPSFromTags(streamInfo);
-            }
-
-            // Use the standard stream bitrate field
-            if (bitrate <= 0 && int.TryParse(streamInfo.BitRate, CultureInfo.InvariantCulture, out var streamVal))
-            {
-                bitrate = streamVal;
             }
 
             // FALLBACK Calculate BPS from total bytes and duration tags
@@ -996,13 +989,16 @@ namespace MediaBrowser.MediaEncoding.Probing
                 }
             }
 
-            // Use the global format info (useful for FLAC or single-stream containers)
-            if (bitrate <= 0 && formatInfo != null)
+            // FALLBACK B: Use the standard stream bitrate field
+            if (bitrate <= 0 && int.TryParse(streamInfo.BitRate, CultureInfo.InvariantCulture, out var streamVal))
             {
-                if (int.TryParse(formatInfo.BitRate, CultureInfo.InvariantCulture, out var formatVal))
-                {
-                    bitrate = formatVal;
-                }
+                bitrate = streamVal;
+            }
+
+            // FALLBACK C: Use the global format info (useful for FLAC or single-stream containers)
+            if (bitrate <= 0 && formatInfo != null && int.TryParse(formatInfo.BitRate, CultureInfo.InvariantCulture, out var formatVal))
+            {
+                bitrate = formatVal;
             }
 
             if (bitrate > 0)
@@ -1228,19 +1224,23 @@ namespace MediaBrowser.MediaEncoding.Probing
 
         private static int GetBPSFromTags(MediaStreamInfo streamInfo)
         {
+            // 1. Check for null streamInfo or Tags
             if (streamInfo?.Tags is null)
             {
-                return 0;
+                return 0; // Changed from null
             }
 
+            // 2. Fetch the value from the dictionary
             var bps = GetDictionaryValue(streamInfo.Tags, "BPS-eng") ?? GetDictionaryValue(streamInfo.Tags, "BPS");
 
+            // 3. Try to parse and return the integer
             if (int.TryParse(bps, NumberStyles.Integer, CultureInfo.InvariantCulture, out var parsedBps))
             {
                 return parsedBps;
             }
 
-            return 0;
+            // 4. Final fallback if parsing fails
+            return 0; // Changed from null
         }
 
         private static double? GetRuntimeSecondsFromTags(MediaStreamInfo streamInfo)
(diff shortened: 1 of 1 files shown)


MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
```diff
@@ -969,54 +969,41 @@ private MediaStream GetMediaStream(bool isAudio, MediaStreamInfo streamInfo, Med
             }
 
             // Get stream bitrate
-            var bitrate = 0;
+            int bitrate = 0;
 
-            if (int.TryParse(streamInfo.BitRate, CultureInfo.InvariantCulture, out var value))
+            // Extract bitrate from "BPS" tag (FFPROBE per-stream data rate)
```
nyanmisaka: The custom tags `BPS(-eng)` and `NUMBER_OF_BYTES(-eng)` are primarily used only in Matroska containers and are generated using `mkvtoolnix`. FFmpeg does not use these tags.

In some cases, they are more accurate than the bitrate estimated per stream by `ffprobe`, but they are easier to edit, thus susceptible to tampering or misconfiguration. How about guarding these paths so they are only usable when the container is Matroska?

MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
```diff
@@ -969,54 +969,41 @@ private MediaStream GetMediaStream(bool isAudio, MediaStreamInfo streamInfo, Med
...
+
+                if (durationInSeconds is { } dur && dur >= 1 && bytes is { } totalBytes)
                 {
-                    bitrate = value;
+                    bitrate = Convert.ToInt32(totalBytes * 8 / dur, CultureInfo.InvariantCulture);
                 }
             }
 
-            if (bitrate > 0)
+            // FALLBACK B: Use the standard stream bitrate field
```
nyanmisaka: This is the standard way to get bitrate from non-Matroska containers using `ffprobe`. Additionally, Matroska files generated using `ffmpeg` also use this path. I don't think it's appropriate to call it a fallback because most containers use this path.

MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
```diff
@@ -1235,20 +1222,25 @@ private static void SetAudioRuntimeTicks(InternalMediaInfoResult result, MediaIn
...
+
+            // 3. Try to parse and return the integer
             if (int.TryParse(bps, NumberStyles.Integer, CultureInfo.InvariantCulture, out var parsedBps))
             {
                 return parsedBps;
             }
 
-            return null;
+            // 4. Final fallback if parsing fails
+            return 0; // Changed from null
```
nyanmisaka: These new comments 1-4 are a bit excessive. They are the literal meaning of the code.

nyanmisaka (changes requested):

Brady3035:
I apologize for the formatting, this is my first time contributing to Jellyfin, thanks for reviewing my code. :)

Brady3035:
Any reason this hasn't been merged? I made the requested changes.

gnattu:
I’m sorry, but I don’t believe it’s a good idea to blindly trust a user-generated, unverified metadata tag and consider its value more reliable than what ffprobe estimates. 

Even if the user doesn’t maliciously edit the metadata to report a wrong number, it’s extremely easy to create a video with incorrect metadata. Simply re-encode your video using ffmpeg and output it as an mkv file. The original BPS tag will be retained and will still display the bitrate of the **original video**, even though the processed video has a significantly different bitrate value. Considering the popularity of ffmpeg and the tools developed based on it, the number of mkv files carrying incorrect BPS metadata could be substantial.

Unfortunately, mediainfo blindly believes the metadata tag and will report the incorrect value if the BPS tag is not accurate. And yes, ffprobe would report a value that is not very accurate but the estimation at least comes from the file it is looking at, not from a metadata that might not even represent the status of current file.


To reliably calculate the bitrate we have to use some super slow method like `-show_entries packet=size` which will scan almost the whole file which is very slow and too heavy for this use case. This is just very unfortunate that a container as versatile as mkv does not even have a standard core per-stream bitrate/total size field and have to use a user generated tag to record bitrate.

Brady3035:
I see what you're saying, I have to believe that there's a better way to get the bitrate. I will do a bit more research and get back to you, thanks for taking the time to get back to me!

Brady3035:
@gnattu Is it a non-starter to rewrite this to use MediaInfo instead of FFPROBE? I did a bit of testing and it seems to be useful for this purpose but I do understand adding new dependencies isn't ideal. Thoughts? That would be a much more ambitious project.

Renari:
> @gnattu Is it a non-starter to rewrite this to use MediaInfo instead of FFPROBE? I did a bit of testing and it seems to be useful for this purpose but I do understand adding new dependencies isn't ideal. Thoughts? That would be a much more ambitious project.

This is why I originally stopped working on #16248, I thought jellyfin shipped with mediainfo as a dependency but found out that it doesn't and was of the opinion that adding a new dependency on it to fix this issue wouldn't go over well.

Brady3035:
Ok I changed things a bit here, I got rid of the BPS tag due to our discussion above. I found that our per-stream ffprobe calculation was more than accurate enough but it wasn't being parsed by TimeSpan.TryParse(duration, out var parsedDuration) in line 1257 in my commit because TimeSpan.TryParse doesn't support 9 decimal places (nanosecond precision)  which is what ffprobe was returning for the duration. My solution was to regex it down to 7 decimal places which allows it to parse the duration in seconds and thus calculate the bitrate accurately. I tested on a few different containers, and codecs and it seems to work properly. When it comes across codecs ffprobe can't read like interlaced video the previous fallbacks seem to work. Is there anything else that can be improved here?

MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
```diff
@@ -971,7 +971,24 @@ private MediaStream GetMediaStream(bool isAudio, MediaStreamInfo streamInfo, Med
             // Get stream bitrate
             var bitrate = 0;
 
-            if (int.TryParse(streamInfo.BitRate, CultureInfo.InvariantCulture, out var value))
+            // Get average bitrate info from tag "NUMBER_OF_BYTES" and "DURATION" if possible.
```
gnattu, line 974: But aren't these still matroska-specific tags which also carried all the problem BPS tags have?

theguymadmax:
Superseded by #17170

PR #16354, opened 2026-03-06
Fix DbUpdateConcurrencyException in AuthenticateByName with NoLock

### LINKED ISSUE #16353: AuthenticateByName fails with DbUpdateConcurrencyException when LockingBehavior is NoLock
## Description

All `POST /Users/AuthenticateByName` requests fail with `DbUpdateConcurrencyException` when the database `LockingBehavior` is set to `NoLock` (the default after migration to 10.11.x). Authentication credentials are validated successfully, but the subsequent token creation step crashes due to a race condition on the `Users` table `RowVersion` column.

This makes it **impossible to log in** from any client — web browser, Swiftfin, Infuse, etc. Existing sessions with cached tokens continue to work until the token is invalidated.

## Reproduction Steps

1. Fresh install or migrate to Jellyfin 10.11.6 (database migrated to EF Core)
2. `database.xml` has `<LockingBehavior>NoLock</LockingBehavior>` (the default)
3. Have multiple devices registered in the `Devices` table (normal usage over time)
4. Attempt to log in from any client

## Observed Behavior

Authentication succeeds, but the login flow crashes at the token creation step:

```
[INF] Authentication request for <user> has succeeded.
[INF] Current/Max sessions for user <user>: 1/0
[INF] Logging out access token <old-token>
[INF] Creating new access token for user <user-id>
[ERR] Error trying to save changes.
Microsoft.EntityFrameworkCore.DbUpdateConcurrencyException: The database operation was expected to affect 1 row(s), but actually affected 0 row(s); data may have been modified or deleted since entities were loaded.
```

Full stack trace:

```
Microsoft.EntityFrameworkCore.DbUpdateConcurrencyException: The database operation was expected to affect 1 row(s), but actually affected 0 row(s); data may have been modified or deleted since entities were loaded.
   at Microsoft.EntityFrameworkCore.Update.AffectedCountModificationCommandBatch.ThrowAggregateUpdateConcurrencyExceptionAsync(...)
   at Microsoft.EntityFrameworkCore.Update.AffectedCountModificationCommandBatch.ConsumeResultSetWithRowsAffectedOnlyAsync(...)
   at Microsoft.EntityFrameworkCore.Update.AffectedCountModificationCommandBatch.ConsumeAsync(...)
   at Microsoft.EntityFrameworkCore.Update.ReaderModificationCommandBatch.ExecuteAsync(...)
   at Microsoft.EntityFrameworkCore.Update.Internal.BatchExecutor.ExecuteAsync(...)
   at Microsoft.EntityFrameworkCore.Storage.RelationalDatabase.SaveChangesAsync(...)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.StateManager.SaveChangesAsync(...)
   at Microsoft.EntityFrameworkCore.DbContext.SaveChangesAsync(...)
   at Jellyfin.Database.Implementations.JellyfinDbContext.<>c__DisplayClass62_0.<<SaveChangesAsync>b__0>d.MoveNext()
--- End of stack trace from previous location ---
   at Jellyfin.Database.Implementations.Locking.NoLockBehavior.OnSaveChangesAsync(JellyfinDbContext context, Func`1 saveChanges)
   at Jellyfin.Database.Implementations.JellyfinDbContext.SaveChangesAsync(...)
```

Followed immediately by:

```
[ERR] Error processing request. URL POST /Users/AuthenticateByName.
```

## Root Cause Analysis

The authentication flow performs two sequential database writes against the `Users` table:

1. **Log out old access token** — this updates the user record and increments `RowVersion`
2. **Create new access token** — this also tries to update the user record, but the `RowVersion` it loaded is now stale

With `NoLockBehavior`, these operations are not serialized, so the EF Core concurrency check (`WHERE RowVersion = @expected`) fails because step 1 already changed the `RowVersion` that step 2 is checking against.

The problem is **amplified** when multiple threads handle concurrent auth requests for the same user (visible in logs as different thread IDs `[17]`, `[41]` etc. processing simultaneous `AuthenticateByName` requests at the same timestamp). Each thread logs out the other's token, further racing the `RowVersion`.

Once this state is reached, it becomes **self-reinforcing** — every subsequent login attempt fails, the `RowVersion` keeps climbing (observed values of 93+ and 334+ within hours), and clients retry aggressively, making it worse.

## Workaround (being tested)

Changing `database.xml` from:

```xml
<LockingBehavior>NoLock</LockingBehavior>
```

to:

```xml
<LockingBehavior>Pessimistic</LockingBehavior>
```

This serializes concurrent writes so the "logout old token" and "create new token" operations don't race against each other on the `RowVersion` column.

Additionally, to recover from the broken state, stop Jellyfin and run:

```sql
DELETE FROM Devices;
UPDATE Users SET RowVersion = 0 WHERE Id = '<user-id>';
```

Without changing the `LockingBehavior`, the manual DB cleanup only provides temporary relief (~12 hours before the race condition triggers again).

## Suggested Fix

The authentication flow in `SessionManager` should either:

1. Perform the "logout old token" and "create new token" operations within a single atomic transaction/save, so only one `RowVersion` increment occurs
2. Or reload the entity after the logout step before attempting the token creation write
3. Or set the default `LockingBehavior` to something other than `NoLock` for SQLite deployments where concurrent write safety is needed

## Environment

- **Jellyfin version**: 10.11.6
- **Image**: `lscr.io/linuxserver/jellyfin:10.11.6ubu2404-ls22`
- **Database**: SQLite (Jellyfin-SQLite, EF Core migration from 10.10.x)
- **OS**: macOS (Docker Desktop)
- **Clients affected**: All (Jellyfin Web, Swiftfin, Infuse)

johnib:
## Summary

- Fixes `DbUpdateConcurrencyException` that makes login impossible when `LockingBehavior` is `NoLock` (the default after migration to 10.11.x)
- Adds retry logic with client-wins concurrency resolution to `UpdateUserInternalAsync` in `UserManager`
- On `RowVersion` conflict, reloads DB state (refreshing the concurrency token) while preserving the caller's in-memory property changes (`LastActivityDate`, `LastLoginDate`, etc.)

## Root Cause

The `User` entity is a shared in-memory singleton (`UserManager._users` dictionary). During authentication, multiple operations call `UpdateUserAsync` which creates short-lived `DbContext` instances, attaches the same `User` object, and saves. When concurrent auth requests (or even sequential steps within a single auth flow) race, the `[ConcurrencyCheck]`-annotated `RowVersion` becomes stale between `Attach` and `SaveChangesAsync`, causing the exception.

## Test plan

- [ ] Verify login works with `LockingBehavior` set to `NoLock` (the default)
- [ ] Verify concurrent logins for the same user from multiple clients succeed
- [ ] Verify `LastActivityDate` and `LastLoginDate` are correctly persisted after login
- [ ] Verify existing behavior is unchanged with `Pessimistic` and `Optimistic` locking modes

Fixes #16353

### DIFF
--- a/Jellyfin.Server.Implementations/Users/UserManager.cs
+++ b/Jellyfin.Server.Implementations/Users/UserManager.cs
@@ -886,7 +886,7 @@ namespace Jellyfin.Server.Implementations.Users
         private async Task UpdateUserInternalAsync(JellyfinDbContext dbContext, User user)
         {
             const int MaxRetries = 3;
-            for (int attempt = 0; attempt < MaxRetries; attempt++)
+            for (int attempt = 1; attempt <= MaxRetries; attempt++)
             {
                 try
                 {
@@ -901,8 +901,9 @@ namespace Jellyfin.Server.Implementations.Users
                     await dbContext.SaveChangesAsync().ConfigureAwait(false);
                     return;
                 }
-                catch (DbUpdateConcurrencyException) when (attempt < MaxRetries - 1)
+                catch (DbUpdateConcurrencyException ex) when (attempt < MaxRetries)
                 {
+                    _logger.LogWarning(ex, "DbUpdateConcurrencyException on attempt {Attempt} of {MaxRetries} for user {UserId}. Retrying.", attempt, MaxRetries, user.Id);
                     // Another concurrent operation modified the User row, causing the RowVersion
                     // concurrency check to fail. Refresh the original values from the database
                     // while preserving our in-memory property changes (last-writer-wins).
--- a/Jellyfin.Server.Implementations/Users/UserManager.cs
+++ b/Jellyfin.Server.Implementations/Users/UserManager.cs
@@ -885,10 +885,34 @@ namespace Jellyfin.Server.Implementations.Users
 
         private async Task UpdateUserInternalAsync(JellyfinDbContext dbContext, User user)
         {
-            dbContext.Users.Attach(user);
-            dbContext.Entry(user).State = EntityState.Modified;
-            _users[user.Id] = user;
-            await dbContext.SaveChangesAsync().ConfigureAwait(false);
+            const int MaxRetries = 3;
+            for (int attempt = 0; attempt < MaxRetries; attempt++)
+            {
+                try
+                {
+                    var entry = dbContext.Entry(user);
+                    if (entry.State == EntityState.Detached)
+                    {
+                        dbContext.Users.Attach(user);
+                        entry.State = EntityState.Modified;
+                    }
+
+                    _users[user.Id] = user;
+                    await dbContext.SaveChangesAsync().ConfigureAwait(false);
+                    return;
+                }
+                catch (DbUpdateConcurrencyException) when (attempt < MaxRetries - 1)
+                {
+                    // Another concurrent operation modified the User row, causing the RowVersion
+                    // concurrency check to fail. Refresh the original values from the database
+                    // while preserving our in-memory property changes (last-writer-wins).
+                    var entry = dbContext.Entry(user);
+                    var currentValues = entry.CurrentValues.Clone();
+                    await entry.ReloadAsync().ConfigureAwait(false);
+                    entry.CurrentValues.SetValues(currentValues);
+                    entry.State = EntityState.Modified;
+                }
+            }
         }
     }
 }
--- a/tests/Jellyfin.Server.Implementations.Tests/Users/UserUpdateConcurrencyTests.cs
+++ b/tests/Jellyfin.Server.Implementations.Tests/Users/UserUpdateConcurrencyTests.cs
@@ -133,13 +133,20 @@ public sealed class UserUpdateConcurrencyTests : IDisposable
         }
 
         // Step 3: simulate LogSessionActivity's UpdateUserAsync
-        user.LastActivityDate = DateTime.UtcNow;
+        var expectedDate = DateTime.UtcNow;
+        user.LastActivityDate = expectedDate;
         await using (var ctx2 = CreateDbContext())
         {
             ctx2.Users.Attach(user);
             ctx2.Entry(user).State = EntityState.Modified;
             await ctx2.SaveChangesAsync();
         }
+
+        // Verify the final update was persisted.
+        await using var verifyCtx = CreateDbContext();
+        var savedUser = await verifyCtx.Users.FindAsync(user.Id);
+        Assert.NotNull(savedUser);
+        Assert.Equal(expectedDate, savedUser!.LastActivityDate);
     }
 
     /// <summary>
--- a/tests/Jellyfin.Server.Implementations.Tests/Users/UserUpdateConcurrencyTests.cs
+++ b/tests/Jellyfin.Server.Implementations.Tests/Users/UserUpdateConcurrencyTests.cs
@@ -1,0 +1,307 @@
+using System;
+using System.Collections.Concurrent;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using Jellyfin.Database.Implementations;
+using Jellyfin.Database.Implementations.Entities;
+using Jellyfin.Database.Implementations.Locking;
+using Jellyfin.Database.Providers.Sqlite;
+using Jellyfin.Server.Implementations.Users;
+using MediaBrowser.Common;
+using MediaBrowser.Common.Net;
+using MediaBrowser.Controller.Authentication;
+using MediaBrowser.Controller.Configuration;
+using MediaBrowser.Controller.Drawing;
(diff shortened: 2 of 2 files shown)


breneser:
@johnib Do you have the custom image accessible by any chance? I am having the exact same issue, could give it a try

johnib:
@breneser Sure! I've published a patched image based on `linuxserver/jellyfin:10.11.6ubu2404-ls22` with this fix applied:

```
docker pull ghcr.io/johnib/jellyfin:10.11.6-fix-auth-concurrency
```

Available for both `amd64` and `arm64`. It's a drop-in replacement — only the `Jellyfin.Server.Implementations.dll` is replaced on top of the stock image.

I've been running it for a few hours now with no recurrence so far. Let me know how it goes for you.

johnib:
So far this image had proven to over come this issue. @breneser

Jellyfin.Server.Implementations/Users/UserManager.cs
```diff
@@ -885,10 +885,34 @@ private async Task IncrementInvalidLoginAttemptCount(User user)
...
+                        dbContext.Users.Attach(user);
+                        entry.State = EntityState.Modified;
+                    }
+
+                    _users[user.Id] = user;
+                    await dbContext.SaveChangesAsync().ConfigureAwait(false);
+                    return;
+                }
+                catch (DbUpdateConcurrencyException) when (attempt < MaxRetries - 1)
+                {
```
tudoranghelina91, line 905: Would logging help a bit for auditing purposes?

Jellyfin.Server.Implementations/Users/UserManager.cs
```diff
@@ -885,10 +885,34 @@ private async Task IncrementInvalidLoginAttemptCount(User user)
...
+                    {
+                        dbContext.Users.Attach(user);
+                        entry.State = EntityState.Modified;
+                    }
+
+                    _users[user.Id] = user;
+                    await dbContext.SaveChangesAsync().ConfigureAwait(false);
+                    return;
+                }
+                catch (DbUpdateConcurrencyException) when (attempt < MaxRetries - 1)
```
tudoranghelina91: I think it would communicate intent better if the loop would start from one

Jellyfin.Server.Implementations/Users/UserManager.cs
```diff
@@ -885,10 +885,34 @@ private async Task IncrementInvalidLoginAttemptCount(User user)
 
         private async Task UpdateUserInternalAsync(JellyfinDbContext dbContext, User user)
         {
-            dbContext.Users.Attach(user);
-            dbContext.Entry(user).State = EntityState.Modified;
-            _users[user.Id] = user;
-            await dbContext.SaveChangesAsync().ConfigureAwait(false);
+            const int MaxRetries = 3;
+            for (int attempt = 0; attempt < MaxRetries; attempt++)
```
tudoranghelina91: Could an approach where we retrieve the user's concurrency token and do a comparison work? Or is it better performance-wise to just let it fail the fourth time?

tudoranghelina91 (commented):
Just some minor comments

Jellyfin.Server.Implementations/Users/UserManager.cs
johnib: The retry-on-exception approach is the standard EF Core pattern for optimistic concurrency. Pre-fetching the token would add an extra DB round-trip on every save just to avoid an exception that rarely happens — and you'd still need to reload and retry anyway.

Jellyfin.Server.Implementations/Users/UserManager.cs
johnib: Done, changed to 1-based loop.

Jellyfin.Server.Implementations/Users/UserManager.cs
johnib: Added a warning-level log on each retry with attempt number and user ID.

johnib:
It's been a week since this fix was deployed on our custom image and we haven't had a single recurrence of the issue.

@breneser Have you had the chance to try the patched image? Would love to hear if it resolved the issue on your end as well.

pandrez:
@johnib I’m also having the same problem.
I’ll test your image and let you know how it’s working in few days.
Thank you for your fix!

tykateetee:
Also testing the image and will report back if it is working in a few days. Thanks @johnib for the (hopeful) fix!

fOO223Fr:
Facing the same problem.

Testing this image now:
`docker pull ghcr.io/johnib/jellyfin:10.11.6-fix-auth-concurrency`
thanks @johnib for vibing it ;) 

will update with results in few days

theguymadmax:
Superseded by #15368

Issue #16179, opened 2026-02-03
Dolby Vision manifest should use spec-compliant HLS manifest 


DrHurt:
### Description of the bug

When JellyFin converts DV P5 mkv to HLS, it correctly tags the bitstream dvh1 however the HLS manifest contains CODECS=hvc1

This causes my Apple TV to switch to HDR mode instead of DV mode.

Apple HLS specs require CODECS=dvh1 for P5

This could potentially have a detrimental impact on other devices too, cause them to setup an incorrect playback pipeline.

Manually changing the value to dvh1 causes the TV to switch correctly with P5.

### Reproduction steps

Play MKV + DV P5 with a client that only supports mp4 (eg. Apple TV)

### What is the current _bug_ behavior?

tvOS & TV change to HDR mode because of incorrect signalling in the master.m3u8

### What is the expected _correct_ behavior?

master.m3u8 MUST use CODECS = dvh1 with DV P5

JellyFin is generally specs compliant and works very well with Apple devices. This is the first deviation from compliance in my extensive testing.

### Jellyfin Server version

10.11.6

### Environment

```markdown
- OS: macOS
```

### Jellyfin logs

```shell
#EXTM3U
#EXT-X-STREAM-INF:BANDWIDTH=26502305,AVERAGE-BANDWIDTH=26502305,VIDEO-RANGE=PQ,CODECS="hvc1.2.4.L153.B0,ec-3",RESOLUTION=1920x1080,FRAME-RATE=59.94
main.m3u8?
```

DrHurt:
@gnattu Can you please have a look. I believe you've worked on this before.

nyanmisaka:
The DV P5 does not include any HDR10 compatibility layer. Therefore, according to Apple documentation, SUPPLEMENTAL-CODECS should not be added.

https://developer.apple.com/documentation/http-live-streaming/hls-authoring-specification-for-apple-devices-appendixes#The-SUPPLEMENTAL-CODECS-attribute

https://developer.apple.com/documentation/http-live-streaming/hls-authoring-specification-for-apple-devices-appendixes#Example-playlist-with-SDR-Dolby-Vision-HDR10-and-HLG-content-at-resolutions-from-720p-to-4K

<img width="1000" alt="Image" src="https://github.com/user-attachments/assets/f20dea6c-cdc4-4c8f-982b-4e6206c805f4" />

DrHurt:
[...]

Issue #16092, opened 2026-01-23
Update 10.11.6 ruins Dolby Vision video playback on WebOS


WinaxPL:
### Description of the bug

My setup:
Player: Jellyfin App on WebOS LG C4 (local network via Ethernet)
Jellyfin server: Jellyfin Container App on TrueNas Scale v1.2.23
Transcoding options:
Hardware acceleration: NVENC
Enable hardware decoding for:
H264, HEVC, MPEG4, HEVC 10bit, VP9 10bit, HEVC Rext 8/10bit
Enable enhanced NVDEC decoder - enabled
Enable hardware encoding - enabled

After updating Jellyfin on the TrueNas Scale server in the container application from version 11.10.5 to version 10.11.6, .mkv films in Dolby Vision format are played in SDR format with the ‘prefer fMP4-HLS container’ option disabled (the server transcodes the file). When this option is enabled, the situation is the same, but not every time. In the previous version, with this option enabled, Dolby Vision videos would start in HDR mode with remuxing enabled, but would not transcode. I can only watch Dolby Vision videos with the ‘prefer fMP4-HLS container’ option disabled. However, sometimes the same video can start in SDR (the server transcodes). This did not happen in the previous version.

### Reproduction steps

1. I launch the Jellyfin application on WebOS.
2. Run "Tron.Ares.2025.IMAX.2160p.10bit.HDR.DV.WEBRip.6CH.x265.HEVC-PSA.mkv" file (source: https://psa.wf/movie/tron-ares-2025/)
3. The server begins transcoding.

### What is the current _bug_ behavior?

The television does not display the Dolby Vision or HDR indicator when a film is started.

### What is the expected _correct_ behavior?

When you play a film in Dolby Vision format, the television plays the film in Dolby Vision mode without transcoding.

### Jellyfin Server version

10.11.6

### Specify the build version

10.11.6

### Environment

```markdown
- OS: TrueNas Scale 25.04
- Linux Kernel: 6.12.15-production+truenas
- Virtualization: Application (Kubernetes)
- Clients: WebOS Jellyfin Web 10.11.6
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- CPU Model:
- GPU Model:
- Plugins:
- Reverse Proxy: NGINX
- Base URL:
- Networking:
- Jellyfin Data Storage & Filesystem:
- Media Storage & Filesystem:
- External Integrations:
```

### Jellyfin logs

```shell
[...]

PR #16361, opened 2026-03-06
Skip mp4toannexb bitstream filters for fMP4 HLS segments during stream copy

NoFear0411:
## Summary

`DynamicHlsController.GetVideoArguments` applies `mp4toannexb` bitstream filters (`hevc_mp4toannexb` for HEVC, `h264_mp4toannexb` for H.264) unconditionally when copy-streaming to HLS, without checking the segment container type. For HEVC this converts HVCC extradata to Annex B format, which the fMP4 muxer must then reconstruct back into an hvcC box for the init segment. The round-trip corrupts the codec configuration and breaks Dolby Vision signaling (the dvvC box depends on a correct hvcC), causing playback failures on devices receiving fMP4 HLS segments. The same unnecessary conversion applies to H.264 with AVCC-to-Annex B, though without the DoVi breakage.

TS segments genuinely need Annex B conversion — the TS muxer requires start-code-delimited NALs. fMP4 segments use the same MP4 NAL format as the source, so the conversion is both unnecessary and harmful.

`EncodingHelper` already guards this correctly for progressive streams (`OutputContainer == "ts"` at line 7549), but `DynamicHlsController` was missing the equivalent check despite having `segmentContainer` available as a parameter.

### References

- **FFmpeg `hevc_mp4toannexb` docs** — ["Converts an HEVC/H.265 bitstream from length-prefixed mode to start code prefixed mode, as required by some streaming formats like MPEG-2 transport stream."](https://ffmpeg.org/ffmpeg-bitstream-filters.html#hevc_005fmp4toannexb) — explicitly scoped to TS, not MP4/fMP4.
- **ISO 14496-12 (ISOBMFF)** — fMP4 stores NAL units in length-prefixed format with codec configuration in `hvcC` box, same as MKV's `HEVCDecoderConfigurationRecord`. No Annex B conversion involved.
- **ISO 14496-15 §8.3.2** — Defines `hvcC` (HEVC Configuration Box) using `HEVCDecoderConfigurationRecord` with `lengthSizeMinusOne` for NAL unit length fields. This is the format both MKV and fMP4 use natively.
- **MPEG-2 TS (ISO 13818-1)** — Transport stream PES packets carry HEVC in Annex B format (start code `0x00000001` delimited NALs with in-band parameter sets), which is why `hevc_mp4toannexb` exists.
- **Dolby Vision ISOBMFF spec** — The `dvvC`/`dvwC` configuration boxes are stored alongside `hvcC` in the sample entry. When `hevc_mp4toannexb` strips the HVCC extradata to Annex B, the fMP4 muxer reconstructs `hvcC` from in-band parameter sets but cannot reconstruct the `dvvC` box, losing DoVi signaling.
- **EncodingHelper.cs line 7549** — Jellyfin's own progressive stream path already applies the same TS-only guard: `string.Equals(state.OutputContainer, "ts", ...)`. The HLS path was inconsistent.

## Changes

- **`DynamicHlsController.cs`**: Add `segmentContainer` check around the BSF block:
  - TS: unchanged — full BSF chain (`h264_mp4toannexb`/`hevc_mp4toannexb` format conversion + optional metadata removal)
  - fMP4: skip all `mp4toannexb` format conversion, apply only metadata removal when the client cannot handle DoVi or HDR10+

- **`EncodingHelper.cs`**: Add `GetDynamicHdrMetadataRemovalBitStreamArgs` — returns only the metadata removal BSF arguments without the `hevc_mp4toannexb` prefix. The CBS-based filters (`hevc_metadata`, `dovi_rpu`) can parse MP4-format packets directly via HVCC extradata detection and do not depend on `hevc_mp4toannexb` as a prerequisite.

## Impact

| Scenario | TS segments | fMP4 segments |
|----------|-------------|---------------|
| HEVC copy, no metadata removal needed | `hevc_mp4toannexb` | No BSFs (passthrough) |
| HEVC copy, client needs DoVi stripped | `hevc_mp4toannexb,hevc_metadata=remove_dovi=1` | `hevc_metadata=remove_dovi=1` |
| HEVC copy, client needs HDR10+ stripped | `hevc_mp4toannexb,hevc_metadata=remove_hdr10plus=1` | `hevc_metadata=remove_hdr10plus=1` |
| AV1 copy, no metadata removal needed | No BSFs | No BSFs |
| AV1 copy, client needs DoVi stripped | `av1_metadata=remove_dovi=1` | `av1_metadata=remove_dovi=1` |
| AV1 copy, client needs HDR10+ stripped | `av1_metadata=remove_hdr10plus=1` | `av1_metadata=remove_hdr10plus=1` |
| H264 copy | `h264_mp4toannexb` | No BSFs (passthrough) |
| Any codec, NalLengthSize == 0 (Annex B source) | No BSFs (block skipped) | No BSFs (block skipped) |
| Transcode (any codec) | No change — BSF block not entered | No change — BSF block not entered |

## Test plan

- [ ] Play DoVi P5 MKV via fMP4 HLS (copy codec) — verify no `hevc_mp4toannexb` in FFmpeg command, dvvC box intact in init segment
- [ ] Play DoVi P8 MKV via fMP4 HLS (copy codec) — same verification
- [ ] Play HEVC MKV via TS HLS (copy codec) — verify `hevc_mp4toannexb` still present
- [ ] Play HEVC content where DoVi must be stripped (client lacks support) via fMP4 — verify standalone `hevc_metadata=remove_dovi=1` without `hevc_mp4toannexb` prefix
- [ ] Play H264 content via TS HLS — verify `h264_mp4toannexb` still applied

Relates to #16092
Relates to #16179

### DIFF
--- a/Jellyfin.Api/Controllers/DynamicHlsController.cs
+++ b/Jellyfin.Api/Controllers/DynamicHlsController.cs
@@ -1837,13 +1837,26 @@ public class DynamicHlsController : BaseJellyfinApiController
         // See if we can save come cpu cycles by avoiding encoding.
         if (EncodingHelper.IsCopyCodec(codec))
         {
-            // If h264_mp4toannexb is ever added, do not use it for live tv.
             if (state.VideoStream is not null && !string.Equals(state.VideoStream.NalLengthSize, "0", StringComparison.OrdinalIgnoreCase))
             {
-                string bitStreamArgs = _encodingHelper.GetBitStreamArgs(state, MediaStreamType.Video);
-                if (!string.IsNullOrEmpty(bitStreamArgs))
+                if (string.Equals(segmentContainer, "ts", StringComparison.OrdinalIgnoreCase))
                 {
-                    args += " " + bitStreamArgs;
+                    // TS segments require Annex B NAL format; apply format conversion and optional metadata removal.
+                    string bitStreamArgs = _encodingHelper.GetBitStreamArgs(state, MediaStreamType.Video);
+                    if (!string.IsNullOrEmpty(bitStreamArgs))
+                    {
+                        args += " " + bitStreamArgs;
+                    }
+                }
+                else
+                {
+                    // fMP4 segments use MP4 NAL format natively, no format conversion needed.
+                    // Still apply metadata removal if the client cannot handle DoVi or HDR10+.
+                    string metadataRemovalArgs = _encodingHelper.GetDynamicHdrMetadataRemovalBitStreamArgs(state);
+                    if (!string.IsNullOrEmpty(metadataRemovalArgs))
+                    {
+                        args += " " + metadataRemovalArgs;
+                    }
                 }
             }
 
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -1515,6 +1515,48 @@ namespace MediaBrowser.Controller.MediaEncoding
             return null;
         }
 
+        public string GetDynamicHdrMetadataRemovalBitStreamArgs(EncodingJobInfo state)
+        {
+            if (state?.VideoStream is null)
+            {
+                return null;
+            }
+
+            var stream = state.VideoStream;
+
+            if (IsH265(stream))
+            {
+                switch (ShouldRemoveDynamicHdrMetadata(state))
+                {
+                    case DynamicHdrMetadataRemovalPlan.RemoveDovi:
+                        return _mediaEncoder.SupportsBitStreamFilterWithOption(BitStreamFilterOptionType.HevcMetadataRemoveDovi)
+                            ? "-bsf:v hevc_metadata=remove_dovi=1"
+                            : "-bsf:v dovi_rpu=strip=1";
+                    case DynamicHdrMetadataRemovalPlan.RemoveHdr10Plus:
+                        return "-bsf:v hevc_metadata=remove_hdr10plus=1";
+                    default:
+                        return null;
+                }
+            }
+
+            if (IsAv1(stream))
+            {
+                switch (ShouldRemoveDynamicHdrMetadata(state))
+                {
+                    case DynamicHdrMetadataRemovalPlan.RemoveDovi:
+                        return _mediaEncoder.SupportsBitStreamFilterWithOption(BitStreamFilterOptionType.Av1MetadataRemoveDovi)
+                            ? "-bsf:v av1_metadata=remove_dovi=1"
+                            : "-bsf:v dovi_rpu=strip=1";
+                    case DynamicHdrMetadataRemovalPlan.RemoveHdr10Plus:
+                        return "-bsf:v av1_metadata=remove_hdr10plus=1";
+                    default:
+                        return null;
+                }
+            }
+
+            return null;
+        }
+
         public string GetAudioBitStreamArguments(EncodingJobInfo state, string segmentContainer, string mediaSourceContainer)
         {
             var bitStreamArgs = string.Empty;


DrHurt:
@NoFear0411 Should it not skip h264_mp4toannexb in fMP4 as well? (while we're at it)

NoFear0411:
@DrHurt Good catch but it already does. The fix moves the entire `GetBitStreamArgs()` call (which returns both h264_mp4toannexb and hevc_mp4toannexb) behind a `segmentContainer == "ts"` guard. The fMP4 branch only calls `GetDynamicHdrMetadataRemovalBitStreamArgs()` for optional DoVi/HDR10+ metadata stripping. So neither BSF is
  applied to fMP4 segments. I've edited the title and text to be more clear about it.

DrHurt:
Here’s hoping this gets merged into the next release! It’s relatively low risk and provides a meaningful improvement.

NoFear0411:
@DrHurt I hope so too. Also, https://github.com/jellyfin/jellyfin-web/pull/7659 and https://github.com/jellyfin/jellyfin/pull/16362 would give us that sweet spec compliant DoVi playback.

Jellyfin.Api/Controllers/DynamicHlsController.cs
```diff
@@ -1837,13 +1837,26 @@ private string GetVideoArguments(StreamState state, int startNumber, bool isEven
...
+                    if (!string.IsNullOrEmpty(bitStreamArgs))
+                    {
+                        args += " " + bitStreamArgs;
+                    }
+                }
+                else
+                {
+                    // fMP4 segments use MP4 NAL format natively, no format conversion needed.
+                    // Still apply metadata removal if the client cannot handle DoVi or HDR10+.
+                    string metadataRemovalArgs = _encodingHelper.GetDynamicHdrMetadataRemovalBitStreamArgs(state);
```
gnattu, line 1855: Why you are doing this here? We already removed metadata elsewhere as we are doing bitstream filtering as a pipeline, so you are end up calling metadata removal twice. Is there anything I missed?

DrHurt:
> Before we modify the behavior I genuinely want to know how converting mp4 format twice (from and to nal) would corrupt dolby vision. If that is the case it means there is an upstream ffmpeg bug that needs to be fixed.
> 
> Yes double conversion is not efficient but this should not break playback at all.

I haven’t done any testing, but ffmpeg dv support is still unofficial so I wouldn’t be surprised if it’s buggy. 

Efficiencies aside, I think generally the less we mess with streams, the better. Less potential for unexpected behaviour.

NoFear0411:
Today I ran FFmpeg tests with jellyfin-ffmpeg 7.1.3 and fresh test files, to verify whether "hevc_mp4toannexb" actually corrupts Dolby Vision signaling in fMP4 init segments. The results show my original premise was wrong. The previous files I used to test that led to this PR, were with malformed metadata from the start... Apologies, I should sleep before creating PR.

Test setup                                                                                                                                                                                                                                                                                                                
                  
For each source, generated 10s of fMP4 HLS, once with -bsf:v hevc_mp4toannexb and once without. Used the same flags Jellyfin generates for a stream copy:

  ```
/opt/jellyfin-ffmpeg/ffmpeg -analyzeduration 200M -probesize 1G -fflags +genpts -t 10 \
    -i file:"<source>" -map_metadata -1 -map_chapters -1 -threads 0 \
    -map 0:0 -map 0:1 -map -0:s -codec:v:0 copy -tag:v:0 dvh1 -strict -2 \
    [-bsf:v hevc_mp4toannexb] -start_at_zero -codec:a:0 copy \
    -copyts -avoid_negative_ts disabled -max_muxing_queue_size 2048 \
    -f hls -max_delay 5000000 -hls_time 6 -hls_segment_type fmp4 \
    -hls_fmp4_init_filename "init.mp4" -start_number 0 \
    -hls_playlist_type vod -hls_list_size 0 \
    -hls_segment_options movflags=+frag_discont -y output.m3u8
```

Inspected init segments with ffprobe -v trace for hvcC and dvcC/dvvC boxes.

Results

  | Source                           | BSF     | hvcC                             | dvcC/dvvC        | Codec params                    |
  |----------------------------------|---------|----------------------------------|------------------|---------------------------------|
  | MP4 DoVi P5 (dvhe, sparse hvcC)  | with    | 2554 bytes (rebuilt from stream) | 32 bytes, intact | OK                              |
  | MP4 DoVi P5                      | without | 31 bytes (empty, numOfArrays=0)  | 32 bytes, intact | FAIL - unspecified pixel format |
  | MKV DoVi P8.1 (V_MPEGH/ISO/HEVC) | with    | 140 bytes                        | 32 bytes, intact | OK                              |
  | MKV DoVi P8.1                    | without | 140 bytes                        | 32 bytes, intact | OK                              |

Findings

  1. Dolby Vision signaling (dvcC/dvvC) is preserved in all cases; The BSF does not corrupt it.
  2. MKV sources: Identical init segments with or without the BSF. MKV stores the full HEVCDecoderConfigurationRecord (VPS/SPS/PPS), so the Annex B round-trip is lossless.
  3. MP4 sources with sparse hvcC (parameter sets in-band only, e.g. repeat-headers encoded P5): The BSF fixes the init segment by reconstructing a complete hvcC from in-band parameter sets. Without it, the init segment has an empty hvcC (31 bytes) and ffprobe cannot determine codec parameters.
  4. Real playback: DoVi P5 MKV --> fMP4 HLS with BSF, without HLS short segment --> LG C4 webOS triggers Dolby Vision and Atmos correctly (if and when the https://github.com/jellyfin/jellyfin-webos/pull/331 bug reports valid DoVi support for the device.)

Conclusion:

Sadly the BSF is redundant for MKV sources but harmless, and actively beneficial for MP4 sources with sparse extradata. Removing it would cause a regression for those files. I think this PR should be closed.

Jellyfin.Api/Controllers/DynamicHlsController.cs
NoFear0411: There is no duplication. I traced every call site:                                                                                                                                                                                                                                                                        
   
  - GetBitStreamArgs(Video) is called in two places: DynamicHlsController.cs:1843 (HLS path) and EncodingHelper.cs:7552 (progressive path). These are mutually exclusive code paths.                                                                                                                                        
  - ShouldRemoveDynamicHdrMetadata() is called at EncodingHelper.cs:2409, but that's the copy-eligibility check (CanEncoderRemoveDynamicHdrMetadata) — it returns false to reject copy, it doesn't build any BSF arguments.
  - No other code in the pipeline constructs hevc_metadata, av1_metadata, or dovi_rpu BSF strings.

In this PR, the fMP4 branch calls GetDynamicHdrMetadataRemovalBitStreamArgs() instead of GetBitStreamArgs(), not in addition to it. They are two mutually exclusive branches gated by the segmentContainer check.

That said, see my comment below about the test results. The PR has a different problem.

DrHurt:
Thanks for the very extensive testing!

It’s mostly used for mkv sources for devices that don’t support mkv like Apple TV & LG webOS.

So I still think this is a useful change to avoid redundancy & improve efficiency!

Up to you whether you want this reopened.

NoFear0411:
LG webOS introduced DoVi playback from mkv sources, but from my initial testing I still think it is buggy and it is only supported on webOS 25 with specific models:

- 2025 Models | OLED (G5, C5, B5), QNED (92A, 85A), LCD | Shipped with webOS 25 (May 2025)
- 2024 Models | OLED (G4, C4, B4), QNED, LCD | Started Q4 2025 (Now available)
- 2023 Models | OLED (G3, C3, B3), QNED, LCD | Started Late 2025 / Q1 2026
- 2022 Models | OLED (G2, C2, B2), 8K QNED, StanbyME Go | Started Q1 2026
- 2021 Models | StanbyME (Original) | Expected Q1 2026

I don't know if I'm still 100% for removal, because in some tests I saw regression which defeats the purpose of an improvement. Yeah, the double conversion will reduce HW resource usage, but that is negligible compared to a regression or some edge case that I don't see now... I can keep the PR open, maybe it can be revisited and verified by other people interested in testing DoVi playback with both MKV, MP4 source containers

DrHurt:
Yeah, but under what scenario will we be needing mp4 to HLS in Jellyfin (without transcoding that is, which is a whole different pathway).

Also, doesn't that mp4 bug you're talking about imply a bad file? A "full" mp4 file shouldn't really have an empty hvcC as that would cause most players to fail at direct play. 

mp4 fragments are a different story as only the init segment will have the data but subsequent segments won't (necessarily). 

I'm still all for removing this. The less we manipulate files during remuxing, the better.

NoFear0411:
Here are some very extended test results and root cause analysis as to when mp4toannexb is actually needed or not. 
_I had to spend premium API credits to fully trace the issue and verify so apologies for the long text hope it helps._ 

The "bad" MP4 source file was actually a "dvhe" tagged file, I dismissed it as malformed without full verification. @gnattu you were right, this is ffmpeg upstream issue that somehow JF addition of "mp4toannexb" masked the real issue. So up to you and @DrHurt if you want to proceed with merge of this PR.

tl;dr:

The bug is at movenc.c:6589 — the par->extradata_size > 0 check accepts semantically empty extradata (23-byte hvcC header with numOfArrays=0), blocking the first-packet fallback at line 6608 that would correctly extract parameter sets from the bitstream. The BSF accidentally works around this by zeroing the extradata size, allowing the fallback to trigger.

The BSF is a no-op for properly-muxed sources (MKV, dvh1 MP4), but acts as a critical safety net for dvhe-tagged MP4s with sparse hvcC. Removing it would regress playback for these files until upstream ffmpeg is patched.

Long version:

Following up on my earlier tests, I performed additional tests across different source formats, codec tags, and DoVi profiles to determine why I was getting weird results and when exactly the BSF matters. I also traced the issue through the FFmpeg source code to identify the root cause.

Test matrix (jellyfin-ffmpeg 7.1.3, fMP4 HLS output with -tag:v:0 dvh1 -strict -2):

  | Source                       | Container | Codec tag        | DoVi profile             | Source hvcC                     | Without BSF                               | With BSF       | BSF needed? |
  |------------------------------|-----------|------------------|--------------------------|---------------------------------|-------------------------------------------|----------------|-------------|
  | Zack Snyder's Justice League | MP4       | dvhe             | P5 (dvhe.05.06)          | 31 bytes (empty, numOfArrays=0) | 31 bytes, FAIL — unspecified pixel format | 2554 bytes, OK | Yes         |
  | X (2022)                     | MP4       | dvhe             | P5 (dvhe.05.06)          | 31 bytes (empty, numOfArrays=0) | 31 bytes, FAIL — unspecified pixel format | 233 bytes, OK  | Yes         |
  | Fallout S02E04               | MP4       | dvh1             | P8.1+HDR10+ (dvhe.08.06) | 124 bytes (complete)            | 124 bytes, OK                             | 124 bytes, OK  | No          |
  | The Housemaid                | MKV       | V_MPEGH/ISO/HEVC | P5 (dvhe.05.06)          | 159 bytes (complete)            | 159 bytes, OK                             | 159 bytes, OK  | No          |
  | Severance S02E10             | MKV       | V_MPEGH/ISO/HEVC | P8.1+HDR10+ (dvhe.08.06) | complete                        | OK                                        | OK             | No          |

The BSF is only necessary for dvhe-tagged MP4 sources. Both failing files were muxed by Dolby's own DOVI Coding encoder.

Why dvhe specifically?

FFmpeg's codec tag registration (isom_tags.c:124-126) documents the distinction:

-   hev1 → "parameter sets may be in ES"     (in-band allowed, hvcC may be sparse)
-   hvc1 → "parameter sets shall not be in ES" (must be in hvcC)
-   dvhe → "HEVC-based Dolby Vision derived from hev1"

dvhe inherits from hev1, so a sparse hvcC with numOfArrays=0 is spec-valid — parameter sets are stored in-band only. dvh1 inherits from hvc1, so the hvcC must contain VPS/SPS/PPS. MKV always stores complete codec-private data regardless.

Root cause trace through FFmpeg source code:

Path A — dvhe tagged MP4 Without BSF (broken):

  1. Demuxer (mov.c:2418): ff_get_extradata() reads the 23-byte hvcC header (configurationVersion + profile/tier/level fields + numOfArrays=0) into par->extradata. extradata_size = 23.
  2. Muxer extradata copy (movenc.c:6589): Checks par->extradata_size > 0 — this is 23 > 0 = true, so it copies the 23-byte header-only hvcC into track->vos_data. Now vos_len = 23.
  3. First-packet fallback (movenc.c:6608): Checks !trk->vos_len — this is !(23) = false. The fallback is blocked. This fallback would have extracted VPS/SPS/PPS from the first I-frame packet, but because vos_len is non-zero (the useless 23-byte header was accepted), it never triggers.
  4. hvcC writing (movenc.c:1559): track->tag = dvh1 (forced by -tag:v:0) which is not hvc1, so ps_array_completeness = 0 → FLAG_ARRAY_COMPLETENESS not set.
  5. hvcC re-serialization (hevc.c:1177): *data == 1 (configurationVersion) → enters hvcC re-parse path. Reads header fields, sets FLAG_IS_NALFF. At line 1216: num_arrays = 0 → the NALU loop doesn't execute. No VPS/SPS/PPS are added.
  6. Validation skipped (hevc.c:961-967): The "at least one VPS/SPS/PPS" check only runs when FLAG_ARRAY_COMPLETENESS is set. Since dvh1 ≠ hvc1, the flag is not set → empty hvcC passes silently.
  7. Output (hevc.c:1043): Writes numOfArrays = 0 → 23 bytes hvcC content + 8 bytes box header = 31 bytes. No SPS → ffprobe cannot determine pixel format → unspecified pixel format.

  Path B — dvhe tagged MP4 With BSF (works):

  1. BSF init (hevc_mp4toannexb.c:52-54): Parses the 23-byte hvcC. Reads num_arrays = 0 → conversion loop doesn't execute → new_extradata_size = 0. Logs "No parameter sets in the extradata". Sets par_out->extradata = NULL, par_out->extradata_size = 0.
  2. BSF filter (hevc_mp4toannexb.c:135): extradata_parsed = 1 (set during init) → processes packets. Converts length-prefixed NALUs to Annex B (start-code delimited). Prepends zero bytes of extradata to IRAP frames (extradata is empty). First I-frame packet now contains start_code + VPS + start_code + SPS + start_code + PPS + start_code + IDR slices in Annex B format.
  3. Muxer extradata copy (movenc.c:6589): par->extradata_size = 0 → 0 > 0 = false → does not copy. vos_len = 0.
  4. First-packet fallback (movenc.c:6608): !trk->vos_len → true → TRIGGERS. Copies the first I-frame packet data (Annex B) as vos_data.
  5. hvcC writing (hevc.c:1240): *data ≠ 1 (starts with Annex B start code 0x00000001) → enters Annex B parsing path. ff_nal_parse_units_buf parses NALUs. hvcc_parse_nal_unit dispatches VPS/SPS/PPS to hvcc_parse_vps/sps/pps which extract profile, level, chroma format, bit depth. All parameter sets are added to the hvcc arrays.
  6. Output (hevc.c:1043): numOfArrays > 0 → writes complete hvcC with VPS/SPS/PPS = 233–2554 bytes (size varies by source due to SEI NALUs in the first I-frame).

gnattu:
I think we should close this as the original intent is no longer valid after the extensive testing.

We already know that 
- ffmpeg has some problem handling dvhe
- our bsf chain workarounds that problem

Then we should not introduce extra complexity here.

If we want to remove this bsf, we should do it on the ffmpeg side to fix the dvhe handling instead of hacking the bsf chain. Removing dovi metadata is the last-resort resolution to fix the issue that certain video does not even play on some devices, not for sanity use cases like this.

Issue #15814, opened 2025-12-17
Unsupported HDR content results in HLS manifest with multiple video variants (PQ/SDR)


clams4shoes:
### Description of the bug

When Jellyfin encounters an HDR format reported as unsupported by the client (e.g., Dolby Vision Profile 7 or HDR10+) but the client reports support for HDR10, it serves an HLS manifest (`master.m3u8`) containing both:

- an PQ (HDR10) remux variant
- an SDR transcoded variant

This can cause issues for certain clients (e.g., Android TV; see: https://github.com/jellyfin/jellyfin-androidtv/issues/5245), where the SDR variant is selected due to lower codec requirements -- despite the presence of the `VIDEO-RANGE` attribute. In some cases, the client later attempts to switch to the HDR variant, which can result in playback freezing.

In the linked issue, I was informed that HLS manifests served by Jellyfin -- at least during remux/transcode -- should not contain multiple video variants, and that it is a server issue.

#### Example HLS Manifest
```
#EXTM3U
#EXT-X-STREAM-INF:BANDWIDTH=60367918,AVERAGE-BANDWIDTH=60367918,VIDEO-RANGE=PQ,CODECS="hvc1.2.4.L153.B0,mp4a.40.2",RESOLUTION=3840x2160,FRAME-RATE=23.976
main.m3u8?...
#EXT-X-STREAM-INF:BANDWIDTH=60367918,AVERAGE-BANDWIDTH=60367918,VIDEO-RANGE=SDR,CODECS="hvc1.2.4.L150.B0,mp4a.40.2",RESOLUTION=3840x2160,FRAME-RATE=23.976
main.m3u8?...
#EXT-X-IMAGE-STREAM-INF:BANDWIDTH=8241,RESOLUTION=320x180,CODECS="jpeg",URI="Trickplay/320/tiles.m3u8?MediaSourceId=***&ApiKey=***"
```

### Reproduction steps

1. Open the Android TV client or the Web client.
2. Play a video with an unsupported HDR format (e.g., Dolby Vision Profile 7 or HDR10+).
3. Intercept/Inspect the HLS master playlist (`master.m3u8`).

### What is the current _bug_ behavior?

The HLS manifest contains both PQ (HDR10) and SDR video variants.

### What is the expected _correct_ behavior?

The HLS manifest contains only a single video variant/track.

### Jellyfin Server version

10.11.5

### Specify the build version

10.11.5

### Environment

```markdown
[...]

Issue #15675, opened 2025-11-30
DV playback broken on latest build on LG C4 (works on C1)


jacekowski:
### Description of the bug

After upgrading to latest jellyfin/jellyfin:unstable i can't play DV videos anymore on LG C4 (everything works fine on C1). Downgrade to 10.11.2 (10.11.3 is broken) fixes the issue.

Latest version uses different ffmpeg command.

### Reproduction steps

Play DV video on LG C4

### What is the current _bug_ behavior?

Playback hangs immediately.

### What is the expected _correct_ behavior?

Playback starts.

### Jellyfin Server version

Unstable

### Specify the build version

2025112405

### Environment

```markdown
- OS: Debian 12
- Linux Kernel: 6.12.57+deb13-amd64
- Virtualization: Docker
- Clients: LG C4 and LG C1
- Browser: N/A
- FFmpeg Version: 7.1.2-Jellyfin
- Playback Method: Remux
- Hardware Acceleration: none
- GPU Model: none
- Plugins: N/A
- Reverse Proxy: Apache
- Base URL: none
- Networking: bridge
- Jellyfin Data Storage: CEPH
- Media Storage: NFS
- External Integrations: N/A
```

### Jellyfin logs

```shell
This works:

[2025-11-30 20:29:02.448 +00:00] [INF] [7]  MediaBrowser.MediaEncoding.Transcoding.TranscodeManager: "/usr/lib/jellyfin-ffmpeg/ffmpeg" "-analyzeduration 200M -probesize 1G -fflags +genpts  -i file:\"DV_movie.mkv\" -map_metadata -1 -map_chapters -1 -threads 0 -map 0:0 -map 0:1 -map -0:s -codec:v:0 copy -tag:v:0 dvh1 -strict -2 -bsf:v hevc_mp4toannexb -start_at_zero -codec:a:0 copy -copyts -avoid_negative_ts disabled -max_muxing_queue_size 2048 -f hls -max_delay 5000000 -hls_time 1 -hls_segment_type fmp4 -hls_fmp4_init_filename \"f2a118b37f55da8215e05847f000f1cc-1.mp4\" -start_number 0 -hls_segment_filename \"/cache/transcodes/f2a118b37f55da8215e05847f000f1cc%d.mp4\" -hls_playlist_type vod -hls_list_size 0 -hls_segment_options movflags=+frag_discont -y \"/cache/transcodes/f2a118b37f55da8215e05847f000f1cc.m3u8\""

[...]

PR #16362, opened 2026-03-06
Add spec-compliant dvh1 HLS variant for Dolby Vision Profile 5

### LINKED ISSUE #16179: Dolby Vision manifest should use spec-compliant HLS manifest

NoFear0411:
## The Problem

Dolby Vision Profile 5 has no backward-compatible base layer — there is nothing to declare in `SUPPLEMENTAL-CODECS`. Per the [Apple HLS Authoring Specification](https://developer.apple.com/documentation/http-live-streaming/hls-authoring-specification-for-apple-devices-appendixes), P5 requires `CODECS="dvh1.05.XX"` as the primary codec string.

Currently, `GetH265String` always returns `hvc1.*` for HEVC streams regardless of DoVi profile. This creates a mismatch: `DynamicHlsController` already passes `-tag:v:0 dvh1 -strict -2` to FFmpeg for P5 copy-codec streams ([line 1817](https://github.com/jellyfin/jellyfin/blob/master/Jellyfin.Api/Controllers/DynamicHlsController.cs#L1817)), writing a `dvh1` FourCC and `dvvC` configuration box into the fMP4 init segment. But the master playlist declares `CODECS="hvc1.2.4.L150.B0"` — the bitstream says DV, the manifest says plain HEVC.

Spec-compliant clients (Apple TV, webOS 24+) read the manifest first and set up an HDR10 pipeline instead of a Dolby Vision one. The TV switches to HDR mode instead of DV mode, resulting in incorrect tone mapping and dimmer picture. Manually changing `CODECS` to `dvh1` in the manifest resolves it (confirmed by multiple users in #16179).

## The Fix

Add a `dvh1` variant **before** the existing `hvc1` variant in the master playlist for P5 copy-codec streams. Both variants point to the same stream URL — no extra transcoding or storage.

**Before (3 variants):**
```
#EXT-X-STREAM-INF:...,VIDEO-RANGE=PQ,CODECS="hvc1.2.4.L150.B0,ec-3"    ← hvc1 hack
main.m3u8?...
#EXT-X-STREAM-INF:...,VIDEO-RANGE=SDR,CODECS="hvc1.1.4.L150.B0,ec-3"   ← SDR HEVC
main.m3u8?...&VideoCodec=hevc&...
#EXT-X-STREAM-INF:...,VIDEO-RANGE=SDR,CODECS="avc1.640033,ec-3"         ← SDR H.264
main.m3u8?...&VideoCodec=h264&...
```

**After (4 variants):**
```
#EXT-X-STREAM-INF:...,VIDEO-RANGE=PQ,CODECS="dvh1.05.06,ec-3"          ← spec-compliant DV (NEW)
main.m3u8?...
#EXT-X-STREAM-INF:...,VIDEO-RANGE=PQ,CODECS="hvc1.2.4.L150.B0,ec-3"    ← hvc1 hack (unchanged)
main.m3u8?...
#EXT-X-STREAM-INF:...,VIDEO-RANGE=SDR,CODECS="hvc1.1.4.L150.B0,ec-3"   ← SDR HEVC (unchanged)
main.m3u8?...&VideoCodec=hevc&...
#EXT-X-STREAM-INF:...,VIDEO-RANGE=SDR,CODECS="avc1.640033,ec-3"         ← SDR H.264 (unchanged)
main.m3u8?...&VideoCodec=h264&...
```

## Why This Ordering

When two variants have identical `BANDWIDTH`, HLS clients SHOULD prefer the first compatible match. Listing `dvh1` first means:

- **Spec-compliant clients** (Apple TV, webOS 24+): select `dvh1` → DV decoder path activated
- **Legacy clients** that reject `dvh1` in CODECS (older LG TVs, some Android TV): skip `dvh1` → fall through to `hvc1` → detect DV from init segment's `dvvC` box → existing behavior preserved
- **Non-HDR devices**: skip both PQ variants → select SDR HEVC or H.264 transcode

## Why Only P5

Profile 8.1/8.4 already have correct signaling via `SUPPLEMENTAL-CODECS` (`dvh1.08.XX/db1p` for P8.1, `dvh1.08.XX/db4h` for P8.4). These profiles have backward-compatible base layers (HDR10/HLG) that non-DV devices can decode — `SUPPLEMENTAL-CODECS` is the correct mechanism. P5 is the only profile where the primary `CODECS` field must use `dvh1` because there is no compatible base layer to supplement.

## Guard Conditions

The `dvh1` variant is only emitted when all of:
- Stream is copy codec (remux, not transcode)
- `VideoRangeType` is `DOVI` (P5 specifically, not P8.1/P8.4)
- DoVi metadata is present and not being stripped
- `DvProfile` and `DvLevel` are available for building the codec string

## Changes

| File | Change |
|------|--------|
| `HlsCodecStringHelpers.cs` | Add `GetDoviString(int dvProfile, int dvLevel)` returning `dvh1.{profile:D2}.{level:D2}` |
| `DynamicHlsHelper.cs` | Add P5 detection in `GetMasterPlaylistInternal` before the main `hvc1` variant |
| `DynamicHlsHelper.cs` | Add `AppendDoviPlaylist` helper following the same pattern as `AppendPlaylist` |

## References

- Apple HLS Authoring Spec: P5 example uses `CODECS="dvh1.05.06"` with no `SUPPLEMENTAL-CODECS`
- Apple HLS Authoring Spec Rule 1.10: SHOULD use `dvh1` (out-of-band) over `dvhe` (in-band)
- `DynamicHlsController.cs:1817`: already writes `-tag:v:0 dvh1 -strict -2` for P5 fMP4 streams
- `AppendPlaylistSupplementalCodecsField`: correctly returns empty for `VideoRangeType.DOVI` (no compatible BL to supplement)

Fixes #16179

### DIFF
--- a/Jellyfin.Api/Helpers/DynamicHlsHelper.cs
+++ b/Jellyfin.Api/Helpers/DynamicHlsHelper.cs
@@ -428,7 +428,7 @@ public class DynamicHlsHelper
                 .Append('"');
         }
 
-        playlistBuilder.Append(Environment.NewLine);
+        playlistBuilder.AppendLine();
         playlistBuilder.AppendLine(url);
         builder.Append(playlistBuilder);
     }


NoFear0411:
@gnattu hope this satisfies the DoVi P5 compliant signaling while perserving the "hacks" for other/older devices. 

https://github.com/jellyfin/jellyfin/issues/16179#issuecomment-3875064494

NoFear0411:
Spec compliance verification notes:
                                                                                                                                                                                                                                                              
I traced the full pipeline to verify the manifest output matches what FFmpeg actually writes into fMP4 init segments:                                                                                                                                                                                                     

  1. Init segment alignment: DynamicHlsController.cs:1817 writes `-tag:v:0 dvh1 -strict -2` for P5, producing a dvh1 FourCC + dvvC box in the init segment. `AppendDoviPlaylist` now emits `CODECS="dvh1.05.06,..."` - these match. Previously the manifest declared hvc1 while the init segment contained dvh1, violating the Apple HLS Authoring Spec.
  2. Dual-variant safety: Both the dvh1 variant (new) and hvc1 variant (existing) point to the same stream URL. Per HLS spec, when variants share identical BANDWIDTH, players pick the first compatible match. DoVi-capable clients match dvh1 first; legacy clients skip it and fall to hvc1. No extra transcoding or storage cost.
  3. P8.1/P8.4 unaffected: Guard condition `VideoRangeType == DOVI` only matches P5 and the theoretical P10/bl_compat_0. P8 uses `DOVIWithHDR10/DOVIWithHLG` and continues using `SUPPLEMENTAL-CODECS` as before.
  4. DoVi metadata preserved: `ShouldRemoveDynamicHdrMetadata` returns None for `VideoRangeType.DOVI` - RPU NALs and dvvC/dvcC boxes are never stripped for P5.

Companion PR: https://github.com/jellyfin/jellyfin-web/pull/7659 adds Safari DoVi detection so that Apple devices actually declare DOVI support and trigger this code path. Without it, Safari sends no DOVI range type --> server rejects copy --> P5 gets transcoded to SDR.

satmandu:
Could this also get backported to 10.11.x?

NoFear0411:
> Could this also get backported to 10.11.x?

Sure it can, after the maintainers review it and accept it. Hope that happens before the next release.

holow29:
@gnattu What would make you confident? It is extremely frustrating to want to help develop a spec-compliant server and constantly be flummoxed by worries about clients that could change to be spec-compliant but somehow make that the server's problem. I believe this solves an issue with certain Apple devices and players that are spec-compliant, but you are preventing this fix for an unknown potential incompatibility with non-compliant players. At some point, it must be the client's responsibility to be compliant. There are enough server-side hacks as is that only serve at this point to allow clients to remain non-compliant.

I think allowing this in the next release (and not backporting) is time enough for clients to ensure compatibility. After all, you'll allow breaking API changes in future releases, so there is already an implied warning.

gnattu:
> It is extremely frustrating to want to help develop a spec-compliant server and constantly be flummoxed by worries about clients that could change to be spec-compliant but somehow make that the server's problem.

So do I. Unfortunately "it must be the client's responsibility to be compliant" is harder than it looks like because we mostly uses the system player and it is not stable on a lot of devices. It is not unknown devices. We got report from android tv that the HLS player ignores range and picks up fallback when it should not to:
https://github.com/jellyfin/jellyfin/issues/15814

We also know that on LG webOS, dvh1 in manifest is not supported until very recently, and on webOS it has very inconsistent playback support status, depending on the device model and webOS version.

> There are enough server-side hacks as is that only serve at this point to allow clients to remain non-compliant.

Therefore this is inevitable. We cannot force users to upgrade the OS/system to a bug-fixed version, and some of them may not even have the option to. 

So back to your question:
> What would make you confident?

When we change the playback code, there will not be a lot of users complaining playbacks break for them. Unfortunately it happened multiple times, if not every time, we make the server side more spec compliant.

A recent example:
https://github.com/jellyfin/jellyfin/issues/15675

I used to be more confident on making such changes but with recent reports I have to be more careful on "making things technically correct". 

Of cause we cannot predict all use cases, but some regression tests on devices that this change **should not affect** will be very helpful. My biggest concern is that it breaks playback on other devices after all.

> I think allowing this in the next release (and not backporting) is time enough for clients to ensure compatibility. After all, you'll allow breaking API changes in future releases, so there is already an implied warning.

I hope this would work too, but our recent changes to dynamic HDR handling is also merged during the unstable cycle and there are no reports about the compatibility issues during that cycle and its compatibility issue is only being reported after the "stable" release.

I agree that we need a paradigm shift in how we build the playback pipeline. We’ve been choosing the safest path for years by hacking on the server side to fix broken clients, but I believe we should introduce a device-side quirk system. This system would allow devices to report their non-compliant behavior to the server, enabling the server to take necessary mitigating actions. Even with that change, it’s challenging to discover the device-specific quirks until a user notices that playback is no longer working. In such cases, the user may need to manually enable certain quirk options, which, in my opinion, is worse than simply having the video play in an unideal state. Enabling an option that the user is unaware of its purpose is far from user-friendly.

I apologize for my overly timid actions, but I’ve seen too many cases where breaking use case A to fix a bug in use case B. Additionally, fixing bugs for a device that I don’t own or have access to test on is not enjoyable, and playback regressions are usually this kind of bugs.

NoFear0411:
@gnattu I understand the caution and appreciate the detailed explanation. But I think we need to step back and look at the bigger picture because I think this isn't just about DoVi playback.

  Jellyfin's playback pipeline in 2026 should be built on a spec-compliant baseline, not a broken one.

  The current approach inverts the correct architecture: the server produces spec-violating output as the default, and you are afraid to fix it because some non-compliant clients accidentally work with the broken output. This pattern repeats across the entire pipeline, not just DoVi signaling, but CODECS strings, VIDEO-RANGE, BANDWIDTH values, and manifest structure. Every time the server outputs incorrect metadata to accommodate a broken client, every modern spec-compliant device pays the price. 

I believe the right approach should be:

  1. Baseline: spec-compliant pipeline first. The server should produce correct output per the specs it implements. A server releasing in 2026 should target 2026 specs and device capabilities, not be designed around client bugs.
  2. Quirks: device-specific workarounds layered on top. When a specific client is known to be broken (like the Android TV ExoPlayer issue in #15814), targeted overrides handle it. This is the quirk system you described, but it must work FROM a correct baseline downward, not from a broken baseline that punishes every compliant device.

I think that designing the pipeline around legacy client bugs creates a maintenance trap: every new spec-compliant device needs investigation into why it doesn't work, when the answer is almost always "because the server sends wrong metadata." Meanwhile the non-compliant clients that current pipeline is protecting often get incorrect playback anyway, they just fail silently instead of visibly.

  The broader capability detection problem is real, web and TV clients still rely on outdated APIs like "canPlayType()" that return incorrect results for modern codecs and HDR formats. That needs updating too, but outdated client detection shouldn't block fixing the server's output.

The jellyfin-server must have correct baseline and as you said a quirk system to handle known broken non-spec compliant clients, than maintain a broken baseline that holds back every modern device. Even other users are saying the same things over and over again https://github.com/jellyfin/jellyfin/issues/16092#issuecomment-4046197674 and it feels like jf-server media playback pipeline is held hostage by broken devices/clients/players.

gnattu:
> Baseline: spec-compliant pipeline first. The server should produce correct output per the specs it implements. A server releasing in 2026 should target 2026 specs and device capabilities, not be designed around client bugs.

This is exactly what I meant how the quirk system would work. Standard by default and client specific workarounds on top of that. But that is out of scope of this PR.

For this PR, like what I said above, if there are some testing results on devices that should not be affected would be very helpful. We do not need to predict all use cases, but at least do some testing before we make a change that would break some users.

NoFear0411:
The PR scope is targeting DoVi/HLS spec-compliant playback on DoVi/HLS spec-compliant devices. 

All the extensive testing that I have done prior to opening this PR, are on Apple(ipad,iphone,tvOS) and LG(C4,CX,C1) devices. I can confirm that the changes this PR introduces, will help those devices tremendously and enable more smooth DoVi playback. LG is rolling out the latest WebOS 26 firmware to all TVs (going back to 2021 models) and I can 100% confirm the firmware does indeed follow DoVi/HLS spec compliance. So the users on those older WebOS TVs (up to 2021) will not be affected if they just apply the OTA upgrades. 

As for the other DoVi capable devices that are not 100% spec-compliant, I cannot give you concrete evidence that this will help them as well. The worst case scenario that I see here, is that they get some broken DoVi playback on P5 media but some playback on P8/7 media. I think this PR will help identify those non-compliant devices so we can learn and push those vendors to make the changes.

Either way even more vendors are starting to make changes in order to become DoVi/HLS spec-compliant, because commercial streaming platforms demand them, so that they can serve higher quality content, including DoVi. 

It is your decision if you want to close, merge or keep the PR open till more testing data is provided by the users.

Jellyfin.Api/Helpers/HlsCodecStringHelpers.cs
```diff
@@ -346,4 +346,23 @@ public static string GetAv1String(string? profile, int level, bool tierFlag, int
...
+    /// <summary>
+    /// Gets a Dolby Vision codec string.
+    /// </summary>
+    /// <param name="dvProfile">Dolby Vision profile number.</param>
+    /// <param name="dvLevel">Dolby Vision level number.</param>
+    /// <returns>Dolby Vision codec string.</returns>
+    public static string GetDoviString(int dvProfile, int dvLevel)
+    {
+        // FORMAT: dvh1.[profile].[level] (out-of-band parameter sets, recommended by Apple HLS spec Rule 1.10)
+        StringBuilder result = new StringBuilder("dvh1", 12);
```
gnattu: I *think* we should use correct fourcc for other profiles, like Profile 10 with av1 codec, which should use `dav1`. Profile 10.0 also has no compatibility fallback layer

Jellyfin.Api/Helpers/DynamicHlsHelper.cs
```diff
@@ -209,6 +209,22 @@ private async Task<ActionResult> GetMasterPlaylistInternal(
...
 
+        // For DoVi Profile 5, add a spec-compliant dvh1 variant before the hvc1 hack variant.
+        // P5 has no compatible base layer, so SUPPLEMENTAL-CODECS cannot be used.
+        // The dvh1 variant is listed first so spec-compliant clients (Apple TV, webOS 24+)
+        // select it over the hvc1 fallback when both have identical BANDWIDTH.
+        if (state.VideoStream is not null
+            && state.VideoRequest is not null
+            && EncodingHelper.IsCopyCodec(state.OutputVideoCodec)
+            && state.VideoStream.VideoRangeType == VideoRangeType.DOVI
+            && EncodingHelper.IsDovi(state.VideoStream)
```
gnattu: `IsDovi` already checked `state.VideoStream.VideoRangeType == VideoRangeType.DOVI`

Jellyfin.Api/Helpers/DynamicHlsHelper.cs
```diff
@@ -209,6 +209,22 @@ private async Task<ActionResult> GetMasterPlaylistInternal(
...
+            && !_encodingHelper.IsDoviRemoved(state)
```
gnattu: This check is not very useful because Profile 5 has no compatibility layer, which means there is nothing to remove.

Instead, we could check if the client really declared `VideoRangeType.DOVI` as supported to not adding this manifest to clients that do not label dolby vision as supported at all. We can use `state.GetRequestedRangeTypes(state.VideoStream.Codec)` to get the requested video ranges.

Jellyfin.Api/Helpers/DynamicHlsHelper.cs
gnattu: I *think* we should really only testing `state.VideoStream.VideoRangeType == VideoRangeType.DOVI` here.

We should not add this manifest to other kind of Dolby Vision videos.

Jellyfin.Api/Helpers/HlsCodecStringHelpers.cs
NoFear0411: Done. `GetDoviString` now takes an isAv1 parameter and returns dav1.{profile}.{level} for AV1 DoVi, dvh1.{profile}.{level} for HEVC DoVi. This matches the existing pattern in `AppendDvString` (line 488) where dav1 vs dvh1 is already codec-dependent.

Jellyfin.Api/Helpers/DynamicHlsHelper.cs
NoFear0411: Removed. The guard now uses only `VideoRangeType == VideoRangeType.DOVI` which is the precise check for profiles without a compatible base layer (P5 HEVC, P10 bl_compat_id=0 AV1).

Jellyfin.Api/Helpers/DynamicHlsHelper.cs
NoFear0411: Confirmed. The guard is now strictly `VideoRangeType.DOVI` only. From MediaStream.cs, this maps to exactly two cases: P5 (line 791) and P10 bl_compat_id=0 (line 803). P8, P7, DOVIWithHDR10, DOVIWithHLG, and all other DoVi types are excluded, they continue using SUPPLEMENTAL-CODECS as before.

NoFear0411: Companion PR jellyfin/jellyfin-web#7659 adds the client-side Safari DoVi detection so these devices actually declare DOVI in their requested range types and trigger this code path. Without it, the client capability check here would filter them out. Would appreciate a review on that one as well.

Jellyfin.Api/Helpers/DynamicHlsHelper.cs
NoFear0411: Replaced `IsDoviRemoved` with: `state.GetRequestedRangeTypes(state.VideoStream.Codec).Contains("DOVI")`
This ensures the dvh1/dav1 variant is only emitted for clients that explicitly declared DOVI support. Same pattern used in EncodingHelper.cs:1392 and DynamicHlsController.cs:1807.

Jellyfin.Api/Helpers/HlsCodecStringHelpers.cs
```diff
@@ -346,4 +346,25 @@ public static string GetAv1String(string? profile, int level, bool tierFlag, int
...
     }
+
+    /// <summary>
+    /// Gets a Dolby Vision codec string for profiles without a compatible base layer.
+    /// </summary>
+    /// <param name="dvProfile">Dolby Vision profile number.</param>
+    /// <param name="dvLevel">Dolby Vision level number.</param>
+    /// <param name="isAv1">Whether the video codec is AV1 (uses dav1 FourCC instead of dvh1).</param>
+    /// <returns>Dolby Vision codec string.</returns>
+    public static string GetDoviString(int dvProfile, int dvLevel, bool isAv1)
```
nyanmisaka: I recommend passing in the codec name `string codec` instead of a boolean. This is in case the API changes occur when DoVi is extended to H.266 or AV2 in the future.

NoFear0411: Done. Changed bool isAv1 to string codec, the FourCC mapping is now codec-driven, same pattern as AppendDvString (line 507). Easy to extend when H.266/VVC or AV2 DoVi FourCCs are defined.

Jellyfin.Api/Helpers/DynamicHlsHelper.cs
```diff
@@ -917,4 +936,63 @@ private string ReplacePlaylistCodecsField(StringBuilder playlist, StringBuilder
...
+    /// Appends a Dolby Vision variant with dvh1/dav1 CODECS for profiles without a compatible
+    /// base layer (P5 HEVC, P10/bl0 AV1). This enables spec-compliant HLS clients to detect
+    /// DoVi from the manifest rather than relying on init segment inspection.
+    /// </summary>
+    /// <param name="builder">StringBuilder for the master playlist.</param>
+    /// <param name="state">StreamState of the current stream.</param>
+    /// <param name="url">Playlist URL for this variant.</param>
+    /// <param name="bitrate">Bitrate for the BANDWIDTH field.</param>
+    /// <param name="subtitleGroup">Subtitle group identifier, or null.</param>
+    private void AppendDoviPlaylist(StringBuilder builder, StreamState state, string url, int bitrate, string? subtitleGroup)
```
nyanmisaka: Could you move `AppendDoviPlaylist()` next to `AppendPlaylist()`? This would make them more relevant and easier to compare.

NoFear0411: Moved.

Jellyfin.Api/Helpers/HlsCodecStringHelpers.cs
```diff
@@ -346,4 +346,25 @@ public static string GetAv1String(string? profile, int level, bool tierFlag, int
...
+        // HEVC DoVi uses dvh1, AV1 DoVi uses dav1 (out-of-band parameter sets, recommended by Apple HLS spec Rule 1.10)
+        var fourCc = string.Equals(codec, "av1", StringComparison.OrdinalIgnoreCase) ? "dav1" : "dvh1";
+        StringBuilder result = new StringBuilder(fourCc, 12);
+
+        result.Append('.')
+            .AppendFormat(CultureInfo.InvariantCulture, "{0:D2}", dvProfile)
+            .Append('.')
+            .AppendFormat(CultureInfo.InvariantCulture, "{0:D2}", dvLevel);
+
+        return result.ToString();
```
Bond-009, line 368: This could just be a string.format. (unless this will get more complex in the near future with conditionals?)

Jellyfin.Api/Helpers/DynamicHlsHelper.cs
```diff
@@ -355,6 +374,65 @@ private StringBuilder AppendPlaylist(StringBuilder builder, StreamState state, s
...
+        AppendPlaylistFramerateField(playlistBuilder, state);
+
+        if (!string.IsNullOrWhiteSpace(subtitleGroup))
+        {
+            playlistBuilder.Append(",SUBTITLES=\"")
+                .Append(subtitleGroup)
+                .Append('"');
+        }
+
+        playlistBuilder.Append(Environment.NewLine);
```
Bond-009:
```suggestion
        playlistBuilder.AppendLine();
```

Jellyfin.Api/Helpers/HlsCodecStringHelpers.cs
NoFear0411: Hmm `string.Format` is cleaner here, BUT I think keeping `StringBuilder` is better for now. The only reason is because of the DoVi codec string format, that might need profile-specific adjustments. This is only if you decide and build out the spec-compliant signaling for additional profiles/levels.

PR #16368, opened 2026-03-08
Fix remote image language priority to prefer English over no-language

### LINKED ISSUE #13310: Sort and use the best rated pictures for primary pictures

### Description of the bug

When adding new series/movies to the libraries, the automated choice for primary pictures is always the worst.
It uses "Unrated" pictures that aren't as good as others, that also have a ranking.
The same can be seen, when you "edit the pictures" via the three-dot menu.

Two examples:
Image

Image

You can see, that the first pics in the row have a rating of "Unrated" or at least a lower rating and the better ones (with rating 8.0) are later on in the list - so they aren't used in the first place.

It seems to work just fine on logos, thumbnais and backgrounds.

It also should consider the language of the user (in my case 'de').
So the order should be:
- highest rated to lowest rated in the current language
- highest rated to lowest rated in the default or unknown language (perhaps always consider 'en', because most of the stuff is 'en')
- Unrated  in the current language
- Unrated  in the default or unknown language

### Reproduction steps

1. go to your library (movies or series)
2. use the three-dot menu on an item
3. "edit pictures"
4. click on the search icon below "primary" picture

==> no particularly order of the pictures, so the first one, that is taken, can be the worst...

The best rated primary picture (in the current language) should be first in list and automatically used when a new series/movie is found in the library.

### Jellyfin Server version

10.10.0+

### Specify commit id

_No response_

_No response_

### Specify version number

10.10.3.0

### Specify the build version

 10.10.3

### Environment

```markdown
- OS: Debian GNU/Linux 12 (bookworm)
- Linux Kernel: Linux rpi4-1 6.1.21-v8+ #1642 SMP PREEMPT Mon Apr  3 17:24:16 BST 2023 aarch64 GNU/Linux
- Virtualization: none
- Clients: Web, Kodi on nVIDIA Shield
- Browser: Chrome, always updated
- FFmpeg Version: ffmpeg version 7.0.2-Jellyfin Copyright (c) 2000-2024 the FFmpeg developers
- Playback Method: native
```

### Jellyfin logs

```shell
Nothing in the logs for that issue.
I can provide logs, when it makes sense (extra debug logs) - just need to know how to produce them.
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

redinsch:
**Changes**
Swap the priority order in `OrderByLanguageDescending()` so that English images (score 3) are preferred over images with no language (score 2). Previously, languageless images were ranked higher, causing poorly rated images to be selected over well-rated English alternatives for posters and logos.

Backdrop images are unaffected as they have their own dedicated sorting in `ItemImageProvider`.

**Issues**
Fixes #13310

## Test plan
- [x] Added 6 unit tests for `OrderByLanguageDescending` covering all priority combinations
- [x] Tested manually: library scan now selects better-rated English posters/logos over languageless ones

### DIFF
--- a/Jellyfin.Api/Controllers/QuickConnectController.cs
+++ b/Jellyfin.Api/Controllers/QuickConnectController.cs
@@ -52,6 +52,7 @@ public class QuickConnectController : BaseJellyfinApiController
     /// <returns>A <see cref="QuickConnectResult"/> with a secret and code for future use or an error message.</returns>
     [HttpPost("Initiate")]
     [ProducesResponseType(StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
     public async Task<ActionResult<QuickConnectResult>> InitiateQuickConnect()
     {
         try


tests/Jellyfin.Model.Tests/Extensions/EnumerableExtensionsTests.cs
```diff
@@ -0,0 +1,117 @@
+using System.Linq;
+using MediaBrowser.Model.Extensions;
+using MediaBrowser.Model.Providers;
+using Xunit;
+
+namespace Jellyfin.Model.Tests.Extensions
```
Bond-009: use file scoped namespaces

redinsch: changed to file scoped namespace

redinsch:
@Bond-009 @Shadowghost
Just a small heads up regarding my PR – the `OpenAPI - Difference` check failed with an error that seems to be a bug in the check itself (it tried to post a comment that exceeded the size limit). This isn't related to any API changes on my end.

To be safe, I re-merged from master (no conflicts), but the checks now need re-approval to run again. I haven't changed anything else.

No rush at all – just wanted to keep you in the loop! 😊

PR #16370, opened 2026-03-08
Feature: Add clip extraction

olwe1:
> Add clip extraction feature
>
> Adds a new ClipController with four endpoints (POST, GET /Progress via SSE, GET /Download, DELETE) allowing users to extract a time-range clip from any video item. FFmpeg is used for re-encoding with H.264, H.265 or AV1, with AAC audio. Encoding is limited to one concurrent job via a semaphore, progress is streamed to the client, and completed files are cleaned up automatically after 48 hours.

https://features.jellyfin.org/posts/2786/create-clips-from-media
Client PR: https://github.com/jellyfin/jellyfin-web/pull/7664


joshuaboniface:
I don't believe we want this in the core server. We have multiple concerns about the implementation and would require much further discussion about it before accepting this, which we cannot commit to at this time.

We would request that this be implemented as a 3rd party plugin.

PR #16372, opened 2026-03-09
Add playback speed property to session state APIs

### LINKED ISSUE #8971: Add "speed" property to playback state API's

boris324:
Adds a nullable `Speed` float property to `PlaybackProgressInfo`, `PlayerStateInfo`, and the legacy query-parameter progress endpoints so clients can report and query the current playback rate.

The value flows through `SessionManager.UpdateNowPlayingItem` into the session play state and is automatically included in `SessionInfoDto` responses and WebSocket broadcasts.

`PlaybackStartInfo` inherits `PlaybackProgressInfo` so it picks up the property without changes. The JSON-body endpoints (`ReportPlaybackStart`, `ReportPlaybackProgress`) accept it automatically; the legacy query-parameter endpoints get an explicit `speed` parameter.

Closes #8971

### DIFF
--- a/Emby.Server.Implementations/Session/SessionManager.cs
+++ b/Emby.Server.Implementations/Session/SessionManager.cs
@@ -447,6 +447,7 @@ namespace Emby.Server.Implementations.Session
             session.PlayState.CanSeek = info.CanSeek;
             session.PlayState.IsMuted = info.IsMuted;
             session.PlayState.VolumeLevel = info.VolumeLevel;
+            session.PlayState.Speed = info.Speed;
             session.PlayState.AudioStreamIndex = info.AudioStreamIndex;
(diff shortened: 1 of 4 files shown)


Bond-009:
This is already being addressed: https://github.com/jellyfin/jellyfin/pull/9329

sevenrats:
can we ban this joker his AI is obviously just going to keep doing this

Issue #16028, opened 2026-01-15
Live TV bitrate reported as ~382 Mbps with TVHeadend, forcing unnecessary transcoding on Raspberry Pi 5


IndianBartonka:
### Description of the bug

<img width="417" height="317" alt="Image" src="https://github.com/user-attachments/assets/97597151-99ee-4caa-9341-d2ef2ec36997" />

<img width="434" height="295" alt="Image" src="https://github.com/user-attachments/assets/8010968f-f78b-4961-a2a3-8322196dd3a2" />

When using an Xbox TV Tuner with TVHeadend and the Jellyfin TVHeadend plugin, the bitrate for Live TV channels is incorrectly reported as approximately 382 Mbps (specifically 382.7 Mbps). Because this reported bitrate exceeds any client or server limits, the Jellyfin server attempts to transcode or remux the stream.

On a Raspberry Pi 5 (which lacks a hardware H.264 encoder), software transcoding of a perceived 382 Mbps stream causes 100% CPU usage and severe playback stuttering, even when using Jellyfin Media Player with "Unlimited" settings.

### Reproduction steps

Setup TVHeadend with Xbox TV Tuner.

Connect TVHeadend to Jellyfin via the official plugin.

Open any Live TV channel.

Check "Playback Info" – bitrate shows 382.7 Mbps.

Server starts transcoding/remuxing due to "Bitrate exceeds limit".

### What is the current _bug_ behavior?

Raspberry Pi transcodes even though, as far as I know, Jellyfin Media Player shouldn't transcode.

### What is the expected _correct_ behavior?

Normal Bitrate

### Jellyfin Server version

10.11.5

### Specify the build version

10.11.5

### Environment

```markdown
OS: Raspberry Pi OS (Debian 12 Bookworm)
Linux Kernel: 6.12.47+rpt-rpi-2712
Virtualization: None (Bare Metal)
Clients: Jellyfin Media Player (Windows), Android App (Xiaomi Redmi Note 13 Pro), Web Browser (Chrome/Edge)
FFmpeg Version: Jellyfin-FFmpeg 6.x 
Playback Method: Remuxing / Transcoding (forced by incorrect bitrate reporting)
Hardware Acceleration: None (Raspberry Pi 5 lacks a hardware H.264 encoder)
Plugins: TVHeadend, YouTube Metadata, Playback Reporting
Reverse Proxy: None (Direct IP access)
Base URL: None
Networking: Local LAN (Gigabit Ethernet)
[...]

PR #16376, opened 2026-03-09
Fix int32 overflow in QSV rate-control parameter computation

scheilch:
## The Problem

While investigating #16028 I noticed that `GetVideoBitrateParam` computes several QSV rate-control parameters using plain int32 arithmetic:

```csharp
// original code
return FormattableString.Invariant($" -b:v {bitrate} -maxrate {bitrate + 1} -rc_init_occupancy {bitrate * 2} -bufsize {bitrate * 4}");
```

When the source reports no useable bitrate metadata (common with Live TV / MPEG-TS streams from e.g. TVHeadEnd), Jellyfin falls back to a very large estimated value. For a bitrate above roughly 537 million, `bitrate * 4` silently wraps to a negative int32 and FFmpeg receives somthing like `-bufsize -4`.

This is incorrect regardless of whether the encoder backend tolerates it or not.

### Observed command line (before fix)

```
-b:v 1073741823 -maxrate 1073741824 -rc_init_occupancy 2147483646 -bufsize -4
```

### After fix

```
-b:v 50000000 -maxrate 50000001 -rc_init_occupancy 62500000 -bufsize 62500000
```

## What this PR does

1. **Long arithmetic** for `qsvMaxrate`, `qsvInitOcc` and `qsvBufsize`, clamped to `int.MaxValue`, prevents the overflow.
2. **h264_qsv bitrate cap at 50 Mbps** — H.264 Level 4.2 (the most common level for 1080p QSV transcoding) has a CPB limit of 62.5 Mbit, so 50 Mbps keeps derived values safely within spec. HEVC and AV1 encoders have much higher level limits and are not affected.
3. **CPB-aware clamping** — when an H.264 level is known, bufsize, rc_init_occupancy and maxrate are clamped to the level's maximum CPB size (per ITU-T H.264 Table A-1).
4. **General bufsize fix** — the `bufsize = bitrate * 2` computation used by libx264/libx265/AMF/VAAPI etc. now also uses long arithmetic.

## Tests

Added comprehensive unit tests covering:
- Overflow prevention (int.MaxValue/2 inputs)
- Bitrate capping (h264_qsv only, not hevc/av1)
- CPB clamping (Level 4.2, 5.1, 6.0, boundary, no-level)
- Minimum bitrate enforcement (< 1000)
- CPB lookup table correctness (all 19 levels + unknown)

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -1592,9 +1592,9 @@ namespace MediaBrowser.Controller.MediaEncoding
                 // Set (rc_init_occupancy == 2 * bitrate) and (bufsize == 4 * bitrate) to deal with drastic scene changes
                 // Use long arithmetic and clamp to int.MaxValue to prevent int32 overflow
                 // (e.g. bitrate * 4 wraps to a negative value for bitrates above ~537 million)
-                long qsvMaxrate = Math.Min((long)bitrate + 1, int.MaxValue);
-                long qsvInitOcc = Math.Min((long)bitrate * 2, int.MaxValue);
-                long qsvBufsize = Math.Min((long)bitrate * 4, int.MaxValue);
+                int qsvMaxrate = (int)Math.Min((long)bitrate + 1, int.MaxValue);
+                int qsvInitOcc = (int)Math.Min((long)bitrate * 2, int.MaxValue);
+                int qsvBufsize = (int)Math.Min((long)bitrate * 4, int.MaxValue);
 
                 return FormattableString.Invariant($"{mbbrcOpt} -b:v {bitrate} -maxrate {qsvMaxrate} -rc_init_occupancy {qsvInitOcc} -bufsize {qsvBufsize}");
             }
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -1552,14 +1552,15 @@ namespace MediaBrowser.Controller.MediaEncoding
 
             int bitrate = state.OutputVideoBitrate.Value;
 
-            // Bit rate under 1000k is not allowed in h264_qsv
+            // Bit rate under 1000k is not allowed in h264_qsv.
             if (string.Equals(videoCodec, "h264_qsv", StringComparison.OrdinalIgnoreCase))
             {
                 bitrate = Math.Max(bitrate, 1000);
             }
 
-            // Currently use the same buffer size for all encoders
-            int bufsize = bitrate * 2;
+            // Currently use the same buffer size for all non-QSV encoders.
+            // Use long arithmetic to prevent int32 overflow for very high bitrate values.
+            int bufsize = (int)Math.Min((long)bitrate * 2, int.MaxValue);
 
             if (string.Equals(videoCodec, "libsvtav1", StringComparison.OrdinalIgnoreCase))
             {
@@ -1589,7 +1590,13 @@ namespace MediaBrowser.Controller.MediaEncoding
 
                 // Set (maxrate == bitrate + 1) to trigger VBR for better bitrate allocation
                 // Set (rc_init_occupancy == 2 * bitrate) and (bufsize == 4 * bitrate) to deal with drastic scene changes
-                return FormattableString.Invariant($"{mbbrcOpt} -b:v {bitrate} -maxrate {bitrate + 1} -rc_init_occupancy {bitrate * 2} -bufsize {bitrate * 4}");
+                // Use long arithmetic and clamp to int.MaxValue to prevent int32 overflow
+                // (e.g. bitrate * 4 wraps to a negative value for bitrates above ~537 million)
+                long qsvMaxrate = Math.Min((long)bitrate + 1, int.MaxValue);
+                long qsvInitOcc = Math.Min((long)bitrate * 2, int.MaxValue);
+                long qsvBufsize = Math.Min((long)bitrate * 4, int.MaxValue);
+
+                return FormattableString.Invariant($"{mbbrcOpt} -b:v {bitrate} -maxrate {qsvMaxrate} -rc_init_occupancy {qsvInitOcc} -bufsize {qsvBufsize}");
             }
 
             if (string.Equals(videoCodec, "h264_amf", StringComparison.OrdinalIgnoreCase)


Shadowghost (approved):

scheilch:
Hi @gnattu and @nyanmisaka — I hope you're doing well! When you have a moment, I'd really appreciate it if you could take a look at this PR. It's a fairly focused, self-contained fix (no API changes, no new dependencies) and comes with a comprehensive unit-test suite. Of course, there's no rush — I just wanted to gently check in. Thank you both for your time and all the work you put into Jellyfin! 🙏

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -1566,14 +1552,18 @@ private string GetVideoBitrateParam(EncodingJobInfo state, string videoCodec)
...
 
-            // Bit rate under 1000k is not allowed in h264_qsv
+            // Bit rate under 1000k is not allowed in h264_qsv.
+            // Also cap to 50 Mbps: H.264 Level 4.2 (most common for 1080p QSV) has a CPB
+            // limit of 62.5 Mbit, so 50 Mbps keeps derived values within spec. HEVC/AV1
+            // encoders have much higher level limits and are not affected by this cap.
             if (string.Equals(videoCodec, "h264_qsv", StringComparison.OrdinalIgnoreCase))
             {
-                bitrate = Math.Max(bitrate, 1000);
+                bitrate = Math.Clamp(bitrate, 1000, 50_000_000);
```
gnattu: I am still concerned about hard capping at 50Mbps for all transcoding. Intel encoders are capable of doing larger bitrates and we should not prevent it to.

gnattu (changes requested):
Sorry but I think the scope of this PR is out of its intention described in the title.

The title wants to fix the int overflow problem which should be trivial, but the actual PR contains a refactor on encoding parameters for QSV which changes the behavior too much.


For media sources that is unreliable for ffmpeg to get the source bitrate, Jellyfin does have built-in cap to set fallback limit bitrate:
https://github.com/jellyfin/jellyfin/blob/f52005768a7022d7f5fe1d4622e818982d3a06f3/Jellyfin.Api/Helpers/StreamingHelpers.cs#L149-L153

But this only applies to internal liveTV sources (M3U and HDHR). For sources provided by plugins, I *think* it is the responsibility of the plugin to set the `FallbackMaxStreamingBitrate` to prevent the bitrate being unreasonably high.

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
nyanmisaka: Thank you for raising this issue. I did not expect the client to pass in an irrational target bitrate 1000 Mbps for our H264, HEVC and AV1 encoders.

We currently do not transcode video streams to resolutions higher than 4096x4096, so hardcoding a limit of 400 Mbps in `EncodingHelper.GetVideoBitrateParamValue()` is more than sufficient, even for 4k120 AV1 video. This also prevents `-bufsize` from overflowing and causing ffmpeg to fail if your client accidentally passes in `(int32_max / 2 + 1)` as target bitrate.

```
{"bufsize", "set ratecontrol buffer size (in bits)", OFFSET(rc_buffer_size), AV_OPT_TYPE_INT, {.i64 = DEFAULT }, INT_MIN, INT_MAX, A|V|E},
```

QSV encoder automatically clamps the set `-b:v` and `-maxrate` based on the set `-level` option. The hardcoded 50 Mbps limit is far below the hardware capability of 300 Mbps (H264 High 5.1).

Therefore, there is no need to maintain a level<->bitrate mapping table. Having such a table outside the ffmpeg or GPU driver is also unreliable. EncodingHelper is in a difficult-to-test situation, it's best not to add tests until the ffmpeg command string builder has been cleaned up and refactored.

nyanmisaka (changes requested):

gnattu (approved):
I think for this PR Int.Max is fine now. For sane upper limit we could apply case by case.

scheilch:
Thank you all for the thorough and constructive feedback  @Shadowghost for the approval, @gnattu and @nyanmisaka for the detailed review comments. It is genuinely appreciated.

I have reworked the PR based on your input:

**What changed:**
- Removed the 50 Mbps bitrate cap for h264_qsv
- Removed the H.264 CPB-level clamping logic and the level-to-CPB lookup table
- Removed the unit tests (acknowledged  EncodingHelper deserves a proper test harness after the planned refactoring)
- The PR now contains only the minimal overflow fix: long arithmetic for qsvMaxrate, qsvInitOcc, qsvBufsize and the general Bufsize = bitrate * 2, each clamped to int.MaxValue

**On the 400 Mbps upstream cap:**
@nyanmisaka's suggestion of capping GetVideoBitrateParamValue at 400 Mbps is a good idea - it would address the root cause for all encoders in one place. However, since #16376 already prevents the overflow arithmetically, I'd prefer to keep that as a follow-up PR so this one stays as tight as possible. Happy to open it after this is resolved if that approach has agreement.

**Small note on FallbackMaxStreamingBitrate:**
@gnattu is right that plugin-provided sources should set FallbackMaxStreamingBitrate. That said, the current mechanism only activates when LiveStreamId is set. Plugin streams that don't register as LiveTV tuners bypass it entirely. The long-arithmetic fix is therefore still a defensive improvement, even if it only guards against the extreme edge rather than the common misconfiguration.

Thanks again for your time.

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -1638,6 +1612,34 @@ private string GetVideoBitrateParam(EncodingJobInfo state, string videoCodec)
             return FormattableString.Invariant($" -b:v {bitrate} -maxrate {bitrate} -bufsize {bufsize}");
         }
 
+        /// <summary>
+        /// Builds the QSV-specific rate-control parameters for h264_qsv, hevc_qsv and av1_qsv.
+        /// Extracted from <see cref="GetVideoBitrateParam"/> to reduce cognitive complexity.
+        /// </summary>
+        private static string GetQsvRateControlParams(EncodingJobInfo state, string videoCodec, int bitrate)
```
nyanmisaka: Could you merge GetQsvRateControlParams() into GetVideoBitrateParam()? There's nowhere else it's used.

scheilch:
@nyanmisaka Done  I have inlined the QSV rate-control logic back into GetVideoBitrateParam() as you suggested. The extracted method is gone; the three long-arithmetic lines now live directly inside the h264_qsv / hevc_qsv / av1_qsv block where they belong. Thanks for the clear feedback!

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -1604,7 +1590,13 @@ private string GetVideoBitrateParam(EncodingJobInfo state, string videoCodec)
 
                 // Set (maxrate == bitrate + 1) to trigger VBR for better bitrate allocation
                 // Set (rc_init_occupancy == 2 * bitrate) and (bufsize == 4 * bitrate) to deal with drastic scene changes
-                return FormattableString.Invariant($"{mbbrcOpt} -b:v {bitrate} -maxrate {bitrate + 1} -rc_init_occupancy {bitrate * 2} -bufsize {bitrate * 4}");
+                // Use long arithmetic and clamp to int.MaxValue to prevent int32 overflow
+                // (e.g. bitrate * 4 wraps to a negative value for bitrates above ~537 million)
+                long qsvMaxrate = Math.Min((long)bitrate + 1, int.MaxValue);
+                long qsvInitOcc = Math.Min((long)bitrate * 2, int.MaxValue);
+                long qsvBufsize = Math.Min((long)bitrate * 4, int.MaxValue);
```
nyanmisaka:
```suggestion
                int qsvMaxrate = (int)Math.Min((long)bitrate + 1, int.MaxValue);
                int qsvInitOcc = (int)Math.Min((long)bitrate * 2, int.MaxValue);
                int qsvBufsize = (int)Math.Min((long)bitrate * 4, int.MaxValue);
```

nyanmisaka (approved):

nyanmisaka (approved):

Bond-009 (approved):

Issue #935, opened 2019-02-18
Client fetching subtitles takes too long


vitorsemeano:
**Describe the bug**
When a jellyfin client tries to get a specific subtitle from a video file that contains subtitle streams, it takes some minutes to get a response from the server. The requested link must be in the following format: 

http://<server address>/emby/Videos/<video id>/<video id>/Subtitles/3/0/Stream.the subtitle format like vtt or ass>?api_key=<api key>

**To Reproduce**
To reproduce, you must have a video file with embed subtitles (subtitle streams, not hardcoded). Make sure you didn't played the file with subtitles yet. Try to restart the emby service to clean cached info (not sure how it caches the parsed subtitles, so it's a wild guess).

1. Open the webserver client or android app
2. Navigate to a file that meets the video file requirements
3. For the subtitle dropdown, select one subtitle
4. Click to play it
5. It can happen two things:
- The file plays instantly, but without subtitles
- The loading circle keeps spinning for some minutes

**Expected behavior**
After point 4, the video file should play normally with the pretended subtitles.

**Logs**
Can't see any error logs related with this issue, the request made by the client side simply takes a lot of time for a first time. Seeing the logs in jellyfin server only gives the request received and returned a few minutes later.

**Screenshots**

**System (please complete the following information):**
 - OS: ubuntu 18.10 64bit. The clients used was the jellyfin webserver included with the server, and the android app built with this version: https://github.com/vitorsemeano/jellyfin-android/tree/fixSubtitles
 - Browser: Firefox
 - Jellyfin Version: 10.1.0

**Additional context**
This issue could be related with #172

definitio:
This is still an issue.

Rassilion:
Happened to me in docker armhf and docker amd64 images.

ovandriyanov:
Happens to me all the time. Kinda annoying. Is there any workaround? E. g. an API call to force subtitle extraction and caching?

RazeLighter777:
Workaround is to deselect and reselect the subtitles

Rassilion:
@razelighter777 that doesn't work for me. Are you using external subs or internal? I'm using internal subs in mkv container.

ovandriyanov:
@RazeLighter777 I can't do it when the video is played with selected subtitles from the very beginning, e. g. when subtitles are enabled by default in settings, or when I select them before playing a video, since in that case I don't have any user interface at all — only the spinning loading circle at the center of my screen.

[...]

PR #16380, opened 2026-03-10
Respect EnableSubtitleExtraction setting in subtitle delivery

### LINKED ISSUE #1163: Disabling "Allow subtitle extraction on the fly" broken
**Describe the bug**
I disabled the option "Allow subtitle extraction on the fly" to prevent jellyfin from spawning a ffmpeg process to extract subtitles on the fly. Sadly it does spawn the process even if the option is disabled.

As I use network device, the subtitle extraction process saturates the line and the transcoding process cannot go on, so the playback is stuck.

**To Reproduce**

1. Go to 'Transcoding'
2. Click on 'Allow subtitle extraction on the fly' to disable the option
3. Start a movie with subtitles
4. See two processes, one for subtitle extraction and one for transcoding

**Expected behavior**
spawn only one ffmpeg(transcoding) process when "Allow subtitle extraction on the fly" is disabled, hardcode subtitles if needed.

**System:**
 - OS: Docker
 - Browser: Chromium
 - Jellyfin Version: 10.2.2

LTe:
**Changes**

Make `MediaEncoder.CanExtractSubtitles()` check the `EnableSubtitleExtraction` encoding option instead of always returning `true`. This causes `StreamBuilder` to select burn-in (`Encode`) instead of external delivery for embedded subtitles during transcoding, preventing a separate ffmpeg extraction process from competing with the transcode for network bandwidth.

The check is scoped to transcoding only — during DirectPlay, subtitle extraction is still allowed since there is no competing transcode process.

**Issues**

Fixes #1163

Related #935

### DIFF
--- a/MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
+++ b/MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
@@ -1331,8 +1331,7 @@ namespace MediaBrowser.MediaEncoding.Encoder
 
         public bool CanExtractSubtitles(string codec)
         {
-            // TODO is there ever a case when a subtitle can't be extracted??
-            return true;
+            return _configurationManager.GetEncodingOptions().EnableSubtitleExtraction;
         }
 
         private sealed class ProcessWrapper : IDisposable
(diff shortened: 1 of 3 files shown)


nyanmisaka (approved):
This seems to make sense. I applied it locally and did some testing, and it doesn't appear to cause any regressions in `PlayMethod.Transcode`.

Shadowghost (approved):

gnattu (approved):

Bond-009 (approved):

PR #16383, opened 2026-03-10
Fix NameStartsWith filter not matching numeric item names

### LINKED ISSUE #1470: API: /Users/{UserID}/Items with NameStartsWith numeric parameter returns no results
**Describe the bug**
I have the movie '12 Years a Slave' in my library. If I use the API to filter to NameStartsWith=1 I get no results. All other alphabetic characters return correct results. I currently have no other numeric named movies or series in my library.

**To Reproduce**

1. Go to http://localhost:8096/emby/Users/{USERID}/Items/?NameStartsWith=1&Recursive=true&api_key={APIKEY}
2. no items returned.

**Expected behavior**

One movie item should be returned representing '12 Years a Slave'.

 - OS: Docker
 - Browser:  Chrome
 - Jellyfin Version: 10.3.5
 - Reverse proxy: no

LTe:
**Changes**
Normalize `NameStartsWith`, `NameStartsWithOrGreater`, and `NameLessThan` filter input through `ModifySortChunks` before comparing against `SortName`, so zero-padding is applied consistently. Add a `TrimStart('0')` fallback for `NameStartsWith` so numeric prefixes like `1` correctly match zero-padded sort names like `0000000012 years a slave`. Add unit tests covering numeric, alphabetic, and mixed prefix matching for all three filter types.

**Issues**
Fixes #1470

### DIFF
--- a/MediaBrowser.Controller/Entities/BaseItem.cs
+++ b/MediaBrowser.Controller/Entities/BaseItem.cs
@@ -959,7 +959,22 @@ namespace MediaBrowser.Controller.Entities
             return ModifySortChunks(sortable);
         }
 
-        internal static string ModifySortChunks(ReadOnlySpan<char> name)
+        /// <summary>
+        /// Normalizes a user-provided name filter value to match against <see cref="SortName"/>.
+        /// Applies the same zero-padding, diacritics removal, and transliteration as sort name generation.
+        /// </summary>
+        /// <param name="input">The raw filter value from the API.</param>
+        /// <returns>The normalized value suitable for comparison against SortName.</returns>
+        public static string NormalizeSortNameFilter(string input)
+            => ModifySortChunks(input).ToLowerInvariant();
+
+        /// <summary>
+        /// Processes a name by zero-padding numeric chunks to 10 digits for natural sort order,
+        /// removing diacritics, and transliterating non-ASCII characters.
+        /// </summary>
+        /// <param name="name">The input name to process.</param>
+        /// <returns>The processed string with padded numeric chunks.</returns>
+        public static string ModifySortChunks(ReadOnlySpan<char> name)
         {
             static void AppendChunk(StringBuilder builder, bool isDigitChunk, ReadOnlySpan<char> chunk)
             {
--- a/MediaBrowser.Controller/Entities/BaseItem.cs
+++ b/MediaBrowser.Controller/Entities/BaseItem.cs
@@ -969,27 +969,33 @@ namespace MediaBrowser.Controller.Entities
                 }
             }
 
+            return ModifySortChunks(ApplySortCharacterRules(sortable, configuration));
+        }
+
+        private static string ApplySortCharacterRules(string name, ServerConfiguration configuration)
+        {
             foreach (var removeChar in configuration.SortRemoveCharacters)
             {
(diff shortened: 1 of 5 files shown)


sparky3387:
This is great, ive built on this to enforce namestartswith using the sortby column, which makes sense to me instead of returning 0 results, maybe we could combine our PR's to one?

Shadowghost:
@LTe can you check if this is still needed on master? I think I fixed this in the performance PR

Shadowghost (approved):

LTe:
@Shadowghost rebased onto current master - the bug is still there. Verified both ways: hitting `/Items?NameStartsWith=1` against a library with `"12 Years a Slave"` / `"1917"` / `"300"` returns 0 items on master, all expected items on this branch. The included unit tests also fail on master (9/25) and pass here

MediaBrowser.Controller/Entities/BaseItem.cs
```diff
@@ -959,7 +959,22 @@ protected virtual string CreateSortName()
...
 
-        internal static string ModifySortChunks(ReadOnlySpan<char> name)
+        /// <summary>
+        /// Normalizes a user-provided name filter value to match against <see cref="SortName"/>.
+        /// Applies the same zero-padding, diacritics removal, and transliteration as sort name generation.
+        /// </summary>
+        /// <param name="input">The raw filter value from the API.</param>
+        /// <returns>The normalized value suitable for comparison against SortName.</returns>
+        public static string NormalizeSortNameFilter(string input)
+            => ModifySortChunks(input).ToLowerInvariant();
```
cvium: Nitpick: The ToLowerInvariant is wasted in a lot of the calls because the string comparison uses IgnoreCase. Maybe that part should be decided by the caller instead

tests/Jellyfin.Controller.Tests/Entities/UserViewBuilderFilterTests.cs
```diff
@@ -0,0 +1,107 @@
...
+    [InlineData("Avatar", "A", true)]
+    [InlineData("Avatar", "a", true)]
+    [InlineData("Avatar", "B", false)]
+    [InlineData("The Matrix", "m", true)]
+    [InlineData("The Matrix", "t", false)]
+    // Mixed alpha-numeric input: handled via ModifySortChunks padding
+    [InlineData("Apollo 13", "apollo 13", true)]
+    [InlineData("Apollo 13", "apollo 2", false)]
+    // Edge case: "0" should only match items whose name is literally "0"
+    [InlineData("0", "0", true)]
```
cvium, line 39: Might be worth adding the movie "0.0MHz"

cvium (approved):
Some comments but otherwise LGTM

tests/Jellyfin.Controller.Tests/Entities/UserViewBuilderFilterTests.cs
LTe: Good catch, I added this specs and it failed (because of the `.`). I change a code a little bit to take into consideration punctuation and hyphens (added specs for that as well).

theguymadmax:
@LTe At this point we are close to a12.0 release and are hesitant to merge any more fixes/behavioral changes to Master unless it's a crucial fix. So this will have to wait until we cut a 12.0 branch.

PR #16399, opened 2026-03-12
feat: aspect ratio computation / black bar detection (uses trickplay data)

alchemyyy:
Demo:

Here is an example that shows the differences between aspect ratio mode, and a demo of "Detected" aspect ratio, running on my server.

https://github.com/user-attachments/assets/dcab3e41-35b4-4e13-aa41-e321aaf14631

**Changes**

Detect the actual content aspect ratio of videos by analyzing existing trickplay tile images for baked in black bars.

This data is stored as a single new value in the TrickplayInfos table. TrickplayInfoDto now sends this new value, and two other values which denote the ratio and name of a best guess at what industry standard aspect ratio the video is, since computing an aspect ratio from a low resolution thumbnail isn't going to be float-perfect.

For each pixel position in a thumbnail-sized grid, the algorithm tallies how many thumbnails have a non-black pixel there, then applies a 5% quorum threshold to filter out brief aspect ratio changes (opening logos, IMAX sequences). The resulting bounding box determines the content region, and the ratio is snapped to the nearest industry standard (e.g. 2.39 "Modern Anamorphic Scope"). The detected ratio is persisted via a new `DetectedAspectRatio` column on `TrickplayInfo` and exposed through `TrickplayInfoDto` with snapped value and industry name.

**Issues**

This solves limitations with aspect ratio adjustment methods like "Zoom", where media will get cropped if the aspect ratio of the player is beyond that of the video. This would also allow for aspect ratio settings like "Stretch" to operate more correctly if a player allows for multiple options, but I digress as the main attraction is working black bar detection.

Any issue pertaining to "double black bars", "double letterboxing", or the "cover" aspect ratio setting not working as someone expected, etc. should be considered relevant.

Here are some pre-existing posts I found concerning this:

https://github.com/jellyfin/jellyfin/issues/16443

https://www.reddit.com/r/jellyfin/comments/1ri8bi9/black_bars_on_all_sides_media_not_fitting/

https://www.reddit.com/r/jellyfin/comments/utnm2x/can_the_transcoder_remove_black_bars/

https://www.reddit.com/r/jellyfin/comments/10su97l/ripped_content_occasionally_displaying_shrunk_in/


alchemyyy:
forgot to make a note here. I'm leaving the 3 commits unsquashed to show off the different approaches I went through.

My thinking behind the DetectedAspectRatio 0 convention was to make it explicit that the job has been run, and there were no detected bars, without adding a second flag value or having that gross empty-but-not-null "system" I had before.  

Right now the tweaked detection algorithm seems to be doing a great job with my library, and I think is just a far more robust approach overall. 

web client PR is here: https://github.com/jellyfin/jellyfin-web/pull/7671

gnattu:
I do not quite understand what this is for and what the problem is of the old behavior. Please enlighten me. The feature's implementation is quite complex for the profit it provides.

alchemyyy:
No problem! I didn't explain it very well.

This is a way to perform true black bar detection. This solves limitations with aspect ratio adjustment methods like "Zoom", where media will get cropped if the aspect ratio of the player is beyond that of the video. This would also allow for aspect ratio settings like "Stretch" to operate more correctly if a player allows for multiple options, but I digress as the main attraction is working black bar detection.

This solution can detect letterboxing of any kind i.e. horizontal, vertical, or both. I remember seeing comments a long time ago about the feasibility of black bar detection, and the common understanding was that it would be incredibly impractical to have a full media library scan just for that. The complexity in this implementation is due to the fact its filtering against a composite stack of very low resolution jpegs, that are sourced from the existing trickplay data. My reasoning is that this can be added to any library without the need to run through every media file if trickplay data already exists for that library. 

Here are some existing posts about this issue:

https://github.com/jellyfin/jellyfin/issues/16443

https://www.reddit.com/r/jellyfin/comments/1ri8bi9/black_bars_on_all_sides_media_not_fitting/

https://www.reddit.com/r/jellyfin/comments/utnm2x/can_the_transcoder_remove_black_bars/

https://www.reddit.com/r/jellyfin/comments/10su97l/ripped_content_occasionally_displaying_shrunk_in/


Here is an example that shows the differences between aspect ratio mode, and a demo of "Detected" aspect ratio, running on my server.

https://github.com/user-attachments/assets/dcab3e41-35b4-4e13-aa41-e321aaf14631

PR #16408, opened 2026-03-13
Fix two host lifecycle, disable setupServer

kimboslice99:
Two host lifecycle breaks ANCM.

Starting and stopping one host before firing up the real host feels awkward, is there some reason this was chosen over a single host?

I am aware the [IIS documentation was all removed](https://github.com/jellyfin/jellyfin.org/pull/1270) as well, so I guess the question is why leave the variable JELLYFIN_ENABLE_IIS at all?

It felt quite rude to delete all the IIS docs, citing "nobody knows". Nobody thought to look at the pull which added the docs?

Do as you wish I guess. I am not interested in contributing further to a community which will not recognize me as a member of it.


JPVenson:
We decided to degrade the IIS feature to an undocumented one as it seemed the fairest option to outright remove the IIS feature. We have no current interest in supporting IIS officially as nobody from the team wants to deal with an IIS nor runs it themself and the current implementation is as-is. 

This is also apparent as it took over 10 months to report this bug so it seems like there is only a small handful of people running it that way. Disabling the startup server is not advisable as we want to extend its functionality in the future and we are planning another separate asp pipeline for setup purposes.

PR #16409, opened 2026-03-13
Fix missing collection folder posters after initial scans.

### LINKED ISSUE #1200: Library thumbs are not created in 10.3.0-rc1
**Describe the bug**
In 10.3 these are how the libraries look:
image
In 10.2:
image
**Expected behavior**
Same as 10.2.
Centos rpm.

elio42:
**Changes**
Currently the metadata for top level folders is refreshed before the actual scan of the media (LibraryManager.cs line 1197). If the library was empty before the scan there will not be any images to use as collection poster.

I created a post scan task that refreshes metadata for any top level collections which don't have a primary image.

**Issues**
Fixes #1200

**Note**
The OP in issue #1200 claimed that they also had issues with poster extraction besides on first scan. But reading through the issue comments it doesn't seem like anyone was ever able to replicate this.


Shadowghost (approved):

theguymadmax:
@elio42  Please fix the merge conflicts

crobibero (approved):

theguymadmax (approved):

PR #16418, opened 2026-03-14
Fix orphan paths remaining  in collection.xml after library removal.

### LINKED ISSUE #15497: Existing media for non existant libraries is not purged.
### Description of the bug

Ok, so admitteldly this is a pretty niche bug that I found while I fixed [#94](https://github.com/jellyfin/jellyfin-plugin-tmdbboxsets/issues/94) with Pr [#113](https://github.com/jellyfin/jellyfin-plugin-tmdbboxsets/pull/113) from [jellyfin-plugin-tmdbboxsets](https://github.com/jellyfin/jellyfin-plugin-tmdbboxsets).

The problem is that existing media entries in non existent libraries aren't purged. (For eg. by running Clean up collections and playlists.)

I will describe the bug in the context of collections, but this could likely effect more than just collections.

### Reproduction steps

1. Make a library with any number of movies.
2. Put any number of movies in a collection.
3. Delete the library in jellyfin.
   - But not the physical file location of the movies.
4. Run the `Clean up collections and playlists` task

- You will now have an empty collection in your collections
- In `config/data/collections/{your_collection}/collections.xml` you will still be able to find a reference to the movie from the library which you deleted.

Note, when you actually delete the movie from disk it will correctly purge it when running `Clean up collections and playlists`.

I would expect jellyfin not to consider file locations that are not pointed to by any library, even if jellyfin does have access to that file location.

### Jellyfin Server version

10.11.0+

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.2

### Environment

```markdown
- OS: Debian 12
- Linux Kernel: 6.12.43+deb12-amd64
- Virtualization: docker
- Hardware Acceleration: none
- Plugins: jellyfin tmdb boxsets installed
- Jellyfin Data Storage: local M.2 NVME SSD
- Media Storage: local HDD
```

### Jellyfin logs

```shell
n/a
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

TiagoMateus05:
**Problem**
When a Library with movies is removed, if an arbitrary number of movies belonged to a collection, the collection.xml structure would keep the Paths of those movies without being valid. 

**Changes**
Fixed `CleanupCollectionAndPlaylistPathsTask` so Paths that do not belong to a valid Library, do not stay  in the `collection.xml`.
The verification for a valid Path in the `collection.xml` is now done by using the `IFileSystem` and `ILibraryManager` as helper classes, and new functions `GetLinkedChildById` and `IsPathInActiveLibrary`, to better define which Items should be valid for the collection or not. 
The CleanupCollectionsAndPlaylistTask will now remove Orphan Entries, such as: `<Path>/Invalid/path/to/movie<Path>`, where the item does not belong to any Library, even if the file still exists on disk. 
Created a respective unit test to better verify the expected behavior.

**Issues**
Fixes #15497

### DIFF
--- a/Emby.Server.Implementations/ScheduledTasks/Tasks/CleanupCollectionAndPlaylistPathsTask.cs
+++ b/Emby.Server.Implementations/ScheduledTasks/Tasks/CleanupCollectionAndPlaylistPathsTask.cs
@@ -27,6 +27,8 @@ public class CleanupCollectionAndPlaylistPathsTask : IScheduledTask
     private readonly IPlaylistManager _playlistManager;
     private readonly ILogger<CleanupCollectionAndPlaylistPathsTask> _logger;
     private readonly IProviderManager _providerManager;
+    private readonly ILibraryManager _libraryManager;
+    private readonly IFileSystem _fileSystem;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="CleanupCollectionAndPlaylistPathsTask"/> class.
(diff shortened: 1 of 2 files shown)


theguymadmax:
@ TiagoMateus05, thanks for the PR, and apologies for missing that issue with your post. We recently rewrote how Jellyfin handles collections in #16062, which serves as the foundation for the upcoming 12.0 release. 

The issue this PR targets should already be resolved by that change. You can test out that branch, and if it still occurs, I’d recommend submitting a fix after that PR has been merged, as there are significant changes, such as the removal of the CleanupCollection task.
Thanks.

PR #16419, opened 2026-03-14
Add callback for segment data pruning to IMediaSegmentProvider

Shadowghost:
When we remove item data, we should also call the segment providers so they can prune their cached data. This might be necessary if they use related-analysis e.g. when using Chromaprint comparison.

### DIFF
--- a/Jellyfin.Server.Implementations/MediaSegments/MediaSegmentManager.cs
+++ b/Jellyfin.Server.Implementations/MediaSegments/MediaSegmentManager.cs
@@ -182,12 +182,6 @@ public class MediaSegmentManager : IMediaSegmentManager
     /// <inheritdoc />
     public async Task DeleteSegmentsAsync(Guid itemId, CancellationToken cancellationToken)
     {
-        var db = await _dbProvider.CreateDbContextAsync(cancellationToken).ConfigureAwait(false);
-        await using (db.ConfigureAwait(false))
-        {
-            await db.MediaSegments.Where(e => e.ItemId.Equals(itemId)).ExecuteDeleteAsync(cancellationToken).ConfigureAwait(false);
-        }
-
         foreach (var provider in _segmentProviders)
         {
             try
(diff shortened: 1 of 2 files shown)


MediaBrowser.Controller/MediaSegments/IMediaSegmentProvider.cs
```diff
@@ -31,4 +32,16 @@ public interface IMediaSegmentProvider
...
+
+    /// <summary>
+    /// Called when extracted segment data for an item is being pruned.
+    /// Providers should delete any cached analysis data they hold for the given item.
+    /// </summary>
+    /// <param name="itemId">The item whose data is being pruned.</param>
+    /// <param name="cancellationToken">Abort token.</param>
+    /// <returns>A task representing the asynchronous cleanup operation.</returns>
+    Task CleanupExtractedData(Guid itemId, CancellationToken cancellationToken)
+    {
```
JPVenson: unnessesary

Jellyfin.Server.Implementations/MediaSegments/MediaSegmentManager.cs
```diff
@@ -187,6 +187,18 @@ public async Task DeleteSegmentsAsync(Guid itemId, CancellationToken cancellatio
         {
             await db.MediaSegments.Where(e => e.ItemId.Equals(itemId)).ExecuteDeleteAsync(cancellationToken).ConfigureAwait(false);
         }
+
+        foreach (var provider in _segmentProviders)
```
JPVenson, line 185: i would do this before actually removing the segments in case providers want to do somthing based on those segments

MediaBrowser.Controller/MediaSegments/IMediaSegmentProvider.cs
Shadowghost: Why is this unnecessary? I added a default so exsiting do not breask, but I can remove it if that is what you meant

JPVenson: existing would need to recompile anyway and cleanup is something they should always do so i dont think its a good idea to have the default interface implementation here. Also its a pattern we dont really use anywhere else in the code

JPVenson (approved):

PR #16422, opened 2026-03-16
fix(playlists): performance, stability, missing-item handling, export/import, and client management features

trumblejoe:
## Summary

This PR addresses a cluster of long-standing playlist issues and adds several features that clients have been lacking.

### Bug fixes

- **Performance** — `GetPlaylistItems` previously resolved all N linked children before applying pagination, causing O(N) DB hits on every page load regardless of page size. A new `GetManageableItems(startIndex, maxCount)` overload on `Playlist` slices the `LinkedChildren` array first (O(1)), then resolves only the items needed for the requested page.

- **Stable PlaylistItemId** — `LinkedChild.Create()` now eagerly caches `ItemId` on the linked child so `PlaylistItemId` is always unique and available without a lazy resolution round-trip.

- **Missing-item handling (non-destructive)** — Previously `PlaylistManager` subscribed to `ILibraryManager.ItemRemoved` and silently deleted playlist slots whenever a media file was removed from the library. This was harmful in common upgrade scenarios (file re-downloaded with a different filename). The auto-delete handler has been replaced with an `ItemUpdated` subscriber that heals `LinkedChild.Path` when an item's path changes in-place (same Guid, new path — e.g. file renamed, replaced at same location). Slots for truly missing items are left in place; they show up as broken in the API and can be cleaned up explicitly by the user.

- **`MoveItemAsync` bounds crash** — Added guards for empty accessible-children list and out-of-range `newIndex` (`Math.Clamp`, early return on `oldIndexAccessible < 0`).

### New API endpoints

| Method | Route | Description |
|--------|-------|-------------|
| `GET` | `/Playlists/{id}/Items/Broken` | List PlaylistItemIds whose library item can no longer be resolved |
| `DELETE` | `/Playlists/{id}/Items/Broken` | Remove all broken entries (explicit user action) |
| `POST` | `/Playlists/{id}/Items/Order` | Bulk reorder — accepts ordered `EntryIds` array, replaces one-at-a-time `MoveItem` calls for drag-and-drop clients |
| `GET` | `/Playlists/{id}/Export?format=m3u8\|json` | Download playlist as extended M3U8 (absolute paths) or portable Jellyfin JSON (keyed on ProviderIds — IMDB/TMDB/TVDB) |
| `POST` | `/Playlists/Import` | Upload M3U/M3U8/PLS/WPL/ZPL (path-matched) or Jellyfin JSON (ProviderId-matched, title+year fallback); unmatched items skipped |
| `POST` | `/Playlists/{id}/Clone` | Create an owned copy; optional `newName` param, defaults to `"{name} (Copy)"` |
| `POST` | `/Playlists/{id}/Items/Shuffle` | Fisher-Yates shuffle stored server-side so all clients see the same order |
| `GET` | `/Playlists/{id}/Items/Duplicates` | List PlaylistItemIds of second+ occurrences of the same item |
| `DELETE` | `/Playlists/{id}/Items/Duplicates` | Remove all duplicates, keeping first occurrence |

### New model

`MediaBrowser.Model.Playlists.PlaylistExportDto` / `PlaylistExportItem` — portable export format carrying `ProviderIds`, `Name`, `Year`, episode/series metadata, and artist/album fields so a playlist can round-trip between Jellyfin servers that share the same content regardless of local file paths.

## Test plan

- [ ] Large playlist (500+ items): confirm `GET /Playlists/{id}/Items?limit=20` returns in milliseconds with no full-table item resolution
- [ ] Delete a media file from the library; confirm the playlist slot stays present and appears in `GET /Playlists/{id}/Items/Broken`
- [ ] Replace a media file with a renamed upgrade; confirm the playlist slot heals automatically after library scan
- [ ] `DELETE /Playlists/{id}/Items/Broken` removes only unresolvable slots
- [ ] `POST /Playlists/{id}/Items/Order` with full EntryIds array reorders correctly
- [ ] `POST /Playlists/{id}/Items/Shuffle` produces a different stored order; subsequent `GET /Items` reflects it
- [ ] `GET /Playlists/{id}/Export?format=m3u8` downloads a valid extended M3U8
- [ ] `GET /Playlists/{id}/Export?format=json` downloads JSON; re-import via `POST /Playlists/Import` on same server recreates the playlist
- [ ] Import a `.m3u` file with paths present in the library; confirm matching items are added
- [ ] `POST /Playlists/{id}/Clone` creates an independent copy with all items
- [ ] Add a duplicate item; `GET /Items/Duplicates` lists it; `DELETE /Items/Duplicates` removes it

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Emby.Server.Implementations/Playlists/PlaylistManager.cs
+++ b/Emby.Server.Implementations/Playlists/PlaylistManager.cs
@@ -7,6 +7,8 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Text.Json;
+using System.Text.Json.Serialization;
 using System.Threading;
 using System.Threading.Tasks;
 using Jellyfin.Data.Enums;
@@ -15,6 +17,7 @@ using Jellyfin.Extensions;
 using MediaBrowser.Controller.Dto;
 using MediaBrowser.Controller.Entities;
 using MediaBrowser.Controller.Entities.Audio;
+using MediaBrowser.Controller.Entities.TV;
 using MediaBrowser.Controller.Extensions;
 using MediaBrowser.Controller.Library;
 using MediaBrowser.Controller.Playlists;
@@ -33,6 +36,14 @@ namespace Emby.Server.Implementations.Playlists
 {
     public class PlaylistManager : IPlaylistManager, IDisposable
     {
+        private static readonly JsonSerializerOptions PlaylistJsonOptions = new()
+        {
+            WriteIndented = true,
+            PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
+            PropertyNameCaseInsensitive = true,
+            DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull
+        };
+
         private readonly ILibraryManager _libraryManager;
         private readonly IFileSystem _fileSystem;
         private readonly ILibraryMonitor _iLibraryMonitor;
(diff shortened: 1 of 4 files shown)


nielsvanvelzen:
Thank you for your contribution. Unfortunately we won't be accepting it as it violates our [LLM/"AI" Development Policy](https://jellyfin.org/docs/general/contributing/llm-policies).

PR #16431, opened 2026-03-18
Return created API key from POST /Auth/Keys endpoint

### LINKED ISSUE #16258: POST on /Auth/Keys does not return the created api key
### Description of the bug

Hi,

I'm unsure if this is a bug or intended behavior, but the POST /Auth/Keys api endpoint doesn't return the created API key.
This means we need to GET /Auth/Keys and filter by app name which is not necessarily unique.

Is there a reason not to return the API key on creation?
If this is intended behavior feel free to close the issue.

Thanks.

### Reproduction steps

1. Obtain auth token
2. POST /Auth/Keys
3. Nothing gets returned by the API

Nothing get returned.

The created API key gets returned in the response.

### Jellyfin Server version

10.11.6

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.6

### Environment

```markdown
- OS: Debian 13
- Linux Kernel: 6.12.57+deb13-amd64
- Virtualization: No
- Clients: N/A
- Browser: N/A
- FFmpeg Version: N/A
- Playback Method: N/A
- Hardware Acceleration: N/A
- CPU Model: Intel(R) Xeon(R) CPU E5-2470 v2 @ 2.40GHz
- GPU Model: N/A
- Plugins: None
- Reverse Proxy: nginx
- Base URL: None
- Networking: Docker Bridge
- Media Storage & Filesystem: ZFS raidz1
- External Integrations: None
```

### Jellyfin logs

```shell
N/A
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

mango766:
**Changes**

The `POST /Auth/Keys` endpoint currently returns `204 No Content` after creating an API key, which means callers have no way to get the generated access token back without making a follow-up `GET /Auth/Keys` request and filtering by app name (which isn't guaranteed unique). This changes the endpoint to return `200 OK` with the created `AuthenticationInfo`, matching the shape already used by `GET /Auth/Keys`. The `IAuthenticationManager.CreateApiKey` interface and its implementation are updated to return the new key info.

**Issues**

Fixes #16258

### DIFF
--- a/Jellyfin.Api/Controllers/ApiKeyController.cs
+++ b/Jellyfin.Api/Controllers/ApiKeyController.cs
@@ -46,16 +46,16 @@ public class ApiKeyController : BaseJellyfinApiController
     /// Create a new api key.
     /// </summary>
     /// <param name="app">Name of the app using the authentication key.</param>
-    /// <response code="204">Api key created.</response>
-    /// <returns>A <see cref="NoContentResult"/>.</returns>
+    /// <response code="200">Api key created.</response>
+    /// <returns>An <see cref="AuthenticationInfo"/> containing the new api key.</returns>
     [HttpPost("Keys")]
     [Authorize(Policy = Policies.RequiresElevation)]
-    [ProducesResponseType(StatusCodes.Status204NoContent)]
-    public async Task<ActionResult> CreateKey([FromQuery, Required] string app)
+    [ProducesResponseType(StatusCodes.Status200OK)]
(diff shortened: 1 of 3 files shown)


nielsvanvelzen (changes requested):
We shouldn't leak internal models in the API. This does currently happen for the api key responses, we first have to fix that by adding a dto before changing this specific endpoint. Otherwise this would be a good change.

Bond-009:
Closing as stale, no response to review comment

PR #16435, opened 2026-03-19
🌟 feat: Add "KeepOriginalTitle" library option

schullq:
### Description

Implements the "Keep Original Title" library option.
Feature request: https://features.jellyfin.org/posts/32/keep-original-title-option
Discussion: https://github.com/jellyfin/jellyfin-meta/discussions/80

When enabled on a library, movies and series display their original language title instead of the translated title, while still fetching metadata (descriptions, etc.) in the preferred language.

The translated title is kept in the OriginalTitle property instead (I just swapped them).

### Changes

- Added `KeepOriginalTitle` boolean to `LibraryOptions`
- Modified `DtoService` to swap Name/OriginalTitle when option is enabled


dmitrylyzo:
There's another PR with the same purpose: https://github.com/jellyfin/jellyfin-web/pull/7629. It swaps the titles on the client-side and doesn't require server changes.

schullq:
> There's another PR with the same purpose: [jellyfin/jellyfin-web#7629](https://github.com/jellyfin/jellyfin-web/pull/7629). It swaps the titles on the client-side and doesn't require server changes.

Well, as I understand this PR, it only swaps the title on the web client. What happens on all of the other client ?

dmitrylyzo:
> What happens on all of the other client ?

Obviously, non-web-based clients need to be updated.
Also, that PR doesn't force using the original title for all users (library-wise).

UPD. I am not saying it is "bad", just not to miss another approach.

schullq:
Of course, just trying to understand which approach would be best here.

In my opinion, changing the server and the web UI (to select the option) is better than having to update all the clients to achieve the same result, but I don't know what are the common rules when it comes to updating the server (is it best not to touch the server at all, even though it's not a big change).

For your second point, I'll check it, thanks.

nielsvanvelzen (changes requested):
The necessary data is already available client-side. We shouldn't swap this server side as it causes unexpected API behavior.

PR #16439, opened 2026-03-19
build: add Jellyfin.MediaEncoding.Keyframes reference

jumoog:
**Changes**
Added a project reference to `Jellyfin.MediaEncoding.Keyframes` in `Jellyfin.Server.csproj` to expose the keyframes assembly to plugins. The assembly is now copied to the server's output directory, allowing plugins like IntroSkipper to load and use `Jellyfin.MediaEncoding.Keyframes.KeyframeData`.

**Issues**
Plugins cannot load `Jellyfin.MediaEncoding.Keyframes` assembly (FileNotFoundException)

> IntroSkipper.ScheduledTasks.DetectSegmentsTask: An unexpected error occurred during analysis.
System.IO.FileNotFoundException: Could not load file or assembly 'Jellyfin.MediaEncoding.Keyframes, Version=10.11.6.0, Culture=neutral, PublicKeyToken=null'. The system cannot find the file specified.

### DIFF
--- a/Jellyfin.Server/Jellyfin.Server.csproj
+++ b/Jellyfin.Server/Jellyfin.Server.csproj
@@ -64,6 +64,7 @@
     <ProjectReference Include="..\src\Jellyfin.Drawing.Skia\Jellyfin.Drawing.Skia.csproj" />
     <ProjectReference Include="..\src\Jellyfin.LiveTv\Jellyfin.LiveTv.csproj" />
     <ProjectReference Include="..\Jellyfin.Server.Implementations\Jellyfin.Server.Implementations.csproj" />
+    <ProjectReference Include="..\src\Jellyfin.MediaEncoding.Keyframes\Jellyfin.MediaEncoding.Keyframes.csproj" />
     <ProjectReference Include="..\src\Jellyfin.MediaEncoding.Hls\Jellyfin.MediaEncoding.Hls.csproj" />
     <ProjectReference Include="..\src\Jellyfin.Database\Jellyfin.Database.Implementations\Jellyfin.Database.Implementations.csproj" />
   </ItemGroup>


jumoog:
This could also be backported to 10.11, as it is not a breaking change.

Shadowghost (approved):

jumoog:
are there any concerns? would be nice to have it in 12.0

JPVenson (approved):

JPVenson (changes requested):
As discussed in chat, implicitly loading the assembly is not desirable.

Load it explicitly via
https://github.com/jellyfin/jellyfin/blob/622947e37425f3620432995cde5d4a0809d91694/Jellyfin.Server/CoreAppHost.cs#L115

jumoog:
No its not working 

#### 1  `PluginLoadContext` resolves only from the plugin directory

`Emby.Server.Implementations/Plugins/PluginLoadContext.cs:17–32`

```csharp
public PluginLoadContext(string path) : base(true)
{
    _resolver = new AssemblyDependencyResolver(path); // resolves from plugin dir only
}

protected override Assembly? Load(AssemblyName assemblyName)
{
    var assemblyPath = _resolver.ResolveAssemblyToPath(assemblyName);
    if (assemblyPath is not null)
        return LoadFromAssemblyPath(assemblyPath);

    return null; // falls back to AssemblyLoadContext.Default
}
```

`AssemblyDependencyResolver` is seeded with the **plugin's path**. It reads
the plugin's `.deps.json` (if present) and the plugin directory. It has no
knowledge of the server's own assemblies.

When the resolver returns `null`, .NET falls back to `AssemblyLoadContext.Default`.
That only succeeds if the assembly has already been loaded into the default
context — and that is the problem.

#### 2 `Jellyfin.MediaEncoding.Keyframes` is loaded lazily, not eagerly

`Emby.Server.Implementations/ApplicationHost.cs:822–866`

```csharp
protected IEnumerable<Assembly> GetComposablePartAssemblies()
{
    // ...
    yield return typeof(MediaBrowser.MediaEncoding.Encoder.MediaEncoder).Assembly;
    // ...
    yield return typeof(DynamicHlsPlaylistGenerator).Assembly; // ← Jellyfin.MediaEncoding.Hls
    // Jellyfin.MediaEncoding.Keyframes is NOT listed here
    foreach (var i in GetAssembliesWithPartsInternal()) { yield return i; }
}
```

`Jellyfin.MediaEncoding.Keyframes` is only a **transitive** dependency of
`Jellyfin.MediaEncoding.Hls`. It is loaded into the default context the first
time HLS processing actually executes — not at startup. Until that point, it
does not exist in `AssemblyLoadContext.Default`, so the plugin's fallback
resolution fails.

#### 3 `GetAssembliesWithPartsInternal` doesn't help

`Jellyfin.Server/CoreAppHost.cs:115–128`

```csharp
protected override IEnumerable<Assembly> GetAssembliesWithPartsInternal()
{
    yield return typeof(CoreAppHost).Assembly;
    yield return typeof(JellyfinDbContext).Assembly;
    yield return typeof(ServiceCollectionExtensions).Assembly;
    yield return typeof(LiveTvManager).Assembly;
    // Jellyfin.MediaEncoding.Keyframes is not listed here either
}
```

This method exists to enumerate assemblies for MEF/DI scanning. Adding an
assembly here loads it at startup into the default context, which would make
it visible to plugins via the `PluginLoadContext` fallback path. But
`Jellyfin.MediaEncoding.Keyframes` is absent from both lists.

#### 4 Contrast with assemblies that *do* work for plugins

Assemblies like `Jellyfin.Controller` (`IServerApplicationHost`) and
`MediaBrowser.MediaEncoding.Encoder.MediaEncoder` are listed explicitly in
`GetComposablePartAssemblies`. They are loaded eagerly at server startup into
`AssemblyLoadContext.Default`, so the plugin's null-return fallback always
finds them.

`Jellyfin.MediaEncoding.Keyframes` has no such explicit listing and
`KeyframeManager` (registered in DI at `ApplicationHost.cs:571`) does not
trigger eagerly loading the assembly into the default context in time.

### Proposed fix

Add `Jellyfin.MediaEncoding.Keyframes` explicitly to `GetComposablePartAssemblies`
in `ApplicationHost.cs`, alongside the other MediaEncoding assemblies:

```csharp
// MediaEncoding
yield return typeof(MediaBrowser.MediaEncoding.Encoder.MediaEncoder).Assembly;

// Keyframes (required for plugin access via PluginLoadContext fallback)
yield return typeof(Jellyfin.MediaEncoding.Keyframes.KeyframeData).Assembly;
```

This ensures the assembly is eagerly loaded into `AssemblyLoadContext.Default`
at server startup, making it reliably available to any plugin that references it.

Alternatively, `PluginLoadContext.Load()` could be extended to probe the server's
application base directory for assemblies it cannot find in the plugin directory —
a more general solution that would benefit all future server-internal assemblies
that plugins might need.

jumoog:
#16755

Issue #2547, opened 2020-03-09
Subtitles load after a while and can be out of sync when resuming


ikomhoog:
**Describe the bug**
When i watch anime on my phone the subtitles can take up to 30 seconds to appear, eventhough there should have been subtitles already.
when they appear they do show the correct text, but im just missing the first 30 seconds of the conversation.
When i skip through a video or resume, the timing of the subtitles is off, this can be a minor difference, but i also had a few times where it was 2 sentences off.
i have to quit and resume again to try and get the subtitles to line up correctly.
or i have to watch the video from the beginning, then the timing is always correct.(i just miss the first bit when i do that).
Also the issue is there that subtitles dont work at all when playing on the android app.

**System (please complete the following information):**
 - OS: Unraid (some linux version?)
 - Virtualization: Docker
 - Clients: Android and Chrome on Android
 - Browser: Chrome 80
 - Jellyfin Version: 10.5.0
 - Playback: Any
 - Installed Plugins: Opensubtitles(these subs are not always from the plugin), Anime, Audio, MusicBrainz
 - Reverse Proxy: nginx-proxy-manager(docker container for nginx with a userfriendly UI)
 - Base URL: jellyfin.mydomain.com
 - Networking: Bridge
 - Storage: XFS

**To Reproduce**
just try to play a video on your (android)  phone and skip around to a different part of the video.
In the app you wont get subtitles, and on chrome the subtitles will be out of sync.
note: you can also see that the subs will take a while to load on chrome.

**Expected behavior**
The subtitles should have the correct timing.

hypothesis:
there are probably a seperate timer for the subtitles, that timer does not correspond the playback time of the video and gets out of sync when the video needs to buffer like when its skipped.
As for the delay, i think the video starts playing as soon as its buffered, but the subs have a lower priority and thus are only loaded after the video is buffered.

potential fixes, keep in mind that i didnt look at the code (yet) and thus might be totally wrong.
I will try to look into the code soon so i try can help out.

potential fix for the delay before subs appear:
load subtitles before the video, this means that the first transcoding process also can be done while the subtitles are being send, this way you dont have a performance difference.
what i think happens now:
1. Transcode video part.
2. send video part.
3. repeat step 1 and 2 untill the buffer is full enough to play the video
4. send subs
What i think would fix this problem:
1. Transcode video part.
2. (While transcoding) send subs (based on my idea that its the whole file being send)
3. send video part.
[...]

Issue #172, opened 2018-12-15
Subtitle extraction is slow


nvllsvm:
With subtitles disabled, an 11GB mkv started playback almost instantly. With subtitles enabled, it took around 60 seconds before playback began.

`ffprobe` and `mkvinfo` command output.

[30_days_of_night.mkvinfo.txt](https://github.com/jellyfin/jellyfin/files/2683128/30_days_of_night.mkvinfo.txt)
[30_days_of_night.ffprobe.txt](https://github.com/jellyfin/jellyfin/files/2683129/30_days_of_night.ffprobe.txt)

JustAMan:
Isn't this a ffmpeg thing?..

nvllsvm:
Yeah, but there might be something we can do here.

JustAMan:
We may want to add an option like "Don't wait for subtitles to be extracted, start playback straight away", or "Extract subtitles upon import". Opinions?

nvllsvm:
I agree. With a spinning indicator would be great.

MatMaul:
I will have a look, but perhaps instead of extracting subtitles we could convert it on the fly with ffmpeg to webvtt which seems supported by browsers in an HLS stream.
This would cut the need from extracting them.

MatMaul:
And still avoid video transcoding to burn the subtitles.

MatMaul:
No that's the point, you only need to extract because we are providing an out of band srt to the player, which need to be fully there before the playing begins. If the subtitles are in the HLS stream no need to extract, they will be converted by ffmpeg on the fly and added chunk by chunk to the HLS stream.

MatMaul:
Talking text subtitles here. For picture subtitles we don't really have a choice for now indeed, we would need quite some logic client side to handle the overlay.

JustAMan:
I'm not particularly fond of HLS - some my content is in vp9/vorbis, other in x265/aac. Neither fits to hls causing transcoding on the fly.
I'd rather look into DASH as a possible alternative to HLS (or looked into something that would allow me to select audio tracks - right now this html5 api support is... flaky at best).

MatMaul:
DASH looks nice indeed. It seems to supports embedded webvtt subtitles too.

Tthecreator:
[...]

PR #16440, opened 2026-03-19
Remove -copyts and add -flush_packets 1 to subtitle extraction

### LINKED ISSUE #16438: Subtitle extraction produces 0-byte SRT files during playback due to ffmpeg output buffering
### Description of the bug

Embedded SRT subtitle extraction from MKV files produces 0-byte `.srt` files that persist for the entire duration of the extraction (minutes), causing subtitles to be unavailable during playback.

Two issues in `SubtitleEncoder` combine to cause this:

  1. The `-copyts` flag is passed to ffmpeg for subtitle copy-extraction     ([line 656](https://github.com/jellyfin/jellyfin/blob/master/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs#L656),     [line 689](https://github.com/jellyfin/jellyfin/blob/master/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs#L689)).     For `-c:s copy` to SRT, this flag is unnecessary — SRT carries its own text-based timestamps.     With `-copyts`, ffmpeg must sequentially demux the entire video file instead of skipping     non-subtitle packets, making extraction ~5x slower (~8x realtime vs ~44x realtime).

  2. ffmpeg's SRT muxer does not flush output buffers during processing. Subtitle data is tiny     (~60KB total) so the internal buffer never fills. The `.srt` files remain exactly 0 bytes on     disk until ffmpeg finishes processing the entire file and exits. For a 70-minute episode,     this takes ~10 minutes.

Since extraction runs in parallel with playback, the web client requests the subtitle file  immediately. It receives a 0-byte response, shows no subtitles, and does not retry after  extraction completes. The user must stop and restart playback to get subtitles.

  Related: #3001 (same symptom, different root cause), #2547 (subtitles delayed on load — may  share this root cause).

### Reproduction steps

  1. Play an MKV file with embedded SRT subtitle tracks (not previously cached)
  2. Select a subtitle track
  3. Observe that subtitles do not appear
  4. Check the extracted subtitle file on disk at `<data>/data/subtitles/<id>/<stream>.srt`     , it will be 0 bytes while ffmpeg is still running
  5. Wait ~10 minutes for extraction to finish, then stop and restart playback
  6. Subtitles now appear

Subtitle files are 0 bytes for the entire extraction duration (up to 10+ minutes for a  70-minute episode). The player receives empty subtitle data and shows no subtitles.  The user must restart playback after extraction finishes.

Subtitles should be available within seconds of starting playback. Extraction should write  data to disk progressively, or Jellyfin should wait for xtraction to complete before  attempting to serve the subtitle file.

### Jellyfin Server version

10.11.6

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.6

### Environment

```markdown
- OS: Ubuntu 24.04.4 LTS
- Linux Kernel: 6.8.0-106-generic
- Virtualization: Docker (linuxserver/jellyfin:latest)
- Clients: Browser
- Browser: Firefox
- FFmpeg Version: 7.1.3-Jellyfin
- Playback Method: HLS transcode (video direct stream, audio transcode)
- CPU Model: Intel Core i7-4790
- GPU Model: NVIDIA GeForce RTX 4070 SUPER
- Plugins: SubBuzz, Moonfin, JellyBridge, File Transformation, Jellyfin Enhanced
- Reverse Proxy: Caddy
- Base URL: none
- Networking: Bridge
- Jellyfin Data Storage & Filesystem: local NVMe SSD - ext4
- Media Storage & Filesystem: local HDD (mergerfs pool) - ext4
```

### Jellyfin logs

```shell
[21:07:10] [INF] MediaBrowser.MediaEncoding.Subtitles.SubtitleEncoder: /usr/lib/jellyfin-ffmpeg/ffmpeg -i file:"/data/tvshows/The Ministry of Time/Season 1/The Ministry of Time - S01E01 - Time Is What It Is WEBDL-1080p.mkv"
  -copyts -map 0:2 -an -vn -c:s copy "/config/data/data/subtitles/27/278d0e17-97e8-a975-6b12-c2ffcf9aee0f/2.srt" -map 0:3 -an -vn -c:s copy "/config/data/data/subtitles/27/278d0e17-97e8-a975-6b12-c2ffcf9aee0f/3.srt" -map 0:4 -an
  -vn -c:s copy "/config/data/data/subtitles/27/278d0e17-97e8-a975-6b12-c2ffcf9aee0f/4.srt" -map 0:5 -an -vn -c:s copy "/config/data/data/subtitles/27/278d0e17-97e8-a975-6b12-c2ffcf9aee0f/5.srt"

  # Files remain 0 bytes during extraction:
  # -rw-r--r-- 1 root root 0 Mar 19 20:54 2.srt
  # -rw-r--r-- 1 root root 0 Mar 19 20:54 3.srt
  # -rw-r--r-- 1 root root 0 Mar 19 20:54 4.srt
  # -rw-r--r-- 1 root root 0 Mar 19 20:54 5.srt

  # Extraction completes ~2 minutes later (after removing -copyts; ~10 min with -copyts):
  [21:09:06] [INF] MediaBrowser.MediaEncoding.Subtitles.SubtitleEncoder: ffmpeg subtitle extraction completed for file:"/data/tvshows/The Ministry of Time/Season 1/The Ministry of Time - S01E01 - Time Is What It Is
  WEBDL-1080p.mkv" to /config/data/data/subtitles/27/278d0e17-97e8-a975-6b12-c2ffcf9aee0f/2.srt

  # After completion, files have correct data:
  # -rw-r--r-- 1 root root 60926 Mar 19 21:09 2.srt
  # -rw-r--r-- 1 root root 57668 Mar 19 21:09 3.srt
  # -rw-r--r-- 1 root root 61509 Mar 19 21:09 4.srt
  # -rw-r--r-- 1 root root 57511 Mar 19 21:09 5.srt
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

**Verified workaround:** An ffmpeg wrapper script that intercepts subtitle extraction commands and:

  1. Removes `-copyts` (unnecessary for `-c:s copy` to SRT — SRT has its own text timestamps)
  2. Injects `-flush_packets 1` as a per-output option before each subtitle output path

  With both changes, subtitle files have valid data within 3 seconds of extraction starting.

  **Suggested fix in
  [`SubtitleEncoder.cs`](https://github.com/jellyfin/jellyfin/blob/master/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs):**

  - Remove `-copyts` from subtitle extraction commands (lines ~656, ~689)
  - Add `-flush_packets 1` as a per-output option before each subtitle output path

  **Test results** (70-minute MKV, 4 embedded SRT tracks, jellyfin-ffmpeg 7.1.3):

  | | With `-copyts` (current) | Without `-copyts` | Without `-copyts` + `-flush_packets 1` |
  |---|---|---|---|
  | File has data after | never (0 bytes until exit) | never (0 bytes until exit) | **3 seconds** |
  | Extraction completes | ~10 minutes | ~2 minutes | ~2 minutes |
  | Processing speed | ~8x realtime | ~44x realtime | ~44x realtime |

Molier:
**Changes**

  Remove `-copyts` and add `-flush_packets 1` (per-output) to subtitle extraction in `SubtitleEncoder.cs`.

  `-copyts` is unnecessary for `-c:s copy` to SRT (SRT has its own text-based timestamps) and forces ffmpeg to sequentially demux the entire video, making extraction ~5x slower. Without `-flush_packets 1`, ffmpeg's SRT muxer
  buffers all output until the process exits — leaving `.srt` files at 0 bytes for minutes while the web player shows no subtitles.

  With both changes, subtitle files have valid data within 3 seconds of extraction starting.

  | | Current (`-copyts`, no flush) | This PR (no `-copyts`, `-flush_packets 1`) |
  |---|---|---|
  | File has data after | never (0 bytes until exit) | **3 seconds** |
  | Extraction speed | ~8x realtime | ~44x realtime |

  Complements #16257, which handles stale 0-byte files from *failed* extractions. This PR prevents 0-byte files during *successful* extractions.

  **Issues**

  Fixes #16438

### DIFF
--- a/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
+++ b/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
@@ -569,7 +569,7 @@ namespace MediaBrowser.MediaEncoding.Subtitles
                 var outputPaths = new List<string>();
                 var args = string.Format(
                     CultureInfo.InvariantCulture,
-                    "-i {0} -copyts",
+                    "-i {0}",
                     inputPath);
 
                 foreach (var subtitleStream in subtitleStreams)
@@ -594,7 +594,7 @@ namespace MediaBrowser.MediaEncoding.Subtitles
                     outputPaths.Add(outputPath);
                     args += string.Format(
                         CultureInfo.InvariantCulture,
-                        " -map 0:{0} -an -vn -c:s {1} \"{2}\"",
+                        " -map 0:{0} -an -vn -c:s {1} -flush_packets 1 \"{2}\"",
                         streamIndex,
                         outputCodec,
                         outputPath);
@@ -613,7 +613,7 @@ namespace MediaBrowser.MediaEncoding.Subtitles
             var outputPaths = new List<string>();
             var args = string.Format(
                 CultureInfo.InvariantCulture,
-                "-i {0} -copyts",
+                "-i {0}",
                 inputPath);
 
             foreach (var subtitleStream in subtitleStreams)
@@ -639,7 +639,7 @@ namespace MediaBrowser.MediaEncoding.Subtitles
                 outputPaths.Add(outputPath);
                 args += string.Format(
                     CultureInfo.InvariantCulture,
-                    " -map 0:{0} -an -vn -c:s {1} \"{2}\"",
+                    " -map 0:{0} -an -vn -c:s {1} -flush_packets 1 \"{2}\"",
                     streamIndex,
                     outputCodec,
                     outputPath);


IDisposable (approved):
:shipit:

MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
```diff
@@ -594,7 +594,7 @@ private async Task ExtractAllExtractableSubtitlesMKS(
                     outputPaths.Add(outputPath);
                     args += string.Format(
                         CultureInfo.InvariantCulture,
-                        " -map 0:{0} -an -vn -c:s {1} \"{2}\"",
+                        " -map 0:{0} -an -vn -c:s {1} -flush_packets 1 \"{2}\"",
```
nyanmisaka, line 597: `-copyts` should only be needed when providing closed captions (.vtt) via HLS, but we don't do that currently. Therefore, it can be removed when dumping plain caption files.

`-flush_packets` has `auto` as its default value in ffmpeg, but based on your tests, it's clearly not that smart.

One concern I have is that, assuming the video file is large (hundreds of gigabytes and stored remotely) and the caption packets are interleaved with the audio and video packets throughout the MKV file, caption extraction is slow. Enabling `-flush_packets` can write captions to the file immediately, but it doesn't mean the file is completely written. Will the client stop requesting subsequent caption content because it reads an incomplete caption file that was successfully requested the first time? We need to consider SRT, ASS, and PGS captions here.

DrHurt:
PGS (and SRT, ASS) have native built in time stamps and are currently painfully slow to extract so I hope this gets applied to all.

The client I use downloads subtitles once ready so -flush_packets 1 might result in it downloading an incomplete file.

I don’t know about the web client though.

Molier:
Good points, lemme address,
  On the incomplete file concern:

The current behavior is that subtitle files are 0 bytes for the entire extraction duration. The client requests the file, gets an empty response, shows no subtitles, and does not retry. So right now it's not "complete file eventually" - it's "no subtitles at all until you restart playback". -flush_packets 1 can only improve that, since even a partially written SRT is better than 0 bytes.

Looking at the code, `GetReadableFile` awaits `ExtractAllExtractableSubtitles` with a semaphore lock, so the API path shouldn't serve the file mid-extraction anyway. But in practice, the 0-byte files do get served (as described in #16438 with logs), so there may be a separate code path or race condition there.

  What I tested:
  - SRT extraction from local MKV files (1-5GB range), standard episode lengths
  - Web player gets valid subtitle data within ~3 seconds vs never before

  What I haven't tested:
  - Very large remote files where subtitle packets are spread across hundreds of GB
  - PGS or ASS specifically (only SRT)
  - Clients that download the full subtitle file in one request

  For the -copyts removal specifically, I think that's the less controversial part - it's unnecessary for -c:s copy to SRT as
  @nyanmisaka confirmed, and it's what causes the ~5x slowdown.

  Happy to do more targeted testing on PGS/ASS or larger files if that would help move this forward.

DrHurt:
@Molier It would help if you can also provide test results for PGS and ASS as those tend to be particularly slow. If improvements are anywhere near what you're seeing with SRT, it would provide a very tangible improvement to Jellyfin's web based players. Game changer actually.

Molier:
Tested all three formats as requested. Environment: Jellyfin 10.11.6, jellyfin-ffmpeg 7.1.3, Intel i7-4790, media on HDD (SATA), cold cache (page cache dropped between each run via `echo 3 > /proc/sys/vm/drop_caches`).

  Single-stream extraction using `ffmpeg.real` directly inside the container, each config tested independently:

  | Format | File | Config | Total Time | First Non-Zero Data |
  |--------|------|--------|-----------|-------------------|
  | SRT | 69min / 2.2GB | baseline (`-copyts`, no flush) | 14.4s | 14.5s *(at exit)* |
  | SRT | " | no `-copyts` only | 18.4s | 18.4s *(at exit)* |
  | SRT | " | `-flush_packets 1` only | 20.3s | **0.8s** |
  | SRT | " | both fixes | 17.8s | **0.4s** |
  | ASS | 48min / 710MB | baseline | 7.1s | 7.1s *(at exit)* |
  | ASS | " | no `-copyts` only | 6.9s | 6.9s *(at exit)* |
  | ASS | " | `-flush_packets 1` only | 7.0s | **0.3s** |
  | ASS | " | both fixes | 6.9s | **0.3s** |
  | PGS | 22min / 681MB | baseline | 4.7s | 0.6s |
  | PGS | " | no `-copyts` only | 4.5s | 0.4s |
  | PGS | " | `-flush_packets 1` only | 4.5s | 0.3s |
  | PGS | " | both fixes | 4.5s | 0.3s |

  Also tested multi-stream (4 SRT tracks simultaneously, matching what Jellyfin does): baseline 16.2s vs both fixes 20.1s — similar total time, but files have data in under a second with flush.

  **Takeaways:**

  - **SRT and ASS**: without `-flush_packets 1`, output files are 0 bytes for the entire extraction — data only appears when ffmpeg exits. With it, data lands in <1s. This is the actual fix for #16438.
  - **PGS**: doesn't have the 0-byte problem. PGS packets are large enough (~12MB for a 22min episode) that the OS flushes them to disk naturally. Baseline already has data at 0.6s.
  - **`-copyts` removal**: minimal impact on total extraction time with cold HDD cache. Doesn't hurt, but `-flush_packets 1` is doing the heavy lifting.
  - Total extraction speed is roughly the same across configs — the bottleneck is reading the source file, not the subtitle muxer.

  So for @DrHurt's question: ASS gets the same fix as SRT (immediate data vs 0-byte until exit). PGS doesn't need the fix but isn't harmed by it either.

  Re @nyanmisaka's concern about incomplete files being served: the current behavior is that SRT/ASS files are 0 bytes for the full extraction duration — any client that requests them gets nothing and doesn't retry. A partially-written SRT/ASS is strictly better than 0 bytes, and in practice extraction finishes in 7-20s for typical episodes, so the window where a file is incomplete is short.

_formatted and summarized by AI as you can tell_

nyanmisaka (approved):
Since `SubtitleExtractionTimeoutMinutes = 30`, for long videos and slower storage, it's possible that after 10 minutes of extraction, the file might not be flushed to disk, resulting in a file size of 0.

I reread the download endpoint code and found that it shouldn't return data to the client before the ffmpeg process exits naturally. Therefore, any change that speeds up ffmpeg's completion is acceptable: removing `-copyts` makes sense, but the reason why `-flush_packets 1` speeds up ffmpeg's exit is unclear.

gnattu (approved):
It looks low risk enough to me, if it breaks anything it should be very file-specific (I think)

Molier:
> the reason why `-flush_packets 1` speeds up ffmpeg's exit is unclear.

It doesn't? total extraction time is roughly the same across all configs in my tests. 
The win is purely that data hits disk immediately instead of being buffered until exit, which was the issue i encountered which made it so subtitles were not loaded at all.
> Without -flush_packets 1, ffmpeg's SRT muxer
buffers all output until the process exits — leaving .srt files at 0 bytes for minutes while the web player shows no subtitles.

Bond-009 (approved):

DrHurt:
@Molier nyanmisaka says subtitles aren't downloadable until ffmpeg exits so this may not prove as beneficial as we hoped (without further changes)!

Have you tested if the improvements are reflected in web player? (Time for subtitles to first appear)

Right now, the browser doesn’t start rendering PGS until it’s fully downloaded (even if data starts writing to disk immediately).

nyanmisaka:
> > the reason why `-flush_packets 1` speeds up ffmpeg's exit is unclear.
> 
> It doesn't? total extraction time is roughly the same across all configs in my tests. The win is purely that data hits disk immediately instead of being buffered until exit, which was the issue i encountered which made it so subtitles were not loaded at all.
> 
> > Without -flush_packets 1, ffmpeg's SRT muxer
> > buffers all output until the process exits — leaving .srt files at 0 bytes for minutes while the web player shows no subtitles.

Even without `-flush_packets 1`, ffmpeg should have completed writing and flushing a mere KB of text instantly. Otherwise, it's hard not to call it a bug in ffmpeg, as it wastes tens of seconds waiting without doing anything.

DrHurt:
Unfortunately, the speed improvements here aren’t reflected in the web client as it still waits until subtitle extraction is complete.

Tested using 10.11.8

@Molier would you be able to create a supplemental PR to allow web clients to use subtitles as they’re being written to disk. 

I don’t know if web client can read subs directly off the server without downloading as that would be the most logical way (if such access is possible)

Molier:
Can confirm, ran into the same thing on my end. After the merge I tested with a few episodes ( ~2.2GB MKVs with 4 SRT tracks each) and subtitle load times were still 3+ minutes on first play when the HLS transcode was running concurrently.

  Turns out the improvement I initially saw was from recreating my container (`docker compose up --force-recreate`), which cleared stale 0-byte subtitle files from the cache dir. I wrongly attributed that speedup to `-flush_packets 1`, my bad. The flag does what it says (data hits disk in <1s instead of buffering until exit), but as @nyanmisaka pointed out earlier, `GetReadableFile` awaits `ExtractAllExtractableSubtitles` which awaits `process.WaitForExitAsync`, so the HTTP response is held until ffmpeg fully exits regardless. The web player never sees the incremental writes.

  The real killer in my case was I/O contention with two ffmpeg processes (transcode + subtitle extraction) reading the same multi-GB file from a spinning disk. First episode extracted in 15s (transcode barely started), second and third took 3m30s+ each.

  So the changes in this PR are still correct (no reason to buffer subtitle output, no reason for `-copyts` on copy operations), but they don't solve the actual playback delay.

  This is essentially #2547 (open since 2020, 149 comments) and goes back to #172 from 2018. The blocking call chain hasn't changed. For transcode cases specifically, #16380 (burn-in instead of extracting separately) is approved and waiting to merge, that sidesteps the extraction entirely. For everything else, the Subtitle Extract plugin works as a workaround by pre-extracting during library scans so the cache files exist before playback.

  **Neither of those is a real fix** for the core problem though. I've been looking into what it would take to return subtitles before extraction finishes and will follow up with specifics once I've thought it through properly. Any suggestions and relevant context ofc welcome

DrHurt:
A simple yet slightly dirty workaround would be to remove any HTTP blocks and set a timer to keep re-downloading subtitles every 5 seconds once size > 0kb until process exits and the full file is downloaded.

Another idea would be a user option to force remux mkv/mp4 to HLS with webvtt when subtitles are enabled (similar to how you can currently force burn in)

This should work with text subtitles. PGS/VOB would have to remain external or burned.

Final thought is if an API could be provided for player to directly read subtitles from the server without downloading. I don’t know how feasible that is.

Molier:
> A simple yet slightly dirty workaround would be to remove any HTTP blocks and set a timer to keep re-downloading subtitles every 5 seconds once size > 0kb until process exits and the full file is downloaded.
> 
> Another idea would be a user option to force remux mkv/mp4 to HLS with webvtt when subtitles are enabled (similar to how you can currently force burn in)
> 
> This should work with text subtitles. PGS/VOB would have to remain external or burned.
> 
> Final thought is if an API could be provided for player to directly read subtitles from the server without downloading. I don’t know how feasible that is.

I've put together a server-side fix: instead of awaiting full extraction for text subs, poll the output file and return as soon as `-flush_packets 1` puts data on disk. Extraction continues in the background. Tested on HDD with concurrent transcoding — subtitles in <1 second vs 3+ minutes.

About your suggestion, the retry idea. `fetchSubtitles` in the web player does a single fetch with no retry, so that would need client changes across all players. The server-side early return avoids that.

  Embedding subs in HLS as WebVTT is the proper long-term fix (#172) but needs rework across all clients. 
#16380 (burn-in during transcode) covers the transcode case and is awaiting merge.

PR #16444, opened 2026-03-21
migrate local comic providers to server codebase

dkanada:
**Changes**

This pull request extracts all local metadata providers for comics from the Bookshelf plugin. I only made minimal changes to match our code formatting and improve error handling. It's the last step before we can archive that plugin and replace it with the new Google Books and ComicVine providers.

ComicImageProvider: extracts primary image from any comic file
ComicBookInfoProvider: ComicBookInfo XML embedded in CBZ comment as required by specification
InternalComicInfoProvider: ComicInfo XML embedded in CBZ file
ExternalComicInfoProvider: ComicInfo XML next to any comic file

**Issues**

None

### DIFF
--- a/MediaBrowser.Providers/Books/ComicBookInfo/ComicBookInfoProvider.cs
+++ b/MediaBrowser.Providers/Books/ComicBookInfo/ComicBookInfoProvider.cs
@@ -1,0 +1,246 @@
+using System;
+using System.Globalization;
+using System.IO;
+using System.Linq;
+using System.Text.Json;
+using System.Threading;
+using System.Threading.Tasks;
+using Jellyfin.Data.Enums;
+using Jellyfin.Extensions.Json;
+using MediaBrowser.Controller.Entities;
+using MediaBrowser.Controller.Providers;
+using MediaBrowser.Model.IO;
+using MediaBrowser.Providers.Books.ComicBookInfo.Models;
+using Microsoft.Extensions.Logging;
+using SharpCompress.Archives.Zip;
+
+namespace MediaBrowser.Providers.Books.ComicBookInfo;
+
+/// <summary>
+/// ComicBookInfo provider.
+/// </summary>
+public class ComicBookInfoProvider : IComicProvider
+{
+    private readonly ILogger<ComicBookInfoProvider> _logger;
+    private readonly IFileSystem _fileSystem;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="ComicBookInfoProvider"/> class.
+    /// </summary>
+    /// <param name="fileSystem">Instance of the <see cref="IFileSystem"/> interface.</param>
+    /// <param name="logger">Instance of the <see cref="ILogger{ComicBookInfoProvider}"/> interface.</param>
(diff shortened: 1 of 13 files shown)


carif:
What about the unit tests from [ComicBookInfoProviderTest](https://github.com/jellyfin/jellyfin-plugin-bookshelf/blob/master/tests/Jellyfin.Plugin.Bookshelf.Tests/ComicBookInfoProviderTest.cs) and [ComicInfoXmlUtilitiesTest](https://github.com/jellyfin/jellyfin-plugin-bookshelf/blob/master/tests/Jellyfin.Plugin.Bookshelf.Tests/ComicInfoXmlUtilitiesTest.cs) ?
And why not copy over the commit history of the extracted metadata providers and then commit your changes on top?

dkanada:
I'm trying to split the files into smaller PRs - would have split ComicInfo and ComicBookInfo providers too, but they both relied on IComicProvider so I didn't bother with that one.

The commit history is still visible from the other repository so there's no benefit to merging an orphan branch in the server codebase - none of it would have built. We'll still keep Bookshelf around in jellyfin-archive for posterity.

carif:
> The commit history is still visible from the other repository so there's no benefit to merging an orphan branch in the server codebase - none of it would have built. We'll still keep Bookshelf around in jellyfin-archive for posterity.

Personally I'd disagree. You would need to cross-reference the other repository and at least when viewing the changes with git itself and not with GitHub, the information on who made changes for what reasons is lost. But that might just be my workflow on relying more on the git cli than GitHub.

---

Regardless of that discussion I'd still like to see the unit tests being copied over, too. They provide clear documentation on the expected behavior including edge cases.

MediaBrowser.Providers/Books/ComicInfo/ComicInfoReader.cs
```diff
@@ -0,0 +1,218 @@
...
+            return;
+        }
+
+        try
+        {
+            // culture cannot be null here as the method would have returned earlier
+            commitResult(new CultureInfo(culture!));
+        }
+        catch (CultureNotFoundException)
+        {
```
JPVenson: Add logging

JPVenson (changes requested):

MediaBrowser.Providers/Books/ComicBookInfo/ComicBookInfoProvider.cs
```diff
@@ -0,0 +1,246 @@
...
+                    _logger.LogError("ComicBookInfo deserialization failure: {Path}", info.Path);
+                    return new MetadataResult<Book> { HasMetadata = false };
+                }
+
+                return SaveMetadata(comicBookMetadata);
+            }
+        }
+        catch (Exception)
+        {
+            _logger.LogError("failed to load ComicBookInfo metadata: {Path}", info.Path);
```
IDisposable: Wouldn't it be wise to emit some of the `Exception` information?

IDisposable (approved):

IDisposable (approved):

JPVenson (approved):

PR #16456, opened 2026-03-23
Fix FolderStorageInfo to show parent filesystem

joshuaboniface:
**Changes**

A direct implementation using DriveInfo directly on a path does not work as expected. The method will return a DriveInfo object with the given path as both the Name and the RootDirectory, which is not helpful.

Instead, add parsing logic to find the best possible match out of all filesystems on the system for the path, including handling edge cases involving symlinked paths in the chain.

This ensures that the resulting DeviceId is a valid filesystem, allowing it to be used in the UI to show a better description. It also includes the new ResolvedPath which will show, if required, what the Path resolved to after all symlinks are interpolated.

One possible issue here is that walking all drives as-is might become slow(er) on a system with many partitions, but even on my partition-heavy system with over a dozen ZVOLs and remote mounts, this takes under 0.4 seconds including runup time for `dotnet run`, so I imagine this should be fine.

**Issues**
Addresses https://github.com/jellyfin/jellyfin-web/issues/7323

### DIFF
--- a/Jellyfin.Server.Implementations/StorageHelpers/StorageHelper.cs
+++ b/Jellyfin.Server.Implementations/StorageHelpers/StorageHelper.cs
@@ -43,7 +43,7 @@ public static class StorageHelper
             // This is required because simply calling `DriveInfo` on a path returns that path as
             // the Name and RootDevice, which is not at all how this should work.
             var allDrives = DriveInfo.GetDrives();
-            DriveInfo? bestMatch = null;  
+            DriveInfo? bestMatch = null;
             foreach (DriveInfo d in allDrives)
             {
                 if (resolvedPath.StartsWith(d.RootDirectory.FullName, StringComparison.InvariantCultureIgnoreCase) &&
@@ -52,11 +52,12 @@ public static class StorageHelper
                     bestMatch = d;
                 }
             }
-            
-            if (bestMatch is null) {
+
+            if (bestMatch is null)
+            {
                 throw new InvalidOperationException($"The path `{path}` has no matching parent device. Space check invalid.");
             }
-            
+
             return new FolderStorageInfo()
             {
                 Path = path,


Jellyfin.Server.Implementations/StorageHelpers/StorageHelper.cs
```diff
@@ -28,29 +28,49 @@ public static void TestCommonPathsForStorageCapacity(IApplicationPaths applicati
...
         {
-            var driveInfo = new DriveInfo(path);
+            // Fully resolve the given path to an actual filesystem target, in case it's a symlink or similar.
+            string resolvedPath = ResolvePath(path);
+            // We iterate all filesystems reported by GetDrives() here, and attempt to find the best
+            // match that contains, as deep as possible, the given path.
+            // This is required because simply calling `DriveInfo` on a path returns that path as
+            // the Name and RootDevice, which is not at all how this should work.
+            DriveInfo[] allDrives = DriveInfo.GetDrives();
+            DriveInfo bestMatch = null;
```
JPVenson:
```suggestion
            DriveInfo? bestMatch = null;  
```

Jellyfin.Server.Implementations/StorageHelpers/StorageHelper.cs
```diff
@@ -28,29 +28,49 @@ public static void TestCommonPathsForStorageCapacity(IApplicationPaths applicati
...
         try
         {
-            var driveInfo = new DriveInfo(path);
+            // Fully resolve the given path to an actual filesystem target, in case it's a symlink or similar.
+            string resolvedPath = ResolvePath(path);
+            // We iterate all filesystems reported by GetDrives() here, and attempt to find the best
+            // match that contains, as deep as possible, the given path.
+            // This is required because simply calling `DriveInfo` on a path returns that path as
+            // the Name and RootDevice, which is not at all how this should work.
+            DriveInfo[] allDrives = DriveInfo.GetDrives();
```
JPVenson:
```suggestion
            var allDrives = DriveInfo.GetDrives();
```

Jellyfin.Server.Implementations/StorageHelpers/StorageHelper.cs
```diff
@@ -28,29 +28,49 @@ public static void TestCommonPathsForStorageCapacity(IApplicationPaths applicati
...
     /// <param name="path">Path to a folder.</param>
-    /// <returns>The number of bytes available space.</returns>
+    /// <returns>Various details about the parent filesystem containing the directory.</returns>
     public static FolderStorageInfo GetFreeSpaceOf(string path)
     {
         try
         {
-            var driveInfo = new DriveInfo(path);
+            // Fully resolve the given path to an actual filesystem target, in case it's a symlink or similar.
+            string resolvedPath = ResolvePath(path);
```
JPVenson:
```suggestion
            var resolvedPath = ResolvePath(path);
```

Jellyfin.Server.Implementations/StorageHelpers/StorageHelper.cs
```diff
@@ -28,29 +28,49 @@ public static void TestCommonPathsForStorageCapacity(IApplicationPaths applicati
...
+            // We iterate all filesystems reported by GetDrives() here, and attempt to find the best
+            // match that contains, as deep as possible, the given path.
+            // This is required because simply calling `DriveInfo` on a path returns that path as
+            // the Name and RootDevice, which is not at all how this should work.
+            DriveInfo[] allDrives = DriveInfo.GetDrives();
+            DriveInfo bestMatch = null;  
+            foreach (DriveInfo d in allDrives)
+            {
+                if (resolvedPath.StartsWith(d.RootDirectory.FullName) &&
```
JPVenson:
```suggestion
                if (resolvedPath.StartsWith(d.RootDirectory.FullName, StringComparison.InvariantCultureIgnoreCase) &&
```

Jellyfin.Server.Implementations/StorageHelpers/StorageHelper.cs
```diff
@@ -28,29 +28,49 @@ public static void TestCommonPathsForStorageCapacity(IApplicationPaths applicati
...
+                    (bestMatch == null || d.RootDirectory.FullName.Length > bestMatch.RootDirectory.FullName.Length))
```
JPVenson:
```suggestion
                    (bestMatch is null || d.RootDirectory.FullName.Length > bestMatch.RootDirectory.FullName.Length))
```

Jellyfin.Server.Implementations/StorageHelpers/StorageHelper.cs
```diff
@@ -28,29 +28,49 @@ public static void TestCommonPathsForStorageCapacity(IApplicationPaths applicati
...
+                {
+                    bestMatch = d;
+                }
+            }
+            if (bestMatch is null) {
+                throw new InvalidOperationException($"The path `{path}` has no matching parent device. Space check invalid.");
+            }
             return new FolderStorageInfo()
```
JPVenson, line 61:
```suggestion
            }
            
            if (bestMatch is null) {
                throw new InvalidOperationException($"The path `{path}` has no matching parent device. Space check invalid.");
            }
            
            return new FolderStorageInfo()
```

PR #16463, opened 2026-03-25
split openapi workflows between pull request and merge

dkanada:
**Changes**

Initial attempt to move sensitive jobs into a separate workflow for secret usage. I'll likely need to create a few more PRs this week as issues arise since I can't test this easily without a merge.

**Issues**

None

### DIFF
--- a/.github/workflows/openapi/__generate.yml
+++ b/.github/workflows/openapi/__generate.yml
@@ -1,0 +1,44 @@
+name: OpenAPI Generate
+
+on:
+  workflow_call:
+    inputs:
+      ref:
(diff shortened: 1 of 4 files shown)


nielsvanvelzen:
> The workflow is not valid. .github/workflows/openapi-merge.yml (Line: 12, Col: 3): Error calling workflow 'jellyfin/jellyfin/.github/workflows/openapi-generate.yml@c7ac591548bcd89158fbc4c002b42f1dba701400'. The workflow is requesting 'contents: read', but is only allowed 'contents: none'.

PR #16465, opened 2026-03-25
recognize underscore and dot separators for multi-version grouping

upscaylman:
So currently if you have your files named like `Movie_4K.mkv` and `Movie_1080p.mkv` in the same folder, jellyfin doesn't group them as versions. It only recognizes the dash `-` as a separator between the movie name and the version tag.

But a lot of people (myself included) use underscores or dots in their filenames, its pretty common in release groups naming. This just adds `_` and `.` to the check in `IsEligibleForMultiVersion`.

Had to update TestMultiVersion8 because the test data actually uses underscore-separated files that were previously expected NOT to group — now they do, which is the correct behavior imo. Also added 2 new tests specifically for underscore and dot separators.

All naming tests pass (622 tests).

### DIFF
--- a/tests/Jellyfin.Naming.Tests/Video/MultiVersionTests.cs
+++ b/tests/Jellyfin.Naming.Tests/Video/MultiVersionTests.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using Emby.Naming.Common;
@@ -254,7 +255,6 @@ namespace Jellyfin.Naming.Tests.Video
         [Fact]
         public void TestMultiVersion8()
         {
-            // With underscore separator support, these files are now grouped as alternate versions
             var files = new[]
             {
                 "/movies/Iron Man/Iron Man.mkv",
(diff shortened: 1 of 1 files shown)


tests/Jellyfin.Naming.Tests/Video/MultiVersionTests.cs
```diff
@@ -254,6 +254,7 @@ public void TestMultiVersion7()
         [Fact]
         public void TestMultiVersion8()
         {
+            // With underscore separator support, these files are now grouped as alternate versions
```
Shadowghost: Unnecessary comment.
I wonder if with these changes, the 3d type recognition still works. I'd appreciate it if you could check that.

upscaylman: Removed the comment and checked — 3D recognition works fine with underscore separators since `_` is already in the VideoFlagDelimiters. Added an assertion in the test to verify `_3d-hsbs` is correctly detected as Is3D with Format3D="hsbs".

PR #16476, opened 2026-03-26
guard against empty alternate versions in MovieResolver

upscaylman:
When a video has no alternate versions, video.AlternateVersions.Select(...).ToArray() still allocates an unnecessary array. This adds a simple count check and returns Array.Empty<string>() instead, which is consistent with how other arrays are handled in the codebase (AdditionalParts etc).

Small thing I noticed while looking at the resolver code.


Shadowghost:
@upscaylman can you check if this PR is still needed after the latest changes to the alternative version logic?

Bond-009 (approved):

PR #16477, opened 2026-03-26
Use native contains

### LINKED ISSUE #16344: Do not use custom Contains extension for spans
Hello!
I raised this issue here two years ago, please check it:
https://github.com/jellyfin/jellyfin/pull/11188

I decided to close this PR after [AlternateLookup](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.hashset-1.alternatelookup-1?view=net-10.0) was released.
To my surprise, after two years, this extension now has even more references than before (100+). It is now used even with regular strings, thanks to the implicit `string` => `ReadonlySpan<char>` cast.

My point is still the same: this extension makes all Sets effectively useless. I believe the people who added Sets to the codebase intended the search complexity to be closer to O(1) rather than O(n).
Moreover, it is very confusing because it has the exact same name as the native implementation.

I suggest the following plan:
1. Rename this extension to make it clear that it is custom, for example, `ContainsSpan`.
2. Use native overload (in `System.MemoryExtensions`) and LINQ (for regular strings) wherever possible.
3. Use `AlternateLookup` For HashSet, FrozenSet, Dictionary, etc.

Would you like me to start working on this?

Shadowghost:
**Changes**
* Replace our custom `Contains` extension for spans with the native one from LINQ/MemoryExtensions for strings
* Use FrozenSet where possible

**Issues**
Fixes #16344


Bond-009:
I'm not sure if Sets make a difference with such small collection sizes and with the added penalty of string allocations.
I'd like to see some benchmarks before deciding on this.

PR #16487, opened 2026-03-27
Add server-side playback speed support via ffmpeg filters

VictorJulianiR:
On my Samsung TV  (Tizen 1.1.0) setting playbackRate causes audio to be muted at non-1x speeds. This was a issue for me but also mentioned in a couple of issues, links below.

There are 2 PR's one for server and one for web-client. 

I just wanted to change so I could have a fork that will run on my TV. I tested it on my TV at 1.5x2x and 3x speeds. Audio mostly works but sometimes TV can't handle the buffer, even with changes on the StreamState segment buffer size. 

Two small disclaimers, I didn't boot up the developer TV app, but I did send an API request to the server to force transcoding using the atempo algorithm, so I know the algorithm works, but it needs transcoding, so adding it without a setting might not be the best approach here. Curious to hear feedback from you guys. 

Second disclaimer, I did use copilot (code assistance) to code the changes. Since I wouldn't be able to understand the entire structure of the repo by myself for just this quick fix. If any of the PR reviewers that are more familiar with the repo would like to suggest a better way of doing it, I'm all ears. Otherwise, all changes looking good and non-intrusive, and that's what I seek when changing someone's legacy code. I've read the LLM development policy.

Code was tested and project is building and executing. 

### Changes
BaseEncodingJobOptions: New PlaybackSpeed property (nullable double)

EncodingHelper: Generates atempo filter chains for audio (handles 0.5x–3.0x by chaining filters) and setpts=PTS/{speed} for video. Atempo is the most reliable open source algorithm that does this that I could find, and it was avaiable on ffmpeg that's what jellyfin used for transcoding. Probably there could be better algorithms out there, but this is fine enough.

DynamicHlsController: Added playbackSpeed query parameter to all 7 HLS endpoints

StreamState: HLS segment length scales proportionally with playback speed to prevent buffer underruns at higher speeds

### Issues
ex. Fixes jellyfin-web/issues/3314 and jellyfin-webos/issues/96

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -2793,6 +2793,28 @@ namespace MediaBrowser.Controller.MediaEncoding
             return null;
         }
 
+        private static void AddAtempoFilters(List<string> filters, double speed)
+        {
+            // atempo supports 0.5-2.0 per instance; chain for values outside that range
+            var remaining = speed;
+            while (remaining > 2.0)
+            {
+                filters.Add("atempo=2.0");
+                remaining /= 2.0;
+            }
+
+            while (remaining < 0.5)
+            {
+                filters.Add("atempo=0.5");
+                remaining /= 0.5;
+            }
+
+            if (Math.Abs(remaining - 1.0) > 0.001)
+            {
+                filters.Add(string.Format(CultureInfo.InvariantCulture, "atempo={0:F4}", remaining));
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


gnattu:
I think this approach is too heavy for this use case as a lot of hardware may not even keep up with multiplied playback speed for transcoding.

VictorJulianiR:
> I think this approach is too heavy for this use case as a lot of hardware may not even keep up with multiplied playback speed for transcoding.

yes, i could easily handle in a small notebook, however since it's using transcoding i think it should be an option, not the standard behavior...

i guess for the computers that can handle it it's better to have sound then not?

PR #16488, opened 2026-03-27
Add Accept-Language header support and cleanup translations

Shadowghost:
Implements https://github.com/jellyfin/jellyfin-meta/discussions/107

**Changes**
* Add Accept-Language header support for per-request localization
* Use server culture for persisted activity log strings
* Automate GetLocalizationOptions from embedded resources
* Remove unused translation keys


nielsvanvelzen:
Findings from my initial testing:

- An "Accept-Language" header can include multiple languages (with qualities assigned to them), preferably the server uses all of those to infer the fallback languages (e.g. if the request says german -> dutch -> english and we only have a dutch or english translation for a particular string it should use that)
- The "Content-Language" only returns the primary selected language but the server already does a fallback to english (or server default?) so that should be included at minimum in the response header. If my previous point is implemented it should include all known cultures in fallback order.
- Looks like this does not work for WebSockets yet

Otherwise this seems promising from an API user perspective, tested by listing scheduled tasks and receiving translated responses based on the Accept-Language header.

nielsvanvelzen:
For the WebSockets I'm testing with schedules tasks (manually starting one so it sends an update) but they're still in English even when I send an `Accept-Language: nl` header. So that's still not working completely. Otherwise this seems to be working fine.

nielsvanvelzen (approved):
Seems to be working as expected during testing (both HTTP API and WebSockets).

Jellyfin.Server/Middleware/AcceptLanguageMiddleware.cs
```diff
@@ -0,0 +1,137 @@
...
+
+namespace Jellyfin.Server.Middleware;
+
+/// <summary>
+/// Middleware that resolves the <c>Accept-Language</c> request header
+/// to an ordered list of Jellyfin-supported cultures, sets the fallback chain
+/// on <see cref="LocalizationManager.RequestCultureFallback"/>, and writes
+/// the <c>Content-Language</c> response header.
+/// </summary>
+public class AcceptLanguageMiddleware
```
JPVenson: why do we need to reinvent the wheel here if aspcore has that baked in?
https://learn.microsoft.com/en-us/aspnet/core/fundamentals/localization/select-language-culture?view=aspnetcore-10.0

JPVenson (changes requested):

Emby.Server.Implementations/HttpServer/WebSocketManager.cs
```diff
@@ -48,13 +49,21 @@ public async Task WebSocketRequestHandler(HttpContext context)
 
                 WebSocket webSocket = await context.WebSockets.AcceptWebSocketAsync().ConfigureAwait(false);
 
+                // Capture the culture set by RequestLocalizationMiddleware so it can be
```
JPVenson: comment outdated

Emby.Server.Implementations/HttpServer/WebSocketConnection.cs
```diff
@@ -81,6 +87,24 @@ public WebSocketConnection(
...
+        public void ApplyRequestCulture()
+        {
+            if (string.IsNullOrEmpty(RequestUICulture))
+            {
+                return;
+            }
+
+            try
+            {
+                CultureInfo.CurrentUICulture = CultureInfo.GetCultureInfo(RequestUICulture);
```
JPVenson: i could be wrong but allowing untrusted input into the GetCultureInfo was a dangerous operation. I need to look it up though, how against are you to validating against our known list of cultures?

Shadowghost: The value here actually comes from `CultureInfo.CurrentUICulture` after `RequestLocalizationMiddleware` has already constrained it to `LocalizationManager.GetSupportedUICultures()`, so it's not untrusted at this point.

Emby.Server.Implementations/Localization/LocalizationManager.cs
```diff
@@ -54,6 +67,45 @@ public LocalizationManager(
...
+                {
+                    cultures.Add(CultureInfo.GetCultureInfo(code));
+                }
+                catch (CultureNotFoundException)
+                {
+                    // Skip novelty codes (e.g. "pr" Pirate, "jbo" Lojban) that .NET cannot resolve.
+                }
+            }
+
+            return cultures;
```
IDisposable, line 87: Seems like this method builds a complete translation dictionary from JellyFin `_` to .Net `-` but we use a [much smaller dictionary](https://github.com/jellyfin/jellyfin/pull/16488/changes#diff-9e2bb0ee20e6e729f6834662e38326451b01ab8d6eb10c698af942f8c1b0671cR34-R41) the other direction... couldn't/shouldn't we be building the reflexive dictionary at the same time to ensure they're complete?

nielsvanvelzen (changes requested):
With the changes after my last review this no longer works for the HTTP API (websockets do work).

nielsvanvelzen:
Still broken

Shadowghost:
@nielsvanvelzen  what are you sending and what are you expecting to be returned?

nielsvanvelzen:
I'm sending an Accept-Language header in various formats and I always receive the server set language back

edit: ignore the above - I might have had an issue in the testing setup

nielsvanvelzen (approved):
Turns out having working test code helps

JPVenson:
patrick-stewart-my-god

JPVenson (approved):

Emby.Server.Implementations/Localization/LocalizationManager.cs
Shadowghost: Good point, fixed it.

Emby.Server.Implementations/Localization/LocalizationManager.cs
```diff
@@ -508,77 +569,65 @@ private static string GetResourceFilename(string culture)
...
+            options.Sort((a, b) => string.Compare(a.Name, b.Name, StringComparison.OrdinalIgnoreCase));
+            return (options, bcp47Map.ToFrozenDictionary(StringComparer.OrdinalIgnoreCase));
+        }
+
+        private static string GetDisplayName(string cultureCode)
+        {
+            // Resource files use underscores for codes that .NET's CultureInfo doesn't accept directly (e.g. es_419).
+            var lookup = cultureCode.Contains('_', StringComparison.Ordinal)
+                ? cultureCode.Replace('_', '-')
+                : cultureCode;
```
Bond-009:
```suggestion
            var lookup = cultureCode.Replace('_', '-');
```

Bond-009: This replace + GetCultureInfo in a try catch gets used twice in this file, maybe worth it to extract it to a function

Bond-009 (changes requested):

Emby.Server.Implementations/Localization/Core/en-US.json
```diff
@@ -121,8 +93,8 @@
     "TaskUpdatePluginsDescription": "Downloads and installs updates for plugins that are configured to update automatically.",
     "TaskCleanTranscode": "Clean Transcode Directory",
     "TaskCleanTranscodeDescription": "Deletes transcode files more than one day old.",
-    "TaskRefreshChannels": "Refresh Channels",
-    "TaskRefreshChannelsDescription": "Refreshes internet channel information.",
+    "TasksRefreshChannels": "Refresh Channels",
+    "TasksRefreshChannelsDescription": "Refreshes internet channel information.",
```
crobibero: Was it intentional to change these keys?
If the code differed from what was in the localization I think we should just update the code so we don't need to run this change through weblate

Shadowghost: Likely a rebasing issue. Will take a look later.

Bond-009 (approved):

crobibero (approved):

PR #16494, opened 2026-03-27
Add external item provider infrastructure

IDisposable:
# Rationale

While there are already some external projects like [JellySwarrm](https://github.com/LLukas22/Jellyswarrm/) that allow connecting two JellyFin server instances, I'm looking for a different experience.

The basic idea is building a PlugIn that is given API Key credentials and URL of a remote JellyFin server instance (called _remote_ herein). It talks to that _remote_ instance and replicates the contents of specified libraries into the _local_ instance _as-if_ they were present in the _local_ server. 

We do NOT want local-file copies of the _remote_'s library because it might be a multi-terabyte archive that should be treated as a read-only resource.

We don't want to just expose the _remote_'s play-path for those files because that would require the eventual client of the _local_ server, because that would require users and passwords on the _remote_ to be communicated to the _local_'s client.

## Design

```text
  [Local Consumer Server]                    [Remote Provider Server]
  ┌───────────────────────────┐              ┌────────────────────────┐
  │ Plugin: RemoteLibrarySync │──── API ────►│ Plugin: RemoteAccess   │
  │  IExternalItemProvider    │              │  Token validation      │
  │  IMediaSourceProvider     │              │  /RemoteLibrary/Stream │
  │  IStreamRedirectProvider  │              │  HMAC-signed tokens    │
  └───────────────────────────┘              └────────────────────────┘
         │                                           ▲
         │ Items in local DB as first-class          │ HTTP 302 redirect
         │ citizens (SourceType.External)            │ with time-signed token
         ▼                                           │
  ┌───────────────────────────┐                      │
  │ Client (Jellyfin app)     │──────────────────────┘
  │  Sees normal library      │  Direct HTTP to provider
  │  /Videos/{localId}/stream │  after redirect
  └───────────────────────────┘

```

## Coming Soon

Added the initial beta of the plugin pair I'm building. https://github.com/IDisposable/jellyfin-plugin-remote-library

After that, probably going to build out a second one, etc...

It occurs that having done this work, we could split the _current_ file-system driven scanning and such out as-if it were a distinct source-provider and really isolate the library implementation from the changes in the scanning/metadata slurp... but that remains an idea for a MUCH later iteration :)

## Changes

This requires some establishing work on _local_ server's JellyFin server, so it knows how to track the remote content, and understands when/how to serve it up, and *not* to do metadata scans (much like we already don't for Live TV) content.

- Add `SourceType.External` for items backed by a remote provider; derived from a new `ExternalProviderId` column on `BaseItemEntity` so no extra storage field is needed
- Add `IExternalItemProvider` plugin interface; the library scan calls registered providers, upserts returned `ExternalItemInfo` items, and removes items no longer present
- Add `ExternalItemInfo` carrying provider-supplied metadata: name, year, genres, provider IDs, image URLs, media streams, and series/season linking
- Add `IStreamRedirectProvider` plugin interface with an `Order` property; at stream time, registered providers can return a redirect URL for an item; `HttpStreamRedirectProvider` ships as a built-in `int.MaxValue`-ordered fallback that redirects to the item's HTTP path so plugin providers can take priority
- Redirect external items (HTTP 302) in `AudioHelper`, `VideosController`, and `DynamicHlsController` before entering the transcoding pipeline
- Guard `SupportsLocalMetadata`, `ContainingFolderPath`, `GetLocalMetadataFilesToDelete`, and the playback-probe trigger in `MediaSourceManager` against non-file-protocol items
- Add `IExternalItemProvider.IsAvailableAsync` (default: `true`) so a temporarily-offline
  provider skips sync without removing existing items from the library
- Add `StreamToken` utility in `MediaBrowser.Common` — self-contained HMAC-SHA256
  time-limited tokens for redirect URL signing; format:
  `{base64url("{itemId}|{expiresUnixSeconds}")}.{hex(HMAC-SHA256(secret, payload))}`;
  single `token` query parameter; supports `Generate`, `TryParse`, and `TryValidate`
- Add `StreamRedirectResult` record carrying `RedirectUrl` and an optional `PingForwardUrl`
  (reserved for future provider-side transcoding liveness forwarding)
- Update `IStreamRedirectProvider.GetRedirectAsync` to return `StreamRedirectResult?`
  so providers can opt in to ping forwarding when implemented

---
<a href="https://gitkraken.dev/review/github/jellyfin/jellyfin/pull/16494?source=pr_review_chip">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://gitkraken.dev/images/figures/gitkraken-review-badge-dark.svg">
    Open with GitKraken
  </picture>
</a>


Emby.Server.Implementations/Library/HttpStreamRedirectProvider.cs
```diff
@@ -0,0 +1,39 @@
...
+using MediaBrowser.Controller.Entities;
+using MediaBrowser.Controller.Library;
+
+namespace Emby.Server.Implementations.Library;
+
+/// <summary>
+/// Redirects external items with HTTP media paths directly to their remote URL.
+/// Runs last so plugin providers can take priority.
+/// </summary>
+public class HttpStreamRedirectProvider : IStreamRedirectProvider
```
IDisposable, line 13: I am not sure if we'll ever need other `IStreamRedirectProvider` implementations, but this is the project pattern

Emby.Server.Implementations/Library/LibraryManager.cs
```diff
@@ -1196,20 +1217,179 @@ private async Task PerformLibraryValidation(IProgress<double> progress, Cancella
 
             await ValidateTopLibraryFolders(cancellationToken).ConfigureAwait(false);
 
-            var innerProgress = new Progress<double>(pct => progress.Report(pct * 0.96));
+            // Budget: 2% per external provider, 4% for post-scan tasks, remainder for filesystem scan.
```
IDisposable, line 1520: Going to eventually revisit this logic... even in the current version, things stall for longish times while processing the `PostScanTasks`, but it's going to be potentially a much longer process here.

Emby.Server.Implementations/Library/LibraryManager.cs
```diff
@@ -1196,20 +1217,179 @@ private async Task PerformLibraryValidation(IProgress<double> progress, Cancella
...
+            }
+
+            return null;
+        }
+
+        private BaseItem? CreateExternalItem(ExternalItemInfo info, string providerId)
+        {
+            BaseItem? item = info.ItemKind switch
+            {
+                BaseItemKind.Movie => new Movie(),
```
IDisposable, line 1656: Not sure if there's more types we need to support, or even if this factory-like logic exists somewhere else or could be shared.

MediaBrowser.Model/Library/ExternalItemInfo.cs
```diff
@@ -0,0 +1,111 @@
+using System.Collections.Generic;
+using Jellyfin.Data.Enums;
+using MediaBrowser.Model.Entities;
+
+namespace MediaBrowser.Model.Library;
+
+/// <summary>
+/// Describes a single item provided by an external item provider.
+/// </summary>
+public class ExternalItemInfo
```
IDisposable, line 11: This essentially establishes the contract of what metadata we will support coming from external sources. Would love comments on what else should be allowed,  but this seems like the baseline for the support we need.

**NOTE**: This means that any external source provider could offer up this same data, so we could have an external provider plugin build for pretty much any other media source. It would be plausible to even build ones to consume the content stored on a ChannelsDVR or Plex or Kodi server, or content on S3, for example.

src/Jellyfin.Database/Jellyfin.Database.Implementations/Entities/BaseItemEntity.cs
```diff
@@ -114,6 +114,8 @@ public class BaseItemEntity
 
     public string? ExternalSeriesId { get; set; }
 
+    public string? ExternalProviderId { get; set; }
```
IDisposable, line 119: I don't think we should track the external SOURCE in a by-item basis. If an plug-in wants to do that, it can certainly use a compound key string for the ID.

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/Migrations/20260327000000_AddExternalProviderId.Designer.cs
```diff
@@ -0,0 +1,1724 @@
...
+                    b.HasIndex("MediaType", "TopParentId", "IsVirtualItem", "PresentationUniqueKey");
+
+                    b.HasIndex("Type", "SeriesPresentationUniqueKey", "IsFolder", "IsVirtualItem");
+
+                    b.HasIndex("Type", "SeriesPresentationUniqueKey", "PresentationUniqueKey", "SortName");
+
+                    b.HasIndex("IsFolder", "TopParentId", "IsVirtualItem", "PresentationUniqueKey", "DateCreated");
+
+                    b.HasIndex("Type", "TopParentId", "IsVirtualItem", "PresentationUniqueKey", "DateCreated");
+
```
IDisposable, line 397: Probably would behoove us to have an index here. Might revisit as I build out the plugins

jumoog:
first step for S3 🥳

IDisposable:
> first step for S3 🥳

Indeed!

That'll be my 2nd plugin (or OneDrive or Dropbox or whateves)

IDisposable:
Added the initial beta of the plugin pair I'm building. 

Still testing in progress.

https://github.com/IDisposable/jellyfin-plugin-remote-library

nielsvanvelzen:
Huge changes like this should first be discussed with the Jellyfin team if the feature is even wanted and then a proposal should be drafted in our [meta repository](https://github.com/jellyfin/jellyfin-meta) explaining how it will technically be implemented **before** opening pull request(s).

IDisposable:
I certainly can start up a conversation there. I personally don't see this as a huge change, as it is opaque outside the server itself.

IDisposable:
Started the meta ~issue at [jellyfin-meta](https://github.com/jellyfin/jellyfin-meta/issues/117).~ discussion https://github.com/jellyfin/jellyfin-meta/discussions/118

Going to switch this PR to Draft, even though it is essentially complete pending discussion over there.

~EDIT: No sure what happened, I clicked convert to draft and it suddenly is closed :(~

JPVenson:
Converted it to a draft and converted the issue ticket you created to a Discussion

IDisposable:
Rebase will happen if the discussion resolved to approved.

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/Migrations/20260327000000_AddExternalProviderId.Designer.cs
```diff
@@ -0,0 +1,1724 @@
...
+    [Migration("20260327000000_AddExternalProviderId")]
+    partial class AddExternalProviderId
+    {
+        /// <inheritdoc />
+        protected override void BuildTargetModel(ModelBuilder modelBuilder)
+        {
+#pragma warning disable 612, 618
+            modelBuilder.HasAnnotation("ProductVersion", "9.0.9");
+
+            modelBuilder.Entity("Jellyfin.Database.Implementations.Entities.AccessSchedule", b =>
```
IDisposable, line 23: This file seems to be completely wonky since rebasing to current master. I suspect that was the recent "catch up" PR for designer drift. If this proceeds beyond the meta discussion, I'll fix this.

PR #16499, opened 2026-03-28
add transcode bitrate limit

trumanmiller:
**Changes**

add logic to rerun video stream calculations with a lower MaxBitrate if the reason for transcode would require intensive transcoding and bitrate is over TranscodeBitrateLimit

playback info with limit set to 5mbps

interlaced video, limit is applied:
image

direct play, bitrate limit is not applied:
image

jellyfin-web PR: https://github.com/jellyfin/jellyfin-web/pull/7775

**Feature Requests**

adds feature requested by:
https://features.jellyfin.org/posts/2596/separate-bitrate-limit-for-direct-stream-and-transcoding
https://features.jellyfin.org/posts/3763/server-side-maximum-resolution-cap-for-transcoded-video

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -160,6 +160,7 @@
  - [ThreeFive-O](https://github.com/ThreeFive-O)
  - [tjwalkr3](https://github.com/tjwalkr3)
  - [TrisMcC](https://github.com/TrisMcC)
+ - [Truman Miller](https://github.com/trumanmiller)
  - [trumblejoe](https://github.com/trumblejoe)
  - [TtheCreator](https://github.com/TtheCreator)
  - [twinkybot](https://github.com/twinkybot)
--- a/Jellyfin.Api/Helpers/MediaInfoHelper.cs
+++ b/Jellyfin.Api/Helpers/MediaInfoHelper.cs
@@ -12,6 +12,7 @@ using Jellyfin.Data.Enums;
 using Jellyfin.Database.Implementations.Entities;
 using Jellyfin.Database.Implementations.Enums;
 using Jellyfin.Extensions;
+using MediaBrowser.Common.Configuration;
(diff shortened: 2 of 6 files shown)


JPVenson:
that is not what the bitrate limit is build for. Most people do not have fast internet connections so the bitrate limit ensures media can be played in any case. A direct play with 25mbit over a normal DSL connection which a lot of people still have would not work.

trumanmiller:
@JPVenson I think you're referring to the Internet Streaming bitrate limit?

this is separate from that and only applied to transcodes. as some people have fast connections so have no problem serving high bitrate directly but dont have hardware capable of transcoding at those bitrates.

other people have the same issue as detailed here: https://features.jellyfin.org/posts/2596/separate-bitrate-limit-for-direct-stream-and-transcoding

you can see how the config would look on the frontend on the jellyfin-web PR: https://github.com/jellyfin/jellyfin-web/pull/7775

Thedustbustr:
Hi! What is the current status of this pr? I’m curious because I could really benefit from this feature as my hardware can struggle to keep up with high-bitrate transcodes but I have plenty of bandwidth for direct-play. Is there anything needed on the testing side to help move this forward?

trumanmiller:
hey @Thedustbustr,

I just got busy with work and havent had time to push this more. 

i understand hesitance with the re-calculate design, but after several evenings of exploring alternatives i finally settled on this design, and given the chicken-and-egg nature of the problem i think its satisfactory. 

as far as testing, the biggest thing is whether or not the IntensiveTranscodeReasons composite enum contains all the intensive reasons and no direct stream reasons. 

CC @Shadowghost, @JPVenson

ddemarco5:
I'd love to get some more eyes on this. I think it's a nice terse way to implement a separate transcoding bitrate limit.

Issue #15898, opened 2025-12-30
AutoDiscovery returns loopback address when called from a different subnet


ch6574:
### Description of the bug

When Jellyfin receives a `Who is JellyfinServer?` broadcast from a different ipv4 subnet it will reply with the ipv6 loopback address `::1`

The second example in the reproduction below has a UDP broadcast packet sent with source 192.168.2.1 and destination 192.168.1.255, and Jellyfin replies with a unicast packet with source 192.168.1.2 and destination 192.168.1.1, so it knows how to reply to the discovery request, but the payload is unexpected.

### Reproduction steps

Two subnets, 192.168.1.x/24 and 192.168.2.x/24 for ipv4, with a UDP broadcast relay running across them.

Jellyfin itself is bound to all interfaces on the host (i.e. 0.0.0.0:8920 and ::8920, ditto for 8096) 

```
test-192.168.1.1 $ echo -n 'Who is JellyfinServer?' | socat - udp-datagram:255.255.255.255:7359,broadcast
Address":"https://192.168.1.3:8920","Id":"123456...","Name":"jellyfin","EndpointAddress":null}

test-192.168.2.1 $ echo -n 'Who is JellyfinServer?' | socat - udp-datagram:255.255.255.255:7359,broadcast
Address":"https://[::1]:8920","Id":"123456...","Name":"jellyfin","EndpointAddress":null}
```

### What is the current _bug_ behavior?

I don't expect `::1` as a valid reply.

### What is the expected _correct_ behavior?

This is a good question. I'd expect the address in the payload matches the ipv4 address it comes from (e.g. 192.168.1.3 in the example).

I'm guessing there's code that tries to match the interface and it is getting confused due to the subnet mask mismatch, and so defaulting to one of the loopback devices (but this is just a guess).

### Jellyfin Server version

10.11.5

### Specify the build version

10.11.5

### Environment

```markdown
- OS: Debian 13
- Linux Kernel: 6.17.4
- Virtualization: Yes (LXC container)
- Clients: all
- Browser: n/a
- FFmpeg Version: n/a
- Playback Method: n/a
- Hardware Acceleration: n/a
- CPU Model: n/a
- GPU Model: n/a
- Plugins: n/a
- Reverse Proxy: no
- Base URL: see example
- Networking: see example
- Jellyfin Data Storage & Filesystem: n/a
- Media Storage & Filesystem: n/a
- External Integrations: n/a
```

### Jellyfin logs

```shell
[...]

Issue #11564, opened 2024-05-12
AutoDiscoveryHost still returns wrong address


selfhoster1312:
This follows #10005 being closed.

The automatic server discovery code is still broken after upgrade to Jellyfin v10.9.

```
$ echo -n 'Who is JellyfinServer?' | socat - udp-datagram:255.255.255.255:7359,broadcast
{"Address":"http://127.0.0.1:8095/media","Id":"REDACTED","Name":"REDACTED","EndpointAddress":null}
```

The jellyfin server is behind a reverse proxy, and listens only to localhost for security reasons. I'm not sure which setting between `KnownProxies` and `PublishedServerUriySubnet` should be used here, but returning 127.0.0.1 is wrong.

The jellyfin network config is aware of the reverse proxy, as i set in /etc/jellyfin/network.xml:

```
  <KnownProxies>
    <string>https://REDACTED/media</string>
  </KnownProxies>
  <PublishedServerUriBySubnet>
    <string>https://REDACTED/media</string>
  </PublishedServerUriBySubnet>
```

I can understand the need for people to return different URL depending on the subnet querying the server, but for me a single URL is ok.

This is blocking https://github.com/jellyfin/jellyfin-media-player/issues/168 for example. It's very annoying to onboard users to have to type the complete URL.

gnattu:
If you really want to override the announced url, you need to write the full url and set it as `JELLYFIN_PublishedServerUrl` environment variable, only this env variable will be picked up by auto discovery.

selfhoster1312:
Thanks for the quick answer. So from what i can read there is also a [CLI flag](https://github.com/jellyfin/jellyfin/blob/25c23af8659700909995ca9c91c5dafccc6d348b/Jellyfin.Server/StartupOptions.cs#L68). But there is no network.xml setting for this yet?

gnattu:
We usually don't put such super advanced config into xml, and those hidden options have no documentation because we don't expect most users to ever alter them.

Shadowghost:
Just as an FYI: what you have in your config XML is wrong.
The proxy stanza should include the upstream proxy IPs and the override should be `subnet=override`, e.g. `127.0.01/8=https://domain.tld`

If you have local subnets and proxy properly configured you can also use external, internal and all instead of a subnet.

selfhoster1312:
Thanks for your time! I'm gonna close the issue because the workaround is "documented" here.

[...]

PR #16515, opened 2026-03-30
Fix auto-discovery returning loopback address when PublishedServerUrl is set

baykatre:
I was running into this issue on my own setup — my Jellyfin server sits behind a reverse proxy and binds to localhost, so `PublishedServerUrl` was set to `http://localhost:8096`. When I tried discovering the server from my TV on the same LAN, the discovery response came back with `localhost`, which obviously doesn't work from another device.

Dug into it and found that `RespondToV2Message` calls `GetSmartApiUrl`, which returns `PublishedServerUrl` as-is when it's set — no check for whether it actually makes sense for LAN discovery. Same root cause as #15898, where cross-subnet discovery returns `::1`.

The fix is simple: after getting the result from `GetSmartApiUrl`, check if it's a loopback address. If it is, resolve the real interface IP with `GetBindAddress(skipOverrides: true)` and build the URL with `GetLocalApiUrl`. If it's not loopback (custom domain, specific IP, whatever), leave it alone.

To reproduce:
1. Set `PublishedServerUrl` to `http://localhost:8096`
2. From another machine: `echo "who is JellyfinServer?" | socat - udp-datagram:255.255.255.255:7359,broadcast`
3. You'll get `localhost` back — can't connect

After the fix, you get the actual IP.

Addresses #15898
Addresses #11564

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -11,6 +11,7 @@
  - [AThomsen](https://github.com/AThomsen)
  - [barongreenback](https://github.com/BaronGreenback)
  - [barronpm](https://github.com/barronpm)
+ - [baykatre](https://github.com/baykatre)
  - [bilde2910](https://github.com/bilde2910)
  - [bfayers](https://github.com/bfayers)
  - [BnMcG](https://github.com/BnMcG)
--- a/src/Jellyfin.Networking/AutoDiscoveryHost.cs
+++ b/src/Jellyfin.Networking/AutoDiscoveryHost.cs
@@ -97,7 +97,11 @@ public sealed class AutoDiscoveryHost : BackgroundService
 
     private async Task RespondToV2Message(IPEndPoint endpoint, UdpClient broadCastUdpClient, CancellationToken cancellationToken)
     {
-        var localUrl = _appHost.GetSmartApiUrl(endpoint.Address);
+        var smartUrl = _appHost.GetSmartApiUrl(endpoint.Address);
+        var localUrl = IsLoopbackUrl(smartUrl)
+            ? _appHost.GetLocalApiUrl(_networkManager.GetBindAddress(endpoint.Address, out _, skipOverrides: true))
+            : smartUrl;
+
         if (string.IsNullOrEmpty(localUrl))
         {
             _logger.LogWarning("Unable to respond to server discovery request because the local ip address could not be determined");
@@ -117,4 +121,9 @@ public sealed class AutoDiscoveryHost : BackgroundService
             _logger.LogError(ex, "Error sending response message");
         }
     }
+
+    private static bool IsLoopbackUrl(string? url)
+        => Uri.TryCreate(url, UriKind.Absolute, out var uri)
+           && IPAddress.TryParse(uri.Host, out var ip)
+           && IPAddress.IsLoopback(ip);
 }


Shadowghost (approved):

baykatre:
Updated the approach based on the previous review. Instead of bypassing GetSmartApiUrl entirely, we now only fall back to GetBindAddress(skipOverrides: true) when GetSmartApiUrl returns a loopback address. Non-loopback PublishedServerUrl values (custom domains, specific IPs, reverse proxy URLs) are preserved.

This also addresses #15898 where discovery from a different subnet returns ::1.

@Shadowghost would appreciate your review when you have a chance.

Shadowghost:
I think that's fine.
Seems like the commit author isnt your GH email btw, so it doesn't link correctly

baykatre:
> I think that's fine. Seems like the commit author isnt your GH email btw, so it doesn't link correctly

Fixed, thanks for the heads up — was using my work email by mistake.

baykatre:
> I think that's fine. Seems like the commit author isnt your GH email btw, so it doesn't link correctly

Also updated the PR description to comply with the LLM policy. Would appreciate a re-review when you get a chance.

baykatre:
@Shadowghost Updated the approach since your last review — now using GetSmartApiUrl first and only falling back to GetBindAddress when the result is loopback. Could you take another look?

Shadowghost (approved):

baykatre:
@Bond-009 @JPVenson This PR needs a second review to be merged. It's a one-file change that fixes auto-discovery returning loopback addresses when PublishedServerUrl is set. Would be great if you could take a look when you have a chance. :)

JPVenson (approved):

Bond-009 (changes requested):
The whole point of the PublishedServerUrl setting is that the server will advertise it?

baykatre:
> The whole point of the PublishedServerUrl setting is that the server will advertise it?

In my setup I had internal set to http://localhost:8096 and external to the real domain, the server sits behind a reverse proxy and binds to localhost. Internal is meant for how the server listens locally, not for what it should advertise to other devices on the network. But discovery falls into the internal path and returns localhost to my TV, which is useless. The fix only overrides when the result is loopback, real domains and IPs are left untouched. And this change only affects auto-discovery, nothing else uses this code path. Happy to adjust the approach if you think it should be handled differently though.

baykatre:
@Bond-009 happy to adjust anything if you've had a chance to think it over.

baykatre:
@Bond-009 you have any questions or improvements?

baykatre:
> The whole point of the PublishedServerUrl setting is that the server will advertise it?

@Bond-009 any comments or want to discuss something?

baykatre:
Why you marked to my comments as spam? @Bond-009 If you have any additional comment, we can discuss. But I don't understand why this code is waiting. There is no any comment or improvement. 

@Shadowghost @JPVenson you have any additional comment? How can we fix this?

cvium:
You are aggressively pinging maintainers. Stop. We will get to it when we get to it, and constantly pinging people puts your PR at the bottom of the list.

baykatre:
> You are aggressively pinging maintainers. Stop. We will get to it when we get to it, and constantly pinging people puts your PR at the bottom of the list.

Actually, it's not aggressively. If I put a lot of comments in short time intervals, it will be aggressive. The comment drop in here two weeks ago and I return back to continuously. If there is any article of discussing I'll be glad for this. But only thing I want, review and approve. Marking needs work and no respond in two weeks is really sad.

cvium:
It's not really up for debate. I am telling you it's aggressive. We're a small team and this alleged bug is very low priority.

Also I agree with Bond. The point of the published server url is to override the address that is published to clients. Why are you advertising localhost?

baykatre:
> It's not really up for debate. I am telling you it's aggressive. We're a small team and this alleged bug is very low priority.
> 
> Also I agree with Bond. The point of the published server url is to override the address that is published to clients. Why are you advertising localhost?

My intention wasn't to argue, just to send a simple code fix. I apologize for the misunderstanding. Also, I'm not advertising localhost. It's just that when a reverse proxy is used and the server is behind that proxy, it incorrectly publishes itself as localhost. This fix is ​​to resolve that.

cvium:
> Set PublishedServerUrl to http://localhost:8096

Your first step to reproduce is to set this setting. It does not really make sense to set it to localhost. It's an override for non-standard setups that need more control over what is published in LAN discovery.

baykatre:
> > Set PublishedServerUrl to http://localhost:8096
> 
> Your first step to reproduce is to set this setting. It does not really make sense to set it to localhost. It's an override for non-standard setups that need more control over what is published in LAN discovery.

I've been trying to explain this from the beginning. The scenario here is: the URL I want to access externally is different. I can specify it as "external". But when I do that, the reverse proxy redirect doesn't work on localhost. This requires defining localhost internally. As a result, the broadcast address on the LAN appears as localhost. This fix only solves this in isolation. If you have further questions, I can explain with screenshots and scenario examples.

cvium:
> > > Set PublishedServerUrl to http://localhost:8096
> > 
> > 
> > Your first step to reproduce is to set this setting. It does not really make sense to set it to localhost. It's an override for non-standard setups that need more control over what is published in LAN discovery.
> 
> I've been trying to explain this from the beginning. The scenario here is: the URL I want to access externally is different. I can specify it as "external". But when I do that, the reverse proxy redirect doesn't work on localhost. This requires defining localhost internally. As a result, the broadcast address on the LAN appears as localhost. This fix only solves this in isolation. If you have further questions, I can explain with screenshots and scenario examples.

Your specific scenario sounds quite common (maybe I'm misunderstanding?) and we don't recommend setting Published Server URL for that. It sounds to me like your have it misconfigured. Adding your reverse proxy to the [known proxies setting](https://jellyfin.org/docs/general/post-install/networking/reverse-proxy/#forwarded-for-headers) should be enough.

baykatre:
> Your specific scenario sounds quite common (maybe I'm misunderstanding?) and we don't recommend setting Published Server URL for that. It sounds to me like your have it misconfigured. Adding your reverse proxy to the [known proxies setting](https://jellyfin.org/docs/general/post-install/networking/reverse-proxy/#forwarded-for-headers) should be enough.

Just think about it. Does it make any sense to type `localhost` on a LAN? Can anyone connect to a server that's broadcasting as `localhost`? :)

EraYaN:
Hence why you should not set it in that field, that field should contain the LAN resolvable DNS name or IP.

baykatre:
> Hence why you should not set it in that field, that field should contain the LAN resolvable DNS name or IP.

If you enter the IP address on the LAN or a name resolved by DNS in that internal field instead of localhost, when performing reverse proxying, it constantly tries to resolve the IP and returns, causing a loop. You can try redirecting with local DNS. Therefore, updating directly there doesn't make sense in this scenario. The change is in the discovery section. Is there any reason why broadcasting externally as localhost from here would be logical?

EraYaN:
You should put the domain name or IP the reverse proxy listens to there. The one you can use within the LAN to reach the correct host essentially, no matter the layers in between.

Bond-009:
Works as intended

PR #16516, opened 2026-03-30
invalidate stale LinkedChild cache + cleanup broken links

upscaylman:
ran into this while investigating why merged movies sometimes show ghost versions after you delete one of the files.

Two things going on:

GetLinkedChild in BaseItem caches the ItemId for perf, but when the item gets deleted from the library, it never clears that cache. So it keeps trying to look up a dead guid and silently fails. Fixed by nulling ItemId when the lookup returns nothing, so next call will actually try to re-resolve.

GetLinkedAlternateVersions in Video just skips entries that dont resolve but never removes them from the array. So every time you access the versions list it tries (and fails) to resolve the same dead entries. Now it cleans them up when it detects stale links.

Build passes, tested locally by deleting a file from a merged group and checking that the phantom entry goes away after a refresh.

### DIFF
--- a/tests/Jellyfin.Controller.Tests/Entities/VideoLinkedAlternateTests.cs
+++ b/tests/Jellyfin.Controller.Tests/Entities/VideoLinkedAlternateTests.cs
@@ -1,0 +1,129 @@
+using System;
+using System.Linq;
+using MediaBrowser.Controller.Entities;
+using MediaBrowser.Controller.Library;
+using MediaBrowser.Model.IO;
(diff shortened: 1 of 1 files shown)


Shadowghost:
#16062 reworks how linked children work and will take care of this already.

upscaylman:
makes sense, happy to close this if #16062 is landing soon. just wanted to fix the ghost versions issue in the meantime since it's a small targeted change. closing in favor of your rework, thanks for the heads up!

PR #16517, opened 2026-03-30
better primary selection when merging versions with same resolution

upscaylman:
When you merge versions and multiple files have the same width (like two 1080p encodes), the current code just picks whatever comes first basically. Theres no tiebreaker after the width check.

This adds bitrate as a secondary sort, and then a basic quality score from the filename as last resort - remux > bluray > web-dl. Not the most elegant thing ever but it works for the common case where you have a remux and a web-dl at the same resolution and you want the remux as primary.

I know parsing filenames is fragile, happy to hear if theres a better approach. The bitrate tiebreak alone should handle most cases anyway since remuxes have higher bitrate.

### DIFF
--- a/Jellyfin.Api/Controllers/VideosController.cs
+++ b/Jellyfin.Api/Controllers/VideosController.cs
@@ -211,6 +211,8 @@ public class VideosController : BaseJellyfinApiController
                     return 0;
                 })
                 .ThenByDescending(i => i.GetDefaultVideoStream()?.Width ?? 0)
+                .ThenByDescending(i => i.GetDefaultVideoStream()?.BitRate ?? 0)
+                .ThenByDescending(i => GetSourceQualityScore(i.Path))
                 .First();
         }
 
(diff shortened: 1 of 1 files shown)


IDisposable (approved):

Jellyfin.Api/Controllers/VideosController.cs
```diff
@@ -655,4 +657,37 @@ [FromQuery] [RegularExpression(EncodingHelper.ContainerValidationRegex)] string?
...
+            || filename.Contains("blu-ray", StringComparison.OrdinalIgnoreCase))
+        {
+            return 2;
+        }
+
+        if (filename.Contains("web-dl", StringComparison.OrdinalIgnoreCase)
+            || filename.Contains("webdl", StringComparison.OrdinalIgnoreCase))
+        {
+            return 1;
+        }
```
JPVenson, line 689: We do not support that kind of naming scheme and dont want to

JPVenson (changes requested):

theguymadmax:
Closing per Venson's comments.

IDisposable:
Seems like we still should do the tie-break on bitrate, even if we don't want the quality assumptions.

PR #16537, opened 2026-04-01
Fix episode parser misidentifying video resolutions as season/episode

### LINKED ISSUE #1398: TVDB finding episode, but season is marked as Unknown (Daily Show)
**Describe the bug**

Episode with season and episode number (over 100?) are placed into season Unknown, even thou there is a season defined.
I think this might be related to #1180.

**To Reproduce**
Add a file named like my example
The Daily Show/The Daily Show 24x103 - 2019-05-20 Kirsten Gillibrand [WEBDL-720p][AAC 2.0][x264]-TBS.mkv
JF finds the episode and saves metadata. TVDBID is saved and correct. But a dummy season Unknown is created instead of season 24.

**Expected behavior**

Correct season # should be created, not Season Unknown

**Logs**

No errors in logs

**Screenshots**

image

 - OS: Docker
 - Browser: Chrome
 - Jellyfin Version: 10.3.3

stlcompliance:
Added a method to invalidate matches that look like video dimensions in episode parsing.

**Changes**
Removed the hardcoded season-number blacklist that was masking bad regex matches (e.g. 1920x1080 being parsed as S1920E1080).

Replaced it with a direct validation that detects resolution-style patterns in the matched token and invalidates them explicitly.

This shifts the logic from “guess and block absurd numbers” to actually identifying the root cause of the false positives, making parsing behavior more predictable and maintainable.

**Issues**
Fixes #1398

### DIFF
--- a/Emby.Naming/TV/EpisodePathParser.cs
+++ b/Emby.Naming/TV/EpisodePathParser.cs
@@ -183,19 +183,62 @@ namespace Emby.Naming.TV
                     result.Success = result.EpisodeNumber.HasValue;
                 }
 
-                // Invalidate match when the season is 200 through 1927 or above 2500
-                // because it is an error unless the TV show is intentionally using false season numbers.
-                // It avoids erroneous parsing of something like "Series Special (1920x1080).mkv" as being season 1920 episode 1080.
-                if ((result.SeasonNumber >= 200 && result.SeasonNumber < 1928)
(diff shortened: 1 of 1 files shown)


JPVenson (changes requested):
Please fix the build errors (and always check if your code actually builds) and add tests for that change

PR #16539, opened 2026-04-01
Fix subtitle saving

MBR-0001:
**Changes**
https://github.com/jellyfin/jellyfin/commit/d3907afde7ea0abdf0d3e291b0bd572653bae7ac broke subtitle saving, see issue https://github.com/jellyfin/jellyfin-plugin-opensubtitles/issues/201

`_allowedSubtitleFormats` contains formats without the `.`, so `"." + extension` will always return false
image

**Issues**
Fixes: https://github.com/jellyfin/jellyfin-plugin-opensubtitles/issues/201
Fixes: #16544
Fixes: https://github.com/jellyfin/jellyfin-plugin-opensubtitles/issues/202

### DIFF
--- a/MediaBrowser.Providers/Subtitles/SubtitleManager.cs
+++ b/MediaBrowser.Providers/Subtitles/SubtitleManager.cs
@@ -239,8 +239,7 @@ namespace MediaBrowser.Providers.Subtitles
 
         private async Task TrySaveToFiles(Stream stream, List<string> savePaths, Video video, string extension)
         {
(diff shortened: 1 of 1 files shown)


MBR-0001:
would it make more sense to just remove the `"." + extension` part?

theguymadmax:
Yes, just leaving the extension would be better.

LTe:
In case of regression I think we should add specs for that

vinnie1234:
When will this be released?

Issue #16531, opened 2026-04-01
Migration 20260206200000_FixLibrarySubtitleDownloadLanguages failed


TheColin21:
### Description of the bug

After updating from 10.11.6 to 10.11.7 the server does not start anymore as the migration 20260206200000_FixLibrarySubtitleDownloadLanguages fails.

### Reproduction steps

1. Have a 10.11.6 server
2. Update to 10.11.7
3. ~profit~

### What is the current _bug_ behavior?

The server fails to start

### What is the expected _correct_ behavior?

The server performs the migration and starts

### Jellyfin Server version

10.11.7

### Specify the build version

10.11.7

### Environment

```markdown
- OS: TrueNAS 25.10.2.1
- Linux Kernel: 6.12.33
- Virtualization: Docker
- CPU Model: Ryzen 4650G
- GPU Model: Intel Ark A310
- Plugins: don't know and can't look up now
- Reverse Proxy: Traefik
- Base URL:
- Networking:
- Jellyfin Data Storage & Filesystem:
- Media Storage & Filesystem:
- External Integrations:
```

### Jellyfin logs

```shell
[05:27:12] [INF] [9] InternalCodeMigration: Perform migration 20260206200000_FixLibrarySubtitleDownloadLanguages
[05:27:12] [INF] [9] Jellyfin.Server.Migrations.Routines.FixLibrarySubtitleDownloadLanguages: Starting to fix library subtitle download languages.
[05:27:13] [FTL] [9] InternalCodeMigration: Error: Object reference not set to an instance of an object.
[05:27:13] [ERR] [9] InternalCodeMigration: Migration 20260206200000_FixLibrarySubtitleDownloadLanguages failed
System.NullReferenceException: Object reference not set to an instance of an object.
   at Jellyfin.Server.Migrations.Routines.FixLibrarySubtitleDownloadLanguages.PerformAsync(CancellationToken cancellationToken)
   at Jellyfin.Server.Migrations.Stages.CodeMigration.Perform(IServiceProvider serviceProvider, IStartupLogger logger, CancellationToken cancellationToken)
   at Jellyfin.Server.Migrations.JellyfinMigrationService.InternalCodeMigration.PerformAsync(IStartupLogger logger)
   at Jellyfin.Server.Migrations.JellyfinMigrationService.MigrateStepAsync(JellyfinMigrationStageTypes stage, IServiceProvider serviceProvider)
[05:27:13] [FTL] [9] Main: Error while starting server
[...]

PR #16541, opened 2026-04-01
Fix NRE in FixLibrarySubtitleDownloadLanguages migration

MBR-0001:
**Changes**
Couldn't reproduce but added null checks where possible

**Issues**
Fixes https://github.com/jellyfin/jellyfin/issues/16531

Used vscode autocompelte thing


MBR-0001:
bruh

screw u @Shadowghost :P

PR #16575, opened 2026-04-05
Use FFmpeg bluray: protocol for Blu-ray disc playback

u6bkep:
**Changes**

Replace the BDInfo-based approach for Blu-ray playback (which manually picked .m2ts files and built concat lists) with FFmpeg's native `bluray:` protocol. This lets ffmpeg/ffprobe handle playlist selection and stream demuxing via libbluray directly, producing more reliable metadata and playback for BDMV folders and `.mpls` playlist files.

- Add `ParseBlurayPath` to `EncodingHelper` for extracting disc root and playlist number from BDMV paths (`.mpls` files, `BDMV` dirs, disc roots), used by `MediaEncoder`, `SubtitleEncoder`, and `EncodingHelper` itself
- Pass `-playlist` flag to ffmpeg when the media path is an `.mpls` file
- Use `bluray:` input protocol in `MediaEncoder`, `EncodingHelper`, and `SubtitleEncoder`
- Simplify `FetchBdInfo` to only extract chapter marks (libbluray handles streams and runtime via ffprobe)
- Add `IBlurayExaminer.GetDiscInfo(path, playlistName)` overload (with default interface method for ABI compatibility) so BDInfo looks up the specific playlist for chapter marks instead of always using the longest one

## User-facing behavior changes

**BDMV folders** (the existing default): Behavior is largely unchanged for users who add BDMV folders to their libraries the standard way. Jellyfin still auto-selects the main playlist via libbluray (the same library mpv and VLC use), but now delegates stream demuxing to ffmpeg rather than manually selecting `.m2ts` files. This should improve reliability for discs  with complex playlist structures (seamless branching, multi-angle, etc.) and produce more accurate stream metadata (codec details, HDR info, runtime) since ffprobe reads them directly from the transport stream.

**`.mpls` playlist files**: This change adds support for resolving individual `.mpls` files as playable items. When a `.mpls` path is used (e.g., `/disc/BDMV/PLAYLIST/00202.mpls`), Jellyfin extracts the playlist number and passes it to ffmpeg via `-playlist N -i bluray:/disc/root`. This enables plugins to split a single Blu-ray disc into multiple library items (e.g., individual episodes from a TV show disc) by pointing each item at a different `.mpls` file. Chapter marks are looked up per-playlist.

**Planned plugin**: I'm developing a companion plugin
([jellyfin-plugin-thediscdb](https://github.com/u6bkep/jellyfin-plugin-thediscdb)) that uses [TheDiscDb](https://github.com/TheDiscDb/data) to resolve BDMV folders into individual episodes. It computes a content hash  from the disc's `.m2ts` files, looks up the disc in TheDiscDb's database, and creates Episode items pointing to the correct `.mpls` playlist files. This PR provides the server-side foundation that makes that plugin possible.

**Issues**

Addresses Blu-ray disc (BDMV) playback — currently Jellyfin uses BDInfo to find `.m2ts` files and builds concat file lists, which is fragile for multi-clip playlists and misses libbluray features like seamless branching. The `bluray:` protocol delegates all of this to ffmpeg's libbluray integration.

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -1231,7 +1231,21 @@ namespace MediaBrowser.Controller.MediaEncoding
                 arg.Append(canvasArgs);
             }
 
-            if (state.MediaSource.VideoType == VideoType.Dvd || state.MediaSource.VideoType == VideoType.BluRay)
+            if (state.MediaSource.VideoType == VideoType.BluRay)
+            {
+                var (discRoot, playlistNumber) = ParseBlurayPath(state.MediaPath);
+
+                if (playlistNumber.HasValue)
+                {
+                    arg.Append(" -playlist ")
+                        .Append(playlistNumber.Value);
+                }
+
+                arg.Append(" -i bluray:\"")
+                    .Append(discRoot.Replace("\"", "\\\"", StringComparison.Ordinal))
+                    .Append("\" ");
+            }
+            else if (state.MediaSource.VideoType == VideoType.Dvd)
             {
                 var concatFilePath = Path.Join(_configurationManager.CommonApplicationPaths.CachePath, "concat", state.MediaSource.Id + ".concat");
                 if (!File.Exists(concatFilePath))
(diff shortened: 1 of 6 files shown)


Bond-009:
Duplicate of #11833

u6bkep:
@Bond-009 is #11833 going anywhere? no change in 1.5 years.

PR #16583, opened 2026-04-06
Return text subtitles early while extraction continues in background

Molier:
## What this fixes

On cold cache, text subtitle requests block the HTTP response until ffmpeg finishes extracting *every* subtitle stream in the file. On a large MKV with many tracks, that's 30–60+ seconds (in my setup, often 3+ minutes when HLS transcoding is running in parallel). The client sees a hung request and in practice most players either time out or show a spinner long enough that the user just gives up. Issue #2547, referenced by several downstream reports.
### commits
  1. return text subtitles early while extraction continues in background
  2. refactor `GetReadableFile` to reduce cognitive complexity
  3. wrap the returned stream in `TailingFileStream` so the HTTP response delivers the complete file in a single connection

  For text subtitles (SRT, SSA), `GetReadableFile` now returns as soon as the output file exists on disk instead of blocking until ffmpeg exits.
  Combined with `-flush_packets 1` (#16440), first bytes land within ~1 second. Extraction continues in the background, locks are held, remaining streams finish, cleanup runs normally. Falls back to the old blocking path for ASS (needs `SetAssFont` post-processing) and PGS (binary format).
  On slow storage with concurrent HLS transcoding, this reduces subtitle load time from 3+ minutes to <1 second.

## Scope — what this PR does NOT fix

The **convert path** (client requests `.vtt` or `.js`, source is `.srt`) is **unchanged** by this PR. `ConvertSubtitles` → `SubtitleEditParser.Parse` does a synchronous `stream.ReadAllLines().ToList()` before emitting any cues, so the tailing stream is drained fully before the response starts. **Jellyfin Web requests `.js`**, meaning web users see no improvement from this PR alone. A follow-up PR adding an incremental SRT parser + pull-based stream adapter is in progress and will be submitted separately once this one lands.

  ## LLM/AI Use

Per [Jellyfin's AI Use Policy](https://jellyfin.org/docs/general/contributing/llm-policies):
  - I wrote the entire PR description myself.
  - I used Claude (Opus 4.6) to help me navigate the subtitle extraction codepaths during investigation, brainstorm the`TailingFileStream` approach after @gnattu pointed out the truncation bug, and build the test harness I used. The code in the three commits was reviewed by me end-to-end; I understand what each piece is doing and why, and I'm able to address review feedback myself.
  - I tested this PR locally against `v10.11.7` (what my personal Jellyfin server runs) and confirmed it builds cleanly against `master` before submitting. It's been running on my server for a few days with a handful of real users and is holding up well.

### DIFF
--- a/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
+++ b/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
@@ -195,7 +195,13 @@ namespace MediaBrowser.MediaEncoding.Subtitles
                 }
             }
 
-            return AsyncFile.OpenRead(fileInfo.Path);
+            var fileStream = AsyncFile.OpenRead(fileInfo.Path);
+            if (fileInfo.ExtractionTask is not null && !fileInfo.ExtractionTask.IsCompleted)
+            {
+                return new TailingFileStream(fileStream, fileInfo.ExtractionTask);
+            }
+
+            return fileStream;
         }
 
         internal async Task<SubtitleInfo> GetReadableFile(
@@ -977,14 +983,18 @@ namespace MediaBrowser.MediaEncoding.Subtitles
             }
 
             // For text-based subtitles (SRT, SSA): return as soon as the output file has data on disk.
-            // With -flush_packets 1, ffmpeg writes data within ~1 second. The client gets a valid partial
-            // subtitle file immediately instead of waiting minutes for full extraction to complete.
+            // With -flush_packets 1, ffmpeg writes data within ~1 second. The HTTP response is then
+            // wrapped in a TailingFileStream so it streams progressively as ffmpeg keeps writing,
+            // delivering the full file in a single connection without blocking on extraction.
             // ASS is excluded because SetAssFont rewrites the file after extraction (race condition).
             // PGS is excluded because it's a binary bitmap format that needs the complete file.
-            if (IsEarlyReturnSubtitleFormat(outputFormat)
-                && await TryReturnEarly(mediaSource, outputPath).ConfigureAwait(false))
+            if (IsEarlyReturnSubtitleFormat(outputFormat))
             {
-                return info;
+                var extractionTask = await TryReturnEarly(mediaSource, outputPath).ConfigureAwait(false);
+                if (extractionTask is not null)
+                {
+                    return info with { ExtractionTask = extractionTask };
+                }
             }
 
             // PGS (binary), ASS (needs SetAssFont post-processing), or early-return timed out:
@@ -995,9 +1005,10 @@ namespace MediaBrowser.MediaEncoding.Subtitles
 
         /// <summary>
         /// Starts background extraction and waits for the output file to have data.
-        /// Returns true if the file is ready and extraction should continue in the background.
+        /// Returns the still-running extraction task if the file is ready, or null if early return
+        /// was not possible (extraction finished synchronously, or file did not appear in time).
         /// </summary>
-        private async Task<bool> TryReturnEarly(MediaSourceInfo mediaSource, string outputPath)
+        private async Task<Task?> TryReturnEarly(MediaSourceInfo mediaSource, string outputPath)
         {
             // Use CancellationToken.None so extraction continues even if this HTTP request completes
             var extractionTask = ExtractAllExtractableSubtitles(mediaSource, CancellationToken.None);
@@ -1025,12 +1036,12 @@ namespace MediaBrowser.MediaEncoding.Subtitles
                     },
                     TaskScheduler.Default);
 
-                return true;
+                return extractionTask;
             }
 
             // Extraction finished first (fast storage) or file data timed out — await for error handling
             await extractionTask.ConfigureAwait(false);
-            return false;
+            return null;
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


DrHurt:
This is great and will fix an issue that’s been plaguing Jellyfin for years!

I’m guessing the http download would continue slowly until extraction completes (eof) and full subtitle file is downloaded? 

And that the web client is happy to start rendering with a partial file?

Very well done! Thank you.

Molier:
> This is great and will fix an issue that’s been plaguing Jellyfin for years!
> 
> I’m guessing the http download would continue slowly until extraction completes (eof) and full subtitle file is downloaded?
> 
> And that the web client is happy to start rendering with a partial file?
> 
> Very well done! Thank you.

Exactly right on both. 
The HTTP response streams the file from disk as ffmpeg keeps writing (-flush_packets 1 flushes per packet), and browsers/HLS.js happily render a partial SRT/SSA — each cue is self-contained, so whatever's on disk displays immediately and new cues pop in as they land. I teested with 4-track MKVs: subs show in <1s, extraction finishes about 20s later in the background with no client-visible hiccup for me.

Molier:
I did a check for my lib, here's the subtitle codec breakdown on my 100k-stream library:

  | Codec | Count | % of embedded | Covered by this PR? |
  |---|---:|---:|---|
  | subrip (SRT) | 80,017 | 79.9% | ✅ |
  | ass | 9,615 | 9.6% | ❌ (SetAssFont post-processing) |
  | DVDSUB (VobSub) | 5,586 | 5.6% | ❌ (binary) |
  | PGSSUB | 4,468 | 4.5% | ❌ (binary bitmap) |
  | mov_text | 368 | 0.4% | ✅ (extracts to SRT) |
  | DVBSUB | 62 | <0.1% | ❌ (binary) |
  | eia_608 | 36 | <0.1% | ✅ (extracts to SRT) |
  | other | ~4 | <0.1% | — |

  So this PR fixes the cold-cache delay for ~80% of real-world embedded subtitles. ASS is the next biggest chunk but deserves its own PR.
Cleanly handling it means either eliminating `SetAssFont` or moving its substitution out of the on-disk rewrite path, which is a larger discussion. But Id gladly hear opinions on it before making that PR @DrHurt ?

gnattu:
May I ask a question: do we really think a partial subtitle is better than nothing, or do I miss something?

I do not quite understand how it is better than nothing if none of our clients implemented proper re-fetch or real subtitle streaming. Serving incomplete subtitles to clients should be considered a bug, not a feature. If I’m interpreting this correctly, this PR is proudly promoting a buggy behavior as a feature?

I’m happy to be wrong, but this PR really needs some clarifications. If serving incomplete subtitles is all it does without proper client side handling, I’m afraid I cannot accept this PR as-is.

DrHurt:
@Molier ASS definitely deserves a separate PR because it’s a different, more complex rendering pipeline (SubtitlesOctopus) and will likely require more extensive changes and testing.

Like you said, the easier way is probably removing `SetAssFont`, perhaps behind a user toggle to allow faster extraction at expense of less accurate rendering and seeing how SubtitlesOctopus behaves.

PS. Last question: what happens if you seek to the end of movie before subtitles finish downloading? Do subtitles gracefully resume once extraction reaches playback?

Molier:
> May I ask a question: do we really think a partial subtitle is better than nothing, or do I miss something?
> 
> I do not quite understand how it is better than nothing if none of our clients implemented proper re-fetch or real subtitle streaming. Serving incomplete subtitles to clients should be considered a bug, not a feature. If I’m interpreting this correctly, this PR is proudly promoting a buggy behavior as a feature?
> 
> I’m happy to be wrong, but this PR really needs some clarifications. If serving incomplete subtitles is all it does without proper client side handling, I’m afraid I cannot accept this PR as-is.

Ofc, good question and remark, I think youre very correct.
Also I looked at some options to fix it with client side changes, but that was all way more involved for not enough benefit for this more clear fix. 
But very poignant remark, cuz I agree that as is after some thinking, it would break on faster hardware, eg. if the consumer (client) is faster than producer (ffmpeg): consumer hits EOF after the first ~KB, response closes, client receives a truncated file. So you're right I think.

  I'm pushing a follow-up commit that wraps the FileStream in a tailing stream: when `Read` returns 0, it checks whether the extraction task is still running and waits briefly before retrying, only returning real EOF once extraction has completed. The HTTP response then always delivers the  complete file in a single connection, just streamed progressively as ffmpeg writes it. No client-side re-fetch or streaming-subtitle logic required, every existing client handles it because it's just a slow HTTP file download.

  I'll push it shortly and update here once tested.

DrHurt:
Slight tangent, but why doesn't the Jellyfin Web Client use native HLS WebVTT for text based subtitles. That way users get instant subtitles, native platform controls & styling. The server already has everything needed for this.

Molier:
@gnattu	thank you for the help, ive made changes and its ready for review now. 
@DrHurt the changes we hoped for took a bit more changes, but ive gotten fully server side streaming to work for the web clients im pretty sure!

DrHurt:
> @gnattu thank you for the help, ive made changes and its ready for review now. @DrHurt the changes we hoped for took a bit more changes, but ive gotten fully server side streaming to work for the web clients im pretty sure!

I’m genuinely mind blown by what you’ve done here!! Thank you for taking the time to do this.

My only lingering question is what happens if you seek video beyond the downloaded subtitles range, does the player gracefully resume rendering of subtitles once downloaded file catches up?

Molier:
> I’m genuinely mind blown by what you’ve done here!! Thank you for taking the time to do this.
> 
> My only lingering question is what happens if you seek video beyond the downloaded subtitles range, does the player gracefully resume rendering of subtitles once downloaded file catches up?

if the user seeks to 01:30:00 while the downloaded range is at 00:20:00:
  - On clients that parse progressively as bytes arrive (mpv, VLC, Jellyfin Web's subtitle renderer, Infuse): the cue for 01:30:00 simply lights up the moment its bytes arrive in the response body. The seek is invisible to the subtitle layer — it just keeps consuming the stream and rendering whatever cue bracket the current playback position falls into. Typical observed behaviour: subtitles "resume" rendering a few hundred ms to a couple of seconds after the seek, depending on how far ahead of ffmpeg you jumped.
  - On clients that buffer the whole file before parsing (mainly older HTML `<track>` implementations and a few legacy mobile players): they simply wait until the response completes, same as stock Jellyfin today. No regression vs. the current behaviour.

DrHurt:
Amazing! Thanks for confirming. Here’s hoping this gets merged. 

It will be a game changer for so many people!

Molier:
> Amazing! Thanks for confirming. Here’s hoping this gets merged.
> 
> It will be a game changer for so many people!

this won't just yet, the other larger PR will though. Streaming for 100 percent of subs

DrHurt:
> > Amazing! Thanks for confirming. Here’s hoping this gets merged.
> > It will be a game changer for so many people!
> 
> this won't just yet, the other larger PR will though. Streaming for 100 percent of subs

Looking forward to gnattu's thoughts on this PR after your recent update. 

I would suggest you postpone any work on additional formats until this is approved. Jellyfin is currently in the process of replacing the PGS renderer (libpgs) with a new one (libbitsub).

Once this PR and the new PGS renderer get merged, you could then test PGS streaming with libbitsub. If any changes are required on the rendering side, I'm sure the dev behind libbitsub would be happy to help.

DrHurt:
@gnattu any thoughts on this after the latest changes? 

It would potentially provide a much needed speed up for subtitles on WebOS & browsers

soultaco83:
So far been testing this PR. I have had no issues and users have reported better subtitle response

DrHurt:
@Molier can you review this conflict so this can hopefully move forward. This will provide a very meaningful improvement for Jellyfin 12.0.0 if the team agrees to merge it.

Issue #16584, opened 2026-04-07
Add "MaxParallelism" Trickplay option


Starkiller21321:
In 10.11.8 the `Generate Trickplay Images` scheduled task generates images sequentially. This is reasonable for smaller libraries or limited hardware resources.  With larger libraries, this means that the generation can take hours to days to complete due to the backlog of images it needs to create. 
This issue proposes a new Trickplay option that allows administrators to adjust the `MaxParallelism` value to best fit the needs of their library(s) and hardware limitations.

Feature request: https://features.jellyfin.org/posts/3498/multicore-trickplay-generation-by-default

JPVenson:
feature requests are only tracked on our finder instance and not in the GH repo.

PR #16585, opened 2026-04-07
Add "MaxParallelism" Trickplay option

Starkiller21321:
Adds configurable "MaxParallelism" trickplay option.

Feature request: [https://features.jellyfin.org/posts/3498/multicore-trickplay-generation-by-default](https://features.jellyfin.org/posts/3498/multicore-trickplay-generation-by-default)

**Changes**
- Adds new MaxParallelism option to `TrickplayOptions` (default is set to 1). 
- Replaces the sequential foreach loop in `TrickplayImagesTask` with a `Parallel.ForEachAsync` allowing concurrent processing.
- Changes `TrickplayManager._resourcePool` to an instance-level semaphore initialized from the new `MaxParallelism` config value. 

**Issues**
https://github.com/jellyfin/jellyfin/issues/16584

### DIFF
--- a/Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
+++ b/Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
@@ -28,7 +28,7 @@ namespace Jellyfin.Server.Implementations.Trickplay;
 /// <summary>
 /// ITrickplayManager implementation.
 /// </summary>
-public class TrickplayManager : ITrickplayManager
+public class TrickplayManager : ITrickplayManager, IDisposable
 {
     private readonly ILogger<TrickplayManager> _logger;
     private readonly IMediaEncoder _mediaEncoder;
@@ -40,8 +40,9 @@ public class TrickplayManager : ITrickplayManager
     private readonly IApplicationPaths _appPaths;
     private readonly IPathManager _pathManager;
 
-    private static readonly AsyncNonKeyedLocker _resourcePool = new(1);
+    private readonly AsyncNonKeyedLocker _resourcePool;
     private static readonly string[] _trickplayImgExtensions = [".jpg"];
+    private bool _disposed;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="TrickplayManager"/> class.
@@ -75,6 +76,33 @@ public class TrickplayManager : ITrickplayManager
         _dbProvider = dbProvider;
         _appPaths = appPaths;
         _pathManager = pathManager;
+        _resourcePool = new(config.Configuration.TrickplayOptions.MaxParallelism);
+    }
+
+    /// <inheritdoc />
+    public void Dispose()
+    {
+        Dispose(true);
+        GC.SuppressFinalize(this);
+    }
+
+    /// <summary>
+    /// Releases unmanaged and optionally managed resources.
+    /// </summary>
+    /// <param name="disposing"><c>true</c> to release both managed and unmanaged resources; <c>false</c> to release only unmanaged resources.</param>
+    protected virtual void Dispose(bool disposing)
+    {
+        if (_disposed)
+        {
+            return;
+        }
+
+        if (disposing)
+        {
+            _resourcePool?.Dispose();
+        }
+
+        _disposed = true;
     }
 
     /// <inheritdoc />
--- a/Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
+++ b/Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
@@ -29,7 +29,7 @@ namespace Jellyfin.Server.Implementations.Trickplay;
 /// <summary>
 /// ITrickplayManager implementation.
 /// </summary>
-public partial class TrickplayManager : ITrickplayManager
+public sealed partial class TrickplayManager : ITrickplayManager, IDisposable
 {
     private readonly ILogger<TrickplayManager> _logger;
     private readonly IMediaEncoder _mediaEncoder;
@@ -41,7 +41,7 @@ public partial class TrickplayManager : ITrickplayManager
     private readonly IApplicationPaths _appPaths;
     private readonly IPathManager _pathManager;
 
-    private static readonly AsyncNonKeyedLocker _resourcePool = new(1);
+    private readonly AsyncNonKeyedLocker _resourcePool;
     private static readonly string[] _trickplayImgExtensions = [".jpg"];
 
     /// <summary>
@@ -76,6 +76,20 @@ public partial class TrickplayManager : ITrickplayManager
         _dbProvider = dbProvider;
         _appPaths = appPaths;
         _pathManager = pathManager;
+        var maxParallelism = config.Configuration.TrickplayOptions.MaxParallelism;
+        if (maxParallelism < 1)
+        {
+            _logger.LogWarning("Trickplay max parallelism {MaxParallelism} is too small, using value of 1", maxParallelism);
+            maxParallelism = 1;
+        }
+
+        _resourcePool = new(maxParallelism);
+    }
+
+    /// <inheritdoc />
+    public void Dispose()
+    {
+        _resourcePool.Dispose();
     }
 
     /// <inheritdoc />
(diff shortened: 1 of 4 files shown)


Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
```diff
@@ -28,7 +28,7 @@ namespace Jellyfin.Server.Implementations.Trickplay;
 /// <summary>
 /// ITrickplayManager implementation.
 /// </summary>
-public class TrickplayManager : ITrickplayManager
+public class TrickplayManager : ITrickplayManager, IDisposable
```
JPVenson: Are the changes even used anywhere?

JPVenson (changes requested):

Jellyfin.Server.Implementations/Trickplay/TrickplayManager.cs
Starkiller21321: `_resourcePool` is a type of `AsyncNonKeyedLocker` which implements the `IDisposable` interface. This allows `_resourcePool` to act as a concurrency throttle by internally creating a `SemaphoreSlim`. `SemaphoreSlim` has an unmanaged `WaitHandle` that is not released unless `Dispose()` is explicitly called. 

`_resourcePool` is used in the `RefreshTrickplayDataInternal` (line 257) so it can limit the amount of tasks running. `Dispose()` methods were added on lines 82-95 of `TrickplayManager.cs` to properly manage the cleanup and ensure that disposal only occurs once.

Without the `IDisposable`, the class produces the following build error. `Type 'TrickplayManager' owns disposable field(s) '_resourcePool' but is not disposable[CA1001](https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1001)`.

Let me know if you have any questions or need additional clarification.

JPVenson: dude if i want an AI to tell me what you did i would have asked it myself, i asked you to explain the reasoning why you double locked the access.

Starkiller21321: You asked where the changes were used. Asking me to explain the reasoning is a different question. 
My understanding is that there isn't a double lock in place. If one does exist and it's an issue could you give me some more details so I can try and address the problem?
I updated the `_resourcePool` to no longer be a static value, thus allowing it to receive the new updated config value in the constructor.

Starkiller21321: I see what you are referring to now. I'll work on making that change. 
Sorry about that.

Starkiller21321: I'm very sorry that I misunderstood the question. That is entirely my fault. I've removed the double lock in the PR. Please let me know if you have any other feedback. Again, I apologize for the confusion.

gnattu:
Sorry but before blindly implement feature requests based on speculation that this would improve performance, could you provide some concrete numbers, like how much this could make trickplay generation faster on your library?

As far as I understand, the bottleneck is not trickplay job concurrency at all. The user who creates the feature request might not even be aware of it.

For trickplay generation, the bottleneck is actually the speed at which your CPU (or hardware accelerator, if you enabled hardware acceleration) can decode the stream. We already allow you to configure the number of CPU threads you want to allocate to ffmpeg for decoding the stream. Therefore, parallelism at job level doesn’t help at all in this case. Even if you’re using hardware acceleration, you’re simply feeding more concurrent video streams, which makes each stream processing slower because the hardware throughput is also limited, let alone some GPUs like Nvidia has driver level limit of 8 concurrency decode session per GPU.

I understand your intention to help other users and benefit yourself, but I believe this PR is not doing the most helpful thing.

Also: please, do not copy-paste LLM responses directly during the communication. It is very, very disrespectful.

Starkiller21321:
> Sorry but before blindly implement feature requests based on speculation that this would improve performance, could you provide some concrete numbers, like how much this could make trickplay generation faster on your library?
> 
> As far as I understand, the bottleneck is not trickplay job concurrency at all. The user who creates the feature request might not even be aware of it.
> 
> For trickplay generation, the bottleneck is actually the speed at which your CPU (or hardware accelerator, if you enabled hardware acceleration) can decode the stream. We already allow you to configure the number of CPU threads you want to allocate to ffmpeg for decoding the stream. Therefore, parallelism at job level doesn’t help at all in this case. Even if you’re using hardware acceleration, you’re simply feeding more concurrent video streams, which makes each stream processing slower because the hardware throughput is also limited, let alone some GPUs like Nvidia has driver level limit of 8 concurrency decode session per GPU.
> 
> I understand your intention to help other users and benefit yourself, but I believe this PR is not doing the most helpful thing.
> 
> Also: please, do not copy-paste LLM responses directly during the communication. It is very, very disrespectful.

Thank you for the feedback. I will move the pr into a draft while I address these comments and gather additional data around performance.

aidancrane:
[log_20260409.log](https://github.com/user-attachments/files/26601202/log_20260409.log)

Here is a log file of my media library generating trickplay images over 12hours before I cancelled the job and found this PR.

This was after I discovered trickplay was not enabled on my library and enabled it.

Please consider this as supplementary evidence to support the utility of this PR.

> how much this could make trickplay generation faster on your library?

This being not very parallel means you have to wait it out when the media server has limited hardware acceleration.

gnattu:
> I will experiment with ffmpeg threading settings in jellyfin.

image

This is the very first thing you should tune in trickplay settings. 

> This being not very parallel means you have to wait it out when the media server has limited hardware acceleration.

But the media decoding itself is heavy enough to be bottleneck, and it can be parallel with already-present option.

My exact question is how is this better than just let ffmpeg to use more threads. Handling more tasks at the same time does not magically makes your CPU faster, you are just waiting for more tasks at the same time with similar waiting time, if you already configured trickplay to use all your cpu cores with ffmpeg.

aidancrane:
Let me try that. Will let you know.

Starkiller21321:
The following was from an initial performance test with my current setup. I am running additional performance tests on a larger library size. I plan to provide those results when possible.

- OS: Ubuntu 24.04 LTS
- Linux Kernel: 6.8.0

All Trickplay settings left at the default values with the following exceptions,
 - FFmpeg thread count was set to 0 -auto-
 - No hardware acceleration was used.

127 files were present in the library at the time of the scans.
Original Performance Test as a Control Group:
| Process Priority | Time         |
|------------------|--------------|
| Below Normal     | 29 min 38 s  |
| Normal           | 29 min 35 s  |
 
 Performance Test with this new PR:
| MaxParallelism | Process Priority | Time         |
|----------------|------------------|--------------|
| 1              | Below Normal     | 29 min 40 s  |
| 1              | Normal           | 29 min 34 s  |
| 2              | Below Normal     | 21 min 57 s  |
| 2              | Normal           | 22 min 47 s  |
| 4              | Below Normal     | 22 min 16 s  |
| 6              | Below Normal     | 23 min 1 s   |
| 6              | Normal           | 22 min 41 s  |
| 8              | Below Normal     | 24 min 6 s   |
| 8              | Normal           | 24 min 10 s  |

AndreHatt-git:
You should have compared it to with different number of FFmpeg Threads instead of different system priority.

The test was conducted on just over 100 samples, without HW accel on the VM.

| FFmpeg Threads | Time |
| ------------- | ------------- |
| 1 | 98 min 29s  |
| 2 | 57 min 24s  |
| 4 | 37 min 47s  |
| 8 | 28 min 38s  |

I've had little time to dedicate to testing it with this PR recently, so sorry for not doing it properly but it is hard to imagine that it will somehow help with trickplay time generation further compared to ffmpeg thread count setting.

This is mostly insight that the same can be accomplished by just upping FFmpeg Threads which is already present in trickplay settings.

aidancrane:
First I tried my real media server on v10.11.8 (using a containerized jellyfin, Ryzen 5 3600) and found that it was kind of too noisy given all the different media types on there, the other containers running etc. Left it running for a couple days and through a few terabytes and some different threading options but the timings were all over the place.

I decided to use something a bit more scientific so I then tried this guys fork, made a docker container and started using max parallelism. I also switched from using media files of different sizes to just dashcam footage, because it's all the same size-ish. I also switched from my proxmox server to another windows PC (same processor), ran docker desktop etc.  There was about 250GB from start to finish, and I ran it twice just deleting the whole library and re-adding it about midway through (where session 2 ends and session 3 starts). 

image

It seems like more threads more better for ffmpeg, it also looks like parallelism is actually worse for performance over threading (3t/4p performed faster than 2t/4p) (12t/4p faster than 1t/1p) (oversubscribed the CPU on 12t/4p, I guess some cpu governor magic happening there).

I will do some more testing I think, just to be certain, because it took me a while to get this far and my testing is a bit messy.

aidancrane:
Okay, I've completed parallelism testing. I changed the system.xml file each time and let it rip for 30 minutes on dashcam footage, I'll do threading only next.

```
  <TrickplayOptions>
    <ProcessThreads>t</ProcessThreads>
    <MaxParallelism>p</MaxParallelism>
  </TrickplayOptions>
```

You can see the results here.

```
            session_start             session_end  session_duration_mins  threads  parallelism  files_processed  avg_sec_per_file  median_sec_per_file  min_sec  max_sec
0 2026-04-12 17:58:27.972 2026-04-12 18:29:24.825                   30.9        1           12               56             111.1                129.9     46.4    172.2
1 2026-04-12 18:56:45.308 2026-04-12 19:26:46.172                   30.0        2            6              193              54.9                 61.3     15.5     87.7
2 2026-04-12 19:28:04.520 2026-04-12 19:57:54.675                   29.8        3            4              190              37.7                 44.0     18.1     57.5
3 2026-04-12 20:16:20.974 2026-04-12 20:46:33.481                   30.2        4            3              183              29.5                 20.4     15.4     47.7
4 2026-04-12 20:47:35.819 2026-04-12 21:18:35.114                   31.0        6            2              183              20.3                 14.6      6.5     31.4
5 2026-04-12 21:35:54.023 2026-04-12 22:06:17.163                   30.4       12            1              156              11.8                 14.4      5.8     18.4
```

Looks like 2 threads, 6 in parallel is the optimal for my processor. Seems like 1 thread really doesn't give ffmpeg enough room to breathe.

I'll run this test again using only ffmpeg threads and MaxParallelism set to 1 and see if it beats threading and then I guess we'll have a conclusive answer as to which is best (or as I suspect, it makes no difference). Will have to be tomorrow tho because it took a while to get this done.

image

aidancrane:
Okay. Finished testing

```
            session_start             session_end  session_duration_mins  threads  parallelism  files_processed  avg_sec_per_file  median_sec_per_file  min_sec  max_sec
0 2026-04-13 08:21:23.696 2026-04-13 08:50:56.407                   29.5        1            1               30              60.1                 58.8     28.0     96.4
1 2026-04-13 08:52:28.130 2026-04-13 09:23:14.028                   30.8        2            1               61              30.5                 21.9     14.3     53.2
2 2026-04-13 09:24:00.837 2026-04-13 09:54:47.865                   30.8        3            1               80              23.5                 31.9     10.9     35.9
3 2026-04-13 10:05:28.727 2026-04-13 10:36:10.332                   30.7        4            1              103              18.1                 23.9      8.6     30.3
4 2026-04-13 10:46:11.997 2026-04-13 11:16:59.520                   30.8        6            1              137              13.6                 19.0      6.3     22.5
5 2026-04-13 12:40:31.845 2026-04-13 13:11:09.599                   30.6       12            1              160              11.5                 12.4      5.8     22.1
```

image

This is using just threading. It looks like in basically all tests, threading is faster looking at the average processing time per session, which makes sense.

But it also looks like, there is less gains between 4 and 12 threads, which is where parallelism shines and is able to take advantage, using the additional cores to work faster on processing jobs in parallel. Using just 2 threads half's processing time, but then performance gains adding threads drop off.

(looking at my previous chart) In basically all combinations except 1t/12p utilizing all 12 threads with parallelism is faster than a single threaded workload.

So I think this PR is beneficial.

Starkiller21321:
> Okay. Finished testing
> 
> ```
>             session_start             session_end  session_duration_mins  threads  parallelism  files_processed  avg_sec_per_file  median_sec_per_file  min_sec  max_sec
> 0 2026-04-13 08:21:23.696 2026-04-13 08:50:56.407                   29.5        1            1               30              60.1                 58.8     28.0     96.4
> 1 2026-04-13 08:52:28.130 2026-04-13 09:23:14.028                   30.8        2            1               61              30.5                 21.9     14.3     53.2
> 2 2026-04-13 09:24:00.837 2026-04-13 09:54:47.865                   30.8        3            1               80              23.5                 31.9     10.9     35.9
> 3 2026-04-13 10:05:28.727 2026-04-13 10:36:10.332                   30.7        4            1              103              18.1                 23.9      8.6     30.3
> 4 2026-04-13 10:46:11.997 2026-04-13 11:16:59.520                   30.8        6            1              137              13.6                 19.0      6.3     22.5
> 5 2026-04-13 12:40:31.845 2026-04-13 13:11:09.599                   30.6       12            1              160              11.5                 12.4      5.8     22.1
> ```
> 
> image
> This is using just threading. It looks like in basically all tests, threading is faster looking at the average processing time per session, which makes sense.
> 
> But it also looks like, there is less gains between 4 and 12 threads, which is where parallelism shines and is able to take advantage, using the additional cores to work faster on processing jobs in parallel. Using just 2 threads half's processing time, but then performance gains adding threads drop off.
> 
> (looking at my previous chart) In basically all combinations except 1t/12p utilizing all 12 threads with parallelism is faster than a single threaded workload.
> 
> So I think this PR is beneficial.

Thank you for helping test this out. Your post has been quite insightful, and more thorough than what I was attempting to put together.

mitzsch:
Is there a chance that we see this being implemented in v12?  :)

PR #16594, opened 2026-04-07
fix: retain subtitles spanning HLS segment boundaries

LiHRaM:
## Summary

- Fix `FilterEvents` in `SubtitleEncoder` dropping subtitles that span an HLS segment boundary
- The `SkipWhile` predicate used `||` (OR), which dropped any subtitle that merely *started* before the segment — even if it was still on screen. Changed to `&&` (AND) so subtitles are only skipped when fully elapsed
- This bug has been present since the initial port from Emby in December 2018

Partially addresses #2547

### DIFF
--- a/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
+++ b/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
@@ -103,9 +103,9 @@ namespace MediaBrowser.MediaEncoding.Subtitles
 
         internal void FilterEvents(SubtitleTrackInfo track, long startPositionTicks, long endTimeTicks, bool preserveTimestamps)
         {
-            // Drop subs that are earlier than what we're looking for
+            // Drop subs that have fully elapsed before the requested start position
             track.TrackEvents = track.TrackEvents
-                .SkipWhile(i => (i.StartPositionTicks - startPositionTicks) < 0 || (i.EndPositionTicks - startPositionTicks) < 0)
+                .SkipWhile(i => (i.StartPositionTicks - startPositionTicks) < 0 && (i.EndPositionTicks - startPositionTicks) < 0)
                 .ToArray();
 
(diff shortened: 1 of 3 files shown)


MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
gnattu: Although HLS webvtt may be not affected in this path, this will cause negative timestamp in other code paths.

gnattu (commented):
IIRC most clients do not ask the server for subtitle transcoding (maybe android tv is the most popular one that does) so this might not be as helpful as it seems.

And it is also weird for me that you believe the old behavior would cause desync. The old behavior will drop cues at the segment boundary, which is closer to missing subtitles rather than timestamp shifting that the users originally reported. 

The old behavior is still a bug though, but this is a minor fix which cannot address the timestamp shifting issue.

LiHRaM:
Point taken! I'll update the description - my thinking was that this contributes to the overall poor UX around subtitle desync, but it was misleading to say it _causes_ desync, as it's just a boundary condition.

gnattu (approved):

Shadowghost (approved):

tests/Jellyfin.MediaEncoding.Tests/Subtitles/FilterEventsTests.cs
```diff
@@ -0,0 +1,282 @@
+using System;
+using AutoFixture;
+using AutoFixture.AutoMoq;
+using MediaBrowser.MediaEncoding.Subtitles;
+using MediaBrowser.Model.MediaInfo;
+using Xunit;
+
+namespace Jellyfin.MediaEncoding.Subtitles.Tests
```
Bond-009, line 8: use file scoped namespace

Bond-009 (approved):

PR #16597, opened 2026-04-08
fix: use keyframe-aligned segments in subtitle playlists

LiHRaM:
**Changes**
Extract `GetSegmentDurations` from `CreateMainPlaylist`, and reuse in `SubtitleController.GetSubtitlePlaylist`.

**Issues**
Addresses https://github.com/jellyfin/jellyfin/issues/2547

---

Previously, subtitle playlists used fixed-length segments while video playlists used keyframe-aligned segments (when remuxing). After seeking, FFMpeg stats the video from the nearest keyframe, but the subtitle segment starts at a fixed boundary. This would cause the video and subtitles to be out of sync if the keyframe did not line up with the fixed boundary.

Tested locally by running an instance from master, seeking until subtitles were out of sync, restarting from this feature branch and seeing the subtitles align again with the video at the same timestamps as the videos were observed.


gnattu:
May I ask which client are you using to test this feature? as the most common client, the jellyfin-web, does not use HLS vtt as subtitle AFAIK.

LiHRaM:
Hey @gnattu, I'm using Android TV to test this feature

src/Jellyfin.MediaEncoding.Hls/Playlist/DynamicHlsPlaylistGenerator.cs
```diff
@@ -106,6 +104,18 @@ public string CreateMainPlaylist(CreateMainPlaylistRequest request)
         return builder.ToString();
     }
 
+    /// <inheritdoc />
+    public IReadOnlyList<double> GetSegmentDurations(Guid? mediaSourceId, string filePath, int desiredSegmentLengthMs, long totalRuntimeTicks)
+    {
+        if (mediaSourceId is not null
+            && TryExtractKeyframes(mediaSourceId.Value, filePath, out var keyframeData))
```
gnattu, line 104: I used to want to do this for subtitles as well but this operation is actually:
1. not that fast
2. not used for subtitles before

Is ad-hoc extracting fast enough on your system?

Jellyfin.Api/Controllers/SubtitleController.cs
```diff
@@ -368,6 +373,12 @@ public async Task<ActionResult> GetSubtitlePlaylist(
             throw new ArgumentException("segmentLength was not given, or it was given incorrectly. (It should be bigger than 0)");
         }
 
+        var segments = _playlistGenerator.GetSegmentDurations(
```
gnattu, line 378: For transcoded videos, the keyframe interval is different from the original one and that could cause desync as well. And if I understand your code correctly, we are not handling transcoded video? Or have you tried that the subtitle is still in sync after transcode? (I think it is not)

LiHRaM: You're right, let me see what I can do in that case.

EDIT: I've refactored the code to pass `isRemuxing` to `GetSegmentDurations`, so the logic should be identical on both sides.

src/Jellyfin.MediaEncoding.Hls/Playlist/DynamicHlsPlaylistGenerator.cs
LiHRaM: I've been testing the server on a much more powerful computer than I use for my homelab, not sure if it could handle it if this is heavy. :/

DrHurt:
This problem also affects Apple clients using AVPlayer (mkv to fmp4 + webvtt via HLS).

LiHRaM:
Closing as I haven't had time to properly polish and test with slower hardware.

DrHurt:
> Closing as I haven't had time to properly polish and test with slower hardware.

This does address a valid issue though! May I suggest you leave it open so that perhaps someone else takes it over!

There’s also another issue thats causing HLS subtitles to go out of sync which forces me to use subtitles method external with a custom renderer.

https://github.com/jellyfin/jellyfin/issues/16647

PR #16606, opened 2026-04-10
Fix artist metadata not being fetched on initial library scan

theguymadmax:
**Changes**
AudioDB artist metadata had several issues preventing it from populating correctly on initial scan or at all:

1. Recent requests to the AudioDB API return biography as `strBiography` instead of `strBiographyEN`. Added `strBiography` as a fallback when `strBiographyEN` is empty.

2. `AudioDbArtistProvider` ran before `MusicBrainzArtistProvider`, so the MusicBrainz ID needed for AudioDB lookup was always null on first scan. Fixed by implementing `IHasOrder` on `MusicBrainzArtistProvider` with `Order = 0`.

3. Artists derived from audio tags rather than a folder structure have no real library path, causing `GetLibraryOptions` to return null and AudioDB being filtered out of the provider list. Fixed by allowing all providers through when the item is a `MusicArtist` and `libraryTypeOptions` is null.

4. On the performance branch (yet to be merged), `ArtistsValidator` only refreshes new items. However, artists derived from audio tags failed this condition. Added a condition to refresh artists that have never been refreshed (`DateLastRefreshed == DateTime.MinValue`).

**Issues**
Fixes: #16599

### DIFF
--- a/Emby.Server.Implementations/Library/Validators/ArtistsValidator.cs
+++ b/Emby.Server.Implementations/Library/Validators/ArtistsValidator.cs
@@ -50,6 +50,10 @@ public class ArtistsValidator
     public async Task Run(IProgress<double> progress, CancellationToken cancellationToken)
     {
         var names = _itemRepo.GetAllArtistNames();
+        var existingArtistIds = _libraryManager.GetItemIds(new InternalItemsQuery
+        {
+            IncludeItemTypes = [BaseItemKind.MusicArtist]
+        }).ToHashSet();
 
         var numComplete = 0;
         var count = names.Count;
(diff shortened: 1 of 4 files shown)


Shadowghost (approved):

Emby.Server.Implementations/Library/Validators/ArtistsValidator.cs
```diff
@@ -59,8 +63,13 @@ public async Task Run(IProgress<double> progress, CancellationToken cancellation
             try
             {
                 var item = _libraryManager.GetArtist(name);
+                var isNew = !existingArtistIds.Contains(item.Id);
+                var neverRefreshed = item.DateLastRefreshed == DateTime.MinValue;
```
Bond-009:
```suggestion
                var neverRefreshed = item.DateLastRefreshed == default;
```

Bond-009 (approved):

PR #16616, opened 2026-04-10
fix person TotalRecordCount when limit is applied

dkanada:
**Changes**

Right now TotalRecordCount for the Persons API will always return the limit when present since we don't return a QueryResult with the required information. This breaks pagination on the web client.

I tried to limit the blast radius of changes for a 12.0 inclusion. We can't pass QueryResult<BaseItem> from PeopleRepository because a number of other locations expect PersonInfo. I have subsequent pull requests to add prefix filtering and maybe sort order.

**Issues**

None

### DIFF
--- a/Jellyfin.Server.Implementations/Item/PeopleRepository.cs
+++ b/Jellyfin.Server.Implementations/Item/PeopleRepository.cs
@@ -1,14 +1,13 @@
 using System;
 using System.Collections.Generic;
-using System.Collections.Immutable;
 using System.Linq;
 using Jellyfin.Data.Enums;
 using Jellyfin.Database.Implementations;
 using Jellyfin.Database.Implementations.Entities;
-using Jellyfin.Database.Implementations.Entities.Libraries;
 using Jellyfin.Extensions;
 using MediaBrowser.Controller.Entities;
 using MediaBrowser.Controller.Persistence;
+using MediaBrowser.Model.Querying;
 using Microsoft.EntityFrameworkCore;
 
 namespace Jellyfin.Server.Implementations.Item;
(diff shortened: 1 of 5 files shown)


Jellyfin.Server.Implementations/Item/PeopleRepository.cs
```diff
@@ -72,7 +58,18 @@ public IReadOnlyList<string> GetPeopleNames(InternalPeopleQuery filter)
...
+            StartIndex = filter.StartIndex ?? 0,
+            TotalRecordCount = count,
+            Items = dbQuery.AsEnumerable().Select(Map).ToArray(),
+        };
+    }
+
+    /// <inheritdoc/>
+    public IReadOnlyList<string> GetPeopleNames(InternalPeopleQuery filter)
+    {
+        return GetPeople(filter).Items.Select(e => e.Name).ToArray();
```
JPVenson: rather then pulling the full object here and then client side selecting the names we should do that as an EFQuery with a name projection instead
```suggestion        
        using var context = _dbProvider.CreateDbContext();
        var dbQuery = TranslateQuery(context.Peoples.AsNoTracking(), context, filter);
        var count = dbQuery.Count();
        return new QueryResult<PersonInfo>
        {
            StartIndex = filter.StartIndex ?? 0,
            TotalRecordCount = count,
            Items = dbQuery.Select(e => e.Name).ToArray(),
        };
```

JPVenson (changes requested):

Jellyfin.Server.Implementations/Item/PeopleRepository.cs
dkanada: I had originally done this to consolidate the filtering et cetera into a single method but I can see the appeal of performance. Swapped it back to the original behavior.

Shadowghost (approved):

JPVenson (approved):

PR #16630, opened 2026-04-13
Fix DirectoryNotFoundException aborting library scan on missing folders

### LINKED ISSUE #16628: Library scan aborts on DirectoryNotFoundException, preventing new media discovery
## Description

When a media folder has been deleted from disk but its entry remains in the Jellyfin database, the library scan throws a `DirectoryNotFoundException` that causes the entire scan to abort early. As a result, new media added to the library is never discovered until the orphaned database entry is manually removed.

## Steps to Reproduce

1. Have a show/movie folder indexed in Jellyfin
2. Delete the folder from disk without removing it from Jellyfin first
3. Trigger a library scan (manual or scheduled)
4. Add new media folders to the library path
5. Observe that the new media never appears in Jellyfin

## Expected Behavior

The library scan should handle a missing folder path gracefully — log a warning, mark the orphaned item for removal, and continue scanning the rest of the library normally.

## Actual Behavior

In `Folder.ValidateChildrenInternal2()`, when `GetNonCachedChildren()` is called for a folder whose path no longer exists on disk, a `DirectoryNotFoundException` is thrown. This is not caught by the `IOException` or `SecurityException` handlers (since `DirectoryNotFoundException` inherits from `SystemException`, not `IOException` in this context) and falls through to the generic `catch (Exception ex)` block which executes `return`, exiting the entire validation method immediately.

The result:
- Scan completes in seconds regardless of library size (the fast completion is a symptom)
- Orphaned DB entries are logged as "Removing item" but are never actually deleted from the database (the removal code is never reached due to the early `return`)
- New media directories on disk are never discovered
- The scheduled task always reports "Completed" masking the failure entirely

## Root Cause

`MediaBrowser.Controller/Entities/Folder.cs` — `ValidateChildrenInternal2()`:

```csharp
try
{
    nonCachedChildren = GetNonCachedChildren(directoryService);
}
catch (IOException ex)
{
    Logger.LogError(ex, "Error retrieving children from file system");
}
catch (SecurityException ex)
{
    Logger.LogError(ex, "Error retrieving children from file system");
}
catch (Exception ex)
{
    Logger.LogError(ex, "Error retrieving children");
    return;  // <-- exits the entire method, skipping all new item discovery
}
```

`DirectoryNotFoundException` falls through to the catch-all and the `return` prevents:
- The filesystem vs database comparison that would flag the item for removal
- Discovery of any new items on disk
- Recursive validation of sibling folders

## Proposed Fix

Add a specific `DirectoryNotFoundException` catch that sets `nonCachedChildren` to an empty collection rather than returning. This allows the existing logic to naturally flag the orphaned item for removal and the scan to continue:

```csharp
catch (DirectoryNotFoundException ex)
{
    Logger.LogWarning(ex, "Directory not found during scan, marking for removal: {Path}", ContainingFolderPath);
    nonCachedChildren = [];
}
catch (IOException ex)
{
    Logger.LogError(ex, "Error retrieving children from file system");
}
catch (SecurityException ex)
{
    Logger.LogError(ex, "Error retrieving children from file system");
}
catch (Exception ex)
{
    Logger.LogError(ex, "Error retrieving children");
    return;
}
```

## Environment

- Jellyfin version: 10.11.6
- OS: Ubuntu 22.04 (Linux aarch64)
- Library backend: NTFS via ntfs-3g

## Additional Context

Workaround: manually delete the orphaned entry and all child records directly from `jellyfin.db` via SQLite, then re-run the library scan.

ryan-rbw:
When a folder is deleted from disk while its entry remains in the database, `GetNonCachedChildren` throws `DirectoryNotFoundException`. This exception is not a subclass of `IOException` or `SecurityException`, so it falls through to the generic catch block which executes an early `return`, preventing the orphaned entry from being cleaned up and blocking discovery of any new media on disk.

Fix by catching `DirectoryNotFoundException` before `IOException` and leaving `nonCachedChildren` as an empty enumerable, allowing the existing comparison logic to naturally remove the orphaned entry.

Fixes #16628

### DIFF
--- a/MediaBrowser.Controller/Entities/Folder.cs
+++ b/MediaBrowser.Controller/Entities/Folder.cs
@@ -390,6 +390,13 @@ namespace MediaBrowser.Controller.Entities
                 {
                     nonCachedChildren = GetNonCachedChildren(directoryService);
                 }
+                catch (DirectoryNotFoundException ex)
+                {
+                    // The folder no longer exists on disk. Use an empty list so the existing
+                    // database entry is naturally flagged for removal by the comparison below,
+                    // and the rest of the library scan can continue uninterrupted.
+                    Logger.LogWarning(ex, "Directory not found during scan, marking for removal: {Path}", Path);
+                }
                 catch (IOException ex)
                 {
                     Logger.LogError(ex, "Error retrieving children from file system");
--- a/tests/Jellyfin.Controller.Tests/Entities/FolderTests.cs
+++ b/tests/Jellyfin.Controller.Tests/Entities/FolderTests.cs
@@ -1,0 +1,90 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using MediaBrowser.Controller.Entities;
+using MediaBrowser.Controller.Library;
+using MediaBrowser.Controller.LibraryTaskScheduler;
+using MediaBrowser.Controller.Providers;
+using MediaBrowser.Model.IO;
+using MediaBrowser.Model.MediaInfo;
+using Microsoft.Extensions.Logging.Abstractions;
+using Moq;
+using Xunit;
+
+namespace Jellyfin.Controller.Tests.Entities;
+
+public class FolderTests
+{
+    /// <summary>
+    /// Verifies that a <see cref="DirectoryNotFoundException"/> thrown from
+    /// <see cref="Folder.GetNonCachedChildren"/> does not abort the library scan.
+    ///
+    /// Previously the exception fell through to a catch-all handler that executed
+    /// an early <c>return</c>, preventing the orphaned database entry from being
+    /// cleaned up and blocking discovery of any new media on disk.
+    /// </summary>
+    [Fact]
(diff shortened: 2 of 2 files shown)


MediaBrowser.Controller/Entities/Folder.cs
```diff
@@ -390,6 +390,13 @@ private async Task ValidateChildrenInternal2(IProgress<double> progress, bool re
                 {
                     nonCachedChildren = GetNonCachedChildren(directoryService);
                 }
+                catch (DirectoryNotFoundException ex)
```
gnattu, line 393: `DirectoryNotFoundException` is a subclass of `IOException` in dotnet: https://learn.microsoft.com/en-us/dotnet/api/system.io.directorynotfoundexception?view=net-10.0

Which means the change here is effectively a no-op. Your unit test passes even without your change.

theguymadmax:
We already have a fix https://github.com/jellyfin/jellyfin/pull/16062/changes/4b48cad6f79123c25bb3c15c006afd27f0b49e45 for this issue that will be included in Jellyfin 12.0,

JPVenson:
closed as per max comment

PR #16655, opened 2026-04-17
Round HLS output dimensions to even display resolution

LTe:
**Changes**

Instead of predicting dimensions in C#, ask ffmpeg directly. A small helper spawns ffprobe against a synthetic `color=size=WxH,<scale-filter>` source and reads the resulting stream dims. Whatever ffprobe says is, by definition, what the real transcode will produce.

The probed dims are stored on two new nullable fields on the streaming state; the existing output-width/height properties prefer them when set and fall back to the legacy math otherwise. Downstream consumers (playlist writer, transcode manager) are unchanged — they read the same properties as before and now get accurate numbers.

Handling around the probe:

- **SW encoders:** build the actual scale filter and probe it.
- **HW accelerators:** the HW scale filter bakes literal dimensions in at generation time; we read those directly rather than spawning ffprobe.
- **Rotation:** source width/height swapped for `±90°` to mirror the transpose step in the real pipeline.
- **Anamorphic content:** sample aspect ratio is now persisted at scan time (new database column) and forwarded to the probe so mjpeg's `(a*sar)` expression resolves correctly for non-square pixels.

**Issues**

The HLS master playlist declares `RESOLUTION=720x405` for content that transcodes to 720x404 actual pixels. `mediastreamvalidator` catches this as `Playlist video resolution doesn't match parsed video resolution`.

Steps to reproduce

1. Request PlaybackInfo

```bash
BASE="http://localhost:8096"
TOKEN="<your-api-token>"
USER_ID="<your-user-id>"
ITEM="<your-item-id>"

RESP=$(curl -s -X POST \
    -H "Authorization: MediaBrowser Token=\"${TOKEN}\"" \
    -H "Content-Type: application/json" \
    "${BASE}/Items/${ITEM}/PlaybackInfo?UserId=${USER_ID}&MaxStreamingBitrate=3000000&IsPlayback=true&AutoOpenLiveStream=true" \
    -d '{
        "DeviceProfile":{
            "MaxStreamingBitrate":3000000,
            "DirectPlayProfiles":[],
            "TranscodingProfiles":[{
                "Container":"ts","Type":"Video",
                "VideoCodec":"h264","AudioCodec":"aac","Protocol":"hls",
                "Conditions":[{"Condition":"LessThanEqual","Property":"Width","Value":"720","IsRequired":true}]
            }],
            "CodecProfiles":[{
                "Type":"Video","Codec":"h264",
                "Conditions":[{"Condition":"LessThanEqual","Property":"Width","Value":"720","IsRequired":true}]
            }],
            "SubtitleProfiles":[],"ResponseProfiles":[]
        }
    }')

URL="${BASE}$(echo "${RESP}" | jq -r '.MediaSources[0].TranscodingUrl')"
echo "${URL}"
```

Output is a URL like:
```
http://localhost:8096/videos/c8d3b0dd-.../master.m3u8?DeviceId=e2e-001&MediaSourceId=...&VideoCodec=h264&AudioCodec=aac&VideoBitrate=2872187&AudioBitrate=127813&MaxFramerate=24&MaxWidth=720&SegmentContainer=ts&PlaySessionId=...&ApiKey=...
```

2. Fetch the master playlist

```bash
curl -s "${URL}"
```

Output — **note `RESOLUTION=720x405`**:

```
#EXTM3U
#EXT-X-STREAM-INF:BANDWIDTH=3000000,AVERAGE-BANDWIDTH=3000000,VIDEO-RANGE=SDR,CODECS="avc1.42401F,mp4a.40.2",RESOLUTION=720x405,FRAME-RATE=24
main.m3u8?DeviceId=e2e-001&MediaSourceId=...&MaxWidth=720&...
```

3. Run mediastreamvalidator

```bash
mediastreamvalidator "${URL}"
```

Output — **the resolution mismatch is flagged**:

```
Processed 10 out of 10 segments
Playlist max bitrate: 3000.000000 kb/s
Video resolution: 718x404
Error: Playlist video resolution doesn't match parsed video resolution
```

This can be set in jellyfin-web in Avatar > Profile > Playback > Maximum Allowed Video Transcoding Resolution and set to Display Resolution


gnattu:
For this specific use case this is not the best approach IMO.

The actual output dimension is actually very encoder and input dependent, we calculate output dimension differently on different code paths

See https://github.com/jellyfin/jellyfin/blob/1e64e4284224949c3029ec8f1e3bc66e364da6dd/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs#L3374

and 
https://github.com/jellyfin/jellyfin/blob/1e64e4284224949c3029ec8f1e3bc66e364da6dd/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs#L3239

The problem here is that the master playlist is generated **before** the ffmpeg helpers are even called. So current behavior is more of a design limitation, because at the time we create HLS master playlist, we do not know which path would actually be taken, and the actual size would very like be different from the requested size due to normalizations.

It is indeed a spec-compliant problem but this PR is probably not doing in the most correct way. Our playback info api and HLS master playlist api both have this limitation which makes the clients can never accurately get the accurate info of the output stream and the actual encoding method before the playback is started.

If you do have intention to make it better, I suggest you to start with the playbackinfo api so that we can have all the information required before the playback is started, and then we can use the information generated from the actual playback method to make the master playlist.

LTe:
@gnattu I was looking at the files and I can see couple ways that this can be solved.

1. Mimic ffmpeg math inside C# (can drift when someone will change filter or add new one or ffmpeg behavior will change)
2. Run real ffmpeg and produce one segment and probe that (slower, need to spawn process but always right). We can even ask ffmpeg to create playlist file for us with one segment and get resolution from that.

I'm wondering if it is hard rule that we should generate playlist without touching ffmpeg. I guess this can be problematic for NFS shares or some low performance hardware. But on other hand when client will ask for playlist in the next request we will start transcoding anyway.

MediaBrowser.Controller/MediaEncoding/EncodingJobInfo.cs
```diff
@@ -542,6 +533,31 @@ public string GetMimeType(string outputPath, bool enableStreamDefault = true)
...
+            var size = new ImageDimensions(VideoStream.Width.Value, VideoStream.Height.Value);
+
+            var newSize = DrawingUtils.Resize(
+                size,
+                BaseRequest.Width ?? 0,
+                BaseRequest.Height ?? 0,
+                BaseRequest.MaxWidth ?? 0,
+                BaseRequest.MaxHeight ?? 0);
+
+            return newSize.Height / 2 * 2;
```
IDisposable: Should we be rounding up here?

```csharp
return ((newSize.Height + 1) / 2) * 2;
```

MediaBrowser.Controller/MediaEncoding/EncodingJobInfo.cs
```diff
@@ -542,6 +533,31 @@ public string GetMimeType(string outputPath, bool enableStreamDefault = true)
...
+
+        /// <summary>
+        /// Computes the display width for a given output height, preserving the source aspect ratio.
+        /// </summary>
+        private static int ComputeOutputDisplayWidth(int height, int sourceWidth, int sourceHeight, string sampleAspectRatioString)
+        {
+            // Display aspect ratio = storage ratio × sample aspect ratio (pixel shape).
+            var storageRatio = new AspectRatio(sourceWidth, sourceHeight);
+            var sourceDisplayRatio = storageRatio.Compose(AspectRatio.Parse(sampleAspectRatioString));
+            return sourceDisplayRatio.WidthFor(height) / 2 * 2;
```
IDisposable: (likewise, seems like we should be rounding up to the next multiple of two, not dropping a pixel)

MediaBrowser.Controller/MediaEncoding/EncodingJobInfo.cs
LTe: Good question, but floor is correct — ffmpeg's scale uses `trunc(…/2)*2`, so rounding up would make the manifest disagree with the segments in the opposite direction.

MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
```diff
@@ -707,7 +707,7 @@ private async Task<string> ExtractImageInternal(
                 // fsbs crop width in half,set the display aspect,crop out any black bars we may have made
                 Video3DFormat.FullSideBySide => @"crop=iw/2:ih:0:0,setdar=dar=a,crop=min(iw\,ih*dar):min(ih\,iw/dar):(iw-min(iw\,iw*sar))/2:(ih - min (ih\,ih/sar))/2,setsar=sar=1",
                 // htab crop height in half,scale to correct size, set the display aspect,crop out any black bars we may have made
-                Video3DFormat.HalfTopAndBottom => @"crop=iw:ih/2:0:0,scale=(iw*2):ih),setdar=dar=a,crop=min(iw\,ih*dar):min(ih\,iw/dar):(iw-min(iw\,iw*sar))/2:(ih - min (ih\,ih/sar))/2,setsar=sar=1",
+                Video3DFormat.HalfTopAndBottom => @"crop=iw:ih/2:0:0,scale=(iw*2):ih,setdar=dar=a,crop=min(iw\,ih*dar):min(ih\,iw/dar):(iw-min(iw\,iw*sar))/2:(ih - min (ih\,ih/sar))/2,setsar=sar=1",
```
LTe, line 712: While reviewing the 3D filter chains I found a stray `)` in the HalfTopAndBottom filter string, sitting there since November 2013. ffmpeg rejects it with `Invalid chars ')' at the end of expression 'ih)'`. The same broken chain is duplicated in the transcoding helper and the image-extraction helper — both are corrected. This silently broke thumbnail extraction for every HTAB-tagged file during library scan.

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -3502,7 +3502,7 @@ private static string GetFixedSwScaleFilter(Video3DFormat? threedFormat, int req
                         // fsbs crop width in half,set the display aspect,crop out any black bars we may have made the scale width to requestedWidth.
                         break;
                     case Video3DFormat.HalfTopAndBottom:
-                        filter = @"crop=iw:ih/2:0:0,scale=(iw*2):ih),setdar=dar=a,crop=min(iw\,ih*dar):min(ih\,iw/dar):(iw-min(iw\,iw*sar))/2:(ih - min (ih\,ih/sar))/2,setsar=sar=1,scale={0}:trunc({0}/dar/2)*2";
+                        filter = @"crop=iw:ih/2:0:0,scale=(iw*2):ih,setdar=dar=a,crop=min(iw\,ih*dar):min(ih\,iw/dar):(iw-min(iw\,iw*sar))/2:(ih - min (ih\,ih/sar))/2,setsar=sar=1,scale={0}:trunc({0}/dar/2)*2";
```
LTe, line 3574: Ditto https://github.com/jellyfin/jellyfin/pull/16655/changes#r3111079935

gnattu:
Before we move on I want to clarify some misunderstanding:

NO, we do not use ffmpeg's internal normalizations(at least not in all cases) because that is not reliable and actually, not automatic for all encoders. Some encoder will simply reject the input if the dimension is not compatible with the chroma. That is why we have `GetSwScaleFilter/GetHwScaleFilter` which calculates the fixed output size if needed and we **should** use the result of those helpers instead, which means the discussion of how we should normalize the dimension here, is irrelevant, as it will be different from what we actually tells ffmpeg to resize to.

That is why I said the design limitation earlier, because those size calculations are part of the ffmpeg cli building step and will not be called until the HLS stream is requested and ffmpeg is going to be launched.

LTe:
`GetHwScaleFilter` emits filters with literal integer dims (`scale_vaapi=W=1280:H=720`) - dims are computed ahead of time via `GetFixedOutputSize`. `GetSwScaleFilter` emits expression-based filters (`trunc(min(max(iw,ih*a),MW)/2)*2:…`) that ffmpeg evaluates at transcode time.

What if we made SW mirror HW — compute dims in C#, emit literal-integer scale strings? No ffprobe, no reproducing AVExpression semantics in C#, and the helper naturally returns (filter, W, H) — same shape HW already has.

Issue #13568, opened 2025-02-17
Live TV: Schedules Direct EPG Image Pre-Cache Fails Due To image/jpg MIME Type Not Recognized


arazilsongweaver:
### Description of the bug

Schedules Direct electronic program guide (EPG) images fail to download with the following MIME type error message:

```
[2025-02-17 01:00:37.154 +00:00] [WRN] [38] Jellyfin.LiveTv.Guide.GuideManager: Unable to pre-cache "https://json.schedulesdirect.org/20141201/image/30663d7a6c97bfa29442d23a1834772289c8475aaab1ebf1cac4a1de9bd0e14b.jpg?token=TOKEN"
System.ArgumentException: Unable to determine image file extension from mime type image/jpg
   at MediaBrowser.Providers.Manager.ImageSaver.GetStandardSavePath(BaseItem item, ImageType type, Nullable`1 imageIndex, String mimeType, Boolean saveLocally)
   at MediaBrowser.Providers.Manager.ImageSaver.GetSavePaths(BaseItem item, ImageType type, Nullable`1 imageIndex, String mimeType, Boolean saveLocally)
   at MediaBrowser.Providers.Manager.ImageSaver.SaveImage(BaseItem item, Stream source, String mimeType, ImageType type, Nullable`1 imageIndex, Nullable`1 saveLocallyWithMedia, CancellationToken cancellationToken)
   at MediaBrowser.Providers.Manager.ProviderManager.SaveImage(BaseItem item, String url, ImageType type, Nullable`1 imageIndex, CancellationToken cancellationToken)
   at MediaBrowser.Providers.Manager.ProviderManager.SaveImage(BaseItem item, String url, ImageType type, Nullable`1 imageIndex, CancellationToken cancellationToken)
   at Emby.Server.Implementations.Library.LibraryManager.ConvertImageToLocal(BaseItem item, ItemImageInfo image, Int32 imageIndex, Boolean removeOnFailure)
   at Jellyfin.LiveTv.Guide.GuideManager.<>c__DisplayClass24_0.<<PreCacheImages>b__2>d.MoveNext()
```

This pre-caching issue prevents Schedules Direct images from being used in the EPG.

### Reproduction steps

1. Set up the Schedules Direct electronic program guide (EPG) source.
2. Attach Schedules Direct to a Live TV source.
3. Failure occurs when Jellyfin downloads and updates the EPG. (see jellyfin.log)

### What is the current _bug_ behavior?

[...]

PR #16658, opened 2026-04-18
fix: add image/jpg to MIME type extension lookup

jskoetsier:
## Problem

Several external metadata providers (TMDb, Schedules Direct, and others) return `image/jpg` as the Content-Type for JPEG images instead of the standard `image/jpeg` (RFC 2046).

When Jellyfin encounters this non-standard MIME type during library scans, it throws:
```
System.ArgumentException: Unable to determine image file extension from mime type image/jpg
```

This causes:
- Episode thumbnail downloads to fail silently
- Library scans to accumulate errors and eventually be cancelled
- Repeated scan failures every ~60 minutes on large libraries

## Fix

Add `image/jpg` → `.jpg` to the `_extensionLookup` FrozenDictionary in `MimeTypes.cs`.

## Context

- PR #7052 previously added this mapping but it was lost during the migration to FrozenDictionary
- Issue #13568 reports the same bug in 10.10+
- Issue #7050 was the original report

While `image/jpg` is not a registered MIME type, multiple widely-used providers return it, making this a practical necessity.

## Testing

- Library scan that previously crashed after ~60 minutes now completes successfully
- Episode images that previously failed with `image/jpg` errors now save correctly

### DIFF
--- a/MediaBrowser.Model/Net/MimeTypes.cs
+++ b/MediaBrowser.Model/Net/MimeTypes.cs
@@ -132,6 +132,7 @@ namespace MediaBrowser.Model.Net
 
             // Type image
             new("image/jpeg", ".jpg"),
+            new("image/jpg", ".jpg"),
             new("image/tiff", ".tiff"),
             new("image/x-png", ".png"),
             new("image/x-icon", ".ico"),


Shadowghost (approved):
Please rebase on current master and check the tests.

gnattu (approved):

JPVenson (approved):

Bond-009 (approved):
Please also contact TMDb, Schedules Direct, and others to let them know that they're sending an invalid content type. We'll be less strict and accept it, but that doesn't make it correct.

PR #16664, opened 2026-04-19
add experimental settings and require elevated access for certain routes

### LINKED ISSUE #13984: Video streams unauthenticated (VideosController, HlsSegmentController, LiveTVController)
Summary: The current implementation allows direct play without authentication. Transcoded media is partially authenticated (old HLS only has metadata protection, new HLS protects all). The Jellyfin web client does provide an api_key parameter in the URL for all some stream types (except HlsSegmentController because the m3u playlist generated doesn't contain parameters), but it isn't checked.
HlsSegmentController has a comment saying "Can't require authentication just yet due to seeing some requests come from Chrome without full query string"
This also applied to streams in LiveTvController endpoints. Those are also not authenticated. Update: Appears to be resolved in 10.8.9
Potential fix: Require authentication on all video streams, both direct streams as well as any transcoded media.

### LINKED ISSUE #13986: All endpoints in AudioController are unauthenticated
Split from #5415 

Has a comment saying "TODO: In order to authenticate this in the future, Dlna playback will require updating"
Currently affects "GetAudioStream()" and "GetAudioStreamByContainer()" endpoints. Anyone with knowledge about item ID's can retrieve audio from the server, even if not logged in.
Potential fix: Authenticate endpoints

### LINKED ISSUE #13987: Most endpoints in DashboardController are unauthenticated
split from #5415 

Partially authenticated since 10.8b1

As of 10.7, this controller apparently contains admin dashboard (plugin) configuration pages. While the pages alone are not highly sensitive by itself (the controller only returns the list & html pages without config data), the endpoints being completly unauthenticated allows anyone to retrieve information about the server, e.g this leaks the installed plugins.
Potential fix: Require admin privileges (at least user privileges) to access these pages

dkanada:
**Changes**

Once a decision has been made here I can open a corresponding pull request in the web client. I have a few more ideas for experimental settings but I wanted to start small. Each of these four endpoint groups have existing use cases at their current access levels and would benefit from a long migration window, hence the experimental flag.

Assets: Any plugins that use data-controller or anonymously access their assets will break with this change.
Configuration: Not 100% certain but I believe the Live TV settings are retrieved by users.
Media: Now that DLNA has been migrated to a plugin I believe we can require that media streams include authentication.

If any of these are contested I would rather remove them from the pull request and focus on the experimental settings endpoint itself as a feature. Even if a decision is made that some of these endpoints should keep their current access by default, I would argue that all of them should exist as flags for server hardening.

**Issues**

- replace #16292
- fix #13987
- fix #13986
- partially fix #13984


JPVenson (commented):
overall looks fine to me, i would like to discuss the benefit of allowing to lock down those yet now. IIRC we discussed and agreed on a future auth system where api tokes can be created with a scoped context for those instead of adding full auth.

What we need however is a lot more verbose logging in the ConditionalAuthHandler for support purposes to very quickly identify issues with that.

JPVenson (approved):

PR #16668, opened 2026-04-19
Fix actor images not displayed until clicked

### LINKED ISSUE #8103: [Issue]: Displaying actor photos with local metadata
### Please describe your bug

When you install a fresh new Jellyfin server, define all the metadata to be stored locally and create a Jellyfin library based on an existing folder with multiple files, the picture of the actors doesn't show in the interface, even when the scan is 100% finish.

I tried with a Debian server and a Synology server on Docker, on the web interface of Jellyfin or the Android TV app, the problem is the same.

However, if I click on the actor, his photo is displayed correctly.
If I go back to the film, the thumbnails display well, and will also be on next connections:

image

image

image

When I add a new movie in the library (after that Jellyfin was installed), the problem doesn't occur. All the actors that play in the new movies have their profile picture correctly loaded without having to click on it.

### Jellyfin Version

Other

### if other:

10.8.1

### Environment

```markdown
- OS:Debian and DSM
- Virtualization:N/A and Docker
- Clients:Android TV
- Browser:Google Chrome 103
- Hardware Acceleration:None
- Plugins:None
- Reverse Proxy:None
- Networking:Host
- Storage:local
```

### Jellyfin logs

_No response_

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

johnpc:
**Changes**

The "Refresh People" scheduled task used ValidationOnly mode, which skips all remote providers, so it never fetched images or metadata from TMDB. 

Clicking an actor is the only way to populate the image data, because the UI triggers a `FullRefresh`. 

The fix uses `FullRefresh` for people missing an image or overview, with a 30-day cooldown so TMDB isn't hammered for people it has no data for. The 30 day cooldown is also nice, because tmbd may add data over time.

I tested this with my own library:

1. Stock Jellyfin (11.10.8): Ran "Refresh People" -> 34,845 people missing images before, 34,845 after (Zero change).
2. Patched Jellyfin with this PR. Ran "Refresh People" -> 34,845 missing before, 23,805 missing after. ** 11,040 people got images.** (The remaining without images are more obscure people that TMDB doesn't have a photo for)

**Issues**

Fixes #8103

### DIFF
--- a/Emby.Server.Implementations/ScheduledTasks/Tasks/PeopleValidationTask.cs
+++ b/Emby.Server.Implementations/ScheduledTasks/Tasks/PeopleValidationTask.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Jellyfin.Database.Implementations;
+using Jellyfin.Database.Implementations.Entities;
 using MediaBrowser.Controller.Entities;
 using MediaBrowser.Controller.Library;
 using MediaBrowser.Controller.Providers;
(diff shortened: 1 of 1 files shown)


IDisposable:
Should we honor the library-specific setting about metadata refresh frequency?
image

johnpc:
This is a good point. However `Person` items can't be directly associated with a single library. 

The on-demand path uses `TotalDays >= 3` (https://github.com/jellyfin/jellyfin/blob/master/Jellyfin.Api/Controllers/UserLibraryController.cs#L638) so I could change to match that but it feels a bit aggressive.

JPVenson (commented):
I am not not agreeing with that this is an issue, but the peoplevalidator currently gets Hammered during scan so this could potentially massively add to the scan times which is already an pain point.

Is maybe a scheduled task a better solution? Can you please test the impact of the change?

johnpc:
> Can you please test the impact of the change?

I agree it shouldn't run during library scans.

> Is maybe a scheduled task a better solution?

The scheduled task already exists as "Refresh People". I will update this to target that task without also leaking to other callers.

johnpc:
Hey @JPVenson - just checking in to see if this updated strategy moving the logic directly into the Refresh People task makes sense to you!

JPVenson (approved):

Emby.Server.Implementations/ScheduledTasks/Tasks/PeopleValidationTask.cs
```diff
@@ -125,5 +143,63 @@ await context.PeopleBaseItemMap.WhereOneOrMany(dups, e => e.PeopleId)
...
+                var item = _libraryManager.GetPerson(person);
+                if (item is null)
+                {
+                    continue;
+                }
+
+                var hasImage = item.HasImage(ImageType.Primary, 0);
+                var hasOverview = !string.IsNullOrWhiteSpace(item.Overview);
+
+                if ((hasImage && hasOverview) || (DateTime.UtcNow - item.DateLastRefreshed).TotalDays < 30)
```
cvium: Couldn't this filtering be done on the database?

johnpc: Good call! Just added a commit to do it that.

Emby.Server.Implementations/ScheduledTasks/Tasks/PeopleValidationTask.cs
```diff
@@ -125,5 +144,74 @@ await context.PeopleBaseItemMap.WhereOneOrMany(dups, e => e.PeopleId)
...
+                cancellationToken.ThrowIfCancellationRequested();
+
+                try
+                {
+                    if (_libraryManager.GetItemById(personId) is not Person item)
+                    {
+                        continue;
+                    }
+
+                    var hasImage = item.HasImage(ImageType.Primary, 0);
```
JPVenson: I mean in that case and if we are already doing a DB query we can also just look into the DB and see if it has images

JPVenson (changes requested):

Emby.Server.Implementations/ScheduledTasks/Tasks/PeopleValidationTask.cs
johnpc: great point, updated it

johnpc:
@JPVenson just checking in for a re-review, I updated it so that it's all moved to db queries

Emby.Server.Implementations/ScheduledTasks/Tasks/PeopleValidationTask.cs
```diff
@@ -125,5 +143,76 @@ await context.PeopleBaseItemMap.WhereOneOrMany(dups, e => e.PeopleId)
...
+
+        // Phase 3: Refresh images for people missing them (66-100%)
+        subProgress = new Progress<double>((val) => progress.Report((val / 3) + 66));
+        await RefreshPeopleImagesAsync(subProgress, cancellationToken).ConfigureAwait(false);
+    }
+
+    private async Task RefreshPeopleImagesAsync(IProgress<double> progress, CancellationToken cancellationToken)
+    {
+        var thirtyDaysAgo = DateTime.UtcNow.AddDays(-30);
+        var personTypeName = typeof(Person).FullName!;
```
JPVenson: Dont use the type directly here, we have the https://github.com/jellyfin/jellyfin/blob/master/MediaBrowser.Controller/Persistence/IItemTypeLookup.cs#L11 for that

Emby.Server.Implementations/ScheduledTasks/Tasks/PeopleValidationTask.cs
```diff
@@ -125,5 +143,76 @@ await context.PeopleBaseItemMap.WhereOneOrMany(dups, e => e.PeopleId)
...
+                .ToListAsync(cancellationToken)
+                .ConfigureAwait(false);
+
+            var numPeople = people.Count;
+            var numComplete = 0;
+            var numRefreshed = 0;
+
+            _logger.LogDebug("Found {Count} people needing image/overview refresh", numPeople);
+
+            foreach (var entry in people)
```
JPVenson: Better use the pagination helper https://github.com/jellyfin/jellyfin/blob/master/src/Jellyfin.Database/Jellyfin.Database.Implementations/QueryPartitionHelpers.cs#L25C59-L25C80 it also has progress reporting build in

JPVenson (changes requested):
otherwise looks good

johnpc:
Thanks @JPVenson addressed both in [6da6895](https://github.com/jellyfin/jellyfin/pull/16668/commits/6da6895bc660bc936aaa7a5343914ffc2acf688e)

JPVenson (approved):

johnpc:
@cvium do you have additional thoughts?

johnpc:
Please let me know if there are any other merge-blockers, happy to jump in with adjustments as desired by maintainers.

Shadowghost (commented):
Can we make use of QueryPartitionHelpers here?

johnpc:
What do you have in mind @Shadowghost?

I am already using [PartitionEagerAsync](https://github.com/jellyfin/jellyfin/pull/16668/changes#diff-78f714b2321bd0cb62254dfc88889b8ba7583288ac8caca58bf58d24ef7842ffR213-R214) when looping over Person items missing a primary image.

theguymadmax:
I moved the `ValidatePeopleAsync` to run after the Orphan cleanup because when it runs before, the PeopleBaseItemMap is still populated with orphaned entries, so `IsDeadPerson` won't remove the person entity from  BaseItems. 

This prevents it from needing two scans for a proper cleanup:
 1st) cleans the PeopleBaseItemMap.
 2nd) removes the person entity and metadata

johnpc:
~@theguymadmax I'm not sure I follow, would you like me to wait to merge this until after your changes are merged, so I can address any conflict? Or should I change something about the implementation I have here?~

Nevermind, I see https://github.com/jellyfin/jellyfin/pull/17140 has already been merged. I have rebased origin/master and updated this PR

Shadowghost (approved):

crobibero (approved):

Issue #15352, opened 2025-11-02
[10.11.1] API Performance Degradation / Slow UI


verybadsoldier:
### Description of the bug

After upgrading from 10.10.9 to 10.11.1, I instantly experienced a slow and sluggish UI. This is when browsing my library, which consists of movies and music.

To narrow it down, I tested with Postman just a single API endpoint `MusicGenres`:
`https://myJellyfin/MusicGenres`

On 10.11.1 the call takes nearly 7 seconds:

<img width="1732" height="874" alt="Image" src="https://github.com/user-attachments/assets/4103b6e9-bd65-4d79-8cd0-0fc81b2d0519" />

The same call takes on 10.10.9 roughly 500 ms:

<img width="1692" height="856" alt="Image" src="https://github.com/user-attachments/assets/5b325985-bd94-42bc-8f0f-03ea07b65b90" />

My music library contains roughly 4000 albums.

### Reproduction steps

1. Open Postman
2. Send a HTTP GET query to `https://<myDomain>/MusicGenres`
3. Check the response time

### What is the current _bug_ behavior?

UI and API response time slow.

### What is the expected _correct_ behavior?

Similar UI and API responsiveness as in 10.10.x

### Jellyfin Server version

Master

### Specify the build version

10.11.1

### Environment

```markdown
- OS: Docker on Ubuntu Server
- Linux Kernel: Linux minion 6.8.0-63-generic #66-Ubuntu SMP PREEMPT_DYNAMIC Fri Jun 13 20:25:30 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
- Virtualization: yes, VMWare Linux guest on Windows 11 Host
```

### Jellyfin logs

```shell
There is no log output when sending the API call above.
```

### FFmpeg logs

```shell

```

### Additional information

I know there are reports about performance problems on 10.11.1 but I think this one using the MusicGenre-API is really easy to reproduce and easy to compare with 10.10.x

theinfguy:
Am also noticing very slow behavior when scanning music library through external apps such as Jellyfin for Kodi. Maybe related.

anthopare16:
I am also have the same behaviour.

theinfguy:
Can confirm extremely slow Music scanning via Jellyfin for Kodi add-on after updated to 10.11.2 still. It brings the Jellyfin Web UI to the point of being unusable and usually ends in a database lock that requires restart.

[...]

Issue #15090, opened 2025-10-21
Collections library very slow to load since 10.11 upgrade


kandykarter:
### Description of the bug

After upgrading to Jellyfin 10.11, I've had a fairly smooth transition, but the Collections library takes 10-15 seconds to load, where it used to be instant. The individual collections load smoothly and quickly, but the main page takes ages. I have around 90 collections. 

There is no information generated in the logs during this period. 

I have no plugins currently installed beyond the defaults (aside from Kodi Sync Queue). 

### Reproduction steps

1. Open web client (or android app, I haven't tested other clients)
2. Under the My Media section, select Collections.

### What is the current _bug_ behavior?

Collections page takes 10-15 seconds to load. 

### What is the expected _correct_ behavior?

Collections page should load instantly, like the Movies or Shows page does. 

### Jellyfin Server version

10.10.0+

### Specify the build version

10.11.0

### Environment

```markdown
- OS: Debian Linux
- Virtualization: Docker
- Clients: Web, android, etc
- Browser: Firefox, Chrome
- Plugins: Stock, kodi
- Reverse Proxy: Cosmos cloud
- Jellyfin Data Storage: ssd on n100 pc
- Media Storage: nas/cifs mounts
```

### Jellyfin logs

```shell
Nothing happens in the logs when Collections library is accessed
```

### FFmpeg logs

```shell

```

theguymadmax:
Did you perform a full library scan after upgrading?

kandykarter:
Sure did! Assuming that pressing the "Scan all libraries" button in the dashboard is the correct way of doing it.

kandykarter:
I've now tried a full "replace all metadata" scan on my libraries and this collection issue has not improved. I've also tried it with non-web clients (roku, android tv) and it happens there as well. Occasionally it seems to get to a point where collections will not load at all, until I restart the server.

Sc00tz:
I am seeing the same issue, and when I go to "movies" it's totally screwed up with the wrong count, and hardly anything listed..

I'd be okay with just removing my collections and re-add the files to each collection again, but I don't even know how to do that because it's so slow to load.

s1mpleman:
> I am seeing the same issue, and when I go to "movies" it's totally screwed up with the wrong count, and hardly anything listed..
> 
> I'd be okay with just removing my collections and re-add the files to each collection again, but I don't even know how to do that because it's so slow to load. 

[...]

PR #16472, opened 2026-03-26
Parse provider IDs from season and episode folder/file names

IDisposable:
## Summary

- Parses provider IDs (TVDB, TMDB, TVMaze, IMDB) from season folder names and episode file names at resolve time, e.g. `Season 1 [tvdbid-12345]` or `episode.mkv [imdbid-tt0944947]`
- Adds `TmdbSeasonExternalId` and `TmdbEpisodeExternalId` so TMDB IDs are surfaced in the UI for seasons and episodes
- Tests for `SeasonResolver` and `EpisodeResolver` provider ID parsing, plus new `ExternalId` provider URL tests

## Details

`SeasonResolver` and `EpisodeResolver` each gain a `SetProviderIdFromPath` helper that calls the existing `GetAttributeValue` extension to extract bracket-style attributes from the folder/file name. IDs are written via `TrySetProviderId`, which is a no-op when the value is absent, so paths without annotations are unaffected.

Supported tags per item type:

| Item | Tags |
|---|---|
| Season | `tvdbid`, `tmdbid`, `tvmazeid` * |
| Episode | `tvdbid`, `tmdbid`, `tvmazeid`, `imdbid` |

* imdb doesn't have direct Season Ids

## See also 

Corresponding leveraging of the presence of the IDs

[jellyfin-plugin-tvdb PR #239](https://github.com/jellyfin/jellyfin-plugin-tvdb/pull/239)
[jellyfin-plugin-tvmaze PR #65](https://github.com/jellyfin/jellyfin-plugin-tvmaze/pull/65)

PR #16680, opened 2026-04-22
perf(scan): implement phased library scanning and fast metadata refresh

taos15:
**Changes**

- add phased discovery tasks and fast tree scanning for library refresh

- refresh phase 2 from database-backed work items with provider parallelism and request coalescing

- include supporting resolver, migration, scheduler, and SQLite tuning changes for the new scan pipeline
   
**Issues**
This PR try to fix the following issues: #15352 #15070 #15090


Emby.Naming/Video/VideoListResolver.cs
```diff
@@ -197,29 +200,92 @@ private static bool HaveSameYear(IReadOnlyList<VideoInfo> videos)
...
+            var normalizedTestFilename = NormalizeMultiVersionName(testFilename, namingOptions);
+
+            if (!normalizedTestFilename.StartsWith(normalizedFolderName, StringComparison.OrdinalIgnoreCase))
             {
                 return false;
             }
 
             // Remove the folder name before cleaning as we don't care about cleaning that part
-            if (folderName.Length <= testFilename.Length)
+            if (normalizedFolderName.Length <= normalizedTestFilename.Length)
```
IDisposable, line 229: I'm concerned about how this (existing) comment and code syncs up with the fact that **Show Name** folders _can_ have provider IDs embedded (and will support for Season and Episode if PR #16472 gets merged).

Do we know that the folder name's provider IDs isn't significant?.

Emby.Naming/Video/VideoListResolver.cs
```diff
@@ -197,29 +200,92 @@ private static bool HaveSameYear(IReadOnlyList<VideoInfo> videos)
...
+
+            foreach (var ch in value)
+            {
+                switch (ch)
+                {
+                    case '[':
+                        square++;
+                        break;
+                    case ']':
+                        square = square > 0 ? square - 1 : square;
```
IDisposable, line 271:
```csharp
if (square > 0) square--;
```

Seems clearer... there's going to be a if/jump anyway in the emitted code... let's be clear about it.

Emby.Naming/Video/VideoListResolver.cs
```diff
@@ -197,29 +200,92 @@ private static bool HaveSameYear(IReadOnlyList<VideoInfo> videos)
...
+                        break;
+                    case '(':
+                        round++;
+                        break;
+                    case ')':
+                        round = round > 0 ? round - 1 : round;
```
IDisposable, line 277:
```csharp
if (round > 0) round--;
```

Emby.Naming/Video/VideoListResolver.cs
```diff
@@ -197,29 +200,92 @@ private static bool HaveSameYear(IReadOnlyList<VideoInfo> videos)
...
+                        break;
+                    case '{':
+                        curly++;
+                        break;
+                    case '}':
+                        curly = curly > 0 ? curly - 1 : curly;
```
IDisposable, line 283:
```csharp
if (curly > 0) curly--;
```

Emby.Server.Implementations/Dto/DtoService.cs
```diff
@@ -517,7 +517,7 @@ private static int GetChildCount(Folder folder, User user)
         {
             // Right now this is too slow to calculate for top level folders on a per-user basis
             // Just return something so that apps that are expecting a value won't think the folders are empty
-            if (folder is ICollectionFolder || folder is UserView)
+            if (folder is UserView)
```
IDisposable, line 520: This means that we'll now actually count the children of `ICollectionFolder`. Is that a performance degradation we can/should accept?

Emby.Server.Implementations/Library/IgnorePatterns.cs
```diff
@@ -135,16 +138,62 @@ public static class IgnorePatterns
         /// <returns>Whether to ignore the path.</returns>
         public static bool ShouldIgnore(ReadOnlySpan<char> path)
         {
+            var pathString = path.ToString();
+            if (_cache.TryGetValue(pathString, out var cached))
+            {
+                return cached;
+            }
+
+            if (!MightMatchIgnorePattern(path))
```
IDisposable, line 147: Is this clearer?

```csharp
if (MightMatchIgnorePattern(path))
{
   int len = _globs.Length;
   for (int i = 0; i < len; i++)
   {
      if (_globs[i].IsMatch(path))
      {
         return _cache[pathString] = true;
      }
   }
}

return _cache[pathString] = false;
```

Emby.Server.Implementations/Library/IgnorePatterns.cs
```diff
@@ -135,16 +138,62 @@ public static class IgnorePatterns
...
+        }
+
+        private static bool MightMatchIgnorePattern(ReadOnlySpan<char> path)
+        {
+            if (path.StartsWith(".") || path.Contains("/.", StringComparison.Ordinal) || path.Contains("\\.", StringComparison.Ordinal))
+            {
+                return true;
+            }
+
+            return path.Contains("metadata", StringComparison.OrdinalIgnoreCase)
```
IDisposable, line 171: Seems like this list should be a frozen (or mutable) dictionary somewhere, and also given how it's just a long set of `if` checks, the list should likely be ordered in a most-likely to least-likely order to ensure reasonable performance.

Emby.Server.Implementations/Library/IgnorePatterns.cs
```diff
@@ -135,16 +138,62 @@ public static class IgnorePatterns
...
+                || path.Contains("sample", StringComparison.OrdinalIgnoreCase)
+                || path.Contains("minta", StringComparison.OrdinalIgnoreCase)
+                || path.Contains("trickplay", StringComparison.OrdinalIgnoreCase)
+                || path.Contains("TempRec", StringComparison.OrdinalIgnoreCase)
+                || path.Contains("TempSBE", StringComparison.OrdinalIgnoreCase)
+                || path.Contains("eaDir", StringComparison.OrdinalIgnoreCase)
+                || path.Contains("recycle", StringComparison.OrdinalIgnoreCase)
+                || path.Contains("thumb", StringComparison.OrdinalIgnoreCase)
```
IDisposable, line 179: This will match to the later specified `"thumbs.db"` so that pattern on line 184 can be deleted. Same thing for `"extrathumbs"` on line 196.

Emby.Server.Implementations/Library/IgnorePatterns.cs
```diff
@@ -135,16 +138,62 @@ public static class IgnorePatterns
...
+                || path.Contains("thumbs.db", StringComparison.OrdinalIgnoreCase)
+                || path.Contains(".bts", StringComparison.OrdinalIgnoreCase)
+                || path.Contains(".sync", StringComparison.OrdinalIgnoreCase)
+                || path.Contains(".actors", StringComparison.OrdinalIgnoreCase)
+                || path.Contains(".wd_tv", StringComparison.OrdinalIgnoreCase)
+                || path.Contains("lost+found", StringComparison.OrdinalIgnoreCase)
+                || path.Contains("subs", StringComparison.OrdinalIgnoreCase)
+                || path.Contains("snapshot", StringComparison.OrdinalIgnoreCase)
+                || path.Contains("System Volume Information", StringComparison.OrdinalIgnoreCase)
+                || path.Contains("ps3_update", StringComparison.OrdinalIgnoreCase)
```
IDisposable, line 193: Would testing for `"ps3_"` be too broad?

Emby.Server.Implementations/Library/IgnorePatterns.cs
```diff
@@ -135,16 +138,62 @@ public static class IgnorePatterns
...
+                || path.Contains("ps3_vprm", StringComparison.OrdinalIgnoreCase)
+                || path.Contains("extrafanart", StringComparison.OrdinalIgnoreCase)
```
IDisposable, line 195: But not plain _fanart_?

Emby.Server.Implementations/ScheduledTasks/Tasks/RefreshMediaLibraryTask.cs
```diff
@@ -60,6 +59,62 @@ public async Task ExecuteAsync(IProgress<double> progress, CancellationToken can
...
         progress.Report(0);
 
-        await ((LibraryManager)_libraryManager).ValidateMediaLibraryInternal(progress, cancellationToken).ConfigureAwait(false);
+        // Cancel any running phase tasks before starting full scan
+        await CancelAndWaitForPhaseTaskAsync<RefreshMediaLibraryPhase1Task>(cancellationToken).ConfigureAwait(false);
+        await CancelAndWaitForPhaseTaskAsync<RefreshMediaLibraryPhase2Task>(cancellationToken).ConfigureAwait(false);
+
+        // Phase 1: discovery and local NFO metadata (0% - 40% of full scan)
+        var phase1Worker = GetTaskWorker<RefreshMediaLibraryPhase1Task>();
+        void OnPhase1Progress(object? sender, GenericEventArgs<double> e) => progress.Report(e.Argument * 0.4);
```
IDisposable, line 68: 40% for the new-file and read NFO scan seems too large... I would expect the metadata scans to be a much larger proportion of the overall scan time... consider lowering this?

Emby.Server.Implementations/ScheduledTasks/Tasks/RefreshMediaLibraryTask.cs
```diff
@@ -60,6 +59,62 @@ public async Task ExecuteAsync(IProgress<double> progress, CancellationToken can
...
+        }
+    }
+
+    private async Task CancelAndWaitForPhaseTaskAsync<T>(CancellationToken cancellationToken)
+        where T : IScheduledTask
+    {
+        var worker = _taskManager.ScheduledTasks.FirstOrDefault(t => t.ScheduledTask is T);
+        if (worker is ScheduledTaskWorker scheduledTaskWorker)
+        {
+            scheduledTaskWorker.CancelIfRunning();
```
IDisposable, line 107: The way this is constructed, we'll cancel the phase one worker, wait for it to completely die, then cancel the phase two worker... seems like we could/should do the cancels simultaneously and then wait for both to die cumulatively.

Does phase two actually start before phase one ends? 
- If not then we should really not try to cancel phase two if phase one hasn't completed. 
- If they do run in parallel, then would we be wiser to cancel them in opposite order (phase 2, then phase 1), since the phase 2 stuff is the more costly and cancelling more of it is of more value?

Jellyfin.Api/Controllers/LibraryController.cs
```diff
@@ -425,18 +436,149 @@ public ActionResult DeleteItems([FromQuery, ModelBinder(typeof(CommaDelimitedCol
...
+        if (parentId.HasValue && recursive && (includeItemTypes is null || includeItemTypes.Length == 0))
+        {
+            idsToDelete.Add(parentId.Value);
+        }
+        else if (parentId.HasValue)
+        {
+            var query = new InternalItemsQuery(user)
+            {
+                ParentId = parentId.Value,
+                Recursive = recursive,
```
IDisposable, line 510: Why would we want to include the recursive items instead of just _actually_ recursing so that all the intervening parents would be rights-checked?

Jellyfin.Api/Controllers/LibraryController.cs
```diff
@@ -425,18 +436,149 @@ public ActionResult DeleteItems([FromQuery, ModelBinder(typeof(CommaDelimitedCol
...
+
+                idsToDelete.Add(item.Id);
+                if (recursive && item.IsFolder)
+                {
+                    var childQuery = new InternalItemsQuery(user)
+                    {
+                        AncestorIds = [item.Id],
+                        DtoOptions = new DtoOptions(false)
+                    };
+                    idsToDelete.AddRange(_libraryManager.GetItemList(childQuery).Select(i => i.Id));
```
IDisposable, line 550: Design questions

1. If I have rights to delete a parent, and the parent has subfolders, I have right to delete those subfolders... but what if those subfolders have children (items or subsubfolders) that I _don't_ have rights to delete... should those be subfolders actually NOT be deleted?
2. When we do a `_libraryManager.DeleteItem(`, (like we do in the delete-by-ids case above), does _it_ delete the subfolders and items underneath automatically? If so, why are we mining down here only?

Jellyfin.Api/Controllers/LibraryController.cs
```diff
@@ -425,18 +436,149 @@ public ActionResult DeleteItems([FromQuery, ModelBinder(typeof(CommaDelimitedCol
...
+                    _libraryManager.DeleteItem(
+                        item,
+                        new DeleteOptions { DeleteFileLocation = true },
+                        false);
+                }
+            }
+        }
+        else
+        {
+            _itemRepository.DeleteItem(idsToDelete);
```
IDisposable, line 576: Does this delete the child items internally?

Jellyfin.Api/Controllers/LibraryController.cs
```diff
@@ -425,18 +436,149 @@ public ActionResult DeleteItems([FromQuery, ModelBinder(typeof(CommaDelimitedCol
...
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    public ActionResult<int> DeleteItemsByQuery(
+        [FromQuery] Guid? parentId,
+        [FromQuery, ModelBinder(typeof(CommaDelimitedCollectionModelBinder))] BaseItemKind[]? includeItemTypes,
+        [FromQuery] bool recursive = true,
+        [FromQuery] bool deleteFileLocation = false)
+    {
+        var userId = User.GetUserId();
```
IDisposable, line 473: The whole point of this seems to be to grab the items to be deleted, and then gather up the ids of the children (grandchildren, etc) so that we can delete the physical files if requested. 

Design question: What if we made `_itemRepository.DeleteItem(` (or a new function) return a list of Guids it deleted... that way we could just iterate the list and delete the files at the file-system level after the items without all this kerfuffle.

Jellyfin.Server.Implementations/FullSystemBackup/BackupService.cs
```diff
@@ -156,6 +157,16 @@ void CopyDirectory(string source, string target, string[]? exclude = null)
             if (manifest.Options.Database)
             {
                 _logger.LogInformation("Begin restoring Database");
+
```
IDisposable, line 160: The changes in this file seem completely unrelated. Should these be in a distinct PR?

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -640,17 +640,19 @@ public void UpdateOrInsertItems(IReadOnlyList<BaseItemDto> items, CancellationTo
...
+        var existingItems = context.BaseItems.Where(e => ids.Contains(e.Id)).Select(f => f.Id).ToHashSet();
 
         foreach (var item in tuples)
         {
             var entity = Map(item.Item);
             // TODO: refactor this "inconsistency"
             entity.TopParentId = item.TopParent?.Id;
 
-            if (!existingItems.Any(e => e == entity.Id))
+            if (!existingItems.Contains(entity.Id))
```
IDisposable, line 655: Is this change related to the `// TODO: refactor this "inconsistency"` comment above? Does this address it (and thus the comment should be deleted)?

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
```diff
@@ -1060,13 +1142,25 @@ public BaseItemEntity Map(BaseItemDto dto)
...
         entity.Height = dto.Height;
-        entity.Provider = dto.ProviderIds.Select(e => new BaseItemProvider()
+        var providers = dto.ProviderIds.Select(e => new BaseItemProvider()
         {
             Item = entity,
             ProviderId = e.Key,
             ProviderValue = e.Value
         }).ToList();
 
+        var duplicateProvider = providers.GroupBy(p => p.ProviderId).FirstOrDefault(g => g.Count() > 1);
```
IDisposable, line 1152: Seems like we're going to complain about one duplicate provider at a time, and if they somehow clean that one up, complain about the next one, etc...

Would it be better to make this `.Where(g => g.Count() > 1)` and walk the entire set of duplicates in the error message below and build one message to throw with all of the dupes enumerated? In fact, that could easily be done with a `.Select()`

I'm also curious why this is being done at this layer at all... why would we want to throw an `Exception` when simply trying to `Map` to an entity from a DTO?

Jellyfin.Server/Program.cs
```diff
@@ -88,6 +88,8 @@ private static async Task StartApp(StartupOptions options)
             ServerApplicationPaths appPaths = StartupHelpers.CreateApplicationPaths(options);
             appPaths.MakeSanityCheckOrThrow();
 
+            ThreadPool.SetMinThreads(200, 200);
```
IDisposable, line 91: Seems like as config knob would be warranted here.

Jellyfin.Server/Startup.cs
```diff
@@ -85,13 +85,15 @@ public void ConfigureServices(IServiceCollection services)
...
+                ConnectCallback = HttpClientExtension.OnConnect,
+                PooledConnectionLifetime = TimeSpan.FromMinutes(5)
             };
 
             Func<IServiceProvider, HttpMessageHandler> defaultHttpClientHandlerDelegate = (_) => new SocketsHttpHandler()
             {
                 AutomaticDecompression = DecompressionMethods.All,
-                RequestHeaderEncodingSelector = (_, _) => Encoding.UTF8
+                RequestHeaderEncodingSelector = (_, _) => Encoding.UTF8,
+                PooledConnectionLifetime = TimeSpan.FromMinutes(5)
```
IDisposable, line 96: Seems like a configuration knob would be valuable here.

MediaBrowser.Controller/Entities/Audio/MusicAlbum.cs
```diff
@@ -169,21 +169,24 @@ public async Task RefreshAllMetadata(MetadataRefreshOptions refreshOptions, IPro
...
-                percent /= totalItems;
-                progress.Report(percent * 95);
-            }
+            await Parallel.ForEachAsync(
+                items,
+                new ParallelOptions { MaxDegreeOfParallelism = parallelism, CancellationToken = cancellationToken },
+                async (item, ct) =>
+                {
+                    var updateType = await item.RefreshMetadata(refreshOptions, ct).ConfigureAwait(false);
+                    lock (updateLock)
```
IDisposable, line 182: We don't need a lock if we use `Interlocked.Or(ref childUpdateType, updateType)`

MediaBrowser.Controller/Entities/BaseItem.cs
```diff
@@ -2524,6 +2524,11 @@ protected async Task RefreshMetadataForOwnedVideo(MetadataRefreshOptions options
                 video.OwnerId = Id;
             }
 
+            if (!copyTitleMetadata && video is Video ownedVideo)
```
IDisposable, line 2527: CoPilot is complaining that `video is Video` is always true by the time we get here.

MediaBrowser.Controller/Providers/DirectoryService.cs
```diff
@@ -26,7 +26,19 @@ public DirectoryService(IFileSystem fileSystem)
 
         public FileSystemMetadata[] GetFileSystemEntries(string path)
         {
-            return _cache.GetOrAdd(path, static (p, fileSystem) => fileSystem.GetFileSystemEntries(p).ToArray(), _fileSystem);
+            return _cache.GetOrAdd(
+                path,
+                static (p, state) =>
+                {
+                    var entries = state.FileSystem.GetFileSystemEntries(p).ToArray();
```
IDisposable, line 33: Not sure the `.ToArray()` adds any value here, since we're just immediately iterating the entries and shoving them into the FileCache.

MediaBrowser.Model/Configuration/ServerConfiguration.cs
```diff
@@ -245,6 +245,11 @@ public ServerConfiguration()
     /// </summary>
     public int LibraryScanFanoutConcurrency { get; set; }
 
+    /// <summary>
+    /// Gets or sets the memory limit in MB for buffering library scan items before writing to the database.
+    /// </summary>
+    public int LibraryScanMemoryLimitMB { get; set; } = 100;
```
IDisposable, line 251: Are we actually measuring the memory creep? Seems like this is really number of item * 100 give the code above.  Maybe change to be `LibraryScanItemsLimitHundreds` or reconsider?

MediaBrowser.Providers/Manager/MetadataService.cs
```diff
@@ -904,44 +905,106 @@ private async Task<RefreshResult> ExecuteRemoteProviders(MetadataResult<TItemTyp
...
+                        MaxDegreeOfParallelism = Math.Min(3, groupProviders.Count),
+                        CancellationToken = cancellationToken
+                    };
+
+                    var groupResults = new ConcurrentBag<(bool HasMetadata, ItemUpdateType UpdateType, string ErrorMessage, int Failures)>();
+
+                    await Parallel.ForEachAsync(groupProviders, options, async (provider, ct) =>
+                    {
+                        var providerResult = await RunRemoteProviderAsync(provider, temp, logName, replaceData, id, cancellationToken).ConfigureAwait(false);
+                        groupResults.Add(providerResult);
```
IDisposable, line 946: What's the value of caching the groups and then doing the process below instead of inlining that code here? 

We would have to switch to `Interlocked.Or(ref result.UpdateType, ItemUpdateType.MetadataDownload)`, `InterlockedIncrement(ref refreshResult.Failures)` and `Interlocked.CompareExchange(` on the `ErrorMessage` but that's got to be cheaper than the memory pressure of keeping processed groups around.

MediaBrowser.Providers/Manager/MetadataService.cs
```diff
@@ -904,44 +905,106 @@ private async Task<RefreshResult> ExecuteRemoteProviders(MetadataResult<TItemTyp
...
+                    }).ConfigureAwait(false);
+
+                    foreach (var result in groupResults)
+                    {
+                        if (result.HasMetadata)
+                        {
+                            refreshResult.UpdateType |= ItemUpdateType.MetadataDownload;
+                        }
+
+                        if (!string.IsNullOrEmpty(result.ErrorMessage))
```
IDisposable, line 956: Seems like we should track/concatenate all the error messages, not just keep the first one encountered.

MediaBrowser.Providers/Manager/MetadataService.cs
```diff
@@ -904,44 +905,106 @@ private async Task<RefreshResult> ExecuteRemoteProviders(MetadataResult<TItemTyp
...
+                        {
+                            refreshResult.ErrorMessage = result.ErrorMessage;
+                        }
+
+                        refreshResult.Failures += result.Failures;
```
IDisposable, line 961: Looks like `result.Failures` is either zero if there was no `result.ErrorMessage` or `1` if there was... probably could just move `refreshResult.Failures++` up into that loop and not return the extra `int` from the `RunRemoteProviderAsync(` method

MediaBrowser.Providers/Manager/ProviderManager.cs
```diff
@@ -333,8 +327,6 @@ private async Task<IEnumerable<RemoteImageInfo>> GetImages(
 
                 if (!includeAllLanguages && hasPreferredLanguage)
                 {
-                    // Filter out languages that do not match the preferred languages.
-                    //
                     // TODO: should exception case of "en" (English) eventually be removed?
```
IDisposable, line 330: This `// TODO` is stale, we don't use `"en"`, we use whatever's set in `preferredLanguage`.

MediaBrowser.Providers/Manager/ProviderManager.cs
```diff
@@ -1042,26 +1023,35 @@ private async Task StartProcessingRefreshQueue()
 
             var cancellationToken = _disposeCancellationTokenSource.Token;
 
-            while (_refreshQueue.TryDequeue(out var refreshItem, out _))
+            while (!_disposed)
```
IDisposable, line 1026: Is this why we switch from `using` to `var` way up? Seems like the wrong way to handle cancellation.

MediaBrowser.Providers/Manager/ProviderManager.cs
```diff
@@ -1042,26 +1023,35 @@ private async Task StartProcessingRefreshQueue()
...
+                    refreshBatch.Add(refreshItem);
+                }
+
+                if (refreshBatch.Count == 0)
                 {
-                    return;
+                    break;
                 }
 
+                var refreshTasks = refreshBatch.Select(refreshItem => ProcessQueuedRefreshItem(libraryManager, refreshItem, cancellationToken));
```
IDisposable, line 1039: Seems like we could be just `.Add(`ing the `refreshTask` collection up at line 1031 instead of building a `refreshBatch` and then running a `.Select(` over it.

MediaBrowser.Providers/Manager/ProviderManager.cs
```diff
@@ -1079,10 +1069,73 @@ private async Task StartProcessingRefreshQueue()
...
+                    ? RefreshArtist(artist, refreshItem.RefreshOptions, cancellationToken)
+                    : RefreshItem(item, refreshItem.RefreshOptions, cancellationToken);
+
+                await task.ConfigureAwait(false);
+                itemStopwatch.Stop();
+
+                var itemProcessingMs = itemStopwatch.Elapsed.TotalMilliseconds;
+                return (true, queueWaitMs, itemProcessingMs);
+            }
+            catch (OperationCanceledException)
```
IDisposable, line 1102: Better expressed with a filtered catch block as it preserves the call stack (and thus doesn't cause more memory churn)

```csharp
catch (Exception ex) when (ex is not OperationCanceledException)
```

MediaBrowser.Providers/Manager/MetadataService.cs
```diff
@@ -904,44 +905,106 @@ private async Task<RefreshResult> ExecuteRemoteProviders(MetadataResult<TItemTyp
...
+                    MergeData(result, temp, [], replaceData, false);
+                    MergeNewData(temp.Item, id);
+
+                    return (true, ItemUpdateType.MetadataDownload, null, 0);
+                }
+
+                Logger.LogDebug("{Provider} returned no metadata for {Item}", providerName, logName);
+                return (false, ItemUpdateType.None, null, 0);
+            }
+            catch (OperationCanceledException)
```
IDisposable, line 997: Better expressed with a filtered catch block as it preserves the call stack (and thus doesn't cause more memory churn)

```csharp
catch (Exception ex) when (ex is not OperationCanceledException)
```

MediaBrowser.Providers/Manager/ProviderManager.cs
```diff
@@ -1042,26 +1023,35 @@ private async Task StartProcessingRefreshQueue()
...
+                            continue;
+                        }
 
-                    await task.ConfigureAwait(false);
+                        totalQueueWaitMs += result.QueueWaitMs;
+                        totalProcessingMs += result.ProcessingMs;
+                        processedItems++;
+                    }
                 }
                 catch (OperationCanceledException)
```
IDisposable, line 1056: Better expressed with a filtered catch block as it preserves the call stack (and thus doesn't cause more memory churn)

```csharp
catch (Exception ex) when (ex is not OperationCanceledException)
```

MediaBrowser.Providers/Manager/ProviderManager.cs
```diff
@@ -1079,10 +1069,73 @@ private async Task StartProcessingRefreshQueue()
...
+
+        private int GetMetadataRefreshConcurrencyLimit()
+        {
+            var configuredConcurrency = _configurationManager.Configuration.LibraryMetadataRefreshConcurrency;
+            if (configuredConcurrency > 0)
+            {
+                return configuredConcurrency;
+            }
+
+            return Environment.ProcessorCount <= 3
```
IDisposable, line 1122: This would be nicer to ensure we don't saturate all the CPUs on a huge machine. There's very little chance we would want to spawn 12 tasks :)

The `4` could also be a config knob

```csharp
return Math.Clamp(Environment.ProcessorCount - 3, 1, 4)
```

MediaBrowser.Providers/MediaInfo/FFProbeVideoInfo.cs
```diff
@@ -84,67 +84,67 @@ public async Task<ItemUpdateType> ProbeVideo<T>(
...
+                    cancellationToken.ThrowIfCancellationRequested();
                 }
-                else
-                {
-                    mediaInfoResult = await GetMediaInfo(item, cancellationToken).ConfigureAwait(false);
-                }
-
-                cancellationToken.ThrowIfCancellationRequested();
+            }
+            catch (FfmpegException ex) when (ex.Message.Contains("Permission denied", StringComparison.OrdinalIgnoreCase))
```
IDisposable, line 145: Is this multilingual-safe on an OS/.Net install in a different language?

MediaBrowser.Providers/Plugins/Tmdb/TmdbClientManager.cs
```diff
@@ -93,29 +96,26 @@ public TmdbClientManager(IMemoryCache memoryCache)
...
+                key,
+                async ct =>
+                {
+                    await EnsureClientConfigAsync().ConfigureAwait(false);
+
+                    var collection = await _tmDbClient.GetCollectionAsync(
+                        tmdbId,
+                        TmdbUtils.NormalizeLanguage(language, countryCode),
+                        imageLanguages,
+                        CollectionMethods.Images,
```
IDisposable, line 113: Not your change, but find it odd the we don't have/want `CollectionMethods.Videos` and `CollectionMethods.Keywords` and `CollectionMethods.ExternalIds` for collections.

MediaBrowser.Providers/Plugins/Tmdb/TmdbClientManager.cs
```diff
@@ -53,35 +59,32 @@ public TmdbClientManager(IMemoryCache memoryCache)
...
-            }
 
-            return movie;
+            return ExecuteWithCoalescingAsync(
+                key,
+                async ct =>
+                {
+                    await EnsureClientConfigAsync().ConfigureAwait(false);
+
+                    var extraMethods = MovieMethods.Credits | MovieMethods.Releases | MovieMethods.Images | MovieMethods.Videos;
```
IDisposable, line 72: Not your change, but find it odd the we don't have/want `MovieMethods.ExternalIds` for movies.

MediaBrowser.Providers/Plugins/Tmdb/TmdbClientManager.cs
```diff
@@ -695,6 +685,59 @@ public async Task<TMDbConfig> GetClientConfiguration()
...
+            where TResult : class
+        {
+            if (_memoryCache.TryGetValue(key, out TResult? cachedValue))
+            {
+                return cachedValue;
+            }
+
+            if (_completedRequestTimes.TryGetValue(key, out var completedTime)
+                && DateTime.UtcNow - completedTime < TimeSpan.FromSeconds(InflightCoalescingWindowSeconds)
+                && _memoryCache.TryGetValue(key, out cachedValue))
```
IDisposable, line 698: This block seems wrong, specifically that the previous `TryGetValue(` failed, but we have a record of a completion for this query, it was recent enough, and we now get a value from `TryGetValue(`... realistically that's almost never going to happen unless we `Thread.Sleep(` because there's very little chance we're in between some other thread's line 711-722... that's no the slow part... it's the `fetchFunc(` that is slow but we aren't giving it any time to finish up.

If I understand the point of this, if we get a cache miss, I think what we want is a time-bound collection of inflight/outstanding requests... if there's one that for this key, and it hasn't been too long, we should _wait with timeout_ to see if it arrives. If it doesn't then we should fire off our own request. If it does return in that time, the entry should be in the cache so we'll be able to return it... Is that the desired behavior?

If so, then what we want is a coaleascing-time-bound map in `inflightRequests` of in-flight requests, each carrying the request identifier (just any nonce like a static incrementing integer) and a expiry time.

- When we start processing a request, we check the `_memoryCache.TryGetValue(key, ` and if returns `true`, then `return cachedValue`.
- If `false`/not found, we **enter a loop** fetching the entry from the `inflightRequests.TryGetValue(key, out inflight)`.
- If it returns `false`, there's no longer an in-flight request for our `key`, so we `break`.
- If the fetched `inflight` has expired (`inflight.Expires < DateTime.UtcNow`) then we `break` because waiting has exceeded the coalesce window.
- We retry the `_memoryCache.TryGetValue(key, ` and return the `cachedValue` entry if it returns `true`. 
- If `false`/_doesn't succeed_ (or the `cachedValue` is `null` ?), we compute a `nonce` value then we add our **own** in-flight notation for the `key` with a expiration time set to `DateTime.UtcNow + TimeSpan.FromSeconds(InflightCoalescingWindowSeconds)` with our `nonce`
- We execute the `fetchedValue = fetchFunc` 
- When it returns, we add the result `fetchedValue` to the `_memoryCache` 
- **Then** we remove OUR in-flight entry from `inflightRequests` if it exists. If the entry in the `inflightRequest` does NOT match our `nonce`, then we just leave it alone, it's someone else's in-flight that replaced ours because we took too long.
- Then we `return fetchedValue`

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteDatabaseProvider.cs
```diff
@@ -190,10 +190,30 @@ public async Task PurgeDatabase(JellyfinDbContext dbContext, IEnumerable<string>
...
+                {
+                    existingTables.Add(reader.GetString(0));
+                }
+            }
+        }
+
+        var deleteQueries = tableNames
+            .Where(existingTables.Contains)
+            .Select(tableName => $"DELETE FROM \"{tableName}\";")
+            .ToList();
```
IDisposable, line 212: Do we want to exclude the migration history table here?

Emby.Naming/Video/VideoListResolver.cs
taos15: The provider ID tokens are intentionally stripped only for multi-version file matching. 

The problem being solved is: files like `Meg 2 {imdb-tt123} [AMZN]...mkv` and `Meg 2 [tmdbid-456] [Remux]...mkv` inside the same folder should still be recognized as versions of the same movie, even though they embed different provider IDs. 

The test `TestMultiVersionWithDifferentProviderTokens` validates this exact scenario. Provider IDs for series/season/episode resolution are handled by the `resolvers` separately and are unaffected.

Emby.Naming/Video/VideoListResolver.cs
taos15: Both implementation do the same. I'll switch to the explicit conditional form.

Emby.Naming/Video/VideoListResolver.cs
taos15: Both implementation do the same. I'll switch to the explicit conditional form.

Emby.Naming/Video/VideoListResolver.cs
taos15: Both implementation do the same. I'll switch to the explicit conditional form.

Emby.Server.Implementations/Dto/DtoService.cs
taos15: This is intentional. With the phased scan, phase 1 populates items in the database before the UI queries them, so `ICollectionFolder` children are now reliably available for counting. 

The previous skip was because the old scan populated items lazily, making counts unreliable. Since phase 1 discovery runs first and fills the DB, this is no longer a concern.

Emby.Server.Implementations/Library/IgnorePatterns.cs
taos15: Agreed, I'll restructure it.

Emby.Server.Implementations/Library/IgnorePatterns.cs
taos15: I dont agree, these are substring `Contains` checks, not exact-key lookups, so a dictionary wouldn't help. 

The `MightMatchIgnorePattern prefilter` already short-circuits the majority of paths before reaching these checks, which was the main performance goal here.

Emby.Server.Implementations/Library/IgnorePatterns.cs
taos15: Good catch. I'll remove the redundant entries.

Emby.Server.Implementations/Library/IgnorePatterns.cs
taos15: Agreed, I will keep the specific `ps3_update`/`ps3_vprm` matches.

Emby.Server.Implementations/Library/IgnorePatterns.cs
taos15: Plain `fanart` is too broad. It could match legitimate content names. `extrafanart` is a Kodi-specific convention that's safe to ignore.

Emby.Server.Implementations/ScheduledTasks/Tasks/RefreshMediaLibraryTask.cs
taos15: Fair point. Phase 1 is designed to be the fast filesystem-only discovery pass, 40% using the total time early iteration were taking in my test. I want more test to get a better benchmark of phase 1 and phase 2 duration; phase 2 (metadata providers) is typically the bottleneck.

Keeping it as 40% unless I must change it, after some test and feedback we can adjust it. I think this the split of progress is trivial, in the current implementation you can be stuck at certain percentages for hours, and then jump twice the progress within minutes. I plan to revise this later, but I think the PR got complicated enough to bother with this at this stage.

Emby.Server.Implementations/ScheduledTasks/Tasks/RefreshMediaLibraryTask.cs
taos15: Phase 1 and phase 2 are strictly sequential when invoked by the orchestrator;  phase 2 never starts before phase 1 completes. The `CancelAndWait` at the top handles the case where a standalone phase task was manually triggered before the full scan started. Since they're separate `IScheduledTask` registrations with no default triggers (yield break), at most one phase would be running independently at any given time, so sequential cancellation is sufficient.

 That said, you raise a fair point;  if someone somehow manually triggered both phase tasks independently (from the dashboard), they could theoretically overlap. The `_libraryScanCts` sharing in `GetLibraryScanCancellationToken` provides a partial safety net (cancelling one cancels both), but the `IsScanRunning` flag and `LibraryMonitor.Stop/Start` lifecycle aren't guarded against concurrent phases. I could add a `SemaphoreSlim(1,1)` in `LibraryManager` to ensure only one phase runs at a time. Would that address your concern?

Jellyfin.Api/Controllers/LibraryController.cs
taos15: Rights are checked; the parent is checked at `line 492` (`parent.CanDelete(user)`), and each individual descendant item is checked at `lines 517-519` and `534-536` before being added to the delete set. The query approach avoids expensive tree traversal while preserving per-item authorization. This is important for the phased scan where we need efficient bulk operations.

Jellyfin.Api/Controllers/LibraryController.cs
taos15: `_libraryManager.DeleteItem` does handle child deletion from the database. The explicit descendant gathering here is specifically needed for the `deleteFileLocation` feature; we need to collect all physical file paths `BEFORE` deleting from the DB so we can clean up the filesystem afterward. Without this pre-collection, we'd lose the path information after the DB delete. Each item is individually rights-checked before being added to the delete set.

Jellyfin.Api/Controllers/LibraryController.cs
taos15: Yes, the database delete cascades to children. The batched ID collection above is for gathering filesystem paths before the delete, not for implementing the cascade.

Jellyfin.Api/Controllers/LibraryController.cs
taos15: That's a good longer-term design improvement. For this PR, I'm keeping the current approach to minimize the surface area of changes to the repository interface since I have made enough changes to make reviewing this PR a nightmare.

Jellyfin.Server.Implementations/FullSystemBackup/BackupService.cs
taos15: These are SonarQ-driven log level changes (Information -> Debug) to reduce log noise during backup/restore operations. They're in this PR because the `PurgeDatabase` method they relate to was also modified (to validate table existence before purging). I could split them out if preferred, but they're small and directly related to the database changes.

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
taos15: No, this change is unrelated to that TODO. This is a performance optimization; switching from `Array.Any(e => e == id)` `(O(n))` to `HashSet.Contains(id)` `(O(1))` for the existence check during batch upserts. The TODO refers to the `TopParentId` mapping inconsistency on the line below and remains valid.

Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
taos15: Good point on reporting all duplicates at once; I'll change this to aggregate all duplicate provider IDs into a single error message. As for throwing in `Map()`: this is a defensive guard because duplicate provider IDs cause DB unique constraint violations during save. Catching it here produces a clear, actionable error message instead of a cryptic EF Core exception.

Jellyfin.Server/Program.cs
taos15: This addresses thread pool starvation during phased library scans. 

Phase 1 parallelizes folder traversal and phase 2 runs concurrent metadata refreshes; both can exhaust the default thread pool minimum on large libraries, causing slow ramp-up while the pool grows one thread every 500ms. 

The value 200 is a reasonable floor that prevents starvation without being wasteful. I'd prefer to avoid exposing this as user-facing config since incorrect values could cause worse problems; this should "just work."

Jellyfin.Server/Startup.cs
taos15: This is a well-documented best practice [ dotnetfundamentals](https://learn.microsoft.com/en-us/dotnet/fundamentals/networking/http/httpclient-guidelines#pooled-connections) for `SocketsHttpHandler`; it prevents DNS caching issues and stale connections by recycling pooled connections periodically. 5 minutes is the standard recommended value. This is infrastructure plumbing, not user-facing behavior.

Like the previous comment I don't think a config knob is appropriate here.

MediaBrowser.Controller/Entities/Audio/MusicAlbum.cs
taos15: I disagree, the lock isn't just protecting the bitwise OR;  it's also synchronizing the `numComplete` increment and `progress.Report()` call together as an atomic unit. Splitting to `Interlocked.Or` would still require synchronization for the progress counter/reporting, so the lock is simpler and clearer here.

MediaBrowser.Controller/Entities/BaseItem.cs
taos15: Good catch. I will verify and if copilot is right, and the parameter is already typed as Video this indeed is redundant. 

I will verify and simplify if needed.

MediaBrowser.Controller/Providers/DirectoryService.cs
taos15: I disagree,  `.ToArray()` is intentional; this result is cached in a `ConcurrentDictionary` and returned as `FileSystemMetadata[]`. Caching a deferred enumerable would be unsafe (multiple enumeration, potential filesystem changes between reads). The materialization ensures a stable, cached snapshot.

MediaBrowser.Model/Configuration/ServerConfiguration.cs
taos15: That is a fair point. The value is used to estimate when to flush batched items to the database during phase 1 scanning, based on an approximate per-item memory cost.

I can rename this to something clearer like `LibraryScanBatchFlushThresholdMB` or add a comment explaining the heuristic.

MediaBrowser.Providers/Manager/MetadataService.cs
taos15: I disagree, the memory overhead of the grouped results is negligible compared to the actual provider HTTP work being parallelized. The current structure keeps the ordering/parallelism logic easy to follow; pushing interlocked mutations into the parallel loop would make the code harder to reason about for minimal performance gain.

MediaBrowser.Providers/Manager/MetadataService.cs
taos15: We could, the current "last error wins" behavior matches the pre-existing pattern, but I agree it could lose information.

I can consider aggregating the error messages so all providers failures are visible, will that address your concern?

MediaBrowser.Providers/Manager/MetadataService.cs
taos15: I agree, I will simplify this.

MediaBrowser.Providers/Manager/ProviderManager.cs
taos15: Ok, I'll remove this stale TODO.

MediaBrowser.Providers/Manager/ProviderManager.cs
taos15: The change to `while (!_disposed)` is intentional for batch processing. The old pattern dequeued and processed one item at a time. The new pattern dequeues a batch (up to `maxParallelRefreshes` items) and processes them concurrently with `Task.WhenAll`. The loop breaks when the queue is empty (`refreshBatch.Count == 0`), and the `!_disposed` guard prevents re-entering after disposal. 

The cancellation token is still checked inside `ProcessQueuedRefreshItem`. This batching is important for phase 2 throughput; processing metadata refreshes one-by-one was a major bottleneck.

MediaBrowser.Providers/Manager/ProviderManager.cs
taos15: The separation is intentional. First we collect the batch (dequeue phase), then we start all tasks together (execution phase). If we started tasks inside the dequeue loop, early tasks would already be running and potentially completing while we're still dequeuing, making the batch behavior non-deterministic. 

The current approach also keeps the error handling clean; one `try/catch` around `Task.WhenAll` for the entire batch.

Emby.Server.Implementations/Library/IgnorePatterns.cs
IDisposable: I wasn't suggesting looking up the dictionary, I was saying  (poorly, sorry)we could have an array of patterns to loop over for `.Contains(`

Emby.Server.Implementations/ScheduledTasks/Tasks/RefreshMediaLibraryTask.cs
IDisposable: That would be better, I think

MediaBrowser.Controller/Providers/DirectoryService.cs
IDisposable: My comment was that we call .ToList(), and the enumerate the list, which is equivalent to having just done the enumeration on the IEnumerable (because that's what the ToList is doing, plus the overhead of creating a list...

Your comment would apply to lazy enumeration later not the immediate one we are doing here.

MediaBrowser.Providers/Manager/ProviderManager.cs
taos15: I'll switch to the filtered catch pattern.

MediaBrowser.Providers/Manager/MetadataService.cs
taos15: I'll switch to the filtered catch pattern.

MediaBrowser.Providers/Manager/ProviderManager.cs
taos15: This one is different from the other two; the `OperationCanceledException` catch here does `break`; (not `throw`;) to cleanly exit the batch processing loop. Using `catch (Exception ex) when (ex is not OperationCanceledException`) would skip the break and let the cancellation propagate up, which changes the shutdown behavior. The separate catch blocks are intentional here.

MediaBrowser.Providers/Manager/ProviderManager.cs
taos15: There's already a config knob; `LibraryMetadataRefreshConcurrency` is checked first (line 1116). This CPU-based calculation is only the fallback when it's unset (0). The `Math.Clamp` suggestion to cap at 4 makes sense though; I'll apply that to the fallback to avoid saturating high-core machines.

MediaBrowser.Providers/MediaInfo/FFProbeVideoInfo.cs
taos15: Good point. The `"Permission denied"` string comes from ffmpeg's stderr output, which is typically in English regardless of system locale; ffmpeg doesn't localize its error messages ( I least I think it doesnt). However, I should add a fallback check for the POSIX error code (EACCES) or consider catching a more specific exception type if one becomes available. For now, this handles the most common case where ffmpeg can't read a file due to permissions (e.g., network-mounted media with restricted access during scan).

MediaBrowser.Providers/Plugins/Tmdb/TmdbClientManager.cs
taos15: I agreed, not my change. This could be a good follow-up improvement in a separate PR; adding `Videos`, `Keywords`, and `ExternalIds` to collection fetches.

MediaBrowser.Providers/Plugins/Tmdb/TmdbClientManager.cs
taos15: Like above, it is a good follow-up PR.

MediaBrowser.Providers/Plugins/Tmdb/TmdbClientManager.cs
taos15: You're right that the `_completedRequestTimes` check (`lines 696-701`) is nearly useless; if the `_memoryCache` missed at `line 691`, it's extremely unlikely to hit at `line 698` in the microseconds between the two calls. That check was meant as a "just barely expired from cache" safety net but in practice it's dead code.

The actual request coalescing happens via `_inflightRequests.GetOrAdd()` at `line 704`; when multiple callers request the same key concurrently, only the first one executes `fetchFunc`, and all others await the same `Task`. This IS the correct deduplication pattern.

I'll remove the `_completedRequestTimes` check since it adds complexity without real value. The `GetOrAdd` coalescing already handles the concurrent case correctly. Your proposed loop/nonce/expiry design would add robustness for edge cases (e.g., a fetch taking longer than the coalescing window) but I think that's premature complexity — if a fetch takes too long, the right answer is probably to let additional requests fire rather than having callers wait indefinitely.

Will Removing the `_completedRequestTimes` check address your concerns?

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteDatabaseProvider.cs
taos15: `__EFMigrationsHistory` isn't included here; `tableNames` is derived from the EF entity DbSets in the caller (`BackupService`), and the migrations history table isn't a mapped entity. So it's already excluded by the input, not by this code. But I could add an explicit exclusion as a safety guard if you'd prefer.

MediaBrowser.Controller/Providers/DirectoryService.cs
taos15: Looking at line 33 again: `state.FileSystem.GetFileSystemEntries(p)` returns a lazily-evaluated `IEnumerable` that walks the filesystem. The `.ToArray()` materializes it once. This result is then stored in `_cache` (a `ConcurrentDictionary<string, FileSystemMetadata[]>`) and returned for reuse by `GetDirectories()`, `GetFiles()`, etc. Without the `.ToArray()`, we'd either be storing a deferred enumerable that re-walks the filesystem on each access, or we'd need to change the dictionary's value type. The immediate `foreach` on `lines 34-37` is a secondary concern; it populates the per-file `_fileCache`. The `.ToArray()` is doing real work here.

Emby.Server.Implementations/Library/IgnorePatterns.cs
taos15: I see;  you mean extracting the substring patterns like `"trailer"`, `"sample"`, `"thumbs"` etc. into a `string[]` and looping with `Contains()`. That's fair for maintainability. The tradeoff is that the current explicit `if` chain allows different return values per match and the JIT can optimize each `string.Contains` call individually. But if we don't actually need per-pattern return values (which we currently don't; they all return `true`), then an array + loop would be cleaner.

Do you still want me to refactor it?

IDisposable: I think it would be cleaner... and we should make sure the order in the array is such that the most common occurring are listed first.

MediaBrowser.Controller/Providers/DirectoryService.cs
IDisposable: I don't think we're in sync, sorry.

What you're doing is calling .ToList() on an IEnumerator of file system entries. That returns a List<> of them (by enumerating and adding each entry)... then you enumerate that resulting list getting the same entries that were added. That means we're enumerating to get a list of entries, then enumerating the list.  Your concern about storing the enumerable itself would be valid, but we're not doing that.

What I'm suggesting is NOT getting rid of your loop, I'm saying that you doing the enumeration against the original IEnumerable is _exactly_ the same code as what's done by the .ToList() and thus would act the same way without the overhead of the intervening List() and extra enumeration.

```csharp
var entries = state.FileSystem.GetFileSystemEntries(p);
foreach (var entry in entries)
{
    state.FileCache.TryAdd(entry.FullName, entry);
}

return entries;
```

MediaBrowser.Model/Configuration/ServerConfiguration.cs
IDisposable: I guess I'm just quibbling on the `MB` on the name... it's not megabytes... it hundred-items...

MediaBrowser.Providers/Manager/ProviderManager.cs
IDisposable: Understood. However the switch to `!_disposed` **forced** you to remove the `using` in the callers [here](https://github.com/jellyfin/jellyfin/pull/16680/changes#diff-3af4a4e6a4d84ac699cfea27bfb7b223e73c80979ca00e84a81e36d307acc6fbR58-R71) which means they will no longer get the _required_ .Dispose() called against them.

MediaBrowser.Providers/Plugins/Tmdb/TmdbClientManager.cs
IDisposable: Yes, I suspect that simpler is better... and you're right... the logic I suggested _only_ covers the _fetch takes longer than coalesce_ window.  Seems like such an edge case we really shouldn't bother :)

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteDatabaseProvider.cs
IDisposable: I just wasn't sure what what returned from the SQLite query (e.g. if it included the pseudo meta-tables) :)

MediaBrowser.Model/Configuration/ServerConfiguration.cs
taos15: I will rename it to clarify that the items, not MB part.

theguymadmax:
@taos15 As mentioned on Matrix, substantial changes like this should first be drafted in our [meta repository](https://github.com/jellyfin/jellyfin-meta/discussions), explaining how it will technically be implemented before opening pull request(s).

Also, our roadmap is that the Jellyfin 12.0 release will address performance issues, and 13.0 will be a rewrite of the scanner by @JPVenson. So I don't see how this would fit into our timeline.

taos15:
> @taos15 As mentioned on Matrix, substantial changes like this should first be drafted in our [meta repository](https://github.com/jellyfin/jellyfin-meta/discussions), explaining how it will technically be implemented before opening pull request(s).
> 
> 
> 
> Also, our roadmap is that the Jellyfin 12.0 release will address performance issues, and 13.0 will be a rewrite of the scanner by @JPVenson. So I don't see how this would fit into our timeline.  

I’ll open a discussion as asked, but I felt it wouldn’t be fair not to respond to the comments after someone took the time to review this.

I can change this to a draft and/or target a different branch. We don’t need to implement this in the current major release; we can revisit it in the performance major release. That said, I think this is a good starting point for now, and I can try to keep it up to date and as close to master as possible, with feedback from the discussion or inline comments.

Rather than starting over, this could serve as a starting point. Is there a better branch I should target for feedback?

Thanks.

theguymadmax:
Draft mode sounds best; the master branch is correct. Updating the PR and addressing feedback is fine. I just didn’t want you to feel like your time was wasted working on it, since we’re planning a rewrite.

If this can be accepted or used as a starting point, that will be up to Venson and whatever comes out of the discussion.

taos15:
> Daft mode sounds best; the master branch is correct. Updating the PR and addressing feedback is fine. I just didn’t want you to feel like your time was wasted working on it, since we’re planning a rewrite.
> 
> 
> 
> If this can be accepted or used as a starting point, that will be up to Venson and whatever comes out of the discussion.
> 
> 

Thanks. I’ll change it to draft and try to keep it as synchronized with master as I can. Maybe before Venson starts working on the rework, he can give me some feedback so I can try to shape this toward what he has in mind. That said, the discussion should already have clarified those goals.

theguymadmax:
Closing per prior comments: A scanner rewrite is planned for 13.0 by Jellyfin Teams members (https://github.com/jellyfin/jellyfin-meta/discussions/125)

PR #16682, opened 2026-04-22
remove generic configuration endpoint and add explicit ones

KGT1:
**Changes**
This removes the generic Configuration endpoint, which implicitly allows all users to read all configuration files and admins to edit them all via the api. Adds the endpoints used in the Webclient back as explicit ones.

I think this is a better api design, since it doesnt require hacks to add the Models to OpenAPI and hopefully allows these endpoints to be kept stable.

Specifically removed are the endpoints:
```
POST/GET /System/Configuration/database
GET /System/Configuration/branding # has its own endpoint in BrandingController
```

and these are kept / added:

```
POST/GET /System/Configuration/encoding
POST/GET /System/Configuration/network
POST/GET /System/Configuration/metadata
POST/GET /System/Configuration/xbmcmetadata
POST/GET /System/Configuration/livetv
POST /System/Configuration/branding
```
But since the function is broken up, this is breaking OpenAPI.

This also breaks Plugins, that register an IConfigurationFactory and use the generic endpoint to access it (namely jellyfin-plugin-playbackreporting). They will need to add their own endpoint.

Edit:
ive created a [pr](https://github.com/jellyfin/jellyfin-plugin-playbackreporting/pull/126) for jellyfin-plugin-playbackreporting
and have one ready for jellyfin-web ([here](https://github.com/jellyfin/jellyfin-web/compare/master...KGT1:typedConfig))

in generated sdk's changes would boil down to something like this:
```
getNamedConfiguration('network')
updateNamedConfiguration('branding', brandingOptions)
```
would become
```
getNetworkConfiguration()
updateBrandingConfiguration(brandingOptions)
```

**Issues**
no issue but expressed [here](https://github.com/jellyfin/jellyfin/pull/13756#pullrequestreview-2728355839) and [here](https://github.com/jellyfin/jellyfin/pull/13756#issuecomment-2761941729)


nielsvanvelzen (changes requested):
Removing the generic endpoint is definitely the way forward. However we should NOT use the existing configuration types and introduce new DTO's for them instead. Similar to how the branding configuration currently works.

This is to avoid internal types leaking and allows the actual configuration to be detached from the API. There might be cases where we don't want a specific config option to ever be readable/writable from the API (like secrets or deprecated ones).

Jellyfin.Api/Controllers/ConfigurationController.cs
```diff
@@ -142,4 +230,388 @@ public ActionResult UpdateBrandingConfiguration([FromBody, Required] BrandingOpt
 
         return NoContent();
     }
+
+    // Mapping Helpers
+
+    private static ServerConfigurationDto MapToDto(ServerConfiguration config)
```
JPVenson: better to put them into a dedicated class instead of the controller

JPVenson (changes requested):
otherwise fine

MediaBrowser.Model/Configuration/EncodingOptionsDto.cs
```diff
@@ -0,0 +1,309 @@
...
+/// The encoding options DTO for API use.
+/// </summary>
+public class EncodingOptionsDto
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="EncodingOptionsDto" /> class.
+    /// </summary>
+    public EncodingOptionsDto()
+    {
+        EnableFallbackFont = false;
```
JPVenson, line 19: On secound thought, i am not feeling too happy about having the dto defining the default, a dto should be logic-less

JPVenson (changes requested):

nielsvanvelzen (approved):

PR #16685, opened 2026-04-23
added jpg image compression and .jellyignore file support to toggle ignoring from the frontend

Rani-Wehbe:
added new feature to enable or disable JellyIgnore for the .jellyignore files per library
this way users administrators can have maximum functionality over there libraries 
frontend pull request support: https://github.com/jellyfin/jellyfin-web/pull/7850

also added jpeg files compression to make them load much faster 

note: for backwards compatibility and security .ignore files are always ignored and no administrator account can change that from the frontend 

assistance from: used opencode with GitHub Copilot model claude haiku 4.5 free also used CodeGraph mcp


nielsvanvelzen:
Thank you for your contribution. Unfortunately we won't be accepting it as it violates our [LLM/"AI" Development Policy](https://jellyfin.org/docs/general/contributing/llm-policies).

PR #16688, opened 2026-04-23
Fix jellyfin#16152 : "Replace all metadata" Removes all metadata on movies without external id

### LINKED ISSUE #16152: "replace all metadata" removes all metadata on movies without any external id
### Description of the bug

All metadata is lost of movies i manual added the metadata (genre, cast, description, year and so on) when i select "replace all metadata".

### Reproduction steps

1. Add Metadata to an Movie
2. Select "Update Metadata"
3. Replace all Metadata

All Metadata except the Title is lost.
Also this happen if you run replace all metadata on the complete library.

The Metadata should only replaced if there are an "external id"

### Jellyfin Server version

10.11.6

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.6

### Environment

```markdown
I'm using Jellyfin as an Docker Container out from Truenas.
```

### Jellyfin logs

```shell
There are not any entries from the log to this behavior.
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

mfar14:
Files without remote metadata available (orphaned items) were having their manually edited metadata overwritten during automatic refresh operations.

**Changes**
This patch solves the metadata loss issue for files lacking external metadata sources through orphaned item detection in the MergeBaseItemData method of the MetadataService. When a file has no provider IDs (indicating no remote metadata was found), the system identifies it as an "orphaned source."

Rather than merging empty fields and overwriting user-edited metadata, the system skips the field merge entirely and performs an early return. This preserves all existing user-edited metadata. Only metadata settings (like locked fields or language preferences) are updated if explicitly requested, preventing accidental data loss during both automatic refreshes and explicit "Replace all metadata" actions.

**Issues**
Fixes #16152


IDisposable (approved):

mfar14:
@IDisposable Would you please ask another reviewer to confirm that everything is alright and approve the pull request? Thank You

mfar14:
@JPVenson, are you still reviewing this? Please let me know if you think any changes should be made.

MediaBrowser.Providers/Manager/MetadataService.cs
IDisposable, line 1127: Should this be testing `replaceData`?

IDisposable, line 1127: Should this be testing `replaceData`?

IDisposable, line 1122: Should this be testing `replaceData`?

IDisposable, line 1122: Should this be testing `replaceData`?

IDisposable, line 1103: Looks like this code block is essentially completely duplicate of the later version https://github.com/jellyfin/jellyfin/pull/16688/changes/6da4ac86b1de0f44be00999d6360ef160b3f4de9#diff-174eaed87dce839bc188df15984f50ea35700c6be976b375fe005c1fb3a9247bR1345-R1382

Perhaps it would be better to not duplicate the block and just surround the entire "rest of the fields" with an `if (!isOrphanedSource) {` and not do an early `return;`

IDisposable (changes requested):
After resolving the merge conflict, looks like we need to eliminate the duplicate code in MetadataService.cs

PR #16691, opened 2026-04-24
Preserve ffprobe stream indices so the subtitle cache stays aligned

ignacio82:
**Changes**

Stop renumbering embedded `MediaStream.Index` in `FFProbeVideoInfo.Fetch`.
Embedded streams are added to the working list first and keep the native
ffprobe stream index they were given by `ProbeResultNormalizer`; external
subtitle / audio streams are then appended afterwards, where
`AddExternalSubtitlesAsync` / `AddExternalAudioAsync` already compute
`startIndex = max(existing Index) + 1`, so they land past the embedded
range instead of colliding with it.

**Why**

`MediaStream.Index` is both the cache key under
`/config/data/data/subtitles/<dir>/<itemId>/<Index>.<ext>` and the value
the subtitle extractor uses (via `EncodingHelper.FindIndex`) to compute
the ffmpeg `-map 0:N` argument. Both paths implicitly assume it matches
the physical stream position in the file.

Previously, when external subs were appended to a freshly-probed item,
they claimed `Index=0` and every embedded stream's `Index` was bumped
sequentially — silently discarding the real ffprobe index. Extracted
subtitle cache files on disk were keyed by the *old* `Index`, but the
next request read them back by the *new* `Index`, so the controller
started serving mismatched content. Concretely: after dropping an
`.srt` sibling beside an MKV that already had its embedded subs
extracted, selecting a given embedded subtitle track renders the
content of whichever stream used to live at that `Index` — not the one
the label points at.

**Reproduction (Jellyfin 10.11.8)**

1. Play an MKV with several embedded text subtitle tracks, letting
   Jellyfin extract and cache them to
   ` /config/data/data/subtitles/…/<id>/<Index>.ass `.
2. Drop an external `.srt` sibling next to the video (or have the
   OpenSubtitles downloader do it) and refresh metadata.
3. The external SRT is appended with `Index=0`; every embedded stream's
   `Index` shifts by +1; cache files on disk are untouched.
4. Request `/Videos/{item}/{src}/Subtitles/<Index>/0/Stream.<ext>` for
   any embedded subtitle — the response is the *previous* occupant of
   that `Index`. For a file with English / Spanish / Thai / Vietnamese
   subs, selecting \"Spanish (Spain)\" returns the Thai bytes, etc.

**Compatibility**

External streams move from low `Index` values (typically 0, 1, …) to
positions past the highest embedded index (e.g. 21, 22, …).
`MediaStream.Index` is documented as \"Stream index\" (ffprobe-native),
not \"offset in the MediaStreams array\", and a \`grep -rn\` across the
server codebase turns up no caller that assumes externals sit at low
indices — existing consumers look up by \`Index == N\` or iterate.

Previously-cached subtitle files that were extracted *while* the bug
was active will remain mis-keyed until manually wiped (delete
\`/config/data/data/subtitles/<dir>/<itemId>/\` for the affected item,
or the whole subtitle cache root, and let Jellyfin re-extract on next
playback). Happy to follow up with a one-time cache-invalidation
migration if reviewers think that's warranted.

**Issues**

(no existing issue; this was bisected from a downstream client report)

### DIFF
--- a/MediaBrowser.Providers/MediaInfo/FFProbeVideoInfo.cs
+++ b/MediaBrowser.Providers/MediaInfo/FFProbeVideoInfo.cs
@@ -194,20 +194,17 @@ namespace MediaBrowser.Providers.MediaInfo
             IReadOnlyList<MediaAttachment> mediaAttachments;
             ChapterInfo[] chapters;
 
-            // Add external streams before adding the streams from the file to preserve stream IDs on remote videos
-            await AddExternalSubtitlesAsync(video, mediaStreams, options, cancellationToken).ConfigureAwait(false);
-
-            await AddExternalAudioAsync(video, mediaStreams, options, cancellationToken).ConfigureAwait(false);
-
-            var startIndex = mediaStreams.Count == 0 ? 0 : (mediaStreams.Max(i => i.Index) + 1);
-
+            // Add embedded streams FIRST, preserving the ffprobe-native stream indices
+            // stored in MediaStream.Index. The subtitle cache on disk is keyed by this
+            // Index (see SubtitleEncoder.GetSubtitleCachePath), and FindIndex-based
+            // ffmpeg `-map 0:N` extraction also depends on Index matching the physical
+            // stream position. If we renumber embedded streams after externals have
+            // claimed low indices, a later refresh that adds or removes an external
+            // subtitle will shift all embedded Index values and the cache files will be
+            // served under the wrong labels (e.g. selecting "Spanish" shows Thai).
             if (mediaInfo is not null)
             {
-                foreach (var mediaStream in mediaInfo.MediaStreams)
-                {
-                    mediaStream.Index = startIndex++;
-                    mediaStreams.Add(mediaStream);
-                }
+                mediaStreams.AddRange(mediaInfo.MediaStreams);
 
                 mediaAttachments = mediaInfo.MediaAttachments;
                 video.TotalBitrate = mediaInfo.Bitrate;
@@ -231,7 +228,6 @@ namespace MediaBrowser.Providers.MediaInfo
                 {
                     if (!mediaStream.IsExternal)
                     {
-                        mediaStream.Index = startIndex++;
                         mediaStreams.Add(mediaStream);
                     }
                 }
@@ -239,6 +235,14 @@ namespace MediaBrowser.Providers.MediaInfo
                 mediaAttachments = [];
                 chapters = [];
             }
+
+            // External streams get Index values AFTER the embedded range.
+            // AddExternalSubtitlesAsync / AddExternalAudioAsync already compute their
+            // startIndex as `max(existing Index) + 1`, so they fall in sequence after
+            // the highest embedded stream.
+            await AddExternalSubtitlesAsync(video, mediaStreams, options, cancellationToken).ConfigureAwait(false);
+
+            await AddExternalAudioAsync(video, mediaStreams, options, cancellationToken).ConfigureAwait(false);
 
             var libraryOptions = _libraryManager.GetLibraryOptions(video);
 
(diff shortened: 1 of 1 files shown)


Shadowghost:
This will break strm support with external subs/audio.
The whole reason why we add external first is because strms are probed on playback, so we can't populate internal stream indices (and therefore appended external streams), breaking playback

MBR-0001:
I was messing around with this part of the code in [this branch](https://github.com/MBR-0001/jellyfin/tree/fix-ffprobe-subs), more specifically [this commit](https://github.com/MBR-0001/jellyfin/commit/0ac9d2f05b94a7ccae09032e5acfff92483fab7c) in order to fix a bug where subtitles would get downloaded unnecessarily (check commit description)

would that be of any use?
I didn't open a PR yet because I didn't have enough time to test this, I believe this is a very important part that could mess things up, so I planned to do thorough testing

ignacio82:
@Shadowghost thanks for catching this — you're right that the original v1 patch broke the "external `Index` is stable across the no-probe → probe transition" guarantee for `.strm` shortcuts. I traced the flow and the concern is concrete:

- **Library scan** on a fresh `.strm`: `EnableRemoteContentProbe=false` → `mediaInfo=null` → no internal streams known → with v1, externals landed at `0..K-1`.
- **Playback** of the same `.strm`: `MediaSourceManager.GetPlaybackMediaSources` forces a refresh with `EnableRemoteContentProbe=true` → probe returns internals at ffprobe-native indices `0..N` → with v1, externals were re-assigned to `max(internal)+1 = N+1..N+K`.

So an external sub's `Index` shifted between the two snapshots, which is exactly what the original `// preserve stream IDs on remote videos` comment guarded against.

I've pushed [f3ee4359](https://github.com/ignacio82/jellyfin/commit/f3ee435950) which keeps the cache-mismatch fix (embedded streams retain their ffprobe-native `Index`) and also keeps external `Index` stable across both snapshots, by anchoring externals at a fixed high offset:

```csharp
private const int ExternalStreamIndexBase = 1_000_000;

private static int GetExternalStartIndex(List<MediaStream> currentStreams)
{
    if (currentStreams.Count == 0)
    {
        return ExternalStreamIndexBase;
    }
    var maxExisting = currentStreams.Max(i => i.Index);
    return Math.Max(maxExisting + 1, ExternalStreamIndexBase);
}
```

Both `AddExternalSubtitlesAsync` and `AddExternalAudioAsync` now compute their `startIndex` via this helper. Properties of the new scheme:

- Embedded streams keep their ffprobe-native `Index` (cache stays aligned, `EncodingHelper.FindIndex` `-map 0:N` is unaffected).
- External streams always start at `1_000_000`, regardless of whether internals are known yet, so a `.strm`'s externals get the same `Index` in the no-probe snapshot and the post-probe snapshot.
- ffprobe doesn't return indices anywhere near `1_000_000` for real media files, so collisions are impossible in practice. The `Math.Max` guard handles the pathological case where `mediaStreams.Max(i.Index) ≥ 1_000_000`.

This does change the externally-visible magnitude of external `Index` values from low integers to `1_000_000+`. `MediaStream.Index` is `int`, the API serializes it as an integer, and a `grep -rn` across the server turned up no caller that assumes externals sit at low indices — they're looked up by `Index == N` or by iteration. Persisted user prefs (`UserItemData.SubtitleStreamIndex` / `AudioStreamIndex`) for external tracks would need to be re-picked once after upgrade, but the indices were already going to change once for the v1 patch, so this isn't an additional cost; if reviewers want a one-shot migration that maps the old external `Index` to the new range I'm happy to add it.

@MBR-0001 — your branch tackles a separate (also real) bug in the subtitle-download skip logic. I think those two PRs are independent and don't conflict; happy to coordinate if helpful.

dsmaugy:
Whoops I just made https://github.com/jellyfin/jellyfin/pull/16712 which I think is also trying to solve the same problem as this PR. I was considering changing the indexing scheme as well since it seems like a more apt fix, but wouldn't that be a breaking change as all current caches will be invalidated?

JPVenson:
https://jellyfin.org/docs/general/contributing/llm-policies#general-guidelines

PR #16698, opened 2026-04-25
improve openapi endpoint categorization

dkanada:
**Changes**

Initial effort to better categorize endpoints in the OpenAPI specification. This commit only includes the low hanging fruit but there are a number of other potential changes that require some discussion.

1. Should *all* images go to their respective categories? Or just endpoints like `/Branding` which have existing controllers?
2. Merge `HlsSegment` and `DynamicHls` into `HLS` or move to `Audio` and `Video` respectively?
3. Should the password / QuickConnect / reset endpoints move to `Auth` or stay in the `Users` category?
4. We have a mix of singular and plural category names - which one is preferable?

**Issues**

None

### DIFF
--- a/Jellyfin.Api/Controllers/ActivityLogController.cs
+++ b/Jellyfin.Api/Controllers/ActivityLogController.cs
@@ -19,6 +19,7 @@ namespace Jellyfin.Api.Controllers;
 /// </summary>
 [Route("System/ActivityLog")]
 [Authorize(Policy = Policies.RequiresElevation)]
+[Tags("System")]
 public class ActivityLogController : BaseJellyfinApiController
 {
     private readonly IActivityManager _activityManager;
--- a/Jellyfin.Api/Controllers/ApiKeyController.cs
+++ b/Jellyfin.Api/Controllers/ApiKeyController.cs
@@ -14,6 +14,7 @@ namespace Jellyfin.Api.Controllers;
 /// Authentication controller.
 /// </summary>
 [Route("Auth")]
+[Tags("Authentication")]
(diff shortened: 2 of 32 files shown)


Shadowghost (approved):

nielsvanvelzen (approved):
Considering the controller names are messy I'd argue we should **always** add the Tags annotation to them, even if the name is equal. (could be separate PR though)

To answer your questions:

**Should all images go to their respective categories? Or just endpoints like /Branding which have existing controllers?**

I'd say all image endpoints in the same API makes more sense, as they're all the same thing with different inputs.

**Merge HlsSegment and DynamicHls into HLS or move to Audio and Video respectively?**

Yes, but we should really exclude them from the OpenAPI spec. Those endpoints are NEVER meant to be used directly, instead you should only use them via the TranscodingUrl returned by the server on the media info API.

**Should the password / QuickConnect / reset endpoints move to Auth or stay in the Users category?**
I'd say auth. User category should be stuff like getting your own user, listing other users, updating a user etc. Unrelated to authentication.

**We have a mix of singular and plural category names - which one is preferable?**
I'd argue singular, in our SDK's we often add the "Api" suffix and a name like "SystemApi" reads better compared to "SystemsApi"

dkanada:
This pull request is now focused exclusively on tags applied at the controller level. I will open successive PRs to move individual endpoints to new categories and hide HLS / deprecated endpoints from the specification.

nielsvanvelzen (approved):

Bond-009 (approved):

PR #16220, opened 2026-02-11
Fix EPG issues

Shadowghost:
Went through the EPG issues and implemented some fixes. Needs testing by someone who actually has LiveTV and EPG configured

**Changes**

SchedulesDirect API fixes
*  Fixed `/metadata/programs` 1008 errors: Added required trailing slash to POST URL, send full 14-char program IDs instead of deprecated 10-char truncation, use proper `JsonContent.Create serialization` instead of manual `StringBuilder JSON`
* Added 500-item batching for `/metadata/programs/` requests per SD API spec
* Parsed SD error codes from response body in `Request<T>` using `JsonDocument`, with targeted handling per error type:
  * `4001/4003/4004/4005/4008` (permanent account errors): disable SD for server lifetime
  * `4009/4010` (transient login errors): 30-minute backoff
  * `5002` (max image downloads): block image requests until midnight UTC
  * `5003` (max metadata requests): block metadata requests until midnight UTC
  * `4006/403` (token expired): clear token and retry once
* Image and metadata limits are independent
* Daily limits persisted to cache files (`sd-image-limit.txt`, `sd-metadata-limit.txt`) so they survive server restarts
* Cached `/available/countries` response to file with 7-day TTL + in-memory cache

Image deserialization resilience
* New `ImageDataArrayConverter` handles SD returning non-array values (e.g. string error messages) for the data field in image responses, preventing JsonException crashes
* Uses `TrySkip()` instead of `Skip()` for streaming deserialization compatibility

Guide image pre-caching
* SD-aware early termination in `PreCacheImages`: detects SD image download failures, sets a global flag, logs once, and skips remaining SD images for the rest of the scan
* Persists across channel scans via `_sdImageLimitHitDate` instance field that auto-clears after midnight UTC

Controller cleanup
* Moved countries fetch logic out of the controller into SchedulesDirect
* Controller delegates to `ISchedulesDirectService.GetAvailableCountries()`
* Removed unused `IHttpClientFactory` and `IConfigurationManager` dependencies

**Issues**
- Fixes #2548
- Fixes #6103
- Fixes #13147
- Fixes #13637
- Fixes #15399
- Fixes #15782

PR #16701, opened 2026-04-25
Fix stale XMLTV cache for local guide updates

### LINKED ISSUE #6103: Changes in xmltv requires manual purge of caches

itz4blitz:
**Changes**
Refresh local XMLTV caches when a local guide source is newer while keeping the existing age-based cache behavior for remote guides. Add regression coverage for local XML and gzip XMLTV updates, plus a higher-level Live TV service test that proves guide refresh rewrites the `*_channels` cache and surfaces updated channels and programs without manual cache cleanup.

**Issues**
Fixes #6103

**Testing**
- dotnet test \"tests/Jellyfin.LiveTv.Tests/Jellyfin.LiveTv.Tests.csproj\" -c Release

### DIFF
--- a/src/Jellyfin.LiveTv/Listings/XmlTvListingsProvider.cs
+++ b/src/Jellyfin.LiveTv/Listings/XmlTvListingsProvider.cs
@@ -65,7 +65,7 @@ namespace Jellyfin.LiveTv.Listings
 
             if (File.Exists(cacheFile))
             {
-                if (File.GetLastWriteTimeUtc(cacheFile) >= DateTime.UtcNow.Subtract(_maxCacheAge))
+                if (CanUseCache(info.Path, cacheFile))
                 {
                     return cacheFile;
                 }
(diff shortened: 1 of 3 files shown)


theguymadmax:
This should already be address in the master by #16220

PR #16705, opened 2026-04-26
Add .jellyignore support with per-library configuration

Rani-Wehbe:
## Summary

Implement `.jellyignore` file support that complements the existing `.ignore` mechanism, allowing users to optionally filter files and folders during library scans. The feature is configurable per-library via admin UI, while `.ignore` files remain always enforced as a security boundary.

## Motivation

The `.ignore` mechanism provides essential security by preventing malicious or accidental inclusion of unwanted files in libraries. The new `.jellyignore` feature extends this with optional, user-controlled file filtering that can be enabled/disabled per library. This allows power users to leverage gitignore style patterns while maintaining backwards compatibility with existing `.ignore` behavior.

## Changes

**Backend (jellyfin server)**:
- `DotIgnoreIgnoreRule.cs`: Implements .jellyignore pattern matching with support for:
  - Gitignore-style patterns (`*.txt`, `folder/`, negations with `!`)
  - Empty ignore files (ignore entire directory)
  - Invalid regex patterns (gracefully skipped)
  - Cross-OS path normalization (Windows/Unix)
  - Per-library and global configuration with proper precedence
                                (basically like .ignore for simplicity)
- `ServerConfiguration.cs`: Added `EnableJellyignore` property (default: true)
- `LibraryOptions.cs`: Added per-library `EnableJellyignore` toggle

## How It Works
1. `.ignore` files are always enforced as a security boundary
2. `.jellyignore` files are optional, controlled by `EnableJellyignore` setting
3. Per-library setting takes precedence over global setting
4. Patterns use standard gitignore syntax
5. Empty ignore files ignore the entire directory

## Testing
- Empty ignore files correctly ignore entire directories
- Gitignore-style patterns matched correctly (*.txt, folder/, etc.)
- Invalid regex patterns handled gracefully without crashing
- Both `.ignore` and `.jellyignore` coexist properly
- Path normalization works cross-platform (Debian, macOS, TrueNAS (in docker I didnt test this yet), Windows)

## Breaking Changes
None. `.jellyignore` is optional; existing `.ignore` behavior unchanged.


nielsvanvelzen:
This could be implemented as a plugin, there's no need to add this to the core.

PR #16707, opened 2026-04-26
Trim usernames on authentication

### LINKED ISSUE #8897: [Issue]: Can't login if there is space at the end of user name
### Please describe your bug

Just like in title. I think when somebody type their name on phone, their keyboard suggest complete it to their user name, but after that, their keyboard set space at the end (that's useful for messengers i guess). And, user can't login. I think it can be fixed by regex? By the way, that affects all clients on different platforms.

### Jellyfin Version

Other

### if other:

10.8.8.0

### Environment

```markdown
- OS: Debian
- Virtualization:
- Clients: Jellyfin Android, Jellyfin IOS
- Browser: Firefox, Google Chrome
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- Plugins:
- Reverse Proxy: nginx
- Base URL:
- Networking:
- Storage:
```

### Jellyfin logs

```shell
[ERR] Error authenticating with provider "Default"
MediaBrowser.Controller.Authentication.AuthenticationException: Specified user does not exist.
   at Jellyfin.Server.Implementations.Users.DefaultAuthenticationProvider.Authenticate(String username, String password, User resolvedUser)
   at Jellyfin.Server.Implementations.Users.UserManager.AuthenticateWithProvider(IAuthenticationProvider provider, String username, String password, User resolvedUser)
[INF] Authentication request for "User " has been denied (IP: "<IP>").
[ERR] Error processing request: "Invalid username or password entered". URL "POST" "/Users/authenticatebyname".

...

[INF] Authentication request for "User" has succeeded.

# notice the space in the authentication request
```

### FFmpeg logs

_No response_

_No response_

_No response_

### Code of Conduct

bxdxnn:
**Changes**
Trim usernames in the UserManager
**Issues**
Fixes #8897


src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/Migrations/JellyfinDbModelSnapshot.cs
```diff
@@ -15,7 +15,7 @@ partial class JellyfinDbModelSnapshot : ModelSnapshot
         protected override void BuildModel(ModelBuilder modelBuilder)
         {
 #pragma warning disable 612, 618
-            modelBuilder.HasAnnotation("ProductVersion", "9.0.9");
+            modelBuilder.HasAnnotation("ProductVersion", "10.0.7");
```
JPVenson: why the mayor version upgrade?

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/Migrations/20260426115901_TrimUsernameWhitespaces.cs
```diff
@@ -0,0 +1,24 @@
...
+namespace Jellyfin.Server.Implementations.Migrations
+{
+    /// <inheritdoc />
+    public partial class TrimUsernameWhitespaces : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.Sql(
+                "UPDATE \"Users\" SET \"Username\" = TRIM(\"Username\") WHERE \"Username\" != TRIM(\"Username\");");
```
JPVenson: Use raw string literals for that.

Also how would we handle conflicts here?

JPVenson (changes requested):
You should also add preprocessing for

https://github.com/jellyfin/jellyfin/blob/48299ff69450edb9fa6c59e228f32c7b7fb8fded/Jellyfin.Server.Implementations/Users/DefaultPasswordResetProvider.cs#L74

This would also introduce inconsistancies in for example
https://github.com/jellyfin/jellyfin/blob/48299ff69450edb9fa6c59e228f32c7b7fb8fded/Emby.Server.Implementations/Session/SessionManager.cs#L1607C49-L1607C56

where the dto still has potentiality whitespaces which is then used in other parts of the code. IMO if we disallow leading and tailing whitespaces in usernames we should validate against that on the api level instead.

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/Migrations/JellyfinDbModelSnapshot.cs
bxdxnn: Probably because of my .NET version, I've reverted.

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/Migrations/20260426115901_TrimUsernameWhitespaces.cs
bxdxnn: I've added conflict handling by appending a number at the end of the username.

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/Migrations/20260426173709_TrimUsernames.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            migrationBuilder.Sql(
+                """
+                WITH ranked AS (
+                    SELECT "Id", "Username",
+                           ROW_NUMBER() OVER (PARTITION BY TRIM("Username") ORDER BY LENGTH("Username")) as rn
+                    FROM "Users"
+                )
+                UPDATE "Users"
+                SET "Username" = TRIM("Username") || ' ' || ranked.rn
```
IDisposable: As I read this, we're aiming to take `"Marc"`, "'Marc "`, and `"Marc  "` to emit `"Marc"`, `"Marc 2"`, and `"Marc 3"`... which seems quite reasonable. The intent is to avoid unique Username collisions, but it _will_ fail if there's already a `"Marc 2"` from preexisting data.  I wonder if we should really try to be more bullet-proof and just use `NEWID()` instead of the `ranked.rn`.

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/Migrations/20260426173709_TrimUsernames.cs
```diff
@@ -0,0 +1,40 @@
...
+    public partial class TrimUsernames : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.Sql(
+                """
+                WITH ranked AS (
+                    SELECT "Id", "Username",
+                           ROW_NUMBER() OVER (PARTITION BY TRIM("Username") ORDER BY LENGTH("Username")) as rn
```
IDisposable: Probably should also add `, Id` to the `ORDER BY` as a tie-break in case there's a `" Marc "` and a  `"Marc  "` (e.g. both losing 2 spaces in the trim... or even a tab?)

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/Migrations/20260426173709_TrimUsernames.cs
IDisposable: _resolved by appending the user's ID, so GTG_

bxdxnn:
@JPVenson any news?

theguymadmax:
Is this still needed after: https://github.com/jellyfin/jellyfin/pull/13556 & https://github.com/jellyfin/jellyfin-web/pull/6528?

bxdxnn:
> Is this still needed after: #13556 & [jellyfin/jellyfin-web#6528](https://github.com/jellyfin/jellyfin-web/pull/6528)?

These doesn't fix #8897, the space wouldn't be trimmed on login. The PRs you've provided only disallow spaces for new users and trims the name on user creation in Web.

bxdxnn:
It's been 4 months ever since, any status so far?

PR #16715, opened 2026-04-27
hide HLS controllers and deprecate unused endpoints

dkanada:
**Changes**

The most notable change here is the removal of the internal HLS endpoints, but I also hid existing obsolete endpoints and added a few more to the list.

MusicGenres/InstantMix: Seems to be unused in the web client. Superceded by a different route.
MusicGenres/genreName: Also unused and replaced by the corresponding `Genres/genreName` endpoint.
TmdbController: Part of a plugin that we don't want to see in the OpenAPI specification file.

**Issues**

None

### DIFF
--- a/Jellyfin.Api/Controllers/InstantMixController.cs
+++ b/Jellyfin.Api/Controllers/InstantMixController.cs
@@ -320,6 +320,7 @@ public class InstantMixController : BaseJellyfinApiController
     [ProducesResponseType(StatusCodes.Status200OK)]
     [ProducesResponseType(StatusCodes.Status404NotFound)]
     [Obsolete("Use GetInstantMixFromArtists")]
+    [ApiExplorerSettings(IgnoreApi = true)]
     public ActionResult<QueryResult<BaseItemDto>> GetInstantMixFromArtists2(
         [FromQuery, Required] Guid id,
         [FromQuery] Guid? userId,
@@ -358,6 +359,7 @@ public class InstantMixController : BaseJellyfinApiController
     [HttpGet("MusicGenres/InstantMix")]
     [ProducesResponseType(StatusCodes.Status200OK)]
     [ProducesResponseType(StatusCodes.Status404NotFound)]
+    [Obsolete("Use GetInstantMixFromMusicGenreByName")]
     public ActionResult<QueryResult<BaseItemDto>> GetInstantMixFromMusicGenreById(
         [FromQuery, Required] Guid id,
         [FromQuery] Guid? userId,
--- a/Jellyfin.Api/Controllers/LibraryController.cs
+++ b/Jellyfin.Api/Controllers/LibraryController.cs
@@ -114,20 +114,6 @@ public class LibraryController : BaseJellyfinApiController
     }
 
     /// <summary>
-    /// Gets critic review for an item.
-    /// </summary>
-    /// <response code="200">Critic reviews returned.</response>
-    /// <returns>The list of critic reviews.</returns>
-    [HttpGet("Items/{itemId}/CriticReviews")]
-    [Authorize]
-    [Obsolete("This endpoint is obsolete.")]
-    [ProducesResponseType(StatusCodes.Status200OK)]
-    public ActionResult<QueryResult<BaseItemDto>> GetCriticReviews()
-    {
-        return new QueryResult<BaseItemDto>();
-    }
-
-    /// <summary>
     /// Get theme songs for an item.
     /// </summary>
     /// <param name="itemId">The item id.</param>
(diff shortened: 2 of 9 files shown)


MediaBrowser.Providers/Plugins/Tmdb/Api/TmdbController.cs
```diff
@@ -14,6 +14,7 @@ namespace MediaBrowser.Providers.Plugins.Tmdb.Api
     [Authorize]
     [Route("[controller]")]
     [Produces(MediaTypeNames.Application.Json)]
+    [ApiExplorerSettings(IgnoreApi = true)]
```
nielsvanvelzen, line 17: Plugin controllers are allowed to show up in the OpenAPI spec, but this one is a bit of a special case because the plugin is shipped with server so our published documents at api.jellyfin.org currently include it.
So not entirely sure what the correct approach here is: keep it for now and it will be gone once the plugin is extracted or hide it and unhide after extraction.

Jellyfin.Api/Controllers/PluginsController.cs
```diff
@@ -137,6 +137,7 @@ public ActionResult UninstallPluginByVersion([FromRoute, Required] Guid pluginId
     [ProducesResponseType(StatusCodes.Status204NoContent)]
     [ProducesResponseType(StatusCodes.Status404NotFound)]
     [Obsolete("Please use the UninstallPluginByVersion API.")]
+    [ApiExplorerSettings(IgnoreApi = true)]
     public ActionResult UninstallPlugin([FromRoute, Required] Guid pluginId)
```
nielsvanvelzen, line 139: I'm actually a bit confused why this one is even deprecated, I see enough use-cases to allow a complete plugin uninstall instead of just uninstalling specific versions.
So I'd rather remove the deprecation in this one.

Jellyfin.Api/Controllers/MusicGenresController.cs
```diff
@@ -145,6 +146,8 @@ public ActionResult<QueryResult<BaseItemDto>> GetMusicGenres(
     /// <returns>An <see cref="OkResult"/> containing a <see cref="BaseItemDto"/> with the music genre.</returns>
     [HttpGet("{genreName}")]
     [ProducesResponseType(StatusCodes.Status200OK)]
+    [Obsolete("Use GetGenre instead")]
+    [ApiExplorerSettings(IgnoreApi = true)]
```
nielsvanvelzen: We shouldn't hide endpoints immediately. At least 1 major release should be pushed out with just the deprecation, so the SDK's will message this deprecation to client developers first.

Jellyfin.Api/Controllers/LiveTvController.cs
```diff
@@ -953,6 +955,7 @@ public async Task<ActionResult> CreateSeriesTimer([FromBody] SeriesTimerInfoDto
     [Authorize(Policy = Policies.LiveTvAccess)]
     [ProducesResponseType(StatusCodes.Status404NotFound)]
     [Obsolete("This endpoint is obsolete.")]
+    [ApiExplorerSettings(IgnoreApi = true)]
```
nielsvanvelzen: This endpoint we can just remove entirely right now, it already returns a 404 response. Which is the same behavior for a non-existent endpoint.

Jellyfin.Api/Controllers/LibraryController.cs
```diff
@@ -121,6 +121,7 @@ public ActionResult GetFile([FromRoute, Required] Guid itemId)
     [HttpGet("Items/{itemId}/CriticReviews")]
     [Authorize]
     [Obsolete("This endpoint is obsolete.")]
+    [ApiExplorerSettings(IgnoreApi = true)]
```
nielsvanvelzen: Although this one does return a 200 response, it's response is ALWAYS empty and it does not look like any of our official clients use it. I'd say we can remove this directly as the deprecation was 6 years ago.

Jellyfin.Api/Controllers/InstantMixController.cs
```diff
@@ -358,6 +359,8 @@ public ActionResult<QueryResult<BaseItemDto>> GetInstantMixFromArtists2(
     [HttpGet("MusicGenres/InstantMix")]
     [ProducesResponseType(StatusCodes.Status200OK)]
     [ProducesResponseType(StatusCodes.Status404NotFound)]
+    [Obsolete("Use GetInstantMixFromMusicGenreByName")]
+    [ApiExplorerSettings(IgnoreApi = true)]
```
nielsvanvelzen: We shouldn't hide endpoints immediately. At least 1 major release should be pushed out with just the deprecation, so the SDK's will message this deprecation to client developers first.

nielsvanvelzen (changes requested):

MediaBrowser.Providers/Plugins/Tmdb/Api/TmdbController.cs
dkanada: I chose the latter method because I hope to remove the embedded plugins very soon, although not before the next update.

nielsvanvelzen (approved):

Bond-009 (approved):

Shadowghost (approved):

Issue #14750, opened 2025-09-05
Unstable/RC5: Song pages link to empty "People" (no Ids, no images, links to TMDB not Musicbrainz)


whyman:
### Description of the bug

In UI on song pages there is a new People section, but this is never populated. 

The images are always blank.

You cannot look up the "People" on Musicbrainz or AudioDB, and even if you could, the Ids are copied from the tracks

There is duplication between the "Album Artist" person and the "Artist" person?

### Reproduction steps

1. Go to an Album
2. See  empty People stubs
3. Be sad.

### What is the current _bug_ behavior?

Music artists "People" are always blank

### What is the expected _correct_ behavior?

People should be populated with data from Musicbrainz/AudioDB like they are on artists

### Jellyfin Server version

Unstable

### Specify unstable release number

2025090105

### Specify version number

10.11.RC

### Specify the build version

2025090105

### Environment

```markdown
- OS: Ubuntu 22.04 Docker
```

### Jellyfin logs

```shell
N/A
```

### FFmpeg logs

```shell

```

### Relevant screenshots or videos

Empty and duplicated people
<img width="1864" height="845" alt="Image" src="https://github.com/user-attachments/assets/51e367e9-ca3c-4c64-bcb6-58edebbd89c3" />
People have no IDs, and empty "Film" style Ids
<img width="1856" height="1221" alt="Image" src="https://github.com/user-attachments/assets/fe13139c-5ae2-4780-b6c4-e52ddbf54ecf" />

ghost:
This is fixed in https://github.com/jellyfin/jellyfin-web/pull/7109.

whyman:
Why are people created if they are not supposed to be shown anywhere?

ghost:
That is one of the great mysteries of the universe.  I have the impression that artists started as a kind of person but then it was realised that they are different, eg. metadata sources, or even that artists use square portraits over the rectangular ones used for other people, and then another way to track them was introduced.  Or it could be the other way around.

Technically in the database a person has a person type of Artist or Album Artist but there is a base assumption that a person is a theatrical person and not a musician.  There is then separate tracking of artists on albums and songs through their own columns in BaseItems.

There is more opportunity for classification of Artists if they were better merged with People; for example People can include Composers and Conductors, whereas Artists are only Artisrs or Album Artists.  Maybe a project for after 10.11 is released.

whyman:
[...]

PR #16721, opened 2026-04-28
return music artists from person endpoints

dkanada:
**Changes**

Scans triggered after this pull request is merged will populate the Peoples table with Artist and AlbumArtist entities from all music libraries. The goal is to improve person management and migrate away from hacks that have been used to store people as BaseItem entities. We use PremiereDate for DateOfBirth, ProductionLocations for PlaceOfBirth, Overview for Biography, and lack support for other common fields like DateOfDeath. It will also more closely match APIs from popular metadata providers as seen below.

**Are there separate endpoints for artist "groups" versus individual performers?**

MusicBrainz: Same endpoint with a Person / Group / Orchestra / Band / etc type field.
TheAudioDB: Same endpoint with an intMembers field.
IMDb: Groups and individuals use the `nm*` entity but are not nested.
TheTVDB: Groups don't exist.
OpenLibrary: Groups don't exist.
Goodreads: Groups don't exist.

**Do companies have their own entity or are they only tracked with a field? Relevant for `Studios` endpoint.**

MusicBrainz: Distributors, publishers, labels, etc are stored in a unique entity.
TheAudioDB: Labels are tracked as `strLabel` and `idLabel` on artist and album endpoints.
IMDb: Very comprehensive support with a separate `co*` endpoint.
TheTVDB: Same as IMDb with a unique endpoint and CompanyType field.
OpenLibrary: Publishers tracked in an array on individual book editions.
Goodreads: Publishers (not distributors) are only saved as a string field on books.

**Issues**

None

### DIFF
--- a/Jellyfin.Api/Controllers/ArtistsController.cs
+++ b/Jellyfin.Api/Controllers/ArtistsController.cs
@@ -87,6 +87,7 @@ public class ArtistsController : BaseJellyfinApiController
     /// <returns>An <see cref="OkResult"/> containing the artists.</returns>
     [HttpGet]
     [ProducesResponseType(StatusCodes.Status200OK)]
+    [Obsolete("Use GetPersons")]
     public ActionResult<QueryResult<BaseItemDto>> GetArtists(
         [FromQuery] double? minCommunityRating,
         [FromQuery] int? startIndex,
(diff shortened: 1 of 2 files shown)


IDisposable (approved):
Most of the PR description is not actually part of _this_ change, but useful information. Perhaps reference a new issue and ensure that information lives somewhere until implemented.

Bond-009 (approved):

theguymadmax (changes requested):
Requires web changes first, other wise we'll get people in the album view. So person will be will be listed three times for an album. 
1) Person - Album Artist
2) Person - Artist
3) Artist

theguymadmax:
Ex: https://github.com/jellyfin/jellyfin/issues/14750

theguymadmax (approved):

nielsvanvelzen (approved):

IDisposable (approved):

PR #16733, opened 2026-04-30
Add personalised default profile images

JPVenson:
**Changes**
Adds default profile image generation and migration to a colored background based on the Username with Black letters composing of the Username.

image

### DIFF
--- a/MediaBrowser.Controller/Drawing/IProfileImageService.cs
+++ b/MediaBrowser.Controller/Drawing/IProfileImageService.cs
@@ -1,0 +1,35 @@
+using System;
+using System.IO;
+using System.Threading.Tasks;
+using Jellyfin.Database.Implementations.Entities;
+
+namespace MediaBrowser.Controller.Drawing
+{
+    /// <summary>
+    /// Service for generating default profile images.
+    /// </summary>
+    public interface IProfileImageService
+    {
+        /// <summary>
+        /// Generates a PNG profile image for the given display name and user identity.
+        /// The returned stream must be disposed by the caller; storage is the caller's responsibility.
+        /// </summary>
+        /// <param name="displayName">The display name used to derive the initials rendered on the image.</param>
+        /// <param name="userId">The user's unique identifier, used for deterministic color selection.</param>
+        /// <returns>A <see cref="Stream"/> containing PNG image data.</returns>
+        Task<Stream> GenerateProfileImageAsync(string displayName, Guid userId);
+
+        /// <summary>
+        /// Generates a profile image for the user, saves it to the user's profile image path,
+        /// and updates the user's <see cref="User.ProfileImage"/> in the database.
+        /// Any previous profile image file is deleted before saving.
+        /// </summary>
+        /// <param name="user">The user to generate and save a profile image for.</param>
+        /// <param name="displayName">
+        /// Optional display name override used to derive initials.
+        /// When <c>null</c>, the user's <see cref="User.Username"/> is used.
+        /// </param>
+        /// <returns>A <see cref="Task"/> representing the asynchronous operation.</returns>
+        Task GenerateAndSaveProfileImageAsync(User user, string? displayName = null);
+    }
+}
--- a/src/Jellyfin.Drawing.Skia/ProfileImageGenerator.cs
+++ b/src/Jellyfin.Drawing.Skia/ProfileImageGenerator.cs
@@ -15,21 +15,32 @@ namespace Jellyfin.Drawing.Skia;
 /// </summary>
 public class ProfileImageGenerator : MediaBrowser.Controller.Drawing.IProfileImageService
 {
-    // A palette of pleasant background colors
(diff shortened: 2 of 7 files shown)


nielsvanvelzen (changes requested):
I don't think we really need an API for this. I'd say if there is no uploaded image available we should always fallback to these generated images. Meaning the server will **always** return a profile picture with this change.

JPVenson:
fair enough, i thought the api might be useful if the user would like to override the generated image with its Own version (hence the explicit name parameter). Say if that still should go in your eyes.

nielsvanvelzen:
I don't think there's much worth to that. We don't provide that functionality for the generated library images either.

JPVenson:
I removed the api endpoints

src/Jellyfin.Drawing.Skia/ProfileImageGenerator.cs
```diff
@@ -0,0 +1,185 @@
...
+    public async Task GenerateAndSaveProfileImageAsync(
+        User user,
+        string? displayName)
+    {
+        _logger.LogDebug("Generating profile image for user {UserId}", user.Id);
+
+        if (user.ProfileImage is not null)
+        {
+            try
+            {
```
IDisposable, line 80: It would be better to defer the deletion of the file until after the generation succeeds.

Ideally we would just do this here

```csharp
var oldProfileImagePath = user.ProfileImage?.Path;
```

src/Jellyfin.Drawing.Skia/ProfileImageGenerator.cs
```diff
@@ -0,0 +1,185 @@
...
+        var stream = await GenerateProfileImageAsync(displayName ?? user.Username, user.Id).ConfigureAwait(false);
+        await using (stream.ConfigureAwait(false))
+        {
+            await _providerManager.SaveImage(stream, "image/png", imagePath).ConfigureAwait(false);
+        }
+
+        user.ProfileImage = new ImageInfo(imagePath);
+        await _userManager.UpdateUserAsync(user).ConfigureAwait(false);
+        _logger.LogDebug("Saved profile image for user {UserId} to {Path}", user.Id, imagePath);
+    }
```
IDisposable, line 106: ... and then do the delete here (which would be skipped if the creation/update failed/threw):

```csharp
   if (!string.IsNullOrEmpty((oldProfileImagePath))
   {
         try
         { 
             System.IO.File.Delete(oldProfileImagePath);
             _logger.LogDebug("Deleted previous profile image at {Path}", oldProfileImagePath);
         }
         catch (IOException e)
         {
             _logger.LogError(e, "Failed to delete previous profile image at {Path}", userProfileImagePath);
         }
   }
```

src/Jellyfin.Drawing.Skia/ProfileImageGenerator.cs
```diff
@@ -0,0 +1,185 @@
...
+            {
+                System.IO.File.Delete(user.ProfileImage.Path);
+                _logger.LogDebug("Deleted previous profile image at {Path}", user.ProfileImage.Path);
+            }
+            catch (IOException e)
+            {
+                _logger.LogError(e, "Failed to delete existing profile image at {Path}", user.ProfileImage.Path);
+            }
+
+            await _userManager.ClearProfileImageAsync(user).ConfigureAwait(false);
```
IDisposable, line 89: ... and then we technically no longer need to ever `ClearProfileImageAsync` because we'll just leave it alone if the generation wasn't successful. 

This seems like a happier path in case of transient access to the image storage. We would essentially not clear, fail the generate, and thus leave the image alone until we A) can read the old image again, or B) create a new one

MediaBrowser.Controller/Drawing/IProfileImageService.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// and updates the user's <see cref="User.ProfileImage"/> in the database.
+        /// Any previous profile image file is deleted before saving.
+        /// </summary>
+        /// <param name="user">The user to generate and save a profile image for.</param>
+        /// <param name="displayName">
+        /// Optional display name override used to derive initials.
+        /// When <c>null</c>, the user's <see cref="User.Username"/> is used.
+        /// </param>
+        /// <returns>A <see cref="Task"/> representing the asynchronous operation.</returns>
+        Task GenerateAndSaveProfileImageAsync(User user, string? displayName = null);
```
IDisposable, line 33: All calls to this method never pass the `displayName`, so the entire _if not `null`_ path is never used. Can we simplify or explain when we're anticipating when we will pass something?

jumoog:
Good idea!

JPVenson:
Closed as apparently not acceptable by server.

Replaced by: https://github.com/JPVenson/Jellyfin.BetterDefaultProfilePictures.Plugin

nielsvanvelzen:
Sad, was looking forward to this one 😭

PR #16739, opened 2026-05-01
fix: use symmetric 15s default for skip forward/backward lengths

tcsenpai:
**Problem**
The server-side default for SkipForwardLength is 30 000 ms (30 s) while SkipBackwardLength defaults to 10 000 ms (10 s). This asymmetry is hardcoded in DisplayPreferencesController.cs and propagates to all clients (Android, Android TV, web) that read these user settings from the server.

**Solution**
Set both defaults to 15 000 ms (15 s). This gives a predictable, symmetric seek experience out of the box while still allowing users to customize the values per-client in their profile settings.

**Files changed**
Jellyfin.Api/Controllers/DisplayPreferencesController.cs — 2 lines

### DIFF
--- a/Jellyfin.Api/Controllers/DisplayPreferencesController.cs
+++ b/Jellyfin.Api/Controllers/DisplayPreferencesController.cs
@@ -157,13 +157,13 @@ public class DisplayPreferencesController : BaseJellyfinApiController
         existingDisplayPreferences.SkipBackwardLength = displayPreferences.CustomPrefs.TryGetValue("skipBackLength", out var skipBackLength)
                                                         && !string.IsNullOrEmpty(skipBackLength)
             ? int.Parse(skipBackLength, CultureInfo.InvariantCulture)
-            : 10000;
+            : 15000;
         displayPreferences.CustomPrefs.Remove("skipBackLength");
 
         existingDisplayPreferences.SkipForwardLength = displayPreferences.CustomPrefs.TryGetValue("skipForwardLength", out var skipForwardLength)
                                                        && !string.IsNullOrEmpty(skipForwardLength)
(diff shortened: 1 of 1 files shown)


nielsvanvelzen:
>  and propagates to all clients (Android, Android TV, web) that read these user settings from the server.

I don't think that's entirely true, the Android TV apps still reads it from the custom prefs part and has the (same) defaults [hardcoded as fallback](https://github.com/jellyfin/jellyfin-androidtv/blob/46624419f6120e81342a5e1e0243b484621b644a/app/src/main/java/org/jellyfin/androidtv/preference/UserSettingPreferences.kt#L17). So if this is merged we should make sure to _at least_ update the ATV app and perhaps others.

tcsenpai:
That's a good catch, honestly I just assumed android TV used the same approach.

What would be detrimental in approving it and then changing the android TV app later?

Asking out of curiosity

Sent from [Proton Mail](https://proton.me/mail/home) for Android.

-------- Original Message --------

nielsvanvelzen:
Fwiw these changes are good to go for the server part. Ideally we'd change the Android TV app to use the newer SkipBackwardLength and SkipForwardLength properties from the API but until that's happened we can just bump the default values there at the same time.

nielsvanvelzen (approved):

tcsenpai:
Thanks!

I am available for further clarifications

Sent from [Proton Mail](https://proton.me/mail/home) for Android.

-------- Original Message --------

Shadowghost (approved):

JPVenson (approved):
as a new default feels fine to me

Bond-009 (approved):

Issue #881, opened 2019-02-13
Certain data locations are not using `datadir` properly


joshuaboniface:
**Describe the bug**
Some elements of the media library (cache, or download data perhaps - not sure) continue to try to use the default system location (in Debian, `/var/lib/jellyfin`) even when the `--datadir` option is passed with a different value.

Did some digging and was able to confirm that this issue has two completely different symptoms:

Most of the time, rescanning a library just fails in the browser since the library `itemId` can't be found - there are no server-side errors but the API client must be trying to grab this from a folder that it expects at `/var/lib/jellyfin`, but is not finding (despite the `datadir` being elsewhere).

The second symptom occurrs if a library starts to scan. If the original location does not exist, this results in an occasional error stack trace indicating that ` /var/lib/jellyfin/root/default/<library>` can't be found.

All of these issues disappear if the `datadir` is restored to the default.

**To Reproduce**
1. Move the default data directory to a new location (do not recreate the original data location):
    ```
    $ sudo systemctl stop jellyfin
    $ sudo mv /var/lib/jellyfin /srv/jellyfin
    ```
1. Set `--datadir` on the command line (`/etc/default/jellyfin`) to the new location:
    ```
    $ grep 'DATA_DIRECTORY' /etc/default/jellyfin
    JELLYFIN_DATA_DIRECTORY="/srv/jellyfin"
    ```
1. Start Jellyfin and rescan a library.

**Expected behavior**
No errors, all data is written to `/srv/jellyfin`, all lookups are from `/srv/jellyfin`.

**Logs**

Failure to scan libraries, in browser:
```
Error: null itemId
apiclient.js:739:28
ApiClient.prototype.refreshItem
http://test5:8096/web/bower_components/emby-apiclient/apiclient.js:739:28
onSubmit/<
http://test5:8096/web/bower_components/emby-webcomponents/refreshdialog/refreshdialog.js:79:13
forEach self-hosted:261:13 onSubmit
http://test5:8096/web/bower_components/emby-webcomponents/refreshdialog/refreshdialog.js:78:9
onSubmit self-hosted:977:17 
```

Occasional error when scanning actually results:
```
Feb 11 21:28:02 jf1 jellyfin[515]: [21:28:02] [ERR] Error refreshing owned items for /var/lib/jellyfin/root/default/Movies 
[...]

PR #16745, opened 2026-05-02
Fix VttWriter to convert ASS alignment tags to WebVTT cue position settings

### LINKED ISSUE #16736: WebVTT cue positioning (line:X%) converted to SSA {\anX} literal text by server subtitle parser
### Description of the bug

When the server parses an external WebVTT subtitle file containing cue positioning settings (e.g. `line:10% align:center`), the resulting subtitles contain `{\anX}` as visible literal text prepended to each cue. The positioning is also ignored and subtitles always appear at the bottom.

The source VTT file contains no SSA/ASS tags whatsoever (attached). Server logs confirm the file is being parsed by the WebVTT format parser before being served to the client.

[Big Buck Bunny.vtt.txt](https://github.com/user-attachments/files/27271967/Big.Buck.Bunny.vtt.txt)

### Reproduction steps

1. Add a video with an external .vtt subtitle file containing cue positioning (e.g. `line:10% align:center`)
2. Play the video on any client (tested on Jellyfin Android and Jellyfin Android TV with internal player)
3. Enable the external VTT subtitle track

{\an8} appears as visible literal text at the start of every subtitle cue. Vertical positioning is ignored; subtitles always render at the bottom.

Subtitles should render at the position specified by the VTT cue settings, with no literal tags visible.

### Jellyfin Server version

10.11.8

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.8

### Environment

```markdown
OS: Debian
Virtualization: none
Clients: Android (internal ExoPlayer-based player), Android TV (Chromecast with Google TV, internal ExoPlayer-based player)
FFmpeg Version: Jellyfin FFmpeg
Playback: Direct Play
Hardware Acceleration: QSV
Reverse Proxy: yes
Networking: local
Media Storage & Filesystem: Local HDD
```

### Jellyfin logs

```shell
[DBG] charset "ascii" detected for ".../Big Buck Bunny.vtt"
[DBG] Trying to parse '"vtt"' subtitle using the "WebVTT" format parser
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

Image

### Additional information

_No response_

ferezvi:
VttWriter was hardcoding `line:90%` for all cues and writing ASS alignment tags (e.g. `{\an8}`) as literal text in the output. These tags are introduced by libse when parsing VTT files with position cues, but were never converted 
back to valid VTT cue settings when writing the output.

This fix preserves VTT cue settings through the subtitle pipeline without parsing any subtitle format tags in the writer:

- Adds `VttCueSettings` to `SubtitleTrackEvent` to carry cue settings 
  through the pipeline
- `SubtitleEditParser` maps `p.Style` from libse to the new field — libse 
  already populates this with the raw cue settings string (e.g. 
  `"line:10% align:center"`) when parsing VTT files
- `VttWriter` writes `VttCueSettings` directly to the timing line with no 
  additional parsing, falling back to `line:90%,end` when not present
- `{\anN}` tags introduced by libse are stripped from the text only when 
  `VttCueSettings` is present, leaving SRT and other formats untouched

Tested with VTT files containing line/position cues on Jellyfin Android TV and Android app (integrated player) . Before: tags visible as text, all subtitles at bottom.
After: subtitles appear at correct position, no tag artifacts.

### DIFF
--- a/MediaBrowser.MediaEncoding/Subtitles/VttWriter.cs
+++ b/MediaBrowser.MediaEncoding/Subtitles/VttWriter.cs
@@ -36,8 +36,11 @@ namespace MediaBrowser.MediaEncoding.Subtitles
             foreach (var (tag, position) in _assTagToCuePosition)
             {
                 if (text.StartsWith(tag, StringComparison.Ordinal))
+                {
                     return position;
+                }
             }
+
             return "region:subtitle line:90%";
         }
 
@@ -60,6 +63,7 @@ namespace MediaBrowser.MediaEncoding.Subtitles
                     {
                         endTime = startTime.Add(TimeSpan.FromMilliseconds(1));
                     }
+
                     var text = trackEvent.Text;
                     // TODO: Not sure how to handle these
                     text = NewlineEscapeRegex().Replace(text, " ");
--- a/MediaBrowser.MediaEncoding/Subtitles/VttWriter.cs
+++ b/MediaBrowser.MediaEncoding/Subtitles/VttWriter.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Collections.Generic;
 using System.IO;
 using System.Text;
 using System.Text.RegularExpressions;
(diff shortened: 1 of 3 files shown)


MediaBrowser.MediaEncoding/Subtitles/VttWriter.cs
```diff
@@ -15,6 +16,32 @@ public partial class VttWriter : ISubtitleWriter
         [GeneratedRegex(@"\\n", RegexOptions.IgnoreCase)]
```
IDisposable: Unrelated to your change, but why would we have a case-insensitive regex for a newline character? It's 0x0A all day, every day.

MediaBrowser.MediaEncoding/Subtitles/VttWriter.cs
```diff
@@ -15,6 +16,32 @@ public partial class VttWriter : ISubtitleWriter
...
         private static partial Regex NewlineEscapeRegex();
 
+        [GeneratedRegex(@"^\{\\an\d\}", RegexOptions.IgnoreCase)]
```
IDisposable: Just a question here... are tags _really_ case-insensitive? I would be _very_ surprised to see `\AN9` is supposed to be a  valid tag. If they are not, we shouldn't be using `RegexOptions.IgnoreCase` here.

MediaBrowser.MediaEncoding/Subtitles/VttWriter.cs
```diff
@@ -15,6 +16,32 @@ public partial class VttWriter : ISubtitleWriter
...
+        private static partial Regex AssAlignTagRegex();
+
+        private static readonly Dictionary<string, string> _assTagToCuePosition = new()
+        {
+            ["{\\an1}"] = "position:20%",
```
IDisposable: This works for `\an?` form (e.g. the NumPad layout), but not for the `\a?` where the number is a 

From the [ASS file format specs](http://www.tcax.org/docs/ass-specs.htm)

```text
\a<alignment>            <alignment> is a number specifying the onscreen alignment/positioning of a subtitle.

                                 A value of 1 specifies a left-justified subtitle
                                 A value of 2 specifies a centered subtitle
                                 A value of 3 specifies a right-justified subtitle

                                 Adding 4 to the value specifies a "Toptitle"
                                 Adding 8 to the value specifies a "Midtitle"

                                 0 or nothing resets to the style default (which is usually 2)


                                 eg. {\a1}This is a left-justified subtitle
                                       {\a2}This is a centered subtitle
                                       {\a3}This is a right-justified subtitle
                                       {\a5}This is a left-justified toptitle
                                       {\a11}This is a right-justified midtitle

\an<alignment>         numpad layout
```

From [ASS override tags doc](https://aegisub.org/docs/latest/ass_tags/)

```text
Line alignment

\an<pos>
Specify the alignment of the line. The alignment specifies the position of the line when no [position override](https://aegisub.org/docs/latest/ass_tags/#\pos) or [movement](https://aegisub.org/docs/latest/ass_tags/#\move) is in effect, and otherwise specifies the anchor point of positioning and rotation.

The \an tag uses “numpad” values for the pos, ie. the alignment values correspond to the positions of the digits on the numeric keypad on a regular keyboard:

Bottom left
Bottom center
Bottom right
Middle left
Middle center
Middle right
Top left
Top center
Top right

Line alignment (legacy)

\a<pos>
Specify the alignment of the line using legacy alignment codes from SubStation Alpha. This tag is supported but considered deprecated; you should usually use \an in new scripts instead, as it is more intuitive.
The exception is that \a6 should be used for lazy sign translating, because if you’re going to be lazy you should do it right and save the extra character.

Calculate pos as follows: Use 1 for left-alignment, 2 for center alignment and 3 for right-alignment. If you want sub-titles you’re done. To get top-titles, add 4 to the number, to get mid-titles add 8 to the number:

1: Bottom left
2: Bottom center
3: Bottom right
5: Top left
6: Top center
7: Top right
9: Middle left
10: Middle center
11: Middle right
```

gnattu (approved):

MediaBrowser.MediaEncoding/Subtitles/VttWriter.cs
ferezvi: This predates my change. I'd rather keep this PR focused but i'll be happy to open a separate one for that cleanup.

MediaBrowser.MediaEncoding/Subtitles/VttWriter.cs
ferezvi: You're right, ASS override tags are case-sensitive. Removed IgnoreCase.

MediaBrowser.MediaEncoding/Subtitles/VttWriter.cs
ferezvi: You're right that \a legacy tags are not handled here. However, the only input that reaches VttWriter in practice is VTT, which has no \a equivalent — Jellyfin's parser only generates {\anX} internally. I'm not aware of a code path where \a tags would arrive here, but if there is one (e.g. via ASS direct input), I'd be happy to add support in a follow-up PR once we confirm it's needed.

IDisposable (approved):

Bond-009:
@ferezvi Can you please test if #16805 solves your use-case?

ferezvi:
> @ferezvi Can you please test if #16805 solves your use-case?

Thanks for the pointer.
My use-case: when Jellyfin serves a valid WebVTT file, SubtitleEdit normalizes the cue position settings (line:x%, position:x%) into {\anN} tags prepended to the cue text. This normalization is desirable — it preserves positioning information in a unified internal representation, giving users a consistent experience.
By the time the data reaches VttWriter, the original positioning only exists as those {\anN} tags. Without PR #16745, they render as literal text on screen. PR #16745 strips them and maps them back to valid WebVTT cue position settings, completing the round-trip correctly.
We reviewed #16805 (my wife and I, WAF is important) and it routes VTT conversion through SubtitleEdit.WebVTT.ToText() directly, bypassing VttWriter entirely. Would it make sense to route through VttWriter instead? I'm happy to help explore that approach if it's aligned with the project's direction.
If you have any questions I'd be more than happy to help or contribute in any way I can.

gabeluci:
> when Jellyfin serves a valid WebVTT file, SubtitleEdit normalizes the cue position settings (line:x%, position:x%) into {\anN} tags prepended to the cue text.

I did some testing in Jellyfin server with letting SubtitleEdit convert subtitles between formats, much the same as #16805 - but since it broke the Roku client with subtitles going off the bottom, I decided to leave it alone. I tested that locally with an external ASS file with a bunch of {\an#} tags. Roku always requests it in VTT, and SubtitleEdit converted the ASS position tags into VTT position data. None of the ASS tags were left in the text.

ferezvi:
> gabeluci

Are you using custom subtitles? Internally, custom subtitles strip VTT position tags before delivery. For Roku specifically, the server sends VTT subtitles as-is without any modification — any other format goes through vttwritter first. So VTT it's not actually broken see PR #881.

Could you run a test on an Android device using **Settings → Client Settings → Integrated Player**, or on an Android TV using the default player? In my tests, position tags are visible there.

gabeluci:
> when Jellyfin serves a valid WebVTT file, SubtitleEdit normalizes the cue position settings (line:x%, position:x%) into {\anN} tags prepended to the cue text.

I did some testing, and I do see that when I'm testing with 3.1.9, but not with the current master branch, so something got changed. But what I'm seeing now is the position data being stripped entirely. I see why, it's reporting it cannot support "webvtt" when "vtt" is in the list, which is dumb, so it's converting to SRT first with ffmpeg (as it does with any unsupported type), which loses the position data.

I might submit a pull request for that.

Ideally, my vote (whatever that's worth) is to let SubtitleEdit do all the conversion since it'll do a better job of it than we ever could. I know that would make your PR moot, but I think it would end up with cleaner code in the end.

ferezvi:
@gabeluci That makes sense, and it actually lines up pretty closely with what I’ve been finding while working on this area.

I’ve been building on top of some of the work @Bond-009 already did and experimenting with an extension-based approach for subtitle JSON handling here:
[json-subtitle-helper branch](https://github.com/ferezvi/jellyfin/tree/json-subtitle-helper?utm_source=chatgpt.com)

I also ran into the ffmpeg fallback/conversion issue you mentioned and implemented a possible fix for it in my branch, since the positioning data seems to get lost before the later processing stages.

Feel free to take a look at the branch if you want, and feel completely free to ask me anything about it. I’d be very happy to help or collaborate however useful.

IDisposable (approved):

ferezvi:
Hi

I tested PR #16805 and found several issues:

- webvtt vs vtt format name mismatch causes unnecessary conversions or fallback to ffmpeg, losing position data before it even reaches the writer
- SRT→VTT conversion via SubtitleEdit loses {\anX} positioning entirely — subtitles appear at the bottom with no position data. This is a regression compared to the current VttWriter behavior
- Internal subtitles served as a separate stream also lose positioning completely
- Some output contains invalid VTT tags that Roku and ExoPlayer reject, causing subtitles to not render at all

On the dictionary side, through testing on Android, Android TV and Roku I found that ,center and ,start line alignment suffixes, while valid VTT, cause issues on some players. Only ,end on line:90% is needed for ExoPlayer (Android/Android TV) to display subtitles correctly and not stick them to the bottom. The other suffixes can be safely dropped.
I'd like to add {\an2} explicitly to the dictionary and clean up the alignment suffixes as a small improvement if that helps move this forward.
As a future improvement, applying similar positioning logic to the JSON subtitle format could benefit clients that use it, though this is not currently needed.
Could you advise whether it makes sense to make these small improvements to this PR, or if the project is moving in a different direction? I want to make sure my contributions are aligned with where things are headed.
I'm happy to contribute further in whatever direction makes sense for the project.

MediaBrowser.MediaEncoding/Subtitles/VttWriter.cs
```diff
@@ -12,9 +13,38 @@ namespace MediaBrowser.MediaEncoding.Subtitles
...
+            ["{\\an3}"] = "position:80% line:90%",
+            ["{\\an4}"] = "position:20% line:50%",
+            ["{\\an5}"] = "line:50%",
+            ["{\\an6}"] = "position:80% line:50%",
+            ["{\\an7}"] = "position:20% line:10%",
+            ["{\\an8}"] = "line:10%",
+            ["{\\an9}"] = "position:80% line:10%",
+        };
+
+        [GeneratedRegex(@"\\n")]
```
gnattu: I think we still need RegexOptions.IgnoreCase as `\N` also breaks in ASS,

MediaBrowser.MediaEncoding/Subtitles/VttWriter.cs
```diff
@@ -12,9 +13,38 @@ namespace MediaBrowser.MediaEncoding.Subtitles
...
         private static partial Regex NewlineEscapeRegex();
 
+        [GeneratedRegex(@"^\{\\an\d\}")]
+        private static partial Regex AssAlignTagRegex();
+
+        private static string GetCuePositionFromAssTag(string text)
+        {
+            foreach (var (tag, position) in _assTagToCuePosition)
+            {
+                if (text.StartsWith(tag, StringComparison.Ordinal))
```
gnattu: There are some more complicated ASS composition that might not have `{\anX}` as the exact beginning but I am not sure if such cases are in scope of this PR. For example:
```
{\an4\i1}Left-middle italic text
```

Which should render as:

Image

And it is currently converted to

```
00:00:10.000 --> 00:00:14.000 region:subtitle line:90%,end
{\an4\i1}Left-middle italic text
```

I am personally fine if this PR does not cover those complex cases, as it would push the complexity way beyond its current status, given the complexity of ASS styling itself.

MediaBrowser.MediaEncoding/Subtitles/VttWriter.cs
ferezvi: Thanks for the suggestion! I updated the regex to \\[nN] which explicitly matches both \n and \N via a character class, so IgnoreCase isn't needed.

MediaBrowser.MediaEncoding/Subtitles/VttWriter.cs
ferezvi: Great point! I went ahead and updated the PR to handle those cases. {\an4\i1} is now parsed by ProcessAssBlocks, which extracts the alignment tag and converts \i1 to <i>, producing:
```
00:00:10.000 --> 00:00:14.000 position:20% line:50%
<i>Left-middle italic text</i>
```
Tested on Roku and Android.

nielsvanvelzen (changes requested):
Parsing of subtitle formats shouldn't happen in the writers, because with this current implementation all subtitle formats (like srt) are suddenly supporting ass tags, which we explicitly do NOT support.

ferezvi:
Thanks for the feedback — you were right. Parsing subtitle format tags in the writer was the wrong approach.

This revision removes all ASS tag parsing from VttWriter entirely. Instead, we leverage the fact that libse already does the right thing: when parsing a VTT file, it stores the raw cue settings string (e.g. "line:10% align:center") in `p.Style`. We now carry that value through the pipeline via a new `VttCueSettings` field on `SubtitleTrackEvent`, and VttWriter simply writes it out — no parsing, no dictionaries.

The `{\anN}` tags libse introduces into `p.Text` are stripped only when `VttCueSettings` is present, so SRT files with accidental ASS tags are not affected.

theguymadmax:
Niels feedback has been addressed, so that is no longer a blocking factor.  However, if @Bond-009 's PR (https://github.com/jellyfin/jellyfin/pull/16805) can be made mergeable in time for 12.0, it would supersede this. Otherwise, this could continue to be reviewed/merged for 12.0

ferezvi:
Thanks for the discussion and work on this. I'm closing this in favor of #16805, which appears to be the better long-term solution. Further feedback and testing should be focused there.

PR #16755, opened 2026-05-03
Load Jellyfin.MediaEncoding.Keyframes eagerly in GetComposablePartAssemblies

jumoog:
PluginLoadContext resolves assemblies only from the plugin's own directory via AssemblyDependencyResolver, falling back to AssemblyLoadContext.Default when an assembly is not found there. Jellyfin.MediaEncoding.Keyframes was only a transitive dependency of Jellyfin.MediaEncoding.Hls and was loaded lazily, so it was not guaranteed to be present in the default context when a plugin first needed it, causing a FileNotFoundException at runtime.

Adding it explicitly to GetComposablePartAssemblies ensures it is loaded into AssemblyLoadContext.Default at server startup, making it reliably available to plugins through the standard fallback path.


jumoog:
doesnt work for plugins

PR #16756, opened 2026-05-04
Speed-up LatestItems for Music

Shadowghost:
Due to some N+1 fetching issues and Ancestor resolution problems, loading latest added music was about 4x slower than other LatestItems queries. This should fix it by using proper EXISTS clauses and artist batching for the lookups.

Other approaches I tested for the grouping issue:
 * GROUP BY Album-string with a correlated subquery to pick a representative track - one subquery per album group, ~3s for a few hundred albums
* GROUP BY track.ParentId - fast, but wrong for multi-disc albums where the immediate parent is a Disc folder, not the album

### DIFF
--- a/Emby.Server.Implementations/Dto/DtoService.cs
+++ b/Emby.Server.Implementations/Dto/DtoService.cs
@@ -203,6 +203,39 @@ namespace Emby.Server.Implementations.Dto
                 }
             }
 
+            // Batch-fetch MusicArtist lookups across all items to avoid N+1 queries.
+            IReadOnlyDictionary<string, MusicArtist[]>? artistsBatch = null;
+            HashSet<string>? artistNames = null;
+            foreach (var item in accessibleItems)
+            {
+                if (item is IHasArtist hasArtist)
+                {
+                    foreach (var name in hasArtist.Artists)
+                    {
+                        if (!string.IsNullOrWhiteSpace(name))
+                        {
+                            (artistNames ??= new HashSet<string>(StringComparer.Ordinal)).Add(name);
+                        }
+                    }
+                }
+
+                if (item is IHasAlbumArtist hasAlbumArtist)
+                {
+                    foreach (var name in hasAlbumArtist.AlbumArtists)
+                    {
+                        if (!string.IsNullOrWhiteSpace(name))
+                        {
+                            (artistNames ??= new HashSet<string>(StringComparer.Ordinal)).Add(name);
+                        }
+                    }
+                }
+            }
+
+            if (artistNames is { Count: > 0 })
+            {
+                artistsBatch = _libraryManager.GetArtists(artistNames.ToArray());
+            }
+
             for (int index = 0; index < accessibleItems.Count; index++)
             {
                 var item = accessibleItems[index];
@@ -214,7 +247,8 @@ namespace Emby.Server.Implementations.Dto
                     userDataBatch?.GetValueOrDefault(item.Id),
                     allCollectionFolders,
                     childCountBatch,
-                    playedCountBatch);
+                    playedCountBatch,
+                    artistsBatch);
 
                 if (item is LiveTvChannel tvChannel)
                 {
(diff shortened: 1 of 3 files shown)


Jellyfin.Server.Implementations/Item/BaseItemRepository.Querying.cs
```diff
@@ -125,53 +124,69 @@ public IReadOnlyList<BaseItemDto> GetLatestItemList(InternalItemsQuery filter, C
...
+                    MaxDate = g.Max(e => e.DateCreated),
+                    FirstId = g.OrderByDescending(e => e.DateCreated).ThenByDescending(e => e.Id).Select(e => e.Id).First()
+                })
+                .OrderByDescending(g => g.MaxDate);
+
+            var idsQuery = filter.Limit.HasValue
+                ? topGroupItems.Take(filter.Limit.Value).Select(g => g.FirstId)
+                : topGroupItems.Select(g => g.FirstId);
+
+            firstIds = idsQuery.ToList();
```
JPVenson: Not really happy with this as N can easily be bigger then whats possible to translate into a query

JPVenson (changes requested):

Jellyfin.Server.Implementations/Item/BaseItemRepository.Querying.cs
Shadowghost: True. I hate that this keep coming up...
Will take a look

Emby.Server.Implementations/Dto/DtoService.cs
```diff
@@ -203,6 +203,39 @@ public IReadOnlyList<BaseItemDto> GetBaseItemDtos(
...
+            HashSet<string>? artistNames = null;
+            foreach (var item in accessibleItems)
+            {
+                if (item is IHasArtist hasArtist)
+                {
+                    foreach (var name in hasArtist.Artists)
+                    {
+                        if (!string.IsNullOrWhiteSpace(name))
+                        {
+                            (artistNames ??= new HashSet<string>(StringComparer.Ordinal)).Add(name);
```
IDisposable: I'm all for lazy evaluations, but just creating the empty hashset once before all these loops is going to be cheaper than a null-check for every single artist name. That's so many branch predictions that could go astray.

Jellyfin.Server.Implementations/Item/BaseItemRepository.Querying.cs
```diff
@@ -125,45 +124,54 @@ public IReadOnlyList<BaseItemDto> GetLatestItemList(InternalItemsQuery filter, C
...
-            .Where(groupKeyFilter)
-            .GroupBy(groupKeySelector)
-            .Select(g => new
-            {
-                MaxDate = g.Max(e => e.DateCreated),
-                FirstId = g.OrderByDescending(e => e.DateCreated).ThenByDescending(e => e.Id).Select(e => e.Id).First()
-            })
-            .OrderByDescending(g => g.MaxDate);
+        // Albums whose Id is the parent of any track matching the user's filter.
+        var musicAlbumType = _itemTypeLookup.BaseItemKindNames[BaseItemKind.MusicAlbum]!;
```
IDisposable: This seems like something we could lookup once and cache

IDisposable (approved):

Jellyfin.Server.Implementations/Item/BaseItemRepository.Querying.cs
IDisposable: ... in the old SQL Server days™, we passed stored procs a [table variable](https://learn.microsoft.com/en-us/sql/t-sql/data-types/table-transact-sql?view=sql-server-ver17) filled with the ids and JOINed/EXISTS against that. I'm not sure what SQLite offers in this domain beyond inlined CTEs or temp tables.  Just might be a path worth looking into?

JPVenson: sadly table valued functions are not a thing in sqlite and CTEs are not _yet_ supported in efcore. That would have been a godsend in soooo many places if CTEs were supported

IDisposable: Any thoughts on using https://github.com/borisdj/EFCore.BulkExtensions (which is license compatible)

JPVenson: that tool has been discussed extensively and is scheduled for a later discussion for its inclusion.

JPVenson (approved):

IDisposable (approved):

PR #16758, opened 2026-05-04
Switch to new version scheme

Shadowghost:
**Changes**
Switch to new version scheme, dropping the 10.x prefix


joshuaboniface (approved):

nielsvanvelzen (commented):
Is just 12.0 going to work or do we need to keep it 12.0.0 in the assemblies?

We might need to update the packaging scripts for this change as well, and jellyfin-web of course.

Bond-009 (approved):

thornbill:
Corresponding web PR: https://github.com/jellyfin/jellyfin-web/pull/7885

crobibero (approved):

PR #16769, opened 2026-05-04
Guard against null-overwrite of saved audio/subtitle track selections

llaforest:
Some clients omit AudioStreamIndex or SubtitleStreamIndex in playback progress reports and it causes previously saved track selections to be erased. 

Add .HasValue checks so only explicit track changes are persisted.

### DIFF
--- a/Emby.Server.Implementations/Session/SessionManager.cs
+++ b/Emby.Server.Implementations/Session/SessionManager.cs
@@ -973,7 +973,7 @@ namespace Emby.Server.Implementations.Session
 
             if (user.RememberAudioSelections)
             {
-                if (data.AudioStreamIndex != info.AudioStreamIndex)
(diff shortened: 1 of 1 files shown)


nielsvanvelzen:
Isn't this because a `null` value is to indicate no track selection? 😄

llaforest:
It seems some client simply do not sent it. 

It is not possible to not choose any audio right?
For subtitles, if you disable them it would be -1. So not null.

So when null, it is assumed no changes are desired in that regard.

JPVenson (approved):

JPVenson:
Code change ok, not validating purpose tho

nielsvanvelzen:
Yeah I just checked with the Android TV code and we do in fact send `-1` when the subtitle track is disabled.

Shadowghost (approved):

PR #16786, opened 2026-05-06
Fix fMP4 HLS subtitle timestamp map

### LINKED ISSUE #16647: HLS subtitle X-TIMESTAMP-MAP is misaligned when using fMP4 segments (SegmentContainer=mp4)

- [x] This issue is **not** already reported on GitHub.
- [x] I'm using an up to date version of Jellyfin Server.

### Description of the bug

When requesting HLS with `SegmentContainer=mp4` and `SubtitleMethod=Hls`, the WebVTT subtitle segments contain `X-TIMESTAMP-MAP=MPEGTS:900000,LOCAL:00:00:00.000`, which maps PTS 10s to subtitle time 0. This is correct for TS segments (which start at PTS ~10s), but fMP4 segments start at PTS ~0s. The result is that all HLS-delivered subtitles are displayed approximately 10 seconds late.

This affects all text-based subtitles regardless of source — embedded in MKV, external SRT files, ASS, etc. Jellyfin extracts and converts them all to WebVTT for HLS delivery, and the `X-TIMESTAMP-MAP` is the same in every case. Verified against a video with embedded SRT inside an MKV container.

The `X-TIMESTAMP-MAP` appears to be generated assuming TS PTS conventions regardless of the `SegmentContainer` parameter.

Related discussion: https://github.com/jellyfin/jellyfin/issues/10247

### Reproduction steps

1. Set up a Jellyfin server with content that has text subtitles (SRT, ASS, etc.)
2. Request HLS playback with fMP4 segments and HLS subtitles:
   ```
   /Videos/{id}/master.m3u8?mediaSourceId={id}&VideoCodec=h264,hevc,h265&AudioCodec=aac&SubtitleMethod=Hls&SegmentContainer=mp4&MaxStreamingBitrate=200000000
   ```
3. Play on any Apple platform using AVPlayer (tvOS, iOS, macOS Safari)
4. Enable subtitles via the native player subtitle picker
5. Observe that subtitles appear ~10 seconds after the corresponding dialogue

Alternatively, verify by inspecting the subtitle playlist directly:
```bash
# Fetch a subtitle segment
curl "http://<server>:8096/Videos/{id}/{sourceId}/Subtitles/{index}/stream.vtt?CopyTimestamps=true&AddVttTimeMap=true&StartPositionTicks=900000000&EndPositionTicks=1200000000&ApiKey={token}"

# Note the header: X-TIMESTAMP-MAP=MPEGTS:900000,LOCAL:00:00:00.000
# 900000 / 90000 = 10 seconds PTS offset
# But fMP4 video segments start at PTS ~0, not ~10
```

Subtitles delivered via `SubtitleMethod=Hls` are displayed ~10 seconds late when `SegmentContainer=mp4` is used. The `X-TIMESTAMP-MAP=MPEGTS:900000,LOCAL:00:00:00.000` assumes video PTS starts at ~10s (the TS convention), but fMP4 segments start at PTS ~0s.

When `SegmentContainer=mp4`, the `X-TIMESTAMP-MAP` should reflect the actual fMP4 PTS offset. Since fMP4 segments start at PTS ~0, the mapping should be `MPEGTS:0,LOCAL:00:00:00.000` (or adjusted to match the actual fMP4 PTS).

### Jellyfin Server version

10.11.8

### Specify the build version

10.11.8

### Environment

- OS: Docker
- Virtualization: Docker
- Clients: tvOS 26.2 (Apple TV 4K 3rd gen)
- FFmpeg Version: jellyfin-ffmpeg (bundled with 10.11.8 Docker image)
- Playback Method: HLS direct stream (remux, no transcode for h264/hevc)
- Hardware Acceleration: nvenc
- Networking: Local network

### Jellyfin logs

Not applicable — the issue is in the generated WebVTT subtitle segments, not in server errors.

### Additional information

The workaround is to fetch subtitles separately via the `/Subtitles/{index}/Stream.vtt` REST endpoint (which returns correct absolute timestamps) and render them client-side, bypassing HLS subtitle delivery entirely.

InkAurora:
**Changes**
Fix HLS WebVTT timestamp map generation so fMP4 subtitle playlists use a zero MPEGTS offset while MPEG-TS playlists keep the existing 10 second offset. Pass the HLS segment container through subtitle playlist URLs so subtitle segment requests can map cues against the correct media timeline. Add API test coverage for timestamp map selection.

**Issues**
Fixes #16647

**Testing**
- `dotnet test tests\Jellyfin.Api.Tests\Jellyfin.Api.Tests.csproj --no-restore --verbosity minimal`
- `dotnet test Jellyfin.sln --configuration Release --collect:"XPlat Code Coverage" --settings tests/coverletArgs.runsettings --verbosity minimal`


gnattu:
Isn't `segmentContainer` way too leaky as an API parameter?

If we are going to change API to handle this case, why not just make the offset exposed as an API parameter?

PR #16787, opened 2026-05-06
Allow converting to/from ASS/SSA

gabeluci:
**Changes**
Removed the `SupportsSubtitleConversionTo` method since all it did was say that we won't convert ASS/SSA subtitles to/from anything, when conversion does actually work. This can trigger transcoding when it's not necessary.

I had started a discussion about this: https://github.com/jellyfin/jellyfin-meta/discussions/126


gnattu:
We already have a PR about webvtt conversion: https://github.com/jellyfin/jellyfin/pull/16745

Please remove webvtt part is your intention is to remove a method that is likely not useful

gabeluci:
> We already have a PR about webvtt conversion: #16745

Ha.... I did actually see that PR earlier and forgot and got carried away 😄  I'll remove my changes to VttWriter

ferezvi:
Hi there, worth noting: both VTT and SRT subtitles go through VttWriter, so if #16745 gets merged, position tags from ASS/SSA will be respected after conversion. Similarly, there's work in progress on a PR to support position rendering in web browsers (Chrome for now). Once those are in, the natural next step would be implementing italic, bold, and other inline tags.

gabeluci:
> Hi there, worth noting: both VTT and SRT subtitles go through VttWriter, so if #16745 gets merged, position tags from ASS/SSA will be respected after conversion. Similarly, there's work in progress on a PR to support position rendering in web browsers (Chrome for now). Once those are in, the natural next step would be implementing italic, bold, and other inline tags.

Awesome, good to hear! I recently implemented italic, bold, and font color support in the Roku client.

FYI: I removed the change to `VttWriter` in this PR.

ferezvi:
> Awesome, good to hear! I recently implemented italic, bold, and font color support in the Roku client.

I'm a fan of your work! I actually made a PR to make subtitle borders visible on the Roku client — would love it if you could take a look at it!

gabeluci:
> I'm a fan of your work! I actually made a PR to make subtitle borders visible on the Roku client — would love it if you could take a look at it!

Oh that was you! I did see it and commented on it. It's a great idea.

MediaBrowser.Model/Entities/MediaStream.cs
```diff
@@ -769,40 +769,6 @@ public static bool IsVobSubFormat(string format)
                    || codec.Contains("vobsub", StringComparison.OrdinalIgnoreCase);
         }
 
-        public bool SupportsSubtitleConversionTo(string toCodec)
-        {
-            if (!IsTextSubtitleStream)
-            {
-                return false;
```
gnattu, line 776: Have you checked that the image based subtitles like vobsub would still be blocked to convert and not creating an error?

gabeluci: Sorry it took me so long to look at this.
The two places where `SupportsSubtitleConversionTo` was being called already includes a check for `IsTextSubtitleStream`, so that logic doesn't change. (see the two changes in `StreamBuilder.cs` in this PR)

PR #12579, opened 2024-09-04
Add OriginalLanguage as option to PreferredAudioLanguage

Lampan-git:
**Feature**
https://features.jellyfin.org/posts/454/option-to-play-original-audio-language

**Changes**
- Added originallanguage metadata tag and OriginalLanguage to items
- Added original indicator for audio streams.
- If OriginalLanguage is the picked option for PrefferedAudioLanguage:
  - The OriginalLanguage metadata is used to get the first matching language. If there is no matching, default stream will be used.
  - If there is no OriginalLanguage metadata, original indicator is used.
  - If both exists and the original-indicator-stream's language matches the OriginalLanguage metadata, that one will be used.
  - If both exists and they are not matching, the first matching OriginalLanguage will be used.
  - See the test, SetDefaultAudioStreamIndex_Index_Correct, for how the complete logic works.
 - If the item has no OriginalLanguage metadata it will try to get the metadata from its' parents

**Web**
jellyfin-web PR to enable this: https://github.com/jellyfin/jellyfin-web/pull/6022

PR #16797, opened 2026-05-07
Set default AudioLanguagePreference to OriginalLanguage

Lampan-git:
**Changes**

Now when https://github.com/jellyfin/jellyfin/pull/12579 is merged I personally think that the default settings for a user should change to play Original Language by default and disable `PlayDefaultAudioTrack`. 

I understand why the `PlayDefaultAudioTrack = true` default would have made sense before, even though I don't personally agree with it. However now when you can select Original Language, or any other language for that matter, it will prioritize the default track anyways, so Jellyfin will always pick the best track now and if there is no good choice the `Default` track will be picked anyways, so I believe the `PlayDefaultAudioTrack = true` is not the optimal default anymore. See [this test](https://github.com/jellyfin/jellyfin/blob/e1e18e8da015e7311e62cdb62167d51e90331edd/tests/Jellyfin.Server.Implementations.Tests/Library/MediaSourceManagerTests.cs#L57-L148) how the language selection work, I believe it covers all use cases.

**Issues**
N/A

### DIFF
--- a/src/Jellyfin.Database/Jellyfin.Database.Implementations/Entities/User.cs
+++ b/src/Jellyfin.Database/Jellyfin.Database.Implementations/Entities/User.cs
@@ -50,7 +50,8 @@ namespace Jellyfin.Database.Implementations.Entities
             RememberSubtitleSelections = true;
             EnableNextEpisodeAutoPlay = true;
             EnableAutoLogin = false;
-            PlayDefaultAudioTrack = true;
+            PlayDefaultAudioTrack = false;
+            AudioLanguagePreference = "OriginalLanguage";
             SubtitleMode = SubtitlePlaybackMode.Default;
             SyncPlayAccess = SyncPlayUserAccessType.CreateAndJoinGroups;
         }


src/Jellyfin.Database/Jellyfin.Database.Implementations/Entities/User.cs
```diff
@@ -50,7 +50,8 @@ public User(string username, string authenticationProviderId, string passwordRes
             RememberSubtitleSelections = true;
             EnableNextEpisodeAutoPlay = true;
             EnableAutoLogin = false;
-            PlayDefaultAudioTrack = true;
+            PlayDefaultAudioTrack = false;
```
IDisposable, line 53: Does this imply that we should just eliminate the `PlayDefaultAudioTrack` property from Users?

Lampan-git: I don't believe so. Sometimes I believe people mark their preferred stream as default and would want to always pick it regardless. I don't do it, but I believe some people would not be happy if it was removed.

Issue #13712, opened 2023-06-14
Videos from Google Pixel 4a shot in portrait are played in wrong orientation


Skuair:
### Describe the bug

1. Play a video shot in portrait with a pixel 4a (maybe others ?) (player mode is ExoPlayer, same with VLC)
2. Video is played in portrait but content is in landscape stretched to portrait, so deformed

Here is the playback on android tv:
![PXL_20230614_183705754](https://github.com/jellyfin/jellyfin/assets/49756503/d0838cc7-5781-4d09-a888-c352c7c73ecc)

On android version and web player, the video is played in landscape, but not stretched.

Android:
![Screenshot_20230614-204528](https://github.com/jellyfin/jellyfin/assets/49756503/bc554195-9eac-452e-ac02-75d084c4582a)

Web player:
![image](https://github.com/jellyfin/jellyfin/assets/49756503/8f478e5d-002d-4a7a-9446-e99daee71a00)

Is it a video problem ? I don't have other phones so I cannot test portrait with others, all my other videos from GoPro or camera are landscape.
But android phone and web player are not well played as well (but better, i just have to move my head), so i suppose something on the video.

### FFmpeg logs

```shell
No logs, seems to be direct played.
```

### Media info of the file

```shell
Vidéo
Titre: 1080p HEVC SDR
Codec: HEVC
Étiquette du codec: hvc1
Profil: Main
Niveau: 153
Résolution: 1920x1080
Ratio d'aspect original: 16:9
Entrelacé: No
Images par seconde: 59.680187
Débit: 29625 kbps
Plage vidéo: SDR
Video range type: SDR
Espace couleur: smpte170m
Transfert couleur: smpte170m
Couleurs primaires: bt470bg
Format de pixel: yuvj420p
Images de référence: 1
```

### Application version

0.15.10

### Device information

Sony Bravia KD-49XD8005

### Android version

Android 9

### Jellyfin server version

10.8.10

Skuair:
Still an issue.

Skuair:
Still an issue.

Skuair:
Still an issue.

Skuair:
Thanks.
A note about that, I don't remember to have tested that before but the problem is the same with the app vlc android tv, so I suppose it make sense it is device related.

If server can handle that, could be nice.

nyanmisaka:
```
// pseudo code for ATV
if (device_model == "Sony Bravia KD-49XD8005") {
[...]

PR #16803, opened 2026-05-08
Add videoRotation profile condition

nyanmisaka:
**Changes**
- Add videoRotation profile condition
- Check videoRotation in video stream copy
- Add tests for videoRotation profile condition

**Issues**
- Related to #13712,
  it will not take effect until it is implemented by the Android client

Example for client impl:
```js
h264CodecProfileConditions.push({
    Condition: 'Equals',
    Property: 'VideoRotation',
    Value: '0', // If the client cannot handle +-90/180 rotation at all.
    IsRequired: false
});

hevcCodecProfileConditions.push({
    Condition: 'EqualsAny',
    Property: 'VideoRotation',
    Value: '0|90', // If the client can only handle 90-degree rotations.
    IsRequired: false
});
```


Shadowghost (approved):

dmitrylyzo (commented):
LGTM.
It might be worth adding a unit test for this.

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -2466,6 +2466,17 @@ public bool CanStreamCopyVideo(EncodingJobInfo state, MediaStream videoStream)
                 }
             }
 
+            var requestedRotations = state.GetRequestedRotations(videoStream.Codec);
+            if (requestedRotations.Length > 0)
+            {
+                var rotation = state.VideoStream?.Rotation ?? 0;
+                if (rotation != 0
+                    && !requestedRotations.Contains(rotation.ToString(CultureInfo.InvariantCulture), StringComparison.OrdinalIgnoreCase))
```
IDisposable: Why would a numeric string (e.g. 0 or 90), ever have any case-difference? Shouldn't this just be `StringComparison.Ordinal`?

IDisposable (approved):

MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
nyanmisaka: It was a copy-paste. IgnoreCase is useless here.

nyanmisaka:
> LGTM. It might be worth adding a unit test for this.

Tests have been added.

MediaBrowser.Controller/MediaEncoding/EncodingJobInfo.cs
```diff
@@ -629,6 +629,26 @@ public string[] GetRequestedCodecTags(string codec)
             return Array.Empty<string>();
         }
 
+        public string[] GetRequestedRotations(string codec)
+        {
+            if (!string.IsNullOrEmpty(BaseRequest.Rotation))
```
IDisposable: How about:

```csharp
var rotation = BaseRequest.Rotation;

if (string.IsNullOrEmpty(rotation) && !string.IsNullOrEmpty(codec))
{
    rotation = BaseRequest.GetOption(codec, "rotation");
}

return (rotation ?? "").Split(_separators, StringSplitOptions.RemoveEmptyEntries);
```

IDisposable (approved):

MediaBrowser.Controller/MediaEncoding/EncodingJobInfo.cs
nyanmisaka: Applied.

nielsvanvelzen (approved):
API changes look good

MediaBrowser.Controller/MediaEncoding/BaseEncodingJobOptions.cs
```diff
@@ -91,6 +91,12 @@ public BaseEncodingJobOptions()
         /// <value>The codec tag.</value>
         public string CodecTag { get; set; }
 
+        /// <summary>
+        /// Gets or sets the rotation.
+        /// </summary>
+        /// <value>The video rotation angle, usually 0 or +-90/180.</value>
+        public string Rotation { get; set; }
+
```
dmitrylyzo, line 99: If we don't squash commits on merge, this should be in the first commit, since `Rotation` is used there.

dmitrylyzo (commented):
I would refactor the unrelated methods in a separate PR, but this is up to the server team.
Otherwise, it's fine.

MediaBrowser.Controller/MediaEncoding/BaseEncodingJobOptions.cs
nyanmisaka: `MediaStream.Rotation` was added in 10.10, but at that time there were no use cases that required a new profile condition. https://github.com/jellyfin/jellyfin/commit/bf654bde600fa6a67b19d27cf82f600a47ca3638

dmitrylyzo: Oh, this is another `Rotation`. 😅

IDisposable (approved):

Bond-009 (approved):

PR #14153, opened 2025-05-21
Make name sorting case-insensitive

theguymadmax:
**Changes**
Updated `itemSortBy.Name` to perform case-insensitive sorting.

**Issues**
Fixes #11320

PR #16804, opened 2026-05-08
Use SortName when sorting by name

Shadowghost:
We need to use SortName when sorting by name, since it also strips stuff like "The", etc.


IDisposable (approved):
Just a question... if `SortName` hasn't been set in the metadata editor, does it default to the value of `CleanName`?

Bond-009 (approved):

theguymadmax:
The `ItemSortBy.Name` has been historically `Name`, and  I changed it to `CleanName` in https://github.com/jellyfin/jellyfin/pull/14153 to perform case-insensitive sorting. While this change may be fine, it doesn't fix the issue reported on Matrix that browsing artists & albums is being sorted with articles included.

theguymadmax:
Tested - new commit fixes the ordering issue.

Bond-009 (approved):

PR #4984, opened 2021-01-08
Use SubtitleEdit to parse subtitles

Bond-009:

PR #16805, opened 2026-05-08
Write subtitles using SubtitleEdit

Bond-009:
We've been using SubtitleEdit to parse since 2021
https://github.com/jellyfin/jellyfin/pull/4984

I think it's time we start using it to write too


MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
```diff
@@ -83,13 +86,18 @@ private MemoryStream ConvertSubtitles(
...
-                FilterEvents(trackInfo, startTimeTicks, endTimeTicks, preserveOriginalTimestamps);
+                FilterEvents(subtitle, startTimeTicks, endTimeTicks, preserveOriginalTimestamps);
 
-                var writer = GetWriter(outputFormat);
+                var formatter = GetWriter(outputFormat);
+
+                var text = formatter.ToText(subtitle, "untitled");
+                using (var writer = new StreamWriter(stream, Encoding.UTF8, 1024, true))
+                {
+                    writer.Write(text);
```
IDisposable: If we retain the `cancellationToken` argument (which is currently unused and could be deleted), then this should probably be this or similar:

```csharp
await writer.WriteAsync(text, cancellationToken).ConfigureAwait(false); 
```

MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
```diff
@@ -101,26 +109,24 @@ private MemoryStream ConvertSubtitles(
...
         }
 
-        internal void FilterEvents(SubtitleTrackInfo track, long startPositionTicks, long endTimeTicks, bool preserveTimestamps)
+        internal void FilterEvents(Subtitle track, long startPositionTicks, long endTimeTicks, bool preserveTimestamps)
         {
             // Drop subs that have fully elapsed before the requested start position
-            track.TrackEvents = track.TrackEvents
-                .SkipWhile(i => (i.StartPositionTicks - startPositionTicks) < 0 && (i.EndPositionTicks - startPositionTicks) < 0)
-                .ToArray();
+            track.Paragraphs
```
IDisposable, line 99: Is it problematic that we're mutating the incoming `track`'s `Paragraphs` state here? Will that break anything the caller was expecting to remain (perhaps for _seek-backwards_)?

MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
```diff
@@ -265,51 +271,50 @@ internal async Task<SubtitleInfo> GetReadableFile(
...
+                value = new TimedText10();
                 return true;
             }
 
             value = null;
             return false;
         }
 
-        private ISubtitleWriter GetWriter(string format)
+        private Nikse.SubtitleEdit.Core.SubtitleFormats.SubtitleFormat GetWriter(string format)
```
IDisposable, line 306: Maybe add `using Nikse.SubtitleEdit.Core.SubtitleFormats` above and simplify this declaration to `SubtitleFormat`?

Bond-009: We can't we have our own 'SubtitleFormat' type that we need here

IDisposable: Ugh, bummer

MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
```diff
@@ -265,51 +271,50 @@ internal async Task<SubtitleInfo> GetReadableFile(
...
+                value = new AdvancedSubStationAlpha();
                 return true;
             }
 
             if (string.Equals(format, "json", StringComparison.OrdinalIgnoreCase))
             {
-                value = new JsonWriter();
-                return true;
+                throw new NotImplementedException();
             }
```
gabeluci, line 274: This is problematic because the web client requests subtitles in json format.
But I think this whole method can be rewritten. I'll make a separate comment on the PR.

gabeluci:
I like the idea overall and I think this is the way to go (let SubtitleEdit do all the conversions).

The whole `TryGetWriter` method can be simplified or moved, I think.

In the `SubtitleEncoder` class, the `_subtitleParser` variable is a singleton of the [`SubtitleEditParser`](https://github.com/jellyfin/jellyfin/blob/master/MediaBrowser.MediaEncoding/Subtitles/SubtitleEditParser.cs) class. That class has a dictionary (`_subtitleFormatTypes`) that maps file extensions to the respective `SubtitleFormat` class (or multiple, in the case of txt or xml, for example). That class doesn't expose that dictionary right now, but you could add something like the `TryGetWriter` method in the `SubtitleEditParser` class that accepts a file extension and returns the value from the dictionary (a list of `SubtitleFormat`).

On another note, and I know this isn't your code, but this line needs to be changed:

https://github.com/jellyfin/jellyfin/blob/c169184e019d224a60fcbb9125b05de607f42799/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs#L226-L227

I think the `subtitleStream.Codec ??` should be removed. The codec and the file extension are often different (like "subrip" vs "srt", and "webvtt" vs "vtt"). That makes `_subtitleParser.SupportsFileExtension` return false here:

https://github.com/jellyfin/jellyfin/blob/c169184e019d224a60fcbb9125b05de607f42799/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs#L242

because it's looking up the codec name (e.g. "webvtt") in a list of file extensions. Then it falls back to using ffmpeg to convert it to SRT before converting to the desired format (which could have been the same format the file was already in). That also makes it lose position data that VTT and ASS support.

ferezvi:
Hi! I’ve been working on subtitle-related changes lately PR #16745  and I spent some time reviewing your branch and experimenting on top of it.

I’ve been trying an extension-based approach for some of the subtitle JSON handling here:
https://github.com/ferezvi/jellyfin/tree/json-subtitle-helper

Maybe some of it could be useful here, or we could improve the approach further together. I’d be completely happy to help/test/contribute in any way that’s useful.

Bond-009:
Thank you guys for all the reviews, but this was just a POC to for a possibly better solution to https://github.com/jellyfin/jellyfin/pull/16745
Obviously people like the idea, so I'll put in some more effort to make it mergable

Bond-009:
This is a nice starting point that keeps everything functional and is easy to review.
Further improvements can be handled in future PRs.

Shadowghost (approved):

cvium (approved):

ferezvi:
Great work on this branch — the code is clean, well structured and easy to follow. Using libse as the single conversion layer is a much more maintainable approach.

While testing I found that external WebVTT subtitles lose their position cue settings before reaching the client, even when the client supports them natively.

The root cause is that ffprobe reports the codec as `"webvtt"` which `SupportsFileExtension()` doesn't recognize since the dictionary in `GetSubtitleFormatTypes()` only indexes by `Extension` and not `AlternateExtensions`. This causes the file to fall through to the ffmpeg SRT conversion fallback, stripping all cue settings.

The fix is simple — index `AlternateExtensions` alongside `Extension` in `GetSubtitleFormatTypes()` in `SubtitleEditParser.cs`. This way `"webvtt"` is recognized, the file is served directly, and cue settings are preserved. Tested locally on Android with ExoPlayer.

PR #16807, opened 2026-05-09
Fix artist duplicates

### LINKED ISSUE #16311: Music Library shows mixed cases for artist names, even after fixing metadata and rescanning library, replacing all metadata
Image

This is with the latest release of JellyFin, using the provided Docker.

Shadowghost:
We should not consider casing even on raw lookups because otherwise we'd get duplicate artists for e.g. `Thirty Seconds to Mars` and `Thirty Seconds To Mars`.

**Changes**
Ignore casing on raw artist lookup

**Issues**
Fixes: #16311

### DIFF
--- a/Jellyfin.Server.Implementations/Item/LinkedChildrenService.cs
+++ b/Jellyfin.Server.Implementations/Item/LinkedChildrenService.cs
@@ -1,4 +1,6 @@
 #pragma warning disable RS0030 // Do not use banned APIs
+#pragma warning disable CA1304 // Specify CultureInfo
+#pragma warning disable CA1311 // Specify a culture or use an invariant version
 
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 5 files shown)


JPVenson (approved):
LGTM

Jellyfin.Server.Implementations/Item/LinkedChildrenService.cs
```diff
@@ -62,17 +64,19 @@ public IReadOnlyDictionary<string, MusicArtist[]> FindArtists(IReadOnlyList<stri
     {
         using var dbContext = _dbProvider.CreateDbContext();
 
+        var lowerNames = artistNames.Select(n => n.ToLowerInvariant()).ToArray();
         var artists = dbContext.BaseItems
             .AsNoTracking()
             .Where(e => e.Type == _itemTypeLookup.BaseItemKindNames[BaseItemKind.MusicArtist]!)
-            .Where(e => artistNames.Contains(e.Name))
+            .Where(e => lowerNames.Contains(e.Name!.ToLower()))
```
IDisposable, line 71: Why are we using `ToLowerInvariant()` [to lower the artistNames](https://github.com/jellyfin/jellyfin/pull/16807/changes#diff-212c16e9749d622e1bfc613a4dfd8b39f2313d76950605d3914738f00352ce13R67), but using `ToLower()` here? Shouldn't we be consistent?

IDisposable (approved):

Bond-009 (approved):

Jellyfin.Server.Implementations/Item/LinkedChildrenService.cs
Shadowghost: EFCore does not support invariant

PR #16809, opened 2026-05-09
Add Prometheus metrics for users, sessions, and library

Thomas05000005:
**Changes**

This PR adds Prometheus metrics on top of the existing `/metrics` endpoint (activated by `EnableMetrics`). It introduces an `IMetricsCollector` interface and a single `MetricsHostedService` that refreshes every registered collector every 15 seconds.

Four collectors ship with this change:

- **UserMetrics**
  - `jellyfin_users{state="enabled|disabled|admin"}`
  - `jellyfin_users_authentication_provider{provider="..."}`
  - `jellyfin_users_active_recent`
  - `jellyfin_users_failed_logins`
- **SessionMetrics**
  - `jellyfin_sessions{state="playing|idle"}`
  - `jellyfin_sessions_clients{client="..."}`
  - `jellyfin_streams{play_method="DirectPlay|DirectStream|Transcode"}`
- **LibraryMetrics**
  - `jellyfin_library_items{type="Movie|Series|Episode|Audio|..."}`
- **TranscodingMetrics**
  - `jellyfin_transcoding_sessions{hardware="true|false"}`
  - `jellyfin_transcoding_bitrate_bps`

Metric names are consolidated under labels (one `jellyfin_users` series with a `state` label, instead of separate `_total` / `_active` / `_admin` gauges) to match the Prometheus best-practice feedback that came up on #14920.

**Architecture**

- Collectors are pulled by the hosted service rather than pushed from inside `UserManager` / `SessionManager`, so the managers stay clean.
- Adding more metrics (scan duration, plugin counts...) is just a new `IMetricsCollector` file under `Jellyfin.Server.Implementations/Metrics`.
- File-scoped namespaces, sealed classes, no public mutable state.

**Design notes**

A few intentional choices that are worth flagging up-front for reviewers:

- **Pull model, 15s tick.** The default Prometheus scrape interval is 15 seconds, so refreshing the gauges at the same cadence keeps the metric values fresh at scrape time without doing redundant work between scrapes. The interval is a `private static readonly` constant for now; happy to expose it via configuration if there is appetite for it.
- **Background refresh vs. lazy callbacks.** `prometheus-net` supports lazy callbacks (`Gauge.AddCallback` / observable gauges) which would compute values on scrape only. I went with `BackgroundService` because it matches existing Jellyfin patterns (`RecordingsHost`, `DeviceAccessHost`, `LibraryChangedNotifier`), is easier to reason about, and stays consistent with the direction the previous PR (#14920) ended up taking after review feedback. The `IMetricsCollector` interface keeps a future switch to lazy callbacks isolated to `MetricsHostedService.cs` without touching the collectors themselves.
- **Performance under load.** With #15368 having removed the `UserManager` cache, every tick now hits the database for users / sessions / library counts. On a small home server this is invisible; on a large multi-user instance with concurrent activity it could add measurable DB pressure. If reviewers want to address this proactively, options are (a) raising the interval, (b) caching results inside `MetricsHostedService` between ticks, or (c) switching to lazy callbacks (see point above). I would prefer to wait for actual feedback before adding configuration surface here.
- **Web UI is intentionally out of scope.** Prometheus metrics are designed to be scraped by external tooling (Grafana, varken, jellyfin-exporter, etc.). Native widgets in the Jellyfin Dashboard would duplicate that tooling. If the project wants a "Server health" widget later, it can land as a separate PR on `jellyfin-web` consuming existing JSON endpoints (`/Sessions`, `/Users`, `/Library/Items/Counts`) rather than `/metrics`.

**Tests**

Unit tests added for all four collectors and the hosted service in `tests/Jellyfin.Server.Implementations.Tests/Metrics/` (xUnit + Moq, 14 tests total, all passing locally). They cover:

- `*MetricsTests`: smoke tests on `CollectAsync` for each collector with realistic and empty inputs, plus a `Name` sanity check.
- `MetricsHostedServiceTests`: confirms that `CollectAsync` is never invoked while `EnableMetrics=false`, and that an exception thrown by a collector is caught and logged without bringing down the hosted service.

**Runtime testing performed**

Built against master (.NET 10) and ran on Windows with `--nowebclient --datadir <tmp>`:

| Scenario | Result |
| --- | --- |
| `EnableMetrics=true` -> `GET /metrics` | All `jellyfin_*` series visible |
| Create a 2nd user via API | `state="enabled"` 1 -> 2 after the 15s tick |
| Disable the 2nd user via Policy API | `state="enabled"` 2 -> 1, `state="disabled"` 0 -> 1 |
| Authenticate as `Client="test"` | `jellyfin_sessions_clients{client="test"} 1` appears |
| Logged-in user activity | `jellyfin_users_active_recent` 0 -> 1 |
| `EnableMetrics=false` | `GET /metrics` -> 404 (collectors no-op) |

**Relation to #14920**

#14920 has been inactive since 2025-12-15. This PR picks up the same direction (#3016) but reorganises the code along the lines reviewers asked for there:

- single hosted service driving an `IMetricsCollector` interface (no manual `UpdateUserMetrics()` calls scattered through `UserManager`),
- consolidated metric names using labels (per @jpds' feedback on cardinality),
- file-scoped namespaces (per @Bond-009's feedback).

Happy to close this and rebase onto #14920 if @ysf3rdm comes back to finish it.

**Issues**

#3016

### Code assistance

Used Claude (LLM) to help with code drafting and to format this description.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -154,6 +154,7 @@
  - [Terror-Gene](https://github.com/Terror-Gene)
  - [ThatNerdyPikachu](https://github.com/ThatNerdyPikachu)
  - [ThibaultNocchi](https://github.com/ThibaultNocchi)
+ - [Thomas05000005](https://github.com/Thomas05000005)
  - [thornbill](https://github.com/thornbill)
  - [ThreeFive-O](https://github.com/ThreeFive-O)
  - [tjwalkr3](https://github.com/tjwalkr3)
--- a/Jellyfin.Server.Implementations/Jellyfin.Server.Implementations.csproj
+++ b/Jellyfin.Server.Implementations/Jellyfin.Server.Implementations.csproj
@@ -28,6 +28,7 @@
   <ItemGroup>
     <PackageReference Include="AsyncKeyedLock" />
     <PackageReference Include="Microsoft.EntityFrameworkCore.Relational" />
+    <PackageReference Include="prometheus-net" />
   </ItemGroup>
 
   <ItemGroup>
--- a/Jellyfin.Server.Implementations/Metrics/IMetricsCollector.cs
+++ b/Jellyfin.Server.Implementations/Metrics/IMetricsCollector.cs
@@ -1,0 +1,22 @@
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Jellyfin.Server.Implementations.Metrics;
+
+/// <summary>
+/// Defines a Prometheus metrics collector that refreshes its gauges or counters when invoked.
+/// </summary>
+public interface IMetricsCollector
+{
+    /// <summary>
+    /// Gets the human-readable name of the collector, used in log output.
+    /// </summary>
+    string Name { get; }
+
+    /// <summary>
+    /// Refreshes all metrics owned by the collector.
+    /// </summary>
+    /// <param name="cancellationToken">A cancellation token.</param>
+    /// <returns>A <see cref="Task"/> representing the asynchronous operation.</returns>
+    Task CollectAsync(CancellationToken cancellationToken);
+}
--- a/Jellyfin.Server.Implementations/Metrics/LibraryMetrics.cs
+++ b/Jellyfin.Server.Implementations/Metrics/LibraryMetrics.cs
@@ -1,0 +1,61 @@
+using System.Threading;
+using System.Threading.Tasks;
+using Jellyfin.Data.Enums;
+using MediaBrowser.Controller.Entities;
+using MediaBrowser.Controller.Library;
+using Prometheus;
+
+namespace Jellyfin.Server.Implementations.Metrics;
+
+/// <summary>
+/// Exposes Prometheus metrics describing the contents of the media library.
+/// </summary>
+public sealed class LibraryMetrics : IMetricsCollector
+{
+    private static readonly BaseItemKind[] _trackedKinds =
+    [
+        BaseItemKind.Movie,
+        BaseItemKind.Series,
+        BaseItemKind.Season,
+        BaseItemKind.Episode,
+        BaseItemKind.MusicArtist,
+        BaseItemKind.MusicAlbum,
+        BaseItemKind.Audio,
+        BaseItemKind.AudioBook,
+        BaseItemKind.Book,
+        BaseItemKind.Photo,
+        BaseItemKind.Video,
+    ];
+
+    private static readonly Gauge _items = Prometheus.Metrics
+        .CreateGauge("jellyfin_library_items", "Number of items in the media library grouped by item type.", "type");
+
+    private readonly ILibraryManager _libraryManager;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="LibraryMetrics"/> class.
+    /// </summary>
+    /// <param name="libraryManager">The library manager.</param>
+    public LibraryMetrics(ILibraryManager libraryManager)
+    {
+        _libraryManager = libraryManager;
+    }
+
+    /// <inheritdoc />
+    public string Name => nameof(LibraryMetrics);
+
+    /// <inheritdoc />
+    public Task CollectAsync(CancellationToken cancellationToken)
+    {
+        foreach (var kind in _trackedKinds)
+        {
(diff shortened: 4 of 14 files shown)


nielsvanvelzen:
How is this different from #14920 / #14052? Does this expand upon or is it a complete rewrite?

Thomas05000005:
Hi @nielsvanvelzen, thanks for taking a look.

**It's a complete rewrite, not an extension.**

---

**Difference vs #14920** (same goal #3016, different implementation)

This PR applies the changes that reviewers requested on #14920 but that were never applied (PR has been inactive since December 2025):

- **Interface-based architecture** (requested by @JPVenson): `IMetricsCollector` + a single `MetricsHostedService` that pulls the data, instead of `UpdateMetrics()` being injected into `UserManager` / `SessionManager`.
- **File-scoped namespaces** (requested by @Bond-009).
- **Metrics consolidated with Prometheus labels** (requested by @jpds across 5 cardinality comments) — e.g. `jellyfin_users{state="enabled|disabled|admin"}` instead of three separate gauges `_total` / `_active` / `_admin`.
- **`LibraryMetrics` + `TranscodingMetrics`** added (these did not exist in #14920; @oddstr13 had mentioned wanting to add more metrics inside this structure).
- **14 unit tests** with xUnit + Moq (Jellyfin has a dedicated `Jellyfin.Server.Implementations.Tests` project; #14920 had none).

#14920 is currently `CONFLICTING` since #15368 was merged and would need a rebase plus the application of the reviewer feedback. I'm happy to close this PR if @ysf3rdm comes back to finish theirs.

---

**Difference vs #14052** (orthogonal, no conflict)

The two PRs cover entirely different topics:

- **#14052** adds *configuration*: enabling Prometheus via `JELLYFIN_EnableMetrics=true` (env var) and exposing `/metrics` on a dedicated port. It does not add any custom metric.
- **This PR** adds *content*: 24 `jellyfin_*` series (users, sessions, streams, library, transcoding). It does not touch the configuration.

No files in common → both can be merged independently in any order. They are actually complementary: a typical Docker/K8s deployment would use #14052 to enable and scope the port, and this PR to expose useful content.

---

Please let me know if any of the metrics don't match what you'd expect, or if you have suggestions for metrics to add or remove. Happy to adjust the scope.

nielsvanvelzen:
Ok thanks for explaining, then another question I have (which also applies to the other PR's really). Can't this functionality be achieved with a plugin?

Thomas05000005:
Good question.

In theory yes, technically this could be done in a plugin. In practice though, **the Prometheus infrastructure is already in core** and has been for a while:

- 3 `PackageReference`s in `Jellyfin.Server.csproj` and `Emby.Server.Implementations.csproj`: `prometheus-net`, `prometheus-net.AspNetCore`, `prometheus-net.DotNetRuntime`
- `Startup.cs` already calls `UseHttpMetrics()` + `MapMetrics()`
- `ServerConfiguration.cs` already exposes the `EnableMetrics` flag

So the pipe (the `/metrics` endpoint + middleware) is already in core. This PR only adds the content that flows through it (the `jellyfin_*` gauges). Putting the content in a plugin while the pipe stays in core would be architecturally inconsistent.

To go for a proper plugin approach, you'd need to:

1. Remove the `prometheus-net*` references from the core csproj files
2. Remove `UseHttpMetrics()` / `MapMetrics()` from `Startup.cs`
3. Remove `EnableMetrics` from `ServerConfiguration.cs`
4. Create a plugin that reintroduces all of the above through `IPluginServiceRegistrator`

That's a wider architectural change that goes beyond this PR (and would equally apply to #14920 and #14052). If the Jellyfin team wants to take that direction, I'm happy to open a discussion on jellyfin-meta to decide — but it would be a separate call.

Worth noting: external exporters already exist (`jellyfin-exporter`, `jellyfin_exporter`) that hit the Jellyfin REST API without needing a `/metrics` endpoint at all. That's a third option if the project wants to fully move Prometheus out of the server.

Does the team have a clear preference here? If so I can adapt, but it would make sense for all the active Prometheus PRs to follow the same direction.

christiaangoossens:
Having some actual data in the existing metrics endpoint would be really useful. Right now it doesn't really have any use besides process tracking, which you can already do in many other ways. To me this feels like it just belongs with the core metrics functionality.

PR #16813, opened 2026-05-09
Safeguard against invalid GUIDs

Shadowghost:
Apparently people have invalid mediaSourceIds in their database which breaks deletion and migrations that use it.

**Changes**
* Add safeguard against invalid GUIDs

### DIFF
--- a/Jellyfin.Server/Migrations/Routines/MigrateLinkedChildren.cs
+++ b/Jellyfin.Server/Migrations/Routines/MigrateLinkedChildren.cs
@@ -283,9 +283,9 @@ internal class MigrateLinkedChildren : IDatabaseMigrationRoutine
             .Select(id => _libraryManager.GetItemById(id))
             .Where(item => item is not null)
             .ToList();
-        _libraryManager.DeleteItemsUnsafeFast(itemsToDelete!);
+        var deleted = SafeDeleteItems(itemsToDelete!);
 
-        _logger.LogInformation("Removed {Count} wrong-type alternate version items. They will be recreated with the correct type on next library scan.", itemsToDelete.Count);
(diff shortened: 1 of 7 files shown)


gnattu (approved):

jumoog (approved):

Jellyfin.Server/Migrations/Routines/MigrateLinkedChildren.cs
```diff
@@ -431,9 +431,43 @@ private void CleanupStaleFileEntries(JellyfinDbContext context)
...
             .Where(item => item is not null)
             .ToList();
-        _libraryManager.DeleteItemsUnsafeFast(itemsToDelete!);
+        var deleted = SafeDeleteItems(itemsToDelete!);
 
-        _logger.LogInformation("Removed {Count} stale items.", itemsToDelete.Count);
+        _logger.LogInformation("Removed {Count} stale items.", deleted);
+    }
+
+    private int SafeDeleteItems(IReadOnlyCollection<MediaBrowser.Controller.Entities.BaseItem> items)
```
JPVenson: no thats not really the point of the UnsafeFast delete. It doesn't check if references on the filesystem exists or within the database or on the filesytem. Just putting a try-catch around that doesn't really help with much. use the normal Delete method instead if you are not absolutly sure you are only deleting base items that can never have references

JPVenson (changes requested):
does it make sense to also try to convert the mediaSourceId to a Guid in the DB in the same breath now that its properly cleaned up?

Shadowghost:
We need to ignore theinvlaid ones because livetv produces non-guid media source ids

JPVenson (approved):

Bond-009 (approved):

PR #16814, opened 2026-05-09
Add OpenTelemetry instrumentation

Shadowghost:
**Changes**
Add OTEL instrumentation and configuration

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -53,6 +53,12 @@
     <PackageVersion Include="Moq" Version="4.18.4" />
     <PackageVersion Include="NEbml" Version="1.1.0.5" />
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.4" />
+    <PackageVersion Include="OpenTelemetry.Exporter.OpenTelemetryProtocol" Version="1.18.0" />
+    <PackageVersion Include="OpenTelemetry.Extensions.Hosting" Version="1.18.0" />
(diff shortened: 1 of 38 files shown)


jester-xbmc:
I'm probably overly cautious, but from a privacy point of view, maybe it is good to explain why this telemetry needed or added as standard ? 

Also, depending on its usage, a suggestion to make it a plugin and not builtin.

enter-a-random-username:
> I'm probably overly cautious, but from a privacy point of view, maybe it is good to explain why this telemetry needed or added as standard ?
> 
> Also, depending on its usage, a suggestion to make it a plugin and not builtin.

Thought the same thing, but looking over the code, it looks like it is a telemetry endpoint for you. So you can monitor you instance, if you want to.

Shadowghost:
OpenTelemetry is not sending something to any service, you can enable it and when configured, it will send logs, metrics and traces to your own open telemtry compatible monitoring system. Just having it in the code does nothing without configuration.

h1ght:
> I'm probably overly cautious, but from a privacy point of view, maybe it is good to explain why this telemetry needed or added as standard ?
> 
> Also, depending on its usage, a suggestion to make it a plugin and not builtin.

such telemetry is optional, for this you need to configure an endpoint like alloy->tempo/loki. maybe it is intersting for the devs. not really for an enduser.

h1ght:
@Shadowghost what i have found by accident, there is a plugin that is some1 already created. https://github.com/axolmain/jellyfin-otel-plugin

PR #16818, opened 2026-05-10
Episode list on season page should sort by part number to break ties

### LINKED ISSUE #16817: Episode list on season page should sort by part number to break ties
For a show, the default metadata provider identifies the pilot as a single episode. Other providers identify it as two separate parts. I have two files for this episode, and believe that to be common.

To appease the metadata, I have accepted the single episode demarcation, but provided two files: part-1 and part-2. This works, except in one place. The sort order of the season episode list page will put part-2 first. I believe it is doing so by chance. The part-X is missing from the sort key.

tcox2:
**Changes**

For a show, the default metadata provider identifies the pilot as a single episode. Other providers identify it as two separate parts. I have two files for this episode, and believe that to be common.

To appease the metadata, I have accepted the single episode demarcation, but provided two files: part-1 and part-2. This works, except in one place. The sort order of the season episode list page will put part-2 first. I believe it is doing so by chance. The part-X is missing from the sort key.

**Issues**

Fixes #16817

### DIFF
--- a/Emby.Server.Implementations/Sorting/AiredEpisodeOrderComparer.cs
+++ b/Emby.Server.Implementations/Sorting/AiredEpisodeOrderComparer.cs
@@ -1,7 +1,10 @@
 #pragma warning disable CS1591
 
 using System;
+using System.IO;
+using System.Text.RegularExpressions;
 using Jellyfin.Data.Enums;
+using Jellyfin.Extensions;
 using MediaBrowser.Controller.Entities;
 using MediaBrowser.Controller.Entities.TV;
 using MediaBrowser.Controller.Sorting;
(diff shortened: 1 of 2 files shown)


theguymadmax:
Since we implemented multiple versions in 12.0, your naming no longer produces separate versions; it just creates alternative ones. So this fix doesn’t seem necessary

tcox2:
> Since we implemented multiple versions in 12.0, your naming no longer produces separate versions; it just creates alternative ones. So this fix doesn’t seem necessary

hi - thanks for looking
to clarify:  my intent was never to create versions.  just to fix sort order.   how is my change related to versions?

theguymadmax:
```
/series/Season 01/Show S01E01-part-1 - Pilot.mkv
/series/Season 01/Show S01E01-part-2 - Pilot.mkv

/series/Season 01/Show S01E01 pt A - Pilot.mkv
/series/Season 01/Show S01E01 pt B - Pilot.mkv
```
This will not lead to separate versions in 12.0. But will create versions of the same episode.  Your fix is for a workaround that no longer works.

tcox2:
> ```
> /series/Season 01/Show S01E01-part-1 - Pilot.mkv
> /series/Season 01/Show S01E01-part-2 - Pilot.mkv
> 
> /series/Season 01/Show S01E01 pt A - Pilot.mkv
> /series/Season 01/Show S01E01 pt B - Pilot.mkv
> ```
> 
> This will not lead to separate versions in 12.0. But will create versions of the same episode.

so the meaning of 'version' here is not 'substitutes', but rather 'grouping'?   i.e. part 1 is a version, part 2 is a version, and they are keyed to an episode?

happy to revisit upon 12.0

theguymadmax:
Episodes will now be stacked. Whereas before that would lead to separate entries. That behavior is already in the Master branch/unstable builds, if you want to evaluate.

PR #16821, opened 2026-05-10
Allow plugins to customize “Similar” recommendations via composable providers

mkostova2:
This change lets third-party code supply the item IDs used for the Similar row on movies, shows, music, and other supported types, while keeping the existing server-side fallback when no plugin returns results.

## Behavior

- On each Similar request, the server asks registered `IItemSimilarityProvider` implementations, in order (`IHasOrder` when implemented, then name), for the first implementation whose `Supports` method accepts the item.
- If a provider returns one or more GUIDs, those are resolved to DTOs and returned. If every provider returns empty, errors, or declines the item, the previous metadata-based random query is unchanged.
- `GET Libraries/AvailableOptions` gains a `similarityProviders` list so admins and clients can see which similarity plugins are loaded, consistent with subtitle, lyric, and media-segment listings.
- Similar requests now honor the HTTP `CancellationToken` when calling into providers.

## Plugin authors

- Implement `IItemSimilarityProvider<T>` (or `IItemSimilarityProvider` for custom `Supports` / `GetSimilarItems` behavior) from `MediaBrowser.Controller.Providers`.
- Ship your types in a plugin assembly that Jellyfin already composes; `IItemSimilarityProvider` exports are collected in `FindParts` the same way as image and metadata providers (`GetExports<IItemSimilarityProvider>()`). No extra `IPluginServiceRegistrator` boilerplate is required for basic discovery.

## Tests

- `ProviderManager` tests cover similarity `Supports` filtering and full existing provider-ordering behavior.
- API tests cover the `IItemSimilarityProvider<Audio>` surface (success, empty, and exception from the typed `GetSimilarItems` path).

## Breaking / contract

- `LibraryOptionsResultDto` is extended with `similarityProviders`. OpenAPI generation should be run or left to the repo's usual workflow so consumers pick up the new field.

### DIFF
--- a/Emby.Server.Implementations/ApplicationHost.cs
+++ b/Emby.Server.Implementations/ApplicationHost.cs
@@ -692,7 +692,8 @@ namespace Emby.Server.Implementations
                 GetExports<IMetadataProvider>(),
                 GetExports<IMetadataSaver>(),
                 GetExports<IExternalId>(),
-                GetExports<IExternalUrlProvider>());
+                GetExports<IExternalUrlProvider>(),
+                GetExports<IItemSimilarityProvider>());
 
             Resolve<IMediaSourceManager>().AddParts(GetExports<IMediaSourceProvider>());
         }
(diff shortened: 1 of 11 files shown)


JPVenson:
already in implementation by https://github.com/jellyfin/jellyfin/pull/15970

PR #16829, opened 2026-05-11
Improve OriginalLanguage normalization and inheritance

Lampan-git:
**Changes**

Followup to https://github.com/jellyfin/jellyfin/pull/12579. This cleans stuff up so it's easier to use elsewhere in the repo. As @MarcoCoreDuo pointed out in https://github.com/jellyfin/jellyfin/pull/16001, it wasn't really ideal before. 

- Fix the commas in OMDb provider itself.
- Move the parent fallback to GetInheritedOriginalLanguage()
- Normalize OriginalLanguage in the setter.

### DIFF
--- a/Emby.Server.Implementations/Library/MediaSourceManager.cs
+++ b/Emby.Server.Implementations/Library/MediaSourceManager.cs
@@ -440,10 +440,6 @@ namespace Emby.Server.Implementations.Library
 
             if (string.Equals(user.AudioLanguagePreference, "OriginalLanguage", StringComparison.OrdinalIgnoreCase))
             {
-                originalLanguage = !string.IsNullOrWhiteSpace(originalLanguage)
-                    ? originalLanguage.Split(',').FirstOrDefault()
-                    : null;
-
                 if (user.PlayDefaultAudioTrack)
                 {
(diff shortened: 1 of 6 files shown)


MarcoCoreDuo:
Thanks for the PR. Wouldn't it be better to always return the TwoLetterISOLanguageName of the CultureDto? It would return either a language code based on ISO 639-1 (e.g. en) or a combination of ISO 639-1 and ISO 3166-1 codes (e.g. en-US). The same format is also returned by preferredMetadataLanguage.

Lampan-git:
> Thanks for the PR. Wouldn't it be better to always return the TwoLetterISOLanguageName of the CultureDto? It would return either a language code based on ISO 639-1 (e.g. en) or a combination of ISO 639-1 and ISO 3166-1 codes (e.g. en-US). The same format is also returned by preferredMetadataLanguage.

Yeah that make sense. I just added the same normalization I saw elsewhere.

IDisposable (approved):

MarcoCoreDuo (approved):

Shadowghost (approved):

theguymadmax (approved):

PR #16836, opened 2026-05-13
Fix subtitle currentFormat to always use file extention

gabeluci:
**Changes**

The original line was this:

https://github.com/jellyfin/jellyfin/blob/c169184e019d224a60fcbb9125b05de607f42799/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs#L226-L227

The codec and the file extension are often different (like "subrip" vs "srt", and "webvtt" vs "vtt"). That makes `_subtitleParser.SupportsFileExtension` return false here:

https://github.com/jellyfin/jellyfin/blob/c169184e019d224a60fcbb9125b05de607f42799/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs#L242

because it's looking up the codec name (e.g. "webvtt") in a list of file extensions. Then it falls back to using ffmpeg to convert it to SRT before converting to the desired format (which could have been the same format the file was already in). That also makes it lose position data from VTT and ASS files.

I removed `subtitleStream.Codec ??` so it only uses the file extension. At this point in the code, we're only dealing with external subtitles, so `subtitleStream.Path` will always have a value.

#16805 may make this moot, but I have a feeling that may not be ready for the next release. This is a quick and easy fix.

### DIFF
--- a/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
+++ b/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
@@ -223,8 +223,7 @@ namespace MediaBrowser.MediaEncoding.Subtitles
                 };
             }
 
-            var currentFormat = subtitleStream.Codec ?? Path.GetExtension(subtitleStream.Path)
-                .TrimStart('.');
+            var currentFormat = Path.GetExtension(subtitleStream.Path).TrimStart('.');
 
             // Handle PGS subtitles as raw streams for the client to render
             if (MediaStream.IsPgsFormat(currentFormat))


ferezvi:
I see the appeal of simplifying to just the extension, but Codec comes from ffprobe and reflects the actual detected format — a .vtt file could contain SRT data, for example.

The real pain point seems to be the scattered format aliases (webvtt/vtt, subrip/srt). Since we're running into the same issues, maybe we could tackle this together? Something like keeping the ` Codec ?? extension ` fallback but normalizing the format name early:

```
var currentFormat = NormalizeFormat(
    subtitleStream.Codec ?? Path.GetExtension(subtitleStream.Path).TrimStart('.')
);
```

Happy to collaborate on a solution that works for both our use cases.

gabeluci:
I guess that's fair. I have an idea to allow using the codec name.

I want to avoid manually creating a mapping (like "subrip" = "srt", etc.)

gabeluci:
Now that I'm looking at the code again, I'm going to leave this be. That value is used later in the code as a file extension (when it looks up which writer to use) so unless that downstream code is going to change, then it's just wrong to set this value as the codec instead of the file extension.

If the file extension doesn't match the codec, it'll still work, but it'll go through that fallback where it converts to SRT with ffmpeg. This PR doesn't change that - it would have done that before.

ferezvi:
Before closing it, wanted to share an approach that worked for me. Instead of changing the precedence, you can wrap currentFormat with a normalization method and apply it in two more places:

```
// In GetReadableFile
var currentFormat = NormalizeSubtitleFormat(
    subtitleStream.Codec ?? Path.GetExtension(subtitleStream.Path).TrimStart('.'));

// In GetSubtitles, after GetSubtitleStream
inputFormat = NormalizeSubtitleFormat(inputFormat);
outputFormat = NormalizeSubtitleFormat(outputFormat);

// New method
private static string NormalizeSubtitleFormat(string format) =>
    format.ToLowerInvariant() switch
    {
        "webvtt" => "vtt",
        "subrip" => "srt",
        _ => format
    };
```
Tested locally with an external .vtt file — it goes directly to the parser without the ffmpeg fallback. Happy to collaborate if you want to take it further.

gabeluci:
But that's basically saying, "if it's the name of the codec, set it to the extension."

Yes, there could be cases where the codec doesn't match the file extension, but that's going to be very rare, and in those cases, the user can just fix the file name if it doesn't work.

I get the impression all of this code is going to be refactored, so at that point, all of the code can be changed to work with the codec and not the file extension, if that's desired.

I'll leave it to the maintainers to comment.

gabeluci:
Closing in favour of #16998

PR #16837, opened 2026-05-13
Misc fixes

Shadowghost:
Since all of them are small I opted for one PR.

**Changes**
* Order chapter response by start time
* Properly handle cancellation in MediaSegmentManager
* Prevent unecessary log spam in NetworkUtils

### DIFF
--- a/Jellyfin.Server.Implementations/MediaSegments/MediaSegmentManager.cs
+++ b/Jellyfin.Server.Implementations/MediaSegments/MediaSegmentManager.cs
@@ -81,6 +81,11 @@ public class MediaSegmentManager : IMediaSegmentManager
 
             foreach (var provider in providers)
             {
+                if (cancellationToken.IsCancellationRequested)
+                {
+                    break;
+                }
+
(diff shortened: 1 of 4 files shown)


Jellyfin.Server.Implementations/MediaSegments/MediaSegmentManager.cs
```diff
@@ -146,6 +151,15 @@ public async Task RunSegmentPluginProviders(BaseItem baseItem, LibraryOptions li
                         await CreateSegmentAsync(segment, providerId).ConfigureAwait(false);
                     }
                 }
+                catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
+                {
+                    throw;
```
JPVenson, line 153: should that not be a break; or the new break; in L86 an exception?

JPVenson (changes requested):

IDisposable (approved):

JPVenson (approved):
Still not super sure about the usefulness of MediaSegmentManagerL155 but good enough

Bond-009 (approved):

PR #16839, opened 2026-05-13
ImageController: drop Vary: Accept and Content-Disposition: attachment for cacheable image responses

Finomosec:
## Changes

Removes two response headers from `ImageController.GetImageResult` that
prevented effective browser caching of image responses:

- `Vary: Accept` — caused the browser to key its HTTP cache by the
  `Accept` request header. `` tags send `Accept:
  image/avif,image/webp,...` while `fetch()` and the jellyfin-web
  client send `Accept: */*`. The browser then stored separate cache
  entries for the same URL — one per Accept variant — causing repeated
  network round-trips for resources that should have been served from
  cache.
- `Content-Disposition: attachment` — instructed the browser to treat
  the response as a download rather than inline content. This is
  semantically wrong for an endpoint whose primary consumer is ``
  tags and can interfere with browser caching of image responses.

## Issues

Closes _no specific open issue_, but addresses a long-standing
complaint that image loads on library pages stay slow even on repeated
visits.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

Reproduced and verified against jellyfin 10.11.6 in Docker.

**Before the patch** (Chrome resource-timing API):

- 245 image requests on a library view
- 64 of 245 served from cache (~26 %)
- The remaining 181 were full re-downloads even though the responses
  carried `Cache-Control: public, max-age=31536000`

**After the patch:**

- Same library view, repeated visit
- 0 of N requests hit the network (all served from disk cache)
- Cross-request-type caching works: a response fetched once via
  `fetch()` is reused by a subsequent `` request and vice versa

The URL already encodes the cache tag and all size parameters, so
removing `Vary: Accept` does not risk serving the wrong content to
clients with different `Accept` headers in practice — each
(URL + tag + size) tuple maps to exactly one resized image variant
(the format is chosen on first encode and stored as a separate cache
file).

## Checklist:

- [x] My code follows the code style of this project.
- [x] I have performed a self-review of my own code.
- [x] My changes generate no new warnings.

### DIFF
--- a/Jellyfin.Api/Controllers/ImageController.cs
+++ b/Jellyfin.Api/Controllers/ImageController.cs
@@ -2009,9 +2009,6 @@ public class ImageController : BaseJellyfinApiController
 
         Response.ContentType = imageContentType ?? MediaTypeNames.Text.Plain;
         Response.Headers.Append(HeaderNames.Age, Convert.ToInt64((DateTime.UtcNow - dateImageModified).TotalSeconds).ToString(CultureInfo.InvariantCulture));
-        Response.Headers.Append(HeaderNames.Vary, HeaderNames.Accept);
-
-        Response.Headers.ContentDisposition = "attachment";
 
         if (disableCaching)
         {
--- a/Jellyfin.Api/Controllers/ImageController.cs
+++ b/Jellyfin.Api/Controllers/ImageController.cs
@@ -2010,6 +2010,18 @@ public class ImageController : BaseJellyfinApiController
         Response.ContentType = imageContentType ?? MediaTypeNames.Text.Plain;
         Response.Headers.Append(HeaderNames.Age, Convert.ToInt64((DateTime.UtcNow - dateImageModified).TotalSeconds).ToString(CultureInfo.InvariantCulture));
 
+        // CVE-2024-43801 / GHSA-vcmh-9wx9-rfqh: an SVG profile picture rendered
+        // inline by the browser can execute JavaScript in the Jellyfin origin and
+        // exfiltrate the viewer's access token from localStorage. Keep forcing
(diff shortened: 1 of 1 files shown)


nielsvanvelzen (changes requested):
https://github.com/jellyfin/jellyfin/security/advisories/GHSA-vcmh-9wx9-rfqh

Finomosec:
Thanks for the pointer — I missed the security context of #12490, my apologies.

Pushed a follow-up commit (cebeec6) that restores the mitigation: `Content-Disposition: attachment` is now applied conditionally, only when the response is NOT a known inline-safe raster image (JPEG / PNG / WebP / GIF / AVIF / BMP / TIFF / ICO). SVG and any other / unknown content type continue to be served as attachments, so CVE-2024-43801 stays closed.

Rationale: the raster formats can't execute scripts, so forcing the browser to download them was overly broad. Keeping the header on for SVG (the actual exploit vector) preserves the LocalStorage-isolation property; allowing it off for raster formats unblocks browser caching from `` consumers, which was the original motivation for this PR.

The `Vary: Accept` removal is unchanged — that one has no security bearing.

I verified locally against 10.11.6:

- a movie poster (`image/jpeg`) is now served without `Content-Disposition`; an `` and a `fetch()` for the same URL share a single browser-cache entry, and a repeated page visit produces zero network round-trips for image resources
- a profile picture served as `image/svg+xml` still receives `Content-Disposition: attachment`

Happy to adjust the allow-list — let me know if you'd prefer a different policy (e.g. checking `MimeTypes.IsImage` plus an SVG exclusion, or a configurable hardening level).

Finomosec:
Is it ok to merge now? Or do you have any other feedback / change requests?

PR #16840, opened 2026-05-13
ImageProcessor: lock per source image instead of globally on encode

Finomosec:
## Changes

`ImageProcessor.ProcessImage` previously guarded the encoder call with
a single process-wide `AsyncNonKeyedLocker` capped at
`Environment.ProcessorCount`. Encoding two completely unrelated source
images was therefore serialized against the same set of slots, even
though the work is fully independent.

This PR introduces a two-layer locking strategy:

1. **Outer `AsyncKeyedLocker<string>`** keyed by `originalImagePath`.
   Two resize requests for _different_ source images can now run in
   parallel without contending; two requests for the _same_ source are
   still serialized. After acquiring the per-source lock, the cache
   file existence is rechecked so that a request that lost the race
   reuses the just-produced output instead of redoing identical work.
2. **Inner `AsyncNonKeyedLocker`** retained with the existing
   semantics — it acts as a global memory-pressure cap on
   simultaneous encoder invocations (the original intent of the
   `ParallelImageEncodingLimit` configuration).

The `AsyncKeyedLock` package is already a dependency (the namespace
is in `ImageProcessor`'s `using` directives and is used across the
codebase, e.g. `ProviderManager._imageSaveLock`,
`TranscodeManager._transcodingLocks`). The pool-size constants
(`PoolSize = 20`, `PoolInitialFill = 1`) match the pattern used
elsewhere.

## Issues

Closes _no specific open issue_, but is a small correctness/contention
improvement to the image-encoding pipeline.

## Type of change

- [x] Refactor (non-breaking change that improves contention behavior)

## How Has This Been Tested?

Builds locally against `master` (net10.0) with zero warnings. Same
codepath was deployed and exercised manually against a running
10.11.6 Docker container; image responses behave identically and no
regressions were observed. The user-visible effect is a contention-
only one — it does not change any cache layout, request handling, or
output bytes.

## Checklist:

- [x] My code follows the code style of this project.
- [x] I have performed a self-review of my own code.
- [x] My changes generate no new warnings.

### DIFF
--- a/src/Jellyfin.Drawing/ImageProcessor.cs
+++ b/src/Jellyfin.Drawing/ImageProcessor.cs
@@ -42,6 +42,7 @@ public sealed class ImageProcessor : IImageProcessor, IDisposable
     private readonly IImageEncoder _imageEncoder;
 
     private readonly AsyncNonKeyedLocker _parallelEncodingLimit;
+    private readonly AsyncKeyedLocker<string> _perSourceEncodingLock;
 
     private bool _disposed;
 
@@ -72,6 +73,11 @@ public sealed class ImageProcessor : IImageProcessor, IDisposable
         }
 
         _parallelEncodingLimit = new(semaphoreCount);
+        _perSourceEncodingLock = new(o =>
+        {
+            o.PoolSize = 20;
+            o.PoolInitialFill = 1;
+        });
     }
 
     private string ResizedImageCachePath => Path.Combine(_appPaths.ImageCachePath, "resized-images");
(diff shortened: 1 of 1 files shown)


Shadowghost (commented):
1st things first: Please use our PR template.

For the changes:
I think we should rather lock on `cacheFilePath` since in the end that is the deciding factor. AsyncKeyLocker by default uses `Ordinal` but our check is using `OrdinalIgnoreCase` - should be consistent (either switch AsyncKeyLocker or our comparision)

It would also be good to add a comment why `PoolSize=20` was chosen or make it dynamic based on CPU count.

If possible we should also add some tess (at least for the logic part, concurrency is hard to test).

crobibero:
Closing per https://jellyfin.org/docs/general/contributing/llm-policies/

Finomosec:
> Closing per https://jellyfin.org/docs/general/contributing/llm-policies/

Even if my PR was not good enough...
Will you look into the problem itself and solve it?
That would be a good thing, as it makes the UI very slow and unresponsive at times.

Greetings Fino;

Shadowghost:
You can always resubmit with the template being properly filled out if you 100% understand what the changes do.
But I will take a look later today.

Issue #11282, opened 2024-04-02
[Issue] rc_mode not supported!  jellyfin has no config to set the rc_mode


woshilaiba:
I encountered the problem like this:
[h264_vaapi @ 0x1baad80] Driver does not support VBR RC mode (**supported modes: CQP**). 

It says ffmpeg does not support rc_mode VBR,  CBR neither.
**I try "-rc_mode CQP" as it mentioned on ffmpeg directly, and it worked.**

**But jellyfin has no configuration for the RC mode,** so I cannot use the hardware acceleration。
My cpu is intel j5040

FFmpeg logs: from other issue, and mine like this.

             same error with intel N5095

```
/videos/39134344-9b47-fe93-b659-59cc2292c96d/hls1/main/0.ts

[...]

PR #16844, opened 2026-05-14
Fix VAAPI transcoding failure when only CQP rate control is supported

bybrooklyn:
**Changes**

This fixes a VAAPI hardware transcoding failure when the selected VAAPI encoder cannot use Jellyfin's default VBR rate control.

Some VAAPI drivers report only CQP support. Before this change, Jellyfin could still generate commands like `-rc_mode VBR`, which makes FFmpeg fail before writing any video frames:

```text
Driver does not support VBR RC mode (supported modes: CQP).
Error while opening encoder
Conversion failed!
```

The fix keeps the existing VAAPI VBR path when VBR is supported. When probing shows that VBR is unavailable but CQP works, Jellyfin now falls back to CQP and omits the bitrate-control arguments that do not belong with CQP (`-b:v`, `-maxrate`, and `-bufsize`).

This is intentionally limited to VAAPI rate-control selection. It does not change QSV, NVENC, AMF, decoding behavior, or add a new user-facing setting.

**Issues**

Relates to #11282.

**Testing**

- `dotnet test tests/Jellyfin.Controller.Tests/Jellyfin.Controller.Tests.csproj --configuration Release`
- `dotnet test tests/Jellyfin.MediaEncoding.Tests/Jellyfin.MediaEncoding.Tests.csproj --configuration Release`

I could not complete a local full-solution build because this checkout currently fails earlier with `CS9057`: `Jellyfin.CodeAnalysis` references compiler `5.3.0.0`, while the local compiler is `5.0.0.0`.

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -1576,8 +1576,55 @@ namespace MediaBrowser.Controller.MediaEncoding
             return ".ts";
         }
 
-        private string GetVideoBitrateParam(EncodingJobInfo state, string videoCodec)
+        private static bool IsVaapiVideoEncoder(string videoCodec)
         {
+            return string.Equals(videoCodec, "h264_vaapi", StringComparison.OrdinalIgnoreCase)
+                || string.Equals(videoCodec, "hevc_vaapi", StringComparison.OrdinalIgnoreCase)
+                || string.Equals(videoCodec, "av1_vaapi", StringComparison.OrdinalIgnoreCase);
+        }
+
+        private static int GetVaapiQp(EncodingOptions encodingOptions, string videoCodec)
+        {
+            var defaultQp = string.Equals(videoCodec, "h264_vaapi", StringComparison.OrdinalIgnoreCase) ? 23 : 28;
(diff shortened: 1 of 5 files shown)


bybrooklyn:
Just to clarify, I did use GPT-5.5 to help identify and fix the bug here. If that is a big no no, then please reject this. Thank you!

gnattu:
I think we have already answered why encoders only supports CQP cannot and *WILL NOT* be supported?

Just in your linked issue, you can get the answer, and in your linked issue, there is another PR that was closed for the exact same reason.

So I will say it again here:

Jellyfin requires CBR or VBR to limit bitrate reliably to both meet the client's expectation and network bandwidth constraints. CQP cannot do this and will not be supported.

nyanmisaka:
Intel could fix their kernel driver with just one line of code, but they don't want to. Therefore, you need to manually add a kernel parameter to set it correctly.

https://gitlab.freedesktop.org/drm/i915/kernel/-/work_items/8080

bybrooklyn:
> I think we have already answered why encoders only supports CQP cannot and _WILL NOT_ be supported?
> 
> Just in your linked issue, you can get the answer, and in your linked issue, there is another PR that was closed for the exact same reason.
> 
> So I will say it again here:
> 
> Jellyfin requires CBR or VBR to limit bitrate reliably to both meet the client's expectation and network bandwidth constraints. CQP cannot do this and will not be supported.

Thank you for your explaination

PR #16855, opened 2026-05-15
Fix integrated provider images

Shadowghost:
Since integrated providers don't have a repository entry and their manifests are generated on-the-fly they don't have images in the plugin view on web.

**Changes**
* Include provider images in the repo
* Allow Plugins to reference bundled resources


MediaBrowser.Providers/Plugins/ComicVine/Plugin.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// <summary>
+    /// Gets the current plugin instance.
+    /// </summary>
+    public static Plugin? Instance { get; private set; }
+
+    /// <inheritdoc />
+    public override Guid Id => new("3ade6fd1-c76c-4560-b2df-f6bca4c2332f");
+
+    /// <inheritdoc />
+    public override string Name => "Comic Vine";
```
dkanada: As I understand it, neither this one nor the Google Books plugin need to be created. The external link handling only exists to parse local metadata without installing the *actual* plugins from our official repositories. The next stable release will include external plugins for Google Books, Comic Vine, and OpenLibrary so the images should be added to the repository rather than the server.

Shadowghost: Sure, but we should not ship multiple of the same providers. If we have an external provider for Google Books and ComicVine, the link resolution should also exclusively live there.

dkanada: We don't ship multiple of the same providers - we ship them as external plugins, and the server can parse external IDs for Google Books and Comic Vine when they are provided from *any* metadata source. We also provide the same support for IMDb, IMVDb, and zap2it links.

JPVenson (approved):

MediaBrowser.Providers/Plugins/ComicVine/Plugin.cs
Shadowghost: Well yes but for IMDb and zap2it we don't have additional plugins. I'm not sure how we should handle it tbh.
Either we say the external resolution is part of the actual metadata plugin (if there is one) or we handle all the links independently in a secondary plugin that bundles the resolution.

dkanada (approved):
An improvement over the current situation. Long term I'd say this should be removed as soon as we can extract the embedded plugins.

Bond-009 (approved):

PR #16862, opened 2026-05-15
Modernize version display in UI

JPVenson:


Shadowghost (approved):

theguymadmax:
Starting with Jellyfin 12.0 for display purposes, we should use 2 digits (major.minor). As, webUI is doing: https://github.com/jellyfin/jellyfin-web/blob/f24eefc626e3b4e95638ca3c6447f97625619df8/src/utils/versions.ts#L31

theguymadmax (approved):

Bond-009 (approved):

PR #16864, opened 2026-05-15
feat: Add DVD/Blu-Ray playback title selection

DanTheMan827:
**Changes**
Adds the ability to select the title to play from a DVD or Blu-Ray image or unpacked directory structure.

This requires an ffmpeg binary with libdvdnav, and libbluray support.

**Known Issues**
Static streams of such media will be transcoded on the fly with codec copy to an .mkv file and sent.  This means no seeking of the media.

**Issues**

### DIFF
--- a/Jellyfin.Api/Controllers/ItemUpdateController.cs
+++ b/Jellyfin.Api/Controllers/ItemUpdateController.cs
@@ -244,7 +244,7 @@ public class ItemUpdateController : BaseJellyfinApiController
 
         if (item is not Video video)
         {
-            return BadRequest("Item is not a DVD/Blu-ray ISO or unpacked disc.");
+            return BadRequest(_localizationManager.GetLocalizedString("ItemNotDvdBluray"));
         }
 
         IsoType? effectiveIsoType = video.VideoType switch
@@ -278,7 +278,7 @@ public class ItemUpdateController : BaseJellyfinApiController
                 }
             }
 
-            return BadRequest("Item is not a DVD/Blu-ray ISO or unpacked disc.");
+            return BadRequest(_localizationManager.GetLocalizedString("ItemNotDvdBluray"));
         }
 
         var titles = _mediaEncoder.GetIsoTitles(video.Path, effectiveIsoType.Value);
--- a/Jellyfin.Api/Controllers/ItemUpdateController.cs
+++ b/Jellyfin.Api/Controllers/ItemUpdateController.cs
@@ -15,11 +15,13 @@ using MediaBrowser.Controller.Entities.Audio;
 using MediaBrowser.Controller.Entities.TV;
 using MediaBrowser.Controller.Library;
 using MediaBrowser.Controller.LiveTv;
+using MediaBrowser.Controller.MediaEncoding;
 using MediaBrowser.Controller.Providers;
 using MediaBrowser.Model.Dto;
 using MediaBrowser.Model.Entities;
 using MediaBrowser.Model.Globalization;
 using MediaBrowser.Model.IO;
+using MediaBrowser.Model.MediaInfo;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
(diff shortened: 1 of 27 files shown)


Jellyfin.Api/Controllers/ItemUpdateController.cs
```diff
@@ -196,6 +221,70 @@ public ActionResult<MetadataEditorInfo> GetMetadataEditorInfo([FromRoute, Requir
...
+    {
+        var item = _libraryManager.GetItemById<BaseItem>(itemId, User.GetUserId());
+        if (item is null)
+        {
+            return NotFound();
+        }
+
+        if (item is not Video video)
+        {
+            return BadRequest("Item is not a DVD/Blu-ray ISO or unpacked disc.");
```
JPVenson: dont use untranslated responses

MediaBrowser.MediaEncoding/Probing/DvdVideoProber.cs
```diff
@@ -0,0 +1,789 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Microsoft.Extensions.Logging;
+
+namespace MediaBrowser.MediaEncoding.Probing
```
JPVenson: use scoped namespace for new titles

MediaBrowser.MediaEncoding/Probing/DvdVideoProber.cs
```diff
@@ -0,0 +1,789 @@
...
+        /// Parses a DVD VIDEO_TS directory (on disk) and returns full disc information
+        /// including per-title durations resolved from the VTS IFO files.
+        /// </summary>
+        /// <param name="dvdPath">
+        /// Root DVD directory (may contain a VIDEO_TS sub-directory) or the VIDEO_TS
+        /// directory itself.
+        /// </param>
+        /// <param name="logger">Optional logger; warnings are emitted for missing/corrupt IFOs.</param>
+        /// <returns>Parsed disc info; titles list may be empty on failure.</returns>
+        internal static DvdDiscInfo ParseVideoTsDirectory(string dvdPath, ILogger? logger = null)
```
JPVenson: why would the logger be optional here?

MediaBrowser.MediaEncoding/Probing/DvdVideoProber.cs
```diff
@@ -0,0 +1,789 @@
...
+            return disc;
+        }
+
+        /// <summary>
+        /// Parses a DVD ISO image and returns full disc information including per-title durations.
+        /// </summary>
+        /// <param name="isoPath">Path to the ISO image file.</param>
+        /// <param name="logger">Optional logger.</param>
+        /// <returns>Parsed disc info; titles list may be empty on failure.</returns>
+        internal static DvdDiscInfo ParseVideoTsIso(string isoPath, ILogger? logger = null)
```
JPVenson: or here

JPVenson (changes requested):
Please test your own PR before submitting. This is missing database migrations.

Jellyfin.Api/Controllers/ItemUpdateController.cs
DanTheMan827: Could you please direct me where to add these? There are other `BadRequest` invocations that also use hard-coded strings.

DanTheMan827:
> Please test your own PR before submitting. This is missing database migrations.

In regards to database migrations, I don't believe I modified any of those models? The playback title is stored in the media data object.

JPVenson:
you added a new property IsoPlaybackTitle

DanTheMan827:
> you added a new property IsoPlaybackTitle

Yes, but that’s part of the “Data” key.  In the database, that’s stored as a json serialized string.

I did try to make a migration, but IsoPlaybackTitle is nowhere in the generated file, so that’s why it wasn’t included.

holow29:
I think this effectively switches back to ffmpeg/libbluray like #11833, though with concat fallback.

One thing that might be outside of this PR's scope but potential for future work: being able to assign playlists/titles a name and then store that in DB so it can be selected as a "version" - maybe using the existing version functionality. This would be helpful for discs that use seamless branching for different cuts.

DanTheMan827:
@holow29 I suppose so, but this also adds the ability to choose a specific title / playlist if the longest one isn’t correct, although that requires UI support unless there’s a way to provide custom metadata fields to clients that I’m not aware of.

PR #16869, opened 2026-05-18
Preserve external subtitles after strm probing

fabwa:
## What changed

- Preserve external subtitle streams before the forced remote probe used for `.strm` playback sources.
- Restore those preserved external subtitles after the media sources are rebuilt from the probed stream metadata.
- Avoid duplicate restored subtitle paths and reindex restored streams after the existing probed stream indexes.

## Why

Remote `.strm` items, especially HLS URLs, are probed again during `PlaybackInfo` generation. That probe rebuilds media streams from the remote source and can drop already-discovered local sidecar subtitles. The result is that clients receive audio/video streams but no subtitle streams, so the subtitle selector disappears even though the library item has an external subtitle.

This keeps the probed audio/video streams while preserving local external subtitles that were attached to the original media source.

## Validation

- `dotnet build Emby.Server.Implementations/Emby.Server.Implementations.csproj -c Release -p:TreatWarningsAsErrors=false`
- `dotnet test tests/Jellyfin.Server.Implementations.Tests/Jellyfin.Server.Implementations.Tests.csproj -c Release --no-restore`
- Manual verification on a Jellyfin 10.11.8 host with a remote HLS `.strm`: before the equivalent patch, `PlaybackInfo` returned only audio/video streams; after the patch it returned the external subtitle stream and the subtitle `Stream.vtt` endpoint returned `200 OK`.

### DIFF
--- a/Emby.Server.Implementations/Library/MediaSourceManager.cs
+++ b/Emby.Server.Implementations/Library/MediaSourceManager.cs
@@ -171,6 +171,7 @@ namespace Emby.Server.Implementations.Library
         public async Task<IReadOnlyList<MediaSourceInfo>> GetPlaybackMediaSources(BaseItem item, User user, bool allowMediaProbe, bool enablePathSubstitution, CancellationToken cancellationToken)
         {
             var mediaSources = GetStaticMediaSources(item, enablePathSubstitution, user);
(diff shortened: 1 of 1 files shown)


Shadowghost:
You literally wrote `[codex]` in the PR title, please check our [LLM policy](https://jellyfin.org/docs/general/contributing/llm-policies/).

PR #16884, opened 2026-05-20
Move episode path parsing to Rust interop

yvesoddly:
- Build and ship a Rust `jellyfin_naming` native library from the C# project
- Route `EpisodePathParser` through JSON-based Rust interop
- Preserve the existing parser behavior in the Rust implementation

**Changes**

**Issues**


nielsvanvelzen:
what the slop

PR #16907, opened 2026-05-22
Add library option to prefer original title over localized name

JonathanVil:
**Changes**
Adds a `PreferOriginalTitle` option to `LibraryOptions`. When enabled, item names returned by the API will use `OriginalTitle` instead of the localized `Name`. If `OriginalTitle` is absent, it will just use the regular `Name`.
This allows users who prefer seeing titles in their original language (e.g. Swedish titles instead of English translations) to configure this per library.

**Issues**
This addresses https://features.jellyfin.org/posts/32/keep-original-title-option


JonathanVil:
Closing this at it is almost identical to #16435

PR #16914, opened 2026-05-23
Live TV: re-fetch channel icons on guide refresh

danieltutuianu:
## Context

Ran into Live TV channel logos getting stuck after the first import — `Refresh Guide` skips them once a primary image exists, even when the EPG/tuner icon changed (or the URL stayed the same but the file on the server was updated).

In my setup icons come from XMLTV with remote picon URLs; once they were wrong or stale, they stayed that way forever.

## Change

- re-apply icon path/url on every guide refresh (not only when `!HasImage`)
- small `LiveTvChannelImageHelper`, used from `GuideManager` and `ChannelManager`
- a few unit tests

## Test plan

- [x] `dotnet test tests/Jellyfin.LiveTv.Tests --filter LiveTvChannelImageHelperTests`
- [x] Live TV + XMLTV: run Refresh Guide and confirm channel icons update (including when the URL is unchanged but the picon on the server changed)


Shadowghost (approved):

Shadowghost:
@danieltutuianu can you please fix the merge conflicts

danieltutuianu:
> @danieltutuianu can you please fix the merge conflicts

done

Bond-009:
Is this still needed now that #10841 is merged?

danieltutuianu:
> Is this still needed now that #10841 is merged?

Yes, I think it’s still needed.

#10841 fixed the case where the channel icon URL changes (for example after remapping a channel). That was issue #7834.

My PR fixes a different case: the icon URL stays the same, but the image on the server changed. That happens a lot with XMLTV picons.

Right now, if the URL didn’t change, Jellyfin keeps the old cached icon and doesn’t download it again. This PR makes guide refresh re-download the icon even when the URL is the same.

So #10841 and this PR solve related but different problems.

crobibero (approved):

PR #16915, opened 2026-05-23
Extract attachments in one ffmpeg command when dumping

### LINKED ISSUE #16911: Fonts extraction takes longer than necessary

- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-web/issues?q=is%3Aissue) _(I've searched it)_.

### Describe the bug

I played a file with internal ASS subtitles and it worked flawlessly when my server was a new, powerful PC. But when I changed my server to a mini PC, the subtitles would take minutes to appear, so I went to check the issue in the log.

jellyfin-server extracts the fonts one by one from the video file, and considering this file had 49 fonts attached, it took a lot of time. I would like to suggest for you to change the font extraction command to this one:

ffmpeg -dump_attachment:t:0 "fontDir/font00.ttf" -dump_attachment:t:1 "fontDir/font01.otf" [.....] -dump_attachment:t:n "fontDir/fontNN.ttf" -i input.mkv

This command managed to extract the fonts in about 8 seconds, even in my old mini PC from a HDD connected through a USB 2.0 port. It took 60 seconds when they were extracted one by one.

### Reproduction Steps

1. Play any file with ASS subtitles and attached fonts.
2. Check the server log to follow how long the font extraction takes.
3. Better observed in files with dozens of fonts attached.

### Expected/Actual behaviour

I expected it to be as fast as it was before, but over more demanding conditions it was possible to notice the bottleneck at the font extraction step.

### Logs

_No response_

### Server version

10.11.8

### Web version

10.11.8

### Build version

10.11.8

### Platform

Tizen

### Browser

Firefox

### Additional information

_No response_

Shadowghost:
Instead of running multiple sequential ffmpeg runs per attachment, we just dump all of them at once.

**Changes**
Extract attachment with on FFmpeg run

**Issues**
Fixes #16911

### DIFF
--- a/MediaBrowser.MediaEncoding/Attachments/AttachmentExtractor.cs
+++ b/MediaBrowser.MediaEncoding/Attachments/AttachmentExtractor.cs
@@ -1,8 +1,10 @@
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using AsyncKeyedLock;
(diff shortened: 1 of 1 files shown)


nyanmisaka (approved):

Bond-009 (approved):
Maybe we should log exactly which attachments failed to extract?

Shadowghost:
@Bond-009 since we are doing it in batches, I don't see a way to know this except analyzing the output of ffmpeg

Bond-009:
@Shadowghost We check if everything is extracted as expected, so we can report the failures, no?
https://github.com/jellyfin/jellyfin/pull/16915/changes#diff-e81faefb46f008f2237bf279da536be253d255d1969e96224acc2dbb92eaee10R238

PR #16916, opened 2026-05-23
Speedup VirtualFolder modifications

Shadowghost:
When adding or removing virtual folders, we don't need to do a full-blown validation of all folders below the root folder, just the newly added one or the modified library will do.

**Changes**
* Refresh AggregatedFolder only on full scan
* Share refreshOptions and therefore DirectoryService and its cache between folder refreshs


JPVenson (approved):
needs testing

theguymadmax:
Breaks the built-in "Automatically add to collection" library feature.

Jellyfin.Api/Controllers/LibraryStructureController.cs
```diff
@@ -90,7 +90,7 @@ public async Task<ActionResult> AddVirtualFolder(
             libraryOptions.PathInfos = Array.ConvertAll(paths, i => new MediaPathInfo(i));
         }
 
-        await _libraryManager.AddVirtualFolder(name, collectionType, libraryOptions, refreshLibrary).ConfigureAwait(false);
+        _ = _libraryManager.AddVirtualFolder(name, collectionType, libraryOptions, refreshLibrary);
```
dfederm, line 93: Doesn't this make things racy? A caller will call this endpoint and get a 204 and then (reasonably) believes it can query the folder it added. After this change though the caller would need to wait an arbitrary amount of time.

Emby.Server.Implementations/Library/LibraryManager.cs
```diff
@@ -3508,19 +3512,37 @@ public async Task AddVirtualFolder(string name, CollectionTypeOptions? collectio
...
-                }
-                else
-                {
-                    // Need to add a delay here or directory watchers may still pick up the changes
-                    await Task.Delay(1000).ConfigureAwait(false);
-                    LibraryMonitor.Start();
-                }
+                    try
+                    {
+                        await ValidateTopLibraryFolders(CancellationToken.None).ConfigureAwait(false);
```
dfederm, line 3525: If this throws, the catch below just logs and swallows. `LibraryMonitor` then stays stopped for the rest of the process lifetime and there's no filesystem watching until the server restarts.

Consider moving the restart into a `finally` so that it always fires.

Jellyfin.Api/Controllers/LibraryStructureController.cs
dfederm: Also if things end up ultimately failing, the API still returns success.

PR #16928, opened 2026-05-25
Skip video copy-preference ranking when re-encoding is required

kontell:
`GetVideoTranscodeProfile()` ranks transcoding profiles by whether the source video codec can be stream-copied. This makes sense when the source needs remuxing only. 

But when the video stream must be re-encoded (bitrate limit, etc.), the ranking still favours profiles containing the source codec. 

This prevents a more efficient video codec from being selected when it's in a different profile to the source codec.

The fix is to check the already-computed transcode reasons before applying copy-preference ranking. When `VideoReasons` or `ContainerBitrateExceedsLimit` are present, all profiles receive equal video rank and the client's profile ordering determines the target codec. When the video can actually be copied, the existing copy-preference ranking is preserved.

One scenario where this doesn't work is with SubtitleCodecNotSupported, don't see a way to drop the copy-preference ranking for burning in subtitles only.

### DIFF
--- a/MediaBrowser.Model/Dlna/StreamBuilder.cs
+++ b/MediaBrowser.Model/Dlna/StreamBuilder.cs
@@ -859,6 +859,10 @@ namespace MediaBrowser.Model.Dlna
             var videoCodec = videoStream?.Codec;
             var audioCodec = audioStream?.Codec;
 
+            // When the video stream must be re-encoded (not just remuxed), skip the copy-preference
+            // ranking so that client profile ordering determines the target codec.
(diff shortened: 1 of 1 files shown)


gnattu (changes requested):
TranscodeReason is a bad way to detect it because the real playback is way more complicated than this.

Actually with current API even the server does not know if the video is going to be copied or transcoded during streambuilder phase, at this phase, it only knows if the video can be directly played or not.

Direct play here means no any kind of server processing and ffmpeg will not even be called.

The server ONLY knows if the video can be copied after the HLS endpoint has being requested by the client and then the ffmepg helpers will decide which codec to use and can the stream be copied.

The issue you said is also a non-issue because the client now can explicitly specify a preferred transcoding codec.

PR #16936, opened 2026-05-26
Compare old file byte-by-byte to new stream

### LINKED ISSUE #12978: nfo saver makes unneccesary disk writes during library scan

IDisposable:
When computing the NFO contents, we will rewrite the file even if everything is unchanged. This causes the file updated time to change and increases load on the filesystem (and underlying drives), causing other consumers of the NFO files to force a rescan (e.g. ChannelsDVR).

Now does a byte-by-byte comparison and skips the write if the NFO is identitcal.

Also now passes the cancellationToken down to the stream copy so we abort faster if cancelled.

Fixes https://github.com/jellyfin/jellyfin/issues/12978

### DIFF
--- a/MediaBrowser.XbmcMetadata/Savers/BaseNfoSaver.cs
+++ b/MediaBrowser.XbmcMetadata/Savers/BaseNfoSaver.cs
@@ -208,7 +208,7 @@ namespace MediaBrowser.XbmcMetadata.Savers
             Directory.CreateDirectory(directory);
 
             // Compare byte-for-byte before proceeding.
-            if (File.Exists(path) && await IsFileIdenticalAsync(stream, path, cancellationToken).ConfigureAwait(false))
+            if (File.Exists(path) && await stream.IsFileIdenticalAsync(path, cancellationToken).ConfigureAwait(false))
             {
                 return; // Don't save since .nfo is unchanged.
             }
@@ -236,68 +236,6 @@ namespace MediaBrowser.XbmcMetadata.Savers
             if (ConfigurationManager.Configuration.SaveMetadataHidden)
             {
                 SetHidden(path, true);
-            }
-        }
-
-        private static async Task<bool> IsFileIdenticalAsync(Stream stream, string path, CancellationToken cancellationToken)
-        {
-            ArgumentNullException.ThrowIfNull(stream);
-            ArgumentException.ThrowIfNullOrEmpty(path);
-
-            if (!stream.CanSeek)
-            {
-                return false;
-            }
-
-            const int BufferSize = 81920;
-            var originalPosition = stream.Position;
-
-            try
-            {
-                stream.Position = 0;
-
-                using var existingFileStream = new FileStream(
-                    path,
-                    FileMode.Open,
-                    FileAccess.Read,
-                    FileShare.Read,
-                    bufferSize: BufferSize,
-                    FileOptions.Asynchronous);
-
-                if (existingFileStream.Length != stream.Length)
-                {
-                    return false;
-                }
-
-                var streamBuffer = new byte[BufferSize];
-                var existingBuffer = new byte[BufferSize];
-
-                while (true)
-                {
-                    cancellationToken.ThrowIfCancellationRequested();
-
-                    var streamBytesRead = await stream.ReadAsync(streamBuffer.AsMemory(), cancellationToken).ConfigureAwait(false);
-                    var existingBytesRead = await existingFileStream.ReadAsync(existingBuffer.AsMemory(), cancellationToken).ConfigureAwait(false);
-
-                    if (streamBytesRead != existingBytesRead)
-                    {
-                        return false;
-                    }
-
-                    if (streamBytesRead == 0)
-                    {
-                        return true;
-                    }
-
-                    if (!streamBuffer.AsSpan(0, streamBytesRead).SequenceEqual(existingBuffer.AsSpan(0, existingBytesRead)))
-                    {
-                        return false;
-                    }
-                }
-            }
-            finally
-            {
-                stream.Position = originalPosition;
             }
         }
 
--- a/MediaBrowser.XbmcMetadata/Savers/BaseNfoSaver.cs
+++ b/MediaBrowser.XbmcMetadata/Savers/BaseNfoSaver.cs
@@ -198,14 +198,22 @@ namespace MediaBrowser.XbmcMetadata.Savers
 
                 cancellationToken.ThrowIfCancellationRequested();
 
-                await SaveToFileAsync(memoryStream, path).ConfigureAwait(false);
+                await SaveToFileAsync(memoryStream, path, cancellationToken).ConfigureAwait(false);
             }
         }
 
-        private async Task SaveToFileAsync(Stream stream, string path)
+        private async Task SaveToFileAsync(Stream stream, string path, CancellationToken cancellationToken)
         {
             var directory = Path.GetDirectoryName(path) ?? throw new ArgumentException($"Provided path ({path}) is not valid.", nameof(path));
             Directory.CreateDirectory(directory);
+
+            // Compare byte-for-byte before proceeding.
+            if (File.Exists(path) && await IsFileIdenticalAsync(stream, path, cancellationToken).ConfigureAwait(false))
+            {
+                return; // Don't save since .nfo is unchanged.
+            }
+
+            stream.Position = 0;
 
             // On Windows, saving the file will fail if the file is hidden or readonly
             FileSystem.SetAttributes(path, false, false);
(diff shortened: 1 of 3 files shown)


Shadowghost (commented):
Wouldn't hashing the file with something like XXhash3 be simpler?

JPVenson (changes requested):

JPVenson (approved):

MediaBrowser.XbmcMetadata/Savers/BaseNfoSaver.cs
```diff
@@ -231,6 +239,68 @@ private async Task SaveToFileAsync(Stream stream, string path)
...
+                    bufferSize: BufferSize,
+                    FileOptions.Asynchronous);
+
+                if (existingFileStream.Length != stream.Length)
+                {
+                    return false;
+                }
+
+                var streamBuffer = new byte[BufferSize];
+                var existingBuffer = new byte[BufferSize];
```
Bond-009: You should rent these from the arraypool

MediaBrowser.XbmcMetadata/Savers/BaseNfoSaver.cs
```diff
@@ -231,6 +239,68 @@ private async Task SaveToFileAsync(Stream stream, string path)
             }
         }
 
+        private static async Task<bool> IsFileIdenticalAsync(Stream stream, string path, CancellationToken cancellationToken)
```
Bond-009: Should probably move this somewhere else as it's pretty generic
Maybe Jellyfin.Extensions?

Bond-009 (commented):
Another possible optimization is checking if the stream is a MemoryStream and just accessing the underlying buffer directly

MediaBrowser.XbmcMetadata/Savers/BaseNfoSaver.cs
IDisposable: Done

MediaBrowser.XbmcMetadata/Savers/BaseNfoSaver.cs
IDisposable: Done

IDisposable:
> Wouldn't hashing the file with something like XXhash3 be simpler?

It was, but @Bond-009 mentioned I should just do byte-by-byte... 
https://github.com/jellyfin/jellyfin/pull/13547#issuecomment-4423530912

IDisposable:
(rebased to get rid of the irrelevant "API changes")

IDisposable:
> Another possible optimization is checking if the stream is a MemoryStream and just accessing the underlying buffer directly

Done

src/Jellyfin.Extensions/StreamExtensions.cs
```diff
@@ -60,5 +65,143 @@ public static async IAsyncEnumerable<string> ReadAllLinesAsync(this TextReader r
...
+            }
+
+            if (a is MemoryStream ms_a)
+            {
+                var bufferA = ms_a.GetBuffer();
+
+                // Fast path A: if both streams are MemoryStreams, compare directly against each other
+                if (b is MemoryStream ms_b)
+                {
+                    return bufferA.AsSpan(0, (int)ms_a.Length).SequenceEqual(ms_b.GetBuffer().AsSpan(0, (int)ms_b.Length));
```
Bond-009: You should use `TryGetBuffer` as it can fail

IDisposable: Swapped to `TryGetBuffer()` and cleaned up `CanSeek` contract.

src/Jellyfin.Extensions/StreamExtensions.cs
```diff
@@ -12,6 +15,8 @@ namespace Jellyfin.Extensions
     /// </summary>
     public static class StreamExtensions
     {
+        private const int StreamComparisonBufferSize = 65536;
```
IDisposable: There's a `CopyToBufferSize` constant in _MediaBrowser.Model.IOContants_ that might be usable, but I don't think pulling in a dependency is worthwhile for that...

IDisposable:
Added tests and cleaned up the `CanSeek` requirement.

src/Jellyfin.Extensions/StreamExtensions.cs
```diff
@@ -60,5 +65,140 @@ public static async IAsyncEnumerable<string> ReadAllLinesAsync(this TextReader r
...
+                        if (bytesReadA == 0)
+                        {
+                            return true;
+                        }
+
+                        if (!memoryA.Span[..bytesReadA].SequenceEqual(memoryB.Span[..bytesReadB]))
+                        {
+                            return false;
+                        }
+                    }
```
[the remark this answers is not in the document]
IDisposable: Switched to  `Stream.ReadAtLeastAsync` to cover short reads and added tests.

src/Jellyfin.Extensions/StreamExtensions.cs
```diff
@@ -60,5 +65,140 @@ public static async IAsyncEnumerable<string> ReadAllLinesAsync(this TextReader r
...
+
+                        offset += bytesRead;
+                    }
+
+                    return offset == segmentA.Count;
+                }
+                finally
+                {
+                    ArrayPool<byte>.Shared.Return(bufferB);
+                }
```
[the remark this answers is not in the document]
IDisposable: Document the stream position expectations and seek to start if they are seekable.
Fixed fast-path tests to actually exercise the fast path.

IDisposable:
I think the bike shed is clean.

src/Jellyfin.Extensions/StreamExtensions.cs
```diff
@@ -60,5 +65,164 @@ public static async IAsyncEnumerable<string> ReadAllLinesAsync(this TextReader r
...
+            // If b is MemoryStream but a is not, swap them to enable fast path B
+            if (b is MemoryStream && a is not MemoryStream)
+            {
+                (a, b) = (b, a);
+            }
+
+            if (a is MemoryStream streamA && streamA.TryGetBuffer(out var segmentA))
+            {
+                // Fast path A: if both streams are MemoryStreams, compare directly against each other
+                if (b is MemoryStream streamB && streamB.TryGetBuffer(out var segmentB))
```
Bond-009: Technically, if both are MemoryStreams but only B has a public buffer we'll take the slow path.

Bond-009 (approved):
Sorry for sending you down this path

src/Jellyfin.Extensions/StreamExtensions.cs
IDisposable: Good call, I did one more pass to clean that part up.

- Use fast paths only if we can `TryGetBuffer` on `MemoryStream` using segment's `Array`.
- Reduce swap overhead for fast path B.
- Avoid multiple virtcalls by memoizing the `CanSeek`s.
- Overlap slow path stream async reads.

IDisposable:
> Sorry for sending you down this path

I love a nice shed!

src/Jellyfin.Extensions/StreamExtensions.cs
```diff
@@ -60,5 +65,172 @@ public static async IAsyncEnumerable<string> ReadAllLinesAsync(this TextReader r
...
+                b.Position = 0;
+            }
+
+            if (aCanSeek && bCanSeek && b.Length != a.Length)
+            {
+                return false;
+            }
+
+            // MemoryStreams only unlock a fast path if their underlying buffer is exposed via TryGetBuffer.
+            var segmentA = a is MemoryStream streamA && streamA.TryGetBuffer(out var bufA) ? bufA : default;
```
IDisposable, line 152: `streamA` is a `struct`, so `default` gives us an valid struct with a null `.Array` element (and 0 for `.Count`)

src/Jellyfin.Extensions/StreamExtensions.cs
```diff
@@ -60,5 +65,172 @@ public static async IAsyncEnumerable<string> ReadAllLinesAsync(this TextReader r
...
+                var bufferB = ArrayPool<byte>.Shared.Rent(StreamComparisonBufferSize);
+                try
+                {
+                    var memoryA = bufferA.AsMemory();
+                    var memoryB = bufferB.AsMemory();
+                    while (true)
+                    {
+                        cancellationToken.ThrowIfCancellationRequested();
+
+                        var taskA = a.ReadAtLeastAsync(memoryA, memoryA.Length, throwOnEndOfStream: false, cancellationToken).AsTask();
```
IDisposable, line 205: Need the `.AsTask()` so we convert the `ValueTask<int>` to something we can `await` twice.

src/Jellyfin.Extensions/StreamExtensions.cs
```diff
@@ -60,5 +65,172 @@ public static async IAsyncEnumerable<string> ReadAllLinesAsync(this TextReader r
...
+                        var taskB = b.ReadAtLeastAsync(memoryB, memoryB.Length, throwOnEndOfStream: false, cancellationToken).AsTask();
+                        await Task.WhenAll(taskA, taskB).ConfigureAwait(false);
```
IDisposable, line 207: We have to `await` both here to ensure we don't hit a race condition where we are still reading into the memory buffers after they get returned to the `ArrayPool` by the `finally`

src/Jellyfin.Extensions/StreamExtensions.cs
```diff
@@ -60,5 +65,172 @@ public static async IAsyncEnumerable<string> ReadAllLinesAsync(this TextReader r
...
+
+                        var bytesReadA = await taskA.ConfigureAwait(false);
```
IDisposable, line 209: The project bans `.Result`. After `Task.WhenAll`, so we have to use this syntax

Bond-009 (approved):

PR #16940, opened 2026-05-26
Extend studio plugin

Shadowghost:
Reorganizes the plugin after https://github.com/jellyfin/jellyfin-artwork/pull/12


nielsvanvelzen (changes requested):
We already have a plugin for the artwork repository, it's external like all new plugins should be.

https://github.com/jellyfin/jellyfin-plugin-artwork

PR #16946, opened 2026-05-27
Fix CA2007 warnings in InstallationManager

jonesy827:
Wraps the downloaded plugin stream in an explicit `await using` block with `ConfigureAwait(false)`, matching the pattern already used elsewhere (e.g. `LiveStreamHelper`). Also adds `ConfigureAwait(false)` to the `ZipFile.ExtractToDirectoryAsync` call.

No behavioural change — the stream is still disposed after extraction, just with the configured awaiter on the implicit `DisposeAsync`.

Verified locally: warning count for the file drops from 4 to 0, full solution builds clean, all tests pass, and `dotnet format --verify-no-changes` is clean. Smoke-tested by installing two plugins (TVHeadend, LDAP Authentication) from the live repo against a freshly-built server — both extracted to `plugins/` with no errors in the log.

Part of #2149

### DIFF
--- a/Emby.Server.Implementations/Updates/InstallationManager.cs
+++ b/Emby.Server.Implementations/Updates/InstallationManager.cs
@@ -527,42 +527,44 @@ namespace Emby.Server.Implementations.Updates
             using var response = await _httpClientFactory.CreateClient(NamedClient.Default)
                 .GetAsync(new Uri(package.SourceUrl), cancellationToken).ConfigureAwait(false);
             response.EnsureSuccessStatusCode();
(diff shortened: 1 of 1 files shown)


Bond-009 (approved):

PCEWLKR (approved):
Tested locally in a Docker development environment.

dotnet build succeeds
CA2007 warnings appear resolved
Jellyfin starts successfully
Plugin installation flow still works as expected

Reviewed the changed sections in InstallationManager; the stream lifetime and checksum/extraction flow appear preserved while adding ConfigureAwait(false) handling. Dont see any problems here :)

theguymadmax (approved):

Issue #16728, opened 2026-04-29
10.11.8 LiveTV transcoding filling up space


Esslingers:
### Description of the bug

Im using 2 HDHomerun tuners directly in Jellyfin with XMLTV guide data.  When I start playing a live stream on a TV the transcode directory will get filled up and eventually crash the machine.  What I'm seeing in the transcode folder are a few smaller .ts files, and m38u file and 1 large .ts file that just keeps growing. The only way that the large file goes away is if the client stops streaming.  Even with delete segments enabled or even disabled the directory still fills up. 

### Reproduction steps

1. HDHomerun in Jellyfin
2. play live stream and watch transcode directory
3. even with delete segments the directory fills up. 
4. Use either QSV or VAAPI transcoding

### What is the current _bug_ behavior?

1. HDHomerun in Jellyfin
2. play live stream and watch transcode directory
3. even with delete segments the directory fills up. 
4. Use either QSV or VAAPI transcoding

### What is the expected _correct_ behavior?

Transcoding creates several smaller files for the HLS segments instead of appending to a single file.

### Jellyfin Server version

10.11.8

### Specify the build version

10.11.8

### Environment

```markdown
- OS: CentOS9 Stream
- Linux Kernel:
- Virtualization: Docker
- Clients:
- Browser:
- FFmpeg Version: ffmpeg version 7.1.3-Jellyfin Copyright (c) 2000-2025 the FFmpeg developers
- Playback Method: Roku client
- Hardware Acceleration: VAAPi or QSV (Intel iGPU)
- CPU Model:  i5-8500
- GPU Model:  Intel UHD Graphics 630
- Plugins:
- Reverse Proxy: HA Proxy
- Base URL:
- Networking:
- Jellyfin Data Storage & Filesystem:  NVME
- Media Storage & Filesystem: NVME
- External Integrations:
```

### Jellyfin logs

```shell
[2026-04-29 10:58:17.416 -04:00] [INF] [33] Jellyfin.Api.Helpers.MediaInfoHelper: RemoteClientBitrateLimit: 10000000, RemoteIP: "10.10.10.192", IsInLocalNetwork: False
[2026-04-29 10:58:17.416 -04:00] [INF] [33] Jellyfin.LiveTv.LiveTvMediaSourceProvider: Opening channel stream from "Emby", external channel Id: "hdhr_15.1"
[...]

PR #16948, opened 2026-05-27
Clean live TV HLS transcode segments

lollinoo:
## Summary

- Start the HLS segment cleaner for infinite Live TV streams.
- Start the transcoding throttler for infinite HLS video streams from file or HTTP inputs.
- Keep the existing behavior for long finite video files and long finite HLS streams.

## What happens

Live TV media sources are normalized as infinite streams, and media probing clears `RunTimeTicks` when the source has no fixed duration. That means many Live TV HLS transcodes reach `TranscodeManager` with `MediaSource.IsInfiniteStream == true` and `RunTimeTicks == null`.

On current `upstream/master`, the HLS segment cleaner is only enabled when all of these are true:

- input protocol is file or HTTP
- input is video
- transcoding type is HLS
- `RunTimeTicks` exists and is at least five minutes

Infinite Live TV streams fail the final condition because they intentionally do not have a runtime. As a result, no `TranscodingSegmentCleaner` is started for those sessions.

HLS transcoding writes segment files continuously while FFmpeg is running. Without the segment cleaner, those files remain in the transcode directory for the lifetime of the transcode. If the transcode path is backed by RAM, or if several clients start independent HLS transcodes for the same Live TV channel, segment files can accumulate until the transcode storage fills up.

The throttler has a similar finite-duration guard on `upstream/master`: it only starts for file-based video with a runtime of at least five minutes. Infinite HLS Live TV transcodes therefore also miss the normal throttling path.

## Change details

This updates the transcode cleanup and throttling predicates to treat infinite HLS video streams as eligible even when `RunTimeTicks` is not available. The existing finite-stream checks remain in place, so short finite streams are still excluded.

## Impact

Live TV HLS sessions without a fixed timestamp/runtime now get the same ongoing segment cleanup behavior as long finite HLS sessions. This limits transcode directory growth during long-running Live TV playback and reduces the risk of exhausting RAM-backed transcode storage.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -231,7 +231,7 @@
  - [MSalman5230](https://github.com/MSalman5230)
  - [dwandw](https://github.com/dwandw)
  - [Lampan-git](https://github.com/Lampan-git)
-
+ - [Lollinoo](https://github.com/lollinoo)
 # Emby Contributors
 
  - [LukePulverenti](https://github.com/LukePulverenti)
--- a/MediaBrowser.MediaEncoding/Transcoding/TranscodeManager.cs
+++ b/MediaBrowser.MediaEncoding/Transcoding/TranscodeManager.cs
@@ -541,7 +541,7 @@ public sealed class TranscodeManager : ITranscodeManager, IDisposable
 
     private void StartThrottler(StreamState state, TranscodingJob transcodingJob)
     {
-        if (EnableThrottling(state)
+        if (ShouldStartThrottler(state)
             && (_mediaEncoder.IsPkeyPauseSupported
                 || _mediaEncoder.EncoderVersion <= _maxFFmpegCkeyPauseSupported))
         {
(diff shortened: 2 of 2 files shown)


nyanmisaka:
@gnattu I remember you commented on a similar PR years before. It seems that with the influx of users, people are starting to request https://github.com/jellyfin/jellyfin/issues/16728 this feature again. Plex does indeed automatically delete segments generated during transcoding of live tv.

gnattu:
Well infinite streams are different from finite stream and I am still against putting the segment cleaner designed for finite stream onto the infinite stream. Also this PR would not fix the issue you linked because for livetv we have a cursed streaming mode called shared http stream which is not a typical hls playback. Its transcode cache does not even cache HLS segments of the livetv source, instead it copies source into an mpegts file and stream to the clients from that file. This design is inherited from the emby time and it barely works. For HDHR sources the default mode is still the shared http stream, and that is what the OP of the linked issue is using.

Fixing livetv requires much more effort than just patching the segment cleaner unfortunately.

barrelltitor:
If LiveTV needs redone anyway it might be good to have a way to at least not fill up the disk until then. I know how it goes with "temporary solutions", but my disk keeps getting filled up so it would be good to see some sort of hotfix or suggested way of properly cleaning these up until a proper implementation is made

lollinoo:
Yeah, I meant it as a minimal hotfix that works, since Live TV is a feature I use a lot. The simplest thing to do was to fix this problem, which for now is the one that impacts use the most, as I've found myself with full RAM and the VM crashing many times. I'm already working on finding a solution and a better implementation, but it will obviously take time.

PR #16969, opened 2026-05-29
Music logic rework and skip metadata

jagp8:
**Changes**
Move PlayCount and LastPlayedDate updates for audio items from playback start to playback stop, so listening to the first 3 seconds of a song no longer mark it as played.

Add two new thresholds to the Playback - Continue settings (MinAudioResumePct, MaxAudioResumePct) used to determine if a song is considered not started, partially played, or fully played. Below the minimum threshold the song is considered as skipped instead of partially played.

 Add SkipCount and LastSkippedDate to mirror the existing PlayCount/LastPlayedDate pair for skips.

**Issues**
https://features.jellyfin.org/posts/3944/improve-music-category-with-more-data
https://github.com/jyourstone/jellyfin-smartlists-plugin/issues/410


jagp8:
The migration file was generated automatically with the dotnet ef migrations, don't really know how to do it 😅

Emby.Server.Implementations/Library/UserDataManager.cs
```diff
@@ -363,6 +380,23 @@ public bool UpdatePlayState(BaseItem item, UserItemData data, long? reportedPosi
                     data.Played = playedToCompletion = true;
                 }
             }
+            else if (positionTicks > 0 && hasRuntime && item is Audio)
```
dfederm, line 509: I believe this block is unreachable. Line 340 above (`positionTicks > 0 && hasRuntime && item is not AudioBook && item is not Book`) would already music and line 366 (`positionTicks > 0 && hasRuntime && item is AudioBook`) would already match audio books.

If I'm right about that, then `MaxAudioResumePct` isn't read anywhere else and `MinAudioResumePct`  would only impact the skip decision and not the "considered played" decision.

Emby.Server.Implementations/Session/SessionManager.cs
```diff
@@ -825,12 +826,16 @@ private void OnPlaybackStart(User user, BaseItem item)
         {
             var data = _userDataManager.GetUserData(user, item);
 
-            data.PlayCount++;
-            data.LastPlayedDate = DateTime.UtcNow;
-
-            if (item.SupportsPlayedStatus && !item.SupportsPositionTicksResume)
+            // We won't manage Audio here but on OnPlaybackStopped
+            if (item is not Audio)
```
dfederm: `AudioBook : Audio` so audiobooks no longer get `PlayCount++`/`LastPlayedDate` at start

Emby.Server.Implementations/Session/SessionManager.cs
```diff
@@ -1148,6 +1153,26 @@ private bool OnPlaybackStopped(User user, BaseItem item, long? positionTicks, bo
                 playedToCompletion = true;
             }
 
+            // For Audio: PlayCount/Date and SkipCount/Date are managed here at stop time rather than at start time.
+            if (item is Audio && positionTicks.HasValue)
```
dfederm: Similarly since `AudioBook : Audio`, audiobooks now get `SkipCount++` when paused below `MinAudioResumePct` (default 10%), so pausing a 10 hour book at 30 min would be recorded as a skip.

dfederm:
Please consider adding tests

Emby.Server.Implementations/Library/UserDataManager.cs
jagp8: True, I forgot to add && item is not Audio on 340

Emby.Server.Implementations/Session/SessionManager.cs
jagp8: Ahhhhh, ok I'll look for a way to exclude Audibooks here

Emby.Server.Implementations/Session/SessionManager.cs
jagp8: Yep, same as the previous one I'll check!

jagp8:
after screwing it a little I think I've done it

Emby.Server.Implementations/Library/UserDataManager.cs
jagp8: Should be fixed

Emby.Server.Implementations/Session/SessionManager.cs
jagp8: Should be fixed

Emby.Server.Implementations/Session/SessionManager.cs
jagp8: Should be fixed

jagp8:
> Please consider adding tests

I've done my best with the tests, but it was more difficult than copy-paste PlayCount and LastPlayedDate 😅

Shadowghost (approved):

src/Jellyfin.Database/Jellyfin.Database.Implementations/ModelConfiguration/UserDataConfiguration.cs
```diff
@@ -20,6 +20,12 @@ public void Configure(EntityTypeBuilder<UserData> builder)
         builder.HasIndex(d => new { d.UserId, d.ItemId, d.LastPlayedDate });
         builder.HasIndex(d => new { d.UserId, d.Played, d.ItemId });
         builder.HasIndex(d => new { d.UserId, d.IsFavorite, d.ItemId });
+        builder.HasIndex(d => new { d.ItemId, d.UserId, d.SkipCount });
+        builder.HasIndex(d => new { d.UserId, d.SkipCount, d.ItemId });
+        builder.HasIndex(d => new { d.ItemId, d.UserId, d.LastSkippedDate });
+        builder.HasIndex(d => new { d.UserId, d.ItemId, d.LastSkippedDate });
+        builder.HasIndex(d => new { d.ItemId, d.UserId, d.PartiallyPlayed });
+        builder.HasIndex(d => new { d.UserId, d.PartiallyPlayed, d.ItemId });
```
cvium: This is excessive. Every insert and delete needs to update the indexes. You need to look at the queries and make an index that makes sense. And that's assuming we even need another index.

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/Migrations/20260604022501_AddIsPartiallyPlayed.cs
```diff
@@ -0,0 +1,29 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+#nullable disable
+
+namespace Jellyfin.Server.Implementations.Migrations
+{
+    /// <inheritdoc />
+    public partial class AddIsPartiallyPlayed : Migration
```
JPVenson: can you aggregate those migrations pls? no need to have two

JPVenson (changes requested):

src/Jellyfin.Database/Jellyfin.Database.Implementations/ModelConfiguration/UserDataConfiguration.cs
jagp8: Done

src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/Migrations/20260604022501_AddIsPartiallyPlayed.cs
jagp8: Done

jagp8: I deleted both and now are in AddAudioSkipAndPartialPlayTracking

JPVenson (approved):
Database LGTM 

However please keep in mind that this will not make the cut for 12.0 as we are already in the feature freeze for that.

jagp8:
> Database LGTM
> 
> However please keep in mind that this will not make the cut for 12.0 as we are already in the feature freeze for that.

No problem, better late thannever 😂

Do I have to monitor or do something else to the database?

src/Jellyfin.Database/Jellyfin.Database.Implementations/ModelConfiguration/UserDataConfiguration.cs
```diff
@@ -20,6 +20,8 @@ public void Configure(EntityTypeBuilder<UserData> builder)
         builder.HasIndex(d => new { d.UserId, d.ItemId, d.LastPlayedDate });
         builder.HasIndex(d => new { d.UserId, d.Played, d.ItemId });
         builder.HasIndex(d => new { d.UserId, d.IsFavorite, d.ItemId });
+        builder.HasIndex(d => new { d.ItemId, d.UserId, d.SkipCount });
+        builder.HasIndex(d => new { d.UserId, d.SkipCount, d.ItemId });
```
cvium: I don't see any queries on `SkipCount`. Are you sure these indexes are necessary?

jagp8: Hmmmm not really, I'll delete them

PR #9787, opened 2023-05-18
New filters for audio and subtitle languages

TheMelmacian:
**Changes**
New filter values allow to filter movies and series by audio and/or subtitle languages.
Language tags are extracted from the MediaStreams and provided by the FilterController.
To support filtering of TV-Series based on languages (which are linked to the Episodes and not the show itself) I created a sub-query that determines all linked media streams for an ItemId. The sub-query also works for alternate versions of movies.

I also changed the type of TypedBaseItems.OwnerId from TEXT to GUID and added new Indexes.
This was necessary because comparing GUID to TEXT didn't use the new index.
I used `EXPLAIN QUERY PLAN` on the subquery to check index usage.

I also prepared the frontend part for jellyfin-web: https://github.com/jellyfin/jellyfin-web/pull/4684

Solves the following feature requests:
- https://github.com/jellyfin/jellyfin/issues/617
- https://features.jellyfin.org/posts/39/filter-videos-by-audio-stream-language
- https://features.jellyfin.org/posts/1802/filter-content-by-subtitle-language
- https://features.jellyfin.org/posts/690/be-able-to-search-and-filter-by-language-movies-series

---

- The sub-query should also solve some of the filter issues where only the primary version of a movie with alternate versions was considered: (see: https://github.com/jellyfin/jellyfin/issues/6609 and linked/duplicate issues)

Issue #16859, opened 2026-05-15
[12.0] ISO-639-2 language labels create noisy and confusing track lists


854562:
### Description of the bug

A change for 12.0 (probably https://github.com/jellyfin/jellyfin/pull/15947) changed how Jellyfin sources language names for subtitles and audio tracks (it uses [iso6392.txt](https://github.com/jellyfin/jellyfin/blob/master/Emby.Server.Implementations/Localization/iso6392.txt) now). For several languages, this has introduced cluttered and, in some cases, confusing labels in the user interface.

Some examples:

- Visual clutter, such as `Romanian; Moldavian; Moldovan` or `Greek, Modern (1453-)`. The linguistic history of the Greek language is of very little relevance to a media server UI.
- False precision, such as `Dutch; Flemish`. To the user, this could imply a regional variant which is not present in the majority of cases. It would be akin to labelling all English tracks as `English; Australian`.
-  Contradictory labels when using embedded tracks that use language code `spa` for multiple tracks, especially if they use custom track titles for regional variants: `Latin American - Spanish; Castilian - SUBRIP` is now a common occurrence. 

### Reproduction steps

1. Open a client connected to a 12.0/unstable Jellyfin server.
2. Select a media item that contains subtitle or audio tracks for languages with the offending ISO-639-2 descriptors (e.g., Dutch, Spanish, Greek, or Romanian).
3. Observe the labels provided for these tracks.

### What is the current _bug_ behavior?

The UI displays the full string from iso6392.txt. This creates an unpolished (at best) and confusing (at worst) experience for the end user. See the examples above. It creates an inconsistency between English, French, and Portuguese on one side (which look "correct") and Greek, Dutch, and Spanish (among others) on the other.

### What is the expected _correct_ behavior?

Jellyfin should display more "friendly" language names, to make behaviour more consistent with (for example) English, French, and Portuguese, which are displayed as one would expect (with simply their primary name without any qualifiers).

Truncating the language names at the first delimiter (`,` or `;`)  would be a potential solution:

`Dutch; Flemish` would become `Dutch`
`Greek, Modern (1453-)` would become `Greek`
`Spanish; Castilian` would become `Spanish`
`Chinese (Simplified)` would remain `Chinese (Simplified)`, as intended by the linked PR.

But ideally, the language names would be properly localised to the interface language, rather than relying on raw unpolished English strings in a text file.

### Jellyfin Server version

Unstable

### Specify unstable release number

2026051106

[...]

PR #16980, opened 2026-05-30
Improve language filters to only fetch language codes that match the requested items/libraries (follow up to #9787)

TheMelmacian:
#9787 introduced language filters for audio and subtitles. One downside of the original implementation is that all language codes are fetched from the database when building the filter values, regardless of the requested item type or library.
I adapted the code so that the language codes are pre filtered based on the requested library and item types, similar to the genre filter values.

**Changes**
- only return language codes for the requested library and item types
-  enable language filters for season and episode views

### DIFF
--- a/Jellyfin.Api/Controllers/FilterController.cs
+++ b/Jellyfin.Api/Controllers/FilterController.cs
@@ -161,7 +161,7 @@ public class FilterController : BaseJellyfinApiController
         var streamLanguageQuery = new InternalItemsQuery(user)
         {
             // It's possible that different langauges are only available on alternative versions.
-            // To fetch them all, owned items are inlcluded.
+            // To fetch them all, owned items are included.
             IncludeOwnedItems = true,
             IncludeItemTypes = includeItemTypes,
             DtoOptions = new DtoOptions
@@ -194,7 +194,7 @@ public class FilterController : BaseJellyfinApiController
             && !includeItemTypes.Contains(BaseItemKind.Episode))
         {
             // streams are joined on epsiodes not shows or seasons
-            streamLanguageQuery.IncludeItemTypes = [..includeItemTypes, BaseItemKind.Episode];
+            streamLanguageQuery.IncludeItemTypes = [.. includeItemTypes, BaseItemKind.Episode];
         }
 
         if (includeItemTypes.Length == 1
--- a/Jellyfin.Api/Controllers/FilterController.cs
+++ b/Jellyfin.Api/Controllers/FilterController.cs
@@ -158,13 +158,42 @@ public class FilterController : BaseJellyfinApiController
             IsSeries = isSeries
         };
 
+        var streamLanguageQuery = new InternalItemsQuery(user)
+        {
+            // It's possible that different langauges are only available on alternative versions.
+            // To fetch them all, owned items are inlcluded.
+            IncludeOwnedItems = true,
+            IncludeItemTypes = includeItemTypes,
+            DtoOptions = new DtoOptions
+            {
+                Fields = Array.Empty<ItemFields>(),
+                EnableImages = false,
+                EnableUserData = false
+            },
+            IsAiring = isAiring,
+            IsMovie = isMovie,
+            IsSports = isSports,
+            IsKids = isKids,
+            IsNews = isNews,
+            IsSeries = isSeries
+        };
+
         if ((recursive ?? true) || parentItem is UserView || parentItem is ICollectionFolder)
         {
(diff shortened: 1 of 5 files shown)


enter-a-random-username:
This improves the search a lot.

Currently, language tags form music are shown in movie library and so on. Or same with long list of subtitle languages and after activating the filter, no result are left, because it is from another library.

Shadowghost (approved):

theguymadmax (approved):

Jellyfin.Api/Controllers/FilterController.cs
```diff
@@ -158,13 +158,42 @@ public ActionResult<QueryFilters> GetQueryFilters(
...
+            IsSports = isSports,
+            IsKids = isKids,
+            IsNews = isNews,
+            IsSeries = isSeries
+        };
+
         if ((recursive ?? true) || parentItem is UserView || parentItem is ICollectionFolder)
         {
             genreQuery.AncestorIds = parentItem is null ? Array.Empty<Guid>() : new[] { parentItem.Id };
+            streamLanguageQuery.AncestorIds = parentItem is null ? Array.Empty<Guid>() : new[] { parentItem.Id };
```
IDisposable: These could use a local instead of duplicate code

IDisposable (approved):

theguymadmax:
@TheMelmacian you have some formatting issues.

TheMelmacian:
fixed

Jellyfin.Server.Implementations/Item/BaseItemRepository.ByName.cs
```diff
@@ -81,6 +82,46 @@ public IReadOnlyList<string> GetGenreNames()
...
 
+    /// <inheritdoc />
+    public IReadOnlyList<string> GetMediaStreamLanguages(InternalItemsQuery filter, MediaStreamType mediaStreamType)
+    {
+        ArgumentNullException.ThrowIfNull(filter);
+
+        if (!filter.Limit.HasValue)
+        {
+            filter.EnableTotalRecordCount = false;
+        }
```
crobibero: EnableTotalRecordCount is never used, this should be removed

theguymadmax:
@TheMelmacian, can you address the feedback.

enter-a-random-username:
BTW https://github.com/jellyfin/jellyfin/issues/16859 adds issues for the filters. Now `enm` filters for `en`, because ti truncates the language. Both are visible.

theguymadmax (approved):

TheMelmacian:
> BTW #16859 adds issues for the filters. Now `enm` filters for `en`, because ti truncates the language. Both are visible.

This should only affect the displayed language and not the ISO code itself.
The code is coming directly from the database and is send unchanged back to the server when searching for media.
So instead of `English, Middle (1100-1500) (enm)` you will only see `English (enm)`.
While not ideal, it's still distinguishable from `English (en)` or `English (eng)` and doesn't affect the filter functionality.

The bigger "problem" is that ffmpeg currently only reads the old Language field from mkv files, instead of the newer LanguageBCP47 field, which can combine language, script and region information.
For example
- `por-BR` = "Portuguese, Brazil" or "Brazilian Portuguese" becomes `por`
- `zh-Hant` = "Chinese with Traditional Characters" becomes `zh`

So we currently can't distinguish between `por-BR` and `por-PT`, or `spa-ES` and `spa-419`, etc., even if it is specified in the metadata of the matroska container.

enter-a-random-username:
> > BTW #16859 adds issues for the filters. Now `enm` filters for `en`, because ti truncates the language. Both are visible.
> 
> This should only affect the displayed language and not the ISO code itself. The code is coming directly from the database and is send unchanged back to the server when searching for media. So instead of `English, Middle (1100-1500) (enm)` you will only see `English (enm)`. While not ideal, it's still distinguishable from `English (en)` or `English (eng)` and doesn't affect the filter functionality.
> 
> The bigger "problem" is that ffmpeg currently only reads the old Language field from mkv files, instead of the newer LanguageBCP47 field, which can combine language, script and region information. For example
> 
>     * `por-BR` = "Portuguese, Brazil" or "Brazilian Portuguese" becomes `por`
> 
>     * `zh-Hant` = "Chinese with Traditional Characters" becomes `zh`
> 
> 
> So we currently can't distinguish between `por-BR` and `por-PT`, or `spa-ES` and `spa-419`, etc., even if it is specified in the metadata of the matroska container.

Yeah, just noticed the enm filter is now useless, because it shows en results. Pretty sure I tested that before and always worked. So it is not the change from this PR

About other codes - Old/specific languages and not regional stuff
image

crobibero (approved):

PR #16981, opened 2026-05-30
Refactor RemoveVirtualFolder not found handling

PCEWLKR:
Refactors RemoveVirtualFolder to use an explicit check for the target folder's existence instead of relying on a FileNotFoundException to return a 404.

This addresses the existing TODO in LibraryStructureController.cs and keeps the 404 handling at the controller level instead of relying on manager-level exception handling.

Testing

dotnet build -p:AllowMissingPrunePackageData=true
dotnet test -p:AllowMissingPrunePackageData=true
Verified deleting an existing virtual folder returns 204
Verified deleting a nonexistent virtual folder returns 404
Verified /Library/VirtualFolders still returns valid data

### DIFF
--- a/Jellyfin.Api/Controllers/LibraryStructureController.cs
+++ b/Jellyfin.Api/Controllers/LibraryStructureController.cs
@@ -115,15 +115,15 @@ public class LibraryStructureController : BaseJellyfinApiController
             throw new ArgumentNullException(nameof(name));
         }
 
-        var virtualFolders = _libraryManager.GetVirtualFolders(true);
+        var virtualFolder = _libraryManager.GetVirtualFolders(true)
+            .FirstOrDefault(folder => string.Equals(folder.Name, name, StringComparison.Ordinal));
 
-        if (!virtualFolders.Any(folder =>
-                    string.Equals(folder.Name, name, StringComparison.OrdinalIgnoreCase)))
+        if (virtualFolder is null)
         {
             return NotFound("The media library does not exist.");
         }
 
-        await _libraryManager.RemoveVirtualFolder(name, refreshLibrary).ConfigureAwait(false);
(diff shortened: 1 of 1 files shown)


PCEWLKR:
Updated the implementation to avoid path-based filesystem validation from the request input. The existence check now uses the configured virtual folder list from ILibraryManager instead.

I originally modeled the implementation after RenameVirtualFolder, which appears to use a similar path-construction pattern. This change highlighted a few nearby cases that may warrant separate follow-up review, but I'm keeping the scope of this PR limited to RemoveVirtualFolder.

Jellyfin.Api/Controllers/LibraryStructureController.cs
```diff
@@ -105,12 +105,25 @@ public async Task<ActionResult> AddVirtualFolder(
...
         [FromQuery] bool refreshLibrary = false)
     {
-        // TODO: refactor! this relies on an FileNotFound exception to return NotFound when attempting to remove a library that does not exist.
-        await _libraryManager.RemoveVirtualFolder(name, refreshLibrary).ConfigureAwait(false);
+        if (string.IsNullOrWhiteSpace(name))
+        {
+            throw new ArgumentNullException(nameof(name));
+        }
+
+        var virtualFolder = _libraryManager.GetVirtualFolders(true)
```
talha4420, line 118: potential nullpointer from _libraryManager.GetVirtualFolders(true)

Jellyfin.Api/Controllers/LibraryStructureController.cs
```diff
@@ -105,12 +105,25 @@ public async Task<ActionResult> AddVirtualFolder(
...
+            .FirstOrDefault(folder => string.Equals(folder.Name, name, StringComparison.Ordinal));
```
talha4420, line 119: same for folder

talha4420 (commented):
nice to have api testing

PR #16995, opened 2026-06-01
Fix season unknown for flat TV structures

theguymadmax:
In the current master, there is a regression in which episodes in a flat folder structure with proper naming structure gets detected as season unknown. 
```
Show name (year)\
   S01E01.mkv
   S01E02.mkv
```
**Changes**
Updated `NeedsVirtualSeason()` to evaluate unlinked episodes based on their folder location rather than skipping them. 

**Issues**
None reported

### DIFF
--- a/MediaBrowser.Providers/TV/SeriesMetadataService.cs
+++ b/MediaBrowser.Providers/TV/SeriesMetadataService.cs
@@ -210,16 +210,19 @@ public class SeriesMetadataService : MetadataService<Series, SeriesInfo>
             return true;
         }
 
-        // Not yet processed
-        if (episode.SeasonId.IsEmpty())
+        // Episode has been processed and linked to a season, only needs a virtual season
+        // if it isn't already linked to a known physical season by ID or path
+        if (!episode.SeasonId.IsEmpty())
         {
-            return false;
+            return !physicalSeasonIds.Contains(episode.SeasonId)
+                && !physicalSeasonPaths.Contains(System.IO.Path.GetDirectoryName(episode.Path) ?? string.Empty);
         }
 
-        // Episode has been processed, only needs a virtual season if it isn't
-        // already linked to a known physical season by ID or path
-        return !physicalSeasonIds.Contains(episode.SeasonId)
-            && !physicalSeasonPaths.Contains(System.IO.Path.GetDirectoryName(episode.Path) ?? string.Empty);
+        // Episode not yet linked, check if it's in a physical season folder
+        // If yes then skip it, processing not finished
+        // If no then include it, needs Season Unknown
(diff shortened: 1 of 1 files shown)


Shadowghost (approved):

IDisposable (approved):

MediaBrowser.Providers/TV/SeriesMetadataService.cs
```diff
@@ -210,16 +210,19 @@ private static bool NeedsVirtualSeason(Episode episode, HashSet<Guid> physicalSe
             return true;
         }
 
-        // Not yet processed
-        if (episode.SeasonId.IsEmpty())
+        // Episode has been processed and linked to a season, only needs a virtual season
```
IDisposable, line 213: How about?

```csharp
        // Episode has been processed and linked to a season, does not need a virtual season
        // if it is already linked to a known physical season by ID
        if (!episode.SeasonId.IsEmpty() && physicalSeasonIds.Contains(episode.SeasonId))
        {
            return false;
        }

        // Episode not yet linked to a physical season ID, check if it's in a physical season folder
        // If yes then skip it, processing not finished
        // If no then include it, needs Season Unknown
        var episodeDirectory = System.IO.Path.GetDirectoryName(episode.Path) ?? string.Empty;
        return !physicalSeasonPaths.Contains(episodeDirectory);
```

theguymadmax: I would prefer keeping the cases explicit.

jhparker88:
> In the current master, there is a regression in which episodes in a flat folder structure

@theguymadmax Do you think it would make sense to add a test for this particular case to avoid another regression in future?

MediaBrowser.Providers/TV/SeriesMetadataService.cs
IDisposable: Understood... I just felt it was more explicit to do the early `false` return when we _know_ we have a `SeasonId`, and it's linked to an actual _physical season_.

If that isn't the case, then we would have to then request a virtual season only if we don't have one...

theguymadmax:
> > In the current master, there is a regression in which episodes in a flat folder structure
> 
> @theguymadmax Do you think it would make sense to add a test for this particular case to avoid another regression in future?

Yes and No. The work on the [scanner rewrite](https://github.com/jellyfin/jellyfin-meta/discussions/125) is set to start once we release 12.0 RC1, which we are trying to get out this weekend - why you might have noticed a bunch of PR's being merged :-) 

Since that's right around the corner, I feel like anything now would end up being rewritten or just hinder the rewrite. New tests will be included in the rewrite.

Bond-009 (approved):
Not sure why you had to invert the branches, but otherwise OK

PR #16856, opened 2026-05-15
Fix movie recommendations

Shadowghost:
Add a batch path to the local similar-items pipeline so movie recommendations can score and resolve candidates for several source items in one pass instead of N independent queries.

**Changes**
* Restore weighted recommendations for movies

**Issues**
Fixes: #14088

Issue #15342, opened 2025-11-01
Random sorting in GetSimilarItems (PR #14918) breaks recommendation accuracy in "More Like This"


Guamaboy:
## Description of the bug
After upgrading to Jellyfin 10.11.1, the "More Like This" recommendations have become completely inaccurate. Previously, this feature provided relevant suggestions based on genre, tags, and other metadata. Now the recommendations appear random and unrelated to the source content. Probably because it randomizes all of the recommendations not just the first 16 that are displayed

Upon investigation, this appears to be caused by the random sorting feature added to GetSimilarItems in PR #14918.

## Impact
This makes the "More Like This" feature essentially unusable for content discovery, as users cannot rely on it to find related content. The recommendations are now less accurate than having no recommendations at all.
Believe it or not, multiple users use this as a search functionality to look for movies they may not know the name of but know of a show its related to. Leaving that capability to a randomizer breaks its consistency and reliability.

## Suggested fix
Add a configuration option to disable random sorting in GetSimilarItems, or modify the algorithm to maintain similarity-based ordering while only randomizing among items of equal similarity scores. Or maybe randomizing the first 20 - 30 items. Only 16 are ever shown. This way the randomness is still a bit similar. Or maybe only randomize the trailing 6 - 10 items shown that way the first items are always similar. etc

### Reproduction steps

1. Upgrade to Jellyfin 10.11.0 or 10.11.1 from 10.10.7
2. Navigate to any movie or TV show detail page
3. Scroll down to the "More Like This" section
4. Observe the recommended items

### What is the current _bug_ behavior?

The "More Like This" section displays items that appear to be randomly selected and are often completely unrelated to the source content. Genre, tags, and other metadata similarity seem to be ignored or heavily diluted by the randomization.
The current behaviour is even more obvious when watching a Movie in a Collection and none of the other movies even show up in the suggestions.

### What is the expected _correct_ behavior?

"More Like This" should show items that are genuinely similar to the current item, sorted by relevance (matching genres, tags, actors, directors, studios, etc.). The most relevant matches should appear first.

### Jellyfin Server version

Master

### Specify commit id

94fbc56918acc2f1496e2e7cbf9216da536ec436

### Specify version number

10.11.1

### Specify the build version

10.11.1

### Environment

```markdown
- Jellyfin Version: 10.11.1
- Installation Method: Docker
- Client: Web browser
- Browser: Chrome
```

### Jellyfin logs

```shell
[...]

PR #17000, opened 2026-06-01
Add per-user taste-profile recommendations for Movies and Series

twigglits:
Add per-user taste-profile recommendations for Movies and Series

**Changes**
Adds a per-user **taste-profile recommendation engine** that powers
  `/Movies/Recommendations` and `/Items/Suggestions` with personalized,
  metadata-driven results instead of the current random output. It builds a
  weighted profile (genres, tags, people, studios) from each user's locally
  watched and favorited items, then ranks candidates by how well they match that
  profile. The "similar to…" categories source their candidates from the existing
  `ISimilarItemsManager` providers (#16856) and re-rank them with the taste
  profile — so this layers personalization **on top of** the current similarity
  pipeline rather than replacing it. No API routes, request parameters, response
  DTOs, or database schema change.

**Issues**
- `MoviesController.GetMovieRecommendations` lost its similarity criteria when the
    `SimilarTo` filter was removed from `InternalItemsQuery` during the EF Core
    migration (#12798); each "Because you watched X" row emits random movies.
  - `SuggestionsController.GetSuggestions` ranks solely by `OrderBy = Random`.

  Reported in #14088 and #15342; partially addressed by #16856. This PR supplies
  the missing personalization signal.


JPVenson:
Closed as not following our AI guidelines

PR #17004, opened 2026-06-02
fix: index AlternateExtensions in GetSubtitleFormatTypes to prevent lossy VTT roundtrip

ferezvi:
**Changes**
`GetSubtitleFormatTypes()` only indexed `SubtitleFormat.Extension` when building the format lookup dictionary, ignoring `SubtitleFormat.AlternateExtensions`. This caused `SupportsFileExtension("webvtt")` to return `false` — since libse's
`WebVTT` format exposes `"webvtt"` via `AlternateExtensions`, not `Extension` — triggering a fallback to ffmpeg conversion that permanently strips all WebVTT cue settings (`position`, `line`, `align`, `size`), breaking subtitle positioning on clients that support them (e.g. ExoPlayer on Android).

This fix iterates `AlternateExtensions` alongside `Extension` when registering formats, so `"webvtt"` is recognized and the file is served directly with cue settings intact. Any other libse format that exposes alternate extensions benefits automatically.

**Issues**
Fixes #

### DIFF
--- a/MediaBrowser.MediaEncoding/Subtitles/SubtitleEditParser.cs
+++ b/MediaBrowser.MediaEncoding/Subtitles/SubtitleEditParser.cs
@@ -12,7 +12,7 @@ using SubtitleFormat = Nikse.SubtitleEdit.Core.SubtitleFormats.SubtitleFormat;
 namespace MediaBrowser.MediaEncoding.Subtitles
 {
     /// <summary>
-    /// SubStation Alpha subtitle parser.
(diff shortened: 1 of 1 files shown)


Shadowghost (commented):
I think we should dedup within a bucket. If a format's AlternateExtensions happens to include its own primary extension, the same Type is added to that bucket twice.

If format A's alternate extension equals format B's primary, A's type is appended to B's bucket, adding parser candidates. I don't think we use Parse anywhere right now, but preventing that additional roundtrip would be nice either way.

PR #17011, opened 2026-06-03
Refactor lyric data and add TTML parsing

6xingyv:
**Changes**
The existing lyric model stored lyrics as a flat list of lines with cue ranges, while lyric files were tracked separately from the media stream structure used by subtitles and other external media. This made it hard to represent richer lyric formats such as multi-track lyrics, translations, phonetics, background vocals, and syllable timing without adding format-specific fields.

This refactors lyrics into track-based data: main lyrics share one track, translated and phonetic lyrics are represented as separate tracks, each line can reference artist ids, and syllable timing is stored directly on lyric lines. Lyric discovery is kept aligned with the existing media stream approach instead of introducing a separate lyric file info structure. This also adds TTML lyric parsing, adapted from the Kotlin implementation in [`6xingyv/accompanist-lyrics-core`](https://github.com/6xingyv/accompanist-lyrics-core).

**Issues**
[Feature Request](https://features.jellyfin.org/posts/3959/add-a-track-based-lyric-data-model)


ThisIsTeddyBear:
I was planning to put up a pull request and saw, for a work around I implemented my own way of using and getting support for TTML, by converting them to ELRC and then using JS and CSS I have implemented it to work in swapping jelllyfins lyrics page.
@6xingyv You may have a look at this once if you want -> https://github.com/ThisIsTeddyBear/Jellyfin-LyricMotion
Work around 😊

6xingyv:
> I was planning to put up a pull request and saw, for a work around I implemented my own way of using and getting support for TTML, by converting them to ELRC and then using JS and CSS I have implemented it to work in swapping jelllyfins lyrics page. @6xingyv You may have a look at this once if you want -> https://github.com/ThisIsTeddyBear/Jellyfin-LyricMotion Work around 😊

Thanks for this! Acutally I was using a similar way in my own Jellyfish instance(basically replace the current lyrics page with all and parse TTML directly).

And that's why I wrote this PR, we do need an official way to support these all different lyrics formats.

ThisIsTeddyBear:
@6xingyv Is your current PR working properly like parsing and other things working properly ? Should I switch to this and give it a try ? Any known bugs that need addressal ?

6xingyv:
> @6xingyv Is your current PR working properly like parsing and other things working properly ? Should I switch to this and give it a try ? Any known bugs that need addressal ?

It should work properly, I did some tests. But I didn't use it in my server for now.
And I think I might also migrate lyrics parser to plugin, only leaving the data and api in core.

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
```diff
@@ -563,11 +563,21 @@ private void AddExternalLyrics(
...
+                if (currentStreams.Any(i => string.Equals(i.Path, externalLyricFile.Path, StringComparison.Ordinal)))
+                {
+                    continue;
+                }
+
+                currentStreams.Add(externalLyricFile);
             }
+
+            audio.LyricFiles = currentStreams
+                .Where(i => i.Type == MediaStreamType.Lyric && i.IsExternal)
```
gnattu, line 577: _lyricResolver.GetExternalStreams returns stream without `isExternal` flag so this filter does not work as your intended way

MediaBrowser.Model/Lyrics/LyricDto.cs
```diff
@@ -12,8 +13,17 @@ public class LyricDto
...
+    public IReadOnlyList<LyricTrack> Tracks { get; set; } = [];
+
     /// <summary>
     /// Gets or sets a collection of individual lyric lines.
     /// </summary>
-    public IReadOnlyList<LyricLine> Lyrics { get; set; } = [];
+    public IReadOnlyList<LyricLine> Lyrics
+    {
+        get => Tracks.FirstOrDefault(i => i.Type == LyricTrackType.Main)?.Lines ?? [];
+        set => Tracks = [new LyricTrack { Type = LyricTrackType.Main, Lines = value }];
```
gnattu, line 27: you are discarding ALL other tracks whenever lyrics are assigned, which is something we should not do

MediaBrowser.Providers/Lyric/TtmlLyricParser.cs
```diff
@@ -0,0 +1,410 @@
...
+    {
+        if (lines.Count == 0)
+        {
+            return;
+        }
+
+        tracks.Add(new LyricTrack
+        {
+            Type = type,
+            Language = languages.Count == 1 ? languages.First() : null,
```
gnattu, line 170: you are combining all translations into one track. is that the intended behavior?

MediaBrowser.Providers/Lyric/TtmlLyricParser.cs
```diff
@@ -0,0 +1,410 @@
...
+
+    private static bool HasRole(XElement element, string role)
+        => element.Attributes().Any(i => i.Name.LocalName == "role" && i.Value == role);
+
+    private static bool HasAnyRole(XElement element, params string[] roles)
+        => element.Attributes().Any(i => i.Name.LocalName == "role" && roles.Contains(i.Value, StringComparer.Ordinal));
+
+    private static string PreformatTtml(string content)
+        => content
+            .Replace("  ", string.Empty, StringComparison.Ordinal)
```
gnattu, line 374: Excuse me but why we need to replace two spaces here?

MediaBrowser.Providers/Lyric/TtmlLyricParser.cs
```diff
@@ -0,0 +1,410 @@
...
+        var translationLanguages = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+        var phoneticLanguages = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+
+        foreach (var p in document.Descendants().Where(i => i.Name.LocalName == "p"))
+        {
+            var start = ParseTime(GetAttributeValue(p, "begin"));
+            var end = ParseTime(GetAttributeValue(p, "end"));
+            if (!start.HasValue || !end.HasValue)
+            {
+                continue;
```
gnattu, line 69: We are discarding ttml without timing info BUT untimed ttml is allowed by spec: https://help.apple.com/itc/videoaudioassetguide/en.lproj/static.html#itca96ba3a47

It says the timing is optional.

gnattu (changes requested):
There might be more issues but current implementation is still to draft, or in other words "slop". to merge.

It has unwanted side effects, and its parser does not really match the ttml spec. Instead of hand writing our own ttml parser which is easy to make mistakes, is there any libraries that does this for us?

6xingyv:
> There might be more issues but current implementation is still to draft, or in other words "slop". to merge.
> 
> It has unwanted side effects, and its parser does not really match the ttml spec. Instead of hand writing our own ttml parser which is easy to make mistakes, is there any libraries that does this for us?

Thanks for your review! I do have my own [TTML parser](https://github.com/6xingyv/accompanist-lyrics-core), which is already widely used and tested in many android music apps. But the problem is: It is written in Kotlin.... And I'm not familiar with C# and that's why I just use LLM to translated it into C# and the result turned out to be far from perfect 😅

Btw, I have a idea that we should separate these lyrics parsers into standalone plugins just like the lyrics provides plugins. So we dont need to merge every formats manually into the main codebase, and everyone can use their preferred formats. 😍

PR #17013, opened 2026-06-03
Reject unsafe plugin package names in installer

### LINKED ISSUE #16899: Plugin installer should reject package names that resolve outside the plugins directory
### Description of the bug

The plugin installation flow uses the package name from a configured plugin repository as part of the installation directory without validating that it is a safe path segment.

In `InstallationManager.PerformPackageInstallation`, `targetDir` is built from:

```csharp
Path.Combine(_appPaths.PluginsPath, package.Name)
```

If a configured repository returns a package named `../../provally_jellyfin_path_traversal`, the path later resolves outside the configured plugins directory. The same unsanitized path is passed to `ZipFile.ExtractToDirectoryAsync(...)` and then to `PluginManager.SaveManifest(...)`, which writes `meta.json`.

This is admin-only because registering plugin repositories and installing plugins requires elevated privileges. Still, plugin installation should keep plugin files contained under `_appPaths.PluginsPath`.

I have attached a self-contained PoC script:

```text
jellyfin_plugin_path_traversal_poc.py
```

By default, the PoC does not generate a local proof page. It triggers the bug and opens Jellyfin's own `/System/Logs/Log` endpoint so the final human verification screen is served by the target Jellyfin server itself.

### Reproduction steps

1. Start Jellyfin Server from `master`.
2. Ensure an administrator account exists.
3. Run the attached PoC from a machine that the Jellyfin server can call back to:

[jellyfin_plugin_path_traversal_poc.py](https://github.com/user-attachments/files/28137555/jellyfin_plugin_path_traversal_poc.py)

```bash
python3 jellyfin_plugin_path_traversal_poc.py \
  --url http://localhost:8096 \
  --username admin \
  --password '<admin-password>' \
  --attacker-host host.docker.internal
```

For non-Docker/localhost setups, replace `--attacker-host` with an address reachable from the Jellyfin server.

4. The PoC performs the following actions without manual UI interaction:
   - Authenticates as the admin user.
   - Starts a mock plugin repository HTTP server.
   - Registers that repository through `POST /Repositories`.
   - Serves a repository manifest containing package name `../../provally_jellyfin_path_traversal`.
   - Installs the package by `assemblyGuid`.
   - Opens Jellyfin's real log endpoint for final verification.

5. Confirm that the browser is on the target Jellyfin server log page and shows `DirectoryNotFoundException` for the escaped `meta.json` path.

Example PoC output:

```text
[*] Target: http://localhost:8096
[+] Mock repository server listening on 0.0.0.0:60935
[*] Mode: error-screen
[+] Target version: 12.0.0
[+] Authenticated as admin
[*] Trying mock repository URL reachable as http://host.docker.internal:60935/manifest.json
[+] Jellyfin fetched the mock repository via host.docker.internal
[*] Triggering package installation by assemblyGuid
[-] Install endpoint returned HTTP 404: Error processing request.

[*] Vulnerable path construction:
    plugins path : /runtime/data/plugins
    raw target   : /runtime/data/plugins/../../provally_jellyfin_path_traversal_1.0.0
    resolved     : /runtime/provally_jellyfin_path_traversal_1.0.0
[*] Real target verification screen:
    http://localhost:8096/System/Logs/Log?name=log_20260522.log&ApiKey=<redacted>#:~:text=DirectoryNotFoundException
```

Jellyfin accepts a repository package name containing `../` path segments and uses it when constructing the plugin installation path.

In the default PoC mode, the ZIP archive is intentionally empty. On current Jellyfin builds, this does not create the escaped target directory, so `SaveManifest()` fails while trying to write:

```text
/runtime/provally_jellyfin_path_traversal_1.0.0/meta.json
```

The resulting error appears in Jellyfin's own server logs:

```text
System.IO.DirectoryNotFoundException: Could not find a part of the path '/runtime/provally_jellyfin_path_traversal_1.0.0/meta.json'.
   at Emby.Server.Implementations.Plugins.PluginManager.SaveManifest(PluginManifest manifest, String path) in /src/Emby.Server.Implementations/Plugins/PluginManager.cs:line 369
   at Emby.Server.Implementations.Plugins.PluginManager.PopulateManifest(PackageInfo packageInfo, Version version, String path, PluginStatus status) in /src/Emby.Server.Implementations/Plugins/PluginManager.cs:line 439
   at Emby.Server.Implementations.Updates.InstallationManager.PerformPackageInstallation(InstallationInfo package, PluginStatus status, CancellationToken cancellationToken) in /src/Emby.Server.Implementations/Updates/InstallationManager.cs:line 568
```

The PoC also has a `--mode write` option. In that mode, the served ZIP contains a regular file, so `ExtractToDirectoryAsync(...)` creates the escaped target directory and Jellyfin writes both the ZIP payload and `meta.json` outside the plugins directory.

Example write-impact command:

```bash
python3 jellyfin_plugin_path_traversal_poc.py \
  --mode write \
  --url http://localhost:8096 \
  --username admin \
  --password '<admin-password>' \
  --attacker-host host.docker.internal
```

Observed escaped path in write mode:

```text
/runtime/provally_jellyfin_path_traversal_1.0.0/payload.txt
/runtime/provally_jellyfin_path_traversal_1.0.0/meta.json
```

Jellyfin should reject plugin package names that are not safe directory names, or canonicalize the final target directory and verify that it remains under `_appPaths.PluginsPath` before deleting directories, extracting ZIP contents, downloading images, or writing `meta.json`.

For example, package names containing path separators, `.` segments, `..` segments, rooted paths, or other invalid path characters should not be accepted as installation directory names.

### Jellyfin Server version

Master

### Specify commit id

e2fe5c163ae7ca12e862df70fb8d6caa14e86ef2

_No response_

### Specify version number

_No response_

### Specify the build version

12.0.0 local Debug build from master commit e2fe5c163ae7ca12e862df70fb8d6caa14e86ef2

### Environment

```markdown
- OS: Ubuntu 24.04.4 LTS inside Docker
- Linux Kernel: Docker host kernel
- Virtualization: Docker
- Clients: PoC script / Jellyfin API
- Browser: Default system browser opened by the PoC for the Jellyfin log endpoint
- FFmpeg Version: 6.1.1
- Playback Method:
- Hardware Acceleration:
- CPU Model:
- GPU Model:
- Plugins: default bundled plugins only
- Reverse Proxy: none
- Base URL: none
- Networking: Docker bridge/NAT
- Jellyfin Data Storage & Filesystem: bind-mounted local temporary directory
- Media Storage & Filesystem:
- External Integrations:
```

### Jellyfin logs

```shell
Emby.Server.Implementations.Plugins.PluginManager: No local manifest exists for plugin ../../provally_jellyfin_path_traversal. Skipping manifest reconciliation.
Emby.Server.Implementations.Updates.InstallationManager: Package installation failed
System.IO.DirectoryNotFoundException: Could not find a part of the path '/runtime/provally_jellyfin_path_traversal_1.0.0/meta.json'.
   at Emby.Server.Implementations.Plugins.PluginManager.SaveManifest(PluginManifest manifest, String path) in /src/Emby.Server.Implementations/Plugins/PluginManager.cs:line 369
   at Emby.Server.Implementations.Plugins.PluginManager.PopulateManifest(PackageInfo packageInfo, Version version, String path, PluginStatus status) in /src/Emby.Server.Implementations/Plugins/PluginManager.cs:line 439
   at Emby.Server.Implementations.Updates.InstallationManager.PerformPackageInstallation(InstallationInfo package, PluginStatus status, CancellationToken cancellationToken) in /src/Emby.Server.Implementations/Updates/InstallationManager.cs:line 568
Jellyfin.Api.Middleware.ExceptionMiddleware: Error processing request: Could not find a part of the path '/runtime/provally_jellyfin_path_traversal_1.0.0/meta.json'. URL POST /Packages/Installed/provally-repro.
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

The PoC opens the real Jellyfin log URL at the end of execution, with a text fragment targeting `DirectoryNotFoundException`. A screenshot of that Jellyfin-served log page can be attached if useful.

### Additional information

The route parameter itself does not need to contain the traversal payload. In testing, using an encoded traversal value directly in the route returned 404. The reliable path is to keep the route name benign and select the malicious repository package by `assemblyGuid`.

The issue is admin-only, but it is still a plugin installer containment bug: untrusted repository metadata controls a filesystem path that should remain under the Jellyfin plugins directory.

dfederm:
**Changes**

`InstallationManager.PerformPackageInstallation` builds the install directory with `package.Name` straight from a repository manifest, so a name like `../../evil` resolves outside the plugins root and is then passed to `Directory.Delete`, `ZipFile.ExtractToDirectoryAsync`, and `PluginManager.SaveManifest`. Admin-only so not exactly a security issue, but the installer should still validate inputs appropriately.
 
The change validates `package.Name` against null/whitespace, `.`, `..`, path separators, and `Path.GetInvalidFileNameChars`, and verifies the fully resolved target stays under the plugins path.

**Issues**
Fixes #16899

### DIFF
--- a/Emby.Server.Implementations/Updates/InstallationManager.cs
+++ b/Emby.Server.Implementations/Updates/InstallationManager.cs
@@ -521,8 +521,26 @@ namespace Emby.Server.Implementations.Updates
                 return;
             }
 
+            if (!IsValidPackageDirectoryName(package.Name))
+            {
+                _logger.LogError("Refusing to install package with invalid name {PackageName}.", package.Name);
+                throw new InvalidDataException($"Plugin package name '{package.Name}' is not a valid directory name.");
+            }
+
             // Always override the passed-in target (which is a file) and figure it out again
             string targetDir = Path.Combine(_appPaths.PluginsPath, package.Name);
+
+            var pluginsRoot = Path.TrimEndingDirectorySeparator(Path.GetFullPath(_appPaths.PluginsPath));
+            var resolvedTarget = Path.GetFullPath(targetDir);
+            if (!resolvedTarget.StartsWith(pluginsRoot + Path.DirectorySeparatorChar, StringComparison.OrdinalIgnoreCase))
+            {
+                _logger.LogError(
+                    "Refusing to install package {PackageName}: resolved target {Resolved} is outside plugins directory {Root}.",
+                    package.Name,
+                    resolvedTarget,
+                    pluginsRoot);
+                throw new InvalidDataException($"Plugin package name '{package.Name}' resolves outside the plugins directory.");
+            }
 
             using var response = await _httpClientFactory.CreateClient(NamedClient.Default)
                 .GetAsync(new Uri(package.SourceUrl), cancellationToken).ConfigureAwait(false);
@@ -570,6 +588,31 @@ namespace Emby.Server.Implementations.Updates
             await _pluginManager.PopulateManifest(package.PackageInfo, package.Version, targetDir, status).ConfigureAwait(false);
 
             _pluginManager.ImportPluginFrom(targetDir);
+        }
+
+        private static bool IsValidPackageDirectoryName(string? name)
+        {
+            if (string.IsNullOrWhiteSpace(name))
+            {
+                return false;
+            }
+
+            if (name.Equals(".", StringComparison.Ordinal) || name.Equals("..", StringComparison.Ordinal))
+            {
+                return false;
+            }
+
+            if (name.Contains('/', StringComparison.Ordinal) || name.Contains('\\', StringComparison.Ordinal))
+            {
+                return false;
+            }
+
+            if (name.AsSpan().IndexOfAny(Path.GetInvalidFileNameChars()) >= 0)
+            {
+                return false;
+            }
+
+            return true;
         }
 
         private async Task<bool> InstallPackageInternal(InstallationInfo package, CancellationToken cancellationToken)
--- a/Emby.Server.Implementations/Updates/InstallationManager.cs
+++ b/Emby.Server.Implementations/Updates/InstallationManager.cs
@@ -32,6 +32,8 @@ namespace Emby.Server.Implementations.Updates
     /// </summary>
     public class InstallationManager : IInstallationManager
     {
+        private static readonly char[] InvlidPackageNameChars = [.. Path.GetInvalidFileNameChars(), '/', '\\'];
+
         /// <summary>
         /// The logger.
         /// </summary>
@@ -602,12 +604,7 @@ namespace Emby.Server.Implementations.Updates
                 return false;
             }
 
-            if (name.Contains('/', StringComparison.Ordinal) || name.Contains('\\', StringComparison.Ordinal))
-            {
-                return false;
-            }
-
-            if (name.AsSpan().IndexOfAny(Path.GetInvalidFileNameChars()) >= 0)
+            if (name.AsSpan().IndexOfAny(InvlidPackageNameChars) >= 0)
             {
                 return false;
             }
--- a/Emby.Server.Implementations/Updates/InstallationManager.cs
+++ b/Emby.Server.Implementations/Updates/InstallationManager.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Buffers;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.IO;
@@ -32,7 +33,7 @@ namespace Emby.Server.Implementations.Updates
     /// </summary>
     public class InstallationManager : IInstallationManager
     {
-        private static readonly char[] InvlidPackageNameChars = [.. Path.GetInvalidFileNameChars(), '/', '\\'];
+        private static readonly SearchValues<char> InvalidPackageNameChars = SearchValues.Create([.. Path.GetInvalidFileNameChars(), '/', '\\']);
 
         /// <summary>
         /// The logger.
@@ -604,7 +605,7 @@ namespace Emby.Server.Implementations.Updates
                 return false;
             }
 
-            if (name.AsSpan().IndexOfAny(InvlidPackageNameChars) >= 0)
+            if (name.IndexOfAny(InvalidPackageNameChars) >= 0)
             {
                 return false;
             }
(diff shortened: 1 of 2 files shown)


IDisposable (approved):

Emby.Server.Implementations/Updates/InstallationManager.cs
```diff
@@ -572,6 +590,31 @@ private async Task PerformPackageInstallation(InstallationInfo package, PluginSt
...
+            {
+                return false;
+            }
+
+            if (name.Contains('/', StringComparison.Ordinal) || name.Contains('\\', StringComparison.Ordinal))
+            {
+                return false;
+            }
+
+            if (name.AsSpan().IndexOfAny(Path.GetInvalidFileNameChars()) >= 0)
```
IDisposable: Couldn't we have only this and add the various characters tested above to the invalid file name list?

dfederm: Sure, I'll add a new static reasonly char[] to have these and the / and \ checks. The other checks in this method will stay as-is though.

Emby.Server.Implementations/Updates/InstallationManager.cs
```diff
@@ -32,6 +32,8 @@ namespace Emby.Server.Implementations.Updates
     /// </summary>
     public class InstallationManager : IInstallationManager
     {
+        private static readonly char[] InvlidPackageNameChars = [.. Path.GetInvalidFileNameChars(), '/', '\\'];
```
IDisposable: If we use `SearchValues<char>` it should be much faster
```csharp
private static readonly SearchValues<char> InvalidPackageNameChars = 
    SearchValues.Create(
        [.. Path.GetInvalidFileNameChars(), '/', '\\']
    );

...

if (name.AsSpan().IndexOfAny(InvalidPackageNameChars) < 0)
   return false;
```

_also fixes the typo of `Invlid` -> `Invalid`_

IDisposable: If we use `SearchValues<char>` it should be much faster
```csharp
private static readonly SearchValues<char> InvalidPackageNameChars = 
    SearchValues.Create(
        [.. Path.GetInvalidFileNameChars(), '/', '\\']
    );

...

if (name.AsSpan().IndexOfAny(InvalidPackageNameChars) < 0)
   return false;
```

_also fixes the typo of `Invlid` -> `Invalid`_

IDisposable (approved):

jumoog:
I spotted that same issue. I don't think it's necessary to fix it since a running plugin is not restricted.

dfederm:
This is less about security (admins are making these decisions) and more about data validation.

JPVenson (approved):

crobibero (approved):

PR #17033, opened 2026-06-06
Fix exit code for --help and --version CLI options

booxter:
**Changes**

Before the patch, both returned 1 despite accepted CLI conventions.

Also, --version printed:

```
ERROR(S):
  Option 'version' is unknown.
```

After the patch, the error is no longer shown and the exit code for both
options is 0.

**Issues**

### DIFF
--- a/Jellyfin.Server/Program.cs
+++ b/Jellyfin.Server/Program.cs
@@ -1,7 +1,6 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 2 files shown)


Jellyfin.Server/Program.cs
```diff
@@ -72,15 +71,31 @@ public static Task Main(string[] args)
...
+                    .ParseArguments<StartupOptions>(new[] { builtInOption })
+                    .MapResult(static _ => Task.CompletedTask, ErrorParsingArguments);
+            }
+
             // Parse the command line arguments and either start the app or exit indicating error
             return Parser.Default.ParseArguments<StartupOptions>(args)
                 .MapResult(StartApp, ErrorParsingArguments);
         }
 
+        private static bool IsBuiltInOption(string arg)
```
JPVenson, line 91: dont we have much more options then this?

booxter: The other options are handled later, see line 87. This is only meant to handle --help and --version, which are special - they don't start the app. Or do you mean the name is confusing here? Maybe IsHelpOrVersionOption clearer?

PR #17038, opened 2026-06-07
Implement dynamic collections (SmartCollections)

TiagoMateus05:
Implement new entity of SmartCollections, enabling a dynamic query based collection to automatically return a list of matching items per user description. This allows the user to create a SmartCollection with specific criteria, for example, movies from 2015 onwards with at least 6 IMDB rating, and always return the most recent always-current list, including anything added after creating the SmartCollection.

**Changes**

- New Entity: 
    - **SmartCollection.cs**, along with **SmartCollectionFilters.cs** for easier query build;
    - `SmartCollectionsManager.cs`, `SmartCollectionController.cs` for logic and endpoints control and new Database Table to save entity and the Json for later parsing;
    - **Database**: new `SmartCollections` EF Core migration included.
- New EndPoints:
    -  `POST /SmartCollections` — creates a new Smart Collection from a JSON body;
    - `GET /SmartCollections` — retrieves all Smart Collections for the authenticated user;
    - `GET /SmartCollections/{id}` — retrieves a specific Smart Collection by Id;
    - `GET /SmartCollections/{id}/Items`— retrieves, by Evaluation on-the-go, every matching item from the query build.
    
**Performance**
The choice of making the Evaluation be on-the-go was to save space and easily adapt to newly added movies. This can also cause high CPU cost for large libraries, so a default limit of 50 items was implemented along with a 10 minute TTL cache. SmartCollections was tested in 3 different hardwares, with 100k Mock Items, with similar results:

| Scenario | Latency |
|---|---|
| Cold query search (no cache) | ~500ms |
| Cache hit (≤10 min window) | ~40–50ms |

**Testing**
Implements tests to validate the SmartCollections integrity and expected behavior and optional (run by setting a flag) tests to evaluate cache misses performance, cache hit performance and 100k Mock Items Evaluation performance.

**Discussion**
The initial discussion of implementation:
[Discussion](https://github.com/jellyfin/jellyfin-meta/discussions/122)


JPVenson:
before we slap on another collection type feature we need to refactor the existing ones into one unified system so that we do not keep adding _the same_ functionality over and over again.

TiagoMateus05:
Hey!

Just to clarify the scope and what we would like to see in Jellyfin: this implementation doesn't touch or modify any existing collection code, it's entirely additive and independent. A future refactor of the existing collection types could happen independently of this PR without any conflict.

We also searched through open issues and couldn't find any tracking issue for the refactor you're referring to. Our proposal has been open in jellyfin-meta#122 for two months without this concern being raised. 

Can you explain to us where that refactor was mentioned? We would be happy to understand how both can coexist, but it would help to know what we need to consider!

jhparker88:
I'm guessing this would cover [1], [2], [3] and potentially [4], [5].

[1] https://features.jellyfin.org/posts/1921/collections-based-on-tags-actor-director
[2] https://features.jellyfin.org/posts/3743/smart-collections
[3] https://features.jellyfin.org/posts/593/dynamic-smart-collections-defined-by-custom-queries
[4] https://features.jellyfin.org/posts/2008/movie-queues
[5] https://features.jellyfin.org/posts/3562/add-support-for-dynamic-libraries

TiagoMateus05:
@theguymadmax Hey! Sorry for pinging, but both me and @beateixeira04 are new to Jellyfin and we don't know if there is a team ping for the maintainers. We would like a review for the implementation and we also would like to know if there are changes needed to our code to respect Jellyfins guidelines.

PR #17005, opened 2026-06-02
Version-aware playback tracking and multi-version fixes

Shadowghost:
When playing version B of  a movie/episode right now when resuming, we are playing version A from that progress which is incorrect.

This leads to a sideffect though:  The version that has playback data will appear first in the version dropdown

**Changes**
* Fix episode version naming stripping the common string
* Track playback progress on a per version basis
* Track played state for all versions of an item
* Don't exclude versions from data extraction
* Exlcude versions a user does not have access to
* Surface extras for all versions

**Issues**
Fixes #11118
Fixes #11967

**AI Code Assistance**
Claude Opus 4.6 was used for debugging issues and cleaning up the implementation.

PR #7997, opened 2022-06-23
Parse episode extras

SenorSmartyPants:
**Changes**
Subfolder and suffix style extras support for episodes. Includes unit tests for series, season and episode level extras.

Feature parity with [plex ](https://support.plex.tv/articles/local-files-for-tv-show-trailers-and-extras/).

Edit: upon further testing, JF does support -clip2 style suffixes.

PR #17044, opened 2026-06-07
Fixes for multi version handling

Shadowghost:
Split-out of https://github.com/jellyfin/jellyfin/pull/17005 

**Changes**
* Implement version-aware playback/progress tracking and resume
* Fix returned version names
* Fix unplayed/played propagation
* Surface Extras for all versions
* Only mark linked alternate versions as grouped media sources (otherwise "split" will appear on frontend)

**AI Code Assistance**
Claude Opus 4.8 was used to extract this out of the original PR, validation was done manually.


Jellyfin.Server.Implementations/Item/BaseItemRepository.TranslateQuery.cs
```diff
@@ -539,22 +543,36 @@ public IQueryable<BaseItemEntity> TranslateQuery(
...
                     .Select(s => s.SeriesId);
 
-                // Non-series items: resumable if PlaybackPositionTicks > 0
-                var resumableItemIds = context.UserData
-                    .Where(ud => ud.UserId == userId && ud.PlaybackPositionTicks > 0)
-                    .Select(ud => ud.ItemId);
+                // Non-series items: resumable if the item or any of its alternate versions has
+                // PlaybackPositionTicks > 0. Alternate versions (PrimaryVersionId set) are excluded
+                // from the base query, so coalesce their progress onto the primary's id.
+                var resumableMovieIds = context.BaseItems
```
JPVenson: typo?


```suggestion
                var resumableSeriesIds = context.BaseItems
```

JPVenson (approved):
db looks good

Jellyfin.Server.Implementations/Item/BaseItemRepository.TranslateQuery.cs
Shadowghost: This is correct as is, resumableSeriesIds  is defined elsewhere

jhparker88:
> Surface Extras for all versions

Just to clarify, this doesn't include parsing extras from an episode (such as `/series/Dexter/Season 1/Dexter - S01E02 - Second Epi-interview.mkv` etc.) as mentioned and suggested in https://github.com/jellyfin/jellyfin/pull/7997?

Shadowghost:
@jhparker88 no it does not. It just copies extras from the parent entity to the child entity - primarily movies. I'm not sure if we need per-episode extras to be hones.

theguymadmax (approved):

Emby.Server.Implementations/Session/SessionManager.cs
```diff
@@ -1138,6 +1177,12 @@ private bool OnPlaybackStopped(User user, BaseItem item, long? positionTicks, bo
 
             _userDataManager.SaveUserData(user, item, data, UserDataSaveReason.PlaybackFinished, CancellationToken.None);
 
+            // A completed version marks all of its alternate versions played; positions stay per-version.
+            if (data.Played == true && item is Video playedVideo)
+            {
+                playedVideo.PropagatePlayedState(user, true);
```
crobibero, line 1188: **Finishing one version wipes the resume point of the others — contradicting the comment above.**

`PropagatePlayedState(user, true)` relies on the default `resetPosition: true`, which sets `PlaybackPositionTicks = 0` on every sibling version. But the comment on line 1180 says *"positions stay per-version."*

Scenario: user is 30 min into version B, then watches version A to completion → `OnPlaybackStopped` fires for A, `PropagatePlayedState` zeroes B's `PlaybackPositionTicks`, destroying B's in-progress resume point. (The same call in `OnPlaybackProgress` at line 985 has the same effect.)

Either pass `resetPosition: false` here, or fix the comment — but wiping a sibling's distinct resume point looks unintended given the PR's per-version-resume goal.

Emby.Server.Implementations/Session/SessionManager.cs
```diff
@@ -952,6 +979,17 @@ private void OnPlaybackProgress(User user, BaseItem item, PlaybackProgressInfo i
...
             {
                 _userDataManager.SaveUserData(user, item, data, UserDataSaveReason.PlaybackProgress, CancellationToken.None);
+
+                if (data.Played == true && item is Video playedVideo)
+                {
+                    playedVideo.PropagatePlayedState(user, true);
+                }
+            }
+
+            if ((!user.RememberAudioSelections && data.AudioStreamIndex.HasValue)
```
crobibero: **This block is dead code — the intended cross-version stream-selection cleanup never runs.**

`UpdatePlaybackSettings(user, info, data)` is called unconditionally earlier in `OnPlaybackProgress` and already nulls `data.AudioStreamIndex` / `data.SubtitleStreamIndex` in memory when `!RememberAudioSelections` / `!RememberSubtitleSelections`. So by the time this guard evaluates `(!user.RememberAudioSelections && data.AudioStreamIndex.HasValue) || …`, both `HasValue` checks are already `false` for exactly the users the condition targets.

Result: `ResetPlaybackStreamSelections` never fires. If the goal is to clear stored selections across all versions/keys, this needs to key off something other than the already-nulled in-memory `data` (e.g. the incoming `info` values, or the persisted rows).

Jellyfin.Server.Implementations/Item/BaseItemRepository.TranslateQuery.cs
```diff
@@ -742,10 +760,13 @@ public IQueryable<BaseItemEntity> TranslateQuery(
...
         else if (filter.OwnerIds.Length == 0 && filter.ExtraTypes.Length == 0 && !filter.IncludeOwnedItems)
         {
-            // Exclude alternate versions and owned non-extra items from general queries.
-            // Alternate versions have PrimaryVersionId set (pointing to their primary).
+            // Exclude owned non-extra items from general queries.
             // Extras (trailers, etc.) have OwnerId set but also have ExtraType set - keep those.
-            baseQuery = baseQuery.Where(e => e.PrimaryVersionId == null && (e.OwnerId == null || e.ExtraType != null));
+            // Alternate versions (PrimaryVersionId set) are normally excluded too, but resume queries
+            // keep them so the actually-played version can surface instead of collapsing onto the primary.
+            baseQuery = filter.IsResumable == true
```
crobibero, line 771: **Keeping alternate-version rows in resume queries breaks `excludeActiveSessions` on `/Items/Resume`.**

With grouping disabled for `IsResumable` (`EnableGroupByPresentationUniqueKey` returns `false`) and alternate rows kept here, the resume query now returns the *actually-played alternate version's own id*. But `ItemsController.GetResumeItems` builds its `excludeActiveSessions` exclude set from `session.NowPlayingItem.Id`, which is the **displayed/primary** id (set in `OnPlaybackStart` from `info.ItemId`, not the `MediaSourceId`-resolved version).

So while a user is actively playing an alternate version, `GET /Items/Resume?excludeActiveSessions=true` still lists that item, because the returned alternate id never matches the primary exclude id.

Jellyfin.Server.Implementations/Item/OrderMapper.cs
```diff
@@ -34,7 +34,10 @@ public static class OrderMapper
             (ItemSortBy.AirTime, _) => e => e.SortName,
             (ItemSortBy.Runtime, _) => e => e.RunTimeTicks,
             (ItemSortBy.Random, _) => e => EF.Functions.Random(),
-            (ItemSortBy.DatePlayed, _) => e => e.UserData!.Where(f => f.UserId.Equals(query.User!.Id)).OrderBy(f => f.CustomDataKey).FirstOrDefault()!.LastPlayedDate,
+            (ItemSortBy.DatePlayed, not null) => e =>
```
crobibero, line 37: **Narrowing this arm to `not null` makes `DatePlayed` silently sort by name for a null user.**

The arm changed from `(ItemSortBy.DatePlayed, _)` to `(ItemSortBy.DatePlayed, not null)`, and there is no `(ItemSortBy.DatePlayed, null)` arm — so a null-user query falls through to the default `_ => e => e.SortName` and returns results in alphabetical order instead of by play date, with no error. Note `SeriesDatePlayed` has both a `not null` and a `null` arm, so this is inconsistent.

Reachable via `MoviesController.GetMovieRecommendations`, which builds an `InternalItemsQuery(user)` with a possibly-null user and `OrderBy = DatePlayed`. Consider adding an explicit null-user arm mirroring `SeriesDatePlayed`.

crobibero (commented):
Automated review — 4 findings that survived verification as confirmed. Each is posted inline. (Two further *plausible* findings and some cleanup/altitude notes were left out of this batch.)

MediaBrowser.Controller/Library/VersionResumeData.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// <summary>
+        /// Merges the most recently played version's completion state into the supplied user data dto.
+        /// Only completion (played) propagates to the primary; the in-progress resume position stays on
+        /// the version that owns it, which is surfaced directly (e.g. in resume queries) so that playback
+        /// always targets the correct version rather than resuming the primary at another version's offset.
+        /// </summary>
+        /// <param name="dto">The user data dto to update.</param>
+        public void ApplyTo(UserItemDataDto dto)
+        {
+            dto.Played = dto.Played || UserData.Played;
```
crobibero, line 25: **(plausible)** An item can end up marked `Played=true` while still carrying a partial progress bar.

`ApplyTo` sets `dto.Played` and bumps `dto.LastPlayedDate`, but never clears `dto.PlaybackPositionTicks` / `dto.PlayedPercentage`, which come from the *primary's own* user data. If the primary was partially watched and an alternate version is later finished, the DTO carries `Played=true` alongside a non-zero position/percentage — a "watched and resumable" state.

`VersionResumeDataTests.ApplyTo_PropagatesCompletionButNotPosition` actually asserts exactly this (`Played=true` with `PlaybackPositionTicks=1`, `PlayedPercentage=1.0`), so the inconsistent DTO is enshrined as intended. Whether the client then renders both a checkmark and a resume bar is UI-dependent — worth confirming that's acceptable, and if not, clearing the position/percentage when completion is propagated.

Jellyfin.Server.Implementations/Item/BaseItemRepository.TranslateQuery.cs
```diff
@@ -535,26 +539,40 @@ public IQueryable<BaseItemEntity> TranslateQuery(
...
+                baseQuery = hasSeries
+                    ? baseQuery.Where(e =>
+                        (e.Type == seriesTypeName && resumableSeriesIds!.Contains(e.Id))
+                        || (e.Type != seriesTypeName && inProgressIds.Contains(e.Id)))
+                    : baseQuery.Where(e => inProgressIds.Contains(e.Id));
+
+                // When several versions of the same item are in progress, keep only the most recently played one.
+                baseQuery = baseQuery.Where(e => e.Type == seriesTypeName || !context.BaseItems
+                    .Where(s => s.Id != e.Id && (s.PrimaryVersionId ?? s.Id) == (e.PrimaryVersionId ?? e.Id))
+                    .Any(s => inProgress.Where(su => su.ItemId == s.Id).Max(su => su.LastPlayedDate)
```
crobibero: **(plausible)** Duplicate resume entries when two versions share an identical `LastPlayedDate`.

The version-dedup uses a strict `>`: a row `e` is dropped only if a sibling has a *strictly greater* `Max(LastPlayedDate)`. If two in-progress versions of the same item have the exact same `LastPlayedDate`, neither is strictly greater, so **both** survive — and since presentation-key grouping is disabled for resume queries, nothing collapses them, so the item shows up twice in Continue Watching (and inflates `TotalRecordCount`).

Normal playback stamps distinct `DateTime.UtcNow` ticks so this is uncommon, but it's reachable via the `UpdateUserItemData` API (arbitrary `LastPlayedDate`, no reset/propagation) or an import/migration. A tie-break on `Id` (e.g. `> || (== && s.Id < e.Id)`) would make the selection deterministic.

crobibero (commented):
Two additional *plausible* findings — the mechanism is real, but the trigger is narrower/UI-dependent, so treat these as lower priority than the confirmed batch.

Emby.Server.Implementations/Dto/DtoService.cs
```diff
@@ -1257,7 +1269,11 @@ private void AttachBasicFields(BaseItemDto dto, BaseItem item, BaseItem? owner,
 
                 if (options.ContainsField(ItemFields.MediaSourceCount))
                 {
-                    var mediaSourceCount = video.MediaSourceCount;
+                    // Match the per-user filtering of the media sources: versions the user cannot
+                    // access are not selectable, so they must not count towards the badge either.
+                    var mediaSourceCount = user is null
+                        ? video.MediaSourceCount
+                        : video.GetAllVersions().Count(v => v.Id.Equals(video.Id) || v.IsVisibleStandalone(user));
```
crobibero: **(efficiency)** This turns `MediaSourceCount` from a cached property read into per-item DB/library work on every list render.

For every video DTO with a non-null user, `video.GetAllVersions()` → `GetAllItemsForMediaSources()` issues uncached `GetLinkedAlternateVersions` / `GetLocalAlternateVersionIds` / `GetItemById` queries, and `IsVisibleStandalone(user)` walks parents + collection folders per version. On a 100-item library page that's hundreds of round-trips where the old `video.MediaSourceCount` was a single cached value.

Since the overwhelming majority of items have no alternate versions, short-circuit first — e.g. keep `video.MediaSourceCount` and only do the per-user visibility filtering when the item actually has more than one version (`LinkedAlternateVersions.Length > 0 || HasLocalAlternateVersions`).

Emby.Server.Implementations/Library/MediaSourceManager.cs
```diff
@@ -406,6 +410,72 @@ public IReadOnlyList<MediaSourceInfo> GetStaticMediaSources(BaseItem item, bool
...
+                }
+            }
+
+            MediaSourceInfo resumeSource = null;
+            UserItemData resumeData = null;
+            foreach (var source in sources)
+            {
+                if (source.Id is null
+                    || !dataBySourceId.TryGetValue(source.Id, out var data)
+                    || data.PlaybackPositionTicks <= 0)
```
crobibero: **(duplication / altitude)** The "pick the most-recently-played version by `LastPlayedDate`" scan is re-implemented four times in this PR, with divergent predicates:

- `UserDataManager.GetResumeUserDataBatch` — `PlaybackPositionTicks > 0 || Played`
- here (`SetAlternateVersionResumeStates`) — `PlaybackPositionTicks > 0` only (line 460)
- `TVSeriesManager.GetMostRecentlyPlayedVersion` — per-version `GetUserData` (an N+1)
- the SQL `Max(LastPlayedDate)` subquery in `BaseItemRepository.TranslateQuery`

Because the predicates already differ, the resume tile, the media-source default, and Next Up can disagree about which version is "current," and any future change has to be made in all four spots. Consider making `IUserDataManager.GetResumeUserData` the single owner (extended to return the winning version id) and having the other C# call sites delegate to it; the SQL form is the only one that legitimately needs its own implementation.

MediaBrowser.Providers/TV/SeriesMetadataService.cs
```diff
@@ -78,11 +78,73 @@ protected override async Task AfterMetadataRefresh(Series item, MetadataRefreshO
     {
         await base.AfterMetadataRefresh(item, refreshOptions, cancellationToken).ConfigureAwait(false);
 
+        // Note that this only updates the children's SeriesPresentationUniqueKey and SeasonId, not the ParentIndexNumber
+        if (LibraryManager.GetLibraryOptions(item).EnableAutomaticSeriesGrouping)
+        {
+            await UpdateSeriesChildrenInfoAsync(item, cancellationToken).ConfigureAwait(false);
```
crobibero, line 84: **(scope)** These `SeriesMetadataService` changes — `UpdateSeriesChildrenInfoAsync` (syncing `SeriesPresentationUniqueKey` / `SeasonId`) and the `FillMissingEpisodeNumbersFromPath` pass in `CreateSeasonsAsync` — are series-grouping / season-scan fixes unrelated to alternate-version handling.

They also touch the hot refresh path (an extra full `GetRecursiveChildren` pass plus per-episode `UpdateToRepositoryAsync` writes on every refresh). Bundling them into the versioning PR means the two can't be reviewed, bisected, or reverted independently. Consider splitting them into their own PR.

crobibero (commented):
Non-blocking cleanup / altitude notes (efficiency, duplication, scope). No correctness impact — feel free to defer.

crobibero:
The review comments were all generated by claude `/review https://github.com/jellyfin/jellyfin/pull/17044` and not verified. Some of them seem like they could be actual issues, some are nonsense. Please address all notes

Emby.Server.Implementations/Session/SessionManager.cs
Shadowghost: Code comment is wrong, the behavior works as intended, if you complete watching any version, you want all of them marked as watched and not one of them that is partially watched appear in continue watching. at least that's my opinion.
I will fix the comment with the rationale.

crobibero (approved):

nielsvanvelzen:
Please revert the API changes introduced here as we're not allowing more API changes since RC2.

Issue #148, opened 2018-12-14
Playback speed


gerroon:
It will awesome if Jellyfin offers palyback speed adjustment like Youtube and Vlc does. Basically the video can play faster or slower

JustAMan:
Do you want quick-and-dirty version or good one? :) I mean, speed up or slow down a video should not be that big of a deal in `ffmpeg` (but video quality would suffer in any case I think), but most likely would require transcoding. Are you willing to pay the price? :)

Audio is another thing, if you blindly speed it up or down it changes pitch and tone. At least Youtube goes smart and modified the tempo. I think `ffmpeg` has a library for that (`rubberband` should be the one doing this thing), but that is something to be learnt.

gerroon:
I am ok with transcoding, but transcoding will kick in only when the speed is adjusted right?

Cerothen:
I am not entirely familiar with the web player but it should be possible right in the web client, I don't see why playback rate would be a serverside thing.

https://stackoverflow.com/questions/3027707/how-to-change-the-playing-speed-of-videos-in-html5

JustAMan:
Nice find! I wonder what would happen to audio, though... Needs some experiments.

kusochi:
This seems to be something the browser itself can do. On plain html5 video example [https://www.w3schools.com/html/html5_video.asp](https://www.w3schools.com/html/html5_video.asp) try right clicking the video and at least on Firefox there seems to be option to change playback speed in the right click menu. I don't think there's need to transcode it using `ffmpeg`. Is this something that can be added to the player ui as it seems to be a standard feature?

gerroon:
@kusochi I seem to get speed menu only once, I set the playback speed, but then that menu is not available anymore in the right click menu

bugfixin:
I agree this should be a client-side feature. I'm looking at implementing this for jellyfin-web, but would like some input regarding the expected UI for playback speed adjustments.

kusochi:
I like the idea that the setting would be in the same "cogwheel menu" as the video quality setting selection. From there could open a submenu containing all the usual playback speed options like 0.5, 0.75, 1.0, 1.25 and 1.5 similar how they are in Youtube's player settings for example.

wyzdic:
This is a very good idea, I really need this,Is it possible to implement this feature in the near future?

PR #17046, opened 2026-06-08
Add AudioPlaybackRate parameter for server-side atempo filter

jordanmichaelrushing:
**Changes**
Accept an optional AudioPlaybackRate query parameter on the universal audio and HLS audio segment endpoints. When the rate differs from 1.0, force transcoding (bypassing direct play) and apply ffmpeg's atempo filter chain so the speed is baked into the encoded stream. Fixes audio-only HLS transcoding which was missing the audio filter step, and removes a StartTimeTicks guard that incorrectly rejected segment requests generated by Jellyfin's own HLS playlist when resuming from a non-zero position.

**Issues**
https://github.com/jellyfin/jellyfin/issues/148 (This is already closed, but only because feature requests were moved elsewhere. Also they consensus was a frontend only fix, which unfortunately does not work for iOS.

**Notes**
I originally was targeting master until I realized that was 12.x so I switched to 10.11.z. If I should be targeting 12.x, please let me know and I will update this.

### DIFF
--- a/Jellyfin.Api/Controllers/AudioController.cs
+++ b/Jellyfin.Api/Controllers/AudioController.cs
@@ -84,6 +84,7 @@ public class AudioController : BaseJellyfinApiController
     /// <param name="context">Optional. The <see cref="EncodingContext"/>.</param>
     /// <param name="streamOptions">Optional. The streaming options.</param>
     /// <param name="enableAudioVbrEncoding">Optional. Whether to enable Audio Encoding.</param>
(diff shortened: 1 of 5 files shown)


theguymadmax:
Nope, you made the wrong choice. Please re-target the master branch. Also note, we are finalizing our 12.0 release candidates, so there is a feature freeze.  Once we create a 12.0 release branch, new features can be merged into the master.

jordanmichaelrushing:
No worries. I'll close this one out and re-do the work and testing. Thanks @theguymadmax

Issue #2908, opened 2020-04-18
Sort title metadata not read


Sonico98:
**Describe the bug**
Most of my music has embedded sortalbum / albumsort tags, but they are not being read by the server. Instead, the Sort title field is empty on every album.

**System (please complete the following information):**
 - OS: Archlinux 5.6.4
 - Browser: Chrome 81
 - Jellyfin Version: 10.5.4
 - Installed Plugins: Anime
 - Reverse Proxy: none
 - Storage: Local

**To Reproduce**
1. Using either Kid3 or MP3Tag, add the Sort Album tag for an album.
2. Refresh the library's metadata.

**Expected behavior**
The newly added tag should show under "Sort title" when checking the library's metadata

**Screenshots**
![KhH7Irh](https://user-images.githubusercontent.com/61394886/79627635-627c6200-8110-11ea-966b-f55877533b9d.png)

NeuroDawg:
Can this issue be re-opened and addressed by the developers. If not re-opened, I'll start a new issue. I tag my .flac files meticulously with how I want albums/artists/album artists to sort, yet Jellyfin doesn't read/apply those tags to music in its library. And since there is no way to assign sort metadata in batches, it would be nice if Jellyfin used these tags in the metadata files.

bravely-beep:
I also think this should be reopened. These are commonly-used tags, and have a one-to-one mapping with Jellyfin metadata fields. I don't see why they shouldn't be supported.

florianjacob:
Still an issue.

nbabanov:
Hopefully https://github.com/jellyfin/jellyfin/pull/17048 will resolve the issue.

PR #17048, opened 2026-06-08
Add support for ID3 sort tags

nbabanov:
**Changes**

Adds support for reading embedded ID3/Vorbis sort tags from audio files so music is sorted by the values the tagger intended rather than the display name. When probing a track, the title sort tag (e.g. `TSOT`/`titlesort`) is applied to the song's forced sort name, and the album, artist, and album artist sort tags (`TSOA`/`albumsort`, `TSOP`/`artistsort`, `TSO2`/`albumartistsort`) are persisted on the song via three new `BaseItemEntity` columns (`SortAlbum`, `SortArtist`, `SortAlbumArtist`). 

The album and artist metadata services then aggregate these per-song values — picking the most common non-empty sort tag — onto the parent `MusicAlbum` and `MusicArtist`, with the artist matched against the song's primary album artist (then primary artist). All updates respect locked fields and the `ReplaceAllMetadata` refresh option, and a new EF Core migration (`AddAudioSortTags`) adds the backing columns.

> **TL;DR:**
> - Sort fields are now populated with the embedded data in the songs. Before they were empty. (https://github.com/jellyfin/jellyfin/issues/2908)
> - Sort fields are now searchable. 
>   - This is quite usefull if you have music in native language other than english and romanize it in the sort fields. This enables Siri search, because Siri does not support many languages.

**Code assistance**

- The Codex LLM created the migration - which is the biggest part of this PR.
- I tested the changes and they work perfectly on my machine.
- I am open to feedback and will fix any issues. 🙂 

**Issues**

https://github.com/jellyfin/jellyfin/issues/2908

https://features.jellyfin.org/posts/2450/utilize-embedded-sort-tags


Jellyfin.Server.Implementations/Item/BaseItemRepository.TranslateQuery.cs
```diff
@@ -179,12 +179,12 @@ public IQueryable<BaseItemEntity> TranslateQuery(
             {
                 cleanedSearchTerm = $"%{cleanedSearchTerm.Trim('%')}%";
                 var likeSearchTerm = $"%{originalSearchTerm.Trim('%')}%";
-                baseQuery = baseQuery.Where(e => EF.Functions.Like(e.CleanName!, cleanedSearchTerm) || (e.OriginalTitle != null && EF.Functions.Like(e.OriginalTitle, likeSearchTerm)));
+                baseQuery = baseQuery.Where(e => EF.Functions.Like(e.CleanName!, cleanedSearchTerm) || (e.OriginalTitle != null && EF.Functions.Like(e.OriginalTitle, likeSearchTerm)) || (e.SortName != null && EF.Functions.Like(e.SortName, cleanedSearchTerm)));
```
gnattu, line 182: I don't think this change should apply to all item types as it is useless for a lot of things.

PR #17063, opened 2026-06-10
Add playback rate support for audiobooks

jordanmichaelrushing:
**Changes**
- Add an optional `AudioPlaybackRate` query parameter to the universal-audio and HLS audio-segment endpoints.
- A non-1.0 rate forces transcoding (bypassing direct play) and applies ffmpeg's `atempo` filter chain, baking the speed into the stream.
- Audio-only HLS transcoding never added the `-af` audio filter argument; call `GetAudioFilterParam` so atempo (and any existing audio filters like downmix/volume) are applied.
- Remove the `StartTimeTicks` guard in `GetDynamicSegment`. Each segment's start comes from its own `runtimeTicks`, so a `StartTimeTicks` on a segment request is overwritten or ignored rather than something to reject.

**Code assistance**
I used Opus to identify the issue and where I would need to make the changes. I used Opus to also look over the changes to make any suggestions. It suggested to fallback of >2 and < 0.5.

**Issues**
https://github.com/jellyfin/jellyfin/issues/148 (This is already closed, but only because feature requests were moved elsewhere. Also they consensus was a frontend only fix, which unfortunately does not work for iOS.

**Associated Changes**
- [Web Change](https://github.com/jellyfin/jellyfin-web/pull/8024)

### DIFF
--- a/Jellyfin.Api/Controllers/DynamicHlsController.cs
+++ b/Jellyfin.Api/Controllers/DynamicHlsController.cs
@@ -1264,6 +1264,7 @@ public class DynamicHlsController : BaseJellyfinApiController
     /// <param name="context">Optional. The <see cref="EncodingContext"/>.</param>
     /// <param name="streamOptions">Optional. The streaming options.</param>
     /// <param name="enableAudioVbrEncoding">Optional. Whether to enable Audio Encoding.</param>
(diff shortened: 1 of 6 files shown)


jordanmichaelrushing:
Closing after talking with SevenRats to opt for a more general Audiobook Improvements PR to acompany this change along with my Audiobook Chapters work as well.

Issue #14837, opened 2025-09-22
Intermittent crashing when jellyfin tries to access a websocket that has been disposed


Sandermoen:
### This issue respects the following points:

- [x] This is a **bug**, not a question or a configuration issue; Please visit our [forum or chat rooms](https://jellyfin.org/contact/) first to troubleshoot with volunteers, before creating a report.
- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [x] I'm using an up to date version of Jellyfin Server stable, unstable or master; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Description of the bug

Keeping the server running normally in Unraid will make it crash after a random amount of days/hours, I am unsure if this bug is present in installs that are not within Docker and Unraid. I've had the Jellyfin instance running concurrently for a month before and then suddenly it crashes or it might crash within the hour, very random seeminlgy.
This issue has persisted through several Jellyfin versions and is still present now.

### Reproduction steps

1. Install Jellyfin through Unraid
2. Keep it running for a few days/weeks
3. Crashes with error Cannot access a disposed object.

### What is the current _bug_ behavior?

The server crashes after X amount of days with the error System.ObjectDisposedException: Cannot access a disposed object.
Seemingly trying to close a socket connection that doesn't exist anymore. Possibly a race condition?

### What is the expected _correct_ behavior?

The server shouldn't crash and should handle and check that a socket is in an acceptable state/isn't already closed when attempting to close the connection.

### Jellyfin Server version

10.10.0+

### Specify the build version

10.10.7

### Environment

```markdown
- OS: Unraid
- Linux Kernel: 6.1.106-Unraid
- Virtualization: Docker
- Clients: IOS, Android, Browser, Chromecast
- Browser: Chrome 140.0.7339.185
- FFmpeg Version: 7.0.2-Jellyfin
- Playback Method: Direct Play, Remux, Direct Stream, Transcode
- Hardware Acceleration: NVENC
- GPU Model: 1650 Super
- Plugins: None
- Reverse Proxy: None
- Base URL: none
- Networking: Host
[...]

PR #17079, opened 2026-06-12
Close sessions for lost WebSockets to prevent zombie SyncPlay groups

### LINKED ISSUE #15709: Jellyfin 10.11.4 Crashes When WebSocket Client Disconnects Abruptly (Unhandled ObjectDisposedException)
### Description of the bug

Jellyfin crashes entirely when a connected WebSocket client (Chrome browser) disconnects abruptly during shutdown or sleep. Instead of gracefully closing the session, Jellyfin throws an unhandled ObjectDisposedException originating from ManagedWebSocket, which bubbles up to Main and terminates the whole server process.

This occurs consistently when a client with the Dashboard open is shut down without completing a proper WebSocket close handshake.

This appears to be an unhandled WebSocket disconnect path that should be caught and ignored. A client dropping connection should never crash the server.

⸻

Environment
	•	Jellyfin Version: 10.11.4
	•	Platform: Unraid (Docker)
	•	Reverse Proxy: Nginx Proxy Manager (WebSocket headers enabled)
	•	Client Causing Crash: Google Chrome on Windows (192.168.1.6)
	•	Reproducibility: 100%

### Reproduction steps

1.	Open the Jellyfin Dashboard in Chrome on a Windows PC.
	2.	Shut down the PC or disconnect it from the network abruptly (e.g., Wi-Fi drop, sleep, NIC power-off).
	3.	Within seconds, Jellyfin logs show a WebSocket close error followed by an unhandled exception.
	4.	The Jellyfin server process crashes and restarts.

An abrupt client disconnect triggers a fatal, unhandled exception:

WS 192.168.1.6 error receiving data: The remote party closed the WebSocket connection without completing the close handshake
Unhandled exception: System.ObjectDisposedException: Cannot access a disposed object.
...
System.Net.WebSockets.ManagedWebSocket.WriteFrameToSendBuffer(...)
...
OperationCanceledException: The operation was canceled.

Jellyfin then terminates completely.

•	Jellyfin should gracefully handle unexpected WebSocket disconnects.
	•	Client session should close cleanly.
	•	Server should continue running without interruption.

### Jellyfin Server version

10.11.3

### Specify commit id

_No response_

_No response_

### Specify version number

10.11.4

### Specify the build version

10.11.4

### Environment

```markdown
- OS: Unraid
- Linux Kernel:
- Virtualization: No
- Clients: Windows 11 PC
- Browser: Google Chrome
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- GPU Model: AMD 7800xt
- Plugins:
- Reverse Proxy: Nginx Proxy Manager
- Base URL: https://jellyfin.gmadi.me
- Networking:
- Jellyfin Data Storage:
- Media Storage:
- External Integrations:
```

### Jellyfin logs

```shell
WS 192.168.1.6 error receiving data: The remote party closed the WebSocket connection without completing the close handshake
Unhandled exception: System.ObjectDisposedException: Cannot access a disposed object.
...
System.Net.WebSockets.ManagedWebSocket.WriteFrameToSendBuffer(...)
...
OperationCanceledException: The operation was canceled.



(Trimmed for clarity)

[WRN] Emby.Server.Implementations.HttpServer.WebSocketConnection:
WS 192.168.1.6 error receiving data: The remote party closed the WebSocket connection without completing the close handshake

[ERR] Microsoft.AspNetCore.Server.Kestrel: Unexpected exception in TimingPipeFlusher.FlushAsync

System.ObjectDisposedException: Cannot access a disposed object.
   at System.Net.WebSockets.ManagedWebSocket.WriteFrameToSendBuffer(...)
   at System.Net.WebSockets.ManagedWebSocket.SendFrameFallbackAsync(...)

[FATAL] Main: Unhandled Exception
OperationCanceledException: The operation was canceled.
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

Image
Image

### Additional information

_No response_

### LINKED ISSUE #17078: Sessions of silently dropped WebSockets are never closed, leaving zombie SyncPlay groups
### Description of the bug

When a client's WebSocket connection drops silently (mobile network loss, killed WebView, NAT timeout — no close frame is ever sent), the server detects the lost socket via its keep alive watchdog but never closes the session it belongs to.

The watchdog in `SessionWebSocketListener.KeepAliveSockets` only removes the lost socket from its own watchlist — there is a literal `// TODO: handle session relative to the lost webSocket` at that spot. `SessionEnded` is only ever raised through `WebSocketController.OnConnectionClosed`, which depends on the connection's `Closed` event, and that event is only raised when the receive loop in `WebSocketConnection.ReceiveAsync` terminates. On a silent TCP drop the pending `ReceiveAsync` hangs forever, so `Closed` never fires and the session stays "active" indefinitely (the dead socket remains in `WebSocketController._sockets` with `State == Open`).

For SyncPlay this is fatal: the dead session stays in `Group._participants` forever. `WaitingGroupState` has no timeout, so every playback start waits forever on the dead participant and all other group members get stuck on an infinite loading state. The group also never becomes empty, so it is never removed and clients with stored membership keep rejoining it.

### Reproduction steps

1. From a phone (or any client), sign in to the web client and
  create a SyncPlay group.
2. Enable airplane mode on the phone *without* closing the
  browser, so no WebSocket close frame is sent.
3. Wait at least 75 seconds (60s WebSocket-lost timeout + watchdog
  interval).
4. From a second client, open the SyncPlay menu and observe the
  group; also check Dashboard → Active sessions.

The server logs `Lost 1 WebSockets.` and nothing else: the session is never ended, it remains visible in the dashboard indefinitely, and the SyncPlay group survives with the dead participant in it. Any client that joins the group and starts playback gets stuck loading forever, because the group waits for a "ready" signal from the dead session. The only workaround is leaving the group manually or creating a new one.

Once the watchdog declares a socket lost, the session should be closed (when it has no other live connection), so `SessionEnded` fires and consumers such as SyncPlay can remove the dead participant and delete the now-empty group - the same cleanup that already happens on a clean disconnect.

### Jellyfin Server version

10.11.11

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

master / 10.11-dev

### Environment

```markdown
- OS: macOS 26.5.1 (server built and run from source with .NET SDK 10.0.301, `dotnet run --project Jellyfin.Server`)   
- Linux Kernel: none (macOS host)
- Virtualization: none (bare metal)
- Clients: Jellyfin Web (desktop browser + mobile browser on a phone over LAN Wi-Fi)
- Browser: <desktop: es. Safari 26 / Chrome 137> + <telefono: es. Safari on iOS 26 / Chrome on Android>
- FFmpeg Version: none installed (server started with `JELLYFIN_FFMPEG__NOVALIDATION=true`; not relevant - the bug reproduces without any playback)
- Playback Method: none (no media library needed to reproduce)
- Hardware Acceleration: none
- CPU Model: Apple M1 Pro
- GPU Model: Apple M1 Pro (integrated)
- Plugins: none
- Reverse Proxy: none (direct connection)
- Base URL: none
- Networking: host networking, plain HTTP on port 8096; desktop via localhost, phone via LAN IP over Wi-Fi
- Jellyfin Data Storage & Filesystem: local disk, APFS
- Media Storage & Filesystem: none (empty library)
- External Integrations: none
```

### Jellyfin logs

```shell
On an unpatched server the sequence ends after the watchdog detection — note the absence of any session/SyncPlay cleanup afterwards:

[INF]
Emby.Server.Implementations.Session.SessionWebSocketListener:
Sending ForceKeepAlive message to 1 inactive WebSockets.
[INF]
Emby.Server.Implementations.Session.SessionWebSocketListener: 
Lost 1 WebSockets.
  (no further log lines: no session end, no group removal)
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

Related issues describing the user-visible symptom: #4680, #8140.
This report pins down the server-side root cause. PR #<numero-PR> implements the fix together with an integration test that reproduces the scenario (the test fails without the fix).

Eneo-org:
**Changes**

The keep-alive watchdog in `SessionWebSocketListener` detected lost WebSockets but only removed them from its watchlist (the long-standing `// TODO: handle session relative to the lost webSocket`), so on a silent connection drop the receive loop hung forever, `Closed` never fired and the session never ended. The dead session then stayed in its SyncPlay group forever: `WaitingGroupState` has no timeout, so playback waited indefinitely on the zombie participant and the never-empty group kept being rejoined.

- `SessionWebSocketListener`: dispose lost connections so the pending receive loop aborts and `Closed` is raised; `WebSocketController` then releases the connection and `CloseIfNeededAsync` ends the session, letting the existing `SessionEnded` handler in `SyncPlayManager` drop the dead participant and delete the empty group. Sessions with other live connections are unaffected.
- `WebSocketConnection`: widen the receive-loop catch so `Closed` is also raised when the teardown surfaces as `ObjectDisposedException`/`OperationCanceledException` instead of `WebSocketException`.
- New integration test that opens a real WebSocket, creates a SyncPlay group, simulates the silent drop (backdated `LastKeepAliveDate`, no close frame), and asserts the group is removed. The test fails without this fix.

Also verified on real hardware: phone on Wi-Fi switched to airplane mode while in a SyncPlay group; within ~75s the server logs `Lost 1 WebSockets.` → session left group → `Group is empty, removing it.` A clean tab close still removes the group immediately, and a session with two sockets survives losing one of them.

**Code assistance**

The root-cause analysis, the fix and the integration test were developed with the assistance of an LLM coding agent (Claude Fable 5). All changes were reviewed by me, and verified locally with the full test suite plus manual end-to-end testing on a real device.

**Issues**

Fixes #17078

### DIFF
--- a/Emby.Server.Implementations/Session/SessionWebSocketListener.cs
+++ b/Emby.Server.Implementations/Session/SessionWebSocketListener.cs
@@ -246,8 +246,21 @@ namespace Emby.Server.Implementations.Session
                     _logger.LogInformation("Lost {0} WebSockets.", lost.Count);
                     foreach (var webSocket in lost)
                     {
-                        // TODO: handle session relative to the lost webSocket
                         RemoveWebSocket(webSocket);
+
+                        // The connection stopped answering keep-alives, so a close frame will
+                        // never arrive and the pending receive loop would hang forever, keeping
+                        // the session (and e.g. its SyncPlay group membership) alive. Disposing
+                        // the connection aborts the receive loop, which raises Closed and lets
+                        // the session end normally.
+                        try
+                        {
+                            webSocket.Dispose();
+                        }
+                        catch (Exception exception)
+                        {
+                            _logger.LogWarning(exception, "Error disposing lost WebSocket {0}.", webSocket);
+                        }
                     }
                 }
             }
--- a/Emby.Server.Implementations/Session/SessionWebSocketListener.cs
+++ b/Emby.Server.Implementations/Session/SessionWebSocketListener.cs
@@ -259,7 +259,7 @@ namespace Emby.Server.Implementations.Session
                         }
                         catch (Exception exception)
                         {
-                            _logger.LogWarning(exception, "Error disposing lost WebSocket {0}.", webSocket);
+                            _logger.LogWarning(exception, "Error disposing lost WebSocket from {RemoteEndPoint}.", webSocket.RemoteEndPoint);
                         }
                     }
                 }
--- a/tests/Jellyfin.Server.Integration.Tests/SyncPlayLostWebSocketTests.cs
+++ b/tests/Jellyfin.Server.Integration.Tests/SyncPlayLostWebSocketTests.cs
@@ -17,138 +17,124 @@ using MediaBrowser.Controller.Net;
 using Microsoft.Extensions.DependencyInjection;
 using Xunit;
 
-namespace Jellyfin.Server.Integration.Tests
+namespace Jellyfin.Server.Integration.Tests;
+
+public sealed class SyncPlayLostWebSocketTests : IClassFixture<JellyfinApplicationFactory>
 {
-    /// <summary>
-    /// Reproduces the SyncPlay "zombie group" scenario: a client joins a group over
-    /// a WebSocket and then silently stops responding (mobile network drop, killed
-    /// WebView, NAT timeout) without ever sending a close frame. The keep-alive
-    /// watchdog detects the lost socket; the session it belongs to must then be
-    /// closed so that SyncPlay drops the dead participant and removes the empty
-    /// group. If that does not happen, the group keeps waiting on the dead
-    /// participant forever and every other member is stuck on an infinite load.
-    /// </summary>
-    public sealed class SyncPlayLostWebSocketTests : IClassFixture<JellyfinApplicationFactory>
+    private readonly JellyfinApplicationFactory _factory;
+
+    public SyncPlayLostWebSocketTests(JellyfinApplicationFactory factory)
     {
-        private readonly JellyfinApplicationFactory _factory;
+        _factory = factory;
+    }
 
-        public SyncPlayLostWebSocketTests(JellyfinApplicationFactory factory)
+    [Fact]
+    public async Task LostWebSocket_EndsSession_And_RemovesEmptySyncPlayGroup()
+    {
+        var cancellationToken = TestContext.Current.CancellationToken;
+        var client = _factory.CreateClient();
+        var accessToken = await AuthHelper.CompleteStartupAsync(client);
+        client.DefaultRequestHeaders.AddAuthHeader(accessToken);
+
+        var wsClient = _factory.Server.CreateWebSocketClient();
+        wsClient.ConfigureRequest = request =>
+            request.Headers.Authorization = AuthHelper.DummyAuthHeader + $", Token={accessToken}";
+
+        var webSocket = await wsClient.ConnectAsync(
+            new UriBuilder(_factory.Server.BaseAddress)
+            {
+                Scheme = "ws",
+                Path = "websocket"
+            }.Uri,
+            cancellationToken);
+
+        _ = DrainAsync(webSocket, cancellationToken);
+
+        var watched = await WaitForWatchedWebSocketsAsync(TimeSpan.FromSeconds(10), cancellationToken);
+        var connection = Assert.Single(watched);
+
+        using var createResponse = await client.PostAsync(
+            "SyncPlay/New",
+            JsonContent.Create(new NewGroupRequestDto { GroupName = "ZombieGroupRepro" }, options: JsonDefaults.Options),
+            cancellationToken);
+        Assert.Equal(HttpStatusCode.OK, createResponse.StatusCode);
+        Assert.Equal(1, await WaitForGroupCountAsync(client, 1, TimeSpan.FromSeconds(10), cancellationToken));
+
+        connection.LastKeepAliveDate = DateTime.UtcNow - TimeSpan.FromSeconds(180);
+
+        var groupCount = await WaitForGroupCountAsync(client, 0, TimeSpan.FromSeconds(45), cancellationToken);
+        Assert.True(
+            groupCount == 0,
+            $"SyncPlay group still listed {groupCount} group(s) after the WebSocket was lost: "
+            + "the keep-alive watchdog removed the socket from its watchlist without closing "
+            + "the session, leaving a zombie participant in the group (SessionWebSocketListener).");
+    }
+
+    private static async Task DrainAsync(WebSocket webSocket, CancellationToken cancellationToken)
+    {
+        var buffer = new byte[4096];
+        try
         {
-            _factory = factory;
+            while (webSocket.State == WebSocketState.Open)
+            {
+                await webSocket.ReceiveAsync(buffer, cancellationToken);
+            }
         }
-
-        [Fact]
-        public async Task LostWebSocket_EndsSession_And_RemovesEmptySyncPlayGroup()
+        catch
         {
-            var client = _factory.CreateClient();
-            var accessToken = await AuthHelper.CompleteStartupAsync(client);
-            client.DefaultRequestHeaders.AddAuthHeader(accessToken);
-
-            var wsClient = _factory.Server.CreateWebSocketClient();
-            wsClient.ConfigureRequest = request =>
-                request.Headers.Authorization = AuthHelper.DummyAuthHeader + $", Token={accessToken}";
-
-            var webSocket = await wsClient.ConnectAsync(
-                new UriBuilder(_factory.Server.BaseAddress)
-                {
-                    Scheme = "ws",
-                    Path = "websocket"
-                }.Uri,
-                TestContext.Current.CancellationToken);
-
-            // Drain server messages (ForceKeepAlive etc.) but never answer them,
-            // like a client whose connection went dark.
-            _ = DrainAsync(webSocket, TestContext.Current.CancellationToken);
-
-            var watched = await WaitForWatchedWebSocketsAsync(TimeSpan.FromSeconds(10));
-            var connection = Assert.Single(watched);
-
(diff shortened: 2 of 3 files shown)


manueloverride:
This should also fix #14837

Emby.Server.Implementations/Session/SessionWebSocketListener.cs
```diff
@@ -246,8 +246,21 @@ private async void KeepAliveSockets(object? o, EventArgs? e)
...
+                        // the session (and e.g. its SyncPlay group membership) alive. Disposing
+                        // the connection aborts the receive loop, which raises Closed and lets
+                        // the session end normally.
+                        try
+                        {
+                            webSocket.Dispose();
+                        }
+                        catch (Exception exception)
+                        {
+                            _logger.LogWarning(exception, "Error disposing lost WebSocket {0}.", webSocket);
```
cvium: Two things:
1. We use structured logging, so it should be something like `{WebSocket}` and not `{0}`.
2. As far as I can tell, WebSocketConnection does not implement ToString, so won't it just log the name of the class?

Eneo-org: Good catch, fixed — switched to a named placeholder and now logging RemoteEndPoint instead of the connection object (which doesn't override ToString()).

cvium: Thanks, Claude 🙄

cvium (approved):

tests/Jellyfin.Server.Integration.Tests/SyncPlayLostWebSocketTests.cs
```diff
@@ -0,0 +1,154 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Emby.Server.Implementations.Session;
+using Jellyfin.Api.Models.SyncPlayDtos;
+using Jellyfin.Extensions.Json;
+using MediaBrowser.Controller.Net;
+using Microsoft.Extensions.DependencyInjection;
+using Xunit;
+
+namespace Jellyfin.Server.Integration.Tests
```
Bond-009: Use file scoped namespaces for new files

tests/Jellyfin.Server.Integration.Tests/SyncPlayLostWebSocketTests.cs
```diff
@@ -0,0 +1,154 @@
...
+
+        private static async Task<int> WaitForGroupCountAsync(HttpClient client, int expected, TimeSpan timeout)
+        {
+            var stopwatch = Stopwatch.StartNew();
+            var count = -1;
+            while (stopwatch.Elapsed < timeout)
+            {
+                using var response = await client.GetAsync("SyncPlay/List");
+                response.EnsureSuccessStatusCode();
+                using var document = JsonDocument.Parse(await response.Content.ReadAsStringAsync());
```
Bond-009: Parse the stream instead of converting to a string

tests/Jellyfin.Server.Integration.Tests/SyncPlayLostWebSocketTests.cs
```diff
@@ -0,0 +1,154 @@
...
+                catch (InvalidOperationException)
+                {
+                    // The watchdog mutated the set during enumeration; retry.
+                }
+
+                await Task.Delay(100);
+            }
+        }
+
+        private static async Task<int> WaitForGroupCountAsync(HttpClient client, int expected, TimeSpan timeout)
```
Bond-009: This should take a cancellation token

tests/Jellyfin.Server.Integration.Tests/SyncPlayLostWebSocketTests.cs
```diff
@@ -0,0 +1,154 @@
...
+                    await webSocket.ReceiveAsync(buffer, cancellationToken);
+                }
+            }
+            catch
+            {
+                // The server tears the connection down once the watchdog gives up on it.
+            }
+        }
+
+        private async Task<IReadOnlyList<IWebSocketConnection>> WaitForWatchedWebSocketsAsync(TimeSpan timeout)
```
Bond-009: This should also take a cancellation token

Bond-009 (changes requested):

Emby.Server.Implementations/Session/SessionWebSocketListener.cs
Eneo-org: I'm sorry for my ai slop reply. It's my firs pr and I don't want to so something wrong. It won't happen again.

tests/Jellyfin.Server.Integration.Tests/SyncPlayLostWebSocketTests.cs
Eneo-org: I should have fix it.

tests/Jellyfin.Server.Integration.Tests/SyncPlayLostWebSocketTests.cs
```diff
@@ -0,0 +1,140 @@
...
+            new UriBuilder(_factory.Server.BaseAddress)
+            {
+                Scheme = "ws",
+                Path = "websocket"
+            }.Uri,
+            cancellationToken);
+
+        _ = DrainAsync(webSocket, cancellationToken);
+
+        var watched = await WaitForWatchedWebSocketsAsync(TimeSpan.FromSeconds(10), cancellationToken);
```
Bond-009, line 53: Now that they take a cancellation token we can create a linked token with the timeout.
That way we don't have the hacky stopwatch timeout

Bond-009 (approved):

tests/Jellyfin.Server.Integration.Tests/SyncPlayLostWebSocketTests.cs
crobibero: @Bond-009 is this a suggestion to implement before merging, or something that should be done as a follow-up?

Bond-009: It's non-blocking, feel free to merge

Issue #17084, opened 2026-06-12
/Items?Recursive=true&AlbumIds=<> does not return songs where song metadata "Album" and album name mismatch


matt-teahan:
### Description of the bug

When making an API request to `/Items?AlbumIds=<id>&Recursive=true` I found that songs which are under that album on jellyfin web ui may not be returned by that request - only if the Album name in the song's metadata exactly matches the name of the album on jellyfin. If it is blank or anything else it will not be returned.

I believe this is is a bug since this should not be using the metadata (which may not be filled out correctly) when it was requested to find all songs inside of an Album by id. It would be acceptable if the query was by by an album name but not by id.

### Reproduction steps

1. Sign into jellyfin web ui with an admin account
2. Create a music media library
3. Add an album with one song by placing it in the library and ensure it is visible in the web ui
4. Edit the metadata of the song via the web ui and set the "Album" to an empty string
5. Make note of the album id in the URL of the web UI
6. Exit the album, in the web ui before clicking on the album open devtools then click on the album
7. Get any /Items request and copy as curl, paste into postman or similar (or just run the curl command) and modify the URL to be a GET request for `/Items?AlbumIds=<id>&Recursive=true` where <id> is from step 5
8. Observe that the returned JSON is empty
9. Go back to step 4 but edit the metadata to exactly match the name of the jellyfin album
10. Observe that the returned JSON contains the song

### What is the current _bug_ behavior?

Items is empty:
```
{
    "Items": [],
    "TotalRecordCount": 0,
    "StartIndex": 0
}
```

### What is the expected _correct_ behavior?

Items is populated
```
{
    "Items": [
        {
            "Name": "Song",
            .... omitted for brevity ....
        }
    ],
    "TotalRecordCount": 1,
    "StartIndex": 0
}
```

### Jellyfin Server version

Master

### Specify commit id

dd42a121c43721c8984ba0026d6fbed4a526d01f

### Specify the build version

Release

### Environment

```markdown
- OS: Windows 10
- Linux Kernel: none
- Virtualization: none
- Clients: Browser, API via postman
- Browser: Firefox 151.0.4
- FFmpeg Version: Not Applicable
- Playback Method: Not Applicable
- Hardware Acceleration: Not Applicable
- CPU Model: Ryzen 7 5700X3D
- GPU Model: Nvidia RTX 4070 SUPER
- Plugins: None
- Reverse Proxy: None
- Base URL: none
- Networking: Host
[...]

PR #17085, opened 2026-06-12
Fix AlbumIds filtering by Name instead of by Id

matt-teahan:
**Changes**
Replaces query that compares a song's Album metadata string to the Album's name (which was fetched via the id) to instead fetch songs that have a parent album which has an id of a queried id

**Code assistance**
AI not used

**Issues**
#17084

### DIFF
--- a/Jellyfin.Server.Implementations/Item/BaseItemRepository.TranslateQuery.cs
+++ b/Jellyfin.Server.Implementations/Item/BaseItemRepository.TranslateQuery.cs
@@ -586,8 +586,7 @@ public sealed partial class BaseItemRepository
 
         if (filter.AlbumIds.Length > 0)
         {
-            var subQuery = context.BaseItems.WhereOneOrMany(filter.AlbumIds, f => f.Id);
-            baseQuery = baseQuery.Where(e => subQuery.Any(f => f.Name == e.Album));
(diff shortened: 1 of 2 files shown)


IDisposable (approved):

Bond-009 (approved):

Shadowghost (approved):

Jellyfin.Server.Implementations/Item/BaseItemRepository.TranslateQuery.cs
```diff
@@ -586,8 +586,7 @@ public IQueryable<BaseItemEntity> TranslateQuery(
 
         if (filter.AlbumIds.Length > 0)
         {
-            var subQuery = context.BaseItems.WhereOneOrMany(filter.AlbumIds, f => f.Id);
-            baseQuery = baseQuery.Where(e => subQuery.Any(f => f.Name == e.Album));
+            baseQuery = baseQuery.Where(e => filter.AlbumIds.Contains((Guid)e.ParentId!));
```
Bond-009:
```suggestion
            baseQuery = baseQuery.Where(e => e.ParentId.HasValue && filter.AlbumIds.Contains(e.ParentId.Value));
```

matt-teahan: Ack and applied the suggestion. Tested and has the same behaviour.

PR #17086, opened 2026-06-13
Fix missing person thumbnails after initial library scan (issue #8103)

### LINKED ISSUE #8103: [Issue]: Displaying actor photos with local metadata

deltonio2:
**Changes**

Convert remote primary images for person entities to local images during image updates.

When a library is initially scanned with metadata stored locally, person entities can be created with remote image URLs that are not downloaded until the person's page is opened. This results in missing actor thumbnails in movie views despite the image being available.

This change eagerly converts remote person images to local images during the image update process, making actor thumbnails available immediately after the initial scan.

**Code assistance**

GitHub Copilot was used to help identify the relevant code path and propose an initial implementation. The generated code was reviewed, adapted, built, and validated against the reported issue before submission.

**Issues**

Fixes #8103

### DIFF
--- a/Emby.Server.Implementations/Library/LibraryManager.cs
+++ b/Emby.Server.Implementations/Library/LibraryManager.cs
@@ -3624,6 +3624,29 @@ namespace Emby.Server.Implementations.Library
                     personEntity.DateLastSaved = DateTime.UtcNow;
 
                     CreateItems([personEntity], null, CancellationToken.None);
+
+                    // Download person image stubs to ensure images are available during initial scan
+                    if (itemUpdateType == ItemUpdateType.ImageUpdate && personEntity.HasImage(ImageType.Primary))
+                    {
+                        var primaryImage = personEntity.GetImageInfo(ImageType.Primary, 0);
+                        if (primaryImage != null && !string.IsNullOrWhiteSpace(primaryImage.Path)
+                            && primaryImage.Path.StartsWith("http", StringComparison.OrdinalIgnoreCase))
+                        {
+                            try
+                            {
+                                cancellationToken.ThrowIfCancellationRequested();
+                                await ConvertImageToLocal(personEntity, primaryImage, 0, removeOnFailure: false).ConfigureAwait(false);
+                            }
+                            catch (OperationCanceledException)
+                            {
+                                throw;
+                            }
+                            catch (Exception ex)
+                            {
+                                _logger.LogDebug(ex, "Failed to download person image for {Name}, will use lazy loading", personEntity.Name);
+                            }
+                        }
+                    }
                 }
             }
         }


Emby.Server.Implementations/Library/LibraryManager.cs
```diff
@@ -3624,6 +3624,29 @@ private async Task SavePeopleMetadataAsync(IEnumerable<PersonInfo> people, Cance
...
+                            catch (OperationCanceledException)
+                            {
+                                throw;
+                            }
+                            catch (Exception ex)
+                            {
+                                _logger.LogDebug(ex, "Failed to download person image for {Name}, will use lazy loading", personEntity.Name);
+                            }
+                        }
+                    }
```
Shadowghost, line 3649:
```suggestion
                    if (itemUpdateType == ItemUpdateType.ImageUpdate)
                    {
                        await UpdateImagesAsync(personEntity).ConfigureAwait(false);
                    }
```

I think this should be enough to fix it

Shadowghost:
@deltonio2 could you please check my comment and fix the formatting issues, then we can continue with getting this in.

deltonio2:
@Shadowghost I'm sorry, but I have to admit that my expertise is very limited when it comes to a project as large as Jellyfin; I can't determine whether it makes sense to implement only part of the proposed fix.

When I ran the AI prompt I used to generate the suggestion, I got the following response.

> 
> This suggestion is partially relevant, but it isn’t sufficient on its own to fix the bug. Let me explain.
> 
> Why this proposal seems relevant
> 
> - It calls `UpdateImagesAsync(personEntity)`, which will iterate through the person’s images and attempt to convert them from URL paths to local files using `ConvertImageToLocal`.
> - This is consistent with the mechanism already used elsewhere in Jellyfin to “make images exist” on disk.
> 
> Why this isn’t sufficient on its own
> 
> - `UpdateImagesAsync` doesn’t necessarily download an image if it doesn’t need to be refreshed according to `ImageNeedsRefresh`.
> - However, for a stub that has just been created, the image may not be considered “outdated” at the time of the call; the code depends on the state of the image metadata and the presence of the local file.
> - In practice, the method is designed to refresh existing images, not to force the download of a newly added stub.
> 
> Why the More Robust Solution Is Better
> 
> - The code we added to `SavePeopleMetadataAsync` is more straightforward: it explicitly detects the case where a person’s image is initially stored as an HTTP URL and immediately triggers its conversion to a local file.
> - This precisely covers the initial scan/NFO scenario, which is at the heart of bug #8103.
> 
> In short
> 
> - Yes, this suggestion is relevant as an image refresh mechanism.
> - No, it is not sufficient on its own to guarantee the expected behavior regarding the bug, because it relies on the existing image refresh mechanism and does not explicitly force the download of the stub at the time of the initial scan.
> 
> Conclusion
> 
> - `UpdateImagesAsync(personEntity)` can be a good addition, but it is not a minimal and reliable fix on its own.
> - The correct approach remains to download the images when the person is created or added in `SavePeopleMetadataAsync`, and then optionally call `UpdateImagesAsync` if a more comprehensive refresh is desired.

Issue #17065, opened 2026-06-10
GET /Trailers returns HTTP 500 (unhandled NullReferenceException)


NaivEPoi:
### Description of the bug

`GET /Trailers` throws an unhandled `System.NullReferenceException` and returns HTTP 500 for a normal (non-admin) authenticated user. The stack shows `TrailersController.GetTrailers` delegating to `ItemsController.GetItems`, which calls `User.GetIsApiKey()`; `ClaimsPrincipalExtensions.GetClaimValue` then dereferences a null `ClaimsPrincipal`. It reproduces when the request authenticates via the `api_key` query parameter (the `ClaimsPrincipal`/`User` appears to be null in the delegated `GetItems` call). No admin rights are required, so any authenticated user can fault the request handler.

### Reproduction steps

1. Run Jellyfin 10.11.11 (official `jellyfin/jellyfin` Docker image) and create a normal, non-admin user; obtain that user's access token.
2. Call the endpoint authenticating with that token via the query parameter:
   ```
   GET /Trailers?api_key=<non-admin access token>
   ```

### What is the current _bug_ behavior?

HTTP 500 ("Error processing request."). The server log records an unhandled `NullReferenceException` originating in `ClaimsPrincipalExtensions.GetClaimValue` via `GetIsApiKey`, reached from `ItemsController.GetItems` ← `TrailersController.GetTrailers`.

### What is the expected _correct_ behavior?

HTTP 200 with the trailers list (or an empty list when there are none). The endpoint must not raise an unhandled exception; the `ClaimsPrincipal` should be propagated/null-checked in the `GetTrailers → GetItems` path.

### Jellyfin Server version

10.11.11

### Specify the build version

10.11.11

### Environment

```markdown
- OS: Debian GNU/Linux 13 (trixie) — inside the official jellyfin/jellyfin Docker image
- Linux Kernel: 5.15.0-181-generic (host)
- Virtualization: Docker (official jellyfin/jellyfin image)
- Clients: none — reproduced via direct REST API calls (curl)
- Browser: none
- FFmpeg Version: 7.1.3-Jellyfin (bundled; not involved in this bug)
- Playback Method: N/A
- Hardware Acceleration: none
- CPU Model: N/A
- GPU Model: N/A
- Installed Plugins: default bundled only — TMDb, Studio Images, OMDb, MusicBrainz, AudioDB
- Reverse Proxy: none
- Base URL: none (default)
[...]

PR #17094, opened 2026-06-14
Deprecate the redundant /Trailers endpoint

moontwister:
**Changes**
`GET /Trailers` is a thin alias for `GET /Items` with `includeItemTypes=Trailer` — it just forwards to the injected `ItemsController`. Per the review on this PR, the agreed direction is to deprecate it rather than keep maintaining the delegation. I mark the action `[Obsolete]` so it is flagged as `deprecated` in the OpenAPI spec; clients should call the `GetItems` operation with `includeItemTypes=Trailer` instead. No behaviour change beyond the deprecation flag.

**Code assistance**
Used Claude Code. Verified against a running server that the generated OpenAPI spec now marks `GET /Trailers` as `deprecated: true` (operationId `GetTrailers`). No unit test is added: this is an API-surface (attribute) change, and the OpenAPI diff on the PR shows the deprecation.

**Issues**
Re #17065 — the endpoint is a redundant alias, so deprecating it steers clients to the working `GetItems` operation.

### DIFF
--- a/Jellyfin.Api/Controllers/TrailersController.cs
+++ b/Jellyfin.Api/Controllers/TrailersController.cs
@@ -122,6 +122,7 @@ public class TrailersController : BaseJellyfinApiController
     /// <returns>A <see cref="QueryResult{BaseItemDto}"/> with the trailers.</returns>
     [HttpGet]
     [ProducesResponseType(StatusCodes.Status200OK)]
+    [Obsolete("Use GetItems with includeItemTypes=Trailer instead.")]
     public async Task<ActionResult<QueryResult<BaseItemDto>>> GetTrailers(
         [FromQuery] Guid? userId,
         [FromQuery] string? maxOfficialRating,


nielsvanvelzen (changes requested):
I believe the better fix here would be to deprecate this endpoint as it's a complete alias for a different operation. Clients can call the getItems operation directly instead.
If we do want to keep the getTrailers operation we should just make it non-dependent on a different controller.

moontwister:
Thanks for the review! I agree that deprecating the endpoint is reasonable longer-term, since it's effectively an alias for `getItems` with `includeItemTypes=Trailer`.

One thing worth flagging: deprecation alone won't resolve this issue. As long as the endpoint remains callable, it still returns HTTP 500 for any authenticated request because the injected `ItemsController` has no `ControllerContext` (and therefore a null `User`). So existing clients would keep hitting the crash throughout the deprecation period.

So I see two paths:
1. Keep this minimal fix to stop the 500 now, and handle deprecation/decoupling separately.
2. Decouple `GetTrailers` from `ItemsController` in this PR — though since `GetItems` is a ~430-line controller action with several injected dependencies, that's effectively a refactor to extract the shared logic into a service.

Happy to go either way — what would you (and other maintainers) prefer?

nielsvanvelzen:
I believe deprecating without any further changes is the way forward. The replacement for clients is easy enough.

moontwister:
Agreed — that makes sense. I've updated this PR to just deprecate the endpoint: `GetTrailers` is now marked `[Obsolete]` and shows as `deprecated` in the OpenAPI spec, with no other changes. Clients should use the Items endpoint with `includeItemTypes=Trailer` instead. Thanks for the guidance!

nielsvanvelzen (approved):

Shadowghost (approved):

Bond-009 (approved):

PR #17095, opened 2026-06-14
Fix identify on locked items and BoxSet ID references

Shadowghost:
While debugging why manually created BoxSets don't return anything when identifying I encountered two issues:
* By default they are marked as Locked, which prevents identify from returning anything.
* BoxSets can have Tmdb IDs but right now we don't allow it due to it only being not assignable

**Changes**
* Allow remote searches to run and return data, even on locked items
* Use the existing `TmdbBoxSetExternalId` to handle the BoxSet's own Tmdb Id
* Create new `TmdbCollectionExternalId` that handles the Item -> BoxSet relation

**Code assistance**
None


IDisposable (approved):
This  will greatly help my Mind the Gaps plugin :)

PR #17096, opened 2026-06-14
Fall back to the default transcode path when it is unwritable

### LINKED ISSUE #16766: Cannot set Transcode path
### Description of the bug

Setting the "Transcode path" will break Jellyfin server and stop it from starting. 

### Reproduction steps

1. On a Linux system, mount a different drive to /media/storage, which is owned by root:root (my use case is: main installation is on a somewhat small drive, so transcoding shouldn't go there)
2. Create /media/storage/transcode folder
3. In Web UI, set Transcoding path to this new folder
4. --> Any playback will break
5. Set transcode folder to /media/storage (which is also root:root, but contains all media files and access to those works)
6. Does not work
7. Restart Jellyfin
8. Jellyfin won't start up, giving you "access denied" error in the logs (full stack trace in log section)
9. Change back to /media/storage/transcode, but give that folder owner jellyfin:jellyfin (as used, for example on /var/log/jellyfin)
10. Restart Jellyfin server, issue persists

There seems to be no way to set the transcoding path (correctly). Once set incorrectly, Jellyfin server will not start up (e.g. by throwing a warning an defaulting to the default transcoding folder). There is no documentation on how to create a custom transcoding folder with regards to access rights or necessary information, nor where the default folder resides.
When set incorrectly and rebooting Jellyfin, it's hard to find the config file in which to manually change it (and for others with the problem, mine was at /etc/jellyfin/encoding.xml)

If special access rights are required, they should be documented.
Once set (and I'm assuming that at least jellyfin:jellyfin should be correct), the given transcoding path should work.
If it does not work for whatever reason, it should throw a warning and default back to the default setting - which enables changing it in the web UI.

### Jellyfin Server version

10.11.8

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

Release

### Environment

```markdown
- OS: Arch Linux (in an LXC Container in Proxmox)
- Linux Kernel: Linux 7.0.0-3-pve
- Virtualization: LXC Container in Proxmox
- Clients: Web UI
- Browser: Firefox
- FFmpeg Version: n8.1
- Playback Method: Web UI play on a movie and a song
- Hardware Acceleration: Default Settings after enabling Hardware Acceleration
- CPU Model: Ryzen 5 8500G
- GPU Model: Intel Arc A310
- Plugins: AudioDB, Intro Skipper, MusicBrainz, OMDb, Studio Images, TheTVDB, TMDb
- Reverse Proxy: Nginx Proxy Manager
- Base URL:
- Networking: Issue on both LAN and via Internet
- Media Storage & Filesystem: ext4
- External Integrations:
```

### Jellyfin logs

```shell
System.UnauthorizedAccessException: Access to the path '/media/storage1/transcode/.jellyfin-transcode' is denied.
 ---> System.IO.IOException: Permission denied
   at Interop.ThrowExceptionForIoErrno(ErrorInfo errorInfo, String path, Boolean isDirError)
   at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(String path, OpenFlags flags, Int32 mode, Boolean failForSymlink, Boolean& wasSymlink, Func`4 createOpenException)
   at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(String fullPath, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallocationSize, UnixFileMode openPermissions, Int64& fileLength, UnixFileMode& filePermissions, Boolean failForSymlink, Boolean& wasSymlink, Func`4 createOpenException)
   at System.IO.File.OpenHandle(String path, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallocationSize)
   at Jellyfin.Extensions.FileHelper.CreateEmpty(String path)
   at Emby.Server.Implementations.AppBase.BaseApplicationPaths.CheckOrCreateMarker(String path, String markerName, Boolean recursive)
   at Emby.Server.Implementations.AppBase.BaseApplicationPaths.CreateAndCheckMarker(String path, String markerName, Boolean recursive)
   at MediaBrowser.Common.Configuration.EncodingConfigurationExtensions.GetTranscodePath(IConfigurationManager configurationManager)
   at Jellyfin.Server.Program.StartServer(IServerApplicationPaths appPaths, StartupOptions options, IConfiguration startupConfig)
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

moontwister:
**Changes**
A transcoding temp path that exists but is not writable by the Jellyfin user bricks the server: `GetTranscodePath()` creates a `.jellyfin-transcode` marker in that directory, and the unhandled `UnauthorizedAccessException` it throws aborts startup, so the server never starts. Playback also breaks, since transcoding resolves its path the same way. The only recovery is to hand-edit `encoding.xml`. This is the issue in #16766. I make `GetTranscodePath` fall back to the default transcode path when the configured one can't be used, so the server starts and transcoding keeps working. The configured value is left untouched, so a transient cause (for example a drive that isn't mounted yet at boot) self-heals on the next start. Startup logs an error telling the admin the configured path is being ignored and how to fix it.

**Code assistance**
Used Claude Code for the diagnosis, fix and tests. Added `EncodingConfigurationExtensionsTests` covering the unwritable path (falls back to default, config untouched), a writable path (returned as-is), and no configured path (default). I also verified it end-to-end against a real server: I saved a transcode path, made the directory unwritable, and restarted. On master the server fails to start; with this change it starts, logs the warning, keeps the configured path in `encoding.xml`, and serves requests.

**Issues**
Fixes #16766

### DIFF
--- a/Jellyfin.Server/Program.cs
+++ b/Jellyfin.Server/Program.cs
@@ -187,7 +187,14 @@ namespace Jellyfin.Server
                  * Initialize the transcode path marker so we avoid starting Jellyfin in a broken state.
                  * This should really be a part of IApplicationPaths but this path is configured differently.
                  */
-                _ = appHost.ConfigurationManager.GetTranscodePath();
+                var configuredTranscodePath = appHost.ConfigurationManager.GetEncodingOptions().TranscodingTempPath;
+                var transcodePath = appHost.ConfigurationManager.GetTranscodePath();
+                if (!string.IsNullOrEmpty(configuredTranscodePath) && !string.Equals(configuredTranscodePath, transcodePath, StringComparison.Ordinal))
+                {
+                    _logger.LogError(
+                        "Configured transcode path \"{Path}\" is not writable; using the default instead. Fix its permissions or set a writable path in the dashboard",
+                        configuredTranscodePath);
+                }
 
                 // Re-use the host service provider in the app host since ASP.NET doesn't allow a custom service collection.
                 appHost.ServiceProvider = _jellyfinHost.Services;
--- a/MediaBrowser.Common/Configuration/EncodingConfigurationExtensions.cs
+++ b/MediaBrowser.Common/Configuration/EncodingConfigurationExtensions.cs
@@ -18,24 +18,40 @@ namespace MediaBrowser.Common.Configuration
             => configurationManager.GetConfiguration<EncodingOptions>("encoding");
 
         /// <summary>
-        /// Retrieves the transcoding temp path from the encoding configuration, falling back to a default if no path
-        /// is specified in configuration. If the directory does not exist, it will be created.
+        /// Retrieves the transcoding temp path from the encoding configuration. Falls back to a default
+        /// when no path is configured, or when the configured path exists but cannot be used (for example
+        /// it is not writable), so a bad path does not stop the server or transcoding. The directory is
+        /// created if it does not exist.
         /// </summary>
         /// <param name="configurationManager">The configuration manager.</param>
         /// <returns>The transcoding temp path.</returns>
-        /// <exception cref="UnauthorizedAccessException">If the directory does not exist, and the caller does not have the required permission to create it.</exception>
-        /// <exception cref="NotSupportedException">If there is a custom path transcoding path specified, but it is invalid.</exception>
-        /// <exception cref="IOException">If the directory does not exist, and it also could not be created.</exception>
+        /// <exception cref="NotSupportedException">If a configured transcoding path is invalid.</exception>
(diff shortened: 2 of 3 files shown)


gnattu:
I don't think it is a good idea. Write to a folder that the user does not intend to (and potentially takes huge amounts of space) just to make it appears to work is not better than directly fail and let the user to fix the config.

moontwister:
Fair point — silently falling back to a different folder isn't the right behaviour, and I don't have a cleaner approach that avoids that trade-off. Closing this. Thanks for the feedback!

PR #17099, opened 2026-06-14
Follow native interoperability best practices

Bond-009:
https://learn.microsoft.com/en-us/dotnet/standard/native-interop/best-practices

### DIFF
--- a/MediaBrowser.MediaEncoding/Encoder/ApplePlatformHelper.cs
+++ b/MediaBrowser.MediaEncoding/Encoder/ApplePlatformHelper.cs
@@ -1,9 +1,11 @@
 #pragma warning disable CA1031
 
 using System;
+using System.Buffers;
 using System.Linq;
 using System.Runtime.InteropServices;
 using System.Runtime.Versioning;
+using System.Text;
 using Microsoft.Extensions.Logging;
 
 namespace MediaBrowser.MediaEncoding.Encoder;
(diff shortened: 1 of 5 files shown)


Shadowghost:
@Bond-009 this replace #17098 if I understand correctly?

Bond-009:
> @Bond-009 this replace #17098 if I understand correctly?

Correct, we no longer create the c string manually

Shadowghost (approved):

MediaBrowser.MediaEncoding/Encoder/ApplePlatformHelper.cs
```diff
@@ -12,43 +14,38 @@ namespace MediaBrowser.MediaEncoding.Encoder;
...
-        if (osStatus != 0)
+        int osStatus = sysctlbyname(name, Span<byte>.Empty, ref length, IntPtr.Zero, 0);
+        if (osStatus != 0 || length == 0)
         {
-            throw new NotSupportedException($"Failed to get sysctl value for {System.Text.Encoding.UTF8.GetString(name)} with error {osStatus}");
+            throw new NotSupportedException($"Failed to get sysctl value for {name} with error {osStatus}");
         }
 
-        IntPtr buffer = Marshal.AllocHGlobal(length.ToInt32());
+        byte[] buffer = ArrayPool<byte>.Shared.Rent((int)length);
```
gnattu, line 31: I think we should use `(checked((int)length)` here instead, per dotnet doc?

Bond-009: I just replaced `IntPtr` with `nuint` instead

gnattu (approved):

PR #17102, opened 2026-06-14
Clear resume position when importing a watched item from NFO

### LINKED ISSUE #12054: Episodes marked as watched show in "Continue watching"
### Please describe your bug

I have multiple episodes of a show that won't leave the "Continue Watching" area even tho there is no watch state on the media (meaning there is nothing to continue in those videos) and they are already marked as watched. (They are from a season before the one I'm currently watching)

Marking them as watched manually does not help and the episode I actually am at right now does not show up so I have to go to the show and manually search the next episode after a break.

### Reproduction Steps

1. Watch a episode
2. Find it not marked as watched
3. Manually marking it as such also doesn't do something

### Jellyfin Version

10.9.0

### if other:

10.9.5

### Environment

```markdown
- OS: Archlinux
- Linux Kernel: -
- Virtualization: None
- Clients: Every client (Web/Android/3rdparty)
- Browser: -
- FFmpeg Version: -
- Playback Method: Direct
- Hardware Acceleration: Quicksync
- GPU Model: Intel UHD
- Plugins: none
- Reverse Proxy: nginx
- Base URL: none
- Networking: Host
- Storage: NFS
```

### Jellyfin logs

```shell
No relevant log
```

### FFmpeg logs

_No response_

_No response_

Screenshot_20240609_214908_Jellyfin

### Code of Conduct

moontwister:
**Changes**
Fully watched movies and episodes can stay in Continue Watching forever; the common factor in #12054 is NFO watch-status tracking. Continue Watching is the resume query, and "resumable" is defined purely as `PlaybackPositionTicks > 0` (`BaseItemRepository.TranslateQuery`); the resume endpoints set `IsResumable = true` but never `IsPlayed = false`. So any item that is both `Played = true` and `PlaybackPositionTicks > 0` stays there. Normal playback cannot create that state (`UpdatePlayState` zeroes the position on completion, and mark-as-watched uses `MarkPlayed(resetPosition: true)`), but the NFO import path set `Played = true` while leaving any leftover position untouched. I reset `PlaybackPositionTicks` to 0 when an NFO marks an item watched, matching `MarkPlayed(resetPosition: true)` and `UpdatePlayState`.

I did not exclude played items at the query level on purpose: an in-progress rewatch is also `Played = true` with `position > 0` and would be wrongly dropped from Continue Watching. Third-party paths (Trakt/webhook plugins) that mark items played without resetting the position are out of scope here.

**Code assistance**
Used Claude Code for the diagnosis, fix and test. Added `MovieNfoParserTests.Fetch_WatchedItem_ResetsResumePosition` (seed a 30-minute position, import a watched NFO, assert `Played == true` and position == 0); it fails without the fix and passes with it, and the full Jellyfin.XbmcMetadata.Tests suite is green. The reset runs on the same `userData` object and `SaveUserData` call that already persists the played flag, so it takes effect wherever the NFO watch-status import is applied.

**Issues**
Fixes #12054

### DIFF
--- a/MediaBrowser.XbmcMetadata/Parsers/BaseNfoParser.cs
+++ b/MediaBrowser.XbmcMetadata/Parsers/BaseNfoParser.cs
@@ -317,6 +317,14 @@ namespace MediaBrowser.XbmcMetadata.Parsers
                             {
                                 userData.Played = played;
 
+                                // A fully watched item has no resume position. Clearing it keeps the
+                                // item out of "Continue Watching", whose resume query treats any
+                                // PlaybackPositionTicks > 0 as resumable regardless of the played state.
+                                if (played)
+                                {
+                                    userData.PlaybackPositionTicks = 0;
+                                }
+
                                 if (!item.Id.IsEmpty())
                                 {
                                     _userDataManager.SaveUserData(user, item, userData, UserDataSaveReason.Import, CancellationToken.None);
--- a/tests/Jellyfin.XbmcMetadata.Tests/Parsers/MovieNfoParserTests.cs
+++ b/tests/Jellyfin.XbmcMetadata.Tests/Parsers/MovieNfoParserTests.cs
@@ -194,6 +194,29 @@ namespace Jellyfin.XbmcMetadata.Tests.Parsers
             Assert.Equal(_localImageFileMetadata.Name, result.Images[0].FileInfo.Name);
         }
 
+        [Fact]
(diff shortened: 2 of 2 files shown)


moontwister:
Withdrawing this PR. On closer review the change only affects the NFO-import path, whereas the reported issue (#12054) involves natively-played episodes with no NFO involvement — so this does not address that report's actual code path. I'd rather re-investigate the real root cause than leave a mismatched fix open. Apologies for the noise.

Issue #15846, opened 2025-12-21
Shows and seasons not changing watched state when manually marking as watched


byjosue:
### Description of the bug

After updating to versions >= 10.11 when trying to mark something as watched it does not work, particularly when trying to mark a whole series or a season as watched, it only marks the first episode as watched. When trying on the Android TV client it crashes but I assume that's maybe related to the client.

### Reproduction steps

1. Navigate to TV show
2. Mark the show or a season it as watched

### What is the current _bug_ behavior?

After updating the page the show does not remain watched and instead just marks the first episode

### What is the expected _correct_ behavior?

All the episodes contained within the season or show should be marked as watched

### Jellyfin Server version

10.11.5

### Specify the build version

10.11.5

### Environment

```markdown
- OS: OMV
- Linux Kernel:
- Virtualization: Docker
- Clients: Browser, Android TV
- Browser: Firefox
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- CPU Model:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Jellyfin Data Storage & Filesystem:
- Media Storage & Filesystem:
- External Integrations:
```

### Jellyfin logs

```shell
[2025-12-21 01:48:10.784 +00:00] [ERR] [3] Jellyfin.Api.Middleware.ExceptionMiddleware: Error processing request. URL "POST" "/Sessions/Playing/Progress".
System.TypeLoadException: Could not load type 'Jellyfin.Data.Entities.User' from assembly 'Jellyfin.Data, Version=10.11.5.0, Culture=neutral, PublicKeyToken=null'.
   at Jellyfin.Plugin.TubeArchivistMetadata.Plugin.OnWatchedStatusChange(Object sender, UserDataSaveEventArgs eventArgs)
   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.Start[TStateMachine](TStateMachine& stateMachine)
   at Jellyfin.Plugin.TubeArchivistMetadata.Plugin.OnWatchedStatusChange(Object sender, UserDataSaveEventArgs eventArgs)
   at Emby.Server.Implementations.Library.UserDataManager.SaveUserData(User user, BaseItem item, UserItemData userData, UserDataSaveReason reason, CancellationToken cancellationToken)
   at Emby.Server.Implementations.Session.SessionManager.OnPlaybackProgress(User user, BaseItem item, PlaybackProgressInfo info)
[...]

PR #17104, opened 2026-06-14
Don't let stale user-data keys shadow the current played state

### LINKED ISSUE #15795: Mark watched not working for media mis-identified and then corrected
### Description of the bug

If I issue an API command to mark a media item (that was initially mismatched) as watched (in this case) it returns a 200 OK but doesn't mark them as watched. This also happens via UI and tried marking as watched, and the same thing, they just never stick at being watched it just immediately goes from watched back to unwatched (in terms of the tick changing colour)

Monitoring the network tab in chrome dev tools shows the same issue as the API call  a 200 response but with {"PlaybackPositionTicks": 0, "PlayCount": 0, "IsFavorite": false, "Played": false, "Key": "503901", "ItemId": "978b227d683dbf824d134ba62f60db32"}

### Reproduction steps

I'm not sure it's easy to supply replication steps as it has to match an incorrect item but for me the example is 

The Bodyguard (1992) (tt0103855) for some reason matched to a Russian movie called Телохранитель (1992) (tt0197013) which has an aka of The Bodyguard

So the item was stored with 2 new rows with keys tt0197013 and 503901
Correcting the match created 3 new rows with keys tt0103855, 619 and the item id

When changing watched status via the API or UI it returns all 5 rows in the query and the proper 3 keys via `var keys = item.GetUserDataKeys();`

It then updates these 3 correct codes to watched, for example, but the older codes which happen to be the first two results are ignored

Image

Then is returns the first row of the result set which is Played=False as the old keys aren't updated, which is correct, but this leads to the database thinking it's watched and the API/UI thinking it's unwatched

 

I suspect that given UserDataManager.SaveUserData is the lowest point of many paths there should be a method added to clear out any keys for this ItemId/UserId combo that weren't returned as part of the item.GetUserDataKeys() call something like. This would lessen the chance of the issue being seen as old keys would be automatically removed and the correct item would be selected and returned to the API/UI 

```
private static void RemoveUserDataNoLongerAssociatedViaKey(JellyfinDbContext dbContext, User user, BaseItem item, List<string> keys)
{
    var noLongerAssociated = dbContext.UserData
           .Where(f => f.ItemId == item.Id && f.UserId == user.Id && !keys.Contains(f.CustomDataKey));

    if (noLongerAssociated.Any())
    {
        dbContext.UserData.RemoveRange(noLongerAssociated);
    }
 }

```

### Jellyfin Server version

10.11.4

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.4

### Environment

```markdown
- OS: Unraid 7.2.0
- Linux Kernel: Linux 6.12.54-Unraid x86_64
- Virtualization: Docker
- Clients: Web
- Browser: Chrome
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration: QSV
- CPU Model:
- GPU Model:
- Plugins:
- Reverse Proxy: nginx
- Base URL:
- Networking:
- Media Storage & Filesystem:
- External Integrations:
```

### Jellyfin logs

```shell
Logs not required
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

moontwister:
**Changes**
After an item is re-identified (for example fixing a wrong match), marking it watched returns 200 but immediately flips back to unwatched and the API returns `"Played": false` (#15795). Re-identifying leaves the old user-data rows under the previous keys alongside new rows under the current keys. `SaveUserData` writes the played state to all of the item's current keys, but the stale rows keep their old values, and the read paths picked an arbitrary row: `GetUserData` and the cached path in `GetUserDataBatch` did `item.UserData.Where(user).FirstOrDefault()` over every row, so a stale row could shadow the current state; the `GetUserDataBatch` query path compared `CustomDataKey == item.Id.ToString("N")`, which never matches the dashed `Id.ToString()` format actually stored, so it always fell back to `.First()`. When reading I now prefer the row stored under one of the item's current `GetUserDataKeys()`, falling back to the first available row. This is robust regardless of key format because it compares the stored `CustomDataKey` against the same key set `SaveUserData` writes to. The three read sites share a small helper.

**Code assistance**
Used Claude Code for the diagnosis, fix and tests. Added `UserDataManagerKeySelectionTests` (3 cases: a stale row ordered first no longer shadows the current-key row; only-stale rows still resolve; no rows yields the default with the primary key) and ran the full Jellyfin.Server.Implementations.Tests suite green.

**Issues**
Fixes #15795

### DIFF
--- a/Emby.Server.Implementations/Library/UserDataManager.cs
+++ b/Emby.Server.Implementations/Library/UserDataManager.cs
@@ -192,7 +192,10 @@ namespace Emby.Server.Implementations.Library
                 }
                 else
                 {
-                    var userData = item.UserData?.Where(e => e.UserId.Equals(user.Id)).Select(Map).FirstOrDefault();
+                    var keys = item.GetUserDataKeys();
+                    var userData = item.UserData is null
+                        ? null
+                        : SelectCurrentUserData(item.UserData.Where(e => e.UserId.Equals(user.Id)), keys);
                     if (userData is not null)
                     {
                         result[item.Id] = userData;
@@ -200,7 +203,6 @@ namespace Emby.Server.Implementations.Library
                     }
                     else
                     {
-                        var keys = item.GetUserDataKeys();
                         itemsNeedingQuery.Add((item, keys));
                     }
                 }
@@ -230,8 +232,7 @@ namespace Emby.Server.Implementations.Library
                     UserItemData userData;
                     if (userDataByItem.TryGetValue(item.Id, out var itemUserData) && itemUserData.Length > 0)
                     {
-                        var directDataReference = itemUserData.FirstOrDefault(e => e.CustomDataKey == item.Id.ToString("N"));
-                        userData = directDataReference is not null ? Map(directDataReference) : Map(itemUserData.First());
+                        userData = SelectCurrentUserData(itemUserData, keys) ?? Map(itemUserData[0]);
                     }
                     else
                     {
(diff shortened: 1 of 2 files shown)


moontwister:
Closing this. The issue it tracked (#15795) was just closed as a duplicate of #15846, with the fix landing in master for 12.0, and the underlying mis-identification is now handled by the Identify fix in #17146. With no open issue behind it I'd rather not leave a stale PR around. Happy to revisit the read-path handling if the stale-key case resurfaces. Thanks for the triage!

Issue #2923, opened 2020-04-19
Review Use of "\d" in regexes


mark-monteiro:
using `\d` n regexes does not perform as expected, and in many cases `[0-9]` should be preferred (as discovered [here](https://github.com/jellyfin/jellyfin/pull/2915#discussion_r410921281)).

* [x] Review existing uses of `\d` and replace with `[0-9]` where appropriate
* [ ] Consider adding this to the docs somewhere. Not sure where this would be appropriate to document, perhaps some discussion is needed here

d-pizarro:
More Info...

From: https://en.wiktionary.org/wiki/%E0%A9%A9
੩ Represents the number 3 (three) in the Gurmukhi script.

From: https://www.rexegg.com/regex-quickstart.html
Character | Legend | Example | Sample Match
-- | -- | -- | --
\d | Most engines: one digitfrom 0 to 9 | file_\d\d | file_25
\d | .NET, Python 3: one Unicode digit in any script | file_\d\d | file_9੩

ATahma:
This issue can be closed since #5142 has been merged

PR #17112, opened 2026-06-15
Fix series year lost during name parsing

theguymadmax:
In #14824 we stripped the year from the returned series name so titles like `1923 (2022)` would match correctly. That solved that issue, but created another since downstream uses the returned name to parse the year. An issue for series that share a name but have different years, e.g., `Doctor Who (2005)` and `Doctor Who (2024)`. 

For example, a folder named `Doctor Who (2024)` would resolve to `Doctor Who` with no year, causing providers to match it to `Doctor Who (2005)`, the first result.

**Changes**
Instead of discarding the year after stripping it, we now capture it from the same regex match and carry it through `SeriesInfo.Year` to the server resolver, which sets it as `ProductionYear` on the `Series` entity.

**Code assistance**
N/A

**Issues**
A lot of angry Doctor Who fans.

### DIFF
--- a/Emby.Naming/TV/SeriesResolver.cs
+++ b/Emby.Naming/TV/SeriesResolver.cs
@@ -21,7 +21,7 @@ namespace Emby.Naming.TV
         /// Regex that matches titles with year in parentheses. Captures the title (which may be
         /// numeric) before the year, i.e. turns "1923 (2022)" into "1923".
         /// </summary>
-        [GeneratedRegex(@"(?<title>.+?)\s*\(\d{4}\)")]
+        [GeneratedRegex(@"(?<title>.+?)\s*\((?<year>\d{4})\)")]
         private static partial Regex TitleWithYearRegex();
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


Emby.Naming/TV/SeriesResolver.cs
```diff
@@ -43,7 +43,8 @@ public static SeriesInfo Resolve(NamingOptions options, string path)
                     seriesName = titleWithYearMatch.Groups["title"].Value.Trim();
                     return new SeriesInfo(path)
                     {
-                        Name = seriesName
+                        Name = seriesName,
+                        Year = int.TryParse(titleWithYearMatch.Groups["year"].Value, out var year) ? year : null
```
Bond-009:
```suggestion
                        Year = int.TryParse(titleWithYearMatch.Groups["year"].ValueSpan, out var year) ? year : null
```

Emby.Naming/TV/SeriesResolver.cs
```diff
@@ -21,7 +21,7 @@ public static partial class SeriesResolver
         /// Regex that matches titles with year in parentheses. Captures the title (which may be
         /// numeric) before the year, i.e. turns "1923 (2022)" into "1923".
         /// </summary>
-        [GeneratedRegex(@"(?<title>.+?)\s*\(\d{4}\)")]
+        [GeneratedRegex(@"(?<title>.+?)\s*\((?<year>\d{4})\)")]
```
Bond-009: Don't use `\d`
https://github.com/jellyfin/jellyfin/issues/2923

Bond-009 (changes requested):

Shadowghost (approved):

Bond-009 (approved):

PR #17114, opened 2026-06-15
Resolve the Blu-ray/DVD playlist when extracting subtitles from disc folders

### LINKED ISSUE #16987: BDMV folder: on-the-fly PGS/image subtitle extraction fails — ExtractSubtitlesForFile is passed the directory path (still in 10.11.8, re #13041)
### Please describe your bug

On-the-fly / external extraction of image-based (PGS, `hdmv_pgs_subtitle`) subtitles fails for media stored as a **raw Blu-ray (BDMV) folder**. The subtitle `Stream` endpoint errors and the `.sup` is never produced, so any client that renders PGS externally never receives the subtitle.

The logs show the cause: `SubtitleEncoder.ExtractSubtitlesForFile` is handed the **folder path** (a BDMV folder's `MediaSource.Path`, which has no file extension) as the ffmpeg input. ffmpeg can't open a directory → `Unable to get streams` → `ffmpeg subtitle extraction failed` → the `.sup` is missing → the request fails.

Same problem as #13041 / #13042 (both closed as *not planned* / stale). **Still reproduces on 10.11.8** — filing fresh logs in case it can be reopened/prioritized.

**Key evidence that this is an input-path bug, not a decode/codec limitation:**
- **Burn-in works** — forcing transcode with subtitle burn-in renders the PGS correctly, because burn-in rides the *main transcode* ffmpeg command, which opens the source via the resolved Blu-ray playlist (the same resolution that makes video playback work).
- **Single-file sources work** — the identical on-the-fly extraction succeeds for an MKV or a single extracted `.m2ts` (the `.sup` is produced and the subtitle shows).
- Only **standalone extraction from a BDMV folder** fails, because that path uses the raw folder `MediaSource.Path` instead of resolving the playlist like the transcode pipeline does.

### Reproduction Steps
1. Add a movie stored as a raw Blu-ray folder (`BDMV/` + `STREAM/*.m2ts`, no single video file).
2. Select a PGS subtitle track on a client that renders image subtitles externally — or request it directly: `GET /Videos/{itemId}/{mediaSourceId}/Subtitles/{index}/0/Stream.pgssub`.
3. The subtitle never loads; the server log shows the failure below.
4. Contrast: the same title with burn-in (transcode) shows the subtitle, and a single-file `.m2ts`/`.mkv` of the same content extracts fine.

### Expected behavior
Subtitle extraction for a BDMV folder source should resolve the Blu-ray playlist (the same way the transcode/playback pipeline does — `bluray:` protocol or the selected `.mpls`/concat input) and extract the requested stream, exactly as it already does for single-file sources.

### Actual behavior
Extraction is invoked with the raw folder path and fails:

```
[WRN] MediaBrowser.MediaEncoding.Subtitles.SubtitleEncoder: Unable to get streams for File:"/path/to/<MOVIE_FOLDER>"
MediaBrowser.Common.FfmpegException: ffmpeg subtitle extraction failed for file:"/path/to/<MOVIE_FOLDER>"
   at MediaBrowser.MediaEncoding.Subtitles.SubtitleEncoder.ExtractSubtitlesForFile(String inputPath, String args, List`1 outputPaths, CancellationToken cancellationToken)
   at MediaBrowser.MediaEncoding.Subtitles.SubtitleEncoder.ExtractAllExtractableSubtitlesInternal(MediaSourceInfo mediaSource, List`1 subtitleStreams, CancellationToken cancellationToken)
   at MediaBrowser.MediaEncoding.Subtitles.SubtitleEncoder.ExtractAllExtractableSubtitles(MediaSourceInfo mediaSource, CancellationToken cancellationToken)
[ERR] Jellyfin.Api.Middleware.ExceptionMiddleware: Error processing request: "Could not find file '/config/data/subtitles/xx/<itemGuid>/15.sup'". URL "GET" "/Videos/<itemGuid>/<mediaSourceId>/Subtitles/15/0/Stream.pgssub".
```

`MediaSource.Path` here is the BDMV **folder** (no file extension) — that directory path is what gets passed to ffmpeg.

### Jellyfin Version
10.11.8

### Environment
- OS / install: Docker in TrueNAS Scale 25.10.3
- Source: raw Blu-ray folder (BDMV)
- Subtitle codec: PGS (`hdmv_pgs_subtitle`)
- Client: any client rendering PGS externally (subtitle never received)

### Additional context
Likely fix direction: `ExtractAllExtractableSubtitles` / `ExtractSubtitlesForFile` should reuse the BDMV source resolution the transcode pipeline already performs (resolve the `.mpls` / main `.m2ts`, or use the `bluray:` protocol) instead of the raw folder `MediaSource.Path`. Duplicate of #13041 / #13042, still present in 10.11.8.

moontwister:
**Changes**
On-the-fly subtitle extraction fails for media stored as a raw Blu-ray (`BDMV/`) or DVD (`VIDEO_TS/`) folder, so the `.sup`/`.srt` is never produced and clients that render image (PGS) subtitles externally get nothing (#16987). Extraction builds its ffmpeg input from `GetInputArgument(mediaSource.Path, mediaSource)`, which only special-cases `IsoType.BluRay` (ISO images) and not `VideoType.BluRay`/`VideoType.Dvd` (raw disc folders); for a disc-folder source `mediaSource.Path` is a directory, so ffmpeg cannot open it. The transcode/playback pipeline avoids this by using `GetInputPathArgument`, which resolves the disc playlist (`GetPrimaryPlaylistM2tsFiles`/`GetPrimaryPlaylistVobFiles`) and falls back to `GetInputArgument` for ordinary files (that is also why burn-in works). I switched the two main-file subtitle-extraction inputs (`ExtractAllExtractableSubtitlesInternal` and `ExtractTextSubtitle`) to `GetInputPathArgument`; it is identical to `GetInputArgument` for normal files. The external `.mks` input path is unchanged.

**Code assistance**
Used Claude Code for the diagnosis and fix. There is no behaviour change for ordinary files: `GetInputPathArgument` only resolves a playlist for `VideoType.Dvd`/`VideoType.BluRay` and otherwise returns `GetInputArgument(path, mediaSource)` unchanged, so the only difference is exactly the disc-folder case this fixes — and that uses the same resolver the transcode/burn-in pipeline already relies on. Build is clean and the full Jellyfin.MediaEncoding.Tests suite passes. I did not add a unit test: the extraction methods are private and spawn a real ffmpeg process, so this reuses the existing, already-exercised resolver rather than a brittle process-driven test. A full end-to-end check needs a real BDMV/VIDEO_TS folder source, which I don't have to test against.

**Issues**
Fixes #16987

### DIFF
--- a/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
+++ b/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
@@ -723,7 +723,7 @@ namespace MediaBrowser.MediaEncoding.Subtitles
             List<MediaStream> subtitleStreams,
             CancellationToken cancellationToken)
         {
-            var inputPath = _mediaEncoder.GetInputArgument(mediaSource.Path, mediaSource);
+            var inputPath = _mediaEncoder.GetInputPathArgument(mediaSource.Path, mediaSource);
             var outputPaths = new List<string>();
             var args = string.Format(
                 CultureInfo.InvariantCulture,
@@ -912,7 +912,7 @@ namespace MediaBrowser.MediaEncoding.Subtitles
                 {
                     var subtitleStreamIndex = EncodingHelper.FindIndex(mediaSource.MediaStreams, subtitleStream);
 
-                    var args = _mediaEncoder.GetInputArgument(mediaSource.Path, mediaSource);
+                    var args = _mediaEncoder.GetInputPathArgument(mediaSource.Path, mediaSource);
 
                     if (subtitleStream.IsExternal)
                     {


moontwister:
Withdrawing this PR for now. I'm re-verifying every fix I have open against its report's actual root cause before keeping it in review. I'll resubmit once I've fully confirmed the diagnosis. Thanks for your time.

PR #17116, opened 2026-06-15
Skip parsing root-level folders in SeriesResolver

theguymadmax:
While debugging another issue, I noticed that root-level folders were being passed into `SeriesResolver` and unnecessarily processed by `SeriesPathParser`

**Changes**
Add an early return for root-level folders to skip series parsing.

**Code assistance**
N/A

**Issues**
N/A

### DIFF
--- a/Emby.Server.Implementations/Library/Resolvers/TV/SeriesResolver.cs
+++ b/Emby.Server.Implementations/Library/Resolvers/TV/SeriesResolver.cs
@@ -57,6 +57,11 @@ namespace Emby.Server.Implementations.Library.Resolvers.TV
                     return null;
                 }
 
+                if (args.Parent is not null && args.Parent.IsRoot)
+                {
+                    return null;
+                }
+
                 var seriesInfo = Naming.TV.SeriesResolver.Resolve(_namingOptions, args.Path);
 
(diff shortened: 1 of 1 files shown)


Emby.Server.Implementations/Library/Resolvers/TV/SeriesResolver.cs
```diff
@@ -57,6 +57,11 @@ protected override Series Resolve(ItemResolveArgs args)
                     return null;
                 }
 
+                if (args.Parent is not null && args.Parent.IsRoot)
```
Shadowghost, line 60: I think we usually use != here

theguymadmax: Like three years ago :-)  https://github.com/jellyfin/jellyfin/pull/8511

theguymadmax: Also, it's already written that way in the code; in fact, it was changed from  != to is not by Bond, the above PR I linked.

Bond-009 (approved):

Shadowghost (approved):

PR #17119, opened 2026-06-16
Make ItemValues get-or-create concurrency-safe under parallel item saves

fuhrysteve:
`ItemPersistenceService.UpdateOrInsertItems` reads the existing `ItemValues`, then inserts the missing ones through the change tracker with no concurrency guard. Two items saved in parallel that introduce the same new value (a shared genre, studio, tag, ...) both insert it, and the loser violates the unique `IX_ItemValues_Type_Value` index and aborts its transaction. SQLite serialises writes, so this only surfaces on backends with concurrent writers.

**Changes**

The race is settled in C# rather than in SQL. The save is retried, bounded, when a re-read shows the values it meant to create exist now, which is what losing the race looks like from here; any other failure is rethrown untouched. That is the same "let the write be the check, then re-read to see whether the assumption it was made under still holds" shape `SaveImagesAsync` already uses in this class for the item-deleted-underneath race.

Behaviour is unchanged on backends that serialise writes: with no concurrent writer there is never a conflict, so the retry never runs.

Two details worth pointing at during review:

- The retry check runs in the `catch` body, not in the exception filter. A filter runs before the failed attempt unwinds, so the transaction would still be open, and still holding its write lock, while the check reads.
- The whole save is retried rather than just the `ItemValues` insert, because the method runs as one transaction and the values must be resolved before the map rows that reference them are built.

**Previous revision**

The first version of this PR used `INSERT ... ON CONFLICT ("Type", "Value") DO NOTHING`. Per @Shadowghost's review that is not portable, MariaDB has no such clause, so it has been replaced. No dialect-specific SQL is emitted now.

**Tests**

`ItemPersistenceServiceItemValueRaceTests` covers the unique index being enforced, two items in one batch sharing a new value creating it once, an already-existing value being reused, and the retry converging onto the winner's row.

The last of those needs a note. SQLite cannot host a genuine two-connection version of this race, which is the same reason the bug does not occur there: the save under test already holds the write lock, so a competing writer is simply locked out. The test therefore stages the race through the seams the service is constructed with, and the two halves that matter stay real. The conflicting row is written into the failing save's own transaction, so the exception comes from the actual unique index rather than a fabricated one, and the winning row is then committed for real, from its own connection, before the service re-reads.

**Issues**

Reported downstream (PostgreSQL plugin) as JPVenson/Jellyfin.Pgsql#19; the bug is in core.

### DIFF
--- a/Jellyfin.Server.Implementations/Item/ItemPersistenceService.cs
+++ b/Jellyfin.Server.Implementations/Item/ItemPersistenceService.cs
@@ -295,11 +295,13 @@ public class ItemPersistenceService : IItemPersistenceService
         var values = allListedItemValues.Select(e => e.Value).Distinct().ToArray();
         var allListedItemValuesSet = allListedItemValues.ToHashSet();
 
-        var existingValues = context.ItemValues
+        ItemValue[] StoredItemValues() => context.ItemValues
             .Where(e => types.Contains(e.Type) && values.Contains(e.Value))
             .AsEnumerable()
             .Where(e => allListedItemValuesSet.Contains((e.Type, e.Value)))
             .ToArray();
+
+        var existingValues = StoredItemValues();
         var missingItemValues = allListedItemValues.Except(existingValues.Select(f => (MagicNumber: f.Type, f.Value))).Select(f => new ItemValue()
         {
             CleanValue = f.Value.GetCleanValue(),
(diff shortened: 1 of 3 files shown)


Shadowghost (commented):
we should not use raw sql since this potentially behaves differently on different engines.
e.g. On CONFLICT does not exist in mariaDB

Issue #10303, opened 2023-09-28
[Issue]: No way to select video stream to play (like with audio/subtitles)


Moonbase59:
### Please describe your bug

I have several (usually MKV) files that have multiple video streams to select from. KODI and other players let me choose, but Jellyfin 10.8.11 doesn’t—it always selects the first video stream found in a file and offers no selection like it does with audio and subtitle streams.

Screen shot web UI:

![Jellyfin – Mozilla Firefox_002](https://github.com/jellyfin/jellyfin/assets/3706922/5546fedf-4e90-4f62-8b2b-567f8de29ffe)

Screen shot media info screen (it _does_ see both video streams):

![Jellyfin – Mozilla Firefox_001](https://github.com/jellyfin/jellyfin/assets/3706922/e5552704-bbe9-40ef-861d-aa5cbba4f457)

MediaInfo data:
```text
General
Unique ID                                : 105556898050516193135998472102841122669 (0x4F6984E2600F7A3A241645DAD0577F6D)
Complete name                            : /home/matthias/Videos/Movies/Night on Earth (1991).mkv
Format                                   : Matroska
Format version                           : Version 4
File size                                : 3.84 GiB
Duration                                 : 2 h 3 min
Overall bit rate mode                    : Variable
Overall bit rate                         : 4 466 kb/s
Movie name                               : Night on Earth (1991)
Encoded date                             : UTC 2023-09-28 10:29:30
Writing application                      : mkvmerge v72.0.0 ('Minuano (Six-eight)') 64-bit
Writing library                          : libebml v1.4.4 + libmatroska v1.7.1

Video #1
ID                                       : 1
Format                                   : AVC
Format/Info                              : Advanced Video Codec
Format profile                           : Main@L3.1
Format settings                          : CABAC / 3 Ref Frames
Format settings, CABAC                   : Yes
Format settings, Reference frames        : 3 frames
Codec ID                                 : V_MPEG4/ISO/AVC
Duration                                 : 2 h 3 min
Bit rate mode                            : Variable
Bit rate                                 : 2 167 kb/s
Maximum bit rate                         : 2 200 kb/s
Width                                    : 1 280 pixels
Height                                   : 720 pixels
Display aspect ratio                     : 16:9
Frame rate mode                          : Constant
Frame rate                               : 25.000 FPS
Standard                                 : PAL
Color space                              : YUV
Chroma subsampling                       : 4:2:0
Bit depth                                : 8 bits
[...]

PR #17123, opened 2026-06-17
Honor VideoStreamIndex in media playback negotiation

ivenos:
### Changes

The transcode path (`EncodingHelper`) already knows how to map a specific video stream, there was just no way to actually ask for one through PlaybackInfo. This adds `VideoStreamIndex` to the negotiation, right next to the `AudioStreamIndex` and `SubtitleStreamIndex` that are already there: `PlaybackInfoDto`/`OpenLiveStreamDto`/`LiveStreamRequest`, `MediaInfoController`, `MediaInfoHelper`, `MediaOptions`, and the stream URL built in `StreamInfo`.

In `StreamBuilder`, if you ask for a video stream that isn't the default one, it gets written into the stream URL and direct play is switched off, so the stream you picked actually gets mapped during remux/transcode. Asking for the default stream (or not asking at all) behaves exactly like before. And if the index doesn't match a video stream in that source, it just falls back to the default instead of doing anything weird.

Web side is here: https://github.com/jellyfin/jellyfin-web/pull/8051

### Issues

Related to jellyfin/jellyfin-web#1421, https://github.com/jellyfin/jellyfin/issues/10303 and https://features.jellyfin.org/posts/942/support-multi-track-video-quality

### Follow-up / known gap

Clients have to send `VideoStreamIndex` for any of this to do something. The web client is in the companion PR. The native apps (Android TV, Swiftfin, Roku, ...) each need their own picker and won't get it automatically. I looked at the Android TV one but didn't feel solid enough about threading the selected index through its playback launch path to do it here, so I'm leaving that as a follow-up.

### Testing

- `dotnet build` (net10.0), clean.
- `StreamBuilderTests`, 300/300 passing.
- Ran it against a local server with an MKV that has a couple of video tracks, plus a movie with two versions that have different stream layouts on purpose. Checked that the picked index shows up in the transcode URL for the right source, that a bogus index (one pointing at an audio stream, or out of range) falls back to the default instead of breaking, and that picking a stream in one version doesn't bleed into the other.

### DIFF
--- a/Jellyfin.Api/Controllers/MediaInfoController.cs
+++ b/Jellyfin.Api/Controllers/MediaInfoController.cs
@@ -98,6 +98,7 @@ public class MediaInfoController : BaseJellyfinApiController
     /// <param name="userId">The user id.</param>
     /// <param name="maxStreamingBitrate">The maximum streaming bitrate.</param>
     /// <param name="startTimeTicks">The start time in ticks.</param>
+    /// <param name="videoStreamIndex">The video stream index.</param>
     /// <param name="audioStreamIndex">The audio stream index.</param>
     /// <param name="subtitleStreamIndex">The subtitle stream index.</param>
     /// <param name="maxAudioChannels">The maximum number of audio channels.</param>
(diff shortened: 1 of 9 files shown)


gnattu:
I have concerns beyond playback for this. Currently lots of functionalities are designed/implemented on the fact that a single item/file can only have one video stream. Adding this would have larger impact because supporting multi stream video well is not as easy as change the playback api to pass in a stream index. It changes the one to many relationship of metadata/supplemental medias to multi to multi relationship.  Multiple video streams can have multiple codec/resolution and the support for that is missing currently and adding support for that alone is not a small task, because nothing in Jellyfin expects a video item can have multiple video streams.

PR #17146, opened 2026-06-21
Fix Identify returning wrong results

### LINKED ISSUE #14761: identifying won't work if it is missing in main source
### Description of the bug

Something that isn't on TVDB, because the studio released a short movie on an event and YT. It is on TMDB, anidb, IMDB has it. I used identify and selected a release, and it always goes to random movie (similar name if you close both eyes) or so.

### Reproduction steps

1. Get something that isn't on TVDB
2. Use TVDB as main info for `Metadata downloaders (Movies)` `TheTVDB`
3. identify the movie in TMDB or so
4. You will get a movie from TVDB

Identify will not overwrite OR fallback the metadata provider, in case it isn't on TVDB. Tried deleting DB, Refetch metadata and deleting the TVDB ID before `replace all metadata`.

Fallback if there is no match at all or use the metadata rfom identify

### Jellyfin Server version

Master

### Specify commit id

0a2cf69a5500ef3feb2964011e27c5c5bb599de7

_No response_

### Specify version number

_No response_

### Specify the build version

Release

### Environment

```markdown
- OS:
- Linux Kernel:
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Jellyfin Data Storage:
- Media Storage:
- External Integrations:
```

### Jellyfin logs

```shell
No log entires
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

Search
Image

Selection
Image

Conformation image behavior
Image

TVDB....

Image

### Additional information

_No response_

theguymadmax:
When manually identifying an item, if the selected result didn't exist in the highest priority metadata provider, it would fall back to a name search and match the wrong item, causing incorrect metadata and IDs to be saved.

**Changes**
When performing an Identify, reorder the remote providers so the one the user picked from runs first.

**Code assistance**
N/A

**Issues**
Fixes: #14761

### DIFF
--- a/MediaBrowser.Providers/Manager/MetadataService.cs
+++ b/MediaBrowser.Providers/Manager/MetadataService.cs
@@ -831,8 +831,16 @@ namespace MediaBrowser.Providers.Manager
             var isLocalLocked = temp.Item.IsLocked;
             if (!isLocalLocked && (options.ReplaceAllMetadata || options.MetadataRefreshMode > MetadataRefreshMode.ValidationOnly))
             {
-                var remoteResult = await ExecuteRemoteProviders(temp, logName, false, id, providers.OfType<IRemoteMetadataProvider<TItemType, TIdType>>(), cancellationToken)
-                    .ConfigureAwait(false);
+                var remoteProviders = providers.OfType<IRemoteMetadataProvider<TItemType, TIdType>>();
+
+                // When identifying, run the provider the user picked first so the correct IDs are used.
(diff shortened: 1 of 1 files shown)


Shadowghost (approved):

IDisposable (approved):
This is really great, thanks

Bond-009 (approved):

PR #17153, opened 2026-06-22
Fix too many SQL variables in DeleteItem for large batch deletes

joshuaboniface:
**Changes**

Fix a few bugs in the FixIncorrectOwnerIdRelationships migration.

1. Fix the single DeleteItemsUnsafeFast -> DeleteItem(ids) call to use WhereOneOrMany so the id set is bound as a single
json_each parameter, like the rest of the method, making bulk deletes work for unlimited library sizes - my library was breaking hard on this.
2. Add progress indication in steps of 500 like other migrations, to keep the log reflecting what's actually happening.

**Code assistance**
Claude Opus 4.8 for troubleshooting and generation of patch

**Issues**
N/A

### DIFF
--- a/Jellyfin.Server/Migrations/Routines/20260115120000_FixIncorrectOwnerIdRelationships.cs
+++ b/Jellyfin.Server/Migrations/Routines/20260115120000_FixIncorrectOwnerIdRelationships.cs
@@ -136,19 +136,38 @@ public class FixIncorrectOwnerIdRelationships : IAsyncMigrationRoutine
 
         if (allIdsToDelete.Count > 0)
         {
-            // Batch-resolve items for metadata path cleanup, then delete all at once
-            var itemsToDelete = allIdsToDelete
-                .Select(id => _libraryManager.GetItemById(id))
-                .Where(item => item is not null)
-                .ToList();
-            _libraryManager.DeleteItemsUnsafeFast(itemsToDelete!);
+            _logger.LogInformation("Deleting {Count} duplicate database entries...", allIdsToDelete.Count);
 
-            // Fall back to direct DB deletion for any items that couldn't be resolved via LibraryManager
-            var deletedIds = itemsToDelete.Select(i => i!.Id).ToHashSet();
-            var unresolvedIds = allIdsToDelete.Where(id => !deletedIds.Contains(id)).ToList();
-            if (unresolvedIds.Count > 0)
+            // Delete in batches so progress is visible (item resolution and deletion can take a
+            // long time on large libraries) and so we never issue one massive delete transaction.
+            const int deleteBatchSize = 500;
+            var deletedSoFar = 0;
+            for (var offset = 0; offset < allIdsToDelete.Count; offset += deleteBatchSize)
             {
-                _persistenceService.DeleteItem(unresolvedIds);
+                cancellationToken.ThrowIfCancellationRequested();
+
+                var batchIds = allIdsToDelete.GetRange(offset, Math.Min(deleteBatchSize, allIdsToDelete.Count - offset));
+
+                // Resolve items for metadata path cleanup, then delete this batch
+                var itemsToDelete = batchIds
+                    .Select(id => _libraryManager.GetItemById(id))
+                    .Where(item => item is not null)
+                    .ToList();
+                if (itemsToDelete.Count > 0)
+                {
+                    _libraryManager.DeleteItemsUnsafeFast(itemsToDelete!);
+                }
+
+                // Fall back to direct DB deletion for any items that couldn't be resolved via LibraryManager
+                var deletedIds = itemsToDelete.Select(i => i!.Id).ToHashSet();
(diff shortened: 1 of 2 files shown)


Jellyfin.Server/Migrations/Routines/20260115120000_FixIncorrectOwnerIdRelationships.cs
```diff
@@ -136,19 +136,38 @@ private async Task RemoveDuplicateItemsAsync(JellyfinDbContext context, Cancella
...
-            var deletedIds = itemsToDelete.Select(i => i!.Id).ToHashSet();
-            var unresolvedIds = allIdsToDelete.Where(id => !deletedIds.Contains(id)).ToList();
-            if (unresolvedIds.Count > 0)
+            _logger.LogInformation("Deleting {Count} duplicate database entries...", allIdsToDelete.Count);
+
+            // Delete in batches so progress is visible (item resolution and deletion can take a
+            // long time on large libraries) and so we never issue one massive delete transaction.
+            const int deleteBatchSize = 500;
+            var deletedSoFar = 0;
+            for (var offset = 0; offset < allIdsToDelete.Count; offset += deleteBatchSize)
```
cvium, line 145: nit: This can be simplified with https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.chunk?view=net-10.0

Jellyfin.Server/Migrations/Routines/20260115120000_FixIncorrectOwnerIdRelationships.cs
```diff
@@ -136,19 +136,38 @@ private async Task RemoveDuplicateItemsAsync(JellyfinDbContext context, Cancella
...
-
-            // Fall back to direct DB deletion for any items that couldn't be resolved via LibraryManager
-            var deletedIds = itemsToDelete.Select(i => i!.Id).ToHashSet();
-            var unresolvedIds = allIdsToDelete.Where(id => !deletedIds.Contains(id)).ToList();
-            if (unresolvedIds.Count > 0)
+            _logger.LogInformation("Deleting {Count} duplicate database entries...", allIdsToDelete.Count);
+
+            // Delete in batches so progress is visible (item resolution and deletion can take a
+            // long time on large libraries) and so we never issue one massive delete transaction.
+            const int deleteBatchSize = 500;
```
cvium, line 143: 500 is quite small. How many items are we expecting?

cvium (approved):

Jellyfin.Server/Migrations/Routines/20260115120000_FixIncorrectOwnerIdRelationships.cs
joshuaboniface: From my run, it was about 50,000, and deleting them even in 500 batches took a very long time (10+ minutes). It was originally 1000 but it was even slower in comparison.

Jellyfin.Server/Migrations/Routines/20260115120000_FixIncorrectOwnerIdRelationships.cs
joshuaboniface: Interesting, and probably worthwhile going forward - I would like to see *any* potentially-long-running data manipulation migration do batching like this with logging going forward.

Jellyfin.Server/Migrations/Routines/20260115120000_FixIncorrectOwnerIdRelationships.cs
cvium: That's weird. 50k is nothing. Do you know which step is slow?

Shadowghost (approved):

Jellyfin.Server/Migrations/Routines/20260115120000_FixIncorrectOwnerIdRelationships.cs
joshuaboniface: The actual delete itself was painfully slow. The time from start to crash in my first attempt was about 3 minutes. With a 1000 batch, I got done about 3000 in 3 minutes; with a 500 batch I got done all 50,000 in about 10 minutes.

Shadowghost: I don't get why this is so slow, maybe IO bound?

IDisposable (approved):

Jellyfin.Server/Migrations/Routines/20260115120000_FixIncorrectOwnerIdRelationships.cs
joshuaboniface: Possible, though my systems are on a fast Ceph cluster of SSDs. I do have graphs of the various disk parameters (1 minute increments) during this time, and nothing is amiss - no massive increases in I/O or anything like that, so I'm not certain.

PR #17155, opened 2026-06-22
Return the created API key from POST /Auth/Keys

### LINKED ISSUE #16258: POST on /Auth/Keys does not return the created api key

ivenos:
**Changes**
`POST /Auth/Keys` created the key but returned `204` with no body, so the only way to get the token back was to `GET /Auth/Keys` and match on the app name, which isn't unique. It now returns the generated access token with `200`. Added a couple of tests for `AuthenticationManager.CreateApiKey` covering the returned token.

**Code assistance**
Used Claude Opus 4.8 to help navigate the codebase and draft the change.

**Issues**
Fixes #16258

### DIFF
--- a/Jellyfin.Api/Controllers/ApiKeyController.cs
+++ b/Jellyfin.Api/Controllers/ApiKeyController.cs
@@ -1,6 +1,8 @@
 using System.ComponentModel.DataAnnotations;
+using System.Linq;
 using System.Threading.Tasks;
 using Jellyfin.Api.Constants;
+using Jellyfin.Api.Models.ApiKeyDtos;
 using MediaBrowser.Common.Api;
 using MediaBrowser.Controller.Security;
 using MediaBrowser.Model.Querying;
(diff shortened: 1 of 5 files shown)


nielsvanvelzen (changes requested):
The response of creating a key should be the same as getting it. But just like the previous (stale) attempt (#16431) we should first add a new DTO so no internal models are exposed to the api.

ivenos:
Makes sense. Added an `AuthenticationInfoDto` and switched both `GET` and `POST /Auth/Keys` over to it, so the internal model isn't exposed anymore.

ivenos:
Fixed the DTO, it only had `AccessToken`/`AppName`/`DateCreated` so GET was losing `Id`, `DeviceId`, `IsActive` etc. All empty for api keys anyway but it's still a break, so I put them back. Can strip them again if you'd rather.

Other thing: `CreateApiKey` returns `Task<AuthenticationInfo>` now instead of `Task`, which breaks ABI for plugins. apicompat flags it but the bot can't comment on fork PRs so it's not visible here. Ok, or keep the interface as-is?

cvium (approved):

nielsvanvelzen (approved):
API change seem fine

PR #17160, opened 2026-06-22
Truncate ISO-639-2 language display names at first delimiter

### LINKED ISSUE #16859: [12.0] ISO-639-2 language labels create noisy and confusing track lists

854562:
**Changes**
Prevents raw ISO-639-2 values (e.g. "Greek, Modern (1453-)") from cluttering the audio and subtitle display names by truncating them at the first comma or semicolon ("Greek").

This behaviour was a (probably unintended) side effect of #15947 and was introduced in ProbeResultNormalizer in #16567 . Explanation and examples of why this is undesirable in #16859 .

Some examples:
| Before| After |
|--------|--------|
| Romanian; Moldavian; Moldovan | Romanian |
| Greek, Modern (1453-) | Greek |
| Spanish; Castilian | Spanish |
| Asturian; Bable; Leonese; Asturleonese | Asturian |
| Dutch; Flemish | Dutch |

Built locally and verified working.

**Code assistance**
Gemini for double checking

**Issues**
Fixes #16859

### DIFF
--- a/Emby.Server.Implementations/Localization/LocalizationManager.cs
+++ b/Emby.Server.Implementations/Localization/LocalizationManager.cs
@@ -262,6 +262,24 @@ namespace Emby.Server.Implementations.Localization
         }
 
         /// <inheritdoc />
+        public string? GetLanguageDisplayName(string language)
+        {
+            if (string.IsNullOrEmpty(language))
+            {
+                return null;
+            }
+
+            var displayName = FindLanguageInfo(language)?.DisplayName;
+            if (displayName is null)
+            {
+                return null;
+            }
+
+            // Truncate at the first delimiter to avoid cluttered display names
+            return displayName.Split([';', ','], StringSplitOptions.None)[0].Trim();
+        }
+
+        /// <inheritdoc />
         public IReadOnlyList<CountryInfo> GetCountries()
         {
             using var stream = _assembly.GetManifestResourceStream(CountriesPath) ?? throw new InvalidOperationException($"Invalid resource path: '{CountriesPath}'");
--- a/Jellyfin.Server.Implementations/Item/MediaStreamRepository.cs
+++ b/Jellyfin.Server.Implementations/Item/MediaStreamRepository.cs
@@ -175,7 +175,7 @@ public class MediaStreamRepository : IMediaStreamRepository
                 var culture = _localization.FindLanguageInfo(dto.Language);
                 // Truncate at the first delimiter to avoid cluttered display names
                 dto.LocalizedLanguage = culture?.DisplayName is { } name
-                    ? name.Split(';', ',')[0].Trim()
+                    ? name.Split([';', ','])[0].Trim()
                     : null;
             }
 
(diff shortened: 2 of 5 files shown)


Shadowghost (commented):
Can we please get some tests for this.

Additionally, the duplicated logic can be extracted into a helper, so we don't forget to update it in all three places if we need to in the future

crobibero (approved):

theguymadmax (approved):

PR #17170, opened 2026-06-23
Rework bitrate reporting

### LINKED ISSUE #16248: Bitrate is incorrectly reported to frontend for video files.

Shadowghost:
Alternative to #16351

**Changes**
* Stop reporting the container bitrate as the video bitrate, estimate the missing video bitrate instead
  * Guarded to avoid bad estimates. Only applied when the result is positive and every audio stream's bitrate is known, so unaccounted audio is never silently folded into the video number
* Extended `GetEstimatedAudioBitrate` codec coverage (used to fill in missing audio bitrates): added DTS/DCA, Opus, Vorbis, WMA, MP2, TrueHD/MLP; also fixed a latent bug where 3- and 4-channel audio returned null for every codec.
  * Lossless codecs now scale per channel (FLAC/ALAC, TrueHD/MLP) instead of a flat multichannel value
  * Profile-aware DTS handling

**Code assistance**
None

**Issues**
Fixes #16248

### DIFF
--- a/MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
+++ b/MediaBrowser.MediaEncoding/Probing/ProbeResultNormalizer.cs
@@ -254,16 +254,38 @@ namespace MediaBrowser.MediaEncoding.Probing
                 {
                     if (mediaStream.Type == MediaStreamType.Audio && !mediaStream.BitRate.HasValue)
                     {
-                        mediaStream.BitRate = GetEstimatedAudioBitrate(mediaStream.Codec, mediaStream.Channels);
+                        mediaStream.BitRate = GetEstimatedAudioBitrate(mediaStream.Codec, mediaStream.Profile, mediaStream.Channels);
                     }
                 }
 
-                var videoStreamsBitrate = info.MediaStreams.Where(i => i.Type == MediaStreamType.Video).Select(i => i.BitRate ?? 0).Sum();
-                // If ffprobe reported the container bitrate as being the same as the video stream bitrate, then it's wrong
-                if (videoStreamsBitrate == (info.Bitrate ?? 0))
+                // ffprobe frequently omits the per-stream video bitrate (common in MP4/MKV containers).
+                // Estimate the missing video bitrate as the container bitrate minus the combined stream bitrates.
+                var videoStreams = info.MediaStreams.Where(i => i.Type == MediaStreamType.Video).ToList();
+                if (info.Bitrate.HasValue
+                    && videoStreams.Count == 1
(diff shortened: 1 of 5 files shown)


gnattu (approved):
The estimation is still rough but should be better than container bitrate. Although still inaccurate, but should be closer than container bitrate.

nyanmisaka (approved):

Bond-009 (approved):

PR #17127, opened 2026-06-18
Fix HLS segment cleanup silently failing for live TV streams

supersebbo:
Changes

HLS segment cleanup was silently disabled for live TV streams due to two independent bugs. First, the EnableSegmentCleaning predicate in TranscodeManager required RunTimeTicks.HasValue, which is always null for live TV — so the cleaner was never started. Second, even when the cleaner did run, it derived the client's playback position from DownloadPositionTicks, which live TV clients never report (they play at the live edge with no ticks-based position), so the position stayed at zero and no segments were ever deleted. This fix extends EnableSegmentCleaning to include live TV HLS streams via the existing IsSegmentedLiveStream flag, and adds a GetLiveSegmentPositionSeconds fallback that derives the current position from the highest segment index present on disk when the client reports no position.

Code assistance

Claude (claude-sonnet-4-6) was used for: identifying the two-bug root cause, generating the GetLiveSegmentPositionSeconds implementation, writing the unit tests for EnableSegmentCleaning, and debugging guidance during local testing against a live HDHomeRun tuner.

Issues

Fixes #13725
Partially fixes #17120

Issue #16880, opened 2026-05-20
HDHomeRun live TV channel continues streaming on server even after pause or closing Android app using integrated player (Based on ExoPlayer)


DannyLaQc:
### Description of the bug

When playing a live TV channel from an HDHomeRun tuner through Jellyfin Server, the stream on the server side never stops when using the integrated player (based on ExoPlayer), even if I:

Press Pause in the Jellyfin Android app
Close the app completely
Switch to another channel or exit the player (it will steam/use a second tuner channel)

The server keeps the HDHomeRun stream active indefinitely until I manually restart the Jellyfin Server service or stop the stream via the dashboard.

### Reproduction steps

1. Set up an HDHomeRun device with Jellyfin Live TV.
2. Open the Jellyfin Android app.
3. Change the default "Web player" to "Integrated player" in Settings/Clients Setting/Video player type.
4. Play any live TV channel from the HDHomeRun.
5. Pause the playback or completely close the app.
6. Check the Jellyfin server bandwith, cpu usage (dashboard says stopped) or the HDHomeRun status -- the stream remains active.

### What is the current _bug_ behavior?

The stream continues running on the server and on the HDHomeRun indefinitely.

### What is the expected _correct_ behavior?

The server should stop the HDHomeRun stream shortly after the client pauses, stops, or disconnects.

### Jellyfin Server version

10.11.8

### Specify the build version

10.11.8

### Environment

```markdown
- OS: ZimaOS
- Linux Kernel: Custom ZimaOS
- Virtualization: Docker linuxserver/jellyfin : latest
- **Clients: Android App and videotype to "Integrated player" 
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration: bug with or without
- CPU Model: 
- GPU Model:
- Plugins:
- Reverse Proxy: Lan and Wan bug
- Base URL:
- Networking:
- Jellyfin Data Storage & Filesystem:
- Media Storage & Filesystem:
- External Integrations:
```

### Jellyfin logs

```shell
### Jellyfin Server Logs (relevant parts - Live TV HDHomeRun)

[2026-05-19 22:57:33.xxx] [INF] Jellyfin.LiveTv.LiveTvMediaSourceProvider: Opening channel stream from "Emby", external channel Id: "hdhr_4.1"
[2026-05-19 22:57:33.xxx] [INF] Jellyfin.LiveTv.DefaultLiveTvService: Streaming Channel "hdhr_4.1"
[...]

PR #17177, opened 2026-06-24
Close live TV stream on session stop when client omits LiveStreamId

awkimball:
This is the server side of the live TV / HDHomeRun tuner leak I described in https://github.com/jellyfin/jellyfin/issues/16880\#issuecomment-4792467368.

Short version: when a transcoded live TV session stops, the consumer ffmpeg gets killed but the underlying SharedHttpStream is never closed if the client's stop report doesn't include a LiveStreamId. The consumer count stays at 1, so the tuner keeps copying to disk until the server is restarted.

Since #13220 the only thing that closes a live stream on stop is `CloseLiveStreamIfNeededAsync`, and `OnPlaybackStopped` only calls it when `info.LiveStreamId` is set. The `_activeLiveStreamSessions` dictionary already knows which live stream a session is mapped to (it gets populated from playback start and progress), but nothing uses it in reverse when the stop report has no LiveStreamId. Clients that don't echo it back (Swiftfin in my case, and this lines up with the Roku and Android reports in #15769) never trigger the close.

What this does:

- Adds `CloseLiveStreamsForSessionAsync(sessionId)`, which looks up any live stream still mapped to the session in `_activeLiveStreamSessions` and closes it.
- Calls it from `OnPlaybackStopped` (both the normal stop and the negative-PositionTicks guard) and from `CloseIfNeededAsync` when a session ends.
- Does not change the ConsumerCount model or the #13220 dedup behavior. Closing is idempotent because `CloseLiveStreamIfNeededAsync` only acts if the id is still in the map.

What this does not cover:

This handles the stop and disconnect paths. It does not cover the pure pause case in #16880 where the app stays alive and paused (that's the inactive-session sweeper, which only reaps sessions with NowPlayingItem set), and it doesn't touch the `KillTranscodingJobs(closeLiveStream: false)` path. Those are separate follow-ups if that direction sounds reasonable.

Testing:

- Added a unit test in `SessionManagerTests` that starts playback with a LiveStreamId, reports stop without one, and asserts the live stream gets closed.
- Full solution builds, and the unit test project passes (549 passed, 5 skipped).

Related: #16880, #15769, #16728

Relationship to other open live TV PRs:

This is orthogonal to #17128 and #17127 and stacks cleanly with both.

- #17128 (live tv stream buffer unbounded growth) caps how big the buffered/on-disk data gets while a stream is open. It doesn't change the close path, so the tuner still stays open until something actually closes the live stream. This PR is the part that closes it.
- #17127 (HLS segment cleanup) handles leftover transcode segments on the TranscodeManager side. Different layer, no overlap with the session stop path.

None of them touch SessionManager or the live stream close decision, so there's no conflict regardless of merge order.

### DIFF
--- a/Emby.Server.Implementations/Session/SessionManager.cs
+++ b/Emby.Server.Implementations/Session/SessionManager.cs
@@ -317,6 +317,10 @@ namespace Emby.Server.Implementations.Session
                     await CloseLiveStreamIfNeededAsync(session.PlayState.LiveStreamId, session.Id).ConfigureAwait(false);
                 }
 
+                // Close any live streams that are still mapped to this session but were
+                // not reported via PlayState (e.g. the client never sent a clean stop).
+                await CloseLiveStreamsForSessionAsync(session.Id).ConfigureAwait(false);
+
                 await OnSessionEnded(session).ConfigureAwait(false);
             }
         }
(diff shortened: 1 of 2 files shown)


awkimball:
Fixed in #17178, closing this one as it was not the correct approach.

PR #17183, opened 2026-06-25
Add optional TwelveLabs Pegasus plugin (auto-subtitles & summaries)

mohit-twelvelabs:
Hi! I'm Mohit, I work at TwelveLabs (@mohit-twelvelabs).

This adds an **opt-in** TwelveLabs plugin under `MediaBrowser.Providers/Plugins/TwelveLabs/` that uses **Pegasus** (TwelveLabs' video-understanding model) to generate auto-subtitles and summaries for library items — useful for media that ships without captions.

It's fully opt-in and non-breaking: the plugin is inert unless configured with a TwelveLabs API key, and it follows Jellyfin's standard plugin structure (PluginConfiguration + provider wiring in `CoreAppHost`). No default behavior changes.

You can grab a free API key at https://twelvelabs.io — there's a generous free tier. Happy to adjust to match your plugin conventions.

### DIFF
--- a/Jellyfin.Server/CoreAppHost.cs
+++ b/Jellyfin.Server/CoreAppHost.cs
@@ -25,6 +25,7 @@ using MediaBrowser.Controller.Library;
 using MediaBrowser.Controller.Lyrics;
 using MediaBrowser.Controller.Net;
 using MediaBrowser.Controller.Security;
+using MediaBrowser.Controller.Subtitles;
 using MediaBrowser.Controller.Trickplay;
 using MediaBrowser.Model.Activity;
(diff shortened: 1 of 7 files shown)


JPVenson:
we want to move away from build-in plugins you are free to publish that as a standalone plugin but that is not something we want build-in.

mohit-twelvelabs:
Makes total sense — respect the move away from built-in plugins. I'll publish this as a standalone Jellyfin plugin instead so folks who want it can opt in. Thanks for taking a look! 🙏

— Mohit (@mohit-twelvelabs, TwelveLabs)

PR #17191, opened 2026-06-26
Fix path transversal exposure in Plugins

IDisposable:
The request path is not validated to a valid path and could allow escaping the transcode path and downloading of any arbitrary file in GetHlsPlaylistLegacy .

GetHlsAudioSegmentLegacy and GetHlsVideoSegmentLegacy have the same issue, and are NOT behind an Authorize so they are publicly exploitable.

**Changes**

Added a ValidateTranscodePath that verifies that requested file paths start with the transcode path setting. Also ensure that all filename comparisons are OrdinalIgnoreCase because we might be running on a filesystem where filename-casing doesn't have to match. Switched from InvariantCulture because the underlying OS filename comparisons are always byte-wise (with case insensitivity here).

Fixed a similar issue in GetPluginImage

**Code assistance**

None

**Issues**

None (recent security patch didn't address all cases)

### DIFF
--- a/Jellyfin.Api/Controllers/HlsSegmentController.cs
+++ b/Jellyfin.Api/Controllers/HlsSegmentController.cs
@@ -60,11 +60,8 @@ public class HlsSegmentController : BaseJellyfinApiController
     public ActionResult GetHlsAudioSegmentLegacy([FromRoute, Required] string itemId, [FromRoute, Required] string segmentId)
     {
         // TODO: Deprecate with new iOS app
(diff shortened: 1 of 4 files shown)


Bond-009 (approved):
This could use some tests

Shadowghost (approved):

IDisposable:
> This could use some tests

Good point, added them and found another edge case... where there's a plugin at "transcodePath-evil/" I wasn't putting the path separator on the end so it would pass the starts-with test.

PR #17200, opened 2026-06-29
Don't gate -readrate on EnableSegmentDeletion for HLS stream-copy

filipghoulin:
\`-readrate 10\` is applied for HLS stream-copy jobs to prevent premature ffmpeg exit on fast drives — the existing comment says exactly this.

Currently gated on \`EnableSegmentDeletion\`, which is off by default. No comment or doc explains why the protection should only apply when segment deletion is on; looks like an oversight.

Removing the gate so the limit applies to all HLS stream-copy jobs. The existing \`-readrate_catchup\` handles recovery from momentary slowdowns, so this doesn't regress seeking.

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -7349,14 +7349,13 @@ namespace MediaBrowser.Controller.MediaEncoding
                 readrate = 1;
                 inputModifier += " -re";
             }
-            else if (encodingOptions.EnableSegmentDeletion
-                && state.VideoStream is not null
+            else if (state.VideoStream is not null
                 && state.TranscodingType == TranscodingJobType.Hls
                 && IsCopyCodec(state.OutputVideoCodec)
                 && _mediaEncoder.EncoderVersion >= _minFFmpegReadrateOption)
             {
-                // Set an input read rate limit 10x for using SegmentDeletion with stream-copy
-                // to prevent ffmpeg from exiting prematurely (due to fast drive)
+                // Limit read rate for all HLS stream-copy jobs to prevent premature exit
+                // on fast drives, not just when segment deletion is enabled.
                 readrate = 10;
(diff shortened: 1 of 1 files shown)


MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
```diff
@@ -7349,14 +7349,13 @@ public string GetInputModifier(EncodingJobInfo state, EncodingOptions encodingOp
...
-            else if (encodingOptions.EnableSegmentDeletion
-                && state.VideoStream is not null
+            else if (state.VideoStream is not null
                 && state.TranscodingType == TranscodingJobType.Hls
                 && IsCopyCodec(state.OutputVideoCodec)
                 && _mediaEncoder.EncoderVersion >= _minFFmpegReadrateOption)
             {
-                // Set an input read rate limit 10x for using SegmentDeletion with stream-copy
-                // to prevent ffmpeg from exiting prematurely (due to fast drive)
+                // Limit read rate for all HLS stream-copy jobs to prevent premature exit
```
nyanmisaka, line 7357: Users and clients may want transcoding tasks to complete and exit as soon as possible. The default behavior is to impose no limits on transcoding/remuxing speeds, unless the user opts in to speed throttling.

Therefore, a check is always required. For your use case, the check should be modified to `(encodingOptions.EnableThrottling || encodingOptions.EnableSegmentDeletion) && ...`

nyanmisaka (changes requested):

gnattu:
> No comment or doc explains why the protection should only apply when segment deletion is on; looks like an oversight.

No, it is not an oversight because there is no good reason to keep ffmpeg running if there is no reason to.

filipghoulin:
Misunderstood the design — the gate is intentional. When segment deletion is off, letting ffmpeg run at full speed is correct (all segments pre-cached, fast seeking). Rate-limiting it here would only slow that down for no benefit. Thanks for the quick clarification.

PR #17221, opened 2026-07-02
Undo breaking change to "BreakOnNonKeyFrames"

### LINKED ISSUE #16621: Cannot serialize member 'BreakOnNonKeyFrames' of type bool?
### Description of the bug

In the DLNA plugin, all DLNA device profiles (both built-in system profiles and custom user profiles) fail to parse after the changes introduced in [PR #15926](https://github.com/jellyfin/jellyfin/pull/15926). This causes the DLNA plugin to fall back to the `Generic Device` profile for every client, regardless of any custom or built-in profile matching.

The `BreakOnNonKeyFrames` property was changed from `bool` to `bool?` (nullable). 

`System.Xml.Serialization.XmlSerializer` cannot serialize `Nullable<bool>` as an `[XmlAttribute]`. The serializer fails at type reflection time, before it even attempts to read any XML content, which means every profile file fails to parse.

## Fix

Replace `[XmlAttribute]` with `[XmlIgnore]` on the obsolete `BreakOnNonKeyFrames` property. The property is already marked `Obsolete` and should not participate in XML serialization.

This preserves the nullable type for any code that references it, while preventing the `XmlSerializer` reflection failure.

I can submit a PR.

### Reproduction steps

1. Use Jellyfin with DLNA plugin
2. Note all profile parsing errors in logs
3. All clients use Generic profile

Errors deserializing profiles, all clients use Generic profile

No errors deserializing profiles, custom profiles correctly loaded.

### Jellyfin Server version

10.11.8

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

Release

### Environment

```markdown
- OS: Windows 11
- Browser: Chrome
- Plugins: DLNA
```

### Jellyfin logs

```shell
[INF] [12] Jellyfin.Plugin.Dlna.DlnaManager: Using user profile directory "C:\ProgramData\Jellyfin\Server\plugins\configurations\dlna\user"
[ERR] [12] Jellyfin.Plugin.Dlna.DlnaManager: Error parsing profile file: "C:\ProgramData\Jellyfin\Server\plugins\configurations\dlna\user\Samsung TVG.xml"
System.InvalidOperationException: There was an error reflecting type 'Jellyfin.Plugin.Dlna.Model.DlnaDeviceProfile'.
 ---> System.InvalidOperationException: There was an error reflecting property 'TranscodingProfiles'.
 ---> System.InvalidOperationException: There was an error reflecting type 'MediaBrowser.Model.Dlna.TranscodingProfile'.
 ---> System.InvalidOperationException: There was an error reflecting property 'BreakOnNonKeyFrames'.
 ---> System.InvalidOperationException: Cannot serialize member 'BreakOnNonKeyFrames' of type System.Nullable`1[System.Boolean]. XmlAttribute/XmlText cannot be used to encode complex types.
   at System.Xml.Serialization.XmlReflectionImporter.ImportAccessorMapping(MemberMapping accessor, FieldModel model, XmlAttributes a, String ns, Type choiceIdentifierType, Boolean rpc, Boolean openModel, RecursionLimiter limiter)
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

Daniel-Svensson:
This fixes introduced regression in DLNA profile that caused issues due to the XmlSerializer does not support "Nullable<bool>"

This should hopefully fix https://github.com/jellyfin/jellyfin/issues/16621

**Changes**
Revert the change of BreakOnNonKeyFrames  from bool to "Nullable<bool>"

**Issues**
* #16621
* https://github.com/jellyfin/jellyfin-plugin-dlna/issues/224

### DIFF
--- a/MediaBrowser.Model/Dlna/TranscodingProfile.cs
+++ b/MediaBrowser.Model/Dlna/TranscodingProfile.cs
@@ -143,7 +143,7 @@ public class TranscodingProfile
     [DefaultValue(false)]
     [XmlAttribute("breakOnNonKeyFrames")]
     [Obsolete("This is always false")]
-    public bool? BreakOnNonKeyFrames { get; set; }
+    public bool BreakOnNonKeyFrames { get; set; }
 
     /// <summary>
     /// Gets or sets the profile conditions.
--- a/tests/Jellyfin.Model.Tests/Dlna/SerializationTests.cs
+++ b/tests/Jellyfin.Model.Tests/Dlna/SerializationTests.cs
@@ -1,0 +1,98 @@
+using System;
+using System.IO;
+using Emby.Server.Implementations.Serialization;
+using MediaBrowser.Model.Dlna;
+using Xunit;
+
+namespace Jellyfin.Model.Tests.Dlna;
+
+public class SerializationTests
+{
+    [Fact]
+    public void DeviceProfile_XmlSerialization()
+    {
+        var serializer = new MyXmlSerializer();
+        var original = new DeviceProfile
+        {
+            Name = "Test Device",
+            Id = Guid.NewGuid(),
+            MaxStreamingBitrate = 9500000,
+            MaxStaticBitrate = 8500000,
+            MusicStreamingTranscodingBitrate = 192000,
+            MaxStaticMusicBitrate = 5000000,
+            DirectPlayProfiles =
+            [
+                new DirectPlayProfile
+                {
+                    Container = "mp4,mkv",
+                    AudioCodec = "aac,ac3",
+                    VideoCodec = "h264,hevc",
+                    Type = DlnaProfileType.Video
+                }
+            ],
+            TranscodingProfiles =
+            [
+                new TranscodingProfile
+                {
+                    Container = "ts",
+                    Type = DlnaProfileType.Video,
+                    VideoCodec = "h264",
+                    AudioCodec = "aac",
+                    SegmentLength = 6,
+                    Conditions =
+                    [
+                        new ProfileCondition(ProfileConditionType.Equals, ProfileConditionValue.VideoBitDepth, "8")
+                    ]
+                }
+            ],
+            ContainerProfiles =
+            [
+                new ContainerProfile
+                {
+                    Type = DlnaProfileType.Video,
+                    Container = "mkv",
+                    SubContainer = "matroska",
+                    Conditions =
+                    [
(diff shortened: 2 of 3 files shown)


nielsvanvelzen:
This property is deprecated and pending removal, as such it must be kept nullable until it's actually removed per our API deprecation process.

Daniel-Svensson:
@nielsvanvelzen why do you need to make a **BREAKING CHANGE** when adding obsolete?
cannot it be keept a boolean and still be obsolete? 

It is **not normal to make breaking API changes when obsoleting properties**, the normal approach is to first obsolete methods and then make the breaking change **later**. The purpose of the attribute is to signal that you will make changes later and give time to react BEFORE the actual change.
 
 
 As the code stands it would be better to just remove the property so you do the breaking change once instead of twice.
 
SITE NOTE:, it would make sense to remove the XmlAttribute in favor of `XmlIgnore` before you remove the property

nielsvanvelzen:
Moving a property from non-null to nullable is not a breaking change.

Daniel-Svensson:
> Moving a property from non-null to nullable is not a breaking change.

Maybe not for reference types, where nullability is only encoded as metadata and it might only be a source breaking change.
But it **certainly IS** for value types where you change the actual type, if you don't trust me or the linked issues then refer to https://learn.microsoft.com/en-us/dotnet/core/compatibility/library-change-rules#properties-fields-parameters-and-return-values 

> ❌ DISALLOWED: Changing the returned values for a property, field, method return value, or [out](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/method-parameters#out-parameter-modifier) **parameter**

It is **BOTH** a binary breaking change and a source breaking change, consider `bool variable = device.Property` which will fail in both occurrences.

In this case the change also broke the serialization framework used for the class since the property was changed to an unsupported type.

MediaBrowser.Model/Dlna/TranscodingProfile.cs
```diff
@@ -143,7 +143,7 @@ public TranscodingProfile(TranscodingProfile other)
     [DefaultValue(false)]
     [XmlAttribute("breakOnNonKeyFrames")]
```
Daniel-Svensson, line 144:
```suggestion
    [XmlIgnore]
```
Should the property be removed from (de)serialization now or later ?

PR #15575, opened 2025-11-19
Fix unreachable local API URLs in NAT/Docker by respecting Published Server URIs

martenumberto:
### Description
This PR fixes an issue where Jellyfin returns unreachable internal IP addresses (e.g., Docker container IPs like `172.17.0.x`) to clients when `GetApiUrlForLocalAccess` is called. This behavior breaks functionality such as Live TV Direct Stream on Android TV clients running in a LAN environment different from the container's internal network.

**The Problem:**
Previously, `GetApiUrlForLocalAccess` checked for CLI startup overrides but ignored the "Published Server URIs" configured in the Web Dashboard (`NetworkConfiguration`). Consequently, even if a user configured a correct LAN URL (e.g., via `all=` or `internal=`), the server would still broadcast its internal container IP for local access requests.

**The Solution:**
The method has been updated to respect the `NetworkConfiguration`. The logic priority is now:
1. CLI/Startup Override (`--published-server-url`)
2. Dashboard Configuration `internal=` override (Best for LAN access in NAT)
3. Dashboard Configuration `all=` override
4. Fallback to detected Network Interface IP (Standard behavior)

### Related Issues
Fixes #15411

### How to test
1. Run Jellyfin in a Docker container (bridge mode).
2. Navigate to **Dashboard -> Networking**.
3. Set **Published Server URIs** to your actual LAN IP or Domain (e.g., `all=http://192.168.1.50:8096`).
4. Connect a client (e.g., Android TV) and attempt to play a Live TV stream that triggers the `GetApiUrlForLocalAccess` logic (often Direct Stream).
5. **Before fix:** Client receives `http://172.x.x.x:8096...` and fails/timeouts.
6. **After fix:** Client receives `http://192.168.1.50:8096...` or `https://mydomain.com...` and plays successfully.

Issue #15411, opened 2025-11-05
Providing wrong Video Path to clients


martenumberto:
### Description of the bug

Jellyfin 10.11.2 (also earlier Minor Versions) provides the wrong Video Path to the Clients (Android & AndroidTV):
It tells them to open a unreachable Path (for e.g. http://172.19.0.3:8096/LiveTv/LiveStreamFiles/cf7be597148d422eb0c57dd724b0eacb/stream.ts)
This is my Docker IP and not the URL i set in the Network Settings and results in a java.net.SocketTimeoutException.
After that Jellyfin tries to transcode the File, provides another Path (with my Custom URL/Domain) to the Clients and Playback starts.

### Reproduction steps

1. Setup a M3U Tuner in Jellyfin
2. Try to Play a Live Channel
3. Look at the Logcat of the Client (e.g. Jellyfin-AdnroidTV and Jellyfin-Android)
4. See the wrong Path set in the Log
5. Watch it for 10 seconds till the SocketTimeoutException kicks in

### What is the current _bug_ behavior?

Jellyfin Provides the Local (Docker) path instead of the "Publish Server URI" Path. It doesnt matter if its LiveTV or a Local File

### What is the expected _correct_ behavior?

Tell the Clients to use "https://mydomain.com/...." instead of the local (Docker) IP (http://172.19.0.3:8096/...)

### Jellyfin Server version

Master

### Specify the build version

10.11.2

### Environment

```markdown
- OS: Ubuntu 24
- Linux Kernel:
- Virtualization: PVE, Docker
- Clients: WebClient, Jellyfin-Android, Jellyfin-AndroidTV
- Browser: Chrome
- FFmpeg Version: 7.1.2-Jellyfin
- Playback Method: Preffered DirectStream
- Hardware Acceleration: None, QSV
- GPU Model: Intel iGPU
- Plugins: -
- Reverse Proxy: nginx
- Base URL: "all=https//mydomain.com"
- Networking: nginx on 172.19.0.2 (known proxy)
- Jellyfin Data Storage: docker volume
- Media Storage: docker volume, nfs
- External Integrations: strm files
```

### Jellyfin logs

```shell
[2025-11-05 19:14:00.298 +00:00] [INF] [32] Jellyfin.Api.Helpers.MediaInfoHelper: User policy for "MyUser". EnablePlaybackRemuxing: True EnableVideoPlaybackTranscoding: True EnableAudioPlaybackTranscoding: True
[...]

PR #17224, opened 2026-07-02
Fix/livetv published url override

WizardOfYendor1:
Fix - Live TV stream URLs are leaking the docker-internal IP to clients in some cases (in particular using a reverse proxy on an internal LAN).

**Changes**
GetApiUrlForLocalAccess was passing a null source IP into NetworkManager.GetBindAddress. This applies a "published server URL" overrides when a URL is specified. 
With no override applied, it fell back to the server's own network interface and used the docker bridge IP (e,g. 172.23.0.5) into the live TV paths (SharedHttpStream etc) instead of the configured public/reverse-proxy address. This broke Live TV direct-play for clients outside the container's own network. 

**Code assistance**
Used Claude Code to trace the bug end-to-end and to correlated it to live ADB Fire TV device logcat output. 

**Issues**
This issue seems to be similar and/or related to this closed PR...but not quite the same.
Related to jellyfin/jellyfin#15411 and supersedes the approach in the closed jellyfin/jellyfin#15575.

Still a major work in process as the FFMPEG issues noted in the closed 15575 PR needs addressing.

### DIFF
--- a/Emby.Server.Implementations/ApplicationHost.cs
+++ b/Emby.Server.Implementations/ApplicationHost.cs
@@ -955,8 +955,16 @@ namespace Emby.Server.Implementations
         /// <inheritdoc/>
         public string GetApiUrlForLocalAccess(IPAddress ipAddress = null, bool allowHttps = true)
         {
-            // With an empty source, the port will be null
-            var smart = NetManager.GetBindAddress(ipAddress, out _, false);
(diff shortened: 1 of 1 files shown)


WizardOfYendor1:
Closing this and opening a new PR with a more comprehensive approach.

Issue #14689, opened 2025-08-22
Media playback randomly starts over again or jumps back - STILL


quick-red-fox:
### This issue respects the following points:

- [x] This is a **bug**, not a question or a configuration issue; Please visit our [forum or chat rooms](https://jellyfin.org/contact/) first to troubleshoot with volunteers, before creating a report.
- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [x] I'm using an up to date version of Jellyfin Server stable, unstable or master; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Description of the bug

Someone closed my previous issue without actually addressing it. My problem is accross ALL CLIENTs, not just Chrome. Android TV APP and Android Jellyfin APP as well. Please don't close it until it is resolved.

When opening specific video files the playback on the client's device (doesn't matter what kind of device is it) randomly starts over after a period of time or at random points of the media playback. It is always the same length on the same devices until it starts over.
In some cases it occurs after 15-30 seconds after initiating the media, in some cases after maybe 1-3 seconds and in other occassions at random points of the played media.

Until this report I tested the following:

- restarted Jellyfin server (no change)
- disabled all plugins then reinstalled them (no change)
- switched to software encoding instead of hardware encoding (no change)
- made dummy new profiles, started media playback - at first it was good, after a few tries the issue came back just like it was before (temporary fix)
- tested 3 different devices accross the connected ones, multiple times, with all of the tests I did
- deleted cache on all devices used for playback (no change)
- readded the server on all devices used for playback (no change)
- tested the media itself besides Jellyfin to ensure it is not broken (they were not)
- installed GPU driver upgrade, made no change (no change)
- reinstalled Jellyfin using the same installer version as it is now (10.10.7) - made no change
- I store the trickplay images next to the media, deleted them, made no difference
- When I fast forward the part where it stucks and restarts, I solves the issue temporary but it comes back the next time I play the same media or an other one.
- I can't define what causes it becase the time it takes to restarts the media depends on the media itself and user/device combo being used.
[...]

PR #17199, opened 2026-06-29
Add audiobook support for playback rate, CUE chapters, & multi file chapters

jordanmichaelrushing:
**Changes**
- Add atempo transcoding of audiobook files when client requests non 1.0x playback rate (accepts optional parameters) (46d66cfcf25b5a3e729ab84d43421bb12a977c01)
- Add CUE file parsing and setting for audiobook chapters (3c659a61ea12391ecee809e51b162a6dfadb9c1a)
- Handle multiple audiobook files in the same folder and build them as chapters, sending multiple files (7a01626a260b487d296b1c02f1a856e978377d3c)
    - Store multifile artwork so crossing chapters (aka files) shows that files artwork (if different) (For Example: Graphic Audio has their files in parts and has unique art for each major part) (7a01626a260b487d296b1c02f1a856e978377d3c)
- Standardized the needed file naming to squash multiple audio files into a singular Item with chapters (will need to update the Docs) (7a01626a260b487d296b1c02f1a856e978377d3c)

**Code assistance**
- I used Opus to identify the issue and where I would need to make the changes.
- I used Opus to also look over my changes to make any suggestions
- It suggested the fallback of >2 and < 0.5
- Had Opus help me with tests, only code it wrote

**Issues**
- https://github.com/jellyfin/jellyfin/issues/148 (This is already closed, but only because feature requests were moved elsewhere. Also the consensus then was a frontend only fix, which unfortunately does not work for iOS while it's not on a native player.
- https://features.jellyfin.org/posts/590/adjust-audiobook-playback-speed
- https://features.jellyfin.org/posts/1242/android-app-playback-speed-for-audio
- https://features.jellyfin.org/posts/2552/autoplay-next-audiobook-chapter

**Associated Changes**
- [Web Change](https://github.com/jellyfin/jellyfin-web/pull/8107)

**Notes**
- Commit 3 (multifile support) has what I would consider the most controversial change: `item is AudioBook)` in `Emby.Server.Implementations/Library/MediaSourceManager.cs`. I really tried to not touch any main files, but this one was a necessity to make multifiles work with GUID (which is needed to pass multiple files)

PR #17226, opened 2026-07-03
Preserve resume position on near-start playback reports

jordanmichaelrushing:
**Changes**
- Store the previous position before attempting to overwrite play position to help prevent an accidental overwrite of position to 0:00 on HLS transcoding error

**Code assistance**
- Opus for code research
- Opus for reviewing my code change
- Opus for building the table below
- No code written by AI

**Issues**
I searched a bunch of terms in the issues page but found nothing exactly like this. The closest I could find was https://github.com/jellyfin/jellyfin/issues/14689, but my solution I don't think would solve their underlying issue as they wouldn't have gotten past the min % for video or min minutes for audiobooks. However myself and 2 others who use my server reproduced the issue independently. I traced it down to this code path. The gist is that resuming an item that needs transcoding and the server runs into an issue, can result in a state where the resume time gets reset to 0:00.

**Product callout**
There is a slight change in behavior that is worth calling out here, but I think it's worth the change. On Master (this includes 10.11.x) if you're at 50% progress and you decide to set the progress to 4% for some reason and then close out the app, any resumes will start at 0%. For manually seeking this is fine, it's a server setting, makes sense. The issue is if this was either accidental or the existing HLS pattern of starting from 0 then seeking to the resume position errors out on the server side and the user closes the client, it can report that the resume position at that point is 0:00. For movies & audiobooks, this really stinks if you're hours in and an error happens and your progress is lost, especially when you don't know exactly where you were.

I had Claude build a table to help explain the UX patterns:
  | Action | Master | With this PR |
  |---|---|---|
  | Seek back near the start, keep playing | Continues; resume moves forward again once you pass the threshold | Same |
  | Pause near the start, then resume (same session) | Continues from where you paused | Same |
  | **Stop, or close the app, near the start, then reopen** | **Starts over from the beginning** (your progress is lost) | **Resumes where you last were** |
  | **A background/phantom session reports a near-start position** | **Progress lost, starts over** | **Progress kept** |
  | Drag to the very start (0:00) and stop | Starts from the beginning | Starts from the beginning |

### DIFF
--- a/Emby.Server.Implementations/Library/UserDataManager.cs
+++ b/Emby.Server.Implementations/Library/UserDataManager.cs
@@ -312,7 +312,9 @@ namespace Emby.Server.Implementations.Library
         /// <inheritdoc />
         public bool UpdatePlayState(BaseItem item, UserItemData data, long? reportedPositionTicks)
         {
+            var previousPositionTicks = data.PlaybackPositionTicks;
             var playedToCompletion = false;
+            var zeroedNearStart = false;
 
             var runtimeTicks = item.GetRunTimeTicksForPlayState();
 
@@ -328,6 +330,7 @@ namespace Emby.Server.Implementations.Library
                 {
                     // ignore progress during the beginning
                     positionTicks = 0;
+                    zeroedNearStart = true;
                 }
                 else if (pctIn > _config.Configuration.MaxResumePct || positionTicks >= (runtimeTicks - TimeSpan.TicksPerSecond))
                 {
@@ -355,6 +358,7 @@ namespace Emby.Server.Implementations.Library
                 {
                     // ignore progress during the beginning
                     positionTicks = 0;
+                    zeroedNearStart = true;
                 }
                 else if (remainingTimeInMinutes < _config.Configuration.MaxAudiobookResume || positionTicks >= runtimeTicks)
                 {
@@ -368,6 +372,11 @@ namespace Emby.Server.Implementations.Library
                 // If we don't know the runtime we'll just have to assume it was fully played
                 data.Played = playedToCompletion = true;
                 positionTicks = 0;
+            }
+
+            if (zeroedNearStart && previousPositionTicks > 0)
+            {
+                positionTicks = previousPositionTicks;
             }
 
             if (!item.SupportsPlayedStatus)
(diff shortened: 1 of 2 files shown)


Shadowghost (approved):

jordanmichaelrushing:
To give context into why I have this change, there is some bug that I have spent several days trying to track down on one of my client's Android phone. Specifically when she presses pause, turns off her car, and the Bluetooth disconnects, it sends a stop position at 0 ms, which was resetting her stored position to 0 ms (she was hours into an audiobook). I've tried to determine why this happens and had a few theories but it still keeps coming back. I'm going to keep on trying to fix that underlying issue but I still think that it's important to also fix this on the server side to prevent this from happening from any potential other future client issues, whether it's the web repo or Android or iOS. 

I have a couple of other fixes specifically for this issue of only honoring a position of zero when it is not a stop but a storing of position. That way if someone drags their progress to zero, it will honor that but if it's this phantom zero millisecond issue, it won't honor that. It's a little bit of a rats nest of changes so I'm going to put this into a draft state until I feel more comfortable about what changes I think should actually be applied here. Will update when it's good to go again.

jordanmichaelrushing:
Ok I think I went too far and the issue is really with the Android app sending 0ms as the stop position, not near 0. The thought was that the HLS pattern is "start by fetching the amount playhead first then offset" that if it errors out between the playhead and offset calls, that with changes like I have on https://github.com/jellyfin/jellyfin/pull/17199 where multiple files all go into a singular Item that the playhead starting at 30 seconds but the offset is actually multiple hours in the future (because we're on file 15 of 30, but it's 30 seconds into file 15) might lead to the server storing 30 seconds, which is this "near zero" situation which would restart the playhead to actual 0ms if the error resets the app / play etc. 

The error I'm actually seeing is the Android app seems to be throwing an error when bluetooth is abruptly disconnected and the app sends "playback stopped at 0ms" which is not the same thing as "near zero". Two very different issues. I have a fix for the 0ms on the Web app I'm testing and I do have a fix for the Server side. I'm going to keep this PR up in case someone does think this is a worthwhile change still for 2 weeks. Happy to close it out as well. My 0ms fix will be in a separate PR from this.

jordanmichaelrushing:
Going to split into 2 PRs:
- One that handles 0ms instead of near 0 (written but not in a PR state just yet)
- Deterministically selecting the correct data based on current key: https://github.com/jellyfin/jellyfin/pull/17320

Going to close this one out

Issue #16484, opened 2026-03-26
upgrade from libpgs-js to libbitsub (better performance, PGS & VOB/DVD support) for web clients & LG WebOS


DrHurt:
Jellyfin currently uses libpgs-js to render PGS subtitles in Web Browsers and web-based clients like LG WebOS.

May I suggest moving to this more modern, performant fork which uses Web Assembly, WebGPU / WebGL and adds support for VOB (DVD subtitles) allowing us to watch DVD rips without needing a video transcode.

It's under MIT license so compatible with Jellyfin and as a fork of libpgs-js, it should not be too difficult to implement.

https://github.com/altqx/libbitsub

Edit 1: Jellyfin doesn't currently extract VobSub (DVD) subtitles due to an ffmpeg limitation.
Edit 2: Jellyfin's ffmpeg could copy VOB to an mks container, and libbitsub could extract & render sub/idx files client-side.
Edit 3: We now have a pull request to integrate libbitsub into Jellyfin https://github.com/jellyfin/jellyfin-web/pull/7777

Neptune altqx (the dev) is is currently looking into implementing an mks extractor as part of the workflow.

Thanks in advance!

DrHurt:
@nyanmisaka I believe you worked on this before and that you also have an LG TV?

This is a relatively simple change that will provide a tangible improvement to Jellyfin users who rip BluRays and DVDs.

Thanks

nyanmisaka:
> [@nyanmisaka](https://github.com/nyanmisaka) I believe you worked on this before and that you also have an LG TV?
> 
> This is a relatively simple change that will provide a tangible improvement to Jellyfin users who rip BluRays and DVDs.
> 
> Thanks

It wasn't me. It was the authors of libpgs-js who did some testing on LG TV.

Perhaps you should consult him about the possibility of merging the two projects. This would allow for a smoother migration process downstream.

DrHurt:
@Arcus92 

Would you be able to merge VOB support from the above fork into your project to help us get VOB working with Jellyfin?

Thanks

Arcus92:
Hi @DrHurt,
It's been a while but I worked on a VobSub in the original project. The issue was that VOB subtitles are actually two files - a .SUB and .IDX. Requesting both files for one subtitle track would require some backend changes. Not impossible, but doable.

[...]

PR #17227, opened 2026-07-03
Match VobSub MKS subtitle profiles by container

altqx:
**Changes**
Match VobSub subtitle profiles using `Format=vobsub` with `Container=mks` instead of treating `mks` as a subtitle format. This lets clients advertise support for rendering VobSub delivered in an MKS container without implying support for arbitrary MKS subtitles. Add focused tests for internal VobSub extraction, external `.mks` delivery, and external `.idx/.sub` fallback behavior.

**Code assistance**
Codex was used to inspect the subtitle profile matching flow, implement the matcher and test updates, and run focused validation.

**Issues**
Continues #16484

### DIFF
--- a/MediaBrowser.Model/Dlna/StreamBuilder.cs
+++ b/MediaBrowser.Model/Dlna/StreamBuilder.cs
@@ -576,11 +576,8 @@ namespace MediaBrowser.Model.Dlna
                     foreach (var profile in subtitleProfiles)
                     {
                         if (profile.Method == SubtitleDeliveryMethod.External
-                            && (string.Equals(profile.Format, stream.Codec, StringComparison.OrdinalIgnoreCase)
-                                // FFmpeg cannot mux VobSub back into an .idx/.sub pair, so extracted VobSub streams are exposed as .mks.
-                                || (string.Equals(profile.Format, "mks", StringComparison.OrdinalIgnoreCase)
-                                    && stream.IsVobSubSubtitleStream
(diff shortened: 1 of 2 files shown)


gnattu (approved):

gnattu:
A note that this might trigger weird server side subtitle conversion and get failed as vobsub cannot be converted, but it should not hurt as it does not support conversion even from the beginning.

Shadowghost (approved):

nyanmisaka (approved):

dfederm:
> A note that this might trigger weird server side subtitle conversion and get failed as vobsub cannot be converted, but it should not hurt as it does not support conversion even from the beginning.

I think it might be more harmful than you think. The match keeps `Format=vobsub`, so the URL is `Stream.vobsub`, but extraction reports `mks`, so `GetSubtitles` misses the same-format fast path and tries to convert to vobsub (no writer). The intended MKS delivery never happens. Requesting `Stream.mks` would return the extracted file directly.

crobibero:
> > A note that this might trigger weird server side subtitle conversion and get failed as vobsub cannot be converted, but it should not hurt as it does not support conversion even from the beginning.
> 
> I think it might be more harmful than you think. The match keeps `Format=vobsub`, so the URL is `Stream.vobsub`, but extraction reports `mks`, so `GetSubtitles` misses the same-format fast path and tries to convert to vobsub (no writer). The intended MKS delivery never happens. Requesting `Stream.mks` would return the extracted file directly.

Merging as is, if it causes issues we can revert

PR #17240, opened 2026-07-04
Remove unreachable conditions and redundant null checks

iderex:
SonarCloud (S2589) flagged a handful of conditions that can never take their other branch, so the checks were just dead weight:

- LibraryManager / MediaStreamSelector: Enumerable.ToArray()/ToList() never return null, so the '?? []' and '?? BehaviorOnlyForced(...)' fallbacks were unreachable. The MediaStreamSelector fallback never ran, so dropping it doesn't change behavior.
**Changes**

SonarCloud (rule S2589) flagged a few conditions whose "other" branch can
never be reached, so the checks were just dead weight. This cleans them up
without changing runtime behavior:

- `LibraryManager` / `MediaStreamSelector`: `Enumerable.ToArray()` / `.ToList()`
  never return `null`, so the `?? []` and `?? BehaviorOnlyForced(...)` fallbacks
  were unreachable.
- `EncodingHelper`: in both the QSV (d3d11) and RKMPP filter chains the early
  guard already returns whenever the OpenCL path isn't supported, so the final
  `if (…OclSupported)` is always `true` by the time we reach it. I return the
  preferred pipeline directly and drop the now-unreachable trailing return.
- `StreamBuilder.GetDefaultAudioBitrate`: after the `< 2 channels` branch has
  returned, `audioChannels` can no longer be `null`, so `audioChannels ?? 0`
  was redundant. I hoisted it into a local so it's computed once.

No behavior change is intended — each change resolves to exactly what the code
already did at runtime. Existing tests for all three areas pass.

Note: the `MediaStreamSelector` comment mentions an "otherwise OnlyForced"
fallback that, because of the never-null `.ToList()`, never actually ran. This
PR only removes the dead fallback; wiring it up for real would be a behavior
change and is left for a separate PR.

**Code assistance**

Claude Code was used to verify with data-flow
reasoning that each branch is genuinely unreachable, and run the build and
tests. I reviewed every change.

**Issues**

### DIFF
--- a/Emby.Server.Implementations/Library/MediaStreamSelector.cs
+++ b/Emby.Server.Implementations/Library/MediaStreamSelector.cs
@@ -134,7 +134,7 @@ namespace Emby.Server.Implementations.Library
             {
                 // Always load (full/non-forced) subtitles of the user's preferred subtitle language if possible, otherwise OnlyForced behavior.
                 filteredStreams = sortedStreams.Where(s => !s.IsForced && MatchesPreferredLanguage(s.Language, preferredLanguages))
-                    .ToList() ?? BehaviorOnlyForced(sortedStreams, preferredLanguages);
+                    .ToList();
             }
             else if (mode == SubtitlePlaybackMode.OnlyForced)
             {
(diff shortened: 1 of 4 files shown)


Emby.Server.Implementations/Library/MediaStreamSelector.cs
```diff
@@ -134,7 +134,7 @@ public static void SetSubtitleStreamScores(
             {
                 // Always load (full/non-forced) subtitles of the user's preferred subtitle language if possible, otherwise OnlyForced behavior.
```
Shadowghost, line 135: Since the comment is now stale, we should remove it.

Shadowghost (approved):

PR #17242, opened 2026-07-05
Add endpoint for targeted media additions

PhilLaser:
## Summary

- add `POST /Library/Media/Added` for directly adding newly available media paths
- resolve each submitted path under the nearest existing Jellyfin folder parent
- return per-path results instead of reporting the parent folder to the library monitor

## Motivation

`POST /Library/Media/Updated` is useful for filesystem change notifications, but it reports paths to the library monitor and can cause parent-folder refreshes. External automation tools (e.g. Autopulse/Autoscan) that already know the exact new file path need a narrower API surface, especially when media is on mounts that do not provide reliable inotify events.

This endpoint lets callers submit exact media paths and lets Jellyfin resolve/insert the missing child item directly.

## Testing

```text
dotnet build Jellyfin.Api/Jellyfin.Api.csproj --configuration Release
```

Result:

```text
Build succeeded.
0 Warning(s)
0 Error(s)
```

# EDIT 16 July 2026
## Use case

In setups with remote filesystems (nfs, smb, etc.) this PR would allow users to add media immediately (10-20sec after download is finished) to jellyfin instead of having to wait for the next scheduled library scan to happen. 

I'm running this PR since ~3 weeks in an Arr-setup together with a patched version of Autopulse that calls this new endpoint and did not experience any issues so far.

### DIFF
--- a/Jellyfin.Api/Controllers/LibraryController.cs
+++ b/Jellyfin.Api/Controllers/LibraryController.cs
@@ -32,6 +32,7 @@ using MediaBrowser.Model.Configuration;
 using MediaBrowser.Model.Dto;
 using MediaBrowser.Model.Entities;
 using MediaBrowser.Model.Globalization;
+using MediaBrowser.Model.IO;
 using MediaBrowser.Model.Net;
 using MediaBrowser.Model.Querying;
 using Microsoft.AspNetCore.Authorization;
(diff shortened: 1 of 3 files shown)


IDisposable (approved):

PhilLaser:
for informational purposes: extended the initial comment of the PR to cover real world use case

PR #17246, opened 2026-07-06
Show production companies under TV Shows' Studios

Rant423:
**Changes**
The Studios section of TV Shows actually shows networks.

This change will add production companies to it.

Possible follow-up: split Production Companies from Networks.

**Code assistance**
no

**Issues**
n/a

### DIFF
--- a/MediaBrowser.Providers/Plugins/Tmdb/TV/TmdbSeriesProvider.cs
+++ b/MediaBrowser.Providers/Plugins/Tmdb/TV/TmdbSeriesProvider.cs
@@ -258,14 +258,14 @@ namespace MediaBrowser.Providers.Plugins.Tmdb.TV
 
             var studios = Enumerable.Empty<string>();
 
-            if (seriesResult.ProductionCompanies is not null)
-            {
-                studios = studios.Concat(seriesResult.ProductionCompanies.Select(i => i.Name).OfType<string>());
-            }
-
             if (seriesResult.Networks is not null)
             {
                 studios = studios.Concat(seriesResult.Networks.Select(i => i.Name).OfType<string>());
+            }
+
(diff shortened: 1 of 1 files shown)


Shadowghost:
This is intentional an needs to be decoupled but this will require changes on clients too. I'd suggest to rather add both (Networks + Production companies) rather than switching.

Rant423:
> This is intentional an needs to be decoupled but this will require changes on clients too. I'd suggest to rather add both (Networks + Production companies) rather than switching.

Having both would be great, but it's a much bigger change than I wanted to make 😄 (since there is only Studios, it would mean adding a new Networks)

So basically this was the simpler change to make things right (Studios = Production companies). If someone else want to tackle adding both I'll be grateful

Shadowghost:
I wonder if we should just add both to studios for now until we can decouple it

Rant423:
> I wonder if we should just add both to studios for now until we can decouple it

Good idea, so nothing is lost. Will do it

Shadowghost (approved):

dkanada (approved):

theguymadmax (changes requested):
Please move the `if (seriesResult.Networks is not null)` block before the `if (seriesResult.ProductionCompanies is not null)` block so that networks are listed first, followed by production companies, in the list.

Rant423:
> Please move the `if (seriesResult.Networks is not null)` block before the `if (seriesResult.ProductionCompanies is not null)` block so that networks are listed first, followed by production companies, in the list.

Alright, done

theguymadmax (approved):

Rant423:
> ... until we can decouple it

@Shadowghost: https://github.com/jellyfin/jellyfin/pull/17904 and https://github.com/jellyfin/jellyfin-web/pull/8447

PR #17248, opened 2026-07-06
Add Novel job mapping to the Writing department

theguymadmax:
**Changes**
Add Novel job mapping to the Writing department

**Code assistance**
N/A

**Issues**
Novelist like J. R. R. Tolkien were not being listed

### DIFF
--- a/MediaBrowser.Providers/Plugins/Tmdb/TmdbUtils.cs
+++ b/MediaBrowser.Providers/Plugins/Tmdb/TmdbUtils.cs
@@ -48,6 +48,16 @@ namespace MediaBrowser.Providers.Plugins.Tmdb
             PersonKind.Producer
         };
 
+        /// <summary>
+        /// Writing jobs to keep.
+        /// </summary>
+        private static readonly HashSet<string> WriterJobs = new(StringComparer.OrdinalIgnoreCase)
+        {
+            "writer",
+            "screenplay",
+            "novel"
+        };
+
         [GeneratedRegex(@"[\W_-[·]]+")]
         private static partial Regex NonWordRegex();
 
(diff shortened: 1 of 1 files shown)


Shadowghost (approved):

dkanada (approved):

MediaBrowser.Providers/Plugins/Tmdb/TmdbUtils.cs
```diff
@@ -48,6 +48,16 @@ public static partial class TmdbUtils
             PersonKind.Producer
         };
 
+        /// <summary>
+        /// Writing jobs to keep.
+        /// </summary>
+        private static readonly HashSet<string> WriterJobs = new(StringComparer.OrdinalIgnoreCase)
```
Bond-009:
```suggestion
        private static readonly HashSet<string> _writerJobs = new(StringComparer.OrdinalIgnoreCase)
```
And not sure if a hash set makes sense for such a small collection, if so, a FrozenSet probably makes more sense

MediaBrowser.Providers/Plugins/Tmdb/TmdbUtils.cs
```diff
@@ -48,6 +49,16 @@ public static partial class TmdbUtils
...
         };
 
+        /// <summary>
+        /// Writing jobs to keep.
+        /// </summary>
+        private static readonly FrozenSet<string> _writerJobs = new[]
+        {
+            "writer",
+            "screenplay",
+            "novel"
```
Rant423, line 59: Should we also have "Short Story"?

e.g. Philip K. Dick for Minority Report (https://www.themoviedb.org/movie/180-minority-report/cast)

theguymadmax:
@Rant423

I don't think we need to add back all the job roles, it was narrowed in https://github.com/jellyfin/jellyfin/pull/15872

The goal would be to match jobs featured on the TMDB overview page. 

Novel is featured:
lotr

While Short Story is not:
mr

Bond-009 (approved):

PR #17251, opened 2026-07-06
Fix stuck HLS transcode when job dies mid-playlist

jordanmichaelrushing:
**Changes**
When a segment's playlist file already exists on disk but its transcoding job has been killed, the handler would silently do nothing and wait indefinitely for a segment that would never be produced. Now detects the null job and restarts ffmpeg instead of hanging.

Also avoids logging a misleading "cannot serve" warning when the segment actually exists but no transcode job is running, and adds debug logging around segment requests for diagnosing transcode stalls.

**Code assistance**
Code review from Opus

**Issues**
None that I'm aware of. This was something I saw in my logs when moving from wifi to cell or cell to wifi, I believe a secondary job was created because of the new connection and the original was abandoned but my logs were getting bombarded by a wall of `cannot serve {0} as it doesn't exist and no transcode is running` lines. So I investigated and fixed the issue.

### DIFF
--- a/Jellyfin.Api/Controllers/DynamicHlsController.cs
+++ b/Jellyfin.Api/Controllers/DynamicHlsController.cs
@@ -1456,7 +1456,9 @@ public class DynamicHlsController : BaseJellyfinApiController
 
         var segmentExtension = EncodingHelper.GetSegmentFileExtension(state.Request.SegmentContainer);
 
-        TranscodingJob? job;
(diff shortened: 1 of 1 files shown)


gnattu (changes requested):
This is a bit risky and a null job does not mean the old ffmpeg process has stopped as ffmepg is not that stable and it can have a lot of reason to get stuck and/or killed. If we want to make the lifecycle handling better the code you changed is not the right place.

Plus, reusing output from previous transcoding job is not guaranteed to be compatible as network condition changes can easily trigger transcoding parameter changes, and that requires a fresh HLS manifest.

jordanmichaelrushing:
I'll keep this one local then. Thanks

PR #17252, opened 2026-07-07
Fix artists being displayed with albums

theguymadmax:
After https://github.com/jellyfin/jellyfin/pull/15954, the album view returns both artist and albums

**Changes**
Remove `BaseItemKind.MusicArtist` from the `CollectionType.music` case in `GetBaseItemKindsForCollectionType`

**Code assistance**
N/A

**Issues**
1a

### DIFF
--- a/Emby.Server.Implementations/Images/CollectionFolderImageProvider.cs
+++ b/Emby.Server.Implementations/Images/CollectionFolderImageProvider.cs
@@ -31,6 +31,12 @@ namespace Emby.Server.Implementations.Images
             var includeItemTypes = DtoExtensions.GetBaseItemKindsForCollectionType(viewType);
             var recursive = viewType != CollectionType.playlists;
 
+            if (viewType == CollectionType.music)
+            {
+                // Music albums usually don't have dedicated backdrops, so use artist instead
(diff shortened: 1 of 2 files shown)


Shadowghost (approved):

dkanada:
Are you sure this isn't a web issue? It looks like the old value in #15954 was MusicArtist, not MusicAlbum, which makes me question why it's inverted now.

theguymadmax:
Before, we were only using this to return what the image collection folder should use for images.  For music, we overrode it to use artists instead of albums, since albums don't typically have backdrops.  I added a commit to restore the original behavior for the collection images

We introduced a new [behavior](https://github.com/jellyfin/jellyfin/pull/15954/changes#diff-b40abb5fd1633280ce8a0982c06a4e6d804829d1fd4167282471b8f3f2601f8eR319) (this bug), which incorrectly adds artist as a default return type for albums, which is wrong - copies the the override we were using to get better music collection folder images.

Definitely not a web issue; this fix is correct. It affects all clients, such as ATV:

1b

dkanada (approved):

Bond-009 (approved):

PR #17254, opened 2026-07-07
Add additional attribute aliases and improve attribute detection

sjakub:
**Changes**

This adds support for two additional attribute aliases: `tvdb` for `tvdbid` and `imdb` for `imdbid`.

It also fixes an issue that was causing secondary detections to run on modified string range.
It was causing (existing) `tmdb` alias to be ignored if it was followed by something like `tmdbidxyz`,
or IMDB pattern matching to not work if the pattern was followed by something like `imdbidxyz`.
In that case those two tests would fail:

```
("Superman: Red Son [tmdb=618355][tmdbid1=1]", "tmdbid", "618355")
("Superman: Red Son - tt10985510 [imdbid1=tt11]", "imdbid", "tt10985510")
```

Also, if `tmdb` was followed by `tmdbid`, the second attribute would be selected instead of the first one and this test would fail:

```
("Superman: Red Son {tmdb-618355}{tmdbid=1234567}", "tmdbid", "618355")
```

Now the first recognized attribute occurrence is selected, alias or not.

Finally, it ignores attributes with values consisting of only whitespaces, so tests like this won't fail:

```
("[tmdbid= ][tmdbid=223344]", "tmdbid", "223344")
```

**Code assistance**
N/A

**Issues**
N/A


Shadowghost (approved):

theguymadmax (approved):

sjakub:
Hi, I just noticed that in the alias switch it would be nicer to use `ReadOnlySpan<char>.Empty` instead of `string.Empty`. But since it's already been approved, I don't know if I should be adding more commits to this... Should I? :)

Shadowghost:
Feel free to change it

crobibero (approved):

PR #17255, opened 2026-07-07
Add TVDB provider ID support for movies

theguymadmax:
We were only parsing `tmdbid` and `imdbid` for movies, so movies could never be identified via a `tvdbid` provider tag, even though series already support this.

**Changes**
Add `tvdbid` parsing support to the movie resolver

**Code assistance**
N/A

**Issues**
N/A


Shadowghost (approved):

IDisposable (approved):
I find the location of the inner function quite confusing, can we move it below the call-sites?

Bond-009 (approved):

PR #17287, opened 2026-07-10
support external images for audiobooks

dkanada:
**Changes**

At the moment only embedded images are supported for audiobooks. This adds external image support, which is currently referenced in our documentation.

**Code Assistance**

None

**Issues**

None

### DIFF
--- a/MediaBrowser.LocalMetadata/Images/LocalImageProvider.cs
+++ b/MediaBrowser.LocalMetadata/Images/LocalImageProvider.cs
@@ -84,7 +84,7 @@ namespace MediaBrowser.LocalMetadata.Images
             if (item.SupportsLocalMetadata)
             {
                 // Episode has its own provider
-                if (item is Episode || item is Audio || item is Photo)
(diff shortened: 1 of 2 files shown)


theguymadmax (approved):

Shadowghost (approved):

theguymadmax:
@dkanada Can you also add a commit to allow saving images locally at this line: https://github.com/jellyfin/jellyfin/blob/53aafcd38e1f4558ff18f5258d0d46b3a0565783/MediaBrowser.Providers/Manager/ImageSaver.cs#L93

This way, when users enable the option in the library, it will save to the local folder instead of the internal metadata path, as it currently does.

theguymadmax (approved):
Thanks, the option to save next to media works now for audiobooks.

Bond-009 (approved):

jordanmichaelrushing:
I'm putting together a PR for OpenLibrary to add support there as well utilizing this, so thank you for making the change here!

PR #17290, opened 2026-07-10
Fix oversized, mis-aligned, outline-less burned-in mov_text (tx3g) subtitles

Olsro:
This is my first contribution to Jellyfin. I found a bug that bothered me, I am satisfying functionnally with how great this patch works so I want to share it with the community. Thank you

**Code assistance / AI usage**
Code generated by Sonnet 5

I hand tested manually with several contents containing tx3g subtitles to ensure subtitles now looks great in all common transcoded resolution (480p, 720p).

When you will test my patch and switch between several versions of Jellyfin, don't forget to enter your config folder then run this between each test ```sudo find . -iname "*.ass" -delete -print``` : it's crucial to clear the cache so Jellyfin recreates the subtitles each time you want to do a new test.

**Issues**
All of my content containing tx3g subtitles appeared wrong in a very wrong way when burn + transcoding is done by Jellyfin. It seems like no one bothered creating an issue for this problem. It happens only if you burn tx3g subtitles into the video stream. I decided to make screenshots so you can visually understand what the problem is.

Directly loading an affected content on VLC3 on MacOS (I used this as a visual reference, as it looks really great and wanted Jellyfin to produce a similar result for me) : 
image

Transcoding with subtitles burn on that content : 
image

As you can see it's very bad & too disturbing to watch the content this way, which is why I spent time iterating with Claude on this subject.

Transcoding with subtitles burn on that content, with the content from this patch : 
image

Looks very satisfying to see and compares very well to what VLC provides. It's comfortable to read.

**Code changes**
- Extract mov_text to ASS instead of SubRip, passing the mov_text decoder's own -width/-height options so PlayResX/PlayResY match the real video.
- Recenter left-column alignment (both in the Style definition and any per-line \an override).
- Force OutlineColour/BackColour to fully opaque black.
- Replace the track's own Fontsize with one derived from the video's height (calibrated against real-world testing across burn-in pipelines), scaling MarginV and Outline to match, and rescaling any per-line {\fsNN} emphasis overrides proportionally to the new default.

### DIFF
--- a/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
+++ b/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
@@ -501,7 +501,7 @@ namespace MediaBrowser.MediaEncoding.Subtitles
                 && mediaSource.VideoStream is { Width: > 0, Height: > 0 };
         }
 
-        private string GetExtractableSubtitleFormat(MediaStream subtitleStream, MediaSourceInfo mediaSource)
+        private static string GetExtractableSubtitleFormat(MediaStream subtitleStream, MediaSourceInfo mediaSource)
         {
             if (string.Equals(subtitleStream.Codec, "ass", StringComparison.OrdinalIgnoreCase)
                 || string.Equals(subtitleStream.Codec, "ssa", StringComparison.OrdinalIgnoreCase)
@@ -523,7 +523,7 @@ namespace MediaBrowser.MediaEncoding.Subtitles
             }
         }
 
-        private string GetExtractableSubtitleFileExtension(MediaStream subtitleStream, MediaSourceInfo mediaSource)
+        private static string GetExtractableSubtitleFileExtension(MediaStream subtitleStream, MediaSourceInfo mediaSource)
         {
             // Using .pgssub as file extension is not allowed by ffmpeg. The file extension for pgs subtitles is .sup.
             if (string.Equals(subtitleStream.Codec, "pgssub", StringComparison.OrdinalIgnoreCase))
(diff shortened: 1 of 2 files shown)


Olsro:
If anyone has any interest to merge this, I will be motivated to fix the conflicts. Just warn me, thanks

MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
```diff
@@ -1082,6 +1137,119 @@ private async Task SetAssFont(string file, CancellationToken cancellationToken =
...
+                {
+                    var originalOverride = int.Parse(m.Groups[1].Value, CultureInfo.InvariantCulture);
+                    var scaled = (int)Math.Round(targetFontSize * (double)originalOverride / originalFontSize, MidpointRounding.AwayFromZero);
+                    return "\\fs" + scaled.ToString(CultureInfo.InvariantCulture);
+                });
+            }
+
+            return _assOverrideAlignmentRegex.Replace(
+                newText,
+                m => "\\an" + _movTextLeftToCenterAlignment[m.Groups[1].Value]);
```
gnattu, line 1250: I'm not sure if such changes are right because, although this fixes your use case, you are unconditionally force all left aligned tx3g subtitle to be center aligned.

MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
```diff
@@ -1082,6 +1137,119 @@ private async Task SetAssFont(string file, CancellationToken cancellationToken =
...
+                }
+
+                fields[2] = targetFontSize.ToString(CultureInfo.InvariantCulture);
+                // ffmpeg writes OutlineColour/BackColour as 8-digit &HAABBGGRR with the alpha
+                // byte set to "ff". ASS uses an inverted alpha (00 = opaque, ff = fully
+                // transparent), so this renders the outline and shadow completely invisible
+                // regardless of Outline/Shadow width. PrimaryColour/SecondaryColour are
+                // unaffected because ffmpeg writes those as 6-digit RRGGBB with no alpha byte.
+                fields[5] = "&H00000000";
+                fields[6] = "&H00000000";
```
gnattu, line 1224: Similar here, you are forcing all outline/background color to opaque black which discards the intended styling.

MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
```diff
@@ -757,19 +784,25 @@ private async Task ExtractAllExtractableSubtitlesInternal(
...
-                    outputPath);
+                mapOptions.Append(CultureInfo.InvariantCulture, $" -map 0:{streamIndex} -an -vn -c:s {outputCodec}{outputFormatOption} -flush_packets 1 \"{outputPath}\"");
             }
 
+            var args = string.Format(CultureInfo.InvariantCulture, "{0} -i {1}{2}", inputOptions, inputPath, mapOptions);
+
             if (outputPaths.Count > 0)
             {
-                await ExtractSubtitlesForFile(inputPath, args, outputPaths, cancellationToken).ConfigureAwait(false);
+                await ExtractSubtitlesForFile(inputPath, args, outputPaths, movTextOutputPaths, videoStream?.Height ?? 0, cancellationToken).ConfigureAwait(false);
```
gnattu, line 805: How does this handle the conflicts with other subtitles? For example, what if the file already has an ass subtitle?

gnattu (changes requested):
The changes are too aggressive IMO.

Yes for your specific use case, the forcefully restyled subtitles do look much better, but I still question if it is suitable to make such forced restyling a standard jellyfin extraction process.

The normalization you are doing here could very likely break some other users' use case can this is not what we want to do.

MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
Olsro: (I am not a C# dev, so for the sake of transparency the answer below was AI-generated but looks decent to me so I kept it as it is)

ffmpeg's mov_text_tx3g() skips the BoxRecord entirely (tx3g_ptr += 8), so the ASS Alignment it emits comes from the justification byte with the text box that gave it meaning already discarded — "left" ends up meaning "left of the whole frame", not "left of the box". The reference renderer doesn't honour that byte at all: VLC's substx3g.c hardcodes region.align = SUBPICTURE_ALIGN_BOTTOM.

So I've kept the remap, but narrowed it to the Style: line only — the per-line {\anN} remap is gone, since movtextdec.c never emits \an in the first place (it only writes \q \r \b \i \u \fs \fn \1c \1a \2c). That regex was dead code, and removing it also clears one of the SonarCloud findings.

Happy to drop the Style-line remap too if you'd rather not carry a heuristic — the size fix is independent of it and is the part that actually makes these tracks watchable.

MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
Olsro: (I am not a C# dev, so for the sake of transparency the answer below was AI-generated but looks decent to me so I kept it as it is)

Agreed, that was too broad. Narrowed so nothing authored is discarded: an opaque black outline is substituted only when ffmpeg produced a fully transparent OutlineColour, i.e. when the track authored no background and the text would otherwise have no border at all. Any colour the track does declare is kept verbatim, and BorderStyle is never touched.

Context: tx3g has no outline concept, so mov_text_init() feeds the track's background colour and alpha into the ASS OutlineColour slot. With ASS's inverted alpha, a track with no background arrives as &Hff…… and renders nothing. Worth noting ffmpeg's own mov_text muxer writes back_alpha = 255, which arrives as &H0 — opaque black, not transparent — so remuxed tracks keep the outline they have today. All three branches are covered by tests.

MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs
Olsro: (I am not a C# dev, so for the sake of transparency the answer below was AI-generated but looks decent to me so I kept it as it is)

-width/-height are private AVOptions of the mov_text decoder, emitted with a per-stream specifier, so they only reach that one tx3g stream; every other subtitle stream goes through -c:s copy and is never decoded. The normalization is scoped the same way and now runs in the caller rather than inside ExtractSubtitlesForFile, which also drops the signature change that two other call sites on master depend on.

While tightening this I also found that the ASS output format and file extension were chosen unconditionally for mov_text while -width/-height and the normalization were gated on knowing the video dimensions — a source without them would have produced un-normalized ASS at libavcodec's 384x288 fallback, worse than the SubRip it replaced. All three now go through a single ShouldExtractMovTextAsAss check, and the MKS path derives its output codec from the same helper as its extension instead of hardcoding srt.

PR #17298, opened 2026-07-11
Fix Live TV returning unreachable "server-local" streaming URLs to clients.

WizardOfYendor1:
Fix Live TV returning unreachable "server-local" streaming URLs to clients.
  
 **Changes**

   When Jellyfin is running in a docker container (etc) using a bridged network with reverse proxy in front of it, the Live TV playback URL contains the container's internal address, such as `172.x.x.x` in the streaming URL. Thus clients cannot connect. This change rewrites the url returned to clients and keeps the local Live TV buffer URL local for Jellyfin and FFmpeg. 

Rewriting is limited to live streaming, Jellyfin-hosted, /LiveTv/LiveStreamFiles/ paths.  Plug-ins can evade this check, which is of questionable value. With this approach the server keeps using the local address internally (FFmpeg, recording), while the client is given a URL it can actually reach. e.g. FFmpeg keeps the local address; the client gets the public one.
  
I also took the opportunity to fix the url/port parsing issues I encounter/noticed. This makes http:// and https:// explicit and also handles IPv6 port parsing. This is a separate commit on purpose as it has impact to all callers of the GetSmartApiUrl - so it can be pulled out because kinda out of scope, IMO. 

  **Code assistance**

  Codex and Opus was used for codebase research, planning and code review. Fable was used to code review and analyze the later state of my changes. Agentic coding used to generate unit tests and refactoring in some instances from findings in code reviews. Additionally, LLM was used to refine some of my rambling comments and explanations :-). 

  **Testing**
  
  Tested locally in this bridged container scenario and outside scenario(s) (e.g. "host" network) with and without a reverse proxy in front (aka "it works on my machine" :) ). More testing is suggested for scenarios I haven't thought of or considered.

  **Issues**
  
  Related to jellyfin/jellyfin#15411.
  
  Uses an approach with the changes/comments in mind of this closed PR jellyfin/jellyfin#15575 . 
  
  I did not address the `.strm` URL  case discussed in PR  15696 as I think it's a different domain of an issue (though similar) and it's    being tracked there.

### DIFF
--- a/Jellyfin.Api/Helpers/MediaInfoHelper.cs
+++ b/Jellyfin.Api/Helpers/MediaInfoHelper.cs
@@ -555,11 +555,6 @@ public class MediaInfoHelper
             return;
         }
 
-        if (mediaSource.Protocol != MediaProtocol.Http)
-        {
-            return;
-        }
-
         var baseUrl = _serverConfigurationManager.GetNetworkConfiguration().BaseUrl;
         var publishedPath = GetPublishedLiveStreamPath(_appHost.GetSmartApiUrl(request), mediaSource.Path, mediaSource.Protocol, baseUrl);
 
@@ -613,6 +608,20 @@ public class MediaInfoHelper
             return null;
         }
 
-        return smartApiUrl.TrimEnd('/') + relativePath;
+        var prefix = smartApiUrl.TrimEnd('/');
+        if (!string.IsNullOrEmpty(baseUrl))
+        {
+            var includesBaseUrl = Uri.TryCreate(prefix, UriKind.Absolute, out var publishedUri)
+                && Uri.UnescapeDataString(publishedUri.AbsolutePath)
+                    .TrimEnd('/')
+                    .EndsWith(baseUrl, StringComparison.OrdinalIgnoreCase);
+
+            if (!includesBaseUrl)
+            {
+                prefix += baseUrl;
+            }
+        }
+
+        return prefix + relativePath;
     }
 }
--- a/Jellyfin.Api/Helpers/MediaInfoHelper.cs
+++ b/Jellyfin.Api/Helpers/MediaInfoHelper.cs
@@ -99,29 +99,7 @@ public class MediaInfoHelper
     {
         var result = new PlaybackInfoResponse();
 
-        MediaSourceInfo[] mediaSources;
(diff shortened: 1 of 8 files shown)


Jellyfin.Api/Helpers/MediaInfoHelper.cs
```diff
@@ -524,4 +538,81 @@ private void SetDeviceSpecificSubtitleInfo(StreamInfo info, MediaSourceInfo medi
...
+        // Opened live streams always carry a LiveStreamId; this excludes pre-open and plugin/remote sources.
+        if (string.IsNullOrEmpty(mediaSource.LiveStreamId))
+        {
+            return;
+        }
+
+        if (mediaSource.Protocol != MediaProtocol.Http)
+        {
+            return;
+        }
```
Shadowghost: `GetPublishedLiveStreamPath` checks this again

Shadowghost (commented):
When `JELLYFIN_PublishedServerUrl` env var is set, `GetSmartApiUrl` returns it verbatim.
If you then set `PublishedServerUrl=https://media.example.com` without the base path but also configure `BaseUrl=/jellyfin`, the rewrite strips `/jellyfin` from the local path and prepends a smart URL that lacks it.
I don't think this actually happens that often (since it is technically a misconfiguration), but I think it's worth fixing and a test

Jellyfin.Api/Helpers/MediaInfoHelper.cs
WizardOfYendor1: I removed the extra check... my thinking there was to exit early if it's not http. But given the infrequency of that scenario and how "cheap" it is to not early exit;  it's not a big deal. 
I spent a lot of time thinking and looking at the base path issue.  I've included your suggestion - at least as *I* understand this, along with a bunch of tests. There's an uncovered scenario if the reverse proxy rewrites the path.  There's no good way to deal with that, except perhaps read some X header to determine it is coming via a proxy. This is probably VERY uncommon but I thought I should attempt to help the user and log a warning.... I'm so-so this :-)

Shadowghost (approved):

crobibero (approved):

martenumberto:
Hey!
Do you think this is a similar issue? Looking forward to see a fix in v12.0 :)
-> https://github.com/jellyfin/jellyfin/pull/15696

PR #17299, opened 2026-07-11
Fix fMP4 HLS subtitle timestamp map offset

### LINKED ISSUE #16647: HLS subtitle X-TIMESTAMP-MAP is misaligned when using fMP4 segments (SegmentContainer=mp4)

Speenah:
**Changes**

WebVTT subtitles served over HLS always embedded a hardcoded `X-TIMESTAMP-MAP=MPEGTS:900000` (10s) offset, which is only correct for the legacy MPEG-TS segment container — fMP4 segments start at PTS 0, so clients end up rendering subtitles ~10s behind audio/video. This computes the correct offset from the actual HLS segment container in use (0 for `mp4`, 900000 otherwise) in `DynamicHlsHelper.AddSubtitles`, and exposes it as a `vttTimestampMapMpegts` query parameter threaded through to `SubtitleController`, which stays container-agnostic and just echoes whatever offset it's given — matching the approach originally implemented in #16786. Defaults to 900000 for any caller that doesn't pass it, preserving existing behavior.

Verified in a native tvOS AVPlayer client against a 10-bit HEVC source: subtitles were ~10s behind before the fix and in sync after. Also checked the `stream.vtt` response directly via `curl` — its `X-TIMESTAMP-MAP` header changed from `MPEGTS:900000` to `MPEGTS:0`.

**Code assistance**

Used Claude Code throughout: investigating the root cause against a local dev server and a custom WIP native tvOS client, referencing the prior closed PR (#16786) and its review feedback to land on the API-parameter design, implementing the change, and drafting this description.

**Issues**

Fixes #16647

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -144,6 +144,7 @@
  - [sorinyo2004](https://github.com/sorinyo2004)
  - [Soumyadip Auddy](https://github.com/SoumyadipAuddy)
  - [sparky8251](https://github.com/sparky8251)
+ - [Speenah](https://github.com/Speenah)
  - [spookbits](https://github.com/spookbits)
  - [ssenart](https://github.com/ssenart)
  - [stanionascu](https://github.com/stanionascu)
--- a/Jellyfin.Api/Controllers/SubtitleController.cs
+++ b/Jellyfin.Api/Controllers/SubtitleController.cs
@@ -50,6 +50,12 @@ public class SubtitleController : BaseJellyfinApiController
     private readonly ILogger<SubtitleController> _logger;
 
     /// <summary>
+    /// The default MPEGTS value used in the WebVTT X-TIMESTAMP-MAP header, matching the legacy MPEG-TS
+    /// segment container's 10 second PTS offset.
+    /// </summary>
+    internal const long DefaultVttTimestampMapMpegts = 900000;
+
+    /// <summary>
     /// Initializes a new instance of the <see cref="SubtitleController"/> class.
     /// </summary>
     /// <param name="serverConfigurationManager">Instance of <see cref="IServerConfigurationManager"/> interface.</param>
@@ -202,6 +208,7 @@ public class SubtitleController : BaseJellyfinApiController
     /// <param name="endPositionTicks">Optional. The end position of the subtitle in ticks.</param>
     /// <param name="copyTimestamps">Optional. Whether to copy the timestamps.</param>
     /// <param name="addVttTimeMap">Optional. Whether to add a VTT time map.</param>
+    /// <param name="vttTimestampMapMpegts">Optional. The VTT timestamp map MPEGTS value.</param>
     /// <param name="startPositionTicks">The start position of the subtitle in ticks.</param>
     /// <response code="200">File returned.</response>
     /// <returns>A <see cref="FileContentResult"/> with the subtitle file.</returns>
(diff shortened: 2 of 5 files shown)


gnattu:
Can you fix the CI format check? The rest LGTM

Speenah:
@gnattu Fixed in 8f2b5e4b9470e7a3e3c79564aa372475458bae3e

gnattu (approved):

Shadowghost (approved):

PR #17302, opened 2026-07-11
Keep authenticated user entity in sync with persisted login timestamps

### LINKED ISSUE #17301: Successful login immediately reverts the user's LastLoginDate (stale entity overwrite in LogSessionActivity)

- [x] This issue is **not** already reported on GitHub _(I've searched it)_.
- [x] I'm using an up to date version of Jellyfin Server stable (10.11.11), and I verified the code path is unchanged on master.

### Description of the bug

A user's `LastLoginDate` is written correctly on login, but is reverted (usually back to `null`) milliseconds later **by the same login request**. The result is that users who sign in through `/Users/AuthenticateByName` keep showing "never logged in" in the dashboard, seemingly forever.

The sequence (verified on 10.11.11 by API testing and reading the source; the same code exists on master):

1. `SessionManager.AuthenticateNewSessionInternal` calls `UserManager.AuthenticateUser(..., isUserSession: true)`, which loads the `user` entity and then persists the login timestamps via `ExecuteUpdateAsync`:
   https://github.com/jellyfin/jellyfin/blob/v10.11.11/Jellyfin.Server.Implementations/Users/UserManager.cs#L614-L627
   `ExecuteUpdateAsync` bypasses the EF change tracker, so the in-memory `user` entity that `AuthenticateUser` returns still carries the **old** `LastLoginDate`/`LastActivityDate` values. It also does not bump `RowVersion`.

2. Back in `AuthenticateNewSessionInternal`, that stale entity is passed to `LogSessionActivity`:
   https://github.com/jellyfin/jellyfin/blob/v10.11.11/Emby.Server.Implementations/Session/SessionManager.cs#L1633-L1639

3. `LogSessionActivity` checks the stale `user.LastActivityDate` (old/`null`, so the "more than 60 seconds ago" guard passes), sets `user.LastActivityDate` and calls `UpdateUserAsync(user)` — a full-entity save of the stale object:
   https://github.com/jellyfin/jellyfin/blob/v10.11.11/Emby.Server.Implementations/Session/SessionManager.cs#L263-L279
   This writes the stale `LastLoginDate` (typically `null`) back over the value step 1 just persisted. Because step 1 didn't increment `RowVersion`, the `DbUpdateConcurrencyException` catch around this save never fires.

The bug is intermittent in a misleading way: if the user was active less than 60 seconds before signing in, the guard in step 3 skips the clobbering save and `LastLoginDate` survives. So users with constantly-active sessions show plausible values while users who only log in occasionally — arguably the ones the field matters most for — permanently show "never logged in".

`LastActivityDate` is unaffected overall because the clobbering save in step 3 sets it itself.

Master has the same pattern: the `ExecuteUpdateAsync` without refreshing the tracked entity (https://github.com/jellyfin/jellyfin/blob/master/Jellyfin.Server.Implementations/Users/UserManager.cs — `SetProperty(f => f.LastLoginDate, date)`), and the full-entity `UpdateUserAsync(user)` in `LogSessionActivity`.

### Reproduction steps

Using only the REST API against a 10.11.11 server (replace `$HOST`, `$APIKEY`):

1. Create a fresh user:
   ```bash
   curl -X POST "$HOST/Users/New?api_key=$APIKEY" -H 'content-type: application/json' \
     -d '{"Name":"lastlogin-test","Password":"test-pw-1234"}'
   ```
2. Authenticate as that user (exactly what every client does):
   ```bash
   curl -X POST "$HOST/Users/AuthenticateByName" -H 'content-type: application/json' \
     -H 'X-Emby-Authorization: MediaBrowser Client="curl", Device="curl", DeviceId="repro", Version="1.0.0"' \
     -d '{"Username":"lastlogin-test","Pw":"test-pw-1234"}'
   ```
3. Read the user back (any amount of time later):
   ```bash
   curl "$HOST/Users/{userId}?api_key=$APIKEY"
   ```

Observed: `LastLoginDate` is `null` (while `LastActivityDate` is set to the login time). Repeating step 2 and 3 never changes this. The server log shows `Authentication request for "lastlogin-test" has succeeded.` and no errors — the timestamp write happens and is then overwritten.

Expected: `LastLoginDate` reflects the successful authentication.

### Suggested fix

In `UserManager.AuthenticateUser`, after the `ExecuteUpdateAsync`, also update the returned in-memory entity:

```csharp
user.LastActivityDate = date;
user.LastLoginDate = date;
```

That keeps the entity consistent with the database, which both prevents `LogSessionActivity` from clobbering the row (its full-entity save would now carry the fresh values) and makes its 60-second guard correctly skip the redundant write during login. Happy to submit this as a PR if the approach sounds right.

### Jellyfin Server version

10.11.11 (verified); the affected code is unchanged on master.

### Environment

```
- OS: Docker (Synology NAS)
- Clients tested: raw REST API (curl / fetch), so client-independent
```

ElianCodes:
**Changes**
Sets `LastActivityDate`/`LastLoginDate` on the in-memory user entity right after `AuthenticateUser` persists them with `ExecuteUpdateAsync`.

**Issues**
Fixes #17301

**Why**

`ExecuteUpdateAsync` bypasses the EF change tracker, so the entity `AuthenticateUser` returns still carries the old timestamps (and `RowVersion` is not bumped). `SessionManager.AuthenticateNewSessionInternal` passes that stale entity to `LogSessionActivity`, whose 60-second guard then passes (stale `LastActivityDate`) and which saves the entity in full via `UpdateUserAsync` — writing the stale `LastLoginDate` (usually `null`) back over the value persisted milliseconds earlier, without tripping the `DbUpdateConcurrencyException` catch.

Net effect: users who sign in show "never logged in" indefinitely, unless they happened to be active within 60 seconds before signing in (which makes the bug look intermittent). Full analysis with an API-only reproduction is in #17301, verified against 10.11.11.

Keeping the entity in sync fixes the revert and also lets the activity guard correctly skip the redundant full-row write during login.

### DIFF
--- a/Jellyfin.Server.Implementations/Users/UserManager.cs
+++ b/Jellyfin.Server.Implementations/Users/UserManager.cs
@@ -616,6 +616,12 @@ namespace Jellyfin.Server.Implementations.Users
                                 .SetProperty(f => f.LastActivityDate, date)
                                 .SetProperty(f => f.LastLoginDate, date))
                             .ConfigureAwait(false);
+
+                        // ExecuteUpdateAsync bypasses the change tracker, so keep the
+                        // returned entity in sync. Otherwise SessionManager.LogSessionActivity
+                        // saves this (stale) entity in full and reverts LastLoginDate.
+                        user.LastActivityDate = date;
+                        user.LastLoginDate = date;
                     }
 
                     await dbContext.Users


theguymadmax (approved):

reecertc (approved):
The fix correctly keeps the returned user entity in sync with the persisted login timestamps and prevents `LastLoginDate` being overwritten. Don't see any issues.

Shadowghost (approved):

PR #17304, opened 2026-07-11
Normalize invalid PTS from containers for Trickplay generation

### LINKED ISSUE #17294: Trickplay generation runs away indefinitely on files with corrupted Duration metadata
### Description of the bug

If a video file's container reports an implausible Duration value (wildly disagreeing with its own real frame count and actual playable length), "Generate Trickplay Images" does not detect or reject this. It runs away, generating vastly more frame images than the real video length could ever justify, and never completes on its own. Observed: one ~42-minute episode produced 60,000+ trickplay frame JPEGs (implying ~169 hours of "content") before being manually cancelled, still climbing when killed.

### Reproduction steps

1. Have a video file whose container reports an implausible Duration (e.g. years) that wildly disagrees with its real frame count and actual playable length.
2. Trigger trickplay generation for that item (library scan, or manually via the "Generate Trickplay Images" scheduled task).
3. Observe: trickplay generation for that item never completes and continues writing frame images to the temp directory indefinitely (observed: 60,000+ frames and still climbing before being manually killed).

The file had a corrupted `Duration` value in its container metadata (confirmed independently of Jellyfin, via `ffprobe -show_format -show_streams` run directly on the file):
- `duration=322843082.752000` (seconds), implying ~3,736 days (~10.2 years)
- `TAG:DURATION=89678:38:15.485000000`, the same absurd value, embedded in the container itself
- `TAG:NUMBER_OF_FRAMES=60936`
- `r_frame_rate=24/1`, `avg_frame_rate=24/1` (confirmed constant frame rate, not VFR)

60,936 frames at a real 24fps CFR = 2,539 seconds ≈ 42.3 minutes, a completely normal episode. The real video content and per-frame timing were fine. Only the container's stated Duration was corrupted (the corruption itself originates outside Jellyfin, from a third-party muxing tool, but Jellyfin's handling of it is the actual bug being reported here).

Jellyfin's own `RunTimeTicks` for this item (`GET /Items/{id}`) matched the corrupted value exactly, confirming the library scan trusted the corrupted container metadata with no sanity check against the file's own frame count or actual stream length.

When "Generate Trickplay Images" then ran on this file, the `fps=0.1` sampling filter used to extract one trickplay frame per 10 seconds did not correctly decimate to that rate. Instead of respecting the real ~24fps native frame timing, it emitted close to one output frame per real decoded frame. This is consistent with its internal time-pacing being computed against the corrupted multi-year duration rather than the real, sane per-frame timestamps.

- The task hangs on this item indefinitely (severity depends on the "Scan Behavior" setting), making no real progress toward completion.
- Cancelling the task does not reliably clean up the temp directory afterward (see #15426 below, which compounds this): a single file with this kind of corrupted metadata can accumulate an effectively unbounded number of orphaned temp files on disk before anyone notices.

- Sanity-check a file's Duration against its own frame count and frame rate (or an independently-computed stream duration) before trusting it for trickplay frame-pacing. Fall back to a computed duration when the two disagree significantly.
- Regardless of root cause, cap trickplay frame generation at a sane hard limit (e.g. expected-duration/interval plus generous headroom) so a file with corrupted metadata fails cleanly and visibly instead of running away indefinitely.
- Ensure a cancelled or failed trickplay job's temp directory is always cleaned up (see #15426).

### Jellyfin Server version

10.11.10

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.10

### Environment

```markdown
- Jellyfin 10.11.10
- FFmpeg 7.1.3 (jellyfin-ffmpeg)
- Hardware acceleration: Intel QSV/VAAPI
- Trickplay settings: Image Interval 10000ms (0.1 fps), Width 320, hardware-accelerated MJPEG encoding enabled
```

### Jellyfin logs

```shell
n/a
```

### FFmpeg logs

```shell
n/a
```

### Client / Browser logs

n/a

n/a

## Suggested fix
- Sanity-check a file's Duration against its own frame count and frame rate (or an independently-computed stream duration) before trusting it for trickplay frame-pacing. Fall back to a computed duration when the two disagree significantly.
- Regardless of root cause, cap trickplay frame generation at a sane hard limit (e.g. expected-duration/interval plus generous headroom) so a file with corrupted metadata fails cleanly and visibly instead of running away indefinitely.
- Ensure a cancelled or failed trickplay job's temp directory is always cleaned up (see #15426).

nyanmisaka:
This change does not affect the keyframe only mode.

**Changes**
- Normalize invalid PTS from containers for Trickplay generation

**Issues**
- Fixes #17294

### DIFF
--- a/MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
+++ b/MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
@@ -926,6 +926,25 @@ namespace MediaBrowser.MediaEncoding.Encoder
                 throw new InvalidOperationException("EncodingHelper returned empty or invalid filter parameters.");
             }
 
+            // Normalize invalid PTS from containers for non keyframe only mode
+            if (!enableKeyFrameOnlyExtraction)
+            {
+                var fpsFilterIndex = filterParam.IndexOf("fps=", StringComparison.Ordinal);
+                if (fpsFilterIndex >= 0)
+                {
+                    var inputFrameRate = (imageStream.ReferenceFrameRate.HasValue && imageStream.ReferenceFrameRate > 0)
+                        ? imageStream.ReferenceFrameRate.Value : 30;
+
+                    var setPtsFilter = string.Create(CultureInfo.InvariantCulture, $"setpts=N/{inputFrameRate:F3}/TB,");
+
+                    filterParam = filterParam.Insert(fpsFilterIndex, setPtsFilter);
+                }
+                else
+                {
+                    throw new InvalidOperationException("EncodingHelper returned invalid filter parameters.");
+                }
+            }
+
             try
             {
                 return await ExtractVideoImagesOnIntervalInternal(


Shadowghost (approved):

gnattu (approved):

MediaBrowser.MediaEncoding/Encoder/MediaEncoder.cs
```diff
@@ -926,11 +926,28 @@ public async Task<string> ExtractVideoImagesOnIntervalAccelerated(
...
+            {
+                throw new InvalidOperationException("EncodingHelper returned invalid filter parameters.");
+            }
+
             try
             {
                 return await ExtractVideoImagesOnIntervalInternal(
                     (enableKeyFrameOnlyExtraction ? "-skip_frame nokey " : string.Empty) + inputArg,
-                    filterParam,
+                    enableKeyFrameOnlyExtraction ? filterParam : normalizedFilterParam,
```
Bond-009: Instead of this, why don't you modify `filterParam` if enableKeyFrameOnlyExtraction is false?
That way we only execute that branch when needed

nyanmisaka: Done

Bond-009 (approved):

PR #17309, opened 2026-07-12
Smooth SyncPlay ping tracking and tune staleness handling

riseprometheus:
### Smooth SyncPlay ping tracking and tune staleness handling

Hello there! I started poking around some of the syncplay functionality to learn more about the subsystem, and hopefully make some improvements to the desync/laggy behavior my friends and I experience when using syncplay for movie nights. While exploring some of the syncplay/timesync codebase I found logic about how ping spikes are handled when unpausing for a group that I wanted to take a crack at. Upfront I want to mention I used Claude to help with this investigation, and I welcome any corrections to my understanding of this feature since I actively want to learn more about it. 

Important note: I have a separate PR for [jellyfin-web](https://github.com/jellyfin/jellyfin-web/pull/8201) to fix an issue where the ping data was never actually being sent from the client side for the original latency compensation logic. As such, this change is directly tied to the fix to actually get ping data back since that is what I tested against. The jellyfin-web feature could likely be merged independently, but I am unsure if now triggering the originally intended logic would have undesirable side effects. 

Additionally, since this is my first PR to this project, please let me know if the additional debug lines are overly verbose. Happy to remove/tweak/update based on the preference of the maintainers. 

**Changes**
SyncPlay's latency compensation for group resume was driven by the last raw ping report from each client. A single noisy sample, or a client that stopped reporting altogether, could keep an inflated or stale delay in effect indefinitely with no recovery mechanism. This resulted in a flat 500ms for unpauses.

My update applies an asymmetric EWMA to incoming ping samples so the server reacts quickly to a sudden increase in latency but decays slowly on the way down. It also raises the staleness threshold from 30s to 90s since my  local testing against jellyfin-web showed its steady-state ping-reporting interval settles to 60 seconds after an initial burst.

Debug-level logging was added for the smoothed ping values, stale-fallback events, and resume-delay decisions, to make the compensation math observable against a deployment without attaching a debugger.

**Code assistance**
I used Claude code to help trace how how the timesync functionality works after exploring some of the group unpause behavior to learn why it can desync.

**Issues**

None that I could find while searching through issues.


Emby.Server.Implementations/SyncPlay/Group.cs
```diff
@@ -438,19 +463,61 @@ public void UpdatePing(SessionInfo session, long ping)
         {
             if (_participants.TryGetValue(session.Id, out GroupMember value))
             {
-                value.Ping = ping;
+                var previousPing = value.Ping;
+                if (value.LastPingUpdate == DateTime.MinValue)
```
dfederm: This only resets the average for the first ever sample. If a client reports a high ping, stops reporting long enough to be treated as stale, then reports a normal ping again, this path smooths against the old stale value and makes it current again. For example, 5000ms followed by a stale period then 100ms stores 4020ms, so the group goes back to waiting on a stale spike for several more samples. Treat a sample after `PingStaleAfterMs` like a fresh baseline, or reset from `DefaultPing`, and add a test for stale high ping followed by a fresh low ping.

dfederm (commented):
Found one correctness issue in the stale recovery path. A stale high ping can become active again as soon as that client sends a fresh low sample, which defeats the fallback for the first several reporting intervals.

Emby.Server.Implementations/SyncPlay/Group.cs
riseprometheus: Yeah, annoying miss on my part. I will add a quick check to evaluate how stale the last ping is.

riseprometheus: Added the updated check and a test. The more I look at this, the more I want to tweak how these client latencies are communicated to the server, but that can be an adventure for another day.

Emby.Server.Implementations/SyncPlay/Group.cs
```diff
@@ -438,19 +463,62 @@ public void UpdatePing(SessionInfo session, long ping)
         {
             if (_participants.TryGetValue(session.Id, out GroupMember value))
             {
-                value.Ping = ping;
+                var previousPing = value.Ping;
+                if ((DateTime.UtcNow - value.LastPingUpdate).TotalMilliseconds > PingStaleAfterMs)
+                {
+                   // Use the raw sample  when we don't have an established smoothed average
```
dfederm: Nit: indentation. This and the line below need 1 more space character

dfederm (approved):

Emby.Server.Implementations/SyncPlay/Group.cs
riseprometheus: Fixed.

riseprometheus:
Hi there! Is there anything else I need to for the PR?

fmarcac:
While reading this area I noticed a unit bug on one of the lines this PR touches, in `WaitingGroupState.HandleRequest(ReadyGroupRequest ...)`:

```csharp
delayTicks = highestPing * 2 * TimeSpan.TicksPerMillisecond;
delayTicks = Math.Max(delayTicks, context.DefaultPing);
```

The first line produces ticks, but `DefaultPing` is milliseconds (`Group.DefaultPing = 500`). So the intended 500 ms floor is actually 500 ticks, which is 0.05 ms, and it never applies. Any group whose highest ping is under 250 ms gets a resume delay of twice the highest ping with no lower bound at all.

`PlayingGroupState.HandleRequest(UnpauseGroupRequest ...)` does the same calculation entirely in milliseconds and is correct, which is what makes the difference easy to miss here.

The fix is:

```csharp
delayTicks = Math.Max(delayTicks, TimeSpan.FromMilliseconds(context.DefaultPing).Ticks);
```

This matters more after your change than before it, since falling back to `DefaultPing` for a stale session is exactly the case where the floor is supposed to do the work. Happy to leave it to you as part of this PR, or to send it separately if you would rather keep this one focused.

PR #17311, opened 2026-07-12
extract page count from archives and PDFs

dkanada:
**Changes**

This pull request generates accurate durations for all archive and PDF books. Note that database migrations are not required since UserData is already saving progress in this exact format. If you would prefer to save RunTimeTicks as the raw pageCount value then we'll need to migrate UserData entries and modify web to report pageCount instead of ticks.

I see this as a stopgap for #15633 which is why I don't use pageCount directly. I looked at several PDF libraries and settled on this one, but am fine with another choice. We only have two tasks for PDFs that I can see right now: extract page count and rasterize for streaming at some point.

itext: rasterization is through a commercial license
PdfSharpCore: no rasterization
PDFtoImage: supports page count and  rasterization
PDFsharp: somewhat inactive development

**Code assistance**

None

**Issues**

None

### DIFF
--- a/MediaBrowser.Providers/MediaInfo/ProbeProvider.cs
+++ b/MediaBrowser.Providers/MediaInfo/ProbeProvider.cs
@@ -24,6 +24,8 @@ using MediaBrowser.Model.Globalization;
 using MediaBrowser.Model.IO;
 using MediaBrowser.Model.MediaInfo;
 using Microsoft.Extensions.Logging;
+using PDFtoImage;
+using SharpCompress.Archives;
 
 namespace MediaBrowser.Providers.MediaInfo
 {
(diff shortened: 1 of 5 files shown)


Shadowghost (approved):

crobibero (approved):

MediaBrowser.Providers/MediaInfo/ProbeProvider.cs
```diff
@@ -24,6 +24,8 @@
 using MediaBrowser.Model.IO;
 using MediaBrowser.Model.MediaInfo;
 using Microsoft.Extensions.Logging;
+using PDFtoImage;
+using SharpCompress.Archives;
```
Bond-009, line 28: We should use System.IO.Compression instead of SharpCompress

PR #17315, opened 2026-07-13
Parse ReplayGain album gain field

Florin-Popescu:
**Changes**

Hello! First time contributor here.

I'm missing proper support for ReplayGain tags for volume normalization in Music Assistant. This PR parses the ReplayGain tags internally if present in the music files and exposes them on the server API to be usable by 3rd party clients such as Music Assistant. The following tags are parsed if available in music files:

- REPLAYGAIN_TRACK_GAIN
  - was already handled in the code base, I didn't change it
- REPLAYGAIN_ALBUM_GAIN
  - covers the measured loudness over an entire album. Here I chose to add it to the parent album of a track when parsed, since it's a tag in the track files themselves
  - I chose to only set it once for the parent album if the album doesn't have its NormalizationGain property already set. Normally all tracks in an album should have this tag set to the same value if the files are coming from the same source, so this implementation seems sensible to me vs. for example averaging the REPLAYGAIN_ALBUM_GAIN for all tracks in an album and setting that as the album's NormalizationGain
  - It is exposed by the server API using the field AlbumNormalizationGain when querying for a certain track as well as by the NormalizationGain field when querying for an album. The support for AlbumNormalizationGain field for a track was added in #14745 but to my understanding the album gain was internally populated (and therefore exposed by the API) only from jellyfin's own LUFS measurement, but not from a file's ReplayGain tags.
- REPLAYGAIN_REFERENCE_LOUDNESS
  - we can't always assume ReplayGain was ran with its default target loudness of -18 LUFS nor can clients make such assumptions. Music Assistant for example uses -17 LUFS by default for playback. Clients need this information to play the files at the correct volume
  - this required adding a new ReferenceLoudness field in the entity database. Hope I did it correctly, including the migration file.
  - I don't know if it's really necessary, but depending on how the clients parse the library, I thought it useful to add this information to a track's parent album in addition to the track itself. The gain information by itself is not very useful without the reference LUFS for either a track or an album. It's added in the same way as REPLAYGAIN_ALBUM_GAIN: only once if not already present in the album's properties, for the first track parsed for that album. Assumption is that all files were processed to add ReplayGain tags in an album using the same settings.
  - I'm exposing the ReferenceLoudness field in the DTO API only if NormalizationGain is also present. I thought it wouldn't be useful if only the reference loudness were present in a file but not the actual gain values. That would anyway not make much sense (corrupted tags in the files?) but it doesn't hurt to protect for.

**Code assistance**

I'm normally a developer from the embedded world, so my C# experience is limited. I used an LLM to understand the code base, where to plug my changes, but I did the actual implementation and debugging. Tested and confirmed it works on my machine, not just AI slop.

**Issues**

This change just exposes the information but doesn't really fix anything by itself. I'm submitting related PRs to Music Assistant to parse this information:
Jc2k/aiojellyfin#10
music-assistant/server#4784
I haven't found any issues related to ReplayGain or volume normalization in this repo.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -235,6 +235,7 @@
  - [Lampan-git](https://github.com/Lampan-git)
  - [elio42](https://github.com/elio42)
  - [rwebster85](https://github.com/rwebster85)
+ - [Florin-Popescu](https://github.com/Florin-Popescu)
 
 # Emby Contributors
 
--- a/MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
+++ b/MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
@@ -479,7 +479,7 @@ namespace MediaBrowser.Providers.MediaInfo
 
                 if (float.TryParse(trackAlbumGainTag, NumberStyles.Float, CultureInfo.InvariantCulture, out var value) && float.IsFinite(value))
                 {
-                    if (!audio.AlbumEntity.NormalizationGain.HasValue)
+                    if (audio.AlbumEntity is not null && !audio.AlbumEntity.NormalizationGain.HasValue)
                     {
                         audio.AlbumEntity.NormalizationGain = value;
                     }
--- a/MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
+++ b/MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
@@ -468,6 +468,24 @@ namespace MediaBrowser.Providers.MediaInfo
                 }
             }
 
+            TryGetSanitizedAdditionalFields(track, "REPLAYGAIN_ALBUM_GAIN", out var trackAlbumGainTag);
+
+            if (trackAlbumGainTag is not null)
+            {
+                if (trackAlbumGainTag.EndsWith("db", StringComparison.OrdinalIgnoreCase))
+                {
+                    trackAlbumGainTag = trackAlbumGainTag[..^2].Trim();
+                }
+
+                if (float.TryParse(trackAlbumGainTag, NumberStyles.Float, CultureInfo.InvariantCulture, out var value) && float.IsFinite(value))
+                {
+                    if (!audio.AlbumEntity.NormalizationGain.HasValue)
+                    {
+                        audio.AlbumEntity.NormalizationGain = value;
+                    }
+                }
+            }
+
             if (options.ReplaceAllMetadata || !audio.TryGetProviderId(MetadataProvider.MusicBrainzArtist, out _))
             {
                 if ((TryGetSanitizedAdditionalFields(track, "MUSICBRAINZ_ARTISTID", out var musicBrainzArtistTag)
--- a/MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
+++ b/MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
@@ -468,18 +468,18 @@ namespace MediaBrowser.Providers.MediaInfo
                 }
             }
 
-            TryGetSanitizedAdditionalFields(track, "REPLAYGAIN_ALBUM_GAIN", out var trackAlbumGainTag);
-
-            if (trackAlbumGainTag is not null)
+            if (audio.AlbumEntity is not null && !audio.AlbumEntity.NormalizationGain.HasValue)
             {
-                if (trackAlbumGainTag.EndsWith("db", StringComparison.OrdinalIgnoreCase))
-                {
-                    trackAlbumGainTag = trackAlbumGainTag[..^2].Trim();
-                }
+                TryGetSanitizedAdditionalFields(track, "REPLAYGAIN_ALBUM_GAIN", out var trackAlbumGainTag);
 
-                if (float.TryParse(trackAlbumGainTag, NumberStyles.Float, CultureInfo.InvariantCulture, out var value) && float.IsFinite(value))
+                if (trackAlbumGainTag is not null)
                 {
-                    if (audio.AlbumEntity is not null && !audio.AlbumEntity.NormalizationGain.HasValue)
+                    if (trackAlbumGainTag.EndsWith("db", StringComparison.OrdinalIgnoreCase))
+                    {
+                        trackAlbumGainTag = trackAlbumGainTag[..^2].Trim();
+                    }
+
+                    if (float.TryParse(trackAlbumGainTag, NumberStyles.Float, CultureInfo.InvariantCulture, out var value) && float.IsFinite(value))
                     {
                         audio.AlbumEntity.NormalizationGain = value;
                     }


MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
```diff
@@ -468,6 +468,43 @@ private async Task FetchDataFromTags(Audio audio, Model.MediaInfo.MediaInfo medi
...
+            if (trackAlbumGainTag is not null)
+            {
+                if (trackAlbumGainTag.EndsWith("db", StringComparison.OrdinalIgnoreCase))
+                {
+                    trackAlbumGainTag = trackAlbumGainTag[..^2].Trim();
+                }
+
+                if (float.TryParse(trackAlbumGainTag, NumberStyles.Float, CultureInfo.InvariantCulture, out var value) && float.IsFinite(value))
+                {
+                    if (!audio.AlbumEntity.NormalizationGain.HasValue)
```
dfederm: Possible nullref here. `AlbumEntity` is `FindParent<MusicAlbum>()`, which returns null when the track has no `MusicAlbum` parent.

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
```diff
@@ -468,6 +468,43 @@ private async Task FetchDataFromTags(Audio audio, Model.MediaInfo.MediaInfo medi
...
+            {
+                if (trackReferenceLoudnessTag.EndsWith("LUFS", StringComparison.OrdinalIgnoreCase))
+                {
+                    trackReferenceLoudnessTag = trackReferenceLoudnessTag[..^4].Trim();
+                }
+
+                if (float.TryParse(trackReferenceLoudnessTag, NumberStyles.Float, CultureInfo.InvariantCulture, out var value) && float.IsFinite(value))
+                {
+                    audio.ReferenceLoudness = value;
+                    if (!audio.AlbumEntity.ReferenceLoudness.HasValue)
```
dfederm: Same nullref

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
```diff
@@ -468,6 +468,43 @@ private async Task FetchDataFromTags(Audio audio, Model.MediaInfo.MediaInfo medi
...
+                        audio.AlbumEntity.NormalizationGain = value;
+                    }
+                }
+            }
+
+            TryGetSanitizedAdditionalFields(track, "REPLAYGAIN_REFERENCE_LOUDNESS", out var trackReferenceLoudnessTag);
+
+            if (trackReferenceLoudnessTag is not null)
+            {
+                if (trackReferenceLoudnessTag.EndsWith("LUFS", StringComparison.OrdinalIgnoreCase))
```
dfederm: Consider stripping "db" like with the gain. Not sure though how common that is in practice.

Florin-Popescu: I'm not sure either, but it definitely doesn't hurt. Is it ok to do the 2 checks separately like this? I assume you could have either "LUFS" or "dB" but not both.

Bond-009:
REPLAYGAIN_REFERENCE_LOUDNESS should not be used as it deviates from the spec
https://wiki.hydrogenaudio.org/index.php?title=Revised_ReplayGain_specification#De-Facto_extensions

Florin-Popescu:
> REPLAYGAIN_REFERENCE_LOUDNESS should not be used as it deviates from the spec https://wiki.hydrogenaudio.org/index.php?title=Revised_ReplayGain_specification#De-Facto_extensions

I understand your point and while I'm personally anyway using the default -18 value, I also don't see the harm in processing it as long as it's available in the file. A file that has a track gain tag of -6 dB at -18 reference will have the track gain tag at -2 dB if it was scanned at -14 reference. If you parse only the -2dB tag without considering it was relative to a non-standard value, the file will be played too loud. If only some albums in your collection used the non-standard reference for whatever reason, but not the entire library, that defeats the purpose of volume normalization.

What should we do in this situation? Process the reference gain tag when present, translate the track/album gain internally to -18 and only expose the adjusted gain relative to -18 LUFS? I think that might be a good alternative. But ignoring the tag when present in files will cause inconsistent playback volume.

Bond-009:
> But ignoring the tag when present in files will cause inconsistent playback volume.

Correct, but I see that as user error.
It's not our fault if files are tagged incorrectly.

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
Florin-Popescu: Fixed in 605de147c57e855eb6aebd0b696215fa02caadc6

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
Florin-Popescu: No longer relevant if not parsing the reference loudness

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
Florin-Popescu: No longer relevant if not parsing the reference loudness

Florin-Popescu:
> Correct, but I see that as user error. It's not our fault if files are tagged incorrectly.

OK, removed the reference gain. PR greatly simplified due to this.
Should only assign the album gain to the parent album internally and then rely on #14745 to expose the album gain when querying for a track.
Will update music-assistant/server#4784 to assume -18 dB reference level.

CONTRIBUTORS.md
```diff
@@ -296,3 +296,4 @@
  - [tikuf](https://github.com/tikuf/)
  - [Tim Hobbs](https://github.com/timhobbs)
  - [SvenVandenbrande](https://github.com/SvenVandenbrande)
+ - [Florin-Popescu](https://github.com/Florin-Popescu)
```
Bond-009: Wrong list, now you're under Emby instead of Jellyfin

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
```diff
@@ -468,6 +468,24 @@ private async Task FetchDataFromTags(Audio audio, Model.MediaInfo.MediaInfo medi
...
+            if (trackAlbumGainTag is not null)
+            {
+                if (trackAlbumGainTag.EndsWith("db", StringComparison.OrdinalIgnoreCase))
+                {
+                    trackAlbumGainTag = trackAlbumGainTag[..^2].Trim();
+                }
+
+                if (float.TryParse(trackAlbumGainTag, NumberStyles.Float, CultureInfo.InvariantCulture, out var value) && float.IsFinite(value))
+                {
+                    if (audio.AlbumEntity is not null && !audio.AlbumEntity.NormalizationGain.HasValue)
```
Bond-009: Wrap everything in this check, no need to do the work if we won't use it anyway

Bond-009 (changes requested):
Thank you, this is way simpler

CONTRIBUTORS.md
Florin-Popescu: Didn't see there's 2 lists. Fixed in 78b535b

MediaBrowser.Providers/MediaInfo/AudioFileProber.cs
Florin-Popescu: Fixed in f6f0fed

Bond-009 (approved):
Thank you

crobibero (approved):

PR #17320, opened 2026-07-13
Fix: Fetch the correct row matching the most up to date file

jordanmichaelrushing:
**Changes**
The user data read path (GetUserData / GetUserDataBatch) takes the first row for an item without any ordering or key filter, and without an order by the database returns rows in arbitrary order. Writes however, only update rows matching the item's current GetUserDataKeys(). If you're someone like me who edits the underlying files/metadata, the computed keys change and the next save creates rows under the new keys, leaving the old rows orphaned. Reads can then arbitrarily return an orphaned row, so resume positions and played status intermittently revert to stale values. This change resolves the row by the item's current keys, in key order, falling back to remaining rows only when no current-key row exists (so items renamed before this fix keep their resume position).

**Code assistance**
I had Fable write the tests for me, I wrote the code

**Issues**
I couldn't find any. On my end this was resulting in resume position and played status intermittently revert to stale values after an item's metadata is edited.

**Repro**
1) Play an audiobook or track partway so a position is saved
2) Edit metadata: album, album artist, or title
3) Continue listening or mark the item played
4) Re-open the item: position/played state may show the pre-edit values

### DIFF
--- a/Emby.Server.Implementations/Library/UserDataManager.cs
+++ b/Emby.Server.Implementations/Library/UserDataManager.cs
@@ -192,7 +192,8 @@ namespace Emby.Server.Implementations.Library
                 }
                 else
                 {
-                    var userData = item.UserData?.Where(e => e.UserId.Equals(user.Id)).Select(Map).FirstOrDefault();
+                    var userDataRow = ResolveUserDataRow(item, item.UserData?.Where(e => e.UserId.Equals(user.Id)));
+                    var userData = userDataRow is not null ? Map(userDataRow) : null;
                     if (userData is not null)
                     {
                         result[item.Id] = userData;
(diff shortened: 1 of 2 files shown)


Emby.Server.Implementations/Library/UserDataManager.cs
dfederm: This database fallback still selects by `Id.ToString("N")` / `First()`  instead of the new key-ordered `ResolveUserDataRow`, so it bypasses the fix when `item.UserData` isn't preloaded. Note the query above (`allKeys` filter) also drops retired-key rows, so matching the in-memory fallback fully means fetching those too.

Shadowghost (approved):

Emby.Server.Implementations/Library/UserDataManager.cs
jordanmichaelrushing: Nice catches. Working through some changes locally.

jordanmichaelrushing: Updated and also removed the allKeys filter

crobibero (approved):

PR #17333, opened 2026-07-15
Reconnect remote HTTP sources dropped before EOF

### LINKED ISSUE #17332: strm-backed (remote HTTP) VOD library items get no reconnect-on-drop — reconnect flags only apply to Live TV (RequiresLooping)
## Summary

`.strm`-backed library items pointing at a remote HTTP video (e.g. an IPTV VOD URL) get **no reconnect-on-drop handling** if the underlying HTTP connection is cut mid-stream. Playback (or a live transcode) simply dies with a demuxer I/O error instead of retrying, because the ffmpeg reconnect flags (`-reconnect_at_eof`, `-reconnect_streamed`, `-reconnect_delay_max`) are only ever applied when `MediaSource.RequiresLooping` is true — and that flag is only ever set by the Live TV M3U tuner code path, never for a regular library `VideoFile` item.

## Where this happens in the source

`MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs`:
```csharp
if (state.MediaSource.RequiresLooping)
{
    inputModifier += " -stream_loop -1 -reconnect_at_eof 1 -reconnect_streamed 1 -reconnect_delay_max 2";
}
```

`RequiresLooping` is set in exactly one place in the codebase, `src/Jellyfin.LiveTv/TunerHosts/M3UTunerHost.cs`:
```csharp
RequiresLooping = info.EnableStreamLooping,
```

So this reconnect behavior is scoped entirely to Live TV tuner channels. A regular library item whose `MediaSource` is a remote HTTP URL (which is exactly what a `.strm` file resolves to) never gets `RequiresLooping = true`, and therefore never gets the reconnect flags, regardless of protocol or source reliability.

## Real-world reproduction

Environment: Jellyfin 10.11.8, jellyfin-ffmpeg 7.1.3, Intel UHD 770 (Raptor Lake), Docker/LXC on Linux, kernel 6.14.11.

We have a `.strm` file (an IPTV VOD provider URL) that drops its HTTP connection partway through the file. Manually reproduced outside of Jellyfin with the exact command Jellyfin itself generates for this MediaSource (minus the `-ss`/output-specific parts):

```
ffmpeg -analyzeduration 200M -probesize 1G -f matroska -i "http://<provider>/series/.../<id>.mkv" -c copy out.mkv
```

Result, consistently reproducible from byte 0:
```
[http @ ...] Stream ends prematurely at 840675404, should be 6588292445
[matroska,webm @ ...] Read error
[http @ ...] Stream ends prematurely at 840675404, should be 6588292445
    Last message repeated 1 times
[in#0/matroska,webm @ ...] Error during demuxing: Input/output error
```

The server reports a Content-Length of ~6.6GB but the connection is cut after only ~840MB. In real playback through Jellyfin, this manifests as the video freezing/erroring at a reproducible point (independent of client-side seeking) and the transcode/direct-stream session dying — no retry is attempted.

## Why this matters beyond IPTV

Any `.strm`-based library (IPTV-to-strm indexers, network-mounted media exposed as strm, cloud-storage-backed strm setups, etc.) is exposed to this same gap whenever the remote host has any transient connectivity hiccup, not just a permanently broken file — a single dropped TCP connection during totally healthy playback would produce the same unrecoverable failure, since nothing distinguishes "permanently broken source" from "transient network blip" here. Both currently get zero retry.

## Suggested fix

Either:
1. Also set (or independently check) a condition for remote-HTTP `VideoFile`/VOD sources — not just Live TV — so `-reconnect_at_eof 1 -reconnect_streamed 1 -reconnect_delay_max 2` apply there too (without `-stream_loop -1`, since VOD content has a known finite duration and shouldn't loop, just reconnect once and resume from the dropped offset), or
2. Expose this as an explicit, narrower "reconnect on remote source disconnect" behavior for any `MediaSource` where `Protocol == MediaProtocol.Http`, independent of `RequiresLooping`/Live TV at all.

Happy to test a patch against our environment if useful — we have a specific, reliably-reproducing case in hand.

dfederm:
**Changes**

The ffmpeg reconnect flags were only emitted for `MediaSource.RequiresLooping`, which only the Live TV M3U tuner sets. So a remote HTTP `.strm` VOD item whose connection dropped mid-stream got no reconnect, and its server-side transcode/remux died with a demuxer I/O error. This adds `-reconnect 1 -reconnect_delay_max 2` for HTTP inputs (`state.InputProtocol == MediaProtocol.Http`), as an `else if` after the looping branch so Live TV is unchanged.

`-reconnect` is the flag that handles a disconnect before EOF; `-reconnect_at_eof`/`-reconnect_streamed` are omitted since they target live streams and would mishandle finite VOD. This only helps server-side ffmpeg paths (transcode, remux, HLS), and resume depends on the source honoring HTTP Range requests.

**Code assistance**

Used an LLM to trace the reconnect decision through `libavformat/http.c` (confirming a premature close returns `AVERROR(EIO)` and that `-reconnect` is the flag that path checks, whereas `-reconnect_at_eof`/`-reconnect_streamed` do not apply to finite seekable VOD) and to draft the unit tests. The change and flag choice were reviewed against the ffmpeg source rather than taken as generated.

**Issues**

Fixes #17332

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -7415,6 +7415,12 @@ namespace MediaBrowser.Controller.MediaEncoding
             {
                 inputModifier += " -stream_loop -1 -reconnect_at_eof 1 -reconnect_streamed 1 -reconnect_delay_max 2";
             }
+            else if (state.InputProtocol == MediaProtocol.Http)
+            {
+                // Reconnect a remote HTTP source (e.g. a .strm-backed VOD item) if it drops before EOF.
+                // No -stream_loop/-reconnect_at_eof/-reconnect_streamed: VOD is finite and seekable.
+                inputModifier += " -reconnect 1 -reconnect_delay_max 2";
+            }
 
             return inputModifier;
         }
--- a/tests/Jellyfin.MediaEncoding.Tests/EncodingHelperTests.cs
+++ b/tests/Jellyfin.MediaEncoding.Tests/EncodingHelperTests.cs
@@ -1,0 +1,82 @@
+using System;
+using MediaBrowser.Common.Configuration;
+using MediaBrowser.Controller.IO;
+using MediaBrowser.Controller.MediaEncoding;
+using MediaBrowser.Model.Configuration;
+using MediaBrowser.Model.Dto;
+using MediaBrowser.Model.MediaInfo;
+using Microsoft.Extensions.Configuration;
+using Moq;
+using Xunit;
+
+namespace Jellyfin.MediaEncoding.Tests;
+
+public class EncodingHelperTests
+{
+    private static EncodingHelper CreateEncodingHelper()
+    {
+        return new EncodingHelper(
+            Mock.Of<IApplicationPaths>(),
+            Mock.Of<IMediaEncoder>(),
+            Mock.Of<ISubtitleEncoder>(),
+            Mock.Of<IConfiguration>(),
+            Mock.Of<MediaBrowser.Common.Configuration.IConfigurationManager>(),
+            Mock.Of<IPathManager>());
+    }
+
+    private static EncodingJobInfo CreateJobInfo(MediaProtocol protocol, bool requiresLooping)
+    {
+        return new EncodingJobInfo(TranscodingJobType.Progressive)
+        {
+            BaseRequest = new BaseEncodingJobOptions(),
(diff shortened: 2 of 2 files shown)


dfederm:
@therobmilne Feel free to test the fix. I also tested it locally with a synthetic repro, but you metnioned you had a reliable repro which presumably is more of a real-world test than mine.

dfederm:
Nudge @nyanmisaka

PR #17349, opened 2026-07-15
Add per-member SyncPlay buffering timeout so one client cannot stall the group forever

tehreet:
**Changes**

`Group.IsBuffering()` blocks on any buffering member with no time bound, so a client whose
playback/transcode never starts is indistinguishable from ordinary network drift and can hang
a SyncPlay group indefinitely — everyone else stays frozen until someone manually removes the
stuck client.

Commit 1 tracks when each member started its current buffering period and applies a generous
timeout (60 s) while a member is still preparing playback for the first time vs. a tight
timeout (10 s) once it has already caught up and is merely re-buffering. A member that exceeds
its timeout is dropped from the wait (with a warning naming the session) so the rest of the
group proceeds. Stalled usernames are surfaced on `GroupInfoDto` as `BufferingParticipants`
(additive/optional field) so clients can show who is holding up playback.

Commit 2 fixes a flaw in commit 1's "has reported since buffering started" heuristic: because
`SetAllBuffering(true)` (seek, next item, unpause-from-idle) marks everyone buffering up
front, the very first Buffer/Ready report after a seek flipped a still-preparing session onto
the tight 10 s clock — exactly the slow-start case the timeout was meant to protect. Replaced
with `GroupMember.HasCaughtUpSinceReset`, set only on a genuine catch-up and cleared only by a
group-wide reset. (Kept as two commits to preserve the reasoning; happy to squash on merge.)

Tested by building the touched projects and by running the change in production on a private
Jellyfin server: previously a client with a slow-starting transcode regularly hung group
playback; with this change the group resumes and the stuck client is identified in the UI.

**Code assistance**

Developed with LLM assistance (Claude): root-cause analysis and initial patch generation.
All code was human-reviewed, built, and has been running in production on a private server.

**Issues**

None.

### DIFF
--- a/Emby.Server.Implementations/SyncPlay/Group.cs
+++ b/Emby.Server.Implementations/SyncPlay/Group.cs
@@ -27,6 +27,18 @@ namespace Emby.Server.Implementations.SyncPlay
     public class Group : IGroupStateContext
     {
         /// <summary>
+        /// The maximum time, in milliseconds, a session that is still preparing playback for the first
+        /// time in the current buffering cycle (e.g. waiting on a transcode) can keep the group waiting.
+        /// </summary>
+        private const long PreparingTimeoutMs = 60000;
+
+        /// <summary>
+        /// The maximum time, in milliseconds, a session that has already caught up once in the current
+        /// cycle (and is therefore assumed to be re-buffering from network lag) can keep the group waiting.
+        /// </summary>
+        private const long NetworkLagTimeoutMs = 10000;
+
+        /// <summary>
         /// The logger.
         /// </summary>
         private readonly ILogger<Group> _logger;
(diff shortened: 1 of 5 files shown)


Shadowghost (commented):
My main concern is that the timeout is passive, and the drop doesn't itself resume the group.

`IsBuffering()` is now mutating: when a session exceeds its timeout it clears `IsBuffering/BufferingSince` and continues. But clearing the flag is not the same as resuming playback. The `Waiting → Playing/Paused` transition (and the Unpause/resume commands) live only in the request handlers that check `IsBuffering()` leading to the timeout only being evaluated when a Buffer/Ready/IgnoreWait/SessionLeave request arrives or when the client calls `GetInfo()`. In the latter case, we are also not handling the state transition.

Also, if a member ever has `IsBuffering == true` with `BufferingSince == null` (no sure if this can happen), `IsBuffering()` skips the timeout and blocks indefinitely.

Some tests would also be nice.

PR #17350, opened 2026-07-15
Track a stable SyncPlay host and reassign it when the host leaves

tehreet:
**Changes**

SyncPlay groups have no stable notion of "who created / hosts this group". Anything a client
bases on `Participants[0]` is built on `Dictionary<TKey,TValue>` enumeration order, which is
unspecified and shifts whenever an entry is removed (`_participants.Remove` runs on every
SessionLeave) — so "who is host" can silently reassign to an arbitrary member after ordinary
group churn.

Commit 1 tracks `HostUsername` on `Group`, set in `CreateGroup` from the creating session's
username, and exposes it on `GroupInfoDto` (additive, following the same optional-parameter
pattern as other DTO additions) so clients have a stable identity to key off.

Commit 2 handles the host leaving: on SessionLeave, if no remaining participant carries the
host's username (checked by username, so a host with multiple sessions keeps the role until
the last one leaves) and the group is not empty, the longest-standing remaining member is
promoted (`GroupMember` now records `JoinedAt`; ties broken by session id — deterministic and
independent of dictionary order). Reassignment happens before the leave updates go out, so
every subsequent `GetInfo()` carries the new value.

Tested by building the touched projects and by running the change in production on a private
Jellyfin server with regular multi-client SyncPlay sessions, including host-leave scenarios.

**Code assistance**

Developed with LLM assistance (Claude): root-cause analysis and initial patch generation.
All code was human-reviewed, built, and has been running in production on a private server.

**Issues**

None.

### DIFF
--- a/Emby.Server.Implementations/SyncPlay/Group.cs
+++ b/Emby.Server.Implementations/SyncPlay/Group.cs
@@ -115,6 +115,16 @@ namespace Emby.Server.Implementations.SyncPlay
         public string GroupName { get; private set; }
 
         /// <summary>
+        /// Gets the username of the session that created this group. Set once at group
+        /// creation time and never reassigned afterward, regardless of participants joining
+        /// or leaving. Unlike deriving "who is host" from participant/dictionary ordering
+        /// (which is unspecified and shifts once any entry is removed), this is a stable
+        /// identity for the lifetime of the group.
+        /// </summary>
+        /// <value>The host's username.</value>
+        public string HostUsername { get; private set; }
+
+        /// <summary>
         /// Gets the group identifier.
         /// </summary>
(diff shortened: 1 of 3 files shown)


Shadowghost (commented):
Some minor comments form my side:
* Some tests would be nice.
 * Reassignment updates the server field, but the leave broadcast (`SyncPlayGroupLeftUpdate` to the leaver, `SyncPlayUserLeftUpdate` to others) carries no `GroupInfoDto`. So already-connected clients keep their cached host until the next `GetInfo()` (a new joiner, a group-list refresh). I'm not sure if this is the right approach
 * Since this modifies the websocket layer, this change will need to wait for 13.x

PR #11517, opened 2024-05-09
Implement multi-part audiobook support

Samson50:
Modify the server to handle audiobooks which are split into multiple audio files. This assumes all files for a single audiobook are in their own directory and each directory only contains files for a single audiobook. 

By itself, the changes in this merge request allow the server to handle resolving multi-part audiobooks into a single parent AudioBook entity and serve to the client under the normal books view. Playback can be started from the beginning of the audiobook and continue through all chapters but cannot be "resumed" once the play session is stopped.

When combined with the changes in the web pull request, playback can be resumed from the most recent stopping point from the books list view or from the "Continue Listening" entry.

**Changes**
- Changed AudioBook entity to a subclass of Folder - serving as a parent for the chapters/files of an audiobook
- Added entity for AudioBookFile representing the individual playable media files of an audiobook
- Added conditionals for AudioBook/AudioBookFiles to controllers/managers to enable expected playback behavior
 
**Issues**
Fixes #10668

Issue #10668, opened 2023-12-03
[Issue]: Multi-Part Audiobook being treated as folder with indivdual files instead of one book


verybadsoldier:
### Please describe your bug

When having an audiobook in one folder consisting of multiple MP3 files, then JF does not treat it as one audiobook but treats it as a folder showing multiples files.

I have this structure:
```
D:\JELLYTEST\BOOKS
└───audiobooks
    └───John Doe
        └───MyAudiobook
                01 Chapter.mp3
                02 Chapter.mp3
                03 Chapter.mp3
                04 Chapter.mp3
                05 Chapter.mp3
```

Then in JF, the "MyAudiobook" is shown as a folder which you cannot play directly:
![image](https://github.com/jellyfin/jellyfin/assets/364469/094a110a-8dc2-49bd-ad03-dca37024c5fe)

Then, inside this folder I get individual files that I can play:
![image](https://github.com/jellyfin/jellyfin/assets/364469/6d36e590-afce-4565-aefa-c03ef503ce78)

Also, on the landing page, as "New Media", every file as shown as separate audiobook:
![image](https://github.com/jellyfin/jellyfin/assets/364469/cafb4960-8790-43c8-8844-6707b0fc3fd2)

My expectation of this whole thing would be, that "MyAudiobook" is detected as one single audiobook by JF and you can directly play it. Also, on the landing page, it should only show up as one audiobook.

But maybe my expectation is just wrong, thanks!

I usually use JF on Linux as docker, but for this demonstration purpose, I have set up an instance on my Windows machine to make it easier to play around with it.

### Jellyfin Version

Other

### if other:

10.8.13

### Environment

```markdown
- OS: Windows 10
- Browser: Chrome
- Plugins: Bookshelf
- Storage: Local
```

### Jellyfin logs

```shell
[15:16:18] [DBG] [65] Jellyfin.Api.Auth.CustomAuthenticationHandler: AuthenticationScheme: CustomAuthentication was successfully authenticated.
[15:16:18] [DBG] [65] Jellyfin.Api.Auth.CustomAuthenticationHandler: AuthenticationScheme: CustomAuthentication was successfully authenticated.
[15:16:18] [DBG] [78] MediaBrowser.Providers.Folders.CollectionFolderMetadataService: CollectionFolderImageProvider reports change to C:\Users\vbs\AppData\Local\jellyfin\root\default\Bücher
[...]

PR #17362, opened 2026-07-17
Audiobooks: Add multi-file chapter support

jordanmichaelrushing:
**Changes**
- Add support for taking audiobooks split into multiple files in the same folder and combining into one Item, by chapter
- Add support for per file artwork (every file can have it's own artwork, [Graphic Audio](https://www.graphicaudio.net/the-stormlight-archive-1-download-series-set.html) is a prime example)
- Support multifile progress for the api (admin dashboard usecase)

**Code assistance**
- Multiple code reviews and suggestions to tighten up my code
- Had Opus help me with tests, only code it wrote

**Issues**
- https://github.com/jellyfin/jellyfin/issues/10668

**Feature Requests**
- https://features.jellyfin.org/posts/794/audiobook-metadata-support

**Frontend**
I'm at my 3 PR limit over there so I'll update this section when I have access.

**Notes**
- A [previous attempt](https://github.com/jellyfin/jellyfin/pull/11517) had this at 40 files changed
- Split from my much bigger more [general purpose Audiobook PR](https://github.com/jellyfin/jellyfin/pull/17199)

### DIFF
--- a/Emby.Server.Implementations/Dto/DtoService.cs
+++ b/Emby.Server.Implementations/Dto/DtoService.cs
@@ -1252,6 +1252,11 @@ namespace Emby.Server.Implementations.Dto
                     .ToArray();
             }
 
+            if (item is AudioBook audioBookItem && audioBookItem.AdditionalParts.Length != 0)
+            {
+                dto.PartCount = audioBookItem.AdditionalParts.Length + 1;
+            }
+
             // Add video info
             if (item is Video video)
             {
--- a/Emby.Server.Implementations/Library/MediaSourceManager.cs
+++ b/Emby.Server.Implementations/Library/MediaSourceManager.cs
@@ -393,7 +393,8 @@ namespace Emby.Server.Implementations.Library
                 sources = sources
                     .Where(source => !Guid.TryParse(source.Id, out var sourceId)
                         || sourceId.Equals(item.Id)
-                        || _libraryManager.GetItemById<BaseItem>(sourceId, user) is not null)
+                        || _libraryManager.GetItemById<BaseItem>(sourceId, user) is not null
+                        || item is AudioBook)
                     .ToArray();
 
                 foreach (var source in sources)
--- a/Emby.Server.Implementations/Library/Resolvers/Audio/AudioResolver.cs
+++ b/Emby.Server.Implementations/Library/Resolvers/Audio/AudioResolver.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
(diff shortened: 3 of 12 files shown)


seba76:
Hi @jordanmichaelrushing, great work you are doing. I do have one question in regard to your audiobooks changes. What happens if "audiobook" is actually "videobook"?

To give you more context, I have several video tutorials that are made of several mp4 files with chapters. Depending on tutorial it is sometimes nice to look at the video and not only listen to audio. For now I've solved that with custom hacked audiobookshelf where I changed audio tag to video plus small ui changes, but if jellyfin can support something like this out of the box it will be excellent replacement for autiobookshelf.

jordanmichaelrushing:
@seba76 it sounds more similar to a tv show with separate files for each "episode". But it sounds like you want it as a single video? I'm not a Jellyfin member, so honestly my opinion here doesn't really amount to anything, I don't speak on behalf of Jellyfin. But given their answers to questions similar to this I imagine their response will be "This isn't a feature we would want to support". They don't want Jellyfin to become a Swiss Army Knife / jack of all, master of none. They'll be able to give an official word on that, I might suggest the Forum or Matrix and ask them there.

seba76:
@jordanmichaelrushing, thank you for answering but I was more asking how would your branch handle "audiobook" files which are mp4 file rather then m4b. To be more precise would I see video or it would be audio playback only? 

For my use case I have following:
- Vol01_Topic1.mp4 which contains chapters 01 to 05
- Vol02_Topic2.mp4 which contains chapters 06 to 14
...
- Vol10_Topic10.mp4 which contains chapters 95 to 97

In audiobookshelf I see that as one audiobook with chapters 01 to 97 which is extremely nice for navigation, I don't have to remember on which file I'm now abs does that for me and in addition it is extremely easy to jump from chapter to chapter. I've asked abs if they would support video files but I got pretty much the same answer you thought jellyfin crew would give.

jordanmichaelrushing:
Ah. In that case it wouldn't handle them. Audiobook is a subset of Audio which has a list of audio extensions, mp4 is not one of them. Even if you forced the mp4 ext, it still wouldn't work as the type of file determines the player, in which case Jellyfin would attempt to play the audio of the files (no video). I'm not even sure it would be able to play just the audio in that case or if it would error out. Ffmpeg would probably be fine and you'd just have audio, but probably isn't the flow I assume you want and YMMV

seba76:
Thanks for explaining, I guess I need to stick to my patched abs.

pomianowski:
~~ Friendly reminder about the PR

Issue #12254, opened 2024-07-11
[Issue]: Jellyfin crashing upon startup


Gilgamesh0028:
### Please describe your bug

Crashes upon launch. 

### Reproduction Steps

rebooted machine (previously had it working fine pre-reboot), and where upon logon jellyfin is set to auto-start. 
When I received the popup saying it failed to start, i checked the log; saw the xml error, and decided to make an issue.

### Jellyfin Version

10.9.2

### Environment

```markdown
- OS: Windows 10, 22H2
- Linux Kernel: N/A
- Virtualization: N/A
- Clients: Local (via browser)
- Browser: Firefox
- FFmpeg Version: latest bundled (I don't recall, and lack the ability to check via jellyfin cli / GUI
- Playback Method: N/A
- Hardware Acceleration: enabled
- GPU Model: NVIDIA RTX 2080 VENTUS OC
- Plugins: anilist metadata fetcher
- Reverse Proxy: N/A
- Base URL: 127.0.0.1:8096
- Networking: 2.5Gb Ethernet
- Storage:1TB SSD (jellyfin server dir) 4TB HDD (media dir)
```

### Jellyfin logs

```shell
[2024-07-10 21:45:47.211 -06:00] [INF] [1] Main: Jellyfin version: "10.9.2"
[2024-07-10 21:45:47.227 -06:00] [INF] [1] Main: Environment Variables: ["[JELLYFIN_LOG_DIR, C:\Users\ADMIN\Documents\Jellyfin\Server\log]"]
[2024-07-10 21:45:47.229 -06:00] [INF] [1] Main: Arguments: ["C:\Program Files\Jellyfin\Server\jellyfin.dll", "--datadir", "C:\Users\ADMIN\Documents\Jellyfin\Server"]
[2024-07-10 21:45:47.231 -06:00] [INF] [1] Main: Operating system: "Microsoft Windows 10.0.19045"
[2024-07-10 21:45:47.232 -06:00] [INF] [1] Main: Architecture: X64
[2024-07-10 21:45:47.234 -06:00] [INF] [1] Main: 64-Bit Process: True
[2024-07-10 21:45:47.235 -06:00] [INF] [1] Main: User Interactive: True
[2024-07-10 21:45:47.236 -06:00] [INF] [1] Main: Processor count: 16
[2024-07-10 21:45:47.237 -06:00] [INF] [1] Main: Program data path: "C:\Users\ADMIN\Documents\Jellyfin\Server"
[...]

PR #17369, opened 2026-07-17
Prevent unauthenticated re-run of the startup wizard on misconfiguration

Shadowghost:
The `/Startup/*` endpoints are gated by the `FirstTimeSetupHandler`, which grants unauthenticated access whenever `IsStartupWizardCompleted = false`. If that flag is ever cleared on an already-provisioned server (corrupted or partially restored configuration, a manual edit, a botched migration) the entire admin-provisioning surface reopens, letting an unauthenticated caller reset the administrator password.

**Changes**
- At startup, if the wizard is marked incomplete but a configured administrator (admin with a password) already exists, restore the `IsStartupWizardCompleted` flag so the setup endpoints cannot be served unauthenticated on an established server.
- `UpdateStartupUser` now refuses (403) to overwrite the password of an account that already has one, independent of the flag. This is safe for genuine first-time setup, where the freshly initialized user has no password yet.

**Code assistance**
None

**Issues**

### DIFF
--- a/Emby.Server.Implementations/ApplicationHost.cs
+++ b/Emby.Server.Implementations/ApplicationHost.cs
@@ -39,6 +39,8 @@ using Emby.Server.Implementations.SyncPlay;
 using Emby.Server.Implementations.TV;
 using Emby.Server.Implementations.Updates;
 using Jellyfin.Api.Helpers;
+using Jellyfin.Data;
+using Jellyfin.Database.Implementations.Enums;
 using Jellyfin.Drawing;
 using Jellyfin.MediaEncoding.Hls.Playlist;
 using Jellyfin.Networking.Manager;
@@ -417,6 +419,8 @@ namespace Emby.Server.Implementations
         {
             Logger.LogInformation("Running startup tasks");
 
+            EnsureStartupWizardIntegrity();
+
             Resolve<ITaskManager>().AddTasks(GetExports<IScheduledTask>(false));
 
             ConfigurationManager.ConfigurationUpdated += OnConfigurationUpdated;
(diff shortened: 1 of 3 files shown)


Emby.Server.Implementations/ApplicationHost.cs
```diff
@@ -436,6 +440,24 @@ public Task RunStartupTasksAsync()
             return Task.CompletedTask;
         }
 
+        private void EnsureStartupWizardIntegrity()
```
cvium, line 443: This logic (minus config saving) could be put into the FirstTimeStartupHandler (name might be incorrect)?

Shadowghost: Since the rest of the wizard depends on the flag, if we evaluate it in the handler, any step after the user creation will fail

jordanmichaelrushing (approved):

crobibero (approved):

theguymadmax (approved):
Approved, accidently disabled the auto merge.

Bond-009:
Just making sure we actually want this, we've told people before to use this behavior to fix their installs. Not sure what they could do after this change.
https://github.com/jellyfin/jellyfin/issues/12254#issuecomment-2224013816

theguymadmax:
That was to recreate a missing/corrupt migration xml file.  And now there is no migration.xml file since we moved it into the db.

Bond-009:
> That was to recreate a missing/corrupt migration xml file. And now there is no migration.xml file since we moved it into the db.

Sure, it was just one example, but I think it's a pretty common bandaid fix to peoples issues

PR #17374, opened 2026-07-18
Fix: Prevent resume position from being wiped by 0 stop position

jordanmichaelrushing:
**Changes**
There's a bug on the Android app where a harsh bluetooth disconnect sends a stop position of 0ms. If the client doesn't continue playing afterwards, it will store their current position at 0. Regardless of whether the Android issue gets fixed, the server should be able to determine in the vast majority of cases whether the 0 position is intended or not.

**Code assistance**
Opus wrote the tests for me; I wrote the non test code

**Issues**
https://github.com/jellyfin/jellyfin-android/issues/2084

**Notes**
I previously had [this](https://github.com/jellyfin/jellyfin/pull/17226) fix, but that was when I thought the issue was for near zero, not actually zero.

### DIFF
--- a/Emby.Server.Implementations/Library/UserDataManager.cs
+++ b/Emby.Server.Implementations/Library/UserDataManager.cs
@@ -434,9 +434,20 @@ namespace Emby.Server.Implementations.Library
             };
         }
 
-        /// <inheritdoc />
-        public bool UpdatePlayState(BaseItem item, UserItemData data, long? reportedPositionTicks)
+        private long GetCommittedPositionTicks(Guid itemId, Guid userId)
         {
+            using var dbContext = _repository.CreateDbContext();
+            return dbContext.UserData
+                .Where(e => e.ItemId == itemId && e.UserId == userId)
+                .Select(e => e.PlaybackPositionTicks)
+                .OrderByDescending(t => t)
+                .FirstOrDefault();
+        }
+
+        /// <inheritdoc />
+        public bool UpdatePlayState(BaseItem item, UserItemData data, long? reportedPositionTicks, User? user = null, bool wasStopped = false)
+        {
+            var previousPositionTicks = data.PlaybackPositionTicks;
             var playedToCompletion = false;
 
             var runtimeTicks = item.GetRunTimeTicksForPlayState();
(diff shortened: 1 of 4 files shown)


Shadowghost (approved):

Emby.Server.Implementations/Library/UserDataManager.cs
```diff
@@ -495,6 +506,15 @@ public bool UpdatePlayState(BaseItem item, UserItemData data, long? reportedPosi
                 positionTicks = 0;
             }
 
+            if (wasStopped && reportedPositionTicks == 0 && positionTicks == 0 && !playedToCompletion && user is not null && previousPositionTicks > 0 && item.SupportsPositionTicksResume)
```
cvium: I find this very hard to read. Might be good to either do named (composite) variables or write a comment.

Emby.Server.Implementations/Library/UserDataManager.cs
```diff
@@ -434,9 +434,20 @@ private UserItemDataDto GetUserItemDataDto(UserItemData data, Guid itemId)
...
+            return dbContext.UserData
+                .Where(e => e.ItemId == itemId && e.UserId == userId)
+                .Select(e => e.PlaybackPositionTicks)
+                .OrderByDescending(t => t)
+                .FirstOrDefault();
+        }
+
         /// <inheritdoc />
-        public bool UpdatePlayState(BaseItem item, UserItemData data, long? reportedPositionTicks)
+        public bool UpdatePlayState(BaseItem item, UserItemData data, long? reportedPositionTicks, User? user = null, bool wasStopped = false)
```
cvium: I don't think User can be null. A UserItemData implies we have a user. This method also only uses the Id, so you don't really need the whole object.

Emby.Server.Implementations/Library/UserDataManager.cs
jordanmichaelrushing: Sure, I'll weigh both options. I'm not the biggest fan off comments as I think the code should speak for itself, so most likely I'll go this way

Emby.Server.Implementations/Library/UserDataManager.cs
jordanmichaelrushing: Good callouts. I'll optimize this a bit and tackle the Sonar complexity at the same time

jordanmichaelrushing:
@cvium - I updated and split out in to function by type to be more readable and straight-forward. Also switched to UserId, good callouts.

crobibero:
How does this interact with the Rewatching feature? (User > Display > Enable Rewatching in Next Up)

jordanmichaelrushing:
> User > Display > Enable Rewatching in Next Up

@crobibero  Rewatching is unaffected. The restore only triggers on `wasStopped && reportedPositionTicks == 0 && !playedToCompletion`, and it only restores `PlaybackPositionTicks` — it never modifies Played or LastPlayedDate, which are what drive the rewatching Next Up selection (`DetermineNextEpisodeForRewatching` orders by LastPlayedDate among played episodes). Completing a rewatch still hits the `playedToCompletion` path and resets the position exactly as before, so Next Up advances normally. The only change a rewatcher sees is that a 0-position stop report no longer wipes their mid-episode resume point. The same protection this PR adds for first watches. A partial rewatch (Played=true with a nonzero position) is already a reachable state on master via progress updates, so no new state shape is introduced

IDisposable (approved):

crobibero:
@jordanmichaelrushing you have merge conflicts, please fix so we can re-review

PR #17377, opened 2026-07-18
Exempt people from the allowed tags visibility check

### LINKED ISSUE #14926: [bug] No Cast & Crew Section When Allow Items with Tags Set
### Description of the bug

Verified issue on Jellyfin 10.10.7, using the web browser client.

If a user has a "Parental Control" set to only allow media (say movies) to be listed if they have a particular tag set in the media's metadata, then when the user looks at the movie listing, the Cast & Crew section doesn't appear at all. Tested with 4 users on the same Jellyfin server, 2 of which had no parental control tags and could see the Cast & Crew, and 2 of which had the parental control tags and can't see the Cast & Crew.

Also noticed searches for any actor results all come up blank for those folks who can't see the Cast & Crew section. Am thinking that possibly the movie/TV show/media metadata tag rules are being applied to cast and crew as well rather than just the media?

An example of why/how someone uses parental control tags: We have a multilingual household, some members of which can't speak the same languages. We tag movies and TV shows with tags like "language_eng" or "subtitle_jpn" and then set the parental control filters so users only see movies and TV shows in the languages that they use. (Very handy, especially with the LanguageTag plugin which automatically adds language tags to movies and TV shows for you.)

### Reproduction steps

1. Sign in to the Jellyfin web client as an admin.
2. Randomly select a movie with a cast and crew
3. Click/tap the 3 dots menu for the movie and select "Edit metadata".
4. Scroll down to the Tags section and add a new tag "test".
5. Visit the Dashboard and select the Users menu.
6. Pick a user and click/tap the 3 dot menu for that user and select Parental control.
7. Scroll down to the "Allow Items with tags" section.
8. This section has some helper text: "Only show media with at least one of the specified tags."
9. Add "test" as a tag here then scroll to the bottom and hit the Save button.
10. Log out then log in as the user where you edited the parental controls.
11. Select the same movie. You will not be able to see the Cast & Crew section.
12. You also can't find any cast members when using the top bar global search button.

You will not be able to see the Cast & Crew section when a movie has a cast and crew.
You will not be able to globally search for any actor/cast/crewmember by name.

If the movie can be displayed as the movie tag matches the parental control list, the cast and crew section should be displayed, just as if there is no parental control.

Users should be able to search for cast and crew on any movies they are allowed to see via parental controls.

### Jellyfin Server version

10.10.0+

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.10.7 (f4b8aa0ed)

### Environment

```markdown
- OS: Debian 12
- Linux Kernel: 6.1.0-39-amd64
- Virtualization: none
```

### Jellyfin logs

```shell
Nothing relevant
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

mbastian77:
**Changes**

When a user has "Allow items with tags" configured, the tag filter was also applied to `Person` items. People never carry the tags of the media they appear in, so for those users the Cast & Crew section disappeared entirely and searching for any person returned no results.

This exempts people from the allowed tags check in the three places it was applied:

- `Person` now overrides `IsVisible` and skips the allowed tags part of the check. Blocked tags and parental ratings still apply, so nothing that parental controls are meant to protect is weakened. This fixes the Cast & Crew section (filtered in `DtoService.AttachPeople`) and `/Persons` (filtered in `LibraryManager.GetPeopleItems`).
- The allowed-tags predicate in `BaseItemRepository.TranslateQuery` and `BaseItemRepository.ApplyAccessFiltering` no longer filters out `Person` rows (analogous to the existing playlist owner exemption), which fixes person search.

**Issues**

Fixes #14926

### DIFF
--- a/Jellyfin.Server.Implementations/Item/BaseItemRepository.QueryBuilding.cs
+++ b/Jellyfin.Server.Implementations/Item/BaseItemRepository.QueryBuilding.cs
@@ -447,6 +447,7 @@ public sealed partial class BaseItemRepository
         if (filter.IncludeInheritedTags.Length > 0)
         {
             var includeTags = filter.IncludeInheritedTags.Select(e => e.GetCleanValue()).ToArray();
+            var personTypeName = _itemTypeLookup.BaseItemKindNames[BaseItemKind.Person];
             var allowedTagItemIds = context.ItemValuesMap
                 .Where(f => f.ItemValue.Type == ItemValueType.Tags && includeTags.Contains(f.ItemValue.CleanValue))
                 .Select(f => f.ItemId);
@@ -455,7 +456,10 @@ public sealed partial class BaseItemRepository
                 allowedTagItemIds.Contains(e.Id)
                 || (e.SeriesId.HasValue && allowedTagItemIds.Contains(e.SeriesId.Value))
                 || e.Parents!.Any(p => allowedTagItemIds.Contains(p.ParentItemId))
-                || (e.TopParentId.HasValue && allowedTagItemIds.Contains(e.TopParentId.Value)));
+                || (e.TopParentId.HasValue && allowedTagItemIds.Contains(e.TopParentId.Value))
+
+                // People don't carry the tags of the media they appear in and would never match
+                || e.Type == personTypeName);
         }
 
         // Exclude alternate versions (have PrimaryVersionId set) and owned non-extra items.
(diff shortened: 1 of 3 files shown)


Shadowghost (approved):
I think this is correct, but maybe @theguymadmax can do a quick validation

crobibero (approved):

theguymadmax (approved):
I did test this out and it does resolve the issue. The only issue that comes up, is that if people are specially blocked for other users, that check won't be honored anymore.  

A user might have done this as a workaround, since all people are visible regardless of parental rating checks on media when searching. 

IMO that shouldn't block this.

Issue #15075, opened 2025-10-21
Replace All Metadata does not work


alexkahler:
### Description of the bug

Using "Refresh metadata > "Replace All Metadata" for a Specials season and its episodes does not update or replace the metadata. Values are stuck.
Only by moving the season folder with the affected episodes out of the library, scanning for updated files, moving the season folder back into the library (thereby re-adding them to Jellyfin), and once again "scanning for updated files" will force Jellyfin to populate metadata according to your chosen provider correctly.
Or, by manually "resetting" the metadata via the UI, input a dummy obligatory "Title", and then "Replace All Metadata" will populate the metadata again.

Note: There are no NFO files in the season folder with the episode. There _is_ an NFO file for the show in the parent folder; "tvshow.nfo"

### Reproduction steps

It is unknown exactly which steps are needed to reproduce the behavior, but a couple of suspected points are:
1. Have NFO files with the episode, and let that populate
2. Delete the NFO files and the values from the NFO files are "stuck"
3. Have "TMDb" as the primary metadata provider and let metadata populate
4. Change metadata provider to another provider, e.g. "TheTVDb" and replace all metadata again

### What is the current _bug_ behavior?

Metadata, with e.g., episode naming and ordering, is not updated according to the expected metadata provider. This is obvious when episode ordering for a show is different between the providers.

### What is the expected _correct_ behavior?

For Jellyfin always use the metadata provider according to the library settings and to replace all metadata when forcing a refresh.

### Jellyfin Server version

10.10.0+

### Specify the build version

10.10.7

### Environment

```markdown
- OS:
- Linux Kernel:
- Virtualization: Docker
- Clients: Browser
- Browser: Brave 1.18.118
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins: TMDb, TheTVDb
- Reverse Proxy:
- Base URL: none
- Networking:
- Jellyfin Data Storage: RAID1 SSD
- Media Storage: RAID5 HDD
- External Integrations:
```

### Jellyfin logs

```shell
Running a "Replace all metadata" on a single episode:
[...]

PR #4242, opened 2020-10-01
Increase library scan and metadata refresh speed

Spacetech:
Allows the the library scan to perform concurrent subfolder scans and concurrent metadata refreshes. It requires #4223 to be merged for it to work correctly for TV show libraries.

**Scanning**
Some brief background info for how the scanner works:

The scan is initiated on the root folder. This root folder contains a child for each library. For each child it will:
   1. Retrieve all immediate children in the folder
   1. Determine if any of those children were added, updated, or deleted and will update the database accordingly
   1. Filters the children that are folders and recursively runs this scan each of them (back to number `1`).
   1. If metadata refreshing was enabled for this scan:
      1. Refresh metadata for this folder
      1. Refresh metadata for all children
         1. For each child, it will recursively call `4` for each of them.

Once the root folder is done, it runs some post scan tasks.

It's important to note that most of these methods are async, however they are all currently called sequentially.

This change lets it fan out the recursive calls for `3` and the metadata refresh calls for `4ii`.

Fanning out the folder scan is fine since that's mostly running simple directory enumerations / database updates. However running too many metadata refreshes at same time could be a bit CPU intensive. I set it up to limit the amount of concurrent metadata refreshes (defaults to amount of CPU cores). This means it will limit the amount of ffprobe/ffmpeg processes running at once.
From my testing each individual ffprobe is not that expensive so spawning 8 of them on an 8 core system is just fine. In practice you could make this number much higher. On a 24 core system, setting it to 128 makes the metadata scan go very fast while staying around ~30% cpu usage.

**Scan progress**
Something else worth discussing is how the progress indicator works during a scan:
   - It starts at 0% and moves to 5% once the folder enumerates all it's direct children.
   - Moves to 10% after it updates the database based on what was added, updated, or removed.
  -  Moves between 10% and 90% as it's scanning all subfolders.
   - Moves between 90% and 100% as it's refreshing metadata for all subfolders.

So 80% is for subfolder scanning, 10% is for metadata refreshing, and those are the only two that show real progress while running. Personally I have seen the scan stuck around the 90% mark for long periods of time. This is likely due to the fact that metadata refreshing is slower than just enumerating files on the disk.

Based on that, I changed the distribution of the progress in the scan:
   - Moves between 10% and __50%__ as it's scanning all subfolders.
   - Moves between __50%__ and 100% as it's refreshing metadata for all subfolders.

Expanding the breadth of the metadata refresh part will allow users to see progress more frequently.

**Results**
[...]

PR #17381, opened 2026-07-18
Fix music and series full library metadata refresh hang by handrolling a working implementation from movies library refresh

jsphgei-dot:
**Changes**
Before:
 - Both Music and Series methods walked an async for each chain that await-ran RefreshMetadata with no timeout and no failure logs.
After:
 - Both Music and Series's refresh metadata functions now utilize the ILimitedConcurrencyLibraryScheduler interface's enqueue method to concurrently (assigns runners according to cpu core related calculations, falls back to sequential if refresh was queued across the whole library due to anti-deadlock) refresh metadata with respect to the LibraryScanFanoutConcurrency setting under the library settings.
 - each metadata refresh child is capped to 60 seconds using a linked CancellationTokenSource to undercut 100 second http timer and attribute the error to the child item
 - try/catch identifies which child item went wrong
 - each child item becomes its own work item, doesn't take down the whole series's run
 - timeouts and failures now log the item type, name and path with a warning at the end that gives the failed count

**Code Assistance**
 - Used ai assistance to discover code inconsistencies and find the most elegant fix
 - Verified the fix by running a metadata refresh on a consistently (attempted retries, server restart, checked the db status, I'm aware removing and putting back the series fixes the issue) failing/hanging tv series with 200+ (Naruto 2004 and Jojo parts 1-5) episodes and multiple seasons that updated with no hangs after patch test.

**Issues**
Might be related to Issue #15075

### DIFF
--- a/MediaBrowser.Controller/Entities/Audio/MusicAlbum.cs
+++ b/MediaBrowser.Controller/Entities/Audio/MusicAlbum.cs
@@ -192,10 +192,17 @@ namespace MediaBrowser.Controller.Entities.Audio
                         childUpdateType |= updateType;
                     }
                 }
-                catch (OperationCanceledException) when (!cancellationToken.IsCancellationRequested)
+                catch (OperationCanceledException ex)
                 {
+                    if (cancellationToken.IsCancellationRequested)
+                    {
+                        // The caller's token was cancelled, not our own timeout. Propagate it.
+                        throw;
+                    }
+
                     Interlocked.Increment(ref numFailed);
                     Logger.LogError(
+                        ex,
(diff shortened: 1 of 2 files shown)


jsphgei-dot:
Follow-up suggestion (out of scope for this PR): after this fix, `Series.RefreshAllMetadata` and `MusicAlbum.RefreshAllMetadata` (the two `IMetadataContainer` implementations) now share nearly identical boilerplate, bounded concurrency via `ILimitedConcurrencyLibraryScheduler`, a per-item timeout, and error/timeout logging, with only their domain-specific ordering and business rules differing on top.

Worth considering extracting that shared boilerplate (timeout wrapper + logging + scheduler call) into a small helper both can call, so a future third `IMetadataContainer` implementer doesn't have to reinvent it (which is effectively how this bug happened in the first place: two independent hand-rolled loops that both missed the concurrency work done in #4242 for the generic `Folder` path). Happy to open that as a separate PR if there's interest.

Issue #16975, opened 2026-05-30
[bug] UNIQUE constraint failed: UserData.ItemId, UserData.UserId, UserData.CustomDataKey during library cleanup of duplicate media


eliminat:
### Description of the bug

When duplicate items exist in libraries under different `ItemId`s (due to separate files or failed grouping rules) but resolve to the same IMDb/TMDb metadata ID, they generate separate `UserData` rows with the same `(UserId, CustomDataKey)`. 

When library cleanup operations run to delete one of the duplicate entries, Jellyfin tombstones the deleted item's history by updating its `ItemId` in the `UserData` table to the null-GUID `00000000-0000-0000-0000-000000000001`. If multiple rows being updated share the same `(UserId, CustomDataKey)`, updating both to the same null-GUID violates the composite primary key constraint on `UserData` (Primary Key: `ItemId, UserId, CustomDataKey`). This aborts the transaction, crashing the library scan and preventing new media discovery.

### Reproduction steps

1. Add two duplicate copies of the same episode to a library, named such that they index as two distinct items.
2. Mark both episodes as watched for a user so that both get separate `UserData` records (same `UserId` and `CustomDataKey`, but different `ItemId`s).
3. Delete or rename one of the duplicate files on disk.
4. Run a "Scan Media Library" task.
5. The library monitor will find the stale directory, attempt to clean it up, and fail with a primary key constraint error on the SQLite database, aborting the scan.

### What is the current _bug_ behavior?

The library scan aborts prematurely and rolls back all newly discovered library additions.

### What is the expected _correct_ behavior?

* The library cleanup routine should handle watch-state resolution (e.g., merging records or deleting excess histories) instead of executing updates that create duplicate primary keys.
* A database transaction rollback on a single deleted item's watch history should not abort the entire library scan process.

### Jellyfin Server version

10.11.9

### Specify version number

10.11.9

### Specify the build version

10.11.9

### Environment

```markdown
- OS: Arch Linux on Host, Debian GNU/Linux 13 (Trixie) inside Container
- Linux Kernel: 7.0.9-1-cachyos
- Virtualization: Docker (Bare Metal Host)
- Clients: N/A (Server-side background task crash)
- Browser: N/A
- FFmpeg Version: 7.1.3-Jellyfin
- Playback Method: N/A
- Hardware Acceleration: Intel QuickSync / VAAPI (Intel Arc A380, i915 driver) - N/A to this database bug
[...]

Issue #15496, opened 2025-11-13
[BUG][10.11.2] Database Error Deleting Series


WHiZSTA:
Fresh 10.11.2 windows install with non-migrated freshly made library

when trying to delete a series it errors out, with a db error related to foreign keys constraint.

`[2025-11-12 23:26:54.887 -05:00] [INF] [3] Emby.Server.Implementations.Library.LibraryManager: Removing item, Type: "Series", Name: "<redacted>", Path: "E:\Video\TV Shows\<redacted>", Id: 3aee3363-efb3-db04-52f3-5a0e1fecfc44
[2025-11-12 23:26:54.905 -05:00] [INF] [3] Emby.Server.Implementations.Library.LibraryManager: Deleting item path, Type: "Series", Name: "<redacted>", Path: "E:\Video\TV Shows\<redacted>", Id: 3aee3363-efb3-db04-52f3-5a0e1fecfc44
[2025-11-12 23:26:55.363 -05:00] [ERR] [3] Microsoft.EntityFrameworkCore.Database.Command: Failed executing DbCommand ("0"ms) [Parameters=["@__date_1='?' (DbType = DateTime), @__p_0='?' (Size = 3979)"], CommandType='Text', CommandTimeout='30']"
""UPDATE \"UserData\" AS \"u\"
SET \"ItemId\" = '00000000-0000-0000-0000-000000000001',
    \"RetentionDate\" = @__date_1
WHERE \"u\".\"ItemId\" IN (
    SELECT \"p\".\"value\"
    FROM json_each(@__p_0) AS \"p\"
)"
[2025-11-12 23:26:55.366 -05:00] [ERR] [3] Microsoft.EntityFrameworkCore.Query: An exception occurred while executing an 'ExecuteUpdate' operation for context type '"Jellyfin.Database.Implementations.JellyfinDbContext"'."
""Microsoft.Data.Sqlite.SqliteException (0x80004005): SQLite Error 19: 'FOREIGN KEY constraint failed'.
   at Microsoft.Data.Sqlite.SqliteDataReader.NextResult()
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteReader(CommandBehavior behavior)
   at Microsoft.EntityFrameworkCore.Storage.RelationalCommand.ExecuteNonQuery(RelationalCommandParameterObject parameterObject)
[...]

Issue #14295, opened 2025-06-13
[10.11.0-RC1] Foreign Key Errors


theguymadmax:
### Description of the bug

After an upgrade from 10.10.7 to 10.11(Master) a scan library  produces Foreign Key errors

### Reproduction steps

See above

### What is the current _bug_ behavior?

Errors in log leading to  a failed scan

### What is the expected _correct_ behavior?

Scan without error.

### Jellyfin Server version

Master

### Specify commit id

23c25289dae40f292f0ffcf8c83f8ea2f9eebd61

### Specify the build version

2025061219 (Master

### Environment

```markdown
- OS: Linux
- Linux Kernel:6.11.0-26-generic
- Virtualization:Docker
- Plugins: None
- Jellyfin Data Storage: 256GB SSD Ext4
```

### Jellyfin logs

```shell
[2025-06-12 22:54:26.359 +00:00] [INF] [69] Emby.Server.Implementations.IO.LibraryMonitor: Stopping directory watching for path "/MediaD/MediaD/Movies"
[2025-06-12 22:54:26.361 +00:00] [INF] [69] Emby.Server.Implementations.Library.LibraryManager: Validating media library
[2025-06-12 22:54:37.478 +00:00] [INF] [65] Emby.Server.Implementations.Library.LibraryManager: Removing item, Type: "Movie", Name: "Movie", Path: "/MediaD/MediaD/Movies/Movie(1988)/filename.mkv", Id: a7749719-cb96-646c-e4f4-b48eae056775
[2025-06-12 22:54:37.786 +00:00] [INF] [65] Emby.Server.Implementations.Library.LibraryManager: Removing item, Type: "Movie", Name: "Movie", Path: "/MediaD/MediaD/Movies/Movie(1988)/filename.mkv", Id: 552c977c-7169-ddc7-6755-39fbdcd6433e
[2025-06-12 22:54:37.801 +00:00] [ERR] [65] Microsoft.EntityFrameworkCore.Database.Command: Failed executing DbCommand ("2"ms) [Parameters=["@__date_1='?' (DbType = DateTime), @__id_0='?' (DbType = Guid)"], CommandType='Text', CommandTimeout='30']"
""UPDATE \"UserData\" AS \"u\"
SET \"ItemId\" = '00000000-0000-0000-0000-000000000001',
    \"RetentionDate\" = @__date_1
WHERE \"u\".\"ItemId\" = @__id_0"
[...]

PR #17390, opened 2026-07-20
Recreate missing UserData placeholder BaseItem on startup

jvandenbroek:
## Summary

`BaseItemRepository.DeleteItem()` detaches `UserData` by reassigning it to a well-known placeholder `BaseItem` (`00000000-0000-0000-0000-000000000001`), seeded by the `DetachUserDataInsteadOfDelete` migration. If that seeded row is ever missing on a given server (observed in the wild — migration recorded as applied in `__EFMigrationsHistory`, but the row itself absent from `BaseItems`), every subsequent item delete fails with a SQLite `FOREIGN KEY constraint failed` on the `UserData` reassignment `ExecuteUpdate`.

Library scans call `DeleteItem` per-item for entries they believe are gone, including transient cases (e.g. a brief stat failure on a network filesystem during a large recursive scan). The scan swallows this exception per item and continues, so the failure is invisible in normal use — but the item's `DateModified` is still bumped before the failure, so it keeps surfacing as continuous "everything changed" library-sync churn to companion clients (observed via the Kodi Sync Queue plugin: hundreds of items reported "updated" every night) with no visible server-side error apart from log noise.

This is likely related in spirit (same `UserData` tombstone mechanism, different trigger) to #15496, #15578, #14295, and #16975.

## Fix

Add a startup migration routine (`EnsureUserDataPlaceholderItem`) that checks for the placeholder row and recreates it — using the exact same values as the original seed migration — if it's missing, so a server that lost this row for whatever reason self-heals instead of failing this way indefinitely.

## Test plan

- [x] `dotnet build Jellyfin.Server/Jellyfin.Server.csproj` succeeds with no new warnings/errors
- [x] Reproduced the underlying `FOREIGN KEY constraint failed` on a production database missing the placeholder row (confirmed via the exact `UPDATE UserData SET ItemId = <placeholder> ...` statement from `DeleteItem`)
- [x] Manually re-inserted the placeholder row with the same shape this routine produces; confirmed `PRAGMA foreign_key_check` is clean afterward and the previously-failing `UserData` reassignment now succeeds
- [ ] Did not add an automated test — no existing test coverage for this class of startup `IAsyncMigrationRoutine` (checked `tests/Jellyfin.Server.Implementations.Tests`), followed existing convention

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Jellyfin.Server/Migrations/Routines/EnsureUserDataPlaceholderItem.cs
+++ b/Jellyfin.Server/Migrations/Routines/EnsureUserDataPlaceholderItem.cs
@@ -1,0 +1,65 @@
+#pragma warning disable RS0030 // Do not use banned APIs
+using System;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Jellyfin.Database.Implementations;
+using Jellyfin.Database.Implementations.Entities;
+using Jellyfin.Server.Implementations.Item;
+using Jellyfin.Server.ServerSetupApp;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.Logging;
+
+namespace Jellyfin.Server.Migrations.Routines;
+
+/// <summary>
+/// Recreates the placeholder BaseItem used to detach UserData on item deletion
+/// (see migration DetachUserDataInsteadOfDelete) if it is missing.
+/// Without it, every <see cref="BaseItemRepository.DeleteItem"/> call fails with a
+/// FOREIGN KEY constraint violation, which library scans swallow per-item, leaving the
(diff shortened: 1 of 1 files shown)


cvium:
Please review our LLM policy: https://jellyfin.org/docs/general/contributing/llm-policies/

> You must review the output and be able to explain in the PR body - without LLM output as noted above - what is being changed and why. Your PR body (and, if applicable, commit bodies) should be providing context to other developers about why a change was made, and if your name is on it, we want your words and explanations, not an LLM's. If you can't explain what the LLM did, we are not interested in the change.

PR #11191, opened 2024-03-20
Enable nullable for LibraryManager

Bond-009:
Changes nullability of some API params

PR #17394, opened 2026-07-21
Avoid NRE when sorting by user-dependent keys without a user

### LINKED ISSUE #17393: NullReferenceException in UserDataManager.GetUserData when sorting by user-dependent keys without a user
## Description of the bug

A `GET /Items` (or any library query) that is sorted by a user-dependent
key (`PlayCount`, `IsFavoriteOrLiked`, `DatePlayed`, `IsPlayed`,
`IsUnplayed`) but carries no `User` fails with HTTP 500. The request is
dispatched with a `NullReferenceException` inside
`UserDataManager.GetUserData`, which surfaces through the LINQ sort as
`System.InvalidOperationException: Failed to compare two elements in the
array`.

This is reached by any anonymous / API-key request (no `userId`) that
specifies one of the user-dependent sort keys — e.g. a client or
third-party integration calling `/Items?sortBy=PlayCount` with only an API
key. `ItemsController.GetItems` documents that `user == null` throughout
when an API key is used.

## Reproduction steps

1. Make an authenticated (API-key, no user) request to `/Items` sorting by
   a user-dependent key, against a library with at least one item. Example:
   `GET /Items?recursive=true&mediaTypes=Audio&sortBy=PlayCount&api_key=<API_KEY>`
   (no `userId`).
2. Observe the response.

## What is the current (bug) behavior?

The request returns HTTP 500. The server log shows:

```
[ERR] Error processing request. URL "GET" "/Items".
System.InvalidOperationException: Failed to compare two elements in the array.
 ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at Emby.Server.Implementations.Library.UserDataManager.GetUserData(User user, BaseItem item)   # 10.11.z: user.InternalId
   at Emby.Server.Implementations.Sorting.PlayCountComparer.GetValue(BaseItem x)
   at Emby.Server.Implementations.Sorting.PlayCountComparer.Compare(BaseItem x, BaseItem y)
   at System.Linq.Enumerable.EnumerableSorter`2.CompareAnyKeys(...)
   ...
   at MediaBrowser.Controller.Entities.UserViewBuilder.SortAndPage(...)
   at MediaBrowser.Controller.Entities.Folder.PostFilterAndSort(...)
   at Jellyfin.Api.Controllers.ItemsController.GetItems(...)
```

Root cause: `LibraryManager.GetComparer` assigns
`userComparer.User = user` without a null guard. The comparer
(`PlayCountComparer`) then calls `UserDataManager.GetUserData(user, item)`,
which dereferences the null user (`user.InternalId` on 10.11.z; on master
the overload was rewritten and dereferences `user.Id`).

## What is the expected (correct) behavior?

An anonymous/API-key request sorted by a user-dependent key should not
500. The expected behavior is to either (a) return results in a
deterministic fallback order (e.g. `SortName`, which Jellyfin already uses
as the canonical tiebreaker in `ItemsController`), or (b) return a clean
error indicating the combination is unsupported — rather than an opaque NRE
that crashes the request.

## Jellyfin Server version

- Reproduced on **10.11.11** (stable) — NRE at `user.InternalId`.
- Also affects **master** — the `GetUserData(User, BaseItem)` overload was
  rewritten to use `ResolveUserDataRow`; it no longer dereferences
  `user.InternalId` but still dereferences `user.Id`, so the NRE still
  occurs at a different line. `LibraryManager.GetComparer` is identical on
  master and has the same unguarded `userComparer.User = user` assignment.

## Build version

10.11.11 (stable); reproduced against the `release-10.11.z` tag and current
`master`.

## Environment

- Jellyfin 10.11.11 running in a container (podman).
- Library with a mix of Audio items.
- Triggered while developing a third-party Jellyfin plugin (the
  jellyfin-alexa-plugin) that issues `/Items` queries sorted by
  `PlayCount`; the crash was observed against both the plugin's queries and
  the generic `/Items` endpoint.

## Jellyfin logs

See the stack trace in "current bug behavior" above (redacted API key).
The NRE fires per-comparison during the LINQ sort, and when wrapped in a
retry loop (as the plugin does) it repeats until the request budget is
exhausted.

## Additional information

This was discovered while working on a third-party Jellyfin plugin,
[jellyfin-alexa-plugin](https://github.com/paoloantinori/jellyfin-alexa-plugin)
(an Alexa skill plugin for Jellyfin): the plugin's artist-song query sorts
by `PlayCount`, and anonymous/API-key-scope requests crashed with this NRE.
The plugin-side workaround (avoiding the trigger) landed in commit
[paoloantinori/jellyfin-alexa-plugin@25e5446](https://github.com/paoloantinori/jellyfin-alexa-plugin/commit/25e5446);
this issue tracks the underlying server-side bug.

Note: `DateLastMediaAddedComparer` declares `: IUserBaseItemComparer` but
its `Compare`/`GetDate` is `static` and never references `User` — it is
statically mis-tagged as user-dependent. That is a separate (related)
issue and will be filed/handled independently.

paoloantinori:
## Changes

Fixes a `NullReferenceException` in `UserDataManager.GetUserData` that 500-ed any anonymous (API-key, no `userId`) `/Items` request sorted by a user-dependent key (`PlayCount`, `IsFavoriteOrLiked`, `DatePlayed`, `IsPlayed`, `IsUnplayed`).

- `LibraryManager.GetComparer`: when `user` is null and the requested key requires a user (`IUserBaseItemComparer`), substitute the `SortName` comparer so the result stays deterministic instead of crashing. `SortName` is the project's canonical tiebreaker (already used by `ItemsController` for album-by-artist). This closes the gap left when `GetComparer`'s `user` parameter was widened to nullable in #11191.
- `UserDataManager.GetUserData`: `ArgumentNullException.ThrowIfNull(user)` as defense in depth (matches the existing guards on the `SaveUserData` overloads in the same file), so any future null-user caller gets a clean argument exception instead of an opaque NRE deep in a sort comparer.
- `DateLastMediaAddedComparer`: re-tagged from `IUserBaseItemComparer` to `IBaseItemComparer` (its `GetDate` is `static` and never reads `User`), and dropped the unused `User`/`UserManager`/`UserDataManager` properties. Without this, the new `SortName` fallback would wrongly engage for `DateLastContentAdded` on anonymous queries, returning `SortName` order instead of date order.

The crash surfaced as `System.InvalidOperationException: Failed to compare two elements in the array` (wrapping the NRE from the LINQ sort).

## Code assistance

None — hand-written null-guard fix.

## Tests

- `UserDataManagerTests.GetUserData_NullUser_ThrowsArgumentNullException` — reproduces the crash (NRE → now `ArgumentNullException`).
- `LibraryManagerSortTests.Sort_UserDependentKey_NullUser_FallsBackToSortNameWithoutThrowing` — `Sort` with a user-dependent key + null user no longer throws and returns items ordered by the `SortName` fallback (direction preserved).
- `LibraryManagerSortTests.Sort_DateLastContentAdded_NullUser_OrdersByDateNotSortName` — guards that `DateLastContentAdded` still sorts by date with no user (fixture chosen so date-desc and SortName-desc disagree, so a revert is caught).

Full `Jellyfin.Server.Implementations.Tests` suite: 423 passed, 0 failed.

## Issues

Fixes #17393

---

Discovered while working on [jellyfin-alexa-plugin](https://github.com/paoloantinori/jellyfin-alexa-plugin); the plugin-side workaround is [paoloantinori/jellyfin-alexa-plugin\@25e5446](https://github.com/paoloantinori/jellyfin-alexa-plugin/commit/25e5446).

### DIFF
--- a/Emby.Server.Implementations/Library/LibraryManager.cs
+++ b/Emby.Server.Implementations/Library/LibraryManager.cs
@@ -1969,9 +1969,16 @@ namespace Emby.Server.Implementations.Library
         {
             var comparer = Comparers.FirstOrDefault(c => name == c.Type);
 
-            // If it requires a user, create a new one, and assign the user
+            // User-dependent comparers (IUserBaseItemComparer) need a User. With no user
+            // (anonymous/API-key /Items requests), substitute the SortName comparer so the
+            // result stays deterministic instead of 500-ing with an NRE inside the LINQ sort.
             if (comparer is IUserBaseItemComparer)
             {
+                if (user is null)
+                {
+                    return Comparers.FirstOrDefault(c => c.Type == ItemSortBy.SortName);
+                }
+
                 var userComparer = (IUserBaseItemComparer)Activator.CreateInstance(comparer.GetType())!; // only null for Nullable<T> instances
 
                 userComparer.User = user;
--- a/Emby.Server.Implementations/Library/UserDataManager.cs
+++ b/Emby.Server.Implementations/Library/UserDataManager.cs
@@ -238,6 +238,7 @@ namespace Emby.Server.Implementations.Library
         /// <inheritdoc />
         public UserItemData GetUserData(User user, BaseItem item)
         {
+            ArgumentNullException.ThrowIfNull(user);
             var cacheKey = GetCacheKey(user.InternalId, item.Id);
             return _cache.GetOrAdd(
                 cacheKey,
--- a/Emby.Server.Implementations/Sorting/DateLastMediaAddedComparer.cs
+++ b/Emby.Server.Implementations/Sorting/DateLastMediaAddedComparer.cs
@@ -3,34 +3,14 @@
 
 using System;
 using Jellyfin.Data.Enums;
-using Jellyfin.Database.Implementations.Entities;
 using MediaBrowser.Controller.Entities;
-using MediaBrowser.Controller.Library;
 using MediaBrowser.Controller.Sorting;
 using MediaBrowser.Model.Querying;
 
 namespace Emby.Server.Implementations.Sorting
 {
-    public class DateLastMediaAddedComparer : IUserBaseItemComparer
+    public class DateLastMediaAddedComparer : IBaseItemComparer
     {
-        /// <summary>
-        /// Gets or sets the user.
-        /// </summary>
-        /// <value>The user.</value>
-        public User User { get; set; }
-
-        /// <summary>
-        /// Gets or sets the user manager.
-        /// </summary>
-        /// <value>The user manager.</value>
-        public IUserManager UserManager { get; set; }
-
-        /// <summary>
-        /// Gets or sets the user data manager.
-        /// </summary>
-        /// <value>The user data manager.</value>
-        public IUserDataManager UserDataManager { get; set; }
-
         /// <summary>
         /// Gets the name.
         /// </summary>
--- a/tests/Jellyfin.Server.Implementations.Tests/Library/LibraryManagerSortTests.cs
+++ b/tests/Jellyfin.Server.Implementations.Tests/Library/LibraryManagerSortTests.cs
@@ -1,0 +1,105 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using AutoFixture;
+using AutoFixture.AutoMoq;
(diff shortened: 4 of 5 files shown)


Bond-009:
We're done with 10.11 releases

PR #17404, opened 2026-07-22
Fix NullReferenceException in user-data query filters

### LINKED ISSUE #17403: Concurrent IsFavorite/IsLiked queries throw NullReferenceException and stall (release-10.11.z)
**Closing as invalid — filed on a misdiagnosis, apologies for the noise.**

On re-verification, `v12.0-rc3` is built from `master` (identical commit `fc43f15`), and the `FirstOrDefault(f => f.UserId == filter.User!.Id)!.IsFavorite` pattern referenced below does **not** exist there — `master`/rc3 already use the null-safe `.Any(...)` form (in `BaseItemRepository.TranslateQuery.cs`). That unsafe pattern only exists on the separate, older `release-10.11.z` branch.

Crucially, the `NullReferenceException` I observed reproduces on rc3 **against the already-null-safe code**, and only under concurrent requests (a single request succeeds). That means it has a **different root cause** than the null-forgiving deref described here — most likely concurrent `DbContext` use rather than a query-translation null deref. I'm withdrawing this report and will open a correctly-scoped one if further investigation confirms a specific, reproducible cause.

---

_Original (incorrect) report retained below for reference:_

On the 10.11 release line (tested on 12.0-rc3), concurrent `GET /Items?Filters=IsFavorite&Recursive=true` requests intermittently throw a server-side `NullReferenceException` inside EF Core's LINQ expression interpreter and stall. A single such request is fast; the failure only appears under concurrent load.

MaxRink:
**Changes**

Rewrites the `IsLiked`, `IsFavoriteOrLiked`, `IsFavorite`, `IsResumable` and the Series `IsPlayed` user-data filters in `BaseItemRepository.TranslateQuery` to use a null-safe `.Any(...)` form instead of the chained `!.FirstOrDefault(...)!.Property` deref.

The old form cannot be translated to SQL by EF Core and is evaluated client-side in the expression interpreter, where it throws a `NullReferenceException` for items that have no matching `UserData` row. This surfaces intermittently under the concurrent `Filters=IsFavorite&Recursive=true` requests the web client issues (e.g. the Favorites view, one request per item type): a single request is fast (~13 ms on a ~189k-item library) while several concurrent ones throw and stall.

This ports the null-safe approach already present on `master` (where the query translation was refactored into `BaseItemRepository.TranslateQuery.cs`) back to the `release-10.11.z` line, so it is a targeted backport rather than new behaviour.

Note on the `IsResumable` false branch: the rewrite `!Any(... PlaybackPositionTicks > 0)` also correctly matches items that have no `UserData` row at all (previously those NRE'd), which is the intended "not resumable" set.

**Code assistance**

None.

**Issues**

Fixes #17403

### DIFF
--- a/Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
+++ b/Jellyfin.Server.Implementations/Item/BaseItemRepository.cs
@@ -2053,20 +2053,23 @@ public sealed class BaseItemRepository
 
         if (filter.IsLiked.HasValue)
         {
+            var isLiked = filter.IsLiked.Value;
             baseQuery = baseQuery
-                .Where(e => e.UserData!.FirstOrDefault(f => f.UserId == filter.User!.Id)!.Rating >= UserItemData.MinLikeValue);
+                .Where(e => e.UserData!.Any(ud => ud.UserId == filter.User!.Id && ud.Rating >= UserItemData.MinLikeValue) == isLiked);
         }
 
         if (filter.IsFavoriteOrLiked.HasValue)
         {
+            var isFavoriteOrLiked = filter.IsFavoriteOrLiked.Value;
             baseQuery = baseQuery
-                .Where(e => e.UserData!.FirstOrDefault(f => f.UserId == filter.User!.Id)!.IsFavorite == filter.IsFavoriteOrLiked);
+                .Where(e => e.UserData!.Any(ud => ud.UserId == filter.User!.Id && ud.IsFavorite) == isFavoriteOrLiked);
         }
 
         if (filter.IsFavorite.HasValue)
         {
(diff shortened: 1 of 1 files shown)


MaxRink:
Closing this — I filed it based on a misdiagnosis, apologies for the noise.

The NullReferenceException I was chasing was observed on 12.0-rc3, which is built from `master` (identical commit). `master`/rc3 already use the null-safe `.Any(...)` form this PR was backporting to `release-10.11.z`, so this change does not address the crash I actually hit. The remaining unsafe `!.FirstOrDefault(...)!` pattern on `release-10.11.z` was not something I reproduced a failure against, so I don't want to leave a speculative, mislinked PR open. Withdrawing.

PR #17406, opened 2026-07-22
Warm up ItemsByName query plans at startup to avoid ~37s first-request stall

### LINKED ISSUE #17405: 12.0-rc3: ItemsByName queries (Artists/Studios/Genres) with IsFavorite take ~37s cold on a large library; login stalls the same ~37s
### Description of the bug

On 12.0-rc3 (built from `master`, commit `fc43f15`) with a large library (~189,000 items; `ItemValuesMap` ~777k rows), the ItemsByName endpoints (`GET /Artists`, `/Studios`, `/Genres` with `Filters=IsFavorite&Recursive=true`) take **~32 to 38 seconds on every request** (not just cold). The web client issues these when rendering favorites/home surfaces, so the UI stalls for tens of seconds; login (`POST /Users/authenticatebyname`) also stalls the same ~37 s when per-user data is loaded.

> **Correction (see comments):** an earlier revision of this issue claimed these were "~37 s cold, ~30 ms warm". That was wrong, i had only measured a warm `/Persons` call and generalised. Re-measured, `/Artists` is slow on **every** call (32.9 / 32.5 / 32.6 s sequential on production). The cost is per-call query **execution**, not a one-time plan compilation. Details, full stats and candidate benchmarks are in the comments.

Related: #17269 (person-filtered variant, closed as improved-for-12.0).

### Reproduction steps

1. Large library (repro: ~189k items, ~777k `ItemValuesMap` rows), on 12.0-rc3.
2. As an authenticated user, call `GET /Artists?Filters=IsFavorite&Recursive=true&Limit=100` (also `/Studios`, `/Genres`).
3. Observe ~32 to 38 s **on every call**, not just the first.

### What is the current behaviour?

`/Artists`|`/Studios`|`/Genres` with a user-data filter take tens of seconds per request. Parameter isolation shows the cost splits roughly evenly between the ItemsByName base query (`ItemValuesMap`→`BaseItems`) and the `ItemCounts` aggregation (`BuildItemCountsByCleanName`): adding `&enableTotalRecordCount=false` roughly halves the time (32.7 s → 16.4 s). The underlying SQLite executes the equivalent query in milliseconds.

### Expected behaviour

These queries should complete in roughly the time the database needs (milliseconds plus reasonable overhead).

### Jellyfin Server version

12.0-rc3

### Specify commit id

fc43f151a2418cc112e116050a99dd6318917ab0

### Environment

```
- OS: Linux (Docker, TrueNAS SCALE app)
- Jellyfin: 12.0-rc3 (master @ fc43f15)
- Database: SQLite (~1.4 GB jellyfin.db; BaseItems 189,261; ItemValuesMap 777,362)
- Library scale: ~189,000 items
- DB storage: NVMe
```

### Additional information

Benchmarked candidate fixes (all against a copy of the production DB): global `QuerySplittingBehavior.SplitQuery`, no change; `Shadowghost@0363489` "Flatten queries", no change here. Full stats, methodology, and per-parameter breakdown are in the comments.

MaxRink:
**Changes**

Adds `QueryPlanWarmupHost`, an `IHostedService` that, once at startup on a background thread, executes the ItemsByName queries (`GetArtists`/`GetAlbumArtists`/`GetStudios`/`GetGenres`/`GetMusicGenres`) with a user-data filter (`IsFavorite=true, Recursive=true, Limit=1`) and discards the results. This moves the one-time EF Core query-plan compilation cost off the first user request.

**Why**

On a large library (~189k items) these ItemsByName + user-data-filter query shapes take ~37 s on their first execution after a server restart, then ~30 ms once warm. Profiling (see #17405) shows the cold cost is EF Core building the query's shaper/materializer expression tree and `Expression.Compile()` + JIT compiling the generated delegates — CPU-bound on a single core, not SQL (the underlying SQLite query returns in milliseconds). Because the plan is cached per distinct query, the first Favorites-view load (or even login, which triggers per-user data) blocks for the full compile time. Warming these plans at startup pays that cost before any request arrives.

The five shapes are independent and each `ILibraryManager` call opens its own `DbContext`, so they are compiled concurrently via `Parallel.Invoke` to reduce total warm-up wall-time. Warm-up is best-effort: failures are swallowed and logged at debug level, and it never blocks startup.

This is additive and changes no query behaviour or results.

**Notes / status**

- Draft. Verified it **compiles** against `master` (`fc43f15`, i.e. 12.0-rc3) with the .NET 10 SDK; I have **not** yet measured the end-to-end startup warm-up effect on a live large-library instance, and there is no test yet. Feedback welcome on whether a startup hosted-service is the preferred approach vs. `EF.CompileQuery`/precompiled queries or reducing the shaper complexity for these paths.
- The warmed query shapes are a reasonable representative set for the ItemsByName + user-data paths, but may not cover every variant the web client issues.

**Code assistance**

Developed with AI assistance for profiling and drafting.

**Issues**

Fixes #17405

### DIFF
--- a/Jellyfin.Server/QueryPlanWarmupHost.cs
+++ b/Jellyfin.Server/QueryPlanWarmupHost.cs
@@ -1,0 +1,102 @@
+using System;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Jellyfin.Database.Implementations.Entities;
+using MediaBrowser.Controller.Entities;
+using MediaBrowser.Controller.Library;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+
+namespace Jellyfin.Server
+{
+    /// <summary>
+    /// <see cref="IHostedService"/> that pre-compiles the EF Core query plans for the
+    /// ItemsByName endpoints (Artists/AlbumArtists/Studios/Genres/MusicGenres) combined with
+    /// a user-data filter. These query shapes build large expression trees whose first-use
+    /// compilation (Expression.Compile + JIT) can take tens of seconds on large libraries,
+    /// blocking the first request (e.g. loading the Favorites view or logging in). Executing
+    /// them once at startup on a background thread moves that one-time cost off the request path.
+    /// This does not change any query behaviour; the results are discarded.
+    /// </summary>
+    public sealed class QueryPlanWarmupHost : IHostedService
+    {
+        private readonly ILogger<QueryPlanWarmupHost> _logger;
+        private readonly ILibraryManager _libraryManager;
+        private readonly IUserManager _userManager;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="QueryPlanWarmupHost"/> class.
+        /// </summary>
+        /// <param name="logger">The <see cref="ILogger{TCategoryName}"/>.</param>
+        /// <param name="libraryManager">The <see cref="ILibraryManager"/>.</param>
+        /// <param name="userManager">The <see cref="IUserManager"/>.</param>
+        public QueryPlanWarmupHost(
+            ILogger<QueryPlanWarmupHost> logger,
+            ILibraryManager libraryManager,
+            IUserManager userManager)
+        {
+            _logger = logger;
+            _libraryManager = libraryManager;
+            _userManager = userManager;
+        }
+
+        /// <inheritdoc />
+        public Task StartAsync(CancellationToken cancellationToken)
+        {
+            // Fire-and-forget on a background thread; never block startup.
+            _ = Task.Run(() => Warmup(cancellationToken), cancellationToken);
+            return Task.CompletedTask;
+        }
+
+        /// <inheritdoc />
+        public Task StopAsync(CancellationToken cancellationToken) => Task.CompletedTask;
+
+        private void Warmup(CancellationToken cancellationToken)
+        {
+            var user = _userManager.GetUsers().FirstOrDefault();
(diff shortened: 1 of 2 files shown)


MaxRink:
Update after benchmarking on a copy of the affected library (189k items) — **this warm-up approach does not fix the reported `/Artists` slowness, so this PR should not be treated as a fix for #17405.**

The premise here was that the ItemsByName + user-data-filter queries pay a one-time cold plan-compilation cost that warm-up could move off the request path. On re-measurement that premise is wrong: `/Artists?Filters=IsFavorite&Recursive=true` takes ~32–35 s on **every** call (32.9 / 32.5 / 32.6 s sequential), not just the first. Parameter isolation shows the cost is per-call query execution — roughly ~16 s in the ItemsByName base query and ~16 s in the `ItemCounts` aggregation (`&enableTotalRecordCount=false` halves it) — which pre-warming cannot help.

Leaving this open only as a reference for the narrower idea (pre-compiling genuinely cold-only query plans at startup), but it is **not** the fix for the issue. The real cost is characterised with full stats and candidate benchmarks in #17405.

MaxRink:
Closing: benchmarking on a copy of the affected 189k-item library showed this warm-up approach does not fix the problem. /Artists?Filters=IsFavorite&Recursive=true is ~32-35s on every call (32.9/32.5/32.6s sequential), not cold-only — the cost is per-call query execution (~16s ItemsByName base query + ~16s ItemCounts aggregation), which pre-warming cannot address. Full stats and candidate benchmarks are in #17405. Not the right fix; withdrawing.

PR #17409, opened 2026-07-22
Fix trickplay generation for hardware-scaled frames with off-by-a-pixel dimensions

### LINKED ISSUE #17346: Trickplay tiling fails on specific aspect ratio content on AMD GPUs
### Description of the bug

Trickplay image generation fails for a subset of my library with:

System.InvalidOperationException: Image width does not match provided width.
   at Jellyfin.Drawing.Skia.SkiaEncoder.CreateTrickplayTile(ImageCollageOptions options, Int32 quality, Int32 imgWidth, Nullable`1 imgHeight)
   at Jellyfin.Server.Implementations.Trickplay.TrickplayManager.CreateTiles(...)
   at Jellyfin.Server.Implementations.Trickplay.TrickplayManager.RefreshTrickplayDataInternal(...)

The ffmpeg step itself completes successfully — hardware init (d3d11va + opencl) works and frames are produced. The failure happens after ffmpeg, in the Skia tiling stage (CreateTrickplayTile), which asserts that every decoded frame is exactly the configured trickplay width and throws when a frame is off by a pixel or two.

This is the same failure as the confirmed #11409 and the stale #14869, still present in a current stable release. The previous issue (#14869) was closed as stale because it lacked the necessary steps to reliably reproduce the issue. I have since identified that the issue occurs with media having an aspect ratio of approximately 2.2:1, so I'm opening this new issue with the additional reproduction details.

### Reproduction steps

1. Enable hardware acceleration for trickplay (AMD, D3D11VA decode + OpenCL scaling).
2. Add media whose native aspect ratio does not divide cleanly into the configured trickplay width (e.g. a ~2.2:1 cinematic-ratio 1080p file).
3. Run the Generate Trickplay Images task.
4. Generation fails for those files; standard 16:9 SDR content in the same library generates fine.

Trickplay image generation with hardware acceleration enabled for the content with the mentioned aspect ratio fails.

Trickplay tiles are generated regardless of small dimension differences introduced by the hardware scaler, without throwing.

### Jellyfin Server version

10.11.8

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.8

### Environment

```markdown
- OS:Windows 10 IOT LTSC
- Linux Kernel: n/a
- Virtualization: n/a
- Clients: Jellyfin desktop client
- Browser: n/a
- FFmpeg Version: bundled jellyfin ffmpeg
- Playback Method: n/a
- Hardware Acceleration: AMD AMF
- CPU Model: ryzen 7 5700x3d
- GPU Model: radeon rx 9060xt
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Media Storage & Filesystem:
- External Integrations:
```

### Jellyfin logs

```shell
[INF] TrickplayManager: Creating trickplay files at 320 width, for "...MINDHUNTER (2017) - S01E01...[WEBRip-1080p]...x264...mkv"
[INF] MediaEncoder: Trickplay generation: ffmpeg -loglevel error -init_hw_device d3d11va=dx11:,vendor=0x1002 -init_hw_device opencl=ocl@dx11 -filter_hw_device ocl -hwaccel d3d11va -hwaccel_output_format d3d11 ... -vf "fps=0.1,setparams=...,hwmap=derive_device=opencl:mode=read,scale_opencl=w=320:h=144:format=nv12,hwmap=mode=read,format=nv12" -c:v mjpeg -qscale:v 4 -f image2 "...\%08d.jpg"
[ERR] TrickplayManager: Error creating trickplay images.
System.InvalidOperationException: Image width does not match provided width.
   at Jellyfin.Drawing.Skia.SkiaEncoder.CreateTrickplayTile(...)
   at Jellyfin.Server.Implementations.Trickplay.TrickplayManager.CreateTiles(...)
   at Jellyfin.Server.Implementations.Trickplay.TrickplayManager.RefreshTrickplayDataInternal(...)
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

**Why this does not reproduce on all media**

Jellyfin requests a fixed trickplay width (320 here) and computes a proportional height, then hands that target to the hardware scaler (scale_opencl on the D3D11VA → OpenCL path). The Skia tiling stage afterwards asserts that each decoded frame's width equals the configured width and throws otherwise.

Whether the hardware scaler's actual output lands exactly on the requested width depends on two things:

The source geometry. How cleanly the source's dimensions and aspect ratio map to the target width and to even values. A standard 16:9 1080p file scales to 320×180 with no drama. A non-standard ratio — e.g. the ~2.2:1 cinematic framing in the logs above, which produces 320×144 — is the kind of case where the scaler's alignment/rounding can land a pixel or two off the requested width.
The hardware scaler's own rounding/alignment. The OpenCL/D3D11 scaler (and its NV12 alignment requirements) can round differently than Jellyfin's fixed expectation. This is vendor/driver-dependent, which is why reports cluster around specific GPUs.

By contrast, the software trickplay path computes the scale filter deterministically (a trunc(...)*2 expression) and stays consistent with the width Jellyfin expects, so it doesn't hit the assert. That's also the current workaround: disabling hardware acceleration for trickplay makes these files generate successfully.

Net effect on a mixed library: most files generate fine, and a subset — those with unusual aspect ratios or resolutions on the hardware path — fail. It reads as random, but it's deterministic per file.

**Suggested fix direction**

In SkiaEncoder.CreateTrickplayTile, rather than throwing when a decoded frame's dimensions don't match the configured cell size, normalize the frame to the exact cell size before compositing (the class already has a ResizeImage helper). This keeps the tile geometry aligned with the width stored in trickplay metadata — so the web client's fixed-stride slicing stays correct — while tolerating the small dimension differences the hardware scaler introduces. There are two throw sites in that method (first-frame height derivation, and the per-frame composite loop); both would need to tolerate the mismatch.

AhmadRafiq90:
**Changes**
`CreateTrickplayTile` threw when a decoded frame's dimensions didn't exactly match the configured trickplay size. Hardware scalers (d3d11va/opencl) can emit frames a couple of pixels off — a 320px tier receiving 318px frames — aborting generation even though ffmpeg succeeded. Now the cell height is derived from the first frame's aspect ratio, and off-size frames are resized to the exact cell (via existing ResizeImage) instead of throwing. The cell stays locked to the configured width, so tile geometry and stored metadata stay aligned. Exact-size frames still composite directly, unchanged. Changes were tested on local build with a ~2.2:1 1080p source

**Code assistance**
LLM was used to find possible edge cases and check if code missed any obvious best practices

**Issues**
Fixes #17346

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -233,6 +233,7 @@
  - [MSalman5230](https://github.com/MSalman5230)
  - [dwandw](https://github.com/dwandw)
  - [Lampan-git](https://github.com/Lampan-git)
+ - [AhmadRafiq90](https://github.com/AhmadRafiq90)
 
 # Emby Contributors
 
--- a/src/Jellyfin.Drawing.Skia/SkiaEncoder.cs
+++ b/src/Jellyfin.Drawing.Skia/SkiaEncoder.cs
@@ -731,12 +731,9 @@ public class SkiaEncoder : IImageEncoder
                 throw new InvalidDataException("Could not decode image data.");
             }
 
-            if (firstImg.Width != imgWidth)
-            {
-                throw new InvalidOperationException("Image width does not match provided width.");
-            }
-
-            imgHeight = firstImg.Height;
+            var derivedHeight = (int)Math.Round(firstImg.Height * (double)imgWidth / firstImg.Width);
+            derivedHeight += derivedHeight % 2; // keep even for the JPEG encoder
+            imgHeight = derivedHeight;
         }
 
         // Make horizontal strips using every provided image.
@@ -760,17 +757,18 @@ public class SkiaEncoder : IImageEncoder
                     throw new InvalidDataException("Could not decode image data.");
                 }
 
-                if (img.Width != imgWidth)
+                if (img.Width != imgWidth || img.Height != imgHeight.Value)
                 {
-                    throw new InvalidOperationException("Image width does not match provided width.");
+                    // Normalize any off-by-a-few-pixels hardware-scaled frame to the exact
+                    // cell size so the grid stays aligned with the stored trickplay width.
+                    var targetInfo = new SKImageInfo(imgWidth, imgHeight.Value, img.ColorType, img.AlphaType, img.ColorSpace);
+                    using var resized = ResizeImage(img, targetInfo);
+                    canvas.DrawImage(resized, x * imgWidth, y * imgHeight.Value);
                 }
-
-                if (img.Height != imgHeight)
+                else
                 {
-                    throw new InvalidOperationException("Image height does not match first image height.");
+                    canvas.DrawBitmap(img, x * imgWidth, y * imgHeight.Value, DefaultSamplingOptions);
                 }
-
-                canvas.DrawBitmap(img, x * imgWidth, y * imgHeight.Value, DefaultSamplingOptions);
             }
         }
 


gnattu:
The bug makes sense but the approach to fix that is not my preferred way TBH. Hardware acceleration becomes less useful if each and every frame extracted need to be post processed by CPU and for trickplay, it is a lot of images to process. 

Can you be more specific like on what hardware platform we are seeing this commonly? I think this should be fixed on the ffmpeg side.

AhmadRafiq90:
@gnattu I've only reproduced it on AMD radeon rx 9060xt as I only have access to that but after manually running ffmpeg extraction with verbose logging shows that it may not be a hardware specific issue. The source is 1920x872 (SAR 1:1, DAR 240:109) and Jellyfin requests scale_opencl=w=320:h=144. However, the output is:
` 318x144 [SAR 5760:5777 DAR 240:109] `
if w=320 had been used, the SAR would be 108:109, instead 5760:5777 comes out to ~317.06 rounded up to 318 which is exactly the frame size observed and why skia throws currently.It also suggests the hardware scaler cant be trusted to always return the requested width. So let me know how you want me to rework it?

1) The tiling step is already entirely cpu side and the resizing an already decoded 318x144 bitmap is only triggered for some problematic video files not the complete library which only adds a minor performance impact.
2) Recording the measured width instead of asserting the configured one is an option and also gets rid of the per-frame resize cost you mentioned but that would require more changes.
3) investigate the filter side instead if you think scale_opencl should be honoring an explicit w=

nyanmisaka:
Try https://github.com/jellyfin/jellyfin-ffmpeg/releases/tag/v8.1.2-2

gnattu:
amd does not have hardware accelerated mjpeg encoder iirc and probably enabling hardware acceleration is in fact de-acceleration due to the extra overhead introduced by data moving and hardware setup. I really want to change the wording on that page because the term "acceleration" can easily make people think the hardware path, even partial, is always faster, it is not, especially for platforms does not even have a hardware mjpeg encoder. On a lot of platforms it is easy to produce negative performance impact, and now adding this extra overhead it is might be even slower.

But that also means the OpenCL scalar should not slightly change the dimension when it has no reason to, as the software jpeg encoder should be fine with 320 width. If the latest ffmpeg still produces wrong dimension we might need to figure out why.

AhmadRafiq90:
@nyanmisaka @gnattu works correctly w latest ffmpeg

PR #17420, opened 2026-07-23
Write resized-image cache entries atomically

### LINKED ISSUE #17135: Resized-image cache written non-atomically → 0-byte entries served permanently (blank posters/thumbnails)
- [x] This is a bug, not a question or configuration issue.
- [x] Searched existing issues; couldn't find this one.
- [x] Using an up-to-date version (10.11.11).
- [x] Single issue.

### Description of the bug

`ImageProcessor.ProcessImage` (`src/Jellyfin.Drawing/ImageProcessor.cs`) writes the resized image **directly to the final cache path** and gates the cache purely on `File.Exists` — no size/validity check, and the only lock (`_parallelEncodingLimit`) is **non-keyed**, so it caps *total* concurrent encodes but does **not** serialize encodes of the *same* file:

```csharp
if (!File.Exists(cacheFilePath))
{
    using (await _parallelEncodingLimit.LockAsync().ConfigureAwait(false))
    {
        resultPath = _imageEncoder.EncodeImage(originalImagePath, dateModified, cacheFilePath, ...); // direct write
    }
    ...
}

return (cacheFilePath, outputFormat.GetMimeType(), _fileSystem.GetLastWriteTimeUtc(cacheFilePath));
```

`EncodeImage` opens `cacheFilePath` with create/truncate, so the file is **0 bytes for the entire duration of the encode**. Two failure modes follow:

1. **Concurrent reads (transient):** when the library grid requests the same uncached image many times at once (first load), request A starts encoding (file now exists, 0 bytes), and requests B…N hit `!File.Exists(cacheFilePath)` → now **false** → they skip encoding and immediately serve the **0-byte file mid-write**.
2. **Crash/kill mid-encode (permanent):** if the encode is interrupted after the truncate but before the data is written (e.g. OOM/CPU starvation while a hardware/software transcode is hammering the box), the **0-byte file is left at `cacheFilePath` forever**, and `File.Exists` keeps reporting it as cached → that item shows a blank/blurhash placeholder permanently.

### Reproduction steps

1. Pick an item; request an **uncached** size with N parallel requests:
   `GET /Items/{id}/Images/Primary?fillHeight=361&fillWidth=240&format=Webp` ×30 in parallel.
2. **~11 of 30 return `HTTP 200` with `Content-Length: 0`.**
3. After a heavy first-load while the server is under CPU pressure, the on-disk cache is poisoned: on my box **131 of 179** files under `cache/images/resized-images/` were 0 bytes, each served blank to every client until manually deleted.

### What is the current bug behavior?
0-byte images returned (transiently under concurrency, permanently after an interrupted encode); affected posters/thumbnails render blank everywhere.

### What is the expected correct behavior?
A resized image is either fully present in the cache or absent — never a 0-byte file that's trusted and served.

### Suggested fix (verified locally)
Encode to a unique temp file in the same directory and `File.Move(temp, cacheFilePath, overwrite: true)` to publish atomically. With that change the same 30-parallel repro returns **0/30** zero-byte responses, and an interrupted encode leaves a junk temp instead of poisoning the cache. Happy to open a PR.

### Environment
- Jellyfin 10.11.11, Linux (NixOS), aarch64.

LTe:
**Changes**
ImageProcessor encoded resized images to file that need to be served. In the process it created empty file and then fill in with the results. In case of the parallel requests user can get empty file instead of processed image.

Encode to a unique temporary file and publish it with an atomic rename, so a cache entry is either fully present or absent. 

**Code assistance**

I was debugging root cause with LLM

**Issues**

Fixes #17135

### DIFF
--- a/src/Jellyfin.Drawing/ImageProcessor.cs
+++ b/src/Jellyfin.Drawing/ImageProcessor.cs
@@ -214,7 +214,16 @@ public sealed class ImageProcessor : IImageProcessor, IDisposable
                         return (originalImagePath, MimeTypes.GetMimeType(originalImagePath), dateModified);
                     }
 
-                    File.Move(tempFilePath, cacheFilePath, overwrite: true);
+                    try
+                    {
+                        File.Move(tempFilePath, cacheFilePath, overwrite: true);
+                    }
+                    catch (IOException) when (new FileInfo(cacheFilePath) is { Exists: true, Length: > 0 })
+                    {
+                        // Lost the publish race: a concurrent request already wrote this cache
+                        // entry and a reader holds it open (Windows denies replacing an open
+                        // file). The published entry is complete, so serve that one.
+                    }
                 }
                 finally
                 {
--- a/src/Jellyfin.Drawing/ImageProcessor.cs
+++ b/src/Jellyfin.Drawing/ImageProcessor.cs
@@ -218,11 +218,13 @@ public sealed class ImageProcessor : IImageProcessor, IDisposable
                     {
                         File.Move(tempFilePath, cacheFilePath, overwrite: true);
                     }
-                    catch (IOException) when (new FileInfo(cacheFilePath) is { Exists: true, Length: > 0 })
+                    catch (Exception ex) when ((ex is IOException or UnauthorizedAccessException)
+                        && new FileInfo(cacheFilePath) is { Exists: true, Length: > 0 })
                     {
                         // Lost the publish race: a concurrent request already wrote this cache
                         // entry and a reader holds it open (Windows denies replacing an open
-                        // file). The published entry is complete, so serve that one.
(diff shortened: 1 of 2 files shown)


Shadowghost (approved):

cvium:
Do we have any idea how many concurrent calls occur? This PR fixes the atomicity, but not the cache stampede, which is potentially a bigger problem on lower hardware.

LTe:
> Do we have any idea how many concurrent calls occur? This PR fixes the atomicity, but not the cache stampede, which is potentially a bigger problem on lower hardware.

This was one way of creating a issue. I can imagine that two clients will ask for the image in the same time. But I think this issue is more probable when you kill server in the middle of the processing of the image. Then you will left with empty file and poisoned cache.

PR #17422, opened 2026-07-23
Reduce correlated subqueries to improve query performance

Shadowghost:
**Changes**

* Rewrites several `BaseItemRepository` and query-helper hot paths that generated pathological SQL on large libraries.
* The nested correlated `.Any(...Any(...))` filters in `JellyfinQueryHelperExtensions` are flattened into two flat sub-selects plus a `Contains`, so a by-name filter (genres, studios, tags) becomes a set-based semi-join instead of a per-row nested scan.
* The by-name and Latest-items paths no longer use `GroupBy(key).Select(g => g.OrderBy(...).First())`, which EF translates into a correlated per-group subquery; they now materialize once and deduplicate in memory (or read albums directly by `TopParentId` when the query is library-scoped).
* `OneOrManyExpressionBuilder` always wraps its collection in `EF.Parameter`, so EF caches and reuses one compiled plan across calls with different id-sets instead of recompiling per distinct constant set, and every materialized id/value set now goes through `WhereOneOrMany` (single `json_each` parameter), which removes the SQLite variable-limit risk on high-cardinality value types.
* The `DtoService` inherited-image resolution for music albums reuses the batched artist lookup that `GetBaseItemDtos` already builds, instead of resolving the album artist once per album (an N+1).

**Code assistance**
Claude Code (Opus 4.8) was used to explain the EF Core translation of the affected LINQ shapes and to build and run the in-process benchmark harness that produced the numbers below.

**Performance**

Measured with an in-process benchmark harness  against one shared synthetic SQLite library of ~129,700 items (5,000 movies, 1,400 series with 3 seasons of 14 episodes, 2,450 artists / 2,580 albums with 14 tracks each, 6,500 studios, 8,000 tags, 300 genres, ~19,250 item values). Both variants ran against the identical seeded database. Numbers are the median of 3 samples with a cold EF query cache (reset before each sample), which is the case that dominates on large libraries. Lower is better.

| Scenario | master (s) | this branch (s) | Speedup |
| --- | ---: | ---: | ---: |
| GetArtists, unfiltered | 7.681 | 0.129 | ~60x |
| GetArtists, filtered by genre | 0.576 | 0.155 | ~3.7x |
| GetItems filtered by StudioIds | 71.191 | 0.061 | ~1170x |
| GetItems filtered by GenreIds | 3.284 | 0.065 | ~50x |
| GetItems filtered by TagIds | 0.072 | 0.081 | ~1.0x (unchanged) |
| GetItems filtered by IsFavorite | 0.061 | 0.064 | ~1.0x (unchanged) |
| GetStudios, unfiltered | 0.300 | 0.210 | ~1.4x |
| GetGenres, unfiltered | 0.070 | 0.073 | ~1.0x (unchanged) |
| Latest, movies | 0.137 | 0.099 | ~1.4x |
| Latest, music | 0.151 | 0.096 | ~1.6x |
| Latest, tvshows | 0.171 | 0.127 | ~1.3x |
| DtoService album inherited images, 30 albums | 0.144 | 0.099 | ~1.5x |
| OneOrMany, 50 calls with varied id-sets | 277.917 | 0.364 | ~763x |

Notes:

* The `StudioIds` and `GenreIds` rows are the flattened correlated-filter change. Studios show the largest gain because the seeded library has 6,500 studios, so the old nested `.Any(...Any(...))` degraded worst there.
* The unfiltered `GetArtists` row is the removal of `GroupBy(...).Select(g => g.OrderBy(...).First())`.
* `OneOrMany, 50 calls with varied id-sets` is the always-`EF.Parameter` change: master recompiles a plan per distinct constant set, this branch reuses one.
* The `DtoService album inherited images` row compares the old per-album `GetArtist(name)` lookup (one query per album, the N+1) against the batched `FindArtists(names)` lookup this branch reuses. The wall-clock gain is modest on fast local storage; the real win is collapsing 30 database round-trips into 1, which widens sharply on cold production disk (warm samples show 0.025 s versus 0.002 s here, roughly 12x).
* `TagIds`, `IsFavorite`, and unfiltered `GetGenres` are effectively unchanged on purpose. Single-level correlated filters are already handled well by SQLite, so those paths were left alone; the flattening only targets the nested and per-group forms that were pathological.
* Warm-cache numbers follow the same pattern. For example, `StudioIds` warm is ~71.4 s on master versus ~0.007 s here, and unfiltered `GetArtists` warm is ~7.4 s versus ~0.061 s.

**Issues**


Jellyfin.Server.Implementations/Item/BaseItemRepository.Querying.cs
```diff
@@ -126,38 +126,60 @@ public IReadOnlyList<BaseItemDto> GetLatestItemList(InternalItemsQuery filter, C
...
+                .OrderByDescending(e => e.DateCreated)
+                .ThenByDescending(e => e.Id)
+                .Select(e => new { e.Id, e.PresentationUniqueKey });
 
-            var firstIdsQuery = filter.Limit.HasValue
-                ? topGroupItems.Take(filter.Limit.Value).Select(g => g.FirstId)
-                : topGroupItems.Select(g => g.FirstId);
+            var seenKeys = new HashSet<string>();
+            var firstIds = new List<Guid>(limit ?? 0);
+            foreach (var row in orderedIds.AsEnumerable())
```
JPVenson: why not Distinct().Take(limit)

?

JPVenson (approved):

Jellyfin.Server.Implementations/Item/BaseItemRepository.Querying.cs
Shadowghost: We're deduplicating by `PresentationUniqueKey` here not by rows, so distinct won't fix the problem

crobibero (approved):

jhparker88:
@Shadowghost On a previous commit this PR included a benchmark harness, are you planning to bring those tests back with a separate PR as it would allow users to compare performance between releases?

Shadowghost:
I need to clean that up and would put it into a separate PR. Unintentionally pushed it to this PR

Jellyfin.Server.Implementations/Item/BaseItemRepository.ByName.cs
```diff
@@ -205,32 +200,42 @@ private string[] GetItemValueNames(IReadOnlyList<ItemValueType> itemValueTypes,
...
+        // representative id per group, then materialize the representative ids once.
         var masterQuery = TranslateQuery(innerQuery, context, outerQueryFilter);
         var isMusicArtist = returnType == _itemTypeLookup.BaseItemKindNames[BaseItemKind.MusicArtist];
-        var representativeIds = isMusicArtist
-            ? masterQuery
+        List<Guid> representativeIds;
+        if (isMusicArtist)
+        {
+            // For MusicArtist, prefer the entity from a library the user can actually access.
+            // Materilaize to prevent correlated per-group first-row queries which hurt performance.
```
cvium:
```suggestion
            // Materialize to prevent correlated per-group first-row queries which hurt performance.
```

Jellyfin.Server.Implementations/Item/BaseItemRepository.Querying.cs
cvium: Wouldn't [DistinctBy](https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.distinctby?view=net-10.0) achieve the same result?

```C#
orderedIds.AsEnumerable().DistinctBy(r => r.PresentationUniqueKey).Select(r => r.Id).Take(limit.Value);
```

cvium (approved):

Jellyfin.Server.Implementations/Item/BaseItemRepository.Querying.cs
Shadowghost: Yes, that actually works.

enter-a-random-username:
THX @Shadowghost
This seems to fix nearly all endpoints. Slowest thing are `favorites` `person` with 5-10 times slower compared to all others queries favorites.
Plus general favorite pages are slow compared to libraries with a filter for favorites. I don't use favorites a lot and empty once are slower as an item is for shows or movies.

Not listed in compare - Resume is 5x faster for me :heart:

PR #17424, opened 2026-07-23
Implement AudioDb artist search

### LINKED ISSUE #13980: identifying artist via theaudiodb.com ID is hard
### Description of the bug

Jellyfin 10.10.7 but this is a rather old issue that has existed in previous Jellyfin versions as well.

Identifying artist via theaudiodb.com ID is hard. Scenario is that an artist exists on theaudiodb.com but Jellyfin is not finding the artist image likely due to using the wrong artist when maybe multiple arists with the same name exist.

In order to address this mistake there is the option to `Identify` an artist. The dialog allows entering a `MusicBrainz Artist Id` or a `TheAudioDb Artist Id`. But the results are unexpected and difficult to understand.

### Reproduction steps

Example:

Trying to identify artist `Astro` (the house producer from Edinburgh) https://www.theaudiodb.com/artist/183973-Astro via TheAudioDb Artist Id `183973`

leads to the following results page:

1. it's surprosing to see "MusicBrainz" for each entry when trying to identify an artist using an `Theaudiodb ID`
2. no artist uses an image making it very hard to identify the correct artist
3. this becomes especially difficult since multiple arists with identical info exist - there is 16 artists called `astro` with info "MusicBrainz". For users it is impossible to know which artist to pick and know which the correct arist entry is
4. assuming the ID is unique, I'd expect a single entry to show or, even better, when an ID is matched pick that artist right away, without requiring additional user interaction as the matching artist could be identified already.
5. not quite sure what the year info represents in the Search Results

### Jellyfin Server version

10.10.0+

### Specify commit id

_No response_

_No response_

### Specify version number

10.10.7

### Specify the build version

10.10.7

### Environment

```markdown
- OS: Synology DSM 7.2.2-72806 Update 3
```

### Jellyfin logs

```shell
N/A
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

Shadowghost:
**Changes**
* Implement AudioDb artist search
* Fix MusicBrainz metadata

**Code assistance**
None

**Issues**
Fixes: https://github.com/jellyfin/jellyfin/issues/13980

### DIFF
--- a/MediaBrowser.Providers/Plugins/AudioDb/AudioDbArtistProvider.cs
+++ b/MediaBrowser.Providers/Plugins/AudioDb/AudioDbArtistProvider.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Net.Http;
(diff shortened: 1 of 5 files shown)


theguymadmax (approved):

MediaBrowser.Providers/Plugins/AudioDb/AudioDbArtistProvider.cs
```diff
@@ -52,45 +53,180 @@ public AudioDbArtistProvider(IServerConfigurationManager config, IFileSystem fil
...
+
+        private async Task<List<Artist>> FetchArtists(string url, CancellationToken cancellationToken)
+        {
+            using var response = await _httpClientFactory.CreateClient(NamedClient.Default).GetAsync(url, cancellationToken).ConfigureAwait(false);
+            response.EnsureSuccessStatusCode();
+
+            var jsonStream = await response.Content.ReadAsStreamAsync(cancellationToken).ConfigureAwait(false);
+            await using (jsonStream.ConfigureAwait(false))
+            {
+                var obj = await JsonSerializer.DeserializeAsync<RootObject>(jsonStream, _jsonOptions, cancellationToken).ConfigureAwait(false);
```
cvium: I believe this could also be used https://learn.microsoft.com/en-us/dotnet/api/system.net.http.json.httpcontentjsonextensions.readfromjsonasync?view=net-10.0#system-net-http-json-httpcontentjsonextensions-readfromjsonasync-1(system-net-http-httpcontent-system-text-json-jsonserializeroptions-system-threading-cancellationtoken)

cvium (approved):

crobibero (approved):

PR #15986, opened 2026-01-10
Fix N+1 queries and add missing indexes for large database performance

mtrogman:
This PR addresses multiple performance issues affecting users with large libraries in 10.11.x.

**Changes**
- Add missing database indexes on ItemValuesMap, AncestorId, UserData, BaseItemImageInfo, and PeopleBaseItemMap
- Replace multiple COUNT() calls with single grouped query in GetItemCounts
- Batch load ancestor IDs in SaveItems instead of per-item queries
- Add sort-aware includes to prevent N+1 queries when sorting by DatePlayed, PlayCount, etc.
- Use JOIN instead of correlated subquery in PeopleRepository ordering
- Replace unbounded ConcurrentDictionary with LRU cache in DirectoryService
- Use EF.Functions.Like() for case-insensitive searches to utilize indexes
- Add cascade delete for KeyframeData, MediaStreamInfo, BaseItemImageInfo
- Add configurable ContextPoolSize and TranscodingLockPoolSize options
- Add support for custom SQLite pragmas via pragmas.sql file
- Fix DbUpdateConcurrencyException on simultaneous logins (adds retry logic)

**New Configuration Options**

`database.xml` - ContextPoolSize:
```xml
<ContextPoolSize>1024</ContextPoolSize>
```
Sets the DbContext pool size for concurrent database operations. Higher values support more concurrent users. Default is 1024.

`encoding.xml` - TranscodingLockPoolSize:
```xml
<TranscodingLockPoolSize>20</TranscodingLockPoolSize>
```
Sets the pool size for transcoding locks. Higher values allow more concurrent transcoding operations. Default is 20.

`pragmas.sql` - Custom SQLite pragmas:

Place a `pragmas.sql` file in your config directory to apply custom SQLite pragmas on each database connection. The file is auto-discovered from `JELLYFIN_CONFIG_DIR` or `JELLYFIN_DATA_DIR`.

Alternatively, set pragmas via environment variables:
- `JELLYFIN_SQLITE_PRAGMAS` - semicolon-separated pragma statements
- `JELLYFIN_SQLITE_PRAGMAS_FILE` - path to a custom pragma file

Example `pragmas.sql`:
```sql
# Performance tuning for large databases
PRAGMA cache_size=-128000;
PRAGMA mmap_size=3221225472;
PRAGMA temp_store=MEMORY;
PRAGMA wal_autocheckpoint=2000;
```

**Bug Fixes**

[...]

PR #17440, opened 2026-07-25
Auto-scale default SQLite cache_size to fix slow ItemsByName endpoints

MaxRink:
**Changes**

The SQLite provider already exposes a `cacheSize` custom option that feeds `PragmaConnectionInterceptor`, but it defaults to `null`, so no `PRAGMA cache_size` is emitted and SQLite falls back to its 2 MiB default. On large libraries the ItemsByName queries (`/Artists`, `/Studios`, `/Genres`, `/AlbumArtists`) run a `ROW_NUMBER() ... PARTITION BY PresentationUniqueKey` window sort over an `IN (<subquery over the ~777k-row ItemValuesMap>)`, whose working set does not fit in a 2 MiB cache, so the connection thrashes and re-reads pages from disk on every request.

This PR gives `cacheSize` a default when the operator has not set one. It scales to ~1/16 of the memory available to the process (`GC.GetGCMemoryInfo().TotalAvailableMemoryBytes`, which is cgroup/container aware), clamped to [16 MiB, 128 MiB]. Small devices stay light, large installs get enough cache to hold the working set. `cache_size` only affects performance, never results, and operators can still override via the `cacheSize` option.

**Correction on the earlier page-size claim**

My original writeup said "the Jellyfin DB uses a 64 KiB page size, making that only ~32 pages". That is wrong as a general statement, @LTe pointed this out. A fresh 12.0-rc install creates a database with SQLite's default 4096 byte page size (verified). My own databases are 64 KiB, but that is specific to my install (created during a restore somewhere along a multi-year upgrade path), not something Jellyfin sets. Jellyfin does not set `page_size` anywhere in code, in any version. The fix does not depend on page size, so I re-ran everything on a 4 KiB database.

**Measurements**

All against a ~189k item library, PII-redacted copy, only HTTP 200 responses timed, `cacheSize=-262144` (256 MiB) for the "after" column.

4 KiB page size (the default, matching a normal install):

| endpoint | before (2 MiB default) | after |
|---|---|---|
| `/Studios` | ~16 s | **~3.2 s** |
| `/Artists` (favorites) | ~16 s | **~3.3 s** |
| `/Genres` | ~16 s | **~3.3 s** |

64 KiB page size (my install, kept for reference, this is the number I originally posted):

| endpoint | before | after |
|---|---|---|
| `/Studios` | ~35 s | ~2.1 s |
| `/Artists` (favorites) | ~35 s | ~2.2 s |
| `/Genres` | ~35 s | ~2.3 s |

So the effect is smaller on a stock 4 KiB DB than my original 64 KiB numbers implied, but it is still a large win (~5x). Page size shifts the magnitude, not the conclusion: the 2 MiB default is too small for a large library's working set regardless.

Cache-size sweep on the exact representative-id statement via `Microsoft.Data.Sqlite` against a cold DB copy, 4 KiB page size (cold / warm ms):

| cache_size | cold | warm |
|---|---|---|
| 2 MiB (default) | 127 | 117 |
| 8 MiB | 83 | 194 |
| 32 MiB | 75 | 15 |
| 128 MiB | 74 | 15 |
| 256 MiB | 72 | 17 |

The win flattens out by ~128 MiB, which is why the auto-scaled default clamps there.

Adds regression tests (`SqliteCacheSizeTests`) asserting the default is emitted (non-null, the actual regression), negative (interpreted as KiB), deterministic, and clamped to the sane range. Full solution builds clean and the tests pass locally (`dotnet test`, 3/3).

**Documentation**

Since the existing `cacheSize` / `#PRAGMA:` knobs (merged in 10.11) are undocumented, I opened jellyfin/jellyfin.org#1955 to document `database.xml` and cache_size guidance for large libraries, independent of whether this default change lands. Per @LTe's point in #15986, that covers the "leave it to the admin" path too.

**Code assistance**

Investigation, root-cause isolation (Stopwatch instrumentation plus a standalone `Microsoft.Data.Sqlite` probe to rule out EF translation and prove the page-cache cause), the cache-size sweep, benchmarking, and drafting were done with LLM assistance. All code, measurements, and the passing build/tests were reviewed and verified by me.

**Issues**

Refs #17405

### DIFF
--- a/src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/PragmaConnectionInterceptor.cs
+++ b/src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/PragmaConnectionInterceptor.cs
@@ -26,7 +26,14 @@ public class PragmaConnectionInterceptor : DbConnectionInterceptor
     /// Initializes a new instance of the <see cref="PragmaConnectionInterceptor"/> class.
     /// </summary>
     /// <param name="logger">The logger.</param>
-    /// <param name="cacheSize">Cache size.</param>
+    /// <param name="cacheSize">
+    /// The <see href="https://sqlite.org/pragma.html#pragma_cache_size">cache_size</see> pragma value.
+    /// A positive value is a number of pages; a negative value is a size in KiB (page-size independent).
+    /// When <see langword="null"/> no <c>PRAGMA cache_size</c> is emitted and SQLite keeps its 2 MiB
+    /// default. Operators can override it with the <c>cacheSize</c> custom provider option; when unset it
+    /// is auto-scaled to available memory by
+    /// <see cref="SqliteDatabaseProvider.GetDefaultCacheSize"/> (see jellyfin/jellyfin#17405).
+    /// </param>
     /// <param name="lockingMode">Locking mode.</param>
     /// <param name="journalSizeLimit">Journal Size.</param>
     /// <param name="tempStoreMode">The https://sqlite.org/pragma.html#pragma_temp_store pragma.</param>
--- a/src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteDatabaseProvider.cs
+++ b/src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/SqliteDatabaseProvider.cs
@@ -83,7 +83,7 @@ public sealed class SqliteDatabaseProvider : IJellyfinDatabaseProvider
                     .Ignore(RelationalEventId.MultipleCollectionIncludeWarning))
             .AddInterceptors(new PragmaConnectionInterceptor(
                 _logger,
-                GetOption<int?>(customOptions, "cacheSize", e => int.Parse(e, CultureInfo.InvariantCulture)),
+                GetOption<int?>(customOptions, "cacheSize", e => int.Parse(e, CultureInfo.InvariantCulture), () => GetDefaultCacheSize()),
                 GetOption(customOptions, "lockingmode", e => e, () => "NORMAL")!,
                 GetOption(customOptions, "journalsizelimit", int.Parse, () => 134_217_728),
                 GetOption(customOptions, "tempstoremode", int.Parse, () => 2),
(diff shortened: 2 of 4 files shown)


crobibero (changes requested):
Please review our LLM/AI policies.

MaxRink:
@crobibero That Policy has been observed; both the original Issue and PR Body are human-written but based upon LLM-aggregated data and output. ( For example, I explicitly outsourced all benchmark runs and data aggregation to an LLM).
Besides raw tables and data comparisons, everything has been touched, reviewed, and rewritten by me, to roughly the same standard as internal PRs and Issues at my Company

LTe:
> The Jellyfin DB uses a 64 KiB page size, making that only ~32 pages. 

Are you sure about that?

```
⏺ > sqlite3 -readonly /var/lib/jellyfin/config/data/jellyfin.db "PRAGMA page_size; PRAGMA cache_size;"
  4096
  -2000
```

MaxRink:
@LTe That's what my DB currently has:
```
sqlite3 -readonly jellyfin.db "PRAGMA page_size; PRAGMA cache_size;"

  # Production instance
  65536
  -2000

  # 189k-item benchmark copy
  65536
  -2000
```

A fresh DB does default to 4k, you're right. This install is from 2020 and has been upgraded through a lot of release candidates over the years, and i had to restore it once or twice along the way, so the 64k almost certainly came from me and not from Jellyfin. I took the value at face value in the writeup instead of checking it against a clean install, my bad.

I dug into both the where and the so-what:

Going back through my ZFS snapshots, the page size flipped mid-November last year, right after i upgraded to 10.11.x and was chasing some performance issues at the time. Looks like a leftover from a restore or dump/reload i did while poking at it. I also checked the source to be sure: Jellyfin never sets page_size in any version (`git grep page_size` is empty tree-wide, the interceptor only ever emits cache_size, locking_mode, journal_size_limit, synchronous and temp_store), so this was self-inflicted, not a default.

It doesn't change the conclusion though. I re-ran the benchmarks on a 4k copy: the broad ItemsByName endpoints are still slow with the default cache (~16s at 4k vs ~35s at 64k), and raising cacheSize fixes both the same way. Page size shifts the magnitude, but the 2 MiB default cache is just too small for the working set either way.

I opened jellyfin/jellyfin.org#1955 to at least make the existing flags discoverable. But i still think auto-scaling the default with an override is the better UX. Anyone who also keeps a large music library in Jellyfin hits the point where the default stops scaling pretty fast, and "the server's fine, you just had to know about an undocumented PRAGMA" isn't a great out-of-the-box experience.

LTe:
https://github.com/jellyfin/jellyfin/pull/15986#issuecomment-3764485801

There is already settings for this. In my opinion we should leave this to the server administrator and keep defaults without any config changes.

enter-a-random-username:
I tried this, and it feels like a memory leak.
The usage was 3-4 times (and rising) compared to before and normal browsing wasn't faster on current master branch with latest performance fixes applied.

PR #17442, opened 2026-07-25
Fix hyphenated numbers in episode titles parsed as multi-episodes

### LINKED ISSUE #17431: Episode Names with Hypens & Numbers are seen as Multi-Episodes
### Description of the bug

I have three instances of this in different series:

S01E01 The 6-10 to Lubbock [WEBRip-1080p][AV1 Opus]
S05E23 11-59 [HDTV-1080p][x265 AC3]
S01E01 1-23-45 [Bluray-1080p][AV1 Opus]

In each case, Sonarr renamed file names with time in them (colons) to use hyphens.  Import with Jellyfin sees these as multiple episodes, in the case of S01E01 1-23-45 [Bluray-1080p][AV1 Opus] it actually lists the episode in Jellyfin as 1-45 instead of episode 1.

### Reproduction steps

1. Add an episode with this type of name format.
2. Scan the library.

The episode is incorrectly reported as a single file containing multiple episodes.

S01E01 1-23-45 [Bluray-1080p][AV1 Opus] should be seen as S01E01.  Possibly the logic controlling this needs to stop looking for multipart episodes when it his the first [space].

### Jellyfin Server version

10.11.11

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.11

### Environment

```markdown
- OS: Ubuntu 24.04.4 LTS
- Virtualization: Proxmox
- Clients: Web
- Browser: Edge
- Playback Method: Web
- Jellyfin Data Storage & Filesystem: EXT4
- Media Storage & Filesystem: SMB
```

### Jellyfin logs

```shell
Unable to locate log with this.  I will attempt to remove an affected episode and reimport to get details.  I'll return and edit if successful.
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

Shadowghost:
The `Foo Bar 889` fallback expression matched filenames that already carry an SxxEyy marker, reading digits out of the title as an episode range: `S01E01 1-23-45 [Bluray-1080p]` became episodes 1 through 45 and `S01E01 The 6-10 to Lubbock` became episodes 1 through 10.
It never won the primary parse, the Kodi expression above it did, but reached these files through `FillAdditional`'s series-name recovery path, where it also set SeriesName to `S01E01`.

**Changes**
Exclude those names with the same marker pattern the Kodi expression uses, so the fallback steps aside exactly when that rule applies.

**Code assistance**
None

**Issues**
Fixes #17431

### DIFF
--- a/Emby.Naming/Common/NamingOptions.cs
+++ b/Emby.Naming/Common/NamingOptions.cs
@@ -361,7 +361,10 @@ namespace Emby.Naming.Common
                 // Not a Kodi rule as well, but the expression below also causes false positives,
                 // so we make sure this one gets tested first.
                 // "Foo Bar 889"
-                new EpisodeExpression(@".*[\\\/](?![Ee]pisode)(?<seriesname>[\w\s]+?)\s(?<epnumber>[0-9]{1,4})(-(?<endingepnumber>[0-9]{2,4}))*[^\\\/x]*$")
+                // Names carrying an SxxEyy marker are excluded because the Kodi expression above already covers them.
+                // Without that guard this expression reads digits out of the title instead, turning
+                // "S01E01 1-23-45 [Bluray-1080p]" into episodes 1 through 45.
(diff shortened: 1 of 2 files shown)


crobibero (approved):

cvium (approved):
I can't really be bothered to read the regex changes on mobile but I trust the tests.

PR #17428, opened 2026-07-24
fix: stop Clean Cache Directory task from deleting metadata stored inside the cache path

mvanhorn:
## Summary

Makes the "Clean Cache Directory" scheduled task (`DeleteCacheFileTask`) metadata-aware so it never deletes files that live under the configured metadata directory. The task now resolves the effective metadata path through `IServerApplicationPaths.InternalMetadataPath` and, when that directory is the same as or nested inside `CachePath`, excludes every file beneath it from the cache-cleanup pass. The containment check normalizes both paths with `Path.GetFullPath`, compares case-insensitively on Windows and ordinally elsewhere, and requires a trailing directory separator, so a sibling such as `metadata2` next to `metadata` is not protected. A single warning is logged when the metadata directory is detected inside the cache directory. The short-lived `TempDirectory` pass is left unchanged, so genuinely stale cache files are still removed exactly as before.

## Background

`DeleteCacheFileTask` recursively deletes every file older than 30 days under `CachePath`. When a server's metadata directory is configured as the cache directory or a child of it, the task also deletes stored metadata images such as posters, logos, and backdrops; those images only reappear after a full library refresh, which is the root cause behind reports of artwork randomly disappearing over days or weeks. The existing top-level marker system does not protect a metadata folder nested inside the cache folder, so a metadata path that resolves inside the cache path needs to be excluded explicitly. This change prevents the data loss without altering cleanup behavior for genuinely stale cache files.

Fixes #15849

PR #17447, opened 2026-07-25
fix: reject a metadata path nested inside another Jellyfin directory

mvanhorn:
**Changes**

`MakeSanityCheckOrThrow` now validates and markers `InternalMetadataPath`, rejecting a metadata path that resolves inside the configuration, log, plugin, cache, data, or root directory. `ServerConfigurationManager.ValidateMetadataPath` runs the same check so the bad state cannot be created through the UI or API, and `UpdateMetadataPath` establishes the marker on reassignment the way `BaseConfigurationManager.UpdateCachePath` already does for the cache path. Containment comparison normalizes both paths with `Path.GetFullPath`, trims trailing separators, and uses `OrdinalIgnoreCase` only on Windows.

This replaces #17428, which tried to make the Clean Cache Directory task tolerate a nested metadata directory. Per @cvium, that configuration is not supported and the right fix is to validate and refuse rather than accommodate it. #12832 already established the marker-file mechanism; the gap was that `InternalMetadataPath` was never included in the check, so this extends that mechanism rather than adding a parallel one.

**Code assistance**

Written with LLM assistance. The approach was chosen by me from @cvium's review of #17428 and the existing #12832 mechanism; the LLM was used to write the implementation and tests against that decision, and I reviewed the output and can explain any part of it. Testing was run locally, not assumed.

**Testing**

8 new tests in `ServerConfigurationManagerTests`: nested-inside-cache rejected at startup; nested inside cache / data / config rejected via `ReplaceConfiguration`; sibling path accepted and markered; trailing-separator and case-difference variants behave as the plain form; default metadata path creates its marker.

Full `Jellyfin.Server.Implementations.Tests` project: 676 passed, 0 failed, 5 skipped (pre-existing Windows-only skips).

**Issues**

Replaces #17428. Reference implementation: #12832.

### DIFF
--- a/Emby.Server.Implementations/Configuration/ServerConfigurationManager.cs
+++ b/Emby.Server.Implementations/Configuration/ServerConfigurationManager.cs
@@ -76,13 +76,14 @@ namespace Emby.Server.Implementations.Configuration
             ((ServerApplicationPaths)ApplicationPaths).InternalMetadataPath = string.IsNullOrWhiteSpace(Configuration.MetadataPath)
                 ? ApplicationPaths.DefaultInternalMetadataPath
                 : Configuration.MetadataPath;
-            Directory.CreateDirectory(ApplicationPaths.InternalMetadataPath);
+            ApplicationPaths.CreateAndCheckMarker(ApplicationPaths.InternalMetadataPath, "metadata");
         }
 
         /// <summary>
         /// Replaces the configuration.
         /// </summary>
         /// <param name="newConfiguration">The new configuration.</param>
+        /// <exception cref="ArgumentException">If the metadata path is nested inside another Jellyfin directory.</exception>
         /// <exception cref="DirectoryNotFoundException">If the configuration path doesn't exist.</exception>
         public override void ReplaceConfiguration(BaseApplicationConfiguration newConfiguration)
         {
@@ -99,6 +100,7 @@ namespace Emby.Server.Implementations.Configuration
         /// Validates the metadata path.
         /// </summary>
         /// <param name="newConfig">The new configuration.</param>
+        /// <exception cref="ArgumentException">The new metadata path is nested inside another Jellyfin directory.</exception>
         /// <exception cref="DirectoryNotFoundException">The new config path doesn't exist.</exception>
         private void ValidateMetadataPath(ServerConfiguration newConfig)
         {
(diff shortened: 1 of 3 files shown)


cvium:
Please review our [LLM policies](https://jellyfin.org/docs/general/contributing/llm-policies/), and please follow our PR template

mvanhorn:
Both fair. I've rewritten the description using the template, including the Code assistance section.

To answer it directly here too: this was written with LLM assistance, and the reasoning is mine to defend. #17428 tried to accommodate the nested layout; your response was that Jellyfin doesn't support that configuration and the right fix is to validate and refuse. #12832 already established the marker-file mechanism via `MakeSanityCheckOrThrow` -- the gap was that `InternalMetadataPath` was never included in it. This extends that check rather than adding a parallel one, and mirrors `UpdateCachePath` for the reassignment case.

If the approach is still wrong, I'd rather hear it now than iterate.

nielsvanvelzen:
Just saying that LLM assistance was used is not meaningfully disclosing AI usage. What model, and in what way was it used? Is your entire account an agent, or did you just use an autocomplete model in your IDE?

mvanhorn:
I'm a human who works with a lot of agents. For this PR: Claude Opus 5, through Claude Code, not IDE autocomplete.

Concretely: I chose the approach from @cvium's review of #17428 plus the existing `MakeSanityCheckOrThrow` mechanism in #12832. The model wrote the implementation and the 8 tests against that decision. I read the diff before it went up and ran the suite locally (676 passed, 0 failed, 5 pre-existing Windows-only skips). Any line in it, I can explain and defend.

If the policy wants a specific disclosure format in the description itself, tell me the shape you want and I will put it there.

theguymadmax:
Do we really want to disallow the metadata folder inside the data folder? I feel that might be a common occurrence. Here is a recent example of such a case: https://forum.jellyfin.org/t-moved-metadata

mvanhorn:
Fair challenge, and worth settling before this goes further, because it cuts against what #17428 concluded.

The take there was that nesting the folders like this isn't wanted, and that the right fix is to validate the paths and refuse rather than accommodate it. This PR is that validation. If nesting metadata inside the data folder is actually common in practice, then the validation is too strict and the forum case you linked is the counter-example.

Which is it? If the answer is that nested-inside-data should be allowed while other nesting still gets rejected, say so and I'll rescope. I don't think I should be the one picking between you and @cvium on it.

cvium:
I might be way off base here, and I know that people can be eloquent and verbose without AI, but every comment you write just gives me heavy LLM vibes. We not only ask that people disclose their AI usage, but we also require that any community facing communication is wholly and solely written by a human. There are exceptions to this rule, but it needs to be disclosed.

We get a lot of AI generated PRs and a lot of those made by bots or people copy/pasting the LLM output. As a maintainer this is incredibly frustrating and disrespectful as it wastes our time and energy.

As for your PR: Nesting these directories is mostly fine. It is only the cache and transcode directories (cache-like directories) where it shouldn't be allowed as we delete everything inside those folders on a regular basis.

mvanhorn:
Pushed faba53e. Narrowed the check to the cache and transcode directories only, per your note that the rest is fine.

ValidateMetadataPathOrThrow now takes the transcode path from EncodingOptions.TranscodingTempPath rather than assuming a location, and the thrown message names the reason: those two get cleared on a schedule. Config, log, plugin, data and root no longer reject, so the moved-metadata-into-data layout from the forum thread works again.

Tests flipped to match: data and config nesting now assert acceptance, transcode joins cache as a rejection case. 9/9 passing locally.

cvium:
I am closing this as the author has failed to convince me they're not a bot 🫠 The implementation is also very wrong. We use physical files as markers for these directories since docker containers can mount these directories in weird ways, where comparing paths won't work.

PR #17450, opened 2026-07-26
fix(subtitles): skip duplicate download when file already exists

### LINKED ISSUE #17426: Opensubtitles plugin continually downloads duplicate srt files
### Description of the bug

The Opensubtitles plugin downloads duplicate copies of the same subtitle files. Ref: [#12424](https://github.com/jellyfin/jellyfin/issues/12434) The previous ticket was closed.

I have tested on 10.11.11 using Opensubtitles plugin 24.0.0.0 .
I have also tested on 12.0 RC 3 using Open subtitles plugin 24.2605.18.0 .
The duplicate subtitle files are named as such: *.#.srt . The "#" increments for each new copy that is downloded.

### Reproduction steps

Completely removed Jellyfin from my Windows system(C:\Program files\Jellyfin\ & C:\Program data\Jellyfin).
Installed 12.0 RC 3.
Set the plugin repo to: https://nyc1.mirror.jellyfin.org/files/plugin-unstable/manifest.json
Restarted Jellyfin.
Installed Open Subtitles 24.2605.18.0 .
Restarted Jellyfin.
Logged into the plugin using my valid opensubtitles.com account.
Restarted Jellyfin.
Made sure I have a download subtitles enabled and a few languages selected for each of my libraries.
Restarted Jellyfin.
Performed a full media scan.
Manually ran the scheduled task for downloading subtitles.

Subtitle *.srt files are downloaded each time the plugin runs regardless on if the file has been previously downloaded. There is an ever increasing count of *.#.srt copies of the same subtitle file.

Skip downloading a specific subtitle file if it already exists in the media library.

### Jellyfin Server version

10.11.11

### Specify commit id

_No response_

12.0 RC3

### Specify version number

_No response_

### Specify the build version

12.0-rc3

### Environment

```markdown
- OS: Windows 10
- Linux Kernel:
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- CPU Model:
- GPU Model:
- Plugins: Opensubtitles
- Reverse Proxy:
- Base URL:
- Networking:
- Jellyfin Data Storage & Filesystem: local HDD - NTFS
- Media Storage & Filesystem: Unraid CIFS share
- External Integrations:
```

### Jellyfin logs

```shell
[2026-07-23 14:16:48.358 -05:00] [INF] [24] Jellyfin.Plugin.OpenSubtitles.OpenSubtitleDownloader: Remaining subtitle downloads: 996
[2026-07-23 14:16:48.909 -05:00] [INF] [9] MediaBrowser.Providers.Subtitles.SubtitleManager: Saving subtitles to "P:\Movies\#_####.eng.srt"
[2026-07-23 14:16:49.597 -05:00] [INF] [24] Jellyfin.Plugin.OpenSubtitles.OpenSubtitleDownloader: Remaining subtitle downloads: 995
[2026-07-23 14:16:50.111 -05:00] [INF] [12] MediaBrowser.Providers.Subtitles.SubtitleManager: Saving subtitles to "P:\Movies\#_####.spa.srt"
[2026-07-23 14:16:50.339 -05:00] [INF] [24] Jellyfin.Plugin.OpenSubtitles.OpenSubtitleDownloader: Remaining subtitle downloads: 995
[2026-07-23 14:16:50.373 -05:00] [INF] [12] MediaBrowser.Providers.Subtitles.SubtitleManager: Saving subtitles to "P:\Movies\#_####.spa.0.srt"

[2026-07-23 14:27:59.681 -05:00] [INF] [29] MediaBrowser.Providers.Subtitles.SubtitleManager: Saving subtitles to "P:\Movies\###_####_###_#######_#-#\#.####.#.#######.####.######.#####.###-##.####-####.(@#####)\#_#_#_#######_####.eng.0.srt"
[2026-07-23 14:27:59.906 -05:00] [INF] [29] Jellyfin.Plugin.OpenSubtitles.OpenSubtitleDownloader: Remaining subtitle downloads: 995
[2026-07-23 14:27:59.952 -05:00] [INF] [29] MediaBrowser.Providers.Subtitles.SubtitleManager: Saving subtitles to "P:\Movies\###_####_###_#######_#-#\#.####.#.#######.####.######.#####.###-##.####-####.(@#####)\#_#_#_#######_####.spa.0.srt"
[2026-07-23 14:28:00.224 -05:00] [INF] [26] Jellyfin.Plugin.OpenSubtitles.OpenSubtitleDownloader: Remaining subtitle downloads: 995
[2026-07-23 14:28:01.266 -05:00] [INF] [29] MediaBrowser.Providers.Subtitles.SubtitleManager: Saving subtitles to "P:\Movies\###_####_###_#######_#-#\#.####.#.#######.####.######.#####.###-##.####-####.(@#####)\#_#_#_#######_####.spa.1.srt"
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

sergioperezcheco:
The "Download missing subtitles" scheduled task (and manual downloads via the API) were accumulating unbounded duplicate files — movie.en.srt, movie.en.0.srt, movie.en.1.srt, and so on — on every run, as reported in #17426.

Root cause is in SubtitleManager.TrySaveToFiles: when the target subtitle path already existed, the code incremented a counter to find the next free filename rather than treating the existing file as the intended destination. Because the scheduled task re-runs daily and the metadata refresh that populates MediaStreams can lag behind the filesystem, the same subtitle kept getting re-downloaded into a new numbered copy each time.

This replaces the counter loop with an early skip: if the file is already present at the computed path, we log it and return without writing anything. This matches the user expectation expressed in the issue ("Skip downloading a specific subtitle file if it already exists") and stops the duplicate accumulation. FileMode.CreateNew already prevents accidental overwrites on the first write.

Fixes #17426.

### DIFF
--- a/MediaBrowser.Providers/MediaInfo/FFProbeVideoInfo.cs
+++ b/MediaBrowser.Providers/MediaInfo/FFProbeVideoInfo.cs
@@ -553,7 +553,8 @@ namespace MediaBrowser.Providers.MediaInfo
             {
                 var downloadedLanguages = await new SubtitleDownloader(
                     _logger,
-                    _subtitleManager).DownloadSubtitles(
+                    _subtitleManager,
+                    _localization).DownloadSubtitles(
                         video,
                         currentStreams.Concat(externalSubtitleStreams).ToList(),
                         libraryOptions.SkipSubtitlesIfEmbeddedSubtitlesPresent,
--- a/MediaBrowser.Providers/MediaInfo/SubtitleDownloader.cs
+++ b/MediaBrowser.Providers/MediaInfo/SubtitleDownloader.cs
@@ -14,6 +14,7 @@ using MediaBrowser.Controller.Entities.TV;
 using MediaBrowser.Controller.Providers;
 using MediaBrowser.Controller.Subtitles;
 using MediaBrowser.Model.Entities;
+using MediaBrowser.Model.Globalization;
 using Microsoft.Extensions.Logging;
 
 namespace MediaBrowser.Providers.MediaInfo
@@ -22,11 +23,36 @@ namespace MediaBrowser.Providers.MediaInfo
     {
         private readonly ILogger _logger;
         private readonly ISubtitleManager _subtitleManager;
+        private readonly ILocalizationManager _localization;
 
-        public SubtitleDownloader(ILogger logger, ISubtitleManager subtitleManager)
+        public SubtitleDownloader(ILogger logger, ISubtitleManager subtitleManager, ILocalizationManager localization)
         {
             _logger = logger;
             _subtitleManager = subtitleManager;
+            _localization = localization;
+        }
+
+        /// <summary>
+        /// Normalizes an ISO language code (2-letter, 3-letter, or full name) to the
+        /// canonical <see cref="CultureDto.Name"/> so two codes referring to the same
+        /// language compare equal. Returns the input unchanged when no culture matches.
+        /// </summary>
+        /// <remarks>
+        /// Media streams store the language detected by FFprobe via <c>ExternalPathParser</c>,
+        /// which prefers the ISO 639-3 code (e.g. "eng"), while library options and subtitle
+        /// requests commonly use ISO 639-2 codes (e.g. "en"). A plain ordinal compare lets
+        /// an existing subtitle slip through the pre-download guard, causing a redundant
+        /// download that consumes the OpenSubtitles daily quota and re-creates the file.
+        /// </remarks>
+        private string NormalizeLanguage(string language)
+        {
+            if (string.IsNullOrEmpty(language))
+            {
+                return string.Empty;
+            }
+
+            var culture = _localization.FindLanguageInfo(language);
+            return culture?.Name ?? language;
         }
 
         public async Task<List<string>> DownloadSubtitles(
@@ -131,8 +157,13 @@ namespace MediaBrowser.Providers.MediaInfo
             bool isAutomated,
             CancellationToken cancellationToken)
         {
+            // Normalize the requested language so a 2-letter/3-letter code mismatch
+            // (e.g. requested "en" vs. an existing stream stored as "eng") doesn't
+            // bypass these guards and trigger a redundant download.
+            var normalizedLanguage = NormalizeLanguage(language);
+
             // There's already subtitles for this language
-            if (mediaStreams.Any(i => i.Type == MediaStreamType.Subtitle && i.IsTextSubtitleStream && string.Equals(i.Language, language, StringComparison.OrdinalIgnoreCase)))
+            if (mediaStreams.Any(i => i.Type == MediaStreamType.Subtitle && i.IsTextSubtitleStream && string.Equals(NormalizeLanguage(i.Language), normalizedLanguage, StringComparison.OrdinalIgnoreCase)))
             {
                 return false;
             }
@@ -148,14 +179,14 @@ namespace MediaBrowser.Providers.MediaInfo
 
             // There's already a default audio stream for this language
             if (skipIfAudioTrackMatches &&
-                defaultAudioStreams.Any(i => string.Equals(i.Language, language, StringComparison.OrdinalIgnoreCase)))
+                defaultAudioStreams.Any(i => string.Equals(NormalizeLanguage(i.Language), normalizedLanguage, StringComparison.OrdinalIgnoreCase)))
             {
                 return false;
             }
 
             // There's an internal subtitle stream for this language
             if (skipIfEmbeddedSubtitlesPresent &&
-                mediaStreams.Any(i => i.Type == MediaStreamType.Subtitle && !i.IsExternal && string.Equals(i.Language, language, StringComparison.OrdinalIgnoreCase)))
+                mediaStreams.Any(i => i.Type == MediaStreamType.Subtitle && !i.IsExternal && string.Equals(NormalizeLanguage(i.Language), normalizedLanguage, StringComparison.OrdinalIgnoreCase)))
             {
                 return false;
             }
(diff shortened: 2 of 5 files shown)


cvium:
People may not necessarily want to overwrite existing subtitle files. Isn't the real problem that the scheduled task fails to check if the language exists already?

sergioperezcheco:
Fair point — overwriting silently would hide refresh issues and users may intentionally want to replace a bad subtitle. The unbounded counter was the worst symptom though: even when the scheduled task's `HasNoExternalSubtitleTrackWithLanguage` filter and the `SubtitleDownloader` mediaStreams check both pass, the duplicate accumulation still happens in practice (issue #17426 reproduces on 10.11 and 12.0 RC). Tracing it, the likely upstream gap is a language-code mismatch: `SubtitleManager.TrySaveSubtitle` builds the filename from `response.Language.ToLowerInvariant()` (the value the OpenSubtitles plugin returns), but `ExternalPathParser` stores the detected language as `ThreeLetterISOLanguageName`. So when the plugin returns a 2-letter code but the resolver stores "eng"/"fre", neither the task query (`ms.Language == lang`) nor the `SubtitleDownloader` mediaStreams check matches, and the counter keeps appending.

Happy to take one of two directions here, whichever you prefer:

1. Keep the save-path guard but make it non-destructive — if a file already exists at the target path, skip instead of overwrite (current change), since that is the direct fix for the ever-growing `*.0.srt`, `*.1.srt`, ... reported in the issue.
2. Replace this with an upstream fix that normalizes the language code before the mediaStreams comparison in `SubtitleDownloader.DownloadSubtitles` (line 135) and the task query, so existing external subtitles are actually detected and the download is skipped before reaching `TrySaveToFiles`.

Option 2 is more correct but a larger change touching query translation; option 1 is the minimal stop against the reported regression. Let me know which you want and I'll push it.

cvium:
Please review our [LLM policies](https://jellyfin.org/docs/general/contributing/llm-policies/).

gabeluci:
Is this before or after downloading the subtitle?
Free OpenSubtitles accounts have a per-day download limit, so if it downloads the file and then decides it's not needed, it would eat up the download limit unnecessarily (possibly repeatedly if it does the same thing the next day).

sergioperezcheco:
Good catch — yes, this guard sits in TrySaveToFiles, which runs after GetRemoteSubtitles already downloaded the file, so it does burn one download against the daily limit. The pre-download check in SubtitleDownloader (line 135) should short-circuit before that, but it compares mediaStreams[].Language against the requested language with a plain ordinal equals, so a 2-letter/3letter mismatch (e.g. the plugin returns "en" but FFprobe stores the existing file as "eng") lets it slip through and reach the download. I will rework this to normalize the language code before that pre-download comparison so the skip happens before any quota is consumed, rather than catching it at save time.

sergioperezcheco:
@cvium @gabeluci Pushed the rework discussed above. The save-time skip stays (it is the direct stop against the unbounded *.0.srt/*.1.srt files), and I added the upstream fix so the daily-quota concern is addressed too.

The scheduled task and the refresh path both route through SubtitleDownloader.DownloadSubtitles, which has a pre-download guard that should short-circuit before any search happens:

    if (mediaStreams.Any(i => ... && string.Equals(i.Language, language, StringComparison.OrdinalIgnoreCase)))
        return false;

The reason it never fired is a language-code mismatch: FFprobe/ExternalPathParser stores the detected language as the 3-letter ISO 639-3 code ("eng", "fre", ...) while the library option / request is commonly a 2-letter ISO 639-2 code ("en", "fr"). A plain ordinal equals on "eng" vs "en" returns false, the existing external subtitle is treated as absent, and every run walks straight into SearchSubtitles and downloads again.

Fix: inject ILocalizationManager into SubtitleDownloader and resolve both the requested and the stored language to their canonical CultureDto.Name before each of the subtitle/audio/embedded guards, so the codes compare on a normalized form. The skip now happens before the search, so it does not consume an OpenSubtitles download.

Added a unit test (SubtitleDownloaderTests) covering the 2-letter-request-vs-3-letter-stream and the reverse cases; the full Jellyfin.Providers.Tests run is green locally (307 passed). Let me know if you would rather the normalization live in a helper on ILocalizationManager itself rather than a private method on SubtitleDownloader, happy to move it.

sergioperezcheco:
Good question — the File.Exists check in TrySaveToFiles runs after GetRemoteSubtitles has already fetched the bytes, so by itself it does not protect the daily quota. The quota protection for the scheduled task lives one level higher, in SubtitleDownloader.DownloadSubtitles: the guard at the mediaStreams.Any(...) check (around line 166) runs before SearchSubtitles/DownloadSubtitles are ever called, so an already-indexed external subtitle skips the network round-trip entirely.

The gap that issue #17426 surfaces is when MediaStreams lags behind the filesystem — the subtitle file is on disk but the metadata refresh has not picked it up yet, so the line 166 guard misses, the download proceeds, and the TrySaveToFiles check then discards the already-fetched result. That stops the movie.en.0/1/2.srt pile-up (the actual user-reported symptom) but yes, it burns one download in that edge case.

The second commit on this branch tries to narrow that gap from the other side: FFprobe stores the language as the ISO 639-3 code (eng) while library options and requests commonly use the 2-letter code (en), and the old ordinal compare let an existing stream slip past the line 166 guard even when metadata was current. Normalizing both sides through ILocalizationManager.FindLanguageInfo before comparing closes that mismatch path, so fewer requests reach the download stage in the first place.

If you would prefer the file-existence check to also gate the network call, the complication is that the save path is derived from response properties (Language, IsForced, IsHearingImpaired, Format) that are only known after GetRemoteSubtitles returns. I can look at computing the candidate path from the requested language before the fetch and short-circuiting there, but it touches the SubtitleManager.DownloadSubtitles(video, subtitleId) signature, so I wanted to flag the trade-off before going further.

cvium:
Closing this as the author failed to comply with our [LLM policies](https://jellyfin.org/docs/general/contributing/llm-policies/).

PR #17451, opened 2026-07-26
fix(encoding): honor AudioStreamIndex for audio requests on video items

sergioperezcheco:
When requesting an audio stream from a video item via /Audio/{itemId}/stream?audioStreamIndex=N, the server always returned the first audio track regardless of the requested index. Both requests produced identical output and identical FFmpeg commands with no -map selection.

The cause is in EncodingHelper.AttachMediaSourceInfo: the non-video-request branch (used by the audio endpoints) hardcoded null as the desired index when selecting the audio stream, instead of forwarding the AudioStreamIndex from the request. This meant GetMediaStream fell through to its default first-track selection and the caller's index was silently dropped.

This passes state.BaseRequest.AudioStreamIndex so the requested track is actually honored on the transcode path, mirroring what the video-request branch already does.

### DIFF
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -7502,7 +7502,7 @@ namespace MediaBrowser.Controller.MediaEncoding
             }
             else
             {
-                state.AudioStream = GetMediaStream(mediaStreams, null, MediaStreamType.Audio, true);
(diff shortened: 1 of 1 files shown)


gnattu:
This is not the intended way of using the audio api though and it is not guaranteed to work at all if the input source is not a video.

sergioperezcheco:
You're right, thanks for the correction. Honoring AudioStreamIndex through the video playback path only works when the source is actually a video file with embedded streams, and it breaks for standalone audio. I'll close this rather than force a workaround that won't generalize. If there's a preferred place to handle per-track audio selection for audio items, happy to take another crack at it properly.

PR #17456, opened 2026-07-27
Fix extras naming and version assignment

Shadowghost:
**Changes**
* Extras are now named after what they are: "Trailer", "Behind The Scenes", numbered when a type repeats, instead of inheriting the owner's file name
* `TryGetExtraTypeForOwner` hands back the rule it matched so an extra carrying a real title of its own ("Recording the audio-behindthescenes") keeps it, and a corrected name is persisted on the next refresh.
* Remote metadata and image providers no longer run for extras, since an extra has no identity of its own
* For an item with several versions, an extra named after a specific version file is assigned to that version, while folder-based extras and extras named after the item as a whole stay with the primary
  * Every version surfaces all of them, and the `HasTrailer` / `HasSpecialFeature` / `HasThemeSong` / `HasThemeVideo` filters match on any version.
  * Alternate versions and stacked parts are owned items but not extras, so they are no longer counted as such when deciding whether an item's extras changed.

**Code assistance**
Claude Opus 5 was used to trace the issue and write the tests.

**Issues**


crobibero:
How important is this fix? I think it is a bit late to be adding translations

Shadowghost:
Not that important, but since we are re-generating all Extras on the frist scan anyways, it would be a nice to have, otherwise we'll likely need to do another migration in 13.x

crobibero (approved):

theguymadmax (approved):

ScarletCarson:
This may have introduced a regression in bracketed version handling.

Up to 12 these worked:

├── Movie (2005) [imdbid-ttxx]
    ├── Movie (2005) [imdbid-ttxx] - [Original][1080p].mkv
    ├── Movie (2005) [imdbid-ttxx] - [Original][480p].mkv
    └── Movie (2005) [imdbid-ttxx] - [Original].mkv

After updating to 12.0-rc5 (to pick up the schedule-direct fix), these movies are still collpased into one title but the versions are missing and only one arbitrary selection is left.  I've validated backwards to [1080p] and [480p] and they are picked up again.

PR #17460, opened 2026-07-27
Queue person metadata refresh instead of blocking the item request and fix ItemCounts

### LINKED ISSUE #17457: Fast `/Users/<UUID>/Items/<UUID>` enpoint, except person
### Description of the bug

As the title states the endpoint `/Users/<UUID>/Items/<UUID>`  is really slow, compared to everything else, especially after recently merged performance increase from @Shadowghost the difference is massive.

The returned payload is small, (but has counts of all related entities).

_(Search is also the slowest entity even with no match)_

### Reproduction steps

1. Go to an episode or series
2. Click ... on hover while hovering a person and wait

Slow answer ~20x slower compared to anything else on same endpoint

Similar speed would be nice.

### Jellyfin Server version

Master

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

Release

### Environment

```markdown
- OS: Linux
- Linux Kernel:
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- CPU Model:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Jellyfin Data Storage & Filesystem: NVMe
- Media Storage & Filesystem:
- External Integrations:
```

### Jellyfin logs

```shell
-
```

### FFmpeg logs

```shell

```

### Client / Browser logs

-

Endpoint movies popover menu
Image

Endpoint person

Image

### Additional information

My guess is counting all relations is an issue in menu (and the search)

Image

According to api.jellyfin.org, the endpoint has not a lot of options, so recursive or so could be dropped.

(In general, no clue why opening a menu needs a quest, my guess was entries based on permission)

Shadowghost:
**Changes**
* Queue the on-demand person metadata refresh instead of awaiting it.
* Drive the by-name `ItemCounts` query off an item-id sub-select instead of joining the mapping table into the count.
* Stop counting an item more than once when it is mapped more than once. An actor who also directed a film has two `PeopleBaseItemMap` rows for it, and the old join counted that film twice.

**Code assistance**
Claude Opus 5 was used for the tests

**Issues**
Fixes #17457

### DIFF
--- a/Jellyfin.Api/Controllers/UserLibraryController.cs
+++ b/Jellyfin.Api/Controllers/UserLibraryController.cs
@@ -40,6 +40,7 @@ public class UserLibraryController : BaseJellyfinApiController
     private readonly IDtoService _dtoService;
     private readonly IUserViewManager _userViewManager;
     private readonly IFileSystem _fileSystem;
+    private readonly IProviderManager _providerManager;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="UserLibraryController"/> class.
@@ -50,13 +51,15 @@ public class UserLibraryController : BaseJellyfinApiController
     /// <param name="dtoService">Instance of the <see cref="IDtoService"/> interface.</param>
     /// <param name="userViewManager">Instance of the <see cref="IUserViewManager"/> interface.</param>
     /// <param name="fileSystem">Instance of the <see cref="IFileSystem"/> interface.</param>
+    /// <param name="providerManager">Instance of the <see cref="IProviderManager"/> interface.</param>
     public UserLibraryController(
         IUserManager userManager,
         IUserDataManager userDataRepository,
         ILibraryManager libraryManager,
         IDtoService dtoService,
         IUserViewManager userViewManager,
-        IFileSystem fileSystem)
+        IFileSystem fileSystem,
+        IProviderManager providerManager)
     {
         _userManager = userManager;
         _userDataRepository = userDataRepository;
@@ -64,6 +67,7 @@ public class UserLibraryController : BaseJellyfinApiController
         _dtoService = dtoService;
         _userViewManager = userViewManager;
         _fileSystem = fileSystem;
+        _providerManager = providerManager;
     }
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


crobibero (approved):

Jellyfin.Api/Controllers/UserLibraryController.cs
```diff
@@ -94,7 +98,7 @@ public async Task<ActionResult<BaseItemDto>> GetItem(
             return NotFound();
         }
 
-        await RefreshItemOnDemandIfNeeded(item).ConfigureAwait(false);
+        QueueRefreshOnDemandIfNeeded(item);
```
IDisposable, line 101: Doesn't this change mean that the DTO returned from `GettBaseItemDto` is essentially guaranteed to be the unchanged-yet existing data? Seem that make the return value of the `GetItem` pretty useless. We should likely set the right expectation by _not_ returning a DTO from this method, since it's a "before" state.

Jellyfin.Api/Controllers/UserLibraryController.cs
```diff
@@ -112,7 +116,7 @@ public async Task<ActionResult<BaseItemDto>> GetItem(
     [ProducesResponseType(StatusCodes.Status200OK)]
     [Obsolete("Kept for backwards compatibility")]
     [ApiExplorerSettings(IgnoreApi = true)]
-    public Task<ActionResult<BaseItemDto>> GetItemLegacy(
+    public ActionResult<BaseItemDto> GetItemLegacy(
         [FromRoute, Required] Guid userId,
         [FromRoute, Required] Guid itemId)
         => GetItem(userId, itemId);
```
IDisposable, line 122: The legacy version likely should really be doing a "wait" until the queue refresh finishes, so it isn't returning the "before" DTO...

Jellyfin.Api/Controllers/UserLibraryController.cs
Shadowghost: It's different behavior to any other endpoint we have and it is essentially a library refresh every time somebody calls the endpoint.
IMO implicitly refreshing data and not returning what is currently in the DB is bad practice

enter-a-random-username:
Tested a bit
- Person search is now similar to all others requests
- Favorites is faster as well, still 2.5x slower with 0 entries (never marked a person as favorite)
- Menu opens faster as well and down to ~450ms compared to 30-40ms for an episode or series. (I usually want to change the image and not refresh or edit the other metadata)

It is a big improvement, and I guess a similar time could only be achieved if a parameter is added to do less work for places that do not require the information.

Bond-009 (approved):

PR #17466, opened 2026-07-28
Improve People deduplication, fix search and restrict ItemByName responses

### LINKED ISSUE #2207: Actors Appended To Unviewable Libraries Visible
If you create an actor, then add it to a movie.
If that movie is under a library that a user doesn't have access too, that actor still appears in searches.

I think it is greatly needed to have an option to hide all actors appended to items in libraries depending on user access permissions. Or something along these lines, maybe per actor. But I can definitely see the latter being extremely tedious to select all the actors individually with big libraries.

Specifically in this case to be honest, it's a nsfw video, making a nsfw actress picture visible to people that don't have access to that library. Hoping for a fix in the next update or two maybe before someone comes across anything lmfao.

### LINKED ISSUE #17253: [12.0RC2] Genres fail to load in Modern Layout with large libraries
### Description of the bug

When using the "Modern Layout" web interface, movie genres do not load or display correctly. This issue specifically occurs in large libraries that contain a high volume of movies and a diverse set of genres.

### Reproduction steps

1. Open Jellyfin using the web client.
2. Ensure the display settings are set to "Modern Layout".
3. Navigate to the "Movies" library.
4. Click on the "Genres" tab at the top

The UI freezes, fails to fetch the data, or displays an empty page when trying to populate the genre list.

All available genres should load promptly and display their respective poster cards, regardless of the library size.

### Jellyfin Server version

Unstable

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

12.0 RC2

### Environment

```markdown
- OS: Ubuntu 26.04
- Virtualization: KVM
```

### Jellyfin logs

```shell
-
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

### LINKED ISSUE #17491: Search no longer returns People, Genres or Studios
### Description of the bug

Since the search provider rewrite (#16121, commit 07a802d8fa), `/Search/Hints` and `/Items?searchTerm=...` no longer return People, Genres, Studios or other by-name items when the request carries a `userId`. Only media items appear in search results. The `includePeople`, `includeGenres`, `includeStudios` and `includeArtists` parameters still exist and default to true but have no effect.

### Reproduction steps

1. Run Jellyfin from current master (or any 12.0 release candidate).
2. Create a movie library where a movie has an actor (e.g. "Shared Actor"), a genre (e.g. "War") and a studio (e.g. "Warner Bros") in its metadata.
3. Search for the actor, the genre or the studio by name, with a user attached:

```
curl -H "Authorization: MediaBrowser Token=\"<token>\"" \
  "http://localhost:8096/Search/Hints?searchTerm=Shared&userId=<userId>"
```

The by-name items are silently missing from the results:

```json
{"TotalRecordCount": 0, "SearchHints": []}
```

For `searchTerm=War&includeGenres=true` only the BoxSet (a library item) survives, the Genre and the Studio are gone:

```json
{"TotalRecordCount": 1, "SearchHints": [{"Name": "War Duo", "Type": "BoxSet"}]}
```

Passing `includePeople=true`, `includeGenres=true` or `includeStudios=true` explicitly makes no difference. The same searches **without** a `userId` (raw API key) return the by-name items correctly.

The behavior of 10.11.x and of master builds before #16121 (verified on d9ced0d639 with identical media):

```json
{"TotalRecordCount": 1, "SearchHints": [{"Name": "Shared Actor", "Type": "Person"}]}
```

```json
{"TotalRecordCount": 3, "SearchHints": [{"Name": "War", "Type": "Genre"}, {"Name": "War Duo", "Type": "BoxSet"}, {"Name": "Warner Bros", "Type": "Studio"}]}
```

### Jellyfin Server version

Master

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

12.0

### Environment

```markdown
- OS: macOS 26.5 (server built and run from source), also reproduced on a production Linux 12.0.0 instance
- Linux Kernel: n/a
- Virtualization: none
- Clients: direct API calls (curl)
- Browser: n/a
- FFmpeg Version: n/a (not involved)
- Playback Method: n/a
- Hardware Acceleration: none
- CPU Model: Apple Silicon (arm64)
- GPU Model: n/a
- Plugins: none
- Reverse Proxy: none
- Base URL: none
- Networking: localhost
- Jellyfin Data Storage & Filesystem: local, APFS
- Media Storage & Filesystem: local, APFS
- External Integrations: none
```

### Jellyfin logs

```shell
No errors are logged. All requests return HTTP 200 and the by-name items are silently filtered out of the results.
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

_No response_

Shadowghost:
People, genres, studios and artists are not owned by a library, so nothing on those rows says whether a user may see them, that only follows from the items behind the name.

`/Persons` returned every person in the database regardless of library access, while `/Items?includeItemTypes=Person` dropped them entirely because a by-name row has no `TopParentId` to match.
Search dropped them too, one stage earlier: the candidate query in `SqlSearchProvider` runs its own access filter, and that filter was built without the query's type filters, so it had no way to know by-name types were being asked for.

Since #16121 routed `/Search/Hints` and `/Items?searchTerm=` through the search providers, both lost People, Genres and Studios for every request carrying a `userId`.

**Changes**
All of them now keep a by-name row only when at least one item behind its name passes the user's access filter (library, parental rating, tags). The check is skipped entirely for users with no content restrictions, and `InternalPeopleQuery.EnableTotalRecordCount` lets callers drop the total, which is the only part of the query that cannot short-circuit.

The by-name exemption moved out of `TranslateQuery` into `ApplyAccessFiltering`, so it is applied once for every caller instead of only on the item-list path. `SqlSearchProvider` and `SearchManager` build their access filter through a shared `SearchQueryAccessFilter`, which carries the search's include/exclude types and marks items-by-name as wanted unless the search is scoped to a parent - a by-name row has no parent to descend from, so those searches still drop them.

**Code assistance**
Claude Opus 5 was used to create and run a test harness to a) get the generated SQL by EF core and b) test different approaches

**Issues**
Fixes  #2207
Fixes #17491
Fixes #17253


crobibero:
This is a pretty significant change so I think this is better for v13.
If there is a smaller bug fix that can be extracted from this PR I think it is acceptable for v12

Also, I'm not sure if pure deduplication of the People table is correct as these records don't have any Provider IDs attached, so we could easily be merging two distinct people

Shadowghost:
@crobibero I limited the changes to the existing dedupe logic now, so this should be fine for 12.x now

MediaBrowser.Controller/Entities/PeopleHelper.cs
```diff
@@ -35,57 +35,61 @@ public static void AddPerson(ICollection<PersonInfo> people, PersonInfo person)
...
-                {
-                    people.Add(person);
-                }
-                else
-                {
-                    MergeExisting(existing, person);
-                }
+            // If the type is GuestStar and there's already an Actor entry, then promote it to avoid dupes
+            if (person.Type == PersonKind.GuestStar)
+            {
```
crobibero, line 67: I think either this comment is wrong, or the conditional is?

Shouldn't this be a `IsCastKind` check?

Shadowghost: The conditional is correct: `IsSameCredit` only pairs a `GuestStar` with an `Actor` or another `GuestStar`, so an `IsCastKind(existing.Type)` guard here would always be true.
Guarding on `IsCastKind(person.Type)` would be wrong in the other direction: adding a plain `Actor` would then overwrite the existing entry's type with `GuestStar`.

crobibero: I see, I missed the `IsSameCredit(p, person)` to pull the existing user.

crobibero (approved):

Jellyfin.Server.Implementations/Item/PeopleRepository.cs
```diff
@@ -107,9 +117,13 @@ public void UpdatePeople(Guid itemId, IReadOnlyList<PersonInfo> people)
             person.Role = person.Role?.Trim() ?? string.Empty;
         }
 
-        // multiple metadata providers can provide the _same_ person; dedupe case-insensitively.
-        people = people.DistinctBy(e => e.Name.ToLowerInvariant() + "-" + e.Type).ToArray();
-        var personKeys = people.Select(e => e.Name.ToLowerInvariant() + "-" + e.Type).ToArray();
+        // multiple metadata providers can provide the _same_ credit; dedupe case-insensitively.
+        // The role is part of the key because one person can hold several credits of the same type
+        // on an item, e.g. a Writer credited for both the Novel and the Screenplay.
+        people = people.DistinctBy(e => e.Name.ToLowerInvariant() + "-" + e.Type + "-" + e.Role.ToLowerInvariant()).ToArray();
```
IDisposable: If we project the ToLowerInvariant() values of Name, Type, and Role to loweredName, loweredType, and loweredRole before the distincts, we could remove all those repeated calls lower down.

IDisposable (approved):

PR #17468, opened 2026-07-28
Add HDR passthrough for transcoding

### LINKED ISSUE #17467: Add HDR passthrough so transcodes are not forced to SDR
## Problem

Transcoding an HDR source always produces BT.709 SDR, even when the source is HDR10 and the client reports HDR10 support. Tone mapping is the only available path.

Five separate mechanisms pin the output to SDR:

1. Every tone mapping filter hardcodes `p=bt709:t=bt709:m=bt709` (`GetHwTonemapFilter`, and the `tonemapx` argument string in the software chain). These are string literals, so no setting changes them.
2. `GetOverwriteColorPropertiesParam` rewrites the colour properties to BT.709 whenever no tone mapping runs.
3. The filter chains select an 8-bit output format (`nv12` or `yuv420p`).
4. `GetVideoQualityParam` downgrades HEVC Main 10 to Main.
5. `DynamicHlsHelper` writes `VIDEO-RANGE=SDR` and an 8-bit codec string into the HLS playlist.

Turning tone mapping off does not help. It removes the tone curve but still strips the colour tags and the tenth bit, which is what produces the washed out picture that gets reported.

`master_display` and `max_cll` appear nowhere in the codebase, so the HDR metadata handling added to jellyfin-ffmpeg never gets used by the server.

## Proposed approach

Add an `EnableHdrPassthrough` encoding option, off by default. When all of the following hold, the tone mapping filter is dropped, the chain stays 10-bit through to the encoder, and the BT.2020 primaries and PQ or HLG transfer reach the output:

* the source has a static HDR10 or HLG base layer
* the client reports the matching range as supported
* an encoder exists that can emit 10-bit

ffmpeg propagates the mastering display and content light level side data on its own, so static HDR10 metadata survives without the server restating it.

Cases that keep tone mapping:

* Dolby Vision without an HDR10 or HLG base layer, because dropping the RPU leaves invalid colours and no encoder here can rewrite one
* clients that do not declare a matching range, because undeclared capabilities are not the same as HDR being safe to send
* hardware with no 10-bit encoder

Eligibility is decided by naming the encoder that would actually run and asking whether it exists, rather than by an allowlist of accelerators. V4L2 M2M resolves to no encoder because it is 8-bit only, as do RKMPP and VideoToolbox for AV1.

The HLS playlist also has to be corrected. A client configures its decoder from `VIDEO-RANGE` and the codec string, so a stream announced as SDR `hvc1.1` but delivered as Main 10 PQ fails to initialise and never requests a media segment.

## Verification

Confirmed end to end on a running server with `hevc_nvenc` (Ampere) and `hevc_vaapi` (AMD Cezanne). Both produce Main 10 output carrying BT.2020 primaries, the PQ transfer, and mastering display and content light level side data. Playback confirmed in Safari on iPadOS and in Chrome.

I have the change ready and will open a PR referencing this issue. A companion change to jellyfin-web adds the setting and reports the resulting range in the playback stats.

karan-vk:
**Changes**

Adds an `EnableHdrPassthrough` encoding option, off by default, that keeps the source HDR through a transcode instead of tone mapping it to SDR. When the source has a static HDR10 or HLG base layer, the client reports the matching range as supported, and an encoder exists that can emit 10-bit, the tone mapping filter is dropped, the filter chain stays 10-bit through to the encoder, and the BT.2020 primaries and PQ or HLG transfer reach the output.

Five things had to change together, because any one of them alone still yields SDR:

| Location | Before | After |
| --- | --- | --- |
| `GetHwTonemapFilter`, software `tonemapx` string | tone map filter always inserted | no tone map filter under passthrough |
| `GetOverwriteColorPropertiesParam` | rewrites colour properties to BT.709 | keeps the source PQ or HLG tags |
| Filter chains | 8-bit `nv12` or `yuv420p` | `p010` or `yuv420p10le` |
| `GetVideoQualityParam` | HEVC Main 10 downgraded to Main | Main 10 kept |
| `DynamicHlsHelper` | `VIDEO-RANGE=SDR`, `hvc1.1` | `VIDEO-RANGE=PQ` or `HLG`, `hvc1.2` |

The playlist part matters for playback, not just correctness. A client configures its decoder from `VIDEO-RANGE` and the codec string, so a stream announced as SDR `hvc1.1` but delivered as Main 10 PQ fails to initialise. In testing Safari fetched the init segment repeatedly and never requested a media segment.

`TranscodingInfo` also gains `VideoRangeType`, reporting the range that is actually produced: the source range for a copy or a passthrough, SDR when the picture was tone mapped. Without it a client cannot tell whether HDR survived.

**Behaviour when passthrough is not possible**

Tone mapping is kept, unchanged, in every one of these cases:

* the option is off, which is the default
* Dolby Vision without an HDR10 or HLG base layer, because dropping the RPU leaves invalid colours and no encoder here can rewrite one
* the client does not declare a matching range, because undeclared capabilities are not the same as HDR being safe to send
* the output codec is H.264, which has no practical HDR client support
* no 10-bit encoder exists for the configured accelerator

Eligibility is decided by naming the encoder that would run and asking `SupportsEncoder`, rather than by an allowlist of accelerators. Adding an accelerator is one line. V4L2 M2M resolves to no encoder because it is 8-bit only, as do RKMPP and VideoToolbox for AV1.

All eight filter chains carry 10-bit under passthrough: software, NVENC, QSV on VAAPI and on D3D11, AMF, Intel VAAPI, AMD VAAPI with Vulkan, the limited i965 and AMD legacy VAAPI path, RKMPP, and VideoToolbox.

**Testing**

11 new unit tests cover every accept and refuse branch, including HLG, Dolby Vision profile 5, profile 8.1 with HDR10+, H.264 output, an undeclared client, a missing encoder, and V4L2 M2M. Existing suites pass unchanged: Controller 105, MediaEncoding 97, Api 108.

Verified end to end on a running server. `hevc_nvenc` on Ampere and `hevc_vaapi` on AMD Cezanne both produce Main 10 output carrying BT.2020 primaries, the PQ transfer, and mastering display and content light level side data. ffmpeg propagates that side data on its own, so the server does not restate it. Playback confirmed in Safari on iPadOS and in Chrome.

Generated filter graph for the same file and hardware, differing only in whether the client advertises HDR10:

```
SDR client : setparams=...smpte2084..., scale_cuda=w=1280:h=720, tonemap_cuda=format=yuv420p:p=bt709:t=bt709:m=bt709
HDR client : setparams=...smpte2084..., scale_cuda=w=1280:h=720:format=p010
```

**Known limitation**

`SupportsEncoder` answers whether an encoder is present in the ffmpeg build, not whether it can initialise. On a machine where jellyfin-ffmpeg is compiled with AMF but `libamfrt64.so.1` is absent, `hevc_amf` probes as available and fails at runtime. That failure is not specific to this change, since AMF transcoding fails there regardless, but the probe is not a guarantee that the hardware works.

**Code assistance**

**Issues**

Fixes #17467

### DIFF
--- a/Jellyfin.Api/Helpers/DynamicHlsHelper.cs
+++ b/Jellyfin.Api/Helpers/DynamicHlsHelper.cs
@@ -466,9 +466,23 @@ public class DynamicHlsHelper
                     }
                 }
             }
+            else if (_encodingHelper.IsHdrPassthroughAvailable(state, _serverConfigurationManager.GetEncodingOptions()))
+            {
+                // Clients configure their decoder from this field; announcing SDR while
+                // delivering PQ prevents playback.
+                switch (videoRangeType)
+                {
+                    case VideoRangeType.HLG:
+                    case VideoRangeType.DOVIWithHLG:
+                        builder.Append(",VIDEO-RANGE=HLG");
+                        break;
+                    default:
+                        builder.Append(",VIDEO-RANGE=PQ");
+                        break;
+                }
+            }
             else
             {
-                // Currently we only encode to SDR.
                 builder.Append(",VIDEO-RANGE=SDR");
             }
         }
@@ -815,6 +829,14 @@ public class DynamicHlsHelper
                 || string.Equals(state.ActualOutputVideoCodec, "av1", StringComparison.OrdinalIgnoreCase))
             {
                 profileString ??= "main";
+
+                // Passthrough encodes HEVC as Main 10; announcing Main describes an 8-bit
+                // stream and the client fails to initialise on the 10-bit init segment.
+                if (!string.Equals(state.ActualOutputVideoCodec, "av1", StringComparison.OrdinalIgnoreCase)
+                    && _encodingHelper.IsHdrPassthroughAvailable(state, _serverConfigurationManager.GetEncodingOptions()))
+                {
+                    profileString = "main10";
+                }
             }
         }
 
--- a/Jellyfin.Api/Helpers/DynamicHlsHelper.cs
+++ b/Jellyfin.Api/Helpers/DynamicHlsHelper.cs
@@ -441,52 +441,26 @@ public class DynamicHlsHelper
     /// <param name="state">StreamState of the current stream.</param>
     private void AppendPlaylistVideoRangeField(StringBuilder builder, StreamState state)
     {
-        if (state.VideoStream is not null && state.VideoStream.VideoRange != VideoRange.Unknown)
+        if (state.VideoStream is null || state.VideoStream.VideoRange == VideoRange.Unknown)
         {
-            var videoRange = state.VideoStream.VideoRange;
-            var videoRangeType = state.VideoStream.VideoRangeType;
-            if (EncodingHelper.IsCopyCodec(state.OutputVideoCodec))
-            {
-                if (videoRange == VideoRange.SDR)
-                {
-                    builder.Append(",VIDEO-RANGE=SDR");
-                }
-
-                if (videoRange == VideoRange.HDR)
-                {
-                    switch (videoRangeType)
-                    {
-                        case VideoRangeType.HLG:
-                        case VideoRangeType.DOVIWithHLG:
-                            builder.Append(",VIDEO-RANGE=HLG");
-                            break;
-                        default:
-                            builder.Append(",VIDEO-RANGE=PQ");
-                            break;
-                    }
-                }
-            }
-            else if (_encodingHelper.IsHdrPassthroughAvailable(state, _serverConfigurationManager.GetEncodingOptions()))
-            {
-                // Clients configure their decoder from this field; announcing SDR while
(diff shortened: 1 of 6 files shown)


gnattu:
No this is much more involved than what you think. Not all hardware platforms are ready for this and not all clients are ready for this. You are outright crashing transcoding pipeline for a lot of users.

nyanmisaka:
I'm adding some context so you understand the complexity of the preparatory work.
https://github.com/jellyfin/jellyfin-ffmpeg/pull/733

anchpop:
I think one issue with this PR is that it uses _mediaEncoder.SupportsEncoder(...), which doesn't actually tell you if your hardware supports the encoder. I think the handling of subtitles is also incorrect (for burn-in subtitles on HDR content, I think they render at maximum brightness). Still, I appreciated the PR as it was useful for getting this working on my setup

PR #17473, opened 2026-07-29
Fix InvalidOperationException when guide requests an unmatched channel

loguefx:
## Summary

`DefaultLiveTvService.GetProgramsAsync` used `.First(...)` to locate a channel by id, which throws `InvalidOperationException: "Sequence contains no matching element"` when no channel matches. `GuideManager` catches this per channel and logs it at **ERROR**, which spams the log and can leave guides empty. A channel with no matching EPG entry is a **normal** condition, not an error.

## Fix

Use `.FirstOrDefault(...)` and return an empty result (logged at Debug) when there is no match.

```csharp
var channel = channels.FirstOrDefault(i => string.Equals(i.Id, channelId, StringComparison.OrdinalIgnoreCase));
if (channel is null)
{
    _logger.LogDebug("No channel found matching id {ChannelId}; returning no programs", channelId);
    return [];
}
```

## Tests

Adds `DefaultLiveTvServiceTests.GetProgramsAsync_UnmatchedChannelId_ReturnsEmpty`: against `master` it fails with the exception above; with this change it returns empty and the suite stays green.

## Notes

- Small, backward-compatible, server-side change (benefits every client/OS).


crobibero:

PR #17474, opened 2026-07-29
Parse XMLTV guide once per refresh instead of once per channel

loguefx:
## Summary

`XmlTvListingsProvider.GetProgramsAsync` builds a new `XmlTvReader` and calls `GetProgrammes(channelId, ...)` on every call, and that streams and parses the **entire** XMLTV file to filter for a single channel. `GuideManager` calls this once per channel, so for *N* channels the whole file is parsed *N* times — **O(channels × filesize)**. With a large IPTV provider (tens of thousands of channels, multi-GB XMLTV) a guide refresh becomes pathologically slow.

## Fix

Parse the file a single time into a `channelId → programmes` dictionary, cached and keyed by path + last-write-time, then serve each channel from memory.

```csharp
var programsByChannel = GetProgramsByChannel(path, GetLanguage(info), cancellationToken);
if (!programsByChannel.TryGetValue(channelId, out var channelPrograms))
{
    return [];
}

return channelPrograms
    .Where(p => p.EndDate >= startOffset && p.StartDate < endOffset)
    .Select(p => GetProgramInfoWithEtag(p, info))
    .ToList();
```

Fresh `ProgramInfo` objects are still materialized per call because callers mutate them (`Id`/`ChannelId`), so only the lighter parsed entries are cached. The cache rebuilds automatically when a newer file is downloaded (write-time changes). This turns **O(channels × filesize)** into **O(filesize)**.

## Tests

Three new tests in `XmlTvListingsProviderTests` (+ a multi-channel XMLTV test data file): only the requested channel's programmes are returned; the date window is respected; repeated calls return fresh objects so caller mutation cannot corrupt the cache.

## Notes

- Parsing semantics are unchanged — only the number of file reads changes (N → 1).

### DIFF
--- a/src/Jellyfin.LiveTv/Listings/XmlTvListingsProvider.cs
+++ b/src/Jellyfin.LiveTv/Listings/XmlTvListingsProvider.cs
@@ -9,6 +9,7 @@ using System.Linq;
 using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Xml;
 using Jellyfin.Extensions;
 using Jellyfin.XmlTv;
 using Jellyfin.XmlTv.Entities;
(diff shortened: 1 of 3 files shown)


crobibero:

PR #17475, opened 2026-07-29
Add Referer and custom HTTP headers for M3U tuners

loguefx:
## Summary

For M3U tuner sources, `M3UTunerHost.CreateMediaSourceInfo` only ever set a **User-Agent** on the outgoing request. Many IPTV providers return **HTTP 406/403** unless a `Referer` (or another provider-specific header) is present, so those channels fail in Jellyfin while playing fine in VLC/ffmpeg where such headers can be supplied. This also affects DVR, which opens the stream through the same tuner host.

## Fix

Add optional `Referer` and `CustomHttpHeaders` to `TunerHostInfo` and populate `mediaSource.RequiredHttpHeaders` from them alongside the existing User-Agent.

```csharp
if (!string.IsNullOrWhiteSpace(info.Referer))
{
    httpHeaders[HeaderNames.Referer] = info.Referer;
}

if (info.CustomHttpHeaders is not null)
{
    foreach (var header in info.CustomHttpHeaders)
    {
        if (!string.IsNullOrWhiteSpace(header.Name) && !string.IsNullOrWhiteSpace(header.Value))
        {
            httpHeaders[header.Name] = header.Value;
        }
    }
}
```

**Important:** `CustomHttpHeaders` is a `NameValuePair[]`, **not** a `Dictionary<string,string>`. `LiveTvOptions` is persisted via `XmlSerializer`, which throws `NotSupportedException` on `IDictionary` members and would break *all* tuner saves. This mirrors the `NameValuePair[]` pattern already used by `ListingsProviderInfo.ChannelMappings`.

## Tests

`M3UTunerHostTests` (3 tests): the default User-Agent is still applied; a configured `Referer` lands in `RequiredHttpHeaders`; custom headers are applied.

## Notes

- Fully backward-compatible: existing configs with none of the new fields behave exactly as before.
- A matching `jellyfin-web` config UI (inputs for Referer + custom headers) can follow as a separate web PR.

### DIFF
--- a/MediaBrowser.Model/LiveTv/TunerHostInfo.cs
+++ b/MediaBrowser.Model/LiveTv/TunerHostInfo.cs
@@ -1,6 +1,9 @@
 #nullable disable
 #pragma warning disable CS1591
 
+using System;
+using MediaBrowser.Model.Dto;
+
 namespace MediaBrowser.Model.LiveTv
 {
     public class TunerHostInfo
@@ -13,6 +16,7 @@ namespace MediaBrowser.Model.LiveTv
             AllowStreamSharing = true;
             AllowFmp4TranscodingContainer = false;
             FallbackMaxStreamingBitrate = 30000000;
+            CustomHttpHeaders = Array.Empty<NameValuePair>();
         }
 
         public string Id { get; set; }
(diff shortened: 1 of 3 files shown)


crobibero:

PR #17476, opened 2026-07-29
Make Live TV guide refresh scale for large IPTV playlists

loguefx:
## Summary

On a large IPTV/M3U playlist (tens of thousands of channels) a guide refresh could take **hours**. This PR removes several avoidable per-refresh costs so a large refresh completes in minutes, with no behavior change unless a user opts in.

## Changes

**Performance (no behavior change):**
- `CleanDatabase` used `currentIdList.Contains` on a `Guid[]` (O(n·m)); build a `HashSet<Guid>` once for O(1) membership, removing the quadratic tail.
- The per-channel existing-programmes DB query ran even for channels with no EPG; skip it when the provider returned no programmes (stale rows are still removed by `CleanDatabase`).
- Every channel was re-persisted and re-metadata-refreshed (re-downloading its logo) on every refresh; track whether a channel actually changed and only save/refresh when it did.
- Channel icons were re-applied (and thus re-downloaded) even when unchanged; remember the icon source in a provider id and skip re-applying an unchanged icon.
- Lower two per-channel `LogInformation` lines (`M3uParser`, `XmlTvListingsProvider`) to `LogDebug`; they emitted tens of thousands of lines per refresh.

**Opt-in options (default = current behavior):**
- `LiveTvOptions.PreCacheGuideImages` (default `true`). When disabled, channel-logo and programme-image downloads are skipped during refresh and load lazily on first display instead (`ImageController` already converts remote images to local on first request), so dead IPTV image links no longer stall the refresh on network timeouts.
- `LiveTvOptions.HideChannelsWithoutProgrammes` (default `false`). The refresh tags channels that received programmes (`"GuideData"`) and untags those that lost them; when enabled, `GetInternalChannels` filters on that tag so blank channels are excluded at the DB level (pagination/totals stay correct). One refresh is required after enabling before blank channels drop.

First-time import is unchanged (everything is "new"); steady-state refreshes now skip untouched channels entirely.

## Tests

Unit tests for the unchanged-icon skip (`LiveTvChannelImageHelperTests`) and the hide-blank-channels tag filter (`LiveTvManagerChannelFilterTests`).

## Notes

- Happy to split the opt-in `HideChannelsWithoutProgrammes` feature into its own PR if maintainers prefer to keep this purely a performance change.

### DIFF
--- a/MediaBrowser.Model/LiveTv/LiveTvOptions.cs
+++ b/MediaBrowser.Model/LiveTv/LiveTvOptions.cs
@@ -44,5 +44,21 @@ namespace MediaBrowser.Model.LiveTv
         public bool SaveRecordingNFO { get; set; } = true;
 
         public bool SaveRecordingImages { get; set; } = true;
+
+        /// <summary>
+        /// Gets or sets a value indicating whether channels without any guide data are hidden
+        /// from the channel list and guide. Useful for large IPTV playlists (tens of thousands
+        /// of channels) where most channels have no EPG, leaving the guide cluttered with blank rows.
+        /// </summary>
+        public bool HideChannelsWithoutProgrammes { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether channel logos and programme images are
+        /// downloaded and cached locally during the guide refresh. When disabled, images are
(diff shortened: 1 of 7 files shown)


crobibero:

PR #17477, opened 2026-07-29
Improve Live TV DVR reliability: retry stream open, clean up failed captures

loguefx:
## Summary

Two related DVR robustness fixes for flaky IPTV/M3U sources.

### Retry the stream open (D1)

`RecordStream` opened the live stream once, so a single transient hiccup lost the whole recording before the recorder even started. Wrap the open in `OpenWithRetryAsync` (3 attempts, exponential backoff 2s→10s):

```csharp
var liveStreamResponse = await OpenWithRetryAsync(
    ct => _mediaSourceManager.OpenLiveStreamInternal(
        new LiveStreamRequest { ItemId = channel.Id, OpenToken = mediaStreamInfo.OpenToken }, ct),
    StreamOpenMaxAttempts, StreamOpenInitialRetryDelay, StreamOpenMaxRetryDelay,
    _logger, recordingInfo.CancellationTokenSource.Token).ConfigureAwait(false);
```

It retries every exception except `OperationCanceledException` (propagated immediately) and re-throws the final failure. The existing coarse 60s whole-recording retry still applies if all in-line attempts fail.

### Don't leave orphaned artifacts behind (D2)

The `.nfo` metadata sidecar was written from the recorder's `onStarted()` callback — i.e. **before any video bytes were confirmed** — so a capture that never delivered data left a metadata-only phantom item in the library (the recurring "I get a `.nfo` instead of the show" complaint). Now:

- the sidecar is written in finalize and **only when a usable (non-empty) recording exists**;
- the old 0-byte cleanup is replaced by `DeleteOrphanedRecordingArtifacts`, which removes the empty video, the orphaned `.nfo`, and the now-empty recording folder.

```csharp
internal static bool HasUsableRecording(string recordingPath)
    => File.Exists(recordingPath) && new FileInfo(recordingPath).Length > 0;

internal static bool DeleteOrphanedRecordingArtifacts(string recordingPath, ILogger logger)
{
    if (HasUsableRecording(recordingPath)) { return false; }
    TryDeleteFile(recordingPath, logger);
    TryDeleteFile(Path.ChangeExtension(recordingPath, ".nfo"), logger);
    TryDeleteEmptyDirectory(Path.GetDirectoryName(recordingPath), logger);
    return true;
}
```

Net effect: a failed capture leaves nothing behind, and the sidecar is only ever written for a real recording.

## Tests

- `OpenWithRetryAsync`: success-first-try, fail-then-succeed, exhausted-attempts, cancellation.
- Orphan cleanup / `HasUsableRecording`: empty/missing video is removed with its `.nfo`; a real recording keeps both.

## Notes

- A follow-up mid-recording reconnect/resume (so a dropped stream reconnects and finishes as one file) builds on this and will be a separate PR.

### DIFF
--- a/src/Jellyfin.LiveTv/Recordings/RecordingsManager.cs
+++ b/src/Jellyfin.LiveTv/Recordings/RecordingsManager.cs
@@ -51,6 +51,10 @@ public sealed class RecordingsManager : IRecordingsManager, IDisposable
     private readonly SeriesTimerManager _seriesTimerManager;
     private readonly RecordingsMetadataManager _recordingsMetadataManager;
 
+    private const int StreamOpenMaxAttempts = 3;
+    private static readonly TimeSpan StreamOpenInitialRetryDelay = TimeSpan.FromSeconds(2);
+    private static readonly TimeSpan StreamOpenMaxRetryDelay = TimeSpan.FromSeconds(10);
+
     private readonly ConcurrentDictionary<string, ActiveRecordingInfo> _activeRecordings = new(StringComparer.OrdinalIgnoreCase);
     private readonly AsyncNonKeyedLocker _recordingDeleteSemaphore = new();
     private bool _disposed;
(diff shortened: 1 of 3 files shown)


crobibero:

PR #17481, opened 2026-07-29
Pr/livetv reconnect q1

loguefx:
Flaky IPTV/M3U sources routinely drop the connection mid-stream, which previously truncated the recording at the first drop. This records across reconnects (exponential backoff, append/resume to the same file) so a laggy/dropping source still produces a single, complete file. Adds append mode to IRecorder, ffmpeg -reconnect for HTTP inputs, HTTPS→HTTP redirect + required-header handling on the direct path, and RecordWithReconnectAsync with unit tests. Stacked on #17477 (DVR retry + orphaned-artifact cleanup); until that merges this PR's diff includes that commit too.
**Changes**

**Code assistance**

**Issues**

### DIFF
--- a/src/Jellyfin.LiveTv/IO/DirectRecorder.cs
+++ b/src/Jellyfin.LiveTv/IO/DirectRecorder.cs
@@ -1,6 +1,7 @@
 #pragma warning disable CS1591
 
 using System;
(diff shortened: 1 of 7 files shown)


crobibero:

PR #17484, opened 2026-07-29
fix(search): match accented People names in NameContains query

### LINKED ISSUE #17483: Search with diacritics still broken for People (actors/directors), even though item search was fixed
### Description of the bug

After the diacritics search fix (PR #15435, included since 10.11.2) was released, search for items (movie/show titles) with accented characters now works correctly. However, search for People (actors, directors) still does not work reliably with diacritics.

### Reproduction steps

Have a person in the library whose name contains a diacritic, e.g. "René".
Go to the search bar.
Search for "Ren" (without diacritic) → the person appears in results.
Search for "René" (with diacritic) → the person does NOT appear in results.
For comparison, search for a movie title containing a diacritic → works correctly.

Searching "René" should return the same results as searching "Ren", consistent with how item title search now behaves after PR #15435.

After the diacritics search fix (PR #15435, included since 10.11.2) was released, search for items (movie/show titles) with accented characters now works correctly. However, search for People (actors, directors) still does not work reliably with diacritics.

Reproduction steps

Have a person in the library whose name contains a diacritic, e.g. "René".
Go to the search bar.
Search for "Ren" (without diacritic) → the person appears in results.
Search for "René" (with diacritic) → the person does NOT appear in results.
For comparison, search for a movie title containing a diacritic → works correctly.

### Jellyfin Server version

10.11.11

### Specify commit id

Master

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.11 (stable)

### Environment

```markdown
- OS: windows
- Clients:Web
- Browser: Firefox
```

### Jellyfin logs

```shell
No specific logs for this issue
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Additional information

This suggests PR #15435 addressed diacritic normalization for item search but not for the People search path, which appears to use a separate code path/index.

sergioperezcheco:
Searching for a Person whose name contains diacritics (e.g. "René") failed even though searching without the accent ("Ren") worked fine. The root cause was a mismatch between how SQLite and .NET fold case: UPPER(name) on the SQLite side only folds ASCII letters (é stays é), while filter.NameContains.ToUpper() in .NET folds full Unicode (é → É). The Contains comparison between these two differently-folded strings never lined up for accented characters.

This replaces the Contains call with EF.Functions.Like on UPPER(name) and drops the .NET ToUpper() on the search term. SQLite LIKE is ASCII case-insensitive and compares non-ASCII bytes exactly, so the accent matches on both sides. The net effect is that "Ren", "René", "ren", and "rené" all find a person named "René", consistent with how item title search already behaves after #15435.

Added unit tests covering accented, unaccented, and mixed-case search terms against the PeopleRepository.

Fixes #17483

### DIFF
--- a/Jellyfin.Server.Implementations/Item/PeopleRepository.cs
+++ b/Jellyfin.Server.Implementations/Item/PeopleRepository.cs
@@ -284,8 +284,17 @@ public class PeopleRepository(IDbContextFactory<JellyfinDbContext> dbProvider, I
 
         if (!string.IsNullOrWhiteSpace(filter.NameContains))
         {
-            var nameContainsUpper = filter.NameContains.ToUpper();
-            query = query.Where(e => e.Name.ToUpper().Contains(nameContainsUpper));
+            // The previous code compared SQLite's UPPER(name) against a .NET-uppercased search
+            // term. SQLite's UPPER() only folds ASCII letters, so "René" -> "RENé", whereas .NET's
+            // ToUpper() folds accented letters too, so "René" -> "RENÉ". The mismatch meant
+            // searching for an accented name failed to match the identical stored name, even though
+            // searching without the accent ("Ren") matched. Use EF.Functions.Like so the match is
+            // handled by SQLite's case-insensitive LIKE on UPPER(name); LIKE folds ASCII case but
+            // compares non-ASCII characters exactly, so the accented letter lines up on both sides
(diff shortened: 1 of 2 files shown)


theguymadmax:
This did not resolved the linked issue. We are addressing it in https://github.com/jellyfin/jellyfin/pull/17574

Issue #5355, opened 2021-03-03
ToList Usage


shortspider:
**Describe the bug**
In looking at the code I noticed that `ToList` is used quite liberally. Take the following as an example (but there are lots more):
![image](https://user-images.githubusercontent.com/5712543/109748185-6c7f4600-7ba6-11eb-809b-8b5f9b07277b.png)
The problem with the above is that every call to `ToList` creates a copy of the `IEnumerable`. So in the above we create a copy of `_profiles.Values` (this may or may not be needed depending on thread safety) and we then create a another copy after we order and filter the first copy. This second copy at least isn't needed at all since the method states that it returns an `IEnumerable`. This copying is death by 1000 cuts since we need to spend CPU allocating the memory for the copy, doing the copy, and then collecting the original and copy when the GC kicks in.

There are a number of usages like this in the code. I'd like to take a crack at fixing them but wanted to know how best to go about it. One big PR? A PR per class or namespace? I understand this isn't exactly a bug report but it's not a feature request either so I thought I'd post here. Please feel free to close and direct me somewhere else if that's more appropriate.

crobibero:
I would recommend a class per class/namespace, large PRs tend to stick around as they’re less likely to be reviewed

dkanada:
I'd say LoC is a better indicator of review turnaround time. Anything under 200 LoC is usually pretty fast here so if you can fit everything within that size it could be fine :p

cvium:
Your example is not terribly great. The first `.ToList()` is probably superfluous given the lock on the `_profiles` object. But the last `.ToList()` is necessary. LINQ is evaluated lazily, so you have to force the enumeration inside the lock.

shortspider:
@cvium I mean, you still are pointing out that one `ToList` is unneeded so I think it serves as a decent example. The lock does complicate it though.
@crobibero @dkanada thanks for the feedback, I'll try and go by class/namespace while keeping the LoC down.

Bond-009:
@cvium Do we want to keep this open? All the biggest allocations should already be fixed, no?

cvium:
The ToList usage is still a bit excessive in some places, so dunno

felix920506:
Please head to the development channel in our chat rooms to discuss the matter further. https://jellyfin.org/contact

sarah-strawberries:
My classmates @happystar9 and @jesus-sunuc and I are interested in helping out with this; any input on what we should get started with?

sjorge:
[...]

PR #17496, opened 2026-07-30
Fix: Remove unnecessary ToList calls

TheRodzz:
Relates #5355

### DIFF
--- a/Jellyfin.Api/Controllers/DashboardController.cs
+++ b/Jellyfin.Api/Controllers/DashboardController.cs
@@ -53,11 +53,11 @@ public class DashboardController : BaseJellyfinApiController
     public ActionResult<IEnumerable<ConfigurationPageInfo>> GetConfigurationPages(
         [FromQuery] bool? enableInMainMenu)
     {
-        var configPages = _pluginManager.Plugins.SelectMany(GetConfigPages).ToList();
+        var configPages = _pluginManager.Plugins.SelectMany(GetConfigPages);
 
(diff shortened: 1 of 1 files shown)


Jellyfin.Api/Controllers/DashboardController.cs
```diff
@@ -53,14 +53,14 @@ public DashboardController(
     public ActionResult<IEnumerable<ConfigurationPageInfo>> GetConfigurationPages(
         [FromQuery] bool? enableInMainMenu)
     {
-        var configPages = _pluginManager.Plugins.SelectMany(GetConfigPages).ToList();
+        IEnumerable<ConfigurationPageInfo> configPages = _pluginManager.Plugins.SelectMany(GetConfigPages);
```
jakob753951: Why are we using the explicit type?

jakob753951:
Also I don't think this actually resolves #5355 as this only fixes one case of unnecessary `.ToList()` (obviously one fix is better than no fix, but it doesn't resolve the issue completely).

TheRodzz:
> Also I don't think this actually resolves #5355 as this only fixes one case of unnecessary `.ToList()` (obviously one fix is better than no fix, but it doesn't resolve the issue completely).

True, updated the description to say relates instead of resolve.

Jellyfin.Api/Controllers/DashboardController.cs
TheRodzz: fixed

PR #17501, opened 2026-07-31
Skip SIDX in fMP4 HLS segments

alchemyyy:
Prevent FFmpeg's SIDX path from rewriting open-GOP boundary packet timestamps. HLS uses the media playlist for segment indexing and does not require the SIDX box.

**Changes**

Add `skip_sidx` to the MOV flags used for fMP4 HLS video segments:

```text
-hls_segment_options movflags=+frag_discont+skip_sidx
```

FFmpeg’s HLS muxer subsequently appends `+frag_custom+dash+delay_moov`, making the effective set:

```text
+frag_discont+skip_sidx+frag_custom+dash+delay_moov
```

This preserves the existing `frag_discont` behavior and DASH-compatible fragmented MP4 layout while omitting the unnecessary top-level `sidx` box.

**Code assistance**

GPT5.6-Sol was used for testing and validation.

**Issues**

I did not find any pre-existing issues about this.

<br>
<br>
<br>

### Background

Some open-GOP H.264 streams exhibit periodic dropped frames when Jellyfin remuxes them into fMP4 HLS. The affected fragments contain incorrect presentation timestamps before they reach the browser or hls.js.

At one reproduced segment boundary, the packet timing was:

```
Source keyframe:       PTS 18.477, DTS 18.227
Generated fMP4 sample: PTS 18.393, DTS 18.226
Another fMP4 sample:   PTS 18.393
```

FFmpeg moved the segment-opening sample backward by approximately two frames, placing it at the same presentation timestamp as another sample. Chrome rendered one sample and reported the other as dropped. A 165-second playback baseline contained 16 such drops, all aligned with affected segment boundaries.

The problem also reproduced when the generated init segment and fragments were appended as a native fragmented-MP4 Blob, without Jellyfin Web or hls.js processing. This isolates the timestamp corruption to the generated FFmpeg output.

### FFmpeg behavior

For fMP4 HLS output, FFmpeg’s HLS muxer enables:

```
+frag_custom+dash+delay_moov
```

When DASH output is enabled and SIDX generation is not disabled, `movenc` rewrites a boundary packet’s PTS while calculating index timing:

```
pkt->pts = pkt->dts + trk->end_pts - trk->cluster[trk->entry].dts;
```

This is incorrect for certain open-GOP boundaries. The first decode-order sample can present after leading pictures that follow it in decode order. Moving that sample backward can make its PTS collide with one of those leading pictures.

Adding `skip_sidx` bypasses this SIDX-specific timestamp rewrite and preserves the source packet timing.

### Why SIDX can be skipped

Jellyfin’s HLS clients obtain segment locations and durations from the media playlist. They do not require a top-level `sidx` box in each media segment.

This change does not disable FFmpeg’s `dash` MOV flag or alter the `moof`, `traf`, `trun`, `tfdt`, or media payload structure required for fragmented MP4 playback. It only suppresses SIDX generation.

The change is limited to the existing fMP4 video path. MPEG-TS and audio-only fMP4 outputs are unchanged.

### Verification

The workaround was tested by generating both video-only and video+FLAC fMP4 HLS output with:

```
-hls_segment_options movflags=+frag_discont+skip_sidx
```

Compared with the existing output:

- Segment boundaries were unchanged.
- Playlist `EXTINF` durations were unchanged.
- The affected opening samples retained their source PTS.
- The duplicate presentation timestamps were no longer generated.
- Valid fragmented MP4 HLS output was produced without SIDX boxes.

The underlying timestamp rewrite remains an FFmpeg issue and should ultimately be corrected upstream. Skipping SIDX avoids a Jellyfin-FFmpeg or upstream FFmpeg patch for the time being.

### DIFF
--- a/Jellyfin.Api/Controllers/DynamicHlsController.cs
+++ b/Jellyfin.Api/Controllers/DynamicHlsController.cs
@@ -1607,8 +1607,9 @@ public class DynamicHlsController : BaseJellyfinApiController
 
             if (state.VideoStream is not null && state.IsOutputVideo)
             {
-                // fMP4 needs this flag to write the audio packet DTS/PTS including the initial delay into MOOF::TRAF::TFDT
-                hlsArguments += $" {(useLegacySegmentOption ? "-hls_ts_options" : "-hls_segment_options")} movflags=+frag_discont";
+                // fMP4 needs frag_discont to write the audio packet DTS/PTS including the initial delay into MOOF::TRAF::TFDT
+                // HLS does not use SIDX, and skipping it avoids FFmpeg rewriting open-GOP boundary packet PTS
+                hlsArguments += $" {(useLegacySegmentOption ? "-hls_ts_options" : "-hls_segment_options")} movflags=+frag_discont+skip_sidx";
             }
 
             segmentFormat = "fmp4" + outputFmp4HeaderArg;


gnattu (changes requested):
Give concrete example of reproduction, a clear description of what the original issue and a human-verifiable verification step before I can accept this.

If you or your AI did not find any related issue reporting this playback problem it (for years) it might be something else.

alchemyyy:
I've had this problem for quite a while, but never filed an issue for it. It has a pretty stringent set of prerequisites, and the visual effect of the issue can be very hard to notice, doesn't leave any error logs, and IMO its pretty hard to not mistake it for any other kind of jitter.

Since I'm unable to share my actual "production" test files, it was difficult to provide an example. Here is a script to generate a test video which reproduces the issue (note: this script assumes a bare jellyfin installation on debian, specifically it assumes the jellyfin-ffmpeg path): [generate-jellyfin-sidx-test.sh](https://github.com/user-attachments/files/30605485/generate-jellyfin-sidx-test.sh)

You can manually A/B test this in the standard web client (I manually A/B tested it myself and it appears to work). You should get the following playback info: 
image

Enabling the SIDX skipping should result in 0 dropped frames.

In my own personal server this issue is something I encountered pretty regularly with Anime, most likely because of the audio tracks causing remuxing, which is a prerequisite of this issue.

If you'd like to discuss real-world sources more, I'm on the jellyfin discord (alchemyyyyyy) and generally available for DM.

gnattu (approved):
LGTM to me, I have not seen regressions caused by this (yet).

The test video generated demonstrates the frame drop but has no noticeable visual artifacts though.

Shadowghost (approved):

crobibero (approved):

PR #17503, opened 2026-07-31
Use CleanName when sorting by name

theguymadmax:
In #16804 we changed  `ItemSortBy.Name` to use` SortName` instead of `CleanName` to fix an issue with album sorting. That solved the issue at the time, but caused some regressions (see below). The original issue is no longer reproducible with `CleanName` being used, so it is safe to revert the change.

**Code assistance**
N/A

**Issues**
Tracks being sorted by name in the songs tab:

CleanName: 
1a

SortName:
2a


jhparker88:
Shouldn't there be any test coverage for this (cleanname, sortname) to avoid any other regression in future when some other refactoring happens given there were at least three changes to this recently?

Shadowghost (approved):

crobibero (approved):

Issue #17494, opened 2026-07-30
Trickplay generation becomes unresponsive for HEVC Main 10 HDR10 videos


prapador:
### Description of the bug

Trickplay generation consistently fails for HEVC Main 10 HDR10 videos.

The affected media plays correctly in Jellyfin using Direct Play, but Trickplay thumbnails are never generated.

During Trickplay generation, Jellyfin starts the ffmpeg process normally, but after approximately 40 seconds reports:

- Trickplay process unresponsive.
- Stopping trickplay extraction.
- Killing ffmpeg process.

After killing the process, Jellyfin throws a generic `MediaBrowser.Common.FfmpegException`.

The generated ffmpeg command includes the following filter chain:

`setparams -> scale -> tonemapx`

The issue has been reproduced on multiple HEVC Main 10 HDR10 episodes from the same series.

Hardware acceleration for Trickplay is disabled.

### Reproduction steps

1. Enable Trickplay generation.
2. Add a library containing a HEVC Main 10 HDR10 video.
3. Scan the library.
4. Wait for Trickplay generation to start.
5. Observe that Trickplay generation becomes unresponsive and eventually fails.

### What is the current _bug_ behavior?

Trickplay generation becomes unresponsive.

Jellyfin logs:

- Trickplay process unresponsive.
- Stopping trickplay extraction.
- Killing ffmpeg process.

After the process is terminated, Jellyfin throws a `MediaBrowser.Common.FfmpegException` and no Trickplay images are generated.

### What is the expected _correct_ behavior?

Trickplay images should be generated successfully for the video.

### Jellyfin Server version

10.11.11

### Specify the build version

10.11.11

### Environment

```markdown
OS: Debian GNU/Linux 13 (trixie)
Linux Kernel: 6.18.34+rpt-rpi-2712
Virtualization: Docker
Container Image: lscr.io/linuxserver/jellyfin:10.11.11
Clients: Web
Browser: Chrome
FFmpeg Version: 7.1.4-Jellyfin
Playback Method: Direct Play (media playback works correctly)
Hardware Acceleration: Disabled for Trickplay
CPU Model: Raspberry Pi 5 Model B Rev 1.0
GPU Model: Raspberry Pi VideoCore VII
Installed Plugins:
- Missing Episode Fetcher
- OpenSubtitles
- TheTVDB


Networking: Docker bridge (runtipi_tipi_main_network)
Jellyfin Data Storage & Filesystem: Local NVMe SSD - ext4
[...]

Issue #17180, opened 2026-06-25
[v12] Initial Library Scan - Scheduled Tasks block running Scan


ghost:
### Description of the bug

jellyfin_12.0-rc1-amd64, windows 10, empty/new config
Upon first start i added one movie (10731 items) and one show library (Episodes: 13055)

The intial scan run up to 76% which took around 2.5 Hours and then got seemingly stuck at 76%.

While the intial scan was running (at least) two scheduled tasks [Database optimization] and [Refresh people] were triggered twice.

Once very early one - those completed, and once at 76%, at this point the scan did not progress and neither did the scheduled tasks. The logs did not show any error.

After 55 Minutes i manually stopped, through the stop button visible next to a running task, the running [database optimization] and [refresh people] tasks, which were the only tasks visible in the dashboard for a long time seemingly blocking the initial scan and each other. 
The library scan percentage continued to rise at similar pace as before almost immediatly after the scheduled tasks were stopped.

### Reproduction steps

1. Add large library where scan will take 2h+ to complete. (Unsure if only on fresh install)
2. Wait for scheduled tasks to start themselves multiple times during scan.
3. Scheduled tasks, once the scan has progressed for some time - should block the server scan and also themselves no progress.

### What is the current _bug_ behavior?

Scan and Tasks stuck, no progression.

### What is the expected _correct_ behavior?

Scheduled tasks should not block each other or a big running scan.

If this bug is true i assume:
Scheduled tasks should be queued and not triggered while another scheduled task and especially not while a big library scan is running.

### Jellyfin Server version

Unstable

### Specify unstable release number

jellyfin_12.0-rc1-amd64

### Specify the build version

12.0-rc1

### Environment

```markdown
- OS:Windows 10
```

### Jellyfin logs

```shell
[No log entries other than keep alives and chapter extractions before...]

[2026-06-25 00:21:54.530 +02:00] [INF] [17] Emby.Server.Implementations.Chapters.ChapterManager: Stopping chapter extraction for "REDACTED" because a chapter was found with a position greater than the runtime.
[2026-06-25 00:22:02.427 +02:00] [INF] [3] Emby.Server.Implementations.Session.SessionWebSocketListener: Sending ForceKeepAlive message to 1 inactive WebSockets.
[2026-06-25 00:22:50.423 +02:00] [INF] [28] Emby.Server.Implementations.Session.SessionWebSocketListener: Sending ForceKeepAlive message to 1 inactive WebSockets.
[...]

Issue #16083, opened 2026-01-22
Library scan stuck at 87% with infinite parental rating check loop causing 100% CPU usage


mattl135:
### Description of the bug

Library scanning gets stuck at 87% with sustained 100% CPU usage due to an infinite loop repeatedly checking parental ratings on root library folders. Logs show MediaBrowser.Controller.Entities.BaseItem continuously checking the same 6 library folders every 3-5ms. The scan never completes and system remains at high power consumption indefinitely until scan is manually cancelled.

### Reproduction steps

1. Configure Jellyfin with multiple libraries including at least one library with spaces in the display name (e.g., "Home Photos", "Home Videos")
2. Navigate to Dashboard → Scheduled Tasks → "Scan Media Library"
3. Click "Run" to start a full library scan
4. Monitor the scan progress percentage
5. Observe: Scan progresses normally through early percentages (0-80%)
6. Observe: Scan slows significantly and gets stuck at approximately 87% progress
7. Check system CPU usage: CPU utilization spikes to 100%
8. Wait 15+ minutes: Progress remains stuck at 87%, CPU remains elevated

Check debug logs: Look for repeating messages about parental ratings every 3-5ms:
    - "Collections" has no parental rating set.
    - "Home Photos" has no parental rating set.
    - "Home Videos" has no parental rating set.
    - (etc., same 6 messages cycling repeatedly)

10. Scan will not complete even after hours of runtime
11. Must manually cancel the scan to stop the high CPU usage

### What is the current _bug_ behavior?

Scan reaches 87%, triggers an infinite loop of "has no parental rating set.", pegs CPU at 100%, and never completes.

### What is the expected _correct_ behavior?

Library scan should complete normally without getting stuck, consuming reasonable CPU resources, and finish within a reasonable timeframe (previous scans on 10.10.7 completed successfully with the same library configuration).

### Jellyfin Server version

10.11.6

### Specify version number

10.11.6

### Specify the build version

10.11.6

### Environment

```markdown
- OS: TrueNAS SCALE ElectricEel-24.10.2.4
- Linux Kernel: (TrueNAS SCALE default kernel)
- Virtualization: Docker (TrueNAS SCALE official Jellyfin app)
- Clients: Not applicable (issue occurs during library scanning, no playback involved)
- Browser: Not applicable
- FFmpeg Version: 7.1.x (bundled with Jellyfin 10.11.6)
- Playback: Not applicable (scanning issue, not playback)
[...]

PR #17509, opened 2026-07-31
Fix: cycle-guard the parent-chain walks (GetParents / GetCollectionFolders)

streamoverlay-app:
## What
Fixes a silent freeze of the **Trickplay** (~75%) and **Media Segment Scan** (~95%) scheduled tasks caused by an unguarded cyclic parent/owner chain — same bug class as #7891.

## Root cause
`BaseItem.GetParents()` and `LibraryManager.GetCollectionFolders(BaseItem, IEnumerable<Folder>)` are unguarded ancestor-walk `while` loops. On a cyclic parent/owner chain they never terminate, re-entering `GetParent() → GetItemById → _cache.TryGet` each iteration. A process dump shows a worker spinning forever (Cooperative GC, zero monitors held):

```
BitFaster…ConcurrentLruCore.TryGet ← LibraryManager.GetItemById ← BaseItem.GetParent ← LibraryManager.GetCollectionFolders ← GetLibraryOptions ← TrickplayImagesTask / MediaSegmentExtractionTask
```

Both affected tasks call `_libraryManager.GetLibraryOptions(item)` per item → `GetCollectionFolders`.

**BitFaster is not the cause** — for `FastConcurrentLru<Guid, BaseItem>` (reference-type value), `TryGet` is a lock-free `ConcurrentDictionary.TryGetValue` and cannot spin; the spin is Jellyfin's loop re-entering it.

Symptoms: frozen at a %, no error in logs, no ffmpeg/ffprobe spawned (hangs before the encoder), and the server won't shut down gracefully (hard kill required).

## Fix
Add a `HashSet<Guid>` visited-set (the idiom already used in `BaseItem.GetCustomRatingForComparision`, see #7891) to both loops — break on revisit. This also protects all other ancestor-walk callers (`GetTopParent`, `GetAncestorIds`, `GetInheritedTags`, `IsVisibleStandalone`, etc.).

## Changes
- `MediaBrowser.Controller/Entities/BaseItem.cs` — guard `GetParents()`.
- `Emby.Server.Implementations/Library/LibraryManager.cs` — guard `GetCollectionFolders(BaseItem, IEnumerable<Folder>)`.
- `tests/Jellyfin.Controller.Tests/Entities/BaseItemParentsTests.cs` — cyclic a→b→a graph asserts `GetParents()` terminates (hangs pre-patch); plus a non-cyclic chain returns all ancestors.

## Related
- #7891 — same class; the precedent fix (`callstack` guard in `GetCustomRatingForComparision`).
- #16083 — sibling symptom (scan stuck at 87%, infinite `BaseItem` loop); possibly mis-triaged as a dup of #14386.
- #17180 — nearest symptom relative (frozen at 76%, silent), different cause (DB deadlock).
- Not the ffmpeg-side trickplay hangs #11828 / #17294 / #17494 (those spawn ffmpeg / log errors).

### DIFF
--- a/Emby.Server.Implementations/Library/LibraryManager.cs
+++ b/Emby.Server.Implementations/Library/LibraryManager.cs
@@ -2703,8 +2703,17 @@ namespace Emby.Server.Implementations.Library
 
         public List<Folder> GetCollectionFolders(BaseItem item, IEnumerable<Folder> allUserRootChildren)
         {
+            var visited = new HashSet<Guid>();
             while (item is not null)
             {
+                // Guard against a cyclic parent chain (e.g. ParentId set before the parent is
+                // registered during concurrent scans). Without this the loop never terminates
+                // and re-enters the item-id cache forever — the trickplay/segment ~75%/95% freeze.
+                if (!visited.Add(item.Id))
+                {
+                    break;
+                }
+
                 var parent = item.GetParent();
 
                 if (parent is AggregateFolder)
--- a/MediaBrowser.Controller/Entities/BaseItem.cs
+++ b/MediaBrowser.Controller/Entities/BaseItem.cs
@@ -1040,9 +1040,16 @@ namespace MediaBrowser.Controller.Entities
         public IEnumerable<BaseItem> GetParents()
         {
             var parent = GetParent();
+            var visited = new HashSet<Guid>();
 
             while (parent is not null)
             {
+                // Guard against a cyclic parent chain; stop instead of spinning forever.
+                if (!visited.Add(parent.Id))
+                {
+                    break;
+                }
+
                 yield return parent;
 
                 parent = parent.GetParent();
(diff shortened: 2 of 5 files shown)


streamoverlay-app:
This is one of those situations that shouldn't happen, but did happen to me. Trickplay refused to finish for months. Restarts, updated drivers, turned of hardware encoding, multiple Jellyfin updates, etc.. Modest sized collection of media. Not sure what caused this to happen in first place, possibly related to TV recordings. Seems I have a show or movie somehow parented to itself. This is at least a fix for people with similar issues and good practice to add. Does not fix the root cause however, but makes it a non-issue.

Shadowghost:
We should at least add some warning log for the cyclic "bad database".

There are also two more sizes that are an issue:
* `FileRefresher.cs`: `while (!Directory.Exists(item.Path) && !File.Exists(item.Path))`
* `BaseItem.cs`: `OfficialRatingForComparison`

PR #17511, opened 2026-08-01
Load paged item results via split queries to avoid client-side row explosion

### LINKED ISSUE #17510: [v12] Resume / Continue Watching (`/UserItems/Resume`) takes ~8s on a large library, scales with per-user playback history
### Description of the bug

On a large library, the Resume / "Continue Watching" endpoints take ~8s per request, while comparable home-screen endpoints (NextUp, Latest) and even the ItemsByName endpoints stay well under 1s. Because the web client and third-party clients fetch Resume when rendering the home screen, the home hangs for ~8s (the web client visibly stalls on the "Continue Watching" row; a Flutter client I use shows a ~30s spinner on launch as it waits on this plus a couple of follow-up calls).

Two things make it clear this is the Resume query specifically and not general DB slowness:

- **It scales with the size of the library**, not with the number of resumable items returned (the result set is tiny, 12 items).
- **It only affects users that have playback history.** A user with essentially no watch history returns Resume in ~0.35s; users with real history take ~8s. So the query appears to scan the whole library to resolve resumable items for the user rather than using an index on user data.

This looks like the same class of large-library query inefficiency as #17405 and #17192, but on the Resume path. Note #17405's `cache_size` angle (and the fix in #17440) is tied to 64 KiB page-size DBs — this DB uses the default 4 KiB page size and still exhibits the problem, so I don't think #17440 alone would cover this.

### Timings

Measured directly against the API with `curl -w %{time_total}` (server otherwise idle, load ~2):

| Endpoint | Time |
|---|---|
| `GET /UserItems/Resume?Limit=12&MediaTypes=Video&Recursive=true` (user w/ history) | ~7.6s |
| `GET /Users/{uid}/Items/Resume?Limit=12&MediaTypes=Video` (same user) | ~7.1s |
| `GET /Items?Filters=IsResumable&Recursive=true&SortBy=DatePlayed&Limit=12` | ~9.7s |
| `GET /Shows/NextUp?Limit=24` | ~0.5s |
| `GET /Users/{uid}/Items/Latest?Limit=16` | ~0.36s |
| `GET /Studios?Filters=IsFavorite&Recursive=true` | ~0.88s |
| `GET /Genres?Filters=IsFavorite&Recursive=true` | ~0.5s |

Per user (same `/UserItems/Resume` call):

| User | Playback history | Resume time |
|---|---|---|
| A | large | ~8.0s |
| B | large | ~7.5s |
| C | ~none | ~0.36s |

### What I tried

- **"Optimize database" scheduled task** (`OptimizeDatabaseTask`): ran to completion, **no effect** on Resume timing.
- **Bumping the SQLite cache** (`PRAGMA default_cache_size` from the default ~2 MiB to 128 MiB, verified applied): brought Resume from ~8s down to **~5s** warm, but it never drops below that. So a bigger cache helps a bit, but the query is clearly doing a lot of work regardless — consistent with a full scan rather than an index lookup.

For comparison, Emby on the exact same media set / same host serves its Continue Watching instantly, so this isn't the underlying storage or the sheer item count being unreasonable — it's this query path in v12.

### Reproduction steps

1. Large library (see counts below).
2. A user account with a meaningful amount of playback history (several partially-watched items).
3. Call `GET /UserItems/Resume?Limit=12&MediaTypes=Video&Recursive=true` for that user (or just load the web home screen and watch the "Continue Watching" row).
4. Observe multi-second response time that scales with total library size, not with the 12-item result.

### What is the current bug behavior?

Resume / Continue Watching takes ~8s, stalling the home screen for every client.

### What is the expected correct behavior?

Resume should resolve in well under a second, like NextUp and Latest do, since it returns a small, bounded result set.

### Server Configuration

- **Jellyfin version:** 12.0-rc3 (Docker image `jellyfin/jellyfin:preview`)
- **Database:** Jellyfin-SQLite, page_size 4096, WAL, `LockingBehavior=NoLock`
- **Operating System:** Linux (Docker), x64
- **Library counts:** 175,907 items total — 3,372 movies, 588 series, 27,580 episodes, 415 box sets

### Additional notes

Happy to run `EXPLAIN QUERY PLAN` against the generated SQL or provide anonymized DB stats if that helps pin down which index is missing / which scan is happening.

slevin-7:
**Changes**

`GetItems()` and the non-random branch of `GetItemList()` load the full entity graph (up to seven collection includes) in a single query, which multiplies rows per include and leaves EF de-duplicating them client-side — several seconds on large libraries, with `/UserItems/Resume` as the worst case. This switches both call sites to the two-step pattern the random-sort branch already uses: fetch the page of ids first, then load only those items via `AsSplitQuery()` and restore the order in memory. Adding `AsSplitQuery()` to the existing query without the id-projection step actually makes things worse, because the expensive filter re-runs once per include.

**Code assistance**

LLM-assisted profiling and initial patch draft. I reviewed the changes and validated them against my own library (175k items): `/UserItems/Resume` for a user with 82 resumable items goes from ~8s to ~3.6s with this change alone (~0.55s combined with the already-merged f2f66606 + `PrimaryVersionId` index), with identical result sets and no regressions on `/Items`, `/Shows/NextUp`, `/Items/Latest`. Profiling details in #17510.

**Issues**

Fixes #17510 (the materialization half — the `DatePlayed` sort half is already fixed on master)


Shadowghost:
Please adhere to our PR template and LLM policy, pure LLM generated PRs (especially the description) are prohibited.

slevin-7:
Sorry, I should have caught that before submitting. The analysis, profiling, and numbers in the PR are my own work — I used Claude to help translate it into English, since I’m not a native speaker and my own phrasing tends to come out pretty rough. I didn’t realize the “Generated with” note and co-author tag would get pulled in automatically along with that. Next time I’ll write the description myself and at most have individual sentences checked, without tagging the commit message with it.

slevin-7:
I’ll also clean up the commit message (remove the co-author tag) once I’m back at my computer — on mobile right now

Shadowghost:
I also did some local testing, it looks fine so far. But still some things we should consider:

1. The random-sort branch has an empty ID guard, maybe we should also add that to the other ones too?
2. Maybe we should limit the second query to cases where we actually use includes, because when ApplyNavigations attaches nothing (DtoOptions(false) with images/user data off, we don't eally get anything out of it.
3. We should add a comment that because GetItemList is unbounded, every id is now materialised and re-bound as a JSON parameter per split query, incurring memory and parameter cost scaling with result size.
4. Because split queries run outside a transaction, we need to make sure this doesn't bite us.
5. Same codeblock used thrice in the same file, better extract to helper

slevin-7:
Thanks for testing! Addressed all five points:

1. Empty-id guard is now in the shared helper.
2. Done — instead of duplicating the include conditions, LoadItems checks whether ApplyNavigations returned the query unchanged (it only returns a new instance when it actually attaches something). No includes → single direct query as before, no extra roundtrip.
3. Added to the helper's remarks.
4. The order-restore already tolerates this: items deleted between the id projection and the split load just drop out via the ContainsKey filter. Beyond that it's the same class of race the endpoint already had — TotalRecordCount runs as a separate query outside any transaction too, so a mid-flight write could always make count and page disagree slightly. Noted in the remarks as well.
5. Extracted into LoadItems/LoadOrderedItemsById — nice side effect: the random-sort special case collapsed into the same path and is gone entirely.

I've been running this exact revision on my own instance (175k items, rc3 base) since yesterday: /UserItems/Resume sits at ~0.4s warm, result sets match expectations, and random sort still returns properly shuffled pages. No regressions on /Items, /Shows/NextUp or /Items/Latest.

Shadowghost:
Some more comments:
* I think we can improve the unbounded path further: either gate the two-step on `filter.Limit.HasValue` and leave unbounded calls on the single query, or chunk `orderedIds` inside `LoadOrderedItemsById` (maybe ~10k per batch) to cap parameter size - I did not run benchmarks on the actual impact though.
* The ReferenceEquals fast path never executes because `InternalItemsQuery.DtoOptions` defaults to new `DtoOptions()` (all fields), and even new DtoOptions(false) sets EnableImages = true and EnableUserData = true in the constructor. So ApplyNavigations adds `Include(UserData)` (and usually Images, Provider, LockedFields, and LinkedChildEntities whenever IncludeItemTypes is empty) for essentially every caller in the tree. I couldn't find a caller that reaches the fast branch.

Since there are some changes on master also impacting querying, re-testing with merged master would be good.

yhlllry:
邮件已经收到，谢谢。yhlllry这是来自QQ邮箱的自动回复邮件。

slevin-7:
Went through all three points:

* Gated the two-step load on `filter.Limit.HasValue` — unbounded calls now stay on the single query like before. Chunking felt like a lot of machinery for a path that shouldn't be hot anyway, since the row explosion this works around grows with page size.
* You're right about the fast path, I couldn't find a caller that reaches it either — UserData ends up included pretty much everywhere. Dropped it, the unbounded branch now covers the "just run the query as-is" case.
* Merged master and re-ran on a ~32k item library (27.7k episodes). Good call: with the recent EF fixes in, master is already much faster than what I originally measured and the 8s case doesn't reproduce anymore. The split path still wins consistently though: `/UserItems/Resume` goes 0.30s -> 0.21s and a 100-item episode listing with People+Genres fields 2.2s -> 1.96s. Smaller margin than the original numbers, same shape of saving.

Shadowghost (approved):

JPVenson:
Sadly this is the 3rd PR that just throws an "AsSplitQuery" onto it and calls it a day.

https://github.com/jellyfin/jellyfin/pull/15927#issuecomment-3705195650

Shadowghost:
I guess we ended up where we started...

slevin-7:
Fair pushback, but to be clear about what this one actually does: the split query only ever runs over the current page of ids (<= limit), not the whole tree - the real change is materializing the page ids first and only loading navigations for those. And EF split queries reuse the same connection (one command per include), so the sqlite connection overhead argument from #15927 shouldn't apply here.

That said, I'll be honest about where the re-test left us: the 8s case that motivated this is gone on current master, what remains is the 10-30% on paged listings plus pinning random sort to a single evaluation. If the direction is to fix this at the schema level and you don't want query-side changes in the meantime, I'm fine closing this - no hard feelings. Just let me know which way you want to go.

theguymadmax:
Closing per Venson's comments.

PR #17512, opened 2026-08-01
Allow client-rendered graphical subtitles during remux

altqx:
Fix https://github.com/jellyfin/jellyfin-web/pull/8305#issuecomment-5151777645

Keep embedded PGS and VobSub subtitles external when the client advertises support for rendering them. This prevents `SubtitleMethod=Encode` from disabling video stream copy during secondary-audio HLS remuxes.

Regression coverage was added for both PGS and VobSub.

### DIFF
--- a/MediaBrowser.Model/Dlna/StreamBuilder.cs
+++ b/MediaBrowser.Model/Dlna/StreamBuilder.cs
@@ -1582,7 +1582,11 @@ namespace MediaBrowser.Model.Dlna
                     continue;
                 }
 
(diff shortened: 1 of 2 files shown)


nyanmisaka (approved):
It does the trick

theguymadmax (approved):

Shadowghost (approved):

IDisposable (approved):
This complements the recent support of extracting them as external sidecars too!

PR #17514, opened 2026-08-01
Reduce correlated subqueries and parallelize deserialization on the query hot path

dkzetier:
**Changes**
Two changes to the item query. First, ItemSortBy.DatePlayed fell through to a per-row correlated MAX(LastPlayedDate) subquery, which becomes very expensive on large libraries (e.g. /UserItems/Resume, which sorts by DatePlayed). This adds ApplyDatePlayedOrder, mirroring the existing ApplySeriesDatePlayedOrder pattern: pre-aggregate the max played date per PresentationUniqueKey in one grouped query and LEFT JOIN it, instead of a correlated subquery per outer row ([#17422](https://github.com/jellyfin/jellyfin/issues/17422) made that fallback subquery more index-friendly but didn't remove it for the common non-search case; the search + DatePlayed combination still uses the #17422 path). Second, GetItems and GetItemList now spread the per-row JSON deserialization of BaseItemEntity across CPU cores using PLINQ (AsParallel/AsOrdered) instead of deserializing large result sets on a single thread. AsOrdered preserves the database-issued sort order on the list paths, and the random-sort branch is unaffected since it builds a lookup dictionary afterward.

**Code assistance**
Used Claude to profile the query pipeline, diagnose both bottlenecks, and verify them with before/after timing measurements against a live server.

**Issues**
No issue just general performance improvements

### DIFF
--- a/Jellyfin.Server.Implementations/Item/BaseItemRepository.QueryBuilding.cs
+++ b/Jellyfin.Server.Implementations/Item/BaseItemRepository.QueryBuilding.cs
@@ -283,6 +283,15 @@ public sealed partial class BaseItemRepository
             return ApplySeriesDatePlayedOrder(query, filter, context, orderBy);
         }
 
+        // DatePlayed has the same problem as SeriesDatePlayed: the OrderMapper fallback runs a
+        // correlated MAX(LastPlayedDate) subquery per outer row, which is catastrophic on large
+        // result sets. Pre-aggregate the max played date per item in a single grouped query
+        // and left-join it instead.
+        if (!hasSearch && orderBy.Any(o => o.OrderBy == ItemSortBy.DatePlayed))
+        {
+            return ApplyDatePlayedOrder(query, filter, context, orderBy);
+        }
+
         IOrderedQueryable<BaseItemEntity>? orderedQuery = null;
 
         if (hasSearch)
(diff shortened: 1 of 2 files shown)


Shadowghost:
I don't think this is a valid improvment from my tests:
Behavior changes:
* Every other sort field is thrown away. The branch fires on `orderBy.Any(o => o.OrderBy == DatePlayed)` and then hardcodes `OrderBy(MaxDate).ThenBy(SortName)` ascending, discarding the rest of the orderBy array
* Master keys on the item's own `UserData` plus its alternate versions' (`ItemId == e.Id` and `Item.PrimaryVersionId == e.Id`). The PR keys on MAX over everything sharing a `PresentationUniqueKey`. That's equivalent for Video version groups, but not for Series/Season, whose keys are provider-based and shared across libraries, merged duplicates now all sort at the group max.
* Items with a NULL PresentationUniqueKey lose their play date entirely
* I tested the supposed performance gains on my local DB and it was worse for all queries I tested
* `dbQuery.AsEnumerable().AsParallel()` is unsafe and I'd really like to avoide it

Can you check if the perf is better if you run the optimization task before you run your queries?

dkzetier:
I written these changes before https://github.com/jellyfin/jellyfin/pull/17422 existed, and while I referenced that PR as related context, I never re-verified these changes against master once it actually merged. That was my mistake. The approach you did turned out to be genuinely better and faster were this PR's approach isn't. This was a fun thought experiment to dig into, but closing it because of https://github.com/jellyfin/jellyfin/pull/17422. Thanks for the amazing work.

PR #17522, opened 2026-08-03
Store date-only metadata as midnight UTC instead of server-local time

Shadowghost:
**Changes**
Air dates, release dates and birth/death dates are calendar dates, but several metadata providers anchored them to the server's timezone, either via `SpecifyKind(Local).ToUniversalTime()` or by leaving `Kind` as `Unspecified`, which the SQLite `DateTimeKindValueConverter` then converts as local on save. On a `UTC+2` server an episode airing `2026-08-02` was stored as `2026-08-01 22:00:00Z`, so clients that read the date components off the wire rendered it a day early (and the offset shifted again with DST).
This PR adds `AsCalendarDate()` and applies it in the included providers, matching what the NFO parser already does via `AssumeUniversal | AdjustToUniversal`. Existing rows are corrected on the next metadata refresh.

**Code assistance**
Claude Opus 5 was used for tracing the issue across providers once I found the initial culprit.

**Issues**

### DIFF
--- a/MediaBrowser.Providers/Extensions/DateTimeExtensions.cs
+++ b/MediaBrowser.Providers/Extensions/DateTimeExtensions.cs
@@ -1,0 +1,29 @@
+using System;
+
+namespace MediaBrowser.Providers.Extensions;
+
+/// <summary>
+/// Extensions for normalizing dates returned by metadata providers.
+/// </summary>
+internal static class DateTimeExtensions
+{
+    /// <summary>
+    /// Anchors a date-only metadata value to midnight UTC.
+    /// </summary>
+    /// <param name="value">The date reported by the provider.</param>
+    /// <returns>The same calendar date, with <see cref="DateTimeKind.Utc"/>.</returns>
+    public static DateTime AsCalendarDate(this DateTime value)
+    {
+        return DateTime.SpecifyKind(value, DateTimeKind.Utc);
+    }
+
+    /// <summary>
+    /// Anchors an optional date-only metadata value to midnight UTC.
+    /// </summary>
+    /// <param name="value">The date reported by the provider, if any.</param>
+    /// <returns>The same calendar date with <see cref="DateTimeKind.Utc"/>, or <c>null</c>.</returns>
+    public static DateTime? AsCalendarDate(this DateTime? value)
+    {
+        return value.HasValue ? value.Value.AsCalendarDate() : null;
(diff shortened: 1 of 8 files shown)


MediaBrowser.Providers/Extensions/DateTimeExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+internal static class DateTimeExtensions
+{
+    /// <summary>
+    /// Anchors a date-only metadata value to midnight UTC.
+    /// </summary>
+    /// <param name="value">The date reported by the provider.</param>
+    /// <returns>The same calendar date, with <see cref="DateTimeKind.Utc"/>.</returns>
+    public static DateTime AsCalendarDate(this DateTime value)
+    {
+        return DateTime.SpecifyKind(value, DateTimeKind.Utc);
```
cvium, line 17: I don't think this method will behave correctly in all situations. If the raw string representation of the datetime contains an offset, most parsers will convert it to local time, but this method will force it to have kind UTC, which would be incorrect in that situation.

IDisposable: To your point... depending on what we want the behavior to be:

A) the real UTC instant to it's UTC calendar date (so the local day of the value might differ from the UTC day)

```csharp
// Utc: pass-through. Local: converted. Unspecified: treated as local.
static DateOnly UtcDate(DateTime value) => DateOnly.FromDateTime(value.ToUniversalTime());

// Midnight UTC as DateTime
static DateTime UtcMidnight(DateTime value) =>
    UtcDate(value).ToDateTime(TimeOnly.MinValue, DateTimeKind.Utc);
```
B) Keep the original calendar day but declare it to be UTC 

```csharp
// Same Y/M/D, midnight, marked Utc — no conversion.
static DateTime AsUtcMidnight(DateTime value) =>
    DateTime.SpecifyKind(value.Date, DateTimeKind.Utc);
```

Shadowghost: What metadata providers give us is usually the latter, a date with unspecified TZ - but treating that as UTC is best effort because the metadata itself is inconsistent from where that date is gathered: "when did this aird in UTC", "When did this air in the original release timezone" (but here again, we do not know or have the offset), and so on.

PR #17525, opened 2026-08-03
Clear metadata provider cache when provider parts are registered

rlauuzo:
**Changes**

`AddParts` replaces `_metadataProviders` but never invalidated `_metadataProviderCache`, so a second call would keep serving provider arrays from the previous registration. `ClearMetadataProviderCache()` was added for this but had no callers. This calls it at the end of `AddParts`. Not a problem during normal startup, where `AddParts` runs once before anything populates the cache, but nothing enforces that single-call assumption.

**Code assistance**

No.

**Issues**

N/A


IDisposable (approved):

Shadowghost:
Please fill out the PR template.

rlauuzo:
> Please fill out the PR template.

Okay.

crobibero (approved):

PR #17531, opened 2026-08-04
Fix database lock contentions

IDisposable:
There are frequent database locked errors thrown, or excessively long contention intervals that end up degrading the entire system performance. This is especially prevalent in the Scheduled Activities runs, like rebuilding QuickPlay images or Extracting Attachments.

Issues exist in the SQLite layer, due to the busy_timeout PRAGMA not being sent (so it defaults to 0). This results in the driver immediately returning SQLITE_BUSY instead of spinning a bit internally. Then the SQLite .Net driver just sleeps for 150ms, and tries again and then eventually timing out when CommandTimeout expires (60s). This is compounded by the fact that the thread is stalled in the thread pool for the entire time, so waits just keep stacking up.

Another issue causing database locks is that the UpdateOrInsertItems method does a deletion of the related Provider, Image, and Metadata subitems one-by-one for each existing BaseItem. This is done too early in the method, so the database takes a transaction lock on the very first BaseItem and holds it until ALL the items in the batch are computed and processed. Later the changes are actually saved with a SaveChanges() call, but all the intermediate time should have been done without any form of lock active. The fix here is to do the delete in bulk (for all updated BaseItemIds at once) and do it right before the SaveChanges().

Other minor speed ups can be had by switching to Hashset / Dictionary instead of arrays for the processing in the UpdateOrInsertItems method.

Finally, the various locking modes are actually unsafe as coded. The PessimisticLockBehavior tries to hold locks using ReadWriteLockSlim, which CANNOT be held across await calls. OptimisticLockBehavior uses ExecuteAndCapture/ExecuteAndCaptureAsync, but doesn't actually do anything with the captured exception, so the updates act as if they succeeded and the provider layer spins.

**Changes**

Add support for the `busy_timeout` PRAGMA for SQLite and pass it through the database options. By setting this to 10s (by default), the SQLite library can spin internally and not block the thread-pool. If the command timeout expires then the normal .Net SQLite Provider will kick in, but at that point we've already given the database/connection time to decongest.

Use a HashSet for the existing items in the UpdateOrInsertItems to speed up the processing by doing a Contains instead of an Any() linear search O(n) -> O(log n).

Use a Dictionary for the existing item's values to speed up the processing by doing a dictionary lookup instead of a .First() linear scan O(n) -> O(log n).

Make the deletion of the related items (BaseItemProviders, BaseItemImageInfos, BaseItemMetadataFields) to just before the save to hold the lock for far less time.

Use bulk deletion of those related items using WhereOneOrMany() against all the IDs instead of Where() _on each item_

Added unit test to ensure the same set of rows is DELETE and INSERT as before.

Add SerializeWriteLockBehavior to work like the SQLite WAL in-transaction works. Waits using SemaphoreSlime, which IS await compatible. Blocks other writes but allow reads (as they are outside the transaction). Add to the ServiceCollectionExtensions.

Add ability to set the DatabaseLockingBehaviorType via the DatabaseConfigurationOptions.

Add warning to the PessimisticLockBehavior (as it's really not safe to use in a multi-transaction environment)

Fixed OptimisticLockBehavior to actually let the lock exceptions through to the SQLite Provider library where it can be retried instead of just dropping the error on the floor.

**Code assistance**

Claude Code to build unit test and diagnose where blocks were occuring in the bulk processing of QuickPlay images. Also used to derive why the pessimistic lock mode was so broken when I tried it.

**Issues**

Pretty much every database locked error, and a lot of "my system go slow while Scheduled Tasks were running"

### DIFF
--- a/Jellyfin.Server.Implementations/Extensions/ServiceCollectionExtensions.cs
+++ b/Jellyfin.Server.Implementations/Extensions/ServiceCollectionExtensions.cs
@@ -134,6 +134,9 @@ public static class ServiceCollectionExtensions
             case DatabaseLockingBehaviorTypes.Optimistic:
                 serviceCollection.AddSingleton<IEntityFrameworkCoreLockingBehavior, OptimisticLockBehavior>();
                 break;
+            case DatabaseLockingBehaviorTypes.SerializedWrites:
+                serviceCollection.AddSingleton<IEntityFrameworkCoreLockingBehavior, SerializedWriteLockBehavior>();
+                break;
         }
 
         serviceCollection.AddPooledDbContextFactory<JellyfinDbContext>((serviceProvider, opt) =>
--- a/Jellyfin.Server.Implementations/Item/ItemPersistenceService.cs
+++ b/Jellyfin.Server.Implementations/Item/ItemPersistenceService.cs
@@ -259,6 +259,16 @@ public class ItemPersistenceService : IItemPersistenceService
         var ids = tuples.Select(f => f.Item.Id).ToArray();
         var existingItems = context.BaseItems.Where(e => ids.Contains(e.Id)).Select(f => f.Id).ToHashSet();
 
+        // Owned rows of updated items are rewritten wholesale, so clear them in one statement per
+        // table rather than three per item.
+        if (existingItems.Count > 0)
+        {
+            var updatedIds = existingItems.ToArray();
+            context.BaseItemProviders.WhereOneOrMany(updatedIds, e => e.ItemId).ExecuteDelete();
+            context.BaseItemImageInfos.WhereOneOrMany(updatedIds, e => e.ItemId).ExecuteDelete();
+            context.BaseItemMetadataFields.WhereOneOrMany(updatedIds, e => e.ItemId).ExecuteDelete();
+        }
+
         foreach (var item in tuples)
         {
             var entity = BaseItemMapper.Map(item.Item, _appHost);
@@ -397,18 +407,6 @@ public class ItemPersistenceService : IItemPersistenceService
 
                 context.AncestorIds.RemoveRange(existingAncestorIds);
             }
-        }
-
-        // Owned rows of updated items are rewritten wholesale. Issued here, immediately before the
-        // first flush, because a DELETE is what takes SQLite's single write lock: everything above
-        // is reads and change tracking, and holding the lock across it would block all other
-        // writers for the duration. Batched rather than per-item to keep this to three statements.
-        if (existingItems.Count > 0)
-        {
-            var updatedIds = existingItems.ToArray();
-            context.BaseItemProviders.WhereOneOrMany(updatedIds, e => e.ItemId).ExecuteDelete();
-            context.BaseItemImageInfos.WhereOneOrMany(updatedIds, e => e.ItemId).ExecuteDelete();
-            context.BaseItemMetadataFields.WhereOneOrMany(updatedIds, e => e.ItemId).ExecuteDelete();
         }
 
         context.SaveChanges();
--- a/Jellyfin.Server.Implementations/Item/ItemPersistenceService.cs
+++ b/Jellyfin.Server.Implementations/Item/ItemPersistenceService.cs
@@ -257,14 +257,14 @@ public class ItemPersistenceService : IItemPersistenceService
         using var transaction = context.Database.BeginTransaction();
 
         var ids = tuples.Select(f => f.Item.Id).ToArray();
-        var existingItems = context.BaseItems.Where(e => ids.Contains(e.Id)).Select(f => f.Id).ToArray();
+        var existingItems = context.BaseItems.Where(e => ids.Contains(e.Id)).Select(f => f.Id).ToHashSet();
 
         foreach (var item in tuples)
         {
             var entity = BaseItemMapper.Map(item.Item, _appHost);
             entity.TopParentId = item.TopParent?.Id;
 
-            if (!existingItems.Any(e => e == entity.Id))
+            if (!existingItems.Contains(entity.Id))
             {
                 context.BaseItems.Add(entity);
             }
--- a/Jellyfin.Server.Implementations/Item/ItemPersistenceService.cs
+++ b/Jellyfin.Server.Implementations/Item/ItemPersistenceService.cs
@@ -270,10 +270,6 @@ public class ItemPersistenceService : IItemPersistenceService
             }
             else
             {
-                context.BaseItemProviders.Where(e => e.ItemId == entity.Id).ExecuteDelete();
-                context.BaseItemImageInfos.Where(e => e.ItemId == entity.Id).ExecuteDelete();
-                context.BaseItemMetadataFields.Where(e => e.ItemId == entity.Id).ExecuteDelete();
-
                 if (entity.Images is { Count: > 0 })
                 {
                     context.BaseItemImageInfos.AddRange(entity.Images);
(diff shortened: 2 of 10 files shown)


IDisposable:
This completely eliminated the huge problem I was having since I reenabled QuickPlay images (and it's been building for a week, throwing errors all the time). Finished in couple hours ...

Shadowghost:
Can you please check out https://github.com/dotnet/efcore/issues/28135 - the efcore devs there advise against setting the busy_timout pragma

IDisposable:
> Can you please check out [dotnet/efcore#28135](https://github.com/dotnet/efcore/issues/28135) - the efcore devs there advise against setting the busy_timout pragma

That advice seems to be warning that if I set busy_timeout to a number higher than the CommandTimeout, then the command timeout wouldn't be honored because the SQLite side wouldn't be fine-grained enough. Given that I'm defaulting busy_timeout to 10s and CommandTimeout to 60s, that wouldn't apply, would it?

If we were too concerned with the fine-grain... we could make the default 500ms or something?  The big thing is that if we leave it at 0, then we tie up a threadpool for 60s

IDisposable:
Addressed CoPilot review comments.

Shadowghost:
Did some testing and some analysis (all in all I think these changes are good):

#### Moving the DELETE later trades plain SQLITE_BUSY for SQLITE_BUSY_SNAPSHOT, which busy_timeout cannot fix.

From what my local tests showed, `UpdateOrInsertItems` opens a deferred `BEGIN`, reads (taking a WAL read snapshot), and now does its first write at the very end. If any other connection commits inside that window, SQLite refuses the read -> write upgrade with `SQLITE_BUSY_SNAPSHOT` and it returns it immediately without invoking the busy handler, because waiting can't resolve it. WAL database had busy_timeout=10000 set: `upgrade FAILED after 60145ms with rc=5/517: SQLite Error 5: 'database is locked'`

`517` is `SQLITE_BUSY_SNAPSHOT`. The 60s is `Microsoft.Data.Sqlite` retrying a statement that can never succeed on a stale snapshot, until `CommandTimeout` expires. Then the whole batch throws. Master has the same issue but since we only have one query, it's not as prevalent

#### `SerializedWrites` holds the semaphore for a whole transaction and isn't reentrant across DbContexts, leading to contention becoming a hard 60 s stall.

`TransactionStarted` takes the permit and only releases at commit/rollback/connection-close. Anything else needing a write in that window waits on `_acquireTimeout = 60s`, then gives up and proceeds unsynchronized.

Since `ItemPersistenceService` keeps a transaction open for an entire batch, this turns cross-thread write contention into minute-long threadpool stalls. Seems like the 60s backstop is far too long (a few seconds would still be a deadlock backstop while degrading gracefully), and holding the permit for the transaction's lifetime rather than for the write statements undoes this PR's own work in `UpdateOrInsertItems`, where the point was that the long middle section needs no write lock.
This is a bit theoretical since none of our write patterns should run into nested write on a second context inside an open transaction. Nonetheless this would be a guaranteed 60 s stall, so we should at least document it somewhere (and if only in this comment)

#### `_holdsWriteLock` is static while `_writeLock` is per-instance.
Not really a problem right now because we only have a singleton, but still worth fixing IMO.

IDisposable:
Thanks for the extensive comments @Shadowghost. I've moved the delete back up, since it really didn't have the effect I was seeing in testing (perhaps because I was running in a different lock mode at that time). 

I DID keep use of the .WhereOneOrMany() so we delete the rows in bulk with 3 SQL statements instead of 3n.

Fixed the lock ownership static vs. member... that was just silly, sorry.

Tightened up the code comments (was a bit to chatty on review).

Shadowghost:
With the latest changes, `SerializedWrites` behaves the same as `NoLock`, so we can just not add it. Batching + busy_timeout changes should be good enough for now.

IDisposable:
Interestingly, I've seen SerializedWrite act differently after playing a video. Turns out that the jellyfin-authored Playback Reporting is setting the SQLite shared-cache mode to 1, which is a process-global operation and causes the Jellyfin core database access to start getting table/database locks again. See https://github.com/jellyfin/jellyfin-plugin-playbackreporting/blob/master/Jellyfin.Plugin.PlaybackReporting/Data/BaseSqliteRepository.cs#L54-L62

I then changed the core SQLite database connection to have the Cache mode set to SqliteCacheMode.Private instead of SqliteCacheMode.Default and suddenly Playback Reporting doesn't start the lock cascade!

Obviously, I can change that with my database.xml, but I think it's worth discussing if:

A) Maybe PlaybackReporting ought not be forcing process-global settings we have a config for elsewhere

B) Maybe we should consider making the default behavior Private, not Default for the core access... seems very suspicious to allow a plugin dictate/change the caching mode process-wide...

Also, I noticed that same block of Playback Reporting is also play fast with other SQLite options

`_ = raw.sqlite3_config(raw.SQLITE_CONFIG_MULTITHREAD, 1);`

Downgrades SQLite's threading mode from SERIALIZED to MULTITHREAD process-wide, which removes the per-connection mutex which would be genuinely dangerous if it took effect, since EF pools connections. It very likely doesn't take effect: sqlite3_config returns SQLITE_MISUSE once any connection is open, and Jellyfin's database is open long before plugins load. The return value is discarded (_ =), so a failure is silent.  This probably ought to be removed in that plug-in...

All that said, I'll be happy to rip out the SerializedWrite, but it really seems to be completely fixing my accesses now :)

Shadowghost:
That plugin needs a rewrite. Any DB test should not include it being active

JPVenson (changes requested):
Please split this PR.

You have done essentially 3 things, updating the ItemPersistanceService, introduce a new Locking behavior and introduce a new named pragma.

The changes to the ItemPersistanceService look fine at first glance allthough i would argue moving the writing part at the end not the beginning to have the transaction to be as long as possible in a non-blocking state.

The new locking behavior as ghost said has some issues and i need time to evaluate that as well.

The new option i see more critical as it as far bigger implications and can already be configured by anyone that currently may face issues so this i see more as a doc topic.

IDisposable:
> Please split this PR.

Funny story, I originally tried to use the new GitHub stacked PRs to do exactly that... and it [completely failed](https://x.com/IDisposable/status/2084623462509777122?s=20) so I caved after entering 5 different PRs on 5 branches. Sorry.

> The changes to the ItemPersistanceService look fine at first glance allthough i would argue moving the writing part at the end not the beginning to have the transaction to be as long as possible in a non-blocking state.

That's where I originally had it placed, but still was seeing locks and the [review comments made me move it back down](https://github.com/jellyfin/jellyfin/commit/f475451f1c40d505bc57f54624df27dc63d2782c)... wasn't until I figured out that the other plugin was breaking the cache mode that I realized I could put it back... will do.

> The new locking behavior as ghost said has some issues and i need time to evaluate that as well.

That's the one that really is completely opt-in and certainly should be broken out.

> The new option i see more critical as it as far bigger implications and can already be configured by anyone that currently may face issues so this i see more as a doc topic.

Not sure what you're meaning here? Are you talking about the [busy-timeout pragma](https://github.com/IDisposable/jellyfin-jellyfin/tree/feat/add-busy-timeout-pragma) or the change from Default to Private?  That's not exactly obvious how to put in, I can make a sample XML in the PR for those changes, also going to make a real quick PR for the plugin, because that's just wrong.

You didn't mention the [Dictionary](https://github.com/IDisposable/jellyfin-jellyfin/tree/fix/faster-item-values) / [Hashset](https://github.com/jellyfin/jellyfin/compare/master...IDisposable:jellyfin-jellyfin:fix/faster-existing-item-check) changes (that speeds up the processing of UpdateOrInsertItems), I assume you want that as a distinct PR as well?

IDisposable:
Split PRs all are siloed except 17555  builds on 17554 since it reused the HashSet:

#17554 (speed ups of the existing items/existing values) lookups
#17555 (move deletion to later, and do it in bulk with 3 calls instead of 3n calls) - stacks on #17554 because of shared code
#17556 (add busy-timeout pragma and configuration)
#17557 (add SerializedWriteLockBehavior support)
#17558 (switch SQLite cache mode to Private from Default to prevent Playback Reporting plugin from breaking the cache)
#17559 (fix OptimisticLockBehavior silently dropping execution failures)
#17560 (add warning to PessimisticLockBehavior log message)

IDisposable:
@JPVenson @Shadowghost I've split it up into the individual PRs. Closing this one

PR #10321, opened 2023-10-02
Add support for transcode audio file to PCM format

feilongfl:
Support transcoding audio to PCM format; this allows the client to play directly through the I2S bus without decoding. This feature is particularly useful when playing with MCUs.

**Changes**

Add support for transcode audio file to PCM format
1. set file ext name to pcm when audiocodec is pcm_*
2. add pcm params for ffmpegs

Issue #17026, opened 2026-06-05
Progressive audio transcode forces all non-Opus codecs to 48 kHz, ignoring the requested/source sample rate


esmondmissen:
### Description of the bug

When transcoding audio through the progressive `/Audio/{id}/stream` endpoint, the
output sample rate is hardcoded to **48000 Hz** for every codec except Opus. A
44.1 kHz source (e.g. a CD-rip FLAC) is silently resampled to 48 kHz even when
the request explicitly asks for `AudioSampleRate=44100`. The requested rate is
read correctly but then discarded by a clamp.

### Reproduction steps

1. Have a 44.1 kHz source audio file (e.g. 16-bit/44.1 kHz FLAC).
2. Request a transcode on the progressive endpoint with the source rate: `GET /Audio/{id}/stream?Static=false&Container=mp4&AudioCodec=aac&AudioSampleRate=44100&AudioBitRate=192000`
3. Inspect the generated ffmpeg command (or probe the output).

### What is the current _bug_ behavior?

It should return a stream with 44.1k sample rate. Returns 48k sample rate.

### What is the expected _correct_ behavior?

Return an audio stream that is 44.1k sample rate

### Jellyfin Server version

10.11.10

### Specify the build version

10.11.10

### Environment

```markdown
- OS:
- Linux Kernel:
- Virtualization:
- Clients:
- Browser:
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- CPU Model:
- GPU Model:
- Plugins:
- Reverse Proxy:
- Base URL:
- Networking:
- Jellyfin Data Storage & Filesystem:
- Media Storage & Filesystem:
- External Integrations:
```

### Jellyfin logs

```shell
[2026-06-04 20:46:18.920 -04:00] [INF] [24] MediaBrowser.MediaEncoding.Transcoding.TranscodeManager: "/usr/lib/jellyfin-ffmpeg/ffmpeg" "-analyzeduration 200M -probesize 1G  -i file:\"/data/music/Artist/Album/14 Song.flac\" -threads 0 -vn -ab 192000 -ac 2 -acodec libfdk_aac -ar 48000 -movflags empty_moov+delay_moov -id3v2_version 3 -write_id3v1 1 -y \"/cache/transcodes/344483e9ec3cd7ffff240a6a97a57f50.mp4\""
```

### FFmpeg logs

```shell

```

### Root cause

[...]

PR #17537, opened 2026-08-04
Fix PCM audio transcoding to wav returning HTTP 500 and headerless output

vdatanet:
**Changes**

Transcoding audio to PCM is broken in two ways, both from the same block in `GetProgressiveAudioFullCommandLine`. In practice `GET /Audio/{id}/stream.wav` answers **HTTP 500**, and the one PCM route that does answer 200 serves a body with no RIFF header.

```csharp
if (GetAudioEncoder(state).StartsWith("pcm_", StringComparison.Ordinal))
{
    audioTranscodeParams.Add(string.Concat("-f ", GetAudioEncoder(state).AsSpan(4)));
    audioTranscodeParams.Add("-ar " + state.BaseRequest.AudioBitRate);
}
```

**1. `-ar` is fed a bitrate, and that bitrate is optional → HTTP 500.** `-ar` is ffmpeg's *sample rate*; `AudioBitRate` is a *bitrate*, and it is `int?`. When the client sends none — the normal case for `stream.wav` — the concatenation degrades to a bare `-ar`:

```
... -acodec pcm_s16le -f s16le -ar  -ar 48000 -id3v2_version 3 ...
```

```
Expected number for ar but found: -ar
Error parsing options for output file 48000.
Error opening output files: Invalid argument
```

ffmpeg exits 234, the transcode never starts, the request 500s. The sample rate is already appended a few lines below from `state.OutputAudioSampleRate`, so this argument is removed rather than repaired.

**2. The raw muxer is forced even for a real container → no RIFF header.** `-f s16le` selects the raw-PCM muxer whatever the client asked for. A request that *does* carry a bitrate survives bug 1 — `/Audio/{id}/universal` passes `MaxStreamingBitrate` through as `AudioBitRate` — but then writes headerless samples while `Content-Type` still comes from `OutputContainer`, i.e. `audio/wav`. Clients receive `audio/wav` whose first bytes are raw samples, which on a track opening in silence reads as a run of zeros. Forcing the raw muxer was added in #10321 for I2S/MCU clients; that route is preserved by only overriding the muxer when the requested container really is raw PCM.

**3. Dead assignment.** In `StreamingHelpers.GetStreamingState`, `containerInternal = ".pcm"` is written after `state.OutputContainer` has already been read from that same variable, and is never read again — the output extension comes from `state.OutputContainer`. It has no effect and only obscures where the output container comes from, so it is dropped.

Measured against a 96 kHz FLAC source with ffmpeg 8.1.2, replaying the exact command lines the helper builds:

| Route | Before | After |
|---|---|---|
| `stream.wav`, `AudioCodec=pcm_s16le`, no bitrate | ffmpeg exit **234** → HTTP 500 | valid `RIFF`/`WAVE`, `pcm_s16le` 48 kHz stereo |
| `universal`, `Container=wav`, `MaxStreamingBitrate` set | 200, raw samples, no RIFF header | valid `RIFF`/`WAVE` |
| raw PCM route (#10321) | 384000 B headerless | 384000 B headerless (unchanged) |

Found while casting a hi-res track to a DLNA renderer with a sample-rate ceiling: the proxy needs capped PCM of known length, every `stream.wav` variant answered 500, and `universal` answered with an unusable body. The same probe against Emby returns a valid WAV.

Unit tests cover all three: no raw muxer for `wav`/`flac`/`mp3`, raw muxer still forced for the `pcm` container, and no malformed `-ar` when no bitrate is supplied. All three fail on current `master` and pass with this change.

**Code assistance**

Claude Code was used throughout: to trace the failing path from the HTTP route down to the ffmpeg argument list, to reproduce both failure modes by replaying the generated command lines against ffmpeg directly, and to draft the patch, the tests and this description. Every claim above — the exit code, the ffmpeg error text, the byte-level output of each route, and the tests failing before / passing after — was verified by running it, not inferred.

**Issues**

No existing issue found; the failing routes are described above.

### DIFF
--- a/Jellyfin.Api/Helpers/StreamingHelpers.cs
+++ b/Jellyfin.Api/Helpers/StreamingHelpers.cs
@@ -198,11 +198,6 @@ public static class StreamingHelpers
             state.OutputAudioBitrate = encodingHelper.GetAudioBitrateParam(streamingRequest.AudioBitRate, streamingRequest.AudioCodec, state.AudioStream, state.OutputAudioChannels) ?? 0;
         }
 
-        if (outputAudioCodec.StartsWith("pcm_", StringComparison.Ordinal))
-        {
-            containerInternal = ".pcm";
-        }
-
         if (state.VideoRequest is not null)
         {
             state.OutputVideoCodec = state.Request.VideoCodec;
--- a/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
+++ b/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs
@@ -7864,10 +7864,16 @@ namespace MediaBrowser.Controller.MediaEncoding
                 audioTranscodeParams.Add("-acodec " + GetAudioEncoder(state));
             }
 
-            if (GetAudioEncoder(state).StartsWith("pcm_", StringComparison.Ordinal))
+            // The pcm_* encoders emit raw samples that carry no header of their own, so the header
+            // has to come from the muxer. Only force the matching raw muxer when the client actually
+            // asked for a raw container (added in #10321 for I2S/MCU clients): applying it to every
+            // pcm_* codec also strips the RIFF header from a `stream.wav` request, which then serves
+            // headerless PCM behind an audio/wav content type.
+            var audioEncoder = GetAudioEncoder(state);
+            if (audioEncoder.StartsWith("pcm_", StringComparison.Ordinal)
+                && string.Equals(state.OutputContainer, "pcm", StringComparison.OrdinalIgnoreCase))
             {
-                audioTranscodeParams.Add(string.Concat("-f ", GetAudioEncoder(state).AsSpan(4)));
-                audioTranscodeParams.Add("-ar " + state.BaseRequest.AudioBitRate);
+                audioTranscodeParams.Add(string.Concat("-f ", audioEncoder.AsSpan(4)));
             }
 
             var sampleRate = state.OutputAudioSampleRate;
--- a/tests/Jellyfin.Controller.Tests/MediaEncoding/EncodingHelperTests.cs
+++ b/tests/Jellyfin.Controller.Tests/MediaEncoding/EncodingHelperTests.cs
@@ -223,12 +223,51 @@ public class EncodingHelperTests
         Assert.Contains("-ar " + expectedSampleRate, args, StringComparison.Ordinal);
     }
 
-    private static EncodingJobInfo BuildAudioState(string audioCodec, int requestedSampleRate)
+    [Theory]
+    [InlineData("wav")]
+    [InlineData("flac")]
+    [InlineData("mp3")]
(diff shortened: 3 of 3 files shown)


vdatanet:
Follow-up: the description above measured this by replaying the generated command lines against ffmpeg directly. Since then I ran it end to end against real servers, on both `master` and `release-10.11.z`, and the results line up.

### A/B on a server built from this branch vs. `master`

Same server, same library, same requests — only the two source files differ. Source is a 96 kHz FLAC, requesting a 48 kHz ceiling.

| Route | `master` | with this PR |
|---|---|---|
| `stream.wav` + `pcm_s16le`, no bitrate | **HTTP 500** | 200, `RIFF`, 2400150 B |
| `stream.wav` + `pcm_s16le`, uncapped | **HTTP 500** | 200, `RIFF` |
| `universal` + `Container=wav` | 200, first bytes `00000000`, `ffprobe`: `Invalid data found when processing input` | 200, `RIFF`, 4800174 B |
| `stream.flac` (control) | 200, `fLaC` | 200, `fLaC` |

On `master` the server log shows the malformed command line and the exit code directly:

```
[INF] TranscodeManager: /opt/homebrew/bin/ffmpeg -analyzeduration 200M -probesize 1G  -i file:"…/01 - Hi-Res 96kHz.flac"
      -threads 0 -vn -ab 128000 -ac 1 -acodec pcm_s16le -f s16le -ar  -ar 48000 -id3v2_version 3 -write_id3v1 1 -y "…/….wav"
[ERR] TranscodeManager: FFmpeg exited with code 234
MediaBrowser.Common.FfmpegException: FFmpeg exited with code 234
```

Note the `-ar  -ar 48000`. Without an `AudioSampleRate` the tail is worse — `-ar  -id3v2_version 3` — since the bare `-ar` then swallows the next token.

### Also confirmed on 10.11.z with a real library

I backported the same two-file change onto `release-10.11.z` (it applies cleanly; only the new test file has no counterpart there), built a 10.11.11 image, and pointed it at a real 62-track hi-res library.

Requesting a 96 kHz / 24 bit FLAC as capped PCM — the request that returns 500 on stock 10.11.11 — now returns a valid WAV:

```
codec_name=pcm_s16le
sample_rate=48000
channels=2
bits_per_sample=16
format_name=wav
duration=106.945104
```

20533712 bytes, which is exactly `106.945 × 48000 × 2ch × 2B` plus header, and `volumedetect` reports mean −16.1 dB / peak −0.2 dB, so it is full-length real audio rather than a header over silence.

### Unrelated, but noticed while backporting

`release-10.11.z` still has the sample-rate bug fixed on `master` by #17026 — the `!opus` branch just below applies opus's rate clamping to every other codec, so a requested 44100 comes out as 48000. Not touched by this PR; flagging it in case it matters for a backport decision.

gnattu (approved):

Shadowghost (approved):

Bond-009 (approved):

PR #17539, opened 2026-08-04
Cancel embedded subtitle extraction when the request is aborted

### LINKED ISSUE #17499: Embedded subtitle extraction reads the entire remote file and isn't cancelled when playback stops (huge bandwidth waste for .strm/HTTP and rclone sources)
### Description of the bug

When a title is played, Jellyfin extracts embedded text subtitle tracks via
`SubtitleEncoder` (`ffmpeg -i <source> -map 0:x -an -vn -c:s copy <out>.srt`).
For a container like MKV, subtitle packets are interleaved across the whole
duration, so this ffmpeg reads the **entire source file sequentially to EOF**.

Two problems compound for **remote sources** (a `.strm` pointing at an HTTP URL,
or an rclone-mounted library):

1. The extraction downloads the **whole file** over the network (e.g. a 16 GB
   4K remux), regardless of how much of the title was actually watched.
2. The extraction job is **not cancelled when the triggering playback stops**
   (user exits the player / switches titles). It keeps reading to completion in
   the background, continuing to consume bandwidth long after playback ended.

This makes lightweight sampling of a remote title effectively download the full
file, and switching between titles multiplies the wasted bandwidth.

### Reproduction steps

1. Add a movie as a `.strm` file whose content is an `http://` URL served by a
   simple range-capable HTTP server (or mount the library via rclone). Use a
   large file (4K remux) with embedded text subtitle tracks.
2. On the HTTP source, log byte reads / range requests.
3. Start playback in any client, let it play a few seconds, then stop/close the
   player.
4. Observe on the source: an open-ended `GET` (e.g. `Range: bytes=0-`) keeps
   reading at full line speed to the end of the file, well after playback ended.
5. Confirm the reader is the subtitle-extraction ffmpeg (cmdline
   `ffmpeg -i <url> -map 0:x -c:s copy ....srt`); killing that ffmpeg stops the
   read immediately.

Embedded-subtitle extraction reads the entire remote file and continues after
the associated playback session has stopped.

Measured on my setup (Jellyfin 10.11.11, `.strm` → HTTP range source, client
"Jellyfin Web 10.11.11"):

- `21:44:38` — `SubtitleEncoder` starts `ffmpeg ... -map 0:4..0:7 -c:s copy`
  against the source URL.
- `21:44:44` — SessionManager logs **"Playback stopped reported by app Jellyfin
  Web 10.11.11"** (I closed the player).
- `21:44:44 → 21:51:59` — the extraction ffmpeg keeps reading the source at full
  line speed for **~7 more minutes** (a single open-ended `GET`), transferring
  **~7.4 GB** (≈46% of a 16 GB file) with no playback session active.
- The source's transfer rate dropped to **0** the instant I killed that ffmpeg.

- Subtitle extraction should be tied to the playback lifecycle (cancelled when
  the triggering session stops), and/or
- For remote/HTTP/rclone sources it should be possible to disable proactive
  embedded-subtitle extraction, so that sampling/stopping a title does not force
  a full-file download.

### Jellyfin Server version

10.11.11

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.11

### Environment

```markdown
- Server: Jellyfin 10.11.11 (Docker, linux/amd64)
- FFmpeg: ffmpeg 7.1.4-Jellyfin
- Media source: .strm files pointing at an http:// range server (custom); large 4K HDR10+ MKV remuxes with embedded subtitle tracks
- Client: Jellyfin Web 10.11.11 (browser)
```

### Jellyfin logs

```shell
# Subtitle extraction starts:
[21:44:38] [INF] MediaBrowser.MediaEncoding.Subtitles.SubtitleEncoder:
/usr/lib/jellyfin-ffmpeg/ffmpeg -i "http://torrentstreamer:8080/stream/b5e7a67446d53ea25a15c3cea6c2f4e5951b1b56/0/Scary%20Movie%20%282026%29%20Extended%20Version%20WEB-DL%202160p%20HDR10Plus.mkv" \
  -map 0:4 -an -vn -c:s copy "/config/data/subtitles/0d/0d1403c5-567a-3819-3d5d-3b1e96738042/4.srt" \
  -map 0:5 -an -vn -c:s copy ".../5.srt" \
  -map 0:6 -an -vn -c:s copy ".../6.srt" \
  -map 0:7 -an -vn -c:s copy ".../7.srt"

# Playback stops ~6s later, but extraction keeps running:
[21:44:44] [INF] Emby.Server.Implementations.Session.SessionManager:
Playback stopped reported by app Jellyfin Web 10.11.11 playing <title>. Stopped at 1263801 ms

# Extraction finishes reading the file ~7 minutes after playback stopped:
[21:51:59] [INF] MediaBrowser.MediaEncoding.Subtitles.SubtitleEncoder:
ffmpeg subtitle extraction completed for "http://torrentstreamer:8080/stream/.../Scary%20Movie...mkv" to .../7.srt
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

_No response_

### Related issues

- #13029 / PR #13033 fixed "continues reading after HTTP request cancelled" but
  only for `SymlinkFollowingPhysicalFileResultExecutor` (the physical-file
  stream/download path); it does not cover the subtitle-extraction path.
- #14588 documents subtitle extraction reading big files over an rclone remote
  (same `SubtitleEncoder ... -c:s copy` mechanism).

vavallee:
**Changes**
`SubtitleController.EncodeSubtitles` passed `CancellationToken.None` to `ISubtitleEncoder.GetSubtitles`, so an extraction started to serve a subtitle request could not be stopped. Extracting an embedded text track reads the source container to EOF (subtitle packets are interleaved across the whole duration), so closing the player or switching titles left the ffmpeg process reading to completion; for a remote source (`.strm` over HTTP, rclone mount) that means downloading the entire file. The reporter measured ~7.4 GB of a 16 GB remux transferred over seven minutes after playback had already stopped, with the transfer dropping to zero the instant that ffmpeg was killed by hand.

The cancellation machinery below the controller is already complete: `RunSubtitleExtractionProcess` kills ffmpeg on cancellation and returns exit code -1, and `ExtractSubtitlesForFile` then deletes the partial output and rethrows, so no truncated `.srt` is left behind in the subtitle cache. Only the token was missing. This passes `HttpContext.RequestAborted` instead.

Behaviour change worth calling out: an extraction is now abandoned if the client that asked for it goes away, so a second client waiting on the same output (they share an `AsyncKeyedLocker` key) will restart it rather than inherit the finished file. That is the trade the issue asks for — sampling a title should not force a full-file download.

Deliberately out of scope: the remote-subtitle search and download endpoints in the same controller still use `CancellationToken.None`. They write into the media folder and are not bound to playback lifetime, so cancelling them on request abort is a separate decision.

**Code assistance**
Written with Claude Code (root-cause confirmation against the code, and implementation); reviewed, validated, and submitted by me.

**Issues**
Fixes #17499


tests/Jellyfin.Api.Tests/Controllers/SubtitleControllerTests.cs
```diff
@@ -0,0 +1,100 @@
...
+using Moq;
+using Xunit;
+
+namespace Jellyfin.Api.Tests.Controllers;
+
+// Extracting an embedded subtitle track reads the source container to EOF, which for a remote
+// source (.strm/HTTP, rclone mount) means downloading the whole file. These tests pin down that
+// the request's cancellation token reaches the encoder, so the extraction dies with the request
+// instead of running on after the player was closed.
+public sealed class SubtitleControllerTests
```
cvium: Can you tell me the point of these tests?

vavallee: They pin that RequestAborted reaches the encoder instead of CancellationToken.None

If you'd rather not carry a mock-heavy controller test for that, I'm fine dropping them

cvium: Test code is often ugly and slightly hacky, so that's not really the problem for me. It's that the tests don't really validate that the cancellationtoken is used and will cancel the request properly. If someone down the line creates a [linked token](https://learn.microsoft.com/en-us/dotnet/api/system.threading.cancellationtokensource.createlinkedtokensource?view=net-10.0) somewhere in this call stack, this test will fail, but the behaviour might very well be intact. Similarly, the cancellation behaviour could be broken at any point without this test failing.

vavallee: Agreed, I didn't consider those cases, removing the tests

LeifWebber:
Thanks for working on this. I added a concrete Android TV/local-MKV reproduction to https://github.com/jellyfin/jellyfin/issues/17499#issuecomment-5447307448. It shows a downstream effect that may not be covered by this one-file change: after the client stopped, embedded subtitle extraction continued, and when it finished ~154.5 seconds later a stale ASS HLS request started a full `subtitles=.../7.ass + overlay_qsv` transcode. That ghost job then ran until the HLS kill timer removed it.

Could you confirm whether this PR is expected to cancel the HLS burn-in path as well?

The patch changes `SubtitleController.EncodeSubtitles` to use `HttpContext.RequestAborted`, but the burn-in command construction still calls:

`EncodingHelper.GetSubtitleFilePath(..., CancellationToken.None)`

and `DynamicHlsController.GetDynamicSegment` creates its own `CancellationTokenSource` rather than linking it to the aborted request/play session. If an HLS request is already waiting on the same subtitle-cache lock, cancelling the subtitle HTTP request may allow the stale HLS request to acquire the lock, restart/finish extraction, and register/start ffmpeg after the stop event has already passed.

A behavior-level regression test would be especially useful here:

1. Start an ASS HLS request whose embedded subtitle extraction is still running.
2. Abort/stop that playback before extraction completes.
3. Assert the extraction is cancelled and, critically, that no `TranscodingJob` or ffmpeg process is registered/started later for the stale request.

This is not just bandwidth waste: in the reproduced case the ghost burn-in overlapped the active non-burn stream and reduced each 4K transcode from about 1.18x single-job speed to roughly 0.56x.

PR #17455, opened 2026-07-27
Reevaluate pending migrations after each one instead of per stage

Shadowghost:
`JellyfinMigrationService.MigrateStepAsync` computed the pending migration list once and then ran that whole list to completion before re-evaluating, despite its own comment saying the state should be reevaluated because "migrations may alter the migration state".
Since `IMigrator.MigrateAsync` treats its argument as the state to end up in, it reverts everything applied after that target, and backported migrations carry older timestamps than the ones already on master, so on a `10.11.z -> master` upgrade the first migration reverts them and invalidates every remaining entry in the precomputed list.

`AddNormalizedUsername` and `AddUniqueNormalizedUsernameIndex` were swept into a later target's batch without the `UpdateNormalizedUsername` code migration running between them (its history row is deleted by  `AddNormalizedUsername.Down()`), so the unique index was created while every row still held the empty default, failing on any install with more than one user.

**Changes**
* Make the loop run a single migration per iteration and recompute the pending set afterwards, so a reverted migration is picked straight back up and re-applied in the correct interleaved order.
* Add a guard against a migration that runs without recording itself (which would now loop forever), sorts with `StringComparer.Ordinal` to match how EF orders migration ids, and fixes the "There are N migrations" line to count database migrations instead of only code migrations.

**Code assistance**
Claude Code and Opus 5 were used to write some in-memory temporary tests to check the behavior causes

**Issues**
Regression surfaced by #17422 but caused by backported migrations from 10.11.z

Fixes #17458

PR #17540, opened 2026-08-05
Improve season folder resolution: series-prefixed names and empty
  folder handling

AvonWilliams:
## What I found

My season folders are named like "Solar Opposites - Season 1" (Sonarr default). After digging into the code, I noticed SeasonPathParser cant parse these because the season keyword isnt at the start of the cleaned folder name.

This seems to cause some problems:
- Physical seasons end up with IndexNumber = null
- Metadata providers then create virtual seasons that compete with them  
- Episodes get linked to the wrong season
- On scans, things get removed and recreated, triggering webhooks

I also noticed empty folders like an unused "Season 1" sitting next to a real "Invasion - Season 1" folder will create phantom seasons, because the empty-folder check only ran when name parsing failed.

## What I changed

Two small changes:

In SeasonPathParser.cs: after the anchored regex checks, look for season keywords anywhere in the name and try parsing from there. It skips matches with a digit right before them (to avoid episode patterns like "Episode 1 Season 2"), and itll try to strip trailing years so "Season 7 (2023)" comes out as 7 not 72023.

In SeasonResolver.cs: moved the empty-folder check earlier so it runs for all folders, not just ones that failed parsing. Uses a quick top-level check first, then one level of subdirectories (to skip .trickplay/ image folders), and only does full recursive as a last resort.

## Testing

On my server (12.0-rc4, ~250 eps):
- Ran 5+ full scans across rc3 and rc4, saw no more remove/recreate cycles
- Solar Opposites went from 4 real + 4 duplicate virtual seasons to just 4 real ones
- Invasion phantom seasons cleaned up
- Legitimate virtuals for undownloaded content stayed intact
- Scan times unchanged (~8s)

Unit tests: added 14 new cases covering prefixed names, years, international keywords, etc. Full naming suite (701) and SeasonResolver tests (10) all pass, no regressions.

Patch applies cleanly to rc1 through rc4 — the affected files didnt change between them.

I also tested on rc4 which has a LinkedChildrenLoaded guard on Folder (#17455) — seems to work fine alongside these changes, though I havent dug into whether there is any actual interaction.

## Caveats

This is my first contribution here. I tested as thoroughly as I could on my own server but would appreciate a closer look, specially around the year sanitization approach and the directory scanning.

Note: I used Claude Code to help dig through the codebase and iterate on the fix, but the testing was done on my own server and these words are mine. Per the LLM policy, happy to adjust anything that needs changing.

This problem alone was fundamentally breaking my 12.0RC3 install. Every day more episodes would simply disappear,  it  quickly became unusable. So far, this fix seems to be working. I've tested it to the best of my ability with my limited skills and library, on multiple server builds and 12.0 versions.  So far, it seems to be solving the problem. 

Of course I welcome feedback, criticism and suggestions. My goal in doing this is not only to fix it for myself, but to fix it for the community. I urge the community to test on sacrificial servers, and please report back any bugs / fixes.

### DIFF
--- a/Emby.Naming/TV/SeasonPathParser.cs
+++ b/Emby.Naming/TV/SeasonPathParser.cs
@@ -193,10 +193,11 @@ namespace Emby.Naming.TV
                 if (suffix >= 1900 && suffix <= 2099)
                 {
                     var prefix = seasonNumber / 10000;
-                    // Only strip if the remaining prefix is a reasonable season number (1-99)
-                    if (prefix > 0 && prefix < 100)
+                    // Only strip if the remaining prefix is a reasonable season number (1-99).
+                    // prefix is always >= 1 here since seasonNumber >= 10000.
+                    if (prefix < 100)
                     {
-                        return (int)prefix;
+                        return prefix;
                     }
                 }
             }
--- a/Emby.Naming/TV/SeasonPathParser.cs
+++ b/Emby.Naming/TV/SeasonPathParser.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Globalization;
 using System.IO;
+using System.Linq;
 using System.Text.RegularExpressions;
 
 namespace Emby.Naming.TV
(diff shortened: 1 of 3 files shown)


AvonWilliams:
## Update: tested on 12.0-rc4

Rebuilt and deployed against the `v12.0-rc4` tag — patch applies cleanly (no changes to the affected files between rc1 and rc4). Library scan on a ~250-episode library completed without issues; season linkages remained stable.

One observation: rc4 includes a `LinkedChildrenLoaded` flag on `Folder` (#17455) that guards against empty `LinkedChildren` arrays silently dropping child links during persistence. From my testing, this appears to complement the parser fix — the `SeasonPathParser` change ensures physical seasons get correct `IndexNumber` values, and the persistence guard may help those linkages survive save cycles that previously could have lost them.

This is just from my own testing on one server — I would welcome review from someone more familiar with the persistence layer to confirm there is no unintended interaction.

Shadowghost:
Please check https://jellyfin.org/docs/general/contributing/llm-policies/ and adhere to it.

AvonWilliams:
fixed  two easy issues that were flagegd — the `> 0` was dead code since the division always gives at least 1, and the cast was pointless.

AvonWilliams:
Got a bit keen and pushed before i had cleaned everything up — that foreach issue was in the report too. All three should be sorted now.

PR #17541, opened 2026-08-05
Fix by-name endpoints reporting TotalRecordCount=0 next to a populated Items array

vdatanet:
**Changes**

The by-name endpoints report `TotalRecordCount = 0` while returning a populated `Items` array, whenever the request carries no explicit `limit`.

`GetItemValues` in `BaseItemRepository.ByName.cs` is the shared path behind `/Artists`, `/Artists/AlbumArtists`, `/Genres`, `/MusicGenres` and `/Studios`. Its first act is to turn the count off:

```csharp
if (!filter.Limit.HasValue)
{
    filter.EnableTotalRecordCount = false;
}
```

so the count never reaches the result a hundred lines below:

```csharp
if (filter.EnableTotalRecordCount)
{
    result.TotalRecordCount = representativeIds.Count;
}
```

A client that pages on the reported total — the contract every other list endpoint honours — reads a populated response as an empty library. `/Items` and `/Persons` do not go through `GetItemValues` and report the count correctly, which is what makes the inconsistency observable without reading any source.

Measured against `master` (12.0.0) with a 62-track music library:

| Request | Before | After |
|---|---|---|
| `GET /Artists?UserId=…` | `TotalRecordCount=0`, `Items=7` | `TotalRecordCount=7`, `Items=7` |
| `GET /MusicGenres?UserId=…` | `TotalRecordCount=0`, `Items=38` | `TotalRecordCount=38`, `Items=38` |
| `GET /Studios?UserId=…` | `TotalRecordCount=0`, `Items=6` | `TotalRecordCount=6`, `Items=6` |
| `GET /Artists/AlbumArtists?UserId=…` | `TotalRecordCount=0`, `Items=5` | `TotalRecordCount=5`, `Items=5` |
| `GET /Artists?UserId=…&limit=100` | `TotalRecordCount=7` | unchanged |
| `GET /Items?…` (control, different path) | correct | unchanged |

The `limit` column is the tell: the same query reports the total correctly as soon as a limit is present, which is exactly what the removed block keyed on.

**Removing the block costs nothing.** `representativeIds` is materialised into a `List<Guid>` regardless a few lines further down, so the count was already computed and in hand — no extra query, no extra pass. Callers that genuinely want to skip counting still can: `EnableTotalRecordCount = false` is honoured as before, and that is covered by a test.

The block additionally mutated the *caller's* query object rather than a local copy, so an `InternalItemsQuery` instance reused across calls silently lost its total after the first limitless one. Also covered by a test.

Four unit tests are added in `BaseItemRepositoryByNameTotalCountTests`, against an in-memory SQLite context seeded with artists credited on songs (the item-value join the by-name endpoints rely on). Two of them fail on `master` — the limitless count and the query mutation — and the other two pass both before and after, guarding the paging case and the opt-out.

**Code assistance**

Claude Code was used to locate the failing path, reproduce it against a server built from `master`, write the fix and the tests, and draft this description. Everything stated above was verified by running it: the before/after numbers come from HTTP responses against two servers built from this branch with and without the change, and the "fails on master" claim comes from running the new tests against the unmodified file. Nothing here is inferred.

**Issues**

No existing issue found; the failing requests are listed above.

### DIFF
--- a/Jellyfin.Server.Implementations/Item/BaseItemRepository.ByName.cs
+++ b/Jellyfin.Server.Implementations/Item/BaseItemRepository.ByName.cs
@@ -144,11 +144,6 @@ public sealed partial class BaseItemRepository
     {
         ArgumentNullException.ThrowIfNull(filter);
 
-        if (!filter.Limit.HasValue)
-        {
-            filter.EnableTotalRecordCount = false;
-        }
-
         using var context = _dbProvider.CreateDbContext();
 
         var innerQueryFilter = TranslateQuery(context.BaseItems.Where(e => e.Id != EF.Constant(PlaceholderId)), context, new InternalItemsQuery(filter.User)
--- a/tests/Jellyfin.Server.Implementations.Tests/Item/BaseItemRepositoryByNameTotalCountTests.cs
+++ b/tests/Jellyfin.Server.Implementations.Tests/Item/BaseItemRepositoryByNameTotalCountTests.cs
@@ -1,0 +1,199 @@
+using System;
+using System.Linq;
+using Emby.Server.Implementations.Data;
+using Jellyfin.Database.Implementations;
+using Jellyfin.Database.Implementations.Entities;
+using Jellyfin.Database.Implementations.Locking;
+using Jellyfin.Database.Providers.Sqlite;
+using Jellyfin.Server.Implementations.Item;
+using MediaBrowser.Controller;
+using MediaBrowser.Controller.Configuration;
+using MediaBrowser.Controller.Entities;
+using MediaBrowser.Model.Configuration;
+using Microsoft.Data.Sqlite;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.Logging.Abstractions;
+using Moq;
+using Xunit;
+using BaseItemKind = Jellyfin.Data.Enums.BaseItemKind;
+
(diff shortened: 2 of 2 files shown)


Shadowghost:
Should already be fixed by #17466

vdatanet:
I built and ran `Shadowghost:fix-byname-queries` at `d3c7aa6bd2` against a music library to check, and the by-name totals are still zero there:

```
/Artists?UserId=…                  -> TotalRecordCount=0  Items=7
/Artists?UserId=…&limit=500        -> TotalRecordCount=7  Items=7
/Artists/AlbumArtists?UserId=…     -> TotalRecordCount=0  Items=6
/Persons?UserId=…                  -> TotalRecordCount=4  Items=4   (correct)
```

`BaseItemRepository.ByName.cs` isn't among the files #17466 touches, and the block is still at line 147 on that branch. The `EnableTotalRecordCount` work in #17466 is in `PeopleRepository.cs`, which is the `/Persons` path — the one that already reported the total correctly, and still does.

The two don't overlap on any file, so this should merge cleanly either way. Happy to close this if you'd rather fold the one-block change into #17466.

IDisposable (approved):

Shadowghost (approved):

crobibero (approved):

PR #17544, opened 2026-08-05
Let /Audio/{id}/universal reuse a transcode instead of re-encoding per request

vdatanet:
**Changes**

`GET /Audio/{itemId}/universal` re-encodes the whole track on every request, even when the request is byte-for-byte identical to the previous one.

The endpoint accepts no `playSessionId`. It uses the one `MediaInfoHelper.GetPlaybackInfo` produces, which is a fresh GUID per call:

```csharp
result.PlaySessionId = Guid.NewGuid().ToString("N", CultureInfo.InvariantCulture);
```

and hands it to the `StreamingRequestDto`. Since the transcode's output path is

```csharp
var data = $"{state.MediaPath}-{state.UserAgent}-{deviceId!}-{playSessionId!}";
```

no two requests ever land on the same path, so the completed transcode is never reused. `stream.*` already takes a `playSessionId` and reuses it; `universal` is the odd one out.

Measured on `master`, three identical requests for one track, counting ffmpeg invocations in the server log:

| Request | Before | After |
|---|---|---|
| `/Audio/{id}/universal` ×3, same `playSessionId` | **3 ffmpeg invocations** | **1** |
| `/Audio/{id}/universal` ×3, no `playSessionId` | 3 | 3 (unchanged) |
| `/Audio/{id}/stream.mp3` ×3, same `playSessionId` (control) | 1 | 1 |

The change is additive: the parameter is optional, and when it is absent a fresh id is still generated per request, so no existing caller changes behaviour.

**Side effect worth naming, since it is how this was found.** The Xing/LAME header of an MP3 transcode is written blank on the first pass and backfilled once the encoder knows the frame counts. Because every `universal` request started a new transcode, clients received `frames=0 delay=0 padding=0` on *every* request and had no encoder delay or padding to trim, so tracks played with the priming samples at the head and the padding at the tail. With the transcode reused, the completed header comes back from the second request on:

```
before:  Xing  frames=0     delay=0    padding=0
after:   Info  frames=9881  delay=576  padding=870
```

which is what `stream.mp3` already returns today.

**On tests:** there is no unit test here, and I would rather say so than pad the PR. `UniversalAudioController` takes `MediaInfoHelper`, `AudioHelper` and `DynamicHlsHelper` as concrete classes with non-virtual methods, so it cannot be meaningfully faked, and a test that proved anything about transcode reuse would need the whole transcode stack and real media. The verification is the measurement above: a server built from this branch, one built from `master`, the same library and the same requests, counting ffmpeg invocations in the log. Happy to add a test if a reviewer sees a seam I have missed.

**Code assistance**

Claude Code was used to trace the path from the endpoint to the output-path hash, to run the before/after measurements, and to draft this description. Everything above was verified by executing it — the invocation counts come from server logs on two builds, and the header values from parsing the returned MP3s. Nothing is inferred.

**Issues**

No existing issue found.

### DIFF
--- a/Jellyfin.Api/Controllers/UniversalAudioController.cs
+++ b/Jellyfin.Api/Controllers/UniversalAudioController.cs
@@ -83,6 +83,7 @@ public class UniversalAudioController : BaseJellyfinApiController
     /// <param name="maxAudioSampleRate">Optional. The maximum audio sample rate.</param>
     /// <param name="maxAudioBitDepth">Optional. The maximum audio bit depth.</param>
     /// <param name="enableRemoteMedia">Optional. Whether to enable remote media.</param>
+    /// <param name="playSessionId">Optional. The play session id. Repeating it across requests lets the server reuse the transcode it already produced instead of starting a new one; omitted, a fresh id is generated per request.</param>
     /// <param name="enableAudioVbrEncoding">Optional. Whether to enable Audio Encoding.</param>
     /// <param name="enableRedirection">Whether to enable redirection. Defaults to true.</param>
     /// <response code="200">Audio stream returned.</response>
@@ -113,6 +114,7 @@ public class UniversalAudioController : BaseJellyfinApiController
         [FromQuery] int? maxAudioSampleRate,
         [FromQuery] int? maxAudioBitDepth,
         [FromQuery] bool? enableRemoteMedia,
+        [FromQuery] string? playSessionId = null,
         [FromQuery] bool enableAudioVbrEncoding = true,
         [FromQuery] bool enableRedirection = true)
     {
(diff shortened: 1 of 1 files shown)


gnattu (changes requested):
We explicitly removed the arbitrary session id from the api due to security concerns.

vdatanet:
Understood, thanks — I'll drop the parameter; I wasn't aware of that decision.

Would a server-derived key be acceptable instead, or is that also off the table? What I was after is only that two identical requests reuse one transcode — the session id was just the handle that happened to be in the output path. If the transcode's identity were derived from the request's own shape rather than a fresh GUID, the reuse would work with nothing supplied by the client.

For context on why I went looking: three identical `universal` requests currently produce three full re-encodes, measured on master by counting ffmpeg invocations.

Happy to close this if it's not a direction you want to take. And if you can point me at the reasoning behind removing the session id, I'll make sure not to walk into it again.

vdatanet:
Closing this — the security concern is reason enough, and I don't want to leave a rejected PR sitting open.

For anyone who finds this later: the underlying observation still holds, that three identical `universal` requests produce three full re-encodes because the play session id in the transcode's output path is regenerated per call. If it's ever worth addressing, deriving that key server-side from the request rather than accepting one from the client would get the reuse without reintroducing what was removed. Not pursuing it here.

On our side there's no need to wait: `stream.{container}` already honours a play session id and reuses the transcode, so we'll move to that route.

Thanks for the quick review.

PR #17548, opened 2026-08-05
Fix auto-created collections losing custom name on rescan

### LINKED ISSUE #15693: Locking Collection Metadata Not Preventing Changes

### Description of the bug

When editing collection metadata, like the name of the collection, and then locking it for future changes, the XML file for the collection is updated and shows as locked, but the lock is lost and the XML file is then overwritten the next time the library is scanned.  Confirmed this may be a bug in Element troubleshooting room.  Disabling the "automatically add to collection" function in the library resolves this, but just as automatically adding metadata for movies and then locking the metadata works, so should the same for collections metadata.

NOTE: I'm running 10.11.4, but that wasn't an option in the Server version dropdown below

### Reproduction steps

1. Change the title of any auto-created collection
2. Lock the metadata against future changes
2a. _Optional_ Verify the metadata is locked in the collection's XML file
3. Rescan the library
4. The change is lost and the collection title has returned to the default name

Collection titles are losing their changes after the library is rescanned.

Locking the metadata for the collection should prevent the metadata from being overwritten when the library is rescanned (even if the library is set to auto-add to collections), _OR_ if this is working as designed, the ability to lock the collection's metadata should be removed since it doesn't seem to actually lock anything, to avoid future confusion.

### Jellyfin Server version

10.11.3

### Specify commit id

_No response_

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.4

### Environment

```markdown
- OS: Windows 11
- Linux Kernel:
- Virtualization:
- Clients: Browser, Windows app
- Browser: Firefox 145.0.2
- FFmpeg Version:
- Playback Method:
- Hardware Acceleration:
- GPU Model:
- Plugins: OpenSubs
- Reverse Proxy:
- Base URL:
- Networking:
- Jellyfin Data Storage: local SSD
- Media Storage: External HDD
- External Integrations:
```

### Jellyfin logs

```shell
No relevant logs
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

Image

Image

### Additional information

_No response_

kyle678:
**Changes**

Auto-created collections lost their custom name after a library rescan when
"Automatically add to collection" was enabled. `CollectionPostScanTask`
identified existing collections by their mutable `Name`, so when a user
renamed an auto-created collection the lookup missed (the movies still carried
their original `CollectionName`) and the task recreated the collection under
its default name, discarding the rename regardless of whether metadata was
locked. This changes the task to match existing box sets by the stable TMDB
collection id (falling back to the name for legacy box sets without an id), and
stamps newly created box sets with the id so future scans can match them after
a rename. Adds unit tests covering both the renamed-collection and
new-collection paths.

Note: collections whose movies have no TMDB collection id still fall back to
name matching, so a rename of those isn't preserved yet. A membership-based
identity would be the fuller fix and is left as a potential follow-up.

**Code assistance**

Claude Code Opus 4.8 was used to navigate and explain the codebase. It was
also used to draft the unit tests.

**Issues**

Fixes #15693

### DIFF
--- a/Emby.Server.Implementations/Library/Validators/CollectionPostScanTask.cs
+++ b/Emby.Server.Implementations/Library/Validators/CollectionPostScanTask.cs
@@ -84,7 +84,9 @@ public class CollectionPostScanTask : ILibraryPostScanTask
 
                     var tmdbCollectionId = movie.TryGetProviderId(MetadataProvider.TmdbCollection, out var id) ? id : null;
 
-                    var key = string.IsNullOrEmpty(tmdbCollectionId) ? movie.CollectionName : tmdbCollectionId;
+                    var key = string.IsNullOrEmpty(tmdbCollectionId)
+                        ? "name=" + movie.CollectionName
+                        : "id=" + tmdbCollectionId;
 
                     if (!collectionGroups.TryGetValue(key, out var group))
                     {
--- a/Emby.Server.Implementations/Library/Validators/CollectionPostScanTask.cs
+++ b/Emby.Server.Implementations/Library/Validators/CollectionPostScanTask.cs
@@ -9,6 +9,7 @@ using MediaBrowser.Controller.Collections;
 using MediaBrowser.Controller.Entities;
 using MediaBrowser.Controller.Entities.Movies;
 using MediaBrowser.Controller.Library;
+using MediaBrowser.Model.Entities;
 using Microsoft.Extensions.Logging;
 
 namespace Emby.Server.Implementations.Library.Validators;
@@ -46,7 +47,7 @@ public class CollectionPostScanTask : ILibraryPostScanTask
     /// <returns>Task.</returns>
     public async Task Run(IProgress<double> progress, CancellationToken cancellationToken)
     {
-        var collectionNameMoviesMap = new Dictionary<string, HashSet<Guid>>();
+        var collectionGroups = new Dictionary<string, CollectionGroup>();
 
         foreach (var library in _libraryManager.RootFolder.Children)
         {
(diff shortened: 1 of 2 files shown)


Emby.Server.Implementations/Library/Validators/CollectionPostScanTask.cs
```diff
@@ -74,17 +75,24 @@ public async Task Run(IProgress<double> progress, CancellationToken cancellation
...
-                        else
-                        {
-                            collectionNameMoviesMap[movie.CollectionName] = new HashSet<Guid> { movie.Id };
-                        }
+                        continue;
+                    }
+
+                    var tmdbCollectionId = movie.TryGetProviderId(MetadataProvider.TmdbCollection, out var id) ? id : null;
+
+                    var key = string.IsNullOrEmpty(tmdbCollectionId) ? movie.CollectionName : tmdbCollectionId;
```
IDisposable: What if some movie.CollectionName just happens to match some tmdbCollectionId? Seems like we should prefix a disambiguation to the key:

```csharp
var key = string.IsNullOrEmpty(tmdbCollectionId) 
 ? "name=" + movie.CollectionName 
 : "id=" + tmdbCollectionId;
```

IDisposable (changes requested):

Emby.Server.Implementations/Library/Validators/CollectionPostScanTask.cs
kyle678: Done in commit 1750c0c, the key is now prefixed with `name=`/`id=` to disambiguate the two namespaces.

Shadowghost:
You'd need to modify the path if we have name-based collissions because collections metadata directories are name-unique. Having two collections with the same name would result in them using the same metadata directory.

We're also doing two linear scans over all box sets per group, better would be tovuild a dictionary once (id → box set, name → box set) and then use that in th epost scan task.

PR #17556, opened 2026-08-06
Add busy_timeout pragma support

IDisposable:
Add support for the `busy_timeout` PRAGMA for SQLite and pass it through the database options. By setting this to 5s (by default), the SQLite library can spin internally and not block the thread-pool. If the command timeout expires then the normal .Net SQLite Provider will kick in, but at that point we've already given the database/connection time to decongest.

**Changes**

- Add busy timeout support to the PragmaConnectionInterceptor and emit the
PRAGMA if it is set to non-zero.
- Add busy timeout loading from the database.xml configuration. Defaults to 5 seconds if not set
- Add default command-timeout constant (same 60s)

**Code assistance**

**Issues**

### DIFF
--- a/src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/PragmaConnectionInterceptor.cs
+++ b/src/Jellyfin.Database/Jellyfin.Database.Providers.Sqlite/PragmaConnectionInterceptor.cs
@@ -20,6 +20,7 @@ public class PragmaConnectionInterceptor : DbConnectionInterceptor
     private readonly int? _journalSizeLimit;
     private readonly int _tempStoreMode;
     private readonly int _syncMode;
+    private readonly int _busyTimeout;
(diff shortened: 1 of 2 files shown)


cvium:
Won't busy_timeout still block a thread? SQLite is fully synchronous afaik.

IDisposable:
> Won't busy_timeout still block a thread? SQLite is fully synchronous afaik.

Briefly, yes, default is ~5 seconds~ 100ms, fully settable in database.xml.  If desired, I can change the default spin.

IDisposable:
Testing with all the other PRs merged in shows a sweet-spot at 100ms for the default spin

JPVenson:
Why does this new named variable add opposed to using the dynamic setter? You can already set that variable if your setup requires it, otherwise this needs to be tested broadly or we run into a "works better for _me_" situation

IDisposable:
Close it if we don't want it, no worries.

PR #17558, opened 2026-08-06
Switch SQLite connection Cache to Private

IDisposable:
When SQLite's cache mode is set to Default, any settings that another consumer sets will pollute the process-global settings. This allows plugins to change the behavior of the Jellyfin core cache.

**Changes**

Change the SqliteCacheMode to Private from Default

**Code assistance**

None

**Issues**

Lock and cache corruptions when plugins like Playback Reports set their own options


IDisposable:
See this comment for context https://github.com/jellyfin/jellyfin/pull/17531#issuecomment-5188409660

Shadowghost (approved):

crobibero (approved):

Issue #15354, opened 2025-11-02
Enable moving location of jellyfin.db


twsouthwick:
I run jellyfin in a k3s cluster and use iscsi for the database. However, everything else I prefer to use nfs stores to go to my NAS. Currently, the data directory for jellyfin is where the database is as well as things like trickplay, subtitles, etc. Since iscsi shares need to be provisioned to a specific size, I've hit out of memory as trickplays get generated. To fix this, I'd like to move the database to its own directory so I can have a specific store for just the db.

anultravioletaurora:
@twsouthwick just out of curiosity are you using [Longhorn](https://longhorn.io/) for your iSCSI distributed storage? I only ask because Longhorn supports expanding persistent volumes after creation

twsouthwick:
Not right now - I currently use https://github.com/hpe-storage/truenas-csp. It supports expanding, but it grows quite large and I'd rather have the non-db files (i.e. trickplay, subtitles, etc) stored on my NFS shares

IgorAngelini:
I would also like this, I'm running on ZFS and would be nice if I could create a dataset with recordsize of 64K and set the sqlite page size to 65536. Not nearly as damaging as OP's issue, but I thought it would add to the case.

JPVenson:
never put your JF db on a network share! This has caused a lot of issues in the past and i highly discourage anyone from doing that.

PR #17565, opened 2026-08-06
Use DataSource for sqlite db path

twsouthwick:
#15354 added support to move where the db is stored, but there are some hard coded things in the file still for the original path. Tried it out on  my local deployment and hit some issues with migrations. This change will access the DbContext.DataSource to find the actual path instead of hard coding it


Shadowghost (approved):
Maybe even better if we cache the path in Initialise instead of going through a DbContext

twsouthwick:
I considered that, but then it's a separate field and the DbContext has it as part of it so it felt redundant.

PR #17568, opened 2026-08-07
Implement nested boxset rollup

Shadowghost:
If a boxset is part of a boxset and you have folding enabled, this is not handled properly

**Changes**
* Properly fold nested boxsets

**Code assistance**
Opus 5 for the tests

**Issues**

### DIFF
--- a/Emby.Server.Implementations/Collections/CollectionManager.cs
+++ b/Emby.Server.Implementations/Collections/CollectionManager.cs
@@ -335,6 +335,32 @@ namespace Emby.Server.Implementations.Collections
 
             var allBoxSets = GetCollections(user).ToList();
 
+            // Build the BoxSet nesting graph.
+            var boxSetsById = new Dictionary<Guid, BoxSet>();
+            foreach (var boxSet in allBoxSets)
+            {
+                boxSetsById[boxSet.Id] = boxSet;
+            }
+
+            var childBoxSetIds = new Dictionary<Guid, List<Guid>>();
+            var nestedBoxSetIds = new HashSet<Guid>();
+            foreach (var boxSet in allBoxSets)
+            {
+                var children = new List<Guid>();
+                foreach (var child in boxSet.GetLinkedChildren(user))
+                {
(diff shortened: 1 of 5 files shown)


theguymadmax (approved):

IDisposable (approved):

Emby.Server.Implementations/Collections/CollectionManager.cs
```diff
@@ -389,5 +415,43 @@ public IEnumerable<BaseItem> CollapseItemsWithinBoxSets(IEnumerable<BaseItem> it
 
             return results.Values;
         }
+
+        /// <summary>
+        /// Determines whether <paramref name="itemId"/> is a direct member of <paramref name="boxSet"/>
+        /// or of any BoxSet nested within it. <paramref name="visited"/> guards against cycles.
+        /// </summary>
+        private static bool BoxSetContainsItem(
```
IDisposable: Wouldn't it be cheaper to build a  flattened boxsetChildren which is  dictionary index of boxsetId -> Set<childrenId> up above since we're iterating all children of the boxset instead of just skipping the plain items? We don't care about the hierarchy for this question, just the flattened itemIds for all children in the set.

Emby.Server.Implementations/Collections/CollectionManager.cs
```diff
@@ -336,16 +336,42 @@ public IEnumerable<BaseItem> CollapseItemsWithinBoxSets(IEnumerable<BaseItem> it
...
+            foreach (var boxSet in allBoxSets)
+            {
+                var children = new List<Guid>();
+                foreach (var child in boxSet.GetLinkedChildren(user))
+                {
+                    if (child is BoxSet && boxSetsById.ContainsKey(child.Id))
+                    {
+                        children.Add(child.Id);
+                        nestedBoxSetIds.Add(child.Id);
+                    }
```
IDisposable, line 391: Here we could insert into the flattened-per-setId Set of all _leaf_ itemIds

IDisposable (approved):

